PR #4727, opened 2020-03-13
ENH: Add command line url handling and DICOM import from DICOMweb url

Sunderlandkyl:
Adds the urlReceived signal to qSlicerCoreApplication that is emitted when a url beginning with the "slicer://" protocol is processed from the command line.
This also adds handling for "slicer://viewer/..." urls, which will import the specified dataset using DICOMweb.


lassoan:
Thank you, I've pushed a commit with two minor fixes, please have a look and let me know if you find them OK.

Sunderlandkyl:
Thanks, seems good to me!

PR #4728, opened 2020-03-13
Modernize first pass

hjmjohnson:
Apply common C++11 preferred coding approaches.

### DIFF
--- a/Base/QTCLI/vtkSlicerCLIModuleLogic.cxx
+++ b/Base/QTCLI/vtkSlicerCLIModuleLogic.cxx
@@ -1241,18 +1241,18 @@ void vtkSlicerCLIModuleLogic::ApplyTask(void *clientdata)
   // Add a command line flag for the process information structure
   if ( commandType == SharedObjectModule )
     {
-    commandLineAsString.push_back( "--processinformationaddress" );
+    commandLineAsString.emplace_back("--processinformationaddress");
 
     char tname[256];
     sprintf(tname, "%p", node0->GetModuleDescription().GetProcessInformation());
 
-    commandLineAsString.push_back( tname );
+    commandLineAsString.emplace_back(tname);
     }
 
   // Add a command line flag for a file of return types
   if (node0->GetModuleDescription().HasReturnParameters())
     {
-    commandLineAsString.push_back( "--returnparameterfile" );
+    commandLineAsString.emplace_back("--returnparameterfile");
 
     std::ostringstream pidString;
 #ifdef _WIN32
--- a/Modules/Loadable/Texts/qSlicerTextsReader.h
+++ b/Modules/Loadable/Texts/qSlicerTextsReader.h
@@ -32,13 +32,13 @@ class qSlicerTextsReader : public qSlicerFileReader
 public:
   typedef qSlicerFileReader Superclass;
   qSlicerTextsReader(QObject* parent = 0);
-  virtual ~qSlicerTextsReader();
+  ~qSlicerTextsReader() override;
 
-  virtual QString description() const;
-  virtual IOFileType fileType() const;
-  virtual QStringList extensions() const;
+  QString description() const override;
+  IOFileType fileType() const override;
+  QStringList extensions() const override;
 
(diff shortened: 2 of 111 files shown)


Modules/Loadable/Texts/qSlicerTextsReader.h
```diff
@@ -32,13 +32,13 @@ class qSlicerTextsReader : public qSlicerFileReader
 public:
   typedef qSlicerFileReader Superclass;
   qSlicerTextsReader(QObject* parent = 0);
-  virtual ~qSlicerTextsReader();
+  ~qSlicerTextsReader() override override;
```
jcfr: Look like the keyword is duplicated

Modules/Loadable/VolumeRendering/MRML/vtkMRMLShaderPropertyNode.h
```diff
@@ -58,28 +58,28 @@ class VTK_SLICER_VOLUMERENDERING_MODULE_MRML_EXPORT vtkMRMLShaderPropertyNode
   //--------------------------------------------------------------------------
   // MRMLNode methods
   //--------------------------------------------------------------------------
-  virtual vtkMRMLNode* CreateNodeInstance() VTK_OVERRIDE;
+  vtkMRMLNode* CreateNodeInstance() VTK_OVERRIDE;
```
jcfr: `VTK_OVERRIDE` should probably be removed and replaced with `override`

Modules/Loadable/VolumeRendering/qSlicerShaderPropertyReader.h
```diff
@@ -38,17 +38,17 @@ class qSlicerShaderPropertyReader
   typedef qSlicerFileReader Superclass;
   qSlicerShaderPropertyReader(QObject* parent = nullptr);
   qSlicerShaderPropertyReader(vtkSlicerVolumeRenderingLogic* logic, QObject* parent = nullptr);
-  virtual ~qSlicerShaderPropertyReader();
+  ~qSlicerShaderPropertyReader() override override;
```
jcfr: duplicate

Base/QTCLI/vtkSlicerCLIModuleLogic.cxx
```diff
@@ -1241,18 +1241,18 @@ void vtkSlicerCLIModuleLogic::ApplyTask(void *clientdata)
   // Add a command line flag for the process information structure
   if ( commandType == SharedObjectModule )
     {
-    commandLineAsString.push_back( "--processinformationaddress" );
+    commandLineAsString.emplace_back("--processinformationaddress" );
```
jcfr: while updating the file, remove extra space at the end

```
commandLineAsString.emplace_back("--processinformationaddress" );
```

->

```
commandLineAsString.emplace_back("--processinformationaddress");
```

jcfr:
Thanks @hjmjohnson , it looks like commit https://github.com/Slicer/Slicer/pull/4728/commits/9c0164d2474779478938cb1256cb3b0675f789c8 introduce `override override` which is letter fixed in the unrelated commit https://github.com/Slicer/Slicer/pull/4728/commits/6f246738d580c684e751ced4ce30bd5d852e1263.

Could you updates the commit and force push the topic ? :pray:

hjmjohnson:
@jcfr All requested changes have been made.  Local builds completed successfully.

jcfr:
Thanks for updating the commits 🙏 

> re: emplace_back

Since we are not using `std::move` in our object constructors, do you still think it is useful to introduce this change at that time ? 

See https://en.cppreference.com/w/cpp/container/vector/emplace_back

Also after this change, will we have a mix of `push_back` and `emplace_back`, or will `emplace_back` be used everywhere ?

hjmjohnson:
I do think it is useful.  we may not take full advantage of this yet for Slicer objects, it does work for objects that follow the "rule of zero", or other objects.  I see no downsides to using emplace_back in any cases.  Only upsides for object types that support move semantics.

PR #4734, opened 2020-03-16
Updating Python Packages to latest - March 2020 

jamesobutler:
Python packages were last updated November 10th, 2019. Outdated python packages were discovered using the below code snippet and then versions, hashes and dependencies were determined and updated.
```powershell
./SlicerPython.exe -m pip list --outdated
```

Tests were run locally on Windows with the same currently failing tests on the dashboard also only being the ones to fail.

@pieper  This includes updating the pydicom version as you indicated you wanted in https://github.com/Slicer/SlicerGitSVNArchive/pull/1254#issuecomment-597660126.


pieper:
Thank you @jamesobutler !

PR #4737, opened 2020-03-17
ENH: Set imported segmentation names from label value

### LINKED ISSUE #4733: Segmentation loaded from volume has non-descriptive segment names
## Environment
- Slicer version: Slicer-4.11.0 r28813
- Operating system: Linux

## Steps to reproduce the issue:

Loaded the attached volume file as segmentation.
[4962.Oto-label.nrrd.gz](https://github.com/Slicer/Slicer/files/4340012/4962.Oto-label.nrrd.gz)

## What was expected to happen?

The segment names are expected to contain the original label values (1, 2, 4, 8).

## What happened instead?

The segmentation was loaded with segment names Segment, Segment_1, Segment_2, .... See the screen capture below.

image

The segment names did not indicate which original label value they correspond to, which made it difficult to interpret the segmentation.

Sunderlandkyl:
Previously imported segmentations used linearly increasing identifiers when importing (Segment, Segment_1, Segment_2, etc.). Now names are generated using the label value of the segment (Segment_1, Segment_4, Segment_5, etc for label values 1, 4, and 5).

Fixes #4733

### DIFF
--- a/Libs/MRML/Core/vtkMRMLSegmentationStorageNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLSegmentationStorageNode.cxx
@@ -857,7 +857,9 @@ int vtkMRMLSegmentationStorageNode::ReadBinaryLabelmapRepresentation(vtkMRMLSegm
       }
     else
       {
-      currentSegmentID = segmentation->GenerateUniqueSegmentID("Segment");
+      std::stringstream segmentNameSS;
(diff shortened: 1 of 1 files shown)


Sunderlandkyl:
It would be called Segment_3 (and it would also have a label value of 3).

lassoan:
Sounds good, thank you.

PR #4738, opened 2020-03-17
ENH: Improve Slicer Data Bundle DICOM export and import

cpinter:
- Slicer Data Bundle exporter can save to user defined path
- Slicer Data Bundle exporter can use custom tags: study description and series description as separate member variables (because they have defaults in the exporter), and an optional tags map to specify any tag
- Allow Slicer Data Bundle exporter to export custom image to the Secondary Capture series
- Add option to Slicer Data Bundle exporter for copying SDB DICOM file to database or not when importing
- Slicer Data Bundle DICOM plugin now sets up subject hierarchy as all other DICOM plugins when importing. Important so that patient/study/series data are stored somewhere in the scene. The series node is the first volume or (non-slice) model or markup
- Allow CreateDICOMSeries module to export Secondary Capture


lassoan:
Very nice!

cpinter:
Thank you for the review! Much easier to merge it now isn't it? :)

PR #4739, opened 2020-03-17
ENH: Add JPG support to Screen Capture

cpinter:
DCMTK Secondary Capture only supports JPG files, so instead of converting within the application, JPG support has been added to the Screen Capture module


cpinter:
Thanks, Andras!

PR #4744, opened 2020-03-18
ENH: Replace old "scripted" module template by the new "scripteddesigner" template

lassoan:
We had two scripted module templates: scripted and scripteddesigner. The "scripted" template had several limitations: GUI widgets are created from code, parameter node is not used (user choices are not saved in the scene), logic is not created in the widget (so it is not accessible from other modules).

Removed the old scripted module template to make things simpler (and added an example of a dynamically added node selector widget).

See related discussion: https://discourse.slicer.org/t/replace-scripted-module-template-by-scripteddesigner/10736


Utilities/Templates/Modules/Scripted/Resources/UI/TemplateKey.ui
```diff
@@ -178,7 +181,7 @@
       <item row="0" column="1">
        <widget class="QCheckBox" name="invertOutputCheckBox">
         <property name="toolTip">
-         <string>If checked, take screen shots for tutorials. Use Save Data to write them to disk.</string>
+         <string>v</string>
```
jcfr: Look like the tooltip was removed by mistake

lassoan:
What does this SlicerExtensionTemplatesGenerator.cmake do? It seems that the module type list was not up-to-date before but ExtensionWizard still worked fine.

jcfr:
> What does this SlicerExtensionTemplatesGenerator.cmake do?

It re-generates the examples found in https://github.com/Slicer/Slicer/tree/master/Extensions/Testing using the wizard. Building the `SlicerGenerateExtensionTemplates` target allows to do so.

lassoan:
Thanks for the reviews and suggestions, I've fixed them (and will follow up with a pull request that adds a Python scripted CLI module template).

PR #4745, opened 2020-03-19
Fix extension config with depends

jcfr:


jcfr:
> Do we still need to write "NA" if no dependencies?

We could probably revisit this and update the code based to ignore empty list element instead of the string `"NA"`

I suggest we tackle this in a different PR as this on is fixing an issue.

lassoan:
There are two commits with contradicting titles - do we accept both space and ; as separator or only ;?

In general, I prefer having fewer commits (all commits related to a single set of changes squashed into one), with more detailed description, which describes _why_ the changes were made (a short summary of _what_ was changed is useful, too, but file diff shows the exact changes anyway).

jcfr:
> do we accept both space and ; as separator or only ;?

There are extensions that are still providing dependencies as strings like `"ExtA ExtB ExtC"` or `"ExtA;ExtB;ExtC"`, and other one as a list like `ExtA ExtB ExtC`. 

The first commit ensures all existing extensions can successfully be configured.


The second commit removes the documented support for having list of dependency specified as string because I wanted to avoid making the SuperBuild template more complex. Indeed, this means I would need to include lines like these:

```
  # If needed, convert to a list
  list(LENGTH EXTENSION_DEPENDS _count)
  if(_count EQUAL 1)
    string(REPLACE " " ";" EXTENSION_DEPENDS ${EXTENSION_DEPENDS})
  endif()
```

here: 

https://github.com/Slicer/Slicer/blob/a3662ddc2de823c32c66891e344dd713267aa674/Utilities/Templates/Extensions/SuperBuild/SuperBuild.cmake#L13-L14

jcfr:
>  to understand have a single commit where you described what you did and why

Agreed. Thanks for the feedback.

@lassoan Let me know what you think.

jcfr:
@lassoan When you have a chance, would be great if you could approve.

Utilities/Templates/Extensions/Default/CMakeLists.txt
```diff
@@ -10,7 +10,7 @@ set(EXTENSION_CONTRIBUTORS "John Doe (AnyWare Corp.)")
 set(EXTENSION_DESCRIPTION "This is an example of a simple extension")
 set(EXTENSION_ICONURL "http://www.example.com/Slicer/Extensions/TemplateKey.png")
 set(EXTENSION_SCREENSHOTURLS "http://www.example.com/Slicer/Extensions/TemplateKey/Screenshots/1.png")
-set(EXTENSION_DEPENDS "NA") # Specified as a space separated string, a list or 'NA' if any
+set(EXTENSION_DEPENDS "NA") # Specified as a list or 'NA' if any
```
lassoan: This would be more clear:

... or "NA" if the extension does not depend on other extensions

... or "NA" if no dependencies

PR #4746, opened 2020-03-19
ENH: Add new plane markup type

Sunderlandkyl:
This commit adds a new plane type of markup. Planes are defined by 3 control points. The first control point defines the plane origin. An offset can be specified which transforms the origin and vectors of the plane using a 4x4 matrix.

Shift+Left click and drag translates the offset of the plane in 3D and slice views.
The vtkMRMLMarkupsPlaneNode class also includes some convenience functions for generating a plane without placing the markup points (SetOrigin(), SetNormal(), and SetPlaneAxes()) that can be specified in either local or world coordinates.

Future work will include more advanced plane widgets/interactions (movement/rotation along specific axis, adjusting size, etc).

image


lassoan:
Thanks a lot, I'll review and test this tonight.

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsPlaneNode.cxx
```diff
@@ -0,0 +1,659 @@
...
+  if (transformNode)
+    {
+    // Get transform
+    vtkNew<vtkGeneralTransform> transformToWorld;
+    transformNode->GetTransformFromWorld(transformToWorld.GetPointer());
+
+    // Convert coordinates
+    double origin[3] = { 0 };
+    this->GetOriginWorld(origin);
+    transformToWorld->TransformVectorAtPoint(origin, normal, normal);
```
lassoan: I've tried to apply a non-linear transform to a plane and found it confusing that the 3 control points are no longer on the plane. I think a more intuitive way to apply a non-linear transform would be to apply the transform to the 3 control points and then fit a plane on those.

Sunderlandkyl:
Here's a visualization with 0.2-0.5 opacity:
image

lassoan:
This opacity looks very nice.

Sunderlandkyl:
@lassoan I believe that I've addressed the comments.

lassoan:
Once Slicer download page is fixed and users can download from there a version that contains this feature, write a topic on the Slicer forum, asking people to test this and give feedback. Thank you!

sjh26:
The write/read/print macros appear to be buggy on mac / Linux:

http://slicer.cdash.org/viewBuildError.php?buildid=1860686

lassoan:
I'll push a fix.

PR #4760, opened 2020-03-23
ENH: Add pixel size to segment hollow/margin size preview

Sunderlandkyl:


Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorHollowEffect.py
```diff
@@ -110,14 +110,14 @@ def updateGUIFromMRML(self):
       selectedSegmentLabelmapSpacing = selectedSegmentLabelmap.GetSpacing()
       shellThicknessPixel = self.getShellThicknessPixel()
       if shellThicknessPixel[0] < 1 or shellThicknessPixel[1] < 1 or shellThicknessPixel[2] < 1:
-        self.shellThicknessPixel.text = "Not feasible at current resolution."
+        self.shellThicknessLabel.text = "Not feasible at current resolution."
         self.applyButton.setEnabled(False)
       else:
         thicknessMm = self.getShellThicknessMm()
-        self.shellThicknessPixel.text = "Actual: {0} x {1} x {2} mm.".format(*thicknessMm)
+        self.shellThicknessLabel.text = "Actual: {0} x {1} x {2} mm ({3}x{4}x{5} pixels)".format(*thicknessMm, *shellThicknessPixel)
```
jcfr: Assuming `pixel` is a unit, it should probably be in singular form. See https://physics.stackexchange.com/a/384189

@pieper  What do you think ?

pieper: Yes, as used here it would make sense to use the singular "pixel" when used as a unit.

Also, if we want to stick with the style guide, it would be `thicknessMM` (although I'll admit it looks just as odd as `thicknessMm` will probably annoy the mathematician/physicists just as much).
https://www.slicer.org/wiki/Documentation/Nightly/Developers/Style_Guide#Naming

Sunderlandkyl:
I've updated pixels ->pixel and Mm -> MM.

It's worth noting that I've changed the editor effect properties as well (ShellThicknessMm ->ShellThicknessMM, etc).  These parameters will be reset to default if they were saved in a previous scene.

jcfr:
> It's worth noting that I've changed the editor effect properties as well (ShellThicknessMm ->ShellThicknessMM, etc). These parameters will be reset to default if they were saved in a previous scene.

In anticipation of question on discourse forum, should we add an entry to https://www.slicer.org/wiki/Documentation/Nightly/Developers/Tutorials/MigrationGuide#Segmentations ?

lassoan:
I think VTK's style of inconsistent camelcase is really bad (not camelcasing acronyms), as it makes some names very hard to read. I would still follow the recommendation for method names (for sake of consistency), but not for other things, such as parameter names. So, I would prefer keeping ShellThicknessMm. If there is a consensus that we should change capitalization of these parameter names then we have to make sure that all comparisons are case-insensitive.

By the way, the value is not actually millimeter but whatever is the length unit in the scene. If there is a good name suggestion for how to call this "physical length" unit then we could gradually switch to that naming convention.

Sunderlandkyl:
@lassoan OK, I've reverted the parameter case change.

jcfr:
> OK, I've reverted the parameter case change.

Looks like the `MM` version is still used

Sunderlandkyl:
> Looks like the `MM` version is still used

I just reverted the case for the MRML parameters.

jcfr:
> I just reverted the case for the MRML parameters.

Gotcha.

PR #4763, opened 2020-03-24
ENH: Add option to build Slicer using system vtkAddon

Sunderlandkyl:
Adds the Slicer_USE_SYSTEM_vtkAddon option, which is OFF by default.
This allows extensions with dependencies that rely on vtkAddon to be built in custom apps.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -272,6 +272,9 @@ option(Slicer_USE_FOLDERS "Organize build targets into folders" ON)
 mark_as_superbuild(Slicer_USE_FOLDERS)
 mark_as_advanced(Slicer_USE_FOLDERS)
 
+option(Slicer_USE_SYSTEM_vtkAddon "Build Slicer using system vtkAddon" OFF)
+mark_as_superbuild(Slicer_USE_SYSTEM_vtkAddon:BOOL)
+
 #-----------------------------------------------------------------------------
 # External projects related options
 #-----------------------------------------------------------------------------
@@ -859,8 +862,10 @@ set(SlicerExecutionModel_EXTRA_EXECUTABLE_TARGET_LIBRARIES "" CACHE INTERNAL "Sl
 # Set Slicer builtin libraries *_DIR variables
 #-----------------------------------------------------------------------------
 set(GenerateLM_DIR ${Slicer_BINARY_DIR}/Libs/GenerateLM)
-set(vtkAddon_DIR ${Slicer_BINARY_DIR}/Libs/vtkAddon)
 set(vtkITK_DIR ${Slicer_BINARY_DIR}/Libs/vtkITK)
+if (NOT Slicer_USE_SYSTEM_vtkAddon)
+  set(vtkAddon_DIR ${Slicer_BINARY_DIR}/Libs/vtkAddon)
+endif()
 
 #-----------------------------------------------------------------------------
 # Set COVERAGE_{C,CXX}_FLAGS variables
--- a/Libs/CMakeLists.txt
+++ b/Libs/CMakeLists.txt
@@ -19,8 +19,13 @@ set(SlicerExecutionModel_EXTRA_EXECUTABLE_TARGET_LIBRARIES
   CACHE INTERNAL "SlicerExecutionModel extra executable target libraries" FORCE
   )
 
+if (NOT Slicer_USE_SYSTEM_vtkAddon)
+  list(APPEND dirs
+    vtkAddon
+    )
+endif()
+
 list(APPEND dirs
-  vtkAddon
   vtkTeem
   vtkITK
   vtkSegmentationCore
@@ -128,7 +133,6 @@ set(MRML_LIBRARIES ${_mrml_libraries} CACHE INTERNAL "MRML libraries" FORCE)
 # Set variable Slicer_Libs_VTK_WRAPPED_LIBRARIES
 #-----------------------------------------------------------------------------
 set(_vtk_wrapped_libraries
-  vtkAddon
   vtkSegmentationCore
   vtkTeem
   vtkITK
@@ -136,6 +140,12 @@ set(_vtk_wrapped_libraries
   MRMLCore
   MRMLLogic
   )
+
+if (NOT Slicer_USE_SYSTEM_vtkAddon)
+  list(APPEND _vtk_wrapped_libraries
+    vtkAddon
+    )
+endif()
 if(Slicer_USE_PYTHONQT AND VTK_WRAP_PYTHON)
   list(APPEND _vtk_wrapped_libraries MRMLDisplayableManager)
 endif()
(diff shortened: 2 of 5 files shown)


Sunderlandkyl:
This change shouldn't affect the default behavior of Slicer.

jcfr:
Are the source only available in  https://github.com/Slicer/Slicer/tree/master/Libs/vtkAddon or is there an other repo ? 

If not, should we extract the source into its own source repository ?

Sunderlandkyl:
I've mirrored the vtkAddon dir into [IGSIO](https://github.com/IGSIO/IGSIO/tree/master/vtkAddon), so that we can use the vtkStreamingVolumeFrame/Codec classes in other projects as well.

If possible, I think that moving vtkAddon to its own repo would be useful.

jcfr:
>If possible, I think that moving vtkAddon to its own repo would be useful.

I just extracted the history and pushed it into `Slicer/vtkAddons` repository.  See https://github.com/Slicer/vtkAddon

@lassoan @pieper  What do you think ? 

Should we organize files organized like this ?

```
<root>/CMakeLists.txt
<root>/src/...
```


For reference, here are the commands used to extract the history:

```
# Clone
git clone git://github.com/Slicer/Slicer /tmp/Slicer-for-vtkAddon

cd /tmp/Slicer-for-vtkAddon

# Extract history specific to vtkAddon
git-rocket-filter --branch vtkAddon \
  --keep Libs/vtkAddon/ \
  --detach

# Remove empty directories
find . -depth -empty -not -path '*/\.*' -delete

# Publish repo
git remote add vtkAddon git@github.com:Slicer/vtkAddon

# Clone
git clone git@github.com:Slicer/vtkAddon /tmp/vtkAddon

cd /tmp/vtkAddon


# Move files from 'Libs/vtkAddon' directory to the root
git filter-branch --force --tree-filter \
    'mv Libs/vtkAddon/* ./ > /dev/null 2>&1 || true' \
    --tag-name-filter cat -- --all

# Re-publish
git push origin master --force
```

pieper:
Makes good sense to me - thanks 👍

jcfr:
Thanks for the feedback.

For now,  I suggest we include vtkAddon as a [remote module](https://www.slicer.org/wiki/Documentation/Nightly/Developers/Build_system/Remote_Module).

This will keep things simple and avoid us copying the `vtkMacroKitPythonWrap` required to support python wrapping of the module. (this macro is currently found  in Slicer source tree `Slicer/CMake` directory)

Moving forward with the transition to the latest VTK, we could probably revisit this.

Also to support integrated in IGSIO, we will introduce an option named `vtkAddon_WRAP_PYTHON` set to `FALSE` by default. This will be consistent with what @Sunderlandkyl  did:

https://github.com/IGSIO/IGSIO/blob/be28dbad32ea42bbe30dcd9753f017d53b3cdbb5/vtkAddon/CMakeLists.txt#L169-L170

jcfr:
> This allows extensions with dependencies that rely on vtkAddon to be built in custom apps.

@Sunderlandkyl Out of curiosity, what prevent custom app from using the vtkAddon pre-build by Slicer ?

Sunderlandkyl:
@jcfr It resulted in a circular dependency. When building a custom app with SlicerIGSIO included:
- Slicer (w/ SlicerIGSIO) -> Depends on IGSIO and compiles vtkAddon
- IGSIO -> Depends on vtkAddon

jcfr:
Since we could potentially have multiple extension depending on vtkAddon, I think it would make more sens to add the option option `vtkAddon_DIR` to IGSIO

CMakeLists.txt
```diff
@@ -272,6 +272,9 @@ option(Slicer_USE_FOLDERS "Organize build targets into folders" ON)
 mark_as_superbuild(Slicer_USE_FOLDERS)
 mark_as_advanced(Slicer_USE_FOLDERS)
 
+option(Slicer_USE_SYSTEM_vtkAddon "Build Slicer using system vtkAddon" OFF)
+mark_as_superbuild(Slicer_USE_SYSTEM_vtkAddon:BOOL)
```
jcfr: To be very specific, we should support `vtkAddon_DIR` instead of `Slicer_USE_SYSTEM_vtkAddon`

CMakeLists.txt
```diff
@@ -859,8 +862,10 @@ set(SlicerExecutionModel_EXTRA_EXECUTABLE_TARGET_LIBRARIES "" CACHE INTERNAL "Sl
 # Set Slicer builtin libraries *_DIR variables
 #-----------------------------------------------------------------------------
 set(GenerateLM_DIR ${Slicer_BINARY_DIR}/Libs/GenerateLM)
-set(vtkAddon_DIR ${Slicer_BINARY_DIR}/Libs/vtkAddon)
 set(vtkITK_DIR ${Slicer_BINARY_DIR}/Libs/vtkITK)
+if (NOT Slicer_USE_SYSTEM_vtkAddon)
```
jcfr:
```
if (NOT DEFINED vtkAddon_DIR)
```

Libs/MRML/Core/CMakeLists.txt
```diff
@@ -60,6 +60,13 @@ if(MRML_USE_vtkTeem)
   include(${Teem_USE_FILE})
 endif()
 
+#
+# vtkAddon
+#
+if (Slicer_USE_SYSTEM_vtkAddon)
+  find_package(vtkAddon REQUIRED)
+endif()
```
jcfr: The call to `find_package(vtkAddon REQUIRED)` should always be valid

Libs/CMakeLists.txt
```diff
@@ -19,8 +19,13 @@ set(SlicerExecutionModel_EXTRA_EXECUTABLE_TARGET_LIBRARIES
   CACHE INTERNAL "SlicerExecutionModel extra executable target libraries" FORCE
   )
 
+if (NOT Slicer_USE_SYSTEM_vtkAddon)
```
Sunderlandkyl: @jcfr Correct me if I'm wrong, but I'm not sure if we could rely exclusively on vtkAddon_DIR, since it would be already defined by this point.

Slicer_USE_SYSTEM_vtkAddon is still somewhat consistent with the use of Slicer_USE_SYSTEM_VTK/ITK/CTK, etc.

jcfr: VTK/ITK/CTK .. support both `<proj>_USE_SYSTEM` and `<proj>_DIR` options. Setting `<proj>_DIR` means we explicitly provide a location where a config is expected. `<proj>_USE_SYSTEM` is set to ON when `proj` found in any location should be used.

In the case of `vtkAddons`, at that stage we do not support the "use system" approach (e.g there are no `apt-get install vtkAddon` yet)

jcfr: But not worry,  I will revisit the approach while integrating vtkAddon remote module

Sunderlandkyl: Ok, sounds good. Thanks for the clarification!

PR #4767, opened 2020-03-25
Fix miscallenous warnings

jcfr:
An additional warning is being fixed in ITK (see https://github.com/InsightSoftwareConsortium/ITK/pull/1727) and will be backported after the PR is integrated upstream.

### DIFF
--- a/Modules/CLI/ResampleDTIVolume/itkDiffusionTensor3DFSAffineTransform.txx
+++ b/Modules/CLI/ResampleDTIVolume/itkDiffusionTensor3DFSAffineTransform.txx
@@ -28,7 +28,7 @@ DiffusionTensor3DFSAffineTransform<TData>
 
   vnl_matrix_fixed<double, 3, 3> M( 3, 3 );
   M = matrix.GetVnlMatrix();
-  vnl_real_eigensystem             eig( M );
+  vnl_real_eigensystem             eig( M.as_matrix() );
   vnl_matrix_fixed<std::complex<double>, 3, 3 > D( 3, 3 );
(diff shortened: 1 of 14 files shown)


Modules/CLI/ResampleDTIVolume/itkDiffusionTensor3DFSAffineTransform.txx
```diff
@@ -28,15 +28,15 @@ DiffusionTensor3DFSAffineTransform<TData>
 
   vnl_matrix_fixed<double, 3, 3> M( 3, 3 );
   M = matrix.GetVnlMatrix();
-  vnl_real_eigensystem             eig( M );
+  vnl_real_eigensystem             eig( M.as_ref() );
```
jcfr: @hjmjohnson  Would appreciate if you could have a quick look at this patch and also this one https://github.com/Slicer/BRAINSTools/commit/d87ea8f357622c27abc6a2162b4623abe782b995

hjmjohnson: @jcfr This looks fine.  If you can match types, there can be some efficiencies using move constructors.  For a case like this, it probably does not make any significant difference using  `as_ref()` vs `as_matrix()`.

Hans

jcfr: Thanks for the review.

If recall, I tried to use `.as_matrix()` and it failed. I re-compiling locally and will post an update here.

PR #4776, opened 2020-03-26
ENH: Remove redundant reference to volume node in volume rendering display node

### LINKED ISSUE #2628: VR displayable manager should not show display node that are not referenced by displayable node
_This issue was created automatically from an original [Mantis Issue](https://mantisarchive.slicer.org/view.php?id=2628). Further discussion may take place here._

lassoan:
vtkMRMLVolumeRenderingDisplayNode::SetAndObserveVolumeNodeID method was removed, as display node base class already maintains a pointer to the displayed (volume) node.

Updated migration guide (https://www.slicer.org/wiki/Documentation/Nightly/Developers/Tutorials/MigrationGuide/Slicer#Volume_rendering).

Fixes #2628.


lassoan:
> Since roles have been removed, should mrml scene used in test be updated as well ?

Probably not needed, but I'm running the tests now.

> Finally, may be worth checking this page as well: https://www.slicer.org/wiki/Documentation/Nightly/Modules/VolumeRendering#Information_for_Developers

I've updated the relevant example code now. (it was mostly OK already, just the order of operations were not ideal).

lassoan:
All volume rendering related tests passed and manual testing shows that everything seems to work well.

Issue #4685, opened 2020-03-13
Draw effect does not take parent transform into account


slicerbot:
_This issue was created automatically from an original [Mantis Issue](https://mantisarchive.slicer.org/view.php?id=4685). Further discussion may take place here._

lassoan:
Could you please check how difficult would it be to fix this?

Sunderlandkyl:
Still looking at a fix, but this issue applies to both the draw and level-tracing effects.

lassoan:
Fixed.

PR #4777, opened 2020-03-27
BUG: Apply segmentation parent transform to draw effect

Sunderlandkyl:
The polydata for the draw effect was rasterized in the world coordinate system.
If the segmentation node was under a transform node, the transform would not be taken into account when modifying the segment.

This commit transforms the rasterized image to the same coordinate system as the segmentation before appending it to the modifier labelmap.

re #4685

### DIFF
--- a/Modules/Loadable/Segmentations/EditorEffects/qSlicerSegmentEditorAbstractLabelEffect.cxx
+++ b/Modules/Loadable/Segmentations/EditorEffects/qSlicerSegmentEditorAbstractLabelEffect.cxx
@@ -120,12 +120,30 @@ void qSlicerSegmentEditorAbstractLabelEffect::updateMRMLFromGUI()
 }
 
 //-----------------------------------------------------------------------------
-void qSlicerSegmentEditorAbstractLabelEffect::appendPolyMask(vtkOrientedImageData* input, vtkPolyData* polyData, qMRMLSliceWidget* sliceWidget)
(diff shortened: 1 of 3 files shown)


Modules/Loadable/Segmentations/EditorEffects/qSlicerSegmentEditorAbstractLabelEffect.cxx
```diff
@@ -120,13 +120,25 @@ void qSlicerSegmentEditorAbstractLabelEffect::updateMRMLFromGUI()
...
   // Rasterize a poly data onto the input image into the slice view
   // - Points are specified in current XY space
   vtkSmartPointer<vtkOrientedImageData> polyMaskImage = vtkSmartPointer<vtkOrientedImageData>::New();
   qSlicerSegmentEditorAbstractLabelEffect::createMaskImageFromPolyData(polyData, polyMaskImage, sliceWidget);
 
+  // If the segmentation has a parent transform, apply a transformation to polyMaskImage to move it to the same coordinate system.
+  if (segmentationNode && segmentationNode->GetParentTransformNode())
+    {
+    vtkNew<vtkMatrix4x4> worldToSegmentation;
+    segmentationNode->GetParentTransformNode()->GetMatrixTransformFromWorld(worldToSegmentation);
```
lassoan: Check if transform to world is linear (transformNode->IsTransformToWorldLinear()) and if not then log a warning message and ignore the transform (this is what we do in qSlicerSegmentEditorAbstractLabelEffect::imageToWorldMatrix).

Sunderlandkyl: Done!

PR #4780, opened 2020-03-27
BUG: Fix crash in qMRMLSceneTransformModelTest1

lassoan:
qMRMLSceneTransformModelTest1 crashed on exit because deleted items are attempted to be accessed.
Deleting tree widgets before the application quits solved the issue.


lassoan:
@cpinter I've also noticed that if I run this test in interactive mode (-I) then it crashes immediately after almost any drag-and-drop operation. Are you aware of this issue? Do you have any idea why this is happening?

PR #4771, opened 2020-03-25
WIP: Migrate documentation to source repository

jcfr:
See issue #4751

For now, readthedocs will automatically regenerate the documentation each time the branch [Slicer/Slicer@4751-migrate-documentation-to-source-repository](https://github.com/Slicer/Slicer/tree/4751-migrate-documentation-to-source-repository) is updated.

Initial set of changes in this branch was ported over from https://github.com/Slicer/SlicerGitSVNArchive/pull/686

Pending issues:
* [x] Fix integration of external images used in readthedocs
* [x] Update Copyright year
* [ ] Port over existing module wiki page using pandoc
* [ ] Add recommendation for compressing images + enable PR hook for checking size of files

Questions:
* Should we store figure binary sources (e.g pptx used in this commit: https://github.com/Slicer/Slicer/pull/4771/commits/ffd89df68c67c7ba584defc5cccbf5de5df4e319#diff-41e1283d0b9f7e13387615bc3aff4f5d) in the repository ? 
  * @jcfr: I would prefer not. Then, where do we store them ?

PR #4783, opened 2020-03-29
Various updates to contributing doc

jamesobutler:
This updates the contributing doc with some additional changes about the usage of the Slicer forum instead of the mailing list, using github issues, and the transition of information about commit messages from the Slicer wiki to a github hosted document.

I don't believe this work has any duplication of work in #4771 as the contributing doc there is just being converted from MD to RST.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -21,6 +21,7 @@ If you are new to Slicer development and you don't have push access to the Slice
 repository, here are the steps:
 
 1. [Fork and clone](https://help.github.com/articles/fork-a-repo/) the repository.
+2. Run the developer setup script [`SetupForDevelopment.sh`](Utilities/SetupForDevelopment.sh).
 3. Create a branch.
 4. [Push](https://help.github.com/articles/pushing-to-a-remote/) the branch to your GitHub fork.
 5. Create a [Pull Request](https://github.com/Slicer/Slicer/pulls).
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -132,9 +132,9 @@ blog post.
 #### Nightly tests
 
 After changes are integrated, every evening at 10pm EST (3am UTC), Slicer build bots (aka factories)
-will build, test and package Slicer application and all its extensions on Linux, MacOSX
-and Windows. Results are published daily on [CDash](http://slicer.cdash.org/index.php?project=Slicer4)
-and developers introducing changes introducing build or test failures are notified by
+will build, test and package the Slicer application and all its extensions on Linux, MacOSX
+and Windows. Results are published daily on CDash ([Stable](http://slicer.cdash.org/index.php?project=Slicer4) & [Preview](http://slicer.cdash.org/index.php?project=SlicerPreview))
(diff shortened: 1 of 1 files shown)


jamesobutler:
CommitCheck failing because of the usage of `Doc: Commit message type` as included in the commit regarding various commit prefix types.

jcfr:
I just updated the check to accept DOC and also updated the wiki to include `DOC` as a valid prefix. 

See https://www.slicer.org/wiki/Documentation/Nightly/Developers/Style_Guide#Commits

For reference, [here](https://github.com/InsightSoftwareConsortium/ITK/blob/cb712843f93b3488b0ff316beb1ed5d6e0b6ff42/Utilities/Hooks/prepare-commit-msg#L33-L40) are the prefixes used in ITK.

CONTRIBUTING.md
```diff
@@ -132,9 +132,9 @@ blog post.
 #### Nightly tests
 
 After changes are integrated, every evening at 10pm EST (3am UTC), Slicer build bots (aka factories)
-will build, test and package Slicer application and all its extensions on Linux, MacOSX
-and Windows. Results are published daily on [CDash](http://slicer.cdash.org/index.php?project=Slicer4)
-and developers introducing changes introducing build or test failures are notified by
+will build, test and package the Slicer application and all its extensions on Linux, MacOSX
```
jcfr: MacOSX -> macOS

See https://en.wikipedia.org/wiki/MacOS

CONTRIBUTING.md
```diff
@@ -6,7 +6,7 @@ look through the [documentation](https://www.slicer.org/wiki/Documentation/Night
 do better.
 
   * Ask a question on the [Slicer forum](https://discourse.slicer.org)
-  * Submit a feature request or bug, or add to the discussion on the [Slicer issue tracker](https://issues.slicer.org/)
```
jcfr: In  documentation, I suggest we keep `https://issues.slicer.org/`

CONTRIBUTING.md
jamesobutler: Ok, I think I had kept because Slicer 4.10 and 4.11 (as of now) still supports 10.11 as the minimum which is OS X 10.11 and not macOS 10.11. Though once 4.11 uses Qt 5.12.X the minimum will be 10.12 which was a macOS version.

I had also looked into other OS X to macOS but most were CMake related which we can’t really do anything about.

jcfr: Fair point. Since we still support 10.11, that makes sense.

CONTRIBUTING.md
jamesobutler: @jcfr Did you still want this changed? I see you already merged this PR.

jcfr: That would be nice. 

Also,  would be nice to add the following at the end of the document:

```
[slicer-forum]: https://discourse.slicer.org
[slicer-issues]: https://issues.slicer.org
...
```

Then, update usage like this:

`[Slicer forum](https://discourse.slicer.org)`
->
`[Slicer forum][slicer-forum]`

PR #4784, opened 2020-03-30
ENH: Save current application version into MRML scene files

### LINKED ISSUE #4779: Save current application version to scene file
In current scene files, version attribute of MRML element is outdated:

    <MRML  version="Slicer4.4.0" userTags=""> 

We should save current application version and also revision (simple monotonously incremented integer) instead.

lassoan:
Version information (<ApplicationName> <Major>.<Minor>.<Patch> <revision>) is saved into top-level MRML element's "Version" attribute.

Fixes #4779


Modules/Loadable/Data/qSlicerSceneReader.cxx
```diff
@@ -106,17 +107,42 @@ bool qSlicerSceneReader::load(const qSlicerIO::IOProperties& properties)
...
+    QString warningMsg;
+    if (loadedApplication != currentApplication)
+      {
+      warningMsg += tr(" Scene file was saved with %1 application (current application is %2).")
+        .arg(QString::fromUtf8(loadedApplication.c_str()))
+        .arg(QString::fromUtf8(currentApplication.c_str()));
+      }
+    QVersionNumber loadedSceneVersion(loadedMajor, loadedMinor, loadedPatch);
+    QVersionNumber currentVersion(currentMajor, currentMinor, currentPatch);
+    if (loadedSceneVersion > currentVersion)
```
jcfr: Similarly to [QT_VERSION_CHECK](https://github.com/qt/qtbase/blob/e0b89899e3c505edbdece60d6a2d2368a7ef9a01/src/corelib/global/qglobal.h#L61) we could also introduce a `MRML_VERSION_CHECK` macro and then do the following:

```
if (MRML_VERSION_CHECK(loadedMajor, loadedMinor, loadedPatch) > MRML_VERSION_CHECK(currentMajor, currentMinor, currentPatch))
```

Libs/MRML/DisplayableManager/Testing/Cxx/vtkMRMLCameraDisplayableManagerTest1.cxx
```diff
@@ -621,6 +621,7 @@ int vtkMRMLCameraDisplayableManagerTest1(int argc, char* argv[])
   testHelper->AddArguments(argc, const_cast<const char **>(argv));
   vtkStdString savedScene = testHelper->GetTempDirectory();
   savedScene += "/vtkMRMLCameraDisplayableManagerTest1_saved.mrml";
+  scene->SetVersion("Slicer4.4.0"); // Force scene version to be the sam eas in the baseline scene file
```
jcfr: `sam eas` -> same as

Modules/Loadable/Data/qSlicerSceneReader.cxx
jcfr: Similarly, we could have a `SLICER_VERSION_CHECK`.

Libs/CMakeLists.txt
```diff
@@ -59,6 +59,27 @@ if(Slicer_BUILD_CLI_SUPPORT)
   set(ModuleDescriptionParser_USE_PYTHON OFF)
 endif()
 
+set(MRML_APPLICATION_NAME "${Slicer_MAIN_PROJECT_APPLICATION_NAME}")
+
+# Convert version to 0xXXYYZZ format where XX=major, YY=minor, ZZ=patch version in hexadecimal.
+# Add a leading 1 (256*256*256*1) to force having a leading 0 for major version (e.g., 04 instead of just 4)
+math(EXPR MRML_VERSION_HEX "256*256*256*1 + 256*256*${Slicer_MAIN_PROJECT_VERSION_MAJOR}
+  + 256*${Slicer_MAIN_PROJECT_VERSION_MINOR} + ${Slicer_MAIN_PROJECT_VERSION_PATCH}" OUTPUT_FORMAT HEXADECIMAL)
```
jcfr: Could also format like this:

```
math(EXPR MRML_VERSION_HEX "1 << 24 
  | ${Slicer_MAIN_PROJECT_VERSION_MAJOR} << 16
  | ${Slicer_MAIN_PROJECT_VERSION_MINOR} << 8 
  | ${Slicer_MAIN_PROJECT_VERSION_PATCH}"
  OUTPUT_FORMAT HEXADECIMAL)
```

Inspired from what is done for python in [cpython/Include/patchlevel.h](https://github.com/python/cpython/blob/c49016e67c3255b37599b354a8d7995d40663991/Include/patchlevel.h#L29-L35)

jcfr: To be consistent with the new style of the form `Slicer 4.4.0 123456`, should also include `123456` in `MRML_VERSION_HEX` ?

CMake/SlicerVersion.cmake
```diff
@@ -138,3 +92,61 @@ endif()
 
 message(STATUS "Configuring Slicer version [${Slicer_VERSION_FULL}]")
 message(STATUS "Configuring Slicer revision [${Slicer_REVISION}]")
+
```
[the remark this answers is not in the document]
lassoan: If the main application is "Slicer" then we copy ...WC_COMMIT_COUNT_OFFSET and ...REVISION_TYPE values from Slicer to the main application, so Slicer values need to be set first.

Libs/CMakeLists.txt
lassoan: This version is used in C precompiler macros (for example `#if MRML_SUPPORT_VERSION < 0x040000`). If we added the revision then this would not work.

We could use the revision _instead_ of major.minor.patch, but probably we don't need this granularity and it would be bigger change without much (or any) added value.

CMake/SlicerVersion.cmake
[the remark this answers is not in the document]
jcfr: Makes sense.

I missed that, I guess this could be its own commit as it fixes an issue independent of the feature implemented by this commit.

Libs/CMakeLists.txt
jcfr: > used in C precompiler macros (for example #if MRML_SUPPORT_VERSION < 0x040000)

I would prefer the use a macro like `MRML_VERSION_CHECK` because as soon as the minor version go beyond 9, it would be very easy to introduce wrong comparison (either in the main code base or in extension). For example, `MRML_SUPPORT_VERSION < 0x040b00` should be used instead of `MRML_SUPPORT_VERSION < 0x041100`

jcfr: > We could use the revision instead of major.minor.patch, but probably we don't need this granularity and it would be bigger change without much (or any) added value.

Since MRML version  is now coupled with  the Slicer version, after introducing a change into MRML scene, we would be able to test for the version only after the Slicer version is updated.

Instead of appending the Slicer revision, what about using a scheme like `MRML X.Y.Z.M` where:
* X -> Slicer major
* Y -> Slicer minor
* Z -> Slicer patch
* M -> MRML patch

This would allow us to bump the MRML version  while doing development without having to  bump the Slicer version.

lassoan: It's all good for minor and patch version up to 255 (0x... will be compared as a hexadecimal number), but I agree that MRML_VERSION_CHECK() macro would be probably more familiar for developers and it is nicer if we don't expose hex representation in public CMake variables and #defined values.

lassoan: How would you set MRML patch? From the revision? We currently allow either the git hash or the commit-count-based counter to be used as revision. Git has would not allow comparison. Should we remove the option of using git has as revision and only allow revision to be an integer?

jcfr: MRML patch would be an integer set manually in CMake (e.g in a `MRMLVersion.cmake`)

or it could be set by extracting the commit count using a command  like `git rev-list HEAD --count -- ./Libs/MRML/`

lassoan: MRML version is an application level information (many MRML nodes are outside Libs/MRML). Should we use main application revision?

jcfr: To answer the question, we need to define what is the meaning of "MRML Version". 

(1) Is it the version of the MRML Library ?

or

(2) Is it the version of the MRML Library and the description of the required MRML extensions needed to successfully read the scene ? 

I was assuming it was (1) and i was assuming we were conveniently setting the MRML version based on the Slicer version only because the corresponding sources were inside the Slice source tree. 

If we go with the definition (2), the version saved in the file would have to list the modules (and the their version) required to successfully load the scene.

lassoan:
We should probably use the main application version directly as Scene file version.

@jcfr @pieper @sjh26 What was the conclusing at the developer meeting about application version and revision?

pieper:
> What was the conclusing at the developer meeting about application version and revision?

We didn't fully resolve it, but I still like the <major>.<minor>.<number of commits since tag> scheme.

hjmjohnson:
@lassoan Any further work towards finalizing this?

jcfr:
Waiting we converge to a resolution, I will extract the change discussed in https://github.com/Slicer/Slicer/pull/4784#discussion_r400621292 and integrate it, then I will rebase this topic.

CMake/SlicerVersion.cmake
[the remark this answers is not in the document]
jcfr: A dedicated PR was submitted, see https://github.com/Slicer/Slicer/pull/4828

jcfr:
> I will extract the change discussed in #4784 (comment)

see https://github.com/Slicer/Slicer/pull/4828

lassoan:
I've rebased this branch on latest master and fixed all the suggestions.

To clarify the meaning of the version number, I've changed the name to `MRML_APPLICATION_VERSION`. We save into the scene the name+version+revision of the application that created the file, because it is easier to obtain and more specific than inventing some new "MRML version".

I've also added saving the list of extensions into the scene. This could allow displaying a warning if some extensions are not installed that were installed when the scene was saved. It is not considered to be an error or warning if there is a mismatch (many extensions do not affect ability to load or save a scene), but if there are errors then the list of extensions that were present when the scene was saved and the current list of installed extensions are displayed to the user.

Base/QTCore/qSlicerSlicer2SceneReader.cxx
```diff
@@ -1219,8 +1219,8 @@ void qSlicerSlicer2SceneReaderPrivate::importOptionsNode(NodeType& node)
   Q_Q(qSlicerSlicer2SceneReader);
   //$::slicer3::MRMLScene SetErrorMessage "warning: option nodes cannot be imported"
   //$::slicer3::MRMLScene SetErrorCode 1
-  q->mrmlScene()->SetErrorMessage("Warning: option nodes cannot be imported");
-  q->mrmlScene()->SetErrorCode(1);
+  //q->mrmlScene()->SetErrorMessage("Warning: option nodes cannot be imported");
+  //q->mrmlScene()->SetErrorCode(1);
```
pieper, line 1223: At some point we should deprecate slicer2 scenes and remove this dead code.

Base/QTCore/qSlicerCoreApplication.cxx
```diff
@@ -447,6 +447,11 @@ void qSlicerCoreApplicationPrivate::init()
     qDebug() << "Successfully uninstalled extension" << extensionName;
     }
 
+  // Set the list of installed extensions in the scene so that we can warn the user
+  // if the scene is loaded with some of the extensions not present.
+  QString extensionList = model->installedExtensions().join(";");
+  scene->SetExtensions(extensionList.toStdString().c_str());
```
pieper, line 453: Can we also support extension versioning?

Modules/Loadable/Data/qSlicerSceneReader.cxx
pieper: And perhaps in the template an EXTENSION_VERSION_CHECK

pieper (approved):
This is already a great improvement, thanks! 👍 

We can think about versioning extensions at some point, but this is good for now.

There may end up being some false positives because people don't use extensions they have installed or the extension is needed to create the scene but not to work with it later.

We may want to add a mechanism that automatically records the whether the modules from a given extension were entered during the current app session and make a separate list of those to save in the scene.  We could also keep a list of all modules that were used during the session when the scene was created, which might be useful metadata to optionally display.

We may want to make a mechanism so that extensions could "declare" the MRML node types that they will register with the scene.  This way if the app finds that it skips unknown node tags it could look on the list to see whether there is an extension that might handle them.  This could be tied in to extend the suggestion system [in the DICOM module](https://github.com/Slicer/Slicer/blob/master/Modules/Scripted/DICOM/DICOMExtensions.json).

lassoan:
> There may end up being some false positives because people don't use extensions they have installed or the extension is needed to create the scene but not to work with it later.

Yes, actually most extensions do not affect scene loading. To avoid false alarms, the list of extensions is only displayed if there is a scene loading error or warning. Even then, it is just an informational message, it is not even designated as a warning. We now detect and display node instantiation errors to the user, so if an extension that would be needed to read the scene is not installed then you get an error about failing to create a node and a information about which extensions are not installed that were installed when the scene was created. This should give enough hint to the user about what the problem is and what to do about it.

lassoan:
Extension versioning is not needed if extensions are changed only so much that they remain forward compatible (current version of the extension can handle scenes created by a future version of the extension) for the same Slicer major.minor version. If we can create a new Slicer minor version in every quarter then this should be easily doable, but I think it would rarely cause trouble even with a yearly release.

Base/QTCore/qSlicerSlicer2SceneReader.cxx
lassoan: I've removed it now. It was not tested and was surely broken. Slicer2 scenes can be better converted to Slicer3 scenes using Slicer3 application.

Base/QTCore/qSlicerCoreApplication.cxx
lassoan: If we release a new Slicer core version in every 3-6 months then probably we can keep extensions compatible with the corresponding Slicer core and we don't need additional versioning.

Issue #4602, opened 2020-03-13
Cleanup datastore


slicerbot:
_This issue was created automatically from an original [Mantis Issue](https://mantisarchive.slicer.org/view.php?id=4602). Further discussion may take place here._

lassoan:
While datastore works, it breaks the extension manager (open DataStore module, close the window; open Extension Manager => instead of "Install" button, "Download" button is displayed and it downloads the extension instead of installing it).

Also, it is clear that there will be no Midas communities (there are many alternative, more open and more popular data repositories) and at some point the Midas server will be retired.

@Slicer/slicer-core What do you think about removing SampleData module, moving data to into a github repository and add links to important data sets (atlas collection, registration case library, maybe a few more nicer data sets) to SampleData module? We could use the existing [SlicerTestingData](https://github.com/Slicer/SlicerTestingData) or a new SlicerDataStore github repository.

pieper:
I like the idea of migrating from midas to github for this.

I'm not sure what you mean by removing the SampleData module - did you mean DataStore?  If so, yes.

lassoan:
Yes, sorry I meant removing DataStore module.

jcfr:
> removing DataStore

That makes sense. 

Also, if there is a strong push to keep similar functionality we could modify the DataStore module and  leverage GitHub API to upload asset.  That said, I would prefer to avoid the added complexity and associated maintenance cost.

> existing SlicerTestingData or a new SlicerDataStore github repository.

That is a tough questions.

Before answering, may be we should define the role of each. Here is an attempt:

* SlicerTestingData: Collection of datasets used for testing Slicer. These datasets may not have any clinical relevance and may be incomplete. 

* SlicerDataStore: Collection of datasets that are clinically relevant and anatomically correct

Now for practical reason, we could organize all data under the same repository: `SlicerDataStore` and have different type of release: `testing-data-<hashalgo>` and `clinical-data-<hashalgo>`

```
https://github.com/Slicer/SlicerDataStore/releases/download/testing-data-<hashalgo>/<hash>
https://github.com/Slicer/SlicerDataStore/releases/download/clinical-data-<hashalgo>/<hash>
```

lassoan:
[...]

PR #4785, opened 2020-03-30
ENH: Update sample data URLs to use GitHub instead of Midas

lassoan:
Also improved SampleData.downloadFromSource to not load scenes many times if download or loading fails (by default, download and load is only attempted up to 3 times).

re #4602


jcfr:
Looks good. 

Only few minor tweaks

lassoan:
@jcfr I have addressed all your comments, please review.

lassoan:
@jcfr comments have been addressed, please check.

PR #4786, opened 2020-03-30
ENH: Add copy file option to importDicom utility function

cpinter:


Modules/Scripted/DICOMLib/DICOMUtils.py
```diff
@@ -309,15 +309,15 @@ def __exit__(self, type, value, traceback):
...
+def importDicom(dicomDataDir, dicomDatabase=None, copyFiles=False):
   """ Import DICOM files from folder into Slicer database
   """
   try:
     indexer = ctk.ctkDICOMIndexer()
     assert indexer is not None
     if dicomDatabase is None:
       dicomDatabase = slicer.dicomDatabase
-    indexer.addDirectory( dicomDatabase, dicomDataDir )
+    indexer.addDirectory( dicomDatabase, dicomDataDir, copyFiles )
```
jcfr: nitpick:

`addDirectory( dicomDatabase, dicomDataDir, copyFiles )` -> `addDirectory(dicomDatabase, dicomDataDir, copyFiles)`

cpinter: @jcfr I'd prefer that too, but people still use different styles in the code, and when facing with something like this, the question is whether to make that one line inconsistent within the file/function, or fix it to conform to the global guidelines. So should we fix the style line by line that we are touching?

jcfr: When looking at the other lines found in this file, this is the only line that differ in style. It seems reasonable to remove the extra spaces.

Soon we will setup flake8 (with a config file) to enforce the style when code is contributed.

cpinter:
@jcfr I fixed the style, please approve if it looks good. Thanks!

cpinter:
Thanks! Not sleeping much @jcfr ? 😪

Issue #4664, opened 2020-03-13
qMRMLSegmentationEditorWidget: Unecessary error messages


slicerbot:
_This issue was created automatically from an original [Mantis Issue](https://mantisarchive.slicer.org/view.php?id=4664). Further discussion may take place here._

PR #4788, opened 2020-03-30
BUG: Prevent segment editor error messages on scene close

### LINKED ISSUE #4662: Various traceback errors when closing scene after using Segment Editor
_This issue was created automatically from an original [Mantis Issue](https://mantisarchive.slicer.org/view.php?id=4662). Further discussion may take place here._

Sunderlandkyl:
- Modified events are now blocked during parameter node initialization
- Error message is not displayed if no parameter node is found during updateWidgetFromMRML
- More thorough error checking for parameter and segmentation nodes in autocomplete effect

fixes #4662, #4664

### DIFF
--- a/Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.cxx
+++ b/Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.cxx
@@ -1166,11 +1166,16 @@ void qMRMLSegmentEditorWidget::updateWidgetFromMRML()
 
   if (!d->ParameterSetNode)
     {
-    qCritical() << Q_FUNC_INFO << ": Invalid segment editor parameter set node";
+    d->SegmentationNodeComboBox->setEnabled(false);
+    d->EffectsGroupBox->setEnabled(false);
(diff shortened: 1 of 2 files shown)


Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.cxx
```diff
@@ -1166,11 +1166,10 @@ void qMRMLSegmentEditorWidget::updateWidgetFromMRML()
 
   if (!d->ParameterSetNode)
     {
-    qCritical() << Q_FUNC_INFO << ": Invalid segment editor parameter set node";
```
Sunderlandkyl: @lassoan This is to fix the from the steps in https://mantisarchive.slicer.org/view.php?id=4664.
The other solution would be to check if the parameter node exists before calling updateWidgetFromMRML from setMRMLScene.

lassoan: I agree that it makes sense to remove the error message here but before returning we should probably disable the widgets that we would normally update with valid content in this method.

PR #4795, opened 2020-03-31
COMP: Disable copy of all VTK based objects

jcfr:
This commit consistently prevents use of copy-ctor and =operator
for all VTK based objects.

It also replaces VTK_DELETE_FUNCTION with =delete to support
building against VTK >= kitware/vtk@66f02acea (Deprecate: VTK_OVERRIDE,
VTK_FINAL, VTK_DELETE_FUNCTION. All of the above macros are now behind
a VTK_LEGACY_REMOVE check as we require a C++11 compiler and the macros
add no value.)


jcfr:
Thanks.

Some background, my local build is almost without error when building against the latest VTK 9.0

I will then introduce changes (after testing them locally and independently) during the remaining of the evening ...

jcfr:
I will merge after my local build completes.

PR #4799, opened 2020-04-01
COMP: Fix configuration with CMake >= 3.17. CMAKE_DEFAULT_BUILD_TYPE can not be specified

### LINKED ISSUE #4798: Configuring with CMake >= 3.17.0 on Windows fails with "CMAKE_DEFAULT_BUILD_TYPE" being unsupported
Slicer build-system introduces the variable `CMAKE_DEFAULT_BUILD_TYPE` in Nov 2017 (see c808da1d4).

Following the introduction of the same variable in CMake 3.17 (see [CMAKE_DEFAULT_BUILD_TYPE](https://cmake.org/cmake/help/latest/variable/CMAKE_DEFAULT_BUILD_TYPE.html)), configuring on windows is now failing with an error like the following:

```
CMake Error:
   Generator

     Visual Studio 15 2017

   does not support variable

     CMAKE_DEFAULT_BUILD_TYPE

   but it has been specified.
```

Thanks @jjomier for the report

## Steps to reproduce

* Checkout Slice sources
* Download CMake >= 3.17
* Attempt to configure Slicer on windows

## Expected behavior

Successful configuration

## Environment
- Any Slicer
- CMake >= 3.17
- Visual Studio

jcfr:
This commit conditionally sets CMAKE_DEFAULT_BUILD_TYPE variable to fix
configuration error <s>on Windows</s> using CMake >= 3.17.

This change is required because Slicer build-system introduced the
variable CMAKE_DEFAULT_BUILD_TYPE in Nov 2017 (see c808da1) and it triggers
a configuration error following the introduction of the same variable in
CMake 3.17.
See https://cmake.org/cmake/help/latest/variable/CMAKE_DEFAULT_BUILD_TYPE.html

Error fixed by this commit:

```
  CMake Error:
     Generator

       Visual Studio 15 2017

     does not support variable

       CMAKE_DEFAULT_BUILD_TYPE

     but it has been specified.
```

Custom Slicer application may have to be updated to conditionally set
CMAKE_DEFAULT_BUILD_TYPE based on the value of the GENERATOR_IS_MULTI_CONFIG
global property.

Fixes #4798

### DIFF
--- a/CMake/SlicerConfig.cmake.in
+++ b/CMake/SlicerConfig.cmake.in
@@ -112,7 +112,7 @@ endif()
 # --------------------------------------------------------------------------
 # Slicer options
 # --------------------------------------------------------------------------
-set(CMAKE_DEFAULT_BUILD_TYPE "@CMAKE_DEFAULT_BUILD_TYPE@")
+set(Slicer_DEFAULT_BUILD_TYPE "@Slicer_DEFAULT_BUILD_TYPE@")
 set(Slicer_USE_NUMPY "@Slicer_USE_NUMPY@")
 set(Slicer_USE_SCIPY "@Slicer_USE_SCIPY@")
 set(Slicer_USE_PYTHONQT "@Slicer_USE_PYTHONQT@")
--- a/CMake/SlicerInitializeBuildType.cmake
+++ b/CMake/SlicerInitializeBuildType.cmake
@@ -1,15 +1,15 @@
 
 # Default build type to use if none was specified
-if(NOT DEFINED CMAKE_DEFAULT_BUILD_TYPE)
-  set(CMAKE_DEFAULT_BUILD_TYPE "Debug")
+if(NOT DEFINED Slicer_DEFAULT_BUILD_TYPE)
+  set(Slicer_DEFAULT_BUILD_TYPE "Debug")
 endif()
 
 # Set a default build type if none was specified
 if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
 
-  message(STATUS "Setting build type to '${CMAKE_DEFAULT_BUILD_TYPE}' as none was specified.")
+  message(STATUS "Setting build type to '${Slicer_DEFAULT_BUILD_TYPE}' as none was specified.")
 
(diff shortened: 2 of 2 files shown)


jcfr:
I locally tested and it fixes the problem on Windows with Visual Studio while still allowing to pass `CMAKE_DEFUALT_BUILD_TYPE` on linux

jcfr:
That said, it fails when trying to use `Ninja` generator.


```
$ ~/Software/cmake-3.17.0-Linux-x86_64/bin/cmake -DQt5_DIR=$Qt5_DIR -DCMAKE_DEFAULT_BUILD_TYPE:STRING=Release -GNinja ../Slicer
[...]
CMake Error:
  Generator

    Ninja

  does not support variable

    CMAKE_DEFAULT_BUILD_TYPE

  but it has been specified.

```

jcfr:
Thanks for testing  @jamesobutler . It turns out that the variable `CMAKE_DEFAULT_BUILD_TYPE` is now reserved for the `Ninja Multi-Config` generator. 

I renamed it to `Slicer_DEFAULT_BUILD_TYPE` and added [an entry](https://www.slicer.org/w/index.php?title=Documentation/Nightly/Developers/Tutorials/MigrationGuide#Slicer_4.11:_Variable_CMAKE_DEFAULT_BUILD_TYPE_renamed_to_Slicer_DEFAULT_BUILD_TYPE) to the migration guide

jcfr:
Since would impact build system of existing custom slicer application, I am discussing with the CMake team the future of the variable `CMAKE_DEFAULT_BUILD_TYPE`. I will post update here and in the associated CMake discourse post. See https://discourse.cmake.org/t/ninja-multi-config-is-overly-complex/684/10

jcfr:
The current CMake behavior will not be changed.

For reference:

> we already set CMAKE_DEFAULT_BUILD_TYPE.

The `CMAKE_` prefix is reserved for use by upstream CMake.  You should not
be setting variables with such names that are not defined by CMake.

> could we not fail when this variable is set for single-config generator ?

That would mean *other* projects that set it will suddenly get a behavior
change in the future when we activate it for single-config generators.

PR #4800, opened 2020-04-01
ENH: Add interaction handles for transforming markup positions

Sunderlandkyl:
Adds 6 interaction handles to rotate and translate markup control points.
Each of the handles allows constrained motion of the markup control points constrained around/on a single axis.

Handles are only visible if the node is selected (via the vtkMRMLNode::Selected attribute).
This commit also adds some additional helper functions to vtkMRMLSelectionNode for getting the currently selected nodes, detecting if only one node is selected, and deselecting all nodes.

image

Demo video:
https://www.dropbox.com/s/rycp6hxoh1i17rf/markup_handle_demo_1.mp4?dl=0

### DIFF
--- a/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsAngleNode.cxx
+++ b/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsAngleNode.cxx
@@ -22,6 +22,7 @@
 #include "vtkMRMLScene.h"
 
 // VTK includes
+#include <vtkMatrix4x4.h>
 #include <vtkNew.h>
 #include <vtkObjectFactory.h>
 
@@ -30,7 +31,6 @@
 
 //----------------------------------------------------------------------------
 vtkMRMLNodeNewMacro(vtkMRMLMarkupsAngleNode);
-
 
 //----------------------------------------------------------------------------
 vtkMRMLMarkupsAngleNode::vtkMRMLMarkupsAngleNode()
@@ -92,4 +92,60 @@ void vtkMRMLMarkupsAngleNode::UpdateMeasurements()
       }
     }
   this->WriteMeasurementsToDescription();
+}
+
+//---------------------------------------------------------------------------
+void vtkMRMLMarkupsAngleNode::UpdateInteractionHandleToWorldMatrix()
+{
+  if (this->GetNumberOfControlPoints() < 3)
+    {
+    return;
+    }
+
+  double point0_World[3] = { 0.0 };
+  double point1_World[3] = { 0.0 };
+  double point2_World[3] = { 0.0 };
+  this->GetNthControlPointPositionWorld(0, point0_World);
+  this->GetNthControlPointPositionWorld(1, point1_World);
+  this->GetNthControlPointPositionWorld(2, point2_World);
+
+  double epsilon = 1e-5;
+  double handleX_World[3] = { 1.0, 0.0, 0.0 };
+  vtkMath::Subtract(point0_World, point1_World, handleX_World);
+  if (vtkMath::Norm(handleX_World) < epsilon)
+    {
+    return;
+    }
+  vtkMath::Normalize(handleX_World);
+
+  double vectorPoint1ToPoint2_World[3] = { 0.0 };
+  vtkMath::Subtract(point2_World, point1_World, vectorPoint1ToPoint2_World);
+  if (vtkMath::Norm(vectorPoint1ToPoint2_World) < epsilon)
+    {
+    return;
+    }
+  vtkMath::Normalize(vectorPoint1ToPoint2_World);
+
+  if (std::abs(vtkMath::Dot(handleX_World, vectorPoint1ToPoint2_World)) > 1.0 - epsilon)
+    {
+    return;
+    }
+
+  double handleZ_World[3] = { 0.0 };
+  vtkMath::Cross(handleX_World, vectorPoint1ToPoint2_World, handleZ_World);
+  vtkMath::Normalize(handleZ_World);
+
+  double handleY_World[3] = { 0.0 };
+  vtkMath::Cross(handleZ_World, handleX_World, handleY_World);
+  vtkMath::Normalize(handleY_World);
+
+  vtkNew<vtkMatrix4x4> handleToWorldMatrix;
+  for (int i = 0; i < 3; ++i)
+    {
+    handleToWorldMatrix->SetElement(i, 0, handleX_World[i]);
+    handleToWorldMatrix->SetElement(i, 1, handleY_World[i]);
+    handleToWorldMatrix->SetElement(i, 2, handleZ_World[i]);
+    handleToWorldMatrix->SetElement(i, 3, point1_World[i]);
+    }
+  this->InteractionHandleToWorldMatrix->DeepCopy(handleToWorldMatrix);
 }
--- a/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsLineNode.cxx
+++ b/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsLineNode.cxx
@@ -19,8 +19,10 @@
 #include "vtkMRMLMarkupsDisplayNode.h"
 #include "vtkMRMLMarkupsLineNode.h"
 #include "vtkMRMLMarkupsFiducialStorageNode.h"
+#include "vtkMatrix4x4.h"
 #include "vtkMRMLScene.h"
 #include "vtkMRMLUnitNode.h"
+#include "vtkTransform.h"
 
 // VTK includes
 #include <vtkNew.h>
@@ -94,4 +96,46 @@ void vtkMRMLMarkupsLineNode::UpdateMeasurements()
     this->SetNthMeasurement(0, "length", length, unit, printFormat);
     }
   this->WriteMeasurementsToDescription();
+}
+
+//---------------------------------------------------------------------------
+void vtkMRMLMarkupsLineNode::UpdateInteractionHandleToWorldMatrix()
+{
+  Superclass::UpdateInteractionHandleToWorldMatrix();
+  if (this->GetNumberOfControlPoints() < 2)
+    {
+    return;
+    }
+
+  double handleX_World[4] = { 1.0, 0.0, 0.0, 0.0 };
+  this->InteractionHandleToWorldMatrix->MultiplyPoint(handleX_World, handleX_World);
+
+  double point0_World[3];
+  this->GetNthControlPointPositionWorld(0, point0_World);
+  double point1_World[3];
+  this->GetNthControlPointPositionWorld(1, point1_World);
+
+  double vectorPoint0ToPoint1_World[4] = { 0.0 };
+  vtkMath::Subtract(point1_World, point0_World, vectorPoint0ToPoint1_World);
+
+  double angle = vtkMath::DegreesFromRadians(vtkMath::AngleBetweenVectors(vectorPoint0ToPoint1_World, handleX_World));
+  double epsilon = 1e-5;
+  if (angle < epsilon)
+    {
+    return;
+    }
+
+  double rotationVector_Local[3] = { 0.0 };
+  vtkMath::Cross(handleX_World, vectorPoint0ToPoint1_World, rotationVector_Local);
+
+  double origin_World[4] = { 0.0, 0.0, 0.0, 1.0 };
+  this->InteractionHandleToWorldMatrix->MultiplyPoint(origin_World, origin_World);
+
+  vtkNew<vtkTransform> handleToWorldMatrix;
+  handleToWorldMatrix->PostMultiply();
+  handleToWorldMatrix->Concatenate(this->InteractionHandleToWorldMatrix);
+  handleToWorldMatrix->Translate(-origin_World[0], -origin_World[1], -origin_World[2]);
+  handleToWorldMatrix->RotateWXYZ(angle, rotationVector_Local);
+  handleToWorldMatrix->Translate(origin_World);
+  this->InteractionHandleToWorldMatrix->DeepCopy(handleToWorldMatrix->GetMatrix());
 }
--- a/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsLineNode.h
+++ b/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsLineNode.h
@@ -67,6 +67,9 @@ protected:
   void operator=(const vtkMRMLMarkupsLineNode&);
 
   void UpdateMeasurements() override;
+
+  /// Calculates the handle to world matrix based on the current control points
+  void UpdateInteractionHandleToWorldMatrix() override;
 };
 
 #endif
--- a/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsNode.cxx
+++ b/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsNode.cxx
@@ -29,6 +29,9 @@
 // Slicer MRML includes
 #include "vtkMRMLScene.h"
 
+// vtkAddon includes
+#include "vtkAddonMathUtilities.h"
+
 // VTK includes
 #include <vtkAbstractTransform.h>
 #include <vtkBitArray.h>
@@ -40,11 +43,15 @@
 #include <vtkMatrix4x4.h>
 #include <vtkNew.h>
 #include <vtkObjectFactory.h>
+#include <vtkPlane.h>
 #include <vtkPointData.h>
 #include <vtkPolyData.h>
 #include <vtkStringArray.h>
+#include <vtkTransform.h>
 #include <vtkTransformPolyDataFilter.h>
 #include <vtkTrivialProducer.h>
+#include "vtk_eigen.h"
+#include VTK_EIGEN(Dense)
 
 // STD includes
 #include <sstream>
@@ -94,6 +101,8 @@ vtkMRMLMarkupsNode::vtkMRMLMarkupsNode()
   this->CurveCoordinateSystemGeneratorWorld->SetInputConnection(this->CurvePolyToWorldTransformer->GetOutputPort());
 
   this->TransformedCurvePolyLocator = vtkSmartPointer<vtkPointLocator>::New();
+
+  this->InteractionHandleToWorldMatrix = vtkSmartPointer<vtkMatrix4x4>::New();
 }
 
 //----------------------------------------------------------------------------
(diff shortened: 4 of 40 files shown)


Libs/MRML/Core/vtkMRMLSelectionNode.h
```diff
@@ -228,6 +228,15 @@ class VTK_MRML_EXPORT vtkMRMLSelectionNode : public vtkMRMLNode
   /// Get the number of class names in the list
   int GetNumberOfPlaceNodeClassNamesInList() { return static_cast<int>(this->PlaceNodeClassNameList.size()); };
 
+  /// Returns true if this is the only node that is selected
+  bool IsOnlyNodeSelected(vtkMRMLNode*);
```
jcfr, line 232: Since the function is not used, I suggest to remove it for now.

Libs/MRML/Core/vtkMRMLSelectionNode.cxx
```diff
@@ -561,3 +563,62 @@ void vtkMRMLSelectionNode::SetActivePlotChartID(const char* id)
...
+      selectedNodes->AddItem(node);
+      }
+    iter->GoToNextItem();
+    }
+}
+
+//----------------------------------------------------------------------------
+void vtkMRMLSelectionNode::DeselectAllNodes()
+{
+  vtkCollection* nodes = this->Scene->GetNodes();
```
jcfr, line 612: What about iterating over the list returned by `GetSelectedNodes` ?

Modules/Loadable/Markups/VTKWidgets/vtkSlicerMarkupsWidget.h
```diff
@@ -57,12 +57,16 @@ class VTK_SLICER_MARKUPS_MODULE_VTKWIDGETS_EXPORT vtkSlicerMarkupsWidget : publi
   {
     WidgetStateDefine = WidgetStateUser, // click in empty area will place a new point
     WidgetStateTranslateControlPoint, // translating the active point by mouse move
+    WidgetStateOnTranslationHandle,
```
jcfr, line 60: Add comments

Modules/Loadable/Markups/VTKWidgets/vtkSlicerPlaneRepresentation3D.cxx
```diff
@@ -184,6 +184,19 @@ void vtkSlicerPlaneRepresentation3D::UpdateFromMRML(vtkMRMLNode* caller, unsigne
...
+void vtkSlicerPlaneRepresentation3D::UpdateInteractionPipeline()
+{
+  Superclass::UpdateInteractionPipeline();
+
+  vtkMRMLMarkupsPlaneNode* planeNode = vtkMRMLMarkupsPlaneNode::SafeDownCast(this->GetMarkupsNode());
+  if (!planeNode || planeNode->GetNumberOfControlPoints() < 3)
+    {
+    this->InteractionPipeline->Actor->SetVisibility(false);
+    return;
+    }
```
jcfr, line 197: Should a call to `this->InteractionPipeline->Actor->SetVisibility(true);` be added below ?

Modules/Loadable/Markups/VTKWidgets/vtkSlicerMarkupsWidgetRepresentation3D.cxx
```diff
@@ -925,3 +999,15 @@ bool vtkSlicerMarkupsWidgetRepresentation3D::GetNthControlPointViewVisibility(in
...
+//----------------------------------------------------------------------
+void vtkSlicerMarkupsWidgetRepresentation3D::UpdateInteractionPipeline()
+{
+  Superclass::UpdateInteractionPipeline();
+  vtkMRMLViewNode* viewNode = vtkMRMLViewNode::SafeDownCast(this->ViewNode);
+  if (!viewNode)
+    {
+    this->InteractionPipeline->Actor->SetVisibility(false);
+    return;
+    }
```
jcfr, line 1012: Should a call to `this->InteractionPipeline->Actor->SetVisibility(true);` be added below ? If not, may be sensible to add a comment

Modules/Loadable/Markups/VTKWidgets/vtkSlicerMarkupsWidgetRepresentation3D.cxx
```diff
@@ -398,6 +401,9 @@ void vtkSlicerMarkupsWidgetRepresentation3D::CanInteract(
       }
     }
 
+  // See if we can interact with the handle
```
jcfr, line 404: What about describing what is done to check if interaction can effectively be done or not.

Modules/Loadable/Markups/VTKWidgets/vtkSlicerMarkupsWidgetRepresentation3D.cxx
Sunderlandkyl: The same thing applies to both this and the previous comment, but final visibility is controlled by the top level superclass in vtkSlicerMarkupsWidgetRepresentation.

Adding a comment and moving the superclass call to the end of the function to make the order more logical.

Sunderlandkyl:
Changes made, thanks for the feedback!

lassoan:
Thank you this will be awesome!

"Selection" needs to be exposed on GUI. I see two options:
- A. "Select" happens automatically when the user clicks on a widget.
- B. Allow an "interactive" flag that could be activated/deactivated for each widget (multiple widgets can be interactive). This interactive mode would be similar to how Annotation ROI widget can be "locked" (ROI is still shown but no handles), how markups control points can be locked (control points are still shown but you cannot interact with them), and how an interactive widget can be enabled for editing linear transforms in 3D views.

I think I slightly prefer option B (although we agreed in option A last time we talked). It is more similar to how things work now and overall simpler design (no need for creating a new single "selected" node concept - we can still add later if needed, but not necessary for this feature). It would also allow making multiple widgets interactive, which is helpful if you want to adjust 2-3 widgets repeatedly (you don't need to "click to select" and then "click-and-drag to move"; just "click-and-drag to move").

 Few smaller remarks:
- We should make the center of the interaction widget draggable to make it unconstrained translation (in image plane normal view).
- Handles always show up in slice views for all planes. It should be only shown if interactive mode is enabled.
- Handles are always shown for plane widgets in slice views. They should not be shown if the plane is not visible at all (similarly to how the plane fill is not displayed when the plane does not intersect the slice at all)
- It is hard to click on an arrow because it seems that the arrow shaft cannot be picked, only the arrow head. It would be nice if the arrow shaft could be picked as well.

Sunderlandkyl:
@lassoan OK, I've added the "HandlesInteractive" attribute to the markups display node, and added the option to change it from the GUI. The handles are visible when HandlesInteractive is True.

image

I think I've addressed all of the other comments. Let me know if there's something I missed.

finetjul:
Very cool feature, thanks !

From the video, it seems that translations within 2D views are done along the 3D interaction handle vectors. It would seem to me that translations could be done along the view plane x,y directions instead (option?) ?

That widget would be very useful to transform "Models" as well.

Would it be difficult to make it an all-purpose widget (e.g. as part of the Transform module ?) ?

Sunderlandkyl:
It was designed to be (mostly) separate from Markups so that it could be used for all transformable nodes in the future. The only use of markup specific nodes currently is the synchronization of the selected handle between views using the display node.

jcfr:
I wonder if later on the new widget could be used to interactively update the transform:

image

lassoan:
Thank you @Sunderlandkyl this works very well.

A few more things to tune:
- I agree that @finetjul that it could be useful to allow rotation along current plane normal, especially in slice views. Maybe even translation along current view axes could be useful, but this is not critical, since there is already the in-plane translation option (when click-and-dragging the center). It could be a local/global coordinate system switcher (similar to what we saw in MeshMixer), or maybe keyboard shortcut modifier (con: it would be hard to discover these), or additional arrows (con: it would make the widget look a bit too busy).
- Translation by dragging the centerpoint is really useful, it is just a bit difficult to hit the center because the arrow shafts take precedence. Would it be possible to always translate when the mouse pointer is near the center (even when it is also near an arrow shaft)?
- It would be nice to have a stronger color difference in the centerpoint when it is highlighted (on my screen, the centerpoint is quite small and the color change between white/yellow is not really visible).
- Could you add the Display/Interaction/Visible checkbox as a checkable menu item to the context menu of the widget to make it easier to access? (currently, the menu items are Rename point, Delete point, Toggle select point, Edit properties)

Sunderlandkyl:
@lassoan I didn't add the in-plane translation, however I can make that change in a follow-up pull request. All other comments should be addressed.

Center handle is now orange to provide more contrast against the yellow.

image

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsNode.cxx
```diff
@@ -2009,3 +2018,92 @@ void vtkMRMLMarkupsNode::WriteMeasurementsToDescription()
...
+    {
+    this->InteractionHandleToWorld->SetElement(i, 3, origin[i]);
+    }
+
+  if (this->GetNumberOfControlPoints() < 3)
+    {
+    return;
+    }
+
+  // The orientation of the coordinate system is adjusted so that the z axis aligns with the normal of the
```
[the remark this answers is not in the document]
Sunderlandkyl: Tried that initially, but it would result in a circular dependency MRML -> Logic -> MRML.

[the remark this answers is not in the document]
Sunderlandkyl: Tried that initially, but it would result in a circular dependency. MRML->Logic->MRML.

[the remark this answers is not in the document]
lassoan: I see. Maybe we could move the heart of the function into MRML, maybe as a static method of vtkMRMLMarkupsNode.

Libs/MRML/Core/vtkMRMLSelectionNode.cxx
```diff
@@ -561,3 +563,51 @@ void vtkMRMLSelectionNode::SetActivePlotChartID(const char* id)
 {
   this->SetNodeReferenceID(ACTIVE_PLOT_CHART_REFERENCE_ROLE, id);
 }
+
+//----------------------------------------------------------------------------
+void vtkMRMLSelectionNode::GetSelectedNodes(vtkCollection* selectedNodes)
```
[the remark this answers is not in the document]
Sunderlandkyl: No, we do not.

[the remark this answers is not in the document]
lassoan: Then save these code snippets for yourself and remove it from the code for now. We may use them later.

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsNode.cxx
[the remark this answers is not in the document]
jcfr: or into `vtkAddon` as `vtkAddonMathUtilities::FitPlaneToPoints`?

lassoan:
Sorry @Sunderlandkyl it seems I caused a merge conflict, I hope it won't be too difficult to resolve. 

One of the changes in the conflicting commits was a stack corruption caused by calling GetSize() (it set 3 values while only 2 values were allocated on the stack). I've fixed the issue but the problem was most likely that GetSize() is not a commonly used method name in VTK for physical size (but instead more specific terms are used, such bounds, extent, radius, diameter, ...). I've checked how box sizes are defined in VTK and found that in most places "double bounds[6]" is used (xmin, xmax, ymin, ymax, zmin, zmax). I think we should use this, too. Since MRML nodes already have GetBounds(), GetRASBounds(), GetSliceBounds(), we could add something like GetLocalBounds() or GetPlaneBounds() - indicating that the bounds are in the plane's coordinate system.

Sunderlandkyl:
@lassoan It was just due to the addition of the IsDisplayable method. Thanks for catching the stack corruption!

I'll change the plane node function names.

Sunderlandkyl:
Functions are now Get/SetPlaneModelBounds.

lassoan:
> Functions are now Get/SetPlaneModelBounds.

Thank you, that's much more clear (we know how "bounds" is defined in VTK). What does "Model" refer to? If you feel that PlaneBounds is not distinctive enough then I would not use Model (because it is already used for Model nodes) but it could be instead PlaneRegionBounds, PlaneAreaBounds, PlaneAxisBounds, OrientedBounds, ... Sorry for the nitpicking, this should be the last thing to decide on and then it's ready to merge.

Sunderlandkyl:
Ok, I've changed it to PlaneOrientedBounds.

I didn't use PlaneBounds, since there is already a Plane coordinate system (the plane offset). This used to reflect the height/width of the untransformed plane model (hence why I used PlaneModelBounds).

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsLineNode.h
```diff
@@ -67,6 +67,9 @@ class  VTK_SLICER_MARKUPS_MODULE_MRML_EXPORT vtkMRMLMarkupsLineNode : public vtk
   void operator=(const vtkMRMLMarkupsLineNode&);
 
   void UpdateMeasurements() override;
+
+  /// Calculates the handle to world matrix based on the current control points
+  void UpdateInteractionHandleModelToWorld() override;
```
jcfr, line 72: `UpdateInteractionHandleModelToWorld` -> `UpdateInteractionHandleMatrixToWorld` ?

Sunderlandkyl: Done!

lassoan: Sorry, but if we want to be consistent then we need to put "Matrix" to the end of the name: UpdateInteractionHandleToWorldMatrix. This is not something that I would delay the pull request for, so only change it if you change other things anyway.

Sunderlandkyl: Ok, changed to UpdateInteractionHandleToWorldMatrix.

lassoan:
> Ok, I've changed it to PlaneOrientedBounds.
> 
> I didn't use PlaneBounds, since there is already a Plane coordinate system (the plane offset). This used to reflect the height/width of the untransformed plane model (hence why I used PlaneModelBounds).

What coordinate system names are we using: Plane, Local, and World?
What coordinate system these bounds are defined in: Plane or Local?

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsAngleNode.cxx
```diff
@@ -93,3 +93,59 @@ void vtkMRMLMarkupsAngleNode::UpdateMeasurements()
...
+  vtkMath::Normalize(modelZ_World);
+
+  double modelY_World[3] = { 0.0 };
+  vtkMath::Cross(modelZ_World, modelX_World, modelY_World);
+  vtkMath::Normalize(modelY_World);
+
+  vtkNew<vtkMatrix4x4> modelToWorldMatrix;
+  for (int i = 0; i < 3; ++i)
+    {
+    modelToWorldMatrix->SetElement(i, 0, modelX_World[i]);
```
jcfr, line 145: `modelToWorldMatrix` -> `handleToWorldMatrix`

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsAngleNode.cxx
```diff
@@ -93,3 +93,59 @@ void vtkMRMLMarkupsAngleNode::UpdateMeasurements()
...
+    modelToWorldMatrix->SetElement(i, 1, modelY_World[i]);
+    modelToWorldMatrix->SetElement(i, 2, modelZ_World[i]);
+    modelToWorldMatrix->SetElement(i, 3, point1_World[i]);
+    }
+  this->InteractionHandleModelToWorld->DeepCopy(modelToWorldMatrix);
```
jcfr, line 150: `InteractionHandleModelToWorld` -> InteractionHandleToWorldMatrix

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsLineNode.cxx
```diff
@@ -95,3 +97,45 @@ void vtkMRMLMarkupsLineNode::UpdateMeasurements()
...
+    return;
+    }
+
+  double rotationVector_Local[3] = { 0.0 };
+  vtkMath::Cross(modelX_World, vectorPoint0ToPoint1_World, rotationVector_Local);
+
+  double origin_World[4] = { 0.0, 0.0, 0.0, 1.0 };
+  this->InteractionHandleModelToWorld->MultiplyPoint(origin_World, origin_World);
+
+  vtkNew<vtkTransform> modelToWorldMatrix;
```
jcfr, line 134: `modelToWorldMatrix` -> `handleToWorldMatrix`

Sunderlandkyl:
> What coordinate system names are we using: Plane, Local, and World?

Yes, those are the current coordinate systems defined. The main point of confusion I think is that we determined previously that the "Plane" coordinate system is what we used to call "Offset".

> What coordinate system these bounds are defined in: Plane or Local?

Neither. The GetSize function was previously used to define the height and width of the plane model along the x and y axis, before transforming the model to local coordinates. The GetXYZBounds function would be in the "Model" coordinate system.

pieper:
Sorry to jump into the middle of this, but it might be good to also look at the vtkMRMLSliceNode, which is essentially a bounded plane with a sampling grid.  It might be cool if we could unify the naming and coordinate system conventions, and maybe even find ways to unify some of the manipulation methods

lassoan:
Being consistent with slice plane is a good idea. In slice nodes, we use RAS, Slice and XY coordinate systems (and UVW but we can probably ignore that). I think they could be mapped to Markups plane like this:

| Slice                                           | Plane                                                      |
|-------------------------------------------------|------------------------------------------------------------|
| RAS                                             | World                                                      |
| Slice                                           | Local                                                      |
| XY (in pixels, can be offset compared to Slice) | Plane (can be additional offset/rotation compared to Local) |

In Markups "World" coordinate system has been in use for several years and I think it is better than RAS because probably at some point we would switch to LPS coordinate system inside Slicer, too, and World coordinate system would be better for this.

XY is in pixels and I'm not sure if we want to allow scaling in PlaneToLocal transform.

Size size is called "FieldOfView" in slice node, which would not be very intuitive for a plane widget.

Overall, I don't see how we could bring the coordinate system and size/extent/bounds names closer to each other in plane and slice nodes.

@Sunderlandkyl do you have any suggestion?

jcfr:
> I think they could be mapped to Markups plane like this

That is great description. May this could be added to the documentation ?

Sunderlandkyl:
The description of the plane coordinate system didn't match the existing implementation, so I adjusted it to match the description. GetPlaneBounds is now in the Plane coordinate system.

Also made the style changes and added a test for the vtkMRMLMarkupsPlaneNode functions.

hjmjohnson:
@Sunderlandkyl Is this ready for integration, or is there more work to be done?

Sunderlandkyl:
Ready for integration pending final review.

PR #4802, opened 2020-04-02
COMP: Correctly set CMAKE_CXX_FLAGS_CONFIG in SlicerConfig.cmake

Sunderlandkyl:
The variables `ep_common_c_flags`, and `ep_common_cxx_flags` were not set in the scope of `SlicerGenerateSlicerConfig.cmake`, so `Slicer_CMAKE_C_FLAGS_CONFIG` and `Slicer_CMAKE_CXX_FLAGS_CONFIG` were set to empty strings.

Calling `find_package(Slicer)` would then set `CMAKE_CXX_FLAGS` and `CMAKE_C_FLAGS` to empty strings whenever it was called.

Fixed by defining `Slicer_CMAKE_C/CXX_FLAGS_CONFIG` in `SlicerGenerateSlicerConfig.cmake`.


CMake/SlicerGenerateSlicerConfig.cmake
```diff
@@ -21,6 +21,8 @@
 set(Slicer_USE_FILE_CONFIG ${Slicer_USE_FILE})
 
 # Compilation settings
+set(ep_common_c_flags "${CMAKE_C_FLAGS_INIT} ${ADDITIONAL_C_FLAGS}")
+set(ep_common_cxx_flags "${CMAKE_CXX_FLAGS_INIT} ${ADDITIONAL_CXX_FLAGS}")
```
jcfr, line 25: Instead let's do this:

```cmake
set(Slicer_CMAKE_CXX_FLAGS_CONFIG    ${CMAKE_C_FLAGS_INIT} ${ADDITIONAL_C_FLAGS})
set(Slicer_CMAKE_CXX_FLAGS_CONFIG    ${CMAKE_CXX_FLAGS_INIT} ${ADDITIONAL_CXX_FLAGS})
```

The variables `ep_common_c/cxx_flags` are specific to superbuild extensions  and shouldn't be part of the SlicerConfig interface

Sunderlandkyl: Sounds good, done!

jcfr:
For now, I suggest you update `SlicerConfig.cmake.in` to simply remove the setting of these variables:

```
CMAKE_C_FLAGS
CMAKE_CXX_FLAGS
Slicer_CMAKE_C_FLAGS
Slicer_CMAKE_CXX_FLAGS
```


### Explanation


Having an other look it turns that the variable `CMAKE_C/CXX_FLAGS` set in `SlicerConfig.cmake.in`

https://github.com/Slicer/Slicer/blob/0560b7fffc371cbbea1bfd79586a21de3be9fb0f/CMake/SlicerConfig.cmake.in#L470-L471

are overridden by code in `UserSlicer.cmake` :

https://github.com/Slicer/Slicer/blob/0560b7fffc371cbbea1bfd79586a21de3be9fb0f/CMake/UseSlicer.cmake.in#L195-L214

This means that the variables that were set in `SlicerConfig` have no effect.

This is assuming everyone is using the  `find_package(Slicer)` with `include(${Slicer_USE_FILE})`

Sunderlandkyl:
OK, variables removed.

The reason that this came up was because I was trying to link to the vtkAddon targets from within IGSIO by calling `find_package(Slicer)`. I didn't think that calling `include(${Slicer_USE_FILE})` would be necessary in that context.

jcfr:
Thanks for updating.

One last nitpick: Commit message should be changed to:

```
COMP: Remove setting of [Slicer_]CMAKE_(C|CXX)_FLAGS in SlicerConfig.cmake

This commit ensures that only calling `find_package(Slicer)` does NOT set
CMAKE_(C|CXX)_FLAGS to empty strings.

Since the variables ep_common_(c|cxx)_flags were NOT set in the scope of
SlicerGenerateSlicerConfig.cmake, variables Slicer_CMAKE_(C|CXX)_FLAGS_CONFIG 
were set to empty strings. Calling find_package(Slicer) would then set 
CMAKE_(C|CXX)_FLAGS to empty strings.

The problem is fixed by removing the setting of Slicer_CMAKE_(C|CXX)_FLAGS_CONFIG
in SlicerGenerateSlicerConfig.cmake, and [Slicer_]CMAKE_(C|CXX)_FLAGS  in 
SlicerConfig.cmake.in.

Indeed, settings these variables is not needed because variables CMAKE_(C|CXX)_FLAGS
are systematically overriden in UseSlicer.cmake

```

Sunderlandkyl:
Done!

PR #4808, opened 2020-04-06
BUG: Reduce startup warnings from qMRMLSliceControllerWidget

### LINKED ISSUE #4807: Runtime warning ctkDoubleSliderPrivate::toInt out of integer bounds
Recently this warning log appears when starting Slicer:

    ctkDoubleSliderPrivate::toInt : value  10  for singleStep  1e-12  is out of integer bounds !

## Steps to reproduce

Start Slicer, check logs. Tested on Windows, in debug mode.

It is probably related to recent changes in how the slice slider position is set (to fix another runtime warning).

Sunderlandkyl:
Starting a debug build of Slicer would result in a warning message, since the number of single steps for the slider would exceed the max integer value.

Fixed in two ways, although either would independently remove the warning:
- Increase the minimum resolution of the slider to a value that does not cause the integer overflow (0.000000000001 -> 0.00000001).
- When calculating the offset resolution, the slice bounds are not taken into account if the size of the bounds is zero.

fixes #4807

### DIFF
--- a/Libs/MRML/Widgets/qMRMLSliceControllerWidget.cxx
+++ b/Libs/MRML/Widgets/qMRMLSliceControllerWidget.cxx
@@ -1284,7 +1284,10 @@ void qMRMLSliceControllerWidgetPrivate::onSliceLogicModifiedEvent()
   // For thin volumes such as ultrasound, the slice bounds may be smaller than the spacing due to floating point errors.
   // To avoid warnings caused by the step size being larger than the slider maximum, set the step size to the smaller of
   // the either the slice offset or the slice bounds depth.
-  offsetResolution = std::min(offsetResolution, sliceBounds[5] - sliceBounds[4]);
+  if (sliceBounds[5] - sliceBounds[4] > 0.0)
+    {
+    offsetResolution = std::min(offsetResolution, sliceBounds[5] - sliceBounds[4]);
+    }
   q->setSliceOffsetResolution(offsetResolution);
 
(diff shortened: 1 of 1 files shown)


Libs/MRML/Widgets/qMRMLSliceControllerWidget.cxx
```diff
@@ -1284,7 +1284,10 @@ void qMRMLSliceControllerWidgetPrivate::onSliceLogicModifiedEvent()
   // For thin volumes such as ultrasound, the slice bounds may be smaller than the spacing due to floating point errors.
   // To avoid warnings caused by the step size being larger than the slider maximum, set the step size to the smaller of
   // the either the slice offset or the slice bounds depth.
-  offsetResolution = std::min(offsetResolution, sliceBounds[5] - sliceBounds[4]);
+  if (sliceBounds[5] = sliceBounds[4] > 0.0)
```
jcfr, line 1287: For sake of clarify, what about changing this to:

```
   if (sliceBounds[4] > 0.0)
     {
     sliceBounds[5] = sliceBounds[4]
     [...]
     }
```

Sunderlandkyl: Whoops, that's actually a typo. Should be:
```if (sliceBounds[5] - sliceBounds[4] > 0.0)```

Thanks for catching it!

PR #4811, opened 2020-04-09
BUG: Return all data nodes loaded by DICOMUtils.loadLoadables

cpinter:
Only the volume nodes were returned by the DICOMUtils.loadLoadables, which was confusing given the documentation and variable names, and also incorrect because many other types of nodes can be loaded from DICOM (fiducials, segmentations, beams, etc.). This fix allows any non-storage and non-display type node to be included in the returned list.


cpinter:
Thanks @lassoan! I was fairly certain that SH nodes are never created during DICOM loading, only items in the singleton node (unlike MRML scene import, when the SH node of the loaded scene is added temporarily). Now I did some testing and it seems to be true. I didn't want to add it just in case, because it might send the wrong message (that we expect them to be added).
I'm integrating this as is, and fix it later if needed for some reason.

PR #4813, opened 2020-04-10
BUG: Change GUI instructions in error message

JEHoctor:
The "Application Settings" menu is under "Edit", not "View". This directs the user to the right part of the UI.

I do not know when this code runs. I do not know if "Application Settings" menu is _always_ under "Edit". This patch is completely untested. It fails the pre-commit hook (line length).


jcfr:

PR #4815, opened 2020-04-13
BUG: Fix failing slicer_util_without_modules test

jamesobutler:
This fixes the failing [py_nomainwindow_test_slicer_util_without_modules](http://slicer.cdash.org/testDetails.php?test=9961479&build=1881141).  The test was failing due to an invalid expected checksum for the `compute-checksum.txt` file.

Since the test was passing as of March 10th and then failing as of March 13th, this corresponds exactly with the transition from SVN to Git.  I believe during this transition `compute-checksum.txt` went from having `LF` line endings with SVN to having `CRLF` line endings with git.

This is similar to #4814 as it is a fix for a test that started to fail due to a change in a file's line endings.

### DIFF
--- a/Base/Python/slicer/tests/compute-checksum.txt
+++ b/Base/Python/slicer/tests/compute-checksum.txt
@@ -1,1 +1,0 @@
-This is a text file!
--- a/Base/Python/slicer/tests/test_slicer_util_without_modules.py
+++ b/Base/Python/slicer/tests/test_slicer_util_without_modules.py
@@ -1,4 +1,6 @@
+import io
 import os
+import tempfile
 import unittest
 import slicer.util
 
@@ -12,12 +14,19 @@ class SlicerUtilWithoutModulesTest(unittest.TestCase):
     with self.assertRaises(IOError):
       slicer.util.computeChecksum('SHA256', 'compute-checksum-nonexistent.txt')
 
-    input_file = os.path.join(os.path.dirname(__file__), 'compute-checksum.txt')
-    self.assertEqual(slicer.util.computeChecksum('SHA256', input_file), '4a57f3207b97f26a6061f86948483c00b03893ddfef9e82b639ebe66e3aba338')
-    self.assertEqual(slicer.util.computeChecksum('SHA512', input_file), '5080ee92e951c5f8336053f11d278c23f5d26b5eb78805c952960eac0194f357f98b0e350611ce081d4a1e28dd8ea182d3a276c99b1752e0def2de0f47b8b27b')
+    with tempfile.TemporaryDirectory() as tmpdirname:
 
-    with self.assertRaises(ValueError):
-      slicer.util.computeChecksum('SHAINVALID', input_file)
+      input_file = os.path.join(tmpdirname, 'compute-checksum.txt')
+      with io.open(input_file, 'w', newline='\n') as content:
+        content.write('This is a text file!\n')
+
+      self.assertTrue(os.path.exists(input_file))
+
+      self.assertEqual(slicer.util.computeChecksum('SHA256', input_file), '4a57f3207b97f26a6061f86948483c00b03893ddfef9e82b639ebe66e3aba338')
+      self.assertEqual(slicer.util.computeChecksum('SHA512', input_file), '5080ee92e951c5f8336053f11d278c23f5d26b5eb78805c952960eac0194f357f98b0e350611ce081d4a1e28dd8ea182d3a276c99b1752e0def2de0f47b8b27b')
+
+      with self.assertRaises(ValueError):
+        slicer.util.computeChecksum('SHAINVALID', input_file)
 
   def test_extractAlgoAndDigest(self):
     with self.assertRaises(ValueError):


hjmjohnson:
@jamesobutler Will you have a chance to address the noted concerns?

jamesobutler:
Yes, I will be able to address this, but only in the evening (5 hours from now).  If you want to address it before then, feel free to go ahead.

jamesobutler:
Well it seemed self-explanatory, but I don't know what to do now after trying a bunch of things.

I've tried using a .gitattributes file like the following example and renormalized things, but it just doesn't work for me on Windows. 
```
# Declare files that will always have LF line endings on checkout.
*.txt text eol=lf 
```

lassoan:
None of our code uses different line endings depending on platform, so it should not be necessary for git to change line endings. We should instruct git to always commit/check out files as is.

jcfr:
> We should instruct git to always commit/check out files as is.

In fact, on windows **text** files are checked out with `windows line endings` and on unix with `unix-line endings`, this the default options when you install git and we should keep it like that.

I will fix the test to write the file while running the test with the expected content and line ending

jamesobutler:
Adding a .gitattributes seems to make sense as I see it present in ITK and also CMake as you pointed out. 

I guess the text file exception is the reason why I wasn't having any success. Thanks for helping out and finding that!

lassoan:
> this the default options when you install git and we should keep it like that.

This has never been a good idea and should not be used for development of platform-independent code. ITK making this mistake is not a good justification either (it will be the problem of ITK developer).

> I will fix the test to write the file while running the test with the expected content and line ending. That way, we don't have to set gitattributes.

Please don't. We should not start introducing platform-dependent behavior unless it is absolutely necessary.

Where would this end? Would we start writing mrml, ini, s4ext, nhdr, stl files, etc. with CRLF line endings, too? Would we upload sample data sets that contain text-like files with LF and CRLF encodings and use different hashes on windows and elsewhere? Implement all file comparisons in platform-dependent way, too? CRLF content would creep into strings stored in memory, too. These would be all terrible.

Please just keep LF everywhere.

jcfr:
Agreed. Since for now, I suspect that on windows files are checked out as CRLF but committed as LF. (this would be the case since `core.autocrlf=true` setting is enabled by default when installing git), there are few options:
1. add `* text=auto` to `.gitattributes`. This ensure file are checked out as CRLF and committed as LF independently of the way git was installed
2. ensure text files are always checked out as `LF` also on windows, we would then expect windows  user to configure their editor to use proper line ending
3. do not prescribe anything regarding text file using `* -text` in `.gitattributes` (this was is done for in Julia language https://github.com/JuliaLang/julia/blob/master/.gitattributes)

For now, I suggest we implement approach (1)

Here are some relevant readings: 
* https://www.edwardthomson.com/blog/git_for_windows_line_endings.html
* https://help.github.com/en/github/using-git/configuring-git-to-handle-line-endings

lassoan:
Sorry, I still don't understand why would you ever want to check out files as CRLF. I've seen Linux/Mac developer tools struggling with CRLF line endings, but nowadays on Windows everything handles files with LF ending. There is really no need to bring in variability of line endings.

Can we force git to checkout/commit all files at Slicer repository level without any line ending manipulation?

jcfr:
> I still don't understand why would you ever want to check out files as CRLF

This is the current behavior on windows, default install of Git has the following:

image

For example, on windows dashboard we have:

image

lassoan:
You don't want to rely on a setting that the user randomly makes _globally_ for all project he is working on. For example, I always use checkout as-is, commit as-is. Users may follow instructions of whatever project's instructions when they install git.

To make the behavior predictable we must set the behavior at the repository level. Either option 2 or 3 would be OK.

jcfr:
>   I always use checkout as-is, commit as-is

Makes sense

> To make the behavior predictable we must set the behavior at the repository level.

Agreed

> Either option 2 or 3 would be OK.

Since all options are enumerated are deterministic, I suggest we instead go with (1). Not doing so means that, source of ITK/VTK/... checkout on windows and sources of Slicer also checkout on windows would have different line ending.

Considering this, I would prefer to be consistent. While not a perfect approach ... I think it will  streamline the developer experience.

lassoan:
Option 1 is not predictable as it produces different files on different platforms.

jamesobutler:
I think a gitattributes for the slicer repo would be good, but I was unable to get it to respect my specification in that file to checkout as LF line endings for text files. It was like it wasn’t overriding the default windows specification of core.autocrlf. Maybe you all can try to get it working, but I had no luck. I committed the file, renormalized things, tried checking out different branches and back to the branch I committed things to, but my editor would still show compute-checksum.txt as having CRLF line endings.

lassoan:
Consistency with VTK would be nice to maintain - but it seems that VTK does not specify crlf for text files (https://github.com/Kitware/VTK/blob/master/.gitattributes).  Overall, for me not having anything (VTK-style) or forcing LF (consistency guaranteed on all platforms) would be both acceptable options.

@jamesobutler Hash files seem to be special case. See [VTK's .gitignore](https://github.com/Kitware/VTK/blob/master/.gitattributes#L23).

jcfr:
@jamesobutler Could you test the updated topic ?  :pray:

jamesobutler:
@jcfr The test on Windows passes for me.

PR #4823, opened 2020-04-15
STYLE: Remove conditional code for CXX< 201103L

hjmjohnson:
Now that C++11 is mandatory, remove conditional code
for older language support.


hjmjohnson:
@jcfr Am I doing something wrong?  It seems that this simple change should not fail so catastrophically.

jcfr:
The CI failing is independent, this will not prevent integration.

Issue #4839, opened 2020-04-17
 error: This CMake does not support the HTTPS protocol.  Ensure that CMake   is compiled with CMAKE_USE_OPENSSL enabled.


MurielCuiCui:
When I build Slicer with CMake, it come out a cmake error,
![image](https://user-images.githubusercontent.com/37406633/79525444-720a9500-8095-11ea-939b-e0720e153ecb.png)
How can I fix it,please help me

jcfr:
It looks like your installation of CMake executable is not able to download files served over https.

From a terminal, could you try to run the following:

```
/path/to/cmake -P /path/to/Slicer/CMake/SlicerCheckCMakeHTTPS.cmake
```

where:

* `/path/to/cmake` is the full path to CMake executable 
* `/path/to/Slicer` is the path to your git checkout of Slicer sources

For example, here is the output of my workstation:

```
~/Software/cmake-3.17.0-Linux-x86_64/bin/cmake -P ~/Projects/Slicer/CMake/SlicerCheckCMakeHTTPS.cmake 
-- Checking if CMake supports https
-- Checking if CMake supports https - ok
```

MurielCuiCui:
I download binary distribution of CMake for win64.  I try what you say,and it seems CMake doesn't support https.  Then I downloaded OpenSSL for windows, CMake again,it doesn't work either.

jcfr:
Since CMake is statically linked against OpenSSL, there is no need to download OpenSSL libraries.

Now,  could you try the following:

1. In your local Slicer checkout, update file [CMake/SlicerCheckCMakeHTTPS.cmake](https://github.com/Slicer/Slicer/blob/master/CMake/SlicerCheckCMakeHTTPS.cmake) like this:

```diff
diff --git a/CMake/SlicerCheckCMakeHTTPS.cmake b/CMake/SlicerCheckCMakeHTTPS.cmake
index 6e28e92d55..4fe209423a 100644
--- a/CMake/SlicerCheckCMakeHTTPS.cmake
+++ b/CMake/SlicerCheckCMakeHTTPS.cmake
@@ -22,7 +22,8 @@ function(slicer_check_cmake_https)
     list(GET status 0 error_code)
     file(REMOVE ${dest})
     if(error_code)
-      message(FATAL_ERROR "error: This CMake does not support the HTTPS protocol. Ensure that CMake is compiled with CMAKE_USE_OPENSSL enabled.")
+      list(GET status 1 error_msg)
[...]

PR #4846, opened 2020-04-17
ENH: Improve SlicerCheckCMakeHTTPS to display underlying error

jcfr:
See #4839


CMake/SlicerCheckCMakeHTTPS.cmake
```diff
@@ -22,7 +22,8 @@ function(slicer_check_cmake_https)
     list(GET status 0 error_code)
     file(REMOVE ${dest})
     if(error_code)
-      message(FATAL_ERROR "error: This CMake does not support the HTTPS protocol. Ensure that CMake is compiled with CMAKE_USE_OPENSSL enabled.")
+      list(GET status 1 error_msg)
+      message(FATAL_ERROR "error: ${error_msg}")
```
lassoan, line 26: It could be still useful to tell users that the error is due to wrong CMake configuration and they may try to download a different version or build on their own. Or, if you don't want to complicate the error message then at least we could write some more information in comments here (developers can find the comments by looking up the filename&linenumber reported by CMake).

jcfr:
Thanks for the review, I added an lined comment. 

If that looks good, I will squash the commits together while integrating

PR #4847, opened 2020-04-17
ENH: Disable backface culling in vtkMRMLModelDisplayNode by default

Sunderlandkyl:


jcfr:
Could this add any performance impact ?

Sunderlandkyl:
Probably it adds some, but I don't notice much of a difference on my desktop.

@lassoan From https://github.com/Slicer/SlicerSurfaceToolbox/pull/17#issuecomment-615348713:
> Disable backface culling in output node by default (maybe disable backface culling by default in model display node, the performance hit nowadays is negligible)

PR #4860, opened 2020-04-18
ENH:select dicom subseries scalar volumes by default

pieper:
When there a scalar volume has the option of being broken
into subseries by a tag such as AcquisitionNumber or
ImageOrientationPatient and there is only such tag a
reasonable heuristic is to prefer loading these subseries
as volumes rather than trying to load the full series, since
the full series is unlikely to be a coherent scalar volume.

This change ups the confidence for loadables for which there
is one set of subseries loadables.

This change also allows a set of loadables that all have the
higest confidence to all be selected.

This change works well on data that was previously corruped on load
and doesn't appear to have an impact otherwise.

An example of a dataset that loads better is series 2 from
patient PD-1-Lung-00008, study 20090123 in this collection:
https://wiki.cancerimagingarchive.net/pages/viewpage.action?pageId=41517500


Modules/Scripted/DICOMLib/DICOMBrowser.py
```diff
@@ -362,8 +362,7 @@ def organizeLoadables(self):
         if loadable.confidence > highestConfidenceValue:
           highestConfidenceValue = loadable.confidence
       for loadable in loadablesBySeries[series]:
-        if loadable.confidence < highestConfidenceValue:
-          loadable.selected = False
+        loadable.selected = loadable.confidence == highestConfidenceValue
```
lassoan, line 365: We had to add this mechanism (keep only a single loadable) for a good reason but I forgot what was that. To avoid this in the future, please add a comment here that explains what the current mechanism does and why (basically what you described in the commit comment).

pieper: Okay, let's keep an eye out for any cases where this leads to trouble.

pieper: Yes, it could happen that too many subseries get loaded, but I'm not a fan of an arbitrary cutoff.  I hope most of those cases would become sequences, or even better get caught by some acquisition-specific code like we had in the Longitudinal PETCT extension.

pieper:
Thanks for the reviews @lassoan!  I went ahead and merged, just so these don't pile up and become stale, but we can always tweak as needed.

lassoan:
Unfortunately, as I was afraid of, this broke DICOM loading. See for example how the data set shared in [this post](https://discourse.slicer.org/t/mouse-pet-ct-to-mri-registration-error/11016/3) shows up:

image

We need to adjust this logic somehow. For example, we could make subseries loading by default if there is spatial overlap within the sequence (slices are in the same position or intersecting).

PR #4861, opened 2020-04-18
ENH: add DICOMUtils.loadByInstanceUID method

pieper:
Sometimes it's useful to store only the instance uid of a slice,
for example with a measurement, but even if you know the
seriesUID for that instance, it doesn't always map to a specific
MRML node, so you may end up loading a full set of nodes
corresponding to different subseries when you only needed
to load the one with this instance.

With this change the loadables are filtered so that only
the highest confidence one containing the instance is
loaded.

### DIFF
--- a/Modules/Scripted/DICOMLib/DICOMUtils.py
+++ b/Modules/Scripted/DICOMLib/DICOMUtils.py
@@ -32,6 +32,10 @@ def loadPatientByUID(patientUID):
       for patientUID in patientUIDs:
         loadedNodeIDs.extend(DICOMUtils.loadPatientByUID(patientUID))
 
+    Note that since there is no way to have a globally unique patient ID,
+    because they are issued by different institutions and there may be
+    name clashes.  db.patients() returns values unique for the current database.
+    See ctkDICOMDatabasePrivate::insertPatient for more details.
   """
   if not slicer.dicomDatabase.isOpen:
     raise IOError('DICOM module or database cannot be accessed')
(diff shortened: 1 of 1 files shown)


Modules/Scripted/DICOMLib/DICOMUtils.py
```diff
@@ -148,6 +152,50 @@ def loadSeriesByUID(seriesUIDs):
...
+  """
+
+  if not slicer.dicomDatabase.isOpen:
+    raise IOError('DICOM module or database cannot be accessed')
+
+  # get the loadables corresponding to this instance's series
+  filePath = slicer.dicomDatabase.fileForInstance(instanceUID)
+  seriesUID = slicer.dicomDatabase.seriesForFile(filePath)
+  fileList = slicer.dicomDatabase.filesForSeries(seriesUID)
+  loadablesByPlugin, loadEnabled = getLoadablesFromFileLists([fileList,])
```
lassoan, line 176: Is the comma needed in `[fileList,]`? I've seen similar syntax at couple of places but normally I don't add comma in such cases and I don't remember having any issues. It is needed when you define a tuple but not for lists, maybe that's the reason? Or is there a special case when the extra comma is useful?

Modules/Scripted/DICOMLib/DICOMUtils.py
```diff
@@ -32,6 +32,10 @@ def loadPatientByUID(patientUID):
       for patientUID in patientUIDs:
         loadedNodeIDs.extend(DICOMUtils.loadPatientByUID(patientUID))
 
+    Note that since there is no way to have a globally unique patient ID,
```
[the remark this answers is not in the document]
pieper: Hmm, I think the rest of the comment actually already says exactly what you describe.  I said "db.patients() returns values unique for the current database" and you said "this patient UID is the internal unique identifier used in the database".

In any case I added this comment since most people have trouble with it, and we really should use patientUID since it's not a valid dicom concept.  So we should no doubt explain it more clearly.  I'll rework the comment.

Modules/Scripted/DICOMLib/DICOMUtils.py
pieper: You're right - it's only needed for the tuple case and not lists.  I can remove it.

PR #4862, opened 2020-04-18
ENH: adds a helper for making numpy changes to scalars

pieper:
Helpful to trigger renders.

E.g. used here:
https://gist.github.com/pieper/f9da3e0a73c70981b48d0747132526d5

### DIFF
--- a/Base/Python/slicer/util.py
+++ b/Base/Python/slicer/util.py
@@ -1020,6 +1020,11 @@ def arrayFromModelPointData(modelNode, arrayName):
   narray = vtk.util.numpy_support.vtk_to_numpy(arrayVtk)
   return narray
 
+def arrayFromModelPointDataModified(modelNode, arrayName):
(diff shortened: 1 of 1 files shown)


Base/Python/slicer/util.py
```diff
@@ -1020,6 +1020,11 @@ def arrayFromModelPointData(modelNode, arrayName):
   narray = vtk.util.numpy_support.vtk_to_numpy(arrayVtk)
   return narray
 
+def arrayFromModelPointDataModified(modelNode, arrayName):
+  """Indicate that modification of a numpy array returned by :py:meth:`arrayFromModelPoints` has been completed."""
+  arrayVtk = modelNode.GetPolyData().GetPointData().GetArray(arrayName)
+  arrayFromModelPointsModified(modelNode)
```
lassoan, line 1026: I would expect here something like `arrayVtk.Modified()`

pieper: Whoops!  Good catch.

Issue #4872, opened 2020-04-20
MRI volume is loaded as several separate small volumes


lassoan:
MRI images are loaded incorrectly (as many separate small volumes) instead of a single volume due to a change in the DICOM module yesterday - see details here: https://github.com/Slicer/Slicer/pull/4860

pieper:
Which MRI volumes?  It doesn't happen for the ones I spot-checked.

As a general fix, maybe we should try limiting to load just subseries where [Acquisition Number](https://dicom.innolitics.com/ciods/mr-image/general-image/00200012) is different.

lassoan:
For some reason, acquisition number _is_ often different within the same series for some reason. I really think we need to check spatial integrity and if it seems that the entire series is one 3D volume (no overlaps between slices) then it should not be split by default.

pieper:
I guess we could only split automatically if there's a geometry warning, we already have that calculation.

lassoan:
That's a good idea! If there is warning without splitting and there is no warning with splitting then it makes sense to split by default.

pieper:
👍 I'm on it.

lassoan:
This should be fixed now.

PR #4873, opened 2020-04-20
BUG: fix dicom subseries priorities

pieper:
This fix loads subseries if there's only one choice and
the there are geometry warnings detected on the full
series.

See: https://github.com/Slicer/Slicer/issues/4872

### DIFF
--- a/Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
+++ b/Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
@@ -160,11 +160,11 @@ class DICOMScalarVolumePluginClass(DICOMPlugin):
     seriesName = self.defaultSeriesNodeName(seriesUID)
 
     # default loadable includes all files for series
-    loadable = DICOMLoadable()
-    loadable.files = files
-    loadable.name = seriesName
-    loadable.tooltip = "%d files, first file: %s" % (len(loadable.files), loadable.files[0])
-    loadable.selected = True
+    allFilesLoadable = DICOMLoadable()
+    allFilesLoadable.files = files
+    allFilesLoadable.name = seriesName
+    allFilesLoadable.tooltip = "%d files, first file: %s" % (len(allFilesLoadable.files), allFilesLoadable.files[0])
+    allFilesLoadable.selected = True
     # add it to the list of loadables later, if pixel data is available in at least one file
 
     # make subseries volumes based on tag differences
@@ -189,7 +189,7 @@ class DICOMScalarVolumePluginClass(DICOMPlugin):
     #
     subseriesFiles = {}
     subseriesValues = {}
-    for file in loadable.files:
+    for file in allFilesLoadable.files:
       # check for subseries values
       for tag in subseriesTags:
         value = slicer.dicomDatabase.fileValue(file,self.tags[tag])
@@ -205,7 +205,7 @@ class DICOMScalarVolumePluginClass(DICOMPlugin):
     loadables = []
 
     # Pixel data is available, so add the default loadable to the output
-    loadables.append(loadable)
+    loadables.append(allFilesLoadable)
 
     #
     # second, for any tags that have more than one value, create a new
@@ -258,20 +258,18 @@ class DICOMScalarVolumePluginClass(DICOMPlugin):
     #
     # now for each series and subseries, sort the images
     # by position and check for consistency
+    # then adjust confidence values based on warnings
     #
     for loadable in loadables:
       loadable.files, distances, loadable.warning = DICOMUtils.getSortedImageFiles(loadable.files, self.epsilon)
 
-    # by default, prefer the all-files loadable
-    loadables[0].confidence = .55
-    if subseriesCount == 1 and loadables[0].warning != "":
+    if subseriesCount == 1 and allFilesLoadable.warning != "":
       # there was a sorting warning and
       # only one kind of subseries, so it's probably correct
-      # to make the subseries a higher confidence than the default all-files version
-      for subseriesLoadable in loadables[1:]:
-        if subseriesLoadable.warning == "":
-          subseriesLoadable.confidence = .55
-          loadables[0].confidence = .45
+      # to have lower confidence in the default all-files version
+      for loadable in loadables:
+        if loadable != allFilesLoadable and loadable.warning == "":
+          allFilesLoadable.confidence = .45
 
     return loadables
 
--- a/Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
+++ b/Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
@@ -262,6 +262,8 @@ class DICOMScalarVolumePluginClass(DICOMPlugin):
     for loadable in loadables:
       loadable.files, distances, loadable.warning = DICOMUtils.getSortedImageFiles(loadable.files, self.epsilon)
 
+    # by default, prefer the all-files loadable
+    loadables[0].confidence = .55
     if subseriesCount == 1 and loadables[0].warning != "":
       # there was a sorting warning and
       # only one kind of subseries, so it's probably correct
@@ -269,9 +271,7 @@ class DICOMScalarVolumePluginClass(DICOMPlugin):
       for subseriesLoadable in loadables[1:]:
         if subseriesLoadable.warning == "":
           subseriesLoadable.confidence = .55
-    else:
-      # prefer the all-files loadable
-      loadables[0].confidence = .55
+          loadables[0].confidence = .45
 
     return loadables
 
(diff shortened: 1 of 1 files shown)


pieper:
This one works for me on the tcia lung ct and on the mouse mr, so it should be good to merge unless you see issues or have an idea to catch more special cases.

pieper:
Actually hold on - the logic is not quite right.

lassoan:
Test this with the MRI stored in this data set: https://discourse.slicer.org/t/mouse-pet-ct-to-mri-registration-error/11016/5?u=lassoan

pieper:
Yes, I did test with that mouse mr.

pieper:
Okay, this should be good to go.

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
```diff
@@ -268,6 +262,17 @@ def examineFiles(self,files):
...
 
+    # by default, prefer the all-files loadable
+    loadables[0].confidence = .55
+    if subseriesCount == 1 and loadables[0].warning != "":
+      # there was a sorting warning and
+      # only one kind of subseries, so it's probably correct
+      # to make the subseries a higher confidence than the default all-files version
+      for subseriesLoadable in loadables[1:]:
+        if subseriesLoadable.warning == "":
+          subseriesLoadable.confidence = .55
```
lassoan, line 273: Is there a specific reason we both increase subseries confidence and decrease loadables confidence? It may be simpler for other DICOM plugins to coexist with the default Scalar Volume plugin if it did not use many different confidence values.

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
```diff
@@ -268,6 +262,17 @@ def examineFiles(self,files):
     for loadable in loadables:
       loadable.files, distances, loadable.warning = DICOMUtils.getSortedImageFiles(loadable.files, self.epsilon)
 
+    # by default, prefer the all-files loadable
+    loadables[0].confidence = .55
```
lassoan, line 266: At this point, do we know that all loadables belong the same series and loadables[0] refers to the full series?

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
```diff
@@ -268,6 +262,17 @@ def examineFiles(self,files):
...
+    if subseriesCount == 1 and loadables[0].warning != "":
```
lassoan, line 267: We may have more subseries if "allowLoadingByTime" is set to True. This means that if the user allows loading subseries by time then it may seemingly break loading of volumes that were previously loaded correctly (split based on AcquisitionNumber). To prevent this, what do you think about putting subseries keys in decreasing priority order (I think the current order would be correct), and using the _first_ set of split subseries (corresponding to the highest priority key)?

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
pieper: It really doesn't matter, because currently we only use the max anyway.  But in the bigger picture I think it's correct to have a below-default confidence in this case.  Someday we may want to use this information for something so it should reflect something about the result of examining the files.

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
pieper: Yes, that's the case.

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
pieper: That could work, but wouldn't it be better to generate a sequence if we detect a subseries that is based on time?

On a similar note, I'm thinking that it's pretty awkward to have one checkbox per subseries in the advanced panel of the browser - makes it hard to switch between choices.  Perhaps we should make one compound loadable that represents a subseries option.  Or maybe the subseries should always be sequences with the frames flagged by the split value.

lassoan: One checkbox per subseries might be useful in some cases (to be able to load particular time points or orientations), but I agree that it can be quite inconvenient, especially when it is more than a few nodes. I like both ideas - compound loadable and sequence. Since it is easier to split a sequence to nodes than construct a sequence from a set of nodes, and sequences work much better when there are more than a few items, I think it would be better to always load a subseries as a sequence.

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
lassoan: OK, then we should keep the confidence value at 0.5 for the loadable that we are most confident about and slightly lower it for the others.

It is better not to increase confidence above 0.5 because that range is quite crowded already (there are plugins that have set their own confidence at slightly higher than 0.5 to take over loading from the default volume reader).

pieper: Okay, that's fair - the scalarvolume plugin is meant to be the fallback if no other plugins grab the series.

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
pieper: I agree though that relying on the order in the list could be fragile.  I'll save a dedicated variable pointing to the all-files loadable.

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
pieper: Do you have a sequence importer already?  The DICOMVolumeSequencePlugin appears to only do export.

lassoan: Yes, the sequence importer is in https://github.com/fedorov/MultiVolumeImporter/blob/master/MultiVolumeImporterPlugin.py

In fact, this importer already splits a series into subseries based on selected tags. Maybe we should just remove splitting by subseries from scalar volume plugin and let the multivolume importer handle this?

pieper: Yes, I like that in general for things that are naturally sequential like perfusion.  I notice that MultiVolumeImporter doesn't use AcquisitionNumber to detect sequences, and that may be okay.  The lung CT data has different acquisitions that correspond to different parts of the body at different FOVs so maybe more natural as multiple scalar volumes.  So I think we should just wrap this up and leave it for now, but also make sure the confidence for Sequence loadables is higher for any of the time-related subseries.

By the way, testing the MultiVolumeImporter it generates this message at the end (loading works because it's completed before this error).


```
DICOM plugin failed to load 'dMRI_dir107_PA - as a 108 frames Volume Sequence by ImagePositionPatient+InstanceNumber' as a 'MultiVolume'.
  File "/home/researcher/Downloads/Slicer-4.11.0-2020-04-19-linux-amd64/bin/../lib/Slicer-4.11/qt-scripted-modules/MultiVolumeImporterPlugin.py", line 700, in load
    sequenceBrowserModule = slicer.modules.sequencebrowser
AttributeError: module 'modules' has no attribute 'sequencebrowser'

```
I'm guessing you know an easy fix for this?
https://github.com/fedorov/MultiVolumeImporter/blob/master/MultiVolumeImporterPlugin.py#L699

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
pieper: Okay, this is now updated

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
pieper: Moved this to a new issue: https://github.com/fedorov/MultiVolumeImporter/issues/41

I'll go ahead and merge this, since we know it addresses an issue in the current master.  We can open something new if more corner cases come up.

lassoan:
I agree that this should work for now. We can consolidate the scalar volume and multivolume plugins later.

cpinter:
In the latest preview release (r29000 / dfef957) I encountered a problem with a CT that used to be loaded as a normal volume, but now it is loaded as separate volumes for each slice. I can fix it by switching to advanced mode in the DICOM browser and select only the complete volume, but I think it is very confusing for the average user.
Is this a remaining bug, or intended?

image

pieper:
With the current master this should only happen if the all-slices volume had geometry warnings but the subseries versions did not.  What warning was on the complete volume?

cpinter:
I don't see any warnings in the log or the python window, but it still loads all loadables (the complete CT and the per-slice ones as well).

pieper:
Thanks for the report!

I went ahead and pushed a fix to master.

https://github.com/Slicer/Slicer/commit/ad521846728894f08ea520100f16467fcc318986

cpinter:
Thanks a lot, @pieper! I will try tomorrow's nightly and let you know

cpinter:
@pieper your last change fixed the particular issue that I had. Thank you so much for the super fast fix!

PR #4878, opened 2020-04-21
Use equal default for C++11

hjmjohnson:


jcfr:
Thanks :pray: 

I will try to build the topic later to confirm the VTK wrapping behave.

hjmjohnson:
@jcfr Are there issues with VTK wrapping?

jcfr:
Looking at the git history associated with the VTK wrapping tools, I saw some changed related to handling of C++11 features (enum, default initializer, ...)

This is why I think we shouldn't rush that change

hjmjohnson:
@jcfr Can we please take this to completion?  If you want to reject it, then let's take care of that now. Monitoring outstanding patches is time-consuming.

lassoan:
@jcfr are there things that you are sure that are safe to change or things that should not be changed? So many files have changed so that if these changes are not merged soon then it'll be very hard to merge them due to conflicting changes.

jcfr:
I am building locally and testing, I will integrate when this is done.

PR #4879, opened 2020-04-21
WIP: Detect and prohibit symbolic links in build path

hjmjohnson:
Symbolic links in the build path are not supported
due to incorrect build-time relative path resolution
that is needed.  The relative paths are not stable
across the symlinks.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,6 +50,15 @@ if(NOT DEFINED Slicer_VERSION_PATCH)
 endif()
 project(Slicer VERSION "${Slicer_VERSION_MAJOR}.${Slicer_VERSION_MINOR}.${Slicer_VERSION_PATCH}")
 
+
+#-----------------------------------------------------------------------------
+# Slicer builds fail when build directory is a simlink
+get_filename_component(Slicer_BINARY_DIR_realdir "${Slicer_BINARY_DIR}" REALPATH)
+if( NOT "${Slicer_BINARY_DIR_realdir}" STREQUAL "${Slicer_BINARY_DIR}")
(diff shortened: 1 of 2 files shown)


jcfr:
To follow up on this, there is related to an underlying issue in CMake, see https://gitlab.kitware.com/cmake/cmake/-/issues/16228

Based on my discussion with @bradking not even sure the detection we want to implement will work:

> > detect if CMAKE_BINARY_DIR includes symlinks and aborts ?

> Yes, though the very same bug may cause get_filename_component REALPATH
> to actually put a symlink back in to the path and cause the detection
> not to work.



@hjmjohnson To address the particular issue you reported, I will update the build system to check if `lsb_release` works with python3 and skip the generation of that additional launcher if it is the case.

jcfr:
Closing. Superseded by #4887 4887

hjmjohnson:
@jcfr Thanks! Slicer builds worked for me on the previously failing machine.  (not a clean build, but no longer failing as it used to).

jcfr:
Thanks again for your patience with this 🙏 

Are you able to run both Slicer and also PythonSlicer (found in python-install/bin in the build tree)

hjmjohnson:
```
./python-install/bin/PythonSlicer  
Python 3.6.7 (default, Apr 18 2020, 17:42:36) 
[GCC Clang 11.0.0 ] on linux2
Type "help", "copyright", "credits" or "license" for more information. 
```

I don't have access to the machine to test the Slicer binary. I only have ssh access at the moment.

jcfr:
Great, since `PythonSlicer` is also a launcher ... i am glad it also works in your case :tada:

PR #4880, opened 2020-04-22
Unstructured grid probe fixes

lassoan:
Various fixes related to unstructured grid loading, display, and probing.


lassoan:
Merging this now so that it gets into tonight's release.

PR #4882, opened 2020-04-22
BUG: Fix crash in qMRMLSubjectHierarchyModel

Sunderlandkyl:
If a null QStandardItem* was passed to updateSubjectHierarchyItemFromItemData, it would cause a crash.
Fixed by adding a null-pointer check and an error message.

### DIFF
--- a/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyModel.cxx
+++ b/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyModel.cxx
@@ -1219,6 +1219,11 @@ void qMRMLSubjectHierarchyModel::updateSubjectHierarchyItemFromItemData(vtkIdTyp
     qCritical() << Q_FUNC_INFO << ": Invalid subject hierarchy";
     return;
     }
+  if (!item)
+    {
+    qCritical() << Q_FUNC_INFO << ": Invalid item";
+    return;
+    }
 
   qSlicerSubjectHierarchyAbstractPlugin* ownerPlugin =
     qSlicerSubjectHierarchyPluginHandler::instance()->getOwnerPluginForSubjectHierarchyItem(shItemID);


cpinter:
What was the use case in which this crashed?

Sunderlandkyl:
It happened when I accidentally re-parented some markup nodes onto each other when I was trying to delete them, and then when I deleted them, it caused the crash. I spent some time last night trying to reproduce the exact issue, but wasn't able to.

It's related to drag and drop since it happens during delayedItemChanged(). I think it has to do with multiple items being re-parented in the widget and then deleted in quick succession before the delayedItemChanged().

Managed to reproduce it this morning, and encountered another crash (https://github.com/Slicer/Slicer/pull/4884).

Here's how I managed to reproduce it this morning:
- Change the delay at this line to some larger number to make it easier to cause the issue (I used 5 seconds).
https://github.com/Slicer/Slicer/blob/1bc83586fc406a8f4cb86250e683eac34ae22980/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyModel.cxx#L1595
- Add 2 nodes to the scene (I used markup nodes)
- Re-parent one of the nodes onto the other
- Delete one of the nodes

Here is the stack trace:
>	qSlicerSubjectHierarchyModuleWidgets.dll!qMRMLSubjectHierarchyModel::updateSubjectHierarchyItemFromItem(__int64 shItemID, QStandardItem * item) Line 1161	C++
 	qSlicerSubjectHierarchyModuleWidgets.dll!qMRMLSubjectHierarchyModel::delayedItemChanged() Line 1615	C++
 	[External Code]	
 	qSlicerBaseQTGUI.dll!qSlicerApplication::notify(QObject * receiver, QEvent * event) Line 394	C++
 	[External Code]	
 	qSlicerBaseQTCore.dll!qSlicerCoreApplication::exec() Line 773	C++
 	SlicerApp-real.exe!`anonymous namespace'::SlicerAppMain(int argc, char * * argv) Line 62	C++
 	[External Code]

Issue #4681, opened 2020-03-13
No /usr/include on macOS Mojave


slicerbot:
_This issue was created automatically from an original [Mantis Issue](https://mantisarchive.slicer.org/view.php?id=4681). Further discussion may take place here._

pieper:
@Slicer/slicer-developer does anyone have an idea how best to fix this?  To me this seems like something to fix at the CMake level, but I guess we could just set the environment variable as part of our build.

ihnorton:
Update [this](https://github.com/Slicer/Slicer/blob/47deb76d7556e40de4e25e585c4b24a63a153da5/CMake/SlicerInitializeOSXVariables.cmake#L43) to use/require [`CMAKE_OSX_SYSROOT`](https://cmake.org/cmake/help/latest/variable/CMAKE_OSX_SYSROOT.html) instead of `CMAKE_OSX_DEPLOYMENT_TARGET`? Not clear to me if `ExternalProject_Add` automatically passes that down to the sub-project (it is not currently passed through by Slicer AFAICT).

jcfr:
Since:
* numpy is now installed from a wheel, the numpy build problem reported in the original is obselete. 
* this issue affect only non CMake project. Indeed, having `CMAKE_OSX_SYSROOT` set is sufficient for CMake project.

Could you test the following patch:

```diff
diff --git a/CMake/ExternalProjectForNonCMakeProject.cmake b/CMake/ExternalProjectForNonCMakeProject.cmake
index efafbdfa75..a316cb30bd 100644
--- a/CMake/ExternalProjectForNonCMakeProject.cmake
+++ b/CMake/ExternalProjectForNonCMakeProject.cmake
@@ -215,6 +215,10 @@ if(APPLE)
   endif()
   set(CMAKE_OSX_FLAGS \"\${osx_arch_flags} \${osx_version_flag} \${osx_sysroot}\")
 
+  if(NOT DEFINED ENV{SDKROOT})
+    set(ENV{SDKROOT} \"\${CMAKE_OSX_SYSROOT}\")
+  endif()
+
   _ep_var_append(_ep_CFLAGS \"\${CMAKE_OSX_FLAGS}\")
   _ep_var_append(_ep_CXXFLAGS \"\${CMAKE_OSX_FLAGS}\")
   _ep_var_append(_ep_LDFLAGS \"\${CMAKE_OSX_FLAGS}\")

```

pieper:
[...]

PR #4857, opened 2020-04-18
Fix OpenSSL build issues on Mac 10.15 XCode 11 clean OS

hjmjohnson:
The SDK must be provided in the environment for OpenSSL to compile.

OpenSSL include directory name is 'openssl' all lower case letters.  Use that to avoid 100's of compiler warnings about case mis-match.

PR #4883, opened 2020-04-23
COMP: Fix OpenSSL build on macOS setting SDKROOT in build script

jcfr:
This commit fixes a regression introduced in  ddeb0a33a (COMP: Fix
OpenSSL build on Mac 10.15 XCode 11.4.1 (#4857)) originally intended
to fix issue #4681

Setting the SDKROOT environment variable worked as expected on a regular
developer build but failed on a dashboard build where the invocation of
"make" is wrapped using a "launcher" [1]

This commit addresses the issue by writing a CMake script that will set
the environment and then invoke make.

Last, since  file descriptors expect by the make jobserver are not kept
open by the CMake process, the build script explicitly unset MAKEFLAGS
to avoid "warning: -jN forced in submake: disabling jobserver mode."
For more details, see [2]

[1] https://cmake.org/cmake/help/latest/manual/ctest.1.html#ctest-build-step
[2] https://stackoverflow.com/questions/29910944/is-it-possible-to-pass-through-gnu-make-jobserver-environment-to-a-submake-ser

### DIFF
--- a/SuperBuild/External_OpenSSL.cmake
+++ b/SuperBuild/External_OpenSSL.cmake
@@ -91,6 +91,17 @@ set(ENV{VS_UNICODE_OUTPUT} \"\")
   # This is required when building Slicer using a cronjob where the
   # default environment is restricted.
   set(ENV{PATH} \"/usr/sbin:\$ENV{PATH}\")
+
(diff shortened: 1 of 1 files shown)


jcfr:
Merging, change tested on Linux and macOS

PR #4890, opened 2020-04-25
ENH: Always use filenames that user typed in save data dialog

lassoan:
When user typed an invalid filename in save data dialog (such as just a filename base without extension or with an invalid extension) then the validator rejected it and user had to either forced to type the extension or revert to the original value. When user clicked on save after entering an invalid filename (without pressing enter or escape) then the filename was reverted to the original without warning to the user.

It was also not possible to choose a different file extension just by typing a filename with a different extension.

This commit fixes these problems and makes consistent with other software (Notepad, Word, Visual Studio Code, Gimp, Camtasia, Snagit):
- If a known file extension is entered but not the one that is selected in the listbox then the entered filename is used as is.
- If an unknown file extension is entered then the file extension selected in the listbox is appended to the entered filename.

See https://discourse.slicer.org/t/confusing-save-dialog-behavior/11251/15 for discussion.

### DIFF
--- a/Base/QTGUI/qSlicerSaveDataDialog.cxx
+++ b/Base/QTGUI/qSlicerSaveDataDialog.cxx
@@ -74,67 +74,42 @@ qSlicerFileNameItemDelegate::qSlicerFileNameItemDelegate( QObject * parent )
 }
 
 //-----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


pieper:
Thanks - I didn't have a chance to build it but the code looked good.

PR #4896, opened 2020-04-28
BUG: Fix segment closed surface display after smoothing/margin effect

### LINKED ISSUE #4895: Using Smooth effect in Segment Editor hides the segment in 3D
In Segment Editor, when using the Smooth effect (any method), the segment that was being smoothed disappears from 3D. Hiding and showing the 3D view solves it. It seems to me that the closed surface representation is being removed but not re-converted.

## Steps to reproduce

- Smooth a segment in Segment Editor

## Expected behavior

- The segment is smoothed and the result can be seen in both 2D and 3D (if turned on)

## Environment
- Slicer version: Slicer-4.11.0 r29005
- Operating system: Windows

Sunderlandkyl:
Invokes RepresentationModified after derived representations are reconverted in vtkSlicerSegmentationsModuleLogic::SetBinaryLabelmapToSegment(). This ensures that the display manager is kept up-to-date.
Fixes #4895


cpinter:
@Sunderlandkyl Thank you so much for fixing this so quickly!

Issue #4772, opened 2020-03-26
Move legacy Editor module to separate extension


lassoan:
Move Editor module to separate extension.

lassoan:
@jcfr could you copy the editor module (Modules\Scripted\Editor and Modules\Scripted\EditorLib ) to a separate extension with history? After that I can take care of the rest.

jamesobutler:
Given the discussion at https://github.com/Slicer/Slicer/pull/4904#issuecomment-956351314, should this issue be tagged with the `status:wontfix` label and then be closed?

PR #4904, opened 2020-04-30
Remove legacy modules

lassoan:
This is a proposal for removing legacy Editor module and Charts infrastructure from Slicer. Editor is replaced by Segment Editor module. Charts is replaced by Plots module. By removing them, we can reduce Slicer core maintenance efforts and force extension developers to use up-to-date infrastructure.

Slicer5 may be a good opportunity to make such breaking changes rather than removing them in an upcoming patch release, but this is up for discussion.

We could make some features available in an extension (SlicerLegacyModules), but it would require some effort, and some features would be very hard to implement (e.g., CLI support of double arrays). Therefore, it would be better if this would not be necessary and instead we could upgrade all extensions to not rely on legacy modules.


jcfr:
i am currently swamped until Monday, I will review this along with the other issues you asked me question for on Monday.

pieper:
Yes, I'm concerned about extensions and also about users who rely on old modules for their own work.  Good to give people some advanced warning.

lassoan:
OK. I'm building the Extension index to see what the impact is.

fedorov:
Related to this, are there any maintenance commitments for the earlier releases? Can we say that the last Slicer 4 release will be maintained for some period of time, so that the users of legacy extensions who for whatever reason cannot migrate will be able to use legacy modules in the latest Slicer 4 release for some time?

lassoan:
I don't think free maintenance would be offered for Slicer4. Paying Slicer core maintenance and development from various grants is already challenging enough. Spending time with Slicer4 would directly hinder Slicer stable maintenance and Slicer preview enhancements, so I would avoid it.

lassoan:
Uses legacy charts infrastructure:
- PerkTutor: actively maintained, need to be upgraded to use tables instead of double arrays
- SlicerRT: actively maintained, need to be upgraded to use tables instead of double arrays
- Chest_Imaging_Platform
- FiberViewerLight
- IASEM
- LongitudinalPETCT
- OpenCAD
- PetSpectAnalysis
- SlicerAstro
- MarginCalculator: obsolete, can be removed

Uses legacy editor infrastructure:
- Cardiac_MRI_Toolkit
- CardiacAgatstonMeasures
- Chest_Imaging_Platform
- IASEM
- iGyne
- LongitudinalPETCT
- mpReview
- NeedleFinder
- PET-IndiC
- SegmentationWizard
- SlicerPathology
- CarreraSlice: obsolete, can be removed
- GraphCutSegment: obsolete, can be removed
- WindowLevelEffect: obsolete, can be removed

pieper:
Interesting - thanks for running those tests.  Based on the results I suggest we consider leaving the code in place for the time being.  Some of those extensions are unlikely to be updated soon.  Perhaps we can flag them in the extension interface as legacy extensions that we may not be able to help support.  We can also hide any user level access to the legacy modules and generate warnings when the code is used.

lassoan:
OK, I'll submit an issue to each extension's repository about plans for removing these components and give some time for everybody to respond.

jamesobutler:
> Some of those extensions are unlikely to be updated soon.

Can't users still continue to use Slicer4 if they want to use these extensions? Or would the download access for Slicer4.10 extensions using extension manager being taken down?

I would expect if the legacy modules would be remove, these extensions would be removed from the supposed "Slicer 5.0" branch of the extensions index and then added again once they are updated to be compatible with Slicer 5.0.

lassoan:
I've posted issues for all the impacted extensions:

- PerkTutor: https://github.com/PerkTutor/PerkTutor/issues/68
- SlicerRT: https://github.com/SlicerRt/SlicerRT/issues/142
- Chest_Imaging_Platform: https://github.com/acil-bwh/SlicerCIP/issues/29
- FiberViewerLight: https://github.com/NIRALUser/FiberViewerLight/issues/3
- IASEM: https://github.com/blowekamp/Slicer-IASEM/issues/8
- LongitudinalPETCT: https://github.com/QIICR/LongitudinalPETCT/issues/17
- OpenCAD: https://github.com/vnarayan13/Slicer-OpenCAD/issues/8
- PetSpectAnalysis: https://github.com/gti-fing/SlicerPetSpectAnalysis/issues/2
- SlicerAstro: https://github.com/Punzo/SlicerAstro/issues/109
- Cardiac_MRI_Toolkit: https://github.com/carma-center/carma_slicer_extension/issues/5
- CardiacAgatstonMeasures: https://github.com/BRAINSia/CardiacAgatstonMeasures/issues/1
- iGyne: https://github.com/gpernelle/iGyne/issues/2
- mpReview: https://github.com/SlicerProstate/mpReview/issues/168
- NeedleFinder: (issues disabled)
- PET-IndiC: https://github.com/QIICR/PET-IndiC/issues/6
- SegmentationWizard: https://github.com/QTIM-Lab/SlicerSegmentationWizard/issues/2
- SlicerPathology: https://github.com/SBU-BMI/SlicerPathology/issues/103
- CarreraSlice: https://github.com/ikolesov/CarreraSlice/issues/12
- GraphCutSegment: https://github.com/Slicer/SlicerGraphCutSegment/issues/2

> Can't users still continue to use Slicer4 if they want to use these extensions? Or would the download access for Slicer4.10 extensions using extension manager being taken down after Slicer 5.0 becomes stable?

Yes, Slicer-4.10 will remain available, but we'll not rebuild all the extensions every night for it and if any compatibility issues come up then we won't be able to fix them.

> I would expect if the legacy modules were removed, these extensions would then be removed from the supposed "Slicer 5.0" branch of the extensions index and then added again once they are updated to be compatible with Slicer 5.0.

I agree, it would make sense to review the extensions list for Slicer5 and remove those that are known to be not compatible.

pieper:
> 
> >     Can't users still continue to use Slicer4 if they want to use these extensions? Or would the download access for Slicer4.10 extensions using extension manager being taken down after Slicer 5.0 becomes stable?
> 
> Yes, Slicer-4.10 will remain available, but we'll not rebuild all the extensions every night for it and if any compatibility issues come up then we won't be able to fix them.

Agreed, plus sometimes users need to get bug fixes and features from the latest slicer and use them with legacy modules/workflows.  People can sometimes get by with running two versions at the same time, but passing data back and forth can be a challenge.  I wouldn't want to put effort into maintaining old code, but I feel we should leave them in place until we know it causes trouble somehow.

cpinter:
SlicerRT does not use the Charts infrastructure. I did this migration in https://github.com/SlicerRt/SlicerRT/commit/8f9155f94399be71e747d3d45b1f6c4152caa139
It does use vtkMRMLDoubleArrays, but it will be quick to change.

cpinter:
A perhaps wild idea: couldn't we release a 4.12 just before removing all these modules and the other changes that affect the users most (such as switch to dark style)? That way we'd have the old infrastructure and a snapshot of a pretty stable Slicer4, and we wouldn't have to force the people who cannot use Slicer5 for some reason to use 4.10, which is 1.5 years old now, and is seriously behind in features compared to the nightly.

lassoan:
We have made lots of breaking changes in 4.11 already (Python 2->3, Sequences, RAS/LPS, etc.). It would be better if we could say that we only make such breaking changes in major releases.

If we find that removal of these classes would cause major problem for extension developers then we can still delay this a (e.g., just hide them from the GUI as Steve suggested; or work with extension developers who still maintain their extensions and make sure they upgrade before we remove anything).

jamesobutler:
If there is a Slicer 4.12, will this earlier statement

> Slicer-4.10 will remain available, but we'll not rebuild all the extensions every night for it and if any compatibility issues come up then we won't be able to fix them.

become the following? 

"Slicer-4.12 will remain available, but we'll not rebuild all the extensions every night for it and if any compatibility issues come up then we won't be able to fix them." 

This would be when stable Slicer 5.0 is released shortly after the Slicer 4.12 release

fedorov:
> I've posted issues for all the impacted extensions

@lassoan do you expect transition would need to be drastically different for those extensions? Maybe you could post instructions for transition on some example? 

At least some of those extensions have been developed under projects that ended, and are not backed up by a maintenance team. Original authors may no longer be around. So having some kind of guidance on what should be done to switch to SegmentEditor would definitely not hurt.

Currently, the issues you posted suggest questions should be asked on the developer forum. It might be better if instead sample transition could be posted on the forum, and if anything is not covered by those instructions, developers of the individual extensions could follow up, so there is a single thread of communication related to this effort.

lassoan:
Thank you @fedorov , good points. A migration guide is useful if many developers need to upgrade many extensions in a similar way.

I've checked now and unfortunately all extensions (except 4) are abandoned. The submitted issues will confirm this. Since all these extensions already have build errors, they don't even show up in the extension manager; or they don't have build errors (because they only contain Python scripted modules) but the module fails to load. Regardless of keeping around legacy modules or not, these extensions cannot be used with Slicer-4.11 for more than a year.

There are only four extensions that currently work and could be impacted by removal of legacy modules. PerkLab will upgrade PerkTutor and SlicerRT.

**So, we are really just talking about what to do with PET-IndiC and mpReview extensions.** Since upgrading these extensions would be probably less work and provide a long-term solution, I think that is the most reasonable way forward. I can help with the upgrade as needed - for example, I can give it a go and upgrade from Editor to Segment Editor, then maintainers review/test/fix up things.

fedorov:
@lassoan this makes sense, at lest for the extensions you mentioned. It is very generous of you to offer to help with the upgrade. I would definitely welcome such help - I have no doubt you would spend order of magnitude less time doing that than me (the only person working with me right now has no Slicer development experience, and has no relevant funding to work on mpReview). Let me know how you would like to proceed.

Adding @chribaue, since he developed PET-IndiC as part of the now expired QIICR project.

fedorov:
@lassoan let's continue the mpReview discussion in https://github.com/SlicerProstate/mpReview/issues/168.

hjmjohnson:
> * CardiacAgatstonMeasures: [BRAINSia/CardiacAgatstonMeasures#1](https://github.com/BRAINSia/CardiacAgatstonMeasures/issues/1) (single Python2 scripted module, last commit 5 years ago)

This will not be upgraded.  That project is no longer being pursued. Please remove from the extension manager.

hjmjohnson:
> > * CardiacAgatstonMeasures: [BRAINSia/CardiacAgatstonMeasures#1](https://github.com/BRAINSia/CardiacAgatstonMeasures/issues/1) (single Python2 scripted module, last commit 5 years ago)
> 
> This will not be upgraded. That project is no longer being pursued. Please remove from the extension manager.

SEE:  https://github.com/Slicer/ExtensionsIndex/pull/1702

jamesobutler:
@lassoan Is this PR still on target for the [4.13 milestone](https://github.com/Slicer/Slicer/milestone/1), or is this happening as part of the [5.0.0 milestone](https://github.com/Slicer/Slicer/milestone/4) ?

chribaue:
@lassoan A while ago you mentioned that the Slicer team might move the Legacy Editor into a separate extension rather than dumping the whole module, so that you don't have to maintain it as part of the Slicer core anymore. This way interested developers would still be able use the Legacy editor in their extensions, but they would be responsible for fixing issues in this Legacy editor extension.

Is this still the plan?

jamesobutler:
@chribaue From what was discussed in #5962, since Slicer core developers weren't going to maintain a Legacy "Editor" extension it didn't make sense for them to proceed with that transfer work #4772. Slicer developers would spend time helping others transition extensions that are currently using the legacy "Editor" module to instead use the "Segment Editor" module. This time spent would be more valuable to others than spending time transferring data into a new repository. I suppose if you're interested in maintaining an extension with Legacy Editor module and don't want to use "Segment Editor", then if you can volunteer your time of transferring the Editor module and setting it up as a new extension that would be appreciated!

lassoan:
>  interested developers would still be able use the Legacy editor in their extensions, but they would be responsible for fixing issues in this Legacy editor extension.
Is this still the plan?

Plans have changed. Developers of extensions that used the legacy Editor either upgraded to Segment Editor or did not show interest in maintaining their extension at all (except the mpReview and PET extensions). Since fixing the currently broken Editor module and splitting it out into a separate extension would be very significant effort and would only help 1-2 projects, and only temporarily (until those move to Segment Editor, too), it is much better use of everyone's time to move those projects to use Segment Editor.

chribaue:
Thank you for the update!

Our 2 extensions (PET-IndiC and PETTumorSegmentation) that used the Legacy Editor are also fully functional without the Legacy Editor. We're ready to live without the Legacy Editor. It was just a nice option to have for some people who were used to it. Once you remove the Legacy Editor from Slicer, I'll remove the modules that use it from our extensions.

fedorov:
@lassoan I have a fellow who may be able to contribute to migration of mpReview, but my question from May 2020 remains - where does one start with such migration? Is there a guide, or example of a module that was successfully migrated, which could be suitable for someone who is willing to learn, but is not an expert Slicer developer?

lassoan:
> where does one start with such migration?

I think the main change is to switch from using labelmap volume to segmentation node.

One simple approach, if you want to make the transition with minimum amount of change in the existing module then you can add support for segmentation nodes (and keep supporting labelmap volume nodes, too). For example, if your module used a labelmap volume node as input, you can update the node selector so that it takes a segmentation node as well. If the user selects a segmentation node input then you can export it to a temporary labelmap volume node and run the processing the same way as before.

If you want to simplify your code and/or take full advantage of all the additional features that segmentations provide (3D display of segmentation, overlapping segments, using DICOM terminologies for defining segment content, smooth DICOM Segmentation Object import/export, etc.) then you may decide drop support for labelmaps and work only with segmentations.

lassoan:
> could be suitable for someone who is willing to learn, but is not an expert Slicer developer?

I would also be happy to help that person. Maybe starting with a meeting reviewing the project requirements and priorities, because there are a lot of features in mpReview that make the extension update more complicated and might not that important anymore.

fedorov:
Andras, thank you! Let me discuss this internally, and I will follow up. 

No matter what happens with mpReview, I believe the decision to remove legacy Editor is the right one. I want to make sure my comments above are not interpreted to imply the opposite.

jcfr:
@lassoan  Thanks again for moving forward with this :100: :pray: 

When you have a chance, could you also update these wiki entries accordingly:

* https://www.slicer.org/wiki/Documentation/Nightly/Developers/Tutorials/MigrationGuide#Slicer_5.0:_Removed_Editor_module

* https://www.slicer.org/wiki/Documentation/Nightly/Developers/Tutorials/MigrationGuide#Slicer_5.0:_Removed_Charts_and_DoubleArrays_module

* It may also be sensible to reference the MRML nodes and relevant API that were removed in https://www.slicer.org/wiki/Documentation/Nightly/Developers/Tutorials/MigrationGuide#Slicer_5.0:_API_changes_since_4.10

lassoan:
I've added this section. Let me know if there is anything else that should be added.

PR #4921, opened 2020-05-08
BUG: Make using long paths safe in python !WIP!

cpinter:
DICOM databases are sometimes stored in long paths, which are not handled well on Windows, and the result in these cases is files not found.

The new longPath function in slicer.util converts the path string to one that can be safely used on Windows for paths longer than 260 characters.

Probably this won't work like this on Mac and Linux, so if after the discussion we decide to use this, then it will have to be changed to be only used on Windows.

Note that in the PR I moved some functions in slicer.util that were on the bottom of the file in the wrong group (there are groups in util.py like UI, Modules, VTK, etc.) to their corresponding groups.

### DIFF
--- a/Base/Python/slicer/util.py
+++ b/Base/Python/slicer/util.py
@@ -376,6 +376,68 @@ def setSliceViewerLayers(background='keep-current', foreground='keep-current', l
         if sliceLogic:
           sliceLogic.FitSliceToAll()
 
+def setToolbarsVisible(visible, ignore=None):
+  """Show/hide all existing toolbars, except those listed in
+  ignore list.
+  """
+
+  for toolbar in mainWindow().findChildren('QToolBar'):
+    if ignore is not None and toolbar in ignore:
+      continue
+    toolbar.setVisible(visible)
+
+  # Prevent sequence browser toolbar showing up automatically
+  # when a sequence is loaded.
+  # (put in try block because Sequence Browser module is not always installed)
+  try:
+    import slicer
+    slicer.modules.sequences.autoShowToolBar = visible
+  except:
+    # Sequences module is not installed
+    pass
+
+def setMenuBarsVisible(visible, ignore=None):
+  """Show/hide all menu bars, except those listed in
+  ignore list."""
+  for menubar in mainWindow().findChildren('QMenuBar'):
+    if ignore is not None and menubar in ignore:
+      continue
+    menubar.setVisible(visible)
+
+def setPythonConsoleVisible(visible):
+  """Show/hide Python console."""
+  mainWindow().pythonConsole().parent().setVisible(visible)
+
+def setStatusBarVisible(visible):
+  """Show/hide status bar"""
+  mainWindow(verbose=False).statusBar().setVisible(visible)
+
+def setViewControllersVisible(visible):
+  """Show/hide view controller toolbar at the top of slice and 3D views"""
+  import slicer
+  lm = slicer.app.layoutManager()
+  for viewIndex in range(lm.threeDViewCount):
+    lm.threeDWidget(viewIndex).threeDController().setVisible(visible)
+  for sliceViewName in lm.sliceViewNames():
+    lm.sliceWidget(sliceViewName).sliceController().setVisible(visible)
+  for viewIndex in range(lm.tableViewCount):
+    lm.tableWidget(viewIndex).tableController().setVisible(visible)
+  for viewIndex in range(lm.plotViewCount):
+    lm.plotWidget(viewIndex).plotController().setVisible(visible)
+
+def forceRenderAllViews():
+  """Force rendering of all views"""
+  import slicer
+  lm = slicer.app.layoutManager()
+  for viewIndex in range(lm.threeDViewCount):
+    lm.threeDWidget(viewIndex).threeDView().forceRender()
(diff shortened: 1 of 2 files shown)


Base/Python/slicer/util.py
```diff
@@ -2334,90 +2422,13 @@ def pip_install(req):
...
-  modulePanel = findChild(mainWindow(), "ModulePanel")
-  modulePanel.helpAndAcknowledgmentVisible = visible
-
-def setDataProbeVisible(visible):
-  """Show/hide Data probe at the bottom of module panel."""
-  widget = findChild(mainWindow(), "DataProbeCollapsibleWidget")
-  widget.setVisible(visible)
+  :return string: Safe path
+  """
+  return u"\\\\?\\" + path.replace('/', '\\')
```
pieper, line 2434: Didn't you want a platform check here?  Other places we use `if os.name == 'nt':`

Also just for reference PATH_MAX on macos is 1024 and NAME_MAX is 255 (in
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/sys/syslimits.h)

Modules/Scripted/DICOMLib/DICOMUtils.py
```diff
@@ -116,7 +116,7 @@ def getDatabasePatientUIDByPatientID(patientID):
       continue
     firstFile = filePaths[0]
     # Get PatientID from first file
-    currentPatientID = slicer.dicomDatabase.fileValue(firstFile, "0010,0020")
+    currentPatientID = slicer.dicomDatabase.fileValue(slicer.util.longPath(firstFile), "0010,0020")
```
pieper, line 119: Seems like we might want to eventually move this `longPath` method to C++, perhaps in mrml or other fundamental place since this could pop up anywhere.  Maybe there's a way to wrap all IO libraries to use this longpath notation automatically?

Base/Python/slicer/util.py
cpinter: Yes, it will be needed. Since the application I use this in is Windows only, I submitted this as is for now, until we all agree that this will be necessary. @lassoan or @jcfr may have a different idea.

cpinter:
@lassoan @jcfr please chime in if you have a minute. This is a very simple fix, and if you agree with the general approach then I'll add the platform check and then we can continue discussing. Thanks!

lassoan:
Sorry, I haven't answered this because this issue is not simple at all. There are trivial things, such as:
- make it a no-op on linux/mac
- only add the UNC prefix if it has not been added already
- allow disabling it (just in case an OS or file system does not support it)

There are not-so-trivial issues, such as UNC path is only allowed for absolute paths, there is no substitution for /, ., and .., and handling of server shares (\\), and system devices (\\.\).

These are all fixable, probably by adopting David Gobbi's excellent [vtkDICOMFilePath.cxx](https://github.com/dgobbi/vtk-dicom/blob/master/Source/vtkDICOMFilePath.cxx) class.

However, the main problem is that this must to be handled at library level, most probably in CTK, because:
- We would really need to store relative paths in the DICOM database so that the database is relocatable (so that you can move it elsewhere; or that you can use it on a USB stick, which is mapped to a different drive letter on every computer). I'm afraid that this change would interfere with this.
- It is unlikely that slicer.dicomDatabase.fileValue is the only affected method. I had problems in the past importing anything when DICOM database directory was too long. As a workaround, we implemented CTK DICOM infrastructure in CTK so that folder structure is kept short (e.g., .../patientid/studyinstanceuid/seriesinstanceuid would make sense logically but would often exceed the maximum path length), but you can still quite easily run into path length issues.

Finally, you might not need to use UNC paths if you can control your environment, i.e., you know you only need to support NTFS file system and you can enable NTFS long path (https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#enable-long-paths-in-windows-10-version-1607-and-later). We have already added WindowsApplicationUseUtf8.manifest to CTK, the manifest to enabl long path support could be added similarly (`add_executable(${ARGN} ${CTK_CMAKE_DIR}/WindowsApplicationUseUtf8.manifest ${CTK_CMAKE_DIR}/WindowsApplicationLongPathAware.manifest`).

cpinter:
Thanks a lot @lassoan ! This is exactly why I kept the PR super simple (so some of your feedback I expected), because I wanted to know what else is involved. There is a chance that I can work on this in a month or two, let me check with the funding partner.

If we can identify a subset of this that makes sense to implement now (relatively simple but safe and points to the right direction), then I can do those now.

cpinter:
I made a round of improvements, please check it out.
In the meantime I'll look for other places to use this.

cpinter:
@lassoan @pieper I made some fixes that could be enough to let this be merged as a first step in the right direction.

What I haven't done yet is check where else in the code it could be used.

cpinter:
@pieper @lassoan Do you think this could be integrated? With my changes the basic issues are solved, and we can add its usage to other places as we find them.

cpinter:
Great, thanks! I will just amend the commit message not to include the WIP and force push.

cpinter:
@pieper I amended (only) the message. Could you please re-approve? Thank you!

cpinter:
Thanks a lot @pieper!

PR #4923, opened 2020-05-14
BUG: ScreenCapture- image series files inadvertently deleted 

hherhold:
When creating image series in ScreenCapture module, deleteTemporaryFiles() would always be called at the end of the try: block. This should not be done if "image series" is selected.


lassoan:
Thank you, good catch. In order to merge the commit, the commit message has to start with a specific prefix that indicates what kind of change this is (ENH:, BUG:, DOC:, ...).

Could you add `BUG:` prefix to the commit message? (you need to amend the commit and force-push to your master branch again). If you cannot figure out how to do these in git then let us know.

hherhold:
Hey @lassoan, thanks. Fixed, I think - let me know if it doesn't look ok.

PR #4929, opened 2020-05-19
BUG: Return GetCurvePoints in the correct coordinate system

Sunderlandkyl:
Due to changes in the input of the curve generator, curve points were being returned in the world coordinate system. Fixed by passing the input points in the local coordinate system.
This also ensures that the output curve transform is up-to-date if the parent transform node changes.

### DIFF
--- a/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsCurveNode.cxx
+++ b/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsCurveNode.cxx
@@ -72,12 +72,12 @@ vtkMRMLMarkupsCurveNode::vtkMRMLMarkupsCurveNode()
   this->TriangleFilter = vtkSmartPointer<vtkTriangleFilter>::New();
   this->TriangleFilter->SetInputConnection(this->CleanFilter->GetOutputPort());
 
-  this->SurfaceToWorldTransformer = vtkSmartPointer<vtkTransformPolyDataFilter>::New();
-  this->SurfaceToWorldTransformer->SetTransform(vtkNew<vtkGeneralTransform>());
-  this->SurfaceToWorldTransformer->SetInputConnection(this->TriangleFilter->GetOutputPort());
+  this->SurfaceToLocalTransformer = vtkSmartPointer<vtkTransformPolyDataFilter>::New();
+  this->SurfaceToLocalTransformer->SetTransform(vtkNew<vtkGeneralTransform>());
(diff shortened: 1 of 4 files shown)


Modules/Loadable/Markups/MRML/vtkMRMLMarkupsCurveNode.cxx
```diff
@@ -1273,22 +1274,14 @@ void vtkMRMLMarkupsCurveNode::OnSurfaceModelTransformChanged()
     }
 
   vtkSmartPointer<vtkGeneralTransform> surfaceToWorldTransform = vtkGeneralTransform::SafeDownCast(
-    this->SurfaceToWorldTransformer->GetTransform());
+    this->SurfaceToLocalTransformer->GetTransform());
   if (!surfaceToWorldTransform)
     {
     surfaceToWorldTransform = vtkSmartPointer<vtkGeneralTransform>::New();
-    this->SurfaceToWorldTransformer->SetTransform(surfaceToWorldTransform);
+    this->SurfaceToLocalTransformer->SetTransform(surfaceToWorldTransform);
```
lassoan, line 1281: We set surfaceToWorldTransform in a SurfaceToLocalTransformer. One of the names should be fixed.

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsCurveNode.h
```diff
@@ -210,7 +210,7 @@ class  VTK_SLICER_MARKUPS_MODULE_MRML_EXPORT vtkMRMLMarkupsCurveNode : public vt
 protected:
   vtkSmartPointer<vtkCleanPolyData> CleanFilter;
   vtkSmartPointer<vtkTriangleFilter> TriangleFilter;
-  vtkSmartPointer<vtkTransformPolyDataFilter> SurfaceToWorldTransformer;
+  vtkSmartPointer<vtkTransformPolyDataFilter> SurfaceToLocalTransformer;
```
lassoan, line 213: Can you define somewhere what this local coordinate system is? How it is related to world

Sunderlandkyl:
Changes made.

Markup line node remains straight, but curve node points are non-linearly transformed. This does not match the current behavior, but it does match the original behavior from before the bug was introduced (with the revamped curve generator and ShortestDistanceOnSurface curve type).

lassoan:
Thank you very much, this looks good!

Is there anything you wanted to test or change or is this ready to merge?

Sunderlandkyl:
Nope, nothing else, thanks Andras!

lassoan:
Thank you very much for the quick fix!

Issue #4925, opened 2020-05-16
Improvements to the Texts module


karamellpelle:
Here are some suggestions for the _Texts_ module

* Use monospaced font. In this way you can make tables etc. This will also let your notes stand out compared to the application font for a better user experience.
* It should be possible to just type in your notes directly without clicking the Edit button. And remove the Save and Cancel buttons. I didn’t see those buttons, so when I tried to Clone the text node in Data view, the text was not copied.
* The editor field should look editable. Now it's just a white rectangle.

lassoan:
The second and third bullet points are already implemented. You can enable direct editing in advanced section.

Edit/Save is needed if you want to edit text while you expect that other modules may change the node at the same time.
Using a monospaced font is a good idea. Pull request with the change is welcome.

lassoan:
Switched to a monospaced font.

PR #4931, opened 2020-05-21
ENH: Texts module with monospaced font

karamellpelle:
Monospaced font for `qMRMLTextEditWidget` as suggested in #4925. This is done by [setting the font family to _Courier_](https://forum.qt.io/topic/35999/solved-qplaintextedit-how-to-change-the-font-to-be-monospaced/5).

**Question:** `QTextEdit` has a property `acceptRichText` which defaults to _true_. Therefore, if you paste formatted text into our widget, the content will have formatting too. Should we set `acceptRichText` to _false_ to make clear that the data is plain text? (The function `void qMRMLTextWidget::updateMRMLFromWidget()` converts content to plain text).

### DIFF
--- a/Modules/Loadable/Texts/Widgets/Resources/UI/qMRMLTextWidget.ui
+++ b/Modules/Loadable/Texts/Widgets/Resources/UI/qMRMLTextWidget.ui
@@ -15,7 +15,13 @@
   </property>
   <layout class="QVBoxLayout" name="verticalLayout">
    <item>
-    <widget class="QTextEdit" name="TextEdit"/>
+    <widget class="QTextEdit" name="TextEdit">
+     <property name="font">
+      <font>
+       <family>Courier</family>
+      </font>
+     </property>
+    </widget>
    </item>
    <item>
     <layout class="QHBoxLayout" name="ButtonLayout">


jcfr:
Thanks @Sunderlandkyl for the review and thanks @karamellpelle for the contribution :pray:

lassoan:
> Question: QTextEdit has a property acceptRichText which defaults to true. Therefore, if you paste formatted text into our widget, the content will have formatting too. Should we set acceptRichText to false to make clear that the data is plain text? (The function void qMRMLTextWidget::updateMRMLFromWidget() converts content to plain text).

We could disable rich text, but if it works well then we could just as well support it. I've added comment with more details in the forum discussion: https://discourse.slicer.org/t/text-annotations/11558/18

karamellpelle:
So I should create a patch for the `acceptRichText` option too? That's also an easy fix in _Qt Designer_.

How do you recompile after editing? Are there for example some arguments/targets I can pass to `make` in order to speed up build time? It takes too long time to scan through a whole build every time. I'm just started learning Slicer development.

@lassoan: I saw your comment in the forum; I'll take a look at that later :)

lassoan:
You can build the inner project (Slicer-build/Slicer.sln). You can go even deeper but then you need to think about what needs to be rebuilt when you make a change.

PR #4934, opened 2020-05-26
ENH: Add absolute way to specify markups curve and line thickness

cpinter:
Curve diameter so far could not be set independently form markup control point diameter, but to use curves to show e.g. intravascular catheter with specific diameter, we need the ability to set the displayed curve diameter to an absolute value.

We also need to keep the ability to specify the size relative to the markups size, because for some other use cases that is the preferable option, so we have similar options for this as for glyph sizing:
- LineThickness (already existed, size relative to glyph)
- LineDiameter (new double value, absolute value in length unit, such as mm)
- CurveLineRadiusMode (enum: UseLineThickness and UseLineDiameter for now, but later we may add more options, such as vary radius according to scalar value associated with control points)

### DIFF
--- a/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsDisplayNode.cxx
+++ b/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsDisplayNode.cxx
@@ -88,7 +88,9 @@ vtkMRMLMarkupsDisplayNode::vtkMRMLMarkupsDisplayNode()
   // Set active component defaults for mouse (identified by empty string)
   this->ActiveComponents[GetDefaultContextName()] = ComponentInfo();
 
+  this->CurveLineSizeMode = vtkMRMLMarkupsDisplayNode::UseLineThickness;
   this->LineThickness = 0.2;
+  this->LineDiameter = 1.0;
 
   // Line color variables
   this->LineColorFadingStart = 1.;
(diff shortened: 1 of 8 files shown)


cpinter:
I noticed that in the 2D views the control point and line size is constant regardless the zoom factor and so cannot be used to see an actual catheter. I guess this should be changed as well. Any ideas what would be the best way?

lassoan:
> I noticed that in the 2D views the control point and line size is constant regardless the zoom factor and so cannot be used to see an actual catheter. 

You can choose between absolute and screen-relative glyph size in both 2D and 3D views, using UseGlyphScale display property.

cpinter:
I see. For some reason it seemed that it was screen-relative in 2D but absolute in 3D, but now I can see that it's screen-relative in both cases by default. OK that solves this issue, thanks!

cpinter:
@lassoan Is there any issue with this PR of can it be merged?

lassoan:
The logic looks good, I'm only thinking about how we should improve naming, because CurveLineRadiusMode -> UseLineDiameter is a bit confusing. Instead of CurveLineRadiusMode, what do you think about CurveLineSizeMode, CurveLineSizingMode, or something similar?

cpinter:
Yes I have been wondering about this as well. CurveLineSizeMode would be better.

cpinter:
I renamed the variable, please see the updated branch.

PR #4938, opened 2020-05-27
ENH: Added json file format for markups

lassoan:
Markups did not have a generic storage node that worked for all for types
of markups. This new storage nodes aims to fill this need. It is also meant
to be easily extensible in potential markup subclasses.

The new generic storage exports markups to JSON for ease of interface with
the web, python and other applications.

The existing Markups Fiducial reader is left as-is and is still the default
for fiducials.

The original pull request (https://github.com/Slicer/SlicerGitSVNArchive/pull/1017)
has been simplified: no intermediate translation map is created but node
is directly read from json document and streamed to file using json write functions.

Limitations:
- Measurements are not read/written to file
- Schema file has not been created

Co-authored-by: Johan Andruejol <johan.andruejol@kitware.com>


lassoan:
@jcfr @pieper @fedorov

This is ready for integration, but I'm OK to wait a bit to see if discussions about adopting DICOM-compliant JSON will proceed.

Currently, the file is just a direct mapping of MRML node attributes to JSON name:value pairs:

```json
{
    "@schema": "https://raw.githubusercontent.com/slicer/slicer/master/Modules/Loadable/Markups/Resources/markups-schema.json#",
    "markupsType": "Fiducial",
    "coordinateSystem": "RAS",
    "locked": false,
    "labelFormat": "%N-%d",
    "controlPoints": [
        {
            "id": "2",
            "label": "Testing label",
            "description": "description with spaces",
            "associatedNodeID": "testingAssociatedID",
            "position": [-9.9, 1.1, 0.87],
            "orientation": [1.0, 0.0, 0.0, 0.0, 0.9999939076577904, -0.003490651415223732, 0.0, 0.003490651415223732, 0.9999939076577904],
            "selected": false,
            "locked": true,
            "visibility": false,
            "positionStatus": "defined"
        },
        {
            "id": "3",
            "label": "MarkupsFiducial-2",
            "description": "",
            "associatedNodeID": "",
            "position": [0.0, 0.0, 0.0],
            "orientation": [1.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 1.0],
            "selected": true,
            "locked": false,
            "visibility": true,
            "positionStatus": "undefined"
        },
        {
            "id": "4",
            "label": "",
            "description": "",
            "associatedNodeID": "",
            "position": [0.0, 0.0, 0.0],
            "orientation": [1.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 1.0],
            "selected": true,
            "locked": false,
            "visibility": true,
            "positionStatus": "undefined"
        },
        {
            "id": "5",
            "label": "Label, commas, two",
            "description": "Description one comma \"and two quotes\", for more testing",
            "associatedNodeID": "",
            "position": [0.0, 0.0, 0.0],
            "orientation": [1.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 1.0],
            "selected": true,
            "locked": false,
            "visibility": true,
            "positionStatus": "undefined"
        },
        {
            "id": "6",
            "label": "Label with end quotes \"around the last phrase\"",
            "description": "\"Description fully quoted\"",
            "associatedNodeID": "",
            "position": [0.0, 0.0, 0.0],
            "orientation": [1.0, 0.0, 0.0, 0.0, 1.0, 0.0, 0.0, 0.0, 1.0],
            "selected": true,
            "locked": false,
            "visibility": true,
            "positionStatus": "undefined"
        }
    ],
    "display": {
        "visibility": true,
        "opacity": 1.0,
        "color": [0.4, 1.0, 1.0],
        "selectedColor": [1.0, 0.5, 0.5],
        "propertiesLabelVisibility": true,
        "pointLabelsVisibility": false,
        "textScale": 3.0,
        "glyphType": "Sphere3D",
        "glyphScale": 1.0,
        "glyphSize": 5.0,
        "useGlyphScale": true,
        "sliceProjection": false,
        "sliceProjectionUseFiducialColor": true,
        "sliceProjectionOutlinedBehindSlicePlane": false,
        "sliceProjectionColor": [1.0, 1.0, 1.0],
        "sliceProjectionOpacity": 0.6,
        "lineThickness": 0.2,
        "lineColorFadingStart": 1.0,
        "lineColorFadingEnd": 10.0,
        "lineColorFadingSaturation": 1.0,
        "lineColorFadingHueOffset": 0.0,
        "handlesInteractive": false,
        "snapMode": "toVisibleSurface"
    }
}
```

pieper:
LGTM.  Thanks for considering the dicom option.

lassoan:
@pieper Do you think there is a chance to make DICOM SR simple enough so that it would not be a major burden for users? Or we should just go with whatever the simplest for Slicer and use DICOM SR as an export/import format only?

pieper:
Good question - I think @dclunie is the only one who has tried to implement this so far, and only in abstracted examples, not in the context of a real application.  I liked your suggestions in the last email so let's see if he responds to that.

Of course there's going to be a lot of Slicer-specific data in these files, or at least concepts like orientation and lock status that aren't in the standard.  We can use our on coded concepts for those.  Plus we want to include display info, which David doesn't want to see in SR (but he can't prevent us from putting it in with our own codes).

We'll also have a lot of cases where we don't have the so-called "composite context" (patient/study ids) that are needed to create valid dicom.

There's also the tricky issue of the instance UID.  If you give someone an example file and they make their own by hand, it seems likely to me that they would just copy the UID from the example, which would be the exact opposite of what you would want.

So I guess the answer has to be that we just need to make dicom an import/export option only and use something simple as the informal research format.  I've just convinced myself that you should just merge this as-is.  Then we use SR only when the whole workflow is in dicom, which we should still try to encourage in general.

@fedorov any thoughts?

fedorov:
TL;DR: 👍 I think this PR is a great start and is way better than what Slicer does right now saving markups.

We also discussed the DICOM option a bit in this thread: https://discourse.slicer.org/t/markupsline-fcsv-loads-as-markupsfiducials/11545 (by the way, I did forward that thread to David Clunie when the discussion started), and my main issue was I could not see how DICOM could be used to support the variety of use cases in Slicer, which often deals with data that is not and cannot become DICOM. Because of this I think it is appropriate to use non-DICOM representation, but keep around attributes that can be used to create a DICOM representation (such as Series/InstanceUID of the image being annotated that could be used to populate composite context). I don't know if Andras had something in mind for this, but even if DICOM JSON SR can be simple enough to meet the requirements of Andras, I cannot see how it can be used for all of Slicer needs. It is also important to keep in mind that "It is not expected that the JSON representation will be used as the persistent form, but rather that the existing DICOM binary object storage infrastructure will be used" (quote from [DICOM Sup 219](ftp://medical.nema.org/medical/dicom/supps/Frozen/sup219_fz_JSONSR.pdf)), so JSON DICOM SR is not a replacement of DICOM SR.

On the topic of "is there a chance to make DICOM SR simple enough so that it would not be a major burden for users", I think that simplification should come from the implementation. If the DICOM SR use is restricted to a specific template, and Slicer (or ideally a standalone library used by Slicer) provides API to parameterize creation of those complex representation, user may not care much about what representation is used to serialize the content (like we do with [dcmqi](https://github.com/QIICR/dcmqi) and [dcmqi-specific JSON files](https://github.com/QIICR/dcmqi/tree/master/doc/examples) parameterizing the conversion). Outside of Slicer, other implementations that support specific templates can hide those complexities too while reading content generated by Slicer, such as [highdicom](https://github.com/MGHComputationalPathology/highdicom). In the end, you don't expect the user to create those JSON files by hand, right? The example above is already quite complex for that.

I do think it is a very valuable exercise to explore mapping appropriate use cases of markups into DICOM representation, but it doesn't need to happen in this PR. I don't think there is any better way other than DICOM to put the outputs produced by Slicer into a bigger context of medical imaging data.

Sorry for the long response, and maybe not completely to the point. I think this kind of topics are best discussed in a conversation, hope we can get to that sometime!

fedorov:
Forgot to mention, I did think that the simplifications suggested by Andras to David C make sense, but I don't have the depth to anticipate whether this can break anything. Would be great to hear from David C about those. But, again, I do not think any simplification of JSON DICOM SR would address the fundamental concerns of using it as THE format for Slicer markups.

lassoan:
@pieper @fedorov Thanks for your valuable insights.

There are a number of software that use DICOM as a basis of their data storage. I thought that this could have been a good experiment to see how it could work for markups and if it turns out to be successful then slowly expand this to other data types (images, segmentations, etc.).

All the issues you've raised are valid and the solutions may indeed not be trivial, may even require some extra infrastructure development, but possible to solve.

Private data could be stored in private fields, as it is done in many other software. This would have been a follow-up request to David (to be able to incorporate private fields with a simple syntax, too).

We could use internal UIDs (such as Slicer DICOM database UIDs, MRML node ID, filename, or combination of these), or completely omit them and generate them when exporting data into the Slicer DICOM database.

Let's see what David says, but I'm not very hopeful either.

jcfr:
Should the Markups documentation be updated as well to describe the format `.mrk.json` ?  See https://slicer.readthedocs.io/en/latest/user_guide/modules/markups.html

In the project where we originally introduced support for saving json as annotation, we were also saving other information (see [here](https://github.com/BICCN/cell-locator/blob/master/Modules/Loadable/Splines/MRML/vtkMRMLMarkupsSplinesStorageNode.cxx#L71) and [here](https://github.com/BICCN/cell-locator/blob/604de038819daf91751ebb4a54f1cbfd7e06e0b9/Modules/Loadable/Splines/MRML/vtkMRMLMarkupsSplinesStorageNode.cxx#L244-L280)), when I will update that project to use an updated Slicer version ... I am wondering what will be the best past forward.

May be it will be a matter of creating a custom Json storage node deriving from  `vtkMRMLMarkupsJsonStorageNode` and then adding a custom file reader for `.celllocator.json` (or similar)

lassoan:
> Should the Markups documentation be updated as well to describe the format .mrk.json ?

Good point, we should get used to doing this.

> May be it will be a matter of creating a custom Json storage node deriving from vtkMRMLMarkupsJsonStorageNode and then adding a custom file reader for .celllocator.json (or similar)

Yes. I've split the reader/writer methods into smaller methods to make custom implementations easier, but probably this may need to be further refined when there is a concrete implementation that uses this.

Instead of .celllocator.json, you may consider using something like .cell.mkp.json to indicate that it is a markup file (useful if you want to load the file into a plain Slicer that does not have celllocator module installed).

fedorov:
@lassoan it might be a good idea to include schema version information (as a version attribute, or by versioning the schema URL referenced). Somehow I think this format will evolve over time ;-)

I searched, and it seems there is no recommended way in JSON Schema to do schema versioning. Some suggestions are discussed here: https://gist.github.com/mattyod/3608613 and https://github.com/json-schema-org/understanding-json-schema/issues/116.

jcfr:
> I've split the reader/writer methods into smaller methods to make custom implementations easier

Perfect.

>  .cell.mkp.json to indicate that it is a markup file

Makes sense

> it might be a good idea to include schema version information

Agreed.

jcfr:
I am addressing related compiler warnings on my workstation along side with a compiler specific issue:

```
In file included from /home/jcfr/Projects/Slicer/Libs/MRML/Core/vtkMRMLCoreTestingMacros.h:409:0,
                 from /home/jcfr/Projects/Slicer/Modules/Loadable/Markups/Testing/Cxx/vtkMRMLMarkupsStorageNodeTest2.cxx:19:
/home/jcfr/Projects/Slicer/Modules/Loadable/Markups/Testing/Cxx/vtkMRMLMarkupsStorageNodeTest2.cxx: In function ‘int vtkMRMLMarkupsStorageNodeTest2(int, char**)’:
/home/jcfr/Projects/Slicer/Modules/Loadable/Markups/Testing/Cxx/vtkMRMLMarkupsStorageNodeTest2.cxx:257:75: error: invalid initialization of non-const reference of type ‘std::__cxx11::string& {aka std::__cxx11::basic_string<char>&}’ from an rvalue of type ‘std::__cxx11::basic_string<char>’
     vtkSmartPointer<vtkMRMLMarkupsFiducialStorageNode>::New(), tempFolder + "/vtkMRMLMarkupsStorageNodeTest2-fiducial-temp.fcsv"));
                                                                           ^
[...]
```

jcfr:
It turns out in cell locator, we are currently saving multiple annotation per json file. After updating to the latest Slicer, I will then probably create the file with extension `.cell.mrks.json`

I am also thinking to also add support for format `.mrks.json` in  Slicer where multiple annotations are saved in a single json file or we could generalize the current json format to always be a list of annotations.


```json
{
    "@schema": "https://raw.githubusercontent.com/slicer/slicer/master/Modules/Loadable/Markups/Resources/markups-schema.json#",
    "markupsType": "ClosedCurve",
    "coordinateSystem": "LPS",
    "locked": false,
    "labelFormat": "%N-%d",
    "controlPoints": [
        {
            "id": "1",
            "label": "",
            "description": "",
            "associatedNodeID": "vtkMRMLScalarVolumeNode1",
            "position": [76.09203482376901, 53.664027452970817, -10.214302062988282],
            "orientation": [-1.0, -0.0, -0.0, -0.0, -1.0, -0.0, 0.0, 0.0, 1.0],
            "selected": true,
            "locked": false,
            "visibility": true,
            "positionStatus": "defined"
        },
        [...]
    ],
    "display": {
        "visibility": true,
        "opacity": 1.0,
        "color": [0.4, 1.0, 1.0],
        [...]
        "handlesInteractive": false,
        "snapMode": "toVisibleSurface"
    }
}
```
 would be changed into

```json
{
    "@schema": "https://raw.githubusercontent.com/slicer/slicer/master/Modules/Loadable/Markups/Resources/markups-schema.json#",
    "annotations": [
        {
            "markupsType": "ClosedCurve",
            "coordinateSystem": "LPS",
            "locked": false,
            "labelFormat": "%N-%d",
            "controlPoints": [
                {
                    "id": "1",
                    "label": "",
                    "description": "",
                    "associatedNodeID": "vtkMRMLScalarVolumeNode1",
                    "position": [76.09203482376901, 53.664027452970817, -10.214302062988282],
                    "orientation": [-1.0, -0.0, -0.0, -0.0, -1.0, -0.0, 0.0, 0.0, 1.0],
                    "selected": true,
                    "locked": false,
                    "visibility": true,
                    "positionStatus": "defined"
                },
                [...]
            ],
            "display": {
                "visibility": true,
                "opacity": 1.0,
                "color": [0.4, 1.0, 1.0],
                [...]
                "handlesInteractive": false,
                "snapMode": "toVisibleSurface"
            }
        }
    ]
}
```


Currently, saving such file would not be possible through the regular save dialog.

image

lassoan:
I agree that exporting/import of multiple markups into a single file will be a useful feature.
GUI could be right-click on a folder in the Data module (subject hierarchy plugin would add a new option of exporting markups to file) and we could also add an export section to Markups module.

Import could be handled by the current reader.

Since a single reader can manage single-markup and multi-markup files, I think we can keep using the same extension.

If we export multiple nodes into a single file then we also need to save the node name of each markups node in the JSON file.

lassoan:
Thank you for all the useful feedback. I've added a (versioned) schema file, which is a quite comprehensive formal description of the file format. All tests pass, I'll merge it now.

PR #4940, opened 2020-05-28
COMP: Update minimum required CMAKE_OSX_DEPLOYMENT_TARGET to 10.13

jamesobutler:
This is for work https://github.com/Slicer/DashboardScripts/pull/27 which updates the factory macOS build to use Qt 5.15.0.

Deployment target >= 10.13 is the one officially supported by Qt 5.15. See [Qt 5.15 supported platforms](https://doc-snapshots.qt.io/qt5-5.15/supported-platforms.html).

This will support the following macOS versions:
- 10.15: Catalina (2019)
- 10.14: Mojave (2018)
- 10.13: High Sierra (2017)

The last forum discussion about updating minimum osx deployment target can be found [here](https://discourse.slicer.org/t/poll-which-version-of-macos-do-you-use-to-run-slicer/6055).

---------------------------
macOS 10.13 High Sierra is compatible with the following hardware ([source](https://support.apple.com/en-us/HT208969)):
- MacBook introduced in late 2009 or later
- MacBook Air introduced in late 2010 or later
- MacBook Pro introduced in mid 2010 or later
- Mac mini introduced in mid 2010 or later
- iMac introduced in late 2009 or later
- Mac Pro introduced in mid 2010 or later

According to https://gs.statcounter.com/os-version-market-share/macos/desktop/worldwide as of April 2020 this would mean there would be support for 84.23% of Mac devices worldwide.

### DIFF
--- a/CMake/SlicerDashboardScript.TEMPLATE.cmake
+++ b/CMake/SlicerDashboardScript.TEMPLATE.cmake
@@ -34,7 +34,7 @@ dashboard_set(WITH_DOCUMENTATION  FALSE)
 dashboard_set(Slicer_BUILD_CLI    ON)
 dashboard_set(Slicer_USE_PYTHONQT ON)
 
-dashboard_set(QT_VERSION          "5.10.0")
+dashboard_set(QT_VERSION          "5.15.0")
 dashboard_set(Qt5_DIR             "${DASHBOARDS_DIR}/Support/Qt${QT_VERSION}/${QT_VERSION}/gcc_64/lib/cmake/Qt5")
 
 #   Source directory : <DASHBOARDS_DIR>/<Slicer_DASHBOARD_SUBDIR>/<Slicer_DIRECTORY_BASENAME>-<Slicer_DIRECTORY_IDENTIFIER>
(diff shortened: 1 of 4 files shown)


jamesobutler:
I've updated the getting started doc to now recommend macOS Catalina for Slicer and also include a specification about minimum requirement for macOS version.

jcfr:
For reference, I posted a note indicating that will update the operating system on the macOS factory tomorrow morning. See https://discourse.slicer.org/t/poll-which-version-of-macos-do-you-use-to-run-slicer/6055/6?u=jcfr

jcfr:
It turns the system can't simply be updated, Kitware sysadmins will proceed to a clean installation in the next few days. See https://discourse.slicer.org/t/poll-which-version-of-macos-do-you-use-to-run-slicer/6055/7?u=jcfr

jamesobutler:
@jcfr is the hardware listed [here](https://www.slicer.org/wiki/Documentation/Nightly/Developers/Factory#Hardware) still up-to-date? Because it looks like the ATI Radeon HD 5770 GPU isn’t Metal compatible which is required as of macOS 10.14

jcfr:
> is the hardware listed here still up-to-date?

I just checked and yes, this is correct.

image


> Because it looks like the ATI Radeon HD 5770 GPU isn’t Metal compatible which is required as of macOS 10.14

Thanks for pointing this out,. It looks like both macOs machines have the same graphic card, I will mention that to our sysadmin and discuss a plan to move forward with the macOS build machine.

jamesobutler:
Take a look at https://support.apple.com/en-us/HT208898 which says 

> Don't upgrade your Mac Pro (Mid 2010) or Mac Pro (Mid 2012) to macOS Mojave directly from macOS versions prior to 10.13.6.

So you could on your own make this transition to High Sierra if the likely goal is to stick with this same build machine.

jcfr:
Thanks for pointing this out. I believe this is the approach our Sysadmins are following.

PR #4943, opened 2020-05-28
BUG: unset cursor on scene close in slice views

vovythevov:
This fixes the following issue:
 - Start to edit a segmentation by painting in a slice
 - Close the scene
 -> The cursor in the Slice view is still showing the Paint cursor even though you're not painting anymore.

Note: This fix should probably be propagated to the 3D view and any other view that may have cursor change (maybe Chart view ?).

### DIFF
--- a/Libs/MRML/Widgets/qMRMLSliceWidget.cxx
+++ b/Libs/MRML/Widgets/qMRMLSliceWidget.cxx
@@ -94,6 +94,12 @@ void qMRMLSliceWidgetPrivate::endProcessing()
 }
 
 // --------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


lassoan:
Thanks for working on this. It is indeed a bug, but it should be fixed at a bit higher level. I'll try to find a better place and get back to you.

lassoan:
Thanks again for taking the time to fix this, I would close this pull request in favor of this more comprehensive fix: https://github.com/Slicer/Slicer/commit/f49930e4dc9e78d8771ec9f65df8f25bc571958b

vovythevov:
Awesome ! Thanks for the fix !

PR #4948, opened 2020-05-29
Use correct Qt case in build names and comments

jamesobutler:
This updates "QT" to the correct case of "Qt". I first noticed this in the nightly package build names on the dashboard. 

This doesn't update the case for any variables which might be specifying it as "QT" in the context of an all CAPS variable name. It also doesn't update file names such as "qSlicerBaseQTCoreExport" which should really by "qSlicerBaseQtCoreExport".


Base/QTCLI/qSlicerCLIExecutableModuleFactory.h
```diff
@@ -21,7 +21,7 @@
 #ifndef __qSlicerCLIExecutableModuleFactory_h
 #define __qSlicerCLIExecutableModuleFactory_h
 
-// SlicerQT includes
+// SlicerQt includes
```
jcfr, line 24: Since we are modifying the files, I suggest to use `// Slicer includes`

Base/QTCore/Testing/Cxx/CMakeLists.txt
```diff
@@ -120,7 +120,7 @@ if(BUILD_TESTING)
   # Add Tests
   #
 
-  # Remark: qSlicerModuleFactoryManager class is tested within Applications/SlicerQT/Testing
+  # Remark: qSlicerModuleFactoryManager class is tested within Applications/SlicerQt/Testing
```
jcfr, line 123: `Applications/SlicerQt/Testing` -> `Applications/SlicerApp/Testing`

Base/QTCLI/qSlicerCLIExecutableModuleFactory.h
jamesobutler: Specifically changing `// SlicerQt includes` to `// Slicer includes` across all files used?

jcfr: Indeed, since the files live in the `Slicer/Base/*` directories, I think simply using `Slicer` makes sense.

On the other hand, files living in `Libs/*` are referenced using `MRML includes, ...`

PR #4953, opened 2020-06-01
COMP: Update FindVcvars with newer compiler versions

jamesobutler:
This fixes a configure failure noticed for the SlicerRadiomics extension (see [here](http://slicer.cdash.org/viewConfigure.php?buildid=1927747)) due to the recent upgrade of the compiler used on the factory Windows build machine.

```
-- Setting Vcvars_MSVC_VERSION to '1924' as MSVC_VERSION was `1924`
CMake Error at D:/D/P/Slicer-0/CMake/FindVcvars.cmake:161 (message):
  failed to convert msvc_version [1924].  It is not a known version number.
```

### DIFF
--- a/CMake/FindVcvars.cmake
+++ b/CMake/FindVcvars.cmake
@@ -116,7 +116,9 @@ cmake_minimum_required(VERSION 3.13.4)
 # Global variables used only in this script (unset at the end)
 set(_Vcvars_MSVC_ARCH_REGEX "^(32|64)$")
 set(_Vcvars_MSVC_VERSION_REGEX "^[0-9][0-9][0-9][0-9]$")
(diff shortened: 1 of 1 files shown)


jcfr:
Thanks for the fix 🙏 

Could you submit a similar change to https://github.com/scikit-build/cmake-FindVcvars and then reference the change in this PR ? 

See https://github.com/scikit-build/cmake-FindVcvars#integrating-the-module-in-your-project

jcfr:
To streamline the process, you should now have write access to https://github.com/scikit-build/cmake-FindVcvars

jamesobutler:
I've included the note about the commit that was added to the cmake-FindVcvars repository.  This PR is currently a copy implementation for integrating.  The FindVcvars.cmake file here in Slicer doesn't include https://github.com/scikit-build/cmake-FindVcvars/commit/ae24f7cf4d192e6ec9cfbeded343a0abcbed5dca which those changes could also be copied over here if you want.

PR #4956, opened 2020-06-02
ENH: Set default display precision of length unit to 4

lassoan:
By default, length values over 100mm were displayed without decimal digits by default (precision=3), which may be insufficient for some medical applications, since measurement of such lengths is no uncommon and image resolution is usually in the magnitude of 0.1mm.

Example: 104.123mm distance was displayed as "104.mm" before this change and it is displayed as "104.2mm" after this change.


lassoan:
Thanks for the review.

PR #4971, opened 2020-06-10
BUG: Update qMRMLTreeView to fix visibility toggle

jcfr:
This commit fixes a regression introduced in 2f932d8 / r28510 (ENH: Remove
model hierarchy nodes) and ensures visibility of both display and displayable
nodes can be toggled using a qMRMLTreeView associated with a "Displayable"
scene model (qMRMLSceneDisplayableModel)

### DIFF
--- a/Libs/MRML/Widgets/qMRMLTreeView.cxx
+++ b/Libs/MRML/Widgets/qMRMLTreeView.cxx
@@ -827,8 +827,13 @@ bool qMRMLTreeView::clickDecoration(const QModelIndex& index)
 void qMRMLTreeView::toggleVisibility(const QModelIndex& index)
 {
   vtkMRMLNode* node = this->sortFilterProxyModel()->mrmlNodeFromIndex(index);
-  vtkMRMLDisplayNode* displayNode =
-    vtkMRMLDisplayNode::SafeDownCast(node);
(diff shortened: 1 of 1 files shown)


lassoan:
The change looks good, if it's needed. 

Is there any reason to use displayable hierarchy instead of subject hierarchy? Could we just delete displayable scene model altogether to reduce maintenance and testing burden? Scene models are extremely complicated (and still have a number of small issues) so we should do everything to keep things as simple and small as possible.

jcfr:
Thanks for the review. Since the change was easy enough to implement, I went with it.

> Could we just delete displayable scene model altogether to reduce maintenance and testing burden?

Ultimately, we should consider this.

> Is there any reason to use displayable hierarchy instead of subject hierarchy?

This is in context of the [Q3DCExtension](https://github.com/DCBIA-OrthoLab/Q3DCExtension), [AnglePlanes-Extension](https://github.com/DCBIA-OrthoLab/AnglePlanes-Extension) and [EasyClip-Extension](https://github.com/DCBIA-OrthoLab/EasyClip-Extension).

Waiting we discuss with the PI to better understand the use cases and evaluate what need to be changed to use (and improve) the new markups annotation, I am focusing on other issues .... and this was easier (at least in the short term)

lassoan:
OK!

PR #4972, opened 2020-06-11
Fix vtkMRMLMarkupsJsonStorageNode warnings

jcfr:


Modules/Loadable/Markups/MRML/vtkMRMLMarkupsJsonStorageNode.cxx
```diff
@@ -728,7 +728,7 @@ vtkMRMLMarkupsNode* vtkMRMLMarkupsJsonStorageNode::AddNewMarkupsNodeFromFile(con
     displayNode = vtkMRMLMarkupsDisplayNode::SafeDownCast(markupsNode->GetDisplayNode());
     if (displayNode && markup.HasMember("display"))
       {
-      success == success && this->Internal->UpdateMarkupsDisplayNodeFromJsonValue(displayNode, markup);
+      success = success && this->Internal->UpdateMarkupsDisplayNodeFromJsonValue(displayNode, markup);
```
jcfr, line 731: @lassoan  Does this change seems reasonable ?

lassoan: Yes, this is correct. Good catch!

PR #4978, opened 2020-06-12
COMP: Remove Slicer_ITK_VERSION_MAJOR configure option

### LINKED ISSUE #4866: Remove visibility of "Slicer_ITK_VERSION_MAJOR"
Developer burden can be reduced by removing false choice options.  There is really no option except for ITK v5.

Remove visibility, and perhaps complete existence of the Slicer_ITK_VERSION_MAJOR variable.

Leave it at the cost of developer burden for needing to consider it's values.

## Additional context

Attempting to reduce developer burden.

jamesobutler:
This closes #4866.

There is really no option except for ITK v5.  This can be brought back sometime in the future if there is ever a need to support multiple versions during a ITK v5/v6 transition

Also included minor commit to update version of ITK to find when using system ITK as part of build process.


jcfr:
I initially kept the change anticipating ITK 6. But since you submit this change, looks good to me :+1:

Issue #4803, opened 2020-04-03
Improve colors in "Dark mode"


jcfr:
Color of web links in Dark mode has too much contrast.

## Steps to reproduce

Slicer Slicer. Then, observe web links in welcome module or in the help tab of any modules.

For example, here is the welcome module:

![image](https://user-images.githubusercontent.com/219043/78324388-910c1180-7542-11ea-9707-42d1f3a0958b.png)

## Expected behavior

Web links should be represented with a lighter color like the one used on the "stackoverflow" website in dark mode.

Normal: #3ca4ff
Hover: #378ad3

![image](https://user-images.githubusercontent.com/219043/78324477-cd3f7200-7542-11ea-8d7f-e5b4b73b6b62.png)

pieper:
On a related note, the default background of the python console in dark mode needs to be fixed too.

pieper:
@rkikinis also reports:  dark mode: black on dark grey is difficult to read

![image](https://user-images.githubusercontent.com/126077/78926413-0c9e1f00-7a6b-11ea-84d1-629abb12a882.png)

pieper:
Also in the dicom metadata search bar the text is not easy to read:

![image](https://user-images.githubusercontent.com/126077/79462366-62bc1680-7fc5-11ea-8358-431a36d265f3.png)

When the text is selected you can see it:

![image](https://user-images.githubusercontent.com/126077/79462404-6fd90580-7fc5-11ea-8492-40aefadebc54.png)

jamesobutler:
- [ ] Update the base color palette to fix issues such as link text color being difficult to read and the disabled text being dark and unreadable as seen in the "Models" module in images above.
- [ ] Python Console. Update colors when in Dark Slicer to something like ![image](https://user-images.githubusercontent.com/15837524/84601586-49add580-ae4f-11ea-8dae-7a3fd58882e9.png)
- [ ] DICOM Issues: The dicom metadata search box is updating from the dark base color to white again because of these lines which reset the palette of the search box to white when the filter warning is not shown.  Instead it should be restoring the palette used by parent widgets.  The text being white is correct for dark mode as it is on a dark color.
https://github.com/commontk/CTK/blob/17373933b81472450f85cc250cea1380610e09b3/Libs/DICOM/Widgets/ctkDICOMTableManager.cpp#L172-L184

lassoan:
What would you all think about hardcoding Python console colors into the style class (the same way as we do it for all other colors in the application)?

[...]

PR #4993, opened 2020-06-19
Improve colors in dark mode

jamesobutler:
This tweak makes the colors a little bit darker which matches more closely the Dark Mode that macOS does automatically for the regular Slicer style.  See https://discourse.slicer.org/t/automatic-dark-mode-on-mac-in-recent-slicer-preview-releases/12077/6. I've maintained the current highlight color(as used for StackOverflow dark theme) because the blue in macOS is a very macOS specific selection color.  Part of #4803.

| Current | This PR | macOS |
|----------|---------|---------|
|slicer-dark-mode-updated|slicer-dark-mode-updated-again3|macos-dark-mode|
|slicer-dark-mode-updated2|slicer-dark-mode-updated2-again3|macos-dark-mode-3|
|slicer-dark-mode-updated3|slicer-dark-mode-updated3-again3|macos-dark-mode-2|


Base/QTGUI/qSlicerDarkStyle.cxx
```diff
@@ -41,31 +41,31 @@ QPalette qSlicerDarkStyle::standardPalette()const
...
-  palette.setColor(QPalette::Window, "#353535");
+  palette.setColor(QPalette::Window, "#323232");
   palette.setColor(QPalette::WindowText, Qt::white);
-  palette.setColor(QPalette::Disabled, QPalette::WindowText, "#7f7f7f");
-  palette.setColor(QPalette::Base, "#2d2d2d");
-  palette.setColor(QPalette::Disabled, QPalette::Base, "#323234");
-  palette.setColor(QPalette::AlternateBase, "#424242");
+  palette.setColor(QPalette::Disabled, QPalette::WindowText, "#6d6d6d");
+  palette.setColor(QPalette::Base, "#1e1e1e");
+  palette.setColor(QPalette::Disabled, QPalette::Base, "#2d2d2d");
```
jamesobutler, line 48: It appear in the macOS dark mode control, the disabled base color is the same as the base color if we want to match this state as well. This can be seen in the current screenshots of the Models module.

Base/QTGUI/qSlicerDarkStyle.cxx
```diff
@@ -41,31 +41,31 @@ QPalette qSlicerDarkStyle::standardPalette()const
...
-  palette.setColor(QPalette::Disabled, QPalette::ButtonText, "#7f7f7f");
+  palette.setColor(QPalette::Disabled, QPalette::ButtonText, "#6d6d6d");
   palette.setColor(QPalette::BrightText, Qt::red);
   // Color roles used mostly for 3D bevel and shadow effects.
   palette.setColor(QPalette::Light, "#828284");  // Lighter than Button color.
-  palette.setColor(QPalette::Midlight, "#5c5c5d");  // Between Button and Light.
+  palette.setColor(QPalette::Midlight, "#5a5a5b");  // Between Button and Light.
   palette.setColor(QPalette::Dark, "#232323");  // Darker than Button.
-  palette.setColor(QPalette::Mid, "#2c2c2c");  // Between Button and Dark.
+  palette.setColor(QPalette::Mid, "2b2b2b");  // Between Button and Dark.
```
jcfr, line 62: `2b2b2b` -> `#2b2b2b` ?

jamesobutler: Indeed that is a typo. I will fix that too

Base/QTGUI/qSlicerDarkStyle.cxx
jcfr: Do I understand correctly saying that:
* on Linux and Windows, we will be using our custom qSlicerDaskStyle
* on macOS, we will *not* have an explicit dark mode available and we will let the operating system taking care of it ?

And finally, when you asked "It appear in the macOS dark mode control, the disabled base color is the same as the base color if we want to match this state as well", it means that we will update our own `qSlicerDarkStyle` palette to that Linux and Windows dark theme match macOS system one ?

jamesobutler:
With the update of the disabled base to be like that on macOS
| Previous | Latest Update | macOS|
|-----------|----------------|---------|
|slicer-dark-mode-updated2-again3|slicer-dark-mode-updated2-again4|macos-dark-mode-3|

jcfr:
> re: macOS

the module name with  black color seems off:

image

Do we have control over this ?

Base/QTGUI/qSlicerDarkStyle.cxx
jamesobutler: This PR does not remove the Style selection for macOS, but that is something we should do as the latest Qt version we are using now supports letting macOS control the dark mode based on if the OS is using light or dark.  This PR just adjusts the qSlicerDarkStyle so that Linux and Windows can have a similar experience to what users on macOS will have automatically.

jamesobutler:
That black text issue is likely because the macOS version is trying to apply some of the palette of the qSlicerStyle while the OS is changing some things automatically. I don't have a mac on hand to test this at the moment. It is not an issue on Windows and likely not on Linux as well.

jamesobutler:
Based on this exercise that I'm doing of making Windows and Linux dark versions look like the automated dark mode done by macOS I really should only do this after making macOS not apply these style files where there could be a confusion of what is done automatically by the OS and was it being applied by the style (making the text in the combobox incorrectly dark color).

Base/QTGUI/qSlicerStyle.cxx
```diff
@@ -193,15 +194,16 @@ QRect qSlicerStyle::subControlRect(ComplexControl control, const QStyleOptionCom
...
-  palette.setColor(QPalette::Shadow, "#5a5a5a");
-  palette.setColor(QPalette::AlternateBase, QColor("#e4e4fe"));
+  // TODO: Use standardPalette as defined in qSlicerLightStyle
+
+  #ifdef Q_OS_WIN
+    // Qt on macOS already dynamically picks light/dark theme based on the OS setting
+    QSettings settings("HKEY_CURRENT_USER\\Software\\Microsoft\\Windows\\CurrentVersion\\Themes\\Personalize", QSettings::NativeFormat);
+    if (settings.value("AppsUseLightTheme") == 0)
+      {
+      // TODO: Use standardPalette as defined in qSlicerDarkStyle
```
jamesobutler, line 204: @lassoan Can you provide guidance, or commit the change yourself, to use the standardPalette() option from qSlicerLightStyle and qSlicerDarkStyle here in qSlicerStyle? I'm trying not to duplicate the individual palette set options from those classes here.

lassoan: Since qSlicerDarkStyle depends on qSlicerStyle, the dark palette has to be either in qSlicerStyle or in a third class that both classes can access. Since this situation is temporary (the palette will move into qSlicerDarkStyle when Qt fully supports Windows and Linux dark mode), I think it is simpler not to add the palette to a third class, but create a `QPalette qSlicerStyle::standardDarkPalette()const` method and use that in both qSlicerStyle and qSlicerDarkStyle.

jamesobutler: Ok, I will try that!

Base/QTGUI/qSlicerStyle.cxx
```diff
@@ -191,17 +192,73 @@ QRect qSlicerStyle::subControlRect(ComplexControl control, const QStyleOptionCom
...
+  QPalette palette =  standardLightPalette();
+
+  #ifdef Q_OS_WIN
+    // Qt on macOS already dynamically picks light/dark theme based on the OS setting
+    QSettings settings("HKEY_CURRENT_USER\\Software\\Microsoft\\Windows\\CurrentVersion\\Themes\\Personalize", QSettings::NativeFormat);
+    if (settings.value("AppsUseLightTheme") == 0)
+      {
+      QPalette palette =  standardDarkPalette();
+      }
+  #endif
```
jamesobutler, line 205: I'm doing something wrong here as even if dark mode is specified for apps on Windows, it is always using the standardLightPalette instead of the standardDarkPalette when using the "Slicer" style which is OS controlled.

lassoan: Probably you do not want to redefine "QPalette" in the if block.

jamesobutler: ah yes. Fixed.

jamesobutler:
I've pushed updates that I've confirmed only on Windows where there the default style is still "Slicer" but chooses the "Slicer Light" mode if the Windows setting "Choose your default app mode" is "Light" or chooses the "Slicer Dark" mode if the Windows setting is "Dark".  There is also the option to override the OS specification by picking the options "Slicer Light" or "Slicer Dark". 
image

I don't believe this will change anything about the Dark text used in some widgets on macOS as the SlicerLight style is still being applied and then macOS is automatically inverting the colors to dark mode if that is the current system setting.

Just a reminder, there are other colors issues as described in #4803 that aren't being addressed in this PR.

lassoan:
Thank you for the update. This should work well.

The only issue I see is that bright palette only sets 9 colors (not all the 24 colors as the dark palette). This means that it does not force a bright palette and on Mac you end up having half of the colors with the default palette, so on Mac you could end up with a palette consisting half dark, half bright colors.

I think the default palette should specify minimal number of colors (maybe none?) and bright and dark palettes should set all colors.

Could you investigate Mac color issues if we gave you remote access to a Mac?

jcfr:
I suggest to add an entry in the documentation to describe the theme.  An other entry could be added to https://slicer.readthedocs.io/en/latest/user_guide/settings.html#panels

It would briefly describe the 3 themes. Optionally image could be included, there would have to be uploaded at https://github.com/Slicer/Slicer/releases/tag/docs-resources

Docs/user_guide/settings.md
```diff
@@ -69,6 +69,18 @@ List of modules that appear in the Favorites toolbar:
 
 To add a module, drag&drop it from the *Modules* list above. Then use the advanced panel (>>) to reorganize/delete the modules within the toolbar.
 
+### Appearance
+
+#### Style
+
+The overall theme of Slicer is controlled by the selected Style:  `Slicer`, `Light Slicer` or `Dark Slicer`
+
+`Slicer` is the default and it sets the style based on theme settings set by the operating system. For example, on Windows if [dark mode][windows-dark-mode] is turned on for apps, then the `Dark Slicer` style will be used upon launching Slicer.
```
jcfr, line 78: What do you think of being more explicit. Indeed after looking at the implementation of the Slicer style, it looks like the dark palette will only be used on windows when `Slicer` is the default style ? 

Raw:
```
The overall theme of Slicer is controlled by the selected Style:  `Slicer`, `Light Slicer` or `Dark Slicer`

`Slicer` is the default and it sets the style based on theme settings set by the operating system.

The table below summarizes how the appearance of the application is updated when the `Slicer` style is selected:

|             |  Dark Operating System Theme |  Light Operating System Theme |
|-------------|------------------------------|-------------------------------|
| **Linux**   | `Light Slicer`               | `Light Slicer`                |
| **macOS**   | `Light Slicer`               | `Light Slicer`                |
| **Windows** | `Dark Slicer`                | `Light Slicer`                |

`Light Slicer` and `Dark Slicer` are options used to override the system setting.


```

Rendered:

The overall theme of Slicer is controlled by the selected Style:  `Slicer`, `Light Slicer` or `Dark Slicer`

`Slicer` is the default and it sets the style based on theme settings set by the operating system.

The table below summarizes how the appearance of the application is updated when the `Slicer` style is selected:

|             |  Dark Operating System Theme |  Light Operating System Theme |
|-------------|------------------------------|-------------------------------|
| **Linux**   | `Light Slicer`               | `Light Slicer`                |
| **macOS**   | `Light Slicer`               | `Light Slicer`                |
| **Windows** | `Dark Slicer`                | `Light Slicer`                |

`Light Slicer` and `Dark Slicer` are options used to override the system setting.

jcfr: I wonder if there is a better way to convey the fact that `Light Slicer` will always be used on macOs and Linux when `Slicer` is the selected style.

lassoan: For users the behavior is the following: "Slicer theme" sets dark/light mode based on operating system settings. Slicer dark/light theme: dark/light mode is used, regardless of operating system settings. Current limitation: "Slicer theme" always uses light mode on Linux.

lassoan:
I've tested this on Windows and the dark them looks very nice! Thank you @jamesobutler.

I could not try the theme switching because for me the bright option does not appear. I guess a complete rebuild from scratch would fix the issue, but do you know by any chance how to force the bright options to appear without a full rebuild? 

Considering that text visibility is now very good in dark mode and on Mac dark mode is controlled by the operating system already, I think we can merge this.

Docs/user_guide/settings.md
jcfr: Gotcha. I missed that comment:

https://github.com/Slicer/Slicer/blob/9bacb34fa5e26c02d79501af30d8c137ce2c4e4c/Base/QTGUI/qSlicerStyle.cxx#L199

How does this work ?

jamesobutler:
> do you know by any chance how to force the bright options to appear without a full rebuild?

I don’t remember exactly what I did but it might’ve just been a reconfigure/generate then build all again for the super build. This not being a build from scratch.

I’m not sure how exactly the automatic changing of colors on macOS works but the one line that isn’t changing is this color definition of `palette.setColor(QPalette::Window, Qt::white)` and macOS was already changing it to something dark. I’m not really sure how toggling the styles will actually work on macOS. I don’t think Apple would encourage style override options in the application as we’ve done. I haven’t had the time to investigate on other platforms besides Windows. Even the newer Ubuntu 20.04 has a system dark mode, but I don’t know much about it.

For those in the states, enjoy your 4th of July weekend 🎆

lassoan:
@jamesobutler Were you able to test this on Mac? Let me know if you need remote access.

jamesobutler:
I only got as far as cherry-picking these commits onto the source used for that shared Mac. It’s needs a configure and build, but I haven’t had a lot of time outside of work this week.

 I did seem to notice when using the current build on the shared Mac that the current “Slicer” style actually looked like the “fusion” style. Toggling would apply styles correctly but on start it wouldn’t be correct. I’m surprised more Mac users haven’t commented about it if they use the macOS system light mode. It is the more light gray look instead of mostly white as the current slicer style details.

pieper:
Yes, the mac style has always been a bit different.  It's never bothered me.  I don't use the dark mode because the icons are hard to see and the python console used to be unusable.  I could test if needed.

image|690x321, 50%

lassoan:
Terminal colors will be fixed by this PR: https://github.com/Slicer/Slicer/pull/5028

I agree that icons look bad in dark mode. I think this would be the last major thing to fix before set dark mode the default. Since we need to fix the icons anyway (make them higher resolution and all of them consistent), it could be a design requirement to make them compatible with both light and dark backgrounds. I think once the Slicer website and extension manager is fixed, the Kitware designers may be able to help with the icons, too (but we should probably ask others, too, maybe we could speed things up).

Docs/user_guide/settings.md
jcfr: After documenting this behavior, moving forward with the integration of this makes sense.

jcfr: That said, should we should keep `Slicer Light` ? 

If not, should we introduce a build settings to change the default ? In context of custom  application, it is important to ensure a style is consistently used.

lassoan: Yes, we need all three themes:
- Slicer: automatic bright or dark (based on OS settings)
- Slicer Light: always light (regardless of OS settings)
- Slicer Dark: always dark (regardless of OS settings)

jamesobutler:
@pieper I'll take you up on helping to test this PR on macOS.  I haven't had a lot of free time to spend continuing to debug issues on other platforms.  The goal is to have each platform have the same style where the "Slicer" style is based on the OS setting and will look like SlicerLight (mostly white) if in light mode or SlicerDark if in dark mode.

pieper:
I merged your PR changes into a master branch build of Slicer and I'm testing on macOS 10.14 (not the very latest).  I'm using Qt.14.1

Right now it's not behaving right and sometimes gets in unusable states if you switch the system theme while Slicer is open (some UI elements change, some do not).  
image

There's other cross-talk too - like you can start up with Slicer stye you get the slicer-mac-off-white, but toggling the system gets you to white Slicer style like you get on windows.

I suspect if we test with the latest qt and the latest macOS we might see different things.   I agree with Andras that the ideal is to have three modes: slicer light, slicer dark, and follow system settings.  It's a bit tedious to debug though since not all UI elements are responding the same way to the two different mode settings and the fresh-restart vs already running condition.  I don't know if there are lingering Qt issues or if there are style overrides in the CTK or Slicer code causing this.

lassoan:
It may be normal that dynamic switching does not work. Are things consistent after a restart?

pieper:
As far as I can tell yes.

jamesobutler:
> There's other cross-talk too - like you can start up with Slicer stye you get the slicer-mac-off-white, but toggling the system gets you to white Slicer style like you get on windows.

This isn't the intended behavior. On start of Slicer with the "Slicer" style the one selected in settings, it should be either the true SlicerDark or the SlicerLight (white style) and not the off-white "Fusion" standard style that has been the typical Slicer look on mac.

Docs/user_guide/settings.md
jcfr: Agreed, I realized my last comment was confusing. 

Instead of `should we should keep Slicer Light ?`, I meant to ask if  we should keep `Slicer Light` as the default theme ? 

If not, I suggest we introduce a build setting allowing to  choose which theme is the default one.

jamesobutler:
I've pushed some additional updates to fix some other areas that weren't respecting the current style palette which included the qSlicerAboutDialog and the qSlicerErrorReportDialog.

I've also included an update that removes a specification of a gray color that is currently seen between slice viewers to now use the QPalette::Window specification instead.  This in combination with https://github.com/commontk/CTK/pull/923 address the issue described in https://github.com/Slicer/Slicer/issues/4803#issuecomment-644255623.

The DICOM database now updates correctly when changing style within the Slicer session
| Dark | Light |
|------|--------|
|image|image|

Now that the space between the slice views uses the QPalette::Window specification, the main window looks like the following:
| Dark | Light |
|------|--------|
| image| image|

lassoan:
@jamesobutler Would you mind cherry-picking changes proposed in https://github.com/Slicer/Slicer/pull/5028 to consolidate all dark-mode related changes in a single pull request?

jamesobutler:
@lassoan I've cherry-picked your commit to fix color issues with the python console to this PR.

jamesobutler:
- [ ] Currently the text color in Module finder dialog isn't updating appropriately when the style is changed in the session. The text color of the module selector in the toolbar however does change correctly. This image shows starting Slicer in dark mode and then switching to the Slicer light style. image

lassoan:
Module finder: `item->setForeground` in qSlicerModulesListView.cxx is set based on the module status (successfully loaded, failed to load, etc.). The color is taken from the style sheet but then when the style is switched then there is no update. @jamesobutler do you know if there is an event that could be observed in qSlicerModulesListView that is called after the style is changed? We could call `this->updateModules();` to force a refresh of the colors.

jamesobutler:
@lassoan I believe you could try using the QEvent ApplicationPaletteChange to observe when we change the palette and then call `updateModules()`.  The `paletteChanged` signal of a QGuiApplication object was deprecated in favor of this QEvent.

lassoan:
I have a fix for the module finder. If there is no objection, I merge this branch along with that fix.

jamesobutler:
That's good with me @lassoan !

lassoan:
Thanks a lot @jamesobutler for all your work on this. The changes have now been merged in https://github.com/Slicer/Slicer/commit/566a81923988fd3a3e8443523bd76fe8d50b62ee, https://github.com/Slicer/Slicer/commit/2054a88780bf961bf0b0f4b2f7c74275371170f2, https://github.com/Slicer/Slicer/commit/044c7667ec82ee1b62ed6304f52d96f3915432ce

PR #4995, opened 2020-06-19
Doc computer requirement updates

jamesobutler:
Updates here are motivated by the change in https://github.com/mhalle/slicer4-download/pull/17 to link to https://slicer.readthedocs.io/en/latest/user_guide/getting_started.html#system-requirements from the download page.

Please check the OS versions I specified especially for Slicer stable.  I'm not exactly sure what is the lowest known well working version of Slicer stable on a popular linux distro like Ubuntu.

### DIFF
--- a/Docs/user_guide/getting_started.md
+++ b/Docs/user_guide/getting_started.md
@@ -6,6 +6,18 @@ Welcome to the 3D Slicer community. This page contains information that you need
 
 3D Slicer runs on any Windows, Mac, or Linux computer that was released in the last 5 years. Older computers may work (depending mainly on graphics capabilites). Slicer can also run on virtual machines and docker containers.
 
+### Operating system versions
+
+|               |Windows                                        |macOS                             |Linux                                     |
+|---------------|-----------------------------------------------|----------------------------------|------------------------------------------|
+|Stable Release |Windows 10 or later<br>Windows 8.1<br>Windows 7|OS X Mavericks (10.9) or later    |Ubuntu 16.04 or later<br>CentOS 7 or later|
+|Preview Release|Windows 10 or later<br>Windows 8.1<br>Windows 7|macOS High Sierra (10.13) or later|Ubuntu 18.04 or later<br>CentOS 7 or later|
+
+**Notes:**
+- It is always recommended to use the latest public release of Windows or macOS.  For Linux, this would be the latest LTS (Long-term-support) version.
+- Slicer is expected to work on the vast majority of desktop and server Linux distributions. The system is required to provide at least GLIBC 2.17 and GLIBCCC 3.4.19. For more details, read [here](https://www.python.org/dev/peps/pep-0599/#the-manylinux2014-policy).
+- Only 64-bit Slicer installers are available to download. Developers can attempt to build 32-bit versions on their own if they need to run Slicer on a 32-bit operating system. That said, this should be carefully considered as many clinical research tasks, such as processing of large CT or MR volumetric datasets, require more memory than can be accommodated with a 32-bit program.
+
 ### Recommended hardware configuration:
 - Memory: more than 4GB (8 or more is recommended). As a general rule, have 10x more memory than the amount of data that you load.
 - Display: a minimum resolution of 1024 by 768 (1280 by 1024 or better is recommended).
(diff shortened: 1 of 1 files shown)


Docs/user_guide/getting_started.md
```diff
@@ -6,29 +6,35 @@ Welcome to the 3D Slicer community. This page contains information that you need
 
 3D Slicer runs on any Windows, Mac, or Linux computer that was released in the last 5 years. Older computers may work (depending mainly on graphics capabilites). Slicer can also run on virtual machines and docker containers.
 
+## Operating system versions
```
jcfr, line 9:
```
### Operating system versions
```

Docs/user_guide/getting_started.md
```diff
@@ -6,29 +6,35 @@ Welcome to the 3D Slicer community. This page contains information that you need
...
+
+|               |Windows                                        |macOS                             |Linux                                     |
+|---------------|-----------------------------------------------|----------------------------------|------------------------------------------|
+|Stable Release |Windows 10 or later<br>Windows 8.1<br>Windows 7|OS X Mavericks (10.9) or later    |Ubuntu 16.04 or later<br>CentOS 7 or later|
+|Preview Release|Windows 10 or later<br>Windows 8.1<br>Windows 7|macOS High Sierra (10.13) or later|Ubuntu 18.04 or later<br>CentOS 7 or later|
+
+**Notes:**
+- It is always recommended to use the latest public release of your operating system.
+- Most recent versions of popular Linux distributions should work.
+- Only 64-bit Slicer versions are available to download. Developers can attempt to build 32-bit versions on their own if they need to run Slicer on a 32-bit operating system.
```
jcfr, line 19: I like the previous text associated with `32 bit versus 64 bit`, may be we could keep it. Or even introduce a `Frequently Asked Question` sub-section ?

Docs/user_guide/getting_started.md
```diff
@@ -6,29 +6,35 @@ Welcome to the 3D Slicer community. This page contains information that you need
...
-32 bit versus 64 bit: Many clinical research tasks, such as processing of large CT or MR volumetric datasets, require more memory than can be accommodated with a 32 bit program. Therefore, we only make 64-bit Slicer versions available. Developers can build 32-bit version on their own if they need to run Slicer on a 32-bit operating system.
-
 ## Installing 3D Slicer
 
 To download Slicer, click [here](http://download.slicer.org/).
 
 ![](https://github.com/Slicer/Slicer/releases/download/docs-resources/getting_started_download.png)
 
-The "Preview Release" of 3D Slicer is updated every night and contains latest developments: new features and fixes. The "Stable Release" is updated about once a year nightly and is more rigorously tested.
+The "Preview Release" of 3D Slicer is updated every night and represents the latest development including new features and fixes.
```
jcfr, line 36: `updated every night` -> `updated daily (process starts at 11pm ET and takes few hours to complete)`

Docs/user_guide/getting_started.md
```diff
@@ -6,29 +6,35 @@ Welcome to the 3D Slicer community. This page contains information that you need
...
+The "Stable Release" is updated about once a year and is more rigorously tested.
```
jcfr, line 37: `updated about once a year` -> `updated few times a year`

Docs/user_guide/getting_started.md
```diff
@@ -6,29 +6,35 @@ Welcome to the 3D Slicer community. This page contains information that you need
...
 
+## Operating system versions
+
+|               |Windows                                        |macOS                             |Linux                                     |
+|---------------|-----------------------------------------------|----------------------------------|------------------------------------------|
+|Stable Release |Windows 10 or later<br>Windows 8.1<br>Windows 7|OS X Mavericks (10.9) or later    |Ubuntu 16.04 or later<br>CentOS 7 or later|
+|Preview Release|Windows 10 or later<br>Windows 8.1<br>Windows 7|macOS High Sierra (10.13) or later|Ubuntu 18.04 or later<br>CentOS 7 or later|
+
+**Notes:**
+- It is always recommended to use the latest public release of your operating system.
```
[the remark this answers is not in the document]
jamesobutler: This was mostly written in the context of Windows and macOS where every version is really an LTS version.  I could word it differently for Linux in the context of choosing latest LTS version.

Docs/user_guide/getting_started.md
jamesobutler: Yeah the previous text seemed to have been written in a time where 32-bit was more common than 64-bit or at least closer to half and half, so I was attempting to re-write it in the manner of expecting 64-bit with the option to choose 32-bit.  Even when running the suggested latest version of macOS it only supports running 64-bit applications so it gets a little more tricky of a comment as 32-bit support gets less and less supported.

jcfr: Good point.

What about something like this then:

```
- Only 64-bit Slicer installers are available to download. Developers can attempt to build 32-bit versions on their own if they need to run Slicer on a 32-bit operating system.  That said, this should be carefully considered as many clinical research tasks, such as processing of large CT or MR volumetric datasets, require more memory require more memory than can be accommodated with a 32 bit program.
```

jamesobutler: Yes, that sounds good.  I've updated the commit with this change.

PR #4996, opened 2020-06-20
BUG: Fix SimpleITK build error in debug mode

lassoan:
Trying to fix SimpleITK build error due to missing python_d.lib by undoing a change that was added to consistently force using release-mode Python in all external projects (by setting PYTHON_DEBUG_LIBRARY).
This change was necessary for VTK but seems to cause problems for SimpleITK.

re #4991

@cpinter Please confirm that this fixes the build error for you. You may need to delete PYTHON_DEBUG_LIBRARY from SimpleITK's CMake cache (or start a clean build).


lassoan:
This fixes all the debug-mode build errors for me, starting from a clean build. I will merge this if there is no specific objection.

jcfr:
Great news.

Could you squash the two commits together and update the commit message accordingly referencing the one that is being reverted.

The current history associated with this PR seems confusing :pray:

jcfr:
Thanks @lassoan

PR #5003, opened 2020-06-23
BUG: Add missing import to ScriptedLoadableModule

adamaji:
This commit fixes the following error when using takeScreenshot from a
ScriptedLoadableModuleLogic subclass:
```
  File "path\to\slicer\Python\slicer\ScriptedLoadableModule.py", line
  344, in takeScreenshot
      imageData = vtk.vtkImageData()
      NameError: name 'vtk' is not defined
```


jcfr:
Thanks for your contribution @adamaji  🙏

Issue #5008, opened 2020-06-24
SH plugin action not added to view context menu


pieper:
If I follow the same pattern for defining a context menu action for a view and the tree view the action only shows up in the subject hierarchy tree menu and not in the view

## Steps to reproduce

Save this module script as ViewContextMenuTest in a directory and then start slicer with the `--additional-module-path` pointing to that directory. 

Once Slicer starts, place a fiducial and right click in the view vs the Data module tree.  The item only shows up in the data tree.

```
import os
import vtk, qt, ctk, slicer
import warnings
from slicer.ScriptedLoadableModule import *

from SubjectHierarchyPlugins import AbstractScriptedSubjectHierarchyPlugin

class ViewContextMenuTest(ScriptedLoadableModule):
    def __init__(self, parent):
        ScriptedLoadableModule.__init__(self, parent)
        parent.title = "Test Subject Hierarchy Plugin Loader"
        parent.categories = [""]
        parent.contributors = ["Steve Pieper, Isomics, Inc."]
        parent.helpText = ""
        parent.hidden = not slicer.util.settingsValue('Developer/DeveloperMode', False, converter=slicer.util.toBool)

        #
        # register subject hierarchy plugin once app is initialized
        #
        def onStartupCompleted():
            import SubjectHierarchyPlugins
            from ViewContextMenuTest import ViewContextMenuTestSubjectHierarchyPlugin
            scriptedPlugin = slicer.qSlicerSubjectHierarchyScriptedPlugin(None)
            scriptedPlugin.setPythonSource(ViewContextMenuTestSubjectHierarchyPlugin.filePath)
            print('ViewContextMenuTestWidget loaded')
        slicer.app.connect("startupCompleted()", onStartupCompleted)


class ViewContextMenuTestWidget(ScriptedLoadableModuleWidget):
  """
  dummy module widget to allow reloading in developer mode
  """
  def setup(self):
    ScriptedLoadableModuleWidget.setup(self)

# debug helper
def whoami():
    import sys
    return "ViewContextMenuTest: " + sys._getframe(1).f_code.co_name


class ViewContextMenuTestSubjectHierarchyPlugin(AbstractScriptedSubjectHierarchyPlugin):

  # Necessary static member to be able to set python source to scripted subject hierarchy plugin
  filePath = __file__

  def __init__(self, scriptedPlugin):
    print(whoami())
    print('ViewContextMenuTestWidget calling superclass')
[...]

PR #5009, opened 2020-06-24
BUG: add plugin actions to view context

pieper:
Addresses #5008. Could be considered a fix unless there's
a better suggestion.

### DIFF
--- a/Modules/Loadable/SubjectHierarchy/Widgets/qSlicerSubjectHierarchyPluginLogic.cxx
+++ b/Modules/Loadable/SubjectHierarchy/Widgets/qSlicerSubjectHierarchyPluginLogic.cxx
@@ -510,6 +510,10 @@ void qSlicerSubjectHierarchyPluginLogic::onDisplayMenuEvent(vtkObject* displayNo
   foreach (qSlicerSubjectHierarchyAbstractPlugin* plugin, qSlicerSubjectHierarchyPluginHandler::instance()->allPlugins())
     {
     plugin->hideAllContextMenuActions();
+    foreach (QAction *action, plugin->viewContextMenuActions())
+      {
+      d->ViewMenu->addAction(action);
+      }
     plugin->showViewContextMenuActionsForItem(itemID, eventDataMap);
     }
 


lassoan:
I think the current implementation was intentional, but I don't remember the details. @cpinter do you remember? Can you check if it works in the spine project (menu does not get cluttered with extra items)?

pieper:
I don't think it was intentional because as far as I can tell it's not working 😉   But yes, if there's an example to follow that would be great.  I hunted around quite a bit but didn't find anything I could make work.

cpinter:
Currently the two types of context menus cannot be used at the same time due to a limitation described here: https://github.com/Slicer/Slicer/blob/master/Modules/Loadable/Annotations/SubjectHierarchyPlugins/AnnotationsSubjectHierarchyPlugin.py#L96-L107

It seems that the plugin needs to be registered at a different time. I'm investigating how this can be prevented.

cpinter:
As I remember, this difference was also due to the view menu whitelist, see https://github.com/Slicer/Slicer/blob/master/Modules/Loadable/SubjectHierarchy/Widgets/qSlicerSubjectHierarchyPluginLogic.cxx#L625, so if it also works with this change, then it may be a good solution, but let me finish looking into this more.

The whitelist can be used like this:
```
    pluginHandler = slicer.qSlicerSubjectHierarchyPluginHandler.instance()
    pluginLogic = pluginHandler.pluginLogic()
    whitelistedActions = ['CustomAction']
    pluginLogic.setDisplayedViewMenuActionNames(whitelistedActions)
```
In your example the action has to be named like this:
```
self.viewAction.objectName = 'CustomAction'
```

pieper:
Thanks for your help @cpinter !  Unfortunately the whitelist option isn't working for me either.

cpinter:
The problem is that registering the plugin happens sooner than setting its python source, so the view context menus are returned empty at the time of registering. I can see two options other than the one suggested here (which adds the view context menu actions every time when the menu is requested, and I suspect it would break the whitelist function):

1. Call register with delay in the abstract plugin, something like this:
```
  def __init__(self, scriptedPlugin):
    self.scriptedPlugin = scriptedPlugin

    qt.QTimer.singleShot(50, self.register) # Here register was called simply using self.register
```

It would be better to tie this to startupCompleted, but unfortunately setting up these non-core plugins happens later than that, so that signal is never caught. I tried to find a good way to check if startup has been completed, but haven't been able to. I'm thinking about something like this:

```
  def __init__(self, scriptedPlugin):
    self.scriptedPlugin = scriptedPlugin

    # Register plugin when startup is completed
    def onStartupCompleted():
      qt.QTimer.singleShot(50, self.register)
    if slicer.app.isStartupCompleted:
      qt.QTimer.singleShot(50, self.register)
    else:
      slicer.app.connect("startupCompleted()", onStartupCompleted)
```

Delaying is still needed, because if the modules do the setup on startup completed, then we need to make sure it happens sooner than registering.


2. Add the registration step to the commands that need to be called manually:

```
import SubjectHierarchyPlugins
from ViewContextMenuTest import ViewContextMenuTestSubjectHierarchyPlugin
scriptedPlugin = slicer.qSlicerSubjectHierarchyScriptedPlugin(None)
scriptedPlugin.setPythonSource(ViewContextMenuTestSubjectHierarchyPlugin.filePath)
pluginHandlerSingleton = slicer.qSlicerSubjectHierarchyPluginHandler.instance()
pluginHandlerSingleton.registerPlugin(scriptedPlugin)
print('ViewContextMenuTestWidget loaded')
```

None of the three solutions is perfect, but to me maybe the last one seems to be the cleanest. I tested the whitelist function and it works with both the above.

pieper:
I added the two suggested lines to `onStartupCompleted`

```
pluginHandlerSingleton = slicer.qSlicerSubjectHierarchyPluginHandler.instance()
pluginHandlerSingleton.registerPlugin(scriptedPlugin)
```

but this did not work.  It also didn't work if I called them in a timer after startup completed.

Do you think it will be possible to have a scripted view context menu without major changes?  If there's a lot of work required I can look for ways to bypass the SHPlugin, but of course I'd prefer to have a unified solution that can serve as an example.

cpinter:
Hm. Is there a debug message in the log about a plugin being already registered? It is possible that the plugin was registered before being set up, and this second call was a no-op.

Well the obvious workaround is to have two plugins, one handling the tree context menu and the other the view context menu as described in the Annotations plugin I referred to above.

If this is about fixing the plugins so that the two can work from the same plugin, then it will require some work, probably not much. If you try the first approach I suggested does it work?

Also, it would help a lot if I got an answer to the question about whether there is a way to determine if the startupComplete event already happened.

cpinter:
I confirm that the second suggestion doesn't work without applying some of the changes from the first. It seems that we need to apply the delayed registration change in order to have the view context menus work the same way as the item context menu (and thus being able to use both from the same plugin).

pieper:
> Hm. Is there a debug message in the log about a plugin being already registered? It is possible that the plugin was registered before being set up, and this second call was a no-op.

No, there's no debug message about duplicate registrations that I can see.


> Well the obvious workaround is to have two plugins, one handling the tree context menu and the other the view context 
menu as described in the Annotations plugin I referred to above.

I'm sorry, I'm not following - the Annotations plugin appears not to define any view context menu actions?  Is there a python example that does?

> If this is about fixing the plugins so that the two can work from the same plugin, then it will require some work, probably not much. If you try the first approach I suggested does it work?

No, I don't need this to be one plugin, two would be fine.  But even if I take out `itemContextMenuActions` and `showContextMenuActionsForItem` from the class I still don't get my view action to show in the context menu.

> Also, it would help a lot if I got an answer to the question about whether there is a way to determine if the startupComplete event already happened.

Sorry, I didn't see that question.  Yes, I'm sure `startupComplete` has happened because I register the callback with that signal is triggered.

I'm still testing with the code in #5008 and when I trace through it seems that the actions are never added to the context menu, which is why this PR is one fix.  I don't understand your comment above about "so the view context menus are returned empty at the time of registering."  Maybe we could look at that more closely?  To me it seems that all the loadable module C++ code will have already run by the time `startupComplete` is triggered so it should be ready for the scripted plugin.

cpinter:
> I'm sorry, I'm not following - the Annotations plugin appears not to define any view context menu actions? Is there a python example that does?

There is a few line long description in that plugin about how to make it work in general. It's there because the Annotations plugin was the first scripted plugin, so I made that an example, containing all the information and empty functions.

> No, I don't need this to be one plugin, two would be fine. But even if I take out itemContextMenuActions and showContextMenuActionsForItem from the class I still don't get my view action to show in the context menu.

It's not about taking out those functions, it's about registering them differently, as explained in the Annotations plugin (I added a link to those comments above)

cpinter:
@lassoan We had a quick chat with @pieper and sorted this out. I further simplified Steve's code and added it as an example to script repository: https://www.slicer.org/wiki/Documentation/Nightly/ScriptRepository#Subject_hierarchy_plugin_offering_view_context_menu_action

pieper:
Fixed with improved example code - thanks @cpinter !

PR #5018, opened 2020-06-26
ENH: Update transform node name when the transform is inverted

lassoan:
When inverting a transform by clicking "Invert" button, it is easy to lose track how many times invert was clicked and the user does not know if the transform ends up being the same as the original or inverted.

We change the behavior so that a (-) suffix as appended to or removed from the transform's name each time "Invert" is clicked.

image

image

### DIFF
--- a/Libs/MRML/Core/vtkMRMLTransformNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLTransformNode.cxx
@@ -39,6 +39,7 @@ Version:   $Revision: 1.14 $
 #include <vtkPoints.h>
 #include <vtkThinPlateSplineTransform.h>
 #include <vtkTransform.h>
+#include <vtksys/SystemTools.hxx>
 
 // STD includes
 #include <sstream>
@@ -1191,6 +1192,22 @@ void vtkMRMLTransformNode::Inverse()
   this->StorableModifiedTime.Modified();
   this->Modified();
   this->TransformModified();
+}
+
+//----------------------------------------------------------------------------
+void vtkMRMLTransformNode::InverseName()
+{
+  std::string nodeName = (this->GetName() ? this->GetName() : "");
+  const std::string inverseSuffix(" (-)"); // this could be moved to a member variable
+  if (vtksys::SystemTools::StringEndsWith(nodeName, inverseSuffix.c_str()))
+    {
+    nodeName.erase(nodeName.size() - inverseSuffix.size(), inverseSuffix.size());
+    }
+  else
+    {
+    nodeName += inverseSuffix;
+    }
+  this->SetName(nodeName.c_str());
 }
 
 //----------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


Modules/Loadable/Transforms/SubjectHierarchyPlugins/qSlicerSubjectHierarchyTransformsPlugin.cxx
```diff
@@ -408,7 +408,9 @@ void qSlicerSubjectHierarchyTransformsPlugin::invert()
     shNode->GetItemDataNode(currentItemID) );
   if (transformNode)
     {
+    MRMLNodeModifyBlocker blocker(transformNode);
     transformNode->Inverse();
+    transformNode->InverseName();
```
pieper, line 413: Wouldn't it make more sense to put the `InverseName` inside the `Inverse` method of the node rather than having duplicate logic in the GUI code.  Also if anyone outside uses the `Inverse` method they will get consistent results.

Libs/MRML/Core/vtkMRMLTransformNode.cxx
```diff
@@ -1193,6 +1194,22 @@ void vtkMRMLTransformNode::Inverse()
   this->TransformModified();
 }
 
+//----------------------------------------------------------------------------
+void vtkMRMLTransformNode::InverseName()
```
pieper, line 1198: I feel like this should be an independent member variable like `HasBeenInverted` and the display should be just a gui feature.  Although I guess the benefit of changing the name is that it will appear consistently on all comboboxes, list view, etc.  But it will also be in the default filename, which is kind of odd.  Overall I guess it's fine to change the name.

Modules/Loadable/Transforms/SubjectHierarchyPlugins/qSlicerSubjectHierarchyTransformsPlugin.cxx
lassoan: I've though about that, too. In fact, I've implemented like that initially. But it then I moved it out to a separate function to not change current behavior, as modules may not expect this side effect. For example, if Inverse() would include InverseName() then result of SetName(); Inverse(); would differ from Inverse(); SetName().

I talked to @ungi about this, too, and he preferred keeping these two methods separate (at MRML level).

Libs/MRML/Core/vtkMRMLTransformNode.cxx
lassoan: This is an interesting point. We actually have a notion of "inverted" transform (shown in the Information section in Transforms module), but it is really about the consistency of the node name and content. If you invert the transform and save it then you do want to reflect that in the name.

lassoan:
@pieper Thank you for the review and raising these questions. After talking to @ungi and thinking about it a bit more, I would lean towards keeping the current implementation (separate Inverse/InverseName and actually change the transform name), unless you disagree.

lassoan:
@pieper please check and approve if you agree that it's OK like this (merging is blocked if there are "changes requested" in the review)

Libs/MRML/Core/vtkMRMLTransformNode.cxx
pieper: Yes, makes sense.

PR #5024, opened 2020-06-30
New module finder dialog

lassoan:
This commit replaces module selector combobox search and "All modules" submenu by a new module finder dialog.

Module menu's "All modules" section could not show all the modules in one screen.

Solution: Replaced the module toolbar's combobox and the "All modules" list with a  "Module finder" window, which is brought up when user clicks module search icon or hits Ctrl/Cmd + F.
It can select a module using the same number of clicks as the combobox (type characters, press arrow-down key N times, then press Enter to go to that module).

Built-in modules can be hidden, so it is easy to find modules installed by extensions.
Module category is shown (e.g., Cleaner module is available in Surface models->Advanced category).
Developer information can be optionally displayed: module type (CLI, C++ loadable, scripted loadable), location, dependencies.

Demo video:

https://youtu.be/bhPbT0pu2K8

### DIFF
--- a/Base/QTCore/qSlicerUtils.h
+++ b/Base/QTCore/qSlicerUtils.h
@@ -26,6 +26,8 @@
 
 #include "qSlicerBaseQTCoreExport.h"
 
+class qSlicerAbstractCoreModule;
+
 class Q_SLICER_BASE_QTCORE_EXPORT qSlicerUtils
 {
 
@@ -52,6 +54,11 @@ public:
   /// \note Associated \a fileName is expected to match the following
   /// regular expression: "(lib)?qSlicer.+Module\\.(so, dll, dylib)"
   static bool isLoadableModule(const QString& filePath);
+
+  /// Return \a true is the module is for testing purposes.
+  /// These modules are for testing and for troubleshooting and normally should not be displayed
+  /// to end users.
+  static bool isTestingModule(qSlicerAbstractCoreModule* module);
 
   /// Look for target file in build intermediate directory.
   /// On windows, the intermediate directory includes: . Debug RelWithDebInfo Release MinSizeRel
--- a/Base/QTGUI/qSlicerModuleFinderDialog.cxx
+++ b/Base/QTGUI/qSlicerModuleFinderDialog.cxx
@@ -1,0 +1,404 @@
+/*==============================================================================
+
+  Program: 3D Slicer
+
+  Copyright (c) Kitware Inc.
+
+  See COPYRIGHT.txt
+  or http://www.slicer.org/copyright/copyright.txt for details.
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+
+==============================================================================*/
+
+// STD includes
+#include <algorithm>
+
+// Qt includes
+#include <QKeyEvent>
+#include <QPushButton>
+
+// Slicer includes
+#include "qSlicerAbstractCoreModule.h"
+#include "qSlicerAbstractModule.h"
+#include "qSlicerApplication.h"
+#include "qSlicerModuleFactoryFilterModel.h"
+#include "qSlicerModuleFactoryManager.h"
+#include "qSlicerModuleManager.h"
+#include "qSlicerModuleFinderDialog.h"
+#include "qSlicerUtils.h"
+#include "ui_qSlicerModuleFinderDialog.h"
+
+// --------------------------------------------------------------------------
+// qSlicerModuleFinderDialogPrivate
+
+//-----------------------------------------------------------------------------
+class qSlicerModuleFinderDialogPrivate: public Ui_qSlicerModuleFinderDialog
+{
+  Q_DECLARE_PUBLIC(qSlicerModuleFinderDialog);
+protected:
+  qSlicerModuleFinderDialog* const q_ptr;
+
+public:
+  qSlicerModuleFinderDialogPrivate(qSlicerModuleFinderDialog& object);
+
+  /// Convenient method regrouping all initialization code
+  void init();
+
+  void makeSelectedItemVisible();
+
+  QString CurrentModuleName;
+};
+
+// --------------------------------------------------------------------------
+// qSlicerModuleFinderDialogPrivate methods
+
+// --------------------------------------------------------------------------
+qSlicerModuleFinderDialogPrivate::qSlicerModuleFinderDialogPrivate(qSlicerModuleFinderDialog& object)
+  :q_ptr(&object)
+{
+}
+
+// --------------------------------------------------------------------------
+void qSlicerModuleFinderDialogPrivate::init()
+{
+  Q_Q(qSlicerModuleFinderDialog);
+
+  this->setupUi(q);
+
+  qSlicerCoreApplication * coreApp = qSlicerCoreApplication::application();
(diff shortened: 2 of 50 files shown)


finetjul:
LGTM, thanks. 

I wonder if optionally searching through the module documentation could be useful. For example, someone may want to "smooth" an image, unfortunately, the corresponding filter is "Gaussian Blur Image Filter". If "smooth" was mentionned in the filter documentation (unfortunately not the case), it could be found that way. (same thing for the SimpleFilters module that should list all the available ITK filters in the documentation)

WDYT ?

lassoan:
Adding an option to search in the documentation would be a good idea. Do you have a suggestion for how it could be implemented? (I think you are the original author of qSlicerModuleFactoryFilterModel)

Base/QTCore/qSlicerUtils.h
```diff
@@ -53,6 +53,8 @@ class Q_SLICER_BASE_QTCORE_EXPORT qSlicerUtils
   /// regular expression: "(lib)?qSlicer.+Module\\.(so, dll, dylib)"
   static bool isLoadableModule(const QString& filePath);
 
+  static bool isTestingModule(const QStringList& categories);
```
jcfr, line 56: `hasTestingCategory` ?

Base/QTGUI/qSlicerModuleFinderDialog.cxx
```diff
@@ -0,0 +1,342 @@
...
+#include <QPushButton>
+
+// QtGUI includes
+#include "qSlicerAbstractCoreModule.h"
+#include "qSlicerAbstractModule.h"
+#include "qSlicerApplication.h"
+#include "qSlicerModuleFactoryFilterModel.h"
+#include "qSlicerModuleFactoryManager.h"
+#include "qSlicerModuleManager.h"
+//#include "qSlicerModulesMenu.h"
```
jcfr, line 35: Remove

Base/QTGUI/qSlicerModuleFinderDialog.cxx
```diff
@@ -0,0 +1,342 @@
...
+==============================================================================*/
+
+// STD includes
+#include <algorithm>
+
+// Qt includes
+#include <QKeyEvent>
+#include <QPushButton>
+
+// QtGUI includes
```
jcfr, line 28: `// Slicer includes`

Base/QTGUI/qSlicerModuleFinderDialog.cxx
```diff
@@ -0,0 +1,342 @@
...
+};
+
+// --------------------------------------------------------------------------
+// qSlicerModuleFinderDialogPrivate methods
+
+// --------------------------------------------------------------------------
+qSlicerModuleFinderDialogPrivate::qSlicerModuleFinderDialogPrivate(qSlicerModuleFinderDialog& object)
+  :q_ptr(&object)
+{
+  //this->ModulesMenu = nullptr;
```
jcfr, line 66: Remove

Base/QTGUI/qSlicerModuleFinderDialog.cxx
```diff
@@ -0,0 +1,342 @@
...
+  Q_Q(qSlicerModuleFinderDialog);
+
+  this->setupUi(q);
+
+  qSlicerCoreApplication * coreApp = qSlicerCoreApplication::application();
+  qSlicerAbstractModuleFactoryManager* factoryManager = coreApp->moduleManager()->factoryManager();
+
+  qSlicerModuleFactoryFilterModel* filterModel = this->SlicerModulesListView->filterModel();
+
+  // Hide modules that do not have GUI (use cannot switch to them)
```
jcfr, line 81: `use cannot switch to them` -> `user cannot switch to them`

Base/QTGUI/qSlicerModuleFinderDialog.cxx
```diff
@@ -0,0 +1,342 @@
...
+      }
+    }
+
+  if (module)
+    {
+    d->CurrentModuleName = module->name();
+
+    d->ModuleNameLineEdit->setText(module->name());
+
+    QString type = QLatin1String("Core");
```
jcfr, line 158:
```
// Use "inherits" instead of "qobject_cast" because "qSlicerBaseQTCLI" depends on "qSlicerQTGUI"
```

Base/QTGUI/qSlicerModuleFinderDialog.h
```diff
@@ -0,0 +1,64 @@
...
+
+// CTK includes
+#include <ctkPimpl.h>
+
+// Slicer includes
+#include "QDialog.h"
+#include "qSlicerBaseQTGUIExport.h"
+
+/// Forward declarations
+class qSlicerModuleFinderDialogPrivate;
```
jcfr, line 29: Sort alphabetically

Base/QTGUI/qSlicerModuleFinderDialog.h
```diff
@@ -0,0 +1,64 @@
...
+==============================================================================*/
+
+#ifndef __qSlicerModuleFinderDialog_h
+#define __qSlicerModuleFinderDialog_h
+
+// CTK includes
+#include <ctkPimpl.h>
+
+// Slicer includes
+#include "QDialog.h"
```
jcfr, line 25: `QDialog.h` -> `QDialog` 

Move to `// Qt includes` sections

finetjul:
I like the idea of a custom Qt model role with keywords. 

A simpler approach could be to have whole documentation in Qt::ToolTipRole and filter on it. It would then display the documentation on long mouse over.

Base/QTCore/qSlicerUtils.h
lassoan: This method determines if the module is for testing only, the current implementation is to check the category but in the future it may use different mechanisms.

I'll add a comment to clarify this.

Base/QTGUI/qSlicerModuleFinderDialog.cxx
lassoan: ok!

Base/QTGUI/qSlicerModuleFinderDialog.cxx
lassoan: ok

Base/QTGUI/qSlicerModuleFinderDialog.cxx
lassoan: ok

Base/QTGUI/qSlicerModuleFinderDialog.cxx
lassoan: ok

Base/QTGUI/qSlicerModuleFinderDialog.cxx
lassoan: ok

Base/QTGUI/qSlicerModuleFinderDialog.h
lassoan: ok

Base/QTGUI/qSlicerModuleFinderDialog.h
lassoan: ok

Base/QTCore/qSlicerUtils.h
jcfr: Thanks.

Would it make sense to have the following signature then:

```
static bool isTestingModule(qSlicerAbstractCoreModule* module);
```

lassoan:
I've fixed all the code review comments, added full-text search option, and reviewed and fixed all core modules help/acknowledgement sections.

jcfr:
:+1:  :fire:

PR #5025, opened 2020-07-02
ENH: add writtenNodes prop to scripted writer

pieper:
A qSlicerFileWriter uses the writtenNodes QStringList variable
to indicate which nodes were written when the write method is
invoked.  Prior to this change, a qScriptedFileWriter had no
way to set this variable, so, for example, the Save Data
dialog would always report a failure when saving with a scripted
writer.

This change makes the method virtual in the super class so that
it can be made a wrapped property in the subclass.

This allows usage as in the example below.

class MarkupsFcsvFileWriter:
    def __init__(self, parent):
        self.parent = parent

    def description(self):
        return 'Markup points as fcsv'

    def fileType(self):
        return 'MarkupsFile'

    def canWriteObject(self, obj):
        return isinstance(obj, slicer.vtkMRMLMarkupsNode)

    def extensions(self, obj):
        return ['Fiducial csv (*.fcsv)']

    def write(self, properties):
        print("parent", self.parent)
        print("Write", properties)

        if (not "nodeID" in properties
                or not "fileName" in properties):
            logging.error("Bad properties passed to MarkupsFcsvFileWriter.write")
            return False
        markupNode = slicer.mrmlScene.GetNodeByID(properties["nodeID"])
        if not markupNode or not self.canWriteObject(markupNode):
            logging.error("Bad MarkupNode passed to MarkupsFcsvFileWriter.write")
            return False
        storageNode = slicer.vtkMRMLMarkupsFiducialStorageNode()
        slicer.mrmlScene.AddNode(storageNode)
        fileName = properties["fileName"]
        if fileName.endswith(".fcsv.fcsv"):
            fileName = fileName[:-5]
        properties["fileName"] = fileName
        storageNode.SetFileName(fileName)
        storageNode.SetURI(None)
        result = storageNode.WriteData(markupNode)
        if result:
            self.parent.writtenNodes = [markupNode.GetID()]
        return bool(result)

### DIFF
--- a/Base/QTCore/qSlicerScriptedFileWriter.h
+++ b/Base/QTCore/qSlicerScriptedFileWriter.h
@@ -37,6 +37,9 @@ class Q_SLICER_BASE_QTCORE_EXPORT qSlicerScriptedFileWriter
 {
   Q_OBJECT
 
+  /// This property allows the writer to report back what nodes it was able to write
+  Q_PROPERTY(QStringList writtenNodes READ writtenNodes WRITE setWrittenNodes)
+
 public:
   typedef qSlicerFileWriter Superclass;
   qSlicerScriptedFileWriter(QObject* parent = nullptr);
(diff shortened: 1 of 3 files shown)


Base/QTCore/qSlicerScriptedFileWriter.h
```diff
@@ -71,6 +74,19 @@ class Q_SLICER_BASE_QTCORE_EXPORT qSlicerScriptedFileWriter
...
   bool write(const qSlicerIO::IOProperties& properties) override;
 
+  /// Added so node writers can report back written nodes
+  /// \sa qSlicerFileWriter::writtenNodex()
+  void addWrittenNode(const QString& writtenNode);
+
+  /// Reimplemented to propagate to python methods
+  /// \sa qSlicerFileWriter::writtenNodes()
+  QStringList writtenNodes()const override {
+    return Superclass::writtenNodes();
```
lassoan, line 84: Did you do this because you had to expose the method on public interface and did not want to modify method visibility in the base class? If yes, then please add a comment that explains it.

pieper: Including this no-op override of `writtenNodes` is not required, but to me it makes this code more readable and self contained.  I'll add more comments.

pieper:
Thanks for the review @lassoan  👍

PR #5026, opened 2020-07-03
BUG: Fix name of node loaded from filename containing multiple dots

lassoan:
When loading filenames, such as segmentation.1.seg.nrrd, the loaded node was named "segmentation" by default (everything after the first "." character was ignored).

Fixes issue reported at https://discourse.slicer.org/t/cannot-rename-markupsfiducial-in-load-data-window/11994/7?u=lassoan


lassoan:
Thanks for the review!

PR #5027, opened 2020-07-03
BUG: Increase button margin for better visual appearance

lassoan:
Text on buttons almost touched the border (especially in button boxes), which did not look nice. This commit reverts button margin to the default value.

Before this change (current Slicer Preview Release):

image

With this proposed change:

image


jcfr:
> I'm not sure what the overall impact of that will be, but I think the more slicer style follows the Qt defaults the better looking everything will be.

Also worth noting that the original style was developed while being based on style used almost a decade ago.

All of these stylistic changes are very exciting

PR #5028, opened 2020-07-04
ENH: Take Python console colors from style

lassoan:
Requires merging of #4993 first! 

Python console colors were stored in application settings and so the same colors were used regardless of chosen style (dark or light Slicer).
Since setting up color scheme for the Python console takes quite an effort (manual selection of 7 colors) so probably very few people did it.

To make things simpler and make Python console match the current style, with this commit we remove customization option for console colors and use colors from the style's palette.

Partial fix for #4803.

Before this change (with default settings):

image

After this change, bright style:

image

After this change, dark style:

image

### DIFF
--- a/Base/QTApp/qSlicerMainWindow.cxx
+++ b/Base/QTApp/qSlicerMainWindow.cxx
@@ -494,6 +494,8 @@ void qSlicerMainWindowPrivate::setupUi(QMainWindow * mainWindow)
       this->PythonConsoleToggleViewAction = new QAction("", this->ViewMenu);
       this->PythonConsoleToggleViewAction->setCheckable(true);
       }
+    q->pythonConsole()->setScrollBarPolicy(Qt::ScrollBarAsNeeded);
+    this->updatePythonConsolePalette();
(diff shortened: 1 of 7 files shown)


lassoan:
@jcfr @pieper @cpinter @jamesobutler please have a look at the color scheme above. Do you think we can switch to using these colors (hardcoded in the styles, so users cannot change individual colors anymore)?

jcfr:
That makes sense to me. Thanks for working on this.

jamesobutler:
Yes, getting the colors from the palette used by the application is good. Thanks for doing this! I would only suggest removing the comments like ` //black` as the color is specific to the palette so therefore the comment would only make sense specific to the light theme.

Also the blue used for `import numpy as np` appears different in your image for light vs dark. Could you update your example images with the color you currently are specifying to use?

lassoan:
Ok, I'll make the comments more clear.

The screenshots correspond to the code that I get with including your pull request that refines the colors. So, your PR must be merged first (I'll update the description with this info).

lassoan:
It is now included in https://github.com/Slicer/Slicer/pull/4993, so closing this PR.

Issue #4966, opened 2020-06-08
Save all configuration settings locally


lassoan:
Currently, Slicer code and settings are installed in 2 locations:
- Slicer core and additionally installed Python packages in: c:\Users\andra\AppData\Local\NA-MIC
- Slicer*.ini files and extensions in c:\Users\(username)\AppData\Roaming\NA-MIC\

It would be better if all these would be saved in a single location (in the Slicer core installation folder):
- it would make Slicer fully portable: could run from a USB stick, or Slicer could be used as a pre-configured viewer on a DICOM CD
- in shared installation environments, all users would work with the same preconfigured settings and set of extensions and Python packages

jcfr:
Makes sense.

Ideally, functions [slicerUserSettingsFilePath](https://github.com/Slicer/Slicer/blob/276c72e150f62d719f75d1606a47f959557454b2/Base/QTCore/qSlicerCoreApplication.cxx#L1224-L1228) and [slicerRevisionUserSettingsFilePath](https://github.com/Slicer/Slicer/blob/276c72e150f62d719f75d1606a47f959557454b2/Base/QTCore/qSlicerCoreApplication.cxx#L1230-L1252) should return path based on build settings [Slicer_CPACK_NSIS_INSTALL_REQUIRES_ADMIN_ACCOUNT](https://github.com/Slicer/Slicer/blob/47deb76d7556e40de4e25e585c4b24a63a153da5/CMake/SlicerApplicationOptions.cmake#L123-L146)

lassoan:
> Ideally, functions [slicerUserSettingsFilePath](https://github.com/Slicer/Slicer/blob/276c72e150f62d719f75d1606a47f959557454b2/Base/QTCore/qSlicerCoreApplication.cxx#L1224-L1228) and [slicerRevisionUserSettingsFilePath](https://github.com/Slicer/Slicer/blob/276c72e150f62d719f75d1606a47f959557454b2/Base/QTCore/qSlicerCoreApplication.cxx#L1230-L1252) should return path based on build settings [Slicer_CPACK_NSIS_INSTALL_REQUIRES_ADMIN_ACCOUNT](https://github.com/Slicer/Slicer/blob/47deb76d7556e40de4e25e585c4b24a63a153da5/CMake/SlicerApplicationOptions.cmake#L123-L146)

Since Python packages are installed within the Slicer home folder, it is no longer an option to install Slicer into a protected folder (writeable only by admins).

[...]

Issue #4722, opened 2020-03-13
Use relative paths for files copied into the DICOM database


slicerbot:
_This issue was created automatically from an original [Mantis Issue](https://mantisarchive.slicer.org/view.php?id=4722). Further discussion may take place here._

PR #5029, opened 2020-07-05
ENH: Make Slicer application fully portable

lassoan:
Slicer install tree was portable but not fully self-contained, as settings and extensions were stored in the user profile folder.

Added a new option Slicer_STORE_SETTINGS_IN_HOME_DIR (enabled by default) that makes Slicer store all settings and extensions in the application home folder, within (organizationName) subfolder (as QSettings default constructor saves settings file into (organizationName) subfolder by default), making the application fully portable.

To still allow having settings that are common to all installed application versions for a user (e.g., DICOM database folder, confirmation popup suppressions, ...), the common non-revision-specific is still stored in the user profile directory by default. However, if the user places an (applicationName).ini file in the application home folder/(organizationName) then that file is used instead.

By default, (applicationName)=Slicer and (organizationName)=NA-MIC and therefore the .ini files are: (applicationHome)/NA-MIC/Slicer.ini and (applicationHome)/NA-MIC/Slicer-(revision).ini; and extensions are stored in (applicationHome)/NA-MIC/Extensions-(revision).

re #4966

Prerequisites to merge:
- [x] Merge https://github.com/commontk/AppLauncher/pull/114
- [x] Update AppLauncher
- [x] Update AppLauncher hash in Slicer

### DIFF
--- a/Base/Python/slicer/slicerqt.py
+++ b/Base/Python/slicer/slicerqt.py
@@ -101,13 +101,18 @@ def initLogging(logger):
 initLogging(logging.getLogger())
 
 def getSlicerRCFileName():
-  """Return slicer resource script file name '~/.slicerrc.py'"""
+  """Return application startup file (Slicer resource script) file name.
+  If a .slicerrc.py file is found in slicer.app.slicerHome folder then that will be used.
+  If that is not found then the path defined in SLICERRC environment variable will be used.
+  If that environment variable is not specified then .slicerrc.py in the user's home folder
+  will be used ('~/.slicerrc.py')."""
   import os
-  if 'SLICERRC' in os.environ:
-    rcfile = os.environ['SLICERRC']
-  else:
-    import os.path
-    rcfile = os.path.expanduser( '~/.slicerrc.py' )
+  rcfile = os.path.join(slicer.app.slicerHome,".slicerrc.py")
+  if not os.path.exists(rcfile):
+    if 'SLICERRC' in os.environ:
+      rcfile = os.environ['SLICERRC']
+    else:
+      rcfile = os.path.expanduser( '~/.slicerrc.py' )
   rcfile = rcfile.replace('\\','/') # make slashed consistent on Windows
   return rcfile
 
--- a/Base/QTApp/qSlicerMainWindow.cxx
+++ b/Base/QTApp/qSlicerMainWindow.cxx
@@ -639,7 +639,9 @@ QList<qSlicerIO::IOProperties> qSlicerMainWindowPrivate::readRecentlyLoadedFiles
     {
     settings.setArrayIndex(i);
     QVariant file = settings.value("file");
-    fileProperties << file.toMap();
+    qSlicerIO::IOProperties properties = file.toMap();
+    properties["fileName"] = qSlicerApplication::application()->toSlicerHomeAbsolutePath(properties["fileName"].toString());
+    fileProperties << properties;
     }
   settings.endArray();
 
@@ -654,7 +656,9 @@ void qSlicerMainWindowPrivate::writeRecentlyLoadedFiles(const QList<qSlicerIO::I
   for (int i = 0; i < fileProperties.size(); ++i)
     {
     settings.setArrayIndex(i);
-    settings.setValue("file", fileProperties.at(i));
+    qSlicerIO::IOProperties properties = fileProperties.at(i);
+    properties["fileName"] = qSlicerApplication::application()->toSlicerHomeRelativePath(properties["fileName"].toString());
+    settings.setValue("file", properties);
     }
   settings.endArray();
 }
--- a/Base/QTCore/qSlicerCoreApplication.cxx
+++ b/Base/QTCore/qSlicerCoreApplication.cxx
@@ -46,6 +46,7 @@
 //  - Slicer_ORGANIZATION_DOMAIN
 //  - Slicer_ORGANIZATION_NAME
 //  - SLICER_REVISION_SPECIFIC_USER_SETTINGS_FILEBASENAME
+//  - Slicer_STORE_SETTINGS_IN_APPLICATION_HOME_DIR
 #include "vtkSlicerConfigure.h"
 
 #ifdef Slicer_USE_PYTHONQT
@@ -91,6 +92,9 @@
 # include <vtkMRMLCommandLineModuleNode.h>
 #endif
 #include <vtkMRMLScene.h>
+
+// CTK includes
+#include <ctkUtils.h>
 
 // CTKLauncherLib includes
 #include <ctkAppLauncherEnvironment.h>
@@ -472,6 +476,18 @@ QSettings* qSlicerCoreApplicationPrivate::instantiateSettings(bool useTmp)
     {
     q->setApplicationName(q->applicationName() + "-tmp");
     }
+#ifdef Slicer_STORE_SETTINGS_IN_APPLICATION_HOME_DIR
+  // If a Slicer.ini file is available in the home directory then use that,
+  // otherwise use the default one in the user profile folder.
+  // Qt appends organizationName to the directory set in QSettings::setPath, therefore we must include it in the folder name
+  // (otherwise QSettings() would return a different setting than app->userSettings()).
+  QString iniFileName = QDir(this->SlicerHome).filePath(QString("%1/%2.ini").arg(q->organizationName()).arg(q->applicationName()));
+  if (QFile(iniFileName).exists())
+    {
+    // Use settings file in the home folder
+    QSettings::setPath(QSettings::IniFormat, QSettings::UserScope, this->SlicerHome);
+    }
+#endif
   QSettings* settings = this->newSettings();
   if (useTmp && !q->coreCommandOptions()->keepTemporarySettings())
     {
@@ -1133,8 +1149,8 @@ QString qSlicerCoreApplication::defaultScenePath() const
 {
   QSettings* appSettings = this->userSettings();
   Q_ASSERT(appSettings);
-  QString defaultScenePath = appSettings->value(
-        "DefaultScenePath", QStandardPaths::writableLocation(QStandardPaths::DocumentsLocation)).toString();
+  QString defaultScenePath = this->toSlicerHomeAbsolutePath(appSettings->value(
+        "DefaultScenePath", QStandardPaths::writableLocation(QStandardPaths::DocumentsLocation)).toString());
 
   return defaultScenePath;
 }
@@ -1148,7 +1164,7 @@ void qSlicerCoreApplication::setDefaultScenePath(const QString& path)
     }
   QSettings* appSettings = this->userSettings();
   Q_ASSERT(appSettings);
-  appSettings->setValue("DefaultScenePath", path);
+  appSettings->setValue("DefaultScenePath", this->toSlicerHomeRelativePath(path));
 }
 
 //-----------------------------------------------------------------------------
@@ -1189,7 +1205,8 @@ QString qSlicerCoreApplication::temporaryPath() const
   Q_D(const qSlicerCoreApplication);
   QSettings* appSettings = this->userSettings();
   Q_ASSERT(appSettings);
-  QString temporaryPath = appSettings->value("TemporaryPath", this->defaultTemporaryPath()).toString();
+  QString temporaryPath = qSlicerCoreApplication::application()->toSlicerHomeAbsolutePath(
+    appSettings->value("TemporaryPath", this->defaultTemporaryPath()).toString());
   d->createDirectory(temporaryPath, "temporary"); // Make sure the path exists
   return temporaryPath;
 }
@@ -1251,8 +1268,16 @@ QString qSlicerCoreApplication::slicerUserSettingsFilePath()const
 //-----------------------------------------------------------------------------
 QString qSlicerCoreApplication::slicerRevisionUserSettingsFilePath()const
 {
+#ifdef Slicer_STORE_SETTINGS_IN_APPLICATION_HOME_DIR
+  this->userSettings(); // ensure applicationName is initialized
+  QString filePath = QString("%1/%2").arg(this->slicerHome()).arg(this->organizationName());
+  QString prefix = this->applicationName();
+#else
   QFileInfo fileInfo = QFileInfo(this->userSettings()->fileName());
+  QString filePath = fileInfo.path();
   QString prefix = fileInfo.completeBaseName();
+#endif
+
   QString suffix = "-" + this->revision();
   bool useTmp = this->coreCommandOptions()->settingsDisabled();
   if (useTmp)
@@ -1261,7 +1286,7 @@ QString qSlicerCoreApplication::slicerRevisionUserSettingsFilePath()const
     useTmp = true;
     }
   QString fileName =
-      QDir(fileInfo.path()).filePath(QString("%1%2%3.ini")
+      QDir(filePath).filePath(QString("%1%2%3.ini")
                                      .arg(prefix)
                                      .arg(SLICER_REVISION_SPECIFIC_USER_SETTINGS_FILEBASENAME)
                                      .arg(suffix));
@@ -1277,19 +1302,16 @@ void qSlicerCoreApplication::setTemporaryPath(const QString& path)
 {
   QSettings* appSettings = this->userSettings();
   Q_ASSERT(appSettings);
-  appSettings->setValue("TemporaryPath", path);
+  appSettings->setValue("TemporaryPath", this->toSlicerHomeRelativePath(path));
   this->applicationLogic()->SetTemporaryPath(path.toUtf8());
 }
 
 //-----------------------------------------------------------------------------
 QString qSlicerCoreApplication::defaultExtensionsInstallPath() const
 {
-  QSettings* appSettings = this->userSettings();
-  Q_ASSERT(appSettings);
 #ifdef Slicer_BUILD_EXTENSIONMANAGER_SUPPORT
-  return QFileInfo(appSettings->fileName()).dir().filePath(Slicer_EXTENSIONS_DIRNAME);
+  return QFileInfo(this->slicerRevisionUserSettingsFilePath()).dir().filePath(Slicer_EXTENSIONS_DIRNAME);
 #else
-  Q_UNUSED(appSettings);
   return QString();
 #endif
 }
@@ -1298,7 +1320,8 @@ QString qSlicerCoreApplication::defaultExtensionsInstallPath() const
 QString qSlicerCoreApplication::extensionsInstallPath() const
 {
   QSettings settings(this->slicerRevisionUserSettingsFilePath(), QSettings::IniFormat);
-  return settings.value("Extensions/InstallPath", this->defaultExtensionsInstallPath()).toString();
+  return qSlicerCoreApplication::application()->toSlicerHomeAbsolutePath(
+    settings.value("Extensions/InstallPath", this->defaultExtensionsInstallPath()).toString());
 }
 
 //-----------------------------------------------------------------------------
@@ -1308,7 +1331,8 @@ void qSlicerCoreApplication::setExtensionsInstallPath(const QString& path)
     {
     return;
     }
-  this->revisionUserSettings()->setValue("Extensions/InstallPath", path);
+  this->revisionUserSettings()->setValue("Extensions/InstallPath",
+    qSlicerCoreApplication::application()->toSlicerHomeRelativePath(path));
 #ifdef Slicer_BUILD_EXTENSIONMANAGER_SUPPORT
   Q_ASSERT(this->extensionsManagerModel());
   this->extensionsManagerModel()->updateModel();
@@ -1892,4 +1916,43 @@ void qSlicerCoreApplication::showConsoleMessage(QString message, bool error/*=tr
   {
     std::cout << message.toLocal8Bit().constData() << std::endl;
   }
+}
+
+// --------------------------------------------------------------------------
+QString qSlicerCoreApplication::toSlicerHomeAbsolutePath(const QString& path) const
+{
+  Q_D(const qSlicerCoreApplication);
+  return ctk::absolutePathFromInternal(path, d->SlicerHome);
+}
+
+// --------------------------------------------------------------------------
+QString qSlicerCoreApplication::toSlicerHomeRelativePath(const QString& path) const
+{
+  Q_D(const qSlicerCoreApplication);
+  return ctk::internalPathFromAbsolute(path, d->SlicerHome);
+}
+
+// --------------------------------------------------------------------------
+QStringList qSlicerCoreApplication::toSlicerHomeAbsolutePaths(const QStringList& paths) const
+{
+  Q_D(const qSlicerCoreApplication);
+  QStringList absolutePaths;
+  foreach(QString path, paths)
+    {
+    absolutePaths << this->toSlicerHomeAbsolutePath(path);
+    }
+  return absolutePaths;
+}
+
+
+// --------------------------------------------------------------------------
+QStringList qSlicerCoreApplication::toSlicerHomeRelativePaths(const QStringList& paths) const
+{
+  Q_D(const qSlicerCoreApplication);
+  QStringList relativePaths;
+  foreach(QString path, paths)
+    {
+    relativePaths << this->toSlicerHomeRelativePath(path);
+    }
+  return relativePaths;
 }
--- a/Base/QTCore/qSlicerCoreApplication.h
+++ b/Base/QTCore/qSlicerCoreApplication.h
@@ -471,6 +471,22 @@ public:
   /// If error is true then the message is printed on stderr, otherwise on stdout.
   Q_INVOKABLE void showConsoleMessage(QString message, bool error=true) const;
 
+  /// Converts relative path to absolute path using slicerHome directory.
+  /// Returns absolute path unchanged.
+  Q_INVOKABLE QString toSlicerHomeAbsolutePath(const QString& path) const;
+
+  /// Converts paths within slicerHome directory to relative paths.
+  /// Leaves other paths unchanged.
+  Q_INVOKABLE QString toSlicerHomeRelativePath(const QString& path) const;
+
+  /// Converts relative path to absolute path using slicerHome directory.
+  /// Returns absolute path unchanged.
+  Q_INVOKABLE QStringList toSlicerHomeAbsolutePaths(const QStringList& path) const;
+
+  /// Converts paths within slicerHome directory to relative paths.
+  /// Leaves other paths unchanged.
+  Q_INVOKABLE QStringList toSlicerHomeRelativePaths(const QStringList& path) const;
+
 public slots:
 
   /// Restart the application with the arguments passed at startup time
--- a/Base/QTCore/qSlicerExtensionsManagerModel.cxx
+++ b/Base/QTCore/qSlicerExtensionsManagerModel.cxx
@@ -43,6 +43,7 @@
 #include <qRestResult.h>
 
 // QtCore includes
+#include "qSlicerCoreApplication.h"
 #include "qSlicerExtensionsManagerModel.h"
 #include "vtkSlicerConfigure.h"
 #include "vtkSlicerVersionConfigure.h"
@@ -492,9 +493,9 @@ void qSlicerExtensionsManagerModelPrivate::addExtensionPathToApplicationSettings
 {
   Q_Q(qSlicerExtensionsManagerModel);
   QSettings settings(q->extensionsSettingsFilePath(), QSettings::IniFormat);
-  QStringList additionalPaths = settings.value("Modules/AdditionalPaths").toStringList();
+  QStringList additionalPaths = qSlicerCoreApplication::application()->toSlicerHomeAbsolutePaths(settings.value("Modules/AdditionalPaths").toStringList());
   settings.setValue("Modules/AdditionalPaths",
-                    appendToPathList(additionalPaths, q->extensionModulePaths(extensionName)));
+    qSlicerCoreApplication::application()->toSlicerHomeRelativePaths(appendToPathList(additionalPaths, q->extensionModulePaths(extensionName))));
 }
 
 // --------------------------------------------------------------------------
@@ -502,9 +503,9 @@ void qSlicerExtensionsManagerModelPrivate::removeExtensionPathFromApplicationSet
 {
   Q_Q(qSlicerExtensionsManagerModel);
   QSettings settings(q->extensionsSettingsFilePath(), QSettings::IniFormat);
-  QStringList additionalPaths = settings.value("Modules/AdditionalPaths").toStringList();
+  QStringList additionalPaths = qSlicerCoreApplication::application()->toSlicerHomeAbsolutePaths(settings.value("Modules/AdditionalPaths").toStringList());
   settings.setValue("Modules/AdditionalPaths",
-                    removeFromPathList(additionalPaths, q->extensionModulePaths(extensionName)));
+    qSlicerCoreApplication::application()->toSlicerHomeRelativePaths(removeFromPathList(additionalPaths, q->extensionModulePaths(extensionName))));
 }
 
 // --------------------------------------------------------------------------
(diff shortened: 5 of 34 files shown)


lassoan:
This would allow using Slicer as a viewer on a DICOM DVD/USB stick, the only thing would need to be fixed for this is to make DICOM database store relative paths for images that are copied into the database (currently all images and thumbnails are stored with absolute paths).

lassoan:
Unfortunately, there are a number of absolute paths in the .ini files that would cause problems (see below). So, this is not a full solution yet. @jcfr what do you think, should we use <APPLAUNCHER_DIR> placeholder in the .ini files, too?

---
Slicer.ini:
- DatabaseDirectory_0.6.3=C:/Users/andra/OneDrive/Documents/SlicerDICOMDatabase_2
- DefaultScenePath=C:/Users/andra/OneDrive/Documents
- TemporaryDirectory=C:/Users/andra/AppData/Local/Temp/Slicer
- Path=C:/Users/andra/AppData/Local/Temp/Slicer/RemoteIO

Slicer-NNN.ini:

- [LogFiles]
- AdditionalPaths=C:/D/S4D/Slicer-build/NA-MIC/Extensions-29209/BreastImplantAnalyzer/lib/Slicer-4.11/qt-scripted-modules
- InstallPath=C:/D/S4D/Slicer-build/NA-MIC/Extensions-29209
- [LibraryPaths]
- [PYTHONPATH]

cpinter:
I think the variable name `Slicer_STORE_SETTINGS_IN_HOME_DIR` is misleading. Most people think of the user's folder when they hear "home dir". It would be better to call it `Slicer_STORE_SETTINGS_IN_APPLICATION_HOME_DIR` or `Slicer_STORE_SETTINGS_IN_APP_HOME_DIR` or just `Slicer_STORE_SETTINGS_IN_APP_DIR`.

pieper:
I'd very much like to see the application be relocatable.  Running Slicer off a CD was the original motivation for having the launcher in the first place (so we could hand out CD at RSNA in like 2002 or something).

I agree with @cpinter and prefer `Slicer_STORE_SETTINGS_IN_APP_DIR` for clarity.

Making everything relative to the app path makes sense to me.  It'll also help sandbox the app for security.

lassoan:
What to do with `.slicerrc.py` file? These are the candidate locations for the file:
- application home directory
- location specified in SLICERRC environment variable
- ~/.slicerrc.py

Should we add any more locations (e.g., specify file in Slicer.ini or Slicer-NNN.ini)?
Would the search order above work (application home directory would have the highest priority)?

pieper:
I think it would be fine to look in the application directory for a .slicerrc.py and in the user's home directory.  I also like the idea of a .slicerrc-NNN.py in case you wanted to do something specific to the version.

If we did put it in the .ini file we could also include a application settings panel entry that allows you to browse to set the location and edit it with a simple text editor (and include a link to the readthedocs with information about what you can put there).  I still feel like .slicerrc.py should be used with care, since it could be error prone, but we may as well make it fairly easy to use.

jcfr:
>  @jcfr what do you think, should we use `<APPLAUNCHER_DIR>` placeholder in the .ini files, too?

Instead, we should probably use `<APPLAUNCHER_SETTINGS_DIR>`  (see [here](https://github.com/commontk/AppLauncher/blob/ef9cd0ad6e0866c0e7a94b44b5ca13528452a7be/Base/ctkAppLauncherSettings.h#L67) for description) for settings value that are common to both the launcher and slicer through the use of the AppLauncherLib (`LibraryPaths`, `PYTHONPATH`).

For regular settings only relevant to Slicer (e.g `LogFiles`, `InstallPath`,  ..,) we should introduce an other special placehoder variable called `<SLICER_SETTINGS_DIR>`. 

We would also need to provide an interface for reading/writing settings and ensure the placeholder is properly handled. This means directly using `QSettings()` method would not work out of the box.

Base/Python/slicer/slicerqt.py
```diff
@@ -101,13 +101,18 @@ def initLogging(logger):
...
+  If that is not found then the path defined in SLICERRC environment variable will be used.
+  If that environment variable is not specified then .slicerrc.py in the user's home folder
+  will be used ('~/.slicerrc.py')."""
   import os
-  if 'SLICERRC' in os.environ:
-    rcfile = os.environ['SLICERRC']
-  else:
-    import os.path
-    rcfile = os.path.expanduser( '~/.slicerrc.py' )
+  rcfile = os.path.join(slicer.app.slicerHome,".slicerrc.py")
```
jcfr, line 110: `os.path.join(slicer.app.slicerHome,".slicerrc.py")` -> `os.path.join(slicer.app.slicerHome, '.slicerrc.py')`

Base/Python/slicer/slicerqt.py
```diff
@@ -101,13 +101,18 @@ def initLogging(logger):
...
+  if not os.path.exists(rcfile):
+    if 'SLICERRC' in os.environ:
+      rcfile = os.environ['SLICERRC']
+    else:
+      rcfile = os.path.expanduser( '~/.slicerrc.py' )
```
jcfr, line 115: `rcfile = os.path.expanduser('~/.slicerrc.py')`

lassoan:
Thanks for the feedback. The implementation is now complete.

All paths that refer to to files and folders within the SlicerHome folder are now stored with relative path in settings. Extensions and Slicer-NNN.ini are stored within SlicerHome folder, therefore they are relocatable.

Base/QTCore/qSlicerCoreApplication.cxx
```diff
@@ -1872,3 +1896,85 @@ void qSlicerCoreApplication::showConsoleMessage(QString message, bool error/*=tr
...
+// --------------------------------------------------------------------------
+QString qSlicerCoreApplication::toSlicerHomeAbsolutePath(const QString& path) const
+{
+  Q_D(const qSlicerCoreApplication);
+  if (path.isEmpty())
+    {
+    return path;
+    }
+  QFileInfo fileInfo(path);
+  if (fileInfo.isRelative())
```
jcfr, line 1909: That is clever, that way there is no need for any special placeholder and it is always relative to Slicer home :+1:

Base/QTCore/qSlicerRelativePathMapper.cxx
```diff
@@ -0,0 +1,164 @@
+/*=========================================================================
+
+  Library:   CTK
```
jcfr, line 3: Update License

Base/QTCore/qSlicerRelativePathMapper.h
```diff
@@ -0,0 +1,113 @@
+/*=========================================================================
+
+  Library:   CTK
```
jcfr, line 3: Header

Base/QTCore/qSlicerCoreApplication.h
```diff
@@ -471,6 +471,22 @@ class Q_SLICER_BASE_QTCORE_EXPORT qSlicerCoreApplication : public QApplication
   /// If error is true then the message is printed on stderr, otherwise on stdout.
   Q_INVOKABLE void showConsoleMessage(QString message, bool error=true) const;
 
+  /// Converts relative path to absolute path using slicerHome directory.
+  /// Returns absolute path unchanged.
```
jcfr, line 475:
```
/// \sa toSlicerHomeRelativePath
/// \sa toSlicerHomeAbsolutePaths
```

Base/QTCore/qSlicerCoreApplication.h
```diff
@@ -471,6 +471,22 @@ class Q_SLICER_BASE_QTCORE_EXPORT qSlicerCoreApplication : public QApplication
...
+  Q_INVOKABLE QString toSlicerHomeAbsolutePath(const QString& path) const;
+
+  /// Converts paths withing slicerHome directory to relative paths.
+  /// Leaves other paths unchanged.
```
jcfr, line 479:
```
/// \sa toSlicerHomeAbsolutePath
/// \sa toSlicerHomeRelativePaths
```

Base/QTCore/qSlicerCoreApplication.h
```diff
@@ -471,6 +471,22 @@ class Q_SLICER_BASE_QTCORE_EXPORT qSlicerCoreApplication : public QApplication
...
+  Q_INVOKABLE QString toSlicerHomeRelativePath(const QString& path) const;
+
+  /// Converts relative path to absolute path using slicerHome directory.
+  /// Returns absolute path unchanged.
```
jcfr, line 483:
```
/// \sa toSlicerHomeRelativePaths
/// \sa toSlicerHomeAbsolutePath
```

Base/QTCore/qSlicerCoreApplication.h
```diff
@@ -471,6 +471,22 @@ class Q_SLICER_BASE_QTCORE_EXPORT qSlicerCoreApplication : public QApplication
...
+  Q_INVOKABLE QStringList toSlicerHomeAbsolutePaths(const QStringList& path) const;
+
+  /// Converts paths withing slicerHome directory to relative paths.
+  /// Leaves other paths unchanged.
```
jcfr, line 487:
```
/// \sa toSlicerHomeAbsolutePaths
/// \sa toSlicerHomeRelativePath
```

jcfr:
> Instead, we could make a smooth transition: deprecate usage of QSettings and remove its usage from Slicer core now;

Anticipating the change and to simplify this:

```
QString extensionInstallPath = coreApp->toSlicerHomeAbsolutePath(
    coreApp->revisionUserSettings()->value("Extensions/InstallPath").toString());
```

should we have method like `revisionUserSettings()` returns `qSlicerSettings` (or even `ctkSettings`) object that would take care of calling `toSlicerHomeAbsolutePath/toSlicerHomeRelativePath` ? 

Last, is there any absolute path save in within CTK widgets ?

lassoan:
> should we have method like revisionUserSettings() returns qSlicerSettings (or even ctkSettings) object that would take care of calling toSlicerHomeAbsolutePath/toSlicerHomeRelativePath ?

Yes, if we use a placeholder then we could do this automatically. Unfortunately, QSettings `value()` and `setValue()` methods are not virtual. Fortunately, we may be able to intercept and rewrite information in the settings file by registering a custom file format (https://doc.qt.io/qt-5/qsettings.html#registerFormat). The file format could be the default ini file, we would just rewrite the placeholder value in QString and QStringList values. I'm looking into this now.

> Last, is there any absolute path save in within CTK widgets ?

You are right. Unfortunately, there are a few:
- Few classes in PluginFramework: I would not touch these
- ctkCommandLineParser.cpp: I don't know what it uses settings for, maybe we can leave it as is
- ctkDICOMBrowser.cpp: we'll have to fix the DICOM browser
- ctkPathLineEdit.cpp: stores history; it is not a huge issue if history always uses absolute paths, but we can add a method to specify QSettings that should be used instead of the default QSettings (which would use the custom storage format)

jcfr:
> Few classes in PluginFramework: I would not touch these

Agreed

> ctkCommandLineParser.cpp: I don't know what it uses settings for, maybe we can leave it as is

It is `false` [by default](https://github.com/commontk/CTK/blob/3ae43ea441fd759c30ed234dfa9b3c419ae1de8f/Libs/Core/ctkCommandLineParser.cpp#L236) and we are [not using](https://github.com/Slicer/Slicer/blob/47deb76d7556e40de4e25e585c4b24a63a153da5/Base/QTCore/qSlicerCoreCommandOptions.cxx) this feature in Slicer. 

Description of the feature is [here](https://github.com/commontk/CTK/blob/3ae43ea441fd759c30ed234dfa9b3c419ae1de8f/Libs/Core/ctkCommandLineParser.h#L52-L53)

```
 * <li>QSettings support. Default values for arguments can be read from
 *     a QSettings object.</li>
```

> ctkDICOMBrowser.cpp: we'll have to fix the DICOM browser

Agreed. Could be addressed as part of #4722

> ctkPathLineEdit.cpp: stores history; it is not a huge issue if history always uses absolute paths, but we can add a method to specify QSettings that should be used instead of the default QSettings (which would use the custom storage format)

That would be reasonable.

## Other approach

I am wondering if this type of settings could be updated explicitly on application shutdown or by providing our own  sync method that would call https://doc.qt.io/qt-5/qsettings.html#sync.

Rewriting  of settings value using `qSlicerCoreApplication::toSlicerHomeRelativePath` could be done by:
* (1) iterating over all settings
* or (2) only by updating settings explicitly registered within the application

lassoan:
Unfortunately, value, setValue, and sync methods are not virtual, so we could not reliably override them.

I'm now trying to add pathValue() and setPathValue() methods to ctkSettings to simplify the syntax. Instead of:

    QString extensionInstallPath = coreApp->toSlicerHomeAbsolutePath(coreApp->revisionUserSettings()->value("Extensions/InstallPath").toString());

we could then use:

    QString extensionInstallPath = coreApp->revisionUserSettings()->pathValue("Extensions/InstallPath").toString()

jamesobutler:
Isn’t QSettings something at the heart of a QApplication? I use it more so that other Slicer modules for the custom modules I create. It allows for defining user settings for various widgets at the user location and default settings for all user accounts at the system level. Having multiple users using the same application with the same default settings with the ability for user settings is important for me. I often also store and apply presets for various custom modules in this manner. Not sure what I would do if the plan is to phase out the QSettings() usage in the Slicer application. Being fully portable hasn’t been a concern at all for me.

lassoan:
Slicer has been always using more than just a simple QSettings() for storing application settings (e.g., it uses a common userSettings and revision-specific revisionUserSettings).

Since Python packages must be installed in the Slicer home directory, it cannot be shared between users anymore. Since extensions can depend on various Python packages (with potentially incompatible version requirements), extensions cannot be simply shared between multiple installations of the same Slicer version for the same user. Therefore, we need to make extensions install within the Slicer home directory, too.

Since we have everything in one folder, it makes a lot of sense to go one step further and allow the application to be portable. This opens a up a very important use case: bundling a preconfigured Slicer instance to exported data sets as a viewer. It also allows shared installation for many users in a server environment (as extensions can be all copied into the shared Slicer home directory), and also allows easy distribution preconfigured Slicer environments on training courses.

QSettings() will not go away completely, it will be just simpler and safer to use app->userSettings() and app->revisionUserSettings().

jcfr:
_Comments below are for information only and to capture some idea of future discussion_

> app->userSettings() and app->revisionUserSettings().

Moving forward, if we only support building Slicer with `Slicer_STORE_SETTINGS_IN_HOME_DIR` set to True, I am wondering if we should only have one function called `app->settings()`.

At application startup, we would check if home folder is writable or not ... and have a relocatable application or not. 

Function `qSlicerCoreApplication::isRelocatableInstall()` could even be added.

lassoan:
> I am wondering if we should only have one function called app->settings()

It is still useful to allow sharing some user preferences (dark/light mode, DICOM database folder, dialog "don't show again" settings, etc.). With the proposed changes in this pull request, it is possible to choose between a fully isolated application (by having a Slicer.ini file in the Slicer Home folder) or sharing some user preferences between application instances (by not having Slicer.ini in Slicer Home).

jamesobutler:
Ok just wanted to throw out some commercial usage of Slicer as this is pretty different from the motivation of making it easy to share the installation with data for sharing in various cases.  We still prefer the installation of Slicer/SlicerCustomApp in a system location not a user location. This is because customers have many users/researchers and they setup a local instance and use multiple Windows user accounts for access on the local machine and not a server environment. Installation is restricted to certain machines. Installing of different versions and installation of extensions not included is also restricted.

lassoan:
You can still keep the current behavior in your custom application by setting Slicer_STORE_SETTINGS_IN_APPLICATION_HOME_DIR to OFF in your custom application.

However, Slicer_STORE_SETTINGS_IN_APPLICATION_HOME_DIR actually makes shared installations safer and more predictable, too, as you can preinstall extensions for all users and you can prevent non-admin users from installing extensions.

Do you want your users to be able install extensions or Python packages?

jamesobutler:
> You can still keep the current behavior in your custom application by setting Slicer_STORE_SETTINGS_IN_APPLICATION_HOME_DIR to OFF in your custom application.

Ok sounds good 👍 

> However, Slicer_STORE_SETTINGS_IN_APPLICATION_HOME_DIR actually makes shared installations safer and more predictable, too, as you can preinstall extensions for all users and you can prevent non-admin users from installing extensions.

This could be beneficial if we want to preinstall some extension that hasn't made it into our SlicerCustomApp yet. We wouldn't support users installing/removing extensions.

> Do you want your users to be able install extensions or Python packages?

At this time, we do not want users to install extensions or python packages.  We control the experience in a very specific manner.

lassoan:
I've moved the path replacement logic to CTK (ctkCoreSettings class) - @jcfr please have a look: https://github.com/commontk/CTK/pull/922

Once this CTK prerequisite is merged, I'll finalize this PR.

jcfr:
> please have a look: commontk/CTK#922

Thanks for working on this :pray: , I just posted few comments.

Base/QTApp/qSlicerMainWindow.cxx
```diff
@@ -605,30 +605,34 @@ QList<qSlicerIO::IOProperties> qSlicerMainWindowPrivate::readRecentlyLoadedFiles
 {
   QList<qSlicerIO::IOProperties> fileProperties;
 
-  QSettings settings;
-  int size = settings.beginReadArray("RecentlyLoadedFiles/RecentFiles");
+  ctkCoreSettings* settings = qSlicerCoreApplication::application()->userSettings();
```
jcfr, line 608: I think  we can  keep it as `ctkCoreSettings settings;`

Base/QTGUI/qSlicerIOManager.cxx
```diff
@@ -142,22 +141,22 @@ void qSlicerIOManagerPrivate::readSettings()
...
+  settings->endGroup();
 }
 
 //-----------------------------------------------------------------------------
 void qSlicerIOManagerPrivate::writeSettings()
 {
   Q_Q(qSlicerIOManager);
-  QSettings settings;
-  settings.beginGroup("ioManager");
+  ctkCoreSettings* settings = qSlicerCoreApplication::application()->userSettings();
```
jcfr, line 151: `ctkCoreSettings settings;`

Base/QTApp/qSlicerMainWindow.cxx
lassoan: We would need to copy the application home directory from usersettings object and I don't think it is copied (I think copying is disabled for ctkCoreSettings due to the private member implementation class).

jcfr: To address this, may be we could also set the home directory as a dynamic property to the application instance?

Base/QTCore/qSlicerExtensionsManagerModel.cxx
```diff
@@ -481,19 +482,23 @@ QStringList removeFromPathList(const QStringList& paths, const QStringList& path
...
 // --------------------------------------------------------------------------
 void qSlicerExtensionsManagerModelPrivate::removeExtensionPathFromApplicationSettings(const QString& extensionName)
 {
   Q_Q(qSlicerExtensionsManagerModel);
-  QSettings settings(q->extensionsSettingsFilePath(), QSettings::IniFormat);
-  QStringList additionalPaths = settings.value("Modules/AdditionalPaths").toStringList();
-  settings.setValue("Modules/AdditionalPaths",
+  ctkCoreSettings settings(q->extensionsSettingsFilePath(), QSettings::IniFormat);
+  settings.setApplicationHomeDirectory(qSlicerCoreApplication::application()->slicerHome());
+  settings.setApplicationHomePlaceholder("<APPLAUNCHER_DIR>");
```
jcfr, line 499: Consider using `<APPLAUNCHER_SETTINGS_DIR>` , also the `ApplicationHomeDirectory` may have to be set to match the location of the settings file instead.

lassoan:
I've tested this on Windows and seems to work well.

@jcfr could you test in Linux?
@pieper could you test it on Mac?

Installing a larger extension package that contains loadable and CLI modules (such as SlicerRT or SlicerIGT) and testing if the modules run correctly should be enough. Slicer-(revision).ini and all installed extensions should show up in the (applicationHome)/NA-MIC folder.

Maybe you could also test if you put a Slicer.ini file into (applicationHome)/NA-MIC folder then it overrides the global ~/Slicer.ini file.

pieper:
To test did you build this branch and create packages of it and also the extensions?

lassoan:
I've built and packaged extensions myself, but probably extension packages that you download from the dashboard should work, too.

Base/QTCore/qSlicerExtensionsManagerModel.cxx
jcfr: To ensure `PythonSlicer` works, could you updated to use `APPLAUNCHER_SETTINGS_DIR` ?

jcfr:
> @jcfr could you test in Linux?

A package is being built on metroxplex

jcfr:
Build failed with the following error:

```
/CMakeFiles/qSlicerBaseQTCore.dir/qSlicerCoreApplication.cxx.o -c /work/Slicer/Base/QTCore/qSlicerCoreApplication.cxx
In file included from /work/Slicer/Base/QTCore/qSlicerCoreApplication.cxx:63:0:
/work/Slicer/Base/QTCore/qSlicerCoreApplication_p.h:41:29: fatal error: ctkCoreSettings.h: No such file or directory
compilation terminated.
[2280/5319] Building CXX object Base/QTCore/CMakeFiles/qSlicerBaseQTCore.dir/qSlicerCoreCommandOptions.cxx.o
ninja: build stopped: subcommand failed.
```

lassoan:
```
/work/Slicer/Base/QTCore/qSlicerCoreApplication_p.h:41:29: fatal error: ctkCoreSettings.h: No such file or directory
compilation terminated.
```

ctkCoreSettings.h is the new class in CTK. Probably you just need to do a top-level build to get CTK updated.

jcfr:
Should the topic be rebased or CTK updated ? 

It currently checkout this version:

https://github.com/Slicer/Slicer/blob/c3572674e2726db84628085cae4af5a690bc15eb/SuperBuild/External_CTK.cmake#L58-L62

Whereas this one should be used:

commontk/CTK@f0bb5d7822dee7214b8458c8e2cfd4ecab756c03

lassoan:
You are right, the branch has to be rebased on latest Slicer master branch to get the new CTK hash.

lassoan:
I've rebased the branch now. It should work now.

jcfr:
Build in progress, will report back once complete.

jcfr:
> Build in progress, will report back once complete.

Linux build was successful :+1: 

## Testing

Prerequisites:
* Delete `~/.config/NA-MIC` directories

Summary:
* :heavy_check_mark:  Slicer starts
* :heavy_check_mark:  Installing extension package (e.g `SlicerRT`)
* :negative_squared_cross_mark: Slilcer starts with installed extension

### Analysis

After installing `SlicerRT`, Slicer fails to load the extension (see `Error message` section below)

Specifying `--launcher-verbose` option, we can see that the `UserAdditionalSettings` is not found:

```
./Slicer --launcher-verbose 
[...]
info: OrganizationDomain [www.na-mic.org]
info: OrganizationName [NA-MIC]
info: ApplicationName [Slicer]
info: ApplicationRevision [29218]
info: SettingsFileName [/home/jcfr/Downloads/Slicer-4.11.0-2020-07-14-linux-amd64/bin/SlicerLauncherSettings.ini]
info: SettingsDir [/home/jcfr/Downloads/Slicer-4.11.0-2020-07-14-linux-amd64/bin]
info: UserAdditionalSettingsDir [/home/jcfr/.config/NA-MIC]
info: UserAdditionalSettingsFileName []
info: UserAdditionalSettingsFileBaseName []
[...]
info: <APPLAUNCHER_DIR> -> [/home/jcfr/Downloads/Slicer-4.11.0-2020-07-14-linux-amd64]
info: <APPLAUNCHER_NAME> -> [Slicer]
info: <APPLAUNCHER_SETTINGS_DIR> (RegularSettings) -> [/home/jcfr/Downloads/Slicer-4.11.0-2020-07-14-linux-amd64/bin]
info: <APPLAUNCHER_SETTINGS_DIR> (UserAdditionalSettings) -> [<APPLAUNCHER_SETTINGS_DIR-NOTFOUND>]
info: <APPLAUNCHER_SETTINGS_DIR> (AdditionalSettings) -> [<APPLAUNCHER_SETTINGS_DIR-NOTFOUND>]
[...]
```

For comparison, after installing `SlicerRT` from a preview build without this PR, the following is reported:

```
./Slicer --launcher-verbose
[...]
info: OrganizationDomain [www.na-mic.org]
info: OrganizationName [NA-MIC]
info: ApplicationName [Slicer]
info: ApplicationRevision [29154]
info: SettingsFileName [/home/jcfr/Downloads/Slicer-4.11.0-2020-06-17-linux-amd64/bin/SlicerLauncherSettings.ini]
info: SettingsDir [/home/jcfr/Downloads/Slicer-4.11.0-2020-06-17-linux-amd64/bin]
info: UserAdditionalSettingsDir [/home/jcfr/.config/NA-MIC]
info: UserAdditionalSettingsFileName []
info: UserAdditionalSettingsFileBaseName []
[...]
info: <APPLAUNCHER_SETTINGS_DIR> (RegularSettings) -> [/home/jcfr/Downloads/Slicer-4.11.0-2020-06-17-linux-amd64/bin]
info: <APPLAUNCHER_SETTINGS_DIR> (UserAdditionalSettings) -> [<APPLAUNCHER_SETTINGS_DIR-NOTFOUND>]
info: <APPLAUNCHER_SETTINGS_DIR> (AdditionalSettings) -> [<APPLAUNCHER_SETTINGS_DIR-NOTFOUND>]
```

Finally, starting Slicer specifying `./Slicer --launcher-additional-settings ./NA-MIC/Slicer-29218.ini` ensure the `SlicerRT` extension can be loaded.

### Proposed  solutions

To address this, the application launcher should be updated so that the path to the user additional settings is read from the main settings instead of being inferred from the QSettings.

See https://github.com/commontk/AppLauncher/blob/b1d6dad48b11c2770a72ad879812995213ccfc6e/Base/ctkAppLauncherSettings.cpp#L60-L64

Optionally, function `ctkAppLauncherSettingsPrivate::userAdditionalSettingsDir()` could return a path  configured into the main launcher settings file using the key `userAdditionalSettingsDir`


### Error message



```
./Slicer 
  Error(s):
    Cannot load library /home/jcfr/Downloads/Slicer-4.11.0-2020-07-14-linux-amd64/NA-MIC/Extensions-29218/SlicerRT/lib/Slicer-4.11/qt-loadable-modules/libqSlicerBeamsModule.so: (libqSlicerBeamsModuleWidgets.so: cannot open shared object file: No such file or directory)
  Error(s):
    Cannot load library /home/jcfr/Downloads/Slicer-4.11.0-2020-07-14-linux-amd64/NA-MIC/Extensions-29218/SlicerRT/lib/Slicer-4.11/qt-loadable-modules/libqSlicerDicomRtImportExportModule.so: (libvtkSlicerDicomRtImportExportModuleLogic.so: cannot open shared object file: No such file or directory)
  Error(s):
    Cannot load library /home/jcfr/Downloads/Slicer-4.11.0-2020-07-14-linux-amd64/NA-MIC/Extensions-29218/SlicerRT/lib/Slicer-4.11/qt-loadable-modules/libqSlicerDicomSroImportExportModule.so: (libvtkSlicerDicomSroImportExportModuleLogic.so: cannot open shared object file: No such file or directory)
  Error(s):
    Cannot load library /home/jcfr/Downloads/Slicer-4.11.0-2020-07-14-linux-amd64/NA-MIC/Extensions-29218/SlicerRT/lib/Slicer-4.11/qt-loadable-modules/libqSlicerDoseAccumulationModule.so: (libvtkSlicerDoseAccumulationModuleLogic.so: cannot open shared object file: No such file or directory)
  Error(s):
    Cannot load library /home/jcfr/Downloads/Slicer-4.11.0-2020-07-14-linux-amd64/NA-MIC/Extensions-29218/SlicerRT/lib/Slicer-4.11/qt-loadable-modules/libqSlicerDoseComparisonModule.so: (libqSlicerDoseComparisonSubjectHierarchyPlugins.so: cannot open shared object file: No such file or directory)
  Error(s):
    Cannot load library /home/jcfr/Downloads/Slicer-4.11.0-2020-07-14-linux-amd64/NA-MIC/Extensions-29218/SlicerRT/lib/Slicer-4.11/qt-loadable-modules/libqSlicerDoseVolumeHistogramModule.so: (libqSlicerDoseVolumeHistogramSubjectHierarchyPlugins.so: cannot open shared object file: No such file or directory)
  Error(s):
    Cannot load library /home/jcfr/Downloads/Slicer-4.11.0-2020-07-14-linux-amd64/NA-MIC/Extensions-29218/SlicerRT/lib/Slicer-4.11/qt-loadable-modules/libqSlicerDosxyzNrc3dDoseFileReaderModule.so: (libvtkSlicerRtCommon.so: cannot open shared object file: No such file or directory)
  Error(s):
    Cannot load library /home/jcfr/Downloads/Slicer-4.11.0-2020-07-14-linux-amd64/NA-MIC/Extensions-29218/SlicerRT/lib/Slicer-4.11/qt-loadable-modules/libqSlicerExternalBeamPlanningModule.so: (libqSlicerBeamsSubjectHierarchyPlugins.so: cannot open shared object file: No such file or directory)
  Error(s):
    Cannot load library /home/jcfr/Downloads/Slicer-4.11.0-2020-07-14-linux-amd64/NA-MIC/Extensions-29218/SlicerRT/lib/Slicer-4.11/qt-loadable-modules/libqSlicerIsodoseModule.so: (libvtkSlicerIsodoseModuleWidgets.so: cannot open shared object file: No such file or directory)
  Error(s):
    Cannot load library /home/jcfr/Downloads/Slicer-4.11.0-2020-07-14-linux-amd64/NA-MIC/Extensions-29218/SlicerRT/lib/Slicer-4.11/qt-loadable-modules/libqSlicerPlanarImageModule.so: (libvtkSlicerPlanarImageModuleLogic.so: cannot open shared object file: No such file or directory)
  Error(s):
    Cannot load library /home/jcfr/Downloads/Slicer-4.11.0-2020-07-14-linux-amd64/NA-MIC/Extensions-29218/SlicerRT/lib/Slicer-4.11/qt-loadable-modules/libqSlicerPlastimatchPyModule.so: (libvtkSlicerPlastimatchPyModuleLogic.so: cannot open shared object file: No such file or directory)
  Error(s):
    Cannot load library /home/jcfr/Downloads/Slicer-4.11.0-2020-07-14-linux-amd64/NA-MIC/Extensions-29218/SlicerRT/lib/Slicer-4.11/qt-loadable-modules/libqSlicerPlmProtonDoseEngineModule.so: (libvtkSlicerPlmProtonDoseEngineModuleLogic.so: cannot open shared object file: No such file or directory)
  Error(s):
    Cannot load library /home/jcfr/Downloads/Slicer-4.11.0-2020-07-14-linux-amd64/NA-MIC/Extensions-29218/SlicerRT/lib/Slicer-4.11/qt-loadable-modules/libqSlicerRoomsEyeViewModule.so: (libvtkSlicerRoomsEyeViewModuleLogic.so: cannot open shared object file: No such file or directory)
  Error(s):
    Cannot load library /home/jcfr/Downloads/Slicer-4.11.0-2020-07-14-linux-amd64/NA-MIC/Extensions-29218/SlicerRT/lib/Slicer-4.11/qt-loadable-modules/libqSlicerSegmentComparisonModule.so: (libvtkSlicerSegmentComparisonModuleLogic.so: cannot open shared object file: No such file or directory)
  Error(s):
    Cannot load library /home/jcfr/Downloads/Slicer-4.11.0-2020-07-14-linux-amd64/NA-MIC/Extensions-29218/SlicerRT/lib/Slicer-4.11/qt-loadable-modules/libqSlicerSegmentMorphologyModule.so: (libvtkSlicerSegmentMorphologyModuleLogic.so: cannot open shared object file: No such file or directory)
  Error(s):
    Cannot load library /home/jcfr/Downloads/Slicer-4.11.0-2020-07-14-linux-amd64/NA-MIC/Extensions-29218/SlicerRT/lib/Slicer-4.11/qt-loadable-modules/libqSlicerVffFileReaderModule.so: (libvtkSlicerVffFileReaderLogic.so: cannot open shared object file: No such file or directory)
When loading module  "BatchStructureSetConversion" , the dependency "DicomRtImportExport" failed to be loaded.
libvtkSlicerRoomsEyeViewModuleLogicPythonD.so: cannot open shared object file: No such file or directory
libvtkSlicerExternalBeamPlanningModuleLogicPythonD.so: cannot open shared object file: No such file or directory
libvtkSlicerDicomRtImportExportModuleLogicPythonD.so: cannot open shared object file: No such file or directory
libvtkSlicerDoseAccumulationModuleLogicPythonD.so: cannot open shared object file: No such file or directory
libvtkSlicerBeamsModuleLogicPythonD.so: cannot open shared object file: No such file or directory
libvtkSlicerDoseVolumeHistogramModuleLogicPythonD.so: cannot open shared object file: No such file or directory
libvtkSlicerSegmentComparisonModuleLogicPythonD.so: cannot open shared object file: No such file or directory
libvtkSlicerPlastimatchPyModuleLogicPythonD.so: cannot open shared object file: No such file or directory
libvtkSlicerPlanarImageModuleLogicPythonD.so: cannot open shared object file: No such file or directory
libvtkSlicerDoseComparisonModuleLogicPythonD.so: cannot open shared object file: No such file or directory
libvtkSlicerDicomSroImportExportModuleLogicPythonD.so: cannot open shared object file: No such file or directory
libvtkSlicerIsodoseModuleLogicPythonD.so: cannot open shared object file: No such file or directory
libvtkSlicerPlmProtonDoseEngineModuleLogicPythonD.so: cannot open shared object file: No such file or directory
libvtkSlicerSegmentMorphologyModuleLogicPythonD.so: cannot open shared object file: No such file or directory
libvtkSlicerBeamsModuleMRMLPythonD.so: cannot open shared object file: No such file or directory
libvtkSlicerDoseVolumeHistogramModuleMRMLPythonD.so: cannot open shared object file: No such file or directory
libqSlicerDoseVolumeHistogramSubjectHierarchyPlugins.so: cannot open shared object file: No such file or directory
libqSlicerDoseComparisonSubjectHierarchyPlugins.so: cannot open shared object file: No such file or directory
libqSlicerDicomRtImportExportSubjectHierarchyPlugins.so: cannot open shared object file: No such file or directory
libqSlicerBeamsModuleWidgets.so: cannot open shared object file: No such file or directory
libqSlicerPlmProtonDoseEngineDoseEngines.so: cannot open shared object file: No such file or directory
libqSlicerBeamsSubjectHierarchyPlugins.so: cannot open shared object file: No such file or directory
libqSlicerExternalBeamPlanningModuleWidgets.so: cannot open shared object file: No such file or directory
libqSlicerIsodoseSubjectHierarchyPlugins.so: cannot open shared object file: No such file or directory
When loading module  "DvhComparison" , the dependency "DoseVolumeHistogram" failed to be loaded.
When loading module  "IGRTWorkflow_SelfTest" , the dependency "DicomRtImportExport" failed to be loaded.
```

lassoan:
Thanks a lot for the thorough testing and analysis! 

We would still want to allow sharing a single Slicer.ini between all Slicer installs (which don't have a local Slicer.ini). So, the Slicer-NNN path cannot be stored in Slicer.ini. 


Slicer looks for Slicer.ini in <applauncher_dir>/organization and if it is not found there then it looks for it in userprofile/organization folder. Currently, Slicer-NNN.ini is only looked for in <applauncher_dir>/organization or userprofile/organization folder, depending on CMake option chosen during Slicer build.

Launcher currently looks for these files only in userprofile/organization.

We should resolve these inconsistencies. 

I think the best behavior would be to look for both ini files in <applauncher_dir>/organization folder first and if it is not found there then in userprofile/organization folder. What do you think?

jcfr:
> We would still want to allow sharing a single Slicer.ini between all Slicer installs (which don't have a local Slicer.ini). So, the Slicer-NNN path cannot be stored in Slicer.ini.

Agreed. 

> [...] What do you think?

To clarify, the idea is to set `userAdditionalSettingsDir` in `SlicerLauncerSettings.ini` only if `Slicer_STORE_SETTINGS_IN_HOME_DIR` is `TRUE`

That way, the launcher will lookup the `Slicer-NNN.ini` in the expected location and this will nicely address the problem.

jcfr:
@lassoan I am currently building and testing this PR, did you squash and rebase or did you push any other changes ?

lassoan:
I pushed a change in qSlicerExtensionsManagerModelTest (fixes this test, no change in actual application code).

PR #5030, opened 2020-07-06
ENH: Improve performance for markup nodes

Sunderlandkyl:
Removed unnecessary modified events caused by setting the SurfaceCostFunctionType or SurfaceDistanceWeightingFunction to the same value.
Improved speed when calling SetControlPointPositionsWorld by pausing the calculation of measurements until all of the points have been set.

### DIFF
--- a/Modules/Loadable/Markups/MRML/vtkCurveGenerator.cxx
+++ b/Modules/Loadable/Markups/MRML/vtkCurveGenerator.cxx
@@ -1015,9 +1015,13 @@ int vtkCurveGenerator::GetSurfaceCostFunctionType()
 }
 
 //------------------------------------------------------------------------------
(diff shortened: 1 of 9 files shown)


lassoan:
If we may want to override UpdateMeasurementsInternal in derived classes then it is a bit more clear if it is a simple protected member (and not private implementation in vtkInternal).

jcfr:
Thinking more about it, current proposal  makes more sense. As it allows to external  call `UpdateMeasurements` 



--------------

For reference, here is my now obsolete explanation ... 

The virtual method would always be in the public class and would keep the name `UpdateMeasurements`.
That would have the advantage of keeping the surface API simpler.
The one in `vtkInternal` would be like:

```
void vtkMRMLMarkupsNode::vtkInternal::UpdateMeasurements()
{
  if (this->IsUpdatingPoints)
    {
    return;
    }

  this->External->UpdateMeasurements();
}
```

PR #5034, opened 2020-07-08
BUG: Fix invalid oversampling in segmentation geometry dialog

### LINKED ISSUE #5032: Specify geometry button in Segment Editor applies oversampling factor twice
When user wants to reduce spacing by a factor of 2, it actually gets reduced by a factor of 4.

## Steps to reproduce

- Load MRHead
- Create segmentation, new segment, and paint something
- Click on Specify geometry button, choose current segmentation as reference, and oversampling factor 2.0
- Go to data module, select segmentation node, open MRML node information section, Node information subsection
- `Reference image geometry = 0;0;0.6499977111816406;-86.9698963165283;-0.5;0;0;134.17860412597656;0;-0.5;0;117.03569793701172;0;0;0;1;200;341;254;255;102;211;` => correct
- Segment_1 binary labelmap representation has: `Spacing: 0.25 0.25 0.324999` => Incorrect

## Expected behavior

- Segment_1 binary labelmap representation should have: `Spacing: 0.5 0.5 0.65`

## Environment
- Slicer version: Slicer-4.11.0-2020-07-04

Sunderlandkyl:
When oversampling a segmentation, the oversampling factor would be doubled if the same segmentation node was set as the reference geometry.
While updating the segmentation node, modification events would cause widget to recalculate the output geometry. The specific action that caused the update was setting the segmentation node ReferenceImageGeometry reference.

Fixed by blocking modified events from the segmentation node until the resampling has been applied, and by not changing the ReferenceImageGeometry reference if the same segmentation node is being used as the reference geometry node.

Also fixes memory leak in vtkSlicerSegmentationGeometryLogic.

fixes #5032

### DIFF
--- a/Modules/Loadable/Segmentations/Logic/vtkSlicerSegmentationGeometryLogic.cxx
+++ b/Modules/Loadable/Segmentations/Logic/vtkSlicerSegmentationGeometryLogic.cxx
@@ -64,7 +64,7 @@ vtkSlicerSegmentationGeometryLogic::vtkSlicerSegmentationGeometryLogic()
 vtkSlicerSegmentationGeometryLogic::~vtkSlicerSegmentationGeometryLogic()
 {
   this->SetInputSegmentationNode(nullptr);
-
+  this->SetSourceGeometryNode(nullptr);
   if (this->OutputGeometryImageData)
     {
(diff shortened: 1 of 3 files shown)


pieper:
👍

PR #5039, opened 2020-07-10
ENH: Add different angle measurement modes

cpinter:
Angle modes allow calculating angles in different ways:
- Unsigned: Minimal angle, between [0..180)
- Oriented signed: Oriented angle [-180..180), using rotation axis in right-hand rule
- Oriented positive: Oriented angle [0..360), using rotation axis in right-hand rule

### DIFF
--- a/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsAngleNode.cxx
+++ b/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsAngleNode.cxx
@@ -33,6 +33,8 @@ vtkMRMLNodeNewMacro(vtkMRMLMarkupsAngleNode);
 
 //----------------------------------------------------------------------------
 vtkMRMLMarkupsAngleNode::vtkMRMLMarkupsAngleNode()
+  : AngleMode(Unsigned)
+  , OrientationRotationAxis{0.0, 0.0, 1.0}
 {
   this->MaximumNumberOfControlPoints = 3;
   this->RequiredNumberOfControlPoints = 3;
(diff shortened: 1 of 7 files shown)


cpinter:
@lassoan I did not add the GUI for the rotation axis unit vector in the Markups module. I think the users will not want to use it, as the modes work well most of the cases. In the cases where the vector needs to be given accurately, it is calculated automatically and need to be set programmatically.

Another thing that may need discussion is the radius of the arc. In the case of long arc (>180) I cut the radius in half because it looked really strange. So there is a jump in radius after 180, but I don't see it strange or annoying. If you do let me know.

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsAngleNode.h
```diff
@@ -43,6 +43,14 @@ class  VTK_SLICER_MARKUPS_MODULE_MRML_EXPORT vtkMRMLMarkupsAngleNode : public vt
 
   const char* GetIcon() override {return ":/Icons/MarkupsAngleMouseModePlace.png";}
 
+  enum AngleModes
```
jcfr, line 46: Could you document the enum ?   Adding something like this would work

```
\sa SetAngleModeToUnsigned()
\sa SetAngleModeToOrientedSigned()
\sa SetAngleModeToOrientedPositive()
```

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsAngleNode.cxx
```diff
@@ -33,6 +33,8 @@ vtkMRMLNodeNewMacro(vtkMRMLMarkupsAngleNode);
 
 //----------------------------------------------------------------------------
 vtkMRMLMarkupsAngleNode::vtkMRMLMarkupsAngleNode()
+  : AngleMode(0)
```
jcfr, line 36: I am also wondering if `AngleMode` should be changed to `AngleMeasurementMode`

lassoan, line 36: use enum instead of 0

lassoan: AngleMode is shorter than AngleMeasurementMode, and it is not strictly just measurement but display, too (angle is displayed slightly differently depending on the mode). However, if you don't find it obvious what "AngleMode" refers to then I agree that a longer, more specific name could be better.

@pieper would "AngleMode" self-explaining enough for you or do you think "AngleMeasurementMode" would be more clear?

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsAngleNode.cxx
```diff
@@ -45,18 +47,179 @@ vtkMRMLMarkupsAngleNode::~vtkMRMLMarkupsAngleNode() = default;
 void vtkMRMLMarkupsAngleNode::WriteXML(ostream& of, int nIndent)
 {
   Superclass::WriteXML(of,nIndent);
+
+  vtkMRMLWriteXMLBeginMacro(of);
+  vtkMRMLWriteXMLEnumMacro(angleMode, AngleMode);
+  vtkMRMLWriteXMLVectorMacro(surfaceDistanceWeightingFunction, OrientationRotationAxis, double, 3);
```
lassoan, line 53: surfaceDistanceWieghtingFunction copy-paste error at several places

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsAngleNode.cxx
```diff
@@ -45,18 +47,179 @@ vtkMRMLMarkupsAngleNode::~vtkMRMLMarkupsAngleNode() = default;
...
+  vtkMRMLPrintBeginMacro(os, indent);
+  vtkMRMLPrintEnumMacro(AngleMode);
+  vtkMRMLPrintVectorMacro(OrientationRotationAxis, double, 3);
+  vtkMRMLPrintEndMacro();
+}
+
+//----------------------------------------------------------------------------
+int vtkMRMLMarkupsAngleNode::GetAngleMode()
+{
+  return this->AngleMode;
```
lassoan, line 94: Modified() call is missing. Also, if you just need to get/set value then use the standard get/set macro instead.

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsAngleNode.cxx
```diff
@@ -45,18 +47,179 @@ vtkMRMLMarkupsAngleNode::~vtkMRMLMarkupsAngleNode() = default;
...
+  double angle_Rad = acos(vtkMath::Dot(vector1, vector2));
+  double angle_Deg = vtkMath::DegreesFromRadians(angle_Rad);
+  if (this->AngleMode == Unsigned)
+    {
+    return angle_Deg;
+    }
+  else
+    {
+    double vector1_vector2_Cross[3] = {0.0, 0.0, 0.0};
+    vtkMath::Cross(vector1, vector2, vector1_vector2_Cross);
```
lassoan, line 203: Good, this is nice and simple now.

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsAngleNode.h
```diff
@@ -59,7 +67,47 @@ class  VTK_SLICER_MARKUPS_MODULE_MRML_EXPORT vtkMRMLMarkupsAngleNode : public vt
 
   /// Copy node content (excludes basic data, such as name and node references).
   /// \sa vtkMRMLNode::CopyContent
-  vtkMRMLCopyContentDefaultMacro(vtkMRMLMarkupsAngleNode);
+  vtkMRMLCopyContentMacro(vtkMRMLMarkupsAngleNode);
+
+  /// Get angle calculation mode, determining what angle to calculate from the given points.
+  /// \return \sa AngleMode (0:Unsigned, 1:OrientedSigned, 2:OrientedPositive)
```
lassoan, line 73: Enum values (0, 1, 2) are not part of the public API, they are internal implementation details that we should be able to change in the future without breaking backward compatibility. So, we should not include them in the API documentation.

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsAngleNode.h
```diff
@@ -71,6 +119,19 @@ class  VTK_SLICER_MARKUPS_MODULE_MRML_EXPORT vtkMRMLMarkupsAngleNode : public vt
...
+
+protected:
+  /// Rotation axis for calculating right-hand rule angles.
+  /// Right-hand rule: angle is positive if p2->p1 is rotated to p2->p3 by counter-clockwise rotation
+  /// Applies if \sa AngleMode is oriented (OrientedSigned or OrientedPositive).
+  double OrientationRotationAxis[3];
+
+  /// Angle calculation mode, determining what angle to calculate from the given points.
+  /// Available options:
+  /// 0: Unsigned: Minimal angle, between [0..180)
```
lassoan, line 131: remove 0, 1, 2

Modules/Loadable/Markups/Resources/UI/qSlicerMarkupsModule.ui
```diff
@@ -1070,6 +1070,47 @@
...
+     <layout class="QGridLayout" name="gridLayout_2">
+      <item row="0" column="0">
+       <widget class="QLabel" name="label_8">
+        <property name="text">
+         <string>Angle mode:</string>
+        </property>
+       </widget>
+      </item>
+      <item row="0" column="1">
+       <widget class="QComboBox" name="angleModeComboBox">
```
lassoan, line 1090: If we add angle mode to the GUI then we must allow the user to set the rotation axis, too. You can use a ctkMatrixWidget.

cpinter:
I addressed all comments. I haven't changed the name of the attribute though, because there doesn't seem to be an agreement about it yet.

lassoan:
After some small fixes it is integrated in https://github.com/Slicer/Slicer/commit/2b3e7af4dc6a4733f1f55ad139bd68daa1ed34ac

Issue #4968, opened 2020-06-09
Report detailed error message about why node save failed


lassoan:
Save data dialog may show several popups when nodes cannot be saved. The popups only tell that saving failed, but not why or what the user can do about it.

We should show a single popup and/or show comments (or error icons) in the save data dialog table. Error message should be more specific.

See for example these related issues:
https://mantisarchive.slicer.org/view.php?id=4714
https://github.com/Slicer/SlicerGitSVNArchive/commit/d596340895ed6e54bc164a6715b3b030654129ee

lassoan:
It should not be hard to improve this:
1. we could easily replace repetitive popups with a single popup at the end; adding a new column for status text or icon would not be hard either
2. we could store a LastErrorDescription member variable in MRML storage nodes, which could contain a more specific, human-readable error message

pieper:
Yes, and for loading too.

Leengit:
For those who weren't at the 6/9/20 hangout, the idea is to modify storage nodes to collect warnings and errors and then to present them nicely in a summary dialog box for the user ... not just outputting developer-style text to the console.  I'd like to sanity check these possibilities:

1. We could replace every invocation of the vtkWarningMacro in every Slicer/Libs/MRML/Core/vtkMRML*StorageNode.cxx file with a new macro storageNodeWarningMacro.  The new macro would be in charge of (1) logging the warning in a user-friendly way, perhaps via a new (overridable) vtkMRMLStorageNode::logWarningMessage method, and (2) passing the arguments to vtkWarningMacro for traditional handling.  At opportune places near the end of some vtkMRML*StorageNode methods we'd call a new vtkMRMLStorageNode::showWarningMessage method, to summarize the collected information into a single user-friendly dialog box.  (And likewise with "Warning" replaced with "Error" in all of the above.)
2.     I see that vtkWarningWithObjectMacro (which is invoked by vtkWarningMacro) checks whether the object->HasObserver("WarningEvent") and, if so, invokes the event.  Instead of idea #1, could I add some sort of an observer that collects and creates a user-friendly version of the messages?  Although this collects all the messages, I don't know how a post-processing that displays the summary in a single dialog box would get invoked -- perhaps much like idea #1, we could somehow invoke the observer in some distinct way, at opportune places in the vtkMRML*StorageNode methods.
3.     Maybe, I am significantly off the mark and there is a better approach?  If so, what am I not considering?

Thanks!

lassoan:
[...]

PR #5045, opened 2020-07-16
ENH: Display warnings and errors in the Save Data dialog

Leengit:
This is the code associated with https://github.com/Slicer/Slicer/issues/4968 for collecting the user-friendly messages at the time of saving.  Still to do is presenting the collected messages in the Base/QTGUI/qSlicerSaveDataDialog.cxx.

### DIFF
--- a/Base/QTGUI/qSlicerSaveDataDialog.cxx
+++ b/Base/QTGUI/qSlicerSaveDataDialog.cxx
@@ -19,9 +19,11 @@
 ==============================================================================*/
 
 /// Qt includes
+#include <QApplication>
 #include <QComboBox>
 #include <QDate>
 #include <QDebug>
+#include <QLabel>
 #include <QLineEdit>
 #include <QMessageBox>
 #include <QRegExp>
@@ -49,6 +51,7 @@
 #include <vtkCollection.h>
 #include <vtkDataFileFormatHelper.h> // for GetFileExtensionFromFormatString()
 //#include <vtkMRMLHierarchyNode.h>
+#include <vtkMRMLMessageCollection.h>
 #include <vtkMRMLScene.h>
 #include <vtkMRMLStorageNode.h>
 #include <vtkMRMLSceneViewNode.h>
@@ -271,7 +274,7 @@ void qSlicerSaveDataDialogPrivate::populateItems()
   for (it = nodes.begin(); it != nodes.end(); it++)
     {
     vtkMRMLSceneViewNode *svNode = vtkMRMLSceneViewNode::SafeDownCast(*it);
-    // skip "Master Scene View" since it contains the same ndoes as the scene
+    // skip "Master Scene View" since it contains the same nodes as the scene
     if (svNode->GetName() && std::string("Master Scene View") == std::string(svNode->GetName()))
       {
       continue;
@@ -379,6 +382,11 @@ void qSlicerSaveDataDialogPrivate::populateScene()
 
   this->FileWidget->setCellWidget(row, FileDirectoryColumn, sceneDirectoryButton);
 
+  // Scene User Messages
+  vtkMRMLStorableNode * const storableNode = vtkMRMLStorableNode::SafeDownCast(this->object(row));
+  QWidget * const userMessagesItem = this->createUserMessagesItem(storableNode);
+  this->FileWidget->setCellWidget(row, UserMessagesColumn, userMessagesItem);
+
   // Scene Selected
   QTableWidgetItem* selectItem = this->FileWidget->item(row, SelectColumn);
   selectItem->setFlags(selectItem->flags() | Qt::ItemIsUserCheckable);
@@ -480,6 +488,10 @@ void qSlicerSaveDataDialogPrivate::populateNode(vtkMRMLNode* node)
   // File Directory
   QWidget* directoryWidget = this->createFileDirectoryWidget(fileInfo);
   this->FileWidget->setCellWidget(row, FileDirectoryColumn, directoryWidget);
+
+  // User messages
+  QWidget *userMessagesItem = this->createUserMessagesItem(storableNode);
+  this->FileWidget->setCellWidget(row, UserMessagesColumn, userMessagesItem);
 
   // Select modified nodes by default
   QTableWidgetItem* selectItem = this->FileWidget->item(row, SelectColumn);
@@ -710,6 +722,93 @@ ctkDirectoryButton* qSlicerSaveDataDialogPrivate::createFileDirectoryWidget(cons
 }
 
 //-----------------------------------------------------------------------------
+void qSlicerSaveDataDialogPrivate
+::makeUserMessages(vtkMRMLStorableNode *node, QWidget *userMessagesWidget)
+{
+  QHBoxLayout * const userMessagesLayout = userMessagesWidget ?
+    dynamic_cast<QHBoxLayout *>(userMessagesWidget->layout()) : nullptr;
+
+  if (!userMessagesLayout)
+    {
+    return;
+    }
+
+  static const QPixmap okayIcon = qApp->style()->standardPixmap(QStyle::SP_DialogApplyButton);
+  static const QPixmap warningIcon = qApp->style()->standardPixmap(QStyle::SP_MessageBoxWarning).scaledToHeight(18);
+  static const QPixmap errorIcon = qApp->style()->standardPixmap(QStyle::SP_MessageBoxCritical).scaledToHeight(18);
+  static const QPixmap questionIcon = qApp->style()->standardPixmap(QStyle::SP_MessageBoxQuestion).scaledToHeight(18);
+
+  const vtkMRMLStorageNode * const snode = node ? node->GetStorageNode() : nullptr;
+
+  // Clear out any existing items in the userMessagesLayout
+  QLayoutItem *layoutItem = nullptr;
+  while ((layoutItem = userMessagesLayout->takeAt(0)) != nullptr)
+    {
+    delete layoutItem;
+    }
+  // Put new widgets into the userMessagesLayout
+  if (snode && snode->GetUserMessages()->GetNumberOfMessages() > 0)
+    {
+    // There is at least one message from the storage node.
+    const vtkMRMLMessageCollection * const messageCollection = snode->GetUserMessages();
+    const int numberOfMessages = messageCollection->GetNumberOfMessages();
+
+    for (int index = 0; index < numberOfMessages; ++index)
+      {
+      const unsigned long messageType = messageCollection->GetNthMessageType(index);
+      const std::string messageText = messageCollection->GetNthMessageText(index);
+      // Locate the appropriate icon.
+      const QPixmap *icon = nullptr;
+      switch (messageType)
+        {
+        case vtkCommand::WarningEvent:
+          icon = &warningIcon;
+          break;
+        case vtkCommand::ErrorEvent:
+          icon = &errorIcon;
+          break;
+        default:
+          icon = &questionIcon;
+          break;
+        }
+      // Use icon in widget.  SetToolTip for widget.  Install widget in layout.
+      QLabel *messageLabel = new QLabel;
+      messageLabel->setPixmap(*icon);
+      messageLabel->setToolTip(QString(tr(messageText.c_str())));
+      userMessagesLayout->addWidget(messageLabel);
+      }
+    }
+  userMessagesWidget->show();
+}
+
+//-----------------------------------------------------------------------------
+void qSlicerSaveDataDialogPrivate
+::clearUserMessages()
+{
+  for (int row = 0; row < this->FileWidget->rowCount(); ++row)
+    {
+    vtkMRMLStorableNode* storableNode;
+    vtkMRMLStorageNode* snode;
+    if ((storableNode = vtkMRMLStorableNode::SafeDownCast(this->object(row))) && (snode = storableNode->GetStorageNode()))
+      {
+      snode->GetUserMessages()->ClearMessages();
+      }
+    }
+}
+
+//-----------------------------------------------------------------------------
+QWidget *qSlicerSaveDataDialogPrivate
+::createUserMessagesItem(vtkMRMLStorableNode *node)
+{
+  QHBoxLayout * const userMessagesLayout = new QHBoxLayout;
+  userMessagesLayout->setEnabled(true);
+  QWidget * const qWidget = new QWidget;
+  qWidget->setLayout(userMessagesLayout);
+  this->makeUserMessages(node, qWidget);
+  return qWidget;
+}
+
+//-----------------------------------------------------------------------------
 QString qSlicerSaveDataDialogPrivate::extractKnownExtension(const QString& fileName, vtkObject* object)
 {
   qSlicerCoreIOManager* coreIOManager =
@@ -787,12 +886,15 @@ bool qSlicerSaveDataDialogPrivate::save()
     // restore the root directory only if the scene is not saved
     this->restoreAfterSaving();
     }
+  // Save was a total success.  The user messages are no longer relevant.
+  this->clearUserMessages();
   return true;
 }
 
 //-----------------------------------------------------------------------------
 bool qSlicerSaveDataDialogPrivate::saveNodes()
 {
+  bool doneWithSaveDataDialog = true;
   QMessageBox::StandardButton forceOverwrite = QMessageBox::Ignore;
   QList<qSlicerIO::IOProperties> files;
   const int sceneRow = this->findSceneRow();
@@ -823,42 +925,51 @@ bool qSlicerSaveDataDialogPrivate::saveNodes()
     QString format = this->format(row);
     qSlicerIOOptions* options = this->options(row);
     vtkMRMLNode* node = vtkMRMLNode::SafeDownCast(this->object(row));
+    vtkMRMLStorableNode* const storableNode = vtkMRMLStorableNode::SafeDownCast(this->object(row));
+    vtkMRMLStorageNode* const snode = storableNode ? storableNode->GetStorageNode() : nullptr;
 
     if (file.fileName().isEmpty())
       {
-      QMessageBox::warning(this, tr("Saving node..."),
-                           tr("Node %1 not saved, file name is empty.").arg(nodeNameItem->text()));
+      if (snode)
+        {
+        snode->GetUserMessages()->AddMessage(vtkCommand::WarningEvent,
+          (tr("Node %1 not saved, file name is empty.").arg(nodeNameItem->text())).toStdString());
+        }
+      else
+        {
+        QMessageBox::warning(this, tr("Saving node..."),
+          tr("Node %1 not saved, file name is empty.").arg(nodeNameItem->text()));
+        }
+      doneWithSaveDataDialog = false;
       continue;
       }
 
     // check if the file already exists
     if (file.exists())
       {
-      if (forceOverwrite == QMessageBox::NoToAll)
+      QMessageBox::StandardButton answer = forceOverwrite;
+      if (answer == QMessageBox::Ignore)
         {
-        continue;
+        answer = QMessageBox::question(this, tr("Saving node..."),
+          tr("The file: %1 already exists."
+            " Do you want to replace it ?").arg(file.absoluteFilePath()),
+          QMessageBox::Yes | QMessageBox::No |
+          QMessageBox::YesToAll | QMessageBox::NoToAll,
+          QMessageBox::Yes);
+        if (answer == QMessageBox::YesToAll || answer == QMessageBox::NoToAll)
+          {
+          forceOverwrite = answer;
+          }
         }
-      if (forceOverwrite == QMessageBox::Ignore)
+      if (answer == QMessageBox::No || answer == QMessageBox::NoToAll)
         {
-        QMessageBox::StandardButton answer =
-          QMessageBox::question(this, tr("Saving node..."),
-                                tr("The file: %1 already exists."
-                                   " Do you want to replace it ?").arg(file.absoluteFilePath()),
-                                QMessageBox::Yes | QMessageBox::No |
-                                QMessageBox::YesToAll | QMessageBox::NoToAll,
-                                QMessageBox::Yes);
-        if (answer == QMessageBox::YesToAll)
+        if (snode)
           {
-          forceOverwrite = QMessageBox::YesToAll;
+          snode->GetUserMessages()->AddMessage(vtkCommand::WarningEvent,
+            (tr("The file: %1 already exists. Not replaced.").arg(file.absoluteFilePath())).toStdString());
           }
-        else if (answer == QMessageBox::NoToAll)
-          {
-          forceOverwrite = QMessageBox::NoToAll;
-          }
-        if (answer == QMessageBox::No || answer == QMessageBox::NoToAll)
-          {
-          continue;
-          }
+        doneWithSaveDataDialog = false;
+        continue;
         }
       }
 
@@ -889,15 +1000,42 @@ bool qSlicerSaveDataDialogPrivate::saveNodes()
                               QMessageBox::Yes | QMessageBox::No, QMessageBox::Yes);
       if (answer == QMessageBox::No)
         {
+        if (snode)
+          {
+          snode->GetUserMessages()->AddMessage(vtkCommand::ErrorEvent,
+            (tr("Cannot write data file: %1. Stopping save operation at user's request.").arg(file.absoluteFilePath())).toStdString());
+          }
         return false;
         }
+
+      if (snode)
+        {
+        snode->GetUserMessages()->AddMessage(vtkCommand::WarningEvent,
+          (tr("Cannot write data file: %1.").arg(file.absoluteFilePath())).toStdString());
+        }
+      doneWithSaveDataDialog = false;
+      continue;
       }
 
-    // clean up node after saving
+    // clean up node after successfully saving
+    if (snode)
+      {
+      snode->GetUserMessages()->ClearMessages();
+      }
     nodeNameItem->setCheckState(Qt::Unchecked);
     nodeStatusItem->setText("Not Modified");
     }
-  return true;
+
+  // Updating the messages is useful only if the dialog box is not about to go away.  doneWithSaveDataDialog==false
+  // means that the dialog box is not about to go away but there are other that it could remain, so, just in case, we
+  // will prepare the messages as if the dialog box is not going away.
+  for (int row = 0; row < this->FileWidget->rowCount(); ++row)
+    {
+    updateUserMessagesItem(row);
+    }
+  this->updateSize();
+
+  return doneWithSaveDataDialog;
 }
 
 //-----------------------------------------------------------------------------
@@ -1297,6 +1435,15 @@ void qSlicerSaveDataDialogPrivate::updateOptionsWidget(int row)
 }
 
 //-----------------------------------------------------------------------------
+void qSlicerSaveDataDialogPrivate
+::updateUserMessagesItem(int row)
+{
+  vtkMRMLStorableNode * const node = vtkMRMLStorableNode::SafeDownCast(this->object(row));
+  QWidget * const userMessagesWidget = this->FileWidget->cellWidget(row, UserMessagesColumn);
+  this->makeUserMessages(node, userMessagesWidget);
+}
+
+//-----------------------------------------------------------------------------
 void qSlicerSaveDataDialogPrivate::showMoreColumns(bool show)
 {
   this->FileWidget->setColumnHidden(OptionsColumn, !show);
@@ -1397,19 +1544,21 @@ void qSlicerSaveDataDialogPrivate::enableNodes(bool enable)
       {
       continue;
       }
-    QTableWidgetItem* item = this->FileWidget->item(i, FileNameColumn);
+    QTableWidgetItem* fileNameItem = this->FileWidget->item(i, FileNameColumn);
     if (enable)
       {
-      item->setFlags(item->flags() | Qt::ItemIsEnabled);
+      fileNameItem->setFlags(fileNameItem->flags() | Qt::ItemIsEnabled);
       }
     else
       {
-      item->setFlags(item->flags() & ~Qt::ItemIsEnabled);
+      fileNameItem->setFlags(fileNameItem->flags() & ~Qt::ItemIsEnabled);
       }
     QWidget* fileFormatWidget = this->FileWidget->cellWidget(i, FileFormatColumn);
     fileFormatWidget->setEnabled(enable);
     QWidget* fileDirectoryWidget = this->FileWidget->cellWidget(i, FileDirectoryColumn);
     fileDirectoryWidget->setEnabled(enable);
+    QWidget* userMessagesWidget = this->FileWidget->cellWidget(i, UserMessagesColumn);
+    userMessagesWidget->setEnabled(enable);
     }
 }
 
--- a/Libs/MRML/Core/vtkMRMLMessageCollection.cxx
+++ b/Libs/MRML/Core/vtkMRMLMessageCollection.cxx
@@ -1,0 +1,121 @@
+/*==============================================================================
+
+  Program: 3D Slicer
+
+  Copyright (c) Kitware Inc.
+
+  See COPYRIGHT.txt
+  or http://www.slicer.org/copyright/copyright.txt for details.
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+
+  This file was originally developed by Lee Newberg, Kitware Inc.
+
+==============================================================================*/
+
+
+#include "vtkMRMLMessageCollection.h"
+
+// MRML includes
+#include "vtkCommand.h"
+
+//----------------------------------------------------------------------------
+vtkStandardNewMacro(vtkMRMLMessageCollection);
+
+//----------------------------------------------------------------------------
+void
+vtkMRMLMessageCollection
+::PrintSelf(ostream& os, vtkIndent indent)
+{
+  Superclass::PrintSelf(os,indent);
+  os << indent << "Messages: " << &this->Messages << "\n";
+  for (int i = 0; i < this->GetNumberOfMessages(); i++)
+    {
+    const unsigned long messageType = this->GetNthMessageType(i);
+    const std::string messageText = this->GetNthMessageText(i);
+    os << indent << "MessagesMember: " << messageType << " "
+       << messageText << "\n";
+    }
+}
+
+//----------------------------------------------------------------------------
+vtkMRMLMessageCollection::Message
+::Message(unsigned long messageType, const std::string &messageText)
+  : MessageType(messageType), MessageText(messageText)
+{
+}
+
+//----------------------------------------------------------------------------
+int
+vtkMRMLMessageCollection
+::GetNumberOfMessages() const
+{
+  return this->Messages.size();
+}
+
+//----------------------------------------------------------------------------
+int
+vtkMRMLMessageCollection
+::GetNumberOfMessagesOfType(unsigned long messageType) const
+{
+  int response = 0;
+  for (int i = 0; i < this->Messages.size(); ++i)
+    {
+    response += static_cast<int>(GetNthMessageType(i) == messageType);
+    }
+  return response;
+}
+
+//----------------------------------------------------------------------------
+int
+vtkMRMLMessageCollection
+::GetNumberOfMessagesOfType(const char *eventName) const
+{
+  return GetNumberOfMessagesOfType(vtkCommand::GetEventIdFromString(eventName));
+}
+
+//----------------------------------------------------------------------------
+unsigned long
+vtkMRMLMessageCollection
+::GetNthMessageType(int index) const
+{
+  return this->Messages.at(index).MessageType;
+}
+
+//----------------------------------------------------------------------------
+std::string
+vtkMRMLMessageCollection
+::GetNthMessageText(int index) const
+{
+  return this->Messages.at(index).MessageText;
+}
+
+//----------------------------------------------------------------------------
+void
+vtkMRMLMessageCollection
+::AddMessage(unsigned long messageType, const std::string &messageText)
+{
+  this->Messages.push_back({messageType, messageText});
+}
+
+//----------------------------------------------------------------------------
+void
+vtkMRMLMessageCollection
+::ClearMessages()
+{
+  this->Messages.clear();
+}
+
+//----------------------------------------------------------------------------
+vtkMRMLMessageCollection::vtkMRMLMessageCollection()
+{
+}
+
+//----------------------------------------------------------------------------
+vtkMRMLMessageCollection::~vtkMRMLMessageCollection()
+{
+}
--- a/Libs/MRML/Core/vtkMRMLMessageCollection.h
+++ b/Libs/MRML/Core/vtkMRMLMessageCollection.h
@@ -1,0 +1,89 @@
+/*==============================================================================
+
+  Program: 3D Slicer
+
+  Copyright (c) Kitware Inc.
+
+  See COPYRIGHT.txt
+  or http://www.slicer.org/copyright/copyright.txt for details.
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+
+  This file was originally developed by Lee Newberg, Kitware Inc.
+
+==============================================================================*/
+
+#ifndef __vtkMRMLMessageCollection_h
+#define __vtkMRMLMessageCollection_h
+
+// MRML includes
+#include "vtkMRML.h"
+
+// VTK includes
+#include <vtkIdTypeArray.h>
+
+// STD includes
+#include <string>
+#include <vector>
+
+///
+/// A class for recording warnings and errors associated with this
+/// vtkMRMLStorableNode.  A substantially similar vtkMessageCollection
+/// is planned for vtkAddon.  When that is released, this will be
+/// replaced by that.
+class VTK_MRML_EXPORT vtkMRMLMessageCollection : public vtkObject
+{
+public:
+  vtkTypeMacro(vtkMRMLMessageCollection,vtkObject);
+  void PrintSelf(ostream& os, vtkIndent indent) override;
+  static vtkMRMLMessageCollection* New();
+
+  ///
+  /// Return the total number of reported messages.
+  virtual int GetNumberOfMessages() const;
+
+  ///
+  /// Return the number of reported messages of the specified event
+  /// type
+  virtual int GetNumberOfMessagesOfType(unsigned long messageType) const;
+
+  ///
+  /// Return the number of reported messages of the specified event
+  /// type
+  virtual int GetNumberOfMessagesOfType(const char *eventName) const;
+
+  // Get the Nth message from the message vector
+  virtual unsigned long GetNthMessageType(int index) const;
+  virtual std::string GetNthMessageText(int index) const;
+
+  // Append a message to the message vector
+  virtual void AddMessage(unsigned long messageType, const std::string &messageText);
+
+  // Clear the message vector
+  virtual void ClearMessages();
+
+protected:
+  vtkMRMLMessageCollection();
+  ~vtkMRMLMessageCollection() override;
+  vtkMRMLMessageCollection(const vtkMRMLMessageCollection&);
+  void operator=(const vtkMRMLMessageCollection&);
+
+  ///
+  /// A helper class that describes a single message
+  struct Message
+    {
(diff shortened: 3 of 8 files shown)


Libs/MRML/Core/vtkMRMLStorageNode.cxx
```diff
@@ -144,6 +144,8 @@ void vtkMRMLStorageNode::WriteXML(ostream& of, int nIndent)
       else
         {
         vtkWarningMacro("WriteXML: unable to convert relative file path to absolute, still using " << this->FileName);
+        GetVTKMessageLog()->SetMessage(vtkMessageLog::Message(vtkCommand::WarningEvent,
```
pieper, line 147: Looks like these entries in the message log are always the same as the corresponding macro text.  In this case I'd prefer to see something like `vtkMRMLStorageNodeWarningMacro` that internally uses `vtkWarningMacro` and sets the message log.  Same for error cases.

Base/QTGUI/qSlicerSaveDataDialog.cxx
```diff
@@ -67,7 +67,7 @@ namespace
 }
 
 //-----------------------------------------------------------------------------
-qSlicerFileNameItemDelegate::qSlicerFileNameItemDelegate( QObject * parent )
+qSlicerFileNameItemDelegate::qSlicerFileNameItemDelegate(QObject* parent)
```
pieper, line 70: Putting the const and spacing changes into a different commit would make it easier to review the changes specific to this PR.

Libs/MRML/Core/vtkMRMLStorageNode.h
```diff
@@ -26,6 +27,76 @@ class vtkStringArray;
...
+    ///
+    /// Function object for checking whether a Message has a specified
+    /// EventID.
+    struct EqualEventId
+    {
+      EqualEventId(event_type eventId) : EventId(eventId) {}
+      bool operator()(const Message &message) const
+        { return message.EventId == this->EventId; }
+
+      event_type EventId;
```
pieper, line 57: Since you are using these with values like `vtkCommand::ErrorEvent` they aren't really `EventId`s as much as `MessageType`s.  `MessageType` would make a better field name.

Also maybe this could be implemented as a subclass of a `vtkVariantArray` so that you wouldn't need to invent a new API for the common operations.

Base/QTGUI/qSlicerSaveDataDialog.cxx
lassoan: Has actually anything changed in the dialog (other than style)? I've already spent 5 minutes with reviewing changes, but all that I noticed were trivial changes and don't see any functional difference. Maybe I just did not pay enough attention, but now I don't want to spend another 5-10 minutes with reading through these changes again. I agree with Steve that you could submit these style changes in another code pull request so that we can focus on the functional changes here.

pieper: Yes, I waded through a lot before I found the substantive changes - the message log text is appended to the error dialogs if the data doesn't save.  I'd include the line numbers here but that would mean I need to hunt for them again 😄

Libs/MRML/Core/vtkMRMLStorageNode.h
```diff
@@ -369,6 +440,9 @@ class VTK_MRML_EXPORT vtkMRMLStorageNode : public vtkMRMLNode
   static const char *GetCoordinateSystemTypeAsString(int id);
   static int GetCoordinateSystemTypeFromString(const char *name);
 
+  const vtkMessageLog *GetVTKMessageLog() const { return &this->MessageLog; }
+  vtkMessageLog *GetVTKMessageLog() { return &this->MessageLog; }
```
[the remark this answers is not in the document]
Leengit: Will do.

Libs/MRML/Core/vtkMRMLStorageNode.cxx
Leengit: The current text for these SetMessage calls is just a placeholder.  We should definitely change these.  The brevity level, format, etc. for these user-friendly messages will become more obvious (to me) once I start seeing them in the dialog box in the upcoming implementation.

Base/QTGUI/qSlicerSaveDataDialog.cxx
Leengit: I will certainly do that next time.  Would it help if I put in another commit to this pull request that undoes these spacing changes?

Libs/MRML/Core/vtkMRMLStorageNode.h
Leengit: I will change EventId to MessageType.  I will take a look at `vtkVariantArray`.

Libs/MRML/Core/vtkMRMLStorageNode.cxx
```diff
@@ -280,10 +284,17 @@ void vtkMRMLStorageNode::ReadXMLAttributes(const char** atts)
         if (vtksys::SystemTools::FileExists(secondCollapsedFullPath.c_str(), true) == false)
           {
           vtkWarningMacro("ReadXMLAttributes: file list member " << filename << " is relative, but not found relative to the mrml root directory. Trying to find the path relative to the archetype doesn't work either: " << secondCollapsedFullPath << " not found, going back to using path from mrml root dir of " << collapsedFullPath);
+          GetVTKMessageLog()->SetMessage(vtkMessageLog::Message(vtkCommand::WarningEvent,
```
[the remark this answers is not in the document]
Leengit: I will change `SetMessage()` to `AddMessage()`.  I will remove that message.

Libs/MRML/Core/vtkMRMLStorageNode.cxx
```diff
@@ -280,10 +284,17 @@ void vtkMRMLStorageNode::ReadXMLAttributes(const char** atts)
...
+              std::string("ReadXMLAttributes: file list member ") + filename +
+              " is relative, but not found relative to the mrml root directory. Trying to find the path relative to the archetype doesn't work either: " +
+              secondCollapsedFullPath + " not found, going back to using path from mrml root dir of " + collapsedFullPath));
           }
         else
           {
           vtkWarningMacro("ReadXMLAttributes: file list member " << filename << " is relative, but not found relative to the mrml root directory: " << collapsedFullPath << " is invalid. Found the file relative to the archetype, so using path  " <<  secondCollapsedFullPath);
+          GetVTKMessageLog()->SetMessage(vtkMessageLog::Message(vtkCommand::WarningEvent,
```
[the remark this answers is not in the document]
Leengit: I will remove that message.

Libs/MRML/Core/vtkMRMLStorageNode.h
```diff
@@ -26,6 +27,76 @@ class vtkStringArray;
...
 #include <vector>
+#include <algorithm>
+
+///
+/// A class for recording warnings and errors associated with this
+/// vtkMRMLStorableNode.  This cannot be hidden within the
+/// vtkMRMLStorableNode class scope because the Python wrapping code
+/// vtkMRMLStorageNodePython.cxx cannot wrap the GetVTKMessageLog
+/// method.
+class vtkMessageLog
```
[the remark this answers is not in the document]
Leengit: At present, `vtkMessageLog` is a helper class within `Libs/MRML/Core/vtkMRMLStorageNode.h`.  I am thinking that when we decide that it is ready for more general use, we can have it be standalone, inherit from `vtkObject`, etc.

At https://github.com/Slicer/Slicer/issues/4968 @pieper wrote
> In terms of MRML, only vtkMRMLStorable subclasses are stored, so the vtkMRMLStorageNode is the correct place to focus the error sink checking.

At https://github.com/Slicer/Slicer/issues/4968 @lassoan wrote
> Exactly. For now, you can implement just a few checks in vtkMRMLStorageNode. You don't even need to bother with vtkErrorSink, just do some trivial checks as a proof of concept. We'll add more checks later.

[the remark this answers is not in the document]
Leengit: I will change it to `vtkMessages`.

Leengit:
d190b0a has these name changes:
- class vtkMessageLog -> class vtkMessages
- event_type -> unsigned long
- EventId -> MessageType
- size_type -> int
- GetVTKMessageLog() -> GetUserMessages()
- SetMessage(...) -> AddMessage(...)
- vtkMRMLStorageNode member: vtkMessageLog MessageLog -> vtkMessages Messages

Libs/MRML/Core/vtkMRMLStorageNode.h
```diff
@@ -26,6 +27,76 @@ class vtkStringArray;
...
+      event_type EventId;
+    };
+
+    event_type EventId;
+    std::string MessageText;
+  };
+
+  ///
+  /// The kind of integer that is used for indexing the message vector
+  using size_type = typename std::vector<Message>::size_type;
```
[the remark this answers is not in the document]
Leengit: On my system, `std::vector<Message>::size_type` is `unsigned long`.  I will use `int`, but if it should be `unsigned` or `unsigned long`, please let me know.

Libs/MRML/Core/vtkMRMLStorageNode.h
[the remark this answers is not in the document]
pieper: I'd still vote for making it a stand alone class inheriting from `vtkOjbect` and putting it in the `vtkAddons` library.  The ability to access and inspect it from Python will be very useful, and also the ability to add observers.

Libs/MRML/Core/vtkMRMLStorageNode.cxx
pieper: +1 for `AddMessage`

Base/QTGUI/qSlicerSaveDataDialog.cxx
pieper: It looks like we will have another couple rounds of discussion, so yes, if you can revert the spacing changes and put them in a different PR that would make things easier.

Libs/MRML/Core/vtkMRMLStorageNode.h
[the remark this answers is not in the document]
pieper: Using the `vtkVariantArray` or other standard vtk container would sidestep this issue.

Base/QTGUI/qSlicerSaveDataDialog.cxx
Leengit: Commit https://github.com/Slicer/Slicer/pull/5045/commits/e350b0e2c61beb6973ef1e2e5afbf3dfb9da4103 reverts the const, Self::, this->, and spacing changes, leaving only those changes that are most specific to the present pull request.

Libs/MRML/Core/vtkMRMLStorageNode.h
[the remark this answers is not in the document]
pieper: The more I think about it, two `vtkStringArray`s like `ErrorMessages` and `WarningMessages` instance variables might actually be all we need.  This would get rid of the helper class, and also provide the python wrapping and observability.  What do you think of this simpler solution?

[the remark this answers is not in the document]
Leengit: Pros: `vtkStringArray` provides all sorts of methods that have proved useful, at least in other circumstances.  Its integer type for indexing is already fixed independent of the system on which it is compiled (unlike the case for `std::vector<...>`), so we will not have to worry about issues relating to exposing `int` when the underlying `std::vector<...>` instead supports its own `size_type`.

Cons: having exactly two instances of `vtkStringArray` supports exactly two message types, errors and warnings; in this storage node context or some future context will we want to support a larger or flexible subset of the message types from `VTK/Common/Core/vtkCommand.h`?

Proposal: we leave this as is for now (i.e., using a `std::vector<typename vtkMessages::Message>` object rather than a pair of `vtkStringArray` objects) and we revisit once we have implemented the `Slicer/Base/QTGUI/qSlicerSaveDataDialog.cxx` side of things.  It is plausible that the needs in `qSlicerSaveDataDialog.cxx` will advise this situation.

Leengit:
Commit https://github.com/Slicer/Slicer/commit/e350b0e2c61beb6973ef1e2e5afbf3dfb9da4103 rebases to the current master branch.

Libs/MRML/Core/vtkMRMLStorageNode.h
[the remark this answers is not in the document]
pieper: Sure, it's not technically wrong the way it is now, and yes, it's good to prototype the whole thing to get a feel for how all the pieces work.

I'll still push for re-using existing VTK concepts wherever possible, not just for the python and event benefits, but also it's more maintainable and readable in a large code base because people know the naming scheme, trust the implementation isn't buggy, know it will compile across platforms, etc.  The slicer policy is to follow the coding conventions of the parent class, and VTK is good at this.  Here you might use an std map of event ids to vtkStringArray for example, so there's only one instance variable; this pattern is used a lot in VTK code.

Libs/MRML/Core/vtkMRMLStorageNode.h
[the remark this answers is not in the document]
Leengit: Correction: we will change `GetVTKMessageLog()` / `GetMessageLog()` to `GetUserMessages()`.

Leengit:
I just pushed to this pull request a proof-of-concept implementation of displaying the user messages to the user.  Currently the message is always "Message: all good!" and it is simple text.  Feedback on the code itself or any aspect of the UIX would be much appreciated.

Base/QTGUI/qSlicerSaveDataDialog.cxx
```diff
@@ -823,42 +880,50 @@ bool qSlicerSaveDataDialogPrivate::saveNodes()
...
     vtkMRMLNode* node = vtkMRMLNode::SafeDownCast(this->object(row));
+    vtkMRMLStorageNode* const storageNode = vtkMRMLStorageNode::SafeDownCast(this->object(row));
 
     if (file.fileName().isEmpty())
       {
-      QMessageBox::warning(this, tr("Saving node..."),
-                           tr("Node %1 not saved, file name is empty.").arg(nodeNameItem->text()));
+      if (storageNode)
+        {
+        storageNode->GetUserMessages()->AddMessage({vtkCommand::WarningEvent,
```
pieper, line 889: I think it would make more sense to move this logic so that the save button is not enabled unless the proper fields have been filled out.  Basically add a validation function that is run when any ui elements are edited.

Libs/MRML/Core/vtkMRMLStorageNode.h
```diff
@@ -26,6 +27,71 @@ class vtkStringArray;
...
 #include <vector>
+#include <algorithm>
+
+///
+/// A class for recording warnings and errors associated with this
+/// vtkMRMLStorableNode.  This cannot be hidden within the
+/// vtkMRMLStorableNode class scope because the Python wrapping code
+/// vtkMRMLStorageNodePython.cxx cannot wrap the GetUserMessages
+/// method.
+class vtkMessages
```
pieper, line 38: I'm still not a fan of this class 😛

Leengit: I hear you.  Let's figure this out.

I fear that `std::map<vtkStringArray>` may suffer from some of the same problems as `std::vector<vtkMessages::Message>`.  Either way, the use of the standard library means that we deal with the system-specific integer types such as `size_type` that the standard library employs.

The `std::vector<vtkMessages::Message>` approach allows us to add additional fields to the `Message` struct (such as a time stamp, \_\_LINE\_\_, or cetera) should we decide that we want them later.

Perhaps a hybrid approach that uses `vtkVariantArray<vtkMessage>` for some new `vtkMessage` struct (not plural!) is the way to go.

Regardless of the previous paragraphs, we can have any new structures inherit from `vtkObject`.

Base/QTGUI/qSlicerSaveDataDialog.cxx
Leengit: I fear that I am not understanding what you are saying.  With the current implementation, 

1. any user messages from the `vtkMRMLStorageNode` nodes that occur _before_ the Save Data Dialog box is displayed will get shown to the user in a new last column of that dialog box.
1. when the Save button is clicked, if there are any additional warnings and errors then the dialog box will get updated to also have the new messages.  

Am I correct that you are suggesting a step between those two?  That is, if any warnings or errors are displayed in that first step, we should gray out the Save button and otherwise make the user do something in order to have the right to Save?

Or am I way off base and the two implemented steps above are not what you were expecting?

Leengit:
At the moment I am displaying "Message: All good!" for debug purposes.  Normally that will be blank unless there are actual user-style Warning or Error messages.  At present Warning and Error messages change that text to something like
```
Warning: Could not find my navel.
Error: Toe jam football.
```

Assuming that this is on the right track, having actual warning and error icons and other prettier formatting would be a next step.

Screenshot from 2020-07-24 10-50-58

Libs/MRML/Core/vtkMRMLStorageNode.h
Leengit: The `vtkVariantArray` discussion is continued below.  Marking this thread as resolved.

Libs/MRML/Core/vtkMRMLStorageNode.h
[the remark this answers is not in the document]
Leengit: The `std::map` approach, etc. is now discussed in another thread.  I am marking this thread as resolved.

Leengit:
As I am trying to exercise my new code -- by reading in an `MR-head.nrrd` or `3DHeadScene.mrml` file, rotating an image, and then saving it back out -- I am finding that no node returned from a `qSlicerSaveDataDialogPrivate->object(row)` call is (a sublcass of) a `vtkMRMLStorageNode` object.  Thus, I am never finding any node for which any user messages could have been recorded for display in the Save Data Dialog.  What am I doing wrong?

Libs/MRML/Core/vtkMRMLStorageNode.h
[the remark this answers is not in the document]
Leengit: This conversation is continued in another thread and I am marking this thread as resolved.

Libs/MRML/Core/vtkMRMLStorageNode.h
pieper: It's very tempting to migrate everything to std container, but the vtk ones do a nice job of smoothing out any of the cross-platform issues (the worst of those days are gone, but they still crop up sometimes and it's really nice when all the code gets fixed in one place with some cmake or other magic).  So yes, I like your hybrid approach.

We increasingly rely on having scripted control over the application, so making sure the API is wrapped completely is a nice byproduct of using vtkObject derivations unless there's a strong performance or other reason not to.

Leengit:
My ability to put substantial time into this issue ends at the end of the month, Friday.  I would appreciate your thoughts on whether the current implementation is ready to merge or, if not, what improvements will get it to something worth merging.

Libs/MRML/Core/vtkMRMLStorageNode.h
Leengit: I  want to stick to the current implementation with `std::vector<vtkMessage>` rather than switching to an implementation with `vtkVariantArray`. 

Details: `vtkVariantArray` is not a C++ template, but a container that handles a mixture of types.  I think that you are suggesting that we make use of its ability to contain ordinary pointers to `vtkObjectBase` objects.  So, `vtkMRMLStorageNode` would implement methods to handle memory management, and the setting and parsing of the components of a `vtkMessage` (singular) subclass of `vtkObject`.  In anticipation of having messages for other than `vtkMRMLStorageNode` objects, I am thinking that we would create a `vtkMessages` (plural) class to handle that memory management, setting, and parsing, and we would have `vtkMRMLStorageNode` use that.

However that reasoning brings us back almost full circle to the current design, which differs only in that it uses a `std::vector<vtkMessage>` struct instead of `vtkVariantArray(vtkObjectBase *)` calls.  I note that the `vtkMRMLStorageNode` already has 3 members that are `std::vector<...>` and 2 methods that return `std::vector<...>`, so I think this additional use of `std::vector<...>` is likely to work.  Because `std::vector` accepts actual objects, not merely pointers to them, it is going to save me having to worry about memory management.  And it has the advantage that it is already written and has been through a bit of debugging.

In short, I want to stick to the current implementation with `std::vector<vtkMessage>` rather than switching to an implementation with `vtkVariantArray`.

lassoan: I don't think vtkVariantArray is applicable here.

VTK style would be to create a vtkMessage class (based on vtkObject, therefore having Python wrapping) and for convenience, add a vtkMessageCollection class (based on vtkCollection, with some additional methods that are useful for setting/getting/procesing messages). However, this would be an overkill for a trivially small message class (that only contains a string and an int).

Probably we could just add a message struct, std::deque<message>, and few methods directly to the storage node, but since you already implemented a separate class, we can just use that (it is somewhat nicer in a separate class, so that it can be moved around easier in future code refactorings). Since vtkMessages is not a VTK class, its name must not begin with "vtk" and it has to be a private implementation (not in the public header file). Its features can be exposed via a few new methods in vtkMRMLStorageNode.

pieper: Stepping back a bit, I think my main point is that this all feels like too much code to address the actual use case.

I think we could accomplish the task at hand with a simple string to capture any messages.  If there's more than one message they can be separated by newlines.  And each line can be prefixed by 'Warning' or 'Error' and the string can be placed directly in the gui.

I see that the original mockup from @lassoan included the count of the number of errors and warnings but I'm not sure that's very meaningful to users.  The main thing is to flag them when something happens and then they need to read the details of the message.

Leengit: I have spent the morning adding `vtkObject` support to `vtkMessages` and in response to @lassoan's comment just above I am changing the name of latter to `vtkMessageCollection`.  It is now in the pull request as commit https://github.com/Slicer/Slicer/pull/5045/commits/aa236538876d8a31ac7f0fa4053ed000bf4b33a3.

Base/QTGUI/qSlicerSaveDataDialog.cxx
```diff
@@ -1296,6 +1402,17 @@ void qSlicerSaveDataDialogPrivate::updateOptionsWidget(int row)
   this->FileWidget->resizeColumnToContents(OptionsColumn);
 }
 
+//-----------------------------------------------------------------------------
+void qSlicerSaveDataDialogPrivate
+::updateUserMessagesItem(int row)
+{
+vtkMRMLStorableNode * const node = vtkMRMLStorableNode::SafeDownCast(this->object(row));
+  QString allUserMessages = makeUserMessages(node);
```
jcfr, line 1410:
```
this->makeUserMessages
```

Base/QTGUI/qSlicerSaveDataDialog.cxx
```diff
@@ -1296,6 +1402,17 @@ void qSlicerSaveDataDialogPrivate::updateOptionsWidget(int row)
   this->FileWidget->resizeColumnToContents(OptionsColumn);
 }
 
+//-----------------------------------------------------------------------------
+void qSlicerSaveDataDialogPrivate
+::updateUserMessagesItem(int row)
+{
+vtkMRMLStorableNode * const node = vtkMRMLStorableNode::SafeDownCast(this->object(row));
```
jcfr, line 1409: Adding the extra `const` is not something we do consistently ... I suggest to remove it

Libs/MRML/Core/vtkMRMLStorageNode.cxx
```diff
@@ -395,6 +431,20 @@ void vtkMRMLStorageNode::PrintSelf(ostream& os, vtkIndent indent)
     (this->GetDefaultWriteFileExtension() ? this->GetDefaultWriteFileExtension() : "(none)") << "\n";
 
   os << indent << "TempFileName: " << (this->TempFileName ? this->TempFileName : "(none)") << "\n";
+  if (this->UserMessages)
+    {
+    os << indent << "UserMessages: " << &this->UserMessages << "\n";
+    for (int i = 0; i < this->UserMessages->GetNumberOfMessages(); i++)
+      {
+      vtkMessageCollection::Message message = this->UserMessages->GetMessage(i);
+      os << indent << "UserMessagesMember: (" << message.MessageType << ") "
```
jcfr, line 440: To be consistent with how `FileListMember` and `URIListMember` are displayed, what about removing the parenthesis ?

Base/QTGUI/qSlicerSaveDataDialog.cxx
```diff
@@ -1296,6 +1402,17 @@ void qSlicerSaveDataDialogPrivate::updateOptionsWidget(int row)
...
+  QString allUserMessages = makeUserMessages(node);
+  QTableWidgetItem* const userMessagesItem = this->FileWidget->item(row, UserMessagesColumn);
+  Q_ASSERT(userMessagesItem);
```
jcfr, line 1412: Remove assert, if we expect such error ... a proper warning should be reported to the user.

Libs/MRML/Core/vtkMRMLStorageNode.h
```diff
@@ -26,6 +27,77 @@ class vtkStringArray;
 
 // STD includes
 #include <vector>
+#include <algorithm>
+
+///
+/// A class for recording warnings and errors associated with this
+/// vtkMRMLStorableNode.
+class VTK_MRML_EXPORT vtkMessageCollection : public vtkObject
```
jcfr, line 35: Since this is added to the public API, I suggest to move this into its own file.

Likely a file to be added to https://github.com/Slicer/vtkAddon

Libs/MRML/Core/vtkMRMLStorageNode.h
```diff
@@ -26,6 +27,77 @@ class vtkStringArray;
...
+{
+public:
+  vtkTypeMacro(vtkMessageCollection,vtkObject);
+  void PrintSelf(ostream& os, vtkIndent indent) override;
+  static vtkMessageCollection* New();
+
+  ///
+  /// A helper class that describes a single message
+  struct Message
```
jcfr, line 44: This should  probably be maintained in the private implementation.

Libs/MRML/Core/vtkMRMLStorageNode.h
```diff
@@ -26,6 +27,77 @@ class vtkStringArray;
...
+  /// type
+  virtual int GetNumberOfMessagesOfType(const char *eventName) const
+    { return GetNumberOfMessagesOfType(vtkCommand::GetEventIdFromString(eventName)); }
+
+  // Get the nth message from the message vector
+  virtual Message GetMessage(int index) const
+    { return this->MessageVector.at(index); }
+
+  // Append a message to the message vector
+  virtual void AddMessage(const Message &message)
```
jcfr, line 87: `AddMessage(unsigned long type, const std::string& text)`

Libs/MRML/Core/vtkMRMLStorageNode.h
```diff
@@ -26,6 +27,77 @@ class vtkStringArray;
...
+  // Clear the message vector
+  virtual void ClearMessages() { this->MessageVector.clear(); }
+
+protected:
+  vtkMessageCollection();
+  ~vtkMessageCollection() override;
+  vtkMessageCollection(const vtkMessageCollection&);
+  void operator=(const vtkMessageCollection&);
+
+  std::vector<Message> MessageVector;
```
jcfr, line 99: `Messages`

Libs/MRML/Core/vtkMRMLStorageNode.h
lassoan: It would be useful to be able to quickly distinguish between warning (save was successful but there may be some limitations) and error messages (save failed). Displaying the number of messages can be informative, too, but it is not essential - I agree that we could just display the error or warning icon, without any text).

I feel, too, that this is a bit too much infrastructure for implement this little feature. However, if we consider generalizing and expanding this in the future and use it in other places (for example using it for collecting messages while running CLI modules or other processing), or it is planned to be moved to VTK proper, then it might be justified.

> I am changing the name of latter to vtkMessageCollection

If we agree to keep this class then it should be renamed to something more specific. 1. It is in MRML library, so it should be vtkMRML... (if we plan to move it to VTK then we don't need the MRML prefix but then it would need to be moved to vtkAddon, but for this it would be nice to ask VTK core people if they were interested in adding vtkLogMessage and vtkLogMessageCollection classes to VTK). 2. Since there are many kinds of messages, a more specific name would be better something like vtkMRMLLogMessageCollection.

In the long term, the best would be to add a vtkMessage class to VTK, which could store message information (text, type, group name, class name, file name, line number, etc.) and VTK logging macros would use them as event data in warning/error events. This way VTK would not need to create a formatted string from all the input information and applications would not need to parse this string to extract relevant information (making logging and interpretation of logs simpler and faster).

pieper: I agree it would be a valuable and easy fix to VTK to add a log management feature like this - all the log messages go through the macros anyway, so you could also keep information about the vtkObject that generated the log message.  In this case it would make sense to integrate with the sink code to intercept output from helper libraries too.  It would be cool also to use that for CLIs but also modules could use that mechanism to create logs associated with workflows in a structured way.  Obviously that's a more ambitious project than Lee signed up for.

Base/QTGUI/qSlicerSaveDataDialog.cxx
Leengit: Will do.

Base/QTGUI/qSlicerSaveDataDialog.cxx
Leengit: I think of this `const` as a unit test to prevent a regression.  If I try to modify my local variable `node`, I want the compiler to let me know.

Libs/MRML/Core/vtkMRMLStorageNode.cxx
Leengit: Will do.

Base/QTGUI/qSlicerSaveDataDialog.cxx
Leengit: I don't expect that a user could cause such an error; but a pull request could.  I will remove it.

Libs/MRML/Core/vtkMRMLStorageNode.h
Leengit: Since `vtkAddon` is not part of `github.com/Slicer/Slicer`, might I leave `vtkMessageCollection` where it is until I can get an implementation into and released from `vtkAddon`?  Perhaps I should rename the current one as `vtkMRMLMessageCollection` so that when `vtkMessageCollection` is released from vtkAddon there won't be a naming conflict ... and then I can easily change the calls from `vtkMRMLStorageNode`, etc. to use the `vtkAddon` implementation and `vtkMessageCollection` naming.

Libs/MRML/Core/vtkMRMLStorageNode.h
Leengit: I am not sure that I understand.  `vtkMessageCollection::Message` is a return value for a public method and would need to be public.

Libs/MRML/Core/vtkMRMLStorageNode.h
Leengit: Will add.

Libs/MRML/Core/vtkMRMLStorageNode.h
Leengit: Will do.

Libs/MRML/Core/vtkMRMLStorageNode.h
Leengit: I will do the renaming to `vtkMRMLMessageCollection`.

Leengit:
As requested, the commits have been merged into a single commit, https://github.com/Slicer/Slicer/pull/5045/commits/9fc7a42f2cc2358c682633d574b145fc42e51ad4.

Leengit:
Here is a screenshot of what the Save Data Dialog looks like after one approves the file overwriting of every other file.
Screenshot from 2020-07-27 18-33-22

pieper:
Looks pretty good, but it would be nicer to use some kind of icons, like a green check vs a yellow triangle and red octagon and then have the detail text in a tooltip or other popup.  Some of the error/warning messages can be a few lines long.

lassoan:
Thanks for the screenshot. Messages can be several sentences long and multiple lines. We can display 1-2 icons and maybe a few words and show details in tooltip or popup window.

Leengit:
I will implement the idea of icons with the longer text in tooltips.

Base/QTGUI/qSlicerSaveDataDialog.cxx
Leengit: At today's Slicer Hangout we discussed trying the approach of showing icons for warnings and errors, with a tool tip having the more complete message.

Leengit:
I just checked in a version that uses icons with Tool Tips instead of text in the User Messages column of the Save Data Dialog.  Feedback would be appreciated.  To do items that I see are:

1. Some icons are too big.  I need to figure out how to shrink them or find better sized ones.  Any thoughts?
1. I am thinking to remove the green check marks that appear if there have been no warnings or errors _yet_.  My worry is that the green check marks will be misinterpreted as meaning that the save has _already_ been successful.  Any thoughts?
1. I will put the `vtkMRMLMessageCollection` code in new files named `vtkMRMLMessageCollection.h` and `vtkMRMLMessageCollection.cxx`.

Here is what the Save Dialog box looks like if I select "No to all" for replacing files.  Hovering the mouse cursor over one of the icon gives the message text as a tool tip ... though I couldn't get my screen capture to show that.

Screenshot from 2020-07-29 09-12-22

pieper:
How about just setting the background and text color of the user messages cells?  Green with no text, yellow with "Warning" in black text, and red with "Error" in white text.

Leengit:
> How about just setting the background and text color of the user messages cells? Green with no text, yellow with "Warning" in black text, and red with "Error" in white text.

If there is at least one warning and at least one error that could get tricky with the background.    Besides, I like icons better.  :-)

Leengit:
I just pushed what I hope is the final version for this pull request.  Please review.  Rather than allow feature-creep to keep this open longer, let's submit any improvements as a new Github issue.

jcfr:
Thanks @Leengit  :+1: 

Could you include here a screenshot ?  :pray:

Leengit:
Screenshot from 2020-07-29 13-44-29

lassoan:
The column is very wide for just showing icons. Maybe we could just add an icon to the first column (by setting the icon in DecorationRole for the item), or removing the text from the column header (or replace the text with an icon).

Leengit:
With this commit just now, I have narrowed the column width.

Leengit:
@jcfr, @pieper, if your "changes requested" reviews have been satisfied would you indicate that at GitHub?

jamesobutler:
I think the intended commit title of "ENH: Report detailed error message about why node save failed" has been lost.  It is currently just titled "Squashed commits:"

Leengit:
> I think the intended commit title of "ENH: Report detailed error message about why node save failed" has been lost. It is currently just titled "Squashed commits:"

Thanks.  Fixed with force commit https://github.com/Slicer/Slicer/commit/fb0426cb87611f9ea5905b5401b50fa40313990d.

Libs/MRML/Core/vtkMRMLStorageNode.cxx
```diff
@@ -395,6 +406,20 @@ void vtkMRMLStorageNode::PrintSelf(ostream& os, vtkIndent indent)
     (this->GetDefaultWriteFileExtension() ? this->GetDefaultWriteFileExtension() : "(none)") << "\n";
 
   os << indent << "TempFileName: " << (this->TempFileName ? this->TempFileName : "(none)") << "\n";
+  if (this->UserMessages)
+    {
+    os << indent << "UserMessages: " << &this->UserMessages << "\n";
```
lassoan, line 411: just call PrintSelf of UserMessges

Leengit:
@lassoan: I would like to get this into tonight's nightly build so that I can use tomorrow to make any needed fixes.  If you have any last minute comments, they would be appreciated.  Otherwise, marking your review as approved would be appreciated.  Thanks!

lassoan:
I am currently reviewing this.

Base/QTGUI/qSlicerSaveDataDialog.cxx
```diff
@@ -709,6 +721,105 @@ ctkDirectoryButton* qSlicerSaveDataDialogPrivate::createFileDirectoryWidget(cons
...
+  static const QPixmap errorIcon = qApp->style()->standardPixmap(QStyle::SP_MessageBoxCritical).scaledToHeight(18);
+  static const QPixmap questionIcon = qApp->style()->standardPixmap(QStyle::SP_MessageBoxQuestion).scaledToHeight(18);
+
+  const vtkMRMLStorageNode * const snode = node ? node->GetStorageNode() : nullptr;
+
+  // Clear out any existing items in the userMessagesLayout
+  QLayoutItem *layoutItem = nullptr;
+  while ((layoutItem = userMessagesLayout->takeAt(0)) != nullptr)
+    {
+    delete layoutItem;
```
lassoan, line 747: Deleting items is tricky and risky. It is better to create them once and just show/hide as appropriate.
For example, if you simply delete `layoutItem` here then its parent may want to delete it again when the save dialog is destroyed (potentially causing crash during application exit).
A simpler solution could be to just show a single icon (maybe on a toolbutton) and just show/hide this button and change the icon as appropriate. Shown only one icon - if there are warning(s) and error(s) then just show a single error icon. If there are only warning(s) then show warning icon.

Base/QTGUI/qSlicerSaveDataDialog.cxx
```diff
@@ -709,6 +721,105 @@ ctkDirectoryButton* qSlicerSaveDataDialogPrivate::createFileDirectoryWidget(cons
...
+          break;
+        }
+      // Use icon in widget.  SetToolTip for widget.  Install widget in layout.
+      QLabel *messageLabel = new QLabel;
+      messageLabel->setPixmap(*icon);
+      messageLabel->setToolTip(QString(tr(message.MessageText.c_str())));
+      userMessagesLayout->addWidget(messageLabel);
+      }
+    }
+  // else
```
lassoan, line 780: Remove commented out code.

jcfr:
Thanks @lassoan 

@Leengit  Could you update the commit message writing a text explaining the problem solved, the approach and relevant technical details ? 

Instead of 

```
ENH: Changing type and variable names according to feedback.
BUG: Reverting const, Self, this->, and space adds in qSlicerSaveDataDialog.cxx.
ENH: Using UserMessages instead of Messages as member name.
ENH: Adding User Messages column to Save Data Dialog.
ENH: Using GetStorageNode().  Clearing user messages when done.
ENH: vtkMessages -> vtkMessageCollection and adding vtkObject support.
ENH: Variable and class renaming -> vtkMRMLMessageCollection.
ENH: Using icons in UserMessages column of Save Data Dialog
ENH: Implemented vtkMRMLMessageCollection.h and vtkMRMLMessageCollection.cxx
ENH: Removing file-okay icon for save data dialog; shrinking others.
ENH: Narrowing User Messages column by renaming it to Notes
```

you would have something  like 

```
This commit updates the XXX dialog to include a new column displaying saving warnings and
errors associated with XXX. It addresses issue #4968.

To support this, the storage node have been updated to include XXX, .... 

```

Base/QTGUI/qSlicerSaveDataDialog.cxx
```diff
@@ -889,15 +1012,42 @@ bool qSlicerSaveDataDialogPrivate::saveNodes()
                               QMessageBox::Yes | QMessageBox::No, QMessageBox::Yes);
       if (answer == QMessageBox::No)
         {
+        if (snode)
+          {
+          snode->GetUserMessages()->AddMessage({vtkCommand::ErrorEvent,
+              qPrintable(tr("Cannot write data file: %1. Stopping save operation at user's request.").arg(file.absoluteFilePath()))});
```
lassoan, line 1018: qPrintable is equivalent to str.toLocal8Bit().constData(), which is for writing to the console only. Application's code page is UTF-8, which may note be the same as the console. In all locations where you convert QString to std::string or char*, you need to https://doc.qt.io/qt-5/qstring.html#toStdString or https://doc.qt.io/qt-5/qstring.html#toUtf8, or use qUtf8Printable instead. Since we don't use qUtf8Printable anywhere in Slicer and there are alternatives, I would not start using it here.

Libs/MRML/Core/vtkMRMLMessageCollection.h
```diff
@@ -0,0 +1,98 @@
...
+{
+public:
+  vtkTypeMacro(vtkMRMLMessageCollection,vtkObject);
+  void PrintSelf(ostream& os, vtkIndent indent) override;
+  static vtkMRMLMessageCollection* New();
+
+  ///
+  /// A helper class that describes a single message
+  struct Message
+  {
```
lassoan, line 47: Please use current indentation style. I don't like the current style either and we'll update it soon, but until then it is better to keep the style consistent.

Libs/MRML/Core/vtkMRMLMessageCollection.h
```diff
@@ -0,0 +1,98 @@
...
+    Message(unsigned long messageType, const std::string &messageText);
+    ///
```
lassoan, line 49: Adding a function object perhaps more elegant, but it means 20+ lines of extra code to save 2-3 lines of code at a one place (in GetNumberOfMessagesOfType). If you want to be fancy then use a lambda function, then at least the code would not be much longer (or even 1-2 lines shorter) than doing a trivial iteration through the list).

Libs/MRML/Core/vtkMRMLMessageCollection.h
```diff
@@ -0,0 +1,98 @@
+/*=========================================================================
+ *
+ *  Copyright NumFOCUS
```
lassoan, line 3: Use standard Slicer header

Libs/MRML/Core/vtkMRMLMessageCollection.cxx
```diff
@@ -0,0 +1,132 @@
+/*=========================================================================
+ *
+ *  Copyright NumFOCUS
```
lassoan, line 3: Use standard Slicer header.

Libs/MRML/Core/vtkMRMLMessageCollection.h
```diff
@@ -0,0 +1,98 @@
...
+  /// type
+  virtual int GetNumberOfMessagesOfType(unsigned long messageType) const;
+
+  ///
+  /// Return the number of reported messages of the specified event
+  /// type
+  virtual int GetNumberOfMessagesOfType(const char *eventName) const;
+
+  // Get the nth message from the message vector
+  virtual Message GetMessage(int index) const;
```
lassoan, line 79: "Message" type is not Python-wrapped. Add GetNthMessageText and GetNthMessageType (or just GetMessageText, GetMessageType) methods to make this class usable from Python.

Libs/MRML/Core/vtkMRMLStorageNode.cxx
Leengit: Done.

Base/QTGUI/qSlicerSaveDataDialog.cxx
Leengit: The `takeAt(0)` command is successively removing the currently first widget from the parent.  "[The following code fragment shows a safe way to remove all items from a layout.](https://doc.qt.io/qt-5/qlayout.html#takeAt)"  So, it should all be good.

This late in the game, I'd rather not switch the functionality to showing a single icon.

Base/QTGUI/qSlicerSaveDataDialog.cxx
Leengit: Done.

Base/QTGUI/qSlicerSaveDataDialog.cxx
Leengit: Very good.  I have switched to using `.toStdString()`.

Libs/MRML/Core/vtkMRMLMessageCollection.h
Leengit: Done.

Libs/MRML/Core/vtkMRMLMessageCollection.h
Leengit: I eliminated the `EqualMessageType` function object and went with the "2-3 lines of code" in `GetNumberOfMessagesOfType`.

Libs/MRML/Core/vtkMRMLMessageCollection.h
Leengit: Done.

Libs/MRML/Core/vtkMRMLMessageCollection.cxx
Leengit: Done.

Libs/MRML/Core/vtkMRMLMessageCollection.h
Leengit: I replaced `GetMessage(int)` with `GetNthMessageType(int)` and `GetNthMessageText(int)`.

Leengit: And I moved the `struct vtkMRMLMessageCollection::Message` from public to protected.

PR #5046, opened 2020-07-17
ENH: Make exception message more user-friendly

### LINKED ISSUE #5011: Make memory allocation error message more user friendly
When Slicer runs out of memory, the displayed error message is too technical. We should display a message that is understandable for most users.

See details in this discussion: https://discourse.slicer.org/t/bad-allocation-exception/12172/5

yehan0223:
Changed the exception message to make it more understandable by
users with less technical background.

Additional system-specific suggestion on adding RAM is given only
when std::bad_alloc is captured.

Fixes #5011 

Throw bad_alloc on Linux:
throw_bad_alloc

Throw std::exception
throw_exception

### DIFF
--- a/Base/QTGUI/qSlicerApplication.cxx
+++ b/Base/QTGUI/qSlicerApplication.cxx
@@ -25,6 +25,7 @@
 #include <QtGlobal>
 #include <QMainWindow>
 #include <QSurfaceFormat>
+#include <QSysInfo>
 
 #if defined(Q_OS_WIN32)
   #include <QtPlatformHeaders\QWindowsWindowFunctions> // for setHasBorderInFullScreen
@@ -393,13 +394,29 @@ bool qSlicerApplication::notify(QObject *receiver, QEvent *event)
     {
     return QApplication::notify(receiver, event);
     }
-  catch( std::exception& exception )
+  catch( std::bad_alloc& exception )
     {
     QString errorMessage;
-    errorMessage = tr("%1 has caught an internal error.\n\n").arg(this->applicationName());
-    errorMessage += tr("You may be able to continue from this point, but results are undefined.\n\n");
-    errorMessage += tr("Suggested action is to save your work and restart.");
-    errorMessage += tr("\n\nIf you have a repeatable sequence of steps that causes this message, ");
+    errorMessage = tr("%1 has caught an application error, ").arg(this->applicationName());
+    errorMessage += tr("please save your work and restart.\n\n");
+    errorMessage += tr("The application has run out of memory. ");
+    if (!QSysInfo::kernelType().compare(tr("winnt")))
+      {
+      errorMessage += tr("Increasing virtual memory size in system settings or adding more RAM may fix this issue.\n\n");
+      }
+    else if (!QSysInfo::kernelType().compare(tr("linux")))
+      {
+      errorMessage += tr("Increasing swap size in system settings or adding more RAM may fix this issue.\n\n");
+      }
+    else if (!QSysInfo::kernelType().compare(tr("darwin")))
+      {
+      errorMessage += tr("Increasing free disk space or adding more RAM may fix this issue.\n\n");
+      }
+    else
+      {
+      errorMessage += tr("Adding more RAM may fix this issue.\n\n");
+      }
+    errorMessage += tr("If you have a repeatable sequence of steps that causes this message, ");
     errorMessage += tr("please report the issue following instructions available at http://slicer.org\n\n\n");
     errorMessage += tr("The message detail is:\n\n");
     errorMessage += tr("Exception thrown in event: ") + exception.what();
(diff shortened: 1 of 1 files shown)


Base/QTGUI/qSlicerApplication.cxx
```diff
@@ -396,21 +396,24 @@ bool qSlicerApplication::notify(QObject *receiver, QEvent *event)
...
-    errorMessage += tr("\n\nIf you have a repeatable sequence of steps that causes this message, ");
+    errorMessage = tr("%1 has caught an exception, ").arg(this->applicationName());
+    errorMessage += tr("please save your work and restart.\n\n");
+    errorMessage += tr("If you have a repeatable sequence of steps that causes this message, ");
     errorMessage += tr("please report the issue following instructions available at http://slicer.org\n\n\n");
     errorMessage += tr("The message detail is:\n\n");
     errorMessage += tr("Exception thrown in event: ") + exception.what();
+    if (!tr("bad allocation").compare(exception.what()) || !tr("std::bad_alloc").compare(exception.what()))
+      {
+        errorMessage += tr("\n\nIncreasing system memory may resolve this issue\n\n");
```
lassoan, line 407: Most users stop reading after a couple of words, so it would be better to start with this specific message and add the general stuff after that.

I think we could also improve the memory allocation message:
"The application has run out of memory." and then
- For Windows "Increasing virtual memory size in system settings or adding more RAM may fix this issue".
- For Linux "Increasing swap size in system settings or adding more RAM may fix this issue".
- For Mac "Increasing free disk space or adding more RAM may fix this issue".

Base/QTGUI/qSlicerApplication.cxx
```diff
@@ -396,21 +396,24 @@ bool qSlicerApplication::notify(QObject *receiver, QEvent *event)
...
-    errorMessage += tr("You may be able to continue from this point, but results are undefined.\n\n");
-    errorMessage += tr("Suggested action is to save your work and restart.");
-    errorMessage += tr("\n\nIf you have a repeatable sequence of steps that causes this message, ");
+    errorMessage = tr("%1 has caught an exception, ").arg(this->applicationName());
+    errorMessage += tr("please save your work and restart.\n\n");
+    errorMessage += tr("If you have a repeatable sequence of steps that causes this message, ");
     errorMessage += tr("please report the issue following instructions available at http://slicer.org\n\n\n");
     errorMessage += tr("The message detail is:\n\n");
     errorMessage += tr("Exception thrown in event: ") + exception.what();
+    if (!tr("bad allocation").compare(exception.what()) || !tr("std::bad_alloc").compare(exception.what()))
```
pieper, line 405: Here it would be better to compare the exception type directly, rather than the exception.what() string, which may be different on depending on the platform (especially it may not match the localized string from tr).

https://en.cppreference.com/w/cpp/memory/new/bad_alloc

Base/QTGUI/qSlicerApplication.cxx
```diff
@@ -396,21 +396,24 @@ bool qSlicerApplication::notify(QObject *receiver, QEvent *event)
...
+      }
     qCritical() << errorMessage;
     if (this->commandOptions()->isTestingEnabled())
       {
       this->exit(EXIT_FAILURE);
       }
     else
       {
-      QMessageBox::critical(this->mainWindow(),tr("Internal Error"), errorMessage);
+      QMessageBox::critical(this->mainWindow(),tr("System Exception"), errorMessage);
```
lassoan, line 416: Users don't know what "Exception" means. I think "Internal error" was not too bad - maybe not entirely fair, as Slicer kind of took the blame for something that might have been caused by a bad workflow. Maybe we can change it to "Application error"?

yehan0223: Subject and message body were changed to "Application error".

Base/QTGUI/qSlicerApplication.cxx
yehan0223: System kernels are checked before giving the additional suggestion.

Base/QTGUI/qSlicerApplication.cxx
yehan0223: I changed the method of checking exception types. The new method should be able to capture the target exception type robustly.

I did not find a good way to compare the exception type directly. The cppreference uses std::strcmp(what(), other.what()) for comparing exception types in all the exception pages. I avoided using cstring as an additional lib. Please let me know if there are any good resources on such type comparison topic.

yehan0223:
Hi Steve and Andras, the commit has been updated. Please take another pass and let me your thoughts. I am new to Slicer and Github, and this is the first issue I ever work on. So feel free to let me know if there is any thing I need to pay attention to in my commit. Thanks.

Base/QTGUI/qSlicerApplication.cxx
```diff
@@ -396,10 +397,29 @@ bool qSlicerApplication::notify(QObject *receiver, QEvent *event)
   catch( std::exception& exception )
     {
     QString errorMessage;
-    errorMessage = tr("%1 has caught an internal error.\n\n").arg(this->applicationName());
-    errorMessage += tr("You may be able to continue from this point, but results are undefined.\n\n");
-    errorMessage += tr("Suggested action is to save your work and restart.");
-    errorMessage += tr("\n\nIf you have a repeatable sequence of steps that causes this message, ");
+    errorMessage = tr("%1 has caught an application error, ").arg(this->applicationName());
+    errorMessage += tr("please save your work and restart.\n\n");
+    if (!tr(exception.what()).compare(std::bad_alloc().what()))
```
pieper, line 402: Comparing the strings would probably work, but it's not really the right thing to do.  Instead we should provide a specific catch body for std::bad_alloc and then a generic catch(...) body.

I didn't know exactly how this worked either, but I looked it up and wrote this for testing.  If you change around the order of the statements in the body of the try block you can see how it catches the various exceptions.

```c++
#include <exception>
#include <iostream>

int main()
{
  char *buffer = nullptr;
  try {
    unsigned long billion = 1000 * 1000 * 1000;
    buffer = new char[100000 * billion ];
    std::exception std_exception;
    throw 0;
    throw std_exception;
  } catch (std::bad_alloc e) {
    std::cerr << "caught bad alloc" << std::endl;
    std::cerr << e.what() << std::endl;
    exit(1);
  } catch (std::exception e) {
    std::cerr << "generic exception" << std::endl;
    std::cerr << e.what() << std::endl;
    exit(2);
  } catch (...) {
    std::cerr << "something odd happened" << std::endl;
    exit(2);
  }
  delete [] buffer;
  std::cout << "no problems" << std::endl;
  exit(0);
}
```

yehan0223: Hi Steve, thanks for your comment. I've added the specific catch for std::bad_alloc before the generic catch as suggested. Some repeated codes are generated due to the addition, but I think the current form maintains good readability. I didn't add the last "catch ( ... )" as in the example since other modules may deliberately throw event for their own purpose.  Let me know what you think. Thanks!

lassoan:
It would be slightly better translate (tr) the entire message instead of line-by-line, but I'm also OK with keeping this as is (and improve it when we actually get to translate this text).

pieper:
Thanks for the contribution!

yehan0223:
Thank you both for the constructive reviews!

PR #5047, opened 2020-07-19
BUG: Fix segment editor masking for Set-type effects

### LINKED ISSUE #4984: Segment editor masking does not prevent deletion
Masking in Segment Editor correctly prevents painting outside the mask, but does not prevent deleting outside the mask.

Example:

1. Initial state:

image

Editable area is "inside mask" (mask is the yellow segment).

https://1drv.ms/u/s!Arm_AFxB9yqHu7UbAvQQixgpzI0RFA?e=eEat8g

2. Margin grow segment 1 (green) by 6mm -> correct result:

image

3. Undo growing and now shrink by 6mm -> ERROR! shrinking is not restricted to the masked region (places where the segmentation changed but should not have are marked with red arrows)

image

Sunderlandkyl:
Segment editor effects that used ModificationModeSet (Margin, Threshold, AutoComplete, etc.) would adhere to the mask when adding new regions, but would not prevent the deletion of regions within the mask.

Fixed by:
- Not always including the edited segment in the mask, so that we can accurately determine actual mask region
- Adding the regions of the segment labelmap outside of the mask to the modifier labelmap

fixes #4984

### DIFF
--- a/Libs/MRML/Core/vtkMRMLSegmentationNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLSegmentationNode.cxx
@@ -655,22 +655,6 @@ bool vtkMRMLSegmentationNode::GenerateEditMask(vtkOrientedImageData* maskImage, 
     return false;
     }
 
-  // Always allow paint inside edited segment
-  if (paintInsideSegments)
-    {
-    // include edited segment in "inside" mask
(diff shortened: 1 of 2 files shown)


lassoan:
Great, thanks a lot!

Sunderlandkyl:
Sorry for the noise. This PR is superceded by https://github.com/Slicer/Slicer/pull/5048.

PR #5052, opened 2020-07-21
ENH: Make volume centering apply a transform in Volumes module

### LINKED ISSUE #4952: Make volume centering apply a transform
Center volume function (in Volumes module and in Add data/Center option) hardens the centering transform on the volume. It should apply the centering transform but should not harden it so that no information is discarded, the same transform can be applied to other nodes, and the transformation can be undone.

See more information in this discussion: https://discourse.slicer.org/t/volume-rendering-not-centered/11735

yehan0223:
Make the center volume function in Volumes module apply a transform
instead of hardening the centering transform directly, so that no
information is discarded.

The created transform is a linear transform named CenteringTransform
on the volume node in Data module.

Fixes #4952

before centering
before_centering

after centering
after_centering

It's also suggested to do the same for the Add data/center option, the actual execution of reading and centering the input data is done in a separate place. I traced that functional from qSlicerIO to  qSlicerVolumesReader, vtkSlicerVolumeLogic::AddArchetypeVolume and storageNode->SetCenterImage(options & vtkSlicerVolumesLogic::CenterImage), which is used in many other places. My concern is that changing the loading functions may have other impacts. So if possible let's keep it in the current form. If needed, we can create another issue for centering option during adding files.

All in all, this pull request fixes the issue described in #4952 subject.

### DIFF
--- a/Libs/MRML/Widgets/qMRMLVolumeInfoWidget.cxx
+++ b/Libs/MRML/Widgets/qMRMLVolumeInfoWidget.cxx
@@ -29,6 +29,7 @@
 #include "ui_qMRMLVolumeInfoWidget.h"
 
 // MRML includes
+#include <vtkMRMLLinearTransformNode.h>
 #include <vtkMRMLScene.h>
 #include <vtkMRMLScalarVolumeNode.h>
 #include <vtkMRMLScalarVolumeDisplayNode.h>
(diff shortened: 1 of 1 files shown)


Libs/MRML/Widgets/qMRMLVolumeInfoWidget.cxx
```diff
@@ -368,12 +369,22 @@ bool qMRMLVolumeInfoWidget::isCentered()const
...
     }
-  d->VolumeNode->SetOrigin(origin);
+  double volumeOrigin[3];
+  d->VolumeNode->GetOrigin(volumeOrigin);
+  vtkNew<vtkMatrix4x4> transformMatrix;
+  vtkNew<vtkMRMLLinearTransformNode> centeringTransform;
+  transformMatrix->SetElement(0, 3, centerOrigin[0]-volumeOrigin[0]);
+  transformMatrix->SetElement(1, 3, centerOrigin[1]-volumeOrigin[1]);
+  transformMatrix->SetElement(2, 3, centerOrigin[2]-volumeOrigin[2]);
+  centeringTransform->SetName("CenteringTransform");
```
jcfr, line 384: I suggest to name the transform by including the volume name.

instead of `CenteringTransform` it would be `<volumeNodeName>-CenteringTransform` (e.g `CTChest-CenteringTransform`)

yehan0223: That's a good point, it improves the clarity since no volume node would have the same name as the others'. I've made the changes. 
renameTransform

yehan0223:
@jcfr Can we merge this commit and close the issue?

lassoan:
I'll merge this.

jcfr:
Thanks @lassoan

yehan0223:
Thanks! @lassoan  @jcfr  @pieper

lassoan:
Merged in https://github.com/Slicer/Slicer/commit/422bc4fd0363971bf35703e0e60cfa78222572b7

Moved the logic to volume node so that it can be used without GUI (it would have been a bit nicer to add it to Volumes module logic, but that would not have been accessible by the MRML widget).

PR #5060, opened 2020-07-24
BUG: Fix Segment statistics ignores parent transform

### LINKED ISSUE #4942: Segment statistics ignores parent transform
Segment statistics module silently ignores parent transform, which may lead to unexpected results.

This could be fixed by cloning segmentation to a temporary node, hardening the transform, and computing everything on that. 

Transformation of master volume should be taken into account, too (but probably that is done already).

Sunderlandkyl:
In order to ensure that the segment statistics are calculated in world coordinates, the segmentation is duplicated to a temporary segmentation node and the transform is hardened.

fixes #4942

### DIFF
--- a/Modules/Scripted/SegmentStatistics/SegmentStatistics.py
+++ b/Modules/Scripted/SegmentStatistics/SegmentStatistics.py
@@ -182,28 +182,34 @@ class SegmentStatisticsWidget(ScriptedLoadableModuleWidget):
   def onApply(self):
     """Calculate the label statistics
     """
-    if not self.outputTableSelector.currentNode():
-      newTable = slicer.mrmlScene.AddNewNodeByClass("vtkMRMLTableNode")
-      self.outputTableSelector.setCurrentNode(newTable)
-    # Lock GUI
-    self.applyButton.text = "Working..."
-    self.applyButton.setEnabled(False)
-    slicer.app.processEvents()
-    # set up parameters for computation
-    self.logic.getParameterNode().SetParameter("Segmentation", self.segmentationSelector.currentNode().GetID())
-    if self.scalarSelector.currentNode():
-      self.logic.getParameterNode().SetParameter("ScalarVolume", self.scalarSelector.currentNode().GetID())
-    else:
-      self.logic.getParameterNode().UnsetParameter("ScalarVolume")
-    self.logic.getParameterNode().SetParameter("MeasurementsTable", self.outputTableSelector.currentNode().GetID())
-    # Compute statistics
-    self.logic.computeStatistics()
-    self.logic.exportToTable(self.outputTableSelector.currentNode())
-    # Unlock GUI
-    self.applyButton.setEnabled(True)
-    self.applyButton.text = "Apply"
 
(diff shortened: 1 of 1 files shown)


Modules/Scripted/SegmentStatistics/SegmentStatistics.py
```diff
@@ -423,6 +423,17 @@ def computeStatistics(self):
...
+    if not segmentationNode.GetParentTransformNode() is None:
+      # Create a temporary segmentation and harden the transform to ensure that the statistics are calculated
+      # in world coordinates
+      transformedSegmentationNode = slicer.vtkMRMLSegmentationNode()
+      transformedSegmentationNode.Copy(segmentationNode)
+      transformedSegmentationNode.HideFromEditorsOn()
+      slicer.mrmlScene.AddNode(transformedSegmentationNode)
+      transformedSegmentationNode.HardenTransform()
+      self.getParameterNode().SetParameter("Segmentation", transformedSegmentationNode.GetID())
+
```
lassoan, line 436: add a try/except block to make sure that we restore the original segmentation node even if an exception is thrown

Sunderlandkyl: OK, done!

Modules/Scripted/SegmentStatistics/SegmentStatistics.py
```diff
@@ -422,20 +422,38 @@ def computeStatistics(self):
...
+        segmentationNode.GetSegmentation().GetSegmentIDs(visibleSegmentIds)
+      if visibleSegmentIds.GetNumberOfValues() == 0:
+        logging.debug("computeStatistics will not return any results: there are no visible segments")
+
+      # update statistics for all segment IDs
+      for segmentIndex in range(visibleSegmentIds.GetNumberOfValues()):
+        segmentID = visibleSegmentIds.GetValue(segmentIndex)
+        self.updateStatisticsForSegment(segmentID)
+    except:
+      logging.error("computeStatistics: Error when calculating statistics")
```
lassoan, line 451: This is good in the sense that now the parameter node remains consistent even if an error occurs. The only not nice thing is that now we swallow the error: we just log the error, no stack trace, and the caller method and the user has no way of knowing that something is wrong (other than no output is generated).

It would be nicer not to catch the exception or re-raise it (the simplest I think is to remove the "except:" branch and just keep "try" and "finally"). Try/except should be also added in the GUI to catch the error, display it to the user, and leave the module in a consistent state (similarly how it is done in current module template https://github.com/Slicer/Slicer/blob/master/Utilities/Templates/Modules/Scripted/TemplateKey.py#L270-L285). Now it does not work well - for example if an exception is thrown then the apply button gets stuck.

Sorry for just keep asking more but this part of the module was just not implemented well and it would be nice to fix it now that we touch it anyway. Thank you!

Sunderlandkyl: Ok, thanks! Added the try/except block to the onApply function.

PR #5064, opened 2020-07-28
BUG: Add support for downloading PyPI package linux aarch64

jamesobutler:
This was introduced by 111b131e733b10f2f7fbcb8f5d4284e1281cf96d where Pillow was updated from 7.0.0 to 7.2.0.  The new version now has both a manylinux1 and manylinux2014 option.

This should address @adamrankin 's comment at https://github.com/Slicer/Slicer/commit/111b131e733b10f2f7fbcb8f5d4284e1281cf96d#commitcomment-40980312

### DIFF
--- a/SuperBuild/External_python-dicom-requirements.cmake
+++ b/SuperBuild/External_python-dicom-requirements.cmake
@@ -45,9 +45,11 @@ if(NOT Slicer_USE_SYSTEM_${proj})
   #  - Pillow-7.2.0-cp36-cp36m-win_amd64.whl
   #  - Pillow-7.2.0-cp36-cp36m-macosx_10_10_x86_64.whl
   #  - Pillow-7.2.0-cp36-cp36m-manylinux1_x86_64.whl
+  #  - Pillow-7.2.0-cp36-cp36m-manylinux2014_aarch64.whl
   pillow==7.2.0 --hash=sha256:ffe538682dc19cc542ae7c3e504fdf54ca7f86fb8a135e59dd6bc8627eae6cce \
                 --hash=sha256:ec29604081f10f16a7aea809ad42e27764188fc258b02259a03a8ff7ded3808d \
-                --hash=sha256:0a80dd307a5d8440b0a08bd7b81617e04d870e40a3e46a32d9c246e54705e86f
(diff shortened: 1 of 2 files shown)


adamrankin:
also this from scipy:

066c513d90eb3fd7567a9e150828d39111ebd88d3e924cdfc9f8ce19ab6f90c9

jamesobutler:
What system are you on? That hash for scipy is the hash for the source file and not a wheel.  See https://pypi.org/project/scipy/#copy-hash-modal-cd35c84c-60a5-4834-be4d-c8b05bf5592e

adamrankin:
Linux petergrp-t4.robarts.ca 4.4.154 #1 SMP Thu Mar 7 18:30:53 CST 2019
aarch64 aarch64 aarch64 GNU/Linux

jcfr:
I suggest to change the commit message from `BUG: Fix unable to match PyPI package if newer linux version` to  `BUG: Add support for downloading PyPI package linux aarch64`

adamrankin:
Using
`sudo apt-get install gfortran libopenblas-dev liblapack-dev`
allowed it to be built

jcfr:
If there wheels available, it makes sense to list the corresponding hash. That said, I would prefer we do not add the hash of the sdist by default, I suggest we append it only if the system is effectively aarch64

jamesobutler:
Note I didn't append the scipy src hash if the system is aarch64.  Unfortunately there aren't any newer manylinux wheels available for scipy at the moment and building scipy from source is a little more involved to make sure it builds correctly every time.  @adamrankin If you want to contribute more as it relates to aarch64 that would be helpful, or you can just wait for scipy to provide a wheel and then add the hash.

jcfr:
> Note I didn't append the scipy src hash if the system is aarch64.

Makes sense. We shouldn't include hash of source distribution by default.

> building scipy from source is a little more involved to make sure it builds correctly every time.

We could introduce a global option called `Slicer_ALLOW_PYTHON_PACKAGE_INSTALL_WITH_SDIST_FALLBACK`  (I couldn't come up with a better name) ... if specified the hash of the sdist would be included as a possible alternative.

The caveats is that this would introduce additional maintenance burden ... but would help in the context of aarch64 building.

PR #5065, opened 2020-07-28
COMP: disable parameter serializer by default

pieper:
This feature wasn't presently being used and it led to
multiply defined symbols in CLI builds for some
tests and extensions.

https://discourse.slicer.org/t/the-slicerdmri-extension-is-currently-unavailable/12276/21

https://github.com/pnlbwh/ukftractography/pull/130

Thanks @sjh26 for shared pain and @jcfr for suggesting the fix.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -324,7 +324,7 @@ CMAKE_DEPENDENT_OPTION(
   "Slicer_BUILD_ITKPython" OFF)
 mark_as_superbuild(Slicer_INSTALL_ITKPython)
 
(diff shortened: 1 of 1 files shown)


pieper:
Force merged without review so it'll get into tonight's night build and we already discussed it on the developer call.

jcfr:
> Force merged without review so it'll get into tonight's night build and we already discussed it on the developer call.

Makes sense :+1: 

Thanks for taking care of this, and since issues have been documented with  link to discourse thread. It would be easy for anyone to address it later on.

PR #5067, opened 2020-07-29
Improved model display

lassoan:
This pull request contains two improvements for model display:

1. Changes default model color to yellow from the old default mid gray. This make default scenes more interesting and model intersections visible over grayscale images.

2. Slightly changes inside (backface) color so that inside/outside surface can be visually distinguished. Change is specified as relative HSV offset (can be changed in Models module GUI).

**Example**

Old:

image

New:

image

### DIFF
--- a/Libs/MRML/Core/vtkMRMLDisplayNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLDisplayNode.cxx
@@ -74,9 +74,10 @@ vtkMRMLDisplayNode::vtkMRMLDisplayNode()
   this->ScalarRange[0] = 0;
   this->ScalarRange[1] = 100;
 
-  this->Color[0] = 0.5;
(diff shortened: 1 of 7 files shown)


lassoan:
I've played a bit with various offset combinations and ended up changing the hue as well because saturation did not make any difference when the color was already low in saturation. Brightness differences are a bit harder to notice because there are brightness changes in the model due to lighting anyway. For an obvious brightness difference you need to darken the color quite drastically, but then the overall color may become too dark (problem for darker initial colors and darker areas that are not orthogonal to the view plane).

All the offset values are exposed in the Models module, so everybody can experiment with different colors and parameters and see if we can come up with better defaults.

lassoan:
I'll merge this now so that everyone can experiment with this and tune the defaults as needed.

PR #5070, opened 2020-07-30
COMP: Fix -Wunused-variable warning in vtkSlicerSegmentationGeometryLogic

jcfr:
This commit fixes the following warning:

```
  /path/to/Slicer/Modules/Loadable/Segmentations/Logic/vtkSlicerSegmentationGeometryLogic.cxx:130:28: warning: unused variable ‘sourceSegmentationNode’ [-Wunused-variable]
     vtkMRMLSegmentationNode* sourceSegmentationNode = vtkMRMLSegmentationNode::SafeDownCast(this->SourceGeometryNode);
                              ^
```

### DIFF
--- a/Modules/Loadable/Segmentations/Logic/vtkSlicerSegmentationGeometryLogic.cxx
+++ b/Modules/Loadable/Segmentations/Logic/vtkSlicerSegmentationGeometryLogic.cxx
@@ -129,7 +129,8 @@ std::string vtkSlicerSegmentationGeometryLogic::CalculateOutputGeometry()
   vtkMRMLAnnotationROINode* sourceRoiNode = vtkMRMLAnnotationROINode::SafeDownCast(this->SourceGeometryNode);
   vtkMRMLSegmentationNode* sourceSegmentationNode = vtkMRMLSegmentationNode::SafeDownCast(this->SourceGeometryNode);
 
-  if (sourceVolumeNode || this->IsSourceSegmentationWithBinaryLabelmapMaster())
+  if (sourceVolumeNode
+      || (sourceSegmentationNode && this->IsSourceSegmentationWithBinaryLabelmapMaster()))
     {
     //TODO: Fractional labelmaps cannot be used yet as source, as DetermineCommonLabelmapGeometry only supports binary labelmaps
     return this->CalculateOutputGeometryFromImage();


Modules/Loadable/Segmentations/Logic/vtkSlicerSegmentationGeometryLogic.cxx
```diff
@@ -127,7 +127,6 @@ std::string vtkSlicerSegmentationGeometryLogic::CalculateOutputGeometry()
   // Determine source type
   vtkMRMLScalarVolumeNode* sourceVolumeNode = vtkMRMLScalarVolumeNode::SafeDownCast(this->SourceGeometryNode);
   vtkMRMLAnnotationROINode* sourceRoiNode = vtkMRMLAnnotationROINode::SafeDownCast(this->SourceGeometryNode);
-  vtkMRMLSegmentationNode* sourceSegmentationNode = vtkMRMLSegmentationNode::SafeDownCast(this->SourceGeometryNode);
 
   if (sourceVolumeNode || this->IsSourceSegmentationWithBinaryLabelmapMaster())
```
lassoan, line 131:
```suggestion
  if (sourceVolumeNode
    || (sourceSegmentationNode && this->IsSourceSegmentationWithBinaryLabelmapMaster()))
```

Modules/Loadable/Segmentations/Logic/vtkSlicerSegmentationGeometryLogic.cxx
```diff
@@ -127,7 +127,6 @@ std::string vtkSlicerSegmentationGeometryLogic::CalculateOutputGeometry()
   // Determine source type
   vtkMRMLScalarVolumeNode* sourceVolumeNode = vtkMRMLScalarVolumeNode::SafeDownCast(this->SourceGeometryNode);
   vtkMRMLAnnotationROINode* sourceRoiNode = vtkMRMLAnnotationROINode::SafeDownCast(this->SourceGeometryNode);
-  vtkMRMLSegmentationNode* sourceSegmentationNode = vtkMRMLSegmentationNode::SafeDownCast(this->SourceGeometryNode);
```
lassoan, line 130: this line must be kept

Modules/Loadable/Segmentations/Logic/vtkSlicerSegmentationGeometryLogic.cxx
jcfr: Thanks for the review. I updated the commit message and implemented your suggestion.

Also, since the function `IsSourceSegmentationWithBinaryLabelmapMaster` already include the check ... could you confirm that the suggested change is to improve readability of the current code and minimize introduction of issue when working on the code later on ?

https://github.com/Slicer/Slicer/blob/6751d8e6d8cdc8e7c221ed2165d902cd0c066554/Modules/Loadable/Segmentations/Logic/vtkSlicerSegmentationGeometryLogic.cxx#L349-L364

lassoan:
Thanks for the investigation. So, then the code was correct and sourceSegmentationNode check is unnecessary after all (since it is checked in IsSourceSegmentationWithBinaryLabelmapMaster anyway). It is up to you - you can keep it as is, or remove the sourceSegmentationNode check to make things a bit simpler.

PR #5073, opened 2020-07-31
BUG: Make sure subject hierarchy context menu is always up-to-date

cpinter:
If a plugin was registered later than the tree view created, then the context menus from the new plugin were not available. With this change, the context menu for each tree view is re-generated if a new plugin was registered since the last context menu generation.

Re https://github.com/SlicerMorph/SlicerMorph/issues/71


cpinter:
Excellent, thanks for the super quick merge!

PR #5074, opened 2020-07-31
COMP: fix signed/unsigned comparison introduced by pull request 5045.

Leengit:
Fixes warning for comparison of signed and unsigned integer.

### DIFF
--- a/Libs/MRML/Core/vtkMRMLMessageCollection.cxx
+++ b/Libs/MRML/Core/vtkMRMLMessageCollection.cxx
@@ -63,9 +63,12 @@ vtkMRMLMessageCollection
 ::GetNumberOfMessagesOfType(unsigned long messageType) const
 {
   int response = 0;
-  for (int i = 0; i < this->Messages.size(); ++i)
+  for (int i = 0; i < static_cast<int>(this->Messages.size()); ++i)
     {
-    response += static_cast<int>(GetNthMessageType(i) == messageType);
+    if (this->GetNthMessageType(i) == messageType)
+      {
+      ++response;
+      }
     }
   return response;
 }


Libs/MRML/Core/vtkMRMLMessageCollection.cxx
```diff
@@ -63,7 +63,8 @@ vtkMRMLMessageCollection
 ::GetNumberOfMessagesOfType(unsigned long messageType) const
 {
   int response = 0;
-  for (int i = 0; i < this->Messages.size(); ++i)
+  const unsigned numberOfMessages = static_cast<unsigned>(this->Messages.size());
+  for (unsigned i = 0; i < numberOfMessages; ++i)
     {
     response += static_cast<int>(GetNthMessageType(i) == messageType);
```
lassoan, line 69: Bool should not be casted to int. You can use the ternary operator or a simple `if` to increment by 1.

jcfr: I suggest something like this:

```cpp
  int response = 0;
  for (int i = 0; i < static_cast<int>(this->Messages.size()); ++i)
    {
    if (this->GetNthMessageType(i) == messageType)
      {
      response++;
      }
    }
  return response;
```

jcfr:
Consider also using a more descriptive commit message. Looking at the commit history may help.

Libs/MRML/Core/vtkMRMLMessageCollection.cxx
Leengit: Will do simple `if`.

jcfr:
Thanks, I will integrate updating the commit message to the following:

```
COMP: Fix -Wsign-comp warning in vtkMRMLMessageCollection

This commit fixes a regression introduced in b30151532 (ENH: Display warnings
and errors in the Save Data dialog.)
```

Leengit:
> Thanks, I will integrate updating the commit message to the following:
> 
> ```
> COMP: Fix -Wsign-comp warning in vtkMRMLMessageCollection
> 
> This commit fixes a regression introduced in b30151532 (ENH: Display warnings
> and errors in the Save Data dialog.)
> ```

Great!  Thank you.

PR #5075, opened 2020-08-01
ENH: Improve display performance if TransformModifiedEvent is invoked

Sunderlandkyl:
When a TransformModified event was invoked on any node in the scene, all model display pipelines were updated. This caused a noticeable drop in performance when there were a large number of nodes in the scene.

Fixed by not setting UpdateFromMRMLRequested to true when a transform was updated, and instead calling  OnMRMLDisplayableModelNodeModifiedEvent.

The performance of OnMRMLDisplayableModelNodeModifiedEvent was also improved, as it was not necessary to call UpdateClipSlicesFromMRML, and SetUpdateFromMRMLRequested functions when only a display node was modified.

### DIFF
--- a/Libs/MRML/DisplayableManager/vtkMRMLModelDisplayableManager.cxx
+++ b/Libs/MRML/DisplayableManager/vtkMRMLModelDisplayableManager.cxx
@@ -621,12 +621,12 @@ void vtkMRMLModelDisplayableManager::ProcessMRMLNodesEvents(vtkObject *caller,
           break;
           } // else fall through
       case vtkCommand::ModifiedEvent:
-        VTK_FALLTHROUGH;
       case vtkMRMLModelNode::MeshModifiedEvent:
+      case vtkMRMLTransformableNode::TransformModifiedEvent:
         requestRender = this->OnMRMLDisplayableModelNodeModifiedEvent(displayableNode);
         break;
       default:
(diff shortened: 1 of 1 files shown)


Libs/MRML/DisplayableManager/vtkMRMLModelDisplayableManager.cxx
```diff
@@ -621,12 +621,11 @@ void vtkMRMLModelDisplayableManager::ProcessMRMLNodesEvents(vtkObject *caller,
           break;
           } // else fall through
       case vtkCommand::ModifiedEvent:
-        VTK_FALLTHROUGH;
       case vtkMRMLModelNode::MeshModifiedEvent:
+      case vtkMRMLTransformableNode::TransformModifiedEvent:
         requestRender = this->OnMRMLDisplayableModelNodeModifiedEvent(displayableNode);
         break;
       default:
-        this->SetUpdateFromMRMLRequested(true);
```
lassoan, line 629: Could you add a comment here about what kind of events we expect to have here (if any)?

Sunderlandkyl: OK. We don't expect any other types of events, so I added a comment to reflect that.

lassoan:
Perfect, thank you.

PR #5076, opened 2020-08-01
BUG: Fix crash when chaning volume rendering display nodes

lassoan:
Slicer often crashed when manipulated display nodes, for example switched between CPU/GPU rendering mode or added/removed volume rendering display nodes from scripts.

The problem was that volume rendering displayable manager used simple pointers to display nodes and volume nodes, which could become invalid, while the nodes were removed from the scene.

### DIFF
--- a/Modules/Loadable/VolumeRendering/MRMLDM/vtkMRMLVolumeRenderingDisplayableManager.cxx
+++ b/Modules/Loadable/VolumeRendering/MRMLDM/vtkMRMLVolumeRenderingDisplayableManager.cxx
@@ -88,6 +88,7 @@ public:
     }
     virtual ~Pipeline()  = default;
 
+    vtkWeakPointer<vtkMRMLVolumeRenderingDisplayNode> DisplayNode;
     vtkSmartPointer<vtkVolume> VolumeActor;
     vtkSmartPointer<vtkMatrix4x4> IJKToWorldMatrix;
   };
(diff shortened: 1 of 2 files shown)


Modules/Loadable/VolumeRendering/MRMLDM/vtkMRMLVolumeRenderingDisplayableManager.cxx
```diff
@@ -503,67 +552,86 @@ void vtkMRMLVolumeRenderingDisplayableManager::vtkInternal::AddDisplayNode(vtkMR
...
-    {
-    const PipelineMultiVolume* pipelineMulti = dynamic_cast<const PipelineMultiVolume*>(pipeline);
+    PipelineMultiVolume* pipelineMulti = dynamic_cast<PipelineMultiVolume*>(pipeline);
     if (pipelineMulti)
       {
       // Remove volume actor from multi-volume actor collection
       this->MultiVolumeMapper->RemoveInputConnection(pipelineMulti->ActorPortIndex, 0);
       this->MultiVolumeActor->RemoveVolume(pipelineMulti->ActorPortIndex);
 
+      // Decrease next actor port index
```
lassoan, line 595: @cpinter Should not we update port numbers in other pipelines (decrement the port index of those pipelines which come after the deleted actor)?

lassoan:
I've merged this now so that it gets in tonight's nightly. @cpinter it would be still useful if you could have a look when you have time. Thank you!

Modules/Loadable/VolumeRendering/MRMLDM/vtkMRMLVolumeRenderingDisplayableManager.cxx
cpinter: Yes, based on the code I think you are right.

That would mean that if I show two volumes, and then I hide the first one, then unexpected things should happen. I tested it with four volumes, showing/hiding them in lots of different ways then adjusting settings, and it seems to work. Maybe it's just a coincidence. However, what you suggest seems valid.

cpinter:
I like the simplification of the different maps by including the display node in the pipeline an simply storing vectors. I don't see anything that could be a problem.

I still don't understand why everything worked without decrementing the actor port index, but if you have tested it after adding the decrementing step and it works, then I'm OK with it.

Thank you!

Issue #4976, opened 2020-06-11
Show new Slicer Preview Release on download page when extension builds are completed


lassoan:
In certain timezones (most of Europe) there is a high chance that people don’t find any extensions for the release they have just downloaded.

It should be possible to somehow synchronize the time the new Slicer Preview Release package shows up on the download page - with the time when extension builds are complete.

Maybe it could be as simple as having an embargo period between 10pm EST and 12pm EST: packages that are uploaded during this time would all show up at the end of the period (at 12pm EST).

See this discussion: https://discourse.slicer.org/t/extension-manager-problems/11946/5?u=lassoan

lassoan:
This remains an issue that makes Slicer's reputation worse in Europe.

See yet another user report: https://discourse.slicer.org/t/modifications-to-cardiac-module-or-suggestion-to-module-for-annotation-of-geometrical-components/17917/13

Probably there are hundreds of disappointed users that simply give up instead of trying a few hours later.

jamesobutler:
To document it here, this issue was brought up again at https://discourse.slicer.org/t/no-segmenteditorextraeffects-in-v4-13/20773

lassoan:
Another complaint here:

https://discourse.slicer.org/t/extensions-for-4-13-r30577-not-available-fro-windows/21819/4

This continues to be a source of user confusion/frustration in Europe and Asia.

lassoan:
Extension builds on Slicer-4.13 seem to finish at about 7am EST. We could add a bit of margin and **pause any updates of the download site between 10pm and 9am EST**. That would ensure that users will not see incomplete releases.

@jcfr now that we can make changes to the download server more easily, could you look into making this change?

lassoan:
The current behavior is unexpected and confusing. I still need to keep explaining this issue to many users and even developers. 

@jcfr Now that we have more control over the download.slicer.org site, **could we disable any updates on download.slicer.org between 10pm and 10am EST to prevent the Slicer Preview Release download link to update before all the extensions are built**?

jcfr:
The updates of the download page currently happens twice an hour based on this rule.
See https://crontab.guru/#21,51_*_*_*_*

Following today's discussion, the provision script along with the current crontab will be updated to instead use the following rule:
https://crontab.guru/#0_10_*_*_*

![image](https://user-images.githubusercontent.com/219043/196465184-af35eb07-c8a4-431e-afb4-cfd5982dd4ec.png)

jcfr:
[...]

PR #5082, opened 2020-08-03
ENH: Update extension CDash driver script to package then test

jcfr:
This change is intended to make the Slicer extensions packages
quickly available to the user and independently of the time needed
to run all tests.


lassoan:
This change is good for robustness but makes the time period even longer when no extensions are available yet for the new nightly Preview Release. Maybe we should solve this issue first: https://github.com/Slicer/Slicer/issues/4976

jcfr:
> makes the time period even longer when no extensions are available

Agreed, I think you comment applied to https://github.com/Slicer/Slicer/pull/5081 as this one focuses on updating the Extension driver script and not the main application script.

If there are no other objection , I will go ahead and integrate this one later this afternoon.

>  Maybe we should solve this issue first: #4976

Agreed. I will follow up.

PR #5085, opened 2020-08-04
ENH: Improve markup plane widget appearance

### LINKED ISSUE #4930: Enhance markups plane widget appearance
It would be useful to display plane boundary in 3D view (similarly how it is already done in slice view, but maybe with a thin tube instead of just a line).

We would also need flags to control what to display: boundary, filling, or both. Maybe we don't need new MRML attribute but the existing EdgeVisibility and Representation properties could be used.

Sunderlandkyl:
Improves the appearance of plane widgets by adding a border around the plane and allowing control of plane fill/border opacity.

fixes #4930

image

### DIFF
--- a/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsDisplayNode.cxx
+++ b/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsDisplayNode.cxx
@@ -84,6 +84,10 @@ vtkMRMLMarkupsDisplayNode::vtkMRMLMarkupsDisplayNode()
 
   this->PropertiesLabelVisibility = true;
   this->PointLabelsVisibility = false;
+  this->FillVisibility = true;
+  this->OutlineVisibility = true;
+  this->FillOpacity = 0.5;
+  this->OutlineOpacity = 1.0;
 
   // Set active component defaults for mouse (identified by empty string)
   this->ActiveComponents[GetDefaultContextName()] = ComponentInfo();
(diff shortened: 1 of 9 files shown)


Modules/Loadable/Markups/MRML/vtkMRMLMarkupsDisplayNode.cxx
```diff
@@ -231,6 +243,10 @@ void vtkMRMLMarkupsDisplayNode::CopyContent(vtkMRMLNode* anode, bool deepCopy/*=
   vtkMRMLCopyFloatMacro(LineColorFadingSaturation);
   vtkMRMLCopyFloatMacro(LineColorFadingHueOffset);
   vtkMRMLCopyBooleanMacro(HandlesInteractive);
+  vtkMRMLCopyBooleanMacro(FillVisibility);
+  vtkMRMLCopyBooleanMacro(OutlineVisibility);
+  vtkMRMLCopyFloatMacro(FillOpacity);
+  vtkMRMLCopyFloatMacro(OutlineOpacity);
```
lassoan, line 249: Could you add it to PrintSelf method, too?

Sunderlandkyl: Done!

lassoan:
This is awesome. How difficult would it be to allow translating/rotating the plane by interacting with the border?

Sunderlandkyl:
It should be possible. What operations would you expect to be mapped to each action?

lassoan:
Great!

Since there is currently no consistent interaction style across different widgets in Slicer, we can just choose a mapping that seems the most intuitive (and maybe that is used in similar applications).

It could work similarly to Transforms module interaction widget: left-click-and-drag rotates, shift-left-click-and-drag translates; with refinements, such as rotating along one axis (the clicked axis; or when clicking near a corner then rotate along the plane normal). But it could work similarly to the slice reformat plane widget, which translates the plane when you click-and-drag it. We could also translate when the user clicks inside the plane (maybe only if interaction is enabled?).

Maybe we could go with anything that you find intuitive and then refine it based on user feedback.

finetjul:
Nice !

Btw, it would be great if the slice view "Reformat" widget get replaced by this Plane markup (no point visible, only Interaction axes widget, and no translucency)

PR #5087, opened 2020-08-06
Remove the code responsible of VERY LOW priority of SlicerAppReal (WIP)

achataigner:


achataigner:
For now this pull request is simply here to show to Andras Lasso what I've done to change SlicerAppReal process priority:
https://discourse.slicer.org/t/process-nice-priority-on-linux/6661/9

I'm just waiting for Andras feedback, I will close this pull request afterwards.

lassoan:
Thanks a lot for working on this. It would be great if you could add an environment variable (SLICER_PROCESSING_THREAD_PRIORITY) to control the behavior (similar to [this](https://github.com/Slicer/Slicer/blob/43300c7b9a3b27265c2bc736693cb881f05e3ea9/Modules/Loadable/Colors/Logic/vtkSlicerColorLogic.cxx#L68-L83)). If the variable is not specified then the current behavior should be preserved. If you specify a value (an integer value is expected) then that can be passed directly to both to the [linux](https://linux.die.net/man/2/setpriority) and [windows](https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-setthreadpriority) priority setting function.

achataigner:
Modifications done.
Tested only on Ubuntu 18.04 LTS

Docs/developer_guide/build_instructions/overview.md
```diff
@@ -38,3 +38,16 @@ Customized editions of Slicer can be generated without changing Slicer source co
 - `Slicer_EXTENSION_SOURCE_DIRS`: Defines additional extensions that will be included in the application package as built-in modules. Full paths of extension source directories has to be specified, separated by semicolons.
 
 More customization is available by using [SlicerCustomAppTemplate](https://github.com/KitwareMedical/SlicerCustomAppTemplate) project maintained by Kitware.
+
+## Runtime environment variables
```
jcfr, line 42: Thanks for adding the documentation.

lassoan: Yes, now that documentation is tightly coupled with the source code, it is feasible to describe and maintain such details in the documentation. All the work invested in the infrastructure is paying off.

jcfr: Agreed. Definitely the way to go

PR #5096, opened 2020-08-12
BUG: Fix extent calculation in IsLabelInMask

Sunderlandkyl:
The binary labelmap extent was not transformed to the mask coordinate system when calculating the resampled binary labelmap extent. Fixed by transforming the binary labelmap extent.


Sunderlandkyl:
It caused issues when collapsing labelmap layers that had different geometry, IsLabelInMask would falsely return that the labelmaps did not overlap.

PR #5100, opened 2020-08-13
ENH: Improve segment threshold histogram for floating point volumes

Sunderlandkyl:
The number of bins used for the selected region histogram was calculated with the following formula:
  ```max(scalarRangeMax - scalarRangeMin + 1, 1000)```

This worked well for integer type images, but resulted in a low resolution for floating point types.
The number of bins for floating point types is now automatically set to the maximum allowed (1000).

### DIFF
--- a/Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorThresholdEffect.py
+++ b/Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorThresholdEffect.py
@@ -813,7 +813,11 @@ class SegmentEditorThresholdEffect(AbstractScriptedSegmentEditorEffect):
     maxNumberOfBins = 1000
     masterImageData = self.scriptedEffect.masterVolumeImageData()
     scalarRange = masterImageData.GetScalarRange()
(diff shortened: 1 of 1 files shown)


Sunderlandkyl:
The histogram appearance is still not great with images that have a highly bimodal intensity distribution, since most of the interesting differences are relegated to a small number of bins on either end, while the middle is largely empty.

lassoan:
Can you post a screenshot?

Sunderlandkyl:
Sorry, I didn't mean a bimodal distribution. This behavior happens when there is a large scalar range, with most of the intensities concentrated in a few small areas.

This is a T1/T2 image:
image

Most of the values are under 5.0, while the total intensity range is 100.0. The histogram is accurate, however I could imagine that there would be an issue if the scalar range was much larger.

Sunderlandkyl:
Probably the solution would be to add controls to adjust the minimum and maximum range of the histogram so that you could zoom in on any particular range.

lassoan:
I think this is OK, but if we really want to make this nicer then we can show the histogram between 1st at 99th percentile + some fixed margin. The same way as it is done in automatic window/level. We could even try to use the current window/level to set the min/max range of the histogram.

PR #5104, opened 2020-08-15
BUG: Fix running of multiple Slicer instances

lassoan:
When started multiple Slicer instances exactly at the same time, only one log file was created and MRB file loading was failed due to insufficient timestamp (1 second precision) and random number generation with uninitialized seed.

Fixed by adding a 3-digit random counter to log file name and switching to C++ random number generators.


che85:
I just tested and it works! Thanks for adding this!

PR #5107, opened 2020-08-17
BUG: Fix model slice display crash when lookup table node is null

Sunderlandkyl:
Loading a scene that used an invalid LUT for the model scalar overlay, would cause a crash.
Fixed by checking that the LUT exists before attempting to access it.

### DIFF
--- a/Libs/MRML/DisplayableManager/vtkMRMLModelSliceDisplayableManager.cxx
+++ b/Libs/MRML/DisplayableManager/vtkMRMLModelSliceDisplayableManager.cxx
@@ -605,7 +605,10 @@ void vtkMRMLModelSliceDisplayableManager::vtkInternal
         // that lookup table original range.
         vtkSmartPointer<vtkLookupTable> dNodeLUT = vtkSmartPointer<vtkLookupTable>::Take(displayNode->GetColorNode() ?
           displayNode->GetColorNode()->CreateLookupTableCopy() : nullptr);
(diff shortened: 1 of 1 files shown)


Libs/MRML/DisplayableManager/vtkMRMLModelSliceDisplayableManager.cxx
```diff
@@ -605,8 +605,11 @@ ::UpdateDisplayNodePipeline(vtkMRMLDisplayNode* displayNode, const Pipeline* pip
         // that lookup table original range.
         vtkSmartPointer<vtkLookupTable> dNodeLUT = vtkSmartPointer<vtkLookupTable>::Take(displayNode->GetColorNode() ?
           displayNode->GetColorNode()->CreateLookupTableCopy() : nullptr);
-        dNodeLUT->SetAlpha(hierarchyOpacity * displayNode->GetSliceIntersectionOpacity());
-        mapper->SetLookupTable(dNodeLUT);
+        if (dNodeLUT)
+          {
+          dNodeLUT->SetAlpha(hierarchyOpacity* displayNode->GetSliceIntersectionOpacity());
+          mapper->SetLookupTable(dNodeLUT);
```
lassoan, line 611: Shouldn't we still call mapper->SetLookupTable if dNodeLUT is nullptr?

Sunderlandkyl: Yeah, you're right. Changed!

PR #5108, opened 2020-08-17
BUG: Fix crash when loading closed surface segmentations

Sunderlandkyl:
When loading a segmentation from closed surface, the vtkMRMLSegmentationStorageNode will cause a crash if one of the blocks in the seg.vtm is unable to be loaded. Fixed by checking that the block could be read before trying to access it, and logging an error message if it could not.

I haven't figured out what causes a segmentation to be written missing some blocks, but this fixes the error when it does occur.

### DIFF
--- a/Libs/MRML/Core/vtkMRMLSegmentationStorageNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLSegmentationStorageNode.cxx
@@ -923,6 +923,11 @@ int vtkMRMLSegmentationStorageNode::ReadPolyDataRepresentation(vtkMRMLSegmentati
     {
     // Get poly data representation
     vtkPolyData* currentPolyData = vtkPolyData::SafeDownCast(multiBlockDataset->GetBlock(blockIndex));
+    if (!currentPolyData)
+      {
+      vtkErrorMacro("ReadPolyDataRepresentation: Could not read block " << blockIndex);
+      continue;
+      }
 
     // Set master representation if it has not been set yet
     // (there is no global place to store it, but every segment field data contains a copy of it)


Libs/MRML/Core/vtkMRMLSegmentationStorageNode.cxx
```diff
@@ -923,6 +923,11 @@ int vtkMRMLSegmentationStorageNode::ReadPolyDataRepresentation(vtkMRMLSegmentati
     {
     // Get poly data representation
     vtkPolyData* currentPolyData = vtkPolyData::SafeDownCast(multiBlockDataset->GetBlock(blockIndex));
+    if (!currentPolyData)
+      {
+      vtkErrorMacro("ReadPolyDataRepresentation: Could not read block " << blockIndex);
```
lassoan, line 928: If we can access the segment id or name then it could be nice to include that in the log message.

Libs/MRML/Core/vtkMRMLSegmentationStorageNode.cxx
```diff
@@ -923,6 +923,11 @@ int vtkMRMLSegmentationStorageNode::ReadPolyDataRepresentation(vtkMRMLSegmentati
...
+      continue;
```
lassoan, line 929: Can it cause any issue elsewhere that the polydata is not set? We could set an empty polydata object in the representation instead.

If we create a new empty segment, is it initialized with an empty polydata or no representation is added at all?

Libs/MRML/Core/vtkMRMLSegmentationStorageNode.cxx
Sunderlandkyl: Unfortunately, we don't have the name or id. They are normally read from the FieldData of the polydata.

Libs/MRML/Core/vtkMRMLSegmentationStorageNode.cxx
Sunderlandkyl: In this case we would avoid creating the segment entirely (since we don't know the name or id), so there wouldn't be any issues if the polydata wasn't set.

When we create a new segment, an empty polydata is set as the representation.

Sunderlandkyl:
I found out how to reproduce the missing block:
- Load MRHead
- Create a segmentation
- Add 3 segments
- Draw something on the 1st and 3rd segments
- Set master representation to Closed surface
- Save segmentation
- Load segmentation

Libs/MRML/Core/vtkMRMLSegmentationStorageNode.cxx
lassoan: OK, then this is really the best we can do at this point.

lassoan:
We need to figure out something with vtkXMLMultiBlockDataWriter ignoring empty polydata. Maybe we could save a polydata with a single vertex when the actual polydata is empty.

PR #5109, opened 2020-08-18
BUG: Fix layout errors in qSlicerSaveDataDialog

### LINKED ISSUE #2443: add a run-time warning when trying to save a transformed .vtk file
_This issue was created automatically from an original [Mantis Issue](https://mantisarchive.slicer.org/view.php?id=2443). Further discussion may take place here._

lassoan:
Fix the following problems in qSlicerSaveDataDialog:
- when a long directory path was used then the dialog did not fit the screen on some systems and "Save" button was not visible (fixed by using ctkPathLineEdit instead of ctkDirectoryButton)
- fixed lots of widgets appearing and disappearing before save dialog is shown (recent regression introduced by status icon display)
- remove separate "Notes" column for saving status (added icon to filename column instead, where it is much more visible, take less space, and simpler to implement)
- fix deselection of successfully saved items

image

Also add warning if saving in Analyze format or rotated volume in VTK format - fixes #2443


pieper:
Didn't look in detail, but sounds like good fixes 👍

PR #5112, opened 2020-08-19
STYLE: Remove deprecated member from crosshair node

cpinter:
The navigation feature has been removed but this member was kept for API compatibility. Now that we are approaching 5.0, it can be removed to simplify the codebase.

See some more details here: https://discourse.slicer.org/t/crosshair-with-mouse-drag-drop-navigation/13045


cpinter:
OK, done.

PR #5127, opened 2020-08-22
BUG: Update EditBox.py to use setExclusive method

carjohnson:
Use method to enable the group exclusion checking property.
Direct assignment of this property results in the message:
      AttributeError: 'exclusive' does not exist on QActionGroup and creating new attributes on C++ objects is not allowed
and the EditBox will not open.

### DIFF
--- a/Modules/Scripted/EditorLib/EditBox.py
+++ b/Modules/Scripted/EditorLib/EditBox.py
@@ -34,7 +34,7 @@ class EditBox(VTKObservationMixin):
     self.effectCursors = {}
     self.effectActionGroup = qt.QActionGroup(parent)
     self.effectActionGroup.connect('triggered(QAction*)', self._onEffectActionTriggered)
-    self.effectActionGroup.exclusive = True
+    self.effectActionGroup.setExclusive(True)
     self.currentEffect = None
     self.undoRedo = UndoRedo()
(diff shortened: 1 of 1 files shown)


pieper:
Yes, I like the commit message @jamesobutler suggests.

@carjohnson, perhaps the easiest is to make a new PR with the same fix and the new commit message, plus add the url for this PR for reference.  If you aren't comfortable with that one of us can just add your fix.  Thanks again for the contribution.

lassoan:
@carjohnson You can "amend" the last commit and "force-push" - if you use TortoiseGit as git client on Windows, then you can do the followings:
- Select Commit in the right-click menu
- Select "Amend last commit" 
- Edit the commit message
- Click Commit
- Select Push in the right-click menu
- Check "Override known changes" option
- Click Push

carjohnson:
I've created a new pull request with your suggestions - reference: https://github.com/Slicer/Slicer/pull/5129

I am using the legacy version of the segment editor because I only needed a few simple editing tools for the application I am building. I found it easiest to override the classes in the python scripts in EditorLib. I will revisit the Segment Editor module to see how I can incorporate it into my project.

jamesobutler:
Superseded by #5129.

PR #5128, opened 2020-08-23
ENH: Add option to render markups on top of everything else

Sunderlandkyl:
By enabling vtkMRMLMarkupsDisplayNode::AlwaysOnTop, the markup widget will be rendered on top of all other objects in the scene.
This option can be toggled on and off from the markups display node ui.

### DIFF
--- a/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsDisplayNode.cxx
+++ b/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsDisplayNode.cxx
@@ -102,6 +102,9 @@ vtkMRMLMarkupsDisplayNode::vtkMRMLMarkupsDisplayNode()
   this->LineColorFadingSaturation = 1.;
   this->LineColorFadingHueOffset = 0.;
 
+  this->OccludedVisibility = false;
+  this->OccludedOpacity = 0.3;
+
   this->HandlesInteractive = false;
 
   // Line color node
@@ -147,6 +150,8 @@ void vtkMRMLMarkupsDisplayNode::WriteXML(ostream& of, int nIndent)
   vtkMRMLWriteXMLBooleanMacro(outlineVisibility, OutlineVisibility);
   vtkMRMLWriteXMLFloatMacro(fillOpacity, FillOpacity);
   vtkMRMLWriteXMLFloatMacro(outlineOpacity, OutlineOpacity);
+  vtkMRMLWriteXMLBooleanMacro(occludedVisibility, OccludedVisibility);
+  vtkMRMLWriteXMLFloatMacro(occludedOpacity, OccludedOpacity);
   vtkMRMLWriteXMLEndMacro();
 }
 
@@ -183,6 +188,8 @@ void vtkMRMLMarkupsDisplayNode::ReadXMLAttributes(const char** atts)
   vtkMRMLReadXMLBooleanMacro(outlineVisibility, OutlineVisibility);
   vtkMRMLReadXMLFloatMacro(fillOpacity, FillOpacity);
   vtkMRMLReadXMLFloatMacro(outlineOpacity, OutlineOpacity);
+  vtkMRMLReadXMLBooleanMacro(occludedVisibility, OccludedVisibility);
+  vtkMRMLReadXMLFloatMacro(occludedOpacity, OccludedOpacity);
   vtkMRMLReadXMLEndMacro();
 
   // Fix up legacy markups fiducial nodes
@@ -247,6 +254,8 @@ void vtkMRMLMarkupsDisplayNode::CopyContent(vtkMRMLNode* anode, bool deepCopy/*=
   vtkMRMLCopyBooleanMacro(OutlineVisibility);
   vtkMRMLCopyFloatMacro(FillOpacity);
   vtkMRMLCopyFloatMacro(OutlineOpacity);
+  vtkMRMLCopyBooleanMacro(OccludedVisibility);
+  vtkMRMLCopyFloatMacro(OccludedOpacity);
   vtkMRMLCopyEndMacro();
 }
 
@@ -425,6 +434,8 @@ void vtkMRMLMarkupsDisplayNode::PrintSelf(ostream& os, vtkIndent indent)
   vtkMRMLPrintBooleanMacro(OutlineVisibility);
   vtkMRMLPrintFloatMacro(FillOpacity);
   vtkMRMLPrintFloatMacro(OutlineOpacity);
+  vtkMRMLPrintBooleanMacro(OccludedVisibility);
+  vtkMRMLPrintFloatMacro(OccludedOpacity);
   vtkMRMLPrintEndMacro();
 }
 
--- a/Modules/Loadable/Markups/VTKWidgets/vtkSlicerAngleRepresentation3D.cxx
+++ b/Modules/Loadable/Markups/VTKWidgets/vtkSlicerAngleRepresentation3D.cxx
@@ -69,6 +69,20 @@ vtkSlicerAngleRepresentation3D::vtkSlicerAngleRepresentation3D()
   this->ArcActor->SetMapper(this->ArcMapper);
   this->ArcActor->SetProperty(this->GetControlPointsPipeline(Unselected)->Property);
 
+  this->LineOccludedMapper = vtkSmartPointer<vtkPolyDataMapper>::New();
+  this->LineOccludedMapper->SetInputConnection(this->TubeFilter->GetOutputPort());
+
+  this->ArcOccludedMapper = vtkSmartPointer<vtkPolyDataMapper>::New();
+  this->ArcOccludedMapper->SetInputConnection(this->ArcTubeFilter->GetOutputPort());
+
+  this->LineOccludedActor = vtkSmartPointer<vtkActor>::New();
+  this->LineOccludedActor->SetMapper(this->LineOccludedMapper);
+  this->LineOccludedActor->SetProperty(this->GetControlPointsPipeline(Unselected)->OccludedProperty);
+
+  this->ArcOccludedActor = vtkSmartPointer<vtkActor>::New();
+  this->ArcOccludedActor->SetMapper(this->ArcOccludedMapper);
+  this->ArcOccludedActor->SetProperty(this->GetControlPointsPipeline(Unselected)->OccludedProperty);
+
   this->HideTextActorIfAllPointsOccluded = true;
 }
 
@@ -188,6 +202,8 @@ void vtkSlicerAngleRepresentation3D::UpdateFromMRML(vtkMRMLNode* caller, unsigne
 
   this->UpdateRelativeCoincidentTopologyOffsets(this->LineMapper);
   this->UpdateRelativeCoincidentTopologyOffsets(this->ArcMapper);
+  this->UpdateOccludedRelativeCoincidentTopologyOffsets(this->LineOccludedMapper);
+  this->UpdateOccludedRelativeCoincidentTopologyOffsets(this->ArcOccludedMapper);
 
   double diameter = ( this->MarkupsDisplayNode->GetCurveLineSizeMode() == vtkMRMLMarkupsDisplayNode::UseLineDiameter ?
     this->MarkupsDisplayNode->GetLineDiameter() : this->ControlPointSize * this->MarkupsDisplayNode->GetLineThickness() );
@@ -198,6 +214,8 @@ void vtkSlicerAngleRepresentation3D::UpdateFromMRML(vtkMRMLNode* caller, unsigne
 
   this->LineActor->SetVisibility(lineVisibility);
   this->ArcActor->SetVisibility(lineVisibility && markupsNode->GetNumberOfDefinedControlPoints(true) == 3);
+  this->LineOccludedActor->SetVisibility(this->MarkupsDisplayNode && this->LineActor->GetVisibility() && this->MarkupsDisplayNode->GetOccludedVisibility());
+  this->ArcOccludedActor->SetVisibility(this->MarkupsDisplayNode && this->ArcActor->GetVisibility() && this->MarkupsDisplayNode->GetOccludedVisibility());
 
   int controlPointType = Active;
   if (this->MarkupsDisplayNode->GetActiveComponentType() != vtkMRMLMarkupsDisplayNode::ComponentLine)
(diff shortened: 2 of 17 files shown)


lassoan:
Thank you, I like that this is a very simple solution. However, it breaks consistency of the 3D display: there is no way to distinguish between occluded and actually visible control points. What worked extremely well in another project in the past was to render "always-on-top" actor with reduced opacity in addition to the regular actor (with full opacity).

See for example what you can achieve by rendering the same curve using a regular and an always-on-top node:

image

Sunderlandkyl:
Okay, I've added another actor with controllable opacity for each markup type:

image

lassoan:
I've tested this and found some issues:
- If "Occluded" is checked then the curve appears on the top, regardless of Opacity settings (with both depth peeling enabled/disabled).
- "Occluded" seems to work backwards (curve not occluded if "Occluded" is enabled): the name is a bit unclear and unusual, too. "3D Projection" would be a simple name and the intent would be similar to "2D Projection" (show the markup in the view "no matter what"), and a similar "3D Projection Opacity" option could control its opacity. Unfortunately, it would not be a very accurate wording in term of engineering. "Occluded parts visibility/opacity in 3D" would be accurate but too long. @ungi what do you think? 
- Transform interaction is always visible (regardless of checkbox state)
- Markups/Display/Advanced section should be non-collapsed by default

image

I've rebased my branch on latest master, cherry-picked your commit, and built - it may be possible that there is something wrong with my computer/build, but I did not have any merge conflicts or any related changes in my branch, so it is not very likely. But if this works well on your computer then I can try to build again.

ungi:
I agree "Occluded" for me too means that something is not visible. If the markup becomes visible by checking that box, I would call the checkbox "Show in front" or "Always visible". The Slider also needs a more descriptive name. E.g. if you call it "Occlusion", then maximum slider value could make the occluded parts invisible. If you call it "See through" then maximum slider value could make the occluded parts completely visible.

Sunderlandkyl:
>Transform interaction is always visible (regardless of checkbox state)

Does this mean that the interaction handles are visible regardless of the "Occluded" check box, or the Interaction "Visible" check box?

> I agree "Occluded" for me too means that something is not visible. If the markup becomes visible by checking that box, I would call the checkbox "Show in front" or "Always visible".

Ok, I will change the name to "Always visible".

> The Slider also needs a more descriptive name. E.g. if you call it "Occlusion", then maximum slider value could make the occluded parts invisible. If you call it "See through" then maximum slider value could make the occluded parts completely visible.

When you say "occluded parts", do you mean the parts of the markup that is occluded, or the rest of the scene covered up by the markup? "See through" to me would suggest that at maximum value the occluded parts of the markup should be fully transparent.

ungi:
You are right, the labels should talk about the markup, not other objects. So "Occlusion" and "See through" are not good words. How about "Fade"? When Fade is 0%, I would expect the markup to be completely visible everywhere. When Fade is 100%, then I would expect parts behind other things to be invisible. (Fade 100% would be the same as "Always visible"=OFF.)

Sunderlandkyl:
Labels changed to "Always visible" and "Fade".

image

lassoan:
"Always visible" conflicts with "Visibility". Somehow we should indicate that it refers to the occluded parts of the markup. "Occluded parts visible" or something similar would not be too bad.

> Does this mean that the interaction handles are visible regardless of the "Occluded" check box, or the Interaction "Visible" check box?

Yes. Also, opacity of the top rendering layer is always 100%. Maybe you haven't pushed the latest version of your code on the branch? I can check a bit later if I got the branch right and rebuilt everything.

ungi:
How about "Occlusions visible"? I'm happy with "Occluded parts visible" too, just trying to come up with something shorter, so we save space for widgets on the right side.

lassoan:
"Occlusions visible" sounds OK to me. If native speakers don't cringe then it should work. @Sunderlandkyl @pieper what do you think?

Sunderlandkyl:
That sounds fine to me. Are we sticking with fade for the opacity slider?

sjh26:
"Show Occluded" ?

lassoan:
"Show occluded" sounds good, but "show" is an action and we try to use state (visible/visibility)  instead. "Occluded visible" or "Occluded visibility" would be more consistent (with an "Opacity" slider - the same way as everywhere else).

There was a build error in the version that I tried but the latest version works very well! I've found only one minor issue: line seems to be in front of the glyphs (which is especially visible when when fade=0.0):

image

If you are already modifying the display properties, we could add a "3D Display" and "Slice Display" subsections (similarly to Models module), either in Display/Advanced (non-collapsed by default) or directly in Display (collapsed by default). You could put "Occluded visible" + "Opacity" slider in 3D section, and 2D projection settings in slice section. It would be nice if you could make the projection settings a bit more consistent, too (labels aligned to the left, visibility checkbox instead of eye, opacity in the same line).

Sunderlandkyl:
- Options are now "Occluded Visibility/Opacity".
- Added the 2D Display/3D Display sections.
- Control points are now rendered on top of the representation.

image

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsDisplayNode.cxx
```diff
@@ -102,6 +102,9 @@ vtkMRMLMarkupsDisplayNode::vtkMRMLMarkupsDisplayNode()
   this->LineColorFadingSaturation = 1.;
   this->LineColorFadingHueOffset = 0.;
 
+  this->OccludedVisibility = false;
+  this->OccludedOpacity = 0.75;
```
lassoan, line 106: I find this default a bit too high - it is barely distinguishable from full-opacity. What do you think about setting in to 0.3 or 0.5?

0.75:

image

0.5:

image

0.3:

image

Modules/Loadable/Markups/VTKWidgets/vtkSlicerAngleRepresentation3D.cxx
```diff
@@ -69,6 +69,27 @@ vtkSlicerAngleRepresentation3D::vtkSlicerAngleRepresentation3D()
   this->ArcActor->SetMapper(this->ArcMapper);
   this->ArcActor->SetProperty(this->GetControlPointsPipeline(Unselected)->Property);
 
+  this->LineOccludedMapper = vtkSmartPointer<vtkPolyDataMapper>::New();
+  this->LineOccludedMapper->SetInputConnection(this->TubeFilter->GetOutputPort());
+  this->LineOccludedMapper->SetRelativeCoincidentTopologyLineOffsetParameters(0, -25000);
```
lassoan, line 74: Add explanation to magic number -25000. Maybe even better: define a helper function in a base class to set these offset parameters.

Modules/Loadable/Markups/VTKWidgets/vtkSlicerMarkupsWidgetRepresentation3D.cxx
```diff
@@ -633,6 +651,23 @@ void vtkSlicerMarkupsWidgetRepresentation3D::UpdateFromMRML(vtkMRMLNode* caller,
     controlPoints->TextProperty->SetOpacity(this->MarkupsDisplayNode->GetOpacity() * hierarchyOpacity);
     controlPoints->TextProperty->SetFontSize(static_cast<int>(5. * this->MarkupsDisplayNode->GetTextScale()));
 
+    controlPoints->OccludedProperty->DeepCopy(controlPoints->Property);
+    if (this->MarkupsDisplayNode->GetOccludedVisibility() && this->MarkupsDisplayNode->GetOccludedOpacity() > 0.0)
+      {
+      // If OccludedVisibility is enabled, then the depth of the widget will be 0.0.
+      // The point should never be occluded, so set tolerance to 1.0.
+      controlPoints->SelectVisiblePoints->SetTolerance(1.0);
```
lassoan, line 659: It would be nice if we would still use point visibility and display occluded labels with OccludedOpacity. If this is complicated then just write there in a comment that it could be done and what are the complications.

Sunderlandkyl: It's possible if we change when the vtkSelectVisiblePoints filter is updated, as it would need to be updated before the points are rendered.
I don't think it should be too complicated.

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsDisplayNode.cxx
Sunderlandkyl: Right. I actually meant to set it to 0,25. It was 0.75 from when I had changed the variable to be "Fade".
I think 0.3 makes sense.

Modules/Loadable/Markups/VTKWidgets/vtkSlicerMarkupsWidgetRepresentation3D.cxx
Sunderlandkyl: Should occluded control points be interactive?

ungi: I cannot think of an application that would require the user to move always-on-top markups. Because it would be hard to intuitively feel which way they go in 3D. But I can imagine that people would like to click on them to center the 2D views on a particular fiducial, or show other annotations. E.g. imagine an interactive anatomical atlas. People could click on these points inside the 3D anatomy to e.g. display their names, or some measurements related to them.

lassoan: I agree, it would be very useful to be able to interact with all visible landmarks, even if occluded. Maybe in some cases it could be useful to lock them, but I don't have anything specific application in mind, so keeping them interactive for now should be good.

Sunderlandkyl:
OK, the labels now have an occluded actor, and I've added Set/Get for the OccludedRelativeOffset.

Modules/Loadable/Markups/VTKWidgets/vtkSlicerAngleRepresentation3D.cxx
```diff
@@ -69,6 +69,20 @@ vtkSlicerAngleRepresentation3D::vtkSlicerAngleRepresentation3D()
   this->ArcActor->SetMapper(this->ArcMapper);
   this->ArcActor->SetProperty(this->GetControlPointsPipeline(Unselected)->Property);
 
+  this->LineOccludedMapper = vtkSmartPointer<vtkPolyDataMapper>::New();
```
lassoan, line 72: It would be more consistent to keep the same order (all actors then all mappers), and it would be better to keep ...Mapper and ...OccludedMapper closer to each other everywhere, because if we want to add a mapper then we most likely need to add an OccludedMapper as well.

Modules/Loadable/Markups/VTKWidgets/vtkSlicerAngleRepresentation3D.cxx
```diff
@@ -188,6 +202,8 @@ void vtkSlicerAngleRepresentation3D::UpdateFromMRML(vtkMRMLNode* caller, unsigne
 
   this->UpdateRelativeCoincidentTopologyOffsets(this->LineMapper);
```
lassoan, line 203: It would make things safer (harder to forget about occluded mapper) and a bit more compact if `UpdateRelativeCoincidentTopologyOffsets` would take two arguments, the non-occluded and the occluded mapper:

    this->UpdateRelativeCoincidentTopologyOffsets(this->LineMapper, this->LineOccludedMapper);

Modules/Loadable/Markups/VTKWidgets/vtkSlicerMarkupsWidgetRepresentation3D.cxx
```diff
@@ -149,6 +181,8 @@ vtkSlicerMarkupsWidgetRepresentation3D::vtkSlicerMarkupsWidgetRepresentation3D()
 
   this->AccuratePicker = vtkSmartPointer<vtkCellPicker>::New();
   this->AccuratePicker->SetTolerance(.005);
+
+  this->OccludedRelativeOffset = -25000; /// Default occluded relative offset
```
lassoan, line 185: Can you describe in the commend why the value is -25000 (why not less, why not more)?

Modules/Loadable/Markups/VTKWidgets/vtkSlicerMarkupsWidgetRepresentation3D.cxx
```diff
@@ -627,15 +677,34 @@ void vtkSlicerMarkupsWidgetRepresentation3D::UpdateFromMRML(vtkMRMLNode* caller,
...
 
+    controlPoints->OccludedProperty->SetColor(color);
+    controlPoints->OccludedTextProperty->SetColor(color);
+    controlPoints->OccludedTextProperty->SetFontSize(controlPoints->TextProperty->GetFontSize());
+    if (this->MarkupsDisplayNode->GetOccludedVisibility() && this->MarkupsDisplayNode->GetOccludedOpacity() > 0.0)
+      {
+      // To prevent some rendering artifacts, and to ensure that the occluded actor does not block point visibility,
+      // the maximum opacity of the occluded actor is required to be almost, but not fully opaque.
+      double occludedOpacity =
+        std::min(0.99999999, this->MarkupsDisplayNode->GetOccludedOpacity() * opacity);
```
lassoan, line 698: This value is a bit risky, as 0.99999999 can only be stored in double, and not in float (in float it would be stored as 1.0, see https://www.h-schmidt.net/FloatConverter/IEEE754.html). Using 0.9999999 can be stored in single-precision float, but to avoid any numerical issues, we could use a very safe value such as 0.99 (that would be even non-opaque if stored in an unsigned char alpha channel).

Modules/Loadable/Markups/VTKWidgets/vtkSlicerMarkupsWidgetRepresentation3D.cxx
```diff
@@ -809,7 +890,7 @@ int vtkSlicerMarkupsWidgetRepresentation3D::RenderOpaqueGeometry(
     {
     ControlPointsPipeline3D* controlPoints = reinterpret_cast<ControlPointsPipeline3D*>(this->ControlPoints[i]);
     if (controlPoints->Actor->GetVisibility())
-      {
+    {
```
lassoan, line 893: indentation

Modules/Loadable/Markups/Widgets/qMRMLMarkupsDisplayNodeWidget.cxx
```diff
@@ -596,4 +608,26 @@ void qMRMLMarkupsDisplayNodeWidget::onOutlineOpacitySliderWidgetChanged(double o
...
+    }
+  d->MarkupsDisplayNode->SetOccludedVisibility(visibility);
+}
+
+//-----------------------------------------------------------------------------
+void qMRMLMarkupsDisplayNodeWidget::setOccludedOpacity(double OccludedOpacity)
+{
+  Q_D(qMRMLMarkupsDisplayNodeWidget);
+  if (!d->MarkupsDisplayNode)
+  {
```
lassoan, line 629: indentation

Modules/Loadable/Markups/VTKWidgets/vtkSlicerMarkupsWidgetRepresentation3D.cxx
jcfr: > https://www.h-schmidt.net/FloatConverter/IEEE754.html

This website is neat. Thanks for sharing

For future reference:

* using `0.99999999`
image

* using `0.99`
image

PR #5129, opened 2020-08-24
BUG: Fix invalid use of QActionGroup exclusive attribute

carjohnson:
The exclusive attribute for QActionGroup was replaced starting in Qt 5.14 with the exclusionPolicy attribute. Using the setExclusive method will set the exclusive or exclusionPolicy attribute depending on the Qt 5 version being used.

Reference: https://github.com/Slicer/Slicer/pull/5127


jcfr:
Thanks @carjohnson and @jamesobutler :pray:

PR #5130, opened 2020-08-24
BUG: Fix crash when filtering by name in subject hierarchy

cpinter:
If filtering by name affected items that were highlighted as reference of the selected item, then a crash may have occurred. It is fixed by checking if the item to be highlighted is actually allowed by the filter (and a valid proxy item exists).

### DIFF
--- a/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyTreeView.cxx
+++ b/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyTreeView.cxx
@@ -1817,7 +1817,7 @@ void qMRMLSubjectHierarchyTreeView::applyReferenceHighlightForItems(QList<vtkIdT
   foreach (vtkIdType highlightedItemID, d->HighlightedItems)
     {
     QStandardItem* item = sceneModel->itemFromSubjectHierarchyItem(highlightedItemID, nameColumn);
-    if (item)
+    if (item && this->sortFilterProxyModel()->filterAcceptsItem(highlightedItemID))
       {
       item->setBackground(Qt::transparent);
       }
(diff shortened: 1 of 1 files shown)


Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyTreeView.cxx
```diff
@@ -1882,9 +1883,9 @@ void qMRMLSubjectHierarchyTreeView::applyReferenceHighlightForItems(QList<vtkIdT
       if (referencedItem && referencedItem != itemID)
         {
         QStandardItem* item = sceneModel->itemFromSubjectHierarchyItem(referencedItem, nameColumn);
-        if (item && !d->HighlightedItems.contains(referencedItem))
+        if (item && !d->HighlightedItems.contains(referencedItem) && this->sortFilterProxyModel()->filterAcceptsItem(referencedItem))
           {
-          item->setBackground(d->IndirectReferenceColor);
+          item->setBackground(QColor::fromRgb(255, 255, 170));
```
jcfr, line 1888: Should these attribute be removed as well:

https://github.com/Slicer/Slicer/blob/231fc4f35f63606fa81ffe53b330efdc2477159b/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyTreeView.cxx#L145-L148

cpinter: No, sorry, this is a merge mistake due to a very recent change in the same file and the same function. I actually need to add those instead of the colors. I'll fix this asap.

cpinter:
I pushed the fix. Thanks @jcfr for catching this!

jcfr:
Thanks @cpinter

PR #5105, opened 2020-08-15
Update to tagged ITK 5.1.1 version

jamesobutler:
This updates the ITK version that Slicer uses to an official ITK tagged version (5.1.1).  I first ran into a build error related to TBB that was seemingly the result of the change in https://github.com/InsightSoftwareConsortium/ITK/commit/51792db03b24d1aa7e80524998064c1745fe4d99. The version of TBB currently specified in the BRAINSTools (the Slicer version) superbuild is actually a version of v2020.2 as seen [here](https://github.com/Slicer/BRAINSTools/blob/slicer-2020-04-14-v5.2.0/SuperBuild/External_TBB.cmake#L27) so I expect we can update the version to at least that high in the future without issues with BRAINSTools. I also changed to download the TBB binaries from the [official TBB github repo](https://github.com/oneapi-src/oneTBB) instead of the [Slicer/SlicerBinaryDependencies](https://github.com/Slicer/SlicerBinaryDependencies) repo. I'm guessing the TBB binaries were not previously offered in a easy public way on github before or that the binaries were not built with a compatible compiler for the Slicer factory build machine?

I updated the TBB version to 2019_U9 because the minimum mac OS X version for this update is 10.11 ([source](https://github.com/oneapi-src/oneTBB/blob/2019_U9/doc/Release_Notes.txt)). Starting with TBB v2020.0 the minimum macOS version is 10.14 ([source](https://github.com/oneapi-src/oneTBB/blob/v2020.0/doc/Release_Notes.txt)) which is greater than the current Slicer minimum macOS version of 10.13. Also starting with TBB version v2020.0 it drops Windows 8.1 and Windows 7 support and requires the Visual Studio 2017 compiler or later.

I've updated SimpleITK to v2.0rc3 which is using a more recent ITK 5.1.x version than we were previously using.  I've specified the SimpleITK upstream repo because there are currently no Slicer specific patches needed.  If using the [Slicer/SimpleITK](https://github.com/Slicer/SimpleITK) fork is still desired, I can change it back and push a new branch to the fork based on the new git commit hash. I'm expecting an update to using a SimpleITK official 2.0.0 tag in the near future.

Currently building in release mode on Windows 10 with standard configuration except for enabling parallel build. Will report results here once finished.

PR #5132, opened 2020-08-25
BUG: remove unavailable extension suggestion

pieper:
Since the Longitudinal PET/CT extension
is not currently available, remove
it from the suggestion list for now.

Longer term: once the new extension manager
is in place, consider auto-generating
suggestions based on a query of available
extension.  We should also explore adding
tags to s4ext files that could explicitly
declare the data types they are appropriate
for.

Co-authored-by: Sonia Pujol <spujol@bwh.harvard.edu>

### DIFF
--- a/Modules/Scripted/DICOM/DICOMExtensions.json
+++ b/Modules/Scripted/DICOM/DICOMExtensions.json
@@ -1,16 +1,6 @@
 {
   "extensions": [
     {
-      "name": "LongitudinalPETCT",
-      "repository": "https://github.com/QIICR/LongitudinalPETCT",
-      "typeDescription": "PET Images",
-      "tagValues": {
-        "Modality": [
-          "PT"
-        ]
-      }
-    },
-    {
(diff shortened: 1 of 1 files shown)


jcfr:
Should an issue be created to track the fact the suggestion should be re-added ? 

Also, I am wondering if  the associated build issues (see link below) will be fixed by the ITK update (https://github.com/Slicer/Slicer/pull/5105) or if it is a matter of using an updated API ... 

http://slicer.cdash.org/index.php?project=SlicerPreview&filtercount=1&showfilters=1&field1=buildname&compare1=63&value1=Longitudinal

pieper:
Hi @jcfr - we discussed this a bit after you left the call - basically the extension isn't currently being maintained so we wanted to remove it from the popup to avoid confusion.  Not known if it's just the build error or other issues need to be maintained.

I added an issue linking to this PR for future reference.

lassoan (approved):
Looks good to me.

PR #5136, opened 2020-08-25
STYLE: make python errors readable in dark style

pieper:
Using Qt::red was not bright enough for some users, so
increase the luminance a bit.

Not that to experiment with different color options one
can use commands like this:

slicer.app.pythonConsole().errorTextColor = qt.QColor("#ff4444")


pieper:
Here's the proposed change (top) vs previous default.
image

pieper:
Thanks for the reviews guys 👍 

I grepped around for it and it was only in the AnnotationROI gui and I think the other color will work fine there too.  I also think this brigher color is more consistent with the Qt definition of [BrightText](https://doc.qt.io/qt-5/qpalette.html#ColorRole-enum), that it should contrast with Dark so we should be good.

lassoan:
This is a standard palette entry, so all the standard Qt widgets may use it, too, but making it brighter should not be an issue there either.

pieper:
Yes, because it is a standard role we should make it match the definition of the role and this change makes it fit more closely.

PR #5138, opened 2020-08-26
ENH: Add option to toggle the markup measurements included in the des…

Sunderlandkyl:
…cription

Adds the MeasurementsInDescription attribute to vtkMRMLMarkupsNode, that controls the markups added to the description by WriteMeasurementsToDescription.
If MeasurementsInDescription is empty, then the description will be empty, and the measurement label in the 3D/2D views will not be displayed.

MeasurementsInDescription can be controlled from the markups subject hierarchy plugin. Each measurement adds a checkbox to the context menu to enable/disable display.

image

### DIFF
--- a/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsNode.cxx
+++ b/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsNode.cxx
@@ -46,6 +46,7 @@
 #include <vtkPointData.h>
 #include <vtkPolyData.h>
 #include <vtkStringArray.h>
+#include <vtksys/SystemTools.hxx>
 #include <vtkTransform.h>
 #include <vtkTransformPolyDataFilter.h>
 #include <vtkTrivialProducer.h>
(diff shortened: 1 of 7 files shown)


Modules/Loadable/Markups/MRML/vtkMRMLMarkupsNode.cxx
```diff
@@ -2084,9 +2095,26 @@ void vtkMRMLMarkupsNode::WriteMeasurementsToDescription()
       this->PropertiesLabelText += "\n" + measurementText;
       }
     }
+
+  if (description.empty())
+    {
+    this->PropertiesLabelText.clear();
```
lassoan, line 2101: We may need to use the description for other things, so if no measurement is selected to be shown in description then just clear it once and then don't clear it every time (it would fix a problem in Path Explorer module in SlicerIGT).

lassoan:
Discussed with @Sunderlandkyl and he achieved what he needed in a different way (just by hiding the entire label), so closing this pull request for now.

Issue #4383, opened 2020-03-13
Display slice intersection of parallel slices as rectangle


slicerbot:
_This issue was created automatically from an original [Mantis Issue](https://mantisarchive.slicer.org/view.php?id=4383). Further discussion may take place here._

PR #5142, opened 2020-08-27
ENH: Enable a copy of the Data Probe image to follow the cursor

adamaji:
This commit adds an option in the Data Probe to draw a copy of the magnified region that follows the cursor. Having the region closer to where the cursor is pointing makes it easier to compare the zoomed out view with the zoomed in view, and potentially allows for more space in the side bar if you wanted to minimize the Data Probe module.

Some issues with the current implementation are 
- the floating region doesn't follow on the shift+mousemove shortcut
- the floating region (and also the regular Data Probe imageLabel) doesn't show while editing segmentations, where it would likely be the most helpful

image

cursor_slicer

### DIFF
--- a/Modules/Scripted/DataProbe/DataProbe.py
+++ b/Modules/Scripted/DataProbe/DataProbe.py
@@ -74,6 +74,7 @@ class DataProbeInfoWidget(object):
     self.iconsDIR = modulePath + '/Resources/Icons'
 
     self.showImage = False
+    self.undocked = False
 
     # Used in _createMagnifiedPixmap()
     self.imageCrop = vtk.vtkExtractVOI()
@@ -204,6 +205,7 @@ class DataProbeInfoWidget(object):
         self.layerIJKs[layer].setText( "" )
         self.layerValues[layer].setText( "" )
       self.imageLabel.hide()
+      self.floatingImageLabel.hide()
       self.viewerColor.hide()
       self.viewInfo.hide()
       self.viewerFrame.hide()
@@ -275,12 +277,37 @@ class DataProbeInfoWidget(object):
         self.imageLabel.setPixmap(pixmap)
         self.onShowImage(self.showImage)
 
+        # show the floating image region
+        if self.undocked:
+          if not self.floatingImageLabel.isVisible():
+            self.floatingImageLabel.show()
+
+          point = qt.QCursor().pos()
+          offset = qt.QPoint(10, 10)  # offset the region to keep away from cursor
+          regionPos = point + offset
+          self.floatingImageLabel.move(regionPos)
+
+          # need to create a new pixmap as the floating region may not be the same size as the docked widget
+          pixmap = self._createMagnifiedPixmap(
+            xyz, sliceLogic.GetBlend().GetOutputPort(), self.floatingImageLabel.size, color)
+          self.floatingImageLabel.setPixmap(pixmap)
+
+          # make the undocked region circular
+          width = self.floatingImageLabel.width
+          height = self.floatingImageLabel.height
+          side = min(width, height)
+          region = qt.QRegion(0, 0, side, side, qt.QRegion.Ellipse)
+          self.floatingImageLabel.setMask(region)
+
     if hasattr(self.frame.parent(), 'text'):
       sceneName = slicer.mrmlScene.GetURL()
       if sceneName != "":
         self.frame.parent().text = "Data Probe: %s" % self.fitName(sceneName,nameSize=2*self.nameSize)
       else:
         self.frame.parent().text = "Data Probe"
+
+    if not (self.undocked and self.showImage):
+      self.floatingImageLabel.hide()
 
   def generateViewDescription(self, xyz, ras, sliceNode, sliceLogic):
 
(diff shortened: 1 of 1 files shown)


pieper:
Hi - thanks for the contribution 👍 

It could make sense to unify this implementation with the RevealCursor provided by the CompareVolumes module.

https://github.com/pieper/CompareVolumes/blob/master/CompareVolumes.py#L604-L773

https://www.slicer.org/wiki/Documentation/Nightly/Modules/CompareVolumes

adamaji:
Hi Steve, not sure I understand -- do you mean to also add an option for the CompareVolumes reveal cursor to show a floating view of the fg over the bg for comparison instead of the cutaway view that the module currently uses? Or to show the cutaway in the floating view?

pieper:
> Hi Steve, not sure I understand -- do you mean to also add an option for the CompareVolumes reveal cursor to show a floating view of the fg over the bg for comparison instead of the cutaway view that the module currently uses? Or to show the cutaway in the floating view?

I didn't have a specific plan, but since they provide similar functionality it might make sense to unify them somehow.

> It appears that in https://github.com/pieper/CompareVolumes/blob/a78f04592488e45eb7700679acd418ba0240651a/CompareVolumes.py#L693-L703 it gets the image data to render the various layers. Unfortunately there doesn't appear to be a `slicer.app.layoutManager().sliceWidget("Red").sliceLogic().GetSegmentationLayer()` type method to then call

Good point, I don't think segmentations are exposed that way and maybe would never be given the way displayable managers work.  We would need to grab the rendered image and zoom it up, which would be pixelated.

Probably the cleaner implementation overall would be to make a new instance of a sliceView and synchronize have it float above the regular app window following the mouse like this.  Then use the jump slice with centering option and zoom it in using the slice node to get the magnifier effect without pixelation.

This would allow it to automatically get all the displayable types without needing any special knowledge of the internals.  We could also use the existing infrastructure to decide which content gets displayed and how.

lassoan:
Today's screens are very high resolution. I've played with built-in magnifier tool in Windows and even at 800% zoom level pixelation was not an issue at all (and of course you can apply filtering to remove it).

However, after playing with the Windows built-in magnifier tool, I don't really see a magnifier in Slicer. The built-in magnifier just works so well! There are various display modes (full-screen, docked, floating lens), smoothing, convenient keyboard shortcuts, and lots of other features.

The Windows built-in magnifier also solve the most important limitation of the current Slicer magnifier, difficulty in placing points accurately. If you just zoom in the view but don's slow down the mouse then you can see better where you want to place your points, but you cannot do it, because the mouse moves relatively too quickly.

Reveal cursor still relevant, as the operating system cannot reveal layers in Slicer viewer and mouse speed is not an issue, as the goal is just data exploration and so high-precision mouse movement is not needed.

@adamaji **What is the main motivation for the zoomed-in view?** Is it hard to zoom in? Or you don't know where in the image you are when you zoom in? Maybe we could address these issues specifically - either by improving image navigation and zooming in Slicer, or enabling magnifier tool provided by the operating system's or [third-party software](https://crazytechtricks.com/windows-magnifier-alternatives/#:~:text=%20Top%2010%20Best%20Free%20Windows%20Magnifier%20Alternatives,by%20using%20the%20mouse.%20Moreover%2C%20it...%20More%20).

pieper:
It's true that on the mac I also use the system zoom functions (in the Accessibility section) and I never turn on the pixel replication zoom in the DataProbe.

adamaji:
>Or you don't know where in the image you are when you zoom in?

@lassoan I think this is a big part of it. The UX proposed in this PR feels a little better to me, I think because the context of seeing where the cursor is both at the 100% view and the magnified view at the same time is important. The problem with docking the magnified region (like with the current Data Probe) is having it far away from the cursor, so your eye has to dart back and forth for the context. After trying out the built-in magnification tools on Windows and MacOS I wasn't sure how to offset the floating region to achieve this, but perhaps one of the third party apps that you linked supports behavior like this.

One use case where the context might be helpful would be for segmenting m-mode cardiac ultrasound images, where we want to segment peaks and troughs precisely across a very wide horizontally oriented image. Normally with the Slicer zoom this would require a lot of panning to segment the whole image.

image

lassoan:
>>Or you don't know where in the image you are when you zoom in?
>
> @lassoan I think this is a big part of it. The UX proposed in this PR feels a little better to me, I think because the context of seeing where the cursor is both at the 100% view and the magnified view at the same time is important. 

This need would be better addressed by this issue: https://github.com/Slicer/Slicer/issues/4383
All we need is to display boundaries of one slice view in another when slice views are parallel.
It may be implemented at the upcoming project week as part of this project: https://projectweek.na-mic.org/PW35_2021_Virtual/Projects/InteractiveSliceIntersections/

> One use case where the context might be helpful would be for segmenting m-mode cardiac ultrasound images, where we want to segment peaks and troughs precisely across a very wide horizontally oriented image. Normally with the Slicer zoom this would require a lot of panning to segment the whole image.

The problem is that the crosshair corresponds to the mouse pointer and not the markup control point position (there are typically a couple of mm offset, depending on where the use started to drag near the control point). Since neither the markups control point position nor the markup curve is visible in the magnified view, this view is actually not useful for adjusting markup control point positions.

I'm closing this issue for now, because there does not seem to be a way to fix this pull request and there is a good workaround (using built-in operating system features) and better long-term solution (using multiple slice views to view the same region at different zoom levels).

PR #5143, opened 2020-08-27
ENH: Allow switch between json and fcsv format for markups

lassoan:
Added a dedicated markups writer plugin that offers loading both json and fcsv formats for saving any markup node type. This allows easy switching between formats in save dialog:

image

Default format is now json for all types (as user can easily change in save dialog). If fcsv format is used then a warning icon is displayed about the lossy storage (no popup is displayed, the warning can be dismissed by closing the save dialog by clicking a button or hitting escape or enter key):

image


lassoan:
If there is no objection then I merge this today.

PR #5147, opened 2020-08-28
BUG: Fix duplicate python logging messages in the log

jamesobutler:
Python INFO, WARNING, and ERROR messages generated by the python logging module created duplicate entries as seen in the Slicer log file and in the Error Log Dialog.  This commit removes python logging messages from showing in the python console as they are already included in the Slicer log and should be viewed there.

When using logging across many scripted modules, this current duplication makes reading the Slicer log and using the Error Log Dialog **much** more difficult to understand.

### Before:
```
[DEBUG][Qt] 27.08.2020 22:56:22 [] (unknown:0) - Switch to module:  "ScreenCapture"
[INFO][Python] 27.08.2020 22:56:25 [Python] (C:/Users/james/AppData/Local/NA-MIC/Slicer 4.11.0-2020-08-26/bin/../lib/Slicer-4.11/qt-scripted-modules/ScreenCapture.py:805) - Write C:\Users\james\Documents\SlicerCapture\image_00004.png
[INFO][Stream] 27.08.2020 22:56:25 [] (unknown:0) - Write C:\Users\james\Documents\SlicerCapture\image_00004.png
```
image

### After:
```
[DEBUG][Qt] 27.08.2020 23:13:22 [] (unknown:0) - Switch to module:  "ScreenCapture"
[INFO][Python] 27.08.2020 23:13:25 [Python] (C:/Users/james/AppData/Local/NA-MIC/Slicer 4.11.0-2020-08-26/bin/../lib/Slicer-4.11/qt-scripted-modules/ScreenCapture.py:805) - Write C:\Users\james\Documents\SlicerCapture\image_00005.png
```
image

### DIFF
--- a/Base/Python/slicer/slicerqt.py
+++ b/Base/Python/slicer/slicerqt.py
@@ -64,9 +64,7 @@ def initLogging(logger):
   Initialize logging by creating log handlers and setting default log level.
   """
 
-  # Prints debug messages to Slicer application log.
-  # Only debug level messages are logged this way, as higher level messages are printed on console
-  # and all console outputs are sent automatically to the application log anyway.
+  # Prints debug, info, warning and error messages to Slicer application log.
   applicationLogHandler = SlicerApplicationLogHandler()
   applicationLogHandler.setLevel(logging.DEBUG)
   # We could filter out messages at INFO level or above (as they will be printed on the console anyway) by adding
@@ -74,18 +72,6 @@ def initLogging(logger):
   # but then we would not log file name and line number of info, warning, and error level messages.
   applicationLogHandler.setFormatter(logging.Formatter('%(message)s'))
   logger.addHandler(applicationLogHandler)
-
-  # Prints info message to stdout (anything on stdout will also show up in the application log)
-  consoleInfoHandler = logging.StreamHandler(sys.stdout)
-  consoleInfoHandler.setLevel(logging.INFO)
-  # Filter messages at WARNING level or above (they will be printed on stderr)
-  consoleInfoHandler.addFilter(_LogReverseLevelFilter(logging.WARNING))
-  logger.addHandler(consoleInfoHandler)
-
-  # Prints error and warning messages to stderr (anything on stderr will also show it in the application log)
-  consoleErrorHandler = logging.StreamHandler(sys.stderr)
-  consoleErrorHandler.setLevel(logging.WARNING)
-  logger.addHandler(consoleErrorHandler)
 
   # Log debug messages from scripts by default, as they are useful for troubleshooting with users
   logger.setLevel(logging.DEBUG)


jcfr:
Since the intent of:

https://github.com/Slicer/Slicer/blob/d31576db7802e99772e589ade784f769ed735d4f/Base/Python/slicer/slicerqt.py#L67-L71

was to only log Debug message, I am wondering if updating the ctor of `SlicerApplicationLogHandler` to accept a level parameter and the update the `emit` function to ignore message smaller than the expected level.

jamesobutler:
I think having the python logging only in the Slicer log (Report A Bug/Error Log Dialog) will be a more consistent way to tell users to review the log instead of viewing the python console to the review logging. `[Info][Python]` instead of `[Info][Stream]` I do think is better as well to know where it originated.

lassoan:
This is an interesting experiment, but it makes the Python console unusable (you cannot ask users of the Python console to look up response to their Python commands in some other GUI window). It also breaks Jupyter notebooks (as you cannot view the error log from the Jupyter notebook), running Slicer without main window (there is no error log window then), etc.

jamesobutler:
Response to python commands are still viewable. Python logging module responses are not included.

Below is an image of what the behavior looks like now
image

What won't happen anymore, but will still be reported in the log:
image
`[ERROR][Python] 28.08.2020 00:39:48 [Python] (<console>:1) - This is an error`

cpinter:
I'm surprised this has been merged so quickly without a chance for discussion, given that it is a really big change. If this is absolutely necessary for some reason, then please make it optional.

During development of Python based classes I always use the Python console to see the errors, because it's very apparent where they happened between the other messages (print, logging.info), and because that's a window I already use for the development. Having the error messages in the Error log window and the log file are separate places that I'd need to monitor continuously making the development workflow more fragmented and error-prone, and to be frank the Error log window is not usable enough for something like this (the user has to click a message to see it, and I find grouping somewhat undeterministic).

Please either revert this or add a Python or Application setting flag to use the old way asap. Until then I'll try not to update my working copies. Thank you!

jamesobutler:
@jcfr so the method should be what you described in the below?

> (1) updating the ctor of SlicerApplicationLogHandler to accept a level parameter
> (1) and updating the emit function to ignore message smaller than the expected level
> would achieve the desired goal.

jamesobutler:
Requirements:
- Console output so that developers can see everything that is going on easily without having to open the log.. VTK errors, Qt errors, python logging module generated messages and other python console output
- No double python logging in the actual Slicer Log file which is also displayed by the Error Report Dialog
- preferably the Slicer log detailing that anything generated from python is tagged with the [Python] tag. 
- Output detailing python interactor inputs by the user could be displayed as [Stream] but really are also [Python]

jcfr:
Requirements makes sense.

> I'm surprised this has been merged so quickly without a chance for discussion, given that it is a really big change. If this is absolutely necessary for some reason, then please make it optional.

@cpinter You are right, the change turns out to be larger than anticipated. Lesson learned.

We will be reverting  shortly.

In  the meantime, all errors are logged to the console expect the one associated with `logging.[error|info|...]` 

image

cpinter:
Thank you, @jcfr ! I am fine with any change as long as I can keep seeing these messages with some .slicerrc.py or other setting.

jamesobutler:
I’m fine with reverting as long as we know a path forward first about how to address the double python logging issue in the Slicer log. The double python logging issue has been very frustrating to the users that my group provides Slicer to.

jcfr:
> The double python logging issue has been very frustrating to the users that my group provides Slicer to.

Agreed. We should address this.

In addition of requirements outlined in https://github.com/Slicer/Slicer/pull/5147#issuecomment-682563384

Should message logged using `logging.error`, `logging.info`, ... be logged to the python console ? 

@cpinter Could you confirm that this the behavior you expect ? 

If yes, we just need to make sure these are not double logged.

cpinter:
> Should message logged using logging.error, logging.info, ... be logged to the python console ?

Yes, I think that is a requirement of many developers

> Could you confirm that this the behavior you expect ?

I agree with the points in the comment.

> If yes, we just need to make sure these are not double logged

Yes please. Again, to me an option is also acceptable. @lassoan since you also had doubts about this, please let us know what you think.

Thank you!

lassoan:
I agree that seeing the same message twice in the log may be unexpected (and I was slightly annoyed by it, too), but they have two different roles: 1. saving all console content, 2. recording all details of a Python error.

If you use a log viewer and filter for console messages then you can see what the user saw. If you want to see error location, etc. then you look at the message created by the error logger.

I like @cpinter suggestion, too, to make the behavior configurable of we see that there is strong differences in developer preferences, but first I would try to see if we can find a solution that everybody is comfortable with. 

> The double python logging issue has been very frustrating

Which log are you unhappy with: Python console, Application log window, Report a bug? Maybe we can find a way to make the error log easier to read but keeping Python errors displayed on the console.

jamesobutler:
The below items are the things that users are unhappy with.  For example seeing 6 warnings and 6 errors, but really there was only 3 unique errors and 3 unique warnings that actually occurred. Users are instructed to view the log using the filter options in the Error Log Dialog or also go to Report A Bug to view the log.  They are not expected to the view the python interactor to view logging messages.


[DEBUG][Qt] 27.08.2020 22:56:22 [] (unknown:0) - Switch to module:  "ScreenCapture"
[INFO][Python] 27.08.2020 22:56:25 [Python] (C:/Users/james/AppData/Local/NA-MIC/Slicer 4.11.0-2020-08-26/bin/../lib/Slicer-4.11/qt-scripted-modules/ScreenCapture.py:805) - Write C:\Users\james\Documents\SlicerCapture\image_00004.png
[INFO][Stream] 27.08.2020 22:56:25 [] (unknown:0) - Write C:\Users\james\Documents\SlicerCapture\image_00004.png
```
![image](https://user-images.githubusercontent.com/15837524/91516864-8a4bb580-e8ba-11ea-9903-aee2ff855633.png)

cpinter:
> they have two different roles: 1. saving all console content, 2. recording all details of a Python error.

> If you use a log viewer and filter for console messages then you can see what the user saw. If you want to see error location, etc. then you look at the message created by the error logger.

This is a very good point. I guess what is frustrating to one user/developer is the desired behavior to another. But the most important thing is to keep it consistent. So if the logging messages appear, then it will be twice in the log. So an option would make sense.

It is apparent that the opinion of these frustrated users is very important to Kitware, so I'm OK with their preference being the default and the rest of us having to adapt.

jcfr:
Updated Requirements (based on comment from @jamesobutler at https://github.com/Slicer/Slicer/pull/5147#issuecomment-682563384 and @cpinter at https://github.com/Slicer/Slicer/pull/5147#issuecomment-682586742):

* Console output so that developers can see everything that is going on easily without having to open the log. VTK errors, Qt errors, python logging module generated messages and other python console output
* No double python logging in the actual Slicer Log file which is also displayed by the Error Report Dialog
* Preferably the Slicer log detailing that anything generated from python is tagged with the [Python] tag.
* Output detailing python interactor inputs by the user could be displayed as `[Stream]` but really are also `[Python]`
* Calling `logging.info`, `logging.error`, ... should both log to file, error log and python console

Questions:

* Should the error log be improved to include an "Origin" filter. Origin would be "Qt", "Stream" or "Python" ?

jamesobutler:
Ideally things like `logging.error` would show in the stream but then not get counted twice in the Slicer log. So `[Error][Python] line_number “This is error”` would be present in the Slicer log but `[Error][Stream] “This is error”` would not be present. The latter doesn’t provide any new information that isn’t already included in the former. I don’t think it’s necessary to mention in the log that it was present in the stream.

PR #5151, opened 2020-08-28
BUG: Fix interactability of occluded markup control points

### LINKED ISSUE #5150: Interactability of occluded markup control points is dependent on order that actors were added
In the current nightly release, markups are still interactive if they are behind an actor that was added after the markup actor.
Markup added later are correctly occluded by previously added nodes.

This is caused by the fact that the vtkSelectVisiblePoints is executed during RenderOpaqueGeometry before the other actors have been rendered.

## Steps to reproduce

- Place markup fiducials in the scene
- Create an opaque object (model, segmentation, etc.)
- Position camera so that control points are occluded behind the object
- Control points can still be interacted with, even though they are not visible

## Expected behavior

Markups should be not be interactive if they are occluded by any other actors, regardless of the order that they were added to the renderer.

## Environment
- Slicer version: Slicer-4.11.0-2020-08-27

Sunderlandkyl:
Markups were still interactive if they were behind an actor that was added after the markup actor. Markup that were added later were correctly occluded by previously added nodes.
Fixed by disconnecting SelectVisiblePoints from the label pipeline and only manually calling it at the start of RenderOverlay().

fixes #5150

### DIFF
--- a/Modules/Loadable/Markups/VTKWidgets/vtkSlicerMarkupsWidgetRepresentation3D.cxx
+++ b/Modules/Loadable/Markups/VTKWidgets/vtkSlicerMarkupsWidgetRepresentation3D.cxx
@@ -94,8 +94,11 @@ vtkSlicerMarkupsWidgetRepresentation3D::ControlPointsPipeline3D::ControlPointsPi
   this->SelectVisiblePoints = vtkSmartPointer<vtkSelectVisiblePoints>::New();
   this->SelectVisiblePoints->SetInputData(this->LabelControlPointsPolyData);
   this->SelectVisiblePoints->SetTolerance(0.0); // we will set tolerance in world coordinate system
+  this->SelectVisiblePoints->SetOutput(vtkNew<vtkPolyData>());
 
-  this->PointSetToLabelHierarchyFilter->SetInputConnection(this->SelectVisiblePoints->GetOutputPort());
+  // The SelectVisiblePoints filter should not be added to any pipeline with SetInputConnection.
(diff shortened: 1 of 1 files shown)


Modules/Loadable/Markups/VTKWidgets/vtkSlicerMarkupsWidgetRepresentation3D.cxx
```diff
@@ -94,8 +94,9 @@ vtkSlicerMarkupsWidgetRepresentation3D::ControlPointsPipeline3D::ControlPointsPi
   this->SelectVisiblePoints = vtkSmartPointer<vtkSelectVisiblePoints>::New();
   this->SelectVisiblePoints->SetInputData(this->LabelControlPointsPolyData);
   this->SelectVisiblePoints->SetTolerance(0.0); // we will set tolerance in world coordinate system
+  this->SelectVisiblePoints->SetOutput(vtkNew<vtkPolyData>());
 
-  this->PointSetToLabelHierarchyFilter->SetInputConnection(this->SelectVisiblePoints->GetOutputPort());
+  this->PointSetToLabelHierarchyFilter->SetInputData(this->SelectVisiblePoints->GetOutput());
```
lassoan, line 99: Add a comment here to ensure that nobody will accidentally change the code back to SetInputConnection.

Sunderlandkyl: Done!

PR #5156, opened 2020-08-30
BUG: Fix duplicate python logging messages in the log for regular users

jamesobutler:
The motivation of this change is to improve user experience when viewing the Error Log Dialog and the slicer log as seen in the Report A Bug Dialog.  When there are, for example, 3 errors and 3 warnings, the user will see 3 errors and 3 warnings.  This is instead of 6 errors and 6 warnings where half were indicated from python and the other half were indicated from the stream.

This proposed change only shows python logging output in the console stream while in Developer Mode. For regular users not in developer mode they will no longer see double python logging (one from python and one from the stream).  

Developers will still have the double logging, but this appears to be desired behavior as they want knowledge of everything from the console stream in addition to everything from the python source.  The following code below indicates a method to prevent the double logging, but as it says, it would remove the line that includes the log file name and line number which is arguably the more important of the double logged messages.
https://github.com/Slicer/Slicer/blob/03113368de2e2afac71ade15ba0058b5e78dc39d/Base/Python/slicer/slicerqt.py#L72-L74

### Developer Mode:
```
[DEBUG][Qt] 27.08.2020 22:56:22 [] (unknown:0) - Switch to module:  "ScreenCapture"
[INFO][Python] 27.08.2020 22:56:25 [Python] (C:/Users/james/AppData/Local/NA-MIC/Slicer 4.11.0-2020-08-26/bin/../lib/Slicer-4.11/qt-scripted-modules/ScreenCapture.py:805) - Write C:\Users\james\Documents\SlicerCapture\image_00004.png
[INFO][Stream] 27.08.2020 22:56:25 [] (unknown:0) - Write C:\Users\james\Documents\SlicerCapture\image_00004.png
```
image

### Regular Users:
```
[DEBUG][Qt] 27.08.2020 23:13:22 [] (unknown:0) - Switch to module:  "ScreenCapture"
[INFO][Python] 27.08.2020 23:13:25 [Python] (C:/Users/james/AppData/Local/NA-MIC/Slicer 4.11.0-2020-08-26/bin/../lib/Slicer-4.11/qt-scripted-modules/ScreenCapture.py:805) - Write C:\Users\james\Documents\SlicerCapture\image_00005.png
```
image

### DIFF
--- a/Base/Python/slicer/slicerqt.py
+++ b/Base/Python/slicer/slicerqt.py
@@ -1,5 +1,4 @@
 import logging
-import sys
 
 import ctk
 import qt
@@ -48,6 +47,7 @@ class SlicerApplicationLogHandler(logging.Handler):
 
     def emit(self, record):
         try:
+            # Add application log entry
             msg = self.format(record)
             context = ctk.ctkErrorLogContext()
             context.setCategory(self.category)
@@ -56,8 +56,8 @@ class SlicerApplicationLogHandler(logging.Handler):
             context.setFunction(record.funcName)
             context.setMessage(msg)
             threadId = f"{record.threadName}({record.thread})"
-            slicer.app.errorLogModel().addEntry(qt.QDateTime.currentDateTime(), threadId,
-                                                self.pythonToCtkLevelConverter[record.levelno], self.origin, context, msg)
+            slicer.app.errorLogModel().postEntry(qt.QDateTime.currentDateTime(), threadId,
+                                                 self.pythonToCtkLevelConverter[record.levelno], self.origin, context, msg)
         except:
             self.handleError(record)
 
@@ -71,25 +71,13 @@ def initLogging(logger):
     # Prints debug messages to Slicer application log.
     # Only debug level messages are logged this way, as higher level messages are printed on console
     # and all console outputs are sent automatically to the application log anyway.
-    applicationLogHandler = SlicerApplicationLogHandler()
-    applicationLogHandler.setLevel(logging.DEBUG)
+    slicer.pythonApplicationLogHandler = SlicerApplicationLogHandler()
+    slicer.pythonApplicationLogHandler.setLevel(logging.DEBUG)
     # We could filter out messages at INFO level or above (as they will be printed on the console anyway) by adding
     # applicationLogHandler.addFilter(_LogReverseLevelFilter(logging.INFO))
     # but then we would not log file name and line number of info, warning, and error level messages.
-    applicationLogHandler.setFormatter(logging.Formatter('%(message)s'))
-    logger.addHandler(applicationLogHandler)
-
-    # Prints info message to stdout (anything on stdout will also show up in the application log)
-    consoleInfoHandler = logging.StreamHandler(sys.stdout)
-    consoleInfoHandler.setLevel(logging.INFO)
-    # Filter messages at WARNING level or above (they will be printed on stderr)
-    consoleInfoHandler.addFilter(_LogReverseLevelFilter(logging.WARNING))
-    logger.addHandler(consoleInfoHandler)
-
-    # Prints error and warning messages to stderr (anything on stderr will also show it in the application log)
-    consoleErrorHandler = logging.StreamHandler(sys.stderr)
-    consoleErrorHandler.setLevel(logging.WARNING)
-    logger.addHandler(consoleErrorHandler)
+    slicer.pythonApplicationLogHandler.setFormatter(logging.Formatter('%(message)s'))
+    logger.addHandler(slicer.pythonApplicationLogHandler)
 
     # Log debug messages from scripts by default, as they are useful for troubleshooting with users
     logger.setLevel(logging.DEBUG)
--- a/Base/QTGUI/Resources/UI/qSlicerSettingsPythonPanel.ui
+++ b/Base/QTGUI/Resources/UI/qSlicerSettingsPythonPanel.ui
@@ -17,7 +17,7 @@
    <item>
     <widget class="QGroupBox" name="ShellDisplayGroupBox">
      <property name="title">
-      <string>Shell Display</string>
+      <string>Python Interactor</string>
      </property>
      <property name="flat">
       <bool>true</bool>
@@ -46,14 +46,28 @@
       <item row="1" column="0">
        <widget class="QLabel" name="PromptFontLabel">
         <property name="text">
-         <string>Shell font:</string>
+         <string>Font:</string>
         </property>
        </widget>
       </item>
       <item row="1" column="1">
-       <widget class="ctkFontButton" name="pythonFontButton">
+       <widget class="ctkFontButton" name="ConsoleFontButton">
         <property name="text">
          <string/>
+        </property>
+       </widget>
+      </item>
+      <item row="2" column="0">
+       <widget class="QLabel" name="LogLevelLabel">
+        <property name="text">
+         <string>Log level:</string>
+        </property>
+       </widget>
+      </item>
+      <item row="2" column="1">
+       <widget class="QComboBox" name="ConsoleLogLevelComboBox">
+        <property name="toolTip">
+         <string>Log messages at this level and above are displayed in the Python interactor.</string>
         </property>
        </widget>
       </item>
--- a/Base/QTGUI/qSlicerApplication.cxx
+++ b/Base/QTGUI/qSlicerApplication.cxx
@@ -47,6 +47,7 @@
 #include <ctkErrorLogStreamMessageHandler.h>
 #include <ctkITKErrorLogMessageHandler.h>
 #include <ctkMessageBox.h>
+#include <ctkPimpl.h>
 #ifdef Slicer_USE_PYTHONQT
 # include "ctkPythonConsole.h"
 #endif
@@ -156,6 +157,9 @@ public:
 #ifdef Slicer_USE_QtTesting
   ctkQtTestingUtility*    TestingUtility;
 #endif
+#ifdef Slicer_USE_PYTHONQT
+  ctkErrorLogLevel::LogLevel PythonConsoleLogLevel;
+#endif
 };
 
 
@@ -177,6 +181,9 @@ qSlicerApplicationPrivate::qSlicerApplicationPrivate(
 #endif
 #ifdef Slicer_USE_QtTesting
   this->TestingUtility = nullptr;
+#endif
+#ifdef Slicer_USE_PYTHONQT
+  this->PythonConsoleLogLevel = ctkErrorLogLevel::Warning;
 #endif
 }
 
@@ -274,6 +281,18 @@ void qSlicerApplicationPrivate::init()
   this->ErrorLogModel->registerMsgHandler(new ctkITKErrorLogMessageHandler);
   this->ErrorLogModel->registerMsgHandler(new ctkVTKErrorLogMessageHandler);
   this->ErrorLogModel->setAllMsgHandlerEnabled(true);
+
+#ifdef Slicer_USE_PYTHONQT
+  // Make ITK, VTK, Qt error messages show up in the Python console
+  QSettings* userSettings = q->userSettings();
+  ctkErrorLogLevel::LogLevel level = ctkErrorLogLevel::logLevelFromString(userSettings->value("Python/ConsoleLogLevel").toString());
+  if (level >= 0)
+    {
+    q->setPythonConsoleLogLevel(level);
+    }
+  QObject::connect(this->ErrorLogModel.data(), SIGNAL(entryAdded(QDateTime, QString, ctkErrorLogLevel::LogLevel, QString, ctkErrorLogContext, QString)),
+    q, SLOT(logToPythonConsole(QDateTime, QString, ctkErrorLogLevel::LogLevel, QString, ctkErrorLogContext, QString)));
+#endif
 
   q->setupFileLogging();
   q->logApplicationInformation();
@@ -382,6 +401,11 @@ QSettings* qSlicerApplicationPrivate::newSettings()
 
 //-----------------------------------------------------------------------------
 // qSlicerApplication methods
+
+#ifdef Slicer_USE_PYTHONQT
+CTK_GET_CPP(qSlicerApplication, ctkErrorLogLevel::LogLevel, pythonConsoleLogLevel, PythonConsoleLogLevel);
+CTK_SET_CPP(qSlicerApplication, ctkErrorLogLevel::LogLevel, setPythonConsoleLogLevel, PythonConsoleLogLevel);
+#endif
 
 //-----------------------------------------------------------------------------
 qSlicerApplication::qSlicerApplication(int &_argc, char **_argv)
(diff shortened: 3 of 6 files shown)


cpinter:
Thank you for offering an alternative for removing the log to console function! To me having access to the python warnings and errors only in developer mode is an acceptable solution.

When working with users using mostly Python-based extensions and modules we usually asked them to look at the Python window for errors because it was much more probable to have Python related errors than core C++ related ones, and the red text was very easy to spot. But although users seem to find the Python window easier than the log for some reason, I think it is an OK sacrifice to ask them for logs.

lassoan:
Fully agree with @cpinter . The python console was just a much better source of spotting Python errors than the "Log messages" windows, but as an experiment, we can try to hide them in non-developer mode.

@jamesobutler Will uncaught exceptions still show up in the Python console for users (so that we can tell users to check out the Python console when something is wrong)?

To be honest, I would rather see more errors showing up in the Python console. For example, when I run a C++ logic function in Python then it may log an error message in the "Log window", but I would need to take my hand over to the mouse, click that button, scroll to the error message to see it - which I rarely do. But maybe we should just fix the "Log window" - make it dockable, make it auto-scroll to last message, make it easier to see and copy all messages at once, etc. and then we would not need to rely on the Python console that much.

jamesobutler:
@lassoan Yes errors such as tracebacks still show up.
image


> then we would not need to rely on the Python console that much.

I would encourage improving the items in Slicer that are actually referred to as the log for communicating to users how to see if something is wrong. This being instead of using the python console for all logging issues (not just python) as a workaround to avoid viewing the real Slicer log in Report a Bug or in the Error Log dialog.

pieper:
FWIW, when developing I typically open Slicer in a terminal so I can see error messages from VTK and other low C++ code.  I wouldn't need to see it in the python console too, but I guess I would not mind.  I rarely use the error log window.

I'm going to vote 👎  on the idea of changing logging behavior in developer vs. non-developer mode.  It's very common to ask users to report what they see when trying to track down and issue and having different behavior in different modes will only make that more confusing.

jamesobutler:
@pieper With this change, if in DeveloperMode, you would see all typical logging. Nothing would be lost.

> It's very common to ask users to report what they see when trying to track down and issue and having different behavior in different modes will only make that more confusing.

I would hope the consistent message to this would be "to review the log". The log can be filtered in the Error Log dialog or the full log can be viewed in the Report A Bug dialog.

pieper:
> I would hope the consistent message to this would be "to review the log". The log can be filtered in the Error Log dialog or the full log can be viewed in the Report A Bug dialog.

If we go that way we do need to put in the effort to make the log dialog more usable.  Having the full log displayed when the window opens would be a good start.  Currently the first operation for me is always typing Ctrl-A.  Personally I have never found it helpful to filter the logs, but I don't mind that the feature is there if others use it.

Similarly it might be good if the icon in the lower right were invisible unless there is a real error message in the log.  Currently it is a subtle change from disabled to enabled.  Could also be good to have a warning icon that appears where there is warning content, and no icon if there are only info messages.

jamesobutler:
Log improvement ideas from @lassoan 
- make it auto-scroll to last message
  - I think this could definitely be implemented for both the Error Log dialog and for the log section of the Report A Bug dialog.
- maybe we should just fix the "Log window", make it dockable
  - Not sure where in the main window would be best for this change. Near the error log dialog button in the QStatusBar in the same area as where the python interactor is docked would seem best, but could cause confusion and I think only one DockWidget can be in the bottom area.
- make it easier to see and copy all messages at once,
  - Is this regarding copying from the Error Log dialog? Report A Bug dialog already has a button to copy log messages to the clipboard for easy pasting elsewhere.

Log improvement ideas from @pieper 
- Having the full log displayed when the window opens would be a good start
  - Some logs are too long to be displayed all at once, but I think opening it with it auto-scrolled to the last message would help. Just like the python console or other console can't show everything at once, but will have a scroll bar when needed.
- it might be good if the icon in the lower right were invisible unless there is a real error message in the log. Currently it is a subtle change from disabled to enabled
  - This is a good idea to only show the button that opens the error log dialog from the QStatusBar only when there are actual errors/warnings
- have a warning icon that appears where there is warning content, and no icon if there are only info messages.
  - As related to the previous idea, would the button in the QStatusBar be shown if there are only info messages? Then instead of no info icon, it would just be that the button wouldn't be visible.


Good ideas to help improve the log experience for regular users! Do you think these changes should be incorporated into this PR or organized into a new git issue with follow up PRs?

lassoan:
> I would hope the consistent message to this would be "to review the log". The log can be filtered in the Error Log dialog or the full log can be viewed in the Report A Bug dialog.

I agree, "view the log" should be done in the "Log view" window, it's just way too inconvenient to use it now.

>> make it auto-scroll to last message
>
> I think this could definitely be implemented for both the Error Log dialog and for the log section of the Report A Bug dialog.

"Report a bug" is tuned for users to report errors and "Log messages" window is for interactive use. I think we can keep it this way.

>>  make it dockable
>
> Not sure where in the main window would be best for this change

We could do exactly the same as for the Python console.

>> make it easier to see and copy all messages at once,
>
> Is this regarding copying from the Error Log dialog? 

I was thinking about adding copy feature to "Log messages" window, but maybe we don't need it. We can keep that function in the "Report a bug" window (as it has more information, such as line numbers).

> I think only one DockWidget can be in the bottom area.

You can have any number of dock widgets in any area, split or stacked in any way. However, since log window is more of a "tall" window, it is probably better to dock it to the right by default.

>> Having the full log displayed when the window opens would be a good start
>
> Some logs are too long to be displayed all at once, but I think opening it with it auto-scrolled to the last message would help. Just like the python console or other console can't show everything at once, but will have a scroll bar when needed

Constantly merging all error messages could have a performance impact and could be hard to find errors because there is no color highlighting (which by the way would be a good improvement, too). Maybe when the window is opened (or a "Select new messages" button in the window is clicked) then all unseen messages could be selected.

>> have a warning icon that appears where there is warning content, and no icon if there are only info messages.
>
> As related to the previous idea, would the button in the QStatusBar be shown if there are only info messages? Then instead of no info icon, it would just be that the button wouldn't be visible

I would prefer to have at least some very subtle gray icon to open the window anytime. It would make it clear that there are no errors/warnings and users could find the log window the same place all the time. With the icon, the main issue is that we have false alarms that we haven't fixed yet (we used to have lots of subject hierarchy item IDs not found after closing the scene, etc.).

Probably we should hide info level messages by default.

cpinter:
Great discussion! I support the idea of relying more on the Error Log window than so far. In that case the errors may even be kept out of the Python interactor (@pieper is right in that a difference in modes just introduces confusion - something I thought of as well but didn't see a better alternative) It has been alright for a debugging tool, but quite unusable for using during development.

What I'd consider essential before being able to use it for development:
* Scroll to bottom automatically
* Dockability
* Show more of each message by default. One of the main reasons for me not using this window is that I need to click each error message to see what they actually are. See this example:
image
I think a simple solution would be to show the messages in their entirety, however, it would make the rows too high in docked mode. An option:
  * Add a "simple view" mode, where instead of a table more than half of which is occupied by metadata (date, level, source), the actual messages would be shown in simple text. So there would be two modes: Simple and Table
  * Enable message filtering in this simple text view as well (just keep the buttons on the top - except for the "All" button, which just takes up valuable space in docked mode)
  * Color-code the levels in the simple text mode (debug: gray, info: black (white in dark mode), warning: orange, error: red)

Nice to have:
* If we keep the duality of viewing and reporting errors, then there should be a link from the Error Log window to the report an error window. A small font blue link would suffice.

jamesobutler:
So @pieper, @lassoan, @pieper, with improvements to the Error log window would it then be ok to remove the logging messages from the python console? This was something I was trying to accommodate in the DeveloperMode option for @cpinter, but the concern about it being different in one mode versus another would indicate the move to remove it. Removing it will allow for there to not be duplicate messages in the full Slicer log (from the stream and one from python) which is my primary main concern. This duplication is what has confused users where 6 errors and 6 warnings are really just 3 unique errors and 3 unique warnings.

pieper:
Thanks for all the thought about this.  This will really improve things for users and developers.

I would want python traceback errors and print statements to show up in the python console and in the error log.  Any other messages, like `vtkErrorMacro` text can only be in the error log and `stdout`/`stderr`.

I assume we can do that without double messages anywhere.

+1 for having a simple mode default for the error log, basically just a text area.  The messages could be color coded by log level and it could autoscroll to the latest.  Can have a checkbox to enable advanced mode with filtering and searching.

+1 for having the error log button visible all the time, being green or gray when only info text, but turning to yellow triangle on warning and red hexagon when error text is present.

+1 for a blue text link in the error log taking you to the report a bug dialog

+1 for making it docking.  I would probably use the bottom of the main app window with python on one side and error log on the other, ideally with a splitter so the python window could be wider.

Note we also have the [status bar](https://doc.qt.io/qt-5/qstatusbar.html) that we could use to display an extra level of alert with any error text. (try it with  `slicer.util.mainWindow().statusBar().showMessage("Feedback", 2000)`.  That would be handy to see right below the python console if there's an error state.

jamesobutler:
Ok so I will proceed with the method originally introduced in https://github.com/Slicer/Slicer/commit/4ccf136921f6ecb22e546115bc8f56ea6b6ba8be which still supports python tracebacks and print messages showing in the python console, but python logging messages (info, warning, error) will no longer be visible there. This allows consistent console output between developer mode and regular mode instead of what was proposed in the first commit of this PR which was to remove python console logging output from the console only when not in DeveloperMode.

Then additional improvements will be made to the various Error Log as discussed above.

pieper:
Thank you! 🙏

cpinter:
@jamesobutler Can you please make sure you integrate these changes all at once? Several weeks without seeing the python errors in a convenient manner still seem like a lot. Thank you!

jamesobutler:
Yes, I will plan on adding all the changes together.

jamesobutler:
@lassoan Is there any concern with things related to SlicerJupyter or no-main-window actions with these proposed changes?

lassoan:
I guess we will just change the _default_ log level, so users can still adjust the log level to make logging.warning and logging.error messages sent to stderr.

jamesobutler:
Which log level specification would you change? Because the changes in https://github.com/Slicer/Slicer/commit/4ccf136921f6ecb22e546115bc8f56ea6b6ba8be would be included again which removes the code that does `# Prints error and warning messages to stderr `

lassoan:
We could keep the log handler and just adjust the default levels. Currently, "debug" level messages are not displayed in the console (only logged in log view). My understanding is that you would increase this to "error" level (so that even errors would only show up in the log view). Users should be able to change this log level in the handler back to debug (or even lower).

lassoan:
I've had a fresh look at this.

Using the recently added `slicer.app.showConsoleMessage()` method we can log to the application console (Slicer process output) without logging into the Python console - achieving the same behavior for Python log messages as for VTK log messages.

If a developer wants to see the Python log messages in the Python console then this command can be added to the application startup script: 

    slicer.pythonApplicationLogHandler.logToConsoleMode = SlicerApplicationLogHandler.PYTHON_CONSOLE_OUTPUT

To disable console output completely:

    slicer.pythonApplicationLogHandler.logToConsoleMode = SlicerApplicationLogHandler.NO_CONSOLE_OUTPUT

@jamesobutler would this fulfill your needs?

I'm also thinking about an even better, more consistent design: send all log messages (VTK, Python, Qt, etc) to the Python console from the CTK logger. We could then see all the errors and warnings resulting from the executed command.

jamesobutler:
Thanks for your work on this! I will try it out tomorrow.

Maybe we can also revisit #5176 to improve the error log view as well to make it more usable as a way to filter out the errors from a long log file. This PR was trying to help by removing the duplication so this would be a step in the right direction.

lassoan:
I've tested [logging of Qt, ITK, VTK messages into the Python log](https://github.com/lassoan/Slicer/tree/log-vtk-errors-to-python-console) and the result is quite nice:

image

It would make sense to add this feature to Slicer core. If we show ITK, VTK, Qt log messages in the Python console then we should show Python log messages there, too.

jamesobutler:
Yes that is very nice looking. See all the errors of stuff in the python console would be very helpful. Of course as long as it isn’t duplicated in the application log.

jamesobutler:
@lassoan With the work you've done for displaying all the logging in the python interactor https://github.com/Slicer/Slicer/pull/5156#issuecomment-1277051461, when doing Slicer development, would we even need the separate Windows console window since it would now be the same as just looking at the python interactor? That would be nice since we would see all the log messages, but see it in rendered colored format and also have it being interactive for coding in the same window as well.

lassoan:
I've implemented the new design that collects all messages (coming from Python, VTK, ITK, Qt, ... loggers) and displays them in the Python console. These messages could be hard to distinguish from the output of the command; and since messages are logged via a queued connection, they appear after the output of the commend. Therefore, I added a prefix to each message (`^` indicate that the message is for the code above):

image

What do you think? How does it look?

We could make the Python console logging level configurable (not display any logs, or only errors, or even debug/info messages).

jamesobutler:
I haven’t tried it but I wonder if coloredlogs (https://pypi.org/project/coloredlogs/) could be used to help distinguish the different messages levels and distinguish from the command output.

> We could make the Python console logging level configurable (not display any logs, or only errors, or even debug/info messages).

Yes that would be good because maybe some would find it too distracting if a certain command was logging a lot while they were trying various things out Causing their previous commands to be pushed out of view.

lassoan:
`coloredlogs` only works on an ANSI terminal, so we would need to implement ANSI control character sequences in the Python console to take advantage of that. It would be easier to add coloring API directly, but it is already somewhat difficult to find 3 well-visible, different-enough colors that all users like, so it would be hard to find another 3 that are slightly different from the existing ones. So, coloring could be used as an additional hint, but it would require some effort to find the new colors (for both light and dark theme) and implement the logic, so I would not do it in the first phase. (You can give it a go if you have time and interest in trying it out.)

> > We could make the Python console logging level configurable (not display any logs, or only errors, or even debug/info messages).
> 
> Yes that would be good because maybe some would find it too distracting if a certain command was logging a lot while they were trying various things out

I've added an option for this in Application Settings / Python / Log level.

lassoan:
I've force-pushed the log redirection to the Python console. It is ready for review/testing.

The feature allows displaying errors from all sources (not just from Python but VTK, ITK, QT, ...), and the log display level is configurable in application settings / Python. It also solves duplicate logging of Python log messages in the application log.

lassoan:
I'm thinking about committing this to make it easier for everyone to try it. (We can always update or revert the commit if it does not turn out to work well.)

pieper (approved):
Yes, let's merge and try it out.

pieper:
But there are build errors from CI that look real.

lassoan:
> But there are build errors from CI that look real.

I think this error is due to CI not using the latest CTK:

    error: ‘logLevelFromString’ is not a member of ‘ctkErrorLogLevel’

lassoan:
I'll merge this now and will keep an eye on the dashboard and fix if any issues come up.

PR #5158, opened 2020-09-01
ENH: Add customization of markup text font and background appearance

Sunderlandkyl:
Adds additional options to the vtkMRMLMarkupsDisplayNode to customize text display:
- TextFontFamily
- TextFontFamily;
- TextBold;
- TextItalic;
- TextShadow;
- TextBackgroundColor
- TextBackgroundVisibility;
- TextBackgroundOpacity

image

### DIFF
--- a/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsDisplayNode.cxx
+++ b/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsDisplayNode.cxx
@@ -29,12 +29,19 @@
 #include <vtkNew.h>
 #include <vtkObjectFactory.h>
 #include <vtkPiecewiseFunction.h>
+#include <vtkTextProperty.h>
+#include <vtksys/SystemTools.hxx>
 
 // STL includes
 #include <sstream>
 
 const char* vtkMRMLMarkupsDisplayNode::LineColorNodeReferenceRole = "lineColor";
 const char* vtkMRMLMarkupsDisplayNode::LineColorNodeReferenceMRMLAttributeName = "lineColorNodeRef";
+
+static const int SHADOW_H_OFFSET_INDEX = 0;
+static const int SHADOW_V_OFFSET_INDEX = 1;
+static const int SHADOW_BLUR_RADIUS_INDEX = 2;
+static const int SHADOW_COLOR_INDEX = 3;
 
 //----------------------------------------------------------------------------
 vtkMRMLNodeNewMacro(vtkMRMLMarkupsDisplayNode);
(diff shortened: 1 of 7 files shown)


Sunderlandkyl:
Changing to use a vtkTextProperty to contain all text display options makes sense to me.

jcfr:
Thanks @Sunderlandkyl  for accommodating, these changes are great.

Sunderlandkyl:
@lassoan I've updated the MRMLNode to use a vtkTextProperty object.

If you think that we will need a WriteXML/ReadXML macro for vtkTextProperty objects for other nodes, then I'll go ahead and convert the serialization functions to Read/WriteXML macros.

lassoan (commented):
This is great, thank you. I have only one small comment (Set/SetAndObserve) and a few trivial ones.

lassoan (commented):
Thank you very much, I've added a few mostly trivial comments. We could slightly simplify the code if GetTextProperty() method would be available in the public API but SetAndObserveTextProperty() would be kept protected. Because then we could easily ensure that the returned text property is always valid and we would not need extra checks and default value initializations at a number of places. We'll probably add a few more text properties later, so it is worth investing time into getting a simpler, more straightforward API.

Sunderlandkyl:
> We could slightly simplify the code if GetTextProperty() method would be available in the public API but SetAndObserveTextProperty() would be kept protected.

The one reason that I could think that we might want to keep SetAndObserveTextProperty public would be so that you could set the same vtkTextProperty on multiple markups, and be able to update them all without having to update their vtkTextProperty individually.

lassoan:
> The one reason that I could think that we might want to keep SetAndObserveTextProperty public would be so that you could set the same vtkTextProperty on multiple markups, and be able to update them all without having to update their vtkTextProperty individually.

OK, if you think this would be a useful feature then keep it like this, just clarify the documentation (what happens if nullptr is set).

I think we should still make GetTextProperty always return a valid text property (by replacing nullptr with a pointer to a default text property object; either when calling SetTextProperty(nullptr) or when calling GetTextProperty()). It would make the display node a bit more complex/opaque, but simplify classes that use it, and it would also centralize where default text properties are set.

Sunderlandkyl:
OK, I have no particular use case that requires setting the same ```vtkTextPropery``` to multiple nodes, so I've just removed the function to set it. The ```GetTextProperty``` method should now always return a valid pointer,

lassoan (approved):
Thank you, it looks good now. Sorry for all the nitpicks.

pieper:
If you think it would help to use the same text property for multiple nodes, then it would be good to make a new `vtkMRMLTextPropertyNode` that others could observe through references so the relationship would be preserved through scene save and restore.  But I agree I'm not sure it will come up often enough to be worth that effort.

lassoan:
Saving/reloading is a good point - and now that I think about it, sequences, scene views, and undo/redo may be confused by shared data. It is probably safer to keep object members of different nodes independent (as it is done in the final implementation).

Sunderlandkyl:
Thanks for reviewing! Better to do things right the first time then fixing them later.

PR #5162, opened 2020-09-02
COMP: Update FindVcvars with MSVC 1927 support

jamesobutler:
Based on the changes made in https://github.com/scikit-build/cmake-FindVcvars/commit/67b96c3a8dd0db2f4d04c989db12f273a02580af


jcfr (approved):
Thanks for updating https://github.com/scikit-build/cmake-FindVcvars 😄

PR #5165, opened 2020-09-02
BUG: Fix deletion of segments when running smoothing effect

Sunderlandkyl:
When running the smoothing effect, the current segment could be deleted if the mask was an incorrect datatype.
The mask image was copied from the modifier labelmap, and could be a data type other than unsigned char, which would cause the vtkImageMask filter to fail.
Fixed by setting the mask datatype to VTK_UNSIGNED_CHAR.

See topic here: https://discourse.slicer.org/t/13295


lassoan (approved):
Nice fix, thank you!

PR #5166, opened 2020-09-02
COMP: Use CMake wrapping modules provided by vtkAddon

jcfr:
This commit simplifies future maintenance removing the copy of the modules
available in Slicer by using of the one provided by vtkAddon.


jcfr:
@lassoan @pieper  Any objection in de-duplicating our custom wrapping macros using only the copy made available in vtkAddon
project ?

pieper:
No objection from me. 👍

lassoan:
No objection from me either, good idea.

PR #5172, opened 2020-09-03
COMP: Fix Linux compile error in vtkMRMLMarkupsDisplayNode

Sunderlandkyl:


lassoan (approved):
thank you, looks good

PR #5173, opened 2020-09-04
DOC: Improve Slicer logo appearance in ReadTheDocs

lassoan:
SVG image is scaled quite badly in Chrome - black strokes of the Slicer logo look pixelated.

Switch to using a png image instead.

image


jcfr:
This makes sense.

Do we anticipate the `3D-Slicer-Mark.svg` to be used in the application itself ? 

If yes, tell me where we should put it and can make sure it is copied over when we build the documentation

lassoan:
Qt does a much better job scaling and rasterizing SVGs and you can already use SVGs as icons. When we replace the icon for Slicer5 then we can probably move this icon to the appropriate location and use it. Until then I think we can just leave it where it is.

jcfr (approved):

PR #5177, opened 2020-09-07
BUG: Fix segment color not updating in subject hierarchy

lassoan:
When segment color was changed by calling segment->SetColor then subject hierarchy widget was not updated.

Fixed by calling vtkMRMLSubjectHierarchyNode::SubjectHierarchyItemModifiedEvent when segment is modified.


fedorov:
Thank you @lassoan! I tested, and the issue appears to be resolved!

PR #5178, opened 2020-09-08
COMP: avoid JDK warning from Swig on mac

pieper:
Fixes the issue described here: https://github.com/SimpleITK/SimpleITK/issues/1190


lassoan (approved):
Looks good to me.

PR #5179, opened 2020-09-09
Minor logging improvements

jamesobutler:
These are pulled out from #5176 and represent small individual improvements to logging outside of the conversation about how the console view of logging messages should be changed.
- The error log toolbutton in the status bar now has an icon that reflects the highest logged level state since viewing the log widget.
- The log text in the ErrorReport Dialog area is no longer editable. I've made it read-only as I don't think it was intended to be editable there.

### DIFF
--- a/Base/QTApp/Resources/UI/qSlicerErrorReportDialog.ui
+++ b/Base/QTApp/Resources/UI/qSlicerErrorReportDialog.ui
@@ -62,6 +62,9 @@
        <height>50</height>
       </size>
      </property>
+     <property name="readOnly">
+      <bool>true</bool>
+     </property>
     </widget>
    </item>
    <item row="4" column="0" colspan="2">
@@ -83,6 +86,13 @@
        </property>
        <property name="autoDefault">
         <bool>false</bool>
+       </property>
+      </widget>
+     </item>
+     <item>
+      <widget class="QCheckBox" name="LogFileEditCheckBox">
+       <property name="text">
+        <string>Edit Log</string>
        </property>
       </widget>
      </item>
--- a/Base/QTApp/qSlicerErrorReportDialog.cxx
+++ b/Base/QTApp/qSlicerErrorReportDialog.cxx
@@ -62,6 +62,7 @@ qSlicerErrorReportDialog::qSlicerErrorReportDialog(QWidget* parentWidget)
   QObject::connect(d->RecentLogFilesComboBox, SIGNAL(currentIndexChanged(QString)), this, SLOT(onLogFileSelectionChanged()));
   QObject::connect(d->LogCopyToClipboardPushButton, SIGNAL(clicked()), this, SLOT(onLogCopy()));
   QObject::connect(d->LogFileOpenPushButton, SIGNAL(clicked()), this, SLOT(onLogFileOpen()));
+  QObject::connect(d->LogFileEditCheckBox, SIGNAL(clicked(bool)), this, SLOT(onLogFileEditClicked(bool)));
 
   connect(d->ButtonBox, SIGNAL(rejected()), this, SLOT(close()));
 
@@ -101,4 +102,11 @@ void qSlicerErrorReportDialog::onLogFileOpen()
 {
   Q_D(qSlicerErrorReportDialog);
   QDesktopServices::openUrl(QUrl("file:///"+d->RecentLogFilesComboBox->currentText(), QUrl::TolerantMode));
+}
+
+// --------------------------------------------------------------------------
+void qSlicerErrorReportDialog::onLogFileEditClicked(bool editable)
+{
+  Q_D(qSlicerErrorReportDialog);
+  d->LogText->setReadOnly(!editable);
 }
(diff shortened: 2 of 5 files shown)


lassoan (commented):
Thank you.

Icon update looks good!

> The log text in the ErrorReport Dialog area is no longer editable. I've made it read-only as I don't think it was intended to be editable there.

The log text was editable to allow text selection using keyboard (you can select using mouse, but that's really slow compared to Shift+PageUp/Dn for long logs) and allow user to sanitize content (remove patient information, cut out irrelevant parts, etc.) without relying on an external editor. Unless there is a specific concern, I would leave it editable.

jamesobutler:
> The log text was editable to allow text selection using keyboard (you can select using mouse, but that's really slow compared to Shift+PageUp/Dn for long logs) 

You can still ctrl-a (select all) when the QPlainTextEdit is read-only. Is there another keyboard action that you are referring to that isn't available when it is read-only? For non-keyboard based selection of the entire text, I thought that is what the "Copy log messages to clipboard" button is for. Is there a situation for selecting a large amount of text (too cumbersome for mouse click), but that isn't the entire log?

> allow user to sanitize content (remove patient information, cut out irrelevant parts, etc.) without relying on an external editor.

Isn't that what the "Open log file in editor" button is for? Why shouldn't we rely on an external editor? Doesn't every OS have some standard editor?

lassoan:
> You can still ctrl-a (select all) when the QPlainTextEdit is read-only. Is there another keyboard action that you are referring to that isn't available when it is read-only?

You cannot use caret for selecting text. Selecting more text than fits in the window but not all the text is really hard in read-only mode.

> Isn't that what the "Open log file in editor" button is for?

Shell edit is not set up correctly in all environments, so we cannot rely on it. Of course, select all, copy all, etc. then paste into an editor, edit there, and copy the result is doable, but a few extra steps.

Is there any particular problem with keeping the text editable?

A more important usability issue is that users don't realize that they can select logs of previous sessions. Would you have a suggestion for making this more obvious? Maybe converting the combobox to a list or table and maybe move it to the side (so that the user can see multiple entries) would help?

jamesobutler:
> Is there any particular problem with keeping the text editable?

The issue is users accidentally deleting fields in the log area and then providing an edited log without knowledge

jamesobutler:
Could there be an button option to edit the viewed log or not? Then that would make it a clear deliberate action if the user wants to edit it.  Just opening the dialog it appears to be viewable text, not editable text.

lassoan:
> The issue is users accidentally deleting fields in the log area and then providing an edited log without knowledge

In Slicer, users are supposed to check and sanitize the log before submission. If an external editor is used for that then corruption can happen in the external editor the same way, you just have a few extra points of failure because you need to get the text to the external editor.

If in a custom application you determine that log review/sanitizing is no needed (e.g., there are no privacy concerns because log is transferred inside the same organization) then a direct web upload function could be added and then it would make sense to disable editing by default. An "Edit in place" checkbox could be added then (unchecked by default) or editing could be disabled completely.

jamesobutler:
To improve regular Slicer behavior of viewing the log, would it be acceptable to add a checkbox/button to enable editing of the log where editing is not on by default?

lassoan:
> To improve regular Slicer behavior of viewing the log, would it be acceptable to add a checkbox/button to enable editing of the log where editing is not on by default?

It would be one extra click (and some more clutter in the GUI), but it would be fine for me.

jamesobutler:
I've updated the Error Report Dialog update to add the checkbox that will make editing the log a deliberate action. Users can edit the log by checking the "Edit Log" checkbox.
image

lassoan:
Thank you for the updates!

In-place editing of logs looks good.

Showing a strong "Info" icon when there can be errors/warnings in the log is a bit confusing to me. It is displayed even when there are no info messages at all. This change will make things even less intuitive with the docked log window, as you would change the icon when the user would just click into the window. I think it would be better to show an error/warning icon if there are any errors or warnings in the window, not just since it was last seen (only clear it when you click "Clear"). It would be even better to show number of errors/warnings (the same error/warning/info count could be used on the filter buttons). Maybe even info messages right next to the icon(s) as it is done in VS Code:

image

We should merge the "Report a bug" window changes. I would integrate the error window changes with all the other improvements that you have started, as these log window changes by themselves would not clearly improve the current behavior.

lassoan:
I've merged the https://github.com/Slicer/Slicer/pull/5179/commits/7e6a62dccd1b1f17fdb0b8e2946a75286af6d71b commit. I would suggest to close this PR and continue in #5176. I will review those changes along with proposed changes in CTK soon.

jamesobutler:
> Showing a strong "Info" icon when there can be errors/warnings in the log is a bit confusing to me. It is displayed even when there are no info messages at all. 

I could easily change it so that it doesn't reset after displaying the current error report dialog.  I was thinking that this change to the icon was incrementally better than showing the "Error" icon when there are only warning messages that are present which is the current case.  The other work related to the dock widget or console has TBD timeline on it based on my free time.

jcfr:
@jamesobutler  Thanks again for your patience and for all the work you put into Slicer :pray:

jamesobutler:
Closing this now and will look into error and warning counts displayed in the QStatusBar

PR #5183, opened 2020-09-11
ENH: Improve markup interaction handle appearance

Sunderlandkyl:
Reduces the total interaction widget size, as well as the individual size of the handles by 50%.
Increases the resolution of the sphere and tube sources.

Appearance before:
image

Appearance after:
image


lassoan (approved):
Looks good to me, thank you!

Issue #5182, opened 2020-09-11
vtkSlicerMarkupsWidgetRepresentation3D not exposed in python


pieper:
`markupsDisplayableManager.GetWidget(fiducialsNode.GetDisplayNode())` returns `MRMLDisplayableManagerPython.vtkMRMLAbstractWidget` not `vtkSlicerMarkupsWidgetRepresentation3D` meaning that you can't call any of the subclass methods.

## Steps to reproduce

Fresh scene, add a fiducial, then paste this in console:
```
fiducialsNode = slicer.util.getNode('F')
layoutManager = slicer.app.layoutManager()
threeDWidget = layoutManager.threeDWidget(0)
threeDView = threeDWidget.threeDView()
className = "vtkMRMLMarkupsDisplayableManager"
markupsDisplayableManager = threeDView.displayableManagerByClassName(className)
fiducialsWidget =  markupsDisplayableManager.GetWidget(fiducialsNode.GetDisplayNode())
fiducialsRepresentation =  fiducialsWidget.GetRepresentation()

print(fiducialsWidget.GetClassName())
print(fiducialsRepresentation.GetClassName())
print(fiducialsWidget.__repr__())
print(fiducialsRepresentation.__repr__())

assert fiducialsWidget.GetClassName() in fiducialsWidget.__repr__()
assert fiducialsRepresentation.GetClassName() in fiducialsRepresentation.__repr__()
```
## Expected behavior

The widget and representations types should match their class names. 

## Environment
- Slicer version: Slicer-4.11-09-08
- Operating system: Mac + 10.15

lassoan:
Only standard "MRML", "Logic", "Widgets" libraries are imported into `slicer` namespace. To access classes in other libraries, such as widget representations in vtkSlicerMarkupsModuleVTKWidgets, you need to import them manually. Without that, the Python wrapper cannot get to the exact class, but gets stuck somewhere higher in the hierarchy.

`import vtkSlicerMarkupsModuleVTKWidgetsPython` before you try to access the class will fix the problem.

I guess the main question is if these classes are part of the official Python API of Slicer. If they are, then probably it would be better to import them into the Slicer namespace. @jcfr what do you think?

pieper:
Yes, importing is a good workaround 👍 

If they are part of the public C++ API I would say yes, they should be imported to be consistent with the rest of the application.

jcfr:
Makes sense. If not alreay done, I will submit a PR tmrw.

[...]

PR #5184, opened 2020-09-12
BUG: Ensure Slicer VTKWidget classes are available in python. See #5182

jcfr:


lassoan (approved):
Looks good to me (actually, I've just learned from you how this can be done)

PR #5186, opened 2020-09-15
COMP: Attempt to fix extension package upload updating timeout from 120 to 240

jcfr:
Despite of having an error like the following reported on CDash, the
package is still available in the extensions manager. This commit updates
the timeout to provide more time to the server for acknowledging the
upload and avoid to report a failure.

Co-authored-by: Sam Horvath <sam.horvath@kitware.com>


sjh26 (approved):

lassoan:
There is also a `TIMEOUT` value (https://cmake.org/cmake/help/v3.0/command/file.html). If adjusting INACTIVITY_TIMEOUT does not fix the problem then we may try to change TIMEOUT value as well.

PR #5188, opened 2020-09-17
BUG: Fix crash when scaling markups widget in slice view

jcfr:
This commit ensures the transformation reference point can be retrieved
for both 2D and 3D representations.


lassoan (approved):
Looks good to me, thank you.

PR #5189, opened 2020-09-18
ENH: Remove Ctrl-A shortcut for loading data

lassoan:
Ctrl-O is the standard shortcut for loading data in most other applications, and Slicer already support this shortcut.

Ctrl-A was kept for backward compatibility, but it cause frustration to many users, as it prevented using this shortcut for selecting all items in lists and tables. For example, there was no keyboard shortcut for selecting all markup control points or segments in a list or all cells in a table.

This has been discussed at the last Slicer weekly meeting and everybody agreed on this change.


pieper (approved):
LGTM 👍

PR #5192, opened 2020-09-20
BUG: fix failing volume rendering test

pieper:
The test was failing because of an error message
from the logic class that the share directory
was not set:
```
ERROR: In D:\D\P\Slicer-0\Modules\Loadable\VolumeRendering\Logic\vtkSlicerVolumeRenderingLogic.cxx, line 1243
vtkSlicerVolumeRenderingLogic (00000222CD51C6D0): Failed to load presets: Share directory *NOT* set !
```
See: http://slicer.cdash.org/testDetails.php?test=10139125&build=2017840

In normal use in the application, this share directory is set
at the Qt level, but that path is not available to this vtk-only
test.

But the path is known to CMake, so here we pass in the path
as a command line argument to the test.


lassoan (approved):
Looks good to me, thank you.

Issue #4920, opened 2020-05-08
Standard Python package "lzma" is not available


lassoan:
Several Python packages, such as pandas or skimage complains that `lzma` package (that is supposed to be included in Python) is not available.

For example, error when trying to import skimage:

```
from skimage.morphology import skeletonize
```

```
ModuleNotFoundError                       Traceback (most recent call last)
In  [5]:
Line 3:     

File C:\Users\andra\AppData\Local\NA-MIC\Slicer 4.11.0-2020-05-06\lib\Python\Lib\site-packages\skimage\__init__.py, in <module>:
Line 133:   from .data import data_dir

File C:\Users\andra\AppData\Local\NA-MIC\Slicer 4.11.0-2020-05-06\lib\Python\Lib\site-packages\skimage\data\__init__.py, in <module>:
Line 61:    import pooch

File C:\Users\andra\AppData\Local\NA-MIC\Slicer 4.11.0-2020-05-06\lib\Python\Lib\site-packages\pooch\__init__.py, in <module>:
Line 7:     from .processors import Unzip, Untar, Decompress

File C:\Users\andra\AppData\Local\NA-MIC\Slicer 4.11.0-2020-05-06\lib\Python\Lib\site-packages\pooch\processors.py, in <module>:
Line 8:     import lzma

File C:\Users\andra\AppData\Local\NA-MIC\Slicer 4.11.0-2020-05-06\lib\Python\Lib\lzma.py, in <module>:
Line 27:    from _lzma import *

ModuleNotFoundError: No module named '_lzma'
```

jcfr:
We would need to first update python-cmake-buildsystem to support building lzma module

See https://github.com/python-cmake-buildsystem/python-cmake-buildsystem/blob/a1ce7c2cf346ce62648d8c3e9e5cae7f655d4309/cmake/extensions/CMakeLists.txt#L30-L31

pieper:
I've gotten this message before too, but it was never really an issue if you aren't using that format.

lassoan:
It would be nice to add wsgiref.simple_server, too. The lack of this package seems to prevents running Jupyter server in Slicer's Python environment. See https://github.com/Slicer/SlicerJupyter/issues/45.

PR #5194, opened 2020-09-22
BUG: Add lzma support to Python

lassoan:
A number of of Python packages complained that lzma package was not available (displayed a warning or refused to load), such as pandas, skimage, meshio.

Fixes https://github.com/Slicer/Slicer/issues/4920


lassoan:
I have only tested this on Windows.

It would be great if @pieper could test on Mac and @jcfr on Linux.

SuperBuild/External_lzma.cmake
```diff
@@ -0,0 +1,86 @@
...
+    set(${proj}_LIBRARY ${EP_INSTALL_DIR}/lib/lzma.a)
+  endif()
+
+else()
+  ExternalProject_Add_Empty(${proj} DEPENDS ${${proj}_DEPENDENCIES})
+endif()
+
+mark_as_superbuild(
+  VARS
+    ${proj}_DIR:PATH
```
jcfr, line 82: This is not needed as there is no Config file in the project

lassoan:
@jcfr I saw that you merged the https://github.com/python-cmake-buildsystem/python-cmake-buildsystem/pull/272 - thank you!

Are you going to finalize this PR (make the change related to lzma_DIR, update git branch/hash to official python-cmake-buildsystem, and squash all commits)?

jcfr:
Yes. I am in the process of doing so.

SuperBuild/External_LZMA.cmake
```diff
@@ -8,27 +8,32 @@ set(${proj}_DEPENDENCIES "")
 ExternalProject_Include_Dependencies(${proj} PROJECT_VAR proj DEPENDS_VAR ${proj}_DEPENDENCIES)
 
 if(Slicer_USE_SYSTEM_${proj})
-  unset(${proj}_DIR CACHE)
-  unset(${proj} CACHE)
-  find_package(${proj} REQUIRED)
+  unset(LZMA_INCLUDE_DIR CACHE)
```
lassoan, line 11: Is there a reason for using LZMA instead of ${proj}?

SuperBuild/External_LZMA.cmake
```diff
@@ -8,27 +8,32 @@ set(${proj}_DEPENDENCIES "")
...
   ExternalProject_SetIfNotDefined(
     Slicer_${proj}_GIT_REPOSITORY
-    "${EP_GIT_PROTOCOL}://github.com/nextgis-borsch/lib_lzma.git"
+    "${EP_GIT_PROTOCOL}://github.com/Slicer/lib_lzma.git"
     QUIET
     )
 
   ExternalProject_SetIfNotDefined(
     Slicer_${proj}_GIT_TAG
-    "v5.2.2"
```
lassoan: According to Python-3.6.7\PCbuild\readme.txt, lzma v5.2.2 is supposed to be used.

jcfr: Great. I will update the branch then

jcfr:
Commit message that will be used:

```
ENH: Add support for importing lzma python package

This commit updates python build system and adds LZMA external project
to allow the use of Python packages like pandas, skimage or meshio that would 
otherwise complain that lzma package is not available (displayed a warning or
refused to load).

Note that the official LZMA project do not include a CMake-based build-system. To streamline
the integration we forked a project adding such support from https://github.com/nextgis-borsch/lib_lzma
to https://github.com/Slicer/lib_lzma. Forking the project ensures Slicer can be built even if the upstream
project is renamed or removed.

LZMA v5.2.2 is used to satisfy CPython  requirements for windows build.
See https://github.com/python/cpython/blob/3.7/PCbuild/readme.txt

Fixes #4920

```

SuperBuild/External_LZMA.cmake
jcfr: * This allow easy grep for these variables in the code base 
* Consistent with the other external projects

SuperBuild/External_LZMA.cmake
```diff
@@ -0,0 +1,96 @@
...
+   OR NOT DEFINED LZMA_LIBRARY) AND NOT Slicer_USE_SYSTEM_${proj})
+
+  ExternalProject_SetIfNotDefined(
+    Slicer_${proj}_GIT_REPOSITORY
+    "${EP_GIT_PROTOCOL}://github.com/Slicer/lib_lzma.git"
+    QUIET
+    )
+
+  # Use v5.2.2 to satisfy CPython 3.7 Windows build requirements
+  # See https://github.com/python/cpython/blob/3.7/PCbuild/readme.txt
```
jamesobutler, line 35: https://github.com/python/cpython/blob/v3.6.7/PCbuild/readme.txt
I think you want to specify for 3.6.7 which is the version we currently use.  Technically 3.7 also uses v5.2.2 so technically the comment isn't wrong, but not the version we are using.

jcfr: Good catch :pray:

SuperBuild/External_LZMA.cmake
lassoan: It is important to list all the variables we set for grepping, but that could be done once, in the documentation of the script. Right now, ${proj} and project name seems to be randomly used throughout all External_*.cmake files (I could not deduce any logic of why one or the other is used).

I prefer using `${proj}` as much as possible, because it removes the need for copy/paste and search/replace when creating or updating External_*.cmake files and it makes them easier to compare (you only see the real differences). For example, try to compare External_sqlite.cmake (it uses ${proj} everywhere) and External_lzma.cmake files with/without using ${proj} everywhere.

jcfr: Sounds good, I will update then. Also, it should be `External_LZMA.cmake`  instead of `External_lzma.cmake`

lassoan: Thanks you, sounds good!

Just curious, how do you decide between lzma and LZMA?

jcfr: I used LZMA to be consistent with the the variable names  (e.g `LZMA_INCLUDE_DIR`)

Usually, it is also good to be consistent with the name of the CMake project (which would be `lzma`. See [here](https://github.com/nextgis-borsch/lib_lzma/blob/02e9a98e1771391f4f98df6153ff9b0c8188a52d/CMakeLists.txt#L29)).

But the official  project reference it as `LZMA` (see https://sourceforge.net/p/lzmautils/wiki/Home/) and there are occurrences of FindLZMA.cmake (for example [here](https://github.com/microsoft/vcpkg/blob/master/ports/vtk/FindLZMA.patch))

jcfr:
Instead of adding documentation like I did here:
https://github.com/Slicer/Slicer/pull/5194/commits/713c2ac73ba61c93e2275f32d18fbef726d10e03

I am thinking to instead to something like this:

```
# CMake variables supported by this external project:
#
# * LZMA_INCLUDE_DIR
# * LZMA_LIBRARY
# 
# To learn more about external project configuration, see https://slicer.readthedocs.io/en/latest/developer_guide/api.html#build-system
```

We would basically add a new "Build System" category to the SlicerAPI with a sub-section called external project where we would explain how the external project generally work.

Cc: @lassoan

PR #5197, opened 2020-09-22
STYLE: fix typo in Erase effect help text

rat01:
`Erase fom` > `Erase from`


lassoan (approved):
Good catch, thank you for the fix!

PR #5199, opened 2020-09-23
ENH: Enable markup preview when holding shift key

Sunderlandkyl:
Allows markups (and other widgets) to process events when the crosshair position is being changed with shift + mouse move.
This is done by processing ActionSetCrosshairPosition in CanProcessInteractionEvent, rather than ProcessInteractionEvent.


lassoan (commented):
Thank you, this is good. I've also tested and it works very intuitively. I added a few comments to explain the rationale behind our choices.

lassoan:
I've just noticed one thing that we should change: Shift+MouseMove in place mode should work the same way in 3D views, too.

Sunderlandkyl:
Thanks! I've updated the comments and added the same behavior to vtkMRMLCameraWidget.

lassoan (approved):
Looks good to me, thank you!

PR #5226, opened 2020-10-01
ENH: Update Create Fiducial Markup icon

smrolfe:
This commit updates the 'Create Fiducial Markup' icon in the Markups module. The single fiducial is replaced by a set and the arrow is removed. The goal is to help users distinguish between the function of this button and the place fiducial point button.

This update was suggested by @lassoan in the forum discussion:
https://discourse.slicer.org/t/surface-semilandmarks-created-inadequately/13704/4?u=smrolfe


pieper (approved):
LGTM 👍 

Anyone else want to comment before we merge?

jamesobutler:
It needs an appropriate git commit prefix as it has failed that check.

jamesobutler (commented):
I'm guessing this image is trying to mimic the Markups module icon image?

smrolfe:
@jamesobutler Correct, we can change the number/representation of points if you have a suggestion.

jamesobutler:
I think 3 fiducial points in the triangle orientation is enough compared to the 6 currently shown. The icon for MarkupsAngle, MarkupsCurve and MarkupsClosedCurve just use 3 points. It is interesting that those use blue points in the icon while the fiducial one is a red point. I can’t tell if it is, but I would reuse the same red color as the single fiducial placement icon for consistency between that markup type.

smrolfe:
Thanks, I'll make this change and resubmit.

PR #5222, opened 2020-10-01
Markups curve measurements

cpinter:
This PR includes 
* Infrastructure in measurements to accommodate markups control point measurements
* Curve measurements class that calculates curvature of curve markups nodes
  * Also can interpolate control point measurements (such as radius) and show it in the poly data. This feature will be tested through SlicerVMTK, coming soon
* Ability for markups to show scalars

The PR has the original commits so that you can see the incremental changes if you want. Please let me know if I should squash now, or otherwise if there is anything to fix/change.
Note: I will rebase and force push soon.

PR #5235, opened 2020-10-05
ENH: Markups curve measurements

cpinter:
- Infrastructure in measurements to accommodate markups control point measurements
- Curve measurements class that calculates curvature of curve markups nodes
  - Also can interpolate control point measurements (such as radius) and show it in the poly data. This feature will be tested through SlicerVMTK, coming soon
- Ability for markups to show scalars

Squashed commits:

STYLE: Fix copy-paste error in segment editor node doc

ENH: Add test for markups curvature computation

BUG: Fix scalar display without user selection

The user had to switch the scalar array away and back in order to be able to show the scalars from the first array.

ENH: Allow switching active scalar using assign attributes

BUG: Fix issue that changed selection when modifying the curve

- When the curve was modified, then the ctkVTKDataSetModel was rebuilt, and during that the current index was changed in the combobox, which triggered SetActiveScalars in the qMRMLScalarsDisplayWidget. Instead of observing the currentArrayChanged signal we observe the activated signal, which is only triggered when the user changes the selection
- This is also a better way to fix crash fixed in the previous commit by blockSignals

BUG: Fix crash caused by circular calls when modifying curve

When modifying the curve that displays scalars (e.g. move control point), there was a crash because of circular calls:

ctkVTKDataSetModel::onDataSetModified -> populateDataSet -> ctkVTKDataSetArrayComboBox::onCurrentIndexChanged -> qMRMLScalarsDisplayWidget::setActiveScalarName -> vtkMRMLDisplayNode::SetActiveScalar -> vtkMRMLMarkupsDisplayNode::UpdateAssignedAttribute -> vtkMRMLDisplayNode::UpdateScalarRange -> vtkMRMLMarkupsNode::GetCurveWorld (this->CurvePolyToWorldTransformer->Update() triggers pipeline update) -> ctkVTKDataSetModel::onDataSetModified

Fixed by blocking signals in qMRMLScalarsDisplayWidget::setActiveScalarName for the time of the vtkMRMLDisplayNode::SetActiveScalar call

ENH: Highlight of scalar colored curves with specular lighting

ENH: Active curve highlighting when scalar display is on

- Highlighting the curve on hover when scalar visibility is on by setting the specular lighting to maximum if scalar visibility is on, and back to 0 if off
- Fix scalar range update bug by implementing UpdateAssignedAttribute function in markups display node
- Fix crash

ENH: Add scalar display in 3D curve representation

ENH: Use ScalarsDisplayWidget from Markups module

- Make qMRMLMarkupsDisplayNodeWidget subclass QMRMLWidget not QWidget
- Add new Scalars section in the Display section of the Markups module with the ScalarsDisplayWidget
- Sink display node functions to generic display node with empty body to allow Models-specific features without casting to model display node

ENH: New Scalars display widget from Models module Scalars section

In order to use the scalar coloring features for other data types, it was necessary to create a new reusable widget for the related settings. The widgets from the Scalars section of the Models module were outsourced to a new Libs/MRML/Widgets/qMRMLScalarsDisplayWidget class. It seems to work perfectly within the Models module.

ENH: Add control point data interpolation

ENH: Add PedigreeIDs to curve; Use self-calculating measurement for lines

ENH: Create length measurement class that calculates internally

- Subclass measurement class to vtkMRMLMeasurementLength that calculates curve measurement itself
  - Curve markup populates the measurement collection on construction and calls Execute on enabled measurements when updating
- Markups node on ReadXMLAttributes and UpdateMeasurementsInternal does not remove all measurements only initializes them all
- Make GetUnitNode utility function public

ENH: Add members to vtkMRMLMeasurement to store more data

- Add control point values, enabled flag, and node reference to measurements
- Make its list vtkCollection in node

ENH: Add curve measurement calculator for markups curves !WIP!


cpinter:
This is the rebased and squashed PR for the original one https://github.com/Slicer/Slicer/pull/5222.

cpinter:
@lassoan This is ready to merge, please approve if you agree with the content. The only difference between the former PR and this one is squashing and resolving the merge conflicts. Thanks!

Libs/MRML/Widgets/qMRMLScalarsDisplayWidget.cxx
```diff
@@ -0,0 +1,611 @@
...
+    case 4:
+      this->setScalarRangeMode(qMRMLScalarsDisplayWidget::DirectMapping);
+      break;
+    default:
+      break;
+    }
+}
+
+// --------------------------------------------------------------------------
+qMRMLScalarsDisplayWidget::ControlMode qMRMLScalarsDisplayWidget::scalarRangeMode() const
```
lassoan, line 336: It would be nicer to have a single map that converts from combobox-item-index to range-mode and use that everywhere.
It is also quite confusing that `setScalarRangeMode` takes a combobox-item-index but `scalarRangeMode` returns a range-mode enum.
Maybe everything is correct, but at least things should be made more clear with more explicitly named arguments and variables.
Moreover, users of this class would probably expect to be able to use vtkMRMLDisplayNode::UseXYZ enums directly instead of thinking about converting to some internal enums.

Libs/MRML/Widgets/qMRMLScalarsDisplayWidget.h
```diff
@@ -0,0 +1,124 @@
...
+  /// Get the (first) current display node
+  vtkMRMLDisplayNode* mrmlDisplayNode()const;
+  /// Get current display nodes (if multi selection)
+  QList<vtkMRMLDisplayNode*> mrmlDisplayNodes()const;
+
+  bool scalarsVisibility()const;
+  QString activeScalarName()const;
+  vtkMRMLColorNode* scalarsColorNode()const;
+
+  enum ControlMode
```
lassoan, line 56: I think we should remove this and just use the modes defined in vtkMRMLDisplayNode.

Modules/Loadable/Markups/MRML/vtkCurveGenerator.cxx
```diff
@@ -628,12 +641,21 @@ int vtkCurveGenerator::GeneratePointsFromFunction(vtkPoints* inputPoints, vtkPoi
     previousPoint[0] = curvePoint[0];
     previousPoint[1] = curvePoint[1];
     previousPoint[2] = curvePoint[2];
+
+    // Calculate pedigree ID for point
```
lassoan, line 645: Please describe in the comment how it is calculated? It should be clear that pedigreeID is not the ID of the geometrically closest control point.

Modules/Loadable/Markups/MRML/vtkCurveMeasurementsCalculator.cxx
```diff
@@ -0,0 +1,352 @@
...
+    }
+  if (polyData->GetNumberOfPoints() == 0 || polyData->GetNumberOfLines() == 0)
+    {
+    // vtkErrorMacro("CalculatePolyDataCurvature: No points or lines in input poly data");
+    return false;
+    }
+
+  // Note: This algorithm has been ported from CurveMaker
+  //       https://github.com/tokjun/CurveMaker/blob/master/CurveMaker/CurveMaker.py
+  // TODO: This routine does not consider a closed loop. If a closed loop is specified,
```
lassoan, line 162: Could you add closed curve computation, too? We would just need to wrap around at both ends (to compute curvature at one end, you would need to use point(s) from the other end).

Modules/Loadable/Markups/MRML/vtkCurveMeasurementsCalculator.cxx
```diff
@@ -0,0 +1,352 @@
...
+
+  currentLength = sqrt( (prevPoint[0]-prevMeanPoint[0])*(prevPoint[0]-prevMeanPoint[0])
+                      + (prevPoint[1]-prevMeanPoint[1])*(prevPoint[1]-prevMeanPoint[1])
+                      + (prevPoint[2]-prevMeanPoint[2])*(prevPoint[2]-prevMeanPoint[2]) );
+  length += currentLength;
+  meanKappa = meanKappa / length;
+
+  // Set curvature array to output
+  polyData->GetPointData()->AddArray(curvatureValues);
+  polyData->GetPointData()->SetActiveScalars("Curvature"); //TODO: Do it here or in displayable manager when turning off curvature visibility?
```
lassoan, line 261: Multiple measurements may produce arrays, so probably we should not set the active scalars here.

Modules/Loadable/Markups/Testing/Python/MarkupsCurveMeasurementsTest.py
```diff
@@ -0,0 +1,58 @@
...
+if curvePointData.GetArrayName(1) != 'Curvature':
+  exceptionMessage = "Unexpected data array name in curve: " + str(curvePointData.GetArrayName(1))
+  raise Exception(exceptionMessage)
+
+curvatureArray = curvePointData.GetArray(1)
+if curvatureArray.GetMaxId() != curvePointData.GetNumberOfTuples()-1:
+  exceptionMessage = "Unexpected number of values in curvature data array: " + str(curvatureArray.GetMaxId())
+  raise Exception(exceptionMessage)
+
+if curvatureArray.GetMaxNorm() - 0.9816015970208652 > 0.0001:
```
lassoan, line 50:
```suggestion
if abs(curvatureArray.GetMaxNorm() - 0.9816015970208652) > 0.0001:
```

lassoan (commented):
This is great, thank you! I only have a handful of small comments - see inline.

cpinter:
Thanks @lassoan! I'll do these asap and update the PR.

Libs/MRML/Core/vtkMRMLMeasurement.cxx
```diff
@@ -53,12 +55,28 @@ void vtkMRMLMeasurement::Initialize()
   this->SetDerivationCode(nullptr);
   this->SetUnitsCode(nullptr);
   this->SetMethodCode(nullptr);
+  this->SetControlPointValues(nullptr);
+  //this->SetPolyDataValues(nullptr);
```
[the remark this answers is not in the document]
cpinter: Do you really mean #undef? I don't know how it could be used here. I'll use #define and then #ifdef's unless you tell me otherwise.

Libs/MRML/Core/vtkMRMLMeasurement.h
```diff
@@ -37,9 +39,20 @@ class VTK_MRML_EXPORT vtkMRMLMeasurement : public vtkObject
...
   virtual void Initialize();
 
+  /// Clear value and derived variables
+  virtual void ClearValue();
+
   /// Copy one type into another (deep copy)
   virtual void Copy(vtkMRMLMeasurement* aEntry);
 
+  /// Perform calculation on \sa InputMRMLNode and store the result internally
+  virtual void Compute() {};
```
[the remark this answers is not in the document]
cpinter: I haven't updated all the measurements yet. At least the one in vtkMRMLMarkupsAngleNode::UpdateMeasurementsInternal() is the way it was.

Libs/MRML/Widgets/qMRMLScalarsDisplayWidget.cxx
```diff
@@ -0,0 +1,611 @@
...
+
+  this->setupUi(q);
+
+  // Without this infinite loop of widget update/VTK data set update occurs when updating a VTK data set
+  // that is generated by an algorithm that  temporarily removes all arrays from its output temporarily
+  // (for example the vtkGlyph3D filter behaves like this).
+  // The root cause of the problem is that if none option is not enabled then the combobox
+  // automatically selects the first array, which triggers a data set change, which removes all arrays,
+  // which triggers a widget update, etc. - until stack overflow occurs.
+  //this->ActiveScalarComboBox->setNoneEnabled(true);
```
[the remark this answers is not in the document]
cpinter: OK. Someone has kept it there for a warning I guess and I didn't remove it in case it was useful, but I agree that it doesn't seem particulatly useful now.

Libs/MRML/Widgets/qMRMLScalarsDisplayWidget.cxx
```diff
@@ -0,0 +1,611 @@
...
+  if (d->CurrentDisplayNodes.size() > 0)
+    {
+    return d->CurrentDisplayNodes[0];
+    }
+
+  return nullptr;
+}
+
+//------------------------------------------------------------------------------
+QList<vtkMRMLDisplayNode*> qMRMLScalarsDisplayWidget::mrmlDisplayNodes()const
```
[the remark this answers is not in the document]
cpinter: Yes it is, for supporting the SH folders in the Models module

Libs/MRML/Widgets/qMRMLScalarsDisplayWidget.cxx
cpinter: Makes sense. I did wo things:
- Remove the ControlMode (which had a strange name anyway, and it contained the same options as vtkMRMLDisplayNode::ScalarRangeFlagType but in a slightly different order: Manual is with index 3 in ControlMode but with 0 in the ScalarRangeFlagType), and just use vtkMRMLDisplayNode::ScalarRangeFlagType instead
- Reorder the items in the combobox to correspond to the enum values in vtkMRMLDisplayNode::ScalarRangeFlagType, and made the item 1 the default (the same one: Data scalar range (auto), which I renamed from Active scalar (auto))

lassoan:
> Do you really mean #undef? I don't know how it could be used here. I'll use #define and then #ifdef's unless you tell me otherwise.

I thought using undef because this is how CMake converts `#cmakedefine` to C code, for example:

```
/* #undef Slicer_VTK_RENDERING_USE_OpenGL_BACKEND */
#define Slicer_VTK_RENDERING_USE_OpenGL2_BACKEND
#define Slicer_VTK_USE_QVTKOPENGLWIDGET
```

But it is true that we don't use this now in hand-written Slicer code and if it is not intuitive then it is not good. So, we can just use `#define` (we already use `//#define` in two files to exclude performance profiling code).

> +  /// Perform calculation on \sa InputMRMLNode and store the result internally
+  virtual void Compute() {};
I haven't updated all the measurements yet. At least the one in vtkMRMLMarkupsAngleNode::UpdateMeasurementsInternal() is the way it was.

If that's not too much trouble, it would be nice if you could implement that. This whole measurement mechanism is not trivially simple and would be better not to complicated with legacy mechanism lying around.

> +QList<vtkMRMLDisplayNode*> qMRMLScalarsDisplayWidget::mrmlDisplayNodes()const
Yes it is, for supporting the SH folders in the Models module

I see, it is useful then. Please add a comment that "it is for supporting changing properties in case of multiple nodes are selected (e.g., in subject hierarchy)". Why there aren't any other widgets that do this (at least I couldn't find any other examples).

lassoan (approved):

PR #5236, opened 2020-10-06
BUG: Set segment name from color table when the labelmap has one label

Sunderlandkyl:


lassoan (approved):
Looks good to me, thank you.

PR #5237, opened 2020-10-07
ENH: Improve subject hierarchy performance

lassoan:
GetItemByDataNode was slow because it had to search the whole tree for an item that has a specific MRML node associated with it.

Made it faster by adding a vtkMRMLNode* to vtkSubjectHierarchyItem* cache.

### DIFF
--- a/Libs/MRML/Core/vtkMRMLSubjectHierarchyNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLSubjectHierarchyNode.cxx
@@ -103,9 +103,10 @@ public:
   /// The ID is resolved to pointer after import ends, and this member is set to INVALID_ITEM_ID.
   vtkIdType TemporaryParentItemID;
 
-  /// Item cache to speed up lookup by ID that needs to be performed many times.
+  /// Item and data node cache to speed up lookups that are needed many times.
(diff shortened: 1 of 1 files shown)


lassoan:
@cpinter can you have a look?

Libs/MRML/Core/vtkMRMLSubjectHierarchyNode.cxx
```diff
@@ -709,30 +715,36 @@ vtkSubjectHierarchyItem* vtkSubjectHierarchyItem::FindChildByID(vtkIdType itemID
     {
     return itemIt->second;
     }
-  else
-    {
-    vtkWarningMacro("FindChildByID: Item cache does not contain requested ID " << itemID);
-    }
+  // It is not an error if item is not found in cache. It happens normally when
+  // scene has just been closed and widgets are updating themselves and trying to look up their selected item.
```
cpinter, line 719: Great, hopefully this will get rid of all those annoying error messages.

cpinter (approved):
Sorry, I did have a look yesterday and it seemed good. Thanks a lot!

lassoan:
Thanks for the review.

PR #5243, opened 2020-10-12
BUG: Remove P shortcut from incorrect location

smrolfe:
This commit removes a 'P' shortcut implemented in the wrong place. It will be reimplemented but the current definition is interfering with customized shortcuts.

From @lassoan:

> I’ve found a ‘P’ key shortcut in qSlicerMarkupsModuleWidget. This is implemented at completely wrong place, so it should be removed.
> Since “P” shortcut would be generally useful, I think it should be handled in vtkSlicerMarkupsWidget, the same way how “Delete”, “Backpace” and “S” keys are handled already. We also plan to make widget GUI to widget event mapping (‘P’ key press -> pick point) customizable by users and modules.”

### DIFF
--- a/Modules/Loadable/Markups/qSlicerMarkupsModuleWidget.cxx
+++ b/Modules/Loadable/Markups/qSlicerMarkupsModuleWidget.cxx
@@ -682,8 +682,6 @@ qSlicerMarkupsModuleWidget::qSlicerMarkupsModuleWidget(QWidget* _parent)
   : Superclass( _parent )
     , d_ptr( new qSlicerMarkupsModuleWidgetPrivate(*this) )
 {
-  this->pToAddShortcut = nullptr;
-
   this->volumeSpacingScaleFactor = 10.0;
 }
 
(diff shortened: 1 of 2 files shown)


jamesobutler (commented):
Would you be able to restore the shortcut in the correct place in this PR as well?

smrolfe:
Thanks @jamesobutler, I'm happy to coordinate with @lassoan and reimplement following his suggestions as soon as we have time. I'd prefer just to remove for now if possible, since it's interfering with our extension's custom shortcuts. As it is, this shortcut is not functional, so removing should not cause any problems.

smrolfe:
Looping in @pieper who worked with us on the SlicerMorph shortcuts.

lassoan:
It should be very easy to implement this properly in vtkSlicerMarkupsWidget:
- Add a new `WidgetEventControlPointAdd` event enum value (similar to `WidgetControlPointInsert`)
- Add default mapping of "P" keypress to a new "WidgetEventControlPointAdd" event in the constructor
- Implement ProcessControlPointAdd method (similar to ProcessControlPointInsert, but actually implement adding the control point at the current position instead of delegating it to derived classes)

pieper (approved):
LGTM 👍

smrolfe:
Thanks @pieper!

jamesobutler:
Could someone create a git issue to remember to add the "P" shortcut back into Slicer core?

lassoan:
> Could someone create a git issue to remember to add the "P" shortcut back into Slicer core?

Done - https://github.com/Slicer/Slicer/issues/5247

PR #5246, opened 2020-10-13
BUG: Fix crash when selecting a lookup table for slice distance color

### LINKED ISSUE #5241: Crash when selecting some color node for model distance encoded projection display
Slicer crashes if certain color node is chosen for model distance-encoded projection.

## Steps to reproduce

- Load this scene
- Go to models module
- Select Segment_1
- in Display/Slice display/Color table -> choose CoolTint2 => ERROR: the application crashes

## Environment
- Slicer version: Slicer-4.11.2020-09-30

Sunderlandkyl:
The pointer to the lookup table was not set when using a lookup table color node.
Fixed by setting the lut pointer to the correct object.

fixes #5241


lassoan (approved):
perfect, thank you!

Issue #4806, opened 2020-04-04
Extension manager is slow


lassoan:
From @rkikinis:

Operating system: Mac OS 10.15.4
Slicer version: Slicer-4.11.0-2020-04-01-macosx-amd64
Actual behavior:
Something is broken/bizarre with the extension manager. I need to click twice for the window to appear and the history tab is empty. After I install some extensions, i do not get a prompt to restart. When I close the extension manager and exit slicer, then the extension manager window pops up on its own and prompts for a restart.

See related discussion here: https://discourse.slicer.org/t/problematic-behavior-of-the-extension-manager/10986

lassoan:
> I need to click twice for the window to appear

Maybe you just need to wait more. If your application is [not signed](https://github.com/Slicer/Slicer/issues/4724) then it may take minutes for the integrated web browser to launch. Could you try clicking only once and wait for several minutes?

> the history tab is empty

I could not reproduce this issue on Windows.

> After I install some extensions, i do not get a prompt to restart.

You may need to wait more. After the installation seems to be completed, it may take several minutes the restart button to become active.

> When I close the extension manager and exit slicer, then the extension manager window pops up on its own and prompts for a restart.

This may be the second instance of the extension manager (or just the single instance did not quit).

I find this behavior quite frustrating, too. Most likely all the problems are in the javascript and javascript/C++ communication layer (with sometimes extra slowdowns caused by the midas3 server), so probably @pieper would be the best person to debug.

pieper:
I think this has nothing to do with javascript  : )

All the issues described by @rkikinis sound like symptoms of a slow response from the midas server.  As part of the discussion of the thread linked below I saw delays of over 8 minutes at some points and other times near instant response, which would explain why Ron's issues are not reproducible consistently.  This indicates that [this transition project](https://www.slicer.org/wiki/Documentation/Labs/ExtensionsServer) needs to get some attention (probably funding would help move it along...).

https://discourse.slicer.org/t/download-slicer-org-and-extension-catalogue-seems-slows/10835/11?u=pieper

lassoan:
I suspect that the delays are in local javascript execution because for example after an extension download and installation is completed, you may need to wait several minutes for the "Restart" button to become active. The server should not be able to slow down anything other than the download.

Of course delays may still be side effects of slow server-side processing: there may be hundreds of background tasks, such as query list of extensions and download images, and the browser may delay other tasks until some background tasks are completed.

pieper:
[...]

Issue #4635, opened 2020-03-13
Intermittent crash on macOS  when opening the Extensions Manager


slicerbot:
_This issue was created automatically from an original [Mantis Issue](https://mantisarchive.slicer.org/view.php?id=4635). Further discussion may take place here._

lassoan:
Probably fixed by https://github.com/Slicer/Slicer/pull/5249/commits/f0a7a9f9f4e6b2ea2a026b10f6b1b3899bf1fe97

PR #5249, opened 2020-10-14
ENH: Make extension manager more user-friendly

lassoan:
- Display a message popup (and wait cursors) while extension manager is starting up to let the users know that they need to wait (helps with making slow startup more tolerable - see #4806)
- Prevent creating multiple extension manager windows (while the extension manager was starting up, it was possible to create additional instances)
- Display confirmation popup if the user wants to close/restart the extension manager while extensions are still being downloaded or installed
- Make search box work in all tabs
- Make "Install extension from file" more discoverable by having a dedicated button
- Make "Settings" button have the same options, regardless of which tab is selected (it was confusing for users that some options were missing sometimes)
- Fix uninitialized extension manager model (which might have caused crash described in #4635)

For the future: Extension metadata is queried from the extension server quite extensively. This slows down extension manager startup and filtering and increases load on the server. Performance could be improved by implementing more local caching.


cpinter:
These are very useful for the current version. Does this mean that we are not expecting a new extension manager in the near future (for 5.0)?

lassoan:
The extension manager server replacement may slightly change how the appstore webpage looks like (content of "Install extensions" tab) and the server API. None of these are related to the application-level improvements implemented in this pull request.

Base/QTGUI/qSlicerApplication.cxx
```diff
@@ -733,6 +738,28 @@ void qSlicerApplication::setHasBorderInFullScreen(bool hasBorder)
...
+
+  // Startup of extension manager can take tens of seconds.
+  // Display a popup to let the user know.
+  QApplication::setOverrideCursor(QCursor(Qt::BusyCursor));
+  QDialog* startupInProgressDialog = new QDialog(this->mainWindow(), Qt::FramelessWindowHint | Qt::Dialog);
+  QVBoxLayout* layout = new QVBoxLayout();
+  startupInProgressDialog->setLayout(layout);
+  layout->setMargin(20);
+  QLabel* label = new QLabel();
+  label->setText(tr("Extension manager is starting, please wait..."));
```
jcfr, line 758: for consistency
`Extension manager` ->  `Extensions manager`

Base/QTGUI/qSlicerExtensionsRestoreWidget.cxx
```diff
@@ -412,6 +430,12 @@ void qSlicerExtensionsRestoreWidgetPrivate
 ::downloadAndInstallNextExtension()
 {
   Q_Q(qSlicerExtensionsRestoreWidget);
+  if (!q->extensionsManagerModel())
+    {
+    // extension manager model is not set yet
+    qWarning() << Q_FUNC_INFO << " failed: extension manager model is invalid";
```
jcfr, line 436: `extension manager` -> `extensions manager`

Base/QTCore/qSlicerExtensionsManagerModel.h
```diff
@@ -340,6 +340,9 @@ public slots:
 
   void updateModel();
 
+  /// Number of operations in progress
+  int activeTaskCount() const;
```
jcfr, line 344: Move this outside of the public slots and instead add a property:

```
Q_PROPERTY(int activeTaskCount READ activeTaskCount)
```

To be consistent, I am wondering if this should be renamed to `numberOfActiveTasks`. I have no strong feeling about it.

jcfr (commented):
Thanks for working on this, looks good (I haven't tried to compile it)

 May also be worth checking if code like this still  work:

https://github.com/Slicer/Slicer/blob/403902710ce34505795044485881efdf915722a1/Modules/Scripted/DMRIInstall/DMRIInstall.py#L111-L127

Base/QTCore/qSlicerExtensionsManagerModel.h
lassoan: ...Count is Qt style, numberOf... is VTK style. We don't use them very consistently. Since in this class the numberOf... is already used, I'll switch to that.

jcfr: > Count is Qt style, numberOf... is VTK style. We don't use them very consistently.

Good point. I added a note to https://www.slicer.org/wiki/Documentation/Nightly/Developers/Style_Guide#Naming

lassoan: To not break our own rules then, I'll make the other numberOf... methods/properties in this class deprecated and use ...Count for all. There are a couple of (not too many) other At classes that use numberOf, but for now I'll leave them alone.

PR #5252, opened 2020-10-15
DOC: Update operating system versions for recent Slicer stable

jamesobutler:
The recent slicer stable is referring to Slicer-4.11.20200930

### DIFF
--- a/Docs/user_guide/getting_started.md
+++ b/Docs/user_guide/getting_started.md
@@ -8,10 +8,9 @@ Welcome to the 3D Slicer community. This page contains information that you need
 
 ### Operating system versions
 
-|               |Windows                                        |macOS                             |Linux                                     |
-|---------------|-----------------------------------------------|----------------------------------|------------------------------------------|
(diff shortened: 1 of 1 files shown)


lassoan:
Thank you, good catch!

Readthedocs is already versioned (I've just enabled building of stable tag):
https://slicer.readthedocs.io/en/stable/
https://slicer.readthedocs.io/en/latest/

Since the doc always refers to the "current" version (that users can choose in the readthedocs version selector), I think we can remove reference to stable/preview version and only write about what is applicable to the current version (remove the table and just list OS requirements for the current master version).

Could you update the pull request accordingly? Thank you!

jamesobutler:
I've pushed an update to remove the Stable/Preview specification.  

I guess this means I'm unable to update the current stable documentation to fix the operating system version requirements because it has already been released and there is no maintenance branch for building the latest stable documentation.

lassoan:
Good point. Although we don't plan to release not patches for 4.11, we can still create a branch from the 20200930 version and update the documentation. I've created a v4.11 for this purpose and enabled documentation generation for it: https://slicer.readthedocs.io/en/v4.11/ (and disabled documentation generation for the `stable` tag to keep things more clear).

You can send a pull request to the v4.11 branch.

lassoan (approved):
thank you, this looks good

PR #5253, opened 2020-10-15
DOC: Update operating system versions to reflect state of Slicer-4.11

jamesobutler:
This is a follow-up to #5252 for the v4.11 branch.

Documentation is versioned so there is not a need for both Stable and Preview information.


lassoan (approved):
perfect, thank you

PR #5254, opened 2020-10-15
BUG: Fix crash in qMRMLScalarsDisplayWidget

Sunderlandkyl:
qMRMLScalarsDisplayWidget::setMRMLDisplayNode() will cause a crash when called with nullptr.
Fixed by checking if the display node exists before accessing it.

### DIFF
--- a/Libs/MRML/Widgets/qMRMLScalarsDisplayWidget.cxx
+++ b/Libs/MRML/Widgets/qMRMLScalarsDisplayWidget.cxx
@@ -150,15 +150,23 @@ void qMRMLScalarsDisplayWidget::setMRMLDisplayNodes(QList<vtkMRMLDisplayNode*> d
 {
   Q_D(qMRMLScalarsDisplayWidget);
 
-  // Set first node as current node (the properties of which that the widget displays)
-  if (displayNodes.size() > 0)
-    {
-    this->setMRMLDisplayNode(displayNodes[0]);
-    }
+  displayNodes.removeAll(nullptr);
+
+  // Only the first display node is observed
+  qvtkReconnect(
+    (d->CurrentDisplayNodes.size() > 0 ? d->CurrentDisplayNodes[0] : nullptr),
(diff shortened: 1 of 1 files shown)


Libs/MRML/Widgets/qMRMLScalarsDisplayWidget.cxx
```diff
@@ -179,9 +179,11 @@ void qMRMLScalarsDisplayWidget::setMRMLDisplayNode(vtkMRMLDisplayNode* displayNo
     displayNode, vtkCommand::ModifiedEvent, this, SLOT(updateWidgetFromMRML()) );
 
   d->CurrentDisplayNodes.clear();
-  d->CurrentDisplayNodes << displayNode;
-
-  displayNode->UpdateScalarRange();
+  if (displayNode)
+    {
+    d->CurrentDisplayNodes << displayNode;
```
lassoan, line 184: If displayNode is nullptr then we should clear d->CurrentDisplayNodes.

lassoan (commented):
Thank you for the fix!

The code that manages a single display node and a list of display nodes is a bit messy. That's probably why Csaba missed the error that you found and why your fix needs a little further fixing.

Could you clean things up by moving content of `setMRMLDisplayNode(vtkMRMLDisplayNode*)` into `setMRMLDisplayNodes(QList<vtkMRMLDisplayNode*>)`? That would mean that all display node setter methods would end up calling the same method (`setMRMLDisplayNodes(QList<vtkMRMLDisplayNode*>`). Now `setMRMLDisplayNodes(QList<vtkMRMLDisplayNode*>)` calls out to `setMRMLDisplayNode(vtkMRMLDisplayNode*)`, which internally quietly removes other display nodes, but then `setMRMLDisplayNodes(QList<vtkMRMLDisplayNode*>)` sets the other display nodes again.

Sunderlandkyl:
Sounds good, the code is now consolidated in setMRMLDisplayNodes.

lassoan (approved):
Perfect, thank you!

PR #5256, opened 2020-10-20
ENH: Add support for closed curves to curvature calculation

cpinter:
- Since for closed curves the two ends meet, instead of setting the end curvature values to zero, we use the adjacent values
- Force measurements to implement Compute function (and thus using the new method for measurement calculation) by making it pure virtual
- Add test for closed curve curvature calculation

### DIFF
--- a/Libs/MRML/Core/vtkMRMLMeasurement.cxx
+++ b/Libs/MRML/Core/vtkMRMLMeasurement.cxx
@@ -15,8 +15,6 @@ or http://www.slicer.org/copyright/copyright.txt for details.
 // STD include
 #include <sstream>
 
-vtkStandardNewMacro(vtkMRMLMeasurement);
-
 //----------------------------------------------------------------------------
 vtkMRMLMeasurement::vtkMRMLMeasurement()
   : InputMRMLNode(nullptr)
--- a/Libs/MRML/Core/vtkMRMLMeasurementConstant.cxx
+++ b/Libs/MRML/Core/vtkMRMLMeasurementConstant.cxx
@@ -15,26 +15,29 @@
 
 ==============================================================================*/
 
-#include "vtkMRMLMarkupsClosedCurveNode.h"
-
 // MRML includes
-#include "vtkCurveGenerator.h"
-#include "vtkCurveMeasurementsCalculator.h"
+#include "vtkMRMLMeasurementConstant.h"
 
-// VTK includes
-#include <vtkNew.h>
-#include <vtkObjectFactory.h>
+vtkStandardNewMacro(vtkMRMLMeasurementConstant);
 
 //----------------------------------------------------------------------------
-vtkMRMLNodeNewMacro(vtkMRMLMarkupsClosedCurveNode);
-
-//----------------------------------------------------------------------------
-vtkMRMLMarkupsClosedCurveNode::vtkMRMLMarkupsClosedCurveNode()
+vtkMRMLMeasurementConstant::vtkMRMLMeasurementConstant()
 {
-  this->CurveClosed = true;
-  this->CurveGenerator->SetCurveIsClosed(true);
-  this->CurveMeasurementsCalculator->SetCurveIsClosed(true);
 }
 
 //----------------------------------------------------------------------------
-vtkMRMLMarkupsClosedCurveNode::~vtkMRMLMarkupsClosedCurveNode() = default;
+vtkMRMLMeasurementConstant::~vtkMRMLMeasurementConstant() = default;
+
+//----------------------------------------------------------------------------
+void vtkMRMLMeasurementConstant::PrintSelf(ostream& os, vtkIndent indent)
+{
+  Superclass::PrintSelf(os,indent);
+}
+
+//---------------------------------------------------------------------------
+void vtkMRMLMeasurementConstant::Compute()
+{
+  // No action is needed because this class is a constant variant of the dynamic
+  // vtkMRMLMeasurement. Typically all measurements calculate their own value from
+  // its input MRML node. This class is to be able to store constant measurements.
+}
--- a/Modules/Loadable/Markups/MRML/vtkCurveMeasurementsCalculator.h
+++ b/Modules/Loadable/Markups/MRML/vtkCurveMeasurementsCalculator.h
@@ -47,6 +47,14 @@ public:
   //@}
 
   //@{
+  /// This indicates whether the curve loops back in on itself,
+  /// connecting the last point back to the first point (disabled by default).
+  vtkSetMacro(CurveIsClosed, bool);
+  vtkGetMacro(CurveIsClosed, bool);
+  vtkBooleanMacro(CurveIsClosed, bool);
+  //@}
+
+  //@{
   /// Set/Get flag determining whether to calculate curvature
   vtkSetMacro(CalculateCurvature, bool);
   vtkGetMacro(CalculateCurvature, bool);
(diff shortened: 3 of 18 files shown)


Modules/Loadable/Markups/MRML/vtkCurveMeasurementsCalculator.h
```diff
@@ -46,6 +46,14 @@ class VTK_SLICER_MARKUPS_MODULE_MRML_EXPORT vtkCurveMeasurementsCalculator : pub
   vtkCollection* GetMeasurements();
   //@}
 
+  //@{
+  /// This indicates whether the curve should loop back in on itself,
+  /// connecting the last point back to the first point (disabled by default).
+  vtkSetMacro(CurveIsLoop, bool);
+  vtkGetMacro(CurveIsLoop, bool);
```
lassoan, line 53: We use "loop" and "closed" terms quite randomly. It would be better to use only one , consistently. Since closed is used in 6 headers files (many of them are public and would be really hard to change), while loop is used in only 2 headers and only vtkCurveGenerator uses it publicly, I think we should stick to ClosedCurve everywhere (and we can change/deprecate the few existing usages of "loop").

Modules/Loadable/Markups/Testing/Python/MarkupsCurveMeasurementsTest.py
```diff
@@ -44,15 +44,57 @@
...
   exceptionMessage = "Unexpected number of data arrays in curve: " + str(curvePointData.GetNumberOfArrays())
   raise Exception(exceptionMessage)
+
+print('Open curve curvature test finished successfully')
+
+#
+# Test closed curve curvature computation
+#
+closedCurveNode = slicer.mrmlScene.AddNewNodeByClass('vtkMRMLMarkupsClosedCurveNode')
+pos = [0]*3
```
lassoan, line 69: operator* is unsafe for creating vectors, as items may be duplicated (instead of created as independent instances)

```
>>> matrix=[np.zeros(3)]*3
>>> matrix
[array([0., 0., 0.]), array([0., 0., 0.]), array([0., 0., 0.])]
>>> matrix[2][1]=5
>>> matrix
[array([0., 5., 0.]), array([0., 5., 0.]), array([0., 5., 0.])]
```

I know that in this case it is a simple list and elements are independent instances, but it is better not to get used to the habit of using `*3`. Instead, use np.zeros(3) to create a vector.

Modules/Loadable/Markups/Testing/Python/MarkupsCurveMeasurementsTest.py
```diff
@@ -44,15 +44,57 @@
...
+for i in range(curveNode.GetNumberOfControlPoints()):
+  curveNode.GetNthControlPointPosition(i, pos)
+  posVector = vtk.vtkVector3d()
+  posVector.Set(pos[0], pos[1], pos[2])
+  closedCurveNode.AddControlPoint(posVector)
```
lassoan, line 74: you can pass the vector in the vtkVector3d constructor: `closedCurveNode.AddControlPoint(vtk.vtkVector3d(pos))`

lassoan (commented):
Looks good, thank you, just added a few trivial comments.

Modules/Loadable/Markups/MRML/vtkCurveMeasurementsCalculator.h
cpinter: OK. I was just trying to follow conventions. I can make the terminology more consistent by getting rid of "loop"

lassoan: Yes, the naming was already inconsistent in different classes (mainly due to bringing in curve maker filter from another module), just wanted to make sure that we are converging towards a consistent naming scheme.

cpinter:
@lassoan I updated the pull request as you suggested and as we discussed. In addition to that I decided to make the Compute function in the measurement base class pure virtual to avoid its deprecated usage, and thus had to create the vtkMRMLMeasurementConstant subclass for that type of usage (used in two tests). Also worth mentioning that I didn't touch the vtkSlicerMarkupsWidgetRepresentation::ClosedLoop variable name, because it has "closed" in its name already and didn't find an obvious better choice. Let me know if it's fine by you.

lassoan:
> @lassoan I updated the pull request as you suggested and as we discussed.

I don't see any updates. Have you pushed it?

> In addition to that I decided to make the Compute function in the measurement base class pure virtual to avoid its deprecated usage, and thus had to create the vtkMRMLMeasurementConstant subclass for that type of usage (used in two tests).

Good idea. I can see how vtkMRMLMeasurementConstant can be a useful class for assigning properties to a markups node (not just in tests). Although we can use node attributes for storing extra metadata for a markups node, these attributes cannot be displayed easily on the GUI, not saved to the json file, etc.

> Also worth mentioning that I didn't touch the vtkSlicerMarkupsWidgetRepresentation::ClosedLoop variable name, because it has "closed" in its name already and didn't find an obvious better choice. Let me know if it's fine by you.

vtkSlicerMarkupsWidgetRepresentation::ClosedLoop is called CurveClosed in vtkMRMLMarkupsNode. It would be more clear to use CurveClosed everywhere.

cpinter:
It seems that I did push my changes but in another branch. I renamed ClosedLoop, amended the last commit, then cherry picked those commits here. Now this should be complete.

Libs/MRML/Core/vtkMRMLMeasurement.cxx
```diff
@@ -337,3 +335,9 @@ vtkMRMLNode* vtkMRMLMeasurement::GetInputMRMLNode()
 
   return nullptr;
 }
+
+//----------------------------------------------------------------------------
+void vtkMRMLMeasurement::Compute()
```
lassoan, line 340: Does this compile? you defined `virtual void Compute() = 0;` in the header.

Modules/Loadable/Markups/Testing/Python/MarkupsCurveMeasurementsTest.py
```diff
@@ -44,15 +44,132 @@
...
   raise Exception(exceptionMessage)
+
+print('Open curve curvature test finished successfully')
+
+#
+# Test closed curve curvature computation
+#
+
+closedCurveNode = slicer.mrmlScene.AddNewNodeByClass('vtkMRMLMarkupsClosedCurveNode')
+pos = [0]*3
```
lassoan, line 70:
```suggestion
pos = np.zeros(3)
```

Libs/MRML/Core/vtkMRMLMeasurementConstant.cxx
```diff
@@ -0,0 +1,40 @@
...
+
+//----------------------------------------------------------------------------
+void vtkMRMLMeasurementConstant::PrintSelf(ostream& os, vtkIndent indent)
+{
+  Superclass::PrintSelf(os,indent);
+}
+
+//---------------------------------------------------------------------------
+void vtkMRMLMeasurementConstant::Compute()
+{
```
lassoan, line 39: Add a comment here why no action is needed.

lassoan (commented):
It looks good, I just I don't understand how pure virtual function with a body can work. Can you check?
There are also two trivial comments.

Libs/MRML/Core/vtkMRMLMeasurement.cxx
cpinter: An oversight. I'll remove the body. Sorry!

lassoan (approved):
Thank you!

PR #5258, opened 2020-10-22
ENH: Make logs of previous sessions more easily accessible

lassoan:
"Report a bug" dialog had a combobox for selecting logs from previous sessions but users had trouble finding it.
This commit replaces the single-line combobox with a list widget, which is easier to discover and it also easier to browse through the files.

image

### DIFF
--- a/Base/QTApp/Resources/UI/qSlicerErrorReportDialog.ui
+++ b/Base/QTApp/Resources/UI/qSlicerErrorReportDialog.ui
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>702</width>
-    <height>441</height>
+    <width>607</width>
+    <height>638</height>
    </rect>
   </property>
   <property name="sizePolicy">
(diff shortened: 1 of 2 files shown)


jamesobutler (commented):
Yes I think it is a common behavior to have to find the correct log which contains the details of something that recently went wrong. This doesn’t always mean the last file if the user attempted a restart of Slicer a few times. Therefore having a list instead of a combo box helps out when flipping through logs because it is easier to not get lost with what log you’ve already checked.

I think there could be a few additional minor improvements as well on this front related to improving the understanding that there are multiple logs to look through.
- Users frequently don’t understand that the currently selected log upon opening this dialog represents the “current session” log and it is dynamically changing after doing other things in the Slicer session. Often times they incorrectly assume this first one was the last Slicer instance log. Maybe an additional text or indicator of it being the “current session” would help.
- The displayed paths are mainly the same between each other except for the file base name. This makes it difficult to discern differences between entries in the list. Maybe we could display only the file base name or some simpler formatted name that just shows the Slicer version and time with the full path as the hover text for the entry. The full path is not needed as much since there is the button to open the log file directly. Side note: I’ve personally added a button in my application to open the log file directory which allows for easily copying of all log files using the file explorer. I find this more valuable than the open log file because I’d rather copy the file(and other files) itself instead of opening it to copy the content in the file.
- When a user is toggling through the various logs they are often trying to find where something went wrong and this is often at the end of the log. Maybe upon opening the dialog and also upon switching currently selected log to view, the log file content area can auto scroll to the bottom. With this change the log content will noticeably change upon a new selection which doesn’t happen currently because the first 20 lines or so are the standard startup log lines and you only see the time stamp of those first 20 or so lines changing.

lassoan:
Thanks a lot for the review.

> Users frequently don’t understand that the currently selected log upon opening this dialog represents the “current session” log and it is dynamically changing after doing other things in the Slicer session. Often times they incorrectly assume this first one was the last Slicer instance log. Maybe an additional text or indicator of it being the “current session” would help.

Good point, it would be nice to implement this, but the currently used ctkPathListWidget does not support this (it needs the full paths as inputs and displays the full paths).

> The displayed paths are mainly the same between each other except for the file base name. This makes it difficult to discern differences between entries in the list. Maybe we could display only the file base name or some simpler formatted name that just shows the Slicer version and time with the full path as the hover text for the entry. The full path is not needed as much since there is the button to open the log file directly.

Agree, this would be nice, too. The currently used ctkPathListWidget does not support this. We could use a simple list widget, but I like the appearance of the current widget (with the file icons). It would be probably easy to add the icons in a regular listbox, too.

> Side note: I’ve personally added a button in my application to open the log file directory which allows for easily copying of all log files using the file explorer. I find this more valuable than the open log file because I’d rather copy the file(and other files) itself instead of opening it to copy the content in the file.

I agree that this is simpler, but more risky. I would prefer if users reviewed the logs, remove all sensitive information, and share this edited text, instead of blindly sending/uploading the entire log somewhere.

> When a user is toggling through the various logs they are often trying to find where something went wrong and this is often at the end of the log. Maybe upon opening the dialog and also upon switching currently selected log to view, the log file content area can auto scroll to the bottom. With this change the log content will noticeably change upon a new selection which doesn’t happen currently because the first 20 lines or so are the standard startup log lines and you only see the time stamp of those first 20 or so lines changing.

Both the top and bottom has its own advantages. The first error is the most important error (the rest may be just consequences) and it is easier to find it if the log is read from the top. Also, top of the log contains the most important information, so it is good that it can be seen easily, without any scrolling. If the users sanitize the logs (review the entire content) then starting from the top is more natural. I agree with the usefulness of starting at the bottom, too, but it is not clearly better than starting at the top. Maybe a quick parsing and highlighting of errors/warnings could help finding the problems easier.

---

Overall, I don't see anything that would be clear or very easy to improve (not hard either, but switching to a different widget, using custom data for storing the full path, highlighting the current log with custom formatting, add icon, etc. could take 1-2 hours). If you agree that these changes improve things then I would merge this. If you are not sure if this is for the better then I would abandon this pull request and maybe revisit in the future.

cpinter:
Thanks @lassoan, this is useful. I also agree with the comments of @jamesobutler. I'd just like to add, in case someone will do these requests later, that once we add postfixes (current session), we could also add the size of the log file like (92KB). Many times when I look for the relevant log, I find it by size, as most of them are too small to contain what I look for.
My vote is to merge this as is if these additions are unreasonable to do without major changes.

pieper:
I agree this is a tricky one to solve in general.  One thing that people no doubt have trouble with is the time encoding in the log name.  Even the date format may confuse some people.  How about a tool tip that summarizes each log, like "35 minute session started Thu Oct 22, 8:01 am (25 errors, 104 warnings 354 info)".  This could help people recognize the session where the issue happened.

lassoan:
I agree with all the proposed improvement ideas but I don't think I can work on this more right now. In the past few days I've been using this new layout and it feels it works much better than the old one, so I'll merge this as is. We can improve things in future commits.

PR #5260, opened 2020-10-22
ENH: Update qSlicerWebWidget to allow export to PDF

jcfr:
Since WebEngine objects are not exposed in PythonQt, this commit updates
the existing qSlicerWebWidget to support using existing PDF export
functionality built-in WebEngine widget.

For example:

```
  url = "file:///home/jcfr/Downloads/index.html"
  outputPath = "/tmp/test.pdf"

  webWidget = slicer.qSlicerWebWidget()

  def loadFinished(ok):
    if not ok:
      print("failed to load URL %s" % url)
      return
    webWidget.printToPdf(outputPath);

  def pdfPrintingFinished(filePath, success):
    print("export completed [%s]: %s" % (success, filePath))

  webWidget.connect("loadFinished(bool)", loadFinished)
  webWidget.connect("pdfPrintingFinished(QString,bool)", pdfPrintingFinished)
  webWidget.setUrl(url)
```

### DIFF
--- a/Base/QTGUI/qSlicerWebWidget.cxx
+++ b/Base/QTGUI/qSlicerWebWidget.cxx
@@ -145,6 +145,9 @@ void qSlicerWebWidgetPrivate::init()
   QObject::connect(this->WebView, SIGNAL(loadProgress(int)),
                    this->ProgressBar, SLOT(setValue(int)));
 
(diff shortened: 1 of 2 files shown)


pieper (approved):
LGTM 👍

PR #5263, opened 2020-10-24
Add ArchLinux installation information

pieper:
Based on information here:

https://discourse.slicer.org/t/could-not-load-dicom-data/14211/3


lassoan:
I'm just curious, is it normal that stripping breaks the binary or it exposes some hidden error? (to be honest, I don't know what stripping does)

Issue #5259, opened 2020-10-22
Segment Editor: Add warning (overridable) when user hits "Apply" on not-visible segment


hherhold:
I will sometimes inadvertently hit “apply” on an editing operation with a segment that is not visible. This results in a rather long delay, followed by “undo”, to get back to where I was.

It would be nice to have a warning dialog that lets you know the selected segment is not visible, and do you wish to proceed. As noted by @lassoan, this should be a Yes/No dialog with a "Do not show again", as this behavior (being able to modify segments that are not visible) is desired in certain situations.

hherhold:
@lassoan - You mentioned "It would be a helper function in C++ that could be called using a single line from both C++ and Python scripted effects."

Is the idea to add a new dialog in Base/QTApp, similar to qSlicerAboutDialog, qSlicerErrorReportDialog, etc, or something less heavyweight? It's likely to only be called from Segment Editor effects, so it probably makes more sense to implement it somewhere in Modules/Loadable/Segmentations/EditorEffects?

Thanks!

lassoan:
> Is the idea to add a new dialog in Base/QTApp, similar to qSlicerAboutDialog, qSlicerErrorReportDialog, etc, or something less heavyweight?

It is specific to segmentations and it is a widget, so it should be implemented in Modules\Loadable\Segmentations\Widgets. Since confirmation popup dialog with "do not show again" checkbox is already implemented in CTK, we don't need a new class, just a helper function in qMRMLSegmentEditorWidget. Name could be `bool confirmCurrentSegmentVisible()`. This method would show a popup if the current segment is not visible and return with true if the current segment is visible or the user confirmed he wants to operate on a hidden segment. It could be implemented similarly to the scene close confirmation popup (https://github.com/Slicer/Slicer/blob/master/Base/QTApp/qSlicerMainWindow.cxx#L727-L753). The method would be called in "Apply" methods of segment editor effects.

I would also consider having 3 buttons: "Currently selected segment is hidden. Would you like to make it visible? Yes, No, Cancel"

hherhold:
Got it. Thanks!

hherhold:
Hi Andras,

I have everything finished and working. I'm trying to merge HEAD back into my fork before my pull request, but the pre-commit git hook is failing on Docs/user_guide/getting_started.md with lines longer than 80 chars.

Should I fix this file and add it to my pull request?

Thanks!

jcfr:
[...]

PR #5274, opened 2020-10-28
ENH: Segment Editor: Add warning (overridable) when user hits "Apply" on not-visible segment #5259

hherhold:
Added qSlicerSegmentEditorAbstractEffect::confirmCurrentSegmentVisible() method to check for segment visibility on Segment Editor operations with "Yes", "No", or "Cancel". Dialog is only shown if "Apply" is clicked when a segment is invisible. On "Yes", the operation is performed and the segment is made visible, on "No", the operation is performed and the segment remains invisible, and on "Cancel", the operation is NOT performed and the segment remains invisible.

This is currently only implemented for Islands operations but is generally useable by any Segment Editor affect with an Apply button.

This dialog is currently always shown, but a "Don't ask again" could be added.


lassoan:
Sorry for the slow response, I've been preparing for a demo and could not check this. I'll work on it today.

hherhold:
No rush. Thanks for looking at it, and let me know if it looks ok or if I need to change anything.

Thanks!!!

lassoan:
Thank you very much, added the check to all effects, enabled "don't show again" option in popup, added application settings to adjust "don't show again" behavior, and merged as https://github.com/Slicer/Slicer/commit/8744261ce329983ae354e6f5663a3a435f49c890.

PR #5278, opened 2020-10-29
DOC: Add doxygen documentation link to developer guide

lassoan:


jcfr:
Look like there is an issue integrating the index page:

image

lassoan:
Yes, thanks, I know, I use readthedocs to test if I got the index right (it is not correct yet). I'm trying to create an index separately for the user and developer manual.

PR #5292, opened 2020-11-04
BUG: use exec not execfile for scripted MRMLDM

pieper:
vtkMRMLScriptedDisplayableManager::SetPythonSource evaluates
a python code string to load a the source file.

The code used execfile, which worked in python2 but does not
exist in python3.

This changes the code to use exec, open, and read in order
to emulate the execfile behavior.

### DIFF
--- a/Libs/MRML/DisplayableManager/vtkMRMLScriptedDisplayableManager.cxx
+++ b/Libs/MRML/DisplayableManager/vtkMRMLScriptedDisplayableManager.cxx
@@ -297,7 +297,7 @@ void vtkMRMLScriptedDisplayableManager::SetPythonSource(const std::string& pytho
     PyObject * pyRes = nullptr;
     if (pythonSource.find(".py") != std::string::npos)
       {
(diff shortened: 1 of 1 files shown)


lassoan:
In most SetPythonSource methods, we now use `d->PythonCppAPI.instantiateClass` instead of exec - see https://github.com/Slicer/Slicer/blob/master/Base/QTCore/qSlicerScriptedFileWriter.cxx#L149

Have you tried if instantiateClass works? It would be better to use the same mechanisms everywhere, unless it does not work for some reason.

lassoan:
@pieper do you have an example of a working scripted displayable manager? I would try if using instantiateClass works and if yes then I would replace the custom instantiation with that call.

pieper:
No, I started looking at doing an experiment but ran into this bug and didn't get back to it.  Would be glad to see this migrate to be consistent with the mechanism used elsewhere, I just didn't find the time.

lassoan:
I had another look and it is actually good. `d->PythonCppAPI.instantiateClass` is used in Qt classes.

PR #5293, opened 2020-11-04
CRASH: fix observer manager events and priorities

pieper:
The method gives the option of providing a list of priorities
with one priority per event in two lists.

The check logic was incorrect however, because in the case
where no priorities had been provided (null list pointer) the
'if' condition was triggered and the warning macro would
indirect through the null priorities pointer.

Now the two checks have been separated and each has a custom
warning message so that pointers are only used when they are
known to be non-null.


lassoan (commented):
Good catch, thank you. The change looks good except it seems that "CRASH:" is not a valid commit comment prefix (the commit test rejected it).

pieper:
Thanks - I thought we used to use CRASH, but BUG works well enough so I force pushed.

lassoan (approved):

PR #5294, opened 2020-11-05
ENH: Add markups curvature mean and max measurements

cpinter:
* Curvature mean and max measurements are added to curve markup. These two are disabled by default and get enabled when the user enables curvature calculation for the markup itself.
* They are constant measurements (because these values are calculated in vtkCurveMeasurementsCalculator and it would not make much sense recalculating them in each measurement) and set by vtkCurveMeasurementsCalculator.
* They show up in the description as "curvature mean" and "curvature max"
* Their unit is "kappa" given that this was the only option I could find. We can remove the unit name if this seems incorrect.

### DIFF
--- a/Modules/Loadable/Markups/MRML/vtkCurveMeasurementsCalculator.cxx
+++ b/Modules/Loadable/Markups/MRML/vtkCurveMeasurementsCalculator.cxx
@@ -296,6 +296,25 @@ bool vtkCurveMeasurementsCalculator::CalculatePolyDataCurvature(vtkPolyData* pol
   length += currentLength;
   meanKappa = meanKappa / length;
 
+  // Set mean and max curvature to measurements
+  // Calculate and set interpolated control point measurements in poly data
+  for (int index=0; index<this->Measurements->GetNumberOfItems(); ++index)
+    {
+    vtkMRMLMeasurement* currentMeasurement = vtkMRMLMeasurement::SafeDownCast(this->Measurements->GetItemAsObject(index));
+    if (!currentMeasurement || !currentMeasurement->GetEnabled())
+      {
+      continue;
+      }
+    if (currentMeasurement->GetName() && !strcmp(currentMeasurement->GetName(), this->GetMeanCurvatureName()))
+      {
+      currentMeasurement->SetValue(meanKappa);
+      }
+    else if (currentMeasurement->GetName() && !strcmp(currentMeasurement->GetName(), this->GetMaxCurvatureName()))
+      {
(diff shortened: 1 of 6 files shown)


lassoan (commented):
Thank you , this looks good.

The computed curvature value (kappa) is 1.0/radius_of_curvature. Radius of curvature would be probably easier to interpret for users (as it is a radius of a circle), but it is not a well-behaved parametrization, as for a straight line the radius would be infinite (and "radius of curvature" is also somewhat counterintuitive in the sense that larger value means "less curved").

So, keep the "curvature" metric, just fix the unit to be 1/length. Use UCUM standard notation: `mm-1`. Since we don't have a units engine and we only have a unit node for length, we need to get the unit from length and add the `-1` manually.

I've verified that radius is in fact 1/kappa by the test below. It would be great if you could add it to MarkupsCurveMeasurementTests.py:

```python
#
# Test curvature computation for a circle-shaped closed curve
#

radius = 35
numberOfControlPoints = 40
import math
closedCurveNode=slicer.mrmlScene.AddNewNodeByClass("vtkMRMLMarkupsClosedCurveNode")
for controlPointIndex in range(numberOfControlPoints):
    angle = 2.0*math.pi * controlPointIndex/numberOfControlPoints
    closedCurveNode.AddControlPoint(vtk.vtkVector3d(radius * math.sin(angle), radius * math.cos(angle), 0.0))

# Turn on curvature calculation in curve node
closedCurveNode.SetCalculateCurvature(True)

curvePolyData = closedCurveNode.GetCurveWorld()
curvatureArray = curvePolyData.GetPointData().GetArray('Curvature')

if curvatureArray.GetNumberOfValues() < 10:
  exceptionMessage = "Many values are expected in the curvature array, instead  found just %d" % curvatureArray.GetNumberOfValues()
  raise Exception(exceptionMessage)

if abs(curvatureArray.GetRange()[0] - 1/radius) > 1e-4:
  exceptionMessage = "Unexpected minimum in curvature data array: " + str(curvatureArray.GetRange()[0])
  raise Exception(exceptionMessage)
if abs(curvatureArray.GetRange()[1] - 1/radius) > 1e-4:
  exceptionMessage = "Unexpected maximum in curvature data array: " + str(curvatureArray.GetRange()[1])
  raise Exception(exceptionMessage)

print('Radius of curvature computation is verified successfully')
```

cpinter:
Thanks @lassoan !

I made the change about unit, see in the constructor of vtkMRMLMarkupsCurveNode.

The test section is added and passes as well.

lassoan (approved):
Looks good, thank you!

PR #5301, opened 2020-11-17
ENH: Use GPU for volume rendering by default

lassoan:
CPU volume renderer is very slow and does not support custom shaders and other features.
CPU volume renderer was still kept as default for better hardware compatibility, but nowadays it is hard to find graphics systems nowadays that cannot render a volume using GPU volume renderer.

Also changed the default rendering quality to "Normal", as for GPUs the adaptive settings does not work very reliably (sometimes makes the rendering slower).

See related discussion at https://discourse.slicer.org/t/best-computer-for-running-slicer/14579


cpinter:
Couldn't agree more. Thank you!

cpinter (approved):

pieper (approved):
Yes, makes good sense 👍

jcfr:
Agreed. That change is very welcome.

jcfr:
> re: adaptive does not work well

I am wondering if the switch to vtk 9 will help.

Is this something we should bring up with the vtk devs ?

lassoan:
Adaptive mode relies on measuring rendering time and adjusting quality based on how much time the rendering took.

This feedback loop works for CPU processing because we can measure the rendering time accurately and we know how the rendering time is expected to change as we change rendering parameters.

However, when rendering on the GPU, we cannot measure the rendering time accurately (lots of processing done in background threads in the CPU (in the graphics driver) and in the GPU, which interfere with subsequent rendering times, there are also various optimization heuristics done in the driver, which affects rendering time), and it is hard to predict how the rendering time will be affected by changing rendering parameters because we are not aware of everything that happens in the GPU driver and hardware.

It could be maybe a feature request to the VTK team to improve the rendering quality/speed control loop by not just using the last rendering time but by storing dozens of previous rendering times obtained with different rendering settings and estimate best parameters from these. However, it would not be a very high priority request, because rendering times on GPU are often so good that you don't need adaptive rendering at all.

lassoan:
Thanks for the reviews, merged.

PR #5311, opened 2020-11-20
BUG: Remove sampledata not available for commercial use

jcfr:
This commit removes "CTA abdomen (Panoramix)" dataset

See https://discourse.slicer.org/t/should-we-include-download-of-sampledata-not-available-for-commercial-use/14717


lassoan:
Maybe instead of removing it, we could replace it with a similar image (to cause less disruption, e.g., for tutorials that use this image). @pieper do you have an idea where we could get an angio CT from?

pieper:
I'm not sure they are quite comparable, but it looks like we should be able to use one of these datasets: https://www.ncbi.nlm.nih.gov/pmc/articles/PMC6122162/

We can also use TCIA images - I'll keep an eye out for something similar.  Not sure about the contrast/quality.

lassoan:
I've checked the internet archive and the sample data sets were publicly distributed without restrictions in 2011 May.

image

image

It was kind of an advertisement - "These images are best viewed with OsiriX image navigation software !", and distribution and usage were not _restricted_. I don't think these terms can be changed retrospectively. For creative work, one could argue that lack of copyright notice does not mean free use, but this is not creative work, but a patient consented to make parts of their medical record public.

pieper:
I'm sure you are right @lassoan, in general it's better to prefer data with clear license, in this case I'd say we should avoid the breaking change unless someone complains.

lassoan:
I've sent a message via osirix-viewer website about this data set:

>I would like to learn more about patient consent related to images distributed at https://www.osirix-viewer.com/resources/dicom-image-library/.
>
>Could you please provide patient consent documentation for the PANORAMIX data set, to see the exact wording of the terms that the patient has agreed to (that is broadly described as "You are not authorized to redistribute or sell them, or use them for commercial purposes" on the OsiriX website)?

Their answer will be interesting not just because of this particular data set but in general to see how a company responds when asked about a sample data set. Probably we should move towards replacing all sample data sets with public data sets managed by others (for example, TCIA or other data sets released as part of research studies) and save link to the original repository of the data.

lassoan:
> I'm sure you are right @lassoan, in general it's better to prefer data with clear license, in this case I'd say we should avoid the breaking change unless someone complains.

I agree. Let's close this issue, no need to spend more time with this unless somebody complains.

PR #5313, opened 2020-11-23
BUG: Fix performance issue when copying vtkMRMLStreamingVolumeNode

Sunderlandkyl:
Copying a vtkMRMLStreamingVolumeNode would cause the frame to be decoded, leading to to poor performance when copying entire sequences.
Fixed by preventing decoding during node copy.

### DIFF
--- a/Libs/MRML/Core/vtkMRMLStreamingVolumeNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLStreamingVolumeNode.cxx
@@ -26,6 +26,7 @@ Care Ontario.
 #include <vtkAlgorithmOutput.h>
 #include <vtkCallbackCommand.h>
 #include <vtkCommand.h>
+#include <vtkMatrix4x4.h>
 
 // vtkAddon includes
 #include <vtkStreamingVolumeCodecFactory.h>
@@ -279,7 +280,6 @@ bool vtkMRMLStreamingVolumeNode::EncodeImageData(bool forceKeyFrame/*=false*/)
     return false;
     }
 
-
   vtkSmartPointer<vtkStreamingVolumeFrame> frame = vtkSmartPointer<vtkStreamingVolumeFrame>::New();
   if (!this->Codec->EncodeImageData(imageData, frame, forceKeyFrame))
     {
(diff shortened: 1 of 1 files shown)


jcfr:
The change makes sense, I am wondering if we should take this opportunity to add the `CopyInProgress` ivar to the base class and update the function `CopyContent`

@lassoan What do you think ?

lassoan:
> The change makes sense, I am wondering if we should take this opportunity to add the CopyInProgress ivar to the base class and update the function CopyContent

I think this is a special case and it should not be necessary in general. Have you seen similar flags in other classes?

Sunderlandkyl:
Are there any changes needed, or can this be integrated?

lassoan:
Calling the Superclass::CopyContent with some special flags make things a bit opaque, especially because vtkMRMLVolumeNode::CopyContent is already quite complicated (it copies image data, emits image modified events, etc.).

It would be simpler, more explicit if we did not call Superclass::CopyContent in vtkMRMLStreamingVolumeNode::CopyContent but call vtkMRMLDisplayableNode::CopyContent instead and then implement all necessary volume copying operations there. It would result in some code duplication (vtkMRMLTensorVolumeNode::CopyContent and parts of vtkMRMLVolumeNode::CopyContent), but it is just a couple of lines, and so it should not be a problem.

Sunderlandkyl:
@lassoan Ok, I've duplicated some of the CopyContent code from vtkMRMLTensorVolumeNode and vtkMRMLVolumeNode in a way that avoid decoding the frame of the source node.

Libs/MRML/Core/vtkMRMLStreamingVolumeNode.cxx
```diff
@@ -340,16 +340,45 @@ void vtkMRMLStreamingVolumeNode::ReadXMLAttributes(const char** atts)
...
+      producer ? producer->GetOutputDataObject(
+        this->ImageDataConnection->GetIndex()) : nullptr);
+
+    vtkSmartPointer<vtkImageData> targetImageData = sourceImageData;
+    if (deepCopy && sourceImageData)
+      {
+      targetImageData = vtkSmartPointer<vtkImageData>::Take(sourceImageData->NewInstance());
+      targetImageData->DeepCopy(sourceImageData);
+      }
+      this->SetAndObserveImageData(targetImageData);
```
lassoan, line 368: indentation

Libs/MRML/Core/vtkMRMLStreamingVolumeNode.cxx
```diff
@@ -340,16 +340,45 @@ void vtkMRMLStreamingVolumeNode::ReadXMLAttributes(const char** atts)
...
+  if (!streamingVolumeNode)
     {
-    return;
+    // Other node is not a streaming volume node.
+    // We don't need to worry about it trying to decode the frame.
+    Superclass::CopyContent(anode, deepCopy);
+    }
+  else
+    {
+    vtkMRMLDisplayableNode::CopyContent(anode, deepCopy);
```
lassoan, line 353: Describe here why we don't call SuperClass::CopyContent.

lassoan (commented):
Thank you, this looks good, clean, just address the two trivial comments.

Sunderlandkyl:
Thanks, changes made.

lassoan (approved):
Thank you, looks good.

PR #5318, opened 2020-11-25
STYLE: Simplify external projects re-using EP_BINARY_DIR variable

jcfr:


pieper (approved):
Looks like a good change 👍

PR #5319, opened 2020-11-25
COMP: Fix "command line too long error" using ninja generator on windows

jcfr:
This commit is an attempt to avoid error reported when using Ninja generator
on Windows.

Co-authored-by: Stephen Aylward <Stephen.Aylward@kitware.com>

See also https://discourse.slicer.org/t/build-failure-using-visual-studio-2015/7351

### DIFF
--- a/Applications/SlicerApp/Testing/Python/JRC2013Vis.py
+++ b/Applications/SlicerApp/Testing/Python/JRC2013Vis.py
@@ -102,13 +102,13 @@ class JRC2013VisWidget(ScriptedLoadableModuleWidget):
       import subprocess
       dcmqrscpExeOptions = (
         '/bin',
-        '/../CTK-build/CMakeExternals/Install/bin',
+        '/../CTK-b/CMakeExternals/Install/bin',
         '/../DCMTK-install/bin',
-        '/../DCMTK-build/bin',
-        '/../DCMTK-build/bin/Release'
-        '/../DCMTK-build/bin/Debug'
-        '/../DCMTK-build/bin/RelWithDebInfo'
-        '/../DCMTK-build/bin/MinSizeRel'
+        '/../DCMTK-b/bin',
+        '/../DCMTK-b/bin/Release'
+        '/../DCMTK-b/bin/Debug'
+        '/../DCMTK-b/bin/RelWithDebInfo'
+        '/../DCMTK-b/bin/MinSizeRel'
         )
 
       dcmqrscpExePath = None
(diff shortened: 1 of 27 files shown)


pieper (commented):
This seems practical, but a shame we need to take these little steps.

Does the `-install` not cause problems?  Maybe it should be changed to `-i`.

jcfr:
>  Does the -install not cause problems? 

Since header files are all located in a single directory, the command line doesn't go beyond the limit when using an install tree. Problem is when invoking the compiler with a long list of include directories.

We could also install project like VTK ... but:
* as developers we often have to tweak VTK code, rebuild and re-run Slicer ... it would make that process even more tedious
* also building external VTK module against a VTK build tree works only for VTK8.


> Maybe it should be changed to -i.

For sake of consistency, that would make sens. I will update the PR

lassoan:
I don't like that readability compromised because of a short-term workaround. For example, how would I know if "i" refers to install, include, interface,...?

If ninja generator is important then switch to passing inputs as parameter files. If we go on like this then when VTK or ITK adds a couple of more modules then we may get into the impossible situation that we need to use folder names that have 0 or negative length.

jcfr:
Overall, I agree with you.

> ninja generator is important then switch to passing inputs as parameter files

I am still trying to get confirmation if ninja was used or not (I just assumed) along with exacts version of CMake, Visual Studio, .... 

In the meantime, I am marking this PR as a draft

lassoan:
CMake always uses parameter files when passing input to Visual C++ compiler and linker. There are a very few rare exceptions, such as maybe the resource compiler and custom tools in CTK that might not accept parameter files, but we should address those issues instead of further restricting path lengths.

Installing VTK does not seem like a bad solution either.
- Modifying VTK and re-run Slicer is not all that common (maybe a few percent of my time is spent on re-running Slicer with modified VTK) and I measured VTK9 installation time on Windows takes 24 seconds - which should be tolerable.
- "building external VTK module against a VTK build tree works only for VTK8" - it should be enough if it works on VTK9. In the long term we would like to be able to build from an "SDK" anyway.

jcfr:
Next week, I will work on reproducing the issue so that we either submit an issue tp upstream CMake or fix the relevant build system.

For reference, the following configuration was used:

> I was generating Ninja and using VS2017 native x64 window.
> Ninja version 1.8.2
> CMake version 3.18.2
> Windows 10

lassoan:
We have a lot of open pull requests that makes it harder to see if there are active pull requests that require attention. What do you think about closing this pull request, maybe adding an issue that summarizes the issue and refers to the branch for a potential solution?

jcfr:
Since documentation section [Developer Guide » Build Instructions » Windows » Set up source and build folders](https://slicer.readthedocs.io/en/5.0/developer_guide/build_instructions/windows.html#set-up-source-and-build-folders) provides instructions allowing to easily avoid the problem, we will revisit this in the future.

PR #5320, opened 2020-11-25
BUG: Do not skip adding hidden and excluded nodes to subject hierarchy

cpinter:
The purpose of hidden and excluded nodes is not to show them in the UI. Since subject hierarchy is supposed to have an item to every supported node in the scene, and the hidden/excluded attributes can change, it does not make sense to not add them to subject hierarchy when adding the node or loading a scene containing the node.

### DIFF
--- a/Modules/Loadable/SubjectHierarchy/Widgets/qSlicerSubjectHierarchyPluginLogic.cxx
+++ b/Modules/Loadable/SubjectHierarchy/Widgets/qSlicerSubjectHierarchyPluginLogic.cxx
@@ -222,11 +222,10 @@ void qSlicerSubjectHierarchyPluginLogic::onNodeAdded(vtkObject* sceneObject, vtk
 
     // Add subject hierarchy node for the added data node
     // Don't add to subject hierarchy automatically one-by-one if importing scene, because the SH nodes may be stored in the scene and loaded
-    // Also abort if invalid or hidden node or if explicitly excluded from subject hierarchy before even adding to the scene
-    if ( scene->IsImporting()
-      || !node
-      || node->GetHideFromEditors()
-      || node->GetAttribute(vtkMRMLSubjectHierarchyConstants::GetSubjectHierarchyExcludeFromTreeAttributeName().c_str()) )
+    // Also abort if invalid or hidden node. The HideFromEditors flag is not considered to work dynamically, meaning that
+    // we don't expect changes on the UI when we set it after adding it to the scene, so not adding it to SH should not cause issues.
+    // The GetSubjectHierarchyExcludeFromTreeAttributeName attribute on the other hand is dynamic, so we add the node to SH despite that.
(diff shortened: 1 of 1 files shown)


lassoan:
Was there any specific problem with the current behavior?

I find it very useful that SH tree can show nodes, regardless of hidden or not, and use it extensively in many scripted modules, because vtkMRMLScriptedModuleNode is hidden by default but I can make it show up in the SH tree. I often use this module node as parent for all other nodes that this module creates.

The only things that is broken since the display hierarchy rework that I cannot show/hide all the nodes that are below the scripted module node because it is not a displayable node.

Maybe we can address both issues at once: If you find that showing hidden nodes in SH is a big problem then we could make vtkMRMLScriptedModuleNode a visible node and make it a displayable node.

cpinter:
The problem with the current behavior is that you load a scene in which there is a node that was excluded from SH using the attribute, then the node will not be added to SH at all. So:
- Scene that you save has the node in SH but as excluded, so not shown in the tree views
- Scene that you load will not have the node in SH at all, so even if you remove the excluded attribute, it will be outside of SH so will not be shown in the tree views

This change should not have any effect on how the exclude or HideFromEditor flags behave, only allow the nodes to be added to the SH node, and let the proxy model decide if they should be shown or not.

lassoan:
Thanks for the clarification, I guess I just missed an important "not"

> it does not make sense to **not** add them to subject hierarchy when adding the node or loading a scene containing the node

It is good then.

lassoan (approved):
Looks good to me.

cpinter:
Thanks Andras! Do you want me to update the commit message to be clearer? And include the original problem description?

lassoan:
I guess I was confused by the negated negatives ("do not skip adding", "does not make sense to not add them"). Something like this would be simpler, more direct:

```
BUG: Always add hidden and excluded nodes to subject hierarchy

The purpose of hidden and excluded nodes is hide them in the UI. However, hidden/excluded attributes can change, so we need to have these nodes in subject hierarchy so that these changes can be monitored. The proxy model can decide if they should be shown in the UI or not.
```

By the way, keeping all hidden nodes slowed down the regular node selector very significantly. Doesn't this slow down the SH tree? It triples the number of nodes that are monitored by SH - by keeping track of a display and storage node for each data node; and we also have extra 100+ color nodes. Would we have a SH item for each color, display, and storage node in the scene file?

cpinter:
> slow down the SH tree

This is a good point. Maybe I should re-add the HideFromEditors exception. That flag is already not considered work dynamically, meaning that we don't expect changes on the UI when we set it after adding it to the scene.

cpinter:
I re-added the hidden nodes and edited the commit message based on your suggestion and added a paragraph about hidden nodes. Thanks for the review!

lassoan (approved):
Looks good, thank you.

PR #5322, opened 2020-11-26
Markups measurement storage

cpinter:
Two commits: a bugfix and the storage feature.

### DIFF
--- a/Libs/MRML/Core/vtkMRMLMeasurement.cxx
+++ b/Libs/MRML/Core/vtkMRMLMeasurement.cxx
@@ -33,7 +33,11 @@ std::string vtkMRMLMeasurement::GetValueWithUnitsAsPrintableString()
 {
   if (!this->PrintFormat)
     {
-    return  "";
+    return "";
+    }
+  if (!this->ValueDefined)
+    {
+    return "(undefined)";
     }
   char buf[80] = { 0 };
   snprintf(buf, sizeof(buf) - 1, this->PrintFormat, this->Value, this->Units);
@@ -46,6 +50,7 @@ void vtkMRMLMeasurement::Initialize()
   this->SetEnabled(true);
   this->SetName(nullptr);
   this->SetValue(0.0);
+  this->ValueDefined = false;
   this->SetUnits(nullptr);
   this->SetPrintFormat("%5.3f %s");
   this->SetDescription(nullptr);
@@ -64,6 +69,7 @@ void vtkMRMLMeasurement::Initialize()
 void vtkMRMLMeasurement::ClearValue()
 {
   this->SetValue(0.0);
+  this->ValueDefined = false;
 
   // Note: this->SetControlPointValues(nullptr); is not called here, because if we clear
   // this then every time certain things in the markups node changes that calls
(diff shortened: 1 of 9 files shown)


cpinter:
Hold on just realized I still need to save the control point data.

lassoan (commented):
Thank you the saving/loading works, but I've found a few problems or unclear behavior with measurements:

1. Use standard JSON for storing per-control-point measurements for two reasons: 1. you don't want to force users to implement a custom parser when json is already perfectly capable of storing vectors or vector of vectors (we already do this for control point position storage). 2. Streaming << operator is just not good enough for writing double values. It is not does not guarantee that a double value survives writing out and reading in without any change. If you hardcode stream formatting to use many decimals then you may not lose information, but you may end up with lots of extra decimals in the file.

I've implemented this change, you can just cherry-pick this commit: https://github.com/lassoan/Slicer/commit/d05f3574d38f7b57395930aaab30746801bd2ee7

2. If I save a markups with radius array and then load it, then Radius is disabled. If I click Enabled then "Radius: Insufficient input" shows up in Measurements section. Why? How can I set (and save and reload) a constant measurement?

If I call c.GetNthMeasurement(3).SetValue(44.5) then the value shows up in description column but still "Radius: Insufficient input"  is displayed in Measurements section

3. While I'm moving a control point, "Active Scalar: PedigreeIDs" is displayed. When I release the mouse then it switches back to the correct selection (Curvature)

4. If Data scalar range = auto and I edit the curve (so that its maximum curvature gets higher or lower) the scalar range is not recomputed (I need to switch to manual and back to auto to make it refresh). It should refresh automatically.

5. Measurements section does not look nice:

image

You could put the enabled/disabled table in a collapsible groupbox. Make the table row height minimum necessary (currently it is very tall), make the table fill the available horizontal space, do not scroll the table (now I can only see 1.5 items by default).

6. Error at startup. When starting Slicer in debug mode, it logs this message (probably harmless, but pollutes the error log):

Algorithm vtkCurveGenerator(000002216BFC5F90) returned failure for request: vtkInformation (000002216BD43100)
  Debug: Off
  Modified Time: 190454
  Reference Count: 1
  Registered Events: (none)
  Request: REQUEST_DATA
  FORWARD_DIRECTION: 0
  ALGORITHM_AFTER_FORWARD: 1
  FROM_OUTPUT_PORT: 0

cpinter:
1. Thanks a lot @lassoan for doing the vector read/write change! I should have noticed that there is a way already to do this without custom implementation.

3. I have seen this too. The scalar array selection logic is quite complex and I decided not to bother trying to fix this as it would probably take several hours. Do you want me to work on this?

I'll work on the rest today.

cpinter:
@lassoan I fixed some of the issues but have a few questions:

-2: I changed the code so that it doesn't write "Insufficient input". But as it doesn't have a single value, it always writes 0.000. If there is a control point array and value equals 0 then should I write "Control point values (see scalar display)" instead?

-5: I tried a bunch of stuff but I can't get rid of the scrollbar without manually trying to calculate the actual height of the table (which is impossible now with the different DPIs we support). If I disable the scrollbar it simply cuts off the rest. I could change this section to not use a table widget, but I opted for a table widget so that I don't have to manually delete widgets when updating the section.

cpinter:
@lassoan I cherry-picked your commit and solved most of the issues in another commit.

What I didn't solve is #3 (see above), and in #5 "make the table fill the available horizontal space" because in that case due to an endless loop the horizontal size of the widget grows indefinitely, and maybe the scrollbar issue is still present (also see above).

If these are not blocking, please integrate it. Thank you!

lassoan (approved):
Thank you @cpinter, this should be good to go. We can take care of the remaining issues in follow-up commits.

lassoan:
Thank you for the nice work @cpinter. I keep forgetting that we now have the documentation in the repository so we can keep documenting the features that we are adding. Could you add a short description of markups measurements to markups module user and developer documentation page?

cpinter:
Yes of course.

cpinter:
See https://github.com/Slicer/Slicer/pull/5342 for the documentation

PR #5324, opened 2020-11-27
BUG: Fix markup position jumping when using interaction handles

### LINKED ISSUE #5257: Plane position jumps when interaction handles are used
[Give a short summary of what the problem is]

## Steps to reproduce
Create a random plane, enable the interaction handles and increase the glyph size to make them visible.
Hover over one of the straight arrows, when it becomes yellow (highlighted) try to pan the plane in the same direction as arrow pointing at. (to the left in the screenshot). 

Notice that plane immediately displaces quite a bit in negative direction of the arrow, and then gradually starts to move in the correct direction. 

This is a big issue when you are trying to make subtle changes to the planes position. 

image

## Environment
- Slicer version: 4.11.20200930 r29402 / 002be18

Sunderlandkyl:
Previous implementation of interaction handle translation/rotation tried to align the position of the handle to the current mouse position, which had the issue that the markup handle would "jump" if the user didn't click exactly on the center of the handle.
This is fixed by transforming the markup handles based on the difference between the current and previous mouse positions.

fixes #5257

### DIFF
--- a/Modules/Loadable/Markups/VTKWidgets/vtkSlicerMarkupsWidget.cxx
+++ b/Modules/Loadable/Markups/VTKWidgets/vtkSlicerMarkupsWidget.cxx
@@ -973,12 +973,15 @@ void vtkSlicerMarkupsWidget::TranslateWidget(double eventPos[2])
     // Only perform constrained translation if the length of the axis is non-zero.
     if (vtkMath::Norm(translationAxis_World) > 0)
       {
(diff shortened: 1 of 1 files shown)


lassoan (approved):
Looks good, thank you very much for fixing this.

cpinter:
Thank you, this makes the interaction handles much easier to use if precision is needed!

PR #5325, opened 2020-11-28
ENH: Adds logic and interaction for Markup position status

smrolfe:
This commit adds functionality for using unplaced fiducial points as discussed in this [thread](https://discourse.slicer.org/t/proposed-changes-to-markups-to-support-unplaced-points/12714). The main application will be for creating landmark templates. No changes are made to the Control Points table since the current table is planned to be replaced with a new widget. The updates in this commit:

1. add a new position status: "PositionMissing" for points with undefined positions that should not attempt to be placed,
2 handle visibility of undefined\missing points
3. displays position of undefined\missing points as [0,0,0]
4. fix the existing “Add new markup at origin” button and places an undefined point,
5. add a button to set position of highlighted control point as undefined,
6. add a button to toggle "PositionMissing" status of highlighted control point
7. change the point placement logic to default to placing the first undefined point if one exists.

@lassoan


smrolfe:
I'm adding the to-do list from the developer's call here:

1. Icon displaying status in table, format following the segment editor
2. "Place" button in table column
3. Highlight point being placed and scroll to this position in table
4. Update new table buttons with higher resolution and transparent background

Thanks @lassoan, @pieper, and @jcfr for your input.

smrolfe:
This latest commit addresses the deletion of preview points by adding an "AutoCreated" status to control points and adds cycling of states in the position status column back to "Defined" at the original position. It also fixes a two small bugs in the current Markups module:

- stretching the last column of the control points table 

- allowing control point table interactions after exiting point placement mode with a right click.

lassoan (commented):
Thank you, this will be awesome!

A few things to fix/improve:

- If control point is made visible manually (click eye icon) then it is displayed in the viewers, although no valid RAS position is displayed. Update logic in the displayable manager so that if the position is undefined then the point is not displayed (even if its visibility is enabled). Showing an undefined point in viewers would be very confusing especially because the position is (correctly) not displayed in the table.
- If you want to replace a defined point then you need to click twice (first click undefines, second click edits). However, two click is counted as a double-click and detected as an edit trigger. Disable all the edit triggers for the position status column.
- Start the column tooltip with a short, clear name, such as "Position status" (you can add some more help text after that). This helps when you write user manuals to refer to the column, because the column does not have a text title, just an icon.
- Add explicit setting of a position state using right-click menu items (as in the segment list). It is somewhat inconvenient that you need to keep clicking until you get the icon you want, instead of just choosing it from a menu (as it is done in the segments list)
- When you set "edit" position mode for a control point then set the previously edited control point position back to undefined. It is confusing why we can have multiple control points in edit state.
- I see these warnings in the log: `Generic Warning: In C:\D\S4\Modules\Loadable\Markups\MRML\vtkMRMLMarkupsNode.cxx, line 1757 Position status set to missing for point index: 0`
- make the markups name or description column expanding
- lock the GUI until a markups node is selected: old bug, but makes things confusing
- there is a white halo around the position state icons: use the same icons that are used in segment list and create icons with similar size and style for the new X icon
- probably there should be a way to lock the markup list (not allow to add more points). There is a maximum number of control points variable that could be used, but it is not exposed on the GUI and if the max number is reached then a new markups node is created automatically when the last point is placed (when in persistent placement mode), which is not ideal (the user probably did not want to create a new markup).
- Scroll the table to the markup position that gets into edit state (make this behavior optional)
- Move out the markups table into a separate widget, into the Widgets subfolder, similarly to qSlicerSimpleMarkupsWidget. This is important for two reasons: it makes the module widgets simpler, and allows using this widget in any other module. It could be called qSlicerMarkupsControlPointsTableWidget.

smrolfe:
Thanks for your review @lassoan!

lassoan:
Fix the file conflicts and let us know if the pull request is ready for review/merge.

lassoan:
Let us know if this is ready for review.

smrolfe:
Thanks @lassoan I have the toolbar implemented and am working on the logic that updates it. I'll post when it's done, or if I get stuck.

lassoan:
@smrolfe let us know if this is ready for review.

smrolfe:
Thanks @lassoan, it is ready for review. Could you advise me on the best way to correct the typo in a prior commit message?

smrolfe:
A few notes on the most recent changes in this version:
- Markups module Control Point menu locked until a markup node is created
- Option to set max number of markup points from the Markup module removed
- Button to lock markup point number added to Control Point menu of the Markups module. Insertions and deletions are prevented, but edits are allowed.
- Same option added to the More Options menu in the Markups toolbar
- Unset last/all point position added under the Delete menu in the Markups toolbar
- Rename "Persistent placement" mode "AutoPlace"
- Keyboard shortcut "Shift +A" creates a new node, can continue adding points in AutoPlace mode
- Keyboard shortcut  "Shift +Z" toggles AutoPlace mode
- Keyboard shortcut "Space" toggles placement mode. No point is placed, so if it is accidentally triggered, placement mode can be toggled back with the same key or right mouse click. One slight issue is that the mode is only entered when there is some mouse interaction. It works fine as is, but it would be preferable to eliminate this lag if possible.
- I implemented a separate persistence mode called "AutoPlace Nodes" to distinguish between auto-placement of control points and markup nodes. However, after testing, I removed this option, since we found this option to be confusing to use and the added functionality limited, especially given the new keyboard shortcuts.

Modules/Loadable/Markups/Widgets/qMRMLMarkupsToolBar.cxx
```diff
@@ -0,0 +1,530 @@
...
+  q->addWidget(this->MarkupsNodeSelector);
+
+  connect(this->MarkupsNodeSelector, SIGNAL(currentNodeChanged(vtkMRMLNode*)),
+    q, SIGNAL(activeMarkupsNodeChanged(vtkMRMLNode*)) );
+  connect(this->MarkupsNodeSelector, SIGNAL(currentNodeChanged(vtkMRMLNode*)), q, SLOT(onMarkupsNodeChanged()));
+
+  // set up keyboard shortcuts
+  q->addCreateNodeShortcut("Shift+A");
+  q->addTogglePersistenceShortcut("Shift+Z");
+  q->addPlacePointShortcut("Space");
```
jamesobutler, line 100: If I'm typing to rename a node and I rename the node "A Zoo" will it create a node, toggle persistence and toggle placement mode?

smrolfe: Thanks @jamesobutler, these shortcuts work regardless of the mouse placement, but if a text field is active then the shortcuts are blocked, so this should not be an issue. Three button shortcuts that we tested were not easy to use from a workflow standpoint, but I could replace Shift with Ctrl to make it more distinct.

lassoan: I would prefer not to use single keys or Shift+keys or space as keyboard shortcuts in Slicer core, because it could cause conflicts with more specialized workflows and modules, which want to define their own, simplest-possible shortcuts. For example, space is already used in Segment Editor for toggling between last two segment editor effects - using the same key for markups placement enable/disable would probably not work at all. These kind of shortcuts can be temporarily added in a dedicated landmarking module (similarly how Segment Editor module temporarily adds its shortcuts).

lassoan:
> Could you advise me on the best way to correct the typo in a prior commit message?

In general, if you want to fix a commit in the middle of the history then you can choose to rebase the repository on itself and choose commits you want to edit (for example TortoiseGit has a convenient GUI for this). But you don't need to worry about this, as probably there will be many more small changes requested and in the end we can squash the final version into a single commit.

lassoan (commented):
Thank you for all these developments, these will be awesome.

There are a number of small things to improve:
- If number of control points are fixed then do not show "delete" option in toolbar. It could be disabled but it would be better if it was replaced by "Unset".
- We need to keep all mouse modes on the main toolbar. It is OK to not have the submenu on the toolbar but we should always clearly see what is the current mouse mode by glancing at "Mouse interaction" toolbar. If there is no active markup node selected then the button could be either disabled or could make the markups toolbar visible.
- AutoPlace: the behavior is currently placing multiple points/markups. If we keep the current behavior (which I think is OK) then the name should reflect this. Maybe something like "Place multiple"
- After starting to defined the position of a predefined control point, right-click should unset/delete it (now it just stays in edit mode)
- Add "Edit properties..." function to markups toolbar so that you can quickly open the current node in the markups module. It could be in the node selector (I think it can be enabled by a single flag in the node selector).
- When "Editing properties..." is called then the markups tree view should scroll to where the currently selected markup node is.
- When I opened Markups module using "Edit properties..." from Data module then "Control points" section was disabled. Need to make sure that the widget updates properly.
- If point position is not defined then don't show it in views. This can be fixed in the markups displayable manager. Do not make the control point hidden when the point gets undefined.
- When autoplace is enabled then every second ROI markups node have different color.
- With the addition of all these great features, control points section in markups module GUI has become quite messy. Need to simplify it / make it prettier. A few ideas:
  - move click to jump slices closer to the list (it could be a combobox: click action -> none, jump slices - offset, jump slices - centered, ...)
  - Move interaction in views and Lock number of control points into a single row: Interaction
  - Change "Hide RAS" to "Show position" (or "Show coordinates"). Maybe replace with a comobox: "Coordinates" -> hide, local, world
  - Filtering will be added to markups the same way as for segments. Make sure to reserve space for this.
- I've added more comments in the source code - search for `TODO:markups`

smrolfe:
Thanks @lassoan for your review. I will plan to make these improvements this week.

lassoan:
@smrolfe do you have any updates? Now that the extensions manager update and VMTK build issues are getting sorted out, we getting close to the new stable release and it would be important to include your enhancements.

smrolfe:
Hi @lassoan Thanks, I am getting close. I'm debugging a few issues and should have an update shortly. Will you be joining the developers meeting tomorrow? I can plan to check in then if so.

smrolfe:
> If point position is not defined then don't show it in views. This can be fixed in the markups displayable manager. Do not make the control point hidden when the point gets undefined.

One question I have is on this improvement. After looking at the markups displayable manager, it was not clear to me on how to do this without hiding the point. Would you be able to give me a little advice on how to implement?

lassoan:
> Hi @lassoan Thanks, I am getting close. I'm debugging a few issues and should have an update shortly. Will you be joining the developers meeting tomorrow? I can plan to check in then if so.

Yes, I'll join the developer meeting tomorrow.

> One question I have is on this improvement. After looking at the markups displayable manager, it was not clear to me on how to do this without hiding the point. Would you be able to give me a little advice on how to implement?

Instead of just checking GetNthControlPointVisibility, you need to always check for the position status of the markup. Maybe you can add a helper function (something like `GetNthControlPointPositionVisibility`) that you can call instead.

smrolfe:
I've added the updates from the last code review aside from updating the markups JSON schema. When I try to commit the updated version of this file, I'm getting errors related to the length of two lines already in the file: https://github.com/Slicer/Slicer/blob/dd407fb22cb2c50527a59161ba6e0125e4e14bc6/Modules/Loadable/Markups/Resources/Schema/markups-schema-v1.0.2.json#L102

and 

https://github.com/Slicer/Slicer/blob/dd407fb22cb2c50527a59161ba6e0125e4e14bc6/Modules/Loadable/Markups/Resources/Schema/markups-schema-v1.0.2.json#L163

Is there a way to get around this requirement, or to break up the string over multiple lines in the JSON format?

lassoan:
You may need to re-run SetupForDevelopment.sh.

If it does not help you can just temporarily rename the precommit hook file in the hooks folder.

lassoan:
@smrolfe have you addressed all my [comments above](https://github.com/Slicer/Slicer/pull/5325#pullrequestreview-693433769)?

smrolfe:
@lassoan Yes all your comments are addressed in the last commit.

smrolfe:
@lassoan Yes all your comments are addressed in the last commit.

muratmaga:
any updates on this? We are waiting the features in these for a collaborative project.

lassoan:
I'm reviewing and testing the code and will provide feedback soon by Friday.

lassoan:
I've rebased your branch on latest master and fixed a few crashes and force-pushed the result to this branch. Switch to this new branch (discard your old branch with all the merges, etc.) to continue the work on it.

I've tested this latest version and although it has rough edges, it probably makes sense to get it out sooner so that more people can provide feedback and finalize the implementation based on these feedbacks. However, there are a few things that must be fixed because they would cause issues for current users.

High priority (must be fixed before integration):
- Place button in the mouse mode toolbar must not be disabled. It would be a major regression. It could bring up the markups toolbar instead when no markup is active.
- "Point number locked" gets disabled automatically when clicking to place points. Instead, it should stop and disable point placement when all points are defined.
- Add button to the toolbar that opens the Markups module.
- "Edit position of highlighted fiducials" behaves inconsistently. If I select 3 (already placed) control points then the first one becomes being placed, the second get undefined, the third remains defined.
- Right-click menu actions don't always work. For example, I experienced that the third point out of 4 points did not change the state.
- Make the status column narrow (same way as in Segment list). If RAS coordinates are not displayed then status column can be very wide.

Medium priority:
- Right-click menu action names are very long and not clear.
  - "Edit position of highlighted fiducial(s)" -> "Place point" (only show it if a single control point is selected)
  - "Set position missing for highlighted fiducial(s)" -> "Skip point" (only show if at least one of the points are not skipped already)
  - "Unset position of highlighted fiducial(s)" -> "Undefine point position" (only show if not all the points are undefined already)
  - "Restore position of highlighted fiducial(s)" -> "Restore last known position" (only show if position is not placed already for all the points)
- Lots of errors like this: "Cannot create accessible child interface for object:  QTableWidget(0x227fe713370, name = "activeMarkupTableWidget")  index:  45"
- The toolbar should start with the add buttons, because workflows should generally go from left to right and you need to create a markup first.
- Do not turn off point's visibility when unsetting the position.
- Add a skip point button to the toolbar.

Low priority:
- "Point number locked" flag meaning is unclear.
- "Visibility" and "Locked" sounds inconsistent. If we use "locked" word then we should use "visible".
- It is easy to miss the point status column (the widget is very wide and you need to scroll to see the last column). Maybe it would help if the status column would come before the RAS columns.
- When hovering over an item in the status column then it should show the state in the tooltip (placed, undefined, missing)

smrolfe:
Thanks for your helpful review @lassoan! I have a couple clarification questions I am hoping you can help me with.

> "Point number locked" gets disabled automatically when clicking to place points. Instead, it should stop and disable point placement when all points are defined.

I am not seeing this behavior, unless I am misunderstanding. Currently, the place mode button starts point placement in a new node of the same type if the active node has reached the maximum number of points allowed. This behavior is identical to the angle and line node creation. When persistence placement mode is active, placement mode is exited when the maximum point number is reached.

> Right-click menu actions don't always work. For example, I experienced that the third point out of 4 points did not change the state.

I am having trouble replicating this error. Could you let me know if you are able share a scene file or screen capture from your test?

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsNode.h
```diff
@@ -264,6 +269,7 @@ class  VTK_SLICER_MARKUPS_MODULE_MRML_EXPORT vtkMRMLMarkupsNode : public vtkMRML
 
   /// Clear out the node of all control points
   virtual void RemoveAllControlPoints();
+	virtual void UnsetAllControlPoints();
```
jamesobutler, line 272: It looks like your indentation is messed up in several places where it has become overindented. This comment is just pointing out one location, but appears to be happening across several files in this PR.

smrolfe: Thanks @jamesobutler  I seem to have a had an issue with my editor, I'll take a look.

smrolfe:
This commit:

- restores functionality to the mouse mode place button
- updates to the place\add node button icons in the  Markups module and Markups toolbar,
- reorders the Markups toolbar to follow the workflow and improve readability,
- adds tooltip to the Markups toolbar "more options" button,
- adds a button to legacy Annotations module to the Markups toolbar,
- adds line and ROI node create-and-place buttons to the Annotations module to replace unmarked buttons in the mouse mode toolbar,
- updates the position status column in the Markups module control points table to be narrow, similar to the segment list in the Segment Editor,
- clarifies position status button tooltips in the Markups Module, and
- fixes synchronization bug between the "Fixed point number" status displayed in the Markups module and toolbar.

Two planned changes were made and then reversed after testing:
- Mouse mode place button raising the Markups toolbar when no active node is present. This was confusing when the toolbar was already displayed, as the button looks active, but does not appear to have a function, especially as most users will expect this button to start placing markup points. Allowing this button to create a default markup type (fiducial node) is in line with the users expectations of this button's function and also brings up the Markups toolbar.
- Separating the place and create functionality in the Markups toolbar. When a user is placing a markup type that has a max number of points, the place button will either become inactive (when no other markups are present) or will revert to placing points in a previously placed node that has not reached the maximum number. This behavior was inconvenient with the mouse mode placement button and results in points being placed in unexpected nodes when there are multiple markups in the scene. We revised this by allowing the place button to initiate markup placement of the current node type, with node creation allowed when the max points are reached.

Base/QTGUI/Resources/qSlicerBaseQTGUI.qrc
```diff
@@ -87,6 +88,8 @@
         <file alias="SlicerInteractivePlotting.png">Icons/Medium/SlicerInteractivePlotting.png</file>
         <file alias="SlicerDatabase.png">Icons/Medium/SlicerDatabase.png</file>
         <file alias="SlicerImportDICOM.png">Icons/Medium/SlicerImportDICOM.png</file>
+		    <file alias="SlicerPointNumberLock.png">Icons/Medium/SlicerPointNumberLock.png</file>
+		    <file alias="SlicerPointNumberUnlock.png">Icons/Medium/SlicerPointNumberUnlock.png</file>
```
jamesobutler, line 92: Appears you still have some indentation issues going on here.

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsNode.h
```diff
@@ -245,17 +247,20 @@ class  VTK_SLICER_MARKUPS_MODULE_MRML_EXPORT vtkMRMLMarkupsNode : public vtkMRML
     PointStartInteractionEvent,
     PointEndInteractionEvent,
     CenterPointModifiedEvent,
+		FixedNumberOfControlPointsModifiedEvent,
```
jamesobutler, line 250: Indentation issue

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsNode.h
jamesobutler, line 272: Indentation issues here in this file

Modules/Loadable/Markups/Resources/Schema/markups-schema-v1.0.2.json
```diff
@@ -30,586 +30,651 @@
...
-                                    "$id": "#markup/type",
-                                    "type": "string",
-                                    "title": "Basic type",
-                                    "enum": ["Fiducial", "Line", "Angle", "Curve", "ClosedCurve", "Plane", "ROI"]
+                          "properties": {
+                            "type": {
+                              "$id": "#markup/type",
+                              "type": "string",
+                              "title": "Basic type",
+                              "enum": [ "Fiducial", "Line", "Angle", "Curve", "ClosedCurve", "Plane", "ROI" ]
```
jamesobutler, line 38: Apologies for nitpicking on this, but looks like an unintended change to indentation here that is causing more line diffs in this PR than there really are.

Modules/Loadable/Markups/Resources/UI/qSlicerMarkupsModule.ui
```diff
@@ -409,88 +501,83 @@
...
-           <item>
-            <layout class="QHBoxLayout" name="horizontalLayout_7">
+         <string>Advanced</string>
+        </property>
+        <property name="collapsed">
+         <bool>true</bool>
+        </property>
+        <layout class="QVBoxLayout" name="verticalLayout">
+         <item>
+          <layout class="QHBoxLayout" name="horizontalLayout_7">
```
jamesobutler, line 580: This diff change necessary as well? Or indentation issue?

Modules/Loadable/Transforms/Widgets/Resources/qSlicerTransformsModuleWidgets.qrc
```diff
@@ -1,7 +1,7 @@
 <!DOCTYPE RCC><RCC version="1.0">
   <qresource>
-	<file>Icons/GlyphIcon.png</file>
-	<file>Icons/GridIcon.png</file>
-	<file>Icons/ContourIcon.png</file>
+  <file>Icons/GlyphIcon.png</file>
+  <file>Icons/GridIcon.png</file>
+  <file>Icons/ContourIcon.png</file>
```
jamesobutler, line 5: This diff necessary?

jamesobutler (commented):
In recent commit https://github.com/Slicer/Slicer/pull/5325/commits/816478f3238402f908e5b6aa76dfa3e920840d31, there is the complete removal of `qSlicerMouseModeToolBar.cxx` intended? There is still a `qSlicerMouseModeToolBar.h` and `qSlicerMouseModeToolBar_p.h`.

smrolfe:
Thanks @jamesobutler, I am still testing\making a few changes and will make sure these get incorporated.

smrolfe:
@lassoan and @jamesobutler this pull request is ready for review. In addition to the whitespace and minor fixes I have made a couple changes to the MouseModeToolBar place button, in response to our conversation on Tuesday.

- When there are no active markups in the scene, the place button will toggle the Markups toolbar. Since it is a toggle, it should be clear something is happening, even if the toolbar was previously displayed. 
- The generic place icon is used for the place button in this case, to help differentiate from the active place mode options.

lassoan:
This is getting quite usable and intuitive now. There are just a few things to address.

> When there are no active markups in the scene, the place button will toggle the Markups toolbar. 

I like the new behavior but there are two things that would be necessary to make this look clear and consistent.

- [ ] Change the icon. The icon could be composed of multiple markups or somehow indicate it will show a toolbar. The icon with the "M" looks bad and it is not clear what it means and we cannot have letters in icons (the "M" will not make sense when we translate "markup" word to a different language).
- [ ] Add a drop-down menu with a show/hide toolbar. Without that, it is very confusing that there is no way to show/hide the toolbar with the button anymore (I need to remove all markup nodes from the scene to be able to show/hide the toolbar with this button again).

Other findings:

- [ ] "Lock the number of markup points" must prevent placement (make the button disabled). I tis really bad now that I keep placing points and suddenly, without any warning, a new markup node is created. Instead, it should be made very clear to the user that placement of points in this markup node is completed.
- [ ] Do not close the eye icon of a control point in the list when its position is unset. You can just keep as is (reflecting the visibility state of the point), or if you want to better differentiate the unset points then gray out the icon or create a special visibility icon that is shown for unset points.
- [ ] "Lock the number of markup points" button should be renamed to "Lock the number of control points"
- [ ] Run all the tests and see if there are any regressions (tests failing or timing out that work fine in the latest master). I saw that there are memory leaks due to an observation not removed, but maybe they are not due to your changes.

muratmaga:
@lassoan 
"Lock the number of markup points" must prevent placement (make the button disabled). I tis really bad now that I keep placing points and suddenly, without any warning, a new markup node is created. Instead, it should be made very clear to the user that placement of points in this markup node is completed."

is a complicated one. I remember in discourse discussions that people want to continue placing lots of lines quickly, without having to do any additional clicks of node creation etc.  While in theory I agree with you, this is a one of those things that we should probably push the behavior as is, and get a feedback and iterate over.

jamesobutler:
I would have to agree that I would find it very confusing if I was suddenly creating multiple nodes rather than adding points to a specific node. Adding multiple points to a markups fiducial list makes sense however being in a state that adds multiple line modes does not. I think multi node creation should be a deliberate action rather than a passive automatic action. From how Slicer typically works, users specifically declare to create a new node.

muratmaga:
" was suddenly creating multiple nodes rather than adding points to a specific node", but thats the default behavior of the place mode for lines and angles for years now!

E.g., in the stable, enable persistent mode, switch to angle or line tool, and you will see new nodes will be created one after another, without any user input. So the current behavior in revision for the fiducial nodes (with fixed amount of points in them), is consistent with that behavior.  
 
I personally never liked (I think all node creation actions should be explicit and initiated by the user),  but some people apparently benefiting from this behavior. 

I am not against revising and changing it. But whatever the change is, it needs to be consistent across different markup types. I don't think one type of behavior for angle/line nodes vs fiducial/curves (again with predetermined number of points in them) would make sense. 

The other concern I have is that I really need to see a form of this markup being incorporated to the Slicer stable for SlicerMorph users and hopefully a version that we have consensus on. That's why my suggestion is to push changes to the preview, let the community know and collect feedback and revise (prior to stable).

smrolfe:
Thanks @lassoan I have addressed the two issues you identified and am testing updates that address your other findings. I'll update the thread again when that's done.

smrolfe:
This commit updates the user interface to separate the create node and place function. The last changes to markups toolbar update eliminated my concerns with this structure and I think it's relatively easy and intuitive to use now.

I'm currently building the tests for this branch. To compare to the master, should I build it on my machine and compare the test outputs?

 @lassoan @jamesobutler @muratmaga

lassoan:
> I'm currently building the tests for this branch. To compare to the master, should I build it on my machine and compare the test outputs?

Yes, you run the tests on your computer and check that there are no more failed tests on your computer compared to the [dashboard](https://slicer.cdash.org/index.php?project=SlicerPreview).

A few more code review comments:
- [  ] Important: The API should only be changed if it is absolutely necessary. For example, renaming `qSlicerMarkupsPlaceWidget::deleteButton()` to `qSlicerMarkupsPlaceWidget::deleteUnsetButton()` would break many extensions that use this method. If we really want to rename a method then we can add the new method and make the old method deprecated (keep the old method but add a comment to its documentation that it is deprecated and if it is called then log a warning that the method is deprecated). Please revert this deleteButton name change (it is enough to just document that the delete button may just unset point). Either revert these changes or add new methods and deprecate the old methods: placeModeActionTriggered, onHideCoordinateColumnsToggled.
- [ ] int currentColorTableIndex; -> this must not be public
- [ ] void updateCurrentColorTableIndex(); -> fix capitalization (VTK class methods start with capital)
- [ ] bool getNewMarkupsColor(double currentColor[3]); -> fix capitalization
- [ ] /// Delete last placed markup point.  void modifyLastPoint(); -> comment name is inconsistent with method name
- [ ] class QSpinBox; is not needed in qSlicerMarkupsModuleWidget.h
- [ ] remove dead code (//void activeMarkupsNodeChanged(vtkMRMLNode* activeMarkupsNode);) in qSlicerMarkupsModuleWidget.h

smrolfe:
@lassoan I've found and corrected a small bug in the order of placing undefined points and completed fixes for the issues in your previous comments. The tests are re-running now.

Modules/Loadable/Markups/VTKWidgets/vtkSlicerMarkupsWidget.cxx
```diff
@@ -808,7 +862,7 @@ void vtkSlicerMarkupsWidget::Leave(vtkMRMLInteractionEventData* eventData)
       {
       interactionContext = eventData->GetInteractionContextName();
       }
-    markupsDisplayNode->SetActiveComponent(vtkMRMLMarkupsDisplayNode::ComponentNone, -1, interactionContext);
+    //markupsDisplayNode->SetActiveComponent(vtkMRMLMarkupsDisplayNode::ComponentNone, id, interactionContext);
```
jamesobutler, line 865: This dead code? Or accidental commented out code?

smrolfe: Thanks @jamesobutler, this is dead code. I have just removed.

jamesobutler:
As you get closer to be ready to be merged, currently this PR is failing the required "CommitCheck" status check which makes sure commit messages following standard prefixes documened at https://github.com/Slicer/Slicer/blob/master/CONTRIBUTING.md#how-to-write-commit-messages-. Currently this PR fails the check due to commits such as `Merge branch 'Slicer:master' into LMUS` which are merge commits. Squashing down commits whether one big one, or organized smaller ones based on enhancements versus other bug fixes that exist in Slicer `master` that are being fixed here will then allow for the required check to pass here.
image

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsNode.cxx
```diff
@@ -702,7 +755,7 @@ void vtkMRMLMarkupsNode::RemoveNthControlPoint(int pointIndex)
     {
     this->InvokeCustomModifiedEvent(vtkMRMLMarkupsNode::PointPositionUndefinedEvent, static_cast<void*>(&pointIndex));
     }
-  this->InvokeCustomModifiedEvent(vtkMRMLMarkupsNode::PointModifiedEvent, static_cast<void*>(&pointIndex));
+  //this->InvokeCustomModifiedEvent(vtkMRMLMarkupsNode::PointModifiedEvent, static_cast<void*>(&pointIndex));
```
jamesobutler, line 758: Commented out code that was triggering PointModifiedEvent

smrolfe:
@jamesobutler I have a question on squashing the commits on my branch. Could you advise if this is the correct approach, or if there is a better method? To consolidate the 28 commits on the LMUS branch, I reset the head by 28 commits and then created a new commit with the changes: 
`git reset --soft HEAD~28`
`git commit -am "ENH: Improve markups control points state management and add markups toolbar"`

jamesobutler:
Yes, resetting the head like that to add a new commit followed by force pushing the branch is one of many ways to do it. 

There is also this suggestion offered by @lassoan to another user in a different PR with the similar issue
> you can squash all the commits in TortoiseGit: right-click on the CTK folder, click "Git show log", select all your commits, right-click, choose "Combine to one commit".

I personally use TortoiseGit to manage squashing of fix-up commits into a subset of organized commits for a large bit of work if there are separate things going on. I find this a bit more user friendly then attempting interactive rebasing through command line git. I'll use command line or some other GitHub based application for more simple things like creating new branches, but will leave squashing fix-ups and rebasing to bring in latest commits from the main development branch through the rebase process using TortoiseGit. https://tortoisegit.org/docs/tortoisegit/tgit-dug-rebase.html

TortoiseGit will stop in the rebase process when there are merge conflicts which will ultimately happen when squashing down fix-up commits in a more complicated branch of many commits. Instead of using TortoiseGit's editor to resolve conflicts, I will actually keep my VS Code IDE up which will reflect the branch during the process, handle the conflicts using VS Code, then stage the changes and go back to my TortoiseGit window to actually commit the change.

lassoan:
Important: Before doing any operation that changes commit history (squashing commits, rebasing, etc.), create a branch from your current branch as a backup (you can even make a local copy of the entire repository, if you are not confident), because you can lose your work if you make a mistake.

jamesobutler:
Currently your branch is behind `master`. See the following at https://github.com/smrolfe/Slicer/tree/LMUS. 
image
You have incorporated those diffs into your 1 commit which is why there are a lot more line diffs showing in this PR now. You will need to rebase your branch so that it instead reflects as "1 commit head Slicer:master".

Base/QTCore/qSlicerExtensionsManagerModel.cxx
```diff
@@ -1741,6 +1741,12 @@ bool qSlicerExtensionsManagerModel::installExtension(
   const ExtensionMetadataType& extensionIndexMetadata =
     Self::parseExtensionDescriptionFile(extensionIndexDescriptionFile);
 
+  // Enable or disable the extension if state was not already set
+  if (!extensionMetadata.contains("enabled"))
+    {
+    extensionMetadata.insert("enabled", d->NewExtensionEnabledByDefault);
+    }
+
```
jamesobutler, line 1749: This might be here due to resolving some merge conflicts incorrectly. This diff would be a partial revert of https://github.com/Slicer/Slicer/commit/fb23d84810549856ebc73f7a9f1c95793bf48fc1. You should just remove this diff change from your 1 commit now `ENH: Improve control point state management and add markups toolbar`.

smrolfe: Got it, thank you!

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsNode.cxx
```diff
@@ -992,10 +1044,12 @@ void vtkMRMLMarkupsNode::SetNthControlPointPositionOrientationWorldFromArray(
...
 
+  if (positionStatus != vtkMRMLMarkupsNode::PositionPreview)
+    {
+    controlPoint->AutoCreated = false;
+    }
   // Transform orientation matrix to world
-  this->TransformOrientationMatrixFromWorldToNode(pos, orientationMatrix_World, controlPoint->OrientationMatrix);
-
-  if (associatedNodeID)
+  this->TransformOrientationMatrixFromWorldToNode(pos, orientationMatrix_World, controlPoint->OrientationMatrix);  if (associatedNodeID)
```
lassoan, line 1052: split this into two lines

lassoan:
I've reviewed the code and tested the current implementation throughout this week and found the followings.

Overall, most issues are sorted out, except the change in behavior in sticky active (highlighted) state of control points. I understand that there is a need to visually quickly cross-reference control points in markups table widgets and viewers. However, the current implementation (the mouse pointer moves away from a control point it remains active (highlighted color)) has serious issues:
- It would be hard to implement a sticky active state correctly. There can be any number of active control points - one for each context (desktop, left virtual reality controller, etc). There can be any number of markups control points table widget: there is one in markups module, but modules can add their own control points tables (for example, Fiducial registration wizard module displays two tables, one for fixed and another for moving markups). It would be hard to synchronize the very volatile selection state of the table widget with the persistent state of a MRML node (that is even stored on disk). There can be multiple markups display nodes (that store the active control point) for each markups data node (that is associated with each table widget). There can be multiple control points selected in a table widget, but there can only be one active control point per display node.
- Active state makes selection state not hidden (the color is used for temporary highlighting the active control point and showing selected/unselected state), which is not an issue if it is just temporary (while the mouse is hovered over the point) but it is much bigger problem if the active state is more sticky.
- Active state setting is not symmetric: if user selects a control point then the point gets highlighted in viewers, but if the user hovers over a control point in a viewer then it is _not_ selected in markups table widgets.
- Lots of `Cannot create accessible child interface for object:  QTableWidget(0x222861e3540, name = "activeMarkupTableWidget")` errors logged due to changing the selection in an incorrect state (the newly added of modified item seems to be invalid). Changing a table widget's content and selection model at the same time is error-prone. Instead of changing selection model, something like scrolling to the current control point (to ensure the item is visible), change display style (background coloring, font, etc.) would be safer and simpler.

Let's collect the kinds of states of control points that we have currently:

    - MRML control point selected/unselected: stored in markups data node, shown in table widgets (using checkboxes; synchronized across all table widgets), shown in views (using colors)
    - MRML active control point: stored in display node, one for each context (mouse, left/right VR controller), shown in views, **not shown in table widgets**
    - table widget current item: shown in table, only one item can be current in a table at a time, but there can be many table widgets; it is not saved in MRML, **not shown in views**
    - table selected items: shown in the table widget, there can be many tables, there can be many items selected, selection is not synchronized across table widgets; it is not saved in MRML, **not shown in views**

Potential approaches that would allow cross-referencing between table widgets and views:

- We could make MRML active point displayed in tables (the background color of the active row could change, maybe the table could scroll there) table could scroll there
- We could set the control point that the mouse is hovered over in table widget (without clicking, similarly how hovering over a control point in a view makes it active, even without clicking). In the default (desktop) context, in the first display node of the markup node.
- We could fix "right-click menu -> edit properties" action to not just switch to Markups module and select the markups node, but also select the control point in the control points section.
- Add an optional "Quick-select" mode to table widget. In this state, the control point that gets selected will be set to selected state in MRML (and all control points that are not selected in the table widget will be made unselected in MRML). When control point selected/unselected state is changed in MRML then item selection all the table widgets (that have quick-select enabled) would be updated accordingly. In this state, the checkbox column could be hidden. This would make MRML point selection volatile, which would break Fiducial registration module (which uses control point selection state to indicate which points need to be used during the registration) and VMTK centerline extraction module (which uses control point selection state to indicate that a point is inflow or outflow). So, quick-select would need to be optional.

Other issues:

High priority:
- [ ] When a control point is unset then the curve is not updated -- the point should be excluded from the curve generation
  - due to cannot select item (invalid?) - should we just scroll there
- [ ] Slicer crashes if a qSlicerMarkupsPlaceWidget is created and scene is set and then a markups node is set, due to an infinite loop in qSlicerMarkupsPlaceWidget::updateWidget. A IsUpdatingWidgetFromMRML flag can be added to prevent recursive calls to updateWidget (if the widget is already being updated then just return if called again). Search for IsUpdatingWidgetFromMRML in Slicer source code to see examples.

Low priority (can be done after integration):
- [ ] *Display* markups toolbar -> *Toggle* markups toolbar (because the same menu is used for showing/hiding)
- [ ] Display markups toolbar icon is not very informative. A design like this would be better: MarkupsDisplayToolBar
- [ ] Node selector and color selector in the toolbar must have the same height as the toolbar buttons
- [ ] Add a spacer between the new markup buttons and the node selector button

smrolfe:
This commit:
- Saves the active control point table row number in the markups node, used set the first unplaced point placed. This fixes a bug causing the active point to be modified when moving the mouse into the Module window during point placement.
- Updates curves when the point position status changes
- Fixes the recursive calls to qSlicerMarkupsPlaceWidget

lassoan:
This works well. I've only found one problem that would be important to fix (it was much improved but not fully fixed by your last commit), and all the rest can be addressed in follow-up commits.

High priority:
- [ ] When a control point is unset then the curve is updated - good! But then when moving a point the curve gets messed up. For example, draw a curve with 5 points, make the 3rd hidden and move the 4th and 5th point.

Low priority (can be done after integration):
- [ ]  vtkMRMLColorTableNodeRandom color table is not designed for creating nice, distinguishable colors. Instead, use MediumChartColors, DarkBrightChartColors, LightPaleChartColors. Make the color index start at 0 (not at 3) and wrap around if the color index reaches the number of values in the color table. Remove UpdateCurrentColorTableIndex (fuse it into GetNewMarkupsColor). Make GetNewMarkupsColor return a double* (you need to add a `double LastReturnedMarkupsColor[3];` in the class to store the vector that is returned) and add `VTK_SIZEHINT(3)` to make the method conveniently usable from Python.
- [ ] It is inconsistent that ROIs have all the same color if created one by one but have different colors if "Place multiple control points" is enabled.
- [ ] *Display* markups toolbar -> *Toggle* markups toolbar (because the same menu is used for showing/hiding)
- [ ] Display markups toolbar icon is not very informative. A design like this would be better: MarkupsDisplayToolBar
- [ ] Node selector and color selector in the toolbar must have the same height as the toolbar buttons
- [ ] Add a spacer between the new markup buttons and the node selector button

smrolfe:
Thanks @lassoan. I just pushed the fix for the curve update bug. Would you like me to combine these commits?

lassoan:
Awesome!

Yes, please squash the commits and rebase on latest master, and it should be ready to merge.

I'll merge it today if there are no objections.

muratmaga:
Great! 
@smrolfe when you have the chance can you do the short videos we discussed (how it was before and how it is now, and also one for how to use LM templates, and skipping/resetting LMs).

smrolfe:
@muratmaga Sounds good!

@lassoan It should be all set now, thanks again for all your help in getting this done!

smrolfe:
I'd like to pause on this while I submit one final bug fix. I'll update the thread shortly.

lassoan:
OK, no problem, take your time.

lassoan:
While testing, I've found a small inconsistency (does not have to be fixed before integration, but would be nice):
- [ ] When clicking the control point table to edit a point then the point position is filled immediately. This position gets deleted if I move over to a view and leave the view and the edit icon disappears from the table. It would be better if starting editing would just change the icon to edit (would not show the coordinates) and after going over a view and coming back, the edit icon would stay displayed in the table while placement mode is enabled. This would make it clear for the user which point will be the next to place (without need to hover over any view) and would ensure that the table state would not be changed just because the user temporarily moves the mouse over a view.

smrolfe:
@lassoan I found one more place where the curve points were being updated without checking the position status, which is corrected in this commit. I will make the change to the point position display, but am also ok to move forward with the integration now.

lassoan:
Please fix the point position/edit state issue, but no need to rush, if you cannot confidently implement and test it today then tlit can go in tomorrow.

lassoan:
I've added an issue to track all the tasks that we left unresolved in this pull request: https://github.com/Slicer/Slicer/issues/5889
Please refer to it when you submit the fixes.

jamesobutler:
There appears to be a new failing test for qSlicerMouseModeToolBarTest1
https://slicer.cdash.org/test/19435567

Issue #2118, opened 2020-03-12
Editor module documentation hyperlink is broken


slicerbot:
_This issue was created automatically from an original [Mantis Issue](https://mantisarchive.slicer.org/view.php?id=2118). Further discussion may take place here._

PR #5326, opened 2020-11-30
Minor updates to ITK and SimpleITK version

jamesobutler:
Based on desire to update ITK to 5.1.2 in Slicer sooner rather than later (see https://discourse.itk.org/t/patches-for-itk-5-1-2/3635/11), this PR bumps the ITK version to that and SimpleITK also gets a minor update. With SimpleITK v2.0.1 it doesn't need to specify `Module_SimpleITKFilters_GIT_TAG`. This is because https://github.com/SimpleITK/SimpleITK/commit/b71184d1282deff2550026e87e8ce430022e04ea specified that a later ITK version no longer needed the custom SimpleITKFilters remote git tag.  ITK 5.1.2 is even later and meets the criteria to not need the custom SimpleITKFilters remote git tag.

Marking as draft first as I haven't tested the build yet.

### DIFF
--- a/SuperBuild/External_ITK.cmake
+++ b/SuperBuild/External_ITK.cmake
@@ -36,7 +36,7 @@ if(NOT DEFINED ITK_DIR AND NOT Slicer_USE_SYSTEM_${proj})
 
   ExternalProject_SetIfNotDefined(
     Slicer_${proj}_GIT_TAG
-    "v5.1.1"
+    "v5.1.2"
     QUIET
     )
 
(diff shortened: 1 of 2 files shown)


N-Dekker:
Thank you very much, James (@jamesobutler)! For the record, I especially appreciate a Slicer upgrade to ITK 5.1.2 as it includes a bug fix that I proposed, which affects loading transform files into Slicer: [ITK pull request #2118 - BUG: Fix MatrixOffsetTransformBase SetFixedParameters if too few params](https://github.com/InsightSoftwareConsortium/ITK/pull/2118 )

jamesobutler:
I tested on Windows and the tests that failed are the same as the ones that are already failing (http://slicer.cdash.org/viewTest.php?onlyfailed&buildid=2077006).

I have good confidence that this update can be merged. @N-Dekker I'm assuming you tested your ITK fix as part of your work with Slicer since it appears you discovered the issue using Slicer originally.

lassoan (approved):
Thanks a lot @jamesobutler, this looks good to me.

PR #5327, opened 2020-12-01
ENH: Allow showing nodes by drag-and-dropping into views from subject hierarchy tree

lassoan:
A node can be shown in a view by drag-and-dropping it from the subject hierarchy into a view.
If a volume is dragged to a 3D view then volume rendering is displayed.
To show a node in a single view only, hide the node (click on the eye icon in subject hierarchy) and then drag-and-drop it into a view.
To restore visibility of a node in all views, choose "Show in all views" in subject hierarchy item visibility.

**Demo video:**

**https://youtu.be/muIlmbUSz3A**

### DIFF
--- a/Libs/MRML/Core/vtkMRMLSubjectHierarchyNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLSubjectHierarchyNode.cxx
@@ -3654,3 +3654,13 @@ vtkMRMLSubjectHierarchyNode* vtkMRMLSubjectHierarchyNode::GetSubjectHierarchyNod
 
   return scene->GetSubjectHierarchyNode();
 }
+
+//----------------------------------------------------------------------------
+void vtkMRMLSubjectHierarchyNode::ShowItemsInView(vtkIdList* itemIDs, vtkMRMLAbstractViewNode* viewNode)
+{
+  SubjectHierarchyItemsShowInViewRequestedEventData eventData;
+  eventData.itemIDsToShow = itemIDs;
+  eventData.viewNode = viewNode;
+  this->InvokeEvent(SubjectHierarchyItemsShowInViewRequestedEvent, &eventData);
+  // The event will be processed by qSlicerSubjectHierarchyPluginHandler
+}
--- a/Libs/MRML/Widgets/qMRMLUtils.cxx
+++ b/Libs/MRML/Widgets/qMRMLUtils.cxx
@@ -19,8 +19,11 @@
 ==============================================================================*/
 
 // Qt includes
-#include <QStyle>
+#include <QMimeData>
 #include <QPainter>
+#include <QStyle>
+#include <QUrl>
+#include <QUrlQuery>
 
 // CTK includes
 #include "ctkVTKWidgetsUtils.h"
@@ -180,5 +183,28 @@ void qMRMLUtils::qColorToColor(const QColor &qcolor, double* color)
     color[0] = qcolor.redF();
     color[1] = qcolor.greenF();
     color[2] = qcolor.blueF();
+    }
+}
+
+//------------------------------------------------------------------------------
+void qMRMLUtils::mimeDataToSubjectHierarchyItemIDs(const QMimeData* mimeData, vtkIdList* idList)
+{
+  if (!mimeData->hasFormat("text/uri-list") || !idList)
+    {
+    return;
+    }
+  idList->Reset();
+  foreach(QUrl url, mimeData->urls())
+    {
+    if (!url.isValid() || url.isEmpty())
+      {
+      continue;
+      }
+    if (url.scheme() != "mrml" || url.host() != "scene" || url.path() != "/subjecthierarchy/item")
+      {
+      continue;
+      }
+    QUrlQuery query(url.query());
+    idList->InsertNextId(query.queryItemValue("id").toLong());
     }
 }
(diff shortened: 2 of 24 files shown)


Modules/Loadable/VolumeRendering/SubjectHierarchyPlugins/qSlicerSubjectHierarchyVolumeRenderingPlugin.cxx
```diff
@@ -184,88 +186,159 @@ void qSlicerSubjectHierarchyVolumeRenderingPlugin::showVisibilityContextMenuActi
...
+  double rasBounds[6] = { 0.0 };
+  volumeNode->GetRASBounds(rasBounds);
+  double cameraFocalPoint[3] =
+    {
+    (rasBounds[0] + rasBounds[1]) / 2.0,
+    (rasBounds[2] + rasBounds[3]) / 2.0,
+    (rasBounds[4] + rasBounds[5]) / 2.0,
+    };
+
+  // Get view nodes to reset the FOV of
```
cpinter, line 212: Seems like an unfinished comment

cpinter:
This looks like a useful, intuitive feature, thank you!

I guess it makes sense to only show an item in the slice view you drop it into, given that the views are not linked by default.
However, this is still something that I find strange. The slice views are linked in a way implicitly because the different orientation of the three make up the three main orientations, so one (at least I) would assume that this means that they are meant to show the same data but sliced in a different way. If they were all axial, I'd assume otherwise. This is a conceptual idea, so I think the PR can be merged (I had only one extremely small comment), but I'd still revisit the "linkedness" of the slice views at some point.

Libs/MRML/Core/vtkMRMLSubjectHierarchyNode.cxx
```diff
@@ -3654,3 +3654,13 @@ vtkMRMLSubjectHierarchyNode* vtkMRMLSubjectHierarchyNode::GetSubjectHierarchyNod
...
   return scene->GetSubjectHierarchyNode();
 }
+
+//----------------------------------------------------------------------------
+void vtkMRMLSubjectHierarchyNode::ShowItemsInView(vtkIdList* itemIDs, vtkMRMLAbstractViewNode* viewNode)
+{
+  SubjectHierarchyItemsShowInViewRequestedEventData eventData;
+  eventData.itemIDsToShow = itemIDs;
+  eventData.viewNode = viewNode;
+  this->InvokeEvent(SubjectHierarchyItemsShowInViewRequestedEvent, &eventData);
```
pieper, line 3664: Is this struct available to an observer in python using the decorator?  If not we could use a vtkObject subclass instead.

pieper (dismissed):
Overall I love it - looks like it will really simplify some common use cases.

I do agree with Csaba though that there may be some other useful modes.  Something like the CompareVolumes functionality to set up a rows of Axi/Sag/Cor or Axi/Sag/Cor/Vol is handy if people have enough screen space.

Libs/MRML/Core/vtkMRMLSubjectHierarchyNode.cxx
lassoan: The SubjectHierarchyItemsShowInViewRequestedEvent (and the associated event data struct) is for internal use only, to allow qMRML widgets to call qSlicer functions (qMRML classes cannot directly depend on qSlicer classes).

vtkMRMLSubjectHierarchyNode::ShowItemsInView or qSlicerSubjectHierarchyPluginHandler::showItemsInView method can be used to request view of subject hierarchy items in a view.

I'll add comments to clarify this.

lassoan:
> However, this is still something that I find strange. The slice views are linked in a way implicitly because the different orientation of the three make up the three main orientations, so one (at least I) would assume that this means that they are meant to show the same data but sliced in a different way. If they were all axial, I'd assume otherwise. This is a conceptual idea, so I think the PR can be merged (I had only one extremely small comment), but I'd still revisit the "linkedness" of the slice views at some point.

View linking is implemented for volume display (if you drag-and-drop a volume into any view of a view group then the image will be displayed in all views in that group if slice linking is enabled). We could implement the same behavior for display of other node types, too. I would merge this as is for now, and do further tunings based on what we experience and users request.

Modules/Loadable/VolumeRendering/SubjectHierarchyPlugins/qSlicerSubjectHierarchyVolumeRenderingPlugin.cxx
lassoan: I think this is grammatically correct, but I'll rephrase it to make it more clear.

Libs/MRML/Widgets/qMRMLUtils.cxx
```diff
@@ -182,3 +185,26 @@ void qMRMLUtils::qColorToColor(const QColor &qcolor, double* color)
...
+    return;
+    }
+  idList->Reset();
+  foreach(QUrl url, mimeData->urls())
+    {
+    if (!url.isValid() || url.isEmpty())
+      {
+      continue;
+      }
+    if (url.scheme() != "mrml" || url.host() != "scene" || url.path() != "/subjecthierarchy/item")
```
jcfr, line 203: Do we have documentation for this scheme or should we consider this an implementation details ?

jcfr (commented):
Thanks for the contribution 🙏

May also be worth adding few line in the Data module doc to describe the new behavior.

lassoan:
> May also be worth adding few line in the Data module doc to describe the new behavior.

Yes, absolutely. I need to get into the habit of updating the documentation along with code changes.

Libs/MRML/Widgets/qMRMLUtils.cxx
lassoan: This is implementation detail for now, as this URL is only applicable within the current scene of the current Slicer instance. 

The same approach could be used for opening real universal URLs directly in the viewer, but I don't have any meaningful use case in mind right now.

PR #5334, opened 2020-12-04
COMP: Update FindVcvars with MSVC 1928 support

jamesobutler:
Based on the changes made in https://github.com/scikit-build/cmake-FindVcvars/commit/a4d72a435d39b1adcc717c6202350a5d06056bfc


jcfr (approved):
Thanks @jamesobutler  :pray:

PR #5342, opened 2020-12-14
STYLE: Add doc for markups measurements and scalar display

cpinter:
- Sections about scalars and measurements added in Markups user documentation
- Measurements section collapsed by default

### DIFF
--- a/Docs/user_guide/modules/markups.md
+++ b/Docs/user_guide/modules/markups.md
@@ -10,7 +10,7 @@ This module is used to create and edit markups (fiducial list, line, angle, curv
 
 ### Place new markups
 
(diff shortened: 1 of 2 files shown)


lassoan (approved):
Thank you, it looks good to me!

Note that a "DOC:" commit comment prefix has been added recently that can be used for documentation updates.

cpinter:
> Note that a "DOC:" commit comment prefix has been added recently that can be used for documentation updates.

That was my first choice but the pre-commit hook rejected it. I guess I need to run SetupForDevelopment script again?

lassoan:
It seems this was only added to the github checks. I've submitted a pull request to add it to the local hooks, too.

PR #5345, opened 2020-12-16
EHN: Removing hard-coded markup types from DM

RafaelPalomar:
This commit adds Markup nodes registration capabilities to  `vtkMRMLMarkupsDisplayableManager` and removes the need to hard-code markup nodes. The following changes have been applied:

- Adding a factory method (`CreateInstance()`) in `vtkSlicerMarkupsWIdget`. This method is *pure virtual*, thus forcing sub-classes to define it.
- Adding a `CreateIntance()` to all the existing markups widgets.
- Adding a `RegisterMarkup()` function to `vtkMRMLMarkupsDisplayableManager`. This function will register a new markup and the associated needed items (short name for new fiducials and markup widget type associated to the node).
- Replacing the `std::set<std::string> Focus` member used to keep track of registered nodes, by `std::map<std::string, vtkSlicerMarkupsWidget>` (keeping association between nodes registered and corresponding wiget types), and `std::map<std::string, std::string>` (keeping association between nodes registered and corresponding short names).

### DIFF
--- a/Modules/Loadable/Markups/MRMLDM/vtkMRMLMarkupsDisplayableManager.cxx
+++ b/Modules/Loadable/Markups/MRMLDM/vtkMRMLMarkupsDisplayableManager.cxx
@@ -88,18 +88,20 @@ vtkStandardNewMacro (vtkMRMLMarkupsDisplayableManager);
 //---------------------------------------------------------------------------
 vtkMRMLMarkupsDisplayableManager::vtkMRMLMarkupsDisplayableManager()
 {
-  this->Focus.insert("vtkMRMLMarkupsAngleNode");
-  this->Focus.insert("vtkMRMLMarkupsFiducialNode");
-  this->Focus.insert("vtkMRMLMarkupsLineNode");
-  this->Focus.insert("vtkMRMLMarkupsCurveNode");
-  this->Focus.insert("vtkMRMLMarkupsClosedCurveNode");
-  this->Focus.insert("vtkMRMLMarkupsPlaneNode");
+  this->RegisterMarkup(vtkMRMLMarkupsAngleNode::New(), vtkSlicerAngleWidget::New(), "A");
+  this->RegisterMarkup(vtkMRMLMarkupsFiducialNode::New(), vtkSlicerPointsWidget::New(), "F");
+  this->RegisterMarkup(vtkMRMLMarkupsLineNode::New(), vtkSlicerLineWidget::New(), "L");
+  this->RegisterMarkup(vtkMRMLMarkupsCurveNode::New(), vtkSlicerCurveWidget::New(), "C" );
+  this->RegisterMarkup(vtkMRMLMarkupsClosedCurveNode::New(), vtkSlicerClosedCurveWidget::New(), "C");
+  this->RegisterMarkup(vtkMRMLMarkupsPlaneNode::New(), vtkSlicerPlaneWidget::New(), "P");
 
   this->Helper = vtkMRMLMarkupsDisplayableManagerHelper::New();
   this->Helper->SetDisplayableManager(this);
   this->DisableInteractorStyleEventsProcessing = 0;
 
-  this->Focus.insert("vtkMRMLMarkupsNode");
+  //NOTE: I think this should be removed. Every node handled by
+  //this DM requires an associated widget
+  //this->Focus.insert("vtkMRMLMarkupsNode");
 
   // by default, this displayableManager handles a 2d view, so the SliceNode
   // must be set when it's assigned to a viewer
@@ -119,6 +121,14 @@ vtkMRMLMarkupsDisplayableManager::~vtkMRMLMarkupsDisplayableManager()
   this->Helper->Delete();
 
   this->SliceNode = nullptr;
+
+  for (const auto& item : this->MarkupsNodesWidgets)
+    {
+    item.second->Delete();
+    }
+
+  this->MarkupsNodesWidgets.clear();
+  this->MarkupsNames.clear();
 }
 
 //---------------------------------------------------------------------------
@@ -136,6 +146,21 @@ void vtkMRMLMarkupsDisplayableManager::PrintSelf(ostream& os, vtkIndent indent)
     {
     os << indent << "No slice node" << std::endl;
     }
+}
+
+//---------------------------------------------------------------------------
+void vtkMRMLMarkupsDisplayableManager::RegisterMarkup(vtkMRMLMarkupsNode* node,
+                                                      vtkSlicerMarkupsWidget* widget,
+                                                      const std::string& name)
+{
+  if (!node)
+    {
+    vtkErrorMacro("RegisterMarkup: invalid node.");
+    }
+
+  this->MarkupsNodesWidgets[node->GetClassName()] = widget;
+  this->MarkupsNames[node->GetClassName()] = name;
+  node->Delete();
 }
 
 //---------------------------------------------------------------------------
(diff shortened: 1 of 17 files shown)


Modules/Loadable/Markups/MRMLDM/vtkMRMLMarkupsDisplayableManager.cxx
```diff
@@ -561,51 +586,37 @@ bool vtkMRMLMarkupsDisplayableManager::IsCorrectDisplayableManager()
...
-    {
-    nodeName = "C";
-    }
-  else if (markupsNodeClassName == "vtkMRMLMarkupsClosedCurveNode")
-    {
-    nodeName = "C";
-    }
-  else if (markupsNodeClassName == "vtkMRMLMarkupsPlaneNode")
+
+  for (const auto &item : this->MarkupsNames)
```
pieper, line 604: Why not use find_if with a lambda?  I know the loop is sure to be short but I think it would be a good pattern.

Modules/Loadable/Markups/MRMLDM/vtkMRMLMarkupsDisplayableManager.cxx
```diff
@@ -88,18 +88,20 @@ vtkStandardNewMacro (vtkMRMLMarkupsDisplayableManager);
...
-  this->Focus.insert("vtkMRMLMarkupsLineNode");
-  this->Focus.insert("vtkMRMLMarkupsCurveNode");
-  this->Focus.insert("vtkMRMLMarkupsClosedCurveNode");
-  this->Focus.insert("vtkMRMLMarkupsPlaneNode");
+  this->RegisterMarkup(vtkMRMLMarkupsAngleNode::New(), vtkSlicerAngleWidget::New(), "A");
+  this->RegisterMarkup(vtkMRMLMarkupsFiducialNode::New(), vtkSlicerPointsWidget::New(), "F");
+  this->RegisterMarkup(vtkMRMLMarkupsLineNode::New(), vtkSlicerLineWidget::New(), "L");
+  this->RegisterMarkup(vtkMRMLMarkupsCurveNode::New(), vtkSlicerCurveWidget::New(), "C" );
+  this->RegisterMarkup(vtkMRMLMarkupsClosedCurveNode::New(), vtkSlicerClosedCurveWidget::New(), "C");
+  this->RegisterMarkup(vtkMRMLMarkupsPlaneNode::New(), vtkSlicerPlaneWidget::New(), "P");
```
pieper, line 96: To me it would make more sense for each markup type to register itself so that this class wouldn't presume that any of them are available.  That would simplify maintenance since all would follow the same pattern.

pieper (commented):
This looks good to me, modulo some comments below.

Modules/Loadable/Markups/MRMLDM/vtkMRMLMarkupsDisplayableManager.cxx
```diff
@@ -138,6 +148,21 @@ void vtkMRMLMarkupsDisplayableManager::PrintSelf(ostream& os, vtkIndent indent)
...
+void vtkMRMLMarkupsDisplayableManager::RegisterMarkup(vtkMRMLMarkupsNode* node,
+                                                      vtkSlicerMarkupsWidget* widget,
+                                                      const std::string& name)
+{
+  if (!node)
+    {
+    vtkErrorMacro("RegisterMarkup: invalid node.");
+    }
+
+  this->MarkupsNodesWidgets[node->GetClassName()] = widget;
```
lassoan, line 161: We may have multiple widgets for the same node class. For example, all Python scripted widgets will use the same vtkSlicerMarkupsScriptedWidget class. We may also add different widgets for the same markups node (for example, we may add slightly different widgets or widget representations for the same vtkMRMLMarkupsPlaneNode).

As a consequence, we cannot use the node class name as key in the MarkupsNodesWidgets map. Instead, we could specify a "Type" attribute in the widget class, which would return a string (e.g., "Points", "Plane", "ClosedCurve", "CurvedPlane", ...).

To make sure that the design is usable in Python, it would be important to implement a Python scripted widget (it can be a test in Slicer core or in your extension).

Modules/Loadable/Markups/MRMLDM/vtkMRMLMarkupsDisplayableManager.cxx
```diff
@@ -138,6 +148,21 @@ void vtkMRMLMarkupsDisplayableManager::PrintSelf(ostream& os, vtkIndent indent)
...
+  this->MarkupsNames[node->GetClassName()] = name;
+  node->Delete();
```
lassoan, line 163: this will not be needed (the caller takes care of this)

Modules/Loadable/Markups/MRMLDM/vtkMRMLMarkupsDisplayableManager.cxx
lassoan: Actually, this should be just a method call to the widget object

Modules/Loadable/Markups/MRMLDM/vtkMRMLMarkupsDisplayableManager.cxx
```diff
@@ -942,40 +953,24 @@ vtkSlicerMarkupsWidget * vtkMRMLMarkupsDisplayableManager::CreateWidget(vtkMRMLM
...
+      widget = (item.second)->CreateInstance();
+      break;
+      }
     }
-  else
+
+  // If the widget was successfully createdc
+  if (widget)
     {
-    vtkErrorMacro("vtkMRMLMarkupsDisplayableManager::CreateWidget failed: invalid display node class " << markupsNode->GetClassName());
```
lassoan, line 972: Keep the main success scenario in the main branch and use if branches for handling errors (and return if needed) - similarly as it was done before. It makes the code easier to read.

Modules/Loadable/Markups/MRMLDM/vtkMRMLMarkupsDisplayableManager.h
```diff
@@ -138,8 +145,12 @@ class  VTK_SLICER_MARKUPS_MODULE_MRMLDISPLAYABLEMANAGER_EXPORT vtkMRMLMarkupsDis
   /// \sa IsManageable(vtkMRMLNode*), IsCorrectDisplayableManager()
   virtual bool IsManageable(const char* nodeClassName);
 
-  /// Contains class names of markups nodes that this displayable manager can handle
-  std::set<std::string> Focus;
+  /// Contains types of markups nodes that this displayable manager can
+  /// handle as well as the associated widgets.
+  std::map<std::string, vtkSlicerMarkupsWidget*> MarkupsNodesWidgets;
```
lassoan, line 150: Include in the variable name what is mapped to what. For example MarkupsTypeToWidget

Modules/Loadable/Markups/MRMLDM/vtkMRMLMarkupsDisplayableManager.h
```diff
@@ -138,8 +145,12 @@ class  VTK_SLICER_MARKUPS_MODULE_MRMLDISPLAYABLEMANAGER_EXPORT vtkMRMLMarkupsDis
...
+
+  /// Relates types of markups and node naming strings
+  std::map<std::string, std::string> MarkupsNames;
```
lassoan, line 153: this should not be necessary (the widget can tell what should be the default node name prefix).

lassoan (commented):
Thanks a lot @RafaelPalomar, this will be great!

Generally it is good. The only major improvement to do is to ensure that the design is compatible with future introduction of Python scripted markup widgets/representations/nodes (that would allow users to easily create new widgets by slightly modifying behavior or representation of existing widgets).

It is important to make the very first design Python-ready because otherwise we would need to break all the custom markup nodes in all extensions when we introduce Python scripted markups later.

Using markup type to identify markups may require changes in the markups placement toolbar. This should not be a problem, because that toolbar button must be changed anyway. It is untenable that you cannot tell what happens when the user clicks in a view (if a new markup is created or a new control points is added to an existing markup). Probably we should remove the dropdown menu from the markups placement icon (keep it as a simple button) and move markups creation to a separate toolbar.

PR #5346, opened 2020-12-16
BUG: Enabling builtin ctypes in python

RafaelPalomar:
This fixes a segfault occurring in some linux distributions. Suspected
incompatibility between libffi>=3.4 and Slicer cpython, but not 100% confirmed. The proposed changes seem harmless anyways.

For discussion see [https://discourse.slicer.org/t/cant-start-latest-stable-on-ubuntu-20-04/14029/35](https://discourse.slicer.org/t/cant-start-latest-stable-on-ubuntu-20-04/14029/35)

### DIFF
--- a/SuperBuild/External_python.cmake
+++ b/SuperBuild/External_python.cmake
@@ -91,7 +91,7 @@ if((NOT DEFINED PYTHON_INCLUDE_DIR
       )
   elseif(UNIX)
     list(APPEND EXTERNAL_PROJECT_OPTIONAL_CMAKE_ARGS
-        #Fixes segfault in some linux distributions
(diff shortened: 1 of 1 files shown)


pieper (approved):
LGTM 👍

jcfr:
Thanks for the contribution. I will test and report back. This change may also be required so that we can notarize Slicer app on macos. I will report back.

lassoan:
I tested this on Windows and there does not seem to be any issue. It would be nice to merge this, as pull requests are piling up.

lassoan:
Any objection to merge this?

jcfr (approved):

Issue #123, opened 2020-03-12
Normalized Images are not renamed properly


slicerbot:
_This issue was created automatically from an original [Mantis Issue](https://mantisarchive.slicer.org/view.php?id=123). Further discussion may take place here._

PR #5347, opened 2020-12-16
ENH: Add DOC commit comment prefix to commit message check hook

lassoan:


commit-msg
```diff
@@ -91,10 +92,14 @@ Valid commit types are:
...
   COMP:  - a fix for a compilation issue, error or warning,
+  DOC:   - a change made to documentation files,
   ENH:   - new functionality added to the project,
   PERF:  - a performance improvement,
-  STYLE: - a change that does not impact the logic or execution of the code.
-  (improve coding style, comments, documentation).
+  STYLE: - a code change that does not impact the logic or execution of the code.
+  (improve coding style, comments, documentation in source code).
+
+Also consider adding reference to related issue by adding "re #123" (without quotes,
```
jcfr, line 101: Or `See #123`

jcfr, line 101:
```suggestion
Also consider adding reference to related issue by adding "re #123" (without quotes,
Also consider adding reference to related issue by adding "See #123" (without quotes,
```

jamesobutler:
Does there still need to be git hook check for commit message prefix when there is already the usage of the CommitCheck GitHub app on PRs?

cpinter:
I think it is useful to be able to catch this as early as possible. Then the contributor can fix it locally before pushing

lassoan:
Thank you, committed with the suggested changes.

PR #5349, opened 2020-12-17
ENH: Pluggable Markups

RafaelPalomar:
This PR makes the Markups system pluggable, this is, new Markups can be defined in modules other than the Markups module. This PR introduces the following changes:

- Markups can be registered through `vktSlicerMarkupsLogic`.
- MRML Markups nodes have been extended to hold attributes that previously were hard-coded in other components (i.e., name of the markup was hard-coded in the registration of the markup with the selection node) .
- `vtkMRMLMarkupsDisplayableManager` has been modified to use the markups registration factory.
- vtkSlicerMarkupsWidgets have been extended with a `CreateInstance()` function that acts as a factory method, thus allowing widget type resolution and creation in run-time.
-  Markup subject hierarchy plugin has been modified to use the markups registration.
- The markups defined within the Markups module have been registered using this new architecture.
- GUI has been modified to adapt dynamically to registered markups.

TODO:

- [x]  Modify the Markups module widgets to dynamically add/remove new markups creation buttons
- [x] Modify Markups module widgets to dynamically deal with additional widgets
- [x] Bug correction: Icon names for markups
- [x] Replace QString by const QString& in widget
- [x] Change class-based markups registration by name-based markups registration
- [x] Bring consistency to Markups naming (created from widget vs. created through selection node).
- [x] Order of markups buttons and white/blacklisting


Modules/Loadable/Markups/VTKWidgets/vtkMRMLMarkupsRegistrationFactory.h
```diff
@@ -0,0 +1,111 @@
...
+  static vtkMRMLMarkupsRegistrationFactory* GetInstance();
+
+  /// This is a singleton pattern New.  There will only be ONE
+  /// reference to a vtkMRMLMarkupsRegistrationFactory object per process.
+  /// Clients that call this must call Delete on the object so that the reference
+  /// counting will work. The single instance will be unreferenced when
+  /// the program exits.
+  static vtkMRMLMarkupsRegistrationFactory* New();
+
+  /// This reigsters a new markup.
```
jcfr, line 59:
```suggestion
/// This registers a new markup.
```

Modules/Loadable/Markups/VTKWidgets/vtkMRMLMarkupsRegistrationFactory.h
```diff
@@ -0,0 +1,111 @@
...
+  /// \param markupsNode vtkMRMLMarkups node to register.
+  /// \param markupsWidget vtkSlicerMarkupsWidget associated to the markup node.
+  void RegisterMarkup(vtkMRMLMarkupsNode *markupsNode,
+                      vtkSlicerMarkupsWidget*markupsWidget);
+
+  /// This returns an isntance to a corresponding vtkSlicerMarkupsWidget associated
```
jcfr, line 65:
```suggestion
  /// This returns an instance to a corresponding vtkSlicerMarkupsWidget associated
```

Modules/Loadable/Markups/VTKWidgets/vtkMRMLMarkupsRegistrationFactory.h
```diff
@@ -0,0 +1,111 @@
...
+  /// to the MRML node class specified.
+  /// \param nodeClass registered class to retrieve the associated widget.
+  /// \return pointer to associated vtkSLicerMarkupsWidget or nullptr if the MRML node
+  /// class is not registered.
+  vtkSlicerMarkupsWidget* GetWidgetByMarkupsNodeClass(const std::string& nodeClass) const;
+
+  /// This checks if the MRML node specified is registered.
+  /// \param nodeClass MRML node class name to check.
+  /// \return true, if the node is registered in the factory, fasle otherwise.
```
jcfr, line 74:
```suggestion
  /// \return true, if the node is registered in the factory, false otherwise.
```

Modules/Loadable/Markups/qSlicerMarkupsModule.cxx
```diff
@@ -119,6 +133,24 @@ void qSlicerMarkupsModule::setup()
 {
   this->Superclass::setup();
 
+   vtkMRMLMarkupsRegistrationFactory::GetInstance()->RegisterMarkup(
+     vtkMRMLMarkupsAngleNode::New(), vtkSlicerAngleWidget::New());
```
jcfr, line 137: Ideally, we should only register the widget and the method would be renamed from `RegisterMarkup` to `RegisterMarkupsWidget`

Similarly, the class `vtkMRMLMarkupsRegistrationFactory` would be renamed `vtkMRMLMarkupsWidgetFactory`

Modules/Loadable/Markups/MRMLDM/vtkMRMLMarkupsDisplayableManager.cxx
```diff
@@ -939,43 +916,23 @@ vtkSlicerMarkupsWidget * vtkMRMLMarkupsDisplayableManager::CreateWidget(vtkMRMLM
...
-  else if (vtkMRMLMarkupsPlaneNode::SafeDownCast(markupsNode))
-    {
-    widget = vtkSlicerPlaneWidget::New();
-    }
-  else
+  // Create a widget of the associated type if the node matches the registered nodes
+  vtkSlicerMarkupsWidget* widget =
+    vtkSlicerMarkupsWidget::SafeDownCast(
+      vtkMRMLMarkupsRegistrationFactory::GetInstance()->GetWidgetByMarkupsNodeClass(markupsNode->GetClassName())
+    )->CreateInstance();
```
jcfr, line 924:
```suggestion
  vtkSlicerMarkupsWidget* widget =
    vtkSlicerMarkupsWidget::SafeDownCast(
      this->GetMRMLApplicationLogic()->GetWidgetByMarkupsNodeClass(markupsNode->GetClassName())
    )->CreateInstance();
```

Modules/Loadable/Markups/MRMLDM/vtkMRMLMarkupsDisplayableManager.cxx
```diff
@@ -561,51 +553,36 @@ bool vtkMRMLMarkupsDisplayableManager::IsCorrectDisplayableManager()
 //---------------------------------------------------------------------------
 bool vtkMRMLMarkupsDisplayableManager::IsManageable(vtkMRMLNode* node)
 {
-  return (this->Focus.find(node->GetClassName()) != this->Focus.end());
+  if (node == nullptr)
+    {
+   vtkErrorMacro("Is Manageable: invalid node.");
```
cpinter, line 558: Indentation only has 3 spaces, four required.

Modules/Loadable/Markups/MRMLDM/vtkMRMLMarkupsDisplayableManager.cxx
```diff
@@ -561,51 +553,36 @@ bool vtkMRMLMarkupsDisplayableManager::IsCorrectDisplayableManager()
...
+  return vtkMRMLMarkupsRegistrationFactory::GetInstance()->IsRegistered(node->GetClassName());
 }
 
 //---------------------------------------------------------------------------
 bool vtkMRMLMarkupsDisplayableManager::IsManageable(const char* nodeClassName)
 {
-  return nodeClassName && (this->Focus.find(nodeClassName) != this->Focus.end());
+  if (nodeClassName == nullptr)
+    {
+   return false;
```
cpinter, line 569: Indentation

Modules/Loadable/Markups/SubjectHierarchyPlugins/qSlicerSubjectHierarchyMarkupsPlugin.cxx
```diff
@@ -33,6 +33,13 @@
 #include "qSlicerTerminologyItemDelegate.h"
 #include "vtkSlicerTerminologiesModuleLogic.h"
 
+#include "vtkMRMLMarkupsFiducialNode.h"
```
cpinter, line 36: Do you really need to include all the node classes?

Modules/Loadable/Markups/VTKWidgets/vtkMRMLMarkupsRegistrationFactory.h
```diff
@@ -0,0 +1,111 @@
...
+
+  // Singleton management functions.
+  static void classInitialize();
+  static void classFinalize();
+
+  friend class vtkMRMLMarkupsRegistrationFactoryInitialize;
+  typedef vtkMRMLMarkupsRegistrationFactory Self;
+
+  // Keeps track of the registered nodes and corresponding widgets
+  std::map<std::string, vtkSmartPointer<vtkSlicerMarkupsWidget>> MarkupsWidgetsMap;
```
cpinter, line 91: As I remember if you don't add a space between ">>" then Linux compilation will fail. So instead:

`std::map<std::string, vtkSmartPointer<vtkSlicerMarkupsWidget> > MarkupsWidgetsMap;`

lassoan: I think this is not an issue anymore in recent gcc versions.

Modules/Loadable/Markups/qSlicerMarkupsModule.cxx
RafaelPalomar: Thanks @jcfr . In our last meeting we decided to try to integrate the registration in the markups logic itself, but I will keep this idea of only registering the widget when I propose the new changes.

Modules/Loadable/Markups/MRMLDM/vtkMRMLMarkupsDisplayableManager.cxx
RafaelPalomar: Thanks for the tip @cpinter, you are right.

Modules/Loadable/Markups/VTKWidgets/vtkMRMLMarkupsRegistrationFactory.h
RafaelPalomar: As @lassoan points out this is not an issue in recent gcc versions. The question is whether it makes sense to have it still as `> >` to make it more backwards-compatible. Any opinions @lassoan @cpinter?

Modules/Loadable/Markups/MRMLDM/vtkMRMLMarkupsDisplayableManager.cxx
RafaelPalomar: @cpinter, which tools do you use to check for code styling?

Modules/Loadable/Markups/MRMLDM/vtkMRMLMarkupsDisplayableManager.cxx
jcfr: The advantage of adding the registration logic to `vtkMRMLApplicationLogic` is that we can:
* avoid using our own factory and rely on the fact there a single instance of the application
* support access to function related to widget registration from displayable manager (since they derive from `vtkMRMLAbstractLogic` having the  method [GetMRMLApplicationLogic](https://github.com/Slicer/Slicer/blob/87adcd2211d69826b5bdba8508b831ae4b919295/Libs/MRML/Logic/vtkMRMLAbstractLogic.h#L145-L146))

Using `vtkSlicerMarkupsLogic` could be confusing as we can have multiple instance of the logic

jcfr: >  forced to use abstract types for the registration

This could be addressed by moving `vtkMRMLMarkupsNode` and `vtkMRMLMarkupsJsonStorageNode` to `Libs/MRML/Core`  (`vtkMRMLMarkupsStorageNode` is already there)

Modules/Loadable/Markups/VTKWidgets/vtkMRMLMarkupsRegistrationFactory.h
jcfr: I would favor consistency and suggest to use  `> >`

Then, let's create an issue documenting the fact we need to update the entire code base replacing `> >` with `>>`

Modules/Loadable/Markups/SubjectHierarchyPlugins/qSlicerSubjectHierarchyMarkupsPlugin.cxx
RafaelPalomar: I'll check it out. Probably not. Thanks for the feedback!

Modules/Loadable/Markups/qSlicerMarkupsModule.cxx
RafaelPalomar: As I see it at this point, we need the node-widget association. The displayable manager will try to decide which widget to create, based on the display node. However, several markups are sharing a common display node. For disambiguation one can retrieve the `vtkMRMLMarkupsNode` from the `vtkMRMLDisplayNode` and check with the registered `vtkMRMLarkupsNode`s.

Alternatively, instead of keeping the association on the registration, I think we could keep the associated vtkMRMLMarkupsNode on the `vtkSlicerMarkupsWidget`. @jcfr, @lassoan let me know your thoughts on this.

RafaelPalomar:
@jcfr, @lassoan, @pieper This is ready now for a review.

Modules/Loadable/Markups/qSlicerMarkupsModuleWidget.cxx
```diff
@@ -626,68 +608,58 @@ void qSlicerMarkupsModuleWidgetPrivate::setPlaceModeEnabled(bool placeEnable)
...
       {
-      return "";
+      QPushButton *markupCreatePushButton = new QPushButton();
+      markupCreatePushButton->setIcon(QIcon(markupsNode->GetPlacementIcon()));
+      layout->addWidget(markupCreatePushButton,
+                        i / this->createMarkupsButtonsColumns,
+                        i % this->createMarkupsButtonsColumns);
+
+      QObject::connect(markupCreatePushButton, &QPushButton::clicked,
+                       q, [q, markupsNode] {q->onCreateMarkupByClass(markupsNode->GetClassName());});
```
jcfr, line 657: @lassoan @pieper Lambda use

RafaelPalomar:
@lassoan, @pieper, @jcfr I made the changes we discussed in the last meeting and resolved the merge conflicts. I think this is ready to be reviewed.

jcfr:
To follow up on discussion related to mapper.

In this example:
* clicking on button1 will display `data: data1`, 
* and clicking on button2 will only display `data: data2`

```python
mapper = qt.QSignalMapper()

button1 = qt.QPushButton("Button 1")
button2 = qt.QPushButton("Button 2")

def dosomething(data):
  print("data: %s" % data)

button1.connect("clicked()", mapper, "map()")
mapper.setMapping(button1, "data1")

button2.connect("clicked()", mapper, "map()")
mapper.setMapping(button2, "data2")

mapper.connect("mapped(const QString&)", dosomething)

button1.show()
button2.show()
```

RafaelPalomar:
> To follow up on discussion related to mapper.

Thanks @jcfr for the snippet. This PR uses now `QSignalMapper` instead of lambdas.

Modules/Loadable/Markups/Logic/vtkSlicerMarkupsLogic.cxx
```diff
@@ -1561,3 +1561,63 @@ vtkMRMLMarkupsJsonStorageNode* vtkSlicerMarkupsLogic::AddNewJsonStorageNodeForMa
...
+                    " is already registered.");
+    return;
+    }
+
+  this->MarkupsWidgetsMap[markupsNode->GetMarkupName()] = associatedWidget;
+  this->MarkupsNodesMap[markupsNode->GetMarkupName()] = node;
+}
+
+//----------------------------------------------------------------------------
+vtkSlicerMarkupsWidget* vtkSlicerMarkupsLogic::GetWidgetByMarkupsName(const char* markupName) const
```
lassoan, line 1602: Markups node already have a "Name", so it is a bit confusing what name you mean here.
It should be something like "Type" instead.

Modules/Loadable/Markups/Logic/vtkSlicerMarkupsLogic.cxx
```diff
@@ -1561,3 +1561,63 @@ vtkMRMLMarkupsJsonStorageNode* vtkSlicerMarkupsLogic::AddNewJsonStorageNodeForMa
...
+  if (search != this->MarkupsWidgetsMap.end())
+    {
+    return search->second;
+    }
+
+  return nullptr;
+}
+
+//----------------------------------------------------------------------------
+vtkMRMLMarkupsNode* vtkSlicerMarkupsLogic::GetNodeByMarkupsName(const char* markupName) const
```
lassoan, line 1614: type - see above
Rename of "name" to "type" (or class or something similar) should be done everywhere.

Modules/Loadable/Markups/Logic/vtkSlicerMarkupsLogic.cxx
```diff
@@ -1561,3 +1561,63 @@ vtkMRMLMarkupsJsonStorageNode* vtkSlicerMarkupsLogic::AddNewJsonStorageNodeForMa
...
+
+  // Check that the class is not already registered
+  if (this->MarkupsWidgetsMap.count(markupsNode->GetMarkupName()))
+    {
+    vtkWarningMacro("RegisterMarkup: Markups node " << markupsNode->GetMarkupName() <<
+                    " is already registered.");
+    return;
+    }
+
+  this->MarkupsWidgetsMap[markupsNode->GetMarkupName()] = associatedWidget;
```
lassoan, line 1597: Name of map variable should contain both what maps to what, something like MarkupsTypeToWidgetMap and MarkupsTypeToNodeMap. It would probably also better to map the markups type to a struct because you always need to add/remove widget and node together. It would be also easier to specify additional properties in the future (you would not need to create a new map for each additional property).

Modules/Loadable/Markups/Logic/vtkSlicerMarkupsLogic.h
```diff
@@ -243,6 +245,35 @@ class VTK_SLICER_MARKUPS_MODULE_LOGIC_EXPORT vtkSlicerMarkupsLogic :
...
+  /// \param markupsNode MRMLMarkups node to be registered.
+  /// \param markupsWidget vtkSlicerWidget associated to the MRMLMarkups node registered.
+  void SetMarkup(vtkMRMLMarkupsNode *markupsNode, vtkSlicerMarkupsWidget* markupsWidget);
+
+  /// This returns an isntance to a corresponding vtkSlicerMarkupsWidget associated
+  /// to the indicated markups name.
+  /// \param nodeClass registered class to retrieve the associated widget.
+  /// \return pointer to associated vtkSLicerMarkupsWidget or nullptr if the MRML node
+  /// class is not registered.
+  vtkSlicerMarkupsWidget* GetWidgetByMarkupsName(const char* markupName) const;
```
lassoan, line 260: GetMarkupsWidgetByType

Modules/Loadable/Markups/Logic/vtkSlicerMarkupsLogic.h
```diff
@@ -243,6 +245,35 @@ class VTK_SLICER_MARKUPS_MODULE_LOGIC_EXPORT vtkSlicerMarkupsLogic :
...
+
+  /// This returns an isntance to a corresponding vtkMRMLMarkupsNode associated
+  /// to the indicated markups name.
+  /// \param nodeClass registered class to retrieve the associated widget.
+  /// \return pointer to associated vtkSLicerMarkupsWidget or nullptr if the MRML node
+  /// class is not registered.
+  vtkMRMLMarkupsNode* GetNodeByMarkupsName(const char* markupName) const;
```
lassoan, line 267: GetMarkupsNodeByType

Modules/Loadable/Markups/Logic/vtkSlicerMarkupsLogic.h
```diff
@@ -243,6 +245,35 @@ class VTK_SLICER_MARKUPS_MODULE_LOGIC_EXPORT vtkSlicerMarkupsLogic :
...
+
+  /// This returns a constant reference to the markups-widgets association map.
+  const std::map<std::string, vtkSmartPointer<vtkSlicerMarkupsWidget> >& GetMarkupsWidgets()const
```
lassoan, line 270: It would be better to avoid such low-level access, unless it is unavoidable. I think it is not Python-wrappable either (could you check?).

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsAngleNode.h
```diff
@@ -63,6 +64,12 @@ class  VTK_SLICER_MARKUPS_MODULE_MRML_EXPORT vtkMRMLMarkupsAngleNode : public vt
   /// Get node XML tag name (like Volume, Model)
   const char* GetNodeTagName() override {return "MarkupsAngle";}
 
+  /// Get markup name
+  const char* GetMarkupName() override {return "Angle";};
+
+  /// Get markup short name
+  const char* GetMarkupShortName() override {return "A";};
```
lassoan, line 71: A name can be shortened for many reasons. It would be better to have a method name that tells more about its intent.
If it is for choosing node name prefix then the method could be `GetDefaultNodeNamePrefix`.

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsNode.h
```diff
@@ -131,6 +131,7 @@ class  VTK_SLICER_MARKUPS_MODULE_MRML_EXPORT vtkMRMLMarkupsNode : public vtkMRML
   void PrintSelf(ostream& os, vtkIndent indent) override;
 
   virtual const char* GetIcon() {return "";};
+  virtual const char* GetAddIcon() {return "";};
```
lassoan, line 134: What happens if the user does not override the method? Do we display some generic markup icon?

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsROINode.h
```diff
@@ -58,16 +58,24 @@ class  VTK_SLICER_MARKUPS_MODULE_MRML_EXPORT vtkMRMLMarkupsROINode : public vtkM
   /// Print out the node information to the output stream
   void PrintSelf(ostream& os, vtkIndent indent) override;
 
-  const char* GetIcon() override {return ":/Icons/MarkupsCurveMouseModePlace.png";}
+  const char* GetIcon() override {return ":/Icons/MarkupsROIMouseModePlace.png";}
+  const char* GetAddIcon() override {return ":/Icons/MarkupsROIMouseModePlaceAdd.png";}
```
lassoan, line 62: Why don't we need the place icon here as well? These names could be more clear:
- Icon: - that is shown in the subject hierarchy tree, etc.
- AddIcon: - icon with  +, to add a new markups node
- AddPointIcon or PlacePointIcon: - icon with arrow, to add a point

Modules/Loadable/Markups/MRMLDM/vtkMRMLMarkupsDisplayableManager.cxx
```diff
@@ -553,66 +546,75 @@ bool vtkMRMLMarkupsDisplayableManager::IsCorrectDisplayableManager()
...
   if (selectionNode->GetActivePlaceNodeClassName() == nullptr)
     {
     return false;
     }
-  // the purpose of the displayableManager is hardcoded
-  return this->IsManageable(selectionNode->GetActivePlaceNodeClassName());
 
+  vtkSmartPointer<vtkMRMLNode> node;
+  node.TakeReference(this->GetMRMLScene()->CreateNodeByClass(placeNodeClassName));
+  vtkMRMLMarkupsNode* markupsNode = vtkMRMLMarkupsNode::SafeDownCast(node);
```
lassoan, line 557: A single command would convey the intent a bit more clearly:
vtkSmartPointer<vtkMRMLNode> node = vtkSmartPointer<vtkMRMLNode>::Take(this->GetMRMLScene()->CreateNodeByClass(placeNodeClassName));

Modules/Loadable/Markups/VTKWidgets/vtkSlicerAngleWidget.cxx
```diff
@@ -38,6 +38,22 @@ vtkSlicerAngleWidget::vtkSlicerAngleWidget()
 //----------------------------------------------------------------------
 vtkSlicerAngleWidget::~vtkSlicerAngleWidget() = default;
 
+//----------------------------------------------------------------------
+vtkSlicerMarkupsWidget* vtkSlicerAngleWidget::CreateInstance() const
```
lassoan, line 42: I'm just curious, have you tried using vtkObjectFactory for this?
Since this code needs to be repeated in each class, it would be better to create a macro and/or template function so that it can be implemented by a single line of code (probably in the header). It would also allow us to change the instantiation at a single place, if a change is needed in the future.

Modules/Loadable/Markups/Widgets/qSlicerMarkupsAdditionalWidget.h
```diff
@@ -0,0 +1,51 @@
...
+
+// Qt Slicer includes
+#include "qSlicerWidget.h"
+
+// Markups Widgets includes
+#include "qSlicerMarkupsModuleWidgetsExport.h"
+
+class vtkMRMLMarkupsNode;
+
+class Q_SLICER_MODULE_MARKUPS_WIDGETS_EXPORT
```
lassoan, line 32: I don't understand the role of this class.
I would expect that there would be a MarkupsAdditionalOptionsWidget base class that the markups module widget would show whenever a suitable node type is selected.
These "additional options" widgets could be registered along with the markups node and widgetrepresentation, but probably it would be more flexible if they would be registered independently (because then each widget could return in its canManage() method if it is applicable to the selected node and the markups module widget would show/hide accordingly; or the widget could even show/hide/disable itself depending on what node is selected).

lassoan (commented):
This is great!
I have added comments mostly about name changes, to make things a bit more intuitive; and a few small design adjustments. But this is quite close to be ready.

Modules/Loadable/Markups/Logic/vtkSlicerMarkupsLogic.h
RafaelPalomar: Yes, this can be avoided. I've made the changes accordingly.

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsROINode.h
RafaelPalomar: @lassoan I can only see two different icons per markup in the `Markups/Resources/Icons` directory

lassoan: The basic icon type (without arrow or +) is in this folder: https://github.com/Slicer/Slicer/tree/master/Modules/Loadable/Markups/SubjectHierarchyPlugins/Resources/Icons

It would be better to define all 3 icon at the same place.

Modules/Loadable/Markups/VTKWidgets/vtkSlicerAngleWidget.cxx
RafaelPalomar: @lassoan I did try with `vtkObjectFactory` but it seems it does not operate on `vtkSlicerWidget`s. I have created a macro that is used in the widget's header file (single line).

Modules/Loadable/Markups/Widgets/qSlicerMarkupsAdditionalWidget.h
RafaelPalomar: @lassoan, this class plays the role you describe. `qSlicerMarkupsAdditionalWidget` is an abstract base class for `qSlicerMarkupsCurveSettingsWidget`, `qSlicerMarkupsAngleMeasurementsWidget`, `qMRMLMarkupsROIWidget` and potentially any markups-specific new widgets. These widgets only appear when the corresponding markup is selected.

Note that these additional widgets are not registered using `vtkSlicerMarkupsLogic` but `qSlicerMarkupsWidget`:
```cpp
qSlicerAbstractModuleRepresentation* qSlicerMarkupsModule::createWidgetRepresentation()
{
  // Create and configure module widget.
  qSlicerMarkupsModuleWidget* moduleWidget = new qSlicerMarkupsModuleWidget();

  // Register markups
  moduleWidget->setMarkup("Fiducial");
  moduleWidget->setMarkup("Line");
  moduleWidget->setMarkup("Angle", new qSlicerMarkupsAngleMeasurementsWidget);
  moduleWidget->setMarkup("Curve", new qSlicerMarkupsCurveSettingsWidget);
  moduleWidget->setMarkup("ClosedCurve");
  moduleWidget->setMarkup("Plane");
  moduleWidget->setMarkup("ROI", new qMRMLMarkupsROIWidget);

  // Set the number of columns for the grid of "add markups buttons" to the number of markups
  // regitered in this module.
  moduleWidget->setCreateMarkupsButtonsColumns(7);

  return moduleWidget;
}
```

Modules/Loadable/Markups/VTKWidgets/vtkSlicerAngleWidget.cxx
RafaelPalomar: I tried `vtkObjectFactory`, but it does not seem to operate in Markups VTKWidgets. I created a macro to have a 1-line declaration in the VTKWidgets header files as you suggest.

Modules/Loadable/Markups/Widgets/qSlicerMarkupsAdditionalWidget.h
lassoan: OK. To make things more clear, add documentation to this class, rename `setMarkup` to something that reflects what it actually does, and decouple the markup registration from the additional widget registration, because additional modules may want to register more widgets (without registering new markup types) and a widget may handle multiple markup types (e.g., new import/export or visualization options that can work with several markup types).

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsNode.h
RafaelPalomar: Here is what happens with a new markup with no icons specified:

screenshot

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsROINode.h
RafaelPalomar: Thanks, @lassoan. 

- I moved all icons to `Markups/Resources/Icons` 
- I renamed a couple to keep the naming consitency
- I added new `GenericMarkup` icons to be used by default (The M icon in the picture below)

Here is the result:
screenshot

jcfr (commented):
Some pending review I forgot to submit ...

Modules/Loadable/Markups/Logic/vtkSlicerMarkupsLogic.h
```diff
@@ -243,6 +245,35 @@ class VTK_SLICER_MARKUPS_MODULE_LOGIC_EXPORT vtkSlicerMarkupsLogic :
...
   vtkMRMLMarkupsJsonStorageNode* AddNewJsonStorageNodeForMarkupsType(std::string markupsType);
 
+  /// Registers a markup and its corresponding widget to be handled by the Markups module
+  /// For a markup to be handled by this module (processed by the displayable
+  /// manager, UI and subject hierarchy) it needs to be registered using this function.
+  /// \param markupsNode MRMLMarkups node to be registered.
+  /// \param markupsWidget vtkSlicerWidget associated to the MRMLMarkups node registered.
+  void SetMarkup(vtkMRMLMarkupsNode *markupsNode, vtkSlicerMarkupsWidget* markupsWidget);
+
+  /// This returns an isntance to a corresponding vtkSlicerMarkupsWidget associated
```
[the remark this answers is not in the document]
RafaelPalomar: Just pushed

RafaelPalomar:
@lassoan, @jcfr, @pieper this has now the changes requested by @lassoan in the review and correction to typos highlighted by @jcfr.

Yesterday, in the devs hangout, we discussed the possibility to create a singleton to register the `qSlicerMarkupsAdditionalOptionsWidget`s. The aim is to avoid exposing `qSlicerXXXModule` and `qSlicerXXXModuleWidget` across modules. I will explore how to do this and propose new changes. This PR is already functional and it is up to you if you want to merge it now (in which case I would make a new PR) or wait until new code to solve this issue is added to this PR. Let me know your thoughts.

lassoan:
@RafaelPalomar thanks a lot for all your hard work on this. I would wait for you to implement this factory class and then integrate it.

jcfr:
> implement this factory class and then integrate it.

I am looking at the code now, and will provide additional comment shortly.

jcfr:
@RafaelPalomar Do you know where the function `qSlicerMarkupsModuleWidget::setMarkup` is ? Or if it has been renamed or omitted during one of the recent rebase ?

jcfr (changes requested):
Here are few suggestions and discussion points.

Now, to move forward with the registration of markups, I suggest to add class called

```
qSlicerMarkupsFactory
```

with the following methods:

```
  Q_INVOKABLE static qSlicerMarkupsFactory* instance();

  /// Allows cleanup of the singleton at application exit
  static void setInstance(qSlicerMarkupsFactory* instance);
```

as well as the following in the implementation:

```cpp
qSlicerMarkupsFactory *qSlicerMarkupsFactory::m_Instance = nullptr;

class qSlicerMarkupsFactoryCleanup
{
public:
  inline void use()   {   }

  ~qSlicerMarkupsFactoryCleanup()
    {
    if (qSlicerMarkupsFactoryCleanup::m_Instance)
      {
      qSlicerMarkupsFactoryCleanup::setInstance(nullptr);
      }
    }
};
static qSlicerMarkupsFactoryCleanup qSlicerMarkupsFactoryCleanupGlobal;

....
```

Then,  in `qSlicerMarkupsModule::setup()` you would perform the relevant registration with calls like:

```cpp  qSlicerMarkupsFactory::instance()->registerMarkup(vtkNew<vtkMRMLMarkupsAngleNode>(), QList<QWidget*>() << new qSlicerMarkupsAngleMeasurementsWidget());
```

References:
* [qSlicerSegmentEditorEffectFactory.h](https://github.com/Slicer/Slicer/blob/master/Modules/Loadable/Segmentations/EditorEffects/qSlicerSegmentEditorEffectFactory.h) and [qSlicerSegmentEditorEffectFactory.cxx](https://github.com/Slicer/Slicer/blob/master/Modules/Loadable/Segmentations/EditorEffects/qSlicerSegmentEditorEffectFactory.cxx)
* [qSlicerSubjectHierarchyPluginHandler.h](https://github.com/Slicer/Slicer/blob/master/Modules/Loadable/SubjectHierarchy/Widgets/qSlicerSubjectHierarchyPluginHandler.h) and [qSlicerSubjectHierarchyPluginHandler.cxx](https://github.com/Slicer/Slicer/blob/master/Modules/Loadable/SubjectHierarchy/Widgets/qSlicerSubjectHierarchyPluginHandler.cxx)

RafaelPalomar:
> @RafaelPalomar Do you know where the function `qSlicerMarkupsModuleWidget::setMarkup` is ? Or if it has been renamed or omitted during one of the recent rebase ?

@jcfr `qSlicerMarkupsModuleWidget::setMarkups` was removed in favor of `qSlicerMarkupsModuleWidget::associatedMarkups()` that returns a `QStringList` you can use to stream your markup types in.

Modules/Loadable/Markups/qSlicerMarkupsModuleWidget.cxx
```diff
@@ -132,12 +131,21 @@ class qSlicerMarkupsModuleWidgetPrivate: public Ui_qSlicerMarkupsModule
...
   QPixmap     SlicerVisibleIcon;
   QPixmap     SlicerInvisibleIcon;
+
+  // Dynamic list of create markups push buttons
+  QList<QPushButton*> ceateMarkupsPushButtons;
+  unsigned int createMarkupsButtonsColumns;
+
+  // List of markups that will be handled by the UI
+  // NOTE: The order in this list determines the order of appearance of the buttons in the UI
+  QStringList associatedMarkups;
```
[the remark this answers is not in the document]
RafaelPalomar: @jcfr I observe that in `qSlicerMarkupsModuleWidgetPrivate` some of the member variable names are capitalized and others not. Should we capitalize them all?

Modules/Loadable/Markups/qSlicerMarkupsModule.cxx
```diff
@@ -146,9 +175,33 @@ void qSlicerMarkupsModule::setup()
...
 
 //-----------------------------------------------------------------------------
-qSlicerAbstractModuleRepresentation * qSlicerMarkupsModule::createWidgetRepresentation()
+qSlicerAbstractModuleRepresentation* qSlicerMarkupsModule::createWidgetRepresentation()
 {
-  return new qSlicerMarkupsModuleWidget;
+
+  // Create and configure the additional widgets
+  auto angleMeasurementsWidget = new qSlicerMarkupsAngleMeasurementsWidget();
+  angleMeasurementsWidget->associatedMarkups() << "Angle";
```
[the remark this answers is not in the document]
RafaelPalomar: @jcfr, @lassoan I totally agree with @jcfr, there are a lot of hard-coded strings. There are two details to have into account. First is that this form cannot be enforced (maybe?) in new markups as opposed to e.g., GetMarkupName() which is defined in the parent node already for all the child markups. Slicer is heavily based on passing `char *`, maybe until the code base moves to `std::string` it should be:
```
static  const char MARKUP_NAME[] = "Angle" 
```

Modules/Loadable/Markups/Logic/vtkSlicerMarkupsLogic.cxx
```diff
@@ -217,13 +217,21 @@ void vtkSlicerMarkupsLogic::ObserveMRMLScene()
...
+    vtkNew<vtkMRMLMarkupsLineNode> line;
+    vtkNew<vtkMRMLMarkupsAngleNode> angle;
+    vtkNew<vtkMRMLMarkupsCurveNode> curve;
+    vtkNew<vtkMRMLMarkupsClosedCurveNode> closedCurve;
+    vtkNew<vtkMRMLMarkupsPlaneNode> plane;
+    vtkNew<vtkMRMLMarkupsROINode> roi;
+
+    selectionNode->AddNewPlaceNodeClassNameToList(fiducial->GetClassName(), fiducial->GetAddIcon(), "Fiducial");
+    selectionNode->AddNewPlaceNodeClassNameToList(line->GetClassName(), line->GetAddIcon(), "Line");
+    selectionNode->AddNewPlaceNodeClassNameToList(angle->GetClassName(), angle->GetAddIcon(), "Angle");
```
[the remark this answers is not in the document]
RafaelPalomar: > I was also thinking we could simply the interface and add the function `vtkMRMLSelectionNode::AddNewPlaceNodeClassNameToList(vtkMRMLMarkupsNode*)` but `vtkMRMLMarkupsNode` is not available in `Libs/MRML/Core`

@jcfr, @lassoan, @pieper in December last year we discussed about the possibility of moving Markups core classes to `Libs/MRML/Core`. Though there were arguments in favor of this, keeping the MRML Core as simple as possible prevailed. I'll be happy to help on this if you decide to make a move in that direction sometime in the future.

RafaelPalomar:
> Now, to move forward with the registration of markups...

@lassoan, @pieper.  @jcfr has a good idea on how to implement this! There are a couple of points that need discussion:

1. We need to decouple the registration of markups and  the registration of optionalWidgets. In the suggestion by @jcfr 1 markup can be associated to n optionalWidgets. But also 1 optionalWidget could operate on n markups (e.g. CurveSettingsWidget operates on open and closed curves). I would suggest to use the factory to keep the two registries (markups and optionalWidgets).

2. When it comes to register associated types, widgets, etc. a common pattern is use of streams. Maybe we could use it like that.

My suggestion to address 1 and 2 would be something like this:

```cpp
  auto curveSettingsWidget = new qSlicerMarkupsCurveSettingsWidget();
  curveSettingsWidget->associatedMarkups() << "Curve" << "ClosedCurve";

qSlicerMarkupsFactory::instance()->associatedMarkupTypes() << "Curve" << "ClosedCurve" << "Fiducial";
qSlicerMarkupsFactory::instance()->additionalOptionsWidgets() << curveSettingsWidget;
```

lassoan:
> > Now, to move forward with the registration of markups...
> 
> @lassoan, @pieper. @jcfr has a good idea on how to implement this! There are a couple of points that need discussion:
> 
> 1. We need to decouple the registration of markups and  the registration of optionalWidgets. In the suggestion by @jcfr 1 markup can be associated to n optionalWidgets. But also 1 optionalWidget could operate on n markups (e.g. CurveSettingsWidget operates on open and closed curves). I would suggest to use the factory to keep the two registries (markups and optionalWidgets).
> 2. When it comes to register associated types, widgets, etc. a common pattern is use of streams. Maybe we could use it like that.
> 
> My suggestion to address 1 and 2 would be something like this:
> 
> ```c++
> auto curveSettingsWidget = new qSlicerMarkupsCurveSettingsWidget();
> curveSettingsWidget->associatedMarkups() << "Curve" << "ClosedCurve";
> 
> qSlicerMarkupsFactory::instance()->associatedMarkupTypes() << "Curve" << "ClosedCurve" << "Fiducial";
> qSlicerMarkupsFactory::instance()->additionalOptionsWidgets() << curveSettingsWidget;
> ```

I agree that 2 separate factories make sense, because it is not a lot of extra code or complexity and it is much more flexible.

> auto curveSettingsWidget = new qSlicerMarkupsCurveSettingsWidget();
> curveSettingsWidget->associatedMarkups() << "Curve" << "ClosedCurve";

We cannot enforce from outside to handle certain widget types. Only the widget knows which markups nodes it can handle and it may use complex logic to decide it (not just checking the markups type). The widget could provide a method that tells if it supports a particular markups node and if it does not support the currently selected markups node then the module widget could hide it. In the future, it would be also useful if it supported multi-select (e.g., for being able to add widget that combines markups or computes joint statistics or exports multiple nodes, ...)

> ```c++
> qSlicerMarkupsFactory::instance()->associatedMarkupTypes() << "Curve" << "ClosedCurve" << "Fiducial";
> ```

Since this is for MRML classes, it would be nice to have the factory in the logic. Since the module logic is already a factory, it could handle this, too (or it can be a separate class).

> qSlicerMarkupsFactory::instance()->additionalOptionsWidgets() << curveSettingsWidget;

I don't understand how this would work. How the factory learns about that additionalOptionsWidgets are changed? Why not use add/remove methods? (add method could take a list, if you really want to be able to pass a list, but also need to check what the Python wrapping supports)

Modules/Loadable/Markups/qSlicerMarkupsModule.cxx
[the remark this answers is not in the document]
lassoan: The widget should decide for itself what markups it supports.

Modules/Loadable/Markups/qSlicerMarkupsModuleWidget.cxx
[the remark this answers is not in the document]
lassoan: Yes, definitely be consistent within the class. Maybe have a look at a couple of similar classes and follow the convention that is used more.

RafaelPalomar:
@jcfr, @lassoan, @pieper this should be ready for a review.

pieper:
Hi @RafaelPalomar - I'm not seeing any test code; do you have a test that demonstrates what a plugin looks like from the external developer point of view?  We'll want that anyway to for regression testing but also it will be easier to review the new code in the context of such an example.

RafaelPalomar:
> Hi @RafaelPalomar - I'm not seeing any test code; do you have a test that demonstrates what a plugin looks like from the external developer point of view? We'll want that anyway to for regression testing but also it will be easier to review the new code in the context of such an example.
 
Hello @pieper - thanks for having a look at this! I'm currently creating new unit tests to validate the new functionality; I did not want to invest much time on testing until having a somehow definitive design. I have tried the new functionality with this module [https://github.com/ALive-research/Slicer-LiverAnalysis/tree/feature/liver-line-markup](https://github.com/ALive-research/Slicer-LiverAnalysis/tree/feature/liver-line-markup). We can use this code to create our regression tests.

Modules/Loadable/Markups/Testing/Python/PluggableMarkupsSelfTest.py
```diff
@@ -0,0 +1,128 @@
...
+        raise Exception("Create PushButton for " + markupNode.GetMarkupType() + "is present after unregistration")
+
+    logging.info('Processing completed')
+
+    self.delayDisplay('Test passed!')
+
+  def checkPushButtonExists(self, widget, name):
+
+    pushButtonName ='Create' + str(name) + 'PushButton '
+    self.delayDisplay('Checking' + pushButtonName + 'exists')
```
jcfr, line 125:
```suggestion
    self.delayDisplay("Checking '%s' exists" % pushButtonName)
```

Modules/Loadable/Markups/Testing/Python/PluggableMarkupsSelfTest.py
```diff
@@ -0,0 +1,128 @@
...
+      if self.checkPushButtonExists(markupsWidget, markupNode.GetMarkupType()) is None:
+        raise Exception("Create PushButton for " + markupNode.GetMarkupType() + "is not present")
+
+    markupsLogic = slicer.modules.markups.logic()
+    if markupsLogic is None:
+      raise Exception("Could't get the Markups module logic")
+
+    # Unregister Markups and check the buttons are gone
+    for markupNode in self.standardMarkupsNodes:
+      self.delayDisplay('Unregistering ' + markupNode.GetMarkupType())
```
jcfr, line 113:
```suggestion
      self.delayDisplay("Unregistering %s" % markupNode.GetMarkupType())
```

Modules/Loadable/Markups/Testing/Python/PluggableMarkupsSelfTest.py
```diff
@@ -0,0 +1,128 @@
...
+
+  def runTest(self):
+    self.setUp()
+    self.test_UnregisterDefaultMarkups()
+
+  def test_UnregisterDefaultMarkups(self):
+
+    self.delayDisplay("Starting the unregister default markups test")
+
+    # Move to markups module
```
jcfr, line 93:
```
    # Open markups module
```

Modules/Loadable/Markups/Testing/Python/PluggableMarkupsSelfTest.py
```diff
@@ -0,0 +1,128 @@
...
+
+  def setUp(self):
+    slicer.mrmlScene.Clear(0)
+    self.standardMarkupsNodes = [slicer.vtkMRMLMarkupsROINode(),
+                                 slicer.vtkMRMLMarkupsPlaneNode(),
+                                 slicer.vtkMRMLMarkupsClosedCurveNode(),
+                                 slicer.vtkMRMLMarkupsCurveNode(),
+                                 slicer.vtkMRMLMarkupsAngleNode(),
+                                 slicer.vtkMRMLMarkupsLineNode(),
+                                 slicer.vtkMRMLMarkupsFiducialNode()]
```
jcfr, line 83:
```suggestion
    self.standardMarkupsNodes = [
      slicer.vtkMRMLMarkupsAngleNode(),
      slicer.vtkMRMLMarkupsClosedCurveNode(),
      slicer.vtkMRMLMarkupsCurveNode(),
      slicer.vtkMRMLMarkupsFiducialNode()
      slicer.vtkMRMLMarkupsLineNode(),
      slicer.vtkMRMLMarkupsPlaneNode(),
      slicer.vtkMRMLMarkupsROINode(), 
      ]
```

jcfr: Alphabetical order

Modules/Loadable/Markups/Widgets/Testing/Cxx/CMakeLists.txt
```diff
@@ -4,17 +4,47 @@ set(KIT ${PROJECT_NAME})
 set(INPUT ${MRMLCore_SOURCE_DIR}/Testing)
 
 #-----------------------------------------------------------------------------
-set(KIT_TEST_SRCS
+set(TEST_SOURCES
```
jcfr, line 7: Variable `KIT_TEST_SRCS` is used across the code base

```bash
$ ack --cmake TEST_SOURCES -l | wc -l
2
$ ack --cmake KIT_TEST_SRCS -l | wc -l
35

```

Modules/Loadable/Markups/qSlicerMarkupsModuleWidget.cxx
```diff
@@ -1411,7 +1411,7 @@ void qSlicerMarkupsModuleWidget::onDeleteMarkupPushButtonClicked(bool confirm /*
     rows << row;
     }
   // sort the list
-  qSort(rows);
```
jcfr, line 1513: Could you address these in a separate pull request and review the remaining occurences as well:

```
Modules/Loadable/Annotations/Widgets/qMRMLAnnotationTreeView.cxx
390:  qSort(selected.begin(),selected.end());

Modules/Loadable/Markups/qSlicerMarkupsModuleWidget.cxx
1513:  qSort(rows);
2032:  qSort(rows);
2237:  qSort(rows);

Libs/MRML/Widgets/qMRMLTableModel.cxx
651:  qSort(mrmlIndexList.begin(), mrmlIndexList.end(), qGreater<int>());

```

RafaelPalomar: This is now in Slicer/Slicer#5557

RafaelPalomar:
@jcfr, @lassoan, @pieper this is ready for review. I have made the following changes:

- Unit testing for new classes
- Expanded testing for relevant existing classes
- A testing module showcasing the use of the pluggable markups
- Integration testing for  pluggable markups through a python self test.
- I also made minor modifications to solve bugs I found when creating the tests.

Modules/Loadable/Markups/VTKWidgets/vtkSlicerAngleWidget.h
```diff
@@ -43,9 +43,13 @@ class VTK_SLICER_MARKUPS_MODULE_VTKWIDGETS_EXPORT vtkSlicerAngleWidget : public
   /// Standard methods for a VTK class.
   vtkTypeMacro(vtkSlicerAngleWidget,vtkSlicerMarkupsWidget);
 
+  /// Create instance of the markups widget
+  vtkSlicerMarkupsWidgetCreateInstanceMacro(vtkSlicerAngleWidget);
+
   /// Create the default widget representation and initializes the widget and representation.
   void CreateDefaultRepresentation(vtkMRMLMarkupsDisplayNode* markupsDisplayNode, vtkMRMLAbstractViewNode* viewNode, vtkRenderer* renderer) override;
 
+
```
pieper, line 52: Nitpick: the extra empty line is a touch non-standard.

Modules/Loadable/Markups/VTKWidgets/vtkSlicerClosedCurveWidget.cxx
```diff
@@ -60,6 +60,7 @@ void vtkSlicerClosedCurveWidget::CreateDefaultRepresentation(
   rep->UpdateFromMRML(nullptr, 0); // full update
 }
 
+
```
pieper, line 63: Same as above, we try to have one blank line between methods.

Modules/Loadable/Markups/VTKWidgets/vtkSlicerCurveWidget.cxx
```diff
@@ -67,6 +67,7 @@ void vtkSlicerCurveWidget::CreateDefaultRepresentation(
   rep->UpdateFromMRML(nullptr, 0); // full update
 }
 
+
```
pieper, line 70: Here too

Modules/Loadable/Markups/VTKWidgets/vtkSlicerMarkupsWidget.h
```diff
@@ -34,6 +34,7 @@
 
 #include "vtkMRMLMarkupsNode.h"
 
+//----------------------------------------------------------------------
```
pieper, line 37: Another nitpick: we don't usually have these.

Modules/Loadable/Markups/VTKWidgets/vtkSlicerPointsWidget.cxx
```diff
@@ -50,3 +50,4 @@ void vtkSlicerPointsWidget::CreateDefaultRepresentation(
   rep->SetMarkupsDisplayNode(markupsDisplayNode);
   rep->UpdateFromMRML(nullptr, 0); // full update
 }
+
```
pieper, line 53: Nitpick: empty line

pieper (dismissed):
I didn't read every single line but overall it looks really good 👍 

Just a few formatting nitpicks but otherwise I think it looks good.  Let's see what @Sunderlandkyl thinks when he tries using it.

Sunderlandkyl:
Looks good!

I just have some comments about `qSlicerMarkupsAdditionalOptionsWidget`.

- Add a `mrmlMarkupsNode()` getter so that we don't need to inherit `qSlicerMarkupsAdditionalOptionsWidgetPrivate` in the private classes of derived widget to get the Markups node.
- Change `setMRMLMarkupsNode()` to be a slot, and add a `setMRMLNode(vtkMRMLNode*)` slot so that we can attach `currentNodeChanged(vtkMRMLNode*)` signals if needed.
- We should remove or update the various `setMRMLMarkupsNode()/setCurrentNode()/setMRMLMarkupsROINode()/etc.` functions in the derived classes to use the base class setter/getter.

Also, TestLine is currently added to the vtkMRMLSelectionNode as a placeable node type.

RafaelPalomar:
@jcfr, @lassoan, @pieper I believe I have addressed all the comments from @Sunderlandkyl

jcfr:
:tada: :arrow_right:  :champagne:

jcfr (approved):

PR #5350, opened 2020-12-17
ENH: Allow dropping SH folders into views

cpinter:
As an addition to recent work about dropping SH items into views (https://github.com/Slicer/Slicer/pull/5327), this commit adds the ability to drag&drop folders as well. In that case all the child items (recursively) of the folder are shown in the view the folder has been dropped. See alsohttps://github.com/NA-MIC/ProjectWeek/blob/master/PW34_2020_Virtual/Projects/SubjectHierarchyFolders/README.md

### DIFF
--- a/Modules/Loadable/SubjectHierarchy/Widgets/qSlicerSubjectHierarchyFolderPlugin.cxx
+++ b/Modules/Loadable/SubjectHierarchy/Widgets/qSlicerSubjectHierarchyFolderPlugin.cxx
@@ -648,3 +648,25 @@ void qSlicerSubjectHierarchyFolderPlugin::setApplyColorToBranchEnabledForItem(vt
 
   folderDisplayNode->SetApplyDisplayPropertiesOnBranch(enabled);
 }
+
+//-----------------------------------------------------------------------------
+bool qSlicerSubjectHierarchyFolderPlugin::showItemInView(vtkIdType itemID, vtkMRMLAbstractViewNode* viewNode, vtkIdList* allItemsToShow)
+{
+  vtkMRMLSubjectHierarchyNode* shNode = qSlicerSubjectHierarchyPluginHandler::instance()->subjectHierarchyNode();
+  if (!shNode)
+    {
+    qCritical() << Q_FUNC_INFO << ": Failed to access subject hierarchy node";
+    return false;
+    }
+
+  std::vector<vtkIdType> folderContentItemIDs;
+  shNode->GetItemChildren(itemID, folderContentItemIDs, true);
+  for (auto childItem : folderContentItemIDs)
+    {
+    qSlicerSubjectHierarchyAbstractPlugin* ownerPlugin = qSlicerSubjectHierarchyPluginHandler::instance()->getOwnerPluginForSubjectHierarchyItem(childItem);
+    if (ownerPlugin)
+      {
+      ownerPlugin->showItemInView(childItem, viewNode, allItemsToShow);
+      }
+    }
+}
--- a/Modules/Loadable/SubjectHierarchy/Widgets/qSlicerSubjectHierarchyFolderPlugin.h
+++ b/Modules/Loadable/SubjectHierarchy/Widgets/qSlicerSubjectHierarchyFolderPlugin.h
@@ -137,6 +137,9 @@ public:
   /// \param itemID Subject Hierarchy item to show the visibility context menu items for
   void showVisibilityContextMenuActionsForItem(vtkIdType itemID) override;
 
+  /// Show the contents of the colder in a selected view.
+  bool showItemInView(vtkIdType itemID, vtkMRMLAbstractViewNode* viewNode, vtkIdList* allItemsToShow) override;
+
 public:
   /// Create folder under specified item
   /// \param parentNode Parent item for folder to create


pieper (approved):
LGTM 👍 

This will be great as a start.  There are some other things we should consider, like adding tags or classes to views that would act similarly to the ViewGroup concept, so that when you drag a folder to a view it would be shown in all views with a matching tag (e.g. so you could show axi/sag/cor with a single drop).  ViewGroup is already used to control things like crosshair mode, which you would want to control independently from display behavior, but view group is only a single value currently.  We could use the existing mrml node attributes for this tagging concept.

cpinter:
I agree. As we also started discussing in the original "SH drop into views" PR, it would be great if the 3-orientation groups would behave as one place to show one data. Then I think we should at last link those views by default.

Modules/Loadable/SubjectHierarchy/Widgets/qSlicerSubjectHierarchyFolderPlugin.cxx
```diff
@@ -648,3 +648,25 @@ void qSlicerSubjectHierarchyFolderPlugin::setApplyColorToBranchEnabledForItem(vt
...
+    }
+
+  std::vector<vtkIdType> folderContentItemIDs;
+  shNode->GetItemChildren(itemID, folderContentItemIDs, true);
+  for (auto childItem : folderContentItemIDs)
+    {
+    qSlicerSubjectHierarchyAbstractPlugin* ownerPlugin = qSlicerSubjectHierarchyPluginHandler::instance()->getOwnerPluginForSubjectHierarchyItem(childItem);
+    if (ownerPlugin)
+      {
+      ownerPlugin->showItemInView(childItem, viewNode, allItemsToShow);
```
lassoan, line 669: allItemsToShow must contain all the items that will be shown in this operation (i.e., all children). This is important because for example if the owner plugin thinks that only one volume will be shown then it will put it into the background. However, if you tell the plugin that multiple volumes will be shown then it will place them in different layers (foreground, background).

It would be costly to assemble a full list of items here in the folder plugin (and in all other plugins). Instead, we should assemble the allItemsToShow list once, by recursively adding all item IDs in qSlicerSubjectHierarchyPluginHandler::showItemsInView.

qSlicerSubjectHierarchyPluginHandler::showItemsInView should probably still call showItemInView for only those items that the user actually selected (itemIDsToShow), because it would allow custom folder plugins to implement custom strategies to show items in that folder.

lassoan (commented):
Thank you, this will be useful! I have just one comment, should be easy to fix.

lassoan:
> There are some other things we should consider, like adding tags or classes to views that would act similarly to the ViewGroup concept, so that when you drag a folder to a view it would be shown in all views with a matching tag (e.g. so you could show axi/sag/cor with a single drop). ViewGroup is already used to control things like crosshair mode, which you would want to control independently from display behavior, but view group is only a single value currently. We could use the existing mrml node attributes for this tagging concept.

This is already implemented for volumes. You can enable/disable it by linking/unlinking views. If views are linked then drag-and-dropping a volume into a slice view puts that volume into all slice views in that group. (Since volume rendering can be computationally expensive, I was not sure if we wanted to add it to 3D views as well.)

This could be implemented similarly for all other node types. Visibility of all non-volume displayable nodes are controlled the same way, so we would need to implement a single mechanism and that would take care of all nodes.

Modules/Loadable/SubjectHierarchy/Widgets/qSlicerSubjectHierarchyFolderPlugin.cxx
cpinter: Good point! I'll do it right now in a follow-up PR.

cpinter: Just one thing: I don't understand the "it would be costly to assemble a full list of items here in the folder plugin" part.

You mean that in case more than one folders are selected and dropped, the children of those should be added in `qSlicerSubjectHierarchyPluginHandler::showItemsInView`? But if we do that then it would not make much sense what I did in the Folders plugin because if we handle folders in the plugin handler, then we could handle not just the allItemsToShow, but the whole dropping part...

lassoan: showItemInView method needs a full list of all items, including all children items of all selected folder items, recursively. This operation is costly, especially when we want to show a model hierarchy containing hundreds of items; and it would need to be performed hundreds of times (once for each shown item).

> You mean that in case more than one folders are selected and dropped, the children of those should be added in qSlicerSubjectHierarchyPluginHandler::showItemsInView?

Even when a single folder is dropped, you would determine all the children items in advance (and pass it on via allItemIds list), so that the owner plugins do not need to look up recursively all the other items again and again. It would be still the responsibility of the folder to call showItemInView for its children (and the folder plugin, especially custom folder plugins could decide how they actually show the content of that folder).

There is certainly a potential issue that a folder plugin could decide not to show some items in its subfolder, while the top-level logic would assume that all children are shown. However, I think the only way to resolve this would be to do the show operation in two phases: 1. ask folder plugins about what they would display and create the allItemIds list based on this, 2. call showItemInView using this allItemIds. It is doable, but a bit more complicated, so I would only implement it if we find that the assumption that all children get displayed is often violated and is causing issues.

cpinter: > This operation is costly, especially when we want to show a model hierarchy containing hundreds of items; and it would need to be performed hundreds of times (once for each shown item).

First of all, I don't understand this part. showItemInView will only run once for each folder, won't it?

Regardless of this, I implemented the code that collects all child items recursively. I did it in two steps, first added all the child items in a set (it is possible to select the folder as well as its contents in which case these items would be added twice or more if there are more levels), then added the unique item IDs in a new vtkIdList.

However, it occurred to me that not only folders can have children, but anything (like the branch curves after branch splitting in VMTK, or the beams under the plan in RT studies). In this case the children of folders would be shown, but not the children of other items.
What do you think about this? Should I instead expand the tree in the plugin handler and call showItemInView for all the children from there and remove the function from the folder plugin?

lassoan: >  it occurred to me that not only folders can have children, but anything 

Very good point. This would make it quite complicated to assign responsibility of showing children items to each parent item.

I would vote for making qSlicerSubjectHierarchyPluginHandler::showItemsInView gather all the children item IDs recurseively and call ownerPlugin->showItemInView for each, because it seems the simplest implementation. If we later find that this is not sufficient then we can still make the design more sophisticated.

cpinter: I completely agree. Change coming up soon.

cpinter: See https://github.com/Slicer/Slicer/pull/5353

PR #5352, opened 2020-12-18
ENH: Add option to automatically release resources when volume rendering is hidden

lassoan:
When rendering large volumes, it may be necessary to unload hidden volume from GPU memory.
See also: https://discourse.slicer.org/t/setting-visibility-to-off-in-volume-rendering-does-not-unload-the-volume-from-texture-memory/15127

Solution: Added "Auto-release resources" checkbox (in Volume Rendering module advanced settings). If it is checked then volume that do not have visible volume rendering are automatically unloaded from GPU memory.


cpinter:
This is a useful feature and the implementation looks good to me. Thank you @lassoan !

cpinter (approved):

lassoan:
Thanks for the review!

PR #5353, opened 2020-12-20
ENH: Show child items of any dropped SH item in view

cpinter:
As discussed in the PR https://github.com/Slicer/Slicer/pull/5350 it may be confusing if only the folders would show their children. Also the plugin handler needs to collect all the shown child items so that the individual plugins have the complete item list to best show the items (e.g. volumes in background and foreground). So instead of the folder plugin managing its children when dropping it, the plugin handler collects and shows all the dropped items and their children, recursively.


lassoan (approved):
Looks good, thank you!

PR #5362, opened 2021-01-01
DOC: Remove outdated macOS build instructions

jamesobutler:
This cleans up the macOS build instructions document which was full of old OS X version information that is no longer needed in the Slicer latest macOS build instructions page.  Slicer latest has a minimum version of macOS 10.13 and now primarily uses Qt 5.15 as the preferred Qt version.  Old platform information can always be viewed again in older versions of the documentation. The latest documentation should correspond to the current latest code.

Note that many of the "Common Errors" sections have been removed as they were for outdated platforms, or greatly outdated builds including Qt4, or the issue was simply fixed and no longer a "Common Error".

I'm not a regular macOS user, but I know these instructions were incorrect and could use a cleanup. @pieper Your knowledge on building Slicer with this platform will be helpful for improving this document.

### DIFF
--- a/Docs/developer_guide/build_instructions/macos.md
+++ b/Docs/developer_guide/build_instructions/macos.md
@@ -8,98 +8,12 @@ The prerequisites listed below are required to be able to configure/build/packag
 ```
 xcode-select --install
 ```
-- El Capitan is what most developers use.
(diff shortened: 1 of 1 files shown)


pieper (approved):
Thanks for the cleanup! 👍

PR #5363, opened 2021-01-01
BUG: Fix Qt printer support for Slicer installations

Sunderlandkyl:
For installed versions of Slicer, calling QPrinterInfo::availablePrinterNames() would return an empty list, while built versions would have the list of printers populated as normal.
The lack of available printers made it impossible to set the output of QPrinter to NativeFormat.

By bundling the printsupport plugin with the installer, the available printers can be found.

### DIFF
--- a/CMake/SlicerCPack.cmake
+++ b/CMake/SlicerCPack.cmake
@@ -31,6 +31,7 @@ if(NOT Slicer_USE_SYSTEM_QT)
   set(SlicerBlockInstallQtPlugins_subdirectories
     audio
     imageformats
+    printsupport
     sqldrivers
     )
     if(Slicer_BUILD_WEBENGINE_SUPPORT)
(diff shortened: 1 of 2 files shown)


lassoan (approved):
Looks good to me. The printsupport package is available with LGPL license, so it should be no problem to be included in the package (https://doc.qt.io/qt-5/qtprintsupport-index.html#:~:text=The%20Qt%20Print%20Support%20module%20is%20available%20under,version%202.%20See%20Qt%20Licensing%20for%20further%20details.)

jcfr:
@timthirion @sjh26 this should be helpful

lassoan:
Yes, printing works very well with Qt. It is also surprisingly easy to print html to pdf just by using core Qt classes (see LungCTAnalyzer extension).

PR #5364, opened 2021-01-02
ENH: Remove unused logos from Slicer resource

jamesobutler:
This is some cleanup to remove unused logo files as I was reviewing Slicer logo usage in the source code. These appear to be primarily from ported resources from Slicer3 days and were maybe used in some old modules that are no longer included in Slicer. These files of course can still be accessed by checkout the source code of older versions, but with this PR merged they will no longer be in the latest Slicer source code.


pieper (approved):
Yes, cleaning this up makes sense 👍

PR #5366, opened 2021-01-03
ENH: New Slicer logo

jamesobutler:
This PR updates to the new Slicer logo originally brainstormed and thoroughly discussed in https://discourse.slicer.org/t/proposed-new-slicer-logo/11295.

These images files originated from the google folder link on this post: https://discourse.slicer.org/t/proposed-new-slicer-logo/11295/97. 
I will also upload a zip file here in case that link expires for some reason: [SlicerNewLogo.zip](https://github.com/Slicer/Slicer/files/5762176/SlicerNewLogo.zip)

Windows Desktop:
image
Windows Task Bar:
image
Windows Explorer:
image
image
Slicer Spash Screen (includes minor change of http to https):
image
Slicer Welcome:

| Light  Slicer | Dark Slicer |
|--------------|-------------|
|image|image|

Slicer About Dialog:

| Light  Slicer | Dark Slicer |
|--------------|-------------|
|image|image|


lassoan (approved):
These look great, thanks for working on it.
I think we should release a new stable release now and merge this change after that (in the hope of next stable release will be Slicer 5).

cpinter:
Thanks for working on this! It would be great if the discourse icon could be changed around the same time.

jamesobutler:
I've detailed all the places to update the Slicer logo in https://github.com/Slicer/Slicer/issues/5367 including the discourse icon.

jcfr:
Thanks @jamesobutler :pray: 

>  think we should release a new stable release now and merge this change after that 

Makes sense, I am adding this a topic for tomorrow hangout.

jamesobutler:
Following up on this PR, is the current plan to finalize this branding change just before the next stable release? Or only once confirmed that the next stable will be Slicer 5 instead of a Slicer 4.11.YYYYMMDD release?

jcfr:
Thanks for the follow up.

Here is a summary of what we discussed:
* Move forward with updating readthedocs, website, etc ... independently of Slicer 5.x release
* Switch logo in the application just before releasing Slicer 5.x

@sjh26 @lassoan @pieper : Did I miss anything ?

lassoan:
Since we started to replace Slicer logo in most places, I think it is time to replace the logo in the application, too. Any objections?

jamesobutler:
If there is going to be another snapshot release (4.14?), then it seems best to hold off on this as this should probably coincide with the Slicer 5.0 release marking a visible milestone. Slicer.org's new design probably should've rolled out at the 5.0.0 release as well. https://download.slicer.org/ will need to be updated to the new download backend platform and have the new logo to match and is tracked as part of https://github.com/Slicer/Slicer/issues/5367.

Slicer 5.0 is seemingly turning into a less of a major jump to something new because the previous thought 4.10 ->5.0 jump has had 4.11.20200926 and 4.11.20210226 and those have been listed as stable releases on the website which are close to what 5.0 will end up being. Maybe 5.0 will be most noticeable in the actual removal of old and outdated modules like Editor.

Side Note:
It would be nice if the [5.0.0 milestone](https://github.com/Slicer/Slicer/milestone/4) was filled out with the things that need to be completed before the 5.0.0 release to help with some prioritization. I'm really not sure how it is related to the [4.13.0 milestone](https://github.com/Slicer/Slicer/milestone/1)? My observation of how the milestones are currently used is that it never really seems like progress is made towards a release as more things just get added. Not sure how best to address this. It should at least match some of the Month times mentioned in the [roadmap](https://github.com/Slicer/Slicer/wiki/Roadmap).

lassoan:
> it seems best to hold off on this as this should probably coincide with the Slicer 5.0 release marking a visible milestone

I agree that ideally the logo should be swapped out everywhere at once, at Slicer5 release day. However, that would require concerted effort from many people at the same time - when also Slicer5 is released. Quite likely things would break, because it is just so hard to do many changes at once.

Maybe we should hold off the logo update in the application until the next interim release (that we'll do right after the extension manager, VMTK, and SlicerVirtualReality extensions are fixed)?

Slicer-4.13.0 milestone content is the intended scope for the next interim release. What is already clear that will not fit in this milestone are in Slicer-5.0.0, but I would expect that many other of the 91 items that are currently in Slicer-4.13.0 will be moved to Slicer-5.0.0. We will review these in the coming weeks, before the interim release.

lassoan:
Superseded by https://github.com/Slicer/Slicer/pull/5877

PR #5369, opened 2021-01-05
COMP: Update vtkMRMLReadXMLStdStringVectorMacro to fix VTK9 build error

jcfr:
This commit updates the macro to use "std::string" because "vtksys::String"
is not available when building against VTK 9.

It fixes a regression introduced in commit 1e0b70925 (BUG: Fixed MRML
string vector attribute writing)


lassoan (approved):

PR #5370, opened 2021-01-06
Fix extensions tests

lassoan:
Make extension description comparison insensitive to line ending.
Make SVN optional.
Fix expected order of build artifacts submission.

### DIFF
--- a/Extensions/CMake/SlicerFunctionGenerateExtensionDescription.cmake
+++ b/Extensions/CMake/SlicerFunctionGenerateExtensionDescription.cmake
@@ -153,7 +153,7 @@ function(slicer_generate_extension_description_test)
     EXTENSION_STATUS "Development"
     )
   execute_process(
-    COMMAND ${CMAKE_COMMAND} -E compare_files
+    COMMAND ${CMAKE_COMMAND} -E compare_files --ignore-eol
       ${CMAKE_CURRENT_BINARY_DIR}/SlicerToKiwiExporter.s4ext
       ${Slicer_SOURCE_DIR}/Extensions/CMake/Testing/extension_description_without_depends.s4ext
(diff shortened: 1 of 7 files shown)


Extensions/CMake/SlicerFunctionGenerateExtensionDescription.cmake
```diff
@@ -193,7 +193,7 @@ function(slicer_generate_extension_description_test)
     EXTENSION_STATUS ""
     )
   execute_process(
-    COMMAND ${CMAKE_COMMAND} -E compare_files
+    COMMAND ${CMAKE_COMMAND} -E compare_files --ignore-eol
```
jcfr, line 196: Is this option available with the minimum required version of CMake associated with Slicer ?

lassoan: I assumed that it is, I have not checked it.

jcfr:
If it is not the case, let's just exclude the test or add a comment.

Otherwise, looks good.

jcfr:
As a follow up, I will push a branch completely removing svn support ... I have it as a wip branch on my workstation

jcfr (dismissed):

lassoan:
ignore-eol option was introduced in [CMake-3.14](https://cmake.org/cmake/help/v3.14/manual/cmake.1.html#run-a-command-line-tool)

Required CMake version:
- Windows -> documentation states 3.15.1
- Linux, Mac -> https://github.com/Slicer/Slicer/blob/master/CMakeLists.txt#L1 -> 3.13.4

Is Slicer can be built using CMake 3.13?

lassoan:
@jcfr Thanks for the review. I've added the CMake version check and merged.

PR #5376, opened 2021-01-08
ENH: exposing module logic in vtk mrml application logic

RafaelPalomar:
The purpose of this PR is to frame the discussion around improving the exposure of `vtkSlicerModuleLogics` to other Slicer components (e.g., displayable managers) through `vtkMRMLApplicationLogic`. A discussion on this can be found in [https://discourse.slicer.org/t/improving-the-exposure-of-vtkslicermodulelogic-s-to-other-components/15375](https://discourse.slicer.org/t/improving-the-exposure-of-vtkslicermodulelogic-s-to-other-components/15375)

The following changes are proposed:

- Add module logic registration/unregistration methods, as well as query methods in `vtkMRMLApplicationLogic`. Unit testing has been added.
- Use registration/unregistration methods in `qSlicerModuleFactoryManager` for automatic registration of module logics on registration/unregistration of modules.
- `qSlicerWidget` has been extended to hold a pointer to the application logic. In this way, `qSlicerWidgets` can have access to all modules logics.
- Different modules have been undergoing a minor refactoring to exploit this new mechanics in favor to the former way. Note that this does not mean any change in the old API (it is till available and unchanged), but the new code is shorter and more tidy.

### DIFF
--- a/Base/QTGUI/qSlicerWidget.cxx
+++ b/Base/QTGUI/qSlicerWidget.cxx
@@ -23,28 +23,42 @@
 
 #include "qSlicerWidget.h"
 
+// MRML includes
+#include <vtkSlicerApplicationLogic.h>
+
 // VTK includes
 
 //-----------------------------------------------------------------------------
 class qSlicerWidgetPrivate
 {
+  Q_DECLARE_PUBLIC(qSlicerWidget);
+protected:
+  qSlicerWidget* const q_ptr;
+
 public:
+  qSlicerWidgetPrivate(qSlicerWidget& object);
+
   QPointer<QWidget>                          ParentContainer;
+  vtkSlicerApplicationLogic*                 AppLogic;
 };
+
+//-----------------------------------------------------------------------------
+qSlicerWidgetPrivate::qSlicerWidgetPrivate(qSlicerWidget& object)
+  : q_ptr(&object)
+{
+  this->AppLogic = nullptr;
+}
 
 //-----------------------------------------------------------------------------
 qSlicerWidget::qSlicerWidget(QWidget * _parent, Qt::WindowFlags f)
   :QWidget(_parent, f)
-  , d_ptr(new qSlicerWidgetPrivate)
+  , d_ptr(new qSlicerWidgetPrivate(*this))
 {
 }
 
 //-----------------------------------------------------------------------------
 qSlicerWidget::~qSlicerWidget() = default;
 
-//-----------------------------------------------------------------------------
-//CTK_SET_CPP(qSlicerWidget, vtkSlicerApplicationLogic*, setAppLogic, AppLogic);
-//CTK_GET_CPP(qSlicerWidget, vtkSlicerApplicationLogic*, appLogic, AppLogic);
 //-----------------------------------------------------------------------------
 
 void qSlicerWidget::setMRMLScene(vtkMRMLScene* scene)
@@ -55,4 +69,18 @@ void qSlicerWidget::setMRMLScene(vtkMRMLScene* scene)
     {
     emit mrmlSceneChanged(scene);
     }
+}
+
+//-----------------------------------------------------------------------------
+void qSlicerWidget::setAppLogic(vtkSlicerApplicationLogic* logic)
+{
+  Q_D(qSlicerWidget);
+  d->AppLogic = logic;
+}
+
+//-----------------------------------------------------------------------------
+vtkSlicerApplicationLogic* qSlicerWidget::appLogic()const
+{
+  Q_D(const qSlicerWidget);
+  return d->AppLogic;
 }
--- a/Libs/MRML/Logic/Testing/Cxx/vtkMRMLApplicationLogicTest1.cxx
+++ b/Libs/MRML/Logic/Testing/Cxx/vtkMRMLApplicationLogicTest1.cxx
@@ -39,6 +39,7 @@ int SliceLogicsTest();
 int SliceOrientationPresetInitializationTest();
 int TemporaryPathTest();
 int CreateUniqueFileNameTest(std::string tempDir);
+int AddModuleLogicTest();
 
 //-----------------------------------------------------------------------------
 int vtkMRMLApplicationLogicTest1(int argc, char *argv [])
@@ -57,6 +58,7 @@ int vtkMRMLApplicationLogicTest1(int argc, char *argv [])
   CHECK_EXIT_SUCCESS(SliceOrientationPresetInitializationTest());
   CHECK_EXIT_SUCCESS(TemporaryPathTest());
   CHECK_EXIT_SUCCESS(CreateUniqueFileNameTest(tempDir));
+  CHECK_EXIT_SUCCESS(AddModuleLogicTest());
   return EXIT_SUCCESS;
 }
 
@@ -238,4 +240,70 @@ int CreateUniqueFileNameTest(std::string tempDir)
     tempDir + "/CreateUniqueFileNameTest_2.nii.gz");
 
   return EXIT_SUCCESS;
+}
+
+
+//-----------------------------------------------------------------------------
+int AddModuleLogicTest()
+{
+  vtkNew<vtkMRMLApplicationLogic> appLogic;
+  std::string module_name = "TestModule";
+
+  // Registration of a module logic should work
+  {
+    vtkNew<vtkMRMLAbstractLogic> moduleLogic;
+    appLogic->SetModuleLogic(module_name.c_str(), moduleLogic.GetPointer());
+
+    CHECK_POINTER(appLogic->GetModuleLogic(module_name.c_str()),
+                  moduleLogic.GetPointer());
+  }
+
+  // Getting a pointer to a logic that it is gone should give nullptr.
+  // Note that moduleLogic from last scope is gone. GetModuleLogic should return nullptr
+  {
+    CHECK_NULL(appLogic->GetModuleLogic(module_name.c_str()));
+  }
+
+  // Trying to register a module logic associated to an already registered module
+  // should trigger an error.
+  {
+    vtkNew<vtkMRMLAbstractLogic> moduleLogic;
+    TESTING_OUTPUT_ASSERT_WARNINGS_BEGIN();
+    appLogic->SetModuleLogic(module_name.c_str(), moduleLogic.GetPointer());
+    TESTING_OUTPUT_ASSERT_WARNINGS(1);
+    TESTING_OUTPUT_ASSERT_WARNINGS_END();
+  }
+
+  // Trying to get a logic that has not been registered should return nullptr
+  {
+    const vtkMRMLAbstractLogic* retval = appLogic->GetModuleLogic((module_name+"a").c_str());
+
+    // return value should be nullptr
+    CHECK_NULL(retval);
+  }
+
+  // Passing nullptr to GetModuleLogic should trigger an error and return nullptr
+  {
+    // Triggers an error
+    TESTING_OUTPUT_ASSERT_ERRORS_BEGIN();
+    const vtkMRMLAbstractLogic* retval = appLogic->GetModuleLogic(nullptr);
+    TESTING_OUTPUT_ASSERT_ERRORS(1);
+    TESTING_OUTPUT_ASSERT_ERRORS_END();
+
+    // return value should be nullptr
+    CHECK_NULL(retval);
+  }
+
+  // Trying to unregister a non-registered module logic
+  {
+    appLogic->SetModuleLogic((module_name+"a").c_str(), nullptr);
+  }
+
+  // Removeing an already registered module logic should succeed
+  {
+    // Successfully removes module logic association
+    appLogic->SetModuleLogic(module_name.c_str(), nullptr);
+
+    const vtkMRMLAbstractLogic* retval = appLogic->GetModuleLogic((module_name).c_str());
+  }
 }
--- a/Libs/MRML/Logic/vtkMRMLApplicationLogic.cxx
+++ b/Libs/MRML/Logic/vtkMRMLApplicationLogic.cxx
@@ -56,6 +56,7 @@
 
 // STD includes
 #include <cassert>
+#include <map>
 #include <sstream>
 
 // For LoadDefaultParameterSets
@@ -86,7 +87,7 @@ public:
   vtkSmartPointer<vtkMRMLViewLinkLogic> ViewLinkLogic;
   vtkSmartPointer<vtkMRMLColorLogic> ColorLogic;
   std::string TemporaryPath;
-
+  std::map<std::string, vtkWeakPointer<vtkMRMLAbstractLogic> > ModuleLogicMap;
 };
 
 //----------------------------------------------------------------------------
@@ -822,4 +823,49 @@ void vtkMRMLApplicationLogic::EditNode(vtkMRMLNode* node)
 {
   // Observers in qSlicerCoreApplication listen for this event
   this->InvokeEvent(EditNodeEvent, node);
+}
+
+//----------------------------------------------------------------------------
+void vtkMRMLApplicationLogic::SetModuleLogic(const char* moduleName,
+                                             vtkMRMLAbstractLogic* moduleLogic)
+{
+  if (!moduleName)
+    {
+    vtkErrorMacro("AddModuleLogic: invalid module name.");
+    return;
+    }
+
+  // If no logic is provided, erase the module-logic association.
+  if (!moduleLogic)
+    {
+    this->Internal->ModuleLogicMap.erase(moduleName);
+    return;
+    }
+
+  // Check that the module has not any registered logic.
+  if (this->Internal->ModuleLogicMap.count(moduleName))
+    {
+    vtkWarningMacro("AddModuleLogic: Logic already register for " << moduleName);
+    return;
+    }
+
+  this->Internal->ModuleLogicMap[moduleName] = moduleLogic;
+}
+
+//----------------------------------------------------------------------------
+vtkMRMLAbstractLogic* vtkMRMLApplicationLogic::GetModuleLogic(const char* moduleName) const
+{
+  if (!moduleName)
+    {
+    vtkErrorMacro("GetModuleLogic: invalid module name");
+    return nullptr;
+    }
+
+  //Check that the logic is registered.
+  if (this->Internal->ModuleLogicMap.count(moduleName) == 0)
+    {
+    return nullptr;
+    }
+
+  return this->Internal->ModuleLogicMap.at(moduleName);
 }
--- a/Libs/MRML/Logic/vtkMRMLApplicationLogic.h
+++ b/Libs/MRML/Logic/vtkMRMLApplicationLogic.h
@@ -235,6 +235,19 @@ public:
   /// Requests the application to show user interface for editing a node.
   virtual void EditNode(vtkMRMLNode* node);
 
+  /// Sets a module with its corresponding logic to the application logic.
+  /// \param moduleName name of the module.
+  /// \param moduleLogic pointer to logic to be associated to the module. If this
+  /// parameter is nullptr, then the module logic will be removed from the application
+  /// logic.
+  void SetModuleLogic(const char* moduleName, vtkMRMLAbstractLogic* moduleLogic);
+
+  /// Gets a constant pointer to module logic associated with a given module
+  /// \param moduleName name of the module associated to the logic
+  /// \return constant pointer to vtkMRMLAbstractLogic corresponding to the
+  /// logic associated to th logic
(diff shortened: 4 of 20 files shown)


Base/QTGUI/qSlicerWidget.cxx
```diff
@@ -56,3 +70,16 @@ void qSlicerWidget::setMRMLScene(vtkMRMLScene* scene)
...
 }
+
+//-----------------------------------------------------------------------------
+void qSlicerWidget::setAppLogic(vtkSlicerApplicationLogic* logic)
+{
+  Q_D(qSlicerWidget);
+  d->AppLogic = logic;
+}
+
+vtkSlicerApplicationLogic* qSlicerWidget::appLogic()const
```
jcfr, line 81:
```suggestion
//-----------------------------------------------------------------------------
vtkSlicerApplicationLogic* qSlicerWidget::appLogic()const
```

Libs/MRML/Logic/vtkMRMLApplicationLogic.cxx
```diff
@@ -823,3 +824,68 @@ void vtkMRMLApplicationLogic::EditNode(vtkMRMLNode* node)
...
+  //       way to indicate there is that module that has no logic, as opposed to, the module
+  //       is not registered (could be that the module does not exists at all).
+  // if (!moduleLogic)
+  //   {
+  //   vtkErrorMacro("RegisterModuleLogic: module logic.");
+  //   return;
+  //   }
+
+  // Check that the module has not any registered logic.
+  if (this->Internal->ModuleLogicMap.count(moduleName))
```
jcfr, line 849:
```suggestion
if (this->Internal->ModuleLogicMap.count(moduleName) > 0)
```

lassoan: We should use the `empty()` method. In this case performance does not matter, but it is good practice and even for readability it is nicer to check if a container is empty instead of counting number of elements.

jcfr:
```suggestion
  if (!this->Internal->ModuleLogicMap.empty())
```

Libs/MRML/Logic/vtkMRMLApplicationLogic.cxx
```diff
@@ -823,3 +824,68 @@ void vtkMRMLApplicationLogic::EditNode(vtkMRMLNode* node)
...
   this->InvokeEvent(EditNodeEvent, node);
 }
+
+//----------------------------------------------------------------------------
+void vtkMRMLApplicationLogic::RegisterModuleLogic(const char* moduleName,
+                                                  vtkMRMLAbstractLogic* moduleLogic)
+{
+  if (!moduleName)
+    {
+    vtkErrorMacro("RegisterModuleLogic: invaild module name.");
```
jcfr, line 834:
```suggestion
    vtkErrorMacro("RegisterModuleLogic: invalid module name.");
```

Libs/MRML/Logic/Testing/Cxx/vtkMRMLApplicationLogicTest1.cxx
```diff
@@ -239,3 +241,98 @@ int CreateUniqueFileNameTest(std::string tempDir)
...
+    TESTING_OUTPUT_ASSERT_WARNINGS(1);
+    TESTING_OUTPUT_ASSERT_WARNINGS_END();
+
+    // return value should be nullptr
+    CHECK_NULL(retval);
+  }
+
+  // Passing nullptr to GetModuleLogic should trigger an error and return nullptr
+  {
+    //Triggers an error
```
jcfr, line 300:
```suggestion
    // Triggers an error
```

Libs/MRML/Logic/vtkMRMLApplicationLogic.cxx
```diff
@@ -823,3 +824,68 @@ void vtkMRMLApplicationLogic::EditNode(vtkMRMLNode* node)
   // Observers in qSlicerCoreApplication listen for this event
   this->InvokeEvent(EditNodeEvent, node);
 }
+
+//----------------------------------------------------------------------------
+void vtkMRMLApplicationLogic::RegisterModuleLogic(const char* moduleName,
```
jcfr, line 829: We should be able to use `const std::string& moduleName` as it is now wrapper in python.

Libs/MRML/Logic/vtkMRMLApplicationLogic.cxx
```diff
@@ -823,3 +824,68 @@ void vtkMRMLApplicationLogic::EditNode(vtkMRMLNode* node)
...
+                                                  vtkMRMLAbstractLogic* moduleLogic)
+{
+  if (!moduleName)
+    {
+    vtkErrorMacro("RegisterModuleLogic: invaild module name.");
+    return;
+    }
+
+  // NOTE: Should we accept registation of modules that have no logic (e.g. Welcome)
```
jcfr, line 838: registation -> registration

Libs/MRML/Logic/vtkMRMLApplicationLogic.cxx
```diff
@@ -823,3 +824,68 @@ void vtkMRMLApplicationLogic::EditNode(vtkMRMLNode* node)
...
+
+  this->Internal->ModuleLogicMap[moduleName] = moduleLogic;
+}
+
+//----------------------------------------------------------------------------
+void vtkMRMLApplicationLogic::UnregisterModuleLogic(const char* moduleName)
+{
+  if (!moduleName)
+    {
+    vtkErrorMacro("UnregisterModuleLogic: invaild module name.");
```
jcfr, line 863:
```suggestion
vtkErrorMacro("UnregisterModuleLogic: invalid module name.");
```

Libs/MRML/Logic/vtkMRMLApplicationLogic.cxx
jcfr, line 834:
```suggestion
    vtkErrorMacro("RegisterModuleLogic: invalid module name.");
```

Libs/MRML/Logic/vtkMRMLApplicationLogic.h
```diff
@@ -235,6 +235,21 @@ class VTK_MRML_LOGIC_EXPORT vtkMRMLApplicationLogic
   /// Requests the application to show user interface for editing a node.
   virtual void EditNode(vtkMRMLNode* node);
 
+  /// Registers a module with its corresponding logic
+  /// \param moduleName name of the module.
+  /// \param moduleLogic pointer to logic to be associated to the module.
+  void RegisterModuleLogic(const char* moduleName, vtkMRMLAbstractLogic* moduleLogic);
```
jcfr, line 241: For consistency, I think we should change `RegisterModuleLogic/UnregisterModuleLogic` to `SetModuleLogic`

Since we are associating an already instantiated logic with the application logic and not asking the application to instantiate a module logic for us, this would be more consistent with the current use of register/unregister across the application .

jcfr (changes requested):

Libs/MRML/Logic/vtkMRMLApplicationLogic.h
jcfr: cc: @lassoan

Libs/MRML/Logic/vtkMRMLApplicationLogic.cxx
lassoan: It would be nice to harmonize this in Slicer overall. Until we have `const char*` at some places, it is a bit complicated to set a string because if we get a string from somewhere as const char* then we always need to check if it is nullptr and if it is then replace it with "" (because setting nullptr as std::string& input immediately crashes the application).

Libs/MRML/Logic/vtkMRMLApplicationLogic.h
lassoan: Agreed. This is a simple cache, so I would just use Add/Remove/Get or Set/Get (remove by setting nullptr). Maybe Set/Get is simpler.

jcfr:
Also, when when writing commit message, prefer imperative "mood" (Enabling -> Enable, Exposing -> Expose). See https://slicer.readthedocs.io/en/latest/developer_guide/contributing.html#how-to-write-commit-messages

Libs/MRML/Logic/vtkMRMLApplicationLogic.cxx
RafaelPalomar: Actually, this line is not to check whether the map is empty, but to check whether `moduleName` exists in the map. `std::map::empty()` does not apply here.

Libs/MRML/Logic/vtkMRMLApplicationLogic.h
RafaelPalomar: @lassoan, @jcfr I would prefer to use Add/Remove/Get methods. Set/Get is simpler, but it is also more unclear an less regular (depending on the parameter you pass, you get the opposite behavior out of the function). Also, during module load, there will be modules that has no logic and will trigger runtime warnings ("SetModuleLogic: module is not registered"). 

A related question here would be: do we accept setting of modules with no logic (`nullptr`)?

Libs/MRML/Logic/vtkMRMLApplicationLogic.cxx
lassoan: In this case, `count()` would be fine (count is guaranteed to be 1, so it will not look for additional items). In general, it is more efficient to call `count()` only if you actually need count and otherwise getting the first element is faster (as it stops after it has found one element).

Libs/MRML/Logic/vtkMRMLApplicationLogic.cxx
```diff
@@ -823,3 +824,68 @@ void vtkMRMLApplicationLogic::EditNode(vtkMRMLNode* node)
...
+  //       way to indicate there is that module that has no logic, as opposed to, the module
+  //       is not registered (could be that the module does not exists at all).
+  // if (!moduleLogic)
+  //   {
+  //   vtkErrorMacro("RegisterModuleLogic: module logic.");
+  //   return;
+  //   }
+
+  // Check that the module has not any registered logic.
+  if (!this->Internal->ModuleLogicMap.empty())
```
lassoan, line 849: Changing the module logic is a valid operation. It would be more clear if the method name would be changed to SetModuleLogic.

Libs/MRML/Logic/vtkMRMLApplicationLogic.cxx
```diff
@@ -823,3 +824,68 @@ void vtkMRMLApplicationLogic::EditNode(vtkMRMLNode* node)
...
+    {
+    vtkWarningMacro("RegisterModuleLogic: Logic already register for " << moduleName);
+    return;
+    }
+
+  this->Internal->ModuleLogicMap[moduleName] = moduleLogic;
+}
+
+//----------------------------------------------------------------------------
+void vtkMRMLApplicationLogic::UnregisterModuleLogic(const char* moduleName)
```
lassoan, line 859: If we rename RegisterModuleLogic to SetModuleLogic then we can remove this method.

Libs/MRML/Logic/vtkMRMLApplicationLogic.cxx
```diff
@@ -823,3 +824,68 @@ void vtkMRMLApplicationLogic::EditNode(vtkMRMLNode* node)
...
+  if (!moduleName)
+    {
+      vtkErrorMacro("GetModuleLogic: invalid module name");
+      return nullptr;
+    }
+
+  //Check that the logic is registered.
+  if(this->Internal->ModuleLogicMap.count(moduleName) == 0)
+    {
+      vtkWarningMacro("GetModuleLogic: requested logic not registered");
```
lassoan, line 887: I think warning log is not needed here. It can be normal that you want to check if a module logic has been already set. Just returning nullptr is enough to signal to the caller that there is no such module logic.

Modules/Loadable/Markups/Widgets/qSlicerSimpleMarkupsWidget.cxx
```diff
@@ -485,20 +472,20 @@ void qSlicerSimpleMarkupsWidget::onMarkupsControlPointSelected(int row, int colu
   Q_UNUSED(column)
   Q_D(qSlicerSimpleMarkupsWidget);
 
-  if (d->JumpToSliceEnabled)
-    {
-    vtkMRMLMarkupsNode* currentMarkupsNode = vtkMRMLMarkupsNode::SafeDownCast( this->currentNode() );
-    if ( currentMarkupsNode == nullptr )
-      {
+  if (d->JumpToSliceEnabled) {
+    vtkMRMLMarkupsNode *currentMarkupsNode =
```
lassoan, line 476: can be a single line

Modules/Loadable/Markups/Widgets/qSlicerSimpleMarkupsWidget.cxx
```diff
@@ -485,20 +472,20 @@ void qSlicerSimpleMarkupsWidget::onMarkupsControlPointSelected(int row, int colu
...
+        vtkMRMLMarkupsNode::SafeDownCast(this->currentNode());
+    if (currentMarkupsNode == nullptr) {
```
lassoan, line 478: { should be in new line

Modules/Loadable/Markups/Widgets/qSlicerSimpleMarkupsWidget.cxx
```diff
@@ -485,20 +472,20 @@ void qSlicerSimpleMarkupsWidget::onMarkupsControlPointSelected(int row, int colu
...
       return;
-      }
-    if (d->MarkupsLogic == nullptr)
-      {
-      qCritical("qSlicerSimpleMarkupsWidget::onMarkupsControlPointSelected failed: Cannot jump, markups module logic is invalid");
+    }
+    vtkSlicerMarkupsLogic *markupsLogic = vtkSlicerMarkupsLogic::SafeDownCast(this->appLogic()->GetModuleLogic("Markups"));
+    if (markupsLogic == nullptr) {
```
lassoan, line 482: { should be in new line

lassoan (commented):
Nice work @RafaelPalomar, it is great that you implemented all the simplifications that are now possible and added tests. If you fix the remaining open comments then it should be ready to merge.

Libs/MRML/Logic/vtkMRMLApplicationLogic.h
lassoan: All these naming options could work, we just need to choose a one that is the most consistent with already established naming conventions.

I've checked what methods are now used in MRML for accessing maps (see details below):
- Get/Set/Remove 4x (in one case "Unset" is used instead of "Remove")
- Get/Add/Remove 2x

Add/Remove is tend to be used where "Add" means that a new item will be present (we do not want to overwrite an existing item, that would be an error). For example AddSegment clearly communicates the expectation that the new segment will be added and will not overwrite any existing segments. In contrast, "Set" effectively means "AddOrUpdate", i.e., if there is an existing key then we want to overwrite that.

Since in this case we want the "AddOrUpdate" behavior (e.g., when we reload a scripted module then we'll want to update the logic) and Get/Set/Remove is used more overall anyway, **we should use Get/Set/Remove**.

To be consistent with most other Get/Set/Remove implementations, Set(..., nullptr) should be equivalent with Remove.

---

vtkMRMLNode:
  GetAttribute
  SetAttribute
  RemoveAttribute

vtkMRMLScriptedModuleNode:
  void SetParameter(const std::string& name, const std::string& value);
  void UnsetParameter(const std::string& name);
  std::string GetParameter(const std::string& name) const;

vtkMRMLSegmentationDisplayNode:
  bool GetSegmentDisplayProperties(std::string segmentID, SegmentDisplayProperties &properties);
  void SetSegmentDisplayProperties(std::string segmentID, SegmentDisplayProperties &properties);
  void SetSegmentDisplayPropertiesToDefault(const std::string& segmentId);
  void RemoveSegmentDisplayProperties(std::string segmentID);

vtkSegment:
  vtkDataObject* GetRepresentation(std::string name);
  bool AddRepresentation(std::string type, vtkDataObject* representation);
  bool RemoveRepresentation(std::string name);

vtkSegmentation:
  bool AddSegment(vtkSegment* segment, std::string segmentId = "", std::string insertBeforeSegmentId = ""); 
  void RemoveSegment(std::string segmentId);
  vtkSegment* GetSegment(std::string segmentId); 
vtkSegment:

vtkSegmentation:
  void SetTag(std::string tag, std::string value);
  void RemoveTag(std::string tag);
  bool GetTag(std::string tag, std::string &value);

Libs/MRML/Logic/vtkMRMLApplicationLogic.cxx
```diff
@@ -57,6 +57,7 @@
 // STD includes
 #include <cassert>
 #include <sstream>
+#include <map>
```
jcfr, line 60:
```
#include <sstream>
#include <map>
```

->

```
#include <map>
#include <sstream>
```

Libs/MRML/Logic/vtkMRMLApplicationLogic.cxx
```diff
@@ -823,3 +824,48 @@ void vtkMRMLApplicationLogic::EditNode(vtkMRMLNode* node)
...
+  this->Internal->ModuleLogicMap[moduleName] = moduleLogic;
+}
+
+//----------------------------------------------------------------------------
+vtkMRMLAbstractLogic* vtkMRMLApplicationLogic::GetModuleLogic(const char* moduleName) const
+{
+  if (!moduleName)
+    {
+      vtkErrorMacro("GetModuleLogic: invalid module name");
+      return nullptr;
```
jcfr, line 861: Fix indent

Libs/MRML/Logic/vtkMRMLApplicationLogic.cxx
```diff
@@ -823,3 +824,48 @@ void vtkMRMLApplicationLogic::EditNode(vtkMRMLNode* node)
...
+    }
+
+  //Check that the logic is registered.
+  if(this->Internal->ModuleLogicMap.count(moduleName) == 0)
+    {
+      return nullptr;
```
jcfr, line 867: Fix indent

Modules/Loadable/Data/qSlicerDataModule.cxx
```diff
@@ -85,12 +86,19 @@ QStringList qSlicerDataModule::dependencies() const
...
-    qSlicerCoreApplication::application()->moduleManager()->module("Cameras");
   vtkSlicerCamerasModuleLogic* camerasLogic =
-    vtkSlicerCamerasModuleLogic::SafeDownCast(camerasModule ? camerasModule->logic() : nullptr);
+    vtkSlicerCamerasModuleLogic::SafeDownCast(this->appLogic()->GetModuleLogic("Cameras"));
+  // NOTE: here we assume that camerasLogic with a nullptr value can be passed
+  // to the qSlicerSceneReader. Therefore we trigger a warning but don't return
+  // immediately.
+  if (!camerasLogic)
+    {
+    qCritical() << Q_FUNC_INFO << ": Cameras module is not found";
```
jcfr, line 100: I am wondering if the warning should be completely removed as no warning were displayed before.

jcfr (commented):
Few nitpicks and it will be good to integrate

Modules/Loadable/Data/qSlicerDataModule.cxx
RafaelPalomar: I'm not sure myself, @jcfr. I tried to be cautions on the event of passing nullptr to `qSlicerSceneReader` a few lines below. As for now, it does not seem to trigger any unnecessary warning, but on the other hand it seems ulikely that the Cameras module logic is not in place at that point. @lassoan, any opinions on this?

RafaelPalomar:
> Few nitpicks and it will be good to integrate

@jcfr, @lassoan, If you are not thinking of squashing everything in one big commit, let me know and I will squash the code-review commits into the corresponding original ones. After that you can do further squashing to your liking before the merge.

Thank you both for the quick replies and the insightful code reviews.

jcfr:
Yes, it would be great if you could re-organize the commits and squash the one that should be together.

jcfr:
Ideally, the code base should be in a "buildable" state at each commit.

Modules/Loadable/Data/qSlicerDataModule.cxx
lassoan: I agree with @RafaelPalomar that it is better to be cautious and add the warning. If we find that there are false alarms (warnings are logged while starting up or shutting down the application) then we can refine the logic accordingly.

RafaelPalomar:
@jcfr, @lassoan review commits have been squashed with the original development. I have checked that every commit compiles without errors. I think it is ready for merge now.

Issue #5385, opened 2021-01-14
Crash when creating markups place widget


cpinter:
Slicer crashes when creating a qSlicerMarkupsPlaceWidget

See failing tests
http://slicer.cdash.org/test/16300684
http://slicer.cdash.org/test/16300948

## Steps to reproduce

In Python interactor:
`slicer.qSlicerMarkupsPlaceWidget()`

## Expected behavior

No crash

## Environment
- Slicer version: Slicer-4.13.0-2021-01-13
- Operating system: Windows 10

cpinter:
The crash is this line:
https://github.com/Slicer/Slicer/commit/914a566394997848a696f553b1ef8419139dd13a#diff-1eb70c7d69da9c2b93ac1b5fb596c73e083f9d6c51759b3ccf2ad1991678c5a8R115
The problem is that the `appLogic()` call returns nullptr

PR #5386, opened 2021-01-14
BUG: Fix crash when creating qSlicerMarkupsPlaceWidget and qSlicerSim…

cpinter:
…pleMarkupsWidget

Accessing the Markups module logic the way it was changed to did not work and lead to a crash. Since all the change to this class in the Markups module refactoring was related to accessing the Markups logic, it made sense to simply revert those changes.

Same applied to qSlicerSimpleMarkupsWidget.

Re https://github.com/Slicer/Slicer/issues/5385

### DIFF
--- a/Applications/SlicerApp/Testing/Cpp/qSlicerModuleFactoryManagerTest1.cxx
+++ b/Applications/SlicerApp/Testing/Cpp/qSlicerModuleFactoryManagerTest1.cxx
@@ -24,9 +24,10 @@
 #include <qSlicerModuleFactoryManager.h>
 #include <qSlicerCoreModuleFactory.h>
 #include <qSlicerCoreApplication.h>
+#include <vtkSlicerApplicationLogic.h>
 
-// STD includes
-
+// VTK includes
+#include <vtkNew.h>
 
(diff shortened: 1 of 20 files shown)


lassoan (dismissed):
If it fixes the crash (and at least some of the 17 new test errors) then we should merge this. @RafaelPalomar is then not under time pressure to submit a fix.

cpinter:
The two tests in the referenced issue that I linked pass with this change.

lassoan:
Wait, don't merge it yet, I'll see if I can quickly fix qSlicerWidget to get the applogic from the application (instead of expecting the user to set it).

lassoan:
I've tuned this a bit further - now we use the appLogic method in the application (according to the new design by @RafaelPalomar) but we get and store the markups module logic similarly as we did it before (according to the changes of @cpinter).

I'm running the Slicer automated tests now, if all looks good then I'll merge this.

lassoan:
Added some more fixes. Now all tests succeed on Windows (when starting Visual Studio using the launcher).

cpinter:
Thanks @lassoan

PR #5388, opened 2021-01-15
ENH: Add VoxelValueQuantities and VoxelValueUnits to volume nodes

lassoan:
Volumes were sometimes incorrectly loaded or saved because semantics of the volume - quantity of the voxel value - was not defined in the volume node.

Failing cases (that this commit fixes):
- When a displacement field was written to a MRML node then vector voxels stored displacements in RAS coordinate system. When this volume was saved then image grid was transformed to LPS but displacement values remained in RAS because the image writer had no way of knowing that voxels represent vectors that should be transformed along with the image axes.
- When a displacement field was read from a volume then image grid was converted to RAS but vectors that are stored in the voxel were not.
- When an color image was loaded into Slicer then saved as NRRD, the color array kind got converted to vector (instead of RGB or RGBA)

Open question: how should we store voxel information
- Option A. Use coded entries  (coding scheme designator, code value, code meaning) to specify voxel quantity and unit for each axis: flexible but complicated
- Option B: Specify enums in vtkMRMLVolumeNode to specify voxel quantity (and some internal representation but maybe coded entries for units). Voxel quantity enums could refer to the DICOM terminologies (http://dicom.nema.org/medical/dicom/current/output/chtml/part16/sect_CID_7180.html)

Option A. Pros:
- Flexibility of coded entries is nice. We can store any concept.
Option A Cons:
- The application may not recognize all codes and it would be difficult to know what codes are accepted. They may slightly differ
- Coding of even very simple things gets complicated. For example, setting that a volume is RGB requires 9 strings. This might have small performance penalty and adds lot of extra code (50-100 lines of code to get/set terms instead of 5-10; dozens of lines of code to robustly compare two quantity lists instead of a couple of lines).
- Even basic things may not have standard DICOM terms. For example, DICOM has terms for red, green, blue component of a color image, but there is no term for alpha. It does not have concept for general vectors, such as covariant, contravariant, gradient, normal, etc.
- Comparison may become non-trivial, as there could be equivalent codes in multiple coding schemes.

We could also use a mixed model - be able to save coded entries from DICOM import and use it during DICOM export, but not use it to decide how to use or interpret the voxels. However, this could lead to data inconsistency (enums state that it is an RGB image, coded entries state that it is a displacement image - which one should be trusted?).

Option A is implemented now but I feel it way too complicated, so I am thinking about switching to Option B. @pieper @fedorov I would like to hear your opinion before I do this, to ensure that it is a good choice (and I don't need to switch back to Option A later).


fedorov:
Considering that it is highly unlikely this mechanism will be used by anything other than Slicer, it is more important to have simplicity than comprehensiveness. I vote for B augmented by a (versioned) coding scheme maintained in Slicer documentation, mapping each of the items in the enum to a proper code tuple. Where standard codes exist, we should map enum items to those, and if they don't - use a Slicer-specific code (e.g., coding scheme "99SLICER"), which could later be substituted for a standard code. The file header can (should?) refer to that page for the definitions.

pieper:
+1 for option B.  This is consistent with the Slicer design where we map external representations to the needs of the Slicer application.

Thank you for taking this on.  I totally agree that adding quantity and units is critical to making the application easier to use by enabling it to automatically treat data as the user would expect.  👍

lassoan:
Thanks for the feedback. I've changed the implementation (and forced pushed in to this pull request) to have only a VoxelVectorType enum to specify that it is a generic vector (independent components or some unknown vector quantity), spatial vector (which inverts its signs when the image is converted between LPS/RAS), and color RGB/RGBA (this is not used right now, but we can use it to initialize things like independent/non-independent components when doing volume rendering).

PR #5392, opened 2021-01-18
ENH: Always interpolate control point measurements in markups

cpinter:
See discussion here https://discourse.slicer.org/t/perpendicular-cross-sectional-area-along-centerline/15556/103


lassoan (approved):

PR #5394, opened 2021-01-19
BUG: Position interaction handles at plane origin in world coordinates

Sunderlandkyl:
Interaction handles were incorrectly placed at the origin in local coordinates. Resulted in incorrect handle position if the plane was transformed.

### DIFF
--- a/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsPlaneNode.cxx
+++ b/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsPlaneNode.cxx
@@ -789,9 +789,9 @@ vtkMatrix4x4* vtkMRMLMarkupsPlaneNode::GetPlaneToPlaneOffsetMatrix()
 //---------------------------------------------------------------------------
 void vtkMRMLMarkupsPlaneNode::UpdateInteractionHandleToWorldMatrix()
 {
-  double handleX_World[3] = { 0.0 };
-  double handleY_World[3] = { 0.0 };
-  double handleZ_World[3] = { 0.0 };
+  double handleX_World[3] = { 0.0, 0.0, 0.0 };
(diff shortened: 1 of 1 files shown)


Modules/Loadable/Markups/MRML/vtkMRMLMarkupsPlaneNode.cxx
```diff
@@ -789,26 +789,26 @@ vtkMatrix4x4* vtkMRMLMarkupsPlaneNode::GetPlaneToPlaneOffsetMatrix()
 //---------------------------------------------------------------------------
 void vtkMRMLMarkupsPlaneNode::UpdateInteractionHandleToWorldMatrix()
 {
-  double handleX_World[3] = { 0.0 };
-  double handleY_World[3] = { 0.0 };
-  double handleZ_World[3] = { 0.0 };
+  double handleX_World[3] = { 0.0, 0.0, 0.0 };
```
lassoan, line 792: Does the change make any functional different or is it only for making it clear what the code does exactly? (I thought that `={0.0}` initializes each element to 0.0)

lassoan (commented):

Could you add a reference to the related bug report (if there was any)?

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsPlaneNode.cxx
Sunderlandkyl: It's just for clarity. I seem to recall from a previous PR that this was the preferred style for array initialization.

lassoan: OK, I just wanted to make sure that `={0}` was not wrong.

Sunderlandkyl:
@lassoan There is no associated bug report to reference.

lassoan (approved):

PR #5395, opened 2021-01-19
BUG: Fix terminology model update

cpinter:
When changing terminology of a node using the delegate, the model was not updated properly because apparently only the color and the terminology entry setting invoked the item changed signal, and with the previous order name was not updated at the correct time when the name was auto-generated.


lassoan (dismissed):
Thank you, looks good.

cpinter:
@lassoan Had to fix a merge conflict due to a very recent commit that changed this file. Please approve and integrate. Thanks

lassoan (approved):
Thank you!

PR #5396, opened 2021-01-20
COMP: Fix version check in extension bundled in custom applications

jcfr:
This commit ensures that Slicer_VERSION_* variables are not reset
to 0 after calling "find_package(Slicer)" in bundled extensions.


lassoan (approved):
Thank you, looks good to me!

PR #5400, opened 2021-01-25
STYLE: Make some members Qt properties

cpinter:
Some changes that I made during the project week and could be useful.

### DIFF
--- a/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyTreeView.h
+++ b/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyTreeView.h
@@ -69,6 +69,8 @@ class Q_SLICER_MODULE_SUBJECTHIERARCHY_WIDGETS_EXPORT qMRMLSubjectHierarchyTreeV
   Q_PROPERTY(QString nameFilter READ nameFilter WRITE setNameFilter)
   Q_PROPERTY(QStringList nodeTypes READ nodeTypes WRITE setNodeTypes)
   Q_PROPERTY(QStringList hideChildNodeTypes READ hideChildNodeTypes WRITE setHideChildNodeTypes)
+  Q_PROPERTY(vtkIdType currentItem READ currentItem WRITE setCurrentItem DESIGNABLE false)
+  Q_PROPERTY(QList<vtkIdType> currentItems READ currentItems WRITE setCurrentItems DESIGNABLE false)
 
 public:
   typedef QTreeView Superclass;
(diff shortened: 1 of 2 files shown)


jcfr (approved):

lassoan (commented):
This is good in the sense that we use members as properties whenever it is possible. However, it also breaks all extensions that have Python modules using members of this widget that were previously Q_INVOKABLE.

To have a softer transition (in 1-2 years), we could add new members (such as selectedSegmentID, or maybe even play with ID/ID capitalization but that's a clever hack and may cause confusion) and deprecate the old ones (log a warning message whenever they are used).

cpinter:
Good point. Would re-adding Q_INVOKABLE prevent breaking the extensions? I mean do you think it works even if the function is that of a property?

jcfr:
> Would re-adding Q_INVOKABLE prevent breaking the extensions? I mean do you think it works even if the function is that of a property?

I think it should ensure that both signature are available. It depends on how PythonQt is implemented. Could you give a try ?

cpinter:
@jcfr Sure I will. In the meantime I switched to another branch to investigate https://github.com/Slicer/Slicer/issues/5401 but will do soon.

cpinter:
It seems that the getter cannot be a property and a function at the same time. So even with Q_INVOKABLE, the function does not appear, only the property.

lassoan:
Yes, they need to be slightly different names.

cpinter:
OK good to know. I guess let's close this PR then. Sorry for the noise.

Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyTreeView.h
```diff
@@ -69,6 +69,8 @@ class Q_SLICER_MODULE_SUBJECTHIERARCHY_WIDGETS_EXPORT qMRMLSubjectHierarchyTreeV
   Q_PROPERTY(QString nameFilter READ nameFilter WRITE setNameFilter)
   Q_PROPERTY(QStringList nodeTypes READ nodeTypes WRITE setNodeTypes)
   Q_PROPERTY(QStringList hideChildNodeTypes READ hideChildNodeTypes WRITE setHideChildNodeTypes)
+  Q_PROPERTY(vtkIdType currentItem READ currentItem WRITE setCurrentItem DESIGNABLE false)
+  Q_PROPERTY(QList<vtkIdType> currentItems READ currentItems WRITE setCurrentItems DESIGNABLE false)
```
jcfr, line 73: You could update the PR to mark `setCurrentItem` and `setCurrentItems` as Q_INVOKABLE ?

cpinter: Those are slots, so are already callable.

lassoan:
> OK good to know. I guess let's close this PR then. Sorry for the noise.

OK, closing it then.

PR #5408, opened 2021-01-27
ENH: Always copy camera when importing MRML scene

jcfr:
Properties of the scene-to-import camera nodes are always copied into the existing nodes having the same name. This is done when a camera node is about to be added to the scene.


jcfr:
There are no references to `CopyImportedCameras` or `CopyCamera`  in Slicer extension sources.

PR #5409, opened 2021-01-27
Remove commented or unused code from qMRMLThreeDView and vtkMRMLSliceLogic

jcfr:


jcfr:
Functions `Initialize/IsInitialized` are not used in Slicer extensions

PR #5410, opened 2021-01-27
ENH: Simplify view widgets initialization from view nodes

jcfr:
This PR integrates changes originally contributed in  [lassoan/SlicerGitSVNArchive@cleanup-view-widget-creation2](https://github.com/lassoan/SlicerGitSVNArchive/tree/cleanup-view-widget-creation2) and  [lassoan/SlicerGitSVNArchive@cleanup-view-widget-creation](https://github.com/lassoan/SlicerGitSVNArchive/tree/cleanup-view-widget-creation).

Simplifies initialization of slice nodes adding the method `vtkMRMLSliceLogic::AddSliceNode(const char* layoutName)`
and removes the hard-coded viewer color names by ensuring the function
`vtkMRMLAbstractViewNode::GetLayoutColor()` is consistently used.

Associates the attribute `LayoutName` with the camera node, deprecates
the `ActiveTag` property and introduce the even `LayoutNameModifiedEvent`
equivalent to `ActiveTagModifiedEvent`.

Removes explicit association of view logics with the layout manager and
instantiated view widgets. Instead, view logics are implicitly instantiated
with each view widget controller and associated with the current MRML scene
in `qMRMLThreeDViewControllerWidget::setMRMLScene()`.


jcfr:
Few issues:
* The update of `qMRMLThreeDViewControllerWidget` with a valid view node is missing, the view controller button are not enabled.
* The threeD view label are not updated (this can be observed when switching to dual 3D view)

jcfr:
* ThreeD controller is now enabled :heavy_check_mark: 
* When view are linked, camera is not synchronized between viewer

jcfr:
Remaining tests to fix when built against VTK8:

```
	161 - vtkMRMLCameraDisplayableManagerTest1 (Failed)
	164 - vtkMRMLThreeDReformatDisplayableManagerTest1 (Failed)
	219 - qMRMLSliceControllerWidgetTest (Failed)
	221 - qMRMLSliceWidgetTest2_fixed.nrrd (Failed)
	227 - qMRMLUtf8Test1_cube-utf8.mrml (Failed)
	302 - vtkSlicerCamerasModuleLogicCopyImportedCamerasTest (Failed)
	359 - vtkMRMLMarkupsNodeTest3 (Failed)
	368 - vtkMarkupsAnnotationSceneTest (Failed)
	391 - qMRMLModelDisplayNodeWidgetTest2 (Failed)
	407 - py_nomainwindow_SegmentationsModuleTest2 (Failed)
```

jcfr:
> * The threeD view labels are not updated (this can be observed when switching to dual 3D view). 

This is now fixed in e91f37e :heavy_check_mark: 

@lassoan It took me some time to find this one. I am wondering if we should have a mode to display explicit warnings when a connection failed to be established. Using https://github.com/KDAB/GammaRay allows me realized that the slot `updateViewFromMRMLCamera` was not connected

jcfr:
On my workstation, remaining tests failure when building against VTK8 are the following:

```
vtkMRMLMarkupsNodeTest3
py_nomainwindow_SegmentationsModuleTest2
```

lassoan:
> > * The threeD view labels are not updated (this can be observed when switching to dual 3D view).
> 
> This is now fixed in [e91f37e](https://github.com/Slicer/Slicer/commit/e91f37efd65ca761c8baefee31cf0370115292e3) ✔️
> 
> @lassoan It took me some time to find this one. I am wondering if we should have a mode to display explicit warnings when a connection failed to be established. Using https://github.com/KDAB/GammaRay allowed me realized that the slot `updateViewFromMRMLCamera` was not connected

Do you mean that this method does not log a warning when a connection is specified incorrectly (similar to the warnings that Qt logs)?

lassoan:
> On my workstation, remaining tests failure when building against VTK8 are the following:
> 
> ```
> vtkMRMLMarkupsNodeTest3
> py_nomainwindow_SegmentationsModuleTest2
> ```

These are probably not regressions, but it is hard to know for sure - there have been no linux and macos nightly tests for a good while now.

jcfr:
> Do you mean that this method does not log a warning when a connection is specified incorrectly (similar to the warnings that Qt logs)?

If the connection is specified incorrectly (e.g non existing slot), I think a warning is displayed.

I am just wondering if we could find a way to systematically avoid issues like the one I fixed in e91f37e

When the incorrect signature `this->qvtkReconnect(d->CameraNode, vtkMRMLCameraNode::CameraInteractionEvent, this, SLOT(updateViewFromMRMLCamera()));` was used, an empty connection id was returned. 

To conclude, what I suggest is likely overkill and this is just a "bug" that needed to be fixed.

jcfr:
This topic is ready for final review.

Note that it will  be squashed before integration.

cpinter (dismissed):
After a quick read through it is obvious that the code has become simpler and more readable and straightforward. Thanks a lot for working on this @jcfr and @lassoan !

jcfr:
After re-running all tests last night, these two are failing because I updated `cub.mrml` and the tests still expect the deprecation warning.

```
	124 - vtkMRMLSceneTest2_cube.mrml (Failed)
	125 - vtkMRMLSceneImportTest_cube.mrml (Failed)
```

jcfr:
All `vtkMRMLScene` tests are now passing.

Remaining two failing tests are `vtkMRMLMarkupsNodeTest3` and `py_nomainwindow_SegmentationsModuleTest2` which also fail without this PR.

jcfr:
* [ ] Update https://www.slicer.org/wiki/Documentation/Nightly/ScriptRepository#Show_a_slice_view_outside_the_view_layout
* [ ] Rename `LayoutName` to `ViewIDInLayout`
* [ ] Add documentation

jcfr:
> Rename LayoutName to ViewIDInLayout

I started to work on this and then realized that this is consistent with the `LayoutName` already associated with [vtkMRMLAbstractViewNode](https://github.com/Slicer/Slicer/blob/58b38ab63329d8efd2632a0d6ba1aa96c70b411e/Libs/MRML/Core/vtkMRMLAbstractViewNode.h#L62-L69)

Is it worth changing the view node `LayoutName` to `ViewIDInLLayout` and adding support for older scene ?

cc: @lassoan

jcfr:
### Show a slice view outside the view layout

See https://www.slicer.org/wiki/Documentation/Nightly/ScriptRepository#Show_a_slice_view_outside_the_view_layout

_Currently working on updating this example_

### Show a 3D view outside the view layout

See https://www.slicer.org/wiki/Documentation/Nightly/ScriptRepository#Show_a_3D_view_outside_the_view_layout

```python
layoutName = "Test3DView"
layoutLabel = "T3"
layoutColor = [1.0, 1.0, 0.0]
# ownerNode manages this view instead of the layout manager (it can be any node in the scene)
viewOwnerNode = slicer.mrmlScene.AddNewNodeByClass("vtkMRMLScriptedModuleNode")

# Create MRML node
viewLogic = slicer.vtkMRMLViewLogic()
viewLogic.SetMRMLScene(slicer.mrmlScene)
viewNode = viewLogic.AddViewNode(layoutName)
viewNode.SetLayoutLabel(layoutLabel)
viewNode.SetLayoutColor(layoutColor)
viewNode.SetAndObserveParentLayoutNodeID(viewOwnerNode.GetID())

# Create widget
viewWidget = slicer.qMRMLThreeDWidget()
viewWidget.viewLabel = layoutLabel
viewWidget.viewColor = qt.QColor.fromRgbF(*layoutColor)
viewWidget.setMRMLScene(slicer.mrmlScene)
viewWidget.setMRMLViewNode(viewNode)
viewWidget.show()
```

```diff
layoutName = "Test3DView"
layoutLabel = "T3"
layoutColor = [1.0, 1.0, 0.0]
# ownerNode manages this view instead of the layout manager (it can be any node in the scene)
viewOwnerNode = slicer.mrmlScene.AddNewNodeByClass("vtkMRMLScriptedModuleNode")

# Create MRML node
-viewNode = slicer.vtkMRMLViewNode()
-viewNode.SetName(layoutName)
-viewNode.SetLayoutName(layoutName)
+viewLogic = slicer.vtkMRMLViewLogic()
+viewLogic.SetMRMLScene(slicer.mrmlScene)
+viewNode = viewLogic.AddViewNode(layoutName)
viewNode.SetLayoutLabel(layoutLabel)
viewNode.SetLayoutColor(layoutColor)
viewNode.SetAndObserveParentLayoutNodeID(viewOwnerNode.GetID())
-viewNode = slicer.mrmlScene.AddNode(viewNode)

# Create widget
viewWidget = slicer.qMRMLThreeDWidget()
viewWidget.viewLabel = layoutLabel
viewWidget.viewColor = qt.QColor.fromRgbF(*layoutColor)
viewWidget.setMRMLScene(slicer.mrmlScene)
viewWidget.setMRMLViewNode(viewNode)
viewWidget.show()
```

jcfr:
@lassoan  I just updated the example for the 3D View, see https://github.com/Slicer/Slicer/pull/5410#issuecomment-772184310

It now makes use of the convenience method `vtkMRMLViewLogic::AddViewNode()` that I also updated to return the node and add it to the scene.

Later this evening, I will work on updating the slice view ones, it currently crashes.

lassoan:
> @lassoan I just updated the example for the 3D View and Slice view, see [#5410 (comment)](https://github.com/Slicer/Slicer/pull/5410#issuecomment-772184310)
> 
> It now makes use of the convenience methods like `vtkMRMLViewLogic::AddViewNode()` and `vtkMRMLSliceLogic::AddSliceNode()` that I also updated to return the node and add it to the scene.
> 
> **Note**: I am working on simplifying the Slice example one (for example the composite nodes are now automatically added)

`layoutName = "Test3DView"` => add a comment here that this refers to view ID in layout description

`viewNode = slicer.vtkMRMLViewNode()` => we need to use slicer.mrmlScene.AddNewNodeByClass because that's how all the default view settings are set into view nodes (Application Settings / Views).

```
viewWidget.viewLabel = layoutLabel
viewWidget.viewColor = qt.QColor.fromRgbF(*layoutColor)
```
Why do we still need to set these? Shouldn't it retrieve these from the view node?

jcfr:
> layoutName = "Test3DView" => add a comment here that this refers to view ID in layout description (and therefore it should be set to a value that is not used in any of the layouts owned by the layout manager)

All set. I updated the comment above.

jcfr:
> viewNode = slicer.vtkMRMLViewNode() => we need to use slicer.mrmlScene.AddNewNodeByClass because that's how all the default view settings are set into view nodes (Application Settings / Views updates the default view nodes in the scene).

Agreed. This is already the case because the `vtkMRMLSliceLogic` and `vtkMRMLViewLogic` are used and internally they make sure of `CreateNodeByClass`

jcfr:
> ```
> viewWidget.viewLabel = layoutLabel
> viewWidget.viewColor = qt.QColor.fromRgbF(*layoutColor)
> ```
> Why do we still need to set these? Shouldn't the widget retrieve these from the view node?

All set


After this is integrated, I think we could look into supporting this instead: 

```diff
-sliceLogics = slicer.app.applicationLogic().GetSliceLogics()
-viewWidget.setSliceLogics(sliceLogics)
+viewWidget.setSliceLogics(slicer.app.applicationLogic().GetSliceLogics())
-sliceLogics.AddItem(viewWidget.sliceLogic())
-viewWidget.show()
```

Indeed, the `viewWidget` has all needed information to successfully add its own logic to the global list of slice logics. We could even make all the lines obsolete.

jcfr:
I started to document the layout infrastructure in https://github.com/Slicer/Slicer/pull/5433 and will plan to finalize.

@lassoan @cpinter  In the meantime, I would like to move forward with the integration of this PR.  Or should we do a stable release of current master first ?

lassoan:
We can merge this now and if we don't encounter any issues for 4-5 days then we can create the stable release from that. If we run into issues that we cannot quickly fix and we want to move on with the VTK upgrade then we can create the release from the version before this change.

lassoan:
> Indeed, the `viewWidget` has all needed information to successfully add its own logic to the global list of slice logics. We could even make all the lines obsolete.

Very good point. We usually don't require module developers to create logic classes manually, they just need to create MRML nodes and add widgets to their module GUIs.

lassoan:
> Is it worth changing the view node attribute `LayoutName` to `ViewIDInLLayout` and adding support for older scene ?

We can change the code later and just explain everything in the documentation.

cpinter:
Build fails due to a trivial syntax error. Fix is here:
https://github.com/Slicer/Slicer/pull/5435

I think it would make sense to give my write permissions so that I can quick fix stuff like this before any of you wakes up. It would give the community an additional 5 hours window of hot support possibility.

jcfr:
> It might make sense to give my write permissions so that I can quick fix stuff like this

Makes sense, it turns out you already part of the [slicer-core](https://github.com/orgs/Slicer/teams/slicer-core/members) which has [Maintainer](https://docs.github.com/en/github/setting-up-and-managing-organizations-and-teams/repository-permission-levels-for-an-organization) role for all Slicer organization repositories. 

It turns out that only administrator can merge without having the pull request check being green.

lassoan:
Indeed, it is not possible to bypass code review requirement for specific groups or individuals. I would trust Csaba to be added as admin, but it would not be nice if we could only manage our software process by adding core developers as repository owners. I think there are two sustainable ways to resolve this:
- Option A: Consider code reviews essential and enforce them to everyone (including admins).
- Option B: Consider code reviews optional. Allow a group of trusted developers to push to master branch without review.

I think option A would be the right thing to do for a large (and this is what we have now, except we have not blocked admin access yet), but if we think this is too inconvenient or slows things down too much then we can try option B.

pinging @pieper, too

cpinter:
Just a quick argument about option A: I think it's good that at least some of us has direct write access, because in an emergency that quick commit can be made without haveing to wait for someone else to be able to look at the PR. If we had 20 core developers maybe this would not be a problem, but the number of people that regularly participate in code reviews is only around 5.

jcfr:
>  slows things down too much

To speed things up, I suggest:
* we introduce a way to mark commit as "trivial" such commit would not require "CircleCI" to complete
* we implicitly consider commit starting with `DOC:` and `STYLE: ` as "trivial"
* commit updating `SuperBuild/*` or `SuperBuild.cmake` files would ensure CircleCI returns early with a green status

pieper:
I'd vote for giving Csaba the right to bypass review for the very practical time zone issues he raised.  I think we can all agree to go through the review process for all but the very most trivial changes.  That's worked well so far.   Even for DOC and STYLE it's not a huge burden.

jamesobutler:
I'm a bit partial to Option A as described by Andras. There is no having to agree how a developer should interpret "trivial".  Even if trivial, I do not think it is critical to have the change in master for ~5 hours. The change can be used locally for the developer trying to build master.  This is assuming the trivial fix was also trivial to find the source of the issue that broke things.

PR #5411, opened 2021-01-28
BUG: Fix vtkSegmentationConverter::GetCheapestPath if cost >= 10000000

Sunderlandkyl:
For some Segmentation conversion methods vtkSegmentationConverterRule::GetConversionInfiniteCost() (10000000) is used to represent a last resort conversion path that should not be used unless it is the only one that can be found.
GetCheapestPath initialized the cheapestPathCost using 10000000. This meant that the cheapest path would never be found if its cost was >= 10000000.

Fixed by initializing cheapestPathCost to VTK_UNSIGNED_INT_MAX.

See SlicerRt/SlicerRT#177

### DIFF
--- a/Libs/vtkSegmentationCore/vtkSegmentationConverter.cxx
+++ b/Libs/vtkSegmentationCore/vtkSegmentationConverter.cxx
@@ -303,7 +303,7 @@ std::string vtkSegmentationConverter::GetConversionParameterDescription(const st
 //----------------------------------------------------------------------------
 vtkSegmentationConverter::ConversionPathType vtkSegmentationConverter::GetCheapestPath(const ConversionPathAndCostListType &pathsCosts)
 {
-  unsigned int cheapestPathCost = vtkSegmentationConverterRule::GetConversionInfiniteCost();
+  unsigned int cheapestPathCost = VTK_UNSIGNED_INT_MAX;
   unsigned int cheapestPathNumberOfConversions = 0;
   ConversionPathType cheapestPath;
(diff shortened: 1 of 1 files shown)


lassoan (approved):
Thank you for the lightning-fast fix!

Libs/vtkSegmentationCore/vtkSegmentationConverter.cxx
```diff
@@ -303,7 +303,7 @@ std::string vtkSegmentationConverter::GetConversionParameterDescription(const st
 //----------------------------------------------------------------------------
 vtkSegmentationConverter::ConversionPathType vtkSegmentationConverter::GetCheapestPath(const ConversionPathAndCostListType &pathsCosts)
 {
-  unsigned int cheapestPathCost = vtkSegmentationConverterRule::GetConversionInfiniteCost();
```
cpinter, line 306: Would it make more sense to change vtkSegmentationConverterRule::GetConversionInfiniteCost()? If the value it returns causes bugs, then it should not be left as is.

Sunderlandkyl: I don't think the value of GetConversionInfiniteCost is an issue. 

It was only a problem for initializing the cheapestPathCost because it was also used as the cost of planar contour to ribbon conversion. It wouldn't matter what the value was in this case, it would still cause the issue.

cpinter: I see. Thanks for the explanation

PR #5412, opened 2021-01-28
ENH: Double click on type selects terminology

cpinter:
Besides clicking Select in the terminology selector dialog, now double-clicking on the type has the same effect. It makes selecting terminology faster in simple cases (when modifiers do not need to be set).


lassoan (approved):
This will be very useful, thank you!

PR #5413, opened 2021-01-28
DOC: Update macOS build instructions

piiq:
This PR updates build instructions for macOS. The updates include:

- Hints for debugging the build.
- Improved wording of the packaging section.
- Hints for debugging packaging errors when libarchive takes homebrew dependencies instead of local ones.
- Minor typos and formatting changes.

@pieper @lassoan would be nice if you have a look.


Docs/developer_guide/build_instructions/macos.md
```diff
@@ -5,74 +5,108 @@
...
 
 ## Configure and generate Slicer solution files
 
-- Configure using the following commands. By default `CMAKE_BUILD_TYPE` is set to `Debug` (replace `/path/to/QtSDK` with the real path on your machine where QtSDK is located):
-```
+- Configure using the following commands. By default `CMAKE_BUILD_TYPE` is set to `Debug` (replace `/path/to/Qt` with the real path on your machine where QtSDK is located):
+
+```bash
 mkdir Slicer-SuperBuild-Debug
 cd Slicer-SuperBuild-Debug
```
pieper, line 51: maybe these should be changed to `/opt/s`

Docs/developer_guide/build_instructions/macos.md
```diff
@@ -102,16 +140,43 @@ See list of issues common to all operating systems on [Common errors](common_err
...
+
+```bash
 xcode-select --install
 ```
 
 ### dyld: malformed mach-o: load commands size (...) > 32768
 
-Path of source or build folder is too long. For example building Slicer in */User/somebody/projects/something/dev/slicer/slicer-qt5-rel* may fail with malformed mach-o error, while it succeeds in */sq5* folder. To resolve this error, move source and binary files into a folder with shorter full path and restart the build from scratch (the build tree is not relocatable).
+Path of source or build folder is too long. For example building Slicer in */User/somebody/projects/something/dev/slicer/slicer-qt5-rel* may fail with malformed mach-o error, while it succeeds in */opt/s* folder. To resolve this error, move source and binary files into a folder with shorter full path and restart the build from scratch (the build tree is not relocatable).
+
```
pieper, line 158: We could clarify here that only the build tree needs to be in a short path.  Source can be anywhere.

Docs/developer_guide/build_instructions/macos.md
```diff
@@ -102,16 +140,43 @@ See list of issues common to all operating systems on [Common errors](common_err
...
+#### LibArchive pointing to a nonexistent path
+
+If a packaged Slicer is launched on another mac and it crashes with the error log saying that
+
+```
+Library not loaded: /usr/local/opt/zstd/lib/libzstd.1.dylib
+  Referenced from: Slicer.app/Contents/lib/Slicer-4.13/libarchive.17.dylib
+```
+
+It means that `libarchive` has has found homebrew versions of some of it's requirements, rather than local ones. For the packaged version of Slicer to run on other machines none of the prerequisites should be installed via homebrew. For example `lz4` and `zstd` are bundled with `subversion` and `rsync` so if you have these two application installed via homebrew, `libarchive` will grab them from `/usr/local/opt/` and the packaged Slicer will not run on other machines. The solution is either to remove them from from homebrew with `brew remove lz4` and `brew remove zsdt` or to change the `$PATH` so that the local build folder goes before `/usr/local/opt/`. After doing this Slicer should be rebuilt and repackaged.
```
pieper, line 182: Is there a github issue filed for this?  would be good to file one is Slicer so we can look at how to force libarchive to use the right versions of these libraries.  If it means changing libarchive we can work to get it fixed upstream, like [this issue was](https://github.com/libarchive/libarchive/pull/1395).

pieper (approved):
This is really appreciated!

I had a couple small comments and maybe we should change to using `/opt/s` as the superbuild directory name consistently?

pieper:
It's already a big improvement but I'll wait to merge until I hear from you about the comments.

Docs/developer_guide/build_instructions/macos.md
jamesobutler: On Windows path length of source and build tree is enforced to 40 characters. Does this need to be changed and also enforced on other platforms? [PreventDirWithTooManyChars.cmake](https://github.com/Slicer/Slicer/blob/767551b4b8cf85f9d3193af1a5ff4ff8f3fdf728/CMake/PreventDirWithTooManyChars.cmake)

https://github.com/Slicer/Slicer/blob/767551b4b8cf85f9d3193af1a5ff4ff8f3fdf728/CMakeLists.txt#L85-L92

pieper: I believe that @jcfr said that the new vtk will have a lot fewer python related shared libraries so we can test if that removes this as a practical concern on mac. But yes, a note on the windows build instructions would make sense.

Docs/developer_guide/build_instructions/macos.md
```diff
@@ -5,74 +5,108 @@
...
 - XCode command line tools must be installed:
-```
+
+```bash
 xcode-select --install
 ```
+
 - A CMake version that meets at least the minimum required CMake version [here](https://github.com/Slicer/Slicer/blob/master/CMakeLists.txt#L1)
-- Large File Storage for git is required. (`brew install git-lvs`)
+- Large File Storage for git is required. (`brew install git-lfs`)
```
jcfr, line 14: We should double check but I think this is not required anymore.

Docs/developer_guide/build_instructions/macos.md
```diff
@@ -5,74 +5,108 @@
...
+- Setting `CMAKE_OSX_DEPLOYMENT_TARGET` CMake variable specifies the minimum macOS version a generated installer may target. So it should be equal to or less than the version of SDK you are building on. Note that the SDK version is set using `CMAKE_OSX_SYSROOT` CMake variable automatically initialized during CMake configuration.
 
 ## Checkout Slicer source files
 
 Notes:
+
 - While it is not enforced, we strongly recommend you to *avoid* the use of *spaces* for both the `source directory` and the `build directory`.
-- Due to maximum path length limitations during build the build process, source and build folders must be located in a folder with very short total path length. This is expecially critical on Windows and MacOS. For example, `/sq5` has been confirmed to work on MacOS.
+- Due to maximum path length limitations during build the build process, source and build folders must be located in a folder with very short total path length. This is expecially critical on Windows and MacOS. For example, `/opt/s` has been confirmed to work on MacOS.
```
jcfr, line 25: On a recent version of macOs, is it an issue to use '/opt' ? see https://apple.stackexchange.com/questions/399421/opt-folder-no-longer-accessible-in-finder-on-catalina

Docs/developer_guide/build_instructions/macos.md
```diff
@@ -5,74 +5,108 @@
...
-```
+- Start a terminal and type the following (you can replace 4 by the number of processor cores in the computer. You can find out the number of available cores by running `sysctl -n hw.ncpu`):
+
+```bash
 cd ~/Projects/Slicer-SuperBuild
 make -j4
 ```
 
-When using the -j option, the build will continue past the source of the first error. If the build fails and you don't see what failed, rebuild without the -j option. Or, to speed up this process build first with the -j and -k options and then run plain make. The -k option will make the build keep going so that any code that can be compiled independent of the error will be completed and the second make will reach the error condition more efficiently.
+### Debugging the build process
```
jcfr, line 89: Should we move this onto its own page ? So that it can linked from Linux and macOS build instructions?

Docs/developer_guide/build_instructions/macos.md
jcfr: Ditto. With VTK9, we won't have the PythonD library anymore.

jcfr (commented):
Thanks for the contribution 🙏

Few minor comments and it should be good.

Docs/developer_guide/build_instructions/macos.md
lassoan: We already have a note in Windows build instructions on using short paths: "Due to maximum path length limitations during build the build process, source and build folders must be located in a folder with very short (couple of characters) total path length."  (https://slicer.readthedocs.io/en/latest/developer_guide/build_instructions/windows.html#set-up-source-and-build-folders)

Docs/developer_guide/build_instructions/macos.md
pieper: I agree, I'm pretty sure I didn't install it in my latest mac and it doesn't show up in `brew list`

Docs/developer_guide/build_instructions/macos.md
pieper: I was able to add it to the dock on the side of the finder as a shortcut, so that hasn't been a problem.

We should point out if we haven't that you need to use `sudo` for `mkdir` and `chown` in `/opt`

Docs/developer_guide/build_instructions/macos.md
piiq: Maybe common errors and debugging can be on the same page?

Docs/developer_guide/build_instructions/macos.md
piiq: There wasn't an issue filed for this so I've [created one](https://github.com/Slicer/Slicer/issues/5415).

Docs/developer_guide/build_instructions/macos.md
lassoan: Please keep linux and mac instructions on completely separate pages. We should only put very special things on the "Common errors" page: that are common to all operating systems _and_ not part of the build instructions (exceptions). But we may even consider removing that section (depending on how its content evolves), because it is much easier to find information if everything is on a single page.

Trying to reduce duplication in build instructions leads really bad results, as we learned in the past. It is obviously bad for developers (I had to keep scrolling the documentation pages really carefully and figure out for each section if it is applicable to my platform). But it is also bad for maintainers and results in lower quality documentation, because when you update build instructions for linux or mac then you either end up just writing instructions without testing on the other platform; or you build and edit instructions on linux and mac at the same time and make mistakes because you are too distracted.

Docs/developer_guide/build_instructions/macos.md
piiq: @pieper , do you have anything apart from slicer build folders in your `/opt`? 

I've read the linked stackexchange post and took another look in my `/opt`. On my machine it looks like `/opt` was brought in by migration assistant from an older laptop. When I was setting up the Slicer build I didn't have problems writing to `/opt` because the folder was already there and is contained files that were created by me like 6 years ago, before the actual computer was assembled on the factory.
On the other hand when I tried to create the `/sq5` folder the system didn't let me (even with `sudo`) because `/` "is" the mentioned read-only system volume. 

It might be a case what @jcfr is talking about.
In order to test this someone with a mac that had 10.15+ pre-installed on the new machine and who didn't use migration assistant from the older mac to check if `/opt` is available and writable with or without `sudo`

pieper: On my old machine I used `/s`.  When I moved to my new mac manually (no migration assistant) and started using `/opt` because it was the shortest path that already existed and I could write to (with sudo).  There's also a `/otp/cisco` that's part of the hospital vpn and it is owned by root so it was created by the installer.  So I think it's a reasonable place to use.  I also put an `/opt/tmp` and '/opt/data` for stuff I want local but don't want in my home directory (don't really care if it gets lost).

piiq: @pieper and your new mac came with Catalina, right?

pieper: > @pieper and your new mac came with Catalina, right?

yes

piiq:
Thanks for all the comments @pieper @jcfr @lassoan!

I've updated the PR:

- included consistent references to `/opt/s` as the advised build folder path
- removed references to deprecated requirements
- separated build and debug sections while keeping them on the same page

pieper (approved):
👍

pieper:
If there are no other suggestions let's merge this.

Docs/developer_guide/build_instructions/macos.md
```diff
@@ -5,113 +5,177 @@
...
+- Setting `CMAKE_OSX_DEPLOYMENT_TARGET` CMake variable specifies the minimum macOS version a generated installer may target. So it should be equal to or less than the version of SDK you are building on. Note that the SDK version is set using `CMAKE_OSX_SYSROOT` CMake variable automatically initialized during CMake configuration.
 
 ## Checkout Slicer source files
 
 Notes:
+
 - While it is not enforced, we strongly recommend you to *avoid* the use of *spaces* for both the `source directory` and the `build directory`.
-- Due to maximum path length limitations during build the build process, source and build folders must be located in a folder with very short total path length. This is expecially critical on Windows and MacOS. For example, `/sq5` has been confirmed to work on MacOS.
+- Due to maximum path length limitations during build the build process, build folders must be located in a folder with very short total path length. This is expecially critical on Windows and MacOS. For example, `/opt/s` has been confirmed to work on MacOS.
```
jamesobutler, line 24: Note there is merge conflicts due to an update of this line in https://github.com/Slicer/Slicer/commit/8a3730d3e5860502bff203c61054a5d6415303eb. During rebase it should be resolved to the following:
```suggestion
- Due to maximum path length limitations during build the build process, build folders must be located in a folder with very short total path length. This is expecially critical on Windows and macOS. For example, `/opt/s` has been confirmed to work on macOS.
```

Docs/developer_guide/build_instructions/macos.md
```diff
@@ -5,113 +5,174 @@
...
+- Due to maximum path length limitations during build the build process, build folders must be located in a folder with very short total path length. This is expecially critical on Windows and macOS. For example, `/opt/s` has been confirmed to work on macOS.
 
 Check out the code using `git`:
-- Clone the github repository</p>
-```cd MyProjects
+
+- Clone the github repository
+
+```bash
+cd MyProjects
```
lassoan, line 31: Above it is written that source folder length matters, too. What is the recommended source folder location then?

Docs/developer_guide/build_instructions/macos.md
```diff
@@ -5,113 +5,174 @@
...
 git clone git://github.com/Slicer/Slicer.git
 ```
+
 The `Slicer` directory is automatically created after cloning Slicer.
+
 - Setup the development environment:
-```cd Slicer
```
lassoan, line 31: Is the recommended sourcelocation `/opt/Slicer`?

Docs/developer_guide/build_instructions/macos.md
```diff
@@ -5,113 +5,174 @@
...
+- Configure using the following commands. By default `CMAKE_BUILD_TYPE` is set to `Debug` (replace `/path/to/Qt` with the real path on your machine where QtSDK is located):
+
+```bash
+mkdir /opt/s
+cd /opt/s
+cmake \
+  -DCMAKE_OSX_DEPLOYMENT_TARGET:STRING=10.13 \
+  -DCMAKE_BUILD_TYPE:STRING=Debug \
+  -DQt5_DIR:PATH=/path/to/Qt/lib/cmake/Qt5 \
+  /path/to/Slicer
```
lassoan, line 55: We should replace /path/to/Slicer with the recommended source location.

Docs/developer_guide/build_instructions/macos.md
lassoan: Could you add a link to the issue to this paragraph?

lassoan (changes requested):
I've consolidated the merge conflicts, squashed all the commits, fixed the co-authors. It is almost ready to go except we should clarify what location we recommend for the source files and use that location consistently throughout the page. @piiq could you address this?

Docs/developer_guide/build_instructions/macos.md
pieper: I've never seen a problem with the length of the source path.  I always use /Users/pieper/slicer/latest/Slicer with no problem.

piiq: The source location is irrelevant on macOS. Should there be a recommended source location?

pieper: Above where it talks about the length being important we could clarify that they source path doesn't matter, only the build path.

lassoan (approved):
Looks good, thank you!

PR #5414, opened 2021-01-28
BUG: Fix calculation immediately after enabling measurement

cpinter:


lassoan (approved):
Perfect, thank you for the fix!

PR #5416, opened 2021-01-29
ENH: Show colors in terminology selector

cpinter:
The recommended colors for are shown as a color square for each type in the terminology navigator widget.

image

### DIFF
--- a/Modules/Loadable/Terminologies/Widgets/qSlicerTerminologyNavigatorWidget.cxx
+++ b/Modules/Loadable/Terminologies/Widgets/qSlicerTerminologyNavigatorWidget.cxx
@@ -128,13 +128,17 @@ public:
   /// Reset current type modifier name and container object
   void resetCurrentTypeModifier();
 
-  // Set name from current selection and set it to name text box
+  /// Set name from current selection and set it to name text box
   void setNameFromCurrentTerminology();
-  // Set recommended color from current selection to color picker
+  /// Set recommended color from current selection to color picker
   void setRecommendedColorFromCurrentTerminology();
-  // Get recommended color stored in terminology.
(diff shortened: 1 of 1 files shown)


Modules/Loadable/Terminologies/Widgets/qSlicerTerminologyNavigatorWidget.cxx
```diff
@@ -1170,49 +1170,62 @@ void qSlicerTerminologyNavigatorWidget::populateTypeTable()
...
-    }
-    else
-    {
-      d->tableWidget_Type->setRowCount(types.size());
-    }
+  // Show none item only if search term is empty (if user is searching then they want an actual type)
+  int noneTypeExists = 0;
+  QTableWidgetItem* noneItem = nullptr;
+  if (searchTerm.empty())
+  {
```
lassoan, line 1179: indentation

lassoan (approved):
This looks great, thank you!

cpinter:
I fixed the indentation and added a check not to show color if the type did not have a color to begin with, like this
image

lassoan:
When does it happen that the type does not have a color? When it has a modifier? In that case, it would be better to show the first available color (usually they are the same or very similar for all modifiers).

cpinter:
Yes it happens when they have modifiers. I think I can get the first modifier. Not sure how much it will slow things down, probably not noticeably.

cpinter:
Done, the first modifier's color is used if found.

lassoan (changes requested):
This will be good, there is just a build error on linux according to CircleCI.

cpinter:
I fixed the error.

lassoan (approved):

lassoan:
Thank you!

PR #5417, opened 2021-01-29
ENH: Make slice views aligned with image axes by default

### LINKED ISSUE #5379: Make slice views aligned with image axes by default

New users keep struggling because slice views not aligned with image axes by default.

https://discourse.slicer.org/t/rotate-to-volume-plane/8032/2
https://discourse.slicer.org/t/nearby-fiducial-points-at-different-slices/15309/6
https://discourse.slicer.org/t/read-prostate-mr-image-failed/3423/8
https://discourse.slicer.org/t/slice-number-change/4358/2
https://discourse.slicer.org/t/segmentation-editor-how-to-disable-painting-on-adjacent-slices/1459/4
https://discourse.slicer.org/t/mri-dwi-images-load-with-wrong-orientation/482

Instead of trying to educate users about why Slicer behaves like this, we should change the behavior so that new users get what they expect. It should be easy to enable the current behavior (keep current slice view orientation, aligned with anatomical axes by default).

Add a "Reset orientation on show" checkbox (enabled by default for new installations) for visibility context menu in Data module, similarly to the existing "Reset field of view on show" context menu.

If this option is enabled then when the volume is propagated in slice views, eye icon is clicked in data tree for a volume, or volume drag-and-dropped into slice views, slice views would be reset to their default orientation (to make sure that different views would not be snapped to the same orientation) and then "rotate to volume plane" would be activated.

For single-slice volumes:
- If slice views are not linked, the slice could be added to the first slice view that does not show any volume yet.
- If slice views are linked then, the slice could be added to the first slice view group that does not show any volume yet.

I don't have any other good ideas, but open to suggestion.

## Additional context

Part of the "Make simple things simple and complex things possible" effort.

lassoan:
Users were confused by showing reformatted images by default, therefore image display actions are changed to facilitate axis-aligned display.
Slice views are rotated to be aligned with closest image axes if:
- Volume selection is propagated - for example when a volume is loaded or a CLI completes execution. Unless if a view is excluded by setting vtkMRMLSliceCompositeNode::DoPropagateVolumeSelection to false or "show" option is disable in Add data dialog.
- In Data module / Subject hierarchy tree -> eye icon is clicked. Unless "Reset view orientation on show" checkbox (checked by default) is unchecked in context menu of the eye icon.
- Dragging a volume from Data module / Subject hierarchy tree to a view. Unless "Reset view orientation on show" checkbox (checked by default) is unchecked in context menu of the eye icon.

fixes #5379

### DIFF
--- a/Libs/MRML/Core/vtkMRMLSliceNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLSliceNode.cxx
@@ -2187,8 +2187,10 @@ void vtkMRMLSliceNode::RotateToAxes(vtkMatrix4x4 *referenceToRAS, int sliceNorma
   // 8 combinations of X, Y vector directions, 2 vectors (slice X and Y)
   double sliceXAxisDirection[3] = { 0.0 };
   double sliceYAxisDirection[3] = { 0.0 };
+  double sliceZAxisDirection[3] = { 0.0 };
   vtkAddonMathUtilities::GetOrientationMatrixColumn(this->SliceToRAS, 0, sliceXAxisDirection);
   vtkAddonMathUtilities::GetOrientationMatrixColumn(this->SliceToRAS, 1, sliceYAxisDirection);
+  vtkAddonMathUtilities::GetOrientationMatrixColumn(this->SliceToRAS, 2, sliceZAxisDirection);
   double minAngleDiff = -1;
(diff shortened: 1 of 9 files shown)


pieper (approved):
I guess this makes sense - I'm very used to the other behavior but I can't really think of any way this will degrade typical usage.

lassoan:
I've played with it for a while and it feels right and we can fine-tune the behavior as needed.

PR #5421, opened 2021-01-31
STYLE: Use macOS in comments instead of old Mac OS X

jamesobutler:
This includes updates to use macOS term in code comments instead of older Mac OS X name.  There a few additional commits that remove workarounds for old mac versions.


pieper (approved):
👍

PR #5428, opened 2021-02-02
BUG: Avoid crash in DisplayableManager by checking for null scene pointer

jcfr:
This commit fixes crashes when using the python interactor by improving check for MRML scene validity in ProcessMRMLNodesEvent() and GetSelectionNode()


lassoan (commented):
Looks good, thank you!

I would just replace DM by "displayable manager" (or maybe "DisplayableManager") in the commit comment to make it easier to find this commit in the future by searching.

jcfr:
>  replace DM by "displayable manager" (or maybe "DisplayableManager") 

The pre-commit hook preventing me from doing so but I will describe it in the content

PR #5430, opened 2021-02-03
ENH: Add properties for easy visibility setting of SH tree columns

cpinter:
Instead of
`self.ui.SubjectHierarchyTreeView.setColumnHidden(self.ui.SubjectHierarchyTreeView.model().idColumn, True)`
we can now use
`self.ui.SubjectHierarchyTreeView.idColumnVisible = False`

Same for color and transform columns.

### DIFF
--- a/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyTreeView.cxx
+++ b/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyTreeView.cxx
@@ -1068,6 +1068,66 @@ QStringList qMRMLSubjectHierarchyTreeView::hideChildNodeTypes()const
   return this->sortFilterProxyModel()->hideChildNodeTypes();
 }
 
+//--------------------------------------------------------------------------
+void qMRMLSubjectHierarchyTreeView::setVisibilityColumnVisible(bool visible)
+{
+  this->setColumnHidden(this->model()->visibilityColumn(), !visible);
+}
+
+//--------------------------------------------------------------------------
+bool qMRMLSubjectHierarchyTreeView::visibilityColumnVisible()
+{
+  return !this->isColumnHidden(this->model()->visibilityColumn());
+}
+
+//--------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


lassoan (approved):
Very nice simplification, thank you.
Just for completeness, would you add visibility and and description columns as well?

cpinter:
Makes sense, I added the description column visibility property as well.

lassoan (dismissed):
>  visibility and and description

Was there any specific reason you did not include visibility column? Not a big issue, we can merge as is, but it would be nice to be able to consistently show/hide all volumes.

cpinter:
You're right there is no good reason. Initially I thought visibility is always needed and forgot about the new description column. But it's indeed better to include all of them for consistency.

Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyTreeView.cxx
```diff
@@ -1068,6 +1068,66 @@ QStringList qMRMLSubjectHierarchyTreeView::hideChildNodeTypes()const
...
+//--------------------------------------------------------------------------
+void qMRMLSubjectHierarchyTreeView::setVisibilityColumnVisible(bool visible)
+{
+  this->setColumnHidden(this->model()->visibilityColumn(), !visible);
+}
+
+//--------------------------------------------------------------------------
+bool qMRMLSubjectHierarchyTreeView::visibilityColumnVisible()
+{
+  return !this->isColumnHidden(this->model()->idColumn());
```
lassoan, line 1080: Copy-paste error
```suggestion
  return !this->isColumnHidden(this->model()->visibilityColumn());
```

lassoan (changes requested):
Thank you, almost there!

Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyTreeView.cxx
cpinter: Bah... ok :)

lassoan (approved):
Looks good, thank you!

PR #5431, opened 2021-02-03
BUG: Fix incorrect/missing volume node copy content macros

Sunderlandkyl:
In vtkMRMLVolumeNode, vtkMRMLVectorVolumeDisplayNode was used as the class name in vtkMRMLCopyContentMacro, rather than vtkMRMLVolumeNode.
Fixed by updating the macro with the correct class name.

vtkMRMLVectorVolumeNode was missing a vtkMRMLCopyContentDefaultMacro() so HasCopyContent() always returned false.
This prevented vector volume nodes from being displayed in the list of proxy nodes in the Sequences module.
Fixed by adding the missing macro.


jcfr:
Good catch. This change makes sense. 

Should we also fix this one?

https://github.com/Slicer/Slicer/blob/b00c14e21767dec1224c163b529dfc2d2784beb3/Libs/MRML/Core/vtkMRMLDiffusionImageVolumeNode.h#L47-L49

Sunderlandkyl:
Yes, I've added a fix for vtkMRMLDiffusionImageVolumeNode as well.

jcfr (approved):

PR #5435, opened 2021-02-04
COMP: Fix build error in cameras test

cpinter:
Related to https://github.com/Slicer/Slicer/pull/5410


pieper (approved):
👍

jcfr:
Thanks for the fix. Interestingly the use of `not` didn't trigger issue when using gcc on Linux. May be it is defined as a macro

PR #5440, opened 2021-02-07
Report data loading errors

lassoan:
Implement error reporting for loading data (similarly how it was recently implemented for *saving* data) and a two other minor fixes.

image

### DIFF
--- a/Libs/MRML/Core/vtkMRMLMessageCollection.cxx
+++ b/Libs/MRML/Core/vtkMRMLMessageCollection.cxx
@@ -23,6 +23,11 @@
 // MRML includes
 #include "vtkCommand.h"
 
+namespace
+{
+  const int SEPARATOR_MESSAGE_TYPE = vtkCommand::PropertyModifiedEvent;
+}
+
 //----------------------------------------------------------------------------
 vtkStandardNewMacro(vtkMRMLMessageCollection);
 
@@ -106,6 +111,12 @@ vtkMRMLMessageCollection
 }
 
 //----------------------------------------------------------------------------
+void vtkMRMLMessageCollection::AddSeparator()
+{
+  this->AddMessage(SEPARATOR_MESSAGE_TYPE, "\n--------\n");
+}
+
+//----------------------------------------------------------------------------
 void
 vtkMRMLMessageCollection
 ::ClearMessages()
@@ -116,11 +127,15 @@ vtkMRMLMessageCollection
 //----------------------------------------------------------------------------
 vtkMRMLMessageCollection::vtkMRMLMessageCollection()
 {
+  this->CallbackCommand = vtkSmartPointer<vtkCallbackCommand>::New();
+  this->CallbackCommand->SetCallback(vtkMRMLMessageCollection::CallbackFunction);
+  this->CallbackCommand->SetClientData(this);
 }
 
 //----------------------------------------------------------------------------
 vtkMRMLMessageCollection::~vtkMRMLMessageCollection()
 {
+  this->SetObservedObject(nullptr);
 }
 
 //----------------------------------------------------------------------------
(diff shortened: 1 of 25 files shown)


Libs/MRML/DisplayableManager/vtkMRMLViewInteractorStyle.cxx
```diff
@@ -366,9 +366,13 @@ bool vtkMRMLViewInteractorStyle::DelegateInteractionEventDataToDisplayableManage
...
-    //this->SetMouseCursor(VTK_CURSOR_DEFAULT);
-    this->DisplayableManagers->GetNthDisplayableManager(0)->SetMouseCursor(VTK_CURSOR_DEFAULT);
+    // None of the displayable managers can process the event, just ignore it.
+    // If click events (non-keyboard events) cannot be processed here then
+    // indicate this by setting the mouse cursor to default.
+    if (eventData->GetType() != vtkCommand::KeyPressEvent && eventData->GetType() != vtkCommand::KeyReleaseEvent)
+      {
+      this->DisplayableManagers->GetNthDisplayableManager(0)->SetMouseCursor(VTK_CURSOR_DEFAULT);
+      }
     return false;
```
pieper, line 376: Looks like a useful cleanup, but this is unrelated to the file io status.

Modules/Loadable/Markups/VTKWidgets/vtkSlicerCurveWidget.cxx
```diff
@@ -38,6 +38,9 @@ vtkSlicerCurveWidget::vtkSlicerCurveWidget()
   this->SetEventTranslationClickAndDrag(WidgetStateOnWidget, vtkCommand::RightButtonPressEvent, vtkEvent::AltModifier,
     WidgetStateScale, WidgetEventScaleStart, WidgetEventScaleEnd);
 
+  // Accept Ctrl+MouseMove (and process as simple mouse move) so that this widget keeps the focus when the user moves
+  // the mouse while holding down Ctrl key for inserting a point.
+  this->SetEventTranslation(WidgetStateOnWidget, vtkCommand::MouseMoveEvent, vtkEvent::ControlModifier, WidgetEventMouseMove);
```
pieper, line 43: Another change unrelated to file io

pieper (approved):
A couple unrelated changes - they look fine so it's okay with me to merge them with this PR or separate them out if you want.

pieper:
This is a really nice addition - thanks!

lassoan:
Thanks for the review, I'll separate those two unrelated changes into new commits.

lassoan:
I've checked and the two unrelated changes are in separate commits already.

I've just started to not push changes directly to the Slicer repository anymore so that I work the same way as all other non-admin developers. I did not want to create a new pull request for the two unrelated small fixes because I did not want to add too much noise, but maybe in the future I'll create separate pull request for each commit, no matter how small they are.

Libs/MRML/Core/vtkMRMLMessageCollection.cxx
```diff
@@ -23,6 +23,11 @@
 // MRML includes
 #include "vtkCommand.h"
 
+namespace
+{
+  const int SEPARATOR_MESSAGE_TYPE = vtkCommand::PropertyModifiedEvent;
```
jcfr, line 28: Is the use of `vtkCommand::PropertyModifiedEvent` arbitrary ?

lassoan: Yes, completely arbitrary, I just chose an event ID that has a name that makes some sense (separator is when we switch to another group, which is kind of moving to a different property) and not likely to be used otherwise.

lassoan: I'll add a comment to explain this in a follow-up commit.

PR #5441, opened 2021-02-08
BUG: Fix segment subject hierarchy visibility when display node modified

Sunderlandkyl:
The visibility of the segment subject hierarchy visibility was not being updated when segment display properties were modified.
Fixed by adding the slot qSlicerSubjectHierarchySegmentationsPlugin::onDisplayNodeModified() that is invoked when the display node is modified.
This slot updates all segment items from the corresponding display node.

See discussion here: https://discourse.slicer.org/t/visibility-of-segments-from-data-module-vs-segment-editor-segmentations/15888


lassoan (approved):
Looks good to me.
Thank you for the very quick fix!

PR #5442, opened 2021-02-08
ENH: Extend capabilities of subject hierarchy attribute filtering

cpinter:
In addition to allowing one attribute name (and value for that attribute) to be filtered, now multiple attributes can be specified that control which items show up in the subject hierarchy trees and comboboxes with more granularity.
- addItemAttributeFilter / addNodeAttributeFilter: add new attribute filter that can filter by attribute name and value, and class name in case of node filter
- removeItemAttributeFilter / removeNodeAttributeFilter: remove certain filters defined by only attribute name or each property
- includeItemAttributeNamesFilter: convenience function to show only items that contain any of the given attributes with this name (removes all previously added include item attribute filters)
- includeNodeAttributeNamesFilter: convenience function to show only items for data nodes that contain any of the given attributes with this name (removes all previously added include node attribute filters)
- excludeItemAttributeNamesFilter: convenience function to hide items that contain any of the given attributes with this name (removes all previously added exclude item attribute filters)
- excludeNodeAttributeNamesFilter: convenience function to hide items for data nodes that contain any of the given attributes with this name (removes all previously added exclude node attribute filters)
Each of these are disabled (empty string list) by default. The legacy filters were kept for backward compatibility:
- attributeNameFilter: show only items that contain an attribute with this name
- attributeValueFilter: show only items that contain the attribute specified in attributeNameFilter and have a given value

Test case added to SubjectHierarchy/Testing/Python/SubjectHierarchyGenericSelfTest.py

Add two convenience functions to slicer.util for getting SH children items: getSubjectHierarchyItemChildren and getSubjectHierarchyItemAllChildren

### DIFF
--- a/Base/Python/slicer/util.py
+++ b/Base/Python/slicer/util.py
@@ -1188,6 +1188,31 @@ class NodeModify(object):
   def __exit__(self, type, value, traceback):
     self.node.EndModify(self.wasModifying)
 
+
+
+#
+# Subject hierarchy
+#
+def getSubjectHierarchyItemChildren(parentItem=None, recursive=False):
+  """Convenience method to get children of a subject hierarchy item.
+  :param vtkIdType parentItem: Item for which to get children for. If omitted
+         or None then use scene item (i.e. get all items)
+  :param bool recursive: Whether the query is recursive. False by default
+  :return: List of child item IDs
+  """
+  children = []
+  shNode = slicer.mrmlScene.GetSubjectHierarchyNode()
+  # Use scene as parent item if not given
+  if not parentItem:
+    parentItem = shNode.GetSceneItemID()
+  childrenIdList = vtk.vtkIdList()
+  shNode.GetItemChildren(parentItem, childrenIdList, recursive)
+  for childIndex in range(childrenIdList.GetNumberOfIds()):
+    children.append(childrenIdList.GetId(childIndex))
+  return children
+
+
+
 #
 # MRML-numpy
 #
--- a/Modules/Loadable/SubjectHierarchy/Testing/Python/SubjectHierarchyGenericSelfTest.py
+++ b/Modules/Loadable/SubjectHierarchy/Testing/Python/SubjectHierarchyGenericSelfTest.py
@@ -70,7 +70,7 @@ class SubjectHierarchyGenericSelfTestTest(ScriptedLoadableModuleTest):
   def setUp(self):
     """ Do whatever is needed to reset the state - typically a scene clear will be enough.
     """
-    slicer.mrmlScene.Clear(0)
+    slicer.mrmlScene.Clear()
 
     self.delayMs = 700
 
@@ -101,10 +101,10 @@ class SubjectHierarchyGenericSelfTestTest(ScriptedLoadableModuleTest):
     self.section_AddNodeToSubjectHierarchy()
     self.section_CLI()
     self.section_CreateSecondBranch()
-    self.section_AttributeFilter()
     self.section_ReparentNodeInSubjectHierarchy()
     self.section_LoadScene()
     self.section_TestCircularParenthood()
+    self.section_AttributeFilters()
 
     logging.info('Test finished')
 
@@ -127,7 +127,11 @@ class SubjectHierarchyGenericSelfTestTest(ScriptedLoadableModuleTest):
     self.dicomZipFilePath = subjectHierarchyGenericSelfTestDir + '/TestDicomCT.zip'
     self.expectedNumOfFilesInDicomDataDir = 10
     self.tempDir = subjectHierarchyGenericSelfTestDir + '/Temp'
-    self.sceneFileName = self.tempDir + '/SubjectHierarchyGenericSelfTestScene.mrml'
+    self.genericTestSceneFileName = self.tempDir + '/SubjectHierarchyGenericSelfTestScene.mrml'
+
+    self.attributeFilterTestSceneFileUrl = TESTING_DATA_URL + 'SHA256/83e0df42d178405dccaf5a87d0661dd4bad71b535c6f15457344a71c4c0b7984'
+    self.attributeFilterTestSceneChecksum = 'SHA256:83e0df42d178405dccaf5a87d0661dd4bad71b535c6f15457344a71c4c0b7984'
+    self.attributeFilterTestSceneFileName = 'SubjectHierarchyAttributeFilterTestScene.mrb'
 
     self.invalidItemID = slicer.vtkMRMLSubjectHierarchyNode.GetInvalidItemID()
 
@@ -160,7 +164,7 @@ class SubjectHierarchyGenericSelfTestTest(ScriptedLoadableModuleTest):
     self.delayDisplay("Clear scene",self.delayMs)
 
     # Clear the scene to make sure there is no crash (closing scene is a sensitive operation)
-    slicer.mrmlScene.Clear(0)
+    slicer.mrmlScene.Clear()
 
     # Make sure there is only one subject hierarchy node after closing the scene
     self.assertEqual( slicer.mrmlScene.GetNumberOfNodesByClass('vtkMRMLSubjectHierarchyNode'), 1 )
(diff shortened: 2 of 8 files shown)


Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyComboBox.h
```diff
@@ -57,6 +57,26 @@ class Q_SLICER_MODULE_SUBJECTHIERARCHY_WIDGETS_EXPORT qMRMLSubjectHierarchyCombo
...
+  /// Filter to hide items that contain any of the given attributes with this name. Empty by default
+  /// Overrides \sa includeItemAttributeNamesFilter
+  Q_PROPERTY(QStringList excludeItemAttributeNamesFilter READ excludeItemAttributeNamesFilter WRITE setExcludeItemAttributeNamesFilter)
+  /// Filter to hide items for data nodes that contain any of the given attributes with this name. Empty by default
+  /// Overrides \sa includeNodeAttributeNamesFilter
+  Q_PROPERTY(QStringList excludeNodeAttributeNamesFilter READ excludeNodeAttributeNamesFilter WRITE setExcludeNodeAttributeNamesFilter)
+
+  /// Filter to show only items that contain an attribute with this name. Empty by default
+  /// Note: Deprecated, kept only for backwards compatibility. Sets and returns the first attribute in \sa includeNodeAttributeNamesFilter
+  Q_PROPERTY(QString attributeNameFilter READ attributeNameFilter WRITE setAttributeNameFilter)
```
lassoan, line 73: I think converting something to a property changes how it can be accessed from Python (you cannot access it as class.member() and class.setMember() anymore, only as class.member). So, this change would break existing modules. It would be nice to make this change for consistency, but it could be better to leave it as is for backward compatibility.

Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyComboBox.h
```diff
@@ -57,6 +57,26 @@ class Q_SLICER_MODULE_SUBJECTHIERARCHY_WIDGETS_EXPORT qMRMLSubjectHierarchyCombo
...
+  /// Filter to show only items that contain any attribute given in \sa includeItemAttributeNamesFilter with the value.
+  /// If empty, then existence of the attributes is enough to show.
+  /// Exact match is required. Empty by default
+  /// Note: Deprecated, kept only for backwards compatibility. Works consistently with the previous operation.
+  Q_PROPERTY(QString attributeValueFilter READ attributeValueFilter WRITE setAttributeValueFilter)
```
lassoan, line 78: We cannot deprecate this. qMRMLNodeCombobox::addAttribute (http://apidocs.slicer.org/master/classqMRMLNodeComboBox.html#a056b54be78c6053b6fd21d51b2b1d045) proven to work be a very simple and powerful way of filtering nodes. In the future we would like to switch over from qMRMLNodeCombobox to qMRMLSubjectHierarchyComboBox (by making qMRMLNodeCombobox an alias to qMRMLSubjectHierarchyComboBox), so it would be important to get closer to feature parity and not get farther from it.

It could be good to have various convenience functions for setting filtering parameters (e.g., set several attribute filters with as a stringlist), but ultimately I think we cannot avoid having filters specified with:
- QString ClassName
- QString AttributeName
- QVariant AttributeValue
- bool Include

The first 3 same as qMRMLNodeComboBox::addAttribute, the last selects between include/exclude as you have nicely implemented.

lassoan (commented):
Thank you, this will be great, we should just make sure we move towards reaching feature parity with qMRMLNodeComboBox instead of getting farther from it. We are quite close to being able to replace qMRMLNodeComboBox by qMRMLSubjectHierarchyComboBox.

Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyComboBox.h
cpinter: It has already been a property.

Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyComboBox.h
cpinter: I implemented the filters as you suggest, and added extra functions so that the additional granularity can be used.

Base/Python/slicer/util.py
```diff
@@ -1188,6 +1188,39 @@ def __enter__(self):
...
+
+#
+# Subject hierarchy
+#
+def getSubjectHierarchyItemChildren(parentItem=None, recursive=False):
+  """Convenience method to get children of a subject hierarchy item.
+  :param vtkIdType parentItem: Item for which to get children for. If omitted
+         or None then use scene item (i.e. get all items)
+  :param bool recursive: Whether the query is recursive. False by default
+  :return: List of child items
```
lassoan, line 1201:
```suggestion
  :return: List of child item IDs
```

Base/Python/slicer/util.py
```diff
@@ -1188,6 +1188,39 @@ def __enter__(self):
...
+  # Use scene as parent item if not given
+  if not parentItem:
+    parentItem = shNode.GetSceneItemID()
+  childrenIdList = vtk.vtkIdList()
+  shNode.GetItemChildren(parentItem, childrenIdList, recursive)
+  for childIndex in range(childrenIdList.GetNumberOfIds()):
+    children.append(childrenIdList.GetId(childIndex))
+  return children
+
+def getSubjectHierarchyItemAllChildren(parentItem=None):
```
lassoan, line 1214: Do we really need this? Users could simply call `getSubjectHierarchyItemChildren(parentItem, True)` themselves.

Modules/Loadable/SubjectHierarchy/Testing/Python/SubjectHierarchyGenericSelfTest.py
```diff
@@ -428,9 +410,135 @@ def section_TestCircularParenthood(self):
...
+    shTreeView.show()
+
+    shProxyModel = shTreeView.sortFilterProxyModel()
+
+    def testAttributeFilters(filteredObject, proxyModel):
+      # Check include node attribute name filter
+      self.assertEqual(shProxyModel.acceptedItemCount(shNode.GetSceneItemID()), 9)
+      filteredObject.includeNodeAttributeNamesFilter = ['Markups.MovingInSliceView']
+      self.assertEqual(shProxyModel.acceptedItemCount(shNode.GetSceneItemID()), 6)
+      filteredObject.addNodeAttributeFilter('Sajt')
```
lassoan, line 453: It is funny, but maybe it is better to change it to something that conveys the intent better for English-speaking readers ('dummy' or 'nonexisting'), or add a comment to the line that it is just some random string (corresponding to a non-existing  attribute name)

Modules/Loadable/SubjectHierarchy/Testing/Python/SubjectHierarchyGenericSelfTest.py
```diff
@@ -428,9 +410,135 @@ def section_TestCircularParenthood(self):
...
+
+    # Create test SH tree view
+    shTreeView = slicer.qMRMLSubjectHierarchyTreeView()
+    shTreeView.setMRMLScene(slicer.mrmlScene)
+    shTreeView.show()
+
+    shProxyModel = shTreeView.sortFilterProxyModel()
+
+    def testAttributeFilters(filteredObject, proxyModel):
+      # Check include node attribute name filter
```
lassoan, line 449: It could be nice if you could add here in a comment attributes of those 9 nodes so that the test can be verified by code review.

lassoan (dismissed):
This is very nice! I only added a few nitpicks as comments.

Modules/Loadable/SubjectHierarchy/Testing/Python/SubjectHierarchyGenericSelfTest.py
cpinter: I realized this as well. I created a test scene first to test the filters, but then the test snippet became the automatic test code and the scene the test scene, and now it would be a hassle replacing the file. If this is not too big a problem I would keep it as is.

Base/Python/slicer/util.py
cpinter: I added this because if we call it without any argument then we get all SH items. If you want I can remove it.

lassoan: I would vote for removing it, just to keep the API a bit smaller (it is already quite big, and growing).

Modules/Loadable/SubjectHierarchy/Testing/Python/SubjectHierarchyGenericSelfTest.py
lassoan: OK, just add a comment why "Sajt" (is there an attribute like this in the scene?)

Base/Python/slicer/util.py
cpinter: OK one minute.

Modules/Loadable/SubjectHierarchy/Testing/Python/SubjectHierarchyGenericSelfTest.py
cpinter: I added an ASCII tree as comment in this function

lassoan (approved):
Looks good, thank you.

cpinter:
Thank you for the review!

PR #5448, opened 2021-02-11
COMP: Add aarch64 architecture to Slicer operating system macro

adamrankin:
Working towards a aarch64 build of Slicer. This macro is used (among many places probably) to craft the URL to download the AppLauncher binary


jcfr:
Thanks for the contribution, I suggest we move forward with integration after this issue is addressed: https://github.com/commontk/AppLauncher/issues/120

How does that sound ?

PR #5450, opened 2021-02-11
ENH: Performance improvements

lassoan:
These changes make scene loading many times faster if there are hundreds of nodes in the scene.

- Pause rendering during batch processing and data loading for better performance. When the operation is completed, rendering is resumed and any pending rendering requests are processed.
- Make show/hide model slice intersections faster by not removing and re-adding all actors, but only removing/adding those that have changed.


pieper (approved):
This is great!  This will help SlicerDMRI immediately.

PR #5451, opened 2021-02-11
ENH: Improve module access scripts in slicer.util

lassoan:
Add function to return module logic.
Improve module widget function to return the actual Python class instead of the C++ wrapper.
Use exceptions for error handling (instead of just printing error message) to make sure that the error is noticed and the most relevant error message is displayed.
Add missing function documentation.

### DIFF
--- a/Base/Python/slicer/util.py
+++ b/Base/Python/slicer/util.py
@@ -194,8 +194,7 @@ def lookupTopLevelWidget(objectName, verbose = True):
     if hasattr(w,'objectName'):
       if w.objectName == objectName: return w
   if verbose:
-    import sys
-    print("Failed to obtain reference to '%s'" % objectName, file=sys.stderr)
+    raise RuntimeError("Failed to obtain reference to '%s'" % objectName)
   return None
 
 def mainWindow(verbose = True):
@@ -205,7 +204,7 @@ def pythonShell(verbose = True):
   from slicer import app
   console = app.pythonConsole()
   if not console and verbose:
-    print("Failed to obtain reference to python shell", file=sys.stderr)
+    raise RuntimeError("Failed to obtain reference to python shell")
   return console
 
 def showStatusMessage(message, duration = 0):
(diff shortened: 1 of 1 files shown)


pieper (approved):
Is there any compatibility issue related to returning the python class compared to the c++ wrapper?

lassoan:
Yes, the behavior is now slightly different, but probably very few people used these functions, as they were not advertised anywhere, not documented, and they did not do that much (they just replaced a single-line operation with a single command). Now that it can do a bit more, it makes more sense to recommend it over the basic C++ API in examples and tutorials.

Base/Python/slicer/util.py
```diff
@@ -194,8 +194,7 @@ def lookupTopLevelWidget(objectName, verbose = True):
     if hasattr(w,'objectName'):
       if w.objectName == objectName: return w
   if verbose:
-    import sys
-    print("Failed to obtain reference to '%s'" % objectName, file=sys.stderr)
+    raise RuntimeError("Failed to obtain reference to '%s'" % objectName)
```
jcfr, line 197: Now we raise an exception. It should probably be done all the time and not only when verbose is True.

Base/Python/slicer/util.py
```diff
@@ -205,7 +204,7 @@ def pythonShell(verbose = True):
   from slicer import app
   console = app.pythonConsole()
   if not console and verbose:
-    print("Failed to obtain reference to python shell", file=sys.stderr)
+    raise RuntimeError("Failed to obtain reference to python shell")
```
jcfr, line 207: Same comment here.

Base/Python/slicer/util.py
```diff
@@ -928,64 +927,131 @@ def saveScene(filename, properties={}):
 #
 
 def moduleSelector():
+  """Return module selector widget.
+  Throws a RuntimeError exception if there is no module selector (for example, the application runs without a main window).
```
jcfr, line 931: There is a way to document exception. See https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html#the-sphinx-docstring-format

Base/Python/slicer/util.py
```diff
@@ -928,64 +927,131 @@ def saveScene(filename, properties={}):
...
   module = app.moduleManager().module(moduleName)
   if not module:
-    import sys
-    print("Could not find module with name '%s" % moduleName, file=sys.stderr)
-    return None
+    raise RuntimeError("Could not find module with name '%s'" % moduleName)
   return module
 
 def getModuleGui(module):
+  """Deprecated. Use getModuleWidget function instead"""
```
jcfr, line 981: I suggest to use the 'deprecated directive.  See https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-deprecated

Base/Python/slicer/util.py
```diff
@@ -928,64 +927,131 @@ def saveScene(filename, properties={}):
...
+      # This widget does not have a logic instance in its widget class
+      logic = None
+  else:
+    logic = module.logic()
+  if not logic:
+    raise RuntimeError("Could not find module widget representation with name '%s'" % module.name)
+  return logic
 
 def modulePath(moduleName):
+  """Get module logic object.
```
jcfr, line 1049: This docstring seems incorrect

PR #5452, opened 2021-02-11
BUG: Fix vtkMRMLVolumeRenderingDisplayableManagerTest1

Sunderlandkyl:
Volume rending logic switched from directly instantiating an ROI node to using AddNewNodeByClass.
Fixed test by registering the ROI nodes to the test scene.


lassoan (approved):
Looks good, thank you!

PR #5453, opened 2021-02-11
ENH: Observe markup measurements to trigger necessary updates

cpinter:
By observing the Modified event of each measurement in the markup nodes, updating the measurements is possible when any of them are e.g. enabled programmatically.

### DIFF
--- a/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsNode.cxx
+++ b/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsNode.cxx
@@ -19,14 +19,13 @@
 
 // MRML includes
 #include "vtkCurveGenerator.h"
+#include <vtkEventBroker.h>
 #include "vtkMRMLMarkupsDisplayNode.h"
 #include "vtkMRMLMarkupsStorageNode.h"
 #include "vtkMRMLStaticMeasurement.h"
 #include "vtkMRMLSelectionNode.h"
 #include "vtkMRMLTransformNode.h"
 #include "vtkMRMLUnitNode.h"
-
-// Slicer MRML includes
 #include "vtkMRMLScene.h"
 
 // vtkAddon includes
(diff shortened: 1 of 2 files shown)


lassoan (commented):
This will be good, but I think adding/removing observations could be made a bit more clear. The main problem now is that observers are not removed when a measurement is removed, so the measurements hangs around (kept alive by the observer, because event observation increases reference count on the observed object). An elegant solution would be to use vtkEventBroker to manage observations, because that can automatically remove observations when the observed object gets deleted.

cpinter:
Right, removing them is not implemented. I just took care of removing them when quitting. OK I'll look into improving this.

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsNode.cxx
```diff
@@ -103,6 +103,8 @@ vtkMRMLMarkupsNode::~vtkMRMLMarkupsNode()
 
   if (this->Measurements)
     {
+    this->Measurements->RemoveAllItems(); // Trigger modified event to unobserve measurements
```
lassoan, line 106: Do we still need this line? Usually we want to avoid firing modified events in destructors because it can cause a cascade of calls that may assume that this object is still valid while in fact it may be already partially destructed.

lassoan (commented):
Looks good to me. Have you checked that the measurement object gets deleted when it is removed from the collection?

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsNode.cxx
cpinter: There is no leak on exit when deleted so I think it's not needed. I removed it.

cpinter:
> Have you checked that the measurement object gets deleted when it is removed from the collection?

Yes now I have.

lassoan (approved):
Thank you!

PR #5454, opened 2021-02-11
ENH: Miscellaneous ROI display improvements

Sunderlandkyl:
List of improvements:
- Default fill opacity set to 20%
- Disable fill visibility for volume rendering by default
- Add outline in 3D views
- Remove out of plane ROI outline in 2D
- Increase interaction handle size by 2x in 3D views
- Enable right-click menu for scale handles
- Add interaction handle visibility context menu appear in subject hierarchy widget visibility column
- Hide interaction handles when out of plane
- Enable ROI interaction handles by default


Sunderlandkyl:
Here is what the markups ROI currently looks like in 2D and 3D:
image

lassoan (approved):
Looks good to me. Thanks a lot for the quick fixes/improvements!

PR #5457, opened 2021-02-13
ENH: Remove deprecated sitkUtils API

jamesobutler:
@lassoan You marked these methods deprecated on August 10th, 2017. Considering it has now been 3.5 years since then and the deprecation was originally present in stable release 4.8.0 (released October 18th 2017) I think it has been enough time to execute on the removal of the code.


lassoan (approved):
Thank you for the PR. I agree that these can go now.

Issue #4652, opened 2020-03-13
Runtime errors due to failure to create temp folder


slicerbot:
_This issue was created automatically from an original [Mantis Issue](https://mantisarchive.slicer.org/view.php?id=4652). Further discussion may take place here._

lassoan:
If temporary path is not writable then a new temporary path is generated.

PR #5458, opened 2021-02-13
BUG: Ensure temporaryPath is writable

lassoan:
To ensure that CLI modules, sample data downloads, etc. work it is necessary to have a writable temporary directory.
In some cases the path specified in application settings does not point to a valid, writable location anymore.

To fix this, the application checks at startup that the temporary path is writeable. If the check fails then the path is reset to the default.

see #4652

### DIFF
--- a/Base/QTCore/qSlicerCoreApplication.cxx
+++ b/Base/QTCore/qSlicerCoreApplication.cxx
@@ -322,6 +322,17 @@ void qSlicerCoreApplicationPrivate::init()
   modifiedRequestCallback->SetCallback(vtkSlicerApplicationLogic::RequestModifiedCallback);
   vtkEventBroker::GetInstance()->SetRequestModifiedCallback(modifiedRequestCallback);
 
+  // Check if temporary folder is writeable
+  QTemporaryFile fileInTemporaryPathFolder(
+    QFileInfo(q->temporaryPath(), "_write_test_XXXXXX.tmp").absolutePath());
+  if (!fileInTemporaryPathFolder.open())
+    {
+    QString newTempFolder = q->defaultTemporaryPath();
+    qWarning() << Q_FUNC_INFO << "Setting temporary folder to " << newTempFolder
+      << " because previously set " << q->temporaryPath() << " folder was not writeable";
+    q->setTemporaryPath(newTempFolder);
+    }
+
   this->AppLogic->SetTemporaryPath(q->temporaryPath().toUtf8());
   vtkPersonInformation* userInfo = this->AppLogic->GetUserInformation();
   if (userInfo)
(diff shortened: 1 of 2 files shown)


Base/QTCore/qSlicerCoreApplication.cxx
```diff
@@ -322,6 +322,17 @@ void qSlicerCoreApplicationPrivate::init()
...
   vtkEventBroker::GetInstance()->SetRequestModifiedCallback(modifiedRequestCallback);
 
+  // Check if temporary folder is writeable
+  QTemporaryFile fileInTemporaryPathFolder(
+    QFileInfo(q->temporaryPath(), "_write_test_XXXXXX.tmp").absolutePath());
+  if (!fileInTemporaryPathFolder.open())
+    {
+    QString newTempFolder = q->defaultTemporaryPath();
+    qWarning() << Q_FUNC_INFO << "Setting temporary folder to " << newTempFolder
+      << " because previously set " << q->temporaryPath() << " folder was not writeable";
```
jcfr, line 332:
```suggestion
      << " because previously set " << q->temporaryPath() << " folder is not writable";
```

Base/QTGUI/qSlicerSettingsModulesPanel.cxx
```diff
@@ -161,7 +161,7 @@ void qSlicerSettingsModulesPanelPrivate::init()
                       "showModules", SIGNAL(showModulesChanged(QStringList)));
   qSlicerRelativePathMapper* relativePathMapper = new qSlicerRelativePathMapper(
     this->TemporaryDirectoryButton, "directory", SIGNAL(directoryChanged(QString)));
-  q->registerProperty("Modules/TemporaryDirectory", relativePathMapper,
+  q->registerProperty("TemporaryPath", relativePathMapper,
```
jcfr, line 164: I suggest we add an entry to  https://www.slicer.org/wiki/Documentation/Nightly/Developers/Tutorials/MigrationGuide. 

For example, a category called `Slicer 5.0: API changes since 4.10 / Settings` could  be added.

lassoan: Added https://www.slicer.org/wiki/Documentation/Nightly/Developers/Tutorials/MigrationGuide/Slicer#Slicer_5.0_:_Temporary_path. If we have more settings changes then we may merge them under a Settings section.

jcfr: Thanks. I tweaked formatting using `<code>...</code>`

jcfr (approved):

lassoan:
Thanks for all the thorough reviews.

PR #5459, opened 2021-02-13
ENH: Move unit and recent log file paths storage to common application settings

lassoan:
Units and recent log file paths were stored in the revision-specific application settings, which was not ideal because it generated warning messages when Slicer is launched from a read-only location (and revision-specific settings file is not created in the user profile folder).

see #4966

### DIFF
--- a/Base/QTGUI/qSlicerApplication.cxx
+++ b/Base/QTGUI/qSlicerApplication.cxx
@@ -797,12 +797,12 @@ int qSlicerApplication::numberOfRecentLogFilesToKeep()
 {
   Q_D(qSlicerApplication);
 
-  QSettings* revisionUserSettings = this->revisionUserSettings();
+  QSettings* userSettings = this->userSettings();
 
   // Read number of log files to store value. If this value is missing,
   // then the group considered non-existent
   bool groupExists = false;
-  int numberOfFilesToKeep = revisionUserSettings->value(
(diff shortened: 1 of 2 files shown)


jcfr:
I am wondering if we should make the display of log file list and associated a bit smarter then. 

For example, we could only list/remove log files that match name and revision.

lassoan:
> I am wondering if we should make the display of log file list and associated a bit smarter then.
> 
> For example, we could only list/remove log files that match name and revision.

Application name is matched via the ini file name. All revisions are indeed listed, but this is not necessarily bad (you can get log file of a crashing revision by running a non-crashing revision). The log file list was already shared between instances of the same revision, so the behavior is not fundamentally different, there is just no filtering by revision.

jcfr:
> Application name is matched via the ini file name. All revisions are indeed listed, but this is not necessarily bad (you can get log file of a crashing revision by running a non-crashing revision). The log file list was already shared between instances of the same revision, so the behavior is not fundamentally different, there is just no filtering by revision.

Good point. And the default number of file is initialized to `10` which seems reasonable

See https://github.com/commontk/CTK/blob/f670e1d0e513bae9d08257b883aaf669f546053d/Libs/Core/ctkFileLogger.cpp#L49-L54

jcfr (approved):

PR #5460, opened 2021-02-13
STYLE: Make slicer.util API documentation more consistent

lassoan:
Use standard sphinx syntax for exceptions and deprecation.
Remove verbose parameter (always throw exception instead).

Based on feedback from @jcfr - https://github.com/Slicer/Slicer/pull/5451#pullrequestreview-589153421

### DIFF
--- a/Base/Python/slicer/util.py
+++ b/Base/Python/slicer/util.py
@@ -206,13 +206,13 @@ def lookupTopLevelWidget(objectName):
   raise RuntimeError("Failed to obtain reference to '%s'" % objectName)
 
 def mainWindow():
-  """Get Python console widget (ctkPythonConsole object)
+  """Get main window widget (qSlicerMainWindow object)
 
-  :return: main window object, or None if there is no main window
+  :return: main window widget, or ``None`` if there is no main window
   """
   try:
     mw = lookupTopLevelWidget('qSlicerMainWindow')
-  except:
+  except RuntimeError:
     # main window not found, return None
     # Note: we do not raise an exception so that this function can be conveniently used
     # in expressions such as `parent if parent else mainWindow()`
@@ -231,7 +231,7 @@ def pythonShell():
   return console
 
 def showStatusMessage(message, duration = 0):
-  """Display message in the status bar.
+  """Display ``message`` in the status bar.
   """
   mw = mainWindow()
   if not mw or not mw.statusBar:
@@ -362,7 +362,7 @@ def addParameterEditWidgetConnections(parameterEditWidgets, updateParameterNodeF
 
   The function is useful for calling updateParameterNodeFromGUI method in scripted module widgets.
 
-  Note: Not all widget classes are supported yet. Report any missing classes at discourse.slicer.org.
+  .. note:: Not all widget classes are supported yet. Report any missing classes at https://discourse.slicer.org.
 
   Example::
 
@@ -585,7 +585,7 @@ def setStatusBarVisible(visible):
   If there is no main window or status bar then the function has no effect.
   """
   mw = mainWindow()
-  if not mw:
+  if mw is None:
     return
   try:
     statusBar = mw.statusBar()
@@ -1209,7 +1209,7 @@ def setModulePanelTitleVisible(visible):
   If there is no main window then the function has no effect.
   """
   mw = mainWindow()
-  if not mw:
+  if mw is None:
     return
   modulePanelDockWidget = mw.findChildren('QDockWidget','PanelDockWidget')[0]
   if visible:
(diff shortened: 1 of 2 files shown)


Base/Python/slicer/util.py
```diff
@@ -529,20 +557,42 @@ def setToolbarsVisible(visible, ignore=None):
...
-  mainWindow(verbose=False).statusBar().setVisible(visible)
+  """Show/hide status bar
+
+  If there is no main window or status bar then the function has no effect.
+  """
+  mw = mainWindow()
+  if not mw:
+    return
+  try:
+    statusBar = mw.statusBar()
```
jcfr, line 591: this try/except is not needed. From the Qt documentation:

```
Returns the status bar for the main window. This function creates and returns an empty status bar if the status bar does not exist.
```

Base/Python/slicer/util.py
```diff
@@ -577,6 +627,7 @@ def forceRenderAllViews():
 def loadNodeFromFile(filename, filetype, properties={}, returnNode=False):
   """Load node into the scene from a file.
 
+  :raises RuntimeError: in case of failure
```
jcfr, line 630: Let's add `:raises RuntimeError: in case of failure` at the end after return (usual convention)

Base/Python/slicer/util.py
```diff
@@ -1120,43 +1202,66 @@ def reloadScriptedModule(moduleName):
...
+
   If the title bar is not visible then it is not possible to drag and dock the
-  module panel to a different location."""
-  modulePanelDockWidget = mainWindow().findChildren('QDockWidget','PanelDockWidget')[0]
+  module panel to a different location.
+
+  If there is no main window then the function has no effect.
+  """
+  mw = mainWindow()
+  if not mw:
```
jcfr, line 1212:
```suggestion
  if mw is None:
```

Base/Python/slicer/util.py
```diff
@@ -1120,43 +1202,66 @@ def reloadScriptedModule(moduleName):
...
 
 def setApplicationLogoVisible(visible):
-  """Show/hide application logo at the top of module panel."""
-  widget = findChild(mainWindow(), "LogoLabel")
+  """Show/hide application logo at the top of module panel.
+
+  If there is no main window then the function has no effect.
+  """
+  mw = mainWindow()
+  if not mw:
```
jcfr, line 1227:
```suggestion
  if mw is None:
```

Base/Python/slicer/util.py
```diff
@@ -1120,43 +1202,66 @@ def reloadScriptedModule(moduleName):
...
 
 def setModuleHelpSectionVisible(visible):
-  """Show/hide Help section at the top of module panel."""
-  modulePanel = findChild(mainWindow(), "ModulePanel")
+  """Show/hide Help section at the top of module panel.
+
+  If there is no main window then the function has no effect.
+  """
+  mw = mainWindow()
+  if not mw:
```
jcfr, line 1238:
```suggestion
  if mw is None:
```

Base/Python/slicer/util.py
```diff
@@ -1120,43 +1202,66 @@ def reloadScriptedModule(moduleName):
...
 
 def setDataProbeVisible(visible):
-  """Show/hide Data probe at the bottom of module panel."""
-  widget = findChild(mainWindow(), "DataProbeCollapsibleWidget")
+  """Show/hide Data probe at the bottom of module panel.
+
+  If there is no main window then the function has no effect.
+  """
+  mw = mainWindow()
+  if not mw:
```
jcfr, line 1249:
```suggestion
  if mw is None:
```

Base/Python/slicer/util.py
jcfr: If we keep the `try/except`, I suggest we use a specific exception

Base/Python/slicer/util.py
lassoan: According to the documentation that you linked in the previous code review, "raises" is just _before_ the :return:. I've updated the code now accordingly. Is this OK, or I should move "raises" just _after_ the :return:?

Base/Python/slicer/util.py
lassoan: fixed

Base/Python/slicer/util.py
jcfr: Ideally, it should have the following form:

```

def public_fn_with_sphinxy_docstring(name, state=None):
    """This function does something.

    :param name: The name to use.
    :type name: str.
    :param state: Current state to be in.
    :type state: bool.
    :returns:  int -- the return code.
    :raises: AttributeError, KeyError

    """
    return 0

```

Copied from  https://pythonhosted.org/an_example_pypi_project/sphinx.html

lassoan: Done - "raises" now comes after "returns". All the other conversations in this PR are resolved, so if this looks OK then please approve. Thanks for the review!

lassoan:
I've merged this, as it also fixes a test error (introduced by the previous commit).

PR #5462, opened 2021-02-15
BUG: fix tmp file path calculation

pieper:
See discussion here:

https://github.com/Slicer/Slicer/commit/d518c36aca64583d003b3a405f2308aceedf65db#commitcomment-47154875

Co-authored-by: Andras Lasso <lassoan@queensu.ca>


jamesobutler:
@pieper This change would be needed to make it show as being co-authored here before merging
```
BUG: fix tmp file path calculation 

See discussion here:

d518c36#commitcomment-47154875
```
should be 
```
BUG: fix tmp file path calculation 

See discussion here:

d518c36#commitcomment-47154875

Co-authored-by: Andras Lasso lassoan@queensu.ca
```

pieper:
Yes, I know, but dinner was served : )

PR #5466, opened 2021-02-17
BUG: Fix python wrapping of vtkSlicerCLIModuleLogic using VTK9

### LINKED ISSUE #5266: vtkSlicerCLIModuleLogic Python class is not found
Using Slicer-4.11.20200930:

```ptyhon
>>> slicer.vtkSlicerCLIModuleLogic()
(qSlicerBaseQTCLIPython.vtkSlicerCLIModuleLogic)00000167BA933648
```

Using latest master:

```python
>>> slicer.vtkSlicerCLIModuleLogic()
TypeError: this class cannot be instantiated
```

I’ve debugged into wrappython.exe and the problem is that for some reason the hierarchy cannot find the parent’s class in the hierarchy and so it determines that vtkSlicerCLIModuleLogic is not a VTK class, therefore it looks for a public constructor (instead of New()), which is of course is not found.

Originally reported here: https://discourse.slicer.org/t/calls-to-cli-makes-tests-to-fail/13835/10?u=lassoan

jcfr:
Fixes #5266

### DIFF
--- a/Base/QTCLI/CMakeLists.txt
+++ b/Base/QTCLI/CMakeLists.txt
@@ -94,6 +94,14 @@ set(KIT_target_libraries
   ModuleDescriptionParser ${ITK_LIBRARIES}
   MRMLCLI
   )
+if(VTK_WRAP_PYTHON AND ${VTK_VERSION} VERSION_GREATER_EQUAL "8.90")
+  # HACK Explicitly list transitive VTK dependencies because _get_dependencies_recurse
+  # used in vtkAddon/CMake/vtkMacroKitPythonWrap.cmake only recurses over dependencies
+  # that are VTK python wrapped.
+  list(APPEND KIT_target_libraries
+    SlicerBaseLogic
+    )
+endif()
 
 if(Slicer_USE_QtTesting)
   list(APPEND KIT_SRCS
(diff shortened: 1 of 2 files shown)


lassoan (approved):
Thank you, it looks good. What is the reason this has only came up for VTK9?

jcfr:
> What is the reason this has only came up for VTK9?

The CMake infrastructure to collect hierarchy files has been revamped in VTK9 and the module we maintained in `vtkAddon` does not handle the case of `vtkLibrary1` -> `qSlicerLibrary1` -> `vtkLibrary2`. 

See https://discourse.vtk.org/t/issue-with-using-vtkpythonutil-getobjectfrompointer-for-custom-vtk-classes/4100/ and https://github.com/Slicer/vtkAddon/commit/5df87dee75c049320943f445200b4cbe7309f811

lassoan:
Thanks for the explanation and congratulations for being able to track down this issue, so deeply in the Python wrapping infrastructure. Hopefully such issues will not come up too often, because it seems quite difficult to identify the problem.

jcfr:
Considering I had spent a fair amount of time when working on the initial transition, it didn't take too long to address this particular one. 

That said, it would have taken a significant amount time without that initial investment.

> Hopefully such issues will not come up too often

This is a very special case and I don't foresee this happening in extensions.

PR #5467, opened 2021-02-17
Improve dicom export robustness

lassoan:
Small fixes for errors found while investigating failing DICOM export into folders that have special characters in their names.


pieper (approved):
👍

Issue #5061, opened 2020-07-25
Create markups ROI


lassoan:
Annotation ROI nodes have many limitations. We need complete rework, based on markups infrastructure.

## Describe the solution you'd like

Have a new vtkMRMLMarkupsROINode, similar to vtkMRMLAnnotationROINode but with these additional features:
- [x] can be rotated
- [x] can be transformed (#1636)
- [x] has label
- [x] its color can be set
- [x] has inside-out mode (https://github.com/Slicer/Slicer/issues/1596)

Additional specific requests:

High priority:
- [x] Storage: use a single "orientation" matrix for storing orientation (instead of separate xAxis, yAxis, zAxis"), as it is more consistent with how we usually manage orientation
- [x] Storage: "sideLenghts" is very clear definition for a box, but not so good if we generalize the ROI (e.g., to ellipsoid); "diameter" or "size" could be better, as they are probably still clear enough and applicable to more shapes than just a box
- [x] Storage: "origin" is consistent with how we represent a coordinate system, but in case of a ROI, it may be more informative to have the "center" displayed in the file (if users create a file manually or from an external software, it could be more intuitive if they specify the center); "center" is also a bit more clear than "origin" (it may not be clear for everyone that origin is in a corner)
- [x] Storage: it would be better to display ROI properties (origin, size, etc.) near the beginning (before control points, measurments, display sections), because these files will be directly viewed and edited by users, so we should always start with the most important information
- [x] Make filling 20% opacity by default (and set it to 0% when used as volume rendering clipping ROI)
- [x] Translation/rotation handles are too small in 3D (about half the size compared to slice views)
- [x] When a ROI is already the active widget and user clicks place mode again then a new widget should be created (now it resets the current widget). At least until we improve the widget toolbar (we should have separate options for editing selected segmentation or create a new one).
- [x] When the ROI is in bounding box mode and has many control points and then you switch to box mode then the ROI seems to remain in bounding box mode.
- [x] We need outline (wireframe) display in 3D view (for example, for volume rendering clipping that would be a good representation)
- [x] After placing a ROI (at least when it is done, interactively, initiated via the GUI) then make the interaction handles displayed by default, because now after you place the ROI it is very hard to adjust it.
- [x] Make right-click menu work on all interaction handle points
- [x] Make right-click work on the outline (but maybe on the filled area as well, if it is >0%) so that interaction handles can be activated/deactivated in a viewer (it can be hard to find a ROI node in the Data module)
[...]

PR #5468, opened 2021-02-18
ENH: Improve markup ROI interaction

Sunderlandkyl:
List of improvements:
- When an ROI is already active and the user clicks place mode, create a new ROI rather than modify the old one.
- Correctly switch between box and bounding box ROI types.
- Enable hover and interaction (right-click) on ROI outline.
- Change box ROI type to place a single fiducial at the center of the ROI.


Sunderlandkyl:
@lassoan this should address the high-priority items in #5061.

lassoan (approved):
Thank you very much, it looks good!

PR #5471, opened 2021-02-19
BUG: Fix MainWindow.statusBar() calls in util.py

Sunderlandkyl:
In several places in util.py, statusBar was called as a property, rather than as a method.
This caused an error when setStatusBarVisible was called.

Fixed by changing mw.statusBar -> mw.statusBar()


lassoan (approved):
Good catch, thank you!

PR #5472, opened 2021-02-19
BUG: Fix scene not updating after toggling subject hierarchy item visibility

lassoan:
When hitting a key on a subject hierarchy item that did not have an owner, the scene stopped updating.

It was due to starting but not ending a batch processing event when no owner plugin was found for the item.


cpinter:
Makes sense, thank you!

PR #5475, opened 2021-02-19
BUG: Fix table view losing table node selection after batch processing

lassoan:
When table view node was not set in a qMRMLTableView but the table node was set correctly, a scene batch update resulted in resetting the table node.

See user report here: https://discourse.slicer.org/t/cant-modify-a-table-in-tables-module-after-using-landmark-registration-module/16069

qMRMLTableView is allowed to be used without a view node, so fixed the problem by ignoring scene batch processing update if view node is not set.


pieper (approved):
👍

PR #5477, opened 2021-02-21
ENH: Improve default axis-aligned display of single-slice volume

lassoan:
When "Reset view orientation on show" was enabled then single-slice volumes fully appeared in all view (view normal aligned with slice normal), with different rotations.
This resulted in redundant and slightly confusing views.

Fixed the problem by forcing slice view normal to be aligned with the volume slice normal only if a volume is requested to be shown in a single view.
If a volume is shown in multiple views then each slice view is aligned with the nearest volume axis.

Before the fix:

image

After the fix:

image

### DIFF
--- a/Libs/MRML/Logic/vtkMRMLSliceLogic.h
+++ b/Libs/MRML/Logic/vtkMRMLSliceLogic.h
@@ -239,7 +239,10 @@ public:
   void GetBackgroundSliceBounds(double sliceBounds[6]);
 
   /// Rotate slice view to match axes of the lowest volume layer (background, foreground, label).
-  void RotateSliceToLowestVolumeAxes();
+  /// \param forceSlicePlaneToSingleSlice If the volume is single-slice and forceSlicePlaneToSingleSlice
+  /// is enabled then slice view will be aligned with the volume's slice plane. If the flag is disabled
+  /// of the volume has more than one slice then the slice view will be rotated to the closest orthogonal axis.
+  void RotateSliceToLowestVolumeAxes(bool forceSlicePlaneToSingleSlice = true);
 
   ///
   /// adjust the node's field of view to match the extent of current background volume
--- a/Modules/Loadable/Volumes/SubjectHierarchyPlugins/qSlicerSubjectHierarchyVolumesPlugin.cxx
+++ b/Modules/Loadable/Volumes/SubjectHierarchyPlugins/qSlicerSubjectHierarchyVolumesPlugin.cxx
@@ -28,7 +28,7 @@
 // Slicer includes
 #include "qMRMLSliceWidget.h"
 #include "qSlicerApplication.h"
-#include "qSlicerLayoutManager.h" 
+#include "qSlicerLayoutManager.h"
 #include "vtkSlicerApplicationLogic.h"
 
 // MRML includes
@@ -460,7 +460,10 @@ void qSlicerSubjectHierarchyVolumesPlugin::showVolumeInAllViews(
           // Set to default orientation before rotation so that the view is snapped
           // closest to the default orientation of this slice view.
           sliceNode->SetOrientationToDefault();
-          sliceWidget->sliceLogic()->RotateSliceToLowestVolumeAxes();
+          // If the volume is shown in only one view and the volume is a single slice then
+          // make sure the view is aligned with that, otherwise just snap to closest view.
+          bool forceSlicePlaneToSingleSlice = (numberOfCompositeNodes == 1);
+          sliceWidget->sliceLogic()->RotateSliceToLowestVolumeAxes(forceSlicePlaneToSingleSlice);
           }
         if (resetFov)
           {
--- a/Libs/MRML/Core/vtkMRMLSliceNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLSliceNode.cxx
@@ -2114,7 +2114,7 @@ void vtkMRMLSliceNode::SetSliceOffset(double offset)
 }
 
 //----------------------------------------------------------------------------
-void vtkMRMLSliceNode::RotateToVolumePlane(vtkMRMLVolumeNode *volumeNode)
+void vtkMRMLSliceNode::RotateToVolumePlane(vtkMRMLVolumeNode *volumeNode, bool forceSlicePlaneToSingleSlice/*=true*/)
 {
   if (volumeNode == nullptr)
     {
@@ -2141,7 +2141,7 @@ void vtkMRMLSliceNode::RotateToVolumePlane(vtkMRMLVolumeNode *volumeNode)
     }
 
   int volumeAxisIndexForSliceZ = -1;
-  if (volumeNode->GetImageData() != nullptr)
+  if (forceSlicePlaneToSingleSlice && volumeNode->GetImageData() != nullptr)
     {
     int dims[3] = { 0, 0, 0 };
     volumeNode->GetImageData()->GetDimensions(dims);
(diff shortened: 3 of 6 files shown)


Libs/MRML/Logic/vtkMRMLSliceLogic.h
```diff
@@ -239,7 +239,10 @@ class VTK_MRML_LOGIC_EXPORT vtkMRMLSliceLogic : public vtkMRMLAbstractLogic
   void GetBackgroundSliceBounds(double sliceBounds[6]);
 
   /// Rotate slice view to match axes of the lowest volume layer (background, foreground, label).
-  void RotateSliceToLowestVolumeAxes();
+  /// \param forceSlicePlaneToSingleSlice If the volume is single-slice and forceSlicePlaneToSingleSlice
+  /// is enabled then slice view will be aligned with the volume's slice plane. If the flag is disabled
+  /// of the volume has more than one slice then the slice view will be rotated to the closest orthogonal axis.
```
cpinter, line 244: of => or?

cpinter (dismissed):

Libs/MRML/Logic/vtkMRMLSliceLogic.h
lassoan: good catch!

lassoan:
@cpinter thanks for the review!

Modules/Loadable/Volumes/SubjectHierarchyPlugins/qSlicerSubjectHierarchyVolumesPlugin.cxx
```diff
@@ -460,7 +460,10 @@ void qSlicerSubjectHierarchyVolumesPlugin::showVolumeInAllViews(
           // Set to default orientation before rotation so that the view is snapped
           // closest to the default orientation of this slice view.
           sliceNode->SetOrientationToDefault();
-          sliceWidget->sliceLogic()->RotateSliceToLowestVolumeAxes();
+          // If the volume is shown in only one view and the volume is a single slice then
+          // make sure the view is aligned with that, otherwise just snap to closest view.
+          bool forceSlicePlaneToSingleSlice = (numberOfCompositeNodes == 1);
```
jcfr, line 465: Is the `numberOfCompositeNodes` related to the number of layer and not the number of slice got a given viewer ?

cpinter: It is related to the number of views. A composite node can control the layers as it has members `SetBackground/Foreground/LabelVolumeID` which we use to set what to show in a 2D view

lassoan: Yes, we make the forcing strategy dependent on number of views, because if the user asks Slicer to show a single-slice volume in a specific view (e.g., drag-and-drop in a single view) then we want to show the complete slice in that view; but if the user wants to show that single-slice volume in 3 orthogonal views then we don't need to force all three of them to show the complete slice (only _one_ of the 3 orthogonal views will snapped to the image axes so that it will show the full slice, but not all 3). This essentially restores the previous snapping behavior for multiple views, while keeping the new behavior for single view (important for drag-and-drop).

Another option would be to dynamically change the layout depending on what the user requests to display. For example, if display of a regular 3D volume is requested then add 3 slice views to the layout; if display of a single-slice volume is displayed then add one slice view to the layout. But this would be a drastic change, we would need to think through the use cases more.

jamesobutler: > if display of a regular 3D volume is requested then add 3 slice views to the layout; if display of a single-slice volume is displayed then add one slice view to the layout. 

My group automatically changes the layout if we are looking at a regular 3D volume versus a 2D volume. We automatically choose layout based on the dimensions of the volume.  The only additional thing you have to consider is when a transform is applied to a 2D volume how to display it such as whether it is automatically in a reformat view.

lassoan: > My group automatically changes the layout if we are looking at a regular 3D volume versus a 2D volume.

This would be a quite big change in how Slicer works, but could make things convenient, especially for new users. How do you deal with displaying multiple volumes? How the user can indicate that both volumes should be displayed in the same views (blended) or you need to open a new set of views? How the user indicates that he would like to see a 3D view or not (especially when showing two volumes in two view groups, because then you could have a 3-over-3 or 4-over-4 layout)?

jamesobutler: A lot of the customizations we have made reduce all the different things you can do in Slicer, but it is simpler for new users and helps automate certain small tasks to get to a view that is likely where they want to start. I would also be willing to meet and demo some customizations to give the other Slicer developers some ideas. I know it can be difficult to fully understand just with a text description.

First main change as it relates to view stuff is that we don't allow users to see the full expanded Slice controller options. In our custom module panel area we have the following area that is persistent across all types of analysis work.
image
It is like a simplified version of the slice controller area. Primary volume defines the Foreground volume layer and the Secondary volume defines the Background/reference volume layer. There is the overlay vertical slider. Over on the right side there is a toggle button for Overlay or Side-By-Side. The icon changes based on what is current. Below that the button with the arrows is a "Swap" button to quickly swap volumes between primary/secondary.

>  How do you deal with displaying multiple volumes?

We have for a while only supported 2 volumes displayed in the layout manager at once. You're either in "Overlay" mode in the same slice group or you are in "Side-By-Side" where we have 2 slice groups with 1 volume per group. The primary volume is shown in slice viewers like R/Y/G/1 and secondary volumes while in Side-By-Side are indicated by R+/Y+/G+/1+

> How the user can indicate that both volumes should be displayed in the same views (blended) or you need to open a new set of views?

This is handled by the logic tied with the Overlay/Side-By-Side button. Our default selection is for overlay. So if your primary volume is 3D (x/y/z dims >1) our decided best layout for the volume is the "Four-up" layout. If you select a secondary volume that is also 3D and in overlay mode we would propagate that volume as the background volume in the Four-up slice viewers. We also detect if the volume bounds overlap, so in the case that they don't overlap, we automatically change to Side-By-Side mode as otherwise you couldn't see both and lock them out of overlay mode.

> How the user indicates that he would like to see a 3D view or not (especially when showing two volumes in two view groups, because then you could have a 3-over-3 or 4-over-4 layout)?

Our preferred layout for 2 3D Volumes in side-by-side mode is a Three-over-three view. We don't normally find the 3D view all that helpful outside of Volume Rendering. We do most of our work in the 2D slice viewers. That is the reasoning for us defining the preferred layout to be Three-over-three. Once a user changes a selection in the primary volume or secondary volume combobox it always goes to our preferred layout for that combination of volumes. If the user wants to go to a different layout such as 4-over-4 we allow them to change the layout to whatever. We prioritize the majority workflow case to provide convenience and let the minority cases deal with the preferred layout stuff that we enforce.

lassoan: These are excellent points (I especially like the check for overlap between volumes to decide to show it overlaid or in separate views). It would be nice to see what customizations you implemented and what could be utilized in a general-purpose Slicer application. Could you join one of the Slicer weekly meetings?

PR #5479, opened 2021-02-22
BUG: Fix default storage node creation in LoadMarkupsFromJson

Sunderlandkyl:
The GetJsonStorageNodeClassNameForMarkupType was supposed to return the default Json storage node "vtkMRMLMarkupsJsonStorageNode" if a specific storage node wasn't registered for the markup type, but it instead returned an empty string.
Fixed by changing default return to the correct value.


lassoan (commented):
Looks good to me. Just a nitpick: use "markups" everywhere (not "markup"). By using "markups" as a fixed name prefix we never need to decide if "markup" or "markups" is more appropriate.

Sunderlandkyl:
Thanks! Changed Markup -> Markups.

lassoan (approved):
Thank you, looks good.

PR #5482, opened 2021-02-23
ENH: Add Shift shortcut for changing ROI size without moving origin

Sunderlandkyl:
Pressing the shift button while scaling the ROI will adjust the size of the ROI symmetrically, keeping the origin in the same location.
Re #5061

### DIFF
--- a/Modules/Loadable/Markups/VTKWidgets/vtkSlicerROIWidget.cxx
+++ b/Modules/Loadable/Markups/VTKWidgets/vtkSlicerROIWidget.cxx
@@ -38,7 +38,11 @@
 vtkStandardNewMacro(vtkSlicerROIWidget);
 
 //----------------------------------------------------------------------
-vtkSlicerROIWidget::vtkSlicerROIWidget() = default;
+vtkSlicerROIWidget::vtkSlicerROIWidget()
+{
+  this->SetEventTranslationClickAndDrag(WidgetStateOnScaleHandle, vtkCommand::LeftButtonPressEvent, vtkEvent::AltModifier,
+    WidgetStateSymmetricScale, WidgetEventSymmetricScaleStart, WidgetEventSymmetricScaleEnd);
+}
 
 //----------------------------------------------------------------------
 vtkSlicerROIWidget::~vtkSlicerROIWidget() = default;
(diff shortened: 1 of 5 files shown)


Modules/Loadable/Markups/VTKWidgets/vtkSlicerROIWidget.cxx
```diff
@@ -63,8 +67,52 @@ void vtkSlicerROIWidget::CreateDefaultRepresentation(
...
+  // of the widget.
+  double eventPos[2]
+    {
+    static_cast<double>(eventData->GetDisplayPosition()[0]),
+    static_cast<double>(eventData->GetDisplayPosition()[1]),
+    };
+
+  if (state == WidgetStateScale)
+    {
+    bool symmetricScale = eventData->GetModifiers() & vtkEvent::ShiftModifier;
```
lassoan, line 95: At this point, we should not check the modifier anymore, because it would prevent dynamic remapping of GUI->widget events. Instead, Shift+MouseMove should be mapped to a new widget event trio (something like WidgetStateSymmetricScale, WidgetStateSymmetricScaleStart, WidgetStateSymmetricScaleEnd).

lassoan (commented):
This will be good, I would just suggest one (hopefully small) design change - see inline

Sunderlandkyl:
Ok. I've added the separate state SymmetricScale states and events.
I just realized that holding shift and scaling still moves the slice intersection. Not sure if this is desirable behavior or not.

lassoan (commented):
Thank you, the implementation looks perfect now.

>I just realized that holding shift and scaling still moves the slice intersection. Not sure if this is desirable behavior or not.

I've checked this and moving of the crosshair is not good (it messes up the slice views, so it is hard to set volume clipping region). But I've tried changing it to AltModifier and then it works perfectly.

One more thing: could you add description of this shortcut to the [documentation](https://slicer.readthedocs.io/en/latest/user_guide/modules/markups.html)? The structure of the page is not great, maybe you could just add a section about ROI editing to [How to](https://slicer.readthedocs.io/en/latest/user_guide/modules/markups.html#how-to) section.

Sunderlandkyl:
Changed to Atl and added a section about modifying ROI to docs.

lassoan (approved):
This is perfect, thank you!

PR #5485, opened 2021-02-25
ENH: Add separate node attributes for interaction handle visibility

Sunderlandkyl:
Markup display nodes now have the option to show/hide interaction handle types separately (Translation, Rotation, Scale).
All interaction handle types are shown by default if interaction visibility is enabled. For ROI display nodes, translation and rotation handles are hidden by default.

Visibility of handle types can be controlled from the markup display node widget, or through the subject hierarchy context menu.

This commit also reduces the thickness of the translation arrows by 25%.

Re #5061

image

### DIFF
--- a/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsDisplayNode.cxx
+++ b/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsDisplayNode.cxx
@@ -125,6 +125,9 @@ vtkMRMLMarkupsDisplayNode::vtkMRMLMarkupsDisplayNode()
   this->ActiveColor[2] = 0.0;
 
   this->HandlesInteractive = false;
+  this->TranslationHandleVisibility = true;
+  this->RotationHandleVisibility = true;
+  this->ScaleHandleVisibility = true;
 
   // Line color node
   vtkNew<vtkIntArray> events;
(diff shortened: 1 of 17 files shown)


lassoan (dismissed):
Looks good to me, thank you!
Have you tried to make the 3 checkboxes children of the ToggleHandleInteractive action?

Sunderlandkyl:
Yes, unfortunately that prevents the ToggleHandleInteractive action from being clickable.

lassoan:
> Yes, unfortunately that prevents the ToggleHandleInteractive action from being clickable.

OK, no problem.

However, I've noticed that we are getting quite inconsistent with naming. We should standardize across all widgets how we name the enabling/disabling interaction and setting modes. It is just for the GUI, no need to change variable names in code.

Currently:

image

image

image

Enable/disable:
- Interaction in 3D view (for transforms, in SH menu)
- Interaction / Visible in 3D view (for transforms, in Transforms module)
- Interaction handles visible (for markups, in SH menu)

Modes:
- Interaction / Enable translation, Enable rotation, Enable scaling (for transforms, in Transforms module)
- Interaction handles / Translate, Rotate, Scale (for markups, in SH menu)

Proposed:

Something like this may be OK (but maybe you can come up with something better):

- [ ] Interaction
- Interaction options
  - [ ] Translate
  - [ ] Rotate
  - [ ] Scale

For transforms, for now, we would still add "in 3D view" because there is no widget in 2D.
I would call "Interaction options" instead of something like "Interaction modes" because in the feature we may add local/global options (translate/rotate along rotated or world coordinate system axes).

Maybe Enable translation, Enable rotation, Enable scaling would be better (as by clicking we don't actually translate/rotate/scale, only allow the user to perform these)?

lassoan:
By the way, ability to enable each mode is separately was a great idea, it significantly improves usability of the widgets!

Sunderlandkyl:
Makes sense, I've changed the menu:
image

lassoan (approved):
Thank you, it looks good!

PR #5487, opened 2021-02-26
ENH: Ensure that remote cache folder is writable

lassoan:
macOS started to manage application temporary and cache folders more actively, i.e., remove them after certain time period (or at system restart or application exit) to improve security.

Cache folder by default was initialized to be a subfolder of temporary folder. However, when the system removed the temporary folder, the cache folder has become invalid.

To fix this, implemented an automatic mechanism to check the cache folder at startup and reset it to a system-recommended location if the folder is not writable.

See related user error report: https://discourse.slicer.org/t/slicer-4-11-on-m1-not-loading-mrb-file/16228/7

The pull request also contains an unrelated commit with a trivial fix of missing imports in slicer.util.

### DIFF
--- a/Base/Python/slicer/util.py
+++ b/Base/Python/slicer/util.py
@@ -1063,6 +1063,7 @@ def getModuleWidget(module):
   :return: module widget object
   :raises RuntimeError: if the module does not have widget.
   """
+  import slicer
   if isinstance(module, str):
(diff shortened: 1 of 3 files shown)


pieper (approved):
Thanks for fixing this!

LGTM

PR #5488, opened 2021-02-26
ENH: Update slicerMacroBuildApplication adding SPLASHSCREEN_DISABLED option

jcfr:
This commit allows to configure custom application without application launcher
splashscreen


CMake/SlicerMacroBuildApplication.cmake
```diff
@@ -285,6 +286,8 @@ macro(slicerMacroBuildApplication)
     LICENSE_FILE
 
     TARGET_NAME_VAR
+
+    APPLICATION_DEFAULT_ARGUMENTS # space separated list
```
lassoan, line 290: Is this change intentional?

CMake/SlicerMacroBuildApplication.cmake
```diff
@@ -307,11 +310,15 @@ macro(slicerMacroBuildApplication)
 
   set(expected_defined_vars
     NAME
-    LAUNCHER_SPLASHSCREEN_FILE
     APPLE_ICON_FILE
     WIN_ICON_FILE
     LICENSE_FILE
     )
+  if(NOT SLICERAPP_SPLASHSCREEN_DISABLED)
```
lassoan, line 317: I much prefer using positive terms for variable names (SLICERAPP_SPLASHCREEN_ENABLED instead of SLICERAPP_SPLASHSCREEN_DISABLED) to avoid dispersing error-prone and hard-to-read double negatives (NOT ..._DISABLED).

Would it be possible to use SLICERAPP_SPLASHCREEN_ENABLED variable name and initialize it to TRUE by default?

lassoan (commented):
Thank you, it looks good overall, just added two minor comments.

CMake/SlicerMacroBuildApplication.cmake
jcfr: Yes, this provide  way to specify additional default arguments. It is the variable `SLICERAPP_APPLICATION_DEFAULT_ARGUMENTS` concatenated to `_launcher_application_default_arguments`

It is space separated because is is handled as such in the launcher:
https://github.com/commontk/AppLauncher/blob/1367de4c6efde0c11e87835fb7245ea2b05074aa/Base/ctkAppLauncher.cpp#L272-L281

CMake/SlicerMacroBuildApplication.cmake
jcfr: Since the function `cmake_parse_arguments` accept option, and specifying option means the variable will be set to True, following this approach was sensible and consistent with how the launcher itself is configured.

https://github.com/commontk/AppLauncher/blob/1367de4c6efde0c11e87835fb7245ea2b05074aa/CMake/ctkAppLauncher.cmake#L436

jcfr: In  the Slicer macro, I could do something like this:

```
set(_spashscreen_enabled 1)
if(SLICERAPP_SPLASHSCREEN_DISABLED)
  set(_spashscreen_enabled 0)
endif()
```

and then test the variable `_spashscreen_enabled`. That seems to be a reasonable tradeoff ?

lassoan: The proper solution would be to be able to specify default value for undefined variables. I thought that this is already possible or easy to add, but if it is not then it is fine to keep the SLICERAPP_SPLASHSCREEN_DISABLED variable name as is (no need for introducing the local variable).

jcfr: I could indeed switch from having `SPLASHSCREEN_DISABLED` specified as an option (meaning if specified it is `ON` and `OFF` otherwise) to having a single arg parameter.

After switching to a single arg parameter, it will indeed be possible to have a default value. I will do that.

jcfr:
@lassoan I revisited the approach to account for your comment, let me know if you have other suggestions.

lassoan (approved):
This looks perfect, thank you.

Issue #5114, opened 2020-08-19
DICOM Plugins icon


spujol:
The icon (>>) to display the list of DICOM plugins in the DICOM module is black and there is little contrast with the background color in Dark mode. 
<img width="1680" alt="Screen Shot 2020-08-19 at 1 00 02 PM" src="https://user-images.githubusercontent.com/1847492/90649163-62988580-e208-11ea-8c3a-8679b8de2937.png">

Would it be possible to change the size and color of the icon to make it more visible to the end-user?

pieper:
If anyone is digging around in that code, it would be great the plugin list had a scrollbar.  We have enough plugins now that they get squished if there are many installed.  Also. splitter between the loadable list and the patient/study/series panes would be nice.

dsomani125:
My team wants to work on this issue. @ISD14-LNMIIT

jcfr:
Thanks @dsomani125 :pray:

pieper:
> My team wants to work on this issue. @ISD14-LNMIIT

Great!

Some of us discussed this on the last developer hangout and thought a good solution would be to most the checklist of plugins into a collapsable frame in the module panel (maybe below the `DICOM database settings`).  We thought this would simplify the layout.

lassoan:
> My team wants to work on this issue

@dsomani125 We haven't heard from you since you commented on this. Do you plan to work on this? Do you need help to get started?

dsomani125:
> > My team wants to work on this issue
> 
> @dsomani125 We haven't heard from you since you commented on this. Do you plan to work on this? Do you need help to get started?

Actually I will not be able to work on this issue for now. 
Will surely work with you in future. 
Sorry for the inconvenience.

lassoan:
OK, no problem. I'll unassign the issue then.

MichaelColonel:
> If anyone is digging around in that code, it would be great the plugin list had a scrollbar. We have enough plugins now that they get squished if there are many installed.

As a temporal solution of that issue.
![image](https://user-images.githubusercontent.com/3785912/109125014-a0004180-775c-11eb-8553-39e83a5dfe7e.png)

 I can try to move that element into a collapsible frame in a module panel later, when i used to python scripting.

lassoan:
@MichaelColonel this looks good. Could you send a pull request with the change?

MichaelColonel:
[...]

PR #5491, opened 2021-02-27
ENH: DICOM plugins layout is moved from browser to module panel

MichaelColonel:
Another solution for an issue #5114

### DIFF
--- a/Docs/user_guide/modules/dicom.md
+++ b/Docs/user_guide/modules/dicom.md
@@ -137,7 +137,7 @@ Slicer supports sending of DICOM items to a remote server using DICOMweb protoco
 - **Load**: click this button to load currently selected loadables into slicer.
 - **Loaded data**: shows all content currently loaded into the scene, which can be displayed in viewers by clicking the eye icon
 
-![](https://github.com/Slicer/Slicer/releases/download/docs-resources/module_dicom_basic.png)
+![](https://github.com/Slicer/Slicer/releases/download/docs-resources/module_dicom_basic_rev02.png)
 
 Additional options:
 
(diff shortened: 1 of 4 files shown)


lassoan:
Thanks a lot, this will be great! I'll review and test this tomorrow.

Modules/Scripted/DICOM/Resources/UI/DICOM.ui
```diff
@@ -6,7 +6,7 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>372</width>
+    <width>615</width>
```
lassoan, line 9: Keep this value the same. Larger value may increase minimum width of the DICOM module.
```suggestion
    <width>372</width>
```

lassoan (commented):
This looks great and works well, thanks a lot.

I've just added a trivial comment inline and the DICOM module documentation (https://github.com/Slicer/Slicer/blob/master/Docs/user_guide/modules/dicom.md) needs a small update, too: 
- Update this text: `Plugin selector section: you can choose which plugins will be allowed to examine the selected series for loading. This section is displayed if the double-arrow button on the left side of loadable items list is pushed.`
- Update screenshots with these new ones that I've uploaded: https://github.com/Slicer/Slicer/releases/download/docs-resources/module_dicom_basic_rev02.png,  https://github.com/Slicer/Slicer/releases/download/docs-resources/module_dicom_advanced_rev02.png

MichaelColonel:
I've updated documentation and UI file.

Docs/user_guide/modules/dicom.md
```diff
@@ -155,7 +155,7 @@ Advanced loading (allows loading DICOM data sets using non-default options):
 - **Examine button:** Runs each of the DICOM Plugins on the currently selected series and offers the result in the Loadable items list table.
```
lassoan, line 155: Please update the "Plugin selector section" above (no double-arrow, etc.)

lassoan (commented):
Thank you, it is almost perfect, just need to update the "Plugin selector section" description.

Docs/user_guide/modules/dicom.md
MichaelColonel: Could you check, that description is OK?

lassoan: Yes, it is perfect, thank you.

lassoan (dismissed):
This all looks good now, thank you!
I'll merge it now.

lassoan:
Squashed and merged your changes. Thank you for your contribution!

PR #5495, opened 2021-03-01
ENH: Hide markup line in 3D view if the number of control points != 2

Sunderlandkyl:
Previously, creating a vtkMRMLMarkupsLineNode with 2 control points and then removing one or both would leave a line in the 3D view.
This commit hides the line actor if there aren't 2 defined control points.


lassoan (commented):
Thank you, indeed the lines were left there. However, after this change, do we still see the line during preview?

The method's signature is `int GetNumberOfDefinedControlPoints(bool includePreview=false)`, which suggests that when there is one defined and one preview point then the line is not shown.

Sunderlandkyl:
Yes, because we call GetNumberOfDefinedControlPoints(true).

lassoan (approved):
I don't know how I missed that! It is good then, sorry for the noise.

PR #5496, opened 2021-03-01
BUG: Fix paint effect drawing extra strokes after interrupted

lassoan:
If Segment Editor Paint effect was interrupted by hitting Escape then next time it was used again, a stray stroke was added because IsPainting state variable was not reset.

See user error report at https://discourse.slicer.org/t/issue-with-brush-or-eraser/16270


Sunderlandkyl (approved):
LGTM!

PR #5499, opened 2021-03-02
BUG: Fix mouse flicker when hovering over interaction handles

Sunderlandkyl:
While hovering over interaction handles, the mouse was constantly flickering between interactive and non-interactive modes due to missing MouseMoveEvent/Move3DEvent event translations.


lassoan (approved):
Perfect, thank you!

PR #5501, opened 2021-03-03
BUG: Fix linux package ensuring OpenGL_GL_PREFERENCE is considered with VTK9

### LINKED ISSUE #5228: VTK9: Linux package requires  libGLX.so
Linux package built with VTK9 requires `libGLX.so.0`

To ensure the Slicer can run on system where only libGL is available, we will have to find out why the setting of `OpenGL_GL_PREFERENCE` to `LEGACY` originally introduced in 23d8144 does not have any effect on the VTK9 build.

I suspect it may be related to [CMP0072](https://cmake.org/cmake/help/latest/policy/CMP0072.html)

## Steps to reproduce

Build Slicer using the latest [SlicerBuildEnvironment](https://github.com/Slicer/SlicerBuildEnvironment/) docker image and set option `Slicer_VTK_VERSION_MAJOR` to `9`

## Expected behavior

On older system that do not provide the GL Vendor Neutral Dispatch libraries (`libOpenGL.so` and `libGLX.so`), Slicer should starts without reporting message like this one:

```
./Slicer 
/tmp/Slicer-4.13.0-2020-10-01-linux-amd64/bin/SlicerApp-real: error while loading shared libraries: libGLX.so.0: cannot open shared object file: No such file or directory

```

## Environment
- Slicer version: Slicer-4.??.?-YYYY-MM-DD
- Operating system: Windows / Linux / Mac + which version

### LINKED ISSUE #5286: missing libOpenGL.so on ubuntu 20.10
The download gives this error:
```
$ ~/Downloads/Slicer-4.13.0-2020-10-31-linux-amd64/Slicer 
/home/pieper/Downloads/Slicer-4.13.0-2020-10-31-linux-amd64/bin/SlicerApp-real: error while loading shared libraries: libOpenGL.so.0: cannot open shared object file: No such file or directory
pieper@sdp-november:~$ cat /etc/issue
Ubuntu 20.10 \n \l
```

4.11-20200930 works fine on the same machine.

jcfr:
To ensure the Slicer can run on system where only libGL is available,
this commit updates VTK to make sure setting OpenGL_GL_PREFERENCE
to LEGACY works as expected. For reference, setting to LEGACY value
was originally introduced in Slicer@23d8144 (BUG: Fix linux package
removing dependency to GLVND libraries).

List of changes:

```
$ git shortlog d76bf153dd..60b13991d3
      COMP: Remove custom FindOpenGL to ensure OpenGL_GL_PREFERENCE is considered
```

Fixes #5228
Fixes #5286

### DIFF
--- a/SuperBuild/External_VTK.cmake
+++ b/SuperBuild/External_VTK.cmake
@@ -177,7 +177,7 @@ if((NOT DEFINED VTK_DIR OR NOT DEFINED VTK_SOURCE_DIR) AND NOT Slicer_USE_SYSTEM
     set(_git_tag "97904fdcc7e73446b3131f32eac9fc9781b23c2d") # slicer-v8.2.0-2018-10-02-74d9488523
     set(vtk_egg_info_version "8.2.0")
   elseif("${Slicer_VTK_VERSION_MAJOR}" STREQUAL "9")
-    set(_git_tag "d76bf153dd694d9fc5d84fcee10702091f60ace0") # slicer-v9.0.20201111-733234c785
+    set(_git_tag "60b13991d37445f70a8e24604a45047022e6cb18") # slicer-v9.0.20201111-733234c785
     set(vtk_egg_info_version "9.0.20201111")
   else()
     message(FATAL_ERROR "error: Unsupported Slicer_VTK_VERSION_MAJOR: ${Slicer_VTK_VERSION_MAJOR}")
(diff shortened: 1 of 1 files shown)


jcfr:
@RafaelPalomar @pieper When you have a chance, would appreciate if you could test.

pieper (dismissed):
@jcfr thanks for the fix 👍 

Based on the description I'd suggest just merging this and testing the factory builds.  Or if you think there's a serious risk of regressions I can package up a release build for testiong.

jcfr:
Thanks for the review. 

I just did additional testing, and I confirm VTK itself is built against the LEGACY OpenGL libraries but project depending on VTK are not, I am working on addressing this.

PR #5508, opened 2021-03-05
COMP: Undefined class "QTemporaryFile" when building using SlicerCAT

kerim371:
Without this `#include <QTemporaryFile>` I got compilation error telling me that `QTemporaryFile` is undefined. With this little change I could compile on Windows 10 MSVC2019, Qt 5.15.2


lassoan (approved):
The change makes sense. Thank you for your contribution.

PR #5509, opened 2021-03-06
ENH: Add custom volume rendering presets to the top of the preset list by default

lassoan:
See user request:
https://discourse.slicer.org/t/i-want-to-boot-up-with-my-very-own-color-map-and-a-black-background/16383/10?u=lassoan


pieper (approved):
Thanks for adding this! 👍

PR #5512, opened 2021-03-09
COMP: Fix qMRMLMarkupsROIWidget build error against Qt 5.10

jcfr:
This commit removes uses of "PlaceholderText" in UI files that was
originally introduced in 0ba962c12 (ENH: Add Markups ROI node)

It fixes the following error:

  error: ‘PlaceholderText’ is not a member of ‘QPalette’

QPalette::PlaceholderText was introduced in Qt 5.12
See https://github.com/qt/qtbase/commit/ebd3a13b807c6af2684b42d3912549caf7ef82aa


Sunderlandkyl (approved):
Thanks, LGTM!

lassoan:
The change looks OK and does not make any harm, I'm just wondering if what Qt version do we still need to support? Current build instructions state that Qt-5.15 is required to build Slicer.

jcfr:
> I'm just wondering if what Qt version do we still need to support?

This is mainly motivated because I am still using Qt 5.10 ... :facepalm: 

We will also need to update this at some point:

https://github.com/Slicer/Slicer/blob/ee2c0a5c043abd60edc2f2c1bfd1c88a2afcb44e/CMakeLists.txt#L161-L165

jamesobutler:
I guess how it read it was that Qt5 was the perquisite and not a specific version.

lassoan:
I see, OK, it can be interpreted like this, too.

Should we update the text from `Qt5: Download...` to `Qt 5.15: Download...`?
We could also write a general comment to prerequisites to clarify that these are not recommendations but requirements: `Using any other versions of the listed tools or libraries are not tested and may lead to build or runtime errors`.

Issue #5480, opened 2021-02-22
Hide markups in slice views when their region does not intersect the slice


lassoan:
## Is your feature request related to a problem? Please describe.

It is confusing for users that lines are visible on neighbor slices. Similarly, it can be confusing that curves and angles fade out gradually over a few slices. Plane and ROI widgets work well (they already hidden when they do not intersect the slice).

## Describe the solution you'd like

Hide lines, curves, and angles markups in the slice view if they don't intersect the slice view (with some tolerance, probably half slice spacing tolerance would work).

## Describe alternatives you've considered

Reducing maximum fadeout distance was also considered but it would not work very well for curves.

## Additional context

See this discussion: https://discourse.slicer.org/t/remove-markup-lines-on-adjacent-slices/16136/4

aldenshoup:
Hello, this fix seems to not be present in the latest version

lassoan:
Please provide more information, such as description of steps to reproduce and a few screenshots.

aldenshoup:
Hello,

My understanding is that a part of the line should only appear where it crosses the slice in the 2D slice. This line crosses multiple slices but the whole line is visible in all of the slices it crosses. I have toggled all of the options under "Display".
<img width="929" height="1234" alt="Image" src="https://github.com/user-attachments/assets/eabf8d92-b398-407f-aa88-823ab709eacb" />

lassoan:
The slice is not modeled as an infinitely thin plane, but there is a thickness, specified by `ColorFadingStart` and `ColorFadingEnd` parameters of the line markup. You can reduce that to avoid showing a long line segment:

```
line=getNode('L')
line.GetDisplayNode().SetLineColorFadingStart(0.1)
line.GetDisplayNode().SetLineColorFadingEnd(0.1)
```

If you want to show the intersection with a marker then you can use a more recent Slicer version and enable showing of direction markers:

<img width="1036" height="1237" alt="Image" src="https://github.com/user-attachments/assets/35d9ebbb-6c3e-4ece-8bc2-3825cd2837b6" />

<img width="1412" height="558" alt="Image" src="https://github.com/user-attachments/assets/ffd7ec00-7625-4d60-a321-d413f48ceb86" />

PR #5513, opened 2021-03-10
ENH: Hide markups in slice views when they do not intersect the slice

Sunderlandkyl:
This commit hides the line actor if all of the points on the markup curve are outside of the slice by 0.5 * slice thickness.
Re #5480


lassoan (commented):
Thank you, this will be good!
Could you create a helper function that is used at both places to reduce code duplication?

Sunderlandkyl:
Ok, made a function to check for slice intersection in vtkSlicerMarkupsWidgetRepresentation2D.

lassoan (approved):
Looks perfect, thank you.

PR #5439, opened 2021-02-07
COMP: Update python packages to latest

jamesobutler:
Python packages were last updated October 25th, 2020. Outdated python packages were discovered using the below code snippet and then versions, hashes and dependencies were determined and updated.

In this PR I'm only updating python packages that Slicer downloads from pypi. This means not updating SimpleITK and VTK python packages.
```
PS C:\Users\JamesButler\AppData\Local\NA-MIC\Slicer 4.13.0-2021-03-09\bin> ./PythonSlicer.exe -m pip list --outdated
Package         Version   Latest    Type
--------------- --------- --------- -----
certifi         2020.6.20 2020.12.5 wheel
chardet         3.0.4     4.0.0     wheel
Deprecated      1.2.10    1.2.11    wheel
dicomweb-client 0.50.2    0.52.0    wheel
GitPython       3.1.11    3.1.14    wheel
idna            2.10      3.1       wheel
numpy           1.19.2    1.19.5    wheel
packaging       20.4      20.9      wheel
Pillow          8.0.1     8.1.2     wheel
pip             20.3.3    21.0.1    wheel
pydicom         2.0.0     2.1.2     wheel
PyGithub        1.53      1.54.1    wheel
PyJWT           1.7.1     2.0.1     wheel
requests        2.24.0    2.25.1    wheel
scipy           1.5.3     1.5.4     wheel
setuptools      50.3.2    54.1.1    wheel
smmap           3.0.4     4.0.0     wheel
urllib3         1.25.11   1.26.3    wheel
wheel           0.35.1    0.36.2    wheel
```

- I'm unable to upgrade `idna`  because `requests` requires idna >=2.5, <3
- I'm unable to upgrade `PyJWT` because `PyGithub` requires pyjwt<2.0
- I'm unable to upgrade `smmap` to latest because `gitdb` has the requirement smmap>=3.0.1,<4, so I have it at 3.0.5.

Packages I'm unable to upgrade to their latest release because they dropped support for Python 3.6 (see https://github.com/Slicer/Slicer/issues/5014#issuecomment-742599890):
- `numpy` >= v1.20.0
- `scipy` >= v1.6.0

PR #5514, opened 2021-03-11
ENH: Update dicomweb-client version to 0.52.0

lassoan:
This version contains fix that allows connection to dcm.js DICOMweb server.


jamesobutler:
This update is also available in #5439 if you want to provide review there. @jcfr gave a previous approval, but I issued additional improvements to the updater script

pieper (approved):
Do you mean server.dcmjs.org?  Not clear from the commit what was fixed so maybe a clearer commit message or link to the dicomweb-client commit could help.

In general I agree with @jamesobutler that this will be better addressed as part of the update script, but also fine to merge this independently if it fixes an issue.

lassoan:
I meant compatibility with this server: https://github.com/dcmjs-org/dicomweb-server

It is not urgent, so  #5439 gets integrated soon then it is fine. I'm closing this so that we can focus on #5439.

Issue #5261, opened 2020-10-23
Save length unit in markups files


lassoan:
"Length" unit of the scene must be saved into markups json files. See more information here: https://discourse.slicer.org/t/add-unit-to-the-fcsv-header/13716/2

jamesobutler:
I believe this issue is now complete as of @cpinter's commit 38aa3b2e471a8a517c3ae1809fce13d24fb6ec1b. The following is a section from a mrk.json file created using Slicer 4.13.0-2020-12-29.
```json
            "measurements": [
                {
                    "name": "length",
                    "enabled": true,
                    "value": 145.2813077003837,
                    "printFormat": "%-#4.4gmm"
                }
            ],
```

lassoan:
This issue refers to saving unit of coordinate values (length unit of the scene) into the markups, which has not been implemented yet.

jamesobutler:
I see. "Length" seemed to indicate that it was the value that was needing to be saved when here specifically it is the "Units" that we want to also save.

Line Node can have a measurement length value of "145" which is in units of "mm".
Angle Node can have measurement angle value of "45" which is in units of "degrees".

SlicerMorph:
Just a reminder that as of latest stable this is still open issue.  See a single markups fiducial saved. Just looking at the contents of this file, there is no way to tell that the coordinates reported are in millimeters. 

```
{
    "@schema": "https://raw.githubusercontent.com/slicer/slicer/master/Modules/Loadable/Markups/Resources/Schema/markups-schema-v1.0.0.json#",
    "markups": [
        {
            "type": "Fiducial",
            "coordinateSystem": "LPS",
            "locked": false,
            "labelFormat": "%N-%d",
            "controlPoints": [
                {
                    "id": "1",
                    "label": "F-1",
                    "description": "",
                    "associatedNodeID": "",
                    "position": [-284.4686648501363, -5.684341886080802e-14, 60.49046321525884],
                    "orientation": [-1.0, -0.0, -0.0, -0.0, -1.0, -0.0, 0.0, 0.0, 1.0],
                    "selected": true,
                    "locked": false,
                    "visibility": true,
                    "positionStatus": "defined"
                }
            ],
            "measurements": [],
            "display": {
                "visibility": true,
                "opacity": 1.0,
                "color": [0.4, 1.0, 0.0],
                "selectedColor": [0.29617761501487757, 0.013031204699778745, 0.9606012054627299],
[...]

PR #5516, opened 2021-03-11
ENH: Save coordinate system unit to markups json files

lassoan:
Also improve saving of units into markups measurements and make error message display during loading.

see https://github.com/Slicer/Slicer/issues/5261

### DIFF
--- a/Modules/Loadable/Markups/Resources/Schema/markups-schema-v1.0.1.json
+++ b/Modules/Loadable/Markups/Resources/Schema/markups-schema-v1.0.1.json
@@ -1,0 +1,537 @@
+    {
+        "$schema": "http://json-schema.org/draft-07/schema",
+        "$id": "https://raw.githubusercontent.com/slicer/slicer/master/Modules/Loadable/Markups/Resources/Schema/markups-v1.0.1-schema.json#",
+        "type": "object",
+        "title": "Schema for storing one or more markups",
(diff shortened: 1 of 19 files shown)


Modules/Loadable/Markups/Resources/Schema/markups-schema-v1.0.1.json
```diff
@@ -0,0 +1,532 @@
...
+                                    "anyOf": [
+                                        {
+                                            "type": "string",
+                                            "title": "Units of control point coordinates",
+                                            "description": "Control point coordinate values are specified in this length unit. Specified in UCUM.",
+                                            "default": "mm",
+                                            "enum": ["mm", "um"]
+                                        },
+                                        {
+                                            ttt
```
Sunderlandkyl, line 66: Remove?

lassoan: Good catch thank you! It seems I've not committed the final version.

lassoan:
I've made some simplifications in how units in measurements are specified (moved more into the measurement base class) and added one more test, @Sunderlandkyl please have a look. Thanks!

Sunderlandkyl (approved):
I've never really looked at the vtkMRMLMeasurement code, but it looks good to me.

PR #5518, opened 2021-03-11
BUG: Fix missing ITKv5 migration items for MorphologicalContourInterpolator

jamesobutler:
```
BUG: Fix missing ITKv5 migration items for MorphologicalContourInterpolator

Updated itkMorphologicalContourInterpolator.h based on functional changes as of (https://github.com/KitwareMedical/ITKMorphologicalContourInterpolation, version 5169b16186a33042274968cca7add9a59cc56d48).

Commits included here:
Dženan Zukić (1):
      COMP: use STL's unordered map instead of KWSYS's hash map

Mathew Seng (1):
      STYLE: Rename ITK_DISALLOW_COPY_AND_ASSIGN to ITK_DISALLOW_COPY_AND_MOVE
```


jamesobutler:
Changes from `COMP: use STL's unordered map instead of KWSYS's hash map` helped fix this build error I was running into:
image

jamesobutler:
Note that I did not copy itkMorphologicalContourInterpolator.h and itkMorphologicalContourInterpolator.hxx directly from https://github.com/KitwareMedical/ITKMorphologicalContourInterpolation as most changes were just changes to fit into ITK's style guide.

dzenanz (dismissed):

dzenanz:
Well now is a fine time to synchronize this copy with the upstream. The style changes should not hurt, and it would make it clear that these few files can just be copied when updated. Which also begs the question: why not just enable the module when building ITK for Slicer?

jamesobutler:
I've added a commit to switch to building MorphologicalContourInterpolator as a ITK remote module. I'm guessing it was included in the vtkITK directory because when it was originally added it wasn't available like that?

lassoan (approved):
This is perfect, thank you!

This filter was added to Slicer's repository because at the time it was not available in Slicer's ITK version.

SuperBuild/External_ITK.cmake
```diff
@@ -104,7 +104,6 @@ if(NOT DEFINED ITK_DIR AND NOT Slicer_USE_SYSTEM_${proj})
     -DITK_LEGACY_REMOVE:BOOL=OFF   #<-- Allow LEGACY ITKv4 features for now.
     -DITK_LEGACY_SILENT:BOOL=ON    #<-- Silence for initial ITKv5 migration.
```
dzenanz, line 105: Probably turn this OFF sometime. Is now a good time?

SuperBuild/External_ITK.cmake
```diff
@@ -146,6 +145,8 @@ if(NOT DEFINED ITK_DIR AND NOT Slicer_USE_SYSTEM_${proj})
       -DModule_ITKReview:BOOL=ON
       -DModule_MGHIO:BOOL=ON
       -DModule_ITKIOMINC:BOOL=ON
+      -DModule_MorphologicalContourInterpolation:Bool=ON
```
dzenanz, line 148: Maybe `BOOL` uppercase to be consistent with the other specifications?

jamesobutler: Done

SuperBuild/External_ITK.cmake
jamesobutler: Yeah I was thinking the same thing. Was planning to do this next to see what gets shown. Same with trying `ITK_LEGACY_REMOVE` as ON.

dzenanz: Maybe just silence now, and legacy when updating to 5.2.0 final? Flipping this switch should be its own commit, for ease of finding it in the log.

jamesobutler: Currently prepping testing with latest ITK with 5.2.0 almost released. Just been having to solve these various build issues with using Visual Studio 16.9 to build Slicer and ITK.

dzenanz (approved):

lassoan:
Looks good to me. @jcfr would you like to have a look?

jcfr:
Thanks everyone for working on this :100:

PR #5520, opened 2021-03-12
BUG: Update VTK to fix crash and avoid process output polluting by debug leaks

jcfr:
List of changes:

```
$ git shortlog e88e47c967..1076d8bc00 --no-merges
Andras Lasso (2):
      [Backport] Prevent vtkDebugLeaks from polluting process output if no leaks are found
      [Backport] Fix crash in vtkOpenGLFramebufferObject::GetMultiSamples()
```


jcfr:
@lassoan Is there discourse discussion or issue we should reference ?

lassoan:
I don't think there were any discussions about these (just in the VTK merge request).

PR #5523, opened 2021-03-12
ENH: Improve handling of internal transforms for ROI and plane markups

Sunderlandkyl:
List of changes:
- Renamed "ROI"/"Plane" space to "Object".
- Renamed "Local" space to "Node".
- Improved handling of ROI ObjectToWorld matrix.
  - ObjectToWorld is a matrix generated from ObjectToNode and NodeToWorld transforms that will always represent the transform from the axis-aligned "Object" space, to world space, ensuring that the ROI axes remain orthogonal.
- Moved scale handle indices from vtkMRMLMarkupsROINode to vtkMRMLMarkupsROIDisplayNode.
- Add DLL export to interaction handle pipelines so that they can be inherited by subclasses in other extensions.
- Add vtkMRMLMarkupsROINode::IsPointInROI() and IsPointInROIWorld() functions.
- Added vtkMRMLMarkupsNodeTest5 to test ROI functions, applying transforms, and transform hardening.

### DIFF
--- a/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsPlaneNode.h
+++ b/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsPlaneNode.h
@@ -39,10 +39,18 @@
 /// Markups is intended to be used for manual marking/editing of point positions.
 ///
 /// Coordinate systems used:
-///   - Local: Local coordinates
-///   - World: All parent transforms on node applied to local.
-///   - Plane: Plane coordinate space (Origin of plane at 0,0,0, XYZ axis aligned to XYZ unit vectors).
-///            Can have additional offset/rotation compared to local.
+///   - Object: Coordinate system of the plane rectangle. Computed from Base coordiante system by ObjectToBaseMatrix.
+///     Origin of the coordinate system is in the plane rectangle center.
+///     X and Y axes of the coordinate system are parallel to the first and second axes of the plane rectangle.
+///     Z axis of the coordinate system is the plane normal.
+///   - Base: Coordinate system computed from markup control point positions.
+///     Origin of the coordinate system is in the first control point.
+///     X coordinate system is the direction from the first control point to the second.
+///     Y coordinate system is orthogonal to the X coordinate system, in the plane specified by the first 3 control points.
+///     Z axis is the cross product of X and Y coordinate system.
+///   - Node: Coordinate system of the markup node. Coordinates of the control points are stored in this coordinate system.
+///   - World: Patient coordinate system (RAS). Transform between Node and World
+///     coordinate systems are defined by the parent transform of the node.
 /// \ingroup Slicer_QtModules_Markups
 class  VTK_SLICER_MARKUPS_MODULE_MRML_EXPORT vtkMRMLMarkupsPlaneNode : public vtkMRMLMarkupsNode
 {
@@ -109,14 +117,12 @@ public:
   vtkSetVector6Macro(PlaneBounds, double);
 
   /// The normal vector for the plane.
-  /// Calculated as the vector perpendicular to the plane containing the 3 markup points, and transformed by the PlaneToPlaneOffsetMatrix.
   void GetNormal(double normal[3]);
   void SetNormal(const double normal[3]);
   void GetNormalWorld(double normal[3]);
   void SetNormalWorld(const double normal[3]);
 
   /// The origin of the plane.
-  /// Calculated as the location of the 0th markup point, and translated by the PlaneToPlaneOffsetMatrix.
   void GetOrigin(double origin[3]);
   void SetOrigin(const double origin[3]);
   void GetOriginWorld(double origin[3]);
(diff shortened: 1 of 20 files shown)


lassoan (commented):
Thank you, this will be good, I just added a few suggestions for making code simpler (less duplications) and robust (ensure orthogonal axes...).

lassoan:
I had to merge the content of this PR in https://github.com/Slicer/Slicer/pull/5547, but I don't close this PR because the comments should be still addressed.

In addition to those comments, two more:

1. Please add an API to set the ROIToLocalMatrix. In vtkSlicerCropVolumeLogic I had to add all these lines to set the ROI orientation:

```
    MRMLNodeModifyBlocker blocker(markupsROI);
    markupsROI->GetROIToLocalMatrix()->DeepCopy(volumeIJKToROILocal);
    markupsROI->UpdateControlPointsFromROI();
    markupsROI->Modified();
```

2. The concept of having a "local" and a "ROI" coordinate system is not very clear because "ROI" coordinate system supposed to mean the local transformed by LocalToROI; but when I write "ROI" in the code then it can also be interpreted as "ROI node", which is the "local" coordinate system. Also, the same concept comes up for plane ROI, transform, slice intersection widget, etc., so it would be nice if we could use the same terms for all cases. For example, the node coordinate system could be "node" and the transformed be "local" (but this one would be very confusing in the short term because "local" currently refers to "node"); or "local" (the same way as now) and "object" (object would mean ROI, plane, ...), or ...?

Sunderlandkyl:
> The concept of having a "local" and a "ROI" coordinate system is not very clear because "ROI" coordinate system supposed to mean the local transformed by LocalToROI; but when I write "ROI" in the code then it can also be interpreted as "ROI node", which is the "local" coordinate system. Also, the same concept comes up for plane ROI, transform, slice intersection widget, etc., so it would be nice if we could use the same terms for all cases. For example, the node coordinate system could be "node" and the transformed be "local" (but this one would be very confusing in the short term because "local" currently refers to "node"); or "local" (the same way as now) and "object" (object would mean ROI, plane, ...), or ...?

Probably I didn't describe it well enough in documentation, but this was my original intent behind the "ROI" and "Plane" coordinate systems:

In the ROI coordinate system, the center of the ROI sits at [0,0,0], and is axis-aligned. This way, it's easy to calculate if a point is inside the ROI by transforming the point to ROI coordinates and checking that it is within the axis-aligned bounds. Similarly, in the the Plane coordinate system, the center of the plane sits at [0,0,0], is axis aligned, and its normal points along the Z-axis.

The ROI/Plane coordinates could both be renamed to "Object" for consistency. Do you have a suggestion of how the coordinate systems should be organized?

lassoan:
> The ROI/Plane coordinates could both be renamed to "Object" for consistency. Do you have a suggestion of how the coordinate systems should be organized?

No need for reorganization - the ROI/Plane coordinate systems are good. The only reason they are confusing is that the node itself is called ROI/Plane, too, and "Local" name does not clarify things either. Instead of GetROIToLocalMatrix(), what do you think about something like these: GetObjectToNodeMatrix, GetLocalToNodeMatrix, GetInternalToNodeMatrix, GetAxesToNodeMatrix?

Sunderlandkyl:
@lassoan I've made a number of changes, including renaming the ROI/Plane space to "Object", and the "Local" space to "Node".

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsPlaneNode.h
```diff
@@ -39,9 +39,9 @@
 /// Markups is intended to be used for manual marking/editing of point positions.
 ///
 /// Coordinate systems used:
-///   - Local: Local coordinates
+///   - Node: Local node coordinates
 ///   - World: All parent transforms on node applied to local.
```
lassoan, line 43:
```suggestion
///   - World: All parent transforms on node applied to Node coordinates.
```

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsPlaneNode.h
```diff
@@ -39,9 +39,9 @@
...
-///   - Plane: Plane coordinate space (Origin of plane at 0,0,0, XYZ axis aligned to XYZ unit vectors).
+///   - Object: Plane coordinate space (Origin of plane at 0,0,0, XYZ axis aligned to XYZ unit vectors).
```
lassoan, line 44: Origin by definition is (0,0,0), so that's not helping. Is the origin the center of the rectangle?

Specify that X and Y axes are parallel to the first and second axes of the plane's rectangle and Z axis is the plane normal.

lassoan, line 44: Define ObjectOffset coordinate system, too.

lassoan (commented):
Thank you, the renaming looks good. I've added a few comments. Have you responded to all my previous comments?

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsROINode.cxx
```diff
@@ -511,6 +604,78 @@ void vtkMRMLMarkupsROINode::SetSize(double x, double y, double z)
...
+
+  double yAxis_Local[3] = { 0.0, 1.0, 0.0 };
+  double yAxis_World[3] = { 0.0, 1.0, 0.0 };
+  this->GetYAxisLocal(yAxis_Local);
+  localToWorldTransform->TransformVectorAtPoint(center_Local, yAxis_Local, yAxis_World);
+
+  double zAxis_Local[3] = { 0.0, 0.0, 1.0 };
+  double zAxis_World[3] = { 0.0, 0.0, 1.0 };
+  this->GetZAxisLocal(zAxis_Local);
+  localToWorldTransform->TransformVectorAtPoint(center_Local, zAxis_Local, zAxis_World);
```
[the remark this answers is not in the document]
Sunderlandkyl: Added a new function GenerateOrthogonalMatrix, which ensures that the transformed axes form an orthogonal right-handed coordinate system.

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsROINode.cxx
```diff
@@ -138,6 +138,62 @@ void vtkMRMLMarkupsROINode::CreateDefaultDisplayNodes()
...
+  double center_Local[3] = { 0.0, 0.0, 0.0 };
+  this->GetCenter(center_Local);
+
+  double xAxis_Local[3] = { 1.0, 0.0, 0.0 };
+  this->GetXAxisLocal(xAxis_Local);
+  transform->TransformVectorAtPoint(center_Local, xAxis_Local, xAxis_Local);
+  double xAxisSize = vtkMath::Normalize(xAxis_Local);
+
+  double yAxis_Local[3] = { 0.0, 1.0, 0.0 };
+  this->GetYAxisLocal(yAxis_Local);
```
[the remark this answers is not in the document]
Sunderlandkyl: Let me know if the GenerateOrthogonalMatrix simplifies the process enough. Internally it still does some vector-by-vector operations, but now it's all collected in one place. I'm not sure if there's a way around it since the applied transform may not be linear.

lassoan (approved):
Thank you! It all looks good.

I've ended up renaming plane "ObjectOffset" coordinate system to "Base", resulting in World->Node->Base->Object coordinate systems.

PR #5524, opened 2021-03-13
BUG: Fix import of python modules on Linux selectively stripping libraries

### LINKED ISSUE #5474: Importing the numpy C-extensions failed. [slicer 4.11.20200930, archlinux]
When starting slicer I get a lot of errors concerning the numpy C-extensions.
...
Importing the numpy C-extensions failed.
...
Original error was: 

```
libgfortran-2e0d59d6.so.5.0.0:
ELF load command address/offset not properly aligned
```

Workaround:

```
install -c -s /usr/lib/libgfortran.so.5.0.0 \
 /opt/3dslicer/lib/Python/lib/python3.6/site-packages/numpy.libs/libgfortran-2e0d59d6.so.5.0.0
```

I also did:

```
install -c -s /usr/lib/gcc/x86_64-pc-linux-gnu/5.5.0/libgfortran.so.3.0.0 \
 /opt/3dslicer/lib/Python/lib/python3.6/site-packages/scipy.libs/libgfortran-ed201abd.so.3.0.0
```

After this slicer is starting without error messages.

Regards
Wolfgang

jcfr:
This commit fixes "ELF load command address/offset not properly aligned" runtime
error reported when attempting to import some python modules.

Instead of systematically stripping all libraries, this commit updates the
packaging rules Strip symbols of selected libraries for which ones
(1) stripping has a significant impact
and (2)  stripping does not prevent the successful import of the module.

Example of error reported:

```
  $ ./bin/PythonSlicer -c "from scipy.special import _ufuncs as u; from pprint import pprint as pp; pp(dir(u))"
  [...]
  ELF load command address/offset not properly aligned
```

Fixes #5474

### DIFF
--- a/CMake/SlicerBlockInstallPython.cmake
+++ b/CMake/SlicerBlockInstallPython.cmake
@@ -44,8 +44,12 @@ To create a Slicer package including python libraries, you can *NOT* provide you
     REGEX "/test/" EXCLUDE
     ${extra_exclude_pattern}
     )
+  # Strip symbols of selected libraries for which ones (1) stripping has a
+  # significant impact and (2) stripping does not prevent the import of
+  # of the module.
+  # See https://github.com/Slicer/Slicer/issues/5474
   slicerStripInstalledLibrary(
-    PATTERN "${Slicer_INSTALL_ROOT}lib/Python/${PYTHON_STDLIB_SUBDIR}/*.so"
+    PATTERN "${Slicer_INSTALL_ROOT}lib/Python/${PYTHON_STDLIB_SUBDIR}/_SimpleITK.*.so"
     COMPONENT Runtime
     )
 


jcfr:
Copied from https://github.com/Slicer/Slicer/issues/5474#issuecomment-797827128:

> Since systematically stripping  symbols from all shared library found in  `python-install/lib/python3.6/site-packages` is leading to import error, we will instead selectively strip symbols.
> 
> Here is the size of the `python-install/lib/python3.6/site-packages` directory obtained after applying different stipping strategy:
> 
> ```
> 419M: No stripping
> 393M: Stripping all *.so libraries
> 398M: Stripping only _SimpleITK*.so library
> ```
> 
> _Size obtained on a workstation wherer Slicer was compiled using gcc 5.2.1_

jcfr:
After testing locally, I confirm that only SimpleITK python module is stripped and that the other libraries are not.

jcfr:
Since local testing confirmed the change addresses the issue, to ensure the next round of preview build include the fix, I am bypassing our standard review process.

lassoan:
Awesome! We would just need to fix the display configuration on the Linux dashboard machine to see (most of) the tests passing.

sjh26:
Yep, I am still working on sorting the linux tests

PR #5529, opened 2021-03-17
BUG: Restore collapsed state of Advanced section of Markups display w…

cpinter:
…idget

It has been probably accidentally left uncollapsed earlier when working on improvements in the widget.


cpinter:
Not sure if the new Interaction handles section is uncollapsed by default but in any case I haven't touched that.

lassoan (approved):

cpinter:
Thanks for merging!

PR #5530, opened 2021-03-17
ENH: Use flow layout in Sample Data module

cpinter:
The current three-column box layout resizes the module panel if it was narrower than what could fit three thumbnails, which was an inconvenience especially on smaller monitors or smaller window sizes, because the user had to manually resize the panel after switching to another module if they wanted to see the viewers in the usual size.
By using flow layout the thumbnails naturally fit the available horizontal space.

https://user-images.githubusercontent.com/1325980/111494297-40cb9680-8736-11eb-988c-18a8ead0e61e.mp4

### DIFF
--- a/Modules/Scripted/SampleData/SampleData.py
+++ b/Modules/Scripted/SampleData/SampleData.py
@@ -288,7 +288,6 @@ class SampleDataWidget(ScriptedLoadableModuleWidget):
     Download buttons are organized in collapsible GroupBox with one GroupBox
     per category.
     """
-    numberOfColumns = 3
     iconPath = os.path.join(os.path.dirname(__file__).replace('\\','/'), 'Resources','Icons')
     desktop = qt.QDesktopWidget()
     mainScreenSize = desktop.availableGeometry(desktop.primaryScreen)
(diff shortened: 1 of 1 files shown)


pieper (dismissed):
Thank you thank you thank you!  This has been an annoyance.

This is good as is, but a small side comment I noticed when I looked into this is that the size of the thumbnails seems to be based on the [size of the user's screen](https://github.com/Slicer/Slicer/blob/aab9e013f450bd95291d5d5c69952091ebeda68b/Modules/Scripted/SampleData/SampleData.py#L293-L294), which can sometimes be very big compared to the app window size (e.g. on a 4k monitor).  On reflection I think the thumbnail size should be based on the app window size.

What do you think?  It can be a distinct change from this PR.

cpinter:
@pieper I made the change. The thumbnails are more or less the same size with the window size I usually use.

jcfr (approved):

pieper:
Thanks again, this is great.  Thanks for the review @jcfr 👍

lassoan:
Thank you @cpinter. Could you check if you can fix this test easily?
https://slicer.cdash.org/test/16920394

cpinter:
Yes I'll check.

cpinter:
Yes it was an easy fix. See https://github.com/Slicer/Slicer/pull/5541

PR #5531, opened 2021-03-17
BUG: launcher executable now using app name to compose the filepath

fbordignon:
Untested
SlicerJupyter is using this function to determine the Slicer Launcher file path and it is failing for a custom app name.

### DIFF
--- a/Base/QTCore/qSlicerCoreApplication.cxx
+++ b/Base/QTCore/qSlicerCoreApplication.cxx
@@ -1314,7 +1314,8 @@ QString qSlicerCoreApplication::launcherExecutableFilePath()const
     {
     return QString();
     }
-  return this->slicerHome() + "/Slicer" + qSlicerUtils::executableExtension();
+  QString appName = this->applicationName().replace("-tmp", "");
+  return this->slicerHome() + "/" + appName + qSlicerUtils::executableExtension();
 }
 
 //-----------------------------------------------------------------------------


jcfr:
Thanks for the patch :pray:  This makes sense.

Few minor comments:
* Is there an issue in https://github.com/Slicer/SlicerJupyter or forum discussion we could reference ?
* Could you associated your email with your GitHub account so that your are listed as a Slicer contributor ?  See https://github.com/Slicer/Slicer/graphs/contributors

Commit message to use for integration:

```
BUG: Fix launcher executable file path in custom application

This commit fixes function "qSlicerCoreApplication::launcherExecutableFilePath()" originally
introduced in a19473a99 (ENH: Add "launcherExecutableFilePath()" and "launcherSettingsFilePath()"
methods to qSlicerCoreApplication) and is a follow-up of 371cd469c (ENH: Extend application API
adding environment()/startupEnvironment()).
```

fbordignon:
1 - I don't think so. I did not create an issue because I figure it was easier to just open a PR
2 - Ok done

Do I need to update the commit message as you suggested or it will appear on the merge commit? Thanks

jcfr:
> Do I need to update the commit message as you suggested or it will appear on the merge commit?

No need to worry about it, I will do a "squash + merge" and update accordingly.

jcfr:
Change has been locally tested, I will move forward with the integration.

Issue #5486, opened 2021-02-26
Curvature measurement circular calls


cpinter:
When the curvature measurement is enabled, an error is thrown
`UpdateInformation invoked during another request.  Returning failure to algorithm vtkCurveMeasurementsCalculator(000001B69EF24640)`, and calculation is slower than before.

The problem was introduced in commit "ENH: Observe markup measurements to trigger necessary updates" (add171b12a282be1434a95432a61fd4c94d26007) by adding in `vtkMRMLMarkupsNode::ProcessMRMLEvents` a handle of vtkMRMLMeasurement events to call UpdateAllMeasurements. See call stack below in comment.

A solution could be to either use different events for modified input and output, or use other event like ControlPointsModified or similar.

## Steps to reproduce

- Add curve
- Enable curvature measurement

## Expected behavior

- No error
- Faster calculation

## Environment
- Slicer version: Slicer-4.13.0-2020-02-26
- Operating system: Windows

cpinter:
>	vtkCommon-8.2.dll!vtkExecutive::CheckAlgorithm(const char  method, vtkInformation  request) Line 806	C++
 	vtkCommon-8.2.dll!vtkDemandDrivenPipeline::UpdateInformation() Line 379	C++
 	vtkCommon-8.2.dll!vtkStreamingDemandDrivenPipeline::Update(int port, vtkInformationVector * requests) Line 398	C++
 	vtkCommon-8.2.dll!vtkStreamingDemandDrivenPipeline::Update(int port) Line 392	C++
 	vtkCommon-8.2.dll!vtkAlgorithm::Update(int port) Line 1459	C++
 	vtkCommon-8.2.dll!vtkAlgorithm::Update() Line 1453	C++
 	vtkSlicerMarkupsModuleMRML.dll!vtkMRMLMarkupsCurveNode::UpdateMeasurementsInternal() Line 1172	C++
 	vtkSlicerMarkupsModuleMRML.dll!vtkMRMLMarkupsNode::UpdateAllMeasurements() Line 2086	C++
 	vtkSlicerMarkupsModuleMRML.dll!vtkMRMLMarkupsNode::ProcessMRMLEvents(vtkObject  caller, unsigned long event, void  callData) Line 258	C++
 	vtkSlicerMarkupsModuleMRML.dll!vtkMRMLMarkupsCurveNode::ProcessMRMLEvents(vtkObject  caller, unsigned long event, void  callData) Line 1220	C++
[...]

PR #5535, opened 2021-03-18
BUG: Fix circular markups measurement update calls

cpinter:
The custom modified event InputDataModifiedEvent has been introduced in vtkMRMLMeasurement that is invoked only when such members are modified that should trigger update of the measurement. By observing this event for updating the measurements, circular calls are prevented.

A concrete effect is that curvature computation and scalar display update is fast again.

Fixes https://github.com/Slicer/Slicer/issues/5486

### DIFF
--- a/Libs/MRML/Core/vtkMRMLMeasurement.cxx
+++ b/Libs/MRML/Core/vtkMRMLMeasurement.cxx
@@ -153,6 +153,54 @@ void vtkMRMLMeasurement::Copy(vtkMRMLMeasurement* src)
 }
 
 //----------------------------------------------------------------------------
+void vtkMRMLMeasurement::SetEnabled(bool enabled)
+{
+  vtkDebugMacro(<< this->GetClassName() << " (" << this << "): setting Enabled to " << enabled);
+  if (this->Enabled != enabled)
+    {
+    this->Enabled = enabled;
+    this->Modified();
+    this->InvokeEvent(InputDataModifiedEvent);
+    }
+}
+
+//----------------------------------------------------------------------------
+void vtkMRMLMeasurement::EnabledOn()
+{
+  this->SetEnabled(true);
+}
+
+//----------------------------------------------------------------------------
+void vtkMRMLMeasurement::EnabledOff()
+{
+  this->SetEnabled(false);
+}
+
+//----------------------------------------------------------------------------
+void vtkMRMLMeasurement::SetUnits(std::string units)
+{
+  vtkDebugMacro(<< this->GetClassName() << " (" << this << "): setting Units to " << units);
+  if (this->Units != units)
+    {
+    this->Units = units;
+    this->Modified();
+    this->InvokeEvent(InputDataModifiedEvent);
+    }
+}
+
+//----------------------------------------------------------------------------
+void vtkMRMLMeasurement::SetPrintFormat(std::string format)
+{
+  vtkDebugMacro(<< this->GetClassName() << " (" << this << "): setting PrintFormat to " << format);
+  if (this->PrintFormat != format)
+    {
+    this->PrintFormat = format;
+    this->Modified();
+    this->InvokeEvent(InputDataModifiedEvent);
+    }
+}
+
+//----------------------------------------------------------------------------
 void vtkMRMLMeasurement::SetQuantityCode(vtkCodedEntry* entry)
 {
   if (!entry)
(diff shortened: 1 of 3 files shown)


lassoan (dismissed):
Looks great, thank you!

Libs/MRML/Core/vtkMRMLMeasurement.cxx
```diff
@@ -152,6 +152,54 @@ void vtkMRMLMeasurement::Copy(vtkMRMLMeasurement* src)
 #endif
 }
 
+//----------------------------------------------------------------------------
+void vtkMRMLMeasurement::SetEnabled(bool enabled)
+{
+  vtkDebugMacro(<< this->GetClassName() << " (" << this << "): setting Enabled to " << enabled);
+  if (this->Enabled != enabled)
+  {
```
jcfr, line 160:
```suggestion
    {

File: Libs/MRML/Core/vtkMRMLMeasurement.cxx [hunk 2]
```diff
@@ -152,6 +152,54 @@ void vtkMRMLMeasurement::Copy(vtkMRMLMeasurement* src)
...
+    this->Enabled = enabled;
+    this->Modified();
+    this->InvokeEvent(InputDataModifiedEvent);
+  }
```
jcfr, line 164:
```suggestion
    }
```

Libs/MRML/Core/vtkMRMLMeasurement.cxx
```diff
@@ -152,6 +152,54 @@ void vtkMRMLMeasurement::Copy(vtkMRMLMeasurement* src)
...
+{
+  this->SetEnabled(false);
+}
+
+//----------------------------------------------------------------------------
+void vtkMRMLMeasurement::SetUnits(std::string units)
+{
+  vtkDebugMacro(<< this->GetClassName() << " (" << this << "): setting Units to " << units);
+  if (this->Units != units)
+  {
```
jcfr, line 184:
```suggestion
    {
```

Libs/MRML/Core/vtkMRMLMeasurement.cxx
```diff
@@ -152,6 +152,54 @@ void vtkMRMLMeasurement::Copy(vtkMRMLMeasurement* src)
...
+    this->Units = units;
+    this->Modified();
+    this->InvokeEvent(InputDataModifiedEvent);
+  }
```
jcfr, line 188:
```suggestion
    }
```

Libs/MRML/Core/vtkMRMLMeasurement.cxx
```diff
@@ -152,6 +152,54 @@ void vtkMRMLMeasurement::Copy(vtkMRMLMeasurement* src)
...
+//----------------------------------------------------------------------------
+void vtkMRMLMeasurement::SetPrintFormat(std::string format)
+{
+  vtkDebugMacro(<< this->GetClassName() << " (" << this << "): setting PrintFormat to " << format);
+  if (this->PrintFormat != format)
+  {
+    this->PrintFormat = format;
+    this->Modified();
+    this->InvokeEvent(InputDataModifiedEvent);
+  }
```
jcfr, line 200:
```suggestion
    }
```

Libs/MRML/Core/vtkMRMLMeasurement.cxx
```diff
@@ -152,6 +152,54 @@ void vtkMRMLMeasurement::Copy(vtkMRMLMeasurement* src)
...
+}
+
+//----------------------------------------------------------------------------
+void vtkMRMLMeasurement::SetPrintFormat(std::string format)
+{
+  vtkDebugMacro(<< this->GetClassName() << " (" << this << "): setting PrintFormat to " << format);
+  if (this->PrintFormat != format)
+  {
```
jcfr, line 196:
```suggestion
    {
```

cpinter: Oops, right! Coming up.

cpinter: By the way we should really switch to aligned brackets...

cpinter: Fix done.

jcfr:
I am wondering if the problem could be fixed by adding a fuzzy compare in these location:

https://github.com/Slicer/Slicer/blob/bcc9a8d5d0bada278944395a22e49cc7dfa18d0e/Libs/MRML/Core/vtkMRMLMeasurement.cxx#L307

https://github.com/Slicer/Slicer/blob/bcc9a8d5d0bada278944395a22e49cc7dfa18d0e/Libs/MRML/Core/vtkMRMLMeasurement.cxx#L348

lassoan:
> I am wondering if the problem could be fixed by adding a fuzzy compare in these location:
> 
> https://github.com/Slicer/Slicer/blob/bcc9a8d5d0bada278944395a22e49cc7dfa18d0e/Libs/MRML/Core/vtkMRMLMeasurement.cxx#L307
> 
> https://github.com/Slicer/Slicer/blob/bcc9a8d5d0bada278944395a22e49cc7dfa18d0e/Libs/MRML/Core/vtkMRMLMeasurement.cxx#L348

Comparison can be quite complicated (how do you determine the tolerance?) and expensive (result can be per curve point).

lassoan (approved):
I would merge this as is, as it is a clean, efficient solution to the repeated measurement updates.

PR #5536, opened 2021-03-19
ENH: Make default slice view orientations configurable

lassoan:
User can choose between radiology and neurology view directions (patient left/right on screen right).

### DIFF
--- a/Base/QTGUI/Resources/UI/qSlicerSettingsViewsPanel.ui
+++ b/Base/QTGUI/Resources/UI/qSlicerSettingsViewsPanel.ui
@@ -127,6 +127,20 @@
             </item>
            </widget>
           </item>
+          <item row="2" column="0">
+           <widget class="QLabel" name="SliceViewOrientationLabel">
+            <property name="text">
+             <string>View orientation:</string>
+            </property>
+           </widget>
+          </item>
+          <item row="2" column="1">
+           <widget class="ctkComboBox" name="SliceViewOrientationComboBox">
+            <property name="currentIndex">
+             <number>-1</number>
+            </property>
+           </widget>
+          </item>
          </layout>
         </widget>
        </item>
--- a/Base/QTGUI/qSlicerSettingsViewsPanel.cxx
+++ b/Base/QTGUI/qSlicerSettingsViewsPanel.cxx
@@ -27,6 +27,7 @@
 
 // CTK includes
 #include <ctkVTKAbstractView.h>
+#include <ctkComboBox.h>
 
 // --------------------------------------------------------------------------
 // qSlicerSettingsViewsPanelPrivate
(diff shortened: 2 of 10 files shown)


pieper:
I'm really torn on this.  We have been very consistently using the radiological convention and it has allowed us to be unambiguous to answer orientation questions.  Given the history of this issue, I think that before we enable this switch we should turn the orientation marker by default and even add L and R (and other letters or letter combinations for reformats) to the sides of the slice views so it is very explicit.

lassoan:
I agree that placing axis letters at top/bottom/left/right of the slice views (double or triple letters in case of oblique views) would be a more common and nicer way of representing orientations than the arrow/box/human figure, we should implement that at some point for sure. They are also somewhat more subtle than the orientation marker in the corner, so we could maybe enable orientation marker by default if we use this style.

I'm not sure if we need to enforce orientation marker display more if we expose this option. The option will only be found if someone looks for it and right next to it there is the option of showing the orientation marker. Also, someone can already use reformat module to flip slice orientation.

By the way, this is the post that triggered this work: https://twitter.com/gadevenyi/status/1372200767994597385

Most other image viewer software that I've tried makes it easy to flip views very easily. I wanted to check if it is easy to change the default in Slicer and it was not too hard. It just seemed unreasonable to ask users to always go to Reformat module and choose a view, then move a slider, then choose another view and move a slider to get patient-right-is-screen-right orientation.

If you feel very strongly about this then maybe I can hide this from application settings and force users to edit the Slicer.ini file, but I'm not sure if it helped that much overall.

pieper:
This is sort of like nifti files, where the neuroscience community adopted their own convention and it kind of leaked out.  It's definitely an issue in the clinical research world where both radiological and neurological software meet.  I would not be surprised if there have been clinical consequences.  I'm afraid that, for example, someone might send a slicer screenshot and the recipient wouldn't know that the setting had been changed on that slicer version.  I'd feel more comfortable if we implemented the side letters before exposing this option.  This would also help neuroscientist be aware when the images are displayed in radiological convention.

lassoan:
Adding a new orientation marker style would not mean that user would suddenly start using orientation markers. What do you think about automatically enabling the orientation marker if someone changes the slice view orientation? (we could use the arrow style for now and later could add the letters at the border)

pieper:
> What do you think about automatically enabling the orientation marker if someone changes the slice view orientation?

Yes, this would be good.  I think the axis arrows are the most clear (human is too subtle).

gdevenyi:
Origin of the twitter thread here ✋🏻 

I'm okay with hiding it within the config file, and I'd be 👍🏻 for having orientation labels always be present. Coming from MINC-land to NIFTI-land I was surprised at all the possible ambiguities of orientation/display, depending on software and on-disk format (Slicer vs ITKsnap vs FSL(eyes,view) vs MRIcron, MRICrogl) and I agree there's a risk of someone changing the settings and it not being obvious when collaborating.

lassoan:
@pieper Now if the user switches to the non-default view then orientation marker is turned on automatically.

pieper (approved):
Thanks for adding that, this makes me more comfortable with it.

PR #5537, opened 2021-03-19
Several small fixes

lassoan:
This pull request contains 4 unrelated fixes, lumped together to reduce overhead in creating and reviewing the pull request.


pieper (approved):
LGTM.

lassoan:
This .vtk file format incompatibility is really ugly (caused us some embarrassment today when a SlicerHeart tool that has been used successfully clinically suddenly created a file that did not work with their EP guidance system).

The change of file format must be reverted in vtk master, too (at least, by default the old, compatible format should be use), before the reputation of this file format get tainted. I've added a comment on [VTK forum](https://discourse.vtk.org/t/legacy-polydata-file-compatibility/5354/4).

pieper:
Yes, I was following the story on the SlicerHeart notifications.  This kind of backwards compatibility breakage is very bad practice and I agree it needs to be reverted in VTK.

jcfr:
> I agree it needs to be reverted in VTK.

We are currently discussing this, I will follow up when I have more details.

PR #5541, opened 2021-03-20
BUG: Fix SampleData test that failed since last change

cpinter:


lassoan (approved):
Looks good, thank you.

PR #5545, opened 2021-03-23
BUG: Fix logic that determines if a module is built-in

lassoan:
Modules in Extensions-<revision> folder must be always classified as not built-in, even if that folder is within the application folder.
Previously Extensions-<revision> folder only contained extension modules on macOS, but now since applications are fully portable, this folder contains extension modules on all operating systems.


cpinter (approved):
Looks good, thanks so much!

PR #5547, opened 2021-03-24
Make Crop volume module compatible with Markups ROI node

lassoan:
This pull request makes Crop volume work with the new markups ROI. Markups ROI node has many advantages over Annotation ROI, for example that it can be rotated without applying a transform.


Sunderlandkyl (approved):
Thanks, it looks good to me. I'll address the comments in the #5523.

PR #5548, opened 2021-03-25
COMP: Update code to allow both C++11 and C++17 compilation

kerim371:


kerim371:
With those changes I could compile SlicerCAT with C++17 enabled

lassoan:
The pull request has been merged. Thank you for your contribution.

PR #5551, opened 2021-03-26
DOC: Add Glossary of commonly used terms

lassoan:
In response to user request: https://discourse.slicer.org/t/glossary-of-terms/16659/2


Docs/user_guide/getting_started.md
```diff
@@ -149,3 +149,51 @@ Browse the [User Guide](user_interface) section to find quick overview of the ap
...
+- Extent: Range of integer coordinates along 3 axes. Typically defined by 6 values, for IJK axes: I_min, I_max, J_min, J_max, K_min, K_max. Both minimum and maximum values are inclusive, therefore size of an array is (I_max - I_min + 1) x (J_max - J_min + 1) x (K_max - K_min + 1).
+- Fiducial: Represents a point in 3D space. The terms originates from image-guided surgery, where "fiducial markers" are used to mark point positions.
+- Frame: One time point in a time sequence. To avoid ambiguity, this term is not used to refer to a slice of a volume.
+- Geometry: Specify location and shape of an object in 3D space.
+- Intensity: Typically refers to value of a voxel.
+- IJK: Voxel coordinate system axes. Each integer coordinate value combination corresponds to a voxel center.
+- ITK: [Insight Toolkit](https://itk.org/). Software library that Slicer uses for most image processing operations.
+- Labelmap (or labelmap volume, labelmap volume node): Volume node that has discrete (integer) voxel values. Typically each value corresponds to a specific structure or region. This allows compact representation of non-overlapping regions in a single 3D array.
```
jcfr, line 172: Could also indicate the relationship with segment managed in the segment editor ?

Docs/user_guide/getting_started.md
```diff
@@ -149,3 +149,51 @@ Browse the [User Guide](user_interface) section to find quick overview of the ap
...
+- LPS: Left-posterior-superior anatomical coordinate system. Most commonly used coordinate system in medical image computing. Slicer stores all data in LPS coordinate system on disk (and converts to/from RAS when writing to or reading from disk).
+- Markups: Simple geometric objects and measurements that the user can place in viewers. Markups module can be used to create such objects. There are several types, such as fiducial list, line, curve, plane, ROI.
```
jcfr, line 174: Add link to markups module ?

Docs/user_guide/getting_started.md
```diff
@@ -149,3 +149,51 @@ Browse the [User Guide](user_interface) section to find quick overview of the ap
 The application has a large and very friendly and helpful user community. We have poeple who will happy to help with simple questions, such as how to do a specific task in Slicer, and we have a large number of engineering and medical experts who can give you advice with how to solve complex problems.
 
 **If you have any questions, go to the [Slicer forum](https://discourse.slicer.org) and ask us!**
+
+## Glossary
+
+Terms used in various fields of medical and biomedical image computing and clinical images are not always consistent. This section defines terms that are commonly used in 3D Slicer, especially those that may have different meaning in other contexts.
+
+- Annotation: Simple geometric objects and measurements that the user can place in viewers. Annotations module can be used to create such objects, but it is deprecated - being replaced by markups.
+- Bounds: Describes bounding box of a spatial object along 3 axes. Typically defined by 6 floating-point values: X_min, X_max, Y_min, Y_max, Z_min, Z_max.
```
jcfr, line 158: Variables like X_min could instead be listed as `X_min`, ....

Docs/user_guide/getting_started.md
```diff
@@ -149,3 +149,51 @@ Browse the [User Guide](user_interface) section to find quick overview of the ap
...
+- Orientation marker: Arrow, box, or human shaped marker to show axis directions in slice views and 3D views.
+- RAS: Right-anterior-superior anatomical coordinate system. Coordinate system used internally in Slicer. It can be converted to/from LPS coordinate system by inverting the direction of the first two axes.
+- Reference: Can mean many things. It typically means a secondary input (data object, coordinate frame, geometry, etc.) for an operation.
+- Region of interest (ROI): Specifies a box-shaped region in 3D. Can be used for cropping volumes, clipping models, etc.
+- Registration: The process of aligning objects in space. Result of the registration is a transform, which transforms the "moving" object to the "fixed" object.
+- Resolution: Voxel size of a volume, typically specified in mm/pixel. It is rearely used in the user interface because its meaning is slightly misleading: high resolution value means large spacing, which means coarse (low) image resolution.
```
jcfr, line 185: Rearely -> rarely

Docs/user_guide/getting_started.md
```diff
@@ -149,3 +149,51 @@ Browse the [User Guide](user_interface) section to find quick overview of the ap
...
+- Volume (or volume node, scalar volume, image): MRML node storing 3D array of voxels. Indices of the array are typically referred to as IJK. Range of IJK coordinates are called extents. Geometry of the volume is specified by its origin (position of the IJK=(0,0,0) point), spacing (size of a voxel along I, J, K axes), axis directions (direction of I, J, K axes in the reference coordinate system). 2D images are single-slice 3D volumes, with their position and orientation specified in 3D space.
+- [Voxel](https://en.wikipedia.org/wiki/Voxel): One element of a 3D volume. It has a rectangular prism shape. Coordinates of a voxel correspond to position of its center point. Value of a voxel may be a simple scalar value or a vector.
+- Window/level (or window width/window level): Specifies linear mapping of voxel values to brightness of a displayed pixel. Window is the size of the intensity range that is mapped to the full displayable intensity range. Level is the voxel value that is mapped to the center of the full displayable intensity range.
```
jcfr, line 197: Move this one to then to keep the list in alphapeitcal order

Docs/user_guide/getting_started.md
```diff
@@ -149,3 +149,51 @@ Browse the [User Guide](user_interface) section to find quick overview of the ap
...
+
+## Glossary
+
+Terms used in various fields of medical and biomedical image computing and clinical images are not always consistent. This section defines terms that are commonly used in 3D Slicer, especially those that may have different meaning in other contexts.
+
+- Annotation: Simple geometric objects and measurements that the user can place in viewers. Annotations module can be used to create such objects, but it is deprecated - being replaced by markups.
+- Bounds: Describes bounding box of a spatial object along 3 axes. Typically defined by 6 floating-point values: X_min, X_max, Y_min, Y_max, Z_min, Z_max.
+- Brightness/contrast: Specifies linear mapping of voxel values to brightness of a displayed pixel. Brightness is the linear offset, contrast is the multiplier.
+- Cell: Data cells are simple topological elements like lines, polygons, tetrahedra, etc.
+- Coordinate system (or coordinate frame, reference frame, space): Specified by position of origin, axis directions, and distance unit. All coordinate systems in 3D Slicer are right-handed.
```
jcfr, line 161: In a former post, it i recall you also described the pre and post multiple convention used in ITK and VTK, and the relationship with computer graphics and image processing (if I recall). I am wondering if we could integrate that info in the glossary also.

jcfr (commented):
This is great, thanks for working on this 🙏

lassoan:
Thank you @jcfr, I've addressed all your comments.

jcfr (dismissed):

Docs/user_guide/getting_started.md
```diff
@@ -149,3 +149,51 @@ Browse the [User Guide](user_interface) section to find quick overview of the ap
...
+- Extension: A collection of modules that is not bundled with the core application but can be downloaded and installed using the Extensions manager.
+- [Extensions manager](extensions_manager): A software component of Slicer that allows browsing, installing, uninstalling extensions in the [Extensions catalog (also known as the Slicer app store)](http://slicer.kitware.com/midas3/slicerappstore) directly from the application.
+- [Extensions index](https://github.com/Slicer/ExtensionsIndex): A repository that contains description of each extension that the Extension catalog is built from.
+- Extent: Range of integer coordinates along 3 axes. Typically defined by 6 values, for IJK axes: `I_min`, `I_max`, `J_min`, `J_max`, `K_min`, `K_max`. Both minimum and maximum values are inclusive, therefore size of an array is `(I_max - I_min + 1)` x `(J_max - J_min + 1)` x `(K_max - K_min + 1)`.
+- Fiducial: Represents a point in 3D space. The terms originates from image-guided surgery, where "fiducial markers" are used to mark point positions.
```
jcfr, line 166:
```suggestion
- Fiducial: Represents a point in 3D space. The term originates from image-guided surgery, where "fiducial markers" are used to mark point positions.

```

Docs/user_guide/getting_started.md
```diff
@@ -149,3 +149,51 @@ Browse the [User Guide](user_interface) section to find quick overview of the ap
...
+- Brightness/contrast: Specifies linear mapping of voxel values to brightness of a displayed pixel. Brightness is the linear offset, contrast is the multiplier.
+- Cell: Data cells are simple topological elements like lines, polygons, tetrahedra, etc.
+- Coordinate system (or coordinate frame, reference frame, space): Specified by position of origin, axis directions, and distance unit. All coordinate systems in 3D Slicer are right-handed.
+- Extension: A collection of modules that is not bundled with the core application but can be downloaded and installed using the Extensions manager.
+- [Extensions manager](extensions_manager): A software component of Slicer that allows browsing, installing, uninstalling extensions in the [Extensions catalog (also known as the Slicer app store)](http://slicer.kitware.com/midas3/slicerappstore) directly from the application.
+- [Extensions index](https://github.com/Slicer/ExtensionsIndex): A repository that contains description of each extension that the Extension catalog is built from.
```
jcfr, line 164:
```suggestion
- [Extensions index](https://github.com/Slicer/ExtensionsIndex): A repository that contains a description of each extension that the Extension catalog is built from.
```

Docs/user_guide/getting_started.md
```diff
@@ -149,3 +149,51 @@ Browse the [User Guide](user_interface) section to find quick overview of the ap
...
+- Extent: Range of integer coordinates along 3 axes. Typically defined by 6 values, for IJK axes: `I_min`, `I_max`, `J_min`, `J_max`, `K_min`, `K_max`. Both minimum and maximum values are inclusive, therefore size of an array is `(I_max - I_min + 1)` x `(J_max - J_min + 1)` x `(K_max - K_min + 1)`.
+- Fiducial: Represents a point in 3D space. The terms originates from image-guided surgery, where "fiducial markers" are used to mark point positions.
+- Frame: One time point in a time sequence. To avoid ambiguity, this term is not used to refer to a slice of a volume.
+- Geometry: Specify location and shape of an object in 3D space.
+- Intensity: Typically refers to value of a voxel.
```
jcfr, line 169:
```suggestion
- Intensity: Typically refers to the value of a voxel.
```

Docs/user_guide/getting_started.md
```diff
@@ -149,3 +149,51 @@ Browse the [User Guide](user_interface) section to find quick overview of the ap
...
+- Labelmap (or labelmap volume, labelmap volume node): Volume node that has discrete (integer) voxel values. Typically each value corresponds to a specific structure or region. This allows compact representation of non-overlapping regions in a single 3D array. Most software use a single labelmap to store an image segmentation, but Slicer uses a dedicated segmentation node, which can contain multiple representations (multiple labelmaps to allow storing overlapping segments; closed surface representation for quick 3D visualization, etc.).
+- LPS: Left-posterior-superior anatomical coordinate system. Most commonly used coordinate system in medical image computing. Slicer stores all data in LPS coordinate system on disk (and converts to/from RAS when writing to or reading from disk).
+- Markups: Simple geometric objects and measurements that the user can place in viewers. [Markups module](modules/markups) can be used to create such objects. There are several types, such as fiducial list, line, curve, plane, ROI.
+- Master volume: Voxel values of this volume is used during segmentation by those effects that rely on intensity of an underlying volume.
```
jcfr, line 175:
```suggestion
- Master volume: Voxel values of this volume are used during segmentation by those effects that rely on intensity of an underlying volume.
 ```

File: Docs/user_guide/getting_started.md [hunk 11]
```diff
@@ -149,3 +149,51 @@ Browse the [User Guide](user_interface) section to find quick overview of the ap
...
+- [MRML (Medical Reality Markup Language)](https://en.wikipedia.org/wiki/Medical_Reality_Markup_Language): Software library for storage, visualization, and processing of information objects that may be used in medical applications. The library is designed to be reusable in various software applications, but 3D Slicer is the only major application that is known to use it.
+- Model (or model node): MRML node storing surface mesh (consists of triangle, polygon, or other 2D cells) or volumetric mesh (consists of tetrehedral, wedge, or other 3D cells)
```
jcfr, line 177:
```suggestion
- Model (or model node): MRML node storing surface mesh (consists of triangle, polygon, or other 2D cells) or volumetric mesh (consists of tetrahedral, wedge, or other 3D cells)
```

Docs/user_guide/getting_started.md
```diff
@@ -149,3 +149,51 @@ Browse the [User Guide](user_interface) section to find quick overview of the ap
...
+- Volume (or volume node, scalar volume, image): MRML node storing 3D array of voxels. Indices of the array are typically referred to as IJK. Range of IJK coordinates are called extents. Geometry of the volume is specified by its origin (position of the IJK=(0,0,0) point), spacing (size of a voxel along I, J, K axes), axis directions (direction of I, J, K axes in the reference coordinate system). 2D images are single-slice 3D volumes, with their position and orientation specified in 3D space.
+- [Voxel](https://en.wikipedia.org/wiki/Voxel): One element of a 3D volume. It has a rectangular prism shape. Coordinates of a voxel correspond to position of its center point. Value of a voxel may be a simple scalar value or a vector.
```
jcfr, line 196:
```suggestion
- [Voxel](https://en.wikipedia.org/wiki/Voxel): One element of a 3D volume. It has a rectangular prism shape. Coordinates of a voxel correspond to the position of its center point. Value of a voxel may be a simple scalar value or a vector.
```

Docs/user_guide/getting_started.md
```diff
@@ -149,3 +149,51 @@ Browse the [User Guide](user_interface) section to find quick overview of the ap
...
+- [Extensions index](https://github.com/Slicer/ExtensionsIndex): A repository that contains description of each extension that the Extension catalog is built from.
+- Extent: Range of integer coordinates along 3 axes. Typically defined by 6 values, for IJK axes: `I_min`, `I_max`, `J_min`, `J_max`, `K_min`, `K_max`. Both minimum and maximum values are inclusive, therefore size of an array is `(I_max - I_min + 1)` x `(J_max - J_min + 1)` x `(K_max - K_min + 1)`.
+- Fiducial: Represents a point in 3D space. The terms originates from image-guided surgery, where "fiducial markers" are used to mark point positions.
+- Frame: One time point in a time sequence. To avoid ambiguity, this term is not used to refer to a slice of a volume.
+- Geometry: Specify location and shape of an object in 3D space.
+- Intensity: Typically refers to value of a voxel.
+- IJK: Voxel coordinate system axes. Each integer coordinate value combination corresponds to a voxel center.
```
pieper, line 170: Change to:

- IJK: Voxel coordinate location within a 3D array.  By VTK convention, K indexes the slice, J indexes the row, and i indexes the column.  Note that numpy uses the opposite ordering convention, where `a[K][J][I]`. Sometimes this memory layout is described as I being the fastest moving index and K being the slowest moving.

Docs/user_guide/getting_started.md
pieper: Side comment: the term `master` is being deprecated in many contexts.  Any objection to using the term `reference` instead?  This would be consistent with dicom usage and is also makes sense that you can have multiple reference volumes for the same segmentation.

Docs/user_guide/getting_started.md
```diff
@@ -149,3 +149,51 @@ Browse the [User Guide](user_interface) section to find quick overview of the ap
...
+- Model (or model node): MRML node storing surface mesh (consists of triangle, polygon, or other 2D cells) or volumetric mesh (consists of tetrehedral, wedge, or other 3D cells)
+- Module: A Slicer module is a software component consisting of a graphical user interface (that is displayed in the module panel when the module is selected), a logic (that implements algorithms that operate on MRML nodes), and may provide new MRML node types, displayable managers (that are responsible for displaying those nodes in views), input/output plugins (that are responsible for load/save MRML nodes in files), and various other plugins. Modules are typically independent and only communicate with each other via modifying MRML nodes, but sometimes a module use features provided by other modules by calling methods in its logic.
+- Node (or MRML node): One data object in the scene. A node can represent data (such as an image or a mesh), describe how it is displayed (color, opacity, etc.), stored on disk, spatial transformations applied on them, etc.
```
pieper, line 179: Add:
There is a C++ class hierarchy to define the common behaviors of nodes, such as the property of being storable on disk or being geometrically transformable.   The structure of this class hierarchy can be inspected in the code or [in the API documentation](http://apidocs.slicer.org/master/classvtkMRMLStorableNode.html).

pieper (changes requested):
This is great 👍 

I'm sure we'll think of other things as people ask questions but no reason not to merge this quickly and advertise it on the forum.

Docs/user_guide/getting_started.md
lassoan: I've added the resampling and modeling transform to the updated version. There is a lot to tell about coordinate systems and transformations, so it could make sense to add a new "Transformations" section either as a new page or as a section in "Registration" page and refer to that for details.

pieper: Agreed, we don't want too much in the glossary.  The collaborative writing process is working well and I don't mind adding or reviewing blocks of material.  [This coordinate systems wiki page](https://www.slicer.org/wiki/Coordinate_systems) gets cited a lot, so maybe we incorporate that and add transform material.

Docs/user_guide/getting_started.md
lassoan: We use "reference volume" for too many things. Most importantly, reference volume is already used for segmentation: it is the volume that specifies reference geometry (that is used for determining initial segmentation geometry and exported labelmap geometry). Maybe it could be "Guiding volume"?

We also use "master representation" for the representation that all the other representations are computed from. If anyone feels strongly about it, it could be renamed to "source" representation.

I would not replace "master" with a word that would make the meaning less clear, but if we can find better words then we could change them.

@cpinter

lassoan: "Reference volume" is heavily overloaded and used for other purposes in segmentation. We could replace the word if we find better alternative. Maybe we could use "Guide volume" (and "Source representation").

lassoan:
I've made all the suggested edits.

Docs/user_guide/getting_started.md
```diff
@@ -149,3 +149,51 @@ Browse the [User Guide](user_interface) section to find quick overview of the ap
...
+- Fiducial: Represents a point in 3D space. The term originates from image-guided surgery, where "fiducial markers" are used to mark point positions.
+- Frame: One time point in a time sequence. To avoid ambiguity, this term is not used to refer to a slice of a volume.
+- Geometry: Specify location and shape of an object in 3D space. See "Volume" term for definition of image geometry.
+- Image intensity: Typically refers to the value of a voxel. Displayed pixel brightness and color is computed from this value based on the chosen window/level and color lookup table.
+- IJK: Voxel coordinate system axes. Integer coordinate values correspond to voxel center positions. IJK values are often used as coordinate location within a 3D array. By VTK convention, and I indexes the column, J indexes the row, K indexes the slice. Note that numpy uses the opposite ordering convention, where `a[K][J][I]`. Sometimes this memory layout is described as I being the fastest moving index and K being the slowest moving.
```
jcfr, line 170:
```suggestion
- IJK: Voxel coordinate system axes. Integer coordinate values correspond to voxel center positions. IJK values are often used as coordinate location within a 3D array. VTK convention uses `I` to index the columns, `J` to index the rows and `K` to index the slices. On the other hand, numpy uses the opposite ordering convention `a[K][J][I]`. Sometimes the numpy memory layout is described as `I` being the fastest moving index and `K` being the slowest moving one.
```

jcfr: It may also be worth indicating that in both cases, the  ‘row major’ or C-style convention is used to organize the data. See https://agilescientific.com/blog/2018/12/28/what-is-the-fastest-axis-of-an-array

Docs/user_guide/getting_started.md
pieper: github now creates the 'main' branch by default.  I still like 'reference' volume in the context of segmentation but I agree it's not super-specific.  'background' could be good, as could 'source' or 'input'.  We use 'input' often for non-interactive segmentations, so that could be consistent.

jcfr: >  find better alternative

Relevant read https://www.theserverside.com/opinion/Master-slave-terminology-alternatives-you-can-use-right-now

jcfr:
I added few more suggestion that could be integrated in a follow up commit

Docs/user_guide/getting_started.md
lassoan: We already use "Reference volume" for something else in segmentation (this is the volume that determines initial geometry of the segmentation and it is used when exporting the segmentation to labelmap), so we cannot use it to replace master volume.

Background volume is also used in the same context. Master volume is set as background volume by default. It could be very confusing if we started to use this for something else, too. "Source" and "Input" could be good (there is no name clash in the context of segmentation), but it would be somewhat of a regression to replace a specific term with a generic term ("input volume" is used at many places; "source volume" is not very specific either and it somewhat suggests that it is "the" source of segmentation).

Either "Guide volume" or "Input volume" sounds good to me.

@cpinter what do you think?

jcfr: > github now creates the 'main' branch by default

On a related note, I updated `slicer.org` repo to use `main` as the default branch name. See https://github.com/Slicer/slicer.org/commit/5ef131bdb8125fbe91b85e7722d35d61568ddd05

pieper: To pick up again on this side conversation, another idea could be to use the term "Evidence" volume.

This would be consistent with the dicom Current Requested Procedure Evidence Sequence as discussed [in highdicom](https://github.com/MGHComputationalPathology/highdicom/issues/62).

jamesobutler: My group uses the term "Primary" instead of "Master".

lassoan: Evidence volume: Maybe DICOM will make "evidence volume" a familiar term in a few years, but for now it sounds foreign and unnecessarily abstract to me.

Primary volume: Appointing a "primary volume" at the scene level could be good. It could also be a good name for the reference volume (i.e., the _first_ selected master volume, which determines the geometry of the segmentation). However, the input volume that is currently selected in the segment editor widget is just guiding the segmentation right now - it is not permanent, distinguished role.

ITK-Snap calls it the "main image". In MITK, it is called "Patient image". They are permanently assigned to the segmentation, so you cannot simply replace them as in Slicer.

I like the "guiding volume" term (it well matches to actual role), but if it does not sound good to you then maybe we could simply use "input volume" as @pieper suggested above - at least that is very simple and already used throughout Slicer extensively.

pieper: Weighing the options do I lean toward 'Input volume' for consistency with the rest of the app and because it is logical from the point of view of the algorithms like GrowCut.  It also doesn't imply permanence like 'primary volume' would.

I see your logic for guiding, but that has a connotation of intelligence which is wouldn't make sense here.  Perhaps 'guide' volume could work.

PR #5552, opened 2021-03-26
DOC: Bold Glossary terms to improve readability

agirault:
cc: @lassoan


pieper (approved):
Thanks!  That was my thought to when I saw the formatted page. 👍

agirault:
Credits to @sreekanthArikatla for the suggestion 🙂

PR #5553, opened 2021-03-27
BUG: Prevent duplicate paths added to environment variables

lassoan:
Path environment variables could grow over 32k when installing a few extensions.
There were some checks trying to prevent adding duplicate paths, but they had logic flaws and some checks were missed.

I've discovered this because Elastix failed to launch because the environment was huge, with lots of duplicate entries in path, pythonpath, and librarypath. Hopefully this will fix the issue (I could not reproduce the massive amount of duplication in a debug-mode local build that occurred in a release build, with installing and updating many extensions).


lassoan:
@jcfr do you have any comments fir this?

jcfr (approved):

PR #5556, opened 2021-03-29
BUG: Fix new segments appearing as hidden

### LINKED ISSUE #5533: SegmentEditor Redo does not restore segment visibility
When undoing segment creation and then redoing, the segment visibility will be off, thus the user does not actually see the action redone.

## Steps to reproduce

https://user-images.githubusercontent.com/1325980/111513282-8b560e80-8748-11eb-8fb5-b58cdd97c70c.mp4

## Expected behavior

Segment visibility should be on so that the redo action is fully successful.

## Environment
- Slicer version: Slicer-4.11.20210226

lassoan:
Segment selector widget in qMRMLSliceControllerWidget made segments hidden when the segmentation display node was updated earlier than the segmentation node.
The problem was that qMRMLSliceControllerWidget made all segments that were not in the segment selector widget hidden, and segments were not yet in the segment selector widget when they have just been added but node modified event was not yet invoked.

Fixed by not modifying visibility of those segments that are not yet in the segment selector widget.

Fixes #5533

### DIFF
--- a/Libs/MRML/Widgets/qMRMLSegmentSelectorWidget.cxx
+++ b/Libs/MRML/Widgets/qMRMLSegmentSelectorWidget.cxx
@@ -357,6 +357,21 @@ QStringList qMRMLSegmentSelectorWidget::selectedSegmentIDs()
 }
 
 //-----------------------------------------------------------------------------
+QStringList qMRMLSegmentSelectorWidget::segmentIDs()
+{
+  Q_D(qMRMLSegmentSelectorWidget);
+  QStringList allSegmentIDs;
+  // Update checkbox states in checkable combobox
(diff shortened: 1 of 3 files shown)


cpinter:
Nice investigation, and the fix looks elegant as well. I approve this now but will test it in the app I'm working with as well. Thanks a lot!

cpinter (approved):

cpinter:
It works in the custom app as well.

I'd be curious what caused this regression but the main point is that it is fixed. Thanks again! I'll merge it because I don't see any reason not to.

lassoan:
The order of EndModify call for segmentation and its display node was changed at a few places when manual Start/EndModify was replaced by ModifyBlocker. The order should not matter, it just caused problems because the logic was n the slice controller widget was flawed and if the "wrong" order was used then it modified the visibility.

PR #5563, opened 2021-04-03
BUG: Make cropping of vector volumes more intuitive

lassoan:
Crop volume requires the output volume node type to be the same as the input volume node type.
There was no option to create vector volume, therefore it was not possible to pre-create the output volume for vector volumes.
The error only appeared in the log.

Fixed by adding display of the error message in the module widget and added the option of creating vector volume output.

Addresses the problem reported by a user: https://discourse.slicer.org/t/crop-volume-reports-error-when-input-is-a-color-volume/16857


pieper (approved):
LGTM 👍

PR #5567, opened 2021-04-05
ENH: Optimize markups batch update speed

### LINKED ISSUE #5564: fcsv files are slow to load due to extra computation
See: https://discourse.slicer.org/t/read-performance-between-json-and-fcsv

There is a lot of updating of display nodes (scalar range and parametric curves) each time a point is inserted.

## Steps to reproduce

Access data linked in discussion above (3400 fiducials).  Json data loads instantly, fcsv takes a few minutes.

## Expected behavior

The load speed should be basically the same for both.

## Environment
- Slicer version: Slicer-4.22.20210226
- Operating system: mac big sur

lassoan:
Prevent updating the internal vtkPoints structure, interaction handles, and measurements from control points between StartModify/EndModify calls.
These are performed automatically in EndModify().

This makes point insertion fast, even if points are added one by one (but between StartModify/EndModify calls).

fixes #5564 (slow fcsv loading)

### DIFF
--- a/Libs/MRML/Core/vtkMRMLNode.h
+++ b/Libs/MRML/Core/vtkMRMLNode.h
@@ -473,6 +473,24 @@ public:
   /// \sa GetDisableModifiedEvent()
   vtkGetMacro(ModifiedEventPending, int);
 
+  /// Returns the number of times a custom modified event is requested
+  /// but not yet invoked on the node.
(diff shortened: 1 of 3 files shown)


Sunderlandkyl (approved):
LGTM!

PR #5568, opened 2021-04-05
ENH: Use parallel transport for more robust curve coordinate system computation

### LINKED ISSUE #5559: Instability in GetCurvePointToWorldTransformAtPointIndex when curve points are too close to each other
When the control points of a curve are too close to each other, then the X/Y directions of the transform given by GetCurvePointToWorldTransformAtPointIndex are unstable regarding small changes in the curve.

A workaround is to reduce the number of curve points (= increase the distance between the curve points) by setting a lower number of points per segment (SetNumberOfPointsPerInterpolatingSegment).

A better fix would be to improve vtkFrenetSerretFrame so that it skips points until reaching a minimum distance between one point and the next considered point.

## Steps to reproduce

https://user-images.githubusercontent.com/1325980/113170498-857b2580-923e-11eb-9a1c-9955b86f0d1d.mp4

## Expected behavior

The orientation of the clip at the end of the device does not flicker.
In other words, the X/Y axes of the transform gven by GetCurvePointToWorldTransformAtPointIndex are more stable.

## Environment
- Slicer version: Slicer-4.13.0-2021-03-21

lassoan:
Previously, "consistent" variant of Frenet-Serret method (in SplineDrivenImageSlicer VTK remote package) was used in GetCurvePointToWorldTransformAtPointIndex to get coordinate frame along the curve. However, this implementation had random initial normal direction and it was not robust when points were too close to each other.

This commit switches to a parallel transport method, which is more robust and predictable. Initial direction is set so that +X direction is preferred as normal direction (if curve tangent is exactly +X then +Y direction is preferred as binormal direction).

Fixes #5559

### DIFF
--- a/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsCurveNode.cxx
+++ b/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsCurveNode.cxx
@@ -38,7 +38,7 @@
 #include <vtkCommand.h>
 #include <vtkCutter.h>
 #include <vtkDoubleArray.h>
-#include <vtkFrenetSerretFrame.h>
+#include <vtkParallelTransportFrame.h>
 #include <vtkGeneralTransform.h>
 #include <vtkGenericCell.h>
 #include <vtkLine.h>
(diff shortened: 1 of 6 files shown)


cpinter (approved):
Looks great, thank you very much!

PR #5571, opened 2021-04-06
BUG: Fix Segment Editor Undo/Redo button states

cpinter:
Undo/Redo buttons were enabled before having done anything in the module. A new function has been added that does both types of checks that are needed to decide enabled state for these buttons, and this function is called both from updateWidgetFromMRML and onSegmentationHistoryChanged.

### DIFF
--- a/Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.cxx
+++ b/Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.cxx
@@ -1268,9 +1268,7 @@ void qMRMLSegmentEditorWidget::updateWidgetFromMRML()
   this->updateEffectsSectionFromMRML();
 
   // Undo/redo section
-
-  d->UndoButton->setEnabled(!d->Locked);
-  d->RedoButton->setEnabled(!d->Locked);
+  this->updateUndoRedoButtonsState();
 
   // Masking section
   this->updateMaskingSection();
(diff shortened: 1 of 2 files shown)


cpinter:
Linux build fails on the CI. It seems to be related to https://github.com/Slicer/Slicer/pull/5568. Doing a Linux VM build...

lassoan:
Docker image that contains prerequisites (VTK, ITK, vtkAddon, etc.) is rebuilt every night. So, if any of these libraries are changed during the day then CI build will fail until next day.

jcfr:
> So, if any of these libraries are changed during the day then CI build will fail until next day.

Currently we fail fast if any file changed in the `SuperBuild` directory, should we disable that behavior and instead try to build anyway ?

https://github.com/Slicer/Slicer/blob/b538adb1cf2ade75aaef945c788713212b46b274/.circleci/config.yml#L15-L24

lassoan:
I think the fast failure is good. If we want to improve things then we could check if the dependencies are changed (and rebuild if they have changed) a few times a day, instead of just once a day.

cpinter:
OK. My VM build succeeded as well.

How does this PR look? Thanks!

cpinter:
@lassoan can you please take a look at the change? It's just a few lines. Thank you!

lassoan (approved):
Looks good, thank you!

cpinter:
Excellent, thanks!

PR #5574, opened 2021-04-07
BUG: Fix qMRMLCollapsibleButton pallete not updating on style change

### LINKED ISSUE #5135: Some widgets not updating appropriately on style change
There are a few widgets that aren't updating when changing Slicer Style and then continuing to use the session. If I find anymore examples and this issue is still open, I will add them here.

- The qMRMLMarkupsDisplayNodeWidget in the Markups module:
image

- The qMRMLSegmentationRepresentationsListView in the Segmentations module:
image

jamesobutler:
This closes #5135.

Open Slicer and go to a module that uses a qMRMLCollapsibleButton such as the Markups module Display section. Then change the Slicer style in the settings area.

| Current | This PR |
|----------|---------|
|image|image|

### DIFF
--- a/Libs/MRML/Widgets/qMRMLCollapsibleButton.cxx
+++ b/Libs/MRML/Widgets/qMRMLCollapsibleButton.cxx
@@ -33,37 +33,10 @@ qMRMLCollapsibleButton::qMRMLCollapsibleButton(QWidget* parentWidget)
   :Superclass(parentWidget)
 {
   this->setAutoFillBackground(true);
-  this->computePalette();
 }
 
 // --------------------------------------------------------------------------
 void qMRMLCollapsibleButton::changeEvent(QEvent* event)
 {
-  if (event->type() == QEvent::ParentChange)
-    {
-    this->computePalette();
-    }
   this->Superclass::changeEvent(event);
-}
-
-void qMRMLCollapsibleButton::computePalette()
-{
-  QColor backgroundColor =
-    this->style()->standardPalette().color(QPalette::Window);
-  QObject* ancestor = this;
-  while(ancestor->parent())
-    {
-    ancestor = ancestor->parent();
-    if (qobject_cast<qMRMLCollapsibleButton*>(ancestor)||
-        qobject_cast<ctkCollapsibleButton*>(ancestor))
-      {
-      backgroundColor = backgroundColor.darker(108);
-      backgroundColor = QColor::fromHsvF(backgroundColor.hueF(),
-                                         backgroundColor.saturationF()*1.2,
-                                         backgroundColor.valueF());
-      }
-    }
-  QPalette newPalette = this->palette();
-  newPalette.setColor(QPalette::Window, backgroundColor);
-  this->setPalette(newPalette);
 }
--- a/Libs/MRML/Widgets/qMRMLCollapsibleButton.h
+++ b/Libs/MRML/Widgets/qMRMLCollapsibleButton.h
@@ -39,7 +39,6 @@ public:
   ~qMRMLCollapsibleButton() override = default;
 protected:
   void changeEvent(QEvent* event) override;
-  void computePalette();
 };
 
 #endif


jamesobutler:
re https://github.com/Slicer/Slicer/issues/5135#issuecomment-799773831. 

> It appears that ctkCollapsibleButton is used more often than qMRMLCollapsibleButton. 

> It appears qMRMLCollapsibleButton.cxx hasn't been modified in over 9 years. Based on what little the class is doing does it make sense to even keep qMRMLCollapsibleButton around? I'm not sure which MRML logic was intended to be used in this widget.

I'm currently maintaining qMRMLCollapsibleButton, but removing the custom `computePalette` method.

jcfr:
> I'm currently maintaining qMRMLCollapsibleButton, but removing the custom computePalette method.

Makes sense, this will help  customizing application using style-sheet.

What is the different with/without the `computePalette` function when using the regular "light" style ?

jamesobutler:
If you start Slicer with "Dark Slicer" and change to "Light Slicer"

| Current | This PR |
|----------|---------|
|image|image|

jcfr:
> If you start Slicer with "Dark Slicer" and change to "Light Slicer"

To clarify, I was wondering what was the difference with/without your patch when using only "Light Slicer". The idea is to document the effect introduced by the removal of "computePalette" ?

jamesobutler:
Not sure exactly what you are asking. If you are only ever using 1 style such as starting Slicer with "Dark Slicer" or starting Slicer with "Light Slicer" you would never see the issue that is being solved here. It only happened when switching Styles. 

If toggling any combination of "Light Slicer" with a different style, this PR makes sure the style is appropriate and not inverted as in the previous posted images of the "Current" state.

jamesobutler:
If there is a certain set of steps that you would like me to test with and without the changes in this PR, let me know and I can provide the results.

jcfr:
Adding this seems to improve things:

```diff
diff --git a/Libs/MRML/Widgets/qMRMLCollapsibleButton.cxx b/Libs/MRML/Widgets/qMRMLCollapsibleButton.cxx
index 810adfa677..70d706e6af 100644
--- a/Libs/MRML/Widgets/qMRMLCollapsibleButton.cxx
+++ b/Libs/MRML/Widgets/qMRMLCollapsibleButton.cxx
@@ -39,7 +39,7 @@ qMRMLCollapsibleButton::qMRMLCollapsibleButton(QWidget* parentWidget)
 // --------------------------------------------------------------------------
 void qMRMLCollapsibleButton::changeEvent(QEvent* event)
 {
-  if (event->type() == QEvent::ParentChange)
+  if (event->type() == QEvent::ParentChange || event->type() == QEvent::StyleChange)
     {
     this->computePalette();
     }
```

That said, I suspect similar change would have to be implemented in few other widgets

jamesobutler:
It would seem to be desired that all uses of a collapsible button would have a similar style. There are currently only 2 modules (Markups and Texts) that are using qMRMLCollapsibleButtons while many more modules are using ctkCollapsibleButtons.

Therefore should the correct action be to make the collapsible button styling usage across Slicer to that of ctkCollapsibleButton or to qMRMLCollapsibleButtons with the styled darker area? It seems confusing to have inconsistent styling of the two. 

image

jcfr:
> It seems confusing to have inconsistent styling of the two.

Agreed, both should indeed be consistent, this means we should probably move the code to ctk and document the behavior.

The issue is that directly setting the palette is not compatible with setting `styleSheet` (see experiment below), if we would like to have both, I think we should update code in Slicer and CTK to skip the palette update if an application  attribute is set.

To allow this, we could add the attribute `AA_DisableWidgetCustomPalette` to this enum:

https://github.com/Slicer/Slicer/blob/f17e701bf19d678c61d6fd4e2d904bfa3fee845d/Base/QTCore/qSlicerCoreApplication.h#L116-L120

value of the custom attribute would be initialized like this:

```
AA_DisableWidgetCustomPalette = ctk::AA_DisableWidgetCustomPalette
```

where `AA_DisableWidgetCustomPalette` would be added to [Widgets/ctkWidgetsUtils.h](https://github.com/commontk/CTK/blob/master/Libs/Widgets/ctkWidgetsUtils.h)

Then, code currently calling `setPalette` would be updated to skip the palette update if the attribute `AA_DisableWidgetCustomPalette` is set.


### Experiment: setting stylesheet along use of computePalette

Trying to use the snippet above (see https://github.com/Slicer/Slicer/pull/5574#issuecomment-815873273) along with:

```python
collapsible1.styleSheet = "background-color: red"
```

does not allow to set the background using stylesheet.

## Occurrences of setPalette in Slicer and CTK

Slicer:

```
ack setPalette
Libs/MRML/Widgets/qMRMLViewControllerBar.cxx
183:  this->BarWidget->setPalette(palette);
186:  this->ViewLabel->setPalette(labelPalette);

Libs/MRML/Widgets/qMRMLVolumeWidget.cxx
89:  this->PopupWidget->setPalette(popupPalette);

Libs/MRML/Widgets/qMRMLSliceControllerWidget.cxx
130:  this->SliceOffsetSlider->spinBox()->setPalette(spinBoxPalette);

Libs/MRML/Widgets/qMRMLCollapsibleButton.cxx
68:  this->setPalette(newPalette);

Base/QTGUI/qSlicerSettingsCachePanel.cxx
161:  d->UsedCacheSizeLabel->setPalette(palette);
162:  d->FreeCacheSizeLabel->setPalette(palette);

Base/QTGUI/DesignerPlugins/qSlicerWidgetPlugin.cxx
41:  widget->setPalette(style.standardPalette());

Base/QTGUI/qSlicerSettingsStylesPanel.cxx
316:  app->setPalette(newStyle->standardPalette());

Base/QTGUI/qSlicerModulePanel.cxx
340:  this->HelpLabel->setPalette(p);

```

CTK:

```
ack setPalette
Libs/Widgets/ctkConsole.cpp
1163:  d->setPalette(pal);

Libs/Widgets/Testing/Cpp/ctkPopupWidgetEventTranslatorPlayerTest1.cpp
193:  focusPopup->setPalette(palette);

Libs/Widgets/Testing/Cpp/ctkDynamicSpacerTest1.cpp
81:  spacer1->setPalette(QPalette(Qt::red));
82:  spacer2->setPalette(QPalette(Qt::blue));

Libs/Widgets/Testing/Cpp/ctkSearchBoxTest1.cpp
46:  search.setPalette(p);

Libs/Widgets/Testing/Cpp/ctkPopupWidgetTest1.cpp
104:  focusPopup->setPalette(palette);
185:    popup->setPalette(p);

Libs/Widgets/Testing/Cpp/ctkDynamicSpacerTest2.cpp
61:  spacer->setPalette(QPalette(Qt::red));

Libs/Widgets/ctkCrosshairLabel.cpp
291:    this->setPalette(palette);

Libs/DICOM/Widgets/ctkDICOMTableView.cpp
147:  this->leSearchBox->setPalette(palette);

Libs/DICOM/Widgets/ctkDICOMTableManager.cpp
180:  searchBox->setPalette(palette);

Libs/DICOM/Widgets/ctkDICOMObjectListWidget.cpp
392:  d->metadataSearchBox->setPalette(palette);

Libs/Visualization/VTK/Widgets/ctkVTKScalarsToColorsWidget.cpp
162:      this->XSpinBox->setPalette(wrongPalette);
175:      this->XSpinBox->setPalette(wrongPalette);
643:  d->XSpinBox->setPalette(this->palette());

```

cc: @lassoan

lassoan:
About the specific widget: it would help if collapsible section background color would slightly differ from the regular background (slightly darker/lighter for each child level).

In general: Since stylesheets can more properties than just colors, it makes a lot of sense to support them better. The main problem is how to support custom colors in custom widgets. I think you cannot specify custom palette colors, that's why they are hardcoded (or recently semi-hardcoded - we check if the palette is dark or bright and choose between two hardcoded colors). It seems that you can specify custom colors in qss (see https://wiki.qt.io/Dynamic_Properties_and_Stylesheets). Could we check in the widget if qss specifies a custom color and if it does then use that and otherwise use the semi-hardcoded color based on the palette? This way we would not need to introduce a custom application attribute.

pieper:
To the extent qss is like css then there are powerful ways to have widgets with special classes and also give rules for widgets classes dependent on the classes of whatever parent they may be under.  As long as we are careful with our conventions we should be able to support a wide range of customization options.

lassoan:
I've merged this (https://github.com/Slicer/Slicer/commit/0daa3b92680706c1e98b6585a676545f41c0b8a2), as the issue is too serious to be delayed while thinking about a perfect solution. We can revisit this by improving or removing the widget later.

PR #5575, opened 2021-04-08
BUG: Fix closeTemporaryDatabase to re-open original database only if it exists

jcfr:
This commit avoids reporting error on system where no original database
was created.


cpinter (approved):
LGTM

PR #5577, opened 2021-04-08
ENH: Improve DICOM examine and load performance

lassoan:
If only first file of a series is needed then retrieve only that.

Update git hashes of MultiVolumeImporter and CTK to include related API and performance improvements.


pieper (approved):

pieper:
Nice 👍

PR #5581, opened 2021-04-12
BUG: Fixed label positions in zoomed in view

lassoan:
When zooming in on a very small object, the axis labels of the purple box were drifting very far away.

How to reproduce: load MRHead, set volume spacing to 1000x smaller (`getNode('MRHead').SetSpacing(0.001, 0.001, 0.0013`), fit to slice views, center 3D view, hit "R" key in 3D view to zoom to fit.

The fix was to center the text in the text actor's coordinate system.

Before the fix:

image

After the fix:

image


pieper (approved):
👍

PR #5582, opened 2021-04-12
ENH: Support taking screenshot without displaying message

jamesobutler:
This change allows `takeScreenshot` to be called without displaying a `delayDisplay` message. I'm using it for non-testing purposes so I don't need a `delayDisplay`.

### DIFF
--- a/Base/Python/slicer/ScriptedLoadableModule.py
+++ b/Base/Python/slicer/ScriptedLoadableModule.py
@@ -301,12 +301,12 @@ class ScriptedLoadableModuleLogic(object):
     """
     slicer.util.clickAndDrag(widget,button=button,start=start,end=end,steps=steps,modifiers=modifiers)
 
-  def takeScreenshot(self,name,description,type=-1):
+  def takeScreenshot(self,name,description,type=-1,displayDescription=True):
     """ Take a screenshot of the selected viewport and store as and
     annotation snapshot node. Convenience method for automated testing.
 
-    If self.enableScreenshots is False then only a message is displayed but screenshot
-    is not stored. Screenshots are scaled by self.screenshotScaleFactor.
+    If self.enableScreenshots is False a screenshot is not stored.
+    Screenshots are scaled by self.screenshotScaleFactor.
 
     :param name: snapshot node name
     :param description: description of the node
@@ -314,10 +314,11 @@ class ScriptedLoadableModuleLogic(object):
       Valid values: slicer.qMRMLScreenShotDialog.FullLayout,
       slicer.qMRMLScreenShotDialog.ThreeD, slicer.qMRMLScreenShotDialog.Red,
       slicer.qMRMLScreenShotDialog.Yellow, slicer.qMRMLScreenShotDialog.Green.
+    :param displayDiscription: whether to show description in a delayDisplay
     """
-
-    # show the message even if not taking a screen shot
-    slicer.util.delayDisplay(description)
+    if displayDescription:
+      # Displaying a description is allowed even if not taking a screenshot
+      slicer.util.delayDisplay(description)
 
     if not self.enableScreenshots:
       return


lassoan:
This function is intended for testing purposes. It cannot be even used outside (e.g., from the Python console or Jupyter notebooks, without instantiating or accessing a module logic class).

Instead of improving this, we should use ScreenCapture module (see few examples [here](https://www.slicer.org/wiki/Documentation/Nightly/ScriptRepository#Capture)). If ScreenCapture is not good enough then improve it (or add convenience functions to slicer.util).

If you use this function because of annotation screenshots then we need to figure out some other solution because we plan to retire Annotations module. We have now Line and ROI in markups module, we'll add report generation in the future, and we don't plan to add screenshot support to markups, so we will probably deprecate Annotations module soon and remove in 1-2 years.

jamesobutler:
I'm currently using in a custom module's ScriptedLoadableModuleLogic class. We use it to create a vtkMRMLAnnotationSnapshotNode for a segmentation whenever it gets edited/applied. Then upon our custom save events we save the segmentation and the latest snapshot node to a simple image format (png). We have built a report generator that uses the screenshots and other saved statistics tables. The generated PDF report includes the screenshots and stats in addition to generated plot views to then detail things like tumor growth over various timepoints during a study.

The ScreenCapture module doesn't currently have a method of storing a captured snapshot as it appears to always want to save it to disk immediately. Has there been thoughts of storing screen capture images in a mrml scene node like annotations does?

jamesobutler:
It appears that the vtkMRMLAnnotationSnapshot node is similar to the vtkMRMLSceneViewNode. Does vtkMRMLSceneViewNode have a similar deprecated fate? Or should I switch to using the SceneView module? Or maybe there is some other node type for storing image data captured from the layout window?

lassoan:
Saving screenshots to the scene (and not directly to file) was too complicated and inconvenient for most users for just taking screenshots, that's why no replacement is planned right now. Report generation is useful and it is very easy to implement it in an application specific way (generate html and print to pdf), but it is not that easy to create a generic report generator framework (define generic template and mapping to MRML node contents).

We'll keep Scene Views and will modernize it by replacing the storage backend with sequences. It will simplify things and will make Scene Views much more powerful and compatible with scene-level undo/redo, etc.

You can already update your implementation to be future-proof and start collecting screenshots into a sequence node. You don't even have to add sequence browser node and proxy node, so the volumes will not clutter the scene. You could add the option to [`captureImageFromView`](https://github.com/Slicer/Slicer/blob/98ab60ad0ee0e341dbceb7860e238cbb1f4dac1a/Modules/Scripted/ScreenCapture/ScreenCapture.py#L974) to write the captured image to a volume node instead of a file. For example, if you pass an optional `volumeNode` argument then the image would be written to that volume (and if you set filename=None then the image would not be written to file).

Thanks for your patience and sorry for not just integrating your pull request but trying to push for making things better. I think the extra effort of switching to mainstream modules will pay off mid/long term. We'll create a roadmap document somewhere to tell well ahead about all planned deprecations in Slicer. We just need to figure out what's the best place for it.

jamesobutler:
@lassoan In addition to the ScreenCapture module updates, you mentioned that this `testScreenshot` method is used for testing purposes, but it is in the ScriptedLoadableModuleLogic class instead of the ScriptedLoadableModuleTest class. Should it be moved to ScriptedLoadableModuleTest to make it more clear and restrict its usage for testing only?

Also, it appears in b55525344bcd255940d9da4b623555284e891bdc from 2014, you deprecated the `delayDisplay` method in ScriptedLoadableModuleLogic in favor of the `delayDisplay` in slicer.util. Can this now be removed? It's also a testing related method in ScriptedLoadableModuleLogic which really shouldn't be there.
https://github.com/Slicer/Slicer/blob/243f1b16f9c86e389686cc6596017515faac90a3/Base/Python/slicer/ScriptedLoadableModule.py#L288-L294

There is also a `delayDisplay` in ScriptedLoadableModuleTest which appears to have a `requestedDelay` argument that was "deprecated" in favor of the preferred `msec` argument. This was changed in 1f4c8d23648329bf2e3258c0a0792b35d79014cf from Dec 2018. Can `requestedDelay` now be removed to simplify that method?

jamesobutler:
Closing this PR since the testing method `takeScreenshot` is now in the ScriptedLoadableModuleTest class as of 71f236bfacffec1d6c01036840dfc8d89be0563c. Therefore using it for non-test purposes isn’t supported.

lassoan:
The option of taking screenshot into a volume node can be useful in general, so adding it to ScreenCapture module is a good idea. I'm just cleaning this up and push it.

lassoan:
It is merged now: https://github.com/Slicer/Slicer/commit/5afa6ae90127f7a4eccd6b67f7cc37b69a0a2007

PR #5583, opened 2021-04-12
STYLE: Remove unnecessary trailing semicolon in python code

jamesobutler:
I noticed an instance of a trailing semicolon being used in some python code. I used the following script to find other similar uses and then removed the trailing semicolon after manually reviewing each instance. Some were part of some embedded logic that wasn't python getting passed elsewhere or were some C++ examples in comments of a python file.
```python
import os
search_directory = os.path.join(os.getenv('UserProfile'), "Documents", "GitHub", "Slicer")
for root, dirs, files in os.walk(search_directory):
    for filename in files:
        if filename.endswith(".py"):
            filepath = os.path.join(root, filename)
            with open(filepath, "r") as open_file:
                lines = open_file.readlines()
                for index, line in enumerate(lines):
                    if line.endswith(";\n"):
                        print(filepath)
                        print(f"LINE: {index+1}:" + line)
```

### DIFF
--- a/Applications/SlicerApp/Testing/Python/CLIEventTest.py
+++ b/Applications/SlicerApp/Testing/Python/CLIEventTest.py
@@ -105,7 +105,7 @@ class CLIEventTestTest(ScriptedLoadableModuleTest):
   def _testCLIStatusEventTest(self, wait_for_completion):
     self.delayDisplay('Testing status events for a normal execution of a CLI')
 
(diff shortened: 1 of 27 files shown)


lassoan (approved):
Great, thank you!

jcfr:
Thanks

Could you add the script to https://www.slicer.org/wiki/Documentation/Nightly/ScriptRepository ?

PR #5586, opened 2021-04-13
Screen capture to node

jamesobutler:
This is a follow up to requests made in https://github.com/Slicer/Slicer/pull/5582#issuecomment-818265101. 

With the changes a user could capture a single image to a vtkMRMLVectorVolumeNode to then save at a later time. This would be used to replace the functionality of the Annotation's module screenshot functionality with storage in a vtkMRMLAnnotationSnapshotNode. vtkMRMLVectorVolumeNode was chosen so color images could be supported.

Specifically using the Screen Capture GUI I can set "Capture all views", set number of images to Single, clear out the output directory to not save to disk directly and then choose a vtkMRMLVectorVolumeNode from a node combobox, and also select View Controllers in the advanced section to keep the slice controllers in the output. 
image

Upon saving the vtkMRMLVectorVolumeNode as a PNG file, the output is not exactly correct as the coordinate system is off as things are flipped. @lassoan Can you provide guidance about what is the syntax for fixing this issue? Do you manipulate the vtkImageData prior to setting it to be observed by the vtkMRMLVectorVolumeNode?
VectorVolume

Also note that having looked at the ScreenCapture code it could definitely use a cleanup organization as most functionality appears just added on top of other things and so it is a bit difficult to understand in the GUI which options are relevant to the type of capture I'm trying to do. For this PR I just fit code into the current style instead of trying to refactor the module.

### DIFF
--- a/Modules/Scripted/ScreenCapture/ScreenCapture.py
+++ b/Modules/Scripted/ScreenCapture/ScreenCapture.py
@@ -316,6 +316,14 @@ class ScreenCaptureWidget(ScriptedLoadableModuleWidget):
       "Maximum number of images to be captured (without backward steps and repeating).")
     advancedFormLayout.addRow("Maximum number of images:", self.maxFramesWidget)
 
+    self.volumeNodeComboBox = slicer.qMRMLNodeComboBox()
+    self.volumeNodeComboBox.nodeTypes = ["vtkMRMLVectorVolumeNode"]
+    self.volumeNodeComboBox.renameEnabled = True
+    self.volumeNodeComboBox.noneEnabled = True
+    self.volumeNodeComboBox.setToolTip("Select a volume node to store the image capture which allows the image to be saved at a later time.")
+    self.volumeNodeComboBox.setMRMLScene(slicer.mrmlScene)
+    advancedFormLayout.addRow("Output volume node:", self.volumeNodeComboBox)
+
     self.transparentBackgroundCheckBox = qt.QCheckBox(" ")
     self.transparentBackgroundCheckBox.checked = False
     self.transparentBackgroundCheckBox.setToolTip("If checked, images will be captured with transparent background.")
@@ -423,7 +431,7 @@ class ScreenCaptureWidget(ScriptedLoadableModuleWidget):
     self.sequenceEndItemIndexWidget.connect('valueChanged(double)', self.setSequenceItemIndex)
     self.outputTypeWidget.connect('currentIndexChanged(int)', self.updateOutputType)
     self.videoFormatWidget.connect("currentIndexChanged(int)", self.updateVideoFormat)
-    self.singleStepButton.connect('toggled(bool)', self.numberOfStepsSliderWidget, 'setDisabled(bool)')
+    self.singleStepButton.connect('toggled(bool)', self.updateSingleStepOptions)
     self.maxFramesWidget.connect('valueChanged(int)', self.maxFramesChanged)
     self.videoLengthSliderWidget.connect('valueChanged(double)', self.setVideoLength)
     self.videoFrameRateSliderWidget.connect('valueChanged(double)', self.setVideoFrameRate)
@@ -436,6 +444,7 @@ class ScreenCaptureWidget(ScriptedLoadableModuleWidget):
     self.updateOutputType()
     self.updateVideoFormat(0)
     self.updateViewOptions()
+    self.updateSingleStepOptions(self.singleStepButton.checked)
 
   def maxFramesChanged(self):
     self.numberOfStepsSliderWidget.maximum = self.maxFramesWidget.value
@@ -448,6 +457,10 @@ class ScreenCaptureWidget(ScriptedLoadableModuleWidget):
     if not self.createdOutputFile:
       return
     qt.QDesktopServices().openUrl(qt.QUrl("file:///"+self.createdOutputFile, qt.QUrl.TolerantMode))
+
+  def updateSingleStepOptions(self, checkedState):
+    self.numberOfStepsSliderWidget.setDisabled(checkedState)
+    self.volumeNodeComboBox.setEnabled(checkedState)
 
   def updateOutputType(self, selectionIndex=0):
     isVideo = self.outputTypeWidget.currentText == "video"
@@ -677,10 +690,15 @@ class ScreenCaptureWidget(ScriptedLoadableModuleWidget):
       if numberOfSteps < 2:
         if imageFileNamePattern != self.snapshotFileNamePattern or outputDir != self.snapshotOutputDir:
           self.snapshotIndex = 0
-        [filename, self.snapshotIndex] = self.logic.getNextAvailableFileName(outputDir, imageFileNamePattern, self.snapshotIndex)
+        if outputDir:
+          [filename, self.snapshotIndex] = self.logic.getNextAvailableFileName(outputDir, imageFileNamePattern, self.snapshotIndex)
+        else:
+          filename = None
         view = None if captureAllViews else self.logic.viewFromNode(viewNode)
-        self.logic.captureImageFromView(view, filename, transparentBackground)
-        self.logic.addLog("Write "+filename)
+        volumeNode = None if numberOfSteps>1 else self.volumeNodeComboBox.currentNode()
+        self.logic.captureImageFromView(view, filename, transparentBackground, volumeNode=volumeNode)
+        if filename:
+          self.logic.addLog("Write "+filename)
       elif self.animationModeWidget.currentText == "slice sweep":
         self.logic.captureSliceSweep(viewNode, self.sliceStartOffsetSliderWidget.value,
           self.sliceEndOffsetSliderWidget.value, numberOfSteps, outputDir, imageFileNamePattern,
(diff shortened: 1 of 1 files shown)


Modules/Scripted/ScreenCapture/ScreenCapture.py
```diff
@@ -1050,10 +1068,13 @@ def captureImageFromView(self, view, filename, transparentBackground=False):
       imageClipper.Update()
       capturedImage = imageClipper.GetOutput()
 
-    writer = self.createImageWriter(filename)
-    writer.SetInputData(self.addWatermark(capturedImage))
-    writer.SetFileName(filename)
-    writer.Write()
+    if isinstance(volumeNode, slicer.vtkMRMLVectorVolumeNode):
```
lassoan, line 1071: If volume node type is not correct (e.g., labelmap volume) then raise an exception.

Modules/Scripted/ScreenCapture/ScreenCapture.py
```diff
@@ -971,7 +989,7 @@ def getSliceOffsetResolution(self, sliceNode):
 
     return sliceOffsetResolution
 
-  def captureImageFromView(self, view, filename, transparentBackground=False):
+  def captureImageFromView(self, view, filename=None, transparentBackground=False, volumeNode=None):
```
lassoan, line 992: Things are getting complex. Please add method documentation.

Modules/Scripted/ScreenCapture/ScreenCapture.py
```diff
@@ -1050,10 +1068,13 @@ def captureImageFromView(self, view, filename, transparentBackground=False):
...
 
-    writer = self.createImageWriter(filename)
-    writer.SetInputData(self.addWatermark(capturedImage))
-    writer.SetFileName(filename)
-    writer.Write()
+    if isinstance(volumeNode, slicer.vtkMRMLVectorVolumeNode):
+      volumeNode.SetAndObserveImageData(capturedImage)
+    if filename:
+      writer = self.createImageWriter(filename)
+      writer.SetInputData(self.addWatermark(capturedImage))
```
lassoan, line 1075: Move `capturedImage = self.addWatermark(capturedImage)` before starting to write to volume node and file, so that the same content is written to both outputs.

Modules/Scripted/ScreenCapture/ScreenCapture.py
```diff
@@ -220,6 +220,12 @@ def setup(self):
     hbox.addWidget(self.lightboxImageFileNameWidget)
     outputFormLayout.addRow("Output file name:", hbox)
 
+    self.volumeNodeComboBox = slicer.qMRMLNodeComboBox()
```
lassoan, line 223: Enable None and Rename options.
Move it to advanced section (in most cases, users would want to write to file and prefer not store in the scene).
Add tooltip that explains what this node is used for.
Only enable this node selector if "single" option is enabled.

Modules/Scripted/ScreenCapture/ScreenCapture.py
```diff
@@ -671,16 +682,23 @@ def onCaptureButton(self):
...
           self.snapshotIndex = 0
-        [filename, self.snapshotIndex] = self.logic.getNextAvailableFileName(outputDir, imageFileNamePattern, self.snapshotIndex)
+        if outputDir:
+          [filename, self.snapshotIndex] = self.logic.getNextAvailableFileName(outputDir, imageFileNamePattern, self.snapshotIndex)
+        else:
+          filename = None
         view = None if captureAllViews else self.logic.viewFromNode(viewNode)
-        self.logic.captureImageFromView(view, filename, transparentBackground)
-        self.logic.addLog("Write "+filename)
+        self.logic.captureImageFromView(view, filename, transparentBackground, volumeNode=self.volumeNodeComboBox.currentNode())
```
lassoan, line 699: Use volumeNode=None if numberOfSteps>1 (otherwise we would keep writing images into the same node). Or, implement writing the output into a sequence node (you could create a sequence node and browser node and use the chosen volume node as proxy node).

lassoan (approved):
This will be useful, thank you! See comments inline.

lassoan:
> Can you provide guidance about what is the syntax for fixing this issue? Do you manipulate the vtkImageData prior to setting it to be observed by the vtkMRMLVectorVolumeNode?

All images are defined in 3D space. You can set the IJKToRAS matrix of the volume so that it shows up with the desired orientation in a slice view. Usually we set IJKToRAS to diag(-1,-1,1,1) by default for 2D images.

Modules/Scripted/ScreenCapture/ScreenCapture.py
```diff
@@ -1050,10 +1083,20 @@ def captureImageFromView(self, view, filename, transparentBackground=False):
...
-    writer = self.createImageWriter(filename)
-    writer.SetInputData(self.addWatermark(capturedImage))
-    writer.SetFileName(filename)
-    writer.Write()
+    capturedImage = self.addWatermark(capturedImage)
+    if isinstance(volumeNode, slicer.vtkMRMLVectorVolumeNode):
+      ijkToRas = vtk.vtkMatrix4x4()
+      ijkToRas.SetElement(0, 0, -1)
+      ijkToRas.SetElement(1, 1, -1)
+      volumeNode.SetIJKToRASMatrix(ijkToRas)
```
jamesobutler, line 1091: @lassoan This still isn't changing the results of the output for some reason. I'm still getting that flipped image. Am I doing some wrong here?

lassoan: This should work well. It should make the image appear correctly in the red slice view if you use Slicer default view orientations. Do you see that ijkToRas matrix makes a difference?

Modules/Scripted/ScreenCapture/ScreenCapture.py
```diff
@@ -1050,10 +1083,20 @@ def captureImageFromView(self, view, filename, transparentBackground=False):
...
+      volumeNode.SetAndObserveImageData(capturedImage)
+    else:
+      raise ValueError("Invalid vector volume node.")
```
lassoan, line 1094: First check if volumeNode is not None, then check the volume type and raise an exception if it is not correct.

Modules/Scripted/ScreenCapture/ScreenCapture.py
```diff
@@ -436,6 +449,7 @@ def setup(self):
     self.updateOutputType()
     self.updateVideoFormat(0)
     self.updateViewOptions()
+    self.updateSingleStepOptions(self.singleStepButton.checked)
```
lassoan, line 452: It would be nicer to enable volume node output also when the number of output images is set to 1 (without clicking "single" option). You can do all these checks in updateOutputType, where all kinds of checks and GUI states updates are done already (no need for a new callback function).

lassoan (commented):
Thank you, this is almost perfect, just added a few more small comments. It would be also nice if you added a test case for volume node export, just to make sure the feature keeps working well.

Modules/Scripted/ScreenCapture/ScreenCapture.py
jamesobutler: The number of output images is not defined as "Single" by any other interaction currently. If the `numberOfStepsSliderWidget` value becomes 1 it does not set the "Single" option checked. Are you suggesting this should be the case? Therefore the `setNumberOfSteps` slot would call to check `singleStepButton`.

Changes to `updateOutputType` would not make sense as currently I'm not adding a new option to this combobox. It still has the options "Image Series", "Video" and "lightbox image". I don't think "Video" or "lightbox image" selection would ever enable volume node output qMRMLNodeComboBox.

Modules/Scripted/ScreenCapture/ScreenCapture.py
jamesobutler: I have set the ijkToRas matrix and you can see below that it is returning the (-1,-1,1,1) diagonal, but the output still looks flipped. See image below
image

Modules/Scripted/ScreenCapture/ScreenCapture.py
lassoan: "Single" button is a useful shortcut for temporarily changing the number of request output images to 1. It has the same effect as moving "Number of images" slider to 1. We just had to add this button because many users did not realize they can create single screenshots by moving the slider to 1.

I don't mind if you keep the separate callback function. It is just a matter of balancing to avoid both too large callback functions and too many callback functions (both decrease readability of the code).

Modules/Scripted/ScreenCapture/ScreenCapture.py
lassoan: Do you see that changing ijkToRas matrix makes a difference? (if you set diag(1,1,1,1) the image is flipped compared to diag (-1,1,1,1))

Can you take a screenshot with "Axis" orientation marker displayed?

Modules/Scripted/ScreenCapture/ScreenCapture.py
jamesobutler: I guess I just don't know how to handle its initial state appropriately where it also makes sense. `setNumberOfSteps` is not called at startup and it doesn't make sense to include it any of the current slots (`updateOutputType`/`updateViewOptions`).

jamesobutler: If you don't want a slot I could do the following in the setup
```python
self.volumeNodeComboBox.enabled = False
.
.
.
self.singleStepButton.connect('toggled(bool)', self.numberOfStepsSliderWidget, 'setDisabled(bool)')
self.singleStepButton.connect('toggled(bool)', self.volumeNodeComboBox, 'setEnabled(bool)')
```

Modules/Scripted/ScreenCapture/ScreenCapture.py
jamesobutler: Ok I have identified what has made things confusing. Not sure what to do to make it both valid when viewing in slice views and upon save to PNG. 

With no SetIJKToRAS calls in the capture image method:
- The VectorVolume as viewed in the slice views needs a (-1,1,1,1) IJKToRAS matrix to flip it to the expected view
image
- However when saving the node to PNG using the Slicer save dialog, it actually needs a (1,-1,1,1) IJKToRAS matrix to flip it to the expected view.
VectorVolume

jamesobutler:
I've added a simple test case to confirm that a new vector volume node is updated and has image data after using the new output to volume node functionality added in this PR.

lassoan:
Thank you, this is merged now: https://github.com/Slicer/Slicer/commit/5afa6ae90127f7a4eccd6b67f7cc37b69a0a2007

I was somewhat confused by the overlapping https://github.com/Slicer/Slicer/pull/5582 merge request. I think valuable part of these two pull requests are now merged.

PR #5591, opened 2021-04-14
ENH: Move testing methods in scripted logic class to test class

jamesobutler:
As I found out in https://github.com/Slicer/Slicer/pull/5582#issuecomment-818070072 there are some methods in ScriptedLoadableModuleLogic class that is intended for testing purposes only. Therefore the work (originally mentioned in https://github.com/Slicer/Slicer/pull/5582#issuecomment-818758575) in this PR moves these methods to the ScriptedLoadableModuleTest class and update a few tests that were using the testing methods from the logic class. 

Part of this is enforcing the removal of some deprecated methods in the ScriptedLoadableModuleLogic class. `delayDisplay` usage in ScriptedLoadableModuleLogic has been deprecated since Dec 10th 2014 and `clickAndDrag` usage since Dec 6th 2016.

### DIFF
--- a/Applications/SlicerApp/Testing/Python/FiducialLayoutSwitchBug1914.py
+++ b/Applications/SlicerApp/Testing/Python/FiducialLayoutSwitchBug1914.py
@@ -102,10 +102,6 @@ class FiducialLayoutSwitchBug1914Logic(ScriptedLoadableModuleLogic):
   """
   def __init__(self):
     ScriptedLoadableModuleLogic.__init__(self)
-    # test difference in display location and then in RAS if this is too fine
-    self.maximumDisplayDifference = 1.0
-    # for future testing: take into account the volume voxel size
-    self.maximumRASDifference = 1.0
 
   def getFiducialSliceDisplayableManagerHelper(self,sliceName='Red'):
     sliceWidget = slicer.app.layoutManager().sliceWidget(sliceName)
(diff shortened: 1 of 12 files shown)


Applications/SlicerApp/Testing/Python/FiducialLayoutSwitchBug1914.py
```diff
@@ -203,12 +219,14 @@ def run(self,enableScreenshots=0,screenshotScaleFactor=1):
...
       if rasDiff != 0.0:
         rasDiff = math.sqrt(rasDiff)
-      print('Checking the difference between fiducial RAS position',seedRAS,'and volume RAS as derived from the fiducial display position',volumeRAS,': ',rasDiff)
+      print('Checking the difference between fiducial RAS position',seedRAS,
+            'and volume RAS as derived from the fiducial display position',volumeRAS,': ',rasDiff)
       if rasDiff > self.maximumRASDifference:
-        self.delayDisplay("RAS coordinate difference is too large as well!\nExpected < %g but got %g" % (self.maximumRASDifference, rasDiff))
-        return False
+        slicer.util.delayDisplay("RAS coordinate difference is too large as well!\nExpected < %g but got %g" % (self.maximumRASDifference, rasDiff))
+        self.delayDisplay('Test failed!')
```
lassoan, line 226: Why do we still use self.delayDisplay?
Instead of moving an assert later down (that repeats the rasDiff>self.maximumRASDifference test) it would be simpler to raise an exception (the exception makes the exception description visible in the output and makes the test fail).

Applications/SlicerApp/Testing/Python/FiducialLayoutSwitchBug1914.py
```diff
@@ -224,47 +242,20 @@ def run(self,enableScreenshots=0,screenshotScaleFactor=1):
...
       imageStats.Update()
       meanVal = imageStats.GetMean()
-      self.delayDisplay("Mean of image difference = %g" % meanVal)
+      slicer.util.delayDisplay("Mean of image difference = %g" % meanVal)
       if meanVal > 5.0:
         # raise Exception("Image difference is too great!\nExpected <= 5.0, but got %g" % (meanVal))
         print("Image difference is too great!\nExpected <= 5.0, but got %g" % (meanVal))
-        return False
+        self.delayDisplay('Test failed!')
+      self.assertFalse(meanVal > 5.0)
```
lassoan, line 258: same as above, just raise an exception in the if branch

lassoan (commented):
Thank you for clearing things up. I've added a few trivial comments in line.

Applications/SlicerApp/Testing/Python/FiducialLayoutSwitchBug1914.py
jamesobutler: I can fix the assert stuff. 

Regarding self.delayDisplay are you thinking everything should be using slicer.util.delayDisplay instead of also having a delayDisplay method in ScriptedLoadableModuleTest that then calls slicer.util.delayDisplay? Having the method in the test class helps understand what type of common methods would be use in the testing class.

Here in this test class it does seem weird to have usages of both slicer.util.delayDisplay and self.displayDisplay so I can make it consistent and just use self.delayDisplay.

lassoan: Having a separate delayDisplay in ScriptedLoadableModuleTest makes sense because it may allow changing the scale of the delay for all tests. For example, if a developer watches the test while running he wants the message to be displayed for a long enough so that the text can be read. However, if the test is run automatically they the delay can be set to a much shorter value.

delayDisplay in ScriptedLoadableModuleTest may call slicer.util.delayDisplay internally to reduce code duplication.

jamesobutler: Ok I've pushed updates to the assert stuff. Also test classes updated in this PR are using delayDisplay of ScriptedLoadableModuleTest instead of `slicer.util.delayDisplay`. Anything delayDisplay usage in ScriptedLoadableModuleLogic uses `slicer.util.delayDisplay`.

lassoan (approved):

PR #5594, opened 2021-04-15
ENH: Add set terminology action to markups view menu

cpinter:
When right-clicking a markup in the views, the subject hierarchy plugin will offer an additional action for setting the terminology. It pops up the terminology selector that sets the color and the terminology to the markup node.


cpinter:
https://user-images.githubusercontent.com/1325980/114849909-27bb1180-9dd8-11eb-972a-3506045fea53.mp4

Modules/Loadable/Markups/SubjectHierarchyPlugins/qSlicerSubjectHierarchyMarkupsPlugin.cxx
```diff
@@ -764,6 +773,100 @@ void qSlicerSubjectHierarchyMarkupsPlugin::toggleSelectPoint()
...
+  // Get markups node
+  QString nodeID = d->ViewMenuEventData["NodeID"].toString();
+  vtkMRMLNode* node = scene->GetNodeByID(nodeID.toUtf8().constData());
+  vtkMRMLMarkupsNode* markupsNode = vtkMRMLMarkupsNode::SafeDownCast(node);
+  if (!markupsNode)
+    {
+    qCritical() << Q_FUNC_INFO << ": Failed to get markups node by ID " << nodeID;
+    return;
+    }
+  markupsNode->CreateDefaultDisplayNodes();
```
lassoan, line 815: Explain in a comment why is it necessary to create display nodes.

Modules/Loadable/Markups/SubjectHierarchyPlugins/qSlicerSubjectHierarchyMarkupsPlugin.cxx
```diff
@@ -139,6 +143,10 @@ void qSlicerSubjectHierarchyMarkupsPluginPrivate::init()
   this->ToggleSelectPointAction->setObjectName("ToggleSelectPointAction");
   QObject::connect(this->ToggleSelectPointAction, SIGNAL(triggered()), q, SLOT(toggleSelectPoint()));
 
+  this->SetTerminologyAction = new QAction("Set markup terminology", q);
```
lassoan, line 146: I think "Edit" instead of "Set" would be better (similarly to "Edit properties...") and since it opens a new window, we need to add "..." to the end. So it would become "Edit markups terminology...".

The order might be improved, too. Currently, all the "point" actions are at the top, and then come all the markup actions. It would be probably better to move this option down, just below or above "Delete markup".

Modules/Loadable/Markups/SubjectHierarchyPlugins/qSlicerSubjectHierarchyMarkupsPlugin.cxx
```diff
@@ -139,6 +143,10 @@ void qSlicerSubjectHierarchyMarkupsPluginPrivate::init()
...
+  this->SetTerminologyAction->setObjectName("SetTerminologyAction");
```
lassoan, line 147: To be consistent with DeleteNodeAction name, either add "Node" to the action name to make it clear that it is for the entire node, not for a point (we may use terminology in the future to specify what a markup point is); or remove "Node" from both actions.

To be consistent with "EditPropertiesAction" name, I think it would be better to replace "Set" by "Edit".

Modules/Loadable/Markups/SubjectHierarchyPlugins/qSlicerSubjectHierarchyMarkupsPlugin.cxx
```diff
@@ -764,6 +773,100 @@ void qSlicerSubjectHierarchyMarkupsPlugin::toggleSelectPoint()
...
+
+  // Convert string list to VTK terminology entry. Do not check success, as an empty terminology is also a valid starting point
+  vtkNew<vtkSlicerTerminologyEntry> terminologyEntry;
+  const char* oldTerminologyAttribute = markupsNode->GetAttribute(vtkSegment::GetTerminologyEntryTagName());
+  std::string oldTerminologyString(oldTerminologyAttribute ? oldTerminologyAttribute : "");
+  terminologiesLogic->DeserializeTerminologyEntry(oldTerminologyString, terminologyEntry);
+
+  // If auto generated flags are not initialized, then set them to the default
+  // (color: on, name: off - this way color will be set from the selector but name will not)
+  bool nameAutoGenerated = false;
```
lassoan, line 831: When a terminology is selected then it should be possible to use it as node name. However, currently, when the popup is displayed and a term is selected the "Reset to default" button is grayed out for the name, so it is hard to make the selected terminology to be the node name (you need to type something, then the button becomes enabled, and you can click it ).

The reset to default button should be enabled (so that it can be clicked).

An alternative solution (nicer, but more work) would be to keep track if the node's name was automatically generated and if it was then make the terminology window automatically update the node name when a term is selected. Unfortunately, there is no mechanism to determine if a node name was automatically generated or not, so that would need to be implemented.

lassoan (commented):
Thank you, it looks good but I added a few small questions inline.

Modules/Loadable/Markups/SubjectHierarchyPlugins/qSlicerSubjectHierarchyMarkupsPlugin.cxx
cpinter: "Delete markup" is singular, "Edit markups terminology..." is plural. Which one should we change to make them consistent?

lassoan: You are right, we should use "markup" on the user interface. I just got used to using "markups" everywhere in the code.

Modules/Loadable/Markups/SubjectHierarchyPlugins/qSlicerSubjectHierarchyMarkupsPlugin.cxx
cpinter: Actually I think it is already implemented like what you say would be the nicer way. See `vtkSlicerTerminologiesModuleLogic::GetNameAutoGeneratedAttributeName()` in the code. The way I disabled it here, is because in the rest of this plugin it is also disabled, see `qSlicerSubjectHierarchyMarkupsPlugin::getDisplayColor` function. We can enable name auto generated, it was just a decision we made in the past (and I don't remember the reasons anymore and they are not commented so just decided to follow the convention).

lassoan: What we miss for the "nicer" solution is that we don't know if the node's name was automatically generated. In some cases it is very clear if the name is auto-generated, for example when you place a markup by clicking on the toolbar. It is also clear when a user manually renames a node that it is not auto-generated. However, when a module generates a name then it may or may not be very specific and sometimes you would want to keep that name, but sometimes it is so generic that you would prefer to replace it.

The bug in the current implementation is the initial state of the "reset name to default" button. It should become enabled when a term is selected and the term is different from the current node name.

cpinter: I see. Yes the name in this case is originally auto-generated.

I'll look into the reset name button state thing.

cpinter: The logic behind the reset button is quite complex actually, and I'm pretty sure if I change something to fix this scenario I'll brak at least another one. @lassoan so I tinhk we should discuss this.

(Btw a workaround for not being able to change the name is to change the name text and then the reset button becomes enabled)

lassoan (approved):
Thanks for the updates. The reset button initial button state need to be fixed, but it is not a regression and it does not have to block this pull request.

PR #5596, opened 2021-04-15
BUG: Fix use of qrcc when building against Python3.9

phcerdan:
This commit replaces usage of base64.encodestring for base64.encodebytes
that was deprecated in python 3.1 and removed in 3.9.

See https://bugs.python.org/issue39351

### DIFF
--- a/Utilities/Scripts/qrcc.py
+++ b/Utilities/Scripts/qrcc.py
@@ -64,7 +64,7 @@ def compileResources(in_path, out_file, args):
 
     os.remove(tmp_path)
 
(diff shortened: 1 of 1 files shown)


jcfr (dismissed):

jcfr:
Could you change the commit message to:

```
BUG: Fix use of qrcc when building against Python3.9

This commit replaces usage of base64.encodestring for base64.encodebytes
that was deprecated in python 3.1 and removed in 3.9.

See https://bugs.python.org/issue39351
```

PR #5600, opened 2021-04-18
Improve model and sequence loading defaults

lassoan:
- Improve model loading (display scalars, center in 3D view)
- Improve sequence loading (show sequence toolbar if sequence is loaded as part of a scene; show toolbar in separate row)
- Add output array name option to Probe volume with model module

These are quite independent commits, do not squash when merging.

### DIFF
--- a/Modules/Loadable/Sequences/qSlicerSequencesModule.cxx
+++ b/Modules/Loadable/Sequences/qSlicerSequencesModule.cxx
@@ -63,8 +63,9 @@ public:
   virtual ~qSlicerSequencesModulePrivate();
   QTimer UpdateAllVirtualOutputNodesTimer;
   qMRMLSequenceBrowserToolBar* ToolBar;
-  bool SequenceBrowserModuleOwnsToolBar{true};
+  bool SequencesModuleOwnsToolBar{true};
   bool AutoShowToolBar{true};
+  vtkWeakPointer<vtkMRMLSequenceBrowserNode> SequenceBrowserToShow;
 
 };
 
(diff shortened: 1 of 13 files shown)


Modules/Loadable/Sequences/qSlicerSequencesModule.cxx
```diff
@@ -105,8 +106,8 @@ void qSlicerSequencesModulePrivate::addToolBar()
 
   this->ToolBar->setWindowTitle("Sequence browser");
   this->ToolBar->setObjectName("SequenceBrowserToolBar");
-  mainWindow->addToolBar(this->ToolBar);
-  this->SequenceBrowserModuleOwnsToolBar = false;
+  mainWindow->addToolBar(Qt::BottomToolBarArea, this->ToolBar);
```
jamesobutler, line 109: Bottom toolbar area would be below the slice views? What about as a second row in the top toolbar area? I know you can already drag toolbars to create a second row.

lassoan: Initially I wanted to put the toolbar in the second row of the top area, but when I saw that you cannot set a preferred row but you have to add a toolbar break then I assumed that it would interfere with manual arrangement and automatic save/restore of toolbar positions. Placing it in the bottom  was easy and looked not nice.

After reading your comment, I've tested the toolbar break behavior and it worked perfectly, too.

Considering that placing the sequence toolbar can in a second row is a smaller change visually and that we will have to add more toolbars (for markups placement and editing, favorite layouts, etc.) it makes more sense to add all the toolbars to the top area by default. I'll update the code accordingly.

lassoan:
If there are no further comments then I'll merge this later this afternoon.

PR #5601, opened 2021-04-19
BUG: Fix failing FiducialLayoutSwitchBug1914 test case

jamesobutler:
This fixes failing test `py_FiducialLayoutSwitchBug1914` as observed on the CDash for all platforms. As part of this I simplified the widget class down as you can the reload and test to rerun things. Those specific widgets weren't needed and is now based on some other tests that specifies a widget class that just use the base class ScriptedLoadableModuleWidget.

Failing test:
https://slicer.cdash.org/test/17315997

Introduced by changes made in 71f236bfacffec1d6c01036840dfc8d89be0563c.


lassoan (approved):

PR #5603, opened 2021-04-20
ENH: Scalar Volume Segment Statistics use volume threshold

nathanbmnt:
Currently scalar volume segment statistics include every voxel value in the segmentation, even if they are not visible because they are outside of the display thresholds. This PR allows the option to only include voxel values within volume thresholds for statistics.
image

I have also put the logic for getting the array of voxel values into a method called `getVoxelsInSegment`. This will be helpful for anyone creating a plugin which is a child class of `ScalarVolumeSegmentStatisticsPlugin`.

### DIFF
--- a/Modules/Scripted/SegmentStatistics/SegmentStatisticsPlugins/ScalarVolumeSegmentStatisticsPlugin.py
+++ b/Modules/Scripted/SegmentStatistics/SegmentStatisticsPlugins/ScalarVolumeSegmentStatisticsPlugin.py
@@ -1,6 +1,9 @@
 import vtk, slicer
+from vtk.util.numpy_support import vtk_to_numpy
 from SegmentStatisticsPlugins import SegmentStatisticsPluginBase
 from functools import reduce
+import numpy as np
+import qt
 
 
 class ScalarVolumeSegmentStatisticsPlugin(SegmentStatisticsPluginBase):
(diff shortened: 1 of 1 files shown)


pieper:
Interesting observation.  My reaction is that this should be an option that is not enabled by default, and when it's enabled you get a second set of statistics that is flagged as being on the thresholded data.  My reasoning is that like the window/level setting It's not clear that people would expect the display threshold to impact their quantification.  It might make sense to have an optation that creates a new volume node with the display "baked in" similar to what hardening does in the transforms module, and then we wouldn't need to make the segment statistics more complex.

Modules/Scripted/SegmentStatistics/SegmentStatisticsPlugins/ScalarVolumeSegmentStatisticsPlugin.py
```diff
@@ -35,6 +37,40 @@ def computeStatistics(self, segmentID):
       # Input grayscale node does not contain valid image data
       return {}
 
+    voxelsInSegment = self.getVoxelsInSegment(segmentationNode, segmentID, grayscaleNode)
+
+    # Only compute stats using voxels within the display node thresholds for the volume
+    threshold_min = grayscaleNode.GetDisplayNode().GetLowerThreshold()
+    threshold_max = grayscaleNode.GetDisplayNode().GetUpperThreshold()
+    arrayThresholded = voxelsInSegment[voxelsInSegment >= threshold_min]
+    arrayThresholded = arrayThresholded[arrayThresholded <= threshold_max]
```
jamesobutler, line 46: You should maintain current functionality of being able to calculate statistics no matter if the voxel is within the threshold or not. If you want the option of also applying threshold to ignore those voxels in statistics calculations, I would suggest to create another parameter setting. This might exist inside the specific plugin parameters section, or in the Advanced SegmentStatistics module if applicable to other plugins.

lassoan:
You would normally use masking settings in Segment Editor to create segments appropriately for the first place - taking account the volume's intensity range (using Threshold effect -> Use for masking) instead of applying some secondary masking based on current display settings.

I can imagine that there could be workflows where a dynamic secondary masking makes things more convenient, but I would not establish a precedent in Slicer core where display settings have any influence on quantification results. To get this new computation method integrated we would also need to do some more validation work, for example there may be a change in how multi-component scalar values are quantified, what is done with NaN values, need to check how threshold bounds should be interpreted (< or <=; > or >=), we would need to add tests, etc.

Instead of merging this change into the Slicer core, I would recommend to add it as a new plugin in an extension (just clone this plugin and rename and add a tiny scripted module that registers the plugin). The name could be something like "DisplayedVolumeStatistics" or "ThresholdedVolumeStatistics". If you don't want to create a new extension for this then you can send a pull request to the https://github.com/PerkLab/SlicerSandbox extension.

jamesobutler:
To support `ThresholdedVolumeStatistics(ScalarVolumeSegmentStatisticsPlugin)` without having to override the large `computeStatistics` method, would changes be approved to support plugging in this option of applying stats based on threshold in the custom statistics plugin outside of Slicer core?

jamesobutler:
The use case is segmenting a tumor in an ultrasound image. The volume statistic is important, but also brightness can be an important metric. The desire is to use that same segment to get statistics that can quantify necrosis in the tumor. With ultrasound, thresholding out the noise floor to remove the speckle in the necrotic regions where it is known to not have signal is desired. Is there a workflow in Segment Editor to support this task?

nathanbmnt:
Pushed a commit which makes using thresholding an option for ScalarVolumeSegmentStatistics. Off by default.

I would like to be able to create a child class of ScalarVolumeSegmentStatisticsPlugin and pull voxel values with the logic in the parent class. Then in the child class run my own custom stats on those voxels. Would a Slicer core pull request just adding the `getVoxelsInSegment` method in this PR be acceptable?

lassoan:
> To support ThresholdedVolumeStatistics(ScalarVolumeSegmentStatisticsPlugin) without having to override the large computeStatistics method, would changes be approved to support plugging in this option of applying stats based on threshold in the custom statistics plugin that would be outside of Slicer core?

Yes, it should be no problem to reorganize the plugin in Slicer core to reduce the amount of code that needs to be overridden in derived classes.

> The use case is segmenting a tumor in an ultrasound image. 

I'm not sure how segment statistics is related to segmentation. Do you mean you identify multiple blobs in some way and then you want to use statistics to decide if is a tumor or not? For that, it could be better to work with the entire histogram not just a few parameters. You can then easily discard any intensity range. You may also consider applying some preprocessing (more that just thresholding) that makes the image invariant to parameter changes, such as dynamic range, power, and threshold.

jamesobutler:
We segment a tumor in an ultrasound image. We then use the SegmentStatistics plugins to provide information on the volume of our segment. We also want to use that same segmented area to provide an additional metric about the tumor as it relates to necrosis inside the tumor. Therefore we generate positivity % of positive pixels values inside the segment. To do this we must first threshold out speckle noise from areas that we know to be empty in the necrotic region of the tumor. This has been a successful technique for us and has provided a simple analysis workflow. 

> Yes, it should be no problem to reorganize the plugin in Slicer core to reduce the amount of code that needs to be overridden in derived classes.

Ok we will do this. Creation of multiple objects or additional manual preprocessing steps increases complexity for our users and something we are not interested in.

Generally we take the approach of “what you see is what you get” and Slicer often at times breaks that paradigm and so we solve that in our custom application to improve the user experience. It is unexpected for our users to get results based on image data that they can’t see actively being displayed in the slice views.

lassoan:
Would you burn in the window/level conversion into the image, too? If you apply thresholding then you may have to  would be somewhat inconsistent if you only applied the thresholding part.

Thank you for contributing this feature and

nathanbmnt:
>Would you burn in the window/level conversion into the image, too? 

What do you mean by conversion of W/L?

lassoan:
I mean that if you want to quantify the displayed values (and not the stored values) then you may want to not just apply the threshold but also the linear scaling/offset ([window width/level](https://slicer.readthedocs.io/en/latest/user_guide/user_interface.html#adjusting-image-window-level)) as well.

nathanbmnt:
I've created a new PR https://github.com/Slicer/Slicer/pull/5609 which only includes the `getVoxelsInSegment` method from this PR.

PR #5609, opened 2021-04-21
ENH: ScalarSegmentStatsPlugin has method to get vtkStencil from segment

nathanbmnt:
I am making a child class of ScalarVolumeSegmentStatisticsPlugin outside of Slicer core, and I would like to make use of the code that already exists for aligning the segment to the volume image data and creating a vtkStencil. This refactor adds a method getStencilForVolume to ScalarVolumeSegmentStatisticsPlugin which returns a vtkStencil.

An example child class would be:
```
class ThresholdedStatsPlugin(ScalarVolumeSegmentStatisticsPlugin):
    ...
    def computeStatistics(self, segmentID):
        segmentationNode = slicer.mrmlScene.GetNodeByID(self.getParameterNode().GetParameter("Segmentation"))
        grayscaleNode = slicer.mrmlScene.GetNodeByID(self.getParameterNode().GetParameter("ScalarVolume"))
        stencil = self.getStencilForVolume(segmentationNode, segmentID, grayscaleNode)
        threshold_lower, threshold_upper = 10, 100
        # Use vtkStencil and vtkImageData to compute stats...
```

### DIFF
--- a/Modules/Scripted/SegmentStatistics/SegmentStatisticsPlugins/ScalarVolumeSegmentStatisticsPlugin.py
+++ b/Modules/Scripted/SegmentStatistics/SegmentStatisticsPlugins/ScalarVolumeSegmentStatisticsPlugin.py
@@ -14,7 +14,6 @@ class ScalarVolumeSegmentStatisticsPlugin(SegmentStatisticsPluginBase):
     #... developer may add extra options to configure other parameters
 
   def computeStatistics(self, segmentID):
-    import vtkSegmentationCorePython as vtkSegmentationCore
     requestedKeys = self.getRequestedKeys()
 
     segmentationNode = slicer.mrmlScene.GetNodeByID(self.getParameterNode().GetParameter("Segmentation"))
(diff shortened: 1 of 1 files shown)


lassoan:
We are happy to integrate refactoring of the class (reorganizing methods), but switching the algorithm to use numpy instead of VTK would be a major change. Computing a histogram and basic statistics may seem like a trivial operation but it is not simple at all. There are many details that varies across implementations (how do you choose number of histogram bins, if you don't use a histogram then computation may be very slow for large arrays - is this speed acceptable?, how do you handle images with multiple components, inf and NaN values, etc.).

We would not like to change the current implementation, unless there is a strong reason for it, because then we would need to revalidate it. [Current SegmentStatistics tests](https://github.com/Slicer/Slicer/blob/0daa3b92680706c1e98b6585a676545f41c0b8a2/Modules/Scripted/SegmentStatistics/SegmentStatistics.py#L707-L926) only checks basic cases and does not test any non-trivial cases. We would need to invest time into improving these testing, otherwise any change could introduce regressions.

I would recommend to update this pull request to only rearrange methods but do not change the computation method.

nathanbmnt:
> computation may be very slow for large arrays

I created a segmentation that was 342x513x452 voxels and this PR took 1.77 seconds to get all stats. In Slicer develop the same volume same segmentation took 1.35 seconds.

> how do you handle images with multiple components, inf and NaN values, etc.).

At the moment Scalar Volume Statistics handles NaN and inf incorrectly. Here are stats for a segmentation of all NaNs: min and max are at about the double precision limits (^299 instead of ^308 though?). And I don't know where that 65535 for median is coming from.
image

All inf has incorrect minimum and median
image

This PR returns no columns for all NaN because those values dissapear with `array = array[array > float('-inf')]`
image

This PR returns proper values for all inf stats
image

pieper:
Why not make a subclass plugin for doing the stats calc in numpy.  Then people can use whichever makes sense and we can more easily see if issue crop up in one or the other (perhaps if enabled it would automatically compare with vtk and flag any disagreements).  I agree with @lassoan that changing the default would need to be strongly motivated.  The nan and inf cases don't occur much in practice, but bin size issues might and would be less obvious.

>  In Slicer develop the same volume same segmentation took 1.35 seconds.

If that was a debug build it's probably a lot slower than the release would be.  Did you time that?

lassoan:
I'm torn on this.

The 23% slowdown is significant, the numpy implementation is much more opaque to us (we practically cannot debug into it and cannot make significant changes to it, or replicate the exact behavior in C++ if we needed that somewhere else), users don't gain anything from this internal change, and it requires more effort from all of us (add more tests, do more analysis of slowdown and increased memory usage in a number of scenarios, review all code and tests).

On the other hand numpy is probably tested by more people than these two VTK classes, and these VTK classes are not particularly sophisticated, so we don't lose much by not using them.

If we make a change then it would also make sense to look at what ITK can offer (via SimpleITK; or by adding vtkITK wrapping for the relevant classes).

Considering the amount of work that we would still need to put in aVTK to numpy switch and that there would be no benefits for the users, @pieper's suggestion of doing the switch in the new class seems to be more attractive to me.

nathanbmnt:
> If that was a debug build it's probably a lot slower than the release would be. Did you time that?

I timed using an install of slicer 4.11. I modified  "C:\Program Files\Slicer 4.11.0-2020-09-17\lib\Slicer-4.11\qt-scripted-modules\SegmentStatisticsPlugins\ScalarVolumeSegmentStatisticsPlugin.py" to add time.time() to the beginning and end of computeStatistics.

What is the binning issue? Are you guys saying there is binning occurring in vtkImageHistogramStatistics to estimate median and standard deviation rather than use every value?

I've modified this PR so now the only change is that there is a method which inputs the segment and volume, and outputs the vtkStencil of the segment in the volume coordinates. There is no numpy or change of logic.

lassoan (approved):
This looks good to me. Thanks for your patience.

PR #5611, opened 2021-04-21
ENH: Add restoreGUIState/saveGUIState to qSlicerMainWindow public API

jcfr:
This is useful when writing custom applications overriding the default
application close behavior.

See https://www.slicer.org/wiki/Documentation/Nightly/ScriptRepository#Override_application_close_behavior

### DIFF
--- a/Base/QTApp/qSlicerMainWindow.h
+++ b/Base/QTApp/qSlicerMainWindow.h
@@ -117,6 +117,22 @@ public slots:
   virtual void on_PasteAction_triggered();
   virtual void on_ViewExtensionsManagerAction_triggered();
 
+  /// Read GUI state from application settings and update the user interface accordingly.
+  ///
+  /// GUI state includes:
+  /// - main window state and geometry (only if MainWindow/geometry application setting is
+  ///   enabled or force argument is set to true)
+  /// - current view layout ID
+  /// - favorite modules
+  /// - recently loaded files
+  ///
+  /// \sa restoreGUIState()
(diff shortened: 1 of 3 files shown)


Base/QTApp/qSlicerMainWindow.h
```diff
@@ -64,6 +64,18 @@ class Q_SLICER_BASE_QTAPP_EXPORT qSlicerMainWindow : public QMainWindow
   /// \sa moduleSelector(), pythonConsole()
   Q_INVOKABLE ctkErrorLogWidget* errorLogWidget()const;
 
+  /// Read windows settings and update user interface.
+  ///
+  /// \sa writeSettings()
+  /// \sa qSlicerMainWindowPrivate::setupRecentlyLoadedMenu
+  Q_INVOKABLE void readSettings();
```
lassoan, line 71: Make them virtual so that more things can be saved in derived classes in custom applications.

lassoan (commented):
Thank you, this will be useful. There are just a few formalities to address, because the poorly named and documented methods that are OK in private are not good enough in the public API. See details inline.

jcfr:
Thanks for the review. It all makes sense. Once I am done working on transition to supporting the new package+extension infrastructure (discourse post to follow, and update tmrw), I will work on this.

lassoan:
@jcfr I've updated the code based on my review comments. Please have a look. Thank you!

jcfr:
Locally tested :+1:

PR #5613, opened 2021-04-22
ENH: add support for Scanco file format

dzenanz:
This file format is frequently used in small animal CT imaging.


jcfr (commented):
Thanks for the patch. Could you update the documentation ? See https://slicer.readthedocs.io/en/latest/user_guide/data_loading_and_saving.html#supported-data-formats

dzenanz:
Doc update added.

lassoan:
Thank you! It seems that this module can read/write a couple of more file extensions:

https://github.com/KitwareMedical/ITKIOScanco/blob/a913b8736e53901edef3be2529d1eacdb3e5d88b/src/itkScancoImageIO.cxx#L53-L56

Is there a specific reason for just adding .isq to Slicer?

dzenanz:
The other formats are less well supported, and I don't have testing data for them. So I would rather not advertised them as "supported".

lassoan (approved):

dzenanz:
Thank you for merging, and quickly too.

Issue #5616, opened 2021-04-23
Windows Taskbar grouping


fbordignon:
When I pin the Slicer launcher to the windows taskbar, it works for launching slicer but the actual slicerapp-real.exe icon is exhibited on a separate button. i.e. I end up with the shortcut for the launcher pinned and with the actual running slicer in another button. From what I gather the problem is that Slicer.exe and SlicerApp-real.exe are separate processes so windows treat them as separate programs from the taskbar perspective.

It would be great if the icons could be grouped to save taskbar space.

I've tried to add
```
#include<shobjidl_core.h>
SetCurrentProcessExplicitAppUserModelID(L"NAMIC.3DSlicer.4.11");
```
to the main functions of qSlicerApplicationMainWrapper.cxx and to Main.cpp of CTKAppLauncherLib with no success. They remain as separate icons at the taskbar. I am not sure of the exact position where the call need to be inserted.

Followed the instructions on https://docs.microsoft.com/en-us/windows/win32/api/shobjidl_core/nf-shobjidl_core-setcurrentprocessexplicitappusermodelid

lassoan:
It would be a nice touch to group the launcher and the real executable into the same taskbar button.

I would recommend to use the Qt implementation (https://doc.qt.io/qt-5/qtwin.html#setCurrentProcessExplicitAppUserModelID) and make it the very first call. It is too late if you set it when a window is already displayed.

The launcher will have to read the model ID from the launcher settings file, because the exact same executable is used for all applications (building the launcher requires commercial Qt license due to building Qt statically).

If you pin a shortcut (and not a running application) then Windows cannot possibly know what the AppUserModelID would be of the running process. Probably you should set the AppUserModelID in the shortcut (see https://stackoverflow.com/questions/61806380/how-to-set-appusermodelid-of-application-pinned-to-taskbar). Since the shortcut is created during installation by NSIS, you need to find out how to specify the model ID for a shortcut using NSIS. Some potentially relevant pages: https://nsis.sourceforge.io/WinShell_plug-in, https://github.com/safing/nsis-shortcut-properties.

You may find this page useful: https://jcutrer.com/windows/find-aumid

fbordignon:
Hi guys, I don't know if I need to work on this any further or if it is a matter of testing.
Let me know if you need anything more from my side.

Thanks.

lassoan:
We'll try to get this reviewed and merged before the new stable release (within a few weeks).

lassoan:
[...]

PR #5619, opened 2021-04-27
ENH: Add windows AppID to group taskbar icons

fbordignon:
Tying to work on issue #5616
Setting the CLSID and AppUserModelID on the shortcut of the slicer installation package successfully groups the shortcut with the SlicerApp-real.exe. If the user pins the Slicer.exe launcher from the installation dir to the taskbar, it also works. Not sure why.
I was unsure whether to use `Slicer_MAIN_PROJECT_VERSION_FULL` or `Slicer_VERSION_FULL` (appname as well), please take a look at the appid construction. I've tested it on a SlicerCAT for our app, if you can be so kind and check it against Slicer it would be great.


Base/QTApp/qSlicerApplicationMainWrapper.cxx
```diff
@@ -24,11 +24,16 @@
 #if defined (_WIN32) && !defined (Slicer_BUILD_WIN32_CONSOLE)
 # include <windows.h>
 # include <vtksys/Encoding.hxx>
+# include<shobjidl_core.h>
```
lassoan, line 27: AppUserModelID should be set both in case of console and non-console application (not just when `!defined (Slicer_BUILD_WIN32_CONSOLE)`).

Base/QTApp/qSlicerApplicationMainWrapper.cxx
```diff
@@ -24,11 +24,16 @@
...
 
 int __stdcall WinMain(HINSTANCE hInstance,
                       HINSTANCE hPrevInstance,
                       LPSTR lpCmdLine, int nShowCmd)
 {
+  // set AppUserModelID to group Slicer windows under the same icon on the taskbar.
+  #define AUMID Slicer_ORGANIZATION_NAME "." Slicer_MAIN_PROJECT_APPLICATION_NAME "." Slicer_MAIN_PROJECT_VERSION_FULL
+  SetCurrentProcessExplicitAppUserModelID(vtksys::Encoding::ToWide(AUMID).c_str());
```
lassoan, line 35: This is a special API, but SetCurrentProcessExplicitAppUserModelID seems to be supported on Windows7, so it should be OK.
https://docs.microsoft.com/en-us/windows/win32/api/shobjidl_core/nf-shobjidl_core-setcurrentprocessexplicitappusermodelid

CMake/SlicerCPackShortcutHelper.nsh
```diff
@@ -0,0 +1,133 @@
+; Copyright Safing ICS Technologies GmbH. Use of this source code is governed by the Apache 2 license that can be found in the LICENSE file.
+; Adapted from https://github.com/safing/nsis-shortcut-properties/blob/b5ede73fc49665449731df5d0f945192329b0ef6/shortcut-properties.nsh
```
lassoan, line 2: Can you confirm that it does not require user access elevation (can run the installer as a regular user)?

Is there a minimum required Windows version for this to work? If it works on Windows7 (it does not matter if it displays an error or warning, as long as it can be installed) then it is good. If it requires anything above Windows7 then we need to update the documentation with that requirement (https://slicer.readthedocs.io/en/latest/user_guide/getting_started.html#operating-system-versions).

lassoan (commented):
This is pretty hardcore and can cause some issues if we are not careful, but I think it should work and would polish the user experience. I've added a number of comments inline.

CMake/SlicerCPack.cmake
```diff
@@ -390,6 +390,18 @@ if(CPACK_GENERATOR STREQUAL "NSIS")
...
   slicer_verbose_set(CPACK_PACKAGE_EXECUTABLES "..\\\\${EXECUTABLE_NAME}" "${PACKAGE_APPLICATION_NAME}")
 
+  # Set AppID for the shortcut
+  set(CPACK_NSIS_INCLUDE_SHORTCUT_SCRIPT "\\\"${Slicer_SOURCE_DIR}/CMake/SlicerCPackShortcutHelper.nsh\\\"")
+  STRING(REPLACE "/" "\\\\" CPACK_NSIS_INCLUDE_SHORTCUT_SCRIPT  ${CPACK_NSIS_INCLUDE_SHORTCUT_SCRIPT} )
+  # Random CLSID generated with uuidgen.exe
+  set(CPACK_NSIS_CREATE_ICONS_EXTRA
+    "  !include ${CPACK_NSIS_INCLUDE_SHORTCUT_SCRIPT}
+  !insertmacro ShortcutSetAppID \\\"$SMPROGRAMS\\\\$STARTMENU_FOLDER\\\\${PACKAGE_APPLICATION_NAME}.lnk\\\"\
+ \\\"{066519E1-76BE-4EE0-B1FF-B5650EB23BC9}\\\"\
```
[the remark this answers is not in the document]
fbordignon: Ok, I've used this uuid as a root namespace uuid and generated a new one based on it and the AppID

Base/QTApp/qSlicerApplicationMainWrapper.cxx
```diff
@@ -24,11 +24,16 @@
...
 # include <windows.h>
 # include <vtksys/Encoding.hxx>
+# include<shobjidl_core.h>
 
 int __stdcall WinMain(HINSTANCE hInstance,
                       HINSTANCE hPrevInstance,
                       LPSTR lpCmdLine, int nShowCmd)
 {
+  // set AppUserModelID to group Slicer windows under the same icon on the taskbar.
+  #define AUMID Slicer_ORGANIZATION_NAME "." Slicer_MAIN_PROJECT_APPLICATION_NAME "." Slicer_MAIN_PROJECT_VERSION_FULL
```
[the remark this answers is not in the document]
fbordignon: Sure, I've defined a CMAKE variable for the AppId and used it both on CPack and C++. Great catch!

Base/QTApp/qSlicerApplicationMainWrapper.cxx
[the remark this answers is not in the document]
fbordignon: Committed together with the fixes.

CMake/SlicerCPack.cmake
```diff
@@ -390,6 +390,18 @@ if(CPACK_GENERATOR STREQUAL "NSIS")
...
+ \\\"${Slicer_ORGANIZATION_NAME}.${APPLICATION_NAME}.${Slicer_MAIN_PROJECT_VERSION_FULL}\\\"
```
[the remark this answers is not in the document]
fbordignon: Ok, agree.

Base/QTApp/qSlicerApplicationMainWrapper.cxx
fbordignon: The solution I proposed for this one is not so elegant. Please advise.

CMake/SlicerCPackShortcutHelper.nsh
fbordignon: I can confirm that no elevation is required. I could install at my local user with no problem.
Confirmed that it runs on win7 using the demo nsis project from the linked original github project, i.e. tested the nsh script outside the slicer build environment on win7 and it worked.

fbordignon:
I've tested it again this week and the slicer organization was missing. Adding it to the sanity checks made it available for me at the appid concatenation.

fbordignon:
I've reverted the sanity check as it was crashing cmake

jamesobutler (commented):
I can confirm that it successfully group the 2 now.

| Current | This PR |
|----------|---------|
|image |image|



I did notice there is a name mismatch of the application name in File Explorer and the window titles. "3D Slicer 4.13.0-2021-09-27" vs "Slicer 4.13.0-2021-09-27".
image

Probably related to the shortcut?  @lassoan Should the application shortcut be "Slicer" or "3D Slicer"?
image
image

lassoan:
> I did notice there is a name mismatch of the application name in File Explorer and the window titles. "3D Slicer 4.13.0-2021-09-27" vs "Slicer 4.13.0-2021-09-27".

If the console is not displayed (this is how official releases are created) then the behavior does not change. If the console is displayed then the window group will contain titles starting with "Slicer" and "3D Slicer" but the group is named as "Slicer..." (it probably comes by the shortcut name). This is a very subtle change and since the shortcut name, shortcut folder, process name, console window title all start with "Slicer", the change should not cause any confusion. So, I think the new behavior would be fine.

---

Unfortunately, there is a serious side-effect of setting a static, globally unique identifier using SetCurrentProcessExplicitAppUserModelID: you could not install multiple Slicer instances completely independently. For example:
- install Slicer (it creates the shortcut and assigns the AppUserModelID to the shortcut)
- copy the application folder to another folder (or install to another folder)
- start Slicer -> they both show up in the same group => not nice but could be tolerable
- pin Slicer to the task bar -> regardless of which instance you were running, always the same shortcut will be pinned to the taskbar => this is a serious issue, which prevents merging of this pull request as is

It would be better if the launcher would get its own AppUserModelID using GetCurrentProcessExplicitAppUserModelID (it contains the full path of the application, so each Slicer installation would have its own unique ID) and save it in an environment variable (using base64 or similar encoding that works regardless of the application code page), and the launched application would take that environment variable, reconstruct the AppUserModelID and set it using SetCurrentProcessExplicitAppUserModelID. In this case, both grouping and pinning should work without the need to mess with the shortcut and each Slicer installation would be independent. @fbordignon could you please explore if this method indeed works?

lassoan:
I did some testing and I can confirm that explicitly setting the AppUserModelID for the application and console window of the launcher makes the console and the application show up in the same group. AppUserModelID can be generated on-the-fly from the launcher's full path, so that can be used to generate a quasi-unique AppUserModelID for each Slicer instance (e.g., by using organization.application.truncatedMD5hashOfFullPath as AppUserModelID).

If the launcher is a console window then pinning the group on the taskbar will pin the launcher, which is good.

If a non-console launcher is used and the application is pinned then the taskbar will pin SlicerApp-real.exe by default, which is not good (it cannot launch the application). Fortunately, there is a nice solution for this: we need to set the relaunch command as explained [here](https://stackoverflow.com/questions/57676759/how-to-allow-launcher-application-to-be-pinned-to-the-taskbar).

The approach described here satisfies all the needs that the proposal in the pull request (console and application window are grouped, application can be pinned) without the issue of grouping completely unrelated application install trees.

Since the solution provided in this pull request has a serious limitation and there is an [alternative approach](https://github.com/Slicer/Slicer/issues/5616#issuecomment-984314471) that solves the same issues but does not suffer from that limitation, I close this pull request now.

@fbordignon thanks for all your work in exploring this problem and providing an almost-perfect solution. It would be great if you could implement the described alternative design, but I understand if you cannot afford to spend more time with this.

PR #5620, opened 2021-04-27
COMP: Fix configruation of qSlicerSimpleMarkupsWidgetTest1 with QtTesting OFF

jcfr:
This commit fixes a regression introduced in 0242927e3 (ENH: Add unit tests
for markups factory)

This pull request fixes the following error:

```
CMake Error at CMake/SlicerMacroSimpleTest.cmake:61 (message):
  error: Target 'qSlicerMarkupsModuleWidgetsCxxTests' is not defined !
  Modules/Loadable/Markups/Widgets/Testing/Cxx/CMakeLists.txt:46 (simple_test)
```


lassoan (approved):
I didn't really follow all the details but it generally looks good. Was there any specific reason to make the test available without QtTesting?

jcfr:
It doesn't depend on QtTesting and this was the issue was a copy/paste problem. In a later commit we will bumb the minimum cmake, and re introduce use of the macro building the test driver and associate it with an object file ...

PR #5629, opened 2021-05-05
Add support for uploading application package to Girder based server

jcfr:
This PR was incorrectly labelled,  it has been renamed and updated.

Changes related to the support for the new extension frontend and backend are published in https://github.com/Slicer/Slicer/pull/5637


jcfr:
Integrating after manually validating.

PR #5631, opened 2021-05-07
BUG: Fix missing attributes when proxy node is changed

riep:
Attributes of proxy nodes are lost after vtkSlicerSequencesLogic::UpdateProxyNodesFromSequences.
Before updating the proxy node copy all its attributes.
Naming of variable "source" and "target" may be confusing with existing targetProxyNode and sourceDataNode variable names.

### DIFF
--- a/Modules/Loadable/Sequences/Logic/vtkSlicerSequencesLogic.cxx
+++ b/Modules/Loadable/Sequences/Logic/vtkSlicerSequencesLogic.cxx
@@ -386,7 +386,7 @@ void vtkSlicerSequencesLogic::UpdateProxyNodesFromSequences(vtkMRMLSequenceBrows
       }
 
     // Get the current target output node
-    vtkMRMLNode* targetProxyNode=browserNode->GetProxyNode(synchronizedSequenceNode);
+    vtkMRMLNode* targetProxyNode = browserNode->GetProxyNode(synchronizedSequenceNode);
     if (targetProxyNode!=nullptr)
       {
       // a proxy node with the requested role exists already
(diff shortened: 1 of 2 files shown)


cpinter (commented):
Probably it would be enough to call vtkMRMLNode::CopyContent instead of creating this new function `CopyNodeAttributes`

riep:
@lassoan What do you think about this modification. This was inspired by one of your posts https://discourse.slicer.org/t/are-mrml-attributes-saved-in-sequences/7698/2

lassoan:
Node attributes are already copied, see [here](https://github.com/Slicer/Slicer/blob/72852d7bfd6920a424092d724d26c84ffdd950a7/Libs/MRML/Core/vtkMRMLNode.cxx#L139).

What is the problem that you are trying to solve with this change?

riep:
It is true but an attribute copy should be done first in the other way: target -> source.

riep:
Thanks Andreas, could you be a bit more specific? How can we specify such thing.

lassoan:
The option can be enabled for each synchronized node:

image

riep:
ok thank you, this is working.

lassoan:
@riep I see that you closed the PR and then reopened it. Is there still anything that is not synchronized properly without your proposed changes?

riep:
Hi @lassoan,
I was able to reproduce this bug in slicer and in a slicelet:

2. Add an attribute to a proxyNode
3. Save the scene in mrb
4. close Slicer
5. Load the mrb
6. Play with the replay 
7. The proxy node have lost its attribute

For some reason, you need to restart and reload to reproduce the bug. Does it ring any bell ?

lassoan:
This works well for me. Please provide more specific instructions or maybe a screen capture video.

PR #5633, opened 2021-05-09
ENH: Set default VTK version to 9

lassoan:
Slicer Preview Releases are built with VTK9 and there are no major issues, therefore we update the default VTK version to 9.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -328,7 +328,7 @@ mark_as_superbuild(Slicer_INSTALL_ITKPython)
 option(Slicer_BUILD_PARAMETERSERIALIZER_SUPPORT "Build Slicer with parameter serializer support" OFF)
 mark_as_superbuild(Slicer_BUILD_PARAMETERSERIALIZER_SUPPORT)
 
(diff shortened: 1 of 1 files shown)


jcfr:
Should wait for https://github.com/Slicer/Slicer/pull/5629 to be finalized and integrated so that we do release 4.14, and then update the default VTK to 9 in 4.15 ?

cpinter:
I agree to wait a bit. For the Extension Manager update as @jcfr mentions, but also because some of the prominent extensions do not build with VTK9 (VMTK, VR - btw do we have a more complete list?). Also it makes sense to me making the change with the release so that we have a clear cut and can identify potential related issues raised by the community (i.e. "I have a problem with this basic feature in my slightly special setup in 4.15").

jcfr (approved):

jcfr:
Let's also check if the requirements/configure/build documentation need to be updated to reflect this. (e.g GPU requirements)

lassoan:

jamesobutler:
@lassoan I believe this can be merged now.

PR #5637, opened 2021-05-11
Integrate with updated infrastructure managing extensions

jcfr:


jcfr:
Prior re-factoring has been pushed into a dedicated pull request. See https://github.com/Slicer/Slicer/pull/5733

PR #5640, opened 2021-05-12
Sh combobox improvements

cpinter:
1. Introduce None item in subject hierarchy
    - If enabled, an "extra item" (similarly to the same concept in qMRMLNodeComboBox) is added as the top item under the scene. This item has the name defined in the noneItemText property, and is selectable but no drag&drop is enabled. When selecting it, the current item will have the ID 0 (= INVALID_ITEM_ID)
    - This enables having empty selections in comboboxes after having made a valid selection

    image

2. Add flag in subject hierarchy combobox for showing parents in title
    - Sometimes it is undesired to see the name of each parent after selecting an item in the combobox (it looks like "[ParentName] / [ParentName] / [SelectedItemName]"). If disabled, only the name of the selected item is shown ("[SelectedItemName]")
    - On by default not to change the current behavior

3. If an include attribute node filter was applied, hierarchy (patient/study/folder) items that were left without shown children were still visible, while they were supposed to only be shown if any child was shown

4. If an include attribute node filter was applied, hierarchy items were still selectable, but due to the filter only the items accepted based on the filter are supposed to be interacted with

### DIFF
--- a/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyComboBox.cxx
+++ b/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyComboBox.cxx
@@ -58,6 +58,7 @@ public:
 public:
   int MaximumNumberOfShownItems;
   bool AlignPopupVertically;
+  bool ShowCurrentItemParents;
 
   qMRMLSubjectHierarchyTreeView* TreeView;
   QLabel* NoItemLabel;
@@ -68,6 +69,7 @@ qMRMLSubjectHierarchyComboBoxPrivate::qMRMLSubjectHierarchyComboBoxPrivate(qMRML
   : q_ptr(&object)
   , MaximumNumberOfShownItems(20)
   , AlignPopupVertically(true)
+  , ShowCurrentItemParents(true)
   , TreeView(nullptr)
   , NoItemLabel(nullptr)
 {
@@ -469,6 +471,60 @@ void qMRMLSubjectHierarchyComboBox::setAlignPopupVertically(bool align)
   d->AlignPopupVertically = align;
 }
 
+//--------------------------------------------------------------------------
+bool qMRMLSubjectHierarchyComboBox::noneEnabled()const
+{
+  if (!this->model())
+    {
+    return false;
+    }
+  return this->model()->noneEnabled();
+}
+
+//--------------------------------------------------------------------------
+void qMRMLSubjectHierarchyComboBox::setNoneEnabled(bool enable)
+{
+  if (!this->model())
+    {
+    return;
+    }
+  this->model()->setNoneEnabled(enable);
+}
+
+//--------------------------------------------------------------------------
+QString qMRMLSubjectHierarchyComboBox::noneDisplay()const
+{
+  if (!this->model())
+    {
+    return QString();
+    }
+  return this->model()->noneDisplay();
+}
+
+//--------------------------------------------------------------------------
+void qMRMLSubjectHierarchyComboBox::setNoneDisplay(const QString& displayName)
+{
+  if (!this->model())
+    {
+    return;
+    }
+  this->model()->setNoneDisplay(displayName);
+}
+
+//--------------------------------------------------------------------------
+bool qMRMLSubjectHierarchyComboBox::showCurrentItemParents()const
+{
+  Q_D(const qMRMLSubjectHierarchyComboBox);
+  return d->ShowCurrentItemParents;
+}
+
+//--------------------------------------------------------------------------
+void qMRMLSubjectHierarchyComboBox::setShowCurrentItemParents(bool show)
+{
+  Q_D(qMRMLSubjectHierarchyComboBox);
+  d->ShowCurrentItemParents = show;
+}
+
 //-----------------------------------------------------------------------------
 void qMRMLSubjectHierarchyComboBox::showPopup()
 {
@@ -520,8 +576,10 @@ void qMRMLSubjectHierarchyComboBox::showPopup()
   else
     {
     // Height based on the number of items
-    const int numberOfRows = qMin(displayedItemCount, d->MaximumNumberOfShownItems);
-    popupHeight = numberOfRows * d->TreeView->sizeHintForRow(0);
+    const int numberOfShownShItems = qMin(displayedItemCount, d->MaximumNumberOfShownItems);
+    const int numberOfRows = (this->noneEnabled() ? numberOfShownShItems + 1 : numberOfShownShItems);
+    const int referenceRowHeight = (this->noneEnabled() ? d->TreeView->sizeHintForRow(1) : d->TreeView->sizeHintForRow(0));
+    popupHeight = numberOfRows * referenceRowHeight;
 
     // Add tree view margins for the height
     // NB: not needed for the width as the item labels will be cropped
@@ -602,6 +660,14 @@ void qMRMLSubjectHierarchyComboBox::mousePressEvent(QMouseEvent* e)
 void qMRMLSubjectHierarchyComboBox::updateComboBoxTitleAndIcon(vtkIdType selectedShItemID)
 {
   Q_D(qMRMLSubjectHierarchyComboBox);
+
+  // Hide popup
+  QFrame* container = qobject_cast<QFrame*>(this->view()->parentWidget());
+  if (container)
+    {
+    container->hide();
+    }
+
   vtkMRMLSubjectHierarchyNode* shNode = d->TreeView->subjectHierarchyNode();
   if (!shNode)
     {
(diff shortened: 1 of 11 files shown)


Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyComboBox.cxx
```diff
@@ -469,6 +471,60 @@ void qMRMLSubjectHierarchyComboBox::setAlignPopupVertically(bool align)
   d->AlignPopupVertically = align;
 }
 
+//--------------------------------------------------------------------------
+bool qMRMLSubjectHierarchyComboBox::noneItemEnabled()const
```
lassoan, line 475: It would be better to have the same API as qMRMLNodeComboBox. So, the method name should be `noneEnabled`.

Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyComboBox.cxx
```diff
@@ -469,6 +471,60 @@ void qMRMLSubjectHierarchyComboBox::setAlignPopupVertically(bool align)
...
+{
+  if (!this->model())
+    {
+    return false;
+    }
+  return this->model()->noneItemEnabled();
+}
+
+//--------------------------------------------------------------------------
+void qMRMLSubjectHierarchyComboBox::setNoneItemEnabled(bool enable)
```
lassoan, line 485: `setNoneEnabled`

Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyComboBox.cxx
```diff
@@ -469,6 +471,60 @@ void qMRMLSubjectHierarchyComboBox::setAlignPopupVertically(bool align)
...
+{
+  if (!this->model())
+    {
+    return;
+    }
+  this->model()->setNoneItemEnabled(enable);
+}
+
+//--------------------------------------------------------------------------
+QString qMRMLSubjectHierarchyComboBox::noneItemText()const
```
lassoan, line 495: It would be better to mirror qMRMLNodeComboBox API -> rename to `noneDisplay`.

Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyComboBox.cxx
```diff
@@ -469,6 +471,60 @@ void qMRMLSubjectHierarchyComboBox::setAlignPopupVertically(bool align)
...
+{
+  if (!this->model())
+    {
+    return QString();
+    }
+  return this->model()->noneItemText();
+}
+
+//--------------------------------------------------------------------------
+void qMRMLSubjectHierarchyComboBox::setNoneItemText(const QString& displayName)
```
lassoan, line 505: It would be better to mirror qMRMLNodeComboBox API -> rename to `setNoneDisplay`.

Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyComboBox.h
```diff
@@ -56,6 +56,17 @@ class Q_SLICER_MODULE_SUBJECTHIERARCHY_WIDGETS_EXPORT qMRMLSubjectHierarchyCombo
...
+  /// the scene item for indicating 'None' selection.
+  Q_PROPERTY(bool noneItemEnabled READ noneItemEnabled WRITE setNoneItemEnabled)
+  /// This property controls the name that is displayed for the None item.
+  /// "None" by default.
+  /// \sa noneItemEnabled
+  Q_PROPERTY(QString noneItemText READ noneItemText WRITE setNoneItemText)
+  /// This property controls whether hierarchy information is included in the title text.
+  /// If enabled (which is the default), then the title typically looks like "[ParentName] / [ParentName] / [SelectedItemName]".
+  /// If disabled, the title will only be "[SelectedItemName]".
+  Q_PROPERTY(bool showParentsInTitle READ showParentsInTitle WRITE setShowParentsInTitle)
```
lassoan, line 69: It is not really a title. In a Qt combo-box it is called the "lineEdit", which might not the most intuitive name either, but at least not we invented the name just use an existing name.

lassoan (approved):
It looks good, I've added just a few small comments. It would be nice if you could add at least some minimal test, which would also server as an example.

Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyComboBox.cxx
cpinter: First I followed these function names but then I thought that although the `noneEnabled` and `noneDisplay` are unambiguous in a combobox, they are (especially `noneDisplay`) not very expressive of what they does in a subject hierarchy model or tree view. Because I had to introduce the same properties in these two other classes as well. I'll wait for your answer before changing these.

Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyComboBox.h
cpinter: I just tried to follow the convention I found. See `updateComboBoxTitleAndIcon` function. At least it is definitely not a non-existent name.

cpinter:
Thanks for the review @lassoan ! I added a test, but had some comments about your renaming requests so I'll wait until we settle on those before I push.

Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyComboBox.cxx
lassoan: I don't particularly like the existing names but it is important to make this new combobox a drop-in replacement of vtkMRMLNodeComboBox. If you want, you can add alternative names (different methods all calling the same underlying method).

Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyComboBox.h
lassoan: `updateComboBoxTitleAndIcon` is not in the public API, so it was not important there to have a meaningful or consistent name. I don't have much better suggestions, so if you don't like any of the ones that I listed below any better then I'm OK with the current name, too.

showCurrentItemParents
showCurrentItemParentsInLineEdit
showParentsInLineEdit

lassoan:
Now that you are modifying this code (add a top-level item), would you be able to add back the top-level "Scene" item? It would be really useful to be able to access scene-level features (that now requires clicking in a "white" area in the SH tree that even people who know what to look for have often trouble finding).

cpinter:
> add back the top-level "Scene" item

Do you mean by default to the SH tree in the Data module? Yes the white area issue is quite annoying. I think we removed it back then to save valuable horizontal space, but I don't see a problem re-adding it.

Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyComboBox.cxx
cpinter: That's a good point about trivial switch from qMRMLNodeComboBox. I changed these properties and functions back.

Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyComboBox.h
cpinter: I renamed this to `showCurrentItemParents`. I think LineEdit is even worse than Title because it cannot be edited in this case.

cpinter:
I changed everything we discussed. Once we agree on the scene item issue I'll add a commit in this PR for it.

lassoan:
> > add back the top-level "Scene" item
> 
> I probably wouldn't change the default not to modify all the SH trees used in all the modules/extensions/etc.
> 
> However, the white area issue is quite annoying. I think we removed it back then to save valuable horizontal space, but I don't see a problem re-adding it to the Data module (Maybe the DICOM module as well? Maybe not because it is mainly intended for seeing the scene content)

You are right, it makes sense to keep the current behavior by default (to not impact existing modules that might not expect the change), but we can enable it explicitly when we know we have enough space.

cpinter:
I re-added the scene item only in the Data module. I reduced the margins from 9/9 to 4/4 within the tab widget tabs (as already the case in other modules where space is important). It looks like this:
image

@lassoan Please approve if it looks good to you. Thanks!

cpinter:
(Note: I think back when we removed the scene item we had the transform column show the transform names instead of the icons so horizontal space looked very restricted)

lassoan (approved):
Looks good, thank you very much.

PR #5643, opened 2021-05-13
Markups measurements fixes

lassoan:


Libs/MRML/Core/vtkMRMLMeasurement.cxx
```diff
@@ -353,10 +406,25 @@ const char* vtkMRMLMeasurement::GetLastComputationResultAsPrintableString()
...
   bool modified = false;
+  if (units)
+    {
+    if (this->Units != units)
+      {
+      this->Units = units;
+      modified = true;
+      }
+    double displayCoefficient = 1.0;
+    if (this->DisplayCoefficient != displayCoefficient)
```
cpinter, line 421: This is I think is a bit hard to read, i.e. this if looks like `displayCoefficient` should come as argument that is set. Instead I'd rename the local variable to `defaultDisplayCoefficient`, or even make a const member with this name, or just use the 1.0 scalar in the if.

Libs/MRML/Core/vtkMRMLMeasurement.cxx
```diff
@@ -408,26 +490,51 @@ void vtkMRMLMeasurement::SetValue(double value, vtkMRMLUnitNode* unitNode, int l
...
+
+//---------------------------------------------------------------------------
+vtkMRMLUnitNode* vtkMRMLMeasurement::GetUnitNode(const char* quantityName)
+{
+  if (!quantityName || strlen(quantityName)==0)
+    {
+    return nullptr;
     }
-  if (lastComputationResult != ComputationResult::NoChange)
+  if (!this->InputMRMLNode || !this->InputMRMLNode->GetScene())
```
cpinter, line 518: I think it is a valid scenario not having an input node for constant measurements.

Libs/MRML/Core/vtkMRMLMeasurement.h
```diff
@@ -153,26 +195,27 @@ class VTK_MRML_EXPORT vtkMRMLMeasurement : public vtkObject
...
-  vtkCodedEntry* UnitsCode{nullptr};      // cubic millimeter
-  vtkCodedEntry* MethodCode{nullptr};     // Sum of segmented voxel volumes
+  vtkSmartPointer<vtkCodedEntry> QuantityCode;   // volume
+  vtkSmartPointer<vtkCodedEntry> DerivationCode; // min/max/mean
+  vtkSmartPointer<vtkCodedEntry> UnitsCode;      // cubic millimeter
+  vtkSmartPointer<vtkCodedEntry> MethodCode;     // Sum of segmented voxel volumes
   ComputationResult LastComputationResult{InsufficientInput};
 
-  /// Per-control point measurements
+  /// Per-control point measurements.
```
cpinter, line 215: Is this a style requirement?

cpinter (dismissed):
Thank you @lassoan! I just added a few minor comments / questions.

Libs/MRML/Core/vtkMRMLMeasurement.cxx
lassoan: We need a reference to the scene to get the unit node from the scene. I'll add documentation to the method that unit can only be retrieved if an input node is set (it should be always possible to find a node that can be associated with a measurement) and a warning message if the input node is not set.

Libs/MRML/Core/vtkMRMLMeasurement.cxx
lassoan: Good point. I've updated the code to let the user specify the display coefficient along with the unit and use the current display coefficient as default.

Libs/MRML/Core/vtkMRMLMeasurement.h
lassoan: I don't think it is a strong requirement, but when there is no "." then the doxygen rendering looks like the text is trimmed or accidentally has not finished.

I'm not sure if somebody was just interrupted while writing the documentation and left it unfinished or that's the complete documentation:

image

It is clear that this is all the documentation:

image

lassoan:
I've addressed all your comments, let me know if you have any remaining findings.

cpinter (approved):
Looks good, thank you very much!

PR #5646, opened 2021-05-17
DOC: add an apt update step to linux build lines

pieper:
Make sure that the package lists are updated before installing to avoid
errors on fresh installs (e.g. a fresh VM on a gcp's Ubuntu image
will not have updated packages by default).


lassoan (approved):
LGTM

PR #5648, opened 2021-05-19
DOC: fix debian 10 install instructions

pieper:
Thanks @ Issakomi 👍 

See https://discourse.slicer.org/t/simple-installation-of-3d-slicer-source-code-ubuntu-20-0/17662/8


lassoan (approved):
I haven't tried this, but it looks good to me

Issue #5649, opened 2021-05-19
Build error using smart pointer and vtkSetGet macros in debug mode


pieper:
Build error due to ambiguous overload

## Steps to reproduce

Mac build in debug mode

## Expected behavior

Should build fine, but instead get build errors like this:

```
[ 18%] Building CXX object Libs/MRML/Core/CMakeFiles/MRMLCore.dir/vtkMRMLMeasurement.cxx.o
In file included from /Users/pieper/slicer/latest/Slicer/Libs/MRML/Core/vtkMRMLMeasurement.cxx:10:
/Users/pieper/slicer/latest/Slicer/Libs/MRML/Core/vtkMRMLMeasurement.h:158:3: error: use of overloaded operator '<<' is ambiguous (with operand types 'vtkOStreamWrapper'
      and 'vtkSmartPointer<vtkCodedEntry>')
  vtkGetObjectMacro(QuantityCode, vtkCodedEntry);
  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/opt/s/VTK/Common/Core/vtkSetGet.h:238:54: note: expanded from macro 'vtkGetObjectMacro'
                << "): returning " #name " address " << this->name );   \
                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~
/opt/s/VTK/Common/Core/vtkSetGet.h:467:33: note: expanded from macro 'vtkDebugMacro'
   vtkDebugWithObjectMacro(this,x)
   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~
/opt/s/VTK/Common/Core/vtkSetGet.h:576:18: note: expanded from macro 'vtkDebugWithObjectMacro'
    vtkmsg << "" x << "\n\n";                                                    \
    ~~~~~~~~~~~~~^
/opt/s/VTK/Common/Core/vtkOStreamWrapper.h:71:22: note: candidate function
  vtkOStreamWrapper& operator << (const vtkSmartPointerBase&);
                     ^
/opt/s/VTK/Common/Core/vtkOStreamWrapper.h:74:22: note: candidate function
  vtkOStreamWrapper& operator << (void*);
                     ^
/opt/s/VTK/Common/Core/vtkOStreamWrapper.h:87:22: note: candidate function
  vtkOStreamWrapper& operator << (bool);
                     ^
[...]

PR #5652, opened 2021-05-21
COMP: Fix build error on Windows

cpinter:
Errors like this occurred when compiling vtkMRMLMeasurement on Windows (using VTK8):
C:\d\Slicer\Libs\MRML\Core\vtkMRMLMeasurement.h(158,3): error C2666: 'vtkOStreamWrapper::operator <<': 5 overloads have similar conversions [C:\d\S4D\Slicer-build\Libs\MRML\Core\MRMLCore.vcxproj] [C:\d\S4D\Slicer.vcxproj]

They were due to using vtkGetObjectMacro for vtkSmartPointer objects.

### DIFF
--- a/Libs/MRML/Core/vtkMRMLMeasurement.cxx
+++ b/Libs/MRML/Core/vtkMRMLMeasurement.cxx
@@ -556,3 +556,33 @@ vtkMRMLUnitNode* vtkMRMLMeasurement::GetUnitNode(const char* quantityName)
   // that case hundreds of warnings would be thrown in a non erroneous situation.
   return unitNode;
 }
+
+//----------------------------------------------------------------------------
+vtkCodedEntry* vtkMRMLMeasurement::GetQuantityCode()
+{
+  return this->QuantityCode;
+}
+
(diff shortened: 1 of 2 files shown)


jamesobutler:
This should resolve #5649?

pieper (approved):
Yes, it looks like this will fix the vtk8 builds.  We need to keep this compatible until we can finish deprecating vtk8 after the next release.

cpinter:
After reading the issue @jamesobutler linked I wonder if we should backport the VTK fix instead. This is a very simple fix though (and the same trivial Get functions can be found for smart pointers elsewhere in the code), so not sure if it's worth the trouble especially that we'll phase out VTK8 soon (hopefully after fixing all the still maintained extensions that are broken with VTK9).

jcfr:

If we would like to keep using the macro, applying the patch referenced here should do it:
https://github.com/Slicer/Slicer/issues/5649#issuecomment-844634361

cpinter:
I merged this because it seems that nobody has a strong objection against this, and will fix VTK8 build.

PR #5654, opened 2021-05-25
Bug Fix: Added retry logic to allow DICOM transfer of DICOM-SEG files

robertsoakes:
The default configuration provided with the DCMTK `dicomscu` program used to send data from Slicer does not allow the transfer of DICOM-SEG. This patch provides: 

* an alternative configuration that includes the needed presentation context UID needed for DICOM-SEG files to be sent using DICOM and DICOMweb (using `dicomscu`)
* retry logic that first attempts to transfer the DCM instance using the standard configuration and (if that fails) will attempt a second transfer using the fallback configuration
* modifies the module `CMakeLists` to include the alternative config as a resource file


pieper (approved):
This looks reasonable to me.  Thanks for the contribution 👍

Modules/Scripted/DICOMLib/DICOMProcesses.py
```diff
@@ -344,9 +350,10 @@ def processStdoutLine(self, line):
 
 
 class DICOMSender(DICOMProcess):
-  """Code to send files to a remote host
-  (Uses storescu from dcmtk)
+  """ Code to send files to a remote host.
+      (Uses storescu from dcmtk.)
   """
+  dicomseg_altconfig_path = 'DICOM/dcmtk/storescu.cfg'
```
lassoan, line 356: Give a more meaningful name (a name that reflects why this alternative configuration exists).
Also, make it more future-proof by either using a more generic name (that allows us to add more customizations)  or by storing a list of alternative configuration filenames.

Either of these could work:

```python
extended_dicom_config_path = 'DICOM/dcmtk/storescu.cfg'
```

```python
alternative_dicom_config_paths = ['DICOM/dcmtk/storescu.cfg']
```

Modules/Scripted/DICOMLib/DICOMProcesses.py
```diff
@@ -436,21 +443,50 @@ def send(self):
         if not self.progressCallback(f"Sent {file} to {self.destinationUrl.host()}:{self.destinationUrl.port()}"):
           raise UserWarning("Sending was cancelled, upload is incomplete.")
 
-  def start(self,file):
+  def start(self, file, config=None, config_profile='Default', retry=False):
```
lassoan, line 446: Adding a Bool flag and introducing recursion would make things complicated and fragile. Instead, it would be much cleaner if you factored out the parts that you want to re-run into a separate method and call that method with various parameters from the `start` method.

jamesobutler (commented):
The required CommitCheck check is failing to pass and preventing direct merging because of various merge commits in the history for this branch. @robertsoakes Are you able to resolve this?

I don't use DICOM functionality in Slicer at all so @lassoan will be able to provide better review than myself.

Modules/Scripted/DICOMLib/Resources/DICOM/dcmtk/storescu.cfg
```diff
@@ -0,0 +1,62 @@
+#
+#  Copyright (C) 2003-2021, 3DSlicer
+#
+#  Author:  Rob Oakes
+#
+#  Purpose: Storescu configuration optimized for DICOM-SR and DICOM-SEG files
```
lassoan, line 6: The file name is very generic and in contrast the purpose is very specific. We could either add support for multiple configuration files (each with very name and content very specific for a certain use case) or add just one generic configuration file (with a generic name, which would contain all the customizations that we want to make in Slicer). I think the latter is somewhat simpler and we can switch to using a list of configuration files if we find that we cannot fulfill all use cases with a single configuration file.

Modules/Scripted/DICOMLib/Resources/DICOM/dcmtk/storescu.cfg
```diff
@@ -0,0 +1,62 @@
...
+#TransferSyntax4 = MPEG4HighProfile/Level4.2For3DVideo
+#TransferSyntax5 = MPEG4StereoHighProfile/Level4.2
+
+# ============================================================================
+[[PresentationContexts]]
+# ============================================================================
+
+[StorageCompressedAndUncompressed]
+#
+# For DICOM-SR and DICOM-SEg SOP classes, only uncompressed transmission is supported.
```
lassoan, line 50: I don't think SEG is a DICOM term (SR is). In the DICOM standard it is referred to as `Segmentation`.

Modules/Scripted/DICOMLib/Resources/DICOM/dcmtk/storescu.cfg
```diff
@@ -0,0 +1,62 @@
...
+#
+#
+# Segmentation transfer syntaxes
+#
+PresentationContext1 = 1.2.840.10008.5.1.4.1.1.66.4\Uncompressed
+
+# ============================================================================
+[[Profiles]]
```
lassoan, line 58: Add a link to the original, unmodified DCMTK configuration and mark very clearly which lines were added/modified compared to the original and why. This is important because time to time we will have to update the configuration from DCMTK and re-apply all the modifications manually.

lassoan (commented):
Thank you, this will be useful.

robertsoakes:
@jamesobutler Thanks for the heads up. I'll see if I can get it squared with the `master` branch via a rebase so that it's cleaner and the commit titles are happy. I did the work on a Windows machine and needed a branch that had not yet been merged. (It now has.)

Modules/Scripted/DICOMLib/Resources/DICOM/dcmtk/storescu.cfg
robertsoakes: @lassoan  Thank you for the comments and feedback. I'll incorporate your suggestions and push a new branch tomorrow.

lassoan: It would be great if we could finalize this pull request while our memory is still fresh and there are no merge conflicts.

robertsoakes: > Add a link to the original, unmodified DCMTK configuration and mark very clearly which lines were added/modified compared to the original and why. This is important because time to time we will have to update the configuration from DCMTK and re-apply all the modifications manually.

The configuration file, `storescu.cfg`, provides the set of DICOM transfer options that are sent as part of the DICOM presentation context. This includes what types of rules and compression options are supported and the storage modalities that are presented to the PACS.

By far the most important component of the file is the list of presentation contexts found under the `PresentationContexts.StorageCompressedAndUncompressed`. I'm not sure if it's part of the DICOM standard or for some other reason, but storescu chose to hard code 128 values covering some of the most common storage modalities.

To make the alternative configuration I provided, I grabbed a copy of the DCMTK config that was included as part of the development source, removed nearly all of the presentation contexts and added the modalities for the DICOM seg files I needed. My thinking was that as additional modalities are discovered that are part of 3DSlicer's core feature set they can just be added by creating additional context values. Examples:

```ini
...
PresentationContext2 = SegmentationStorage
PresentationContext3 = SegmentedVolumeRenderingVolumetricPresentationStateStorage
PresentationContext4 = SurfaceScanPointCloudStorage
PresentationContext5 = SurfaceSegmentationStorage
...
```

My question here: because it is not controlled by Slicer, is it better to include a copy of the original config in our sources, or to clearly document where it came from and the extensions needed?

robertsoakes: > It would be great if we could finalize this pull request while our memory is still fresh and there are no merge conflicts.

Sorry for the delay, I ended up getting pulled away because of some needs on a separate project. I'm hoping to get back to this cleaned up tomorrow or Tuesday.

Modules/Scripted/DICOMLib/DICOMProcesses.py
robertsoakes: I will add more documentation to the alternative configuration describing where it came from and how it can be extended. A single configuration file should be enough (hopefully) to cover all of the specialty storages covered by slicer. I'll also adopt the property name you propose in this comment.

Modules/Scripted/DICOMLib/Resources/DICOM/dcmtk/storescu.cfg
robertsoakes: I think you're right. The standard refers to `Segmentation` templates for SR documents, and DICOM-SEG is technically a special type of Structured Report. Because of their role within many image processing applications, though, many programs refer to them separately from SR, though.

You can find references, for example, to DICOM-SEG in the documentation from libraries such as [PyDicom Seg](https://github.com/razorx89/pydicom-seg), [highdicom](https://github.com/razorx89/pydicom-seg), and OHIF.

Modules/Scripted/DICOMLib/Resources/DICOM/dcmtk/storescu.cfg
robertsoakes: Agreed. I'll adopt this approach into the merge request when I revise it tomorrow.

Modules/Scripted/DICOMLib/Resources/DICOM/dcmtk/storescu.cfg
lassoan: I'm just curious, is there a formal specification of a SR template called DICOM-SEG?

robertsoakes: No, I don't think there is. You are right, the template seems to be "Segmentation."

There are some places where `SEG` is used distinctively to differentiate segmentations from more general structured reports or annotations. [`SEG` is also used to specify the modality of the file](http://dicom.nema.org/medical/Dicom/2018d/output/chtml/part03/sect_C.8.20.html) in comparison to [`SR` for general structured report instances or annotations](https://dicom.innolitics.com/ciods/cr-image/general-series/00080060) or [`REG` for spatial registration files](http://dicom.nema.org/dicom/2013/output/chtml/part03/sect_C.20.html).

The practice of calling the resources DICOM-SEG, DICOM-REG, etc seems to pretty common, at least amongst the software development communities where I am active. (Maybe taken from the modality?) It also appears in semi-formal resources, such as the writings and [presentations of David Clunie](https://cdn.ymaws.com/siim.org/resource/resmgr/mimi18/presentations/18cmimi_ml-clunie.pdf), so I take it as an "informal semi-canonical" thing.

Modules/Scripted/DICOMLib/Resources/DICOM/dcmtk/storescu.cfg
robertsoakes: The file has been renamed to `storescu-seg.cfg`, and the description has been greatly expanded to provide information about why it is needed, how to extend, and provide the URL to the source file in the DCMTK source code (GitHub).

Modules/Scripted/DICOMLib/DICOMProcesses.py
robertsoakes: The method has been refactored to remove the recursion. The transfer logic, method that waits until the tx ends, and process status have been added to a helper method `dicomSend` that is called twice: once with the default configuration and a second time with the alternative configuration.

Modules/Scripted/DICOMLib/Resources/DICOM/dcmtk/storescu.cfg
robertsoakes: > It would be great if we could finalize this pull request while our memory is still fresh and there are no merge conflicts.

I've pulled the most recent changes from the `master` branch, which hopefully should resolve any potential merge conflicts. Please let me know if you would like me to restructure the PR.

robertsoakes: Everything should be ready for a second review.

lassoan:
FYI, I'm reviewing this and will test and merge today.

lassoan:
Thanks a lot for your contribution and addressing review comments. The commits have been squashed and merged in https://github.com/Slicer/Slicer/commit/21f8ba46beb67ae169387dd94237b0d7f8087302.

PR #5656, opened 2021-05-26
Extensions manager tweaks

jcfr:
Selection of commits ready for integration that were initially associated with https://github.com/Slicer/Slicer/pull/5637


jcfr:
Merging. Changes have been locally tested.

PR #5657, opened 2021-05-27
ENH: Add additional saved parameters to vtkMRMLMarkupsPlaneNode

Sunderlandkyl:
Adds missing attributes to markups plane node XML Read/Write/Copy/Print, as well as the new vtkMRMLMarkupsPlaneJsonStorageNode class that can save additional plane data in the json file.

PlaneBounds (renamed to Size), and AutoSizeScalingFactor attributes in vtkMRMLMarkupsPlaneNode were not included in WriteXML/ReadXMLAttributes/CopyContent/PrintSelf functions. This commit adds the relevant attributes to the functions.

### DIFF
--- a/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsPlaneJsonStorageNode.cxx
+++ b/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsPlaneJsonStorageNode.cxx
@@ -1,0 +1,246 @@
+/*==============================================================================
+
+  Program: 3D Slicer
+
+  Portions (c) Copyright Brigham and Women's Hospital (BWH) All Rights Reserved.
+
+  See COPYRIGHT.txt
+  or http://www.slicer.org/copyright/copyright.txt for details.
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+
+  This file was originally developed by Kyle Sunderland, PerkLab, Queen's University
+  and was supported through CANARIE's Research Software Program, Cancer
+  Care Ontario, OpenAnatomy, and Brigham and Women's Hospital through NIH grant R01MH112748.
+
+==============================================================================*/
+
+#include <vtkCodedEntry.h>
+#include "vtkMRMLMarkupsPlaneJsonStorageNode.h"
+#include "vtkMRMLMarkupsDisplayNode.h"
+#include "vtkMRMLMarkupsPlaneNode.h"
+
+#include "vtkMRMLMessageCollection.h"
+#include "vtkMRMLScene.h"
+#include "vtkSlicerVersionConfigure.h"
+
+#include "vtkDoubleArray.h"
+#include "vtkObjectFactory.h"
+#include "vtkStringArray.h"
+#include <vtksys/SystemTools.hxx>
+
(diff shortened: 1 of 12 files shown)


lassoan (commented):
This looks great, thank you! Please update the schema (https://github.com/Slicer/Slicer/blob/master/Modules/Loadable/Markups/Resources/Schema). Since we add new fields, I think we should increase the version to v1.0.2.

lassoan:
> Ok, sounds good. Should the plane/ROI specific fields be prefixed by plane*/roi*?

Probably it results in simpler, more readable code if we do not use prefixes. For example, `markups['myPlane']['origin'] ` is nicer than `markups['myPlane']['planeOrigin'] `. Also, if we don't use prefixes then the same `markup['origin']` code can be used to get the center of the markup, regardless of it is a plane or ROI.

Sunderlandkyl:
Ok, added the v1.0.2 schema.

Modules/Loadable/Markups/Resources/Schema/markups-schema-v1.0.2.json
```diff
@@ -0,0 +1,618 @@
...
+                            "description": "Stores a single markup.",
+                            "default": {},
+                            "required": ["type"],
+                            "additionalProperties": true,
+                            "properties": {
+                                "type": {
+                                    "$id": "#markup/type",
+                                    "type": "string",
+                                    "title": "Basic type",
+                                    "enum": ["Fiducial", "Line", "Angle", "Curve", "ClosedCurve", "Plane"]
```
lassoan, line 38: We should add ROI. (here and maybe elsewhere in the schema, too)

Modules/Loadable/Markups/Resources/Schema/markups-schema-v1.0.2.json
```diff
@@ -0,0 +1,618 @@
...
+                                    "additionalItems": false,
+                                    "items": { "type": "number" },
+                                    "minItems": 16,
+                                    "maxItems": 16
+                                },
+                                "orientation": {
+                                    "$id": "#markup/orientation",
+                                    "type": "array",
+                                    "title": "Markups orientation",
+                                    "description": "3x3 orientation matrix of the markups representation in node coordinates.",
```
lassoan, line 163: to clarify this, add something like "(orientation[0], orientation[3], orientation[6]) is the x vector of the ... coordinate system in the ... coordinate system"

lassoan (commented):
Thanks a lot, these are very important fixes and improvements. I've added a few comments inline.

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsPlaneJsonStorageNode.cxx
```diff
@@ -0,0 +1,247 @@
...
+  double zAxis_Node[3] = { 0.0, 0.0, 0.0 };
+  planeNode->GetAxes(xAxis_Node, yAxis_Node, zAxis_Node);
+  double origin_Node[3] = { 0.0, 0.0, 0.0 };
+  planeNode->GetOrigin(origin_Node);
+  for (int i = 0; i < 3; ++i)
+    {
+    orientation[3 * i] = xAxis_Node[i];
+    orientation[3 * i + 1] = yAxis_Node[i];
+    orientation[3 * i + 2] = zAxis_Node[i];
+    orientation[3 * i + 3] = origin_Node[i];
```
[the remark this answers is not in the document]
Sunderlandkyl: Thanks for catching this. It shouldn't contain translation.

Sunderlandkyl:
Thanks! Changes made.

lassoan (approved):

PR #5660, opened 2021-05-28
Fix plugin factory issues

cpinter:


lassoan (approved):
Thank you, this looks good! Much nicer, cleaner than before.

I've checked all the extensions and qSlicerSegmentEditorEffectFactory is not used anywhere, so the API change will not cause any issue there.

cpinter:
Thank you! Merging then.

PR #5662, opened 2021-05-30
ENH: Add automatically generated CLI module documentation

lassoan:
This is a work-in-progress pull request, which I share to get feedback before going on with making changes in all the affected files.

Switch to MyST markdown parser (https://myst-parser.readthedocs.io/) from recommonmark for two reasons:
1. Sphinx has deprecated recommonmark and recommends MyST now. See https://github.com/readthedocs/recommonmark/issues/221.
2. MyST allows accessing all reStructuredText features with markdown syntax! This is truly amazing, because we can use the simple and intuitive markdown syntax that everyone knows and still access all rst features. It adds essential features that are missing from markdown, such as including other markdown (or rST) files, linking to subsections, etc.

Generate CLI module documentation from module descriptor XML files automatically, during documentation setup. Generating the documentation for all modules takes a few seconds. The generated markdown file can be included in a manually created markdown file, which makes it easy to add custom content and cross-reference content from other documentation pages.

Move all documentation from rst to markdown:
- this makes the entire documentation use a single markup syntax and the same that is used in the forum, wiki, readme files, etc.
- code often become corrupted when added to the script repository because rST required adding indentation, which - depending on text editor settings - messed up indentation of Python code
- GitHub can preview markdown formatting


cpinter:
This is great, thanks for working on this!

lassoan:
This is ready for review now!

Almost all modules are now documented in readthedocs! See https://slicer--5662.org.readthedocs.build/en/5662/user_guide/modules/index.html

Only about 15-20 are missing, this task is tracked in https://github.com/Slicer/Slicer/issues/5664.

Also moved all documentation to markdown with myST markdown. Confirmed that *all* rST features that we used in Slicer documentation works well using this markdown style (but without the extremely fragile and cumbersome rST syntax).

Docs/_extracli/BRAINSDWICleanup.xml
```diff
@@ -0,0 +1,44 @@
+<?xml version="1.0" encoding="utf-8"?>
```
jcfr, line 1: For each remote module let's add a comment documenting which XML descriptions should be updated. For example, for `BRAINSTools`, such comment would be added here:

https://github.com/Slicer/Slicer/blob/7cce89a83f86ea9d542bae7b090912da422ce64b/SuperBuild.cmake#L314-L315

Including in the comment the set of bash command download the files would be ideal, something like this would do it for `BRAINSTools`. Similar comment would be added for DMRI, ... :

```bash
cd Slicer
for moduleName in \
  BRAINSDWICleanup \
  BRAINSDeface \
  BRAINSFit \
  BRAINSLabelStats \
  BRAINSROIAuto \
  BRAINSResample \
  BRAINSResize \
  BRAINSStripRotation \
  BRAINSTransformConvert;
do
  echo "Downloading ${moduleName}/${moduleName}.xml"
  curl -L# https://raw.githubusercontent.com/BRAINSia/BRAINSTools/590c371cecfc9e6f4dfed6dcc6416f67bccdaaf6/${moduleName}/${moduleName}.xml -o Docs/_extracli/${moduleName}.xml;
done
```

jcfr: Or (1) a script called `Utilities/Scripts/UpdateDocsExtraCLIDescription.sh` could be added  ... this script would extract the relevant `GIT_TAG`/`GIT_REPOSITORY` info from `SuperBuild.cmake`, and (2) add a comment to `SuperBuild.cmake` indicating this script should be executed.

That said this may be more error prone as it would rely on the exact syntax in the CMake file and be fragile ...

Docs/conf.py
```diff
@@ -36,12 +35,20 @@
...
-    'recommonmark',
+    'myst_parser',
     'sphinx_markdown_tables',
-    'notfound.extension',  # Show a better 404 page when an invlid address is entered
+    'notfound.extension',  # Show a better 404 page when an invalid address is entered
+]
+
+myst_enable_extensions = [
+    "colon_fence",  # Allow code fence using ::: (see https://myst-parser.readthedocs.io/en/latest/using/syntax-optional.html#syntax-colon-fence)
+    "linkify",  # Allow automCreate links from URLs (it is sufficient to write https://google.com instead of <https://google.com>)
```
jcfr, line 45:
```suggestion
    "linkify",  # Allow automatic creation of links from URLs (it is sufficient to write https://google.com instead of <https://google.com>)
```

Docs/conf.py
```diff
@@ -207,3 +213,69 @@
...
+# These md files are included at the top and bottom of each CLI module
+# documentation file (user_guide\modules\*.md). Custom content, such as
+# tutorials, screenshots, etc. can be added in between these includes.
+#
+# A copy of all CLI module descriptor XML files for modules that are
+# not part of the Slicer repository (e.g., BRAINS toolkit) are stored in
+# _extracli subfolder. Content of this folder must be updated manually
+# whenever they are updated in the original location. The process could
+# be automated with some effort, but since these bundled libraries do
+# not change frequently, manual update takes less work overall.
```
jcfr, line 234: > re: automation [...]  do not change frequently, manual update takes less work overall

Agreed. I suggest to add a comment to `SuperBuild.cmake` with the bash  command to run. See my previous suggestion.

Docs/conf.py
```diff
@@ -207,3 +213,69 @@
...
+excludenames = [
+    'CLIROITest.xml',
+    'TestGridTransformRegistration.xml',
+    'DiffusionTensorTest.xml',
+    ]
+
+# Output folder that contains all generated markdown files.
+outpath = os.path.join(docsfolder, "_moduledescriptions")
+os.makedirs(outpath, exist_ok=True)
+descriptionfile = open(os.path.join(outpath, '_readme_.txt'), 'w')
```
jcfr, line 257:
```suggestion
with open(os.path.join(outpath, '_readme_.txt'), 'w') as descriptionfile:
```

Docs/conf.py
```diff
@@ -207,3 +213,69 @@
...
+descriptionfile.write("Content of this folder is automatically generated by Docs/conf.py from CLI module descriptor XML files")
+descriptionfile.write("during building of documentation. The files are automatically regenerated when needed.")
+descriptionfile.close()
```
jcfr, line 260:
```suggestion
```

Docs/conf.py
```diff
@@ -207,3 +213,69 @@
...
+    'CLIROITest.xml',
+    'TestGridTransformRegistration.xml',
+    'DiffusionTensorTest.xml',
+    ]
+
+# Output folder that contains all generated markdown files.
+outpath = os.path.join(docsfolder, "_moduledescriptions")
+os.makedirs(outpath, exist_ok=True)
+descriptionfile = open(os.path.join(outpath, '_readme_.txt'), 'w')
+descriptionfile.write("Content of this folder is automatically generated by Docs/conf.py from CLI module descriptor XML files")
```
jcfr, line 258:
```suggestion
    descriptionfile.write("Content of this folder is automatically generated by Docs/conf.py from CLI module descriptor XML files")
```

Docs/conf.py
```diff
@@ -207,3 +213,69 @@
...
+descriptionfile.write("during building of documentation. The files are automatically regenerated when needed.")
```
jcfr, line 259:
```suggestion
    descriptionfile.write("during building of documentation. The files are automatically regenerated when needed.")
```

Docs/conf.py
```diff
@@ -207,3 +213,69 @@
...
+descriptionfile.close()
+
+def _generatemd(dom, docsfolder, outpath, xslt, suffix):
+    """Helper function to create markdown file from CLI module description XML file using XSLT"""
+    xsltpath = os.path.join(docsfolder, xslt)
+    transform = ET.XSLT(ET.parse(xsltpath))
+    infile = str(transform(dom))
+    outfile = open(os.path.join(outpath, os.path.splitext(name)[0]+suffix+'.md'), 'w', encoding='utf8')
```
jcfr, line 267:
```suggestion
        with open(os.path.join(outpath, os.path.splitext(name)[0]+suffix+'.md'), 'w', encoding='utf8') as outfile:
```

Docs/conf.py
```diff
@@ -207,3 +213,69 @@
...
+descriptionfile = open(os.path.join(outpath, '_readme_.txt'), 'w')
+descriptionfile.write("Content of this folder is automatically generated by Docs/conf.py from CLI module descriptor XML files")
+descriptionfile.write("during building of documentation. The files are automatically regenerated when needed.")
+descriptionfile.close()
+
+def _generatemd(dom, docsfolder, outpath, xslt, suffix):
+    """Helper function to create markdown file from CLI module description XML file using XSLT"""
+    xsltpath = os.path.join(docsfolder, xslt)
+    transform = ET.XSLT(ET.parse(xsltpath))
+    infile = str(transform(dom))
```
jcfr, line 266:
```suggestion
    content = str(transform(dom))
```

jcfr (approved):
Before of few nitpicks, it looks great.

This is a great improvement ! 

I really like this: https://myst-parser.readthedocs.io/en/latest/using/syntax.html#syntax-directives

jcfr:
I have no problem with a follow up commit addressing my comments, there are minor and can addressed at a later time. And I wouldn't mind taking care of it ...

Docs/_extracli/BRAINSDWICleanup.xml
lassoan: We should have a solution that scales well for documenting all the extensions, not just those that are bundled with Slicer and that works for all module types, not just CLI. This would allow us to include documentation of _all_ extensions in the Slicer documentation, the same way as for bundled extensions. This would result in a documentation that is similar to the old wiki, where all the extensions are documented, not only Slicer core modules. Building the documentation should be a lightweight process and so it should not include installing or building Slicer - so that it can be easily handled by readthedocs.

This could be achieved for example by having all necessary information in the extension description files (in json). There could be a link to a markdown file for the extension and a markdown file for each module (or the markdown file content could be embedded in the json file) and for CLI modules there could be a link to an additional XML file.

Until we have a good design for this, I think we should go minimal - and do all updates manually. Since there is very small development activity in BRAINS and other bundled extensions, in particular in CLI modules, and especially at such level that would result in documentation changes, I would expect that we will not need to do any updates for 1-2 years (and by then we should have the more generic extension documentation infrastructure implemented).

Docs/conf.py
lassoan: I would rather focus on implementing a good final solution rather than making the current workaround a bit more convenient.

jcfr: > I would rather focus on implementing a good final solution rather than making the current workaround a bit more convenient.

Make sense. Then, let's create an issue to track the suggestion about json description file you described above ?

Docs/_extracli/BRAINSDWICleanup.xml
jcfr: > I think we should go minimal - and do all updates manually

Agreed, in the mean time what about adding a comment like this near relevant remote extensions ? 

```
# Consider updating CLI description files found in Docs/_extracli directory
```

Docs/conf.py
lassoan: OK!

Docs/_extracli/BRAINSDWICleanup.xml
lassoan: Good idea, I'll add these comments.

Docs/conf.py
jcfr: >  let's create an issue to track the suggestion about json description file you described above ?

See https://github.com/Slicer/Slicer/issues/5669

Docs/conf.py
```diff
@@ -207,3 +213,69 @@
...
+    'DiffusionTensorTest.xml',
+    ]
+
+# Output folder that contains all generated markdown files.
+outpath = os.path.join(docsfolder, "_moduledescriptions")
+os.makedirs(outpath, exist_ok=True)
+with open(os.path.join(outpath, '_readme_.txt'), 'w') as descriptionfile:
+    descriptionfile.write("Content of this folder is automatically generated by Docs/conf.py from CLI module descriptor XML files")
+    descriptionfile.write("during building of documentation. The files are automatically regenerated when needed.")
+descriptionfile.close()
```
jcfr, line 260:
```suggestion
```

jcfr: This is not needed anymore

Docs/conf.py
```diff
@@ -207,3 +213,69 @@
...
+
+def _generatemd(dom, docsfolder, outpath, xslt, suffix):
+    """Helper function to create markdown file from CLI module description XML file using XSLT"""
+    xsltpath = os.path.join(docsfolder, xslt)
+    transform = ET.XSLT(ET.parse(xsltpath))
+    content = str(transform(dom))
+        with open(os.path.join(outpath, os.path.splitext(name)[0]+suffix+'.md'), 'w', encoding='utf8') as outfile:
```
jcfr, line 267:
```suggestion
    with open(os.path.join(outpath, os.path.splitext(name)[0]+suffix+'.md'), 'w', encoding='utf8') as outfile:
```

Docs/conf.py
```diff
@@ -207,3 +213,69 @@
...
+    outfile.write(infile)
```
jcfr, line 268:
```suggestion
        outfile.write(content)
```

Docs/conf.py
```diff
@@ -207,3 +213,69 @@
...
+    outfile.close()
```
jcfr, line 269:
```suggestion
```

jcfr (approved):

lassoan:
@jcfr thanks for the thorough review and great suggestions.

PR #5666, opened 2021-06-01
COMP: use ITK to build ANTs

simonoxen:
Hi, I'm working on a [ANTs extension](https://github.com/simonoxen/SlicerANTs) and adding these modules to ITK build allows me to build ANTs against Slicer's ITK. They are used in [ANTs ITK Super Build](https://github.com/ANTsX/ANTs/blob/master/SuperBuild/External_ITKv5.cmake).

Would it be possible to include?


lassoan (commented):
Adding the denoising and interpolator should be no problem. 

DCMTK is built outside ITK and passed to it from outside. If enabling the DCMTK module does not download or build DCMTK again then it should be OK, too. Can you check if ITK still uses the common DCMTK that the Slicer superbuild provides?

simonoxen:
Indeed the DCMTK was not used. I Removed it.

lassoan (approved):
Thank you, this looks good to me. I'll just do a test build and if there are no other comments then I'll merge later today.

PR #5668, opened 2021-06-01
ENH: Allow multiple extensions to be excluded from line length check

Sunderlandkyl:
Expands the regex so that files with multiple extensions can still match the regex if the tail of the string contains an excluded extension, instead of requiring a complete match.


lassoan (approved):
Looks great, thank you!

PR #5670, opened 2021-06-02
Update user_interface.md

alfredomp:
With @katiamar63 we added more details to the calculation of the auto W/L to include "percentiles" as in the current description is suggests to use top and bottom 0.1 of the entire range.
We based this modification on this codes:
- https://github.com/Slicer/Slicer/blob/0094e9a35bd266cc8b0e677858dabce797c9928f/Libs/MRML/Core/vtkMRMLScalarVolumeDisplayNode.cxx#L748
- https://github.com/Kitware/VTK/blob/f56f68786f5d1120a73343801208c031dae749e4/Imaging/Statistics/vtkImageHistogramStatistics.cxx#L67

### DIFF
--- a/Docs/user_guide/user_interface.md
+++ b/Docs/user_guide/user_interface.md
@@ -99,7 +99,7 @@ The toolbar icons that switch between these mouse modes are shown from left to r
 
 Medical images typically contain thousands of gray levels, but regular computer displays can display only 256 gray levels, and the human eye also has limitation in what minimum contrast difference it can notice (see [Kimpe 2007](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3043920/) for more specific information). Therefore, medical images are displayed with adjustable brightness/contrast (window/level).
 
-By default 3D Slicer uses window/level setting that is specified in the DICOM file. If it is not available then window/level is set so that the entire intensity range of the image (except top/bottom 0.1%, to not let a very thin tail of the intensity distribution to decrease the image contrast too much).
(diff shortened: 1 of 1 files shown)


Docs/user_guide/user_interface.md
```diff
@@ -99,7 +99,7 @@ The toolbar icons that switch between these mouse modes are shown from left to r
 
 Medical images typically contain thousands of gray levels, but regular computer displays can display only 256 gray levels, and the human eye also has limitation in what minimum contrast difference it can notice (see [Kimpe 2007](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3043920/) for more specific information). Therefore, medical images are displayed with adjustable brightness/contrast (window/level).
 
-By default 3D Slicer uses window/level setting that is specified in the DICOM file. If it is not available then window/level is set so that the entire intensity range of the image (except top/bottom 0.1%, to not let a very thin tail of the intensity distribution to decrease the image contrast too much).
+By default 3D Slicer uses window/level setting that is specified in the DICOM file. If it is not available then window/level is set so that the entire intensity range of the image (except top/bottom 0.1%, calculated using percentiles, to not let a very thin tail of the intensity distribution to decrease the image contrast too much).
```
lassoan, line 102:
```suggestion
By default 3D Slicer uses window/level setting that is specified in the DICOM file. If it is not available then window/level is set to contain the entire intensity range of the image (except top/bottom 0.1%, calculated using percentiles, to not let a very thin tail of the intensity distribution to decrease the image contrast too much).
```

lassoan:
Thank you!

PR #5674, opened 2021-06-04
ENH: Improve directory and file selection by using native dialogs

jamesobutler:
This aims to improve directory selection by using the native dialog instead of Qt's own widget-based dialog design.

See 3D Slicer forum discussion of how users have been doing other methods to avoid using widgets that use Qt's widget-based dialog.
https://discourse.slicer.org/t/using-native-file-dialog-for-selecting-files-and-folders/17926

| Qt widget-based dialog | Native dialog (on Windows)|
|-------------------------|--------------------------------|
|image|image|

This PR currently contains changes to update the main Slicer Load dialog to choose directories and files using a native dialog. I have not made any changes to the ctkDICOMBrowser dialog tied to import buttons used in the DICOM module as it had some custom widgets into the widget-based dialog which wouldn't be possible when using a native dialog. @lassoan Do you have ideas about what you might want to do here? Would be willing to contribute changes as I don't use DICOM functionality at all so I'm not familiar with the usage in Slicer.

Some limitations of using a native dialog is that you can use Qt functionality such as QTimers and such to manipulate it. It is a modal dialog when opened. However navigation of directories and files is much easier using a native dialog and the current Qt widget based dialog often slows down this navigation process leading others to use drag-and-drop from a native dialog to avoid it entirely.

Will require https://github.com/commontk/CTK/pull/973 and if merged, the upstream and new git tag would be updated here in this PR.

### DIFF
--- a/SuperBuild/External_CTK.cmake
+++ b/SuperBuild/External_CTK.cmake
@@ -70,7 +70,7 @@ if(NOT DEFINED CTK_DIR AND NOT Slicer_USE_SYSTEM_${proj})
 
   ExternalProject_SetIfNotDefined(
     Slicer_${proj}_GIT_TAG
-    "01ab373d01e6e8718af42f856278b61bb18fbcfe"
+    "9f12861724155f9929de9be8500c9eabd38507ac"
     QUIET
     )
 
--- a/Base/QTGUI/Resources/UI/qSlicerSaveDataDialog.ui
+++ b/Base/QTGUI/Resources/UI/qSlicerSaveDataDialog.ui
@@ -53,9 +53,6 @@
      <property name="text">
       <string>Change directory for selected files</string>
      </property>
-     <property name="options">
-      <set>ctkDirectoryButton::DontUseNativeDialog</set>
-     </property>
     </widget>
    </item>
    <item row="0" column="6">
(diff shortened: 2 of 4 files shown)


SuperBuild/External_CTK.cmake
```diff
@@ -64,13 +64,13 @@ if(NOT DEFINED CTK_DIR AND NOT Slicer_USE_SYSTEM_${proj})
 
   ExternalProject_SetIfNotDefined(
     Slicer_${proj}_GIT_REPOSITORY
-    "${EP_GIT_PROTOCOL}://github.com/commontk/CTK.git"
+    "${EP_GIT_PROTOCOL}://github.com/jamesobutler/CTK.git"
```
lassoan, line 67:
```suggestion
    "${EP_GIT_PROTOCOL}://github.com/commontk/CTK.git"
```

SuperBuild/External_CTK.cmake
```diff
@@ -64,13 +64,13 @@ if(NOT DEFINED CTK_DIR AND NOT Slicer_USE_SYSTEM_${proj})
...
     QUIET
     )
 
   ExternalProject_SetIfNotDefined(
     Slicer_${proj}_GIT_TAG
-    "01ab373d01e6e8718af42f856278b61bb18fbcfe"
+    "dicom-browser-import-dialog"
```
lassoan, line 73:
```suggestion
    "c4f1eeead234022e74de978126e3a54f648880e5"
```

lassoan (commented):
Just need to update CTK as suggested in the inline comments and it should be ready to go.

SuperBuild/External_CTK.cmake
jamesobutler: See https://github.com/commontk/CTK/pull/977 before this can happen

jcfr: > See commontk/CTK#977 before this can happen

This has just been merged. Thanks for the reminder.

Issue #5651, opened 2021-05-20
qMRMLWindowLevelWidget set manual min/max not working.


fbordignon:
`setAutoWindowLevel(qMRMLWindowLevelWidget::ManualMinMax)`

If in auto mode, changes to Manual W/L instead of manual min max

## Steps to reproduce
Python console:
```
a = slicer.qMRMLWindowLevelWidget()
b = getNode("MyNode")
a.setMRMLVolumeNode(b)
a.show()
a.setAutoWindowLevel(slicer.qMRMLWindowLevelWidget.ManualMinMax)
```

## Expected behavior

Should change to manual min/max?

## Environment
- Slicer version: Slicer-4.11
- Operating system: Windows 10

lassoan:
Good catch. The editing mode changes correctly to ManualMinMax (label are set to Min/Max and moving the end of a slider changes the minimum value not the window), but the AutoManualComboBox selection is not updated. It would be great if you could fix this.

![image](https://user-images.githubusercontent.com/307929/119001054-a536ec00-b959-11eb-8f5b-d2bcb3408f11.png)

fbordignon:
I will try to contribute on this, maybe just update the combobox.

PR #5676, opened 2021-06-04
Improving qMRMLVolumeWidget range bounds functionality

jamesobutler:
This PR aims to improve range bounds behavior for widgets that use qMRMLVolumeWidget including the Window Level and Threshold widgets. See https://discourse.slicer.org/t/improving-qmrmlvolumewidget-range-bounds/17821 for full discussion about this work.

I have added a "..." toolbutton next to the sliders in the window/level and threshold UI files that is similar in nature to what is used in the Slicer qMRMLRangeWidget. 

Shown in the image below is what things look like now. Pressing the "..." toolbutton will show/hide the popup widget contain widgets for setting the bounds of the main slider widget. The previous hover action that would trigger the popup widget was very frustrating to users as in most cases they were looking to change bounds. This toolbutton now makes changing bounds a deliberate action.

A second commit here also sets the threshold slider bounds based on the scalar range of the set MRML volume node. Thresholding outside of the scalar range does not make sense and does nothing. When things do nothing this often confuses users. Users can still change the bounds using the "..." toolbutton in the threshold widget.

image

### DIFF
--- a/Libs/MRML/Core/vtkMRMLScalarVolumeDisplayNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLScalarVolumeDisplayNode.cxx
@@ -53,8 +53,6 @@ vtkMRMLScalarVolumeDisplayNode::vtkMRMLScalarVolumeDisplayNode()
   this->AutoWindowLevel = 1;
   this->AutoThreshold = 0;
   this->ApplyThreshold = 0;
-  //this->LowerThreshold = VTK_SHORT_MIN;
-  //this->UpperThreshold = VTK_SHORT_MAX;
 
   // try setting a default grayscale color map
   //this->SetDefaultColorMap(0);
@@ -163,6 +161,7 @@ void vtkMRMLScalarVolumeDisplayNode::SetInputImageDataConnection(vtkAlgorithmOut
     vtkEventBroker::GetInstance()->AddObservation(
       inputImageDataAlgorithm, vtkCommand::ModifiedEvent, this, this->MRMLCallbackCommand );
     }
+
 }
 
 //----------------------------------------------------------------------------
--- a/Libs/MRML/Widgets/Resources/UI/qMRMLVolumeThresholdWidget.ui
+++ b/Libs/MRML/Widgets/Resources/UI/qMRMLVolumeThresholdWidget.ui
@@ -50,16 +50,34 @@
     </widget>
    </item>
    <item row="1" column="0" colspan="2">
-    <widget class="ctkRangeWidget" name="VolumeThresholdRangeWidget"/>
+    <layout class="QHBoxLayout" name="horizontalLayout">
+     <item>
+      <widget class="ctkRangeWidget" name="VolumeThresholdRangeWidget"/>
+     </item>
+     <item>
+      <widget class="QToolButton" name="RangeButton">
+       <property name="text">
+        <string>...</string>
+       </property>
+       <property name="icon">
+        <iconset>
+         <normaloff>:Icons/SliceMoreOptions.png</normaloff>:Icons/SliceMoreOptions.png</iconset>
+       </property>
+       <property name="checkable">
+        <bool>true</bool>
+       </property>
+      </widget>
+     </item>
+    </layout>
    </item>
   </layout>
  </widget>
  <customwidgets>
-   <customwidget>
-     <class>ctkRangeWidget</class>
-     <extends>QWidget</extends>
-     <header>ctkRangeWidget.h</header>
-   </customwidget>
+  <customwidget>
+   <class>ctkRangeWidget</class>
+   <extends>QWidget</extends>
+   <header>ctkRangeWidget.h</header>
+  </customwidget>
  </customwidgets>
  <resources/>
  <connections/>
--- a/Libs/MRML/Widgets/qMRMLVolumeThresholdWidget.cxx
+++ b/Libs/MRML/Widgets/qMRMLVolumeThresholdWidget.cxx
@@ -16,6 +16,7 @@
 ==============================================================================*/
 
 // qMRML includes
+#include "qMRMLSpinBox.h"
 #include "qMRMLVolumeThresholdWidget.h"
 #include "qMRMLVolumeWidget_p.h"
 #include "ui_qMRMLVolumeThresholdWidget.h"
@@ -249,11 +250,34 @@ void qMRMLVolumeThresholdWidget::setMaximum(double max)
 }
 
 // --------------------------------------------------------------------------
+void qMRMLVolumeThresholdWidget::updateWidgetFromMRMLVolumeNode()
+{
+  Q_D(qMRMLVolumeThresholdWidget);
+  this->Superclass::updateWidgetFromMRMLVolumeNode();
+
+  // When scalar range changes, maintain current threshold values, but update range
+  const double min = d->VolumeDisplayNode->GetLowerThreshold();
+  const double max = d->VolumeDisplayNode->GetUpperThreshold();
+
+  double range[2];
+  d->scalarRange(d->VolumeDisplayNode, range);
+  d->DisplayScalarRange[0] = range[0];
+  d->DisplayScalarRange[1] = range[1];
+  const double minRangeValue = std::min(min, range[0]);
+  const double maxRangeValue = std::max(max, range[1]);
+  d->setRange(minRangeValue, maxRangeValue);
+
+}
+
+// --------------------------------------------------------------------------
 void qMRMLVolumeThresholdWidget::updateWidgetFromMRMLDisplayNode()
 {
   Q_D(qMRMLVolumeThresholdWidget);
 
-  this->Superclass::updateWidgetFromMRMLDisplayNode();
+  // Don't want to call qMRMLVolumeWidget::updateWidgetFromMRMLDisplayNode which would call updateRangeForVolumeDisplayNode
+
+  this->setEnabled(d->VolumeDisplayNode != nullptr &&
+                   d->VolumeNode != nullptr);
 
   if (!d->VolumeDisplayNode)
     {
@@ -275,6 +299,10 @@ void qMRMLVolumeThresholdWidget::updateWidgetFromMRMLDisplayNode()
 
   const double min = d->VolumeDisplayNode->GetLowerThreshold();
   const double max = d->VolumeDisplayNode->GetUpperThreshold();
+  const double minRangeValue = std::min(min, d->MinRangeSpinBox->value());
+  const double maxRangeValue = std::max(max, d->MaxRangeSpinBox->value());
+  d->setRange(minRangeValue, maxRangeValue);
+
   d->VolumeThresholdRangeWidget->setValues(min, max );
 
   d->VolumeThresholdRangeWidget->blockSignals(wasBlocking);
--- a/Libs/MRML/Widgets/qMRMLVolumeThresholdWidget.cxx
+++ b/Libs/MRML/Widgets/qMRMLVolumeThresholdWidget.cxx
@@ -15,10 +15,6 @@
 
 ==============================================================================*/
 
-// CTK includes
-#include <ctkDoubleSpinBox.h>
-#include <ctkPopupWidget.h>
-
 // qMRML includes
 #include "qMRMLVolumeThresholdWidget.h"
 #include "qMRMLVolumeWidget_p.h"
@@ -72,6 +68,10 @@ void qMRMLVolumeThresholdWidgetPrivate::init()
 
   this->connect(this->AutoManualComboBox, SIGNAL(currentIndexChanged(int)),
                 q, SLOT(setAutoThreshold(int)));
+
+  this->RangeButton->setMenu(this->OptionsMenu);
+  this->RangeButton->setPopupMode(QToolButton::InstantPopup);
+
 }
 
 
@@ -143,17 +143,6 @@ void qMRMLVolumeThresholdWidget::setAutoThreshold(ControlMode autoThreshold)
       d->VolumeDisplayNode->SetApplyThreshold(0);
       }
     d->VolumeDisplayNode->EndModify(disabledModify);
-
-    if (autoThreshold == qMRMLVolumeThresholdWidget::Manual)
-      {
-      d->PopupWidget->setAutoShow(true);
-      d->PopupWidget->showPopup();
-      }
-    else
-      {
-      d->PopupWidget->setAutoShow(false);
-      d->PopupWidget->hidePopup();
-      }
 
     if (oldAuto != d->VolumeDisplayNode->GetAutoThreshold() ||
         oldApply != d->VolumeDisplayNode->GetApplyThreshold())
--- a/Libs/MRML/Widgets/qMRMLVolumeThresholdWidget.cxx
+++ b/Libs/MRML/Widgets/qMRMLVolumeThresholdWidget.cxx
@@ -127,30 +127,42 @@ void qMRMLVolumeThresholdWidget::setAutoThreshold(ControlMode autoThreshold)
 {
   Q_D(qMRMLVolumeThresholdWidget);
 
-  if (d->VolumeDisplayNode)
+  if (!d->VolumeDisplayNode)
     {
-    int oldAuto = d->VolumeDisplayNode->GetAutoThreshold();
-    int oldApply = d->VolumeDisplayNode->GetApplyThreshold();
+    return;
+    }
+  int oldAuto = d->VolumeDisplayNode->GetAutoThreshold();
+  int oldApply = d->VolumeDisplayNode->GetApplyThreshold();
 
-    int disabledModify = d->VolumeDisplayNode->StartModify();
-    if (autoThreshold != qMRMLVolumeThresholdWidget::Off)
-      {
-      d->VolumeDisplayNode->SetApplyThreshold(1);
-      d->VolumeDisplayNode->SetAutoThreshold(
-        autoThreshold == qMRMLVolumeThresholdWidget::Auto ? 1 : 0);
-      }
-    else
-      {
-      d->VolumeDisplayNode->SetApplyThreshold(0);
-      }
-    d->VolumeDisplayNode->EndModify(disabledModify);
+  int disabledModify = d->VolumeDisplayNode->StartModify();
+  if (autoThreshold == qMRMLVolumeThresholdWidget::Off)
+    {
+    d->VolumeDisplayNode->SetApplyThreshold(0);
+    }
+  else
+    {
+    d->VolumeDisplayNode->SetApplyThreshold(1);
+    d->VolumeDisplayNode->SetAutoThreshold(
+      autoThreshold == qMRMLVolumeThresholdWidget::Auto ? 1 : 0);
+    }
 
-    if (oldAuto != d->VolumeDisplayNode->GetAutoThreshold() ||
-        oldApply != d->VolumeDisplayNode->GetApplyThreshold())
-      {
-      emit this->autoThresholdValueChanged(autoThreshold);
-      }
+  if (!oldApply && autoThreshold == qMRMLVolumeThresholdWidget::Manual)
+    {
+    // Previously the threshold was turned off and now it is set to manual.
+    // Since the default threshold range is VTK_SHORT_MIN to VTK_SHORT_MAX,
+    // we don't want these values to appear on the GUI but instead set
+    // the threshold range to the full scalar range of the volume (because
+    // this corresponds to the previous state of the thresholding: having
+    // the full scalar range of the volume in the threshold range).
+    d->VolumeDisplayNode->SetThreshold(d->DisplayScalarRange[0], d->DisplayScalarRange[1]);
+    }
 
+  d->VolumeDisplayNode->EndModify(disabledModify);
+
+  if (oldAuto != d->VolumeDisplayNode->GetAutoThreshold() ||
+      oldApply != d->VolumeDisplayNode->GetApplyThreshold())
+    {
+    emit this->autoThresholdValueChanged(autoThreshold);
     }
 }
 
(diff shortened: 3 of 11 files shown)


Libs/MRML/Widgets/Resources/UI/qMRMLVolumeThresholdWidget.ui
```diff
@@ -49,17 +49,31 @@
      </item>
     </widget>
    </item>
-   <item row="1" column="0" colspan="2">
+   <item row="1" column="0" colspan="3">
     <widget class="ctkRangeWidget" name="VolumeThresholdRangeWidget"/>
    </item>
+   <item row="1" column="3">
+    <widget class="QToolButton" name="RangeButton">
```
lassoan, line 56: Please use the same button type and size constraints for the threshold `...` as for the W/L `...` (so that the `...` buttons appear the same size). See how they appear on my computer:

image

Libs/MRML/Widgets/qMRMLVolumeThresholdWidget.cxx
```diff
@@ -263,6 +264,10 @@ void qMRMLVolumeThresholdWidget::updateWidgetFromMRMLDisplayNode()
   // even if the values have been just set programmatically/automatically.
   bool wasBlocking = d->VolumeThresholdRangeWidget->blockSignals(true);
 
+  double range[2];
+  d->scalarRange(d->VolumeDisplayNode, range);
```
lassoan, line 268: Scalar range is cached in vtkImageData but the first time it is computed it is a relatively expensive operation (min/max operation on all voxel values). It can significantly slow down live volume display and volume sequence replay. Make sure that this method is only called if the Volumes module is active and Display section is open.

It is similar to how `qSlicerScalarVolumeDisplayWidget::updateHistogram()` is protected from being called all the time: it is only updated continuously when the histogram section is shown and not collapsed.

lassoan (commented):
Thank you, this is a great improvement! The new widgets are simple and effective.

We need to be very careful when touching anything around the volume display pipeline because during live volume display (e.g., for ultrasound image sequences) we need to update the volume dozens of times per second. Any unnecessary extra computation can decrease the fps. See details in inline comments.

Libs/MRML/Widgets/qMRMLVolumeThresholdWidget.cxx
```diff
@@ -263,6 +264,10 @@ void qMRMLVolumeThresholdWidget::updateWidgetFromMRMLDisplayNode()
...
+  d->setRange(range[0], range[1]);
```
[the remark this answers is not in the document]
jamesobutler: I switched between two volumes with different scalar ranges (MRHead and CTChest) and it worked well. So MRHead was 0 to 297 or something and then CTChest had a lower threshold that was a negative number and it switched to it appropriately.

> The range must be large enough to contain the current threshold value.

This brings up a point I hadn’t considered which is if the scalar range is say 0 to 297 and someone had set the lower threshold to -400 or something, then should the bounds of the threshold widget get set to -400 and 297? I guess so. However should users even be allowed to set a lower threshold below the scalar min value or an upper threshold above the scalar max value? Or should it clamp to this scalar range if they set it to a value outside of the range? Then the threshold range bounds widget would need to support changing bounds only within the scalar range (0 to 297, 50 to 200) and not support cases that go outside the range (-400 to 400).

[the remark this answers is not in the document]
lassoan: I added comment on exactly this (I hope it did not get lost). The GUI must not ever constrain what you can set in MRML. For example, you must allow setting threshold = 10 for an ultrasound image, and keep that, regardless of what the current image's scalar range is. If you replay an image sequence and happen to come across an all-black image, it would be terrible if it forced the threshold to be 0.

Note that I don't talk about switching between volume nodes, but changing vtkImageData (by sequence replay, updates via OpenIGTLink, etc).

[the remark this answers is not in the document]
jamesobutler: Using current Slicer:
```python
import SampleData
volumeNode = SampleData.SampleDataLogic().downloadMRHead()
volumeNode.GetDisplayNode().GetLowerThreshold()
# -32768.0
volumeNode.GetDisplayNode().GetUpperThreshold()
# 32768.0
volumeNode.GetImageData().GetScalarRange()
# (0.0, 279.0)
```
Above shows if I load the MRHead volume into the scene, it has a scalar range of 0 to 297, but the currently set lower threshold and upper threshold is no where near this. The MRML display node has values that are inconsistent with what is shown by the threshold widget. The threshold widget shows -600 to 600 though the actual values by the mrml node is -32768 to 32768. Why does this display volume have an initial lower and upper threshold that is nowhere near the scalar range?

Something would need to change before I change it to the threshold bounds so that they at least include the currently set threshold values as otherwise for the MRHead the widget would show bounds of -32768 to 32768.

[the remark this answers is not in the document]
lassoan: I guess there is some kind of initialization of the display node when you assign it to the volume node. There the threshold could be initialized to the scalar range. Since thresholding is disabled by default, it should not be a problem that it is tightly set to the scalar range.

Libs/MRML/Widgets/Resources/UI/qMRMLVolumeThresholdWidget.ui
jamesobutler: I have pushed an update to make these look the same. See example image below:
image

Libs/MRML/Widgets/qMRMLVolumeThresholdWidget.cxx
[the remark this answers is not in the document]
jamesobutler: Ok I'll look into this initialization so then the widget can truly reflect the currently set MRML threshold values.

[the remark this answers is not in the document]
jamesobutler: This is where the display node is being initialized to -32768 to 32768 and it not getting updated based on the scalar range when an volume node is connected.
https://github.com/Slicer/Slicer/blob/74d3ec4bb1996e73d82470195847a81703979488/Libs/MRML/Core/vtkMRMLScalarVolumeDisplayNode.cxx#L91

Libs/MRML/Widgets/qMRMLVolumeThresholdWidget.cxx
jamesobutler: I've updated the logic so that `scalarRange` is only called once as part of `updateRangeForVolumeDisplayNode` of the qMRMLVolumeWidget class. This is how it is currently in the Slicer `master` branch.

Libs/MRML/Widgets/qMRMLVolumeThresholdWidget.cxx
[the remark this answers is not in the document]
jamesobutler: @lassoan I've pushed an additional commit to initialize a scalar volume node with a threshold range that matches the scalar range instead of the current VTK_SHORT_MIN and VTK_SHORT_MAX values.

jamesobutler:
To handle the now-required "Check for spelling errors" check, I have rebased this PR against latest master to include the commit that added the github action.
image

jamesobutler:
@lassoan Could you provide review on this work again? I believe everything is working appropriately, specifically your previous point that if the threshold of the display node is outside the bounds of the current widget bounds, then the widget bound updates appropriately so that the widget is not constraining the actual node.

Libs/MRML/Core/vtkMRMLScalarVolumeDisplayNode.cxx
```diff
@@ -163,6 +162,10 @@ void vtkMRMLScalarVolumeDisplayNode::SetInputImageDataConnection(vtkAlgorithmOut
     vtkEventBroker::GetInstance()->AddObservation(
       inputImageDataAlgorithm, vtkCommand::ModifiedEvent, this, this->MRMLCallbackCommand );
     }
+
+  double range[2];
```
lassoan, line 166: Initialize to something (e.g., `{0.0, 0.0}` or `{0.0, 255.0}`) to ensure that you have a meaningful value even it GetDisplayScalarRange fails and returns without initializing.

Libs/MRML/Core/vtkMRMLScalarVolumeDisplayNode.cxx
```diff
@@ -163,6 +162,10 @@ void vtkMRMLScalarVolumeDisplayNode::SetInputImageDataConnection(vtkAlgorithmOut
...
+  this->GetDisplayScalarRange(range);
+  this->Threshold->ThresholdBetween(range[0], range[1]);
```
lassoan, line 168: Add a comment explaining why we initialize it to the scalar range by default.

Libs/MRML/Widgets/qMRMLVolumeThresholdWidget.cxx
```diff
@@ -60,10 +60,6 @@ void qMRMLVolumeThresholdWidgetPrivate::init()
 
   this->Superclass::init();
   this->setupUi(q);
-  this->VolumeThresholdRangeWidget->minimumSpinBox()->setDecimalsOption(
-    ctkDoubleSpinBox::DecimalsByKey|ctkDoubleSpinBox::DecimalsByShortcuts);
-  this->VolumeThresholdRangeWidget->maximumSpinBox()->setDecimalsOption(
-    ctkDoubleSpinBox::DecimalsByKey|ctkDoubleSpinBox::DecimalsByShortcuts);
```
lassoan, line 66: Why was this deleted? Is it now set in the .ui file?

Libs/MRML/Widgets/qMRMLVolumeThresholdWidget.cxx
```diff
@@ -286,7 +277,26 @@ void qMRMLVolumeThresholdWidget::updateWidgetFromMRMLDisplayNode()
...
+
   d->VolumeThresholdRangeWidget->setValues(min, max );
 
   d->VolumeThresholdRangeWidget->blockSignals(wasBlocking);
 }
+
+// --------------------------------------------------------------------------
+void qMRMLVolumeThresholdWidget::setPopupWidgetVisible(bool visible)
+{
+  Q_D(qMRMLVolumeThresholdWidget);
```
lassoan, line 293: don't we have a `showPopup(bool show)` method that we could call?

Libs/MRML/Widgets/qMRMLWindowLevelWidget.cxx
```diff
@@ -381,3 +393,17 @@ void qMRMLWindowLevelWidget::updateWidgetFromMRMLDisplayNode()
...
     }
 }
+
+// --------------------------------------------------------------------------
+void qMRMLWindowLevelWidget::setPopupWidgetVisible(bool visible)
+{
+  Q_D(qMRMLWindowLevelWidget);
+  if (visible)
+  {
+    d->PopupWidget->showPopup();
```
lassoan, line 403: use showPopup(bool) if available

lassoan (commented):
It looks good, I just added a few trivial comments that should be easy to address.

Libs/MRML/Widgets/qMRMLVolumeThresholdWidget.cxx
```diff
@@ -286,7 +277,26 @@ void qMRMLVolumeThresholdWidget::updateWidgetFromMRMLDisplayNode()
...
+  if (visible)
+  {
+    d->PopupWidget->showPopup();
+  }
+  else
+  {
+    d->PopupWidget->hidePopup();
+  }
```
jcfr, line 301:
```suggestion
  d->PopupWidget->showPopup(visible);
```

See https://github.com/commontk/CTK/blob/c4f1eeead234022e74de978126e3a54f648880e5/Libs/Widgets/ctkBasePopupWidget.h#L178

jamesobutler: I have tried the below, but haven't had success. What would be the syntax?
```c++
this->connect(this->RangeButton, SIGNAL(toggled(bool)),
                q, SLOT(this->PopupWidget->showPopup(bool)));
```
```c++
this->connect(this->RangeButton, SIGNAL(toggled(bool)),
                q, SLOT(q->PopupWidget->showPopup(bool)));
```

Libs/MRML/Widgets/qMRMLVolumeThresholdWidget.cxx
jamesobutler: I believe this was an accident when I first started working on this. I have restored this code.

Libs/MRML/Widgets/qMRMLVolumeThresholdWidget.cxx
jamesobutler: Ok I've simplified `setPopupWidgetVisible` with the `showPopup(bool)` method rather than trying to get rid of `setPopupWidgetVisible` completely.

lassoan: The syntax is object1, SIGNAL(method1), object2, SLOT(method2) - something like this:

    this->connect(this->RangeButton, SIGNAL(toggled(bool)),
                this->PopupWidget, SLOT(showPopup(bool)));

jamesobutler: Makes sense. Thanks! I confirmed it worked and I have pushed an update with this change.

lassoan (commented):
Thanks a lot. All the code changes look good and the behavior is much simpler and more intuitive!

However, as I tested it I ran into three quite significant glitches:

- The popup widgets remain visible even when navigating away from the button that bring up the widget. For example, if you click the `...` button and show the Python interactor then you get the image below. qMRMLRangeWidget does not have this issue.

image

- When typing a min or max value in the popup that is smaller than the current range then the value is reset to the previous valid value but the current range is set the typed value. For example, you load MRHead and try to delete the `6` from the `600` value then the top intensity range is immediately set to `0`. Either let change the slider range and current range when the value is typed (as it is done in qMRMLRangeWidget) or prevent changing the both the slider range and the current range.
- If the volume's scalar range changes then the min/max values in the popup are updated correctly, but the slider does not change (its min/max should be updated according to the new values). If you then make a small change in the min/max values in the popup then it completely messes up the window/level values (e.g., sets window to 0, level to 2000). To reproduce it: load MRHead sample, run the code below, and try to change the max value in the popup.

```python
vol=getNode('MRHead')
a=slicer.util.arrayFromVolume(vol)
a[:]=a[:]+1000
slicer.util.arrayFromVolumeModified(vol)
```

jamesobutler:
@lassoan I have changed things so that the min/max range spinboxes are no longer in a popup widget but rather in a QMenu like is done in qMRMLRangeWidget. This avoids the issue of the popup widget appearing over the python interactor even though the related widget is hidden underneath.
image

The popup widget appearing over the python interactor happens with the current widget in `master`, but not as bad since moving your mouse off the widget would hide it.
image


> - When typing a min or max value in the popup that is smaller than the current range then the value is reset to the previous valid value but the current range is set the typed value. For example, you load MRHead and try to delete the `6` from the `600` value then the top intensity range is immediately set to `0`. Either let change the slider range and current range when the value is typed (as it is done in qMRMLRangeWidget) or prevent changing the both the slider range and the current range.

I have changed the min/max range spinboxes to process on `editingFinished`. See if this handles what you were noting. I'm not sure I exactly follow the behavior you were trying to describe.

lassoan:
I like the menu-style popup, it is not just more robust but looks cleaner, simpler, too.

Videos for reproducing the two remaining problems:

When typing a min or max value in the popup that is smaller than the current range then the value is reset to the previous valid value but the current range is set the typed value:

https://user-images.githubusercontent.com/307929/122308312-22626c00-ceda-11eb-8b11-ad620e8b98da.mp4

If the volume's scalar range changes then the min/max values in the popup are updated correctly, but the slider does not change:

https://user-images.githubusercontent.com/307929/122308356-3d34e080-ceda-11eb-9a86-e6797821827f.mp4

jamesobutler:
@lassoan I've noticed that my `ENH: Initialize scalar volume node threshold to scalar range` sets an appropriate initial threshold based on the volume scalar range, but for some reason it has broken the update event. Setting the threshold value manually with code or with the widget does update the actual display even though it updates the threshold values.  Did I make changes in a wrong place? Without the commit, threshold works again, but still suffers from the display node not initializing with an appropriate initial threshold based on the associated volume node scalar range.

jamesobutler:
I pushed a workaround commit that reverts the previous commit that was trying to set a better initial threshold range of the vtkMRMLScalarVolumeDisplayNode. Instead of improving the initial threshold of the display node based on the image data that gets connected to it the workaround commit instead sets the display node threshold when the volume display node is set for the VolumeWidgets (qMRMLVolumeThresholdWidget/qMRMLWindowLevelWidget). I don't necessarily like this as if a volume node with scalar range say 0 to 300 but with a threshold set of -10 to 400 for some reason, it would become 0 to 300 upon setting the node for the widget. I'd rather the widget just reflect the state of the MRML node rather than the widget changing the MRML node threshold.

lassoan:
> I'd rather the widget just reflect the state of the MRML node rather than the widget changing the MRML node threshold.

Do you mean that the in the current version of the proposed implementation, the widget changes MRML node content even if the user did not move the slider? This must be changed, because if GUI widgets changed MRML nodes without the user interacting with the widget, it would lead to serious issues. Could you think about alternative designs? I'll look at the code tomorrow.

jamesobutler:
The commit `potential squash with previous, thresholding works again` as currently written is enforcing threshold to be the scalar range bounds. So if a volume threshold was set to -300 to 300, but the scalar range was 0 to 200, then the display node threshold values would get updates to 0 to 200 upon selecting the node for the widget.  

I was trying in `ENH: Initialize scalar volume node threshold to scalar range` to set the initial threshold values based on the scalar range when a volume node was linked to a volume display node, however this wasn't working appropriately. I was likely doing this in the wrong method (`SetInputImageDataConnection`) of vtkMRMLScalarVolumeDisplayNode?

What is happening in Slicer's `master` branch is that MRHead is initialized and has a threshold of VTK_SHORT_MIN to VTK_SHORT_MAX. However the threshold widget shows that it is -600 to 600 which is not what it is actually. I'm trying to fix this. Due to improvements in this PR, the threshold widget now appropriately shows that it is from VTK_SHORT_MIN to VTK_SHORT_MAX, however I'm wanting the vtkMRMLScalarVolumeNode to be initialized with a vtkMRMLScalarVolumeDisplayNode that has threshold initialized to the scalar range. So upon loading MRHead, the threshold widget would have a range of 0 to 297 because the scalar range of MRHead is 0 to 297 and the display node would have been initialized to set the threshold to 0 to 297.

jamesobutler:
@lassoan Have you had the chance to look at the latest in this PR yet?

lassoan:
I've been testing the file dialog PR. It all looks good, so I'll merge that and then get to this one.

jamesobutler:
Ok thanks so much!

jamesobutler:
@lassoan Do you think you will be able to take a look at this work before Project Week activities next week?

lassoan:
@jamesobutler I've pushed a commit that avoids modifying the MRML node by GUI widget without user interaction. Please test. If you find that it works well then I'll squash & merge.

jamesobutler:
@lassoan The threshold widget still doesn't appear to reflect the current threshold values of the mrml node. Upon loading the volume it has a threshold of VTK_SHORT_MIN to VTK_SHORT_MAX but the widget is not reflecting the real threshold value.
image

Then upon changing the threshold level from "Off" to "Manual" it is showing the real threshold value, but the volume display node was not initialized based on an appropriate range of the volume scalar range. It should be initialized to 0 to 297 here, but since the node is initialized to VTK_SHORT_MIN to VTK_SHORT_MAX, that is being reflected here.
image

lassoan:
When the user moves the slider then this works nicely. Indeed, when the user switches Threshold mode from "Off" to "Manual" then the automatic -32768 -> 32767 values get exposed, which is not ideal. We could either block this path (only allow switching to Manual by moving the slider) or have a way to distinguish manually set threshold range from full threshold range. For example, we could initialize the threshold to an invalid range to indicate that it has not been set yet.

jamesobutler:
I was trying in 239079eb26f22157532d0b1daa15f7cb398bc968  to initialize the Display Node with an appropriate initial threshold based on scalar range when a volume node’s image data is connected. Is there something possible in that logic? Maybe I wasn’t putting in the correct method. I was using SetInputDataConnection.

lassoan:
We cannot initialize the threshold range from the volume node for several reasons. For example, the result would depend on the order of setting display node, setting image data, and manually setting threshold range. We could add a method that would need to be called explicitly to initialize the threshold, but it would be hard to ensure that the method is called wherever a volume and display node is created and they are associated with each other. It would be also quite arbitrary to hardcode a threshold range based on the volume that is first associated with the display node. 

So, the only clear solution that I see is to leave the threshold range undefined by default. We just need to find a simple and intuitive way to transition from undefined threshold to a user-defined threshold.

If we losing the manually set threshold values when the user switches off the threshold then we could simply reset the default threshold when the user switches from Off to Manual threshold. If we want to keep the threshold when the user switches between different modes then we would either need to add a flag or use a special threshold range to indicate that it is undefined.

lassoan:
I've pushed a fix that addresses the `changing the threshold mode from "Off" to "Manual"` transition. Please test.

jamesobutler:
Ok your changes appear to work. From what I had before the threshold bounds would get forced to the scalar range upon setting the node to the threshold widget.  Your changes instead will set the threshold bounds to the scalar range upon turning on threshold with the widget. So if you set threshold to -300 to 300 for MRHead, and then set threshold to manual, then it sets the threshold to 0 to 279. I have squashed down the commits to accept your changes.

For historical purposes my changes had `setMRMLVolumeDisplayNode` in qMRMLVolumeWidget.cxx end with:
```c++
  // Initialize to the scalar range by default because it does not make sense to haver lower/upper threshold values
  // that are outside of the scalar bounds. It would only make a corresponding qMRMLVolumeThresholdWidget
  // have large bounds making the usable range of the slider smaller.
  double range[2];
  d->scalarRange(d->VolumeDisplayNode, range);
  d->VolumeDisplayNode->SetThreshold(range[0], range[1]);

  this->updateWidgetFromMRMLDisplayNode();
```

During some of this testing I have also observed confusing behavior with the threshold mode selection where setting a specific threshold doesn't put it into Manual mode like the Window/Level widget I believe does. Related to being able to set the threshold mode through code and having it reflect appropriately in the widget. I know Window/Level widget has similar weird behavior issues with setting mode and the widget not reflecting appropriately as discussed in #5651 and could be addressed in a separate PR.

lassoan (approved):
Thank you, it looks good. I'll merge this now. You can create bug reports or pull requests for any remaining issues or inconsistencies.

Thank you for all your hard work on this. These widgets works so much better now!

PR #5677, opened 2021-06-05
ENH: Add export/import of markups to/from table node

lassoan:
New "Export/import table" section is added to Markups module.

image


lassoan:
I'll merge this now so that users can give it a try and we can refine it later.

PR #5682, opened 2021-06-08
DOC: Fix variable name in slice orientation change example

ClaudioHoffmann:


lassoan (approved):
Good catch, thank you!

PR #5683, opened 2021-06-08
DOC: Add documentation page for Terminologies module

cpinter:


cpinter:
Still need to test the internal links.

Docs/user_guide/modules/terminologies.md
```diff
@@ -0,0 +1,50 @@
+# Terminologies
+
+## Overview
+
+The Terminologies module is a place where the loaded terminology and anatomical contexts can be explored and managed. The same widget (called terminology navigator) appears when editing the "color" of certain types of data nodes (such as models and markups) in the [Data](data.html) module as well as the structures in [Segment Editor](segmenteditor.html), and is used to set the identity and color of the nodes or segments in question.
+Default terminology contexts are the DICOM master list, which is a full list of terminology categories and types, and the 3D Slicer General Anatomy list, which is a subset of the master list, and the entries of which are associated with the labels in the GenericAnatomyColors color table (and the colors matched). Some terminology categories allow the selection of an anatomic region, which come from a different list. There is only one default anatomic context dictionary, which is also called the DICOM master list (but for anatomic contexts). Custom contexts can also be defined.
```
jamesobutler, line 6: Tried to rework some of the sentences here since it was a bit of a bear to read and understand.

```suggestion
Default terminology contexts are the DICOM master list, which is a full list of terminology categories and types, and the 3D Slicer General Anatomy list, which is a subset of the master list. The terminology entries are associated with the labels in the GenericAnatomyColors color table (and the colors matched). Some terminology categories allow the selection of an anatomical region, which come from a different list. There is only one default anatomical context dictionary, which is also called the DICOM master list. Custom contexts can also be defined.
```

Is there one or two? I was getting confused by how these two sentences relate.

> Default terminology contexts are the DICOM master list, which is a full list of terminology categories and types, and the 3D Slicer General Anatomy list, which is a subset of the master list.

> There is only one default anatomical context dictionary, which is also called the DICOM master list.

Docs/user_guide/modules/terminologies.md
```diff
@@ -0,0 +1,50 @@
...
+The contexts were created as part of the [QIICR](https://github.com/QIICR) project, in the [dcmqi](https://github.com/QIICR/dcmqi) toolkit.
+
+## Use cases
+
+Define terminology and anatomic region of a data node such as model or markup, or an segment in a segmentation, so that it can be identified unambiguously, and associated with standardized codes when exported to DICOM.
```
jamesobutler, line 11:
```suggestion
Define terminology and anatomical region of a data node such as model or markup, or a segment in a segmentation, so that it can be identified unambiguously, and associated with standardized codes when exported to DICOM.
```

Docs/user_guide/modules/terminologies.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Panels and their use
+
+In data trees and lists where the color of the objects are shown in a column as a color swatch, then double-clicking the color brings up the terminology selector dialog. These can be found for data nodes in the [Data](data.html), [Models](models.html), [Markups](markups.html) modules, and for the segments within segmentation nodes in the [Segmentations](segmentations.html) and [Segment Editor](segmenteditor.html) modules.
```
jamesobutler, line 15:
```suggestion
In data trees and lists where the color of the objects is shown in a column as a color swatch, then double-clicking the color brings up the terminology selector dialog. These can be found for data nodes in the [Data](data.html), [Models](models.html), [Markups](markups.html) modules, and for the segments within segmentation nodes in the [Segmentations](segmentations.html) and [Segment Editor](segmenteditor.html) modules.
```
the color _is_ shown

Docs/user_guide/modules/terminologies.md
```diff
@@ -0,0 +1,50 @@
...
+It shows the terminology types in all categories in a searchable list. Custom name and color can be assigned while keeping the selected terminology entry.
+
+Opening up the panes on the left and right, additional options become visible:
+![](https://github.com/Slicer/Slicer/releases/download/docs-resources/terminology_selector_advanced.png)
+
+- Left pane
+  - Category selection can be narrowed (all selected by default).
+  - On the top the terminology context can be changed, as well as new ones loaded from `.json` files.
+- Right pane
+  - Anatomic region can be selected. There are types where anatomic region selection is enabled. These can be types that can be located in multiple part of the body, such as "Blood clot", "Mass", or "Cyst".
```
jamesobutler, line 29:
```suggestion
  - Anatomical region can be selected. There are types where anatomical region selection is enabled. These can be types that can be located in multiple parts of the body, such as "Blood clot", "Mass", or "Cyst".
```

Docs/user_guide/modules/terminologies.md
```diff
@@ -0,0 +1,50 @@
...
+![](https://github.com/Slicer/Slicer/releases/download/docs-resources/terminology_selector_basic.png)
+
+It shows the terminology types in all categories in a searchable list. Custom name and color can be assigned while keeping the selected terminology entry.
+
+Opening up the panes on the left and right, additional options become visible:
+![](https://github.com/Slicer/Slicer/releases/download/docs-resources/terminology_selector_advanced.png)
+
+- Left pane
+  - Category selection can be narrowed (all selected by default).
+  - On the top the terminology context can be changed, as well as new ones loaded from `.json` files.
```
jamesobutler, line 27:
```suggestion
  - On the top, the terminology context can be changed, as well as new ones loaded from `.json` files.
```

Docs/user_guide/modules/terminologies.md
```diff
@@ -0,0 +1,50 @@
...
+  - Anatomic region can be selected. There are types where anatomic region selection is enabled. These can be types that can be located in multiple part of the body, such as "Blood clot", "Mass", or "Cyst".
+
+## How to
+
+- Find entries
+  - Start typing the name of the category/type/region in the search box above the column
+- Load new terminology/anatomic context
+  - Click the Load button next to the context drop-down and select JSON from local storage
+- Create custom terminology/anatomic context
+  - Start from an existing JSON file, such as the DICOM master list for [terminologies](https://github.com/Slicer/Slicer/blob/master/Modules/Loadable/Terminologies/Resources/SegmentationCategoryTypeModifier-DICOM-Master.json) or [anatomic contexts](https://github.com/Slicer/Slicer/blob/master/Modules/Loadable/Terminologies/Resources/AnatomicRegionAndModifier-DICOM-Master.json). Remove the entries you do not need. Validate the JSON file with the [validator online tool](http://qiicr.org/dcmqi/#/validators).
```
jamesobutler, line 38: I could be wrong, but I believe it should be anatomical over antatomic. I mentioned in other comments similar changes.

fedorov:
I am sorry, it is a very busy time for me right now. I will comment by the end of the week.

lassoan (dismissed):
Thank you @cpinter for writing this up and for @jamesobutler for the comments. I've incorporated @jamesobutler's comments, added some more information to the overview section, and fixed the links. It looks good to me now.

cpinter:
Throughout the codebase and the UI it the term "anatomic" is used (and not anatomical as changed by @jamesobutler). I originally took it from [the schema](https://github.com/Slicer/Slicer/blob/master/Modules/Loadable/Terminologies/Resources/AnatomicRegionAndModifier-DICOM-Master.json). I don't think it is a good idea to call things differently in the documentation and the UI, even if it sounds better that way. @fedorov do you have any comments about this?

jamesobutler:
Maybe there is a difference in English region that I’m not familiar with, but it is “anatomical position”, “anatomical region” and therefore appropriate as “anatomical context”. Searching some online dictionaries if you search “anatomic” it will take you to the entry “anatomical”.  

There is the word “atomic”, but an unrelated word here. Anyways, in general you would never say an “anatomic region” as it is an “anatomical region”.

If it was used incorrectly in the code, then I’m not sure what to say about how that should be changed then.

cpinter:
If you read my comment again I don't debate the correctness of the terms. However the terminology context schema and thus everything based on it uses the terms defined there for unambiguity. Changing "anatomic" to "anatomical" in the code should not be a big deal, but changing it in the schema will have repercussions. That's why I asked Andrey's opinion about it (not using mention this time to give him the peace he asked for due to his busy week).

jcfr:
Re: anatomic vs anatomical

What about the following:
- keep anatomical in the doc
- add clarifying sentence to mention anatomic is used in the code
- merge the change
- create an issue to track the fact, schema, ... must be updated
- address UI updates

lassoan:
Sounds like a good plan to me. The non-trivial part is that to avoid disruptions due to mismatch between context name in an existing segmentation and in Slicer, we would need to implement automatic search in all contexts if an item is not found in the context that is named in the file. The need for this feature has come up before, when we realized that we cannot include version number in the schema because of the current strict name match requirement, maybe we already have an open issue for it.

lassoan (approved):
I've added a note about anatomic/anatomical difference and a specific issue to make it consistent in the future (https://github.com/Slicer/Slicer/issues/5689). The documentation page is correctly generated and looks good, so I'll merge this now.

Docs/user_guide/modules/terminologies.md
```diff
@@ -0,0 +1,47 @@
+# Terminologies
+
+## Overview
+
+The Terminologies module provides a coding system for specifying anatomy, clinical findings, or other clinical terms. For example, the coding system can be used to specify what anatomical parts are stored in a segmentation node. Using standard codes reduces the chance of data-entry errors and improves interoperability. The coding system in Slicer is compatible with coding used in DICOM information objects and allows storing SNOMED-CT (Systematized Nomenclature of Human and Veterinary Medicine - Clinical Terms) or any other terminologies.
```
fedorov, line 5: I would rephrase the opening statement that the purpose of module was originally to support semantic annotation of segments in a segmentation node, which was subsequently applied to other types of annotations (if I understand correctly the following text - I have not used Terminologies with Markups myself). This is accomplished by utilizing standard coded terms that describe specific attributes of the segment: 1) general **category** of segment (e.g., ; 2) specific **type** of the segment within the broad category (e.g., ), and 3) where applicable, **anatomic region** where the segmented entity is located. Category and Type attributes can be accompanied by **modifiers** that provide more specificity in the situations where the value of the attribute corresponds to paired entities. Terminologies module follows the conventions for the attributes that accompany Segmentation objects as defined in the DICOM standard (see http://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.8.20.4.html#table_C.8.20-4). 

While each of the attributes above can be populated using codes from user-defined sources, and can come from different sources of coded terms, Terminologies module provides pre-populated pick-lists, called **contexts**, that contain terms extracted from the DICOM standard, which in turn relies on the SNOMED-CT ontology.

Docs/user_guide/modules/terminologies.md
```diff
@@ -0,0 +1,47 @@
...
+- Left pane
+  - Allows selection of a subset of categories (all categories selected by default).
+  - On the top, the terminology context can be changed, as well as new ones loaded from `.json` files.
+- Right pane
+  - Anatomical region can be selected for items that can be located in multiple parts of the body, such as "Blood clot", "Mass", or "Cyst".
+
+## How to
+
+- Find items: Start typing the name of the category/type/region in the search box above the column.
+- Load new terminology/anatomical context: click the Load button next to the context drop-down and select JSON from local storage.
```
fedorov, line 42: This may be a useful pointer for this item: https://qiicr.gitbook.io/dcmqi-guide/opening/coding_schemes

Docs/user_guide/modules/terminologies.md
```diff
@@ -0,0 +1,47 @@
...
+- Create custom terminology/anatomical context: Start from an existing JSON file, such as the DICOM master list for [terminologies](https://github.com/Slicer/Slicer/blob/master/Modules/Loadable/Terminologies/Resources/SegmentationCategoryTypeModifier-DICOM-Master.json) or [anatomical contexts](https://github.com/Slicer/Slicer/blob/master/Modules/Loadable/Terminologies/Resources/AnatomicRegionAndModifier-DICOM-Master.json). Remove the entries you do not need. Validate the JSON file with the [validator online tool](http://qiicr.org/dcmqi/#/validators).
+
+## References
+
+- [DCMQI wiki](https://github.com/QIICR/dcmqi/wiki)
```
fedorov, line 47: DCMQI user guide should be referenced instead: https://qiicr.gitbook.io/dcmqi-guide/.

Docs/user_guide/modules/terminologies.md
```diff
@@ -0,0 +1,47 @@
...
+
+Opening up the panes on the left and right, additional options become visible:
+
+![](https://github.com/Slicer/Slicer/releases/download/docs-resources/terminology_selector_advanced.png)
+
+- Left pane
+  - Allows selection of a subset of categories (all categories selected by default).
+  - On the top, the terminology context can be changed, as well as new ones loaded from `.json` files.
+- Right pane
+  - Anatomical region can be selected for items that can be located in multiple parts of the body, such as "Blood clot", "Mass", or "Cyst".
```
fedorov, line 37: I would say that Anatomic region is redundant when segmentation defines the entire organ or anatomic structure, but can be specified for segmentation types that identify an entity located within a certain region.

fedorov:
> Throughout the codebase and the UI it the term "anatomic" is used (and not anatomical as changed by @jamesobutler). I originally took it from the schema. I don't think it is a good idea to call things differently in the documentation and the UI, even if it sounds better that way. @fedorov do you have any comments about this?

I don't think it is critical, and I am fine with either one. FWIW, I think it is more important to be consistent between using "segment" and "segmentation" in this article, and I am not sure it is consistent currently (I thought Terminology-specified terms accompany segments, not segmentations directly?).

Docs/user_guide/modules/terminologies.md
cpinter: Can you please suggest an edit that we can simply paste? It would be much more efficient than me trying to understand your logic then do some iterations until it's fine with you. Thanks.

Docs/user_guide/modules/terminologies.md
cpinter: Again please suggest edit. Thank you.

Issue #5618, opened 2021-04-25
Markups rendering very slow, reading z-buffer and re-uploading data to GPU


pieper:
I created a scene with a 94 markups lines and it runs at only about 11 FPS.  Measured with Rock mode turned on, but feels the same when interacting.

## Steps to reproduce

[mrb for testing](https://www.dropbox.com/s/6h3076k6i981wa2/slow-markups-2021-04-25-Scene.mrb?dl=0)

## Expected behavior

It should be a very lightweight rendering operation, but some of the features appear to use extra resources.

## Environment
- Slicer version: Slicer-4.13-2021-04-12
- Operating system: Mac + big sur 11.2.3

## Other info

Profiling indicates a big hit is in `vtkSlicerLineRepresentation3D::RenderOverlay` which reads back the Zbuffer with `vtkSelectVisiblePoints` which basically breaks the GPU pipelining.

Also the vtk pipeline that gets triggered by `vtkOpenGLRenderer::UpdateGeometry` during the `vtkOpenGLRenderer::DeviceRender`. takes a chunk of time too and seems to be re-uploading data every frame, although that may be unavoidable.

Time sample file is [here](https://www.dropbox.com/s/d7w4f3c5el5vf5w/Sample%20of%20Slicer%204.13.0-2021-04-12.txt?dl=0).

Turning off the PropertyLabel improves performance to about 14 FPS.

If there is only one markup performance is 60 FPS.

If I put the same number of control points into a single `MarkupNode` there is no obvious slowdown.

Suggestion: we should find a way to only update `vtkSelectVisiblePoints` once per render frame.

pieper:
@Sunderlandkyl @lassoan

lassoan:
Yes, we know about this slowdown and I agree that z buffer should be only retrieved once per rendering. Maybe vtkSelectVisiblePoints could be improved to have a singleton cache that would store the last result for each render window on would only retrieve from the GPU again if needed.

lassoan:
We have run into a similar problem in SlicerHeart, too. It seems that markups cannot widely adopted until we this performance issue is fixed.

[...]

PR #5687, opened 2021-06-10
ENH: Improve markups rendering performance

Sunderlandkyl:
This PR contains 2 improvements to the performance of rendering markups fiducials and visibility detection of control points.

- Improve 3D markups fiducial performance with vtkGlyph3DMapper
  - vtkGlyph3D has a higher overhead when used for a large number of points, since it duplicates the glyph polydata once for each point, and then uploads the duplicated polydata to the mapper.
  - vtkGlyph3DMapper has a higher performance as it uploads the glyph once, and renders the same polydata multiple times on the GPU.

- Improve performance of markups visibility detection 
  - Previously, each ControlPointsPipeline3D had its own vtkSelectVisiblePoints filter to determine if the markups were occluded by opaque geometry in the render window. The filters would all download the ZBuffer data separately, resulting in a performance loss.
  - This commit adds the vtkFastSelectVisiblePoints filter which exposes the option to specify the Z-buffer pointer as a vtkFloatArray. As a result, we now only get the contents of the Z-buffer once, and re-use the same data for each control point pipeline.

Re #5618

### DIFF
--- a/Modules/Loadable/Markups/VTKWidgets/CMakeLists.txt
+++ b/Modules/Loadable/Markups/VTKWidgets/CMakeLists.txt
@@ -12,6 +12,8 @@ set(${KIT}_INCLUDE_DIRECTORIES
 set(${KIT}_SRCS
   vtk${MODULE_NAME}GlyphSource2D.cxx
   vtk${MODULE_NAME}GlyphSource2D.h
+  vtkFastSelectVisiblePoints.cxx
+  vtkFastSelectVisiblePoints.h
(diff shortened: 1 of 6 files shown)


lassoan (approved):
Thank you, this works very well!

200 line markups rendering speed increased from 1-2 fps to 7-8 fps!

We could consider moving vtkFastSelectVisiblePoints to vtkAddon, or even VTK proper once we finished with all performance optimizations, but for now we can keep things where they are now.

PR #5688, opened 2021-06-10
ENH: Add context menu when right-clicking in slice or 3D view

lassoan:
Currently, the only actions in the view context menu (when clicking on empty area) then only interaction mode can be changed.
More actions can be added later.

image

Let me know if there are other actions that you would like to see in this menu.

### DIFF
--- a/Modules/Loadable/SubjectHierarchy/Widgets/qSlicerSubjectHierarchyPluginLogic.cxx
+++ b/Modules/Loadable/SubjectHierarchy/Widgets/qSlicerSubjectHierarchyPluginLogic.cxx
@@ -35,6 +35,7 @@
 #include "qSlicerSubjectHierarchyRegisterPlugin.h"
 #include "qSlicerSubjectHierarchyFolderPlugin.h"
 #include "qSlicerSubjectHierarchyOpacityPlugin.h"
+#include "qSlicerSubjectHierarchyViewContextMenuPlugin.h"
 #include "qSlicerSubjectHierarchyVisibilityPlugin.h"
 
 // MRML includes
@@ -45,6 +46,7 @@
 
 // Slicer includes
 #include "qSlicerApplication.h"
+#include "vtkSlicerApplicationLogic.h"
 
 // Qt includes
 #include <QAction>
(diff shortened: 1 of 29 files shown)


pieper:
This looks great, thanks for working on it!

Some ideas for context menu items:
* Edit properties of background/foreground/label/segmentation node
* Edit Segment (use hovered segment like DataProbe)
* Show this view only / restore layout
* Change this view to {axial,sagittal,coronal,threeD,table...)
* Split view horizontally/vertically
* Copy image to clipboard
* Segment background
* Access view controller widget via right click (easier than moving mouse and dealing with popup)

cpinter:
Very cool, thank you @lassoan!

As a fun fact, this is the fruition of this NA-MIC week project from 2013: https://www.na-mic.org/wiki/2013_Project_Week:SimplifyRendererMouseInteraction :)

Modules/Loadable/SubjectHierarchy/Widgets/qSlicerSubjectHierarchyPluginLogic.cxx
```diff
@@ -123,7 +125,9 @@ qSlicerSubjectHierarchyPluginLogic::~qSlicerSubjectHierarchyPluginLogic() = defa
 void qSlicerSubjectHierarchyPluginLogic::registerCorePlugins()
 {
   qSlicerSubjectHierarchyPluginHandler::instance()->registerPlugin(
-    new qSlicerSubjectHierarchyFolderPlugin());
+    new qSlicerSubjectHierarchyViewContextMenuPlugin());
+  qSlicerSubjectHierarchyPluginHandler::instance()->registerPlugin(
+    new qSlicerSubjectHierarchyViewContextMenuPlugin());
```
cpinter, line 130: The new view context menu plugin seems to be registered twice, but the folder plugin is now missing.

lassoan: Thank you, good catch! I've noticed this and fixed locally just haven't pushed the change yet, as I'm still working on adding some of the actions that @pieper recommended - and for this implementing a system for specifying sections and weights so that we can have separators between unrelated set of actions and have better control over what appears where.

cpinter: Wow, very nice! This will be a complete solution from the start then. Thanks Andras!

PR #5692, opened 2021-06-13
ENH: Add interface to register default sequence storage nodes

Sunderlandkyl:
Currently, it is not possible in a module to add a default SequenceStorageNode for a data node class in the Slicer core.
This commit adds an interface so that Extensions can now add default sequence storage node classes for specific data types by registering the data node/storage node in the default sequence node.

Example of use:
````
  defaultSequenceNode = slicer.mrmlScene.CreateNodeByClass("vtkMRMLSequenceNode")
  defaultSequenceNode.RegisterAdditionalStorageNode("vtkMRMLStreamingVolumeNode", "vtkMRMLStreamingVolumeSequenceStorageNode")
  slicer.mrmlScene.AddDefaultNode(defaultSequenceNode)
  defaultSequenceNode.UnRegister(None)
````

### DIFF
--- a/Libs/MRML/Core/vtkMRMLStorableNode.h
+++ b/Libs/MRML/Core/vtkMRMLStorableNode.h
@@ -123,7 +123,9 @@ public:
   virtual bool AddDefaultStorageNode(const char* filename = nullptr);
 
   /// Class name of the default sequence storage node that is instantiated by CreateDefaultSequenceStorageNode
-  /// Default: "vtkMRMLSequenceStorageNode"
+  /// The value is not stored in the scene but it has to be set manually
+  /// (for example, in the corresponding default node in the scene).
+  /// Default value: "vtkMRMLSequenceStorageNode"
   /// \sa CreateDefaultSequenceStorageNode
   vtkSetMacro(DefaultSequenceStorageNodeClassName, std::string);
   vtkGetMacro(DefaultSequenceStorageNodeClassName, std::string);
--- a/Libs/MRML/Core/vtkMRMLStorableNode.h
+++ b/Libs/MRML/Core/vtkMRMLStorableNode.h
@@ -57,6 +57,10 @@ public:
   /// Write this node's information to a MRML file in XML format.
   void WriteXML(ostream& of, int indent) override;
 
+  /// \brief Copy node contents from another node of the same type.
+  /// Reimplemented to copy default sequence storage node class.
+  void Copy(vtkMRMLNode* node) override;
+
   /// Copy node content (excludes basic data, such as name and node references).
   /// \sa vtkMRMLNode::CopyContent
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Review by lassoan — COMMENTED** [review]
The current design is that the default sequence storage node is provided by the storable node. If this is not flexible enough then we have two options:
- Option A: improve the current design. For example, add a member variable that stores the default sequence storage class in a string that can be overridden by modifying the default node.
- Option B: replace the current design. The proposed design looks OK (with the small improvement that I described in the inline comment), but it does not look much better or more flexible than the current design. Maybe the current limitations that I did not realize, and it is unfixe. If this is the case, then we should remove the old mechanism and use the new mechanism for the volume sequence storage node.

I would prefer to at least give Option A a try, because it woulbed a less disruptive change.

File: Libs/MRML/Core/vtkMRMLStorableNode.h [hunk 1]
```diff
@@ -118,6 +122,12 @@ class VTK_MRML_EXPORT vtkMRMLStorableNode : public vtkMRMLNode
   /// storage node is not created and the method returns with true.
   virtual bool AddDefaultStorageNode(const char* filename = nullptr);
 
+  /// Class name of the default sequence storage node that is instantiated by CreateDefaultSequenceStorageNode
+  /// Default: "vtkMRMLSequenceStorageNode"
+  /// \sa CreateDefaultSequenceStorageNode
```
lassoan, line 127: add a comment that it is not stored in the scene

lassoan (commented):
This looks good (just added one trivial comment).

Check if we need to change any extensions (probably not). The easiest is to get all source code of all extensions is to [build the ExtensionIndex](https://www.slicer.org/wiki/Documentation/Nightly/Developers/Build_ExtensionsIndex).

Libs/MRML/Core/vtkMRMLStorableNode.h
```diff
@@ -118,6 +122,12 @@ class VTK_MRML_EXPORT vtkMRMLStorableNode : public vtkMRMLNode
   /// storage node is not created and the method returns with true.
   virtual bool AddDefaultStorageNode(const char* filename = nullptr);
 
+  /// Class name of the default sequence storage node that is instantiated by CreateDefaultSequenceStorageNode
+  /// Default: "vtkMRMLSequenceStorageNode"
```
lassoan, line 126:
```suggestion
  /// The value is not stored in the scene but it has to be set manually
  /// (for example, in the corresponding default node in the scene).
  /// Default value: "vtkMRMLSequenceStorageNode"
```

lassoan:
I've merged this now. @Sunderlandkyl it would be great if you could check if there is no impact on SlicerIGT and PerkTutor extensions.

Issue #5695, opened 2021-06-15
Building Superbuild VTK fails with missing Qt component (Qt5X11Extras)  in Linux systems


RafaelPalomar:
Building the Superbuild VTK fails with the following error
```
-- SplineDrivenImageSlicer: Building as a Remote VTK module
CMake Error at /usr/lib64/cmake/Qt5/Qt5Config.cmake:28 (find_package):
  Could not find a package configuration file provided by "Qt5X11Extras" with
  any of the following names:

    Qt5X11ExtrasConfig.cmake
    qt5x11extras-config.cmake

  Add the installation prefix of "Qt5X11Extras" to CMAKE_PREFIX_PATH or set
  "Qt5X11Extras_DIR" to a directory containing one of the above files.  If
  "Qt5X11Extras" provides a separate development package or SDK, be sure it
  has been installed.
Call Stack (most recent call first):
  GUISupport/Qt/CMakeLists.txt:77 (find_package)
```
## Steps to reproduce

1. Configure Slicer in a Linux system without the Qt5X11Extras component installed.
2. Try to build

## Expected behavior

The detection of a missing component should be done during configuration of Slicer and not during the process of building

## Environment
- Slicer version: Slicer-v4.11.20210226 
- Operating system: Gentoo Linux (I believe this operates the same in all linux systems)

PR #5696, opened 2021-06-15
COMP: Check for Qt5X11Extras on Linux

RafaelPalomar:
`Qt5X11Extras` is a Qt componnent required to build the Superbuild `VTK` in
the current configuration, however, this component is not required when
configuring Slicer. In abscence of installed `Qt5X11Extras` the build will
fail on `VTK`.

his commit adds Qt5X11Extras as a required component for Linux systems
and closes Slicer/Slicer#5695

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -576,6 +576,13 @@ endif()
     Xml XmlPatterns
     Svg Sql
     )
+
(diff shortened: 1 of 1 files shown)


jcfr (approved):

jcfr:
Thanks for the patch :pray:

PR #5700, opened 2021-06-17
ENH: Make Clone action clone a segment when a segment is selected in subject hierarchy

lassoan:
Previously, "Clone" action duplicated the entire segmentation node, even when right-clicked on a segment.
This commit changes the behavior so that if the user right-clicks on a Segment and chooses Clone then only that selected segment will be cloned.


Modules/Loadable/Segmentations/SubjectHierarchyPlugins/qSlicerSubjectHierarchySegmentationsPlugin.cxx
```diff
@@ -712,10 +712,19 @@ void qSlicerSubjectHierarchySegmentationsPlugin::onSegmentAdded(vtkObject* calle
...
+      {
+      positionUnderParent = segmentation->GetSegmentIndex(segmentId);
+      }
+
     // Add the segment in subject hierarchy to allow individual handling (e.g. visibility)
     vtkIdType segmentShItemID = shNode->CreateHierarchyItem(
       segmentationShItemID, (segment->GetName() ? segment->GetName() : ""),
-      vtkMRMLSubjectHierarchyConstants::GetSubjectHierarchyVirtualBranchAttributeName());
+      vtkMRMLSubjectHierarchyConstants::GetSubjectHierarchyVirtualBranchAttributeName(),
+      segmentation->GetSegmentIndex(segmentId));
```
cpinter, line 727: This should be `positionUnderParent` I guess.

cpinter (changes requested):
Thanks Andras! It looks good, I just added one thing that I think should fixed.

Modules/Loadable/Segmentations/SubjectHierarchyPlugins/qSlicerSubjectHierarchySegmentationsPlugin.cxx
lassoan: Yes! Good catch.

PR #5704, opened 2021-06-21
Fix markups edit properties

lassoan:
Fixes https://discourse.slicer.org/t/crash-on-selecting-point-properties-after-using-surface-cut/18234/2


lassoan:
@Sunderlandkyl I'll merge this now, as pull requests related to markups are piling up, but it would be nice if you could have a look at this later.

Sunderlandkyl:
Sorry, I missed this. It looks good to me.

PR #5710, opened 2021-06-28
DOC: Update plugin logic example based on recent renames

jamesobutler:
Methods were renamed as part of recent work in 598dea0b5f2acadffde161111bc071eff0dc8b25.

`setDisplayedViewMenuActionNames` and `availableViewMenuActionNames` were originally added as part of 62fe31fb179e5c904f5d8157739483dee6669a44 which was included in Slicer v4.11.20200930 and v4.11.20210226.


lassoan (approved):
Thank you!

PR #5716, opened 2021-07-01
ENH: Add context managers to show while waiting

fepegar:
These are some potentially useful utils coming from https://github.com/fepegar/SlicerParcellation/blob/master/lib/GeneralUtils.py.

@lassoan's comments were:

> There is indeed some repeating code pattern with the waitcursor and printing of exception messages which would be nice to simplify. Please send a pull request to Slicer core with the content of GeneralUtils.py to start a discussion. Initial comment: the overridecursor has to be restored when an exception occurs ("finally" is too late because the popup window is displayed in "except" and it is not nice if the waitcursor is still shown when the user gets the error popup). messageContextManager: It seems that it is intended for progress reporting, but for thatyou can already use the statusbar or delaydisplay. A context manager for error messages could be more useful: it would take a general error message (that it would display if there is an exception) and would offer to show the exception text if user clicks "Details" button. Something like this: https://github.com/Slicer/Slicer/blob/aaa9f918a4ee4f35bf09e05c5d7e53d083b2e4e3/Modules/Scripted/DICOMLib/DICOMSendDialog.py#L91-L99

My replies:

> the overridecursor has to be restored when an exception occurs ("finally" is too late because the popup window is displayed in "except" and it is not nice if the waitcursor is still shown when the user gets the error popup)

I'm not sure which popup window you mean. Something created by the user? `showWaitCursor` doesn't show a popup window.

> messageContextManager: It seems that it is intended for progress reporting, but for that you can already use the statusbar or delaydisplay.

The status bar might not be very obvious. This is how I've used this in a module (0:35, here being played at x2): https://youtu.be/Gg02F9DGDc4?t=35. Not sure how I would use delayDisplay with the context manager. I wouldn't like buttons in the window.

> A context manager for error messages could be more useful: it would take a general error message (that it would display if there is an exception) and would offer to show the exception text if user clicks "Details" button. 

Thanks. I have added a `tryWithErrorDisplay` context manager.

Comments:

1. Maybe the names can be improved
1. The goal of these is two remove a bit of verbosity from users' code.
1. I imported `contextmanager` to be able to use the decorator. This seems forbidden in `slicer.util`, though. Another option is to use the class syntax for context managers. Why are all the imports inside the functions? Why are many of them in one line, this is against [PEP8](https://www.python.org/dev/peps/pep-0008/#imports).
1. Why is the file so long? Why not splitting it into different submodules and importing everything here if necessary?

Here are some snippets to try these:

```python
import time
import random

def work():
  time.sleep(random.randrange(1, 3))

def fail():
  work()
  int('Szia')

showThings = True

#
with showWaitCursor(show=showThings):
  work()

#
with showWaitCursor(show=showThings):
  fail()

#
with messageContextManager('¡Hola!', show=showThings):
  work()

#
with messageContextManager('¡Hola!', show=showThings):
  fail()

#
with tryWithErrorDisplay(show=showThings):
  work()

#
with tryWithErrorDisplay(show=showThings):
  fail()

#
getPythonConsoleWidget().show()
with peakPythonConsole(show=showThings):
  print('This should just wait a bit')
  slicer.app.processEvents()
  work()

# This one needs to be pasted with a blank line at the end so it runs
getPythonConsoleWidget().hide()
slicer.app.processEvents()
time.sleep(2)
with peakPythonConsole(show=showThings):
  print('I will close soon!')
  slicer.app.processEvents()
  work()


```

### DIFF
--- a/Base/Python/slicer/util.py
+++ b/Base/Python/slicer/util.py
@@ -2429,6 +2429,170 @@ def createProgressDialog(parent=None, value=0, maximum=100, labelText="", window
       setattr(progressIndicator, key, value)
   return progressIndicator
 
+
+from contextlib import contextmanager
+
+@contextmanager
+def displayPythonShell(display=True):
+  """Show the Python console while the code in the context manager is being run.
+
+  The console stays visible only if it was visible already.
+
+  :param display: If show is False, the context manager has no effect.
+
+  .. code-block:: python
+
+    with slicer.util.displayPythonShell():
+      slicer.util.pip_install('nibabel')
+
+  """
+  import slicer
+
+  def dockableWindowEnabled():
+    import qt
+    return toBool(qt.QSettings().value("Python/DockableWindow"))
+
+  def getConsole():
+    return mainWindow().pythonConsole().parent() if dockableWindowEnabled() else pythonShell()
+
+  if display:
+    console = getConsole()
+    consoleVisible = console.visible
+    console.show()
+    slicer.app.processEvents()
+  try:
+    yield
+  finally:
+    if display:
+      console.setVisible(consoleVisible)
+
+class WaitCursor:
+  """Display a wait cursor while the code in the context manager is being run.
+
+  :param show: If show is False, no wait cursor is shown.
+
+  .. code-block:: python
+
+    import time
+
+    n = 2
+    with slicer.util.MessageDialog(f'Sleeping for {n} seconds...'):
+      with slicer.util.WaitCursor():
+        time.sleep(n)
+
+  """
+  def __init__(self, show=True):
+    """Set the cursor to waiting mode while the code in the context manager is being run.
+
+    :param show: If show is False, this context manager has no effect.
+
+    .. code-block:: python
+
+      import time
+
+      with slicer.util.WaitCursor():
+        time.sleep(2)
+    """
+    self.show = show
+
+  def __enter__(self):
+    import qt, slicer
+    if self.show:
+      qt.QApplication.setOverrideCursor(qt.Qt.WaitCursor)
+      slicer.app.processEvents()
+
+  def __exit__(self, type, value, traceback):
+    if self.show:
+      import qt
+      qt.QApplication.restoreOverrideCursor()
+
+class MessageDialog:
+  def __init__(self, message, show=True, logLevel=None):
+    """Log the message and show a message box while the code in the context manager is being run.
+
+    :param message: Text shown in the message box.
+    :param show: If show is False, no dialog is shown.
+    :param logLevel: Log level used to log the message. Default: logging.INFO
+
+    .. code-block:: python
+
+      import time
+
+      n = 2
+      with slicer.util.MessageDialog(f'Sleeping for {n} seconds...'):
+        with slicer.util.WaitCursor():
+          time.sleep(n)
+
+    """
+    import logging
+    import slicer
+
+    if logLevel is None:
+      logLevel = logging.INFO
+    if not isinstance(logLevel, int):
+      raise ValueError(f'Invalid log level: {logLevel}')
+
+    self.message = message
+    self.show = show and not slicer.app.testingEnabled()
+    self.logLevel = logLevel
+    self.box = None
+
+  def __enter__(self):
+    import logging
+    logging.log(self.logLevel, self.message)
+
+    if self.show:
+      import qt, slicer
+      self.box = qt.QMessageBox()
+      self.box.setStandardButtons(qt.QMessageBox.NoButton)
+      self.box.setText(self.message)
+      self.box.show()
+      slicer.app.processEvents()
+
+  def __exit__(self, type, value, traceback):
+      if self.show:
+        self.box.accept()
+
+@contextmanager
+def tryWithErrorDisplay(message=None, show=True, waitCursor=False):
+  """Show an error display with the error details if an exception is raised.
+
+  :param message: Text shown in the message box.
+  :param show: If show is False, the context manager has no effect.
+  :param waitCursor: If waitCrusor is set to True then mouse cursor is changed to
+     wait cursor while the context manager is being run.
+
+  .. code-block:: python
+
+    import random
+
+    def risky():
+      if random.choice((True, False)):
+        raise Exception('Error while trying to do some internal operations.')
+
+    with slicer.util.tryWithErrorDisplay("Risky operation failed."):
+      risky()
+  """
+  try:
+    if waitCursor:
+      import slicer, qt
+      slicer.app.setOverrideCursor(qt.Qt.WaitCursor)
+    yield
+    if waitCursor:
+      slicer.app.restoreOverrideCursor()
+  except Exception as e:
+    import slicer
+    if waitCursor:
+      slicer.app.restoreOverrideCursor()
+    if show and not slicer.app.testingEnabled():
+      if message is not None:
+        errorMessage = f'{message}\n\n{e}'
+      else:
+        errorMessage = str(e)
+      import traceback
+      errorDisplay(errorMessage, detailedText=traceback.format_exc())
+    raise
+
 def toBool(value):
   """Convert any type of value to a boolean.
 
--- a/Base/Python/slicer/ScriptedLoadableModule.py
+++ b/Base/Python/slicer/ScriptedLoadableModule.py
@@ -200,15 +200,10 @@ class ScriptedLoadableModuleWidget:
     """Reload scripted module widget representation and call :func:`ScriptedLoadableModuleTest.runTest()`
     passing ``kwargs``.
     """
-    try:
+    with slicer.util.tryWithErrorDisplay("Reload and Test failed."):
       self.onReload()
       test = slicer.selfTests[self.moduleName]
       test(msec=int(slicer.app.userSettings().value("Developer/SelfTestDisplayMessageDelay")), **kwargs)
-    except Exception as e:
-      import traceback
-      traceback.print_exc()
-      errorMessage = "Reload and Test: Exception!\n\n" + str(e) + "\n\nSee Python Console for Stack Trace"
-      slicer.util.errorDisplay(errorMessage)
 
   def onEditSource(self):
     filePath = slicer.util.modulePath(self.moduleName)
--- a/Modules/Scripted/DICOMLib/DICOMSendDialog.py
+++ b/Modules/Scripted/DICOMLib/DICOMSendDialog.py
@@ -88,17 +88,12 @@ class DICOMSendDialog(qt.QDialog):
     self.cancelRequested = False
     okButton = self.bbox.button(self.bbox.Ok)
 
-    try:
-      #qt.QApplication.setOverrideCursor(qt.Qt.WaitCursor)
+    with slicer.util.tryWithErrorDisplay("DICOM sending failed."):
       okButton.enabled = False
       DICOMLib.DICOMSender(self.files, address, protocol, aeTitle=aeTitle, progressCallback=self.onProgress)
       logging.debug("DICOM sending of %s files succeeded" % len(self.files))
       self.close()
-    except Exception as result:
-      import traceback
-      slicer.util.errorDisplay("DICOM sending failed: %s" % str(result), parent=self.parent().window(), detailedText=traceback.format_exc())
 
-    #qt.QApplication.restoreOverrideCursor()
     okButton.enabled = True
     self.sendingIsInProgress = False
 
--- a/Modules/Scripted/DICOMPatcher/DICOMPatcher.py
+++ b/Modules/Scripted/DICOMPatcher/DICOMPatcher.py
@@ -128,8 +128,8 @@ class DICOMPatcherWidget(ScriptedLoadableModuleWidget):
     pass
 
   def onPatchButton(self):
-    slicer.app.setOverrideCursor(qt.Qt.WaitCursor)
-    try:
+    with slicer.util.tryWithErrorDisplay("Unexpected error.", waitCursor=True):
+
       import tempfile
       if not self.outputDirSelector.currentPath:
         self.outputDirSelector.currentPath =  tempfile.mkdtemp(prefix="DICOMPatcher-", dir=slicer.app.temporaryPath)
@@ -154,12 +154,6 @@ class DICOMPatcherWidget(ScriptedLoadableModuleWidget):
       if self.normalizeFileNamesCheckBox.checked:
         self.logic.addRule("NormalizeFileNames")
       self.logic.patchDicomDir(self.inputDirSelector.currentPath, self.outputDirSelector.currentPath)
-
-    except Exception as e:
-      self.addLog(f"Unexpected error: {str(e)}")
-      import traceback
-      traceback.print_exc()
-    slicer.app.restoreOverrideCursor()
 
   def onImportButton(self):
     self.logic.importDicomDir(self.outputDirSelector.currentPath)
--- a/Modules/Scripted/SegmentStatistics/SegmentStatistics.py
+++ b/Modules/Scripted/SegmentStatistics/SegmentStatistics.py
@@ -182,7 +182,7 @@ class SegmentStatisticsWidget(ScriptedLoadableModuleWidget):
     """Calculate the label statistics
     """
 
-    try:
+    with slicer.util.tryWithErrorDisplay("Failed to compute results.", waitCursor=True):
       if not self.outputTableSelector.currentNode():
         newTable = slicer.mrmlScene.AddNewNodeByClass("vtkMRMLTableNode")
         self.outputTableSelector.setCurrentNode(newTable)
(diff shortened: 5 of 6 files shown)


Base/Python/slicer/util.py
```diff
@@ -2395,6 +2395,116 @@ def createProgressDialog(parent=None, value=0, maximum=100, labelText="", window
...
   return progressIndicator
 
+def getPythonConsoleWidget():
+  import slicer
+  return slicer.util.mainWindow().pythonConsole().parent()
+
+from contextlib import contextmanager
+
+@contextmanager
+def peakPythonConsole(show=True):
```
jcfr, line 2405: With this PR, we would now have the following functions:

| slicer.util API | equivalent API |
|---|---|
| slicer.util.pythonShell | slicer.app.pythonConsole |
| slicer.util.getPythonConsoleWidget | slicer.util.mainWindow().pythonConsole().parent() |

There are two issues with introducing the new function `getPythonConsoleWidget`:
* function is inconsistently named. Name `pythonShellWidget` would be consistent
* The suffix `Widget` is not always "correct". Indeed, when the settings `Python/DockableWindow` is false, the console is standalone dialog.

Considering this, I suggest to avoid adding the function `getPythonConsoleWidget` and rename `peakPythonConsole` to `raisePythonShell` and update the implementation to consider the value of `Python/DockableWindow` settings to conditionally get the parent widget.

For context:

https://github.com/Slicer/Slicer/blob/aaa9f918a4ee4f35bf09e05c5d7e53d083b2e4e3/Base/QTApp/qSlicerMainWindow.cxx#L456-L477

Base/Python/slicer/util.py
```diff
@@ -2395,6 +2395,116 @@ def createProgressDialog(parent=None, value=0, maximum=100, labelText="", window
...
+    qt.QApplication.setOverrideCursor(qt.Qt.WaitCursor)
+    slicer.app.processEvents()
+  try:
+    yield
+  finally:
+    if show:
+      qt.QApplication.restoreOverrideCursor()
+
+@contextmanager
+def messageContextManager(message, show=True):
```
jcfr, line 2454: Some context manager have the suffix `ContextManager` and some other ones to not. We should be consistent.

Also, we already have one context manager implemented as a class and it is called `NodeModify` that can only be used as context manager and not as a callable function.

It looks like some of the proposed context manager should also be only context manager, hence these suggestions:

* `showWaitCursor`
  * rename to `WaitCursor` and convert into a class like `NodeModify`
 
* `tryWithErrorDisplay`
  * instead introduce a manager called `slicer.util.raises` with the possibility to choose with option to catch.
  * always hide the dialog with testing mode is enabled  (`slicer.app.testingEnabled()`)
  * inspired from [pytest](https://docs.pytest.org/en/6.2.x/assert.html#assertions-about-expected-exceptions)
 
* `messageContextManager`
  * renamed to `MessageDialog`
  * consider approaching this as a decorator ? 
  * integrate with testing mode  (`slicer.app.testingEnabled()`)

Implementation of `NodeModify`:

https://github.com/Slicer/Slicer/blob/aaa9f918a4ee4f35bf09e05c5d7e53d083b2e4e3/Base/Python/slicer/util.py#L1343-L1351

jcfr (changes requested):

Base/Python/slicer/util.py
fepegar: Thanks, @jcfr. I think I have addressed this.

Base/Python/slicer/util.py
fepegar: Thanks, @jcfr. I hadn't seen `NodeModify`. What is the rationale for choosing decorator vs class?

> with the possibility to choose with option to catch.

I'm not sure what you mean here. Can you maybe sketch some pseudo code?

> integrate with testing mode

What do you mean by "integrate" here? I'm not sure what you're asking me to do.

jcfr: re: integrate with testing code

If tests are running on the CI, no dialog should be shown. Considering that when we run tests in CI, we pass the parameter `--testing`, the function `slicer.app.testingEnabled()` returns True .. and we should test for this to make sure no dialog are shown. Failing to do would require user input ... which is not possible when running tests.

fepegar: Oh, sure. The same as above, then. Thanks.

fepegar: I've turned those two generators into classes and integrated all of these with `testingEnabled`.

> instead introduce a manager called slicer.util.raises with the possibility to choose with option to catch.

I'm still not sure how this is related to pytest's raises, nor why we are only converting some of these into classes.

Base/Python/slicer/util.py
```diff
@@ -2395,6 +2395,130 @@ def createProgressDialog(parent=None, value=0, maximum=100, labelText="", window
...
+      qt.QApplication.setOverrideCursor(qt.Qt.WaitCursor)
+      slicer.app.processEvents()
+
+  def __exit__(self, type, value, traceback):
+    if self.show:
+      import qt
+      qt.QApplication.restoreOverrideCursor()
+
+class MessageDialog:
+  def __init__(self, message, show=True):
```
jcfr, line 2462: It may be sensible to introduce `logLevel` parameter set to `INFO`

Base/Python/slicer/util.py
jcfr: >  I'm still not sure how this is related to pytest's raises

Instead of hard coding the general python class `Exception`, it was suggestion it is passed as a parameter.

Also , the function would be renamed from `tryWithErrorDisplay` to `raises`

and would be used like this:

```
with slicer.util.raises(Exception):
  [...]
```

jcfr: Generally speaking it also bad practice to catch the general exception, it should be specific one ...

fepegar: The goal of this manager is "try to do something and if you get _any_ error, show it to me". That's why I'm "catching" (not really, it's just to show the message on the dialog) the general exception. The idea is that the user of this context manager doesn't know a priori which exceptions may be raised.

jcfr: This is a relevant reading: https://stackoverflow.com/questions/4990718/about-catching-any-exception#4992124

If the goal is to catch the exception to let the user know about .. then we could consider re-raising it afterward to make sure it can be handled later by the application  ?

Base/Python/slicer/util.py
```diff
@@ -2395,6 +2395,140 @@ def createProgressDialog(parent=None, value=0, maximum=100, labelText="", window
       setattr(progressIndicator, key, value)
   return progressIndicator
 
+
+from contextlib import contextmanager
+
+@contextmanager
+def raisePythonShell(raise=True):
```
lassoan, line 2402: `raise` may be misleading (users may think that it is related to raising exceptions). I think "display" would be simpler and more clear.

Base/Python/slicer/util.py
```diff
@@ -2395,6 +2395,140 @@ def createProgressDialog(parent=None, value=0, maximum=100, labelText="", window
...
+      console.setVisible(consoleVisible)
+    console = getConsole()
+    consoleVisible = console.visible
+    console.show()
+    slicer.app.processEvents()
+  try:
+    yield
+  finally:
+    if show:
+      console.setVisible(consoleVisible)
```
fepegar, line 2443: Hmm I think I messed this up. This is not why I committed originally.

lassoan:
> I'm not sure which popup window you mean. Something created by the user? showWaitCursor doesn't show a popup window

My point is that most often you need both waitcursor + display of error message in a popup (if an exception is occurred). Therefore, it is important that when an exception occurs then the waitcursor is restored in the `except` branch and not the `finally` branch, because if you have popup display (by user's exception handling or other context manager) then that may display a popup, and you want the cursor to be restored to the normal cursor by the time the popup is displayed. Maybe the code works like this already, just make sure to check that it works like this (if a long-running operation fails with an exception then you should see an error popup and the mouse cursor should be normal).

Base/Python/slicer/util.py
fepegar: I added it, but I'm not sure it's what you expected.

fepegar:
> My point is that most often you need both waitcursor + display of error message in a popup (if an exception is occurred). Therefore, it is important that when an exception occurs then the waitcursor is restored in the except branch and not the finally branch, because if you have popup display (by user's exception handling or other context manager) then that may display a popup, and you want the cursor to be restored to the normal cursor by the time the popup is displayed. Maybe the code works like this already, just make sure to check that it works like this (if a long-running operation fails with an exception then you should see an error popup and the mouse cursor should be normal).

Ok, I see. I don't think that's the current behavior. I'll see if I can fix it.

lassoan:
> The status bar might not be very obvious. This is how I've used this in a module (0:35, here being played at x2): https://youtu.be/Gg02F9DGDc4?t=35. Not sure how I would use delayDisplay with the context manager. I wouldn't like buttons in the window.

I find the windows popping up a bit messy. It is acceptable (and definitely much better than not having feedback at all), but it is not something that I would like to see as a common pattern.

Showing a single progress window with updating progress bar and status text is more appropriate. You could make it a popup, but I think it would be nicer to have it in the module widget (right below the "Update visualization" button; similarly to CLI module progress reporting). Or, if you want users to be able to read what's going on (and scroll back to see all messages) then dumping all messages into a scrollable text browser widget may work, too (similarly to DICOM patcher). If we find that many modules need to do lengthy synchronous processing then we could even create a custom widget, which would combine these two (progress text, progress bar, and a textbox showing all previous messages in a collapsible section).

fepegar:
@lassoan 

About the cursor: if the user catches the exception and shows the popup, then WaitCursor won't see that an exception is raised, right? I don't know how to solve that.

About the message box: ok, I agree it's a bit invasive. The issue with the progress bar is that you would need to pass around the bar, the number of steps, and the current step to all child methods, functions, etc. that need to update it, right? Also, sometimes it's a single step so a progress bar maybe wouldn't make much sense. For example, I've been using it during the installation of PyTorch, which might be a bit lengthy.

fepegar:
I see what you mean by using the status bar when I download a large CT from Sample Data. What I don't like is that the cursor stays the same + the GUI can't be used, and that's not very nice. I suppose these issues are what I'm trying to overcome with this PR.

Base/Python/slicer/util.py
fepegar: The guy in the post is doing exactly what I've done...

I agree with the re-raising. I'll add it now.

lassoan:
> About the cursor: if the user catches the exception and shows the popup, then WaitCursor won't see that an exception is raised, right? I don't know how to solve that.

Can this be solved by nesting the errordisplay and waitcursor context managers in the right order? If not, then it could make sense to have a context manager instead, which sets/restores the wait cursor and also displays the error message (if exception occurs).

> About the message box: ok, I agree it's a bit invasive. The issue with the progress bar is that you would need to pass around the bar, the number of steps, and the current step to all child methods, functions, etc. that need to update it, right? Also, sometimes it's a single step so a progress bar maybe wouldn't make much sense. For example, I've been using it during the installation of PyTorch, which might be a bit lengthy.

The main thing that I don't like about the many popups is the flashing of popups of various sizes. It may also be disruptive (depending on the operating system and window manager) if the popups appear over other applications (that the user switched to, waiting for the lengthy operation to complete), and they may even keep raise the application to the top (making the computer unusable for any other tasks).

It should not be necessary to pass around anything. What we usually do is to have a hook in the logic that provides a string and completion% value to a callback function. If you want multi-level progress reporting then you can rescale the received completion% value in the callback function. See for example the logCallback in https://github.com/Slicer/Slicer/blob/master/Modules/Scripted/DICOMPatcher/DICOMPatcher.py (it just has a string parameter, but you can add a percentage parameter as well).

If you have no idea about the number of steps then you can set maximum=0 and then you get an infinite progress bar:
```
p=slicer.util.createProgressDialog(maximum=0)
```

You can hide the "Cancel" button if you don't want your lengthy operation to be interruptible (in general, it should be interruptible, though):

```
p.setCancelButton(None)
```

lassoan:
> I see what you mean by using the status bar when I download a large CT from Sample Data. What I don't like is that the cursor stays the same + the GUI can't be used, and that's not very nice. I suppose these issues are what I'm trying to overcome with this PR.

I agree that the status bar makes the most sense for background tasks. I guess the intent was to make data download a background task (I think it was that, around Slicer3). There are now nice packages for downloading data in the background, multiple threads, so probably it could be made a background task again relatively easily.

lassoan:
Sorry for asking for so much, but I cannot resist asking for this: could you also add a context manager for `slicer.app.pauseRender()` and `slicer.app.resumeRender()`? (this temporarily pauses updates of all views, to prevent showing temporary content in viewers and make the processing faster - for very large and complex scenes or lots of updates)

Base/Python/slicer/util.py
```diff
@@ -2395,6 +2395,139 @@ def createProgressDialog(parent=None, value=0, maximum=100, labelText="", window
...
+
+from contextlib import contextmanager
+
+@contextmanager
+def displayPythonShell(display=True):
+  """Show the Python console while the code in the context manager is being run.
+
+  The console stays visible only if it was visible already.
+
+  :param show: If show is False, the context manager has no effect.
```
jcfr, line 2407:
```suggestion
  :param display: If show is False, the context manager has no effect.
```

Base/Python/slicer/util.py
```diff
@@ -2395,6 +2395,139 @@ def createProgressDialog(parent=None, value=0, maximum=100, labelText="", window
...
+      qt.QApplication.setOverrideCursor(qt.Qt.WaitCursor)
+      slicer.app.processEvents()
+
+  def __exit__(self, type, value, traceback):
+    if self.show:
+      import qt
+      qt.QApplication.restoreOverrideCursor()
+
+class MessageDialog:
+  def __init__(self, message, show=True, logLevel='INFO'):
```
jcfr, line 2462:
```suggestion
  def __init__(self, message, show=True, logLevel=logging.INFO):
```

Base/Python/slicer/util.py
```diff
@@ -2395,6 +2395,139 @@ def createProgressDialog(parent=None, value=0, maximum=100, labelText="", window
...
+    self.message = message
+    self.show = show and not slicer.app.testingEnabled()
+    self.logLevel = logLevel
+    self.box = None
+
+  def __enter__(self):
+    import logging
+    numericLevel = getattr(logging, self.logLevel.upper(), None)
+    if not isinstance(numericLevel, int):
+        raise ValueError(f'Invalid log level: {self.logLevel}')
```
jcfr, line 2486:
```suggestion
```

Base/Python/slicer/util.py
```diff
@@ -2395,6 +2395,139 @@ def createProgressDialog(parent=None, value=0, maximum=100, labelText="", window
...
+    :param show: If show is False, no dialog is shown.
+
+    .. code-block:: python
+
+      import time
+
+      n = 2
+      with slicer.util.messageContextManager(f'Sleeping for {n} seconds...'):
+        time.sleep(n)
+    """
```
jcfr, line 2475:
```suggestion
    """
    if not isinstance(logLevel, int):
      raise ValueError(f'Invalid log level: {logLevel}')
```

Base/Python/slicer/util.py
```diff
@@ -2395,6 +2395,139 @@ def createProgressDialog(parent=None, value=0, maximum=100, labelText="", window
...
+    if self.show:
+      import qt
+      qt.QApplication.restoreOverrideCursor()
+
+class MessageDialog:
+  def __init__(self, message, show=True, logLevel='INFO'):
+    """Log the message and show a message box while the code in the context manager is being run.
+
+    :param message: Text shown in the message box.
+    :param show: If show is False, no dialog is shown.
```
jcfr, line 2466:
```suggestion
    :param show: If show is False, no dialog is shown.
    :param logLevel: Log level used to log the message.
```

Base/Python/slicer/util.py
```diff
@@ -2395,6 +2395,139 @@ def createProgressDialog(parent=None, value=0, maximum=100, labelText="", window
...
+
+  def __enter__(self):
+    import logging
+    numericLevel = getattr(logging, self.logLevel.upper(), None)
+    if not isinstance(numericLevel, int):
+        raise ValueError(f'Invalid log level: {self.logLevel}')
+    originalLevel = logging.root.level
+    logging.basicConfig(level=numericLevel)
+    logging.info(self.message)
+    logging.basicConfig(level=originalLevel)
```
jcfr, line 2490:
```suggestion
    logging.log(logLevel, self.message)
```

I think we should respect the overall level configured in the application and not force it

Base/Python/slicer/util.py
```diff
@@ -2395,6 +2395,139 @@ def createProgressDialog(parent=None, value=0, maximum=100, labelText="", window
...
+
+    if self.show:
+      import qt, slicer
+      self.box = qt.QMessageBox()
+      self.box.setStandardButtons(qt.QMessageBox.NoButton)
+      self.box.setText(self.message)
```
jcfr, line 2496: The message box icon should be set based on the log level. See https://doc.qt.io/qt-5/qmessagebox.html#Icon-enum

Something like this ... 

```python
If logLevel >= logging.CRITICAL:
  icon = qt.QMessageBox.Critical
elif logLevel >= logging.WARNING:
  icon = qt.QMessageBox.Warning
else:
  icon = qt.QMessageBox.Information
```

jcfr:
> nesting the errordisplay and waitcursor context managers in the right order?

context manager can indeed be nested.
See https://rednafi.github.io/digressions/python/2020/03/26/python-contextmanager.html#nesting-contexts

Base/Python/slicer/util.py
fepegar: I thought this would be best as well, but didn't go for it because it needs to `import logging` outside the definition and for some reason (why??) that doesn't happen in `slicer.util`.

lassoan:
I've made some small improvements, fixes, rebased on latest master and tested everything and it all looks good. I'll merge this now.

lassoan:
The branch is now merged. Thank you for your contributions @fepegar!

fepegar:
Awesome! Thanks, @lassoan.

PR #5719, opened 2021-07-02
BUG: Improve performance of Grow from seeds Segment Editor effect

lassoan:
On a large (400x400x1500) volume with 20 seed segments, Grow from seeds completed in about 50 second, but segment updates added about 30 seconds. There also seemed to be multiple updates for a single user action.

To avoid multiple updates, added previewComputationInProgress flag.
To reduce update time, preview segmentation is filled using ImportLabelmapToSegmentationNode method.

With the changes, initial update takes about 55 seconds, subsequent updates take about 11 seconds.

Remaining problems:
- It would be better to avoid temporary labelmap creation (if we keep this, then before importing we probably need to set the segmentation's parent transform to the labelmap).
- Final update is still slow.

### DIFF
--- a/Modules/Loadable/Segmentations/EditorEffects/Python/AbstractScriptedSegmentEditorAutoCompleteEffect.py
+++ b/Modules/Loadable/Segmentations/EditorEffects/Python/AbstractScriptedSegmentEditorAutoCompleteEffect.py
@@ -48,6 +48,8 @@ class AbstractScriptedSegmentEditorAutoCompleteEffect(AbstractScriptedSegmentEdi
     self.extentGrowthRatio = 0.1  # extent of seed region will be grown outside by this much
     self.minimumExtentMargin = 3
 
+    self.previewComputationInProgress = False
+
 
   def __del__(self, scriptedEffect):
     super(SegmentEditorAutoCompleteEffect,self).__del__()
@@ -55,7 +57,7 @@ class AbstractScriptedSegmentEditorAutoCompleteEffect(AbstractScriptedSegmentEdi
     self.observeSegmentation(False)
 
   @staticmethod
-  def isBackgroundLabelmap(labelmapOrientedImageData):
+  def isBackgroundLabelmap(labelmapOrientedImageData, label=None):
     if labelmapOrientedImageData is None:
       return False
     # If five or more corner voxels of the image contain non-zero, then it is background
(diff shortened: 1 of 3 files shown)


lassoan:
@Sunderlandkyl It would be great if you could review this and fix up the remaining issues or come up with an alternative solution. the problem is that `self.scriptedEffect.modifySegmentByLabelmap(previewNode, segmentID, newSegmentLabelmap, slicer.qSlicerSegmentEditorAbstractEffect.ModificationModeSet)` takes 5 seconds for each segment in a 400x400x1500 volume with 20 segments - an extra 100 seconds for each update. Grow from seeds update takes a few seconds but then we need to wait 1.5 minutes for the preview to update!

Having an API to set an entire non-overlapping segmentation from a labelmap would help (I used import from labelmap node, but it would be slightly better to be able to import directly from vtkOrientedImageData). It would be even better if that method could also take an updateExtent input, because we should be able to get that information from Grow from seeds.

This fix is needed for https://discord.com/channels/843934857620357130/856893699275227166/860509600825344010

Sunderlandkyl:
I've been testing the improvements with masking on MRHead, and I've managed to run into errors mostly when testing masking settings:

```
  File "E:\d\s\S4W\Slicer-build\lib\Slicer-4.13\qt-scripted-modules\SegmentEditorEffects\AbstractScriptedSegmentEditorAutoCompleteEffect.py", line 277, in onPreview
    self.preview()
  File "E:\d\s\S4W\Slicer-build\lib\Slicer-4.13\qt-scripted-modules\SegmentEditorEffects\AbstractScriptedSegmentEditorAutoCompleteEffect.py", line 541, in preview
    previewSegment.SetName(inputSegment.GetName())
AttributeError: 'NoneType' object has no attribute 'SetName'
```

lassoan:
Yes, this PR is just a proof of concept. I did not explicitly match the segment IDs between the input and preview segmentation, so probably if you delete segments then it gets messed up. Having a robust built-in API to set/update a complete segmentation from a labelmap would solve this.

Sunderlandkyl:
I'm not able to push to the PR branch, so here are the changes to the preview update, including the segment ID indexing that we discussed:
https://github.com/Sunderlandkyl/Slicer/tree/grow-from-seeds-update-perf

lassoan:
Thanks a lot @Sunderlandkyl. I've cherry-picked your fix, tested it (worked great!), squashed with the existing commit, and will merge now so that it gets into the nightly build.

PR #5721, opened 2021-07-03
ENH: "Apply to all segments" checkbox added in Segment Editor "Smoothing" effect

rbumm:
Added a new checkbox and logic that enables serial smoothing of all visible segments currently open in the Segment Editor. If no segments are visible there is a logging message. The last selected segment gets restored.

The progress is shown as a statusmessage.

### DIFF
--- a/Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorSmoothingEffect.py
+++ b/Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorSmoothingEffect.py
@@ -78,6 +78,12 @@ If segments overlap, segment higher in the segments table will have priority. <b
     self.jointTaubinSmoothingFactorSlider.pageStep = 0.1
     self.jointTaubinSmoothingFactorLabel = self.scriptedEffect.addLabeledOptionsWidget("Smoothing factor:", self.jointTaubinSmoothingFactorSlider)
 
+    self.applyToAllVisibleSegmentsCheckBox = qt.QCheckBox()
+    self.applyToAllVisibleSegmentsCheckBox.setToolTip("Apply smoothing effect to all visible segments in this segmentation node. \
+                                                      This operation may take a while.")
+    self.applyToAllVisibleSegmentsCheckBox.objectName = self.__class__.__name__ + 'ApplyToAllVisibleSegments'
+    self.applyToAllVisibleSegmentsLabel = self.scriptedEffect.addLabeledOptionsWidget("Apply to all segments:", self.applyToAllVisibleSegmentsCheckBox)
+
     self.applyButton = qt.QPushButton("Apply")
     self.applyButton.objectName = self.__class__.__name__ + 'Apply'
     self.applyButton.setToolTip("Apply smoothing to selected segment")
@@ -87,6 +93,7 @@ If segments overlap, segment higher in the segments table will have priority. <b
     self.kernelSizeMMSpinBox.connect("valueChanged(double)", self.updateMRMLFromGUI)
     self.gaussianStandardDeviationMMSpinBox.connect("valueChanged(double)", self.updateMRMLFromGUI)
     self.jointTaubinSmoothingFactorSlider.connect("valueChanged(double)", self.updateMRMLFromGUI)
+    self.applyToAllVisibleSegmentsCheckBox.connect("stateChanged(int)", self.updateMRMLFromGUI)
     self.applyButton.connect('clicked()', self.onApply)
 
     # Customize smoothing brush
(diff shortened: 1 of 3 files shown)


Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorSmoothingEffect.py
```diff
@@ -158,17 +169,26 @@ def updateMRMLFromGUI(self):
     self.scriptedEffect.setParameter("KernelSizeMm", self.kernelSizeMMSpinBox.value)
     self.scriptedEffect.setParameter("GaussianStandardDeviationMm", self.gaussianStandardDeviationMMSpinBox.value)
     self.scriptedEffect.setParameter("JointTaubinSmoothingFactor", self.jointTaubinSmoothingFactorSlider.value)
+    applyToAllSegments = 1 if self.applyToAllSegmentsCheckBox.isChecked() else 0
+    self.scriptedEffect.setParameter("ApplyToAllSegments", applyToAllSegments)
```
lassoan, line 173: Rename to "ApplyToAllVisibleSegments"

Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorSmoothingEffect.py
```diff
@@ -180,8 +200,27 @@ def onApply(self, maskImage=None, maskExtent=None):
       self.scriptedEffect.saveStateForUndo()
       if smoothingMethod == JOINT_TAUBIN:
         self.smoothMultipleSegments(maskImage, maskExtent)
+      elif applyToAllSegments == '1':
```
lassoan, line 203:
```suggestion
      elif applyToAllVisibleSegments:
```

Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorSmoothingEffect.py
```diff
@@ -158,17 +169,26 @@ def updateMRMLFromGUI(self):
...
 
+  def showStatusMessage(self, msg, timeoutMsec=500):
+        slicer.util.showStatusMessage(msg, timeoutMsec)
+        slicer.app.processEvents()
+
   def onApply(self, maskImage=None, maskExtent=None):
     """maskImage: contains nonzero where smoothing will be applied
     """
     smoothingMethod = self.scriptedEffect.parameter("SmoothingMethod")
+    applyToAllSegments = self.scriptedEffect.parameter("ApplyToAllSegments")
```
lassoan, line 191:
```suggestion
    applyToAllVisibleSegments = int(self.scriptedEffect.parameter("ApplyToAllVisibleSegments")) !=0 if self.scriptedEffect.parameter("ApplyToAllSegments") else False
```

lassoan (commented):
This will be great, thank you!

I've added a few comments inline.

Also, please update the user manual and the developer manual with the description of the new option:

https://github.com/rbumm/Slicer/edit/master/Docs/user_guide/modules/segmenteditor.md

https://github.com/rbumm/Slicer/edit/master/Docs/developer_guide/modules/segmenteditor.md

Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorSmoothingEffect.py
```diff
@@ -78,6 +78,11 @@ def setupOptionsFrame(self):
     self.jointTaubinSmoothingFactorSlider.pageStep = 0.1
     self.jointTaubinSmoothingFactorLabel = self.scriptedEffect.addLabeledOptionsWidget("Smoothing factor:", self.jointTaubinSmoothingFactorSlider)
 
+    self.applyToAllSegmentsCheckBox = qt.QCheckBox()
+    self.applyToAllSegmentsCheckBox.setToolTip("Apply smoothing effect to all visible segments in this segmentation node. This operation may take a while.")
+    self.applyToAllSegmentsCheckBox.objectName = self.__class__.__name__ + 'ApplyToAllVisibleSegments'
+    self.applyToAllSegmentsMMLabel = self.scriptedEffect.addLabeledOptionsWidget("Apply to all segments:", self.applyToAllSegmentsCheckBox)
```
lassoan, line 84: Remove MM. It refers to millimeter (physical distance unit) - therefore it is not relevant for this variable's name.

lassoan (approved):
This works great, thank you! I'll squash and merge this now.

lassoan:
Thanks a lot for your contribution! The new feature will be available in the Slicer Preview Release tomorrow.

It would be awesome if you could do the same for Scissors, Margin, and Hollow effects as well.

rbumm:
Great, will do the same implementation for Margin and Hollow soon. 

@lassoan concerning scissors I am a bit unsure how to implement this mechanism because we do not have an Apply button here. Don't we need the mouse and cut individual segments ?  Any idea here ?

lassoan:
Same for Scissors. Instead of "apply" button the processing is applied when the mouse button is released. There are combination of options that do not make much sense (such as fill mode with "overwrite all" masking mode) and we could warn the user when he chooses such combinations, but I don't think it is necessary.

rbumm:
Ok. In Hollow, Margins and Scissor effect: Processing for visible segments only?

lassoan:
Yes, visible segments only. This allows the user to easily select a subset of segments. Also, editing of invisible segments should be generally avoided because you don't see the results.

rbumm:
Still do not fully understand the scissor plan: You normally select a segment and cut away / through this segment. 
Using the test spine from PW: If you cut or trim one specific vertebra,  you want to be able to make identical cuts to the other vertebra?

PR #5723, opened 2021-07-04
BUG: Revert to previous markups placement behavior

### LINKED ISSUE #5713: Unable to place markups points while moving mouse
I've determined that as of https://github.com/Slicer/Slicer/commit/6b8566e6384b34d86c353a3a2cd93c873673018a, I'm unable to place markups points while moving the mouse which is when I'm placing points quickly. I first observed this when using the Surface Cut segment editor effect.

cc: @lassoan 

## Steps to reproduce
1. Open Slicer
2. Set placement mode to persistent
3. Move your mouse while trying to click to place points. You will observe that the points don't get placed.
4. Stop moving your mouse then click to place a point. The point will be placed.

https://user-images.githubusercontent.com/15837524/123838262-cfc18080-d8d9-11eb-998b-45273e885faa.mp4

## Expected behavior

Points should be placed even when moving the mouse quickly and clicking.

## Environment
- Slicer version: 3D Slicer 4.13.0-2021-06-25 (http://slicer.kitware.com/midas3/api/rest?method=midas.bitstream.download&name=Slicer-4.13.0-2021-06-25-win-amd64.exe&checksum=461ec99a00bdc963b62e636efdfba3ed)
- Operating system: Windows

lassoan:
For a long time, markup control points were placed on mouse button press. A week ago we changed the implementation to place a point on click (no mouse move between button press and release), which several users did not like because it required the users to pause while placing points (instead of keep placing points while moving the mouse). See more details here: https://discourse.slicer.org/t/feedback-requested-requirements-to-place-a-fiducial/18407

The only difference is that now point position is finalized on button release (instead of button press) to be consistent with drawing behavior of other software (Microsoft Office, Inkscape, Gimp, etc.).

fixes #5713

### DIFF
--- a/Libs/MRML/DisplayableManager/vtkMRMLAbstractWidget.cxx
+++ b/Libs/MRML/DisplayableManager/vtkMRMLAbstractWidget.cxx
@@ -456,3 +456,35 @@ vtkMRMLInteractionNode* vtkMRMLAbstractWidget::GetInteractionNode()
     }
   return viewNode->GetInteractionNode();
 }
+
+//---------------------------------------------------------------------------
+int vtkMRMLAbstractWidget::ProcessButtonClickEvent(vtkMRMLInteractionEventData* eventData)
+{
+  if (eventData->GetMouseMovedSinceButtonDown())
+    {
+    return false;
+    }
+
+  int clickEvent = 0;
+  switch (eventData->GetType())
+    {
+    case vtkCommand::LeftButtonReleaseEvent:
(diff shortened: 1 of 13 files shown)


jamesobutler (commented):
Thanks for working on this! Would you like me to build this branch to review before you merge? Or do you plan on merging and let others review with the next day Slicer preview build? Our July 4th holiday is over so I’m looking to try this new version to then update my group’s Slicer version to the latest with this change.

lassoan:
OK, I merge this now so that everyone can test it more easily.

PR #5728, opened 2021-07-07
BUG: Fix the check for 3D while tool logic is 2D in levelTracing tool

sedghi:
LevelTracing is implemented in 2D and should work on the first and last slicer; however, there is a check that doesn't allow this for when only one slice is available (drag and drop of one dicom file)

For context see: https://discourse.slicer.org/t/level-tracing-on-a-single-slice-from-a-3d-volume/16693/4

### DIFF
--- a/Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorLevelTracingEffect.py
+++ b/Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorLevelTracingEffect.py
@@ -163,11 +163,6 @@ class LevelTracingPipeline:
     ijk = self.effect.xyToIjk(xy, self.sliceWidget, masterImageData, parentTransformNode)
     dimensions = masterImageData.GetDimensions()
 
-    for index in range(3):
-      # TracingFilter crashes if it receives a seed point at the edge of the image,
-      # so only accept the point if it is inside the image and is at least one pixel away from the edge
-      if ijk[index] < 1 or ijk[index] >= dimensions[index]-1:
-        return
     self.tracingFilter.SetInputData(masterImageData)
     self.tracingFilter.SetSeed(ijk)
 
--- a/Libs/vtkITK/itkLevelTracingImageFilter.txx
+++ b/Libs/vtkITK/itkLevelTracingImageFilter.txx
@@ -338,11 +338,14 @@ LevelTracingImageFilter<TInputImage,TOutputImage>
     offsetY = neighbors[zeroIndex][1];
     pixTemp[0] = pix[0] + offsetX;
     pixTemp[1] = pix[1] + offsetY;
-    val = inputImage->GetPixel(pixTemp);
-    if(val < threshold)
+    if(region.IsInside(pixTemp))
       {
-      found = true;
-      break;
+      val = inputImage->GetPixel(pixTemp);
+      if(val < threshold)
+        {
+        found = true;
+        break;
+        }
       }
     }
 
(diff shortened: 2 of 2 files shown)


Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorLevelTracingEffect.py
```diff
@@ -163,11 +163,14 @@ def preview(self,xy):
...
+    # LevelTracing implementation is 2D, the next set of lines should be uncommented when
+    # 3D LevelTracing is implemented 
+    
+    # for index in range(3):
       # TracingFilter crashes if it receives a seed point at the edge of the image,
       # so only accept the point if it is inside the image and is at least one pixel away from the edge
-      if ijk[index] < 1 or ijk[index] >= dimensions[index]-1:
-        return
+      # if ijk[index] < 1 or ijk[index] >= dimensions[index]-1:
+        # return
```
pieper, line 173: I guess you confirmed it doesn't crash in this case?  Can you update the comment and remove the dead code?

pieper (changes requested):

Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorLevelTracingEffect.py
lassoan: The check can only be removed along the slice normal axis, but the check must remain along the two other axes. Unless you find evidence that the crash is fixed (there is a relevant code change and your own tests confirm that you can click on boundary pixels and there is no crash).

sedghi: This is not built and tested, I might have misunderstood @lassoan in the discourse forum. I thought commenting these lines would be the solution and everyone will be happy

sedghi: I will build and test now

lassoan: Thanks for working on this @sedghi.

The test is needed in the slice plane (unless you prove otherwise), but not outside the slice plane.

No need to build, you can just change the .py file and restart Slicer.

lassoan: I've checked what I wrote in my commend on discourse. I meant that you can simply remove the check *on your own computer* - because you know that you must not click at boundary voxels. But in Slicer core this workaround is not good enough because users will not know that they will make the application crash by clicking at boundary voxels.

sedghi:
@lassoan @pieper 

I tested the build with the check commented out and the video is available below. It seems to be running without any error

The data used for rendering is not PHI, it is from tcia-mamo public dataset and [idc](https://viewer.imaging.datacommons.cancer.gov/viewer/1.2.826.0.1.3680043.8.498.29634459996408214234588479670784657165)

https://user-images.githubusercontent.com/7490180/125015762-2186ab00-e03e-11eb-8c71-8998bc44563d.mp4

Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorLevelTracingEffect.py
```diff
@@ -163,11 +163,9 @@ def preview(self,xy):
     ijk = self.effect.xyToIjk(xy, self.sliceWidget, masterImageData, parentTransformNode)
     dimensions = masterImageData.GetDimensions()
 
-    for index in range(3):
-      # TracingFilter crashes if it receives a seed point at the edge of the image,
-      # so only accept the point if it is inside the image and is at least one pixel away from the edge
-      if ijk[index] < 1 or ijk[index] >= dimensions[index]-1:
-        return
+    # Current LevelTracing implementation is 2D. In case of 3D implementation in future,
```
lassoan, line 166: I've tested this and without the check the filter makes the application crash (using CTChest sample data set, red slice view, when I move the mouse at the anterior boundary of the image near the). However, it does not seem to be hard to add a check in the filter to prevent the crash. I'll have a look.

lassoan:
I've added bounds check inside the level trace filter. Now there is no need to check the seed position in the Segment Editor effect.

Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorLevelTracingEffect.py
sedghi: It is weird, I tried on many cases and it didn't crash. Thanks for looking deep into the check and having this merged

Issue #5336, opened 2020-12-05
RedGreenBlue color node is not usable for volumes


lassoan:
When choosing a procedural color node as a color node for a scalar volume then the values are not mapped correctly.

## Steps to reproduce

- Load MRHead
- Choose RedGreenBlue as color node
- Adjust window/level => ERROR: no matter what you do, red color will not be used
- Create a copy of RedGreenBlue, edit values
- Colors corresponding to X values do not match actual values

It seems that 0-255 range of the color transfer function is mapped to the intensity window specified by volume display node's window width/level.

## Expected behavior

Probably the best behavior would be to use the color transfer function for direct mapping (ignore window/level). This would be expected, especially when the user specifies x->RGB mapping using a color transfer function.

If that would be too much of a trouble for some reason then at least the behavior should be changed so that the entire scalar range of the colormap would be used by default (e.g., red, green, and blue colors for RedGreenBlue color node).

## Environment
- Slicer version: Slicer-4.11.0-2020-09-30
- Operating system: Windows 10

lassoan:
This has been reported by another user: https://discourse.slicer.org/t/unexpected-behavior-of-color-when-using-lut-with-color-count-255/18542/2

Sunderlandkyl:
Ok, I'll take a look at it.

lassoan:
@Sunderlandkyl Thanks for working on this. Could you look into vtkMRMLScalarVolumeDisplayNode if we could skip the MapToWindowLevelColors filter (connect image directly to MapToColors input; or maybe set window/level values so that the filter does not change the voxel values) when "direct mapping" is enabled?

![image](https://user-images.githubusercontent.com/307929/125165227-1cbc1700-e164-11eb-9684-68df12932f3b.png)

We could use the ScalarRangeFlag in vtkMRMLDisplayNode to indicate that the color mapping should be used as is (e.g., if it is set to UseDirectMapping then it would skip the window/level mapping, otherwise it would keep the current pipeline).

lassoan:
Fixed in https://github.com/Slicer/Slicer/pull/5748/commits/febdf4f7fd3c348bbfee99a08130be5507e1a3bb. There is no GUI for activating direct mapping option (when voxel values are assigned to colors directly by the color node), but it can be enabled by setting ScalarRangeFlag in the volume's display node to vtkMRMLDisplayNode::UseDirectMapping.

PR #5731, opened 2021-07-09
BUG: Fix RedGreenBlue color node range

Sunderlandkyl:
The scalar range of the RedGreenBlue color table was previously set to (-6.0, 6.0), which caused problems when trying to render volumes that were mapped to (0, 255).
Fixed by changing the scalar range of the RedGreenBlue volume to (0.0, 255.0).

Re #5336


Sunderlandkyl:
The only other color node currently in the core that might have this issue is RandomIntegers, which has a scalar range of (VTK_INT_MIN, VTK_INT_MAX).

lassoan:
The problem is not with this color table. It works on models as is, and it should not be necessary to change. The issue is how it is applied to volumes: that instead of applying the values directly to voxel values, the 0-255 range of the colormap is used for coloring the displayed range of the volume.

lassoan:
I'm closing this PR because probably we'll not want to change the colormap but how it is used in the slice display pipeline.

PR #5735, opened 2021-07-10
BUG: Show error popup if loading via command-line has failed

lassoan:
When "Open with..." function was used on macOS (or the filename was specified as a command-line argument), then the application did not report any loading errors.


pieper (approved):
LGTM 👍

PR #5736, opened 2021-07-10
ENH: Implementation of a 3D stablizer in 3D view

rbumm:
Problem:  
Rotating 3D models in 3D view with the mouse may cause tumbling effects and often leads to the impression of inaccuracy. Apart from this, permanent corrections are exhausting. 

Solution:  
Pressing the "a" on the keyboard now toggles between "limited rotation mode" (Azimuth only) and "full rotation mode" (all axes). The result is a smooth and reproducible rotation "on demand".  
  
Keys:  
"a" -> toggle 3D stabilizer

To do:

Indication to the user that stabilizer is active ? pls suggest where

If accepted-> documentation

### DIFF
--- a/Libs/MRML/DisplayableManager/vtkMRMLCameraWidget.cxx
+++ b/Libs/MRML/DisplayableManager/vtkMRMLCameraWidget.cxx
@@ -42,6 +42,7 @@ vtkMRMLCameraWidget::vtkMRMLCameraWidget()
 {
   this->MotionFactor = 10.0;
   this->MouseWheelMotionFactor = 1.0;
+  this->CameraTiltLocked = false;
 
   this->StartEventPosition[0] = 0.0;
   this->StartEventPosition[1] = 0.0;
@@ -94,6 +95,9 @@ vtkMRMLCameraWidget::vtkMRMLCameraWidget()
 
   this->SetKeyboardEventTranslation(WidgetStateIdle, vtkEvent::NoModifier, 0, 0, "plus", WidgetEventCameraZoomIn);
   this->SetKeyboardEventTranslation(WidgetStateIdle, vtkEvent::NoModifier, 0, 0, "minus", WidgetEventCameraZoomOut);
+
+  // Toggle tiltLock
+  this->SetKeyboardEventTranslation(WidgetStateIdle, vtkEvent::ControlModifier, 0, 0, "b", WidgetEventToggleCameraTiltLock);
 
   // Reset camera
 
@@ -300,6 +304,10 @@ bool vtkMRMLCameraWidget::ProcessInteractionEvent(vtkMRMLInteractionEventData* e
       this->SaveStateForUndo();
       this->CameraNode->RotateAround(vtkMRMLCameraNode::S, true);
       break;
+
+    case WidgetEventToggleCameraTiltLock:
+        this->CameraTiltLocked = !this->CameraTiltLocked;
+        break;
 
     case WidgetEventCameraReset:
       this->SaveStateForUndo();
@@ -610,8 +618,15 @@ bool vtkMRMLCameraWidget::ProcessRotate(vtkMRMLInteractionEventData* eventData)
 
   bool wasCameraNodeModified = this->CameraModifyStart();
 
-  camera->Azimuth(rxf);
-  camera->Elevation(ryf);
+  if (this->CameraTiltLocked == true)
+    {
+    camera->Azimuth(rxf);
+    }
+  else
+    {
+    camera->Azimuth(rxf);
+    camera->Elevation(ryf);
+    }
   camera->OrthogonalizeViewUp();
 
   this->CameraModifyEnd(wasCameraNodeModified, true, true);
--- a/Libs/MRML/DisplayableManager/vtkMRMLCameraWidget.h
+++ b/Libs/MRML/DisplayableManager/vtkMRMLCameraWidget.h
@@ -108,6 +108,8 @@ public:
     WidgetEventCameraWheelZoomIn, // same as WidgetEventCameraZoomIn but with using wheel scaling factor
     WidgetEventCameraWheelZoomOut,
 
+    WidgetEventToggleCameraTiltLock,
+
     WidgetEventCameraReset,
     WidgetEventCameraResetTranslation,
     WidgetEventCameraResetRotation,
@@ -164,6 +166,7 @@ protected:
 
   double MotionFactor;
   double MouseWheelMotionFactor;
+  bool CameraTiltLocked;
 
   vtkWeakPointer<vtkMRMLCameraNode> CameraNode;
 
--- a/Docs/user_guide/user_interface.md
+++ b/Docs/user_guide/user_interface.md
@@ -131,6 +131,7 @@ Default orientation axes: A = anterior, P = posterior, R = right, L = left, S = 
 - **Spin** continuously spins the view around.
 - **Rock** continuously rocks the view left-to-right.
 - **Zoom in/out** slightly zooms in/out the view. Convenient buttons for touchscreens.
+- **Tilt Lock** can be toggled using `Ctrl` + `b` keyboard shortcut. In tilt lock mode 3D view rotation is restricted to the azimuth axis (left-right direction) by disabling rotation around elevation axis (up-down direction).
 
 ### Slice View
 
@@ -224,6 +225,7 @@ view will not activate the view.
 | `Shift` + `Home` or `Shift` + `Keypad 7`| rotate to view from inferior |
 | `right-click` + `drag up/down` | zoom view in/out |
 | `Ctrl` + `mouse wheel` | zoom view in/out |
+| `Ctrl` + `b` | toggle tilt lock |
 | `+` / `-` | zoom view in/out |
 | `middle-click` + `drag` | pan (translate) view |
 | `Shift` + `left-click` + `drag` | pan (translate) view |


pieper:
The stabilizing mode looks simple and elegant.  The 's' hotkey is already used by the native vtk interactor along 'w' to go between wireframe and shaded mode, and that's sometimes useful for debugging so we might pick a different key.

I'm reluctant to add the other hotkey additions though as it seems they are duplicates of existing hotkeys (pageup, pagedown, home, end, etc).  It's better to leave them open for use by widgets or module.

Another nice-to-have would be to make these hotkeys discoverable in the app, perhaps by adding a list AxesWidget.

rbumm:
Then I would vote for "a" (axes), present on most of the keyboards at the identical spot ? 

I can take out the other keys, no problem.

pieper:
> Then I would vote for "a" (axes), present on most of the keyboards at the identical spot ?

Yes, I think 'a' would be good, thanks

jamesobutler:
I'm not familiar enough with how hotkeys work in context of other events. Currently the "a" key is a shortcut to apply a segment with the "Draw" Segment Editor effect. Does it know if you are meaning to apply the segment, or meaning to toggle rotation mode? Does it depend on which view was last clicked (red slice view or 3D View)?

rbumm:
It depends on the view you last clicked and only toggles if you clicked the 3D view. I just tested Draw, the "a" applied the effect and did not toggle the rotation mode in 3D.  Only if you clicked the 3D view the effect was toggled.

lassoan:
Keyboard shortcuts are always conflicting and non-discoverable and if observed in the render window then theu only work if you click in the window first. We can use them, but rather see this as a drop-down menu in the View/transform mouse mode button and as a right-click menu option. We could also define handles on the purple box so that if you grab one edge of the box then you rotate around the corresponding axis.

pieper:
Agreed, we should make all the interface actions discoverable and visible also be more explicit to indicate focus.  For now Rudolf's change is small and consistent with existing behavior since he finds it useful I'd vote to go ahead.

rbumm:
I found the drop down menu too clicky and time consuming, but the right click option should be in anyway, to show the user something has changed. Should I add it ?

rbumm:
Shall I add anything more, Steve?

pieper (dismissed):
LGTM 👍 

Does anyone else have comments before merging?

lassoan:
don't merge yet, I'm adding comments now

Libs/MRML/DisplayableManager/vtkMRMLCameraWidget.cxx
```diff
@@ -322,6 +326,10 @@ bool vtkMRMLCameraWidget::ProcessInteractionEvent(vtkMRMLInteractionEventData* e
       this->SaveStateForUndo();
       this->Dolly(0.8);
       break;
+    case WidgetEventToggleStabilizer:
```
lassoan, line 329: "Stabilizer" could mean many things. For me, it means camera stabilization (removing shaking), and we already have a transform stabilizer module in SlicerIGT for this purpose.

It should be called something like `WidgetEventToggleCameraTiltLock`.

Libs/MRML/DisplayableManager/vtkMRMLCameraWidget.cxx
```diff
@@ -610,8 +618,13 @@ bool vtkMRMLCameraWidget::ProcessRotate(vtkMRMLInteractionEventData* eventData)
 
   bool wasCameraNodeModified = this->CameraModifyStart();
 
-  camera->Azimuth(rxf);
-  camera->Elevation(ryf);
+  if (this->stabilize3D) {
```
lassoan, line 621: There are axis-preferred rotation modes (when you cannot tilt by more than 90 degrees in a selected axis), which I think would work better than such a rigid locking. Try to load a mesh file in "3D Viewer" built-in Windows application to see how that works.

The current implementation is very simple, though, so it is OK to keep it as is, the improvement is just for considering it in the future.

Libs/MRML/DisplayableManager/vtkMRMLCameraWidget.cxx
```diff
@@ -610,8 +618,13 @@ bool vtkMRMLCameraWidget::ProcessRotate(vtkMRMLInteractionEventData* eventData)
...
+      camera->Azimuth(rxf);
+  }
+  else {
```
lassoan, line 624: need to fix the brace indentation style

Libs/MRML/DisplayableManager/vtkMRMLCameraWidget.h
```diff
@@ -164,6 +166,7 @@ class VTK_MRML_DISPLAYABLEMANAGER_EXPORT vtkMRMLCameraWidget : public vtkMRMLAbs
 
   double MotionFactor;
   double MouseWheelMotionFactor;
+  bool stabilize3D = false;
```
lassoan, line 169:
```suggestion
  bool CameraTiltLocked = false;
```

lassoan, line 169: Initialization of variables are inconsistent. Either initialize all the other variables here or initialize all the variables in the constructor.

Libs/MRML/DisplayableManager/vtkMRMLCameraWidget.cxx
```diff
@@ -322,6 +326,10 @@ bool vtkMRMLCameraWidget::ProcessInteractionEvent(vtkMRMLInteractionEventData* e
       this->SaveStateForUndo();
       this->Dolly(0.8);
       break;
+    case WidgetEventToggleStabilizer:
+        if (this->stabilize3D) this->stabilize3D = false;
```
lassoan, line 330: Replace it by:

this->stabilize3D = !this->stabilize3D;

Libs/MRML/DisplayableManager/vtkMRMLCameraWidget.cxx
```diff
@@ -95,6 +95,10 @@ vtkMRMLCameraWidget::vtkMRMLCameraWidget()
   this->SetKeyboardEventTranslation(WidgetStateIdle, vtkEvent::NoModifier, 0, 0, "plus", WidgetEventCameraZoomIn);
   this->SetKeyboardEventTranslation(WidgetStateIdle, vtkEvent::NoModifier, 0, 0, "minus", WidgetEventCameraZoomOut);
 
+  // Toggle stabilizer
+
+  this->SetKeyboardEventTranslation(WidgetStateIdle, vtkEvent::NoModifier, 0, 0, "a", WidgetEventToggleStabilizer);
```
lassoan, line 100: `a` keyboard shortcut without modifier is already used in several extensions and this may disrupt those. For example, for moving devices in SlicerHeart we use the WASD and QE keys (with Shift for fine adjustment). It would be also too easy to accidentally hit the `a` key and then users would have no idea why their 3D view got permanently broken (until an application restart).

It would be better to use a shortcut with a modifier, something that is close to the cursor arrow keys (which rotate the 3D view). For example, Ctrl+`,` or Ctrl+`.` or Ctrl+`/`.

lassoan (commented):
It is not a full solution (no GUI, not discoverable, state not saved in scene, not synchronized with linked views, it is a rigid lock, etc.), but it is a small, isolated change that is not likely to cause any trouble if the shortcut uses a Ctrl or Alt modifier.

I've added a couple of comments, all of them should be easy to fix.

Libs/MRML/DisplayableManager/vtkMRMLCameraWidget.cxx
rbumm: Will check this out.

Libs/MRML/DisplayableManager/vtkMRMLCameraWidget.h
rbumm: Initialized in the constructor now.

rbumm: renamed it this way

rbumm:
Made the changes already, works locally, but it will take a while until I will be able to push them. Thanks for the helpful comments.

rbumm:
I did the changes, but am quite sure I did something wrong when I merged and pushed this branch again to my rep. 

At least, my recent TiltLock additions seem to be in now.  Are you able to merge that ? Sorry!

lassoan:
Thank you, it looks good, except Ctrl-x is the standard shortcut for "cut" operation (see complete list of standard shortcuts on various platforms [here](https://doc.qt.io/qt-5/qkeysequence.html#standard-shortcuts)).

Most of the convenient left-hand-side keyboard shortcuts are already taken, but you could use Ctrl-`e`, `b`, `r` on the left or `m`, `,`, `.`, `/`, `j`, `k`, `l`, `;`, `'`, `u`, `i`, `\` on the right, as they are not commonly used and I think they are not used in Slicer either.

rbumm:
> Thank you, it looks good, except Ctrl-x is the standard shortcut for "cut" operation (see complete list of standard shortcuts on various platforms [here](https://doc.qt.io/qt-5/qkeysequence.html#standard-shortcuts)).
> 
> Most of the convenient left-hand-side keyboard shortcuts are already taken, but you could use Ctrl-`e`, `b`, `r` on the left or `m`, `,`, `.`, `/`, `j`, `k`, `l`, `;`, `'`, `u`, `i`, `\` on the right, as they are not commonly used and I think they are not used in Slicer either.

I changed the hotkey to Ctrl+"b".

lassoan:
Merged. Thank you @rbumm for your contribution!

PR #5738, opened 2021-07-11
ENH: Use consistent spelling of "Color" instead of "Colour" 

jamesobutler:
This change was motivated after seeing different spellings within the same error message or comment. The American spelling of "Color" was chosen to match already existing node and class names like "vtkMRMLColorTableNode".

`"vtkMRMLColorTableNode::SetColor: ERROR: can't set a colour if not a user defined colour table`
is now
`"vtkMRMLColorTableNode::SetColor: ERROR: can't set a color if not a user defined color table`

I could also address "Grey" -> "Gray" if desired as well. Currently the default color node ID is "vtkMRMLColorTableNodeGrey", but would become "vtkMRMLColorTableNodeGray" if this is determined to be an appropriate change. 
- Some documentation details about `Change color mapping from grayscale`
- itkGrayscaleGrindPeakImageFilter
- vtkMRMLColors::sliceGray()
- vtkImageRectangularSource::vtkSetMacro(InsideGraySlopeFlag,int);


pieper:
Thanks for doing this.   Yes, we should be consistent with [rule 4 of our style guidelines](https://www.slicer.org/wiki/Documentation/Nightly/Developers/Style_Guide#Naming).

While I'd generally prefer gray over grey, I don't think it's a big enough issue to justify breaking backward compatibility.

Modules/Loadable/Annotations/MRML/vtkMRMLAnnotationAngleNode.cxx
```diff
@@ -381,7 +381,7 @@ int vtkMRMLAnnotationAngleNode::SetControlPoint(int id, double newControl[3])
 }
 
 //---------------------------------------------------------------------------
-double* vtkMRMLAnnotationAngleNode::GetPointColour()
+double* vtkMRMLAnnotationAngleNode::GetPointColor()
```
lassoan, line 384: Annotations will be gone, so it would be better to just leave the API as is (it would be frustrating for developers to update their code to according to the change and then see the classes completely removed soon after).

lassoan (commented):
Thank you, the changes look good to me, except I would not touch the Annotations classes, as they will be deprecated soon and developers should switch to Markups instead.

lassoan:
It would make sense to replace grey by gray in comments and private variable names. In public members, `...Gray...` methods could be added and `...Grey...` methods could be deprecated (they would still work but log a warning). In a few years, `...Grey...` methods could be removed.

Modules/Loadable/Annotations/MRML/vtkMRMLAnnotationAngleNode.cxx
jamesobutler: What if I deprecate methods such as `GetPointColour` and call to `GetPointColor`?

If users are to be using Markups rather than Annotations, that deprecation should be indicated to them now.

lassoan: We should bug the users to move on to use Markups and not to spend any more time with Annotations.

Deprecation of Annotations module is planned in several steps - see the road map at https://github.com/Slicer/Slicer/wiki/Roadmap

jamesobutler: It appears that the "Deprecate Annotations module - April 2021" has been completed, but there hasn't been anything done in the code to indicate the deprecation?

lassoan: Correct. We have made the decision and told users and developers when the topic came up, but we haven't updated the documentation or added warnings in the code and user interface yet.

lassoan:
There are a lot of files and so there will be merge conflicts if we keep this PR open for too long.

Is there any objection to merging it except the public method name changes (that are in obsolete Annotations module anyway)?

Modules/Loadable/Annotations/MRML/vtkMRMLAnnotationAngleNode.cxx
jcfr: Adding an entry to the migration guide would be sufficient: https://www.slicer.org/wiki/Documentation/Nightly/Developers/Tutorials/MigrationGuide#Slicer_5.0:_API_changes_since_4.10

Moving forward with the integration makes sense.

It may also be sensible to update the spellchecker to consider the words   `colour` and `behaviour` was "invalid"

jcfr (approved):
Makes sense to move forward. 

@jamesobutler  When you have a chance, let me now what you think of: https://github.com/Slicer/Slicer/pull/5738#discussion_r670810765

Modules/Loadable/Annotations/MRML/vtkMRMLAnnotationAngleNode.cxx
jamesobutler: It does appear that codespell has a en_GB->en_US  dictionary so it may support setting a language to en_US. I'll investigate

PR #5746, opened 2021-07-16
DOC: Remove duplicate linking to DICOM module documentation

jamesobutler:
I just noticed this and figured I would go ahead and submit a change.

I removed a link that was linking to the Slicer Wiki DICOM module page (https://www.slicer.org/wiki/Documentation/Nightly/Modules/DICOM) since that page is now telling users to go to the DICOM module ReadTheDocs page (https://slicer.readthedocs.io/en/latest/user_guide/modules/dicom.html) and that is already linked to users in a sentence a few lines down.


lassoan (approved):
Looks good to me, thank you.

PR #5748, opened 2021-07-16
ENH: Update scalar volume display to use full color node scalar range

Sunderlandkyl:
Within vtkMRMLScalarVolumeDisplayNode, scalar values are mapped using window/level to a unsigned char scalar range of 0 to 255.
While most color nodes in the scene use a range of 0 to 255, some nodes use different ranges (ex. RedGreenBlue uses -6.0 to 6.0).

In order to make full use of all colors in the table, we now copy the color node lookup table, and map it to the expected 0 to 255 scalar range from vtkImageMapToWindowLevelColors.

This commit also adds support for using the scalar range flag of UseDirectMapping to map the values of the image directly onto the scalar range of the table, bypassing the window/level.

### DIFF
--- a/Libs/MRML/Core/vtkMRMLScalarVolumeDisplayNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLScalarVolumeDisplayNode.cxx
@@ -73,7 +73,8 @@ vtkMRMLScalarVolumeDisplayNode::vtkMRMLScalarVolumeDisplayNode()
   this->MapToWindowLevelColors->SetLevel(128.);
 
   this->MapToColors->SetOutputFormatToRGBA();
-  this->MapToColors->SetInputConnection(this->MapToWindowLevelColors->GetOutputPort() );
+  // This input may be changed later if ScalarRangeFlag is modified.
+  this->MapToColors->SetInputConnection(this->MapToWindowLevelColors->GetOutputPort());
 
   this->ExtractRGB->SetInputConnection(this->MapToColors->GetOutputPort());
   this->ExtractRGB->SetComponents(0,1,2);
@@ -169,6 +170,20 @@ void vtkMRMLScalarVolumeDisplayNode
 ::SetInputToImageDataPipeline(vtkAlgorithmOutput *imageDataConnection)
 {
   this->Threshold->SetInputConnection(imageDataConnection);
+
+  if (this->GetScalarRangeFlag() == vtkMRMLDisplayNode::UseDirectMapping)
+    {
+    // Bypass window/level to show table scalar range directly
+    this->MapToColors->SetInputConnection(imageDataConnection);
+    }
+  else
+    {
+    // Reconnection is expensive operation, therefore only do it if needed
+    if (this->MapToColors->GetInputConnection(0, 0) != this->MapToWindowLevelColors->GetOutputPort())
+      {
+      this->MapToColors->SetInputConnection(this->MapToWindowLevelColors->GetOutputPort());
+      }
+    }
   this->MapToWindowLevelColors->SetInputConnection(imageDataConnection);
 }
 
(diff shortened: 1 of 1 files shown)


Libs/MRML/Core/vtkMRMLScalarVolumeDisplayNode.cxx
```diff
@@ -563,6 +573,16 @@ void vtkMRMLScalarVolumeDisplayNode::UpdateLookupTable(vtkMRMLColorNode* newColo
         }
       }
     }
+
+  if (lookupTable && this->GetScalarRangeFlag() != vtkMRMLDisplayNode::UseDirectMapping)
```
lassoan, line 577: We should only copy if the range is not already (0,255)

Libs/MRML/Core/vtkMRMLScalarVolumeDisplayNode.cxx
```diff
@@ -169,6 +169,16 @@ void vtkMRMLScalarVolumeDisplayNode
 ::SetInputToImageDataPipeline(vtkAlgorithmOutput *imageDataConnection)
 {
   this->Threshold->SetInputConnection(imageDataConnection);
+
+  if (this->GetScalarRangeFlag() == vtkMRMLDisplayNode::UseDirectMapping)
+    {
+    // Bypass window/level to show table scalar range directly
+    this->MapToColors->SetInputConnection(imageDataConnection);
```
lassoan, line 176: Reconnection can be an expensive operation. Check the current input and only set the input again the the input is actually different (see for example in vtkMRMLVolumeRenderingDisplayableManager.cxx).

Libs/MRML/Core/vtkMRMLScalarVolumeDisplayNode.cxx
```diff
@@ -73,7 +73,7 @@ vtkMRMLScalarVolumeDisplayNode::vtkMRMLScalarVolumeDisplayNode()
   this->MapToWindowLevelColors->SetLevel(128.);
 
   this->MapToColors->SetOutputFormatToRGBA();
-  this->MapToColors->SetInputConnection(this->MapToWindowLevelColors->GetOutputPort() );
+  this->MapToColors->SetInputConnection(this->MapToWindowLevelColors->GetOutputPort());
```
lassoan, line 76: Add a comment here that this may be later changed.

lassoan (commented):
Thank you. It is great that you managed to implement all these with such a few, simple changes. I've added a few small comments inline, mostly to ensure that performance is not impacted, even a little bit, for live volume display (e.g., live ultrasound image display).

lassoan:
One more thing: add the related bug number to the commit comment.

Libs/MRML/Core/vtkMRMLScalarVolumeDisplayNode.cxx
Sunderlandkyl: I always assumed that SetInputConnection would check to see if the input is already connected. The function does return at this point if the connection hasn't changed: [vtkAlgorithm.cxx#L1000-L1009](https://github.com/Kitware/VTK/blob/3a01239340d7f02dfe0ff45a1d0d464437e78e75/Common/ExecutionModel/vtkAlgorithm.cxx#L1000-L1009).

Is the check performed by SetInputConnection expensive?

lassoan: Yes, I think those calls to before the method return propagate information request throughout the entire pipeline. As far as I remember, it made a few fps difference in liver ultrasound display. It would be useful if you could check with a profiler if the check is still needed, as things may have changed since I tested this a few years ago.

Sunderlandkyl: I didn't see any significant difference between the two options under RelWithDebInfo. In any case, I've added a check to make sure that the input is different.

Sunderlandkyl:
@lassoan Comments should be addressed.

lassoan (approved):
Thanks for the updates, looks good.

PR #5749, opened 2021-07-21
ENH: Add public method to change threshold and WindowLevel bounds

jamesobutler:
This is a small change to add some public methods for convenience to manipulate the bounds of Volume Threshold and WindowLevel widgets through code rather than just through the GUI.


lassoan (commented):
Thank you, this will be useful (since these are not stored in MRML, there is no other way of storing them). If we add "set" methods then we must also add "get" methods (and make sure they are accessible from Python, preferably by using them in some Python tests).

pieper:
Yes, probably the `Q_PROPERTY` pattern is the best for this as used all over the Slicer and CTK widgets.

jamesobutler:
`setThresholdBounds` was available to the Python interface for me during my testing. Is `Q_PROPERTY` still needed or is this only for adding a property object rather than getting the value from a getter?

pieper:
Yes, the setter is wrapped because it's a slot.  For a `Q_PROPERTY` you need to define the getter but from python you access like a python member variable rather than invoking the getter or setter directly.  It looks like properties are more consistent with the implementation here.

pieper (approved):
Makes sense to me 👍

PR #5750, opened 2021-07-23
DOC: Add programmatic installation section to developer guide 

allemangD:
This script was built to install Slicer with BoneTextureExtension in a container. See https://github.com/DCBIA-OrthoLab/SlicerCMF-docker/pull/2. I've included the same script here, with some modifications.

The added script `slicer-download.sh` downloads and extracts Slicer to a target directory. It accepts parameters to select slicer version and pre-installed extensions; I've documented these in a comment at the top of the script, however I don't have any `--help` text or similar set up. The script depends on `rsync`, `curl`, and `jq`; I'm not sure the best place to document this.

It can also be used to install extensions into an existing installation of Slicer by setting `-d dest` to the Slicer root directory.

The script only supports the Linux `.tar.gz` binaries. It may be possible to convert only the _extension_ installation to be cross-platform; however cross-platform _Slicer_ installation would be significantly more complex.

### DIFF
--- a/Docs/developer_guide/script_repository/gui.md
+++ b/Docs/developer_guide/script_repository/gui.md
@@ -1,3 +1,10 @@
+## Install Slicer
+
+There are different approaches to install Slicer and extensions programmatically:
+
+- Install Slicer manually and install extensions by using `slicer.app.extensionsManagerModel()`. See example [below](script_repository.md#download-and-install-extension) and in [install-slicer-extension.py](https://github.com/pieper/SlicerDockers/blob/master/slicer-plus/install-slicer-extension.py)
+- Directly interact with the REST API endpoints of https://slicer-packages.kitware.com using `curl` and `jq`. See [slicer-download.sh](https://github.com/Slicer/SlicerDocker/blob/master/scripts/slicer-download.sh)
+
 ## Launch Slicer
 
 ### Open a file with Slicer at the command line


Utilities/Scripts/slicer-download.sh
```diff
@@ -0,0 +1,107 @@
...
+
+# args: [-v release_id] [-e extension] [-d dest] [-s]
+# Installs extensions into slicer at dest. multiple -e arguments are accepted; each
+#   extension will be installed.
+# If -s is present, slicer itself is installed at dest.
+# If -v is not present, assume 4.11.20210226 (current slicer stable)
+
+# Usage:
+# Install slicer to ./stable
+#   slicer-dl.sh -v 4.11.20210226 -d stable -s
```
jcfr, line 11:
```suggestion
#   slicer-download.sh -v 4.11.20210226 -d stable -s
```

Utilities/Scripts/slicer-download.sh
```diff
@@ -0,0 +1,107 @@
...
+
+# Add BoneTextureExtension to that installation
+#   slicer-dl.sh -v 4.11.20210226 -d stable -e BoneTextureExtension
```
jcfr, line 14:
```suggestion
#   slicer-download.sh -v 4.11.20210226 -d stable -e BoneTextureExtension
```

Utilities/Scripts/slicer-download.sh
```diff
@@ -0,0 +1,107 @@
...
+
+# Do both steps at once
+#   slicer-dl.sh -v 4.11.20210226 -d stable -s -e BoneTextureExtension
```
jcfr, line 17:
```suggestion
#   slicer-download.sh -v 4.11.20210226 -d stable -s -e BoneTextureExtension
```

Utilities/Scripts/slicer-download.sh
```diff
@@ -0,0 +1,107 @@
+#!/usr/bin/env bash
+
```
jcfr, line 2:
```suggestion

# This script downloads and extracts Slicer and selected extensions to a target directory.
#
```

Utilities/Scripts/slicer-download.sh
```diff
@@ -0,0 +1,107 @@
...
+# Usage:
+# Install slicer to ./stable
+#   slicer-dl.sh -v 4.11.20210226 -d stable -s
+
+# Add BoneTextureExtension to that installation
+#   slicer-dl.sh -v 4.11.20210226 -d stable -e BoneTextureExtension
+
+# Do both steps at once
+#   slicer-dl.sh -v 4.11.20210226 -d stable -s -e BoneTextureExtension
+
```
jcfr, line 18:
```suggestion

set -e
set -o pipefail

```

jcfr (commented):
Thanks for the contribution  :pray: It looks great.

After addressing few nitpick,  it will be ready for integration.

I suggest to update the documentation adding a section to "Install Slicer" to this page:
https://slicer.readthedocs.io/en/latest/developer_guide/script_repository.html

pieper (changes requested):
I suggest adding a `Utilities/Scripts/slicer-download-README.md` with the info from your commit message above.

Also you should add `jq` to the list of dependencies.

allemangD:
> I suggest to update the documentation adding a section to "Install Slicer" to this page:
> https://slicer.readthedocs.io/en/latest/developer_guide/script_repository.html

Would it make more sense to move the script to the "Getting Started" page of the docs? I can see an argument for either, as the script is more intended for developer use, in containers, etc. I've added a short section about the different installers in the developer guide as you suggested; see 296d6e4.

> Also you should add `jq` to the list of dependencies.

Good catch. I've added it to the developer documentation and updated the script comments in 296d6e4.

> I suggest adding a `Utilities/Scripts/slicer-download-README.md` with the info from your commit message above.

Since there would already be a section in the slicer documentation and in the comments of the script, I'm not sure it makes sense to duplicate the content a third time in a readme file. There are also no other readme files in Utilities/Scripts.

> It would be great if it could query the server to get the latest release or nightly version automatically.

I haven't found a way through the API to get the latest _nightly_ version; however I was able to get the latest release version as a default. This is done in 528d915.

pieper:
> I'm not sure it makes sense to duplicate the content a third time in a readme file.

No, definitely don't want duplicate information.  How about converting the comment at the top of the script into a usage message that prints if the script is run with no arguments or with -h/--help.

> I haven't found a way through the API to get the latest _nightly_ version;

Maybe @jcfr has a way to do this with the new server infrastructure

Docs/developer_guide/script_repository/gui.md
```diff
@@ -1,3 +1,21 @@
+## Install Slicer
+
+### Graphical Installers
+
+Graphical installers for Windows and Mac are available at https://download.slicer.org/. See more information in {doc}`/user_guide/getting_started`.
+
+There are also Linux binaries available as `.tar.gz` archives; extract the archive and launch the contained `Slicer` executable.
+
+### Command-line installer
```
jcfr, line 9:
```suggestion
Install Slicer and extensions using the downloader script
```

jcfr:
> I haven't found a way through the API to get the latest nightly version;

Here it is:

```
curl -s 'https://slicer-packages.kitware.com/api/v1/app/5f4474d0e1d8c75dfc705482/draft?limit=1&sort=created&sortdir=-1' | jq ".[0].meta.revision"
```

In a nutshell, list all draft revision sorted by created date and limit the output to 1 (the most recent)

lassoan:
We have a very few bash scripts and one more should not be a problem but I would much prefer to use Python instead. Python can be used on all platforms, you can easily bring in extra features as needed, there is no need to learn yet another syntax, way to debug, document, test, etc.

jcfr:
> but I would much prefer to use Python instead

That would be nice but this go beyond the scope of this effort. 

I suggest to create an issue motivating the change and associating the tag "good first issue"

Docs/developer_guide/script_repository/gui.md
```diff
@@ -1,3 +1,19 @@
+## Install Slicer
+
+### Download installers from the website
+
+Installers are available at https://download.slicer.org/. See detailed instructions at {doc}`/user_guide/getting_started`.
+
+Install Slicer and extensions using the downloader script
```
jcfr, line 7:
```suggestion
### Install Slicer and extensions using the downloader script
```

Docs/developer_guide/script_repository/gui.md
```diff
@@ -1,3 +1,19 @@
...
+
+Installers are available at https://download.slicer.org/. See detailed instructions at {doc}`/user_guide/getting_started`.
+
+Install Slicer and extensions using the downloader script
+
+On Linux, use [`slicer-download.sh`](https://github.com/Slicer/Slicer/blob/master/Utilities/Scripts/slicer-download.sh) to install Slicer and selected extensions to a target directory.
+
+:::{note}
+
+`slicer-download.sh` depends on `curl`, `rsync`, and `jq`; be sure these are available on your system before using the installer.
```
jcfr, line 13:
```suggestion
`slicer-download.sh` depends on `curl`, `rsync`, and `jq`; be sure these are available on your system before using the script.
```

jamesobutler:
Just curious, how often would the generation of this docker be run to request a download of Slicer from the server and request a download of BoneTextureExtension from the server?

allemangD:
@jcfr @lassoan It's also worth noting that, unless there is a way to download archives of the Windows and MacOS releases, the slicer installation in this script can only work with Linux. 

It should be fairly straightforward to make the extension installation portion of the script cross-platform. I agree that Python would be better suited there.

---

@jamesobutler In the SlicerCMF-docker project, this will only be invoked at image build time; I would expect that only to happen during development or deployment; not when scaling.

Even in development, if the build cache isn't invalidated, it should only be invoked when changing the Slicer or extension versions; you can see in the linked PR that I download Slicer separately from the extensions to take advantage of that. https://github.com/allemangD/SlicerCMF-docker/blob/33e9e18a81817a5e2d140e77266218deda0ab4a6/src/Docker/Dockerfile#L27-L31

jcfr:
> how often would the generation of this docker be run to request a download of Slicer from the server and request a download of BoneTextureExtension from the server?

We anticipate the docker image would be updated if any of the associated extensions include new features.

Considering that @allemangD and the @SlicerCMF team are actively working on the addressing issues in the associated extensions, the update would happen fairly often while it stabilizes.

We still need to discuss how to trigger the update of the docker image when extensions are updated

pieper:
Github actions with a build matrix could address making images with extensions pre-installed for all platforms.

jcfr:
> a way to download archives of the Windows and MacOS releases

On macOS, the following could be done:

```
filename=Slicer-4.13.0-2021-07-25-macosx-amd64.dmg
pkg_base="$(basename ${pkg} .dmg)"
vol_name=/Volumes/${pkg_base}
hdiutil attach -mountpoint ${vol_name} ${pkg_base}.dmg
```

On windows, the NSIS installer has options enabling it to be silently extracted in a target directory.

jcfr:
> Github actions with a build matrix could address making images with extensions pre-installed for all platforms.

Indeed, the CI aspect would be straightforward to do. 

That said, the focus should likely be first on linux based docker image as this is the most prevalent use of docker. 

Then, we could focus on having a window base image using one of [these](https://hub.docker.com/_/microsoft-windows-servercore).

And finally, for macOS, one could evaluate [these](https://github.com/sickcodes/Docker-OSX).

lassoan:
Is there a reason for having this script in the Slicer repository? The logic for downloading Slicer and extensions is not related to Slicer versions but to the download infrastructure, so it should not be versioned with the source code of the application but with the download site. Therefore, hosting it in slicer_download (or maybe SlicerDocker) repository would be more appropriate.

I would be also curious of why bash was chosen to implement this. It seems that the whole script could have been implemented in 5-10 lines in Python. It is just so long and complicated because bash is so limited. It could even be a pip-installable Python package, so it could be downloaded and run with two lines of code on any platform (it would take care of its dependencies).

jcfr:
> I would be also curious of why bash was chosen to implement this

Using bash is really easy within docker as it doesn't requite to first install python.

It is true that now docker support multi stage build (see [here](https://docs.docker.com/develop/develop-images/multistage-build/)), a python  image could be used to run the download script and create the target directory.

To summarize:
* we just capitalize on the fact we already add readily available bash script implemented and aimed for simplicity here.
* we spent a little more time to contribute the script and make it available for others.

> Is there a reason for having this script in the Slicer repository?

We considered adding to https://github.com/Slicer/SlicerDocker but this script was of general use, and adding it to the `Utilities/Scripts` folder seems like a natural fit.

Following this discussion, adding to this repo would likely be a better fit ? 

See https://github.com/Slicer/SlicerBuildEnvironment

lassoan:
> Following this discussion, adding to this repo would likely be a better fit ?
> See https://github.com/Slicer/SlicerBuildEnvironment

Yes. It would be better. Any repository that is more about to downloading and installing Slicer packages would be more relevant.
Of the repositories that I know, https://github.com/Slicer/SlicerDocker would be the most useful, as in this repository we already download and install Slicer and extensions and create docker image from them.

Could we consolidate https://github.com/Slicer/SlicerDocker and https://github.com/Slicer/SlicerBuildEnvironment, https://github.com/DCBIA-OrthoLab/SlicerCMF-docker?

SlicerDocker does almost the same as SlicerCMF-docker, but also a few more things, such as installing Python packages (this is useful for extensions that pip-install Python packages), set up display server, remote desktop access, etc. SlicerCMF docker should be based on some generic Slicer docker. This would not just simplify maintenance of the docker images but it would ensure that not only SlicerCMF but other extensions could use the same Slicer base installation.

There is also some redundancy between https://github.com/Slicer/SlicerDocker and https://github.com/Slicer/SlicerBuildEnvironment. Probably we should just remove the unmaintained images (slicer-build, slicer-dependencies, ...).

allemangD:
> SlicerDocker does almost the same as SlicerCMF-docker, but also a few more things [...]. SlicerCMF docker should be based on some generic Slicer docker. 

SlicerCMF-docker is set up that way because (1) we did not want to build Slicer or extensions in docker and (2) we did not want any overhead associated with display server, web server, remote desktop etc. The only things we need access to are the CLI module executables.

I based the work off of @pieper's https://github.com/pieper/slicerdockers; yet another similar repository.

With all that said, it doesn't really matter if Slicer is built in Docker or release binaries are pulled via the `slicer-download.sh`; we just didn't want that computation running on our hardware. Ideally there would be a maintained `Slicer` base image with a Slicer installation, minimal dependencies, and some mechanism to install extensions.

It seems an additional image that adds GUI dependencies to that (display server, remote desktop, etc) would fill the remaining needs.

lassoan:
> Ideally there would be a maintained Slicer base image with a Slicer installation, minimal dependencies, and some mechanism to install extensions.

Fully agree. What we have is too fragmented with lots of similarities and redundancies. Probably we should just have a minimal docker without GUI, then based on that another docker with GUI. Various additional images (CMF, Jupyter Notebook, web server, etc.), could start off either of these two images. These images could also contain helper scripts that would make it easy to install extensions, Python packages, create default .ini file, etc. These images would be built every day.

> Let's discuss on Tuesday developer call. It would be good to decide what Slicer core should support and coordinate as compared to individual projects that can do what they want.

Sounds good!

jcfr:
> https://github.com/DCBIA-OrthoLab/SlicerCMF-docker

Our understanding is that this repository is specific to SlicerCMF project and will include packages like [clusterpost](https://github.com/juanprietob/clusterpost). @juanprietob may be able to comment further.

> SlicerBuildEnvironment vs SlicerDocker

SlicerBuildEnvironment repository is intended to provide images allowing to build Slicer. It currently has a `Docker` folder and was intended to be expanded to include virtual machine as well.

SlicerDocker is where all dockerfile are intended to be. It could be extended to include more images similarly to what is done in the repo [jupyter/docker-stacks](https://github.com/jupyter/docker-stacks).
See https://jupyter-docker-stacks.readthedocs.io/en/latest/using/selecting.html#image-relationships

lassoan:
> Our understanding is that this repository is specific to SlicerCMF project

Creating new Slicer docker images from scratch (instead of some base image) further increases fragmentation. If the CMF project figures out how to build a lean image then I would like to see all the other projects benefit from it; and when any project developer fixes a bug or makes further improvements then all other projects should benefit from it.

Of course you will add your own extensions, Python packages, and other project-specific settings in specialized dockerfiles, but we should share as much as possible.

> See https://jupyter-docker-stacks.readthedocs.io/en/latest/using/selecting.html#image-relationships

This looks nice. It would be great to have something similar. We can manage Slicer building and running in different repositories, because nothing is shared between them (but if it simplifies something then it should not be a problem to have them in the same repository either).

jcfr:
> > Our understanding is that this repository is specific to SlicerCMF project

> Creating new Slicer docker images from scratch (instead of some base image) further increases fragmentation.

The SlicerCMF images are really specific and aim to be integrated with DSCI (Data Storage for Computation and Integration) execution framework. See https://github.com/DCBIA-OrthoLab/dsci

Here are other examples of such images: 
https://github.com/DCBIA-OrthoLab/fly-by-cnn/tree/master/Docker

There are no particular optimization, the SlicerCMF docker image is intended to have Slicer and CLIs, the goal  is to support execcution of CLIs only.

lassoan:
Optimization was just an example. Even things like the choice of base image (ubuntu, debian, ... which version) would be very useful to coordinate between different groups. We could save a lot of time when we need to upgrade or troubleshoot any problems.

jcfr:
To summarize today's hangout:
* Add the download script to https://github.com/Slicer/SlicerDocker (in a `scripts` folder ?)
* Update documentation in this PR with the following:

```diff
-### Install Slicer and extensions using the downloader script
+### Install Slicer and extensions programmatically
+
+There are different approaches to address this
+* Call Slicer application invoking the `slicer.app.extensionManagerModel.installExtension()`. See [here]()
+* Directly interact with the REST API endpoints of https://slicer-packages.kitware.com using `curl` and `jq`. See [here]()

-On Linux, use [`slicer-download.sh`](https://github.com/Slicer/Slicer/blob/master/Utilities/Scripts/slicer-download.sh) to install  Slicer and selected extensions to a target directory.

-:::{note}
-
-`slicer-download.sh` depends on `curl`, `rsync`, and `jq`; be sure these are available on your system before using the script.
-
-:::
```

allemangD:
> To summarize today's hangout:
> 
> * Add the download script to https://github.com/Slicer/SlicerDocker (in a `scripts` folder ?)
> * Update documentation in this PR with a text like the following:

These are both done.

In `gui.md` I link to https://github.com/allemangD/SlicerDocker/blob/master/scripts/slicer-download.sh, which doesn't exist yet. That link should work once https://github.com/Slicer/SlicerDocker/pull/41 is merged but we should probably double-check that.

jcfr:
> That link should work once Slicer/SlicerDocker#41 is merged but we should probably double-check that.

The PR has been merged

allemangD:
The link was broken; I accidentally used my fork of SlicerDocker. That's fixed in f27a93e.

PR #5751, opened 2021-07-24
COMP: Simplify application package upload logic removing midas support

jcfr:
This commit removes support to upload to the deprecated midas based
server and makes the upload to the Girder based server required ensuring
it reports an error in case of upload failure.


jcfr:
I don't expect any issues with this set of changes. That would appreciate a sanity check in case I missed something.

pieper (approved):
I see no issues.  It looks much cleaner this way 👍

jcfr:
Thanks for the review. The key improvement is that upload will not fail because of midas server failure to respond.

Issue #5323, opened 2020-11-27
Segmentation imported from plain nrrd file is saved with smaller extent than the original


lassoan:
When a plain nrrd file is loaded as segmentation then the reference geometry is not set and so when the segmentation is saved, the extent is reduced to the minimum necessary. We should preserve the original extent.

See details here:
https://discourse.slicer.org/t/segmentation-is-cropped-to-minimum-extent/14788

lassoan:
@Sunderlandkyl you have fixed this, haven't you?

Sunderlandkyl:
Not sure. I'll double check.

Sunderlandkyl:
Should be fixed by 2d1e1d819a8cf410b025e8f4c523280860b5c736.

PR #5754, opened 2021-07-26
BUG: Save segmentation imported from plain nrrd using original geometry

Sunderlandkyl:
When loading a segmentation from a regular nrrd (rather than seg.nrrd), the reference image geometry of the original image was not saved.
Due to the missing geometry conversion parameter, exporting/saving the segmentation would cause the segmentation to be exported/saved at minimum extent. Fixed by setting the reference image geometry conversion parameter to the size of the imported nrrd.

Re #5323

### DIFF
--- a/Libs/MRML/Core/vtkMRMLSegmentationStorageNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLSegmentationStorageNode.cxx
@@ -677,6 +677,8 @@ int vtkMRMLSegmentationStorageNode::ReadBinaryLabelmapRepresentation(vtkMRMLSegm
 
     if (numberOfSegments == 0)
       {
+      // No segment metadata. We are loading from a plain volume (not seg.nrrd).
+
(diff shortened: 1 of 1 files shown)


lassoan (commented):
Thank you, it looks good to me! Does it work for nifti as well?

Sunderlandkyl:
It should, it's not nrrd specific.
It occurs to me that the other possibility would be to change the default behavior. If we are exporting/saving a segmentation that doesn't have reference geometry specified, then we default to the "minimum extent":

https://github.com/Slicer/Slicer/blob/fa6e5482035dd4e0619a132961ec3ba4ee680ce0/Libs/vtkSegmentationCore/vtkSegmentation.cxx#L1990-L2010

lassoan:
> It occurs to me that the other possibility would be to change the default behavior. If we are exporting/saving a segmentation that doesn't have reference geometry specified, then we default to the "minimum extent":

I don't think we have any better option. None of the extent options guarantee that the original geometry is preserved. So, the current implementation is good.

lassoan (approved):

PR #5755, opened 2021-07-26
ENH: Trim whitespaces from node names when generating filenames from them

### LINKED ISSUE #4523: Save dialog: Trim spaces when node name contain space at the end or begining
_This issue was created automatically from an original [Mantis Issue](https://mantisarchive.slicer.org/view.php?id=4523). Further discussion may take place here._

lassoan:
Filenames starting or ending with spaces may be hard to manage. Therefore, we remove leading and trailing whitespace characters when generating filename from node name (in Save data dialog and in segmentation export).

fixes #4523


pieper (approved):
Makes sense to me.

Issue #5734, opened 2021-07-10
Right-click context menu usability issues


jamesobutler:
- [x] If you right-click on a markup point and turn on the "Interaction" visibility action with visible handles, then right-click on the markup cannot bring up the markup context menu anymore. It will only bring up the slice view context menu. Therefore you have to manually navigate to the markups module to turn off the "Interaction" visibility to be able to use the markup right-click context menu again.
  - Remaining task: display translate/rotate/scale handles over control points (https://github.com/Slicer/Slicer/pull/5756#issuecomment-887771981)
- [x] When I'm using the "Draw" Segment Editor effect and use "Right-click" to apply the segment, the slice view context menu is appearing and then I have to click away from it and then the segment is applied. Maybe this effect needs an actual apply button? Or different method to end the interaction to apply? Or the context menu logic should know if the draw effect is active to not bring up context menu stuff?
- [ ] If you are using the Surface Cut or Draw Tube segment editor effect from SegmentEditorExtraEffects extension, you can right-click on a markup point and select "Delete markup" which will delete the `segmentMarkupNode` object used by the effect and mess up its usage. Maybe there should be customizations done to hide some markups context menu actions while using Segment Editor effects? I'm not sure if there are other uses of markups in other modules where the default "Delete Markup" action could mess up some internal logic where a markup node is being used.
  - Remaining task: hide "Delete markup node" menu action

## Environment
- Slicer version: Slicer-4.13.0-2021-07-09
- Operating system: Windows

lassoan:
> If you right-click on a markup point and turn on the "Interaction" visibility action with visible handles, then right-click on the markup cannot bring up the markup context menu anymore

I could not reproduce this. Could you debug into this? Does the code get to qSlicerSubjectHierarchyMarkupsPlugin::showViewContextMenuActionsForItem?

> When I'm using the "Draw" Segment Editor effect and use "Right-click" to apply the segment

Could you try if you can use double-click event would work well for finishing the placement? You can also hit the `Enter` key and of course it would not hurt to have an "Apply" button either.

> If you are using the Surface Cut or Draw Tube segment editor effect from SegmentEditorExtraEffects extension, you can right-click on a markup point and select "Delete markup"

[...]

PR #5756, opened 2021-07-26
ENH: Show context menu if clicking on markups interaction handle

lassoan:
Previously, interaction handles (translate/rotate/scale) did not take right-click events and so the view context menu was shown, so it was hard to turn off interaction handles.

re #5734

### DIFF
--- a/Modules/Loadable/Markups/VTKWidgets/vtkSlicerMarkupsWidget.cxx
+++ b/Modules/Loadable/Markups/VTKWidgets/vtkSlicerMarkupsWidget.cxx
@@ -72,9 +72,13 @@ vtkSlicerMarkupsWidget::vtkSlicerMarkupsWidget()
   this->SetEventTranslation(WidgetStateOnWidget, vtkMRMLInteractionEventData::LeftButtonClickEvent, vtkEvent::NoModifier, WidgetEventJumpCursor);
   this->SetEventTranslation(WidgetStateOnWidget, vtkCommand::LeftButtonDoubleClickEvent, vtkEvent::NoModifier, WidgetEventAction);
 
-  this->SetEventTranslation(WidgetStateOnWidget, vtkCommand::RightButtonPressEvent, vtkEvent::NoModifier, WidgetEventReserved);
-  this->SetEventTranslation(WidgetStateOnWidget, vtkCommand::RightButtonReleaseEvent, vtkEvent::NoModifier, WidgetEventReserved);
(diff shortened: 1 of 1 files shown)


jamesobutler (commented):
I am able to toggle off interaction "Translate" handles now, but I'm observing different context menu options when right-clicking on the center ROI fiducial with the "Translate" handle visible or not. Is this expected behavior?

|Right-click translate handle off | Right-click translate handle on|
|----------------------------------|-----------------------------------|
|image|image|

lassoan:
> I am able to toggle off interaction "Translate" handles now, but I'm observing different context menu options when right-clicking on the center ROI fiducial with the "Translate" handle visible or not. Is this expected behavior?

Yes, it is the correct behavior. Unlike control points, handles cannot be toggled, renamed, or deleted.

jamesobutler:
What is confusing is that I'm clicking on the center ROI fiducial point, not on the red or green translate line. Why is the fiducial point include in the translate handle then? It is confusing when the fiducial point is acting as a fiducial and then changes to acting as a handle.

lassoan:
Within one context (e.g., mouse pointer, left virtual reality controller, right virtual reality controller) you can always select only one component of one widget. It is just the coincidence that there is a control point below the handle. If the translate handle is highlighted (orange disk turns into yellow) then you will manipulate that. If the control point is highlighter (e.g., yellow crosshair turns into green) then you will manipulate that. You can easily hide the interaction handles if they get into your way.

jamesobutler:
> If the translate handle is highlighted (orange disk turns into yellow) then you will manipulate that. If the control point is highlighter (e.g., yellow crosshair turns into green) then you will manipulate that.

What's confusing is that it doesn't highlight in the view where I enabled the translate handle visibility. It highlights when I hover over the point in other slice views. That's why I though I was clicking on the fiducial and not the handle. It is clear to me that if I highlighted over the lines for the translate handle that the right-click context menu would be related to the handle, but it's confusing to me why I can't for instance rename this markup point with the translate handles on even though I'm right-clicking on the fiducial.

https://user-images.githubusercontent.com/15837524/127054003-3919f4bf-cf62-4f7a-a909-d72953cad2d4.mp4

lassoan:
Depending on the shape and sizing method of the fiducial may indeed occlude the translation handle, which may be confusing. Since the handles have priority over control points it would be probably better to render the handles over the control points in slice views (similarly as it is done in 3D views). I'll merge this as is and will address display of handles over control points in a separate issue.

Issue #5757, opened 2021-07-27
Grow from seeds effect preview in 3D is not updating


lassoan:
When "Grow from seeds" auto-update is enabled and editing seeds using Paint effect, preview is updated in slice views, but not in 3D.

## Steps to reproduce

- Load this scene (MRHead with two segments, Grow from seeds auto-update is active): https://1drv.ms/u/s!Arm_AFxB9yqHxecKBZLZhzZ-swJGMA?e=tzTVCZ
- Paint with the Segment_1 in the red view. Grow from seeds output is not updated in 3D! Sometimes it is not updated in slice views either.

## Expected behavior

Disabling and enabling "Show 3D" for the results preview updates the 3D preview, but it should happen automatically when seeds are edited.

## Environment
- Slicer version: Slicer-4.13.0-2021-07-25
- Operating system: Windows

lassoan:
@Sunderlandkyl please check. Probably these update issues are caused by the recent performance optimizations.

lassoan:
Fixed in https://github.com/Slicer/Slicer/commit/e8ed0d97861359462293156fbf4de7dd6cb8a56b

PR #5758, opened 2021-07-27
BUG: Force update of grow from seeds 3D closed surface preview

Sunderlandkyl:
Due to a previous optimization that set the labelmap representation of the preview segments directly, the 3D preview closed surface was no longer automatically updating. To fix this, we remove the existing closed surface so that it will be reconverted after the segmentation is updated.

When the functions for setting multiple segments are added, we will no longer need this call as the closed surface update will be handled by the segmentation logic.

Re #5757


lassoan (approved):
Looks good, thanks for the quick fix.

PR #5759, opened 2021-07-27
ENH: Get node from volume display widgets using Python interface

jamesobutler:
I found exposing to the python interface getting the volume node of some volume widgets helpful. Some other widgets have it exposed from the python interface using calls such as `currentNode()` and now that is possible with these widgets using their already named methods of `mrmlVolumeNode()` and `volumeNode()`. 

It would be nice if getting the node for the associated qMRML/qSlicer widgets had a more consistent naming scheme, but I'll leave that for a more comprehensive future change.


lassoan (approved):
Looks good to me, thank you.

Issue #5338, opened 2020-12-09
Segment Editor "overwrite visible segments" mode sometimes allow overlap


lassoan:
If "Overwrite visible" mode is enabled and there is a hidden segment then visible segments are not overwritten (which would be the correct behavior) but they overlap.

See more information here: https://discourse.slicer.org/t/masking-overwrite-visible-allowing-overlap-instead-of-overwriting/14857/8

## Steps to reproduce

- Load this scene: https://1drv.ms/u/s!Arm_AFxB9yqHxM1HwpPPM-NsSSPYHw?e=eXGuya
- Select Segment_2
- Click Apply in Margin effect
- Purple segment overlaps with the visible green segment => ERROR

## Expected behavior

Purple segment should overwrite the visible green segment.

## Environment
- Slicer version: Slicer-4.13.0-2020-12-07
- Operating system: Windows 10

lassoan:
@Sunderlandkyl have you looked at this issue? I remember you worked on something similar but I'm not sure it was exactly this.

Sunderlandkyl:
No, I'll take a look at it.

lassoan:
@Sunderlandkyl you have fixed a lot of segmentation issues since February. Is this one still unresolved?

jamesobutler:
Following the instructions I provided at https://discourse.slicer.org/t/masking-overwrite-visible-allowing-overlap-instead-of-overwriting/14857/8, I am still able to replicate the observed behavior using Slicer 4.13.0-2021-07-22.

Sunderlandkyl:
Sorry, this fell off of my radar. I'll take a look on Monday.

lassoan:
Fixed in https://github.com/Slicer/Slicer/pull/5760/commits/286d0112a4826101291a70f2b7330eaa567e2efd

Issue #5584, opened 2021-04-13
Segment Editor's Erase effect "Erase all segments" option only erase current segment


lassoan:
Segment Editor's Erase effect, if "Erase all segments" checkbox is checked then painting strokes in views should erase from all segments, but it only erases from the currently segment.

## Steps to reproduce

- Load MRHead
- Create 3 segments and paint a few strokes.
- Click Erase effect, check "Erase all segments" option
- Click-and-drag in slice views => only the currently selected segment is erased from

## Expected behavior

The strokes should erase content from all segmentes.

## Environment
- Slicer version: Slicer-4.13.0-2021-03-27
- Operating system: Windows10

lassoan:
This issue is still present in Slicer-4.13.0-2021-07-24.

jcfr:
@Sunderlandkyl  If you didn't already start to work on this, we would like to assign this issue to a new developer  that just joined our team. What do you think ?

Sunderlandkyl:
Currently working on it. It's been rolled into some other updates that I'm making. Should be done soon.

PR #5760, opened 2021-07-27
BUG: Fix invalid overlap when splitting segments

Sunderlandkyl:
When modifying a segment the list of segments that need to be overwritten does not normally include segments that are on the same layer (since they will be overwritten by default). This list was populated before the selected segment is modified. This causes segments on the same layer to not be overwritten if the selected segment is separated to a new layer.

Fixed by populating the list of segments to overwrite after the selected segment has been modified (and potentially moved to a new layer).

Re #5338


lassoan (approved):
Perfect, thank you!

lassoan:
Is this going to fix this issue as well? https://github.com/Slicer/Slicer/issues/5584

Sunderlandkyl:
No, that will require a separate fix.

PR #5761, opened 2021-07-27
ENH: add QPageLayout argument to printToPdf method

mauigna06:
Allows to print PDFs with custom margins

### DIFF
--- a/Base/QTGUI/qSlicerWebWidget.h
+++ b/Base/QTGUI/qSlicerWebWidget.h
@@ -103,6 +103,7 @@ public slots:
 
   void onDownloadFinished(QNetworkReply* reply);
 
+  ///@{
   /// Renders the current content of the page into a PDF document and saves it in the location specified in filePath.
   ///
   /// The page size and orientation of the produced PDF document are taken from the values specified in pageLayout.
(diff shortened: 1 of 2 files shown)


lassoan (commented):
Thank you, it looks good, except existing signals that are connected to the `printToPdf(const QString& filePath);` slot will not work anymore. Since the default parameter is also very convoluted, it would be probably better to add a new slot without default argument value (`void printToPdf(const QString& filePath, const QPageLayout& pageLayout)`).

Base/QTGUI/qSlicerWebWidget.h
```diff
@@ -112,6 +112,7 @@ public slots:
   /// If a file already exists at the provided file path, it will be overwritten.
   /// \sa QWebEnginePage::printToPdf
   void printToPdf(const QString& filePath);
```
lassoan, line 114: Nitpick: could you add doxyven group (something like @{ @}) around the documentation and declaration of these two methods?

mauigna06: done

lassoan (approved):

PR #5762, opened 2021-07-29
ENH: Allow FOV adjustment without slice view reorientation

riep:
The useful FitSliceToAll function in ApplicationLogic may change the slice view orientation. User may want to keep the current orientation and simply adjust the FOV.

### DIFF
--- a/Libs/MRML/Logic/vtkMRMLApplicationLogic.cxx
+++ b/Libs/MRML/Logic/vtkMRMLApplicationLogic.cxx
@@ -514,7 +514,7 @@ void vtkMRMLApplicationLogic::PropagatePlotChartSelection()
 }
 
 //----------------------------------------------------------------------------
-void vtkMRMLApplicationLogic::FitSliceToAll(bool onlyIfPropagateVolumeSelectionAllowed /* =false */)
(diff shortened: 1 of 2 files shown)


Libs/MRML/Logic/vtkMRMLApplicationLogic.cxx
```diff
@@ -514,7 +514,7 @@ void vtkMRMLApplicationLogic::PropagatePlotChartSelection()
 }
 
 //----------------------------------------------------------------------------
-void vtkMRMLApplicationLogic::FitSliceToAll(bool onlyIfPropagateVolumeSelectionAllowed /* =false */)
+void vtkMRMLApplicationLogic::FitSliceToAll(bool onlyIfPropagateVolumeSelectionAllowed /* =false */, bool rotationAllowed /* =true */)
```
cpinter, line 517: `rotationAllowed` is not a good name here. It is not about it being allowed (as opposed to the other argument, which handles the case when slice composite nodes do not allow propagate volume selection), but being applied. I suggest `resetOrientation` instead.

cpinter:
Thanks! I added one comment, otherwise it looks good to me.

Libs/MRML/Logic/vtkMRMLApplicationLogic.cxx
riep: Done

pieper (approved):
Looks good to me 👍

PR #5765, opened 2021-08-03
ENH: Add prefix for oblique slice offset

riep:
Currently, no prefix is printed in the slice offset label for oblique slices. This may induce flickering of the slice offset slider when switching view orientations, for instance.

### DIFF
--- a/Libs/MRML/Widgets/qMRMLSliceControllerWidget.cxx
+++ b/Libs/MRML/Widgets/qMRMLSliceControllerWidget.cxx
@@ -78,7 +78,7 @@ qMRMLSliceControllerWidgetPrivate::qMRMLSliceControllerWidgetPrivate(qMRMLSliceC
   qMRMLOrientation axialOrientation = {qMRMLSliceControllerWidget::tr("S: "), qMRMLSliceControllerWidget::tr("I <-----> S")};
   qMRMLOrientation sagittalOrientation = {qMRMLSliceControllerWidget::tr("R: "), qMRMLSliceControllerWidget::tr("L <-----> R")};
   qMRMLOrientation coronalOrientation = {qMRMLSliceControllerWidget::tr("A: "), qMRMLSliceControllerWidget::tr("P <-----> A")};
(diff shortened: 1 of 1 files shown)


jcfr (changes requested):

Nitpick: Consider updating the PR removing the `onedrive` directory.

@lassoan Do you recall if we are relying on the fact the name is an empty string ? 


_For reference, `Oblique` is referenced in the literature (see https://pubs.rsna.org/doi/10.1148/rg.255055044))_

riep:
Sorry I try to correct the typo in the branch name. I am not sure I can do it if a PR is started.

lassoan:
Fixed size policy is risky. There is no single width value that is always enough and never too much. The length unit format can be changed dynamically, so this width should be more dynamically changing, too. You might try to set a minimum size based on percentage of width, but avoid hard coding a maximum value. 

Normally the variance of the width of this field is negligible. Why do you think it is causing a prpbl for you? Have you added many custom buttons next to the slider? Do you have many slice views side by side, which makes them very narrow? 

Slice orientation preset changes very rarely. How does it cause flicker (repeated changes) for you?

riep:
When you load a volume an play with the reformat the label size of the reformat view is changing. I will try to find an other approach to avoid this. This is not critical but as I was on this , I thought it could be a little improvement.

lassoan:
> I thought it could be a little improvement.

Yes, a fixed field size can improve the display in some cases, but it may cause problems in other cases.

Adding the "O:" prefix might improve consistency of the formatting a little bit, but it does not help with interpreting the value (in fact, it may be a bit confusing, as "O" might be the abbreviation of a direction in some contexts). To improve this label, it should be actually computed automatically from AxisLabels of the view. In case of oblique view, we would get 2 or 3 letter combinations, such as LA <--> RP, or LAS <--> RPI.

If you can devote more time to improve the orientation label (compute from AxisLabels) or figure out a way to stabilize the offset value display without hardcoding fixed width then we would be very happy to integrate those changes.

lassoan:
I close this pull request now to keep the pull request list manageable, but feel free to reopen if you have time to work on this more and further improve things.

PR #5769, opened 2021-08-05
BUG: Disable DCMTK ICU default for Superbuild

tbirdso:
Fixes an observed issue where building on Windows with Anaconda in PATH
could result in a path being found for nonexistent ICU libraries.

Co-authored-by: Jean-Christophe Fillion-Robin <jcfr@kitware.com>


tbirdso:
cc @jcfr

lassoan:
We have run into build issues when Anaconda was installed, so it is awesome that you managed to find the root cause of the issue.

I'm not sure that simply disabling ICU is a good solution, though. After this change, have you tried to load DICOM files that contained patient names with international characters with various encodings? You can find in the web DICOM files with various encodings for such testing (I don't remember where exactly, but if you cannot find them then let me know and I can help).

jcfr:
On the window factory, there is no icu library and the option end up being turns off.

We will check what the current values on Linux and macOS.

lassoan:
OK, then the change will probably not cause regressions. DICOM data sets that you can use to verify this: http://www.dclunie.com/images/charset/index.html - It is enough if you import the data sets and have a look at the DICOM browser. There is a picture that shows how the international characters should look like. As far as I remember, there were a few special encodings that did not work, but most of them should be good.

tbirdso:
Comparing the completed superbuild with `ICU=OFF` to the previous stable release 4.11.20210226 on Windows, patient DICOM data loads and displays the same in both versions:

dicom-patient-id

In both cases a few characters are missing compared with the ground truth image distributed with the original data set, @lassoan are these the special encodings you mentioned? Image for comparison:

charsettests screenshot

lassoan (approved):
Thank you for the testing. It all looks good to me.

jcfr:
> We will check what the current values on Linux and macOS.

For future reference, the option was also `OFF` on macOS and Linux builds.

jcfr:
To avoid the following linking error observed within a pre-existing Linux build, deleting the folder `DCMTK-build` is required.

> ```
> [  2%] Linking CXX executable ../../bin/ofstd_tests
> ../../lib/libofstd.so.16.3.6.6: undefined reference to `ucnv_getName_55'
> ../../lib/libofstd.so.16.3.6.6: undefined reference to `ucnv_setFromUCallBack_55'
> ../../lib/libofstd.so.16.3.6.6: undefined reference to `UCNV_TO_U_CALLBACK_SKIP_55'
> ../../lib/libofstd.so.16.3.6.6: undefined reference to `ucnv_open_55'
> ../../lib/libofstd.so.16.3.6.6: undefined reference to `ucnv_convertEx_55'
> ../../lib/libofstd.so.16.3.6.6: undefined reference to `UCNV_TO_U_CALLBACK_STOP_55'
> ../../lib/libofstd.so.16.3.6.6: undefined reference to `UCNV_FROM_U_CALLBACK_STOP_55'
> ../../lib/libofstd.so.16.3.6.6: undefined reference to `UCNV_FROM_U_CALLBACK_SKIP_55'
> ../../lib/libofstd.so.16.3.6.6: undefined reference to `ucnv_setToUCallBack_55'
> ../../lib/libofstd.so.16.3.6.6: undefined reference to `ucnv_getFromUCallBack_55'
> ../../lib/libofstd.so.16.3.6.6: undefined reference to `ucnv_close_55'
> ../../lib/libofstd.so.16.3.6.6: undefined reference to `u_errorName_55'
> ```

PR #5771, opened 2021-08-05
BUG: Fix setting of plot view label

### LINKED ISSUE #5770: Plot name cannot be modified
The plot view name cannot be modified. Also, please note that a warning pops up when switching to a view with a plot 

## Steps to reproduce
Add a plot view in slicer and run the following command

```python
plotNode = slicer.app.layoutManager().plotWidget(0).mrmlPlotViewNode()
plotNode.SetLayoutLabel("myplot")
```

## Expected behavior
Plot view label should be changed to myplot

## Environment
- Slicer master
- Operating system: Linux 20.04

lassoan:
There was an incorrect null-pointer check that prevented update of view label from plot view node.

While fixing the problem, also noticed that view label setting is not consistent among various view widgets. This commit harmonizes all of them:
- view widget delegates setViewLabel and viewLabel calls to view controller widget
- view controller widget uses view node's LayoutLabel property for getting/setting the view label

fixes #5770


riep:
Hi, thank you for the fix!
I noticed also that setting the names in the xml definition of the layout does not work. It is the same for you? Maybe it is related to our slicelet implementation...

lassoan:
The referenced pull request fixes that, too.

lassoan:
> Considering removing `#include <ctkLogger.h>` and `static ctkLogger logger("org.slicer.libs.qmrmlwidgets.qMRMLChartViewControllerWidget");`

OK, I'll remove this.

Do you know what are the differences and potential advantages of this ctkLogger approach compared to just using standard Qt logging functions?

PR #5781, opened 2021-08-09
BUG: Fix binary labelmap file export using color table

Sunderlandkyl:
When exporting binary labelmaps to file, the color table was not being taken into account.
Fixed by adding an argument for labelmap values to vtkSlicerSegmentationsModuleLogic::GenerateMergedLabelmapForAllSegments().

Also added additional documentation to vtkSlicerSegmentationsModuleLogic.h.


lassoan (approved):
This is perfect, thank you!

PR #5784, opened 2021-08-10
ENH: qSlicerWebWidget: Log javascript message if developer mode is enabled

jcfr:
The purpose of this changes is to help identify issues within `qSlicerWebWidget` by inspecting the Slicer log file.


lassoan:
It would be a bit more clear if we added a `javaScriptConsoleMessageLoggingEnabled` property (with get/set methods) and initialize the default value from `developerMode` setting.

Issue #5773, opened 2021-08-06
Incoming folder in DICOM database folder not emptied after ctk.ctkDICOMRetrieve()


rbumm:
[Give a short summary of what the problem is]

After DICOM networking retrieval with ctk.ctkDICOMRetrieve() (reading from local or remote PACS)  the DICOM database folder contains two folders: 
"dicom"
"incoming"

Directly after reading (and hours later) the size of "incoming" is about the same as "dicom", several GB depending on the data requested.   

## Expected behavior

"dicom" should contain the data, "incoming" should be empty after retrieval. 

## Possible cause

"incoming" seems not to be routinely emptied after each successful retrieval. 

## What I tried

Deleted all files in "incoming" which did not have a negative effect when I reopened Slicer with the same database.  

## Environment
- Slicer version: Slicer-4.13.0-2021-07-29
- Operating system: Windows

pieper:
It should be enough to add `os.remove(dicomFilePath)` between the two lines below.  The incoming folder would then be emptied as the files are added to the database.

https://github.com/Slicer/Slicer/blob/21f8ba46beb67ae169387dd94237b0d7f8087302/Modules/Scripted/DICOMLib/DICOMProcesses.py#L341-L342

PR #5785, opened 2021-08-11
ENH: DICOM - Added deletion of files

rbumm:
in the incoming directory after DICOM network retrieval

### DIFF
--- a/Modules/Scripted/DICOMLib/DICOMProcesses.py
+++ b/Modules/Scripted/DICOMLib/DICOMProcesses.py
@@ -339,6 +339,7 @@ class DICOMListener(DICOMStoreSCPProcess):
       if self.fileToBeAddedCallback:
         self.fileToBeAddedCallback()
       self.indexer.addFile(self.dicomDatabase, dicomFilePath, True)
+      os.remove(dicomFilePath)
       logging.debug("done indexing")
       self.lastFileAdded = dicomFilePath
       if self.fileAddedCallback:


rbumm:
as suggested by @pieper here: https://github.com/Slicer/Slicer/issues/5773#issuecomment-896279704

pieper (approved):
LGTM 👍 

Thanks for trying this out and adding it.

pieper:
Yes, but here the `copyFile` flag is true, so the indexer has the file in the database folder and doesn't need the one from incoming.  So this code should be fine as-is.

lassoan:
Copyfile flag is taken into account when the indexing request is processed some time in the future. You could just add `self.indexer.waitForImportFinished()` after the `self.indexer.addFile` call, before deleting the original, but that would mean that displayable fields would be updated after each file, which might slow down things significantly.

In DICOMWebBrowser, we collect files for each series in a separate folder and when download is complete then we index the folder and delete the files (see [here](https://github.com/lassoan/SlicerDICOMwebBrowser/blob/413755a16cbf0fb50408dd8e805d2e711ddb323d/DICOMwebBrowser/DICOMwebBrowser.py#L845-L861)). The idea is to index a bunch of files at once because then indexing can be much faster (there is cached information during indexing that reduces the number of database reads and update of displayed fields - such as statistics on total number of frames, last update times, etc. are computed only once).

You should be able to avoid file copy and delete completely by moving the file to its final path (anywhere within the DICOM database folder) before requesting indexing (see for example in the [DICOMwebBrowser how to generate unique storage path](https://github.com/lassoan/SlicerDICOMwebBrowser/blob/413755a16cbf0fb50408dd8e805d2e711ddb323d/DICOMwebBrowser/DICOMwebBrowser.py#L733)) and set `copyFile=False`.

pieper:
This background indexing feature is good, since it sped up the overall process, but bad that it broke the api contract.  The `copyFile` flag should mean that the copy is completed when the `addFile` call returns so that the caller doesn't need to know about the rest of the import process.

rbumm:
Adding self.indexer.waitForImportFinished() did not prolong the retrieval operation and resulted in two usable lung CTs and an empty incoming folder. 

BTW in 4.13 in the DICOM Browser, I always need to "Examine" the datasets first before I can "Load" them. The Doubleclick no longer works for loading. This seems to be new and isn't that misleading for new users?

rbumm:
Should self.indexer.waitForImportFinished() be added in the script then and pushed again? I am just asking to know what is expected and what the normal procedure would be ... Thanks

pieper:
> Should self.indexer.waitForImportFinished() be added in the script then and pushed again? I am just asking to know what is expected and what the normal procedure would be ... Thanks

I would say so, yes, that's the easiest.  I believe you can just push to the same branch and make a new PR with that line added before the delete.

lassoan:
> Changing the behavior of an API call sounds like the very definition of breaking backwards compatibility. 

OK, I see what you mean then. I probably should have documented the change better (update API documentation, migration guide, etc.). It is hard to press myself to spend more time with writing documentation, seeing how reluctant people are (myself included) to read documentation.

pieper:
I agree, it's hard to document things completely and hard to know if anyone will read/understand the documentation when they actually need it.  In this case I looked at the docstring for `addFile` and didn't  look at the rest of the class where the background behavior is well described; maybe someone new to the code wouldn't have made the assumption I did.

In general though I think it preferred to add new flags or methods for new behavior instead of changing behavior of existing APIs unless they are actually buggy.  Remember we [had that issue with vtk files](https://discourse.vtk.org/t/legacy-polydata-file-compatibility/5354/3) not so long ago.  In this specific case of Slicer's use of CTK dicom we could have changed our code to use a `requestFileIndex` or similar.

lassoan:
> In general though I think it preferred to add new flags or methods for new behavior instead of changing behavior of existing APIs unless they are actually buggy.

The default behavior of CTK did not change. `BackgroundImportEnabled` flag is set to false by default to preserve backward compatibility. When I implemented and enabled this feature in Slicer, I've updated Slicer core and all extensions accordingly.

I'll add some more documentation to the methods when I modify CTK next time.

rbumm:
Just tested again with a big dataset (9GB), all is fine and in sync, incoming empty, but needed 45 min retrieval time.  I needed to create a pull request from a different branch because I already deleted the previous one. 

@jcfr 

\>> Please, consider creating a follow up pull request updating the docstring of `DICOMListener`

What exactly should I do here ? .

PR #5787, opened 2021-08-11
ENH: Update DICOMListener with waitForImportFinished()

rbumm:
Add self.indexer.waitForImportFinished() before calling os.remove() to make a DICOM retrieval followed by emptying the "incoming" directory more failsafe.  
There may be a possible interaction with self.indexer as a background process

### DIFF
--- a/Modules/Scripted/DICOMLib/DICOMProcesses.py
+++ b/Modules/Scripted/DICOMLib/DICOMProcesses.py
@@ -311,9 +311,30 @@ class DICOMListener(DICOMStoreSCPProcess):
   def __init__(self, database, fileToBeAddedCallback=None, fileAddedCallback=None):
     self.dicomDatabase = database
     self.indexer = ctk.ctkDICOMIndexer()
+    # Enable background indexing to improve performance.
+    self.indexer.backgroundImportEnabled = True
     self.fileToBeAddedCallback = fileToBeAddedCallback
     self.fileAddedCallback = fileAddedCallback
     self.lastFileAdded = None
+
+    # A timer is used to ensure that indexing is completed after new files come in,
+    # but without enforcing completing the indexing after each file (because
+    # waiting for indexing to be completed has an overhead).
+    autoUpdateDelaySec = 10.0 
+    self.delayedAutoUpdateTimer = qt.QTimer() 
+    self.delayedAutoUpdateTimer.setSingleShot(True) 
+    self.delayedAutoUpdateTimer.interval = autoUpdateDelaySec * 1000 
+    self.delayedAutoUpdateTimer.connect('timeout()', self.completeIncomingFilesIndexing)
+
+    # List of received files that are being indexed
+    self.incomingFiles = []
+    # After self.incomingFiles reaches maximumIncomingFiles, indexing will be forced
+    # to limit disk space usage (until indexing is completed, the file is present both
+    # in the incoming folder and in the database) and make sure some updates are visible
+    # in the DICOM browser (even if files are continuously coming in).
+    # Smaller values result in more frequent updates, slightly less disk space usage,
+    # slightly slower import.
+    self.maximumIncomingFiles = 400
 
     databaseDirectory = self.dicomDatabase.databaseDirectory
     if not databaseDirectory:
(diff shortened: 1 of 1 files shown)


pieper:
Let's let Andras have another look before merging, but this looks correct to me.  Thanks for the contribution and your patience.

lassoan:
`self.indexer.waitForImportFinished()` update all displayed fields, which can be a lengthy operation (the more files you have in the database it takes longer). It would be better not to run it after each and every file import, but after every 100 files or so. You can collect the `dicomFilePath` values in a list (`dicomFilePaths`) and when the list reaches a certain size (e.g., 100) then call `self.indexer.waitForImportFinished()`, remove each file in `dicomFilePaths` list, then clear the `dicomFilePaths` list.

rbumm:
Even if I did not have a notable speed difference in the 18 cases tested today? "incoming" swoll on and off with to around 500 files each series ...

lassoan:
If the files are large and/or the network is slow enough then the network transfer may remain the performance bottleneck, but in other cases too frequent `waitForImportFinished` may cause slowdown.

Could you measure the time that `addFile` and `waitForImportFinished` takes? I would expect that one of them will take longer and longer.

rbumm:
Add file time is 0.02700 s in the beginning and is at 0.10399 s after 17000 files and 45 min. It constantly rises as the database grows. 

Index wait time stays near zero or is displayed as zero all the time.

lassoan:
It most likely means that background indexing is already turned off in the indexer. I've had a look at the code an know why. It is because DICOMProcesses.py creates its own indexer and by default background indexing is disabled. 

0.1s per file is a significant slowdown. It means that a 300-frame image is imported in 30 seconds instead of a few seconds.

You can make DICOM import 10x faster by enabling background indexing and only call waitForImportFinished once after each few hundred files. Since DICOM receiving and importing are done in parallel, the unnecessarily slow import might not be a performance bottleneck if network transfer is relatively slow. However, as the database grows, the import may start to become slower than the network transfer, making the entire process take more time.

rbumm:
That sounds logical, hopefully, I can come up with a working code and test that out. The retrieval model for 18 cases in a row is available.

rbumm:
Next test: The complete removal of DICOMWaitForImportFinish and os.remove() line did not change the increasing add file times in a 15 cases / 33 studies block retrieval.

rbumm:
Next Test: Commenting out self.indexer.addFile() results the 8.7 GB to be retrieved to "incoming" in 1745 seconds. 

Secondary import of that data into the DICOM database from incoming takes 80 s.

rbumm:
Doing the same test with a stopped DICOM listener takes 2034 s to read the dataset described before to the incoming directory. Stopping the DICOM listener process within Slicer - to my understanding - stops background indexing.

rbumm:
My preliminary conclusion is, that there may be a potential performance benefit if the indexer does not addFile() for every received file. The difference of 1745 (dicom listener loaded)  vs. 2043 s (dicom listener not loaded) remains unclear to me, maybe it was related to local network speed. 

Adding waitForImportFinished() will probably not add additional speed, but I do not see that it slows down the retrieval. Probably it is only necessary if you addDirectory().

rbumm:
Coming back to this I did another test and measured the times between retrieving a bit multi-case dataset. 

      self.indexer.addFile(self.dicomDatabase, dicomFilePath, True)

      self.indexer.waitForImportFinished()

      os.remove(dicomFilePath)

addFile() time constantly grows during the retrieval process, waitForImportFinished() and remove() stay constantly low and hardly use any time at all. 
To straighten things out I close this specific PR now because adding waitForImportFinished does not make much sense to me any longer. If you disagree pls notify me. 

There would be room for speed optimization by implementing addDirectory instead of addFile but for this the listener would have to know when a  retrieval process into incoming is actually finished or it would have to periodically check incoming for remaining files.

lassoan:
Can you commit your updated and tested code?

rbumm:
Committed the updated and tested code.

lassoan:
Thanks a lot @rbumm. I've pushed an update with two important fixes:
- Only delete those files from the incoming folder that we know about (and not all the files in the folder). The incoming folder may contain files that are being received (but Slicer has not been notified about) or files that are created by another instance of Slicer.
- waitForImportFinished + file remove is called after reaching a certain limit on number of files. This ensures that disk usage is not doubled in case continuously receiving many files. It also allows seeing progress (new data sets appearing in the database) during lengthy continous transfers.

I've coded this blindly, did not test at all, so it would be great if you could test it.

rbumm:
Thanks @lassoan , looks good. I will test the code as soon as I reach a place with good internet, probably Thursday, and report back.

rbumm:
@lassoan  still looking good. I was able to test your new code with the 15 cases import, all worked well and the procedure took 1854.44 s, only slightly longer than the 1700 s  used by my last modification. 

After the retrieval, "incoming" was empty and all datasets in the database were readable. 

It is certainly good to remove only known files from "incoming", something that was also mentioned by @pieper before. Thanks for your help

rbumm:
And why did this take so long (1854 s)? Actually, I think this is quite fast because my 15 cases have 43 DICOM datasets with series description "Lung\*" and modality "CT"

lassoan:
Thanks for the testing, I'll merge this now.

PR #5789, opened 2021-08-11
ENH: use measurment frame to color dti slices

pieper:
The color-by-orientation slice rendering mode
should take into account the measurement frame
of the tensor acquistion.

This information is stored in the
vtkMRMLDiffusionTensorVolumeNode
and needs to be copied to the
vtkMRMLDiffusionTensorVolumeNode.

### DIFF
--- a/Libs/MRML/Core/vtkMRMLDiffusionTensorVolumeDisplayNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLDiffusionTensorVolumeDisplayNode.cxx
@@ -32,6 +32,7 @@ Version:   $Revision: 1.2 $
 #include <vtkImageMapToWindowLevelColors.h>
 #include <vtkImageShiftScale.h>
 #include <vtkImageThreshold.h>
+#include <vtkMatrix4x4.h>
 #include <vtkObjectFactory.h>
 #include <vtkSphereSource.h>
 #include <vtkVersion.h>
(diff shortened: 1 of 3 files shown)


Libs/MRML/Core/vtkMRMLDiffusionTensorVolumeDisplayNode.cxx
```diff
@@ -226,6 +227,14 @@ void vtkMRMLDiffusionTensorVolumeDisplayNode::UpdateImageDataPipeline()
 
 }
 
+//----------------------------------------------------------------------------
+void
+vtkMRMLDiffusionTensorVolumeDisplayNode::SetTensorRotationMatrix(vtkMatrix4x4 *rotationMatrix)
```
lassoan, line 232: It would be important to include the coordinate system and transform direction in the name: RAS/LPS (this is especially critical in the storage node) and ToRAS or FromRAS. For example, TensorMeasurementFrameToRASRotation or TensorToRASRotation names would be specific enough.

pieper: Yes, we should rationalize all the naming conventions for these methods.  Some places use Measurement Frame, others use TensorRotationMatrix.  The wording I chose here seems most consistent with the rest of this API, which assumes all of this is in RAS.  Let's let @ljod look at this when she's back from vacation.

lassoan: If the same name is already used at many places and it is used fully consistently then it is probably not worth the trouble of changing the name now, just provide a clear definition in the API documentation.

ljod (dismissed):
These look good to me.

pieper:
Before the fix, note that the corpus callosum is green.
 
image

After the fix it turns the proper red color (RGB is L/R, A/P, I/S).

image

Libs/MRML/Core/vtkMRMLDiffusionTensorVolumeDisplayNode.cxx
pieper: All the tensors and gradient calculations are in RAS and that's not explicit in any of the method names so this method name is consistent with the rest of the code.

Node documentation already specifies this:

https://github.com/Slicer/Slicer/blob/40087afee09fa4779e1b809423d927add1904cb7/Libs/MRML/Core/vtkMRMLDiffusionTensorVolumeNode.h#L29-L30

PR #5794, opened 2021-08-12
ENH: captureImageFromView processEvents is optional

nathanbmnt:
I have a python scripted module which hides segmentations and changes slice offsets before screenshotting, every time a segmentation is modified. After screenshot the slice offsets and segmentation visibilities are restored. I use `ScreenCaptureLogic().captureImageFromView(view, volumeNode=screenshot_node)` but it is very jarring for the user when `slicer.app.processEvents()` is called which makes the offsets and seg update visible to the user before being restored.

I noticed that my screenshots are no different if I remove `slicer.app.processEvents()` so this PR makes that step optional for anyone who wants to take a screenshot without updating what is actually visible to the user.

### DIFF
--- a/Modules/Scripted/ScreenCapture/ScreenCapture.py
+++ b/Modules/Scripted/ScreenCapture/ScreenCapture.py
@@ -1007,7 +1007,7 @@ class ScreenCaptureLogic(ScriptedLoadableModuleLogic):
 
     return sliceOffsetResolution
 
-  def captureImageFromView(self, view, filename=None, transparentBackground=False, volumeNode=None):
+  def captureImageFromView(self, view, filename=None, transparentBackground=False, volumeNode=None, processEvents=True):
     """
     Capture an image of the specified view and store in the specified object.
 
@@ -1015,8 +1015,10 @@ class ScreenCaptureLogic(ScriptedLoadableModuleLogic):
     :param filename: Filename of the desired output file. If none, no file will be written.
     :param transparentBackground: Set the background to be transparent for single-view captures.
     :param volumeNode: Vector volume node to store the capture image. If none, no vector volume node will be updated.
+    :param processEvents: Whether to process events before capturing the image. If true the view will visibly update before capture.
     """
-    slicer.app.processEvents()
+    if processEvents:
+      slicer.app.processEvents()
     if view:
       if type(view)==slicer.qMRMLSliceView or type(view)==slicer.qMRMLThreeDView:
         view.forceRender()


lassoan (commented):
`slicer.app.processEvents()` ensures that up-to-date content is rendered I don't understand how removing it would help any use case. Can you  provide a screen capture video that illustrates your problem?

To document segmentation changes, you can record what the user saw and/or record the actual segmentation (not just some screenshot of it).

nathanbmnt:
Here is a video showing how slicer.app.processEvents makes it visible that the slice offsets are changing every time I click. When slicer.app.processEvents is removed from captureImageFromView the screenshots are the same but the slice offsets changing is not visible.
https://user-images.githubusercontent.com/40271210/129369443-d253e7c1-fb6a-4bb3-9316-251c90db29cd.mp4

jamesobutler:
The purpose is part of some custom workflow where Screenshot are silently collected at time of modifying a segment to then be included and saved to disk next to the segmentation file. These screenshots are taken with the slice offset at the center of the segment in all views which is not necessarily the location where the segment was being modified. These changes occur silently and then the state is restored at the end of the process. Evidently @nathanbmnt can make those changes of slice offset, other segmentation visibilities, and take the screen shot without necessarily rendering these changes to the user. Rendering all the changes can be jarring as it flashes quickly.

More generally the question may be how to do certain actions silently to the user by handling when certain display updates are rendered to the user in the layout.

lassoan:
The flickering of the views certainly looks bad, so I understand that you want to avoid it. I'm surprised that simply removing `slicer.app.processEvents()` allows you to capture a screenshot of the updated view content, without showing anything to the user. It might not work reliably on all operating systems and under all circumstances. Integrating this small change would be very easy, but it would be better if you could explore this a bit more, because both you and Slicer core would benefit from a better-understood, more robust design.

Instead of relying on a transient state not becoming visible for users, you could render screenshots using view nodes that are outside of the view layout (see example [here](https://slicer.readthedocs.io/en/latest/developer_guide/script_repository.html#show-a-slice-view-outside-the-view-layout)). The views may not have to be visible to the user (it may be hidden or minimized or outside the visible screen area).

nathanbmnt:
Most of our screenshots are of the whole axial+sagittal+coronal 4-up view. Is it possible to setup multiple views using the technique [here](https://slicer.readthedocs.io/en/latest/developer_guide/script_repository.html#show-a-slice-view-outside-the-view-layout) into a single widget? Maybe a qt widget with a grid layout could be made and each slice widget is put in it?

lassoan:
You can certainly stitch screenshots taken from multiple views into a single image, using either VTK or Qt. See for example how it can be done using VTK here: https://github.com/Slicer/Slicer/blob/4de64d927e06e32832631673cbe8185053400ad1/Modules/Scripted/ScreenCapture/ScreenCapture.py#L1362-L1402

nathanbmnt:
Thanks Andras. I'm using a custom slice view outside the view layout, but an odd glitch is happening where segmentations loaded from file are invisible in the custom slice view. However, new segmentations I create in scene are visible. This glitch does not happen in vanilla Slicer only in my custom app. My custom app has code has custom code for loading in segmentations could you think of any reason the custom slice view would not show segmentations loaded from file?

In this screenshot Markup_4 (pink) is the only one visible in my external slice view window. Markups 1-3 (blue, orange, yellow) are invisible because they were loaded from file.
image

jamesobutler:
@nathanbmnt It seems to indicate that the custom app is setting the display node of the segmentation nodes to be visible only in specific view nodes rather than all view nodes. That would explain why it isn't shown in the different view node that is not "Red", "Yellow" or "Green". Code that would remove view node IDs and only support it in specific views would utilize methods such as `RemoveAllViewNodeIDs` and `AddViewNodeID`.

nathanbmnt:
ScreenCapture is functioning with the external slice view widget as long as I feed in the sliceview into `captureImageFromView`

Closing this PR as the external view works well enough for my use case.

lassoan:
Thank you. I think your current solution is cleaner and more efficient than the temporary hijacking of user's views.

nathanbmnt:
Unfortunately screen capture is not working when the widget is hidden, I've made a discourse post about it [here](https://discourse.slicer.org/t/screenshotting-a-hidden-external-view-widget/19258)

lassoan:
For future reference the final solution was this:

> Showing the widget and then immediately hiding it right after creation does not work unfortunately. I can however screenshot when the widget is outside of the monitor bounds, so I’ve just used qwidget.setGeometry to put it out of bounds

PR #5795, opened 2021-08-13
COMP: Fix Girder_v1 upload of extension with semi-colon in any metadata

jcfr:
This commit generalizes to all metadata the change originally introduced
in 23e8fb730f (COMP: Fix upload of extension with semi-colon in description
metadata) and intended to fix a regression previously introduced in
8f33d93e8c (ENH: Ensure dependency metadata is published with extension
packages)

It fixes issue reported in https://github.com/KitwareMedical/slicer-extensions-webapp/issues/44


jcfr:
After confirming the fix work as expected (see https://github.com/KitwareMedical/slicer-extensions-webapp/issues/44#issuecomment-898262586), I am merging.

PR #5798, opened 2021-08-15
BUG: ExtensionsManager: Fix searchBox update after switching tab

jcfr:
This commit ensures the search box text is properly reset when
using Girder_v1 server API integration.

In that case, the value associated with "q" query parameter should
be retrieved.


jcfr:
@lassoan  Since I created this PR without having the chance to compile, it would be great if you  could test.

_Fun fact: Currently in ✈️ with my mac 💻  and focusing on web dev  but without any Slicer build tree ..._

lassoan (approved):
Tested and it works well.

PR #5799, opened 2021-08-16
ENH: Segment Editor effects placed into categories

jamesobutler:
This PR aims to improve understanding of different segment editor effects by placing them into categories.

Based on original planning from @upersad and @lassoan at https://discourse.slicer.org/t/segment-editor-user-interface-improvements/9029

| Current | This PR |
|----------|---------|
|image|image|

### DIFF
--- a/Modules/Loadable/Segmentations/Widgets/Resources/UI/qMRMLSegmentEditorWidget.ui
+++ b/Modules/Loadable/Segmentations/Widgets/Resources/UI/qMRMLSegmentEditorWidget.ui
@@ -272,6 +272,7 @@
      <property name="alignment">
       <set>Qt::AlignJustify|Qt::AlignVCenter</set>
      </property>
+     <layout class="QFormLayout" name="formLayout_2"/>
     </widget>
    </item>
    <item>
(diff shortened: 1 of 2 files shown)


lassoan:
Thanks for working on this. Improving the Segment Editor layout would be great. It is a very hard task, so we need to involve the community to bring their ideas and concerns. Would you mind reopen any of the related discussions on the Slicer forum (or create a new discussion and reference the old threads)? Thanks!

jamesobutler:
See https://discourse.slicer.org/t/organization-of-segment-editor-effects-into-categories/19208 for a new thread about placing segment editor effects into categories.

jamesobutler:
The discourse post resulted in feedback that was generally negative of putting the segment editor effects into categories for better understanding of the type of effect.  For now they will continue to be roughly organized by default order. See
https://github.com/Slicer/Slicer/blob/fb486aedd8c887981f973b037f6a08cda2571fb8/Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.cxx#L332-L344

lassoan:
The issue was addressed in this commit: https://github.com/Slicer/Slicer/commit/1b94b0d5330b277c698f58d2fe09901875ba2643. Therefore, this pull request is not relevant anymore.

Issue #5752, opened 2021-07-25
Improve markups rendering performance


lassoan:
To ensure that rendering of markups for reasonable-sized scenes (something like 500 markups) is fluid to render even with interaction and label display enabled:

`vtkSlicerMarkupsWidgetRepresentation3D::GetViewScaleFactorAtPosition`
When I attach VerySleepy to a release-mode version (which is not totally reliable) then I keep seeing `vtkSlicerMarkupsWidgetRepresentation3D::GetViewScaleFactorAtPosition` as a performance bottleneck. Computing the scale factor should be a very inexpensive operation (essentially multiplication with a 4x4 matrix). However, how we compute the scale now (using `Renderer->WorldToDisplay`) is very costly.

Making `vtkSlicerMarkupsWidgetRepresentation3D::GetViewScaleFactorAtPosition` return a fixed value (instead of actually computing the scale) seem to make a significant difference. Could you verify this and if you confirm that this has significant impact, could you think about a solution? Either fixing inefficiency of `WorldToDisplay` or by implementing a fast/batch version of it.

Reduce the number of actors
We added a separate actor for displaying almost identical things, just with different colors. We could modulate the color (and even the markup type) using a point data array.

Could you try to merge the Unselected+Selected+Active pipeline and the Project+ProjectBack pipeline? That would allow us to go from 5 actors to 2 actors per control point, which could be a significant improvement.

Check if profiling `RelWithDebInfo` brings up anything else
For example, would disabling `vtkDebugLeaks` make a difference? It is quite expensive now (collects call stack, etc.) and I think it is enabled for release-mode builds on the factory.

Connor-Bowley:
@lassoan  I can confirm that on Linux (Ubuntu 20.04) that having `vtkSlicerMarkupsWidgetRepresentation3D::GetViewScaleFactorAtPosition` return a fixed value has significant impact (using perf and time for rough comparisons). 

Can you provide more information on how you would recommend computing the scale factor without use of WorldToDisplay?

Connor-Bowley:
Also, were you creating 1 markup node with many markups, or did you have many markup nodes with only 1 markup per node?

Connor-Bowley:
https://github.com/Connor-Bowley/Slicer/commit/5b966f55238c39410735ecc073c52d7f2310d9f1

[...]

PR #5800, opened 2021-08-16
Improve markups rendering performance

Connor-Bowley:
Added fast `WorldToDisplay` to `vtkMRMLInteractionEventData` which will cache the Composite Projection Transform matrix from the camera on first use.

This seems to increase the FPS of scenes with high amounts of fiducial markups by 1-2 FPS. Therefore I do not think this solves the whole of the performance problem mentioned in https://github.com/Slicer/Slicer/issues/5752, but is a marginal improvement towards it.

### DIFF
--- a/Libs/MRML/DisplayableManager/vtkMRMLCameraWidget.cxx
+++ b/Libs/MRML/DisplayableManager/vtkMRMLCameraWidget.cxx
@@ -615,7 +615,7 @@ bool vtkMRMLCameraWidget::ProcessRotate(vtkMRMLInteractionEventData* eventData)
     return true;
     }
 
-  int *size = this->Renderer->GetRenderWindow()->GetSize();
+  const int *size = this->Renderer->GetRenderWindow()->GetSize();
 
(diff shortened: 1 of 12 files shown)


Connor-Bowley:
@lassoan @jcfr This should be ready for review now. All the changes from `int*` to `const int*` were to compile with changes I had made to VTK that didn't end up getting accepted. There is nothing wrong with those changes, but they don't really affect the performance. If you are not opposed I would like to keep them in because, in my opinion, it makes the code more clear that we are using those values and not changing them.

lassoan (approved):
Thank you, I've tested this and it works well. It indeed measurably improves markups rendering performance. I agree that the added `const` keywords are good, not necessary, but makes the intent of the code a bit more clear.

Connor-Bowley:
The `const`s were added to allow a different improvement I had made to compile, but that change ended up breaking VTK's interface to be useful so it was backed out. I agree the `const`s became unnecessary at that point, but as you said they give better intent and I had already done the (pretty trivial) work of putting them in, so I just didn't remove them.

PR #5801, opened 2021-08-16
BUG: Fix wrapping for methods in vtkMRMLMarkupsNode

Sunderlandkyl:
Some methods in vtkMRMLMarkupsNode returned double* but did not contain a size hint. As a result the wrapped methods were not usable in python. Fixed by adding VTK_SIZEHINT to the following methods:
- GetNthControlPointPosition
- GetNthControlPointOrientationMatrix
- GetNthControlPointNormal


lassoan (approved):
Looks good, thank you!

PR #5803, opened 2021-08-17
BUG: Update interaction handle orientation when transform is hardened

### LINKED ISSUE #5539: Interactive handles of markups plane don't transform well after transform is hardened
interactiveHandles of plane don't transform well after transform is hardened

## Steps to reproduce

1. Create a plane
2. Activate the interaction handles
3. Create a new linear transform and set up a rotation
4. Apply the linear transform to the plane
5. Harden the transform

Screenshot9

## Expected behavior

Screenshot_1

## Environment
- Slicer version: Slicer-4.13.0-2021-03-09
- Operating system: Windows x64

Sunderlandkyl:
The plane node interaction handle to world matrix was not updated when the transform was hardened.
Fixed by updating the matrix in vtkMRMLMarkupNode::OnTransformNodeReferenceChanged.

Also prevents the plane position from "flickering" by pausing modified events while vtkMRMLNode::ApplyTransform is being called.

Fix #5539


lassoan (approved):
Looks good to me thank you!

Issue #4989, opened 2020-06-16
Add "Export as..." entry to Data context menu


pieper:
## Is your feature request related to a problem? Please describe.

The Save As dialog requires many clicks to perform file saving, but for some use case people want to just move data to another app and need a streamlined method.

## Describe the solution you'd like

* When one or more nodes are selected in the Data module and the context menu (right click) there would be an "Export as..." entry.  
* Activating the "Export as..." would bring up a submenu.  
  * If all the selected nodes are of the same storable type, then the submenu would contain one entry for each file type supported by the storable.
  * If the nodes are of different types, then the submenu would have one entry "`N` nodes to default formats" where N is the number of selected nodes
  * If the selection includes only non-storable nodes the "Export as..." option would be disabled
  * If the selection is mixed storable and non-storable the submenu would say "`M` of `N` nodes to default formats"

On selecting, a directory browser dialog would open, and when a valid directory is picked the action would use the [temporary storage node method](https://discourse.slicer.org/t/markupsline-fcsv-loads-as-markupsfiducials/11545/37?u=pieper) to save in the indicated formats so the scene is not modified.  File basenames would be the node names.

## Describe alternatives you've considered

See [discussion here](https://discourse.slicer.org/t/markupsline-fcsv-loads-as-markupsfiducials/11545).

## Additional context

Discussion of SH plugin implementation between @pieper and @cpinter :

> Hi Csaba -

So you probably saw the pull request - small thing I figured we should fix.

I think I need it anyway, but I'm still stuck on what I'm trying to do, so I hope you can give a pointer (and not say it's impossible!).

Basically after talking it over with the SlicerMorph team we decided we want to add a context menu action for Export, which would have a submenu with all the valid file extensions for the selected node type.  Triggering the action would bring up a folder dialog and then save the selection.

But we also want to do this for multiple selections, and I can't find a way to add my action to the multiple select menu, which seems to be hardcoded to be delete and toggle visibility.

So my two questions are:

* can I generate a dynamic submenu in my plugin?

* can I add an action to the multiple select context menu?

Thanks in advance!
-Steve

Hi Steve,

 

This is great! I’m glad subject hierarchy is being used in so many projects. Of course it still has its limitations (and bugs).

 

* can I generate a dynamic submenu in my plugin?

[...]

PR #5805, opened 2021-08-18
ENH: Incorporate ExportAs module into Slicer core

ebrahimebrahim:
See issue #4989

[ExportAs](https://github.com/SlicerMorph/SlicerMorph/tree/master/ExportAs) is a Slicer module that was created for the extension SlicerMorph.
This commit incorporates ExportAs into Slicer core, renaming the module from
`ExportAs` to `DataSubjectHierarchyPlugin`.

This subject hierarchy plugin is associated with the Data module as a whole,
rather than with any specific module. This is why I am organizing it as
a `SubjectHierarchyPlugin` under `Modules/Loadable/Data`. Please let me know if
there are better ways to go about organizing this.  

I followed `Modules/Loadable/Annotations/SubjectHierarchyPlugins/`
as a model for how to set this up, since it is an example of a subject hierarchy
plugin that is written in python and added under a loadable module. 

cc: @pieper 

---

Some problems:

(1) Since the module is now initialized in
`Modules/Loadable/Data/qSlicerDataModule.cxx`, the class derived from
`ScriptedLoadableModule` is no longer needed. But removing it did away with
data that attributes the module code to the original author. Here is the part
that was deleted:
```
class ExportAs(ScriptedLoadableModule):
    def __init__(self, parent):
        parent.title = "SlicerMorph ExportAs Plugin"
        parent.categories = [""]
        parent.contributors = ["Steve Pieper, Isomics, Inc."]
        parent.helpText = ""
        parent.hidden = not slicer.util.settingsValue('Developer/DeveloperMode', False, converter=slicer.util.toBool)

        #
        # register subject hierarchy plugin once app is initialized
        #
        def onStartupCompleted():
            import SubjectHierarchyPlugins
            from ExportAs import ExportAsSubjectHierarchyPlugin
            scriptedPlugin = slicer.qSlicerSubjectHierarchyScriptedPlugin(None)
            scriptedPlugin.name = "ExportAs"
            scriptedPlugin.setPythonSource(ExportAsSubjectHierarchyPlugin.filePath)
        slicer.app.connect("startupCompleted()", onStartupCompleted)
```
I wonder, how should the attribution be incorporated?

(2) While the *Export as* context menu option seems to work great, I wasn't
able to test the *Export transformed as* option. It didn't show up when I
thought it should. We can probably move forward and address that feature later.
Or maybe I misunderstood how it's meant to work.

(3) Not all tests pass on my linux system. However CDash shows that there is currently a handful of failing tests on linux,
so this might not be due to my changes. Some tests may be failing because of quirks in my particular environment.
Here are the tests that failed on my system:
```
cmake_slicer_generate_extension_description_test (Failed)
py_cmake_slicer_extensions_index_build_with_upload (Failed)
py_cmake_slicer_extensions_index_build_with_upload_using_ctest (Failed)
qMRMLLayoutManagerVisibilityTest (Failed)
qMRMLNavigationViewTest1 (Failed)
qMRMLThreeDWidgetTest1 (Failed)
qMRMLNavigationViewEventTranslatorPlayerTest1 (Failed)
qMRMLThreeDWidgetEventTranslatorPlayerTest1 (Failed)
qSlicerUtilsTest1 (Failed)
qSlicerExtensionsManagerModelTest (Failed)
py_AnnotationsTestingAddManyROIs (Failed)
py_PluggableMarkupsSelfTest (Failed)
py_nomainwindow_SegmentationsModuleTest2 (Failed)
py_SlicerRestoreSceneViewCrashIssue3445 (Failed)
py_SurfaceToolbox (Failed)
```

### DIFF
--- a/Modules/Loadable/Data/SubjectHierarchyPlugins/DataSubjectHierarchyPlugin.py
+++ b/Modules/Loadable/Data/SubjectHierarchyPlugins/DataSubjectHierarchyPlugin.py
@@ -1,0 +1,147 @@
+import logging
+import vtk, qt, ctk, slicer
+from slicer.ScriptedLoadableModule import *
+
+from SubjectHierarchyPlugins import AbstractScriptedSubjectHierarchyPlugin
+
+class DataSubjectHierarchyPlugin(AbstractScriptedSubjectHierarchyPlugin):
+  """ Scripted subject hierarchy plugin for the ExportAs module.
+  """
+
+  # Necessary static member to be able to set python source to scripted subject hierarchy plugin
+  filePath = __file__
+
+  def __init__(self, scriptedPlugin):
+    AbstractScriptedSubjectHierarchyPlugin.__init__(self, scriptedPlugin)
+
+    pluginHandlerSingleton = slicer.qSlicerSubjectHierarchyPluginHandler.instance()
+    self.subjectHierarchyNode = pluginHandlerSingleton.subjectHierarchyNode()
+
+    self.exportAsAction = qt.QAction(f"Export as...", scriptedPlugin)
+    self.exportTransformedAsAction = qt.QAction(f"Export transformed as...", scriptedPlugin)
+    self.menu = qt.QMenu("Plugin menu")
+    self.transformedMenu = qt.QMenu("Transformed plugin menu")
+    self.exportAsAction.setMenu(self.menu)
+    self.exportTransformedAsAction.setMenu(self.transformedMenu)
+
+  #
+  # item context menus are what happens when you right click on a selected line
+  #
+  def itemContextMenuActions(self):
+    """the actions that could be shown for any of this plugins items"""
+    return [self.exportAsAction, self.exportTransformedAsAction]
+
+  def showContextMenuActionsForItem(self, itemID):
+    """Set actions visible that are valid for this itemID"""
+    # reset all menus
+    self.exportAsAction.visible = False
+    self.exportAsAction.enabled = False
+    self.exportTransformedAsAction.visible = False
+    self.exportTransformedAsAction.enabled = False
+    self.menu.clear()
+    self.transformedMenu.clear()
+    # check if this is parent or leaf
+    childIdList = vtk.vtkIdList()
+    self.subjectHierarchyNode.GetItemChildren(itemID, childIdList)
+    if childIdList.GetNumberOfIds() > 0:
+      self.showContextMenuActionsForParentItem(itemID, childIdList)
+    else:
+      self.showContextMenuActionsForLeafItem(itemID)
+
+  def parseWriteFormats(self, writeFormats):
+    # convert ('Markups JSON (.json)', 'Markups Fiducial CSV (.fcsv)')
+    # to "Markups *.mrk.json *.json *.fcsv"
+    # and {'.mrk.json': 'Markups JSON *.mrk.json', '.fcsv': 'Markups Fiducial CSV *.fcsv')
+    allExtensionsFilter = ""
+    filtersByExtension = {}
+    formatsByExtension = {}
+    for writerExtension in writeFormats:
+      extension = writerExtension.split()[-1][1:-1]
+      formatsByExtension[extension] = writerExtension
+      allExtensionsFilter += " *" + extension
+      filtersByExtension[extension] = f"{' '.join(writerExtension.split()[:-1])} *{extension}"
+    return allExtensionsFilter, filtersByExtension, formatsByExtension
+
+  def showContextMenuActionsForParentItem(self, itemID, childIdList):
+    # only enable if all children are same class
+    itemDataNodes = []
+    for childIDIndex in range(childIdList.GetNumberOfIds()):
+      childID = childIdList.GetId(childIDIndex)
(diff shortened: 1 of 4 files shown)


Modules/Loadable/Data/SubjectHierarchyPlugins/DataSubjectHierarchyPlugin.py
```diff
@@ -0,0 +1,147 @@
...
+
+  def __init__(self, scriptedPlugin):
+    AbstractScriptedSubjectHierarchyPlugin.__init__(self, scriptedPlugin)
+
+    pluginHandlerSingleton = slicer.qSlicerSubjectHierarchyPluginHandler.instance()
+    self.subjectHierarchyNode = pluginHandlerSingleton.subjectHierarchyNode()
+
+    self.exportAsAction = qt.QAction(f"Export as...", scriptedPlugin)
+    self.exportTransformedAsAction = qt.QAction(f"Export transformed as...", scriptedPlugin)
+    self.menu = qt.QMenu("Plugin Menu")
```
cpinter, line 22: In Subject hierarchy basically all plugins use sentence case. If that's not a big deal please change these.

cpinter:
Thanks for the contribution. I like adding a Data SH plugin for this, and I agree with the structure of the files. I added a minor comment, otherwise I'm fine with the change.

I'll leave (1) and (2) for @pieper to answer, and about (3) let me know if you want me to run the test on Windows.

pieper:
Thanks for taking this on, I'm happy to see this move into the Slicer core.

Regarding (1), I know you were following a precedent, but I'd much rather see this code in `Modules/Scripted`.  Calling python code from C++ should be avoided in general because it makes debugging, maintenance, and further development much more difficult.  By keeping this as a scripted module it can serve as a working example for extension developers who need a template for adding their own subject hierarchy (data module) plugins.  The subject hierarchy is well documented, but complex enough that self-contained working examples that provide real functionality are important.

Regarding (2) this sounds like a regression that should be fixed before the code is merged.  Exporting transformed data was a use case specifically requested by SlicerMorph users.

As for (3), about the tests, it should be easy to run all the tests with and without the module enabled to see if there are any regressions specifically related to this module.  I suggest running them in a VM to avoid any cross-talk with your desktop.

Also I'd suggest the name `ExportAsDataPlugin` because it is much more specific and descriptive than `DataSubjectHierarchyPlugin`.  "Data" and "SubjectHierarchy" are redundant at this point since the original SubjectHierarchy replaced the Data module in the user interface.  Also that name will not clash with SlicerMorph's ExportAs implementation so it can be turned off when no longer needed but won't break in the mean time.

The current module has a clear purpose and is a good size for a plugin (not to complex).  If we have other plugins for the Data module they should be implemented in distinct modules unless there is a clear reason for them to depend on each other (or share a superclass).  They could all share the `DataPlugin` suffix to indicate they belong together.  We could also consider putting them in a `Modules/Scripted/DataPlugins` directory if there are a few of them.

lassoan:
My main concern is that the feature is just not mature enough to be included in Slicer core. Before [these issues](https://github.com/SlicerMorph/SlicerMorph/issues/72) are fixed, I don't think it is a good idea to add it to the Slicer core.

About where to put the module: I agree that registering a Python plugin from C++ is not nice and it should be either a clean C++ or Python solution (we just chose the hybrid solution in Annotations module as a temporary hack, as Annotations module will be removed soon anyway). While Python is essential in many projects, it is still just an optional dependency of Slicer, so I would prefer keep basic Slicer features, such as data export implemented without depending on Python. The feature is really small, so it should be no problem to translate it to C++.

cpinter:
I agree with many of the previous points, but let me give a reason why I thought this structure could be good enough:

* The Annotations module will be removed soon, so I think the most important thing here is that we have a Python SH plugin for a C++ module in Slicer core that can be used as an example later if needed
    * By the way for this I encourage adding all the empty functions from the Annotations SH plugin to indicate the developers
what functions are available without having to see the C++ abstract SH plugin
    * Calling Python from C++ is not nice, but debugging hybrid C++/Python is practically impossible, so it won't make things worse (or there is something I don't know about debugging Python/C++, for example when a C++ call stack simply says python call and there ends, where to pick it up in Python)
    * That said, a standalone Python SH plugin is also very useful for this purpose, so the separate module makes sense too
    * Porting this to C++ would be nice and clean, but then we'll be left without a Python SH plugin after the removal of Annotations
* The name of the plugin if it was indeed the SH plugin of the Data module follows the conventions
    * For now, every SH plugin in Slicer core has `SubjectHierarchyPlugin` in their names, because there are various types of plugins, and calling it simply `ExportAsDataPlugin` could be confusing for this reason I think

I agree that the pending issues should be fixed before integrating this though

lassoan:
> the most important thing here is that we have a Python SH plugin for a C++ module in Slicer core that can be used as an example later if needed

This could be accomplished by a test, too. But I agree that keeping a Python implementation can be acceptable (the plugin can be converted between Python and C++ anytime anyway). 

The important prerequisite for the merge is fixing the usability issues.

ebrahimebrahim:
See https://github.com/Slicer/Slicer/issues/4989#issuecomment-904779802

Issue #5673, opened 2021-06-03
Bump from C++11 to C++14


jamesobutler:
ITK has recently updated their master branch to bump from C++11 to C++14 as a minimum. Slicer should follow to continue following closely with latest ITK development.

See https://discourse.itk.org/t/may-we-start-using-c-14-at-the-master-branch/4172
https://github.com/InsightSoftwareConsortium/ITK/commit/daec0fdd703b1bca16aeb07bc4e0a01101da950a
https://github.com/InsightSoftwareConsortium/ITK/commit/4e812d60743a3ed7c09647e7f3f0e8498583c2da
https://github.com/InsightSoftwareConsortium/ITK/commit/22bce725aec99889d8428c6d8400ca5fb76cd921

- Remember to update build requirement documentation.
- Remember to update SlicerBlockPlatformCheck which is already quite out of date
https://github.com/Slicer/Slicer/blob/4d13db4c5d707b96844ff41e072aa0f30ce7fbf6/CMake/SlicerBlockPlatformCheck.cmake#L58-L72

lassoan:
Thanks for the information. We indeed need to be careful when we update to this ITK version and make the necessary documentation updates and checks. Probably we will keep the current ITK for the upcoming interim release (that comes within a few weeks) but switch before Slicer5 (planned for this summer).

Do you know if ITK already has platform checks in place? If yes then we may just rely on those, or maybe copy-paste them into Slicer so that users get feedback more quickly.

jamesobutler:
For updating to ITK 5.3 which has the bump to C++14, VS2015 (v140 toolset) is not a supported compiler (https://github.com/InsightSoftwareConsortium/ITK/blob/master/Documentation/SupportedCompilers.md). Therefore additional changes to the ones integrated in #5776 will be necessary to error at the usage of v140 during configuration.

jamesobutler:
Optimization to latest minimum c++ version I’m sure is still needed but the bump of the minimum version was completed in https://github.com/Slicer/Slicer/commit/bc2b3f4049ef32486cf43ad16903820fc8767f27.

PR #5807, opened 2021-08-19
Enable GrowCut remote module in ITK

dzenanz:
GrowCut remote module in ITK has recently been updated to mirror Slicer's improvements. This PR contains pre-requisites to removing this code from Slicer and refactoring it to use the version coming from ITK.

I have not begun the Slicer refactoring, and I would prefer if somebody else did that. I could support them by making changes inside `ITKGrowCut`.

### DIFF
--- a/SuperBuild/External_ITK.cmake
+++ b/SuperBuild/External_ITK.cmake
@@ -146,6 +146,7 @@ if(NOT DEFINED ITK_DIR AND NOT Slicer_USE_SYSTEM_${proj})
       -DModule_ITKIOMINC:BOOL=ON
       -DModule_IOScanco:BOOL=ON
       -DModule_MorphologicalContourInterpolation:BOOL=ON
+      -DModule_GrowCut:BOOL=ON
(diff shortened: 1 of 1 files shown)


jamesobutler:
`ENH: require at least C++14` will be part of #5673 to resolve. That priority doesn’t appear to be immediate at the moment unless you’d like to complete this work here sooner.

jamesobutler:
`ITKGrowCut` can now be used now that #5954 has completed the prerequisites.

dzenanz:
Just enabling ITK's module is easy, refactoring https://github.com/Slicer/Slicer/blob/master/Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorGrowFromSeedsEffect.py to use ITK's implementation instead of https://github.com/Slicer/Slicer/blob/master/Modules/Loadable/Segmentations/Logic/vtkImageGrowCutSegment.cxx is the hard part.

sjh26 (approved):
lgtm.  We will merge this, then look at the refactoring

PR #5811, opened 2021-08-23
ENH: Improve handling of markups control point orientation

Sunderlandkyl:
Previously the memory layout used to store for control point orientation in memory was: [x0, x1, x2, y0, y1, y2, z0, z1, z2], which was different from the implementation of vtkMatrix3x3. This commit changes the memory orientationto the one used in VTK: [x0, y0, z0, x1, y1, z1, x2, y2, z2].

The method vtkMRMLMarkupsNode::SetNthControlPointOrientationMatrixWorld previously did not actually set the orientation in world coordinates, it simply used the orientation matrix in world coordinates to directly update the OrientationMatrix. This commit transforms the orientation matrix from world to node coordinate system, and also adds a GetNthControlPointOrientationMatrixWorld function.

This commit also adds a test to check the orientation of each control point before and after applying linear and b-spline transform nodes to it.

### DIFF
--- a/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsNode.cxx
+++ b/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsNode.cxx
@@ -40,6 +40,7 @@
 #include <vtkCollection.h>
 #include <vtkParallelTransportFrame.h>
 #include <vtkGeneralTransform.h>
+#include <vtkMatrix3x3.h>
 #include <vtkMatrix4x4.h>
 #include <vtkNew.h>
 #include <vtkObjectFactory.h>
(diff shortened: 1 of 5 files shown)


Modules/Loadable/Markups/MRML/vtkMRMLMarkupsNode.cxx
```diff
@@ -923,6 +923,44 @@ void vtkMRMLMarkupsNode::SetNthControlPointPositionWorldFromArray(
...
+  if (!this->GetParentTransformNode())
+    {
+    std::copy_n(orientationMatrix_Node, 9, orientationMatrix_World);
+    }
+  else
+    {
+    // Transform orientation matrix from world
+    vtkNew<vtkGeneralTransform> nodeToWorldTransform;
+    this->GetParentTransformNode()->GetTransformToWorld(nodeToWorldTransform);
+    nodeToWorldTransform->TransformVectorAtPoint(position_Node, &orientationMatrix_Node[0], &orientationMatrix_World[0]);
```
lassoan, line 939: usually we store matrices like this in an array:

```txt
m0 m1 m2
m3 m4 m5
m6 m7 m8
```

and

```txt
m0 m1 m2 m3
m4 m5 m6 m7
m8 m9 m10 m11
m12 m13 m14 m15
```

Is this convention followed here, too?

If yes, then the vectors that we need to transform are not `orientationMatrix_Node[i], orientationMatrix_Node[i+1], orientationMatrix_Node[i+2]`, but `orientationMatrix_Node[i], orientationMatrix_Node[i+3], orientationMatrix_Node[i+6`].

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsNode.h
```diff
@@ -433,9 +433,9 @@ class  VTK_SLICER_MARKUPS_MODULE_MRML_EXPORT vtkMRMLMarkupsNode : public vtkMRML
   /// Set orientation as 9 values: x, y, and z axis directions, respectively.
   void SetNthControlPointOrientationMatrix(int n, double orientationMatrix[9]);
   /// Set orientation as 9 values: x, y, and z axis directions, respectively.
-  /// Important: this method just calls SetNthControlPointOrientationMatrix and
-  /// does not convert from world coordinate yet.
   void SetNthControlPointOrientationMatrixWorld(int n, double orientationMatrix[9]);
+  /// Get orientation as 9 values: x, y, and z axis directions, respectively.
```
lassoan, line 437: Until now, orientation was not tested/incomplete. Now that you fix it, we should do it properly.

The order of elements is not consistent now:
- (x0 x1 x2 y0 y1 y2 z0 z1 z2) order: this is used in the documentation here and in the new function that you added
- (x0 y0 z0 x1 y1 z1 x2 y2 z2) order: this is used [in the schema](https://github.com/Slicer/Slicer/blob/39be4408f084b3e5d632b373a4985cd7fa6c661b/Modules/Loadable/Markups/Resources/Schema/markups-schema-v1.0.2.json#L163) and in [in vtkMatrix3x3](https://github.com/Kitware/VTK/blob/9e838ccf55a67da763399693505907d6ce7222c3/Common/Math/vtkMatrix3x3.cxx#L54-L62)

We don't really have the liberty to choose this, because vtkMatrix3x3 will not change. If we chose a different interpretation then it would lead to errors in the future. So, I think we should standardize on the vtkMatrix3x3 interpretation, even if previous behavior was different.

lassoan (commented):
Thank you, this will be great, just need to clean up the array element ordering.

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsNode.cxx
Sunderlandkyl: No, convention is not followed here. See: https://github.com/Slicer/Slicer/blob/39be4408f084b3e5d632b373a4985cd7fa6c661b/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsNode.h#L111-L113

I take it then that we should change it to use the normal matrix convention?

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsNode.h
Sunderlandkyl: > So, I think we should standardize on the vtkMatrix3x3 interpretation, even if previous behavior was different.

Ok, this addresses my previous question. I will update the class to use the order of elements from VTK.

lassoan: Sounds good. Let me know when it is ready for review.

Sunderlandkyl:
@lassoan Ok, I've updated orientation element ordering and added a test to check that adding parent transforms affects the orientation in world coordinates as expected.

lassoan (approved):
This looks perfect, thank you.

PR #5814, opened 2021-08-24
Maximize and Minimize feature

jahnavi0801:
This feature can be found on the controller bars which helps the user to maximize and minimize each view. 

* `qMRMLSliceControllerWidget` for Red Slice only, yellow slice only, and Green slice only.
* `qMRMLThreeDViewControllerWidget` for 3D view only.

**Working of Red Slice button**

Minimized Red slice
Maximized Red slice

**Working of 3D View**

Minimized 3D view
Maximized 3D View


Libs/MRML/Widgets/qMRMLSliceControllerWidget.cxx
```diff
@@ -1997,6 +2007,40 @@ void qMRMLSliceControllerWidget::fitSliceToBackground()
   d->SliceLogic->EndSliceNodeInteraction();
 }
 
+void qMRMLSliceControllerWidget::MaximizeLayout() {
+  Q_D(qMRMLSliceControllerWidget);
+  const char* MaxLayoutName = d->MRMLSliceNode->GetLayoutName();
```
pieper, line 2012: Local variables should start with lower case

Libs/MRML/Widgets/qMRMLSliceControllerWidget.h
```diff
@@ -167,6 +169,8 @@ public slots:
   /// \sa setSliceView();
   void fitSliceToBackground();
 
+  void MaximizeLayout();
```
pieper, line 172: method names on classes inheriting from Qt classes should start with lower case letters.

Libs/MRML/Widgets/qMRMLThreeDViewControllerWidget.h
```diff
@@ -117,6 +117,9 @@ public slots:
   void setRulerType(int type);
   void setRulerColor(int color);
 
+  void Maximize3D();
```
pieper, line 120: again, lower case for method name

Libs/MRML/Widgets/qMRMLThreeDViewControllerWidget.h
```diff
@@ -128,6 +131,9 @@ protected slots:
 private:
   Q_DECLARE_PRIVATE(qMRMLThreeDViewControllerWidget);
   Q_DISABLE_COPY(qMRMLThreeDViewControllerWidget);
+
+public:
+  bool  Maximized3D;
```
pieper, line 136: here too, lower case

pieper (changes requested):
I like this feature and it's good to have but I believe we could improve the implementation and avoid hard coding layout names in the widget code.  The problem is that this makes assumptions about what layouts users or custom apps might want to switch among.

It could make sense for the view controllers to know the name of the layout that will maximize their view type, or it may be better for them to generate a new custom layout that maximizes their view explicitly.  This would scale better when there are many view controllers and you want to maximize any of them (e.g. if you are in one of the Compare layouts).

Similarly, I think it would be better to have the View node keep track of the layout that was active before maximizing one view and then swap back to that layout when minimizing.

Otherwise there are some code style issues I pointed out.  Not sure I flagged them all.

Libs/MRML/Widgets/qMRMLSliceControllerWidget.cxx
```diff
@@ -1997,6 +2007,40 @@ void qMRMLSliceControllerWidget::fitSliceToBackground()
...
+  vtkMRMLLayoutNode* layoutNode = vtkMRMLLayoutNode::SafeDownCast(this->mrmlScene()->GetSingletonNode("vtkMRMLLayoutNode", "vtkMRMLLayoutNode"));
+  if(QString::fromUtf8(MaxLayoutName).contains("Red")) {
+    if(!this->maximumView) {
+      layoutNode->SetViewArrangement(vtkMRMLLayoutNode::SlicerLayoutOneUpRedSliceView);
+      this->maximumView = true;
+    }
+    else{
+      layoutNode->SetViewArrangement(vtkMRMLLayoutNode::SlicerLayoutFourUpView);
+      this->maximumView = false;
+    }
```
jamesobutler, line 2022: > I think it would be better to have the View node keep track of the layout that was active before maximizing one view and then swap back to that layout when minimizing.

I agree with @pieper here. If I'm in "Side by side" layout (Red on left, Yellow on right), then when I maximize the Red view (Red Only) and then minimize it again, I would expect it go to my previous layout of "Side by side" and not "Four-Up".

lassoan:
Thanks a lot @jahnavi0801. I've reviewed your commit and concluded that to generalize it for all view layouts, removing hardcoded, repeated logic, add GUI with proper feedback (updating icon and tooltip depending on current state; hide the button for standalone widgets; add maximize/restore action to view context menu), and fix all the other issues raised by the reviewers would be easier for me to quickly implement than asking you to do it.

I've pushed my updates to this branch and I also created a short video of how it looks now:

https://user-images.githubusercontent.com/307929/131253976-f216217c-5f49-4270-839e-b2cb9e10526e.mp4

jahnavi0801:
* @lassoan, @pieper, and @jamesobutler Thank you for reviewing and updating the PR. Given that the feature is working is there anything else you want me to do regarding this PR. 
* Currently it is showing "Merging is blocked".

lassoan:
I'll push an update soon that adds maximize/restore on double-click, updates the user manual, and references related issues.

lassoan:
All the planned changes are now completed and committed into this branch. A video that shows all 3 ways to maximize/restore views:

https://user-images.githubusercontent.com/307929/131416102-3eb25cfc-83ee-4691-bd6e-a118a5b3deb6.mp4

If there are no particular concerns then I'll merge this tomorrow so that everybody can easily try it and give feedback.

I'll update the documentation screenshots in a few weeks. I did not want to spend time with updating images until users had a chance to provide feedback.

jamesobutler (commented):
Any conflict with Double left click to expand viewer and the Draw segment editor effect’s “Left double-click to add a point and fill the contour.”?

When maximizing/minimizing the layout does the request for current layout reflect the new maximized view or does it return the original restored layout enum regardless of maximize/minimize? I think I would expect the former. 

Does the Slice layout toolbar current action icon still continue to reflect the current layout?

lassoan:
> Any conflict with Double left click to expand viewer and the Draw segment editor effect’s “Left double-click to add a point and fill the contour.”?

No conflicts. The widget system manages routing of the events cleanly. The view only gets a chance to process double-click events if nothing else needed it (not over a markup, not needed by Segment Editor effects. etc.). As all other widget events, this can be easily disabled or remapped to any other mouse or keyboard shortcuts.

> When maximizing/minimizing the layout does the request for current layout reflect the new maximized view or does it return the original restored layout enum regardless of maximize/minimize? I think I would expect the former.

As I wrote in the commit comment, the current layout ID is not changed. There is just an additional MaximizedView node reference that can be set in the layout node. If that is set then the layout manager sends an XML to CTK that only displays a single view. E.g., you are still in 4-up layout, just temporarily made one view to take up the entire layout area.
 This is similar to how we keep the layout "Dual 3D" even if you resize the 3D views so that the slice views are not visible anymore.

> Does the Slice layout toolbar current action icon still continue to reflect the current layout?

Yes. The current layout ID or description in the MRML layout node does not change.

jamesobutler:
@lassoan If you maximize a view, do you think it would be helpful to also restore the original layout by triggering the layout in the layout toolbar selector? So if I maximize the red slice of the Four-Up view, then accidentally forget my original layout is Four-Up and then choose Four-Up in the layout toolbar selector, should that restore it as well? Currently it does nothing I'm assuming because it detects that there is no layout change to process.

lassoan:
I agree, it make sense to restore the view layout in this case, even if it is the same as the current layout. I'll push an update today.

Issue #5804, opened 2021-08-18
Build on Ubuntu fails due to strict Qt version check


lassoan:
When following linux build instructions on Ubuntu 20.04, the [prerequisites that are installed by the provided command](https://slicer.readthedocs.io/en/latest/developer_guide/build_instructions/linux.html#ubuntu-20-04-focal-fossa) are not sufficient to build Slicer: I got a CMake error message that minimum Qt-5.12.1 is required.

Either the build instructions have to be updated so that it gets the required Qt version, or the Qt version check has to be relaxed.

lassoan:
@RafaelPalomar @pieper what do you think would be the more appropriate solution?

Is it hard to install Qt-5.15.1 on linux? If it is hard then maybe we could add a CMake command to the linux build instructions to lower the required Qt version to 5.12.8 (this is that I have on my ubuntu by defaul) by setting Slicer_REQUIRED_QT_VERSION Cmake variable (as suggested here: https://github.com/Slicer/Slicer/commit/117185ded4709017ec6b2c43d54bc0e1c2d053d9).

pieper:
I would suggest changing the minimum version back to 15.12.8 to match the ubuntu packages.  I just did a build successfully from scratch with that setting.

jamesobutler:
Can Debian 9 no longer build Slicer with the minimum version check event set to Qt 5.12.8?
https://slicer.readthedocs.io/en/latest/developer_guide/build_instructions/linux.html#debian-testing-bullseye-and-debian-9

> Debian 9 Qt 5.7 packages will not work with current Slicer 4.11. Checked 2020-08-19.

pieper:
I suspect that going too far back in the Qt versions will lead to other issues.  It looks like ubuntu isn't offering updated packages due to some Qt license issues so maybe we should just encourage people to build Qt from source.

lassoan:
Ubuntu-20.04.1 works with Qt-5.12.8. I set the minimum version to this, because this would alleviate a lot of frustration and it is very easy to do. We can increment the Qt version number if someone has time to figure out or write down how to get more recent version of Qt on a linux system.

PR #5815, opened 2021-08-24
BUG: Relax Qt version requirements

lassoan:
Official builds and tests are performed with Qt-5.15 on all platforms,
but since it is not easy to get latest Qt versions on Linux, the version
requirement is relaxed on Linux so that it builds on Ubuntu-20.04.

See https://github.com/Slicer/Slicer/issues/5804 for more details.


pieper (approved):
Thanks 👍

PR #5817, opened 2021-08-24
COMP: Update FindVcvars with MSVC 1929 support

jamesobutler:
Based on https://github.com/scikit-build/cmake-FindVcvars/pull/1

Required to successfully build an extension such as SlicerRadiomics with latest version of Visual Studio.

### DIFF
--- a/CMake/FindVcvars.cmake
+++ b/CMake/FindVcvars.cmake
@@ -116,7 +116,7 @@ cmake_minimum_required(VERSION 3.13.4)
 # Global variables used only in this script (unset at the end)
 set(_Vcvars_MSVC_ARCH_REGEX "^(32|64)$")
 set(_Vcvars_MSVC_VERSION_REGEX "^[0-9][0-9][0-9][0-9]$")
-set(_Vcvars_SUPPORTED_MSVC_VERSIONS 1928 1927 1926 1925 1924 1923 1922 1921 1920
+set(_Vcvars_SUPPORTED_MSVC_VERSIONS 1929 1928 1927 1926 1925 1924 1923 1922 1921 1920
                                     1916 1915 1914 1913 1912 1911 1910 1900
                                     1800 1700 1600 1500 1400)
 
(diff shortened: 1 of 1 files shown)


lassoan (approved):

lassoan (commented):
I've just realized that we lost some granularity of the versions here. We don't know anymore the minor version  (in fact, the version is rounded down to the closest integer). How did you check that we did not need this information and rounding down does not cause a problem?

jamesobutler:
> How did you check that we did not need this information and rounding down does not cause a problem?

I just tested it as the current logic of a MSVC_Version matching a Visual Studio version uniquely was no longer true. MSVC 1928 corresponds to both 16.8 AND 16.9 and MSVC 1929 corresponds to both 16.10 and 16.11. As seen in the code below it finds the same path as if you used the previous logic specifying a specific version. I would guess you can use the integer because I don't think it is possible to have install 2 different 16.X versions at the same time.

```cmake
  if(vs_version VERSION_GREATER_EQUAL "15.0")
    # Query the VS Installer tool for locations of VS 2017 and above.
    string(REGEX REPLACE "^([0-9]+)\.[0-9]+$" "\\1" vs_installer_version ${vs_version})
    # vs_installer_version can be "16" here
    cmake_host_system_information(RESULT _vs_dir QUERY VS_${vs_installer_version}_DIR)
    # which successfully returns "C:/Program Files (x86)/Microsoft Visual Studio/2019/Community"
    if(_vs_dir)
      list(APPEND _vs_installer_paths "${_vs_dir}/VC/Auxiliary/Build")
      # which ultimately sets Vcvars as "C:/Program Files (x86)/Microsoft Visual Studio/2019/Community/VC/Auxiliary/Build/vcvars64.bat"
    endif()
  else()
```
image

jamesobutler:
Does the previous explanation make sense of why this change is appropriate?

jamesobutler:
I will proceed with merging if no additional comments. This will allow for successful building of extensions that utilize this file when building with the latest versions of Visual Studio.

Issue #5812, opened 2021-08-23
Using erase tool outside of segmentation crashes slicer


nathanbmnt:
If using the erase tool outside of a specific segmentation Slicer crashes.

## Steps to reproduce
[SegmentationEraseCrashExample.zip](https://github.com/Slicer/Slicer/files/7034356/SegmentationEraseCrashExample.zip)
Load FusedVolume.mha and then Markup_2.seg.nrrd from SegmentationCrashExample.zip into Slicer.
Go to the Segment Editor and click Erase, "Edit in 3D views". The sphere to edit the segmentation appears even when the mouse is not over the segmentation. Maybe the bounds are corrupted? If the mouse is clicked here, outside of the segmentation Slicer crashes.
![image](https://user-images.githubusercontent.com/40271210/130503043-dff72767-2518-4225-a6eb-4acd011485b5.png)
This also happens when erasing the segmentation in a slice view.

The following is printed to the command line for Slicer built with command line visible:

> The update extent specified in the information for output port 0 on algorithm vtkTrivialProducer(000001D3B471D190) is 9 17 80 80 0 0, which is outside the whole extent 9 17 83 80 0 0.
> 
> This data object does not contain the requested extent.
> 
> GetScalarPointer: Pixel (9, 80, 0) not in memory.
> Current extent= (9, 17, 83, 80, 0, 0)
> 
> GetScalarPointer: Pixel (9, 80, 0) not in memory.
> Current extent= (9, 17, 83, 80, 0, 0)

## Expected behavior

After clicking with the erase tool the segmentation is not changed and Slicer does not crash.

## Environment
- Slicer version: Slicer-4.13.0-2021-08-19
- Operating system: Windows 10

jamesobutler:
@nathanbmnt I'm unable to replicate this issue with the specific data that you've provided.

lassoan:
After lots attempts I was able to reproduce it:

https://user-images.githubusercontent.com/307929/130617335-3501c015-7de1-4920-95a3-1eb394799369.mp4

The problem is that `vtkSegmentationModifier::GetExtentIntersection` returns an invalid intersection (resampledSegmentLabelmap = 0,39, 83,193, 0,0; extent = 0,39,  0,13,  0,0; -> updateExtent = 0,39,  83,13,  0,0 !! 83>13) and the vtkImageThreshold crashes if it gets an invalid UpdateExtent instead of just not doing anything.

[...]

PR #5818, opened 2021-08-24
BUG: Fix crash when erasing regions outside segmentation

Sunderlandkyl:
If the modifier labelmap and resampled segmentation labelmaps do not overlap, then GetExtentIntersection will generate an invalid extent. Passing this extent to vtkImageThreshold::UpdateExtent will cause a crash.

Fixed by not applying the segment mask to the modifier if the extent is invalid. This should not cause any issues since if there is no overlap between the modifier/segment labelmaps, nothing should need to be erased.

Fix: #5812

### DIFF
--- a/Libs/vtkSegmentationCore/vtkSegmentationModifier.cxx
+++ b/Libs/vtkSegmentationCore/vtkSegmentationModifier.cxx
@@ -241,6 +241,23 @@ void vtkSegmentationModifier::GetExtentIntersection(const int extentA[6], const 
 }
 
 //-----------------------------------------------------------------------------
+bool vtkSegmentationModifier::IsExtentValid(int extent[6])
+{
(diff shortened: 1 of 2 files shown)


lassoan (commented):
This looks very nice and clear. We should check the extent before all other UpdateExtent() calls. Please also submit a bug report to VTK and include the link for reference (unless it is very hard to describe or reproduce the problem with just using VTK outside Slicer).

lassoan (approved):
Thank you, looks perfect!

Sunderlandkyl:
> Please also submit a bug report to VTK and include the link for reference (unless it is very hard to describe or reproduce the problem with just using VTK outside Slicer).

I'll make a minimum working example and submit the bug report to VTK.

PR #5819, opened 2021-08-24
ENH: Move Segment Editor effects into vertical grid

jamesobutler:
Based on brainstorming in https://discourse.slicer.org/t/segment-editor-effect-button-size-consistency/19214.

The motivation for this change is to better use space in the Segment Editor widget to improve usability. Effect options are now higher up and the buttons take up less overall space by not including text for the buttons.

The number of columns in the gridlayout for the effects can be managed by `setEffectColumnCount`. So if a customization has been made for `setEffectNameOrder` and `unorderedEffectsVisible`, they can also change column count to whatever they please.

| Current | This PR |
|----------|---------|
|image|image|


lassoan:
This looks great! Just minor comments:
- it would be good to have an option for effect list position (bottom = as before; left = as in the current PR, right = as in the current PR but on the right side of the widget)
- I don't think we need the "Effects" label
- It would be nice to expose the EffectButtonStyle (icon+text or only icon) for backward compatibility and also for cases when the developers want bigger buttons (e.g., because there are just a few effects)

pieper:
> * right = as in the current PR but on the right side of the widget

Yes, I was going to suggest putting the buttons on the right of the widget should be the default

cpinter:
Thanks @jamesobutler ! I like the change as well. Hopefully this new arrangement will also improve the layout related issues in the widget (it was so complex that sometimes unexpected things happened in the layout). I also agree with the comments of @lassoan . Removing the option to use the widget the old way might be a problem for some users, so it's good to keep that door open.

jamesobutler:
I have discovered a negative when the segment editor effects are placed on the right side of the widget which is the buttons change position when the module panel changes to needing a vertical scroll bar or not. In the image below I'm clicking in the same spot can toggle between two effects because the buttons are moving. It's this type of behavior that we want to avoid.  Clicking an effect shouldn't change the location of the effects.

image
image

jamesobutler:
I've added expandable widget back and moved segment effects to the left, but still under the add/delete segment row. Other various layout changes. Of course difficult to figure out best size policies, but see how you like the current state.
image

lassoan:
I like this latest version very much! It is almost perfect.

I just have two minor comments:
- The layout can get a bit too wide. For example, Scissors effect layout should be optimized a bit so that it is narrower (and taller, if needed)
- If I scroll down so that I only see the Add/Remove/Show... buttons on top and then switch to "No editing" then the view scrolls up. When I switch to another effect then I have to scroll down to see all the options. It would be nice if somehow the previous scroll position could be restored (or ideally it would not scroll up).

I think neither of these should prevent the merge.

@pieper @cpinter what do you think? Should we merge this so that everyone gets a chance to try it?

jamesobutler:
> The layout can get a bit too wide. For example, Scissors effect layout should be optimized a bit so that it is narrower (and taller, if needed)

When clicking on the scissors effect it can make the left side panel wider if it is currently too small to fit the effect options. Same goes with clicking the threshold effect when the module panel width is small. Any segment editor effect added to Slicer could ultimately do this too unless we enforce some max width that the effect options can be. If greater than that defined width it could show a horizontal scroll bar instead of making the entire module panel area wider. This issue of variable size effect options is why clicking different segment editor effects can make the module panel size wider.

pieper:
It looks like an improvement worth merging so people can try it out and give feedback.

lassoan:
OK, let's do it! I'll update the threshold and scissors effect options layouts to not force the Segment editor widget to be very wide, and I'll merge it.

Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorThresholdEffect.py
```diff
@@ -211,38 +212,51 @@ def setupOptionsFrame(self):
...
+    histogramBrushFrame.addStretch()
+
     self.histogramView = ctk.ctkTransferFunctionView()
+    self.histogramView.setMinimumWidth(30)
+    # qt.QSizePolicy(qt.QSizePolicy.Expanding, qt.QSizePolicy.Expanding)
+    # fails on some systems, therefore set the policies using separate method calls
+    #qSize = qt.QSizePolicy()
+    #qSize.setHorizontalPolicy(qt.QSizePolicy.Expanding)
+    #self.histogramView.setSizePolicy(qSize)
+    #self.histogramView.setMaximumWidth(50)
```
jamesobutler, line 258: Is this some dead code/

lassoan: Good catch, it was just testing code that I left in. I've removed it now.

lassoan:
Merged in https://github.com/Slicer/Slicer/commit/1b94b0d5330b277c698f58d2fe09901875ba2643.

PR #5820, opened 2021-08-24
ENH: Move module panel logo into DataProbe collapsible area

### LINKED ISSUE #3797: Slicer logo steals considerable space from the module UIs
_This issue was created automatically from an original [Mantis Issue](https://mantisarchive.slicer.org/view.php?id=3797). Further discussion may take place here._

jamesobutler:
This closes #3797.

This is motivated to better utilize valuable vertical space in the module panel.

Originally discussed in https://discourse.slicer.org/t/segment-editor-user-interface-improvements/9029
> the 3DSlicer logo on the top left of every module (as it is now) is using up valuable space and it is not necessary. We all know what software we are using.

I am setting the logo label hidden by default rather than removing the label entirely because Slicer custom apps may still want the logo utilized. Is this an appropriate assumption? Or should it be removed and Slicer custom apps reimplement if they feel they need a logo at the top of the module panel?

### DIFF
--- a/Base/QTApp/Resources/UI/qSlicerMainWindow.ui
+++ b/Base/QTApp/Resources/UI/qSlicerMainWindow.ui
@@ -46,13 +46,6 @@
       <number>0</number>
      </property>
      <item>
-      <widget class="QLabel" name="LogoLabel">
-       <property name="alignment">
-        <set>Qt::AlignLeading|Qt::AlignLeft|Qt::AlignTop</set>
-       </property>
-      </widget>
-     </item>
-     <item>
       <widget class="qSlicerModulePanel" name="ModulePanel"/>
      </item>
      <item>
@@ -66,7 +59,15 @@
        <property name="text">
         <string>Data Probe</string>
        </property>
-       <layout class="QHBoxLayout" name="horizontalLayout"/>
+       <layout class="QHBoxLayout" name="horizontalLayout">
+        <item>
+         <widget class="QLabel" name="LogoLabel">
+          <property name="alignment">
+           <set>Qt::AlignLeading|Qt::AlignLeft|Qt::AlignBottom</set>
+          </property>
+         </widget>
+        </item>
+       </layout>
       </widget>
      </item>
     </layout>
--- a/Modules/Scripted/DataProbe/DataProbe.py
+++ b/Modules/Scripted/DataProbe/DataProbe.py
@@ -52,7 +52,7 @@ indicated by the mouse position.
       print("No Data Probe frame - cannot create DataProbe")
       return
     self.infoWidget = DataProbeInfoWidget(parent)
-    parent.layout().insertWidget(0,self.infoWidget.frame)
+    parent.layout().addWidget(self.infoWidget.frame)
 
   def showZoomedSlice(self, value=False):
     self.showZoomedSlice = value
@@ -68,6 +68,7 @@ class DataProbeInfoWidget:
     self.CrosshairNodeObserverTag = None
 
     self.frame = qt.QFrame(parent)
+    self.frame.hide()  # have hidden by default until event triggered to show
     self.frame.setLayout(qt.QVBoxLayout())
 
     modulePath = slicer.modules.dataprobe.path.replace("DataProbe.py","")
@@ -206,11 +207,15 @@ class DataProbeInfoWidget:
       self.viewerColor.hide()
       self.viewInfo.hide()
       self.viewerFrame.hide()
+      self.frame.hide()
+      slicer.util.setApplicationLogoVisible(True)
       return
 
     self.viewerColor.show()
     self.viewInfo.show()
     self.viewerFrame.show()
+    self.frame.show()
+    slicer.util.setApplicationLogoVisible(False)
 
     # populate the widgets
     self.viewerColor.setText( " " )
--- a/Modules/Scripted/DataProbe/DataProbe.py
+++ b/Modules/Scripted/DataProbe/DataProbe.py
@@ -203,17 +203,14 @@ class DataProbeInfoWidget:
         self.layerNames[layer].setText( "" )
         self.layerIJKs[layer].setText( "" )
         self.layerValues[layer].setText( "" )
-      self.imageLabel.hide()
       self.viewerColor.hide()
       self.viewInfo.hide()
       self.viewerFrame.hide()
-      self.showImageFrame.show()
       return
 
     self.viewerColor.show()
     self.viewInfo.show()
     self.viewerFrame.show()
-    self.showImageFrame.hide()
 
     # populate the widgets
     self.viewerColor.setText( " " )
@@ -266,14 +263,6 @@ class DataProbeInfoWidget:
       self.displayableManagerInfo.show()
     else:
       self.displayableManagerInfo.hide()
-
-    # set image
-    if (not slicer.mrmlScene.IsBatchProcessing()) and sliceLogic and hasVolume and self.showImage:
-      pixmap = self._createMagnifiedPixmap(
-        xyz, sliceLogic.GetBlend().GetOutputPort(), self.imageLabel.size, color)
-      if pixmap:
-        self.imageLabel.setPixmap(pixmap)
-        self.onShowImage(self.showImage)
 
     if hasattr(self.frame.parent(), 'text'):
       sceneName = slicer.mrmlScene.GetURL()
@@ -382,31 +371,6 @@ class DataProbeInfoWidget:
     # hide this for now - there's not much to see in the module itself
     self.goToModule.hide()
 
-    # image view: To ensure the height of the checkbox matches the height of the
-    # viewerFrame, it is added to a frame setting the layout and hard-coding the
-    # content margins.
-    # TODO: Revisit the approach and avoid hard-coding content margins
-    self.showImageFrame = qt.QFrame(self.frame)
-    self.frame.layout().addWidget(self.showImageFrame)
-    self.showImageFrame.setLayout(qt.QHBoxLayout())
-    self.showImageFrame.layout().setContentsMargins(0, 3, 0, 3)
-    self.showImageBox = qt.QCheckBox('Show Zoomed Slice', self.showImageFrame)
-    self.showImageFrame.layout().addWidget(self.showImageBox)
-    self.showImageBox.connect("toggled(bool)", self.onShowImage)
-    self.showImageBox.setChecked(False)
-
-    self.imageLabel = qt.QLabel()
-
-    # qt.QSizePolicy(qt.QSizePolicy.Expanding, qt.QSizePolicy.Expanding)
-    # fails on some systems, therefore set the policies using separate method calls
-    qSize = qt.QSizePolicy()
-    qSize.setHorizontalPolicy(qt.QSizePolicy.Expanding)
-    qSize.setVerticalPolicy(qt.QSizePolicy.Expanding)
-    self.imageLabel.setSizePolicy(qSize)
-    #self.imageLabel.setScaledContents(True)
-    self.frame.layout().addWidget(self.imageLabel)
-    self.onShowImage(False)
-
     # top row - things about the viewer itself
     self.viewerFrame = qt.QFrame(self.frame)
     self.viewerFrame.setLayout(qt.QHBoxLayout())
@@ -474,15 +438,6 @@ class DataProbeInfoWidget:
   def onGoToModule(self):
     m = slicer.util.mainWindow()
     m.moduleSelector().selectModule('DataProbe')
-
-  def onShowImage(self, value=False):
-    self.showImage = value
-    if value:
-      self.imageLabel.show()
-    else:
-      self.imageLabel.hide()
-      pixmap = qt.QPixmap()
-      self.imageLabel.setPixmap(pixmap)
 
 #
 # DataProbe widget


lassoan:
I would be happy to see the logo going away. It takes up too much and too valuable screen space. When we rework the DataProbe then we can add the logo display there (when there is nothing to show because the mouse pointer is not in a view).

pieper:
> When we rework the DataProbe then we can add the logo display there (when there is nothing to show because the mouse pointer is not in a view).

Could we please do this before disabling the logo in the top left.  We had this design in Slicer2 and it worked really well.  It should be just a couple line change in the current DataProbe.  I can do it if needed.

I know the lack of space in the module panel is a real issue and that the logo appears useless to many people but having it visible by default (for example in screenshots) has been considered important for branding by key Slicer supporters.

jamesobutler:
Can you provide an image of what it looked like previously with the logo in the DataProbe? With this change the logo would be with the DataProbe, but still visible by default and taking up vertical space in the module panel area?

pieper:
slicer2 looked like this, where the logo was visible by default, but when your mouse pointer was over a slice view then the area where the logo was turned into a zoom of the slice view (like the "Show Zoomed Slice" option in the current DataProbe).  If you want to see it in action I could probably find a VHS tape ;-P   Or maybe it still runs?

image

Currently the DataProbe looks like this when your mouse is not over a slice view, but that blank space could be replaced by the logo that would disappear when there is text to display. (We could also remove the Show Zoomed Slice mode to gain space, since I don't believe it's widely used.  Maybe move the check to the DataProbe module or the Application Settings).



image

jamesobutler:
I’m guessing the Slicer logo made its way into the main application widget area because on macOS there isn’t an icon for the window with the logo? On Windows the logo in the widget area is fairly redundant because it is already present in the top left next to the Window title. Is there a consistent styling for macOS apps about where to display a logo?

jamesobutler:
I've modified my commit so that the module panel logo is moved from the top left of the module panel to the bottom right of the DataProbe collapsible area. Therefore when the DataProbe is collapsed, the module panel logo is not visible anymore.

| Current | This PR |
|----------|----------|
|image|image|

pieper:
Hi all -

I checked with [Ron](https://en.wikipedia.org/wiki/Ron_Kikinis), who many know is the person who really made Slicer happen as the original visionary, active promoter, dedicated user and tester, and long time PI on many grants that have and continue to fund the use and support of Slicer.  He asked me to pass this on:

> The logo is a critical element of the branding of 3D Slicer. Having a recognizable brand helps with recognition in the user and and developer communities and with our fund raising. Visibility of the logo is very important to help prevent dilution of the brand.

> I feel very strongly that maintaining the visibility and prominence of the Slicer logo at the top of the module panel is very important. I am strongly opposed to removing it and making it invisible. While we would gain a little space for engineering purposes, we would loose a lot in conveying that this is the 3D Slicer app.

So let's see what we can do to keep the logo visible.  Perhaps it could be made into a button the brings up the Data module as a popup; that could be very handy for quickly navigating the app, changing visibility, etc without losing context of the current module.

jamesobutler:
> I am strongly opposed to removing it and making it invisible. 

Is he ok with the current iteration in this PR of having it in the DataProbe collapsible area rather than at the top?

> While we would gain a little space for engineering purposes, we would loose a lot in conveying that this is the 3D Slicer app.

I guess as another user has pointed out, when actively using 3D Slicer, you don't need a large logo telling you that you're using 3D Slicer. That's something you already know as a user. It would be a shame to hurt usability of 3D Slicer by prioritizing obtrusive branding. It taking up valuable vertical space in the module panel is obtrusive when the user has to scroll up and down in a module because of it.

Someone viewing a screenshot or a screen capture video is very different than an actual 3D Slicer user. If it needs to be clear that the screenshot is of 3D Slicer, shouldn't that be on the creator of the screenshot/video to make that clear either with labeling/context of the screenshot or with a title screen for the video?

Since branding seems important for fundraising, is the fundraising part primarily completed with screenshots and videos or actual 3D Slicer use cases?

cpinter:
> If it needs to be clear that the screenshot is of 3D Slicer, shouldn't that be on the creator of the screenshot/video to make that clear either with labeling/context of the screenshot or with a title screen for the video?

Lots of scientific papers not written by the active part of the community use Slicer, and these are important part of the dissemination. If you skim such papers you'll find that many times they don't even cite Slicer, and/or they write the name wrong (Slicer3D, Slice3D, etc.), and we can maybe discover them via searching things like "Segment Editor". Or just bumping into them accidentally and seeing the 3D view's background gradient. If these researchers need to make the logo visible explicitly for the publication, they simply won't.

I think by moving the logo to the DataProbe in a way that it's visible most of the time would keep both the space-aware users and the promoters happy.

> Perhaps it could be made into a button the brings up the Data module as a popup; that could be very handy for quickly navigating the app, changing visibility, etc without losing context of the current module.

This is a quite interesting idea, but then we'd need to indicate that it's an active button and it would not be obvious what it is used for. Or maybe it could become the module selector? By clicking the brand logo you find components of the application - this seems intuitive to me...

In any case if the DataProbe approach seems acceptable for everyone and it works nicely then my vote goes there.

jamesobutler:
On Windows, the logo and application name are always present in the top left window title area which is how Windows users know what application is being used. Is this logo and application name not acceptable enough even though it would be present in all screenshots on Windows? If it needs to be bigger and elsewhere in the main application, has Slicer developers defined the design requirements for the logo such as "The 3D Slicer logo must be at least X pixels in size or X% of the Window"? Is there already a 3D Slicer design requirements page on the Wiki/ReadTheDocs that can be used by developers when making various decisions?
image

pieper:
> Or maybe it could become the module selector? By clicking the brand logo you find components of the application - this seems intuitive to me...
> 

I like this idea!  We could replace the search button and module name in the toolbar with a redesigned logo area.  Having the module search button be bigger would be logical, and we could move the 3D Slicer name up and have the module combo box below it, which would make sense for screenshots.

This would also save space in the toolbar, where the module selector takes up a lot of valuable real estate.

pieper:
> defined the design requirements for the logo such as "The 3D Slicer logo must be at least X pixels

Effectively, yes, by making the logo the size it currently is.  I would like to follow Ron's guidance that it not be made any smaller.

> Is there already a 3D Slicer design requirements page on the Wiki/ReadTheDocs that can be used by developers when making various decisions?

Yes, quite a bit of work went into that.  There are several cross-linked wiki pages, probably [this is a good place to start](https://www.slicer.org/wiki/Slicer3:Interface_Design_and_Usability).  As you can see on those pages, usability and space efficiency are high priorities, as is recognizability of the brand.  That's why I'm suggesting we make the logo functional and not just branding.

It's true that the windows title bar introduces some duplication that's not present on mac or linux (although linux varies).  Maybe we can think of creative solutions there too.

pieper:
By the way @rkikinis was trying to comment directly on this thread but said he was locked out for some reason.  Does anyone know why?

jamesobutler:
No moderation is enabled as far as I can see. The only thing I can think of to limit someone from commenting is not being currently signed into GitHub when viewing this page.

lassoan:
Github sometimes has this issue (your comment cannot be saved), I run into it a few times a year. For me the problem always went away in a few ten minutes.

rkikinis:
> This closes #3797.
> 
> This is motivated to better utilize valuable vertical space in the module panel.
> 
> Originally discussed in https://discourse.slicer.org/t/segment-editor-user-interface-improvements/9029
> 
> > the 3DSlicer logo on the top left of every module (as it is now) is using up valuable space and it is not necessary. We all know what software we are using.
> 
> I am setting the logo label hidden by default rather than removing the label entirely because Slicer custom apps may still want the logo utilized. Is this an appropriate assumption? Or should it be removed and Slicer custom apps reimplement if they feel they need a logo at the top of the module panel?

Hi, in my opinion displaying the Logo is "necessary", as I explained in my previous response.

jamesobutler:
Hi @rkikinis, the current state of this PR is that it is still showing the Logo, just in another location. It is positioned in the DataProbe widget as was originally suggested by @pieper based on how it was in Slicer 2 days. Do you find this acceptable? Also do you have thoughts on styling as indicated at https://github.com/Slicer/Slicer/pull/5820#issuecomment-908410769?

rkikinis:
https://www.slicer.org/wiki/Slicer3:Interface_Design_and_Usability#General_Design_Guidelines_that_Reinforce_the_Brand

This page and the dependent pages might help

jamesobutler:
The only thing logo related that I've found in those resources so far is:

> Preserve the footprint of space beside Slicer's logo; do not include other logos in the main application GUI.

Moving the logo into the DataProbe widget area as in this PR doesn't appear to violate this policy. 

There are other logos in the application, such as the python logo is visible in the toolbar area, and logos are present in various acknowledgement text.

rkikinis:
Again, my main point is visibility and proper exposure. Hiding the logo in the bottom of the module does not work for me. For instance, you see the Nike swish on the side of the shoe where it is prominently visible, not on the inside.

jamesobutler:
@rkikinis What do you think of about the following statement from https://github.com/Slicer/Slicer/pull/5820#issuecomment-908410769. Do you know of another application that we can use for example that details the logo somewhere other than the top left window icon area? That would be great to know how Slicer can do something similar.
> On Windows, the logo and application name are always present in the top left window title area which is how Windows users know what application is being used. Is this logo and application name not acceptable enough even though it would be present in all screenshots on Windows?
image

rkikinis:
@jamesobutler ,
this is not the case on the mac. Also, just to reset the conversation: I was directly involved in making decisions about the brand. The actual work was done by Wendy Plesniak who is both a designer and an artist. I felt from the inception that branding was very important for slicer. My feelings on this point have not changed.

User oriented industry has mostly their logo in the upper left with some spacing. I could not find a single example where the logo is hidden at the bottom. The general functionality associated with a logo is to take you to the home page or start page.

https://www.nike.com/?cp=90083564018_search_%7CNike%7CGOOGLE%7C71700000038284603%7CGN_X_X_X_X-Device_X_Legacy_Core-NikeBrand_Exact%7Ce%7Cc&gclsrc=aw.ds&gclid=Cj0KCQjwg7KJBhDyARIsAHrAXaFXJ_McXc40lmToxangCM8jX7sxFFuYzzkiAVKygLT-kJBa09JHrjQaAkkWEALw_wcB&gclsrc=aw.ds

https://www.apple.com/

https://www.starbucks.com/

https://www.google.com/

https://www.microsoft.com/en-us/

jamesobutler:
> this is not the case on the mac.

Based on Slicer download statistics for the latest Slicer 4.11 stable releases, Windows downloads have accounted for nearly 75% of all downloads (see below image). That is a well established majority. Does there need to be a logo elsewhere in the program to satisfy the other 25%? The current logo position is not that many pixels away from the logo and application name in the expected top left of the application.
image

I appreciate the examples you have provided for uses of a logo for branding, but unfortunately these are examples of webpages.  I was looking for examples of how Desktop applications display the logo for users. For instance are there examples of macOS applications that display the application logo somewhere in the main application widget area? On Windows, as it relates to branding on desktop applications I'm not seeing this as a standard that users are familiar with. The standard appears to be icon and application name in the top left and then colors used in the Window primarily tying to the brand.

lassoan:
I checked 30 applications on my computer (open-source professional applications like Slicer, but also commercial applications and system applications) and they all use about 2-3% of the vertical space (about a single line of text) of the client area for displaying its name and logo. (The client area is the application window except menubar, toolbar, statusbar.) See the images [here](https://1drv.ms/u/s!Arm_AFxB9yqHxoodFDXMW2B8x_Kyvg?e=Gc78jL). Even on websites, the application name and logo may be displayed with a large size in the opening page, but when you start scrolling or click on content links then the logo either goes away or remains in a thin single-line header.

In contrast, Slicer uses about 15% of the vertical space, which is 5x larger than the standard. To make things worse, this space is in the most valuable area of the application window, where all modules display their GUI, making scrolling necessary much more frequently. For each click on a button that is not visible because of the reduced space, the user needs 5-6 extra clicks and two mouse moves; or a few mouse wheel gestures.

This is not something that people can just get used to, but it is a constant source of frustration, even for me, who is a very heavy user, this is annoying and whenever I need to do something complex then I run a python code snippet to hide the logo from the GUI. Maybe the only people who are not bothered by this are those who use a large, high-resolution screen, but they are less than 10% of our users (based on slicer.org website analytics only about 11% of users have HiDPI screens - larger than full HD resolution - and many of those can be laptops with high-resolution screens).

We need to go back to the original requirements that led to displaying the application name&logo like this and come up with a solution that fulfills those requirements in a way that it does not violate industry standards or hurts the usability of the application.

rkikinis:
This might be a working solution:

“Even on websites, the application name and logo may be displayed with a large size in the opening page, but when you start scrolling or click on content links then the logo remains in a thin single-line header.“

Sent from my mobile phone. 
This email is for non-work related messages

jamesobutler:
Thanks @lassoan for putting those images together of various desktop applications. That's what I was doing on my own, but providing images to everyone is helpful. I agree with the philosophy of trying to use industry standards of desktop application design as familiar design will be easier for users to navigate. I'm sure this will change over the decades as operating systems change their design as well.

I do not see how a dynamic scrolling to move logo into a line entry could be possible using the GUI frameworks that we are working with. I think the branding in the application needs to be discussed specifically related to Desktop applications and not web site design. A Desktop application has different expected navigation techniques compared to a web site and also has its own expected usage of logo and title. I would only recommend applying web site design practices to 3D Slicer, if the development was primarily focused on giving 3D Slicer some unified experience if it ran as a desktop app and also ran within a browser, however I do not believe that is a current top priority.

rkikinis:
@jamesobutler You lost me there. I felt that the suggestion raised by Andras addressed both sets of requirements: large initial logo and small when in use. I don’t understand your response. 

Sent from my mobile phone. 
This email is for non-work related messages

jamesobutler:
Large initial logo and small after scrolling is something that I don’t believe is possible when using Qt (the toolkit that creates the GUI). Therefore not a viable solution for a desktop application.

rkikinis:
Not possible is difficult to address head on.  How about we park the large logo in the welcome module and have the small one in the top left with some distance to the remaining items on that row? I am open for other workarounds if someone has a better idea

Sent from my mobile phone. 
This email is for non-work related messages

lassoan:
We could add the logo to the title bar of the module panel dockable widget title bar (see [here](https://stackoverflow.com/questions/2386527/how-to-display-icon-in-qdockwidget-title-bar)). At application startup a large logo would be displayed and we could switch to the small version when switching to a different module.

Photoshopped mockup:

image

This could be combined with the idea of making the icon functional (bring up the module finder) and displaying the module selector in the title. It could free up some space in the toolbar and it would make module selection cleaner (the selector is in the same widget as the module GUI), but it might make the module title bar a bit too crowded.

image

rkikinis:
How about moving the three folder icons instead and leaving a little space between the logo and the rest?

Sent from my mobile phone. 
This email is for non-work related messages

jamesobutler:
I've attempted putting the logo into the QMenuBar area like Visual Studio where the the title/resize bar is merged with the menu bar. Unfortunately Qt (5 at least) does not do a good job of supporting something like this. I have put the logo label widget in the menu bar area by `setCornerWidget` and set window flags with the FramelessWindowHint. From a google others have tried things like this, but often rely on hacky methods that are apparently not very stable as OS versions change.
image
image
Original:
image

lassoan:
Thanks for testing this! Could you try putting the logo (and maybe even the module selector widgets) into the module panel title bar?

jamesobutler:
Here's another experiment with the logo as an unmovable entry in the toolbar.

Rough prototyping code
```python
logo_label = slicer.util.mainWindow().findChild(qt.QLabel, "LogoLabel")
logo_label.setSizePolicy(qt.QSizePolicy.Minimum, qt.QSizePolicy.Minimum)
pixmap=logo_label.pixmap.scaled(qt.QSize(80, 60), qt.Qt.KeepAspectRatio, qt.Qt.SmoothTransformation)
logo_label.setPixmap(pixmap)
logo_toolbar = qt.QToolBar("Logo")
logo_toolbar.addWidget(logo_label)
logo_toolbar.setParent(slicer.util.mainWindow())
logo_toolbar.setMovable(False)
main_tool_bar = slicer.util.mainWindow().findChild(qt.QToolBar, "MainToolBar")
slicer.util.mainWindow().insertToolBar(main_tool_bar, logo_toolbar)
```

image

jamesobutler:
This is moving the logo widget from inside the dock widget to the dock widget title bar area. Not really saving vertical space as just moving to another area. I don't see this as a true viable solution of trying to minimize vertical space. Adding additional functionality to it is possible, but still not solving the issue of maximizing the efficiency of the vertical space.
```python
logo_label = slicer.util.mainWindow().findChild(qt.QLabel, "LogoLabel")
logo_label.setSizePolicy(qt.QSizePolicy.Minimum, qt.QSizePolicy.Minimum)
pixmap=logo_label.pixmap.scaled(qt.QSize(80, 60), qt.Qt.KeepAspectRatio, qt.Qt.SmoothTransformation)
logo_label.setPixmap(pixmap)
dock_widget=slicer.util.mainWindow().findChild(qt.QDockWidget, "PanelDockWidget")
dock_widget.setTitleBarWidget(logo_label)
logo_label.setAlignment(qt.Qt.AlignCenter)
```
image

lassoan:
These both look very promising.

Adding the logo to the toolbar is quite nice, as it only makes the toolbar about half line taller, and the logo has a reasonable size. However, it takes up some space from the toolbar.

Adding the logo to the module panel title bar is also good, because it just adds about 1 extra line (big improvement over the current logo, which adds 5) and we could use that space for the module selector (which is much more reasonable to have in the module panel title than in the toolbar).

rkikinis:
@jamesobutler @lassoan Thanks for working on this. It inspired me to do some mock up myself. In my mind module selector and module shortcut belong together. That is why I would move the folders with the logo. See attached.
Screen Shot 2021-08-31 at 8 14 56 AM

cpinter:
> This could be combined with the idea of making the icon functional (bring up the module finder) and displaying the module selector in the title. It could free up some space in the toolbar and it would make module selection cleaner (the selector is in the same widget as the module GUI), but it might make the module title bar a bit too crowded.
> 
> image

I really like this! I completely agree with @lassoan 's statement "we could use that space for the module selector (which is much more reasonable to have in the module panel title than in the toolbar)".

> @jamesobutler @lassoan Thanks for working on this. It inspired me to do some mock up myself. In my mind module selector and module shortcut belong together. That is why I would move the folders with the logo. See attached.
> Screen Shot 2021-08-31 at 8 14 56 AM

Yes, and this one is along these lines as well, applying the idea about moving the folder icons in the line of the logo. The drawback in my view is I don't see the logic behind looking for the load/save functions right there. However, it "keeps" the module selection controls just above the (defult location of the) module panel, while using the space next to the logo for something different but not crowding that space too much. I like this a lot too.

Thanks all for the great conversation!

jamesobutler:
This was discussed during today's weekly hangout with @pieper and @lassoan present. I've pushed an additional commit to move the logo in the DataProbe collapsible area to be left justified and to show/hide based on the DataProbe being used. When manipulating objects in the module panel area, the regular data probe information is empty and so the Application logo is shown instead. When the user moves their mouse over a slice view, the data probe becomes active with the corresponding widget information shown and the logo hidden during this time.  Therefore, this application logo will be hidden in a screenshot if the mouse is currently over the slice views.

This change is primarily motivated so that when an effect such as "Paint" is selected in the Segment Editor module, you can see more of the effect options and masking options at once rather than having to scroll to see the various elements. #5819 does further optimization of this module to improve space efficiency, but both help improve usability.

https://user-images.githubusercontent.com/15837524/131572663-fd096ca3-4b73-4944-8483-3229c9c2d2a6.mp4

pieper:
To me the version in your video looks just like what we discussed and I find that a logical solution.  Thanks for putting that together @jamesobutler.

I did talk with @rkikinis and he was suggesting something more or less like the layout mocked up below.  That is with a little more space between the logo and the modules and the double row of module controls so the logo is a bit bigger.  Let's see what he thinks after seeing the video of the Data Probe option.

image

rkikinis:
Thanks to everyone for working on this. How about we use @jamesobutler movie for the large logo and @pieper mock up  for the permanent? I believe that it addresses the majority of the requirements that were raised.

jamesobutler:
I've actually noticed that the "Show zoomed slice" checkbox in the DataProbe area is not never seen in my latest iteration because I hide the entire DataProbeInfoWidget frame and only show the logo. I need to resolve this in some manner. Any suggestions?
image

@rkikinis Were you suggesting the logo in the DataProbe widget hiding and showing and then also include the logo at the top of the module area based on the other mockup where 2 logos are shown when mousing over the module panel area? Or were you suggesting moving the logo to the DataProbe and then reverting that change at some point for the more permanent of putting it at the top of the module panel?

lassoan:
We will have logo+application name in two places (the window title and in the Data probe). Are we considering adding a third location (in the header of the module panel)?

@jamesobutler The height of the DataProbe must not change when moving the mouse over/away from the viewers. It is always a distracting flickering and when the module panel layout is complex then it may even be more distracting, slower, may make scroll bar appear/disappear, etc.

I think the "Show zoomed slice" feature can be removed. It is useless, as it does not show any of the overlays (markups, segmentations, model intersections, etc.), it does not slow down the mouse cursor (so it is very hard to position the mouse accurately), and when you drag-and-drop markup points then the crosshair position is not the markup control position. Simply zooming in the image, or using the Magnifier feature built-into the operating system both work much better.

rkikinis:
> On Aug 31, 2021, at 10:19 PM, Andras Lasso ***@***.***> wrote:
> 
> 
> We will have logo+application name in two places (the window title and in the Data probe). Are we considering adding a third location (in the header of the module panel)?
> 
The window title is different on the mac, so only 2 (we advertise as a multiplatform software).

> @jamesobutler The height of the DataProbe must not change when moving the mouse over/away from the viewers. It is always a distracting flickering and when the module panel layout is complex then it may even be more distracting, slower, may make scroll bar appear/disappear, etc.
> 
> I think the "Show zoomed slice" feature can be removed. It is useless, as it does not show any of the overlays (markups, segmentations, model intersections, etc.), it does not slow down the mouse cursor (so it is very hard to position the mouse accurately), and when you drag-and-drop markup points then the crosshair position is not the markup control position. Simply zooming in the image, or using the Magnifier feature built-into the operating system both work much better.
> 
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub, or unsubscribe.
> Triage notifications on the go with GitHub Mobile for iOS or Android. 
> 



—
Ron Kikinis
***@***.***
--------
This email is intended for non-work related messages

rkikinis:
@rkikinis Were you suggesting the logo in the DataProbe widget hiding and showing and then also include the logo at the top of the module area based on the other mockup where 2 logos are shown when mousing over the module panel area? 

Yes

> On Aug 31, 2021, at 10:03 PM, James Butler ***@***.***> wrote:
> 
> 
> I've actually noticed that the "Show zoomed slice" checkbox in the DataProbe area is not never seen in my latest iteration because I hide the entire DataProbeInfoWidget frame and only show the logo. I need to resolve this in some manner. Any suggestions?
> 
> 
> @rkikinis Were you suggesting the logo in the DataProbe widget hiding and showing and then also include the logo at the top of the module area based on the other mockup where 2 logos are shown when mousing over the module panel area? Or were you suggesting moving the logo to the DataProbe and then reverting that change at some point for the more permanent of putting it at the top of the module panel?
> 
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub, or unsubscribe.
> Triage notifications on the go with GitHub Mobile for iOS or Android. 
> 



—
Ron Kikinis
***@***.***
--------
This email is intended for non-work related messages

jamesobutler:
Do we need the logo to be in the DataProbe area if the suggestion is that we must keep the logo at the top of the Module panel?

lassoan:
I feel that the showing the logo at 3 locations would be way too much. @jamesobutler can you show actual screenshots of how would that look like?

If I had to choose between showing the logo in the module selector header or the Data Probe, I would vote for the module selector header, as it would allow moving the module selector there. This may be a better choice also because the user can hide the Data Probe, it would solve the issue of flickering Data Probe (due to changing size as moving over/away from views), and in the future we may want to move the Data Probe to the status bar.

rkikinis:
Ok

Sent from my mobile phone. 
This email is for non-work related messages

lassoan:
Great, it seems that we have converged to a solution that is acceptable for everyone.

@jamesobutler could you implement the custom module panel header widget containing double-line-high Slicer logo on the left, module selector combobox, prev/next module, module history, and module finder on the right (as shown [here](https://github.com/Slicer/Slicer/pull/5820#issuecomment-909737930) + add a search icon)? We would not change the Data Probe.

rkikinis:
Thanks to all participants for their contribution 

Sent from my mobile phone. 
This email is for non-work related messages

jamesobutler:
It's currently unclear to me how we will close #3797 specifically. This is because the logo in the module panel area appears to have a vertical dimension of about 80 pixels while a single toolbar row is about 40 pixels. Therefore if we put a 2-row set of module selector widgets in the module panel area, that leaves no room to make the logo smaller to address #3797. If that issue is not going to be addressed should it then be closed?

| Current | Logo with Module Selector Mockup |
|----------|----------------------------------------|
|image|image|

If the Logo is removed entirely or instead moved to the DataProbe widget area, then the module panel gains that 80 pixels back. Currently, the logo is approximately ~8% (80/1030) of the vertical space. This matches what was specified in the original mantis issue for #3797 back originally by @cpinter.

lassoan:
This is hard. Indeed, if we add two lines in the header then we only reduce the header size by about 25-30% (red line). If we only have a single line for module selection then we can reduce the header to about half size (orange line), but then there is no space for prev/next/history buttons.

image

cpinter:
> 8%

I think by making that 8% useful that's a great improvement. Nowadays with so many module toolbars enabled (sequences, VR...), the toolbar space has become valuable as well.
Also we could use ctkFlowLayout in the module selector part. If the module panel is wide, it only takes one line. If someone makes it narrow, they lose a bit of vertical space by braking the module selection row (prev/next buttons jump to second row).

> Acknowledgement

It is an interesting idea. Indeed we don't really use that section day by day. Not sure if the user would recognize that the acknowledgement section that is exiled to be next to the Slicer logo is module specific.

> Module panel header

Yes that is another UI element that takes space. But if we want to keep the widget dockable, then we need that, otherwise how do we detach it? Not sure how many users use the docking feature (I barely if ever). But for example I know of people who put the Python interactor tabbed with the module panel in the dock widget.

@lassoan 's screenshot shows nicely that the new Slicer logo takes up less space vertically than the old one (due to those planes), so a one-line module selector could be useful there. I think those controls with a flow layout would be a good tradeoff.

jamesobutler:
My screenshots were showing the widgets after setting them with `setTitleBarAreaWidget` and not just in the main widget area of the dock widget. The current Slicer logo contains the thin plane lines around the sphere and those do get close to the top title bound. With your screenshots with the new logo you’re able to crop more around the sphere as the sphere can stay the same relative size but the logo vertical size reduces.

lassoan:
> Yes that is another UI element that takes space. But if we want to keep the widget dockable, then we need that

I've checked this. Setting the logo as title widget keeps the widget dockable by drag-and-drop as usual. So we can use the module panel title area without losing any functionality.

PR #5822, opened 2021-08-26
BUG: Add missing Python3.dll to the install package

### LINKED ISSUE #5816: Python3.dll stub missing
Some python libs use python3.dll stub instead of python36.dll. The stub is not being packaged to the final Slicer installer

## Steps to reproduce

Slicer python console:

```
pip_install("paramiko")
import paramiko
```

Error:

```
...\Slicer 4.13.0-2021-08-19\lib\Python\Lib\site-packages\bcrypt\__init__.py", line 25, in <module>
    from . import _bcrypt  # type: ignore
ImportError: DLL load failed: The specified module could not be found.
```

When building Slicer, I can see that python3.dll is present at the folder `python-install/bin` but it is not copied to the final Slicer-build output. If I copy the python3.dll from `python-install` to the bin folder, the issue is fixed.

## Expected behavior

python3.dll stub should be present inside Slicer bin folder.

## Environment
- Slicer version: Slicer-4.13.0-2021-08-19

Thanks, Fernando.

ps. I've tried to look into the cmake stack but I could not make sense of it.

lassoan:
fixes #5816

### DIFF
--- a/CMake/SlicerBlockInstallPython.cmake
+++ b/CMake/SlicerBlockInstallPython.cmake
@@ -78,12 +78,12 @@ To create a Slicer package including python libraries, you can *NOT* provide you
     install(FILES "${PYTHON_LIBRARY_PATH}/${PYTHON_LIB_BASE}.dll"
       DESTINATION bin
       COMPONENT Runtime)
-    # Copy Python3.dll
+    # Copy python3.dll
     # The file name is always exactly this, as this file is guaranteed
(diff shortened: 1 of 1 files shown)


lassoan:
@jcfr Please check if this is an appropriate way to add python3.dll to the package on Windows. Thank you!

lassoan:

jcfr (approved):
Consider squashing the commits together when integrating

PR #5824, opened 2021-08-26
BUG: Fix Erase effect "Erase all segments" options

### LINKED ISSUE #5584: Segment Editor's Erase effect "Erase all segments" option only erase current segment

Sunderlandkyl:
When "Erase all segments" was enabled, only the active segment was erased because minimumOfAllSegments was not enabled. In addition, labelmaps on other layers were not modified. Fixed by setting minimumOfAllSegments to True and marking all segment IDs to be overwritable when erasing all segments.

Fix #5584

### DIFF
--- a/Modules/Loadable/Segmentations/EditorEffects/qSlicerSegmentEditorAbstractEffect.cxx
+++ b/Modules/Loadable/Segmentations/EditorEffects/qSlicerSegmentEditorAbstractEffect.cxx
@@ -457,6 +457,12 @@ void qSlicerSegmentEditorAbstractEffect::modifySegmentByLabelmap(vtkMRMLSegmenta
     segmentIDsToOverwrite.clear();
     }
 
+  if (modificationMode == qSlicerSegmentEditorAbstractEffect::ModificationModeRemoveAll)
+    {
+    // If we want to erase all segments, then mark all segments as overwritable
+    segmentIDsToOverwrite = allSegmentIDs;
+    }
+
   // Create inverted binary labelmap
   vtkSmartPointer<vtkImageThreshold> inverter = vtkSmartPointer<vtkImageThreshold>::New();
   inverter->SetInputData(modifierLabelmap);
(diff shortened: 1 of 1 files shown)


lassoan (commented):
Thank you for working on this. It works well except it seems to make segment masking settings ignored.

If there are two overlapping segments and masking setting is Editable area=Segment_1 then erase all should only erase parts of Segment_2 where it overlaps with Segment_1.

lassoan:
Example video:

https://user-images.githubusercontent.com/307929/131196791-5be7a145-b79c-4629-8ad3-bdc944074c3e.mp4

Since Editable area is Segment_1 only, Segment_2 should only be erased where it overlaps with Segment_1.

Sunderlandkyl:
I don't think this is a new issue. It seems to have been the behavior for erase all at least since 4.10.2.

lassoan:
Thanks for checking. It is better that it is not a regression, it means we can merge this PR as is.

However, the current behavior does not fully respect the masking settings, therefore at some point we should fix it. Since your memory about this is still fresh, it could be a good opportunity to take care of it now. But it is up to you, you can either work on it now or add an issue to track it.

Sunderlandkyl:
Ok, I've created created an issue to track the bug here: https://github.com/Slicer/Slicer/issues/5831.

lassoan (approved):
Thank you, this is ready to be merged then.

PR #5832, opened 2021-08-31
ENH: Left/Right dock widgets given ownership to bottom corner areas

jamesobutler:
Originally discussed at today's hangout with @pieper @lassoan 
Note that this is somewhat related to #5820 which is also improving the defaults to value module panel vertical space more. That PR achieves more space given back to the module panel by moving the application logo to a new spot. That PR has quite the conversation going on as it relates to the logo so I wasn't going to add the changes in this PR there to avoid multiple conversations going on at once especially if this proposed change is as controversial.

Since the module panel is typically not a majority of the width dimension, this means the python interactor in the bottom widget area still has a wide aspect ratio. From my time using the python interactor I find that of typical screen size (1920x1080 at 100 to 125% scale) and aspect ratio (16:9), that the full width of the widget hasn't been used. Therefore, making it slightly less wide where it only docks below the central widget still seems appropriate.

This change maintains the current Slicer default of having the python interactor dock widget shown.

The screenshot here is also showing changes from #5820 FYI.

|Current where bottom dock widget in control of corners| This PR Docked on the left | This PR Docked on the right |
|--------------------------------------------------------------|----------------------|------------------------|
|image|image|image|

### DIFF
--- a/Base/QTApp/qSlicerMainWindow.cxx
+++ b/Base/QTApp/qSlicerMainWindow.cxx
@@ -496,6 +496,17 @@ void qSlicerMainWindowPrivate::setupUi(QMainWindow * mainWindow)
     qWarning("qSlicerMainWindowPrivate::setupUi: Failed to create Python console");
     }
 #endif
+
+  //----------------------------------------------------------------------------
+  // Dockable Widget Area Definitions
+  //----------------------------------------------------------------------------
+  // Setting the left and right dock widget area to occupy the bottom corners
+  // means the module panel is able to have more vertical space since it is the
+  // usual left/right dockable widget. Since the module panel is typically not a
+  // majority of the width dimension, this means the python interactor in the
+  // bottom widget area still has a wide aspect ratio.
+  q->setCorner(Qt::BottomLeftCorner, Qt::LeftDockWidgetArea);
+  q->setCorner(Qt::BottomRightCorner, Qt::RightDockWidgetArea);
 }
 
 //-----------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


pieper:
Looks great - extra space for free!  Glad it was a simple fix.  I have no objections to the implementation.

lassoan (commented):
I've tested this and it works well. If someone uses two Slicer instances side-by-side or uses monitor in portrait orientation then the Python interactor becomes hardly usable at the bottom, but then it can be docked to the top to use the full width of the application window.

Please just describe the rationale for this setting in the source code and not just in commit comments. If the reason for the corner adjustment is very clear and easily accessible for developers then that reduces the chance that the code gets removed in the future during refactoring.

jamesobutler:
@lassoan I have added comments to the code describing the rationale.

jamesobutler:
I will note that since QDockWidget positioning is saved as part of the window state, this change won't populate to users if they already have a Slicer.ini. They would have to restore defaults in settings or manually do the following code to trigger the newly defined state. If we force it after the window state has been restored that could work, but then would no longer truly be fully restoring the window to last state.
```python
slicer.util.mainWindow().setCorner(qt.Qt.BottomLeftCorner, qt.Qt.LeftDockWidgetArea)
slicer.util.mainWindow().setCorner(qt.Qt.BottomRightCorner, qt.Qt.RightDockWidgetArea)
```

Base/QTApp/qSlicerMainWindow.cxx
```diff
@@ -496,6 +496,17 @@ void qSlicerMainWindowPrivate::setupUi(QMainWindow * mainWindow)
...
 #endif
+
+  //----------------------------------------------------------------------------
+  // Dockable Widget Area Definitions
+  //----------------------------------------------------------------------------
+  // Setting the left and right dock widget area to occupy the bottom corners
+  // means the module panel is able to have more vertical space since it is the
+  // usual left/right dockable widget. Since the module panel is typically not a
+  // majority of the width dimension, this means the python interactor in the
+  // bottom widget area still has a wide aspect ratio.
```
lassoan, line 507:
```suggestion
  // bottom widget area still has a wide aspect ratio.
  // If application window is narrow then the Python interactor can be docked to the top
  // to use the full width of the application window.
```

Issue #5835, opened 2021-09-02
Update Acknowledgments link to latest Acknowledgments documentation page


jamesobutler:
![image](https://user-images.githubusercontent.com/15837524/131856640-8c1655ea-e03a-4e9c-8a6d-80363f7f1ffb.png)

https://github.com/Slicer/Slicer/blob/1d0816b9c254e1fbf61901d6ec6a238b8e1b5c19/Base/QTCore/qSlicerCoreApplication.cxx#L1631

Old Acknowledgments link: https://www.slicer.org/wiki/Documentation/4.x/Acknowledgments
New Acknowledgments link: https://slicer.readthedocs.io/en/latest/user_guide/about.html#acknowledgments 

There appears to be duplicate information on the old link compared to the new link though the wiki page doesn't indicate that the information has been moved elsewhere. The only unique information on the wiki page is the table regarding all the grants ever received https://www.slicer.org/wiki/Documentation/4.x/Acknowledgments#Grants. Is this grants table up-to-date? Where should it live going forward? On the website?

The new Acknowledgments link matches what the Slicer website links to https://slicer.readthedocs.io/en/latest/user_guide/about.html#acknowledgments
![image](https://user-images.githubusercontent.com/15837524/131857548-c4bb5568-cd6f-4f13-9f9f-fcb2cc2e4920.png)

lassoan:
The grant list table looks up-to-date but it is certainly not complete. I think it would make sense to add it to readthedocs (version it along with the software, the same way as it was versioned on the wiki, too) instead of adding it to slicer.org (one text for all Slicer versions).

I've created a [pull request](https://github.com/Slicer/Slicer/pull/5836) to show how it would look like: https://slicer--5836.org.readthedocs.build/en/5836/user_guide/about.html#funding-sources

By moving the table to readthedocs we lose the sortable headers, but if someone needs to do analytics on the table then it can be copy-pasted into Excel (I've tested this and it works fine). If sortable headers are crucial then we could set it up with some Javascript package on slicer.org, but it would be more effort to implement and maintain it.

@rkikinis @pieper
**What do you think about moving funding source list from the [wiki](https://www.slicer.org/wiki/Documentation/4.x/Acknowledgments#Grants) to the [new Slicer documentation](https://slicer--5836.org.readthedocs.build/en/5836/user_guide/about.html#funding-sources) so that it can be better maintained and integrated with the rest of the documentation?**

pieper:
[...]

PR #5836, opened 2021-09-02
DOC: Adding funding grant list

lassoan:
Move list of funding grants from the old wiki (https://www.slicer.org/wiki/Documentation/4.x/Acknowledgments#Grants)

This pull request is to test how the documentation page would look like if we included the funding grant list. Motivated by question raised in https://github.com/Slicer/Slicer/issues/5835.


jamesobutler (commented):
At first I thought some text in the table was getting cutoff, but realized you have to scroll down the long list to be able to see the horizontal scroll bar. The Wiki table word wraps long text in cells.
image
image

jcfr:
> but realized you have to scroll down the long list to be able to see the horizontal scroll bar.

I made a similar observation and initially thought the dates and PIs were dropped from the new listing.

Alternatively, we could list the grants and have them listed on the landing page. This would give us more flexibility to present the list.

cpinter:
Once the format is finalized I can add the Spanish grant we have received partially for infrastructure - if you think it's appropriate to add that is.

lassoan:
@cpinter this PR is merged, you can now add the Spanish grant.

Issue #5823, opened 2021-08-26
Slicer build failed with error C2666: '==': 5 overloads have similar conversions


upupming:
[Give a short summary of what the problem is]

Slicer build is errored after following the `windows.md` doc.

## Steps to reproduce

[describe what you did and what happened - attach screenshots if possible]

Build with command:

```
"C:\Program Files\CMake\bin\cmake.EXE" --build c:/D/S4/build --config Debug --target ALL_BUILD -j 14 --
```

I will get this error:

```bash
[build]     qMRMLSegmentSelectorWidget.cxx
[build]     qMRMLSliceControllerWidget.cxx
[build]     qMRMLSliceInformationWidget.cxx
[build] C:\D\S4\Libs\MRML\Widgets\qMRMLSceneModel.cxx(1298,3): error C2666: '==': 5 overloads have similar conversions [C:\D\S4\build\Slicer-build\Libs\MRML\Widgets\qMRMLWidgets.vcxproj] [C:\D\S4\build\Slicer.vcxproj]
[build]   C:\D\S4\build\VTK\Common\Core\vtkUnicodeString.h(276,27): message : could be 'bool operator ==(const vtkUnicodeString &,const vtkUnicodeString &)' (compiling source file C:\D\S4\Libs\MRML\Widgets\qMRMLSceneModel.cxx) [C:\D\S4\build\Slicer-build\Libs\MRML\Widgets\qMRMLWidgets.vcxproj]
[build]   C:\D\S4\build\VTK\Common\Core\vtkSmartPointerBase.h(118,1): message : or       'bool operator ==(const vtkSmartPointerBase &,vtkObjectBase *)' (compiling source file C:\D\S4\Libs\MRML\Widgets\qMRMLSceneModel.cxx) [C:\D\S4\build\Slicer-build\Libs\MRML\Widgets\qMRMLWidgets.vcxproj]
[build]   C:\D\S4\build\VTK\Common\Core\vtkSmartPointerBase.h(118,1): message : or       'bool operator ==(vtkObjectBase *,const vtkSmartPointerBase &)' (compiling source file C:\D\S4\Libs\MRML\Widgets\qMRMLSceneModel.cxx) [C:\D\S4\build\Slicer-build\Libs\MRML\Widgets\qMRMLWidgets.vcxproj]
[...]

PR #5837, opened 2021-09-06
COMP: fix vtkWeakPointer in Q_ASSERT

pieper:
Weak pointer doesn't automatically resolve to pointer in Q_ASSERT.

See discussion here:
https://github.com/Slicer/Slicer/issues/5823

### DIFF
--- a/Libs/MRML/Widgets/qMRMLSceneModel.cxx
+++ b/Libs/MRML/Widgets/qMRMLSceneModel.cxx
@@ -1295,7 +1295,7 @@ void qMRMLSceneModel::onMRMLSceneDeleted(vtkObject* scene)
   Q_UNUSED(scene);
 #ifndef QT_NO_DEBUG
   Q_D(qMRMLSceneModel);
-  Q_ASSERT(scene == d->MRMLScene);
+  Q_ASSERT(scene == d->MRMLScene.GetPointer());
 #endif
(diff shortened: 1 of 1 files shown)


upupming (approved):

lassoan (commented):
This change it harmless and if it helps in some cases then it makes sense. However, we need to add a comment there, otherwise next time a developer changes something around there it may be reverted to the simpler `(scene == d->MRMLScene)` comparison.

@pieper Were you able to reproduce the issue? Does it happen only within a Q_ASSERT?

Libs/MRML/Widgets/qMRMLSceneModel.cxx
```diff
@@ -1295,7 +1295,7 @@ void qMRMLSceneModel::onMRMLSceneDeleted(vtkObject* scene)
   Q_UNUSED(scene);
 #ifndef QT_NO_DEBUG
   Q_D(qMRMLSceneModel);
-  Q_ASSERT(scene == d->MRMLScene);
+  Q_ASSERT(scene == d->MRMLScene.GetPointer());
```
lassoan, line 1298:
```suggestion
  Q_ASSERT(scene == d->MRMLScene.GetPointer());  // GetPointer() is needed because on certain compilers (scene == d->MRMLScene) comparison is ambiguous
```

Issue #5831, opened 2021-08-28
Segment Editor's Erase effect "Erase all segments" option does not respect masking settings


Sunderlandkyl:
Using the "Erase all segments" feature of the erase effect does not respect current masking settings.
When the editable area is restricted to the inside of a single segment, we can erase both the area within that segment, as well as the area inside the currently selected segment.

## Steps to reproduce

See this comment: https://github.com/Slicer/Slicer/pull/5824#issuecomment-907518593

## Expected behavior

We should only be able to erase segments inside of the editable area.

Likely this is due to the fact that the current segment is always included in the mask so that we can "add back" segments in some circumstances.

## Environment
- Slicer version: Slicer-4.13.0-2021-08-26
- Operating system: All

PR #5839, opened 2021-09-07
BUG: Ensure that masking settings are fully observed for most effects

Sunderlandkyl:
The current region previously always added the current segment to the editable area if we were using one of the "EditAllowedInside*Segment*" masking modes.

This commit changes the behavior to only remove the current segment from the mask when:
  1) The masking setting is set to inside a single segment
  2) Modification mode is ModificationModeRemove

In this circumstance, we add the "Mask" segment to the current segment.

Re #5831


lassoan (approved):
This works very well, thank you!

PR #5842, opened 2021-09-09
DOC: Add ICEX grant to funding list

cpinter:


lassoan:
Thank you
Is it the collaborative virtual reality grant?
Is there an official project page or a repository that could be linked?

cpinter:
It is that one, yes. There is no official project page. The funding agency has not published such a page, and we have not created a public page for it.

lassoan (approved):
perfect, thank you

Issue #2934, opened 2020-03-12
Console output logging for Windows


slicerbot:
_This issue was created automatically from an original [Mantis Issue](https://mantisarchive.slicer.org/view.php?id=2934). Further discussion may take place here._

jamesobutler:
@lassoan I see on a Windows failed test case (http://slicer.cdash.org/test/16313164) what looks to be actual output which I've not seen on the Windows factory platform before. Was this issue intentionally fixed?  Or is the output that I'm seeing some different?

lassoan:
SlicerApp-real.exe is still built as a GUI application - without console output.

The test you referred to uses a different executable (D:/D/P/Slicer-0-build/Slicer-build/bin/Release/qMRMLWidgetsCxxTests.exe), which is a console application and therefore CTest can capture the output of it.

We fixed the launcher 1-2 years ago so that it can forward output of launched console applications to its own output.

lassoan:
After https://github.com/Slicer/Slicer/commit/2d7cb882ef1a3d513f03fcfd6500a3e1a4df1b82, the SlicerApp-real.exe application is a console application by default, so its output can be captured.

To make AppLauncher (Slicer.exe) behave the same way as on linux, it could be updated to detect if a console is available (using AttachConsole) and if that's the case then write the child process output to the console using WriteConsoleOutput. This would make Slicer.exe act as a console application when started from a console, but would not create a console when Slicer.exe is started from GUI.

jamesobutler:
https://github.com/Slicer/Slicer/commit/2d7cb882ef1a3d513f03fcfd6500a3e1a4df1b82 was reverted in https://github.com/Slicer/Slicer/commit/0db2e6e8f7c4d3d947a2a55fb7fed7b213f618b6 due to Slicer startup issues.

lassoan:
I've fixed the issue (a wrong flag was used to decide if WIN32 build flag should be added). I'll test it some more and if everything works well, I'll push the fixed commit.

jamesobutler:
@lassoan Would you consider this issue to be complete now?

lassoan:
The only limitation is that the default launcher on Windows is not a console application. [Workarounds are described in the documentation](https://slicer.readthedocs.io/en/latest/developer_guide/debugging/tips.html#console-output-on-windows) and users can overwrite `Slicer.exe` `bin/PythonSlicer.exe` (which is a console launcher), so I think we can close this issue.

PR #5843, opened 2021-09-09
ENH: Build SlicerApp-real.exe as a console application to enable console output

lassoan:
If SlicerApp-real.exe is built as a console application then it is possible to capture and redirect its output (to files, display, any other software), and test outputs also become visible on dashboard builds.

If the launcher is still built as a Windows GUI application then no command console is displayed when Slicer is started.

Added instructions (in Docs\developer_guide\debugging\overview.md) for accessing the console output when launcher is a Windows GUI application.

re #2934


lassoan:
After this is integrated, the AppLauncher could be updated to detect if a console is available (using [AttachConsole](https://docs.microsoft.com/en-us/windows/console/attachconsole)) and if that's the case then write output to is (using [WriteConsoleOutput](https://docs.microsoft.com/en-us/windows/console/writeconsoleoutput)). This would make Slicer.exe fully behave as a console application when a console is available but would not create a console when Slicer.exe is started from GUI.

pieper:
Sounds like really nice behavior.

lassoan:

sjh26 (approved):
🦖

lassoan:
@sjh26 thanks for the review, I'll merge it now.

PR #5846, opened 2021-09-13
Fix extensions manager integration

jcfr:


Base/QTCore/qSlicerCoreApplication.cxx
```diff
@@ -425,7 +425,7 @@ void qSlicerCoreApplicationPrivate::init()
 
   // Clear extension server settings with requested serverAPI changed
   QSettings extensionsSettings(model->extensionsSettingsFilePath(), QSettings::IniFormat);
-  QString savedServerAPI = extensionsSettings.value("Extensions/ServerAPI").toString();
+  int savedServerAPI = extensionsSettings.value("Extensions/ServerAPI").toInt();
```
lassoan, line 428: ServerAPI enum value is an internal implementation detail that may change in the future and we should avoid exposing it directly in the settings. We could have ServerAPIToString/ServerAPIFromString methods in qSlicerExtensionsManagerModel and use those to convert between code strings that can be stored in the .ini file and displayed to users (`Midas_v1`, `Girder_v1`) and the internal enum.

lassoan (commented):
It looks good, I just added a small comment.

Base/QTCore/qSlicerCoreApplication.cxx
jcfr: > ServerAPIToString/ServerAPIFromString methods in qSlicerExtensionsManagerModel

Great idea. Following commit implement this.

PR #5853, opened 2021-09-14
BUG: DICOMVolumeSequencePlugin.py  -Export Volume Sequences

carjohnson:
These modifications allow for the DICOM export of a 4D Volume Sequence

- Invalid syntax when parsing the time string 'tm' in function datetimeFromDicom
   Change from 'string.split' to 'str.split'

- Added missing tags in export function for 'Patient Birth Date' and 'Patient Sex' required by the class DICOMExportScalarVolume
   See issue: https://discourse.slicer.org/t/exporting-sequences-as-dicoms-does-not-work/18935


lassoan (approved):
This looks great, thank you!

PR #5859, opened 2021-09-15
ENH: Add GitHub workflow for updating Slicer.crt certificate bundle

jcfr:


jcfr:
Merging, the workflow has been developed and tested independently in https://github.com/jcfr/ci-sandbox

PR #5862, opened 2021-09-15
Markups constrain curves to surface

Connor-Bowley:
Update to allow all markup curve types (open/closed, spline/linear/etc) to be able to constrain to a model surface. The current curve type "Shortest distance on surface" is unchanged in its behavior.

It is possible to use this new option in conjunction with the Curve Cut from the Dynamic Modeler. However, this new option (e.q. Spline, constrain to surface) does _not_ work with the Boundary Cut from the Dynamic Modeler. I made no changes to either, so I do not know if this makes sense for what Boundary Cut is supposed to be compared to Curve Cut, but it was surprising to me. Note: Boundary Cut still works with "Shortest distance on surface" as it did before.

Because I found it surprising that the Boundary Cut did not work with the Spline/Linear/etc constrained to surface, I would like that to be documented somewhere, but I am not sure where.


Modules/Loadable/Markups/MRML/vtkCurveGenerator.cxx
```diff
@@ -627,12 +628,28 @@ int vtkCurveGenerator::GeneratePointsFromFunction(vtkPoints* inputPoints, vtkPoi
...
     double sampleParameter = double(pointIndex) / ((double)(totalNumberOfPoints - 1));
     double curvePoint[3];
     this->ParametricFunction->Evaluate(&sampleParameter, curvePoint, nullptr);
+
+    if (inputSurface)
+      {
+      vtkIdType id;
+      double closestPointDist;
+      int subId;
+      //adjust curvePoint to be on the surface
```
jcfr, line 649:
```suggestion
      // adjust curvePoint to be on the surface
```

Modules/Loadable/Markups/MRML/vtkCurveGenerator.cxx
```diff
@@ -627,12 +628,28 @@ int vtkCurveGenerator::GeneratePointsFromFunction(vtkPoints* inputPoints, vtkPoi
...
   double previousPoint[3] = { 0.0 };
   for (int pointIndex = 0; pointIndex < totalNumberOfPoints; pointIndex++)
     {
     double sampleParameter = double(pointIndex) / ((double)(totalNumberOfPoints - 1));
     double curvePoint[3];
     this->ParametricFunction->Evaluate(&sampleParameter, curvePoint, nullptr);
+
+    if (inputSurface)
+      {
+      vtkIdType id;
```
jcfr, line 646:
```suggestion
      vtkIdType cellId = -1;

File: Modules/Loadable/Markups/MRML/vtkCurveGenerator.cxx [hunk 3]
```diff
@@ -627,12 +628,28 @@ int vtkCurveGenerator::GeneratePointsFromFunction(vtkPoints* inputPoints, vtkPoi
...
+      double closestPointDist;
+      int subId;
```
jcfr, line 648:
```suggestion
      int subId = -1;
```

Modules/Loadable/Markups/MRML/vtkCurveGenerator.cxx
```diff
@@ -627,12 +628,28 @@ int vtkCurveGenerator::GeneratePointsFromFunction(vtkPoints* inputPoints, vtkPoi
...
   for (int pointIndex = 0; pointIndex < totalNumberOfPoints; pointIndex++)
     {
     double sampleParameter = double(pointIndex) / ((double)(totalNumberOfPoints - 1));
     double curvePoint[3];
     this->ParametricFunction->Evaluate(&sampleParameter, curvePoint, nullptr);
+
+    if (inputSurface)
+      {
+      vtkIdType id;
+      double closestPointDist;
```
jcfr, line 647:
```suggestion
      double closestCurvePoint[3] = { 0.0 };
      double closestCurvePointDist2 = VTK_DOUBLE_MAX;
```

Modules/Loadable/Markups/MRML/vtkCurveGenerator.cxx
```diff
@@ -627,12 +628,28 @@ int vtkCurveGenerator::GeneratePointsFromFunction(vtkPoints* inputPoints, vtkPoi
...
+      int subId;
+      //adjust curvePoint to be on the surface
+      this->SurfaceCellLocator->FindClosestPoint(curvePoint, curvePoint, id, subId, closestPointDist);
+      }
+
     outputPoints->InsertNextPoint(curvePoint);
```
jcfr, line 653:
```suggestion
      this->SurfaceCellLocator->FindClosestPoint(curvePoint, closestCurvePoint, cellId, subId, closestPointDist);
      outputPoints->InsertNextPoint(closestCurvePoint);
      }
    else
      {
      outputPoints->InsertNextPoint(curvePoint);
      }
```

jcfr:
### Comment related to continuous integration

Considering the branch was created before changes have been integrated to `master`, the test implemented in [.circleci/config.yml](https://github.com/Slicer/Slicer/blob/b4c6004ef2709bf71843b17a3364e2c0d22e455e/.circleci/config.yml#L10-L24) return changed file in the `SuperBuild` directory.

For reference, the following set of command is used by the CircleCI worker:

```
git clone --no-checkout git@github.com:Slicer/Slicer.git
cd Slicer/
git fetch --force origin '+refs/pull/5862/head:refs/remotes/origin/pull/5862'
CIRCLE_BRANCH=pull/5862
CIRCLE_SHA1=dc337fc0dfb9f5f0f639f03d92a1adabf2451ec4
git checkout --force -B "$CIRCLE_BRANCH" "$CIRCLE_SHA1"
git --no-pager log --no-color -n 1 --format='HEAD is now at %h %s'
```

and the test returns the following:

```
git diff --name-only master
```

Truncated output:

```
[...]
SuperBuild.cmake
SuperBuild/External_CTK.cmake
SuperBuild/External_ITK.cmake
SuperBuild/External_VTK.cmake
```

Modules/Loadable/Markups/MRML/vtkCurveGenerator.cxx
```diff
@@ -627,13 +628,34 @@ int vtkCurveGenerator::GeneratePointsFromFunction(vtkPoints* inputPoints, vtkPoi
...
     {
     double sampleParameter = double(pointIndex) / ((double)(totalNumberOfPoints - 1));
     double curvePoint[3];
     this->ParametricFunction->Evaluate(&sampleParameter, curvePoint, nullptr);
-    outputPoints->InsertNextPoint(curvePoint);
+
+    if (inputSurface)
+      {
+      vtkIdType cellId = -1;
+      double closestCurvePointDist2 = std::numeric_limits<double>::max();
```
jcfr, line 647:
```suggestion
      double closestCurvePointDist2 = VTK_DOUBLE_MAX;
```

In other classes making use `FindClosestPoint`, the closest point is initialized with `VTK_DOUBLE_MAX`. For sake of consistency, I suggest we keep the same convention.

https://github.com/Slicer/Slicer/blob/b4c6004ef2709bf71843b17a3364e2c0d22e455e/Modules/Loadable/Markups/VTKWidgets/vtkSlicerCurveRepresentation2D.cxx#L361-L371

Connor-Bowley: Is there any plan/desire to update from using `VTK_DOUBLE_MAX` in Slicer to the more standard and widely known `std::numeric_limits`?

jcfr: I am not aware of effort looking to remove these.
Looking at VTK master, it is still in used. See https://github.com/Kitware/VTK/search?q=VTK_DOUBLE_MAX

If you would like, you could discuss this on [VTK discourse](https://discourse.vtk.org). Cc'ing [Sean McBride](https://discourse.vtk.org/u/seanm/summary) may be sensible as they have been very [active](https://github.com/Kitware/VTK/commits?author=seanm) in modernizing the code base.

Connor-Bowley: Updated

jcfr:
Also, providing a table like this:

```
| Before | After |
|--|--|
| <image_before> | <image_after> |
```

would be helpful to understand how the UI associated with the `Curve Settings` changed.

jcfr:
To summarize, here are the next steps:
1. update documentation
2. provide screenshot
3. collect feedback from @lassoan and others
4. draft a post similar to [New feature: draw curve on surface](https://discourse.slicer.org/t/new-feature-draw-curve-on-surface/10431)

lassoan:
Thank you for working on this. Improving constraining markups to surface meshes could be improved for sure.

Snapping of curve points to closest surface points is very simple, so if it was a reasonably good way of snapping curves to surfaces then it would make sense to add it to the curve generator. However, there is no one single "correct" way of constraining a curve to a surface, and simply snapping to closest surface point works very poorly on concave surfaces. That's why we use two non-trivial snapping algorithms now:
- projection to surface based on surface normals (in curve Resample section)
- shortest path search on surface (in Curve settings section)

Neither of them is perfect. The projection based method is sensitive to quality of surface normals and complexity of the surface - we could probably smooth the normals (or compute or allow the user to specify a projection center point or plane) and use those for point projection. The shortest path search on surface method does not cross through cells, so the created path can be very noisy - it would be nice to port [this](https://twitter.com/nmwsharp/status/1337492932845244417?s=19) algorithm into VTK and add an option to use it here.

> Because I found it surprising that the Boundary Cut did not work with the Spline/Linear/etc constrained to surface, I would like that to be documented somewhere, but I am not sure where.

Since the Dynamic Modeler module is installed along with Slicer core, it should be documented in Slicer readthedocs - https://slicer.readthedocs.io/en/latest/user_guide/modules/index.html#surface-models

Connor-Bowley:
@lassoan Just for clarity, are you ok with adding this with the use of snapping of curve points to closest surface point now and possibly updating later, when say the geodesic flipping algorithm you linked to is in VTK, or are you requesting changes for this pull request?

lassoan:
> are you ok with adding this with the use of snapping of curve points to closest surface point now

Sorry, I'm not convinced. Snapping to closest point is probably the simplest  possible method. That is a big advantage. However, the problem is that it works well so rarely, so even if we make it available as an option, I would not build it into the curve generator, but add it as a subsequent step, similarly to the other curve snapping options. In term of GUI, it should be added as an additional option in either Curve settings (next to shortest path on surface) or in resampling section.

It could also make sense to move projection method from the resampling section to the Curve settings section so that it can be applied dynamically (without resampling). In fact, this probably this what what you actually wanted to achieve.

Connor-Bowley:
> I would not build it into the curve generator, but add it as a subsequent step, similarly to the other curve snapping options

The "Shortest distance on surface" is in the curve generator right now, and that was the reason I put my changes in the curve generator. I hadn't thought to look at the resampling.

> It could also make sense to move projection method from the resampling section to the Curve settings section so that it can be applied dynamically (without resampling). In fact, this probably this what what you actually wanted to achieve.

I agree that this should do what I wanted. It makes sense to me to pull this from the resampling to the general curve creation so it is always available even when not resampling. 

So this is what I propose:
1) I refactor so the resampling and general curve creation both use the same methods for projection to surface (not inside the curve generator). The projection options will be either just the surface normal method or the surface normal method and the shortest distance to surface method. 
2) The "Shortest distance on surface" remains in the curve generator as it is (as far as I can tell) actually generating a curve instead of just projecting one.
3) I do some GUI and backend magic to prevent people from trying to use both "Shortest distance on surface" and another projection to surface method.

My reasoning for leaving "Shortest distance on surface" alone in the curve generator is this: we seem to have 2 distinct ideas going on here. 1) curve generation 2) projecting a curve to a surface. Algorithms like "Shortest distance on surface" or the geodesic flipping algorithm you linked are (in my mind) legitimately generating curves that have the property of being on a surface.

Connor-Bowley:
@lassoan Does this sound good to you?

lassoan:
> The "Shortest distance on surface" is in the curve generator right now, and that was the reason I put my changes in the curve generator. I hadn't thought to look at the resampling.

I see, that sounds logical. The only issue with putting this simple closest distance snapping method into the filter that it does not really work (only when all the curve points are close to the surface), and the normal projection method is not very robust.

> It makes sense to me to pull this from the resampling to the general curve creation so it is always available even when not resampling.

Exactly.

> The projection options will be either just the surface normal method or the surface normal method and the shortest distance to surface method.

There are a few possible combinations:
- A. Snap control points to closest points on surface and connect them using shortest distance on surface.
- B. Project all curve points using surface normals.
- C. Project control points using surface normals and connect them using shortest distance on curve.

We currently use A and B, so we need to keep providing those. If you want, you can add C in addition, but I don't think it is necessary, because usually there are not too many control points, and the user can ensure that they are all close to the surface (projection is a must when the curve points are constrained to the surface, because it can be hard to ensure that all curve points are close to the surface everywhere).

> My reasoning for leaving "Shortest distance on surface" alone in the curve generator is this: we seem to have 2 distinct ideas going on here. 1) curve generation 2) projecting a curve to a surface.

Yes. One group of methods is "snap then interpolate" the other is "interpolate then snap". We now have one of each. In the future we may add more methods in both groups.

lassoan (changes requested):
Setting the state to "request changes" to make it more clear that code update is needed (and not code review).

lassoan:
I got a build error:

    28>C:\D\S4\Modules\Loadable\Markups\MRML\vtkMRMLMarkupsCurveNode.cxx(1687,1): error C2027: use of undefined type 'std::array<double,3>'

Adding `#include <array>` to vtkMRMLMarkupsCurveNode.cxx solved it.

lassoan:
I've tested the code and it works very nicely! There was only one major issue: when either the markups node or the surface node was under a transform then the curve was not snapped to the surface anymore. I saw that you removed code that handled transforms - probably you need to add that back. The transforms on the surface or the markup node can also be changed dynamically, so transform modified events should be observed (I think these worked, too, just make sure the mechanisms remain functional).

Connor-Bowley:
Clarification questions regarding transforms: I understand that a curve should snap to a transformed model node even as it is moving. If one applies a transformation to a _curve_ does it makes sense to only apply the transform to the control points only and not the actual curve?

lassoan:
> Clarification questions regarding transforms: I understand that a curve should snap to a transformed model node even as it is moving. If one applies a transformation to a _curve_ does it makes sense to only apply the transform to the control points only and not the actual curve?

I see now, there is indeed subtle difference in the end result depending on the order of operations.

- Option A. interpolate curve points, transform curve points, snap curve points to surface
- Option B. transform control points, interpolate curve points, snap curve points to surface- Option C. snap control points to surface, interpolate curve points

Option C would be very bad, it would result in control points not actually snapped to the surface.

Option A and B is quite similar, but we should use Option A, which is what we use already when we don't constrain to surface. Option B (transform control points and then interpolate) would lead to poor results when we use a non-linear warping transform.

lassoan:
> For completeness, Option D would be interpolate curve points, snap curve points to surface, transform curve points and control points
> 
> This option makes the most sense to me as it actually transforms the curve. Notably this is _not_ what the existing functionality was for shortest distance on surface

I'm not sure I understand this correctly. If you transform the curve points and control points _after_ snapping to the surface then the points would not be on the surface anymore. The visible end result must be that the curve is on the surface under all circumstances, no matter if the transforms on the curve and model were hardened or not.

Connor-Bowley:
> The visible end result must be that the curve is on the surface under all circumstances, no matter if the transforms on the curve and model were hardened or not.

I guess part of where I am getting hung up on is that every option seems like it is breaking a promise in some way. Option D breaks the promise that the curve is on the surface. Options A-C breaks the promise that the curve gets transformed by the transform (like a curve without snap to surface is transformed). I don't know which is more important: snap to surface or make transforms consistent with most of the other items (models, ROI, segmentations)

Connor-Bowley:
I feel like I am going down a rabbit hole a little bit here. Lets take Option A: interpolate, transform, snap

What does GetCurve() return vs GetCurveWorld()? There doesn't really exist a curve before the world transform that remotely resembles the post transform post snap curve. GetCurve could return
1) The curve with no snapping
2) The curve that snaps to the same model with the control points pre transform
3) The curve from GetCurveWorld run through an inverse transform?

All of these seem wrong.

lassoan:
> I feel like I am going down a rabbit hole a little bit here. Lets take Option A: interpolate, transform, snap
> 
> What does GetCurve() return vs GetCurveWorld()? There doesn't really exist a curve before the world transform that remotely resembles the post transform post snap curve. GetCurve could return
> 
> 1. The curve with no snapping
> 2. The curve that snaps to the same model with the control points pre transform
> 3. The curve from GetCurveWorld run through an inverse transform?
> 
> All of these seem wrong.

GetCurve() should return the curve before any constraint or transform is applied. GetCurveWorld() returns the curve after all transforms and constraints are applied. This should be specified in the node API documentation to make the behavior clear.

PR #5865, opened 2021-09-16
ENH: Add RenderBlocker context manager to call pauseRender/resumeRender

Sunderlandkyl:
This commit adds a context a manager that calls slicer.app.pauseRender() on enter and slicer.app.resumeRender() on exit.

Previously this would be written as:
```
try:
  slicer.app.pauseRender()
  # Do things
finally:
  slicer.app.resumeRender()
```

Now can be written as:
```
with slicer.util.RenderBlocker():
  # Do things
```

### DIFF
--- a/Base/Python/slicer/util.py
+++ b/Base/Python/slicer/util.py
@@ -1350,6 +1350,31 @@ class NodeModify:
   def __exit__(self, type, value, traceback):
     self.node.EndModify(self.wasModifying)
 
+class RenderBlocker:
+  """
+  Context manager to conveniently pause and resume view rendering. This makes sure that we are not displaying incomplete states to the user.
+  Pausing the views can be useful to improve performance and ensure consistency by skiping all rendering calls until the current code block has completed.
+
+  Code blocks such as::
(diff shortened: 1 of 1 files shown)


lassoan (commented):
Thank you, This looks good to me.

Could you add the excellent usage example to the class documentation?

Also describe there why blocking the rendering is useful (to improve performance and ensure consistency of all view contents - prevent all renderings until all computations in the code block are completed).

lassoan (changes requested):
Setting the state to "request changes" to make it more clear that code update is needed (and not code review).

lassoan:
@Sunderlandkyl let me know if you prefer if I make the little requested change. We would like to close all the pending pull requests so that we can better see which requires review or more work. Thank you.

Sunderlandkyl:
Sorry, forgot about this PR. I'll update it in a minute.

Sunderlandkyl:
Ok. it should be updated.

lassoan (approved):
Thank you!

PR #5870, opened 2021-09-17
Associate valid download url with CDash submission

### LINKED ISSUE #5768: Download link reported on CDash reference incorrect server
Download link reported on CDash reference incorrect server

## Steps to reproduce

* Go to https://slicer.cdash.org/index.php?project=SlicerPreview&date=2021-08-03

* Click on the "package" icon: image

* Then observe that the link indicated `slicer.kitware.com`

image

## Expected behavior

It should reference `https://slicer-packages.kitware.com/`

## Environment

NA

jcfr:
Fixes #5768


jcfr:
Updated downloading is now associated:

image

image

Remaining problem is that the `SHA512` seems to be invalid. Indeed, clicking on the link returns the following:

```json
{
    "message": "File not found.",
    "type": "rest"
}
```

Hash are different:
```
| | hash |
|--|--|
| local | 5179f4becefcb1a5e311c47dd1a074571b21a6a2b5f6d35adc0e48516a42334ce3b159e35dbb5269105ee3eeb7bc122102755115129017a1b6459faaccd76946 |
| server | a5bd2b52dabddf97e109181a7b240691c3856616dc8ea69fd3850db2e72180b67267b6ec35aaf1941cd64df9158d7b4e1d429b2370aeab2a33ff1f7a285e458b |
```

jcfr:
### Success on a first upload

Additional testing confirm that the hash is properly computed and the link is valid when extension is uploaded for the first time.

`62195019e96769538e5fa724e20f62598845ce809f0a2143c25d342de6dbce99959fa443138bb9a65d61ffdb2d39e7ecf8f4678030d770c05859f2e3d888594e`

image

image


### Issue on a follow-up build and upload of the same extension

That said,  following a second attempt to package and upload the same extension, the server skip the update of the package and then the hash of the computed link using the local package is different from the package already updated.

jcfr:
Considering the behavior described in the previous comment is "expected" (see https://github.com/Slicer/Slicer/issues/5888 for more details), I will integrate the change.

Indeed, initial submission of packages for either Slicer applications or extensions will be **valid**.

PR #5872, opened 2021-09-17
ENH: Use markups fiducial point as cursor in Endoscopy module

lassoan:
The user can more easily adjust size and appearance (scaling, projection, occluded visibility, etc.) more flexibly than a model node.
In the module logic class, developers can choose between using a model or a markups fiducial.


pieper (approved):
Good improvement 👍

lassoan:
Thanks for the review.

PR #5828, opened 2021-08-27
ENH: Multiple color bars via displayable manager support

MichaelColonel:
This is a WIP of the issue #4891.

Color bar will be controlled via vtkMRMLColorBarDisplayNode. Display node can observe either vtkMRMLColorTableNode directly or vtkMRMLDisplayableNode and then get a color table node for color bar visualization.

If there are no obvious errors or mistakes, the next step will be modification of the "Colors" module GUI in order to change properties of the selected color bar actor.

PR #5874, opened 2021-09-18
Slice annotation disabled

riep:
If a user wants to edit custom corner annotation texts, having disabled the sliceViewAnnotationWidget of slicer, there is still an empty text that interfere with the user edition.
If sliceViewAnnotationWidget is turned off, we should remove observers and do not create other.


pieper (approved):
I didn't have a chance to try it, but this code looks okay to me.

It seems there's almost no testing for some of these fairly sophisticated features.  I'm not sure how best to automate testing of this, so for now please manually check that the other features work well with this change.

lassoan:
The data probe was ported from TCL to Python very early and then new features were added under time pressure, without much thinking about a good design. Since it is a performance-critical core feature it would greatly benefit from a complete redesign to make it more robust and flexible.

Moving color bar to separate displayable manager (https://github.com/Slicer/Slicer/pull/5828#) will simplify things a bit. Corner annotation content should also be more flexible and allow placing any content to any corner, including content of any MRML text node. This will allow many modules to display content in corner annotations at once instead of the current situation when there can only be one owner module of each corner annotation.

PR #5875, opened 2021-09-18
ENH: Undo Redo segment editor buttons stay in same position

jamesobutler:
This moves the Undo and Redo buttons in the Segment Editor module to be under the effects groupbox. This is motivated by the desire for these buttons to not move when different segment editor effects are selected.

See https://discourse.slicer.org/t/new-segment-editor-layout-vertical-effect-toolbar/19649/9?u=jamesobutler for more details about this development.

| Current | This PR |
|----------|---------|
|image|image|

### DIFF
--- a/Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.cxx
+++ b/Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.cxx
@@ -260,6 +260,9 @@ public:
   /// Button group for the effects
   QButtonGroup EffectButtonGroup;
 
+  /// Button group for the UndoRedoGroupBox
+  QButtonGroup UndoRedoButtonGroup;
+
   /// These volumes are owned by this widget and a pointer is given to each effect
   /// so that they can access and modify it
   vtkOrientedImageData* AlignedMasterVolume;
@@ -497,6 +500,9 @@ void qMRMLSegmentEditorWidgetPrivate::init()
 
   this->EffectButtonGroup.setExclusive(true);
   QObject::connect(&this->EffectButtonGroup, SIGNAL(buttonClicked(QAbstractButton*)), q, SLOT(onEffectButtonClicked(QAbstractButton*) ) );
+
+  this->UndoRedoButtonGroup.addButton(this->UndoButton);
+  this->UndoRedoButtonGroup.addButton(this->RedoButton);
 
   // Create layout for effect options
   QVBoxLayout* layout = new QVBoxLayout(this->EffectsOptionsFrame);
@@ -1179,14 +1185,33 @@ void qMRMLSegmentEditorWidget::updateEffectList()
     auto gridLayout = dynamic_cast<QGridLayout*>(d->EffectsGroupBox->layout());
     gridLayout->addWidget(effectButton, rowIndex, columnIndex);
     if(columnIndex == d->EffectColumnCount - 1)
-    {
+      {
       columnIndex = 0;
       ++rowIndex;
-    }
+      }
     else
-    {
+      {
       ++columnIndex;
+      }
     }
+
+  // Set UndoRedoGroupBox buttons with same column count as EffectsGroupBox
+  rowIndex = 0;
+  columnIndex = 0;
+  QList<QAbstractButton*> undoRedoButtons = d->UndoRedoButtonGroup.buttons();
+  foreach(QAbstractButton* button, undoRedoButtons)
+    {
+    auto undoRedoGridLayout = dynamic_cast<QGridLayout*>(d->UndoRedoGroupBox->layout());
+    undoRedoGridLayout->addWidget(button, rowIndex, columnIndex);
+    if(columnIndex == d->EffectColumnCount - 1)
+      {
+      columnIndex = 0;
+      ++rowIndex;
+      }
+    else
+      {
+      ++columnIndex;
+      }
     }
 }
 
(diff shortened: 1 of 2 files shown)


Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.cxx
```diff
@@ -1188,6 +1201,25 @@ void qMRMLSegmentEditorWidget::updateEffectList()
...
+  // Add UndoRedoGroupBox buttons with same column count as EffectsGroupBox
+  rowIndex = 0;
+  columnIndex = 0;
+  foreach(QAbstractButton* button, undoRedobuttons)
+    {
+    button->show();
+    auto undoRedoGridLayout = dynamic_cast<QGridLayout*>(d->UndoRedoGroupBox->layout());
+    undoRedoGridLayout->addWidget(button, rowIndex, columnIndex);
+    if(columnIndex == d->EffectColumnCount - 1)
+    {
```
lassoan, line 1214: Indentation

Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.cxx
```diff
@@ -1188,6 +1201,25 @@ void qMRMLSegmentEditorWidget::updateEffectList()
...
+      columnIndex = 0;
+      ++rowIndex;
+    }
+    else
+    {
```
lassoan, line 1219: Indentation

Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.cxx
```diff
@@ -1141,6 +1147,13 @@ void qMRMLSegmentEditorWidget::updateEffectList()
     QLayoutItem *child;
     while ((child = d->EffectsGroupBox->layout()->takeAt(0)) != 0);
     }
+  QList<QAbstractButton*> undoRedobuttons = d->UndoRedoButtonGroup.buttons();
+  foreach (QAbstractButton* button, undoRedobuttons)
+    {
+    button->hide();
```
lassoan, line 1153: Do not show/hide the buttons here. It would interfere with qMRMLSegmentEditorWidget::undoEnabled() or any other low-level customizations that developers might apply (we don't guarantee anything when developers to low-level customizations, but there is also no reason to unnecessarily modify widget properties).

Check undoEnabled method anyway, probably it would be better to show/hide the entire frame instead of show/hide each button.

Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.cxx
```diff
@@ -1141,6 +1147,13 @@ void qMRMLSegmentEditorWidget::updateEffectList()
...
+    QLayoutItem *child;
+    while ((child = d->UndoRedoGroupBox->layout()->takeAt(0)) != 0);
```
lassoan, line 1155: Are you sure you need to remove the widgets from the layout? I think if you add them to the desired layout position then if the position is different then they are removed automatically (but if the position remains the same then nothing is changed and we can avoid potential flickering).

Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.cxx
```diff
@@ -1188,6 +1201,25 @@ void qMRMLSegmentEditorWidget::updateEffectList()
       ++columnIndex;
     }
     }
+
+  // Add UndoRedoGroupBox buttons with same column count as EffectsGroupBox
+  rowIndex = 0;
+  columnIndex = 0;
+  foreach(QAbstractButton* button, undoRedobuttons)
+    {
+    button->show();
```
lassoan, line 1210: should not be needed

lassoan (commented):
Thank you, looks good overall, just added a few nitpicks in inline comments.

Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.cxx
jamesobutler: I did this just to make it similar logic as the effect frame in case other buttons made it into this new frame. I think you’re right that adding all the widgets in the grid layout again in a different position when redefining effect column count that it will remove the conflict of a widget already present in the specified cell.

lassoan: The effect buttons initialization is not perfect. On some computers (maybe slower ones, maybe better visible in debug mode and/or when many effects are registered) there is some flickering when the segment editor widget displayed first. Maybe the buttons are momentarily appear over the screen before they got moved into the layout. I would not delay this pull request with investigation of this small issue, just mention it because it might not be ideal to copy how the effect button layout is created/updated.

jamesobutler:
Pushed updates that I need to test as a sanity test. Mouse clicked on the wrong thing and hit "Rebuild" instead of "Build" on my Slicer project solution...so I'm waiting on it

jamesobutler:
@lassoan Ok I've pushed and tested updates with the Undo Redo frame being shown/hidden based on the `setUndoEnabled` method.

lassoan (approved):

lassoan:
Thanks a lot, this works very nicely, saving a bit more space.

jamesobutler:
We probably need to update screenshots on https://slicer.readthedocs.io/en/latest/user_guide/modules/segmenteditor.html prior to the stable release with the latest visual changes to the Segment Editor module.

Issue #5367, opened 2021-01-04
New Slicer logo rollout


jamesobutler:
The new Slicer logo was originally brainstormed and thoroughly discussed in https://discourse.slicer.org/t/proposed-new-slicer-logo/11295. The next step is the rollout to replace all the old Slicer logo usages.
![image](https://user-images.githubusercontent.com/15837524/103544638-6999fa80-4e6e-11eb-94c4-5c42cd2f63f8.png)

The new Slicer logo images files can be found at the google folder link on this post: https://discourse.slicer.org/t/proposed-new-slicer-logo/11295/97. 
I will also upload a zip file here in case that link expires for some reason: [SlicerNewLogo.zip](https://github.com/Slicer/Slicer/files/5762176/SlicerNewLogo.zip)

Primary Sites:
- [x] Slicer Desktop Icon and other uses in the source code: See ~~https://github.com/Slicer/Slicer/pull/5366~~ https://github.com/Slicer/Slicer/pull/5877
- [x] API Documentation: https://apidocs.slicer.org/master/index.html -- the icon is now updated, just the site is not updated (see #4906)
- [x] Organization Website: https://www.slicer.org/
  - [X] New website design: https://github.com/Slicer/slicer.org/tree/landing-page-redesign. 
![image](https://user-images.githubusercontent.com/15837524/103547055-e24e8600-4e71-11eb-8864-7aaf783b8750.png)
  - Old website design: https://github.com/Slicer/slicer.org/tree/slicer-org and https://github.com/mhalle/slicer4-download. if new website design hasn't rolled out yet
- [x] Slicer download subdomain (https://download.slicer.org/)
- [x] Discourse Forum Logo: https://discourse.slicer.org/
- [x] Discourse Forum Favicon: https://discourse.slicer.org/
- [x] ~~Slicer Wiki: https://www.slicer.org/wiki/Main_Page~~ -- we'll keep the old logo there, since there is/will be nothing for Slicer5
- [X] ReadTheDocs Logo: https://slicer.readthedocs.io/en/latest/
- [x] ~~ReadTheDocs Image resources https://github.com/Slicer/Slicer/releases/download/docs-resources/user_interface_main_window.png~~ -- the new logo will gradually replace the old one as new screenshots are created
- [x] ~~Slicer sticker design: as seen in this [post](https://discourse.slicer.org/t/feedback-on-slicer-sticker-design/6395)~~ -- designer of the new Slicer sticker will use the new logo
[...]

PR #5877, opened 2021-09-19
ENH: Switch to new Slicer logo in module panel

lassoan:
This is a follow-up to https://github.com/Slicer/Slicer/pull/5820 (did not force-push to the same pull request to preserve the content of that branch).

If this pull request is merged then these pull requests can be closed:
- [ ] https://github.com/Slicer/Slicer/pull/5820
- [ ] https://github.com/Slicer/Slicer/pull/5366

---

Switch to the new Slicer logo: it is more compact than the old one and works with both dark and bright background (see #5367).

The logo was moved to the module panel dockable widget's header to save space. See discussion at https://github.com/Slicer/Slicer/pull/5820 for more details.

Welcome module before:

image

Welcome module after:

image

**Available height for module GUI increased by 18%:**

image

**Module panel can be still drag-and-dropped:**

https://user-images.githubusercontent.com/307929/133943218-8ce39c58-cfbe-4f25-a60d-91c22a8f8be8.mp4

---

For **crisp rendering of the new icon**, Qt::AA_UseHighDpiPixmaps application option is now enabled.
If higher-resolution variants of icons are provided by ...@2x.png, ...@3x.png, ... names then they are now used on high-DPI displays.

Before:

image

After:

image

Setting up high-resolution icons on high-DPI screens will come in a follow-up commit (basically we just need to rename our high-resolution icon versions to follow the basename@2x.png, basename@3x.png, ... naming convention).


jamesobutler (commented):
Thanks for looking into the higher quality pixmaps for icons! I was still using the same resolution when replacing images for the new slicer logo as part of #5366. Do you think it will be ok to include the new logo here in the module panel title, but still have the old logo for the app icon? I know it would be great to include the saved space sooner but maybe it should be tied to the transition of the other logo updates?

cpinter:
Thank you @lassoan for implementing this! It looks perfect to me, and a good compromise between branding and space.

lassoan:
> Do you think it will be ok to include the new logo here in the module panel title, but still have the old logo for the app icon? I know it would be great to include the saved space sooner but maybe it should be tied to the transition of the other logo updates?

Makes sense. I'll add your other logo replacement commit to this pull request.

lassoan:
Now all the Slicer logos are updated to the new one:

image

pieper (approved):
Looks good to me 👍

PR #5887, opened 2021-09-22
ENH: Improve icon set

lassoan:
This is a work in progress and the pull request is submitted so that we can start discussion before proceeding with further updates.

A few new icons are added that use more modern style (flat, duotone) and have higher resolution, and look OK over both light and dark backgrounds. All icons are based on Google material design icon set (https://fonts.google.com/icons), edited in Inkscape, exported in native 24x24 and double 48x48 resolution.

image

image

A summary of the overall effort is described on this labs page: https://github.com/Slicer/Slicer/wiki/Modern-Look

Questions:
- Does the workflow of getting SVG from "Google material design" icon set, editing in Inkscape, exporting to 24x24 and 48x48 sound good? *Update:* with carefully designed (SVG with coordinate units set to pixels, all lines aligned to pixel centers, etc.) it is possible to use a single SVG icon in the resource and Qt can scale it. This would simplify icon updates (we would only need to store one file). This is already done for a few icons. We may not want to use this for all icons (e.g., for those that contain text, because I don't know what fonts Qt can render), and we may not have all icons in SVG, so for some icons we may still need to use multiple bitmaps.
- Is this minimalistic style, web-like generally a good direction? It would mean that we would need to replace all the icons that come from the Qt theme (as they are more detailed, 3D-like style). There would still be some inconsistency, as buttons themselves have a kind of 3D/gradient look (but maybe that could be removed by adjusting the color table).
- Should we use the same colors in light and dark mode? Switching over to different icons or somehow dynamically recolor all the icons could result in better contrast, but could be much harder to implement. Or should we add shadows (that would improve contrast but would make it a bit more difficult to create the icons). Or should we just go all-in on dark mode and remove light mode completely?
- What colors should we use? We could use the same 2-3 colors for all icons, or we could use different colors for different categories (e.g., file operations are orange, markups are blue, segmentations is green, ...) but we might run out of nice, distinguishable colors that work well over both light and dark background; and it may be difficult to be consistent. However, more colors can make it easier to find buttons. For example, above the blue module selection buttons separate nicely from the yellow file load/save icons.


pieper:
This is looking good to me.  I like the "practical" flat look with few colors compared to the 2.5D shaded look and being able to more easily change overall themes is a popular feature.  I wish we could do a wholesale replacement but using doing something incremental to test the svg workflow makes sense.

jamesobutler (commented):
Yes the Google icon set definitely has a lot of options that are easily recognizable. I would be all for it!

I'm sure you were just quickly prototyping, but whenever you like I can provide feedback about the icon used to replace the current one. For instance the Slicer save icon in your images shown is the "Save Alt" rather than the preferred "Save". "Save Alt" is very similar to "File Download" which is why I was confused.
image
image
image

jamesobutler:
Regarding colors, yeah that is a tough one. I feel like most times between light and dark modes recently it has been a switch between dark and white colors. Organization of certain colors may work, though I'm not sure if it would look a bit cartoony or just like a bunch of different colors intermixed especially with the icons being used throughout the modules and not just on the toolbar. 

Thinking about OHIF, all the icons are pretty much using the same coloring, but without labels would be a bit difficult to find related things since it all kind of blends together.
image

Looking at say Microsoft Office applications, their ribbon interface generally has similarly colored icons with minor/subtle coloring (for organization) making the difference. Some reds for destructive, green for new things, purple for communication, etc. Maybe icons are brought from the generally grey color scheme and then if organization wants to be added to it, then hints of color are added, but the icon not changing complete color?
image

pieper:
Yes, having a few semantic concepts (new, delete, communicate...) mapped to colors as part of the theme is the best way to deal with light/dark theme changes.

lassoan:
Labels are generally _not_ used in toolbars. Even in ribbons, text is usually limited to groupboxes and rarely used features. Outlook is an outlier among Office applications in that all toolbar buttons are labeled - maybe there is space because there are not too many features to expose and all of them can be considered rarely used?,

Word:

image

PowerPoint:

image

Excel:

image

If we simply displayed labels for toolbar buttons then it would use 4x more space, because it would make every button about 2x higher, and we would need two rows of toolbars to show the same amount of buttons:

image

jamesobutler:
Yes I don't think labels should be shown in a common sized row toolbar. OHIF can do it with limited options and Office can do it with such a large ribbon space.

cpinter:
Thanks for working on this!

> SVG

Great! Having SVG wheneve we can will simplify things.

> Is this minimalistic style, web-like generally a good direction?

How widely used is the Google icon set? Is it something we can expect to be a quasi standard in many applications? Not sure how many of the Qt icons we use are not present in the google one but if we make a list of all icons, or a spreadsheet putting current/new side by side we can crowd source the replacement. I am happy to help with a little icon hunting.

> Light/dark mode

Do we have any sense or statistics on how many users use dark mode? It isn't the default yet (I deleted the ini and it didn't change so I guess not), so it is hard to gauge the acceptance of a possible removal of the light mode, however I wouldn't have a problem with that.
In any case keeping two sets of icons seems like a too large extra maintenance burden.

> What colors should we use?

I think the closer we move in style to OHIF the better. We are sister projects, and it seems to me that in the future Slicer will serve as "OHIF desktop" providing strong rendering, interactive, and real-time support, and vice versa, OHIF providing lightweight image viewing before going into "full mode". So a blue basis seems the best to me.
I like the thematic coloring @jamesobutler suggests as there is in Office (never really noticed really but yes it seems plausible it is the intent), with a base color and the modifiers thematized and grouped. What @lassoan suggests with dividing along file/markup/segmentation seems like a good idea as well. All depends on the google icon set, what it includes and how I guess.

lassoan:
Lots of great ideas!

>> Is this minimalistic style, web-like generally a good direction?
> I think the closer we move in style to OHIF the better.

Harmonizing icons (and maybe look&feel) with OHIF is a good idea and it is the right time to do it. Why would the user need to learn two different brightness/contrast or zoom icons? Unfortunately, OHIF uses FontAwesome, which is not free. Inexpensive, probably would not hurt anyone to pay for it, but still, it is an extra hassle for developers to deal with and some explaining for us to do ("yes, Slicer is really completely free of all obligations except..."). Also, many of the icons in OHIF are not very meaningful and heavily overloaded (the search icon is used as zoom, the next ">" icon is used as angle measurement, etc.).

@pieper do you think OHIF people are happy with their icons? Would they be interested in developing/maintaining a common icon set? Probably Slicer needs about 10x more icons, so they would get much more than they give (especially if we move away from fontawesome), but it would still require effort from them, mainly because they would probably need to revise some of their existing icons and it would take more coordination when adding or changing icons. You could also ask them about icon coloring - do they think it is a good idea to use two-color icons (as discussed above, base color with some modifier indicating function)?

> How widely used is the Google icon set? Is it something we can expect to be a quasi standard in many applications?

It is relatively new. Until recently people went with fontawesome and other large, quasi-free icon sets. Google's icon set changes things, as it is completely restriction-free. Its main disadvantage that it is highly web-focused. I guess that's why there is not even a standard file open/save icon (but only download/upload icon).

>> What colors should we use?
> base color and the modifiers thematized and grouped

This sounds good. Using more than one color can improve appearance and make the icons a bit more functional. Gray base color is not great if we want to use the same icons with both dark and light theme.

>> Light/dark mode
> possible removal of the light mode

Dropping light mode would make it much easier to make everything look optimal.

However, it would not be nice to lose the ability to customize color scheme for custom applications. I'm not sure though, if themes are the best way for customizing colors as theme colors are very limited compared to what you can do with custom style sheets (QSS) for that. @sjh26 do you know how QSS and themes are related? If someone uses QSS does that make themes obsolete?

jamesobutler:
> Its main disadvantage that it is highly web-focused. I guess that's why there is not even a standard file open/save icon (but only download/upload icon).
As mentioned in https://github.com/Slicer/Slicer/pull/5887#pullrequestreview-763078589, I do think there is a standard "Save" icon. 

image

| Load | Load Alt 2 (DICOM Import) | Save |
|------|--------------|------|
|image|image|image|

^ Something of the sort as replacement, though Load DICOM should really be merged with the regular Load button. It's confusing that to load DICOM you have to go into a different load area.

Also which Style?

|Outlined| Filled| Rounded| Sharp|Two-Tone|
|---------|-------|-----------|--------|-----------|
|image|image|image|image|image|

lassoan:
There is no _file_ open.

The folder open icon practically does not exist either, they just assigned the "folder open" name to the "folder" icon. They look exactly the same. It does not indicate "open" in any way. I don't think these are sufficient.

OK, yes, there is a "save" icon, but it is complete garbage. Sorry, I saw it but chose to ignore it. All the shapes are all off, the hole is not in the center, the designer thought that the rectangle at the top is a label, etc. I would not use this image.

How a 3.5" floppy disk looked like:

image

Visual Studio (and other software that I can quickly find on my computer) gets it right:

image

The DICOM module shows a DICOM browser. I don't think it will get merged with the file open function, because it has file open, import/export, network push/pull features. I agree that "Add file..." should redirect users to DICOM import, though. It will however not happen before Slicer5.

Overall, it is not a big deal that a number of standard icons are missing, especially if a consistent design is well established by hundreds of other icons, but still, it requires some extra effort, or incorporation of additional icon sets. I think getting icons from other sets (such as https://roundicons.com/native-line-solid-icons-pack/ for a one-time fee of $500 for all rights that Slicer developers and commercial partners may ever need) can be a good deal - but only if we don't plan to share the icons across many projects (e.g., sharing roundicons icons between Slicer and OHIF could be too much).

cpinter:
I think Ebatinca can offer resources to design some missing icons in SVG, especially those few that are prominent.

pieper:
> I think Ebatinca can offer resources to design some missing icons in SVG, especially those few that are prominent.

I like this idea.  Why don't we pick a small subset of icons, like the editor effects or the markups and see how hard it is for a designer to make a set that's consistent with the google icons.  I can also help support this with resources if needed.  If this approach works it'll be more consistent with our overall open approach and perhaps benefit the community more broadly.



> @pieper do you think OHIF people are happy with their icons? Would they be interested in developing/maintaining a common icon set? Probably Slicer needs about 10x more icons, so they would get much more than they give (especially if we move away from fontawesome), but it would still require effort from them, mainly because they would probably need to revise some of their existing icons and it would take more coordination when adding or changing icons. You could also ask them about icon coloring - do they think it is a good idea to use two-color icons (as discussed above, base color with some modifier indicating function)?

I started a thread on the OHIF community forum:
https://community.ohif.org/t/icon-set-and-theme/46

jcfr:
re: itkwidget 

After talking with @thewtex , he indicated switching switching the relevant build to the new "source" for icons wouldn't be a problem.
For reference, here is the current resource maintained for this project: https://github.com/InsightSoftwareConsortium/itk-viewer-icons

jcfr:
re: paraview

After talking with @cquammen 

> For some background, ParaView's icons have been undergoing modernization and updating over the last several releases. [ParaView] uses a large number of SVG icons now that were designed by @SteveJordanKW, and they have a mostly consistent style. [We] can view them here if interested and use them if [we] wish: https://gitlab.kitware.com/paraview/paraview/-/tree/master/Qt/Components/Resources/Icons 

And here are some more comments redacted for conciseness:
 
* Unifying icons across applications could have advantages if there was significant overlap in the user bases for the applications in question. Is there evidence that that is the case?

* Another question: how many common functions are there between applications? Loading and saving data, taking screenshots, maybe arrows for different purposes, undo/redo, etc. There are a few, but maybe not that many?

lassoan:
I had a look at this again while discussing a pull request (#6662).

@jcfr Thanks for the additional information on ParaView icons. I like the unique, recognizable style of these icons, but it would take much more effort to create new icons in this style. We could not expect average developers to create these kind of nice designs.

I've tried to create icons in Google material icons style, and I've found it was a very easy to get consistent-looking, expressive icons. I drew these in about an hour in Inkscape (these are 24px SVGs with lines snapped to integer positions and so they are rendered pixel-perfect):

MarkupsLine MarkupsFiducial MarkupsAngle MarkupsROI MarkupsPlane MarkupsClosedCurve MarkupsOpenCurve

We could draw these kind of monochrome icons pretty easily. Maybe we could add some color to indicate common themes, such as "add" could be always green (e.g., green + on the markup icon would indicate that a new markup will be created), delete could be always red, etc.

**What do you all think? Is there any objection against replacing all the icons with simple line drawing icons similar to Google material icons?**

jamesobutler:
> I drew these in about an hour in Inkscape (these are 24px SVGs with lines snapped to integer positions and so they are rendered pixel-perfect):

If you could document your process for this that would be great (I'm not sure where to put this info). Then others including myself could learn to create new icons in this same method/style moving forward. When a new feature needs a new icon we could say "See X document" to create an acceptable SVG icon that would be theme compatible.

Punzo:
> What do you all think? Is there any objection against replacing all the icons in Slicer with simple line drawing icons similar to Google material icons?

It would be great to have a material-like theme for the Slicer icons

> I've tried to create icons in Google material icons style, and I've found it was a very easy to get consistent-looking, expressive icons. I drew these in about an hour in Inkscape (these are 24px SVGs with lines snapped to integer positions and so they are rendered pixel-perfect):
> 
> MarkupsLine MarkupsFiducial MarkupsAngle MarkupsROI MarkupsPlane MarkupsClosedCurve MarkupsOpenCurve

is the color in rgb = [241,195,54] ? I picked it, but just wanted to be sure.

Punzo:
> I used the default: image
> 
> image

Experienced that it would be better to get the icons in 24 px for the optical size, since Qt will not shrink them properly, while the grow works well.

Punzo:
after experimenting a little bit, for coloring the material icons a good approach can be:

1) get material icon with this parameters:
image

2) in inkscape:
   A) Create a new layer (menu Layer -> addLayer or Shift+Ctrl+N)
   B) select the layer on the menu in the right       
image
   C) fill the icon (shortcut U)  with Threshold = 100, grow/shrink= 0.01 px, close gaps = none, color rgb = [241,195,54]
image
   D) set visibility of the new layer to off 
   E) delete the black icon
   F) set visibility of the new layer to on
   G) save

jtomeck:
Hi all! For those who don't know me, I work at Kitware with @SteveJordanKW and @jcfr and I am part of the UI/UX team.

@jcfr shared these with me and I think the icons look quite nice in the material style! I love the idea of creating a full material icon set for Slicer for consistency. My only critiques are surrounding color. Yellow text / icons on a white background are often hard to read, so my suggestion is to use a dark gray for the icons in light mode and a light gray in dark mode, where the only color on the icon is just an accent if needed / possible (ie. you could have the shape icons have a translucent colored face or faces). Other than that, nice work!

cpinter:
Once the guidelines are laid down, I'm happy to help with redesigning some of the icons, especially those that I originally made (there may be a dozen or so), but others too if we want to crowdsource.

wendples:
Greetngs all -- I've been coming up with discussion points and tinkering with material symbols icons, and icons derived from that set. Going to post excerpts from my working notes for future discussion. Sets of posts to follow with some screenshots -- startng with some topics for discussion to inform the overall design approach we might take...

Screenshot 2023-09-25 at 2 25 49 PM

wendples:
Screenshot 2023-09-25 at 2 38 39 PM

wendples:
Definitely worth looking at Material Symbol Icons for deriving our own set; we benefit from a solid design process, and it has turned out to be really quick and easy to work from downloaded svgs. Since this is a **variable font**, icons of a particular optical size are styled to look best at that each optical size. Rendering that icon at a different resolution than its optical size will generate a different look ( prob not as handsome) as using an icon at the optical size matching that resolution. No strokes – even linear elements are defined as filled shapes. Linear thickness changes with weight, grade and optical size. So choosing those  settings well is important.  An example: 
Screenshot 2023-09-25 at 2 44 31 PM

wendples:
After a bit of comparing the various combinations,  I'm liking the grade=0, weight=200, optical size=24 configuration. An example:
Screenshot 2023-09-25 at 2 45 52 PM

wendples:
For instance, here's one table of variations... weight 200, grade 0 have good readability, and low glare in dark mode. Definitely open for discussion...
Screenshot 2023-09-25 at 2 48 28 PM

wendples:
Trying out some of the icons Andras used in his example, and those james pointed out (load versus upload, etc)
Screenshot 2023-09-25 at 3 24 37 PM

wendples:
Some mockups to see how the character of these feel with Slicer...

Screenshot 2023-09-25 at 3 34 00 PM

wendples:
Screenshot 2023-09-25 at 3 34 57 PM

jamesobutler:
We should probably separately consider UX changes such as reorganizing the toolbuttons in the Module Selector toolbar. I don't really know what the "Module Select" toolbutton would do in place of the "Modules: " text label next to the module combobox. 

From a UX perspective, I think the module select toolbar is flawed being a QToolBar because ToolBars are flexible objects that can be customized to be shown/hidden, moved around to reorder or moved to different toolbars on top/right/bottom/left. The Module Selector and associated history toolbuttons are actually Primary Navigation elements that should be consistently in the same location and not editable by the user. It is primary navigation as it navigates to specific pages in the module panel dock widget which is critical for using the various functionalities of Slicer.

wendples:
I agree with this James! This goes to the bigger discussion about how much to change the application interface -- with big changes come the need to update a lot of training materials as well. So worth discussing. I'm going to post a few mockups from my notes. Then will take a pause for more discussion on these things.

Screenshot 2023-09-25 at 3 36 43 PM

Screenshot 2023-09-25 at 3 36 12 PM

Screenshot 2023-09-25 at 3 36 19 PM


Screenshot 2023-09-25 at 3 47 20 PM

wendples:
For your review - some screenshots at bottom of light and dark mode slicer. i hope we can discuss at a future dev meeting that's low on urgent topics!

Screenshots feature:
- new main toolbar icons
- new view configuration icons

link to the working notebook: https://docs.google.com/document/d/1OYhRzBFjwT6dUOIDVL_II8ZQ8QUwDl68wbtt3eIV1ao/edit?usp=sharing

In particular, I'd like to know about color variations and what you think is important:

- particularly dark mode -- do you prefer more contrast for use of dark mode in dark environment?
- combination of stroke and flat color ok -- more, less, more limited palette or zero color?
- mouse mode icons: can we use simpler symbols to represent complicated sets of options to select from?
- moving toward simpler icons: preferences from board1 and board2 .

So much fun!
LightSlicerMockup-2024-01-23
DarkSlicerMockup-2024-01-23

jamesobutler:
👏🏻 great work!
image
These board 2 designs shown above are definitely interesting as I assume they are to be inspired by the 3D Slicer logo. Although interesting for 3D Slicer, for all the 3D Slicer custom applications with their own branding, it would be nice to not have to replace icons as well. Those 3D Slicer logo inspired icons may be considered too much 3D Slicer branding for a custom application. The board 1 designs are generic enough to work in both 3D Slicer and for 3D Slicer custom applications.

jcfr:
Looks like it may be sensible to integrate a subset of these changes before closing this pull request.

PR #5890, opened 2021-09-24
BUG: Do not hide SH combobox popup when selected item is modified

cpinter:
The popup was hidden both when selection was changed and when selected item was modified. Now it is not hidden when the item is modified, because it prevented selection when streaming data into Slicer. In that case if the proxy node was selected and it was modified because of the new data item appearing in the corresponding sequence, the popup was collapsed, and if the stream was fast, it could disappear in a fraction of a second.


lassoan (approved):
Looks good to me.

cpinter:
Thank you!

PR #5891, opened 2021-09-24
ENH: Add object name to center button in 3D view header

cpinter:
An object name facilitates unambiguous access for the widget.

The name "CenterButton_Header" was chosen in order to differ from the "CenterButton" in the popup of the same toolbar.


lassoan (approved):
LGTM

cpinter:
Thanks Andras!

Issue #5889, opened 2021-09-24
Markups control point placement improvements


lassoan:
This is a follow-up to pull request https://github.com/Slicer/Slicer/pull/5325. It contains fixes and improvements that were identified during the review but has not been completed before the integration:

- [ ] When clicking the control point table to edit a point then the point position is filled immediately. This position gets deleted if I move over to a view and leave the view and the edit icon disappears from the table. It would be better if starting editing would just change the icon to edit (would not show the coordinates) and after going over a view and coming back, the edit icon would stay displayed in the table while placement mode is enabled. This would make it clear for the user which point will be the next to place (without need to hover over any view) and would ensure that the table state would not be changed just because the user temporarily moves the mouse over a view.
- [ ]  vtkMRMLColorTableNodeRandom color table is not designed for creating nice, distinguishable colors. Instead, use MediumChartColors, DarkBrightChartColors, LightPaleChartColors. Make the color index start at 0 (not at 3) and wrap around if the color index reaches the number of values in the color table. Remove UpdateCurrentColorTableIndex (fuse it into GetNewMarkupsColor). Make GetNewMarkupsColor return a double* (you need to add a `double LastReturnedMarkupsColor[3];` in the class to store the vector that is returned) and add `VTK_SIZEHINT(3)` to make the method conveniently usable from Python.
- [ ] It is inconsistent that ROIs have all the same color if created one by one but have different colors if "Place multiple control points" is enabled.
- [ ] Lot of "child widget is not accessible warnings" when selecting a control point in the table
- [ ] The down-arrow button on the place widget's place button should be hidden if the force single/force multiple option is chosen. 
- [ ] Fix failing toolbar test: qSlicerMouseModeToolBarTest1
- [ ] Update documentation (https://slicer.readthedocs.io/en/latest/user_guide/modules/markups.html#)
- [x] *Display* markups toolbar -> *Toggle* markups toolbar (because the same menu is used for showing/hiding)
- [x] Node selector and color selector in the toolbar must have the same height as the toolbar buttons
- [x] Add a spacer between the new markup buttons and the node selector button

muratmaga:
There is a bug that the locked (template) state of the fiducial node is not retained after save, when loading it into a new scene. 

[...]

PR #5892, opened 2021-09-24
BUG: Corrects fixed point number property on loading markups

smrolfe:
This commit fixes a bug causing the fixed point number property to not be read while loading a markup node from file. This bug is documented in issue #5889


smrolfe:
I have added fixes for the following three issues also documented in issue #5889:
 1. Display markups toolbar -> Toggle markups toolbar (because the same menu is used for showing/hiding)
 2. Node selector and color selector in the toolbar must have the same height as the toolbar buttons
 3. Add a spacer between the new markup buttons and the node selector button

lassoan (approved):
Thank you, it looks good!

PR #5895, opened 2021-09-25
COMP: Ensure stdlib and site-packages python modules are always byte-compiled

jcfr:
This ensures that running Slicer from a read-only location will load the
byte-compiled modules.


jcfr:
Tested locally on Linux, macOS and Windows

PR #5899, opened 2021-09-25
Markups Toolbar build error and Annotations place fixes

jamesobutler:
This fixes a build error I ran into. I was building Slicer latest (cfc69c5fcd8e119dcf3060561c999c8a9baebd93).

> 33>C:\Slicer\Modules\Loadable\Markups\Widgets\qMRMLMarkupsToolBar.cxx(30,10): fatal error C1083: Cannot open include file: 'QTest': No such file or directory

I also noticed some incorrect tooltips when placing Annotation ruler and Annotation ROI node. 

Also have made slight layout improvement so annotations place buttons are compressed together.
| Current | This PR |
|----------|---------|
|image|image|

cc: @lassoan, @smrolfe

### DIFF
--- a/Modules/Loadable/Annotations/Resources/UI/qSlicerAnnotationModuleWidget.ui
+++ b/Modules/Loadable/Annotations/Resources/UI/qSlicerAnnotationModuleWidget.ui
@@ -48,7 +48,7 @@
              <string>Create-and-place</string>
            </property>
            <layout class="QGridLayout" name="createLayout">
-             <item row="0" column="1" >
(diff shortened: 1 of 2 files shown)


jamesobutler:
@lassoan Can you give this a review so it can be merged before the nightly build begins? I wouldn't want another build error to mess up the nightly build.

lassoan (approved):
Thank you, it all looks good to me. I did not see any build error without the "QTest" removal, but it is not necessary there, so it is good to remove it anyway.

PR #5901, opened 2021-09-26
ENH: Create markups node put into menu for toolbar

jamesobutler:
This addresses https://github.com/Slicer/Slicer/issues/5889#issuecomment-927342225 to put the create markup node actions in the Markups Toolbar within a menu of a QToolButton. Markups is a pluggable framework where more markups can be registered and added into Slicer and creating a new button in the toolbar for each registered markup is not sustainable as it will only make the Markups toolbar larger.

|Current | This PR |
|---------|----------|
| image| image|


lassoan:
The design is still discussed here: https://github.com/Slicer/Slicer/issues/5889#issuecomment-927368089

jamesobutler:
Updated image
image

jamesobutler:
Closing PR as proposed changes do appear that they will be accepted and I'm not proceeding to resolve merge conflicts.

PR #5902, opened 2021-09-27
ENH: Make it clearer the meaning of the markups buttons

jamesobutler:
The create markups buttons in the Markups module didn't have any tooltips or text to describe what they did. The Markups toolbar at least has tooltips.

Registered markups are now given a tooltip and button text for their create buttons in the Markups module. This makes it easier for users to learn what they are and be able to use the Markups Toolbar which does not have text.

| Current | This PR |
|----------|---------|
|image|image|


jamesobutler:
Note that I added button text to these markup create buttons based on the amount of current markups in Slicer core. Similar to Segment Editor rework recently, we could consider just having tooltips when there is a really large group of buttons. At minimum the added tooltips will be helpful here.

lassoan:
I like how these buttons with the labels look very much. Now that Segment Editor don't use labels anymore, it would be kind of an outlier (other toolbars don't show labels) and it consumes more vertical space (but it can be scrolled away). Let's see what others think.

jamesobutler:
One of the benefits when you click on Segment Editor effect is that there is an options panel that reveals the name of the effect that is not hidden in a tooltip. With the Create Markups buttons in the Markups module, I don’t think there’s anywhere else in the GUI that defines the name of the markup. It’s currently only in tool tips or node class.

jcfr:
Some application (e.g gmail) have a density option:

image

Wondering if showing/hiding text near button could be part of customization step after user first install the application ?

jcfr:
To answer the specific question, in Slicer app, I have no issue having the text  near the button.  It is indeed a nice way for the user to learn the meaning of each icons.

And we could still have a programmatic way to show/hide them easily for sake of customization.

jamesobutler:
image

Rebasing this PR against latest to see if it resolve the issue of the docs staying stuck in "Expected - Waiting for status to be reported"? @jcfr Any ideas why it is getting stuck?

jcfr:
>  Any ideas why it is getting stuck?

@jamesobutler I suspect this was related to https://github.com/Slicer/Slicer/issues/5958 that has just been addressed.

PR #5903, opened 2021-09-27
BUG: wrong `tbb` subfolder name on Linux

kerim371:
Slicer uses prebuilt `tbb` binaries and I noticed that Slicer uses wrong subfolder name when setting `tbb_libdir` cmake variable.
You can simply check that by cloning the `tbb` release used by Slicer and see folder tree


lassoan (approved):
Thank you for the fix. Indeed the subfolder name in https://github.com/oneapi-src/oneTBB/releases/download/2019_U9/tbb2019_20191006oss_lin.tgz package is indeed gcc4.8.

PR #5904, opened 2021-09-27
ENH: check DICOM instance before stopping listener

rbahegne:
Using a slicelet with a custom DICOM browser and closing the app with this DICOM browser open triggers a python error caused by the dicom listener stopped as it does not have any DICOMInstance.
This simple check fix this error.

### DIFF
--- a/Modules/Scripted/DICOM/DICOM.py
+++ b/Modules/Scripted/DICOM/DICOM.py
@@ -803,7 +803,8 @@ class DICOMWidget(ScriptedLoadableModuleWidget):
       wasBlocked = self.ui.toggleListener.blockSignals(True)
       self.ui.toggleListener.checked = False
       self.ui.toggleListener.blockSignals(wasBlocked)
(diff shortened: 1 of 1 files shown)


Modules/Scripted/DICOM/DICOM.py
```diff
@@ -803,7 +803,8 @@ def onListenerStateChanged(self, newState=None):
       wasBlocked = self.ui.toggleListener.blockSignals(True)
       self.ui.toggleListener.checked = False
       self.ui.toggleListener.blockSignals(wasBlocked)
-      slicer.modules.DICOMInstance.stopListener()
+      if hasattr(slicer.modules, 'DICOMInstance'):
```
lassoan, line 806:
```suggestion
      if hasattr(slicer.modules, 'DICOMInstance'):  # custom applications may not have the standard DICOM module
```

lassoan (approved):
Thank you, it makes sense.

PR #5905, opened 2021-09-27
ENH: Remove vtkMRMLMarkupsROINode::GetBoundsROI() and add documentation

Sunderlandkyl:
The GetBoundsROI function was confusingly named and did not do what users expected (returning the bounds of the ROI in node coordinates).

This commit removes the GetBoundsROI function, and adds additional documentation to explain which functions can be used to determine the ROI axis-aligned bounding box and orientated bounding box. Since the function was only used in one location, it should be safe to remove.

### DIFF
--- a/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsROINode.h
+++ b/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsROINode.h
@@ -128,9 +128,6 @@ public:
     return this->ObjectToWorldMatrix;
     };
 
-  /// Get the bounds of the ROI in the ROI coordinate system.
-  void GetBoundsROI(double bounds[6]);
-
   /// ROIType represents the method that is used to calculate the size of the ROI.
   /// BOX ROI does not require control points to define a region, while the size of a BOUNDING_BOX ROI will be defined by the control points.
   vtkGetMacro(ROIType, int);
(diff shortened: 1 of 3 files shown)


Modules/Loadable/Markups/MRML/vtkMRMLMarkupsROINode.h
```diff
@@ -168,15 +164,24 @@ class  VTK_SLICER_MARKUPS_MODULE_MRML_EXPORT vtkMRMLMarkupsROINode : public vtkM
...
   vtkMRMLStorageNode* CreateDefaultStorageNode() override;
 
-  /// Create default storage node or nullptr if does not have one
+  /// Create default display node or nullptr if does not have one
   void CreateDefaultDisplayNodes() override;
 
+  /// Reimplemented to recalculate the axis-aligned bounds of the ROI.
+  /// If the ROI is rotated, this function will not reflect the oriented bounds defined by the ROI.
+  /// To get the planes that define the oriented bounding box, use GetPlanes()/GetPlanesWorld().
+  /// \sa GetPlanes(), GetPlanesWorld()
```
lassoan, line 173: Could you describe the difference between the two methods?

lassoan (commented):
Thank you, it looks good, I just added one small comment.

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsROINode.h
Sunderlandkyl: Ok, done!

PR #5906, opened 2021-09-28
BUG: Make property slots public in SH tree view

cpinter:
The setter slots for the attribute properties in the Subject Hierarchy tree view were protected by mistake, which caused a build error when setting any of them in Qt designer, and in general not usable for the tree view.


lassoan (approved):
thanks, it looks good to me

PR #5910, opened 2021-09-29
ENH: Add mouse gesture to finish markup point placement by double-click

lassoan:
Double-click is used in many software (PowerPoint, Inkscape, Gimp, ...) to finish point editing. It also makes double-click a more useful action during point placement (previously, a double-click resulted in placing a control point and maximizing/restoring the view).

See discussion here: https://discourse.slicer.org/t/new-feature-maximize-view/19581


jamesobutler (approved):
Yes, this is better thanks.

PR #5919, opened 2021-09-29
ENH: Reorder and rename control point position status options for clarity

jamesobutler:
Originally from discussion starting at https://discourse.slicer.org/t/changes-to-the-markups-module/19871/3?u=jamesobutler

"Clear" is chosen over "Unset" since it is less technical. Action order is based on the cycle order when click in the position status column of the markups control point table.

| Current | This PR |
|----------|---------|
| image| image|
|image|image|

Also including minor commit `STYLE: Use control point terminology in context menu` which should been included in 1669e8c94d54d6763e381e837ae90113ed31deb7 as it relates to consistently using the "control point" terminology rather than "point" or some other incorrect uses such as "fiducial" or "markup".

| Current | This PR |
|----------|---------|
|image|image|

`STYLE: Use correct control point terminology in dialog`
| Current | This PR |
|----------|---------|
|image|image|

### DIFF
--- a/Modules/Loadable/Markups/qSlicerMarkupsModuleWidget.cxx
+++ b/Modules/Loadable/Markups/qSlicerMarkupsModuleWidget.cxx
@@ -480,7 +480,11 @@ void qSlicerMarkupsModuleWidgetPrivate::setupUi(qSlicerWidget* widget)
   QTableWidgetItem *positionHeader = this->activeMarkupTableWidget->horizontalHeaderItem(this->columnIndex("Position"));
   positionHeader->setText("");
   positionHeader->setIcon(QIcon(":/Icons/Large/MarkupsPositionStatus.png"));
-  positionHeader->setToolTip(QString("Click in this column to set/unset control points."));
+  positionHeader->setToolTip(QString("Click in this column to modify the control point position state.\n\n"
+                                     "- Edit: Enter place mode to modify the control point position in the slice views\n"
+                                     "- Skip: 'Place multiple control points' mode skips over the control point entry\n"
+                                     "- Restore: Set the control point position to its last known set position\n"
+                                     "- Clear: Clear the defined control point position, but do not delete the control point"));
   this->activeMarkupTableWidget->setColumnWidth(this->columnIndex("Position"), 10);
 
   // listen for changes so can update mrml node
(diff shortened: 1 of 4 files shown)


Modules/Loadable/Markups/qSlicerMarkupsModuleWidget.cxx
```diff
@@ -480,7 +480,7 @@ void qSlicerMarkupsModuleWidgetPrivate::setupUi(qSlicerWidget* widget)
   QTableWidgetItem *positionHeader = this->activeMarkupTableWidget->horizontalHeaderItem(this->columnIndex("Position"));
   positionHeader->setText("");
   positionHeader->setIcon(QIcon(":/Icons/Large/MarkupsPositionStatus.png"));
-  positionHeader->setToolTip(QString("Click in this column to set/unset control points."));
+  positionHeader->setToolTip(QString("Click in this column to define the position status of the control points."));
```
lassoan, line 483: "define the position status of the control points" sounds really complicated. I don't think many users could make sense of it. Do you have an idea how to simplify? Maybe something like "modify control point state"? Maybe we could add a bullet-point list in the tooltip (we do it at a few places), explaining each state:

"
- edit: ...
- undefined: ...
- missing: ...
- ...
"

Modules/Loadable/Markups/qSlicerMarkupsModuleWidget.cxx
```diff
@@ -2257,14 +2257,9 @@ void qSlicerMarkupsModuleWidget::onRightClickActiveMarkupTableWidget(QPoint pos)
...
       this, SLOT(onResetMarkupPushButtonClicked()));
-  QAction* unsetFiducialAction =
-      new QAction(QString("Unset position of highlighted fiducial(s)"), &menu);
-  menu.addAction(unsetFiducialAction);
-  QObject::connect(unsetFiducialAction, SIGNAL(triggered()),
-      this, SLOT(onUnsetMarkupPushButtonClicked()));
 
   QAction* missingFiducialAction =
-      new QAction(QString("Set position missing for highlighted fiducial(s)"), &menu);
+      new QAction(QString("Skip position for highlighted fiducial(s)"), &menu);
```
lassoan, line 2262: "Skip position" is not very clear. Maybe "mark as missing"?

jamesobutler: "Mark as position is missing" isn't very clear in how that is different from a cleared position. "Skip" defines what functionally is going to happen with this control point in placement mode, however just stating "Missing" doesn't indicate how it will be used any differently from "Clear".

lassoan: Using a predefined set of named landmarks (template) and explicitly marking some of them as not available is a feature that I saw in several of our projects. The need is not related to minor implementation details such as indexing, but it is about differentiating between "position is undefined" and "position is defined to be unknown". You may of want to skip points for various reasons (not necessarily because it is missing), so "skip" may be more accurate than "missing". I don't have a problem with using the "skip" word. What I find confusing is that you don't skip the position, but you skip _definition_ of the position, or _placement_ of the position, or exclude the point from placement.

lassoan: I can image that deleting a point can be acceptable in certain workflows. However, this approach has two important shortcomings:
1. If you delete the point then there is no information on why it was removed.
2. Removing a point is an irreversible operation. When landmarking result is reviewed, point positions may be modified, which includes changing, setting, unsetting positions. It is possible to implement restoring of skipped landmarks in each landmarking module, but since this need is such a common need, it makes sense to implement it only once, cleanly, in Slicer core.

jamesobutler: So you may set it to be skipped, but then later decide that it shouldn't be skipped?

jamesobutler: Trying to think of clear wording for the action when dealing with the control point lists of line nodes and angle nodes.

Modules/Loadable/Markups/qSlicerMarkupsModuleWidget.cxx
jamesobutler: image

Modules/Loadable/Markups/qSlicerMarkupsModuleWidget.cxx
jamesobutler: I've updated to "Skip placement..."
image

lassoan: I like "Skip placement".

I also like the icons for Cut/Copy/Paste actions. What do you think about adding the state icons there so that the actions can be recognized more easily?

jamesobutler: Good idea! Done.
image

lassoan (approved):
Thanks for the updates, all looks good.

PR #5922, opened 2021-09-30
Remove support for uploading extensions to midas

jcfr:


jcfr:
Relevant changes have been backported to branch [v4.11](https://github.com/Slicer/Slicer/commits/v4.11) and tested for few days already.

PR #5924, opened 2021-09-30
BUG: Fix crash in Markups storage node if setting point beyond maximum

Sunderlandkyl:
Calling vtkMRMLMarkupsFiducialStorageNode::SetPointFromString() with an index beyond the maximum number of control points would result in a crash. To reproduce, place a line and then copy+paste a control point using the Markups module widget.

Fixed by checking that the specified index is under the maximum, and that the control point exists before attempting to access it.


lassoan (approved):
Looks good, thank you.

Issue #5915, opened 2021-09-29
Markups: Mouse mode is no longer reflected by mouse cursor shape


lassoan:
Anyone else noticing that the WindowLevel mouse mode cursor icon is no longer present while in that mouse mode? I can adjust window level while in the mode, but the correct cursor icon is missing.

A potentially related issue that also must be fixed is the failing toolbar test: qSlicerMouseModeToolBarTest1.

## Environment
- Slicer version: Slicer-4.11.0-2021-09-27
- Operating system: Windows

@smrolfe

smrolfe:
Thanks, I'm working on this issue now.

jamesobutler:
Closed by dd8d8a5650af40601cd85091f23c3f66b86b8ccc.

PR #5927, opened 2021-10-01
BUG: Restores mouse mode cursor shape

smrolfe:
This commit addresses issue [#5915](https://github.com/Slicer/Slicer/issues/5915). Cursor shape is now set from the selection node active place node class instead of the mouse mode toolbar actions removed in  09d12ae6f7598fe8490b691d17a35dcca5bd344a. qSlicerMouseModeToolBarTest1 is updated to remove tests of interaction between the selection node and removed mouse mode toolbar actions.

### DIFF
--- a/Base/QTGUI/Testing/Cxx/qSlicerMouseModeToolBarTest1.cxx
+++ b/Base/QTGUI/Testing/Cxx/qSlicerMouseModeToolBarTest1.cxx
@@ -9,6 +9,7 @@
 #include "qSlicerMouseModeToolBar.h"
 
 // MRML includes
+#include <vtkMRMLCoreTestingMacros.h>
 #include <vtkMRMLInteractionNode.h>
 #include <vtkMRMLScene.h>
 #include <vtkMRMLSelectionNode.h>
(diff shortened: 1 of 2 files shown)


Base/QTGUI/qSlicerMouseModeToolBar.cxx
```diff
@@ -383,9 +383,44 @@ void qSlicerMouseModeToolBarPrivate::updateCursor()
   int currentInteractionMode = interactionNode->GetCurrentInteractionMode();
   if (currentInteractionMode != vtkMRMLInteractionNode::Place)
     {
-    interactionNode->SetCurrentInteractionMode(vtkMRMLInteractionNode::Place);
+    if (interactionNode->GetCurrentInteractionMode() == vtkMRMLInteractionNode::ViewTransform)
+      {
+      q->changeCursorTo(QCursor());
```
lassoan, line 388: How is it possible that we add all this code for setting the cursor here and we do not remove from anywhere else? Where was it originally implemented? Can that code be removed?

lassoan (commented):
Thanks a lot for working on this, I added two questions/comments.

Base/QTGUI/qSlicerMouseModeToolBar.cxx
smrolfe: This change restores the code to update the cursor shape where previously implemented, with a few changes. I removed it while testing since it no longer worked when the actions menu was removed from the mouse mode toolbar. With the markup node displayed in the toolbar, it did not seem as important to provide the visual cue to the placement node type with the cursor, so I did not add it back immediately. I'm still not sure it's necessary for control point placement, but agree that its needed for the WindowLevel action 

In this commit, I've updated the original code to update the cursor shape from the selection node instead of the mouse mode place actions menu.

Base/QTGUI/Testing/Cxx/qSlicerMouseModeToolBarTest1.cxx
```diff
@@ -60,67 +75,11 @@ int qSlicerMouseModeToolBarTest1(int argc, char * argv[] )
...
   QString activeActionText;
   activeActionText = activePlaceActionText(mouseToolBar);
-  std::cout << "Active action text = " << qPrintable(activeActionText) << std::endl;
+  std::cout << "Active place action text = " << qPrintable(activeActionText) << std::endl;
 
-  // get the selection and interaction nodes that the mouse mode tool bar
+  // get interaction nodes that the mouse mode tool bar
   // listens to
   vtkMRMLNode *mrmlNode;
-  vtkMRMLSelectionNode *selectionNode = nullptr;
```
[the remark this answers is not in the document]
smrolfe: The mouse mode toolbar test has been updated to check for the default button state (toggle the Markups button), and default cursor (translate/rotate). I can add the a check for the WindowLevel mode. Testing the placement mode will require adding markup nodes to the scene, which I think would be better added as a test of the new Markups Toolbar, does this seem correct?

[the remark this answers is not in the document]
lassoan: It would be better to test this toolbar (including placement mode) independently from the markups toolbar. Just adding nodes to the scene and modifying the selection node should update the toolbar's state correctly.

Base/QTGUI/qSlicerMouseModeToolBar.cxx
lassoan: OK, thanks for the explanation.

lassoan (approved):
I've improved the test a bit and tested everything and it all looks good.

PR #5928, opened 2021-10-04
ENH: Make segment effect cursor hotspot position clearer

jamesobutler:
Originally from https://discourse.slicer.org/t/user-confusion-about-mouse-position-due-to-segment-editor-effect-cursor-artwork/19978. Issuing this PR now so others can try it out 

The "Draw" effect cursor was often mistaken for having the hotspot positioned at the end of the Pencil rather than the end of the blue arrow in the cursor.  The blue arrow has been replaced with the standard white arrow cursor arrow and a white border has been added around the effect image to make it clearer that the hotspot position is not in the effect image.

| Current| This PR |
|---------|---------|
|image|image|


jamesobutler:
With this change it appears that CursorBaseArrow.png is no longer used. Not sure if there would be a reason to keep it around? Looks like other icons adopted it, but don't use the base image to form a new image in the code. Instead, a completely new image was created such as MouseWindowLevelMode.png which includes a blue arrow cursor.
https://github.com/Slicer/Slicer/blob/863fbd58be5ead083055e64203371e7f2fedaca4/Base/QTGUI/Resources/Icons/MouseWindowLevelMode.png

jamesobutler:
@cpinter as suggested by JC, can you provide your thoughts here about the changes here?

if no objections, I’ll proceed with merging later today.

PR #5930, opened 2021-10-04
BUG: Fix control point copy paste with 0 maximum control points

Sunderlandkyl:
Fixes a bug introduced by the fix in 1b21b7208594f3b6c1ed387ae648c3e9d757f1d0.
Pasting a control point from the Markups module widget would be prevented in markups nodes that had MaximumNumberOfControlPoints of 0.
Fixed by also checking that MaximumNumberOfControlPoints is greater than 0 before returning.

### DIFF
--- a/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsFiducialNode.cxx
+++ b/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsFiducialNode.cxx
@@ -53,7 +53,7 @@ void vtkMRMLMarkupsFiducialNode::WriteXML(ostream& of, int nIndent)
 //----------------------------------------------------------------------------
 void vtkMRMLMarkupsFiducialNode::ReadXMLAttributes(const char** atts)
 {
-  int disabledModify = this->StartModify();
+  MRMLNodeModifyBlocker blocker(this);
 
   Superclass::ReadXMLAttributes(atts);
 
@@ -62,7 +62,12 @@ void vtkMRMLMarkupsFiducialNode::ReadXMLAttributes(const char** atts)
   vtkMRMLReadXMLIntMacro(requiredNumberOfControlPoints, RequiredNumberOfControlPoints);
   vtkMRMLReadXMLEndMacro();
 
-  this->EndModify(disabledModify);
+  // In scenes created by Slicer version version 4.13.0 revision 30287 (built 2021 - 10 - 05).
+  // The value used to represent unlimited control points has been changed to -1.
+  if (this->MaximumNumberOfControlPoints == 0)
+    {
+    this->MaximumNumberOfControlPoints = -1;
+    }
 }
 
 //----------------------------------------------------------------------------
--- a/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsFiducialNode.cxx
+++ b/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsFiducialNode.cxx
@@ -62,7 +62,7 @@ void vtkMRMLMarkupsFiducialNode::ReadXMLAttributes(const char** atts)
   vtkMRMLReadXMLIntMacro(requiredNumberOfControlPoints, RequiredNumberOfControlPoints);
   vtkMRMLReadXMLEndMacro();
 
-  // In scenes created by Slicer version version 4.13.0 revision 30287 (built 2021 - 10 - 05).
+  // In scenes created by Slicer version version 4.13.0 revision 30287 (built 2021-10-05).
   // The value used to represent unlimited control points has been changed to -1.
   if (this->MaximumNumberOfControlPoints == 0)
     {
--- a/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsFiducialStorageNode.cxx
+++ b/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsFiducialStorageNode.cxx
@@ -304,7 +304,7 @@ bool vtkMRMLMarkupsFiducialStorageNode::SetPointFromString(vtkMRMLMarkupsNode *m
     return false;
     }
 
-  if (pointIndex >= markupsNode->GetMaximumNumberOfControlPoints())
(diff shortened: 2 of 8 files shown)


lassoan (commented):
Good catch, thanks for fixing it.

Could you review all usages of MaximumNumberOfControlPoints that its validity is checked before it is used?

Also, This >0 check is correct, as by default the maximum number is set to 0, but I think we should change the default to -1 and accept all values >=0. This is unrelated to your fix, but it would be a good opportunity to fix it now that we touch related code; and I think it would reduce chances of errors due to incorrect API usage in the future. Let me know what you think - if you find this too much work or too risky or don't have time for it then we could reconsider.

Sunderlandkyl:
Sure, I'll take a look at it. It looks like it is only used in 4 Slicer core classes.

Sunderlandkyl:
What would we do about Markups nodes in existing scenes that were saved with MaximumNumberOfControlPoints 0?

lassoan:
> What would we do about Markups nodes in existing scenes that were saved with MaximumNumberOfControlPoints 0?

Is MaximumNumberOfControlPoints saved into the scene or in the mrk.json file?

Sunderlandkyl:
It's saved to the scene only for Fiducial nodes here: https://github.com/Slicer/Slicer/blob/9ed089072c41c17f3de32615a8dd7ce508fc7813/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsFiducialNode.cxx#L48

lassoan:
We could convert the 0 value to -1 in vtkMRMLMarkupsFiducialNode::ReadXMLAttributes. However, it would be nice if we would overwrite the value only in scenes that were created with earlier Slicer versions. We could do [this](https://github.com/Slicer/Slicer/blob/9ed089072c41c17f3de32615a8dd7ce508fc7813/Libs/MRML/Core/vtkMRMLFiducialListStorageNode.cxx#L384-L391). The only issue could be that I think the [scene is currently only set for storage nodes](https://github.com/Slicer/Slicer/blob/9ed089072c41c17f3de32615a8dd7ce508fc7813/Libs/MRML/Core/vtkMRMLParser.cxx#L143-L148). I don't know why is that and I think it should be no problem to enable this for every node, and that would help with parsing old scenes correctly. It would be great if you could set the scene for all nodes and run Slicer tests to see if any issue comes up. If there are issues or you don't have time to test it then we can just keep changing 0 to -1 for all scene versions (and describe in a comment why we do it).

Sunderlandkyl:
Ok, I've made it so that for now we change 0 to -1 for all scene versions. When I have time to do the proper testing, I'll update it so that we only change the value for old scenes.

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsFiducialStorageNode.cxx
```diff
@@ -304,7 +304,7 @@ bool vtkMRMLMarkupsFiducialStorageNode::SetPointFromString(vtkMRMLMarkupsNode *m
     return false;
     }
 
-  if (pointIndex >= markupsNode->GetMaximumNumberOfControlPoints())
+  if (markupsNode->GetMaximumNumberOfControlPoints() > 0 && pointIndex >= markupsNode->GetMaximumNumberOfControlPoints())
```
lassoan, line 307: Shouldn't we check for >=0 ?
```suggestion
  if (markupsNode->GetMaximumNumberOfControlPoints() >= 0 && pointIndex >= markupsNode->GetMaximumNumberOfControlPoints())
```

Modules/Loadable/Markups/qSlicerMarkupsModuleWidget.cxx
```diff
@@ -1477,7 +1477,7 @@ void qSlicerMarkupsModuleWidget::onAddMarkupPushButtonClicked()
     return;
     }
   // get the active node
-  bool definedStatus = d->MarkupsNode->GetMaximumNumberOfControlPoints() != 0;
+  bool definedStatus = d->MarkupsNode->GetMaximumNumberOfControlPoints() >= 0;
```
lassoan, line 1480: This `definedStatus` name is not relevant here (probably just some copy-paste remnant). Could you change it to the simple check that we do elsewhere?

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsFiducialNode.cxx
```diff
@@ -62,7 +62,12 @@ void vtkMRMLMarkupsFiducialNode::ReadXMLAttributes(const char** atts)
   vtkMRMLReadXMLIntMacro(requiredNumberOfControlPoints, RequiredNumberOfControlPoints);
   vtkMRMLReadXMLEndMacro();
 
-  this->EndModify(disabledModify);
+  // In previous versions 0 was used to represent an unlimited number of control points.
```
lassoan, line 65:
```suggestion
  // In scenes created by Slicer version version 4.13.0 revision 30287 (built 2021-10-05)
  // and earlier versions, 0 was used to represent an unlimited number of control points.
```

lassoan (commented):
This looks good, thank you, just a few nitpicks in comments.

Sunderlandkyl:
Done!

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsFiducialNode.cxx
```diff
@@ -62,7 +62,12 @@ void vtkMRMLMarkupsFiducialNode::ReadXMLAttributes(const char** atts)
   vtkMRMLReadXMLIntMacro(requiredNumberOfControlPoints, RequiredNumberOfControlPoints);
   vtkMRMLReadXMLEndMacro();
 
-  this->EndModify(disabledModify);
+  // In scenes created by Slicer version version 4.13.0 revision 30287 (built 2021 - 10 - 05).
```
lassoan, line 65:
```suggestion
  // In scenes created by Slicer version version 4.13.0 revision 30287 (built 2021-10-05).
```

lassoan:
Thank you for the fix and all the follow-up improvements!

PR #5923, opened 2021-09-30
COMP: add QtMultimedia Python wrapper from CTK

atracsys-sbt:
(This PR originates from [this discussion](https://github.com/Slicer/Slicer/pull/5909))

Support for the QtMultimedia component has been recently [added to CTK](https://github.com/commontk/CTK/pull/992).
This PR simply forward this support to Slicer.

PR #5931, opened 2021-10-05
ENH: Update CTK to include QtMultimedia Python wrappers

jcfr:
List of CTK changes:

```
$ git shortlog 34a390a4..ec9fcee5 --no-merges
Sylvain Bernhardt (1):
      ENH: Add Python wrapping for Qt multimedia component (PR-992)
```

### DIFF
--- a/Base/QTCore/Resources/Certs/make-ca.sh
+++ b/Base/QTCore/Resources/Certs/make-ca.sh
@@ -288,6 +288,13 @@ for tempfile in ${TEMPDIR}/certs/*.tmp; do
   echo "Certificate:  ${certname}"
   echo "Keyhash:      ${keyhash}"
 
+  if [[ "${certname}" == "DST Root CA X3" ]]; then
+    echo "*************************************"
+    echo "* Skipping ${certname}"
+    echo "*************************************"
+    continue
+  fi
+
   # Import certificates trusted for SSL/TLS into
   # GnuTLS certificate bundle
   if test "${satrust}x" == "Cx"; then
--- a/CMake/SlicerBlockInstallQt.cmake
+++ b/CMake/SlicerBlockInstallQt.cmake
@@ -7,6 +7,17 @@ set(QT_INSTALL_LIB_DIR ${Slicer_INSTALL_LIB_DIR})
     "Qt5::Gui"
     )
 
+  # Multimedia Dependencies
+  if("Qt5::Multimedia" IN_LIST QT_LIBRARIES)
+    find_package(Qt5 REQUIRED
+      COMPONENTS
+        MultimediaWidgets
+      )
+    list(APPEND QT_LIBRARIES
+      "Qt5::MultimediaWidgets"
+      )
+  endif()
+
   # WebEngine Dependencies
   if("Qt5::WebEngine" IN_LIST QT_LIBRARIES)
     find_package(Qt5 REQUIRED
--- a/CMake/SlicerBlockInstallQt.cmake
+++ b/CMake/SlicerBlockInstallQt.cmake
@@ -7,17 +7,6 @@ set(QT_INSTALL_LIB_DIR ${Slicer_INSTALL_LIB_DIR})
     "Qt5::Gui"
     )
 
-  # Multimedia Dependencies
-  if("Qt5::Multimedia" IN_LIST QT_LIBRARIES)
-    find_package(Qt5 REQUIRED
-      COMPONENTS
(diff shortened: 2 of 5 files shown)


lassoan:
@jcfr this looks like the change that I've just reverted (https://github.com/Slicer/Slicer/pull/5923) because it [breaks the install package](https://discourse.slicer.org/t/slicer-wont-start/20017/4?u=lassoan). Have you managed to fix this problem?

jcfr:
On windows, here is the call to `windeployqt.exe` configured in `cmake_install.cmake`:

```
execute_process(COMMAND "D:/Support/Qt2/5.15.2/msvc2019_64/bin/windeployqt.exe" --release --no-compiler-runtime --no-translations --no-plugins -core -widgets -multimedia -network -opengl -printsupport -xml -xmlpatterns -svg -sql -webengine -webenginewidgets -webchannel -script -test -gui -qml -quick -quickwidgets -webenginecore -positioning -designercomponents "${CMAKE_INSTALL_PREFIX}/bin/SlicerApp-real.exe")
```

References:

* list of libraries for initializing list of components passed to `windeployqt.exe`

https://github.com/Slicer/Slicer/blob/bd6021edf5414aaacb3b5d25a43b47e5b8443268/CMake/SlicerBlockInstallQt.cmake#L6-L35

* call to `windeployqt.exe`

https://github.com/Slicer/Slicer/blob/bd6021edf5414aaacb3b5d25a43b47e5b8443268/CMake/SlicerBlockInstallQt.cmake#L154-L181

lassoan:
It seems that `multimediawidgets` library is missing. I guess only `multimedia` library is specified because PythonQt uses a single flag to enable both `multimedia` and `multimediawidgets` Qt libraries.

jcfr:
This is now ready for review. The latest commit is expected to fix the packaging issue.

jcfr:
Alternatively, we could introduce an option called `Slicer_BUILD_AUDIO_SUPPORT` and then list `Qt5::MultimediaWidgets` here: 

https://github.com/Slicer/Slicer/blob/1a6186cde4676f9935bf2668cc0c78954ce4db21/CMakeLists.txt#L580-L588

Doing something like this:

```diff
  set(Slicer_REQUIRED_QT_MODULES
    Core Widgets
-    Multimedia
    Network OpenGL
    PrintSupport # Required by "Annotations" module
    UiTools #no dll
    Xml XmlPatterns
    Svg Sql
    )
+if(Slicer_BUILD_AUDIO_SUPPORT)
+  list(APPEND Slicer_REQUIRED_QT_MODULES
+    Multimedia
+    MultimediaWidgets
+    )
+endif()
```

atracsys-sbt:
While testing the new sound inclusion in slicer, I realized something weird: in the python terminal, I can play sound no problem. However, running the exact same code (with an absolute path to a .wav file) in a scripted module does nothing (no sound, no error)...

lassoan:
> While testing the new sound inclusion in slicer, I realized something weird: in the python terminal, I can play sound no problem. However, running the exact same code (with an absolute path to a .wav file) in a scripted module does nothing (no sound, no error)...

Maybe you need to pump the event loop. Try adding `slicer.app.processEvents()` to see if it solves the issue. If you can upload a minimal example that reproduces the issue then I can have a look.

atracsys-sbt:
> 
> 
> > While testing the new sound inclusion in slicer, I realized something weird: in the python terminal, I can play sound no problem. However, running the exact same code (with an absolute path to a .wav file) in a scripted module does nothing (no sound, no error)...
> 
> Maybe you need to pump the event loop. Try adding `slicer.app.processEvents()` to see if it solves the issue. If you can upload a minimal example that reproduces the issue then I can have a look.

I tried, but it didn't seem to work. Here is the [scripted module](https://github.com/Slicer/Slicer/files/7294749/module1.zip) I use to test this problem.

lassoan:
Thanks for the example it made things very clear. Since you create the QSound object as a local variable, it gets immediately deleted when you exit the method. You need to keep the object alive, for example by storing it in a member variable.

lassoan (dismissed):
This looks perfect to me.

jcfr:
Note that including `Qt5::Multimedia` only if `Slicer_BUILD_MULTIMEDIA_SUPPORT ` is `ON` is valid since it was originally introduced to support use of `QSound`. See https://github.com/Slicer/Slicer/commit/6a2af8e4c2a95497d5962335b9bfc7fa75a2641f

PR #5932, opened 2021-10-05
COMP: Markup logic getpointer

pieper:
Fixes build on mac.

 % cc --version
Apple clang version 13.0.0 (clang-1300.0.29.3)
Target: x86_64-apple-darwin20.6.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin


lassoan:
@pieper do you know why these `.GetPointer()` calls are needed? It is not just ugly and unnecessary, but a new source of build failures because it just occurs in very special build toolsets. 

Does the error message provide the list of alternative interpretations that make the comparison ambiguous?

Instead of adding `.GetPointer()` can you fix the error by adding/removing includes?

jcfr:
Could this be a VTK 8 build ?

lassoan:
Thanks for the additional info. It seems that the problem is that two weak pointers are compared, which can be either compared as weakpointer==weakpointer or vtkobject==weakpointer.

What I don't understand why the compiler don't have a preference order and go through that? Why doesn't it use the weakpointer==weakpointer version?

Can you create a minimal example that we could submit to the VTK issue tracker?

Can you check if removing this line fixes the issue for you?

https://github.com/Kitware/VTK/blob/f0105d7ad3a80732ceb3ef54756782fde5cf139c/Common/Core/vtkWeakPointerBase.h#L115

pieper:
It does seem like these lines should cover this case:

https://github.com/Kitware/VTK/blob/f0105d7ad3a80732ceb3ef54756782fde5cf139c/Common/Core/vtkWeakPointerBase.h#L100-L103

pieper:
Yes, the code builds with VTK 9 on this compiler.

The difference seems to be in the header, where vtk9 has some extra operator overloads.  So either we (a) merge this `GetPointer` fix, or (b) patch our pre-9 vtk fork with these extra cases, or we (c) say that vtk8 builds are not supported anymore.  I would vote for (a) short term, then (c) soon.  @jcfr @lassoan ?

https://github.com/Kitware/VTK/blob/f0105d7ad3a80732ceb3ef54756782fde5cf139c/Common/Core/vtkWeakPointerBase.h#L99-L115

https://github.com/Slicer/VTK/blob/v9.0.1/Common/Core/vtkWeakPointer.h#L248-L278

PR #5933, opened 2021-10-06
BUG: Fix slice intersection representation point default case

gaoyi:
When no intersection point was detected, return the center of slice view.
But the index was wrong before


lassoan (approved):
Thank you, good catch!

Issue #5936, opened 2021-10-08
Intersection point moves when rotating with Ctrl+Alt+Drag


gaoyi:
[Give a short summary of what the problem is]

## Steps to reproduce

The same question has been asked on the Forum where I put a video showing the situation:

https://discourse.slicer.org/t/rotation-center-moves-in-ctrl-alt-drag/19986

Copied from there:

I’m trying to use the Ctrl+Alt+Drag to rotate the volume interactively. However the Center seems moving with rotation.

But this only seems happen when I have a customized layout (shown in the video). This moving-center does not happen in standard layout such as the Four-Up and others.

Andras pointed out there that the problem is that the near-parallel planes indeed intersect, and a checking of close-to-parallelism is needed.

It is therefore added in the pull request.

## Expected behavior

The rotation center should not move with rotating, just as the situation in the standard layout such as the Four-Up and others.

## Environment
I'm tested on the system (version) below but I believe the problem is cross platform.
- Slicer version: Slicer-4.11.20210226
- Operating system:  Linux Mint 19.3

jamesobutler:
@lassoan since you merged #5937, this issue can be closed now?

lassoan:
Yes, good catch, it is fixed now.

PR #5937, opened 2021-10-08
PERF: two slice intersection lines under 3 degrees are not counted

gaoyi:
when computing rotation center among slice intersection lines, if
two lines intersect under 3 degrees, they are considered to be
parallel and their intersection point is not counted.

Refs:
https://discourse.slicer.org/t/rotation-center-moves-in-ctrl-alt-drag/19986
and
https://github.com/Slicer/Slicer/issues/5936


lassoan:
Don't create a new pull request, just force-push to the same branch and github will update this pull request automatically.

jamesobutler:
@lassoan Can you provide review again? It appears the user has issued changes, but just has things into 2 commits which can be safely Squashed and Merged at time of approving and merging.

lassoan:
We will get back to this after https://github.com/cpinter/Slicer/tree/interactive-slice-intersection-4 is integrated.

lassoan:
There is no interference with https://github.com/Slicer/Slicer/pull/6008, so this can be merged now.

PR #5949, opened 2021-10-13
BUG: Fix segfault when delete markup in its EndPlacementEvent

Connor-Bowley:
Using a smart pointer instead of raw pointer so the lifetime of
activeWidget is extended in the case it is deleted by interaction event


lassoan:
Thank you, it looks good!

PR #5950, opened 2021-10-16
ENH: Add custom lib `_INCLUDE_DIRS` to `SlicerConfig.cmake`

kerim371:
The commit introduces new target type `SLICER_CUSTOM_TARGETS`.
SlicerCAT lib (`slicerMacroBuildApplication()`) is  an example of that.
It adds `${lib_name}_INCLUDE_DIRS` to the `SlicerConfig.cmake`.
This allow to easilly link to the library from external modules.

It also possible to set this prop to any custom lib target, set `${lib_name}_INCLUDE_DIRS` cache var and then this `_INCLUDE_DIRS` var will be added to `SlicerConfig.cmake`. This makes possible to easilly add custom libraries and link to them from external modules.


kerim371:
If possible, please give me feedback.
Probably in addition to this I could try to add these `_INCLUDE_DIRS ` to the target property

PR #5966, opened 2021-10-27
ENH: Organize menu bar with a few new options

jamesobutler:
I decided to give the Menu Bar some attention to more cleanly organize actions and add a few new ones. I primarily was inspired by Visual Studio Code's menu bar as a menu with lots of organized options.

## File Menu

| Current | This PR |
|----------|---------|
|image|image|

## Edit Menu

| Current | This PR |
|----------|---------|
|image|image|

## View Menu

| Current | This PR |
|----------|---------|
|image|image|

## Help Menu

| Current | This PR |
|----------|---------|
|image|image|


jamesobutler:
@jcfr @lassoan @pieper I've reverted to the "Add" terminology instead of "Load", but I do notice that in the Welcome module it is using the "Load" terminology which is inconsistent. Should the Welcome module be changed to "Add Data"? Or File->Add Data changed to File->Load Data to match the Welcome module? Is the Welcome module more well known or the Menu bar?
image

I've updated the current "DICOM" action to be "Add DICOM Data" as currently it the action just is describing a file format and it is unclear what it is for.

I've also adopted the design principle described at https://docs.microsoft.com/en-us/windows/win32/uxguide/cmd-menus#icons which details adding icons in the menu bar for commonly used actions and not to feel obligated to add icons for every action. This design change is most noticeable in my updated Help menu.

> - Consider providing menu item icons for:
>   - The most commonly used menu items.
>   - Menu items whose icon is standard and well known.
>   - Menu items whose icon well illustrates what the command does.
> - If you use icons, don't feel obligated to provide them for all menu items. Cryptic icons aren't helpful, create visual clutter, and prevent users from focusing on the important menu items.

## File Menu

| Current | This PR |
|----------|---------|
|image|image|

## Edit Menu

| Current | This PR |
|----------|---------|
|image|image|

## View Menu

| Current | This PR |
|----------|---------|
|image|image|

## Help Menu

| Current | This PR |
|----------|---------|
|image|image|

jamesobutler:
With the new icons to support the new menu bar options, I should figure out how to best use dark/light versions based on the  dark/light theme.

jamesobutler:
I've pushed updates to handle switching from the dark icons for the light theme and the light icons for the dark theme. This happens on palette changed. Below is an example

| Light Theme | Dark Theme |
|----------|---------|
|image|image|

lassoan:
Thanks for the updates, The icons make the menu nicer and some actions more recognizable.

Coloring of the icons is an interesting option. Based on my experiments so far, using the same color for both dark and light mode seemed feasible (for example color #f7bf21) and it greatly simplifies things. We discussed it with one of Kitware's UI designers and he confirmed that it is possible to find colors that are visible over both dark and light backgrounds. If we want to change icon colors (which could be nice, because it could have stronger contrast) then the option that you implemented would be usable, but it would be better to avoid duplicating all icons (a bit harder to create and maintain them), for example by recoloring the icons automatically.

jamesobutler:
@lassoan I've pushed updates to my `ENH: Update action icons in Menu Bar` that should improve maintainability by recoloring the SVG's if a color is chosen for the icon that is different from the default black. They are currently set to be the same color as the text color in the menus to match. Black is the default download color from https://fonts.google.com/icons.

jamesobutler:
Any additional thoughts on this work? I have one question that hasn’t been answered yet quoted below as it relates to past discussion on terminology for actions that are in the File menu.

> @jcfr @lassoan @pieper I've reverted to the "Add" terminology instead of "Load", but I do notice that in the Welcome module it is using the "Load" terminology which is inconsistent. Should the Welcome module be changed to "Add Data"? Or File->Add Data changed to File->Load Data to match the Welcome module? Is the Welcome module more well known or the Menu bar?
> image

pieper:
I think the words are close enough that this hasn't been a source of confusion, but I agree it's better to be consistent and use Add.

jamesobutler:
I've pushed an additional commit here to make the actions in the Welcome module use the same "Add" terminology as the File Menu and also use the same icons. "Add Data" button is now on the left and "Add DICOM Data" on the right to match the order in the Load/Save ToolBar which has this left/right order and to match the top/down order in the File menu. "Add Data" is situated to be more important than "Add DICOM Data".

The "Customize Slicer" button is now also using the same icon as the "Application Settings" action.
| Current | This PR |
|----------|---------|
|image|image|

jamesobutler:
Rebased against develop to fix merge conflicts.

jamesobutler:
Any additional thoughts on this? Would be nice to have the updated and better organized menu options for Slicer 5.

jamesobutler:
Rebased against `master` to fix merge conflicts.

lassoan:
@jamesobutler is there anything in this pull request that is not covered in the now merged https://github.com/Slicer/Slicer/pull/6185?

jamesobutler:
The remaining part of this pull request includes adding improved icons for the updated menu bar options. This had been the part with debate about whether to use google icons from https://fonts.google.com/icons with recoloring based on current style applied, or get a Kitware designer to design brand new icons for some actions in the menu bar, or some other option. It was the hold of up deciding on how to improve icons which is why I created #6185 to include all other changes except for icon rework.

jamesobutler:
Closing this PR now that the remaining update action icons work has been revived in https://github.com/Slicer/Slicer/pull/6517.

PR #5971, opened 2021-10-29
ENH: Move RobustStatisticsSegmenter and SimpleRegionGrowingSegmentation modules to legacy category

lassoan:
While RobustStatisticsSegmenter and SimpleRegionGrowingSegmentation modules work quite nicely for easy segmentation tasks (segmenting single structure with good contrast compared to surrounding structures), they are much less versatile and convenient than current tools in the Segment Editor. Specifically, Grow from seeds effect allows segmenting multiple structures at once, it is fast and offers even quicker incremental updates, masking, live 3D preview, etc.

To make it more clear for users, that we recommend to use the Segment Editor, it makes sense to de-emphasize these old modules by showing them under Legacy category in the module tree.

See details in this discussion: https://github.com/Slicer/Slicer/pull/5971


jamesobutler:
@lassoan Should Annotations module be in the legacy category as well to encourage use of Markups module instead?

PR #5986, opened 2021-11-03
Improvements to displayed markup type including "Point List" terminology

### LINKED ISSUE #5976: Make displayed markup node type names more user-friendly
It is not clear for users that "Fiducial node" means a "fiducial list". It is not nice that a closed curve is referred to as "ClosedCurve node" on the GUI.

Use a different string for markup type in code and files (simple machine-readable string, such as `Fiducial`, `ClosedCurve`) and on the GUI (human-readable, translatable string, such as `fiducial list`, `closed curve`).

Automatic conversion from CamelCase to space separators could work, but would not be translatable to different languages.

## Additional context

https://discourse.slicer.org/t/delete-control-point-delete-fiducial-pop-up-confirm-box/20430/14

jamesobutler:
This closes #5976 and is based on @pieper's original suggestion of a new method that returned a string for markup type that was appropriate to be used in the GUI unlike `GetMarkupType()` which is used for internal logic for registering markups.

This makes the GUI easier to read and understand such as "Delete ClosedCurve" -> "Delete Closed Curve". Also the change from the term "Fiducial" -> "Point List" to represent a group of control points. Note that this improves usability for users, but does not yet make things simpler for developers. Developers will still use `slicer.mrmlScene.AddNewNodeByClass("vtkMRMLMarkupsFiducialNode")` to create what is shown as a "Point List" which is a bit confusing.

## Markups module and toolbar updates

| Current | This PR |
|----------|---------|
|image|image|

- Other considerations could be to update the default short node name from "F" to "PL" or something to match the updated markup type displayed to users. Thoughts on a better short node name?

## Right-click context menu updates

| Current | This PR |
|----------|---------|
|image|image|
|image|image|


jcfr (approved):

jcfr:
@smrolfe

smrolfe:
This will be a big improvement! I like the consistency of using "PL" for the node name.

jamesobutler:
Updated with the change of the short node name from "F" to "PL" to match displayed name of "Point List".
image

jcfr:
I suggest you move commit `ENH: Add displayed markup type name for improved reading in the GUI ` into  its own PR so that we can integrate the API improvement.

That way, the renaming of `Fiducial` to `PointList` could be discussed independently.

jamesobutler:
I believe based on our conversation of “Point List” at https://discourse.slicer.org/t/delete-control-point-delete-fiducial-pop-up-confirm-box/20430/18?u=jamesobutler with @lassoan, @pieper, @cpinter and others from SlicerMorph that is was well agreed upon. I don’t think there’s more to discuss?

jcfr:
> I don’t think there’s more to discuss?

Great. Thanks for including the discussion pointer.

Consider updating these test that assume the name of the fiducial:

```
Applications/SlicerApp/Testing/Python/SlicerMRBMultipleSaveRestoreLoopTest.py
140:      fidNode = slicer.util.getNode('F')
151:    fidNode = slicer.util.getNode('F')

Applications/SlicerApp/Testing/Python/SlicerMRBMultipleSaveRestoreTest.py
155:    fidNode = slicer.util.getNode('F')
163:    fidNode = slicer.util.getNode('F')
208:    fidNode = slicer.util.getNode('F')
```

jcfr:
> Developers will still use slicer.mrmlScene.AddNewNodeByClass("vtkMRMLMarkupsFiducialNode") to create what is shown as a "Point List" which is a bit confusing.

Agreed. 

Could you create an issue to capture the required changes to ensure naming is consistent with the developer API:

* Rename `vtkMRMLMarkupsFiducialNode` to `vtkMRMLMarkupsPointListNode`
* Deprecate the methods with `Fiducial` and add equivalent ones with `Point` in the name (if needed)
* Call `RegisterNodeClass` with former tag name

jamesobutler:
Regarding the tests with `fidNode = slicer.util.getNode('F')` usage, they surprisingly still pass. This is because some older API is still producing node names like "F". It appears going down this path is based on internal usage of markup type. This will likely be updated as part of cleaning up the markups API to remove the older "Fiducial" related methods from AnnotationFiducial days. 
```python
markupsLogic = slicer.modules.markups.logic()
markupsLogic.AddFiducial()
```
Things now have "Markups" and "Control Point" terminology used in the API. There's so many different methods doing the same type thing that I think there will need to be some deprecations as supporting all the various names of the year is just making the API too confusing about which version to use.  I've written up #5990 regarding this topic.

@jcfr I've written up #5989 detailing the work related to the API using the new name. I've also added the link to the discourse post about the decision from "Fiducial" to "Point List" for easier back tracking to the original source of the decision.

jamesobutler:
I can revert the change of the default "Point List" node name from "PL" to "F" if that is desired to proceed with this PR. Further discussion on this topic begins over at https://github.com/Slicer/Slicer/issues/5989#issuecomment-960417399.

lassoan:
This is a huge change - impacts all tutorials, previous and future discussions about markups, etc., but since there is an agreement that this is a painful but necessary step in the good direction, I'm in favor.

@jamesobutler I don't think we need to update all the tutorials but I think we should at least update the .md files in the repository (search for all usages of "fiducial") and update the node name to P or F or some other letter that everybody is comfortable with. Then I think this could be merged.

We can update the screenshots in the documentation after all the GUI updates settle down a bit more.

jamesobutler:
I've pushed another commit to update the readthedocs documentation with "Point List" instead of "Fiducial".

I've also reverted back to using the default "F" name for vtkMRMLMarkupsFiducialNode objects which are displayed as "Point List".

lassoan:
Thanks a lot @jamesobutler for implementing this and the feedback from everyone else. This is now merged. "Fiducial" term is now retired.

PR #5987, opened 2021-11-03
ENH: Improved interaction when camera is scaled

kerim371:
Before that user interaction (especially dragging) used to suffer when camera was scaled (aspect ratio is changed).
If the camera is stretched out along X dir then translation along X was slower.
This commit handles that and now user feel comfortable when the camera is scaled.
A C++ test was added to not accidently break that enhancement.
P.S. Scaling of camera is important in some areas (like seismic exploration) when different axes represent different units (distance and time for example)


lassoan:
Thank you for your contribution. The changes have been hardened, cleaned up, tested, and merged in 0c8fcb1917fe3efac6929da2a232af1ff213c6bd.

PR #5994, opened 2021-11-04
Fix deprecated declarations warnings anticipating transition to Qt6

jcfr:


Libs/MRML/Widgets/qMRMLLabelComboBox.cxx
```diff
@@ -126,20 +126,6 @@ qMRMLLabelComboBox::qMRMLLabelComboBox(QWidget* newParent)
 // --------------------------------------------------------------------------
 qMRMLLabelComboBox::~qMRMLLabelComboBox() = default;
 
-// ----------------------------------------------------------------
-void qMRMLLabelComboBox::printAdditionalInfo()
```
lassoan, line 130: I don't know about any particular use of this method, but was there any specific motivation to remove it?

jcfr: It addresses a warning about deprecated use of  'endl'. See details in commit message.

I can also revisit and instead address the warning.

lassoan: In general, we should just change the API for a good reason. Since it seems easy to update and maintain this method and it may be useful as a kind of a "PrintSelf", I think we should keep it.

jcfr: > it may be useful as a kind of a "PrintSelf"

Indeed, that was the original motivation when we started to develop the widgets ..

In the meantime, I will fix the warnings and keep the method.

jcfr: This will be "properly" fixed after https://github.com/commontk/CTK/pull/1008 is integrated.

lassoan (approved):
LGTM

jcfr:
Local Linux build was successful.

PR #5998, opened 2021-11-05
Enabling CCache builds

RafaelPalomar:
This PR enables the use of CCache for Slicer and external projects. This should work in any Unix system (Mac OS and Linux). To make this happen, this PR includes the following changes:
  - A `Slicer_USE_CCACHE` variable to enable/disable CCache builds (disabled by default).
  - CMake infrastructure (ExternalProjectGeneratePatchStep) to patch External Projects over using lists of `.patch` files.
  - A set of patches that will add the needed CMake code to relevant External Projects for enabling CCache.


lassoan:
Thank you for working on this.

How do you use ccache? I've read the ["Why bother?" section](https://ccache.dev/) and it is not clear how ccache could be useful for me. I very rarely do a clean build and when I do then I want to make sure that it is completely clean and no previously cached content can play any role. I very rarely build Slicer in the same build mode in multiple folders because it is just too big (a debug-mode build is about 50GB). I also don't see how this could work for debug symbol (pdb) files. If there are no debug symbols then it does more harm than good. Caching also slightly slows down all original builds, which is the most common operation on developer computers.

Maybe it could be useful for continuous integration? We cache all the dependencies in the docker image, but not the Slicer build files, so there could be relevant caching. In our docker image could we replace the compiler executable by ccache and avoid all the patching of the CMakeLists.txt files?

As I understand, ccache replaces the compiler. Since we already pass the compiler executable (CMAKE_CXX_COMPILER) to all external projects, couldn't we just change the value of this variable instead of patching CMakeLists.txt files? Many of Slicer dependencies are large, well established libraries, so that they don't have explicit support for ccache either means that they can either use ccache without CMakeLists.txt file changes or users/developers don't use ccache.

RafaelPalomar:
> Thank you for working on this.
> 
> How do you use ccache? I've read the ["Why bother?" section](https://ccache.dev/) and it is not clear how ccache could be useful for me. I very rarely do a clean build and when I do then I want to make sure that it is completely clean and no previously cached content can play any role. I very rarely build Slicer in the same build mode in multiple folders because it is just too big (a debug-mode build is about 50GB). I also don't see how this could work for debug symbol (pdb) files. If there are no debug symbols then it does more harm than good. Caching also slightly slows down all original builds, which is the most common operation on developer computers.
> 
> Maybe it could be useful for continuous integration? We cache all the dependencies in the docker image, but not the Slicer build files, so there could be relevant caching. In our docker image could we replace the compiler executable by ccache and avoid all the patching of the CMakeLists.txt files?
> 
> As I understand, ccache replaces the compiler. Since we already pass the compiler executable (CMAKE_CXX_COMPILER) to all external projects, couldn't we just change the value of this variable instead of patching CMakeLists.txt files? Many of Slicer dependencies are large, well established libraries, so that they don't have explicit support for ccache either means that they can either use ccache without CMakeLists.txt file changes or users/developers don't use ccache.



> Thank you for working on this.
> 
> How do you use ccache? I've read the ["Why bother?" section](https://ccache.dev/) and it is not clear how ccache could be useful for me. I very rarely do a clean build and when I do then I want to make sure that it is completely clean and no previously cached content can play any role. I very rarely build Slicer in the same build mode in multiple folders because it is just too big (a debug-mode build is about 50GB). I also don't see how this could work for debug symbol (pdb) files. If there are no debug symbols then it does more harm than good. Caching also slightly slows down all original builds, which is the most common operation on developer computers.
> 
> Maybe it could be useful for continuous integration? We cache all the dependencies in the docker image, but not the Slicer build files, so there could be relevant caching. In our docker image could we replace the compiler executable by ccache and avoid all the patching of the CMakeLists.txt files?
> 
> As I understand, ccache replaces the compiler. Since we already pass the compiler executable (CMAKE_CXX_COMPILER) to all external projects, couldn't we just change the value of this variable instead of patching CMakeLists.txt files? Many of Slicer dependencies are large, well established libraries, so that they don't have explicit support for ccache either means that they can either use ccache without CMakeLists.txt file changes or users/developers don't use ccache.

@lassoan I just opened a thread in Discouse about this ([Using CCache](https://discourse.slicer.org/t/speeding-up-slicer-builds-with-ccache/20497))

lassoan:
OK, let's continue the discussion there.

CMakeLists.txt
```diff
@@ -1,5 +1,14 @@
 cmake_minimum_required(VERSION 3.13.4)
 
+#-----------------------------------------------------------------------------
+# CCache
+#-----------------------------------------------------------------------------
+option(Slicer_USE_CCACHE "Build ${PROJECT_NAME} using CCache." OFF)
+find_program(CCACHE_PROGRAM ccache)
+if(CCACHE_PROGRAM)
+    set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE "${CCACHE_PROGRAM}")
```
jcfr, line 9: This would compete with https://github.com/Kitware/CMake/blob/4fd2d62613af072204910b303fc1506804279df1/Modules/CTestUseLaunchers.cmake#L70

lassoan:
We have a lot of open pull requests that makes it harder to see if there are active pull requests that require attention. What do you think about closing this pull request, maybe adding an issue to keep track of the idea and refer to this branch?

RafaelPalomar:
> We have a lot of open pull requests that makes it harder to see if there are active pull requests that require attention. What do you think about closing this pull request, maybe adding an issue to keep track of the idea and refer to this branch?

Yes, closing this PR makes sense. I'll add an issue about this. Optionally, if you think it could be useful @jcfr proposed a valid way to use `ccache` through symlinking. Maybe this could be documented in readthedocs?

lassoan:
OK.

> if you think it could be useful @jcfr proposed a valid way to use ccache through symlinking. Maybe this could be documented in readthedocs?

It could be added as a "Tip" in this section: https://slicer.readthedocs.io/en/latest/developer_guide/build_instructions/linux.html#build-slicer

PR #5999, opened 2021-11-05
ENH: add WebServer module

pieper:
Code imported from Steve Pieper's development repository.

https://github.com/pieper/SlicerWeb


jcfr:
Thanks for the contribution.

To give the code a chance to mature and be more broadly tested. What do you think of:
- transferring the repository to the Slicer organization
- creating an extension

Later, we could then add the module to the core as a Slicer remote like we do for the Surface Toolbox (see `Slicer/SuperBuild.cmake`)

pieper:
@jcfr I went with this option based on a discussion with @lassoan [on the Slicer discourse](https://discourse.slicer.org/t/how-to-load-nifti-file-from-web-browser-link/18664/23).  We felt there was a motivating use case to have this feature in the core and little downside for anyone who's not using the module.

Regarding where to keep the code, my experience has been that having many repositories for core modules leads to extra burden for maintainers since we need multiple PRs to fix common issues.

pieper:
I've marked this as draft for now so we can discuss.  There are some TODO items in the code and we should think about which of those are important enough to put in place before merging this.  The most important in my mind is putting an explicit start/stop button in the module (currently it starts on `enter`).

Docs/user_guide/modules/webserver.md
```diff
@@ -0,0 +1,73 @@
+# WebServer
+
+## Overview
+
+Creates a fairly simple but pretty powerful web server that can respond to http(s) requests with data from the current Slicer or modify the Slicer state.  This module is meant to be the basis for implemeting web applications that use Slicer as a remote render / computation engine or for controlling your Slicer instance for interaction with other system code like shell scripts or other applications.
+
+There are three basic types of endpoints:
+- **Static:** Hosts files out of the module's `docroot` like any standard http server.  Currently this is used just for examples.
```
lassoan, line 8: We should separate potentially clinically useful examples from from tests and pure technology demonstrations. Clinically useful stuff should be at the top and the rest (timeimage, timecube, spinning cube) should be in a separate section, but preferably in a separate page (can be linked from the main page).

Docs/user_guide/modules/webserver.md
```diff
@@ -0,0 +1,73 @@
...
+
+The web server is integrated with the Qt event loop so it can be used together with the interactive session.
+
+This code has been developed over a number of years in a separate repository where there are additional experiments demontrating other potential uses.  The version in Slicer core has been stripped down to address the most common expected use cases.  See [https://github.com/pieper/SlicerWeb](https://github.com/pieper/SlicerWeb).
+
+Also note that the code should be considered somewhat experimental and a likely security risk.  Do not expose web server endpoints on the public internet without careful consideration.
+
+## Current Features
+
+The server launches on port 2016 by default when you enter the module.  Access logs are shown in the GUI by default and can be routed to the console if persistence is needed (the GUI logs are cleared periodically).  If port 2016is in use, other ports are checked sequentially until an open port is found, allowing more than one Slicer web server to run concurrently.
```
lassoan, line 20:
```suggestion
The server launches on port 2016 by default when you enter the module.  Access logs are shown in the GUI by default and can be routed to the console if persistence is needed (the GUI logs are cleared periodically).  If port 2016 is in use, other ports are checked sequentially until an open port is found, allowing more than one Slicer web server to run concurrently.
```

Docs/user_guide/modules/webserver.md
```diff
@@ -0,0 +1,73 @@
...
+The demo page shows most of the features of the Slicer web api and is hosted using the Static api.
+
+## Remote Control
+
+The web server can also be accessed via other commands such as `curl`.
+
+For example, these command can be used to download data and change the screen layout:
+
+```
+curl -X POST localhost:2016/slicer/repl --data "import SampleData; volumeNode = SampleData.SampleDataLogic().downloadMRHead(); slicer.app.layoutManager().setLayout(slicer.vtkMRMLLayoutNode.SlicerLayoutOneUpRedSliceView)"
```
lassoan, line 33: `repl` is a bit too abstract (and it is not really a loop but we really just perform a single command). What are some other endpoint names that are commonly used for this kind of functionality?

Some examples that look more readable to me:
- localhost:2016/slicer/execute
- localhost:2016/slicer/exec
- localhost:2016/slicer/cmd

Docs/user_guide/modules/webserver.md
```diff
@@ -0,0 +1,73 @@
...
+Other endpoints allow get/set of transforms and fiducials.
+
+## Panels and their use
+
+Currently there is only the logging console and the demo buttons.  Further controls may be added depending on needs.
+
+
+## Related modules
+
+- The [OpenIGTLink](https://github.com/openigtlink/SlicerOpenIGTLink) Extension has some similar functionality cusomized for image guided therapy applications.  It should be preferred for integration with imaging devices and use in a clinical setting.
```
lassoan, line 62:
```suggestion
- The [OpenIGTLink](https://github.com/openigtlink/SlicerOpenIGTLink) Extension has some similar functionality customized for image guided therapy applications.  It should be preferred for integration with imaging devices and use in a clinical setting or setting up continuous high-throughput image and transform streams.
```

Modules/Scripted/WebServer/WebServer.py
```diff
@@ -0,0 +1,1507 @@
+# slicer imports
+from __main__ import vtk, qt, ctk, slicer
+from slicer.ScriptedLoadableModule import *
+
+# python imports
+try:
+    from BaseHTTPServer import HTTPServer
```
lassoan, line 7: Is this Python 2/3 compatibility? We can drop Python2.

Modules/Scripted/WebServer/WebServer.py
```diff
@@ -0,0 +1,1507 @@
...
+import os
+import pydicom
+import random
+import select
+import sys
+import socket
+import string
+import time
+try:
+    import urlparse
```
lassoan, line 23: Python 2/3 compatibility?

Modules/Scripted/WebServer/WebServer.py
```diff
@@ -0,0 +1,1507 @@
...
+    responseBody = fp.read()
+    fp.close()
+    return contentType, responseBody
+
+
+#
+# SlicerRequestHandler
+#
+
+class SlicerRequestHandler(object):
```
lassoan, line 363: Could we move out each request handler to a separate file?
It would make this file shorter and simpler and it could server as example for extensions defining and registering their own custom request handlers.

lassoan (commented):
Thank you, this is great! I added some comments and suggestions inline.

I think it is good to have it in the Slicer core, as the amount of code is quite small and we don't get much benefit from other half-integrated extensions either (LandmarkRegistration, SurfaceToolbox).

lassoan:
Some more findings:

- The web server is started when the module GUI is opened. Instead, there should be start/stop buttons. Maybe the server state could be persistent (server would automatically start with Slicer if the user started the server and did not stop it).
- There should be also a simple way to start Slicer from the command-line with a server listening at a specified port (with specified handlers enabled).
- Processing a `repl` request on Windows takes about 2-3 seconds (even if I just execute a `print()` function). This is very noticeable, even when it is used for opening an image.
- `http://localhost:2016/slicer/slice` returns with error 404 if no image is loaded. Since most newcomers would click through the sample API calls from top to bottom, it would be probably sufficient to add loading of a sample image above the slice call.
- Links on the http://localhost:2016/ServerTests page do not work ("Error loading page" message is displayed).
- It would be nice to add an example where there is some basic widgets and interaction, for example, a slider to browse slices?

lassoan:
I've tried this WebServer module to open a file in a running Slicer instance by running a .bat file and it works very nicely using this command (`%1` refers to the first command-line argument in .bat files):

    PythonSlicer.exe -c "import requests; filename=r'%1'.replace('\\','/'); requests.post(url='http://localhost:2016/slicer/repl', data='slicer.app.loadFiles([\''+filename+'\'])', timeout=15)"

The only thing that is not nice that there is a few-second delay before Slicer starts to process the command. It might be an issue with the QSocketNotifier. We tried to use it in SlicerJupyter but it had issues (100% CPU use for sure, but maybe performance problems, too). First we had to switch to simple polling in Windows, but then problems were found on Linux, too; so we ended up using polling on all platforms.

Docs/user_guide/modules/webserver.md
pieper: Good point.  Internally it uses python `exec` so that would be the best choice in terms of clarity and documenting the feature.  I'll add it and deprecate `repl`.

pieper:
Thanks for the review - good feedback 👍 

I'll do some more tweaks and test on windows.  Good topic to discuss in tomorrow's call.

jcfr:
Few high level comments

There are few aspects: 
* registering handlers for exposing Slicer functionalities through REST API
* serving web pages
* REPL

This is what why I suggest to split `WebServer.py` into different "logic" or even modules:
- SlicerRESTApi: Provide a way to handle request, register new request handlers, describe available handlers  (ideally using the [OpenAPI](https://swagger.io/specification/) specification). For now, I suggest to focus on integrating this in Slicer core.
- SlicerREPL
- SlicerWebServer: I think we should approach this as a list of web/html/javascript components (may be these could even be distributed using npm?) and provide a gallery of examples show casing how to use these.

jcfr:
> Good topic to discuss in tomorrow's call.

I will not be available tomorrow. That said, I will look forward the meeting notes.

lassoan:
I agree that the module should be split up to a module that has the infrastructure to register/unregister handlers (plugins). I suggested the same in my comments - have handlers in separate files; and have API that allows other modules to register their handlers, too.

I like the WebServer module name because it describes the scope of the module nicely and does not limit itself to a particular style of web API (although right now all handlers use REST API).

Distributing components via npm is a great idea (once we have some usable components and not just technology demos).

Modules/Scripted/WebServer/WebServer.py
```diff
@@ -0,0 +1,1507 @@
...
+    moduleDirectory = os.path.dirname(slicer.modules.webserver.path.encode())
+    self.docroot = moduleDirectory + b"/Resources/docroot"
+
+  def start(self):
+    """Set up the server"""
+    self.stop()
+    self.port = SlicerHTTPServer.findFreePort(self.port)
+    self.logMessage("Starting server on port %d" % self.port)
+    self.logMessage('docroot: %s' % self.docroot)
+    # example: certfile = '/Users/pieper/slicer/latest/SlicerWeb/localhost.pem'
```
mauigna06, line 1500: Hi Steve. Here is some feedback:
The default pem file path and an "autostart" flag (telling if the WebServer should start automatically at application launch) could be read from application settings. Or autostart could be turned off by default and you could launch Slicer with specifying a small script that starts the WebServer with the parameters you like (--python-code "import WebServer; server=WebServer.... ; server.start(...)“.

mhalle:
It might be worth looking at FastAPI, which provides a really easy way to create an API from python code, and includes a high performance web server. Self-documenting and uses typing as well.

https://fastapi.tiangolo.com/

pieper:
> It might be worth looking at FastAPI, which provides a really easy way to create an API from python code, and includes a high performance web server. Self-documenting and uses typing as well.
> 
> https://fastapi.tiangolo.com/

This could be good.  The problem I found with pretty much all python web servers is that they want to "own" the event loop where here event loop is owned by the Qt application.

@haehn is trying to recruit some students to help with this as part of another project so we can hope to get some more experiments going soon.

mhalle:
> This could be good. The problem I found with pretty much all python web servers is that they want to "own" the event loop where here event loop is owned by the Qt application.

It looks like there are ways to integrate the Qt event loop with asyncio, which in turn runs FastAPI through another layer (uvicorn or similar). On the other hand, I don't know if async event would be what you want.

lassoan:
I've fixed most of the review comments and TODOs. I haven't tested the DICOM handler.
It would be great if this could be integrated before Slicer5, for example to allow running Slicer as a viewer for SimpleITK/ITKPython.

lassoan:
@pieper Is there anything else you want to add/change in this module? If not, then I think we should rebase and merge it, as remote control of the application is quite important to be able to better compete with other viewers that start up faster and/or are more integrated into the Python environment.

I'm thinking about adding an ITKSphinxExample for using Slicer as a viewer at the [upcoming May 20 ITK](https://hackmd.io/WL6AxKj6Sy2O9GKtgj0LYA#Why-Attend) event and this feature would be a prerequisite. It may take a few days to get all the  issues ironed out on all platforms, so there is not too much time left.

pieper:
@lassoan yes, I am at a good spot with it.  The only [outstanding conflice](https://github.com/Slicer/Slicer/pull/5999/conflicts) is that I needed to upgrade pydicom.  I don't think that will be an issue so let's merge it and we if anything pops up we can deal with it independently.

jamesobutler:
The `master` branch is now failing CI checks. https://github.com/Slicer/Slicer/actions/runs/2310736284

pieper:
Thanks for the heads up @jamesobutler - looks like some simple format issues.  It would be nice if the lint just created a PR automatically.  I guess I can fix these up in the browser (I'm away from my regular setup).

jcfr:
> it would be nice if the lint just created a PR

This should be possible - I am taking note of this and will evaluate our options.

PR #6003, opened 2021-11-08
ENH: Remove Edit Box action

jamesobutler:
The Edit Box class was removed when the legacy "Editor" module was removed from Slicer in https://github.com/Slicer/Slicer/commit/39283db420baf502fa99865c9d5d58d0e5295a6e.

The Edit Box action was visible in Help->Keyboard Shortcuts and is now removed with this PR.

image


lassoan (approved):
Goot catch, thank you.

PR #5954, opened 2021-10-19
Update ITK to post-5.3rc03 including requiring at least C++14

jamesobutler:
This is part of work for #5673.

C++14 is the new minimum with CMake 3.16.3 as set by ITK 5.3 requirements (https://github.com/InsightSoftwareConsortium/ITK/blob/v5.3rc01/Documentation/SupportedCompilers.md).

This will enable future GrowCut remote ITK module work as @dzenanz was beginning in #5807.

PR #6012, opened 2021-11-10
DOC: Update build instructions with references to VS2022

jamesobutler:
This updates the Windows build instructions with instructions on how to use Visual Studio 2022. This new version has been released and is the primary download available at https://visualstudio.microsoft.com/downloads/ so people building Slicer for the first time will now be downloading this version. I have tested and Slicer builds successfully with Visual Studio 2022 and using the latest v143 toolset. Visual Studio 2022's biggest change is itself is now a 64-bit application.

I've also included another commit to recommend the usage of Qt 5.15.2 instead of 5.15.1.

### DIFF
--- a/Docs/developer_guide/build_instructions/linux.md
+++ b/Docs/developer_guide/build_instructions/linux.md
@@ -24,7 +24,7 @@ code of slicer, generating the project files and build the project.
 In addition, Slicer requires a set of support libraries that are not includes as
 part of the *superbuild*:
 
-- Qt5 with the components listed below. Qt version 5.15.1 is recommended, other Qt versions are not tested and may cause build errors or may cause problems when running the application.
+- Qt5 with the components listed below. Qt version 5.15.2 is recommended, other Qt versions are not tested and may cause build errors or may cause problems when running the application.
   - Multimedia
   - UiTools
   - XMLPatterns
--- a/Docs/developer_guide/build_instructions/windows.md
+++ b/Docs/developer_guide/build_instructions/windows.md
@@ -7,7 +7,7 @@
   - Note: CMake must be able to find `git.exe` and `patch.exe`. If git is installed in the default location then they may be found there, but if they are not found then either add the folder that contains them to `PATH` environment variable; or set `GIT_EXECUTABLE` and `Patch_EXECUTABLE` as environment variables or as CMake variables at configure time.
 - [Visual Studio](https://visualstudio.microsoft.com/downloads/)
   - any edition can be used (including the free Community edition)
-  - when configuring the installer, enable `Desktop development with C++` and in installation details, check `MSVC v142 - VS2019 C++ x64...` (Visual Studio 2019 v142 toolset with 64-bit support) - in some distributions, this option is not enabled by default
+  - when configuring the installer, enable `Desktop development with C++` and in installation details, check `MSVC v143 - VS2022 C++ x64...` (Visual Studio 2022 v143 toolset with 64-bit support) - in some distributions, this option is not enabled by default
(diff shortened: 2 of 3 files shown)


lassoan (commented):
Thank you, it is great that you have tested with VS2022 IDE and v143 toolset and it's good news that it works.

We need to decide if we want to upgrade to v143 toolset on the factory machines. If we update to v143 there then the build instructions should be updated as you did in this PR. If we don't update the toolset version on the factory machine yet then v143 should be added to the "Other compiler versions" for now.

Do you know about any specific advantage of switching to using the v143 toolset?

Do you know what is the minimum CMake version for using VS2022 and v143?

jamesobutler:
Here's the corresponding https://github.com/Slicer/DashboardScripts/pull/43 for what the factory machines use.

I don't have any specific knowledge of advantage of the v143 toolset, only that for new developers this is the default option when downloading Visual Studio 2022. They would have to pick a non-default option to still use the v142 toolset. Using the defaults therefore is the primary motivation for the change.

The "Visual Studio 17 2022" CMake generator was introduced in CMake version 3.21 ([source](https://cmake.org/cmake/help/v3.21/generator/Visual%20Studio%2017%202022.html)). I've personally used CMake 3.21.4.

lassoan:
Saving a few clicks during VS2021 installation would be nice and we should document that VS2021 with v143 toolset is known to work. However, I'm not sure that it makes it worth increasing the minimum CMake version and updating the factory machines configuration.

I would lean towards just adding a note that VS2021 + v143 configuration has been tested and it worked, but would not require users to update yet.

@jcfr @sjh26 @pieper what do you think?

jamesobutler:
It seems impossible to install Visual Studio 2019 anymore. So not sure how to instruct users new to building and getting all the build tools how to find a download link to Visual Studio 2019 if we expect them to still use an older CMake version (pre-3.21.0). https://visualstudio.microsoft.com/downloads/ is only about Visual Studio 2022 with no options for the 2019 IDE.

Therefore if we are instructing users to download Visual Studio 2022, since it is the only thing readily available, I would suggest that the Factory build machines use at least the same Visual Studio 2022 IDE with either v143 toolset or still use the v142 toolset if adverse to the new toolset.

lassoan:
I'm not concerned about the IDE version (I think build toolsets can be installed without an IDE), although there could be some CMake troubles (older CMake not recognizing VS2021 or v143 toolset).

The main question for me if we should raise minimum CMake version requirement and transition to v143 build toolset. Both can be risky (lead to errors that may not be immediately obvious). I'm also not sure

jamesobutler:
Reminder that to update to ITK 5.3.0, which is going to be released any day now, CMake minimum version will also have to be increased. In #5954 there is `COMP: Update minimum CMake version from 3.13.4 to 3.16.3`.

jamesobutler:
> I'm also not completely sure about Python compatibility - can it cause any problem if we use a different toolset version than what Python package builders use?

Binary compatibility information is available at https://docs.microsoft.com/en-us/cpp/porting/binary-compat-2015-2017?view=msvc-170

> You can mix binaries built by different versions of the v140, v141, v142, and v143 toolsets. However, you must link by using a toolset at least as recent as the most recent binary in your app. 

There is no problem with us using v143 toolset if someone builds their python package with v140-v143. There would however be a problem if we still use v142 and then a python package is using the v143 toolset.

lassoan:
Updated the changes so that build instructions still use v142 toolset (since that is used in official builds) and added note about v143 as an occasionally tested configuration that appears to work well. I'm closing this pull request as the VS2022 instructions are now included, there are now conflicting changes, and by keeping the PR open would make the pull request list more cluttered (making harder to find active pull requests). Once the factory machines will be updated to use v143 toolset we will update the documentation to reflect that.

PR #6018, opened 2021-11-11
ENH: Move Annotations module to Legacy category

jamesobutler:
Annotations module has been replaced by the Markups module. This moves the Annotations module to the legacy category to make it clear that any remaining objects are legacy and will be removed in the future.

Deprecating of the Annotations module is mentioned in the [Slicer roadmap](https://github.com/Slicer/Slicer/wiki/Roadmap). Moving it to the Legacy category does not hide the module from users, but does put it into a more appropriate category rather than being the first "main" module in the module selector menu.

| Current | This PR |
|----------|---------|
|image|image|

### DIFF
--- a/Modules/Loadable/Annotations/qSlicerAnnotationsModule.cxx
+++ b/Modules/Loadable/Annotations/qSlicerAnnotationsModule.cxx
@@ -161,7 +161,7 @@ QIcon qSlicerAnnotationsModule::icon() const
 //-----------------------------------------------------------------------------
 QStringList qSlicerAnnotationsModule::categories() const
 {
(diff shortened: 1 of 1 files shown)


jamesobutler:
I could also remove from the "Informatics" category so that it is only seen in the context of being fully part of the "Legacy" category.
image

lassoan:
Thank you. I agree, we should remove it from the Informatics category, too.

jamesobutler:
I have updated the commit to remove it from the "Informatics" category so that now it is only within the "Legacy" category.

lassoan (approved):
Thank you. There were no objections, so we should proceed with this change.

PR #6020, opened 2021-11-12
BUG: Fix place control point button menu button always visible

### LINKED ISSUE #5914: Markups: Down-arrow on place button should not be always visible
The down-arrow button on the place widget's place button should be hidden if the force single/force multiple option is chosen. 

## Environment
- Slicer version: Slicer-4.11.0-2021-09-27
- Operating system: Windows

@smrolfe

jamesobutler:
This closes #5914.

The Place control point QToolButton was changed in https://github.com/Slicer/Slicer/commit/09d12ae6f7598fe8490b691d17a35dcca5bd344a (specifically [here](https://github.com/Slicer/Slicer/commit/09d12ae6f7598fe8490b691d17a35dcca5bd344a#diff-b15fee96f47b8f3a2987141a9caae794d55d9a578a07a32a1febafac506d523eR53-R55)) to use the `QToolButton::MenuButtonPopup` type for [ToolButtonPopupMode](https://doc.qt.io/qt-5/qtoolbutton.html#ToolButtonPopupMode-enum). Previously it was the default `QToolButton::DelayedPopup`. DelayedPopup would automatically hide the menu arrow when there was no menu, however the `QToolButton::MenuButtonPopup` option always shows the button even when there is no menu.

To avoid showing a menu button with no menu, we can switch to the `QToolButton::DelayedPopup` mode when we do not have a menu for the QToolButton.

```python
place_widget = slicer.qSlicerMarkupsPlaceWidget()
place_widget.setMRMLScene(slicer.mrmlScene)
place_widget.setCurrentNode(slicer.mrmlScene.AddNewNodeByClass("vtkMRMLMarkupsFiducialNode"))
place_widget.show()
```
image

```python
place_widget.setPlaceMultipleMarkups(place_widget.ForcePlaceMultipleMarkups)
place_widget.show()
```
image

```python
place_widget.setPlaceMultipleMarkups(place_widget.ShowPlaceMultipleMarkupsOption)
place_widget.show()
```
image

### DIFF
--- a/Modules/Loadable/Markups/Widgets/qSlicerMarkupsPlaceWidget.cxx
+++ b/Modules/Loadable/Markups/Widgets/qSlicerMarkupsPlaceWidget.cxx
@@ -643,6 +643,8 @@ void qSlicerMarkupsPlaceWidget::setPlaceMultipleMarkups(PlaceMultipleMarkupsType
   if (d->PlaceButton)
     {
     d->PlaceButton->setMenu(d->PlaceMultipleMarkups == ShowPlaceMultipleMarkupsOption ? d->PlaceMenu : nullptr);
+    // Changing to DelayedPopup mode will hide menu button to avoid confusion of empty menu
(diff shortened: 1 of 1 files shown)


lassoan (approved):
This works great, thank you!

PR #6021, opened 2021-11-12
BUG: Fix unset control point position options visible in default place widget

### LINKED ISSUE #5917: Markups: Hide unset markup point position option from place widget by default
The updated qSlicerMarkupsPlaceWidget seems to create weird situations when it is used elsewhere such as in the SurfaceCut effect area. The place button has a menu drop down but there is no addition menu options as it is always in continuous placement mode.

By default, markups place widget should have the same behavior as before - delete placed points, and not offer unset points. Visibility of unset actions should be made a property of the widget so that it can be easily enabled/disabled in Qt Designer.

## Environment
- Slicer version: Slicer-4.11.0-2021-09-27
- Operating system: Windows

@smrolfe

jamesobutler:
This closes #5917.

Having unset control point position options visible in the place widget by default can lead to weird situations such as the functionality enabled for the SurfaceCut effect. This effect doesn't have a set number of control points and would not benefit from unsetting points. Having unset points not visible by default will maintain previous behavior prior to 09d12ae.

The options is turned on only for the Markups toolbar.

```python
place_widget =slicer.qSlicerMarkupsPlaceWidget()
place_widget.setMRMLScene(slicer.mrmlScene)
place_widget.setCurrentNode(slicer.mrmlScene.AddNewNodeByClass("vtkMRMLMarkupsFiducialNode"))
place_widget.show()
```
Place a point and then you will see the default place widget does not have unset options shown by default.

image

Markups ToolBar has the unset options shown by default.
image

### DIFF
--- a/Modules/Loadable/Markups/Widgets/Testing/Python/MarkupsWidgetsSelfTest.py
+++ b/Modules/Loadable/Markups/Widgets/Testing/Python/MarkupsWidgetsSelfTest.py
@@ -171,10 +171,15 @@ class MarkupsWidgetsSelfTestTest(ScriptedLoadableModuleTest):
     placeWidget.deleteAllMarkupsOptionVisible = True
     self.assertTrue(placeWidget.deleteAllMarkupsOptionVisible)
 
-    placeWidget.deleteAllMarkupsOptionVisible = False
-    self.assertFalse(placeWidget.deleteAllMarkupsOptionVisible)
-    placeWidget.deleteAllMarkupsOptionVisible = True
-    self.assertTrue(placeWidget.deleteAllMarkupsOptionVisible)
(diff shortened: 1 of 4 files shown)


lassoan (commented):
Thank you, this is almost perfect. There is just one thing that would be nice to get fixed: if the place or delete button menu does not have any items then clicking on the down-arrow has no effect. It would be much better if in this case the down-arrow button would not appear at all.

lassoan:
> There is just one thing that would be nice to get fixed

I see that this is addressed in #6020.

jamesobutler:
Yes, not showing the delete button menu arrow button here when there is no option is included here as part of `updateDeleteButton`. Same type thing for the place button was done in #6020.

Issue #6011, opened 2021-11-10
Bad alloc error when loading an sequence on linux


lassoan:
Slicer displays a "Bad alloc" popup when loading or browsing a sequence. It prevents sequence replay because the error popup is displayed at each frame change.

The problem did not exist in Slicer-4.13.0-2021-10-06-linux-amd64 release, and appeared in Slicer-4.13.0-2021-10-07-linux-amd64. Since the only change that might have been related was the upgrade of the gcc compiler, the error is most likely caused by that.

When doing a local build with gcc 9.3.0 the problem does not occur.

## Steps to reproduce

- Go to Sample Data module
- Load "CT Cardio Sequence"
- "Application error" popup is displayed with "Exception thrown in event: std::bad_alloc" message.

## Expected behavior

No error popup should be displayed.

## Environment
- Slicer version: Slicer-4.13 2021-11-10
- Operating system: Ubuntu 20.04

lassoan:
Attaching gdb and catching thrown exceptions resulted in this call stack:

```
Thread 1 "SlicerApp-real" hit Catchpoint 1 (exception thrown), 0x00007f34f0e42672 in __cxa_throw () from /lib/x86_64-linux-gnu/libstdc++.so.6
(gdb) trace
Note: breakpoint 1 also set at pc 0x7f34f0e42672.
Tracepoint 2 at 0x7f34f0e42672
(gdb) bt
#0  0x00007f34f0e42672 in __cxa_throw () from /lib/x86_64-linux-gnu/libstdc++.so.6
#1  0x00007f34f0e39173 in std::__throw_bad_alloc() () from /lib/x86_64-linux-gnu/libstdc++.so.6
#2  0x00007f34f8aba711 in bool std::__detail::__regex_algo_impl<__gnu_cxx::__normal_iterator<char const*, std::string>, std::allocator<std::sub_match<__gnu_cxx::__normal_iterator<char const*, std::string> > >, char, std::regex_traits<char>, (std::__detail::_RegexExecutorPolicy)0, false>(__gnu_cxx::__normal_iterator<char const*, std::string>, __gnu_cxx::__normal_iterator<char const*, std::string>, std::match_results<__gnu_cxx::__normal_iterator<char const*, std::string>, std::allocator<std::sub_match<__gnu_cxx::__normal_iterator<char const*, std::string> > > >&, std::basic_regex<char, std::regex_traits<char> > const&, std::regex_constants::match_flag_type) ()
[...]

PR #6024, opened 2021-11-12
BUG: Fix bad alloc error from sequence string validation on Linux

Sunderlandkyl:
std::regex was causing a bad alloc error with the updated gcc compiler.
To fix the error, the regex used for format string validation is switched from std::regex to kwsys.

Re: #6011


lassoan (approved):
This looks perfect, thank you. I'll merge this and we'll see tomorrow if it fixes the error as expected.

Issue #5014, opened 2020-06-25
Update Python version from 3.6.7 to 3.9.10


jcfr:
This was initially discussed in https://github.com/Slicer/Slicer/issues/4859

The following PR would need to be finalized:

- <s>[ ] python-cmake-buildsystem/python-cmake-buildsystem#264</s>
- [x] python-cmake-buildsystem/python-cmake-buildsystem#267
- [x] When Slicer's Python is upgraded then post a comment to MONAILabel issue https://github.com/Project-MONAI/MONAILabel/pull/378 (so that they can remove Python-3.6 support from MONAILabel)

jamesobutler:
Python packages that have dropped Python 3.6 support:
- Numpy >=1.20.0
- Scipy >=1.6.0

We won't be able to upgrade to these new versions until the Python version included in Slicer is upgraded.

jamesobutler:
Additionally the following python packages not included in Slicer by default have dropped Python 3.6 support (mostly per [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html)):
- scikit-image >=0.18.0
- Matplotlib >= 3.4.0
- IPython >= 7.17.0
- pandas >= 1.2.0

jcfr:
> The following PR would need to be finalized

Will have a look in early January

lassoan:
Thank you @jamesobutler for helping with this. I've already started to encounter issues with lack of packages for Python-3.6, so it is time to move on.

lassoan:
Python-3.6 is starting to become a problem.

For example, we had to stop updating SlicerJupyter, because recent xeus-python versions require Python-3.7. (see https://github.com/Slicer/SlicerJupyter/issues/62)

lassoan:
Yet another important reason to update Python in Slicer ASAP: https://github.com/Project-MONAI/MONAILabel/pull/378

jamesobutler:
GitPython >= 3.1.22 has dropped support for Python 3.6 as well

jamesobutler:
Note that per [NEP29](https://numpy.org/neps/nep-0029-deprecation_policy.html#) projects have begun to drop Python 3.7 support. 

For example the following projects have dropped Python 3.7 support already for their upcoming releases.
- Numpy >= 1.22.0
- Scipy >= 1.8.0

Also to note that Python 3.10 has been released as of Oct. 4th 2021

fbordignon:
[...]

PR #6026, opened 2021-11-15
Update Slicer python to 3.9.9

fbordignon:
Most part of the work was done on project python-cmake-buildsystem. I've opened a [PR](https://github.com/python-cmake-buildsystem/python-cmake-buildsystem/pull/282) over there as well. This current PR is linking to the commit on the tip of the branch over there. I've based my work on a branch by colleagues including @jcfr @dbrnz Dan Dees. Thank you. 

Updating to 3.9 was preferred to 3.10 as discussed on #5014 

Summary of changes:

1. Added libffi as a dependency to build python and use it as an external lib. We need to check if it works on other environments as well. It should, because I've only configured to set the include dirs and lib variables. It is hosted on my personal github for now. I don't mind keeping it there forever, but if you think we should move to anywhere else feel free to suggest modifications. ffi is needed by ctypes which dropped the sources of ffi sometime in the past few years.
2. Added a script to set the dll search dirs, as of python > 3.8  does not search the PATH anymore for dlls, so we need to explicitly call os.add_dll_directory on windows. I've done that by installing a script at python-install/bin and calling this script every time there is the need to install libs. Also the applauncher sets the PYTHONSTARTUP as the mentioned script and sets the LibraryPaths as an environment variable to be passed to python so the script knows what to add to the dll search path.
3. Updated python packages
4. Updated python version wherever I could find a reference to 3.6.x, 3.6 or 36
5. Fixed a bug on defining the zlib include dirs and lib path

### DIFF
--- a/SuperBuild/External_python-dicom-requirements.cmake
+++ b/SuperBuild/External_python-dicom-requirements.cmake
@@ -41,24 +41,24 @@ if(NOT Slicer_USE_SYSTEM_${proj})
   set(requirements_file ${CMAKE_BINARY_DIR}/${proj}-requirements.txt)
   file(WRITE ${requirements_file} [===[
   # [pydicom]
-  pydicom==2.2.1 --hash=sha256:444b5b7289135ff5ea76dfc69d3597dcfde1cd050ca387f709d777f35701242d
+  pydicom==2.2.2 --hash=sha256:6ecb9c6d56a20b2104099b8ef8fe0f3664d797b08a0e0548fe0311b515b32308
   # [/pydicom]
-  # [Pillow]
-  # Hashes correspond to the following packages:
-  #  - Pillow-8.3.2-cp36-cp36m-macosx_10_10_x86_64.whl
-  #  - Pillow-8.3.2-cp36-cp36m-manylinux_2_17_aarch64.manylinux2014_aarch64.whl
-  #  - Pillow-8.3.2-cp36-cp36m-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
-  #  - Pillow-8.3.2-cp36-cp36m-manylinux_2_5_x86_64.manylinux1_x86_64.whl
-  #  - Pillow-8.3.2-cp36-cp36m-win_amd64.whl
-  Pillow==8.3.2 --hash=sha256:11eb7f98165d56042545c9e6db3ce394ed8b45089a67124298f0473b29cb60b2 \
-                --hash=sha256:2f23b2d3079522fdf3c09de6517f625f7a964f916c956527bed805ac043799b8 \
-                --hash=sha256:e5a31c07cea5edbaeb4bdba6f2b87db7d3dc0f446f379d907e51cc70ea375629 \
-                --hash=sha256:8f284dc1695caf71a74f24993b7c7473d77bc760be45f776a2c2f4e04c170550 \
-                --hash=sha256:a048dad5ed6ad1fad338c02c609b862dfaa921fcd065d747194a6805f91f2196
-  # [/Pillow]
   # [six]
   six==1.16.0 --hash=sha256:8abb2f1d86890a2dfb989f9a77cfcfd3e47c2a354b01111771326f8aa26e0254
   # [/six]
+  # [Pillow]
+  # Hashes correspond to the following packages:
+  #  - Pillow-8.4.0-cp39-cp39-macosx_10_10_x86_64.whl
+  #  - Pillow-8.4.0-cp39-cp39-macosx_11_0_arm64.whl
+  #  - Pillow-8.4.0-cp39-cp39-manylinux_2_17_aarch64.manylinux2014_aarch64.whl
+  #  - Pillow-8.4.0-cp39-cp39-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
+  #  - Pillow-8.4.0-cp39-cp39-win_amd64.whl
+  Pillow==8.4.0 --hash=sha256:5503c86916d27c2e101b7f71c2ae2cddba01a2cf55b8395b0255fd33fa4d1f1a \
+                --hash=sha256:4acc0985ddf39d1bc969a9220b51d94ed51695d455c228d8ac29fcdb25810e6e \
+                --hash=sha256:0b052a619a8bfcf26bd8b3f48f45283f9e977890263e4571f2393ed8898d331b \
+                --hash=sha256:b8831cb7332eda5dc89b21a7bce7ef6ad305548820595033a4b03cf3091235ed \
+                --hash=sha256:3eb1ce5f65908556c2d8685a8f0a6e989d887ec4057326f6c22b24e8a172c66b
+  # [/Pillow]
   # [retrying]
   retrying==1.3.3 --hash=sha256:08c039560a6da2fe4f2c426d0766e284d3b736e355f8dd24b37367b0bb41973b
   # [/retrying]
--- a/SuperBuild/External_python.cmake
+++ b/SuperBuild/External_python.cmake
@@ -9,6 +9,7 @@ if(NOT Slicer_USE_SYSTEM_python)
   list(APPEND ${proj}_DEPENDENCIES
     bzip2
     CTKAPPLAUNCHER
+    LibFFI
     LZMA
     zlib
     sqlite
@@ -50,8 +51,8 @@ if((NOT DEFINED PYTHON_INCLUDE_DIR
 
   set(python_SOURCE_DIR "${CMAKE_BINARY_DIR}/Python-${Slicer_REQUIRED_PYTHON_VERSION}")
 
-  set(_download_3.6.7_url "https://www.python.org/ftp/python/3.6.7/Python-3.6.7.tgz")
-  set(_download_3.6.7_md5 "c83551d83bf015134b4b2249213f3f85")
+  set(_download_3.9.10_url "https://www.python.org/ftp/python/3.9.10/Python-3.9.10.tgz")
+  set(_download_3.9.10_md5 "1440acb71471e2394befdb30b1a958d1")
 
   ExternalProject_Add(python-source
     URL ${_download_${Slicer_REQUIRED_PYTHON_VERSION}_url}
@@ -122,7 +123,7 @@ if((NOT DEFINED PYTHON_INCLUDE_DIR
 
   ExternalProject_SetIfNotDefined(
     Slicer_${proj}_GIT_TAG
-    "1feb43e4bb2f3461747487a85234bdd9839d44e3"
+    "bb45aa7a4cfc7a5a93bc490c6158f702d1a2226f"
     QUIET
     )
 
@@ -161,6 +162,8 @@ if((NOT DEFINED PYTHON_INCLUDE_DIR
       -DLZMA_LIBRARY:FILEPATH=${LZMA_LIBRARY}
       -DZLIB_INCLUDE_DIR:PATH=${ZLIB_INCLUDE_DIR}
       -DZLIB_LIBRARY:FILEPATH=${ZLIB_LIBRARY}
+      -DLibFFI_INCLUDE_DIR:PATH=${LibFFI_INCLUDE_DIR}
+      -DLibFFI_LIBRARY:FILEPATH=${LibFFI_LIBRARY}
       -DSQLite3_INCLUDE_DIR:PATH=${sqlite_INCLUDE_DIR}
       -DSQLite3_LIBRARY:FILEPATH=${sqlite_LIBRARY}
       -DENABLE_SSL:BOOL=${PYTHON_ENABLE_SSL}
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -175,7 +175,7 @@ mark_as_superbuild(Slicer_REQUIRED_QT_VERSION)
 # Python requirements
 #-----------------------------------------------------------------------------
 if(NOT DEFINED Slicer_REQUIRED_PYTHON_VERSION)
-  set(Slicer_REQUIRED_PYTHON_VERSION "3.6.7")
+  set(Slicer_REQUIRED_PYTHON_VERSION "3.9.10")
 endif()
 mark_as_superbuild(Slicer_REQUIRED_PYTHON_VERSION)
 
--- a/SuperBuild/External_LibFFI.cmake
+++ b/SuperBuild/External_LibFFI.cmake
@@ -1,0 +1,80 @@
+
+set(proj LibFFI)
+
+# Set dependency list
+set(${proj}_DEPENDENCIES "")
+
+# Include dependent projects if any
+ExternalProject_Include_Dependencies(${proj} PROJECT_VAR proj DEPENDS_VAR ${proj}_DEPENDENCIES)
+
+if(Slicer_USE_SYSTEM_${proj})
+  unset(LibFFI_INCLUDE_DIR CACHE)
+  find_path(LibFFI_INCLUDE_DIR ffi.h)
+
+  unset(LibFFI_LIBRARY CACHE)
+  find_library(LibFFI_LIBRARY NAMES ffi libffi)
+endif()
+
+# Sanity checks
+if(DEFINED ${proj}_DIR AND NOT EXISTS ${${proj}_DIR})
+  message(FATAL_ERROR "${proj}_DIR variable is defined but corresponds to nonexistent directory")
+endif()
+
+if(NOT DEFINED ${proj}_DIR AND NOT Slicer_USE_SYSTEM_${proj})
+
+  set(EP_SOURCE_DIR ${CMAKE_BINARY_DIR}/${proj})
+  set(EP_BINARY_DIR ${CMAKE_BINARY_DIR}/${proj}-build)
+  set(EP_INSTALL_DIR ${CMAKE_BINARY_DIR}/${proj}-install)
+
+  ExternalProject_SetIfNotDefined(
+    Slicer_${proj}_GIT_REPOSITORY
+    "${EP_GIT_PROTOCOL}://github.com/python-cmake-buildsystem/libffi.git"
+    QUIET
+    )
+
+  ExternalProject_SetIfNotDefined(
+    Slicer_${proj}_GIT_TAG
+    "cb4a78e39009646935badbeddb48d54c1fe97d63" # libffi-cmake-buildsystem-v3.4.2-2021-06-28-f9ea416
+    QUIET
+    )
+
+  ExternalProject_Add(${proj}
+    ${${proj}_EP_ARGS}
+    GIT_REPOSITORY "${Slicer_${proj}_GIT_REPOSITORY}"
+    GIT_TAG "${Slicer_${proj}_GIT_TAG}"
+    SOURCE_DIR ${EP_SOURCE_DIR}
+    BINARY_DIR ${EP_BINARY_DIR}
+    INSTALL_DIR ${EP_INSTALL_DIR}
+    CMAKE_CACHE_ARGS
+      # Compiler settings
+      -DCMAKE_C_COMPILER:FILEPATH=${CMAKE_C_COMPILER}
+      -DCMAKE_C_FLAGS:STRING=${ep_common_c_flags}
+      # Install directories
+      -DCMAKE_INSTALL_PREFIX:PATH=<INSTALL_DIR>
+    DEPENDS
+      ${${proj}_DEPENDENCIES}
+    )
+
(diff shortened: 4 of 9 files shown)


SuperBuild/External_python-dicom-requirements.cmake
```diff
@@ -40,41 +40,31 @@ endif()
 if(NOT Slicer_USE_SYSTEM_${proj})
   set(requirements_file ${CMAKE_BINARY_DIR}/${proj}-requirements.txt)
   file(WRITE ${requirements_file} [===[
-  # [pydicom]
-  pydicom==2.2.1 --hash=sha256:444b5b7289135ff5ea76dfc69d3597dcfde1cd050ca387f709d777f35701242d
-  # [/pydicom]
-  # [Pillow]
+  pydicom==2.2.2 --hash=sha256:6ecb9c6d56a20b2104099b8ef8fe0f3664d797b08a0e0548fe0311b515b32308
```
jamesobutler, line 43: Did you use the [python_package_updater.py](https://github.com/Slicer/Slicer/blob/c72d890515ccca70e4173dfcedd370e197c44ae9/Utilities/Scripts/python_package_updater.py) to update all the hashes here? That script relies on the entries in the generated requirements file to be within `# [pydicom]` and `# [/pydicom]` blocks and is used for maintenance purposes for updating hashes automatically.
https://github.com/Slicer/Slicer/blob/c72d890515ccca70e4173dfcedd370e197c44ae9/Utilities/Scripts/python_package_updater.py#L96-L97

fbordignon: I had some trouble running the script and ended up pasting the results by hand. I will try it again.

fbordignon:
Issue: The build was successful but the install steps for the python packages did not install anything 🤦‍♂️
I thought I was running the script and running the install command but I am only passing the install command as args to the script. I will work on it a little bit more and get back.

jcfr:
> I've opened a PR over there as well.

Thanks so much for working on this and for the detailed description :pray: :100: 

This is outstanding.

I will review and work on integrating the change into the `python-cmake-buildsystem` and also post comments here regarding the install rule and support for the new [add_dll_directory](https://docs.python.org/3/library/os.html#os.add_dll_directory)


Relevant porting notes:

> DLL dependencies for extension modules and DLLs loaded with [ `ctypes` ](https://docs.python.org/3/library/ctypes.html#module-ctypes) on Windows are now resolved more securely. Only the system paths, the directory containing the DLL or PYD file, and directories added with [ `add_dll_directory()` ](https://docs.python.org/3/library/os.html#os.add_dll_directory) are searched for load-time dependencies. Specifically, `PATH` and the current working directory are no longer used, and modifications to these will no longer have any effect on normal DLL resolution. If your application relies on these mechanisms, you should check for [ `add_dll_directory()` ](https://docs.python.org/3/library/os.html#os.add_dll_directory) and if it exists, use it to add your DLLs directory while loading your library. Note that Windows 7 users will need to ensure that Windows Update KB2533623 has been installed (this is also verified by the installer). (Contributed by Steve Dower in [bpo-36085](https://bugs.python.org/issue36085).)

Source: https://docs.python.org/3/whatsnew/3.8.html#bpo-36085-whatsnew

SuperBuild/External_python-dicom-requirements.cmake
fbordignon: I've reverted the compatibility with python_package_updater.py
The problem was that I needed python 3.9 to run and the script and get the hashes, but I was building it so I needed to change the script a little bit and ended up messing the blocks. good catch.

fbordignon:
Great! Thanks @jcfr 
I think I fixed the calls using the script I've created for the add_dll_directory
I will leave Slicer building from scratch to be sure. Tomorrow I will report back.

fbordignon:
I believe the build is working now. SimpleITK was having trouble finding the dll dirs but I've updated the environment script to add the LibrayPaths and it worked.
One thing that may be strange is that the script that I've created is being called for every platform but it checks if it is on windows before doing anything. Hence, the only collateral is having the package `sys` loaded after calling the script.

I don't know if the script `add_dll_directories_windows.py` is the best solution to the problem. Please feel free to criticize or suggest a better solution. I would be happy to implement and/or discuss alternatives.

fbordignon:
There was an issue with python3.dll because I had removed the .def usage for python 3.10 due to that it is not provided anymore on that version, but on 3.9 it is. I reverted back to using it and it is working fine!
I've deployed my company's customizations in python installing the following dependencies successfully:

<details>
  <summary markdown="span">requirements.txt</summary>

  ```c
  PySide2==5.15.2
  pyqtgraph
  cudnn-python-wrappers==1.0.0
  dlisio==0.3.5
  h5py>=2.10.0
  Jinja2==2.11.1
  natsort==6.2.0
  numpy>=1.18.1
  opencv-python-headless
  pint==0.11
  scikit-image>=0.17.2
  scikit-learn>=0.22.0
  scikit-optimize>=0.8.1
  scipy>=1.7.2
  tensorflow-gpu>=2.1.0
  tensorflow>=2.1.0
  recordtype==1.3.0
  edt>=2.0.1
  numba>=0.49.1
  matplotlib>=3.2.1
  lasio==0.25.1   
  pandas>=1.0.5
  camelot-py[cv]==0.8.2
  markdown2==2.3.9
  pyzmq>=19.0.0
  bokeh==2.3.0
  panel==0.11.0
  numexpr>=2.7.1
  plotly==4.14.3
  Flask==1.1.2
  cftime>=1.3.0
  detect_delimiter==0.1.1
  chardet==3.0.4
  paramiko>=2.7.2
  pytesseract==0.3.7
  netCDF4==1.5.4
  loguru>=0.5.3
  toolz==0.11.1
  xarray>=0.16.2
  zarr==2.5.0
  dask[complete]>=2.30.0
  dask-image>=0.4.0
  connected-components-3d>=3.2.1
  pathvalidate==2.5.0
  ```

</details>

lassoan:
Thank you for your continued work on this. Instead of removing that code (even if it was just a hack), it would be better to keep it there and only run it if Python version is 3.10. It will help when Python 3.10 support will be added.

fbordignon:
Ok, but in this case, for python 3.10 the hack was working to get the build to finish but the final python3.dll generated was not correct. We will have to look into this again for 3.10.

lassoan:
> Ok, but in this case, for python 3.10 the hack was working to get the build to finish but the final python3.dll generated was not correct. We will have to look into this again for 3.10.

OK, maybe just add a comment there then, describing what you did and how much it worked and what was the remaining issue. Or maybe just add a "Python-3.10 support" issue to the issue tracker and describe what you have learned there. Just to make sure that we don't need to restart investigating this problem from scratch when time comes for Python-3.10.

fbordignon:
I've built this branch with SlicerCAT and without it. Both cases are now building it from start to finish without any errors and generating the installer package. I've tested it only on Windows. Seems to be working fine. I did not run the automated tests. 

I believe we can start reviewing this PR and if someone can be kind enough to test it on Linux and Mac it would be perfect. I am available to discuss any issues and help fix them.

If you wish I can upload the installer package for testing.
Preview of what I am getting:
image
ps. paramiko is not built into Slicer but I've installed it with pip and imported it because I had problems in the past with python3.dll and wanted to test it.

lassoan:
I've tested the build on Ubuntu 20.4 (where current Slicer master build was successful). A completely clean build (starting from an empty build foder) failed with this error:

```
[ 49%] Building C object CMakeBuild/libpython/CMakeFiles/_freeze_importlib.dir/home/perklab/D/Slicer-SuperBuild-Release/Python-3.9/Python/getversion.c.o
[ 49%] Building C object CMakeBuild/libpython/CMakeFiles/_freeze_importlib.dir/home/perklab/D/Slicer-SuperBuild-Release/Python-3.9/Python/graminit.c.o
[ 50%] Building C object CMakeBuild/libpython/CMakeFiles/_freeze_importlib.dir/home/perklab/D/Slicer-SuperBuild-Release/Python-3.9/Python/import.c.o
[ 50%] Building C object CMakeBuild/libpython/CMakeFiles/_freeze_importlib.dir/home/perklab/D/Slicer-SuperBuild-Release/Python-3.9/Python/importdl.c.o
[ 50%] Building C object CMakeBuild/libpython/CMakeFiles/_freeze_importlib.dir/home/perklab/D/Slicer-SuperBuild-Release/Python-3.9/Python/initconfig.c.o
/home/perklab/D/Slicer-SuperBuild-Release/Python-3.9/Python/initconfig.c:28:4: error: #error "PLATLIBDIR macro must be defined"
   28 | #  error "PLATLIBDIR macro must be defined"
      |    ^~~~~
/home/perklab/D/Slicer-SuperBuild-Release/Python-3.9/Python/initconfig.c: In function ‘config_read’:
/home/perklab/D/Slicer-SuperBuild-Release/Python-3.9/Python/initconfig.c:1754:68: error: ‘PLATLIBDIR’ undeclared (first use in this function)
 1754 |         status = CONFIG_SET_BYTES_STR(config, &config->platlibdir, PLATLIBDIR,
      |                                                                    ^~~~~~~~~~
/home/perklab/D/Slicer-SuperBuild-Release/Python-3.9/Python/initconfig.c:764:49: note: in definition of macro ‘CONFIG_SET_BYTES_STR’
  764 |     config_set_bytes_string(config, config_str, str, "cannot decode " NAME)
      |                                                 ^~~
/home/perklab/D/Slicer-SuperBuild-Release/Python-3.9/Python/initconfig.c:1754:68: note: each undeclared identifier is reported only once for each function it appears in
 1754 |         status = CONFIG_SET_BYTES_STR(config, &config->platlibdir, PLATLIBDIR,
      |                                                                    ^~~~~~~~~~
/home/perklab/D/Slicer-SuperBuild-Release/Python-3.9/Python/initconfig.c:764:49: note: in definition of macro ‘CONFIG_SET_BYTES_STR’
  764 |     config_set_bytes_string(config, config_str, str, "cannot decode " NAME)
      |                                                 ^~~
[ 50%] Building C object CMakeBuild/libpython/CMakeFiles/_freeze_importlib.dir/home/perklab/D/Slicer-SuperBuild-Release/Python-3.9/Python/marshal.c.o
[ 50%] Building C object CMakeBuild/libpython/CMakeFiles/_freeze_importlib.dir/home/perklab/D/Slicer-SuperBuild-Release/Python-3.9/Python/modsupport.c.o
make[5]: *** [CMakeBuild/libpython/CMakeFiles/_freeze_importlib.dir/build.make:1311: CMakeBuild/libpython/CMakeFiles/_freeze_importlib.dir/home/perklab/D/Slicer-SuperBuild-Release/Python-3.9/Python/initconfig.c.o] Error 1
make[5]: *** Waiting for unfinished jobs....
[ 51%] Linking C shared library ../../../lib/python3.9/lib-dynload/_decimal.so
[ 51%] Built target extension_decimal
make[4]: *** [CMakeFiles/Makefile2:3992: CMakeBuild/libpython/CMakeFiles/_freeze_importlib.dir/all] Error 2
make[3]: *** [Makefile:141: all] Error 2
make[2]: *** [CMakeFiles/python.dir/build.make:121: python-prefix/src/python-stamp/python-build] Error 2
make[1]: *** [CMakeFiles/Makefile2:1180: CMakeFiles/python.dir/all] Error 2
make: *** [Makefile:95: all] Error 2
perklab@perklabseg:~/D$ 
```

fbordignon:
Thanks for the test. The circleci build is configured and issuing the same
error.

Em sáb., 20 de nov. de 2021 03:58, Andras Lasso ***@***.***>
escreveu:

lassoan:
You can test this easily, too, on a Windows PC. Install Ubuntu from the Microsoft store and follow the Slicer Linux build instructions (much simpler than on Windows, just copy-paste a couple of lines of code to the console).

fbordignon:
Question: Why don't we use python as a CMAKE external project and use the official build scripts with Makefile and msvc solution?

lassoan:
> Question: Why don't we use python as a CMAKE external project and use the official build scripts with Makefile and msvc solution?

See a few reasons here: https://github.com/python-cmake-buildsystem/python-cmake-buildsystem#overview

Have a look at the current official CPython build for windows. It is a mix of VS project files, batch files, and python scripts. Something like 100 files. Just for a single operating system and compiler. It is an embarrassment. The main reason seems to be that at least some CPython core developers hated (most likely just did not know) CMake. Also, the current mess works well enough for many people and it remains less risky to just keep patching it rather than replacing it completely.

fbordignon:
Hi all. I've successfully built this on Windows. On Linux there is a patch that I needed to apply to PythonQt to make it build successfully.
I've commented lines [37, 38 and 39](https://github.com/MeVisLab/pythonqt/blob/fb16d6e5a773be521ff5dc2b15932a5ccaa0a223/src/PythonQtPythonInclude.h#L37) of PythonQtPythonInclude.h.
```c++
#undef _POSIX_C_SOURCE
#undef _POSIX_THREADS
#undef _XOPEN_SOURCE
```
For some reason those defines are now needed when importing Python.h and they are not defined again after the undefs. Should we open an issue on PythonQt?
I've updated to python 3.9.9 as @jamesobutler suggested because it only patches a few files from 3.9.8. The build on python-cmake-buildsystem is working fine with a few tests failing, which already failed for python 3.6 as well. So I figured it is not a big deal.

pieper:
> Should we open an issue on PythonQt?

In general yes, if we think it's a legitimate issue that needs fixing better to get in upstream.

fbordignon:
> > Should we open an issue on PythonQt?
> 
> In general yes, if we think it's a legitimate issue that needs fixing better to get in upstream.

My bad, I was misunderstanding the configure_file function from cmake. Now the needed symbols are defined on CMake and it is building on linux as well.

fbordignon:
I've been testing Slicer with Python 3.9.9 on Windows for the last few days and did not find relevant problems, therefore I don't have any pending issues for this PR. I am available to discuss/solve issues that arise at the code review.

fbordignon:
Note that the ci build is failing because it is not building python. It is trying to use a prebuilt 3.6 and fails at the end while packaging.

jcfr:
Update regarding https://github.com/python-cmake-buildsystem/python-cmake-buildsystem:
* Python 3.7.x support has been added and integrated
* Integration of python 3.8.x and 3.9.x support are being finalized

jcfr:
>  Integration of python 3.8.x and 3.9.x support are being finalized

Support for 3.8.x and 3.9.x has been integrated upstream. 

I will soon focus on updating Slicer :rocket:

fbordignon:
Thanks @jcfr I am compiling Slicer right now with the https://github.com/python-cmake-buildsystem/python-cmake-buildsystem/commit/0b5ae7112feb9b69e1e6c49f93e91ca692acaad6
Not finished yet, but seems to be working. 

One issue that may be revisited is the issue with the add_dll_directory:

> I will review and work on integrating the change into the `python-cmake-buildsystem` and also post comments here regarding the install rule and support for the new [add_dll_directory](https://docs.python.org/3/library/os.html#os.add_dll_directory)
> 
> Relevant porting notes:
> 
> > DLL dependencies for extension modules and DLLs loaded with [ `ctypes` ](https://docs.python.org/3/library/ctypes.html#module-ctypes) on Windows are now resolved more securely. Only the system paths, the directory containing the DLL or PYD file, and directories added with [ `add_dll_directory()` ](https://docs.python.org/3/library/os.html#os.add_dll_directory) are searched for load-time dependencies. Specifically, `PATH` and the current working directory are no longer used, and modifications to these will no longer have any effect on normal DLL resolution. If your application relies on these mechanisms, you should check for [ `add_dll_directory()` ](https://docs.python.org/3/library/os.html#os.add_dll_directory) and if it exists, use it to add your DLLs directory while loading your library. Note that Windows 7 users will need to ensure that Windows Update KB2533623 has been installed (this is also verified by the installer). (Contributed by Steve Dower in [bpo-36085](https://bugs.python.org/issue36085).)
> 
> Source: https://docs.python.org/3/whatsnew/3.8.html#bpo-36085-whatsnew

jcfr:
> the issue with the add_dll_directory

Indeed, I will be reviewing and integrating all the work you have done :pray: :100:

fbordignon:
4b6f700 points to the latest master commit of python cmake and fixes include and libdirs for zlib and ffi.
Also some delimiters for python packages that I messed up earlier and fixed now (setuptools and pip).
I've built on windows with those changes. Looks good

SuperBuild/External_python.cmake
```diff
@@ -87,9 +91,9 @@ if((NOT DEFINED PYTHON_INCLUDE_DIR
   if(WIN32)
     list(APPEND EXTERNAL_PROJECT_OPTIONAL_CMAKE_ARGS
         # Depends on libbz2
-        -DBUILTIN_BZ2:BOOL=OFF
+        -DBUILTIN_BZ2:BOOL=ON
```
jcfr, line 94: Comment above indicates the following:

>   Force modules that statically link to zlib or libbz2 to not be built-in.
>   Otherwise, when building in Debug configuration, the Python library--which
>   we force to build in Release configuration--would mix Debug and Release C
>   runtime libraries.

Unless we confirm it is not an issue anymore, we will likely have to build them as shared.

SuperBuild/External_ffi.cmake
```diff
@@ -0,0 +1,86 @@
+
+set(proj ffi)
```
jcfr, line 2: for consistency, I will likely rename the project to `libffi`. See http://sourceware.org/libffi/

SuperBuild/External_ffi.cmake
```diff
@@ -0,0 +1,86 @@
...
+
+if(NOT DEFINED ffi_DIR AND NOT Slicer_USE_SYSTEM_${proj})
+
+  set(EP_SOURCE_DIR ${CMAKE_BINARY_DIR}/${proj})
+  set(EP_BINARY_DIR ${CMAKE_BINARY_DIR}/${proj}-build)
+  set(EP_INSTALL_DIR ${CMAKE_BINARY_DIR}/${proj}-install)
+
+  ExternalProject_SetIfNotDefined(
+    Slicer_${proj}_GIT_REPOSITORY
+    "${EP_GIT_PROTOCOL}://github.com/fbordignon/libffi.git"
```
jcfr, line 30: Before integrating, we will create a fork of `libffi/libffi` into the Slicer organization that will:
* have a default `welcome` branch describing the fork and available branch
* history of attempt to contribute CMake support to libffi
* describe the naming convention for branches

It will be similar to what we have for these projects:
https://github.com/search?q=topic%3A3d-slicer-fork+org%3ASlicer+fork%3Atrue

SuperBuild/External_ffi.cmake
jcfr: Will also probably renamed the variables in `python-cmake-buildsystem` to have the prefix `LIBFFI_`

SuperBuild/External_python.cmake
jcfr: I confirmed that keeping the original option still allow to compile CPython and import `binascii` and `bz2` modules. I am then reverting this changes back to their original value.

lassoan:
This is huge! Thank you @fbordignon @jcfr @jamesobutler and everyone else for making this happen.

fbordignon:
Thank you guys, this is awesome!

jamesobutler:
SimpleITK build passes on Windows now as of https://github.com/Slicer/Slicer/commit/a1c44fab014a6be93e5917fe63ffee3a98dba597. This was an incremental build after it first failed.

However, currently observing issues on launch of Slicer related to VTK. @jcfr Are you observing the same thing? I don't think doing a clean build would resolve this issue.
```
  File "<string>", line 5, in <module>
  File "<string>", line 6, in <module>
  File "C:\S5R\VTK-build\bin\Lib\site-packages\vtk.py", line 30, in <module>
    all_m = importlib.import_module('vtkmodules.all')
  File "C:\S5R\python-install\Lib\importlib\__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "C:\S5R\VTK-build\bin\Lib\site-packages\vtkmodules\all.py", line 25, in <module>
    from .vtkViewsInfovis import *
ImportError: DLL load failed while importing vtkViewsInfovis: The specified module could not be found.
```

jcfr:
> I don't think doing a clean build would resolve this issue

Indeed, I will make sure the `sitecustomize` module responsible to call `slicer_dll_directories.add()` is also imported when embedding python in the app. It looks like I missed something when integrating the changes

jcfr:
> This error causes multiple Slicer modules to not load properly (all scripted loadable?).

This should be fixed in https://github.com/Slicer/Slicer/pull/6138

PR #6027, opened 2021-11-17
BUG: Fix erase when using the selected segment as mask

Sunderlandkyl:
Erasing a single segment would "add-back" the mask segment. When the selected segment and mask segment were the same, this would result in nothing changing. Fixed by only adding the mask segment back if it is not the selected segment.

See discussion: https://discourse.slicer.org/t/erase-using-a-segment-as-only-editable-area/20519

Also adds tests to SegmentationsModuleTest2.py to detect future regressions in masking behavior.

### DIFF
--- a/Modules/Loadable/Segmentations/EditorEffects/qSlicerSegmentEditorAbstractEffect.cxx
+++ b/Modules/Loadable/Segmentations/EditorEffects/qSlicerSegmentEditorAbstractEffect.cxx
@@ -353,9 +353,6 @@ void qSlicerSegmentEditorAbstractEffect::modifySegmentByLabelmap(vtkMRMLSegmenta
       vtkOrientedImageDataResample::ModifyImage(maskImage, thresholdMask, vtkOrientedImageDataResample::OPERATION_MAXIMUM);
       }
 
-    vtkNew<vtkOrientedImageData> maskWithCurrentSegment;
-    maskWithCurrentSegment->DeepCopy(maskImage);
-
     vtkSmartPointer<vtkOrientedImageData> segmentLayerLabelmap =
       vtkOrientedImageData::SafeDownCast(segment->GetRepresentation(segmentationNode->GetSegmentation()->GetMasterRepresentationName()));
     if (segmentLayerLabelmap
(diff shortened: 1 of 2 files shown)


Modules/Loadable/Segmentations/EditorEffects/qSlicerSegmentEditorAbstractEffect.cxx
```diff
@@ -300,7 +300,10 @@ void qSlicerSegmentEditorAbstractEffect::modifySegmentByLabelmap(vtkMRMLSegmenta
   // Prevent disappearing and reappearing of 3D representation during update
   SlicerRenderBlocker renderBlocker;
 
-  vtkSmartPointer<vtkOrientedImageData> modifierLabelmap = modifierLabelmapInput;
+  // Prevent the original modifier labelmap from being modified
+  vtkNew<vtkOrientedImageData> modifierLabelmap;
```
lassoan, line 304: Copying the modifier labelmap could might require significant extra memory and maybe even cause perceivable amount of delay if the the modifier labelmap is large. Could we make the copy only when it is necessary (e.g., mask is the same as the selected segment)?

lassoan (commented):
Thanks a lot for the fix. Adding the tests is very useful, too, as the code has become very complex over time. I've just added one comment about possible performance improvement.

Modules/Loadable/Segmentations/EditorEffects/qSlicerSegmentEditorAbstractEffect.cxx
Sunderlandkyl: I think that a deep copy is necessary any time that we want to use any sort of masking, as the mask is applied directly to the modifier labelmap. The alternatives would be to document that the input modifier labelmap can be changed by the function, or to change the way that masking works to that the modifier labelmap is not changed. In any case, I've moved the DeepCopy call so that it won't be called if no masking options are enabled.

lassoan: > document that the input modifier labelmap can be changed by the function

This would be OK, but probably developers would not expect it and it might cause problems.

If the modifier is only copied when masking is enabled then I think that should be good enough (there is some overhead due to masking anyway, so one extra buffer and a copy would probably not be noticeable).

lassoan (approved):
Thank you, this should work well.

PR #6029, opened 2021-11-17
BUG: Fix Segment table update for Segmentation replay in Sequences

### LINKED ISSUE #5953: Segmentation Sequences Can't handle segment being deleted
## Steps to reproduce

1. Load a sequence volume
2. Create a segmentation on the first frame of the volume with one segment
3. Go to the sequences module, create a new sequence and set the segmentation as its proxy node, check save changes
image
4. change to the second frame and edit the segment there
image
5. delete the segment, the segment is removed from the segments table as expected. 
image
But if you scroll to the first frame the segment is still not in the segments table even though it shows up in the sliceviews
image
6. Click Add in the segmentation module, no segment is created, this shows up in console:
`class QStandardItem *__cdecl qMRMLSegmentsModelPrivate::insertSegment(class QString,int) : Item mismatch when inserting segment item with ID  "Segment_2"`

## Expected behavior

The segment table updates to show the segment on frames which have the segment, and no segment on frames without the segment.

## Environment
- Slicer version: Slicer-4.13.0-2021-09-14

Sunderlandkyl:
When vtkMRMLSegmentationNode were copied from recorded sequences, no events were emitted that would cause the table to update. This commit adds SegmentationChangedEvent to vtkMRMLSegmentationNode, which is triggered whenever the vtkSegmentation object is changed. If the event is observed by qMRMLSegmentsModel, it will trigger a reset of the model.

This commit also makes some changes to qMRMLSegmentsTableView and qMRMLSegmentEditorWidget to ensure that the widget states are consistent during Segmentation node Sequence replay.

Fixes #5953

### DIFF
--- a/Libs/MRML/Core/vtkMRMLSegmentationNode.h
+++ b/Libs/MRML/Core/vtkMRMLSegmentationNode.h
@@ -291,9 +291,11 @@ public:
   vtkMRMLColorTableNode* GetLabelmapConversionColorTableNode();
 
   /// ReferenceImageGeometryChangedEvent is fired when the ReferenceImageGeometry node reference is Added, Modified, or Removed
+  /// SegmentationChangedEvent is fired when a different vtkSegmentation object is set into the node.
   enum
   {
-    ReferenceImageGeometryChangedEvent = 23000
+    ReferenceImageGeometryChangedEvent = 23000,
+    SegmentationChangedEvent
   };
 
 protected:
--- a/Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.cxx
+++ b/Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.cxx
@@ -1270,9 +1270,16 @@ void qMRMLSegmentEditorWidget::updateWidgetFromMRML()
   d->SurfaceSmoothingSlider->blockSignals(wasBlocked);
 
   QString selectedSegmentID;
-  if (d->ParameterSetNode->GetSelectedSegmentID())
+  if (d->ParameterSetNode->GetSelectedSegmentID() && strcmp(d->ParameterSetNode->GetSelectedSegmentID(), "") != 0)
     {
     selectedSegmentID = QString(d->ParameterSetNode->GetSelectedSegmentID());
+
+    // Check if selected segment ID is invalid.
+    if (!d->SegmentationNode->GetSegmentation()
+      || d->SegmentationNode->GetSegmentation()->GetSegmentIndex(d->ParameterSetNode->GetSelectedSegmentID()) < 0)
+      {
+      selectedSegmentID.clear();
+      }
     }
 
   // Disable adding new segments until master volume is set (or reference geometry is specified for the segmentation).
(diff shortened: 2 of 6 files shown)


Libs/MRML/Core/vtkMRMLSegmentationNode.h
```diff
@@ -291,9 +291,11 @@ class VTK_MRML_EXPORT vtkMRMLSegmentationNode : public vtkMRMLDisplayableNode
   vtkMRMLColorTableNode* GetLabelmapConversionColorTableNode();
 
   /// ReferenceImageGeometryChangedEvent is fired when the ReferenceImageGeometry node reference is Added, Modified, or Removed
+  /// SegmentationChangedEvent is fired when the vtkSegmentation object is changed
```
lassoan, line 294:
```suggestion
  /// SegmentationChangedEvent is fired when a different vtkSegmentation object is set into the node.
```

Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.cxx
```diff
@@ -1270,9 +1270,16 @@ void qMRMLSegmentEditorWidget::updateWidgetFromMRML()
   d->SurfaceSmoothingSlider->blockSignals(wasBlocked);
 
   QString selectedSegmentID;
-  if (d->ParameterSetNode->GetSelectedSegmentID())
+  if (d->ParameterSetNode->GetSelectedSegmentID() && strcmp(d->ParameterSetNode->GetSelectedSegmentID(), "") != 0)
     {
     selectedSegmentID = QString(d->ParameterSetNode->GetSelectedSegmentID());
+
+    // Selected segment ID is not valid.
```
lassoan, line 1277: Rephrase (`Check if ... is invalid`) or move the comment inside the branch (just above `selectedSegmentID.clear();`)

Modules/Loadable/Segmentations/Widgets/qMRMLSegmentsModel.cxx
```diff
@@ -241,6 +241,7 @@ void qMRMLSegmentsModel::setSegmentationNode(vtkMRMLSegmentationNode* segmentati
     d->SegmentationNode->AddObserver(vtkSegmentation::SegmentModified, d->CallBack, -10.0);
     d->SegmentationNode->AddObserver(vtkSegmentation::SegmentsOrderModified, d->CallBack, -15.0);
     d->SegmentationNode->AddObserver(vtkMRMLDisplayableNode::DisplayModifiedEvent, d->CallBack, -15.0);
+    d->SegmentationNode->AddObserver(vtkMRMLSegmentationNode::SegmentationChangedEvent, d->CallBack, 10.0);
```
lassoan, line 244: Could you add a comment on why we cannot use default priority? (it would be nice to document it for the other events as well)

Modules/Loadable/Segmentations/Widgets/qMRMLSegmentsTableView.cxx
```diff
@@ -740,13 +748,20 @@ void qMRMLSegmentsTableView::setSelectedSegmentIDs(QStringList segmentIDs)
...
+    validSelection = true;
     QItemSelectionModel::QItemSelectionModel::SelectionFlags flags = QFlags<QItemSelectionModel::SelectionFlag>();
     flags.setFlag(itemSelectionFlag);
     flags.setFlag(QItemSelectionModel::Rows);
     d->SegmentsTable->selectionModel()->select(index, flags);
     // Afther the first segment, we append to the current selection
     itemSelectionFlag = QItemSelectionModel::Select;
     }
+
+  // The list of segment IDs was either empty, or all IDs were invalid.
```
lassoan, line 760: Similar as above. If I just read this comment then it suggest that at this point, the list of segment IDs is empty. To make it very clear, either rephrase ("Check if...") or if we keep it as a statement then move it to the same line as the `if` or inside the branch.

lassoan (commented):
Thank you, it looks good! I just added a few trivial comments inline.

Modules/Loadable/Segmentations/Widgets/qMRMLSegmentsModel.cxx
Sunderlandkyl: Default priority is fine in this case. I'm not sure if or why it is required for the other events.

lassoan (approved):
Great, thank you!

PR #6030, opened 2021-11-18
BUG: Fix markups toolbar node selector not using specific node label

jamesobutler:
The markups toolbar node selector was not getting a specific nodeTypeLabel from GetTagByClassName so it was just using the fallback "node" term.

| Current | This PR |
|----------|----------|
| image| image|

It is weird that `GetTagByClassName("vtkMRMLMarkupsNode")` is returning an empty string during startup which is why it was proceeding to the end of `qMRMLNodeComboBox::nodeTypeLabel()` to then get the "node" term. I would have expected it to get the node tag name of "Markups" (as seen below). I'm assuming it has something to do with the creation of the Markups ToolBar prior to the markups nodes being registered where it just doesn't know what "vtkMRMLMarkupsNode" is yet. Maybe something weird with "Volumes" is similar which is why there is that same type special case of "Volume" specified at the end of `nodeTypeLabel()`.
https://github.com/Slicer/Slicer/blob/19ed4f21a6bc3b17b418d49174080b4dea49908f/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsNode.h#L146-L147

I'm not using what GetNodeTagName would return unlike other nodes as it would produce statements such as "Select a Markups" which is weird since  "Markups" is a plural term and "Select a" expects to have a singular term like "Markup" to follow it.

### DIFF
--- a/Libs/MRML/Core/vtkMRMLNode.h
+++ b/Libs/MRML/Core/vtkMRMLNode.h
@@ -325,6 +325,14 @@ public:
   /// \note Subclasses should implement this method.
   virtual const char* GetNodeTagName() = 0;
 
+  /// Get node type display name (like "Closed Curve", "Markup", etc).
+  ///
+  /// \note Subclasses should override this method to provide a more appropriate and translatable name.
+  virtual const char* GetTypeDisplayName()
+  {
+    return this->GetNodeTagName();
+  }
+
   /// \brief Set a name value pair attribute.
   ///
   /// Fires a vtkCommand::ModifiedEvent.
--- a/Libs/MRML/Core/vtkMRMLScene.cxx
+++ b/Libs/MRML/Core/vtkMRMLScene.cxx
@@ -640,6 +640,24 @@ const char* vtkMRMLScene::GetTagByClassName(const char *className)
 }
 
 //------------------------------------------------------------------------------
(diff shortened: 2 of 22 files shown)


lassoan:
Tag name is a machine readable name, not translated, so qMRMLNodeComboBox should not use it for display.

What do you think about generalizing "GetMarkupTypeDisplayName"? We could add a "GetTypeDisplayName" to the vtkMRMLNode.h, which would call GetNodeTagName() by default, but could be overridden to return a more appropriate name.

Markups is a prefix that means that means "related to Markups module" and does not refer to multiple objects. I find the "markups" prefix hard to read (it is too easy to interpret it as something referring to multiple objects), but unfortunately this was how the node name was introduced many years ago and we kept consistency using it in the API. It would be hard to move to "markup" in source code, but should be fine to use "markup" in strings displayed to users.

lassoan:
I've fixed the initialization issue and implemented a general solution for specifying node type display names for abstract nodes. @jamesobutler please have a look.

lassoan (commented):
I've made some design improvements and fixed he noneDisplay text update issue. It should be good to go, but any feedback is welcome.

jamesobutler:
Thanks for helping improve this 🙏🏻 ! Much appreciated. I have built these latest updates and can confirm that the Markups toolbar node selector correctly says "Select a Markup" at startup. ✔️

PR #6032, opened 2021-11-18
BUG: Unblock segment table signals when no Segmentation node is set

Sunderlandkyl:
If no Segmentation node was set, endResetModel() was not called. This left the table signals blocked, since qMRMLSegmentsTableView::modelReset() was not triggered.


lassoan (approved):
Looks good, thank you.

PR #6033, opened 2021-11-18
ENH: Allow using markups ROI for specifying segment geometry

### LINKED ISSUE #5934: Allow using markups ROI for specifying segment geometry
In latest Slicer master, annotation ROI nodes are much harder to access, therefore we need to make sure that markups ROI can be used everywhere where annotation ROI nodes. One of the last missing places is the "Specify geometry" feature in Segment Editor (button next to the Master volume selector).

Allow specifying segment geometry using Markups ROI node. Orientation of the ROI should be used to set the internal labelmap's orientation.

## Additional context

[Add any other context or screenshots about the feature request here.]

Sunderlandkyl:
Fixes #5934


lassoan (approved):
Thank you, the implementation is nice and simple and it works beautifully, even for complicated cases, such as transformed segmentation and markups ROI.

PR #6034, opened 2021-11-19
BUG: Fix Markups ROI visibility when all control points are removed

### LINKED ISSUE #5898: Cannot delete ROI using markups place widget
Cannot delete ROI using markups place widget

## Steps to reproduce

Create a markups place widget, use it to create a markups ROI. When pressing the delete button, the control point is deleted but the ROI remains intact. 

## Expected behavior

It should be possible to reset a ROI using markups place widget.

Maybe the "Delete all" button should not just delete the control points but reset the widget. There should be a way to make the trashcan icon call this reset method (without using button menu). 

## Environment
- Slicer version: Slicer-4.13.0-2021-09-24
- Operating system: Windows

Sunderlandkyl:
When all control points were removed from the ROI node, the ROI would remain visible. In addition, attempting to re-place the ROI would only allow one point to be placed.

Fixed visibility issues by disabling the ROI actors if there are no control points.
Fixed the re-placement issue by changing the maximum number of control points from -1 to 2 for "Box" type ROI when all control points were removed.

Fixes #5898


lassoan (approved):
Works great, thank you.

Issue #5989, opened 2021-11-04
Update Markups API to make consistent name change Fiducial to Point List


jamesobutler:
As completed in https://github.com/Slicer/Slicer/pull/5986, a vtkMRMLMarkupsFiducialNode was updated from being displayed as a "Fiducial" to instead by displayed as a "Point List". Things are now confusing for developers that to create a "Point List", one must call something like `slicer.mrmlScene.AddNewNodeByClass("vtkMRMLMarkupsFiducialNode")`.

As suggested by @jcfr in https://github.com/Slicer/Slicer/pull/5986#issuecomment-960305078

> to ensure naming is consistent with the developer API:
> 
> * Rename `vtkMRMLMarkupsFiducialNode` to `vtkMRMLMarkupsPointListNode`
> * Deprecate the methods with `Fiducial` and add equivalent ones with `Point` in the name (if needed)
> * Call `RegisterNodeClass` with former tag name

There should also be updates to the documentation and examples that use newer versions of the API and with the latest terminology of "Point List" instead of "Fiducial". See example below
https://github.com/Slicer/Slicer/blob/76d201be68b353530f09ee027e6b33b3f542ce5f/Docs/developer_guide/script_repository/markups.md#L11-L23

lassoan:
Node name is used as point name prefix, therefore using "PL" as default node name would result in point names PL-1, PL-2, ... ("point list 1", "point list 2"). Also, when a second node is added then point names would be PL_1-1, PL_1-2, PL_1-3..., which sound very complicated.

Therefore, it would be better to use "P" as default node name for point lists.

We could also consider changing how unique node names are generated for point lists. Instead of P, P_1, P_2, ... we could use P, Q, R, S, T U,... Z.

jamesobutler:
> Therefore, it would be better to use "P" as default node name for point lists.

Currently Markups Plane nodes use “P” as the default name so it would have to be changed if we have that letter to “Point List”.

smrolfe:
The PL_1-1, PL_1-2, PL_1-3 style of naming of points is also consistent with the way open and closed curve points are named.

lassoan:
@jamesobutler do you think there is anything left other than renaming vtkMRMLMarkupsFiducialNode to vtkMRMLMarkupsPointListNode?

[...]

PR #6035, opened 2021-11-19
ENH: Remove Slicer3 Fiducial List classes

jamesobutler:
`vtkMRMLFiducialListNode` was originally marked as deprecated in June 2012 (14d653a) which is over 9 years ago. It was stated that it was for backwards compatibility for Slicer 3 fiducial lists and that the Annotation module was to be used instead. Slicer 4 was originally released in November 2011.

This PR proposes the removal of the old `vtkMRMLFiducialListNode` classes following some research I posted about at https://github.com/Slicer/Slicer/issues/5989#issuecomment-973297681. To prevent adding more complexity to the Slicer core in regards to Fiducial/Point List functionality this PR trims the oldest usage. 

`vtkMRMLFiducialListNode` is legacy to `vtkMRMLAnnotationFiducialNode` which is legacy to `vtkMRMLMarkupsFiducialNode` which will potentially be legacy to `vtkMRMLMarkupsPointListNode` when it is created soon. Slicer 4.X can be used to load `vtkMRMLFiducialListNode` objects and if they want to be used in Slicer 5 they can be converted to a newer type using Slicer 4.X. Older releases of Slicer have been available using instructions at https://www.slicer.org/wiki/Documentation/Nightly/FAQ/General#Where_can_I_download_older_release_of_Slicer_.3F to https://slicer-packages.kitware.com/#collection/5f4474d0e1d8c75dfc70547e/folder/5f4474d0e1d8c75dfc705482.

Slicer builds normally on Windows and all currently passing tests are still passing.


lassoan (approved):
Agreed. Since this node was deprecated in Slicer4, we don't need to support it in Slicer5.

Issue #5664, opened 2021-05-31
Document all Slicer core modules on readthedocs


lassoan:
Documentation of some Slicer core modules are almost all on readthedocs (see https://github.com/Slicer/Slicer/pull/5662). These are the only remaining ones that need to be ported from the Slicer wiki/reviewed/rewritten:

@lassoan:
- [x] user_guide/modules/sequences
- [x] user_guide/modules/cropvolumesequences (added through https://github.com/Slicer/Slicer/pull/6656)
- [x] user_guide/modules/screencapture
- [x] user_guide/modules/welcome
- [x] user_guide/modules/sampledata
- [x] user_guide/modules/dataprobe
- [x] user_guide/modules/simplefilters (added through https://github.com/Slicer/Slicer/pull/6656)
- [x] user_guide/modules/multivolumeimporter (added through https://github.com/Slicer/Slicer/pull/6658)
- [x] user_guide/modules/multivolumeexplorer (added through https://github.com/Slicer/Slicer/pull/6658)
- [x] user_guide/modules/editor
- [x] user_guide/modules/datastore

@pieper:
- [x] user_guide/modules/landmarkregistration
- [x] user_guide/modules/dmriinstall
- [x] user_guide/modules/endoscopy
- [x] user_guide/modules/labelstatistics

@cpinter:
- [x] #6019

@Sunderlandkyl:
- [x] user_guide/modules/dynamicmodeler (added through https://github.com/Slicer/Slicer/pull/6640)
- [x] user_guide/modules/colors
- [x] user_guide/modules/markups (update) (added through https://github.com/Slicer/Slicer/pull/6643)

@jcfr or @sjh26:
- [x] user_guide/modules/eventbroker (added through ae2e56b493138eaa9d9082951f49bb3d693a2a9d)
- [x] user_guide/modules/extensionwizard (added through https://github.com/Slicer/Slicer/pull/6634)
- [x] user_guide/modules/vectortoscalarvolume (added through https://github.com/Slicer/Slicer/pull/6635)
- [x] user_guide/modules/performancetests (added through https://github.com/Slicer/Slicer/pull/6668)
- [x] user_guide/modules/selftests (added through https://github.com/Slicer/Slicer/pull/6669)

@RafaelPalomar:
- [x] user_guide/modules/cropvolume (added through https://github.com/Slicer/Slicer/pull/6663)
- [x] user_guide/modules/surfacetoolbox (added through https://github.com/Slicer/Slicer/pull/6666)

[...]

PR #6036, opened 2021-11-23
DOC: Remove Editor module documentation

jamesobutler:
re #5664

The Editor module was removed in https://github.com/Slicer/Slicer/commit/39283db420baf502fa99865c9d5d58d0e5295a6e. The Segment Editor module was the replacement.

### DIFF
--- a/Docs/developer_guide/script_repository/volumes.md
+++ b/Docs/developer_guide/script_repository/volumes.md
@@ -452,7 +452,7 @@ print(point_Ras)
 
 ### Get the values of all voxels for a label value
 
-If you have a background image called ‘Volume’ and a mask called ‘Volume-label’ created with the Editor you could do something like this:
+If you have a background image called ‘Volume’ and a mask called ‘Volume-label’ created with the Segment Editor you could do something like this:
 
 ```python
 import numpy
--- a/Docs/user_guide/modules/data.md
+++ b/Docs/user_guide/modules/data.md
@@ -76,7 +76,7 @@ Operations (accessible in the context menu of the nodes by right-clicking them):
 - Operations for specific node types:
     - **Volumes**: icon, Edit properties and additional information in tooltip
         - **'Register this...'** action to select fixed image for registration. Right-click the moving image to initiate registration
-        - **'Segment this...'** action allows segmenting the volume, for example, in the Editor module
+        - **'Segment this...'** action allows segmenting the volume, for example, in the Segment Editor module
         - **'Toggle labelmap outline display'** for labelmaps
     - **Models**: icon, Edit properties and additional information in tooltip
     - **SceneViews**: icon, Edit properties and Restore scene view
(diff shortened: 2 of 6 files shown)


lassoan (dismissed):
Thank you, these changes look good to me.

jcfr:
Similarly to what we do for extension in  https://github.com/Slicer/ExtensionsIndex/tree/master/ARCHIVE, I  think we should include a section in the documentation to list the past modules.

What do you think of adding a section `Modules / Retired Modules` or  `Modules / Retired`

| `Modules / Retired Modules` |  `Modules / Retired` |
|--|--|
| image |   image |

Docs/user_guide/modules/retiredmodules.md
```diff
@@ -0,0 +1,12 @@
+The modules listed below have been retired and are no longer present in 3D Slicer. This page exists as a convenient marker for historical details about retired modules.
+
+Retiring modules reduces burden on slicer developers.
+
+A short list of reasons why a module may have been retired
+
+- The module has been superseded with another, and is not recommended
+- The module has no development support, and no longer runs properly
```
jcfr, line 8: The list does not seem to be rendered properly:

image

jcfr (commented):
Thanks, looks good to me.

Docs/user_guide/modules/retiredmodules.md
jamesobutler: I noticed that as well. Not sure why. I'll push again to see if something changes.

jamesobutler:
A retired modules page has been added.
image

I've also updated the "Legacy" category in the documentation to include more modules based on what has already been set as Legacy in the Slicer application
image

The documentation had "Label Statistics" module under the legacy category, but in the Slicer application it was not under the legacy category. I've added a commit to move it to the legacy category to match.

Docs/user_guide/modules/retiredmodules.md
jamesobutler: With latest push they are still not rendering appropriately. Not sure why

jcfr:
Thanks for working on consolidating the documentation :pray:. It will  help reduce confusion from users.

Since you are working on this, what do you think of updating the module page of each legacy module to include a warning section with a message like this:

```
This module is deprecated and will be removed in the future. To implement new workflow, we recommend to use <name-and-link-to-alternative-module>.
```

When possible and available, it may also be worth linking to the discussion page where we last discussed the upcoming "retirement" (aka move to legacy) of the module.

jcfr:
All of that said, moving the commits listed below to a new PR called `Add retired modules page and consolidate legacy category` may be sensible. That way we could integrate this PR that was originally intended to update Editor documentation.

```
DOC: Add retired modules page
DOC: Move legacy modules to legacy category
ENH: Move Label Statistics to Legacy category 
```

jamesobutler:
New PR for the legacy and retired modules documentation work: https://github.com/Slicer/Slicer/pull/6039

PR #6039, opened 2021-11-24
Add retired modules page and consolidate legacy category

jamesobutler:
A retired modules page has been added.
image

I've also updated the "Legacy" category in the documentation to include more modules based on what has already been set as Legacy in the Slicer application
image

| Current | This PR |
|---------|-------|
|image|image|

The documentation had "Label Statistics" module under the legacy category, but in the Slicer application it was not under the legacy category. I've added a commit to move it to the legacy category to match.


jamesobutler:
From @jcfr 

https://github.com/Slicer/Slicer/pull/6036#discussion_r756109065

> The list does not seem to be rendered properly:
> 
> image


https://github.com/Slicer/Slicer/pull/6036#issuecomment-977936124

> Thanks for working on consolidating the documentation :pray:. It will  help reduce confusion from users.
> 
> Since you are working on this, what do you think of updating the module page of each legacy module to include a warning section with a message like this:
> 
> ```
> This module is deprecated and will be removed in the future. To implement new workflow, we recommend to use <name-and-link-to-alternative-module>.
> ```
> 
> When possible and available, it may also be worth linking to the discussion page where we last discussed the upcoming "retirement" (aka move to legacy) of the module.

Docs/user_guide/modules/retiredmodules.md
```diff
@@ -0,0 +1,11 @@
+The modules listed below have been retired and are no longer present in 3D Slicer. This page exists as a convenient marker for historical details about retired modules.
+
+Retiring modules reduces burden on slicer developers.
+
+A short list of reasons why a module may have been retired:
```
jcfr, line 5:
```suggestion
A short list of reasons why a module may have been retired:

```

jcfr: @jamesobutler Adding an empty line after the `:` may help I think.

jamesobutler: I've pushed an update for this, but I think I had the empty line originally and it didn't work. Will see with the generated preview

jcfr: Look like the html include a list ... wondering if this is a side effect of the custom styling:

image

It also turns out the retired page is lacking a title:

image

jamesobutler: After providing a heading title to the new page, it appropriately renders the bullet points.

jamesobutler:
Regarding adding a warning message at the top of the Legacy module pages to indicate that it was deprecated and the suggested replacement. I'm not sure yet of the correct syntax to make a rendered message like the following:
image

jcfr:
> I'm not sure yet of the correct syntax to make a rendered message like the following

This can be using the following syntax:

```md
:::{warning}

This module is deprecated and will be removed in the future. To implement new workflow, we recommend to use [name](name).


:::
```


Here are the corresponding references:
* https://slicer.readthedocs.io/en/latest/developer_guide/build_instructions/linux.html#build-slicer
* https://github.com/Slicer/Slicer/blob/00d29391a77b3307155d68f130ba2bff88fc445d/Docs/developer_guide/build_instructions/linux.md?plain=1#L161-L165

jamesobutler:
I have added a commit that adds warning messages to the top of legacy modules detailing they are deprecated and what are the recommended modules to use instead.  Please let me know if the below recommendations should be changed.

- Annotations -> Markups
- Expert Automated Registration -> General registration (SlicerElastix)
- Label Statistics -> Segment Statistics
- Resample Scalar Volume -> Crop Volume
- Robust Statistics Segmenter -> Segment Editor
- Simple Region Growing Segmentation -> Segment Editor

Example of warning message:
image

Docs/user_guide/modules/annotations.md
```diff
@@ -1,8 +1,14 @@
 # Annotations
 
+:::{warning}
+
+This module is deprecated and will be removed in the future. It is recommended to use the [Markups](markups.md) module.
```
jcfr, line 5: Documentation  links are currently specified with and without the `.md` suffix.

Being explicit makes sense and may help reduce confusion when writing documentation.

```bash
$ cd Slicer/Docs
for file in $(ls -1 user_guide/modules/); do \
  name=$(echo $file | cut -d"." -f1); \
  ack -hi "\[\w+\s*\]\(\w*"$name; \
  ack -hi "\[\w+\s*\]\(modules\/\w*"$name; \
done
```

image

jcfr:
>    Resample Scalar Volume -> Crop Volume

I suggest to also update the description to mention it can be used for resampling as well.

https://github.com/Slicer/Slicer/blob/2515768aaf70c161d781ff41f36f2a0655c88efb/Modules/Loadable/CropVolume/qSlicerCropVolumeModule.cxx#L46-L50

Docs/user_guide/modules/annotations.md
jamesobutler: I've pushed a commit with changes to explicitly include the .md suffix when linking module pages. See if I fixed up everyplace correctly.

lassoan (commented):
Thank you, it looks good to me. The only change I would suggest is to add https://slicer--6039.org.readthedocs.build/en/6039/user_guide/modules/retiredmodules.html page to an index file, as some generators might not include the page without it, and it is also a bit harder to navigate the TOC tree if the page does not show up in there.

jcfr:
Thanks :pray: 

And since I noticed on casing "issue, I rebased and also added an extra commit to consistently reference the application or project using `Slicer` instead of `slicer`

jcfr:
> adding page to page to an index file

@jamesobutler I am done changing the topic, I will let you take care of it if that works for you

Docs/user_guide/modules/index.md
```diff
@@ -170,3 +170,7 @@ brainslabelstats.md
 performancetests.md
 selftests.md
 ```
+
+## Retired Modules
+
+See the list of [retired modules](retiredmodules.md).
```
jamesobutler, line 176: @lassoan In this index file you are suggesting it be something more like the following?

```
## Retired Modules
```{toctree}
:maxdepth: 1
retiredmodules.md
```

jamesobutler: Feel free to commit directly to this PR any suggested changes here. I'm signing off working for today.

lassoan: > In this index file you are suggesting it be something more like the following?

Yes, it should be added into a toctree.

jamesobutler:
In https://github.com/Slicer/Slicer/commit/cf62bcfc89d4fc2606a84ac51f741a93d7037299 Issued by @lassoan the “Label Statistics” module, the non-BRAINS version, was removed from Slicer Core. By that same logic should the BRAINS version be removed from Slicer? If so, then `ENH: Move Label Statistics to Legacy category` isn’t really needed, but instead removing the documentation linked entry should be done instead along with removing the module from Slicer.

Docs/user_guide/modules/resamplescalarvolume.md
```diff
@@ -1,3 +1,9 @@
+:::{warning}
+
+This module is deprecated and will be removed in the future. It is recommended to use the [Crop Volume](cropvolume.md) module.
```
lassoan, line 3: "Resample Scalar Volume" is the only module that allows simple setting of output image spacing. "Resample scalar/vector/dwi volume" module requires setting of origin, spacing, axis directions, and extents as well. Crop volume module does not allow setting anisotropic spacing. So, this module is not completely redundant. We could slightly improve "Resample scalar/vector/dwi volume" module (to allow automatic setting of image origin, directions, and extents) and then remove "Resample scalar volume" module completely.

lassoan:
> In [cf62bcf](https://github.com/Slicer/Slicer/commit/cf62bcfc89d4fc2606a84ac51f741a93d7037299) Issued by @lassoan the “Label Statistics” module, the non-BRAINS version, was removed from Slicer Core. By that same logic should the “Label Statistics (BRAINS)” version be removed from Slicer? If so, then `ENH: Move Label Statistics to Legacy category` isn’t really needed, but instead removing the documentation linked entry should be done instead along with removing the module from Slicer. The Slicer documentation has already had this BRAINS version listed as a legacy module.

Good catch, the module should be indeed moved into the retired list. If you have time to do it then please do so, if not then let me know and I can either find time to do it or just merge this PR without this fix. A lot of files are modified, so it would be better not to keep the PR open for too long to avoid chances of merge conflicts.

Docs/user_guide/modules/resamplescalarvolume.md
jamesobutler: It would seem confusing that "Legacy" would not equal "Deprecated and will be removed in the future". 

Since "Resample Scalar Volume" is already in the "Legacy" category in the Slicer application, I suggest to keep the deprecation warning. It will be removed in the future, but as indicated there is additional work needed to make it a good experience (single resample module) before it is removed. Whatever the future may hold, "Resample Scalar Volume" module specifically should not be relied on as existing in the future.
image

How about the following modification?
```suggestion
This module is deprecated and will be removed in the future. It is recommended to use the [Resample Scalar/Vector/DWI Volume](ResampleScalarVectorDWIVolumeParameters.md) module or the [Crop Volume](cropvolume.md) module.

```

Docs/user_guide/modules/index.md
jamesobutler: It's unclear to me what to do here as we aren't listing retired module pages like how the other modules are listed in categories. We aren't keeping retired module documentation, but linking to another page (retiredmodules.md) from this index page that details the removed modules in a table of information.

lassoan: My only problem is that when you open [this page](https://slicer--6039.org.readthedocs.build/en/6039/user_guide/modules/retiredmodules.html) then nothing is selected in the document tree on the left.

image

jamesobutler: Yeah I don't know what to do about that. I could as a workaround not have an individual retiredmodules.md page and instead put all the content of that page at the bottom of the index.md page.

lassoan: OK, I'll try to fix it then. Don't make forced pushes until I commit a change.
Was this the last open discussion?

jamesobutler: Ok sounds good. Yes this is the last open discussion.

PR #6040, opened 2021-11-24
ENH: Add ROI edge handles to slice views

Sunderlandkyl:
In the same way that ROI corner handles can be used to change the size of an ROI in the 3D view, this commit adds handles to the ROI edges that can be used in Slice views.

image

Re #5061


lassoan (approved):
This is awesome, thank you!

PR #6041, opened 2021-11-25
ENH: Clean up markups node API

lassoan:
Deprecate all method names for that either of these applies:
- use old terms: fiducial or markup (instead of control point)
- include "array", "vector", "pointer" in the name: this is redundant, as input arguments can be deduced from the type, and return type for vectors is almost always vtkVectorNd (as it is safe, and simple to use)

For methods that return a 3D vector, add a Python friendly variant that returns a vtkVector3d. This allows the user to get a vector in one line instead of two lines (first line creates a vector, second line sets its value).

### DIFF
--- a/Applications/SlicerApp/Testing/Python/SlicerMRBMultipleSaveRestoreLoopTest.py
+++ b/Applications/SlicerApp/Testing/Python/SlicerMRBMultipleSaveRestoreLoopTest.py
@@ -48,7 +48,7 @@ class SlicerMRBMultipleSaveRestoreLoop(ScriptedLoadableModuleTest):
 
   def __init__(self, methodName='runTest', numberOfIterations=5, uniqueDirectory=True, strict=False):
     """
-    Tests the use of mrml and mrb save formats with volumes and fiducials.
+    Tests the use of mrml and mrb save formats with volumes and point lists.
     Checks that scene views are saved and restored as expected after multiple
     MRB saves and loads.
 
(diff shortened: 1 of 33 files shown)


jamesobutler (commented):
Thanks for the additional update of the various comments and variables and overall organization of the deprecated methods in the classes. I will test building locally and testing locally but may not be immediate due to the Thanksgiving holiday here. If you would like to proceed sooner, feel free to go ahead.

Applications/SlicerApp/Testing/Python/SlicerMRBMultipleSaveRestoreTest.py
```diff
@@ -88,17 +88,15 @@ def test_SlicerMRBMultipleSaveRestore(self):
     import SampleData
     mrHeadVolume = SampleData.downloadSample("MRHead")
 
-    # Place a fiducial
-    markupsLogic = slicer.modules.markups.logic()
+    # Place a control point
+    pointListNode = slicer.mrmlScene.AddNewNodeByClass("vtkMRMLMarkupsFiducialNode")
```
jamesobutler, line 92: This change is causing `py_SlicerMRBMultipleSaveRestoreTest` to fail because it uses the `GetNodeTagName` and not the `GetDefaultNodeNamePrefix` so the added node name is "MarkupsFiducial" and not "F", but it is trying to get the node by doing `pointListNode = slicer.util.getNode('F')`

Applications/SlicerApp/Testing/Python/SlicerMRBMultipleSaveRestoreLoopTest.py
```diff
@@ -91,19 +91,17 @@ def test_SlicerMRBMultipleSaveRestoreLoop(self):
     import SampleData
     mrHeadVolume = SampleData.downloadSample("MRHead")
 
-    # Place a fiducial
-    markupsLogic = slicer.modules.markups.logic()
+    # Place a control point
+    pointListNode = slicer.mrmlScene.AddNewNodeByClass("vtkMRMLMarkupsFiducialNode")
```
jamesobutler, line 95: This change is causing `py_SlicerMRBMultipleSaveRestoreLoopTest` to fail because it uses the `GetNodeTagName` and not the `GetDefaultNodeNamePrefix` so the added node name is "MarkupsFiducial" and not "F", but it is trying to get the node by doing `pointListNode = slicer.util.getNode('F')`

jamesobutler (commented):
Made some inline comments about why there are 2 new tests that are failing due to this PR.

Applications/SlicerApp/Testing/Python/SlicerMRBMultipleSaveRestoreLoopTest.py
lassoan: Thank you for testing!
I fixed this issue but apparently I did not pushed that version. I've force pushed the updated version now.

PR #6045, opened 2021-11-25
BUG: Fix intensity masking for Islands effect

### LINKED ISSUE #6043: Segment Editor masking interferes with Islands effect
Keep largest island mode of Islands effect does not do anything if Editable intensity range is set.

## Steps to reproduce

- Load this scene: https://1drv.ms/u/s!Arm_AFxB9yqHx6YY2CKB_jms9V_W6A?e=MOB6U1 (it contains several segmented islands and editable intensity range that includes all the islands)
- In Islands effect, select "Keep largest island" mode and click Apply

=> ERROR: wait cursor is displayed for a few seconds but nothing changes

Disabling "Editable intensity range" and clicking Apply again works well.

## Expected behavior

Only the largest island should remain.

## Environment
- Slicer version: Slicer-4.11.0-2021-11-24
- Operating system: Windows10

Sunderlandkyl:
Segment editor effects that use "ModificationModeSet" restore the existing segment outside of the masking region. The mask was not being applied, and so the entire segment was restored. Fixed by only restoring the segment outside the mask area.

This commit also adds tests for intensity masking with island effects, and fixes issues that caused the test to fail when run with a main window. (Uninitialized labelmaps + unhandled VTK_SIGNED_CHAR data type in vtkImageLabelOutline).

Fixes #6043

### DIFF
--- a/Libs/MRML/Logic/vtkImageLabelOutline.cxx
+++ b/Libs/MRML/Logic/vtkImageLabelOutline.cxx
@@ -254,6 +254,10 @@ void vtkImageLabelOutline::ThreadedExecute(vtkImageData *inData,
     vtkImageLabelOutlineExecute(this, inData, (char *)(inPtr),
       outData, outExt, id);
     break;
+  case VTK_SIGNED_CHAR:
+    vtkImageLabelOutlineExecute(this, inData, (signed char *)(inPtr),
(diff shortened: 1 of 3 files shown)


lassoan (approved):
Thanks for the very quick fix, it looks good and works well.

PR #6046, opened 2021-11-25
ENH: Improve Markups rotation handle appearance

Sunderlandkyl:
Currently the Markups rotation handle "hangs" as a single arc around the rotation axis, which looks strange if the translation handles are hidden. This commit connects the arc with a right angle so that it forms an enclosed shape.

Before:
image

After:
image

Re #5061


lassoan (approved):
Thank you, this looks nice!

PR #6052, opened 2021-11-30
DOC: Fix unintended link in ResampleScalarVectorDWIVolme documentation

jamesobutler:
Saw this minor issue when looking through some documentation.

The generated documentation was creating a hyperlink to the webpage "Transforms.it".


lassoan (approved):
Good catch!

PR #6053, opened 2021-12-01
COMP: Improve "CI" GitHub Action workflow uploading package as artifact

jcfr:
This commit copies the package to `GITHUB_WORKSPACE` instead of `RUNNER_TEMP`. This
is required to workaround issue discussed in actions/runner#965 (Incorrect context
paths in container)

It also adds a step for uploading the generated Slicer package as an
artifact with a retention time of 1 day.


lassoan:
Is there a way to download the created install package?

We could even build a SlicerNotebook image with it. Testers could run that in the web browser using https://mybinder.org/, without the need to download and install the package on a Linux box.

PR #6054, opened 2021-12-01
ENH: Update FindVcvars for future MSVC versions

jamesobutler:
This is based on FindVcvars.cmake as of https://github.com/scikit-build/cmake-FindVcvars/commit/a57b8b9df60ba1bd650a81f7f28c8b9ea4f96beb.

### DIFF
--- a/CMake/FindVcvars.cmake
+++ b/CMake/FindVcvars.cmake
@@ -111,15 +111,20 @@
 
 cmake_minimum_required(VERSION 3.13.4)
 
-# TODO Support lookup of "Microsoft Visual C++ Compiler for Python 2.7 (x86, amd64)"
-
 # Global variables used only in this script (unset at the end)
 set(_Vcvars_MSVC_ARCH_REGEX "^(32|64)$")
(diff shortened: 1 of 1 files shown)


lassoan (approved):

I'm just curious, what this script is used for? I can already build Slicer with VS2022 and v143 toolset.

jamesobutler:
Some extensions such as SlicerRadiomics having some configuring that relies on this CMake file. When building those extensions with a newer MSVC version that is not yet specified in this CMake file, then it will fail to build. For example when Visual Studio 2022 17.1 version is release it will probably be MSVC 1931 instead of 1930. This PR is already prepping for the existence of 1930-1939 as being reserved for Visual Studio 2022 versions. 

This file was previously only updated when new versions came out which resulted in issues. For example, you can't build current Slicer stable 4.11.20210226 and build SlicerRadiomics against it without updating this FindVCVars.cmake. That's because https://github.com/Slicer/Slicer/blob/v4.11.20210226/CMake/FindVcvars.cmake only supports up to MSVC 1928 (Visual Studio 2019 16.8/16.9) however Visual Studio 2019 16.10/16.11 have MSVC 1929.

PR #6055, opened 2021-12-01
COMP: Update Landmark Registration

jamesobutler:
```
$ git shortlog --topo-order --no-merges 290d481..551cee5
Andras Lasso (1):
      Update test data download URL

James Butler (2):
      ENH: Use non-deprecated markups API
      ENH: Use point terminology instead of fiducial
```


lassoan (approved):
Thank you!

PR #6056, opened 2021-12-01
ENH: Clean up qSlicerMarkupsModuleWidget to use appropriate Control Point terminology

jamesobutler:
This matches changes other recent changes to consistent usage of "control point" instead of "fiducial". "Control point" is already appropriately used in the Markups toolbar, but was not in the Markups module widget.

| Current | This PR |
|----------|---------|
| image| image|


lassoan (approved):

PR #6058, opened 2021-12-01
DOC: Clarify "size" description in markups JSON schema

mikebind:
Attempted to modify the description of the "size" item so that it more clearly distinguished between characterizing the size of a markups object as a radius from a center point vs a diameter across a center point.  Schema change was suggested here: https://discourse.slicer.org/t/explanation-needed-for-json-files-created-by-markup/20842/7

### DIFF
--- a/Modules/Loadable/Markups/Resources/Schema/markups-schema-v1.0.3.json
+++ b/Modules/Loadable/Markups/Resources/Schema/markups-schema-v1.0.3.json
@@ -149,7 +149,7 @@
                                     "$id": "#markup/size",
                                     "type": "array",
                                     "title": "Size",
-                                    "description": "The size of the markups representation. Ex. size of the ROI or plane markups.",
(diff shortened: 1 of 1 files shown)


Modules/Loadable/Markups/Resources/Schema/markups-schema-v1.0.3.json
```diff
@@ -149,7 +149,7 @@
                                     "$id": "#markup/size",
                                     "type": "array",
                                     "title": "Size",
-                                    "description": "The size of the markups representation. Ex. size of the ROI or plane markups.",
+                                    "description": "The size of the markups representation. Ex. size (=axis-aligned edge lengths) of the ROI or plane markups.",
```
lassoan, line 152:
```suggestion
                                    "description": "The size of the markups representation. For example, axis-aligned edge lengths of the ROI or plane markups.",
```

PR #6060, opened 2021-12-02
ENH: added check for accepted event

fbordignon:
If the close event was treated at an event filter and accepted, do not show the confirmation dialog and perform the remaining on close routines.


lassoan:
The test was fixed and integrated in https://github.com/Slicer/Slicer/commit/7610ddea1b40b569641c77964976718be045dde7

PR #6061, opened 2021-12-02
ENH: Add vtkMRMLMarkupsDisplayNode interaction handle scale property

Sunderlandkyl:
Previously the size of the interaction handles was controlled using the vtkMRMLMarkupsDisplayNode::GlyphScale/GlyphSize properties.

This commit adds the vtkMRMLMarkupsDisplayNode::InteractionHandleScale property, which behaves in the same way as GlyphScale except only affects the interaction handle. The interaction handles are no longer affected by Glyph properties.

Also cleans up the constants used to define the size of the interaction widget.

Re #5061


lassoan (approved):
Thanks a lot. It looks good. I've just made a few minor improvements (save/restore scale into default markups properties, updated documentation, etc.).

PR #6064, opened 2021-12-03
ENH: Center "Save before exit?" window on main window

achataigner:
To center "Save before exit?" window on main window even if the main window is not maximized.

### DIFF
--- a/Base/QTApp/qSlicerMainWindow.cxx
+++ b/Base/QTApp/qSlicerMainWindow.cxx
@@ -630,7 +630,7 @@ bool qSlicerMainWindowPrivate::confirmCloseApplication()
   bool close = false;
   if (!question.isEmpty())
     {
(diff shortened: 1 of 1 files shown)


pieper:
Makes sense to me, thanks for the contribution 👍

Base/QTApp/qSlicerMainWindow.cxx
```diff
@@ -630,7 +630,7 @@ bool qSlicerMainWindowPrivate::confirmCloseApplication()
   bool close = false;
   if (!question.isEmpty())
     {
-    QMessageBox* messageBox = new QMessageBox(QMessageBox::Warning, qSlicerMainWindow::tr("Save before exit?"), question, QMessageBox::NoButton);
+    QMessageBox* messageBox = new QMessageBox(QMessageBox::Warning, qSlicerMainWindow::tr("Save before exit?"), question, QMessageBox::NoButton, q_ptr);
```
lassoan, line 633: Is there any specific reason you used `q_ptr` instead of `q`? A few lines below we already use `q` in a very similar role.

PR #6067, opened 2021-12-03
ENH: Clean up markups API further

jamesobutler:
I did this work of additional markups API cleanup after noticing the Markups module documentation beginning at the top with an example of how to add a control point and it doing `slicer.modules.markups.logic().AddFiducial()` instead of something like `slicer.modules.markups.logic().AddControlPoint()`.

Ran all tests on Windows and only the currently failing tests are still failing.

```
Deprecate all method names for that either of these applies:
- use old terms: fiducial or markup (instead of control point)
```

### DIFF
--- a/Modules/Loadable/Markups/Logic/vtkSlicerMarkupsLogic.h
+++ b/Modules/Loadable/Markups/Logic/vtkSlicerMarkupsLogic.h
@@ -100,11 +100,11 @@ public:
   /// On success, return the id, on failure return an empty string.
   std::string AddNewFiducialNode(const char *name = "F", vtkMRMLScene *scene = nullptr);
 
-  /// Add a new fiducial to the currently active list at the given RAS
-  /// coordinates (default 0,0,0). Will create a list is one is not active.
-  /// Returns -1 on failure, index of the added fiducial
+  /// Add a new control point to the currently active markups fiducial node at the given RAS
+  /// coordinates (default 0,0,0). Will create a markups fiducial node if one is not active.
+  /// Returns -1 on failure, index of the added control point
   /// on success.
-  int AddFiducial(double r=0.0, double a=0.0, double s=0.0);
+  int AddControlPoint(double r=0.0, double a=0.0, double s=0.0);
 
   /// jump the slice windows to the given coordinate
   /// If viewGroup is -1 then all all slice views are updated, otherwise only those views
@@ -121,7 +121,7 @@ public:
   /// \sa FocusCamerasOnNthPointInMarkup
   void FocusCameraOnNthPointInMarkup(const char *cameraNodeID, const char *markupNodeID, int n);
 
-  /// Load a markups fiducial list from fileName, return nullptr on error, node ID string
+  /// Load a markups node from fileName, return nullptr on error, node ID string
   /// otherwise. Adds the appropriate storage and display nodes to the scene
   /// as well.
   char* LoadMarkups(const char* fileName, const char* fidsName=nullptr, vtkMRMLMessageCollection* userMessages=nullptr);
@@ -132,13 +132,15 @@ public:
   char* LoadMarkupsFromFcsv(const char* fileName, const char* nodeName=nullptr, vtkMRMLMessageCollection* userMessages=nullptr);
   char* LoadMarkupsFromJson(const char* fileName, const char* nodeName=nullptr, vtkMRMLMessageCollection* userMessages=nullptr);
 
-  /// Utility methods to operate on all markups in a markups node
-  void SetAllMarkupsVisibility(vtkMRMLMarkupsNode *node, bool flag);
-  void ToggleAllMarkupsVisibility(vtkMRMLMarkupsNode *node);
-  void SetAllMarkupsLocked(vtkMRMLMarkupsNode *node, bool flag);
-  void ToggleAllMarkupsLocked(vtkMRMLMarkupsNode *node);
-  void SetAllMarkupsSelected(vtkMRMLMarkupsNode *node, bool flag);
-  void ToggleAllMarkupsSelected(vtkMRMLMarkupsNode *node);
+  /// Utility methods to operate on all control points in a markups node
+  /// @{
+  void SetAllControlPointsVisibility(vtkMRMLMarkupsNode *node, bool flag);
+  void ToggleAllControlPointsVisibility(vtkMRMLMarkupsNode *node);
+  void SetAllControlPointsLocked(vtkMRMLMarkupsNode *node, bool flag);
+  void ToggleAllControlPointsLocked(vtkMRMLMarkupsNode *node);
+  void SetAllControlPointsSelected(vtkMRMLMarkupsNode *node, bool flag);
+  void ToggleAllControlPointsSelected(vtkMRMLMarkupsNode *node);
+  /// @}
 
   /// Utility method to set up a display node from the defaults.
   /// Point labels visibility and properties label visibility setting is not saved to defaults,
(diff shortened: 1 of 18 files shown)


Modules/Loadable/Markups/Logic/vtkSlicerMarkupsLogic.h
```diff
@@ -132,13 +132,13 @@ class VTK_SLICER_MARKUPS_MODULE_LOGIC_EXPORT vtkSlicerMarkupsLogic :
...
   char* LoadMarkupsFromJson(const char* fileName, const char* nodeName=nullptr, vtkMRMLMessageCollection* userMessages=nullptr);
 
-  /// Utility methods to operate on all markups in a markups node
-  void SetAllMarkupsVisibility(vtkMRMLMarkupsNode *node, bool flag);
-  void ToggleAllMarkupsVisibility(vtkMRMLMarkupsNode *node);
-  void SetAllMarkupsLocked(vtkMRMLMarkupsNode *node, bool flag);
-  void ToggleAllMarkupsLocked(vtkMRMLMarkupsNode *node);
-  void SetAllMarkupsSelected(vtkMRMLMarkupsNode *node, bool flag);
-  void ToggleAllMarkupsSelected(vtkMRMLMarkupsNode *node);
+  // /// Utility methods to operate on all control points in a markups node
```
jcfr, line 135:
```suggestion
  /// Utility methods to operate on all control points in a markups node
  /// @{
```

Modules/Loadable/Markups/Logic/vtkSlicerMarkupsLogic.h
```diff
@@ -132,13 +132,13 @@ class VTK_SLICER_MARKUPS_MODULE_LOGIC_EXPORT vtkSlicerMarkupsLogic :
...
+  void SetAllControlPointsVisibility(vtkMRMLMarkupsNode *node, bool flag);
+  void ToggleAllControlPointsVisibility(vtkMRMLMarkupsNode *node);
+  void SetAllControlPointsLocked(vtkMRMLMarkupsNode *node, bool flag);
+  void ToggleAllControlPointsLocked(vtkMRMLMarkupsNode *node);
+  void SetAllControlPointsSelected(vtkMRMLMarkupsNode *node, bool flag);
+  void ToggleAllControlPointsSelected(vtkMRMLMarkupsNode *node);
```
jcfr, line 141:
```suggestion
  void ToggleAllControlPointsSelected(vtkMRMLMarkupsNode *node);
  /// @}

File: Modules/Loadable/Markups/qSlicerMarkupsModuleWidget.h [hunk 3]
```diff
@@ -365,18 +365,6 @@ public slots:
     qWarning("qSlicerMarkupsModuleWidget::onMissingMarkupPushButtonClicked method is deprecated, please use onMissingControlPointPushButtonClicked instead");
     this->onMissingControlPointPushButtonClicked();
     };
-  /// \deprecated Use onDeleteControlPointPushButtonClicked instead.
-  void onDeleteMarkupPushButtonClicked(bool confirm=true)
```
jcfr, line 369: Is the removal of these slots (`onDeleteMarkupPushButtonClicked` and `onDeleteAllMarkupsInListPushButtonClicked`) intentional ? 

It looks like these were initially added in Slicer@516b64a9b and then deprecated in Slicer@edd0af8a5 along with others like `onMissingMarkupPushButtonClicked` ....

jamesobutler: It’s being replaced by `onDeleteControlPointPushButtonClicked` and `onDeleteAllControlPointsInListPushButtonClicked`. It is due to incorrect names of “Markup” instead of “Control Point” used for button names defined in a UI file.

This means I have missed removing the incorrect name usage of `missingMarkupPushButton` in the UI file as well.

jamesobutler: I have pushed an update to maintain these deprecated slots, but also still update the object names in the UI file.

jamesobutler:
This PR includes documentation updates and deprecated methods as mentioned in a point in #5989.

jamesobutler:
This PR should be ready for merging whenever

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsNode.cxx
```diff
@@ -138,7 +138,7 @@ void vtkMRMLMarkupsNode::ReadXMLAttributes(const char** atts)
 
   vtkMRMLReadXMLBeginMacro(atts);
   vtkMRMLReadXMLBooleanMacro(locked, Locked);
-  vtkMRMLReadXMLStdStringMacro(markupLabelFormat, MarkupLabelFormat);
```
lassoan, line 141: We need to keep this line. It will call `SetMarkupLabelFormat()` which should set ControlPointLabelFormat without logging a warning (because all existing scene files will contain `markupLabelFormat` attribute name). Since the label format is stored in the mkp.json file, too, but we can do it later (if we touch the label formatting for other reasons anyway).

lassoan (changes requested):
Thanks a lot, it looks good except one minor thing around MarkupLabelFormat (see inline comment).

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsNode.cxx
jamesobutler: I've restored this line here, but restoring a mrk.json doesn't seem to log the warning? I've been trying to replicate that before commenting that warning out.

jamesobutler:
The `PlaceMultipleMarkupsType` enum is also another instance of "markup" being misused to mean control point. I haven't yet updated that yet because I guess when a markup has a fixed number of points being in place multiple mode will wrap around to place an actual new markup node to continue placing control points. It's primarily about placing multiple control points but that one weird instance of it being able to create multiple markups does seem to exist.

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsNode.cxx
lassoan: I would expect that the warning is logged if you load a scene that is saved by any earlier Slicer version, because in those scene files the "MarkupLabelFormat" attribute is used.

lassoan:
I'm sorry, I've caused a merge conflict in qSlicerMarkupsModule.ui with a fix that I've committed for the splitter (to allow changing the tree view height). I know this can be frustrating, if you don't have time to rebase and resolve the conflict then I'll do it when merging.

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsNode.cxx
jamesobutler: @lassoan I've tried this again but I don't see a warning log when loading of an older scene. Is this because the mrk.json file has a "labelFormat" field and not a "MarkupLabelFormat" field? You can test and see if you can load a scene and have the warning logged from reading it.

lassoan: It is not related to the json file content, but to the scene content (read XML). The warning should appear when a saved scene is loaded.

jamesobutler: When I load a scene file that includes a mark.json type object it does not log the warning.

jamesobutler: @lassoan I was able to replicate and have removed the logged warning so that the scene file MarkupLabelFormat attribute doesn't trigger a warning on load.

jamesobutler:
@lassoan Do you mind resolving the merge conflicts introduced due to the latest merges?

jamesobutler:
Note this was otherwise ready for merging as well.

jamesobutler:
@lassoan I've gone ahead and resolved the merge conflicts.

lassoan (approved):
This is looks good, thanks a lot for taking care of this.

PR #6068, opened 2021-12-03
ENH: Add Markups ROI InsideOut flag

Sunderlandkyl:
The inside out function can be used to invert the ROI so that points inside the ROI are considered outside, and vice versa. This state is reflected in GetPlanes()/GetPlanesWorld() and IsPointInROI()/IsPointInROIWorld().

This commit also adds implicit functions to the ROI node that can be used by other nodes, and that will invert their sign based on the InsideOut state.

Re #5061


lassoan (approved):
Thank you, this works very nicely. I've just improved the documentation a bit and force-pushed to this branch.

PR #6070, opened 2021-12-05
ENH: Add ability to toggle segment editor tool button style

rbumm:
by pressing the "t" key.

The new tool button layout in the segment editor is elegant, but the icons are hard to remember and will be even more if new effects are added. The tool texts are too slow. This small change adds a shortcut to the "t" key to bring back all button bottom labels for a short moment if needed.

### DIFF
--- a/Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.cxx
+++ b/Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.cxx
@@ -2133,10 +2133,10 @@ void qMRMLSegmentEditorWidget::onMasterVolumeNodeChanged(vtkMRMLNode* node)
 //-----------------------------------------------------------------------------
 void qMRMLSegmentEditorWidget::onEffectButtonClicked(QAbstractButton* button)
 {
+  Q_D(qMRMLSegmentEditorWidget);
   // Get effect that was just clicked
   qSlicerSegmentEditorAbstractEffect* clickedEffect = qobject_cast<qSlicerSegmentEditorAbstractEffect*>(
     button->property("Effect").value<QObject*>() );
-
   this->setActiveEffect(clickedEffect);
 }
 
@@ -3144,6 +3144,23 @@ void qMRMLSegmentEditorWidget::toggleMasterVolumeIntensityMaskEnabled()
     !d->ParameterSetNode->GetMasterVolumeIntensityMask());
 }
 
+//------------------------------------------------------------------------------
+void qMRMLSegmentEditorWidget::toggleEffectButtonLabelVisibility()
+{
+  Q_D(qMRMLSegmentEditorWidget);
+
+  if (d->EffectButtonStyle == Qt::ToolButtonIconOnly)
+    {
+    this->setSizePolicy(QSizePolicy::Ignored, QSizePolicy::Preferred);
+    this->setEffectButtonStyle(Qt::ToolButtonTextUnderIcon);
+    }
+  else
+    {
+    this->setSizePolicy(QSizePolicy::Preferred, QSizePolicy::Preferred);
+    this->setEffectButtonStyle(Qt::ToolButtonIconOnly);
+    }
+}
+
 //-----------------------------------------------------------------------------
 void qMRMLSegmentEditorWidget::undo()
 {
@@ -3264,9 +3281,15 @@ void qMRMLSegmentEditorWidget::installKeyboardShortcuts(QWidget* parent /*=nullp
     QObject::connect(nextShortcut, SIGNAL(activated()), this, SLOT(onSelectSegmentShortcut()));
     }
 
+  // Key i
   QShortcut* toggleMasterVolumeIntensityMaskShortcut = new QShortcut(QKeySequence(Qt::Key_I), parent);
   d->KeyboardShortcuts.push_back(toggleMasterVolumeIntensityMaskShortcut);
   QObject::connect(toggleMasterVolumeIntensityMaskShortcut, SIGNAL(activated()), this, SLOT(toggleMasterVolumeIntensityMaskEnabled()));
+
+  // Key t
+  QShortcut* toggleToolButtonStyleShortcut = new QShortcut(QKeySequence(Qt::Key_L), parent);
+  d->KeyboardShortcuts.push_back(toggleToolButtonStyleShortcut);
+  QObject::connect(toggleToolButtonStyleShortcut, SIGNAL(activated()), this, SLOT(toggleEffectButtonLabelVisibility()));
 }
 
 //-----------------------------------------------------------------------------
--- a/Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.h
+++ b/Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.h
@@ -276,6 +276,9 @@ public slots:
   /// Enable/disable masking using master volume intensity
   void toggleMasterVolumeIntensityMaskEnabled();
 
+  /// Toggle style of effect buttons to see button labels if needed.
+  void  toggleEffectButtonLabelVisibility();
+
   /// Restores previous saved state of the segmentation
   void undo();
 


Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.h
```diff
@@ -276,6 +276,9 @@ public slots:
   /// Enable/disable masking using master volume intensity
   void toggleMasterVolumeIntensityMaskEnabled();
 
+  /// Enable/disable masking using master volume intensity
```
jamesobutler, line 279: Wrong doc text

jamesobutler (commented):
How do you plan to actually use this feature? Are you going to turn the tool button text on and leave it on? If so, this would make more sense as a setting in a Segment Editor application settings panel. Or do you plan to just temporarily show the text for a few seconds and then turn it off?

Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.h
rbumm: I would probably just use it for some seconds to see the full range of effects, then turn it off again to save space.

rbumm:
I would probably just use it for some seconds to see the full range of effects, then turn it off again to save space.

Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.h
```diff
@@ -276,6 +276,9 @@ public slots:
   /// Enable/disable masking using master volume intensity
   void toggleMasterVolumeIntensityMaskEnabled();
 
+  /// Toggle style of effect buttons to see button labels if needed.
+  void toggleToolButtonStyle();
```
lassoan, line 280: For consistency, the method name should include "effect". Also, the style is not a binary state, so it cannot be toggled. Something like toggleEffectButtonLabelVisibility would work.

Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.h
```diff
@@ -276,6 +276,9 @@ public slots:
   /// Enable/disable masking using master volume intensity
   void toggleMasterVolumeIntensityMaskEnabled();
 
+  /// Toggle style of effect buttons to see button labels if needed.
```
lassoan, line 279: Hardcoding a specific label style would not be very bad, but it would be nicer if we remembered what was the label style and toggle between that style and no label.

Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.cxx
```diff
@@ -3264,9 +3279,15 @@ void qMRMLSegmentEditorWidget::installKeyboardShortcuts(QWidget* parent /*=nullp
     QObject::connect(nextShortcut, SIGNAL(activated()), this, SLOT(onSelectSegmentShortcut()));
     }
 
+  // Key i
   QShortcut* toggleMasterVolumeIntensityMaskShortcut = new QShortcut(QKeySequence(Qt::Key_I), parent);
   d->KeyboardShortcuts.push_back(toggleMasterVolumeIntensityMaskShortcut);
   QObject::connect(toggleMasterVolumeIntensityMaskShortcut, SIGNAL(activated()), this, SLOT(toggleMasterVolumeIntensityMaskEnabled()));
+
+  // Key t
+  QShortcut* toggleToolButtonStyleShortcut = new QShortcut(QKeySequence(Qt::Key_T), parent);
```
lassoan, line 3288: This shortcut is already used. It would be better to find one that is not already listed here: https://slicer.readthedocs.io/en/latest/user_guide/user_interface.html#mouse-keyboard-shortcuts

lassoan (commented):
Thank you, this can be indeed useful for those who don't regularly use the Segment Editor.

How could this be discoverable for newcomers?

We must at least add to the Segment Editor documentation but it would be better to show it somewhere in the application, too.

pieper:
I agree the feature sounds useful, but I'm not sure a keyboard shortcut is the best option for this (at least not the best primary option).  How about a small button in the toolbar that exposes the button labels.  It could be a checkbox or it could just be a temporary toggle that keeps the labels visible for as long as the button is depressed.  I think this is as good as a keyboard shortcut since they mouse pointer will already need to be by the tool bar.

Another option: how about displaying the labels whenever the "None" tool is selected?  In that mode space in the GUI is not an issue, and it's likely to be the time when the user may need hints about which effect to select.  The labels would then be hidden while there's an active effect but come back when None is selected.  I think this would be handy for all users and easily discoverable.

rbumm:
> Another option: how about displaying the labels whenever the "None" tool is selected? In that mode space in the GUI is not an issue, and it's likely to be the time when the user may need hints about which effect to select. The labels would then be hidden while there's an active effect but come back when None is selected. I think this would be handy for all users and easily discoverable.

That is a good idea, I like that. Should I realize that?

jamesobutler:
> Another option: how about displaying the labels whenever the "None" tool is selected? In that mode space in the GUI is not an issue, and it's likely to be the time when the user may need hints about which effect to select. The labels would then be hidden while there's an active effect but come back when None is selected. 

This should be done only if the size of the buttons do not change. A problem with the old design was that clicking on various effects, including from the None effect, the effects buttons would move location in the GUI. If the effects are shown when in the “None” effect is active and the buttons are larger and then the buttons go smaller when an effect is selected, that is bad design. That could result in a double click action resulting in the second click acting on a second object rather than the original clicked object. Activating different effects should not change the position of the effects on screen.

pieper:
> This should be done only if the size of the buttons do not change.

I'm not sure I agree with that logic.  It's common for the GUI to change in response to button clicks, especially when details like labels are made invisible to provide space for a new GUI element.

If it's easy enough to try then we could see if we like it or not.

rbumm:
> If it's easy enough to try then we could see if we like it or not.

I see @jamesobutler s point but I will just try to implement the 'None' system for a test. Will leave a shortcut in unless we decide not to like it.

rbumm:
Done.

jamesobutler:
> It's common for the GUI to change in response to button clicks, especially when details like labels are made invisible to provide space for a new GUI element.

I would agree that GUI can change with button clicks when displaying new objects, however in this case the Segment Editor Effects group of effect buttons is the same general group object. If I was to go from the "None" effect, then click a different editor effect and then that same editor effect was no longer under my mouse button, but instead shifted upwards, that is unexpected.

jamesobutler:
If moving of the segment editor effect buttons upon clicking other effect buttons is something with varying opinion, I would suggest a discourse post revisiting this thread https://discourse.slicer.org/t/segment-editor-effect-button-size-consistency/19214.

rbumm:
> The contents of this PR originally discussed in https://discourse.slicer.org/t/new-segment-editor-layout-vertical-effect-toolbar/19649/16.

I was aware of this thread but still felt not comfortable using the segment editor in the preview.

jamesobutler (commented):
Current state of the PR doesn't quite show the effect icons when in the "None" effect. For example, if I'm in say the "Paint" effect (effect text is hidden), and then I go into "AdjustWindowLevel" mouse mode, this automatically changes the effect to "None" so that it isn't also accidentally painting while adjusting window level. It is in the "None" effect, but the effect text is not showing.

image

rbumm:
> Current state of the PR doesn't quite always show the effect name text when in the "None" effect. 

This is true, solved.

lassoan:
When the "None" effect is selected we could show a toggle button or checkbox in the effect options section to show/hide the button labels.

jamesobutler:
> Maybe we could make the label text toggle when the user right-clicks on this particular frame and selects a menu option “button texts”, then save this preference for the next start? 

^From https://discourse.slicer.org/t/new-segment-editor-layout-vertical-effect-toolbar/19649/20.

> I would probably just use it for some seconds to see the full range of effects, then turn it off again to save space.

^ From https://github.com/Slicer/Slicer/pull/6070#issuecomment-986214212

@rbumm Have you changed your mind about wanting this ability of showing the button text to be a persistent setting option so that you don’t have to toggle it on every time you start Slicer? Or do you still prefer that the text on the effect buttons to be remembered?

lassoan:
> Have you changed your mind about wanting this ability of showing the button text to be a persistent setting option so that you don’t have to toggle it on every time you start Slicer?

If the user shows the labels in Segment Editor then it should not be enforced for all embedded qSlicerSegmentEditorWidget instances. If we made it possible to save it in the application settings then a settings key must be added to the widget.

We could also consider using the existing application setting that controls visibility of toolbar button labels. That would be a global setting (applies to all buttons in all toolbars), so it could be OK to apply it to all segment editor widget instances.

rbumm:
> Have you changed your mind about wanting this ability of showing the button text to be a persistent setting option so that you don’t have to toggle it on every time you start Slicer? Or do you still prefer that the text on the effect buttons to be remembered?

No, but at the present PR state, the toggle action changes Slicer's frame layout with no sliding back of the (left) action frame as soon Qt::ToolButtonIconOnly is in effect. 

Maybe unrelated: several modules change the position of the right border of the left action frame, tending to make the left action frame bigger, switching to a new module is not resetting it to the user-defined size. In these cases it is hard to grab the frame border and push it back to the left, you need to move it right before you can move it to the left, which may be confusing to new users.

rbumm:
Testing the shortcut again find it useful as a memo when having not used the new version of the segment editor for a while and forgetting the meaning of the icons.  Probably I am "icon-blind". 

A small feature to keep the module panel width stable with button effect labels "on" was added today.  
Has there been any consensus about this PR yet?

rbumm:
Think nobody really liked that, so I close this PR. No problem

PR #6073, opened 2021-12-07
Fix SH combobox selection issues when closing the scene

cpinter:


Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyTreeView.cxx
```diff
@@ -2399,8 +2413,12 @@ void qMRMLSubjectHierarchyTreeView::onMRMLSceneEndBatchProcess(vtkObject* sceneO
     }
 
   Q_D(qMRMLSubjectHierarchyTreeView);
-  this->setCurrentItems(d->SelectedItemsToRestore);
-  d->SelectedItemsToRestore.clear();
+
+  if (d->SelectedItemsToRestore.size())
```
lassoan, line 2417: `if (!d->SelectedItemsToRestore.empty())` would be nicer

Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyComboBox.cxx
```diff
@@ -745,3 +748,15 @@ void qMRMLSubjectHierarchyComboBox::updateComboBoxTitleAndIcon(vtkIdType selecte
...
+//-----------------------------------------------------------------------------
+void qMRMLSubjectHierarchyComboBox::onMRMLSceneCloseEnded(vtkObject* sceneObject)
+{
+  vtkMRMLScene* scene = vtkMRMLScene::SafeDownCast(sceneObject);
+  if (!scene)
+    {
+    return;
+    }
+
+  this->updateComboBoxTitleAndIcon(0);
```
lassoan, line 761: Using `vtkMRMLSubjectHierarchyNode::INVALID_ITEM_ID` instead of `0` and adding a comment about what it does (e.g., "Update SH combobox title to no selection after scene close") would make things a bit more clear.

lassoan (commented):
Looks good to me, and if it fixes those annoying warnings when closing the scene then it is awesome!

I've just added two trivial comments.

cpinter:
Thank you for the review, requests done.

I'm not sure if this fixes those errors, possibly it fixes some, but this is a change in the right direction anyway. I agree there are too many SH related error messages that are superfluous. And there are some (like about cached item ID discrepancies) that are a bit worrying but apparently do not cause issues. It would need a bit of maintenance work for sure...

lassoan (approved):
Looks good, thank you.

PR #6080, opened 2021-12-13
BUG: Fix vtkMRMLMarkupsNode::GetCenterPositionWorld deprecation warning

Sunderlandkyl:
Whenever a Markups closed curve node is modified the displayable manager calls vtkMRMLMarkupsNode::GetCenterPositionWorld. This logs a warning due to the use of the deprecated GetCenterPositionVector method.

Fixed by using the non-deprecated GetCenterPosition.


lassoan (approved):
Thank you!

PR #6081, opened 2021-12-14
DOC: Update CMake Prerequisite to 3.21.1

nathanbmnt:
I was unable to build Slicer because of many `PDB API call failed` errors. It turns out C-Make > 3.20.4 and < 3.21.1 do not add the `/FS` flag to the command line. I was unlucky enough to be using one of these versions of C-Make, and ITK uses that flag, so it failed to build.

https://discourse.itk.org/t/itk-installation-error/3032/9
https://gitlab.kitware.com/cmake/cmake/-/merge_requests/6390

### DIFF
--- a/Docs/developer_guide/build_instructions/windows.md
+++ b/Docs/developer_guide/build_instructions/windows.md
@@ -2,7 +2,7 @@
 
 ## Install prerequisites
 
-- [CMake](http://www.cmake.org/cmake/resources/software.html) >= 3.15.1
+- [CMake](http://www.cmake.org/cmake/resources/software.html) >= 3.21.1
 - [Git](https://git-scm.com/download/win) >= 1.7.10
   - Note: CMake must be able to find `git.exe` and `patch.exe`. If git is installed in the default location then they may be found there, but if they are not found then either add the folder that contains them to `PATH` environment variable; or set `GIT_EXECUTABLE` and `Patch_EXECUTABLE` as environment variables or as CMake variables at configure time.
 - [Visual Studio](https://visualstudio.microsoft.com/downloads/)


Docs/developer_guide/build_instructions/windows.md
```diff
@@ -2,7 +2,7 @@
 
 ## Install prerequisites
 
-- [CMake](http://www.cmake.org/cmake/resources/software.html) >= 3.15.1
+- [CMake](http://www.cmake.org/cmake/resources/software.html) >= 3.21.1
```
lassoan, line 5:
```suggestion
- [CMake](http://www.cmake.org/cmake/resources/software.html) version >= 3.21.1 (or 3.15.1 <= version < 3.20.4)
```

lassoan:
Great work figuring all this out! Thank you for the doc update, this may save time for many developers.

jamesobutler:
@lassoan The merged commit seems to indicate an appropriate CMake version would be 3.15.4 and up until just before 3.20.4. From the [CMake issue](https://gitlab.kitware.com/cmake/cmake/-/issues/22476) and the [Visual Studio Issue](https://developercommunity.visualstudio.com/t/VS2019-2019-16104-no-longer-passes-FS/1482274) it appears that it isn’t an issue for any of the 3.20.x releases including 3.20.4, 3.20.5 and therefore 3.20.6 as well. 

It appears the issue is specific to 3.21.0 due to a compiler issue in Visual Studio 16.10. From the Microsoft developer the issue was resolved starting in Visual Studio 2022 17.0 preview 5. Therefore the more accurate and simpler note is to not use CMake 3.21.0 unless using Visual Studio 2022.

lassoan:
According to the PR description: "It turns out C-Make > 3.20.4 and < 3.21.1 do not add the /FS flag to the command line" - therefore it seems that 3.20.5, 3.20.6 are bad.

@nathanbmnt @jamesobutler please discuss this between each other and once you have a conclusion please submit a new pull request if needed.

jamesobutler:
The original PR description indicates versions after 3.20.4 are bad (>3.20.4) indicating 3.20.4 itself is good. However the Slicer commit that was merged based on @lassoan’s updates changed it to read that versions before 3.20.4 were good (<3.20.4) indicating that 3.20.4 itself is bad. This seems like an unintentional mistake.

From @N-Dekker ’s ITK post this seems to indicate the 3.20.4 is a working version.

> It looks like those fatal PDB errors don’t appear when using an older CMake version (CMake 3.20.4 or older)! 

From @N-Dekker’s Visual Studio forum post it seems to indicate 3.20.5 is also good:

> CMake 3.21.0 is the first major CMake release that adds <ForceSynchronousPDBWrites> elements to its generates vcxproj files. As a workaround, one may use a different version of CMake, for example CMake 3.20.5. 

@N-Dekker you may be able to help here since you were directly involved with this discovery. Am I reading things correctly that the problem is only when using CMake 3.21.0 and with Visual Studio 16.10.4 with the compiler issue? No issues with CMake 3.20.x? Where CMake 3.21.1 provided a change in behavior to fix the issue and then later Visual Studio 2022 17.0 also provided a fix for the issue?

lassoan:
We should not require using latest CMake release, because a wide range of CMake versions work fine (I've been using various CMake versions on different computers and have not run into this issue). We need to be more specific about what CMake versions are not usable with the official build toolset (v142). If we know about compatibility with other toolsets then we can optionally add that information somewhere, too.

> only when using CMake 3.21.0 and with Visual Studio 16.10.4 with the compiler issue

Is this an issue in the Visual Studio IDE or the build toolset? I guess it's the latter, so we should specify the toolset and exact compiler version.

PR #6084, opened 2021-12-17
Updating script repository with a command to enable crosshair visibility

koeglfryderyk:
Added a line of code that enables crosshair visibility


pieper (approved):
Thanks!

PR #6086, opened 2021-12-19
ENH: added markups node events test

kerim371:
It is aimed to test that events are triggered when appropriate methods are called.
`PointPositionMissingEvent`s and `InteractionEvent`s are not tested yet


kerim371:
Can't understand why the build process fail.
It seems to successively build on my PC (tested it using SlicerCAT)

lassoan:
I've fixed the test and committed in 7610ddea1b40b569641c77964976718be045dde7.

PR #6088, opened 2021-12-21
BUG: duplicated subfolder on scripted modules

fbordignon:
scripted modules were getting a subfolder with duplicated content on final windows build.
image


fbordignon:
I've downloaded the linux nightly build from slicer.org and confirmed that it happens on linux as well.
I did not test if the proposal breaks something on linux, but on windows it seems to work fine. In fact, if one deletes the subfolder in question from the nightly build, slicer works normally.

lassoan:
Yes, the folder is redundant in the install tree. Please check if your code change does not break scripted segment editor effects in the build tree. There is no duplication of the effect files in the build tree (`c:\D\S4R\Slicer-build\lib\Slicer-4.13\qt-scripted-modules\SegmentEditorEffects`) therefore if you remove a copy then it might be the one that is needed for the build tree - just need to be checked.

jcfr:

Also confirming locally that this will not cause unwanted side effects

lassoan (approved):
Thank you, it looks good.

fbordignon:
I can confirm that the segment editor effects works fine without the duplicated folder. The build works as well. Tested on win10

lassoan:
Thank you!

PR #6091, opened 2021-12-22
ENH: check if NumberOfFrames is 1

fedorov:
It is possible to encounter a non-empty NumberOfFrames with the value of "1".
Currently, this would result in a warning that DICOM is multiframe, which can
be confusing.


lassoan (approved):
Looks good to me, thank you.

PR #6094, opened 2021-12-28
Use markups roi in crop volume and volume rendering

lassoan:
- Allow CLI modules to use markups ROI as input
- Update Volume Rendering and Crop Volume modules to use markups ROI node by default
- Fix in markups node API (rename CenterPosition property to CenterOfRotation, because a Center property already exists for ROI)
- Fix in specify geometry popup (set the exact requested output extent)


pieper (approved):
Big improvement!  Thank for doing this.  I didn't test but it looks good.  I suggest merging it and testing in the preview and fixing any regressions that might pop up.

PR #6097, opened 2022-01-03
BUG: Fix Markups place widget not enabled when active node is not set

Sunderlandkyl:
When using a qSlicerMarkupsPlaceWidget, the place button would be disabled unless placement for the active place node was valid (vtkMRMLSelectionNode::GetActivePlaceNodePlacementValid). This could cause the place button to be disabled even when placement for the widget's current Markups node was valid.

Fixed by checking place mode validity for the current markups node in the widget, rather than the active node.


lassoan (approved):
Thank you, this looks good.

PR #6104, opened 2022-01-06
ENH: Added variant radius to markups curve

kerim371:
Before that line thickness of markups curve could only be fixed.
Now it is possible to use scalars values as a variant radius.
In this case we control:
- minimal radius (by absolute or relative value);
- maximum radius in the sense of multiple minimum radius (`vtkTubeFilter::RadiusFactor`)
Two aspects were not solved yet:
- if scalars contain negative values then it would be nice to add some offset before so that all values become > 0;
- seems that slice views may incorrectly display variant radius markups. Need to be checked in the future
[The discussion](https://discourse.slicer.org/t/implement-variable-radius-curve/21334/4)

### DIFF
--- a/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsDisplayNode.h
+++ b/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsDisplayNode.h
@@ -351,6 +351,14 @@ public:
   static const char* GetCurveLineSizeModeAsString(int mode);
   static int GetCurveLineSizeModeFromString(const char*);
 
+  /// Set variant radius for tube filter.
+  /// Values are given from 'VTK_VARY_RADIUS_' macro
+  vtkSetMacro(VaryRadius, int);
+  vtkGetMacro(VaryRadius, int);
+
+  vtkSetMacro(RadiusFactor, double);
+  vtkGetMacro(RadiusFactor, double);
+
   /// Configure line thickness
   /// Thickness is specified relative to markup point size
   /// (1.0 means line diameter is the same as diameter of point glyphs).
@@ -509,6 +517,9 @@ protected:
   int CurveLineSizeMode;
   double LineThickness;
   double LineDiameter;
+
+  int VaryRadius;
+  double RadiusFactor;
 
   double LineColorFadingStart;
   double LineColorFadingEnd;
--- a/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsDisplayNode.h
+++ b/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsDisplayNode.h
@@ -342,6 +342,14 @@ public:
     CurveLineSizeMode_Last // insert new types above this line
     };
 
+  /// Reimplements VTK_VARY_RADIUS_ macros
+  enum LineDiameterMode
+    {
+    Constant = 0, // VTK_VARY_RADIUS_OFF
+    FromScalars = 1, // VTK_VARY_RADIUS_BY_SCALAR
+    AbsoluteFromScalars = 3 // VTK_VARY_RADIUS_BY_ABSOLUTE_SCALAR
+    };
+
   /// Configure mode of determining line radius of markup curves.
   /// Default is relative thickness. Available modes in \sa CurveLineSizeModes
   vtkSetMacro(CurveLineSizeMode, int);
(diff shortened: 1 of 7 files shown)


kerim371:
Visual changes:
image

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsDisplayNode.h
```diff
@@ -351,6 +351,14 @@ class  VTK_SLICER_MARKUPS_MODULE_MRML_EXPORT vtkMRMLMarkupsDisplayNode : public
   static const char* GetCurveLineSizeModeAsString(int mode);
   static int GetCurveLineSizeModeFromString(const char*);
 
+  /// Set variant radius for tube filter.
+  /// Values are given from 'VTK_VARY_RADIUS_' macro
+  vtkSetMacro(VaryRadius, int);
```
lassoan, line 356: It should be something like `LineDiameterMode` (as we already have a `LineDiameter` member that controls the size).

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsDisplayNode.h
```diff
@@ -510,6 +518,9 @@ class  VTK_SLICER_MARKUPS_MODULE_MRML_EXPORT vtkMRMLMarkupsDisplayNode : public
   double LineThickness;
   double LineDiameter;
 
+  int VaryRadius;
+  double RadiusFactor;
```
lassoan, line 522: This RadiusFactor should not be necessary for several reasons:
- If we need some kind of relative sizing then LineDiameter should be sufficient.
- If some non-trivial mapping is needed then I don't think factors are sufficient, but we would need an offset as well. This could be added later, if there is a strong driving use case.
- vtkTubeBender only supports a single scaling factor, so we need to compute a new array from the raw input array anyway.

lassoan (commented):
Thank you for working on this. It requires some updates, but it is a very promising start.

A few things to rework:
- We need to use absolute size mode (if a point has radius of 2.0mm then it should appear as 2.0mm). If you feel very strongly about the need for a relative mode then maybe it could be added as an additional mode.
- vtkTubeFilter distorts the cross-sections at sharp turns, which messes up the radius. It would be nice to try if vtkTubeBender works better and if there is any significant performance impact.
- See some more comments inline.

I'll convert this to a Draft to clearly indicate that this pull request needs some more work.

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsDisplayNode.h
kerim371: Ok, I think I can introduce `enum LineDiameterMode{ Constant = VTK_VARY_RADIUS_OFF(0), VaryByScalar = VTK_VARY_RADIUS_BY_SCALAR(1)  }`. There is also `VTK_VARY_RADIUS_BY_VECTOR(2), VTK_VARY_RADIUS_BY_ABSOLUTE_SCALAR(3), VTK_VARY_RADIUS_BY_VECTOR_NORM(4)` but I think they are unnecessary for now, maybe for the future

kerim371:
@lassoan I agree that we need to keep diameter primarily `absolute` but here are some circumstances about its implementation (I omit `vtkTubeBender` for now as I don't understand yet where and how to apply it).

vtkTubeFilter's `RadiusFactor` controls **the maximum tube radius in terms of a multiple of the minimum radius** and `vtkTubeFilter` has the default value `RadiusFactor = 10.`
Then to keep the radius absolute we should calculate the `RadiusFactor` from the scalars min/max values (we may get it via `vtkDoubleArray::GetValueRange()`) each time the scalars is modified. And probably each time the absolute/relative line thickness is modified. For example we can use absolute/relative line thickness as a thickness for the minimal diameter and the maximum will be set via calculated `RadiusFactor `.

But the main obstacle of this contribution is that for now markups curve doesnt't support adding custom scalars ([see Discourse](https://discourse.slicer.org/t/unable-to-add-scalars-to-vtkmrmlmarkupscurvenode/21341)). Only `PedigreIDs` can be set as scalars.

kerim371:
@lassoan on the picture below you can see that it is now possible to control whether Diameter mode:
- Constant;
- FromScalars;
- AbsoluteFromScalars

As you may notice `vtkTubeFilter` treats Scalars as radius for the tube, that is why instead of 9 mm diameter we have 18 mm.
Probably I should remane `LineDiameterMode` to `LineRadiusMode` ? That would break the markups terminology a little but be more logical from the physical point of view.
image

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsDisplayNode.h
kerim371: Agree, fixed

PR #6105, opened 2022-01-06
BUG: Fix truncation of Segmentations when saving at reference geometry

### LINKED ISSUE #6102: Segmentation may be cropped when saved
If some parts of segments fall outside of the reference geometry then the segmentation is cropped (information is lost) without warning.

Originally the error was reported here: https://discourse.slicer.org/t/editing-segmentation-of-model-and-saving-it/21338

## Steps to reproduce

- Load this segmentation: https://1drv.ms/u/s!Arm_AFxB9yqHx-FOPAelcgsOIud1qA?e=Zz7wZ7
- Save the segmentation with default settings into a new file
- Load the saved segmentation from the new file -> ERROR: the segmentation is cropped.

## Expected behavior

At the minimum, a warning should be displayed that explains that with this settings the segmentation will be cropped, and it would tell that the user should change reference geometry or enable "Crop to minimum extent" option.

If such segmentation (that has segments outside the reference geometry) can be easily created using latest Slicer Preview Release (e.g., by moving segments between segmentations) then we could consider expanding the extents automatically (with the effective extent) so that nothing is lost.

## Environment
- Slicer version: Slicer-4.13.0-2022-01-04

Sunderlandkyl:
Previously, saving a Segmentation with "Crop to minimum extent" disabled, the segmentation would be cropped to exactly the reference geometry extent. This could result in the segment being truncated if the reference geometry was smaller than the effective extent of the segmentation.

Fixed by saving the segmentation using an extent that contains both the reference and effective extents.
The same behavior is fixed when exporting segments to labelmap nodes.

Fix #6102
See discussion: https://discourse.slicer.org/t/editing-segmentation-of-model-and-saving-it/21338

### DIFF
--- a/Libs/MRML/Core/vtkMRMLSegmentationStorageNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLSegmentationStorageNode.cxx
@@ -1181,7 +1181,8 @@ int vtkMRMLSegmentationStorageNode::WriteBinaryLabelmapRepresentation(vtkMRMLSeg
   if (segmentation->GetNumberOfSegments() > 0)
     {
     std::string commonGeometryString = segmentation->DetermineCommonLabelmapGeometry(
-      this->CropToMinimumExtent ? vtkSegmentation::EXTENT_UNION_OF_EFFECTIVE_SEGMENTS : vtkSegmentation::EXTENT_REFERENCE_GEOMETRY);
+      this->CropToMinimumExtent ?
+        vtkSegmentation::EXTENT_UNION_OF_EFFECTIVE_SEGMENTS : vtkSegmentation::EXTENT_UNION_OF_EFFECTIVE_SEGMENTS_AND_REFERENCE_GEOMETRY);
     vtkSegmentationConverter::DeserializeImageGeometry(commonGeometryString, commonGeometryImage, true, scalarType, 1);
     commonGeometryImage->GetExtent(commonGeometryExtent);
     }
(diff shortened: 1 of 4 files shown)


lassoan:
Thanks a lot for the quick fix. I have a few questions:

1. When we export to labelmap in Segmentations module and choose a "Reference volume" is selected then do use that exact geometry (crop if reference volume is smaller than the segmentation)?

image

I think we should crop (or at least offer the user the option to crop) because it is a very frequent need to export a labelmap to a volume node that matches a volume exactly.

2. When we use "Export to files" and use NIFTI or NRRD and select a reference volume, do we crop to the exact reference volume extent? We should, as this is an export feature (it can be expected that some information is lost), maybe a warning could be displayed.

3. If we specify a source geometry in "Specify geometry" popup then is the segmentation cropped to that geometry's extent? It should, or at least there should be an option (it could be a checkbox in the specify geometry to window; and it would be nice to have a warning icon next to the checkbox if cropping will occur - the chosen geometry is smaller than the effective extent of the current segmentation).

Sunderlandkyl:
> 1. When we export to labelmap in Segmentations module and choose a "Reference volume" is selected then do use that exact geometry (crop if reference volume is smaller than the segmentation)? I think we should crop (or at least offer the user the option to crop) because it is a very frequent need to export a labelmap to a volume node that matches a volume exactly.

Yes. The only behavior that is changed is when the reference geometry is defined, and no reference volume is set.

> 2. When we use "Export to files" and use NIFTI or NRRD and select a reference volume, do we crop to the exact reference volume extent? We should, as this is an export feature (it can be expected that some information is lost), maybe a warning could be displayed.

Yes, same behavior as above. I can add a warning.

> 3. If we specify a source geometry in "Specify geometry" popup then is the segmentation cropped to that geometry's extent? It should, or at least there should be an option (it could be a checkbox in the specify geometry to window; and it would be nice to have a warning icon next to the checkbox if cropping will occur - the chosen geometry is smaller than the effective extent of the current segmentation).

No, but I was thinking the same thing. I can follow up with a PR adding this feature.

lassoan (commented):
Thanks for the answers. This is a simple, safe solution for the saving - I'll merge it now.

It'll be great if you add the optional cropping to "specify geometry".

lassoan (approved):

Issue #6074, opened 2021-12-08
Support exporting multiple nodes in one batch


ebrahimebrahim:
We should be able to export multiple nodes with a couple of clicks in the Data module.

This issue can be assigned to me.

## Describe the solution you'd like

- A variation of the "Export to file..." context menu item should be available when right clicking on a folder. This could export all nodes nested under the folder, if they have the same type.
- Perhaps also: right clicking after selecting multiple nodes should also offer the option to export them all, assuming they are of the same type.

In both cases, it should lead to the export dialog, which should be given some reasonable way of deciding how to name the nodes.

## Additional context

- [earlier PR for exporting individual nodes](https://github.com/Slicer/Slicer/pull/5876)
- useful earlier commit message: 3fcc1ebd13a150af335eb8d3918435c66668e2e6
- [announcement introducing export functionality](https://discourse.slicer.org/t/new-export-functionality/20950)

## Questions

1. How should filename be decided when multiple nodes are involved? Some ideas:
  a. Gray out the filename text entry box in the export dialog, and force reasonable default names. Maybe just use the node's name, and perhaps preferring the filename in the associated storage node if it exists.
  b. Use the filename text entry box in the export dialog to allow a choice of "base name" that could be prepended or appended to the names of the exported nodes.
2. How should we handle the case where nodes have different types? Some ideas:
  a. Don't provide export option in the context menu in this case.
  b. Provide option, but raise an error telling user to select nodes having all the same type.
  c. Let the majority node type win, only exporting nodes of that type.
  d. Raise a warning that this is what will be done.
  e. Raise a dialog that allows the user to specify which node type they want to export.
  f. Tweak the export dialog to somehow be able to handle nodes of different types.
3. In the case of exporting a folder, should the entire hierarchy be traversed? Or only the direct children of the folder?

lassoan:
> How should filename be decided when multiple nodes are involved?

Option (a) sounds good to me. You would replace the filename selector by a folder selector.
Also add an option to recreate the subject hierarchy folder structure as file folder structure (it can prevent name clashes).

> How should we handle the case where nodes have different types?

I feel all the options unnecessarily inconvenient for the user. This feature is so useful and powerful that it would be a pity if having some extra node in a folder would interfere with the export.

Could we show for each node class: node type (label), file extension (combobox), options (custom widget)?

> In the case of exporting a folder, should the entire hierarchy be traversed? Or only the direct children of the folder?

[...]

PR #6107, opened 2022-01-10
ENH: Exporting multiple nodes

ebrahimebrahim:
# Description

This PR addresses #6074 and extends the export functionality from #5876 by allowing multiple nodes to be exported in one batch. If you right click a a subject hierarchy item that has children, then "Export to file..." will allow you export the storable nodes that occur in the subtree rooted at your selection. You can choose to include children recursively, or to just include direct children. Multi-select is not supported here; only selecting a single subject hierarchy item with children can lead to the multiple-node-export functionality.

image

Each node _type_ occurring within the set of nodes slated for export will have its own options widget and its own export format dropdown box. 

image

The example in these screenshots does not show the "include children" or "recursively include children" checkboxes, because in this particular example:
- not including children would result in no nodes to export, so it is not an option
- the subtree rooted at the selected node has depth 1, so it doesn't make a difference whether you recursively include children or not-- hence the option is not displayed

So these options are only displayed when they matter. For example:

image

Notice that the filename entry textbox is disabled when multiple nodes are being exported. In that case, filenames are automatically set based on node name. The textbox is enabled when exporting a single node.

# Summary of changes
- The subject hierarchy export plugin has been modified to pass along lists of node IDs, rather than just a single node ID.
- The export dialog has been heavily reworked to handle lists of nodes. There is a checkbox for including children and a checkbox for doing so recursively. When nodes of multiple types are included, some widgets get created per node type.
- The export function in `qSlicerCoreIOManager` has been modified to accept a list of parameter sets, rather than just one parameter set. This allows us to pass to it multiple formats for different node types.
- A serious export-related bug (that I introduced) has been fixed where node IDs were sometimes being looked up in the main scene rather than in the temporary scene used for export.
- A mild export-related bug has been fixed where plane markups would get written by `vtkMRMLMarkupsJsonStorageNode` rather than `vtkMRMLMarkupsPlaneJsonStorageNode`, and similarly for ROIs. This caused them to be written incorrectly.

# Problems found

1. I found one case where exporting a transform fails silently, because success is returned by the storage node even when nothing is written. To see this happen, create a `vtkMRMLTransformNode` (not a subtype of it) and immediately try to export it, without modifying its parameters. Success is returned [because of these lines](https://github.com/Slicer/Slicer/blob/master/Libs/MRML/Core/vtkMRMLTransformStorageNode.cxx#L489-L493), whose intention I'm not understanding.

_This problem does not prevent this contribution from being merged._ It is non-critical and I think its a separate issue.

2. "Node type" currently means "class name". One possibly annoying consequence of this is that different types of markups (line, closed curve, etc.) are considered to be different node types, and therefore get their own export format dropdown and options widget, while in most cases the user would want the same settings for all of them. So when you go to export a folder that contains all your markups, you have to set your desired options separately for each markup subtype. Perhaps the same issue applies to different types of transforms. 

_This problem may need to be resolved_, depending on feedback as to how much of a problem it is. I'm looking for suggestions here. It would not be hard to redefine "node type" in my code, but I'm not sure what definition would make the most sense. It would be nice if there were a solution to this that doesn't treat markups as a special case.

### DIFF
--- a/Base/QTCore/qSlicerCoreIOManager.cxx
+++ b/Base/QTCore/qSlicerCoreIOManager.cxx
@@ -60,7 +60,11 @@ public:
   qSlicerFileReader* reader(const QString& fileName)const;
   QList<qSlicerFileReader*> readers(const QString& fileName)const;
 
-  QList<qSlicerFileWriter*> writers(const qSlicerIO::IOFileType &fileType, const qSlicerIO::IOProperties& parameters)const;
+  QList<qSlicerFileWriter*> writers(
+    const qSlicerIO::IOFileType &fileType,
+    const qSlicerIO::IOProperties& parameters,
+    vtkMRMLScene* scene = nullptr
+  ) const;
 
   QSettings*        ExtensionFileType;
   QList<qSlicerFileReader*> Readers;
@@ -121,12 +125,24 @@ QList<qSlicerFileReader*> qSlicerCoreIOManagerPrivate::readers(const QString& fi
 
 //-----------------------------------------------------------------------------
 QList<qSlicerFileWriter*> qSlicerCoreIOManagerPrivate::writers(
-    const qSlicerIO::IOFileType& fileType, const qSlicerIO::IOProperties& parameters)const
+  const qSlicerIO::IOFileType& fileType,
+  const qSlicerIO::IOProperties& parameters,
+  vtkMRMLScene* scene /*=nullptr*/
+) const
 {
   QString fileName = parameters.value("fileName").toString();
   QString nodeID = parameters.value("nodeID").toString();
 
-  vtkObject * object = this->currentScene()->GetNodeByID(nodeID.toUtf8());
+  if (!scene)
+    {
+    scene = this->currentScene();
+    }
+
+  vtkObject * object = scene->GetNodeByID(nodeID.toUtf8());
+  if (!object)
+    {
+    qWarning() << Q_FUNC_INFO << "warning: Unable to find node with ID" << nodeID << "in the given scene.";
+    }
   QFileInfo file(fileName);
 
   QList<qSlicerFileWriter*> matchingWriters;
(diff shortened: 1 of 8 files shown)


Base/QTCore/qSlicerCoreIOManager.cxx
```diff
@@ -880,34 +883,46 @@ bool qSlicerCoreIOManager::exportNodes(
...
+    // Copy parameters map; we will need to set the nodeID parameter to correspond to the node in the temporary scene
+    qSlicerIO::IOProperties temporarySceneParameters = parameters;
     temporarySceneParameters["nodeID"] = temporaryStorableNode->GetID();
-    temporarySceneParameters["fileName"] = filePaths[i];
 
     // Deduce "fileType" from "fileFormat" parameter; saveNodes will want both
-    qSlicerIO::IOFileType fileType = this->fileWriterFileType(storableNode, parameters.value("fileFormat").toString());
+    qSlicerIO::IOFileType fileType = this->fileWriterFileType(storableNode, parameters["fileFormat"].toString());
+
+    // Add default storage node into the temporary scene. This is sometimes needed.
```
ebrahimebrahim, line 905: This is the part that fixed the following bug I mentioned in my summary:
> A mild export-related bug has been fixed where plane markups would get written by `vtkMRMLMarkupsJsonStorageNode` rather than `vtkMRMLMarkupsPlaneJsonStorageNode`, and similarly for ROIs. This caused them to be written incorrectly.

ebrahimebrahim:
@lassoan This is ready for a first review :grinning:

ebrahimebrahim:
Notes from 2022.01.11 meeting:
- play with the layout so that we don't get the words "export format" and "options" repeated so many times
- for now maybe it is okay that different markups get their own widgets, since chosen options are at least preserved within the slicer session

ebrahimebrahim:
@lassoan I adjusted the appearance to make it less noisy. Below is what it looks like now when there's a single node type vs multiple node types. Further suggestions are welcome!
export

ebrahimebrahim:
When exporting multiple nodes, the title now shows the number, e.g. "Export 12 Nodes," rather than just "Export Multiple Nodes." Thanks @jcfr for the suggestion!

(this pr is ready for review any time)

ebrahimebrahim:
@lassoan @pieper @jcfr @smrolfe @muratmaga
(just pinging some people to look for feedback and to indicate that this is ready for review :grinning:)

pieper (dismissed):
@ebrahimebrahim since Jc plans to merge the python changes today let's hold off another day before finishing the merge.

ebrahimebrahim:
@pieper Thank you for taking the time to review this!

I rebased to a more recent commit and gave this another few tests. I didn't have to change anything (besides doing a completely clean super-build after the python update).

@pieper @jcfr Are we good to merge now? Once the CI check is done.

pieper:
Okay from my end.  @jcfr did you learn everything you need about the python integration so that we are open to merging other PRs?

PR #6108, opened 2022-01-11
Updating minimum supported platforms

jamesobutler:
This PR is issuing some commits for updating minimum requirements to prep for Qt 6 Slicer work. Note that Qt 6.2 and later will be planned for use with Slicer rather than versions 6.0-6.1. Qt 6.0 requirements however are being set as minimum here though additional bumps in minimum requirements (dropping of macOS 10.14) may happen in Qt 6.3 or later.

~~The following commits may look familiar as they are also included in #5954 as they are required for updating to ITK 5.3.~~
- ~~`COMP: Require at least C++14`~~
- ~~`COMP: Update minimum CMake version from 3.13.4 to 3.16.3`~~

## Qt 6 needed changes
C++17 (corresponding compilers with support) such as Visual Studio 2017 15.7 and later (#6237)

**Platforms:**
Windows: Windows 10 and later
macOS: macOS 10.14 and later
Linux: Ubuntu 20.04 and later, CentOS 8.1 and later
Note that CentOS Linux 8 is EOL though ([source](https://www.centos.org/centos-linux-eol/))

---------------------------
`COMP: Update minimum required CMAKE_OSX_DEPLOYMENT_TARGET to 10.14`
Deployment target >= 10.14 is the one officially supported by Qt 6.0. See [Qt 6.0 supported platforms](https://doc.qt.io/archives/qt-6.0/supported-platforms.html).

This PR will have support for the following macOS versions:
- 12.x: Monterey (2021)
- 11.x: Big Sur (2020)
- 10.15.x: Catalina (2019)
- 10.14.x: Mojave (2018)

macOS 10.14 Mojave hardware requirements ([source](https://support.apple.com/kb/SP777), [source](https://support.apple.com/HT211683)):
- MacBook (Early 2015 or newer)
- MacBook Air (Mid 2012 or newer)
- MacBook Pro (Mid 2012 or newer)
- Mac mini (Late 2012 or newer)
- iMac (Late 2012 or newer)
- iMac Pro (2017)
- Mac Pro (Late 2013; Mid 2010 and Mid 2012 models with [recommended Metal-capable graphics cards](https://support.apple.com/HT208898))

According to https://gs.statcounter.com/os-version-market-share/macos/desktop/worldwide as of February 2022 this would mean there would be support for 89.02% of Mac devices worldwide.

---------------------------
## Testing
- ~~✅ C++17 built on Windows 10 using Visual Studio 2019 with v142 toolset completed successfully with this PR. All tests cases were run and passed successfully where the failing tests were the same as the already failing tests.~~
- C++20 built on Windows 10 using Visual Studio 2022 with v143 toolset failed in some dependencies such as ITK as C++20 has removed some syntax options that were deprecated in earlier C++ versions. This PR allows the standard to be set to C++20 as an option for testing.


jamesobutler:
@sjh26 https://github.com/Slicer/slicer-apidocs-builder/pull/2 was issued after observing API docs fail in this PR due to git protocol issue as was similarly fixed in https://github.com/Slicer/DashboardScripts/commit/fb444aa09f9b6610967def92f1fe8774ab96cf13.

hjmjohnson:
@jamesobutler  Can I help decouple a few of these issues? 

 I think that updating cmake to 3.16 could be independent of all other changes.

updating CMAKE_OSX_DEPLOYMENT_TARGET to 10.14 could be independent of all other changes.

Updating to C++17 could be independent of all other changes.

I think those 3 changes should be independent of the other changes, and will allow other decisions to be made with more targeted and with expansive testing (i.e. slower pace).

Let me know what you think.

Hans

jamesobutler:
@jcfr Now that Slicer has bumped to C++14 in https://github.com/Slicer/Slicer/commit/bc2b3f4049ef32486cf43ad16903820fc8767f27, what do you think about the consideration of these additional minimum requirement bumps in preparation for Qt6 in Slicer? This includes bump to C++17, and bump of minimum supported platforms as detailed in the original post.

jamesobutler:
@hjmjohnson @jcfr I have removed the commit bumping to C++17 to make that an independent change (now in #6237). What is left is supported platform version updates and ability to test with C++20.

jamesobutler:
After this `COMP: Update minimum required CMAKE_OSX_DEPLOYMENT_TARGET to 10.14` commit is merged we can revert https://github.com/Slicer/Slicer/commit/81517ef9a6a4a4d69efce347adc3fce9d03b785e and return to using latest `numpy` 1.22.2.

jamesobutler:
Would be interesting if https://download.slicer.org/download-stats/ included specific platform version in addition to the filter for Linux/Mac/Windows as could see if any special considerations if there are a majority of mac Slicer users are still using macOS 10.13 or something (I doubt it based on https://gs.statcounter.com/os-version-market-share/macos/desktop/worldwide).

jamesobutler:
Closing this PR now as commits have ultimately been split up into their own PRs.

PR #6110, opened 2022-01-12
BUG: tbb installation in `Debug` mode

kerim371:
`tbb` lib has naming convention for `Debug/Release` modes.
Slicer used to handle only `Release` tbb libs.
Thus if you try to install Slicer built in `Debug` then it will be impossible to run it.
Or you have to manually copy `Debug` libs to the installation folder.
This commit fixes that so two modes are supported: `Debug` and `Release`

### DIFF
--- a/CMake/SlicerBlockInstallTBB.cmake
+++ b/CMake/SlicerBlockInstallTBB.cmake
@@ -17,6 +17,14 @@ if(WIN32)
       ${TBB_BIN_DIR}/tbb.dll
       ${TBB_BIN_DIR}/tbbmalloc.dll
       ${TBB_BIN_DIR}/tbbmalloc_proxy.dll
+    CONFIGURATIONS Release
+    DESTINATION bin COMPONENT Runtime)
+  install(
+    FILES
+      ${TBB_BIN_DIR}/tbb_debug.dll
+      ${TBB_BIN_DIR}/tbbmalloc_debug.dll
+      ${TBB_BIN_DIR}/tbbmalloc_proxy_debug.dll
+    CONFIGURATIONS Debug
     DESTINATION bin COMPONENT Runtime)
 elseif(APPLE)
   install(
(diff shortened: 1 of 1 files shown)


lassoan (commented):
Thank you for working on this!

- Can you provide instructions for testing this?
- Have you checked that it works an all 3 platforms, in all 4 build modes (Debug, Release, MinSizeRel, RelWithDebInfo)?

kerim371:
@lassoan to test this we need to build Slicer with TBB enabled:
```bash
mkdir d
cd d
cmake .. -DCMAKE_BUILD_TYPE=Debug -DSlicer_VTK_SMP_IMPLEMENTATION_TYPE=TBB
cmake --build . -j8
cd Slicer-build
cpack -B ../../install/d
```
I only tested this on Ubuntu 20.04 with `Debug/Release` modes. I presume this won't work with `MinSizeRel` nor with `RelWithDebInfo`. I don't even have experience of building projects with such configurations.
Also at line `59` I use `if (CMAKE_BUILD_TYPE EQUAL "Debug")` and I'm not sure how this will behave on different platforms with different cmake versions if we say omit `CMAKE_BUILD_TYPE` when configuring the Slicer, or if we pass it as `CMAKE_BUILD_TYPE=DeBuG`.
As a result I think I only try to increase the chances to successively package Slicer in debug mode while the `Release` mode should not be affected. Theoretically...
The solution is mostly taken from [CMake docs](https://cmake.org/cmake/help/latest/command/install.html#:~:text=install(TARGETS%20target%0A%20%20%20%20%20%20%20%20CONFIGURATIONS%20Debug%0A%20%20%20%20%20%20%20%20RUNTIME%20DESTINATION%20Debug/bin)%0Ainstall(TARGETS%20target%0A%20%20%20%20%20%20%20%20CONFIGURATIONS%20Release%0A%20%20%20%20%20%20%20%20RUNTIME%20DESTINATION%20Release/bin))

lassoan (changes requested):
I've just realized that you want to install an application that is built in debug mode.

> Thus if you try to install Slicer built in Debug then it will be impossible to run it.

On Windows (but maybe on other operating systems as well), it is prohibited to redistribute application built in debug mode (see for example [here](https://docs.microsoft.com/en-us/visualstudio/releases/2022/redistribution)). Therefore, a debug-mode built-application on a different computer is expected to fail to start, because even if you added debug-mode TBB libraries to the installation package, the debug-mode MSVC runtime DLLs would be missing.

If you want to install a debug-mode built application on the same developer computer then it is legally allowed, but it only makes sense in very rare circumstances (for example, for some special testing of startup issues on an installed Slicer that you cannot reproduce in the build tree), so I would not modify the build system for this. Instead, you can address these special cases by coping the missing files manually. If this solution is not acceptable for you then we could consider forcing the application to always use release-mode TBB, even for a debug-mode application (similarly yo how it is done for CPython).

kerim371:
@lassoan thank you for information.

I think the manual copying the missing files is completely fine. Because my use case was to check whether the SlicerCAT is able to install my third party deps or not. And as I had only debug version of Slicer I tried to install it.

I think this PR is useless so we can close it. My appologize for disturbing you

PR #6112, opened 2022-01-14
ENH: Add warning messages when segmentation could be cropped

Sunderlandkyl:
When changing the segmentation geometry, or when exporting a segmentation to file/node using a reference volume, it is possible to crop the segmentation.
This may result in a loss of data if the cropped region included part of the segment.

This commit adds warning messages when user attempts an operation that will crop the segmentation. The messages will only appear if the effective extent of the segmentation is not fully contained in the reference geometry. Also adds a check box to qMRMLSegmentationGeometryDialog that allows the reference geometry to be set without cropping the segmentation.


Modules/Loadable/Segmentations/qSlicerSegmentationsModuleWidget.cxx
```diff
@@ -950,12 +954,42 @@ bool qSlicerSegmentationsModuleWidget::exportFromCurrentSegmentation()
...
+        {
+        ctkMessageBox* exportWarningMesssgeBox = new ctkMessageBox(this);
+        exportWarningMesssgeBox->setAttribute(Qt::WA_DeleteOnClose);
+        exportWarningMesssgeBox->setWindowTitle("Export may erase data");
+        exportWarningMesssgeBox->addButton(QMessageBox::StandardButton::Ok);
+        exportWarningMesssgeBox->addButton(QMessageBox::StandardButton::Cancel);
+        exportWarningMesssgeBox->setDontShowAgainVisible(true);
+        exportWarningMesssgeBox->setIcon(QMessageBox::Warning);
+        exportWarningMesssgeBox->setDontShowAgainSettingsKey("Segmentations/AlwaysCropDuringSegmentationNodeExport");
+        exportWarningMesssgeBox->setText("Reference volume does not cover the whole segmentation. "
```
lassoan, line 979: I'm not sure if this message is clear enough for everyone. What do you think about something like this?
"The current segmentation does not completely fit into the new geometry. Do you want to crop the segmentation?"

lassoan (commented):
Thank you, this will be great, I've just added a few comments inline.

Modules/Loadable/Segmentations/qSlicerSegmentationsModuleWidget.cxx
Sunderlandkyl: Sure that message sounds good to me.

Sunderlandkyl:
Comments should be addressed.

lassoan (approved):
Thanks a lot, it looks good. Should we add a few sentences to the user manual, too?

Sunderlandkyl:
Sure, I'll add it to the user manual.

Sunderlandkyl:
Added a couple of sentences to the user doc.

lassoan (approved):
Perfect, thank you.

Issue #6111, opened 2022-01-14
Improve color legend


lassoan:
## Is your feature request related to a problem? Please describe.

Color legend (scalar bar, color bar) feature was completely reworked in [Slicer Preview Release 2022-01-15](https://github.com/Slicer/Slicer/pull/5828/commits/170a0e0edd448fd2c5d8e47ac6de591293fc8354). This issue tracks improvement ideas of this feature:

- [ ] BUG: Built-in procedural color nodes (such as PET-DICOM) should not be editable, but they are
- [ ] BUG: Depending on how the text is scaled (even which side the right label positions are snapped to), the labels sometimes vertically shifted (the higher the label the more it shifts)
- [ ] Remove Color Scalar bar from DataProbe module - it is not needed anymore, and it would reduce the amount of code and related maintenance efforts.
- [ ] Skip empty colors. Color tables, such as `AbdomenColors` contain many empty colors (appear with `(none)` label). In Colors module, there is a "Hide empty colors" option, which hides these from the Colors module GUI. We should hide these colors from the color legend, too. Color table exported from a segmentation contains "Background" label for 0, this should be probably replaced by empty label, too, to prevent it from showing up in the color legend.
- [ ] Add view context menu (e.g., when right-clicking on a volume in a slice view) to show/hide color legend for the volume that is visible at that position, to make color legend feature more discoverable. Probably it could be added to qSlicerSubjectHierarchyVolumesPlugin.
- [ ] Improve visibility of scalar bar for volumes in 3D view. When a slice view is shown in a 3D view then show scalar bars of any volumes that are displayed in those slices.
- [ ] Interactive positioning/sizing of the color legend. It could be implemented by creating a VTKWidget that could capture mouse events when the mouse is over the color legend actor and allow translating (and maybe resizing) using drag-and-drop.
- [ ] Add colormap display to segmentation nodes. For this, we would need a color node, which would contain the segment colors. Either the segmentation display node would update the color node when something is changed, or the color node could be a new class (vtkMRMLSegmentationColorNode) which would observe the associated segmentation display node and update itself automatically when changed.
- [ ] When using color nodes, such as PET-DICOM, number of colors are taken from the number of control points, but they are equally distanced. It would be better to either allow setting the positions (not equally distanced) or allow setting the number of colors
[...]

PR #6114, opened 2022-01-16
BUG: Fix vertical shift for labels

MichaelColonel:
Possible fix for a bug listed in #6111.

### DIFF
--- a/Modules/Loadable/Colors/MRMLDM/vtkMRMLColorLegendDisplayableManager.cxx
+++ b/Modules/Loadable/Colors/MRMLDM/vtkMRMLColorLegendDisplayableManager.cxx
@@ -287,25 +287,23 @@ bool vtkMRMLColorLegendDisplayableManager::vtkInternal::UpdateActor(vtkMRMLColor
     {
     case vtkMRMLColorLegendDisplayNode::Vertical:
       actor->SetOrientationToVertical();
-      actor->SetPosition(position[0] * (1 - size[0]), position[1] * (1 - size[1]));
+      actor->SetPosition(position[0] * (1. - size[0]), position[1] * (1. - size[1]));
       actor->SetWidth(size[0]);
       actor->SetHeight(size[1]);
       if (position[0] < 0.5)
         {
         actor->SetTextPositionToSucceedScalarBar();
-        actor->SetTextPad(2); // make some space between the bar and labels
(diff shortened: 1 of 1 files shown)


lassoan (commented):
Thanks for investigating this. The text padding improves the appearance of the legend (without it the label and colorbar are touching), so I would rather keep it. There seems to be a bug in the Scalar bar actor that makes padding incorrectly shifts the labels or the top of the colorbar in vertical direction. This bug in the scalar bar actor should be fixed instead.

There is also a ParaView scalar bar actor (https://github.com/Slicer/Slicer/blob/d36e96bd544db2489d624c85727194cde6da7484/Modules/Scripted/DataProbe/Logic/vtkPVScalarBarActor.h) that might work better than the scalar bar actor in VTK core. It would be an iron to switch to that. 

vtkPVScalarBarActor is a copy from ParaView: the one in ParaView may have been further improved since it was copied into Slicer. It could worth taking a look at what changes were done in ParaView and port them into Slicer if they seem useful.

MichaelColonel:
I will look at ScalarBarActor in DataProbe module and ParaView.

Question: Why the pad size is negative [negative](https://github.com/Slicer/Slicer/blob/master/Modules/Loadable/Colors/MRMLDM/vtkMRMLColorLegendDisplayableManager.cxx#L302) for vertical orientation? Usually such values are >= 0, and there is no vertical shift for labels is case of positive value. 

`        actor->SetTextPad(-2); // make some space between the bar and labels `

lassoan:
The padding is always zero when the orientation is horizontal. When the orientation is vertical then the sign depends on which side of the view the legend is closer to, because we always make the labels appear towards the center of the view (because longer labels somewhat sticking out towards the center of the view is tolerable but displaying the colorbar farther from the view border is very bad). When we change the "precede" setting (labels precede colorbar or colorbar precede labels) then the positive padding does not add horizontal space anymore but it removes space (makes the labels overlap with the colorbar). Probably instead of specifying negative padding, the scalar bar actor should be fixed to make positive padding add horizontal space regardless of "precede" setting.

The Slicer version of the scalar bar supports string labels that have the minimum and maximum labels centered on the color (normally these labels are displayed at the edge of the colorbar not at the center of the last color), so if we switch to the ParaView scalar bar actor then this feature might need to be added (unless that mode is already supported).

PR #6120, opened 2022-01-20
ENH: reset IJK matrix in volume information

riep:
IJK matrix is not reset when selecting an empty volume node

### DIFF
--- a/Libs/MRML/Widgets/qMRMLVolumeInfoWidget.cxx
+++ b/Libs/MRML/Widgets/qMRMLVolumeInfoWidget.cxx
@@ -186,6 +186,15 @@ void qMRMLVolumeInfoWidget::updateWidgetFromMRML()
     double origin[3] = {0.,0.,0.};
     d->ImageOriginWidget->setCoordinates(origin);
 
+    // Set IJK to RAS direction matrix to identity
+    for (int i=0; i<3; i++)
+      {
+      for (int j=0; j<3; j++)
+        {
+        d->IJKToRASDirectionMatrixWidget->setValue(i,j, i==j ? 1. : 0.);
+        }
+      }
+
     d->ScalarRangeValueLabel->setText("");
 
     d->ScanOrderComboBox->setCurrentIndex(-1);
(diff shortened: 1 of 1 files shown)


Libs/MRML/Widgets/qMRMLVolumeInfoWidget.cxx
```diff
@@ -186,6 +186,15 @@ void qMRMLVolumeInfoWidget::updateWidgetFromMRML()
     double origin[3] = {0.,0.,0.};
     d->ImageOriginWidget->setCoordinates(origin);
 
+    double IJKToRASDirections[3][3] = { {1,0,0}, {0,1,0}, {0,0,1} };
```
lassoan, line 189: Double values should be initialized with double (`1.0` or just `1.`). But I would rather remove this line and just use `i==j ? 1.0 : 0.0`). Also add a comment that we are setting the matrix to identity.

Libs/MRML/Widgets/qMRMLVolumeInfoWidget.cxx
```diff
@@ -186,6 +186,15 @@ void qMRMLVolumeInfoWidget::updateWidgetFromMRML()
...
+    for (int i=0; i<3; i++)
+    {
```
lassoan, line 191: I don't like Slicer's current brace indentation style, but for consistency, please use that (indent the `{` by two spaces).

lassoan (changes requested):
Thank you, it looks good, just added two nitpicky comment inline (they are very easy to fix).

riep:
corrections done

Libs/MRML/Widgets/qMRMLVolumeInfoWidget.cxx
```diff
@@ -186,6 +186,16 @@ void qMRMLVolumeInfoWidget::updateWidgetFromMRML()
     double origin[3] = {0.,0.,0.};
     d->ImageOriginWidget->setCoordinates(origin);
 
+    // Set IJK to RAS direction matrix to identity
+    double IJKToRASDirections[3][3];
```
lassoan, line 190: Is `IJKToRASDirections` still needed?

riep: sorry, this is corrected

lassoan (approved):
Perfect, thank you.

PR #6121, opened 2022-01-20
BUG: Inverse zoom when ctrl+wheel is used

riep:
**Steps to reproduce**
Load Volume
keep control + mouse wheel backward in a 2D view

**Expected result:**
Image becomes bigger during movement, in order to be consistent with other zooms (wheel backward in volume rendering view) 

**Actual result:**
Image becomes smaller during movement

### DIFF
--- a/Libs/MRML/DisplayableManager/vtkMRMLCameraWidget.cxx
+++ b/Libs/MRML/DisplayableManager/vtkMRMLCameraWidget.cxx
@@ -129,8 +129,13 @@ vtkMRMLCameraWidget::vtkMRMLCameraWidget()
     WidgetStateScale, WidgetEventScaleStart, WidgetEventScaleEnd);
   this->SetEventTranslationClickAndDrag(WidgetStateIdle, vtkCommand::RightButtonPressEvent, vtkEvent::NoModifier,
     WidgetStateScale, WidgetEventScaleStart, WidgetEventScaleEnd);
+  // Mousewheel zooms direction is chosen to be consistent with other applications (Ctrl-MouseWheelForward zooms in in PowerPoint),
+  // and with Windows touchpad pinch-to-zoom (pinch to zoom in generates MouseWheelForward event).
   this->SetEventTranslation(WidgetStateIdle, vtkCommand::MouseWheelForwardEvent, vtkEvent::NoModifier, WidgetEventCameraWheelZoomIn);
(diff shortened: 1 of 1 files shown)


jamesobutler (commented):
> Image becomes bigger during movement, in order to be consistent with other zooms (wheel backward in volume rendering view)

I don't think matching the zoom behavior of the 3D View is actually what should be done here. Control+mouse wheel backwards (towards the user) should be zoom out behavior. This is currently correct in a 2D View. The fact that the 3D View will zoom-in when mouse wheel backwards (towards the user), is unexpected and confusing to me. Mouse wheel backwards on say google maps will zoom out view so this appears to be default standard behavior for most users using a mouse wheel.

I was testing current behavior with latest preview Slicer 4.13.0-2021-01-19 on Windows.

I wonder if this mismatch is a confusion between trackpad scrolling and mouse wheel scrolling. My Windows laptop trackpad default settings has an action when two-finger swipe up (away from user) results in the same behavior as external mouse wheel down (towards the user).

lassoan:
Yes, slice view works correctly. Ctrl+MouseWheelBackward zooming out is consistent with how other applications work (such as PowerPoint), and results in correct zoom in/out with the touchpad pinch event on Windows (pinch-to-zoom-in generates Ctrl+MouseWheelForwardEvent).

Ctrl-wheel GUI event is not mapped to any widget event in 3D views, so there is no direct inconsistency. But I agree that it would be nice to have consistent zoom direction of Ctrl-wheel in slice views and wheel in 3D views. The zoom direction of wheeling was chosen to be consistent with the right-click-and-drag action: "pulling back the finger on the wheel" and "pulling back the whole mouse while holding right button down" both zooms in the view. This slight inconsistency is present in slice views as well and people don't complain (although Ctrl-wheel is probably much less used than simple wheel), so it should be acceptable for 3D views as well.

We could consider flipping the right-click-and-drag zoom direction to be consistent with the mouse wheel direction. However, the mouse range of motion from neutral position to forward is about 3x shorter than from neutral position to backward, so inverting the zoom would make it much harder to quickly zoom in to see some small detail and the zoom back. It is also a very commonly used gesture and users would need to re-train muscle memory.

So, flipping mousewheel zoom direction in 3D views and adding Ctrl-mousewheel action (for pinch-to-zoom on Windows touchpads) seem to be the right things to do.

lassoan:
I've force-pushed the suggested change to this branch. It feels good to me but please @riep @jamesobutler test if you have a chance.

riep:
Thanks both of you, I will test it on Linux and windows

jamesobutler (approved):
Yes, I've tested and this mouse-wheel behavior makes sense now in the 3D view.

PR #6050, opened 2021-11-30
Improve i18n support

jcfr:
This PR supersedes https://github.com/Slicer/SlicerGitSVNArchive/pull/1162 and should be consolidated with work contributed in https://github.com/Slicer/Slicer/pull/5021

PR #6123, opened 2022-01-20
Enable i18n

lassoan:
Build Slicer with i18n enabled by default.

It is marked as draft because first need to merge this CTK language box improvement (https://github.com/commontk/CTK/pull/1019), but otherwise it is fully ready for review.

It should get into today's nightly build so that it can be shown at the project presentations tomorrow.

### DIFF
--- a/Base/QTCore/qSlicerCoreApplication.cxx
+++ b/Base/QTCore/qSlicerCoreApplication.cxx
@@ -1941,25 +1941,43 @@ void qSlicerCoreApplication::loadTranslations(const QString& dir)
 }
 
 //----------------------------------------------------------------------------
+QStringList qSlicerCoreApplication::translationFolders()
+{
+#ifdef Slicer_BUILD_I18N_SUPPORT
+  qSlicerCoreApplication* app = qSlicerCoreApplication::application();
+  if (!app)
+    {
+    return QStringList();
+    }
+
+  QStringList qmDirs;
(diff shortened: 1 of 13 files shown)


pieper (approved):
LGTM 👍

pieper:
Do you want to add an update to the CTK tag in this PR too?

jcfr:
> Do you want to add an update to the CTK tag in this PR too?

That is indeed then plan (PR marked as draft)

jcfr:
I suggest to also cherry pick these commits from https://github.com/Slicer/Slicer/pull/6050:

```
ENH: i18n: Add support for generating translation templates
STYLE: i18n: Simplify *.ts files removing location 
```

as well as 

```
ENH: i18n: Add GenerateSlicerTranslationQMFiles target 
```

lassoan:
I agree, we should integrate the developments in #6050, but I cannot do that today. I will work on the module that can update translations from crowdin, as it would be nice to demo that tomorrow.

lassoan:
@sjh26 I've updated the default value of `Slicer_BUILD_I18N_SUPPORT` to `ON` in the latest master yesterday. Still, it seems that Slicer was built with this option disabled (I've only tested on Windows). I did not find any reference to this CMake variable. Could you check on the factory machines? Maybe there is some cached configuration that preserves the previous value of `Slicer_BUILD_I18N_SUPPORT` and so it is not sufficient to just change the default value.

PR #6125, opened 2022-01-21
I18n add targets for generating qm files and translation templates

jcfr:
This PR integrates the more general changes originally proposed in https://github.com/Slicer/Slicer/pull/6050 (initially based of https://github.com/Slicer/SlicerGitSVNArchive/pull/1162)

In a nutshell, it adds the following:
* Generation of `<ModuleName>_untranslated.ts` files in `<CMAKE_BINARY_DIR>/TranslationTemplates`
* Add targets `GenerateSlicerTranslationTemplates` and `GenerateSlicerTranslationQMFiles`
* Support for generating and updating translation files in extension.

#### Translating python scripts

Placeholder build-system infrastructure for supporting translating python scripts has been added. This means that translation files associated with `*.py` scripts are not generated but once we finalize the convention and approach, it will be relatively straightforward to integrate it.

#### `GenerateSlicerTranslationQMFiles`

Currently the behavior of `GenerateSlicerTranslationQMFiles` depends on the value of `Slicer_UPDATE_TRANSLATION` variable:

| `Slicer_UPDATE_TRANSLATION` | behavior of `GenerateSlicerTranslationQMFiles` |
|--|--|
| `OFF` (default) | Generate `*.qm` files |
| `ON` | Generate or update `*.ts` files |


jcfr:
Generated template translation files for Slicer itself are listed below. This is not consistent with the current approach consisting in having a single large `Slicer.ts` file.

Additionally, each "component" translated files are currently expected to be located in the directory `${CMAKE_CURRENT_SOURCE_DIR}/Resources/Translations`.

To summarize, we need to consolidate the current approach.

I suggest the following:
* Consolidate modules `*.ts` files to have only `<ModuleName>_<lang>.ts` (instead of having one for each module library)
* Consolidate all `qSlicerBase....ts` and  `qSlicerApp...ts` into a single `SlicerCore.ts`

### Slicer translation templates

```
Annotations_untranslated.ts
Cameras_untranslated.ts
Colors_untranslated.ts
CropVolume_untranslated.ts
DataStore_untranslated.ts
Data_untranslated.ts
DynamicModeler_untranslated.ts
Markups_untranslated.ts
Models_untranslated.ts
MultiVolumeExplorer_untranslated.ts
Plots_untranslated.ts
qMRMLWidgets_untranslated.ts
qSlicerAnnotationsModuleWidgets_untranslated.ts
qSlicerApp_untranslated.ts
qSlicerBaseQTApp_untranslated.ts
qSlicerBaseQTCLI_untranslated.ts
qSlicerBaseQTCore_untranslated.ts
qSlicerBaseQTGUI_untranslated.ts
qSlicerColorsModuleWidgets_untranslated.ts
qSlicerDataStoreModuleWidgets_untranslated.ts
qSlicerDICOMLibModuleWidgets_untranslated.ts
qSlicerMarkupsModuleWidgets_untranslated.ts
qSlicerModelsModuleWidgets_untranslated.ts
qSlicerModulesCore_untranslated.ts
qSlicerPlotsModuleWidgets_untranslated.ts
qSlicerSegmentationsModuleWidgets_untranslated.ts
qSlicerSequencesModuleWidgets_untranslated.ts
qSlicerSubjectHierarchyModuleWidgets_untranslated.ts
qSlicerTablesModuleWidgets_untranslated.ts
qSlicerTemplateKeyModuleWidgets_untranslated.ts
qSlicerTerminologiesModuleWidgets_untranslated.ts
qSlicerTextsModuleWidgets_untranslated.ts
qSlicerTransformsModuleWidgets_untranslated.ts
qSlicerUnitsModuleWidgets_untranslated.ts
qSlicerVolumeRenderingModuleWidgets_untranslated.ts
qSlicerVolumesModuleWidgets_untranslated.ts
Reformat_untranslated.ts
SceneViews_untranslated.ts
Segmentations_untranslated.ts
Sequences_untranslated.ts
SubjectHierarchy_untranslated.ts
Tables_untranslated.ts
TemplateKey_untranslated.ts
Terminologies_untranslated.ts
Texts_untranslated.ts
Transforms_untranslated.ts
Units_untranslated.ts
ViewControllers_untranslated.ts
VolumeRendering_untranslated.ts
Volumes_untranslated.ts
Welcome_untranslated.ts
```

lassoan:
@jcfr Thanks a lot for working on this!

Based on a couple of days of experience, the single large .ts file seems to work well (and we could probably add a new .ts file for each extension).
- One big advantage of a single Slicer core translation file is that translators don't need to browse many .ts files but they can find all strings that they see on screen at a single location (they just type the text they see on screen and immediately see the text there and can translate it). If they would need to first find the file then it would be inconvenient, as it is often not clear which module provides a particular string that is visible on the Slicer GUI. Also, browsing 150 files just for the Slicer core (and hundreds more for the extensions) could be overwhelming (and just too much scrolling, too much selection of files, etc.).
- A slight advantage of many small .ts files would be that you would see the percentage completion for each module (while now we don't see which parts of Slicer are translated). Another advantage of many small .ts files could be that if we rebuild the translation template each time the module is built then the developer would get the translation-related errors/warnings immediately - but we could enable this and then merge the .ts files into one .ts file when uploading to crowdin.

I don't know if many small files are better or worse for updating the untranslated template files. I haven't experimented with crowdin untranslated file update yet, just uploaded the initial version.

I'll try your branch and we should discuss on the Slicer weekly meeting. There are many other open questions about how to set up the automatic updates, archival, and distribution of compiled .qm translation files, etc.

pieper:
Agreed - thanks Jc!

It's great to see the momentum on this and now is a good time to think through the options for automating as much of this as possible so that it's easier for everyone.  I agree that one big ts file appears to make it easier to use the crowdin interface.

lassoan:
It would be very useful to make lupdate fail while during the build step when `tr()` is used incorrectly, for example like this:

```
C:/D/S4/Modules/Loadable/Volumes/Widgets/qSlicerScalarVolumeDisplayWidget.cxx:113: Cannot invoke tr() like this
C:/D/S4/Modules/Loadable/Volumes/Widgets/qSlicerScalarVolumeDisplayWidget.cxx:117: Cannot invoke tr() like this
C:/D/S4/Modules/Loadable/Volumes/Widgets/qSlicerScalarVolumeDisplayWidget.cxx:117: Cannot invoke tr() like this
```

Unfortunately, lupdate does not have an option to make warnings cause failure (see bug report [here](https://bugreports.qt.io/browse/QTBUG-36478?attachmentOrder=asc)). We could detect errors by using `-silent` option and checking if the process output is empty, but I don't know if this is possible in `add_custom_command`.

lassoan:
Qt seems to recommend to concatenate translation files into a monolithic one - see https://doc.qt.io/qt-5/linguist-programmers.html#deploying-translations (although it is not completely clear for me what this is applicable to). A solution could be to generate many small translation files during build and concatenate them into monolithic files during packaging.

Another important aspect is translation of extensions. It should be possible to translate extensions (and test the translations - if there is enough space, everything is translated, etc.) without building anything. Preferably without installing CMake and Qt.

What do you think about bundling Qt lupdate, lconvert, lrelease the same way as we bundle Qt designer? That would allow us to create a Slicer module that could extract translatable strings from the code, compile ts files, maybe even download ts files from crowdin.

lassoan:
I've created a test project to experiment with things in crowdin and I've found that there is an option to translate all .ts files at once, which allows searching in all files. So, the "One big advantage..." that I mentioned above is not relevant.

image

lassoan:
I've created a topic for this on discourse, detailing requirements, proposed implementation, and a few questions: https://discourse.slicer.org/t/slicer-translation-workflow-using-crowdin/21601

jcfr:
Closing.

This pull-request intended to improve the generation of translation files through build targets (configured by CMake) has been superseded by:
* Slicer specific pull-request associated with the [domain:i18n](https://github.com/Slicer/Slicer/pulls?q=is%3Apr++label%3Adomain%3Ai18n+-label%3Astatus%3Ainvalid+-label%3Astatus%3Awontfix+).
* Introduction of `update_translations.py` script originally introduced through https://github.com/Slicer/Slicer/pull/6674
* Work done at https://github.com/Slicer/SlicerLanguagePacks

PR #6129, opened 2022-01-24
ENH: Add setMRMLScene to qMRMLCollapsibleButton

RafaelPalomar:
This adds the`setMRMLScene` function to `qMRMLCollapsibleButton`, as well
as the qt signaling to propagate the changes of the MRMLScene. This is
useful to propagate the change of MRMLScene to children of
`qMRMLCollapsibleButton`, which is particularly useful when defining UIs
with Qt designer.

### DIFF
--- a/Libs/MRML/Widgets/Testing/CMakeLists.txt
+++ b/Libs/MRML/Widgets/Testing/CMakeLists.txt
@@ -27,6 +27,7 @@ set(TEST_SOURCES
   qMRMLCheckableNodeComboBoxTest.cxx
   qMRMLCheckableNodeComboBoxTest1.cxx
   qMRMLClipNodeWidgetTest1.cxx
+  qMRMLCollapsibleButtonTest.cxx
   qMRMLColorListViewTest1.cxx
   qMRMLColorModelTest.cxx
   qMRMLColorModelTest1.cxx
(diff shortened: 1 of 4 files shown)


pieper:
Out of curiosity  how would the collapsible button change based on the scene?

RafaelPalomar:
Scene changes do not have any effect in the collapsible button itself. However, the collapsible button is a container of other widgets, which can change based on the scene (e.g., qMRMLWidgets).

jcfr:
I suggest to:
* Add `qMRMLObject` class (similar to `qSlicerObject`)
* Update `qSlicerObject` to inherit from `qMRMLObject`
* Update `qMRMLCollapsibleButton` to inherit from `qMRMLObject`

Having the container implementing the `MRML` interface is indeed helpful when using Qt designer to establish connections. 

Example: 
https://github.com/Slicer/Slicer/blob/8381ef3e6d73afc9f15ac0db5ddb7de753c90641/Base/QTCore/qSlicerIO.h#L41-L45

lassoan:
Do we really need a new qMRMLObject? Couldn't we use the existing qMRMLWidget instead? (maybe there could be issues due to multiple inheritance, I don't know)

RafaelPalomar:
> Couldn't we use the existing qMRMLWidget instead?

`qMRMLWidget` is based on `QWidget`, as well as `ctkCollapsibleButton` (indirectly). I haven't try to compile, but I suspect we will get [The Diamond Problem](https://en.wikipedia.org/wiki/Multiple_inheritance#The_diamond_problem) in multiple inheritance

lassoan:
I've cleaned up the formatting and added some more explanation to the class documentation and committed the changes: 31617f6503b3dc5e48f65927cfa3e4d93e298f42

@RafaelPalomar when you create pull requests in the future please allow push access for maintainers so we can commit the result of cleanups, squashes, etc. to the pull request's branch.

RafaelPalomar:
@lassoan, I don't get the option to allow modifications from maintainers when I create PRs for Slicer (I do get the option for other projects, though). It may have to do with GitHub permissions and groups, but I agree with you, this would be useful to have.

PR #6130, opened 2022-01-25
BUG: Fixes broken control point table jump slice function

smrolfe:
This commit corrects a discrepancy in the jump mode indexes. This fixes a bug that [disabled the jump slices function from the markups table](https://discourse.slicer.org/t/markups-jump-slices-functionality-is-broken-on-preview/21602) and a bug that [prevented jumping to locked points](https://discourse.slicer.org/t/syncing-markups-edit-state-with-control-point-lock/21606/2?u=smrolfe).

### DIFF
--- a/Modules/Loadable/Markups/qSlicerMarkupsModuleWidget.cxx
+++ b/Modules/Loadable/Markups/qSlicerMarkupsModuleWidget.cxx
@@ -86,9 +86,9 @@
 #include "vtkPoints.h"
 #include <math.h>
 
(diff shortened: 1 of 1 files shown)


smrolfe:
@lassoan Checking in on the status of this commit, let me know if there's anything else needed.

lassoan (approved):
Thanks for the fix, it works well.

I've reverted the risky change of enum value in slice node, as it was not necessary.

PR #6139, opened 2022-01-26
ENH: Add focus camera to control point right-click menu

hherhold:
Added context menu option for control points in 2d/3d view to focus camera on clicked-on point. Functionally equivalent to same menu option in control point list.


lassoan (approved):
This looks great, thank you!

Issue #5912, opened 2021-09-29
Markups: Fix varying coloring of markups ROI nodes


lassoan:
It is inconsistent that ROIs have all the same color if created one by one but have different colors if "Place multiple control points" is enabled.

Also, vtkMRMLColorTableNodeRandom color table is not designed for creating nice, distinguishable colors. Instead, use MediumChartColors, DarkBrightChartColors, LightPaleChartColors. Make the color index start at 0 (not at 3) and wrap around if the color index reaches the number of values in the color table. Remove UpdateCurrentColorTableIndex (fuse it into GetNewMarkupsColor). Make GetNewMarkupsColor return a double* (you need to add a `double LastReturnedMarkupsColor[3];` in the class to store the vector that is returned) and add `VTK_SIZEHINT(3)` to make the method conveniently usable from Python.

## Environment
- Slicer version: Slicer-4.11.0-2021-09-27
- Operating system: Windows

@smrolfe

lassoan:
@smrolfe what do you think?

smrolfe:
Thanks! I missed this. I will be able to look at it early next week.

smrolfe:
@lassoan This issue is addressed in the commit: https://github.com/smrolfe/Slicer/commit/55f9f5f23ceb091349d277b9ef97ed25e9475caa

smrolfe:
@lassoan @jamesobutler Checking in on the status of this commit, thanks!

jamesobutler:
@lassoan Are there additional things needed here for this issue? Or has https://github.com/Slicer/Slicer/commit/c8daef472497776cb9db67487c2664c384407d31 closed this issue?

lassoan:
Good catch. Yes, this can be closed now.

PR #6142, opened 2022-01-28
BUG: Adds varying colors to ROIs added as a single node

smrolfe:
This commit addresses issue: https://github.com/Slicer/Slicer/issues/5912

The color assigned to a ROI created using the single node placement buttons in the Markups module and Markups Toolbar are varied to be consistent with the varying color of nodes created using continuous placement mode.

The colormap used for the ROIs is changed from the vtkMRMLColorTableNodeRandom to MediumChartColors, which are more distinguishable.

### DIFF
--- a/Modules/Loadable/Markups/Logic/vtkSlicerMarkupsLogic.cxx
+++ b/Modules/Loadable/Markups/Logic/vtkSlicerMarkupsLogic.cxx
@@ -54,6 +54,7 @@
 
 // MRML includes
 #include "vtkMRMLCameraNode.h"
+#include "vtkMRMLColorTableNode.h"
 #include "vtkMRMLHierarchyNode.h"
 #include "vtkMRMLInteractionNode.h"
 #include "vtkMRMLMessageCollection.h"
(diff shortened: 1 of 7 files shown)


smrolfe:
One note on the implementation: I moved the GetNewMarkupsColor utility from the vtkMRMLMarkupsDisplayableManager to vtkMRMLScene for easier access from the Markups Toolbar and Markups Module. This is consistent with the GenerateUniqueName() utility which is called after node creation and seems conceptually similar.

lassoan (dismissed):
Thank you, it works well.

I've refactored the code to move unique markups node color generation from the scene to the markups module logic (no node-specific logic should be done in the scene). Also replaced all the markup node creation code blocks by a markups module logic call (to remove redundancy to reduce the amount of code and force consistent behavior).

lassoan:
@smrolfe @pieper @jamesobutler @jcfr What do you think enabling this feature for every markup type? It would mean that if you create multiple markups, each would have a different color by default.

jamesobutler:
Thinking about MarkupsLineNode for ruler functionality, I don't think having unique colors for each ruler would be something desired. Probably would stick with one color. Rulers are primarily identified uniquely by their labels rather than color.

jamesobutler:
@lassoan The changes from this PR as it relates to adding the markups node appears likely to be the reason why the following Slicer test cases are now failing
- https://slicer.cdash.org/test/21300104
- https://slicer.cdash.org/test/21299758
- https://slicer.cdash.org/test/21299754
- https://slicer.cdash.org/test/21299756

lassoan:
Thank you, I'll fix this now.

lassoan:
I've pushed a [fix](https://github.com/Slicer/Slicer/commit/6db0afbb7d29a9fc7e8340a862f7346241647383).

PR #6147, opened 2022-01-30
ENH: ColorScalarBar is removed from DataProbe module as a part of Issue #6111

MichaelColonel:
The vtkPVScalarBarActor is left untouched.
It might be merged later with vtkSlicerScalarBarActor from Colors module.


lassoan (approved):
This is perfect, thank you.

PR #6124, opened 2022-01-20
ENH: Add interactive slice intersections widget

cpinter:
If "Interaction" flag is enabled then slice intersections can be drag-and-dropped in slice views.

Re #5544

Co-authored-by: Andras Lasso <lasso@queensu.ca>
Co-authored-by: Csaba Pinter <pinter.csaba@gmail.com>

Continuation of https://github.com/Slicer/Slicer/pull/6008

PR #6148, opened 2022-01-31
DOC: Update snippet to turn on slice intersections

dgmato:
Update snippet to turn on slice intersections. This update is needed since the visibility of the slice intersections is not handled anymore from the _vtkMRMLSliceCompositeNode_. Instead, a new node type is used for that: _vtkMRMLSliceDisplayNode_.

More info regarding these changes: [here](https://github.com/Slicer/Slicer/pull/6124).


lassoan (approved):
Looks good to me, thank you!

PR #6153, opened 2022-02-03
ENH: Jump slices when markups interaction handle is clicked

Sunderlandkyl:
This commit adds event translations that will jump to the position of an interaction handle when it is left-clicked. This is the same behavior as with control points.

Re #5061


lassoan (approved):
Looks great, thank you.

PR #6156, opened 2022-02-05
COMP: Remove obsolete and unused CMake module SlicerFunctionToday

### LINKED ISSUE #1974: Package name / Slicer_BUILDDATE should be associated with last commit date instead of the current date of the day.
_This issue was created automatically from an original [Mantis Issue](https://mantisarchive.slicer.org/view.php?id=1974). Further discussion may take place here._

jcfr:
The module became unused following 4ab780076 (BUG: Associate package
name with the date of the last commit. Fixes #1974)

Additionally, starting with CMake 3.1, the TIMESTAMP operation
added to the string command allows to the retrieve current date
in a cross-platform fashion

See https://cmake.org/cmake/help/latest/command/string.html#timestamp


lassoan (approved):
Looks good to me.

PR #6160, opened 2022-02-05
ENH: Add startupWorkingPath property to qSlicerCoreApplication

lassoan:
The current working directory can be changed by any module at any time (for example, because some algorithms or executables use the current working directory for their inputs or outputs).
Therefore, modules that wanted to use the current working directory to interpret command-line arguments, or the Slicer Jupyter kernel that relies on knowing the working directory could not reliably get it.

A new property `startupWorkingPath` was added to qSlicerCoreApplication. This property is initialized from the current working directory at startup and cannot be changed later, therefore it is a reliable way of getting the original working directory.

The name is derived from "current working directory", with "current" replaced by "startup" (similarly to `startupEnvironment` property) and "directory" replaced by "path" (as all other directory path strings are referred to as `...path` in qSlicerCoreApplication and in Qt in general).

### DIFF
--- a/Base/QTCore/qSlicerCoreApplication.h
+++ b/Base/QTCore/qSlicerCoreApplication.h
@@ -74,6 +74,7 @@ class Q_SLICER_BASE_QTCORE_EXPORT qSlicerCoreApplication : public QApplication
   Q_PROPERTY(QString slicerSharePath READ slicerSharePath CONSTANT)
   Q_PROPERTY(QString temporaryPath READ temporaryPath WRITE setTemporaryPath)
   Q_PROPERTY(QString cachePath READ cachePath WRITE setCachePath)
+  Q_PROPERTY(QString startupWorkingPath READ startupWorkingPath CONSTANT)
   Q_PROPERTY(QString launcherExecutableFilePath READ launcherExecutableFilePath CONSTANT)
   Q_PROPERTY(QString launcherSettingsFilePath READ launcherSettingsFilePath CONSTANT)
   Q_PROPERTY(QString slicerDefaultSettingsFilePath READ slicerDefaultSettingsFilePath CONSTANT)
@@ -146,6 +147,9 @@ public:
   /// \sa repositoryRevision()
   /// \sa environment()
   Q_INVOKABLE QProcessEnvironment startupEnvironment() const;
+
+  /// Current working directory at the time the application was started.
+  QString startupWorkingPath() const;
 
   /// \brief Returns the current environment.
   ///
--- a/Base/QTCore/qSlicerCoreApplication.cxx
+++ b/Base/QTCore/qSlicerCoreApplication.cxx
@@ -163,6 +163,7 @@ qSlicerCoreApplicationPrivate::qSlicerCoreApplicationPrivate(
   this->DICOMDatabase = QSharedPointer<ctkDICOMDatabase>(new ctkDICOMDatabase);
 #endif
   this->NextResourceHandle = 0;
+  this->StartupWorkingPath = QDir::currentPath();
 }
 
 //-----------------------------------------------------------------------------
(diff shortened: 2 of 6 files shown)


lassoan:
This property was added because in recent SlicerJupyter versions, IPython is not used anymore (the much cleaner and lighter-weight xeus-python-shell is used instead) and the notebook directory is now only accessible by getting the current working directory.

pieper (approved):
Sounds good.

jcfr:
This makes sense.

> The current working directory can be changed by any module at any time

Indeed, and since there is no unified way of running processing methods, there is no way to wrap such calls using a context manager like [push_dir](https://github.com/scikit-build/scikit-build/blob/43c95b4f8fe677ff551122a099f31cd4472e1449/skbuild/utils/__init__.py#L77-L104) or object like [ctkScopedCurrentDir](https://github.com/commontk/CTK/blob/master/Libs/Core/ctkScopedCurrentDir.h#L33).

Somewhat related to this topic, it may may be sensible to also:
- update CLI infrastructure to restore working directory
- add the [push_dir](https://github.com/scikit-build/scikit-build/blob/43c95b4f8fe677ff551122a099f31cd4472e1449/skbuild/utils/__init__.py#L77-L104) (or similar name) context manager to `slicer.util`
- suggest developer writing generic python code to also include a util module with their context manager ... 
- suggest folks writing loadable module to use `ctkScopedCurrentDir`

jcfr:
Last, may be a good time to include a `working directory` section after the `runtime-environment-variables` section ?
See https://slicer--6160.org.readthedocs.build/en/6160/user_guide/settings.html?#runtime-environment-variables

lassoan:
I agree that making it easier to preserve the current working directory could be useful to reduce the chance that it is changed. But it is hard to enforce it in the entire Slicer source code and it is impossible to enforce in all extensions and Python packages. Therefore the current working directory remains an uncontrollable variable that nobody can rely on.

> Last, may be a good time to include a working directory section after the runtime-environment-variables section ?

Do you mean that we should describe that the current working directory can change anytime and this new application property can be used instead if someone needs the initial working directory?

jcfr:
>  describe that the current working directory can change anytime and this new application property can be used instead if someone needs the initial working directory?

Something  like

```
### Working directory

Similarly to other software, the current directory associated with Slicer corresponds to the folder from which one `Slicer` executable is started using either a relative or absolute path.

Considering it is challenging to enforce that both Slicer and all of the associated extensions consistently restore the current working directory (e.g after updating it to more conveniently write files in a specific directory), Slicer provide a convenient way to access the startup working path through the `startupWorkingPath` application property.

| Python | C++ |
|---|---|
| `slicer.app.startupWorkingPath` | `qSlicerCoreApplication::startupWorkingPath()` |
```

Base/QTCore/qSlicerCoreApplication.h
```diff
@@ -74,6 +74,7 @@ class Q_SLICER_BASE_QTCORE_EXPORT qSlicerCoreApplication : public QApplication
   Q_PROPERTY(QString slicerSharePath READ slicerSharePath CONSTANT)
   Q_PROPERTY(QString temporaryPath READ temporaryPath WRITE setTemporaryPath)
   Q_PROPERTY(QString cachePath READ cachePath WRITE setCachePath)
+  Q_PROPERTY(QString startupWorkingPath READ startupWorkingPath)
```
jcfr, line 77:
```suggestion
  Q_PROPERTY(QString startupWorkingPath READ startupWorkingPath CONSTANT)
```

---- 

_The presence of the CONSTANT attribute indicates that the property value is constant. For a given object instance, the READ method of a constant property must return the same value every time it is called. This constant value may be different for different instances of the object. A constant property cannot have a WRITE method or a NOTIFY signal._

Source: https://doc.qt.io/qt-5/properties.html

lassoan:
Thanks for all the feedbacks, I've added the CONSTANT qualifier and documentation as suggested.

PR #6162, opened 2022-02-07
ENH: Skip message boxes when testing

ebrahimebrahim:
# Summary

The message box utility functions such as `slicer.util.confirmOkCancelDisplay` are used in some extensions in such a way that tests can end up timing out and failing; this is due to the expectation of user interaction.

[SlicerAuto3dgm example that started this discussion](https://github.com/ToothAndClaw/SlicerAuto3dgm/issues/6)
[SlicerMorph example](https://slicer.cdash.org/test/21233619)

This commit addresses the issue by having message box functions such as `slicer.util.confirmOkCancelDisplay` behave differently in a testing situation: they can skip the dialog and return a default result.

# Tests

I tested my changes by running the tests for SlicerAuto3dgm and SlicerMorph and seeing that they indeed skip the message boxes, while still raising those message boxes outside of a testing situation. The failing SlicerMorph test still failed for me for other reasons, but at least it wasn't a timeout.

# Details

The following functions produce informational message boxes with no choice for the user, so they could simply be skipped under testing:

- `infoDisplay` 
- `warningDisplay` 
- `errorDisplay` 

The following functions provide a choice and have a return value, so in order to skip the user interaction I had to make a choice about the default value to be returned:

- `confirmOkCancelDisplay`: We default to Ok (return `True`), ~~but this can be changed by the caller with an optional argument~~
- `confirmRetryCloseDisplay`: We default to Cancel (return `False`).
  - Side note: I added error logging to this one, since the message text is typically some kind of error text.
- `confirmYesNoDisplay`: ~~Caller must specify the test-time return value in a required argument.~~ We default to `Yes`.

# Problem

~~The last one, `confirmYesNoDisplay`, I'm less sure about how to deal with. There's no good reason the user should expect a test to default to `Yes` or `No`, so it makes sense to me that they would have to specify the test-time value. However the required argument makes this an API change, and it makes `confirmYesNoDisplay` more annoying to use. That's why this is a draft at the moment.~~

Resolved; just going with `Yes` as a default.

### DIFF
--- a/Base/Python/slicer/util.py
+++ b/Base/Python/slicer/util.py
@@ -2306,7 +2306,8 @@ def delayDisplay(message, autoCloseMsec=1000):
 def infoDisplay(text, windowTitle=None, parent=None, standardButtons=None, **kwargs):
   """Display popup with a info message.
 
-  If there is no main window then the text is only logged (at info level).
+  If there is no main window, or if the application is running in testing mode (slicer.app.testingEnabled()==True),
+  then the text is only logged (at info level).
   """
   import qt, slicer
   import logging
@@ -2314,6 +2315,9 @@ def infoDisplay(text, windowTitle=None, parent=None, standardButtons=None, **kwa
     windowTitle = slicer.app.applicationName + " information"
   logging.info(text)
   mw = mainWindow()
+  if slicer.app.testingEnabled():
+    logging.info("Testing mode is enabled: Skipping info popup [%s]" % windowTitle)
+    return
   if mw:
     standardButtons = standardButtons if standardButtons else qt.QMessageBox.Ok
     messageBox(text, parent, windowTitle=windowTitle, icon=qt.QMessageBox.Information, standardButtons=standardButtons, **kwargs)
@@ -2321,7 +2325,8 @@ def infoDisplay(text, windowTitle=None, parent=None, standardButtons=None, **kwa
 def warningDisplay(text, windowTitle=None, parent=None, standardButtons=None, **kwargs):
   """Display popup with a warning message.
 
-  If there is no main window then the text is only logged (at warning level).
+  If there is no main window, or if the application is running in testing mode (slicer.app.testingEnabled()==True),
+  then the text is only logged (at warning level).
   """
   import qt, slicer
   import logging
@@ -2329,6 +2334,9 @@ def warningDisplay(text, windowTitle=None, parent=None, standardButtons=None, **
     windowTitle = slicer.app.applicationName + " warning"
   logging.warning(text)
   mw = mainWindow()
+  if slicer.app.testingEnabled():
+    logging.info("Testing mode is enabled: Skipping warning popup [%s]" % windowTitle)
+    return
   if mw:
     standardButtons = standardButtons if standardButtons else qt.QMessageBox.Ok
     messageBox(text, parent, windowTitle=windowTitle, icon=qt.QMessageBox.Warning, standardButtons=standardButtons, **kwargs)
@@ -2336,41 +2344,74 @@ def warningDisplay(text, windowTitle=None, parent=None, standardButtons=None, **
 def errorDisplay(text, windowTitle=None, parent=None, standardButtons=None, **kwargs):
   """Display an error popup.
 
-  If there is no main window then the text is only logged (at error level).
+  If there is no main window, or if the application is running in testing mode (slicer.app.testingEnabled()==True),
+  then the text is only logged (at error level).
   """
   import qt, slicer
   import logging
   if not windowTitle:
     windowTitle = slicer.app.applicationName + " error"
   logging.error(text)
+  if slicer.app.testingEnabled():
+    logging.info("Testing mode is enabled: Skipping error popup [%s]" % windowTitle)
+    return
   mw = mainWindow()
   if mw:
     standardButtons = standardButtons if standardButtons else qt.QMessageBox.Ok
     messageBox(text, parent, windowTitle=windowTitle, icon=qt.QMessageBox.Critical, standardButtons=standardButtons, **kwargs)
 
 def confirmOkCancelDisplay(text, windowTitle=None, parent=None, **kwargs):
-  """Display an confirmation popup. Return if confirmed with OK."""
+  """Display a confirmation popup. Return if confirmed with OK.
+
+  When the application is running in testing mode (slicer.app.testingEnabled()==True),
+  the popup is skipped and True ("Ok") is returned, with a message being logged to indicate this.
+  """
   import qt, slicer
+  import logging
   if not windowTitle:
     windowTitle = slicer.app.applicationName + " confirmation"
+  if slicer.app.testingEnabled():
+    result = True
+    logging.info("Testing mode is enabled: Skipping confirmation popup [%s] and returning %s." % (windowTitle, result))
+    return result
   result = messageBox(text, parent=parent, windowTitle=windowTitle, icon=qt.QMessageBox.Question,
                        standardButtons=qt.QMessageBox.Ok | qt.QMessageBox.Cancel, **kwargs)
   return result == qt.QMessageBox.Ok
 
 def confirmYesNoDisplay(text, windowTitle=None, parent=None, **kwargs):
-  """Display an confirmation popup. Return if confirmed with Yes."""
+  """Display a confirmation popup. Return if confirmed with Yes.
+
+  When the application is running in testing mode (slicer.app.testingEnabled()==True),
+  the popup is skipped and True ("Yes") is returned, with a message being logged to indicate this.
+  """
   import qt, slicer
+  import logging
   if not windowTitle:
     windowTitle = slicer.app.applicationName + " confirmation"
+  if slicer.app.testingEnabled():
+    result = True
+    logging.info("Testing mode is enabled: Skipping question popup [%s] and returning %s." % (windowTitle, result))
+    return result
   result = messageBox(text, parent=parent, windowTitle=windowTitle, icon=qt.QMessageBox.Question,
                        standardButtons=qt.QMessageBox.Yes | qt.QMessageBox.No, **kwargs)
   return result == qt.QMessageBox.Yes
 
 def confirmRetryCloseDisplay(text, windowTitle=None, parent=None, **kwargs):
-  """Display an confirmation popup. Return if confirmed with Retry."""
+  """Display an error popup asking whether to retry, logging the text at error level.
+  Return if confirmed with Retry.
+
+  When the application is running in testing mode (slicer.app.testingEnabled()==True),
+  the popup is skipped and False ("Close") is returned, with a message being logged to indicate this.
+  """
   import qt, slicer
+  import logging
+  logging.error(text)
   if not windowTitle:
     windowTitle = slicer.app.applicationName + " error"
+  if slicer.app.testingEnabled():
+    result = False
+    logging.info("Testing mode is enabled: Skipping retry popup [%s] and returning %s." % (windowTitle, result))
+    return result
   result = messageBox(text, parent=parent, windowTitle=windowTitle, icon=qt.QMessageBox.Critical,
                        standardButtons=qt.QMessageBox.Retry | qt.QMessageBox.Close, **kwargs)
   return result == qt.QMessageBox.Retry
--- a/Base/Python/slicer/util.py
+++ b/Base/Python/slicer/util.py
@@ -2306,115 +2306,105 @@ def delayDisplay(message, autoCloseMsec=1000):
 def infoDisplay(text, windowTitle=None, parent=None, standardButtons=None, **kwargs):
   """Display popup with a info message.
 
-  If there is no main window, or if the application is running in testing mode (slicer.app.testingEnabled()==True),
+  If there is no main window, or if the application is running in testing mode (``slicer.app.testingEnabled() == True``),
   then the text is only logged (at info level).
   """
-  import qt, slicer
-  import logging
-  if not windowTitle:
-    windowTitle = slicer.app.applicationName + " information"
-  logging.info(text)
-  mw = mainWindow()
-  if slicer.app.testingEnabled():
-    logging.info("Testing mode is enabled: Skipping info popup [%s]" % windowTitle)
-    return
-  if mw:
-    standardButtons = standardButtons if standardButtons else qt.QMessageBox.Ok
-    messageBox(text, parent, windowTitle=windowTitle, icon=qt.QMessageBox.Information, standardButtons=standardButtons, **kwargs)
+  import qt, logging
+  standardButtons = standardButtons if standardButtons else qt.QMessageBox.Ok
+  _messageDisplay(logging.INFO, text, None, parent=parent, windowTitle=windowTitle, mainWindowNeeded=True,
+                   icon=qt.QMessageBox.Information, standardButtons=standardButtons, **kwargs)
 
 def warningDisplay(text, windowTitle=None, parent=None, standardButtons=None, **kwargs):
   """Display popup with a warning message.
 
-  If there is no main window, or if the application is running in testing mode (slicer.app.testingEnabled()==True),
+  If there is no main window, or if the application is running in testing mode (``slicer.app.testingEnabled() == True``),
   then the text is only logged (at warning level).
   """
-  import qt, slicer
-  import logging
-  if not windowTitle:
-    windowTitle = slicer.app.applicationName + " warning"
-  logging.warning(text)
-  mw = mainWindow()
-  if slicer.app.testingEnabled():
-    logging.info("Testing mode is enabled: Skipping warning popup [%s]" % windowTitle)
-    return
-  if mw:
-    standardButtons = standardButtons if standardButtons else qt.QMessageBox.Ok
-    messageBox(text, parent, windowTitle=windowTitle, icon=qt.QMessageBox.Warning, standardButtons=standardButtons, **kwargs)
+  import qt, logging
+  standardButtons = standardButtons if standardButtons else qt.QMessageBox.Ok
+  _messageDisplay(logging.WARNING, text, None, parent=parent, windowTitle=windowTitle, mainWindowNeeded=True,
+                   icon=qt.QMessageBox.Warning, standardButtons=standardButtons, **kwargs)
 
 def errorDisplay(text, windowTitle=None, parent=None, standardButtons=None, **kwargs):
   """Display an error popup.
 
-  If there is no main window, or if the application is running in testing mode (slicer.app.testingEnabled()==True),
+  If there is no main window, or if the application is running in testing mode (``slicer.app.testingEnabled() == True``),
   then the text is only logged (at error level).
   """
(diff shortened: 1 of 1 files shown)


lassoan:
For me, it would make sense to either bypass all decision-making dialogs (ok/cancel, retry/close, yes/no) with a default choice; or not bypass any of them.

Bypassing makes things a bit simpler (developers don't need to think about popups blocking automatic tests), and since it is used only for testing, it does not seem risky. The default should be always the first choice that lets the execution continue (ok, close, yes), because the buttons are always ordered so that the first option is the most commonly used one.

We should not change the API. If developers don't like the default choice in the popup during testing then they can add an `if slicer.app.testingEnabled() ... else ...` branch there.

ebrahimebrahim:
> We should not change the API. If developers don't like the default choice in the popup during testing then they can add an `if slicer.app.testingEnabled() ... else ...` branch there.

This makes sense. I made some changes based on this and based on a discussion with @jcfr :
- I'm no longer introducing a "testing value" argument into the message box functions-- it cannot be controlled by caller in this way. If the caller really wants to control the default behavior then they can just `if slicer.app.testingEnabled()` themselves.
- Default for the Yes/No dialog is Yes because that's the first button.

Thank you both for the feedback!

This is now ready for review.

Base/Python/slicer/util.py
```diff
@@ -2337,7 +2343,10 @@ def errorDisplay(text, windowTitle=None, parent=None, standardButtons=None, **kw
   """Display an error popup.
 
   If there is no main window then the text is only logged (at error level).
+  Skipped during testing.
   """
+  if slicer.app.testingEnabled():
+    return
```
jcfr, line 2349: To make sure the test effectively covers code that would otherwise be executed, I suggest to move

```python
  if slicer.app.testingEnabled():
    logging.info("Testing mode is enabled: Skipping error popup [%s]" % windowTitle)
    return
```

just above the call to `messageBox` below.

Doing so would also make sure `logging.error(text)` is called.

Base/Python/slicer/util.py
```diff
@@ -2349,25 +2358,40 @@ def errorDisplay(text, windowTitle=None, parent=None, standardButtons=None, **kw
...
-  """Display an confirmation popup. Return if confirmed with OK."""
+  """Display a confirmation popup. Return if confirmed with OK.
+  While testing, this returns True ("Ok") without displaying any popup.
+  """
   import qt, slicer
+  if slicer.app.testingEnabled():
+    return True
   if not windowTitle:
     windowTitle = slicer.app.applicationName + " confirmation"
   result = messageBox(text, parent=parent, windowTitle=windowTitle, icon=qt.QMessageBox.Question,
```
jcfr, line 2369: Move block above `messageBox`

```python
  if slicer.app.testingEnabled():
    result = True
    logging.info("Testing mode is enabled: Skipping confirmation popup returning %s [%s]" % (result, windowTitle))
    return result
```

Base/Python/slicer/util.py
```diff
@@ -2349,25 +2358,40 @@ def errorDisplay(text, windowTitle=None, parent=None, standardButtons=None, **kw
...
-  """Display an confirmation popup. Return if confirmed with Yes."""
+  """Display a confirmation popup. Return if confirmed with Yes.
+  While testing, this returns True ("Yes") without displaying any popup.
+  """
   import qt, slicer
+  if slicer.app.testingEnabled():
+    return True
   if not windowTitle:
     windowTitle = slicer.app.applicationName + " confirmation"
   result = messageBox(text, parent=parent, windowTitle=windowTitle, icon=qt.QMessageBox.Question,
```
jcfr, line 2382: Move block above `messageBox`

```python
  if slicer.app.testingEnabled():
    result = True
    logging.info("Testing mode is enabled: Skipping confirmation popup returning %s [%s]" % (result, windowTitle))
    return result
```

Base/Python/slicer/util.py
```diff
@@ -2322,7 +2325,10 @@ def warningDisplay(text, windowTitle=None, parent=None, standardButtons=None, **
   """Display popup with a warning message.
 
   If there is no main window then the text is only logged (at warning level).
+  Skipped during testing.
   """
+  if slicer.app.testingEnabled():
+    return
```
jcfr, line 2331: To make sure the test effectively covers code that would otherwise be executed, I suggest to move

```python
  if slicer.app.testingEnabled():
    logging.info("Testing mode is enabled: Skipping warning popup [%s]" % windowTitle)
    return
```

just above the call to `messageBox` below.

Doing so would also make sure `logging.warning(text)` is called.

Base/Python/slicer/util.py
```diff
@@ -2307,7 +2307,10 @@ def infoDisplay(text, windowTitle=None, parent=None, standardButtons=None, **kwa
   """Display popup with a info message.
 
   If there is no main window then the text is only logged (at info level).
+  Skipped during testing.
   """
+  if slicer.app.testingEnabled():
+    return
```
jcfr, line 2313: To make sure the test effectively covers code that would otherwise be executed, I suggest to move

```python
  if slicer.app.testingEnabled():
    logging.info("Testing mode is enabled: Skipping info popup [%s]" % windowTitle)
    return
```

just above the call to `messageBox` below.

Doing so would also make sure `logging.info(text)` is called.

Base/Python/slicer/util.py
```diff
@@ -2307,7 +2307,10 @@ def infoDisplay(text, windowTitle=None, parent=None, standardButtons=None, **kwa
   """Display popup with a info message.
 
   If there is no main window then the text is only logged (at info level).
+  Skipped during testing.
```
jcfr, line 2310: Considering adding note like this:

```
.. note:: When testing mode is enabled, the confirmation dialog is skipped and a message indicating so is also logged.
```

jcfr (commented):

Few additional nitpicks (summary):

* move the skipping code further down
* log a message indicating that a dialog is skipped
* consider refactoring introducing function `slicer.util._messageDisplay(logLevel, text, windowTitle=None, parent=None, standardButtons=None, **kwargs)`
  * with `logLevel` being a value like `logging.INFO`, `logging.WARNING` or `logging.ERROR`
  * using `logging.log(logLevel, text)` instead of `logging.info(text)`, ...
  * for conversion from `logLevel` to text, consider using `logging.getLevelName(logLevel)`


Last, could you also:
* update the docstring associated with `slicer.util.MessageDialog` class to mention it is skipped when testing is enabled
* check that skipping effectively work for this class

Base/Python/slicer/util.py
lassoan: It might be simpler to say that if the application runs in testing mode then instead of showing the message in a popup, the message is just logged.

Base/Python/slicer/util.py
```diff
@@ -2349,25 +2358,40 @@ def errorDisplay(text, windowTitle=None, parent=None, standardButtons=None, **kw
     messageBox(text, parent, windowTitle=windowTitle, icon=qt.QMessageBox.Critical, standardButtons=standardButtons, **kwargs)
 
 def confirmOkCancelDisplay(text, windowTitle=None, parent=None, **kwargs):
-  """Display an confirmation popup. Return if confirmed with OK."""
+  """Display a confirmation popup. Return if confirmed with OK.
+  While testing, this returns True ("Ok") without displaying any popup.
```
lassoan, line 2362: "While testing" is not clear enough. Instead, you can write "when the application is running in testing mode (`slicer.app.testingEnabled()==True`)...". Update it not just here but in every function that we now update.

jcfr: Last, it may be good to add a new section to the developer guide to briefly describe the testing mode, and use of `slicer.app.testingEnabled()`. See https://github.com/Slicer/Slicer/issues/6164 for more details.

ebrahimebrahim:
> Last, could you also:
> 
> * update the docstring associated with `slicer.util.MessageDialog` class to mention it is skipped when testing is enabled
> * check that skipping effectively work for this class

Can you explain what you mean by

> * check that skipping effectively work for this class

Is there a scenario you have in mind that should be checked?

ebrahimebrahim:
Thank you for the feedback @jcfr, @lassoan!

I think all the items have been addressed. This is ready for another look any time.

Base/Python/slicer/util.py
```diff
@@ -2306,71 +2306,112 @@ def delayDisplay(message, autoCloseMsec=1000):
...
   result = messageBox(text, parent=parent, windowTitle=windowTitle, icon=qt.QMessageBox.Question,
                        standardButtons=qt.QMessageBox.Yes | qt.QMessageBox.No, **kwargs)
   return result == qt.QMessageBox.Yes
 
 def confirmRetryCloseDisplay(text, windowTitle=None, parent=None, **kwargs):
-  """Display an confirmation popup. Return if confirmed with Retry."""
+  """Display an error popup asking whether to retry, logging the text at error level.
+  Return if confirmed with Retry.
+
+  When the application is running in testing mode (slicer.app.testingEnabled()==True),
```
jcfr, line 2369:
```suggestion
  When the application is running in testing mode (``slicer.app.testingEnabled() == True``),
```

Base/Python/slicer/util.py
```diff
@@ -2306,71 +2306,112 @@ def delayDisplay(message, autoCloseMsec=1000):
...
+    return result
   result = messageBox(text, parent=parent, windowTitle=windowTitle, icon=qt.QMessageBox.Question,
                        standardButtons=qt.QMessageBox.Ok | qt.QMessageBox.Cancel, **kwargs)
   return result == qt.QMessageBox.Ok
 
 def confirmYesNoDisplay(text, windowTitle=None, parent=None, **kwargs):
-  """Display an confirmation popup. Return if confirmed with Yes."""
+  """Display a confirmation popup. Return if confirmed with Yes.
+
+  When the application is running in testing mode (slicer.app.testingEnabled()==True),
```
jcfr, line 2355:
```suggestion
  When the application is running in testing mode (``slicer.app.testingEnabled() == True``),
```

Base/Python/slicer/util.py
```diff
@@ -2306,71 +2306,112 @@ def delayDisplay(message, autoCloseMsec=1000):
...
   mw = mainWindow()
   if mw:
     standardButtons = standardButtons if standardButtons else qt.QMessageBox.Ok
     messageBox(text, parent, windowTitle=windowTitle, icon=qt.QMessageBox.Critical, standardButtons=standardButtons, **kwargs)
 
 def confirmOkCancelDisplay(text, windowTitle=None, parent=None, **kwargs):
-  """Display an confirmation popup. Return if confirmed with OK."""
+  """Display a confirmation popup. Return if confirmed with OK.
+
+  When the application is running in testing mode (slicer.app.testingEnabled()==True),
```
jcfr, line 2342:
```suggestion
  When the application is running in testing mode (``slicer.app.testingEnabled() == True``),
```

Base/Python/slicer/util.py
```diff
@@ -2306,71 +2306,112 @@ def delayDisplay(message, autoCloseMsec=1000):
...
+    return
   if mw:
     standardButtons = standardButtons if standardButtons else qt.QMessageBox.Ok
     messageBox(text, parent, windowTitle=windowTitle, icon=qt.QMessageBox.Warning, standardButtons=standardButtons, **kwargs)
 
 def errorDisplay(text, windowTitle=None, parent=None, standardButtons=None, **kwargs):
   """Display an error popup.
 
-  If there is no main window then the text is only logged (at error level).
+  If there is no main window, or if the application is running in testing mode (slicer.app.testingEnabled()==True),
```
jcfr, line 2331:
```suggestion
  If there is no main window, or if the application is running in testing mode (``slicer.app.testingEnabled() == True``),
```

Base/Python/slicer/util.py
```diff
@@ -2306,71 +2306,112 @@ def delayDisplay(message, autoCloseMsec=1000):
...
+    return
   if mw:
     standardButtons = standardButtons if standardButtons else qt.QMessageBox.Ok
     messageBox(text, parent, windowTitle=windowTitle, icon=qt.QMessageBox.Information, standardButtons=standardButtons, **kwargs)
 
 def warningDisplay(text, windowTitle=None, parent=None, standardButtons=None, **kwargs):
   """Display popup with a warning message.
 
-  If there is no main window then the text is only logged (at warning level).
+  If there is no main window, or if the application is running in testing mode (slicer.app.testingEnabled()==True),
```
jcfr, line 2320:
```suggestion
  If there is no main window, or if the application is running in testing mode (``slicer.app.testingEnabled() == True``),
```

Base/Python/slicer/util.py
```diff
@@ -2306,71 +2306,112 @@ def delayDisplay(message, autoCloseMsec=1000):
 def infoDisplay(text, windowTitle=None, parent=None, standardButtons=None, **kwargs):
   """Display popup with a info message.
 
-  If there is no main window then the text is only logged (at info level).
+  If there is no main window, or if the application is running in testing mode (slicer.app.testingEnabled()==True),
```
jcfr, line 2309:
```suggestion
  If there is no main window, or if the application is running in testing mode (``slicer.app.testingEnabled() == True``),
```

jcfr:
> @jcfr Can you explain what you mean by
> 
> > * check that skipping effectively work for this class
> 
> Is there a scenario you have in mind that should be checked?

In that particular case, I meant checking that `--testing` can effectively skip the confirmation box that would be shown when instantiating  `MessageDialog`.

The following enables to confirm it is working.

When passing `--testing`, no dialog is show .... 

```bash
./Slicer \
  --testing \
  --no-main-window \
  --disable-modules \
  --python-code "import time; context = slicer.util.MessageDialog('This dialog should not be visible when --testing is passed'); context.__enter__(); slicer.app.processEvents();"
```

Without passing `--testing`, a dialog is show .... 

```bash
./Slicer \
  --no-main-window \
  --disable-modules \
  --python-code "import time; context = slicer.util.MessageDialog('This dialog should not be visible when --testing is passed'); context.__enter__(); slicer.app.processEvents();"
```

These works as expected :heavy_check_mark: 

https://github.com/Slicer/Slicer/blob/f247788d9c625f6e1b53104a4e462eeb33cb1601/Base/Python/slicer/util.py#L2563-L2567

Base/Python/slicer/util.py
```diff
@@ -2509,6 +2537,7 @@ def __exit__(self, type, value, traceback):
 class MessageDialog:
   def __init__(self, message, show=True, logLevel=None):
     """Log the message and show a message box while the code in the context manager is being run.
+    When the application is running in testing mode (slicer.app.testingEnabled()==True), the message box is skipped.
```
jcfr, line 2540:
```suggestion
    When the application is running in testing mode (``slicer.app.testingEnabled() == True``), the message box is skipped.

**Comment by ebrahimebrahim** [comment]
> The following enables to confirm it is working

Thanks for showing how this works; I didn't realize how easy it was to test snippets in the command. I also just noticed [this](https://slicer.readthedocs.io/en/latest/developer_guide/script_repository.html?highlight=--testing#run-python-commands-in-the-slicer-environment) now

File: Base/Python/slicer/util.py [hunk 15]
```diff
@@ -2306,75 +2306,103 @@ def delayDisplay(message, autoCloseMsec=1000):
...
+    testingReturnValue: When the application is in testing mode, this value is returned instead of raising the message box.
+    mainWindowNeeded (bool): If True then the message box will not be raised if there is no mainWindow, but the text is still logged.
+    parent (QWidget): The message box parent; by default it is set to main window by slicer.util.messageBox
+    windowTitle (str): Window title; defaults to a generic title based on log level.
+    kwargs: passed to slicer.util.messageBox
+
+  Returns:
+    The output of slicer.util.messageBox, with two exceptions:
+    - If the application is running in testing mode, then testingReturnValue is returned.
+    - Otherwise, if mainWindowNeeded is True and there is no main window, then None is returned.
```
jcfr, line 2392:
```suggestion
    - Otherwise, if ``mainWindowNeeded`` is True and there is no main window, then None is returned.
```

Base/Python/slicer/util.py
```diff
@@ -2306,75 +2306,103 @@ def delayDisplay(message, autoCloseMsec=1000):
...
+    text (str): Message text
+    testingReturnValue: When the application is in testing mode, this value is returned instead of raising the message box.
+    mainWindowNeeded (bool): If True then the message box will not be raised if there is no mainWindow, but the text is still logged.
+    parent (QWidget): The message box parent; by default it is set to main window by slicer.util.messageBox
+    windowTitle (str): Window title; defaults to a generic title based on log level.
+    kwargs: passed to slicer.util.messageBox
+
+  Returns:
+    The output of slicer.util.messageBox, with two exceptions:
+    - If the application is running in testing mode, then testingReturnValue is returned.
```
jcfr, line 2391:
```suggestion
    - If the application is running in testing mode, then ``testingReturnValue`` is returned.
```

Base/Python/slicer/util.py
```diff
@@ -2306,75 +2306,103 @@ def delayDisplay(message, autoCloseMsec=1000):
...
+  """Displays a messagebox and logs message text; knows what to do in testing mode.
+
+  Args:
+    logLevel: The level at which to log text, e.g. logging.INFO, logging.ERROR
+    text (str): Message text
+    testingReturnValue: When the application is in testing mode, this value is returned instead of raising the message box.
+    mainWindowNeeded (bool): If True then the message box will not be raised if there is no mainWindow, but the text is still logged.
+    parent (QWidget): The message box parent; by default it is set to main window by slicer.util.messageBox
+    windowTitle (str): Window title; defaults to a generic title based on log level.
+    kwargs: passed to slicer.util.messageBox
```
jcfr, line 2387:
```suggestion
  :param logLevel: The level at which to log text, e.g. ``logging.INFO``, ``logging.ERROR``
  :param text: Message text
  :type text: str
  :param testingReturnValue: When the application is in testing mode, this value is returned instead of raising the message box.
  :param mainWindowNeeded: If True then the message box will not be raised if there is no mainWindow, but the text is still logged.
  :type mainWindowNeeded: bool, optional
  :param parent: The message box parent; by default it is set to main window by slicer.util.messageBox
  :type parent: QWidget, optional
  :param windowTitle: Window title; defaults to a generic title based on log level.
  :type windowTitle: str, optional
  :param kwargs: passed to :func:`messageBox`
```

For sake of consistency with existing docstring  style in this file

jcfr (dismissed):

ebrahimebrahim:
> Thanks. Documentation can be reviewed here: https://slicer--6162.org.readthedocs.build/en/6162/developer_guide/slicer.html#slicer.util.confirmOkCancelDisplay

Ah thanks, I was trying to find where those docstrings ended up. I guess the syntax you introduced is from sphinx.

Let me know if I should squash any commits-- I'm leaving it as is otherwise.

jcfr:
:rocket: 

> Let me know if I should squash any commits

Could you squash the last two together?  and add `Co-authored-by: Andras Lasso <lasso@queensu.ca>` :pray:

PR #6172, opened 2022-02-09
ENH: (v4.11 patch) Skip message boxes when testing

ebrahimebrahim:
This is a v4.11 patch mirroring #6162:

> The message box utility functions such as slicer.util.confirmOkCancelDisplay are used in some extensions in such a way that tests can end up timing out and failing; this is due to the expectation of user interaction.

> This commit addresses the issue by having message box functions such as slicer.util.confirmOkCancelDisplay behave differently in a testing situation: they can skip the dialog and return a default result.

I just cherry-picked 95819c1a69bd162cf33adee08304143a2b2d147d which is a more minimal change that the full #6162, then I made a small modification to the docstrings.

[See here](https://github.com/ToothAndClaw/SlicerAuto3dgm/issues/6#issuecomment-1034036421) the discussion about introducing this patch.

Please let me know if it's preferred that the patch only affect `confirmOkCancelDisplay` and not the other `messageBox` functions.

### DIFF
--- a/Base/Python/slicer/util.py
+++ b/Base/Python/slicer/util.py
@@ -2224,7 +2224,8 @@ def delayDisplay(message, autoCloseMsec=1000):
 def infoDisplay(text, windowTitle=None, parent=None, standardButtons=None, **kwargs):
   """Display popup with a info message.
 
-  If there is no main window then the text is only logged (at info level).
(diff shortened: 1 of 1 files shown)


jamesobutler:
@jcfr Should we begin submitting patch fixes/backports to the upcoming branch for the next stable release? Generally there haven't been many patch releases of Slicer and just a new stable minor increment was released instead of a patch release.

PR #6173, opened 2022-02-09
ENH: Add options to toggle individual markups interaction handle axes

### LINKED ISSUE #5558: Add API to Slicer so that a markups node can be moved using the interactions handle only in certain directions

Sometimes is needed to let the user only move a plane with a certain axis of the interactions handle.

Add API to Slicer so that a markups node can be moved using the interactions handle only in certain directions.
Here is one example of what I would like:
`markupsNode = getNode('myMarkupsNodeName')`
`displayNode = markupsNode .GetDisplayNode()`
`displayNode.HandlesInteractiveOn()`
`displayNode.HandlesInteractiveTranslateAlongXAxisOn()`
`displayNode.HandlesInteractiveTranslateAlongYAxisOn()`
`displayNode.HandlesInteractiveTranslateAlongZAxisOff()`
`displayNode.HandlesInteractiveRotateAroundXAxisOff()`
`displayNode.HandlesInteractiveRotateAroundYAxisOff()`
`displayNode.HandlesInteractiveRotateAroundZAxisOn()`

I would like interaction handles show all axes like right now, but only the specified axis (above) to be interactive.

Sunderlandkyl:
Previously, all of the interaction handles could only be enabled/disabled with others of the same type (rotation, translation, scaling).
This commit adds options for enabling disabling the individual XYZ axes for each of the handle types, as well has the handles that allow free transformation.

ROI nodes now display translation handles by default, but only display the free translation point in the center of the ROI.

Closes #5558
Re #5061

### DIFF
--- a/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsDisplayNode.h
+++ b/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsDisplayNode.h
@@ -422,6 +422,7 @@ public:
   /// Get the active color of the markup. This color is used when the mouse pointer hovers over a markup.
   vtkGetVector3Macro(ActiveColor, double);
 
+  //@{
   /// The visibility and interactability of the interaction handles
   vtkGetMacro(HandlesInteractive, bool);
   vtkSetMacro(HandlesInteractive, bool);
@@ -437,11 +438,27 @@ public:
   vtkBooleanMacro(ScaleHandleVisibility, bool);
   void SetHandleVisibility(int handleType, bool visibility);
   bool GetHandleVisibility(int handleType);
+  //@}
 
+  //@{
   /// Get/Set interaction handle size relative to the window size.
   /// Diameter of the interaction handle points is defined as "scale" percentage of diagonal size of the window.
   vtkSetMacro(InteractionHandleScale, double);
   vtkGetMacro(InteractionHandleScale, double);
+  //@}
+
+  //@{
+  /// Get/Set the visibility of the individual handle axes
+  /// The order of the vector is: [X, Y, Z, ViewPlane]
+  /// "ViewPlane" scale/translation allows transformations to take place along the active view plane.
+  /// (ex. center translation point and ROI corner scale handles.
+  vtkSetVector4Macro(RotationHandleComponentVisibility, bool);
+  vtkGetVector4Macro(RotationHandleComponentVisibility, bool);
+  vtkSetVector4Macro(ScaleHandleComponentVisibility, bool);
+  vtkGetVector4Macro(ScaleHandleComponentVisibility, bool);
+  vtkSetVector4Macro(TranslationHandleComponentVisibility, bool);
+  vtkGetVector4Macro(TranslationHandleComponentVisibility, bool);
+  //@}
 
   /// Get data set containing the scalar arrays for this node type.
   /// For markups it is the curve poly data
@@ -453,6 +470,9 @@ public:
   virtual void UpdateAssignedAttribute() override;
 
   virtual void SetScalarVisibility(int visibility) override;
+
+  /// Get flag indicating if the markups node can display scale handles
+  vtkGetMacro(CanDisplayScaleHandles, bool);
 
 protected:
   vtkMRMLMarkupsDisplayNode();
@@ -518,5 +538,11 @@ protected:
   bool RotationHandleVisibility;
   bool ScaleHandleVisibility;
   double InteractionHandleScale;
+
+  bool RotationHandleComponentVisibility[4];
+  bool ScaleHandleComponentVisibility[4];
+  bool TranslationHandleComponentVisibility[4];
+
+  bool CanDisplayScaleHandles;
 };
 #endif
--- a/Modules/Loadable/Markups/Widgets/Resources/UI/qMRMLMarkupsInteractionHandleWidget.ui
+++ b/Modules/Loadable/Markups/Widgets/Resources/UI/qMRMLMarkupsInteractionHandleWidget.ui
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>233</width>
-    <height>66</height>
+    <width>370</width>
+    <height>115</height>
    </rect>
   </property>
   <property name="sizePolicy">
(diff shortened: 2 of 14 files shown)


Modules/Loadable/Markups/MRML/vtkMRMLMarkupsDisplayNode.h
```diff
@@ -443,6 +443,17 @@ class  VTK_SLICER_MARKUPS_MODULE_MRML_EXPORT vtkMRMLMarkupsDisplayNode : public
   vtkSetMacro(InteractionHandleScale, double);
   vtkGetMacro(InteractionHandleScale, double);
 
+  /// Get/Set the visibility of the individual handle axes
```
lassoan, line 446: To associate this documentation block with all the Get/Set functions below, add a group start and group end marker above and below - like this:

https://github.com/Slicer/Slicer/blob/fab4ec043672efc3434ae29fea498979b9ae7b17/Modules/Loadable/Markups/MRML/vtkCurveGenerator.h#L54-L61

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsDisplayNode.h
```diff
@@ -518,5 +529,9 @@ class  VTK_SLICER_MARKUPS_MODULE_MRML_EXPORT vtkMRMLMarkupsDisplayNode : public
   bool RotationHandleVisibility;
   bool ScaleHandleVisibility;
   double InteractionHandleScale;
+
+  bool RotationHandleAxes[4];
```
lassoan, line 533: We need to add "Visible" or "Enabled" to the name.

Also, "handle" and "axis" feels somewhat redundant. Since the free translation handle is not really an axis, probably "handle" would be more appropriate.

So, the name could be something like `RotationHandleVisibility` or `RotationHandleComponentVisibility` (or `...Enabled` instead of `...Visibility`).

Same for scale and translation.

lassoan (commented):
Thanks a lot, this will be great! I've added a few nitpicks inline.

Can you attach a screenshots of the GUI that can modify handle visibility options?

lassoan:
Thank you, this looks quite good. Could you try if you could create a grid layout like this, to reduce the number of labels and make the options a bit more organized:

image

Sunderlandkyl:
Ok, here's the updated widget:
image

Modules/Loadable/Markups/Widgets/qMRMLMarkupsInteractionHandleWidget.cxx
```diff
@@ -104,20 +133,35 @@ void qMRMLMarkupsInteractionHandleWidget::updateWidgetFromMRML()
...
   if (!d->DisplayNode)
     {
     return;
     }
 
-  // Scale handles currently not implemented for representations other than ROI
+  // Scale handles currently not implemented for representations other than ROI and plane
   // Disable by default.
-  if (!vtkMRMLMarkupsROIDisplayNode::SafeDownCast(d->DisplayNode))
+  if (!vtkMRMLMarkupsROIDisplayNode::SafeDownCast(d->DisplayNode) && !vtkMRMLMarkupsPlaneDisplayNode::SafeDownCast(d->DisplayNode))
```
lassoan, line 157: Since markups are now pluggable, we should not hardcode the behavior based on the class but the display node should tell if scaling is enabled or not.

Modules/Loadable/Markups/Widgets/qMRMLMarkupsInteractionHandleWidget.cxx
```diff
@@ -104,20 +133,35 @@ void qMRMLMarkupsInteractionHandleWidget::updateWidgetFromMRML()
...
     {
     // Scale handles not currently implemented for non ROI nodes
     d->scaleVisibilityCheckBox->setVisible(false);
+    d->scaleXCheckBox->setVisible(false);
```
lassoan, line 161: It looks a bit strange that the labels remain there but the checkboxes disappear. Instead of making them disappear could you uncheck and disable the options that are not applicable to the node type? (we could hide the labels as well, but then the layout would jump around when browsing the tree, so it is probably better to keep displaying all rows)

Modules/Loadable/Markups/Widgets/Resources/UI/qMRMLMarkupsInteractionHandleWidget.ui
```diff
@@ -32,71 +32,303 @@
...
+      <widget class="QPushButton" name="moreOptionsCheckBox">
+       <property name="sizePolicy">
+        <sizepolicy hsizetype="Fixed" vsizetype="Fixed">
+         <horstretch>0</horstretch>
+         <verstretch>0</verstretch>
+        </sizepolicy>
+       </property>
        <property name="text">
-        <string>Scale</string>
+        <string>More options...</string>
```
lassoan, line 166: Nitpick: The widget is very crowded on the left side and there is a lot of space on the right. Could you try moving the "More options" button to the top-right corner? If it does not look good or not feasible for some reason then you could also try a tall double-arrow button (similarly to "Module paths" in application settings / modules) to show/hide that section.

lassoan (commented):
Thank you, it is getting really close. I've added a few small comments inline.

In addition to those there are issues with the visibility updates of the additional checkboxes:
- "More options" always makes every additional checkboxes visible (so scale can appear for curves).
- When choosing a different markup in the tree the checkboxes are only hidden, never get displayed (so scale does not appear for ROIs and planes if previously a different markup type was selected).

Sunderlandkyl:
Ok, here's the updated widget:
image

lassoan:
Thanks a lot, this looks good. Just one nitpick (sorry): the small vertical line in the first row was just an artifact when I photoshopped the sketch (included a bit of the checkbox side when copied the region from another part of the image). It does not look bad (it even helps a bit separating the additional options from the basic options), but we don't use this kind of separators elsewhere, so it is probably better not to introduce it for just this. So, I think we could just remove that small vertical line for consistency.

Sunderlandkyl:
Ok, removed.

lassoan (approved):
Thanks a lot. This looks perfect.

PR #6178, opened 2022-02-11
BUG: Fix Qt designer crash

RafaelPalomar:


olevs:
This fixes the issue on Ubuntu 20.04 as well. Now it works running ./Slicer --designer.
Trying to launch Designer from the button inside Slicer or running the SlicerDesigner app still don't work.

lassoan (approved):
Perfect, thank you!

Issue #6168, opened 2022-02-08
vtkSegment default terminology not reading Settings


fbordignon:
vtkSegment sets a hardcoded tag in its constructor that should be read from QSettings:

https://github.com/Slicer/Slicer/blob/7610ddea1b40b569641c77964976718be045dde7/Libs/vtkSegmentationCore/vtkSegment.cxx#L69

I am not sure of the best way to accomplish this. I would try to add a class member to hold the terminology setting and update it when settings are updated or first read.

lassoan:
The default terminology term for new segments is already set from application settings. This initial value gets used only if you don't use the GUI.

fbordignon:
Hey, Andras. Thank you for your response. Ok, so maybe there is a case not treated. I reached vtkSegment starting from
[`qSlicerSubjectHierarchySegmentationsPlugin::convertLabelmapToSegmentation()`](https://github.com/Slicer/Slicer/blob/7610ddea1b40b569641c77964976718be045dde7/Modules/Loadable/Segmentations/SubjectHierarchyPlugins/qSlicerSubjectHierarchySegmentationsPlugin.cxx#L1098)

fbordignon:
Thanks for the tips. I was looking into it and some other conversion functions such as https://github.com/Slicer/Slicer/blob/7610ddea1b40b569641c77964976718be045dde7/Modules/Loadable/Segmentations/SubjectHierarchyPlugins/qSlicerSubjectHierarchySegmentationsPlugin.cxx#L1134
potentially have the same problem. So I will need to implement helper functions to apply the terminology.

I was trying to use https://github.com/Slicer/Slicer/blob/fab4ec043672efc3434ae29fea498979b9ae7b17/Modules/Loadable/Segmentations/Logic/vtkSlicerSegmentationsModuleLogic.cxx#L1641
But it is not static, I will need to get the `vtkSlicerSegmentationsModuleLogic` instance. How can I do that from a `SubjectHierarchyPlugin`?

I was trying to get the `qMRMLSegmentEditorWidget` instance as well, to use the convenience function `defaultTerminologyEntry()` but it does not recognize importing the `qMRMLSegmentEditorWidget.h` How can I get that instance?

cpinter:
Are you writing your code in Python or C++? From what kind of class do you want to access these objects?

cpinter:
This is an example for getting the Volumes logic from C++. Just replace the class name and string and it should work:
```
[...]

PR #6179, opened 2022-02-11
ENH: Fix convert and reparent segmentation from labelmap terminology

fbordignon:
Working on #6168

### DIFF
--- a/Modules/Loadable/Segmentations/SubjectHierarchyPlugins/qSlicerSubjectHierarchySegmentationsPlugin.cxx
+++ b/Modules/Loadable/Segmentations/SubjectHierarchyPlugins/qSlicerSubjectHierarchySegmentationsPlugin.cxx
@@ -28,6 +28,12 @@
 #include "vtkSlicerSegmentationsModuleLogic.h"
 #include "qSlicerSubjectHierarchyPluginHandler.h"
 #include "qSlicerSubjectHierarchyDefaultPlugin.h"
+#include "qSlicerCoreApplication.h"
+#include "qSlicerModuleManager.h"
+#include "qSlicerAbstractCoreModule.h"
+#include "qSlicerAbstractModuleRepresentation.h"
+#include "qMRMLSegmentEditorWidget.h"
+#include "qSlicerScriptedLoadableModuleWidget.h"
 
 // Qt includes
 #include <QDebug>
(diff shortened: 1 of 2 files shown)


Modules/Loadable/Segmentations/SubjectHierarchyPlugins/qSlicerSubjectHierarchySegmentationsPlugin.cxx
```diff
@@ -262,12 +268,24 @@ bool qSlicerSubjectHierarchySegmentationsPlugin::reparentItemInsideSubjectHierar
     return false;
     }
 
+  vtkSlicerSegmentationsModuleLogic* sgmentationsModuleLogic =
```
lassoan, line 271: It is simpler to get module logics via the application logic class - see here: https://github.com/Slicer/Slicer/blob/8694fc789671ef66644d10fff66d9c37826c40bd/Modules/Loadable/Markups/SubjectHierarchyPlugins/qSlicerSubjectHierarchyMarkupsPlugin.cxx#L291-L303

Modules/Loadable/Segmentations/SubjectHierarchyPlugins/qSlicerSubjectHierarchySegmentationsPlugin.cxx
```diff
@@ -262,12 +268,24 @@ bool qSlicerSubjectHierarchySegmentationsPlugin::reparentItemInsideSubjectHierar
...
+    (qSlicerCoreApplication::application() ? vtkSlicerSegmentationsModuleLogic::SafeDownCast(
+    qSlicerCoreApplication::application()->moduleManager()->module("Segmentations")->logic()) : nullptr);
+
+  qSlicerScriptedLoadableModuleWidget* segmentEditorWidget = dynamic_cast<qSlicerScriptedLoadableModuleWidget*>(qSlicerCoreApplication::application() ?
```
lassoan, line 275: This would be very fragile: Segment Editor widget may not have been even instantiated. There could be also multiple segment editor widgets in the application in many modules. Also, how the user would know that the current default terminology entry in one of the Segment Editor widgets in one of the modules controls the terminology used in this conversion?

Instead, we could get the default from the application settings (`Segmentations/DefaultTerminologyEntry` is a global application setting which can be specified in Segmentations section of the application settings). It would be also much easier to access this registry key than a module widget.

lassoan:
I'll convert this to draft to make it more clear that it is not ready for re-review yet.

PR #6184, opened 2022-02-13
ENH: Improve Export 

### LINKED ISSUE #6180: Improve Export to file for multiple nodes
- Enable "Expor to file..." feature for clicking on the scene: it is just a folder the same way as all other folders
- Add an option to recreate folder structure: it would be an amazingly effective way to save a complex hierarchy of nodes
- Allow specifying non-existing subfolder(s) as Directory: it makes it much simpler to save into a new subfolder. If we already have a temporary folder but want to specify a subfolder for the new saved data then currently we need to: 1. click the `...` button, 2. click `New folder`, 3. type the folder name, then 4. click `Select` - we could save all these clicks by auto-creating the specified folder (and all parents). Some file writers create parent folders automatically, so the current behavior is also inconsistent (depending on the nodes, they can or cannot be written to a not-yet-existing folder).

ebrahimebrahim:
This closes #6180 

# Enable export action when clicking on scene
image

# Create non-existing subfolders during export
image

# Add "Preserve hierarchy" option to export
image

I could use some help with testing this feature on windows. I don't have a windows machine, and I don't know how platform-robust all my directory string manipulation is. 

# Minor bug fixes
- dialog was getting broken in certain nested folder situations if the user unchecked and checked certain boxes
- "apply transform" option was not getting displayed in some rare-ish situations

### DIFF
--- a/Base/QTGUI/qSlicerExportNodeDialog.cxx
+++ b/Base/QTGUI/qSlicerExportNodeDialog.cxx
@@ -43,80 +43,6 @@
 
 
 //-----------------------------------------------------------------------------
-static QString forceFileNameExtension(const QString& fileName, const QString& extension, vtkMRMLNode* node)
-{
-  qSlicerCoreIOManager* coreIOManager = qSlicerCoreApplication::application()->coreIOManager();
-  QString strippedFileName = qSlicerCoreIOManager::forceFileNameValidCharacters(fileName);
-  strippedFileName = coreIOManager->stripKnownExtension(strippedFileName, node) + extension;
-  return strippedFileName;
-}
-
-//-----------------------------------------------------------------------------
-static QString defaultFilename(vtkMRMLNode* node, QString extension)
-{
-  const QString unsafeNodeName(node->GetName() ? node->GetName() : "");
-  qSlicerCoreIOManager* coreIOManager = qSlicerCoreApplication::application()->coreIOManager();
-  if (!coreIOManager)
-    {
-    qCritical() << Q_FUNC_INFO << "failed: Core IO manager not found.";
-    return QString();
-    }
-  const QString safeNodeName = qSlicerCoreIOManager::forceFileNameValidCharacters(unsafeNodeName);
-  return forceFileNameExtension(safeNodeName, extension, node);
-}
-
-//-----------------------------------------------------------------------------
-// Return whether list a has any elements that list b doesn't have.
-template <typename T>
-static bool setDifferenceIsNonempty(const QList<T>& a, const QList<T>& b)
-{
-  for (const T & item : a)
-    {
-    if (!b.contains(item))
-      {
-      return true;
-      }
-    }
-  return false;
-}
-
-//-----------------------------------------------------------------------------
-// Return whether all widgets that are part of the given layout are invisible
-// Visibility is considered relative to the given widget relativeTo
-static bool layoutWidgetsAllInvisible(const QLayout* layout, const QWidget* relativeTo)
-{
-  for (int i = 0; i < layout->count(); ++i)
-    {
-    QWidget* widget = layout->itemAt(i)->widget();
-    if (widget && widget->isVisibleTo(relativeTo))
-      {
-      return false;
-      }
-    }
-  return true;
-}
-
-//-----------------------------------------------------------------------------
-// Set up the given ctkPathLineEdit to make it easier to use in a tabbing order.
-// ctkPathLineEdit has a default focus policy of NoFocus. It will usable for tab ordering
-// if it has TabFocus and sets one of its member widgets as the focus proxy.
-// Really this should probably be a fix in the implementation of ctkPathLineEdit, but this is fine for now.
-static void fixCtkPathLineEditTabbing(ctkPathLineEdit* w)
-{
-  // Unfortunately, ctkPathLineEdit does not give direct access to its underlying widgets,
-  // so we have to do this fiddly thing with QObject::findChildren.
-  auto lineEdits = w->findChildren<QLineEdit*>();
-  if (lineEdits.size()!=1)
-    { // This might happen if ctkPathLineWidget ever changes its internals.
-    qWarning() << "Export dialog warning: unable to set up ctkPathLineEdit keyboard focus rules. Tabbing might be weird.";
-    return;
-    }
-
-  w->setFocusPolicy(Qt::TabFocus);
-  w->setFocusProxy(lineEdits[0]);
-}
-
-//-----------------------------------------------------------------------------
 NodeTypeWidgetSet::NodeTypeWidgetSet(QWidget* parent, vtkMRMLStorableNode* storableNode, vtkMRMLScene* scene)
   : prototypeNode{storableNode}
 {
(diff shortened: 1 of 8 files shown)


Base/QTGUI/qSlicerExportNodeDialog.cxx
```diff
@@ -465,14 +467,16 @@ const QList<vtkMRMLStorableNode*> & qSlicerExportNodeDialogPrivate::nodeList() c
...
+  const QList<vtkMRMLStorableNode*>& nodesNonrecursive,
+  const QList<vtkMRMLStorableNode*>& nodesRecursive,
+  vtkMRMLStorableNode* selectedNode,
+  const QHash<QString,QVariant>& nodeIdToSubjectHierarchyPath
 )
 {
   this->MRMLScene = scene;
   this->nodesRecursive = nodesRecursive;
   this->nodesNonrecursive = nodesNonrecursive;
+  this->nodeIdToSubjectHierarchyPath = nodeIdToSubjectHierarchyPath;
```
lassoan, line 479: I've just noticed that member variables start with lowercase. The coding convention in Slicer is to start member variable names with uppercase. It is not essential, but it would be nice if you could update all the member variable names in this class accordingly.

Base/QTGUI/qSlicerExportNodeDialog.cxx
```diff
@@ -793,7 +841,8 @@ void qSlicerExportNodeDialogPrivate::adjustTabbingOrder()
       }
     }
   setTabOrder(this->IncludeChildrenCheckBox,this->RecursiveChildrenCheckBox);
-  setTabOrder(this->RecursiveChildrenCheckBox, this->HardenTransformCheckBox);
+  setTabOrder(this->RecursiveChildrenCheckBox, this->PreserveHierarchyCheckBox);
```
lassoan, line 844: You can set the tab order in Qt Designer. Unless there is a specific reason to do it in code, it would be better to remove these lines and set the tab order there.

Modules/Loadable/SubjectHierarchy/Widgets/qSlicerSubjectHierarchyExportPlugin.cxx
```diff
@@ -43,6 +43,31 @@
 // MRML includes
 #include <vtkMRMLStorableNode.h>
 
+//-----------------------------------------------------------------------------
+// Utility function for getting the subject-hierarchy-based relative path of an item.
+// Starting from fromItemID, this function climbs up the subject hierarchy tree until it reaches toItemID,
+// and it constructs a list of item names that it encounters along the way. The list goes into the return parameter `path`.
+// The list can be thought of as the interval "(fromItem, toItem]"; i.e. it excludes fromItem.
+// The returned bool indicates success.
+static bool getSubjectHierarchyPath(vtkMRMLSubjectHierarchyNode* shNode, vtkIdType fromItemID, vtkIdType toItemID, QStringList& path)
```
lassoan, line 52: Instead of adding it to the global namespace (even with with `static`), it would be more clean to add it as a static method to `qSlicerSubjectHierarchyExportPluginPrivate`.

Modules/Loadable/SubjectHierarchy/Widgets/qSlicerSubjectHierarchyExportPlugin.cxx
```diff
@@ -43,6 +43,31 @@
...
+{
+  path.clear();
+  vtkIdType endItemID = shNode->GetItemParent(toItemID); // One item above the highest we want to go
+  int iteration = 0;
+  for (vtkIdType itemID = shNode->GetItemParent(fromItemID); itemID != endItemID; itemID = shNode->GetItemParent(itemID))
+    {
+    if (itemID==0 || iteration > 9999)
```
lassoan, line 59: It would be much more clear to do an infinite loop detection more explicitly. Since the IDs are numbers and we expect only a couple of them (hierarchy is not expected to be deeper) they can be compared very quickly, so it would be a negligible overhead to put the IDs in an array (not a list or map, because for small structures array is expected to have lower overhead) and before each insertion check if the item is already in it.

If you are concerned about the performance then you can do what you do now with a limit of 100 (hierarchy should be really never go above that) and if that number is exceeded then start again and store the IDs and detect the loop and if you find that indeed there is a loop then report the error (with details, such as the items in the loop).

lassoan (commented):
Thank you, I just added a few nitpicky comments, this will be great!

Base/QTGUI/qSlicerExportNodeDialog.cxx
ebrahimebrahim: It's easier to do it in code in this case, because the tab order gets messed up as widgets get added and removed. I found it to be pretty confusing to try to set the tab order for some things in Qt Designer and for other things in code, hence the approach I took here.

jamesobutler: I personally set the tab order by just having the row index values ordered from starting with lowest index at the top of the ui file. I find this to be most orderly without having to add additional lines setting tab order explicitly.

ebrahimebrahim: In this case rows get added and removed as the checkboxes are manipulated, so I don't think a ui-file-based approach would work.

ebrahimebrahim:
@lassoan Thank you for all the feedback! The only item I think remains to address on this is the implicit/explicit infinite loop thing, where I left some questions.

Modules/Loadable/SubjectHierarchy/Widgets/qSlicerSubjectHierarchyExportPlugin.cxx
lassoan: Explicitly: check each item ID if it is in the parent list already; if it is then we know that there is an infinite loop.
Implicit: set a limit for the maximum number of parent items (e.g., total number of items + 1), if that number is reached then we know that there must have been an infinite loop.

> I'm not putting the IDs in a container,

Not in the current implementation, but it would be very easy to change this. Since we never expect to get into an infinite loop, it makes sense to build both the string list (this will be returned) and ID list (for quick loop check) at the same time.

Base/QTGUI/qSlicerExportNodeDialog.cxx
```diff
@@ -733,6 +780,7 @@ bool qSlicerExportNodeDialogPrivate::populateNodeTypeWidgetSets()
     }
   else
     {
+    this->setWindowTitle(QString("Export 0 Nodes")); // This hopefully never happens, but we want a correct title in case it does.
     qCritical() << Q_FUNC_INFO << "failed: There is nothing to export.";
```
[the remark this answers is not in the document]
lassoan: If you are not sure if warning/error message should be used then a good test is to check the return code (if there is no return code then you make the decision based on what return code you would use):
- if you return with success (you can complete the request, but you suspect that something may be wrong or you encountered an error but you could recover from it) then it is more appropriate to use a warning log; it does not matter that you did not expect this request - if you can perform what was asked then it is not really an error
- if you return with failure (you cannot complete what the method promises to do) then an error message is more appropriate

Qt distinguishes a special error message type - qFatal - that should be logged if the application has to be terminated because of that error. So, the generic error message is qCritical. Since qCritical sounds very serious, sometimes developers use qWarning even for irrecoverable error occurs. Since there is no complete consistency in the Slicer code base; and log type here does not change the behavior of the application in any way (there are only a very few places where it actually changes the behavior), I'm OK with either choice.

Base/QTGUI/qSlicerExportNodeDialog.cxx
lassoan: OK, this is a special case that may make it easier to set tab order for all widgets in code. If you describe this in a comment then it is good enough for me.

Modules/Loadable/SubjectHierarchy/Widgets/qSlicerSubjectHierarchyExportPlugin.cxx
ebrahimebrahim: > Explicitly: check each item ID if it is in the parent list already; if it is then we know that there is an infinite loop.

Ah I see what you mean, but this approach misses some types of infinite loops: an infinite loop where the item ID returned is always different. Okay that's technically not possible because there are only finitely many possible vtkIDType values, but the number could be very large depending on the size of a vtkIDType, so we could still end up with an effectively stalled program.

When protecting against stalling the program, I don't want to make assumptions about the size of vtkIDType and I don't want to make assumptions about whether item ids will always be delivered correctly, i.e. from a small pool of possible ids.

> and ID list (for quick loop check) 

The ID list seems like a lot more overhead for loop check compared to just an integer counter. I'm not sure I'm seeing the advantage of keeping an ID list just for loop check. With my current method, the loop check is an integer comparison, versus iterating through an ID array and doing a bunch of equality checks.

(Sorry if I am still misunderstanding your suggestion!)

lassoan: Since most cases you will have about 2-3 values in that array, it will be plenty fast. Maybe in rare cases it will be as many as 5 values. Iterating through 5 integer values in an array is nothing compared to string manipulations and file IO.

In the extremely rare case of an infinite loop, this int array based check will be much faster, because you will detect the error after a few hundred iterations, so you will save 9000 string manipulations. The upper bound on the number of IDs is not the maximum possible values that the ID can represent but the number of items in the subject hierarchy, which in most cases is much smaller than 9999.

Another big issue with the 9999 is that who guarantees that it will be enough? Sure, usually it will be enough, but do you want to really add to the user documentation that maximum 9999 nodes can be exported using this feature?

The only advantage of the 9999 based solution is that it is a little bit simpler, but considering that it is less robust, less elegant and the performance improvement is either negligible or it actually makes things slower, I think we should not use this magic number based design. I'm open to other suggestions, but we can definitely do better than the current design.

lassoan:
> * One simple solution would be for `ctkPathLineEdit` to always set its `QLineEdit` as its focus proxy. I'm assuming most people who use `ctkPathLineEdit` are not interested in being able to set keyboard focus to the "..." button

This sounds good to me. It would mean that things would work as developers would normally expect, without requiring any extra effort or thinking from the developer. Exposing the child widget would still require developers to realize that it is a composite widget and so they need to take extra steps.

Base/QTGUI/qSlicerExportNodeDialog.cxx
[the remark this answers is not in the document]
ebrahimebrahim: Thank you for these very helpful guidelines!

Then I think "error" is suitable in this case, so I will leave this part unchanged.

ebrahimebrahim:
@lassoan Thank you for all the feedback-- all items should be addressed now and this is ready for another look.

Regarding the CTK-level fix, I made [this PR](https://github.com/commontk/CTK/pull/1026) which [09b8f1b](https://github.com/Slicer/Slicer/pull/6184/commits/09b8f1b22f40ac5a0192e40ca5cf93676b9382d8) depends on in order to work. I don't know how often CTK gets upgraded within Slicer, or whether I should wait for an upgrade to introduce this commit. Maybe it's fine to introduce it now because the only thing that will be broken until CTK is upgraded is tabbing.

lassoan:
We'll get the improvement merged into CTK and update Slicer's CTK today, unless raises any concerns.

lassoan:
This is now merged, thank you for your contribution!

PR #6185, opened 2022-02-13
Organize menu bar with a few new options and refresh Welcome module to match

jamesobutler:
I decided to give the Menu Bar some attention to more cleanly organize actions and add a few new ones.

This work was originally out of #5966, but this PR has no changes in bringing in new icon artwork. Changing of icon artwork for modernization, light/dark mode, etc can require a lot of coordination between others to settle on a final design. This includes past talks of bringing in a Kitware designer to overhaul icons across the application. So for this PR I am only including dev work that I have control of to help the PR make it to a final state that is ultimately merged.

## File Menu

| Current | This PR |
|----------|---------|
|image|image|

## Edit Menu

The Edit->Application Settings action is now using the same icon as "Customize Slicer" in the Welcome module.

| Current | This PR |
|----------|---------|
|image|image|

## View Menu

- View->Module Finder has been newly added. This is because "Ctrl+F" to bring up the Module Finder would only work if the module selector toolbar is visible and the user has the option to hide that toolbar. So now it is an application level action that can be triggered with "Ctrl+F" even with module selector toolbar hidden.

| Current | This PR | This PR |
|----------|---------|---------|
|image|image|image|

## Help Menu

- Help->[Visual Blog](https://www.slicer.org/w/index.php/Slicer4:VisualBlog) has been removed because this Slicer Wiki resource hasn't been updated since 2014/2016! Visual materials related to announcements to the community have already been shared through the Discourse forum or the website.
- I have added more Help menu entries. Most link to the web in some manner which is just a side-effect of Slicer not distributing local/offline documentation materials with the application. 

| Current | This PR |
|----------|---------|
|image|image|

## Welcome Module

I've pushed an additional commit here to make the actions in the Welcome module use the same "Add" terminology as the File Menu and also use the same icons. "Add Data" button is now on the left and "Add DICOM Data" on the right to match the order in the Load/Save ToolBar which has this left/right order and to match the top/down order in the File menu. "Add Data" is situated to be more important than "Add DICOM Data".

| Current | This PR |
|----------|---------|
|image|image|


pieper (dismissed):
Definite improvement, thanks.

jamesobutler:
Should File->Recently Loaded be changed to File->Recently Added to match the “Add” terminology instead of “Load”?

jamesobutler:
We can keep the terminology as it is currently in the PR. It can change when we go towards the following comment originally from https://github.com/Slicer/Slicer/pull/5966#issuecomment-953077396.

> Why changing "Add ... data" to "Load ... data"? "Open" is used universally everywhere. "Import" is used sometimes to open some additional file formats or to indicate that the data is added to current content (instead of replacing it). What do you think about: Open file..., Open folder..., Import DICOM... options? Open file/folder would open a file/folder selector directly (saving one click) and then then show the selected content in "Add data" dialog so that the user could adjust loading options.

If no other feedback, then will proceed with merging tomorrow.

jcfr:
> Should File->Recently Loaded be changed to File->Recently Added to match the “Add” terminology instead of “Load”?

That makes sense :+1: 

Let's also make sure to change these references:

```
Base/QTApp/Resources/UI/qSlicerMainWindow.ui
221:      <string>Recently Loaded</string>

Base/QTGUI/qSlicerSettingsGeneralPanel.cxx
166:                      "Max. number of 'Recently Loaded' menu items",

Base/QTGUI/Resources/UI/qSlicerSettingsGeneralPanel.ui
137:      <string>Max. number of 'Recently Loaded' menu items:</string>
```

cpinter:
Thanks @jamesobutler !

I like the direction, and am OK with all the changes but one.

Is there a good reason for changing the Python interactor shortcut from Ctrl+3 to Ctrl+\` ? I use this shortcut a lot, and also it is propagated manually to basically all the apps I have created (like widgets not children of the main window) so this would cause confusion. Another reason to keep it is that the \` key is not available on many non-US layouts (actually none of the other layouts I know).

jamesobutler:
Sure I can remove that shortcut change. It was motivated by using a typical shortcut for opening a integrated terminal as it is the one for popular Visual Studio Code IDE. @cpinter Do you know what VSCode sets this shortcut to on non-US layouts? I may just have to set it to some key where Qt does a localization change for the Key.

jamesobutler:
I’ll go ahead and remove the shortcut change for the python console. Interesting thing I haven’t considered which is internationalization of keyboard shortcuts in addition to regular text. Therefore being able to customize keyboard shortcuts based on the keys available in the local keyboard layout will be necessary.

lassoan:
> Should File->Recently Loaded be changed to File->Recently Added to match the “Add” terminology instead of “Load”?

I would really like to change the name and behavior of this menu be "Recently used files" (this is the official term for this in the operating system and most other software). Because it can be very hard to find a saved scene after restarting Slicer. Storing the saved scene file name would fix this issue. We might consider adding output filenames to the list in "Export to file".

We could avoid two renames of this menu item if we kept the current name then switching to "used" when the behavior change is implemented.

jamesobutler:
I've looked in my OS in various places and I have not seen the terminology of "recently used". I see variations of simply "Recent {object name}" or "Recent". So specifically the "used" term that I don't see.

image

jamesobutler:
## File Menu

| Current | This PR |
|----------|---------|
|image|image|

lassoan:
> I've looked in my OS in various places and I have not seen the terminology of "recently used"

MRU (most recently used) list is the name of the feature in the Windows API, registry, etc. On the user interface things may keep changing, so it makes sense to check what is used in Windows and commonly used applications.

jamesobutler:
Yes, I was checking on latest Windows and common applications and the terminology was something like “Recent” or “Recent Files”.

If you don’t like “Recently Added”, what about just “Recent”?

lassoan:
Give me a few hours to see if I can add the saved scene file name there. If I succeed then renaming to "Recent" would be perfect.

jamesobutler:
That's a good idea @cpinter!. I've added the "Ctrl+`" as a secondary shortcut for the same action. I've maintained "Ctrl+3" as the primary one for compatibility with non-US based keyboards. In the View Menu the action is shown with the primary, though you can see in the Keyboard Shortcut Reference dialog that there are multiple shortcuts enabled as shown below.
image

Base/QTApp/qSlicerMainWindow.cxx
```diff
@@ -461,7 +461,7 @@ void qSlicerMainWindowPrivate::setupUi(QMainWindow * mainWindow)
     this->PythonConsoleToggleViewAction->setText(qSlicerMainWindow::tr("&Python Interactor"));
     this->PythonConsoleToggleViewAction->setToolTip(qSlicerMainWindow::tr(
       "Show Python Interactor window for controlling the application's data, user interface, and internals"));
-    this->PythonConsoleToggleViewAction->setShortcut(QKeySequence("Ctrl+3"));
+    this->PythonConsoleToggleViewAction->setShortcuts({qSlicerMainWindow::tr("Ctrl+3"), qSlicerMainWindow::tr("Ctrl+`")});
```
lassoan, line 464: Also add this to the user manual (user_interface.md).

jamesobutler: Done. The user interface page now shows:
image

lassoan:
@jamesobutler I've pushed a commit that adds saved scene file names to the recent files list, so we can go on with the rename from "Recently Loaded" to "Recent".

I did not want to rebase and force-push to make sure that no changes get overwritten, but it might have generated some extra commits. These extra commits should disappear if you rebase on latest master.

jamesobutler:
Appears the GitHub action to build Slicer is still using an older version that is doing "Setting C++ standard - C++11" even though https://github.com/Slicer/Slicer/commit/bc2b3f4049ef32486cf43ad16903820fc8767f27 was included in last night's Slicer Preview build.

lassoan:
> Appears the GitHub action to build Slicer is still using an older version that is doing "Setting C++ standard - C++11" even though [bc2b3f4](https://github.com/Slicer/Slicer/commit/bc2b3f4049ef32486cf43ad16903820fc8767f27) was included in last night's Slicer Preview build.

The docker image was refreshed by slicerbot yesterday morning - see here: https://github.com/Slicer/SlicerDocker/commits/master, and the Cxx14 requirement was committed after that.

@jcfr is the image update fully automatic? Is it normal that it was refreshed yesterday morning but not this morning?

jamesobutler:
## File Menu

| Current | This PR |
|----------|---------|
|image|image|

jcfr:
> is the docker image update fully automatic?

Yes, there is a cron job setup using this project. See https://github.com/Slicer/SlicerDockerUpdate#readme

> Is it normal that it was refreshed yesterday morning but not this morning?

No, the build was failing with `g++: internal compiler error: Killed (program cc1plus)`. This should be fixed by https://github.com/Slicer/SlicerDocker/pull/49

lassoan:
> This should be fixed by [Slicer/SlicerDocker#49](https://github.com/Slicer/SlicerDocker/pull/49)

Great, thanks for the fix!

jamesobutler:
@lassoan I've pushed an update to remove `QFileInfo` based on the build error that was alerted from changes in `ENH: Add saved scene filenames to File menu's Recent list`
```
/usr/src/Slicer/Base/QTApp/qSlicerMainWindow.cxx: In member function ‘virtual void qSlicerMainWindow::onFileSaved(const IOProperties&)’:
/usr/src/Slicer/Base/QTApp/qSlicerMainWindow.cxx:1367:3: error: declaration does not declare anything [-fpermissive]
   QFileInfo ;
   ^~~~~~~~~
```

Base/QTApp/qSlicerMainWindow.cxx
```diff
@@ -1299,35 +1312,74 @@ void qSlicerMainWindow::onWarningsOrErrorsOccurred(ctkErrorLogLevel::LogLevel lo
...
+    }
+  QString fileExtension = fileName.toUpper();
+  // Adding every saved file to the recent files list could quickly overwrite the entire list,
+  // therefore we only add the scene file.
+  if (fileName.endsWith(".mrml", Qt::CaseInsensitive)
+    || fileName.endsWith(".mrb", Qt::CaseInsensitive))
+    {
+    // Scene file properties do not contain fileType and it contains screenshot,
+    // which can cause complication when attempted to be stored,
+    // therefore we create a new clean property set.
```
jcfr, line 1375:
```suggestion
    // Since the Scene file properties is not associated with the `fileType` key
    // and can be associated with the `screenshot` key (which can cause complication 
    // when attempted to be stored), we create a new clean property set.

File: Base/QTCore/qSlicerCoreIOManager.cxx [hunk 3]
```diff
@@ -776,6 +782,7 @@ bool qSlicerCoreIOManager::saveNodes(qSlicerIO::IOFileType fileType,
       continue;
       }
     nodes << writer->writtenNodes();
+    emit fileSaved(parameters);
```
jcfr, line 785: similarly to what is done here:

https://github.com/Slicer/Slicer/blob/e5c5ad7801b81c99a0e37b44ce570e9acbfdb7f0/Base/QTCore/qSlicerCoreIOManager.cxx#L611

I suggest to introduce `savedFileParameters` and explicitly set the file type.

Base/QTApp/qSlicerMainWindow.cxx
jcfr: By addressing my later comment, we should be able to further simplify and avoid setting `fileType` here.

Base/QTCore/qSlicerCoreIOManager.cxx
lassoan: There are several other parameters that are used for the file loading (coordinate system, centered, etc). They all need to be saved. Scene saving is a special case, we know that we don't lose information by sanitizing the properties, but in general we need to pass all parameters. So, I'm not exactly sure what you mean to change.

Base/QTApp/qSlicerMainWindow.cxx
```diff
@@ -1299,35 +1312,74 @@ void qSlicerMainWindow::onWarningsOrErrorsOccurred(ctkErrorLogLevel::LogLevel lo
...
+//---------------------------------------------------------------------------
+void qSlicerMainWindow::onFileSaved(const qSlicerIO::IOProperties& fileProperties)
+{
+  Q_D(qSlicerMainWindow);
+  QString fileName = fileProperties["fileName"].toString();
+  if (fileName.isEmpty())
+    {
+    return;
+    }
+  QString fileExtension = fileName.toUpper();
```
lassoan, line 1367:
```suggestion
```

jamesobutler:
Anything I can do to help finalize this work?

lassoan:
I did not understand what changes @jcfr was suggesting. @jcfr if you have any specific concerns then it would be great if you could describe them in more detail. Otherwise I'll just go and merge this tonight. We can make changes later anyway.

lassoan:
Thank you, this is now merged.

PR #6187, opened 2022-02-14
BUG: Fix Markups toolbar not using display names

jamesobutler:
The tooltip "Create new ClosedCurve" is now "Create new Closed Curve".

| Current | This PR |
|----------|---------|
|image|image|


lassoan (approved):
Good catch, thank you!

PR #6188, opened 2022-02-14
ENH: Add connector line for fiducial markup label texts

rbumm:
Label text offsets and a connector line (as in anatomy books)  would be awesome to have when labeling anatomical structures. 

 This is an implementation of label text connector lines for fiducial markups. 

Please see the comments for questions. Thank you.

### DIFF
--- a/CMake/SlicerBlockInstallQt.cmake
+++ b/CMake/SlicerBlockInstallQt.cmake
@@ -191,3 +191,28 @@ set(QT_INSTALL_LIB_DIR ${Slicer_INSTALL_LIB_DIR})
       FILES "${Slicer_INSTALL_ROOT}/bin/designer-real"
       COMPONENT Runtime)
   endif()
+
+  # Bundle Qt language tools with the application if internationalization is enabled.
+  # These tools allow Slicer modules to update and process Qt translation (.ts) files
+  # without requiring installation of Qt.
+  if(Slicer_BUILD_I18N_SUPPORT)
+    foreach(tool lconvert lrelease lupdate)
+      install(PROGRAMS ${qt_root_dir}/bin/${tool}${CMAKE_EXECUTABLE_SUFFIX}
+        DESTINATION ${Slicer_INSTALL_ROOT}/bin COMPONENT Runtime
+        )
+      slicerStripInstalledLibrary(
+        FILES "${Slicer_INSTALL_ROOT}/bin/${tool}"
+        COMPONENT Runtime
+        )
+      if(APPLE)
+        set(dollar "$")
+        install(CODE
+          "set(app ${Slicer_INSTALL_BIN_DIR}/${tool})
+          set(appfilepath \"${dollar}ENV{DESTDIR}${dollar}{CMAKE_INSTALL_PREFIX}/${dollar}{app}\")
+          message(\"CPack: - Adding rpath to ${dollar}{app}\")
+          execute_process(COMMAND install_name_tool -add_rpath @loader_path/..  ${dollar}{appfilepath})"
+          COMPONENT Runtime
+          )
+      endif()
+    endforeach()
+  endif()
--- a/CMake/SlicerMacroBuildApplication.cmake
+++ b/CMake/SlicerMacroBuildApplication.cmake
@@ -556,6 +556,45 @@ macro(slicerMacroBuildApplication)
           OUTPUTVAR extraApplicationToLaunchListForBuildTree
           )
       endif()
+
+      # Make available Qt language tools conveniently on the command line, for example:
+      # Slicer.exe --lconvert ...
+      if(Slicer_BUILD_I18N_SUPPORT)
+        if(NOT QT_LCONVERT_EXECUTABLE)
+          find_program(QT_LCONVERT_EXECUTABLE lconvert HINTS "${QT_BINARY_DIR}" NO_DEFAULT_PATH)
+        endif()
+        if(EXISTS ${QT_LCONVERT_EXECUTABLE})
+          ctkAppLauncherAppendExtraAppToLaunchToList(
+            LONG_ARG lconvert
+            HELP "Start Qt lconvert language utility"
+            PATH ${QT_LCONVERT_EXECUTABLE}
+            OUTPUTVAR extraApplicationToLaunchListForBuildTree
+            )
+        endif()
+        if(NOT QT_LRELEASE_EXECUTABLE)
+          find_program(QT_LRELEASE_EXECUTABLE lrelease HINTS "${QT_BINARY_DIR}" NO_DEFAULT_PATH)
+        endif()
+        if(EXISTS ${QT_LRELEASE_EXECUTABLE})
+          ctkAppLauncherAppendExtraAppToLaunchToList(
+            LONG_ARG lrelease
+            HELP "Start Qt lrelease language utility"
+            PATH ${QT_LRELEASE_EXECUTABLE}
+            OUTPUTVAR extraApplicationToLaunchListForBuildTree
+            )
+        endif()
+        if(NOT QT_LUPDATE_EXECUTABLE)
+          find_program(QT_LUPDATE_EXECUTABLE lupdate HINTS "${QT_BINARY_DIR}" NO_DEFAULT_PATH)
+        endif()
+        if(EXISTS ${QT_LUPDATE_EXECUTABLE})
+          ctkAppLauncherAppendExtraAppToLaunchToList(
+            LONG_ARG lupdate
+            HELP "Start Qt lupdate language utility"
+            PATH ${QT_LUPDATE_EXECUTABLE}
+            OUTPUTVAR extraApplicationToLaunchListForBuildTree
+            )
+        endif()
+      endif()
+
       set(executables)
       if(UNIX)
         list(APPEND executables gnome-terminal xterm)
@@ -613,6 +652,35 @@ macro(slicerMacroBuildApplication)
           PATH "<APPLAUNCHER_SETTINGS_DIR>/../bin/designer-real${CMAKE_EXECUTABLE_SUFFIX}"
           OUTPUTVAR extraApplicationToLaunchListForInstallTree
           )
+      endif()
+
+      # Make available Qt language tools conveniently on the command line, for example:
+      # Slicer.exe --lconvert ...
+      if(Slicer_BUILD_I18N_SUPPORT)
+        if(EXISTS ${QT_LCONVERT_EXECUTABLE} AND NOT APPLE)
+          ctkAppLauncherAppendExtraAppToLaunchToList(
+            LONG_ARG lconvert
+            HELP "Start Qt lconvert language utility"
+            PATH "<APPLAUNCHER_SETTINGS_DIR>/../bin/lconvert${CMAKE_EXECUTABLE_SUFFIX}"
+            OUTPUTVAR extraApplicationToLaunchListForInstallTree
+            )
+        endif()
+        if(EXISTS ${QT_LRELEASE_EXECUTABLE} AND NOT APPLE)
+          ctkAppLauncherAppendExtraAppToLaunchToList(
+            LONG_ARG lrelease
+            HELP "Start Qt lrelease language utility"
+            PATH "<APPLAUNCHER_SETTINGS_DIR>/../bin/lrelease${CMAKE_EXECUTABLE_SUFFIX}"
+            OUTPUTVAR extraApplicationToLaunchListForInstallTree
+            )
+        endif()
+        if(EXISTS ${QT_LUPDATE_EXECUTABLE} AND NOT APPLE)
+          ctkAppLauncherAppendExtraAppToLaunchToList(
+            LONG_ARG lupdate
+            HELP "Start Qt lupdate language utility"
+            PATH "<APPLAUNCHER_SETTINGS_DIR>/../bin/lupdate${CMAKE_EXECUTABLE_SUFFIX}"
+            OUTPUTVAR extraApplicationToLaunchListForInstallTree
+            )
+        endif()
       endif()
 
       include(SlicerBlockCTKAppLauncherSettings)


rbumm:
This works well for "Creating new fiducial" (all lines to each new fiducial drawn) but fails when "Place a control point". The letter causes only one line to be drawn (to the last-placed control point). Any ideas? 

Saving and loading are not yet implemented.

lassoan:
Thanks a lot, this looks very promising!

1. The main issue is that you added only a single vtkLineSource for an entire markup node. Since one markup node can display many control points, you need to add a polydata source (or polydata object) that can display lines for every control point.

Probably the simplest is to add a vtkPolyData object and set its points (control point pos 0, label pos 0 control point pos 1, label pos 1, ...) and line cells (cell 0: 0, 1; cell 1: 2, 3; cell 2: 4, 5; ...).

2. If a label reaches the edge of the view then it should not be pushed any further (outside of the view), but it should be kept in the view (if the control points get very close to the view edge then the label can be displayed at 0 distance - the same way as display without label connector. 

For example, I've tried to increase the label distance so that F6 does not occlude the view. But by the time F6 got out of the way, al the other labels were pushed outside the view.

https://user-images.githubusercontent.com/307929/154786964-3b01ba22-37f0-4ef2-9417-3ce7ec61c183.mp4

3. Hide line when "Control point labels" is disabled.

4. Rename "Label connector" to "Label Distance" or "Label spokes" and put that under "Control Point Labels", with an indentation of 10-20.

It would look something like this:

```
Control Point Labels: [ ]
    Label spokes:   [ ]   ---------||--------------------
```

rbumm:
Now includes labeltext 2D border collision as well as XML read and write of connector line settings in the markups file
 
Before implementing 3D: @lassoan would be great if you could look at // ?? comments in the source, maybe you can propose solutions thanks

rbumm:
The commits of today add 3D label connector lines and line angulations which take into account the RASBounds of the first scalar volume in the scene. The implementation is stable and has quite a good performance in all views. 3D view does not offer border collision or label text collision avoidance yet in order not to make it too fancy. Label texts can now span multilines by putting an \<br> in the label string.

lassoan:
Thanks for the update, it looks good oks great! In the final solution we'll need something more controllable than the first master volume bounds. For example, we could use the bounds of the purple box or bounds of a chosen node. The node could be set by default to the first node a point is placed on.

rbumm:
>> For example, we could use the bounds of the purple box or bounds of a chosen node.

The box may not be correctly adjusted so not showing meaningful bounds and markups and volume nodes are not always connected, so I implemented the use of the threeDview node ->GetFocalPoint() world coordinates as the cutoff points for angulation changes.

rbumm:
The colors of LabelConnectorLines can now be set, in the Markups Widget and are saved and loaded. 
I would need a comment please how much is still missing before this could be merged into a preview. Just for better testing more comments.  
 

image

lassoan:
Thanks a lot, I'll test this!
What's the motivation for using a different color for line than for the text and glyph? (if there is a good reason to use a different color then it's OK, but still by default it would be nice to use the same color - unless you have found that the same color is suboptimal for some reason)

rbumm:
\>> What's the motivation for using a different color

(1) maybe I did not see the most obvious way to do that but (2) viewer attention is a bit "drawn" to the glyphs when using different colors like:

lassoan:
> maybe I did not see the most obvious way to do that but (2) viewer attention is a bit "drawn" to the glyphs when using different colors like:

OK, I can see that. As you can see there is already much more configuration options in the Markups module than anyone would want to see, so we should make an effort to add as little to this as possible. Also, GUI always requires a lot of work to implement, maintain, document, and support (train users, answer questions, etc.). So, I would recommend to figure out a way to automatically generate a suitable line color. For example, you can do something similar that is done for the backside coloring: convert the RGB color to HSV, slightly change the brightness, saturation, or hue parameter, and then convert back to RGB and use that.

rbumm:
I removed the color picker button from "Markups" again and implemented an automatic connector line color algorithm as @lassoan suggested. Distances of label texts in 3D view are slightly modified in addition. 

image

rbumm:
Would be glad if you could shortly comment on how to proceed with this PR,  @lassoan

 Thank ypu.

lassoan:
I'm building it now. Will test it and review the code today.

jamesobutler:
@rbumm Could you rebase this branch to bring in all the latest commits from `master` and let the status checks re-run here? Your branch is currently 219 commits behind `master` which is quite a lot.

rbumm:
I rebased the development branch but obviously did something wrong. 

Ran "Fetch upstream" on Github, this produced the other merge commit. 

Then I pulled to my development computer, and now subsequent build attempts fail. Unfortunately, I can not get rid of the merge commit momentarily.

lassoan:
Thank you, this will be awesome. This is all quite hard, complex development tasks, so what you've done is really impressive.

There are a number of things to fix/improve:

- [ ] Position of the label is not computed correctly in 3D views

https://user-images.githubusercontent.com/307929/163865473-d9519c6e-64c7-4784-80c2-fde593d2347a.mp4

- [ ] Position of the label in slice views seem to use some fixed 45 deg angle instead of depending on the position in the image. If the label is on the left side, near the vertical center of the image then it either jumps up or down a lot, the center is never used, so the labels get crowded at the top or bottom.

https://user-images.githubusercontent.com/307929/163866255-f65755af-eeb7-415b-96e6-1b823cabe0c3.mp4

- [ ] Rename option to "Leader lines" to follow nomenclature of Excel (https://www.extendoffice.com/documents/excel/2351-excel-pie-chart-leader-lines.html) -- if you have other software that uses different term for this then we can discuss
- [ ] Move the option under the "Control point labels" checkbox in Advanced section (indented a bit)
- Hide the leading line if control point label is not visible:
  - [ ] Because all control points are hidden ("Control point labels" option is unchecked)
  - [ ] Because the control point is occluded
  - [ ] Because the label text is empty
- [ ] Scale the line thickness better. On my computer the lines are way too thin (barely visible) in slice views and too thick (compared to the glyph size) in 3D views. Probably scaling it the same way as the line in markups line nodes would work. To avoid adding too many options on the GUI (we would need an absolute/relative flag and a thickness slider), we could add a single thickness display node property, set to 25% by default, and interpret it as relative to glyph size (if glyph size is absolute, then use absolute size; if glyph size is relative then use relative size for the line, too). This thickness probably does not have to be exposed in the GUI (just to keep things simpler).
- [ ] Keeping labels in slice views is a really nice feature. It works very well on the left and bottom sides. It cuts the labels into half at the top and right side, instead of keeping them in the view..

Anything is open to discussion (if some things are extremely hard to implement then we can try to find acceptable solutions that are easier to implement) and if you cannot address all issues then we can find other Slicer developers to help out.

lassoan:
If you got stuck with your branch then you can check out the Slicer:master branch and cherry-pick your only commit (skip the merge commits). You can then force-push this new branch to the rbumm:Branch_markuplabeltexts branch.

lassoan:
@rbumm I've rebased your branch on the latest Slicer master and force-pushed to your fork. You can check out your rbumm:Branch_markuplabeltexts branch and it should build without problems.

rbumm:
> @rbumm I've rebased your branch on the latest Slicer master and force-pushed to your fork. You can check out your rbumm:Branch\_markuplabeltexts branch and it should build without problems.

Thank you @lassoan for the rebase and the thoughtful comments above. Unfortunately, I am stuck in a nasty Visual Studio update problem on my main development computer right now and need to solve that first before being able to check the rebased branch and look into all the details.

rbumm:
> > @rbumm I've rebased your branch on the latest Slicer master and force-pushed to your fork. You can check out your rbumm:Branch\_markuplabeltexts branch and it should build without problems.
> 
> Thank you @lassoan for the rebase and the thoughtful comments above. Unfortunately, I am stuck in a nasty Visual Studio update problem on my main development computer right now and need to solve that first before being able to check the rebased branch and look into all the details.

@lassoan you rebased my branch in Tortoisegit on Slicer master, correct? How did you manage to build the branch after the rebase?  Do I have to perform a 5 h make after I do such a rebase with my S4?

rbumm:
> ..  are likely changes in Slicer that need all the external project updates which would require you to do a new fresh build to not run into any incompatibilities.

Exactly what happened, thanks for the explanation.

rbumm:
> There are a number of things to fix/improve:
> 
>  Position of the label is not computed correctly in 3D views

Please explain, where are they wong? 

> 2022-04-18\_15-12-21.mp4
> 
>  Position of the label in slice views seem to use some fixed 45 deg angle instead of depending on the position in the image. If the label is on the left side, near the vertical center of the image then it either jumps up or down a lot, the center is never used, so the labels get crowded at the top or bottom.
> 
> 2022-04-18\_15-38-11.mp4

Agree, but had no time to realize an alternative yet. This would probably be the final thing to do.  I could probably improve the situation myself, but it will take a long - the actual implementation already took extremely long for me and I would like to have the leader lines included in Slicer as soon as possible.  A solution is difficult if I can not determine the actual text box around a label text in 2D and 3D.  

>  Rename option to "Leader lines" to follow nomenclature of Excel (https://www.extendoffice.com/documents/excel/2351-excel-pie-chart-leader-lines.html) -- if you have other software that uses different term for this then we can discuss

Done. 

>  Move the option under the "Control point labels" checkbox in Advanced section (indented a bit)

Done. 

> Hide the leading line if control point label is not visible:
> 
> *   [ ] Because all control points are hidden ("Control point labels" option is unchecked)

Done

> *   [ ] Because the control point is occluded

Not yet done. In 2D I do not see when occlusion would be a problem - in 3D from my view would it not be good to have a leader line even if the control point is currently occluded - making the user want to turn the 3D model to uncover it? 

> *   [ ] Because the label text is empty

Done. 

>  Scale the line thickness better. On my computer the lines are way too thin (barely visible) in slice views and too thick (compared to the glyph size) in 3D views. Probably scaling it the same way as the line in markups line nodes would work. To avoid adding too many options on the GUI (we would need an absolute/relative flag and a thickness slider), we could add a single thickness display node property, set to 25% by default, and interpret it as relative to glyph size (if glyph size is absolute, then use absolute size; if glyph size is relative then use relative size for the line, too). This thickness probably does not have to be exposed in the GUI (just to keep things simpler).

Done. 25 % suggestion works well. 

>  Keeping labels in slice views is a really nice feature. It works very well on the left and bottom sides. It cuts the labels into half at the top and right side, instead of keeping them in the view..

I made a change, it works on my system, please test. Had some problems mentioned earlier in determining the dimensions of the actual text box around a label text. Any ideas on how to do this in 2D and 3D are welcome.

rbumm:
> Position of the label is not computed correctly in 3D views

\>> Please explain, where are they wong?

I now see it. The label positions get out of balance during rotation.  

\>> Because the label text is empty

Added one more: leader lines are not drawn if "Control Point Labels:" is checked off in "Display"->"Advanced"

rbumm:
There is progress in 2D by implementing some trigonometry. 

Leader Lines are now flexibly pointing to the "center" of the loaded volume and when the volume is moved their orientation stays the same. As this is now fixed, center determination becomes a bit problem. I need to know the center in xytoRAS and am lost here.  Already in the red slice, ras(0,0,0) is not the center but slightly anterior. 



Unfortunately, this does not work in all slice views (see green), because I need a robust, preferable fast method that can be called from tkSlicerMarkupsWidgetRepresentation2D::UpdateAllPointsAndLabelsFromMRML to determine the sliceCenterRAS position IN EACH VIEW . I need help here.

I am away next weekend but it would be great if someone could come up with a solution.

rbumm:
Two short demos of the PR state as of May 3rd 2022 (new push will follow: 

Slice view border collision demo:

https://user-images.githubusercontent.com/18140094/166478805-66d2d105-7b1d-4d52-a04f-9ef9363c7498.mp4

2D Slice view markup rotation demo (with leader lines) 

https://user-images.githubusercontent.com/18140094/166478958-a43d478d-1490-4e4d-84c0-0ebf51cbc5cb.mp4

lassoan:
> I need a robust, preferable fast method that can be called from tkSlicerMarkupsWidgetRepresentation2D::UpdateAllPointsAndLabelsFromMRML to determine the sliceCenterRAS position

I think you can get the slice center's RAS position by multiplying the slice node's XYZOrigin by the slice node's SliceToRAS transform.

rbumm:
Got it working to orientate the leader lines all to volume center point in 2D and 3D, thus saving space.

image

rbumm:
Another idea for later would be to ray out leader lines from one control point label to multiple control points if they all have identical names (f.e. "Nodule").

rbumm:
Leader lines work well now in many tests, however, the thickness of the tubes is not updated well if the user zooms in and out in 3DView. Where would be the entry point for handling the zoom events and how would I force a 3D View update incl markup redraw from there elegantly? @lassoan / @jamesobutler  any idea welcome. Thanks.

rbumm:
> however, the thickness of the tubes is not updated well if the user zooms in and out in 3DView. Where would be the entry point for handling the zoom events and how would I force a 3D View update incl markup redraw from there elegantly?

Please disregard. Problem solved in by implementing a timely update of tube filter radius in vtkSlicerMarkupsWidgetRepresentation3D.cxx. See 96d088d

rbumm:
Probably time for another test @lassoan. I will be traveling until Tuesday the 17th and it would be great if we could get this included next week. In 3D there is no collision or occlusion detection.  Thank you.

rbumm:
I would like to revive this PR. The current version still builds fine on my development system but the Slicer version is probably outdated right now. 

\>>However, if your Slicer build tree was built in late January without any updates, then there are likely changes in Slicer that need >>all the external project updates which would require you to do a **new fresh build** to not run into any incompatibilities.

 @lassoan or @jamesobutler could you help me do this fresh build without the foreseeable pain? What is the easiest and most uncomplicated workflow?  Could you do anything from your side? Just a bit nervous I mess something up.

jamesobutler:
Upon rebasing this branch to include all latest Slicer commits you could rebuild the inner "Slicer-build" project. However if your local build is from January, you really need to do a full fresh build as VTK and ITK and other external dependencies have been updated since then. It's been too long to avoid a full rebuild.

rbumm:
> Upon rebasing this branch to include all latest Slicer commits you could rebuild the inner "Slicer-build" project. However if your local build is from January, you really need to do a full fresh build as VTK and ITK and other external dependencies have been updated since then. It's been too long to avoid a full rebuild.

So I would delete S4 and S4R, fork a new S4 and run the 5 h make ? Then implement the current PR, but how exactly to avoid another 5 h build ?

jamesobutler:
1. Clone https://github.com/rbumm/Slicer to C:\S5
2. Checkout Branch_markuplabeltexts
3. In CMake-GUI the "source code" is C:\S5 and "Where to build the binaries" is "C:\S5R". Configure and generate the Visual Studio (if on Windows) solution . 
4. Run the All Build project in the main Slicer.sln located in C:\S5R -> this is the multiple hour step

rbumm:
\>>Clone [https://github.com/rbumm/Slicer](https://github.com/rbumm/Slicer) to C:\\S5

Must I not fetch the upstream first in rbumm/Slicer/ ? 

Can I do this in the github web interface ?

Should I also fetch the upstream for my Branch\_markuplabeltexts ?  

Fetch both an the fork ?

jamesobutler:
Are you asking for instructions about how to rebase this branch to include the latest Slicer commits? Or just instructions on how to generally build Slicer?

rbumm:
The former. Thank you for your answers.

jamesobutler:
I would suggest the following for you:
- Download GitHub Desktop https://desktop.github.com/
- Clone your fork https://github.com/rbumm/Slicer](https://github.com/rbumm/Slicer to C:\S5 (follow instructions that you will be contributing to upstream to set up this connection to the Slicer remote)
- Checkout Branch_markuplabeltexts
- In the GitHub Desktop menu bar go to Branch->Rebase current branch...
- Select "upstream/main" (aka https://github.com/Slicer/Slicer/tree/main) and it will say something like "This will update Branch_markuplabeltexts by applying its 1 commit on top of upstream/main".
- Press the rebase button. Confirm locally in the branch history of commits that your commit is now on top of latest slicer commits
- Follow instructions to force push this branch (you have rewritten history to bring in the latest Slicer commits).

rbumm:
> * Select "upstream/main" (aka https://github.com/Slicer/Slicer/tree/main) and it will say something like "This will update Branch_markuplabeltexts by applying its 1 commit on top of upstream/main".

upstream/main shows up, but GitHub Desktop reports "This branch is up to date with master. 
image
Is it maybe because my Slicer main branch is still called "master", the upstream now called "main"?

jamesobutler:
Make sure you have done the previous step of “Checkout Branch_markuplabeltexts”. There is a big branch selector drop down in the application next to the repo selection drop down and next to the fetch/push button. It appears you still have master checked out (likely at the step of having cloned your fork)

rbumm:
Oh. 
This was super easy to do now. I thought I had to use "master" and not my "branch_markuplabeltexts"
I have rebased and force pushed the result for testing, although nothing has really changed in my source changes. 
 :-)
Thanks a lot for your help @jamesobutler

jamesobutler:
I've pushed to this branch to fix merge conflicts.

rbumm:
I have rebased this branch only to update recent Slicer changes. 
@lassoan or @jamesobutler  could you review this when you find the time?
Any suggestions on how to proceed?

jcfr:
_For convenience, topic has just been rebased_

lassoan:
Let's merge this right after 5.2 is released so that we have a bit more time for testing and tuning.

rbumm:
Ok, thanks, @lassoan.  Are you aware of a way to predict the xy(z) space a label text will cover in a 2D (3D) view? This is astonishingly difficult to implement - if you have an idea please let me know.

lassoan:
Do you mean that when you specify a label position in 2D (or 3D?) and you are not sure about the reaulting size and position of that label in 2D? I've come across this question before and could not fully answer it (that's why the labels are somewhat closer to control points than I would like). Maybe the best would be to create a small test script that sets up a VTK render window and adds some point glyphs and labels and experiment with that. If we work with a standalone script then we could also ask questions about it on the VTK forum.

lassoan:
It proves the complexity of the problem that it seems that even a VTK core development project was failed to implement this. The [label placement mapper](https://vtk.org/doc/nightly/html/classvtkLabelPlacementMapper.html) class was supposed to implement automatic non-overlapping placement  of labels with priority-based hiding of labels with lower importance in cases where label were too dense. However, as far as I can tell, the implementation is incomplete and has severe bugs in it that make it practically unusable (that's why we don't use any of the dynamic label hiding/relocating features of this mapper in Slicer).

One potential solution would be to fix and finish the implementation of this mapper. It might be easier to implement label placement at this lower level, in the mapper, than to figure out the correct position at the application level, because inside the mapper you have easier access to lots of lower-level information (e.g., the mapper surely knows the bounds of each label). The label placement strategy that you implemented in Slicer could be a new placement mode in the mapper.

rbumm:
> Do you mean that when you specify a label position in 2D (or 3D?) and you are not sure about the reaulting size and position of that label in 2D? I've come across this question before and could not fully answer it (that's why the labels are somewhat closer to control points than I would like). Maybe the best would be to create a small test script that sets up a VTK render window and adds some point glyphs and labels and experiment with that. If we work with a standalone script then we could also ask questions about it on the VTK forum.

Yes, I would need the text box size around a given single or multiline text and position it in a way that the text fits into the space between the left border of the box and the connector line end position. It may be familiar to you because I had asked that further up in this thread ...

rbumm:
I tried to rebase the branch, and again I messed it up and it did not work. Unfortunately, this PR was automatically closed without intention. @Please advise what I should do next as I can not reopen the PR. Thank you

pieper:
I'm not sure why you this can't be reopened.  Can you check out the most recent commit of the PR into a new branch and then create a new PR?  You could include a link to this thread so the comments are accessible for review of the new PR.

rbumm:
Thank you @pieper, I'll leave it as it is until now and hope that someone from the Slicer main owner group can reopen this PR. I have reset my Branch\_markuplabeltexts to [`d632b2`](https://github.com/Slicer/Slicer/commit/d632b2633c07e4d8b8b80b0066f0ce64ebf20072)\-> to the state where I started all this 6 h ago.

rbumm:
Thank you @pieper, I'll leave it as it is until now and hope that someone from the Slicer main owner group can reopen this PR. I have reset my Branch\_markuplabeltexts to [`d632b2`](https://github.com/Slicer/Slicer/commit/d632b2633c07e4d8b8b80b0066f0ce64ebf20072)\-> to the state where I started all this 6 h ago.

jamesobutler:
Saw this comment somewhere

> We're blocking the pull request reopen if the current head isn't a descendant of the stored head sha (which is what the head was when the pull request was closed). We are not allowing the reopen in that case, because there is no good way to tell what changes have happened while a pull request was closed and the head branch has changed.

Where the force push of the `main` branch last week caused this:
image

Not allowing force pushes on the protected `main` branch could help prevent this:
image


Maybe this can help https://gist.github.com/robertpainsi/2c42c15f1ce6dab03a0675348edd4e2c?

lassoan:
> Not allowing force pushes on the protected main branch could help prevent this:

The main branch is protected (no force pushes are allowed), I just temporarily disabled it to force-push a single commit (to make some small fix in a commit without with a simpler version history). Probably @rbumm was unlucky and made a pull right at that time.

We should be able to fix this quickly during the project week and also review, test, and integrate if all looks good.

lassoan:
I've rebased the branch, fixed merge conflicts, did some cleanups, and created a new pull request: https://github.com/Slicer/Slicer/pull/6809

PR #6193, opened 2022-02-16
ENH: Add Markups Point List Node

jamesobutler:
This is work for #5989.

In https://github.com/Slicer/Slicer/commit/8e294598acddf71414bc2f1c2ff7c67b2e871683 the vtkMRMLMarkupsFiducialNode's were given a display name of "Point List" instead of "Fiducial". This PR continues on this work by making an actual Markups Point List node that will be the replacement for Markups Fiducial nodes. The markups module and markups toolbar now create vtkMRMLMarkupsPointListNodes instead of vtkMRMLMarkupsFiducialNodes. I have also updated script repository documentation to use these new vtkMRMLMarkupsPointListNode objects where now the API matches the GUI terminology.

vtkMRMLMarkupsFiducialNode is now a derived class of vtkMRMLMarkupsPointListNode with the plan to ultimately remove this derived class in the future upon giving time to transition to vtkMRMLMarkupsPointListNode.

image

### DIFF
--- a/Modules/Loadable/Markups/Logic/vtkSlicerMarkupsLogic.cxx
+++ b/Modules/Loadable/Markups/Logic/vtkSlicerMarkupsLogic.cxx
@@ -34,6 +34,7 @@
 #include "vtkMRMLMarkupsPlaneDisplayNode.h"
 #include "vtkMRMLMarkupsPlaneJsonStorageNode.h"
 #include "vtkMRMLMarkupsPlaneNode.h"
+#include "vtkMRMLMarkupsPointListNode.h"
 #include "vtkMRMLMarkupsROIDisplayNode.h"
 #include "vtkMRMLMarkupsROIJsonStorageNode.h"
 #include "vtkMRMLMarkupsROINode.h"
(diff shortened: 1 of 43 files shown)


Modules/Loadable/Markups/Logic/vtkSlicerMarkupsLogic.cxx
```diff
@@ -386,10 +387,13 @@ void vtkSlicerMarkupsLogic::RegisterNodes()
   scene->RegisterNodeClass(vtkSmartPointer<vtkMRMLMarkupsJsonStorageNode>::New());
 
   // NOTE: the order of registration determines the order of the create push buttons in the GUI
+  vtkNew<vtkMRMLMarkupsPointListNode> pointListNode;
+  vtkNew<vtkSlicerPointsWidget> pointsWidget;
```
lassoan, line 391: Shouldn't this be vtkSlicerPointListWidget?

lassoan (commented):
Thank you, this refactoring will be useful for bringing the GUI and source code closer to each other.

We should update the file reader so that point list node is generated when reading a .mrk.json file with either a `PointList` or `Fiducial` markup type (or .fcsv file). When a scene is loaded then when the class is `vtkMRMLMarkupsFiducialNode` then a `vtkMRMLMarkupsPointListNode` should be instantiated.

I think there would be less disruptions if the class hierarchy would be vtkMRMLMarkupsNode->vtkMRMLMarkupsFiducialNode->vtkMRMLMarkupsPointListNode, because extensions would mostly recognize the vtkMRMLMarkupsPointListNode as a vtkMRMLMarkupsFiducialNode, so extensions would not need to be changed (except if they use node selectors with `showChildNodeTypes` disabled or checking class with class name comparison instead of `object->IsA` method). Slicer core can be prepared to everywhere accept both vtkMRMLMarkupsFiducialNode and vtkMRMLMarkupsPointListNode. In contrast, if the class hierarchy was vtkMRMLMarkupsNode->vtkMRMLMarkupsPointListNode->vtkMRMLMarkupsFiducialNode then all the extensions would need to be prepared to accept both point list and fiducial nodes everywhere, which would be much harder.

It would take a while to test and stabilize Slicer core and all extensions after this change, so I think we should do it after Slicer5 is released.

jamesobutler:
The plan would also be that if an extension or other code specifically calls to create a vtkMRMLMarkupsFiducialNode that they would be met with a deprecation message that would encourage use of vtkMRMLMarkupsPointListNode with the expectation that the fiducial node files to be removed from the code eventually.

lassoan:
> The plan would also be that if an extension or other code specifically calls to create a vtkMRMLMarkupsFiducialNode that they would be met with a deprecation message that would encourage use of vtkMRMLMarkupsPointListNode with the expectation that the fiducial node files to be removed from the code eventually.

This would not be sufficient. If the class hierarchy of the current PR was used then extensions would not work, as they would not recognize the vtkMRMLMarkupsPointListNode objects that are created by Slicer core modules (users could not use any of point lists that they created in Slicer core in Slicer extensions, until the extension is updated to use the new node type).

jamesobutler:
Closing now due to all of the merge conflicts. Starting over will be required.

PR #6195, opened 2022-02-16
ENH: Simplify OSX Deployment target setting

hjmjohnson:


jamesobutler (commented):
You can remove the updated CMake min version from this PR as that has already been integrated in https://github.com/Slicer/Slicer/commit/664fecaf8d70acdfaba30a473fe35931a4999e47. `ENH: Simplify OSX Deployment target setting` from this PR can be integrated on its own.

jamesobutler (approved):

jcfr:
Thanks @hjmjohnson and @jamesobutler 👌🙏

PR #6196, opened 2022-02-16
BUG: Fix plane/ROI measurements to account for parent transform scaling

Sunderlandkyl:
Previously the plane and ROI Markups did not account for parent transform scaling when calculating the volume/area. Fixed by using GetSizeWorld methods.
The measurements now also automatically update if the parent transform node is changed.

Also fixes the position of plane scale handles when scaled by a parent transform.

Re #5061

### DIFF
--- a/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsPlaneNode.cxx
+++ b/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsPlaneNode.cxx
@@ -952,6 +952,58 @@ void vtkMRMLMarkupsPlaneNode::SetSize(double x, double y)
 }
 
 //----------------------------------------------------------------------------
+void vtkMRMLMarkupsPlaneNode::GetSizeWorld(double size_World[2])
+{
+  vtkNew<vtkMatrix4x4> objectToWorldMatrix;
+  this->GetObjectToWorldMatrix(objectToWorldMatrix);
+  vtkNew<vtkTransform> objectToWorldTransform;
+  objectToWorldTransform->SetMatrix(objectToWorldMatrix);
+
+  double sideVectorX_Object[3] = { this->Size[0], 0.0, 0.0 };
+  double sideVectorY_Object[3] = { 0.0, this->Size[1], 0.0 };
+
+  size_World[0] = vtkMath::Norm(objectToWorldTransform->TransformVector(sideVectorX_Object));
+  size_World[1] = vtkMath::Norm(objectToWorldTransform->TransformVector(sideVectorY_Object));
+}
+
+//----------------------------------------------------------------------------
+double* vtkMRMLMarkupsPlaneNode::GetSizeWorld()
+{
+  this->GetSizeWorld(this->SizeWorld);
+  return this->SizeWorld;
+}
+
+//----------------------------------------------------------------------------
+void vtkMRMLMarkupsPlaneNode::SetSizeWorld(const double size_World[2])
+{
+  this->SetSizeWorld(size_World[0], size_World[1]);
+}
+
+//----------------------------------------------------------------------------
+void vtkMRMLMarkupsPlaneNode::SetSizeWorld(double sizeX_World, double sizeY_World)
+{
+  vtkNew<vtkMatrix4x4> objectToWorldMatrix;
+  this->GetObjectToWorldMatrix(objectToWorldMatrix);
+  vtkNew<vtkTransform> worldToObjectTransform;
+  worldToObjectTransform->SetMatrix(objectToWorldMatrix);
+  worldToObjectTransform->Inverse();
+
+  double sideVectorX_World[3] = { 0.0, 0.0, 0.0 };
+  double sideVectorY_World[3] = { 0.0, 0.0, 0.0 };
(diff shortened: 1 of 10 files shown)


Modules/Loadable/Markups/MRML/vtkMRMLMarkupsPlaneNode.cxx
```diff
@@ -951,6 +951,58 @@ void vtkMRMLMarkupsPlaneNode::SetSize(double x, double y)
   this->Modified();
 }
 
+//----------------------------------------------------------------------------
+void vtkMRMLMarkupsPlaneNode::GetSizeWorld(double size_World[2])
```
lassoan, line 955: vtkSlicerPlaneRepresentation3D::BuildPlane()/vtkSlicerPlaneRepresentation2D::BuildPlane() implementation computes the corner points quite differently compared from what this method does. This has a slight risk of inconsistency between what is displayed and what is computed (even if it works perfectly now, developers may make changes in the future and they may not consistenly do it in all 3 methods). Things may become non-trivial when multiple transforms are applied on the plane (including warping) and/or the bounds of the plane are shifted (so that the plane center is not the same as the control point position), etc.

Could you refactor the computations so that all 3 methods to ensure their consistency? For example, a GetPlaneCornerPointsWorld() method could be added to vtkMRMLMarkupsPlaneNode that would compute the corner positions and that would be used in GetSizeWorld and BuildPlane methods. This new method might also simplify other methods, such as GetRASBounds, GetBounds, CalculatePlaneBounds (and ensure consistency).

Modules/Loadable/Markups/MRML/vtkMRMLMeasurementArea.cxx
```diff
@@ -79,9 +79,9 @@ void vtkMRMLMeasurementArea::Compute()
     }
   else if (planeNode)
     {
-    double size[2] = { 0.0 };
-    planeNode->GetSize(size);
-    area = size[0] * size[1];
+    double size_world[2] = { 0.0 };
```
lassoan, line 82: This kind of initialization is somewhat misleading, as it may suggest that it affects all values in the vector, while it only changes the first value (the others are set to 0.0).

Examples:

```
x[3] => 6.95324e-310 4.94066e-324 6.95324e-310
x[3] = {} => 0 0 0
x[3] = {1.0}; => 1 0 0
x[3] = {1.0, 1.0, 1.0}; => 1 1 1
```

https://onlinegdb.com/0w5Ry2JbL

So, if we initialize to 0.0 then we should use `{}` or if we want to be more explicit then to `{0.0, 0.0}`. Since explicitly writing down all the values is harder to misinterpret, it works for all values (not just 0.0), and we do this at many places already, it is probably better to use `{0.0, 0.0}` here, too.

lassoan (commented):
Thank you! I've added a few small comments.

Sunderlandkyl:
Thanks, I've fixed the initialization and refactored several methods to use the new vtkMRMLMarkupsPlaneNode::GetPlaneCornerPointsWorld method.

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsPlaneNode.cxx
```diff
@@ -951,6 +951,58 @@ void vtkMRMLMarkupsPlaneNode::SetSize(double x, double y)
...
+{
+  vtkNew<vtkMatrix4x4> objectToWorldMatrix;
+  this->GetObjectToWorldMatrix(objectToWorldMatrix);
+  vtkNew<vtkTransform> objectToWorldTransform;
+  objectToWorldTransform->SetMatrix(objectToWorldMatrix);
+
+  double sizeX_Object3[3] = { this->Size[0], 0.0, 0.0 };
```
lassoan, line 962: This `sizeX_Object3` name is not clear. Could you rename it to something like `scaledAxisXVector_Object` or `sideXVector_Object` or something similar(include `Vector` in the name and don't call it size, because you already use size for scalars).

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsPlaneNode.cxx
```diff
@@ -951,6 +951,58 @@ void vtkMRMLMarkupsPlaneNode::SetSize(double x, double y)
...
+//----------------------------------------------------------------------------
+void vtkMRMLMarkupsPlaneNode::SetSizeWorld(double sizeX_World, double sizeY_World)
+{
+  vtkNew<vtkMatrix4x4> objectToWorldMatrix;
+  this->GetObjectToWorldMatrix(objectToWorldMatrix);
+  vtkNew<vtkTransform> worldToObjectTransform;
+  worldToObjectTransform->SetMatrix(objectToWorldMatrix);
+  worldToObjectTransform->Inverse();
+
+  double sizeX_World3[3] = { 0.0, 0.0, 0.0 };
```
lassoan, line 991: rename - similarly as in GetSize

lassoan (commented):
This looks good, thank you. I've added two nitpicky comments about renaming variables, it would be nice if you could address those.

lassoan:
Thank you, merged!

PR #6198, opened 2022-02-17
COMP: Update NSIS manifest to support packaging of files with a long paths

jcfr:
Support packaging of application where install prefix like the following
would lead to paths having their length beyond the 260 characters limit:

```
  C:/path/to/Slicer-build/_CPack_Packages/win-amd64/NSIS/<Slicer_MAIN_PROJECT>-X.Y.Z-YYYY-MM-DD-win-amd64
```

References:
* Section "4.8.1.30 ManifestLongPathAware" at
  https://nsis.sourceforge.io/Docs/Chapter4.html#attribgen
* https://www.tenforums.com/tutorials/51704-enable-disable-win32-long-paths-windows-10-a.html


jcfr:
This approach doesn't seem to address the `failed opening file` error corresponding to the following code:

```cpp
  HANDLE hFile = CreateFile(
    newfn,
    GENERIC_READ,
    FILE_SHARE_READ,
    NULL,
    OPEN_EXISTING,
    FILE_ATTRIBUTE_NORMAL | FILE_FLAG_SEQUENTIAL_SCAN,
    NULL
  );
  if (hFile == INVALID_HANDLE_VALUE)
  {
    ERROR_MSG(_T("%") NPRIs _T("File: failed opening file \"%") NPRIs _T("\"\n"),generatecode?_T(""):_T("Reserve"),newfn);
    return PS_ERROR;
  }
```

Source: https://github.com/kichik/nsis/blob/805c093d30e38792d0f1a954507ee79b68cbd1b8/Source/script.cpp#L5308-L5321

Looking at the commit adding support for `ManifestLongPathAware` (https://github.com/kichik/nsis/commit/d19b13142f41573861d5afeb0db7e5c704d0a65d), it takes care of adding the element `<longPathAware  xmlns="http://schemas.microsoft.com/SMI/2016/WindowsSettings">true</longPathAware>` to the manifest of the application being packaged.

It does **not** allow to package files found in longer paths:
* https://nsis-dev.github.io/NSIS-Forums/html/t-123323.html
* https://stackoverflow.com/questions/24295076/createfile-and-long-device-name

jcfr:
The testing of this was likely done on a system without support for long path enabled, considering that this is a small enough change, we will be merging.

Issue #6203, opened 2022-02-18
Scissors circle effect change/improvement drawing it by the borders


fbordignon:
Hey guys, I want to run by you an idea that I had to perhaps improve the circle scissors effect.
I am always frustrated when I need to segment a perfect circle, since I need to guess the center of the circle to pick it with the scissors.

If we change the following two lines https://github.com/Slicer/Slicer/blob/e5c5ad7801b81c99a0e37b44ce570e9acbfdb7f0/Modules/Loadable/Segmentations/EditorEffects/qSlicerSegmentEditorScissorsEffect.cxx#L418-L419

to 

```cpp
          position[0] = (eventPosition[0] + this->DragStartPosition[0])/2 + radius/2 * sin(angle);
          position[1] = (eventPosition[1] + this->DragStartPosition[1])/2 + radius/2 * cos(angle);
```

The circle center will be the midpoint between the start drag position and the current position. Effectively, the creation of the circle will be controlled by its borders. I find it easier to segment circles that way. Is the original way better for the medical imaging users? Drawing softwares such as inkscape uses shift to modify the behavior between those two options. Here is a quick vid showing how it would be. With the old option, I needed several tries to guess the center of the circle correctly. Appreciate your input on this.

https://user-images.githubusercontent.com/4790289/154724391-3d8c21bd-fc45-4bdd-83b9-fdaa35f287b5.mp4

lassoan:
For approximate segmentation the current "centered" approach usually works better, but it is certainly not good for circle fitting.

Adding a "centered" checkbox would be easy (it could be added as a fourth item to "Shape" list). It would make the GUI a tiny bit more complex, but it should be fine.

However, if you have very specific needs, such as segmenting a cylindrical core then you can do much better. You could ask the user to just place the centerpoint and use that to position a cylinder (you could accept numeric input for the diameter). The user could then conveniently adjust the position and would not be forced to get everything right with a single drag-and-drop. If you the user specifies two centerpoints (on two different slices) and then you can define an entire 3D cylindrical core with a few clicks. If this is a frequent operation then you can even go further and automate it so that the user would not need to click at all but the large inner circle would be extracted fully automatically.

lassoan:
The [Gel dosimetry](https://www.slicer.org/wiki/Documentation/Nightly/Modules/GelDosimetry) module works with a cylindrical ROI, you might get some useful ideas there.

lassoan:
[...]

PR #6204, opened 2022-02-18
ENH: added draw circle by border on scissors segmentation effect

fbordignon:
Enhancement based on #6203

### DIFF
--- a/Docs/developer_guide/modules/segmenteditor.md
+++ b/Docs/developer_guide/modules/segmenteditor.md
@@ -75,6 +75,7 @@ Common parameters must be set using `setCommonParameter` method (others can be s
 | ApplyToAllVisibleSegments | int   | no     | 0           | 0 or 1                                             |
 | Operation                 | enum  | no     | EraseInside | EraseInside, EraseOutside, FillInside, FillOutside |
 | Shape                     | enum  | no     | FreeForm    | FreeForm, Circle, Rectangle                        |
+| ShapeDrawCentered         | int   | no     | 0           | 0 (first click defines rectangle or circle boundary), 1 (click defines shape center) |
 | SliceCutMode              | enum  | no     | Unlimited   | Unlimited, Positive, Negative, Symmetric           |
 | SliceCutDepthMm           | float | no     | 0.0         | >=0.0 (single slice = 0.0)                         |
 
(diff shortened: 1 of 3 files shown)


lassoan (commented):
Thank you, this will be great, thank you! A few small things to improve:
- The option is both for circle and rectangle. Rename variable names and tooltips accordingly. (instead of "circle" we could use "region").
- Add the new parameter to Docs\developer_guide\modules\segmenteditor.md
- The new parameter's default value should be set so that default behavior is still centered. This should also be the case for existing scenes, where the new parameter is missing (and so probably a missing int parameter will be reported to have value = 0)

fbordignon:
> The option is both for circle and rectangle. Rename variable names and tooltips accordingly. (instead of "circle" we could use "region").
The default behavior of the rectangle is by the border.

Option one: Create two parameters, one for the rectangle and another one for the circle, so that the default for the rectangle is by the border and for the circle is centered.

Option two: Change the default behavior of the rectangle to be centered so that the checkbox has one parameter only (regionCenteredMode) which controls the behavior of the rectangle and the circle.

Does any of these seem reasonable?

lassoan:
The behavior is good, just need to rename variables, update tooltip, etc. to reflect this.

fbordignon:
Sorry, I did not understand.
The current behavior of the rectangle is to draw it using the borders and the current behavior of the circle is to draw it starting by the center.
The changes I've committed are modifying only the behavior of the circle controlled by the checkbox.
Are you proposing for the checkbox to affect both the circle and the rectangle?

If that so, the parameter CircleCenteredMode will need to be renamed to RegionCenteredMode with default value=0.
The checkbox circleCenteredCheckBox will be renamed to regionCenteredChecBox, which by default will be enabled for the circle and disabled for the rectangle? How can I accomplish that with only one parameter? Don't I need one parameter for the circle and one for the rectangle defaulting to the current behaviors respectively?

lassoan (approved):
I've fixed the review comments (easier than explain) and tested everything. All looks good, will merge now.

fbordignon:
Ok cool. In summary both behaviors defaults to drawing by the boundary now.
I've noticed a small issue with the checkbox. Issued a PR linked above.

Thanks

PR #6200, opened 2022-02-17
Default cleanups with scripts used to modernize ITK

hjmjohnson:
This applies a set of standard scripts that were created to update many ITK packages to make them consistent
with modern programming recommendations.

PR #6205, opened 2022-02-18
WIP:  Prefer exception throwing string conversions

hjmjohnson:
See comments in https://github.com/Slicer/Slicer/pull/6200#pullrequestreview-887706802 for further changes that would need to be made.

### DIFF
--- a/Base/CLI/Testing/itkTestMain.h
+++ b/Base/CLI/Testing/itkTestMain.h
@@ -127,7 +127,7 @@ int main(int ac, char *av[])
       {
       if( strcmp(av[1], "--with-threads") == 0 )
         {
-        int numThreads = atoi(av[2]);
+        int numThreads = std::stoi(av[2]);
         itk::MultiThreader::SetGlobalDefaultNumberOfThreads(numThreads);
         av += 2;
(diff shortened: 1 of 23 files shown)


lassoan (changes requested):
I'm quite shocked. std::stoi is much worse than I thought. It not just litters the code with exceptions, so that you need to add a wrapper that catches all the exceptions, but it also just catches certain kind of errors.

For example, if you call `std::stoi("123abc")` then you don't get an exception but you get `123` integer as output. I just don't know what people were thinking when they added this function.

I've reviewed usages of atoi and most of the time the optimal error handling behavior would be to set a default value and log a warning (and in some cases to log an error and return with an error).

I agree that it would be nice to add error checks into our string->number conversions. However, simply replacing atof/atoi by stod/stoi does not even come close to the ideal behavior and introduces new problems.

lassoan:
Most likely we'll replace these calls with something more sophisticated than `std::stoi` and `std::stod`. For example, `std::strtol` and `std::strtod` can be used to actually verify validity of the input string. So, this pull request will not be merged as is.

PR #6206, opened 2022-02-18
ENH: Skip empty colors with (none) label in color legend

MichaelColonel:
This PR prevents empty colors with "(none)" label from displaying in color legend. Part of the issue #6111.

"Hide empty Colors" check box controls the behavior.
Left: Empty colors are shown; Right: Empty colors are hidden.
image

### DIFF
--- a/Libs/MRML/Logic/Testing/Cxx/vtkMRMLColorLogicTest1.cxx
+++ b/Libs/MRML/Logic/Testing/Cxx/vtkMRMLColorLogicTest1.cxx
@@ -204,7 +204,8 @@ bool TestCopy()
   originalNode->NamesInitialisedOff();
   originalNode->SetNumberOfColors(6);
   originalNode->GetLookupTable()->SetTableRange(0, 5);
-  originalNode->SetColor(0, "background", 0.0, 0.0, 0.0, 0.0);
+  // Use NoName as color name to not list the "background" color in the color legend.
+  originalNode->SetColor(0, originalNode->GetNoName(), 0.0, 0.0, 0.0, 0.0);
   originalNode->SetColor(1, "one", 0.5, 1.0, 0.0, 0.1);
   originalNode->SetColor(2, "two", 0.5, 0.5, 0.0, 0.3);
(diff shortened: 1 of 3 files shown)


lassoan (commented):
Thank you. Very nice, you've implemented this in a smart, elegant way. This works well for label text = value.

What happens now when the label text = color name? In that case, each color should have the same size (the extra space that became available because of the removed empty colors should be distributed evenly among the visible colors).

MichaelColonel:
> What happens now when the label text = color name?

Do you mean that only unique colors must be shown, or the same several colors in a row must be compressed into one colour?

lassoan:
> > What happens now when the label text = color name?
> 
> Do you mean that only unique colors must be shown, or the same several colors in a row must be compressed into one colour?

Example: we have 6 named non-black colors, associated with values 1, 2, 3, 100, 200, 300; with all black between them. 

If label text = value: the colour legend should look the same as an image that has values between 1 and 300. Check if it works correctly for both skip and non-skip empty color modes. 

If label text = color: every color box should have the same size (2 should have the same size as 200).  If empty colors are skipped then there should be 6 nice big equal size boxes. If empty colours are not skipped then there should be 6 equal size thin non-black labeled slivers over black background.

MichaelColonel:
> If label text = value: the colour legend should look the same as an image that has values between 1 and 300. Check if it works correctly for both skip and non-skip empty color modes.

So the final LUT must have 300 colors with filled gaps between invalid colors?

lassoan:
Actually, I think we don't need the new `SkipEmptyColors` property in the color legend display node. In case of "label text = value" we should always show all colors (because if display an object using the colormap the appearance of the object does not change if labels are assigned to certain colors or not). In case "label text = color name" then we should always skip the empty colors.

MichaelColonel:
I've modified display manager instead of color node, and it should show color legend correctly when labels == names. It also shouldn't glitch with a few number of valid colors in the color node.

lassoan (approved):
This works great, thank you!

PR #6210, opened 2022-02-19
COMP: Remove ExpertAutomatedRegistration module

lassoan:
ExpertAutomatedRegistration module was added as a simple example for showing how ITK can be used in Slicer CLI modules.
Registration quality and configurability of this module is very limited compared to BRAINS, Elastix, and ANTs, which are all available in Slicer.
By removing it avoids user confusion (as the module is not recommended to be used) and reduces maintenance burden for developer.

It is a good time to remove this module now, as it is incompatible with latest ITK version that Slicer uses now.


jcfr:
Cc: @aylward @thewtex

jamesobutler (commented):
As a followup, you will want to update the documentation to move ExpertAutomatedRegistration from the "Legacy" category to the "Retired Modules" category with the final merged commit that is linked.

https://github.com/Slicer/Slicer/blame/e4b8a88c7200b71f96794766b6b0f5d62d54e79c/Docs/user_guide/modules/index.md#L175

lassoan:
Since the module causes build error and it was planned to be removed anyway, I commit this change.

@jamesobutler thanks for the suggestion, I've updated the module list as you suggested

jcfr:
And if need to be, corresponding code could be extracted and move into an independent repository.

PR #6212, opened 2022-02-20
DOC: Fix internal link in retired modules list

jamesobutler:
Found this broken link to the registration documentation page. Published readthedocs pages are html unless there is seemingly an index to pick the markdown name.


jamesobutler:
Appears https://www.commitcheck.com/ is down which is why the required check is still pending.

lassoan (approved):
Thanks for working on this.

The link was correct (something.md#anchor), but it seems that links cannot be used in tables in this way. I've converted the table to list, to avoid the problem.

I've also fixed a couple of other broken links (by checking the error messages in the documentation generation).

PR #6215, opened 2022-02-23
BUG: scissors shapeDrawCenteredCheckBox disabling when using rectangle

fbordignon:
Fixes a small issue introduced on #6204


lassoan (approved):
Thank you, looks good.

PR #6217, opened 2022-02-23
DOC: Run script in MacOS terminal

koegl:
Add instructions on how to run a script from a MacOS terminal
('fixes' issue mentioned in https://discourse.slicer.org/t/running-slicer-from-mac-terminal/18908)


pieper (approved):
Thanks 👍

Issue #5415, opened 2021-01-28
Force libarchive to not use homebrew libraries during packaging on macOS


piiq:
Packaged version of Slicer doesn't work on machines other than it was build if libarchive prerequisites were installed via homebrew.

When trying to run a freshly built and packaged Slicer on macOS the app fails to launch and an error pops up:

```
Library not loaded: /usr/local/opt/zstd/lib/libzstd.1.dylib
  Referenced from: Slicer.app/Contents/lib/Slicer-4.13/libarchive.17.dylib
```

The error indicates that during building `libarchive` has linked some of it's requirements to homebrew versions rather than local ones. For example `lz4` and `zstd` are installed by homebrew with `rsync` and `subversion` so if you have at least one of these two application installed via homebrew, `libarchive` will link these libraries to `/usr/local/opt/` and the packaged Slicer will not run on other machines.

## Steps to reproduce

- Install rsync or subversion via homebrew
- Build Qt
- Build Slicer
- Package Slicer
- Run the packaged application on a different mac

## Expected behavior

Slicer launches

## Environment

- Slicer version: Slicer-4.13.0-2021-01-26-macosx-amd64
- Operating system: Mac 11.1

lassoan:
There is already a check in place for Qt, it could be applied to LibArchive, too.

jcfr:
For reference, this approach could be generalized and improved:

https://github.com/Slicer/Slicer/blob/f1cf41897f6e930305f51c1fe37d9ed91e2bcf82/CMakeLists.txt#L644-L670

PR #6221, opened 2022-02-24
Simplify LibArchive external project and improve associated documentation

jcfr:
STYLE: Simplify LibArchive external project consolidating options
    
This commit removes obsolete comment from 2012 originally introduced
in c463d49680 (COMP: switch to newer LibArchive (3.0.4))

Option ENABLE_OPENSSL became redundant following 1f0f6f9b5 (COMP: Explicitly
set LibArchive options to avoid implicit finding of libraries)

Conditional passing of BUILD_TESTING was first introduced in
b50007add3 (COMP: On unix: use LibArchive 2.7.1 patched, on windows:
use LibArchive 2.8.4) and became irrelevant with c463d49680 (COMP:
switch to newer LibArchive (3.0.4))

-----

DOC: Update macOS build instructions removing obsolete LibArchive workaround

Considering that (1) the build instruction section "LibArchive pointing to
a nonexistent path" was added on Feb 2021 by 5dd8e7f71 (DOC: Update macOS
build instructions) and (2) the related issue #5415 was fixed in pull-request
#5854 from Sep 2021, this commit removes the now obsolete section.

Note that the list of LibArchive options explicitly disabled in the external
project has also been reviewed based on LibArchive update introduced in
e5c5ad7 (COMP: Update LibArchive from 3.4.3 to 3.5.2 to support GCC v11).

-----

DOC: Add note about hard-coded soversion to External_LibArchive.cmake

Since forgetting to update the soversion is a recurrent error (see [1]),
this commit attempt to explicit document was need to be done.

[1] A similar fix was integrated in Nov 2019 by 36c83c838 (COMP: Fix macOS
packaging to account for LibArchive version update).

### DIFF
--- a/Docs/developer_guide/build_instructions/macos.md
+++ b/Docs/developer_guide/build_instructions/macos.md
@@ -164,18 +164,6 @@ warning: target '@rpath/QtGui.framework/Versions/5/QtGui' does not exist...
 
 is present during packaging - doublecheck that Slicer was built with Qt that was built from source and not Qt that was installed from a web-installer or homebrew.
 
-#### LibArchive pointing to a nonexistent path
-
-If a packaged Slicer is launched on another mac and it crashes with the error log saying that
-
-```
-Library not loaded: /usr/local/opt/zstd/lib/libzstd.1.dylib
-  Referenced from: Slicer.app/Contents/lib/Slicer-4.13/libarchive.17.dylib
-```
-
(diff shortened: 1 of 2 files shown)


jcfr:
I was also considering adding this change, but I don't think it is worth introducing more variables and changes ...

```diff
diff --git a/SuperBuild/External_LibArchive.cmake b/SuperBuild/External_LibArchive.cmake
index a9b426cdbe..7783d3745d 100644
--- a/SuperBuild/External_LibArchive.cmake
+++ b/SuperBuild/External_LibArchive.cmake
@@ -96,8 +96,10 @@ if((NOT DEFINED LibArchive_INCLUDE_DIR
       ${${proj}_DEPENDENCIES}
     )
   if(APPLE)
+    set(_soversion 18)
+    set(_library ${EP_INSTALL_DIR}/lib/libarchive.${_soversion}.dylib)
     ExternalProject_Add_Step(${proj} fix_rpath
-      COMMAND install_name_tool -id ${EP_INSTALL_DIR}/lib/libarchive.18.dylib ${EP_INSTALL_DIR}/lib/libarchive.18.dylib
+      COMMAND install_name_tool -id ${_library} ${_library}
       DEPENDEES install
       )
   endif()
````

jamesobutler (approved):
Update instructions make sense

PR #6223, opened 2022-02-24
ENH: Remove unused New Fiducial List action

jamesobutler:
Looking back to Slicer 4.0.0 I haven't been able to find when it was last working. If anyone has insights I can add it to the commit message.

image


lassoan (approved):
Good catch, thank you.

Issue #6182, opened 2022-02-11
Interactive slice intersections are rendered incorrectly when slices are zoomed in


lassoan:
## Steps to reproduce

- Load a volume
- Enable interactive slice intersections
- Zoom in and pan down the red slice view so that the intersection with the green slice gets out of view
- Use Shift+mousemove in the green slice view
- ERROR: the yellow slice intersection is not a straight line anymore but two line segments point towards the middle of the slice view.

https://user-images.githubusercontent.com/307929/153658094-5bda6302-8c1a-447a-8df7-ac4bfc818994.mp4

## Expected behavior

A solid vertical yellow line (no gap) should move where the mouse pointer is moved.

## Environment
- Slicer version: Slicer-4.13.0-2022-02-01
- Operating system: Windows

lassoan:
@cpinter could you ask @dgmato to have a look at this?

dgmato:
Interesting. I will check it out.

lassoan:
We only need to interrupt the line at slice intersection points. If there is no intersection point then we should show a single uninterrupted line. The center of rotation and free translation handle position should could be in the center of the intersection line segment (center of the line section that is visible in the slice view).

dgmato:
@lassoan I think I was able to solve this issue by fixing the slice intersection point definition (i.e., rotation center) in those cases when no slice intersection is found. 

This is the current state:

https://user-images.githubusercontent.com/10816661/155157449-246af5ee-9482-4982-804d-5144181479c7.mp4

Let me know if this is the expected behaviour and I will create commit and push to my branch.

jamesobutler:
This can be closed now that https://github.com/Slicer/Slicer/commit/0092c502359dd44be99260511a4d8bcb136cd457 is in Slicer core?

lassoan:
Yes, good point, it can be closed.

Intersections are still rendered incorrectly if the slices are not intersect each other in the same point, but I'll add a separate issue for that.

PR #6225, opened 2022-02-25
BUG: Fix interactive slice intersections when slices are zoomed in

dgmato:
Fix [issue ](https://github.com/Slicer/Slicer/issues/6182) with rendering of interactive slice intersections when slices are zoomed in and the slice intersection point is not visible in slice view. Updated code is able to identify if the slice intersection point is contained in the slice view FOV and ensure correct rendering of intersection lines.

Re #6182


lassoan:
Thank you, merged.

PR #6226, opened 2022-02-25
DOC: Update build instructions to specify using Visual Studio 2022

jamesobutler:
The factory build machine has been using Visual Studio 2022 and the v143 toolset for the past month as of https://github.com/Slicer/DashboardScripts/commit/7cd961df84997f22bb684d7d2a185a610f23d6a1.

This PR is updating the documentation based on the comment at https://github.com/Slicer/Slicer/pull/6012#issuecomment-983050373 to update it once the factory machines were using VS2022 and the v143 toolset.

### DIFF
--- a/Docs/developer_guide/build_instructions/windows.md
+++ b/Docs/developer_guide/build_instructions/windows.md
@@ -7,7 +7,7 @@
   - Note: CMake must be able to find `git.exe` and `patch.exe`. If git is installed in the default location then they may be found there, but if they are not found then either add the folder that contains them to `PATH` environment variable; or set `GIT_EXECUTABLE` and `Patch_EXECUTABLE` as environment variables or as CMake variables at configure time.
 - [Visual Studio](https://visualstudio.microsoft.com/downloads/)
   - any edition can be used (including the free Community edition)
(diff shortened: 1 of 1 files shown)


lassoan:
I switched to VS2022 to test if everything works well and I find starting of debugging and stopping at the first breakpoint painfully slow (tens of seconds instead of few seconds). Have you experience this regression, too?

jamesobutler:
I have not experienced this as I have not tried yet. I typically build my Slicer in release for production usage.

jcfr:
> find starting of debugging and stopping at the first breakpoint painfully slow (tens of seconds instead of few seconds).

Look like this has been addressed with the most recent version of VS 2022. See https://developercommunity.visualstudio.com/t/debugging-is-very-slow-with-visual-studio-2022/1577264

| Version | Status |
|--|--|
| Visual Studio 2022 version 17.0 | :x:  |
| Visual Studio 2022 version 17.0.1 | :heavy_check_mark:  |


@lassoan , if your version is indeed `17.0`, I suggest you try updating and confirm it works as expected.

Then, we could add a note to the documentation to indicate that version `17.0.1` should be used for best debugging performances.

jamesobutler:
@lassoan Have you tried latest Visual Studio 2022 with the most recent bug fixes?

lassoan:
I've been using VS2022 17.1.0 and did a full clean build and debugging is still extremely slow (about 10x slower than with VS2019). It is not a blocking issue, though.

lassoan (approved):
This looks perfect. Factory build switched to VS2022, this commit updates the build instructions accordingly.

PR #6227, opened 2022-02-25
BUG: Fix export directory line edit focus

ebrahimebrahim:
It was not possible to focus on the directory ctkPathLineEdit via mouse click.


lassoan (approved):
The change looks good, thank you!

Issue #6228, opened 2022-02-27
VTK import fails using PythonSlicer from a Slicer build tree on Windows


lassoan:
`import vtk` fails in PythonSlicer.exe interpreter on Windows:

```
No module named 'logic'
DLL load failed while importing qSlicerBaseQTCLIPython: The specified module could not be found.
Traceback (most recent call last):
  File "D:\D\P\Slicer-0\Applications\SlicerApp\Testing\Python\StandalonePythonImportSlicerTest.py", line 2, in <module>
    import vtk
  File "D:\D\P\Slicer-0-build\VTK-build\bin\Lib\site-packages\vtk.py", line 31, in <module>
    all_m = importlib.import_module('vtkmodules.all')
  File "D:\D\P\Slicer-0-build\python-install\Lib\importlib\__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "D:\D\P\Slicer-0-build\VTK-build\bin\Lib\site-packages\vtkmodules\all.py", line 30, in <module>
    from .vtkRenderingQt import *
ImportError: DLL load failed while importing vtkRenderingQt: The specified module could not be found.
```

This causes `py_standalonepython_import_slicer` test to fail - see https://slicer.cdash.org/test/21535427

## Steps to reproduce

<s>Start `PythonSlicer.exe` in the install tree and type `import vtk`.</s>

Start `PythonSlicer.exe` in the `python-install` directory in the Slicer build tree and type `import vtk`.

## Expected behavior

VTK should be imported successfully.

## Workaround

Adding these lines to c:\D\S4R\python-install\bin\PythonSlicerLauncherSettings.ini fixes the issue:

```
[LibraryPaths]
1\path=C:/Qt/5.15.0/msvc2019_64/bin
size=1
```

## Environment
- Slicer version: Slicer-4.13.0-2022-02-20
- Operating system: Windows

jcfr:
Thanks for the report. I will have a look this evening

PR #6229, opened 2022-02-28
BUG: Fix py_standalonepython_import_slicer test on windows

jcfr:
This commit fixes the test and ensures that python modules compiled
on windows and depending on Qt (e.g `vtkRenderingQt`) can be imported
using PythonSlicer when executed from a Slicer build tree.

See #6228

It sets `QT_LIBRARY_DIR` variable and updates the `LibraryPaths`
launcher settings to append `QT_LIBRARY_DIR` on Unix and `QT_BINARY_DIR`
on Windows.

Running these commands on windows now work as expected:

```
  cd C:\D\S4R
  python-install\bin\PythonSlicer.exe -c "import slicer"

  cd C:\D\S4R\Slicer-build
  ctest.exe -C Release -R py_standalonepython_import_slicer
```

### Background

When configured against Qt4, the variables `QT_PLUGINS_DIR`, `QT_BINARY_DIR`
and `QT_LIBRARY_DIR` were set by the `FindQt4` CMake module.

Considering these variables are not set by the Qt5 modules, following
commit e88cb6649 (ENH: Add preliminary support for Qt5), the
variables `QT_PLUGINS_DIR` and `QT_BINARY_DIR` were explicitly set.

Until the introduction of the `slicer_dll_directories` module in
commit 34e48e8ae (ENH: Upgrade from python 3.6.7 to 3.9.10), there were
no issues with either

1. the missing `QT_BINARY_DIR` from the `LibraryPaths` launcher settings

2. the empty `QT_LIBRARY_DIR` variable

because:

* for build tree:
  - on Linux & macOS, library paths are resolved relying on rpath
    hard-coded in libraries and executables
  - on Windows, the `Paths` launcher settings was already properly set
    with `QT_BINARY_DIR`

* for install tree:
  - Qt libraries were installed in common location already associated
    with `LibraryPaths` and `Paths`

### DIFF
--- a/CMake/SlicerBlockCTKAppLauncherSettings.cmake
+++ b/CMake/SlicerBlockCTKAppLauncherSettings.cmake
@@ -71,7 +71,11 @@ set(SLICER_LIBRARY_PATHS_BUILD
   )
 
 if(NOT Slicer_USE_SYSTEM_QT)
-  list(APPEND SLICER_LIBRARY_PATHS_BUILD ${QT_LIBRARY_DIR})
+  if(WIN32)
+    list(APPEND SLICER_LIBRARY_PATHS_BUILD ${QT_BINARY_DIR})
+  else()
+    list(APPEND SLICER_LIBRARY_PATHS_BUILD ${QT_LIBRARY_DIR})
+  endif()
 endif()
 
 # The following lines allow Slicer to load a CLI module extension that depends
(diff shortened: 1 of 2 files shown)


lassoan:
Great, thanks for the fix!

> on Linux & macOS, library paths are resolved relying on rpath
hard-coded in libraries and executables

Does this mean that on linux we don't rely on the launcher anymore to set up ld paths?

jcfr:
> > on Linux & macOS, library paths are resolved relying on rpath
> > hard-coded in libraries and executables

> Does this mean that on linux we don't rely on the launcher anymore to set up ld paths?

For build tree:

* Strictly speaking, a launcher is not needed when building on linux or macOS (unless rpath is explicitly disabled). Now, we still need one to setup env. variables required for general configuration and module discovery. 

For install tree:

* we always need a launcher for Linux and Windows
* And for macOS, we builtin the CTKAppLauncherLib to read settings from a settings file and avoid specific case ...

PR #6230, opened 2022-02-28
ENH: Added Object names to option group box components.

dwijmistry11:


dwijmistry11:
@lassoan @jcfr please have a look on the changes

when we are trying to access Segment Editor module's -> option frame components -> specific to active Effect. it prints blank layout without any elements in it.

Please guide me if I missed or messed something.

dwijmistry11:
I will submit in new branch

PR #6234, opened 2022-03-01
BUG: Workaround for Slicer hang on scipy import on Windows 11

jamesobutler:
This is related to https://github.com/Slicer/Slicer/commit/542f53c5ba9bf650119707f990e36d74a6c8d36a.

I attempted `import scipy` in the latest Slicer preview 4.13.0-2022-02-28 and the Slicer application would then just hang. This is on my Windows 11 machine so I attempted the same type workaround as https://github.com/Slicer/Slicer/commit/542f53c5ba9bf650119707f990e36d74a6c8d36a and it no longer would cause the application to hang.

An unfortunate side-effect of this workaround is that there is no auto-complete for available methods. `numpy` has nothing show up on `numpy.[tab]` in the python interactor and `scipy` only has `scipy.test()` as an option. If I manually type out some known thing like `scipy.__version__` then it does appropriately print out:
```
>>> scipy.__version__
'1.7.3'
```

### DIFF
--- a/Base/Python/slicer/__init__.py
+++ b/Base/Python/slicer/__init__.py
@@ -52,16 +52,18 @@ for kit in available_kits:
   del kit
 
 #-----------------------------------------------------------------------------
-# Import numpy early, as a workaround for application startup hang on Windows11
-# due to output redirection (only needed for embedded Python, not for standalone).
+# Import numpy and scipy early, as a workaround for application hang in import
+# of numpy or scipy at application startup on Windows 11 due to output redirection
+# (only needed for embedded Python, not for standalone).
 # See details in https://github.com/Slicer/Slicer/issues/5945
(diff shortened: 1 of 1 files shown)


jamesobutler:
> An unfortunate side-effect of this workaround is that there is no auto-complete for available methods. `numpy` has nothing show up on `numpy.[tab]` in the python interactor

I don't appear to be seeing this anymore when testing with latest nightly. Not sure why I was experiencing this.

lassoan (commented):
Thank you. Since scipy is bundled with Slicer, I agree that this workaround need to be extended to scipy, too. Could you add scipy to the comment above the change, too?

jamesobutler:
The comment above the code has been updated with the mention of scipy as well.

PR #6236, opened 2022-03-02
ENH: Replacing lzma v5.2.2 by xz v5.2.5

RafaelPalomar:
This replaces the lzma repository by xz which includes lzma and it is
better maintained.


RafaelPalomar:
@jcfr what do you think about this change?

jcfr:

PR #6237, opened 2022-03-02
COMP: Require at least C++17

jamesobutler:
Qt 6 has a C++ minimum version that is now C++17. https://www.qt.io/blog/qt6-development-hosts-and-targets

This commit bumping to C++17 was originally included in #6108 with other supported platform minimum version bumps. This was moved into its own independent PR based on feedback to separate this decision from the other platform version changes needed for Qt 6 and for other reasons as well.

From my initial testing when #6108 was first submitted Slicer I had the following report:
- ✅C++17 built on Windows 10 using Visual Studio 2019 with v142 toolset completed successfully with this PR. All tests cases were run and passed successfully where the failing tests were the same as the already failing tests.


jamesobutler:
I have re-tested with the latest version of this branch which includes all the most recent VTK and ITK upgrades.
- ✅C++17 built on Windows 11 using Visual Studio 2022 with v143 toolset completed successfully with this PR. All tests cases were run and passed successfully where the failing tests were the same as the already failing tests.
```
1>The following tests FAILED:
1>	  7 - py_cmake_slicer_extensions_index_build_without_upload (Failed)
1>	  8 - py_cmake_slicer_extensions_index_build_with_upload (Failed)
1>	  9 - py_cmake_slicer_extensions_index_build_with_upload_using_ctest (Failed)
1>	 10 - py_cmake_slicer_extensions_index_build_without_upload_using_ctest (Failed)
1>	267 - qSlicerExtensionsManagerModelTest (Failed)
1>	331 - vtkMRMLAnnotationLinesNodeTest1 (Failed)
1>	333 - vtkMRMLAnnotationLinesStorageNodeTest1 (Failed)
```

lassoan:
There are severe Qt5 bugs that are only fixed in Qt6, so we need to switch soon (maybe by the end of this year?). Updating the build requirements seems like a good preparation for that (to give time for developers to find and fix issues related to requiring Cxx17).

jamesobutler:
The proposed change here as a hard check is to provide enough time for developers to have made the switch for the next Slicer stable feature release. I had originally issued the PR for the C++11 to C++14 switch in Oct 2021 which was a bit of time until May 2022 when Slicer 5 was released. If C++14 is kept around I would plan to make it not the default where Slicer factory preview builds would still plan to use C++17 while Slicer Stable would continue to use C++14. This seems to be an appropriate split where new dev happens on the preview side with periods of instability while the stable can continue to be used and extended through extensions.

> It's an overhead of at least a full day, maybe more if you work on more projects.

If we maintain C++14 as a non-default option, you could schedule your day to a convenient time. Would that work for you? Or if you can bear a day or two of fixing up that would be helpful to avoid extra Slicer maintenance time for trying to maintain various build options rather than just the single latest version of the tools. Having one well working configuration rather than multiple configurations that all have their own issues.

cpinter:
> Or if you can bear a day or two of fixing up that would be helpful to avoid extra Slicer maintenance time for trying to maintain various build options rather than just the single latest version of the tools. 

I agree that supporting an older C++ version gives extra maintenance work.

I think that if we announce it clearly to developers that the switch will happen, and give a week or two, then it is fair to expect that during that time the devs who pay attention to news find a day to make the switch. What I wanted to avoid is surprises that may cause me or other people significant delay possibly at a critical time.

So in summary if we make this change well visible to the developers and set a day for its integration then I think that's good. Thanks for the considerations.

jamesobutler:
Post has been made at https://discourse.slicer.org/t/transition-to-c-17-as-the-minimum-standard/23951

jamesobutler:
To provide an update, building on base of https://github.com/Slicer/Slicer/commit/c5689b81efe3e6f8cef407b89fa0822ce3ee1dce, there were no build errors and failed tests were the same tests as previously were failing on the factory machines. The following configuration tested building with C++17 was:
- Windows 11 Pro (22000.739)
- CMake 3.23.2
- Visual Studio 2022 (17.2.5)
- v143 toolset

If someone can approve this PR, then it will make this ready for merging on Monday July 4th as stated in the discourse post.

jamesobutler:
@cpinter @lassoan Could one of you all approve this PR so it is ready for merging tomorrow (July 4th) per the corresponding discourse post about it?

pieper (approved):
Developers may need to recompile, but everyone's been notified and if they are under a deadline or something they can revert this change so it's fine to go ahead IMO.

PR #6238, opened 2022-03-03
DOC: Update export docs with how to set DICOM tags

koegl:
The script explaining how to export a volume to DICOM now includes an explanation on how to set DICOM tags


pieper (approved):
The extra documentation is great, thanks for taking the time to contribute this 👍 

I just tried to clarify the wording a bit.

Issue #6240, opened 2022-03-04
Interpolation is reset on volume when showing other volume in slice view


cpinter:
When a new volume is shown in a slice view, then the old one is set the interpolation flag of the new one, thus resetting it.

## Steps to reproduce

- Load MRHead
- Load CTChest
- Link slice views (maybe not needed)
- Turn off interpolation of background (CTChest)
- Show MRHead in background

Interpolation is turned on in display node of CTChest

## Expected behavior

Interpolation flag stays off for CTChest

## Environment
- Slicer version: Slicer-4.13.0-2022-02-02
- Operating system: Windows 10 and Linux Ubuntu 20.04

PR #6241, opened 2022-03-04
BUG: Fix interpolation reset issue in slice views

cpinter:
When the interpolation was different for the previous and new (background or foreground) volume in the slice views when changing, the previous volume display node was set the interpolation value of the new volume display node. It was because of a signal that was emitted when setting the interpolation on the checkbox in the slice view controller, which triggered setting that interpolation value. By adding blockSignals calls this issue is fixed.

Fixes https://github.com/Slicer/Slicer/issues/6240


cpinter:
I've rebased the branch to the latest revision. It is building now, will amend the PR if all is good.

lassoan (approved):
Thank you, looks good to me.

PR #6242, opened 2022-03-05
Fix dicom uid tags

koegl:
Three DICOM UID tags were corrected.

Solves issue mentioned in [DICOM export failed - UID tags](https://discourse.slicer.org/t/dicom-export-failed-uid-tags/22317)[](https://discourse.slicer.org/c/support/11)


lassoan (changes requested):
The correct name of field (0020,0052) is "Frame of Reference UID". So, the changes suggested in this pull request are not needed.

In earlier Slicer versions the field name was misspelled as "...InstanceUID". If you find any occurrences of the incorrect name in the current Slicer source code then let us know.

lassoan:
Have you submitted this pull request to fix some issue? Maybe you tried to use an old script with the current Slicer version? If there was no specific issue that this pull request fixes then I would close it.

PR #6243, opened 2022-03-07
BUG: Fix broken multiple inheritance in ScriptedLoadableModule classes

allemangD:
Missing calls to `super().__init__()` in the `ScriptedLoadableModule` classes break multiple inheritance. The init method is skipped in any derived classes that come after the ScriptedLoadableModule class in the MRO.

Adding super calls to each of the ScriptedLoadableModule classes fixes this issue and correctly supports multiple inheritance.

- `ScriptedLoadableModule.__init__`
- `ScriptedLoadableModuleWidget.__init__`
- `ScriptedLoadableModuleLogic.__init__`

---

For example, this class fails to initialize VTKObservationMixin:

```python
class MyModuleLogic(ScriptedLoadableModuleLogic, VTKObservationMixin):
  def __init__(self):
    super().__init__()
    
    self.addObserver(...)  # AttributeError on self.Observations
```

_In this case_, swapping the base class order avoids the error as `VTKObservationMixin` _does_ include a super call, however that solution is not perfect. `MyModuleLogic.__base__` is not `ScriptedLoadableModuleLogic` as it should be, and it is impossible to pass the optional `parent` parameter to `ScriptedLoadableModuleLogic.__init__`.

---

This change should not have any backwards compatibility issues, as currently the only way to support mixins is to explicitly call the mixin initializer, in which case the behavior of `super()` is ignored.

---

`ScriptedLoadableModuleTest` cannot be fixed because its base class, `unittest.TestCase`, [does not have a super call](https://github.com/python/cpython/blob/3.10/Lib/unittest/case.py#L357-L387). For such cases it would still be necessary to explicitly call the mixin initializer.

### DIFF
--- a/Base/Python/slicer/ScriptedLoadableModule.py
+++ b/Base/Python/slicer/ScriptedLoadableModule.py
@@ -8,6 +8,7 @@ __all__ = ['ScriptedLoadableModule', 'ScriptedLoadableModuleWidget', 'ScriptedLo
 
 class ScriptedLoadableModule:
   def __init__(self, parent):
+    super().__init__()
     self.parent = parent
     self.moduleName = self.__class__.__name__
 
@@ -76,6 +77,7 @@ class ScriptedLoadableModuleWidget:
     """If parent widget is not specified: a top-level widget is created automatically;
     the application has to delete this widget (by calling widget.parent.deleteLater() to avoid memory leaks.
     """
+    super().__init__()
     # Get module name by stripping 'Widget' from the class name
     self.moduleName = self.__class__.__name__
     if self.moduleName.endswith('Widget'):
@@ -211,6 +213,7 @@ class ScriptedLoadableModuleWidget:
 
 class ScriptedLoadableModuleLogic:
   def __init__(self, parent = None):
+    super().__init__()
     # Get module name by stripping 'Logic' from the class name
     self.moduleName = self.__class__.__name__
     if self.moduleName.endswith('Logic'):


jcfr:
Thanks for the detailed explanation :pray: 

Following b2ba48205ae21f1425ad709fcf6f63e46cbbefe1 (ENH: Add support for building Python 3.6) removing support for python 2.7, only new style python classes are supported and using `super()` makes sense.

For references:
* https://docs.python.org/3/library/functions.html#super
* https://www.python.org/doc/newstyle/
* https://docs.python.org/3/tutorial/classes.html
* https://wiki.python.org/moin/NewClassVsClassicClass?action=recall&rev=20

allemangD:
> ScriptedLoadableModuleTest cannot be fixed because its base class, unittest.TestCase, [does not have a super call](https://github.com/python/cpython/blob/3.10/Lib/unittest/case.py#L357-L387). For such cases it would still be necessary to explicitly call the mixin initializer.

> https://docs.python.org/3/library/functions.html#super

Reading through that documentation more carefully, it seems it _is_ possible to fix the behavior for `ScriptedLoadableModuleTest` as well. @jcfr thanks for posting the link! It helps to read things even if you already know the bulk of it, I learned something new!

> The object-or-type determines the [method resolution order](https://docs.python.org/3/glossary.html#term-method-resolution-order) to be searched. The search starts from the class right after the type.

This indicates that `super(UnitTest, self)` would take the next base type of `self` _after `UnitTest`_. Below is a minimal example of the class hierarchy:

```python
class Mixin:
    def __init__(self):
        super().__init__()
        
        print('Mixin')


class UnitTest:
    def __init__(self):
        # super call cannot be added.
        
        print('UnitTest')


class ModuleTest(UnitTest):
    def __init__(self):
        # next class in MRO after UnitTest (Mixin.__init__)
        super(UnitTest, self).__init__() 
        
        # immediate next class in MRO (UnitTest.__init__) 
        super(ModuleTest, self).__init__()

        print('ModuleTest')


class MyModuleTest(ModuleTest, Mixin):
    def __init__(self):
        super().__init__()

        print('MyModuleTest')

_ = MyModuleTest()  # Mixin / UnitTest / ModuleTest / MyModuleTest
```

There is a risk in this implementation: if a super call is ever added to the `UnitTest` class, then `Mixin.__init__` (and `object.__init__`) would be invoked twice.

I will add commits to this branch introducing the above change, along with a test to check for regression, ensuring that `Mixin.__init__` is only ever called once.

PR #6246, opened 2022-03-07
COMP: Rename xz external project back to LZMA

jcfr:
This commit partially reverts d4e7c2fcc (ENH: Replacing lzma v5.2.2 by
xz v5.2.5) to use the name `LZMA` instead of `xz`.

This ensures the project can be configured specifying external built
LZMA library by specifying `LZMA_INCLUDE_DIR` and `LZMA_LIBRARY` variables.

### DIFF
--- a/SuperBuild/External_LZMA.cmake
+++ b/SuperBuild/External_LZMA.cmake
@@ -1,5 +1,5 @@
 
-set(proj xz)
+set(proj LZMA)
 
 # Set dependency list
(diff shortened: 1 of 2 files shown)


RafaelPalomar (approved):
This looks good to me. It makes sense to keep the LZMA naming instead of xz (xz is the project/application that uses the library). Thanks for having a look at it!

PR #6248, opened 2022-03-08
BUG: Fix build with Slicer_USE_PYTHONQT set to OFF

markasselin:


jcfr:
Thanks for the fix :pray:

PR #6249, opened 2022-03-08
BUG: Fix broken multiple inheritance in ScriptedLoadableModuleTest 

allemangD:
See my comment here: https://github.com/Slicer/Slicer/pull/6243#issuecomment-1061800718

> > ScriptedLoadableModuleTest cannot be fixed because its base class, unittest.TestCase, [does not have a super call](https://github.com/python/cpython/blob/3.10/Lib/unittest/case.py#L357-L387). For such cases it would still be necessary to explicitly call the mixin initializer.
> 
> > https://docs.python.org/3/library/functions.html#super
> 
> Reading through that documentation more carefully, it seems it _is_ possible to fix the behavior for `ScriptedLoadableModuleTest` as well. @jcfr thanks for posting the link! It helps to read things even if you already know the bulk of it, I learned something new!
> 
> > The object-or-type determines the [method resolution order](https://docs.python.org/3/glossary.html#term-method-resolution-order) to be searched. The search starts from the class right after the type.
> 
> This indicates that `super(UnitTest, self)` would take the next base type of `self` _after `UnitTest`_.
> 
> ***[...]***
> 
> There is a risk in this implementation: if a super call is ever added to the `UnitTest` class, then `Mixin.__init__` (and `object.__init__`) would be invoked twice.

I have introduced the above change (c582589a3a75be4028477c681ca0c0dd83049ae5) and added tests corresponding with this and #6243 (44779801c54528a572e9baa217432a1844cfba67).

### DIFF
--- a/Base/Python/slicer/ScriptedLoadableModule.py
+++ b/Base/Python/slicer/ScriptedLoadableModule.py
@@ -299,7 +299,12 @@ class ScriptedLoadableModuleTest(unittest.TestCase):
   """
 
   def __init__(self, *args, **kwargs):
-    super().__init__(*args, **kwargs)
+    super(ScriptedLoadableModuleTest, self).__init__(*args, **kwargs)
+
+    # See https://github.com/Slicer/Slicer/pull/6243#issuecomment-1061800718 for more information.
+    # Do not pass *args, **kwargs since there is no base class after `unittest.TestCase`. This is only relevant to
+    # mixins of derived classes, which should each have the same signature as `object()`.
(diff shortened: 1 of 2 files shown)


allemangD:
Rebased on new master since it's been a couple months since creating the PR. Resolved some conflicts in `test_slicer_util_VTKObservationMixin.py`.

lassoan (approved):

jcfr:
Thanks @allemangD  :100: 

Locally tested :heavy_check_mark:

PR #6254, opened 2022-03-08
BUG: Fix attribute bad access when using a sequence node

xriobe:
In DICOMScalarVolumePluginClass::sliceCornersFromDICOM, the DICOM.instanceUIDs attribute will be empty in this case.

### DIFF
--- a/Base/QTCore/qSlicerCoreApplication.cxx
+++ b/Base/QTCore/qSlicerCoreApplication.cxx
@@ -132,6 +132,7 @@
 //-----------------------------------------------------------------------------
 // Helper function
 
+#ifdef Slicer_USE_PYTHONQT
 namespace
 {
 wchar_t* QStringToPythonWCharPointer(QString str)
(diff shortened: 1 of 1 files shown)


pieper (changes requested):
Thanks for the contribution 👍 

Can you add a bug report that describes what situation this fixes?  Then reference that bug report here so that we can close the bug with the PR merge.

It seems to me that the only way the `DICOM.instanceUIDs` attribute would be missing would be if this code were used from outside the DICOM module - is that how this came up?

xriobe:
Yes this occurs when using the method independently, that's why I did not enter a bug for Slicer since it may not be reproducible, but this modification improves robustness.

pieper:
Thanks for the extra information.  Since these methods are part of the public API even if they were originally intended for internal use it does make sense to make them robust.   Since the change is small and clear I guess we don't need a bug report since developers could always find this PR if context is needed.

xriobe:
So i guess there is no more change requested? The fail of the API documentation should not be related to this.

xriobe:
oups, bad manipulation in my fork, see https://github.com/Slicer/Slicer/pull/6259

PR #6256, opened 2022-03-09
BUG: Remove redundant surface constraint from markups resampling

smrolfe:
This pull request removes the surface constraint options from the Markups Resample menu. The surface constraint function has been moved to the Curve Settings menu and was no longer functional from the Resample menu. The issue was discussed in [this thread](https://discourse.slicer.org/t/unexpected-behavior-of-curve-resampling/22188/2).


lassoan (approved):
Thank you, it looks good. I've also removed the now unused resampling with projection method (as it did not work correctly anyway).

PR #6257, opened 2022-03-09
ENH:Make vtkMRMLMarkupsFiducialStorageNode generic

RafaelPalomar:
This allows `vtkMRMLMarkupsFiducialStorageNode` to be used by markups
defined in external modules (pluggable markups)


lassoan (approved):
This looks fine. Note that .fcsv file format is extremely limited, it only saves a small fraction of information of what is stored in a markups node, so it should not be used in any new projects (it is for backward compatibility only).

PR #6258, opened 2022-03-09
ENH: Improve robustness check in vtkSlicerSegmentationGeometryLogic

xriobe:
It will prevent a crash in CalculateOutputGeometryFromBounds if the source geometry is not set.


cpinter:
Looks good to me. Thanks @xriobe !

PR #6263, opened 2022-03-10
DOC: Add explanation of exporting markups

koegl:
Added two lines of code showing how to export/save markups to an .fcsv file

I thought it would be useful to add this to the script repository - it is already explained how an .fcsv file can be loaded but not how it can be created

### DIFF
--- a/Docs/developer_guide/script_repository/markups.md
+++ b/Docs/developer_guide/script_repository/markups.md
@@ -8,6 +8,12 @@ Markups point list can be loaded from file:
 slicer.util.loadMarkupsFiducialList("/path/to/list/F.fcsv")
 ```
 
+### Save markups to .fcsv file
+```python
+markups_node = slicer.mrmlScene.GetNodeByID("vtkMRMLMarkupsFiducialNode1")
+slicer.util.saveNode(markups_node, "/path/to/markpus/file.fcsv")
+```
+
 ### Adding control points Programmatically
 
 Markups control points can be added to the currently active point list from the python console by using the following module logic command:


lassoan (commented):
Thanks for the suggestion. I really appreciate your efforts for improving Slicer and its documentation.

I'm a bit hesitant to add this example, as the usage of the old .fcsv file format is strongly discouraged for several reasons:
- it is not a standard file format: it cannot be easily parsed due to its multiline header
- it can only store a small fraction of the information that is in a markup node, so it should not be used for saving data, only for exporting
- you can only load an .fcsv file as a point list (none of the other markup types)

**What do you think about adding an example for exporting as a standard CSV file?**

We can do something like this now:

```python
markupsNode = getNode('F')
outputFile = "c:/tmp/points.csv"

tableNode = slicer.mrmlScene.AddNewNodeByClass('vtkMRMLTableNode'))
slicer.modules.markups.logic().ExportControlPointsToTable(markupsNode, tableNode, slicer.vtkMRMLStorageNode.CoordinateSystemLPS)
slicer.util.saveNode(tn, outputFile)
slicer.mrmlScene.RemoveNode(tableNode)
```

But it could be simplified to something like this:

```python
markupsNode = getNode('F')
slicer.modules.markups.logic().ExportControlPointsToCSV(markupsNode, "c:/tmp/points.csv")
```

Could you describe how do you plan to use the exported csv file?
What programming application/environment/programming language do you plan to use this file in?

koegl:
I was just looking for a good way to export landmarks and this was the first that I stumbled upon, so .csv should be good too.

I think the example that you gave makes sense - should we add this example to the DOCs and in the mean time I will try to implement `ExportControlPointsToCSV()` in another pull request?

I'm preparing data for a publication and I can export all my volumes and segmentations to DICOM, but not my landmarks, so I thought I would export them (for example as a .csv) to accompany the DICOMs. So in the end I won't be the only one who will use them, I just want to find a simple format which mostly just stores the coordinates

lassoan:
I've added [methods for easy CSV import/export of control point coordinates](https://github.com/Slicer/Slicer/blob/2d77c6dc644312486a6080b1594cc33c30da1927/Modules/Loadable/Markups/Logic/vtkSlicerMarkupsLogic.h#L296-L305). Examples:

```
markupsNode = getNode('F')
slicer.modules.markups.logic().ExportControlPointsToCSV(markupsNode, "c:/tmp/points.csv")

markupsNode = slicer.mrmlScene.AddNewNodeByClass("vtkMRMLMarkupsCurveNode")
slicer.modules.markups.logic().ImportControlPointsFromCSV(markupsNode, "c:/tmp/points.csv")
```

Still, I would recommend using json file format in general - see a minimal example here: https://slicer.readthedocs.io/en/latest/developer_guide/modules/markups.html#markups-json-file-format-mrk-json

```
{"@schema": "https://raw.githubusercontent.com/slicer/slicer/master/Modules/Loadable/Markups/Resources/Schema/markups-schema-v1.0.0.json#",
"markups": [{"type": "Fiducial", "coordinateSystem": "LPS", "controlPoints": [
    { "label": "F-1", "position": [-53.388409961685827, -73.33572796934868, 0.0] },
    { "label": "F-2", "position": [49.8682950191571, -88.58955938697324, 0.0] },
    { "label": "F-3", "position": [-25.22749042145594, 59.255268199233729, 0.0] }
]}]}
```

Json is more versatile (you can store multiple tables, hierarchies, etc.), more explicit (you add "coordinateSystem" instead of infer from column names), and more future-proof (you can easily save more data in the json file while preserving backward compatibility).

I've added examples of these more preferable formats here: https://slicer.readthedocs.io/en/latest/developer_guide/script_repository.html#markups

With these, I think the original intent of the pull request is fulfilled, so I'm closing it now.

PR #6264, opened 2022-03-11
BUG: Fix Markups curve picking

Sunderlandkyl:
The color legend renderer was erasing the Z-buffer, making it impossible to interact with markups curves outside of the view plane.
Fixed by setting setting disabling the Erase flag on the color legend renderer.


lassoan (approved):
Awesome!!! I was experiencing mysterious picking issues and I'm really glad you solved it. I would not have thought that the color legend renderer could be the culprit.

PR #6265, opened 2022-03-14
ENH: Move "Application settings" action to the top of Edit menu

lassoan:
Edit menu's Cut/Copy/Paste actions are not useful (with a few exceptions, copy-paste is only supported in textboxes where keyboard shortcuts and right-click menu are more convenient to use).
It is fine to have them there to make menu sizes a bit more balanced.

In contrast, "Application Settings" action is used often and it is much harder to click at the bottom of the menu than at the top.

Therefore, this commit moves "Application Settings" action to the top of the Edit menu.

Before this commit:

image

After this commit:

image

### DIFF
--- a/Base/QTApp/Resources/UI/qSlicerMainWindow.ui
+++ b/Base/QTApp/Resources/UI/qSlicerMainWindow.ui
@@ -234,14 +234,14 @@
     <property name="title">
      <string>&amp;Edit</string>
     </property>
+    <addaction name="EditApplicationSettingsAction"/>
+    <addaction name="separator"/>
     <addaction name="CutAction"/>
     <addaction name="CopyAction"/>
     <addaction name="PasteAction"/>
     <addaction name="separator"/>
     <addaction name="EditRecordMacroAction"/>
     <addaction name="EditPlayMacroAction"/>
-    <addaction name="separator"/>
-    <addaction name="EditApplicationSettingsAction"/>
    </widget>
    <widget class="QMenu" name="ViewMenu">
     <property name="title">


jamesobutler (commented):
The reason that is was moved lower was because most applications have Edit/Copy/Paste at the top with the settings or preferences as the last entry. It is unexpected for the application settings to be the first entry instead of the last entry.

jamesobutler:
Just a selection of some applications where there is a typical given order for the Edit menu and that is with Undo/Redo and Cut/Copy/Paste at the top with other options below it.
image
image
image

lassoan:
I know that normally application settings is at the bottom of the Edit menu. That's why I tried to explain why not to follow this convention in Slicer. To summarize: The menu items are ordered approximately in frequency of use. Since Cut/Copy/Paste items are just placeholders (not practically useful in Slicer) they should be below "Application settings".

lassoan:
I liked the consistency with other software but the experiencing the extra difficulty on using the menu (it takes at least 2x longer time to click on the last item vs. the first item) changed my opinion on this.

jcfr:
> I liked the consistency with other software but the experiencing the extra difficulty on using the menu (it takes at least 2x longer time to click on the last item vs. the first item) changed my opinion on this.

Out of curiosity, given the time that will be spent in the setting panel, is it worth moving it back up ?

jamesobutler:
My opinion is that menu items are organized by typical location of options based on a shared standard of other applications. If most used item is to always be the first entry then we would be changing other orders of actions in other menus as well, but that doesn’t seem to be the desired intent here.

lassoan:
If we experience that the current implementation is not optimal then we should improve it.

Another example where I find that the layout is visually looks much nicer but it is harder to use is Help / Report a bug. It is just much more difficult to find it because there are many more items to read through. The icon somewhat helps, but still hard. Maybe moving it higher (so that you need to read through less actions if you read from top-to-bottom) or moving it to the last position (just above or below "About 3D Slicer", to make it stand out more) would make it easier to spot it.

jamesobutler:
We should follow UI guidelines to appropriately design menus where there are unique considerations for various menus such as between menu bar menus and context menus. I’ve begun to read more into this with some linked sections below.

A potential consideration for a commonly used menu bar item is if it is very frequently used maybe it should have an available option from the toolbar area as well. It should also have a standard shortcut key if possible. It currently has an odd “Ctrl+2” shortcut. Have you ever used this?

https://docs.microsoft.com/en-us/windows/win32/uxguide/cmd-menus#standard-menus

> Standard menus
> 
> Use the standard menu organization for programs that create or view documents. The standard menu organization makes common menu items predictable and easier to find.
> For other types of programs, use the standard menu organization only when it makes sense to. Consider organizing your commands and options into more useful, natural categories based on your program's purpose and the way users think about their tasks and goals.
> Standard menu bars
> The standard menu bar structure is as follows. This list shows the menu category and item labels, their order with separators, their access and shortcut keys, and their ellipses.

> File

> New Ctrl+N
> Open... Ctrl+O
> Close Save Ctrl+S
> Save as... Send to Print... Ctrl+P
> Print preview
> Page setup
> 1 <filename>
> 2 <filename>
> 3 <filename>
> ...
> Exit Alt+F4 (shortcut usually not given)

> Edit

> Undo Ctrl+Z
> Redo Ctrl+Y
> <separator>
> Cut Ctrl+X
> Copy Ctrl+C
> Paste Ctrl+V
> <separator>
> Select all Ctrl+A
> <separator>
> Delete Del (shortcut usually not given)
> <separator>
> Find... Ctrl+F
> Find next F3 (command usually not given)
> Replace... Ctrl+H
> Go to... Ctrl+G

> View

> Toolbars
> Status bar
> <separator>
> Zoom Zoom in Ctrl++
> Zoom out Ctrl+-
> <separator>
> Full screen F11
> Refresh F5

> Tools

> <separator>
> Options Help <program name> help F1
> <separator>
> About <program name>
> 

https://docs.microsoft.com/en-us/windows/win32/uxguide/cmd-menus#menu-item-organization-and-order

> - Within a menu, put the groups in their logical order. If there is no logical order, place the most commonly used groups first.
> - Within a group, put the items in their logical order. If there is no logical order, place the most commonly used items first. Put numeric items (such as zoom percentages) in numeric order.

https://docs.microsoft.com/en-us/windows/win32/uxguide/cmd-menus#shortcut-keys

> Assign shortcut keys to the most frequently used menu items. Infrequently used menu items don't need shortcut keys because users can use access keys instead.
> 
> Use Ctrl key combinations for commands that have a large-scale effect, such as commands that apply to an entire document. For example, Ctrl+S saves the current document.

https://docs.microsoft.com/en-us/windows/win32/uxguide/cmd-menus#icons

> Icons
> 
> Consider providing menu item icons for:
> The most commonly used menu items.
> Menu items whose icon is standard and well known.
> Menu items whose icon well illustrates what the command does.

https://docs.microsoft.com/en-us/windows/win32/uxguide/cmd-menus#menu-item-names

> Use the following menu item names for the stated purpose:
> - Options To display program options.
> - Customize To display the program options specifically related to mechanical UI configuration.
> - Personalize To display a summary of commonly used [personalization](https://docs.microsoft.com/en-us/windows/win32/uxguide/glossary) settings.
> - Preferences Don't use. Use Options instead.
> - Properties To display an object's property window.
> - Settings Don't use as a menu label. Use Options instead.

lassoan:
> It currently has an odd “Ctrl+2” shortcut. Have you ever used this?

Yes, I use it often when my hand is on the keyboard. If I'm using mouse then I use the menu.

> https://docs.microsoft.com/en-us/windows/win32/uxguide/cmd-menus#standard-menus

This was created long time ago. Microsoft did not follow it very closely at the time these guidelines were released (the guidelines were too new then) and then it moved on (the guidelines are now too old). For example, VS Code uses File/Preferences; and in MS Word Customize is within Options (so they gave up on the artificial separation of Options and Customize at the top level). Sentence/title case guidelines do not seem to be consistently followed either.

Unfortunately, Microsoft does not have an up-to-date desktop application UI design guidelines document. Maybe it is just too hard to provide developer documentation when the UX paradigm of the operating system and core applications is changing continuously. But even if Microsoft had a clear vision of how things should work, we would have trouble following those closely, as we use Qt, which does not support modern desktop GUI concepts. We also need to keep an eye on macOS conventions and other cross-platform applications to avoid being outliers.

I think we need to do something like we do with the coding guidelines: list some common UX specifications that we try to follow (e.g., some from Microsoft, some from Apple, some from cross-platform software) and specify what we do if they contradict.

jamesobutler:
> We also need to keep an eye on macOS conventions and other cross-platform applications to avoid being outliers.

Agreed. Here’s the latest information regarding the menu bar in macOS with specifics on the Edit menu.
https://developer.apple.com/design/human-interface-guidelines/macos/menus/menu-bar-menus/

lassoan:
I keep using Slicer versions that have settings action at the top and that have the settings action at the bottom of the menu. Having the menu at the bottom is just annoying. Makes each selection much more tedious. Both using touchpad and mouse. I know that I can use Ctrl-2, but still, sometimes I just move up the mouse to the Edit menu and realize that from there it is not an easy move anymore to get to the settings. I think it is particularly frustrating because I know how easy it is to fix it.

lassoan:
I'm closing this to reduce the number of open pull requests. May be revisited later.

PR #6266, opened 2022-03-15
DOC: Add plot view widget example to script repo

ebrahimebrahim:
Add this gist to the script repository: https://gist.github.com/ebrahimebrahim/8b899552da072fdd22727cfbf0e3a804

I wrote this for my use case and then found that it could be useful in general.

Suggestions to change any phrasing in the the documentation or in the gist itself are welcome.

### DIFF
--- a/Docs/developer_guide/script_repository/plots.md
+++ b/Docs/developer_guide/script_repository/plots.md
@@ -62,6 +62,12 @@ plotView = slicer.app.layoutManager().plotWidget(0).plotView()
 plotView.saveAsSVG("c:/tmp/test.svg")
 ```
 
+### Create a plot view widget, and use a class to manage its associated nodes
+
+This example shows how to create a plot view widget without the associated plot chart node. The plot _chart_ widget is meant to fit into a Slicer layout (it has a colored bar with a controller), while the plot _view_ widget is just the part with the actual plot. This example also sets up a class to automate management of the various MRML nodes associated to a plot view, making it easy to update the plotted data from code.
+
+https://gist.github.com/ebrahimebrahim/8b899552da072fdd22727cfbf0e3a804
+
 ### Using matplotlib
 
 Matplotlib may be used from within Slicer, but the default Tk backend locks up and crashes Slicer. However, Matplotlib may still be used through other backends. More details can be found on the [MatPlotLib](http://matplotlib.sourceforge.net/) pages.


lassoan:
Thank you for contributing this example. I've tried to use it but copy-pasting it to the Python console did not work (known issue, line indentation problems, due to difficulty of splitting up the string to commands). Please provide an example of how it can be used via the command line or some other way.

lassoan:
Also, please check out the slicer.util.plot function. Maybe it does something similar to your script and instead of adding a new example, maybe the existing plot function could be improved.

ebrahimebrahim:
> Please provide an example of how it can be used via the command line or some other way.

I added a comment at the beginning of the gist suggesting to run it with `--python-script`.

> Also, please check out the slicer.util.plot function.

My snippet already uses slicer.util.plot. The differences are:
- No plot chart node is created, only the plot view widget
- A class is created that removes the need to keep track of series and table nodes manually.

I think this is useful for an application developer who just wants a widget showing a plot and who wants it to be easy to update.

If it seems not to fit very well in the script repository please feel free to close this PR! I wrote this for my own use and I'm throwing it out there in case it's useful, but I totally understand if it doesn't belong

jcfr:
And here are some suggestion to simplify the gist:
https://gist.github.com/jcfr/4ed6b66389c081d72b9a5793b352df76/revisions#diff-85a2ec031add9c063a4f3d46a1f310064c5dd305d41bdc13a323f6625bb7abf0

lassoan:
To me it seems that the gist is a wrapper over `slicer.util.plot` to give it a more convenient interface. Instead of adding a new layer, let's improve the interface so that such wrappers are not necessary! This is important because `slicer.util.plot` exists just because API of MRML plots is too complicated. It is a convenience layer. If the convenience layer is still too inconvenient then we must not add yet another convenience layer on top of it, but fix it.

Could you describe what were the main issues with the `slicer.util.plot` API and if you have any suggestions for improving it?
- For example, we could change the `nodes` argument from a simple `dict` to class that is similar to what you have in your gist.
- We could think about how to avoid the use of complicated constant names, such as `slicer.vtkMRMLPlotSeriesNode.PlotTypeLine`. Maybe adding `SetTypeToX` methods to the series could help? Or we could add simpler constants to `slicer.util`, or add convenience functions to the new Python class.
- We could add the new Python class that you created in `slicer.util`, implementing all current `slicer.util.plot` features; and implement `slicer.util.plot` function by using this new class.

I think the only new feature of the gist is to create a plot in a widget outside the view layout. This could be added to `slicer.util.plot` (or the new class). For this, we need a small fix: the plot view node's ParentLayoutNode must be set to a node (it can be any node except the scene's singleton layout node; and cannot be empty either because it would mean that it is in the default layout node), to indicate that this view is not managed by the default layout manager.

jcfr:
@lassoan To understand the use case:
* https://github.com/KitwareMedical/lungair-desktop-application/blob/master/Modules/Scripted/Home/HomeLib/plots.py

* https://github.com/KitwareMedical/lungair-desktop-application/blob/1cf6deb7d6cce0f63944aefdd463ea6053211428/Modules/Scripted/Home/Home.py#L308-L313

* https://github.com/KitwareMedical/lungair-desktop-application/blob/1cf6deb7d6cce0f63944aefdd463ea6053211428/Modules/Scripted/Home/Home.py#L280-L283

In a nutshell, there a `QTabWidget` where each tab is a different plot.

And to understand why the a `tabbed layout` is not used. See https://github.com/KitwareMedical/lungair-desktop-application/issues/22#issuecomment-1056035789

lassoan:
It has pros/cons of using the layout manager and so there are use cases when placing views outside the view layout makes sense. I'm completely fine with that.

What we need to fix is the `slicer.util.plot` function (potentially by adding an object-oriented interface, i.e., a helper class), because the fact that somebody felt the need to add a convenience wrapper layer on top of it showed that the API is bad: it's too complex and/or not powerful enough.

Most likely the best solution would be to add an object-oriented interface (e.g., slicer.util.Plotter class), which has a similar interface to what is in the gist, and make `slicer.util.plot` use that class (to preserve backward compatibility).

lassoan:
I don't feel strongly against adding this gist to the script repository (it would be just one more example, which would be useful), but it would be even better to use the code in the gist to improve the plotting API in `slicer.util`.

PR #6270, opened 2022-03-16
Attempt to fix codespell GitHub Actions worklflow

jcfr:


jcfr:
Errors were expected and setting the `skip` parameter as a single line now report thee:

image

jcfr:
Reverted back to having all file exclusion specified on a single line .. and it is failing as expected.

Next commit will fix the issues.

jcfr:
:heavy_check_mark: The last commit effectively fixes the typos ... and the check is now really passing.

:warning: **important** :warning: Before integrating:

1. commit `COMP: Temporally convert ci worklow into a no-op` should be removed
2. Commit should be squashed/re-organized ..

cc: @jamesobutler

jamesobutler:
Yes thanks for cleaning this up. I noticed it hadn’t been working at some point even though during my original testing it had been. Maybe a last minute change of the multi line stuff had broken it. This looks good, so feel free to drop that commit you mentioned and integrate.

jamesobutler (approved):

PR #6274, opened 2022-03-19
BUG: Desginer/Python launcher additional settings points to wrong file

kerim371:
In build tree `additionalSettingsFilePath` of Python and Designer launcher settings point to wrong file.
Also SlicerCAT's app settings has the name prefix equal to SlicerCAT app name.

Discoussed on the [forum](https://discourse.slicer.org/t/pythonslicerlaunchersettings-ini-points-to-wrong-additionalsettingsfilepath/22528).


CMake/SlicerCPack.cmake
```diff
@@ -77,6 +77,13 @@ if(Slicer_BUILD_QT_DESIGNER_PLUGINS)
   file(MAKE_DIRECTORY ${CMAKE_BINARY_DIR}/${Slicer_BIN_DIR})
   # Configure designer launcher
   find_package(CTKAppLauncher REQUIRED)
+  # At the time this script is executed for both Slicer's and SlicerCAT's binary dir 
+  # is set to './d/Slicer-build' and there application settings resides
+  set(app_name "Slicer")
+  if(DEFINED Slicer_MAIN_PROJECT_APPLICATION_NAME AND 
+    NOT Slicer_MAIN_PROJECT_APPLICATION_NAME STREQUAL ${app_name})
```
lassoan, line 84: Is this line needed? Does it have a side effect if we set the variable to the same value?

Also note that `NOT Slicer_MAIN_PROJECT_APPLICATION_NAME STREQUAL ${app_name}` is always true because `Slicer_MAIN_PROJECT_APPLICATION_NAME` is just a string. To get the variable value `${Slicer_MAIN_PROJECT_APPLICATION_NAME}` should have been used.

kerim371:
@lassoan I tried to follow your suggestion.
Just in case I have tried to check cmake behaviour and it seems that passing var names to  `if/else` statements is completely fine. 
As I understood cmake first search if the passed name is variable and if so then it unwraps it and then compares. The following works as expected (return `EQUAL`):
```cmake
set(var1 "apple")
set(var2 "apple")
if(var1 STREQUAL var2)
  message("EQUAL")
else()
  message("NOT EQUAL")
endif()
```
and any variations like `if(${var1} STREQUAL var2)`, `if(${var}1 STREQUAL ${var2})` return `EQUAL`.
When I compare `var1` with unexistent `var3` then I get `NOT EQUAL` that means in that case it is not necessary to check `if(DEFINED ...`

My cmake version is 3.19.2

CMake/SlicerCPack.cmake
```diff
@@ -85,11 +87,11 @@ if(Slicer_BUILD_QT_DESIGNER_PLUGINS)
     # Launcher settings specific to build tree
     APPLICATION_EXECUTABLE ${build_designer_executable}
     DESTINATION_DIR ${CMAKE_BINARY_DIR}/${Slicer_BIN_DIR}
-    ADDITIONAL_SETTINGS_FILEPATH_BUILD "${Slicer_BINARY_DIR}/${Slicer_BINARY_INNER_SUBDIR}/SlicerLauncherSettings.ini"
+    ADDITIONAL_SETTINGS_FILEPATH_BUILD "${Slicer_BINARY_DIR}/${Slicer_MAIN_PROJECT_APPLICATION_NAME}LauncherSettings.ini"
```
lassoan, line 90: Have you intentionally removed `${Slicer_BINARY_INNER_SUBDIR}`?

lassoan, line 90: Have you tested your changes with a regular Slicer build, too?

kerim371:
@lassoan I checked on Ubuntu with clean Slicer built from scratch and the paths now correct.
Ready to merge

jamesobutler:
Looks like most of these changes were integrated as part of https://github.com/Slicer/Slicer/commit/1bdfefe4e275c9b5713503077930fe387aa8a324.

jamesobutler:
@kerim371 If you find there was some changes that were not integrated in https://github.com/Slicer/Slicer/commit/1bdfefe4e275c9b5713503077930fe387aa8a324 that are still needed for custom applications to be configured and packaged correctly please reopen this PR after rebasing this branch with latest changes to fix merge conflicts.

kerim371:
@jamesobutler it seems I don't have privilegies to reopen this PR.
Should I open new one or there are ways to reopen this one?

P.S. in SlicerCAT in the **build** tree we still have wrong path to additional settings file in both Python and Designer settings.

lassoan:
I am an admin but I cannot reopen it either. Most likely due to the rename of `master` branxh to `main`. You need to create a new pull request with the required changes (and refer to this pull request in it).

kerim371:
Reopened in #6463

PR #6275, opened 2022-03-19
BUG: segfault invoked by SettingsDialog deletion

kerim371:
Previous implementation worked only when when mainwindow is deleted after app.
If mainwindow is deleted before app then we get seg fault in destructor.
I.e. we are trying to set parent to destroyed SettingsDialog.
This used to happen with SlicerCAT.

QPointer helps to track the lifetime of a pointer.
So we always know if it is alive.

Steps to reproduce:
1) launch SlicerCAT from command line
1) open `Application settings`
2) close it
3) close application
4) see output to command line

### DIFF
--- a/Base/QTGUI/qSlicerApplication.cxx
+++ b/Base/QTGUI/qSlicerApplication.cxx
@@ -141,8 +141,8 @@ public:
   QSettings* newSettings() override;
 
   QPointer<qSlicerLayoutManager> LayoutManager;
-  ctkToolTipTrapper*      ToolTipTrapper;
-  ctkSettingsDialog*      SettingsDialog;
+  ctkToolTipTrapper* ToolTipTrapper;
+  QPointer<ctkSettingsDialog> SettingsDialog;
 #ifdef Slicer_BUILD_EXTENSIONMANAGER_SUPPORT
   qSlicerExtensionsManagerDialog* ExtensionsManagerDialog;
   bool IsExtensionsManagerDialogOpen;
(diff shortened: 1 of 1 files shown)


Base/QTGUI/qSlicerApplication.cxx
```diff
@@ -178,9 +178,11 @@ qSlicerApplicationPrivate::qSlicerApplicationPrivate(
 qSlicerApplicationPrivate::~qSlicerApplicationPrivate()
 {
   // Delete settings dialog. deleteLater would cause memory leaks on exit.
-  this->SettingsDialog->setParent(nullptr);
-  delete this->SettingsDialog;
-  this->SettingsDialog = nullptr;
+  if (this->SettingsDialog)
+    {
+    this->SettingsDialog->setParent(nullptr);
```
lassoan, line 183: Describe here why we must choose check if non-null (because the main window may have already deleted it) and why we need to set the parent to null (because it may have been assigned to the main window).

Base/QTGUI/qSlicerApplication.cxx
```diff
@@ -141,8 +141,8 @@ class qSlicerApplicationPrivate : public qSlicerCoreApplicationPrivate
   QSettings* newSettings() override;
 
   QPointer<qSlicerLayoutManager> LayoutManager;
-  ctkToolTipTrapper*      ToolTipTrapper;
-  ctkSettingsDialog*      SettingsDialog;
+  ctkToolTipTrapper* ToolTipTrapper;
+  QPointer<ctkSettingsDialog> SettingsDialog;
```
lassoan, line 145: Describe the ownership of the settings dialog: when a main window exists and the dialog is displayed then the main window must be set as parent to ensure correct z order; but that also transfers the ownership of the object, therefore we use a smart pointer to keep track if the object is deleted already by the main window.

lassoan (commented):
Thank you, it looks good, just add comments that describes why all the extra complications with smart pointer and checks and setting parent are necessary.

Base/QTGUI/qSlicerApplication.cxx
kerim371: fixed

Base/QTGUI/qSlicerApplication.cxx
kerim371: fixed

lassoan:
Thank you, merged!

PR #6276, opened 2022-03-20
Update website links to https sites

jamesobutler:
This updates web links to specify the https site specifically. This often removes an additional  redirect from the http to https site. Not all sites have an http equivalent which is why there are still some http links. Through this activity I have come across many dead links which should get cleaned up later.

There are multiple commits here just for some organization while I was doing this process, but they can be squashed down to one if that is better for the git blame commit history.


jcfr:
The file `Base/QTCore/Resources/Certs/Slicer.crt` is generated automatically, we should look into why it doesn't include `https:` by default.

Looking at the raw [certdata.txt](https://raw.githubusercontent.com/mozilla/gecko-dev/master/security/nss/lib/ckfw/builtins/certdata.txt), it is not obvious where the `http://` comes from as the only references to `http://` are in comments mentioning `http://www.chambersign.org`.

The use of openssl for generating the x509 certificates should be reviewed to better understanding what could be done to address this (if it makes sense to address it in that specific context)

References:
* https://github.com/Slicer/Slicer/tree/master/Base/QTCore/Resources/Certs
* https://raw.githubusercontent.com/mozilla/gecko-dev/master/security/nss/lib/ckfw/builtins/certdata.txt

Base/QTGUI/Resources/Icons/ExtensionIncompatible.svg
```diff
@@ -1,15 +1,15 @@
...
-   xmlns:svg="http://www.w3.org/2000/svg"
-   xmlns="http://www.w3.org/2000/svg"
-   xmlns:xlink="http://www.w3.org/1999/xlink"
+   xmlns:dc="https://purl.org/dc/elements/1.1/"
+   xmlns:cc="https://creativecommons.org/ns#"
+   xmlns:rdf="https://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:svg="https://www.w3.org/2000/svg"
+   xmlns="https://www.w3.org/2000/svg"
+   xmlns:xlink="https://www.w3.org/1999/xlink"
    xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
```
jcfr, line 11: Wondering why this reference was not updated, I realized that current tools are not saving new files with `https://`.

For that reason, I suggest we do not update our `svg` files.

Few examples:

### Inkscape 1.1.2 (76b9e6a115, 2022-02-25)

```
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<!-- Created with Inkscape (http://www.inkscape.org/) -->

<svg
   width="210mm"
   height="297mm"
   viewBox="0 0 210 297"
   version="1.1"
   id="svg5"
   inkscape:version="1.1.2 (76b9e6a115, 2022-02-25)"
   sodipodi:docname="drawing.svg"
   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
   xmlns="http://www.w3.org/2000/svg"
   xmlns:svg="http://www.w3.org/2000/svg">
....
```

### Adobe Illustrator 24.2.1

Looking at files generated (in July 2020) using `Adobe Illustrator` (for example `3D Slicer_Mark.svg`), they still reference `http://`. 

```
<?xml version="1.0" encoding="utf-8"?>
<!-- Generator: Adobe Illustrator 24.2.1, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
<svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
         viewBox="0 0 512 513.3" style="enable-background:new 0 0 512 513.3;" xml:space="preserve">
<style type="text/css">

...
```

Base/QTGUI/Resources/UI/qSlicerSettingsStylesPanel.ui
```diff
@@ -68,7 +68,7 @@ p, li { white-space: pre-wrap; }
          <item>
           <widget class="QPushButton" name="RemoveAdditionalStylePathButton">
            <property name="toolTip">
-            <string>&lt;!DOCTYPE HTML PUBLIC &quot;-//W3C//DTD HTML 4.0//EN&quot; &quot;http://www.w3.org/TR/REC-html40/strict.dtd&quot;&gt;
+            <string>&lt;!DOCTYPE HTML PUBLIC &quot;-//W3C//DTD HTML 4.0//EN&quot; &quot;https://www.w3.org/TR/REC-html40/strict.dtd&quot;&gt;
```
jcfr, line 71: Considering that
1. the official RFC reference the `http` link [^1]
2. current test seems to look for `http://` to detect if a content is rich text [^2][^3],
3. the `richtexteditor` built in Qt designer default the simplified text editor [64]
4. new version of Qt designer (checked with Qt5 and Qt6) do not include the dtd in generated ui files (but still matching `<ui version="4.0">` like the Slicer ones), 


I instead suggest we simplify the UI files by re-creating the content using Qt designer. This is required because in case it need to be converted from RichText to HTML.

```
Base/QTGUI/Resources/UI/qSlicerSettingsStylesPanel.ui
Base/QTGUI/Resources/UI/qSlicerSettingsViewsPanel.ui
Modules/Loadable/Units/Resources/UI/qSlicerUnitsSettingsPanel.ui
Modules/Loadable/Units/Widgets/Resources/UI/qMRMLUnitWidget.ui
```

Last, the `QTextDocument::toHtml` function still reference the `http` link. [^5]

[^1]: https://www.w3.org/TR/html401/strict.dtd  (note also that the link redirect to `https`)

[^2]: https://github.com/qt/qttools/blob/ce05219da0750a3be429b396ea0d13980aa14d70/src/designer/src/lib/shared/richtexteditor.cpp#L848-L856

[^3]: https://github.com/qt/qtbase/blob/9db7cc79a26ced4997277b5c206ca15949133240/tests/benchmarks/gui/text/qtextdocument/main.cpp#L48-L52

[^4]: https://github.com/qt/qttools/blame/ce05219da0750a3be429b396ea0d13980aa14d70/src/designer/src/lib/shared/richtexteditor.cpp#L70

[^5]: https://github.com/qt/qtbase/blob/32e24e011ec206f99c61b981cabb6ed5fe06a57b/src/gui/text/qtextdocument.cpp#L2358-L2363

SuperBuild.cmake
```diff
@@ -43,7 +43,7 @@ if(EP_GIT_PROTOCOL STREQUAL "https")
 "adding 'git config' statement between the 'git checkout' and the 'submodule init', it is required "
 "to manually update your global git config to successfully build ${CMAKE_PROJECT_NAME} with "
 "option Slicer_USE_GIT_PROTOCOL set to FALSE. "
-"See http://na-mic.org/Mantis/view.php?id=2731"
+"See https://na-mic.org/Mantis/view.php?id=2731"
```
jcfr, line 46: Reference to the former issue tracker should be switched to `https://mantisarchive.slicer.org/view.php?id=NNN`, in this case it would be https://mantisarchive.slicer.org/view.php?id=2731

Base/QTCore/Testing/Cxx/qSlicerSslTest.cxx
```diff
@@ -103,7 +103,7 @@ void qSlicerSslTester::testHttpsConnection_data()
       << QNetworkReply::HostNotFoundError << 0;
 
   QTest::newRow("slicer-clear-with-redirect-NoError-301")
-      << "http://slicer.org"
+      << "https://slicer.org"
```
jcfr, line 106: This should not be changed

Base/QTGUI/Resources/Icons/ExtensionIncompatible.svg
jamesobutler: Ok we can remove the svg updates here. I think I hadn’t originally updated it because https://sodipodi.sourceforge.net/ wasn’t a valid link though I guess https://sourceforge.net/projects/sodipodi/ is valid.

SuperBuild.cmake
jamesobutler: Thanks for that info! Yes that will fix these current dead links.

Base/QTGUI/Resources/Icons/ExtensionIncompatible.svg
lassoan: These are not website links, but namespaces. They must not be changed. See for example https://stackoverflow.com/questions/30707609/xml-namespace-uri-with-https

Base/QTGUI/Resources/UI/qSlicerSettingsStylesPanel.ui
lassoan: These https occurrences must not be changed. They are not for retrieving content but for identification.

jamesobutler: Changes have been reverted here.

jamesobutler:
@jcfr Can you summarize here the changes you mentioned during hangout about the improvements to the Copyright header section that has been touched here in many of the files?

jcfr:
In `ENH: Update other website links to https version`:

* Revert changes to `Slicer.crt`. See https://github.com/Slicer/Slicer/pull/6276#issuecomment-1074701982
* Revert changes to `MacOSXBundleInfo.plist.in` (the reference to the `dtd` falls in the namespace category).
* Regarding the changes to `markups-schema-vx.y.z.json` (e.g [markups-schema-v1.0.0.json](https://github.com/Slicer/Slicer/blob/46369ebd8a2e2e6fe64da7b6f7bac01900fd2750/Modules/Loadable/Markups/Resources/Schema/markups-schema-v1.0.0.json#L2)), I suggest to also revert them as they likely all in the namespace & identification category. @lassoan ?
* Same remark for `volumes-display-presets-schema-vx.y.z.json`

In `ENH: Update NAMIC website links to https versions`:

* Links like `https://www.na-mic.org/Bug/view.php?id=NNN` should be updated to `https://mantisarchive.slicer.org/view.php?id=NNN`

In `ENH: Update Slicer website links to https versions`

* Since there has been discussion of re hosting the publication database. I suggest to now update links like `http://www.slicer.org/publications/item/view/1953`

> during hangout about the improvements to the Copyright header section that has been touched here in many of the files?

I summarized how we could move forward in https://github.com/Slicer/Slicer/issues/6281

In the mean time, I suggest we drop the update of http to https for the copyright links.

jcfr:
Assuming we consider these commits as "style", we could now also start listing them in `.git-blame-ignore-revs`. See See https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view

lassoan:
> * Regarding the changes to `markups-schema-vx.y.z.json` (e.g [markups-schema-v1.0.0.json](https://github.com/Slicer/Slicer/blob/46369ebd8a2e2e6fe64da7b6f7bac01900fd2750/Modules/Loadable/Markups/Resources/Schema/markups-schema-v1.0.0.json#L2)), I suggest to also revert them as they likely all in the namespace & identification category. @lassoan ?
> * Same remark for `volumes-display-presets-schema-vx.y.z.json`

I would not change http to https in any of the schema files. It would not add any value but it would be a potential source of errors. We may consider changing the protocol in the future if we make any major change in the namespaces.

lassoan:
I also find it useful if the author is set to a bot user for such bulk automatic changes. That makes it very clear that the change is result of an automatic algorithm for human readers. It is also easier to filter these changes out for automatic processing and does not skew code modification statistics.

jamesobutler:
> That makes it very clear that the change is result of an automatic algorithm for human readers.

In this case a bot could have been made to do these changes, but what I did was use basic find/replace tools. Not sure if you would consider those tools to be automatic or manual. It's like manual assisted techniques as there is human approval/disapproval of potential replacements. There was manual checking to determine if there were https versions of the domains. A bot could probably have been made to confirm going to these links produces a valid request, but I did things manually here which would require reviewers to take my word for it rather than trying to prove the changes are indeed correct.

CMakeLists.txt
```diff
@@ -62,7 +62,7 @@ set(Slicer_CMAKE_DIR ${Slicer_SOURCE_DIR}/CMake)
 set(Slicer_EXTENSIONS_CMAKE_DIR ${Slicer_SOURCE_DIR}/Extensions/CMake)
 
 #-----------------------------------------------------------------------------
-# http://www.slicer.org/slicerWiki/index.php/Documentation/Nightly/Developers/DevelopmentWithGit
+# https://www.slicer.org/slicerWiki/index.php/Documentation/Nightly/Developers/DevelopmentWithGit
```
jcfr, line 65: Since the current infrastructure redirects link like

```
https://www.slicer.org/slicerWiki/index.php/Documentation/Nightly/Developers/DevelopmentWithGit
```

to 

```
https://www.slicer.org/w/index.php/Documentation/Nightly/Developers/DevelopmentWithGit
```

I suggest to update occurrences of `www.slicer.org/slicerWiki` to `www.slicer.org/w`

Base/QTCore/qSlicerAbstractCoreModule.h
```diff
@@ -204,7 +204,7 @@ class Q_SLICER_BASE_QTCORE_EXPORT qSlicerAbstractCoreModule : public QObject
   virtual void printAdditionalInfo();
 
   /// Convenience method to return slicer wiki URL
-  QString slicerWikiUrl()const{ return "http://www.slicer.org/slicerWiki/index.php"; }
+  QString slicerWikiUrl()const{ return "https://www.slicer.org/slicerWiki/index.php"; }
```
jcfr, line 207: My comment in https://github.com/Slicer/Slicer/pull/6276/commits/4934ca63e8f4e8f1b548771d17266f1d3896ee08#r844439362 was motivated by the fact trying to browse to `https://www.slicer.org/slicerWiki/index.php` fails.

In this form, the following works `https://www.slicer.org/slicerWiki/index.php/`

That said https://www.slicer.org/w/index.php works as expected and do not return a `404`

Base/Logic/Resources/ColorFiles/DarkBrightChartColors.txt
```diff
@@ -4,7 +4,7 @@
 #Filename: Slicer4_2012_DarkBrightChartColors.txt
 #Title: Slicer4:2012 DarkBrightChartColors
 #PageName: Slicer4:2012_DarkBrightChartColors
-#Source: http://www.slicer.org/slicerWiki/index.php/Slicer4:2012_DarkBrightChartColors
+#Source: https://www.slicer.org/slicerWiki/index.php/Slicer4:2012_DarkBrightChartColors
```
jcfr, line 7: @lassoan  Similarly to svg/markups/... files, should we consider occurrence of such URL in color file as a schema reference ?

Base/Logic/Resources/ColorFiles/SPL-BrainAtlas-2009-ColorFile.txt
```diff
@@ -4,7 +4,7 @@
 #Filename: CCA_Atlases_MRI_Brain_Atlas_2010_ColorFile.txt
 #Title: CCA:Atlases:MRI Brain Atlas 2010:ColorFile
 #PageName: CCA:Atlases:MRI_Brain_Atlas_2010:ColorFile
-#Source: http://www.na-mic.org/Wiki/index.php/CCA:Atlases:MRI_Brain_Atlas_2010:ColorFile
+#Source: https://www.na-mic.org/Wiki/index.php/CCA:Atlases:MRI_Brain_Atlas_2010:ColorFile
```
jcfr, line 7:
```suggestion
#Source: https://www.slicer.org/w/index.php/CCA:Atlases:MRI_Brain_Atlas_2010:ColorFile
```

Base/Logic/Resources/ColorFiles/DarkBrightChartColors.txt
jcfr: After further consideration, there are URL and I think we should be updated. I will create an issue to track the fact we need to find these a more legitimate home outside of the wiki.

jcfr (approved):
This is ready for integration, existing links are all redirected already.

jamesobutler:
@jcfr I've committed an extra commit here for the reduction of those redirects from `www.slicer.org/slicerWiki/...` to `www.slicer.org/w/...`

Base/Logic/Resources/ColorFiles/DarkBrightChartColors.txt
jcfr: For reference, here is the issue https://github.com/Slicer/Slicer/issues/6305

jcfr (approved):

PR #6279, opened 2022-03-22
BUG: Fix vtk referenced before assignment error in GE ABUS DICOM plugin

lassoan:
This commit fixes this error (occurs when trying to load a GE ABUS image):

```
Warning in DICOM plugin GE ABUS when examining loadable 1: US TEST1 [1]: Loading of this image type is experimental. Please verify image size and orientation and report any problem is found.
Warnings detected during load.  Examine data in Advanced mode for details.  Load anyway?
DICOM plugin failed to load '1: US TEST1 [1]' as a 'GE ABUS'.
  File "C:\Users\andra\AppData\Local\NA-MIC\Slicer 4.13.0-2022-03-19\lib\Slicer-4.13\qt-scripted-modules\DICOMLib\DICOMUtils.py", line 790, in loadLoadables
    loadSuccess = plugin.load(loadable)
  File "C:/Users/andra/AppData/Local/NA-MIC/Slicer 4.13.0-2022-03-19/bin/../lib/Slicer-4.13/qt-scripted-modules/DICOMGeAbusPlugin.py", line 212, in load
    acquisitionTransform = self.createAcquisitionTransform(volumeNode, metadata)
  File "C:/Users/andra/AppData/Local/NA-MIC/Slicer 4.13.0-2022-03-19/bin/../lib/Slicer-4.13/qt-scripted-modules/DICOMGeAbusPlugin.py", line 254, in createAcquisitionTransform
    gridImage = vtk.vtkImageData()
UnboundLocalError: local variable 'vtk' referenced before assignment
```

### DIFF
--- a/Modules/Scripted/DICOMPlugins/DICOMGeAbusPlugin.py
+++ b/Modules/Scripted/DICOMPlugins/DICOMGeAbusPlugin.py
@@ -251,6 +251,7 @@ class DICOMGeAbusPluginClass(DICOMPlugin):
 
     # create a grid transform with one vector at the corner of each slice
     # the transform is in the same space and orientation as the volume node
+    import vtk
     gridImage = vtk.vtkImageData()
     gridImage.SetOrigin(*volumeNode.GetOrigin())
     gridImage.SetDimensions(samplingPoints_shape[:3])


lassoan:
@jcfr do you know why importing VTK at the top-level does not work for this DICOM plugin?

jcfr:
> o you know why importing VTK at the top-level does not work for this DICOM plugin?

This is a regression introduced in https://github.com/Slicer/Slicer/commit/3a9490ed1ad6f26de027338ff612e9cd9a967168#diff-82b5c02daa028be37db442e3b6de0f29d02b08b78d29968eb45a7582340dca5d

We should understand why this is happening ... I may have an intuition about it.

jcfr:
It may be specific to windows, would you have sample dateset to share (privately) ?

lassoan:
@jcfr If you don't have time to dig into this now then we could merge the current workaround now and create an issue to investigate later.

jcfr:
I was able to reproduce, if we don't have a fix before tomorrow morning. You suggested plan of action will be sensible.

Since the recent update related to python imports likely introduced similar regress in few places, I would like to take some time to understand and if we can fix it "the right way", restore the specific import statement.

jcfr:
The problem can reproduced by copy/paste the following into the python interactor:

```python
import vtk

def createAcquisitionTransform():
  """The purpose of this function is to illustrate the side effect
  of importing vtk.util.numpy_support.
  
  Issue is most likely related to how internally `vtk.py` imports
  `vtkmodules.all` and `vtkmodules` and then replace the originaly
   `vtk` module available in `sys.modules` with some hacked version
   of `vtkmodules.all`.
  
  See https://github.com/Kitware/VTK/blob/master/Wrapping/Python/vtk.py
  
  The root cause of this behavior is most likely what is causing problem
  disscussed at https://github.com/KitwareMedical/lungair-desktop-application/issues/4#issuecomment-1024540602
  """
  gridImage = vtk.vtkImageData()
  import vtk.util.numpy_support



createAcquisitionTransform()
```

it then raises the following exception:

```pytb
  File "<console>", line 2, in createAcquisitionTransform
UnboundLocalError: local variable 'vtk' referenced before assignment
```



See also https://github.com/KitwareMedical/lungair-desktop-application/issues/4#issuecomment-1024540602

jcfr:
Interestingly the issue is also reproducible using this example:

```python
import os

def test_os():
  print(os)
  import os.path

test_os()
```

it raises

```
UnboundLocalError: local variable 'os' referenced before assignment
```

Connor-Bowley:
From what I can see this is documented, expected Python behavior. The `import vtk.util.numpy_support` essentially declares vtk as a local variable, which is then used unassigned in the prior statement.

References:
https://bugs.python.org/issue35069

jcfr:
Ditto. I will have a proper fix pushed shortly.

jcfr:
(just realized I forgot to click "comment" with this list)

Some references:
* https://bugs.python.org/issue35069 (Unexecuted import in function causes UnboundLocalError)
* https://gist.github.com/yatt/3506154
* https://vbsreddy1.medium.com/unboundlocalerror-when-the-variable-has-a-value-in-python-e34e097547d6
* https://stackoverflow.com/questions/37956270/imported-module-goes-out-of-scope-unbound-local-error

jcfr:
Closing. This is superseded by https://github.com/Slicer/Slicer/pull/6280

PR #6280, opened 2022-03-23
BUG: Consolidate vtk_to_numpy imports to fix UnboundLocalError

jcfr:
This commit ensures attributes associated with vtk modules can be referenced
from functions where "vtk" is imported at the top-level
and "vtk.util.numpy_support" import is delayed.

It fixes a regression introduced in 3a9490ed1 (STYLE: Remove unused
python imports) where imported have also been re-organized.

See https://bugs.python.org/issue35069 (Unexecuted import in function
causes UnboundLocalError)

Note that imports associated with Python/slicer/utils.py have not been
updated. They are all consistently done at the function level.


lassoan (approved):

lassoan:
Thank you, I was not aware of this side effect of imports in Python.

Importing `vtk.something` leading to deletion of `vtk.*` (and not just reassignment of `vtk.something`) is unnecessary and unexpected, so I would consider it as a flaw in Python. But Python developers seem to be pushing back very hard on this, so probably it would be difficult for them to implement the expected behavior. So, I guess we just need to accept this and be more careful with local imports.

PR #6282, opened 2022-03-25
BUG: Fix imageData content copying in vtkMRMLStreamingVolumeNode

Sunderlandkyl:
When calling CopyContent, the target ImageConnection was used as the source. This caused problems when adding un-encoded vtkMRMLStreamingVolumeNode to sequences.


lassoan (approved):
Thank you, it looks good.

PR #6290, opened 2022-03-31
ENH: Close application after save

riep:
After closing the application, a pop-up appears (if data have not been saved previously). After clicking on save, the application does not close.
This PR fixes this issue: after saving, the application closes.

### DIFF
--- a/Base/QTApp/qSlicerMainWindow.cxx
+++ b/Base/QTApp/qSlicerMainWindow.cxx
@@ -642,9 +642,9 @@ bool qSlicerMainWindowPrivate::confirmCloseApplication()
     messageBox->exec();
     if (messageBox->clickedButton() == saveButton)
       {
-      // \todo Check if the save data dialog was "applied" and close the
-      // app in that case
+      // Save data and close the app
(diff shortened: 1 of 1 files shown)


Base/QTApp/qSlicerMainWindow.cxx
```diff
@@ -642,9 +642,9 @@ bool qSlicerMainWindowPrivate::confirmCloseApplication()
     messageBox->exec();
     if (messageBox->clickedButton() == saveButton)
       {
-      // \todo Check if the save data dialog was "applied" and close the
-      // app in that case
+      // Save data and close the app
       this->FileSaveSceneAction->trigger();
```
pieper, line 646: This original comment with the `\todo` indicates that the author was concerned about the case where the save failed or was canceled, in which case you wouldn't want the application to quit since it could mean losing data.  Unless that case is accounted for I don't think we should close the app just because the save dialog was triggered.

riep: Understood, I will implement this check in that case

Issue #6293, opened 2022-03-31
Install extensions is empty


phcerdan:
I am not able to access any content in Extension Managet > Install Extensions tab.
 
![image](https://user-images.githubusercontent.com/3021667/161153014-2c0d35b8-cf31-405a-b895-97a927f4e794.png)

The percentage bar gets into 100% and then white screen. I had no problems in other nightly versions of 4.13.

I thought that maybe it was a problem with my global Slicer.ini, so I removed it to start in a clean-state, but the problem persists.

## Steps to reproduce
Download latest 4.13, go to Extension Managet > Install Extensions tab and see it (in Linux).

## Expected behavior
I expected to see extensions, as in other nightly versions.
I have faced this behavior with Slicer previews from at least one or two months. Before that I remember was working.

## Environment
- Slicer version: Slicer-4.13.0-2022-03-28
- Operating system: Linux / Gnome /  X

lassoan:
This is due to an SSL certificate issue or an overly aggressive firewall.

Is your computer behind a proxy server or corporate firewall?

Can you try installing more Slicer versions and determine which day the problem appeared?

phcerdan:
Not behind a proxy, but it seems related indeed with ca-certificates. I am on Archlinux.

Will explore more and report back. Thanks for the hints @lassoan.

phcerdan:
I can download urls from the Slicer python console, so it's not a SSL problem, as reported by @muratmaga in the discourse.

```python
url = qt.QUrl("https://slicer-packages.kitware.com/api/v1/file/613b26b1342a877cb3bee81e/download")
request = qt.QNetworkRequest(url)
manager = qt.QNetworkAccessManager()
reply = manager.get(request)

while (not reply.isFinished()):
    slicer.app.processEvents()

localFile = qt.QFile("/tmp/downloaded2.zip")
localFile.open(qt.QIODevice.WriteOnly)
localFile.write(reply.readAll());
localFile.close()

print(f"HTTP response code: {reply.attribute(qt.QNetworkRequest.HttpStatusCodeAttribute)}")
print(f"Error code: {reply.error()}")

```
```
>>> print(f"HTTP response code: {reply.attribute(qt.QNetworkRequest.HttpStatusCodeAttribute)}")
HTTP response code: 200
>>> print(f"Error code: {reply.error()}")
Error code: 0
>>> 
```

lassoan:
[...]

PR #6294, opened 2022-04-01
DOC: Describe solution for empty blank extensions manager on ArchLinux

lassoan:
See discussion at https://discourse.slicer.org/t/extension-manager-in-linux-empty/22343/6 and issue #6293 for more details.


pieper:
If this is a persistent issue, why not have the launcher detect the conditions and set the environment variable before launching Slicer?

lassoan:
It does not look very easy to detect the conditions. We could disable the sandbox on Linux in the launcher settings, but I'm not sure if there are any side effects or security implications.

The problem may just go away as people upgrade their systems, Qt makes things more robust, and/or relationship between Google and FOSS community improves.

PR #6299, opened 2022-04-04
Allow removing subject hierachy plugin

riep:
This development comes from the need to remove the right click on markups (in MPR views) in a slicelet application.
Users that implement slicelet may want to have their own  right click  actions.
It is also important to avoid switching into slicer when clicking on "edit properties" action.

### DIFF
--- a/Modules/Loadable/SubjectHierarchy/Widgets/qSlicerSubjectHierarchyPluginHandler.cxx
+++ b/Modules/Loadable/SubjectHierarchy/Widgets/qSlicerSubjectHierarchyPluginHandler.cxx
@@ -118,6 +118,17 @@ qSlicerSubjectHierarchyPluginHandler::~qSlicerSubjectHierarchyPluginHandler()
 }
 
 //---------------------------------------------------------------------------
+bool qSlicerSubjectHierarchyPluginHandler::unRegisterPlugin(QString name)
+{
+  qSlicerSubjectHierarchyAbstractPlugin* currentPlugin = pluginByName(name);
+  foreach(QAction* action, currentPlugin->viewContextMenuActions())
+    {
+    this->m_PluginLogic->unRegisterViewMenuAction(action);
+    }
+  return true;
+}
+
+//---------------------------------------------------------------------------
 bool qSlicerSubjectHierarchyPluginHandler::registerPlugin(qSlicerSubjectHierarchyAbstractPlugin* pluginToRegister)
 {
   if (pluginToRegister == nullptr)
--- a/Modules/Loadable/SubjectHierarchy/Widgets/qSlicerSubjectHierarchyPluginHandler.h
+++ b/Modules/Loadable/SubjectHierarchy/Widgets/qSlicerSubjectHierarchyPluginHandler.h
@@ -135,6 +135,10 @@ public:
   /// \return True if plugin registered successfully, false otherwise
   Q_INVOKABLE bool registerPlugin(qSlicerSubjectHierarchyAbstractPlugin* plugin);
 
+  /// Unregister a plugin
+  /// \return True if plugin is unregistered successfully, false otherwise
+  Q_INVOKABLE bool unRegisterPlugin(QString name);
+
   /// Get list of registered plugins
   Q_INVOKABLE QList<qSlicerSubjectHierarchyAbstractPlugin*> registeredPlugins() { return m_RegisteredPlugins; };
 
--- a/Modules/Loadable/SubjectHierarchy/Widgets/qSlicerSubjectHierarchyPluginLogic.cxx
+++ b/Modules/Loadable/SubjectHierarchy/Widgets/qSlicerSubjectHierarchyPluginLogic.cxx
@@ -69,6 +69,8 @@ public:
 
   /// Menu shown when right-clicking in a slice or 3D view
   QMenu* ViewContextMenu;
+  /// Enable properties action
+  bool editPropertiesActionEnabled = true;
   /// Edit properties action
   QAction* EditPropertiesAction;
   /// Actions from the registered plugins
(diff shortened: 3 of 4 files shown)


jamesobutler:
Using the currently available allowed list of context menu actions not enough control for you? https://slicer.readthedocs.io/en/latest/developer_guide/script_repository.html#use-whitelist-to-customize-view-menu

riep:
I think his code is deprecated and is replaced by https://github.com/Slicer/Slicer/commit/62fe31fb179e5c904f5d8157739483dee6669a44

This is an interesting feature that I have looked into. For our application, it has the following limitations:
1. Cannot have empty list of actions.
2. Is reset after right click in a subject hierarchy tree view. (Actions on right click are filtered only after having selected the markup (with left click).

Modules/Loadable/SubjectHierarchy/Widgets/qSlicerSubjectHierarchyPluginLogic.cxx
```diff
@@ -683,6 +689,27 @@ void qSlicerSubjectHierarchyPluginLogic::addSupportedDataNodesToSubjectHierarchy
 void qSlicerSubjectHierarchyPluginLogic::registerViewContextMenuAction(QAction* action)
 {
   Q_D(qSlicerSubjectHierarchyPluginLogic);
+
+  if (action)
+    {
+    const int viewMenuActionsCount = d->ViewMenuActions.count();
+
+    for(int i = 0; i < viewMenuActionsCount; i++)
```
cpinter, line 697: Space between `for` and `(`

Modules/Loadable/SubjectHierarchy/Widgets/qSlicerSubjectHierarchyPluginLogic.cxx
```diff
@@ -683,6 +689,27 @@ void qSlicerSubjectHierarchyPluginLogic::addSupportedDataNodesToSubjectHierarchy
...
+      {
+      if(action == d->ViewMenuActions.at(i))
+        {
+        d->ViewMenuActions.removeAt(i);
+        break;
+        }
+      }
```
cpinter, line 704: Error handling would be nice

Modules/Loadable/SubjectHierarchy/Widgets/qSlicerSubjectHierarchyPluginLogic.cxx
```diff
@@ -69,6 +69,8 @@ class qSlicerSubjectHierarchyPluginLogicPrivate
 
   /// Menu shown when right-clicking in a slice or 3D view
   QMenu* ViewContextMenu;
+  /// Enable properties action
```
cpinter, line 72: Enable Edit properties action

Modules/Loadable/SubjectHierarchy/Widgets/qSlicerSubjectHierarchyPluginHandler.h
```diff
@@ -135,6 +135,10 @@ class Q_SLICER_MODULE_SUBJECTHIERARCHY_WIDGETS_EXPORT qSlicerSubjectHierarchyPlu
   /// \return True if plugin registered successfully, false otherwise
   Q_INVOKABLE bool registerPlugin(qSlicerSubjectHierarchyAbstractPlugin* plugin);
 
+  /// Unregister a plugin
+  /// \return True if plugin is unregistered successfully, false otherwise
+  Q_INVOKABLE bool unRegisterPlugin(QString name);
```
cpinter, line 140: Looking at other Qt classes the function name should be `unregisterPlugin`

Modules/Loadable/SubjectHierarchy/Widgets/qSlicerSubjectHierarchyPluginHandler.cxx
```diff
@@ -117,6 +117,17 @@ qSlicerSubjectHierarchyPluginHandler::~qSlicerSubjectHierarchyPluginHandler()
   delete this->m_DefaultPlugin;
 }
 
+//---------------------------------------------------------------------------
+bool qSlicerSubjectHierarchyPluginHandler::unRegisterPlugin(QString name)
+{
+  qSlicerSubjectHierarchyAbstractPlugin* currentPlugin = pluginByName(name);
+  foreach(QAction* action, currentPlugin->viewContextMenuActions())
+    {
+    this->m_PluginLogic->unRegisterViewMenuAction(action);
```
cpinter, line 126: This function does not actually unregister plugins, it only removes the view actions provided by the given plugin! There are much more to plugins than doing this  (there are other types of actions, there is a list of registered plugins in this class etc). Please look at what `registerPlugin` does (and the abstract SH plugin class to see what they provide and what needs to be removed) and finish this function. Or rename it to make it obvious what it does and what it doesn't.

Modules/Loadable/SubjectHierarchy/Widgets/qSlicerSubjectHierarchyPluginLogic.h
```diff
@@ -83,6 +83,8 @@ class Q_SLICER_MODULE_SUBJECTHIERARCHY_WIDGETS_EXPORT qSlicerSubjectHierarchyPlu
   void registerCorePlugins();
 
   /// Get all view context menu actions available
+  Q_INVOKABLE void setEditPropertiesEnabled(bool enable);
```
cpinter, line 86: If you have a setter function you need to add a getter function too. Also make it a Qt property.

cpinter (changes requested):
The main problem is that the unregister function does not do what its name suggests (it does not actually unregisters plugins). I also added minor comments.

cpinter:
In general, I think it would best to use the whitelist/blacklist feature as @jamesobutler suggests. If it has bugs, then generally it would be better to fix those bugs than adding a feature working around those bugs.

lassoan:
I agree, we should fix the existing mechanisms instead of introducing a new one. We can already customize/disable "Edit properties" at two granularity levels:
- for all items: use action whitelist
- for selected items (can be used for all items, too): create a plugin that owns the subject hierarchy item (by returning high confidence value for `canOwnSubjectHierarchyItem`) and then return `false` for `canEditProperties`.

Probably the whitelist method would be more appropriate for your use case. The [two limitations](https://github.com/Slicer/Slicer/pull/6299#issuecomment-1087440273) seem like bug/minor improvement that should not be hard to fix.

riep:
I will reopen this PR if needed

PR #6304, opened 2022-04-06
Improve python_package_updater to validate hints

jcfr:
Pair of hints are expected to be specified as CMake comment using the following
convention where `<auto-generated-requirements>` is replaced by the
`python_package_updater` script.

```
# [packagename]
<auto-generated-requirements>
# [/packagename]
```

This commit updates the script to always perform validation and exit
if there are errors.

The option "--validate" was also added to only perform validation without
updating external projects.

List of errors are formatted like this:

```
Searching external projects in /home/jcfr/Projects/ShapeVariationAnalyzer/SuperBuild
Validate external projects
/home/jcfr/Projects/ShapeVariationAnalyzer/SuperBuild/External_python-ShapeVariationAnalyzer-requirements.cmake
  joblib != packaging
```


Utilities/Scripts/python_package_updater.py
```diff
@@ -44,6 +45,60 @@ def parse_pip_list_output(packages_to_update):
...
+    """
+    return "\n".join([f"{filepath}\n  {first} != {second}" for filepath, first, second in mismatches])
+
+
+def validate_external_project_python_package_hints(directory):
+    """Check that external project files all have matching pair of hints.
+
+    Pair of hints are expected to be specified as CMake comment using the following
+    convention where ``<auto-generated-requirements>`` is replaced by the
+    ``python_package_updater`` script.
```
jcfr, line 67: I just realized this sentence may be ambiguous.

jamesobutler:
```suggestion
    convention where ``<auto-generated-requirements>`` is replaced with updated contents by using the
    ``python_package_updater`` script.
```
You thinking something like the above?

jamesobutler (commented):
I ran the updater for the Slicer repo and seem to get some false positives. I had tried the new validation by changing `[/Pillow]` to something else like `[/retrying]` which it detected correctly, but also detected some other things.

```
PS C:\S5R\python-install\bin> ./PythonSlicer.exe "C:\Slicer\Utilities\Scripts\python_package_updater.py"
Searching external projects in C:\Slicer\SuperBuild
Validate external projects
C:\Slicer\SuperBuild\External_OpenSSL.cmake
  1 !=
C:\Slicer\SuperBuild\External_OpenSSL.cmake
  3 !=
C:\Slicer\SuperBuild\External_python-dicom-requirements.cmake
  Pillow != retrying
```

jcfr:
Thanks for testing, the false positive are due to these comments, we need to make the regex more robust.

https://github.com/Slicer/Slicer/blob/ed4845ffca0adfe20d834065c4530d80d817ec49/SuperBuild/External_OpenSSL.cmake#L202-L224

jcfr:
I fixed the regex and also cherry-picked the commit `BUG: Fix python package updater when pip is out of date` from @jamesobutler originally associated with https://github.com/Slicer/Slicer/pull/6590

jcfr:
I tested locally that
* Changing `[/Pillow]` to something else like `[/retrying]` is detected correctly
* Specifying `--validate` allows to skip the update
* The comment from `External_OpenSSL.cmake` are not incorrectly considered as hint
* The updated External projects correspond to what was contributed in https://github.com/Slicer/Slicer/pull/6590

PR #6306, opened 2022-04-07
STYLE: Ignore more bulk changes for git blame

jamesobutler:
This adds some recent commits to the git blame ignore list to help skip over bulk change commits. Let me know if all of these commits are valid for this file, or only if was 100.0% STYLE changes (a few instances include changes to some logic).

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -10,3 +10,23 @@
 # $ git config blame.ignoreRevsFile .git-blame-ignore-revs
 #
 # Ignore changes introduced when doing global file format changes
+# STYLE: Trim trailing whitespace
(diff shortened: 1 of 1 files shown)


jcfr:
| Before | After |
|--|--|
| image | image |
| [Link][link-before] | [Link][link-after] |

[link-before]: https://github.com/Slicer/Slicer/blame/67295561c7f33b65bc763dd6282ecacd7e8c410d/Libs/MRML/Logic/vtkMRMLViewLinkLogic.cxx#L191

[link-after]: https://github.com/jamesobutler/Slicer/blame/git-avoid-bulk/Libs/MRML/Logic/vtkMRMLViewLinkLogic.cxx#L191

PR #6309, opened 2022-04-10
COMP: Update LibArchive version to address compile error on macOS

markasselin:
On macOS Monterey, Slicer compiles fail in LibArchive with the error:
error: variable 'detected_bytes' set but not used

This commit updates the LibArchive dependency version to v3.6.1 to include
upstream commit a4c3c90bb828ab5f01589718266ac5d3fdccb854 which addresses
this issue.


jcfr:
For future reference, see also https://github.com/Slicer/Slicer/pull/6390

Issue #6311, opened 2022-04-13
Segment Editor effect list stuck disabled


cpinter:
In some cases the effect list stays disabled even though it should be enabled.

## Steps to reproduce

- Load MRHead
- Right-click MRHead in Data module and choose "Segment this.."
- Add segment
- Right-click MRHead in Data module and choose "Segment this.."
- Add segment
- Editing tools are not enabled

## Expected behavior

The effects should be enabled

## Environment
- Slicer version: Slicer-4.13.0-2022-04-12
- Operating system: Windows 10

lassoan:
Can we close this (fixed by https://github.com/Slicer/Slicer/pull/6312)?

PR #6312, opened 2022-04-13
BUG: Fix stuck disable state in Segment Editor effect list

cpinter:
- When the new selected segment ID is the same as the old one, the necessary updates were sometimes not called, because the SetSelectedSegmentID method did not trigger updateWidgetFromMRML on the segment editor widget
- Fixed widget base class in ui file to prevent Qt Designer warning


cpinter:
Fixes https://github.com/Slicer/Slicer/issues/6311

lassoan (approved):
Looks good, thank you!

PR #6313, opened 2022-04-13
COMP: Allow building with C++20

jamesobutler:
Visual Studio 2019 16.11 now has the /std:c++20 switch
https://devblogs.microsoft.com/cppblog/msvc-cpp20-and-the-std-cpp20-switch/

This PR allows the standard to be set to C++20 as an option for testing. C++20 built Slicer on Windows 10 using Visual Studio 2022 with v143 toolset failed in some dependencies such as ITK as C++20 has removed some syntax options that were deprecated in earlier C++ versions. However, this PR allows to begin testing and finding areas where there are incompatibilities.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,8 +11,8 @@ endif()
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 message(STATUS "${_msg} - C++${CMAKE_CXX_STANDARD}")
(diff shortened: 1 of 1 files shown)


pieper:
Sounds like a good idea to me.

cpinter:
LGTM

jamesobutler:
Ok, if someone can provide the necessary approval to pass the required checks then we can merge this PR. It's been open for 2 months so looking to reduce the clutter of some Slicer PRs.

lassoan (approved):
Thank you for your contribution - and your patience!

PR #6314, opened 2022-04-13
Update minimum supported macOS versions

jamesobutler:
This PR aims to update the minimum supported platform versions for Slicer on macOS to be versions of macOS that they themselves are supported by Apple.

This will support the following macOS versions:
- 13.x: Ventura (Likely Oct 2022)
- 12.x: Monterey (2021)
- 11.x: Big Sur (2020)
- 10.15: Catalina (2019)

---------------------------------------------------------------
`COMP: Update minimum required CMAKE_OSX_DEPLOYMENT_TARGET to 10.14`:
Although Apple does not list an end-of-support date, macOS High Sierra (10.13) last received a security update in November 2020. https://endoflife.date/macos

macOS Mojave (10.14) hardware requirements
https://support.apple.com/kb/SP777

---------------------------------------------------------------
`COMP: Update minimum required CMAKE_OSX_DEPLOYMENT_TARGET to 10.15`:
Although Apple does not list an end-of-support date, macOS Mojave (10.14) last received a security update in July 2021. https://endoflife.date/macos

The supported Mac models are nearly identical between macOS 10.14 and macOS 10.15 with only the Mac Pro Mid 2010 and Mid 2012 being dropped.

macOS Catalina (10.15) hardware requirements
https://support.apple.com/kb/SP803

According to https://gs.statcounter.com/os-version-market-share/macos/desktop/worldwide as of August 2022 this would mean there would be support for 86.26% of Mac devices worldwide.


jamesobutler:
This also allows to continue using the latest version of `numpy` which now only provides macOS wheels for macOS 10.14 and later. 

In general it is becoming hardware for Python packages to create whls for older macOS versions because situations like Azure DevOps and GitHub Actions both having macOS 10.15 now as the oldest macOS agent.

jamesobutler:
The minimum macOS version was last updated approximately 2 years ago (https://github.com/Slicer/Slicer/commit/dede0266e8495a8337874d8bf39852a7b71e5ff4) when it was updated with support of versions 10.13-10.15.

pieper (approved):
Looks good, thanks 👍

jamesobutler:
@jcfr Do we have a build factory machine that supports this updated minimum now? re https://github.com/Slicer/Slicer/pull/6314#issuecomment-1177547745

-----------------------------------
In addition, macOS Catalina is now unsupported as well. In Slicer, do we still want to explicitly set the `CMAKE_OSX_DEPLOYMENT_TARGET]` to only supported versions of macOS?

Although Apple does not list an end-of-support date, macOS Catalina (10.15) last received a security update in July 2022. https://endoflife.date/macos

> on 12th September 2022 Apple released security updates for macOS Big Sur and Monterey, (and other platforms), but these updates did not include macOS Catalina - it has therefore reached its end of life.

jamesobutler:
As such, the factory-south-macos build has failed configuration for tonight’s build https://slicer.cdash.org/build/2824906/configure

jcfr:
I suggest we revert and integrate only after 5.2 is released. Factory-south-macos is running 10.13.6 and will be used to create the release package.

After 5.2 is released, we will the update system and xcode version on both factory-south-macos and computron.

pieper:
I like your plan Jc.  Here's a PR to revert: https://github.com/Slicer/Slicer/pull/6581

jcfr:
Thanks @pieper , turns out that I created the same PR at the time. I will be merging so that we can move forward.

At the same time, we are also working with our system admins to have factory updates installed on computron so that we can move the Slicer preview nightly there.

PR #6315, opened 2022-04-16
Two minor fixes and improvements

lassoan:


Libs/MRML/Core/vtkMRMLVolumeNode.cxx
```diff
@@ -1066,12 +1066,14 @@ void vtkMRMLVolumeNode::ApplyNonLinearTransform(vtkAbstractTransform* transform)
   double transformedBounds[6] = { 0.0,-1.0,0.0,-1.0,0.0,-1.0 };
   this->GetBoundsInternal(transformedBounds, rasToIJK, true);
   double spacing[3] = { 1.0, 1.0, 1.0 }; // output is specified in IJK coordinate system
-  reslice->SetOutputOrigin(transformedBounds[0] - 0.5 * spacing[0], transformedBounds[2] - 0.5 * spacing[1], transformedBounds[4] - 0.5 * spacing[2]);
+  // transformedBounds is computed so that it includes the voxel corners, therefore the origin is half voxel towards the image center
+  reslice->SetOutputOrigin(transformedBounds[0] + 0.5 * spacing[0], transformedBounds[2] + 0.5 * spacing[1], transformedBounds[4] + 0.5 * spacing[2]);
   reslice->SetOutputSpacing(spacing);
+  const double voxelExpandTolerance = 1e-3; // do not expand the volume with a new voxel if only expanding by 1/1000th of a voxel
```
jcfr, line 1072: Out of curiosity, is this an optimization or does it fix an actual issue ?

jcfr (approved):

Libs/MRML/Core/vtkMRMLVolumeNode.cxx
lassoan: It has been reported before that sometimes the volume size becomes one voxel larger after hardening a non-linear transform on it. If the bounds of the volume become even a tiny bit larger due to numerical imprecision then `ceil()` adds a complete voxel to the output image. The tolerance value ensures that a complete voxel is added only if the transformed bounds are at least 1/1000th voxel larger than the original bounds.

PR #6317, opened 2022-04-18
ENH : Markups line addon - draw an optional circle

chir-set:
@lassoan @pieper @coreDevs

Hi,

Following [this](https://discourse.slicer.org/t/draw-specific-dimension/22934) discussion, I came up with this patch that does the following :

 - use a markups line to draw a circle
 - use control point 0 as centre of the circle
 - set the circle's radius equal to the line length
 - use a vtkDiskSource to draw the circle
 - dynamically update the circle as control points are updated
 - orient the circle's axis orthogonal to the line
 - control the circle's visibility independently in 2D and 3D views
 
It is not yet controllable via UI, only by API. I don't have any clue how
to provide a UI control for this. Your suggestions are most welcome.

Please note that I'm still using a vtkDiskSource in 3D also, and not a
vtkSphereSource. This allows direct access to the markups line control points.
Moreover, orienting the disk's axis perpendicular to the line is quite
straightforward.

It is here proposed as a draft pull request, if you find it worthwhile of course. Else,
no problem at all, it has been a good opportunity to better understand things.
 
Please let me know how you consider this PR.

Regards.

### DIFF
--- a/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsLineNode.h
+++ b/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsLineNode.h
@@ -75,6 +75,20 @@ public:
   /// Return line length (distance between the two line endpoints) in world coordinate system.
   virtual double GetLineLengthWorld();
 
+  ///@{
+  /// Controls whether a circle should be drawn around point 0 in 2D, with radius being the line length
+  vtkGetMacro(DiskVisibility2D, int);
+  vtkSetMacro(DiskVisibility2D, int);
+  vtkBooleanMacro(DiskVisibility2D, int);
+  ///@}
+
+  ///@{
+  /// Controls whether a circle should be drawn around point 0 in 3D, with radius being the line length
+  vtkGetMacro(DiskVisibility3D, int);
+  vtkSetMacro(DiskVisibility3D, int);
+  vtkBooleanMacro(DiskVisibility3D, int);
+  ///@}
+
 protected:
   vtkMRMLMarkupsLineNode();
   ~vtkMRMLMarkupsLineNode() override;
@@ -83,6 +97,9 @@ protected:
 
   /// Calculates the handle to world matrix based on the current control points
   void UpdateInteractionHandleToWorldMatrix() override;
+
+  int DiskVisibility2D { 0 };
+  int DiskVisibility3D { 0 };
 };
 
 #endif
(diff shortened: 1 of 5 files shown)


jamesobutler (commented):
My first thought is that drawing a markup circle should be a new markup type and not hijacking the "Line" markup type. It would not be intuitive to me to select "Line" to then create a circle. 
image

As a new "Circle" (?) markup type you could have 2 placement options. This new markup would essentially be like a special case "ClosedCurve" markup type which would allow 2 points to generate a circle. To place the center point + outside ring defining the radius OR to place 2 points defining the diameter and a circle fit to that. These 2 options exist for the Scissors Segment Editor effect when drawing the Circle shape. That effect has a default option of defining the diameter. 
image

The measurement options enabled for this new markup would be "Diameter" and "Area".
image

lassoan:
We plan to add an ellipsoid option to markups ROI node. So, if you want to draw circle/ellipse in a slice view then you may help with the implementation of this new ROI mode.

If you want to experiment with customizing markups then I agree with @jamesobutler that it would be better not to change the line markup but add a new markup type. Instead, you can create your new markup node, widget, and representation classes and get them registered in the Markups module.

Currently, you can only define new markups in C++, but it would not be hard to add C++ classes that have methods that can be defined in Python (similarly to Segment Editor effects and Subject Hierarchy plugins). Then any user could easily add new markup types or just add new representations to existing markup types (such as a sphere or disk representation). Your contribution to this would be very welcome, too.

chir-set:
> drawing a markup circle should be a new markup type and not hijacking the "Line" markup type

I was sort of fearing this reply.

All markups derive from vtkMRMLMarkupsNode, vtkSlicerMarkupsWidget, vtkSlicerMarkupsWidgetRepresentation2D and it's 3D version. Creating a new markups from scratch, deriving from these classes, is not within my reach. I could theoretically derive from vtkMRMLMarkupsLineNode, vtkSlicerLineWidget etc... Then it might be a hacky work, unfit for Slicer.

As @lassoan believes similarly, I can therefore close this PR.

Thank you.

jamesobutler:
@chir-set You've seem to demonstrate that you can use a template (line markup) and are able to add new functionality to it. I think you may find that you don't have to create all the necessary classes from scratch as there is framework that can create template files for a new markup type. Using the "Extension Wizard" module in Slicer you can add a module that is of "loadablecustommarkups" type. Try this out just to see if you can work from this starting point.

image

Or as @lassoan indicated you may try to add a "Circle" mode option to the "ROI" markup type similar to what is planned for the "Ellipsoid" option for the "ROI" markup.

chir-set:
> Using the "Extension Wizard" module in Slicer you can add a module that is of "loadablecustommarkups" type.

I'll try this and slowly start tinkering again as possible. Thank you.

PR #6322, opened 2022-04-21
BUG: Fix issues with Markups plane interaction scale handles

Sunderlandkyl:
This commit solves 2 issues:
- When loading scenes with a plane of area 0.0mm, the scale handle points were not fully initialized, which could cause a crash. Fixed by always initializing the points, and instead hiding the scale handles during UpdateHandleVisibility for planes with area 0.0mm.
- Scale handles were not displayed at the incorrect locations in 3D while the plane was being moved. Fixed by calling Superclass::UpdateInteractionPipeline() before updating the scale handle, ensuring that the HandleToWorldTransform was updated before calculating scale handle position.

### DIFF
--- a/Modules/Loadable/Markups/VTKWidgets/vtkSlicerPlaneRepresentation3D.cxx
+++ b/Modules/Loadable/Markups/VTKWidgets/vtkSlicerPlaneRepresentation3D.cxx
@@ -330,6 +330,10 @@ void vtkSlicerPlaneRepresentation3D::UpdateFromMRML(vtkMRMLNode* caller, unsigne
 //----------------------------------------------------------------------
 void vtkSlicerPlaneRepresentation3D::UpdateInteractionPipeline()
 {
+  // Final visibility handled by superclass in vtkSlicerMarkupsWidgetRepresentation
+  // This needs be called before updating the handles, since it updates HandleToWorldTransform.
+  Superclass::UpdateInteractionPipeline();
+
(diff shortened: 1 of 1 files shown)


Modules/Loadable/Markups/VTKWidgets/vtkSlicerPlaneRepresentation3D.cxx
```diff
@@ -330,6 +330,9 @@ void vtkSlicerPlaneRepresentation3D::UpdateFromMRML(vtkMRMLNode* caller, unsigne
 //----------------------------------------------------------------------
 void vtkSlicerPlaneRepresentation3D::UpdateInteractionPipeline()
 {
+  // Final visibility handled by superclass in vtkSlicerMarkupsWidgetRepresentation
```
lassoan, line 333: Add a note in the comment that explains why this call must be made first.

lassoan (commented):
Thank you, these look good. I've just added a small suggestion to add in a comment why we need to put that line first (to ensure that the code will not be moved to incorrect location during future refactorings).

Modules/Loadable/Markups/VTKWidgets/vtkSlicerPlaneRepresentation3D.cxx
Sunderlandkyl: Thanks! I've added a comment.

lassoan (approved):
Looks perfect, thank you.

Issue #6287, opened 2022-03-29
Foreground image is replaced when reloading scene


rphellan:
Slicer replaces the foreground image with the wrong image when reloading a scene that contains views with foreground and background images.

## Steps to reproduce

Display MRHead_1 (foreground) and MRHead (background) on the red view of Slicer. This data is available as BuiltIn data.
Make sure that the "Link slice views" button is not activated.
It doesn't matter which layout is used to display the slice views.
Save the scene as one mrb file using the option “Create a Medical Record Bundle containing the scene”.
Reload the saved scene in a new Slicer.

![BugOnReload](https://user-images.githubusercontent.com/5794843/160697673-c215bc6b-d9f6-446f-bf38-1e752c187b4f.png)

## Expected behavior

I expect MRHead_1 (foreground) and MRHead (background) to appear on the red view of Slicer, but only one of them is set as foreground and background at the same time.

## Environment
- Slicer version: Slicer-4.13.0-2022-03-28
- Operating system: Microsoft Windows 10 Home

cpinter:
Hint by @lassoan: scene model of a node selector changes it

jcfr:
Are some relevant references:
* https://github.com/Slicer/Slicer/issues/4081 and topic [4081-ensure-reference-are-up-to-date-on-import](https://github.com/jcfr/SlicerGitSVNArchive/commits/4081-ensure-reference-are-up-to-date-on-import)
* https://github.com/Slicer/Slicer/issues/3462

cpinter:
I gave it a go, but in 2-3 hours could not get beyond the suspicion of @lassoan . However, I was able to confirm that it is the issue, that the scene model, while being populated, triggers selecting the incorrect node in the slice controller widget.

In this specific case, "Ensure reference are updated before NodeAddedEvent is invoked" (#4081) does not seem to apply, as the NodeAddedEvent is not invoked during or after a scene import I believe.

Let me share the relevant call stack, in case it helps:
<details>
<summary>Call stack of the set reference call that sets the incorrect volume node to the slice composite node</summary>c

>	MRMLCore.dll!vtkMRMLNode::SetAndObserveNthNodeReferenceID(const char * referenceRole, int n, const char * referencedNodeID, vtkIntArray * events) Line 1321	C++
 	MRMLCore.dll!vtkMRMLNode::SetNthNodeReferenceID(const char * referenceRole, int n, const char * referencedNodeID) Line 1311	C++
[...]

PR #6323, opened 2022-04-22
BUG: Fix infinite loop loading markups node under singleton transforms

Sunderlandkyl:
When loading a scene that contains markups transformed by singleton transform nodes it is possible to enter an infinite loop calling vtkMRMLTransformNode::GetTransformBetweenNodes(). This can happen if a circular loop is created between the transform nodes as their references are being updated.

Fixed by not processing TransformModifiedEvent events in vtkMRMLMarkupsNode during scene import.

### DIFF
--- a/Libs/MRML/Core/vtkMRMLTransformNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLTransformNode.cxx
@@ -527,6 +527,13 @@ void vtkMRMLTransformNode::GetTransformBetweenNodes(vtkMRMLTransformNode* source
     return;
     }
 
+  // If the number of transforms between the nodes exceeds the max depth threshold, then begin to search
+  // for duplicate transform nodes to ensure that the transform nodes don't contain a loop.
+  // See issue https://github.com/Slicer/Slicer/issues/6355.
+  int maxDepth = 100;
+  int currentDepth = 0;
+  std::set<vtkMRMLTransformNode*> visitedTransformNodes;
+
   if (sourceNode != nullptr && sourceNode->IsTransformNodeMyParent(targetNode))
     {
     // traverse the transform tree from bottom to top, from sourceNode to targetNode
@@ -536,6 +543,16 @@ void vtkMRMLTransformNode::GetTransformBetweenNodes(vtkMRMLTransformNode* source
       if (transformToParent)
         {
         transformSourceToTarget->Concatenate(transformToParent);
+        }
+
+      ++currentDepth;
+      if (currentDepth > maxDepth && !visitedTransformNodes.insert(current).second)
+        {
+        // Max depth exceeded and duplicate transform found. Loop detected.
+        // See issue https://github.com/Slicer/Slicer/issues/6355.
+        vtkGenericWarningMacro("vtkMRMLTransformNode::GetTransformBetweenNodes: Loop detected between transform nodes");
+        transformSourceToTarget->Identity();
+        break;
         }
       }
     }
@@ -548,6 +565,16 @@ void vtkMRMLTransformNode::GetTransformBetweenNodes(vtkMRMLTransformNode* source
       if (transformToParent)
         {
         transformSourceToTarget->Concatenate(transformToParent);
+        }
+
+      ++currentDepth;
+      if (currentDepth > maxDepth && !visitedTransformNodes.insert(current).second)
+        {
+        // Max depth exceeded and duplicate transform found. Loop detected.
+        // See issue https://github.com/Slicer/Slicer/issues/6355.
+        vtkGenericWarningMacro("vtkMRMLTransformNode::GetTransformBetweenNodes: Loop detected between transform nodes");
+        transformSourceToTarget->Identity();
+        break;
         }
       }
     // in transformSourceToTarget we have transform targetNode->sourceNode,


Modules/Loadable/Markups/MRML/vtkMRMLMarkupsNode.cxx
```diff
@@ -222,7 +222,7 @@ void vtkMRMLMarkupsNode::ProcessMRMLEvents(vtkObject *caller,
                                            unsigned long event,
                                            void *callData)
 {
-  if (caller != nullptr && event == vtkMRMLTransformableNode::TransformModifiedEvent)
+  if (this->Scene && !this->Scene->IsImporting() && caller != nullptr && event == vtkMRMLTransformableNode::TransformModifiedEvent)
```
lassoan, line 225: Should we also block the similar update in vtkMRMLMarkupsCurveNode::ProcessMRMLEvents?

lassoan (commented):
Thanks for working on this. It is somewhat concerning that this infinite loop can happen and the breaking the loop here may not address similar loops elsewhere. Can you provide a stack trace of the infinite loop (or a description of how to reproduce the issue)?

Sunderlandkyl:
@lassoan Here is a stack trace, specifically the infinite loop seems to be at [vtkMRMLTransformNode::L533](https://github.com/Slicer/Slicer/blob/d88bfaec0e884eb532c2d3c623a1eb59ba02cccd/Libs/MRML/Core/vtkMRMLTransformNode.cxx#L533-L540).

<details><summary>Stack trace</summary>
<pre>
[External Code]	
vtkCommon-9.1.dll!vtkTransformConcatenation::Concatenate(vtkAbstractTransform * trans) Line 444	C++
vtkCommon-9.1.dll!vtkGeneralTransform::Concatenate(vtkAbstractTransform * transform) Line 238	C++
MRMLCore.dll!vtkMRMLTransformNode::GetTransformBetweenNodes(vtkMRMLTransformNode * sourceNode, vtkMRMLTransformNode * targetNode, vtkGeneralTransform * transformSourceToTarget) Line 533	C++
vtkSlicerMarkupsModuleMRML.dll!vtkMRMLMarkupsNode::ProcessMRMLEvents(vtkObject * caller, unsigned long event, void * callData) Line 228	C++
MRMLCore.dll!vtkMRMLNode::MRMLCallback(vtkObject * caller, unsigned long eid, void * clientData, void * callData) Line 901	C++
vtkCommon-9.1.dll!vtkCallbackCommand::Execute(vtkObject * caller, unsigned long event, void * callData) Line 42	C++
MRMLCore.dll!vtkEventBroker::InvokeObservation(vtkObservation * observation, unsigned long eid, void * callData) Line 841	C++
MRMLCore.dll!vtkEventBroker::ProcessEvent(vtkObservation * observation, vtkObject * caller, unsigned long eid, void * callData) Line 713	C++
MRMLCore.dll!vtkEventBroker::Callback(vtkObject * caller, unsigned long eid, void * clientData, void * callData) Line 920	C++
vtkCommon-9.1.dll!vtkCallbackCommand::Execute(vtkObject * caller, unsigned long event, void * callData) Line 42	C++
vtkCommon-9.1.dll!vtkSubjectHelper::InvokeEvent(unsigned long event, void * callData, vtkObject * self) Line 651	C++
[Inline Frame] MRMLCore.dll!vtkObject::InvokeEvent(unsigned long) Line 230	C++
MRMLCore.dll!vtkMRMLNode::InvokePendingModifiedEvent() Line 557	C++
MRMLCore.dll!vtkMRMLScene::AddNodeNoNotify(vtkMRMLNode * n) Line 1252	C++
MRMLCore.dll!vtkMRMLScene::AddNode(vtkMRMLNode * n) Line 1336	C++
MRMLCore.dll!vtkMRMLScene::Import(vtkMRMLMessageCollection * userMessagesInput) Line 845	C++
MRMLCore.dll!vtkMRMLScene::ReadFromMRB(const char * fullName, bool clear, vtkMRMLMessageCollection * userMessagesInput) Line 3983	C++
qSlicerBaseQTCore.dll!qSlicerSceneBundleReader::load(const QMap<QString,QVariant> & properties) Line 93	C++
qSlicerBaseQTCore.dll!qSlicerCoreIOManager::loadNodes(const QString & fileType, const QMap<QString,QVariant> & parameters, vtkCollection * loadedNodes, vtkMRMLMessageCollection * userMessages) Line 587	C++
qSlicerBaseQTGUI.dll!qSlicerIOManager::loadNodes(const QString & fileType, const QMap<QString,QVariant> & parameters, vtkCollection * loadedNodes, vtkMRMLMessageCollection * userMessages) Line 446	C++
qSlicerBaseQTGUI.dll!qSlicerIOManager::loadNodes(const QList<QMap<QString,QVariant>> & files, vtkCollection * loadedNodes, vtkMRMLMessageCollection * userMessages) Line 468	C++
qSlicerBaseQTGUI.dll!qSlicerDataDialog::exec(const QMap<QString,QVariant> & readerProperties) Line 679	C++
qSlicerBaseQTGUI.dll!qSlicerIOManager::execDelayedFileDialog() Line 363	C++
[External Code]	
qSlicerBaseQTGUI.dll!qSlicerApplication::notify(QObject * receiver, QEvent * event) Line 426	C++
[External Code]	
qSlicerBaseQTCore.dll!qSlicerCoreApplication::exec() Line 929	C++
SlicerApp-real.exe!`anonymous namespace'::SlicerAppMain(int argc, char * * argv) Line 62	C++
[External Code]	
	
</pre>
</details>

And here is a Gist containing the code to reproduce the issue: https://gist.github.com/Sunderlandkyl/4aaaa963da346b1adfaa62781e2ac685

lassoan:
Thanks a lot Kyle, the autóval information and script that reproduces the issue are very useful.

As a short-term fix, the currently proposed change could be integrated. My main concern is that we would ignore transform modification events during scene importing and I'm not sure that after importing is finished, what guarantees that interaction handles and measurements are up-to-date according to the transforms. Can you check if an update is called with a reliable mechanism at the end of the import? 

This error demonstrated that the transform concatenation loop may hang, therefore we should add a loop detection mechanism to it. Since transforms may be retrieved very frequently, and this hang almost never happens, this should be an efficient mechanism, that works with near-zero overhead when there is no loop. For example, we could have a counter and if we reach a large number (e.g., 100) and then start remembering already used node pointers and if we find a recurring one then log the error and return. Could you implement this (in this or a follow-up pull request)? 

It seems that there could be similar issues (any module trying to get transform between nodes while the scene is loading) elsewhere, so for a reassuring fix we should probably rework scene loading. Loading all nodes into a new temporary scene, changing conflicting node IDs, then moving the nodes into the main scene would probably solve this issue and probably some other scene loading related issues, too (for example, some subject hierarchy update related problems). Would you have time to work on this in the next couple of weeks?

Sunderlandkyl:
I did some more testing, and you're right the transform update is not called after scene import is finished. Some markup representations are fine, but curves are not. I'll close this PR and follow up with another one containing a fix addressing the underlying issue.

I'll start working on another solution. Do you think that we should implement both approaches (loop detection + scene loading rework), or just one?

lassoan:
In the short term, loop detection is sufficient.

Making node ID resolution more robust on scene import is long overdue. This means that we can push it further, but it also means that we should really do it. Now that the new stable version is going to be released is a good time to make such complex changes. So, I think we should do it within 1-2 months.

Sunderlandkyl:
Ok, I've updated the pull request to instead check for loops in GetTransformBetweenNodes. Detection of duplicate transforms doesn't start until we reach a depth of 100 transform nodes.

Tests are still running.

jcfr:
Great detective work 👌

Should we consider this for integration in `maintenance/5.0` branch ?

jcfr:
Should we update the warning message to ask user to provide details in a specific issue or discourse post ?

Sunderlandkyl:
Tests seem fine. The change isn't affecting the passing/failing of tests on my computer.

lassoan (commented):
The implementation is very nice and compact. Could you submit a new issue where you describe the problem, add the script+description that reproduces it, and describe that scene loading should be made more robust by resolving all node ID conflicts before inserting nodes into the scene? We should then refer to this issue in a comment here that infinite loops may happen due to bugs like this.

jcfr:
> describe that scene loading should be made more robust by resolving all node ID conflicts before inserting nodes into the scene

Wondering if this is similar to https://github.com/Slicer/Slicer/issues/6287#issuecomment-1096786565, I will look into rebasing [jcfr/Slicer@4081-ensure-reference-are-up-to-date-on-import](https://github.com/jcfr/SlicerGitSVNArchive/compare/add-vtkEventSpy...jcfr:4081-ensure-reference-are-up-to-date-on-import#diff-613613dab40347f3b905957c91fded53R653) introducing the function `vtkMRMLScene::PrepareImport`

Sunderlandkyl:
Created an issue here: https://github.com/Slicer/Slicer/issues/6355.

Sunderlandkyl:
Ok, done.

Just a thought, but would loading the nodes into a temporary scene, and then copying them into the main scene not cause the same issue? If the nodes are copied one by one into the main scene, then would it not be possible to still create a circular transform reference depending on the order that the nodes are copied?

lassoan:
> Just a thought, but would loading the nodes into a temporary scene, and then copying them into the main scene not cause the same issue?

An unresolved reference returns a null transform node, which stops the iteration on the transform hierarchy, so worst case it results in a temporarily incorrect transform. When the parent transform node is added to the scene then this triggers a transform update and the temporarily incorrect transform gets fixed.

Normally nodes and modules don't try to resolve node references during scene loading, so this temporary change in transform should not happen, but we see that there are complicated situations, so it is hard to guarantee that it really never happens. Loading into a separate scene ensures that it cannot lead to establishing incorrect references between nodes.

lassoan (approved):
It looks good to me. Thank you!

PR #6324, opened 2022-04-22
BUG: Fix saving of scalar volume sequence with only one frame

Sunderlandkyl:
When trying to save a volume sequence with a single frame, vtkTeemNRRDWriter raises the following error:
```
[nrrd] nrrdSave:
[nrrd] nrrdWrite: trouble
[nrrd] _nrrdWrite:
[nrrd] nrrdCheck: trouble
[nrrd] _nrrdCheck: trouble with space field
[nrrd] _nrrdFieldCheck_space: trouble
[nrrd] _nrrdFieldCheckSpaceInfo: axis[0] has a direction vector, and so can't have min, max, spacing, or units set
```

This is because the range dimension is omitted by the writer when the size of the first dimension is 1. This results in an error when units are instead assigned to the first spatial dimension.
Fixed by adding an option (vtkTeemNRRDWriter::ForceRangeAxis) to force the writer to include a range axis, even when the size of the first dimension is 1. This option remains disabled by default, and is only enabled by vtkMRMLVolumeSequenceStorageNode.

### DIFF
--- a/Libs/MRML/Core/vtkMRMLVolumeSequenceStorageNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLVolumeSequenceStorageNode.cxx
@@ -425,6 +425,10 @@ int vtkMRMLVolumeSequenceStorageNode::WriteDataInternal(vtkMRMLNode *refNode)
   // Use here the NRRD Writer
 #if Slicer_VERSION_MAJOR > 4 || (Slicer_VERSION_MAJOR == 4 && Slicer_VERSION_MINOR >= 9)
   vtkNew<vtkTeemNRRDWriter> writer;
+  // ForceRangeAxis needs to be emabled for the writer to correctly write image sequences that contain only a single frame.
+  // Without this option the range axis would be omitted for single frame sequences, causing the first axis to be a spatial dimension, rather than range.
+  // This would cause an error when attempting to set units on the first axis.
+  writer->SetForceRangeAxis(true);
   writer->SetVectorAxisKind(nrrdKindList);
 #else
   vtkNew<vtkNRRDWriter> writer;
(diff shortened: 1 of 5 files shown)


Libs/vtkTeem/vtkTeemNRRDWriter.h
```diff
@@ -96,6 +96,11 @@ class VTK_Teem_EXPORT vtkTeemNRRDWriter : public vtkWriter
   void vtkSetSpaceToLPS()  { this->SetSpace(nrrdSpaceLeftPosteriorSuperior); };
   void vtkSetSpaceToLPST() { this->SetSpace(nrrdSpaceLeftPosteriorSuperiorTime); };
 
+  /// Force the addition of a range axis, even when the size of the first image dimension (components, or frame list) is 1.
```
lassoan, line 99: Describe why this is useful

Libs/MRML/Core/vtkMRMLVolumeSequenceStorageNode.cxx
```diff
@@ -425,6 +425,7 @@ int vtkMRMLVolumeSequenceStorageNode::WriteDataInternal(vtkMRMLNode *refNode)
   // Use here the NRRD Writer
 #if Slicer_VERSION_MAJOR > 4 || (Slicer_VERSION_MAJOR == 4 && Slicer_VERSION_MINOR >= 9)
   vtkNew<vtkTeemNRRDWriter> writer;
+  writer->SetForceRangeAxis(true);
```
lassoan, line 428: Describe in a comment why this is required. It would be even better to add a simple test for it.

Sunderlandkyl:
Thanks! I've added the comments, and some additional code to test writing/reading in vtkMRMLSequenceStorageNodeTest1.cxx.

lassoan (approved):
Excellent, thank you. It is very nice that you added the test, it'll make sure this will not break.

lassoan:
@Sunderlandkyl it seems that the new test is failing on linux and mac, could you have a look?

https://slicer.cdash.org/test/22148750

Issue #5916, opened 2021-09-29
Markups: Documentation of Markups module has not been updated to reflect latest changes


lassoan:
Markups module documentation (https://slicer.readthedocs.io/en/latest/user_guide/modules/markups.html#) needs an update after the recent improvements.

When everything is settled then a Slicer forum post must be submitted in the "Support" category with the "feature" tag attached that describes the new features strictly from a user standpoint. No implementation details, only description of new features and changes. Need to add screenshots and embed YouTube videos in the page.

## Environment
- Slicer version: Slicer-4.11.0-2021-09-27
- Operating system: Windows

@smrolfe

jamesobutler:
Appears that there are several screenshots to update here as it relates to the new Markups toolbar that will be needed prior to stable release.

jamesobutler:
@lassoan Do we need to add this to the 4.13 milestone so the documentation matches the new Slicer 5.0 release? I remembered about this issue when linking to it in https://discourse.slicer.org/t/draw-specific-dimension/22934/3?u=jamesobutler.

PR #6326, opened 2022-04-24
DOC: Add Markups toolbar info to module documentation

jamesobutler:
re #5916, this updates the Markups module documentation with latest updates and images regarding the Markups toolbar.


lassoan (approved):
This looks good, thanks a lot.

PR #6327, opened 2022-04-25
COMP: Define Slicer icon for package and installed details

### LINKED ISSUE #5885: Define icons for installer package and for installed application entry
Currently the Slicer installer executable uses a generic NSIS icon and also defines no icon for the entry in the installed programs list.
image
image

Define the appropriate installer icon and to also define the registry key DisplayIcon entry. The following documentation may help.
https://cmake.org/cmake/help/latest/cpack_gen/nsis.html?highlight=nsis%20icon

jamesobutler:
This closes #5885:

## Installer icon:
image

## Installed icon:
image
image

### DIFF
--- a/CMake/SlicerCPack.cmake
+++ b/CMake/SlicerCPack.cmake
@@ -390,6 +390,10 @@ if(CPACK_GENERATOR STREQUAL "NSIS")
   set(PACKAGE_APPLICATION_NAME "${APPLICATION_NAME} ${CPACK_PACKAGE_VERSION}")
   slicer_verbose_set(CPACK_PACKAGE_EXECUTABLES "..\\\\${EXECUTABLE_NAME}" "${PACKAGE_APPLICATION_NAME}")
 
+  get_property(${app_name}_CPACK_NSIS_MUI_ICON GLOBAL PROPERTY ${app_name}_WIN_ICON_FILE)
+  slicer_cpack_set("CPACK_NSIS_MUI_ICON")
+  slicer_verbose_set(CPACK_NSIS_INSTALLED_ICON_NAME "${app_name}.exe")
+
   # -------------------------------------------------------------------------
   # File extensions
   # -------------------------------------------------------------------------


CMake/SlicerCPack.cmake
```diff
@@ -338,8 +338,8 @@ slicer_cpack_set("CPACK_RESOURCE_FILE_LICENSE")
 get_property(${app_name}_CPACK_PACKAGE_DESCRIPTION_SUMMARY GLOBAL PROPERTY ${app_name}_DESCRIPTION_SUMMARY)
 slicer_cpack_set("CPACK_PACKAGE_DESCRIPTION_SUMMARY")
 
-get_property(${app_name}_CPACK_PACKAGE_ICON GLOBAL PROPERTY ${app_name}_APPLE_ICON_FILE)
 if(APPLE)
+  get_property(${app_name}_CPACK_PACKAGE_ICON GLOBAL PROPERTY ${app_name}_APPLE_ICON_FILE)
```
jcfr, line 342: Since the variable `APPLE_ICON_FILE` is expected by `slicerMacroBuildApplication`, I suggest to keep this code as is.

https://github.com/Slicer/Slicer/blob/224995ba039f5bc540197a4aa94e393164106d0a/Applications/SlicerApp/CMakeLists.txt#L66-L78

CMake/SlicerCPack.cmake
```diff
@@ -390,6 +390,15 @@ if(CPACK_GENERATOR STREQUAL "NSIS")
   set(PACKAGE_APPLICATION_NAME "${APPLICATION_NAME} ${CPACK_PACKAGE_VERSION}")
   slicer_verbose_set(CPACK_PACKAGE_EXECUTABLES "..\\\\${EXECUTABLE_NAME}" "${PACKAGE_APPLICATION_NAME}")
 
+  if(WIN32)
+    get_property(${app_name}_CPACK_NSIS_MUI_ICON GLOBAL PROPERTY ${app_name}_WIN_ICON_FILE)
+    slicer_cpack_set("CPACK_NSIS_MUI_ICON")
+  elseif(APPLE)
+    get_property(${app_name}_CPACK_NSIS_MUI_ICON GLOBAL PROPERTY ${app_name}_APPLE_ICON_FILE)
+    slicer_cpack_set("CPACK_NSIS_MUI_ICON")
```
jcfr, line 398: Since NSIS installer is specific to windows, how about removing that part ?

CMake/SlicerCPack.cmake
```diff
@@ -390,6 +390,15 @@ if(CPACK_GENERATOR STREQUAL "NSIS")
...
+  endif()
+  slicer_verbose_set(CPACK_NSIS_INSTALLED_ICON_NAME "${app_name}.exe")
```
jcfr, line 400: I suggest to move this statement inside the `if(WIN32)` clause.

CMake/SlicerCPack.cmake
jamesobutler: Ok I've reverted the change.

CMake/SlicerCPack.cmake
```diff
@@ -390,6 +390,12 @@ if(CPACK_GENERATOR STREQUAL "NSIS")
   set(PACKAGE_APPLICATION_NAME "${APPLICATION_NAME} ${CPACK_PACKAGE_VERSION}")
   slicer_verbose_set(CPACK_PACKAGE_EXECUTABLES "..\\\\${EXECUTABLE_NAME}" "${PACKAGE_APPLICATION_NAME}")
 
+  if(WIN32)
+    get_property(${app_name}_CPACK_NSIS_MUI_ICON GLOBAL PROPERTY ${app_name}_WIN_ICON_FILE)
+    slicer_cpack_set("CPACK_NSIS_MUI_ICON")
+    slicer_verbose_set(CPACK_NSIS_INSTALLED_ICON_NAME "${app_name}.exe")
+  endif()
```
jcfr, line 397:
```suggestion
  # Icon filename (*.ico) used as the main icon for the generated install program.
  get_property(${app_name}_CPACK_NSIS_MUI_ICON GLOBAL PROPERTY ${app_name}_WIN_ICON_FILE)
  slicer_cpack_set("CPACK_NSIS_MUI_ICON")

  # Relative path to the executable that contains the installer icon.
  slicer_verbose_set(CPACK_NSIS_INSTALLED_ICON_NAME "${app_name}.exe")
```

jcfr: Since the code is already enclosed in `if(CPACK_GENERATOR STREQUAL "NSIS")`, it turns out that adding the `if(WIN32)` is not needed at all.

jcfr: If this change looks good, it should then be good for integration.

PR #6332, opened 2022-04-26
BUG: Fix remote cache directory in vtkMRMLSequenceStorageNodeTest1

Sunderlandkyl:
Because the remote cache directory was not specified for the scene in vtkMRMLSequenceStorageNodeTest1, attempting to unzip and read the seq.mrb would fail on Linux and Mac as the application would not have permission to write to the directory.

Fixed by setting the remote cache directory to the temporary testing directory.


Sunderlandkyl:
Haven't been able to test it yet on Linux, but I suspect that this is the issue from the [failing test](https://slicer.cdash.org/test/22148750).

lassoan (approved):
Thanks for the quick fix, it looks good!

lassoan:
The fix worked!

image

PR #6335, opened 2022-04-27
BUG: Fix Slicer crash upon hiding delete all control points option

jamesobutler:
Slicer would crash when following this example in the Slicer developer guide documentation.
https://github.com/Slicer/Slicer/blob/a61c805427e9af8cef631ce6e5ff2e297c3a3325/Docs/developer_guide/api.md?plain=1#L49-L56

This was not caught by the Markups widget test because a node is already set to the widget upon customizing the delete all control points action.
https://github.com/Slicer/Slicer/blob/a61c805427e9af8cef631ce6e5ff2e297c3a3325/Modules/Loadable/Markups/Widgets/Testing/Python/MarkupsWidgetsSelfTest.py#L162-L176


jamesobutler:
Not sure why the slicer/apidocs status check has been failing recently across various PRs. Thoughts @jcfr ?

lassoan (approved):
Good catch, thank you!

PR #6340, opened 2022-04-28
BUG: Fix deprecation warning suggesting method not actually used

jamesobutler:
I noticed this mismatch in the suggestion versus what was actually being used while I was updating some old code to use newer API.

This mismatch of the recommendation versus what is actually used was introduced in https://github.com/Slicer/Slicer/commit/e667de230fb41aca7efa576ebcdeb2621f300aa5.


jamesobutler:
@lassoan I've also included a commit to fix a spelling mistake that was introduced in https://github.com/Slicer/Slicer/commit/3d424ec5cc760446f2c247aaee3294b2e518d691 and causing codespell checks to fail.

PR #6341, opened 2022-04-29
ENH: Add slicer.util.chdir context manager

jcfr:
This commit implements a non thread-safe context manager to change
the current working directory.

Available in Python 3.11 as ``contextlib.chdir`` and adapted from
https://github.com/python/cpython/pull/28271.


jcfr:
Local testing

| image | image | image | image |
|--|--|--|--|
| Windows (overload):heavy_check_mark: | Linux (metroplex) :heavy_check_mark: | macOS (factory-south-macos):heavy_check_mark: | macOS (computron):heavy_check_mark: |

PR #6348, opened 2022-05-03
BUG: Fix on exit Segfault ensuring ExtensionsManagerDialog is deleted once

jcfr:
This commit applies a fix similar to acc618074 (BUG: Segfault at application
exit due to SettingsDialog deletion) and fixes a regression likely introduced
in abd2544d4 (BUG: Fixed application window not clickable when popup is
displayed) where the extension manager instance is associated with the
parent window.

See https://discourse.slicer.org/t/help-for-package-testing-on-macos-monterrey-or-bigsur-system/23249


lassoan (approved):
It looks good!

PR #6349, opened 2022-05-03
ENH: add nearest neighbor interpolation mode to labelmapvolume

fbordignon:
Hardening a non-linear transform on a labelmap results in a linear interpolation of the labels which is undesired.
Relative to [issue discussed on the forums](https://discourse.slicer.org/t/landmark-registration-transform-resampling/23243)

### DIFF
--- a/Libs/MRML/Core/vtkMRMLVolumeNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLVolumeNode.cxx
@@ -1050,7 +1050,19 @@ void vtkMRMLVolumeNode::ApplyNonLinearTransform(vtkAbstractTransform* transform)
     }
 
   reslice->SetInputConnection(this->ImageDataConnection);
-  reslice->SetInterpolationModeToLinear();
+
+  // GetResamplingInterpolationMode does not use VTK_RESLICE... constants because it is an implementation
+  // detail that currently vtkImageReslice is used for resampling.
+  int resamplingMode = this->GetResamplingInterpolationMode();
+  switch (resamplingMode)
+    {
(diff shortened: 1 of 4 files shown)


pieper:
This looks like a good start.  Do you plan to add support for this to `vtkMRMLVolumeNode::ApplyTransform`?

Libs/MRML/Core/vtkMRMLVolumeNode.cxx
```diff
@@ -1050,7 +1050,7 @@ void vtkMRMLVolumeNode::ApplyNonLinearTransform(vtkAbstractTransform* transform)
     }
 
   reslice->SetInputConnection(this->ImageDataConnection);
-  reslice->SetInterpolationModeToLinear();
+  reslice->SetInterpolationMode(this->GetResliceInterpolationMode());
```
fbordignon, line 1053: @pieper I've tried to do that over here.

pieper: Ah, I see, I thought you'd only added the accessors.

pieper (dismissed):
LGTM 👍

lassoan:
Need to add `override`, fix a few style inconsistencies and I would use `resample` instead of `reslice`. I'll push a small update with these.

fbordignon:
Good catch on the override, I completely forgot about it.
I figured that keeping the reslice name would hint anyone overriding this method that it needs to return a valid vtk reslice interpolation mode. I thought about doing some kind of enum but ended up not trying to over engineer it.

lassoan:
I've force-pushed my touchups to this branch, please have a look.

PR #6358, opened 2022-05-07
ENH: Increase maximum number of colors loaded from colortable files

### LINKED ISSUE #6354: Problem loading labelmap color table with lots of labels
A labelmap with 50k labels loads but the color table is messed up.
I figure it is because of:
https://github.com/Slicer/Slicer/blob/5b87a6abf59bc37d0d4baf7362a91e9e8dd4f188/Libs/MRML/Core/vtkMRMLColorTableStorageNode.cxx#L34-L35
Why this limit was imposed? Was it taking too long to load an invalid color table file?
I guess the error check to be done here is to prevent overflowing the `vtkMRMLColorTableNode` number of colors (int). Am I correct in assuming that?

lassoan:
File loading performance is fixed by implementing bulk setting of color values.

Increased maximum number of accepted entries to 1 million as requested in #6354.
Loading of a file with 1 million as the maximum ID takes only a few seconds.
Colors module performance is not great with such huge tables. It takes about 20 seconds
to display a color table with 1 million entries.

fixes #6354

### DIFF
--- a/Libs/MRML/Core/vtkMRMLColorTableNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLColorTableNode.cxx
@@ -1254,26 +1254,18 @@ void vtkMRMLColorTableNode::SetNumberOfColors(int n)
   if (this->GetType() != this->User &&
       this->GetType() != this->File)
     {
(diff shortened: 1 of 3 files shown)


pieper (approved):
Great, thanks for doing this 👍

PR #6359, opened 2022-05-08
BUG: vtk_to_numpy funciton missing in Endoscopy

gaoyi:
The vtk_to_numpy function call was missing. The bug was introduced in a258951f473d74f24a93cff95673ea6163a7b5e1, and it's fixed.

### DIFF
--- a/Modules/Scripted/Endoscopy/Endoscopy.py
+++ b/Modules/Scripted/Endoscopy/Endoscopy.py
@@ -365,7 +365,7 @@ class EndoscopyComputePath:
       if originalPointsPerSegment<pointsPerSegment:
         self.fids.SetNumberOfPointsPerInterpolatingSegment(originalPointsPerSegment)
       # Get it as a numpy array as an independent copy
-      self.path = vtk.util.numpy_support(resampledPoints.GetData())
+      self.path = vtk.util.numpy_support.vtk_to_numpy(resampledPoints.GetData())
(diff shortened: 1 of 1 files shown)


lassoan (commented):
Good catch, thanks a lot for the fix.Would you mind updating the commit comment: funciton->function and add `fixes https://github.com/Slicer/Slicer/issues/6360` to the end?

gaoyi:
Hi Andras,  @lassoan 

Thanks. Will do these two changes.

I noticed that there are new commits in the master branch and could i know the right way to do this merging:

1. checkout the master branch and do everything from beginning and submit a new PR. (and you will delete this PR?)

or

2. based on the current status, I merge the new commits from the master branch... but git will auto-generate a "Merge ..." commit message, which is not compatible with Slicer's commit message requirement. Should I edit that "Merge..." message? I don't think so though...

Thanks!

Best,
yi

pieper:
Hi @gaoyi - you could rebase your branch against the current master and then add your commit.  But also, as admins we have the option to perform the rebase for you as part of the merge process, so I think you don't need to bother and it will end up the same either way.  Thanks for the fixes 👍

gaoyi:
> 

Thanks Steve!

I've created another branch to the current master at:
https://github.com/Slicer/Slicer/pull/6362

Should I close this?

Best,
yi

Issue #6360, opened 2022-05-08
BUG: in Endoscopy module


gaoyi:
[Give a short summary of what the problem is. Note: Before submitting an error report, please test if the problem is reproducible in the latest Slicer Preview Release. If the problem does not occur in the latest Slicer Preview Release then most likely it has been already fixed and there is no need to report it again.]

## Steps to reproduce

The call to vtk_to_numpy function call was removed in a258951f473d74f24a93cff95673ea6163a7b5e1 by mistake. Causing path generation to fail.

It's a small bug. I've added it back and created an PR.

## Expected behavior

Things should be back to normal.

## Environment
- Slicer version: Slicer-4.13.0-2022-05-08
- Operating system: Linux + Ubuntu 22.04

jamesobutler:
Closed by https://github.com/Slicer/Slicer/commit/3d66964c524a58b2583aac051e1310a5019d0039.

PR #6362, opened 2022-05-09
BUG: vtk_to_numpy function missing in Endoscopy

gaoyi:
the bug was introduced in a258951f473d74f24a93cff95673ea6163a7b5e1, and it's fixed here

corresponding to the Issue:

https://github.com/Slicer/Slicer/issues/6360

### DIFF
--- a/Modules/Scripted/Endoscopy/Endoscopy.py
+++ b/Modules/Scripted/Endoscopy/Endoscopy.py
@@ -365,7 +365,7 @@ class EndoscopyComputePath:
       if originalPointsPerSegment<pointsPerSegment:
         self.fids.SetNumberOfPointsPerInterpolatingSegment(originalPointsPerSegment)
       # Get it as a numpy array as an independent copy
-      self.path = vtk.util.numpy_support(resampledPoints.GetData())
+      self.path = vtk.util.numpy_support.vtk_to_numpy(resampledPoints.GetData())
       return
 
     # hermite interpolation functions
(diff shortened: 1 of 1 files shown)


jcfr:

@gaoyi Could you also associated the email `gy99@tsinghua.org.cn` with your GitHub account ? That way your contribution will be linked with your account.

If not, we could also amend the commit to use your other email `Yi Gao <gaoyicn@outlook.com>`

Let us know what you would prefer.

jcfr:
@lassoan @pieper That seems to be a good candidate for backporting to the 5.0 maintenance branch and generate `v5.0.3`

pieper:
> backporting to the 5.0 maintenance branch and generate `v5.0.3`

I think yes but let's give it a week or so to see if here are other patches we want to include.

lassoan:
Yes, creating 1-2 patch releases a couple of weeks apart sounds like a good plan to me. Do we just push the commits we want to include in the next patch release into the maintenance branch?

jcfr:
> Do we just push the commits we want to include in the next patch release into the maintenance branch?

If the changes have been tested, directly pushing makes sense.

If not, proposing a pull-request is sensible.

Either way, the cherry-picked should be amended to include text like this after the title of the commit and before the body:

```
(cherry picked from commit <SHA[:9]>, PR #<N>)
```

For example, see https://github.com/Slicer/Slicer/commit/5a16fb6493ef31f48f3d473bb5c1ce5ab5f0a960

lassoan:
Is this ready to be merged?

PR #6363, opened 2022-05-11
Fix python launcher environment in custom app 

fbordignon:
`PythonLauncherSettings.ini` reference to `SlicerLauncherSettings.ini` was hardcoded.
This was bad for custom apps, preventing PythonSlicer to run inside the proper context. 

Loading libs with PythonSlicer - such as vtk - was failing in a custom app built with SlicerCAT. I've confirmed this issue on SlicerSALT [prebuilt binaries](https://data.kitware.com/#item/5f18b8539014a6d84e32ab24) 

1- Run PythonSlicer
2 - import vtk
>>> import vtk
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'vtk'


lassoan (commented):
Thank you, this looks good to me. Have you tested this change with your custom application to confirm that the fix is required and sufficient?

jamesobutler (commented):
I can confirm that the issue happened only for Slicer custom apps.
```
PS C:\Users\JamesButler\AppData\Local\NA-MIC\Slicer 5.1.0-2022-05-08\bin> ./PythonSlicer.exe
Python 3.9.10 (main, May  8 2022, 23:28:56) [MSC v.1930 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import vtk
>>>
```

```
PS C:\Program Files\MyApp 2.0.0-2022-05-11\bin> ./PythonSlicer.exe
Python 3.9.10 (main, May 10 2022, 10:58:36) [MSC v.1931 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import vtk
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'vtk'
>>>
```

This type of launcher settings issue seems like it may be related to https://discourse.slicer.org/t/slicer-custom-app-crashing-with-testing-arg/23209. @fbordignon Can you test with this change to see what happens when you run `./MyApp.exe --testing`?

fbordignon:
> Have you tested this change with your custom application to confirm that the fix is required and sufficient?

Yes, this change yields a reference to the custom application name and properly sets up the context for PythonSlicer to run.

fbordignon:
> Can you test with this change to see what happens when you run `./MyApp.exe --testing`?

I got the same error as reported in the forums. This change seems to be unrelated or insufficient to solve the issue you linked.

jcfr:
Thanks for the patch, that seems sensible. This is a follow up of https://github.com/Slicer/Slicer/commit/94080fe7af34ae9afe16f4cd055bf946168777c9

Reviewing the commit, it looks there are occurrences of `Slicer` (e.g in status message) that would also need to be updated.

@fbordignon  Could you review these and further update your changes ?

fbordignon:
@jcfr please take a look at the latest commit. There are still some occurrences of Slicer but I believe it is better to keep it like that

jcfr (approved):

jcfr (approved):

PR #6365, opened 2022-05-12
COMP: Do not require building extension from git repository

lassoan:
Developers often complained that their extension build failed for some unknown reason and frequently the issue was that they did not store the source code in a git repository.

It is good to encourage developers to use best practices, such as version control, but making the build fail if they build from a non-version-controlled repository was unnecessarily strong measure.

This commit changes the behavior so that if an extension is built from a non-git folder then the build succeeds, just a warning is displayed during project configuration step and `NA` is used as the extension's WC revision.

### DIFF
--- a/CMake/SlicerMacroExtractRepositoryInfo.cmake
+++ b/CMake/SlicerMacroExtractRepositoryInfo.cmake
@@ -72,7 +72,13 @@ macro(SlicerMacroExtractRepositoryInfo)
 
   if(NOT EXISTS ${MY_SOURCE_DIR}/.git)
 
-    message(AUTHOR_WARNING "Skipping repository info extraction: directory [${MY_SOURCE_DIR}] is not a GIT or SVN checkout")
+    if(DEFINED ${wc_info_prefix}_WC_REVISION)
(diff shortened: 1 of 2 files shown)


pieper:
This makes sense to me - I'll leave it to @jcfr to decide if it's a good move.

jcfr (changes requested):
Since directly updating `CMake/SlicerMacroExtractRepositoryInfo.cmake` would also mean we can build custom application without having the code associated with git repo, this would have to be further improved.

To simply accommodate the extension use case, I instead suggest to consider a different approach. See https://github.com/Slicer/Slicer/pull/6369

jcfr:
Since this was inadvertently integrated as part of https://github.com/Slicer/Slicer/pull/6371/, I am closing and will revert in https://github.com/Slicer/Slicer/pull/6369

lassoan:
> Since this was inadvertently integrated as part of #6371, I am closing and will revert in #6369

Yes, sorry, I tested that change in my working copy and I did not noticed that it got in with this commit. I'm not sure if https://github.com/Slicer/Slicer/pull/6369 is what we need - I'll add a comment there.

PR #6366, opened 2022-05-12
Fix python lint issues in WebServer module

pieper:
See https://github.com/Slicer/Slicer/runs/6398682294?check_suite_focus=true


pieper:
I think I have all the fixes here (even the ones I disagree with).  @jamesobutler or anyone could have a look and we can merge this whenever.

jcfr:
> or anyone could have a look and we can merge this whenever.

Thanks Steve. I am taking care of this now.

lassoan:
Thank you all for getting this in!

PR #6368, opened 2022-05-12
DOC: Improve Webserver module documentation

jcfr:
* Add "Panels and their use" section and remove "Current Features"
* Highlight security risks using "warning" directive
* Re-organize endpoint sections

### DIFF
--- a/Docs/user_guide/modules/webserver.md
+++ b/Docs/user_guide/modules/webserver.md
@@ -13,35 +13,58 @@ The web server is integrated with the Qt event loop so it can be used together w
 
 This code has been developed over a number of years in a separate repository where there are additional experiments demonstrating other potential uses. The version in Slicer core has been stripped down to address the most common expected use cases. See [https://github.com/pieper/SlicerWeb](https://github.com/pieper/SlicerWeb).
 
-Also note that the code should be considered somewhat experimental and a likely security risk.  Do not expose web server endpoints on the public internet without careful consideration.
+
+:::{warning}
+This module should be considered somewhat experimental and a likely security risk. Do not expose web server endpoints on the public internet without careful consideration.
 
 Because the web server uses standard http, there are many off-the-shelf security options, such as firewalls, ssh-tunneling, and authenticating proxies, that can be used to improve security of installations.
+:::
 
-## Current Features
+## Panels and their use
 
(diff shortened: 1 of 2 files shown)


jcfr:
@lassoan @pieper  How should update the module index to reference the module. 

Few options:
* Existing `Developer Tools` or `Utilities` section
* Introduction of new section called `Server`

jcfr:
image

jcfr:
The current link to https://www.dicomstandard.org/dicomweb is broken. A valid on is https://www.dicomstandard.org/using/dicomweb

In the documentation, the terminology used on this page is `RESTful services` and it lists the different services like `[Query](https://www.dicomstandard.org/using/dicomweb/query-qido-rs/)`, [Retrieve](https://www.dicomstandard.org/using/dicomweb/retrieve-wado-rs-and-wado-uri/), ... I suggest we update the documentation to be more specific and indicated which service are partially available and not available at all.

lassoan (approved):

Docs/user_guide/modules/webserver.md
```diff
@@ -2,46 +2,83 @@
...
-For example, these command can be used to download data and change the screen layout:
+### Remote Control
+
+The web server can also be accessed via other commands such as `curl`. A `dict` can be returned as a json object by setting it in the `__execResult` variable and enabling the `Slicer API exec` feature in the `Advanced` section.
+
+For example, these commands may be used to download the MRHead sample data, change the screen layout and return a dictionary including the ID of the loaded volume:
 
 ```
-curl -X POST localhost:2016/slicer/exec --data "import SampleData; volumeNode = SampleData.SampleDataLogic().downloadMRHead(); slicer.app.layoutManager().setLayout(slicer.vtkMRMLLayoutNode.SlicerLayoutOneUpRedSliceView); __execResult = {'volumeNodeID': volumeNode.GetID()}"
+curl -X POST localhost:2016/slicer/exec --data "\
```
lassoan, line 62: This would not work on Windows. Line continuation character is `^` in .bat files and backtick in PowerShell, and I don't think it is practically feasible in the standard old Windows console.

I would add examples that can be copy-pasted as is, on any OS, even if they are a bit harder to read.

PR #6369, opened 2022-05-12
COMP: Do not require building extension from git repository (alternative approach)

jcfr:
Developers often complained that their extension build failed for some
unknown reason and frequently the issue was that they did not store the
source code in a git repository.

It is good to encourage developers to use best practices, such as version
control, but making the build fail if they build from a non-version-controlled
repository was unnecessarily strong measure.

This commit changes the behavior so that if an extension is built from a
non-git folder then the build succeeds, just a warning is displayed during
project configuration step and `NA` is used as the extension's WC revision.

### DIFF
--- a/CMake/SlicerMacroExtractRepositoryInfo.cmake
+++ b/CMake/SlicerMacroExtractRepositoryInfo.cmake
@@ -39,12 +39,23 @@
 #  <var-prefix>_WC_REVISION_NAME - Name associated with <var-prefix>_WC_REVISION_HASH
 #  <var-prefix>_WC_REVISION_HASH - current revision
 #  <var-prefix>_WC_REVISION - Equal to <var-prefix>_WC_REVISION_HASH
+#  <var-prefix>_WC_COMMIT_COUNT - number of commits in current branch
+#  <var-prefix>_WC_LAST_CHANGED_DATE - date of last commit
 #
-# If source directory is not GIT, the macro will return early
-# displaying an warning message:
+# If the source directory is not a GIT repository, the macro will
+# display a warning message like the following:
 #
 #   -- Skipping repository info extraction: directory [/path/to/src] is not a GIT checkout
 
+# and set the following variables:
+#
+#   <var-prefix>_WC_TYPE: "local"
+#   <var-prefix>_WC_URL: "NA"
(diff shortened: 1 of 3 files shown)


CMake/SlicerMacroExtractRepositoryInfo.cmake
```diff
@@ -67,12 +67,13 @@ macro(SlicerMacroExtractRepositoryInfo)
   set(${wc_info_prefix}_WC_TYPE local)
   set(${wc_info_prefix}_WC_URL "NA")
   set(${wc_info_prefix}_WC_ROOT "NA")
+  set(${wc_info_prefix}_WC_REVISION "NA")
```
lassoan, line 70: Unconditionally setting `..._WC_REVISION` would prevent specifying the revision manually.

CMake/SlicerMacroExtractRepositoryInfo.cmake
```diff
@@ -67,12 +67,13 @@ macro(SlicerMacroExtractRepositoryInfo)
...
   set(${wc_info_prefix}_WC_REVISION_NAME "NA")
   set(${wc_info_prefix}_WC_REVISION_HASH "NA")
 
   if(NOT EXISTS ${MY_SOURCE_DIR}/.git)
 
-    message(AUTHOR_WARNING "Skipping repository info extraction: directory [${MY_SOURCE_DIR}] is not a GIT or SVN checkout")
+    message(AUTHOR_WARNING "Skipping repository info extraction: directory [${MY_SOURCE_DIR}] is not a GIT checkout")
```
lassoan, line 76: This would always make the warning appear. Even if the user wanted to specify the revision manually.

jcfr: Agreed, this an optimization I will update this PR and address that aspect

jcfr:
> specifying the revision manually.

This is now captured in https://github.com/Slicer/Slicer/issues/6377 and will be addressed independently of this particular fix.

I suggest to move with integration of this pull request as it simply addresses the original problem

cc: @lassoan

lassoan (dismissed):
It looks good to me. Maybe add to the macro documentation that values are set to `NA` if repository is local.

PR #6371, opened 2022-05-12
BUG: Fix loading of table without schema

gabrielmuller:
Loading a table with no schema sets the schema filename to the empty string. After saving again, the table can no longer be loaded as the schema filename is invalid.

Steps to reproduce:

1. Create an empty table
2. Using "Save Data", save both the MRML scene and table files
3. Close scene
4. Using "Add Data", open the .mrml file
5. Save the MRML scene again
6. Close scene
7. Open the .mrml file again
This should produce the following error:
```Error: Loading C:/Users/user/path/2022-05-12-Scene.mrml - ERROR: In D:\D\P\S-0\Libs\MRML\Core\vtkMRMLStorableNode.cxx, line 322
vtkMRMLTableNode (0000027FCB9D7540): vtkMRMLStorableNode::UpdateScene failed: Failed to read node Table (vtkMRMLTableNode1) using storage node vtkMRMLTableStorageNode1.```


lassoan:
Very nice catch, thank you for reporting and developing a fix and submitting a pull request!

Your fix works, but since this issue may cause similar problems in other storage nodes, I would fix it one level lower, in vtkMRMLStorageNode. There is nothing else to do on your side, I'm just testing this lower-level fix now and if everything goes well then I'll merge it later today.

PR #6132, opened 2022-01-26
STYLE: Upgrade python syntax to 3.9 and newer 

jamesobutler:
Since Python in Slicer was recently upgraded from Python 3.6 to Python 3.9 (https://github.com/Slicer/Slicer/commit/34e48e8aef5dad19ec8a955d6f48a1940846e3f3), I've passed Slicer python files through a syntax upgrader for some cleanup tasks.

This PR uses [`pyupgrade`](https://github.com/asottile/pyupgrade) to automatically upgrade python syntax to newer versions. The script listed below was run for Python 3.6+ syntax with changes committed, followed by Python 3.7+, then Python 3.8+ and finally Python 3.9+.  There were no changes made when specifying Python 3.7+ or 3.8+ which is why there are no commits for those iterations. Python 3.6+ syntax upgrading was previously completed in https://github.com/Slicer/Slicer/commit/4483cc0e6f288b0816b6329f1829d9ef8c5aa81a. 

## Using pyupgrade

- Install: `PythonSlicer -m pip install pyupgrade`
- Running:
  - On 1 file: `PythonSlicer -m pyupgrade --py36-plus MyPythonFile.py`
  - On multiple files: Here is my pyupgrade-script.py written to automate running pyupgrade across all python files in the Slicer repo. It was run by `PythonSlicer pyupgrade-script.py`.
```python
# pyupgrade-script.py
import os
import subprocess

search_directory = "C:/Users/JamesButler/Documents/GitHub/Slicer"
for root, _, files in os.walk(search_directory):
    for file_item in files:
        file_path = os.path.join(root, file_item)
        if os.path.isfile(file_path) and file_path.endswith(".py"):
            subprocess.call(["PythonSlicer", "-m", "pyupgrade", "--py36-plus", file_path])
```

PR #5625, opened 2021-05-01
STYLE: Upgrade python syntax to 3.6 and newer

jamesobutler:
This PR uses [`pyupgrade`](https://github.com/asottile/pyupgrade) to automatically upgrade python syntax to newer versions. This ultimately removes Python 2 compatibility. Slicer seemingly hasn't been trying to maintain Python 2 compatibility given that many libraries have already dropped Python 2 support since it went EOL on January 1st 2020. Slicer already includes some f-string formatting only supported by Python 3.6 and newer (See [usage](https://github.com/Slicer/Slicer/blob/master/Modules/Scripted/DICOMLib/DICOMUtils.py#L838) in DICOMUtils.py). Slicer's current Stable past couple of stable releases (4.11.20210226 and 4.11.20200930) have both included Python 3 as well.

Around early April 2019 @jcfr began using scripts, such as [`futurize`](https://python-future.org/futurize.html), provided by https://python-future.org/ to turn Python 2 code into valid Python 3 code. This was because Slicer was in a transition period and the Slicer nightly would begin using Python 3 while Slicer stable would still be using Python 2. This work was documented at https://www.slicer.org/wiki/Documentation/Nightly/Developers/Tutorials/MigrationGuide#Python_2_to_Python_3 and at https://discourse.slicer.org/t/updating-slicer-to-work-with-python-3/4662/9.

Notably, most of the changes in this PR are changes due to the following commits:
- `STYLE: Update python scripts to use print function for python 3.x`: https://github.com/Slicer/Slicer/commit/7df896e839dc03e1c2f3b5458590ea16e80fcbba
- `STYLE: Update python classes to follow new-style`: https://github.com/Slicer/Slicer/commit/d366e32491d13461c9c6316b3e4ed80fb4279d3f

## Using pyupgrade

- Install: `PythonSlicer -m pip install pyupgrade`
- Running:
  - On 1 file: `PythonSlicer -m pyupgrade --py36-plus MyPythonFile.py`
  - On multiple files: Here is my pyupgrade-script.py written to automate running pyupgrade across all python files in the Slicer repo. It was run by `PythonSlicer pyupgrade-script.py`.
```python
# pyupgrade-script.py
import os
import subprocess

search_directory = "C:/Users/JamesButler/Documents/GitHub/Slicer"
for root, _, files in os.walk(search_directory):
    for file_item in files:
        file_path = os.path.join(root, file_item)
        if os.path.isfile(file_path) and file_path.endswith(".py"):
[...]

PR #6375, opened 2022-05-13
STYLE: Update python scripts with whitespace fixes

jamesobutler:
This is part of #6271 and the commit here is in the style of https://github.com/Slicer/Slicer/commit/a4adeff2e8f6f93a7ea4d465226cc3c4de3f200a. These changes were almost all completed through local automation. This commit once merged should be added to https://github.com/Slicer/Slicer/blob/master/.git-blame-ignore-revs.

This fixes all remaining pycodestyle E2 whitespace error codes.

Bulk of updates performed using autopep8 CLI
with the following command:

  autopep8 --in-place --select E201,E202,E203,E211,E221,E222,E225,E226,E227,E228,E231,E241,E251,E261,E262,E265,E266,E271 $(git ls-files '*.py')

See https://github.com/hhatto/autopep8#readme

References:

* Whitespace after '('
  See https://www.flake8rules.com/rules/E201.html

* Whitespace before ')'
  See https://www.flake8rules.com/rules/E202.html

* Whitespace before ':'
  See https://www.flake8rules.com/rules/E203.html

* Whitespace before '('
  See https://www.flake8rules.com/rules/E211.html

* Multiple spaces before operator
  See https://www.flake8rules.com/rules/E221.html

* Multiple spaces after operator
  See https://www.flake8rules.com/rules/E222.html

* Missing whitespace around operator
  See https://www.flake8rules.com/rules/E225.html

* Missing whitespace around arithmetic operator
  See https://www.flake8rules.com/rules/E226.html

* Missing whitespace around bitwise or shift operator
  See https://www.flake8rules.com/rules/E227.html

* Missing whitespace around modulo operator
  See https://www.flake8rules.com/rules/E228.html

* Missing whitespace after ',', ';', or ':'
  See https://www.flake8rules.com/rules/E231.html

* Multiple spaces after ','
  See https://www.flake8rules.com/rules/E241.html

* Unexpected spaces around keyword / parameter equals
  See https://www.flake8rules.com/rules/E251.html

* At least two spaces before inline comment
  See https://www.flake8rules.com/rules/E261.html

* Inline comment should start with '# '
  See https://www.flake8rules.com/rules/E262.html

* Block comment should start with '# '
  See https://www.flake8rules.com/rules/E265.html

* Too many leading '#' for block comment
  See https://www.flake8rules.com/rules/E266.html

* Multiple spaces after keyword
  See https://www.flake8rules.com/rules/E271.html

PR #6376, opened 2022-05-14
Upgrade python syntax to 3.9 and newer

jamesobutler:
<s>This continually enforces changes made in https://github.com/Slicer/Slicer/commit/4483cc0e6f288b0816b6329f1829d9ef8c5aa81a and https://github.com/Slicer/Slicer/commit/1a85ebabd543907072c72c34247aae849489d33b.

Changes were applied automatically by running:
pre_commit run --all-files</s>

---------

**EDIT**

Similarly to 1a85eba (introduced through PR #6132) and 4483cc0 (introduced through PR #5625), this PR uses [`pyupgrade`](https://github.com/asottile/pyupgrade) to also automatically upgrade python syntax.

## Using pyupgrade

- Install: `PythonSlicer -m pip install pyupgrade`
- Running:
  - On 1 file: `PythonSlicer -m pyupgrade --py39-plus MyPythonFile.py`
  - On multiple files, there are few options:
    - on unix system:
      ```
      PythonSlicer -m pyupgrade --py39-plus $(find . -type f -name "*.py")
      ```
    - on any system, use @jamesobutler `pyupgrade-script.py` written to automate running pyupgrade across all python files in the Slicer repo. It was run by `PythonSlicer pyupgrade-script.py`.
      ```python
      # pyupgrade-script.py
      import os
      import subprocess

      search_directory = "C:/Users/JamesButler/Documents/GitHub/Slicer"
      for root, _, files in os.walk(search_directory):
          for file_item in files:
              file_path = os.path.join(root, file_item)
              if os.path.isfile(file_path) and file_path.endswith(".py"):
                  subprocess.call(["PythonSlicer", "-m", "pyupgrade", "--py39-plus", file_path])
      ```

### DIFF
--- a/Base/Python/slicer/util.py
+++ b/Base/Python/slicer/util.py
@@ -2716,7 +2716,7 @@ def _messageDisplay(logLevel, text, testingReturnValue, mainWindowNeeded=False, 
     if not windowTitle:
         windowTitle = slicer.app.applicationName + " " + logLevelString
     if slicer.app.testingEnabled():
-        logging.info("Testing mode is enabled: Returning %s and skipping message box [%s]." % (testingReturnValue, windowTitle))
+        logging.info(f"Testing mode is enabled: Returning {testingReturnValue} and skipping message box [{windowTitle}].")
         return testingReturnValue
     if mainWindowNeeded and mainWindow() is None:
         return
@@ -2739,7 +2739,7 @@ def messageBox(text, parent=None, **kwargs):
     import logging, qt, slicer
     if slicer.app.testingEnabled():
         testingReturnValue = qt.QMessageBox.Ok
-        logging.info("Testing mode is enabled: Returning %s (qt.QMessageBox.Ok) and displaying an auto-closing message box [%s]." % (testingReturnValue, text))
+        logging.info(f"Testing mode is enabled: Returning {testingReturnValue} (qt.QMessageBox.Ok) and displaying an auto-closing message box [{text}].")
         slicer.util.delayDisplay(text, autoCloseMsec=3000, parent=parent, **kwargs)
         return testingReturnValue
 
--- a/Applications/SlicerApp/Testing/Python/CLISerializationTest.py
+++ b/Applications/SlicerApp/Testing/Python/CLISerializationTest.py
@@ -70,9 +70,9 @@ if __name__ == '__main__':
     temp_dir = os.path.expanduser(getattr(args, "/path/to/temp_dir"))
 
     # Create input/output
-    serializeSeedsOutFile = '%s/%s.acsv' % (temp_dir, 'SeedsSerialized')
-    deserializeSeedsOutFile = '%s/%s.acsv' % (temp_dir, 'SeedsDeSerialized')
-    json_file = '%s/%s.json' % (temp_dir, 'ExecutionModelTourSerialized')
+    serializeSeedsOutFile = f'{temp_dir}/SeedsSerialized.acsv'
+    deserializeSeedsOutFile = f'{temp_dir}/SeedsDeSerialized.acsv'
+    json_file = f'{temp_dir}/ExecutionModelTourSerialized.json'
 
     # Copy .mrml file to prevent modification of source tree
     mrml_source_path = os.path.join(data_dir, 'ExecutionModelTourTest.mrml')
--- a/Base/Python/slicer/ScriptedLoadableModule.py
+++ b/Base/Python/slicer/ScriptedLoadableModule.py
@@ -299,7 +299,7 @@ class ScriptedLoadableModuleTest(unittest.TestCase):
     """
 
     def __init__(self, *args, **kwargs):
-        super(ScriptedLoadableModuleTest, self).__init__(*args, **kwargs)
+        super().__init__(*args, **kwargs)
 
         # See https://github.com/Slicer/Slicer/pull/6243#issuecomment-1061800718 for more information.
         # Do not pass *args, **kwargs since there is no base class after `unittest.TestCase`. This is only relevant to
--- a/Modules/Scripted/DICOMPlugins/DICOMEnhancedUSVolumePlugin.py
+++ b/Modules/Scripted/DICOMPlugins/DICOMEnhancedUSVolumePlugin.py
@@ -1,4 +1,3 @@
-
 import vtk
 
 import slicer
--- a/Modules/Scripted/DataProbe/DataProbe.py
+++ b/Modules/Scripted/DataProbe/DataProbe.py
@@ -169,7 +169,7 @@ class DataProbeInfoWidget:
             value = self.calculateTensorScalars(tensor, operation=operation)
             if value is not None:
                 valueString = ("%f" % value).rstrip('0').rstrip('.')
-                return "%s %s" % (scalarVolumeDisplayNode.GetScalarInvariantAsString(), valueString)
+                return f"{scalarVolumeDisplayNode.GetScalarInvariantAsString()} {valueString}"
             else:
                 return scalarVolumeDisplayNode.GetScalarInvariantAsString()
 
--- a/Modules/Scripted/ImportItkSnapLabel/ImportItkSnapLabel.py
+++ b/Modules/Scripted/ImportItkSnapLabel/ImportItkSnapLabel.py
@@ -26,7 +26,7 @@ class ImportItkSnapLabel(ScriptedLoadableModule):
 # (identified by its special name <moduleName>FileReader)
 #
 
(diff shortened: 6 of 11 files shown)


jamesobutler:
It will be easier to resolve merge conflicts in this PR if #6375 is merged first.

jamesobutler:
I've rebased this PR to fix merge conflicts from the recent whitespace updates.

jamesobutler:
I'm following up on some of my old PRs. @jcfr Do you think we can go ahead and proceed with this one? Let me know if `ENH` is an appropriate commit prefix for some of this CI/utility related work - re `ENH: Update GitHub Actions lint workflow to include pyupgrade`

jamesobutler:
Following up on this old PR here. 

Any objections to proceeding with this one @jcfr, @lassoan, @pieper ?

Base/Python/slicer/util.py
```diff
@@ -2707,7 +2707,7 @@ def _messageDisplay(logLevel, text, testingReturnValue, mainWindowNeeded=False,
     if not windowTitle:
         windowTitle = slicer.app.applicationName + " " + logLevelString
     if slicer.app.testingEnabled():
-        logging.info("Testing mode is enabled: Returning %s and skipping message box [%s]." % (testingReturnValue, windowTitle))
+        logging.info(f"Testing mode is enabled: Returning {testingReturnValue} and skipping message box [{windowTitle}].")
```
lassoan, line 2710: I would prefer not to introduce usage of f-strings in logging functions, because using f-strings results in performing expensive string manipulations even if the message does not end up being printed due to the current log level.

Has pyupgrade recommended this change?

Base/Python/slicer/util.py
```diff
@@ -2730,7 +2730,7 @@ def messageBox(text, parent=None, **kwargs):
     import logging, qt, slicer
     if slicer.app.testingEnabled():
         testingReturnValue = qt.QMessageBox.Ok
-        logging.info("Testing mode is enabled: Returning %s (qt.QMessageBox.Ok) and displaying an auto-closing message box [%s]." % (testingReturnValue, text))
+        logging.info(f"Testing mode is enabled: Returning {testingReturnValue} (qt.QMessageBox.Ok) and displaying an auto-closing message box [{text}].")
```
lassoan, line 2733: Same as above, I would not use f-strings in logging.

lassoan (commented):
Thank you, looks all good, just added two small comments.

Base/Python/slicer/util.py
jamesobutler: Yes `pyupgrade` changes it to .format() and then re-running changes it again to an f-string. The maintainer points to https://github.com/asottile/pyupgrade/issues/503#issuecomment-888267163 in that f-string usage here is not a performance drop.

lassoan: Thanks for the link @jamesobutler, it led to some very interesting discussions.  There seems to be a very strong disagreement between various Python developers on this topic and probably the correct answer depends on the use case. The insignificant performance drop was a measurement error, in the same issue further along the discussion, the actual performance drop was shown to be about 20%. So, this is a real problem. There are additional potential issues with exceptions during string conversion.

However, in our specific these cases here, there is no performance or exception-safety problems, so it should be fine to use f-strings.

My only concern is that using f-strings in logging would make this the default choice everywhere, even where performance matters. But maybe keeping the classic formatting in our log messages is not a very effective way of educating/reminding developers anyway, and if the cost is not to use pyupgrade at all or spend a lot of time with fighting with various lint tools, then it may just not worth it.

So, I'm OK with keeping the suggested changes.

jamesobutler:
I've rebased to fix the merge conflict.

jamesobutler:
I've rebased to update to the latest version of `pyupgrade` and re-run to be in compliance.

Base/Python/slicer/util.py
jcfr: Looking at https://pylint.pycqa.org/en/latest/user_guide/messages/warning/logging-fstring-interpolation.html, the following is indicated:

> Another reasonable option is to use f-string. If you want to do that, you need to enable logging-format-interpolation and disable logging-fstring-interpolation.

Given the fact there is no significant performance impact [^1], enforcing f-string everywhere is reasonable.

That said, I think we should explicitly configure pylint with:
* `logging-fstring-interpolation` -> `disabled`
* `logging-format-interpolation` -> `enabled`

[^1]: https://github.com/Slicer/Slicer/pull/6376/files#r984071600

jamesobutler:
@jcfr I've included W1203 in the ignore of .flake8 configuration which I believe utilizes the pylint codes.
https://pylint.pycqa.org/en/latest/user_guide/messages/warning/logging-fstring-interpolation.html

jcfr:
After looking at this in more details, I realized that `pyupgrade` does not make use of `pylint`, so updating the setting has no effect. I will then remove the last commit.

* [ ] Remove commit `ENH: Ignore pylint W1203 logging-fstring-interpolation `

Strictly speaking, running pyupgrade in the GitHub workflow does not really make sense because (1) we do not yet push back the changes to the open pull-request (this could be done enabling  and (2) rewriting details are not reported back in the output.

Using pre-commit to run against all files may indeed be convenient and may avoid the following:

```
PythonSlicer -m pyupgrade --py39-plus $(find . -type f -name "*.py")
```

and only require the following

```
PythonSlicer -m pip install pre-commit
```

Source: https://discourse.slicer.org/t/locally-execute-pre-commit-linting-locally/22837


This means that the update of the workflow without enabling service like `pre-commit.ci` [^1] or using action like `git-auto-commit-action`[^2] does not really fulfill:

> continually enforces changes made in 4483cc0 and 1a85eba.

[^1]: https://pre-commit.ci/
[^2]: https://github.com/stefanzweifel/git-auto-commit-action

For these reason, I will:
* [ ] update the pull-request to only integrate the changes without updating the workflow.
* [ ] create an issue for adding auto-update of pull-request

jcfr:
@jamesobutler Comparing with the version I obtained with `pyupgrade` with the one you originally introduced in https://github.com/Slicer/Slicer/tree/0baa88d681a9a4a0bdc8a42a1b6bd97cf1f9265e, I noticed that the following different. Did you explicitly update `CLISerializationTest.py` ?

image

jamesobutler:
There was indeed some re-runs of pyupgrade as when you run it once and then immediately run it a second time, it will change more things the second time around. Yes, I might’ve additionally changed some things as well but I don’t remember.

Adding pyupgrade to the pre-commit run enforces the updated syntax for new PRs, but I wasn’t planning for it to be auto fixed by a bot. It would be up to the developer to see the reported issues just like if they had incorrect tab counts reported by pylint which they would manually fix and push.

> Strictly speaking, running pyupgrade in the GitHub workflow does not really make sense because (1) we do not yet push back the changes to the open pull-request (this could be done enabling and (2) rewriting details are not reported back in the output.

Re (1) I suggest not to auto-fix PRs with pre-commit.ci to push changes back as mentioned in
https://github.com/Slicer/Slicer/pull/6573#issuecomment-1274839133 because when working in a multi commit PR the changes would be in a single commit and not squashed down cleanly into the commit that introduced the compliance issue. Auto-fix can be helpful but will reduce the quality of the commit history with fix-me-up type commits.

Re (2) rewriting details are indeed reported back in the GitHub action workflow results (or in pre-commit.ci results) and also locally as you’ve shown in your screenshot above. This provides developers clear instructions about what something should be changed or that they can run pyupgrade locally to apply the changes and commit and squash into their appropriate commit.

jcfr:
Thanks for the detailed answer, this is very helpful :pray: 

I will integrate the change and publish pull-requests respectively adding (1) execution of `pyupgrade` (2) `autoupdate` [^1] through the GitHub workflow.

> not planning for it to be auto fixed by a bot [...] it would be up to the developer to see the reported issues

:+1: 

> rewriting details [...] reported back in the GitHub action workflow results

The screenshot I provided was obtained after running the tool locally, I will investigate further to see how rewritten changed are displayed within the GitHub workflow as I couldn't find an example (yet)


[^1]: https://github.com/Slicer/Slicer/pull/6573#issuecomment-1274839133

jamesobutler:
Here' a screenshot from a private repo with an example of pyupgrade indicating changes:
image

jcfr:
Thanks for sharing the example, I will follow up with a PR including your original commit  updating the `pre-commit` configuration. :rocket:

jcfr:
> I will follow up with a PR including your original commit updating the `pre-commit` configuration.

See https://github.com/Slicer/Slicer/pull/6605

PR #6378, opened 2022-05-14
COMP: Update build-system to support forcing extension revision

jcfr:


CMake/SlicerExtensionCPack.cmake
```diff
@@ -59,6 +59,14 @@ endforeach()
 include(SlicerMacroExtractRepositoryInfo)
 SlicerMacroExtractRepositoryInfo(VAR_PREFIX ${EXTENSION_NAME})
 
+#-----------------------------------------------------------------------------
+# Set extension revision
+#-----------------------------------------------------------------------------
+if(NOT "${${EXTENSION_NAME}_FORCED_REVISION}" STREQUAL "")
```
lassoan, line 65: The variable should be `"${${EXTENSION_NAME}_FORCED_WC_REVISION}"` (`WC_` was missing).

lassoan, line 65: Originally the CMake configuration output only contained `Skipping repository info extraction: directory` message twice, which was not very helpful. After adding `${MY_VAR_PREFIX}` to the `AUTHOR_WARNING` message I got these two more informative messages:
- `Skipping Sandbox repository info extraction: directory`
- `Skipping EXTENSION repository info extraction: directory`

I think we should add `${MY_VAR_PREFIX}` to the warning message.

CMake/SlicerExtensionCPack.cmake
```diff
@@ -59,6 +59,14 @@ endforeach()
...
+  set(${EXTENSION_NAME}_WC_REVISION "${${EXTENSION_NAME}_FORCED_REVISION}")
```
lassoan, line 66:
```suggestion
  set(${EXTENSION_NAME}_WC_REVISION "${${EXTENSION_NAME}_FORCED_WC_REVISION}")
```

CMake/SlicerExtensionCPack.cmake
lassoan, line 65: Should we add this `...FORCED...` variable to the CACHE? If we configure the build with `-DSandbox_FORCED_WC_REVISION=300` but then we configure again without this option then is the forced revision going to be preserved?

lassoan (commented):
Thank you. I've tested this a bit and found a few little issues. Now that we are working on this I think it would be nice to address them.

jcfr:
Thanks for testing and for providing detailed suggestion.

I also checked that rerunning cmake in a already configured build tree kept the forced revision.

jcfr:
cc: @jamesobutler @lassoan

jamesobutler (approved):

Issue #1032, opened 2020-03-12
GetReferencedSubScene: Error message when file is in scene


slicerbot:
_This issue was created automatically from an original [Mantis Issue](https://mantisarchive.slicer.org/view.php?id=1032). Further discussion may take place here._

PR #6380, opened 2022-05-16
BUG: Make built-in procedural color nodes (such as PET-DICOM) read only

MichaelColonel:
Possible fix of the first bug in #6111.

CommonTK PR [#1032](https://github.com/commontk/CTK/pull/1032) must be applied as well.

Drawback: Colour transfer function thumbnail icon and widget representation (ColorNode) isn't updated on the fly. Only if you hover a mouse arrow above it.

Read only for build-in CTF nodes (not editable)
image

Node is editable after being copied.
image


lassoan (commented):
Thank you, it looks good. I'll just touch up the commit messages and test it and if all works well then I'll merge it.

PR #6385, opened 2022-05-18
BUG: Improve slicer.util.messageBox() to avoid blocking in testing mode

jcfr:
This commit is a follow up of c47d816c7 (ENH: Skip message boxes when testing)
and 70bb3b2ce (ENH: Refactor message box utility functions)

### DIFF
--- a/Base/Python/slicer/util.py
+++ b/Base/Python/slicer/util.py
@@ -2480,7 +2480,7 @@ def tempDirectory(key='__SlicerTemp__',tempDir=None,includeDateTime=True):
   return dirPath
 
 
-def delayDisplay(message, autoCloseMsec=1000, parent=None):
+def delayDisplay(message, autoCloseMsec=1000, parent=None, **kwargs):
   """Display an information message in a popup window for a short time.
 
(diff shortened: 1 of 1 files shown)


jcfr:
@lassoan  This will avoid calls like the one done in `CardiacDeviceSimulator` [^1] from blocking and wait for user input.

[^1] https://github.com/SlicerHeart/SlicerHeart/blob/e2237ca4d11ea4bd4758108fadf7e5a0de8022bf/CardiacDeviceSimulator/CardiacDeviceSimulator.py#L75-L77

lassoan (commented):
This is a good idea. I would use `delayDisplay` instead with a few-second delay to preserve the overall behavior of the script when a developer is watching its execution.

jcfr:
@lassoan This should address your comment. Further refactoring could be done, for example `autoCloseMsec` could be added to `messageBox` but I would prefer to tackle this at some other time.

lassoan (approved):
Thank you, it looks good. Maybe use autoCloseMsec=3000 to give some more time to see the message.

PR #6386, opened 2022-05-18
BUG: Ensure vtkMRMLCameraNode layout name can be updated

jcfr:
This commit is a follow up of e78f557cd (ENH: Simplify view widgets
initialization from view nodes) that originally introduced the
`vtkMRMLCameraNode::GetLayoutName/SetLayoutName`.

It makes sure that the layout name associated with a given camera node
can be set to a different value after it had already been set once.


lassoan (approved):
Thank you, this looks good.

PR #6391, opened 2022-05-23
BUG: Escape HTML chars in CLI output

gabrielmuller:
Characters such as `<` in CLI output are interpreted as HTML. This causes some lines to become truncated, or displayed in wrong color (such as red instead of white for standard output).


lassoan (approved):
It would be nice to allow CLI modules to provide rich HTML-formatted output, but the current implementation does not seem to be prepared for that (indicated by placing the module output inside a `<pre>` block). Therefore the proposed change makes sense: it makes plain-text output required and it also ensures that plain-text output is rendered correctly.

Adding HTML output support could be a feature implemented later. It could work for example by requiring the module to start its output with `<html>`. That would switch from verbatim display of plain-text (current display mode) to display the text as html (new display mode).

gabrielmuller:
@pieper Well noted. Tags such as `<filter-progress>` are parsed before the HTML escape, so they should still work.

PR #6394, opened 2022-05-26
BUG: Fix subject hierarchy table showing nodes removed during batch processing

jamesobutler:
Originally reported at https://discourse.slicer.org/t/optimizing-performance-of-events-when-clearing-nodes/23588/9.

Code to replicate original issue
```python
slicer.util.selectModule("Data")
first = slicer.mrmlScene.AddNewNodeByClass("vtkMRMLMarkupsFiducialNode", "First")
second = slicer.mrmlScene.AddNewNodeByClass("vtkMRMLMarkupsFiducialNode", "Second")
third = slicer.mrmlScene.AddNewNodeByClass("vtkMRMLMarkupsFiducialNode", "Third")
slicer.mrmlScene.StartState(slicer.mrmlScene.BatchProcessState)

for node in [third, second]:
  slicer.mrmlScene.RemoveNode(node)

slicer.mrmlScene.EndState(slicer.mrmlScene.BatchProcessState)
fourth = slicer.mrmlScene.AddNewNodeByClass("vtkMRMLMarkupsFiducialNode", "Fourth")

```
After running the above code:

| Current | This PR |
|----------|---------|
|image|image|

I have removed `updateFromSubjectHierarchy` because of it's single usage in Slicer core being removed. This could be reverted if it is desired even though it is code not being used by core. `onMRMLSceneEndBatchProcess` was switched to using the full `rebuildFromSubjectHierarchy` as when removing nodes during batch processing it was not removing them from the widget when batch processing state was ended.


jamesobutler:
> Out of curiosity: have you tried improving the update function to remove these items?

@cpinter I have attempted this, but the subject hierarchy code has been pretty confusing for me to understand between all the item terminology. Do you have any thoughts how to access these items where are no longer in the subject hierarchy node, but are persisting in the model of the various table views observing the subject hierarchy?

For example the following where I remove all items in the subject hierarchy after exiting batch process state, still results in items in the table view.
```python
slicer.util.selectModule("Data")
first = slicer.mrmlScene.AddNewNodeByClass("vtkMRMLMarkupsFiducialNode", "First")
second = slicer.mrmlScene.AddNewNodeByClass("vtkMRMLMarkupsFiducialNode", "Second")
third = slicer.mrmlScene.AddNewNodeByClass("vtkMRMLMarkupsFiducialNode", "Third")
slicer.mrmlScene.StartState(slicer.mrmlScene.BatchProcessState)

for node in [third, second]:
  slicer.mrmlScene.RemoveNode(node)

slicer.mrmlScene.EndState(slicer.mrmlScene.BatchProcessState)
fourth = slicer.mrmlScene.AddNewNodeByClass("vtkMRMLMarkupsFiducialNode", "Fourth")
shNode = slicer.mrmlScene.GetSubjectHierarchyNode()
shNode.RemoveAllItems()
```
image

jamesobutler:
@cpinter I could keep `updateFromSubjectHierarchy`, but remove the early return in the BatchProcessState when items are being removed from the subject hierarchy. Would mean processing the removal and reparenting as it goes instead of doing it once out of the BatchProcessState.

https://github.com/Slicer/Slicer/blob/d24cd5be05e36ac5bf67ad5781abdac6ea5904e6/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyModel.cxx#L1521-L1527
https://github.com/Slicer/Slicer/blob/d24cd5be05e36ac5bf67ad5781abdac6ea5904e6/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyModel.cxx#L1553-L1560

Would produce the expected result as well.
image

cpinter:
> pretty confusing for me to understand between all the item terminology

Are you familiar with Qt's model/view architecture? If not, then either start from there (and then find the Qt model items corresponding to non-existent SH items and remove them from the Qt model), or consider this issue done.

> Would mean processing the removal and reparenting as it goes instead of doing it once out of the BatchProcessState.

Yes, but the whole point of batch processing is to postpone these updates until the end so that the processing itself is fast when you know that many objects are to be processed.

Btw I don't see images uploaded in GitHub comments for around two days now. Hopefully it will be fixed, otherwise it will be hard to follow discussions without seeing the illustrations...

image

jamesobutler:
I'm not sure what is going on your end with the pictures. The pictures are rendering successfully for me on GitHub in my browser and on my mobile device.

Currently I have limited time with this improvement so if removing of `updateFromSubjectHierarchy` and using `rebuildFromSubjectHierarchy` is not a valid improvement then we can close this PR and let this problem of batch processing to persist until it can be fixed in another way.

cpinter:
Sorry this has fallen through the cracks. I'm OK with integrating this. My only problem is that I'm not completely sure what other consequences it will have that we rebuild the hierarchy from scratch - haven't had the time to do thorough testing. I guess we can integrate this and see if it causes problems. If it does it will occur quickly because we all use SH to overview and handle the loaded data.

cpinter:
@jamesobutler @lassoan If you don't see any issues please go ahead with integration whenever appropriate.

cpinter (approved):

PR #6395, opened 2022-05-26
ENH: Optimize processing Segment Editor updates during Batch Processing

jamesobutler:
Originally discussed in https://discourse.slicer.org/t/optimizing-performance-of-events-when-clearing-nodes/23588/5. A `masterVolumeNodeChanged` call in SegmentEditor was 25% of the time when clearing a set of volume nodes from the scene in some custom code that I have.

For example when removing volume nodes, this can trigger a change of the current Master Volume node specification in the Segment Editor module if the current volume node is removed from the scene. When this happens over and over during batch processing there was frequent updating of the master volume node for SegmentEditor logic which can be an expensive operation, especially the masterVolumeNodeChanged call for SegmentEditorThresholdEffect.

### DIFF
--- a/Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.cxx
+++ b/Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.cxx
@@ -1178,7 +1178,7 @@ void qMRMLSegmentEditorWidget::updateEffectList()
 void qMRMLSegmentEditorWidget::updateWidgetFromMRML()
 {
   Q_D(qMRMLSegmentEditorWidget);
-  if (!this->mrmlScene() || this->mrmlScene()->IsClosing())
+  if (!this->mrmlScene() || this->mrmlScene()->IsClosing() || this->mrmlScene()->IsBatchProcessing())
     {
     return;
     }
@@ -1777,12 +1777,35 @@ void qMRMLSegmentEditorWidget::setMRMLScene(vtkMRMLScene* newScene)
 
   // observe close event so can re-add a parameters node if necessary
   this->qvtkConnect(this->mrmlScene(), vtkMRMLScene::EndCloseEvent, this, SLOT(onMRMLSceneEndCloseEvent()));
+  this->qvtkConnect(this->mrmlScene(), vtkMRMLScene::EndBatchProcessEvent, this, SLOT(onMRMLSceneEndBatchProcessEvent()));
 }
 
 //-----------------------------------------------------------------------------
 void qMRMLSegmentEditorWidget::onMRMLSceneEndCloseEvent()
 {
   this->initializeParameterSetNode();
+  this->updateWidgetFromMRML();
+}
+
+//-----------------------------------------------------------------------------
+void qMRMLSegmentEditorWidget::onMRMLSceneEndBatchProcessEvent()
+{
+  Q_D(qMRMLSegmentEditorWidget);
+  if (!this->mrmlScene())
+    {
+    return;
+    }
+  if (d->ParameterSetNode->GetSegmentationNode() != d->SegmentationNodeComboBox->currentNode())
+    {
+    this->setSegmentationNode(d->SegmentationNodeComboBox->currentNode());
+    }
+
+  if (d->ParameterSetNode->GetSourceVolumeNode() != d->SourceVolumeNodeComboBox->currentNode())
+    {
+    this->setSourceVolumeNode(d->SourceVolumeNodeComboBox->currentNode());
+    }
+
+  // force update (clear GUI if no node is selected anymore)
   this->updateWidgetFromMRML();
 }
 
@@ -2003,6 +2026,10 @@ QString qMRMLSegmentEditorWidget::sourceVolumeNodeID()const
 void qMRMLSegmentEditorWidget::onSegmentationNodeChanged(vtkMRMLNode* node)
 {
   Q_D(qMRMLSegmentEditorWidget);
+  if (!this->mrmlScene() || this->mrmlScene()->IsBatchProcessing())
+    {
+    return;
+    }
   this->setSegmentationNode(node);
 }
 
@@ -2066,6 +2093,10 @@ void qMRMLSegmentEditorWidget::setCurrentSegmentID(const QString segmentID)
 void qMRMLSegmentEditorWidget::onSourceVolumeNodeChanged(vtkMRMLNode* node)
 {
   Q_D(qMRMLSegmentEditorWidget);
+  if (!this->mrmlScene() || this->mrmlScene()->IsBatchProcessing())
+    {
+    return;
+    }
   this->setSourceVolumeNode(node);
 }
 
--- a/Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.h
+++ b/Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.h
@@ -494,6 +494,9 @@ protected slots:
   /// Clean up when scene is closed
   void onMRMLSceneEndCloseEvent();
 
+  /// Updates needed after batch processing is done
+  void onMRMLSceneEndBatchProcessEvent();
+
   /// Set default parameters in parameter set node (after setting or closing scene)
   void initializeParameterSetNode();
 


Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.cxx
```diff
@@ -2043,6 +2060,10 @@ void qMRMLSegmentEditorWidget::setCurrentSegmentID(const QString segmentID)
 void qMRMLSegmentEditorWidget::onMasterVolumeNodeChanged(vtkMRMLNode* node)
 {
   Q_D(qMRMLSegmentEditorWidget);
+  if (!this->mrmlScene() || this->mrmlScene()->IsBatchProcessing())
```
lassoan, line 2063: The master volume node selector has a "None" option, so if a node is deleted then that should not lead to any performance issue (the combobox will simply switch to the None option). Are you sure you need to ignore the event?

Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.cxx
```diff
@@ -1763,6 +1764,18 @@ void qMRMLSegmentEditorWidget::onMRMLSceneEndCloseEvent()
...
 
+//-----------------------------------------------------------------------------
+void qMRMLSegmentEditorWidget::onMRMLSceneEndBatchProcessEvent()
+{
+  Q_D(qMRMLSegmentEditorWidget);
+  if (!this->mrmlScene())
+    {
+    return;
+    }
+  // force update (clear GUI if no node is selected anymore)
```
lassoan, line 1775: End result of deleting the current segmentation node must be the same regardless of batch processing was activated or not.

Since the current behavior is to choose some other segmentation node when the current segmentation node gets deleted, this same behavior should be preserved here. You may need to set the current segmentation node to some other existing node (the same way how it is set when batch processing is not activated, just skipping all the intermediate node selection changes).

lassoan (commented):
Thanks a lot for working on this. This will be a nice change that could indeed significantly improve the performance if many segmentation nodes are deleted. I've added a few comments.

Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.cxx
jamesobutler: I’m pretty sure it is just like the Slice views volume node combo box in that if the current selection is removed, it selects the last available index. It won’t select the “None” index. So when deleting the current selection of the master volume node combo box over and over it will keep triggering the event.

jamesobutler: @lassoan I have confirmed that deleting the current selection of the master volume node will choose the last available index. This selects another volume node as the master volume node when there are multiple volumes loaded in the scene. It only selects the "Select master volume to enable editing" option if there are no more volume nodes to choose from.

https://user-images.githubusercontent.com/15837524/170770942-e8dbf9eb-462e-4cb9-a079-2b9416a6cc99.mp4

Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.cxx
jamesobutler: @lassoan I'm not sure I understand. As this PR currently does, it maintains the behavior of when the current segmentation node is removed, it selects the last available index of the combobox which is some other other segmentation node if there was more than 1 originally in the scene.

lassoan: > it selects the last available index of the combobox which is some other other segmentation node if there was more than 1 originally in the scene.

Perfect, I just wanted to confirm that this is the behavior.

Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.cxx
lassoan: I've just tested this and I've found that the result of master volume deletion is different in batch-mode and non-batch-mode.

How to reproduce:

- Load [this scene](https://1drv.ms/u/s!Arm_AFxB9yqHyJd5y3S3XT5BShwTvw?e=lZRaZk) (contains 3 segmentation nodes, each with a different master volume)
- Go to Segment editor module
- Run this script:

```
scene = slicer.mrmlScene
scene.StartState(scene.BatchProcessState)
slicer.mrmlScene.RemoveNode(getNode('MRBrainTumor1'))
scene.EndState(scene.BatchProcessState)
```

Result: master volume is now `Select master volume to enable editing`

- Restart Slicer, load the same scene
- Run this script:
- ```
scene = slicer.mrmlScene
slicer.mrmlScene.RemoveNode(getNode('MRBrainTumor1'))
```

Result: master volume is now `CTChest`

The problem may be that the `setMasterVolumeNode` has been completely missed. I don't think we can skip it, we can just delay it until the batch processing is completed.

It probably happens because even though we return early from `onMasterVolumeNodeChanged`, there is still a call to `updateWidgetFromMRML` (due to node references are being changed, which invokes a parameter node modified event).

Both selecting another random volume and setting master volume to nullptr makes sense (I think I prefer the latter), but what is not OK to have different behavior in batch and non-batch mode.

File: Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.cxx (hunk 2 above)
Comment (lassoan) [reply Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.cxx:1775]: I've checked the behavior on my computer and I've found that it was inconsistent in batch and non-batch modes, similarly to the volume selection.

How to reproduce:

- Load [this scene](https://1drv.ms/u/s!Arm_AFxB9yqHyJd5y3S3XT5BShwTvw?e=lZRaZk) (contains 3 segmentation nodes, each with a different master volume)
- Go to Segment editor module
- Run this script:

```
scene = slicer.mrmlScene
scene.StartState(scene.BatchProcessState)
slicer.mrmlScene.RemoveNode(getNode('Segmentation'))
scene.EndState(scene.BatchProcessState)
```

Result: segmentation node is now `Select a segmentation`

- Restart Slicer, load the same scene
- Go to Segment editor module
- Run this script:

```
scene = slicer.mrmlScene
slicer.mrmlScene.RemoveNode(getNode('Segmentation'))
```

Result: selected segmentation node is now `Segmentation_1`

The problem may be similar here as above, that `setSegmentationNode` has been completely missed. I don't think we can skip it, we can just delay it until the batch processing is completed.

There is an early return from `onSegmentationNodeChanged`, but there are still updates due to `updateWidgetFromMRML` called due to node reference change in segment editor parameter node.

We could change the current behavior to always unselect the segmentation (in both batch and non-batch mode), but the fix should not introduce any difference between batch and non-batch mode.

Comment (jamesobutler) [reply Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.cxx:1775]: ah thanks for the check. I will make sure the segmentation node and master volume node are appropriately set at the end of the process. I had been meaning to confirm that after my assumption that `updateWidgetFromMRML` would've completed those actions.

**Comment by jamesobutler** [comment]
@lassoan I've pushed an update to make sure the regular processing mode and the batch process update both end up in the same final result. They now match the regular processing mode in picking the node at the same index as the previously removed node.

File: Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.cxx [hunk 3]
```diff
@@ -1980,6 +1993,10 @@ QString qMRMLSegmentEditorWidget::masterVolumeNodeID()const
 void qMRMLSegmentEditorWidget::onSegmentationNodeChanged(vtkMRMLNode* node)
 {
   Q_D(qMRMLSegmentEditorWidget);
+  if (!this->mrmlScene() || this->mrmlScene()->IsBatchProcessing())
```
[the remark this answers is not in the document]
jamesobutler: @lassoan Do you think a comment is still necessary here? I'm not sure what I would write that wouldn't just rehash why batch processing state is used (to avoid unnecessary updates).

jamesobutler:
@lassoan could you take another look at this batch processing PR?

jamesobutler:
I've rebased to fix merge conflicts here.

jamesobutler:
Pushed fix for compile error due to change of `MasterVolumeNodeComboBox` to `SourceVolumeNodeComboBox`.

lassoan (approved):
It is hard to verify that there is no side effect of this, but if there were, they should be minor; and overall this is a useful change.

PR #6396, opened 2022-05-26
Allow multiple qSlicerWebWidget instance by injecting scripts into page

jcfr:
Since the script injection associated with `qSlicerWebWidget` was done globally
and only once at the profile level, instantiating first `qSlicerWebWidget`
and then `qSlicerExtensionsInstallWidget` was preventing the customization
done in `qSlicerExtensionsInstallWidget::initializeWebChannelTransport()`
from being effective.

This commit addresses this by injecting the `qwebchannel_appended.js` script
on per page basis. Doing so ensures each `qSlicerWebWidget` (and derived
object like `qSlicerExtensionsInstallWidget`) will inject the expected script.

It follows up on da201359b (ENH: WIP - WebEngine support for
extension manager).


pieper (approved):
Looks good as far as I can tell.  Let's merge and see what happens.

Issue #6379, opened 2022-05-15
Adding new markup control points is very slow if control points table is visible


lassoan:
Adding new markup control points is very slow if Markups module is active and control points table is visible.

See user error report at https://discourse.slicer.org/t/adding-points-to-point-list-markup-becomes-progressively-slower/23419/4

## Steps to reproduce

- Create a `Point list` markup
- Add 100 control points
- Add one more control point - it takes a fraction of a second
- Go to Markups module, open Control points section
- Add one more control point - it takes several seconds

## Expected behavior

Adding a control point should only take a fraction of a second.

## Environment
- Slicer version: Slicer-5.1.0-2022-05-08
- Operating system: Windows10

lassoan:
@smrolfe could you have a look at this? Update of the markups control points table is so slow when you have more than a few ten points that it becomes very hard to use it.

smrolfe:
Thanks, I will look into this.

jamesobutler:
Every row in the control points table is getting called to be updated when updateWidgetFromMRML is being processed. This is increasing the amount of processing required when the number of control points in the table gets larger.

https://github.com/Slicer/Slicer/blob/dcd132b264a741e01ae42ce35d774061bb8fc6c5/Modules/Loadable/Markups/qSlicerMarkupsModuleWidget.cxx#L1075-L1085

lassoan:
I'm experimenting with a model-view-based markup control point table (similar to segment list table). It should be much faster as the current one, but I will know more by tomorrow.

@smrolfe if you can figure out a quick and simple fix for the current code then it's good, but if it looks difficult then you may consider waiting for the result of my experiment before investing a lot of time in it.

smrolfe:
Thanks @lassoan sounds like a good plan. I'll update this thread if I find a quick fix this afternoon.

jcfr:
> model based table

That seems like a sensible path forward.


Once this is addressed, another good candidate for v5.0.3 patch release I think

lassoan:
I've implemented a [prototype for a control points table view](https://github.com/lassoan/Slicer/tree/attic/markups-control-points-table-view) based on how the segments table is implemented. It is nice that it makes search&filtering simpler and allows reuse of the table in other places, but it still creates all model items for each control point, which would make the table still too slow if we have many thousand control points. It would be better to if we implemented a special table model instead, similarly to pqSpreadSheetViewModel, which would only create items for the items requested by the view.

[...]

PR #6398, opened 2022-05-27
Maintenance 5.0 backports

jcfr:
Candidates bug fixes prior to releasing `v5.0.3`


jcfr:
cc: @lassoan @jamesobutler @pieper

jamesobutler (commented):
How about the additional commits?

3694f90688d2a4eed4ba33f4db4c44972114541b - Fixing build error on macOS

General documentation updates:
ec4e6ff7970908d5863968f8e86f7c6c6c691384
887143f2c510989f9a597ea89db78483931acbee
0a9adf0af52e713c6804b68de8cdf9e6fbbd9050
73ed5d4e506825997aac1f6a828cbc36655bf7c2
71d053f1fc9ce2649235334557d78ed3ef9d2d20
69c0fcfc71042ca384969b51672f97f1b533c7e8
3a63b6173cc911709e3caaf1e81ae90156099c61

1bdfefe4e275c9b5713503077930fe387aa8a324 - Fixes for custom app using stable releases

dcd132b264a741e01ae42ce35d774061bb8fc6c5 - Fix gcc build error
f6eb38811b642f629a34b2ded5e393f6c0d417d1 - Fix gcc build error

9a4d0d9b7305379c86636d8514a38d138b885bc1 - testing bug fix

650ed1467a3a6f59b6fe722e56ee8778d99e35e4 - CTK update with bug fix for clipped view content

43e8f5b2407b4c83509406b1acf266e25412eb49 - documentation clarification for Windows builds

Markups Control Point table performance improvements as mentioned in @jcfr's https://github.com/Slicer/Slicer/issues/6379#issuecomment-1129059520
423a07be7b542dc0a6bf313102c84e85bed91e7a
537eb93979d0e18e505dc2e79beb43c1811f2c85

More General documentation updates:
7ce77363e20ccc5f679dacd5fc816af7989395ba
0a71deaed03f69836a09986366a4085a12ef5a62
3858275bf1c29ad3cf57bb2f94f6b2d38b15efbb
33d92d52f547ae0f3455952d9451150c65f30200
5d8c7a4c20ca3ab2a48d5e3535ff980e5a597232

jcfr:
> build error fixes
I do not think these are relevant for the patch release since it already build on the factory.

> Markups Control Point table performance improvements as mentioned

These are indeed good candidates. 
@smrolfe  Did you have a chance to test these somewhat exhaustively ? 


> testing bug fix

:+1:  I will backport as this will avoid to have blocking test associated with stable extension build.

>  CTK update with bug fix for clipped view content

If the corresponding changes have been exhaustively tested, backporting would be sensible. it also means we would have to backport 06a7d6981e5d0aad72aab1713ed2dcaa8ae8b642

Could someone carefully review the changes and confirm ? 



> General documentation updates

Considering these are all related to building or debugging, I do not consider these as being relevant. I even think that we could:
- remove the developer section from the released documentation and suggest user to look at the latest version.

Fixes to links like what were done in https://github.com/Slicer/Slicer/commit/3a63b6173cc911709e3caaf1e81ae90156099c61 and https://github.com/Slicer/Slicer/commit/3858275bf1c29ad3cf57bb2f94f6b2d38b15efbb are more relevant if they pertain to user manual.

I am wondering if there would a way to systematically check links in markdown document ... or provide way to collect them and review.

smrolfe:
> @smrolfe Did you have a chance to test these somewhat exhaustively ?
> 
> >

Yes, I've tested with point lists with up to 4k points. Adding new markups with the control point table visible works well now.

lassoan:
Two more fixes that we should consider backporting:
- https://github.com/Slicer/Slicer/commit/3e7cf17cac10ce9f531f0691854f0f5a8a7242b7
- https://github.com/Slicer/Slicer/commit/197a96b2c7dc97a873e904e7f52c17fc572db18c

We see a good amount of feedback from users about the new stable and preview releases. It could make sense to wait a bit more (maybe a week?) before creating the patch release to allow some more reports to come in and to give some time to test the fixes.

jamesobutler:
Any additional commits being added here or is Slicer 5.0.3 going to be cut with this set of additional commits being added here to the maintenance branch?

jcfr:
Additional candidates:

```
BUG: Fix VTKObservationMixin preventing observed nodes from being released
BUG: Fix broken multiple inheritance in ScriptedLoadableModuleTest
BUG: Update onInstallDownloadFinished to ensure relevant metadata are filtered
BUG: Ensure depends retrieved from Girder_v1 is associated with model
BUG: Fix plane validity status computation in plane markup
BUG: Fix RemoveAllSegments function to actually remove the segments
BUG: Fix too wide qMRMLSegmentEditorWidget if segmentation or master volume name is long
BUG: Fix file loading using qSlicerIOManager::openDialog
BUG: Fix interactive slice intersection widget focus handling
BUG: Fix WebServer failing to load when Slicer built without DICOM support (#6407)
ENH: Update Slicer.crt CA bundle
BUG: Fix subject hierarchy item visibility change in batch mode
BUG: Fix copying of trimmed sequences
ENH: Only show load DICOM toolbar button when DICOM module loaded (#6420)
ENH: Prevent warnings by using the new variants of deprecated markups functions (#6451)
```

If the following are not related to recent features specific to nightly:

```
BUG: Fix extension update
```

Performance (may be):

```
ENH: Improve DICOMweb download speed
```

lassoan:
Are there any commits that we *don't* want to include (other than changing Slicer minor version and revision offset)?

The WebServer module (high value, very low risk of regressions) and the reworked Extensions manager (very high value, moderate risk of regressions) are the two main new features - I would not impede our users by unnecessarily delaying these for 3-4 months. The rest are lots of useful small fixes (medium value, low risk of regressions).

jcfr:
There are currently ~25 commit associated with the `maintenance-5.0-backports` branch (counting from `v5.0.0`), since we allocated 100 commit for the maintenance branch, we could then do the following:
- backport extensions improvements (and bug fixes) into own commit
- backport all documentation changes into own commit

These should allow to have 5.0.3 remaining in the allocated range.

lassoan:
I would vote for backporting everything by default (unless there is a reason to not to backport something), which - as I understand - is Approach 2. But I'm OK with a more selective approach if that's what others would mostly prefer.

The issue might have been that 5.0.0, 5.0.1, 5.0.2 were effectively release candidates, but we don't have this concept in Slicer's versioning scheme.

jcfr:
> re: Approach 2

I am almost done consolidating and re-organizing the backports.

pieper:
Yes, I think until we make some big change I think backporting all makes sense.  We used to do release candidates but it meant an extra step to change from rc to real release and it was never clear exactly when to do that.  Instead we can have releases and then it's pretty clear when a patch is needed.

jcfr:
Following comment https://github.com/Slicer/Slicer/pull/6455#issuecomment-1171402381, I will remove the following commits:

```
* 728befe882 - BUG: Fix loading of markups json from previous versions of Slicer
* e92f1b4c42 - ENH: Backport use of inclusive terminology & language
```

jamesobutler:
This all seems good to me. It has now been 2 months since Slicer 5.0.2 was tagged and other immediate issues haven't come up recently in the past week or two so integrating this now makes sense.

jcfr:
I will initiate the process release process this afternoon or tomorrow, and post an update on discourse.

jcfr:
Backported f968c2101d (`DOC: Add description of loading segmentation with color table`) and squashed it with `DOC: Backport documentation updates`

PR #6399, opened 2022-05-27
BUG: Ensure download request are handled by web widget triggering the request

jcfr:
This commit fixes a regression introduced in 9c0ae2ad8 (BUG: Allow multiple
`qSlicerWebWidget` instance injecting scripts into page) where the observation
of the profile `downloadRequested()` signal was done multiple time.

Since the download request is emitted by the default profile observed by
all web widget instances, this change ignores the request if it does not
originate from the page associated with this web widget instance.


pieper (approved):
LGTM 👍

PR #6401, opened 2022-05-29
DOC: Fix incorrect links

jcfr:
This commit fixes incorrect links identified using [markdown-link-check](https://github.com/tcort/markdown-link-check)

The problematic links have been identified running the following:

```
  cd Slicer
  for f in $(fd -e md); do
     echo $f | xargs npx markdown-link-check;
  done
```

Type of updates are the following:

* Obsolete links due to transition to GitHub
  -> Previous reference was looked in "SlicerGitSVNArchive" and then
     updated with the expected on in "Slicer"

* Inconsistent reference to local document
  -> Missing ".md" suffic have been added

* Non-existent website
  -> Replaced with reference from web.archive.org

* Redirect
  -> Replaced with current URL

* Incorrectly formatted link


Docs/user_guide/modules/brainsfit.md
```diff
@@ -10,7 +10,7 @@ Most frequently used for these scenarios and recommended registration settings.
 
 ### Same Subject: Longitudinal
 
-For this case we're registering a baseline T1 scan with a follow-up T1 scan on the same subject a year later.  The two images are available on the [Midas site](http://midas.kitware.com/item/view/483) as test.nii.gz and test2.nii.gz
+For this case we're registering a baseline T1 scan with a follow-up T1 scan on the same subject a year later.
```
jcfr, line 13: @hjmjohnson Is there an authoritative home for BRAINSTool testing data ? 

I found the file [SimpleITK-Notebooks/blob/master/Data/brains_manifest.json](https://github.com/InsightSoftwareConsortium/SimpleITK-Notebooks/blob/master/Data/brains_manifest.json) ... but looks like the collection on data.kitware.com is not public, data can only be referenced using the api.

For example, here is a summary for accessing `test.nii.gz`

| Status  | URL |
|--|--|
| :heavy_check_mark: | https://data.kitware.com/api/v1/file/hashsum/SHA512/038fa4ba7e74233e35052d87d0e525b2f66b78c0b43267b6ab29afffa7306ba31df69ab0445f6d6b5d04dc21a2eb599297d6387eae5fb452ab4189fbb13aa3a6/download | 
| :x: | https://data.kitware.com/#item/5bf963748d777f2179bc07bf | 
 
Last, should I look into migrating the following data (only accessible internally at Kitware):

image

CONTRIBUTING.md
```diff
@@ -20,14 +20,14 @@ The PR Process, Circle CI, and Related Gotchas
 If you are new to Slicer development and you don't have push access to the Slicer
 repository, here are the steps:
 
-1. [Fork and clone](https://help.github.com/articles/fork-a-repo/) the repository.
+1. [Fork and clone](https://docs.github.com/en/get-started/quickstart/fork-a-repo) the repository.
```
jamesobutler, line 23: I would suggest dropping the “/en/“ from several of these updated links so it doesn’t force users to the English versions of the page. It will attempt to go to whatever language their system is set to.

Docs/developer_guide/extensions.md
```diff
@@ -17,7 +17,7 @@ If you have developed a script or module that you would like to share with other
 
 :::{note}
 
-To build extensions that contain modules implemented in C++, you need to [build Slicer from source](build_instructions/index.md) on your machine; they cannot be built against a binary download.
+To build extensions that contain modules implemented in C++, you   need to [build Slicer from source](build_instructions/index.md) on your machine; they cannot be built against a binary download.
```
jamesobutler, line 20: Unnecessary extra space being added here?

CONTRIBUTING.md
lassoan: It is an interesting question. Slicer's documentation is currently in English only, so "choosing" English may not mean that the user actually prefers English, therefore it could make sense to avoid hardcoding `/en` in outgoing links. On the other hand, the localization efforts are ramping up, at some point we'll enable translating the documentation folder, too, and in a translated documentation it could be nice to propagate the language selection to external links.

Since in the short term (before Slicer documentation is translated) is the better solution and very simple, I agree that we can just go with that. If we get to translate Slicer documentation then we can revisit this (we may need to go through all external links to see which one support multiple languages, but there are not that many, so it should be manageable).

jamesobutler: I don’t think outgoing links would ever need to specify the language as that should be up to the user’s system language setting as used by their browser. If the user has manually selected to read Slicer documentation in say French, if it existed, even though their system is set to English, then an outbound link to GitHub.com documentation should respect the system language of English rather than trying to stay on the manual selection of French in Slicer documentation. I don’t think it’s expected for manual translation selection to persist across various domains.

lassoan: > I don’t think outgoing links would ever need to specify the language

docs.github.com handles language preferences very nicely. It does not look at your browser language preferences (maybe except the first time you open the site?), but instead it stores preferred language in a cookie. It even offers to switch to your preferred language in a top banner when you open a link with a different language hardcoded. If every multi-language sites worked like this then I agree that we should always just remove the language selection.

However, preferred or automatic language selection is not generally available. For example, readthedocs.org requires you to specify a language, [there is no way around it](https://docs.readthedocs.io/en/stable/localization.html#project-with-multiple-translations):

- https://docs.godotengine.org/en/stable/tutorials/assets_pipeline/import_process.html -> valid
- https://docs.godotengine.org/es/stable/tutorials/assets_pipeline/import_process.html -> valid
- https://docs.godotengine.org/stable/tutorials/assets_pipeline/import_process.html -> page does not exist

So, maybe we cannot specify a simple universal policy for external link language codes, but it depends on the site. For readthedocs based documentation sites we must include the language, but for docs.github.com it is probably better to remove it.

jcfr: @lassoan Would it make sense to collect this type of finding into a "Best Practice" section added to https://github.com/Slicer/SlicerLanguagePacks ?

jamesobutler: Thanks for your expertise on the specifics of the language support 👍🏻 . My knowledge is limited in just expecting some system level or other memory of setting to apply the selection. Including notes on this would be nice based on the relevant repository. Maybe a contributing doc update related to links in documentation.

jcfr: In the meantime, are we good moving forward with integration of the proposed changes ?

jamesobutler (approved):

CONTRIBUTING.md
lassoan: Yes, I think the proposed changes are good.

I've added a comment about translation of links [here](https://github.com/Slicer/SlicerLanguagePacks/blob/main/README.md#translation-of-external-links).

PR #6407, opened 2022-06-02
BUG: Fix WebServer failing to load when Slicer built without DICOM support

jamesobutler:
I discovered that the WebServer module fails to load if Slicer is built with the `Slicer_BUILD_DICOM_SUPPORT` option turned off because DICOMRequestHandler.py (part of WebServerLib) is imported at initialization which includes a top level import of `pydicom`. The `pydicom` module is part of the python-dicom-requirement superbuild project which is only included if `Slicer_BUILD_DICOM_SUPPORT` is turned on.

|With DICOM Support| This PR with no DICOM support|
|-----------------------|------------------------------------|
|image|image|

### DIFF
--- a/Modules/Scripted/WebServer/WebServer.py
+++ b/Modules/Scripted/WebServer/WebServer.py
@@ -11,8 +11,6 @@ import qt
 import slicer
 from slicer.ScriptedLoadableModule import *
 
-import WebServerLib
-
 #
 # WebServer
 #
@@ -114,7 +112,8 @@ class WebServerWidget(ScriptedLoadableModuleWidget):
 
     self.enableDICOMHandler = qt.QCheckBox()
     self.enableDICOMHandler.toolTip = "Enable serving Slicer DICOM database content via DICOMweb (stop server to change option)"
-    advancedFormLayout.addRow('DICOMweb API: ', self.enableDICOMHandler)
+    if hasattr(slicer.modules, "dicom"):
+      advancedFormLayout.addRow('DICOMweb API: ', self.enableDICOMHandler)
 
     self.enableStaticPagesHandler = qt.QCheckBox()
     self.enableStaticPagesHandler.toolTip = "Enable serving static pages (stop server to change option)"
@@ -168,7 +167,10 @@ class WebServerWidget(ScriptedLoadableModuleWidget):
     self.logToGUI.checked = slicer.app.userSettings().value("WebServer/logToGUI", True)
     self.enableSlicerHandler.checked = slicer.app.userSettings().value("WebServer/enableSlicerHandler", True)
     self.enableSlicerHandlerExec.checked = slicer.app.userSettings().value("WebServer/enableSlicerHandlerExec", False)
-    self.enableDICOMHandler.checked = slicer.app.userSettings().value("WebServer/enableDICOMHandler", True)
+    if hasattr(slicer.modules, "dicom"):
+      self.enableDICOMHandler.checked = slicer.app.userSettings().value("WebServer/enableDICOMHandler", True)
+    else:
+      self.enableDICOMHandler.checked = False
     self.enableStaticPagesHandler.checked = slicer.app.userSettings().value("WebServer/enableStaticPagesHandler", True)
 
   def updateGUIFromLogic(self):
@@ -206,7 +208,9 @@ class WebServerWidget(ScriptedLoadableModuleWidget):
     self.stopServer()
 
     packageName='WebServerLib'
-    submoduleNames=['DICOMRequestHandler', 'SlicerRequestHandler', 'StaticPagesRequestHandler']
+    submoduleNames=['SlicerRequestHandler', 'StaticPagesRequestHandler']
+    if hasattr(slicer.modules, "dicom"):
+      submoduleNames.append('DICOMRequestHandler')
     import imp
     f, filename, description = imp.find_module(packageName)
     package = imp.load_module(packageName, f, filename, description)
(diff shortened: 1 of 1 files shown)


pieper (approved):
Generally looks good to me, thanks.

It seems like we should at least log an error if the dicom handler is enabled in the settings but not available in the running app.

Also, checking for the `dicomDatabase` makes sense, but this isn't what we would do in C++ code (we'd use the `Slicer_BUILD_DICOM_SUPPORT` cmake-define) so I'm trying to think if we have something more declarative.

Neither of these is a big deal though, so if this change helps your use case we should go ahead and merge.

jamesobutler:
> It seems like we should at least log an error if the dicom handler is enabled in the settings but not available in the running app.

This would need the DICOM handler being disabled by default? I wouldn’t want an error logged under normal conditions of Slicer built with DICOM support.

> Also, checking for the `dicomDatabase` makes sense, but this isn't what we would do in C++ code (we'd use the `Slicer_BUILD_DICOM_SUPPORT` cmake-define) so I'm trying to think if we have something more declarative.

Yeah I initially considered the build time dependency of making WebServer a module built only with DICOM support enabled, but since only one of the WebServer options requires DICOM support I didn’t want to introduce that dependency.

I couldn’t find another usage of where a scripted module determines if DICOM functionality is available. I picked checking the `dicomDatabase` since that is what is directly used in the WebServer logic. There would seem to be a need for plugin framework to add DICOM functionality to a scripted module. In this PR I technically don’t register the “plugin” DICOM handler by checking the existence of `dicomDatabase` which may be the best option?

pieper:
> > It seems like we should at least log an error if the dicom handler is enabled in the settings but not available in the running app.
> 
> This would need the DICOM handler being disabled by default? I wouldn’t want an error logged under normal conditions of Slicer built without DICOM support.

It's a pretty edge case, so probably not worth worrying about.  If the slicer is build with no dicom, the dicomweb setting will never be created because the default is only created when the setting is first checked.  Sorry for brining this up.

> I couldn’t find another usage of where a scripted module determines if DICOM functionality is available.

I didn't either, so I think what you have is fine.  `slicer.app.dicomDatabase` is part of the public API so it seems like a fair check.  Options could be to try importing the DICOM module or checking if `slicer.modules.dicom` exists.

More generally I suppose the plugins should be registering themselves to avoid this kind of problem.  This is how it works for dicom plugins, editor effects, modules, etc and it would open the possibility for an extension or custom app to add a web handler cleanly.  We can save that for a future PR though if we don't need it now.

jamesobutler:
I didn’t try `slicer.modules.dicom` because I wasn’t sure if the DICOM module could possibly finishing loading after the WebServer module loading. I’m not entirely sure if `slicer.dicomDatabase` avoids this initialization order issue.

pieper:
> I didn’t try `slicer.modules.dicom` because I wasn’t sure if the DICOM module could possibly finishing loading after the WebServer module loading. I’m not entirely sure if `slicer.dicomDatabase` avoids this initialization order issue.

That's a good point.  Maybe trying to import DICOM is safest since that will only depend on paths set by the launcher.

pieper:
Also perhaps putting a helper method `slicer.util` would encapsulate whatever option is chosen so it could be changed later.  One way to do that would be to make the build settings available to python, so you could check `slicer.util.cmakeDefines. Slicer_BUILD_DICOM_SUPPORT` directly.

jamesobutler:
Even if I could get `Slicer_BUILD_DICOM_SUPPORT` any potential DICOM supporting module would need to be successfully selected to be loaded by Slicer at startup to enable this DICOM related request handler. Is there dependency between `slicer.dicomDatabase` and other DICOM related modules?

pieper:
We sort of reluctantly added `dicomDatabase` to the main app instance because there wasn't a clear way to expose it as a global singleton that was populated by the scripted module, so it could be present as a member variable other modules that use it are not loaded.  But I still think what you have is fine but that longer term we should have the plugins register themselves.  And I think we should consider exposing build settings to python code.  But neither of these need to hold up merging this PR.  Maybe give it a day in case others want to chime in.

lassoan:
As far as I remember, slicer.dicomDatabase may be None if DICOM support is enabled but you have never opened the DICOM module.

Exposing build options to Python would be nice and could be used here. Checking presence of DICOM module is a correct and safe way, too, as module loading and initialization order is well defined (all modules are loaded before module initializations are started).

We could also consider registering the web server plugin in the DICOM module.

jamesobutler:
> As far as I remember, slicer.dicomDatabase may be None if DICOM support is enabled but you have never opened the DICOM module.

@lassoan It appears that `slicer.dicomDatabase` still returns an object for a version of Slicer with DICOM support even though the DICOM module hasn't been opened yet and there is no valid DICOM database to be found.
```python
slicer.dicomDatabase
slicer.util.selectModule("DICOM")
```
image

lassoan:
OK. Maybe the database object is always created, there is just no database on disk (and so no database is conneced).

jamesobutler:
@lassoan Would you prefer that the check `hasattr(slicer, "dicomDatabase")` becomes `hasattr(slicer.modules, "dicom")`?

lassoan:
> Would you prefer that the check `hasattr(slicer, "dicomDatabase")` becomes `hasattr(slicer.modules, "dicom")`?

Both are fine. Maybe `hasattr(slicer.modules, "dicom")` is a bit higher level, more future-proof check of general DICOM support.

A more elegant solution would be if the DICOM module registered the DICOMWeb server plugin.

jamesobutler:
If have updated to use `hasattr(slicer.modules, "dicom")`.

pieper (approved):
Let's go ahead and finalize this, with the idea of adding a way to register custom handlers as a better option in the future.

lassoan:
> Let's go ahead and finalize this, with the idea of adding a way to register custom handlers as a better option in the future.

We can already register custom handlers (the one with the highest confidence gets to process the request). Therefore, the DICOM handler could be registered in the DICOM module - just somebody would need to take the time to move it there.

PR #6414, opened 2022-06-07
BUG: better control of OpenSSL dirs in multiconfig

kerim371:
Before that it was impossible to bundle SlicerJupyter
xeus error: 

```
"OpenSSL::Crypto" includes non-existent path
"C:/C/r/OpenSSL-install/$(Configuration)/include"
```

Now it is possible to pass `CMAKE_BUILD_TYPE` to overcome this

### DIFF
--- a/SuperBuild/External_OpenSSL.cmake
+++ b/SuperBuild/External_OpenSSL.cmake
@@ -317,14 +317,16 @@ this version of visual studio [${MSVC_VERSION}]. You could either:
 
     set(OpenSSL_DIR ${EP_SOURCE_DIR})
     set(_openssl_base_dir ${OpenSSL_DIR})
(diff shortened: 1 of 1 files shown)


fbordignon:
CMAKE_BUILD_TYPE is for single config generators and not used when generating for ms visual studio, for example. 
Suggestion:

```cmake
    if(DEFINED CMAKE_CONFIGURATION_TYPES)
      set(OpenSSL_DIR ${OpenSSL_DIR}/$<CONFIG>)
      set(_copy_release_directory 1)
    else()
```

this evaluates config at generation time and adds the respective paths to the respective configurations.

jcfr:
Closing. Superseded by https://github.com/Slicer/Slicer/pull/7240

PR #6082, opened 2021-12-14
Remove extensions manager midas3 support & update qSlicerExtensionsManagerModelTest

jcfr:
* [x] Add `Girder_v1` json files to https://github.com/Slicer/Slicer/tree/master/Base/QTCore/Testing/Data/Input
* [x] Upload corresponding extension  to https://github.com/Slicer/SlicerTestingData
* [x] Update [Base/QTCore/Testing/Cxx/CMakeLists.txt](https://github.com/Slicer/Slicer/blob/e2ea8afa61a4d4b0372f3650c7843c56b757a20c/Base/QTCore/Testing/Cxx/CMakeLists.txt#L9-L41) to download extension test packages
* [x] Update `qSlicerExtensionsManagerModelTest` to run test based on `serverAPI`

Issue #5908, opened 2021-09-28
Extension updates not found in stable version


muratmaga:
As described in https://discourse.slicer.org/t/cant-update-extensions-on-the-current-stable/19901

extension manager in the current stable does not find/install any updates for install extensions. If a new stable based on the new infrastructure will be out soon, then it is not very concerning. But it still needs to be checked.

sjh26:
@jcfr will this be an issue with the upcoming Slicer 5 stable?

jamesobutler:
@muratmaga You should take another look at this issue with the newly cut Slicer 5.0 stable release.

Issue #5502, opened 2021-03-03
Add automatic application update notification


lassoan:
## Is your feature request related to a problem? Please describe.

We are trying to move towards releasing a new stable version quarterly. However, users may not be aware that a new Slicer core is available. Users also stop getting extension updates if they are not using the latest stable release.

## Describe the solution you'd like

Implement a mechanism that checks if a new version of Slicer is available. The server that queries for new Slicer versions could also collect some statistics about which Slicer versions are used and how often.

## Additional context

https://discourse.slicer.org/t/updated-slicer-binary-automatic-notification/766
https://discourse.slicer.org/t/snapshot-release-updates-and-user-notification/16341

lassoan:
@jcfr do we know how the Slicer download site will work? It would be quite easy to implement this, but it does not make sense to implement it for the current download site if it will be replaced with another in the near future.

lassoan:
Cautionary tale about introducing telemetry feature into an open-source software without enough open communication with the community: https://www.zdnet.com/article/audacity-reverses-course-on-plans-to-add-opt-in-telemetry-after-outcry

jamesobutler:
See https://discourse.slicer.org/t/stable-update-notification/23655 for more brainstorming related to this issue.

lassoan:
@jcfr could you recommend an API to check for application updates? Ideally, it would be a web API call that would take the current application name, version, revision, and operating system as inputs and it would return information on the new release (version, revision, release webpage URL).

lassoan:
I've found that the download site already has API for this:

https://download.slicer.org/find?os=win&stability=release
https://download.slicer.org/find?os=macosx&stability=release
https://download.slicer.org/find?os=linux&stability=release

The referenced pull request uses this API to check if application update is available.

Issue #5470, opened 2021-02-18
Add "Update Extension" option under Manage Extensions 


muratmaga:
An addition the issue #5469, i think it would be good to have a mechanism to let users know when their installed extensions become stale for the stable versions. Something like this would be a good way:

![image](https://user-images.githubusercontent.com/21285140/108403161-45eb0380-71d3-11eb-8ca3-26cac3d33cbc.png)

lassoan:
This is already implemented:

![image](https://user-images.githubusercontent.com/307929/108404232-9d966880-71ed-11eb-8b6b-f600054d8970.png)

![image](https://user-images.githubusercontent.com/307929/108404244-a0915900-71ed-11eb-89df-3c52faad8425.png)

Previously there was no caching of server answers, which overwhelmed the extension manager and so we did not want to check for updates automatically (which would have further strained the server). Since I've added caching in the extensions manager a few months ago, everything is much faster and robust (see extensions manager in Slicer Preview Release), so maybe now we can enable automatic update check (or even automatic updates). I think the feature is still there, just disabled by default.

lassoan:
Auto-update of extensions at startup (display of a popup that updated extensions are available) is currently disabled, because it caused crashes - see https://github.com/Slicer/Slicer/issues/4641.

I think the crashes do not occur anymore, but to be safe, we could run the check when the extensions manager is opened.

muratmaga:
Didn't know. This is awesome. I would argue against auto-update of extensions. perhaps a check to notify a user, but don't update it without giving them consent. 

Also, is there a similar mechanisms exists for checking whether there are any back ports applied to stable since you install? I do not know how often that's the case, but I think you guys do it from time to time ?

lassoan:
> Also, is there a similar mechanisms exists for checking whether there are any back ports applied to stable since you install? I do not know how often that's the case, but I think you guys do it from time to time ?

We update extensions for latest stable release every night. This is exactly why the "Check for updates" feature is useful.

For preview releases, we only create updated extension packages for the latest version. If Slicer is not changed for a few days then it is possible that you happen to get some updates, but in general if you want to have updated extensions for a preview release then you need to upgrade the Slicer release itself.

pieper:
[...]

PR #6417, opened 2022-06-08
Refactored extensions manager

lassoan:
Reworked the extensions manager to take advantage of new Girder_v1 APIs:
- batch download of extension metadata
- server provides extension dependency metadata

These enabled implementing the following improvements:

- Automatic update check and automatic installation can now be enabled for extensions.
  Previously, hundreds of requests were submitted to query all necessary extension
  metadata, which caused slowdowns and crashes. Now all metadata can be downloaded in
  batch (in 4 requests), very robustly, in 1-2 seconds.
  For testing purpose, batch query may also be disabled with `setServerQueryWithLimitEnabled`.

- If updates are available then a notification marker is displayed over the extensions
  manager toolbar icon.

- Extensions can be bookmarked and all reinstalled in a different Slicer version by a
  single click.
  Previously, all extensions that the user have ever installed was offered to be restored, and there was no option to remove any of them. Bookmarked extensions can be installed without the browser widget, which may be useful in cases where proxy servers block the communication between the extensions manager frontend and backend (because only the backend needs to be accessed).

- List of bookmarks can be edited as a simple text. This makes it easy to transfer favorite extensions between computers or install a set of extensions at training events.

- Extensions Manager is displayed immediately. The user does not need to wait for any downloads.
  All data is reloaded immediately from cache or downloaded asynchronously.

- The "manage" and "restore" tabs are merged into a single tab where user can install/uninstall/update/restore.

- All metadata are downloaded from Extensions Server. The metadata is cached for offline access and to reduce extensions server load.

- Dependent extensions are installed automatically by default, to simplify extension installation.
  Manual installation (user confirmation) can be enabled in extensions settings.

- All dependent extensions (even indirect dependencies) are determined at once before asking the user to install them, saving several extra clicks for some extensions.

- Extension version and update time (when the package was built) is displayed for the installed extension, and also if there was any update.

- Close/restart button is not enabled until batch processing (install all/update all extension) are completed.

- Multiple files can be selected when installing extension from file.

- Extensions server website can be opened from the extensions manager. This makes it easier to
  download extensions for offline installation from file or when firewall or proxy settings prevent
  the extensions manager from working.

- Fixed display of icons of many extensions: icons that were hosted on servers with 301-redirect were not displayed.


lassoan:
Here is a demo video:

https://youtu.be/sKZGeko5wGE

@muratmaga please have a look, several features were motivated by your use cases and feature requests.

@pieper @cpinter @jamesobutler check this out and let me know if you have any comments.

lassoan:
Addresses most issues/feature requests listed at:
- https://github.com/Slicer/Slicer/issues/5908
- https://github.com/Slicer/Slicer/issues/5470
- https://github.com/Slicer/Slicer/issues/5341
- https://github.com/Slicer/Slicer/issues/4936

muratmaga:
@lassoan these look awesome. Thanks.

jamesobutler:
@lassoan I'm going to test this locally. Is `ENH: Refactor extensions manager` the only commit of interest here in this PR? The others are from some unrelated work?

muratmaga:
> This works so nicely that I would love to see this made available for the current stable release, and not wait another 3 months for Slicer-5.2. What do you think?

Yes, please. And the check for stable release updates as well (I liked the Welcome screen option)..

jamesobutler:
In general I'm usually strict in new features and enhancements only being in feature release versions as from a user perspective changes in functionality in a patch release like Slicer 5.0.3 is unexpected. If it was to be included then all the build improvements and documentation backports should be considered for inclusion (https://github.com/Slicer/Slicer/pull/6398#pullrequestreview-987540865) because the patch release would not be under strict consideration. If the desire is to get it out sooner than my choice would be to release the next feature release Slicer 5.2 sooner. This 5.2 version would just have less new features than a typical feature release.

jcfr:
> And the check for stable release updates as well

@muratmaga That specific commit (`BUG: Fix extensions update check`) has already  been back-ported on the maintenance branch. See https://github.com/Slicer/Slicer/pull/6398/commits/9c7dc54f1420cdefe6b25c541d04c67487d2d1c1 and https://github.com/Slicer/Slicer/pull/6398#commits-pushed-9fda87d

jamesobutler:
>  the check for stable release updates as well (I liked the Welcome screen option)..

@jcfr What is being referred to is https://github.com/Slicer/Slicer/issues/5502#issuecomment-1147006078 which is update about a new stable version of Slicer being available to download.

muratmaga:
> In general I'm usually strict in new features and enhancements only being in feature release versions as from a user perspective changes in functionality in a patch release like Slicer 5.0.3 is unexpected.

I am going to disagree with this, as we rely on update extensions functionality quite a bit to keep the SlicerMorph up to date with changes and additions. Currently this is broken, which I would say is an important issue for a stable that's going to be in place for next few months. From practice we know how long it takes to do a new release, even with best of our intentions. So this stable is going to be downloaded an used quite a while.

jcfr:
>  on update extensions functionality quite a bit [...] changes and additions. Currently this is broken

@muratmaga As indicated above by @jamesobutler , if you are talking about the issue we discussed during our last meeting, I this has been fixed and backported on the maintenance branch and will be part of `v5.0.3`. See https://github.com/Slicer/Slicer/pull/6417#issuecomment-1150350019

jamesobutler (commented):
- When I installed the "SegmentEditorExtraEffects" extension from the install tab it downloaded and install "SegmentEditorExtraEffects" and dependent extension "MarkupsToModel", but prior to restart the "Manage Extensions" tab indicated that there were 2 items, but there was only just the primary extension.  After restarting both extensions were present and the manage extension count number matched the number of items in the table.

| Before Restart | After Restart |
|-----------------|---------------|
|image|image|

**Update**: I think this was because it was still filtering using the terms I used for the "Install" extensions tab.
- I think having just the "Check for updates" button is sufficient where you can remove the action in the options menu to reduce the number of items there.
image

- When there are no bookmarks the "Install all bookmarks" button is disabled, however when then are no extensions the "Check for updates" button is still enabled. I would expect consistent logic applied where "Check for updates" is disabled until there are extensions to check for.
image

lassoan:
The patch makes it possible to update extensions, but without auto-update notification and the option being so much hidden, the feature is not practically available. We cannot expect that users will notice updates and update extensions.

I agree that there are tons of new features in this PR and lots of changes, so there is significant risk of regressions. These would be good reasons for excluding it from Slicer-5.0. However, preview releases don't have extension updates, so this feature could only be tested in a stable release. If we wait until Slicer-5.2, we just prevent from users accessing the new feature, but the feature will not get much testing, so it won't get more mature or more stable. Slicer5 has not been announced anyway yet, so our Slicer-5.0.x versions are kind of release candidates, so we don't violate our policies very harshly by allowing in some new features.

jamesobutler:
> Check for updates button must be always enabled because network connectivity comes and goes and the user must be able to trigger an update anytime.

Does the "Check for updates" button need to be enabled when there are no extensions to manage in the table?

jamesobutler (commented):
- [ ] Previously the "Install from file" toolbutton was a normal toolbutton of just an icon + tooltip. Was there a reason for including text on your new toolbuttons? Do you think the new icons don't convey clearly the actions that they represent? The "Check for updates" icon is familiar enough to me, though "Install bookmarked" is not recognizable to me. The general bookmark icon is not the base shape of it so I wouldn't expect it to be bookmarked related unless I read the text. Therefore the icon for "Install bookmarked" seems pointless at the moment.
image
image

jcfr:
I am finalizing https://github.com/Slicer/Slicer/pull/6082, once done. I will rebase this PR.

lassoan:
> Does the "Check for updates" button need to be enabled when there are no extensions to manage in the table? My suggestion would be to have it enabled when there is n>0 extensions.

Good point. Although clicking the button has a somewhat useful function (check if the server is available, if yes, download extension metadata, if not, display an error) there no visible result of the operation is expected, so it may make sense to disable. 

> * The general bookmark icon is not the base shape of it so I wouldn't expect it to be bookmarked related unless I read the text. Therefore the icon for "Install bookmarked" seems pointless at the moment.

All the buttons are standard google material design icons for update, install, and add operations, but their meaning are probably not obvious enough, and we have lots of space there, so adding the text is useful. Since we have the text, we could remove the icons, but I think they make the GUI look a bit nicer. If people don't like the icons then we can remove them.

lassoan:
> I am finalizing #6082, once done. I will rebase this PR.

Only the last commit is needed, the rest were just some unrelated experimental stuff from my development branch that I forgot to remove from my latest push (you can remove it). The only thing I still need to fix is the extensions manager test.

lassoan:
> I think having just the "Check for updates" button is sufficient where you can remove the action in the options menu to reduce the number of items there.

In the button row, "Check for updates" turns into "Update all" button. This is nice because it simplifies the button row.
I've added "Check for updates" to the dropdown menu because sometimes I wanted to check for updates again, even if some updates have been already found. We could remove the update check from the dropdown menu, but maybe we could all use it for a while and then finalize these kind of tweaks.

lassoan:
> Is there a functional need for the new option of "Automatically install all dependencies"? This seems like only allowing users to get into trouble? When would there be a reason to not install a dependency? Won't things not work? If things will work, then is it a true dependency?

This is an interesting topic. Currently, we cannot distinguish between optional and required dependencies, so we can either consider all dependencies are optional, or all dependencies are required. I've added a note to https://github.com/Slicer/Slicer/issues/3695 where we can discuss this topic further.

cpinter:
This is a great addition @lassoan !

> However, preview releases don't have extension updates, so this feature could only be tested in a stable release

The ultimate argument I think for including it in 5.0.3

> From practice we know how long it takes to do a new release, even with best of our intentions. So this stable is going to be downloaded an used quite a while.

Completely agreed. Users are very apparently annoyed with the current extensions manager (slow, a bit clumsy) and are requesting extension updates every week. Having them wait 6 months (realistacally, at least) for this when it's available would be a missed opportunity I think.

jamesobutler:
To allow users to use this update can we at least call it Slicer 5.2 instead of Slicer 5.0.3 to best follow [semantic versioning](https://semver.org/)?

cpinter:
In my opinion it would be an acceptable tradeoff @jamesobutler

jcfr:
> Re: test
> I spent time yesterday updating the tests and i am almost done.

See https://github.com/Slicer/Slicer/pull/6082

jcfr:
Also started to work on re-basing this topic on top of https://github.com/Slicer/Slicer/pull/6082

lassoan:
> To allow users to use this update soon can we at least call it Slicer 5.2.0 instead of Slicer 5.0.3 to best follow [semantic versioning](https://semver.org/)?

Semantic versioning is not applicable to end-user applications, because the "public API" is just so large and complex (for example, it includes the entire GUI). If you interpret the rules strictly then you always end up changing the public API, so you need to increase the major version of the application after almost every commit. It is not just my experience - see for example https://uglow.medium.com/making-sense-of-semantic-versioning-for-end-user-software-applications-a3049d97478b.

Therefore, what we do is not semantic versioning but:
- fixes -> increment tweak version
- features -> increment minor version
- "new generation" of software (marketing decision) -> increment major version

It is not always clear what is a fix and what is a feature (and we don't even attempt to give a clear definition for "new generation"), so there is some flexibility in when we increment which version component.

I agree that this change is more of a feature than a fix (although it mostly fixes broken features and usability problems), but by itself I don't feel it is significant enough to warrant an increment of the minor version.

pieper:
Yes, not so soon after the major release.

lassoan:
I've updated the documentation and tested on linux. On my build, clicking on the "Install" button in the "Install extensions" tab has no effect. @jcfr could you check if it works for you?

@pieper could you test this PR on macOS?

@jamesobutler Thanks for all your tests. Are there any issues that have not been properly addressed yet?

jcfr:
> could you check if it works for you?

:ok_hand:  I just rebased the topic, current in the process of testing and also working on fixing the tests.

Once I am done, I will likely need your help to further improve the tests.

jamesobutler (commented):
I'll continue to test once build errors with the extension manager tests are resolved here.

One thing I've realized as thinking about improvements to the usability of the extensions manager is that the search functionality is not prominent in the widget even though it is likely one of the most used features. It would be better to improve its visibility like how it is done for the Extensions Catalog.  Moving the search widget to be a big dumb search bar area for the left/center area would be an improvement. The current design of it being in the top right corner takes a while to find. Currently the size of the widget is also limiting where it can't even show me the full text of an extensions name such as "SegmentEditorExtraEffects".

Current Design:
image


Extensions Catalog:
image

jamesobutler:
Also I can't seem to install extensions from the Extensions Manager using the latest Slicer preview (Slicer 5.1.0-2022-06-10 76a231d). I keep getting this metadata failure. Maybe due to @jcfr 's recent commits for the extension manager.
image

jcfr:
I will have a look later this afternoon.

lassoan:
> One thing I've realized as thinking about improvements to the usability of the extensions manager is that the search functionality is not prominent in the widget even though it is likely one of the most used features.

Esthetically I find it really nice how that top section looks now. I find it intuitive, too - I start looking for the search box at the top and it is right there. The current implementation is very simple, too, because we rely on the "corner widget" feature of the QTabWidget. Maybe we should test this with a few people who are new to Slicer to confirm that they look for the search function elsewhere? We got an eye tracker recently, this could be a nice test for it to see where the gaze of users go when we tell them to find an extension. I would do this after this has been merged.  Maybe the search box could be made more prominent by giving it the focus to the search box by default 



^ This image reminds me to some naming inconsistency. We call the extensions manager frontend "Extensions Catalog", but in the Extensions Catalog the title of the page is "Slicer Extensions Manager". It would be better to change this text to "Extensions Catalog" to more clearly distinguish the "Extensions Manager" Slicer component from the "Extensions Catalog" website.

lassoan:
The name and design of Slicer's "Extensions manager" comes from Mozilla Firefox "Add-ons manager". Even though I slowly got used to the name, for me "Extension manager" would still sound better. Maybe @pieper and other native speakers could chime in.

jamesobutler:
I would agree that the singular "Extension Manager" seems more in line with other terminology I'm used to. I think the reason for the current name was probably that "well, you could be managing multiple extensions". I think the browsers Chrome and Edge (chromium based) have areas simply called "Extensions" where you go to "manage extensions" and from there they link you to the stores. 

I think people are familiar with the concept of single word terminology like "Settings->Extensions" for various apps. They generally don't have proper names as "Extension Manager". Chrome's "Manage extensions" area is for disable/enable and removing, but not for installing. So the "Extension Manager" may be good for those same type actions, but "Extension Catalog" for finding new extensions to install.

## Managing extensions
Chrome:
image
image
Edge:
image

## Finding new extensions from the store/catalog
https://chrome.google.com/webstore/category/extensions?hl=en
image
https://microsoftedge.microsoft.com/addons/Microsoft-Edge-Extensions-Home?hl=en-US
image

jcfr:
Summary of changes:
* Re-based
* Use of `convertExtensionMetadata` and `filterExtensionMetadata`
* Updated tests
* Added support for disabling batched query introducing `setServerQueryWithOffsetEnabled/serverQueryWithOffsetEnabled`, this was required to keep using a single file as URL substitute


Possible follow up changes:
* Add test for `loaded` and `bookmarked` states
* Consolidate `convertExtensionMetadata` and `filterExtensionMetadata`
* Revisit use of `DescriptionKey` name
* Simplify support for `waitForCompletion` using `qRestAPI::sync`


@lassoan After review the last commit adding `setServerQueryWithOffsetEnabled/serverQueryWithOffsetEnabled`, I will squash it.

Base/QTCore/qSlicerExtensionsManagerModel.cxx
```diff
@@ -243,6 +243,7 @@ class qSlicerExtensionsManagerModelPrivate
   QNetworkAccessManager NetworkManager;
 
   qRestAPI ExtensionsMetadataFromServerAPI;
+  bool UseOffset = true;
```
lassoan, line 246: useOffset

Base/QTCore/qSlicerExtensionsManagerModel.cxx
```diff
@@ -2057,6 +2058,20 @@ bool qSlicerExtensionsManagerModel::updateExtensionsMetadataFromServer(bool forc
   return completedSignalReceived;
 }
 
+// --------------------------------------------------------------------------
+bool qSlicerExtensionsManagerModel::serverQueryWithOffsetEnabled()const
```
lassoan, line 2062: It would be nicer to use the CTK get/set macros

Base/QTCore/qSlicerExtensionsManagerModel.h
```diff
@@ -399,6 +399,13 @@ public slots:
   /// Returns false if waitForCompletion is set to true and metadata cannot be retrieved.
   bool updateExtensionsMetadataFromServer(bool force=false, bool waitForCompletion=false);
 
+
+  /// \brief Enable/disable the use of offset for querying metadata from server.
+  ///
+  /// By default, default offset value is used to query the metadata by batch.
```
lassoan, line 405: Describe here the expected server behavior for the flag enabled and disabled. For enabled: query is repeated with increasing offset value, until 0 items are returned. For disabled I can just guess: one query is expected to return all items, offset value is ignored.

lassoan (commented):
Looks good, just added some trivial comments.

Base/QTCore/qSlicerExtensionsManagerModel.h
jcfr: When useOffset is false, the value -1 is passed meaning that all extensions are returns.

By default, the first query returns the default offset size (50 from memory) ... then the next offset is based on the number of extension already retrieved. 

I am wondering if we should instead always pass -1 and simplify the logic ? 

Though is nice to have a more robust approach that doesn't require an amount of memory depending on the number of extensions ...

lassoan: Based on my limited experience with web APIs, querying with offsets is a common pattern, probably because it is more robust (requires less memory, processing time, etc.) and more scalable (works for any number of items), allows progress reporting, etc. The client-side logic for multiple queries is quite simple and already implemented, so I think it is fine to keep its. It is nice to have the single-query option, too. So, we just need to explain in the documentation why there are these two different query modes, how they work, what are the advantages/disadvantages.

Base/QTCore/qSlicerExtensionsManagerModel.cxx
jcfr: Since all the other internal variable starts with an upper case letter, I suggest to stay consistent.

jcfr:
@lassoan This is now ready.

Once reviewed, I will squash the last three commits together.

I fixed the approach to instead pass `limit=-1` when `ServerQueryWithLimit` is `false`, this effectively ensure to have all data retrieved at once.

Base/QTCore/qSlicerExtensionsManagerModel.cxx
```diff
@@ -409,7 +410,15 @@ void qSlicerExtensionsManagerModelPrivate::queryExtensionsMetadataFromServer()
     parameters["app_revision"] = q->slicerRevision();
     parameters["os"] = q->slicerOs();
     parameters["arch"] = q->slicerArch();
-    parameters["offset"] = QString::number(this->ExtensionsMetadataFromServerQueryResults.size());
+    if (this->ServerQueryWithLimit)
+      {
+      parameters["offset"] = QString::number(this->ExtensionsMetadataFromServerQueryResults.size());
+      parameters["limit"] = QString::number(50); // corresponds to the default limit value also used on the server
```
lassoan, line 416: It is better to rely on the server's default, as only the server knows what limit it prefers. The client has no way of knowing what is an acceptable limit and it should not matter for the client either. So, I would remove this line, maybe just add a comment there explaining why we don't set the limit.

Base/QTCore/qSlicerExtensionsManagerModel.h
```diff
@@ -399,6 +399,17 @@ public slots:
   /// Returns false if waitForCompletion is set to true and metadata cannot be retrieved.
   bool updateExtensionsMetadataFromServer(bool force=false, bool waitForCompletion=false);
 
+  /// \brief Enable/disable the use of limit and offset parameters for querying extension metadata from the extensions server.
+  ///
+  /// If enabled (the default), the metadata are retrieved by batch of at most 50 extensions until
```
lassoan, line 404: If enabled (the default), the server may return metadata for a limited number of extensions at once (for example, 50) to keep the server's load and response time low. The model automatically submits additional queries until all data is retrieved.

jcfr:
After I am done rebuilding and doing a last round of testing, I will squash and integrate.

jcfr:
@lassoan The last commit I pushed ensure install works when AutoCheck has not been enabled already.

Base/QTCore/qSlicerExtensionsManagerModel.cxx
```diff
@@ -1693,7 +1693,7 @@ bool qSlicerExtensionsManagerModel::downloadAndInstallExtension(const QString& e
   // Ensure extension metadata has been downloaded from the server
   if (d->ExtensionsMetadataFromServer.empty())
     {
-    this->updateExtensionsMetadataFromServer(false, true);
+    this->updateExtensionsMetadataFromServer(/* force= */true, /* waitForCompletion= */true);
```
lassoan, line 1696: Does "force" trigger a new download of all metadata of all extensions? That would add unnecessary delay and server load every time an extension is installed. We should only query the server if no metadata is available.

When the extension manager is opened, all extensions metadata are downloaded automatically. Have you run into this issue when using the Slicer GUI or using the extension manager model API from code?

jcfr: > Does "force" trigger a new download of all metadata of all extensions? 

It does not because `updateExtensionsMetadataFromServer` is only called from `downloadAndInstallExtension` when `ExtensionsMetadataFromServer` is empty. 

> When the extension manager is opened, all extensions metadata are downloaded automatically.

This would be the case with `AutoCheck` is enabled, I was testing in the case of a fresh install of Slicer without any settings set. 

In that case, the download of metadata need to happen once when the user click the first time on the "Install" button (associated with `downloadAndInstallExtension`).

jcfr: > Have you run into this issue when using the Slicer GUI

This happened when using the GUI

lassoan: Ok, it's all good then.

I wanted to implement downloading of all extensions metadata when the user opens the extensions manager, but apparently I did not get to it.

jcfr:
@jamesobutler  I plan on integrating this later this evening before the round of nightly build, if you have a chance to quickly check on your side as well. Would be nice :pray:

jcfr:
Merci 🙏

jcfr:
Thanks for the feedback.  

Regarding the mixed content warnings, I indeed have a work in progress notebook going though all extensions and s4ext and effectively listing and/or fixing them.

Thanks for the reminder, it would indeed be nice to address this before announcing the release more broadly.

lassoan:
Probably the icons are not taken from the s4ext file in the extension index but from the top-level CMakeLists.txt file.

BreastImplantAnalyzer has an invalid icon: https://github.com/lancelevine/SlicerBreastImplantAnalyzer/blob/master/CMakeLists.txt

lassoan:
I had a look at a couple of extensions and submitted a few PRs to fix icons - they are all some trivial issues (using outdated Slicer wiki links, using the github file's page instead of the download link, etc).

PR submitted:
- https://github.com/AIM-Harvard/SlicerRadiomics/pull/73
- https://github.com/ETS-vis-interactive/SlicerPRISMRendering/pull/21
- https://github.com/hina-shah/SlicerBatchAnonymize/pull/17
- https://github.com/DCBIA-OrthoLab/ShapeVariationAnalyzer/pull/66
- https://github.com/SNRLab/SNRMeasurement/pull/4
- https://github.com/lassoan/SlicerDRRGenerator.git

@jcfr could you take care of these Kitware extensions?
- https://github.com/NIRALUser/SPHARM-PDM/blob/master/CMakeLists.txt
- https://github.com/KitwareMedical/OsteotomyPlanner/wiki

@jamesobutler could you help submitting PRs to fix icons of these extensions? (maybe not all, just as much time as you can contribute)
- https://github.com/gsi-biomotion/Slicer-RegQA.git
- https://github.com/AldrickF/SlicerRegularizedFastMarching
- https://github.com/Ting-Jia/ModelClip.git
- https://github.com/JonathanRamos/SlicerCleverSegmentation.git
- https://github.com/lancelevine/SlicerBreastImplantAnalyzer.git
- https://github.com/CSIM-Toolkits/SlicerBrainVolumeRefinement.git
- https://github.com/DCBIA-OrthoLab/DatabaseInteractorExtension.git

jamesobutler:
> Probably the icons are not taken from the s4ext file in the extension index but from the top-level CMakeLists.txt file.

Are the icons for the ExtensionsManager/ExtensionsCatalog supposed to be using the links specified in the s4ext files? Is the icon url specified in the s4ext in the Extensions Index used for anything?

lassoan:
> Are the icons for the ExtensionsManager/ExtensionsCatalog supposed to be using the links specified in the Extensions Index s4ext files? Is the icon url specified in the s4ext in the Extensions Index used for anything?

The s4ext file in the ExtensionsIndex repository is only used by the factory machine for getting repository URLs and dependencies. All other information is ignored.

The Extensions Server does not know about s4ext files in the ExtensionsIndex repository, but instead it populates its database from the `ExtensionName.zip/share/Slicer-5.0/ExtensionName.s4ext` file of each uploaded package. This s4ext file created from the content of the CMakeLists.txt file during extension build.

I find the redundancy between the CMakeLists.txt and s4ext files in ExtensionsIndex quite confusing and it is a frequent cause of errors. Maybe we could remove all the other fields from these s4ext files to avoid redundancy. Or we could replace the s4ext files by a single json or csv file in the extensions index that contains the list of extension names, repository urls, and dependencies. That said, I would do this only when we touch the extensions infrastructure anyway. I've created an issue to keep track of this: https://github.com/Slicer/Slicer/issues/6434

lassoan:
Since this PR is closed and we already have an issue for the extension icons, I copy the details to that ticket (as a checklist) and we can continue the discussion there:

https://github.com/Slicer/Slicer/issues/4936

jamesobutler:
> @jamesobutler could you help submitting PRs to fix icons of these extensions? (maybe not all, just as much time as you can contribute)

@lassoan I've completed submitting PRs to all the ones you assigned to me. I've provided links to the PRs in #4936 where it is being tracked.

lassoan:
Awesome, you @jamesobutler. Hopefully maintainers will respond to our PRs.

PR #6420, opened 2022-06-09
ENH: Only show load DICOM toolbar button when DICOM module loaded

jamesobutler:
When the DICOM module was not loaded:

| Before | This PR |
|---------|---------|
|image|image|


pieper (approved):
Makes sense to me.  It would be cleaner of course if the DICOM module added the button but I realize things aren't set up that way so this is a good easy fix.

PR #6428, opened 2022-06-14
BUG: Fix RemoveAllSegments function to actually remove the segments

cpinter:
The segment ID list was called early in vtkSegmentation::RemoveAllSegments, and the segments were thus not properly removed. Besides the memory leak, this caused bugs related to expecting the SegmentRemoved event, such as that the segments were not removed from Subject Hierarchy when using said function.

Fixing segmentation automated test by setting selected segment explicitly, which selection was lost due to now properly removing the segments.


cpinter:
Can someone please take a look at this one-liner fix please? Thank you!

lassoan (approved):
Looks good to me, thank you.

PR #6430, opened 2022-06-15
Fix bugs in qSlicerIOManager::openDialog and interactive slice intersection widget focus handling

lassoan:
Two small fixes, one bug reported by a user, another discovered by @jcfr when fixing build warnings.


Libs/MRML/DisplayableManager/vtkMRMLSliceIntersectionWidget.cxx
```diff
@@ -318,12 +318,17 @@ bool vtkMRMLSliceIntersectionWidget::CanProcessInteractionEvent(vtkMRMLInteracti
...
   if (this->WidgetState == WidgetStateMoveCrosshair
     || this->WidgetState == WidgetStateTranslate
     || this->WidgetState == WidgetStateRotateIntersectingSlices
     || this->WidgetState == WidgetStateBlend
     || this->WidgetState == WidgetStateTranslateSlice
-    || this->WidgetState == WidgetStateZoomSlice)
+    || this->WidgetState == WidgetStateZoomSlice
+    || this->WidgetState == WidgetStateTranslateIntersectingSlicesHandle
+    || this->WidgetState == WidgetStateRotateIntersectingSlicesHandle
+    || this->WidgetState == WidgetStateTranslateSingleIntersectingSliceHandle
```
jcfr, line 330: Should `WidgetStateTouchGesture` and `WidgetStateOnTranslateSingleIntersectingSliceHandle` be specified as well ? 

If yes, then checking `this->WidgetState < WidgetState_Last` would be more robust ?

Modules/Loadable/Volumes/qSlicerVolumesReader.cxx
```diff
@@ -173,9 +173,9 @@ bool qSlicerVolumesReader::load(const IOProperties& properties)
     fileList.GetPointer());
   if (node)
     {
-    if (properties.contains("colorNodeID"))
+    QString colorNodeID = properties.value("colorNodeID", QString()).toString();
```
jcfr, line 176: I suggest to update the commit message to also indicate that the volume reader logic checking `colorNodeID` has been improved.

Libs/MRML/DisplayableManager/vtkMRMLSliceIntersectionWidget.cxx
lassoan: We only want to unconditionally claim the events if we are in the middle of a drag-and-drop operation. 

`WidgetStateOnTranslateSingleIntersectingSliceHandle` just means that the mouse is over a handle, it should not capture events.

`WidgetStateTouchGesture` is for handling touch events. I think events during touch gesture state are intentionally not captured (to allow processing some touch gestures in this displayable manager, but letting other displayable managers to process some other touch events).

Modules/Loadable/Volumes/qSlicerVolumesReader.cxx
lassoan: OK!

PR #6431, opened 2022-06-15
ENH: Improve robustness of condition check if testing is enabled

jcfr:
This commit ensures that the code path expected to be executed when testing
is enabled is done so when the AA_EnableTesting attribute is set and not
only when the command line parameter "--testing-enabled" is passed.

This allows to programmatically enable testing in test without having to
explicitly pass the argument.


jcfr:
Locally tested to ensure the returns value by using `testAttribute` is consistent.

PR #6432, opened 2022-06-15
DOC: Update build instructions to include CentOS 7

jcfr:
Slicer built on CentOS 7 will be available for many Linux distros.

At the time of this commit, Slicer was successfully build, packaged
and ran on:
* Debian 10.12
* Debian 11.3
* Ubuntu 18.04
* Ubuntu 20.04
* Ubuntu 22.04
* Fedora 34

To support running on Debian 10.12, the "Getting Started" section
is also updated to document tricks needed to solve `xinerama` issue.

See https://discourse.slicer.org/t/build-and-package-slicercat-on-manylinux/23443/3

### DIFF
--- a/Docs/developer_guide/build_instructions/linux.md
+++ b/Docs/developer_guide/build_instructions/linux.md
@@ -85,6 +85,29 @@ Install the development tools and the support libraries:
 sudo pacman -S git make patch subversion gcc cmake \
   qt5-base qt5-multimedia qt5-tools qt5-xmlpatterns qt5-svg qt5-webengine qt5-script qt5-x11extras libxt
 ```
+
+### CentOS 7
+:::{note}
+Slicer built on CentOS 7 will be available for many Linux distributions and releases
+:::
+
+Install Qt and CMake as described in [Any Distribution](./linux.md#any-distribution) section.
+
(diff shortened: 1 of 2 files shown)


Docs/developer_guide/build_instructions/linux.md
```diff
@@ -85,6 +85,27 @@ Install the development tools and the support libraries:
...
 ```
+
+### CentOS 7
+:::{note}
+Slicer built on CentOS 7 will be available for many Linux distributions and releases
+:::
+
+Install Qt and CMake as described in [Any Distribution](./linux.md#any-distribution) section.
+
+CentOS 7 comes with pretty old devtoolset (gcc 4.8.5). Install and activate newer toolset (11 in this case):
```
jcfr, line 96:
```suggestions
Since by default CentOS 7 comes with `gcc 4.8.5` having [experimental support for C++14](https://gcc.gnu.org/onlinedocs/gcc-4.8.5/gcc/C-Dialect-Options.html#C-Dialect-Options), the following allows to install and activate the `devtoolset-11` [providing](https://access.redhat.com/documentation/en-us/red_hat_developer_toolset/11/html/11.0_release_notes/dts11.0_release#Changes_in_DTS)  `gcc 11.2.1`
```

jcfr: @kerim371  Could you review the proposed change ?

kerim371: Probably pretty soon Slicer will require C++17 standard and this sentence will be inactual.
Thus I think it would be better to mention that [gcc 11 supports C++20](https://en.cppreference.com/w/cpp/compiler_support) while the minimal required C++ standard to compile Slicer is 14 for now (as I remember).

kerim371 (commented):
If you don't mind please fix this section

jcfr:
Thanks for the feedback :pray:  , this should now be ready for integration

kerim371 (approved):

Issue #6437, opened 2022-06-19
fail to build master branch due to fatal error: QSignalSpy: No such file or directory


hubutui:
## Summary

fail to build master branch:
```
/build/3dslicer-git/src/3dslicer/Base/QTCore/qSlicerExtensionsManagerModel.cxx:31:10: fatal error: QSignalSpy: No such file or directory
   31 | #include <QSignalSpy>
      |          ^~~~~~~~~~~~
compilation terminated.
```
see also the complete build log [here](https://github.com/Slicer/Slicer/files/8934955/log.txt.tar.gz). You could jump to the line 72251 for the error messages.

Maybe we forget to add `testlib` according to Qt official [doc](https://doc.qt.io/qt-5/qsignalspy.html)?

## Steps to reproduce

I build 3dslicer with this [PKGBUILD](https://github.com/archlinuxcn/repo/blob/8c62f8b6aca24634780329e42c49286236610ebf/archlinuxcn/3dslicer-git/PKGBUILD) on ArchLinux with minor modification. I use
```
make -C "build" VERBOSE=1 
```
to print more information.

## Environment
- Slicer version: the master branch, date 20220619
- Operating system: ArchLinux

pieper:
Yes, I ran into this too, and it's because `QSignalSpy` was added in Qt 5.14 but some linux distributions have an older Qt by default (ubuntu 20.04 is on Qt 5.12 for instance).

These instructions describe how to build on linux with the latest Qt:

https://slicer.readthedocs.io/en/latest/developer_guide/build_instructions/linux.html#any-distribution

As a workaround, I just commented out the `QSignalSpy` for a local developer build since extensions aren't needed anyway.  

I don't know about Qt>=5.14 on ArchLinux, but hopefully we'll all move to Qt6 before too long.

https://github.com/Slicer/Slicer/issues/6388

I'm going to close this since I don't believe we should do anything about it, but if others feel differently we could reopen.  Perhaps an ifdef version check so the code still compiles by default would be worth it.

hubutui:
ArchLinux provides [qt 5.15.5+kde+r163](https://archlinux.org/packages/extra/x86_64/qt5-base/). The qt source upstream is https://invent.kde.org/qt/qt.
Qt 6.3.1 is also available. I would follow the workaround until we could build with Qt 6.

jamesobutler:
[...]

PR #6440, opened 2022-06-20
ENH: Do not use QSignalSpy in qSlicerExtensionsManagerModel

lassoan:
QSignalSpy is intended for testing, therefore use qRestAPI's synchronous query instead to get extensions metadata from the server.

To simplify the implementation of synchronous query, extension metadata is now always downloaded using a single query (by specifying `limit=0` query parameter).

see #6437

### DIFF
--- a/Base/QTCore/qSlicerExtensionsManagerModel.cxx
+++ b/Base/QTCore/qSlicerExtensionsManagerModel.cxx
@@ -2007,7 +2007,7 @@ bool qSlicerExtensionsManagerModel::updateExtensionsMetadataFromServer(bool forc
       }
     else
       {
-      qWarning() << "Update extension information from server failed: missing implementation for serverAPI " << this->serverAPI();
+      qWarning() << "Update extension information from server failed: missing implementation for serverAPI" << this->serverAPI();
       return false;
       }
 
--- a/Base/QTCore/qSlicerExtensionsManagerModel.cxx
+++ b/Base/QTCore/qSlicerExtensionsManagerModel.cxx
@@ -28,11 +28,11 @@
 #include <QNetworkRequest>
 #include <QNetworkReply>
 #include <QScopedPointer>
-#include <QSignalSpy>
 #include <QSettings>
 #include <QStandardItemModel>
 #include <QTemporaryFile>
 #include <QTextStream>
+#include <QTimer>
 #include <QUrl>
 #include <QUrlQuery>
 
@@ -164,9 +164,6 @@ public:
 
   int role(const QByteArray& roleName);
 
-  /// This method should only be called after initialization.
-  void queryExtensionsMetadataFromServer();
-
   /// Save/load extensions metadata that was retrieved from the server to a local cache
   /// (in application settings) to avoid too frequent polling of the server.
   void saveExtensionsMetadataFromServerToCache();
@@ -243,9 +240,7 @@ public:
   QNetworkAccessManager NetworkManager;
 
   qRestAPI ExtensionsMetadataFromServerAPI;
-  bool ServerQueryWithLimit = true;
   QMap<QString, QVariantMap> ExtensionsMetadataFromServer;
-  QList<QVariantMap> ExtensionsMetadataFromServerQueryResults;
   QUuid ExtensionsMetadataFromServerQueryUID;  // if not null then it means that query is in progress
 
   QHash<QString, UpdateDownloadInformation> AvailableUpdates;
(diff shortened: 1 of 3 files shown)


Base/QTCore/qSlicerExtensionsManagerModel.cxx
```diff
@@ -2038,35 +1999,47 @@ bool qSlicerExtensionsManagerModel::updateExtensionsMetadataFromServer(bool forc
         return false;
         }
       d->ExtensionsMetadataFromServerAPI.setServerUrl(this->serverUrl().toString() + QString("/api/v1/app/%1/extension").arg(appID));
+      parameters["app_revision"] = this->slicerRevision();
+      parameters["os"] = this->slicerOs();
+      parameters["arch"] = this->slicerArch();
+      // request all metadata in a single response (it makes synchronous query simpler)
+      parameters["limit"] = QString::number(0);
```
jcfr, line 2006: This makes sense, I was initially setting the value to `-1`

> limit (optional): the maximum number of results to return. A limit of 0 (the default) is equivalent to setting no limit.

Source: https://pymongo.readthedocs.io/en/stable/api/pymongo/collection.html#pymongo.collection.Collection.find

Base/QTCore/qSlicerExtensionsManagerModel.cxx
```diff
@@ -2038,35 +1999,47 @@ bool qSlicerExtensionsManagerModel::updateExtensionsMetadataFromServer(bool forc
...
       }
     else
       {
-      qWarning() << "Update extension information from server failed: missing implementation for serverAPI" << this->serverAPI();
+      qWarning() << "Update extension information from server failed: missing implementation for serverAPI " << this->serverAPI();
```
jcfr, line 2010:
```suggestion
      qWarning() << "Update extension information from server failed: missing implementation for serverAPI" << this->serverAPI();
```

> This syntax inserts a space between each item, and appends a newline at the end.

Source: https://doc.qt.io/qt-5/qtglobal.html#qWarning

Base/QTCore/qSlicerExtensionsManagerModel.cxx
lassoan: I had to set the limit to 0 because setting it to -1 makes it return a single item: https://slicer-packages.kitware.com/api/v1/app/5f4474d0e1d8c75dfc705482/extension?app_revision=30338&os=win&limit=-1

jcfr: And since the json used for testing only include one item, I initially missed that.

jcfr: This is also confirmed using either:
* https://slicer-packages.kitware.com/api/v1/app/5f4474d0e1d8c75dfc705482/extension?os=linux&arch=amd64&app_revision=31036&limit=0
* https://slicer-packages.kitware.com/api/v1/app/5f4474d0e1d8c75dfc705482/extension?os=linux&arch=amd64&app_revision=31036&limit=-1

Base/QTCore/qSlicerExtensionsManagerModel.cxx
```diff
@@ -2038,35 +1999,47 @@ bool qSlicerExtensionsManagerModel::updateExtensionsMetadataFromServer(bool forc
...
-  // Wait up to 15 seconds for the update check to complete
-  // (typically takes only a few seconds)
-  int timeoutMsec = 15000;
-  bool completedSignalReceived = spy.wait(timeoutMsec);
-  if (!completedSignalReceived)
+  // Temporarily disable onExtensionsMetadataFromServerQueryFinished call via signal/slot
+  // because we'll call it directly to get returned result.
+  QObject::disconnect(&d->ExtensionsMetadataFromServerAPI,
+    SIGNAL(finished(QUuid)),
+    this, SLOT(onExtensionsMetadataFromServerQueryFinished(QUuid)));
```
jcfr, line 2027: Instead of disconnect/connect, instantiating and using `qGirderAPI` and conditionally calling `.sync()` would also work.

Since until recently, we were supporting both midas and girder, the querying was done using the more generic `qRestAPI` along with explicit handling of the parsing using the static helper function like `qGirderAPI::parseGirderAPIv1Response`.

Now that we fully transitioned, using an instance of `qGirderAPI` would ensure the virtual function `qGirderAPI::parseResponse` is used ....

```diff
- qRestAPI ExtensionsMetadataFromServerAPI;
+ qGirderAPI ExtensionsMetadataFromServerAPI;
```

jcfr: This could also be done at a later time as this PR addresses the problem.

lassoan: I've tried using sync but there were two errors:
1. If the `finished()` signal is not disconnected then `onExtensionsMetadataFromServerQueryFinished` is called (which takes the result) and that makes `sync()` crash.
2. Sync cannot return a `qRestResult`, only a `QList<QVariantMap>`, so I would have needed to implement two different result parsing mechanisms (one for sync and another one for async).

`takeResult` waits for the result anyway (by calling `waitForDone()`) therefore using `sync()` would have just complicated things.

lassoan:
@jcfr if you are OK with this version then I squash the commits and merge.

PR #6441, opened 2022-06-20
BUG: Fix copying of trimmed sequences

### LINKED ISSUE #6435: vtkMRMLSequenceNode::Copy not working on trimmed sequences
## Summary

The vtkMRMLSequenceNode::Copy method does not correctly copy sequences after the RemoveDataNodeAtValue function has been used to remove data nodes from the beginning of the sequence.

## Steps to reproduce

1. Create a new LinearTransform in the Transforms module (for example).
2. Create a new sequence node in the Sequences module.
3. Add the LinearTransform as a data node to the sequence (will be added at index value 0).
4. Add the LinearTransform as a data node to the sequence (will be added at index value 1).
5. Remove the data node at index value 1 from the sequence.
6. Use the commands below to copy the sequence.
7. Inspect the copied sequence in the Sequences module (it will have one entry in it, but both the index value and data node will be empty).

```
trimmedSequence = slicer.mrmlScene.GetNodeByID("vtkMRMLSequenceNode1") # The sequence you just created
copySequence = slicer.vtkMRMLSequenceNode()
copySequence.Copy(trimmedSequence)
slicer.mrmlScene.AddNode(copySequence)
```

## Expected behavior

The copied sequence should contain one entry with index value 1, and the data node should be the LinearTransform.

## Environment
- Slicer version: Slicer-5.1.0-2022-06-15

My initial impression is that this may be due to an assumption in the Copy function that the data nodes in the original sequence and the copied sequence are assumed to have the same IDs in their respective SequenceScenes, but this is not always the case, including when a sequence has been trimmed at the beginning. See: https://github.com/Slicer/Slicer/blob/eb82cc0e2a104043dbdb6ad5c042aa4104f3e9ed/Libs/MRML/Core/vtkMRMLSequenceNode.cxx#L270

lassoan:
fixes #6435

### DIFF
--- a/Libs/MRML/Core/vtkMRMLSequenceNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLSequenceNode.cxx
@@ -245,6 +245,9 @@ void vtkMRMLSequenceNode::Copy(vtkMRMLNode *anode)
     }
   this->SequenceScene=vtkMRMLScene::New();
 
+  // Get data node ID in the target scene from the data node ID in the source scene
+  std::map< std::string, std::string > sourceToTargetDataNodeID;
+
   if (snode->SequenceScene)
     {
     for (int n = 0; n < snode->SequenceScene->GetNodes()->GetNumberOfItems(); n++)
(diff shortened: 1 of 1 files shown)


Sunderlandkyl:
lgtm!

PR #6446, opened 2022-06-22
ENH: Remove custom itkDCMTKFileReader from PETStandardUptakeValueComputation

lassoan:
itkDCMTKFileReader was added in 2012 but it seems that current itkDCMTKFileReader in ITK works just as well.
The 2012 version has a bug (it should use (0x5200, 0x9230) instead of (0x5200, 0x9239) tag - see https://discourse.slicer.org/t/image-position-patient-attribute-0020-0032-in-dicom-file-not-handled/23950/3) therefore it makes sense to switch to the ITK implementation.


jcfr:
:+1:  

The copy of the files was introduced in 7868d8237 back in 2012, since then we have transitioned to the latest ITK and removing our copy makes sense.

jcfr (changes requested):
Consider also updating the test removing references to `   ADDITIONAL_SRCS itkDCMTKFileReader.cxx`

https://github.com/Slicer/Slicer/blob/1026bb846d29fe9099de9ec002e332168da8392a/Modules/CLI/PETStandardUptakeValueComputation/Testing/CMakeLists.txt#L12

pieper (approved):
This makes sense to me.  @hjmjohnson do you have any thoughts?

jcfr (approved):

Issue #6447, opened 2022-06-22
Invalid error from vtk causes vtp file not to load


pieper:
## Summary

A file that loads in slicer 4.11 does not load in 5.x due to invalid logic in Slicer and VTK.

## Steps to reproduce

The attached zip file contains a vtp file that has no points.  This is a valid file that is used in SlicerDMRI to indicate that no data was detected for the brain region that the file corresponds to (this is part of the [WhiteMatterAnalysis](https://github.com/SlicerDMRI/whitematteranalysis) pipeline).  The convention has been used for many years with no problem across several versions of Slicer and VTK.

Now when you try to load as a model (or fiber bundle with SlicerDMRI installed) you get an error:

![image](https://user-images.githubusercontent.com/126077/175128238-c30fb1d3-3a3b-48e0-b30f-eaf6cf19ea2a.png)

The full text is:

```
- Error: Loading /opt/data/SlicerDMRI/vtk9-fail-T_Sup-O_right.vtp - ERROR: In /opt/s/VTK/IO/XML/vtkXMLUnstructuredDataReader.cxx, line 393
vtkXMLPolyDataReader (0x7fe30b116e30): No Points element available in first piece found in file. Reading file may fail.
- Error: Loading /opt/data/SlicerDMRI/vtk9-fail-T_Sup-O_right.vtp - Failed to read node vtk9-fail-T_Sup-O_right_2 (vtkMRMLModelNode6) from filename='/opt/data/SlicerDMRI/vtk9-fail-T_Sup-O_right.vtp'
```

But this is a false negative generated by the catching the `vtkErrorMacro` events in [this PR](https://github.com/Slicer/Slicer/commit/517ded4fd4fb6e69c48b042bc7ade36fadf0f29f).  The underlying problem is that `vtkXMLPolyDataReader` uses `vtkErrorMacro` where it should be using `vtkWarningMacro`.  It was added in [this commit](https://github.com/Kitware/VTK/commit/7f768465185b4dbe04ab23de0a7a6034db4a2e61}.

Another problem that should be addressed is that the contents of the `vtkErrorMacro` string are only reported in the gui and do not get printed to stdout/stderr, meaning that they are not part of any log files generated when using a call like `slicer.util.loadModel` in a script.

## Expected behavior

1. The vtp file should be loaded as an empty file for consistency with earlier versions of Slicer.
2. The strings from any `vtkErrorMacros` should be written to output and not swallowed by the GUI.

## Environment
[...]

PR #6448, opened 2022-06-23
ENH: Log error details if slicer.util file read/write operation fails

lassoan:
see #6447


jcfr (dismissed):
Could you copy in the pull request description and/or commit message, an example of raised exception with the additional details ?

pieper (dismissed):
Thanks 👍

PR #6453, opened 2022-06-28
BUG: Fix loading of markups json from previous versions of Slicer

Sunderlandkyl:
Previous versions of Slicer used the previous "master" branch name to identify the schema.
After rename, old files no longer matched the regex since the branch name was changed, and the urls were no longer lower case.

Fixed by allowing either branch name in the regex and forcing the comparison to be done in lower case.

### DIFF
--- a/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsJsonStorageNode.cxx
+++ b/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsJsonStorageNode.cxx
@@ -53,9 +53,10 @@
 namespace
 {
   const std::string MARKUPS_SCHEMA =
-    "https://raw.githubusercontent.com/Slicer/Slicer/main/Modules/Loadable/Markups/Resources/Schema/markups-schema-v1.0.3.json#";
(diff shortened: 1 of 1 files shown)


Sunderlandkyl:
This should fix the failing tests:
https://slicer.cdash.org/viewTest.php?onlyfailed&buildid=2726949

jamesobutler (approved):
I can confirm failing tests are resolved with these changes.

lassoan:
This is a very serious compatibility issue. Right now markup files written by Slicer-5.1.x are not readable by any previous Slicer versions. I'm wondering if we should revert `master` to `main`. It is nice if the schema id resolves to an URL that is actually accessible, but it is not a requirement. We can choose to keep `master` in the schema ID and avoid compatibility issues with older Slicer versions.

If we keep `main` in the schema ID then we need to release Slicer-5.0.4 quickly, because it is very unexpected that Slicer-5.0 cannot read scenes saved by Slicer-5.1.

Sunderlandkyl:
Ok, I'll submit another PR with the following changes
- change main to master in the written json file
- remove the regex entirely in the current version

Once enough time has passed (Slicer 6.0?) we can change the url back to main.

PR #6456, opened 2022-07-01
Fix public API backward incompatibilities

lassoan:
Recently some words (master, blacklist, etc.) in the API were changed to be more inclusive. However, the old method names must remain functional for 1-2 years to give extension developers time to update.


jamesobutler:
Is the “reference image” part of the term “reference image geometry“ referring to the same thing as “source volume”?

lassoan:
Reference image geometry may be initialized from the source volume, but otherwise they are not related.

jamesobutler:
What thing’s “geometry” is being described? The “reference image” or the “segmentation volume”? What’s the difference in describing something as an “image” versus a “volume”?

lassoan:
> What thing’s “geometry” is being described?

The segmentation's. Whenever a segmentation representation requires origin, spacing, or axis directions then existing segment's geometry, reference image geometry, or source image geometry is used.

> What’s the difference in describing something as an “image” versus a “volume”? 

Image and volume are synonyms. Volume refers to "volumetric image" (3D image). 

"image geometry" means origin, spacing, and axis directions (and maybe also extents).

lassoan:
As there don't seem to be any concerns and the API change broke important extensions, I'll merge this now.

jamesobutler:
Instead of “Reference Image Geometry” would “Segmentation Volume Geometry” or “Segmentation Image Geometry” improve the understanding of what “reference” is referring to?

lassoan:
Segmentation image geometry would refer to the current origin, spacing, axis directions of an image-type representation (such as binary labelmap or fractional labelmap). Reference geometry can be specified separately, even if there is no image-type representation stored in the segmentation.

It may sound unnecessarily complicated and maybe it is. But we did not find any way to make it much simpler.

jamesobutler:
Just wondering why “Reference Geometry” is used but “Reference Volume” was replaced with “Source Geometry” when it appears the “Reference” being referred to is the same thing.

lassoan:
"Reference image" was already used in segmentations for a different concept. Reference indicates that is something related, which is a too weak association (master volume is the main information source for the segmentation, not just related). Reference is already used for way too many things, therefore it is better not to overload the term any further.

We also considered "main" but that would have implied that there are multiple associated volumes and this one is distincted among those. But there is no such list. "Base volume" was another candidate but we felt that source reflected the meaning a bit better.

Issue #6459, opened 2022-07-08
Markup ROI interaction inverted


cpinter:
## Summary

In certain scenarios, the behavior of the ROI interaction control is unexpected, and it depends on the actual camera position. See video:

https://user-images.githubusercontent.com/1325980/177949276-5d0775ab-3b12-4683-8fb5-279ad81a97ad.mp4

## Steps to reproduce

* Load scene (less than 10MB): https://1drv.ms/u/s!Ao_a-dPPX98ZhdwviFcE6ARoz5O5kA?e=DNuxBS
* Move the orange interaction handle to move the ROI in the 3D view
* Rotating the 3D view will change the relationship between the mouse movement and the resulting transform

## Expected behavior

* ROI moves together with the mouse

## Environment
- Slicer version: Slicer-5.0.2 + locally built 5.1-2022-007-01
- Operating system: Windows 10

Sunderlandkyl:
Should be fixed in [661e009](https://github.com/Slicer/Slicer/commit/661e00945f7a1f695e939366bcaf10e9481f2ada).

lassoan:
The issue is fixed (comment/reopen if there are still problems).

cpinter:
Absolutely, thank you! Also for the fix @Sunderlandkyl ! Back in the day didn't notice the update.

PR #6460, opened 2022-07-08
BUG: Fix issue with invalid markup handle interaction

Sunderlandkyl:
In some camera positions the behaviour of the interaction handles is incorrect.
Fixed by ensuring that the reference position is not specified when calculating the last world position using vtkPointPlacer::ComputeWorldPosition.

Re #6459


lassoan (approved):
Looks good to me, thank you.

PR #6467, opened 2022-07-18
ENH: customizable Box Color in 3D view

kerim371:
Box in 3D view is an actor controlled by vtkMRMLViewDisplayableManager.
This manager also controls the renderer background color as well.
Thus it was pretty easy to introduce this feature in the same way.
`vtkMRMLViewNode` now have attribute `double BoxColor[3]`.
User is able to modify box color through vtkMRMLViewNode::SetBoxColor().

Probably it would be good to add the example to [Slicer script repository](https://slicer.readthedocs.io/en/latest/developer_guide/script_repository.html#change-3d-view-background-color):
```python
viewNode = slicer.app.layoutManager().threeDWidget(0).mrmlViewNode()
viewNode.SetBoxColor(1,0,0)
```

### DIFF
--- a/Libs/MRML/Core/vtkMRMLViewNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLViewNode.cxx
@@ -29,6 +29,7 @@ vtkMRMLNodeNewMacro(vtkMRMLViewNode);
 vtkMRMLViewNode::vtkMRMLViewNode()
 {
   this->BoxVisible = 1;
+  this->GetDefaultBoxColor(this->BoxColor);
   this->AxisLabelsVisible = 1;
   this->AxisLabelsCameraDependent = 1;
   this->FiducialsVisible = 1;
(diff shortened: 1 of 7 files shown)


kerim371:
Does ReadTheDocs and Slicer API docs checks didn't succed because of my changes?
Can't find information on that in the logs.

Libs/MRML/DisplayableManager/vtkMRMLViewDisplayableManager.cxx
```diff
@@ -394,6 +410,16 @@ void vtkMRMLViewDisplayableManager::vtkInternal::UpdateAxisVisibility()
   this->External->RequestRender();
 }
 
+//---------------------------------------------------------------------------
+void vtkMRMLViewDisplayableManager::vtkInternal::UpdateAxisColor()
+{
+  double boxColor[3];
```
cpinter, line 416: It's good practice to initialize any array with values. Above as well please.

kerim371: Fixed

Libs/MRML/Core/vtkMRMLViewNode.cxx
```diff
@@ -242,6 +249,13 @@ double* vtkMRMLViewNode::defaultBackgroundColor2()
   return backgroundColor2;
 }
 
+//------------------------------------------------------------------------------
+double* vtkMRMLViewNode::defaultBoxColor()
+{
+  static double boxColor[3] = {1.0, 0.0, 1.0};
```
lassoan, line 255: Having a static member variable in a method is very obscure and could have unintended side effects. Also, naming of the method does not follow VTK conventions. Please change it to:

```
void vtkMRMLViewNode::GetDefaultBoxColor(double color[3])
{
  color[0] = 1.0;
  color[1] = 0.0;
  color[2] = 1.0;
}
```

I know that `defaultBackgroundColor` and `defaultBackgroundColor2` does not follow the naming conventions either but we'll fix that later.

kerim371: Done

jamesobutler:
@lassoan Do you have any additional comments here?

jamesobutler (approved):
Startup:
image
With example customization applied:
image

PR #6468, opened 2022-07-18
ENH: add placeholder to OutlinePolyData otherwise vtkError appears

mauigna06:


lassoan (commented):
I've simplified this a bit (no need for a dummy transform filter input; a simple polydata is sufficient), @mauigna06 please test if this fix is still functional.

mauigna06:
Hi @lassoan. Yes, the fix still works

Thank you

lassoan:
Thank you for your contribution (and for testing the updated fix), this is merged now.

PR #6472, opened 2022-07-19
BUG: Fix sampling distance computation for multi-volume rendering

### LINKED ISSUE #5238: MultiVolume Rendering quality is very low if volume spacing is small
[Give a short summary of what the problem is]
Description from https://discourse.slicer.org/t/multi-volume-rendering-bug/13939/5?u=muratmaga

`The problem is that automatic sampling step computation does not work in the multivolume renderer, so the image spacing has to be approximately 1 to get good-quality rendering. `

## Environment
- Slicer version: Slicer-4.11-2020-09-30
- Operating system: Windows

### LINKED ISSUE #5312: Multi-volume rendering much slower using VTK9
Rendering became much slower in VTK9 using multi-volume. Using 4.11.20200930 the CTChest dataset renders with 35 FPS, and r29477 built with VTK9 only 4 FPS.

## Steps to reproduce

* Use Slicer with VTK9
* Load CTChest sample data
* Turn on Volume rendering with Multi-volume technique
* Rendering is very slow, at 4 FPS (on a 2070 GTX)

## Expected behavior

* Rendering is continuous (CTChest dataset renders with 35 FPS)

## Environment
- Slicer version: r29477 built with VTK9

lassoan:
Multi-volume rendering quality was always quite low because always "adaptive" rendering mode was used (regardless of what was set in the GUI) and the sampling distance was not computed correctly.

fixes #5238
fixes #5312

### DIFF
--- a/Modules/Loadable/VolumeRendering/MRMLDM/vtkMRMLVolumeRenderingDisplayableManager.cxx
+++ b/Modules/Loadable/VolumeRendering/MRMLDM/vtkMRMLVolumeRenderingDisplayableManager.cxx
@@ -55,6 +55,7 @@
 #include <vtkMatrix4x4.h>
 #include <vtkPlane.h>
 #include <vtkPlanes.h>
+#include <vtkPointData.h>
 #include <vtkRenderWindow.h>
 #include <vtkRenderWindowInteractor.h>
 #include <vtkRenderer.h>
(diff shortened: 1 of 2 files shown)


cpinter:
Sorry for not having reacted sooner. This is super great, thank you very much @lassoan !

PR #6477, opened 2022-07-22
ENH: add Bishop and Rotation Minimizing Frames to the endoscopy module

### LINKED ISSUE #6476: Endoscopy jitters a bit
## Summary

While using the endoscopy module for a given path curve when it's playing you can see the camera jitter a bit

## Steps to reproduce

A complex enough path (with enough torsion)

## Expected behavior

Camera doesn't jitter a bit

## Environment
- Slicer version: Slicer-5.1.0-2022-07-20

mauigna06:
Together with:
https://github.com/Slicer/vtkAddon/pull/28

Fixes #6476

### DIFF
--- a/Modules/Scripted/Endoscopy/Endoscopy.py
+++ b/Modules/Scripted/Endoscopy/Endoscopy.py
@@ -43,6 +43,12 @@ NA-MIC, NAC, BIRN, NCIGT, and the Slicer Community.
 # qSlicerPythonModuleExampleWidget
 #
 
+class FrameOrientationTypes:
+    ByNormalPlaneOfCurve = 0
+    BishopFrames = 1
+    RotationMinimizingFrames = 2
+
+
 class EndoscopyWidget(ScriptedLoadableModuleWidget):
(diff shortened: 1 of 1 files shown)


jamesobutler:
I've pushed an update to resolve the lint CI failure.

lassoan:
@mauigna06 Has this ended up significantly helping? As I recall, the change was not ended up as useful as initially seemed.

mauigna06:
Hi Andras.

I may be able give a longer explanation on the weekend but I think the answer is no

mauigna06:
> As I recall, the change was not ended up as useful as initially seemed.

Yes, that's right. Just resampling the spline with enough points got rid of the jittering on the endoscopy simulation.

Also I don't notice if the path is more stable while using bishop frames rather than the current implementation of the endoscopy module. But the current implementation appears to select a better first frame by using the normal of the plane-fit of the spline, making the fly through experience a bit nicer

lassoan:
Thank you for experimenting with this. It has been useful to learn all this even if it did not lead to any source code change in the end.

Issue #4906, opened 2020-04-30
Slicer API documentation is not up-to-date


lassoan:
Slicer API documentation has not been updated since January.

See footer of any documentation page, for example [this](http://apidocs.slicer.org/master/classvtkMRMLSegmentEditorNode.html#a451836bc25febd5d1101f0c09ae6a558): `Generated on Thu Jan 30 2020 21:22:17 for Slicer by   doxygen 1.8.14`

lassoan:
@jcfr or @sjh26 it would be great if you could look into this. Thanks!

sjh26:
The apidocs generation apparently has not been running, and before that was having issues with installing dependencies for the doc builder:

https://circleci.com/gh/Slicer/apidocs.slicer.org

lassoan:
Is there any update on this? Could the documentation generation re-enabled?

jcfr:
This is now fixed for the latest api docs:

![image](https://user-images.githubusercontent.com/219043/92275834-dc459a00-eebd-11ea-9b4c-db30ac299147.png)

See https://apidocs.slicer.org/master/

![image](https://user-images.githubusercontent.com/219043/92275872-ed8ea680-eebd-11ea-93d9-35783c1e3146.png)

Let's keep the issue open as we still need to fix:
- versionning of doxygen documentation when a tag is pushed
- generation of review doxygen (like we do for readthedocs ...)

jamesobutler:
Is API documentation building broken again? It is showing as Feb 19 2021 at http://apidocs.slicer.org/master/
![4785F4B5-A755-4CF3-9A43-0AE599C30546](https://user-images.githubusercontent.com/15837524/126422404-00d59b3b-17c7-4117-9d38-6192033c552b.jpeg)

jcfr:
Thanks for the ping. I will follow up with the comments and/or PR. I have an idea of the problem.

jamesobutler:
I've begun to investigate and have started with issuing https://github.com/Slicer/apidocs.slicer.org/pull/1

jcfr:
Thanks @jamesobutler  I created an issue with some suggestion to move forward: https://github.com/Slicer/apidocs.slicer.org/issues/2

jcfr:
Error related to `slicer-apidocs-builder` is now fixed in https://github.com/Slicer/slicer-apidocs-builder/pull/1

[...]

PR #6481, opened 2022-07-27
ENH: Display git diff when extension update is available

lassoan:
When an extension update is available then a link is added to the extension description to list all the changes between the installed and current version (if the repository is hosted on github and version information is available).

image

When the user clicks the link then the change log is displayed in the default browser:

image

### DIFF
--- a/Base/QTGUI/qSlicerExtensionsLocalWidget.cxx
+++ b/Base/QTGUI/qSlicerExtensionsLocalWidget.cxx
@@ -21,6 +21,7 @@
 // Qt includes
 #include <QAbstractTextDocumentLayout>
 #include <QDebug>
+#include <QDesktopServices>
 #include <QFileInfo>
 #include <QLabel>
 #include <QListWidget>
@@ -61,6 +62,7 @@ public:
     InstalledExtensionSlicerVersionRole,
     InstalledExtensionUpdatedRole,
     OnServerExtensionRevisionRole,
+    ChangeLogUrlRole,
     OnServerExtensionMissingRole,  // confirmed that the extension is missing from the server
     OnServerExtensionUpdatedRole,
     UpdateAvailableRole,
@@ -338,7 +340,7 @@ public:
     QTextOption textOption = this->Text.defaultTextOption();
     textOption.setWrapMode(QTextOption::NoWrap);
     this->Text.setDefaultTextOption(textOption);
-    this->MoreLinkText = tr("More");
+    this->MoreLinkText = tr("More...");
     }
 
   // --------------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


Base/QTGUI/qSlicerExtensionsLocalWidget.cxx
```diff
@@ -701,15 +710,29 @@ QListWidgetItem* qSlicerExtensionsLocalWidgetPrivate::updateExtensionItem(const
     qWarning() << Q_FUNC_INFO << " failed: missing implementation for serverAPI" << serverAPI;
     }
 
+  QString installedRevision = metadata["installed"].toBool() ? metadata["revision"].toString() : QString();
+  QString onServerRevision = metadataFromServer["revision"].toString();
+
+  // Generate change log URL - only if repository is hosted on github
+  QUrl scmUrl = QUrl(metadataFromServer["scmurl"].toString());
+  QString changeLogUrl;
+  if (scmUrl.host().toLower() == "github.com"
```
jamesobutler, line 719: Should be easy to support the couple of extensions on gitlab.com as well

jamesobutler (commented):
Having a quick link to the diff of the commits is a nice added feature 🙏🏻 

On the topic of "Improving user understanding about an available update" I was starting to observe that recently there is a lot more information about an update available to the user which can be overwhelming. There is a lot of info in that blue line.

There is information that a regular user cares about and then there is information that a developer cares about. The info highlighted in red in the image below such as the git hash of the installed version, git hash of the available version and the changelog (reviewing the code diff, since it is not just a release notes) are all information that only a developer will really understand. The alphanumeric git hash is something that is unfamiliar to a non-developer as a version. A non-developer would only understand something like "Version 2.1.0" which is something most extensions do not have so would really just understand the date "2022-07-08" as the version.
image

To address not overshowing info to users that they don't understand, but to still provide info to developers I suggest something maybe like the followingr:
image
- Update button is moved underneath the icon: Currently when there is an update available, as a simple user, I was inclined to press the shiny blue info icon and wasn't immediately drawn to the update button on the far right. It was located between other buttons and was smaller in relative size to the bookmark button. The bookmark button appeared to be more important even when there was an update available. Instead the user would now follow their eye to the shiny extension icons where they would then see the update button.
- Changelog is available, but positioned underneath the update button
- Current installed version number is still displayed in the same position (underneath extension name) with same look and feel as when there is not an available update.
- A tooltip for the Update button would describe the new available version with git hash and date to inspect further. Additionally the download file size would be nice to include in the tooltip as well. This additional info in the update tooltip would reduce the overwhelming amount of additional info to a regular user (non-developer).

Base/QTGUI/qSlicerExtensionsLocalWidget.cxx
lassoan: Good idea. It would be great if you could look it up how it can be done.

jamesobutler:
Without changing the layout of a lot of things at this time, how do we address the issue of too much information about the version update being displayed to the user. The concept of simplifying the display of things for regular users, but providing the more detailed information for more advanced users/developers.

Users may care to know what version is currently installed to provide during debugging. Knowing the version that will be installed however isn't cared by simple users. They will see the new installed version once they install it and can report that if they have issues. Under normal conditions they will just update and not need to check the version of the update. They just care to update when an update is available. Inspecting the new version number or diff log is a more advanced user/developer action.

Base/QTGUI/qSlicerExtensionsLocalWidget.cxx
jamesobutler: The same convention is used for gitlab.

`scmUrl.setPath(QString("%1/compare/%2...%3").arg(scmUrl.path()).arg(installedRevision).arg(onServerRevision));`

For example:
https://gitlab.com/opendose/opendose3d/compare/2d08584afbb71d829d63a00e9f1c7c7898d3b8c5...2a9ebce68627f96da05dda628bada05c3ef827df
which can be used and gets redirected to:
https://gitlab.com/opendose/opendose3d/-/compare/2d08584afbb71d829d63a00e9f1c7c7898d3b8c5...2a9ebce68627f96da05dda628bada05c3ef827df

jcfr:
Rebased for the sake of testing fixes related to apidocs generation - See https://github.com/Slicer/Slicer/issues/4906#issuecomment-1214239522

lassoan:
@jamesobutler I've added gitlab support based on your suggestion and tested everything a bit more. I agree that installed and available versions information could be more subtle, but the version information is already displayed like this in current Slicer (so there is no regression) and implementation of a different layout/behavior would not be trivial (due to the various constraints I described above), so I would not address that in this pull request.

PR #6482, opened 2022-07-27
ENH: Allow selecting volume sequence nodes as CLI module input

lassoan:
If an input `image` parameter's `type` is set to "dynamic-contrast-enhanced", "signal", or "multichannel" then the generated node selector allows selection of `vtkMRMLSequenceNode` node that contains `vtkMRMLScalarVolume` nodes in it.

image

This is useful for allowing running CLI modules such as [PkModeling](https://github.com/QIICR/PkModeling/blob/master/CLI/PkModeling.xml) on volume sequences.


pieper (approved):
This looks good - thanks for jumping on this 👍 

Is this enough to make PKModeling work?

PR #6483, opened 2022-07-28
DOC: Fix typos in various subdirectories

luzpaz:
Found via `codespell -q 3 -L ba,dependees,mater,nd,nin,pinter,pullrequest,ro,serie`


Base/Python/slicer/util.py
```diff
@@ -3590,7 +3590,7 @@ def longPath(path):
     sysInfo = qt.QSysInfo()
     if sysInfo.productType() != 'windows':
         return path
-    # Skip prefixing relative paths as UNC prefix wors only on absolute paths
+    # Skip prefixing relative paths as UNC prefix was only on absolute paths
```
pieper, line 3593: Should be "works"

Extensions/CMake/SlicerExtensionDescriptionSpec.cmake
```diff
@@ -203,7 +203,7 @@ function(slicer_extension_description_spec_defaults_test)
     ENABLED
     STATUS
     )
-  foreach(name IN LISTS required optonal)
+  foreach(name IN LISTS required optional)
```
pieper, line 206: Was this a bug?

pieper: Looks like it was a bug since there's a list named `optional` defined just above.  @jcfr could you look?  Did this have any effect?

pieper (changes requested):
small fixes generally but I'm curious about the one cmake file and hoping Jc or others will comment on it.

Base/Python/slicer/util.py
luzpaz: Fixed in fac873f85ec03aa6fad367907c836ca3c2d74948

Extensions/CMake/SlicerExtensionDescriptionSpec.cmake
luzpaz: Yikes, sorry, didn't mean to change source code.

jcfr: Good catch :ok_hand: 

This is indeed an issue, I will submit an independent pull request including the relevant `Co-authored-by:` trailers.

jcfr: This is being addressed by https://github.com/Slicer/Slicer/pull/6485

jcfr:
Before integrating this, we should understand why the current workflow [codespell.yml](https://github.com/Slicer/Slicer/blob/main/.github/workflows/codespell.yml) is not catching these issues.

jcfr:
I am revisiting the installation of `codespell` in the GitHub Actions workflow so that it is installed from git the least release available on PyPI is from June 2021. See https://pypi.org/project/codespell/#history

Indeed, since then the dictionaries have been updated.

Thanks @luzpaz for improving these and contributing directly to codespell :100: :pray:

jcfr:
It turns out that we are already installing the latest version in the workflow:

https://github.com/Slicer/Slicer/blob/main/.github/workflows/codespell.yml#L12

The "issue" is that the workflow currently only check for files modified or added to pull request.

To effectively run codespell on the entire code base on a regular basis, I will update the workflow so that it runs regularly like we do for `update-slicer-certificate-bundle.yml`:

https://github.com/Slicer/Slicer/blob/38e026f75f78320a2cb579b5fa96f0917ab4f02c/.github/workflows/update-slicer-certificate-bundle.yml

jcfr:
> we are already installing the latest version in the workflow:

It turns out that we run the latest version of `actions-codespell` but not the latest version of the codespell package. See https://github.com/codespell-project/actions-codespell/blob/master/requirements.txt

jcfr:
To address this, I will fork the `actions-codespell` to add support for choosing the version of codespell based on suggestion discussed in https://github.com/codespell-project/actions-codespell/issues/5

luzpaz:
JFYI, I'm also alternating with a different dictionary (that isn't part of codespell) using the `codespell -D` flag

jcfr:
@luzpaz This is ready for final review :rocket:  Thanks again for your patience :pray: 

Note that I also updated the GitHub Action workflow to use https://github.com/Slicer/actions-codespell#readme

jcfr:
@pieper  Note that the code fixes have been integrated independently as part of https://github.com/Slicer/Slicer/pull/6485

pieper (approved):
LGTM 👍

PR #6489, opened 2022-08-04
DOC: Add citation

jamesobutler:
This add visibility to how to cite Slicer from the GitHub repository by the addition of a CITATION.cff file. This would be in addition to what is stated at https://slicer.readthedocs.io/en/latest/user_guide/about.html#how-to-cite.

Resources about CITATION.cff
- https://github.blog/2021-08-19-enhanced-support-citations-github/
- https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-citation-files
- https://citation-file-format.github.io/
- https://github.com/citation-file-format/citation-file-format/blob/1.2.0/schema-guide.md#definitionsreferenceend

I've added the file to my default branch at https://github.com/jamesobutler/Slicer so you can see how it looks.

image

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,0 +1,51 @@
+cff-version: 1.2.0
+message: "Please cite this software using the metadata from 'preferred-citation'."
+title: "3D Slicer"
+repository-code: https://github.com/Slicer/Slicer
+website: https://www.slicer.org/
+preferred-citation:
+  type: article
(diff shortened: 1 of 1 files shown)


jcfr (approved):
Thanks for suggesting this change :100: 

As you indicated, this is consistent with https://slicer.readthedocs.io/en/latest/user_guide/about.html#d-slicer-as-a-platform

PR #6494, opened 2022-08-08
BUG: Make sure ProcessMRMLEvents is called all the way to the base no…

cpinter:
…de class

The transformable node did not call the ProcessMRMLEvents function from its superclass, so the storable node and base MRML node classes did not have a chance to handle some events.


pieper (approved):
This makes sense to me.  If nobody else has comments I'm okay to merge.

cpinter:
Thank you! Merged.

Issue #6439, opened 2022-06-20
Change color of crosshair


RafaelPalomar:
## Is your feature request related to a problem? Please describe.

Some users have pointed out that the color of the crosshair is hard-coded and would like to be able to change it from Python.

## Describe the solution you'd like

Add a color variable to `vtkMRMLCrosshairNode` and use it in `vtkMRMLCrosshairDisplayableManager` and `vtkMRMLCrosshairDisplayableManager3D`.

## Describe alternatives you've considered

N/A

## Additional context

N/A

RafaelPalomar:
For reference, discussion here: [https://discourse.slicer.org/t/change-crosshair-color/23966/7](https://discourse.slicer.org/t/change-crosshair-color/23966/7).

RafaelPalomar:
I'm working on it.

PR #6496, opened 2022-08-09
ENH: Add functionality to change crosshair color

RafaelPalomar:
This closes Slicer/Slicer#6439

### DIFF
--- a/Docs/developer_guide/script_repository/gui.md
+++ b/Docs/developer_guide/script_repository/gui.md
@@ -402,6 +402,14 @@ Crosshair node stores two positions: Cursor position is the current position of 
 
 :::
 
+### Change the crosshair color
+
+```python
+# Get the crosshair node
+crosshairNode = slicer.util.getNode("Crosshair")
+# Set the crosshair color to Red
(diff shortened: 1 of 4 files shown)


jamesobutler (commented):
> Some users have pointed out that the color of the crosshair is hard-coded and would like to be able to change it from Python.

@RafaelPalomar Could you also update the python script repository with an example of how to do this? That way when users ask we can point them to it.

Do you think it is more widely needed where there is a GUI option, like in settings, that will allow to change it?

RafaelPalomar:
@jamesobutler, thanks for having a look at this. I updated the documentation with a python snippet to change the crosshair color.

I would wait with adding any GUI option. Changing the color of the crosshair is probably not a very used feature, and doing it in python is simple enough.

I see the readthedocs tests are failing, but it looks more like an infrastructure problem.

jamesobutler:
@RafaelPalomar Yeah the readthedocs are currently failing because your branch is 10 commits behind `main` which actually includes some fixes. If you rebase, then it should allow for the check to pass.

RafaelPalomar:
> If you rebase, then it should allow for the check to pass.

DONE

jamesobutler (approved):

PR #6497, opened 2022-08-11
Left Mouse 'click-and-drag' to shift between 2d pics test ok

simzhangbest:
left mouse click-and-drag to shift between 2d pics finished, but not too smooth

### DIFF
--- a/Libs/MRML/DisplayableManager/vtkMRMLSliceIntersectionWidget.cxx
+++ b/Libs/MRML/DisplayableManager/vtkMRMLSliceIntersectionWidget.cxx
@@ -154,6 +154,8 @@ void vtkMRMLSliceIntersectionWidget::UpdateInteractionEventMapping()
     }
   if (this->GetActionEnabled(ActionBrowseSlice))
     {
+    this->SetEventTranslationClickAndDrag(WidgetStateIdle, vtkCommand::LeftButtonPressEvent,
+          vtkEvent::NoModifier, WidgetStateIncDecSlice, WidgetStateIncDecSliceStart, WidgetStateIncDecSliceEnd);
     this->SetKeyboardEventTranslation(WidgetStateIdle, vtkEvent::NoModifier, 0, 0, "Right", WidgetEventIncrementSlice);
     this->SetKeyboardEventTranslation(WidgetStateIdle, vtkEvent::NoModifier, 0, 0, "Left", WidgetEventDecrementSlice);
     this->SetKeyboardEventTranslation(WidgetStateIdle, vtkEvent::NoModifier, 0, 0, "Up", WidgetEventIncrementSlice);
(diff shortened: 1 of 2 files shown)


Libs/MRML/DisplayableManager/vtkMRMLSliceIntersectionWidget.cxx
```diff
@@ -1392,6 +1411,28 @@ bool vtkMRMLSliceIntersectionWidget::ProcessZoomSlice(vtkMRMLInteractionEventDat
   return true;
 }
 
+
+// TODO: by simzhangbest mail:dg20330034@smail.nju.edu.cn 
+// mouse left-click-drag shit between pics not smooth enough, may someone modify this
+bool vtkMRMLSliceIntersectionWidget::ProcessIncDecSlice(vtkMRMLInteractionEventData* eventData)
+{
+    // Please adjust this parameter to tune the Inc&Dec speed.
+    const double CONF = 0.01;
```
pieper, line 1420: Rather than hard-code this, it would be better to make it a function of the window size using the same code as `ProcessZoomSlice`.

Libs/MRML/DisplayableManager/vtkMRMLSliceIntersectionWidget.h
```diff
@@ -209,6 +212,8 @@ class VTK_MRML_DISPLAYABLEMANAGER_EXPORT vtkMRMLSliceIntersectionWidget : public
 
   bool ProcessZoomSlice(vtkMRMLInteractionEventData* eventData);
 
+  bool ProcessIncDecSlice(vtkMRMLInteractionEventData* eventData);
```
pieper, line 215: `IncDec` wasn't clear to me on first reading.  Maybe better to call this `Scroll` mode (this would need to be changed everywhere).

Libs/MRML/DisplayableManager/vtkMRMLSliceIntersectionWidget.cxx
```diff
@@ -1392,6 +1411,28 @@ bool vtkMRMLSliceIntersectionWidget::ProcessZoomSlice(vtkMRMLInteractionEventDat
...
+    const int* eventPosition = eventData->GetDisplayPosition();
+    const int* windowSize = this->GetRenderer()->GetRenderWindow()->GetSize();
+
+    // modify by simzhang 0809
+    int deltaY = eventPosition[1] - this->PreviousEventPosition[1];
+
+    this->MoveSlice(deltaY * CONF);
+
+    // if (deltaY > 0) this->IncrementSlice();
+    // else this->DecrementSlice();
```
pieper, line 1431: We don't like to leave dead code in the source so this should be removed once the functionality is in place.

pieper (changes requested):
I'm thinking this makes sense as a left button mouse action, but it's a question whether it should be on by default or maybe be a user preference setting.  Changing window/level used to be the default but we disabled it because it was too easy to accidentally change it when trying to click on a Markup.  I wonder if this would have the same issue.  @lassoan @jcfr @jamesobutler

jamesobutler:
Yes the default mouse mode should be just the regular arrow cursor mode for clicking/translating things. A new mouse mode for manipulating slice offset with left click could be an option though.

pieper:
@simzhangbest are you still able to work on this?  I like the idea of having a stack-scroll mode but we'd need to address the comments above.

lassoan:
This will be a useful feature, but needs more work. I'm closing this PR to reduce the number of open PRs, but we are happy to reopen if you can work on it a bit more.

PR #6498, opened 2022-08-11
STYLE: Add recent python formatting to ignored list for git blame

jamesobutler:
This adds a few of the recent python formatting commits to the git blame ignore list.


jcfr:
Thanks :pray:

PR #6499, opened 2022-08-11
STYLE: Add recent python formatting to ignored list for git blame (5.0)

jamesobutler:
Similar to #6498, but with the appropriate git hashes for the commits on the `5.0` branch.


jcfr:
Thanks @jamesobutler :pray:

PR #6506, opened 2022-08-17
ENH: Segment table opacity changes all segment opacity

jamesobutler:
See https://discourse.slicer.org/t/problem-changing-the-individual-segment-opacity-in-slice-views/24810

Previously the opacity widget in the segments table would only change the 3D segment opacity.

### DIFF
--- a/Modules/Loadable/Segmentations/Widgets/qMRMLSegmentsModel.cxx
+++ b/Modules/Loadable/Segmentations/Widgets/qMRMLSegmentsModel.cxx
@@ -138,7 +138,7 @@ void qMRMLSegmentsModelPrivate::init()
   q->horizontalHeaderItem(q->nameColumn())->setToolTip(qMRMLSegmentsModel::tr("Segment name"));
   q->horizontalHeaderItem(q->visibilityColumn())->setToolTip(qMRMLSegmentsModel::tr("Segment visibility"));
   q->horizontalHeaderItem(q->colorColumn())->setToolTip(qMRMLSegmentsModel::tr("Segment color"));
-  q->horizontalHeaderItem(q->opacityColumn())->setToolTip(qMRMLSegmentsModel::tr("Segment opacity (3D views)"));
+  q->horizontalHeaderItem(q->opacityColumn())->setToolTip(qMRMLSegmentsModel::tr("Segment opacity"));
   q->horizontalHeaderItem(q->statusColumn())->setToolTip(qMRMLSegmentsModel::tr("Segment status"));
 
   q->horizontalHeaderItem(q->visibilityColumn())->setIcon(QIcon(":/Icons/Small/SlicerVisibleInvisible.png"));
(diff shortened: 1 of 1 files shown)


lassoan:
The current behavior (that's why we chose the slider to only adjust the 3D opacity) is intentional. We could change the GUI to make it easier to adjust opacities in slice views, but the proposed change has a couple of issues:
- Decreasing the 2D and 3D opacity together very rarely if ever makes sense. Usually you want to make the 3D representation transparent a little bit so that you can see segments inside a large segment (such as the skin). If we did this by changing the overall opacity then the segment becomes barely visible in slice views. 
- The opacity slider in the slice view controller (in the segmentation layer) and the opacity slider in the subject hierarchy (in the right-click menu of the eye icon) both adjust the overall (2D+3D) opacity and we could say that this commit would make the behavior consistent with that. However, those sliders refer to the opacity of the segmentation (not a segment). Also, there is a correspondence between that slider and a MRML node property, while in the proposed commit one slider is mapped to three different opacity values. By using these sliders you can also see how bad it is that segmentation fades out in slice view when you want to just make it slightly transparent in 3D.
- It is undesirable that the fill opacity is set to the same value as the outline opacity (as the outline is usually requires higher opacity value for optimal visibility).

Could you think of a different way of making 2D fill opacity easily adjustable (independently from 3D opacity and 2D outline opacity)?

jamesobutler:
> Decreasing the 2D and 3D opacity together very rarely if ever makes sense. Usually you want to make the 3D representation transparent a little bit so that you can see segments inside a large segment (such as the skin). If we did this by changing the overall opacity then the segment becomes barely visible in slice views.

Interesting. I have typically had the opacity set the same for 2D and 3D. It is confusing when viewing an object if it is represented differently in 2D vs 3D. It makes sense to utilize the same color for the 2D object as the 3D object so it is easily recognizable. Same goes for opacity. When the opacity for an object in 2D doesn’t match what it is in 3D it appears to be a different object.

What value do you typically use for 3D opacity? Our group has always used 40% for 2D fill and 3D. This has usually been good for still viewing the image data in the 2D slice and also the image data (slice views or volume rendering) shown in 3D along side the segment.

Also Slicer has a default of the segmentation 2D fill at 50% and 3D at 100%. A segment opacity of 50% would then be 25% 2D fill and 50% 3D, but we don’t use it in this configuration as yes it would get too light in the slice views. We would set the segmentation 2D fill and 3D opacity to 100% so that a default segment opacity of say 50% is 2D fill at 50% and 3D at 50%.

jamesobutler:
> This is not usually the case. Opacity in 2D is a must (you very rarely want to completely occlude the underlying slice image), while in 3D it is only used in special cases (because transparency in 3D view causes lot of complexity, loss of 3D perception, may cause rendering artifacts, etc.).

If Opacity control of the 2D Fill is a must while Opacity of 3D is only used in special cases, then the opacity in the Segments table should be for just 2D Fill then instead of 3D? While per segment 3D opacity should go elsewhere in the UI for a less used case?

> It is impossible to decipher the shape (which bumps are on the front surface and which are on the back?):

I guess the shape is not so much a concern for us as seeing the segment with the image data also present in 3D. For us the image data is almost always shown in the 3D View, and not just the segment on its own.
image

lassoan:
It is useful to allow users to quickly reduce the opacity of a segment to see what is inside in it in the 3D view, so I would not take that away. At the same time, there is certainly value in allowing changing 2D fill opacity, too, and so it would be nice to make that available, too. I'm not sure what would be a good way, maybe adding a second opacity column in the segment table? We could also add a third opacity column, to increase/decrease relative opacity overall. To keep GUI complexity under control, maybe we could allow users to choose which column(s) to display in the segment table. Visibility of columns of the segment table are already configurable, there is just no GUI to customize it.

In summary, this needs more work. I close the pull request to reduce the number of open pull requests, but feel free to reopen if you implemented some improvements.

PR #6507, opened 2022-08-20
ENH: Expose segment conversion paths in Python

lassoan:
Previously, complex STL containers were used that were not Python-wrappable. Now conversion paths and parameters are all VTK objects.
This allows converting segmentations using a custom path, when there are multiple conversion paths are available.
For example, multiple conversion paths are available after installing SlicerRT extension, which provides additional segment representations and conversion rules.


pieper (approved):
Nice 👍 To my eye this code is much easier to read.

PR #6508, opened 2022-08-22
DOC: add batch example scripts

pieper:


lassoan (commented):
Thank you, this looks good. Since we switched to 4 spaces for indenting Python code in .py files, we may consider using them in script repository examples, too. I've added two other trivial comments inline.

pieper:
Agreed, we should switch to 4 since it's easier to read and looks more "normal" to most people.  For now though we have almost exclusively 2-space so this will be consistent.  Probably there's a tool that can batch-prettify all the code snippets in one go.

PR #6517, opened 2022-08-29
ENH: Update action icons in Menu Bar

jamesobutler:
Icons originally from https://fonts.google.com/icons.

This work is being pulled out of #5966 where it was left off at https://github.com/Slicer/Slicer/pull/5966#issuecomment-1048141868. This utilizes SVG icons from the google icon set with dynamic adjusting of the fill color value when the Slicer style changes from Light to Dark to maintain high contrast. It currently goes with a consistent white or black fill color rather than some though non-grayscale versions could be implemented as well.

| Slicer 5.0.3 | This PR (light)| This PR (dark)|
|-------------|----------|----------------------|
|image|image|image|
|image|image|image|
|image|image|image|
|image|image|image|


Applications/SlicerApp/qSlicerAppMainWindow.cxx
```diff
@@ -154,39 +156,80 @@ void qSlicerAppMainWindowPrivate::setupUi(QMainWindow * mainWindow)
...
+    QString newHexString = "fill=\"" + hexColor;
+    QByteArray newByteArray = newHexString.toLocal8Bit();
+    const char *newHexChar = newByteArray.data();
+    data.replace("fill=\"#000000", newHexChar);
+    QPixmap pixmap;
+    pixmap.loadFromData(data);
+    icon = QIcon(pixmap);
+  }
+  return icon;
 }
```
pieper, line 233: It seems like this basic code structure gets repeated a lot - is there a way to factor it out?

pieper (changes requested):
Generally this looks good to me.  I didn't try building but it looks safe so other than the redundancy question I'd be okay to merge.

lassoan:
Thank you @jamesobutler. I like these simple icons, they make the application look a bit more modern and if we replace all the other icons with similar style then it will make the style more consistent.

I have just two comments:
- The icons seem blurry in the screenshots. Is this an artifact of the screenshot capture tool or the rendering is actually not pixel perfect? It needs to be tested on both normal and hi-dpi screen.
- Strictly black-and-white icons are a bit too dull for me and we miss out on conveying information in color. We could either use 2-3 colors (as in most MS Office application toolbar icons (can be black/white and 1-2 of the Slicer logo colors) or maybe we could use color for grouping toolbars.

Applications/SlicerApp/qSlicerAppMainWindow.cxx
jcfr: In this PR (or a follow-up one), adding a utility function called `getColorizedIcon(const QString& hexColor, const QString& fileName)` to https://github.com/commontk/CTK/blob/master/Libs/Widgets/ctkWidgetsUtils.h may be sensible.

lassoan:
@jamesobutler I see that you have forced-push some commits yesterday. Is this ready for testing? CI shows some build error.

jamesobutler:
I haven't addressed any coloring beyond black/white which is why I still have it as a draft PR.

lassoan:
Coloring is not a showstopper. Do you see any blurring of the icons or they appear sharp on your computers? Does the build error  on CircleCI look like a real one?

jamesobutler:
> Coloring is not a showstopper. 

I was reading more about coloring of icons in material design and it utilizes alpha coloring for grayscale coloring, but I haven't implemented it yet. Right now I was utilizing the Text color for the standardPalette of the currently selected style.
https://developers.google.com/fonts/docs/material_icons#styling_icons_in_material_design

`ENH: Update Welcome module icons` and `ENH: Update module selector toolbar icons` are new commits that I pushed the yesterday when I was playing around with additional new icons outside of just what was in the Menu bar. I don't quite have the light/dark mode working for the Chat.svg in the Welcome module. I'm also not sure I like my usage of the chevron instead of the regular arrow for the module selector forward/backward. I can pick these off of this PR if you're wanting the changes to the Menu bar specifically sooner rather than later. I'd be happy to discuss at tomorrow's hangout if you'd like.

image

> Do you see any blurring of the icons or they appear sharp on your computers?

@lassoan I don't have any high DPI screens (>1080p) to test with, but their source is google icons and they are SVGs so their ability to scale should be good. I know when the source is a PNG that is based on some image and not pixel perfect then having the 2x scale version of the icon is also needed, but I don't see the need for that with these icons.

> Does the build error on CircleCI look like a real one?

It looks real. I've pushed again and tested locally with success.

Applications/SlicerApp/qSlicerAppMainWindow.cxx
jamesobutler: Corresponding PR https://github.com/commontk/CTK/pull/1040

cpinter:
Again, why? And why delete the branch? I think this one would have been useful, but it is destroyed with no trackable discussion (same as https://github.com/commontk/CTK/pull/928 and https://github.com/Slicer/Slicer/pull/5176)

jamesobutler:
This issue was closed after almost 2 years of no integration. This indicates there were serious issues with the proposed work or a need to reassess the validity of the change. Based on the staleness of the PR any work should be started with a new conversation against latest code. The code exists here for anyone to pick back up.

Icon re-work has been taken over by Wendy. See https://discourse.slicer.org/t/new-icons-for-slicer-styling-to-include-monochromatic-designs-and-optional-accent-colors/35409/6

cpinter:
If this was just about icon rework then alright, I remembered also some reorganization.

PR #6524, opened 2022-09-10
Remove annotations module

lassoan:
The first commit automatically converts annotations nodes (fiducial, ruler, ROI) to markups nodes on scene load.

The second commit removes Annotations module. Annotations module is not removed completely yet, so that it can be used for scene loading.


jamesobutler (commented):
Is it going to be ok that the removal step is happening prior to the May 2023 date stated on the roadmap? https://github.com/Slicer/Slicer/wiki/Roadmap

> Retire Annotations module (replaced by Markups, Screen Capture, Scene views)
>    - [X] Deprecate Annotations module - April 2021
>  :runner: Update all Slicer core modules to work with markups module as well
>    - [ ] Update all Slicer core modules to create markups node by default - May 2022
>    - [ ] Hide Annotations module GUI
>    - [ ] Remove Annotations module, convert annotations node to markups node when loading data from file - May 2023

lassoan:
Thank you @jamesobutler, good catch. I stated with removing Annotations module GUI but then realized that having the annotation nodes in the scene without a GUI to use them is not very useful. That's why coupling GUI removal with automatic conversion to markup nodes makes more sense. I'll start a discussion on the Slicer forum to ensure that developers are OK with the slight change of plan.

jamesobutler:
As part of this process we should probably check the various extensions to see which should be updated to the Markups module similar to what was identified when we dropped the Editor module and some needed to update to Segment Editor.

lassoan:
> As part of this process we should probably check the various extensions to see which should be updated to the Markups module similar to what was identified when we dropped the Editor module and some needed to update to Segment Editor.

Agreed. I had a look at all the extensions that use annotations features and created a summary table of their status (which annotations they use and if they already support markups):

https://1drv.ms/x/s!Arm_AFxB9yqHydRUBei9oHHGcSpSVA?e=9U03cR

I'll submit issues or pull requests to all the repositories that do not support markups yet.

> Is it going to be ok that the removal step is happening prior to the May 2023 date stated on the roadmap? https://github.com/Slicer/Slicer/wiki/Roadmap

We will not include this PR in Slicer-5.2 (expected within a few weeks) but in Slicer-5.4 (expected about 3 months after Slicer-5.2, probably around early next year).

jamesobutler:
@lassoan Some user_guide conflicts are present in this PR. Since Slicer 5.2 has been released, it makes sense to go ahead and remove the annotations module for the Slicer 5.3 development period now. Then it will provide enough time before the Slicer 5.4 release to help with any issues with extensions that still rely on it.

Docs/user_guide/modules/retiredmodules.md
```diff
@@ -5,6 +5,9 @@ The modules listed below have been retired and are no longer present in 3D Slice
 Retiring modules reduces burden on Slicer developers. Most common reason for retiring a module is that a new, improved module is introduced.
 A module may also be retired if it no longer works properly (for example due to changes in Slicer core or third-party libraries) and it is not worth investing time into fixing the issue, because the module is not widely used or alternatives exist.
 
+- **Annotations**
+  - Retired: 2022-09-TBD, [commit TBD](https://github.com/Slicer/Slicer/commit/TBD)
```
jamesobutler, line 9: Once merged, this is a reminder to update the retired date and commit hash.

Docs/user_guide/modules/screencapture.md
```diff
@@ -115,7 +115,7 @@ On Windows, Screen Capture model can automatically download and install ffmpeg w
 
 - [Animator](https://github.com/SlicerMorph/SlicerMorph/tree/master/Docs/Animator#readme) module in [SlicerMorph extension](https://slicermorph.github.io/) allows creating more complex animations, such as cutting through a volume (by changing region of interest), adjusting volume rendering transfer functions, or exploding view of complex model assembly.
 - [Scene Views](sceneviews.md) module can create snapshot of the entire scene content along with a screenshot, which are all saved in the scene.
-- [Annotations](annotations.md) module can create simple screenshots that are saved in the scene. Annotations module is being phased out and Screen Capture module's "Output volume node" feature can be used for saving screenshots in the scene.
+- Capture Toolbar allows creation simple screenshots that are saved in the scene. The feature may be removed in the future. Screen Capture module's "Output volume node" feature can be used for saving screenshots in the scene.
```
jamesobutler, line 118:
```suggestion
- The Capture Toolbar creates simple screenshots that are saved in the scene. The feature may be removed in the future. The Screen Capture module includes an "Output volume node" feature that can be used for saving screenshots in the scene.
```

jamesobutler (approved):
In general this looks good. I only had that one suggested comment change in the Screen Capture documentation file. 

I built locally, packaged and ran tests with success. I saved a scene in Slicer 5.2.1 that has Markups module objects and Annotation module objects. I then reopened it in Slicer with this PR to confirm the upgrade of components on load.

| Slicer 5.2.1 | This PR |
|-------------|----------|
|image|image|

Docs/user_guide/modules/retiredmodules.md
```diff
@@ -5,6 +5,9 @@ The modules listed below have been retired and are no longer present in 3D Slice
 Retiring modules reduces burden on Slicer developers. Most common reason for retiring a module is that a new, improved module is introduced.
 A module may also be retired if it no longer works properly (for example due to changes in Slicer core or third-party libraries) and it is not worth investing time into fixing the issue, because the module is not widely used or alternatives exist.
 
+- **Annotations**
+  - Retired: 2023-01-02, [commit 3c3fb74](https://github.com/Slicer/Slicer/commit/3c3fb74ac329ae2916fa3549699fbc49096ca179)
```
jamesobutler, line 9: I’m pretty sure this commit hash is going to change once you rebase and merge these commits into `main`. I think you’ll need a follow up commit once this PR is merged to get the correct hash on the `main` branch.

lassoan: Good point, thanks. I'll update the commit hash after the merge.

lassoan: Updated in https://github.com/Slicer/Slicer/commit/017b64f1f562a2e40b47b903df6c0daba6ad08e0

jcfr:
This is outstanding :100: Thanks @lassoan and @jamesobutler  Exciting to have the transition from annotations to markups finalized !

PR #6526, opened 2022-09-12
ENH: Update scissors "Apply to visible segments" label

Sunderlandkyl:
The "Apply to all segments" label was misleading, since enabling it caused the scissors effect to only apply on all visible segments.
Changed label from "Apply to all segments" to "Apply to visible segments".

See: https://discourse.slicer.org/t/erase-using-a-segment-as-only-editable-area/20519/13

### DIFF
--- a/Docs/user_guide/modules/segmenteditor.md
+++ b/Docs/user_guide/modules/segmenteditor.md
@@ -157,19 +157,19 @@ Notes:
 
 Grows or shrinks the selected segment by the specified margin.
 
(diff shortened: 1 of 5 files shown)


lassoan (commented):
Thank you for working on this. The change makes sense. As far as I remember, we added such "Apply to all segments" checkbox to other effects in Slicer core and maybe SegmentEditorExtraEffects extension. Could you please check this and update the other effects similarly, too?

Sunderlandkyl:
Updated Hollow, Margin, Smoothing, and segment editor documentation.

lassoan (approved):
Thank you, it looks good!

PR #6528, opened 2022-09-13
Fix use of `run_ctest_` variables

RafaelPalomar:
This commit includes a new check based on the `run_ctest_submit` variable when submitting `PARTS build`. Without this fix, you could try to disable submission to dashboard but the build information will still appear submitted.

This commit modifies the logic of the script to decouple the building of the Slicer package and its upload. This is to support cases where the Slicer package needs to be built, but won't be uploaded. Not having this in place forces the generated package to be uploaded.

### DIFF
--- a/CMake/SlicerDashboardDriverScript.cmake
+++ b/CMake/SlicerDashboardDriverScript.cmake
@@ -546,7 +546,7 @@ ${ADDITIONAL_CMAKECACHE_OPTION}
     #-----------------------------------------------------------------------------
     # Package and upload
     #-----------------------------------------------------------------------------
-    if(WITH_PACKAGES AND (run_ctest_with_packages OR run_ctest_with_upload))
+    if(WITH_PACKAGES AND run_ctest_with_packages)
       message(STATUS "----------- [ WITH_PACKAGES and UPLOAD ] -----------")
 
       if(build_errors GREATER "0")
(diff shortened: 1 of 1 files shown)


CMake/SlicerDashboardDriverScript.cmake
```diff
@@ -566,26 +566,29 @@ ${ADDITIONAL_CMAKECACHE_OPTION}
...
-            slicer_ctest_upload_url(
-              ALGO "SHA512"
-              DOWNLOAD_URL_TEMPLATE "${SLICER_PACKAGE_MANAGER_URL}/api/v1/file/hashsum/%(algo)/%(hash)/download"
-              FILEPATH ${p}
-              )
-            if(run_ctest_submit)
-              ctest_submit(PARTS Upload)
-            endif()
-          endforeach()
+          if(run_ctest_with_upload)
```
jcfr, line 578: This extra check is not required as this code block is already conditionally executed based on:

```
        if(run_ctest_with_upload)
          message(STATUS "Uploading Slicer package URL ...")
```

jcfr:
Closing as trying to force push reports the following:

```
ERROR: Permission to ALive-research/Slicer.git denied to jcfr.
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
```

This pull request as then been superseded by https://github.com/Slicer/Slicer/pull/6585

PR #6533, opened 2022-09-15
BUG: Prevent creation of duplicate markup representations

cpinter:
In some cases, when the creation of the markups widget triggers modified event on the markups node, in a recursive call of vtkMRMLMarkupsDisplayableManagerHelper::AddDisplayNode, a second representation is created, which then will be orphaned, because the map MarkupsDisplayNodesToWidgets will point only to the first one. This causes problems in the form of duplicate, stuck actors. Solved by not allowing modified events to be handled until the creation of the widget is finished.


lassoan (approved):
It looks good, thank you!

PR #6536, opened 2022-09-18
BUG: Make main window destruction more robust

### LINKED ISSUE #6522: destructor of qSlicerMainWindow trying to access invalid memory
## Source code
```.cpp
qSlicerMainWindow::~qSlicerMainWindow()
{
  Q_D(qSlicerMainWindow);
  // When quitting the application, the modules are unloaded (~qSlicerCoreApplication)
  // in particular the Colors module which deletes vtkMRMLColorLogic and removes
  // all the color nodes from the scene. If a volume was loaded in the views,
  // it would then try to render it with no color node and generate warnings.
  // There is no need to render anything so remove the volumes from the views.
  // It is maybe not the best place to do that but I couldn't think of anywhere
  // else.
  vtkCollection* sliceLogics = d->LayoutManager ? d->LayoutManager->mrmlSliceLogics() : nullptr;
  for (int i = 0; i < sliceLogics->GetNumberOfItems(); ++i)
    {
    vtkMRMLSliceLogic* sliceLogic = vtkMRMLSliceLogic::SafeDownCast(sliceLogics->GetItemAsObject(i));
    if (!sliceLogic)
      {
      continue;
      }
    sliceLogic->GetSliceCompositeNode()->SetReferenceBackgroundVolumeID(nullptr);
    sliceLogic->GetSliceCompositeNode()->SetReferenceForegroundVolumeID(nullptr);
    sliceLogic->GetSliceCompositeNode()->SetReferenceLabelVolumeID(nullptr);
    }
}
```

## Problems
In the source code above, there are two problems
1. if `d->LayoutManager ? ` returns `false`, then `sliceLogics` is `nullptr`, and then `sliceLogics->GetNumberOfItems()` would trigger segment fault.
2. Since `d->LayoutManager` is child object of `CentralWidget`, and if I setup my own UI(calls the `setCentralWidget()` again), current `CentralWidget` would be destroyed by Qt then LayoutManager is released too, but `d->LayoutManager ? ` would still return true. Now `d->LayoutManager->mrmlSliceLogics()` is trying to access invalid memory that has already been released.

## Expected behavior

Exit normally.

## Environment
- Slicer version: Slicer-v5.0.3
- Operating system: ubuntu + 18.04

lassoan:
qSlicerMainWindow destructor crashed when a custom widget was used as CentralWidget (not the layout manager).

fixes #6522

### DIFF
--- a/Base/QTApp/qSlicerMainWindow.cxx
+++ b/Base/QTApp/qSlicerMainWindow.cxx
@@ -817,17 +817,21 @@ qSlicerMainWindow::~qSlicerMainWindow()
   // There is no need to render anything so remove the volumes from the views.
   // It is maybe not the best place to do that but I couldn't think of anywhere
   // else.
-  vtkCollection* sliceLogics = d->LayoutManager ? d->LayoutManager->mrmlSliceLogics() : nullptr;
-  for (int i = 0; i < sliceLogics->GetNumberOfItems(); ++i)
+  qSlicerLayoutManager* layoutManager = qSlicerApplication::application()->layoutManager();
+  vtkCollection* sliceLogics = layoutManager ? layoutManager->mrmlSliceLogics() : nullptr;
+  if (sliceLogics)
     {
-    vtkMRMLSliceLogic* sliceLogic = vtkMRMLSliceLogic::SafeDownCast(sliceLogics->GetItemAsObject(i));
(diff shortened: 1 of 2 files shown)


pieper (approved):
Makes sense, thanks 👍

PR #6537, opened 2022-09-18
Add check for available application updates

### LINKED ISSUE #5502: Add automatic application update notification

lassoan:
Now that a new Slicer Stable Releases is planned to be created in every quarter, but extensions are only updated for the latest stable, it is important to keep moving users to the latest Slicer version.

This commit adds an application update check to make users aware that a more recent Slicer Stable Release is available that the currently running Slicer.

Automatic update check is enabled by default. It can be enabled/disabled in the Welcome module and in Application Settings.
Update check can be initiated manually in the Welcome module.

Application updates can be completely disabled using Slicer_BUILD_APPLICATIONUPDATE_SUPPORT CMake variable, or can be temporarily disabled in Slicer.ini (ApplicationUpdates/Enabled=false).

fixes #5502

image

image


jamesobutler (commented):
For an application update notification I think in other programs I’m alerted more upfront with a pop-up dialog on application start? I think like TortoiseGit will do that and some others. 

It makes sense to not do that for extension updates but I think more in your face for an application update would be desired. This would be in addition to the button in the Welcome module. The Welcome module button on its own it’s not large and in your face and could be easy to miss and we don’t want users to miss updating to the latest application version.

jamesobutler:
I wonder if some simple widgets for application update notification or extension update notification would make sense in the QStatusBar area as some permanent widgets so it is present regardless of current selected module. The status bar area seems to be frequently used by other applications for notifications.

lassoan:
@jamesobutler thank you for your insights. You raise excellent points and I would like to continue the discussion of future directions for Slicer's update infrastructure (silent in-place auto-update mechanism similarly to most modern software vs. informative popups for motivating users to manually update), but probably not here but on the Slicer forum.

In this pull request I would like to get comments on how the current feature is implemented and how it is presented to the user. Focusing on changes that we can address before integration into Slicer-5.2 within a few weeks.

lassoan:
> QStatusBar area as some permanent widgets so it is present regardless of current selected module. The status bar area seems to be frequently used by other applications for notifications.

Popups and other more "in-the-face" notifications would imply that users need more convincing. We don't know yet if this is needed or not. Even if current notification is not sufficient we have the option of A. make the notifications more prominent; or B. making the updates easier and/or more automatic. I would get back to this question when we gathered some more experience.

> A statusbar area or “Help->Check For Updates” menu action is a place I would look first when I’m looking to manually check for application updates when I’m not using automatic checks.

I agree, this would be the standard place. If it just switches to the Welcome module and opens the Updates section then it might fit into Slicer-5.2. If we need a separate reusable update widget then it'll have to come later.

> it would be nice to put Release Notes somewhere to view (these maybe on download.slicer.org as well?)

Fully agree, it would be nice to add release notes. The current [download.slicer.org API](https://download.slicer.org/find?os=win&stability=release) does not provide release note information, so this will have to come later.

Note that currently the main motivation for application update is not the release content but the fact that extensions are only kept up-to-date for the latest stable release (not for any of the previous stable releases). We might reconsider this and keep supporting two stable releases, as it is done in most other serious software. That would allow users to delay their upgrade for one release cycle, while now we expect users to update immediately when we create a new stable release (because we immediately stop supporting the "old" release). The supported second stable releases can be the one right before the latest stable, or a designated long-term-support (LTS) release.

There, I could not resist commenting on these long-term questions. After this PR is closed, we'll copy these notes to forum discussions or github issues.

pieper (approved):
Looks great, thank you 👍 

I agree we can refine the way it behaves after some more experience.  

How will be app behave if there is no network or some kind of proxy?

Have we confirmed we will be able to collect analytics from the update checks?  Knowing how often the app is started with which extensions installed (in from where) will be very interesting to see.

lassoan:
> How will be app behave if there is no network or some kind of proxy?

If it cannot get a response, or cannot parse the response as a json object, or the response does not contain a few required fields then the check fails and the message will be that no updates are available (and errors are logged in the application log).

> Have we confirmed we will be able to collect analytics from the update checks? Knowing how often the app is started with which extensions installed (in from where) will be very interesting to see.

If automatic update check is enabled then the application contacts the extension server and the application download server at startup if enough time passed since the cached version expires. This time is not set to 24 hours. So, each Slicer user that have auto-update enabled will trigger a server query once a day when Slicer is used.

jamesobutler:
Ok I was thinking we would make it more in your face as a notification and if people don’t like it then back it off, rather than the other way around as you have now (less in your face and will potentially require making more visible).

lassoan:
Thank you for all the comments. I'll merge this now. I've copied the suggestions about future evolution of this feature to this discourse post: https://discourse.slicer.org/t/stable-update-notification/23655

CMakeLists.txt
```diff
@@ -282,6 +282,9 @@ mark_as_superbuild(Slicer_BUILD_QT_DESIGNER_PLUGINS)
 option(Slicer_BUILD_EXTENSIONMANAGER_SUPPORT "Build Slicer extensions manager" ON)
 mark_as_superbuild(Slicer_BUILD_EXTENSIONMANAGER_SUPPORT)
 
+option(Slicer_BUILD_APPLICATIONUPDATE_SUPPORT "Build Slicer with application update support" ON)
```
jamesobutler, line 285: @lassoan Should this be set to OFF by default when going through the SlicerCustomAppTemplate process as a Slicer custom app shouldn’t be checking for Slicer application updates?

jcfr: > Should this be set to OFF by default when going through the SlicerCustomAppTemplate process 

Indeed, default can be set to off if the main application name is different from Slicer.

((I will be back in the office next week and should be able to further comment.))

lassoan: Yes, custom applications could either disable application updates or set a different default update server URL.

jamesobutler: https://github.com/KitwareMedical/SlicerCustomAppTemplate/pull/43 issued for this

PR #6539, opened 2022-09-20
ENH: Add option to hide the maximize view button

cpinter:
In custom apps usually it is easy to simply find unwanted toolbar elements by object name and hide it. However, in the case of the maximize view button, it is automatically shown if the view can be maximized, which makes it reappear in the custom apps. It is possible to remove it from the layout but it does not work properly in all cases. By adding such an option it is possible to hide it appropriately.

### DIFF
--- a/Libs/MRML/Widgets/qMRMLViewControllerBar.cxx
+++ b/Libs/MRML/Widgets/qMRMLViewControllerBar.cxx
@@ -49,13 +49,6 @@ qMRMLViewControllerBarPrivate::qMRMLViewControllerBarPrivate(
   : QObject(nullptr)
   , q_ptr(&object)
 {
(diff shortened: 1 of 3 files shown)


pieper (approved):
Seems fine to me.  I didn't build it but if it works well for you I'd be okay with merging it.

lassoan:
Looks OK to me, too, but I'm wondering if you really want to just hide this button, or you actually want to disable maximize feature all the places (on double-click, context menu, and view controller). If you want to disable everywhere then it may be better to add a higher-level flag for that somewhere at the layout level.

cpinter:
In this case just hiding it is fine. When it is needed to disable the double click I can change it.

PR #6542, opened 2022-09-20
BUG: Make Web Server module DICOMweb endpoint more robust

lassoan:
Fixed incorrect reading of logging flags from settings (logging was always enabled). Fixed study list loading for cases when optional DICOM fields were not present.

Added OHIF viewer to Web Server API examples.

image


pieper (approved):
LGTM 👍

PR #6546, opened 2022-09-22
ENH: Display available application version in "About" box

lassoan:
If application update is found then display the available version number and link to the download page. If no application update is found then do not display link to download page.

Clarify partially checked meaning in "automatic" update check button.

See discussion: https://discourse.slicer.org/t/stable-update-notification/23655/22


jamesobutler (approved):
I've built and tested. This looks good to address my comments on the discourse post.

lassoan:
Great, thanks for testing!

PR #6550, opened 2022-09-26
ENH: Color legend is to be shown for slice view, visible in the 3D view

MichaelColonel:
Color legend is to be shown in 3D view for volumes that are shown in slice views, that are visible in the 3D view. Part of #6111 issue.

Single color bar in 3D.
image

Multiple color bars in 3D.
image

### DIFF
--- a/Modules/Loadable/Colors/MRMLDM/vtkMRMLColorLegendDisplayableManager.cxx
+++ b/Modules/Loadable/Colors/MRMLDM/vtkMRMLColorLegendDisplayableManager.cxx
@@ -29,6 +29,7 @@
 #include <vtkMRMLSliceLogic.h>
 #include <vtkMRMLSliceNode.h>
 #include <vtkMRMLViewNode.h>
+#include <vtkMRMLModelNode.h>
 #include <vtkMRMLVolumeNode.h>
 
 // VTK includes
@@ -66,7 +67,7 @@ public:
   void Modified();
 
   vtkMRMLSliceCompositeNode* FindSliceCompositeNode();
-  bool IsVolumeVisibleInSliceView(vtkMRMLVolumeNode* volumeNode);
+  bool IsVolumeVisibleInSliceView(vtkMRMLSliceCompositeNode* sliceCompositeNode, vtkMRMLVolumeNode* volumeNode);
 
   // Update color legend
   void UpdateColorLegend();
@@ -160,13 +161,13 @@ bool vtkMRMLColorLegendDisplayableManager::vtkInternal::ShowActor(vtkSlicerScala
 
 //---------------------------------------------------------------------------
 bool vtkMRMLColorLegendDisplayableManager::vtkInternal::IsVolumeVisibleInSliceView(
-  vtkMRMLVolumeNode* volumeNode)
+  vtkMRMLSliceCompositeNode* sliceCompositeNode, vtkMRMLVolumeNode* volumeNode)
 {
   if (!volumeNode)
     {
     return false;
     }
-  if (!this->SliceCompositeNode.GetPointer())
+  if (!sliceCompositeNode)
     {
     return false;
     }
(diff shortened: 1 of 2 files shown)


Modules/Loadable/Colors/MRMLDM/vtkMRMLColorLegendDisplayableManager.cxx
```diff
@@ -254,19 +296,34 @@ bool vtkMRMLColorLegendDisplayableManager::vtkInternal::UpdateActor(vtkMRMLColor
...
       else
         {
         // 3D view
-        // For now don't show color legends for volumes in 3D views.
-        // In the future, color legend could be shown for volumes that are shown in slice views
+        // Color legend is be shown for volumes that are shown in slice views
         // that are visible in the 3D view.
-        visible = false;
+        std::vector< vtkMRMLNode* > sliceNodes;
+        this->External->GetMRMLScene()->GetNodesByClass("vtkMRMLSliceNode", sliceNodes);
```
lassoan, line 312: Since this update function can be called quite often, it is good to pay a little attention to performance.
GetNodesByClass is a relatively expensive function, so it would be better to get the slice nodes (and model display nodes, etc. from the application logic). Something like this:

```
vtkMRMLApplicationLogic* mrmlAppLogic = this->External->GetMRMLApplicationLogic();
        vtkCollection* sliceLogics = mrmlAppLogic ? mrmlAppLogic->GetSliceLogics() : nullptr;
        if (sliceLogics)
          {
          vtkMRMLSliceLogic* sliceLogic;
          vtkCollectionSimpleIterator it;
          for (sliceLogics->InitTraversal(it);
            (sliceLogic = vtkMRMLSliceLogic::SafeDownCast(sliceLogics->GetNextItemAsObject(it)));)
            {
             ...
```

Modules/Loadable/Colors/MRMLDM/vtkMRMLColorLegendDisplayableManager.cxx
```diff
@@ -254,19 +296,34 @@ bool vtkMRMLColorLegendDisplayableManager::vtkInternal::UpdateActor(vtkMRMLColor
...
+        for (auto aNode : sliceNodes)
+          {
+          vtkMRMLSliceNode* sliceNode = vtkMRMLSliceNode::SafeDownCast(aNode);
+          if (sliceNode && sliceNode->GetSliceVisible())
+            {
+            vtkMRMLSliceCompositeNode* sliceCompositeNode = vtkMRMLSliceLogic::GetSliceCompositeNode(sliceNode);
+            visible &= this->IsVolumeVisibleInSliceView(sliceCompositeNode, volumeNode);
```
lassoan, line 319: This would pollute all 3D viewers with color legends from all slice views.

A little improvement would be to ignore slice view nodes in different view groups as the 3D view node (then, for example in a 4-over-4 view, slice nodes in the top row would not make a color legend show up in the bottom row).

However, it would be better to only show the color legend if the slice is shown in that 3D view (eye icon is enabled in the slice view). It could work like this:

```
            vtkMRMLModelNode* sliceModelNode = sliceLogic->GetSliceModelNode();
            vtkMRMLDisplayNode* sliceModelDisplayNode = (sliceModelNode ? sliceModelNode->GetDisplayNode() : nullptr);
            if (sliceModelDisplayNode)
              {
              if (sliceModelDisplayNode->GetVisibility()
                && sliceModelDisplayNode->GetVisibility3D()
                && sliceModelDisplayNode->IsDisplayableInView(viewNode->GetID())
                && this->IsVolumeVisibleInSliceView(volumeNode, sliceLogic->GetSliceCompositeNode()))
                {
                // found a slice view that is displayed in this 3D view and the volume is visible on that slice
                visible = true;
                break;
                }
              }
```

For this to work, we would also need to observe display visibility changes of slice model nodes. Right now, when the eye icon is clicked in the slice view controller, this `UpdateActor` method is not called, so the color legend does not get displayed or hidden correctly (only when the volume node changes). Maybe you could give it a try to add these observers, but if it requires a lot of code or lots of additional observations then we can drop this and just keep the current simpler implementation (act if slice view was displayed in the 3D view).

lassoan (commented):
Thanks a lot, this will be great. I've added a few comments.

Modules/Loadable/Colors/MRMLDM/vtkMRMLColorLegendDisplayableManager.cxx
MichaelColonel: I will add observers to model nodes shortly.
If the eye icon is clicked and the slice view is showed up on a 3D view, the one can toggle visibility (Show Off / Show On) of color bar to update it visibility on a 3D.

MichaelColonel:
Better processing of color bar visibility on the 3D has been added, when the "eye" icon on the slice controller widget is clicked.

lassoan (approved):
Thanks a lot @MichaelColonel. I've refined the observations a bit (color legend displayable manager updates were triggered a bit too frequently, e.g., for any slice offset or zoom operation; while some were missed, e.g., when foreground/background/label layer selection was changed) and force-pushed the change. I wait for the automatic checks to complete and it should be all ready to go.

PR #6554, opened 2022-09-27
BUG: Fix plane area measurement not updating

Sunderlandkyl:
When the size of the plane was changed, the area measurement was not updated. Fixed by calling UpdateAllMeasurements() when SetSize or SetPlaneBounds is called.


lassoan (approved):
Thank you, it looks good.

PR #6559, opened 2022-10-02
ENH: add slicer launch demo to slicer-hosted OHIF

pieper:
WIP

Adds a slicer button to the OHIF toolbar based on request from @lassoan.  Usage and final implementation still open for further discussion.

This is based on work first done at project week in January of 2020:

https://projectweek.na-mic.org/PW33_2020_GranCanaria/Projects/OHIFSlicerBridge/


Modules/Scripted/WebServer/Resources/docroot/browse.html
```diff
@@ -47,7 +54,55 @@
...
+      extensions: [
+        {
+          id: "slicerExtension",
+          getToolbarModule() {
+            return {
+              definitions: [
+                {
+                  id: "slicer",
+                  label: "3D Slicer",
+                  icon: "exclamation-triangle",
```
lassoan, line 66: This looks like an error. I would recommend to use the `cube`, `edit`, or `bars` icon instead.

```suggestion
                  icon: "cube",
```

Modules/Scripted/WebServer/Resources/docroot/browse.html
```diff
@@ -47,7 +54,55 @@
...
+            return {
+              definitions: {
+                launchSlicer: {
+                  commandFn: function() {
+                    function launchSlicer() {
+                      const hrefParts = window.location.href.split('/');
+                      const studyUID = hrefParts[hrefParts.length-1];
+                      let wadoRoot = window.viewer._appConfig.servers.dicomWeb[0].wadoRoot;
+                      wadoRoot = window.encodeURIComponent(wadoRoot);
+                      let url = `slicer://viewer/?studyUID=${studyUID}&dicomweb_endpoint=${wadoRoot}&bulk_retrieve=0`;
```
lassoan, line 85: Opening the image from another computer resulted in the error:

Malformed URL: /dicom (Traceback (most recent call last):
  File "C:\Users\andra\AppData\Local\NA-MIC\Slicer 5.1.0-2022-09-27\lib\Slicer-5.1\qt-scripted-modules\DICOMLib\DICOMUtils.py", line 860, in importFromDICOMWeb

Because Slicer got this URL:

slicer://viewer/?studyUID=1.2.826.0.1.3680043.8.498.17263129467074741651245868834987728570&dicomweb_endpoint=%2Fdicom&bulk_retrieve=0

Would it be possible to automatically prepend the server hostname (http://example.com:2016) to ${wadoRoot} automatically? If not, then we could add this comment:

```suggestion
                      // To allow access from other computers, specify externally accessible host address
                      // in dicomweb_endpoint. For example:
                      // ...&dicomweb_endpoint=http://example.com:2016${wadoRoot}
                      let url = `slicer://viewer/?studyUID=${studyUID}&dicomweb_endpoint=${wadoRoot}&bulk_retrieve=0`;
```

lassoan (commented):
This is awesome, thank you. It works almost perfectly. Just added two small comments.

Modules/Scripted/WebServer/Resources/docroot/browse.html
pieper: In general the `wadoRoot`is a full URL but it's a special case for Slicer's dicomweb that it's just a relative address.  I guess we could detect that case and prefix with the `window.location.origin`.  Can you try that on your end to see if it works?

Something like:
```
if (!wadoRoot.startsWith("http") {
  wadoRoot = window.location.origin + "/" + wadoRoot;
}
```

lassoan: It works, after some syntax fixes:

```js
                      if (!wadoRoot.startsWith("http")) {
                        wadoRoot = window.location.origin + wadoRoot;
                      }
```

Instead of checking if the URL starts with `http`, wouldn't it be nicer to convert relative to absolute url like this (found the code on StackOverflow):

```js
    var getAbsoluteUrl = (function() {
      var a;
      return function(url) {
        if(!a) a = document.createElement('a');
        a.href = url;
        return a.href;
      };
    })();
```

pieper: Yes, that's clever and I think it would work.

lassoan: OK. I've forced pushed an update with this change.

pieper:
If you find this helpful let's go ahead and merge it just to clean up the PR list.

lassoan:
Yes, it's useful, thank you. I'll show tomorrow how it can be used.

PR #6561, opened 2022-10-04
ENH: Add GET /gui/version and POST /mrml Web Server methods

lassoan:
Version query can be used to check if an appropriate Slicer version is running. MRML put can be used to load a file from local path or URL into the scene.


Modules/Scripted/WebServer/WebServerLib/SlicerRequestHandler.py
```diff
@@ -679,15 +693,45 @@ def setViewersLayout(layoutName):
         raise ValueError("Unknown layout name: " + layoutName)
 
     def gui(self, request):
-        """return a png of the application GUI.
-        :param contents: {full, viewers}
-        :param viewersLayout: {fourup, oneup3d, ...} slicer.vtkMRMLLayoutNode constants (SlicerLayout...View)
-        :return: png encoded screenshot after applying params
+        """Manage the application GUI.
+
+        - Path: /gui/version
```
jcfr, line 698: The `version` not being specific to the `gui`, what about `/application/version` ?

jcfr: Or since this call is already associated with the `/slicer` handler, having it under `/slicer/version` could work well ?

jcfr:
As part of this pull request, may also be worth updating the documentation. See https://slicer.readthedocs.io/en/latest/user_guide/modules/webserver.html#slicer-endpoints

lassoan:
@jcfr Thanks for the comment, I've moved `/slicer/gui/version` to `/slicer/system/version` and added full documentation of the current `/slicer` API.

PR #6562, opened 2022-10-04
BUG: Minor VectorToScalarVolume module improvements

### LINKED ISSUE #6478: Redundant line in VectorToScalarVolume.py?
Hi,

This call to `updateButtonStates()` seems odd -- it looks like it doesn't belong in `updateParameterNodeFromGui()` (and also, the first line in the called method accesses attributes on `self._parameterNode` unconditionally, while the next line here checks if it is `None`).

https://github.com/Slicer/Slicer/blob/6931e0d9820a86be8c97f77225bdd3a479980820/Modules/Scripted/VectorToScalarVolume/VectorToScalarVolume.py#L268-L272

lassoan:
When scene was closed or the module was reloaded then the module threw exceptions. Fixed by updating the module to use the latest scripted module template.

Average computation was not correct for integer volumes. For example for a volume containing 30, 50, 100 in 3 channels the average is 60 but it was computed to be 59 (due to casting to int). Fixed by rounding of the average computation result instead of just casting the floating-point value.

fixes #6478

### DIFF
--- a/Modules/Scripted/VectorToScalarVolume/CMakeLists.txt
+++ b/Modules/Scripted/VectorToScalarVolume/CMakeLists.txt
@@ -1,4 +1,3 @@
-
 #-----------------------------------------------------------------------------
 set(MODULE_NAME VectorToScalarVolume)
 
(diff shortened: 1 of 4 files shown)


pieper (dismissed):
LGTM 👍

PR #6566, opened 2022-10-07
BUG: Fix bug in closed curve resampling

cpinter:
The number of segments when resampling a curve was "number of given points - 1", which is fine for open curves, but for closed curves it has to be just the number of given points. Added a simple fix in the widget function.

### DIFF
--- a/Modules/Loadable/Markups/Widgets/qMRMLMarkupsCurveSettingsWidget.cxx
+++ b/Modules/Loadable/Markups/Widgets/qMRMLMarkupsCurveSettingsWidget.cxx
@@ -318,7 +318,7 @@ void qMRMLMarkupsCurveSettingsWidget::onApplyCurveResamplingPushButtonClicked()
     {
     return;
     }
-  bool isClosed = (vtkMRMLMarkupsClosedCurveNode::SafeDownCast(inputNode) != nullptr);
+  bool isClosed = inputNode->GetCurveClosed();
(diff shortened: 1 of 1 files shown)


Modules/Loadable/Markups/Widgets/qMRMLMarkupsCurveSettingsWidget.cxx
```diff
@@ -318,6 +318,7 @@ void qMRMLMarkupsCurveSettingsWidget::onApplyCurveResamplingPushButtonClicked()
     {
     return;
     }
+  bool isClosed = (vtkMRMLMarkupsClosedCurveNode::SafeDownCast(inputNode) != nullptr);
```
lassoan, line 321: We usually use `markupsNode->GetCurveClosed()` to decide if it is a closed curve. It is because curve management infrastructure is a feature of the vtkMRMLMarkupsNode base class. There can be custom markups that take advantage of this infrastructure (and display closed curve) but derive directly from this vtkMRMLMarkupsNode base class and not from vtkMRMLMarkupsClosedCurveNode.

```suggestion
  bool isClosed = inputNode->GetCurveClosed();
```

lassoan (commented):
Thank you, it looks good, I've just added a minor suggestion.

cpinter:
Thanks! Ready to merge whenever I guess

Issue #6565, opened 2022-10-04
Make CORS optional


pieper:
## Is your feature request related to a problem? Please describe.

[CORS](https://reflectoring.io/complete-guide-to-cors/) allows access to the WebServer from any code running in the browser so it isn't appropriate in all situations.

## Describe the solution you'd like

Instead it CORS should be off by default, and only allowed when the user requests it (like we do with the exec API).

## Describe alternatives you've considered

It could be disabled universally but there are valid use cases where it can be useful.

## Additional context

The line here should only be added if CORS is enabled:

https://reflectoring.io/complete-guide-to-cors/

PR #6569, opened 2022-10-08
ENH: make CORS optional in WebServer

pieper:
Option is persistent in settings and default is off.


pieper:
Fixes #6565

lassoan (approved):
It looks good to me. Can the  OHIF viewer still access the server when the viewer is opened on another computer (since the OHIF viewer is hosted on the same server)?

pieper:
Yes, as long as OHIF is hosted from the same server instance (same origin) then the browser will let it request dicomweb resources from that origin.  I didn't get a chance to test though, so you can confirm with your setup.

PR #6572, opened 2022-10-10
ENH: When setting window/level mode also activate it

cpinter:
It can be assumed that when the user wants to change the mode of the window/level mouse mode, they also want to use it right afterwards. Thus, it makes sense to activate it along with setting the mode, to save extra clicks and possible confusion.


lassoan (commented):
I agree, this is a quite common pattern in various software, including very commonly used one, such as MS Office. We also do something like this in Slicer already (in Volume Rendering module we activate cropping if the user user displays the ROI).

Maybe just add to the comment in the code _why_ we switch the mode (to save a button click and reduce user confusion, similarly how it is done elsewhere in Slicer and other software, where adjusting an option of a feature activates that feature)

cpinter:
Thanks @lassoan ! I extended that commit with the reasoning you mentioned.

lassoan (approved):
Thank you!

PR #6563, opened 2022-10-04
Fix lupdate warnings related to tr function calls on private classes

mhdiop:
In the [3D Slicer in My Language](https://chanzuckerberg.com/eoss/proposals/3d-slicer-in-my-language-internationalization-and-usability-improvements/) project, we use the _QT **lupdate** tool_ to extract traductible strings. However, when we run that tool on the [current source code](https://github.com/Slicer/Slicer), we note the presence of some warnings that may affect the translation quality. 

This commit fixes the `cannot invoke tr() like this` warning that is related to _tr function calls_ on private classes. 

More details in this issue can be found [here](https://github.com/mhdiop/Sliceriml/issues/1).

CC: @lassoan @pieper

Issue #6416, opened 2022-06-08
Improve launcher to support running  commands checking for prerequisites


jcfr:
## Is your feature request related to a problem? Please describe.

When Slicer main executable can not be executed due to missing requirements, the user is left to deal with cryptic errors.

For example, see https://github.com/Slicer/Slicer/issues/5251 and https://github.com/Slicer/Slicer/issues/5663

## Describe the solution you'd like

Update the launcher to support running executable _checks_ and reporting useful messages.

All check may be disable specifying `--launcher-disable-checks`

A specific check may be disable specifying `-launcer-disable-check-<name`

Example of setting file supporting a new section called `[Prerequisites]`

```diff
[General]
launcherNoSplashScreen=false
launcherSplashImagePath=<APPLAUNCHER_DIR>/bin/Slicer-SplashScreen.png
launcherSplashScreenHideDelayMs=3000
additionalLauncherHelpShortArgument=-h
additionalLauncherHelpLongArgument=--help
additionalLauncherNoSplashArguments=--no-splash,--help,--version,--home,--program-path,--no-main-window,--settings-path,--temporary-path

+ [Prerequisites]
+
+linuxpackages/help=Check that the required linux packages are installed
+linuxpackages/path=<APPLAUNCHER_DIR>/bin/checkLinuxPackages
+linuxpackages/arguments=
+

[Application]
path=<APPLAUNCHER_DIR>/bin/SlicerApp-real
arguments=
name=Slicer
revision=30996
organizationDomain=www.na-mic.org
organizationName=NA-MIC

[...]
```

  
## Describe alternatives you've considered

The launcher could provide a domain specific language implementing platform specific checks.

## Additional context

* https://github.com/Slicer/Slicer/issues/5251
* https://github.com/Slicer/Slicer/issues/5663
* https://github.com/commontk/AppLauncher

PR #6578, opened 2022-10-11
BUG: Fix blank Extensions manager on Linux due to Chromium sandboxing

### LINKED ISSUE #6577: Extensions manager does not work on some linux systems
## Summary

Extensions manager shows up as an empty window on some linux systems.

Setting QTWEBENGINE_DISABLE_SANDBOX=1 environment variable fixes the issue.

## Environment
- Slicer version: Slicer-5.0.3
- Operating system: Linux only

lassoan:
fixes #6577

@jcfr Please test this, as I've made this change without actually testing it on linux. It would be also nice if you could add some more details to the commit comment (and/or into the code) about why this change is appropriate. Thank you!

### DIFF
--- a/CMake/SlicerBlockCTKAppLauncherSettings.cmake
+++ b/CMake/SlicerBlockCTKAppLauncherSettings.cmake
@@ -142,6 +142,13 @@ if(Slicer_USE_PYTHONQT_WITH_OPENSSL)
     "SSL_CERT_FILE=<APPLAUNCHER_SETTINGS_DIR>/${Slicer_SHARE_DIR}/Slicer.crt"
     )
 endif()
+if(UNIX AND NOT APPLE)
+  # Disable Chromium Sandboxing on Linux because not all systems support it
+  # (see https://github.com/Slicer/Slicer/issues/6577)
+  list(APPEND SLICER_ENVVARS_BUILD
+    "QTWEBENGINE_DISABLE_SANDBOX=1"
+    )
+endif()
 
 # External projects - environment variables
 foreach(varname IN LISTS Slicer_EP_LABEL_ENVVARS_LAUNCHER_BUILD)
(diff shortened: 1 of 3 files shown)


jcfr:
For future reference, the requirements for enabling chromium sandbox are the following:

> * The kernel has to support the anonymous namespaces feature (kernel version 3.8 or later). However, on Debian, Ubuntu, and other Debian-derived distributions, this feature is off by default. It can be turned on by setting /proc/sys/kernel/unprivileged_userns_clone to 1.
> * The kernel has to support the seccomp-bpf feature (kernel version 3.5 or later).
> * Setuid sandboxes are not supported and are thus disabled.

Source: https://doc.qt.io/Qt-5/qtwebengine-platform-notes.html#sandboxing-support

That said, as indicated above and discussed in the post below, updating the requirement for sandboxing would required the enabling of kernel feature (potentially) implying larger security issue. 

> Enabling unprivileged user namespaces can make severe vulnerabilities in the Linux kernel much more easily exploitable. If you did not intend to enable it, you should ensure it is disabled. Numerous vulnerabilities that are found regularly are often only exploitable by unprivileged users if unprivileged user namespaces are supported and enabled by the kernel. Unless you truly need it, just disable it.

Source: https://security.stackexchange.com/questions/209529/what-does-enabling-kernel-unprivileged-userns-clone-do/209533#209533

For now, we are systematically disabling chromium sandboxing. This ensures the extension manager will work on system without the required kernel feature enabled.

An alternative option could have be to move forward with a more complex mechanism requiring to implement support for checks described in https://github.com/Slicer/Slicer/issues/6416. The implemented check would allow to conditional check an env. variable based on the its logic.

Additional references:
* https://chromium.googlesource.com/chromium/src/+/master/docs/design/sandbox.md
* https://chromium.googlesource.com/chromium/src/+/refs/heads/main/docs/linux/sandboxing.md

jcfr:
Since the fix for this was integrated through https://github.com/Slicer/Slicer/pull/6563, I am closing this PR.

> It would be also nice if you could add some more details [...] about why this change is appropriate

@lassoan I added some more details above. I am also wondering which place would be best to reference these details, it could either be an entry on the GitHub Slice wiki or discourse post with a `security` tag.

lassoan:
> Since the fix for this was integrated through https://github.com/Slicer/Slicer/pull/6563, I am closing this PR.

Oops, sorry, I did not notice that commit when I edited and force-pushed a commit comment change.

> I am also wondering which place would be best to reference these details

Since #6577 is referenced in the commit comment and in the source code, what do you think about copying (or adding a reference to) your detailed analysis in #6577.

jcfr:
> what do you think about copying your detailed analysis in https://github.com/Slicer/Slicer/issues/6577.

Good idea. We should be all set :ok_hand:

PR #6583, opened 2022-10-13
BUG: Fix lupdate warnings related to missing Q_OBJECT macro

mhdiop:
In the [3D Slicer in My Language](https://chanzuckerberg.com/eoss/proposals/3d-slicer-in-my-language-internationalization-and-usability-improvements/) project, we use the QT _lupdate_ tool to extract traductible strings. However, when we run that tool on the [current source code](https://github.com/Slicer/Slicer), we note the presence of some warnings that may affect the translation quality.

This commit fixes the `Class 'SomeClassName' lacks Q_OBJECT macro` warning that is related to tr function calls on classes with no Q_OBJET macro declaration.

More details in this issue can be found [here](https://github.com/mhdiop/Sliceriml/issues/1).

@lassoan As discussed during the meeting, the Q_OBJECT macro introduced in `Base/QTGUI/qSlicerExtensionsLocalWidget.cxx` is still causing a compilation issue on my side with the following error message : `'staticMetaObject': variable with internal linkage declared but not defined`. There may be usefull to verify if it's related to my development environment or something else.

CC: @lassoan @pieper

### DIFF
--- a/Base/QTGUI/qSlicerExtensionsLocalWidget.cxx
+++ b/Base/QTGUI/qSlicerExtensionsLocalWidget.cxx
@@ -153,17 +153,17 @@ public:
     if (installed && !loaded)
       {
       this->StatusLabel->setVisible(true);
-      this->StatusLabel->setText(tr("Install pending restart"));
+      this->StatusLabel->setText(qSlicerExtensionsLocalWidget::tr("Install pending restart"));
       }
     else if (scheduledForUpdate)
       {
       this->StatusLabel->setVisible(true);
-      this->StatusLabel->setText(tr("Update pending restart"));
+      this->StatusLabel->setText(qSlicerExtensionsLocalWidget::tr("Update pending restart"));
       }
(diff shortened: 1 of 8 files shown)


lassoan (approved):
Thank you @mhdiop, it looks all good.

The build problem with the Q_OBJECT was due to that those classes that you added Q_OBJECT to were declared in the cxx file. The MOC only deals with .h files, it does not generate code from .cxx files. One solution could have been to move the declaration of those classes into a .h file, but they are declared in the .cxx file for the first place because they are low-level implementation classes that should not be visible developers or translators (because we may want to change the low-level implementation without impacting the translations or public API of the class). I've fixed the issue by simple using the public class (`qSlicerExtensionsLocalWidget`) as namespace for the translation.

mhdiop:
Than you @lassoan for your feedback.

I'll use this marking approach if I face a similar case.

Regards.

PR #6588, opened 2022-10-17
COMP: Fix build error for Application Update off and I18N on

jamesobutler:
This appears to have been a typo introduced in https://github.com/Slicer/Slicer/commit/3ce46788cde6d74167cd5bb7639855dc1e59401e

I ran into this issue while testing https://github.com/KitwareMedical/SlicerCustomAppTemplate/pull/43 where I had `Slicer_BUILD_APPLICATIONUPDATE_SUPPORT` set to `OFF` and `Slicer_BUILD_I18N_SUPPORT` set to `ON`.

### DIFF
--- a/Base/QTGUI/qSlicerSettingsGeneralPanel.cxx
+++ b/Base/QTGUI/qSlicerSettingsGeneralPanel.cxx
@@ -95,7 +95,7 @@ void qSlicerSettingsGeneralPanelPrivate::init()
 #endif
 
   bool applicationUpdateEnabled = false;
-#ifdef Slicer_BUILD_I18N_SUPPORT
+#ifdef Slicer_BUILD_APPLICATIONUPDATE_SUPPORT
   applicationUpdateEnabled = qSlicerApplicationUpdateManager::isApplicationUpdateEnabled();
   if (applicationUpdateEnabled)
     {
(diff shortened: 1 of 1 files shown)


jcfr (dismissed):

jamesobutler:
I might have to put an #ifdef around the following as well?

https://github.com/Slicer/Slicer/blob/367c2bb2426e36ad9c7cb41fdbd4dd45b2ddb972/Base/QTGUI/qSlicerSettingsGeneralPanel.h#L55

lassoan:
> I might have to put an #ifdef around the following as well?

Yes, either use ifdef to conditionally remove method declaration in the .h file, or in the .cxx file use ifdef the method content (but not to remove the whole method).

Base/QTGUI/qSlicerSettingsGeneralPanel.h
```diff
@@ -52,7 +52,9 @@ public slots:
   void openSlicerRCFile();
 
 protected slots:
+#ifdef Slicer_BUILD_APPLICATIONUPDATE_SUPPORT
```
jcfr, line 55: Could you check that moc correctly regenerate the `moc_qSlicerSettingsGeneralPanel.cpp` file by deleting it ? 

Explicitly including `vtkSlicerConfigure.h` may also be sensible.

In the past, we had to explicitly pass macro to moc.

jcfr: Per @lassoan suggestion, if this is effectively an issue, I think it may be easier to only exclude the code in the implementation.

jamesobutler: I've pushed the update to exclude the code in `updateAutoUpdateApplicationFromManager` instead

jamesobutler:
Ok sorry for the pushes here as was building quickly at the end of the work day and trying to sort through other build errors in my custom app due to the VTK incompatibilities with IGSIO code (issues already being tracked in Slicer and corresponding extensions).

jcfr (approved):

PR #6589, opened 2022-10-17
DICOM export fixes

lassoan:
Make it easier to save mrb file embedded into a DICOM file (do not require having a reference DICOM file).

Also fixed a couple of warnings and other minor issues.


pieper (approved):
I didn't try it but the changes look good to me.

Are the subject hierarchy and window/level fixes included here because they are needed to make this work?  If not maybe they should go in a different PR.

lassoan:
They are only related in that these warnings appeared while I tested these changes. Now that Qt and VTK warnings show up in the Python console they are much easier to see immediately when they happen.

PR #6604, opened 2022-10-20
ENH: Add https support to vtkHTTPHandler

lassoan:
It allows referring to files hosted on https servers in the MRML scene.

May be useful to distribute scene files with remote data - see https://discourse.slicer.org/t/embedding-uri-in-a-mrml-scene-file-for-viewing-the-mrml-scene-on-a-remote-slicer-application/25373/9


pieper (approved):
LGTM, lets see if it turns out to be useful.

lassoan:
I agree, I'm not sure how useful as is, because we still end up downloading all data at once.

It has potential in the future: it can make scene loading faster if we don't wait for loading/downloading of all data, or if we don't download/load data at all until it is needed. It would be even more useful if only that part of the data was downloaded that was actually needed, at the appropriate resolution.

PR #6605, opened 2022-10-20
ENH: Update pre-commit config to include pyupgrade

jcfr:
By adding `pyupgrade` to the pre-commit configuration, it enables the following:

* GitHub Action report: Suggested changes are reported in the error log associated with the pre-commit GitHub Action workflow.

* Consolidated local pre-commit checks and updates by running:

  ```
  ```

As described in a4ea79909 (STYLE: Upgrade python syntax to 3.9 and newer), note that:

1. Running the command twice may be needed to ensure that strings associated with `logging` package are updated to use f-string.

2. Revisiting some of the changes automatically proposed may be required.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,3 +13,9 @@ repos:
   hooks:
   - id: flake8
     additional_dependencies: [flake8-bugbear==22.9.23]
+
+- repo: https://github.com/asottile/pyupgrade
+  rev: v3.1.0
+  hooks:
+  - id: pyupgrade
+    args: [--py39-plus]


jcfr:
_Originally posted by @jamesobutler in https://github.com/Slicer/Slicer/issues/6376#issuecomment-1284577129_

Here' a screenshot from a private repo with an example of `pyupgrade` indicating changes:

image

jamesobutler (approved):

jamesobutler:
@jcfr CommitCheck required status check should be replaced by the GitHub Actions "Commit Message Check"

PR #6612, opened 2022-10-21
Fix VTKObservationMixin so that observers are removed if object reference is not kept in Python

jcfr:
If an observer is attached to an object through `VTKObservationMixin`, and if a reference to the object is not kept in Python, then the observer was automatically garbage collected from the `WeakKeyDictionary` even if the VTK object still exists. This means the observer cannot be removed with `removeObserver()`/`removeObservers()`.
    
This commit partially reverts 1e6d068 (BUG: Fix VTKObservationMixin preventing observed nodes from being released) by removing the use of `WeakKeyDictionary`.
    
It also marks `test_releaseNodes` as expected to fail and `test_removeObserver_issue6610` as expected to pass.

### DIFF
--- a/Base/Python/slicer/tests/test_slicer_util_VTKObservationMixin.py
+++ b/Base/Python/slicer/tests/test_slicer_util_VTKObservationMixin.py
@@ -163,6 +163,47 @@ class SlicerUtilVTKObservationMixinTests(unittest.TestCase):
         foo.removeObserver(object2, event, callback2)
         self.assertEqual(len(foo.Observations), 0)
 
+    @unittest.expectedFailure
+    def test_removeObserver_issue6610(self):
+        """test_removeObserver_issue6610
+
+        If an observer is attached to a VTK object through :class:`slicer.util.VTKObservationMixin`,
+        and if that object is itself referenced by another VTK object (e.g `vtkCollection`), removing the
+        observation using :func:`slicer.util.VTKObservationMixin.removeObserver()` is expected to remove
+        the observer even if there was no reference to the object kept in Python.
+
+        See https://github.com/Slicer/Slicer/issues/6610
+        """
+        # Case 1: A reference to the object is kept in Python
+        callback = unittest.mock.Mock()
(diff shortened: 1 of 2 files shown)


jcfr:
@Sunderlandkyl

At your convenience, could you check that if by using the reverted implementation of `VTKObservationMixin` currently associated with this pull request, the problem is still occurring ?  

This would provide valuable insight to help address the root cause :pray: 

cc: @allemangD

Sunderlandkyl:
@jcfr Confirmed that this PR fixes the problem.

lassoan:
I felt the implementation with Python weak references was too complex, as it relied on two very different weak reference counting schemes working together, but I could not pinpoint any particular issues. @Sunderlandkyl finding of Python weak references not knowing about non-Python references is a serious issue.

If we can accept the slight inconvenience of requiring explicitly removing observations then the current implementation in this pull request is good.

If we want to avoid requiring explicit removal of observations then we could either improve this Python weakref-based implementation to take into account VTK reference counting as well, or we could switch to using a [vtkObserverManager](https://apidocs.slicer.org/master/classvtkObserverManager.html). Using the observer manager would allow simpler implementation. However, we usually keep a reference to observed nodes anyway (because they are important nodes), so for a full solution (that would automatically remove references and observations when a node is removed from the scene) we would also need a vtkObject weak reference in Python (e.g., a `vtkWeakPointer<vtkObject>`).

In the short term (for Slicer-5.2) the implementation in this pull request seems appropriate. Later we could consider using a `vtkObserverManager` and make vtkWeakPointer available in Python for more convenience.

Sunderlandkyl:
Should these changes be included in 5.2?

jcfr:
I will revert the change and move forward with the integration as well as create relevant issue to track what need to be done to "properly" address this based on @lassoan and @allemangD comments.

Sunderlandkyl (approved):
Thanks @jcfr looks good to me!

allemangD (approved):
:+1: Thanks JC!

PR #6614, opened 2022-10-21
BUG: Fix crash when exporting segment model node

### LINKED ISSUE #6570: Slicer crashes while on BoneReconstructionPlanner workflow
## Summary

Hi. I got to make the [crash](https://github.com/SlicerIGT/SlicerBoneReconstructionPlanner/issues/63) to happen while debugging on RelWithDebInfo

Here is a screenshot:
BRP_Slicer_bug

## Steps to reproduce

It's hard to reproduce, there are two buttons that sometimes cause it. I have the debugging session open. Can I share it somehow?

## Expected behavior

Slicer doesn't crash

## Environment
- Slicer version: Slicer-5.1.0-2022-10-07

Sunderlandkyl:
Exporting visible segments to model nodes could cause a crash when exporting using the action on the virtual "Segment" items, but wouldn't crash when exporting from the "Segmentation" item. When exporting from segment, the segment parent (the segmentation node) was used as the parent of the folder node. This seems to cause an invalid SH item parent, and could cause a crash later when the model is updated.

Fixed by using the parent of the Segmentation node for the export folder parent.

Fixes #6570


lassoan (approved):
Awesome detective work @Sunderlandkyl and the solution looks good. Thank you for your hard work on getting this in time for the 5.2 release.

PR #6615, opened 2022-10-24
Minor documentation and runtime warning fixes

lassoan:


jcfr:
> re: `BUG: Fix ctkDoubleRangeSlider warning message when entering Volumes`

Thanks for fixing this :pray:

lassoan:
> May be the opportunity to review the following issues: https://github.com/commontk/CTK/issues/570

This particular issue is not due to that the range and step is set in two different methods, but it's true that the API to use these sliders could be more robust.

PR #6618, opened 2022-10-26
BUG: Fix terminology JSON schema URLs

cpinter:
The schema files have been moved into a subdirectory so the URL became invalid. This commit fixes them.

### DIFF
--- a/Modules/Loadable/Terminologies/Resources/AnatomicRegionAndModifier-DICOM-Master.json
+++ b/Modules/Loadable/Terminologies/Resources/AnatomicRegionAndModifier-DICOM-Master.json
@@ -1,6 +1,6 @@
 {
   "AnatomicContextName": "Anatomic codes - DICOM master list",
-  "@schema": "https://raw.githubusercontent.com/qiicr/dcmqi/master/doc/anatomic-context-schema.json#",
+  "@schema": "https://raw.githubusercontent.com/qiicr/dcmqi/master/doc/schemas/anatomic-context-schema.json#",
   "AnatomicCodes": {
     "AnatomicRegion": [
       {
--- a/Modules/Loadable/Terminologies/Resources/SegmentationCategoryTypeModifier-DICOM-Master.json
+++ b/Modules/Loadable/Terminologies/Resources/SegmentationCategoryTypeModifier-DICOM-Master.json
@@ -1,6 +1,6 @@
 {
   "SegmentationCategoryTypeContextName": "Segmentation category and type - DICOM master list",
-  "@schema": "https://raw.githubusercontent.com/qiicr/dcmqi/master/doc/segment-context-schema.json#",
+  "@schema": "https://raw.githubusercontent.com/qiicr/dcmqi/master/doc/schemas/segment-context-schema.json#",
   "SegmentationCodes": {
     "Category": [
       {
--- a/Modules/Loadable/Terminologies/Resources/SegmentationCategoryTypeModifier-SlicerGeneralAnatomy.json
+++ b/Modules/Loadable/Terminologies/Resources/SegmentationCategoryTypeModifier-SlicerGeneralAnatomy.json
@@ -1,6 +1,6 @@
 {
   "SegmentationCategoryTypeContextName": "Segmentation category and type - 3D Slicer General Anatomy list",
-  "@schema": "https://raw.githubusercontent.com/qiicr/dcmqi/master/doc/segment-context-schema.json#",
+  "@schema": "https://raw.githubusercontent.com/qiicr/dcmqi/master/doc/schemas/segment-context-schema.json#",
   "SegmentationCodes": {
     "Category": [
       {


jcfr (commented):
This makes sense, the logic in `vtkSlicerTerminologiesModuleLogic` account for the new URL (support introduced in https://github.com/Slicer/Slicer/commit/56c9d32e0e504de4414739db4bc5ec45e64016d1).

That said, I suggest to prefix the commit title using either `STYLE` or `ENH` as it is not fixing an actual problem. @cpinter Could you confirm ?

https://github.com/Slicer/Slicer/blob/788909d55b280f07ded23c619cea1ea9fb9307ac/Modules/Loadable/Terminologies/Logic/vtkSlicerTerminologiesModuleLogic.cxx#L53-L56

cpinter:
Actually, it seems that the URLs were fixed in the commit you referenced but then broken agai in https://github.com/Slicer/Slicer/commit/5ae5564362b40e43730a289e1fa0531feee8ca73

jcfr:
> then broken again in 5ae5564

Good catch. That said, since the logic used to check if a specific document is associated with a specific schema account for both URL, it was not strictly a breaking change.

Regarding the schema reference, my understanding [^1][^2] is that having the URL being valid is helpful but not strictly required as its purpose is to identify the document.

The application is then responsible to map that to actual schema and perform validation (which we don't do in Slicer).

Also worth noting that the `@schema` convention does not seems to be part of the standard [^3][^4]. There is also a draft proposal [^4] for JSON reference, I am wonder if this means we should use `$schema` instead of `@schema`.

cc: @lassoan @fedorov 

[^1]: https://github.com/Slicer/Slicer/pull/6276#issuecomment-1076814674
[^2]: https://github.com/Slicer/Slicer/pull/6276#issuecomment-1076913131
[^3]: https://datatracker.ietf.org/doc/html/rfc7159
[^4]: https://datatracker.ietf.org/doc/html/draft-zyp-json-schema-04#section-4.1
[^5]: https://datatracker.ietf.org/doc/html/draft-pbryan-zyp-json-ref-03

jcfr:
> I can change the prefix

Changing to `STYLE` may be sensible indeed as runtime behavior is unchanged

lassoan:
Yes, the URL does not have to be an actual download URL. Actual server addresses, organization names, etc. (and therefore the download URL) may change, while the schema name is a unique identifier that is supposed to be unchanged forever.

cpinter:
> is a unique identifier that is supposed to be unchanged forever

In this case I think they should not be URLs at all, because then it suggests that that's where the schema file can be found, which is not really expected.

I changed the prefix.

cpinter:
> use $schema instead of @schema

[This JSON validator](https://www.jsonschemavalidator.net/) also uses the convention with the dollar sign.

For reference, I am working on the issue https://github.com/SlicerRt/SlicerRT/issues/218 and am trying to define the schema. This is why I started looking into this.

lassoan:
> In this case I think they should not be URLs at all, because then it suggests that that's where the schema file can be found, which is not really expected.

That's a good point. `https://raw.githubusercontent.com/qiicr/dcmqi/master/...` is
1. so specific that it misleads people to believe it is an actual download URL, and
2. it is hard to keep it up-to-date.

If we change the schema URL then we should use a simpler URL, such as `http://qiicr.org/schemas/...`. It would be nice if we could even redirect to the actual file on github (it should be possible to do it with some web server redirect rules).

This is what Microsoft does, too. They use a schema names like https://developer.microsoft.com/json-schemas/... (see for example [here](https://learn.microsoft.com/en-us/microsoftteams/platform/resources/schema/localization-schema)), but the schema is actually stored on github, at https://github.com/Microsoft/json-schemas.

cpinter:
> Separately - the URL below is just invalid - it does not exist in any of the schemas that are available in dcmqi. Is this just a bug/typo that should be fixed?

The old (now invalid because it has been moved into the schemas folder) URL is allowed so that terminology files defined before this relocation of the json file are still accepted. You can see in this commit https://github.com/Slicer/Slicer/commit/56c9d32e0e504de4414739db4bc5ec45e64016d1 that the new, valid URL was added, but the old one kept for this sort of backwards compatibility.

fedorov:
Thanks for the clarification, I missed it. It is unfortunate that I didn't think of this consequence while re-organizing that folder back in 2017.

lassoan:
> It is supposed to be the actual download URL, and from what I see in the dcmqi repository in https://github.com/QIICR/dcmqi/tree/master/doc/schemas, they are such downloadable URLs

It is not a hard requirement to make a schema file avaialable at that URL on the public internet. Its primary role is to serve as a unique identifier. It is just a nice convenience feature that this URI may be used to obtain the schema (because then editors can automatically retrieve the schema and use it for documentation, syntax highlighting, validation checks, etc). But an application may have other mechanisms to obtain a schema file based on a schema URI. If you lose access over the download server and the schema URI does not download the schema file anymore, then it may still make sense to keep the same schema URI, especially if files with this schema are widely circulated.

In this specific case, I would not change https://raw.githubusercontent.com/qiicr/dcmqi/master/doc/anatomic-context-schema.json to https://raw.githubusercontent.com/qiicr/dcmqi/master/doc/schemas/anatomic-context-schema.json# because they are both bad. They are both too complex and they are not under QIICR's control. These github URLs will not disappear overnight, but github may change download URL syntax or may arbitrarily limit access for any reasons (security, bandwith use, geographical location, etc.) at any time. If the schema URI is ever changed again, it should be something like https://qiicr.org/schemas/... or https://schema.qiicr.org/..., and they can be redirected to github with redirection rules on the qiicr.org web server.

jamesobutler:
Something as simple as renaming the default branch of https://github.com/QIICR/dcmqi from `master` to `main` would make these `https://raw.githubusercontent.com` type links invalid.

For renaming a branch:
> Although file URLs are automatically redirected, raw file URLs are not redirected. 

https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-branches-in-your-repository/renaming-a-branch#about-renaming-branches

fedorov:
Indeed, you convinced me that the current approach is not good and is not reliable. I wish we had this discussion before those schemas were published!

but:
* since it is not clear if there is support for schema versioning, or established standard practice in JSON-Schema, even if we had a reliable way to dereference the link, it would not be possible to communicate the version of the schema that was used to create the specific document
* I am not convinced dereferencing via qiicr.org is more reliable/stable than the github download link, since it would be dependent on timely domain fee payments and maintenance

Maybe the way to do it, if one would do it over, would be to use versioned DOIs from a service like Zenodo https://zenodo.org/record/1256592 ?

jcfr:
While discussing this during the [2022-11-01 hangout](https://discourse.slicer.org/t/2022-11-01-weekly-meeting/26002), three competing priorities were identified:
* Stable schema URL (do not change often)
* URL allowing to download schema
* Simple schema URL with domain, version and name

We also discussed:
* Having schemas consolidated under `https://slicer.org/schemas/` through the creation of `github.com/Slicer/schemas`
  * Would we apply the same approach to `qiicr.org` and have `qiicr.org/schemas` ? 
  * ASsociated challenge related to maintenance of `qiicr.org`
* Reaching out to maintainer of `dicom.nema.org`

jcfr:
Considering that ensuring domain like `qiccr.org` maintain redirects for the foreseeable future, we are suggesting to:
* keep track of mapping between `Schema URL as identifier` and `Schema URL as downloadable asset` in a json file (conceptually similar [DICOMExtensions.json](https://github.com/Slicer/Slicer/blob/main/Modules/Scripted/DICOM/DICOMExtensions.json))
* implement a C++ logic (that would be python wrapped) allowing to obtain the downloadable URL given a `Schema URL as identifier`

Next steps:
* Define where the mapping file would be located
* Consolidate existing json related logic and add support for schema validation

cpinter:
Thank you for discussing it and the update about the potential solution! Yes it sounds good to me to have such a mapping.

jcfr:
For reference, I am duplicating here a note I also added in the referenced issue:

> The use of https://json-schema.org/ should also be investigated.

PR #6623, opened 2022-10-27
Add resourcepath helper to scripted module

jcfr:
In addition of few docstring fixes, this pull request streamline referencing of resources during them module `setup()`, this function adds the function `resourcePath` to `ScriptedLoadableModule` similar to the one already available in `ScriptedLoadableModuleWidget`.


lassoan (approved):

PR #6626, opened 2022-10-28
BUG: Fix mouse cursor not being changed for views not mapped in layout

Sunderlandkyl:
When changing mouse modes, the cursor is not updated for views that are not mapped in the current layout. This causes an incorrect cursor to be displayed in views that are not part of the layout when the mouse mode is changed.

Example:
- Start in four-up layout
- Change to window/level mouse mode
- Maximize red slice view
- Change to view transform mouse mode
- Restore view layout
- Mouse cursor still shows window/level cursor in all views except the red slice == *Error*

Fixed by updating view cursors even when the view is not mapped in the current layout.

### DIFF
--- a/Base/QTGUI/qSlicerMouseModeToolBar.cxx
+++ b/Base/QTGUI/qSlicerMouseModeToolBar.cxx
@@ -542,10 +542,11 @@ void qSlicerMouseModeToolBar::changeCursorTo(QCursor cursor)
   for (int i=0; i < layoutManager->threeDViewCount(); ++i)
     {
     qMRMLThreeDView* threeDView = layoutManager->threeDWidget(i)->threeDView();
-    if (!threeDView->mrmlViewNode()->IsMappedInLayout())
-      {
-      continue;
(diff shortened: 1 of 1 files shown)


lassoan (commented):
To make sure that such check does not get back to the code in future refactoring, please add a comment to each place where a check was removed that explains why this has to be done for all views.

Sunderlandkyl:
I've added the comment. I realized that this issue will also occur for views that haven't been created yet.

Ex.
- Start in four-up layout
- Switch to window/level mouse mode
- Change layout to Dual 3D layout
- Mouse cursor will be incorrect in the second 3D view.

lassoan (approved):

lassoan:
This does not fix the new view's cursor, does it?
If not, then it would be nice to update the cursor each time the layout is changed.

Sunderlandkyl:
No, this doesn't include a fix for the new view's cursor.
I'll work on a fix for the cursor in new views.

PR #6627, opened 2022-10-28
ENH: Rename Python console to Python interactor

lassoan:
The "interactor" word probably was inherited from KWWidgets many years ago (KWWidgets/vtkKWTclInteractor.cxx), but it is nowadays universally referred to as Python console or terminal. Since we already use the console word many places in documentation and code, with this commit we change all usages of "interactor" to "terminal" to achieve clear, consistent naming.

This topic was discussed on the Slicer forum and no issues were raised (https://discourse.slicer.org/t/rename-python-interactor-to-python-console/25897).

There were only a few usages of "interactor" in the code, those usages are still functional but are now deprecated.


pieper (dismissed):
LGTM

jamesobutler (commented):
> Since we already use the console word many places in documentation and code, with this commit we change all usages of "interactor" to "terminal" to achieve clear, consistent naming.

So you intend to change usages of "interactor" to "terminal"? Or did you mean "interactor" to "console"?

Also the commit and PR title is `ENH: Rename Python console to Python interactor` which is indicating that it is going from "console" -> "interactor", but the changes here are the opposite as it is "interactor" -> "console"

lassoan:
Good catches, thank you, I will force-push a fix in a minute.

PR #6628, opened 2022-10-28
COMP: Fix Qt dependencies for application update

Connor-Bowley:
When building with Slicer_BUILD_APPLICATIONUPDATE_SUPPORT the commontk/qRestApi is used. The qRestApi depends on the Script module of Qt. This commit expresses that dependency and allows Slicer to be built with extension manager support off and application update support on.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -621,7 +621,9 @@ endif()
       WebChannel
       )
   endif()
-  if(Slicer_BUILD_EXTENSIONMANAGER_SUPPORT)
+  # Both "extension manager" and "application update" require qRestApi external
+  # project itself depending on Qt's Script module
+  if(Slicer_BUILD_EXTENSIONMANAGER_SUPPORT OR Slicer_BUILD_APPLICATIONUPDATE_SUPPORT)
     list(APPEND Slicer_REQUIRED_QT_MODULES Script)
   endif()
   if(Slicer_BUILD_I18N_SUPPORT)


lassoan (commented):
This looks good, thank you. Please add a short explanation in comment in the source code, too (something like "Extensions manager and application check requires qRestApi, which depends on Qt's Script module" would be enough).

CMakeLists.txt
```diff
@@ -621,7 +621,7 @@ endif()
       WebChannel
       )
   endif()
-  if(Slicer_BUILD_EXTENSIONMANAGER_SUPPORT)
+  if(Slicer_BUILD_EXTENSIONMANAGER_SUPPORT OR Slicer_BUILD_APPLICATIONUPDATE_SUPPORT)
```
jcfr, line 624:
```suggestion
  # Both "extension manager" and "application update" require qRestApi external
  # project itself depending on Qt's Script module
  if(Slicer_BUILD_EXTENSIONMANAGER_SUPPORT OR Slicer_BUILD_APPLICATIONUPDATE_SUPPORT)
```

Connor-Bowley: Added comment.

Connor-Bowley:
Added comment in the source.

jcfr:
For reference, the "application check for update" feature was contributed in https://github.com/Slicer/Slicer/commit/3ce46788cde6d74167cd5bb7639855dc1e59401e through https://github.com/Slicer/Slicer/issues/5502

PR #6629, opened 2022-10-29
ENH: Replace deprecated lookFromAxis usage

jamesobutler:
This fixes the following warning which began to be displayed with the CTK update in https://github.com/Slicer/Slicer/commit/2e02a3581f25db91edd23a82a59cc91833512dcd

"This function is deprecated. Use lookFromAxis(const ctkAxesWidget::Axis& axis) instead"

The above deprecation can be seen when using latest Slicer preview and running
`slicer.app.layoutManager().threeDWidget(0).threeDController().lookFromAxis(ctk.ctkAxesWidget.Anterior)`.

### DIFF
--- a/Libs/MRML/Widgets/qMRMLThreeDView.cxx
+++ b/Libs/MRML/Widgets/qMRMLThreeDView.cxx
@@ -431,20 +431,6 @@ vtkMRMLViewNode* qMRMLThreeDView::mrmlViewNode()const
 }
 
 // --------------------------------------------------------------------------
-void qMRMLThreeDView::lookFromViewAxis(const ctkAxesWidget::Axis& axis)
-{
-  Q_D(qMRMLThreeDView);
-  if (!d->MRMLViewNode)
-    {
-    qCritical() << "qMRMLThreeDView::lookFromViewAxis: no valid view node.";
-    return;
-    }
-  double fov = d->MRMLViewNode->GetFieldOfView();
-  Q_ASSERT(fov >= 0.0);
-  this->lookFromAxis(axis, fov);
-}
-
-// --------------------------------------------------------------------------
 void qMRMLThreeDView::resetFocalPoint()
 {
   Q_D(qMRMLThreeDView);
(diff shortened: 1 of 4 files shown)


Libs/MRML/Widgets/qMRMLThreeDView.cxx
```diff
@@ -430,20 +430,6 @@ vtkMRMLViewNode* qMRMLThreeDView::mrmlViewNode()const
...
-{
-  Q_D(qMRMLThreeDView);
-  if (!d->MRMLViewNode)
-    {
-    qCritical() << "qMRMLThreeDView::lookFromViewAxis: no valid view node.";
-    return;
-    }
-  double fov = d->MRMLViewNode->GetFieldOfView();
-  Q_ASSERT(fov >= 0.0);
-  this->lookFromAxis(axis, fov);
```
jamesobutler, line 444: @lassoan I removed this `lookFromViewAxis` because it would just be a pass through to `lookFromAxis` because the FOV usage is deprecated.

jcfr: For backward compatibility, what about  keeping `qMRMLThreeDView::lookFromViewAxis` so that it calls `lookFromAxis(axis)`. Optionally, we could also log a deprecation message like `qWarning() << "This function is deprecated. Use lookFromAxis(const ctkAxesWidget::Axis& axis) instead";`

jcfr: Searching through existing code, the function is being used in a few extension:
* https://github.com/search?l=Python&q=lookFromViewAxis&type=Code
* https://sourcegraph.com/search?q=context:global+lookFromViewAxis+lang:python&patternType=standard

For the sake of addressing the runtime warning, for now, I would suggest to keep the function and have it call `lookFromAxis`

jamesobutler: @jcfr I've restored `qMRMLThreeDView::lookFromViewAxis` and have a runtime deprecation message to specify to use  `lookFromAxis` instead.

jcfr (approved):

PR #6633, opened 2022-10-31
ENH: Make DICOM request handler in WebServer more robust

lassoan:
If a single instance was missing or had DICOM non-conformances then it could make an entire request fail. Fixed the issue by just logging any issues and skipping problematic items.


pieper (approved):
LGTM 👍

PR #6635, opened 2022-11-01
DOC: Add "Vector to Scalar Volume" module documentation

jcfr:
This commit adds the documentation historically maintained on the Slicer wiki at https://www.slicer.org/wiki/Documentation/Nightly/Modules/VectorToScalarVolume

### DIFF
--- a/Docs/user_guide/modules/vectortoscalarvolume.md
+++ b/Docs/user_guide/modules/vectortoscalarvolume.md
@@ -23,6 +23,10 @@ While vector volumes can be visualized in the slice viewers, many slicer operati
 |--|
 | Volume of JPG files from the [visible human project](https://www.nlm.nih.gov/research/visible/visible_human.html) converted to a scalar volume using the **Luminance** method. |
 
+## Related extensions and modules
+
+- The [SlicerMorph extension's ImageStacks module](https://github.com/SlicerMorph/SlicerMorph#module-descriptions) supports reading and other operations on volumetric vector (color) image formats.
+
 ## Contributors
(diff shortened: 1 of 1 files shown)


jcfr:
Notes related to the screenshot:
* uploaded to https://github.com/Slicer/Slicer/releases/tag/docs-resources
* rescaled to `800x600` (original size was `1193x972`)
* added using a table to associated the figure with its legend

Questions:
* Do we have guidelines regarding resolution of screeshots ?
* What is the best way for associated legend with figures when using markdown ? 
* Does it make sense to add the screenshot directly in the `Panels and their use` section or should it be added to a dedicated `Gallery` or `Example` section ... ?

cc: @lassoan

pieper (approved):
LGTM.  Added a link to ImageStacks.  Commit check failed due to merge but this should be resolved during squash.

lassoan:
> Do we have guidelines regarding resolution of screeshots ?

I would suggest to use similar size as the other screenshots.

> What is the best way for associated legend with figures when using markdown ?

All captioning is done in this PowerPoint file: https://github.com/Slicer/Slicer/releases/download/docs-resources/source_figures.pptx

After slides are added/modified:
- replace the pptx file in the release assets
- exported the slide to .png file, crop the unnecessary margins, and upload as release asset

> Does it make sense to add the screenshot directly in the Panels and their use section or should it be added to a dedicated Gallery or Example section ... ?

I think it makes sense to add it in the "Panels..." section. We could generate a gallery page that would list all the images.

PR #6637, opened 2022-11-01
ENH: Update VTK to allow setting vtkPoissonDiskSampler seed

jcfr:
List of VTK changes:

```
$ git shortlog 97a187572..5ae31a60d --no-merges
Pranjal Sahu (1):
      [Backport MR-9653] Add seed for reproducible sampling in vtkPoissonDiskSampler
```

Corresponding VTK change can be reviewed using this link:
https://github.com/Kitware/VTK/compare/97a187572...Slicer:VTK:slicer-v9.1.20220125-efbe2afc2?expand=1

### DIFF
--- a/SuperBuild/External_VTK.cmake
+++ b/SuperBuild/External_VTK.cmake
@@ -139,7 +139,7 @@ if((NOT DEFINED VTK_DIR OR NOT DEFINED VTK_SOURCE_DIR) AND NOT Slicer_USE_SYSTEM
 
   set(_git_tag)
   if("${Slicer_VTK_VERSION_MAJOR}" STREQUAL "9")
-    set(_git_tag "97a187572d4000cd820f9fc887f21eaf0bde857c") # slicer-v9.1.20220125-efbe2afc2
+    set(_git_tag "5ae31a60d3f6047fce1ff2bc8311e6975ce07ec2") # slicer-v9.1.20220125-efbe2afc2
     set(vtk_egg_info_version "9.1.20220125")
   else()
     message(FATAL_ERROR "error: Unsupported Slicer_VTK_VERSION_MAJOR: ${Slicer_VTK_VERSION_MAJOR}")


jcfr:
I confirmed locally that:
* VTK build :heavy_check_mark:
* GetSeed/SetSeed are available in the Slicer python console :heavy_check_mark: 

cc: @PranjalSahu

PranjalSahu:
Meshfile: https://drive.google.com/file/d/1Ew_upO7ECZLlJNCYYUf3bIIRvJNupnk9/view?usp=sharing

Code to test is reproducible points are sampled or not:
```python
import vtk
from vtk.util import numpy_support
import numpy as np

def readply(filename):
    a = vtk.vtkPLYReader()
    a.SetFileName(filename)
    a.Update()
    m1 = a.GetOutput()
    return m1

def samplePoints(seed, mesh, radius):
    f = vtk.vtkPoissonDiskSampler()
    f.SetInputData(mesh)
    f.SetRadius(radius)
    f.SetSeed(seed)
    f.Update()
    sampled_points = f.GetOutput()
    sampled_points_np = numpy_support.vtk_to_numpy(sampled_points.GetPoints().GetData())
    return sampled_points_np

inputMesh = readply("PWK.PhJ_.ply")
radius = 0.5

sampled_points1_np = samplePoints(seed=5, mesh=inputMesh, radius=radius)
sampled_points2_np = samplePoints(seed=5, mesh=inputMesh, radius=radius)
diff = sampled_points1_np - sampled_points2_np
assert np.sum(diff.flatten()) == 0

sampled_points3_np = samplePoints(seed=0, mesh=inputMesh, radius=radius)
sampled_points4_np = samplePoints(seed=0, mesh=inputMesh, radius=radius)
diff = sampled_points3_np - sampled_points4_np
assert np.sum(diff.flatten()) == 0
```

jcfr:
After downloading [bunny.ply](https://raw.githubusercontent.com/naucoin/VTKData/master/Data/bunny.ply) and updating the snippet accordingly

```diff
- inputMesh = readply("PWK.PhJ_.ply")
+ inputMesh = readply("/home/jcfr/Downloads/bunny.ply")
```

I am able to run the snippet :100: :rocket:

jcfr:
That said, if I run the same snippet without setting the seed, it also completes without error:

```diff
    f.SetSeed(seed)
    #f.SetSeed(seed)
```

@PranjalSahu Should I attempt to run the filter on specific dataset to also reproduce the failure case ?

PranjalSahu:
The default seed is 0. If you run the code with the previous version then it should give a different result every time.

jcfr:
After realizing I was not using the `PWK.PhJ_.ply`, I tested further and here are few observations:
* Without the patch, the snippet does not fail
* With the patch and by explicitly setting different seed, it is failing

Based on this, it looks like the filter allows to reproduce the same result. 

Also worth noting that by default, the `default_random_engine` is consistently seeded with `1` [^1][^2]

[^1]: https://cplusplus.com/reference/random/default_random_engine/
[^2]: https://cplusplus.com/reference/random/linear_congruential_engine/linear_congruential_engine/

PranjalSahu:
Not able to reproduce the issue. The method is returning the same sub-sampled points on each call. 
This PR could be added later if the user needs multiple sub-sampling results (by changing the seed).

jcfr:
For future reference, corresponding `Slicer/VTK` branch has been published as https://github.com/Slicer/VTK/tree/slicer-v9.1.20220125-efbe2afc2-slicer-pr6637

PR #6642, opened 2022-11-03
ENH: Make viewer font file configurable

lassoan:
VTK's default font file does not support Chinese characters. This commit allows setting a custom font file (such as Google's Noto font - https://fonts.google.com/noto/fonts?noto.continent=Asia&noto.region=CN&noto.script=Hant) in application settings. Since these font files are relatively large (about 5MB for each font family) and it is not clear yet which formats will work the best, the fonts are not yet bundled with the application but for now it will be installed by LanguagePacks extension.

image


pieper (approved):
LGTM 👍

PR #6644, opened 2022-11-03
STYLE: Marking translatable strings in modules

mhdiop:


Modules/Loadable/Data/qSlicerDataModule.cxx
```diff
@@ -71,15 +71,15 @@ QIcon qSlicerDataModule::icon()const
...
+  return QStringList() << "" << tr("Informatics");
 }
 
 //-----------------------------------------------------------------------------
 QStringList qSlicerDataModule::dependencies() const
 {
   QStringList moduleDependencies;
   // Cameras: Required in qSlicerSceneReader
-  moduleDependencies << "Cameras";
+  moduleDependencies << tr("Cameras");
```
lassoan, line 82: This refers to the module name (that is not translated), not to the module title (that is translated). To make it clear that this text must not be translated, it would be safer to mark it as non-translatable. I tried to search for convention on how to make text as non-translatable in C/C++ source code, but I did not find anything. Therefore, we should come up with our own convention.

What do you think about marking non-translatable text using a macro or comment?

Option A: Using comment.

```
someFunction("some string"); // I18N:NO_TR - optional comment describing why not to translate
```

Option B: defining a no-op macro

Define this macro somewhere (in CTK?):
```
#define NO_TR(x) x
```

Usage:

```
someFunction(NO_TR("some string")); // optional comment describing why not to translate
```
abc(NO_TR("some string"));

@pieper @jcfr @cpinter

pieper: Thanks for looking for a convention for this.  I'm kind of surprised since I would have thought this would be fairly common.  I was kind of thinking the convention would be `no_tr()` so that it's visually clear that it's related to `tr()`

lassoan: More similarity to `tr()` would be nice, but `tr()` is a function. Would you make `no_tr()` a function, too?

Qt uses macros for some translation features (QT_TR_...), so using all-caps macros would not be unprecedented.

pieper: Yes, in general I would want all macros to be in all caps, but this might be an exception I would be okay with.  On the other hand making `no_tr` a method instead of a macro would be somehow more consistent and might have some value in the future.  I don't think it would have any performance impact since using strings as identifiers is already an option you would only use when performance isn't critical.  

Good point about the name `noTr` but in this case `tr` doesn't follow Qt convention anyway since it would normally be `translatable` or similar.

Maybe instead we should have a macro like `TOKEN` or `CTK_TOKEN` to be even more explicit about the intent of the string so it never get a `tr` added.

In any case if you still like `NO_TR` that's also fine.

lassoan: I don't find any of these names much better than the others. Let's see if @jcfr and @cpinter have any preferences.

mhdiop: Thanks @lassoan and @pieper for your feedbacks. The translation function will be removed from the string in question in the next commit.

Thanks also for the interesting discussion you have about the convention we should use for marking non-translatable strings. 

I think that the first option (using comments) is easier to integrate on existing codes. However, the second one has the advantage to better highlight non-translatable strings since developers may not really consider comments when reading source codes, unless they find it necessary to do so. 

While you're deciding on the best approach to use, I'll keep marking translatable strings on other modules. The necessary work will be done for the non-translatable ones whenever a marking method is retained.

Modules/Loadable/Data/qSlicerDataModule.cxx
```diff
@@ -91,7 +91,7 @@ void qSlicerDataModule::setup()
   this->Superclass::setup();
 
   vtkSlicerCamerasModuleLogic* camerasLogic =
-    vtkSlicerCamerasModuleLogic::SafeDownCast(this->moduleLogic("Cameras"));
+    vtkSlicerCamerasModuleLogic::SafeDownCast(this->moduleLogic(tr("Cameras")));
```
mhdiop, line 94: By the way, is really this string a translatable one ?

mhdiop: I also have the same doubts when it's about `On` and `Off` strings present on `Modules/Loadable/Data/Logic/vtkSlicerDataModuleLogic.cxx:58`. 

```c++
void vtkSlicerDataModuleLogic::PrintSelf(ostream& os, vtkIndent indent)
{
  this->Superclass::PrintSelf(os, indent);

  os << indent << "AutoRemoveDisplayAndStorageNode: " <<
    (AutoRemoveDisplayAndStorageNodes ? "On" : "Off") << "\n";

 }
```

lassoan: > By the way, is really this string a translatable one ?

It is the module name, so it should not be translated. I just first wanted to find a way how to mark these non-obvious cases in the code to ensure that they don't get translated anywhere.

> doubts when it's about On and Off strings present on

We only translate interface for users. We keep developer communication, such as logs, debugging outputs (such as PrintSelf), non-user-provided file contents (such as fields names and tokens in MRML scene files) the same, regardless of the current language.

Please add all these information (don't translate module name, file content, file extension, developer communication, such as log messages, PrintSelf output) to the [developer manual] (https://github.com/Slicer/SlicerLanguagePacks/blob/main/DevelopersManual.md).

Modules/Loadable/Data/qSlicerDataModule.cxx
lassoan: To note hold up your work with this discussion, let's just use a comment now:

    someFunction(/*no tr*/"some string");

Once we have a decision, we can replace all these comments by the final solution.

Modules/Core/EventBroker/qSlicerEventBrokerModule.cxx
```diff
@@ -53,23 +53,23 @@ vtkMRMLAbstractLogic* qSlicerEventBrokerModule::createLogic()
 //-----------------------------------------------------------------------------
 QStringList qSlicerEventBrokerModule::categories()const
 {
-  return QStringList() << "Developer Tools";
+  return QStringList() << tr("Developer Tools");
```
lassoan, line 56: This would add a “Developer Tools" translatable string under the qSlicerEventBrokerModule context. It means that this category name could be translated differently in each module. Weblate will offer to use the same string by default everywhere, but still, when changes are made then it could easily lead to inconsistencies.

What do you think about using some base class, such as `qSlicerAbstractCoreModule` for all module category name translations?

    qSlicerAbstractCoreModule::tr("Developer Tools");

Modules/Loadable/Data/qSlicerDataModule.cxx
mhdiop: Got it, thanks ! 

The right updates will be done on the developer manual.

For modules, there are some folders that contains codes which should not have translatable strings. We can thereby exclude them from the marking process. It's currently the case of `testing` folders. Can I do the same with modules' `logic` folders without risking to miss out any translatable string ?

I think, as a side note, I can add the list of those folders to the developer manual too so that to precise that they should not contain translatable strings.

Modules/Core/EventBroker/qSlicerEventBrokerModule.cxx
mhdiop: That's a good point, thank you 👍

Modules/Loadable/Data/qSlicerDataModule.cxx
mhdiop: Alright, I'm fine with that. Thanks.

Modules/Loadable/Data/qSlicerDataModule.cxx
lassoan: > For modules, there are some folders that contains codes which should not have translatable strings. We can thereby exclude them from the marking process. It's currently the case of testing folders.

Yes, you can exclude testing folders. Please add to the developer manual the lupdate command-line arguments that you use to extract/update the strings.

> Can I do the same with modules' logic folders without risking to miss out any translatable string ?

Translation of strings coming from logic classes is a complex topic. First we need to make the decision if we allow logic classes to generate any user-visible strings and if yes then where they need to be translated (in logic classes or in GUI classes). You can ignore Logic, MRML, MRMLDM, VTKWidgets folders from your analysis, and I've added an issue (https://github.com/Slicer/Slicer/issues/6647) to remember to deal with them later.

mhdiop: Alright, that's good then. Thanks !

lassoan (commented):
I've added comments at the changes that required modifications, all the others look good to me. Please send follow-up commits to this branch and let us know when you are done and the code is ready for review.

mhdiop:
The code is now updated based on your feedbacks. Commits has also been squashed to a single one.

Thank you.

Modules/Loadable/Data/qSlicerDataModule.cxx
cpinter: Sorry just seeing this. I think `no_tr` would be most consistent with what we see in other places if it's a function, and `NO_TR` if it's a macro. I'm also fine with the comment approach as I see it in the commit right now, as it's pretty obvious (I don't see it easily ignorable since it's in-line, just before the string).

lassoan (approved):
Thank you, all these changes look good to me.

PR #6646, opened 2022-11-04
ENH: Add ability to quickly open to log file location

hannah-morilak:
## Motivation and Changes
The main motivation behind this is that adding the ability to open file log locations makes it far easier for slicer users to send entire log files to the developers that make various slicer apps/extensions/modules. This PR updates the 'report a bug' dialog by replacing the `Copy log message to clipboard` button with a `Open log file location` button. I could have opted to add another button, however for the sake of space I decided to simply replace the former button. 

### Before this PR
image

### After this PR
image

### DIFF
--- a/Base/QTApp/Resources/UI/qSlicerErrorReportDialog.ui
+++ b/Base/QTApp/Resources/UI/qSlicerErrorReportDialog.ui
@@ -70,7 +70,7 @@
      <item>
       <widget class="ctkPushButton" name="LogCopyToClipboardPushButton">
        <property name="text">
-        <string>Copy log messages to clipboard</string>
+        <string>Copy log to clipboard</string>
        </property>
        <property name="autoDefault">
         <bool>false</bool>
@@ -80,7 +80,17 @@
      <item>
       <widget class="ctkPushButton" name="LogFileOpenPushButton">
        <property name="text">
-        <string>Open log file in editor</string>
+        <string>Open log in editor</string>
+       </property>
+       <property name="autoDefault">
+        <bool>false</bool>
+       </property>
+      </widget>
+     </item>
+     <item>
+      <widget class="ctkPushButton" name="LogOpenFileLocationPushButton">
+       <property name="text">
+        <string>Open log location</string>
        </property>
        <property name="autoDefault">
         <bool>false</bool>
--- a/Base/QTApp/qSlicerErrorReportDialog.h
+++ b/Base/QTApp/qSlicerErrorReportDialog.h
@@ -43,6 +43,7 @@ public:
 
 protected slots:
   void onLogFileOpen();
+  void onLogFileLocationOpen();
   void onLogCopy();
   void onLogFileSelectionChanged();
   void onLogFileEditClicked(bool editable);
--- a/Base/QTApp/qSlicerErrorReportDialog.cxx
+++ b/Base/QTApp/qSlicerErrorReportDialog.cxx
@@ -24,6 +24,7 @@
 #include <QFile>
 #include <QTextStream>
 #include <QUrl>
+#include <QFileInfo>
 
 // SlicerApp includes
 #include "qSlicerErrorReportDialog.h"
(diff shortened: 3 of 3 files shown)


pieper:
I guess I don't have a strong opinion on this.  Both functions seem useful, but I can see the desire to reduce clutter.

lassoan:
I don't see why we would need to remove the Copy to clipboard feature. We can replace button text by icon (or add icon in addition to the text) to make things a bit prettier or easier to find the right button.

lassoan:
The motivation for not promoting blindly sending log files is that users should review the log file content and remove any protected health information that might appear there. That is why the log is in an editor - it makes it easy to sanitize and select the relevant part of the log.

Base/QTApp/qSlicerErrorReportDialog.h
```diff
@@ -43,6 +43,7 @@ class Q_SLICER_BASE_QTAPP_EXPORT qSlicerErrorReportDialog :
 
 protected slots:
   void onLogFileOpen();
+  void onLogFileLocationOpen();
   void onLogCopy();
```
jamesobutler, line 47: The Build Slicer GitHub action reveals that `onLogCopy` is still present here while in the cxx file you had removed it. From @lassoan latest comment, you should probably keep the button associated with `onLogCopy` rather than replacing it.

jamesobutler:
> The motivation for not promoting blindly sending log files is that users should review the log file content and remove any protected health information that might appear there. 

I guess for many Slicer users out there they are only dealing with animal data and not patient data so there isn’t concern with revealing protected info. Considering there is a warning text describing this in the error dialog already is that enough? @lassoan should this warning be more prevalent with different text color?

lassoan:
> I guess for many Slicer users out there they are only dealing with animal data and not patient data so there isn’t concern with revealing protected info. Considering there is a warning text about sanitizing patient info in the error dialog already is that enough? @lassoan should this warning be more prevalent with different text color?

It is already bold. There is a lot of text, users would probably not read the warning even if it is a different color. Instead maybe we could display a popup (with a don't show again option) when the user clicks "Open log file location" button to remind users that the log files may contain protected health information, so don't share them without viewing.

@hannah-morilak To allow both the current "send after in-place review and sanitization" and the new "send entire log file blindly" use cases, please don't remove any existing buttons. Instead, you can shorten the text on the buttons to make it a bit more user friendly (e.g., Copy to clipboard, Open in external editor, Open log file location).

hannah-morilak:
~~This PR updates the 'report a bug' dialog by replacing the `Copy log message to clipboard` button with a `Open log file location` button. I could have opted to add another button, however for the sake of space I decided to simply replace the former button.~~ 
**Edit**: The `Open log file location` button was added in addition to the two pre-existing buttons based on PR comments.

### Updated
image

Base/QTApp/qSlicerErrorReportDialog.h
hannah-morilak: Both buttons were kept, so all methods are still in the `h` file.

Base/QTApp/Resources/UI/qSlicerErrorReportDialog.ui
```diff
@@ -80,7 +80,17 @@
...
+       </property>
+       <property name="autoDefault">
+        <bool>false</bool>
+       </property>
+      </widget>
+     </item>
+     <item>
+      <widget class="ctkPushButton" name="LogOpenFileLocationPushButton">
+       <property name="text">
+        <string>Open log file location</string>
```
jamesobutler, line 93: Since in the other buttons you have dropped "file" such as `Open log file in editor` -> `Open log in editor`, then maybe best to keep the same convention here so:
```suggestion
        <string>Open log location</string>
```

hannah-morilak: Updated

lassoan:
It looks good to me. I'll merge as soon as the commit checks are completed.

jamesobutler:
The commit checks have completed with success. Thank you @hannah-morilak for your first contribution to Slicer 🙏🏻

PR #6650, opened 2022-11-07
BUG: CopyContent should call Modified

Connor-Bowley:
vtkMRMLScriptedModuleNode copies parameters from the other node not through the SetParameter function, so it needs to properly call Modified if anything changes.

### DIFF
--- a/Libs/MRML/Core/vtkMRMLScriptedModuleNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLScriptedModuleNode.cxx
@@ -109,7 +109,11 @@ void vtkMRMLScriptedModuleNode::CopyContent(vtkMRMLNode* anode, bool deepCopy/*=
     }
 
   this->SetModuleName(node->GetModuleName());
-  this->Parameters = node->Parameters;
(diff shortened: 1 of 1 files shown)


sjh26 (dismissed):
lgtm

Libs/MRML/Core/vtkMRMLScriptedModuleNode.cxx
```diff
@@ -109,7 +109,11 @@ void vtkMRMLScriptedModuleNode::CopyContent(vtkMRMLNode* anode, bool deepCopy/*=
     }
 
   this->SetModuleName(node->GetModuleName());
-  this->Parameters = node->Parameters;
+  if (this->Parameters != node->Parameters)
+  {
```
lassoan, line 113: Indentation does not follow the current coding conventions (we will switch to a more common indentation style all at once, but we keep the current one for now, for consistency)

Otherwise it looks good to me, too.

Connor-Bowley: Fixed indentation.

sjh26 (approved):

PR #6654, opened 2022-11-11
BUG: Fix fixedNumberOfControlPoints property value in exported markup nodes

lassoan:
The FixedNumberOfControlPoints flag was not copied and at node export a copy of the node is exported.

Also made FixedNumberOfControlPoints and MaximumNumberOfControlPoints properties completely independent. Modifying MaximumNumberOfControlPoints when FixedNumberOfControlPoints flag was changed was unnecessary and had very complicated side effects.

Fixes error reported at https://discourse.slicer.org/t/export-file-as-bug-for-markup-control-points/26188

### DIFF
--- a/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsNode.h
+++ b/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsNode.h
@@ -632,7 +632,7 @@ public:
   vtkGetMacro(RequiredNumberOfControlPoints, int);
 
   /// Maximum number of control points limits the number of markups allowed in the node.
-  /// If maximum number of control points is set to 0 then no it means there
+  /// If maximum number of control points is set to -1 then no it means there
   /// is no limit (this is the default value).
   /// The value is an indication to the user interface and does not affect
   /// prevent adding markups to a node programmatically.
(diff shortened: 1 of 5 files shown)


lassoan:
It seems to be important for @muratmaga to include this in the stable release (so they don't need to wait for this fix for 3 months). I've tested the changes manually and it seems to be all good, related automatic tests pass, so I think we should include this in 5.2.

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsNode.h
```diff
@@ -632,7 +632,7 @@ class  VTK_SLICER_MARKUPS_MODULE_MRML_EXPORT vtkMRMLMarkupsNode : public vtkMRML
   vtkGetMacro(RequiredNumberOfControlPoints, int);
 
   /// Maximum number of control points limits the number of markups allowed in the node.
-  /// If maximum number of control points is set to 0 then no it means there
+  /// If maximum number of control points is set to -1 then no it means there
```
jcfr, line 635:
```suggestion
  /// If maximum number of control points is set to -1 then it means there
```

PR #6657, opened 2022-11-12
BUG: Fix control point state not updated with hidden coordinates

jamesobutler:
Originally mentioned in https://discourse.slicer.org/t/unable-to-clear-all-control-points-at-once-goes-by-threes/26203.

An incorrect selected item increment was using the number of total columns to skip when finding the next selected row and this was incorrect when columns were hidden. Selected Items only include the visible selections so each row would only have 6 selected items when coordinates were hidden instead of 9 items when coordinates were shown.

### DIFF
--- a/Modules/Loadable/Markups/qSlicerMarkupsModuleWidget.cxx
+++ b/Modules/Loadable/Markups/qSlicerMarkupsModuleWidget.cxx
@@ -1653,11 +1653,11 @@ void qSlicerMarkupsModuleWidget::onDeleteControlPointPushButtonClicked(bool conf
   // selected indices for each column in a row, so jump by the number of
   // columns), so can delete without relying on the table
   QList<int> rows;
-  for (int i = 0; i < selectedItems.size(); i += d->numberOfColumns())
+  QModelIndexList selectedIndexes = d->activeMarkupTableWidget->selectionModel()->selectedRows();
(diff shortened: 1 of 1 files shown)


Modules/Loadable/Markups/qSlicerMarkupsModuleWidget.cxx
```diff
@@ -1778,7 +1780,7 @@ void qSlicerMarkupsModuleWidget::onRestoreControlPointPushButtonClicked()
     }
 }
 //-----------------------------------------------------------------------------
-void qSlicerMarkupsModuleWidget::onUnsetControlPointPushButtonClicked()
+void qSlicerMarkupsModuleWidget::onUnsetControlPointPushButtonClicked()  /// HERE
```
lassoan, line 1783:
```suggestion
void qSlicerMarkupsModuleWidget::onUnsetControlPointPushButtonClicked()
```

lassoan (approved):
This looks good to me. Thanks a lot for fixing this so quickly.

Modules/Loadable/Markups/qSlicerMarkupsModuleWidget.cxx
jamesobutler: Thanks for cleaning up my mental note comment. I forgot I left that there

PR #6672, opened 2022-11-16
Fix doc generation warnings and errors

jcfr:


lassoan:
This is great, thank you!

PR #6675, opened 2022-11-16
ENH: Support disconnecting markups toolbar from mouse mode toolbar

jamesobutler:
This change allows for calling to removeAction(ToolBarAction) to remove the markups toolbar connection and have it remain disconnected. Previously it would always add the action back, overwriting the customization. Now the action is always in the set of mouse mode actions, but selectively shown/hidden

This doesn't introduce a change in functionality of how things are currently. It changes from removing/adding actions to showing/hiding actions in the Mouse mode toolbar. In the case of my custom app where we don't use the markups toolbar (reducing functionality on purpose), this PR allows for me to remove the toggle markups toolbar action from the Mouse Mode toolbar and have the place widget action still show/hide, but not reshow the toggle markups button in my custom app.

### DIFF
--- a/Base/QTGUI/qSlicerMouseModeToolBar.cxx
+++ b/Base/QTGUI/qSlicerMouseModeToolBar.cxx
@@ -169,6 +169,8 @@ void qSlicerMouseModeToolBarPrivate::init()
 
   connect(this->PlaceWidgetAction, SIGNAL(triggered()), q, SLOT(switchPlaceMode()));
   this->InteractionModesActionGroup->addAction(this->PlaceWidgetAction);
+  this->PlaceWidgetAction->setVisible(false);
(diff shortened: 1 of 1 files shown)


jamesobutler:
@lassoan Do you see any issue with switching to show/hide logic here instead of add/remove action?

jamesobutler:
@lassoan Any thoughts on this minor change? I'd like to finalize this PR soon if possible.

jamesobutler:
@lassoan Any comments here? Or can this be integrated?

lassoan (approved):
This looks good to me. In general, it is more appropriate to change visibility of actions (as it is proposed in this PR) than add/remove action, because add/remove is unnecessarily strong operation with potential side effects (it may change order of actions, etc.).

PR #6677, opened 2022-11-16
DOC: Improve Module Overview highlighting "rational" and "getting started"

jcfr:
_Related to #6189_

* Update "rational" and "getting started" wording and use sphinx directives to highlight them.
    
* Remove ambiguous description: Considering the purpose of implementing a loadable module (either C++ or scripted) is to have a custom and tailored user experience that can not be obtained using the automatically generated UI associated with CLI modules, this commit removes the sentence.
    
* Remove link to obsolete "Loadable module" tutorial last updated in August 2016.

### DIFF
--- a/Docs/developer_guide/module_overview.md
+++ b/Docs/developer_guide/module_overview.md
@@ -5,22 +5,21 @@ Slicer supports multiple types of modules:
 * [Loadable](#loadable-modules)
 * [Scripted](#scripted-modules)
 
-While the developer has to choose between one of the types to implement its module, the end user will NOT notice the difference as they all share the same look & feel.
-
 The choice for a given type of module is usually based on the type of inputs/parameters for a given module.
-
-The [Extension Wizard](user_guide/modules/extensionwizard.md#extension-wizard) module provides a graphical interface within Slicer to aid in the creation of Slicer extensions and associate modules.
 
 ## Command Line Interface (CLI)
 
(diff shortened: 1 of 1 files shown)


Docs/developer_guide/module_overview.md
```diff
@@ -5,22 +5,23 @@ Slicer supports multiple types of modules:
 * [Loadable](#loadable-modules)
 * [Scripted](#scripted-modules)
 
-While the developer has to choose between one of the types to implement its module, the end user will NOT notice the difference as they all share the same look & feel.
+While the developer has to choose between one of the types to implement its module, the end user will **NOT** notice the difference as they all share the same look & feel.
```
jamesobutler, line 8: > as they all share the same look & feel.

I would argue that CLI modules have a different look & feel compared to Loadable and Scripted.

jcfr: Since we now have the rational highlighted (See below), what about just removing this sentence ?


|  | Rational |
|--|--|
| CLI | image |
| Loadable| image |
| Scripted | image |

jcfr: And your comment makes sense, especially considering the purpose of implementing a `loadable` module (either C++ or scripted) is to have custom and tailored user experience that can not be obtained using the automatically generated UI.

jamesobutler: Yes I think this sentence can be removed now that there are these rationales about when to use whichever one

PR #6682, opened 2022-11-17
ENH: Use column name for plot title when none specified

### LINKED ISSUE #6671: When only `columnNames` parameter is specified (no `title`), names are not used for the series during plotting a histogram
## Summary

I would expect that if at least columnNames are provided (in [this function](https://github.com/Slicer/Slicer/blob/main/Base/Python/slicer/util.py#L3268)), the column names will be used for the produced series -- and thus displayed in the legend. Instead, I have to also add a title parameter, and then the name for the series looks like that:

```
 seriesNode.SetName(title + " " + yColumnName)
```

See: https://github.com/Slicer/Slicer/blob/main/Base/Python/slicer/util.py#L3389

## Steps to reproduce

```
# Generate example input data (volumeNode, segmentationNode, segmentId)
################################################

# Load source volume
import SampleData
volumeNode = sampleDataLogic.downloadMRBrainTumor1()

# Create segmentation
segmentationNode = slicer.vtkMRMLSegmentationNode()
slicer.mrmlScene.AddNode(segmentationNode)
segmentationNode.CreateDefaultDisplayNodes() # only needed for display
segmentationNode.SetReferenceImageGeometryParameterFromVolumeNode(volumeNode)

# Create segment
tumorSeed = vtk.vtkSphereSource()
tumorSeed.SetCenter(-6, 30, 28)
tumorSeed.SetRadius(25)
tumorSeed.Update()
segmentId = segmentationNode.AddSegmentFromClosedSurfaceRepresentation(tumorSeed.GetOutput(), "Segment A", [1.0,0.0,0.0])

# Compute histogram
################################################

# Get voxel values of volume in the segmented region
import numpy as np
volumeArray = slicer.util.arrayFromVolume(volumeNode)
segmentArray = slicer.util.arrayFromSegmentBinaryLabelmap(segmentationNode, segmentId, volumeNode)
segmentVoxels = volumeArray[segmentArray != 0]

# Compute histogram
import numpy as np
histogram = np.histogram(segmentVoxels, bins=50)

# Plot histogram
################################################

slicer.util.plot(histogram, xColumnIndex = 1, columnNames = ['Column no 1', 'Column no 2'])
```

I slightly modified this example: https://slicer.readthedocs.io/en/latest/developer_guide/script_repository.html#get-histogram-of-a-segmented-region

Other references:

https://discourse.slicer.org/t/histogram-of-volume-in-regions-defined-by-segments/2236/9

https://slicer.readthedocs.io/en/latest/developer_guide/slicer.html#slicer.util.plot

## Expected behavior

When I use the code above, I expect Series name to use the name of the column.

## Suggested solution

In [this line ](https://github.com/Slicer/Slicer/blob/main/Base/Python/slicer/util.py#L3388) add a clause for the situation when there is no title specified, yet there is a column name, e.g.:

```
        if title:
            seriesNode.SetName(title + " " + yColumnName)
        else:
             if (columnNames is not None) and (len(columnNames) > columnIndex):
                columnName = columnNames[columnIndex]
                seriesNode.SetName(columnName)
```

Although this should work (needs some testing), it also may look like breaking a 'don't repeat yourself' rule, as a similar case is already present for the table: https://github.com/Slicer/Slicer/blob/main/Base/Python/slicer/util.py#L3345-L3347 . With that being said, I bet my suggested solution can be better refactored.

## Environment
- Slicer version: Slicer-5.0.3-2022-07-08 `r30893 / 7ea0f43`
- Operating system: Mac 11.6.6

mikbuch:
Without this fix there can be situation in which when `columnNames` are provided, but no `title`, the name of the created series is not affected. I would expect that column names would be used as `seriesNode` names.

I tested this fix in my Slicer installation and it worked.

Fixes #6671

### DIFF
--- a/Base/Python/slicer/util.py
+++ b/Base/Python/slicer/util.py
@@ -3405,6 +3405,10 @@ def plot(narray, xColumnIndex=-1, columnNames=None, title=None, show=True, nodes
         seriesNode.SetYColumnName(yColumnName)
         if title:
             seriesNode.SetName(title + " " + yColumnName)
+        else:
+            # When only columnNames are set (no `title` parameter), the user should see
+            # the name of the column as the name of the series.
+            seriesNode.SetName(yColumnName)
         if not chartNode.HasPlotSeriesNodeID(seriesNode.GetID()):
             chartNode.AddAndObservePlotSeriesNodeID(seriesNode.GetID())
 


mikbuch:
Using the example code:

```
# Generate example input data (volumeNode, segmentationNode, segmentId)
################################################

# Load source volume
import SampleData
volumeNode = sampleDataLogic.downloadMRBrainTumor1()

# Create segmentation
segmentationNode = slicer.vtkMRMLSegmentationNode()
slicer.mrmlScene.AddNode(segmentationNode)
segmentationNode.CreateDefaultDisplayNodes() # only needed for display
segmentationNode.SetReferenceImageGeometryParameterFromVolumeNode(volumeNode)

# Create segment
tumorSeed = vtk.vtkSphereSource()
tumorSeed.SetCenter(-6, 30, 28)
tumorSeed.SetRadius(25)
tumorSeed.Update()
segmentId = segmentationNode.AddSegmentFromClosedSurfaceRepresentation(tumorSeed.GetOutput(), "Segment A", [1.0,0.0,0.0])

# Compute histogram
################################################

# Get voxel values of volume in the segmented region
import numpy as np
volumeArray = slicer.util.arrayFromVolume(volumeNode)
segmentArray = slicer.util.arrayFromSegmentBinaryLabelmap(segmentationNode, segmentId, volumeNode)
segmentVoxels = volumeArray[segmentArray != 0]

# Compute histogram
import numpy as np
histogram = np.histogram(segmentVoxels, bins=50)

# Plot histogram
################################################

slicer.util.plot(histogram, xColumnIndex = 1, columnNames = ['Column no 1', 'Column no 2'])
```

now produces the expected outcome:
image

lassoan:
Thank you for the contribution! It seems that there are many unrelated changes in this pull request. Could you please rebase your branch on the latest Slicer main version?

jamesobutler:
It is the targeting of `nightly-main` branch that is the issue. It should be retargeted to `main`.

image

lassoan:
Good catch @jamesobutler. I was able to change the target branch and the diff looks good now.

Base/Python/slicer/util.py
```diff
@@ -3405,6 +3405,12 @@ def plot(narray, xColumnIndex=-1, columnNames=None, title=None, show=True, nodes
         seriesNode.SetYColumnName(yColumnName)
         if title:
             seriesNode.SetName(title + " " + yColumnName)
+        else:
+             # When only columnNames are set (no `title` parameter), the user should see
+             # the name of the column as the name of the series.
+             if (columnNames is not None) and (len(columnNames) > columnIndex):
+                 columnName = columnNames[columnIndex]
+                 seriesNode.SetName(columnName)
```
lassoan, line 3413: Could we simply use `yColumnName` here?

jamesobutler, line 3413:
```suggestion
        else:
            # When only columnNames are set (no `title` parameter), the user should see
            # the name of the column as the name of the series.
            seriesNode.SetName(yColumnName)
```

jcfr:
| Before | After |
|--|--|
| image | image |

mikbuch:
Hi guys, I am not an expert in these automated checks. Should I add some more commits, or correct something in order for this merge to pass? The last automatic check I believe failed because of the wrong commit message (the title).

jcfr:
@mikbuch If the update pull request looks good to you, we will move forward with the integration. Let us know what you think.

Issue #6665, opened 2022-11-14
Typed Parameter Nodes in Python


Connor-Bowley:
## The problem

In Python, it is recommended to use the parameter node concept, as the parameter nodes are in the scene and are automatically handled when saving and loading the scene. One large downside to the parameter node, in my opinion, is that they are essentially just a string to string dictionary. This has two large downsides:

- There is a loss of value typing. The user needs to know the type and cast to it at each usage.
- Compared to named access of a member variable or function, the errors from using the wrong key are harder to track down (especially because there is an implicit default of return empty string in `GetParameter`). It is also more difficult to discover possible parameter keys in the Python interpreter than it is to discover members.

## A possible solution

A solution I have implemented for a custom Slicer application is to create a wrapper around the parameter node that has "typed" member variables. A similar solution was implemented independently for the NousNav project (developed by @allemangD). This involves using Python properties on a custom non-MRML class to allow for pythonic access to a contained parameter node.

```py
# instead of
x = int(untypedParameterNode.GetParameter("x"))
inputNode = untypedParameterNode.GetNodeReference("inputNode")

# it would be
x = typedParameterNode.x # which returns an int
inputNode = typedParameter.inputNode # which returns a vtkMRMLNode of the appropriate type
```

An ideal solution would allow for easy to create wrappers, using some syntax such as the following:

```py
CustomTypedParameterNode = MakeTypedParameterNode(
"CustomTypedParameterNode", [
    Parameter("x", type=int, default=0),
    Parameter("values", type=float, multi=True),
    Parameter("inputNode", type=slicer.vtkMRMLModelNode),
])

parameterNode = CustomTypedParameterNode(self.logic.getParameterNode())
parameterNode.x = 4
parameterNode.values = [1.0, 1.5, 2.0]
parameterNode.inputNode = slicer.mrmlScene.AddNewNodeByClass('vtkMRMLModelNode')
```

This proposed syntax for creating the new type is inspired by the syntax for Python's [collections.namedtuple](https://docs.python.org/3/library/collections.html#collections.namedtuple) and [argparse](https://docs.python.org/3/library/argparse.html).

The generated Python properties (such as `x`, `values`, and `inputNode` above) would use `GetParameter/SetParameter/GetNodeReference/SetNodeReferenceID` under the hood, so as to reuse existing code for saving and loading the item to the scene.

## Describe alternatives you've considered

[...]

PR #6684, opened 2022-11-18
Parameter node wrapper

Connor-Bowley:
Adds a Python parameter node wrapper that offers access to typed properties instead of going through strings and manually serializing/deserializing to string manually.

Adds tests for the wrapper.

Updates VectorToScalarVolume to use new parameter node as an example of use (the tests also serve as example).

I have added a `parameter_nodes.md` file documenting purpose, usage, and extensibility, but would like recommendations on where would be good to link to it from.

Closes https://github.com/Slicer/Slicer/issues/6665

### DIFF
--- a/Base/Python/slicer/tests/test_slicer_parameter_node_wrapper.py
+++ b/Base/Python/slicer/tests/test_slicer_parameter_node_wrapper.py
@@ -1,0 +1,749 @@
+import dataclasses
+import pathlib
+from typing import Annotated
+import unittest
+
+import vtk
+import slicer
+
+from MRMLCorePython import (
+    vtkMRMLNode,
+    vtkMRMLModelNode,
+    vtkMRMLScalarVolumeNode,
+)
+from slicer.parameterNodeWrapper import *
+
+
+def newParameterNode():
+    node = slicer.mrmlScene.AddNewNodeByClass("vtkMRMLScriptedModuleNode")
+    return node
+
+
+@dataclasses.dataclass
+class CustomClass:
+    x: int
+    y: int
+    z: int
+
+
+@parameterNodeSerializer
+class CustomClassSerializer(Serializer):
+    @staticmethod
+    def canSerialize(type_) -> bool:
+        return type_ == CustomClass
+
+    @staticmethod
+    def create(type_):
+        if CustomClassSerializer.canSerialize(type_):
+            return ValidatedSerializer(CustomClassSerializer(), [NotNone(), IsInstance(CustomClass)])
+        return None
+
+    def default(self):
+        return None
+
+    def isIn(self, parameterNode, name: str) -> bool:
+        return parameterNode.HasParameter(name)
+
+    def write(self, parameterNode, name: str, value) -> None:
+        parameterNode.SetParameter(name, f"{value.x},{value.y},{value.z}")
+
+    def read(self, parameterNode, name: str):
+        val = parameterNode.GetParameter(name)
+        vals = val.split(',')
+        return CustomClass(int(vals[0]), int(vals[1]), int(vals[2]))
+
+    def remove(self, parameterNode, name: str) -> None:
+        parameterNode.UnsetParameter(name)
+
+
+@dataclasses.dataclass
+class AnotherCustomClass:
+    a: int
+    b: int
+    c: int
+
+
+# the big difference between this and CustomClassSerializer, is that the
+# default for this is not None
+@parameterNodeSerializer
+class AnotherCustomClassSerializer(Serializer):
+    @staticmethod
+    def canSerialize(type_) -> bool:
+        return type_ == AnotherCustomClass
+
+    @staticmethod
+    def create(type_):
+        if AnotherCustomClassSerializer.canSerialize(type_):
+            return ValidatedSerializer(AnotherCustomClassSerializer(), [IsInstance(AnotherCustomClass)])
+        return None
+
+    def default(self):
+        return AnotherCustomClass(0, 0, 0)
+
+    def isIn(self, parameterNode, name: str) -> bool:
+        return parameterNode.HasParameter(name)
+
+    def write(self, parameterNode, name: str, value) -> None:
+        parameterNode.SetParameter(name, f"{value.a},{value.b},{value.c}" if value is not None else "None")
+
+    def read(self, parameterNode, name: str):
+        val = parameterNode.GetParameter(name)
+        if val == "None":
+            return None
+        else:
+            vals = val.split(',')
+            return AnotherCustomClass(int(vals[0]), int(vals[1]), int(vals[2]))
+
+    def remove(self, parameterNode, name: str) -> None:
+        parameterNode.UnsetParameter(name)
+
+
+# not registering as a @parameterNodeSerializer because we are overriding
+# the serialization of a built in type and we don't want anyone else to accidentally
+# use this.
+class CustomIntSerializer(Serializer):
+    """
+    Stores ints as their negative.
+    Used to test being able to use a specific serializer.
+    """
+    @staticmethod
+    def canSerialize(type_) -> bool:
+        return type_ == int
+
+    @staticmethod
+    def create(type_):
+        if CustomIntSerializer.canSerialize(type_):
+            return ValidatedSerializer(CustomIntSerializer(), [NotNone(), IsInstance(int)])
+        return None
+
+    def default(self):
+        return 0
+
+    def isIn(self, parameterNode, name: str) -> bool:
+        return parameterNode.HasParameter(name)
+
+    def write(self, parameterNode, name: str, value) -> None:
+        parameterNode.SetParameter(name, str(-value))
+
+    def read(self, parameterNode, name: str):
+        return -int(parameterNode.GetParameter(name))
+
+    def remove(self, parameterNode, name: str) -> None:
+        parameterNode.UnsetParameter(name)
+
+
+class TypedParameterNodeTest(unittest.TestCase):
+    def setUp(self):
+        slicer.mrmlScene.Clear(0)
+
+    def test_removes(self):
+        # for each serializer, make sure that calling remove from it
+        # erases all trace of it in the parameterNode
+        numberSerializer = NumberSerializer(int)
+        stringSerializer = StringSerializer()
+        pathSerializer = PathSerializer(pathlib.Path)
+        boolSerializer = BoolSerializer()
+        listSerializer = ListSerializer(NumberSerializer(float))
+
+        parameterNode = newParameterNode()
+
+        numberSerializer.write(parameterNode, "number", 1)
+        stringSerializer.write(parameterNode, "string", "1")
+        pathSerializer.write(parameterNode, "path", pathlib.Path("."))
+        boolSerializer.write(parameterNode, "bool", True)
+        listSerializer.write(parameterNode, "list", [1])
+
+        numberSerializer.remove(parameterNode, "number")
+        stringSerializer.remove(parameterNode, "string")
+        pathSerializer.remove(parameterNode, "path")
+        boolSerializer.remove(parameterNode, "bool")
+        listSerializer.remove(parameterNode, "list")
+
+        self.assertFalse(parameterNode.GetParameterNames())
+
+    def test_isCached(self):
+        @parameterNodeWrapper
+        class ParameterNodeType:
+            int_: int
+            listInt: list[int]
+            custom: Annotated[CustomClass, Default(CustomClass(0, 0, 0))]
+            listCustom: list[CustomClass]
+
+        param = ParameterNodeType(newParameterNode())
+        self.assertTrue(param.isCached("int_"))
+        self.assertTrue(param.isCached("listInt"))
+        self.assertFalse(param.isCached("custom"))
+        self.assertFalse(param.isCached("listCustom"))
+        with self.assertRaises(ValueError):
+            param.isCached("notExistentParameter")
+
+    def test_custom_validator(self):
+        class Is777Or778(Validator):
+            @staticmethod
+            def validate(value):
+                if value != 777 and value != 778:
+                    raise ValueError("Twas not 777 or 778")
+
+        @parameterNodeWrapper
+        class ParameterNodeType:
+            i: Annotated[int, Is777Or778, Default(777)]
+
+        param = ParameterNodeType(newParameterNode())
+
+        with self.assertRaises(ValueError):
+            param.i = -2
+        param.i = 778
+
+    def test_custom_serializer(self):
+        @parameterNodeWrapper
+        class CustomClassParameterNode:
+            # serializer for custom type will be found because it was decorate with @parameterNodeSerializer
+            custom: Annotated[CustomClass, Default(CustomClass(0, 0, 0))]
+
+            # custom serializer for built in type
+            customInt: Annotated[int, CustomIntSerializer()]
+
+        param = CustomClassParameterNode(newParameterNode())
+        self.assertEqual(param.custom.x, 0)
+        self.assertEqual(param.custom.y, 0)
+        self.assertEqual(param.custom.z, 0)
+
+        self.assertEqual(param.customInt, 0)
+
+        param.customInt = 4
+        self.assertEqual(param.parameterNode.GetParameter("customInt"), "-4")
+        self.assertEqual(param.customInt, 4)
+
+    def test_custom_serializer2(self):
+        @parameterNodeWrapper
+        class AnotherCustomClassParameterNode:
+            normalDefault: AnotherCustomClass
+            noneDefault: Annotated[AnotherCustomClass, Default(None)]
+            nonNoneDefault: Annotated[AnotherCustomClass, Default(AnotherCustomClass(1, 2, 3))]
+
+        param = AnotherCustomClassParameterNode(newParameterNode())
+        self.assertEqual(param.normalDefault, AnotherCustomClass(0, 0, 0))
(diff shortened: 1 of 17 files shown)


Connor-Bowley:
Fixed linting errors.

Base/Python/slicer/tests/test_slicer_parameter_node_wrapper.py
```diff
@@ -0,0 +1,510 @@
+import dataclasses
+import pathlib
+from typing import Annotated
+import unittest
+
+import vtk
+import slicer
+
+from _AverageMesh.TypedParameterNodeMk3 import (
```
allemangD, line 9: Should be `from slicer.parameterNodeWraper import (`

Base/Python/slicer/parameterNodeWrapper.py
```diff
@@ -0,0 +1,814 @@
...
+    def supportsCaching():
+        return True
+
+
+class StringSerializer(Serializer):
+    """
+    Serializer for str. This is the type that parameterNode.GetParameter/SetParameter use
+    so it is mainly a pass through.
+    """
+    @staticmethod
```
allemangD, line 279: Nitpick - Making a method static in a derived class seems like a code smell to me; the signature isn't the same anymore. It seems to work, though, so it might not be an issue.

Base/Python/slicer/parameterNodeWrapper.py
```diff
@@ -0,0 +1,814 @@
...
+        self._list.__setitem__(index, item)
+        self._saveList()
+
+    def __iadd__(self, other):
+        self._list.__iadd__(other)
+        self._saveList()
+        return self
+
+    def __add__(self, other):
+        raise NotImplementedError("Adding an ObservedList is not supported. However, += is supported.")
```
allemangD, line 428: Do these need to be forbidden? It seems clear to me that `data.the_list + data.the_other_list` should produce a _new_ list that is not observed.

I'm not sure how that would play with the caching mechanic, though.

I suppose `list(data.the_list) + ...` is a suitable workaround.

Base/Python/slicer/parameterNodeWrapper.py
```diff
@@ -0,0 +1,814 @@
...
+        paramName = lambda index: self._paramName(name, index)
+
+        ret = [self._elementSerializer.read(parameterNode, paramName(index))
+               for index in range(self._len(parameterNode, name))]
+        return ObservedList(parameterNode, self, name, ret)
+
+    def remove(self, parameterNode, name: str) -> None:
+        """Removes this parameter from the node if it exists."""
+        paramName = lambda index: self._paramName(name, index)
+        for index in range(self._len(parameterNode)):
```
allemangD, line 532: Static analyzer revealed that `self._len` and `self._lenSerializer.remove` are both missing the `name` argument here.

However I'm having a hard time working the call stack to see if/how to trigger this method to see if it's actually a problem.

allemangD (commented):
Overall looks good. I was able to play around with it.

Some nitpicks and some holdovers from previous versions that need to be addressed, but I was able to play around with it and I think it can do almost everything I'd need.

I do miss a serializer for `dict`; I think an `ObservableDict` would be a good addition at some point.

I also wonder if there's any value in a serializer for `tuple`?

Base/Python/slicer/tests/test_slicer_parameter_node_wrapper.py
Connor-Bowley: Good catch, yes.

Base/Python/slicer/parameterNodeWrapper.py
Connor-Bowley: I can update to not be static to match.

Base/Python/slicer/parameterNodeWrapper.py
Connor-Bowley: In my opinion, the explicit breaking of the observation helps understandability. If a consensus of people think this should be implicit, I can change it.

Base/Python/slicer/parameterNodeWrapper.py
Connor-Bowley: I will look into this. The way to trigger it should be to do a `list[list[something]]` and make the outer list smaller by setting it to smaller list

```
param.listlistx = [[1],[2]]
param.listlistx = [[3]]
```

allemangD: Ah, yes it is an issue.

Suggest to add to `test_list_of_lists`:


  pl.pop(0)
  self.assertEqual(param.p, [[4, 5, 0, 7]])
  self.assertEqual(param2.p, [[4, 5, 0, 7]])

Might also consider overriding `ObservableList.__del__` to support `del pl[0]`; this is what I tried before but it does not propagate the change to the serializer.

Connor-Bowley: Will add the test. Will look int overriding `__del__`.

Connor-Bowley:
I will work on an ObservableDict so we can use that, but I think it would be good to not block this PR on that (it is added functionality rather than some issue with this PR).

What are you thinking would be the need the `tuple` serialization would fill? Just non-editable lists?

Base/Python/slicer/parameterNodeWrapper.py
Connor-Bowley: Re: __del__

I think that `__delitem__` is the right thing here, which I implemented, but the implementation was wrong (it didn't save the updated list).

Connor-Bowley: See https://github.com/Slicer/Slicer/pull/6684/commits/e3b8b690cbc93be711f1bf69fd53384b106a19eb

Base/Python/slicer/tests/test_slicer_parameter_node_wrapper.py
Connor-Bowley: See https://github.com/Slicer/Slicer/pull/6684/commits/279415ebbfc582816e07a53c458918a33ff07002

Base/Python/slicer/parameterNodeWrapper.py
Connor-Bowley: See https://github.com/Slicer/Slicer/pull/6684/commits/0de798353a335df4f47c58b36e8d3eec58520dfc

Base/Python/slicer/parameterNodeWrapper.py
Connor-Bowley: See also https://github.com/Slicer/Slicer/pull/6684/commits/638e6a5a0f513caccc617fa09513e522be40236f

allemangD:
> I will work on an ObservableDict so we can use that, but I think it would be good to not block this PR on that (it is added functionality rather than some issue with this PR).

I agree.

> What are you thinking would be the need the `tuple` serialization would fill? Just non-editable lists?

I wasn't thinking of anything in particular, but yes. It's a little more convenient to not have to wrap tuples returned from functions.

It would be nice for all those "fundamental" python types like `dict`, `tuple`, `set`, `bytes` etc. to be supported out-of-the-box, but the more I think about it I'm not sure the value added is worth the hassle.

I'd be comfortable merging with only these current serializers, with the expectation that more types be supported in the future if a need arises. It might make sense to convert `parameterNodeWrapper.py` into a package to try and improve organization.

Base/Python/slicer/parameterNodeWrapper.py
allemangD: Oops! Yes, `__delitem__` is what I meant.

Connor-Bowley:
> expectation that more types be supported in the future if a need arises

Definitely agree. And we can always prototype them as custom serializers where they are first needed.

> It might make sense to convert parameterNodeWrapper.py into a package to try and improve organization.

Are you suggesting like adding a `parameterNodeWrapper/` folder in `slicer` and have a setup like follows?

- `slicer/parameterNodeWrapper/__init__.py
- `slicer/parameterNodeWrapper/validators.py
- `slicer/parameterNodeWrapper/serializers.py
- etc

Or are you suggesting like a totally separate python package? I would have some reservations with that because of the strong dependency to the `vtkMRMLScriptedModuleNode` and `vtkMRMLNode`.

allemangD:
I meant a folder containing files like `validators.py` `listSerializer.py`, `dictSerializer.py` or similar. Maybe it would make sense to put all the base classes in some `core.py`? I don't have strong opinions on how exactly it should be organized, but I think the one big module should be broken up if we're to have even more serializers or validators in the future.

Connor-Bowley:
I will squash the current changes, refactor into multiple files and force push a new commit tree.

Connor-Bowley:
Refactored into multiple files.

Connor-Bowley:
One thing that is not supported is doing something like

```py
@parameterNodeWrapper
class CustomParameterNode:
  varname: list[Annotated[str, Choice(['a', 'b', 'c']]
```

which is desirable. I will investigate adding support for this.

Connor-Bowley:
Fixed issue where the element type for the lists couldn't be annotated. Added test for this use case.

Refactored a bit to make the addition of custom serializers more complete, giving them the ability to have their own default validators and values.

Added a new `@parameterNodeSerializer` decorator to register serializers for easy use by the parameterNodeWrapper and the things like `ListSerializer`.

Rebased off of latest main.

Docs/developer_guide/parameter_nodes.md
```diff
@@ -0,0 +1,366 @@
...
+## Overview
+
+Parameter nodes are often used in scripted modules to store data such that it can be easily saved to the MRML scene. They are simply MRML nodes that exist in a scene and store data. One of their most common uses is to save GUI state in module widgets.
+
+The parameter node concept is implemented in C++ in the `vtkMRMLScriptedModuleNode` which has `GetParameter` and `SetParameter` methods for saving arbitrary string data to the scene. While the base `vtkMRMLScriptedModuleNode` is
+great for scene saving, treating all data as strings is not ideal, so a parameter node wrapper was implemented.
+
+## Parameter Node Wrapper Classes
+
+The parameter node wrapper allow wrapping around a `vtkMRMLScriptedModuleNode` parameter node with typed member access. A simple example is as follows.
```
sjh26, line 12:
```suggestion
The parameter node wrapper allows wrapping around a `vtkMRMLScriptedModuleNode` parameter node with typed member access. A simple example is as follows.
```

Docs/developer_guide/parameter_nodes.md
```diff
@@ -0,0 +1,366 @@
...
+  # can now use CustomClass like any other type for building parameterNodeWrappers
+  custom: Annotated[CustomClass, Default(CustomClass(1,2,3))]
+  listOfCustom: list[CustomClass]
+
+```
+
+### Caching
+
+The `vtkMRMLScriptedModuleNode` stores its values as strings, but converting from string to another
+type can be slow. Caching is used to speed up repeated reads under the hood. If the value hasn't been written to the _parameter node wrapper_ then it will use the cached value. Note this means if the underlying `vtkMRMLScriptedModuleNode` parameter node changes not through the wrapper, the cache will not be updated and the cached value will be wrong.
```
sjh26, line 283:
```suggestion
type can be slow. Caching is used to reduce repeated costly reads under the hood. If a value hasn't been written through the _parameter node wrapper_ since the last read, then it will use the cached value. Note this means if the underlying `vtkMRMLScriptedModuleNode` parameter node changes outside of the wrapper, the cache will not be updated and the cached value will be wrong.
```

Docs/developer_guide/parameter_nodes.md
Connor-Bowley: Updated.

Docs/developer_guide/parameter_nodes.md
Connor-Bowley: Updated.

sjh26 (dismissed):

Connor-Bowley:
@jcfr @lassoan @pieper Is this good to merge or did you want to review?

Modules/Scripted/VectorToScalarVolume/VectorToScalarVolume.py
```diff
@@ -1,11 +1,13 @@
 from contextlib import contextmanager
 import logging
+from typing import Annotated
 import qt
 import vtk
 
 import slicer
 from slicer.ScriptedLoadableModule import *
 from slicer.util import VTKObservationMixin
+from slicer.parameterNodeWrapper import *
```
pieper, line 10: I'm not a big fan of importing `*` from modules since then the symbols show up later in the code with no indication where they came from.  I'd usually prefer spelling out the full path in the code (verbose, but explicit) or if that is really too long, then listing out the explicit imports.

pieper (approved):
I think this is worth getting into the preview.  It should simplify bookkeeping 👍

Modules/Scripted/VectorToScalarVolume/VectorToScalarVolume.py
Connor-Bowley: It kind of matched `from slicer.ScriptedLoadableModule import *` above. Could also do something like `import slicer.parameterNodeWrapper as pnw` as a shorter but still explicit middle ground.

pieper: Yes, but for those cases the imported symbols are `ScriptedLoadableModuleWidget` and `VTKObservationMixin` so the full path really would be redundant.

Yes, using `pnw` would be an improvement so someone could easily track where the symbol comes from just by searching in this file.

Connor-Bowley: Updated to use explicit imports.

Connor-Bowley:
Rebased off of main.

Modules/Scripted/VectorToScalarVolume/VectorToScalarVolume.py
```diff
@@ -1,11 +1,17 @@
...
 import vtk
 
 import slicer
 from slicer.ScriptedLoadableModule import *
 from slicer.util import VTKObservationMixin
+from slicer.parameterNodeWrapper import (
+    parameterNodeWrapper,
+    Choice,
+    Default,
+)
```
jcfr, line 14: May be worth also updates the `Scripted*` templates to use the decorator. See https://github.com/Slicer/Slicer/tree/main/Utilities/Templates/Modules

To regenerate the examples in [Extensions/Testing](https://github.com/Slicer/Slicer/tree/main/Extensions/Testing), running the target `SlicerGenerateExtensionTemplates` will do it.

jcfr (dismissed):
Nitpick:
* documentation:
  * link documentation from "Developer Guide / MRML / Advanced topic. If helpful, table could be used if it makes the visual parsing easier.
  * using the admonition like `:::{warning}`, `:::{note}` may be helpful. Here is an example of documentation where these are used. See https://cell-locator.readthedocs.io
* Review the templates to ensure they can be updated in a follow up PR

Modules/Scripted/VectorToScalarVolume/VectorToScalarVolume.py
Connor-Bowley: Will update the templates in a separate PR.

Connor-Bowley:
Updated documentation to use tables. Linked to the top of the developer guide and from MRML overview advanced topics.

jcfr (dismissed):

sjh26 (approved):

jcfr:
Thanks @Connor-Bowley for making this happen :100: :rocket: :tada:

Also thanks to @sjh26 & @allemangD, and everyone else who provided feedback and spent time reviewing :pray:

allemangD:
For the sake of standardization, I wonder if it would be appropriate to add some abstract classes that use this, and make the extension wizard use that abstract class. If not an abc or mixin, I think it would be good to reduce the amount of boilerplate needed to correctly use parameter nodes (preferably wrapped parameter nodes). I'd hope putting that mechanism in the extension wizard or other templates would help encourage module developers to use the same patterns and make static analysis more viable.

A rough idea of the API might be:

```python
class VectorToScalarVolume(SimpleScriptedModule):
    UI_RESOURCE_PATH = 'UI/VectorToScalarVolume.ui'
    PARAMETER_NODE_TYPE = VectorToScalarVolumeParameterNode

    def __init__(self, parent):
        super().__init__(parent)

        self.logic = None

    def setup(self):
        super().setup()  # automatically sets up `self.params`

        self.logic = VectorToScalarVolumeLogic()

        self.ui.inputSelector.currentNodeChanged.connect(self.onGUIChanged)

        ...

    def initializeParameterNode(self):  # no boilerplate; called by the base
        if not self.params.InputVolume:
            ...

    def updateGUIFromParameterNode(self):  # no boilerplate
        self.ui.inputSelector.setCurrentNode(self.params.InputVolume)
        ...

    def updateParameterNodeFromGUI(self):  # no boilerplate
        self.params.InputVolume = self.ui.inputSelector.currentNode

    def onApplyButton(self):
        ...
```

The main idea is there's no boilerplate for managing the parameter node, avoiding recursive gui/parameter updates, etc.

I'm not sure what exactly the scope of this kind of feature should be, or whether a base class is the right mechanism. Maybe a mixin, or method decorators? Either way, I don't think this kind of logic belongs in the existing `ScriptedLoadableModule` classes; some modules customize the workflow significantly and this would just get in the way.

---

I'll think on it more and wait for some preliminary feedback here... if there seems to be a reasonable way forward I'll open a new issue and try prototyping something.

Connor-Bowley:
I plan to update the extension wizard to use `@parameterNodeWrapper` soon.

PR #6694, opened 2022-11-23
[5.2.1] BUG: Fix joint smoothing in Segment Editor

lassoan:
Regression occurred due to VTK API change (ThresholdBetween method was removed from vtkThreshold).

This is a pretty important feature, so we'll need a 5.2.1 patch release soon.

### DIFF
--- a/Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorSmoothingEffect.py
+++ b/Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorSmoothingEffect.py
@@ -456,7 +456,9 @@ If segments overlap, segment higher in the segments table will have priority. <b
         oldOverwriteMode = self.scriptedEffect.parameterSetNode().GetOverwriteMode()
         self.scriptedEffect.parameterSetNode().SetOverwriteMode(slicer.vtkMRMLSegmentEditorNode.OverwriteVisibleSegments)
         for segmentId, labelValue in segmentLabelValues:
-            threshold.ThresholdBetween(labelValue, labelValue)
+            threshold.SetLowerThreshold(labelValue)
+            threshold.SetUpperThreshold(labelValue)
+            threshold.SetThresholdFunction(vtk.vtkThreshold.THRESHOLD_BETWEEN)
             stencil.Update()
             smoothedBinaryLabelMap = slicer.vtkOrientedImageData()
             smoothedBinaryLabelMap.ShallowCopy(stencil.GetOutput())


jamesobutler:
https://github.com/Slicer/Slicer/commit/9f2d8f6cc7f1557e83ffe607612cfdfad077103e was when https://github.com/Slicer/VTK/commit/753e6f95afa70387f7c1b931647368374f7e2363 was included which had the `ThresholdBetween` deprecation removed. So the issue has been present for a little over 6 weeks without it being noticed so probably not a deal breaker for Slicer 5.2.0 to be out a few weeks until Slicer 5.2.1.

lassoan:
Until now I (and probably many others) have been using the latest Slicer Preview Release because the latest stable was always too old. With Slicer-5.0.3 it was different. It was not so old so I could use it for most of my daily work. So, this latest Slicer Preview Release probably did not get much testing.

With this more frequent release schedule we need to do something to motivate at least some users to switch to the latest version.

Should we do "soft launch" similar to 5.0, when we made the release available via the download page, but the release was not announced until it was already 5.0.3? It was a bit confusing though, so I'm not sure it is the best way, so maybe we do this now and do better next time.

Maybe we could add a "release candidate" version (in addition to the stable and preview release). However, it may be a bit too much work with adding yet another track for extension builds, dashboard, etc.

Maybe we could just force ourselves (some core Slicer user group) to switch to the latest Slicer Preview Release for our daily work for at least a month before releasing the new stable based on it. It would be hard, because I would not want to suggest our collaborators to switch to the release candidate just to that they find bugs when they are happy with the current stable release.

lassoan:
> so probably not a deal breaker for Slicer 5.2.0 to be out a few weeks until Slicer 5.2.1.

I agree, it should be fine to get this fixed in 1-2 weeks. Maybe we could just add a link to the Slicer-5.2.1 milestone for the list of known issues.

jamesobutler:
> With this more frequent release schedule we need to do something to motivate at least some users to switch to the latest version.

Reviewing the download stats from October 8th (bug introduction) to November 22nd (today), 70% of downloads were for the stable Slicer 5.0.x release (20182 downloads) compared to 30% of downloads for the Slicer 5.1 Preview (8619 downloads).

pieper:
I tend to use my local debug/release builds for day-to-day work so as long as some of us do that we should catch some issues.  I do tend to use the stable release though for testing extensions just because it's easier to install them than to build them.

jcfr:
Thanks for the fix :100: 

Creating a patch release should not be a problem, after updating the "release" issue templates based on some improvement I collected, I will create a new issue called "Slicer v5.2.1".

lassoan:
I've created a 5.2.1 milestone: https://github.com/Slicer/Slicer/milestone/7

jcfr:
@lassoan As a side note, namespacing the branch with `5.2/` leads to problem like the following when trying to checkout the branch locally after there is already a local branch called `5.2`

```
$ git checkout 5.2/fix-joint-smoothing
fatal: cannot lock ref 'refs/heads/5.2/fix-joint-smoothing': 'refs/heads/5.2' exists; cannot create 'refs/heads/5.2/fix-joint-smoothing'

$ git checkout -b 5.2/fix-joint-smoothing lassoan/5.2/fix-joint-smoothing
fatal: cannot lock ref 'refs/heads/5.2/fix-joint-smoothing': 'refs/heads/5.2' exists; cannot create 'refs/heads/5.2/fix-joint-smoothing'
```

Prefixing with `5.2-` works well 

```
$ git checkout -b  "5.2-fix-joint-smoothing" lassoan/5.2/fix-joint-smoothing 
Branch '5.2-fix-joint-smoothing' set up to track remote branch '5.2/fix-joint-smoothing' from 'lassoan'.
Switched to a new branch '5.2-fix-joint-smoothing'
```

PR #6697, opened 2022-11-23
Improve slicer util to bool docstring

jcfr:


lassoan (approved):
Looks good to me, thank you.

PR #6704, opened 2022-11-26
STYLE: Mark translatable strings in Base/QTGUI

mhdiop:
See https://github.com/Slicer/SlicerLanguagePacks/issues/12


Base/QTGUI/qSlicerModulesMenu.cxx
```diff
@@ -69,10 +69,11 @@ qSlicerModulesMenuPrivate::qSlicerModulesMenuPrivate(qSlicerModulesMenu& object)
   this->NoModuleAction = nullptr;
   this->DuplicateActions = false;
   this->ShowHiddenModules = false;
-  this->TopLevelCategoryOrder << "Wizards" << "Informatics" << "Registration"
-    << "Segmentation" << "Quantification" << "Diffusion" << "IGT"
-    << "Filtering" << "Surface Models" << "Converters" << "Endoscopy"
-    << "Utilities" << "Developer Tools" << "Legacy" << "Testing";
+  this->TopLevelCategoryOrder << qSlicerModulesMenu::tr("Wizards") << qSlicerModulesMenu::tr("Informatics") << qSlicerModulesMenu::tr("Registration")
+    << qSlicerModulesMenu::tr("Segmentation") << qSlicerModulesMenu::tr("Quantification") << qSlicerModulesMenu::tr("Diffusion")
```
lassoan, line 73: We use the abstract module factory class name as context for module category everywhere. If we find that it is hard to enforce this then we may reconsider, but for now let's try to follow this convention.

Base/QTGUI/qSlicerSaveDataDialog.cxx
```diff
@@ -965,7 +965,7 @@ vtkMRMLNode* qSlicerSaveDataDialogPrivate::getNodeByID(char *id, vtkMRMLScene* s
       {
       vtkMRMLSceneViewNode *svNode = vtkMRMLSceneViewNode::SafeDownCast(*it);
       // skip "Master Scene View" since it contains the same nodes as the scene
-      if (svNode->GetName() && std::string("Master Scene View") == std::string(svNode->GetName()))
+      if (svNode->GetName() && std::string(/*no tr*/"Master Scene View") == std::string(svNode->GetName()))
```
lassoan, line 968: Search the full Slicer code base and mark all usages of "Master Scene View" as "no tr", even in VTK classes.

Base/QTGUI/qSlicerSettingsDeveloperPanel.cxx
```diff
@@ -71,20 +71,20 @@ void qSlicerSettingsDeveloperPanelPrivate::init()
 
   // Register settings
   q->registerProperty("Developer/DeveloperMode", this->DeveloperModeEnabledCheckBox,
-                      "checked", SIGNAL(toggled(bool)),
-                      "Enable/Disable developer mode", ctkSettingsPanel::OptionRequireRestart);
+    "checked", SIGNAL(toggled(bool)),
```
lassoan, line 74: You can use no tr for "checked" just to make it very clear.

Base/QTGUI/qSlicerSettingsViewsPanel.cxx
```diff
@@ -296,7 +296,7 @@ void qSlicerSettingsViewsPanel::sliceViewOrientationChangedByUser()
       {
       // Non-default orientation is chosen and no orientation marker is displayed.
       // To ensure that there is no accidental mixup of orientations, show the orientation marker.
-      d->SliceOrientationMarkerTypeComboBox->setCurrentText("axes");
+      d->SliceOrientationMarkerTypeComboBox->setCurrentText(/*no tr*/"axes");
```
lassoan, line 299: Combobox can store user data for each item. It would be nicer to translate the item text and store the user data (the nob-translated text) in application settings and ithe scene.

There is a mapper that allows registering a settings property with userdata. If you don't find it then let me know.

Base/QTGUI/qSlicerWebWidget.cxx
```diff
@@ -386,16 +386,16 @@ void qSlicerWebWidget::onDownloadProgress(qint64 bytesReceived, qint64 bytesTota
   QString unit;
   if (speed < 1024)
     {
-    unit = "bytes/sec";
+    unit = /*no tr*/"bytes/sec";
```
lassoan, line 389: Why not translate?

lassoan (commented):
Very nice work! I added a couple of comments. They are mostly just nuances.

mhdiop:
Thank you @lassoan for your valuable comments. 

I'm going to review them and come you back with a new commit that take them into account.

Regards!

lassoan:
It would be nice if you could make progress on this. There are a lot of modified files and if the changes are not merged for long time then we could keep running into merge conflicts.

Base/QTGUI/qSlicerModelsDialog.cxx
```diff
@@ -61,7 +61,7 @@ void qSlicerModelsDialogPrivate::openAddModelFileDialog()
   QStringList filters = qSlicerFileDialog::nameFilters(q->fileType());
   // TODO add last open directory
   this->SelectedFiles = QFileDialog::getOpenFileNames(
-    this, "Select Model file(s)", "", filters.join(", "));
+    this, qSlicerModelsDialog::tr("Select Model file(s)"), "", filters.join(", "));
```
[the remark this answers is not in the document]
mhdiop: I think that this string represents the dialog's caption and doesn't depend on a count that determines if singular or plural should be used.  However, as requested, I will use this approach wherever I see a case for it.

Base/QTGUI/qSlicerExtensionsManagerWidget.cxx
```diff
@@ -453,7 +453,7 @@ void qSlicerExtensionsManagerWidget::onModelUpdated()
     }
 
   d->tabWidget->setTabText(manageExtensionsTabIndex,
-                           QString("Manage Extensions (%1)").arg(managedExtensionsCount));
+                           tr("Manage Extensions (%1)").arg(managedExtensionsCount));
```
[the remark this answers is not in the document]
mhdiop: Thanks for your feedback, it's added now.

However, I notice that lupdate doesn't handle developper comments the same way. In fact, one can use :

- [translator comments](https://doc.qt.io/qt-5/i18n-source-translation.html#translator-comments) like `//: developer comments` or `/*: developer comments */`
-  the old fashion [ _**tr**_ disambiguation argument](https://doc.qt.io/qt-5/i18n-source-translation.html#disambiguation) like `tr("translatable string", "developer comments")`

Using translator comments will result in a `<extracomment>` tag in the translation file while the use of the disambiguation parameter will appear as a `<comment>` tag.

As discussed in the last meeting, I will test them both on Weblate to verify how they are handled so that to see which comment style we may consider in the project.

Base/QTGUI/qSlicerSettingsViewsPanel.cxx
mhdiop: I understand the problem you raised but I don't find the mapper you are talking about.

I wanted to rewrite the code as follows but I don't have the guarantee that the "None" item's data was initialised to "none". 

```cpp
if (d->SliceOrientationMarkerTypeComboBox->currentData() == "none")
{
  // Non-default orientation is chosen and no orientation marker is displayed.
  // To ensure that there is no accidental mixup of orientations, show the orientation marker.
  int index = d->SliceOrientationMarkerTypeComboBox->currentIndex();
  d->SliceOrientationMarkerTypeComboBox->setItemData(index, /*no tr*/"axes");
  d->SliceOrientationMarkerTypeComboBox->setCurrentText(tr("axes"));
}
```

So I finally mark the current text as translatable (both `none` and `axes`). I planned to update it whenever I have more information on that mapper.

mhdiop:
It's done now. 

Your comments are now taken into account in the last commit.

Thanks.

Base/QTGUI/qSlicerExtensionsManagerWidget.cxx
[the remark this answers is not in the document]
lassoan: > the old fashion [tr disambiguation argument](https://doc.qt.io/qt-5/i18n-source-translation.html#disambiguation) like tr("translatable string", "developer comments")

Disambiguation and comments are for entirely purposes. Disambiguation argument is used at runtime for getting different translation for the same source text. Comments are ignored at runtime.

Here we need to use comments, as we don't need to distinguish between several "Manage Extensions (%1)" strings.

It would be useful to see how comment and extracomment show up in weblate. That will help us decide which one to use in what cases.

Base/QTGUI/qSlicerModelsDialog.cxx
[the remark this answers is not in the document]
lassoan: You are correct, in this case we don't know if the user wants to select one or more files, so there is no need to use the plural translation feature.

Base/QTGUI/qSlicerSettingsViewsPanel.cxx
lassoan: The combobox items should be removed from the .ui file and instead the items (translated text and non-translated userdata) should be added as it is done here (just using qSlicerSettingsViewsPanel::tr instead of QWidget::tr):

https://github.com/Slicer/Slicer/blob/6c67bc6b8f637c5c31ec5aadf404778ce774f4bb/Base/QTGUI/qSlicerSettingsViewsPanel.cxx#L105-L112

Base/QTGUI/qSlicerExtensionsManagerWidget.cxx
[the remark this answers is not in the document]
mhdiop: Got it, you're right.

I will test them both and Weblate and see how they are handled.

Thanks !

Base/QTGUI/qSlicerSettingsViewsPanel.cxx
mhdiop: Ok, it's done now.

By the way, which of the following approach is the best when it's about dealing with commits related to PR comments :

- Add new commits at each update that concerns one or more comments ?
- Always merge all commits and force push the branch  so that there is always one commit related to the PR ?

Thank you !

lassoan: Thank you!

There is no strict rule for additional commits or force push. When there are merge conflicts or things just get too complicated then it makes sense to do a force-push. Otherwise additional commits are nice because they don't invalidate the comments.

Base/QTGUI/qSlicerExtensionsManagerWidget.cxx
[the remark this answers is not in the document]
mhdiop: After some tests, I realise that Weblate doesn't handle the `<comment>` and `<extracomment>` tags the same way.

- Disambiguation strings present on `<comment>` tags are just added to the context name of the string so that to constitute its `Key`. That key is displayed in the `String information` and the `Translation` sections of Weblate. Examples of such cases can be found [here](https://hosted.weblate.org/translate/3d-slicer/3d-slicer/fr/?offset=1&q=qSlicerWelcomeModuleWidget+extension+update+is+available&sort_by=context&checksum=) or [there](https://hosted.weblate.org/translate/3d-slicer/3d-slicer/fr/?checksum=e2279a7500de6c73&sort_by=-priority,position).
- Translator comments present on `<extracomment>` tags are displayed as `String source description` in the `String information` section of Weblate. Examples of such cases can also be found [here](https://hosted.weblate.org/translate/3d-slicer/3d-slicer/fr/?checksum=9e57190062ff79e2&sort_by=context) or [there](https://hosted.weblate.org/translate/3d-slicer/3d-slicer/fr/?offset=1&q=qSlicerSequencesModuleWidget+Limited&sort_by=context&checksum=).

I think then we should use translator comments to integrate comments for translators.

Base/QTGUI/qSlicerSettingsViewsPanel.cxx
mhdiop: Ok, that's good then.

Thank you again for all of your valuable feedback!

mhdiop:
Hi @lassoan.

I write this comment to remind you that all comments has been taken into account.

If there are any more things to add, please let me know!

Thank you!

lassoan (approved):
It looks good, thank you!

PR #6715, opened 2022-12-01
BUG: Update vtkMRMLParser to support custom type ending with "Node" string

aravindskumar98:
For module names that contain multiple "Node" substrings, For example: vtkMRMLROS2NodeNode, the previous implementation using .find() would be restrictive. Considering that the logic is to check whether the className ends with "Node" from the end ( using .rfind() ) is better.

### DIFF
--- a/Libs/MRML/Core/vtkMRMLParser.cxx
+++ b/Libs/MRML/Core/vtkMRMLParser.cxx
@@ -127,7 +127,7 @@ void vtkMRMLParser::StartElement(const char* tagName, const char** atts)
     className = "vtkMRML";
     className += tagName;
     // Append 'Node' prefix only if required
-    if (className.find("Node") != className.size() - 4)
+    if (className.rfind("Node") != className.size() - 4)
       {
       className += "Node";
       }


pieper (approved):
This makes sense to me.  Does anyone see any potential issues?

lassoan (approved):
Thank you, this makes things more robust.

I guess we just haven't had a node type that had not inside its name until now.
 
It is very unfortunate that ROS uses the "node" term as well. To avoid name like "vtkMRMLROS2NodeNode" and potential misunderstandings, it might make sense to not refer to ROS nodes by the "node" word in Slicer modules, such as "rosn", "rosnx", "rode", ... resulting in node name like "vtkMRMLROSNNode", "vtkMRMLROSNxNode", or "vtkMRMLRodeNode". None of these sounds really great to me, but it is probably not that hard to get used to it, and the clarity of the name could worth it.

lassoan:
We have a commit check failure because commit messages must start with specific prefixes. Could you add the `BUG:` prefix to the  commit message and force-push to this same branch?

If you are not sure how to do it, then let us know and we can fix it up (but then you'll probably just show up as coauthor on the commit).

aravindskumar98:
@lassoan I would like to do it! I have changed the commit message, however it says that _First-time contributors need a maintainer to approve running workflows_

jamesobutler:
@aravindskumar98 It appears you have changed the PR title, but not the commit message. See https://github.com/Slicer/Slicer/pull/6715/commits. For the required check to pass the commit message must be the one updated.

jcfr:
> however it says that First-time contributors need a maintainer to approve running workflows

All set. :rocket: 

Thanks @aravindskumar98 for your patience :pray:

jcfr (commented):
One last update, could you amend the commit to use the following commit message ? 

Or that looks to you, let me know and I will take care of amending while integrating. Assuming the tests pass.


```
BUG: Update vtkMRMLParser to custom type ending with "Node" string

The convention for naming MRML node is to follow the pattern `vtkMRML<MyCustomType>Node`.

To support the the case where `<MyCustomType>` itself contains the string "Node",
this commit updates the method `vtkMRMLParser::StartElement(const char* tagName, const char** atts)`
to more reliably initialize `className`variable  based of the `tagName`.

For example to support the custom type "ROS2Node" leading to have the class name "vtkMRMLROS2NodeNode",
this commit changes the implementation to use `std::string::rfind` instead of `std::string::find`.

This allows to successfully parse XML like the following:

<?xml version="1.0" encoding="UTF-8"?>
<MRML version="Slicer 5.3.0 31338" userTags="">
<ROS2Node
  id="vtkMRMLROS2NodeNode1"  name="ROSv2 Node 1"></ROS2Node>
</MRML>
```

aravindskumar98:
@jcfr I've updated the commit message

jcfr:
After checking locally that the tests associated with the `MRML` label still pass, I will merge.

jcfr:
Summary of testing

```
$ ctest -R qMRMLLayoutManagerVisibilityTest -V
[...]
99% tests passed, 1 tests failed out of 278

Label Time Summary:
MRMLCore                          =  68.03 sec*proc (115 tests)
MRMLDisplayableManager            =  70.72 sec*proc (8 tests)
MRMLLogic                         =  11.67 sec*proc (13 tests)
qMRMLWidgets                      = 1597.60 sec*proc (93 tests)
qSlicerAnnotationsModule          =  36.68 sec*proc (17 tests)
qSlicerCropVolumeModule           =   1.14 sec*proc (1 test)
qSlicerMarkupsModule              =  11.89 sec*proc (13 tests)
qSlicerModelsModuleWidgets        =  59.14 sec*proc (2 tests)
qSlicerPlotsModuleWidgets         =  25.34 sec*proc (1 test)
qSlicerSequencesModule            =   4.81 sec*proc (3 tests)
qSlicerSequencesModuleWidgets     =  27.53 sec*proc (1 test)
qSlicerTransformsModuleWidgets    =  36.51 sec*proc (2 tests)
qSlicerVolumeRenderingModule      =  37.36 sec*proc (6 tests)

Total Test time (real) = 420.66 sec

The following tests FAILED:
	180 - qMRMLLayoutManagerVisibilityTest (Failed)
Errors while running CTest
```

Since the test `qMRMLLayoutManagerVisibilityTest` is failing on my workstation independently of this change, I am integrating.

PR #6720, opened 2022-12-06
BUG: Fix scripted module template observer warning

Sunderlandkyl:
Whenever the user leaves the template module, the parameter node observers are removed. If the module is entering after that, setParameterNode() will attempt to remove the observer again, and throw a warning message.

Fixed by checking if the observer exists before removing in setParameterNode().

See: https://discourse.slicer.org/t/warning-of-observer-from-scripted-module-in-slicer-5-0-3/


lassoan (approved):
Looks good, thank you!

baderstine:
I think this same issue happens with this line in 5.6.2: 
[TemplateKey.py line 191](https://github.com/Slicer/Slicer/blob/0ceb67e3853f7a72f5bd9a1494aedf26971c1c2b/Utilities/Templates/Modules/Scripted/TemplateKey.py#L191) 

It seems to be fixed if changed to:
```python
            if self.hasObserver(self._parameterNode, vtk.vtkCommand.ModifiedEvent, self._checkCanApply):
                self.removeObserver(self._parameterNode, vtk.vtkCommand.ModifiedEvent, self._checkCanApply)
```

PR #6723, opened 2022-12-07
BUG: Support running multiple processing tasks in parallel

jcfr:
This commit creates 2 processing threads to ensure at list two CLIS can be executed in parallel.

It was created in the context of this test project and associated discussion. See https://github.com/KitwareMedical/SlicerParallelTest and https://discourse.slicer.org/t/running-a-module-in-parallel/12641/11

### DIFF
--- a/Base/Logic/vtkSlicerApplicationLogic.cxx
+++ b/Base/Logic/vtkSlicerApplicationLogic.cxx
@@ -58,7 +58,6 @@ vtkStandardNewMacro(vtkSlicerApplicationLogic);
 vtkSlicerApplicationLogic::vtkSlicerApplicationLogic()
 {
   this->ProcessingThreader = itk::PlatformMultiThreader::New();
-  this->ProcessingThreadId = -1;
   this->ProcessingThreadActive = false;
 
   this->ModifiedQueueActive = false;
@@ -79,20 +78,9 @@ vtkSlicerApplicationLogic::vtkSlicerApplicationLogic()
 //----------------------------------------------------------------------------
 vtkSlicerApplicationLogic::~vtkSlicerApplicationLogic()
 {
-  // Note that TerminateThread does not kill a thread, it only waits
-  // for the thread to finish.  We need to signal the thread that we
-  // want to terminate
-  if (this->ProcessingThreadId != -1 && this->ProcessingThreader)
+  if (this->ProcessingThreader)
     {
-    // Signal the processingThread that we are terminating.
-    this->ProcessingThreadActiveLock.lock();
-    this->ProcessingThreadActive = false;
-    this->ProcessingThreadActiveLock.unlock();
-
-    // Wait for the thread to finish and clean up the state of the threader
-    this->ProcessingThreader->TerminateThread( this->ProcessingThreadId );
-
-    this->ProcessingThreadId = -1;
+    this->TerminateProcessingThread();
     }
 
   delete this->InternalTaskQueue;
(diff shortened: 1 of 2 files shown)


Base/Logic/vtkSlicerApplicationLogic.cxx
```diff
@@ -163,16 +152,20 @@ void vtkSlicerApplicationLogic::PrintSelf(ostream& os, vtkIndent indent)
...
     this->ProcessingThreadActiveLock.lock();
     this->ProcessingThreadActive = true;
     this->ProcessingThreadActiveLock.unlock();
 
-    this->ProcessingThreadId
-      = this->ProcessingThreader
-      ->SpawnThread(vtkSlicerApplicationLogic::ProcessingThreaderCallback,
-                    this);
+    // Start processing threads (TODO: make the number of threads a setting)
+    int numberOfProcessingThreads = 2;
```
jcfr, line 162: Question:
* What should be the default ? Would it break existing workflow to have this than 1 by default ?
* An entry in the Module settings panel could be added ? 
* Should we also support setting the number of worker thread through command line and/or env variable ?

RafaelPalomar: I just came to know about this PR. Great initiative. I'm glad to see efforts to make the way for parallel processing.

IMO, the safest would be to leave the default to 1 and add both an entry in the module settings and a command line flag to change the behavior. As a downside, this approach may not expose the feature to the users enough to make a good detection of benefits and problems.

lassoan: We could use 4 as a default value. That would make users feel like everything runs in parallel and would probably not overwhelm most configurations. It would need to be able to change this in application settings.

Base/Logic/vtkSlicerApplicationLogic.cxx
```diff
@@ -233,10 +230,18 @@ void vtkSlicerApplicationLogic::TerminateProcessingThread()
     this->ProcessingThreadActive = false;
     this->ProcessingThreadActiveLock.unlock();
 
-    this->ProcessingThreader->TerminateThread( this->ProcessingThreadId );
-    this->ProcessingThreadId = -1;
-
     std::vector<int>::const_iterator idIterator;
+
+    // Terminate processing threads
+    idIterator = this->ProcessingThreadIDs.begin();
```
Connor-Bowley, line 236: Why wouldn't you just use a range based for loop?

```cpp
for (const auto id : this->ProcessingThreadIDs)
  {
    this->ProcessingThreader->TerminateThread(id);
  }
this->ProcessingThreadIDs.clear();
```

Connor-Bowley: I guess the iterator is consistent with the networking threads below, but a range based for loop would be better for both of them.

Base/Logic/vtkSlicerApplicationLogic.cxx
pieper: I think an "auto" setting by default, that sets the number to the number of CPUs minus one could be good.

Note though that having a single CLI process is not always a bad thing, since most ITK algorithms that have been the traditional key CLI use case are already multithreaded and running multiples could lead to thrashing.

Perhaps another extension could be to add a CLI flag in the XML that declares if they are multithreaded and make it so that only one of those is run at a time.  This would be easy to add to the built-in CLIs.

PR #6727, opened 2022-12-09
Parameter node wrapper support for tuple and dict

Connor-Bowley:
Adds support for `tuple[type1, type2, type3, ...]` and `dict[keyType, valueType]`.
Adds an `ObservedDict` similar to the `ObservedList`. I did not think an `ObservedTuple` was  needed due to the immutability of `tuple`s.

### DIFF
--- a/Base/Python/slicer/parameterNodeWrapper/serializers.py
+++ b/Base/Python/slicer/parameterNodeWrapper/serializers.py
@@ -6,7 +6,26 @@ import typing
 import slicer
 
 from .util import splitAnnotations
-from .validators import *
+from .validators import (
+    extractValidators,
+    NotNone,
+    IsInstance,
+)
+
+__all__ = [
+    "parameterNodeSerializer",
+
+    "Serializer",
+    "ValidatedSerializer",
+    "NumberSerializer",
+    "StringSerializer",
+    "PathSerializer",
+    "BoolSerializer",
+    "NodeSerializer",
+    "ListSerializer",
+    "TupleSerializer",
+    "DictSerializer",
+]
 
 
 class Serializer(abc.ABC):
--- a/Base/Python/slicer/parameterNodeWrapper/serializers.py
+++ b/Base/Python/slicer/parameterNodeWrapper/serializers.py
@@ -1,5 +1,6 @@
 import abc
 import collections
+import logging
 import pathlib
 import typing
 
@@ -215,7 +216,7 @@ def createSerializerFromAnnotatedType(annotatedType):
     type_, annotations = splitAnnotations(annotatedType)
     serializer, annotations = createSerializer(type_, annotations)
     if annotations:
-        print(f"Warning: Unused annotations: {annotations}")
+        logging.warning(f"Unused annotations: {annotations}")
     return serializer
 
 
@@ -535,7 +536,7 @@ class ListSerializer(Serializer):
         if ListSerializer.canSerialize(type_):
             args = typing.get_args(type_)
             if len(args) != 1:
-                print("Unexpected list[] type arg length")
+                logging.warning("Unexpected list[] type arg length")
             if len(args) == 0:
                 Exception("Unsure how to handle a typed list with no discernible type")
             return ListSerializer(createSerializerFromAnnotatedType(args[0]))
--- a/Base/Python/slicer/parameterNodeWrapper/serializers.py
+++ b/Base/Python/slicer/parameterNodeWrapper/serializers.py
@@ -192,6 +192,14 @@ def createSerializer(type_, annotations):
     return (ValidatedSerializer(serializer, validators), annotations)
 
 
+def createSerializerFromAnnotatedType(annotatedType):
+    type_, annotations = splitAnnotations(annotatedType)
+    serializer, annotations = createSerializer(type_, annotations)
+    if annotations:
+        print(f"Warning: Unused annotations: {annotations}")
+    return serializer
+
+
 @parameterNodeSerializer
 class NumberSerializer(Serializer):
     """
@@ -511,11 +519,7 @@ class ListSerializer(Serializer):
                 print("Unexpected list[] type arg length")
             if len(args) == 0:
                 Exception("Unsure how to handle a typed list with no discernible type")
-            elementType, elementAnnotations = splitAnnotations(args[0])
-            serializer, annotations = createSerializer(elementType, elementAnnotations)
-            if annotations:
-                print(f"Warning: Unused annotations: {annotations}")
-            return ListSerializer(serializer)
+            return ListSerializer(createSerializerFromAnnotatedType(args[0]))
         return None
 
     def __init__(self, elementTypeSerializer):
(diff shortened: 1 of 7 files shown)


Base/Python/slicer/parameterNodeWrapper/serializers.py
```diff
@@ -192,6 +192,14 @@ def createSerializer(type_, annotations):
     return (ValidatedSerializer(serializer, validators), annotations)
 
 
+def createSerializerFromAnnotatedType(annotatedType):
+    type_, annotations = splitAnnotations(annotatedType)
+    serializer, annotations = createSerializer(type_, annotations)
+    if annotations:
+        print(f"Warning: Unused annotations: {annotations}")
```
allemangD, line 199: Probably better to use [`warnings`](https://docs.python.org/3/library/warnings.html) for this.

And the other `print()` warnings, too, for that matter. Probably not a huge deal but the warnings module seems more semantically correct.

Base/Python/slicer/tests/test_slicer_parameter_node_wrapper.py
```diff
@@ -525,6 +531,102 @@ class ParameterNodeType:
...
+
+        self.assertEqual(param.a, (CustomClass(0, 0, 0), 0))
+
+        param.a = (CustomClass(7, 6, 5), 9)
+        self.assertEqual(param.a, (CustomClass(7, 6, 5), 9))
+
+    def test_dict(self):
+        @parameterNodeWrapper
+        class ParameterNodeType:
+            a: dict[str, int]
```
allemangD, line 572: The same comments about mutable-valued dictionaries; maybe `dict[str, list[int]]` or a nested dict.

Base/Python/slicer/tests/test_slicer_parameter_node_wrapper.py
```diff
@@ -525,6 +531,102 @@ class ParameterNodeType:
         param2 = ParameterNodeType(param.parameterNode)
         self.assertEqual(param2.p, [[4, 5, 0, 7]])
 
+    def test_tuple(self):
+        @parameterNodeWrapper
+        class ParameterNodeType:
+            a: tuple[Annotated[int, Minimum(0)], str]
+            b: Annotated[tuple[float, bool], Default((44.0, False))]
```
allemangD, line 538: > I did not think an `ObservedTuple` was needed due to the immutability of `tuple`s.

There is still the case that one of the tuple's elements is mutable; it might be worth adding something like `tuple[list[int]]` to the test suite. I wouldn't expect issues since the logic is the same as other containers.

Docs/developer_guide/parameter_nodes.md
```diff
@@ -110,20 +112,24 @@ This will make the default value of the `numIterations` parameter 500. If the `n
...
+| `dict` (hinted as `dict[keyType, valueType]`)    | `{}` (empty dictionary)                                                                |
+| `pathlib.Path`                                   | `pathlib.Path()` (which is the current directory)                                      |
+| `pathlib.PosixPath`                              | `pathlib.PosixPath()` (which is the current directory)                                 |
+| `pathlib.WindowsPath`                            | `pathlib.WindowsPath()` (which is the current directory)                               |
+| `pathlib.PurePath`                               | `pathlib.PurePath()` (which is the current directory)                                  |
+| `pathlib.PurePosixPath`                          | `pathlib.PurePosixPath()` (which is the current directory)                             |
+| `pathlib.PureWindowsPath`                        | `pathlib.PureWindowsPath()` (which is the current directory)                           |
+
+Note: For specifying the default of a tuple, use `Annotated[tuple[int, bool], Default((4, True))]`, not `tuple[Annotated[int, Default(4)], Annotated[bool, Default(True)]]`.
```
allemangD, line 132: Is it possible to support this? It seems like it would "just work" but if there's some technical reason why it doesn't maybe add a comment about why and a runtime warning suggesting the correct approach.

Docs/developer_guide/parameter_nodes.md
```diff
@@ -293,7 +299,7 @@ This is the chosen behavior for the following reasons:
 
 Because Python objects are returned by reference, when a cached value is returned and then modified, the modification needs to be written back to the parameter node. Otherwise, the cached value and the parameter node will get out of sync.
 
-This write-on-change behavior has been implemented for the ListSerializer. The ListSerializer does not actually return a `list`, it returns an `ObservedList` that updates the parameter node whenever it is modified. `ObservedList` implements most `list` functions. This allows the following to work seamlessly:
+This write-on-change behavior has been implemented for the serializers for `list`, `tuple`, `dict`, and the `parameterPack`s. The ListSerializer does not actually return a `list`, it returns an `ObservedList` that updates the parameter node whenever it is modified. `ObservedList` implements most `list` functions. This allows the following to work seamlessly:
```
allemangD, line 302: `parameterPack` is missing from this PR.

Docs/developer_guide/parameter_nodes.md
```diff
@@ -331,6 +337,8 @@ The following methods are available for `ObservedList`:
 Similarly for a parameter of `list[list[type]]`, an `ObservedList[ObservedList[type]]` is returned.
 When calling `param.values.append`, `param.values[index] = object` or `+=` in these cases, a normal `list` can be passed in and it will be converted to an `ObservedList`.
 
+There are similar mechanisms in place for `tuple`, `dict`, and the `parameterPack`s.
```
allemangD, line 340: `parameterPack` is missing from this PR.

allemangD (commented):
Nitpick re python package that I overlooked in the initial PR: given `import *` in `__init__.py` it would probably be good to set `__all__` on each module.

Base/Python/slicer/tests/test_slicer_parameter_node_wrapper.py
allemangD: "The same comments" referring to the comments on `TupleSerializer` - https://github.com/Slicer/Slicer/pull/6727#discussion_r1044876943

Docs/developer_guide/parameter_nodes.md
Connor-Bowley: Good catch. Will fix docs.

Docs/developer_guide/parameter_nodes.md
Connor-Bowley: Good catch. Will fix docs. For reference for anyone else, `parameterPack` will be a part of a subsequent PR around the parameter node wrapper.

Docs/developer_guide/parameter_nodes.md
Connor-Bowley: There are a couple reasons I chose not to support this.

Interface consistency:

- Doing `tuple[Annotated[int, Default(4)], Annotated[bool, Default(True)]]` may make sense, but it is the only container that it currently makes sense for.
    - Doing `list[Annotated[int, Default(4)]]` is confusing. What is defaulted?
    - Similarly `dict[Annotated[str, Default("hi")], Annotated[int, Default(4)]]` doesn't make sense.
    - However, in each of these cases setting a default for the whole value makes sense as opposed to piecewise values (`Annotated[list[int], Default([4])]` or `Annotated[dict[str, int], Default({"hi": 4})` both make sense).

Technical reasons:

- I didn't want the serializers to have to deal with either names or custom default values (the `Default` annotation). That seemed like it should be something else's concern.
    - `_ParameterInfo`/`_Parameter`/`_CachedParameter` in `wrapper.py` were added to deal with these.
    - For simplicity, it makes sense for the composite serializers (list, tuple, dict) to just deal with serializers and not the `_Parameter` classes.
    - Work could be done to refactor somewhat and support this for tuple, but I think the added value is low since it does not add functionality, just multiple ways of doing the same thing.

Connor-Bowley: I can summarize this in the (docs? code?) for why this is not currently supported.

Base/Python/slicer/tests/test_slicer_parameter_node_wrapper.py
Connor-Bowley: Added test with dictionary of lists.

Base/Python/slicer/tests/test_slicer_parameter_node_wrapper.py
Connor-Bowley: Added test for tuple of lists (which incidentally is also a test for a 1 value tuple).

Docs/developer_guide/parameter_nodes.md
Connor-Bowley: Fixed.

Docs/developer_guide/parameter_nodes.md
Connor-Bowley: Fixed.

Connor-Bowley:
Added an `__all__` to each file and removed internal usage of wildcard imports.

Connor-Bowley:
Rebased on main and squashed fixup commits.

allemangD (approved):
All those comments have been addressed, thanks @Connor-Bowley. LGTM! :rocket:

allemangD:
I'm going to merge now... been sitting here without any other objections.

PR #6728, opened 2022-12-10
ENH: Add color and 3dSlicerLabel for some DICOM terminology types

lassoan:
This adds colors and moer user-friendly labels to some segmentation types that are used in TotalSegmentator.

@fedorov This is a draft pull request, because I assume that these changes should be done somewhere in DCMQI and then the json files need to be regenerated. Can you give advice about what/how to make this changes in DCMQI?

### DIFF
--- a/Modules/Loadable/Terminologies/Resources/SegmentationCategoryTypeModifier-DICOM-Master.json
+++ b/Modules/Loadable/Terminologies/Resources/SegmentationCategoryTypeModifier-DICOM-Master.json
@@ -2555,6 +2555,11 @@
             "paired": "U"
           },
           {
+            "recommendedDisplayRGBValue": [
+              212,
+              188,
+              102
+            ],
             "CodeMeaning": "Sacrum",
             "CodingSchemeDesignator": "SCT",
             "cid": "4031",
@@ -2565,6 +2570,11 @@
             "paired": "U"
           },
           {
+            "recommendedDisplayRGBValue": [
+              212,
+              188,
+              102
+            ],
             "CodeMeaning": "Coccyx",
             "CodingSchemeDesignator": "SCT",
             "cid": "4031",
@@ -2585,8 +2595,14 @@
             "paired": "Y",
             "Modifier": [
               {
+                "recommendedDisplayRGBValue": [
+                  212,
+                  188,
+                  102
+                ],
                 "CodeMeaning": "Right",
                 "CodingSchemeDesignator": "SCT",
+                "3dSlicerLabel": "right scapula",
                 "cid": "244",
                 "UMLSConceptUID": "C0205090",
                 "CodeValue": "24028007",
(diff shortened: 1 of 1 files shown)


pieper:
@lassoan fyi @dclunie offered to review the labels and make sure any needed coded concepts are available in DICOM.

lassoan:
In this PR I did not add any new codes, only added colors and labels for Slicer. However, I added 41 new codes to TotalSegmentator Slicer module (the big number comes from having a separate code for each rib and each vertebra):

https://github.com/lassoan/SlicerTotalSegmentator/blob/ec0fe26ec7ed596a2d5bf55d2aaa29c3174ad83e/TotalSegmentator/Resources/SegmentationCategoryTypeModifier-TotalSegmentator.term.json

Some of the codes are included in the DICOM standard and therefore are free to use. Some of them are not included, but the terms are trivial (they are available in several other free terminologies) and it is only a small number of codes, so I assumed "fair use" applies.

However, if we want to use standard terminology for anatomical segmentations everywhere then it should be clarified with SNOMED if they are OK with this. If they are not then we can still very easily switch to a freely usable terminology, such as TA2 (the DICOM standard supports this, we could just use the `TA2` coding scheme designator). This would make SNOMED CT largely irrelevant in the medical image segmentation community, so it would make sense for SNOMED to allow free use of their anatomical terms. For example, they could allow free use of terms that are already listed in freely usable anatomical terminologies, such as TA2, MeSH, or FMA.

pieper:
With a little more digging I found this: https://www.dicomstandard.org/docs/librariesprovider2/dicomdocuments/dicom-snomed-ct-license-agreement-2021-renewal-v-final-signed.pdf?sfvrsn=228b8bc3_0 and added the link to wikipedia.  

We should definitely think about which terminologies we want to expose and promote in Slicer.

fedorov:
> @fedorov This is a draft pull request, because I assume that these changes should be done somewhere in DCMQI and then the json files need to be regenerated. Can you give advice about what/how to make this changes in DCMQI?

@lassoan I am sorry, I missed this discussion and GitHub notification!

The location of that file in dcmqi is here: https://github.com/QIICR/dcmqi/blob/master/doc/segContexts/SegmentationCategoryTypeModifier-DICOM-Master.json. The idea was that it would reflect the current list of codes in the standard, but there is not process to update this file with the standard releases. If you are not adding any new codes (or are adding only the SNOMED CT codes that are already in DICOM), then it makes sense to make this PR against the dcmqi version of the file. This way it will (I hope) also propagate into the dcmqi helper webapp that I know some users leverage when constructing dcmqi JSON files: https://qiicr.org/dcmqi/#/seg.

fedorov:
@lassoan could you provide the complete list of new SCT codes that you had to add that are not currently included in the DICOM standard? @dclunie is willing to help adding those to the standard.

luzpaz:
Any more progress here ?

lassoan:
I've decided to add the terms to the TotalSegmentator extension. We can think about updating the terminology on Slicer core, but it is not necessary, as any extension can add terminology files. We can revisit this if some project (maybe IDC) provides an updated list then we can update the default terminologies in Slicer, but we don't need to drive these efforts.

Issue #6692, opened 2022-11-22
Fix commit check


jcfr:
## Summary

Check seems to incorrectly report error even when the associated message should be correct. 

## Steps to reproduce

Commit with the following title:

```
BUG: slicer.util.plot: If no title, use column name as series names 
```

is report failure:

```
Error: Error: The first line has to start with the type (e.g "ENH: Add support for awesome feature"). For more details, see https://slicer.readthedocs.io/en/latest/developer_guide/style_guide.html#commits
```

See https://github.com/Slicer/Slicer/actions/runs/3525070397/jobs/5911316779

## Expected behavior

Check should pass

## Environment

NA

jamesobutler:
@jcfr re https://slicer.readthedocs.io/en/latest/developer_guide/style_guide.html#commits, the subject line is not capitalized because after the commit prefix it is `slicer` and not `Slicer`. However this specific commit title is weird in that a specific method of `slicer.util` is being used which is not `Slicer.util`, therefore I would suggest a re-worded message that doesn't include the method as the start of the subject line after the commit prefix.

Here it describes the commit prefix, followed by a colon, followed by a space, followed by a capitalized A-Z character.
https://github.com/Slicer/Slicer/blob/44d626d08b5c675523ce9793c532c647dc9506bf/.github/workflows/commit-message.yml#L21

PR #6729, opened 2022-12-11
BUG: fix incorrect error message

pieper:
Typo in error message makes debugging more difficult


pieper:
I make this one character fix using the github web editor and it seems that the commit message shows up in a way that fails the check.  I think it's actually fine to ignore unless someone knows how to fix it.

jamesobutler:
@pieper re https://slicer.readthedocs.io/en/latest/developer_guide/style_guide.html#commits, the subject line is not capitalized because after the commit prefix it is `fix` and not `Fix`.

Here it describes the commit prefix, followed by a colon, followed by a space, followed by a capitalized A-Z character.
https://github.com/Slicer/Slicer/blob/a943aa850467ba7b2425c236140f72f9d75ae686/.github/workflows/commit-message.yml#L21

`BUG: fix incorrect error message` fails the regex at 
`BUG: f`

jamesobutler:
This confusion was previously reported in https://github.com/Slicer/Slicer/issues/6692, so we could try and improve the example message to better indicate what the regex means. If enforcing the capitalization is not desired, then updating the Slicer docs about conftributing to remove the note that the subject line should be capitalized.

pieper:
Wow, thanks for pointing that out @jamesobutler.  I did not look at the regexp carefully.  I read the error message and also the link to the Slicer documentation and never thought capitalization would be the issue.  The current message is misleading IMO, since it explicitly complains about the "type" being wrong.  I'll make a PR to change the message (and hope I get the commit message formatted correctly!).

image

I'm not sure starting with a capital letter should be a hard requirement but I'll ignore that for now.  If someone will provide the review approval the commit message can be fixed during the squash and merge I believe.

jamesobutler (approved):

PR #6730, opened 2022-12-11
DOC: Improve error message for commit check

pieper:
See:
https://github.com/Slicer/Slicer/pull/6729
and
https://github.com/Slicer/Slicer/issues/6692

### DIFF
--- a/.github/workflows/commit-message.yml
+++ b/.github/workflows/commit-message.yml
@@ -22,7 +22,7 @@ jobs:
           flags: 'gm'
           excludeDescription: 'true' # optional: this excludes the description body of a pull request
           excludeTitle: 'true' # optional: this excludes the title of a pull request
-          error: 'The first line has to start with the type (e.g "ENH: Add support for awesome feature"). For more details, see https://slicer.readthedocs.io/en/latest/developer_guide/style_guide.html#commits'
+          error: 'The first line has to start with a commit prefix, followed by a colon and space, and then followed by a message with a capital letter (e.g "ENH: Add support for awesome feature"). For more details on other requirements, see https://slicer.readthedocs.io/en/latest/developer_guide/style_guide.html#commits'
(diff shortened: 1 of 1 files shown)


.github/workflows/commit-message.yml
```diff
@@ -22,7 +22,7 @@ jobs:
           flags: 'gm'
           excludeDescription: 'true' # optional: this excludes the description body of a pull request
           excludeTitle: 'true' # optional: this excludes the title of a pull request
-          error: 'The first line has to start with the type (e.g "ENH: Add support for awesome feature"). For more details, see https://slicer.readthedocs.io/en/latest/developer_guide/style_guide.html#commits'
+          error: 'The first line has to start with the type followed by a message with a capital letter (e.g "ENH: Add support for awesome feature"). For more details on other requirements, see https://slicer.readthedocs.io/en/latest/developer_guide/style_guide.html#commits'
```
jamesobutler, line 25: Let’s be a little more specific about what “the type” is referring to. I can see others not knowing this is referring to a commit prefix and that there is also a colon and space needed. A regex can be at times confusing to read. 
```suggestion
          error: 'The first line has to start with a commit prefix, followed by a colon and space, and then followed by a message with a capital letter (e.g "ENH: Add support for awesome feature"). For more details on other requirements, see https://slicer.readthedocs.io/en/latest/developer_guide/style_guide.html#commits'
```

jamesobutler: Note that I don’t know if it’s clearly understood that the colon or space is included in the commit message prefix. So I wrote it as though the prefix is simply the “ENH” part

pieper:
Your suggestion is a nice improvement thanks 👍 

It also exposed another odd boundary case.  The automated acceptance of your suggestion generated a non-conforming commit, but since it will go away with the squash it really doesn't matter.  I guess we just ignore it.

jcfr (dismissed):

PR #6731, opened 2022-12-12
BUG: Fix sample data set labels not appearing on large desktop monitors

lassoan:
Fixes the issue of clipped labels in Sample Data module by using the default icon size as a basis, which is more robust than computing it relative to window size.

Before the fix:

image

After the fix:

image

Tested on both large desktop and small laptop screens.

### DIFF
--- a/Modules/Scripted/SampleData/SampleData.py
+++ b/Modules/Scripted/SampleData/SampleData.py
@@ -300,7 +300,11 @@ class SampleDataWidget(ScriptedLoadableModuleWidget):
         iconPath = os.path.join(os.path.dirname(__file__).replace('\\', '/'), 'Resources', 'Icons')
         mainWindow = slicer.util.mainWindow()
         if mainWindow:
-            iconSize = qt.QSize(int(mainWindow.width / 8), int(mainWindow.height / 6))
+            # Set thumbnail size from default icon size. This results in toolbutton size that makes
+            # two columns of buttons fit into the size of the Welcome module's minimum width
+            # on screens with a various resolution and scaling (see qt.QDesktopWidget().size,
+            # desktop.devicePixelRatioF(), qt.QDesktopWidget().physicalDpiX())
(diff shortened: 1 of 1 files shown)


Modules/Scripted/SampleData/SampleData.py
```diff
@@ -300,7 +300,11 @@ def setCategoriesFromSampleDataSources(categoryLayout, dataSources, logic):
         iconPath = os.path.join(os.path.dirname(__file__).replace('\\', '/'), 'Resources', 'Icons')
         mainWindow = slicer.util.mainWindow()
         if mainWindow:
-            iconSize = qt.QSize(int(mainWindow.width / 8), int(mainWindow.height / 6))
+            # Set thumbnail size from default icon size. This results in toolbutton size that makes
+            # two columns of buttons fit into the size of the Welcome module's minimum width
+            # on screens with a various resolution and scaling (see qt.QDesktopWidget().size,
+            # desktop.devicePixelRatioF(), qt.QDesktopWidget().physicalDpiX())
+            iconSize = qt.QSize(int(mainWindow.iconSize.width() * 6), int(mainWindow.iconSize.height() * 4))
```
jcfr, line 307: Thanks for fixing this 👌🙏

Are the multipliers `6` and `4` required?

lassoan: Without the multipliers the thumbnails would be as small as the toolbar icons - too small to be useful:

image

jcfr: I understand now. Thanks for clarifying.

jcfr (approved):

jcfr:
With this change, the rendering is definitely improved on my workstation:

| Before | After |
|---|---|
| image | image |

jcfr:
Increasing the width of the module panel allow to nicely makes use of the available space:

image

lassoan:
Thanks for testing!

PR #6733, opened 2022-12-12
Document NumPyDataLoader for reading .npy and .npz files

dzenanz:
As suggested in forum discussion
https://discourse.slicer.org/t/import-numpy-file-in-as-image-volume/3653/5

Based on:
https://github.com/pieper/SlicerDMRI/blob/9565f89d16cd72618cd87f1c9046542450e4937b/Modules/Scripted/NIfTIFile/NIfTIFile.py
https://github.com/Slicer/Slicer/blob/9390be6bf76e048e6fd384d8bfd607eba2857b5a/Applications/SlicerApp/Testing/Python/SlicerScriptedFileReaderWriterTest.py

### DIFF
--- a/Docs/user_guide/data_loading_and_saving.md
+++ b/Docs/user_guide/data_loading_and_saving.md
@@ -127,6 +127,7 @@ Readers may support 2D, 3D, and 4D images of various types, such as scalar, vect
 - OpenIGTLink extension:
   - **PLUS toolkit configuration file** (.plus.xml): configuration file for real-time data acquisition from imaging and tracking devices and various sensors
 - Sandbox extension:
+  - - [**NumPy array**](https://numpy.org/devdocs/reference/generated/numpy.lib.format.html#npy-format) (.npy, .npz, reading only): read an array of 1 to 5 dimensions into an image. Axis order for 1D-3D: K, J, I; for 4D: K, J, I, component; for 5D: time, K, J, I, component.
   - **Topcon OCT image file** (.fda, reading only)
 
 ### Models


Modules/Scripted/NumpyDataLoader/NumpyDataLoader.py
```diff
@@ -0,0 +1,143 @@
...
+        if not slicer.app.testingEnabled():
+            return False
+
+        return bool(obj.IsA("vtkMRMLScalarVolumeNode"))
+
+    def write(self, properties):
+        return False
+
+
+class NumpyDataLoaderTest(ScriptedLoadableModuleTest):
```
pieper, line 122: It wold be good to have a real test that creates an array, writes it, then confirms it reads back correctly.  Looks like this test will fail.

pieper (dismissed):
I think this looks really handy.

Modules/Scripted/NumpyDataLoader/NumpyDataLoader.py
```diff
@@ -0,0 +1,143 @@
...
+            vtk_image = vtk.vtkImageData()
+            dshape = numpy_image.shape
+            shape3d = dshape[-3:]  # just the last 3 dimensions
+            shape3dr = shape3d[::-1]  # reverse it
+            vtk_image.SetDimensions(shape3dr)
+            vtk_image.AllocateScalars(vtk.VTK_FLOAT, 1)
+            scalar_volume.SetAndObserveImageData(vtk_image)
+
+            node_array = slicer.util.arrayFromVolume(scalar_volume)
+            zero_indices = [0]*(len(dshape) - 3)
```
jcfr, line 81: Could you add a comment explaining why this is done ?

jcfr (commented):

After addressing a few nitpick and adding a test, this should be good for integration.

Additional, the following documentation should be updated.

See https://github.com/Slicer/Slicer/blob/main/Docs/user_guide/data_loading_and_saving.md#images

Modules/Scripted/NumpyDataLoader/NumpyDataLoader.py
```diff
@@ -0,0 +1,143 @@
...
+    def fileType(self):
+        return "numpy"
+
+    def extensions(self, obj):
+        return ["numpy (*.nii.gz)"]
+
+    def canWriteObject(self, obj):
+        # Only enable this writer in testing mode
+        if not slicer.app.testingEnabled():
+            return False
```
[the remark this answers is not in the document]
dzenanz: Numpy format does not have metadata support, I don't think we should help its spread by having regular write support. More in the [forum discussion](https://discourse.slicer.org/t/import-numpy-file-in-as-image-volume/3653/2?u=dzenanz).

Modules/Scripted/NumpyDataLoader/NumpyDataLoader.py
dzenanz: I will add a test in second force-push. In the first, a fix for the rest of the stuff.

dzenanz:
Do you know how to auto-select the just loaded file in user interface? That does not happen for .npy files, like it happens for .png and .nrrd files.

pieper:
> how to auto-select the just loaded file in user interface?

Probably easiest to use this utility to show the images in the background:
https://github.com/Slicer/Slicer/blob/main/Base/Python/slicer/util.py#L514

lassoan:
> > how to auto-select the just loaded file in user interface?
> 
> Probably easiest to use this utility to show the images in the background: https://github.com/Slicer/Slicer/blob/main/Base/Python/slicer/util.py#L514

`setSliceViewerLayers` would work, but in image readars we usually use `applogic.PropagateVolumeSelection`, so for consistency it may be better to use it here, too:

https://github.com/Slicer/Slicer/blob/3d62cbfd81c08e7961d568be18f2bd39a860a543/Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py#L454-L457

lassoan:
I started to add some more comments, but then I rather made the changes myself:
- add support for both .npy and .npz format
- add support for arrays up to 5D (3D spatial + component + time sequence)
- report loading errors to the user

Docs/developer_guide/script_repository/sequences.md
```diff
@@ -1,192 +1,192 @@
-## Sequences
```
jcfr, line 1: Since some of the markdown file currently have `CRLF` endings, we should bulk fix them. Until this is done, we should avoid updating the entire file.

jcfr: See https://github.com/Slicer/Slicer/pull/6734

dzenanz:
Thanks for the improvements @lassoan. You should have added yourself as a module contributor.

In PyTorch, the usual conventions for dimensions are: [NCHW for 4D](https://pytorch.org/docs/stable/generated/torch.nn.Conv2d.html#torch.nn.Conv2d) tensors/arrays and [NCDHW for 5D](https://pytorch.org/docs/stable/generated/torch.nn.Conv3d.html#torch.nn.Conv3d). I assume that [channel-last convention](https://github.com/Slicer/Slicer/blob/5a3f3381d16c7151b695d2597695951869c9981d/Base/Python/slicer/util.py#L2225-L2241) was taken from ITK, or maybe VTK? How many people rely on this? As this is a new addition, we could only use PyTorch's convention in this reader, if we don't want to change this convention in the rest of Slicer.

lassoan:
> In PyTorch, the usual conventions for dimensions are: [NCHW for 4D](https://pytorch.org/docs/stable/generated/torch.nn.Conv2d.html#torch.nn.Conv2d) tensors/arrays and [NCDHW for 5D](https://pytorch.org/docs/stable/generated/torch.nn.Conv3d.html#torch.nn.Conv3d). I assume that [channel-last convention](https://github.com/Slicer/Slicer/blob/5a3f3381d16c7151b695d2597695951869c9981d/Base/Python/slicer/util.py#L2225-L2241) was taken from ITK, or maybe VTK? How many people rely on this? As this is a new addition, we could only use PyTorch's convention in this reader, if we don't want to change this convention in the rest of Slicer.

Yes, this is the ITK/VTK memory layout. It is easy enough to reorder axes before writing to file, so documenting the convention is probably enough (as it is already described in the updated User guide page). If it turns out to be an issue then we could add axis order as loading option (and an options widget for choosing this order in the "Add data" window).

dzenanz:
Even when reordering the axis for channel to be last, with the updated reader, I get an all-black image of "4 components". Visualizing something (vector norm, component average, or even just the first component) would be more useful than an all-black image.

dzenanz:
Also, we could employ heuristic: the dimensions with largest size could be considered to be spatial. So (10, 4, 368, 640) would be interpreted as 4D image with 4 timepoints and 3D size of 640x368x10 (ijk). Prefer timepoints over components unless size is 3, which can be easily visualized as RGB.

dzenanz:
Of course, all of these complications stem from .npy not having suitable metadata.

lassoan:
Having we could have `auto`, `NCDHW`, `NKJIC` axis options. Auto could implement the heuristics you described.

My only concern is that if sooner or later users will report that the loader randomly breaks. We experienced this with ITK, as ITK vector image reader has such heuristic, too. We were shocked when after being annoyed by seemingly random load failures for a long time we realized that loading fails for all data sets that have 3 or 4 time points simply because it is interpreted as channel data instead of time sequence.

Not much better, but maybe a bit more predictable heuristic could be to decide based on (composite) file extension. For example, `something.torch.npy` could make `NCDHW` order the default.

Why don't you save the result using pynrrd? You could use `nrrd.write` and specify the axis kinds like this:

```python
nrrd.write("test.nrrd", a, {'kinds': ['RGB-color', 'domain', 'domain']})
```

How do you load the image into Slicer? Drag-and-drop is pretty tedious and it also shows the loaded image by default (there is no way to update an existing image). You could solve both tedious drag-and-drop and setting axes by using [slicerio](https://github.com/lassoan/slicerio#view-files-in-3d-slicer):

```python
np.save("path/to/myfile.npy", someTensor),
nodeID = slicerio.server.file_load("path/to/myfile.npy", "NumpyArrayFile", {'axes': 'NCDHW'})
```

`slicerio.server.file_load` loads a data set in Slicer (if Slicer is not running then it launches it). If you load the node from a standard file format then you can reload it by simply calling `slicerio.server.node_reload(id=nodeID)`.

pieper:
+1 for being more explicit either in the naming convention or in giving the user the ability to specify the load pattern. I agree that heuristics can be painful.

jcfr:
_Assuming the pull-request was unintentionally closed, I am re-opening it so that we further improve and reach consensus._

pieper:
> _Assuming the pull-request was unintentionally closed, I am re-opening it so that we further improve and reach consensus._

Yes - sorry, wrong button!

dzenanz:
@lassoan It would be good if you could take over this PR, or start a new PR. Implementing `auto`, `TCDHW`, `TKJIC` (T standing for timepoint) axis options would be great. I think you are more familiar with Slicer's internals and UI, so adding this would be a lot easier for you than me.

lassoan:
As it could take some time for this reader to get more mature, it would make sense to put it in the Sandbox extension, where it could be updates more easily and more frequently.

jcfr:
This would be consistent with the existing modules available in [SlicerSandbox](https://github.com/PerkLab/SlicerSandbox):

image

Then, this PR would only update the following files:

```
Base/Python/slicer/util.py
Docs/developer_guide/script_repository/sequences.md
Docs/user_guide/data_loading_and_saving.md
```

dzenanz:
> put it in the Sandbox extension

Agreed. @lassoan or @jcfr, could one of you do it?

jcfr:
I will take care of it.

jcfr:
This pull request should be integrated after the following one have been reviewed and integrated:
* https://github.com/Slicer/Slicer/pull/7142
* https://github.com/PerkLab/SlicerSandbox/pull/19

dzenanz:
Should this PR be merged or closed?

lassoan:
The reader is avaialable in the Sandbox extension (https://github.com/PerkLab/SlicerSandbox/pull/19), so indeed we can close this pull request.

PR #6736, opened 2022-12-14
BUG: Fix reading of default markup active color

lassoan:
Fixes issue reported at https://discourse.slicer.org/t/markups-save-to-defaults-colors-misremembered/26732


jamesobutler (approved):
I have confirmed through testing that this fixes the issue.

PR #6739, opened 2022-12-18
DOC: Clarify description of volume computation in Segment statistics

lassoan:


Docs/user_guide/modules/segmentstatistics.md
```diff
@@ -19,23 +21,38 @@ Labelmap statistics are calculated using the binary labelmap representation of t
 - Principal axes: the principal axes of rotation of the segment
 - Oriented bounding box: the non-axis aligned bounding box that encompasses the segment. Principal axis directions are used to orient the bounding box.
 
-## Scalar volume statistics
+### Scalar volume statistics
+
+The values are computed from the binary labelmap representation of the segment, for the part that verlaps with the chosen scalar volume.
```
pieper, line 26:
```suggestion
The values are computed from the binary labelmap representation of the segment, for the part that overlaps with the chosen scalar volume.
```

Docs/user_guide/modules/segmentstatistics.md
```diff
@@ -19,23 +21,38 @@ Labelmap statistics are calculated using the binary labelmap representation of t
...
+- If primarily the closed surface representation of the segmentation is used (e.g., 3D visualization, 3D printing) then it may be more appropriate to use the `Closed surface` plugin to compute both the volume and surface of the segment. The values are computed from the closed surface representation of the segmentation that is shown in 3D views.
+- If scalar volume input is set in the module then `Scalar volume` plugin computes image intensity statistics for each segment. In this case, using values provided by the `Scalar volume` plugin makes sense. Surface and volume values are computed by the same method as in `Labelmap` plugin, the only difference is that the values are computed for only that part of the segments that overlap with the bounding box of the scalar volume.
+
 ## Related Modules
 
 - [Segmentations](segmentations.md) module allows changing visualization options, exporting/importing segments to/from other nodes (models, labelmap volumes), and moving or copying segments between segmentation nodes.
```
pieper, line 58:
```suggestion
- [Segmentations](segmentations.md) module allows changing conversion options, such as decimation and smoothing when converting from labelmap to closed surface representations, which are mainly for visualization, but can have an impact on some statistics such as volume and surface area.  The Segmentations module can also be used for exporting/importing segments to/from other nodes (models, labelmap volumes), and moving or copying segments between segmentation nodes.
```

pieper (changes requested):
A few small suggestions.

Docs/user_guide/modules/segmentstatistics.md
```diff
@@ -2,43 +2,60 @@
...
+
+- Surface area mm2: the surface area of the segment in mm2
+- Volume mm3: the volume of the segment in mm3
+- Volume cm3: the volume of the segment in cm3
+
+## Frequently asked questions
+
+### What is the difference between the surface and volume values computed by various plugins?
+
+There are several ways of computing volume and surface of a segment. The main difference is what is the representation used (3D binary image or surface mesh) and if the entire segment is used or only the part that overlaps with the selected scalar volume. Difference between the values should be very small, less than one percent, so it usually does not matter which one is used. Usually it does not matter which plugin is used, but to minimize variance between values, it is recommended to consistently use the same plugin within a study.
```
jcfr, line 49:
```suggestion
There are several ways of computing volume and surface of a segment. The main difference relates to the representation being used (3D binary image or surface mesh) and if the entire segment is used or only the part that overlaps with the selected scalar volume. Difference between the values should be very small, less than one percent, so it usually does not matter which one is used. Usually it does not matter which plugin is used, but to minimize variance between values, it is recommended to consistently use the same plugin within a study.
```

Docs/user_guide/modules/segmentstatistics.md
```diff
@@ -2,43 +2,60 @@
...
+## Computed metrics
 
-Labelmap statistics are calculated using the binary labelmap representation of the segment.
+### Labelmap statistics
+
+The values are computed from the binary labelmap representation of the segment.
 
 - Voxel count: the number of voxels in the segment
 - Volume mm3 the volume of the segment in mm3
 - Volume cm3 the volume of the segment in cm3
```
jcfr, line 13:
```suggestion
- Volume mm3: the volume of the segment in mm3
- Volume cm3: the volume of the segment in cm3
```

Docs/user_guide/modules/segmentstatistics.md
```diff
@@ -2,43 +2,60 @@
...
+- If primarily the closed surface representation of the segmentation is used (e.g., 3D visualization, 3D printing) then it may be more appropriate to use the `Closed surface` plugin to compute both the volume and surface of the segment. The values are computed from the closed surface representation of the segmentation that is shown in 3D views.
+- If scalar volume input is set in the module then `Scalar volume` plugin computes image intensity statistics for each segment. In this case, using values provided by the `Scalar volume` plugin makes sense. Surface and volume values are computed by the same method as in `Labelmap` plugin, the only difference is that the values are computed for only that part of the segments that overlap with the scalar volume.
```
jcfr, line 54:
```suggestion
- If the scalar volume input is set in the module then the `Scalar volume` plugin computes image intensity statistics for each segment. In this case, using values provided by the `Scalar volume` plugin makes sense. Surface and volume values are computed by the same method as in `Labelmap` plugin, the only difference is that the values are computed for only that part of the segments that overlap with the scalar volume.
```

Docs/user_guide/modules/segmentstatistics.md
```diff
@@ -2,43 +2,60 @@
...
+
+## Frequently asked questions
+
+### What is the difference between the surface and volume values computed by various plugins?
+
+There are several ways of computing volume and surface of a segment. The main difference is what is the representation used (3D binary image or surface mesh) and if the entire segment is used or only the part that overlaps with the selected scalar volume. Difference between the values should be very small, less than one percent, so it usually does not matter which one is used. Usually it does not matter which plugin is used, but to minimize variance between values, it is recommended to consistently use the same plugin within a study.
+
+How to choose between plugins:
+- `Labelmap` plugin is a good choice for computing volume of a segment for most cases. The volume is computed from the number of voxels multiplied by the volume of a singlevoxel. By default, surface computation in `Labelmap` plugin is disabled - either enable `Surface mm2` measurement in advanced settings to make this plugin compute surface area; or click `Show 3D` button to create closed surface representation for the segment and then get surface area from `Closed surface` plugin.
```
jcfr, line 52:
```suggestion
- The `Labelmap` plugin is a good choice for computing the volume of a segment for most cases. The volume is computed from the number of voxels multiplied by the volume of a singlevoxel. By default, surface computation in `Labelmap` plugin is disabled - either enable `Surface mm2` measurement in advanced settings to make this plugin compute surface area; or click `Show 3D` button to create closed surface representation for the segment and then get surface area from `Closed surface` plugin.
```

PR #6740, opened 2022-12-19
Parameter node wrapper parameter pack

Connor-Bowley:
Adds the ability to make simple dataclass-like classes that can be automatically serialized by the parameterNodeWrapper.

These classes can use the validators.

Example from added documentation:

```py
from slicer.parameterNodeWrapper import *


@parameterPack
class Point:
    x: float
    y: float


@parameterPack
class BoundingBox:
    # can nest parameterPacks
    topLeft: Point
    bottomRight: Point


@parameterNodeWrapper
class ParameterNodeType:
    # can add them to a @parameterNodeWrapper like any other type
    box: Annotated[BoundingBox,
        Default(BoundingBox(Point(0, 1), Point(1, 0)))]


parameterNode = slicer.mrmlScene.AddNewNodeByClass('vtkMRMLScriptedModuleNode')
param = ParameterNodeType(parameterNode)
# can set wholesale
param.box.topLeft = Point(-4, 5)
# or can set piecewise
param.box.bottomRight.x = 4
param.box.bottomRight.y = -5
```


Base/Python/slicer/parameterNodeWrapper/parameterPack.py
```diff
@@ -0,0 +1,285 @@
...
+    Class with the ability to observe any parameterPack and write on change.
+    """
+    def __init__(self, parameterNode, serializer: Serializer, name: str, startingValue):
+        # Because we are overriding __getattr__ and __setattr__ to pass through to _value
+        # we need to access this class's members directly through __dict__
+        self.__dict__['_parameterNode'] = parameterNode
+        self.__dict__['_serializer'] = serializer
+        self.__dict__['_name'] = name
+        self.__dict__['_value'] = startingValue
+
```
allemangD, line 169: _Generally_ I think it's a smell to manually mess with `__dict__`.  I'd prefer to use `super().__setattr__` and `super().__getattr__`; that'd also set you implement `__getattribute__` instead.

Base/Python/slicer/tests/test_slicer_parameter_pack.py
```diff
@@ -0,0 +1,288 @@
...
+
+        pack6 = ParameterPack(y=6)
+        self.assertEqual(pack6.x, 0)
+        self.assertEqual(pack6.y, 6)
+
+        with self.assertRaises(ValueError):
+            pack.x = None
+        with self.assertRaises(ValueError):
+            pack.x = "hi"
+
```
allemangD, line 66: Might be good to also ensure you can't dynamically add an attribute to the pack; `pack.z = 123`.

Base/Python/slicer/tests/test_slicer_parameter_pack.py
```diff
@@ -0,0 +1,288 @@
...
+        pack2 = ParameterPack("words")
+        pack2.i = 6
+        pack2.j = 5
+
+        self.assertEqual(pack, pack2)
+
+        self.assertEqual(str(pack), "mystr")
+        self.assertEqual(repr(pack), "myrepr")
+
+    def test_serialization(self):
```
allemangD, line 135: Similarly, might be good to ensure you can't dynamically add a property to the observed pack; `param.pack.z = 123`.

allemangD:
Just some minor nitpicks. Overall, looks good to me!

Base/Python/slicer/parameterNodeWrapper/parameterPack.py
```diff
@@ -0,0 +1,285 @@
...
+    # required to make this class work, these names are reserved
+    for reserved in ["_parameterPack_allParameters", "_is_parameterPack"]:
+        if reserved in classtype.__dict__:
+            raise ValueError(f"Cannot use reserved name '{reserved}' in a parameterPack")
+    setattr(classtype, "_parameterPack_allParameters", allParameters)
+    setattr(classtype, "_is_parameterPack", True)
+    return classtype
+
+
+def parameterPack(classtype=None):
```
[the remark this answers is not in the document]
Connor-Bowley: Concur, would be good, but not worth bringing in the typing extensions now.

Base/Python/slicer/tests/test_slicer_parameter_pack.py
Connor-Bowley: Good idea, will add.

Base/Python/slicer/tests/test_slicer_parameter_pack.py
Connor-Bowley: Good idea, will add.

Docs/developer_guide/parameter_nodes.md
```diff
@@ -203,7 +203,106 @@ param.value = "d" # ValueError raised
 
 ### Custom Parameter Types
 
-Custom parameters can be supported with a little bit of work.
+#### parameterPack
+
+It is often useful to group related information together in structures with useful names. Another decorator, `@parameterPack` was added to make this easier. This will make the class behave in a similar manner to Python's `@dataclasses.dataclass`.
```
jcfr, line 208: May also be helpful to add a note describing the impact on the actual names of the attributes set in the the `vtkMRMLScriptedModuleNode`

Base/Python/slicer/tests/test_slicer_parameter_pack.py
Connor-Bowley: Upon further consideration, do we want this?

`@dataclasses.dataclass` doesn't do this.

Are you trying to protect from accidents (typos) or something malicious?

Base/Python/slicer/tests/test_slicer_parameter_pack.py
Connor-Bowley: The observed version does prevent this because it overrides `__setattr__`. I can see an argument that this makes sense because setting a new attribute is a change that can't be observed.

Dynamically adding on the non-observed parameterPack may still be ok to allow because it follows dataclass's lead and is a somewhat more generic structure than the observed side.

Base/Python/slicer/tests/test_slicer_parameter_pack.py
allemangD: I don't think it's a good thing to support... dataclass does let you assign new attributes but `dataclasses.asdict`, `dataclasses.update`, etc. and the generated methods like `__str__`, `__eq__`, etc. don't see them. I'd seriously question any code that does this; it probably needs a dictionary, not a dataclass.

The value that `@parameterPack` adds over `@dataclass` is its validation and serialization features; supporting dynamic attributes seems problematic at best. You can still put a `dict[str, T]` in a `parameterPack` so I don't think there's any missing functionality.

Connor-Bowley: There is difference between supporting and explicitly disallowing. The problem with explicitly disallowing things in Python is that you can't.

I can override `__setattr__` to disallow dynamic addition, but it would add complexity. If I don't disallow it, I would still say it is not supported and people shouldn't do it. Even if I did override `__setattr__`, someone could still do something like `object.__setattr__(pack, "dynamicAdd", someValue)`.

So the question is, is the added complexity worth it. If your concern is people breaking the intended use, I think it is not worth it because people can do that anyway. If the concern is protecting someone typing `pack.valeu = 1` instead of `pack.value = 1`, then maybe (although that is a problem everywhere else in Python too).

allemangD: I see. Probably not worth adding any extra complexity in `__setattr__`; maybe just let it be something that's implicitly _possible_, but not encouraged or guaranteed.

Docs/developer_guide/parameter_nodes.md
Connor-Bowley: Added comment. Updated the documented code block to further explain. Added test to enforce naming.

Connor-Bowley: Also updated to separate by a dot instead of an underscore, which is probably more idiomatic.

Base/Python/slicer/parameterNodeWrapper/parameterPack.py
Connor-Bowley: Updated to use super.

Docs/developer_guide/parameter_nodes.md
```diff
@@ -203,7 +203,106 @@ param.value = "d" # ValueError raised
...
+    # can nest parameterPacks
+    topLeft: Point
+    bottomRight: Point
+
+@parameterNodeWrapper
+class ParameterNodeType:
+    # can add them to a @parameterNodeWrapper like any other type
+    box: Annotated[BoundingBox,
+        Default(BoundingBox(Point(0, 1), Point(1, 0)))]
+
```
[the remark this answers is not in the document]
Connor-Bowley: Fixed.

Docs/developer_guide/parameter_nodes.md
```diff
@@ -203,7 +203,106 @@ param.value = "d" # ValueError raised
...
+class Point:
+    x: float
+    y: float
+
+@parameterPack
+class BoundingBox:
+    # can nest parameterPacks
+    topLeft: Point
+    bottomRight: Point
+
```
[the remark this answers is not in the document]
Connor-Bowley: Fixed.

Docs/developer_guide/parameter_nodes.md
```diff
@@ -203,7 +203,106 @@ param.value = "d" # ValueError raised
...
+It is often useful to group related information together in structures with useful names. Another decorator, `@parameterPack` was added to make this easier. This will make the class behave in a similar manner to Python's `@dataclasses.dataclass`.
+
+```py
+from slicer.parameterNodeWrapper import *
+
+@parameterPack
+class Point:
+    x: float
+    y: float
+
```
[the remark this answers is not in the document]
Connor-Bowley: Fixed.

Docs/developer_guide/parameter_nodes.md
```diff
@@ -203,7 +203,106 @@ param.value = "d" # ValueError raised
...
 ### Custom Parameter Types
 
-Custom parameters can be supported with a little bit of work.
+#### parameterPack
+
+It is often useful to group related information together in structures with useful names. Another decorator, `@parameterPack` was added to make this easier. This will make the class behave in a similar manner to Python's `@dataclasses.dataclass`.
+
+```py
+from slicer.parameterNodeWrapper import *
+
```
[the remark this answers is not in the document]
Connor-Bowley: Fixed.

Connor-Bowley:
Squashed fixup commits and rebased off main.

allemangD (approved):
My previous comments have been addressed, so LGTM! :rocket:

Issue #6741, opened 2022-12-19
Make Slicer API compatible with pathlib


lassoan:
## Is your feature request related to a problem? Please describe.

Developers would like to use pathlib.Path directly as input for Slicer API calls where a path-like object is expected.

## Describe the solution you'd like

Allow developers to pass paths as pathlib.Path object.

## Additional context

https://discourse.slicer.org/t/bug-slicer-util-loadvolume-crashes-when-given-a-pathlib-path-instead-of-a-string/26820

PR #6743, opened 2022-12-21
ENH: Use objects with __str__ as a filename

AlistairMcCutcheon:
Issue: https://github.com/Slicer/Slicer/issues/6741
Topic: https://discourse.slicer.org/t/bug-slicer-util-loadvolume-crashes-when-given-a-pathlib-path-instead-of-a-string/26820


jamesobutler:
@lassoan Warning: about having picked `ENH: Add NumpyDataLoader for reading .npy and .npz files` over into this branch

lassoan (approved):

Issue #6207, opened 2022-02-18
Relocate migration guide to the Slicer wiki


jcfr:
## Is your feature request related to a problem? Please describe.

The `MigrationGuide` wiki page documents list of breaking API changes and how to update code to accommodate these.

See https://www.slicer.org/wiki/Documentation/Nightly/Developers/Tutorials/MigrationGuide

When possible, associated errors are also documented along with link to commit implementing the relevant changes.

Considering the wiki is in "life support" mode and not actively maintained, we need to ensure the information remains documented and discoverer-able using search engines.

## Describe the solution you'd like

The proposal is to keep a similar organization creating the following pages on the wiki:

```
Migration Guide
 |----- Slicer 5.0: API changes since 4.10
 |----- Slicer 4.10: API changes since 4.8
 |----- Slicer 4.8: API changes since 4.0
 |----- Slicer Extension updates
 |----- Obsolete Code Removal
 |----- Transition from ITK4 to ITK5
 |----- Transition from VTK 8.0 to VTK 9.0
 |----- Transition from VTK7-Qt4 to VTK8-Qt5
 |----- Transition from VTK6 to VTK7
 |----- Transition-from-VTK5 to VTK6
```

with the following differences:

* Organize changes with the section title `Slicer X.Y: API changes since X.Y-1` instead of specific `Slicer X.Y: Do this instead of that` top-level page for each change

* Consolidation of pages starting with `Slicer X.Y: ...` in relevant section

* Rename `Obsolete Code Removal` to `Code Modernization & Obsolete Code Removal`

### Page names

The page will **not** be prefixed with their parent page name.

This means the pages will have the following URLs and page names:

```
https://github.com/Slicer/Slicer/wiki/MigrationGuide
...
https://github.com/Slicer/Slicer/wiki/Slicer-4.10:-API-changes-since-4.8
...
https://github.com/Slicer/Slicer/wiki/Transition-from-VTK5-to-VTK6
```

and **NOT** these

```
https://github.com/Slicer/Slicer/wiki/MigrationGuide
...
https://github.com/Slicer/Slicer/wiki/MigrationGuide-Slicer-4.10:-API-changes-since-4.8
...
https://github.com/Slicer/Slicer/wiki/MigrationGuide-Transition-from-VTK5-to-VTK6
```

### Reference to modernization tools and scripts

These should be referenced in the `Code Modernization & Obsolete Code Removal` section.

## Describe alternatives you've considered

[...]

PR #6750, opened 2022-12-23
ENH: Change organization name and domain to Slicer from NA-MIC

### LINKED ISSUE #6722: CPack always uses NA-MIC as the vendor for extension packages
## Summary

https://github.com/Slicer/Slicer/blob/53625c29ec8c212a91ba34750e01ea2a4b25272e/CMake/SlicerExtensionCPack.cmake#L126
CPack's NuGet generator, for example, uses this value for the `<authors>` property in the resulting package. Since `CPackConfig.cmake` is also written from that file, this setting can't be easily replaced.

## Expected behavior

Either `${EXTENSION_CONTRIBUTORS}` or the replaceable `${Slicer_ORGANIZATION_NAME}` seem to make more sense in this context.

jamesobutler:
See discussion at https://discourse.slicer.org/t/inconsistency-between-organization-name-used-for-settings-files-and-for-macos-bundle-identifier/26547

This also closes #6722


jamesobutler:
Installed programs list:
image
Installed to user location:
image
Installed user settings location:
image

jamesobutler:
Would migration of user settings being required from the NA-MIC location to the new Slicer organization location?

lassoan (approved):
I don't think it is a big issue if we don't bring over the settings from the old location, and it is a bit simpler and cleaner. The main problem I see with copying the old settings that it would be complicated to reset the application settings to the default (deleting Slicer.ini would not actually reset to the application default but it would reset to the content in the NA-MIC folder).

Since we recently released a new patch release (Slicer-5.2.2) I think this is a good time to make this change.

@jamesobutler Please update the migration guide and write a post on the Slicer forum after this is integrated (people may need to update their scripts).

jcfr:
> deleting Slicer.ini would not actually reset to the application default but it would reset to the content in the NA-MIC folde

If we want to support this, after a first import, the original settings in NA-MIC folder would be:
1. updated to indicate that the settings have already been migrated (e.g writing in the application revision)
2. or deleted if the user choose to delete these after the first import

This would address the issue outlined by @lassoan 

That said, this would make the settings logic more complex so moving forward with approach suggested above is sensible.

jamesobutler:
You all still want to have the Migration Guide on the slicer.org/wiki (https://www.slicer.org/wiki/Documentation/Nightly/Developers/Tutorials/MigrationGuide) or should it be migrated to the GitHub repository like https://github.com/InsightSoftwareConsortium/ITK/blob/master/Documentation/ITK5MigrationGuide.md considering developers are likely the individuals writing the updates?

Based on the ordering, would a new Slicer 5.3 prefixed entry be placed at position 1.2 as the most recent 5.3 change? Is the 1.1 Slicer 5 changes since 4.10 section still needing to be first entry instead of chronologically mentioned?
image

jcfr:
> re: migration

See https://github.com/Slicer/Slicer/issues/6207

jamesobutler:
Migration guide updated: https://www.slicer.org/wiki/Documentation/Nightly/Developers/Tutorials/MigrationGuide/Slicer#Slicer_5.3:_Organization_name_and_domain_changed_from_NA-MIC_to_Slicer

Discourse post:
https://discourse.slicer.org/t/organization-name-and-domain-changed-from-na-mic-to-slicer/28173

PR #6752, opened 2022-12-26
BUG: Fix SlicerRequestHandler return value when reader does not return any node ID

lassoan:
When Slicer web API was used for loading nodes but the reader did not return any node ID then the API call returned with error.


pieper (approved):
Looks fine to me - merge whenever you are ready.

lassoan:
Thanks for the review!

Issue #6732, opened 2022-12-12
Failing py_SubjectHierarchyGenericSelfTest and qMRMLLayoutManagerTest2 tests


jamesobutler:
## Summary

Starting on December 9th py_SubjectHierarchyGenericSelfTest and qMRMLLayoutManagerTest2 began to fail across all platforms https://slicer.cdash.org/viewTest.php?buildid=2877851&onlyfailed

```
======================================================================
FAIL: test_SubjectHierarchyGenericSelfTest_FullTest1 (SubjectHierarchyGenericSelfTest.SubjectHierarchyGenericSelfTestTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "D:/D/P/S-0-build/Slicer-build/lib/Slicer-5.3/qt-scripted-modules/SubjectHierarchyGenericSelfTest.py", line 111, in test_SubjectHierarchyGenericSelfTest_FullTest1
    self.section_LoadScene()
  File "D:/D/P/S-0-build/Slicer-build/lib/Slicer-5.3/qt-scripted-modules/SubjectHierarchyGenericSelfTest.py", line 384, in section_LoadScene
    self.assertEqual(slicer.mrmlScene.GetNumberOfNodesByClass('vtkMRMLModelNode'), 4)  # Including the three slice view models
AssertionError: 7 != 4
```

The 3 commits from the previous day. One of these caused these tests to begin failing:
- https://github.com/Slicer/Slicer/commit/8d1ff646c59379c9f13e90b026bef1a942e8611f
- https://github.com/Slicer/Slicer/commit/2ed9451921d937f2a73d846bd1bf19b956478492
- https://github.com/Slicer/Slicer/commit/5a3f3381d16c7151b695d2597695951869c9981d

## Environment
- Slicer version: Slicer-5.3.0-2022-12-09
- Operating system: Windows / Linux / Mac

lassoan:
Fixed by commits:
- https://github.com/Slicer/Slicer/pull/6755/commits/c6f55e26ba7eee86899ad09d682c59b27d1df3aa
- https://github.com/Slicer/Slicer/pull/6755/commits/d6cf0fdee97d3e43bb25c4a65c617a28d0be95bc

PR #6755, opened 2022-12-28
Fix 2 failing tests

lassoan:
Fixes the two failing tests. Both tests were due to subtle errors introduced by the recently added vertical slice offset slider feature (https://github.com/Slicer/Slicer/commit/2ed9451921d937f2a73d846bd1bf19b956478492).

Also fixed a extension list difference reporting when an incompatible scene is loaded.


jamesobutler:
This will close #6732

pieper (approved):
This looks fine to me.  I don't know why it fixes the tests, but that doesn't really matter.  I'm okay with merging either before or after the build finishes.

lassoan:
Thanks for the review!

PR #6757, opened 2022-12-29
Parameter node wrapper support for typing.Union

Connor-Bowley:
Adds support for `typing.Union`.
Uses better suited exception types.
Minor cleanup.

### DIFF
--- a/Base/Python/slicer/parameterNodeWrapper/serializers.py
+++ b/Base/Python/slicer/parameterNodeWrapper/serializers.py
@@ -9,6 +9,7 @@ import slicer
 from .util import splitAnnotations
 from .validators import (
     extractValidators,
+    IsNone,
     NotNone,
     IsInstance,
 )
@@ -26,6 +27,7 @@ __all__ = [
     "ListSerializer",
     "TupleSerializer",
     "DictSerializer",
+    "UnionSerializer",
 ]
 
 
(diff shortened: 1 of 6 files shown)


Base/Python/slicer/parameterNodeWrapper/serializers.py
```diff
@@ -819,3 +821,147 @@ def supportsCaching(self):
...
+                else:
+                    if raisedException is not None:
+                        raise raisedException
+                    raise TypeError(f"Unable to write value '{value}' of type '{type(value)}'")
+            except Exception:
+                if wasIn:
+                    self.write(parameterNode, name, oldValue)
+                else:
+                    self.remove(parameterNode, name)
+                raise
```
allemangD, line 944: Really not a fan of this logic... Nested `try`-`except Exception` seems problematic.

I'm not certain that `wasIn`, `oldValue`, `raisedException` are strictly necessary. `validatedSerializer.write` will not be called unless `validatedSerializer.validate()` succeeds; and if an error occurs in `validatedSerializer.write` then it seems appropriate to let that pass through.

If I understand the API correctly, this is equivalent:

```python
def write(self, parameterNode, name: str, value) -> None:
    with slicer.util.NodeModify(parameterNode):
        for index, serializer in enumerate(self._serializers):
            if not serializer.canSerialize(type(value)):
                continue

            try:
                serializer.validate(value)
            except Exception:
                continue

            self.remove(parameterNode, name)
            serializer.write(parameterNode, self._paramName(name, index), value)

            return

    raise TypeError(f"Unable to write value {value!r} of type {type(value)}")
```

It might be appropriate to _require_ `validate()` to raise `ValueError` for failed validation as part of the API, else the exception is regarded as a true error and is not caught.

Could also let `validate()` return `bool` and use it as a predicate.

I'm not sure of the best pattern to choose, since exceptions allow to attach some error message; either way, it'd still be nice to tighten or remove those `except` clauses.

allemangD: The code snippet I included is _not_ equivalent; I'll look at it some more and update if I come up with a better approach.

allemangD: The code above fails tests because they expect `TypeError` to be raised if all `canSerialize` fail; but they expect `ValueError` to be raised if `validate` fails.

It's unclear what to expect if _multiple_ `validate()` fail. We do not have `ExceptionGroup` in Python 3.9.

allemangD: Discussed offline with Connor; there are enough edge cases that the fundamental architecture will be unchanged, but add a separate method for encapsulation.

lassoan (approved):

It would be nice if it was used somewhere in a Slicer core module. If you cannot make it easily usable in any of the Python scripted modules in Slicer core then maybe you could add a link to a module in an extension where it is going to be used.

Connor-Bowley:
I am eventually going to update the scripted module template in the extension wizard to use the parameter node wrapper. We could try to come up with a good example in there for using the different aspects of the parameter node wrapper.

A concern with updating existing core modules is that it would likely break the ability to load old scene files for non-trivial use cases (as each of those use cases had their own way of storing lists / differentiating types / etc).

Note that VectorToScalarVolume uses the parameterNodeWrapper, but not Unions specifically.

Connor-Bowley:
Per offline discussion with @allemangD around https://github.com/Slicer/Slicer/pull/6757#discussion_r1063576230, separated the finding of the best serializer in the union into its own function for readability.

allemangD (approved):

PR #6758, opened 2022-12-29
Volumes module Display options improvements and fixes

### LINKED ISSUE #6738: Volumes module w/l lock does not lock w/l slider
## Summary

In Slicer 4.11.20210226, the Window/Level lock in the Volumes module would disable all presets and the window/level mode option and values. You could not drag the slider when it was locked.

image

Starting in Slicer 5.0.3, the Window/Level lock in the Volumes module still disables all presets, but it does not disable the window/level mode option or spinboxes anymore. You can successfully drag the Window/Level slider to change the Window/Level even when in the locked state. This was likely due to the changes made in https://github.com/Slicer/Slicer/commit/1af95e000940297c5fada31251c82f3a34264960 or associated commits from the originating PR.

image

## Steps to reproduce

1. Open Slicer and load MRHead Sample data set
2. Switch to Volumes module
3. Press the "Lock" window/level button
4. Change the window/level slider and observe that window/level can still be changed.

## Expected behavior

While in the locked mode, the window/level mode and spinboxes to changes the values should be disabled to prevent moving the slider which changes the window/level. This would restore the behavior as it was in Slicer 4.11.20210226 and earlier.

## Environment
- Slicer version: Slicer 5.2.1
- Operating system: Windows

jamesobutler:
| Slicer 5.2.1 | This PR |
|--------|---------|
|image|image|

- Display presets now have a group section which is titled to indicate they are presets. The section is moved to the top as the presets change multiple display settings below (window/level/colormap)
- Window/Level locked state removed
- Window/Level widget reorganized with familiarity of Threshold widget
- Window/Level and Threshold widgets contained in a groupbox for improved clarification about which objects are for W/L vs objects for Threshold.
- Additional tooltips added to widgets in the Display section so users don't have to go to readthedocs to read more details about what they do

------------------------------------------
- `ENH: Organize Window/Level widget similar to Threshold widget` and `ENH: Organize Volumes module W/L and threshold widgets into groupboxes` changes:

| Slicer 5.2.1 | This PR |
|-------|---------|
| imageimage|imageimage|

- `ENH: Remove window level locked functionality` closes #6738

There was a bug with the window/level lock button not disabling the window/level slider due to changes made in https://github.com/Slicer/Slicer/commit/1af95e000940297c5fada31251c82f3a34264960 and associated commits. Rather than restoring this disabling functionality, the window/level lock functionality is being removed here.

Locking was originally added when left-click-and-drag used to always change Window/Level. However with the introduction of mouse modes there is no longer accidental changing of window/level with the mouse.

### DIFF
--- a/Libs/MRML/Widgets/Resources/UI/qMRMLWindowLevelWidget.ui
+++ b/Libs/MRML/Widgets/Resources/UI/qMRMLWindowLevelWidget.ui
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>517</width>
-    <height>55</height>
+    <width>300</width>
+    <height>61</height>
    </rect>
   </property>
   <property name="sizePolicy">
@@ -19,46 +19,27 @@
   <property name="windowTitle">
    <string>Window Level</string>
   </property>
-  <layout class="QGridLayout" name="gridLayout" columnstretch="0,0,1,0,0">
+  <layout class="QGridLayout" name="gridLayout">
    <property name="margin">
     <number>0</number>
    </property>
    <item row="0" column="0">
-    <widget class="ctkDoubleSpinBox" name="WindowSpinBox">
-     <property name="prefix">
-      <string>W: </string>
-     </property>
-     <property name="decimals">
-      <number>2</number>
-     </property>
-     <property name="decimalsOption">
-      <set>ctkDoubleSpinBox::DecimalsByKey|ctkDoubleSpinBox::DecimalsByShortcuts</set>
+    <widget class="QLabel" name="label">
+     <property name="text">
+      <string>Window/Level:</string>
      </property>
     </widget>
    </item>
    <item row="0" column="1">
-    <widget class="ctkDoubleSpinBox" name="MinSpinBox">
-     <property name="prefix">
-      <string>Min:</string>
-     </property>
-     <property name="decimals">
-      <number>2</number>
-     </property>
-     <property name="decimalsOption">
-      <set>ctkDoubleSpinBox::DecimalsByKey|ctkDoubleSpinBox::DecimalsByShortcuts</set>
-     </property>
-    </widget>
-   </item>
-   <item row="0" column="2">
     <widget class="QComboBox" name="AutoManualComboBox">
      <item>
       <property name="text">
-       <string>Auto W/L</string>
+       <string>Auto</string>
       </property>
      </item>
      <item>
       <property name="text">
-       <string>Manual W/L</string>
+       <string>Manual</string>
       </property>
      </item>
      <item>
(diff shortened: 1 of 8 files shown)


pieper (approved):
These look like nice improvements to me - thanks! 👍

lassoan:
Looks good to me, too. Thank you!

For the future, my preference would be to squash something like this into 2 commits (1. remove lock W/L feature; 2. GUI improvement). I don't really have any strong reasons against having many small commits other than it is unnecessary, as these commits were not intended to be used separately, they were not tested independently, and they just generate noise (some more commit comments to read through).

lassoan:
@jamesobutler if you don't mind, it would be nice if you could add reference to https://github.com/Slicer/Slicer/issues/6738 in the relevant commit and squash at least the code change commit with the corresponding user manual change. Thank you!

jamesobutler:
@lassoan I've added #6738 to the relevant `ENH: Remove Window/Level locked functionality` commit.

> and squash at least the code change commit with the corresponding user manual change.

I'm not sure which code change and corresponding user manual change you are referring to. `ENH: Group and title volume display presets` which moved the location of the presets includes the user manual change in the commit already. `DOC: Fix outdated window/level range control instructions` is a doc fix that could be its own PR as it is fixing info that became outdated a while ago. None of the new changes in this PR to the Window/Level widget caused the need for this change.

jamesobutler:
`ENH: Organize Window/Level widget similar to Threshold widget` I had kept this as a separate commit just because I know qMRMLWindowLevelWidget is used in various places outside of just the Volumes module so I didn't want to include it in GUI updates with say the reorganization of the Volume Display presets. If so desired, `ENH: Group and title volume display presets`, `ENH: Organize Volumes module W/L and threshold widgets into groupboxes` and `ENH: Add tooltips to Volumes module display options` could be squashed into a single commit about the Volumes Display section GUI update. Let me know if this is desired.

jcfr:
Thanks @jamesobutler 

On my side, no issues having independent commits.

jcfr (approved):

lassoan:
Thank you for the explanations @jamesobutler, if there is a specific reason not to squash commits then it's OK.

Libs/MRML/Widgets/Resources/UI/qMRMLWindowLevelWidget.ui
```diff
@@ -19,46 +19,27 @@
...
-      <string>W: </string>
-     </property>
-     <property name="decimals">
-      <number>2</number>
-     </property>
-     <property name="decimalsOption">
-      <set>ctkDoubleSpinBox::DecimalsByKey|ctkDoubleSpinBox::DecimalsByShortcuts</set>
+    <widget class="QLabel" name="label">
+     <property name="text">
+      <string>Window/Level:</string>
```
jamesobutler, line 29: @lassoan Regarding this new string and the new tooltips added in this PR, am I missing anything related to marking it for language translating? I just realized that I didn’t do anything for it. Do I need to do anything for strings in .ui files?

lassoan: All strings properties in .ui files are translatable by default, unless marked explicitly as non-translatable. So, there is probably nothing to do. (Other than updating the translation source files on Weblate - that I'll do in a few days).

jamesobutler: Ok thanks for the info!

PR #6759, opened 2022-12-31
BUG: Fix more than one rotation slider can be non-zero

### LINKED ISSUE #6735: More than one rotation slider in transforms module can be non-zero
## Summary

Originally mentioned at https://discourse.slicer.org/t/more-than-one-rotation-slider-in-transforms-module-can-be-non-zero/26713

## Steps to reproduce

1. Go to Transforms module and add a new linear transform node
2. Change the LR, PA, IS slider value and observe that upon changing one it resets the others
3. Change the LR. PA, IS spinbox values directly and observe that more than one rotation slider can be non-zero
4. Change the LR, PA, IS slider value and observe that more than one rotation slider can be non-zero.

https://user-images.githubusercontent.com/15837524/207636840-7c4541a7-cc8f-4c7c-b371-72d17fd7cbdc.mp4

## Expected behavior

Only one rotation slider can be non zero (as according to the Slicer documentation)

https://slicer.readthedocs.io/en/latest/user_guide/modules/transforms.html#edit

> Note: Linear transform edit sliders only show relative translation and rotation because a transformation can be achieved using many different series of transforms. To make this clear to users, only one transform slider can be non-zero at a time (all previously modified sliders are reset to 0 when a slider is moved). The only exception is translation sliders in “translate first” mode (i.e., when translation in global/local coordinate system button is not depressed): in this case there is a only one way how a specific translation can be achieved, therefore transform sliders are not reset to 0.

## Environment
- Slicer version: Slicer-5.2.1
- Operating system: Windows

jamesobutler:
This closes #6735.

Changing spinbox values of sliders was adding duplicate references into the ActiveSliders QStack.

The expected behavior from the documentation is mentioned below and it was previously not being respected.

https://slicer.readthedocs.io/en/latest/user_guide/modules/transforms.html#edit

> "Note: Linear transform edit sliders only show relative translation and rotation because a transformation can be achieved using many different series of transforms. To make this clear to users, only one transform slider can be non-zero at a time (all previously modified sliders are reset to 0 when a slider is moved). The only exception is translation sliders in “translate first” mode (i.e., when translation in global/local coordinate system button is not depressed): in this case there is a only one way how a specific translation can be achieved, therefore transform sliders are not reset to 0."

@lassoan The issue detailed by #6735 appears to have been primarily the result of the changes made in your bug fix commit https://github.com/Slicer/Slicer/commit/48172a2034ef6069844c24f1634403f246ff1bfa. It would be great if you could confirm this bug fix is appropriate.

### DIFF
--- a/Libs/MRML/Widgets/qMRMLTransformSliders.cxx
+++ b/Libs/MRML/Widgets/qMRMLTransformSliders.cxx
@@ -22,7 +22,7 @@
 #include "ui_qMRMLTransformSliders.h"
 
 // Qt includes
-#include <QStack>
+#include <QSet>
 
 // qMRML includes
 #include <qMRMLUtils.h>
@@ -47,7 +47,7 @@ public:
 
   int                                    TypeOfTransform;
   vtkMRMLTransformNode*                  MRMLTransformNode;
-  QStack<qMRMLLinearTransformSlider*>    ActiveSliders;
+  QSet<qMRMLLinearTransformSlider*>    ActiveSliders;
 };
 
 // --------------------------------------------------------------------------
@@ -485,7 +485,6 @@ void qMRMLTransformSliders::onSliderPositionChanged(double position)
   qMRMLLinearTransformSlider* slider =
     qobject_cast<qMRMLLinearTransformSlider*>(this->sender());
   Q_ASSERT(slider);
-  d->ActiveSliders.push(slider);
   QWidget* focusWidget = this->focusWidget();
 
   // If update initiated from spinbox, consider it active, too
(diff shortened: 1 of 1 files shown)


lassoan (commented):
Thanks a lot for working on this. The changes look good to me. I'm just wondering if using a QSet instead of a QStack would be more appropriate. Duplicate entries would be inherently prevented.

jamesobutler:
Yes I wasn't sure why QStack was used in the first place as part of https://github.com/Slicer/Slicer/commit/0fbe01a6da60bf31d9f86075b2984990f4491596 so I had originally avoided changing the type. 

I've pushed an update to switch to QList which appears to work as well.

lassoan (approved):
Great, thank you! I'll merge when the tests are completed.

jamesobutler:
Status checks have passed so I will proceed with merging now.

PR #6760, opened 2023-01-03
ENH: Replace `lib` by `CMAKE_INSTALL_LIBDIR`

RafaelPalomar:
~~This commit modifies `SlicerDirectories.cmake` to add a new CMake variable called `Slicer_LIB_PREFIX`. This variable allows users to specify the directory prefix for installation of 3D Slicer llibraries (e.g., lib or lib64). By default, the value of this variable is set to `lib`, which preserves the previous hard-coded value.~~

This commit modifies `SlicerDirectories.cmake` to replace the hardcoded `lib` part of the library installation path by `CMAKE_INSTALL_LIBDIR`. The use of `CMAKE_INSTALL_LIBDIR` is documented in the following
references:

  - https://cmake.org/cmake/help/latest/command/install.html?highlight=cmake_install_libdire#installing-directories
  - https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html#module:GNUInstallDirs

This is useful when building/installing slicer with `Slicer_SUPERBUILD=OFF`

### DIFF
--- a/CMake/SlicerDirectories.cmake
+++ b/CMake/SlicerDirectories.cmake
@@ -52,11 +52,19 @@ if(NOT DEFINED Slicer_MAIN_PROJECT_APPLICATION_NAME)
 endif()
 
 #-----------------------------------------------------------------------------
+# Defaults
+#-----------------------------------------------------------------------------
+if(NOT DEFINED CMAKE_INSTALL_LIBDIR)
+  set(CMAKE_INSTALL_LIBDIR "lib")
+endif()
+mark_as_superbuild(CMAKE_INSTALL_LIBDIR:PATH)
+
+#-----------------------------------------------------------------------------
 # Slicer relative directories
 #-----------------------------------------------------------------------------
 # for build tree
 set(Slicer_BIN_DIR "bin")
-set(Slicer_LIB_DIR "lib/${Slicer_MAIN_PROJECT_APPLICATION_NAME}-${Slicer_VERSION_MAJOR}.${Slicer_VERSION_MINOR}")
+set(Slicer_LIB_DIR "${CMAKE_INSTALL_LIBDIR}/${Slicer_MAIN_PROJECT_APPLICATION_NAME}-${Slicer_VERSION_MAJOR}.${Slicer_VERSION_MINOR}")
 set(Slicer_INCLUDE_DIR "include/${Slicer_MAIN_PROJECT_APPLICATION_NAME}-${Slicer_VERSION_MAJOR}.${Slicer_VERSION_MINOR}")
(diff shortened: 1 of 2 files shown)


CMake/SlicerDirectories.cmake
```diff
@@ -51,12 +51,18 @@ if(NOT DEFINED Slicer_MAIN_PROJECT_APPLICATION_NAME)
   set(Slicer_MAIN_PROJECT_APPLICATION_NAME ${${Slicer_MAIN_PROJECT}_APPLICATION_NAME})
 endif()
 
+# This opens the possibility to define alternative lib directories
+# (e,g., lib64/ comonly used in contemporary GNU/Linux sytems).
```
jamesobutler, line 55:
```suggestion
# (e.g., lib64/ commonly used in contemporary GNU/Linux systems).
```

RafaelPalomar: @jamesobutler, thanks for having a look.

jamesobutler:
Should we use the same variable name as ITK for this (`CMAKE_INSTALL_LIBDIR`) re what was done in https://github.com/InsightSoftwareConsortium/ITK/commit/5b09d546666f6110250aa7edbb59d4469791b944 and mentioned in this [comment](https://github.com/InsightSoftwareConsortium/ITK/pull/3522#issuecomment-1203264286) directed at @jcfr for changes in the context of Slicer.

RafaelPalomar:
> Should we use the same variable name as ITK for this (`CMAKE_INSTALL_LIBDIR`) re what was done in [InsightSoftwareConsortium/ITK@5b09d54](https://github.com/InsightSoftwareConsortium/ITK/commit/5b09d546666f6110250aa7edbb59d4469791b944) and mentioned in this [comment](https://github.com/InsightSoftwareConsortium/ITK/pull/3522#issuecomment-1203264286) directed at @jcfr for changes in the context of Slicer.

@jamesobutler, that makes sense. @jcfr, @pieper any comment?

pieper:
> any comment?

Consistency with existing convention for the same concept makes good sense to me.

RafaelPalomar:
As suggested by @jamesobutler, I have simply replaced `lib` by `CMAKE_INSTALL_LIBDIR`. The default value of the variable is `lib` so, in practice, this should work seamlessly. I have also tried a non-superbuild where I change the variable and seems to work fine too. All of this has been tried on a GNU/Linux system.

jcfr:
@RafaelPalomar  This is ready for final review and testing.

I suggest to build application with SuperBuild set to `ON` and `CMAKE_INSTALL_LIBDIR` set to `mylib`. Once build, I expect to Slicer to start from the build tree. Next, it would be nice to also test after generating a package.

RafaelPalomar:
@jcfr, I have tested this PR:

  - Slicer runs both on build tree and on the package generated by CPack
  - Setting the `CMAKE_INSTALL_LIBDIR` has the desired effect
  
Thanks a lot!

jcfr:
Thanks for testing 🙏

PR #6761, opened 2023-01-04
ENH: Create Windows Desktop shortcut during install

jamesobutler:
On the Windows platform, running a Slicer installer would create Start Menu shortcuts but did not create a desktop shortcut which is pretty common. This PR will create a desktop shortcut as part of the install process.

Note: The Start Menu folder MUI Page has a "Do not create shortcuts" option which will skip start menu and desktop shortcut creation
image

### DIFF
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -14,3 +14,4 @@ jobs:
           check_filenames: true
           skip: ".git,*.crt,*.svg,*.vtp,*DICOM-Master.json,*SlicerGeneralAnatomy.json,./CMakeLists.txt.user,./CMake/CTestCustom.cmake.in,./License.txt,./COPYRIGHT.txt,./Resources/*.h,./Base/Logic/vtkSlicerApplicationLogicRequests.h,./Base/QTCLI/vtkSlicerCLIModuleLogic.cxx,./Base/QTCore/Resources/Certs/README,./Libs/MRML/Core/vtkMRMLCoreTestingUtilities.cxx,./Libs/vtkITK/vtkITKGrowCutSegmentationImageFilter.cxx,./Libs/vtkITK/itkConstrainedValueMultiplicationImageFilter.h,./Libs/vtkITK/itkLevelTracingImageFilter.h,./Modules/CLI/ExtractSkeleton/tilg_iso_3D.cxx,./Modules/CLI/ResampleDTIVolume/itkDiffusionTensor3D*,./Modules/Loadable/Reformat/Resources/UI/qSlicerReformatModuleWidget.ui,./Utilities/Scripts/runCodespell.sh"
           ignore_words_file: .codespellignore
+          exclude_file: CMake/NSIS.template.in
--- a/CMake/NSIS.template.in
+++ b/CMake/NSIS.template.in
@@ -525,6 +525,10 @@ Function DownloadFile
 FunctionEnd
 !endif
 
+Function finishpageaction
+    CreateShortCut '$DESKTOP\\\\@CPACK_NSIS_DISPLAY_NAME@.lnk' '$INSTDIR\\\\@CPACK_PACKAGE_NAME@.exe'
+FunctionEnd
+
 ;--------------------------------
 ; Define some macro setting for the gui
 @CPACK_NSIS_INSTALLER_MUI_ICON_CODE@
@@ -554,6 +558,11 @@ FunctionEnd
   !insertmacro MUI_PAGE_INSTFILES
   @CPACK_NSIS_INSTALLER_FINISH_TITLE_CODE@
   @CPACK_NSIS_INSTALLER_FINISH_TITLE_3LINES_CODE@
+
+  !define MUI_FINISHPAGE_SHOWREADME ""
+  !define MUI_FINISHPAGE_SHOWREADME_NOTCHECKED
+  !define MUI_FINISHPAGE_SHOWREADME_TEXT "Create Desktop Shortcut"
+  !define MUI_FINISHPAGE_SHOWREADME_FUNCTION finishpageaction
   !insertmacro MUI_PAGE_FINISH
 
   !insertmacro MUI_UNPAGE_CONFIRM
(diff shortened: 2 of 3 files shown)


lassoan:
> create a desktop shortcut which is pretty common

I mostly use default options when overusing software and only have a few desktop icons for dozens of installed applications

jamesobutler:
> Is there a specific reason you would want to add this option?

It's required by our company for our custom application as users are confused when the application is installed and there isn't a desktop icon. So I was contributing this back to the upstream from our fork. Many of our users are not tech savvy. Upon installation they only have to click on the desktop shortcut to launch the application rather than opening Windows start to find it which is often less familiar to macOS users that are running our application on the provide hardware instrument computer. The time for a user to download the application and get up and running is important to us.

Here I tied the desktop shortcut creation to the "Do not create shortcuts" option that exists in the current installer. CMake and some others I believe provide more granular control of which shortcuts are created. Do you have a specific opposition to the creation of this additional shortcut? Is it because multiple versions of Slicer can be installed rather than only a single version that is updated?

jcfr:
Considering that:
* adding the shortcut to the desktop is helping some of our "less tech savvy" users 
* deleting a shortcut from the desktop is easy for "expert" users
it seems reasonable to support this ?

pieper:
I agree it's a normal behavior for windows applications to make a desktop shortcut and I am used to deleting them.  Since we have specific cases of users expecting this behavior I think we should do it.

jcfr (approved):

lassoan:
> Do you have a specific opposition to the creation of this additional shortcut?

1. It would be yet another thing that Slicer does differently than most other software. It is worse for Slicer, which typically has several versions installed side by side.
2. The desktop is personal. We cannot guess or dictate how people use it and what content they put on it. Adding a shortcut there without asking, with every installation would be a little annoyance for many users.

https://apple.stackexchange.com/questions/338917/is-it-a-good-practice-to-create-a-desktop-shortcut-on-mac

I would find it much more user-friendly if there was an option to start Slicer after installation completed (then the user would not need to access the start menu or desktop to start the application at all). In addition, we could fix the "pin to taskbar" feature to allow users to make Slicer easier to start (not easy, due to multiple Slicer versions may be installed at multiple locations on the same system - see https://github.com/Slicer/Slicer/pull/5619).

If you think these would not be not sufficient then having the option to add the desktop shortcut is fine, just don't make it enabled by default (it could be a CMake option to enable desktop shortcut creation by default, so custom apps could easily choose the behavior they prefer). The current "Do not create shortcuts" checkbox is not usable for this, because creating start menu shortcuts is essential. Can we add an additional checkbox there? If not, then adding one more step to the installation could be OK, too, but then it would be nice to get rid of another step (merge the the Welcome and License Agreement steps?).

I expect that the start menu will be improved further on Windows11, because it is currently very difficult to find any application (need to click "All apps" - avoid clicking "More", which would be more intuitive - browse a long list in alphabetical order, open the application's section, then click on the application icon). So, maybe it is not worth investing a lot of time into new developments, such as fixing #5619, but only make small fixes or improvements.

jcfr:
It may be worth looking into transitioning to [CPack WIX Generator](https://cmake.org/cmake/help/latest/cpack_gen/wix.html).

WIX is a modern and actively maintained solution for creating windows installers. 

References:
* https://wixtoolset.org/docs/v3
* https://github.com/wixtoolset/
* https://twitter.com/wixtoolset

lassoan:
It would be nice to switch to a modern installer, but wix is probably not the solution. They have been struggling releasing wix4 for a decade. It was finally released a year ago, but due to this long delay the future is quite uncertain: probably many projects and tooling (such as CMake) will stick to wix3 for a good while, but maintainers probably want to work on wix4, and while this all is going on they have no time to work on things that really matter, such MSIX support. MSIX support requires a [paid wix plugin](https://github.com/wixtoolset/issues/issues/5781). It just does not make sense to go through the trouble of an intermediate toolkit, such as wix, we should just go straight to MSIX. I've added some more details here: https://github.com/Slicer/Slicer/issues/6766

Back to the topic of this PR. None of these modern installers allow low-level access to the computer, such as manipulating a user's desktop. We can do it now, but it is not the direction where things are going. So, I would suggest to accept @jamesobutler contribution but with keeping the current behavior as default. If it is easy to add a checkbox to add desktop shortcut then it's great; otherwise just make this configurable via a CMake option for easy access for custom applications.

jamesobutler:
Ok I will investigate a change to determine a best option for user selection of start menu shortcuts and also for a desktop shortcut while maintaining the default behavior of the existing Slicer installer (prior to this PR).

jamesobutler:
Some additional research here which confirms that various applications do desktop shortcut creation in various ways.
Type 1: Desktop shortcut created by default (no option) - Firefox and Adobe Acrobat Reader
Type 2: No desktop shortcut created (no option) - Zoom Desktop Client, MS Teams (work)
Type 3: Option to create a desktop shortcut (off by default) - CMake, Visual Studio Code

@lassoan Apologies for the delay here. I have pushed updates which I think you will like now as it matches Type 3. I have maintained the default option of not creating a desktop shortcut. On the existing "Finish" page of the Windows NSIS installer I have added an checkbox option to create a desktop shortcut. 

image

jamesobutler:
Interesting about your uninstall issue. I was testing on Windows 10 and I believe my desktop is connected with OneDrive, so maybe indeed a Windows 11 related issue? I will issue a fix if I notice it on my end. I had tested the uninstall and removal of the icon on my system which was working.

Also for some reason the docs status check is not reporting the status even though it appears to have been successful. @lassoan you will likely need to bypass the checks to merge anyways.

lassoan:
The documentation generation might have eventually gone through, or not. So, I've just bypassed this check and merged. Thank you for your nice contribution!

lassoan:
The new option for launching Slicer at the end of the installation is very useful. It is a small but very nice convenience that I don't need to dig in the Windows Start menu (it takes a lot of scrolling to find an application in the Windows11 Start menu).

Issue #6647, opened 2022-11-06
Translate user-visible text coming from logic classes


lassoan:
Most text that is displayed for users is coming from Qt GUI classes and translated using Qt translation infrastructure. However, there are probably cases when the GUI displays messages coming from logic classes. We need to review all the module logics classes (e.g., in all Logic, MRML, MRMLDM, VTKWidgets folders) if they provide any strings that end up being displayed for users and make sure they are translated.

Design options:
- Option A. Add a VTK-based translation class or function (that calls QApplication translate function). This is useful if there are many complex messages that logic classes may generate.
- Option B. Translate messages coming from logic classes in the GUI classes that display them. This may be better if there are only few or simple cases when logic classes provide messages visible to the user.

lassoan:
One example is volume rendering preset names. The preset name comes from node names in the presets.xml scene file (see https://github.com/Slicer/Slicer/pull/6762 for some more details).

lassoan:
A new `MRML_TRANSLATE_NOOP` macro could be added, which could work the same way as `QT_TRANSLATE_NOOP`.

    #define QT_TRANSLATE_NOOP(scope, x) x

Source text for translation should be possible by specifying extra options for `lupdate`.

A global function or singleton class could be added to MRML that would call `QCoreApplication::translate` via a callback mechanism. The callback mechanism is needed to avoid MRML depend on Qt classes. The Qt application would set the callback function pointer to `QCoreApplication::translate` at application startup.

mhdiop:
Hi @lassoan,
Hi @pieper.

I've started working on the implementation of the `vtkMRMLTr` function in the `Libs/MRML/Core/vtkMRML.h` file.

I first wanted to use a function pointer as stated in the previous posts but the `QCoreApplication::translate` has optional parameters  and such approach doesn't seem to be the most suitable one.

I then create a `vtkMRMLTr` function that has the same signature as `QCoreApplication::translate` and that directly calls it to make the translation. I tested it and it works.

However, strings present in the logic classes are not always of the same type. Many of them are `std::string` while others are `char*`. I can of course convert the `QString` returned by `QCoreApplication::translate` to any of those types but my question is which one should I choose as the return type of the translation function.

I know that any choice will have impact on the source code since it may result in many string conversions at translation function calls. Also, I wonder if those types are suitable for storing translated strings without any encoding issue ?

Here are some suggested implementations :

```cpp
[...]

Issue #6305, opened 2022-04-06
Consolidate ColorFiles documentation and hosting


jcfr:
## Is your feature request related to a problem? Please describe.

The page describing the color files shipped withing Slicer were historically hosted on the NA-MIC wiki, later relocated to the slicer wiki.

For example, previous home was https://www.na-mic.org/wiki/CCA:Atlases:MRI_Brain_Atlas_2010:ColorFile and current home is https://www.slicer.org/w/index.php/CCA:Atlases:MRI_Brain_Atlas_2010:ColorFile

The authoritative source for the color files was expected to be the wiki pages and color files were expected to be regenerated using using `labelpage.halle.us` (currently down). See [Historical site for generating color file from wiki page](#historical-site-for-generating-color-file-from-wiki-page) below.

Considering that we are phasing out the documentation from Slicer wiki to https://slicer.readthedocs.io/, we need to also consider the color files documentation and find a sustainable home.
consolidate
## Describe the solution you'd like

The key idea would be to update the [Color](https://slicer.readthedocs.io/en/latest/user_guide/modules/colors.html) module documentation and links to the relevant page describing each color files.

### Approach 1

Add markdown documents along side each color files at https://github.com/Slicer/Slicer/tree/master/Base/Logic/Resources/ColorFiles/ and reference each markdown from the `Color` module documentation.

### Approach 2

Similarly to what is proposed for the Volume Rendering Presets (https://github.com/Slicer/Slicer/issues/5206), create a dedicated repository called `Slicer/SlicerColorFiles` hosting color files and associated markdown.

## Describe alternatives you've considered

NA

### Color files

See https://github.com/Slicer/Slicer/tree/master/Base/Logic/Resources/ColorFiles/

```bash
$ ls -1 Base/Logic/Resources/ColorFiles/
64Color-Nonsemantic.txt
AbdomenColors.txt
Cividis.txt
ColdToHotRainbow.txt
DarkBrightChartColors.txt
DivergingBlueRed.txt
GenericAnatomyColors.txt
GenericColors.txt
HotToColdRainbow.txt
Inferno.txt
LightPaleChartColors.txt
Magma.txt
MediumChartColors.txt
PelvisColor.txt
Plasma.txt
Slicer3_2010_Brain_Labels.txt
Slicer3_2010_Label_Colors.txt
SPL-BrainAtlas-2009-ColorFile.txt
SPL-BrainAtlas-2012-ColorFile.txt
[...]

Issue #5206, opened 2020-09-28
Create SlicerVTKVolumeRenderingPresets repository


jcfr:
## Is your feature request related to a problem? Please describe.

Streamline maintenance and distribution of the volume rendering presets to facilitate (1) re-use and (2) integration of improvements.

## Describe the solution you'd like

Create a repository called `Slicer/SlicerVTKVolumeRenderingTransferFunctions` distributed (for example) as:

- git submodule, cmake external project  (e.g facilitate integration in any project)
- python package (e.g support re-use in Jupyter env.)
- npm package  (e.g to streamline integration in [vtk.js](https://kitware.github.io/vtk-js/index.html))
- VTK remote module (e.g to streamline integration in VTK or paraview based applications)

## Describe alternatives you've considered

NA

## Additional context

See https://discourse.slicer.org/t/2020-09-21-hangout-w-project-week-discussion/13594

lassoan:
The github repository should be not for just volume rendering preset files (new .vp.json file format - WIP), but also color tables (new .csv files), volume display presets (VolumeDisplayPresets.json), and in the future materials (textures, material properties).
 
Some ideas to start brainstorming on name of the repository:
•	SlicerPresetRepository (keep Slicer in the repository name for “advertising” that it is from the Slicer community)
•	SlicerPresets
•	SlicerDisplayPresets (although the colors, textures, etc. may be also used for 3D printing, so display may be a bit too narrow)
•	SlicerRenderPresets (similarly to “display” it may be a bit too narrow)
•	SlicerVizPresets
•	MedVizPresets
•	MedVizStyles

lassoan:
@muratmaga brought up the point that if we add a lot of new presets then we need a better way to browse them.

For this, preset files should have metadata including name, description, authors, keywords, biblio reference, tags or classifiers. Maybe we could use trove-style classifiers - https://pypi.org/classifiers/, which could allow filtering like this (from pypi):

![Image](https://github.com/user-attachments/assets/b1d97509-6e06-4a37-aba4-0e18bc8bf1f9)

Each preset file could be in a separate folder, so that we can keep together the preset file with supporting files, such as thumbnails, metadata json file, maybe a license file (but it would be much better to have a common MIT-like license for the entire repository).

pieper:
[...]

PR #6762, opened 2023-01-05
Improve volume rendering presets

lassoan:
Improve appearance of volume rendering preset selector and add micro-CT presets. Icons are now 2x higher resolution (so they don't look small or blurred anymore on high-resolution screens), larger, and labeled.

Before: (small icons, no text)

image

With changes in this pull request: (large, sharp icons, labelled)

image

Some of the existing presets don't seem to be very useful to me and we may consider removing in the future, but I did not want to do too many changes in one pull request.


lassoan:
@muratmaga This pull requests adds the 3 presets in SlicerMorph as MicroCT-bone-LDR, MicroCT-bone-HDR, MicroCT-mouse. LDR/HDR = low/high dynamic range.

I was not sure about the mouse preset. Is if for micro-CT? Is it for a specific mouse dataset or we can specify a broader application area and choose a name that better reflects that? Should it be just Micro-CT-bone-LDR2?

The preset names like `MicroCT-bone-LDR` are a bit too long and so truncated. Should we use `uCT-bone-LDR` instead?

Instead LDR/HDR (which may be misintepreted as low-dose-rate/high-dose-rate, and a bit long), should we use other acronyms, such as `NR`/`WR` (narrow/wide range), `LB`/`HB` (low/high bit depth)? I wanted to avoid the 8-bit/16-bit as it might be too technical, but maybe using that (`8b`/`16b`) could work, too. Any other ideas?

pieper (dismissed):
LGTM 👍

jamesobutler (commented):
Can we make sure the visible labeling is unique? In the provide screenshot of the changes there are multiple presets labeled as "CT-Coronary-..."

lassoan:
> Can we make sure the visible labeling is unique? In the provided screenshot of the changes there are multiple presets where the visible text is "CT-Coronary-..."

It would be nice, but for that we would need to change the preset name, which would be a little bit risky (some extensions may look up the preset by name). A better solution could be to separate the preset ID (or name) from the displayed label. This is also needed for internationalization, but we don't have a design yet for translating MRML/logic content.

Since this is already an improvement (previously there was no label at all), you can see the complete label when you hover over a preset, and you can also distinguish presets by appearance (difference between presets variants is mainly the color tone), I would incline not to hold up this pull request, but get back to this when we implement translation of VR preset names.

lassoan:
> Is reducing the text size label or adjusting the spacing of the grid possible so that the preset name isn't cut off? Currently there is elided text resulting in multiple labels of "CT-Coronary...".

Probably it is possible to disable eliding, but that would mean lots of wasted space (the longest name would increase the size of every item).

We could increase the icon size to avoid truncated text on some displays in some languages, but it would be very hard to prevent this from happening in all environments.

Reducing point size of the text is not a sustainable option (we cannot reduce the font size whenever the space is tight).

jamesobutler:
My only brainstorm idea for the display of the presets would be like what is done for the Color table selection where the icons are displayed to the left of the text in a single vertical column list. Then this accommodates longer names without having to create abbreviations that may be misinterpreted or run into text elide issues after only ~12 characters.
image

muratmaga:
> Instead LDR/HDR (which may be misintepreted as low-dose-rate/high-dose-rate, and a bit long), should we use other acronyms, such as `NR`/`WR` (narrow/wide range), `LB`/`HB` (low/high bit depth)? I wanted to avoid the 8-bit/16-bit as it might be too technical, but maybe using that (`8b`/`16b`) could work, too. Any other ideas?

I personally think 8bit/16bit is more straightforward than LDR/HDR but either way is fine.

jcfr:
> in multiple labels of "CT-Coronary...".

Since a tooltip is displayed with the fullname, I think the proposed solution is an improvement and worth the tradeoff. Especially considering no text is being displayed in the current version.

jcfr (approved):
In addition of this great improvement, would it be possible to update [developer_guide/modules/volumerendering](https://slicer.readthedocs.io/en/latest/developer_guide/modules/volumerendering.html#examples) by:
* briefly describing `presets.xml`
* explain how to re-generate the thumbnail (may be by linking to an entry in the script repository?)

Related feature requests:
* https://github.com/Slicer/Slicer/issues/5206
* https://github.com/Slicer/Slicer/issues/6305

lassoan:
OK, we can use the 8bit and 16bit in the preset name. I will add 8 and 16 to the icon to avoid the need to replace the icon for different languages.

lassoan:
@muratmaga can you comment the mouse preset? See my questions above.

muratmaga:
> I was not sure about the mouse preset. Is if for micro-CT? Is it for a specific mouse dataset or we can specify a broader application area and choose a name that better reflects that? Should it be just Micro-CT-bone-LDR2?

Sorry I missed this. It is for microCT too. it is just a steeper jump than the other 8 bit image, and tries to render the teeth more white than the rest of the specimen. it should be applicable to other microCT scans of bones too. Micro-CT-bone-LDR2 is fine. Or Micro-CT-skull-8bit...

lassoan:
Thanks for all the reviews and feedbacks. I've added a note about changing VR preset labels to https://github.com/Slicer/Slicer/issues/6647#issuecomment-1373094102.

PR #6765, opened 2023-01-06
ENH: Add missing array size hints for better Python wrapping

lassoan:
VTK Python wrapper requires a size hint in the header file (such as "VTK_SIZEHINT(3)") to be able to return a double* pointer as a float vector. This commit adds the missing size hints.

Example:

    print(slicer.mrmlScene.GetFirstNodeByClass('vtkMRMLCameraNode').GetPosition())

Returns a raw pointer without size hint:

    _00000272e44bc7b0_p_void

Returns a vector if a size hint is added to the method declaration:

    (0.0, 500.0, 0.0)

### DIFF
--- a/Libs/MRML/Core/vtkMRMLAbstractViewNode.h
+++ b/Libs/MRML/Core/vtkMRMLAbstractViewNode.h
@@ -174,12 +174,12 @@ public:
   vtkSetVector3Macro(LayoutColor, double);
   vtkGetVector3Macro(LayoutColor, double);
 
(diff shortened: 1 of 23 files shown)


pieper (approved):
Very nice 👍

lassoan:
Thanks for the reviews.

VTK_FILEPATH hint sound useful, too. 

The interface file generation should allow Auto-complete, type checks, etc. in IDEs. It would be nice if someone could have a look at how to set it up in VS Code and PyCharm.

PR #6767, opened 2023-01-07
DOC: Remove Pixel Medical from commercial partners

cpinter:
- Due to ceasing its operations, Pixel Medical is removed from the commercial partners list
- Add some details to the Ebatinca commercial partner

### DIFF
--- a/Docs/user_guide/about.md
+++ b/Docs/user_guide/about.md
@@ -190,10 +190,9 @@ We invite commercial entities to use 3D Slicer.
 
 ### Commercial Partners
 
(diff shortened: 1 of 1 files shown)


lassoan (approved):
Thanks for the update. It would make sense to add focus areas for all commercial partners.

lassoan:
Don't forget to upsets this page, too: https://www.slicer.org/commercial-use.html

jcfr:
> Update [commercial-use](https://www.slicer.org/commercial-use.html) page

See https://github.com/Slicer/slicer.org/pull/155

cpinter:
Excellent, thank you a lot both!

PR #6769, opened 2023-01-08
ENH: make error report box bigger

pieper:
Qt doesn't offer users any way to resize a message box, even when the detailedText field is displayed with lots of text. I didn't find a good way to resize programmatically either.

This is a problem with something like a python stack trace is displayed and the word wrapping makes it hard to read.

This workraound makes the title artificially bigger so that the message box size is wider than normal.

### DIFF
--- a/Base/Python/slicer/util.py
+++ b/Base/Python/slicer/util.py
@@ -2765,6 +2765,11 @@ def messageBox(text, parent=None, **kwargs):
         slicer.util.delayDisplay(text, autoCloseMsec=3000, parent=parent, **kwargs)
         return testingReturnValue
 
+    # if there is detailed text, make the dialog wider by making a long title
+    if "detailedText" in kwargs:
+        windowTitle = kwargs['windowTitle'] if 'windowTitle' in kwargs else ""
+        kwargs['windowTitle'] = windowTitle + " "*(200-len(windowTitle))
+
     import ctk
     mbox = ctk.ctkMessageBox(parent if parent else mainWindow())
     mbox.text = text
--- a/Base/Python/slicer/util.py
+++ b/Base/Python/slicer/util.py
@@ -2767,8 +2767,9 @@ def messageBox(text, parent=None, **kwargs):
 
     # if there is detailed text, make the dialog wider by making a long title
     if "detailedText" in kwargs:
-        windowTitle = kwargs['windowTitle'] if 'windowTitle' in kwargs else ""
-        kwargs['windowTitle'] = windowTitle + " "*(200-len(windowTitle))
+        windowTitle = kwargs['windowTitle'] if 'windowTitle' in kwargs else slicer.app.applicationName
+        padding = " "*((150 - len(windowTitle)) // 2) # to center the title
+        kwargs['windowTitle'] = padding + windowTitle + padding
 
     import ctk
     mbox = ctk.ctkMessageBox(parent if parent else mainWindow())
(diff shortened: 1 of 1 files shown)


pieper:
Before:
image
After:
image

Tested with this line: `errorDisplay("Error text", detailedText = "this could be a line from a stack trace\n"* 99)`

Does anyone know a better method?

lassoan:
Thanks for giving it a try to fix this. Unfortunately, it does not work for formatted input. For example:

```python
mbox = ctk.ctkMessageBox(mainWindow())
mbox.text = "test with <b>formatted</b> text" + " "*200
mbox.exec_()
```

However, this style sheet trick seem to work more robustly (and a bit more elegant):

```
mbox = ctk.ctkMessageBox(mainWindow())
mbox.text = "test with <b>formatted</b> text"
mbox.detailedText = "some"
mbox.setStyleSheet("QLabel{min-width: 400px;}");
mbox.exec_()
```

pieper:
Yes, that is cleaner code but the layout is now pretty ugly. 

image
 
I agree that adding the spaces to the title is a bit of a hack, but it only changes the `windowTitle` where formatting is not allowed anyway:
`messageBox("Error text", detailedText = "this could be a line from a stack trace\n"* 99, windowTitle="text with <b>formatting</b>")`
results in:
image

other ideas?

jamesobutler:
@pieper I'm not observing the ugly layout you show. What code are you running?

## Before:
image

```python
messageBox("test with <b>formatted</b> text", parent=mainWindow(), detailedText="this could be a line from a stack trace\n"* 99, styleSheet="QLabel{min-width: 400px;}")
```

## After:
image

```python
messageBox("test with <b>formatted</b> text", parent=mainWindow(), detailedText="this could be a line from a stack trace\n"* 99, styleSheet="QLabel{min-width: 400px;}")
```
```python
mbox = ctk.ctkMessageBox(mainWindow())
mbox.text = "test with <b>formatted</b> text"
mbox.detailedText = "this could be a line from a stack trace\n"* 99
mbox.setStyleSheet("QLabel{min-width: 400px;}");
mbox.exec_()
```

lassoan:
The messagebox layout is probably a bit different on macOS, that's why the stylesheet workaround works well on Windows but not on macOS.

Since the title length workaround seem to work on all platform, we should use that - even though it is really ugly.

I've just found two remaining issues:

image

1. If no window title is specified then the window title is now empty.

```
messageBox("test with <b>formatted</b> text", 
           parent=mainWindow(), 
           detailedText="this could be a line from a stack trace\n"* 99)
```

If no window title is specified then we should not just concatenate the spaces but we need to add the application name (that seems to be the default window title).

2. If the text is relatively short then the messagebox has a very wide shape. Maybe 100-150 spaces would be enough.

With 150 spaces:

image

With 100 spaces:

image

Maybe 150 spaces could be a good tradeoff. We should also add an argument (e.g., `titleMinimumLength`) to be able to set a custom width if the default is not good enough).

pieper:
thanks for the testing and suggestions 👍 

I just pushed an update that seems to be a fair compromise.  Agreed it's not a pretty solution but we shouldn't spend too much time on this and from what I have been able to research there's no good solution.

jamesobutler:
Are the differences you are seeing platform differences? Is that just how it looks on macOS? I was confused about why "Hide Details"/"Show Details" was positioned to the left of the "Ok" button.

lassoan:
On macOS, dialog buttons are in reverse order compared to Windows/Linux.

lassoan:
It looks good, but not perfect. On Windows, the window title is aligned to the left.
However, I don't think it is important to make this perfect. Probably it is good enough as is.

pieper:
> However, I don't think it is important to make this perfect. Probably it is good enough as is.

Agreed.  Can you give it a review so we can merge?

lassoan (approved):
It is not a perfect solution but good enough. There is a commit message test failure, but otherwise it is good to go.

pieper:
I fixed the commit message in the squash process.

PR #6771, opened 2023-01-09
Parameter node wrapper minor updates

Connor-Bowley:


Base/Python/slicer/parameterNodeWrapper/validators.py
```diff
@@ -132,8 +133,14 @@ class Choice(Validator):
     Validates that any input value is in the list of valid choices.
     """
 
-    def __init__(self, choices):
-        self.choices = choices
+    def __init__(self, *args):
```
lassoan, line 136: The input arguments are no longer self-explaining, therefore need to document what the method takes as input. Adding test or at least examples would be nice, too.

Base/Python/slicer/parameterNodeWrapper/wrapper.py
```diff
@@ -87,8 +88,10 @@ def _onModified(self, caller, event):
     def write(self, value) -> None:
         self._currentlyWriting = True
         try:
-            super().write(value)
-            self._value = self.parameter.read(self.parameterNode)
+            # important: need to read the value back before calling any observers
```
lassoan, line 91: Add an explanation why it is important.

lassoan (commented):
Thank you, it looks good, just added two small comments.

Base/Python/slicer/parameterNodeWrapper/validators.py
allemangD: I would prefer the caller have to do this conversion; too much magic going on to end up at the final list of choices.

For example:

```python
>>> Choice("Foo", "Bar")
Choice(['F', 'o', 'o', 'B', 'a', 'r'])
```

But intuitively I'd expect `Choice(['Foo', 'Bar'])`.

I don't think it's too bad to require the caller to flatten the choices before passing, and then `Choice`'s only argument is always the sequence of choices to use.

```python
def __init__(self, choices: Iterable):
    self.choices = list(choices)
```

Connor-Bowley: So, in other words, back out this change and keep the original functionality?

I missed the implication of using strings like in your `Choice("Foo", "Bar")` example, so I wouldn't necessarily be opposed to backing out the change.

The change was supposed to make things a little easier, not more confusing.

Connor-Bowley: Removed this commit to go back to old functionality.

Base/Python/slicer/parameterNodeWrapper/wrapper.py
Connor-Bowley: Added explanation

allemangD (approved):

PR #6772, opened 2023-01-09
Add Enum support to parameterNodeWrapper

allemangD:
Add an `EnumSerializer` which stores enums by their name and enforces that only valid instances of that enum may be serialized.

This grants static analysis support to certain existing uses of `Option` validator or `Union` type. Serializing by name also enables attaching metadata to particular enum values.

---

Updated _Vector to Scalar Volume_ module to use `Enum` for its conversion method parameter.

Note the new enum and parameter node definitions:

https://github.com/allemangD/Slicer/blob/9ea61e53281631e9177ce1f34faa8eb53bf0be4e/Modules/Scripted/VectorToScalarVolume/VectorToScalarVolume.py#L83-L103

And populating the `methodSelectorComboBox` using the enum metadata:

https://github.com/allemangD/Slicer/blob/9ea61e53281631e9177ce1f34faa8eb53bf0be4e/Modules/Scripted/VectorToScalarVolume/VectorToScalarVolume.py#L138-L142

### DIFF
--- a/Base/Python/slicer/parameterNodeWrapper/serializers.py
+++ b/Base/Python/slicer/parameterNodeWrapper/serializers.py
@@ -1,5 +1,6 @@
 import abc
 import collections
+import enum
 import logging
 import pathlib
 import typing
@@ -970,3 +971,51 @@ class UnionSerializer(Serializer):
         not update the underlying parameter node, which can be confusing.
         """
         return all([s.supportsCaching() for s in self._serializers])
+
+
+@parameterNodeSerializer
+class EnumSerializer(Serializer):
+    @staticmethod
+    def canSerialize(type_) -> bool:
+        return issubclass(type_, enum.Enum)
+
+    @staticmethod
+    def create(type_):
+        if EnumSerializer.canSerialize(type_):
+            return ValidatedSerializer(EnumSerializer(type_), [IsInstance(type_)])
+        return None
+
+    def __init__(self, enum_):
+        self.enum = enum_
+
+        if len(self.enum) == 0:
+            raise ValueError('Cannot serialize empty Enum.')
+
+    def default(self):
+        # There is no idiomatic way to get a default value of an arbitrary enum,
+        # so just take the first item in iteration order.
+        return next(iter(self.enum))
+
+    def isIn(self, parameterNode, name: str) -> bool:
+        return parameterNode.HasParameter(name)
+
+    def write(self, parameterNode, name: str, value) -> None:
+        key = value.name
+        parameterNode.SetParameter(name, key)
+
+    def read(self, parameterNode, name: str):
+        key = parameterNode.GetParameter(name)
+
+        try:
+            return self.enum[key]
+        except KeyError as ex:
+            raise ValueError(
+                f'Found {key!r} in parameter node but it is not a name in {self.enum!r}'
+            ) from ex
+
+    def remove(self, parameterNode, name: str) -> None:
+        parameterNode.UnsetParameter(name)
+
+    @staticmethod
+    def supportsCaching() -> bool:
+        return True
--- a/Modules/Scripted/VectorToScalarVolume/VectorToScalarVolume.py
+++ b/Modules/Scripted/VectorToScalarVolume/VectorToScalarVolume.py
@@ -1,17 +1,14 @@
 from contextlib import contextmanager
 import logging
-from typing import Annotated
+from typing import Optional
 import qt
 import vtk
+import enum
 
 import slicer
 from slicer.ScriptedLoadableModule import *
 from slicer.util import VTKObservationMixin
-from slicer.parameterNodeWrapper import (
-    parameterNodeWrapper,
-    Choice,
-    Default,
-)
+from slicer.parameterNodeWrapper import parameterNodeWrapper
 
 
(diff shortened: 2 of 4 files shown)


Base/Python/slicer/parameterNodeWrapper/serializers.py
```diff
@@ -970,3 +966,50 @@ def supportsCaching(self):
...
+    def default(self):
+        # There is no idiomatic way to get a default value of an arbitrary enum,
+        # so just take the first item in iteration order.
+        return next(iter(self.enum))
+
+    def isIn(self, parameterNode, name: str) -> bool:
+        return parameterNode.HasParameter(name)
+
+    def write(self, parameterNode, name: str, value) -> None:
+        if not isinstance(value, self.enum):
```
Connor-Bowley, line 996: `write` should assume the value is valid. The `create` method should look like

```py
def create(type_):
    if EnumSerializer.canSerialize(type_):
        return ValidatedSerializer(EnumSerializer(type_),
                                   [IsInstance(type_)])
    return None
```

so the `isinstance` is a validator that can be used by things like the `UnionSerializer`.

Base/Python/slicer/parameterNodeWrapper/serializers.py
```diff
@@ -970,3 +966,50 @@ def supportsCaching(self):
...
+        key = value.name
+        parameterNode.SetParameter(name, key)
+
+    def read(self, parameterNode, name: str):
+        key = parameterNode.GetParameter(name)
+
+        try:
+            return self.enum[key]
+        except KeyError as ex:
+            raise ValueError(f'{value!r} is not a value in {self.enum!r}') from ex
```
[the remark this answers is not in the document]
allemangD: Indeed. I am also updating the message to indicate the key was found in the parameter node but is not present in the enum; it might indicate the enum was changed and an old paremeter node is being read.

Base/Python/slicer/parameterNodeWrapper/serializers.py
allemangD: Wondering if it would also be appropriate to include `Choice(list(type_))` here? It seems appropriate but I'm not entirely certain if there are other implications of this.

Base/Python/slicer/parameterNodeWrapper/serializers.py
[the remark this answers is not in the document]
Connor-Bowley: Note another cause could be someone just writing random values to the name in the underlying parameter node. In general, the `read` functions are assuming that this has not happened and if it does the behavior is undefined.

Base/Python/slicer/parameterNodeWrapper/serializers.py
Connor-Bowley: Agree that adding `Choice` is redundant and not necessary.

Base/Python/slicer/parameterNodeWrapper/serializers.py
```diff
@@ -3,16 +3,12 @@
...
 import slicer
 
 from .util import splitAnnotations
-from .validators import (
-    extractValidators,
-    IsNone,
-    NotNone,
-    IsInstance,
-)
+from .validators import (extractValidators, IsNone, NotNone, IsInstance, Choice, )
```
Connor-Bowley, line 11: Why was this flattened to one line?

Modules/Scripted/VectorToScalarVolume/VectorToScalarVolume.py
```diff
@@ -82,18 +79,27 @@ def __init__(self, parent):
...
+        "Extract single component",
+    )
 
 
 @parameterNodeWrapper
 class VectorToScalarVolumeParameterNode:
     InputVolume: slicer.vtkMRMLVectorVolumeNode
     OutputVolume: slicer.vtkMRMLScalarVolumeNode
-    ConversionMethod: Annotated[str, Choice(_CONVERSION_METHODS), Default(_LUMINANCE)]
+    ConversionMethod: ConversionMethods
```
Connor-Bowley, line 102: I think this will not break loading an old MRML scene. Please confirm.

allemangD: Because the enum names `LUMINANCE`, `SINGLE_COMPONENT` are the same as the old string values `_LUMINANCE = "LUMINANCE"`, `_SINGLE_COMPONENT = "SINGLE_COMPONENT"` the parameter node contents are unchanged.

Base/Python/slicer/parameterNodeWrapper/serializers.py
allemangD: PyCharm "helpfully" formatted things before committing. I'll revert this soon.

Base/Python/slicer/parameterNodeWrapper/serializers.py
```diff
@@ -3,6 +3,7 @@
 import logging
 import pathlib
 import typing
+import enum
```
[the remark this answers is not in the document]
allemangD: Resolved.

Connor-Bowley (approved):

Connor-Bowley:
lgtm

PR #6774, opened 2023-01-10
Curve torsion measurement

cpinter:


Modules/Loadable/Markups/MRML/vtkCurveMeasurementsCalculator.cxx
```diff
@@ -314,6 +326,176 @@ bool vtkCurveMeasurementsCalculator::CalculatePolyDataCurvature(vtkPolyData* pol
...
+    for (int i=0; i<3; ++i)
+      {
+      prevNormBinormal[i] = normBinormal[i];
+      }
+    } // For each line point
+
+  if (!this->CurveIsClosed)
+    {
+    // The torsion for the last cell by definition is 0.0 for open curves
+    torsionArray->InsertValue(linePoints->GetId(numberOfPoints-1), 0.0);
```
lassoan, line 456: We should copy the previous distortion value, because that is a better estimation than 0.

Modules/Loadable/Markups/MRML/vtkCurveMeasurementsCalculator.cxx
```diff
@@ -314,6 +326,176 @@ bool vtkCurveMeasurementsCalculator::CalculatePolyDataCurvature(vtkPolyData* pol
...
+    }
+  else
+    {
+    // Use the adjacent values for closed curve instead of the singular values
+    torsionArray->SetComponent(linePoints->GetId(0), 0,
+      torsionArray->GetValue(linePoints->GetId(1)));
+    torsionArray->InsertValue(linePoints->GetId(numberOfPoints-1),
+      torsionArray->GetValue(linePoints->GetId(numberOfPoints-2)));
```
lassoan, line 464: Do we handle the cases correctly when we only have 0, 1, or 2 points?

Modules/Loadable/Markups/MRML/vtkCurveMeasurementsCalculator.cxx
```diff
@@ -314,6 +326,176 @@ bool vtkCurveMeasurementsCalculator::CalculatePolyDataCurvature(vtkPolyData* pol
...
+    {
+    return false;
+    }
+
+  if (this->Measurements->GetNumberOfItems() == 0)
+    {
+    vtkErrorMacro("No measurements found");
+    return false;
+    }
+  vtkMRMLMeasurement* firstMeasurement = vtkMRMLMeasurement::SafeDownCast(this->Measurements->GetItemAsObject(0));
```
lassoan, line 346: Shouldn't we look for torsion measurements specifically instead of just the first measurement? If there are no torsion measurements enabled then we may skip the torsion computation completely.

Also, if we need the markup node anyway then it may be cleaner to set the markup node (instead of the measurements object) as input in vtkCurveMeasurementsCalculator.

Modules/Loadable/Markups/MRML/vtkCurveMeasurementsCalculator.cxx
```diff
@@ -314,6 +326,176 @@ bool vtkCurveMeasurementsCalculator::CalculatePolyDataCurvature(vtkPolyData* pol
...
+  for (vtkIdType idx=1; idx<numberOfPoints-1; ++idx)
+    {
+    binormal = binormals->GetTuple3(linePoints->GetId(idx+1));
+    binormalNorm = sqrt(binormal[0]*binormal[0] + binormal[1]*binormal[1] + binormal[2]*binormal[2]);
+
+    normBinormal[0] = binormal[0] / binormalNorm;
+    normBinormal[1] = binormal[1] / binormalNorm;
+    normBinormal[2] = binormal[2] / binormalNorm;
+
+    // Local torsion
```
lassoan, line 425: could you add a reference to where the computation comes from?

Modules/Loadable/Markups/MRML/vtkCurveMeasurementsCalculator.h
```diff
@@ -30,7 +30,10 @@
 
 class vtkCallbackCommand;
 
-/// Filter that generates curves between points of an input polydata
+/// Filter that calculates per-point measurements for markups curves.
```
lassoan, line 33: Specify if "per point" is per curve point or per control point.

Modules/Loadable/Markups/MRML/vtkCurveMeasurementsCalculator.h
```diff
@@ -96,12 +115,16 @@ class VTK_SLICER_MARKUPS_MODULE_MRML_EXPORT vtkCurveMeasurementsCalculator : pub
...
+  /// Flag determining whether the filter should calculate torsion
+  bool CalculateTorsion{false};
+
   /// Command handling data array modified events
   vtkCallbackCommand* ControlPointArrayModifiedCallbackCommand;
   /// List of observed control point arrays (for removal of observations)
   vtkCollection* ObservedControlPointArrays;
 
-  std::string CurvatureUnits;
+  std::string CurvatureUnits{"mm-1"};
```
lassoan, line 126: It would be nice to set this based on the length unit in the scene.

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsCurveNode.cxx
```diff
@@ -1410,6 +1437,59 @@ void vtkMRMLMarkupsCurveNode::OnCurvatureMeasurementModified(
...
+      self->CurveMeasurementsCalculator->Update();
+      }
+    return;
+    }
+
+  // measurement is enabled
+  if (self->CurveMeasurementsCalculator->GetCalculateTorsion() && measurement->GetValueDefined())
+    {
+    // measurement was already on, nothing to do
+    return;
```
lassoan, line 1485: How do we know that recomputation is not needed? Shouldn't we check some MTime values?

lassoan (commented):
Thank you, it looks good overall, just added a few comments, mainly just to clarify/document how it all works.

cpinter:
Thanks for the review @lassoan, I'll address the comments soon.

Modules/Loadable/Markups/MRML/vtkCurveMeasurementsCalculator.cxx
```diff
@@ -158,6 +158,18 @@ int vtkCurveMeasurementsCalculator::RequestData(
     outputPolyData->GetPointData()->RemoveArray("Curvature");
     }
 
+  if (this->CalculateTorsion)
+    {
+    if (!this->CalculatePolyDataTorsion(outputPolyData))
+      {
+      return 0;
```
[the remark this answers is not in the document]
cpinter: Yes, makes a lot of sense. I don't return any more immediately.

Modules/Loadable/Markups/MRML/vtkCurveMeasurementsCalculator.cxx
cpinter: I added a check to require at least three points.

Modules/Loadable/Markups/MRML/vtkCurveMeasurementsCalculator.cxx
cpinter: The first measurement at this specific line was used just for getting the node. However, it's indeed cleaner to set/get the MRML node instead, I changed the code accordingly.

Modules/Loadable/Markups/MRML/vtkCurveMeasurementsCalculator.cxx
cpinter: Done

Modules/Loadable/Markups/MRML/vtkCurveMeasurementsCalculator.h
cpinter: OK. It's per curve point

Modules/Loadable/Markups/MRML/vtkCurveMeasurementsCalculator.h
cpinter: See `vtkMRMLMarkupsCurveNode::UpdateMeasurementsInternal`. I added torsion unit setting now.

cpinter:
I addressed all the comments in a separate commit and rebased on the current main. If the commit looks good I think the best would be squashing the last two together and integrating it as two commits. Thanks again.

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsCurveNode.cxx
cpinter: This function handles only enabling/disabling the specific measurements. We discussed it with @lassoan and this part is OK (except that there was a missing call so good that this part has been investigated)

cpinter:
@lassoan Can you please integrate if seems alright? Then the SlicerHeart people can try it in their module. Thank you

Modules/Loadable/Markups/MRML/vtkCurveMeasurementsCalculator.cxx
```diff
@@ -314,10 +329,180 @@ bool vtkCurveMeasurementsCalculator::CalculatePolyDataCurvature(vtkPolyData* pol
...
+  if (numberOfPoints < 3)
+    {
+    vtkErrorMacro("Number of curve points too low for torsion calculation (" << numberOfPoints << "), at least 3 needed");
+    return false;
+    }
+
+  // Initialize torsion array
+  vtkNew<vtkDoubleArray> torsionArray;
+  torsionArray->Initialize();
+  torsionArray->SetName("Torsion");
```
lassoan, line 395: This string should be accessible on the vtkCurveMeasurementsCalculator. Preferably it should be possible to set the array name, but at least it there should be a get method for getting the name.

Modules/Loadable/Markups/MRML/vtkCurveMeasurementsCalculator.cxx
```diff
@@ -314,10 +329,180 @@ bool vtkCurveMeasurementsCalculator::CalculatePolyDataCurvature(vtkPolyData* pol
...
+  vtkIdType numberOfPoints = // Last point in closed curve line is the first point
+    (this->CurveIsClosed ? linePoints->GetNumberOfIds()-1 : linePoints->GetNumberOfIds());
+  if (numberOfPoints < 3)
+    {
+    vtkErrorMacro("Number of curve points too low for torsion calculation (" << numberOfPoints << "), at least 3 needed");
+    return false;
+    }
+
+  // Initialize torsion array
+  vtkNew<vtkDoubleArray> torsionArray;
```
lassoan, line 393: If it does not make things very complicated, then try to reuse the existing array instead of always creating a new one.

Modules/Loadable/Markups/MRML/vtkCurveMeasurementsCalculator.cxx
```diff
@@ -314,10 +329,180 @@ bool vtkCurveMeasurementsCalculator::CalculatePolyDataCurvature(vtkPolyData* pol
...
+    torsionArray->SetComponent(linePoints->GetId(0), 0,
+      torsionArray->GetValue(linePoints->GetId(1)));
+    torsionArray->InsertValue(linePoints->GetId(numberOfPoints-1),
+      torsionArray->GetValue(linePoints->GetId(numberOfPoints-2)));
+    }
+
+  tangent = tangents->GetTuple3(linePoints->GetId(numberOfPoints-1));
+  currentLength = sqrt( tangent[0]*tangent[0] + tangent[1]*tangent[1] + tangent[2]*tangent[2] );
+  length += currentLength;
+  meanTorsion = meanTorsion / length;
```
lassoan, line 473: Length may be 0 (if points coincide), make sure that it does not cause major trouble.

Modules/Loadable/Markups/MRML/vtkCurveMeasurementsCalculator.h
```diff
@@ -66,11 +72,26 @@ class VTK_SLICER_MARKUPS_MODULE_MRML_EXPORT vtkCurveMeasurementsCalculator : pub
...
 
+  //@{
+  /// Set/Get flag determining whether to calculate torsion
+  vtkSetMacro(CalculateTorsion, bool);
+  vtkGetMacro(CalculateTorsion, bool);
+  vtkBooleanMacro(CalculateTorsion, bool);
+  //@}
+
+  /// Get name of average torsion measurement
+  static const char* GetAverageTorsionName() { return "torsion avg"; };
```
lassoan, line 86: We usually use `mean` instead of average, as it is shorter and maybe a bit more exact. We already use `mean` here for curvature, so let's rename these `GetMeanTorsionName` and `torsion mean`.

lassoan (commented):
It looks good, just added a few trivial comments.

Modules/Loadable/Markups/MRML/vtkCurveMeasurementsCalculator.h
cpinter: Ah yes, this was from the algorithm of the first approach that I implemented. I'll rename it.

Modules/Loadable/Markups/MRML/vtkCurveMeasurementsCalculator.cxx
cpinter: I think I tried but it caused some issues. Let me try again.

cpinter:
Thanks for the comments! I amended the commit resolving each of them.

cpinter:
@lassoan Could you please check? Thanks

lassoan:
@cpinter Please have a look at this new test failure, probably related, hopefully easy to fix: https://slicer.cdash.org/test/23944053

cpinter:
Will do soon!

PR #6775, opened 2023-01-10
BUG: Load table with default column type

gabrielmuller:
A table schema specifies metadata about each column of a table. It may also have a special row `<default>`. Unlike other rows, the default row does not correspond to a column of the table, instead representing the default properties of the table's columns.

`vtkMRMLTableStorageNode` was not taking this into account when reading a table from file, so I added a check which skips the `<default>` row when loading a table's columns.

Steps to reproduce:
- Create new table
- Under "New Column Defaults", set "Data type" to another type such as "Double"
- Add values to the table
- Save table
- Load table from file

### DIFF
--- a/Libs/MRML/Core/vtkMRMLTableStorageNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLTableStorageNode.cxx
@@ -21,6 +21,7 @@
 ==============================================================================*/
 
 // MRML includes
+#include "vtkMRMLMessageCollection.h"
 #include "vtkMRMLTableStorageNode.h"
 #include "vtkMRMLTableNode.h"
 #include "vtkMRMLScene.h"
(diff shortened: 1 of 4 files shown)


lassoan:
Thanks a lot for the error report and fix. While reviewing the proposed change, I've realized that the code was not very robust, so I've made a few more changes. It would be nice if you could test it.

gabrielmuller:
Thanks for the changes, I've tested them and they are working well.

While testing I have found a few other problems with tables, but it may be better to open separate issues for these. These problems were already present and not introduced in this PR. What do you think?

* When saving a table with string and int columns, the column order is changed in the saved schema. When reading, the columns are ordered according to the schema. This means column order changes after writing and reading a table.

* A string column where all values are the empty string is not read correctly. Its column name becomes associated with the contents of the next column.

* If a default null value is not specified, creating columns of non-string type generates garbage values.

lassoan:
Thank you for testing! Please submit an issue (and preferably pull request) for these remaining problems.

I agree that column order should not be changed at read/write, except when reading multi-component arrays: multi-component arrays (such as coordinate .x, .y, .z) is read into a single array in VTK, so even if they were not adjacent in the input .csv file, they will be displayed adjacent in Slicer and will be written like that in the csv when the table is saved.

PR #6781, opened 2023-01-14
BUG: Fix computation of offset direction label in slicer view controller

### LINKED ISSUE #6778: Sagittal slice view sign inverted
## Summary

Sagittal slice view sign seems to be inverted. See screenshot

**A** and **B** indicate crosshair position in the left side.
**C** shows slice view value of positive R

Picture1

## Steps to reproduce

- Load sample image MRHead
- Scroll sagittal slice to positive right values
- Place crosshair / look at Data Probe
- Slice view shows positive R, while data probe shows L

## Expected behavior

Sagittal view should show negative R values for left side.

## Environment
- Slicer version: Slicer-5.2.1 and Slicer-5.3.0-2023-01-12
- Operating system: Mac

lassoan:
Sagittal plane offset direction label "R" was incorrect: moving the slider to the right did not move the slice in "R" direction (while for axial and coronal slices, moving the slider to the right moved the slice according to the direction specified in the label).

Instead of changing the incorrect "R" label to "L", this commit removes the hardcoded labels and instead computes the correct label dynamically. If the slice normal direction is not aligned with an axis then the label contains a combination of directions. The order of axes reflects the dominance of the axis. For example, if the plane normal points to anterior and slightly left then the label is AL.

fixes #6778

### DIFF
--- a/Docs/user_guide/user_interface.md
+++ b/Docs/user_guide/user_interface.md
@@ -166,7 +166,7 @@ View Controllers module provides an alternate way of displaying these controller
 - **Slice orientation** displays allows you to choose the orientation for this slice view.
 - **Lightbox** to select a mosiac (a.k.a. contact sheet) view.  Not all operations work in this mode and it may be removed in the future.
 - **Reformat** allows interactive manipulation of the slice orientation.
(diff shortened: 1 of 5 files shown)


pieper (approved):
Nice! 👍

lassoan:
Thanks for the review. I'll merge this now.

PR #6791, opened 2023-01-21
BUG: Fix reading of vector fields from NIFTI files

lassoan:
When ITK reads a vector field from NIFTI image, it changes the index to physical to IJKToLPS, but it does not reorient the vectors from RAS to LPS. This commit handles the special case when reading a vector field from NIFTI format.

Future ITK versions might reorient the vectors - in that case this special case handling should be removed from here.

This will allow reading displacement fields in NIFTI file format, for example, created by VoxelMorph. Currently, VoxelMorph writes an invalid NIFTI file, but it should be fixed soon (see https://github.com/voxelmorph/voxelmorph/issues/438).

### DIFF
--- a/Libs/MRML/Core/vtkITKTransformConverter.h
+++ b/Libs/MRML/Core/vtkITKTransformConverter.h
@@ -69,46 +69,66 @@ public:
   /// Create an ITK transform from a VTK transform.
   /// secondaryTransformItk: Only for backward compatibility. For BSpline transforms that have an additive bulk component, the bulk component is saved
   ///   in the secondary transform.
-  /// preferITKv3CompatibleTransforms: If true then BSpline transform will created as BSplineDeformableTransform and additive bulk transform component is always written
-  ///   in the secondary transform. If false then BSpline transform will be written as BSplineTransform (multiplicative bulk component is saved in a composite transform).
+  /// preferITKv3CompatibleTransforms: If true then BSpline transform will created as BSplineDeformableTransform and additive bulk transform component
+  //    is always written in the secondary transform. If false then BSpline transform will be written as BSplineTransform (multiplicative bulk component
+  //    is saved in a composite transform).
   /// If initialize is set to true then the transform is initialized to be readily usable.
   /// Initialization takes a long time for kernel transforms with many points,
   /// If a transform is created only to write it to file, initialization can be turned off to improve performance.
   static itk::Object::Pointer CreateITKTransformFromVTK(vtkObject* loggerObject, vtkAbstractTransform* transformVtk,
     itk::Object::Pointer& secondaryTransformItk, int preferITKv3CompatibleTransforms, bool initialize = true);
 
-  template <typename T> static bool SetVTKBSplineFromITKv3Generic(vtkObject* loggerObject, vtkOrientedBSplineTransform* bsplineVtk, typename itk::TransformBaseTemplate<T>::Pointer warpTransformItk, typename itk::TransformBaseTemplate<T>::Pointer bulkTransformItk);
+  template <typename T> static bool SetVTKBSplineFromITKv3Generic(vtkObject* loggerObject, vtkOrientedBSplineTransform* bsplineVtk,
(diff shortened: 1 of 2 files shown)


lassoan:
Submitted a related pull request to ITK that allows reading NIFTI files that use "displacement vector" intent, which would further improve compatibility - see https://github.com/InsightSoftwareConsortium/ITK/pull/3884

Libs/MRML/Core/vtkITKTransformConverter.h
```diff
@@ -69,46 +69,66 @@ class vtkITKTransformConverter
   /// Create an ITK transform from a VTK transform.
   /// secondaryTransformItk: Only for backward compatibility. For BSpline transforms that have an additive bulk component, the bulk component is saved
   ///   in the secondary transform.
-  /// preferITKv3CompatibleTransforms: If true then BSpline transform will created as BSplineDeformableTransform and additive bulk transform component is always written
-  ///   in the secondary transform. If false then BSpline transform will be written as BSplineTransform (multiplicative bulk component is saved in a composite transform).
+  /// preferITKv3CompatibleTransforms: If true then BSpline transform will created as BSplineDeformableTransform and additive bulk transform component
+  //    is always written in the secondary transform. If false then BSpline transform will be written as BSplineTransform (multiplicative bulk component
+  //    is saved in a composite transform).
```
jcfr, line 74:
```suggestion
  /// preferITKv3CompatibleTransforms: If true then the BSpline transform will be created as a BSplineDeformableTransform and additive bulk transform component
  ///  is always written in the secondary transform. If false then the BSpline transform is written as a BSplineTransform (multiplicative bulk component
  ///  is saved in a composite transform).
```

Libs/MRML/Core/vtkITKTransformConverter.h
```diff
@@ -969,15 +1001,37 @@ bool vtkITKTransformConverter::SetVTKOrientedGridTransformFromITKImage(vtkObject
...
+  if (displacementVectorCoordinateSystem == vtkMRMLStorageNode::CoordinateSystemLPS)
     {
-    typename GridImageType::PixelType displacementVectorLps=inputIt.Get();
-    *(displacementVectors_Ras++) = -displacementVectorLps[0];
-    *(displacementVectors_Ras++) = -displacementVectorLps[1];
-    *(displacementVectors_Ras++) =  displacementVectorLps[2];
-    ++inputIt;
+    while (!inputIt.IsAtEnd())
+      {
+        // vectors are in LPS, convert them to LPS
```
jcfr, line 1009:
```suggestion
        // vectors are in LPS, convert them to RAS
```

Libs/MRML/Core/vtkMRMLTransformStorageNode.cxx
```diff
@@ -258,9 +258,20 @@ int vtkMRMLTransformStorageNode::ReadFromImageFile(vtkMRMLNode *refNode)
       return 0;
     }
 
+  // ITK reads all images into LPS coordinate system (except special cases)
+  int displacementVectorCoordinateSystem = vtkMRMLStorageNode::CoordinateSystemLPS;
+
+  // NIFTI is a special case: it uses RAS coordinate system for the displacement vectors and ITK does not reorient them
+  // into LPS direction, therefore they are still in RAS.
```
jcfr, line 265: Should a link to `https://github.com/InsightSoftwareConsortium/ITK/pull/3884` be added here ? Along with a comment in `External_ITK.cmake` to remind us of updating the logic here when ITK will effectively reorient vector field values ?

Additionally, should we consider back-porting the corresponding commit to `Slicer/ITK` ?

lassoan: The ITK pull request is about accepting the "displacement vector" intent code (not just the generic "vector"). Reorientation of the vectors is just raised as a question in the pull request. I would integrate this fix now (as it is needed to be able to load VoxelMorph registration results from nifti files) and then implement the additional fixes in ITK without any time pressure.

jcfr (approved):
I suggest to `Squash & Merge`

PR #6793, opened 2023-01-23
Fix i18n qt base regressions

jcfr:
This fixes regressions introduced in https://github.com/Slicer/Slicer/commit/1031050a0a677381ff40030a218702cd45fda668 (BUG: Mark translatable strings in Base/QTCore).

To catch issue like the one associated with the return type, we are also looking into marking `-Wreturn-type` as an error so that the CI effectively fails.


jcfr:
I plan to integrate before 9pm ET so that the changes are integrated before the next round of preview builds.

Connor-Bowley (approved):

PR #6798, opened 2023-01-26
ENH: Bundle Qt language tools

lassoan:
Bundle lconvert, lrelease, lupdate tools with the application to allow Slicer modules to update and process translation (.ts) files without requiring Qt installation.

Required for https://github.com/Slicer/SlicerLanguagePacks/issues/1

### DIFF
--- a/CMake/SlicerMacroBuildApplication.cmake
+++ b/CMake/SlicerMacroBuildApplication.cmake
@@ -556,6 +556,45 @@ macro(slicerMacroBuildApplication)
           OUTPUTVAR extraApplicationToLaunchListForBuildTree
           )
       endif()
+
+      # Make available Qt language tools conveniently on the command line, for example:
+      # Slicer.exe --lconvert ...
+      if(Slicer_BUILD_I18N_SUPPORT)
+        if(NOT QT_LCONVERT_EXECUTABLE)
+          find_program(QT_LCONVERT_EXECUTABLE lconvert HINTS "${QT_BINARY_DIR}" NO_DEFAULT_PATH)
+        endif()
+        if(EXISTS ${QT_LCONVERT_EXECUTABLE})
+          ctkAppLauncherAppendExtraAppToLaunchToList(
+            LONG_ARG lconvert
+            HELP "Start Qt lconvert language utility"
+            PATH ${QT_LCONVERT_EXECUTABLE}
+            OUTPUTVAR extraApplicationToLaunchListForBuildTree
+            )
+        endif()
+        if(NOT QT_LRELEASE_EXECUTABLE)
+          find_program(QT_LRELEASE_EXECUTABLE lrelease HINTS "${QT_BINARY_DIR}" NO_DEFAULT_PATH)
+        endif()
+        if(EXISTS ${QT_LRELEASE_EXECUTABLE})
+          ctkAppLauncherAppendExtraAppToLaunchToList(
+            LONG_ARG lrelease
+            HELP "Start Qt lrelease language utility"
+            PATH ${QT_LRELEASE_EXECUTABLE}
+            OUTPUTVAR extraApplicationToLaunchListForBuildTree
+            )
+        endif()
+        if(NOT QT_LUPDATE_EXECUTABLE)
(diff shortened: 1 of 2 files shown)


lassoan:
@jcfr please check if this is a correct way to bundle these Qt utilities. I did not add a separate launcher, as we will use the executables from Slicer's environment.

@pieper it would be great if you could test this on your mac, ideally in the install tree. The best test is to install the latest main version of https://github.com/Slicer/SlicerLanguagePacks and then:
- select a few languages in `Weblate (latest translations)` section
- leave the `Qt lrelease path` empty in `Settings` section
- click `Update translation files`

CMake/SlicerCPack.cmake
```diff
@@ -111,6 +111,29 @@ if(Slicer_BUILD_QT_DESIGNER_PLUGINS)
...
+# Get Qt root directory
+get_property(_filepath TARGET "Qt5::Core" PROPERTY LOCATION_RELEASE)
+get_filename_component(_dir ${_filepath} PATH)
+set(qt_root_dir "${_dir}/..")
+
+if(Slicer_BUILD_I18N_SUPPORT)
+  foreach(tool lconvert lrelease lupdate)
+    install(PROGRAMS ${qt_root_dir}/bin/${tool}${CMAKE_EXECUTABLE_SUFFIX}
+      DESTINATION ${Slicer_INSTALL_ROOT}/bin COMPONENT Runtime
+      )
```
jcfr, line 123:
```suggestion
      )
      slicerStripInstalledLibrary(
        FILES "${Slicer_INSTALL_ROOT}/bin/${tool}"
        COMPONENT Runtime
        )
```

This is similar to what is done in [CMake/SlicerBlockInstallQt.cmake#L184-L193](https://github.com/Slicer/Slicer/blob/201ab3a1d1bb16938f41db63634bc0190cc08d27/CMake/SlicerBlockInstallQt.cmake#L184-L193) for Qt Designer.

For sake of consistency, I suggest to move this (and the update of `rpath`) to `SlicerBlockInstallQt.cmake`

CMake/SlicerCPack.cmake
```diff
@@ -111,6 +111,29 @@ if(Slicer_BUILD_QT_DESIGNER_PLUGINS)
...
+    if(APPLE)
+      set(dollar "$")
+      install(CODE
+        "set(app ${Slicer_INSTALL_BIN_DIR}/${tool})
+        set(appfilepath \"${dollar}ENV{DESTDIR}${dollar}{CMAKE_INSTALL_PREFIX}/${dollar}{app}\")
+        message(\"CPack: - Adding rpath to ${dollar}{app}\")
+        execute_process(COMMAND install_name_tool -add_rpath @loader_path/..  ${dollar}{appfilepath})"
```
jcfr, line 130: This makes sense.

CMake/SlicerMacroBuildApplication.cmake
```diff
@@ -615,6 +652,31 @@ macro(slicerMacroBuildApplication)
           )
       endif()
 
+      if(EXISTS ${QT_LCONVERT_EXECUTABLE} AND NOT APPLE)
```
jcfr, line 655: To ensure preexisting cache value do not implicitly enable this code path (e.g after turning `Slicer_BUILD_I18N_SUPPORT` to `OFF`), I suggest to wrap this into `if(Slicer_BUILD_I18N_SUPPORT) / endif()`

lassoan: OK!

lassoan:
Thanks a lot for the comments @jcfr, I've updated the files accordingly.

jcfr:
Testing locally on Linux :hourglass:

lassoan:
I've tested on Windows and it worked well in both the build and install trees.

If your linux testing does not bring up any issue then I will merge this before the nightly cutoff time, because the macOS packaging is very slow to do manually.

lassoan:
Awesome, thanks a lot for testing. We should include this in Slicer-5.2.2 patch, too.

jcfr:
> We should include this in Slicer-5.2.2 patch, too.

Once tonight builds complete and we have a chance to test macOS, I will backport this PR.

PR #6799, opened 2023-01-27
BUG: Update ITK to fix read/write of displacement field in NIFTI file format

jcfr:
This commit is a follow-up of 898959159 (BUG: Fix reading of vector fields from NIFTI files)

NIFTI image containing displacement field (`intent = 1006`) was incorrectly read as scalar volume (all 3 vector components were set to the same value).

Fixed by reading NIFTI displacement vector files similarly to general-purpose vector images. For displacement vector files, vector components are converted between NIFTI file's `RAS` coordinate system and ITK's `LPS` coordinate system, unless this conversion is explicitly disabled by setting `ConvertRASDisplacementVectors` to false.

An ITK vector image can be written as NIFTI displacement vector file by setting "intent_code"="1006" in the metadata dictionary of the image.

List of ITK changes:

```
$ git shortlog be914a8b57..eeb7f43c49 --no-merges
Andras Lasso (1):
      [Backport PR-3884] BUG: Fix read/write of displacement field in NIFTI file format
```

References:
* https://github.com/Slicer/Slicer/pull/6791
* https://github.com/Slicer/ITK/pull/7

### DIFF
--- a/Libs/MRML/Core/vtkITKTransformConverter.h
+++ b/Libs/MRML/Core/vtkITKTransformConverter.h
@@ -81,11 +81,9 @@ public:
   template <typename T> static bool SetVTKBSplineFromITKv3Generic(vtkObject* loggerObject, vtkOrientedBSplineTransform* bsplineVtk,
     typename itk::TransformBaseTemplate<T>::Pointer warpTransformItk, typename itk::TransformBaseTemplate<T>::Pointer bulkTransformItk);
 
-  /// Image grid is always in LPS but vectors may be in LPS (default) or RAS (when reading displacement field from NIFTI)
   template<typename T>
   static bool SetVTKOrientedGridTransformFromITKImage(vtkObject* loggerObject, vtkOrientedGridTransform* grid_Ras,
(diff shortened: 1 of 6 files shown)


jcfr:
> another ITK NIFTI fix that I don't think was backported to the Slicer fork

Good catch. The change has been backported and added to this pull request.

lassoan:
Thanks a lot for preparing this PR. Along with the ITK change we need to make a small change in Slicer, too (remove the RAS/LPS flipping, as it is now done in ITK). The implementation is ready, I'm just running the tests now. If all looks good then I'll push the additional commits to this branch.

lassoan:
Thank you, I see that you have already removed the unnecessary RAS/LPS conversion in Slicer your commit. I've extended it with simplifying the transform converter (removing the extra coordinate system parameter that is no longer needed) and I've also fixed the error/warning reporting in the transforms reader (so now when a NIFTI file is read that does not have the proper intent code then a warning is displayed). I've forced push all this to this branch. This should be ready to review/merge.

lassoan (approved):

PR #6801, opened 2023-01-27
BUG: Table column type issues

gabrielmuller:
* BUG: Fix non-string table cells initialized with garbage
When a null value for a non-string type column was not specified,
new cells were showing arbitrary values. This fix tries to initialize
cells in this order:

    1. The null value specified by the user, if compatible with the column type
    1. NUL character, if column is char type
    1. 0 otherwise

* BUG: Remove idtype from column types
idtype is not supported by vtkVariant and its size is machine dependent,
so it seems simpler to remove this option.

* BUG: Fix non-str cols always loaded before str cols
When editing a table, string columns with no extra metadata are not
included in the schema. When the schema file is written, these missing
columns are appended to the schema. This commit reorders the schema
before writing so the schema preserves the order in which the columns
appear in the table.

* I have also opened a merge request for a table related problem in VTK. https://gitlab.kitware.com/vtk/vtk/-/merge_requests/9890

### DIFF
--- a/Libs/MRML/Core/vtkMRMLTableNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLTableNode.cxx
@@ -337,7 +337,19 @@ vtkAbstractArray* vtkMRMLTableNode::AddColumn(vtkAbstractArray* column)
     newColumn = vtkSmartPointer<vtkAbstractArray>::Take(vtkAbstractArray::CreateArray(valueTypeId));
     newColumn->SetNumberOfTuples(numberOfRows);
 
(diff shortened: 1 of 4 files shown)


lassoan (commented):
Thank you for your contribution and sorry for the long delay in the review. You hunted down some tricky issues and provided solutions, which were very useful. Instead of merging your solutions as they were, I've created a new pull request that solves a few limitations of the fixes in this pull request.

Since #6850 is intended to supersede this pull request, I'm closing this pull request now.

It would be great if you could test #6850 to see if it works well for you. Thank you!

PR #6808, opened 2023-02-01
Parameter node wrapper any serializer

Connor-Bowley:
Adds a new serializer for `typing.Any`

Fixes a couple bugs.

### DIFF
--- a/Base/Python/slicer/parameterNodeWrapper/serializers.py
+++ b/Base/Python/slicer/parameterNodeWrapper/serializers.py
@@ -1,6 +1,7 @@
 import abc
 import collections
 import enum
+import importlib
 import logging
 import pathlib
 import typing
@@ -29,6 +30,7 @@ __all__ = [
     "TupleSerializer",
     "DictSerializer",
     "UnionSerializer",
+    "AnySerializer",
 ]
 
 
@@ -536,7 +538,7 @@ class ListSerializer(Serializer):
 
     @staticmethod
     def canSerialize(type_) -> bool:
-        return typing.get_origin(type_) == list
+        return typing.get_origin(type_) == list or type_ == list
 
     @staticmethod
     def create(type_):
@@ -545,7 +547,7 @@ class ListSerializer(Serializer):
             if len(args) != 1:
                 logging.warning("Unexpected list[] type arg length")
             if len(args) == 0:
-                Exception("Unsure how to handle a typed list with no discernible type")
+                args = [typing.Any]
             return ListSerializer(createSerializerFromAnnotatedType(args[0]))
         return None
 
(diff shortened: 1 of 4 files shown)


allemangD:
I have some doubts on `pickle`-ing data into a parameter node... for example a Python upgrade might cause issues loading the pickled data, or upgrading an extension might change the qualified name of some class and render these scenes unusable.

There are also security concerns, since `pickle` allows executing arbitrary code, although I'm not certain how secure MRML is as a whole. It is still better in general to reduce the surface area for those kinds of concerns.

---

I guess all this depends on some semantic decisions: are the contents of a parameter pack intended to be persistent _on disk/network_ or _between versions_, or are they only meant to facilitate more "local" module reloading and pipelining/workflow operations?

Connor-Bowley:
Removed usage of pickle because it would make MRB storage unsafe. Also allow mixing of AnySerializer, list, and dict.

---

To address a couple comments:

> Is there any promise of security when loading an untrusted MRML scene?

Not really, but it shouldn't be able to execute arbitrary code.

> a Python upgrade might cause issues loading the pickled data

I don't think this is an issue, pickle seems to be backwards compatible.

> upgrading an extension might change the qualified name of some class and render these nodes unusable.

I would kind of expect if a type is stored and the name of the type is changed that it would wreak havoc on anything trying to read it, unless you knew a priori what you were trying to read, which is not really possible here.

Note: Changing the serializer names shouldn't break anything, but changing the serialized type names or modules will.

Base/Python/slicer/parameterNodeWrapper/serializers.py
```diff
@@ -1039,6 +1043,7 @@ class AnySerializer(Serializer):
       classes that were created in functions.
     - Can never cache.
```
allemangD, line 1044: I wonder if it is possible to cache the _serializer_, so it doesn't need to be recreated on each read/write. If the name and module of the object's type don't change, the chosen serializer doesn't need to change either and so could persist. That might facilitate caching values in _certain_ circumstances, when the AnySerializer is used with the same type and serializer for that type supports caching.

That might be more difficult to implement than I'd expect, so I defer to you @Connor-Bowley on how feasible you think that would be.

Base/Python/slicer/parameterNodeWrapper/serializers.py
```diff
@@ -1052,39 +1057,48 @@ def create(type_):
...
 
     def _getValueSerializer(self, parameterNode, name: str) -> Serializer:
-        type_ = pickle.loads(bytes.fromhex(self._typeSerializer.read(parameterNode, self._serializerName(name))))
-        return createSerializerFromAnnotatedType(type_)
+        modulename, classname = self._typeSerializer.read(parameterNode, self._typeName(name))
+        if (modulename, classname) == ("builtins", "NoneType"):
+            # "from builtins import NoneType" doesn't work
+            return createSerializerFromAnnotatedType(type(None))
+        else:
+            mod = sys.modules[modulename]
```
allemangD, line 1073: :+1: Definitely less concerned about security issues with `pickle` gone.

Although I wonder if here it would be better to use `importlib.import_module` instead of fetching cached modules? It seems unlikely but not impossible for the module to not yet be imported. I don't expect there to be security issues as `import_module` requires the module to already exist on-disk; it can't execute code stored in the scene.

Base/Python/slicer/parameterNodeWrapper/serializers.py
allemangD: After some more thought I don't think this is a good idea. It is entirely incompatible with use cases on containers like `list[Any]` or un-annotated types like `dict`, etc.

Base/Python/slicer/parameterNodeWrapper/serializers.py
Connor-Bowley: Will update to use `importlib`

Connor-Bowley: Updated.

allemangD (approved):
:+1: thanks for those changes @Connor-Bowley

PR #6809, opened 2023-02-01
ENH: Add connector line for fiducial markup label texts

lassoan:
This is a continuation of https://github.com/Slicer/Slicer/pull/6188 - the branch is rebased to the latest master and a few fixes and simplifications has been made.

---

- add connector lines
- handle border collisions
- enable XML read and writer
- add connector lines in 3D view
- adjust connector line angles in 3D angles according to 3D focal point
- remove connector line color picker button and its overhead
- add automatic connector line color algorithm
- renaming leader lines label
- initialize leader lines variables in vtkMRMLMarkupsDisplayNode
- rename all "labelconnector" variables to "leaderlines"
- restore normal 2D label offest (without leaderlines)
- markup read/write test
- repositioning "Leader lines:" label to -> Advanced section in Markups Display Widget (+ indenting two spaces)
- clean up indents in qMRMLMarkupsDisplayNodeWidget.ui
- make lines only visible if ctrl pts visible and label text present
- lines in occluded state to be discussed
- set leader line to 25% of glyph size in 2D and 3D
- a change in upper and lower border collision handling 2D
- do not draw leader lines when "Control Point Labels:" is checked off in "Display"->"Advanced"
- determine background volume centers in 2D views as leader lines reference
- make leader lines fluently pointing to this center in 2D views
- trigonometry to seperate function in vtkSlicerMarkupsWidgetRepresentation2D preparing 3D
- 3D view: make leader line tube filter radius update in time
- set all 3D label text positions to end of connector line

### DIFF
--- a/Modules/Loadable/Markups/VTKWidgets/vtkSlicerMarkupsWidgetRepresentation2D.cxx
+++ b/Modules/Loadable/Markups/VTKWidgets/vtkSlicerMarkupsWidgetRepresentation2D.cxx
@@ -23,6 +23,7 @@
 #include "vtkGlyph2D.h"
 #include "vtkLabelPlacementMapper.h"
 #include "vtkLine.h"
+#include "vtkLineSource.h"
 #include "vtkMarkupsGlyphSource2D.h"
 #include "vtkMath.h"
 #include "vtkMatrix4x4.h"
@@ -98,6 +99,21 @@ vtkSlicerMarkupsWidgetRepresentation2D::ControlPointsPipeline2D::ControlPointsPi
   this->LabelsMapper->GetAnchorTransform()->SetCoordinateSystemToNormalizedViewport();
   this->LabelsActor = vtkSmartPointer<vtkActor2D>::New();
   this->LabelsActor->SetMapper(this->LabelsMapper);
+
+  // Labels line polydata
+  this->LabelsLinePolyData = vtkSmartPointer<vtkPolyData>::New();
+
+  this->LabelsLineProperty = vtkSmartPointer<vtkProperty2D>::New();
+  this->LabelsLineProperty->SetOpacity(1.);
+  this->LabelsLineProperty->SetLineWidth(this->LabelsLineProperty->GetLineWidth());
+  this->LabelsLineMapper = vtkSmartPointer<vtkPolyDataMapper2D>::New();
+  this->LabelsLineMapper->SetInputData(this->LabelsLinePolyData);
+
+  this->LabelsLineMapper->SetScalarVisibility(true);
+
+  this->LabelsLineActor = vtkSmartPointer<vtkActor2D>::New();
+  this->LabelsLineActor->SetMapper(this->LabelsLineMapper);
+  this->LabelsLineActor->SetProperty(this->LabelsLineProperty);
 }
 
 vtkSlicerMarkupsWidgetRepresentation2D::ControlPointsPipeline2D::~ControlPointsPipeline2D() = default;
@@ -151,6 +167,24 @@ void vtkSlicerMarkupsWidgetRepresentation2D::GetSliceToWorldCoordinates(const do
 }
 
 //----------------------------------------------------------------------
+void vtkSlicerMarkupsWidgetRepresentation2D::GetBackgroundVolumeXYCenter(double centerPos[2])
+{
+  if (!this->GetSliceNode())
+  {
+    return;
+  }
+
+  // TODO: get RAS bounds of the scene from the view node
+  double centerPosRAS[4] = { 0.0, 0.0, 0.0, 1.0 };
+
+  double centerPosDisplay[4] = { 0.0, 0.0, 0.0, 1.0 };
+  this->GetWorldToSliceCoordinates(centerPosRAS, centerPosDisplay);
+
+  centerPos[0] = centerPosDisplay[0];
+  centerPos[1] = centerPosDisplay[1];
+}
+
+//----------------------------------------------------------------------
 void vtkSlicerMarkupsWidgetRepresentation2D::GetWorldToSliceCoordinates(const double worldPos[3], double slicePos[2])
 {
   vtkMRMLSliceNode* sliceNode = this->GetSliceNode();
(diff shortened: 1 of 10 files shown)


jamesobutler:
@lassoan Do you think it is time to go ahead with the integration of this new functionality?

I know our group has some potential interest in using this to generate functionality like https://www.imaios.com/en/vet-anatomy/mouse/mouse-whole-body within 3D Slicer for our types of images (Mouse 3D whole-body ultrasound). We would potentially use it as a training tool where on each slice offset it would show the annotation lines for all of the visible segmentations in that 2D slice view.

rbumm:
I would like to get this finished and really need this for annotation.

lassoan:
Sorry fro the delay, I need to make some improvements before merging. I'll try to work on it in the next couple of days.

rbumm:
Ron (who is visiting us in Chur) and I both agree. It would be fantastic if we could implement the first version of the 2D connectors and then proceed with additional development from there.

Modules/Loadable/Markups/VTKWidgets/vtkSlicerMarkupsWidgetRepresentation2D.cxx
```diff
@@ -98,6 +99,23 @@ vtkSlicerMarkupsWidgetRepresentation2D::ControlPointsPipeline2D::ControlPointsPi
   this->LabelsMapper->GetAnchorTransform()->SetCoordinateSystemToNormalizedViewport();
   this->LabelsActor = vtkSmartPointer<vtkActor2D>::New();
   this->LabelsActor->SetMapper(this->LabelsMapper);
+
+  // Labels line polydata
+  this->LabelsLinePolyData = vtkSmartPointer<vtkPolyData>::New();
+
+  this->LabelsLineProperty = vtkSmartPointer<vtkProperty2D>::New();
+  //this->LabelsLineProperty->SetColor(0.4, 1.0, 1.0);
```
lassoan, line 107: remove dead (commented out) code

Modules/Loadable/Markups/VTKWidgets/vtkSlicerMarkupsWidgetRepresentation2D.cxx
```diff
@@ -157,6 +175,24 @@ void vtkSlicerMarkupsWidgetRepresentation2D::GetSliceToWorldCoordinates(const do
   worldPos[2] = rasw[2]/rasw[3];
 }
 
+//----------------------------------------------------------------------
+void vtkSlicerMarkupsWidgetRepresentation2D::GetBackgroundVolumeXYCenter(double centerPos[2])
```
lassoan, line 179: We cannot assume that there is a volume in the scene (the leading lines should be correctly and consistently displayed if there is no volumes or many volumes in the scene). Instead of using the volume center as center of projection point, we could use the center of the bounding box of the markup's control points.

Modules/Loadable/Markups/VTKWidgets/vtkSlicerMarkupsWidgetRepresentation2D.cxx
```diff
@@ -182,6 +218,77 @@ void vtkSlicerMarkupsWidgetRepresentation2D::GetWorldToSliceCoordinates(const do
 }
 
 
+//----------------------------------------------------------------------
+void vtkSlicerMarkupsWidgetRepresentation2D::GetLabelTextXYOffsets(double backgroundVolumeCenterXY[2], double  slicePos[2], double leadersLineOffset[2])
```
lassoan, line 222: This method is way too overcomplicated. We don't need to deal with triangles, but coordinate geometry operations can be used instead. It could be as simple as something like this:

```
double leadLineDirectionVector[3];
vtkMath::Subtract(controlPoint, centerPoint, leadLineDirectionVector);
vtkMath::Normalize(leadLineDirectionVector);
double labelPoint[3] = 
  {
  controlPoint[0] + leadLineDirectionVector[0] * lineLength,
  controlPoint[1] + leadLineDirectionVector[1] * lineLength,
  controlPoint[2] + leadLineDirectionVector[2] * lineLength
  };
```

Modules/Loadable/Markups/VTKWidgets/vtkSlicerMarkupsWidgetRepresentation2D.cxx
```diff
@@ -200,6 +307,10 @@ void vtkSlicerMarkupsWidgetRepresentation2D::UpdateAllPointsAndLabelsFromMRML(do
     activeControlPointIndex = activeControlPointIndices[0];
     }
 
+  vtkNew<vtkPoints> pts;
```
lassoan, line 310: These variables should not be recreated at each call of the `UpdateAllPointsAndLabelsFromMRML` method (it would cause memory fragmentation and slowdown if you reallocate this array dozens of times per second, as you move a markup). Instead, the points and cells objects should be created once, where `LabelsLinePolyData` is created.

Modules/Loadable/Markups/VTKWidgets/vtkSlicerMarkupsWidgetRepresentation2D.cxx
```diff
@@ -288,13 +400,163 @@ void vtkSlicerMarkupsWidgetRepresentation2D::UpdateAllPointsAndLabelsFromMRML(do
...
+      labelPos[0] = slicePos[0] + labelsOffset / sqrt(2.0);
+      labelPos[1] = slicePos[1] + labelsOffset / sqrt(2.0);
+
+      std::string labelStr;
+      int labelLength = 0;
+
+      labelStr = markupsNode->GetNthControlPointLabel(pointIndex);
+
+      // make label texts multiline if <br> is used
+      std::string from = "<br>";
```
lassoan, line 419: This would be a very expensive operation when there are thousands of points. Do we really need this? It would be simpler and more efficient if we we made sure that the newline character (`\n`) can be properly entered on the GUI. This could be done in a separate pull request.

Modules/Loadable/Markups/VTKWidgets/vtkSlicerMarkupsWidgetRepresentation2D.cxx
```diff
@@ -288,13 +400,163 @@ void vtkSlicerMarkupsWidgetRepresentation2D::UpdateAllPointsAndLabelsFromMRML(do
...
+        int screenWidth = screenSize[0];
+        int screenHeight = screenSize[1];
+        labelLength = longestStringLen;
+
+        // Not working correctly - giving the exact same values in _bb  wherever you place this code
+        //double _bb[4] = { 0.0 };
+        //this->TextActor->GetBoundingBox(this->Renderer, _bb);
+        //this->TextActor->SetTextScaleModeToViewport();
+        double labelCharWidth =  this->MarkupsDisplayNode->GetTextScale() * 3.;
+        double labelTextWidth = labelCharWidth * labelLength;
```
lassoan, line 468: Getting the label size by counting characters is slow and inaccurate. It should be possible to get it from VTK. Check out how it is done in vtkLabelPlacementMapper or vtkDynamic2DLabelMapper. Maybe the best would be to fix vtkLabelPlacementMapper (that supports leader lines).

Modules/Loadable/Markups/VTKWidgets/vtkSlicerMarkupsWidgetRepresentation2D.cxx
```diff
@@ -288,13 +400,163 @@ void vtkSlicerMarkupsWidgetRepresentation2D::UpdateAllPointsAndLabelsFromMRML(do
...
+
+        pts->InsertPoint(ptcnt, lineStartPos[0], lineStartPos[1], 0.);
+        pts->InsertPoint(ptcnt + 1, lineEndPos[0], lineEndPos[1], 0.);
+        line->InsertNextCell(2);
+        line->InsertCellPoint(ptcnt);
+        line->InsertCellPoint(ptcnt + 1);
+        ptcnt += 2;
+
+        // make the connector line color a bit lighter then the selected glyph/text color
+        double color[3] = { 0,0,0 };
```
lassoan, line 545: Do not do all these operations for each point, but do these only once, after all the points are added.

Modules/Loadable/Markups/VTKWidgets/vtkSlicerMarkupsWidgetRepresentation2D.cxx
```diff
@@ -288,13 +400,163 @@ void vtkSlicerMarkupsWidgetRepresentation2D::UpdateAllPointsAndLabelsFromMRML(do
...
+        this->MarkupsDisplayNode->GetSelectedColor(color);
+        double hsv[3] = { 0,0,0 };
+        vtkMath::RGBToHSV(color, hsv);
+        hsv[0] *= 0.8;
+        hsv[1] *= 0.3;
+        vtkMath::HSVToRGB(hsv, color);
+        controlPoints->LabelsLineProperty->SetColor(color);
+        controlPoints->LabelsLineProperty->SetLineWidth(this->ControlPointSize * .25);
```
lassoan, line 553: 0.25 is very thick. It might be better to use something like half of the line thickness, but even better would be to use a 2D actor. Maybe the best would be to fix vtkLabelPlacementMapper.

Modules/Loadable/Markups/VTKWidgets/vtkSlicerMarkupsWidgetRepresentation2D.cxx
```diff
@@ -319,6 +581,10 @@ void vtkSlicerMarkupsWidgetRepresentation2D::UpdateAllPointsAndLabelsFromMRML(do
     controlPoints->LabelControlPointsPolyData->GetPointData()->GetNormals()->Modified();
     controlPoints->LabelControlPointsPolyData->Modified();
 
+    controlPoints->LabelsLinePolyData->Modified();
+    controlPoints->LabelsLinePolyData->SetPoints(pts);
```
lassoan, line 585: set these only once, when the polydata is created

lassoan (commented):
There are a number of issues to fix  before this could be merged - see the comments inline. I've only commented on the 2D representation but all the comments apply to the 3D representation, too. To reduce code duplication, it would be nice to move all the similar functions to the vtkSlicerMarkupsWidgetRepresentation base class.

lassoan:
@rbumm I know that you worked a lot on this and waited for long for feedback so you would like to see this merged as soon as possible. But unfortunately the issues described [above](https://github.com/Slicer/Slicer/pull/6809#pullrequestreview-1525341507) and in commit comments should be addressed to avoid performance issues and to make sure the code works correctly in general (not just for some specific use cases). Do you think you'll have time to work on this?

rbumm:
Likely, yes. However, I must reestablish the Slicer build process, which I anticipate to be a somewhat arduous and time-consuming task before all functionalities are restored. I understand my expertise is just at the threshold required for implementing these central Slicer features. Therefore, I would greatly benefit from prompt feedback on my commits to ensure I'm progressing in the correct direction. My timeline allows for initiating this process no earlier than next week.

rbumm:
Will it be possible to integrate the functionality into the next stable version?

rbumm:
I just noticed the new 'Label' markup in the SlicerExtraMarkups extension, which is working reasonably well. The way the extension adds new markup types makes sense to me. And the concept of manually choosing label position and markup position in "Label" makes it quite flexible.  

@lassoan  please share your thoughts on that project and does it make sense to carry on with this PR, which seems already "late"

PR #6810, opened 2023-02-01
Fix welcome module translation

lassoan:
Welcome module contained text in image (large Welcome label at the top) and non-translatable html resources (collapsible sections). These were replaced by content in the .ui file that can be translated.

image


jcfr (commented):
The simplification makes sense. Now that the information in each sections are simpler, there is no need to use html pages.

After building, I was able to test and it works well.

### Suggest changes

* The file [HTML/stylesheet.css](https://github.com/Slicer/Slicer/blob/main/Modules/Loadable/SlicerWelcome/Resources/HTML/stylesheet.css) should be removed and the associated `qrc` file updated.

* It seems the first commit (`BUG: Replace large "Welcome" image by text`) only removes the icon without updating the code. Since the code is updated in the next commit, changes could either be squashed or better split.

### Notes

Attempting to edit the content with the designer also seems to work well if `Rich Text` mode is enabled.

That said, the generated "html" is very rebase and not maintainable but in the context of the welcome module this is not an issue.

lassoan:
Thanks for the review @jcfr, I've updated the commit based on your comments.

PR #6817, opened 2023-02-08
STYLE: Fix Python lint errors

lassoan:


cpinter:
Oops, I need to get used to this change in indentation, sorry. All the other projects we have still use 2 spaces so the IDEs are set to that...

PR #6818, opened 2023-02-09
BUG: Fix issue with copying curve with CopyContent

cpinter:
When a curve markup is copied using CopyContent the vtkMRMLMarkupsNode::RemoveAllMeasurements removes all measurements but the observations of the special curve measurements (calculated by vtkCurveMeasurementsCalculator) are not re-added. This commit adds these observations on CopyContent.


lassoan (approved):

PR #6819, opened 2023-02-09
BUG: Fixes default progress callback raises user warning in send() function

marcus-wirtz-snkeos:
Since the return argument of the `defaultProgressCallback()` is `None`, the `send()` function of the `DICOMSender` raises a UserWarning:
```python
if not self.progressCallback(f"Sending {file} to {self.destinationUrl.toString()} using {self.protocol}"):
```

This minimal PR prevents the user warning by returning True instead.

### DIFF
--- a/Modules/Scripted/DICOMLib/DICOMProcesses.py
+++ b/Modules/Scripted/DICOMLib/DICOMProcesses.py
@@ -418,8 +418,10 @@ class DICOMSender(DICOMProcess):
     def __del__(self):
         super().__del__()
 
-    def defaultProgressCallback(self, s):
+    @staticmethod
+    def defaultProgressCallback(s):
         logging.debug(s)
+        return True
 
     def send(self):
         self.progressCallback("Starting send to %s using self.protocol" % self.destinationUrl.toString())


Modules/Scripted/DICOMLib/DICOMProcesses.py
```diff
@@ -418,8 +418,10 @@ def __init__(self, files, address, protocol=None, progressCallback=None, aeTitle
     def __del__(self):
         super().__del__()
 
-    def defaultProgressCallback(self, s):
+    @staticmethod
```
lassoan, line 421: Callback is usually not static method, as it uses other methods of the object to display the progress information, determine if the operation should be cancelled, etc. What was the reason for making it staticmethod?

lassoan (commented):
Thank you, good catch, returning True is the right thing to do.

You need to fix the commit message (it must start with a capital letter after `BUG: `) and please also see the comment on the necessity of changing the callback to a static method.

Modules/Scripted/DICOMLib/DICOMProcesses.py
marcus-wirtz-snkeos: True, this particular function was not using `self`, that was the only reason. But I agree that it should inherit the same architecture -> I'll update this.

marcus-wirtz-snkeos:
@lassoan sorry, I messed up the commit history by merging main, here is the updated PR:
https://github.com/Slicer/Slicer/pull/6822

PR #6822, opened 2023-02-10
BUG: Fixes default progress callback raises user warning in send() function

marcus-wirtz-snkeos:
Since the return argument of the defaultProgressCallback() is None, the send() function of the DICOMSender raises a UserWarning:

```python
if not self.progressCallback(f"Sending {file} to {self.destinationUrl.toString()} using {self.protocol}"):
```

This minimal PR prevents the user warning by returning True instead.


lassoan (approved):
Thank you for the fix!

I reviewed how we report progress and request cancel in DICOMLib and found that in DICOMProcesses.py the return value means "continue", while in all other places it means "request cancel". In the future, we'll probably rework the progress callbacks to request cancel by raising an exception, but for now this quick fix is appropriate.

PR #6824, opened 2023-02-11
BUG: Free notifier after request handling

### LINKED ISSUE #6823: Slicer crashing when using custom FileReader using slicerio/SlicerWeb interface
## Summary

We are using a custom file loader written in python that loads images and segmentations from a file.
To look at a list of files, we repeatedly clear the scene and then call the FileReader to open our custom files (which are based on h5) by using the slicerio.server package. After around 20 iterations (this can vary, also depending on the data that is loaded) of clearing the scene and loading another file, Slicer crashes with a Access violation error or similar. The errors are not always the same, but I have attached screenshots of two of them below which I obtained by building slicer in Debug mode (in the Release mode it just freezes).
The same error does not happen when using the File Reader from the Python console in Slicer (I was able to load the same data >100 times without any error). The same code from the Python console executed using the Slicer Web exec api crashes as well after 15-20 iterations. I was so far unable to reproduce the issue if I only loaded the image, no segmentations, but it might just be that this is because of the amount of data loaded. Thank you for any hints on what could cause this!

## Steps to reproduce

The code I used to test this is in the python console is:
`[(slicer.mrmlScene.Clear(),slicer.util.loadNodesFromFile("temp.h5",properties={ "segment_keys":'["Segment1","Segment2","Segment3"]',    "image_key":"image"},    filetype="H5File"),slicer.app.processEvents(),time.sleep(1)) for _ in range(100)]`

The code I used for Slicer Web is:
```
import requests
from pathlib import Path
tempfile_path = Path("temp.h5").resolve().as_posix()
command = '(slicer.mrmlScene.Clear(),slicer.util.loadNodesFromFile(\''+ tempfile_path+ '\',properties={ \'segment_keys\':\'["Segment0", "Segment1", "Segment2", "Segment3"]\',\'image_key\':\'image\'},filetype=\'H5File\'))'
for i in range(100):
    print(requests.post(url="http://localhost:2016/slicer/exec",data=command))
    print(i)
```

The test file you can use is here, needs to be unpacked first: (It's just a h5 file with zeros): [temp.zip](https://github.com/Slicer/Slicer/files/10709197/temp.zip)
The file loader uses this code: 
```
import numpy as np
import slicer
import json

try:
    import h5py
except ModuleNotFoundError:
    slicer.util.pip_install("h5py")
    import h5py


class H5Visualizer(ScriptedLoadableModule):
    """Uses ScriptedLoadableModule base class, available at:
    https://github.com/Slicer/Slicer/blob/master/Base/Python/slicer/ScriptedLoadableModule.py
    """

    def __init__(self, parent):
        self.parent.title = "H5Visualizer"
        self.parent.categories = ["Debug"]
        self.parent.dependencies = []
        self.parent.contributors = ["TEST"]
        self.parent.helpText = "Debug"
        self.parent.acknowledgementText = ""


class H5VisualizerFileReader(object):
    def __init__(self, parent):
        self.parent = parent
        self.slice_space_key = "sliceSpace"
        self.colors = [
            "e6194b",  # Red
            "3cb44b",  # Green
            "ffe119",  # Yellow
            "4363d8",  # Blue
            "f58231",  # Orange
            "911eb4",  # Purple
            "42d4f4",  # Cyan
            "f032e6",  # Magenta
        ]

    def description(self):
        return "H5 files"

    def fileType(self):
        return "H5File"

    def extensions(self):
        return ["H5 Files (*.h5)"]

    def canLoadFile(self, filePath):
        return filePath.endswith(".h5")

    def load_h5(self, h5_path, key_image, key_segment=None):
        with h5py.File(h5_path, "r") as open_h5:
            # load slice space information from h5 file
            slice_space = np.identity(4)

            # load image from h5-file
            image = np.squeeze(open_h5[key_image][()])
            with slicer.util.RenderBlocker():
                # create volume node from image data
                volume_node = slicer.util.addVolumeFromArray(
                    image,
                    ijkToRAS=slice_space.T,
                    name="Volume",
                    nodeClassName="vtkMRMLScalarVolumeNode",
                )

                slicer.util.setSliceViewerLayers(background=volume_node)
                slicer.util.resetSliceViews()

                # Add a new segmentation node and link it to the new volume node
                loaded_nodes = [volume_node]
                if key_segment is not None and len(key_segment) > 0:
                    segm_node = slicer.mrmlScene.AddNewNodeByClass(
                        "vtkMRMLSegmentationNode", "Segmentation"
                    )
                    with slicer.util.NodeModify(segm_node):
                        loaded_nodes.append(segm_node)
                        segm_node.CreateDefaultDisplayNodes()
                        segm_node.SetReferenceImageGeometryParameterFromVolumeNode(
                            volume_node
                        )

                        for i, key in enumerate(key_segment):
                            # Add an empty segment for the mask and fill it with content from h5 file
                            color = self.colors[i % len(self.colors)]
                            color_hex = tuple(
                                int(color[i : i + 2], 16) / 255.0 for i in (0, 2, 4)
                            )
                            # create empty oriented ImageData from above defined volume node
                            oriented_image_data = slicer.modules.segmentations.logic().CreateOrientedImageDataFromVolumeNode(
                                volume_node
                            )
                            segm_node.AddSegmentFromBinaryLabelmapRepresentation(
                                oriented_image_data, key, color_hex
                            )
                            segment = open_h5[key][()]
                            slicer.util.updateSegmentBinaryLabelmapFromArray(
                                segment, segm_node, key, volume_node
                            )

                node_ids = [n.GetID() for n in loaded_nodes]
                return node_ids

    def load(self, properties):
        filePath = properties["fileName"]
        image_key = properties.get("image_key", None)
        segments_json = properties.get("segment_keys", None)
        segments = json.loads(segments_json)
        self.parent.loadedNodes = self.load_h5(filePath, image_key, segments)

        return True

```

The errors that appear during this are (most of the time) one of the two following:
image
image

## Environment
- Slicer version: Slicer-5.3.0-2023-02-01

pieper:
If the request included `processEvents()` or otherwise triggered the event loop the notifier could be
freed before the slot was finished, causing memory corruption.

Fixes #6823

### DIFF
--- a/Modules/Scripted/WebServer/WebServer.py
+++ b/Modules/Scripted/WebServer/WebServer.py
@@ -383,7 +383,6 @@ class SlicerHTTPServer(HTTPServer):
                 self.logMessage('Got complete message of header size %d, body size %d' % (len(requestHeader), len(requestBody)))
                 self.readNotifier.disconnect('activated(int)', self.onReadable)
                 self.readNotifier.setEnabled(False)
-                qt.QTimer.singleShot(0, self.onReadableComplete)
 
                 if len(self.requestSoFar) == 0:
                     self.logMessage("Ignoring empty request")
@@ -461,6 +460,10 @@ class SlicerHTTPServer(HTTPServer):
                 fileno = self.connectionSocket.fileno()
                 self.writeNotifier = qt.QSocketNotifier(fileno, qt.QSocketNotifier.Write)
                 self.writeNotifier.connect('activated(int)', self.onWritable)
+
+                # free notifier after handling request
+                # https://github.com/Slicer/Slicer/issues/6823
+                qt.QTimer.singleShot(0, self.onReadableComplete)
 
         def onWriteableComplete(self):
             self.logMessage("writing complete, freeing notifier")
--- a/Modules/Scripted/WebServer/WebServer.py
+++ b/Modules/Scripted/WebServer/WebServer.py
@@ -294,8 +294,6 @@ class SlicerHTTPServer(HTTPServer):
         self.socket.settimeout(5.)
         if logMessage:
             self.logMessage = logMessage
-        self.notifiers = {}
-        self.connections = {}
         self.requestCommunicators = {}
         self.enableCORS = enableCORS
 
@@ -333,10 +331,6 @@ class SlicerHTTPServer(HTTPServer):
         def registerRequestHandler(self, handler):
             self.requestHandlers.append(handler)
             handler.logMessage = self.logMessage
-
-        def onReadableComplete(self):
-            self.logMessage("reading complete, freeing notifier")
-            self.readNotifier = None
 
         def onReadable(self, fileno):
             self.logMessage('Reading...')
@@ -461,15 +455,6 @@ class SlicerHTTPServer(HTTPServer):
                 self.writeNotifier = qt.QSocketNotifier(fileno, qt.QSocketNotifier.Write)
                 self.writeNotifier.connect('activated(int)', self.onWritable)
 
-                # free notifier after handling request
-                # https://github.com/Slicer/Slicer/issues/6823
-                qt.QTimer.singleShot(0, self.onReadableComplete)
-
-        def onWriteableComplete(self):
-            self.logMessage("writing complete, freeing notifier")
-            self.writeNotifier = None
-            self.connectionSocket = None
-
         def onWritable(self, fileno):
             self.logMessage('Sending on %d...' % (fileno))
             sendError = False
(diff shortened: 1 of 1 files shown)


lassoan (commented):
There are a couple of laces where the function can return earlier. Shouldn't we call self.onReadableComplete in all those cases?

lassoan:
Once we have the final implementation, we'll need to backport the fix to the Slicer-5.2.2 patch release.

pieper:
> There are a couple of places where the function can return earlier. Shouldn't we call self.onReadableComplete in all those cases?

Good point.  Looking at it again I wonder if that code even needs to be there.  These are member variables of the `SlicerRequestCommunicator` which is held in the server until a new connection is made on the same fileno so it should be freed at that point.  The whole connection array would go out of scope when the server instance is freed, so I'm thinking now it's okay to just remove this code.

lassoan (approved):
Thank you, it looks good to me. You may consider squashing before merging.

PR #6828, opened 2023-02-14
Make loadable module title translatable

lassoan:
This pull request changes how loadable module titles are specified (they are now set in the C++ source code instead of in CMakeLists.txt) to make them translatable. The pull request also contains a few related minor fixes for the problems found during developing this.

### DIFF
--- a/Base/QTCore/qSlicerCoreApplication.cxx
+++ b/Base/QTCore/qSlicerCoreApplication.cxx
@@ -452,6 +452,22 @@ void qSlicerCoreApplicationPrivate::init()
 
   this->createDirectory(q->extensionsInstallPath(), "extensions"); // Make sure the path exists
 
+  this->ApplicationLocaleName = "en_US";
+  this->ApplicationLocale = QLocale(ApplicationLocale);
+#ifdef Slicer_BUILD_I18N_SUPPORT
+  if (q->userSettings()->value("Internationalization/Enabled").toBool())
+    {
+    QString localeName = q->userSettings()->value("language", this->ApplicationLocaleName).toString();
+    if (!localeName.isEmpty())
+      {
+      this->ApplicationLocaleName = localeName;
+      this->ApplicationLocale = QLocale(ApplicationLocale);
+      }
+    // We load the language selected for the application
+    qSlicerCoreApplication::loadLanguage();
+    }
+#endif
+
   // Prevent extensions manager model from displaying popups during startup (don't ask for confirmation)
   bool wasInteractive = model->interactive();
   model->setInteractive(false);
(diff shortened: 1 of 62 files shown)


Extensions/Testing/LoadableExtensionTemplate/LoadableModuleTemplate/CMakeLists.txt
```diff
@@ -48,7 +47,6 @@ set(MODULE_RESOURCES
 #-----------------------------------------------------------------------------
 slicerMacroBuildLoadableModule(
   NAME ${MODULE_NAME}
-  TITLE ${MODULE_TITLE}
```
pieper, line 51: It's definitely an improvement to specify the module title in the C++ code with tr, and also to clean up the template and usage in the core.  From what I can tell older loadable modules that use the old style will still work but won't be translatable.  I wonder if we should add a warning message in `CMake/SlicerMacroBuildLoadableModule.cmake` if `TITLE` is passed, with a description of how to update the code to make it localizable.

pieper (approved):
Looks good to me 👍 Just some comments but I don't see changes needed.

Extensions/Testing/LoadableExtensionTemplate/LoadableModuleTemplate/CMakeLists.txt
lassoan: Good point, we should deprecate this option at some point. Now I'll just add a note to the translator's guide.

lassoan: I've added a note here: https://github.com/Slicer/SlicerLanguagePacks/blob/main/DevelopersManual.md#translating-module-title

Base/QTCore/qSlicerCoreApplication.cxx
```diff
@@ -452,6 +452,22 @@ void qSlicerCoreApplicationPrivate::init()
 
   this->createDirectory(q->extensionsInstallPath(), "extensions"); // Make sure the path exists
 
+  this->ApplicationLocaleName = "en_US";
+  this->ApplicationLocale = QLocale(ApplicationLocale);
+#ifdef Slicer_BUILD_I18N_SUPPORT
+  if (q->userSettings()->value("Internationalization/Enabled").toBool())
+    {
+    QString localeName = q->userSettings()->value("language", this->ApplicationLocaleName).toString();
```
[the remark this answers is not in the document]
pieper: Yes, I think all those options could be useful, but I'm not sure which are strictly required.  It might be enough to have a command line option, but maybe also we could automatically inherit the OS locale, at least as a default.  I agree it's a new feature that needs discussion.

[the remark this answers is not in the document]
pieper: New feature request filed: https://github.com/Slicer/Slicer/issues/6833

PR #6829, opened 2023-02-15
Parameter node wrapper template key update

Connor-Bowley:
Updates the TemplateKey to use the parameterNodeWrapper, including its gui connection capabilities.

Makes a few other small parameterNodeWrapper related updates.

### DIFF
--- a/Base/Python/slicer/parameterNodeWrapper/default.py
+++ b/Base/Python/slicer/parameterNodeWrapper/default.py
@@ -6,15 +6,28 @@ class Default:
     Annotation to denote the default value for a parameter.
     """
 
-    def __init__(self, value):
-        self.value = value
+    def __init__(self, value=None, *, generator=None):
+        """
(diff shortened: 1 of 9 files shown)


allemangD:
Ah, the `WithinRange` behavior fix is not so straightforward as I thought. It must behave correctly with spinboxes as well, where infinite bounds _do_ make sense. @Connor-Bowley do you have thoughts on this?

Connor-Bowley:
Per discussion, a fix was applied that ensures `Minimum` and Maximum are not used if `WithinRange` is present.

Additionally, an exception is raised if trying to bind a `QSlider` or `ctkSliderWidget`, which by definition should be bounded, to an unbounded `int` or `float`.

allemangD (approved):
That warning is gone with the template module, and looks like those other issues we addressed are handled correctly.

LGTM! :rocket:

PR #6790, opened 2023-01-20
COMP: Fix vtkMarkupsAnnotationSceneTest build with SceneViews module disabled

jcfr:
This commit fixes build of Slicer (or Slicer custom application) where the
`SceneViews` module has been disabled while having Markups module enabled.

It partially reverts changes integrated in https://github.com/Slicer/Slicer/commit/f44849b014ecf9a245901239db64037556110970 (ENH: Remove Annotations
module).

Fixes https://github.com/Slicer/Slicer/issues/6789

PR #6834, opened 2023-02-23
COMP: Fix vtkMarkupsAnnotationSceneTest build with SceneViews module disabled (part2)

jcfr:
This is a follow-up of d4723511c effectively fixing the build by ensuring the SceneViews module logic library is not linked against the Markups tests.

This further improve change initially integrated in:
* https://github.com/Slicer/Slicer/pull/6790


jcfr:
Merging. Change locally tested in a custom application build (with `SceneViews` module disabled) and I confirm the build completes.

cc: @tbirdso

tbirdso:
Great! It would be nice if this infrastructure could be extended beyond Markups and SceneView modules in the future so that these resolutions do not need to happen on a case-by-case basis, but I understand that that is a larger effort beyond our present scope.

PR #6835, opened 2023-02-23
BUG: Create non-existing segments in ImportLabelmapToSegmentationNode

Sunderlandkyl:
Previously ImportLabelmapToSegmentationNode would only import segments if the specified list of segment IDs matched pre-existing segments, and ignore other segments. This commit will insert new segments that don't match an existing segment ID.

See https://discourse.slicer.org/t/slicer-modules-segmentations-logic-importlabelmaptosegmentationnode-not-workign-with-given-segmentids/27998.


lassoan (approved):
It looks good, thank you.

Issue #6827, opened 2023-02-13
Color legend display not working as expected in all slice views


jamesobutler:
## Summary

As originally mentioned at https://discourse.slicer.org/t/issue-displaying-color-legend-in-other-slice-views/27784, the color legend display is not working as expected in some of the other slice views other than Red/Yellow/Green. 

## Steps to reproduce

```python
# Start Slicer so that on startup it is the Four-Up Layout
slicer.app.layoutManager().setLayout(slicer.vtkMRMLLayoutNode.SlicerLayoutThreeOverThreeView)

import SampleData
sampleDataLogic = SampleData.SampleDataLogic()
mrHead = sampleDataLogic.downloadMRHead()

color_legend_display_node = slicer.modules.colors.logic().AddDefaultColorLegendDisplayNode(mrHead)
color_legend_display_node.SetVisibility(True)
```
If you start Slicer where it begins in the the Three-over-three layout, then the color legend display shows as expected. Something about the views on startup has some impact here.

## Expected behavior

The color legend display should turn on/off for the specified volume in all slice views that the volume is displayed in. It should not be tied to Slice visibility in 3D either.

## Current workaround
```python
slicer.app.layoutManager().setLayout(slicer.vtkMRMLLayoutNode.SlicerLayoutThreeOverThreeView)

import SampleData
sampleDataLogic = SampleData.SampleDataLogic()
mrHead = sampleDataLogic.downloadMRHead()

# Workaround - set slice view visibles before turning on color legend display
layoutManager = slicer.app.layoutManager()
for sliceViewName in layoutManager.sliceViewNames():
  view = layoutManager.sliceWidget(sliceViewName).sliceView()
  sliceNode = view.mrmlSliceNode()
  sliceNode.SetSliceVisible(True)

color_legend_display_node = slicer.modules.colors.logic().AddDefaultColorLegendDisplayNode(mrHead)
color_legend_display_node.SetVisibility(True)
```

## Environment
- Slicer version: Slicer-5.3.0-2022-02-10 and Slicer 5.2.1.
- Operating system: Windows

lassoan:
Probably a fix could be to update visibility of the color legends when a new slice composite node is added (in vtkMRMLColorLegendDisplayableManager::OnMRMLSceneNodeAdded) or when the layout is changed.

MichaelColonel:
I will try to fix the issue.

MichaelColonel:
Problem when changing the layout in the color legend displayable manager:

[...]

PR #6836, opened 2023-02-24
BUG: Fix color legend visibility on non-standard slice views

MichaelColonel:
Possible fix for issue #6827. Some minor bugs can be still present.

### DIFF
--- a/Libs/MRML/Widgets/qMRMLLayoutManager.cxx
+++ b/Libs/MRML/Widgets/qMRMLLayoutManager.cxx
@@ -253,13 +253,19 @@ QWidget* qMRMLLayoutSliceViewFactory::createViewFromNode(vtkMRMLAbstractViewNode
   Q_ASSERT(!this->viewWidget(viewNode));
 
   qMRMLSliceWidget * sliceWidget = new qMRMLSliceWidget(this->layoutManager()->viewport());
+
+  // Set slice logic before setting the slice node in the widget
+  // to allow displayable managers to use the slice logic during initialization
+  // For example, without this the color legend displayable manager would not be able to correctly
+  // initialize new views when switching to a new view layout that has more slice views.
+  sliceWidget->setSliceLogics(this->sliceLogics());
+  this->sliceLogics()->AddItem(sliceWidget->sliceLogic());
+
   sliceWidget->sliceController()->setControllerButtonGroup(this->SliceControllerButtonGroup);
   sliceWidget->setObjectName(QString("qMRMLSliceWidget%1").arg(viewNode->GetLayoutName()));
   // set slice node before setting the scene to allow using slice node names in the slice transform, display, and model nodes
   sliceWidget->setMRMLSliceNode(vtkMRMLSliceNode::SafeDownCast(viewNode));
   sliceWidget->setMRMLScene(this->mrmlScene());
-  sliceWidget->setSliceLogics(this->sliceLogics());
-  this->sliceLogics()->AddItem(sliceWidget->sliceLogic());
 
   return sliceWidget;
 }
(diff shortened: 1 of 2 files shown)


lassoan (commented):
Thanks a lot for working on this. Your fix was very useful in showing what went wrong and a potential way to address the problem. However, I've found that a simpler fix could be to ensure that the slice logic is available by the time the displayable manager is initialized. I've pushed that fix on your branch. It would be great if you could give it some testing to see if it fixes all color legend initialization issues related to layout switching and if it does not introduce any regressions.

If everything works well then we'll have to squash the commits and merge.

MichaelColonel:
Everything works correctly except one general workflow pattern when color legend isn't visible: Select Layout -> Load Volume -> Create Color Legend -> Select other layout with more slice views -> Add volume to an any newly added slice.

``` python
# Start Slicer so that on startup it is the Four-Up Layout
slicer.app.layoutManager().setLayout(slicer.vtkMRMLLayoutNode.SlicerLayoutFourUpView)

import SampleData
mrHead = sampleDataLogic.downloadMRHead()

clDisplayNode = slicer.modules.colors.logic().AddDefaultColorLegendDisplayNode(mrHead)
clDisplayNode.SetVisibility(True)


slicer.app.layoutManager().setLayout(slicer.vtkMRMLLayoutNode.SlicerLayoutThreeOverThreeView)

for color in ["Red+", "Yellow+"]:
  slicer.app.layoutManager().sliceWidget(color).sliceLogic().GetSliceCompositeNode().SetForegroundVolumeID(mrHead.GetID())
```
Screenshot1

If delete a color legend and recreate it again, the color legend displayed as it should.

``` python
# Start Slicer so that on startup it is the Four-Up Layout
slicer.app.layoutManager().setLayout(slicer.vtkMRMLLayoutNode.SlicerLayoutFourUpView)

import SampleData
mrHead = sampleDataLogic.downloadMRHead()

clDisplayNode = slicer.modules.colors.logic().AddDefaultColorLegendDisplayNode(mrHead)
clDisplayNode.SetVisibility(True)


slicer.app.layoutManager().setLayout(slicer.vtkMRMLLayoutNode.SlicerLayoutThreeOverThreeView)

for color in ["Red+", "Yellow+"]:
  slicer.app.layoutManager().sliceWidget(color).sliceLogic().GetSliceCompositeNode().SetForegroundVolumeID(mrHead.GetID())


clDisplayNode.GetScene().RemoveNode(clDisplayNode)
clDisplayNode = slicer.modules.colors.logic().AddDefaultColorLegendDisplayNode(mrHead)
clDisplayNode.SetVisibility(True)
```
Screenshot2

jamesobutler:
@lassoan I can confirm that your current changes fixed the situation I described in #6827. I also can replicate the issue that @MichaelColonel observed in the above comment. Do you know what is going wrong in this situation?

MichaelColonel:
After some additional tests and debugging the main problem remains. The `SliceCompositeNode` is invalid and color legend actor is updated earlier than non-standard slice composite node were added to the scene.

MichaelColonel:
I've update this PR a little, so when one set all manually, Slicer shows color legend for non-standard views correctly.
It doesn't work from the script though.
So for the script rule of thumb should be:
1. Setup slice layout;
2. Everything else.

lassoan:
> updated earlier than non-standard slice composite node were added to the scene.

Would observing "node added" events solve the issue? We could update the display pipeline when we detect that a composite node is added to the scene.

MichaelColonel:
> Would observing "node added" events solve the issue?

I will try it later.

I've made some minor changes, and now color legend should be displayed correctly in 3DSlicer, except execution from the script.

MichaelColonel:
> We could update the display pipeline when we detect that a composite node is added to the scene.

How it can be done, i mean the pipeline update? Another question. What event is invoked when volume node is added for removed from the SliceCompositeNode foreground or background?

jamesobutler (approved):
I can confirm the current changes in this PR
- Fix the issue that the script provided in https://github.com/Slicer/Slicer/issues/6827#issue-1582742933 previously caused
- The issue described in https://github.com/Slicer/Slicer/pull/6836#issuecomment-1447957273 still exists.

@lassoan Since the current changes fix my original issue, I think it would be good to integrate these changes as part of continuous improvement. It is better to have some fix than no fix. It appears questions still arise on how to address the second issue that still exists here. A new GitHub issue can be generated for this remaining work or you can issue a change if you see an obvious set of next steps.

PR #6840, opened 2023-02-25
DOC: Clarify that GetTransformBetweenNodes is always in PostMultiply mode

mauigna06:


lassoan (commented):
Thank you, this may help users.

We don't make a commitment that the transform will be pre or post multiply, so please update the text to something like that "the method may change the PreMulitiply/PostMulitply flags in the transform". Thank you!

Libs/MRML/Core/vtkMRMLTransformNode.h
```diff
@@ -120,29 +120,35 @@ class VTK_MRML_EXPORT vtkMRMLTransformNode : public vtkMRMLDisplayableNode
...
+  /// Resulting transformToWorld multiply order will be overwritten to PostMultiply
   /// \sa GetTransformBetweenNodes
   void GetTransformToWorld(vtkGeneralTransform* transformToWorld);
 
   ///
   /// Get concatenated transforms from world.
+  /// Resulting transformFromWorld multiply order will be overwritten to PostMultiply
   /// \sa GetTransformBetweenNodes
   void GetTransformFromWorld(vtkGeneralTransform* transformToWorld);
+  void GetTransformFromWorld(vtkGeneralTransform* transformFromWorld);
```
lassoan, line 132: Is this change intentional?

mauigna06: yes

lassoan: This line causes a build error. You need to remove the old line.

mauigna06: Oh, I redid the commit so many times that that slipped through

lassoan (approved):

PR #6842, opened 2023-02-27
ENH: Allow line/angle Markups line visibility if control points hidden

Sunderlandkyl:
Previously, the line/arc in line and angle Markups was forced to be hidden if the visibility of any of the control points was disabled. This commit removes the GetAllControlPointsVisible() requirement so that line/angle markups can be displayed even when some of the control points are hidden.


lassoan (approved):
The current behavior was intentional, but without any strong motivation. If you have use cases when showing just the lines (without showing the points) is beneficial then let's do it.

Please add to the commit comment a description of your use case(s) and why the current behavior was bad. This can help in the future if somebody questions this decision and wants to bring back the old behavior.

Sunderlandkyl:
I added a the message to the commit.

There is a workaround by setting line thickness to 100%, but that approach is not intuitive compared to hiding the control points. It also causes display artifacts where the control points/line overlap when the line is transparent.

I realize that the existing behavior was intentional, though I'm not sure what the use-case for it is.

lassoan (approved):
Thank you.

PR #6843, opened 2023-02-27
Parameter node wrapper gui creation

Connor-Bowley:
Implements a GUI creation infrastructure that builds on the existing parameter node wrapper and its GUI connection infrastructure


allemangD (approved):
LGTM; the structure seems the same as we had already discussed so no comment there.

PR #6849, opened 2023-03-01
ENH: Add Soft edge option for Mask volume Segment Editor effect

lassoan:
This reduces edge artifacts on the resulting masked image.

Original behavior (no soft edge):

image

Soft edge (less edge artifacts):

image

Soft edge (very wide edge, even less edge artifacts but the transition zone somewhat eats into the image):

image

### DIFF
--- a/Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorMaskVolumeEffect.py
+++ b/Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorMaskVolumeEffect.py
@@ -102,6 +102,17 @@ Fill inside and outside operation creates a binary labelmap volume as output, wi
         fillValuesSpinBoxLayout.addRow(fillValueLayout)
         self.scriptedEffect.addOptionsWidget(fillValuesSpinBoxLayout)
 
+        # Soft edge
+        self.softEdgeMmSpinBox = slicer.qMRMLSpinBox()
+        self.softEdgeMmSpinBox.setMRMLScene(slicer.mrmlScene)
+        self.softEdgeMmSpinBox.setToolTip("Standard deviation of the Gaussian function that blurs the edge of the mask."
+                                          " Higher value makes the edge softer.")
+        self.softEdgeMmSpinBox.quantity = "length"
+        self.softEdgeMmSpinBox.value = 0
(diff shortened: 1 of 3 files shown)


Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorMaskVolumeEffect.py
```diff
@@ -335,31 +358,82 @@ def maskVolumeWithSegment(segmentationNode, segmentID, operationMode, fillValues
...
+            # because it would result in edge artifacts.
+            # Larger value results in less edge artifact but increased computation time,
+            # so 3.0 is a good tradeoff.
+            gaussianFilter.SetRadiusFactor(3.0)
+            gaussianFilter.Update()
+
+            import vtk.util.numpy_support
+            maskImage = gaussianFilter.GetOutput()
+            nshape = tuple(reversed(maskImage.GetDimensions()))
+            maskArray = vtk.util.numpy_support.vtk_to_numpy(maskImage.GetPointData().GetScalars()).reshape(nshape)
```
pieper, line 417: I feel like we should have utility method for this, since it shows up a lot.

pieper: Perhaps `slicer.util.arrayFromVTKImageData` and `slicer.util.arrayFromVTKImageDataModified`

pieper (approved):
This is nice!

Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorMaskVolumeEffect.py
lassoan: Converting a vtkImageData to an ITK-compatible xarray would be indeed nice. However, this would only work well if vtkImageData stored the image geometry (origin, spacing, directions), which we don't do yet in Slicer.

pieper: No, I just meant that these 4 lines to get a numpy array of a vtkImageData shows up in several places and is kind of ugly.  Implementing xarray would be a different thing, but also good.

PR #6852, opened 2023-03-02
COMP: Report error if unsupported CMake version 3.25.(0|1|2) is used

jcfr:
Due to a regression introduced in CMake 3.25 that will be fixed in CMake 3.25.3, this commit ensures an error message is reported.

The regression was preventing the successful build and installation of the PythonQt external project integrated through CTK.

References:
* https://gitlab.kitware.com/cmake/cmake/-/issues/24567
* https://gitlab.kitware.com/cmake/cmake/-/merge_requests/7655
* https://discourse.slicer.org/t/slicer-custom-app-build-error/28026/8


sjh26 (approved):
lgtm

PR #6854, opened 2023-03-03
DOC: Fix outdated sequence browser example usage

jamesobutler:
I saw this outdated Sequence Browser module usage (`slicer.modules.sequencebrowser`) in the sequences documentation so I figured I would go ahead and issue a minor change for it.


lassoan (approved):
Thank you!

PR #6863, opened 2023-03-07
BUG: Allow multiple DICOM plugins for export with same mean confidence

Sunderlandkyl:
Previously only one plugin was stored within the QMap for a given mean confidence value. This commit changes the QMap to a QMultiMap so that multiple plugins with the same confidence can be displayed.

See discussion here: https://discourse.slicer.org/t/exporting-segmentation-as-dcm/28168

### DIFF
--- a/Modules/Scripted/DICOMLib/Widgets/qSlicerDICOMExportDialog.cxx
+++ b/Modules/Scripted/DICOMLib/Widgets/qSlicerDICOMExportDialog.cxx
@@ -255,7 +255,7 @@ void qSlicerDICOMExportDialog::examineSelectedItem()
     }
 
   // Get exportables from DICOM plugins for selection
-  QMap<QString,QList<qSlicerDICOMExportable*> > exportablesByPlugin;
+  QMultiMap<QString,qSlicerDICOMExportable*> exportablesByPlugin;
   foreach (vtkIdType selectedSeriesItemID, selectedSeriesItemIDs)
     {
     PythonQt::init();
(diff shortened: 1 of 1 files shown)


jcfr:
@Sunderlandkyl Could you consolidate this with commit `COMP: Fix -Wdeprecated-declarations in qSlicerDICOMExportDialog.cxx` associated with https://github.com/Slicer/Slicer/pull/6864 ?

pieper (dismissed):
LGTM 👍

Sunderlandkyl:
@jcfr Sure, I brought over the changes from https://github.com/Slicer/Slicer/pull/6864, although there was a bug causing duplicate entries in the list. I fixed that issue and added back the QList to exportablesByConfidence so that exportables remained grouped by plugin.

To clarify, the previously intended behavior seems to be to create only one entry in the list for each plugin, with the number of exportables displayed as "(# series)". Is this still the intended behavior, or do we want an entry for each exportable?:
image

jcfr:
Thanks for consolidating.

> added back the QList to exportablesByConfidence so that exportables remained grouped by plugin

Makes sense. This was an oversight on my side. The behavior you described is the correct one.

jcfr (approved):

Modules/Scripted/DICOMLib/Widgets/qSlicerDICOMExportDialog.cxx
```diff
@@ -289,24 +289,16 @@ void qSlicerDICOMExportDialog::examineSelectedItem()
...
-        {
-        exportablesByPlugin[plugin].append(exportable);
-        }
+      exportablesByPlugin.insert(plugin, exportable);
     }
   }
   // Map the grouped exportables by confidence values so that the highest confidence is on top
-  QMap<double,QList<qSlicerDICOMExportable*> > exportablesByConfidence;
-  foreach(QList<qSlicerDICOMExportable*> exportablesForPlugin, exportablesByPlugin)
+  QMultiMap<double,QList<qSlicerDICOMExportable*> > exportablesByConfidence;
```
jcfr, line 296: When using `QMultiMap`, the values associated with a given key are managed using a list. For this reason, I think using `QMultiMap<double,qSlicerDICOMExportable*> exportablesByConfidence;` should be enough ?

PR #6865, opened 2023-03-07
Parameter pack invariants + FloatRange

Connor-Bowley:
Adds the ability to enforce invariants on parameterPacks.
Adds tests and documentation for enforcing said invariants.
Adds a `FloatRange` class to be used with `ctkRangeWidget`.


allemangD (approved):
Fairly minimal code changes to support invariants, looks like the real effort would be required by the extension developer. It's good to have this kind of practice codified and encouraged.

I suspect this won't come up for most users so I have no issue with the extra overhead compared with the decorator approach you mentioned before.

Thanks!

PR #6871, opened 2023-03-09
ENH: ParameterNodeWrapper default by assignment

Connor-Bowley:
Update to allow parameterPacks and parameterNodeWrappers to specify default values via assignment instead of the Default annotation. The assignment syntax is simpler and easier to read.

### DIFF
--- a/Base/Python/slicer/parameterNodeWrapper/parameterPack.py
+++ b/Base/Python/slicer/parameterNodeWrapper/parameterPack.py
@@ -176,8 +176,13 @@ def _processParameterPack(classtype):
         membertype, annotations = splitAnnotations(nametype)
 
         serializer, annotations = createSerializer(membertype, annotations)
-        default, annotations = extractDefault(annotations)
-        default = default.value if default is not None else serializer.default()
+        if hasattr(classtype, name):
+            # default via equals
+            default = getattr(classtype, name)
+        else:
+            # default via "Default" class, or default default
+            default, annotations = extractDefault(annotations)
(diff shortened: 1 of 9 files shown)


jcfr:
This is a great improvement :100:

allemangD (approved):
Nice! I am glad the changes to support this were so minimal. 

It is important to note that mutable defaults are still a bad idea here, but that is the same semantics as with `Default(...)` and everywhere else in Python so I don't think it needs any special handling. `dataclasses` handles this via `field(default_factory=...)` but the effort to support something like this would be significant and I suspect it would not be used often for parameter nodes anyway.

Connor-Bowley:
> mutable defaults are still a bad idea here...

Yes, but the `Default` annotation still exists and has a `Default(generator=<callable>)` from in its constructor, which could be used for this.

allemangD:
> Yes, but the `Default` annotation still exists and has a `Default(generator=<callable>)` from in its constructor, which could be used for this.

Ah, now I remember that discussion we already had about this. Thanks for the reminder.

In that case these changes LGTM! :rocket:

jcfr:
> Ah, now I remember that discussion we already had about this. Thanks for the reminder.

Should a comment be added to the code ? (e.g in the docstrings of the Default class?)

Connor-Bowley:
The docstring of constructor for `Default` reads:

>If generator is not None, it will be called each time the default is requested via the self.value
> property.
> Otherwise, value will be returned.

PR #6877, opened 2023-03-10
Import more series information from DICOM images

lassoan:
In addition to series modality and series number, also read these fields from DICOM into metadata in the scene: seriesDescription, seriesDate, seriesTime, contentDate, contentTime, manufacturer, model.

This allows preserving more information when importing and then exporting DICOM images.

Also changed default series number to 301 (from 1) because values <100 are typically used for primary acquisitions, while higher numbers are used for processed data.

### DIFF
--- a/Modules/Scripted/DICOMLib/DICOMPlugin.py
+++ b/Modules/Scripted/DICOMLib/DICOMPlugin.py
@@ -202,6 +202,14 @@ class DICOMPlugin:
         tags['seriesInstanceUID'] = "0020,000E"
         tags['seriesModality'] = "0008,0060"
         tags['seriesNumber'] = "0020,0011"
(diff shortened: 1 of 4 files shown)


pieper:
Are the translation-related changes meant to be part of this PR?

lassoan:
> Are the translation-related changes meant to be part of this PR?

Yes, it was just a small improvement, I didn't want to add a new pull request for it. Please have a look at that, too.

pieper (approved):
LGTM 👍 

You might add a comment about how having the extra dicom tags can be used.

lassoan:
Thanks for the review.

> You might add a comment about how having the extra dicom tags can be used.

These tags are used during DICOM export as default values.

Issue #6878, opened 2023-03-12
Cannot install extensions in latest Slicer Preview Release on Linux


lassoan:
There is a name clash of the settings folder and Slicer executable since changing of the `organization name` from `NA-MIC` to `Slicer` (https://github.com/Slicer/Slicer/pull/6750).

There was no issue on Windows, because the Slicer executable name is Slicer.exe and on macOS the folder structure is special, but on linux there is indeed a name clash. The only way to change this is to modify the Slicer executable or the organization name, because the settings subfolder naming logic logic is hardcoded in the constructor of the QSettings class.

See more information here:
https://discourse.slicer.org/t/install-extensions-is-not-working-on-latest-preview-release/28316/7?u=lassoan

Changing the Slicer executable name is probably not a good idea, as it could cause regressions at many places. Therefore, we need to choose a different organization name.

jcfr:
> Does this need to be specifically prevented?

Good point. Adding an explicit check for this is sensible.

I am thinking about the following:

```diff
diff --git a/CMake/SlicerMacroBuildApplication.cmake b/CMake/SlicerMacroBuildApplication.cmake
index c8428fe796..cb41800f2e 100644
--- a/CMake/SlicerMacroBuildApplication.cmake
+++ b/CMake/SlicerMacroBuildApplication.cmake
@@ -339,6 +339,11 @@ macro(slicerMacroBuildApplication)
   message(STATUS "Configuring ${SLICERAPP_APPLICATION_NAME} application: ${SLICERAPP_NAME}")
   message(STATUS "--------------------------------------------------")
 
+  if(Slicer_ORGANIZATION_NAME STREQUAL SLICERAPP_APPLICATION_NAME)
+    # See https://github.com/Slicer/Slicer/issues/6878
+    message(FATAL_ERROR "Application name must be different from the organization name")
+  endif()
+
   macro(_set_app_property varname)
     set_property(GLOBAL PROPERTY ${SLICERAPP_APPLICATION_NAME}_${varname} ${SLICERAPP_${varname}})
     message(STATUS "Setting ${SLICERAPP_APPLICATION_NAME} ${varname} to '${SLICERAPP_${varname}}'")

```

PR #6879, opened 2023-03-12
Change organization name and domain to slicer.org from Slicer

lassoan:
Commit https://github.com/Slicer/Slicer/commit/b0b9be57cb7d1a1e462143ec74e3d37d271d114d modified the organization name to Slicer from NA-MIC, which caused a name clash with the Slicer executable.

This commit changes the organization name to slicer.org. The final decision on the used organization name has not been made yet, this commit is a quick fix to resolve a blocking issue.

see https://github.com/Slicer/Slicer/issues/6878

We also change the organization domain name from www.slicer.org to slicer.org for consistency.


pieper (approved):
Thanks for the quick fix!  I think slicer.org will work well and I don't see any obvious better choice.

pieper:
I'll let you merge - the commit name could be either BUG or ENH I guess.

lassoan:
Thanks for the review, I'm merging it now.

jcfr:
Thanks for addressing this. Using `slicer.org` consistently is a great idea.

jamesobutler:
From the Control Panel programs and features area on Windows, marking the publisher as "slicer.org" is a unlike any of the other publisher names I have.
image
image

lassoan:
Organization name in the installer (that shows up in your screenshots) could be a different string than the installation folder name, but I don't know if we want to complicate things (add another CMake variable, allow custom applications to override, add to the template, etc).

PR #6864, opened 2023-03-07
Fix warnings

jcfr:

PR #6884, opened 2023-03-14
COMP: Fix missing QSslConfiguration include

RafaelPalomar:
One of the changes introduced in Slicer/Slicer#6864 produces a compilation error (missing QSslConfiguration header) in some GNU/Linux distributions, more specifically on Gentoo, which deploys Qt 5.15.8-r3. In some versions of Qt, this might have been indirecly included in some other Qt headers, thus not producing the error.

This commit solves the issue by including `QSslConfiguration` explicitly. This change should work across operating systems.

### DIFF
--- a/Base/QTCore/qSlicerCoreApplication.cxx
+++ b/Base/QTCore/qSlicerCoreApplication.cxx
@@ -30,6 +30,7 @@
 #include <QNetworkProxyFactory>
 #include <QResource>
 #include <QSettings>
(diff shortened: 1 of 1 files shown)


pieper (approved):
Makes sense to me.  Maybe this will address some other ssl issues too.

@jcfr do you want to weigh in before merging?

jcfr:
Thanks for the fix 🙏 and the descriptive commit message 💯

PR #6895, opened 2023-03-15
ENH: Add option to disable python wrapping on loadable modules

RafaelPalomar:
This makes the python wrapping for loadable module components depend on the `Slicer_USE_PYTHONQT` CMake variable. This avoids python wrapping when `Slicer_USE_PYTHONQT=OFF`, which can save generation of files that won't be used.

NOTE: For discussion purposes, so far, this PR applies the concept only to the MRML components of the Colors module, but the idea would be to apply this across modules and components. Alternatively, the `Slicer_USE_PYTHONQT` variable could be checked on the CMake SlicerBuild macros.

@pieper, @jcfr would these changes make sense to you?

### DIFF
--- a/Modules/Loadable/Colors/MRML/CMakeLists.txt
+++ b/Modules/Loadable/Colors/MRML/CMakeLists.txt
@@ -13,6 +13,11 @@ set(${KIT}_TARGET_LIBRARIES
   ${MRML_LIBRARIES}
   )
 
+set(DISABLE_WRAP_PYTHON)
+if(NOT Slicer_USE_PYTHONQT)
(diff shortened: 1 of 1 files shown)


jcfr:
Generally speaking, fixing this makes sense. 

Instead of adding  code in each module (and extension), I suggest to instead update the Slicer CMake macros to skip wrapping if the option `Slicer_USE_PYTHONQT` is set to OFF.

RafaelPalomar:
> Instead of adding code in each module (and extension), I suggest to instead update the Slicer CMake macros to skip wrapping if the option `Slicer_USE_PYTHONQT` is set to OFF.

That makes sense. I can work on it.

PR #6897, opened 2023-03-16
BUG: Fix warning when changing window/level with mouse mode

jamesobutler:
This fixes an issue that would always log the following warning when changing Window/Level with the mouse mode. This issue was introduced in my commit https://github.com/Slicer/Slicer/commit/2896452063e2bcd5526934cd032f68603fd572ff where I removed the Window/Level locked state for a given volume. This logic was there to previously check if the volume in the slice viewer had an unlocked Window/Level state.
```
[VTK] Warning: In D:\D\P\S-0\Libs\MRML\Logic\vtkMRMLSliceLogic.cxx, line 2457
[VTK] vtkMRMLSliceLogic (00000246B7D8E270): vtkMRMLSliceLogic::VolumeWindowLevelEditable method is deprecated. Volume window level can always be set programmatically.
```


lassoan (approved):

PR #6898, opened 2023-03-17
COMP: Pin more GitHub actions to full length commit SHA

jamesobutler:
This is a followup to https://github.com/Slicer/Slicer/commit/73f549c1e698a087c550c193e849e7a03645113f.

GitHub's security hardening guide recommends this mitigation method. https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

### DIFF
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -8,7 +8,7 @@ jobs:
     name: Check for spelling errors
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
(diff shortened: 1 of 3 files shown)


jcfr:
Thanks 🙏

Out of curiosity, did you manually updated the hash and added comment or is there some tooling to help ?

Having to do something similar, I ended up manually looking up sha and revision.

jamesobutler:
I did it manually. I briefly looked for a tool to make the replacement, but I haven’t found one.

jcfr:
Turns out that looking in the GitHub Actions log can help with this. See SHA reported along with each action used in the corresponding workflow.

image

PR #6277, opened 2022-03-20
Use more inclusive language

jamesobutler:
This PR follows recommendations from https://tools.ietf.org/id/draft-knodel-terminology-09.html which have been adopted by many others in the community for work towards the greater good. Although it is often nice to maintain using current terminology, as it relates to improving inclusivity, this should be higher priority.

This PR includes acting upon what was originally described by @jcfr at https://discourse.slicer.org/t/2021-04-06-hangout/16953/3?u=jamesobutler which detailed:

> Removal of divisive terminology regarding branch name associated with repository hosted in the Slicer and commontk GitHub organization. We will rename `master` to `main`

Renaming of the `master` branch to `main` had been indicated in April 2021, but has not yet been acted upon, until now...

GitHub has improved the ability to rename a branch (https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-branches-in-your-repository/renaming-a-branch) which makes it easy to switch to a new name with minimal consequences. This has to be completed by the admin of the Slicer repository (@jcfr).
Some action items:
- [x] Rename default branch for https://github.com/Slicer/Slicer/tree/master
- [x] Rename default branch for https://github.com/Slicer/ExtensionsIndex/tree/master, re https://github.com/Slicer/ExtensionsIndex/pull/1812
- [x] Rename default branch for https://github.com/Slicer/apidocs.slicer.org/tree/master, re https://github.com/Slicer/apidocs.slicer.org/pull/3

Renaming of the default branch looks something like this:
image

PR #6901, opened 2023-03-20
ENH: Use inclusive language for segmentation source representation

adamaji:
See https://tools.ietf.org/id/draft-knodel-terminology-09.html#name-master-slave

This PR follows earlier work for using more inclusive language and removing usage of "master" in the code (https://github.com/Slicer/Slicer/pull/6277). It seems like this instance made it through?

Also, it's user facing:
image

Before I continue here:
- Is "main" an acceptable replacement in this context?
- Should I be following the deprecation of the earlier API changes as opposed to replacement? (Given that Slicer 5.0.x extensions are no longer being built?)


jamesobutler (commented):
> Should I be following the deprecation of the earlier API changes as opposed to replacement? (Given that Slicer 5.0.x extensions are no longer being built?)

Yes you should mark them as deprecated here in Slicer like the others so that extensions have time to update and so they don't immediately start to fail when using the Slicer Preview. Once the code makes it into Slicer stable, then extensions could be updated to use the new value entirely as the latest versions of Slicer extensions often are only kept compatible with the latest Slicer stable and preview. For example https://github.com/lassoan/SlicerSegmentEditorExtraEffects/pull/53 was integrated because `setSourceVolume` exists in both the latest Slicer stable and preview.

cpinter:
I have understood that now we call it source representation, at least
previous changes pointed to that direction. It is more descriptive of its
actual purpose than "main" as well I think.

jamesobutler (approved):
Thanks for your contribution. I have continued your work to update the terminology from `main` to `source` here and added the various deprecation warnings for changes to the public API.

I have confirmed things build successfully on Windows and that tests are still passing.

jcfr:
To follow up on the integration, since we check for new warnings, the test ` py_nomainwindow_test_slicer_util_save` started to fail.

```
Warning: In /work/Preview/Slicer-0/Libs/MRML/Core/vtkMRMLSegmentationNode.h, line 222
vtkMRMLSegmentationNode (0x45032a0): vtkSegmentation::SetMasterRepresentationToBinaryLabelmap() method is deprecated, please use SetSourceRepresentationToBinaryLabelmap method instead
```

See https://slicer.cdash.org/test/24940170

The following pull request is updating the test:
* https://github.com/Slicer/Slicer/pull/7022

PR #6904, opened 2023-03-21
COMP: Ensure Qt language tools are bundled on macOS

jcfr:
This commit is a follow-up of 0a00473ea (ENH: Bundle Qt language tools).

It factors out the logic for installing the tools into a CMake module included on all platforms.

Notes:
* The module `SlicerBlockInstallQtTools.cmake` was designed to eventually support installing other Qt tools.
* Since the bundling of Qt designer is implemented in two files `SlicerCPack.cmake` and `SlicerCPackBundleFixup.cmake.in` included at either build or packaging time, the refactoring the associated installation code is a more involved task that does not belong to this pull request.


lassoan (approved):

PR #6906, opened 2023-03-22
BUG : Exit cross of Pop-up Save before exit is not functional

cpoetteCertis:
The close button (red cross) is enable only if at least one button from the popup is set to RejectRole. See the following link for more details:

https://forum.qt.io/topic/96001/how-enable-and-add-action-to-closebutton-qmessagebox/4

The cancel button is set to RejectRole because it is set to unused at the following line so it does not change the popup behavior and it enable the close button.


pieper (dismissed):
Yes, this makes sense.  👍

lassoan:
Thank you for your contribution.

> The cancel button is set to RejectRole because it is set to unused at the following line so it does not change the popup behavior and it enable the close button.

The popup behavior is actually changed (it does not matter if the returned `cancelButton` variable is used or not), but in a good way.

Commit message check failed because the message must start with `BUG: ...` - no space before the `:`. Please fix the commit message and force-push the updated commit to this same branch. Thank you!

cpoetteCertis:
Hello! 

I noticed that the build failed and it does not seem to be related to the bug correction I suggested.

What is the procedure to unlock the situation?

Thanks in advance!

Christopher

lassoan:
Correct, the build failure was unrelated. I'll force merge now.

In the future, in such cases you can rebase your branch on the latest main branch and force-push here to re-trigger a build.

cpoetteCertis:
Thank you for the tip!

PR #6910, opened 2023-03-23
ENH: Add 3D cursor as crosshair options with fast picking

laurennlam:
Add new 3D Cursor option in crosshair toolbar.
This allows to continuously pick volume with fast picking.
The basic Qt cursor is replaced by a crosshair.

[Discourse Slicer reference for more information](https://discourse.slicer.org/t/3d-mouse-cursor-for-stereoscopic-displays/28370)

https://user-images.githubusercontent.com/11785751/227495055-829a03d7-b5ab-4d8a-890d-2c53c982fb1f.mp4

@LucasGandel

### DIFF
--- a/Base/QTGUI/qSlicerViewersToolBar.cxx
+++ b/Base/QTGUI/qSlicerViewersToolBar.cxx
@@ -25,7 +25,10 @@
 // Slicer includes
 #include "qSlicerApplication.h"
 #include "qSlicerLayoutManager.h"
+#include "qMRMLLayoutManager.h"
 #include "qMRMLSliceWidget.h"
+#include "qMRMLThreeDView.h"
+#include "qMRMLThreeDWidget.h"
 #include "qSlicerViewersToolBar_p.h"
 
 // SlicerLogic includes
@@ -169,13 +172,28 @@ void qSlicerViewersToolBarPrivate::init()
   QObject::connect(this->CrosshairThicknessMapper, SIGNAL(mapped(int)),
                    this, SLOT(setCrosshairThickness(int)));
 
+  // Crosshair options
+  QActionGroup *crosshairOptionsActions = new QActionGroup(q);
+  crosshairOptionsActions->setExclusive(false);
+
+  this->TrackCursorAction = new QAction(q);
+  this->TrackCursorAction->setText(tr("Track cursor"));
+  this->TrackCursorAction->setToolTip(tr("The crosshair follows the mouse cursor."));
+  this->TrackCursorAction->setCheckable(true);
+  QObject::connect(this->TrackCursorAction, SIGNAL(toggled(bool)), this,
+                   SLOT(setCrosshairTrackCursor(bool)));
+  crosshairOptionsActions->addAction(this->TrackCursorAction);
+
   // Crosshair Menu
   this->CrosshairMenu = new QMenu(tr("Crosshair"), q);
+  this->CrosshairMenu->setToolTipsVisible(true);
   this->CrosshairMenu->addActions(crosshairJumpSlicesActions->actions());
   this->CrosshairMenu->addSeparator();
   this->CrosshairMenu->addActions(crosshairActions->actions());
   this->CrosshairMenu->addSeparator();
   this->CrosshairMenu->addActions(crosshairThicknessActions->actions());
+  this->CrosshairMenu->addSeparator();
+  this->CrosshairMenu->addActions(crosshairOptionsActions->actions());
 
   // Crosshair ToolButton
   this->CrosshairToolButton = new QToolButton();
@@ -407,6 +425,31 @@ void qSlicerViewersToolBarPrivate::setCrosshairMode(int mode)
 }
 
 //---------------------------------------------------------------------------
+void qSlicerViewersToolBarPrivate::setCrosshairTrackCursor(bool track) {
+  //  Q_Q(qSlicerViewersToolBar);
+
+  vtkSmartPointer<vtkCollection> nodes;
+  nodes.TakeReference(this->MRMLScene->GetNodesByClass("vtkMRMLCrosshairNode"));
+  if (!nodes.GetPointer())
+    {
+    return;
+    }
+  vtkMRMLCrosshairNode *node = nullptr;
+  vtkCollectionSimpleIterator it;
+  for (nodes->InitTraversal(it); (node = static_cast<vtkMRMLCrosshairNode *>(
+                                      nodes->GetNextItemAsObject(it)));)
+    {
+
+    // The mouse cursor is hidden when track cursor is on, so make sure that the crosshair is visible
+    if (track && node->GetCrosshairMode() == vtkMRMLCrosshairNode::NoCrosshair)
+      {
+      node->SetCrosshairMode(vtkMRMLCrosshairNode::ShowBasic);
+      }
+    node->SetTrackCursor(track);
+    }
+}
+
+//---------------------------------------------------------------------------
 void qSlicerViewersToolBarPrivate::setCrosshairThickness(int thickness)
 {
   vtkSmartPointer<vtkCollection> nodes;
@@ -544,6 +587,7 @@ void qSlicerViewersToolBarPrivate::setMRMLScene(vtkMRMLScene* newScene)
 void qSlicerViewersToolBarPrivate::updateWidgetFromMRML()
 {
   Q_ASSERT(this->MRMLScene);
+  Q_Q(qSlicerViewersToolBar);
 
   vtkMRMLNode *node;
   vtkCollectionSimpleIterator it;
@@ -629,6 +673,8 @@ void qSlicerViewersToolBarPrivate::updateWidgetFromMRML()
     this->IntersectingSlicesRotationEnabledAction->setChecked(
       this->MRMLAppLogic->GetIntersectingSlicesEnabled(vtkMRMLApplicationLogic::IntersectingSlicesRotation));
     }
+
+  q->change3DViewsCursorTo(crosshairNode->GetTrackCursor() ? Qt::BlankCursor : Qt::ArrowCursor);
 }
 
 //---------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


Libs/MRML/DisplayableManager/vtkMRMLCameraWidget.cxx
```diff
@@ -197,6 +198,29 @@ bool vtkMRMLCameraWidget::CanProcessInteractionEvent(vtkMRMLInteractionEventData
     return this->CanProcessButtonClickEvent(eventData, distance2);
     }
 
+  // Need to handle this event here because we keep handle generic event (rotation, translation,...)
```
jcfr, line 201: Would it be possible to revisit the wording associated with this comment ?

Base/QTGUI/qSlicerViewersToolBar.cxx
```diff
@@ -712,3 +726,32 @@ void qSlicerViewersToolBar::setMRMLScene(vtkMRMLScene* newScene)
...
+
+  // Updated all mapped 3D viewers
+  for (int i = 0; i < layoutManager->threeDViewCount(); ++i)
+    {
+    qMRMLThreeDView *threeDView = layoutManager->threeDWidget(i)->threeDView();
+
+    // The cursor should be updated in all views, not only ones that are in the
+    // current layout. If it is only updated for views that are mapped in the
+    // current layout, then the cursor will be incorrect if we switch to a
+    // layout with different views.
```
jcfr, line 753:
```suggestion
    // The cursor is updated in all views, not only the ones that are in the
    // current layout.
    // This is required to ensure the cursor is also set when switching to a
    // layout with views different from the ones already mapped in the current layout.
```

Libs/MRML/DisplayableManager/vtkMRMLCameraWidget.cxx
```diff
@@ -197,6 +198,29 @@ bool vtkMRMLCameraWidget::CanProcessInteractionEvent(vtkMRMLInteractionEventData
...
+    if (!crosshairNode || crosshairNode->GetCrosshairMode() != vtkMRMLCrosshairNode::Show3DCursor)
+      {
+      return true;
+      }
+
+    crosshairNode->FastPick3DOn();
+    this->ProcessSetCrosshair(eventData);
+    crosshairNode->FastPick3DOff();
+
+    distance2 = 1e10; // we can process this event but we let more specific widgets to claim it (if they are closer)
```
jcfr, line 220:
```suggestion
    // After processing the event (see ProcessSetCrosshair), we artificially set the
    // distance to a large value, this may allow other displayable managers to process
    // the event using other widgets.
    distance2 = 1e10;
```

jcfr (commented):

Waiting @lassoan provides his input, I added few nitpicks related to comments.

Additionally, would it be possible to update the associated discourse post to include:
* a link to the pull-request
* screenshot(s) or video illustrating the feature. This will then be useful when we write release note for the next version.

pieper:
> Additionally, would it be possible to update the associated discourse post to include:
> 
> * a link to the pull-request
> * screenshot(s) or video illustrating the feature. This will then be useful when we write release note for the next version.

👍  to this idea

lassoan (commented):
Adding an option to track the cursor with the crosshair (equivalent with some kind of Shift-lock) would be useful and maybe this pull request is a step towards that direction, However, the current version of the code is just a partial implementation, so 3D and slice views now work inconsistently and the crosshair is broken in slice views if the new mode is enabled.

Could you explain the driving use case? What are your requirements?

Would making this a flag that can be used in combination with all the existing modes and options fulfill your requirements?

Libs/MRML/DisplayableManager/vtkMRMLCameraWidget.cxx
laurennlam: We refactor the code so it's not needed anymore. Thanks for the feedback !

laurennlam:
> Adding an option to track the cursor with the crosshair (equivalent with some kind of Shift-lock) would be useful and maybe this pull request is a step towards that direction, However, the current version of the code is just a partial implementation, so 3D and slice views now work inconsistently and the crosshair is broken in slice views if the new mode is enabled.
> 
> Could you explain the driving use case? What are your requirements?
> 
> Would making this a flag that can be used in combination with all the existing modes and options fulfill your requirements?

Thanks a lot @lassoan for the feedbacks. Indeed, the implementation wasn't the best one. This is not a new crosshair mode but most an option to the crosshair. Thanks to your suggestions, we were able to make a more clearer/cleaner code. We hope it's the correct way now. Following your instructions (and also from @jcfr, thank you ! ), here are the modifications we made:
- Remove the added enum for crosshair mode
- Create a new section on the crosshair with this options which can now be toggled
- Enable tooltips
- Reset picking
- Adding a new video to explains the functionality
- Tagged this MR into the discourse post

We are waiting for your feedbacks if needed. Again, thanks for your time.

lassoan:
Thank you I'll test the updated version today.

lassoan:
Thank you, this is a step in the right direction. Two issues:
- If I move the mouse in the 3D view then I don't see any updates in the Data Probe. Maybe the cursor position is not updated?
- It does not work in slice views. If I move the mouse in the slice view, the crosshair does not move without holding down the Shift key.

Can you write a bit about your use case? I find it a bit limiting that the crosshair always end up in irrelevant position because there is no way to decouple the crosshair from the mouse pointer position. Maybe what the checkbox should control is enabling fast picking and continuous cursor update in 3D views? (so that you can see information about the current mouse position in the Data probe without holding down the Shift key)

LucasGandel:
Thanks @lassoan for taking the time to test.

>  If I move the mouse in the 3D view then I don't see any updates in the Data Probe. Maybe the cursor position is not updated?

Could you explain what should be expected in the Data Probe? Even without using this new functionality, I don't see anything in the Data Probe when moving the mouse in the 3D view, nor when moving the crosshair. 

> It does not work in slice views. If I move the mouse in the slice view, the crosshair does not move without holding down the Shift key.

This was intended, see the explanation about the use case below. This could be implemented if needed. 

> Can you write a bit about your use case?

The goal is to implement a stereoscopic mouse cursor for the 3D view only.
In stereo mode, using the usual "screen-space" mouse pointer to pick/place something in 3D is not optimal. A preferred representation is to have the mouse cursor projected in the 3D scene and rendered as a 3D object.

> Maybe what the checkbox should control is enabling fast picking and continuous cursor update in 3D views?

Do you mean "crosshair" instead of "cursor"? Because that would already be the case. In the current implementation, the checkbox controls fast picking and continuous crosshair update in 3D views only. In addition, the usual mouse cursor is hidden to fulfil the requirement of having a "3D" cursor, which is represented by the crosshair.

> I find it a bit limiting that the crosshair always end up in irrelevant position because there is no way to decouple the crosshair from the mouse pointer position

I see what you mean. Thank you @pieper for your insight too. A context menu sounds good to me.<br/>
We could also hide the mouse cursor only if some stereo mode is activated, but that would not help with the crosshair moving away when you move the mouse out of the 3D view.

We are open to any modification as long as we have something behaving like the stereoscopic mouse cursor described above.<br/>
Do not hesitate if the requirement is still not clear.

pieper:
Is the target device the Looking Glass?

LucasGandel:
> Is the target device the Looking Glass?

The target device is rather the [3D Pluraview](https://www.3d-pluraview.com/en/), but it could also be the Looking Glass. It's really for any stereo device, where using the usual screen space mouse pointer "kills" the 3D feeling.
It could also be used with a 3D mouse.

LucasGandel:
Trying to move forward with this, the main limitation preventing this feature to be merged seems to be the following:

> the crosshair always end up in irrelevant position because there is no way to decouple the crosshair from the mouse pointer position

What about automatically toggling the TrackCursor mode when the mouse cursor enter/leaves the 3D view?<br/>
After choosing the "Track Cursor" behavior in the crosshair menu, nothing would change in the 2D/3D views display.<br/>
Then, as soon as the mouse cursor enters the 3D view, fast picking is enable and the crosshair is updated continuously to follow the hidden mouse cursor.<br/>
When the mouse cursor leaves the 3D view, the crosshair is disabled and set back to its default position.<br/>

Please let me know what you think.

pieper:
> What about automatically toggling the TrackCursor mode when the mouse cursor enter/leaves the 3D view?

The concern about this is that when you leave the 3D view your mouse position is on the edge of the view, which isn't a useful place for the crosshairs to be in general.  That's why the shift-move feels natural to me - I can adjust the crosshair when I want to, but also leave it fixed otherwise.

I'm thinking what you want is a kind of projected shadow of the mouse location that only shows when the mouse enters/leaves the 3D view.  Like a "preview" of where the crosshair will go if you hold shift.  The 2D mouse cursor could be turned off so that it's not distracting.

lassoan:
I agree with @pieper, it seems that you would like a virtual pointer.

Visualization of this pointer and/or the trajectory and the found intersection point could be implemented outside the cursor/crosshair infrastructure using a new displayable manager. The displayable manager would be notified about every mouse events, so you could capture the 2D coordinates of the pointer, update the trajectory line (for example, a vanishing thin line, as it is common in AR/VR applications) and maybe show a dot at the pick position (where that trajectory line hits 3D content).

For any meaningful interaction, you probably need some kind of trigger/button on the input device. You should be able to handle the interaction events that the device generates in the displayable manager.

Maybe you could demo the device at the next developer meeting on Tuesday and we could then better understand how it could all work.

LucasGandel:
Thank you very for your insight, I agree with both of your comments.
What we need is indeed a virtual cursor that acts like a preview of where the crosshair would go.
I am now convinced that having a new displayable manager is the way to go. We will leave the crosshair functionality as it is. The crosshair can then be placed with the virtual cursor while pressing the shift key.

If you confirm this is an acceptable feature to be integrated in the 3DSlicer release, then we will proceed with the new implementation. We will probably start with a simple representation (without the trajectory line) similar to the one available in VTK ([screenshot in MR description](https://gitlab.kitware.com/vtk/vtk/-/merge_requests/9894)).

Unfortunately it won't be possible to stream the content of the stereo view so that you feel the stereo effect, you would just see interlaced images. But I'll check if I can record something with my phone.
Anyway I'm happy to discuss potential solutions if you think this is not an acceptable feature. I plan to join the meeting about the SlicerVESPA extension later today, we could keep 5min at the end to talk about this PR.

Please let me know what you think.

lassoan:
Sounds good to me. I don't have the VESPA meeting in my calendar, please send me an invitation. Thank you!

LucasGandel:
> Sounds good to me. I don't have the VESPA meeting in my calendar, please send me an invitation. Thank you!

Thanks. There is a link in this [SlicerVESPA comment](https://github.com/Slicer/ExtensionsIndex/pull/1925#issuecomment-1489861399), but I forwarded the link to you and Steve directly.

pieper:
I'm doublebooked so I'll join if I can, but please go ahead without me.  I think Andras and I are on the same page.

LucasGandel:
As discussed offline, a better approach that is decoupled from the crosshair should be preferred.
Closing this in favor of #6937.

PR #6913, opened 2023-03-26
DOC: Describe how pip_install can be used in a module

lassoan:


jcfr (commented):
I suggest to expand the "important" section indicating that this approach works well for fast prototyping and early development.

When application aims to be deployed in production, all libraries should be packaged to avoid relying an internet connection.

To support that later use case, the package and its dependencies should be downloaded at build time using an approach based on external projects.

lassoan:
> When application aims to be deployed in production, all libraries should be packaged to avoid relying an internet connection.

In case of small, isolated packages, bundling in the extension may work and would have several advantages. However, it would not work for many packages, because they are big (including dependencies we often see hundreds of megabytes, or in case of pytorch etc. several gigabytes), and they may also change quite frequently. I'm also not sure if pip would be able to resolve version incompatibilities.

So, I would say `pip_install` is a good solution, not just for testing. We should think a bit more about when exactly we recommend bundling a Python package.

jamesobutler:
> However, it would not work for many packages, because they are big (including dependencies we often see hundreds of megabytes, or in case of pytorch etc. several gigabytes)

Regarding the above point, to the user it doesn’t seem to be an issue whether they are waiting X number of minutes to download the extension (with the included python packages) from the extensions manager versus X number of minutes downloading from the web (pip install) at runtime. 

Is there a restriction of the size of an extension hosted on the extensions manager server? Or some other reason why the size of the python packages would be a concern here?

lassoan:
When a module manages download of a Python package then the module can install it as late as possible, only when the user actually wants to use a feature that requires it, the module can ask for user confirmation, warn the user that it will take a long time, etc. In contrast, when you download a large extension package then the extension manager simply stops responding.

Pyhthon packages can also be very big. Each day you would store it in the extension manager. If multiple extensions need the same package and they all bundle it then you would store the package each night multiple times. there is no such issue if Python packages are installed at runtime.

I'm also not sure about lifetime management of bundled packages. What happens if extension A bundles a Python package and extension B bundles a different version of that package? Would you expect that the package is deleted from the extension A? Is it going to break extension A? Is extension A going to break if then extension B is uninstalled? If extensions install their required Python packages at runtime then the extension can simply check if the required version of an extension is available and if not then it installs it (the user just needs to wait when switching between modules that rely on different versions of the same package).

So, overall, I find runtime installation of Python packages much more robust and user-friendly than bundling packages and I think it should be the recommended method (except special cases and custom applications).

jamesobutler:
The build time install and packaging of python packages has followed the Slicer convention of specifying specific versions of packages to use (see [here](https://github.com/Slicer/Slicer/blob/5d069c725aaee7a1e56b0e2832e137b008a8a89c/SuperBuild/External_python-dicom-requirements.cmake#L43-L79)). They are currently stored as part of each Slicer preview release as well. Though if a Slicer core module uses a python package, should that as well only be installed at runtime of that specific core module to keep the package size down?

The Slicer script repository for [installing a python package](https://slicer.readthedocs.io/en/latest/developer_guide/script_repository.html#install-a-python-package) makes no mention about being careful about specific versions. That is fine for R&D stuff, but when using a module should python versioning be more explicit?

lassoan:
> if a Slicer core module uses a python package, should that as well only be installed at runtime of that specific core module to keep the package size down?

Python packages bundled with Slicer are mostly very small, so we would not gain a lot by retrieving them runtime, but Slicer would lose basic features (such as DICOM loading) if a network connection was unavailable. However, if we started to use PyTorch in Slicer core modules then I would probably vote for installing it at runtime instead of bundling it with the installation package (or maybe have a light and an offline version).

> makes no mention about being careful about specific versions

In general, I would encourage developers to use whatever version is installed and install the default (latest) version of a package if it has not been installed yet. We could add a note that in case this is impossible then developers need to check the package version and install an appropriate, but usually these are quite complex problems (see PyTorch, SimpleITK, numpy, etc.), so a simple general advice might not help a lot.

Docs/developer_guide/python_faq.md
```diff
@@ -246,3 +246,9 @@ In our example:
       ${MODULE_NAME}Lib/cool_maths.py
       ${MODULE_NAME}Lib/utils.py
       )
+
+## Can I use any Python package in a Slicer module
+
+Any Python packages can be installed in Slicer's built-in Python environment. A convenience function `slicer.util.pip_install` can be used as shown in the [example in the Script Repository](script_repository.md#install-a-python-package).
+
+Important: Do not install any packages in the global scope (outside of all classes and functions) or in the module class constructor, as it can very significantly slow down the application startup and it may even preven the module from loading.
```
lassoan, line 254:
```suggestion
Important: Do not install any packages in the global scope (outside of all classes and functions) or in the module class constructor, as it can very significantly slow down the application startup and it may even prevent the module from loading.
```

lassoan:
@jcfr @jamesobutler it would be nice to converge to a common agreement so that we can close this PR (we have too many open).

jamesobutler:
I would suggest whatever is recommended to developers is also being followed by Slicer core. If there is a call to be made about when to package python packages with the code versus installing at runtime, then those guidelines should be published to developers. 

I agree that PyTorch is a large package that may be best for install at runtime, though I also see the value of an offline package as well when users are unable to access the internet or is restricted heavily by some corporate IT policy. Instructions to help regular users (non-developers) do this would be appreciated.

@jcfr will have to provide guidance about how much storage space Kitware will allow an individual extension to take up on their server space. For example, the SlicerRT extension on macOS has a package size of 100 MB (see [here](https://slicer-packages.kitware.com/#item/64ad1a97b3c3a78356bfeb15)). Maybe the extensions index documented info should be like PyPI’s standard of the default package size limit being 60 MB (exceptions are made for larger packages)? As a general practice, modules should download from the web the content they need over the limit. They could also host an offline extension zip in their GitHub repository which would also be documented to developers.

Docs/developer_guide/python_faq.md
```diff
@@ -246,3 +246,9 @@ In our example:
       ${MODULE_NAME}Lib/cool_maths.py
       ${MODULE_NAME}Lib/utils.py
       )
+
+## Can I use any Python package in a Slicer module
+
+Any Python packages can be installed in Slicer's built-in Python environment. A convenience function `slicer.util.pip_install` can be used as shown in the [example in the Script Repository](script_repository.md#install-a-python-package).
+
+Important: Do not install any packages in the global scope (outside of all classes and functions) or in the module class constructor, as it can very significantly slow down the application startup and it may even prevent the module from loading.
```
pieper, line 254:
```suggestion
Important: Do not install any packages in the global scope (outside of all classes and functions) or in the module class constructor, as it can very significantly slow down the application startup and it may even prevent the module from loading.

In addition, since installing packages can have side effects to other extensions or to the main application, always include a confirmation dialog indicating that the installation will happen.
```

lassoan:
> I would suggest whatever is recommended to developers is also being followed by Slicer core.

My recommendations would be different for Slicer core (and custom application) developers vs. extension developers. Generally, Python packages are better to be bundled in Slicer core (it would risky to require internet connection for very basic features, such as data loading) while packages are better to be downloaded at runtime for extensions (to keep the extension packages smaller; to only install Python packages that are actually needed, as few and as late as possible; I'm also not sure how version conflict resolution would work without using PyPI, etc).

Of course, there can be exceptions, for example I can imagine that at some point SimpleITK would not be bundled with Slicer anymore; and there are a few extensions that bundle small but tricky Python packages.

@jcfr Could you clarify what you meant by "When application aims to be deployed in production, all libraries should be packaged to avoid relying an internet connection."? Do you mean this for custom application development (to create a standalone package), or you would prefer if all extension developers would bundle all Python packages in their extensions? What happens if different extensions bundle different versions of the same Python package? Which one will be used? If you ininstall a package (e.g., pip installs a different version of the package) then are any of the bundled packages will be uninstalled?

jcfr:
> Could you clarify what you meant by "When application aims to be deployed in production, all libraries should be packaged to avoid relying an internet connection."? Do you mean this for custom application development (to create a standalone package), or you would prefer if all extension developers would bundle all Python packages in their extensions? 

In the context of custom application deployed in enterprise settings, bundling the package is required to create a self contained package.

For extensions bundled in custom application, we are creating an external project responsible to install the python packages directly into the Slicer python site-packages. This makes sense as it is a tightly controlled environment (e.g extensions manager is disabled)

> What happens if different extensions bundle different versions of the same Python package? Which one will be used? If you uninstall a package (e.g., pip installs a different version of the package) then are any of the bundled packages will be uninstalled?

For extension packages made available through the extensions manager, we are currently letting extensions install packages and relying on the testing (done as part of the continuous integration on the dashboard) to report issues.

That said, since the set of packages installed in the Slicer Python environment is not deterministic as it depends on the order in which extensions are built & tested, some issue are probably missed or introduced.

For example, considering the following:
* extension A install `python-foo<=2.0`
* extension B install `python-foo`

Extension B may expect to have the latest version of `python-foo` always installed, but if extension A was installed first, that assumption if broken. On the flip side, if extension B was installed first, then extension A would fail to install the package.

For the user, runtime install of package means that the order in which extensions are installed may lead to different set of python packages and may even "break" a given install of Slicer.
* Pros: easy to implement (one call to `slicer.util.pip_install`)
* Cons: may break the Slicer Python environment in unexpected way

Regarding the bundling of python packages into the an extension, each extension provides its own "site-packages" directory appended to the python path (See below):
* Pros: Slicer Python environment is not modifier, disabling an extension exclude associated package from the python paths. Once an extension is installed, there is no network required.
* Cons: require more work to add an external project


### Path forward

It may be sensible to "document" extension python requirements. Based on the use case (runtime install or extension bundling), this would allow to either call `slicer.util.pip_install` at runtime or generate the external project.

In the extension source tree, this could be done by including a `pyproject.toml`[^1] (or a `requirements.txt`) with the `dependencies` metadata.

Possible infrastructure updates:

* Update `slicer.util.pip_install` to support specifying the `--constraint` parameter[^2].

* Include `pyproject.toml` (or `requirements.txt`) in extension package, then at install time provide multiple `--constraint` argument so that dependency resolution accounting for all enabled extension is done while installing package in the given extension.
  * Add a call like `slicer.util.collect_extension_constraint_file()`

* Add [tomli](https://pypi.org/project/tomli/) to the Slicer python package and provide helper function to read the `dependencies` metadata

* Update Slicer build-system to provide output of `pip freeze` (basically a requirement file that could be used as a constraint)

* Update extension build-system bundling python package so that they also include the output of `pip freeze`

### Immediate next steps

* Update extension to provide a `requirements.txt`
* Update `ScriptedLoadableModule` to provide a property like `python_requirements` corresponding to a dictionary like `{"default": ["itk>5", ""]}`
* Document how to install python dependencies in extensions.

There are much more details to sort out but I think these steps will help setup the foundation.


[^1]: Also worth noting that we are already using this file to organize linting rules for custom application. File is usually `Modules/Scripted/pyproject.toml`
[^2]: https://pip.pypa.io/en/latest/user_guide/#constraints-files



### Example of extension bundling python packages

Directories structure of the installed python packages for SlicerRadiomics extension.



See https://github.com/AIM-Harvard/SlicerRadiomics/blob/master/SuperBuild/External_python-pyradiomics.cmake

jcfr:
This has been finalized.

| https://slicer--6913.org.readthedocs.build/en/6913/developer_guide/python_faq.html#can-i-use-any-python-package-in-a-slicer-module |
|--|
| image |

Docs/developer_guide/python_faq.md
```diff
@@ -348,8 +348,21 @@ In our example:
 
 ## Can I use any Python package in a Slicer module
 
-Any Python packages can be installed in Slicer's built-in Python environment. A convenience function `slicer.util.pip_install` can be used as shown in the [example in the Script Repository](script_repository.md#install-a-python-package).
```
jcfr: @jamesobutler  I just realized I forgot to list you as a co-author on this commit :disappointed:

jamesobutler: No worries!

PR #6916, opened 2023-03-28
View manipulation during markup point placement

lassoan:
Small improvements to allow view manipulation during markup point placement.


lassoan:
Based on feedback from the [forum](https://discourse.slicer.org/t/markup-shortkeys-and-model-interaction/28616/5), this seems to work well.

PR #6917, opened 2023-03-29
ENH: Add translation function for python scripted modules

mhdiop:
see Slicer/SlicerLanguagePacks#8

### DIFF
--- a/Modules/Scripted/DICOM/DICOM.py
+++ b/Modules/Scripted/DICOM/DICOM.py
@@ -12,6 +12,7 @@ from slicer.ScriptedLoadableModule import *
 import DICOMLib
 from slicer.i18n import translate
 
+
 #
 # DICOM
 #
@@ -26,15 +27,12 @@ class DICOM(ScriptedLoadableModule):
         ScriptedLoadableModule.__init__(self, parent)
 
         self.parent.title = translate("DICOM", "DICOM")
-        self.parent.categories = ["", _("Informatics")]  # top level module
+        self.parent.categories = ["", translate("DICOM", "Informatics")]  # top level module
         self.parent.contributors = ["Steve Pieper (Isomics)", "Andras Lasso (PerkLab)"]
-        self.parent.helpText = translate("DICOM", """
-This module allows importing, loading, and exporting DICOM files, and sending receiving data using DICOM networking.
-""")
+        self.parent.helpText = translate("DICOM", "This module allows importing, loading, and exporting DICOM files,"
+                                                  " and sending receiving data using DICOM networking.")
         self.parent.helpText += self.getDefaultModuleDocumentationLink()
-        self.parent.acknowledgementText = translate("DICOM", """
-This work is supported by NA-MIC, NAC, BIRN, NCIGT, and the Slicer Community.
-""")
+        self.parent.acknowledgementText = translate("DICOM", """This work is supported by NA-MIC, NAC, BIRN, NCIGT, and the Slicer Community.""")
         self.parent.icon = qt.QIcon(':Icons/Medium/SlicerLoadDICOM.png')
         self.parent.dependencies = ["SubjectHierarchy"]
 
@@ -362,7 +360,8 @@ class _ui_DICOMSettingsPanel:
         # mechanism.
 
         loadReferencesComboBox = ctk.ctkComboBox()
-        loadReferencesComboBox.toolTip = translate("DICOM", "Determines whether referenced DICOM series are " \
+        loadReferencesComboBox.toolTip = translate(
+            "DICOM", "Determines whether referenced DICOM series are " \
             "offered when loading DICOM, or the automatic behavior if interaction is disabled. " \
             "Interactive selection of referenced series is the default selection")
         loadReferencesComboBox.addItem(translate("DICOM", "Ask user"), qt.QMessageBox.InvalidRole)
@@ -375,8 +374,9 @@ class _ui_DICOMSettingsPanel:
             "currentUserDataAsString", str(qt.SIGNAL("currentIndexChanged(int)")))
 
         detailedLoggingCheckBox = qt.QCheckBox()
-        detailedLoggingCheckBox.toolTip = translate("DICOM", "Log more details during DICOM operations."
-                                           " Useful for investigating DICOM loading issues but may impact performance.")
+        detailedLoggingCheckBox.toolTip = translate(
+            "DICOM", "Log more details during DICOM operations."
+            " Useful for investigating DICOM loading issues but may impact performance.")
         genericGroupBoxFormLayout.addRow(translate("DICOM", "Detailed logging:"), detailedLoggingCheckBox)
         detailedLoggingMapper = ctk.ctkBooleanMapper(detailedLoggingCheckBox, "checked", str(qt.SIGNAL("toggled(bool)")))
         parent.registerProperty(
@@ -569,9 +569,11 @@ class DICOMLoadingByDragAndDropEventFilter(qt.QWidget):
             return
 
         if not DICOMFileDialog.validDirectories(self.directoriesToAdd) or not DICOMFileDialog.validDirectories(self.filesToAdd):
-            if not slicer.util.confirmYesNoDisplay(translate("DICOM", "Import from folders with special (non-ASCII) characters in the name is not supported."
-                                                   " It is recommended to move files into a different folder and retry."
-                                                   " Try to import from current location anyway?")):
+            confirmMessage = translate(
+                "DICOM", "Import from folders with special (non-ASCII) characters in the name is not supported."
+                " It is recommended to move files into a different folder and retry."
+                " Try to import from current location anyway?")
+            if not slicer.util.confirmYesNoDisplay(confirmMessage):
                 self.directoriesToAdd = []
                 return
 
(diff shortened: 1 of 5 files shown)


lassoan:
Thank you, this looks nice. Do you have a solution for the string extraction into .ts file?

Modules/Scripted/DICOM/DICOM.py
```diff
@@ -8,6 +8,7 @@
 import slicer
 from slicer.util import settingsValue, toBool
 from slicer.ScriptedLoadableModule import *
+from slicer.translation import _
```
pieper, line 11: I don't recall if we've discussed this before, but I wonder if underscore is the right symbol to use for this.  People often use it as a dummy variable and that would overwrite this import.  Using `tr` might be less error prone, or even `_tr` since that is unlikely to be used as a variable name.

lassoan: I don't like Python's choice of `_` as translation function name (it is hard to search for it, not self-explaining, [used for other purposes](https://stackoverflow.com/questions/5893163/what-is-the-purpose-of-the-single-underscore-variable-in-python/5893946#5893946)), but [it is the Python standard](https://docs.python.org/3/library/gettext.html#internationalizing-your-programs-and-modules) and it is widely used. If we want a Pythonic solution we cannot use Qt style or invent our own function name.

Specifically about the dummy variable issue: developers seem to agree that in programs that use internationalization, underscore should be used for translation and double-underscore can be used as dummy variable.

If we all agree with the above then I can add it to the Slicer coding guidelines.

pieper: Since we aren't using `gettext` I'm not sure we need to be bound by its conventions.

If we think being pythonic is important, is there a way to make the underscore read-only so that trying to use it as a dummy variable raises an error?  I guess if they do change the value of underscore to some dummy value they will get an error when it is used as a function somewhere else, but that is indirect and not foolproof.

jcfr, line 11:
```suggestion
from slicer.translation import tr as _
```

Base/Python/slicer/translation.py
```diff
@@ -0,0 +1,27 @@
...
+            context_name = os.path.basename(parent_folder)
+            context_name += '.' + os.path.basename(source_file).replace('.py', '')
+            return context_name
+        else:
+            return os.path.basename(source_file).replace('.py', '')
+    else:
+        return os.path.basename(source_file)
+
+
+def _(text):
```
jcfr, line 21:
```suggestion
def tr(text):
```

Modules/Scripted/DICOM/DICOM.py
jcfr: This would be consistent with how django approach this. See https://docs.djangoproject.com/en/dev/topics/i18n/translation/

It would also allow things like 

```python
from slicer.translation import tr_noop as _
```

Base/Python/slicer/translation.py
```diff
@@ -0,0 +1,27 @@
...
+    """translation function for python scripted modules"""
+    from slicer import app
+
+    filename = inspect.stack()[1][1]
```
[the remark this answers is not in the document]
lassoan: I don't think it would be ever time-critical. It is for user-displayable string.
Anything in Python is not really time-critical anyway (you would not implement a hot loop in Python). If you must do it for some reason then you can easily moved the translated string into a member variable.

Modules/Scripted/DICOM/DICOM.py
lassoan: Agreed!

lassoan: An alternative would be to go all Qt. It would mean this syntax:

```
# in module main
translate = slicer.app.translate

...

def MyModuleWidget(...):

    def someFunction(self):
        ...
        myMessage = translate("MyModule", "Press any key to continue")
        myMessage = translate("MyModule", "Press {key} to continue").format(key=key)
        ...
```

The huge advantage would be that we could use pylupdate in pyside6 (can be pip-installed) for string extraction.

The disadvantages would be that pylupdate requires writing in the context name at every translation; and Qt standard is used much less frequently as Python standard.

mhdiop:
Thanks for all your valuable comments.

Whenever you decide of the translation function syntax to keep, I'll update the marked strings accordingly. By now, I keep using the `_()` syntax so that to not suspend the marking process.

I'll vote for either using `_()` or `tr()` function. I think that the Qt Style involving the use of the context as additional parameter might be a little bit heavy to use.

Regards.

Base/Python/slicer/translation.py
```diff
@@ -0,0 +1,27 @@
...
+            context_name += '.' + os.path.basename(source_file).replace('.py', '')
+            return context_name
+        else:
+            return os.path.basename(source_file).replace('.py', '')
+    else:
+        return os.path.basename(source_file)
+
+
+def _(text):
+    """translation function for python scripted modules"""
```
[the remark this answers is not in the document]
mhdiop: Thanks @jcfr, suggested updates will be made on the next commit.

mhdiop:
> Thank you, this looks nice. Do you have a solution for the string extraction into .ts file?

Thanks @lassoan for your feedback. 

For the string extraction, I updated the [sliceri18n package](https://github.com/Slicer/sliceri18n) so that it matches our needs. I test it for ts files generation and it works now.

I then made a [pull request](https://github.com/Slicer/sliceri18n/pull/8) so that such proposition get integrated in the official repository.

To use it, just call the `main.py` script and give it the slicer root folder path as an argument. It will detect all `.py` files and extract all marked strings in them before generating the corresponding ts file.

For testing purpose, you may find a Slicer code base with many python marked strings in [idrissa's repository](https://github.com/idrissa10/Slicer).

Regards.

Base/Python/slicer/translation.py
```diff
@@ -0,0 +1,27 @@
+import os
+import inspect
+
+
+def get_context(source_file):
+    """ get the translation context based on the source file name"""
```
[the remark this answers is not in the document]
mhdiop: Thanks for the suggestions, updates will be made on the next commit.

By the way, where could I find documentation on coding style rules used by the `CI / Lint` so that to avoid making it failed on next commits ?

Thanks in advance !

Modules/Scripted/DICOM/DICOM.py
jcfr: > requires writing in the context name at every translation

In a prior comment [^1], @pieper suggested to use a decorator. Instead I just resurrected the `slicer_lupdate_preprocess.py` script processing the AST and injecting the context based on the module name. See https://gist.github.com/jcfr/fc2bbb1e6b30d08c06f56d403f10145c

### Usage

```
!pip install astor

python slicer_lupdate_preprocess.py -i MyModule.py -o MyModule.lupdate.py
```

```diff
diff MyModule.py MyModule.lupdate.py
7,8c7,8
<         myMessage = _("Press any key to continue")
<         myMessage = _("Press {key} to continue").format(key=key)
---
>         myMessage = _('MyModule', 'Press any key to continue')
>         myMessage = _('MyModule', 'Press {key} to continue').format(key=key)
```

```
!pip install pyside6

$ pyside6-lupdate MyModule.lupdate.py -ts MyModule.ts -tr-function-alias translate+=tr,translate+=_
Updating 'MyModule.ts'...
    Found 2 source text(s) (2 new and 0 already existing)
```

```
$ cat MyModule.ts
<?xml version="1.0" encoding="utf-8"?>
<!DOCTYPE TS>
<TS version="2.1">
<context>
    <name>MyModule</name>
    <message>
        <location filename="MyModule.tr.py" line="8"/>
        <source>Press any key to continue</source>
        <translation type="unfinished"></translation>
    </message>
    <message>
        <location filename="MyModule.tr.py" line="8"/>
        <source>Press {key} to continue</source>
        <translation type="unfinished"></translation>
    </message>
</context>
</TS>
```

[^1]: https://github.com/Slicer/Slicer/pull/6050#pullrequestreview-819563767

Base/Python/slicer/translation.py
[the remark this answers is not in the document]
mhdiop: Why not `Translation function for python scripted modules`. Is the plural form of `function` necessary ?

lassoan:
>  I'll explore the suggested parser libraries to better handle such transformations

Yes, @mhdiop please try to create a transformer (based on use your current code and @jcfr's script with a better parser) that can robustly replace `_(...)` by `translate(...)`.

In the meantime, I'll update this pull request to temporarily use `translate(...)` and use pylupdate to add the new strings to Weblate so that we can test all other aspects of the Python module translation.

mhdiop:
Ok @lassoan, I made a transformer and it's now available [here](https://github.com/mhdiop/slicer-tools/lupdate-preprocess). I used regular expressions for transformations and it didn't require any additional parser library.

If the input is a file, the output will be a file. When the input is a folder, it detects all `.py` files it contains before launching transformations on each of them. The output is therefore a folder with the same folder structure. However, I was wondering if it's necessary for files with no calls to translation function to appear in the output folder ? Can I just skip them ?

Regards.

lassoan:
> I made a transformer and it's now available [here](https://github.com/mhdiop/slicer-tools/tree/main/lupdate-preprocess).

It sounds great! I would commit this pull request as is (using `translate(context, message)`) and then you can introduce the new `_(message)` translation format in a follow-up pull request.

lassoan (approved):
This is good as a first phase for Python module translation. Both text extraction using lupdate (using Qt-6.3) and and runtime translation (using slicer.i18n.translate) works.

Next stage wil be simplification of translation by switching to using `_(message)` function instead of using `translate(context, message)`. @mhdiop will work on this and submit a pull request when it is ready.

PR #6918, opened 2023-03-30
DOC: Fix use of admonition directives in "getting started" and "build instructions"

jcfr:
(1) Switch from

```
:::{note} Warning
:class: warning
```

to

```
:::{warning}
```

(2) Update the type of admonition:

* ArchLinux uses a rolling-release: warning -> cautious
* Build on any Distribution: warning -> important
* `Qt5_DIR` need to be set using the full path: warning -> hint
* recommended to avoid use of spaces: warning -> tip

### DIFF
--- a/Docs/conf.py
+++ b/Docs/conf.py
@@ -19,6 +19,7 @@
 import lxml.etree as ET
 import os
 import sys
+from datetime import date
 
 sys.path.insert(0, os.path.abspath('../Base/Python'))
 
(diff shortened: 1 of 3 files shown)


jcfr:
| Before | After |
|--|--|
|  image  | image|
| image| image |
|image | image |
| image | image |

That said the icon are not rendered as expected:

| image |
|--|
| https://myst-parser.readthedocs.io/en/latest/syntax/admonitions.html#admonitions |

PR #6920, opened 2023-04-02
Mark translatable strings in python scripted modules

mhdiop:
See Slicer/SlicerLanguagePacks#21


lassoan (changes requested):
This is going to be good. Please update https://github.com/Slicer/Slicer/blob/main/Utilities/Scripts/update_translations.py file to include the transformation step (replace `_(message)` by `translate(context, message)`)

A possible complication with temporarily transforming the source file is that the filename will not match. So, the complete processing would be:
- modify the the .py files in the filesystem
- rename the .py files in the .ts file (from the original to the modified)
- run lupdate
- rename the .py files in the .ts file (from the modified to the original)

mhdiop:
> Thus, I wanted to add a follow-up pull request to this one. However, when I commit my updates, I have this folowing precommit error message :
> 

The problem is now resolved and for reference, the solution I used is explained [here](https://stackoverflow.com/questions/14267441/automatically-apply-git-update-index-chmod-x-to-executable-files)

> Please update https://github.com/Slicer/Slicer/blob/main/Utilities/Scripts/update_translations.py file to include the transformation step (replace `_(message)` by `translate(context, message)`)

It's done now and a [new PR](6929) is created for that.

Regards.

Modules/Scripted/DICOMLib/DICOMExportScalarVolume.py
```diff
@@ -107,24 +108,24 @@ def export(self):
         """
         cliparameters = {}
         # Patient
-        cliparameters['patientName'] = self.tags['Patient Name']
-        cliparameters['patientID'] = self.tags['Patient ID']
-        cliparameters['patientBirthDate'] = self.tags['Patient Birth Date']
-        cliparameters['patientSex'] = self.tags['Patient Sex'] if self.tags['Patient Sex'] else "[unknown]"
-        cliparameters['patientComments'] = self.tags['Patient Comments']
+        cliparameters['patientName'] = self.tags[_('Patient Name')]
```
lassoan, line 111: I don't think these should _not_ be translated. They are set in DICOMScalarVolumePlugin.py, etc. where they are not translated either.

lassoan: Yes, sorry, I meant that these should not be translated. I'm not sure why these keys chosen to be human-readable-looking strings.

mhdiop: Thanks @lassoan and @pieper for your feedback. They will be taken into account in the next commits.

Regards.

lassoan:
Thank you, integrated in https://github.com/Slicer/Slicer/commit/60c659f33b672d8e8a7193f0741a2f56aaac080c

Issue #6922, opened 2023-04-04
Plane handler visibility is not updated correctly on using slice slider 


malbi:
## Summary

Reproduced in Slicer-5.3.0-2023-04-01-linux-amd64.

When a plane interaction handler is displayed in 2D views for a markup plane, its visibility is not updated correctly when changing slice.

## Steps to reproduce

- Create a markup plane
- Enable the translate interaction option
- Change slice displayed using the slider
- The interaction handler visibility is one slice update off and is correctly update when mouse pass over it.

https://user-images.githubusercontent.com/26160797/229798545-03eb3818-74b1-49d7-a8c7-3e2414fecb2d.mp4

## Expected behavior

Interaction handler disappear when it is not on the slice displayed.

## Environment

- Slicer version: Slicer-5.3.0-2023-04-01-linux-amd64.
- Linux Ubuntu 20.04.1

lassoan:
Thank you, I was able to reproduce the behavior on Windows, too.

@Sunderlandkyl could you have a look at this?

PR #6923, opened 2023-04-04
BUG: Interaction handler visibility is updated when changing displayed slice

malbi:
This is a proposal to fix issue #6922


Sunderlandkyl:
Thanks, lgtm!

lassoan:
Thank you @malbi, you've found the root cause of the issue. @Sunderlandkyl is looking into this now, trying to find a fix that does not require manually calling `UpdateInteractionPipeline()` again (it is already called in `Superclass::UpdateFromMRML(caller, event, callData);` but that is probably done too early, before doing).

malbi:
Hi
Do you have an update? Let me know if I can be of any help.

lassoan:
@Sunderlandkyl I remember you worked on it, did we end up getting a fix integrated for this?

lassoan:
@malbi could you please check if the problem is fixed in the latest Slicer Preview Release?

malbi:
@lassoan thanks for the update, I have tried with Slicer-5.3.0-2023-05-21-linux-amd64 and I still encounter the same behaviour.

lassoan:
Superseded by https://github.com/Slicer/Slicer/pull/6992

PR #6924, opened 2023-04-04
BUG: Fix unable to change foreground volume window level with mouse mode

jamesobutler:
When the slice viewer only had a foreground volume populated, the user was unable to change the window level with the mouse mode.

This was introduced by https://github.com/Slicer/Slicer/commit/4a73c3a7599324fdf66a0cf154bb2f85e7eb63d3


lassoan (approved):
It looks good, thanks a lot for the fix.

PR #6931, opened 2023-04-09
DOC: updated run solution for Debian 10.12

kerim371:
I just did the test of running SlicerCAT on Debian 10.12. To solve `QT_QPA_PLATFORM` error I needed to create symlink. But this time I needed to create it to `libxcb-util.so.0.0.0`. As I there was no `libxcb-util.so`.
Most likely the correct library is always `libxcb-util.so.0.0.0`. But I'm not sure.
We could leave these two options so the user could check his system.

This solution also comes from the link above.

### DIFF
--- a/Docs/user_guide/getting_started.md
+++ b/Docs/user_guide/getting_started.md
@@ -101,9 +101,13 @@ Debian 10.12 users may encounter an error when launching Slicer:
 
     Available platform plugins are: xcb.
 
-[The solution](https://forum.qt.io/topic/93247/qt-qpa-plugin-could-not-load-the-qt-platform-plugin-xcb-in-even-though-it-was-found/81) is to create symlink:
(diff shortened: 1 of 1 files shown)


kerim371:
I remember I added this solution almost a year ago.
The [link](https://forum.qt.io/post/653414) points exactly to the solution with `libxcb-util.so.0.0.0` library.
And `libxcb-util.so.0.0.0` works correctly (just tested).

But the previous aswer on the same Qt forum includes `libxcb-util.so` wich doesn't exist on my fresh `Debian 10.12 tester`.
Probably I've mistaken a year ago and the correct solution always includes `libxcb-util.so.0.0.0`. 
But I think we could leave these two options so the user could check what version of libarary he have.

pieper (approved):
LGTM; thanks for the update 👍

PR #6932, opened 2023-04-09
ENH: Add option from subject hierarchy to set segment opacity

cpinter:
https://user-images.githubusercontent.com/1325980/230777122-254432bc-54ce-4e13-8913-e2fb3d6c678d.mp4

### DIFF
--- a/Modules/Loadable/Segmentations/SubjectHierarchyPlugins/qSlicerSubjectHierarchySegmentsPlugin.cxx
+++ b/Modules/Loadable/Segmentations/SubjectHierarchyPlugins/qSlicerSubjectHierarchySegmentsPlugin.cxx
@@ -39,11 +39,12 @@
 #include "vtkMRMLSubjectHierarchyNode.h"
 
 // Qt includes
+#include <QAction>
 #include <QDebug>
 #include <QIcon>
-#include <QMessageBox>
-#include <QAction>
 #include <QMenu>
+#include <QMessageBox>
+#include <QWidgetAction>
 
 // Slicer includes
 #include "qSlicerApplication.h"
@@ -52,6 +53,9 @@
 
 // MRML widgets includes
 #include "qMRMLNodeComboBox.h"
+
+// CTK includes
+#include "ctkDoubleSlider.h"
 
 //-----------------------------------------------------------------------------
 /// \ingroup SlicerRt_QtModules_Segmentations
@@ -71,6 +75,9 @@ public:
   QAction* ShowAllSegmentsAction{nullptr};
   QAction* JumpSlicesAction{nullptr};
   QAction* CloneSegmentAction{nullptr};
+  QAction* OpacityAction{nullptr};
+  QMenu* OpacityMenu{nullptr};
+  ctkDoubleSlider* OpacitySlider{nullptr};
 };
 
 //-----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


Modules/Loadable/Segmentations/SubjectHierarchyPlugins/qSlicerSubjectHierarchySegmentsPlugin.cxx
```diff
@@ -105,6 +112,20 @@ void qSlicerSubjectHierarchySegmentsPluginPrivate::init()
...
+  this->OpacitySlider = new ctkDoubleSlider(this->OpacityMenu);
+  this->OpacitySlider->setOrientation(Qt::Horizontal);
+  this->OpacitySlider->setRange(0.0, 1.0);
+  this->OpacitySlider->setSingleStep(0.1);
+  QObject::connect(this->OpacitySlider, SIGNAL(valueChanged(double)), q, SLOT(setOpacityForCurrentItem(double)));
+  QWidgetAction* opacityAction = new QWidgetAction(this->OpacityMenu);
+  opacityAction->setDefaultWidget(this->OpacitySlider);
+  this->OpacityMenu->addAction(opacityAction);
+
+  this->OpacityAction = new QAction("Opacity",q);
```
lassoan, line 126:
```suggestion
  this->OpacityAction = new QAction(qSlicerSubjectHierarchySegmentsPlugin::tr("Opacity"),q);
```

Modules/Loadable/Segmentations/SubjectHierarchyPlugins/qSlicerSubjectHierarchySegmentsPlugin.cxx
```diff
@@ -105,6 +112,20 @@ void qSlicerSubjectHierarchySegmentsPluginPrivate::init()
...
+  this->OpacityAction->setToolTip("Set segment opacity in the sub-menu");
```
lassoan, line 127:
```suggestion
  this->OpacityAction->setToolTip(qSlicerSubjectHierarchySegmentsPlugin::tr("Set segment opacity in the sub-menu"));
```

Modules/Loadable/Segmentations/SubjectHierarchyPlugins/qSlicerSubjectHierarchySegmentsPlugin.cxx
```diff
@@ -988,3 +1036,45 @@ bool qSlicerSubjectHierarchySegmentsPlugin::showItemInView(
     qSlicerSubjectHierarchyPluginHandler::instance()->pluginByName("Segmentations") );
   return segmentationsPlugin->showItemInView(segmentationItemId, viewNode, allItemsToShow);
 }
+
+//---------------------------------------------------------------------------
+void qSlicerSubjectHierarchySegmentsPlugin::setOpacityForCurrentItem(double opacity)
+{
+  // Get currently selected node and scene
+  vtkMRMLSubjectHierarchyNode* shNode = qSlicerSubjectHierarchyPluginHandler::instance()->subjectHierarchyNode();
```
lassoan, line 1044: This 30 lines to get the display node and segment ID is very similar to those in `qSlicerSubjectHierarchySegmentsPlugin::showVisibilityContextMenuActionsForItem(vtkIdType itemID)`. It would be nice to factor it out into a helper function to avoid repetition.

lassoan (commented):
This is very useful, thank you, just added a few trivial comments.

Modules/Loadable/Segmentations/SubjectHierarchyPlugins/qSlicerSubjectHierarchySegmentsPlugin.cxx
cpinter: I thought about this, but the purpose of such convenience function is not clear. It could return the segmentation node, its display node, and the segment ID based on the item ID, but then the method signature would be quite confusing (we are not in Python where we can return tuples), and using it would not save a noticeable amount of lines of code, I'd guess about ten, or maybe nothing, depending on how we check the result.

lassoan: I had a look at this and the saving would be significant (about 80-100 lines of code) but the code would also become a bit more complex, so the overal gain is not much. The change would be also somewhat complicated and error-prone to do. Maybe this refactoring could be done later when other changes are made to this class.

lassoan (approved):

cpinter:
Thank you Andras!

Modules/Loadable/Segmentations/SubjectHierarchyPlugins/qSlicerSubjectHierarchySegmentsPlugin.cxx
cpinter: Thanks for looking into this! If we only consider the functional code I agree this could save several dozens of lines of code. As I was considering the new lines we would need to add (sanity checks, validating output, etc), then it started to seem that from the line count point of view it may not be worth it.

PR #6935, opened 2023-04-10
BUG: Fix Python wrapping of methods using enum arguments

lassoan:
`typedef enum {value1, value2, ...} someTypeName;` is not supported by VTK's Python wrapper (the method that uses such enums are not Python-wrapped). Need to use `enum someTypeName {value1, value2, ...};` syntax instead.

Fixes issue reported here: https://discourse.slicer.org/t/how-to-set-clipping-method-of-vtkmrmlclipmodelsnode/28816

### DIFF
--- a/Base/QTCLI/vtkSlicerCLIModuleLogic.h
+++ b/Base/QTCLI/vtkSlicerCLIModuleLogic.h
@@ -31,8 +31,6 @@ class MRMLIDMap;
 
 #include "qSlicerBaseQTCLIExport.h"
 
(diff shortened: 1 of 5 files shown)


jcfr:
Looks like the branch https://github.com/Slicer/Slicer/tree/integ was pushed into the main repo by mistake, I am deleting it.

jcfr:
That said, these two commits have been integrated:

image

@lassoan I will let you cleanup and update relevant pull-requests if needed,

PR #6937, opened 2023-04-11
Update crosshair cursor position on 3D view mouse move events

LucasGandel:
Following discussions on #6910, here is a new proposal for the addition of a virtual cursor in 3DSlicer.
The new approach consists in a model node being updated on CursorPositionModifiedEvent events.

As discussed with @lassoan, this MR:
1. applies the required changes to the 3D view interactor style for the cursor position to be updated in the 3D view too.
2. provides a small python script to test how the feature would behave once fully integrated: [TrackCursor.zip](https://github.com/Slicer/Slicer/files/11201330/TrackCursor.zip)

The goal is to confirm that such feature could be integrated in its own menu, and identify what would be the options that are exposed to users in terms of rendering properties, cursor models, Slice visibility... 

A few notes regarding the status of the example script compared to the expected feature:
- The mouse pointer should be hidden if the virtual cursor is enabled and the mouse enters the 3D view
- Picking of the cursor model must be disabled (I tried using SelectableOff but it does not seem to have any effect)
- The commented code can be added back to try a line representation, but not sure which direction makes more sense for the line
- The marker should probably have a constant size in screen space coordinates

### DIFF
--- a/Libs/MRML/DisplayableManager/vtkMRMLThreeDViewInteractorStyle.cxx
+++ b/Libs/MRML/DisplayableManager/vtkMRMLThreeDViewInteractorStyle.cxx
@@ -98,6 +98,19 @@ bool vtkMRMLThreeDViewInteractorStyle::DelegateInteractionEventToDisplayableMana
     {
     // set "inaccurate" world position
     ed->SetWorldPosition(worldPosition, false);
+
+    // update the cursor position on mouse move
+    if (this->GetCameraNode() != nullptr
+      && this->GetCameraNode()->GetScene() != nullptr
+      && inputEventData->GetType() == vtkCommand::MouseMoveEvent)
+      {
+      vtkMRMLScene* scene = this->GetCameraNode()->GetScene();
+      vtkMRMLCrosshairNode* crosshairNode = vtkMRMLCrosshairDisplayableManager::FindCrosshairNode(scene);
+      if (crosshairNode)
+        {
+        crosshairNode->SetCursorPositionRAS(worldPosition);
+        }
+      }
     }
   ed->SetMouseMovedSinceButtonDown(this->MouseMovedSinceButtonDown);
   ed->SetAccuratePicker(this->AccuratePicker);


LucasGandel:
@lassoan Is there anything we could do to move forward with this change? We can integrate the feature in a new menu if it is more convenient for you to test.
Many thanks.

LucasGandel:
Friendly ping @lassoan .
Or could this be merged as-is and have the stereo mouse cursor implemented in an extension and available through the extension manager?
Thanks.

Libs/MRML/DisplayableManager/vtkMRMLThreeDViewInteractorStyle.cxx
```diff
@@ -105,6 +105,17 @@ bool vtkMRMLThreeDViewInteractorStyle::DelegateInteractionEventToDisplayableMana
 
   ed->SetAttributesFromInteractor(this->GetInteractor());
 
+  if (inputEventData->GetType() == vtkCommand::MouseMoveEvent)
+  {
```
lassoan, line 109: indentation

Libs/MRML/DisplayableManager/vtkMRMLThreeDViewInteractorStyle.cxx
```diff
@@ -105,6 +105,17 @@ bool vtkMRMLThreeDViewInteractorStyle::DelegateInteractionEventToDisplayableMana
...
+    // Update the cursor position
+    vtkMRMLScene * scene = this->GetCameraNode()->GetScene();
+    vtkMRMLCrosshairNode * crosshairNode = vtkMRMLCrosshairDisplayableManager::FindCrosshairNode(scene);
+    if (crosshairNode)
+    {
+      crosshairNode->SetCursorPositionRAS(worldPosition);
```
lassoan, line 115: only set it if it is valid (`QuickPick` returned true)

Maybe you can move all the new lines inside the `if (this->QuickPick(...))` branch.

Libs/MRML/DisplayableManager/vtkMRMLThreeDViewInteractorStyle.cxx
```diff
@@ -105,6 +105,17 @@ bool vtkMRMLThreeDViewInteractorStyle::DelegateInteractionEventToDisplayableMana
 
   ed->SetAttributesFromInteractor(this->GetInteractor());
 
+  if (inputEventData->GetType() == vtkCommand::MouseMoveEvent)
+  {
+    // Update the cursor position
+    vtkMRMLScene * scene = this->GetCameraNode()->GetScene();
```
lassoan, line 111: check if cameranode is not nullptr

lassoan (commented):
There are some trivial issues, but overall this change will be OK.

lassoan:
> The mouse pointer should be hidden if the virtual cursor is enabled and the mouse enters the 3D view

You may need to implement an API to change or hide the default arrow cursor.

> Picking of the cursor model must be disabled (I tried using SelectableOff but it does not seem to have any effect)

Maybe `SelectableOff` might need some extra attention when using quick picking? You can debug into the quick picking code to see why the cursor actor is not excluded from picking.

> The commented code can be added back to try a line representation, but not sure which direction makes more sense for the line
The marker should probably have a constant size in screen space coordinates

You can do what is done in markups module: scale based on the camera focal point position. You will still have some depth cue from size (closer objects appear larger), but the object size will remain reasonable.

You may also consider displaying the cursor as a markup fiducial point, which already offers relative sizing. There are a variety of cursor shapes and you can add new shapes if needed. Markups also support occluded visibility (so that you can see the marker even if it would have been partially occluded by opaque (or almost-opaque) objects. You can lock the markup to prevent interference with interactions.

LucasGandel:
Thank you very much for your help and consideration @lassoan .
I have addressed your comments, and solved the open questions in this updated version of the script: 
[TrackCursor.zip](https://github.com/Slicer/Slicer/files/11653634/TrackCursor.zip)

In the script:
- Hiding the mouse cursor: `slicer.app.setOverrideCursor` is called based on the validity of the cursor position
- `QuickPick` will pick any actor writing to the ZBuffer. An opacity of 0.9999 is given to the markups node's display node to work around this.
- Markups node are now used, thank you very much for the suggestions.

If the behavior is OK, what would be the best way to provide this functionality in the official Slicer releases? Should I implement an extension that add a button to Slicer's UI, and that calls the content of the script when this button is pressed?

LucasGandel:
@lassoan The content of the python script will be moved to an extension (ongoing work [here](https://github.com/KitwareMedical/SlicerVirtualMouseCursor)). 
Can you do a second pass and merge this MR if everything is fine please?

LucasGandel:
Rebased. Checks have passed.

@jcfr please review

jcfr (approved):

PR #6942, opened 2023-04-17
BUG: Mark translatable strings in SegmentStatistics module

mhdiop:
See Slicer/SlicerLanguagePacks#21

### DIFF
--- a/Modules/Scripted/SegmentStatistics/SegmentStatistics.py
+++ b/Modules/Scripted/SegmentStatistics/SegmentStatistics.py
@@ -6,6 +6,8 @@ import vtk
 
 import slicer
 from slicer.ScriptedLoadableModule import *
+from slicer.i18n import tr as _
+from slicer.i18n import translate
 
 from SegmentStatisticsPlugins import *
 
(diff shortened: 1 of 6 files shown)


Modules/Scripted/SegmentStatistics/SegmentStatistics.py
```diff
@@ -332,9 +333,9 @@ def setup(self):
         self.doneButton.connect('clicked()', lambda: self.done(1))
 
     def _addPluginOptionWidgets(self):
-        description = "Edit segment statistics plugin parameters:"
+        description = _("Edit segment statistics plugin parameters:")
         if self.pluginName:
-            description = "Edit " + self.pluginName + " plugin parameters:"
+            description = _("Edit {pluginName} plugin parameters:").format(pluginName=self.pluginName)
```
lassoan, line 338: We need to introduce a separate plugin Title property, which is displayed on the GUI and it is translated, while plugin name is not translatable.

Modules/Scripted/SegmentStatistics/SegmentStatisticsPlugins/ScalarVolumeSegmentStatisticsPlugin.py
```diff
@@ -143,55 +144,55 @@ def getMeasurementInfo(self, key):
         # @moselhy also could not find DICOM quantity code for "median"
 
         info["voxel_count"] = \
-            self.createMeasurementInfo(name="Voxel count", description="Number of voxels", units="voxels",
-                                       quantityDicomCode=self.createCodedEntry("nvoxels", "99QIICR", "Number of voxels", True),
-                                       unitsDicomCode=self.createCodedEntry("voxels", "UCUM", "voxels", True))
+            self.createMeasurementInfo(name="Voxel count", description=_("Number of voxels"), units="voxels",
```
lassoan, line 147: Need to add a measurement Title property, which is displayed and translated.

Column header is a bit complicated. We need to introduce a "Title" property to tables (it can be stored in table column metadata). If column title is specified then on the GUI the column title is displayed instead of the column name. This requires changes in the Tables module. Once those changes are completed, we can complete translation of this module.

lassoan (commented):
Very good, but we need to translate measurement names and column names. Neither of them is easy.

lassoan:
I'll merge this as is and translation of plugin title and column title can be addressed in a separate pull request.

PR #6946, opened 2023-04-17
ENH: Add vtkITKGrowCut filter

Sunderlandkyl:
This commit adds a new VTK filter vtkITKGrowCut, which uses the itk::FastGrowCut implementation.
The new filter is used in the "Grow from seeds" segment editor effect.
SegmentationsModuleTest2 now includes a Grow from seeds test.

### DIFF
--- a/Libs/vtkITK/vtkITKGrowCut.cxx
+++ b/Libs/vtkITK/vtkITKGrowCut.cxx
@@ -251,7 +251,8 @@ void vtkITKGrowCut::SetDistancePenalty(double distancePenalty)
   if (this->DistancePenalty != distancePenalty)
     {
     this->DistancePenalty = distancePenalty;
-    // TODO: Due to a bug in ITK, we currently need to reset the filter when changing the distance penalty.
+    // TODO: Due to a bug in ITK (https://github.com/InsightSoftwareConsortium/ITKGrowCut/issues/18),
+    // we currently need to reset the filter when changing the distance penalty.
     // Otherwise the internal computed distance values will not be recomputed, and the change to distance penalty will not be applied.
     this->Reset();
     this->Modified();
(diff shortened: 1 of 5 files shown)


lassoan (commented):
Thank you @Sunderlandkyl this works very well. I've found only one small issue that would need to be fixed before integrating: seed locality does not have an effect unless you cancel and re-initialize.

Example:

Default:

image

Setting Seed locality to 10 has no effect:

image

After Cancel and Initialize it provides correct result:

image

It would probably need to be fixed in ITK (or at least file a bug report in ITK) to invalidate the internal computed distances if the distance penalty value is modified.

But as an immediate workaround in Slicer (to enable merging without waiting for the fix in ITK) could be to cancel and initialize the filter.

lassoan:
@Sunderlandkyl the number of open issues for the new stable release (https://github.com/Slicer/Slicer/milestone/6) is decreasing quite rapidly, so it seems that we can cut the new release soon. It would be nice if you could fix the small seed locality update issue (see above) by early next week because then this PR could be included in this release. Thanks in advance!

Sunderlandkyl:
Thanks for the review Andras!
I've added a Reset() call when the distance penalty is changed.

Libs/vtkITK/vtkITKGrowCut.cxx
```diff
@@ -0,0 +1,259 @@
...
+}
+
+//-----------------------------------------------------------------------------
+void vtkITKGrowCut::SetDistancePenalty(double distancePenalty)
+{
+  vtkDebugMacro(<< this->GetClassName() << " (" << this << "): setting DistancePenalty to " << distancePenalty);
+  if (this->DistancePenalty != distancePenalty)
+    {
+    this->DistancePenalty = distancePenalty;
+    // TODO: Due to a bug in ITK, we currently need to reset the filter when changing the distance penalty.
```
lassoan, line 254:
```suggestion
    // TODO: Due to a bug in ITK (https://github.com/InsightSoftwareConsortium/ITKGrowCut/issues/18),
    // we currently need to reset the filter when changing the distance penalty.
```

lassoan (approved):
Thank you! I've added an ITK bug report (https://github.com/InsightSoftwareConsortium/ITKGrowCut/issues/18)

PR #6947, opened 2023-04-18
ENH: Initialize global pause render state for newly created views

Sunderlandkyl:
Previously, newly created views were initialized with the default pause render state of 0. This caused warning messages if the layout was changed between pauseRender/resumeRender calls as the render state of newly created views would not match the existing ones.

This commit tracks the global pauseRender count, and uses it to initialize the pauseRender state of newly created views. This removes most instances of the warning message:
> [Qt] int __cdecl ctkVTKAbstractView::resumeRender(void) Cannot resume rendering, pause render count is already 0!

### DIFF
--- a/Libs/MRML/Widgets/qMRMLLayoutManager.cxx
+++ b/Libs/MRML/Widgets/qMRMLLayoutManager.cxx
@@ -1212,23 +1212,35 @@ QWidget* qMRMLLayoutManager::viewWidget(vtkMRMLNode* viewNode) const
 }
 
 //------------------------------------------------------------------------------
+QList<QWidget*> qMRMLLayoutManager::viewWidgets() const
+{
+  Q_D(const qMRMLLayoutManager);
+  QList<QWidget*> viewWidgets;
+  for (qMRMLLayoutViewFactory* factory : this->mrmlViewFactories())
+    {
+    for (int i = 0; i < factory->viewCount(); ++i)
+      {
+      viewWidgets.append(factory->viewWidget(i));
+      }
+    }
+  return viewWidgets;
+}
+
+//------------------------------------------------------------------------------
 void qMRMLLayoutManager::setRenderPaused(bool pause)
 {
   // Note: views that are instantiated between pauseRender() calls will not be affected
   // by the specified pause state
   Q_D(qMRMLLayoutManager);
-  qMRMLLayoutViewFactory* sliceViewFactory = this->mrmlViewFactory("vtkMRMLSliceNode");
-  foreach(const QString& viewName, sliceViewFactory->viewNodeNames())
-    {
-    ctkVTKAbstractView* view = this->sliceWidget(viewName)->sliceView();
-    view->setRenderPaused(pause);
-    }
 
-  qMRMLLayoutViewFactory* threeDViewFactory = this->mrmlViewFactory("vtkMRMLViewNode");
-  foreach(const QString& viewName, threeDViewFactory->viewNodeNames())
+  QList<QWidget*> viewWidgets = this->viewWidgets();
+  for (QWidget* widget : viewWidgets)
     {
-    ctkVTKAbstractView* view = this->threeDWidget(viewName)->threeDView();
-    view->setRenderPaused(pause);
+    qMRMLAbstractViewWidget* viewWidget = qobject_cast<qMRMLAbstractViewWidget*>(widget);
+    if (viewWidget)
+      {
+      viewWidget->setRenderPaused(pause);
+      }
     }
 
   if (pause)
--- a/Libs/MRML/Widgets/qMRMLLayoutManager.cxx
+++ b/Libs/MRML/Widgets/qMRMLLayoutManager.cxx
@@ -1221,21 +1221,7 @@ void qMRMLLayoutManager::setRenderPaused(bool pause)
   foreach(const QString& viewName, sliceViewFactory->viewNodeNames())
     {
     ctkVTKAbstractView* view = this->sliceWidget(viewName)->sliceView();
-    if (!pause && !view->isRenderPaused())
-      {
-      // This view is already resumed and a resume is requested.
-      // This is probably because the view has just been added
-      // and so it missed a few PauseRender requests.
-      // Do not try to resume render to avoid logging of a warning,
-      // and just log a debug message here to help with troubleshooting if
-      // any problem comes up related to pausing rendering.
-      qDebug() << Q_FUNC_INFO << "Resume render request is ignored for view "
-        << viewName << ", probably the view has just been created";
-      }
-    else
-      {
-      view->setRenderPaused(pause);
-      }
+    view->setRenderPaused(pause);
     }
 
   qMRMLLayoutViewFactory* threeDViewFactory = this->mrmlViewFactory("vtkMRMLViewNode");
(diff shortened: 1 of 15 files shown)


Libs/MRML/Widgets/qMRMLLayoutManager.cxx
```diff
@@ -260,6 +263,10 @@ QWidget* qMRMLLayoutSliceViewFactory::createViewFromNode(vtkMRMLAbstractViewNode
   sliceWidget->setMRMLScene(this->mrmlScene());
   sliceWidget->setSliceLogics(this->sliceLogics());
   this->sliceLogics()->AddItem(sliceWidget->sliceLogic());
+  for (int i = 0; i < this->layoutManager()->globalPauseRenderCount(); ++i)
```
lassoan, line 266: This should be done for all new views (not just for 3D and slice views). Could we do it at one place, right after we creating the view, in qMRMLLayoutViewFactory::onViewNodeAdded?

Libs/MRML/Widgets/qMRMLLayoutManager.cxx
```diff
@@ -1244,6 +1237,15 @@ void qMRMLLayoutManager::setRenderPaused(bool pause)
     ctkVTKAbstractView* view = this->threeDWidget(viewName)->threeDView();
     view->setRenderPaused(pause);
     }
+
+  if (pause)
+    {
+    d->GlobalPauseRenderCount += 1;
```
lassoan, line 1243:
```suggestion
    d->GlobalPauseRenderCount++;
```

Libs/MRML/Widgets/qMRMLLayoutManager.cxx
```diff
@@ -1244,6 +1237,15 @@ void qMRMLLayoutManager::setRenderPaused(bool pause)
...
+    }
+  else
+    {
+    d->GlobalPauseRenderCount = std::max(d->GlobalPauseRenderCount - 1, 0);
```
lassoan, line 1247: Let's print a warning if it goes below 0.

Libs/MRML/Widgets/qMRMLLayoutManager_p.h
```diff
@@ -155,6 +155,7 @@ public slots:
   /// Top-level windows created and managed by this class that host additional viewports
   /// (outside the main application window).
   QMap<QString, ViewportInfo>  DetachedViewports;
+  int GlobalPauseRenderCount{ 0 };
```
lassoan, line 158: It is not quite global, just all the views that this layout manager owns. Could you rename it to "AllViewsPauseRenderCount" or something similar?

Libs/MRML/Widgets/qMRMLLayoutManager.cxx
```diff
@@ -1221,21 +1228,7 @@ void qMRMLLayoutManager::setRenderPaused(bool pause)
...
-      // and just log a debug message here to help with troubleshooting if
-      // any problem comes up related to pausing rendering.
-      qDebug() << Q_FUNC_INFO << "Resume render request is ignored for view "
-        << viewName << ", probably the view has just been created";
-      }
-    else
-      {
-      view->setRenderPaused(pause);
-      }
+    view->setRenderPaused(pause);
```
lassoan, line 1231: Please still keep the warning message if trying to unpause render on a view that has no paused rendering on it. Thanks to your fixes, we should no longer see the message.

lassoan (commented):
Thank you, this will be very good. I've added a few comments inline.

Libs/MRML/Widgets/qMRMLLayoutManager.cxx
Sunderlandkyl: This isn't the warning message that we are used to seeing, This is a debug message that was hiding the warning message for slice views. The actual warning message is still in CTK: https://github.com/commontk/CTK/blob/master/Libs/Visualization/VTK/Widgets/ctkVTKAbstractView.cpp#L285.

lassoan: It is very good then, thanks for the clarification.

Libs/MRML/Widgets/qMRMLLayoutViewFactory.cxx
```diff
@@ -407,23 +406,13 @@ void qMRMLLayoutViewFactory::onViewNodeAdded(vtkMRMLAbstractViewNode* node)
...
-    view = threeDWidget->threeDView();
-    }
-  qMRMLSliceWidget* sliceWidget = qobject_cast<qMRMLSliceWidget*>(viewWidget);
-  if (sliceWidget)
-    {
-    view = sliceWidget->sliceView();
-    }
-
-  if (view)
+  qMRMLAbstractViewWidget* abstractView = qobject_cast<qMRMLAbstractViewWidget*>(viewWidget);
```
lassoan, line 409: it is not a view but a widget, so a name like "viewWidget" would be better

Libs/MRML/Widgets/qMRMLPlotWidget.cxx
```diff
@@ -116,37 +119,47 @@ void qMRMLPlotWidget::setMRMLPlotViewNode(vtkMRMLPlotViewNode* newPlotViewNode)
   d->PlotController->setMRMLPlotViewNode(newPlotViewNode);
 }
 
-// --------------------------------------------------------------------------
+//--------------------------------------------------------------------------
+void qMRMLPlotWidget::setMRMLAbstractViewNode(vtkMRMLAbstractViewNode* newPlotViewNode)
+{
+  this->setMRMLPlotViewNode(vtkMRMLPlotViewNode::SafeDownCast(newPlotViewNode));
```
lassoan, line 125: we should log a warning if a wrong view node type is provided (not just here but in all other widget classes, too).

Libs/MRML/Widgets/qMRMLTableWidget.cxx
```diff
@@ -133,22 +149,22 @@ qMRMLTableView* qMRMLTableWidget::tableView()const
...
-//---------------------------------------------------------------------------
-QString qMRMLTableWidget::viewLabel()const
+// --------------------------------------------------------------------------
+qMRMLViewControllerBar* qMRMLTableWidget::controllerWidget()const
 {
   Q_D(const qMRMLTableWidget);
-  return d->TableController->viewLabel();
-}
+  return this->tableController();
+}
```
lassoan, line 170: newline is missing at the end?

lassoan (commented):
Thank you, this looks great, I've just added a few trivial comments inline.

Libs/MRML/Widgets/qMRMLLayoutViewFactory.cxx
Sunderlandkyl: "viewWidget" is already in use for the QWidget* variable. How about "abstractViewWidget"?

lassoan: or maybe rename the `viewWidget` above to `widget` (as it is more generic)

Sunderlandkyl: Ok, I've renamed it "viewWidget"->"widget", and "abstractView"->"viewWidget".

lassoan (approved):
This looks good, thank you!

PR #6948, opened 2023-04-18
BUG: Mark translatable strings in Segmentations module

mhdiop:
See Slicer/SlicerLanguagePacks#12

### DIFF
--- a/Modules/Loadable/Segmentations/EditorEffects/qSlicerSegmentEditorEraseEffect.cxx
+++ b/Modules/Loadable/Segmentations/EditorEffects/qSlicerSegmentEditorEraseEffect.cxx
@@ -52,7 +52,7 @@ qSlicerSegmentEditorEraseEffect::qSlicerSegmentEditorEraseEffect(QObject* parent
  : Superclass(parent)
  , d_ptr( new qSlicerSegmentEditorEraseEffectPrivate(*this) )
 {
-  this->m_Name = QString("Erase");
+  this->m_Name = QString(/*no tr*/"Erase");
   this->m_AlwaysErase = true;
   this->m_ShowEffectCursorInThreeDView = true;
 }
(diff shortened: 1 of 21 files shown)


Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.cxx
```diff
@@ -328,15 +328,16 @@ qMRMLSegmentEditorWidgetPrivate::qMRMLSegmentEditorWidgetPrivate(qMRMLSegmentEdi
   this->EffectNameOrder
     // Thresholding is the the starting point for most segmentations
     // (it can often create usable segmentation by itself, or used to define intensity range for painting)
-    << "Threshold"
+    << qMRMLSegmentEditorWidget::tr("Threshold")
```
lassoan, line 331: These are effect names here (not effect titles) therefore they must not be translated

Modules/Loadable/Segmentations/EditorEffects/qSlicerSegmentEditorEraseEffect.cxx
```diff
@@ -52,7 +52,7 @@ qSlicerSegmentEditorEraseEffect::qSlicerSegmentEditorEraseEffect(QObject* parent
  : Superclass(parent)
  , d_ptr( new qSlicerSegmentEditorEraseEffectPrivate(*this) )
 {
-  this->m_Name = QString("Erase");
+  this->m_Name = QString(/*no tr*/"Erase");
```
lassoan, line 55: We need to introduce a new `Title` property to segment editor effects. It would be the string that is displayed on the GUI as the effect's name (while `Name` would remain non-translated and that would be used to refer to the effect programmatically).

lassoan (commented):
Looks very good, but we must translate the effect names. It is not a simple task (need to introduce a new property to segment editor effects) but you should be able to manage it. Let me know if you need any help.

lassoan:
@mhdiop What do you think? Are you going to give it a try to implement the new `Title` property for effects? It could be a new method that by default (if not overridden) would return the effect name.

mhdiop:
Thanks @lassoan for your valuable feedback.

> @mhdiop, are you going to give it a try to implement the new `Title` property for effects?

I'm currently away for [Eid](https://en.wikipedia.org/wiki/Eid_al-Fitr) purposes. I can however work on this before the next meeting.

Regards.

lassoan:
I've fixed the small issues and will merge this pull request as is.

@mhdiop You can submit the effect title implementation in a new pull request.

mhdiop:
Thank you @lassoan, it's OK for me. I'll work on the effect title implementation and submit a new PR.

Regards.

PR #6951, opened 2023-04-19
BUG: Restore synchronized sequence table header visibility

Sunderlandkyl:
The horizontalHeaderVisible for tableWidget_SynchronizedSequenceNodes was set to "false". Changed horizontalHeaderVisible back to "true".


lassoan (approved):
Good catch, thank you!

PR #6954, opened 2023-04-20
DOC: Add note to data loading page on LPS/RAS coordinate systems

lassoan:


pieper (approved):
If you are okay with my changes we can go with this or keep iterating.

Docs/user_guide/data_loading_and_saving.md
```diff
@@ -58,6 +58,18 @@ Multiple nodes can be exported at once by placing them into a folder and then by
 
 ## Supported Data Formats
 
+:::{note}
+
+### On use of LPS/RAS coordinate systems
+
+DICOM and most medical imaging software uses the **LPS coordinate system** for storing all data. Axis directions of this coordinate system are patient left, posterior, and superior. Unit is millimeters. The choice of origin is arbitrary because only relative differences have meaning, so there is no universal standard, but it is often some geometric center of the imaging system is chosen to be near the center of some object of interest.
```
lassoan, line 65:
```suggestion
DICOM and most medical imaging software uses the **LPS coordinate system** for storing all data. Axis directions of this coordinate system are patient left, posterior, and superior. Unit is millimeters. The choice of origin is arbitrary because only relative differences have meaning, so there is no universal standard, but it is often set to some geometric center of the imaging system, or it is chosen to be near the center of an object of interest.
```

lassoan:
@pieperyour changes looked good, I've just made a small clarification, please check https://github.com/Slicer/Slicer/pull/6954/commits/88bab46d203e85a8a37ee99da989985143962d54

pieper (approved):
Yes, looks great.  I'll merge and we can always start another PR if we want to tweak further.

jamesobutler:
Thanks for the added clarification to the docs! This is frequently a confusion among users and even the continued differences of how it is saved vs how it is used internally can be confusing, but at least this helps some to understand what is going on.

pieper:
I agree, there's a lot of confusion about this.  I wonder if we should be more "vocal" when loading data where we don't really know the LPS/RAS status.  Like a dialog box with a "don't show again" option saying that we had to make an assumption and pointing people to the docs.

lassoan:
I usually immediately close any unsolicited popups after reading the first 1-2 words and I think most people do the same (maybe reaching 5 words). So, the end result is just some annoyance if the popup keeps coming up and mild discomfort if the popup has a "do not show again" checkbox.

I think the only really good solution to this will be to switch to LPS in Slicer, but that's a big task.

pieper:
I disagree.  Yes, popups are annoying but less so than data being in mysterious coordinate systems with no clue why.  The fact that we get so many questions about this indicates that there are many many more people having trouble and not even realizing they should be looking at documentation or asking on the forum.  Switching Slicer to LPS isn't a bad option, but it won't solve this problem since there are other systems that use RAS and lots of existing data that is not explicit about the coordinate system in use.  

If not a popup then maybe we have a decoration/flag in the subject hierarchy that indicates assumptions were made when importing the data.  A tooltip could provide more details.  A context menu item could offer to pick the other option.

lassoan:
I see that the LPS<->RAS conversion is unexpected and would be nice to do something about it but a popup would be ineffective (pretty good discussion [here](https://ux.stackexchange.com/questions/4260/do-users-read-pop-up-messages-notifications-error-messages-etc)).

The decoration+corrective action in the subject hierarchy would be a nice approach: it would never block the user if everything works correctly, and user would have a chance to correct the problem anytime when he realizes that something is off. Adding the right-click menu in subject hiearchy would be very easy to implement, but the decoration would be hard, as the item icon is provided by the owner plugin and so we would need to implement infrastructure to be allow other plugins to request icon overlays (or implement an extra status column and allos plugins to add messages there and provide custom contxt menu items for it).

jamesobutler:
+1 for allowing users to easily swap the coordinate system between LPS and RAS when they notice something is wrong. Creating a transform node and applying it to the object is too difficult for novice/occasional users.

PR #6956, opened 2023-04-22
BUG: Fix dynamic update of transform visualization

lassoan:
Transform displayable managers rely on node reference to get notified about input "glyph points" node changes. Glyph points node designates where to draw glyphs. User can select it in Transforms module / Display / Advanced / Source points.

When points of a markup node change then the markup node does not invoke a Modified event, but a custom PointModified event. However, the vtkMRMLTransformDisplayNode (in MRML core) does not know about markup node custom events (in Markups module).

A solution could have been to move markups nodes into MRML core, but moving all node types into MRML core is not a sustainable strategy.

Instead, MRML node reference infrastructure was enhanced with a new feature: instead of just specifying an event list, it is now also possible to request observing ContentModifiedEvents that a node provides. (ContentModifiedEvents property was originally developed for sequence recording, to allow Sequences module to detect changes in any node type.) This new feature allows any node to observe content changes in any other node, without knowing any internal details about that node.

BUG: Test

### DIFF
--- a/Libs/MRML/Core/Testing/vtkMRMLNodeTest1.cxx
+++ b/Libs/MRML/Core/Testing/vtkMRMLNodeTest1.cxx
@@ -100,12 +100,22 @@ public:
   vtkObject* LastMRMLEventCaller;
   unsigned long LastMRMLEventId;
 
+  enum
+  {
+    ContentModifiedEvent1 = 25000,
+    ContentModifiedEvent2 = 25001,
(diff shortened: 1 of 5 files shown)


pieper (approved):
The motivation makes sense.  I didn't follow the details of the implementation but I'm fine with merging this to fix the original issue.

lassoan:
Thanks for the review!

PR #6962, opened 2023-05-04
Mark markups module

idrissa10:
Mark markups module


lassoan (approved):
Thank you, I've made a few adjustments (mostly reverted all the unrelated code formatting changes) and with those it is ready for integration.

PR #6963, opened 2023-05-07
ENH: Support special characters when saving scene to DICOM SC

cpinter:
Exporting scene packaged in MRB as secondary capture (SC) now supports special characters in DICOM metadata, for example patient name.


lassoan (commented):
Thanks, this will be good, but I think it needs a small adjustment.

If `self.optionalTags` contain special characters then they are already encoded with some other default non-UTF-8 encoding, so setting the character set here is too late. Probably it did not show up for you as an error because you only had ASCII strings in the template.

If we add the character set specification to the `dump` string, because then dump2dcm would already write all tags with UTF-8 encoding. Specifying the character set with `-k` would be no longer needed.

cpinter:
Thanks for the review @lassoan ! Do you mean that adding the same thing to the previous dcmtk call? I'm not sure I understand it totally clear. Thanks!

cpinter:
@lassoan can you please answer the above question? I'm not sure what is needed exactly. Thank you!

cpinter:
Thanks! The problem is that the template file is generated with `CreateDICOMSeries` CLI and there is no option for setting character set (SpecificCharacterSet=ISO_IR 192). Why is it problematic to ensure UTF8 on this level? I think this would be a solution to potential problems regardless the proper procedure that users/developers need to take.

cpinter:
@lassoan Have you seen the above comment? Thank you!

lassoan:
CreateDICOMSeries already uses UTF8.

The problem is with the template file creation. The template is created by [dump2dcm](https://github.com/Slicer/Slicer/pull/6963/files#diff-b818e2344fcc0a7f37d60b562150559db3a8eaeedce2faae85a72d5c6f0fd5e1R145). CreateDICOMSeries uses the template file as input.

lassoan:
After your skype message I understand what happens better. You create a custom referenceFile using CreateDICOMSeries and use that instead of the default one (that is generated from DICOM tags using dump2dcm).

What Slicer version are you using? Because in recent Slicer versions, CreateDICOMSeries already sets utf8 character set. I've just printed this from a series exported by CreateDICOMSeries in Slicer-5.3.0-2023-04-25:

    (0008,0005) CS [ISO_IR 192]                             #  10, 1 SpecificCharacterSet

Isn't this sufficient? If you specify additional fields on the command line using `-k`, do you find that you need to set character set again? If it works with a recent CreateDICOMSeries implementation then the change in this pull request may not be needed at all.

Separate topic: When the dump2dcm is used for creating the referenceFile then the file may not be created correctly, as the fields may contain special characters and SpecificCharacterSet is not set yet. To fix this, the SpecificCharacterSet could be set in [dumpFromTags](https://github.com/cpinter/Slicer/blob/daf219af64e701cb7b890bc63ceb20f1973b360f/Modules/Scripted/DICOMLib/DICOMExportScene.py#L199). The fix is probable trivial, but since you don't use this dump2dcm-based referenceFile generation, you don't need this fix if you don't have time.

cpinter:
I can confirm that with the latest Slicer the series created using CreateDICOMSeries contains the character set tag with the same value as you show. However, the exported scene still breaks the accented characters.

To me adding the -k switch seems to be a simple and harmless fix. If you consider it potentially problematic to add this switch to the DICOM export plugin in Slicer, then I'll create a fork for my project because I need to fix this issue. Thank you!

lassoan:
No need to fork for sure, I just wanted to make sure the implementation is robust and contains all necessary changes and not more.

I see that img2dcm app is used here, which by default sets encoding to `lSO IR 100` unless you specify `--no-latin1`. Could you check if setting `--no-latin1` instead of `'-k', f'SpecificCharacterSet=ISO_IR 192'` works? If it does not work then just add a comment there that we don't understand why, but SpecificCharacterSet must be specified again here.

One last question. This is not new behavior, you don't have to change things if it works well for you as is. But it may be something useful for you to test and fix if needed. -- On Windows, passing UTF8 string through the command line can be problematic (unless in your system settings you [set the system code page to UTF8](https://stackoverflow.com/a/57134096)), because the string may be converted to the system code page and the system code page may not be able to represent all special characters. Therefore, I'm not sure if setting UTF8 strings via `-k` will always work. Maybe it does, because maybe conversion only happens if you use the console and not when you launch a process using Python exec API. If you find that you can pass a Japanese character in the series description correctly with `-k` then it is all good (you can add a comment around the `-k` calls that it is all good for UTF8 strings to help future developers). If it does not work then we could add all strings (`seriesDescription` and all `self.optionalTags`) to the input `dump` string rather than pushing through the command line.

cpinter:
Thank you very much! I'll try both the `--no-latin1` switch and passing a Japanese character.

cpinter:
It seems that using `--no-latin-1` solves the issue as well. Like this:

```
        args = [
            '-k', f'SeriesDescription={seriesDescription}',
            '--no-latin1']
```

I tried it with Japanese characters and they seem to be preserved too. Although it puts an extra character in front for some reason:

image

lassoan:
OK, let's go with the `--no-latin1` option then. Add a comment there that with this the UTF8 encoding of the input file is preserved and without this the character set would be forced to Latin1. We could then merge this now.

We could later revisit the passing of UTF8 strings through command line. For now, just add a comment there noting that it could be safer to pass all these fields through the input file instead of using command-line arguments.

cpinter:
Please check the updated commit and let me know if you intended any of the comments differently. Thanks!

lassoan (commented):
This will be good, just fix the lint error.

cpinter:
Sorry @lassoan not getting any error... What needs to be fixed?

Modules/Scripted/DICOMLib/DICOMExportScene.py
```diff
@@ -148,7 +148,9 @@ def createDICOMFileForScene(self):
         # cmd = "img2dcm -k 'SeriesDescription=Slicer Data Bundle' -df %s/template.dcm %s %s" % (saveDirectoryPath, imageFile, self.sdbFile)
         seriesDescription = 'Slicer Data Bundle' if self.seriesDescription is None else str(self.seriesDescription)
         args = [
-            '-k', f'SeriesDescription={seriesDescription}']
+            '-k', f'SeriesDescription={seriesDescription}',
+            '--no-latin1']  # With this option the UTF8 encoding of the reference file is preserved, otherwise the character set would be forced to Latin1
+        #TODO: It could be safer to pass these fields through the reference file instead of using command-line arguments (from character set point of view)
```
jamesobutler, line 153: A linting error https://github.com/Slicer/Slicer/actions/runs/5154914330/jobs/9283929557?pr=6963
```suggestion
        # TODO: It could be safer to pass these fields through the reference file instead of using command-line arguments (from character set point of view)
```

jamesobutler: Also if this is actually future TODO work, maybe write up a GitHub issue to track it instead? In line TODOs seem to get lost in a big project

cpinter: @lassoan suggested adding a comment about this https://github.com/Slicer/Slicer/pull/6963#issuecomment-1572286973

All I want is get this into Slicer. Thanks.

cpinter:
Linting error fixed.

lassoan (approved):

PR #6970, opened 2023-05-13
BUG: Fix IsPluginInstalled() when plugin installed in application build tree

jcfr:
This commit is a follow-up of 008f8b957 (ENH: Make Slicer application fully
portable) and f7edf41e7 (BUG: Fix logic that determines if a module is built-in).

It ensures that in the context of extension installed from file when running the application from a build tree, the function `qSlicerAbstractCoreModule::isInstalled()` returns the expected value.

### DIFF
--- a/Base/Logic/vtkSlicerApplicationLogic.cxx
+++ b/Base/Logic/vtkSlicerApplicationLogic.cxx
@@ -9,7 +9,13 @@
 // Slicer includes
 #include "vtkSlicerApplicationLogic.h"
 #include "vtkMRMLColorLogic.h"
-#include "vtkSlicerConfigure.h" // For Slicer_BUILD_CLI_SUPPORT
+// For:
+//  - Slicer_BUILD_CLI_SUPPORT
+//  - Slicer_BUILD_EXTENSIONMANAGER_SUPPORT
+//  - Slicer_EXTENSIONS_DIRBASENAME
+//  - Slicer_ORGANIZATION_DOMAIN
+//  - Slicer_ORGANIZATION_NAME
+#include "vtkSlicerConfigure.h"
 #include "vtkSlicerTask.h"
 
 // MRML includes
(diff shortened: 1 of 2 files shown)


Base/Logic/vtkSlicerApplicationLogic.cxx
```diff
@@ -737,15 +743,46 @@ bool vtkSlicerApplicationLogic::IsPluginInstalled(const std::string& filePath,
...
-    return !itksys::SystemTools::FileExists(
+    bool isAppBuildTree = itksys::SystemTools::FileExists(
           std::string(applicationHomeDir).append("/CMakeCache.txt").c_str(), true);
+    if (isAppBuildTree)
+      {
+#ifdef Slicer_BUILD_EXTENSIONMANAGER_SUPPORT
+      // Check if plugin is installed as an extension
+      {
+      std::string extensionInstallDir = itksys::SystemTools::GetRealPath(
+            applicationHomeDir + "/" Slicer_ORGANIZATION_DOMAIN "/" Slicer_EXTENSIONS_DIRBASENAME);
```
jcfr, line 761: Instead of checking if the directory is like `<applicationHome>/<org-name-or-domain>/Extensions-`, we would consolidate the approach already used in `vtkSlicerApplicationLogic::IsEmbeddedModule` by updating the signature of `vtkSlicerApplicationLogic::IsPluginInstalled` to accept a revision parameter and instead look for `Extensions-<revision>` in the plugin path.

jcfr:
@lassoan @jamesobutler Would appreciate a cursory check if you have a chance. 

It is ready as it is. 

My comment is more a nice to have / potential future improvements

Base/Logic/vtkSlicerApplicationLogic.cxx
```diff
@@ -737,15 +743,46 @@ bool vtkSlicerApplicationLogic::IsPluginInstalled(const std::string& filePath,
     return false;
     }
 
+  // If the plugin is found in the application build directory (or one of its descendent), it is not
+  // considered to be installed. This is generally true unless the plugin is found in the extension
```
lassoan, line 747: Do we have somewhere a definition of "installed" and "built-in" module?

lassoan (approved):
This looks good to me.

Base/Logic/vtkSlicerApplicationLogic.cxx
jcfr: > not installed = in build tree ?

Since extension can be installed in the build tree in the "special" folder `<applicationHome>/<org-name-or-domain>/Extensions-12345`, saying this is not exactly true.

> not built-in = provided by extension?

Not exactly correct either as module can be specific using option like `--additional-module-path/--additional-module-paths` and in that care they are not installed using the extension manager.

----

I create a "mermaid" diagram to include in our documentation

Reference:
* https://github.blog/2022-02-14-include-diagrams-markdown-files-mermaid/
* https://sphinxcontrib-mermaid-demo.readthedocs.io/en/latest/

lassoan: I guess what I miss is a very high-level description of what installed and built-in flags are meant to distinguish between.

For example, built-in could refer to "not being manually installed by the user" or "source code is in Slicer source tree". Both could make sense, both could be useful to know.

Issue #6972, opened 2023-05-15
Correctly handle non-modified terminology type properties


cpinter:
## Summary

If selecting "No type modifier" for a terminology type that has modifiers, then the non-modified properties (e.g. color) should be correctly used instead of leaving them undefined/invalid.

See details https://discourse.slicer.org/t/terminologies-questions/27384/20

## Steps to reproduce

First, need to edit the terminology json to specify a color for the "non-modified" type. For example add this
```
            "recommendedDisplayRGBValue": [
              203,
              179,
              77
            ],
```
above this line
https://github.com/Slicer/Slicer/blob/main/Modules/Loadable/Terminologies/Resources/SegmentationCategoryTypeModifier-SlicerGeneralAnatomy.json#L1554
(the file can be found in Windows Slicer installations here: `c:\Users\[UserName]\AppData\Local\slicer.org\Slicer 5.3.0-2023-05-14\share\Slicer-5.3\qt-loadable-modules\Terminologies\SegmentationCategoryTypeModifier-SlicerGeneralAnatomy.term.json`)

* Load TinyPatient sample data (need to have Developer mode enabled)
* Go to Data module
* Double-click color of the `Body_Contour` segment
* Select `Frontal bone` terminology type

## Expected behavior

Color should be yellow instead of the undefined gray.

## Environment
- Slicer version: Slicer 5.3.0-2023-05-14
- Operating system: Windows 11

smrolfe:
I created a [pull request](https://github.com/Slicer/Slicer/pull/6973) with my suggested solution. Assigning a type with no modifiers the color of the first modifier would be consistent with the assignment of the color box icon in the Property types viewer.

cpinter:
Thank you very much @smrolfe (and sorry for the long process)! Unless I'm missing something this can be closed.

PR #6973, opened 2023-05-16
BUG: Update Terminologies default color behavior

smrolfe:
This commit changes the default color assignment for Terminologies with modifiers. 

Currently if there is no modifier selected, the type is assigned the "invalid color", as discussed in this [forum post](https://discourse.slicer.org/t/terminologies-questions/27384/28) and [issue](https://github.com/Slicer/Slicer/issues/6972). This commit assigns the color of the first modifier to the type. This is the heuristic that is followed for the assignment of the color box displayed in the property type viewer.


lassoan:
@smrolfe thanks a lot for working on this. The behavior should be to use the color, name, and terminology of the non-modified term. @cpinter what do you think?

smrolfe:
@lassoan this would be best, but in the examples I've looked at, the non-modified term does not have a color specified: https://discourse.slicer.org/t/terminologies-questions/27384/24?u=smrolfe

muratmaga:
> @smrolfe thanks a lot for working on this. The behavior should be to use the color, name, and terminology of the non-modified term. @cpinter what do you think?

@lassoan we can decorate our custom jsons with RGB properties for non-modified state. None of the the Slicer's bundled JSON files do that AFAIK. For our use case, L/R, L and R (or unmodified state) are going to have same color. So it is really up to you guys how to handle it. But even if you do add the RGBs, the bug is still there (see the issue here https://github.com/Slicer/Slicer/issues/6972).

cpinter (commented):
Can you please describe the change in more detail in the commit message? I may be missing something but I don't understands what this fixes exactly and how. Thanks!

For consistency I'd use `this->CurrentTypeObject->GetRecommendedDisplayRGBValue` instead of `this->recommendedColorForType`, it will make the code more readable.

Also, invalid values need to be checked in every case.

Thank you!

smrolfe:
> For consistency I'd use `this->CurrentTypeObject->GetRecommendedDisplayRGBValue` instead of `this->recommendedColorForType`, it will make the code more readable.

`this->recommendedColorForType` is an existing function that checks the colors assigned to the modifiers if no color was found for the type, and returns the first one found. It's used to create the color preview when the property type table is populated. 

I will update the pull request shortly with your suggestions, thanks @cpinter!

smrolfe:
@lassoan @cpinter @muratmaga 

This pull request is ready for review.

PR #6977, opened 2023-05-16
BUG: Mark translatable strings in sequences module

idrissa10:


lassoan (approved):
Thank you!
I've made a few minor fixes. With those it is good to go.

PR #6984, opened 2023-05-26
ENH: Add Pickable segment display property to allow disable point pic…

cpinter:
…king

### DIFF
--- a/Libs/MRML/Core/vtkMRMLSegmentationDisplayNode.h
+++ b/Libs/MRML/Core/vtkMRMLSegmentationDisplayNode.h
@@ -58,6 +58,8 @@ public:
     double Opacity3D{1.0};
     double Opacity2DFill{1.0}; // This one is used for labelmap volume related operations (color table, merged labelmap)
     double Opacity2DOutline{1.0};
+    /// Pickable by interactions flag. If true, then the point under the mouse on the segment surfaces in the 3D views is computed, otherwise skipped.
+    bool Pickable{true};
 
     // Initialize with default values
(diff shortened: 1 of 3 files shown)


cpinter:
https://github.com/Slicer/Slicer/assets/1325980/ac32aa1b-bda8-42aa-b04b-81bc15a52a6b

Libs/MRML/Core/vtkMRMLSegmentationDisplayNode.h
```diff
@@ -58,6 +58,8 @@ class VTK_MRML_EXPORT vtkMRMLSegmentationDisplayNode : public vtkMRMLDisplayNode
     double Opacity3D{1.0};
     double Opacity2DFill{1.0}; // This one is used for labelmap volume related operations (color table, merged labelmap)
     double Opacity2DOutline{1.0};
+    /// Pickable by interactions
```
lassoan, line 61: This is not very clear. I've kind of almost already forgotten what exactly we wanted to use this for and how it is related to the `Selectable` node property, and this short comment did not help me to understand what this flag is for. Maybe also mention that it is only for 3D views.

lassoan (commented):
This is very nice and clear. I only had a small comment on improving the documentation, because the subtle difference in behavior difference it may not be obvious (many developer may not even know what "picking" refers to).

cpinter:
Thank you for the review! I added more information in the comment about the Pickable property. I hope this explains it well enough.

lassoan (approved):
Thank you, it looks good.

PR #6991, opened 2023-05-30
BUG: Update vtkCacheManager.cxx

maddin200:
remove unused variable

### DIFF
--- a/Libs/MRML/Core/vtkCacheManager.cxx
+++ b/Libs/MRML/Core/vtkCacheManager.cxx
@@ -219,7 +219,6 @@ int vtkCacheManager::LocalFileExists ( const char *uri )
 {
   int index;
   std::string uriString (uri);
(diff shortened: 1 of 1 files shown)


lassoan (approved):
Thank you for working on this. The change looks good to me.

However, now that we touch this method, would you have time to make this method a bit nicer?

This method functionally looks mostly OK, but has many style issues, such as index is not initialized when defined, then later it is set in an `if` statement; uri is not checked for nullptr; there i sa comment `is this the correct index???` that indicates that the developer had limited understanding; there are C-style casts to `(int)` that we should be able to do without; inconsistent spacing around parentheses; returnin simple values in parantheses; etc.

lassoan:
You need to fix the commit message (append the prefix `BUG: ` or `STYLE: `). You can also push fixes for all your other code findings into this branch.

PR #6953, opened 2023-04-20
Add support for registering scripted IOOptionsWidget

jcfr:

PR #6993, opened 2023-05-30
Misc scripted module fixes

jcfr:
This pull-request aims to integrate fixes & changes implemented while working toward finalizing:
* https://github.com/Slicer/Slicer/pull/6953

## Summary

* Add support for checking that scripted IOs are registered
* Add support for checking that scripted file dialogs are registered
* Ensure python exceptions in `ScriptedModuleDiscoveryTest` are considered
* Improve comments in `ScriptedModuleDiscoveryTest.py`
* Fix setting of extension module paths when `slicer_add_python` testing macros
* Fix setting module paths associated with `slicer_add_python` testing macros

### Improvements to `slicer_add_python`

* Ensures that arguments like "--disable-modules" and "--disable-*-modules" are respected when adding tests using the slicer_add_python testing macros.

### Ensure python exceptions in `ScriptedModuleDiscoveryTest` are considered

Workaround the fact using PASS_REGULAR_EXPRESSION ignores return code by checking for output characteristics of Python exceptions.
    
This is needed because the message "ModuleAWidget finalized" is displayed early on before potential python exceptions.
    
Note this is a partial workaround as non-python errors would not cause the test to fail.
    
See https://github.com/jcfr/cmake-add_test-PASS_REGULAR_EXPRESSION

### DIFF
--- a/Applications/SlicerApp/Testing/Python/CMakeLists.txt
+++ b/Applications/SlicerApp/Testing/Python/CMakeLists.txt
@@ -198,6 +198,17 @@ slicer_add_python_test(
 if(UNIX)
 set_tests_properties(py_nomainwindow_${testname} PROPERTIES
   PASS_REGULAR_EXPRESSION "ModuleAWidget finalized"
+  # Workaround the fact using PASS_REGULAR_EXPRESSION ignores return code
+  # by checking for output characteristics of Python exceptions.
+  #
+  # This is needed because the message "ModuleAWidget finalized" is displayed
+  # early on before potential python exceptions.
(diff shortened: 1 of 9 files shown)


lassoan (approved):
Thank you, the changes look good to me.

jamesobutler:
@jcfr Thank you for your continued work on this. Please ping me when you’d like me to test your implementation at https://github.com/Slicer/Slicer/pull/6953. I see you have a little bit of WIP work still there. I’ll provide review to confirm that it works for my use case. Thanks again!

PR #6995, opened 2023-05-31
ENH: Add VolView link to WebServer examples

pieper:


Modules/Scripted/WebServer/Resources/docroot/index.html
```diff
@@ -152,6 +152,21 @@ <h3>modelviewer example</h3>
     </li>
   </ul>
 
+  <ul>
+    <li>
+      <h3>VolView examples</h3>
+      <p>Load a volume in Slicer to see it in VolView.</p>
+      <p>Be sure the slicer api and cors are enabled.</p>
+      <p><a href='https://kitware.github.io/volview/'>volview from kitware</a> can load data from slicer directly using a <a href='https://kitware.github.io/volview/doc/loading_data.html#loading-remote-data-via-urls'>url parameter<a>.</p>
+      <a href='https://volview.netlify.app/?names=%5bmrhead.nrrd%5d&urls=%5bhttp://localhost:2016/slicer/volume%5d'>https://volview.netlify.app/?names=%5bmrhead.nrrd%5d&urls=%5bhttp://localhost:2016/slicer/volume%5d</a><br>
```
lassoan, line 161: `slicer/volume` returns the currently loaded volume, so instead of using `mrhead.nrrd` as name we should use something more generic

pieper: Good point!

lassoan (approved):
It looks good to me.

lassoan:
Could we set up the Slicer web server to allow VolView hosted on netlify to connect to Slicer without enabling CORS (if we enable CORS then we open up Slicer to any apps on any website)?

pieper:
I was thinking we could examine incoming requests and approve CORS on a site-by-site basis.  I.e. we could maintain an approved list of https domains in the settings and if a CORS request came in from a non-approved site we could prompt the user to either add the site to an accepts list, a reject list, or get prompted each time there's a connection.

Something similar could be done for the exec API since it's an extra level of security concern.

PR #6997, opened 2023-05-31
ENH: One draw effect point does not prevent use in other slice offset

jamesobutler:
If you are in the "Draw" segment editor effect and click to place one point in the given slice offset and then you scroll to another slice offset, you will be unable to place more draw effect points in that offset. It is confusing because that single point does not create an visual UI element, so you are left confused that you actually had started drawing on another slice offset. You currently have to right-click to apply (which does nothing because there is <2 points) and then you can draw on the new slice offset. 

This was leading to confusion by our group's users. I have added code so that if you have placed a single point and then scroll to another slice offset, then the draw effect's points are cleared/reset. 

I still think it is a little confusing when there are >1 points and the draw effect goes into a "dashed" line state which is difficult to see. It currently looks like a graphical issue. I would rather the draw effect lines stay the same.

### DIFF
--- a/Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorEffects/SegmentEditorDrawEffect.py
+++ b/Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorEffects/SegmentEditorDrawEffect.py
@@ -137,6 +137,10 @@ class SegmentEditorDrawEffect(AbstractScriptedSegmentEditorLabelEffect):
             if pipeline.activeSliceOffset:
                 offset = abs(currentSliceOffset - pipeline.activeSliceOffset)
                 if offset > 0.01:
+                    if pipeline.rasPoints.GetNumberOfPoints() == 1:
+                        # One placed point is not visible to the user so clear the state upon changing slice offset
+                        pipeline.resetPolyData()
+                        return
                     lineMode = "dashed"
             pipeline.setLineMode(lineMode)
             pipeline.positionActors()


jamesobutler:
> I still think it is a little confusing when there are >1 points and the draw effect goes into a "dashed" line state which is difficult to see. It currently looks like a graphical issue. I would rather the draw effect lines stay the same.

Below is the code being referenced in regards to the display of the draw effect after changing slice offset.
https://github.com/Slicer/Slicer/blob/5e7d60c87711a14902f9001330321e1e494a7e6b/Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorDrawEffect.py#L123-L132

jamesobutler:
The dashed line doesn't really work when using the draw tool in a continuous point placing mode (left+click-and-drag). As shown below it very much looks like a rendering issue. 

image
image

The dashed rendering is better when placing individual points where there are longer distances between points.
image

lassoan:
Allowing users to peek into neighbor slices during drawing is a very useful feature, but I agree that the dashed line appearance is not clear enough (and when the line segments are too short - for example due to continuous drawing). @jamesobutler do you have an idea how we could preserve this feature but make it more clear to users how it works?

jamesobutler:
> Allowing users to peek into neighbor slices during drawing is a very useful feature

I think it would be logical to support drawing on the 2D viewport and allow scrolling to other slice offsets and maintain points (if n>1). Applying the end interaction however would apply to the current offset. Scrolling to another slice and then wanting to apply would seem confusing to a user that it isn’t applying at their current slice.

Otherwise changing the color of the draw lines instead of changing to dashed could indicate it was drawn on a non-active slice offset.

pieper:
Another option would be to change the cursor to something like an arrow with an X when you scroll away from the slice where the drawing will take place.

jamesobutler:
So currently if you place some points with the draw effect and then scroll to another slice offset, the display changes to dashed lines. From that new slice offset you cannot draw new points, but you can apply the drawing and it will apply to that new slice offset, not the original slice offset. 

Therefore what is the meaning of changing the draw lines to dashed and preventing placing of new points if you are allowed to apply the dashed drawing to the new slice offset? Using a forbidden cursor when changing the slice offset would seem like the desire is to prevent applying the drawing on an offset where it was not originally drawn on? Is that a bug?

pieper:
> but you can apply the drawing and it will apply to that new slice offset, not the original slice offset.

Yes, this seems like a bug to me.  I would expect it to apply to the original slice.

jamesobutler:
Is it reasonable to clear the points if only 1 is placed and the slice offset is changed (`ENH: One draw effect point does not prevent use in other slice offset` in this PR)? Essentially defining that the drawing is invalid and not locking it to a slice offset. 

Or should the individual point be maintained, but don't prevent placing new points on a new slice offset? This would bring that first point to the new offset where the second point was placed and locking it in there.

jamesobutler:
Is this commit in this PR good to integrate? We have had customers specifically run into this issue of placing a point in a slice view, then changing slice offset being confused why the "draw effect was broken".

pieper (changes requested):
Were you able to test this? I'm thinking that resetting the polydata should only happen if you are not on the active slice (i.e. when `if offset > 0.01:` is true).  Currently it looks like it would always reset a single point when a slice node event comes in, such as panning or something.

jamesobutler:
Yes I tested this and demo'd it working during the hangout meeting. I'm not sure if you had joined the meeting at that time.

pieper:
Okay, then I think just moving the code down into that if clause is all we need.

pieper (approved):
LGTM 👍

PR #6998, opened 2023-06-02
Choose file reader writer based on confidence

lassoan:
Until now, reader and writer plugins could only return a Boolean value indicating if they can read or write a certain file. For file reading, composite file extensions (.seg.nrrd instad of just .nrrd) were used to find the most suitable plugin, but this did not work when a file was saved with a generic file extension (nrrd, json, txt, ...). It also caused an issue when a more specific reader plugin was registered before a generic plugin (e.g., segmentation reader was registered before volume reader). Finally, there was no way to prioritize writer plugins.

This commit introduces a new `canLoadFileConfidence` method to readers and writers. This allows a plugin to indicate its confidence that it is the most suitable plugin for reading/writing a file. Reader plugins are updated to determine confidence based on a quick inspection of file content when a file with generic file extension is encountered. Quick inspection means that the header is not parsed, just a few hundred bytes are read and a string is searched in it.

### DIFF
--- a/Base/QTCore/qSlicerFileReader.cxx
+++ b/Base/QTCore/qSlicerFileReader.cxx
@@ -58,6 +58,22 @@ bool qSlicerFileReader::canLoadFile(const QString& fileName)const
 }
 
 //----------------------------------------------------------------------------
+double qSlicerFileReader::canLoadFileConfidence(const QString& fileName)const
+{
+  if (!this->canLoadFile(fileName))
+    {
+    return 0.0;
+    }
+  int longestExtensionMatch = 0;
+  QStringList res = this->supportedNameFilters(fileName, &longestExtensionMatch);
+  // If longer extension is matched then the confidence that this is a good reader is
+  // slightly higher. For example, for "somefile.seg.nrrd", a reader that is specifically
+  // for ".seg.nrrd" files get slightly higher confidence than readers that of generic ".nrrd" files.
+  double confidence = 0.5 + 0.01 * longestExtensionMatch;
+  return confidence;
+}
+
+//----------------------------------------------------------------------------
 QStringList qSlicerFileReader::supportedNameFilters(const QString& fileName, int* longestExtensionMatchPtr /* =nullptr */)const
 {
   if (longestExtensionMatchPtr)
--- a/Base/Python/slicer/tests/test_slicer_util_save.py
+++ b/Base/Python/slicer/tests/test_slicer_util_save.py
@@ -15,10 +15,118 @@ class SlicerUtilSaveTests(unittest.TestCase):
         shutil.rmtree(slicer.app.temporaryPath + '/SlicerUtilSaveTests', True)
 
     def test_saveNode(self):
-        node = slicer.util.getNode('MR-head')
+        """Test that nodes are saved correctly and that they are loaded correctly with the default reader
+        even if they are saved with generic (.nrrd) and not composite (.seg.nrrd) file extension."""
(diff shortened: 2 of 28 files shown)


cpinter:
This will be great! Thank you Andras!

Base/QTCore/qSlicerFileReader.cxx
```diff
@@ -57,6 +57,22 @@ bool qSlicerFileReader::canLoadFile(const QString& fileName)const
...
+  if (!this->canLoadFile(fileName))
+    {
+    return 0.0;
+    }
+  int longestExtensionMatch = 0;
+  QStringList res = this->supportedNameFilters(fileName, &longestExtensionMatch);
+  // If longer extension is matched then the confidence that this is a good reader is
+  // slightly higher. For example, for "somefile.seg.nrrd", a reader that is specifically
+  // for ".seg.nrrd" files get slightly higher confidence than readers that of generic ".nrrd" files.
+  double confidence = 0.5 + 0.01 * longestExtensionMatch;
```
cpinter, line 72: I like this idea :)

jamesobutler:
I have a custom reader for a directory of files (I did not create this storage system):
- SequenceName
  - Seq_Frame_0000
    - ImageInfo.txt
    - Image.tiff
  - Seq_Frame_0001
    - ImageInfo.txt
    - Image.tiff
  - Seq_Frame_0002
    - ImageInfo.txt
    - Image.tiff
  - Seq_Frame_0003
    - ImageInfo.txt
    - Image.tiff
  - SequenceInfo.txt

where the SequenceInfo.txt is acting like a Slicer mrml file in that it points to other image files. The SequenceInfo.txt has details about loading the individual frames within the directory as a sequence. From your description I could use quick inspection of the SequenceInfo.txt, which has a generic file extension, to know how to actually read that file for loading. Currently I have had to rely on the name "SequenceInfo.txt" for identifying that it should use the custom reader. The confidence could help if a user renames this file as it could still read the content for figuring it out.

However if I choose "Add Directory" in the add data dialog and select the "SequenceName" directory level (or drag-and-drop this directory into slice views), would there be a confidence method where it detects that a SequenceInfo.txt is present in the directory and only adds that file to the data dialog instead of also adding the individual ImageInfo.txt files and Image.tiff files as well? This is more about using confidence of a file for exclusion of adding other files for individual loading.

lassoan:
> The SequenceInfo.txt has details about loading the individual frames within the directory as a sequence. From your description I could use quick inspection of the SequenceInfo.txt, which has a generic file extension, to know how to actually read that file for loading. 

Yes, this is indeed a good application of the new mechanism.

> However if I choose "Add Directory" in the add data dialog and select the "SequenceName" directory level (or drag-and-drop this directory into slice views), would there be a confidence method where it detects that a SequenceInfo.txt is present in the directory and only adds that file to the data dialog instead of also adding the individual ImageInfo.txt files and Image.tiff files as well?

"Choose directory to add" in "Add data" window adds all files in the folder, so it will be inspected file by file. You could have a raeder that detects all the .txt and .tiff files, claim them with high confidence (preventing other readers from reading them), and when reading is requested then it would just store their filenames. When the "Add data" window is closed, your reader could reactivate itself and inspect all the stored filenames and determine how to load them.

I plan to implement a mechanism like this for loading DICOM images via the "Add data" window. The new "DICOM" reader would claim all DICOM files with high confidence (preventing ITK readers from processing it) and then import and load them using the DICOM module.

PR #7000, opened 2023-06-04
BUG: Set segment Pickable display property to true if not specified

cpinter:
Pickable is a new segment display property, so when loading existing segmentations it will be missing from the display node. In this case the default true value needs to be specified, otherwise no segments will be pickable (i.e. no fiducials or other points can be placed on them).

This commit also fixes saving of the Pickable property. Now it is saved and loaded as a boolean property, as "true" or "false" instead of 0 or 1.


cpinter:
This PR fixes a critical bug. Without this fix, no fiducials can be placed on segmentations loaded from file, because by default the Pickable property will be false for all segments if not specified.

lassoan (approved):
Looks good, thank you.

PR #7001, opened 2023-06-05
COMP: Speed up incremental build

achataigner:
vtkSlicerVersionConfigureInternal.h is renamed vtkSlicerVersionConfigureMinimal.h.

#include "vtkSlicerVersionConfigure.h" is replaced by #include "vtkSlicerVersionConfigureMinimal.h" when it is possible, therefore no compilation is required if git commit is changed.

### DIFF
--- a/Base/Logic/Testing/vtkSlicerVersionConfigureTest1.cxx
+++ b/Base/Logic/Testing/vtkSlicerVersionConfigureTest1.cxx
@@ -34,7 +34,7 @@ int vtkSlicerVersionConfigureTest1(int /*argc*/, char * /*argv*/ [])
   CHECK_STRING_DIFFERENT(Slicer_WC_REVISION, "");
   CHECK_STRING_DIFFERENT(Slicer_REVISION, "");
 
-  // From vtkSlicerVersionConfigureInternal
+  // From vtkSlicerVersionConfigureMinimal
   CHECK_STRING(Slicer_OS_LINUX_NAME, "linux");
   CHECK_STRING(Slicer_OS_MAC_NAME, "macosx");
   CHECK_STRING(Slicer_OS_WIN_NAME, "win");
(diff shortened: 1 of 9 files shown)


achataigner:
Hi all, could I have some feedback on this branch ?

achataigner:
Thanks a lot for your feedback, I will apply these changes tomorrow.

achataigner:
Done

jcfr (commented):
The "internal" was used to indicate the file was not intended to be directly included.

It was in it own file for convenience so that it could be re-configured at build time [^1]

What about calling the file  `vtkSlicerVersionConfigure.h.in`.

[^1]: https://github.com/Slicer/Slicer/blob/main/CMake/SlicerConfigureVersionHeaderTarget.cmake

lassoan:
The idea is to split version configuration file to two parts:
- Minimal: contains version major/minor/patch, whith changes only when a new release is created. 
- Full: contains git hash, revision, etc. that changes with each commit (in addition to the minimal version info)

At most places we only need information from "Minimal". At all those places it makes sense to include only that, because it is rarely changing. If we include the "Full" version information everywhere then more stuff are rebuilt when there is  a new commit.

Since we already had a "...Internal" configuration, it made sense to reuse that. The rename was necessary to reflect that it is not an internal implementation detail anymore but it is part of the public API that modules should use when the minimal version information is sufficient.

achataigner:
Thank you Andras for your explanations.

jcfr:
Before doing so, would it possible to split  into two commits:
- first one renaming the file and updating its header (remove the warning + include description)
- second one doing the speed optimization by changing what is included.

Would you have time to help with this as well ?

achataigner:
Done

jcfr:
Thanks @achataigner :pray: 

I tweaked the wording to provide additional guidance.

jcfr:
Next, we should also update https://github.com/Slicer/Slicer/blob/main/CMake/SlicerConfigureVersionHeaderTarget.cmake

jcfr (approved):

achataigner:
Thanks a lot

PR #7002, opened 2023-06-06
ENH: Add fractional addition to Add/Subtract Slice Composting pipeline

sjh26:
- the Add/Subtract pipeline now respects the opacity slider used for alpha blend
- The opacity slider is used to determine the fraction of the foreground image added to the background

Alpha Blend Behavior:
AlphaBlendPipeline

| Old Add Behavior | New Add Behavior |
|--|--|
| OldAddPipeline_mid | NewAddPipeline |


jcfr (approved):

pieper (approved):
Makes sense to me 👍

PR #7014, opened 2023-06-10
Improve Python FAQ

jcfr:
This pull request adds the following entries:
* What is the Slicer Python Environment?
* What is the PythonSlicer executable?
* What is the Python Console?
* How to access Slicers Python API via an external program while Slicer is running?

Additionally, it update the Slicer API to document the following:
* `slicer.app`
* `slicer.mrmlScene`
* `slicer.modules`
* `slicer.moduleNames`

Summary of this pull request also posted on discourse at https://discourse.slicer.org/t/updated-python-faq-slicer-api-documentation/29967

<details><summary>Click here for preview screenshots</summary>

| | |
|--|--|
| Slicer API | image |
| Python FAQ | image |
|   | image |

</details>

### DIFF
--- a/Base/Python/slicer/__init__.py
+++ b/Base/Python/slicer/__init__.py
@@ -1,4 +1,129 @@
-""" This module sets up root logging and loads the Slicer library modules into its namespace."""
+""" This module sets up root logging and loads the Slicer library modules into its namespace.
+
+
+.. warning::
+
+  These following attributes are only set in the Python environment embedded
+  in the Slicer main application ``SlicerApp-real`` (launched by the Slicer executable):
+
+  * :attr:`app`
+  * :attr:`mrmlScene`
+  * :attr:`modules`
+  * :attr:`moduleNames`
+
+  This means that they are not set in the Python environment of ``python-real``
+  (launched by the ``PythonSlicer`` executable).
+
+.. py:attribute:: app
+
+  This is set to the singleton instance of ``qSlicerApplication``::
+
+    >>> slicer.app
+    qSlicerApplication (qSlicerApplication at: 0x7ffd066a07a0)
+
+  .. warning::
+
+    This attribute is only set in the Python environment embedded in the Slicer
+    main application.
+
+
+.. py:attribute:: mrmlScene
+
+  This is set to the instance of the MRML Scene::
+
+      >>> slicer.mrmlScene
+      <MRMLCore.vtkMRMLScene(0x2797cb0) at 0x7fbfedb82520>
+
+      >>> slicer.mrmlScene == slicer.app.mrmlScene()
+      True
+
+  .. warning::
+
+    This attribute is only set in the Python environment embedded in the Slicer
+    main application.
+
+
+.. py:attribute:: modules
+
+  This object provides access to all instantiated Slicer modules.
+
+  For each instantiated Slicer module, an attribute named after the lower-cased
+  Slicer module name (``slicer.module.<modulename>``) is associated with the
+  corresponding instance of ``qSlicerAbstractCoreModule``.
+
+  For example::
+
+    >>> slicer.modules.volumes
+    qSlicerVolumesModule (qSlicerVolumesModule at: 0x4869000)
+
+    >>> slicer.modules.volumes.inherits("qSlicerAbstractModule")
+    True
+
+    >>> slicer.modules.volumes.inherits("qSlicerLoadableModule")
+    True
+
+  .. warning::
+
+    These attributes are only set in the Python environment embedded in the Slicer
+    main application.
+
+  Additionally for all scripted modules (`qSlicerScriptedLoadableModule`), these
+  additional attributes are also set:
+
+  * the attribute ``<moduleName>Instance`` is set to the corresponding instance
+    of :class:`~slicer.ScriptedLoadableModule.ScriptedLoadableModule`.
+
+    For example::
+
+      >>> slicer.modules.VectorToScalarVolumeInstance
+      <VectorToScalarVolume.VectorToScalarVolume object at 0x7fbfc1d1aa60>
+
+      >>> isinstance(slicer.modules.VectorToScalarVolumeInstance, slicer.ScriptedLoadableModule.ScriptedLoadableModule)
+      True
+
+  * the attribute ``<moduleName>Widget`` is set to the corresponding instance
+    of :class:`~slicer.ScriptedLoadableModule.ScriptedLoadableModuleWidget`.
+
+    For example::
+
+      >>> slicer.util.selectModule(slicer.modules.vectortoscalarvolume)
+
+      >>> slicer.modules.VectorToScalarVolumeWidget
+      <VectorToScalarVolume.VectorToScalarVolumeWidget object at 0x7fbfbc055cd0>
+
+      >>> isinstance(slicer.modules.VectorToScalarVolumeWidget, slicer.ScriptedLoadableModule.ScriptedLoadableModuleWidget)
+      True
+
+    .. note::
+
+      The ``<moduleName>Widget`` attribute is:
+
+      * Only set after the module has been displayed at least once.
+
+      * Updated when reloading the module using :func:`slicer.util.reloadScriptedModule`.
+
+
+.. py:attribute:: moduleNames
+
+  This object provides access to all instantiated Slicer module names.
+
+  The object attributes are the Slicer modules names, the associated
+  value is the module name.
+
+  For example:
+
+    >>> slicer.moduleNames.Volumes
+    'Volumes'
+
+    >>> slicer.moduleNames.VectorToScalarVolume
+    'VectorToScalarVolume'
+
+  .. warning::
+
+    These attributes are only set in the Python environment embedded in the Slicer
+    main application.
+
+"""
 
 
 # -----------------------------------------------------------------------------
@@ -17,17 +142,15 @@ def _createModule(name, globals, docstring):
 # Create slicer.modules and slicer.moduleNames
 
 _createModule('slicer.modules', globals(),
-              """This module provides an access to all instantiated Slicer modules.
+              """This object provides an access to all instantiated Slicer modules.
 
-The module attributes are the lower-cased Slicer module names, the
-associated value is an instance of ``qSlicerAbstractCoreModule``.
+For more details, see the generated Slicer API documentation.
 """)
 
 _createModule('slicer.moduleNames', globals(),
-              """This module provides an access to all instantiated Slicer module names.
+              """This object provides an access to all instantiated Slicer module names.
 
-The module attributes are the Slicer modules names, the associated
-value is the module name.
+For more details, see the generated Slicer API documentation.
 """)
 
 # -----------------------------------------------------------------------------
--- a/Docs/developer_guide/python_faq.md
+++ b/Docs/developer_guide/python_faq.md
@@ -2,7 +2,63 @@
 
 Frequently asked questions about how to write Python scripts for Slicer.
 
-## How to access Slicers Python API via an external program while Slicer is running?
+## What is the Slicer Python environment?
+
+You can consider each Slicer installation as a virtual Python environment - the same way
+as you create virtual environments using python or conda.
+
+## What is the PythonSlicer executable?
+
+`PythonSlicer` is an executable provided in the `bin` directory of the Slicer installation.
+It is a Python interpreter that allows access to all of the packages installed in Slicer.
+This makes it possible to use the Slicer Python environment from outside the application, such
+as for batch processing, command-line operations, terminal-based interactive session and
+integration with IDEs.
+
+This means you can use `PythonSlicer` as a replacement for the regular Python interpreter
+(`python` or `python3`) to take advantage of the installed packages in the Slicer environment.
+
+:::{warning}
+Please note that the running Python context is **not** available when using `PythonSlicer`.
+As a result, many objects, including the application instance, the MRML scene and the loaded
+modules, are not available.
+:::
+
+:::{tip}
+To install additional packages, you can use the {func}`slicer.util.pip_install()` function.
+:::
+
+## What is the Python Console?
+
+This describes the [](/user_guide/user_interface.md#python-console) available through
+the [user interface](/user_guide/user_interface.md).
+
+It allows to access the running Python context including the application instance,
+the MRML scene and all the loaded modules through these :mod:`slicer` attributes:
+* {attr}`slicer.app`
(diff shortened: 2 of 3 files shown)


Docs/developer_guide/python_faq.md
```diff
@@ -2,6 +2,85 @@
...
+## What is the Slicer Python Environment?
+
+You can consider each Slicer installation as a virtual Python environment - the same way
+as you create virtual environments using python or conda.
+
+## What is the PythonSlicer executable?
+
+`PythonSlicer` is an executable provided in the `bin` directory of the Slicer installation.
+It’s the same version of Python available through the Slicer application, but it acts like
+a normal `python` (or `python3`) executable in terms of running scripts and can be used t
```
jamesobutler, line 14:
```suggestion
a normal `python` (or `python3`) executable in terms of running scripts and can be used to
```

Docs/developer_guide/python_faq.md
```diff
@@ -2,6 +2,85 @@
...
+You may use {func}`slicer.util.pip_install()` to install additional packages.
+:::
+
+## What is the Python Console?
+
+This describes the [](/user_guide/user_interface.md#python-console) available through
+the [user interface](/user_guide/user_interface.md).
+
+
+It allows to access to the running application context including the application instance,
```
jamesobutler, line 35:
```suggestion
It allows to access the running application context including the application instance,
```

Docs/developer_guide/python_faq.md
```diff
@@ -2,6 +2,85 @@
...
+- [SlicerWeb](/user_guide/modules/webserver.md): Exposes Slicer's Python environment as a web service
+  that can respond to http(s) requests with data from the current application state or modify the
+  application state. This is well-suited for applications that already use web requests.
+
+- [SlicerOpenIGTLink](https://github.com/openigtlink/SlicerOpenIGTLink): A lightweight socket-based
+  protocol for real-time data transfer. This is useful for applications that need to send many
+  requests per second (e.g., continuous data streaming) or for clients that only have access to
+  sockets and don't want to reimplement complex protocols like http. In most cases, it performs well
+  for sending requests in the range of 10 to 100 requests per second.
+  There is a native Python implementation: [pyigtl](https://pypi.org/project/pyigtl/).
```
jamesobutler, line 69: Can you describe if you would use `pyigtl` in addition to SlicerOpenIGTLink or as an alternative to SlicerOpenIGTLink? Any reason for picking one over the other?  Maybe provide info regarding if SlicerOpenIGTLink can be manipulated through python as well even though OpenIGTLinkIF is originally written in C++.

Docs/developer_guide/python_faq.md
```diff
@@ -2,6 +2,85 @@
...
+
+It allows to access to the running application context including the application instance,
+the MRML scene and all the loaded modules through these :mod:`slicer` attributes:
+* {attr}`slicer.app`
+* {attr}`slicer.mrmlScene`
+* {attr}`slicer.modules`
+* {attr}`slicer.moduleNames`
+
+:::{hint}
+Running scripts (or code) using the command-line option `--python-script` (or `pythoin-code`) is equivalent
```
jamesobutler, line 43:
```suggestion
Running scripts (or code) using the command-line option `--python-script` (or `python-code`) is equivalent
```

Base/Python/slicer/__init__.py
```diff
@@ -1,4 +1,108 @@
...
+.. py:attribute:: app
+
+  This is set to the singleton instance of ``qSlicerApplication``::
+
+    >>> slicer.app
+    qSlicerApplication (qSlicerApplication at: 0x7ffd066a07a0)
+
+  .. warning::
+
+    This attribute is set only in the context of a running Slicer application.
```
lassoan, line 13: It is not very clear what we mean by "in the context of a running Slicer application". Something like this may be better:
```suggestion
    This attribute is only set in the Python environment embedded in the Slicer main application SlicerApp-real (launched by Slicer executable). The attribute is not set in the Python environment of python-real (launched by PythonSlicer executable).
```

Base/Python/slicer/__init__.py
```diff
@@ -1,4 +1,108 @@
...
+
+      >>> slicer.mrmlScene
+      <MRMLCore.vtkMRMLScene(0x2797cb0) at 0x7fbfedb82520>
+
+      >>> slicer.mrmlScene == slicer.app.mrmlScene()
+      True
+
+  .. warning::
+
+    This attribute is set only in the context of a running Slicer application.
```
lassoan, line 28: Instead of repeating this somewhat vague warninng several times, it could be better to create a subsection for all these attributes that are set only in the main Slicer GUI application.

Base/Python/slicer/__init__.py
```diff
@@ -1,4 +1,108 @@
...
+
+    For example::
+
+      >>> slicer.modules.VectorToScalarVolumeInstance
+      <VectorToScalarVolume.VectorToScalarVolume object at 0x7fbfc1d1aa60>
+
+      >>> isinstance(slicer.modules.VectorToScalarVolumeInstance, slicer.ScriptedLoadableModule.ScriptedLoadableModule)
+      True
+
+  * the attribute ``<moduleName>Widget`` is set to the corresponding instance
```
lassoan, line 68: What if the Python module widget is reloaded? Does it point to the current widget instance or the first? The first module widget may become obsolete if the module widget is reloaded.

Docs/developer_guide/python_faq.md
```diff
@@ -2,6 +2,85 @@
...
 
+## What is the Slicer Python Environment?
+
+You can consider each Slicer installation as a virtual Python environment - the same way
+as you create virtual environments using python or conda.
+
+## What is the PythonSlicer executable?
+
+`PythonSlicer` is an executable provided in the `bin` directory of the Slicer installation.
+It’s the same version of Python available through the Slicer application, but it acts like
```
lassoan, line 13:
```suggestion
It is the same version of Python available through the Slicer application, but it acts like
```

Docs/developer_guide/python_faq.md
lassoan, line 14: PythonSlicer acts as a usual `python` executable in every sense, not just in term of running scripts. For example, it can be used as an interactive console, too. So, I'm not sure if we need to narrow its scope (we could just remove the "in term of..." part).

Docs/developer_guide/python_faq.md
```diff
@@ -2,6 +2,85 @@
...
+a normal `python` (or `python3`) executable in terms of running scripts and can be used t
+ access all the packages installed in Slicer.
+
+This means you can use it in place of regular python to make use of the Slicer environment
+for batch processing and command-line operations.
```
lassoan, line 18: Some examples would be great here.

Docs/developer_guide/python_faq.md
```diff
@@ -2,6 +2,85 @@
...
+application, the MRML scene and the loaded modules are not available.
+:::
+
+:::{tip}
+You may use {func}`slicer.util.pip_install()` to install additional packages.
+:::
+
+## What is the Python Console?
+
+This describes the [](/user_guide/user_interface.md#python-console) available through
```
lassoan, line 31: Is the empty `[]` valid syntax?

Docs/developer_guide/python_faq.md
lassoan, line 35: Do you mean *Python* context?
```suggestion
It allows to access to the running Python context including the application instance,
```

Docs/developer_guide/python_faq.md
lassoan: In Slicer SlicerOpenIGTLink should be used (pyigtl just uses Python threading, which is not really parallelization).
pyigtl can be used everywhere else.

lassoan (commented):
Thank you, these changes are very useful. I've added some inline comments.

Docs/developer_guide/python_faq.md
jamesobutler: Since this is Slicer documentation, does it make sense to include essentially general OpenIGTLink usage here? It seems like different uses of IGTLink whether with Slicer or some pure Python implementation might be better in the OpenIGTLink documentation.

lassoan: I agree that conceptually it is redundant to write about these interprocess communication options in the Slicer core documentation, but it has huge practical value to see a summary of all options at one place (similarly how we write about extensions that provide registration modules or DICOM plugins). The descriptions are fairly high-level, so they should not require a lot of maintenance.

Docs/developer_guide/python_faq.md
```diff
@@ -2,6 +2,85 @@
...
+:::{tip}
+You may use {func}`slicer.util.pip_install()` to install additional packages.
+:::
+
+:::{versionchanged} 5.2.0
+The `Python interactor` was renamed to `Python Console`.
+See [related discussion](https://discourse.slicer.org/t/rename-python-interactor-to-python-console/25897).
+:::
+
+## How to access Slicers Python API via an external program while Slicer is running?
```
pieper, line 56: Slicer's

pieper:
Thanks Jc!

I didn't read very carefully.  I suggest we merge, read it in context, and then make new PRs for any improvements.

Docs/developer_guide/python_faq.md
jcfr: It is indeed, it means that corresponding header is used in the generated text. 

> If no text is given or it is an auto-link, e.g. `[](#dest)` or `<project:#dest>`, then MyST will attempt to resolve an implicit text. For example, if the destination is a heading, then the heading text will be used as the link text, or if the destination is a figure/table then the caption will be used as the link text. Otherwise, the link text will be the destination itself.

See https://myst-parser.readthedocs.io/en/latest/syntax/cross-referencing.html#explicit-vs-implicit-link-text

Docs/developer_guide/python_faq.md
jcfr: Very good point, I revisited the wording:

```
`PythonSlicer` is an executable provided in the `bin` directory of the Slicer installation.
It is a Python interpreter that allows access to all of the packages installed in Slicer.
This makes it possible to use the Slicer Python environment from outside the application, such
as for batch processing, command-line operations, terminal-based interactive session and
integration with IDEs.

This means you can use `PythonSlicer` as a replacement for the regular Python interpreter
(`python` or `python3`) to take advantage of the installed packages in the Slicer environment. 

:::{warning}
Please note that the running Python context is **not** available when using `PythonSlicer`.
As a result, many objects, including the application instance, the MRML scene and the loaded
modules, are not available.
:::
```

Docs/developer_guide/python_faq.md
jcfr: Ditto. Here is an issue tracking this:
* https://github.com/Slicer/Slicer/issues/7015


Since depending on the needed capabilities batch processing may have to be implemented through the ``Python Console by leveraging `--python-script` and `--no-main-windows`, I expanded the `What is the Python Console` section with the following:


```
:::{hint}
Running scripts (or code) using the command-line option `--python-script` (or `python-code`)
is equivalent to running code in the Python Console.

Combined with the use of `--no-main-window`, this is useful for implementing batch processing
pipelines leveraging capabilities only available in the context of a running application.
For example, this applies to the [Segment Editor effects](/user_guide/modules/segmenteditor.html#effects).
:::
```

Base/Python/slicer/__init__.py
jcfr: Ditto. Here is an issue tracking this:
* https://github.com/Slicer/Slicer/issues/7016

I the mean time, I will update the consistently update the waning based on your suggestion.

Docs/developer_guide/python_faq.md
jcfr: I think I found a nice compromise in the context of the API documentation:

image

jcfr:
Thanks everyone for the feedback :pray: 

To summarize, I think I addressed most of the concerns

I also created the following new issues to track future improvements:
* https://github.com/Slicer/Slicer/issues/7015
* https://github.com/Slicer/Slicer/issues/7016

If there are no other concerns, I would to move forward with integration.

> re: pyigtl

Should we improve the wording of the FAQ entry `## How to access Slicer's Python API via an external program while Slicer is running?` or are we happy with it ?

Docs/developer_guide/python_faq.md
lassoan: Thanks for the explanation.

lassoan:
Just two small clarifications that could be done Beedie merging:

1. Can you describe what happens to the ...WidgetInstance attribute if a Python scripted loadables module if reloaded?
2. Add a comment that pyigtl is intended to be used outside Slicer, to send/receive data to/from Slicer.

jcfr:
> re: Add a comment that pyigtl is intended to be used outside Slicer, to stream data to/from Slicer.


```diff
diff --git a/Docs/developer_guide/python_faq.md b/Docs/developer_guide/python_faq.md
index 36a8081072..05cb02d553 100644
--- a/Docs/developer_guide/python_faq.md
+++ b/Docs/developer_guide/python_faq.md
@@ -69,9 +69,12 @@ There are several ways to access Slicer's Python API from an external program wh
 - [SlicerOpenIGTLink](https://github.com/openigtlink/SlicerOpenIGTLink): A lightweight socket-based
   protocol for real-time data transfer. This is useful for applications that need to send many
   requests per second (e.g., continuous data streaming) or for clients that only have access to
-  sockets and don't want to reimplement complex protocols like http. In most cases, it performs well
-  for sending requests in the range of 10 to 100 requests per second.
-  There is a native Python implementation: [pyigtl](https://pypi.org/project/pyigtl/).
+  sockets and prefer to avoid the complexity of protocols like HTTP. In most cases, it performs well
+  for sending requests at a rate of 10 to 100 requests per second.
+
+  A native Python implementation, [pyigtl](https://pypi.org/project/pyigtl/) , is available for
+  use outside the Slicer application. It can be used to both stream data from Slicer and stream
+  data to Slicer.
 
 - [SlicerJupyter](https://github.com/Slicer/SlicerJupyter#readme): A protocol for interacting with
   Slicer using standard Jupyter kernel protocol (simple protocol built on top of ZeroMQ middleware).
```

jcfr:
> re: Can you describe what happens to the ...WidgetInstance attribute if a Python scripted loadables module if reloaded?

I checked and is indeed updated.

image


https://github.com/Slicer/Slicer/blob/d4ba0f61075206109d3678d5d70c52bd741e64a3/Base/Python/slicer/util.py#L1364-L1367

PR #7018, opened 2023-06-12
BUG: Skip camera undo when undo is not enabled on vtkMRMLCameraNode

Sunderlandkyl:
Undo states were being added to the stack when the camera was modified, even if undo was not enabled for the camera.


lassoan (approved):
This looks good, thank you!

PR #7020, opened 2023-06-13
Update Debian distribution names in linux.md

benzwick:
Debian 12 "bookworm" released on June 10th, 2023 is the new stable distribution.


jcfr:

I added the `DOC: ` prefix while doing a `Squash & merge`.

PR #7022, opened 2023-06-13
BUG: Update test_slicer_util_save to use non-deprecated method

jcfr:
This fix is a follow of these commits:
* 37416c2e5 (ENH: Use inclusive language for segmentation source representation)
* 5204e7a31 (ENH: Improve default description selection in Add data window)


jamesobutler (approved):
Thanks!

PR #7028, opened 2023-06-18
ENH: Add menu actions to hide or show selected items in subject tree

benzwick:
This change adds two menu actions in the subject hierachy tree view context menu to hide or show the selected items.
This is similar to the existing "Toggle visibility" menu action.

### DIFF
--- a/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyTreeView.cxx
+++ b/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyTreeView.cxx
@@ -74,6 +74,14 @@ protected:
   qMRMLSubjectHierarchyTreeView* const q_ptr;
 
 public:
+
+  enum VisibilityAction
+  {
+    Show,
+    Hide,
+    ToggleVisibility
+  };
+
   qMRMLSubjectHierarchyTreeViewPrivate(qMRMLSubjectHierarchyTreeView& object);
 
   virtual void init();
@@ -87,6 +95,11 @@ public:
   void applyTransformToItem(vtkIdType itemID, const char* transformNodeID);
   vtkMRMLTransformNode* appliedTransformToItem(vtkIdType itemID, bool& commonToAllChildren);
   vtkMRMLTransformNode* firstAppliedTransformToSelectedItems();
+
+  void setSubjectHierarchyItemVisibility(vtkIdType itemID, VisibilityAction visibilityAction);
+
+  /// Set visibility of selected subject hierarchy items
+  void setVisibilityOfSelectedItems(VisibilityAction visibilityAction);
 
   void updateColors();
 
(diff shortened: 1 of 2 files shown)


cpinter:
* If I understand correctly would these new items only show up when multiple items are selected?
* It is a bit strange to me to have three actions for visibility. Would it make sense to remove the toggle option? I think its usefulness will decrease considerably once we have the show and the hide options.

pieper:
Don't take away toggle please, I often use that, for example, to select a set of segments to view by turning them off, and then selecting a whole range and picking toggle so that only the previously invisible ones become visible.  This becomes essentially "show only the selected items" with a bit more flexibility since you can select which ones to toggle.

Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyTreeView.cxx
```diff
@@ -1346,7 +1358,26 @@ void qMRMLSubjectHierarchyTreeView::toggleSubjectHierarchyItemVisibility(vtkIdTy
...
+      }
+    case 1:
+      {
+      visible = 1;
+      break;
+      }
+    case 2:
+      {
+      visible = (ownerPlugin->getDisplayVisibility(itemID) > 0 ? 0 : 1);
+      }
```
jamesobutler, line 1377: Probably want to consider the usage of an official enum for clarity about what 0, 1, and 2 mean here:

https://github.com/Slicer/Slicer/blob/ed6fd97d5d098d384fe4e09c6fac2cd7c7e4fd83/Libs/MRML/Core/vtkMRMLSliceCompositeNode.h#L138-L145
https://github.com/Slicer/Slicer/blob/ed6fd97d5d098d384fe4e09c6fac2cd7c7e4fd83/Libs/MRML/Widgets/qMRMLSliceControllerWidget.cxx#L1079-L1090

jamesobutler (dismissed):
This seems to work and I also see the value of hide/show the selected items regardless of their current visibility state.

https://github.com/Slicer/Slicer/assets/15837524/51912f41-5f7d-400f-ad13-fbe9fd5c6247

cpinter (approved):
Looks good to me. @benzwick you can click merge unless you'd like to change something first. Thanks for the contribution!

Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyTreeView.cxx
lassoan: We could use QItemSelectionModel::SelectionFlag enums:
        Select         = 0x0002,
        Deselect       = 0x0004,
        Toggle         = 0x0008,

benzwick: That sounds like a good idea. Will the meaning Deselect=Hide and Select=Show be clear in this context? Or would it be better to define a new enum?

lassoan (approved):
To accelerate this, I've added the enum and force-pushed the change to this branch. It should be good to merge.

PR #7031, opened 2023-06-19
DOC: Update linux.md to add note about Debian 12 CMake

benzwick:
Slicer is not compatible with CMake 3.25.1 that ships with Debian 12 (bookworm).


pieper (approved):
Looks good to me 👍

jamesobutler:
@jcfr Any idea why the "Build Slicer" workflow is pending here? This is preventing the integration of this PR.

lassoan:
Sometimes readthedocs jobs are not completed. I've merged this manually now.

PR #7035, opened 2023-06-21
BUG: Fix the application's applicationLocale object

lassoan:
The applicationLocale was always set to the default language (en-us).


jamesobutler (approved):
Changes make sense

Issue #6990, opened 2023-05-30
faild to install the debug and release version of slicer multipy times


ML-lilith:
## Steps to reproduce
set up based on https://slicer.readthedocs.io/en/latest/developer_guide/build_instructions/windows.html. 

![image](https://github.com/Slicer/Slicer/assets/125838507/f9ff8cb3-6c84-4760-a44c-be92a1039977)

tried different cmake version but all in the allwoed range. 
![image](https://github.com/Slicer/Slicer/assets/125838507/ff587963-aead-49ee-8df4-5c8525b86763)
also tried to use visualstudio 2019.   all faild no matter the release version and debug version. 
problem(happened many times):
## CMake configure and generate successful. 

```
Setting C++ standard
Setting C++ standard - C++17
Selecting Windows SDK version 10.0.22000.0 to target Windows 10.0.22621.
Configuring Slicer organization domain [slicer.org]
Configuring Slicer organization name [slicer.org]
Configuring Slicer default home module [Welcome]
Configuring Slicer default favorite modules [Data, Volumes, Models, Transforms, Markups, SegmentEditor]
Configuring Slicer text of disclaimer at startup [Thank you for using %1!

This software is not intended for clinical use.]
Configuring Slicer requires admin account [OFF]
Configuring Slicer install root [$LOCALAPPDATA/slicer.org]
Configuring Slicer release type [Experimental]
Configuring Slicer version [5.3.0-2023-05-29]
Configuring Slicer revision [31764]
Configuring VTK
  Slicer_VTK_RENDERING_BACKEND is OpenGL2
  Slicer_VTK_SMP_IMPLEMENTATION_TYPE is TBB
  Slicer_VTK_VERSION_MAJOR is 9
Configuring Slicer with Qt 5.15.2 (using modules: Core, Widgets, Network, OpenGL, PrintSupport, UiTools, Xml, XmlPatterns, Svg, Sql, Multimedia, MultimediaWidgets, WebEngine, WebEngineWidgets, WebChannel, Script, LinguistTools, Test, )
Setting QT_PLUGINS_DIR: C:/Qt/Qt5.15.2/5.15.2/msvc2019_64/plugins
Setting QT_BINARY_DIR: C:/Qt/Qt5.15.2/5.15.2/msvc2019_64/bin
Setting QT_LIBRARY_DIR: C:/Qt/Qt5.15.2/5.15.2/msvc2019_64/lib
[...]

PR #5050, opened 2020-07-20
Documentation build instructions

RafaelPalomar:
**This is a work-in-progress branch**

As suggested in the last meeting, I started a documentation section for building Slicer on Linux using different distributions. I welcome comments on structure and contents and appreciate feedback on the following questions:

- I added the section to the `index.rst` file. It seems, however, that the section link does not appear in the left column when visiting some other content (e.g., content in the user guide)?
- The content is inspired on the current Slicer wiki content. There are a few links to other articles (e.g., link to https://www.slicer.org/wiki/Documentation/Nightly/Developers/DevelopmentWithGit). What to to with this type of content: keep external link? open an new section in the readthedocs documentation?
- In the same way, in the Wiki, there are links to presentations and pdf files.  Should we transfer the files and link them?
- Change the `building_slicer.md` to `.rst` format? For adding tips, warnings, notes, etc., one need to embed `html` code into the markdown, which is not very elegant.

TODOs:
- [ ] Add installation of pre-requisites for Fedora
- [ ] Add installation of pre-requisites for Gentoo
- [ ] Add a note to use `-DSlicer_USE_SYSTEM_curl=ON` for Gentoo/Arch while #5049 is open

PR #7036, opened 2023-06-21
DOC: Update build instructions in windows.md

benzwick:
Update build instructions as suggested by @jcfr in https://github.com/NousNav/NousNav/issues/355#issuecomment-1601495014

### DIFF
--- a/Docs/developer_guide/build_instructions/windows.md
+++ b/Docs/developer_guide/build_instructions/windows.md
@@ -34,10 +34,10 @@ Slicer relies on a number of large third-party libraries (such VTK, ITK, DCMTK),
 
 ## Set up source and build folders
 
-- Create source folder. This folder will be referred to as `<Slicer_SOURCE>` in the following. Recommended path: `C:\D\S4`
+- Create source folder. This folder will be referred to as `<Slicer_SOURCE>` in the following. Recommended path: `C:\D\S`
   - Due to maximum path length limitations during build the build process, source and build folders must be located in a folder with very short (couple of characters) total path length.
   - While it is not enforced, we strongly recommend you to avoid the use of spaces for both the source directory and the build directory.
-- Create build folder. This folder will be referred to as `<Slicer_BUILD>` in the following. Recommended path: `C:\D\S4R` for release-mode build, `C:\D\S4D` for debug-mode build.
+- Create build folder. This folder will be referred to as `<Slicer_BUILD>` in the following. Recommended path: `C:\D\SR` for release-mode build, `C:\D\SD` for debug-mode build.
   - You cannot use the same build tree for both release or debug mode builds. If both build types are needed, then the same source directory can be used, but a separate build directory must be created and configured for each build type.
   - How to decide between Debug and Release mode?
     - Release mode build: runs at same speed as official build, requires less disk space (about 15GB); but step-by-step debugging is not available
@@ -65,9 +65,9 @@ cd C:\D
 "C:\Program Files\CMake\bin\cmake.exe" ^
   -G "Visual Studio 17 2022" -A x64 ^
   -DQt5_DIR:PATH=C:/Qt/5.15.2/msvc2019_64/lib/cmake/Qt5 ^
-  -S C:\D\S4 -B C:\D\S4R
+  -S C:\D\S -B C:\D\SR
 
-"C:\Program Files\CMake\bin\cmake.exe" --build C:\D\S4R --config Release
+"C:\Program Files\CMake\bin\cmake.exe" --build C:\D\SR --config Release
 ```
 
 Debug mode:
@@ -78,9 +78,9 @@ cd C:\D
 "C:\Program Files\CMake\bin\cmake.exe" ^
   -G "Visual Studio 17 2022" -A x64 ^
   -DQt5_DIR:PATH=C:/Qt/5.15.2/msvc2019_64/lib/cmake/Qt5 ^
-  -S C:\D\S4 -B C:\D\S4D
+  -S C:\D\S -B C:\D\SD
 
-"C:\Program Files\CMake\bin\cmake.exe" --build C:\D\S4D --config Debug
+"C:\Program Files\CMake\bin\cmake.exe" --build C:\D\SD --config Debug
 ```
 
 ### Using graphical user interface (alternative solution)
@@ -135,7 +135,7 @@ Errors due to missing Python libraries (or other Python related errors, such as 
 
 ### Custom Slicer and CTK widgets do not show up in Qt designer
 
-Qt Designer can only use designer plugins (in `c:\D\S4R\Slicer-build\bin\designer` and `c:\D\S4R\CTK-build\CTK-build\bin\designer`) that are built in the same mode (e.g. `Debug` or `Release`).
+Qt Designer can only use designer plugins (in `c:\D\SR\Slicer-build\bin\designer` and `c:\D\SR\CTK-build\CTK-build\bin\designer`) that are built in the same mode (e.g. `Debug` or `Release`).
 
 In pre-built Qt packages (downloaded from Qt website) Qt Designer is only provided in `Release` mode. If Qt is built from source in `Release` mode (default) or `Release and Debug` mode then Qt designer will be in `Release` mode. In all these cases, Qt Designer can be used only if Slicer is built in `Release` mode.
 
--- a/Docs/developer_guide/build_instructions/windows.md
+++ b/Docs/developer_guide/build_instructions/windows.md
@@ -59,20 +59,28 @@ Specify source, build, and Qt location and compiler version and start the build 
 
 Release mode:
 
-```
-mkdir C:\D\S4R
-cd /d C:\D\S4R
-"C:\Program Files\CMake\bin\cmake.exe" -G "Visual Studio 17 2022" -A x64 -DQt5_DIR:PATH=C:\Qt\5.15.2\msvc2019_64\lib\cmake\Qt5 C:\D\S4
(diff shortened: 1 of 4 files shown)


Docs/developer_guide/build_instructions/windows.md
```diff
@@ -56,20 +56,30 @@ Specify source, build, and Qt location and compiler version and start the build
 
 Release mode:
 
-```
+```bat
+cd C:\D
 mkdir C:\D\S4R
```
jcfr, line 61: We can also remove the explicit creation of the build directory

Docs/developer_guide/build_instructions/windows.md
```diff
@@ -56,20 +56,30 @@ Specify source, build, and Qt location and compiler version and start the build
...
-cd /d C:\D\S4R
-"C:\Program Files\CMake\bin\cmake.exe" -G "Visual Studio 17 2022" -A x64 -DQt5_DIR:PATH=C:\Qt\5.15.2\msvc2019_64\lib\cmake\Qt5 C:\D\S4
-"C:\Program Files\CMake\bin\cmake.exe" --build . --config Release
+
+cmake ^
+  -G "Visual Studio 17 2022" -A x64 ^
+  -DQt5_DIR:PATH=C:/Qt/5.15.2/msvc2019_64/lib/cmake/Qt5 ^
+  -S C:\D\S4 -B C:\D\S4R
```
jcfr, line 66: For reference, since the minimum required version of CMake is 3.16, the `-S` and `-B` options are available. See https://cmake.org/cmake/help/v3.16/manual/cmake.1.html#options

Docs/developer_guide/build_instructions/windows.md
```diff
@@ -56,20 +56,30 @@ Specify source, build, and Qt location and compiler version and start the build
...
 
-```
+```bat
+cd C:\D
 mkdir C:\D\S4R
-cd /d C:\D\S4R
-"C:\Program Files\CMake\bin\cmake.exe" -G "Visual Studio 17 2022" -A x64 -DQt5_DIR:PATH=C:\Qt\5.15.2\msvc2019_64\lib\cmake\Qt5 C:\D\S4
-"C:\Program Files\CMake\bin\cmake.exe" --build . --config Release
+
+cmake ^
```
jcfr, line 63: @benzwick For consistency, the instructions in `developer_guide/extensions.md` should also be updated.


@pieper @lassoan @jamesobutler

I suggested simply using `cmake` instead of the more verbose `"C:\Program Files\CMake\bin\cmake.exe"`

jcfr (commented):
Thanks for the update :pray:

Docs/developer_guide/build_instructions/windows.md
jamesobutler: The default option for the cmake installer (at least on Windows) is to not add CMake to the system PATH. A novice builder may run into issues here

image

Docs/developer_guide/build_instructions/windows.md
lassoan: Why would we remove the explicit creation of the build directory? We usually start a clean build by removing the entire build folder and it is useful that this script recreates it (otherwise CMake would dump error files etc. in the level above).

jcfr: Specifying `-B C:\D\S4R` takes care of creating the build directory.

Docs/developer_guide/build_instructions/windows.md
lassoan: I agree, I would vote for keeping `C:\Program Files\CMake\bin\cmake.exe` because it works by default, while `cmake` does not.

benzwick: OK, I guess I forgot that I selected that option when I installed CMake.

Docs/developer_guide/build_instructions/windows.md
benzwick: In that case, I think we can also remove these commands:
```
cd C:\D
```

jcfr: Assuming that the current drive doesn't need to be selected, removing `cd C:\D` would be sensible.

That said, I think keeping that command is a helpful hint indicating that everything will be taking place in that sub-directory.

Docs/developer_guide/build_instructions/windows.md
benzwick: I changed it back to `C:\Program Files\CMake\bin\cmake.exe`

Docs/developer_guide/build_instructions/windows.md
benzwick: I removed only the `mkdir` command.

benzwick:
Should we also be using the `--parallel [<jobs>]` or `-j [<jobs>]` [option](https://cmake.org/cmake/help/latest/release/3.12.html#command-line) to specify a parallel build level in CMake?

jcfr:
> Should we also be using the `--parallel [<jobs>]` or `-j [<jobs>]`

A tip similar to the one added here could be added. See https://slicer.readthedocs.io/en/latest/developer_guide/build_instructions/linux.html#build-slicer

On windows  the env. variable `NUMBER_OF_PROCESSORS` can be useful.

I also realized that the tip on linux may be incorrect, I think it should be adjusted like this[^1]:

```diff
- As a rule of thumb, many use the `number of CPU threads + 1` as the number of parallel builds.
+ As a rule of thumb, many use the `number of CPU cores  - 1` as the number of parallel builds.
```

I think the `number of CPU threads + 1` was adapted from the guidelines associated with Gentoo instructions[^2][^3] where the `-l` is also used.

[^2]: https://wiki.gentoo.org/wiki/MAKEOPTS
[^3]: https://web.archive.org/web/20140717224138/http://preney.ca/paul/archives/341

References:
* https://github.com/Slicer/Slicer/pull/5050

[^1]: https://community.nxp.com/t5/i-MX-Processors/How-can-I-reduce-the-number-of-parallel-yocato-builds/m-p/1068778/highlight/true#M157030

cc: @RafaelPalomar

RafaelPalomar:
> I also realized that the tip on linux may be incorrect, I think it should probably be adjusted like this[1](#user-content-fn-1-6cd15c14e0b7be56043601a3c57eb2be):
> 
> ```diff
> - As a rule of thumb, many use the `number of CPU threads + 1` as the number of parallel builds.
> + As a rule of thumb, many use the `number of CPU cores  - 1` as the number of parallel builds.
> ```

@jcfr  you are right. In addition, we should update the GNU/Linux build instructions to more modern distributions.

benzwick:
Should we add this option until #6990 is fixed?
```bat
  -T "version=14.34"
```
See [Visual Studio Toolset Selection](https://cmake.org/cmake/help/latest/variable/CMAKE_GENERATOR_TOOLSET.html#visual-studio-toolset-selection)

benzwick:
> > I also realized that the tip on linux may be incorrect, I think it should probably be adjusted like this[1](#user-content-fn-1-6cd15c14e0b7be56043601a3c57eb2be):
> > ```diff
> > - As a rule of thumb, many use the `number of CPU threads + 1` as the number of parallel builds.
> > + As a rule of thumb, many use the `number of CPU cores  - 1` as the number of parallel builds.
> > ```
> 
> @jcfr you are right.

Fixed in 99a02a1266c3a5d6b6b0eab3c22faba729b47f51

lassoan:
> Should we add this option until #6990 is fixed?
> 
> ```batchfile
>   -T "version=14.34"
> ```
> 
> See [Visual Studio Toolset Selection](https://cmake.org/cmake/help/latest/variable/CMAKE_GENERATOR_TOOLSET.html#visual-studio-toolset-selection)

Toolset selection is mostly just a theoretical possibility, as it is hard to get anything but the latest toolset if you use the Visual Studio Community.

Setting a specific version would also be more restrictive than needed. Setting something like <14.36 would be OK, but by requiring exactly 14.34 we would cause unnecessary failures.

Docs/developer_guide/build_instructions/windows.md
```diff
@@ -56,20 +56,28 @@ Specify source, build, and Qt location and compiler version and start the build
...
-cd /d C:\D\S4R
-"C:\Program Files\CMake\bin\cmake.exe" -G "Visual Studio 17 2022" -A x64 -DQt5_DIR:PATH=C:\Qt\5.15.2\msvc2019_64\lib\cmake\Qt5 C:\D\S4
-"C:\Program Files\CMake\bin\cmake.exe" --build . --config Release
+```bat
+cd C:\D
+
+"C:\Program Files\CMake\bin\cmake.exe" ^
+  -G "Visual Studio 17 2022" -A x64 ^
+  -DQt5_DIR:PATH=C:/Qt/5.15.2/msvc2019_64/lib/cmake/Qt5 ^
+  -S C:\D\S4 -B C:\D\S4R
```
jamesobutler, line 65: “S4” was short for Slicer version 4, however we are now on Slicer 5 versions so “S5” would be more accurate as a short description.

The change to S4 to S5 would be propagated elsewhere as well.
```suggestion
  -S C:\D\S5 -B C:\D\S5R
```

benzwick: Could we change it to S instead of S5 so it doesn't need to be changed for each new major version?

jcfr: That makes sense.

jamesobutler: I have gone ahead and pushed a commit with that change.

jamesobutler (dismissed):

jamesobutler (approved):

PR #7038, opened 2023-06-22
ENH: Add "Test" Button to ScriptedLoadableModule UI

HarryDC:
Some modules don't behave well under Reload which means that Reload & Test cannot be used.
This adds a Test button to the UI of ScriptedLoadableModule and enables access to the unit tests

### DIFF
--- a/Base/Python/slicer/ScriptedLoadableModule.py
+++ b/Base/Python/slicer/ScriptedLoadableModule.py
@@ -131,6 +131,8 @@ class ScriptedLoadableModuleWidget:
         """
         if moduleName == self.moduleName:
             self.cleanup()
+            settings = qt.QSettings()
+            settings.setValue(f'{self.moduleName}/_ReloadAndTest', self.reloadTestMenuButton.defaultAction().objectName)
             slicer.app.moduleManager().disconnect(
                 'moduleAboutToBeUnloaded(QString)', self._onModuleAboutToBeUnloaded)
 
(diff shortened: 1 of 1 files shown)


lassoan (commented):
Thanks for your contribution. I agree it can make sense to just test without reload (although in general reloading can/should be fixed to make testing easier). While there may be space for a third button, it seems a bit wasteful to have 3 buttons for reload, reload&test, test; and it could also make a bit more difficult to choose the correct button each time you want to test.

What do you think about adding a dropdown menu (using ctkMenuButton) to the "Reload and Test" button that would allow you to enable/disable reloading before testing? The button title could switch between "Test" and "Reload and Test" depending on that flag.

You could store in application settings if reloading before testing is disabled for a module (so that you don't need to keep using the dropdown menu for disabling the reloading).

HarryDC:
Forgot to post a picture, here is the interface right now: 
image

I ran into the reloading issue while adding to SlicerPipelines, and it starts coming up when the scripted module that's being reloaded imports other python modules that would need reloading. 

Changing the two to a ctkMenuButton is fine with me, when you mention storing the default that should be per module I assume or ?

HarryDC:
@lassoan I was looking around the ctk toolbox widgets and it looks like there isn't anything that really could implement the "default" button just by itself. We can add a `ctkMenuButton` that would let the user choose between test and reload and test but that is always going to be a button. I'm fine with that. But i'm not sure there is a simple way to setup a button with a menu attach to select the functionality for the button. This might make the UI more complicated that intended. Especially in a place that is meant for development.

Let me know what you think, thanks

lassoan:
vtkMenuButton should work well. See for example the "AutoRun" button in CLI modules, such as "Gaussian Blur Image Filter":

image

jcfr:
Topic has been rebased against latest `main`

What about adding `ctkMenuButton` so that there are the following buttons available in the first row ?

* `Reload`
* `Test`
* `Restart Slicer`

Where `Test` would be updated to use a `ctkMenuButton` with a sub-option called `Test and Reload` toggled to `on` by default.

If systematically toggling `Test and Reload` is tedious ... we could probably have the default saved in the settings ?

jamesobutler:
This individual “Test” action appears to be the same as the individual “SlicerPipelines” button located in the “SelfTests” module which can trigger the module’s unit tests back-to-back without reload. That requires a switch of the active module from SlicerPipelines to SelfTests, but maybe something @HarryDC can use if adding another action to the reload, reload and test section is not desired.

HarryDC:
@jamesobutler How do you navigate or enable the "SelfTest" module, is that a build option ?

lassoan:
> What about adding `ctkMenuButton` so...

@jcfr your entire suggestion sounds perfect to me.

@HarryDC would you be ready to implement it? (it would be just a couple of extra lines of code compared to what you have already implemented)

The "Test" button in SelfTests module is useful, but it still makes sense to offer "Test without reload" feature in the module GUI (it is easier to find, no need to switch to another module).

jcfr:
> How do you navigate or enable the "SelfTest" module, is that a build option 

For completeness, after enabling [Developer Mode](https://slicer.readthedocs.io/en/latest/user_guide/settings.html#developer-mode0, it should be available in `Modules -> Testing -> SelfTests`.

HarryDC:
@lassoan @jcfr Updated with menu button, i went for the `QToolButton` rather than the `ctkMenuButton` it seemed that it needed less logic (i.e. work ;) ) for managing the on button click logic. The `QAction` of the menu can be directly used in the button through `setDefaultAction`. Maybe I missed something but `ctkMenuButton` derives from `QPushButton` and doesn't expose the menu `QAction` to the button itself. 

Here's how the UI looks right now
image

jcfr (approved):

jcfr:
@HarryDC After you are done testing, we will squash & merge with the following message:

```
ENH: Add "Test" Button to ScriptedLoadableModule UI (#7038)

Improve the reloading and testing experience for ScriptedLoadableModule
by adding a "Test" button to the module's user interface available when
developer mode is enabled.

This enhancement allows direct access to the unit tests and addresses
the issue with modules not behaving well under reload.

The preferred action for each module (ReloadAndTest, Test) is now stored
in the settings under the key "<ModuleName>/PreferredReloadAndTestAction".

```

HarryDC:
@jcfr looks good on my side

PR #7041, opened 2023-06-23
DOC: Add note on build error with latest MSVC toolset 14.36

lassoan:
see #6990


benzwick:
Also, reconfiguring the VTK build as described [here](https://discourse.slicer.org/t/c1001-internal-compiler-error/29604) worked for me when building Slicer but not NousNav (it fixed the IOIOSS error but then there were others). Maybe it's better to remove that suggestion. It's also not clear to new users how to reconfigure VTK in this context. For reference, I used:
```
cd /d C:\D\S4R\VTK-build
cmake -DVTK_MODULE_ENABLE_VTK_IOIOSS=NO ..\VTK
```

lassoan:
@jcfr how far is the VTK update? We have all these discussions and evaluating various workarounds, while updating VTK would just fix all these at once.

jcfr:
> how far is the VTK update? We have all these discussions and evaluating various workarounds, while updating VTK would just fix all these at once.

VTK has just been updated :rocket: :heavy_check_mark: 

For reference:
* #7010

lassoan:
This note is no longer needed, since VTK is now updated.

PR #7049, opened 2023-06-26
BUG: Fix file close in vtkSlicerVolumesLogic::InitializeDefaultVolumeDisplayPresets

jcfr:
This commit is a follow-up of c72d89051 (ENH: Add volume display preset selection to slice view right-click menu)

It ensures that file is closed only after all JSON stream parsing is completed.

See https://rapidjson.org/md_doc_stream.html#FileStreams

### DIFF
--- a/Modules/Loadable/Volumes/Logic/vtkSlicerVolumesLogic.cxx
+++ b/Modules/Loadable/Volumes/Logic/vtkSlicerVolumesLogic.cxx
@@ -1602,14 +1602,16 @@ void vtkSlicerVolumesLogic::InitializeDefaultVolumeDisplayPresets()
     {
     vtkErrorMacro("vtkSlicerVolumesLogic::InitializeDefaultVolumeDisplayPresets failed: Error parsing the file '" << displayPresetsFilename << "'.");
     fclose(fp);
+    return;
     }
-  fclose(fp);
+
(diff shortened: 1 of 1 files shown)


lassoan (commented):
For me it is not obvious from the Rapidjson documentation that the stream is still used after ParseStream. Is there anything specific that you have found that suggested that we cannot close the file after ParseStream?

If we think that the file must not be closed until jsonRoot is still used then we need to add a comment in the code to ensure that the fclose calls are not refactored in the future.

jcfr:
> Is there anything specific that you have found that suggested that we cannot close the file after ParseStream?

Since the file had a number of character < 4096, it is working as expected.

That said, "if it runs out of characters in the buffer, it will read the next part from file.". For that reason the descriptor should remains open.

Inspecting the code confirms this:

https://github.com/Tencent/rapidjson/blob/973dc9c06dcd3d035ebd039cfb9ea457721ec213/include/rapidjson/filereadstream.h#L66-L83

jcfr:
Since the file is "small", we could revisit and read the complete file into a string by interface with `RapidJON` using `StringStream`

lassoan:
Should we close this? As far as I remember we concluded that this PR is not needed after all (the content is read from the stream).

PR #7050, opened 2023-06-26
COMP: Fix potentially invalid pointer in vtkSlicerSegmentationsModuleLogic

jcfr:


lassoan (approved):
Good catch, thank you for the fix.

PR #7055, opened 2023-06-28
DOC: Fix typo in `Transforms.dox`

luzpaz:
s/spacial/spatial/

Edit: it could also be 'special' but in context seemed like the former

### DIFF
--- a/Modules/Loadable/Transforms/Documentation/Transforms.dox
+++ b/Modules/Loadable/Transforms/Documentation/Transforms.dox
@@ -2,7 +2,7 @@
 \defgroup Slicer_QtModules_Transforms Transforms
 \ingroup Slicer_QtModules
 This module is used for creating, editing, and displaying transformations.
-Transformation nodes are used in Slicer to define spacial relationships between different nodes (such as volumes, models, markups, ROI's, or other transform nodes) or between the nodes and the global RAS space.
+Transformation nodes are used in Slicer to define spatial relationships between different nodes (such as volumes, models, markups, ROIs, or other transform nodes) or between the nodes and the global RAS space.
 You can establish these relations by moving nodes from the Transformable list to the Transformed list or by dragging the nodes under the Transformation nodes in the Data module.
 
 */


Modules/Loadable/Transforms/Documentation/Transforms.dox
```diff
@@ -2,7 +2,7 @@
 \defgroup Slicer_QtModules_Transforms Transforms
 \ingroup Slicer_QtModules
 This module is used for creating, editing, and displaying transformations.
-Transformation nodes are used in Slicer to define spacial relationships between different nodes (such as volumes, models, markups, ROI's, or other transform nodes) or between the nodes and the global RAS space.
+Transformation nodes are used in Slicer to define spatial relationships between different nodes (such as volumes, models, markups, ROI's, or other transform nodes) or between the nodes and the global RAS space.
```
jamesobutler, line 5: Here is another one. It may be misused in other places as well?
```suggestion
Transformation nodes are used in Slicer to define spatial relationships between different nodes (such as volumes, models, markups, ROIs, or other transform nodes) or between the nodes and the global RAS space.
```

luzpaz: I found:

https://github.com/Slicer/Slicer/blob/571cba35f5715c344101b808c1283edec20adeb5/Modules/Loadable/CropVolume/Logic/vtkSlicerCropVolumeLogic.cxx#L676

but it seems correct in context, right?

jamesobutler: Yes that one is acceptable. Thanks for looking!

jamesobutler:
@luzpaz If you can squash down the second commit into the first, the commit message status check will pass.

lassoan:
Could we remove these dox files (move their content to header file of the module class)?

luzpaz:
> Could we remove these dox files (move their content to header file of the module class)?

Out of the scope of this PR unless someone wants to jump in and do that, by all means

Issue #7056, opened 2023-06-28
License file contains typo


luzpaz:
## Summary

Heads up: there is a typo in the License.txt file https://github.com/Slicer/Slicer/blob/183584877ff421e18a7bdeff893bf651ad14ffd1/License.txt#L82
should obviously be `REQUIRES`

## Steps to reproduce

n/a

## Expected behavior

n/a

## Environment

n/a

RafaelPalomar:
Thanks for reporting @luzpaz!

PR #7057, opened 2023-06-28
DOC: Fix typo in `License.txt`

RafaelPalomar:
This fixes a small typo on the `License.txt` file. This closes Slicer/Slicer#7056

### DIFF
--- a/License.txt
+++ b/License.txt
@@ -79,7 +79,7 @@ Sublicense ("Contribution Agreement").
    your breach of any of the terms of this Agreement.
 
 6. YOU WARRANT THAT TO THE BEST OF YOUR KNOWLEDGE YOUR CONTRIBUTION
-   DOES NOT CONTAIN ANY CODE THAT REQURES OR PRESCRIBES AN "OPEN
+   DOES NOT CONTAIN ANY CODE THAT REQUIRES OR PRESCRIBES AN "OPEN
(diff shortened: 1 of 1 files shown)


jcfr:
For reference, here are the prior updates to the license:
* https://github.com/Slicer/Slicer/commits/main/License.txt

jcfr:
To avoid updating the license file, it was included to the list of file to ignore in `.github/workflows/codespell.yml`. 

That said, I don't see any reason for not moving forward. Historically we have fixed typo. See  https://github.com/Slicer/Slicer/commit/ced0ab6d84996f5f802d4cd7b36b334c8083f676

If we move forward with this, the following file will  have to be updated:

https://github.com/Slicer/Slicer/blob/d1385c51350f9565cec8e3b12fc08ba05bc473c1/.github/workflows/codespell.yml#L15

cc: @lassoan @pieper

jcfr (approved):

jcfr:
I also suggest to add the following at the end of the commit message:

```
Co-authored-by: luzpaz <luzpaz@users.noreply.github.com>
```

jcfr (changes requested):
Consider adding missing `Co-authored-by: `

lassoan:
This is a trivial typo, so I don't see any potential issue with changing the license file to fix this.

PR #7059, opened 2023-06-28
ENH: Make ctk ErrorLogWidget dockable

Punzo:
ref https://discourse.slicer.org/t/log-window-cannot-be-hidden-in-enviroments-with-no-windows-manager/29633

@pieper @lassoan @jamesobutler can you please review? thanks!

### DIFF
--- a/Base/Python/slicer/util.py
+++ b/Base/Python/slicer/util.py
@@ -620,6 +620,17 @@ def setPythonConsoleVisible(visible):
     mw.pythonConsole().parent().setVisible(visible)
 
 
+def setErrorLogVisible(visible):
+    """Show/hide Error log window.
+
+    If there is no main window then the function has no effect.
+    """
+    mw = mainWindow()
+    if not mw:
+        return
+    mw.errorLogDockWidget().setVisible(visible)
+
+
 def setStatusBarVisible(visible):
     """Show/hide status bar
 
--- a/Base/QTApp/Resources/UI/qSlicerMainWindow.ui
+++ b/Base/QTApp/Resources/UI/qSlicerMainWindow.ui
@@ -19,6 +19,9 @@
    <property name="allowedAreas">
     <set>Qt::LeftDockWidgetArea|Qt::RightDockWidgetArea</set>
    </property>
+   <property name="windowTitle">
+    <string>Module Panel</string>
+   </property>
    <attribute name="dockWidgetArea">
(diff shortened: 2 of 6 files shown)


Punzo:
I broke CTRL+0 action, I will add a commit

lassoan (commented):
Thank you this works very nicely.

I'm just wondering if we should make it work the same way as the Python console, i.e., hitting Ctrl-0 should toggle the window instead of bringing it to the top. For me it would make it much easier to quickly peek at the log window (currently I can bring it up by Ctrl-0 but then I need to use the mouse to close it).

@pieper @jcfr Do you agree that we should change the behavior of Ctrl-0 to toggle the log window instead of always showing it (to be consistent with  the Python console and Module panel)?

Punzo:
@lassoan fixed the action/shortcut in https://github.com/Slicer/Slicer/pull/7059/commits/7d8d9ce323e1092dd136374a19da1c753fe50b60

lassoan:
@Punzo This looks very nice. I have only one usability concern: docking the the log window to the bottom by default leads to a relatively narrow window. I can only see a few list entries and about 10 lines of log:

image

Docking to the side looks good:

image

We should either log to the side by default, but even better, it would be nice if we changed the layout so that if the log window is docked to the top or bottom, its layout would switch to horizontal:

image

It could be implemented by using a grid layout and adding a "layoutDirection" property to the ctkErrorLogWidget that is updated when the dock location is changed; or maybe even better, a ctkFlowLayout could be used to automatically switch between horizontal/vertical layout depending on the shape of the window.

pieper:
+1 for making the keyboard shortcut toggle the window.

👍 for making the layout nicer too

Punzo:
@lassoan @pieper I have applied the feedback, see:
https://github.com/commontk/CTK/pull/1100
https://github.com/Slicer/Slicer/pull/7059/commits/5ce1bbd18316e3e05b4cc35afa0aaae6567dafe9
https://github.com/Slicer/Slicer/pull/7059/commits/d7ff16bd4b24cc07d6fc46a6b4512ceba17544a9

lassoan (approved):
I've tested it and made tweaked the behavior a bit more (the "new error" notification is dismissed now when the user interacts with the log window; the log window is raised to the top when shown; etc.) and it all looks good to me now.

Build fails due to the required CTK update.

PR #7064, opened 2023-07-01
BUG: Fix "Build Slicer" CI workflow in the context of push event

jcfr:
This commit is a follow-up of 2d6d136e94 (ENH: Skip steps in required job rather than skipping workflow) that should ensure the "dorny/paths-filter" action works in the context of push event.

See https://github.com/dorny/paths-filter/issues/88#issuecomment-847003742


jcfr:
Fix worked as expected:

image

Issue #6648, opened 2022-11-06
Warping transform is not taken into account in volume rendering


lassoan:
## Summary

It is [documented](https://slicer.readthedocs.io/en/latest/user_guide/modules/volumerendering.html#limitations) that "The volume must not be under a warping (affine or non-linear) transformation. To render a warped volume, the transform must be hardened on the volume.". However, we cannot expect users to know about this limitation.

## Environment
- Slicer version: Slicer-5.1.0-2022-11-06
- Operating system: all

lassoan:
Potential solutions when the volume is under a warping transform:
- Option A: Disable showing the volume rendering 3D view and show a warning in Volume Rendering module that explains what the problem is and shows a button to harden the transform on the volume.
- Option B: Clone the volume internally, harden the warping transform on it, and volume render that. The warped volume could be stored in the vtkMRMLVolumeRenderingDisplayNode node so that it can be shared between all volume renderers; or in the pipeline in vtkMRMLVolumeRenderingDisplayableManager.

pieper:
I like option B, but maybe we should be more explicit since it will take resources and add time to computing.  That is, how would we know when it's okay to discard the clone?  Should the clone be saved with the scene?

How about when the user tries to make a volume under a non-linear transform visible Slicer won't display it, but offers a button to clone and harden the volume and display the clone instead.  This would then be explicit in the scene and the user could manage when it's no longer needed.  We could also keep a reference from the clone to the original so that it could be updated if the source volume or transform is modified.

pieper:
Maybe we should also make a Harden dialog that is something like a the geometry selection dialog in the segment editor or the interface of the Crop Volume module so that users can control the resampling parameters.

lassoan:
> That is, how would we know when it's okay to discard the clone

We already have a flag to automatically free up resources on hiding volume. It could be used for this, too.

We do this automatic internal cloning and hardening in the Segment Editor and nobody complained so far. Even for large volumes, the computation is not too long (up to maybe 10-20 seconds).

> How about when the user tries to make a volume under a non-linear transform visible Slicer won't display it, but offers a button to clone and harden the volume and display the clone instead

That's about what I meant by Option A. Would you display a popup window? We would need some GUI solution that works reasonably when the user drag-and-drops the image to the 3D view (or enables volume rendering in the right-click menu in the Subject Hierarchy eye icon).

[...]

PR #7065, opened 2023-07-04
BUG: Do not display volume rendering if volume is under non-linear transform

lassoan:
This commit just hides the volume and logs a message if a non-linearly transformed volume is attempted to be displayed using volume rendering.

This is a partial fix, as it would be better if the user was warned on the GUI in advance that the volume will not be displayed; and it would be even better to allow correct display of the volume regardless of what transform it is under. Still, this fix is useful as it prevents a volume from appearing in incorrect position.

see #6648


pieper (approved):
This seems logical to me.  I was thinking we could just harden the transform internally so that it renders as expected even if it's slow, but we can always add that in the future.

lassoan:
Thanks for the review. Implementing hardening internally would be better, but it would be a bit more work (we would need to expose the hardening API in a way that it could be used without modifying the existing node or creating a new node; we would also need to ensure that the resampling is done automatically when the volume or transform is changed, and not recomputed unnecessarily).

pieper:
I was also wondering if we can make a workaround recipe for this using the CLI autoexecute feature with one of the resampling CLIs.

PR #7074, opened 2023-07-08
ENH: Allow setting custom editor for .py files

lassoan:
When "Developer mode" is enabled in Application Settings, "Edit" button opens the module .py file in the default program associated with .py files.
This is most commonly an editor, but some users associate .py files with a Python launcher (such as "c:/windows/py.exe"). In these cases, the module .py file is attempted to be launched instead of edited.

This commit adds a new "Editor for .py files" option Application settings / Python. If a text editor is specified there then that program will be launched to edit .py files.
If user holds down Shift key while clicking on "Edit" button then an editor is not launched but instead the full path is copied to the clipboard to allow the user to open it in an editor manually.

Also changed icon of opening of .sliccerrc.py file (file open button usually opens a file selector dialog, replaced by "Go" button = arrow to the right).

### DIFF
--- a/Base/Python/slicer/ScriptedLoadableModule.py
+++ b/Base/Python/slicer/ScriptedLoadableModule.py
@@ -167,7 +167,8 @@ class ScriptedLoadableModuleWidget:
 
         # edit python source code
         self.editSourceButton = qt.QPushButton("Edit")
-        self.editSourceButton.toolTip = "Edit the module's source code."
+        self.editSourceButton.toolTip = ("Edit the module's source code. Shift+Click to copy the source code path to the clipboard."
+                                         " Default editor can be changed in Python section of Application Settings.")
         self.editSourceButton.connect('clicked()', self.onEditSource)
 
         self.editModuleUiButton = None
@@ -225,7 +226,26 @@ class ScriptedLoadableModuleWidget:
 
     def onEditSource(self):
         filePath = slicer.util.modulePath(self.moduleName)
-        qt.QDesktopServices.openUrl(qt.QUrl("file:///" + filePath, qt.QUrl.TolerantMode))
+
+        # On Shift+Click copy the module path instead of opening in editor
+        if slicer.app.keyboardModifiers() == qt.Qt.ShiftModifier:
+            absFilePath = os.path.abspath(filePath)
+            slicer.app.clipboard().setText(absFilePath)
+            slicer.util.delayDisplay(f"Module file path '{absFilePath}' is copied to clipboard")
+            return
+
+        editor = slicer.app.settings().value("Python/Editor")
+        if editor:
+            # User specified custom editor for .py files
+            import subprocess
+            try:
+                # Use startup environment to avoid Python environment issues with text editors implemented in Python
+                subprocess.Popen([editor, filePath], env=slicer.util.startupEnvironment())
+            except:
+                slicer.util.errorDisplay(f"Failed to open file:\n\n{filePath}\n\nusing editor: {editor}")
+        else:
+            # User default file association for opening .py file in editor
+            qt.QDesktopServices.openUrl(qt.QUrl("file:///" + filePath, qt.QUrl.TolerantMode))
 
 
 class ScriptedLoadableModuleLogic:
--- a/Base/QTGUI/qSlicerSettingsGeneralPanel.cxx
+++ b/Base/QTGUI/qSlicerSettingsGeneralPanel.cxx
@@ -133,8 +133,7 @@ void qSlicerSettingsGeneralPanelPrivate::init()
     context.evalScript(QString("slicerrcfilename = getSlicerRCFileName()\n"));
     QVariant slicerrcFileNameVar = context.getVariable("slicerrcfilename");
     this->SlicerRCFileValueLabel->setText(slicerrcFileNameVar.toString());
-    QIcon openFileIcon = QApplication::style()->standardIcon(QStyle::SP_DialogOpenButton);
-    this->SlicerRCFileOpenButton->setIcon(openFileIcon);
+    this->SlicerRCFileOpenButton->setIcon(QIcon(":Icons/Go.png"));
     QObject::connect(this->SlicerRCFileOpenButton, SIGNAL(clicked()), q, SLOT(openSlicerRCFile()));
     }
   else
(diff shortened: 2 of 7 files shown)


pieper (approved):
Looks good - thanks! 👍

Base/Python/slicer/ScriptedLoadableModule.py
```diff
@@ -167,7 +167,8 @@ def createHLayout(elements):
 
         # edit python source code
         self.editSourceButton = qt.QPushButton("Edit")
-        self.editSourceButton.toolTip = "Edit the module's source code."
+        self.editSourceButton.toolTip = ("Edit the module's source code. Shift+Click to copy the source code path to the clipboard."
+                                         " Default editor can be changed in Python section of Application Settings.")
```
jcfr, line 171:
```suggestion
                                         " The default editor can be changed in the Python section of the Application Settings.")
```

Base/Python/slicer/ScriptedLoadableModule.py
```diff
@@ -225,7 +226,26 @@ def onReloadAndTest(self, **kwargs):
 
     def onEditSource(self):
         filePath = slicer.util.modulePath(self.moduleName)
-        qt.QDesktopServices.openUrl(qt.QUrl("file:///" + filePath, qt.QUrl.TolerantMode))
+
+        # On Shift+Click copy the module path instead of opening in editor
```
jcfr, line 230:
```suggestion
        # On Shift+Click copy the module path instead of opening it in the editor
```

Base/Python/slicer/ScriptedLoadableModule.py
```diff
@@ -225,7 +226,26 @@ def onReloadAndTest(self, **kwargs):
...
+        if slicer.app.keyboardModifiers() == qt.Qt.ShiftModifier:
+            absFilePath = os.path.abspath(filePath)
+            slicer.app.clipboard().setText(absFilePath)
+            slicer.util.delayDisplay(f"Module file path '{absFilePath}' is copied to clipboard")
+            return
+
+        editor = slicer.app.settings().value("Python/Editor")
+        if editor:
+            # User specified custom editor for .py files
```
jcfr, line 239:
```suggestion
            # User specified a custom editor for .py files
```

Base/Python/slicer/ScriptedLoadableModule.py
```diff
@@ -225,7 +226,26 @@ def onReloadAndTest(self, **kwargs):
...
+            import subprocess
+            try:
+                # Use startup environment to avoid Python environment issues with text editors implemented in Python
```
jcfr, line 242:
```suggestion
                # Use the startup environment to avoid Python environment issues with text editors implemented in Python
```

Base/Python/slicer/ScriptedLoadableModule.py
```diff
@@ -225,7 +226,26 @@ def onReloadAndTest(self, **kwargs):
...
+                subprocess.Popen([editor, filePath], env=slicer.util.startupEnvironment())
+            except:
+                slicer.util.errorDisplay(f"Failed to open file:\n\n{filePath}\n\nusing editor: {editor}")
+        else:
+            # User default file association for opening .py file in editor
```
jcfr, line 247:
```suggestion
            # Rely on the default file association for opening the .py file
```

Base/QTGUI/qSlicerSettingsGeneralPanel.cxx
```diff
@@ -249,7 +248,22 @@ void qSlicerSettingsGeneralPanel::openSlicerRCFile()
...
+
+  QString editor = qSlicerApplication::application()->userSettings()->value("Python/Editor").toString();
+  if (editor.isEmpty())
+    {
+    QDesktopServices::openUrl(QUrl("file:///" + slicerRcFileName, QUrl::TolerantMode));
+    }
+  else
+    {
+    QProcess process;
+    // Use startup environment to avoid Python environment issues with text editors implemented in Python
```
jcfr, line 260:
```suggestion
    // Use the startup environment to avoid Python environment issues with text editors implemented in Python
```

Docs/developer_guide/python_faq.md
```diff
@@ -296,6 +296,16 @@ In most places, unix-type separators can be used instead of backslash. This is c
 
 See more information in Python documentation: https://docs.python.org/3/tutorial/introduction.html?#strings
 
+## How to modify a Python scripted module
+
+If `Developer mode` is enabled in application settings then `Reload and Test` section is displayed at the top of user interface of Python scripted modules. This section contains buttons for convenient editing of the module source code (`.py` file) and user interface (`.ui` file). Clicking the `Edit` button opens the module source code in the program associated with `.py` files, as defined in operating system settings. This behavior can be overridden by specifying a text editor (such as VS Code, Sublime Text, ...) in application settings: `Editor for .py files` in `Python` section.
```
jcfr, line 301:
```suggestion
If the `Developer mode` is enabled in the application settings then a `Reload and Test` section is displayed at the top of the user interface of Python scripted modules. This section contains buttons for convenient editing of the module source code (`.py` file) and user interface (`.ui` file). By default, clicking on the `Edit` button opens the module source code in the program associated with `.py` files, as defined in the operating system settings. This behavior can be overridden by specifying a text editor (such as VS Code, Sublime Text, ...) in the application settings: `Editor for .py files` in the `Python` section.
```

Docs/developer_guide/python_faq.md
```diff
@@ -296,6 +296,16 @@ In most places, unix-type separators can be used instead of backslash. This is c
...
+
+On Windows, VS Code text editor is installed by default at:
+
+```txt
+C:/Users/YourUserName/AppData/Local/Programs/Microsoft VS Code/Code.exe
+```
```
jcfr, line 307:
````suggestion
:::{tip}
On Windows, VS Code text editor is installed by default at:

```txt
C:/Users/YourUserName/AppData/Local/Programs/Microsoft VS Code/Code.exe
```
:::
````

jcfr:
I should have clicked "Request change" instead of "Comment". That said, the comments may be still be reviewed and integrated independently.

Docs/developer_guide/python_faq.md
lassoan: Thanks for the suggestions @jcfr. An your explain why "the" is needed in "If the `Developer mode` is enabled"?

jcfr: It initially felt more natural ... and your question prompted me to further look into it. See https://learnenglish.britishcouncil.org/grammar/english-grammar-reference/definite-article

Looking at this page `Developer Mode` is use without `the`.
See https://developer.apple.com/documentation/xcode/enabling-developer-mode-on-a-device

Looking at the current Slice code base, there is one occurrence of `when the [developer mode]...` and all the other ones are `developer mode` all lower cased (unless at the start of a sentence)

Most of the occurrences at `developer mode` with a few `Developer mode`.

For sake of consistency with how we define module panel[^1], I think we should use `Developer Mode`

[^1]: https://www.slicer.org/wiki/Documentation/Nightly/Developers/Style_Guide/UI#Text

PR #7077, opened 2023-07-09
ENH: Load DICOM files using DICOM module by default

### LINKED ISSUE #5726: Disable DICOM import from "Add data" dialog
DICOM import via "Add data" dialog has many limitations, such as cannot deal with varying slice spacing, 4D volumes, cannot import any other objects than images, etc. and has issues, such as potentially loading the image in incorrectly slice order.

DICOM loading this way should be disabled and preferable replaced by importing via the DICOM module. To match the previous functionality more closely, the import should be done via a temporary DICOM database and should be fast (at least time to first image should be quick and simple).

## Environment
- Slicer version: Slicer-4.11 and later
- Operating system: all

lassoan:
Drag-and-dropping a DICOM file made Slicer use ITK DICOM reader, which often did not load the data correctly.

This commit adds a new "DICOM import" reader type, which is selected by default for DICOM files. This reader imports the files into the Slicer DICOM database and switches to the DICOM module.

The workaround of interpreting all files as DICOM by default while in DICOM module has now been removed. "Add data" window is brought up when loading a file, regardless of what is the active module. The workaround surprised users when they could not load non-DICOM data while they were in DICOM module.

If a user wants to read the DICOM file using ITK reader, it is possible by changing "Description" in the "Add data" module to "Volume".

fixes #5726

### DIFF
--- a/Base/QTCore/qSlicerScriptedFileReader.cxx
+++ b/Base/QTCore/qSlicerScriptedFileReader.cxx
@@ -278,7 +278,7 @@ double qSlicerScriptedFileReader::canLoadFileConfidence(const QString& file)cons
     return this->Superclass::canLoadFileConfidence(file);
     }
 
-  if (PyFloat_Check(result))
+  if (!PyFloat_Check(result))
     {
     qWarning() << d->PythonSource
                << " - In" << d->PythonClassName << "class, function 'canLoadFileConfidence' "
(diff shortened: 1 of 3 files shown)


pieper (approved):
This looks good.  I have often been confused that data drops were rejected when the dicom module was active, and similarly there's been confusion about the difference between Add Data and the DICOM module.  This nicely fixes both.

PR #7080, opened 2023-07-11
BUG: Update VTK to fix regression preventing from loading older vtp files

jcfr:
This ensures that reading an empty "vtkPolyData" reports a warning instead of an error message.

List of VTK changes:

```
$ git shortlog 6c02c9b376..8e0f1f9af7 --no-merges
      [Backport MR-10295] Allow const std::vector<vtkSmartPointer<T>> in Python (part 2)

Steve Pieper (1):
      [Backport MR-9648] BUG: error for non-fatal xml issue
```


pieper (approved):
👍

PR #7078, opened 2023-07-10
ENH: Add flip and rotate buttons to Reformat module

lassoan:
Users quite often ask about how to flip or rotate slice views. Slicers in Reformat module were suitable for this but were not very convenient or intuitive.

This commit adds "Flip H", "Flip V", "Rotate CW", "Rotate CCW" buttons to the Orientation section of the Reformat module. Internally they just rotate the slice view by 90 or 180 degrees around various slice axes. "Rotate to volume plane" button is also added for completeness.

Normal direction setting buttons and slice orientation sliders were renamed according to their actual function (normal vector is set to LR, PA, IS anatomical directions - not X, Y, Z; rotation happens around horizontal, vertical, normal slice axes - not around LR, PA, IS anatomical directions).

image

PR #7082, opened 2023-07-11
BUG: Fix SlicerOrientationSelectorTestTest to account for widget rename

jcfr:
This commit is a follow-up of e75785816 (ENH: Add flip and rotate buttons to Reformat module (#7078)) accounting for the slide widget rename introduced in:
* https://github.com/Slicer/Slicer/pull/7078


lassoan (approved):
It looks good, thank you for fixing this.

PR #7084, opened 2023-07-11
BUG: Fix behavior after invisible segment confirmation popup

### LINKED ISSUE #6748: Mouse behavior is incorrect after "Operate on invisible segment" popup
## Summary

Mouse behavior is incorrect after initiating editing operation on a hidden segment and the "Operate on invisible segment" popup is displayed.

Reported at: https://discourse.slicer.org/t/3d-rotation-without-mouse-press-after-popup-operate-on-invisible-segment/26889

## Steps to reproduce

- Download Sample Data CTChest.
- Create a segment
- Hide the segment
- Activate Scissor effect
- Click in the 3D view
- Popup appears: “Operate on invisible segment?”
- Press cancel.
- Move cursor in 3D view (without mouse button press)

The 3D view rotates as the mouse is moved around.

## Expected behavior

After the popup is dismissed, the view should work correctly (only rotates if the mouse button is depressed).

## Workaround

After the popup is displayed, the proper mouse behavior can be achieved by deactivating the effect and clicking in the view. 

## Environment
- Slicer version: Slicer-5.2.1
- Operating system: all

lassoan:
When the user starts a segment editing operation on an invisible segment a popup is displayed to confirm that it is intentional (usually not). This popup left some effects in inconsistent state (3D view rotation is initiated, some feedback actors were not cleaned up). This commit avoids these inconsistent states.

fixes #6748

### DIFF
--- a/Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorDrawEffect.py
+++ b/Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorDrawEffect.py
@@ -63,13 +63,19 @@ class SegmentEditorDrawEffect(AbstractScriptedSegmentEditorLabelEffect):
         anyModifierKeyPressed = callerInteractor.GetShiftKey() or callerInteractor.GetControlKey() or callerInteractor.GetAltKey()
 
         if eventId == vtk.vtkCommand.LeftButtonPressEvent and not anyModifierKeyPressed:
+
             # Make sure the user wants to do the operation, even if the segment is not visible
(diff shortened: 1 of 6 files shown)


jcfr (commented):
Should the template be updated to include support for aborting ? Or should such support be moved into the base class ?

jcfr (approved):

jcfr:
Note: I only did a cursory check and didn't try to run after applying the proposed patches.

lassoan:
> Should the template be updated to include support for aborting ? Or should such support be moved into the base class ?

Good idea, I've added it to the template.

jcfr (approved):

PR #7089, opened 2023-07-13
BUG: support color DICOM data series

pieper:
Without this change the vtkITK code complained that a series of color dicom data could not be loaded.

It appears that when the check for grayscale was added it was only tested on single instance files, not on series in multiple files.

This change invokes the correct reader based on whether it's a single file or not.

### DIFF
--- a/Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
+++ b/Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
@@ -380,7 +380,10 @@ class DICOMScalarVolumePluginClass(DICOMPlugin):
         if grayscale:
             reader = vtkITK.vtkITKArchetypeImageSeriesScalarReader()
         else:
-            reader = vtkITK.vtkITKArchetypeImageSeriesVectorReaderFile()
+            if len(files) > 1:
+                reader = vtkITK.vtkITKArchetypeImageSeriesVectorReaderSeries()
+            else:
+                reader = vtkITK.vtkITKArchetypeImageSeriesVectorReaderFile()
         reader.SetArchetype(files[0])
         for f in files:
             reader.AddFileName(f)


pieper:
@lassoan it looks like you added color support to the scalar volume plugin with https://github.com/Slicer/Slicer/commit/9e56d075ee7c9edf5803ebe7307d72da070db76b#diff-ddb2464152946e0fec0e7fcd1ec6cc4aefcd6b8a9ad3ee82a136564525f0c66f, but it wasn't working for a conversion of the visible human male photos, so based on the context I think you probably only tested with a single-file color instance, perhaps a multiframe ultrasound?

The patch here works to load the series of RGB dicom instances.  Can you test that it also works for the data that you originally were using?

pieper:
fyi @fedorov

lassoan:
My plugin is for reading 2D+t (grayscale or color) ultrasound, X-ray, MRI as volume sequence. Do you think my plugin can/should use somehow the class that you've just fixed?

pieper:
> My plugin is for reading 2D+t (grayscale or color) ultrasound, X-ray, MRI as volume sequence.

Which plugin are you referring to?  My change is for the DICOMScalarVolumePlugin in the core.  This change should allow it to read a series of RGB instance while still supporting the other cases you listed.

fedorov:
@pieper I hope this will also correctly parse the geometry of the RGB series?

lassoan:
> > My plugin is for reading 2D+t (grayscale or color) ultrasound, X-ray, MRI as volume sequence.
> 
> Which plugin are you referring to?

DICOMImageSequencePlugin

> This change should allow it to read a series of RGB instance while still supporting the other cases you listed.

I worked on DICOMImageSequencePlugin, which only needed loading of single-slice color images. I also added this feature to the DICOMScalarVolumePlugin, but I did not try to load multi-slice color volumes, so it is great that you added this.

DICOMImageSequencePlugin does not need multi-slice support, so I think your change is good as is (no changes are needed in DICOMImageSequencePlugin).

lassoan (approved):
THank you, the change looks good to me.

In the long term, we could consider merging all these DICOM readers into a single class.

pieper:
> @pieper I hope this will also correctly parse the geometry of the RGB series?

I did some debugging on the VHP dataset and know why it loads the way it does in Slicer.

It turns out that the GDCM code that Slicer uses via ITK explicitly sets the direction cosines to identity for secondary captures:

https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/ThirdParty/GDCM/src/gdcm/Source/MediaStorageAndFileFormat/gdcmImageHelper.cxx#L731-L740

And for secondary captures, he ignores PixelSpacing and instead uses NominalScannedPixelSpacing:

https://github.com/InsightSoftwareConsortium/ITK/blob/master/Modules/ThirdParty/GDCM/src/gdcm/Source/MediaStorageAndFileFormat/gdcmImageHelper.cxx#L1265-L1273

I'm checking with @dclunie to see if he agrees that this is incorrect and we should change ITK.  We could make a PR to ITK proper, but also could put the fix in Slicer's fork to get it in quicker.

pieper:
> In the long term, we could consider merging all these DICOM readers into a single class.

I agree, now that we have more experience with this we really should refactor all this.  I'd like to see all the logic around understanding the dicom structure split out into a separate place, possibly highdicom.  These plugins could then just be responsible for mapping high level dicom concepts into the corresponding Slicer concepts.

lassoan:
I agree that the best would be have this logic outside Slicer to maximize the impact, and nowadays it would need to be in Python. Image reconstruction would be a completely new territory for highdicom, but maybe Chris would be OK to add such logic in the package.

dclunie:
FYI, here is another secondary capture slice-by-slice DICOM dataset that adds ImagePositionPatient and ImageOrientationPatient (greyscale ultrasound) "https://piurimaging.com/files/DICOM/MedianNerve_Volume.zip" and probably triggers the same ITH and Slicer issues

dclunie:
I agree that:
1. GDCM should not ignore ImageOrientationPatient if it is present (as opposed to supplying a default value if not)
2. GDCM should not ignore PixelSpacing if present (as opposed to always using NominalScannedPixelSpacing) - indeed use of PixelSpacing for SC when "calibrated" has long been standardized behavior ([CP 586](https://dicom.nema.org/medical/dicom/Final/cp586_ft.pdf))

pieper:
Okay, thanks @dclunie I'll work on fixing GDCM/ITK.

In the meantime I found another issue with ITK's DCMTK reader that prevents that from working on this data (or apparently any data that has a preamble).

pieper:
See:

https://github.com/InsightSoftwareConsortium/ITK/issues/4108

and

https://github.com/InsightSoftwareConsortium/ITK/issues/4109

pieper:
Fixes for the two issues:

https://github.com/InsightSoftwareConsortium/ITK/pull/4110

https://github.com/InsightSoftwareConsortium/ITK/pull/4111

pieper:
Once the two PRs get merged the VHP data will load in Slicer using either DCMTK or GDCM (confirmed on my local build).  Since it may take a while to get them accepted upstream we should merge them into Slicer's ITK fork.  I'm hoping @jcfr can help me with that when he has a chance.

pieper:
> Thanks for submitting merge request in upstream ITK, waiting these are integrated in ITK proper, I will move forward with backporting into our `Slicer/ITK` fork.

Just FYI @jcfr both PRs will require a little tweaking based on feedback from the community so I'll let you know when they are ready to be backported.  I hope to get to it tomorrow, but if not sometime next week.

PR #7091, opened 2023-07-14
BUG: Fix CPACK_NSIS_INSTALL_SUBDIRECTORY being overwritten

jamesobutler:
Introduced by https://github.com/Slicer/Slicer/commit/43dd44e85c210fc09375997bd435da61adbe136b

I was using this along with `mark_as_superbuild(Slicer_CPACK_NSIS_INSTALL_SUBDIRECTORY "${Slicer_ORGANIZATION_NAME}")` in the CMakeLists.txt of my Slicer custom app to change the default system level install location from "C:/Program Files/{ApplicationName}" to "C:/Program Files/{OrganizationName}/{ApplicationName}

### DIFF
--- a/CMake/SlicerCPack.cmake
+++ b/CMake/SlicerCPack.cmake
@@ -354,7 +354,9 @@ endif()
 # -------------------------------------------------------------------------
 if(CPACK_GENERATOR STREQUAL "NSIS")
 
-  set(Slicer_CPACK_NSIS_INSTALL_SUBDIRECTORY "")
+  if(NOT DEFINED Slicer_CPACK_NSIS_INSTALL_SUBDIRECTORY)
+    set(Slicer_CPACK_NSIS_INSTALL_SUBDIRECTORY "")
+  endif()
   slicer_cpack_set("CPACK_NSIS_INSTALL_SUBDIRECTORY")
 
   set(_nsis_install_root "${Slicer_CPACK_NSIS_INSTALL_ROOT}")
--- a/CMake/SlicerCPack.cmake
+++ b/CMake/SlicerCPack.cmake
@@ -354,6 +354,9 @@ endif()
 # -------------------------------------------------------------------------
 if(CPACK_GENERATOR STREQUAL "NSIS")
 
+  # Conditionally set the variable so that a different value
+  # can be passed to the inner build (usually done by a custom
+  # application).
   if(NOT DEFINED Slicer_CPACK_NSIS_INSTALL_SUBDIRECTORY)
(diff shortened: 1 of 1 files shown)


CMake/SlicerCPack.cmake
```diff
@@ -354,7 +354,9 @@ endif()
 # -------------------------------------------------------------------------
 if(CPACK_GENERATOR STREQUAL "NSIS")
 
-  set(Slicer_CPACK_NSIS_INSTALL_SUBDIRECTORY "")
+  if(NOT DEFINED Slicer_CPACK_NSIS_INSTALL_SUBDIRECTORY)
+    set(Slicer_CPACK_NSIS_INSTALL_SUBDIRECTORY "")
```
lassoan, line 358: Is there a difference in behavior if `Slicer_CPACK_NSIS_INSTALL_SUBDIRECTORY` is undefined vs. empty?

If yes, then we should described that in a comment.

If not (in most cases this is so, because the string value of an undefined CMake variable is `""`) then it would be simpler to just remove this line and add a comment that mentioning that `Slicer_CPACK_NSIS_INSTALL_SUBDIRECTORY` may be set externally.

jcfr: Being explicit makes sense as it is consistent with the `slicer_cpack_set` reporting an error if the variable is not defined.

jcfr (dismissed):

CMake/SlicerCPack.cmake
lassoan: That's good it just need to be described in a comment. Otherwise this line may get removed during future refactoring.

CMake/SlicerCPack.cmake
```diff
@@ -354,7 +354,9 @@ endif()
 # -------------------------------------------------------------------------
 if(CPACK_GENERATOR STREQUAL "NSIS")
 
-  set(Slicer_CPACK_NSIS_INSTALL_SUBDIRECTORY "")
+  if(NOT DEFINED Slicer_CPACK_NSIS_INSTALL_SUBDIRECTORY)
```
jcfr, line 357:
```
  # Conditionally set the variable so that a different value
  # can be used to the inner build in the context of custom application.
  if(NOT DEFINED Slicer_CPACK_NSIS_INSTALL_SUBDIRECTORY)
```

CMake/SlicerCPack.cmake
jcfr: See suggestion above

CMake/SlicerCPack.cmake
lassoan: Looks good, thank you!

PR #7096, opened 2023-07-14
ENH: Save DICOM instance UID in image sequence plugin

ungi:


jcfr (approved):
Could someone also review this similar change I proposed for SlicerHeart ?

https://github.com/SlicerHeart/SlicerHeart/pull/28

I think we should also provide a constant in Slicer instead of an hard-coded string.

lassoan:
> Could someone also review this similar change I proposed for SlicerHeart ?

I've reviewed it, I just had a comment/question before merging.

PR #7098, opened 2023-07-15
BUG: Fix rendering of SaveDataDialog checkboxes in Dark Mode

### LINKED ISSUE #5185: Difficult to see checkboxes within a table in Dark Mode
Originally from https://discourse.slicer.org/t/selection-boxes-in-save-dialog-box-is-invisible-in-dark-mode/13442

image

It appears in light mode that border of the checkbox is #c8c8c8 and based on the values specified for Dark Mode, this color appears to be defined by QPalette::Mid (see below). 

**Slicer Light**
Oddly in the Slicer light style Mid and Light are the same value when they shouldn’t be. Read about QPalette [ColorRoles](https://doc.qt.io/qt-5/qpalette.html#ColorRole-enum).
https://github.com/Slicer/Slicer/blob/c565268cf68c09b54e0e65fbfed5bd5faa404401/Base/QTGUI/qSlicerStyle.cxx#L228-L232
**Slicer Dark**
https://github.com/Slicer/Slicer/blob/c565268cf68c09b54e0e65fbfed5bd5faa404401/Base/QTGUI/qSlicerStyle.cxx#L264-L268

It is harder to see in dark mode because QPalette::Mid is closer to QPalette::Base(the filled color), than it is in light mode. This is a weird case of a widget (QCheckbox) that is using the color of QPalette::Base even though it is within another widget (QTableWidget) which also uses QPalette::Base by default for its contents. Setting it to use the QPalette::AlternateBase could potentially work assuming that this widget continues to not use alternatingRowColors.

AaronGee1:
Set a stylesheet that sets the unchecked background to alternate base for QCheckBox in options column and QAbstractItemView in filename column.

fixes #5185

References:
* https://doc.qt.io/qt-5/stylesheet-reference.html#paletterole
* https://doc.qt.io/qt-5/stylesheet-reference.html#list-of-stylable-widgets

### DIFF
--- a/Base/QTGUI/qSlicerSaveDataDialog.cxx
+++ b/Base/QTGUI/qSlicerSaveDataDialog.cxx
@@ -120,6 +120,14 @@ qSlicerSaveDataDialogPrivate::qSlicerSaveDataDialogPrivate(QWidget* parentWidget
 
   this->FileWidget->setItemDelegateForColumn(
     FileNameColumn, new qSlicerFileNameItemDelegate(this));
+  this->FileWidget->setStyleSheet(QStringLiteral(
+    "QAbstractItemView::indicator:unchecked {\n"
+    "   background-color: palette(alternate-base);\n"
+    "}\n"
+    "QCheckBox::indicator:unchecked {\n"
+    "   background-color: palette(alternate-base);\n"
+    "};"
+  ));
   this->FileWidget->verticalHeader()->setVisible(false);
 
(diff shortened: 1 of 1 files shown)


jcfr:
Before/after preview is available here: https://github.com/Slicer/Slicer/issues/5185#issuecomment-1146811385

jcfr:
| Before | After |
|--|--|
| image | image |
| image | image |

PR #7102, opened 2023-07-17
Fix "Model",  "plane markups" and "segmentation closed surface" slice intersections display

### LINKED ISSUE #7101: "Model", "Plane Markups" and "Segmentation Closed Surface"  Slice intersections are broken
## Summary

Model slice intersections worked, at least as of Slicer-5.3.0-2023-04-23 but does not work at least as of Slicer-5.3.0-2023-07-13 and I can reproduce the issue with my local build.  Issue reported by Zora Kikinis with FiberBundles in with SlicerDMRI, but the same issue happens for any model node.

**Update**: Similar issue can be observed with either the Plane markups or a segmentation wotj its master representation being a closed surface. See https://github.com/Slicer/Slicer/pull/7102#pullrequestreview-1533562575

## Steps to reproduce

* Download sample data
* Segment, e.g. sphere brush, and make visible in 3D
* Export to models
* Go to Model Display and enable slice visibility

## Expected behavior

Models should display in slice view but it doesn't.

Instead we get these error messages:
```
Input for connection index 0 on input port index 0 for algorithm vtkCompositeDataGeometryFilter (0x600003b17000) is of type vtkPolyData, but a vtkCompositeDataSet is required.
```

The message is being generated at this call:

https://github.com/Slicer/Slicer/blob/3acf6800e3578015ffd4d23404fcd1ff9bd1337d/Libs/MRML/DisplayableManager/vtkMRMLModelSliceDisplayableManager.cxx#L482

It appears that with the use of this filter was introduced by this commit:

https://github.com/Slicer/Slicer/commit/72914989516fbb7e593e4e1fe1894d65b2c93dab

But something must have changed in a recent version of VTK and needs to be investigated.

## Environment
See summary

pieper:
With the current VTK a `vtkCompositeDataGeometryFilter` cannot be used with a `vtkPolyData` so we use a `vtkGeometryFilter` instead.

Fixes:
* #7101

### DIFF
--- a/Libs/MRML/DisplayableManager/vtkMRMLModelSliceDisplayableManager.cxx
+++ b/Libs/MRML/DisplayableManager/vtkMRMLModelSliceDisplayableManager.cxx
@@ -58,7 +58,7 @@
 #include <vtkWeakPointer.h>
 
 // VTK includes: customization
-#include <vtkCompositeDataGeometryFilter.h>
+#include <vtkGeometryFilter.h>
 #include <vtkPlaneCutter.h>
 #include <vtkSampleImplicitFunctionFilter.h>
 
@@ -84,7 +84,7 @@ public:
     vtkSmartPointer<vtkTransformFilter> ModelWarper;
     vtkSmartPointer<vtkPlane> Plane;
     vtkSmartPointer<vtkPlaneCutter> Cutter;
-    vtkSmartPointer<vtkCompositeDataGeometryFilter> GeometryFilter; // appends multiple cut pieces into a single polydata
+    vtkSmartPointer<vtkGeometryFilter> GeometryFilter;
     vtkSmartPointer<vtkSampleImplicitFunctionFilter> SliceDistance;
     vtkSmartPointer<vtkProp> Actor;
     };
@@ -311,7 +311,7 @@ void vtkMRMLModelSliceDisplayableManager::vtkInternal
   Pipeline* pipeline = new Pipeline();
   pipeline->Actor = actor.GetPointer();
   pipeline->Cutter = vtkSmartPointer<vtkPlaneCutter>::New();
-  pipeline->GeometryFilter = vtkSmartPointer<vtkCompositeDataGeometryFilter>::New();
+  pipeline->GeometryFilter = vtkSmartPointer<vtkGeometryFilter>::New();
   pipeline->SliceDistance = vtkSmartPointer<vtkSampleImplicitFunctionFilter>::New();
   pipeline->TransformToSlice = vtkSmartPointer<vtkTransform>::New();
   pipeline->NodeToWorld = vtkSmartPointer<vtkGeneralTransform>::New();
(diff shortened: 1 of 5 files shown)


lassoan (commented):
It's great that you could figure this out so quickly. Do you know if this is a change in vtkPlaneCutter or in vtkCompositeDataGeometryFilter? It looks more like a vtkPlaneCutter refactoring that again broke backward compatibility. VTK should really improve its processes so that these things don't happen out of the blue.

Have you tested if this change works with earlier VTK versions? If not then it could be nice to keep backward VTk compatibility for a little while by making this change conditional on `#if (VTK_VERSION_NUMBER >= VTK_VERSION_CHECK(9, 2, 20230212))`

Probably all these are broken, too:
- Modules\Loadable\Segmentations\MRMLDM\vtkMRMLSegmentationsDisplayableManager2D.cxx
- Modules\Loadable\Markups\VTKWidgets\vtkSlicerPlaneRepresentation2D.cxx
- Libs\vtkSegmentationCore\vtkBinaryLabelmapToClosedSurfaceConversionRule.cxx

No need to fix them in this pull request, so if you don't have time to look at them now then just add a bug report so that we don't forget about it.

pieper:
I don't think we should complicate the code with version checks since the superbuild pins the dependency and nobody should really be building Slicer with other VTK versions.

 As far as I can tell those other features are all working fine without the error message so it must have been something specific to the model pipeline.  Probably due to the different inputs: https://github.com/Kitware/VTK/blob/master/Filters/Core/vtkPlaneCutter.h#L42-L48

jcfr:
The regression was likely introduced by https://github.com/Kitware/VTK/commit/f36cb499c19fd7ba0ad35f94236c6a2bab3432cd (vtkPlaneCutter: Change output type for vtkUniformGridAMR)



@spyridon97 Do you anticipate any issue with us transitioning our pipeline from using `vtkGeometryFilter` instead of `vtkCompositeDataGeometryFilter` ?

spyridon97:
If I remember correctly, vtkCompositeDataGeometryFilter uses vtkGeometryFilter or vtkDataSetSurfaceFilter to extract the surface and and then merges the result of each block into one polydata. If you execute the vtkGeometryFilter or vtkDataSetSurfaceFilter on a composite dataset, it will create the surface of each block and the result will be a composite dataset instead of polydata.

vtkPlaneCutter broke backwards compatibility because it didn't really work for vtkUniformGridAMR, and because of the introduction of vtkPartitionedDatasetCollection.

jcfr:
Thanks for the feedback @spyridon97 

Considering the change was introduced in a minor release, for the sake of backward compatibility, do you think the support for `vtkPolyData` could be re-introduced (for example by delegating to `vtkGeometryFilter`) ?

Ideally, a runtime deprecation warning with guidance on which filter to use instead would be great.

jcfr:
## vtkSlicerPlaneRepresentation2D

After applying a similar fix, the plane is correctly intersected with the slice view:

| Before | After |
|--|--|
| image | image |

jcfr:
## vtkMRMLSegmentationsDisplayableManager2D

After applying a similar fix, the "closed surface" segmentation is also correctly intersected with the slice view:

| Before | After |
|--|--|
| image | image |

PR #7114, opened 2023-07-20
DOC: Add Fedora 37+ launch troubleshooting

Villafruela:
- Add checklist to fix execution issues: verify rights, set +x, run from terminal.
- Document workaround for "Failed to obtain launcher executable name!" by moving binary to user’s home Software folder.

---

_Original description updated by @jcfr_

### DIFF
--- a/Docs/user_guide/getting_started.md
+++ b/Docs/user_guide/getting_started.md
@@ -133,20 +133,42 @@ The included libcrypto.so.1.1 in the Slicer installation is incompatible with th
     $SLICER_ROOT/lib/Slicer-5.xx/libcrypto.*
 :::
 
-If you double-clicked on the app and it doesn't launch, you can try this (in order):
-- verify execution rights of the app (Right click on the app > properties > executable as a program checked) or in terminal with :
+If double-clicking the app does not launch it, try the following steps in order:
 
-    ls -la
+1. Verify execution rights of the app
 
-You should have something like "-rwx-r-xr-x" at the beginning of the line of the executable. Sometimes certain unrar/unzip applications requires you to decompress the archive in sudo mode (in order to restore the rights scheme on the files of the archive). If you do not have execution rights you can add them with this command in a terminal
+   - Right-click on the app → *Properties* → make sure **“Executable as a program”** is checked.
+   - Or check from a terminal:
 
-    chmod +x Slicer
+     ```shell
+     ls -la
+     ```
 
-- then, launch the app from terminal :
+     You should see something like `-rwxr-xr-x` at the beginning of the line for the executable.
 
-    ./Slicer
+     :::{note}
+     Some archive managers (e.g., certain `unrar`/`unzip` tools) may require running extraction with `sudo` to restore the correct file permissions.
+     :::
 
-- if you still get the "error: Failed to obtain launcher executable name !" message, try to move the executable in your Software Home folder and double click on the app, it should launch.
+     If execution rights are missing, add them with:
+
+     ```shell
+     chmod +x Slicer
+     ```
+
+2. Launch from the terminal
+
+   ```shell
+   ./Slicer
+   ```
+
+3. If you get the error
+
+   ```
+   error: Failed to obtain launcher executable name !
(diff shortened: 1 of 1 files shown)


Docs/user_guide/getting_started.md
```diff
@@ -123,6 +123,15 @@ The included libcrypto.so.1.1 in the Slicer installation is incompatible with th
 
     $SLICER_ROOT/lib/Slicer-4.xx/libcrypto.*
 
+If you double-clicked on the app and it doesn't launch, you can try this (in order):
+- verify execution rights of the app (Right click on the app > properties > executable as a program checked) or in terminal, in the app folder:
+    
+    chmod +x Slicer
```
pieper, line 129: This shouldn't be needed since the file will be executable after extracting from the tar file.  Did you ever have it happen that this was not the case?

Docs/user_guide/getting_started.md
```diff
@@ -123,6 +123,15 @@ The included libcrypto.so.1.1 in the Slicer installation is incompatible with th
...
+- launch the app from terminal :
+    
+    ./Slicer
+- if you get the "error: Failed to obtain launcher executable name !" message : move the app in your Software Home folder and double click on the app, it should launch.
```
pieper, line 133: It would help if you could provide more context for this.  Is there something special on Fedora that files are downloaded by default to a place where they can't be executed?

pieper (changes requested):
Thank you for suggesting these changes 👍 

We want the documentation to be complete and helpful.  Since differences across linux versions can be a complex topic we want to be sure we are specifically addressing issues in the correct spot (e.g. what's unique to Fedora vs what's generic to linux).

Docs/user_guide/getting_started.md
Villafruela: Some unzip software have special usage. Like Info-Zip 3.0 for instance, where you need to unzip as root if you want to restore the rights scheme on the files. Some also configure their unzip software to restore files only in 666 (rw-rw-rw-) for security reasons. So you can't pass this option to troubleshoot i think (or at least you have to check that permissions are 755...)

Docs/user_guide/getting_started.md
Villafruela: On fedora, only User's Home folders get execution rights. But for some reason, i do not know why but i experienced it : if you unpack the files in the download folder : you get the error message, if you unpack in the "Applications" folder, it just work fine... That's why i suggested the addition to the documentation.

Docs/user_guide/getting_started.md
pieper: Thanks, that makes more sense.  Maybe you can make a small edit to the documentation text summarizing this.

Docs/user_guide/getting_started.md
pieper: This also makes sense and would be nice to mention in the text.  Maybe some documentation about installing applications in these systems.

Docs/user_guide/getting_started.md
Villafruela: Sure i'll be happy to...

Villafruela: Would you be interested in a french translation of the documentation ?

pieper (dismissed):
LGTM 👍

Docs/user_guide/getting_started.md
pieper: > Would you be interested in a french translation of the documentation ?

Yes!  In fact, there's a project devoted to that and any help would be greatly appreciated!

There is an internationalization group and infrastructure that you can learn about here: 

https://projectweek.na-mic.org/PW39_2023_Montreal/Projects/3DSlicerInternationalization/

https://chanzuckerberg.com/eoss/proposals/3d-slicer-in-my-language-internationalization-and-usability-improvements/

The first fully-French tutorial was recently held, but there is a lot more to be done:

https://sliceriml.github.io/

@spujol is the PI of these projects.

jcfr (approved):

PR #7117, opened 2023-07-22
ENH: Make it easier to load of .vtk image files

lassoan:
.vtk files tried to be loaded as Model node by default. Loading failed when attempted to load an image file without manually setting Description to Volume in "Add data" window.

This commit makes .vtk image file loaded as image by default, avoiding user questions such as this one: https://discourse.slicer.org/t/error-exporting-and-importing-vtk/30733/2

### DIFF
--- a/Modules/Loadable/Models/qSlicerModelsReader.cxx
+++ b/Modules/Loadable/Models/qSlicerModelsReader.cxx
@@ -101,6 +101,36 @@ qSlicerIOOptions* qSlicerModelsReader::options()const
   return options;
 }
 
+//----------------------------------------------------------------------------
+double qSlicerModelsReader::canLoadFileConfidence(const QString& fileName)const
+{
+  double confidence = Superclass::canLoadFileConfidence(fileName);
+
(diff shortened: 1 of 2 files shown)


pieper (approved):
I wish there was a cleaner way, but this looks as good as anything.  Thanks for adding this!

lassoan:
The confidence estimation mechanism does not need to provide 100% correct answer, but the check has to be inexpensive (in term of I/O and CPU usage). The current implementation of looking at the beginning of the file fulfills these requirements. 

Parsing the header would be "cleaner" and it would give more accurate answer (something like 99.99% accuracy instead of 99%), but it could take magnitudes more time.

Or so your meant something else should be cleaner?

pieper:
Yes, I agree this is a good solution.  What I meant by cleaner is that it's a shame that the vtk format is overloaded with no easy way to introspect the metadata.  We could make a convention to save as .vol.vtk and .pd.vtk or something so we could be more confident when loading (like seg.nrrd or .nii.gz to minimize uncertainty) but the solution you provided is more than reasonable.

PR #7127, opened 2023-07-26
COMP: Update CTK removing code specific to Qt 4

jcfr:
List of CTK changes:

```
$ git shortlog 2b16f9dff..ee5508d6e --no-merges
James Butler (2):
      COMP: Update C++ source files removing code specific to Qt 4
      ENH: Remove custom CTK Dialog options

      COMP: Revert removal of custom CTK Dialog options
```

### DIFF
--- a/SuperBuild/External_CTK.cmake
+++ b/SuperBuild/External_CTK.cmake
@@ -73,7 +73,7 @@ if(NOT DEFINED CTK_DIR AND NOT Slicer_USE_SYSTEM_${proj})
 
   ExternalProject_SetIfNotDefined(
     Slicer_${proj}_GIT_TAG
-    "2b16f9dff6ce6a1ddf5589eea4878e3cad0fc951"
+    "ee5508d6eeec882eb39358346c1823b6c0bbe825"
     QUIET
     )
 


jcfr:
It turns out that attempting to build against CTK after integrating the following pull request is failing:
* https://github.com/commontk/CTK/pull/1127

Associated error:

```
/home/jcfr/Projects/Slicer-Release/Slicer-build/Modules/Scripted/DICOMLib/Widgets/ui_qSlicerDICOMExportDialog.h:185:64: error: ‘ShowDirsOnly’ is not a member of ‘ctkPathLineEdit’
  185 |         PathLineEdit_OutputFolder->setOptions(ctkPathLineEdit::ShowDirsOnly);
      |                                                                ^~~~~~~~~~~~
```

jcfr:
The issue is that `ctkDirectoryButton` and `ctkPathLineEdit` both derive from `QWidget` and client code uses reference to `ctkPathLineEdit::ShowDirsOnly`

To ensure existing code can still compile, we should revert the removal and instead associated a deprecated warning with the use of the obsolete API.

cc: @jamesobutler

jcfr:
Related pull request:
* https://github.com/commontk/CTK/pull/1128

jamesobutler:
Yes to update the Slicer you’ll need my changes that I have at https://github.com/jamesobutler/Slicer/tree/jamesobutler-ctk-drop-qt4 which I was using when doing my own testing.

I figured it was such a minor change, but I can work to add it as a deprecation on the CTK side.

Modules/Scripted/DICOMLib/Widgets/Resources/UI/qSlicerDICOMExportDialog.ui
```diff
@@ -255,7 +255,7 @@ If unchecked, the exported dataset will be added to the DICOM database.</string>
          <set>ctkPathLineEdit::Dirs|ctkPathLineEdit::Drives|ctkPathLineEdit::Executable|ctkPathLineEdit::NoDot|ctkPathLineEdit::NoDotDot|ctkPathLineEdit::PermissionMask|ctkPathLineEdit::Readable|ctkPathLineEdit::Writable</set>
         </property>
         <property name="options">
-         <set>ctkPathLineEdit::ShowDirsOnly</set>
+         <set>QFileDialog::ShowDirsOnly</set>
```
jamesobutler, line 258: @jcfr If you don't want this change, you can revert it

jcfr: The proposed change makes sense, I suggest we apply it once we have addressed:
* https://github.com/commontk/CTK/issues/1129

jcfr: > I figured it was such a minor compatibility change, but I can work to add it as a deprecation on the CTK side.

I will do a search to assess if the this is used on the C++ side of Slicer extensions.

The issues is that the `ctkDirectoryButton::Option` and `ctkDirectoryButton::Option` enum values are not available anymore from the class as they derive from `QWidget` and not from `QFileDialog`.

PR #7138, opened 2023-07-31
Fix external project download extract timestamp warnings

jcfr:
This fixes `DOWNLOAD_EXTRACT_TIMESTAMP` warnings related to `CMP0135`[^1] introduced in CMake >= 3.24.

[^1]: https://cmake.org/cmake/help/latest/policy/CMP0135.html

### Summary

* Updates `ExternalProjectDependency` CMake module based on the latest version provided by https://github.com/commontk/artichoke
* Updates `CTK` so that its bundled version of `ExternalProjectDependency` is also the most recent
* Updates `Swig` and `python` external projects to explicitly set `DOWNLOAD_EXTRACT_TIMESTAMP` to `1` if using  CMake >= 3.24. This is needed because these projects are directly using `ExternalProject_Add` (without relying on `ExternalProject_Include_Dependencies`)

### DOWNLOAD_EXTRACT_TIMESTAMP warning

This will avoid warnings like the following:

```
CMake Warning (dev) at /path/to/cmake-3.26.4-linux-x86_64/share/cmake-3.26/Modules/ExternalProject.cmake:3091 (message):
  The DOWNLOAD_EXTRACT_TIMESTAMP option was not given and policy CMP0135 is
  not set.  The policy's OLD behavior will be used.  When using a URL
  download, the timestamps of extracted files should preferably be that of
  the time of extraction, otherwise code that depends on the extracted
  contents might not be rebuilt if the URL changes.  The OLD behavior
  preserves the timestamps from the archive instead, but this is usually not
  what you want.  Update your project to the NEW behavior or specify the
  DOWNLOAD_EXTRACT_TIMESTAMP option with a value of true to avoid this
  robustness issue.
```

### DIFF
--- a/CMake/ExternalProjectDependency.cmake
+++ b/CMake/ExternalProjectDependency.cmake
@@ -92,14 +92,17 @@ endif()
 #.rst:
 # .. cmake:variable:: EP_GIT_PROTOCOL
 #
-# The value of this variable is controlled by the option ``<SUPERBUILD_TOPLEVEL_PROJECT>_USE_GIT_PROTOCOL``
-# automatically defined by including this CMake module. Setting this option allows to update the value of
-# ``EP_GIT_PROTOCOL`` variable.
+# The value of this variable is always set to ``https``.
 #
-# If enabled, the variable ``EP_GIT_PROTOCOL`` is set to ``git``. Otherwise, it is set to ``https``.
-# Since GitHub removed unauthenticated access through the git protocol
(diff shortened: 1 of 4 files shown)


jcfr:
* Local reconfigure using CMake 2.22.2 :heavy_check_mark: 
* Local configuration using CMake 3.26.4 confirm that the warnings are addressed :heavy_check_mark:

PR #7141, opened 2023-08-01
ENH: Update MultiVolumeImporter to load as volume sequence by default

lassoan:
Time sequence of volumes can be loaded as multivolume (vtkMRMLMultiVolumeNode) or volume sequence (vtkMRMLSequenceNode). Volume sequence is a more generic representation and many more modules use it than the old multivolume representation.

This commit changes the default so that when a user loads a 4D volume from DICOM then it is loaded into a volume sequence by default.

Users can revert to the old behavior by selecting in Application settings / DICOM / Preferred multi-volume import format -> multi-volume. For loading a single image using non-default settings, enable "Advanced" option in DICOM module, click Examine, and choose the appropriate loadable (loadables are offered for both multivolme and volume sequence representation, application settings specify which is selected by default).

### DIFF
--- a/SuperBuild.cmake
+++ b/SuperBuild.cmake
@@ -237,7 +237,7 @@ list_conditional_append(Slicer_BUILD_MultiVolumeExplorer Slicer_REMOTE_DEPENDENC
 
 Slicer_Remote_Add(MultiVolumeImporter
   GIT_REPOSITORY ${EP_GIT_PROTOCOL}://github.com/fedorov/MultiVolumeImporter.git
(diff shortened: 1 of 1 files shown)


jcfr:
For reference, list of `MultiVolumeImporter` change:

```
$ git shortlog c14758931..198e6b320 --no-merges 
Andras Lasso (1):
      Read 4D volumes as volume sequence by default
```

_See https://github.com/fedorov/MultiVolumeImporter/compare/c14758931...198e6b320_

jcfr (approved):

PR #7142, opened 2023-08-01
ENH: Update slicer.util.updateVolumeFromArray to support 1D NumPy array

jcfr:
This was created based of changes originally contributed through:
* https://github.com/Slicer/Slicer/pull/6733


jcfr:
I confirm that running the tests associated with the `NumPyDataLoader` associated with the following pull request work as expected when used with this patch:
* https://github.com/PerkLab/SlicerSandbox/pull/19

Without this patch, the corresponding test fails with `RuntimeError: Unsupported numpy array shape: (256,)`

Issue #6912, opened 2023-03-24
Build of the extensions index is blocking due to extension waiting for user input


jcfr:
## Summary

After connecting to the windows factory using VNC, I was able to observe that the ExtraMarkups extension was showing a popup dialog and expecting the user to click.

<img width=75% src="https://user-images.githubusercontent.com/219043/227523221-7a3f6995-a7f8-4e0c-a31f-1f3632cbd08d.png"/>

## Steps to reproduce

NA

## Expected behavior

No popup should be displayed.

## Environment

* Observed in the context of the Preview build on Windows (overload factory) associated with fbc2ede

jcfr:
The function `qSlicerStenosisMeasurement3DModuleWidget::installExtensionFromServer` should be improved to skip the popup when running test mode.

https://github.com/vmtk/SlicerExtension-VMTK/blob/5c31f18ba995fa97494b302e23f8a8d16cdc4f8d/StenosisMeasurement3D/qSlicerStenosisMeasurement3DModuleWidget.cxx#L416-L466

Ideally, the function should be contributed to Slicer itself.

jcfr:
cc: @chir-set @lassoan

jcfr:
Once the function `installExtensionFromServer` is added to Slice core, the following should be updated:

| Project | URL |
|--|--|
| SlicerMorph | https://github.com/SlicerMorph/SlicerMorph/blob/c86b3dd507ab0df9ad0fef721b1da1d69cc37d43/SegmentEndocranium/SegmentEndocranium.py#L51-L61 |
| slicer-notebook | https://github.com/Slicer/SlicerDocker/blob/4fc3be96fd28cf97677650a76ac83f209efe834b/slicer-notebook/install.sh |
| Slicer documentation | https://github.com/Punzo/Slicer/blob/08fe346009554f82e63e7d7aaaaec06c16f0a233/Docs/developer_guide/script_repository/gui.md?plain=1#L1137-L1147 |
| DMRIInstall | https://github.com/Punzo/Slicer/blob/08fe346009554f82e63e7d7aaaaec06c16f0a233/Modules/Scripted/DMRIInstall/DMRIInstall.py#L111-L125 |
| kaapana | https://github.com/kaapana/kaapana/blob/0a3b2e14f64510671c9c9dcb7ef73853002d8940/services/applications/slicer-workbench/docker/files/install-slicer-extension.py#L8 |

jcfr:
Related discourse post(s):
[...]

PR #7148, opened 2023-08-03
Update install extension from server return bool

jcfr:


Docs/developer_guide/script_repository/gui.md
```diff
@@ -1223,6 +1223,25 @@ On some systems, *shell=True* must be specified as well.
...
+extensionName = 'SlicerIGT'
+em = slicer.app.extensionsManagerModel()
+em.interactive = False  # prevent display of popups
+restart = True
+if not em.installExtensionFromServer(extensionName, restart):
+  raise ValueError(f"Failed to install {extensionName} extension")
+```
+
+:::{deprecated} 5.4
+The previous method used is now deprecated and no longer recommended.
```
lassoan, line 1242: What previous method? Can you describe it (specify method name, etc )?

Base/QTCore/qSlicerExtensionsManagerModel.cxx
```diff
@@ -1714,13 +1714,14 @@ bool qSlicerExtensionsManagerModel::downloadAndInstallExtensionByName(const QStr
...
 //-----------------------------------------------------------------------------
-void qSlicerExtensionsManagerModel::installExtensionFromServer(const QString& extensionName, bool restart)
+bool qSlicerExtensionsManagerModel::installExtensionFromServer(const QString& extensionName, bool restart)
 {
   Q_D(qSlicerExtensionsManagerModel);
 
   if (this->isExtensionInstalled(extensionName))
     {
-    return;
+    // Already installed; nothing to do
```
lassoan, line 1723: Maybe we check if an update is available for an extension and if a flag is specified that the latest version of the extension should be installed then we should install that update.

lassoan (commented):
Thank you, it looks good. Just added two comments inline.

Docs/developer_guide/script_repository/gui.md
jcfr:
```
The approach presented below is now deprecated and no longer recommended.
```

jcfr: Instead of this, I was thinking to create a python package called `slicer-install` (or `slicer-env` or ... ) that would provide a way to:
* install Slicer
* install extensions
* ... 

This would allows to have approach linked of https://github.com/Slicer/Slicer/issues/6912#issuecomment-1482827047 all simplified. 

Also copied below for convenience:


> [...] following should be updated:
> 
> | Project | URL |
> |--|--|
> | SlicerExtension-VMTK | https://github.com/search?q=repo%3Avmtk%2FSlicerExtension-VMTK%20installExtensionFromServer&type=code |
> | SlicerMorph | https://github.com/SlicerMorph/SlicerMorph/blob/c86b3dd507ab0df9ad0fef721b1da1d69cc37d43/SegmentEndocranium/SegmentEndocranium.py#L51-L61 |
> | slicer-notebook | https://github.com/Slicer/SlicerDocker/blob/4fc3be96fd28cf97677650a76ac83f209efe834b/slicer-notebook/install.sh |
> | Slicer documentation | https://github.com/Punzo/Slicer/blob/08fe346009554f82e63e7d7aaaaec06c16f0a233/Docs/developer_guide/script_repository/gui.md?plain=1#L1137-L1147 |
> | DMRIInstall | https://github.com/Punzo/Slicer/blob/08fe346009554f82e63e7d7aaaaec06c16f0a233/Modules/Scripted/DMRIInstall/DMRIInstall.py#L111-L125 |
> | kaapana | https://github.com/kaapana/kaapana/blob/0a3b2e14f64510671c9c9dcb7ef73853002d8940/services/applications/slicer-workbench/docker/files/install-slicer-extension.py#L8 |

lassoan: I think we should just remove the old installation method. It would just confuse people.

Having a slicer-install pyrhon package would be great. Ideally, it should be usable for installing Slicer in Google colab.

Since slicerio package already can launch Slicer and call its methods (using web API), we may consider adding Slicer installation to this package. It could then download, install, and run Slicer without requiring the user to install or set up anything. Or slicerio could just use slicer-install for installing Slicer.

Base/QTCore/qSlicerExtensionsManagerModel.cxx
jcfr: Now tracker in https://github.com/Slicer/Slicer/issues/7149

Docs/developer_guide/script_repository/gui.md
jcfr: This is now tracked in:
* https://github.com/Slicer/Slicer/issues/7150

PR #7151, opened 2023-08-04
ENH: Update ExtensionsManager install API to check for update

jcfr:
Add support for the `update` parameter in the `qSlicerExtensionsManagerModel::installExtensionFromServer` function.

Users can now set the `update` parameter to `true` to schedule the update of an already installed extension to the latest version available on the server.

The update will take effect during the next application restart.

If the extension is not already installed, this parameter has no effect.


Base/QTCore/qSlicerExtensionsManagerModel.h
```diff
@@ -398,15 +398,22 @@ public slots:
...
+  /// \param update Set to true to schedule the extension for an update on the next application restart (default: false).
   ///
   /// \return \c true if the extension is successfully installed or already installed.
   ///
   /// \sa setInteractive
   /// \sa isExtensionInstalled, installExtension, updateExtensionsMetadataFromServer, downloadAndInstallExtensionByName
+  /// \sa scheduleExtensionForUpdate
   /// \sa qSlicerCoreApplication::testAttribute, qSlicerCoreApplication::AA_EnableTesting, qSlicerCoreApplication::restart
-  bool installExtensionFromServer(const QString& extensionName, bool restart = true);
+  bool installExtensionFromServer(const QString& extensionName, bool restart = true, bool update = false);
```
jcfr, line 416: Since we just introduced the API, it is not too late to have `update` before `restart`.

```diff
- bool restart = true, bool update = false
+ bool update = false, bool restart = true
```

@lassoan  Any preference ?

jcfr:
### Testing

_To deal with the fact my local build didn't have any matching extension on the server, I forced the revision to `31916`._

**Install the Sandbox extension:**

```
rm -rf slicer.org/Extensions-* && \
./Slicer \
  --no-main-window \
  --exit-after-startup \
  --python-code \
    "em=slicer.app.extensionsManagerModel(); \
     em.interactive = False; \
     em.slicerRevision = '31916'; \
     restart = False; \
     em.installExtensionFromServer('Sandbox', restart);"
```

```
[...]
Installing the extension(s) without asking for confirmation (interactive mode is disabled)
Retrieving extension metadata for Sandbox extension
Retrieving Sandbox extension files (extensionId: 64cb3a4a16826926710e3de1)
Downloading Sandbox extension (item_id: 64cb3a4a16826926710e3de1, file_id: 64cb3a4f16826926710e3de8)
Installed extension Sandbox (64cb3a4a16826926710e3de1) revision e163d61
```

**Edit `slicer.org/Extensions-31921/Sandbox.s4ext`**

To ensure the check for update is triggered, edit the file `Sandbox.s4ext` changing the value associated with the  `revision` key.

```diff
- revision e163d61
+ revision e163d62
```

**Trigger the update:**

```
./Slicer \
  --no-main-window \
  --exit-after-startup \
  --python-code \
    "em=slicer.app.extensionsManagerModel(); \
     em.interactive = False; \
     em.slicerRevision = '31916'; \
     restart = False; \
     update = True; \
     em.installExtensionFromServer('Sandbox', restart, update);"
```

```
[...]
Update found for Sandbox extension: 'e163d62' installed, 'e163d61' available, 
Retrieving extension metadata for Sandbox extension
Retrieving Sandbox extension files (extensionId: 64cb3a4a16826926710e3de1)
Downloading Sandbox extension (item_id: 64cb3a4a16826926710e3de1, file_id: 64cb3a4f16826926710e3de8)
Sandbox extension scheduled for update
Installed extension Sandbox (64cb3a4a16826926710e3de1) revision e163d61
Successfully updated extension "Sandbox"
```

jcfr:
If the extension is already installed and need to be updated, specifying `em.interactive = True` does not lead to any pop-up dialog.

**Question** Should we explicitly handle this case by asking confirmation ?

jcfr:
@lassoan  Given the added complexity, I am leaning toward having a new API called `updateExtensionFromServer(const QString& extensionName, bool restart = true)`

lassoan:
If update option was not part of the install method then it might be somewhat simpler for Slicer but slower and much more complex for users. Users would always need to call 2 methods (install and update) when they install an extension. Calling just install without update would not make sense, because it would leave the application in an unknown state (an extension is installed, but it may lack features or fixes if the user happened to install that one extension earlier). It is not by chance that `pip install` has an update flag (instead of having a separate `pip update` command).

jcfr:
Let's keep the update option then

Should have a updateAllExtensionsFromServer ? 

If not, it would not be possible to update all of them easily.

PR #7155, opened 2023-08-06
ENH: Add Slicer website primary links to Welcome module

jamesobutler:
This aims to update the Welcome module to have the similar navigation links on the Slicer landing website to also be on the Slicer welcome module.

image

Below is a quick demo of the re-work. Other changes where to be concise while showing all information that is desired to be conveyed without clicking various expandable widgets.

https://github.com/Slicer/Slicer/assets/15837524/34d5372a-2a85-4b64-a06a-d2b48fc9d3e7

I did not replicate the "Twitter" button because to be accurate I didn't want to create a new button with the text of "X" since Twitter has been renamed to "X". Something about this on the Slicer website will have to be updated as well. Another alternative could be "Social". I instead replaced it with an "Acknowledgement" button that links to the Acknowledgement webpage. However, the existing "Acknowledgement" text browser may be consider redundant. I wanted a 6th button to fill out the 6th color from the Slicer logo.

I currently find it a little weird that the links at the top of the Welcome module which are more so a "Getting Started" informative section are above the new links. We could consider flip/flopping that organization, or removing them now that the large "Documentation" linked button would bring them to the ReadTheDocs page where https://slicer.readthedocs.io/en/latest/user_guide/getting_started.html#using-slicer could be a better informative area.


pieper:
I like the idea here a lot.  Small point, the Download button seems odd since the user will have already downloaded to run the app.  Maybe this could be Check for Updates?

In terms of using the extra button, maybe a link to either the feature request section of discourse or the gitub issues page.

Or maybe more helpful, could we link to a chatbot that's good at answering basic user questions?  Like bing or bard or something?

jamesobutler:
> I like the idea here a lot. Small point, the Download button seems odd since the user will have already downloaded to run the app. Maybe this could be Check for Updates?

Yes I was thinking a similar thing. Currently there is a separate button (see below) that appears when there is a new application update available and that button links to https://download.slicer.org/ just like this new "Download" button is doing currently. I was thinking maybe the new Download button with link could still exist if people are wanting to update their Slicer preview build, but for Stable releases the new Download button could indicate an update available rather than having the separate button.

image


> In terms of using the extra button, maybe a link to either the feature request section of discourse or the gitub issues page.
> Or maybe more helpful, could we link to a chatbot that's good at answering basic user questions? Like bing or bard or something?

I chose "Acknowledgements" as the extra button because it was one of the other links in the "About" menu on the website that seemed best as a different category for the last button to cover. "Citing" could also be an important option instead of "Acknowledgements".

The website contains major navigation that is:
- Application
  - Getting Started 
  - Download
- Support
  - Training
  - Documentation
  - Forum
  - Get Help
- Solutions
  - SlicerIGT ..etc
- Developers
  - GitHub
  - Developer Guide
  - Quality Dashboard
  - Download Statistics
  - Extensions
  - Contribute
- About
  - Contact
  - License
  - Citing
  - Commercial Use
  - 3D Slicer Enabled Research
  - Acknowledgements
  - Twitter
  - YouTube

pieper:
That first colorful button seems like a very prominent place so using it for a feature that may be a bit confusing or redundant.

I mentioned the chatbot option since we have talked a lot about it on discourse but we don't have it on the website or in the app.  It's good that people can ask real people on the forum, but I also sometimes see simple questions going unanswered possibly because many people think that someone else can answer just as easily or that it would take too long to give a complete answer.  For some of these questions it seems the bots do pretty well.  Some people may also feel more comfortable asking a bot rather than posting in public.  That said, I don't know if we trust the bots well enough to feature them.

jamesobutler:
Yes the chat bot is certainly a good idea to keep track of. Would you be willing to write up a GitHub issue for it? I’m not sure how it would work in terms of providing accurate info for the specific version of Slicer being used (5.2.x vs 5.4.x) or even if some behavior changes in a preview build.

pieper:
Good point - I'll make a different issue around chatbots.

jcfr:
For reference, here is the Qt Creator one:

jcfr:
Idea:
* Welcome layout (with getting started, and step by step guide, along with view showing info)
* Badge after learning through Slicer, going through the training

PR #7156, opened 2023-08-06
ENH: Add DICOM direct load reader

jamesobutler:
This is used for directly loading DICOM files into the scene without adding to the DICOM database. The DICOM module however is still used in the logic for loading. Currently there is the "DICOM import" reader option from the Add Data dialog which will import it into the DICOM database, but will not load it into the scene.

Based on https://slicer.readthedocs.io/en/latest/developer_guide/script_repository.html#load-dicom-files-into-the-scene-from-a-folder

### DIFF
--- a/Modules/Scripted/CMakeLists.txt
+++ b/Modules/Scripted/CMakeLists.txt
@@ -25,6 +25,7 @@ endif()
 if(Slicer_BUILD_DICOM_SUPPORT)
   list(APPEND modules
     DICOM
+    DICOMDirectReader
     DICOMLib
     DICOMPlugins
     DICOMPatcher
--- a/Modules/Scripted/DICOMDirectReader/DICOMDirectReader.py
+++ b/Modules/Scripted/DICOMDirectReader/DICOMDirectReader.py
@@ -30,7 +30,7 @@ class DICOMDirectReaderFileReader:
         self.parent = parent
 
     def description(self):
-        return 'DICOM direct load'
+        return _('DICOM direct load')
 
     def fileType(self):
         return 'DICOMDirectLoad'
--- a/Modules/Scripted/DICOMDirectReader/DICOMDirectReader.py
+++ b/Modules/Scripted/DICOMDirectReader/DICOMDirectReader.py
@@ -1,0 +1,68 @@
+import os
+import slicer
+from slicer.i18n import tr as _
+from slicer.i18n import translate
+from slicer.ScriptedLoadableModule import ScriptedLoadableModule
+
+
+class DICOMDirectReader(ScriptedLoadableModule):
+    """A module for loading Revvity Living Image files."""
+
+    def __init__(self, parent):
+        """Initialize this class."""
+        ScriptedLoadableModule.__init__(self, parent)
+        self.parent.title = _("DICOMDirectReader")
+        self.parent.categories = ["", translate("qSlicerAbstractCoreModule", "Informatics")]  # top level module
+        parent.dependencies = ["DICOM"]
+        parent.contributors = [""]
+        parent.helpText = _("This module allows direct loading of DICOM files without importing into the DICOM database.")
+        parent.acknowledgementText = _("This work is supported by NA-MIC, NAC, BIRN, NCIGT, and the Slicer Community.")
+        # don't show this module - it is only for registering a reader
+        parent.hidden = True
+        self.parent = parent
+
+
+class DICOMDirectReaderFileReader:
+    """This reader claims DICOM files with higher-than default confidence and loads them directly into the scene.
+    """
+
+    def __init__(self, parent):
+        self.parent = parent
+
+    def description(self):
+        return 'DICOM direct load'
+
+    def fileType(self):
+        return 'DICOMDirectLoad'
+
+    def extensions(self):
+        return ['DICOM (*.dcm)', 'DICOM (*)']
+
+    def canLoadFileConfidence(self, filePath):
+        import pydicom
+        if pydicom.misc.is_dicom(filePath):
+            # This is a DICOM file, so we return higher confidence than the default 0.5
+            # to import DICOM files directly into the scene.
+            confidence = 0.6
+            if slicer.util.selectedModule() == "DICOM":
+                return confidence - 0.05  # rank "DICOM direct load" below "DICOM import"
+            return confidence + 0.05  # rank "DICOM direct load" above "DICOM import"
+        else:
+            return 0.0
+
+    def load(self, properties):
+        filePath = properties['fileName']
+        dicomFilesDirectory = os.path.dirname(os.path.abspath(filePath))
+
+        loadedNodeIDs = []
+
+        from DICOMLib import DICOMUtils
+        with DICOMUtils.TemporaryDICOMDatabase() as db:
+            DICOMUtils.importDicom(dicomFilesDirectory, db)
+            patientUIDs = db.patients()
+            for patientUID in patientUIDs:
+                loadedNodeIDs.extend(DICOMUtils.loadPatientByUID(patientUID))
+
+        self.parent.loadedNodes = loadedNodeIDs
+
+        return bool(self.parent.loadedNodes)
--- a/Applications/SlicerApp/Testing/Python/ScriptedModuleDiscoveryTest/ModuleE_WithFileWriter_WithoutWidget.py
+++ b/Applications/SlicerApp/Testing/Python/ScriptedModuleDiscoveryTest/ModuleE_WithFileWriter_WithoutWidget.py
@@ -1,3 +1,4 @@
+from slicer.i18n import tr as _
 from slicer.ScriptedLoadableModule import *
 
 SOMEVAR = 'E'
@@ -26,7 +27,7 @@ class ModuleE_WithFileWriter_WithoutWidgetFileWriter:
         self.parent = parent
 
     def description(self):
-        return 'My writer file type'
+        return _('My writer file type')
 
     def fileType(self):
         return 'MyWriterFileType'
--- a/Applications/SlicerApp/Testing/Python/ScriptedModuleDiscoveryTest/ModuleF_WithFileReader_WithoutWidget.py
+++ b/Applications/SlicerApp/Testing/Python/ScriptedModuleDiscoveryTest/ModuleF_WithFileReader_WithoutWidget.py
@@ -1,3 +1,4 @@
+from slicer.i18n import tr as _
 from slicer.ScriptedLoadableModule import *
 
 SOMEVAR = 'F'
@@ -26,7 +27,7 @@ class ModuleF_WithFileReader_WithoutWidgetFileReader:
         self.parent = parent
 
     def description(self):
-        return 'My reader file type'
+        return _('My reader file type')
 
     def fileType(self):
         return 'MyReaderFileType'
--- a/Applications/SlicerApp/Testing/Python/SlicerScriptedFileReaderWriterTest.py
+++ b/Applications/SlicerApp/Testing/Python/SlicerScriptedFileReaderWriterTest.py
@@ -2,6 +2,7 @@ import logging
 import os
 
 import slicer
+from slicer.i18n import tr as _
 from slicer.ScriptedLoadableModule import *
 
 
@@ -34,7 +35,7 @@ class SlicerScriptedFileReaderWriterTestFileReader:
         self.parent = parent
 
     def description(self):
-        return 'My file type'
+        return _('My file type')
 
     def fileType(self):
         return 'MyFileType'
@@ -101,7 +102,7 @@ class SlicerScriptedFileReaderWriterTestFileWriter:
         self.parent = parent
 
     def description(self):
-        return 'My file type'
+        return _('My file type')
 
     def fileType(self):
         return 'MyFileType'
--- a/Modules/Scripted/DICOM/DICOM.py
+++ b/Modules/Scripted/DICOM/DICOM.py
@@ -898,7 +898,7 @@ class DICOMFileReader:
         self.parent = parent
 
     def description(self):
-        return 'DICOM import'
+        return _('DICOM import')
 
     def fileType(self):
         return 'DICOMFileImport'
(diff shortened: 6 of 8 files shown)


Modules/Scripted/CMakeLists.txt
```diff
@@ -25,6 +25,7 @@ endif()
 if(Slicer_BUILD_DICOM_SUPPORT)
   list(APPEND modules
     DICOM
+    DICOM2
```
pieper, line 28: I like the idea of this, but it would be nice to have a more descriptive name, like DICOMDirectReader.

pieper (changes requested):

Modules/Scripted/CMakeLists.txt
jamesobutler: Thanks for the suggestion! I have updated it to DICOMDirectReader.

Modules/Scripted/DICOMDirectReader/DICOMDirectReader.py
```diff
@@ -0,0 +1,68 @@
...
+
+class DICOMDirectReaderFileReader:
+    """This reader claims DICOM files with higher-than default confidence and loads them directly into the scene.
+    """
+
+    def __init__(self, parent):
+        self.parent = parent
+
+    def description(self):
+        return 'DICOM direct load'
```
pieper, line 33: Should this be marked translatable?

jamesobutler: Yes it appears a reader's `description` should be marked translatable per
https://github.com/Slicer/Slicer/blob/7ef59614c2458e4f693ab77834e6726cf6ae0540/Modules/Loadable/Volumes/qSlicerVolumesReader.cxx#L90-L93

I had copied based on the "DICOM import" reader which didn't translate it. I've pushed an additional commit to mark various scripted file reader's description as being translatable. @lassoan can probably confirm this action with the translation work he has been working on.
https://github.com/Slicer/Slicer/blob/7ef59614c2458e4f693ab77834e6726cf6ae0540/Modules/Scripted/DICOM/DICOM.py#L900-L901

lassoan:
DICOM reading is probably more complicated than most users would expect, so simplification is very useful. I can review this later this week.

Modules/Scripted/DICOMDirectReader/DICOMDirectReader.py
```diff
@@ -0,0 +1,68 @@
...
+        dicomFilesDirectory = os.path.dirname(os.path.abspath(filePath))
+
+        loadedNodeIDs = []
+
+        from DICOMLib import DICOMUtils
+        with DICOMUtils.TemporaryDICOMDatabase() as db:
+            DICOMUtils.importDicom(dicomFilesDirectory, db)
+            patientUIDs = db.patients()
+            for patientUID in patientUIDs:
+                loadedNodeIDs.extend(DICOMUtils.loadPatientByUID(patientUID))
```
jcfr, line 64: Once the volume is loaded the MRML attribute `DICOM.instanceUIDs` is set. This allows to link back to the series and get associated tags.

Ability to load data without having them permanently in the database is indeed helpful, especially context of a custom application intended to review a case and then discard it.

That said, by using the `TemporaryDICOMDatabase`,it prevents future lookup of tags.

I am wondering if the API could be improved to accept a list of tags that would be associated to a dictionary or set a MRML attribute to the volume node. 

Similar to the `TagsToPrecache` concept available in CTK

cc: @tbirdso

pieper: Good point @jcfr.  I think if the dicom data is not in the database we should remove the DICOM.instanceUIDs attributes since the having them present is not consistent with the way they are used.

jcfr: Keeping `DICOM.instanceUIDs` is useful as it indicates the provenance.

To accommodate this, we could support a loading option called `cacheTagsAsMRMLAttributes`.

@jamesobutler  Did you come up with a convention for the attribute keys ?  (e.g `DICOM.instanceUID.(<UID>).tags.(<tag>)` where  `<UID>` is a specific `instanceUID` and `tag` is formatted as `NNNN,NNNN` .

Here is an example of attribute key and associated value:

| Key | Value |
|---|---|
| `DICOM.instanceUID.(2.25.272791004175754683422124020696884278231).tags.(0008,0060)` | `CT`|
| `DICOM.instanceUID.(2.25.272791004175754683422124020696884278231).tags.(0028,0030)` | `0.390625, 0.390625`|
```

Alternatively, a `vtkMRMLTableNode` with few columns could be created and added as a reference to the volume node. 

The columns would be: `InstanceUID`, `Tag`, `Attribute`, `Value`  (we would also add `VR`)

lassoan: DICOM is not a file format but distributed database. The DICOM files are database dumps that practically cannot be read and stored in some other file format or tag:value pairs (e.g., it is not feasible to store all DICOM metadata in volume node attributes). Therefore, if we implement a simplified DICOM image loader then I see two options:
- Option A: Make the simplified reader use the application's DICOM database, just avoid the extra clicks and import and load everything with default options.
- Option B: Make the simplified reader use a temporary DICOM database only for the time of importing and loading and then cut off all connections to DICOM. The loaded nodes would not be related to DICOM in any way - no DICOM UIDs, no provenance, no metadata (other than maybe a few very basic tags). We can give temporary access to modules to DICOM UIDs and filenames at the end of data loading before the cutoff (see option CC below for details).

See details about how I arrived to this conclusion and the options I considered below.

---

It is not practically feasible to store all potentially necessary tags as node attributes. We could only store a limited set of basic data elements in node attributes, because in general:
1. Tags can be very complex to encode in strings: tags in a DICOM information object are organized in a tree (using sequence elements), data elements with the same tag may be repeated many times, `(group,element)` alone cannot describe private tags (but you also need to use private creator strings), etc.
2. Values can be complex and inefficient to store in strings: there are many value representation types (including sequences), multiple values may need to be stored in one element, in case of private tags the value representation may be unknown, there could be values of enormous size, which would be slow to encode/decode and would take up a lot of storage space, etc.

Therefore, I would not try to solve metadata access by storing more DICOM data elements in volume node attributes.

Instead, we could offer some alternative for looking up filename based on UID:
- Option AA: Filenames could be listed in volume node attributes. THe problem is that data nodes should not reference filenames directly. It is not just because of breaking design concepts, but practical things like they would become obsolete when the scene is saved and loaded on a different computer.
- Option BB: We could store input filenames in storage nodes. However, we would then need to store the input DICOM files in the scene, too (otherwise saving the scene and loading it on a different computer would end up with many unresolved filename references). This would result in redundant data storage (we would store both the original input DICOM files and the imported data), which could be prohibitive, and could also mean we never really know when the DICOM loading is finished.
- Option CC: We could prvode a temporary access to the input DICOM files while the DICOM loading is finalized: give a chance to modules to inspect nodes loaded from DICOM (and do perform additional DICOM UID based lookups as needed) right after loading. This would be easy to implement, because we could just need to emit a signal when DICOM loading is completed (any modules that want to further inspect loaded data would still have access to the DICOM UIDs and file references in the temporary DICOM database), and then we would delete the DICOM UIDs from the volume node attributes and switch back from the temporary DICOM database to the main DICOM database. Unfortunately, this would not fulfill all the requirements, because we cannot (or do not want to - due to performance reasons) to retrieve and store all DICOM metadata that may be ever needed for any purpose at the time of DICOM loading (for example, we may want to retrieve some additional DICOM metadata for DICOM export). It could be also an issue that DICOM data sets that were loaded before installing some extensions may not work with those extensions (because some required tags were not retrieved when the data set was loaded).

Overall, all options are bad. They all have serious limitations, which would prevent using full set of features that users would expect for DICOM data. Having nodes in the scene that may behave like DICOM data in some circumstances but some DICOM features would not be supported would be very confusing for users and lots of extra work for Slicer core and extension developers. I think we should avoid this.

jamesobutler: > Option B: Make the simplified reader use a temporary DICOM database only for the time of importing and loading and then cut off all connections to DICOM. The loaded nodes would not be related to DICOM in any way

This is how I have currently developed it and have used it in my own work. I’m not aware of other modules that specifically need DICOM metadata elements. If there are some I would expect users would utilize the DICOM database. This direct load method is purely to allow users to load the data nicely using the DICOM module’s file readers and allow the user to quickly view the image data when they don’t care about viewing/keeping the metadata in the database.

In my Slicer custom app I have only read (0080,0060) Modality and stored that key for some our logic about what display options to give for it. This reader could potentially accept advanced options about what tags to store as node attributes. 

> Option BB: We could store input filenames in storage nodes. 

My Slicer custom app also creates a default storage node and sets the fileName to the filepath used, but this is mainly because my custom app doesn’t allow duplicate loading of the same image data and I used storage node paths to confirm that. Also it doesn’t supporting saving out the volume node. It is purely for viewing the image data and doing things like segmentation.

pieper: > > Option B: Make the simplified reader use a temporary DICOM database only for the time of importing and loading and then cut off all connections to DICOM. The loaded nodes would not be related to DICOM in any way
> 
> This is how I have currently developed it and have used it in my own work. I’m not aware of other modules that specifically need DICOM metadata elements. If there are some I would expect users would utilize the DICOM database.

This is my opinion too.  Having the DICOM.instanceUIDs on a loaded volume that isn't in the database would break the exiting convention and lead to confusion.  Having the dicom tags as node attributes would just mean people would need to invent new ways to interpret those.

On the other hand being able to more easily manage multiple databases would be a nice optional feature.  I often use custom database directories for groups dicom data, like for a group of patients in an experiment or something.   Also if you put the data into a special-purpose database you have the option to use advanced loading modes.  Currently it's several steps to manage data, first to switch the database, then import the data.  Making this easier with a drag-and-drop would be nice.

So we could give users the option of (1) using a temporary database that goes away right away, (2) loading to the current database, or (3) selecting a directory to store the new sqlite files (the default could be in the directory you are importing from).

We could also make it easier to switch the current slicer.dicomDatabase.  Currently it's hidden a few clicks down in the DICOM module and doesn't have a list of recently used directories.

lassoan: Very interesting discussion.

Thinking a bit more about the described needs and workflows, I start to be a bit worried that using the "quick" method (bypassing the DICOM database) could lead to basic features (such as DICOM export) "randomly" not working.

For example, I load an image using the "quick" method, I segment it, and want to export the result as a DICOM segmentation object, but it fails, because it is not possible to save a DICOM segmentation object if the source image has no DICOM instance UIDs. I would be really frustrated and would consider Slicer's DICOM export feature to be broken.

Another example: If I load a CT image using the "quick" method then some of the Chest Imaging Platform's image analysis modules would not work, because they rely on reading the CT reconstruction kernel from the DICOM database. It would be completely unexpected that some features don't work if the DICOM image is loaded using the "quick" method.

> On the other hand being able to more easily manage multiple databases would be a nice optional feature.

This is a good point. I've updated the DICOM database some time ago to use relative path for referencing image files, so the database files can stored in the same folders as the DICOM files and can be moved anywhere together.

Maybe we should just make it easier to create a local database in the DICOM files folder? There may be some complications (e.g., the DICOM images are not always stored in writable folder), but maybe these are more manageable than the randomly breaking features depending on the DICOM file was loaded "quickly" or not.

Maybe we should just make it easier to load all the data sets right after importing using default options, and make it easy to remove the folder from the database? Maybe we should just automatically clear the DICOM database when Slicer quits?

jamesobutler: Should there be a separate node object that better indicates it is a DICOM object rather than regular vtkMRMLScalarVolumeNode that would have no DICOM database connection? I have been loading some DCM using the “quick” method and saving segmentations as my preferred .seg.nrrd type instead of a DICOM Segmentation object. I don’t care about the DICOM meta data components but only care about the image data and getting quantifiable segmentation results with my segmentation node object.

> Maybe we should just automatically clear the DICOM database when Slicer quits?

That could possibly work. Or no clearing if the database was purely a memory object and not corresponding to data saved to disk in a file

jamesobutler: If loaded .DCM files that are then segmented (or manipulated in other ways) are encourage to remain in .DCM world (such as saving as DICOM Segmentation) then currently that is not very clear. If I go to save my volume and segmentation node to disk with the normal "Save Data Dialog" I'm only presented with the non-dicom types which is why I prefer .seg.nrrd instead of .dicom.seg
image

It took my quite a while to find "Export to DICOM" for a segmentation object which I could only find in a right-click context menu for the segmentation node in a subject hierarchy area. This was not present in the file menu or within "Save Data Dialog". The concept between "Saving a file to disk" or "Exporting a file to disk" is confusing as a user.

pieper: > Maybe we should just automatically clear the DICOM database when Slicer quits?

Maybe when you load the data you could be given and explicit option to make it a temporary database that would have some value.  Often when I'm just testing some things I create a database in the /tmp directory on unix because I don't care if it goes away when the machine reboots.

But in general I use databases to organize data that I want to use over multiple Slicer sessions.

pieper: @jamesobutler there's no question that things are confusing right now with respect to dicom, which is why it's not the default save format.  I think the main question at hand in this PR is making sure things don't break or become more confusing for people.

jcfr: Thanks everyone for the very informative feedback. 

### DICOM Database lifecycle improvement

Improving how we manage the life cycle of the DICOM database seems like a reasonable step forward.

I envision these cases:

* Persistent database: This the current behavior where the same database is always reused (expect when `--testing` is specified)
* Session database: A new database is created at startup 

These command line options/settings would then be available:
* `--dicom-database-path=`: Path to an existing database 
* `--dicom-database-clear-on-exit`
* `--dicom-database-create-new` (equivalent to the behavior associated with `--testing`)

Passing `--dicom-database-path` and `--dicom-database-create-new` is an error case.

### DICOM Loading strategy

Loading DICOM using a temporary database (or ephemeral database) is supported. The corresponding MRML Node will still have its `DICOM.instanceUIDs` attribute set but there will be no way to retrieve corresponding DICOM information.

PR #7159, opened 2023-08-07
STYLE: Consistently use `canceled` and `canceling` spelling

Leengit:
These changes to make the spelling consistent compile and `ctest` on my local system.  However, _will they break folks who use 3D Slicer by accessing something by the previous spelling?_

These changes are achieved from Linux `bash` with:
```bash
cd Slicer/Slicer
egrep -Ril 'cancell(ed|ing)' * |
    tr \\n \\0 |
    xargs -r0 sed -i -r -e 's/(cancel)l(ed|ing)/\1\2/gi'
```

Also `signalling` -> `signaling` and `labelled` -> `labeled`.

### DIFF
--- a/Applications/SlicerApp/Testing/Python/CLIEventTest.py
+++ b/Applications/SlicerApp/Testing/Python/CLIEventTest.py
@@ -183,7 +183,7 @@ class CLIEventTestTest(ScriptedLoadableModuleTest):
 
     # Testing a the status event when canceling
     def test_CLIStatusEventTestCancel(self):
-        self.delayDisplay('Testing status events when cancelling the execution of a CLI')
+        self.delayDisplay('Testing status events when canceling the execution of a CLI')
 
         tempFile = qt.QTemporaryFile("CLIEventTest-outputFile-XXXXXX")
         self.assertTrue(tempFile.open())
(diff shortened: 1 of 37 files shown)


pieper:
I don't think we should change the API except in the context of a major release, and even then probably having a deprecation phase of a major release cycle would make sense since this is just a stylistic change.  Even better would be to support both spellings for methods and enums for a while.  Comments and log messages don't matter of course.  When we make the change we might also want to run your script on all the extensions and create PRs to switch to the new form.

Also it would be good to note the system you used for your code snippet.  I guess you used bash on linux.  It matters because  I've noticed macOS xargs behaves differently sometimes and it can be a pain to debug.

jcfr:
Thanks for the contribution. 

Since both spelling are valid in US English, I don't think we should consistently use the most prevalent form.

See https://www.scribbr.com/us-vs-uk/cancelled-or-canceled/

Leengit:
> I don't think we should change the API except in the context of a major release

Of course.  I take it from your response that this in fact does represent a change in the API.  So, let's not do it.

> Also it would be good to note the system you used for your code snippet.

Thanks.  I will push that correction before withdrawing this pull request.

pieper:
Thank @Leengit 

Yes, since the enums and the method names change it's a pretty big change.  I agree with Jc, that since it's not a clear-cut spelling error we can just let it stand as-is.

Leengit:
> It seems that the form with two l is the most prevalent.

The web sites that opine on this (including our corporate style guide) indicate that a single `l` is the US spelling and a double `ll` is the UK spelling.  Of course, both will find some use on both sides of the pond.

PR #7163, opened 2023-08-11
ENH: Add loadable warning in case of loading failure

cpoetteCertis:
The loadable warning flag can be used to manage loading failure if not empty

### DIFF
--- a/Modules/Scripted/DICOMLib/DICOMBrowser.py
+++ b/Modules/Scripted/DICOMLib/DICOMBrowser.py
@@ -150,7 +150,7 @@ class SlicerDICOMBrowser(VTKObservationMixin, qt.QWidget):
 
         self.dicomVisualBrowser.visible = self.useExpertimentalVisualDICOMBrowser
         self.dicomBrowser.visible = not self.useExpertimentalVisualDICOMBrowser
-        self.loadableTableFrame.visible = not self.useExpertimentalVisualDICOMBrowser
+        self.loadableTableFrame.visible = (not self.useExpertimentalVisualDICOMBrowser) and self.advancedView
         self.actionButtonsFrame.visible = not self.useExpertimentalVisualDICOMBrowser
         if self.useExpertimentalVisualDICOMBrowser and self.dicomVisualBrowser.patientsTabWidget().count == 0:
             self.dicomVisualBrowser.onShowPatients()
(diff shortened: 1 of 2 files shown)


jcfr:

Could you provide more details regarding the workflow that required you to explicitly set the `warning` attribute of the `DICOMLoadable` [^1] ?

My understanding is that the `warning` is usually set in the loading code of specific plugin. I suspect a specific plugin is not setting the `warning` attribute which in turn prevent the following code from working as expected:

https://github.com/Slicer/Slicer/blob/3f64c3780139447be3b0f2bbee3896dd3716ac54/Modules/Scripted/DICOMLib/DICOMBrowser.py#L561-L575


[^1]: https://github.com/Slicer/Slicer/blob/3f64c3780139447be3b0f2bbee3896dd3716ac54/Modules/Scripted/DICOMLib/DICOMPlugin.py#L24-L64


cc: @pieper @lassoan

lassoan:
I agree that it could be useful to provide a mechanism to determine result of a loading, but `warning` property of `DICOMLoadable` class is used for something else (for storing result of examine step as a user-displayable warning message), so we should not overwrite this property during loading.

We could specify new properties in the loadable, such as `loadingSuccess` (True or False) and `loadingMessage` (user-displayable message about the loading result). Or, we could add an optional `loadSuccess` (list of flags, one for each loadable) or `loadResults` argument (list of success flag, message, maybe evennodeIDs for each loadable) to  `loadLoadables`  method that would return more detailed information about which loadable failed and why.

@cpoetteCertis Could you write a bit more about your use case? What information you need and what do you do with it?

cpoetteCertis:
Dear All,

Both solutions seems suitable for us! 

Here are more details:
Depending on the loading issues, a node can be added (for example if irregular volume geometry is detected, the user can keep it or discard it). In that case, we need to provide all the necessary information for the given patient. 
Or the node is not added (for example if a scalar volume could not be loaded), in that case, no need to go further.

The problem is that currently we have no information about the loading state.

 Thanks to this modification, we simply check if the warning message is empty (success) or not empty (failure or issue while loading). If it is not empty, we do checks to know if a node has been created or not... 

But with your suggestions, it could be much easier!

lassoan:
@jcfr do you have any preference (add a list argument that is updated with loading status; add properties to each loadable object)?

lassoan (approved):
I've updated the code to store the loading result into a new `loadSuccess` attribute of the loadable.

PR #6541, opened 2022-09-20
ENH: Edit orientations in flythrough (DRAFT)

HarryDC:
Just for review with Raj ... 

Can edit an orientation for each control point
stopped at calcualting the list of orientations

Issue #6502, opened 2022-08-15
Upgrade endoscopy to support arbitrary camera poses and key frame interpolation


jadh4v:
## Is your feature request related to a problem? Please describe.

We need a flythrough animation feature in [VPAW](https://github.com/KitwareMedical/vpaw/issues/9). This is to aid in communicating aspects of a patient’s airway/anatomy to other clinicians. Current Endoscopy module comes very close to satisfying our requirements. However, it has some missing feature such as:
- Cannot attach/modify camera poses (camera always looking along the path).
- Needs keyframe interpolation algorithm(s).

## Describe the solution you'd like

Description of suggested solution copied / referenced from the [posted comment here](https://github.com/KitwareMedical/vpaw/issues/9#issuecomment-1142406408).
> A flythrough path is defined by creating a set of keyframes that store camera poses. The camera poses are interpolated between keyframes to produce smooth camera movement along the camera path.

Based on our internal discussions and from the [discourse discussion thread](https://discourse.slicer.org/t/upgrading-endoscopy-extension-with-camera-interpolation/23546), we have decided to modify the Endoscopy module to provide full support for a camera flythrough and keyframe interpolation. We can borrow specific implementations from CameraPath[^1] as/if needed.

[^1]: https://github.com/KitwareMedical/Slicer-CameraPath

### Keyframe Interpolation

(From Discourse)
- *Add some pure-VTK camera interpolation functionality, perhaps just one class based on [vtkQuaternionInterpolator](https://vtk.org/doc/nightly/html/classvtkQuaternionInterpolator.html#details) with some extra code to handle changes in field of view and maybe other parameters. This could go in `vtkAddon` if it’s C++ or in Endoscopy logic if it could be done in python (I’d think it could).*

### User Interface

(From Discourse)
- We could use the 3D view camera state itself as a way to insert keyframes. Initially we were considering having a [flight mode](https://github.com/KitwareMedical/Slicer/commit/9af308920309ec8cef890b98078574ef36de3737) for the 3D view to assist in flying around and placing keyframes, but since we can define the path curve first, it is probably not needed. The user can just fly along the path and stop and insert keyframes wherever necessary.

- Regarding display of keyframes we can simply use a camera mesh model to indicate the position or use Plane markups

- Along with camera pos and fov, would it make sense to also interpolate volume property, etc for better local control of visualization?
[...]

Issue #2579, opened 2020-03-12
Transform Widget:  easy enhancements for more user-friendly manual registrations


slicerbot:
_This issue was created automatically from an original [Mantis Issue](https://mantisarchive.slicer.org/view.php?id=2579). Further discussion may take place here._

lassoan:
It would be also useful to have widget in 2D and 3D views (similarly to markups widgets interactors). It should be possible to set any position for center of rotation and should be possible to translate/rotate both in world and object coordinate systems.

mauigna06:
I'd really like an implementation that works as this one:

https://youtu.be/7RMermYOOYY?t=98

Please note that it has two modes:
1) One to modify the handles
2) One to modify the object

Others are invited to give their opinions

--------------------------------

Could a roadmap of at least 5 steps be detailed?

mauigna06:
Ooh, here there is already a discussion about it:
https://discourse.slicer.org/t/new-transform-widget-for-slicer/20483

mauigna06:
> Last project week I started moving the interaction handles from Markups and making them more generic, and adding them to Transform nodes. This is the branch that contains the changes: [GitHub - Sunderlandkyl/Slicer at interaction_display_manager](https://github.com/Sunderlandkyl/Slicer/tree/interaction_display_manager)

@Sunderlandkyl I'm interested on implementing this, do you recommend starting from Slicer's preview release or your branch quoted above? :)

lassoan:
This had been implemented in Slicer-5.7:

https://github.com/Slicer/Slicer/pull/7562

PR #7165, opened 2023-08-11
ENH: Endoscopy fly through with user's camera transforms

Leengit:
The Endoscopy module currently supports a fly though, where the user can specify points on a path, and the module interpolates the points with splines and _computes_ a camera orientation at each point of interpolation.

The goal with this pull request is to allow the user to _specify_ a camera orientation (and possibly a camera translation too) at specified points.  These user-specified transformations will be interpolated to the same set of points on a spline so that the fly though will use the user-specified transformations (and their interpolations) instead of the currently implemented, module-computed orientations.

### DIFF
--- a/Modules/Scripted/Endoscopy/Endoscopy.py
+++ b/Modules/Scripted/Endoscopy/Endoscopy.py
@@ -304,7 +304,7 @@ class EndoscopyWidget(ScriptedLoadableModuleWidget, VTKObservationMixin):
             self.logic.cleanup()
         self.logic = EndoscopyLogic(inputCurve)
 
-        numberOfControlPoints = len(self.logic.resampledCurve)
+        numberOfControlPoints = self.logic.resampledCurve.GetNumberOfControlPoints()
 
         # Update frame slider range
         self.frameSlider.maximum = max(0, numberOfControlPoints - 2)
@@ -378,7 +378,7 @@ class EndoscopyWidget(ScriptedLoadableModuleWidget, VTKObservationMixin):
     def flyToNext(self):
         currentStep = int(self.frameSlider.value)
         nextStep = currentStep + self.skip + 1
-        if nextStep > len(self.logic.resampledCurve) - 2:
+        if nextStep > self.logic.resampledCurve.GetNumberOfControlPoints() - 2:
             nextStep = 0
         self.frameSlider.value = nextStep
 
@@ -388,13 +388,15 @@ class EndoscopyWidget(ScriptedLoadableModuleWidget, VTKObservationMixin):
         if (
             self.logic is None
             or self.logic.resampledCurve is None
-            or not 0 <= resampledCurvePointIndex < len(self.logic.resampledCurve)
+            or not 0 <= resampledCurvePointIndex < self.logic.resampledCurve.GetNumberOfControlPoints()
         ):
             return
 
-        resampledCurvePointIndex = int(resampledCurvePointIndex)
-        cameraPosition = self.logic.resampledCurve[resampledCurvePointIndex]
-        focalPointPosition = self.logic.resampledCurve[resampledCurvePointIndex + 1]
+        cameraPosition = np.zeros((3,))
+        self.logic.resampledCurve.GetNthControlPointPositionWorld(resampledCurvePointIndex, cameraPosition)
+
+        focalPoint = np.zeros((3,))
+        self.logic.resampledCurve.GetNthControlPointPositionWorld(resampledCurvePointIndex + 1, focalPoint)
 
         toParent = vtk.vtkMatrix4x4()
         self.transform.GetMatrixTransformToParent(toParent)
@@ -407,7 +409,7 @@ class EndoscopyWidget(ScriptedLoadableModuleWidget, VTKObservationMixin):
         # Y vector is aligned with the curve's plane normal.
         # This can be used for example to show a reformatted slice
         # using with SlicerIGT extension's VolumeResliceDriver module.
-        zVec = (focalPointPosition - cameraPosition) / np.linalg.norm(focalPointPosition - cameraPosition)
+        zVec = (focalPoint - cameraPosition) / np.linalg.norm(focalPoint - cameraPosition)
         yVec = self.logic.planeNormal
         xVec = np.cross(yVec, zVec)
         xVec /= np.linalg.norm(xVec)
@@ -428,7 +430,7 @@ class EndoscopyWidget(ScriptedLoadableModuleWidget, VTKObservationMixin):
         # Set the camera & cameraNode
         with slicer.util.NodeModify(self.cameraNode):
             self.cameraNode.SetPosition(cameraPosition)
-            self.cameraNode.SetFocalPoint(*focalPointPosition)
+            self.cameraNode.SetFocalPoint(*focalPoint)
             self.cameraNode.GetCamera().OrthogonalizeViewUp()
 
         self.cameraNode.ResetClippingRange()
@@ -504,14 +506,15 @@ class EndoscopyWidget(ScriptedLoadableModuleWidget, VTKObservationMixin):
 
 
 class EndoscopyLogic:
-    """Compute path given an input curve.
+    """Compute a path based on an input curve.
 
-    Path is stored in `resampledCurve` member variable as a numpy array and resampled path
-    is generated using `vtkMRMLMarkupsCurveNode.ResamplePoints()`.
+    The resulting path is stored in the `resampledCurve` member variable as a `slicer.vtkMRMLMarkupsCurveNode` object.
+    The points of the resampled path are generated using the `vtkMRMLMarkupsCurveNode.ResamplePoints()` method.
 
     Example:
       logic = EndoscopyLogic(inputCurve)
-      print(f"computed path has {logic.resampledCurve} elements")
+      print(f"computed path has {logic.resampledCurve.GetNumberOfControlPoints()} elements")
+
     """
 
     def __init__(self, inputCurve, dl=0.5):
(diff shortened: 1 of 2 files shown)


jcfr:
* https://github.com/Slicer/Slicer/issues/6502
* https://github.com/Slicer/Slicer/pull/6541

Related effort:
* https://github.com/SlicerMorph/SlicerMorph/tree/master/Animator
* https://github.com/KitwareMedical/Slicer-CameraPath ([discourse](https://discourse.slicer.org/t/support-for-keyframe-based-animation/6885) and [wiki](https://www.slicer.org/wiki/Documentation/Labs/FlyThroughNavigation))

Modules/Scripted/Endoscopy/Endoscopy.py
```diff
@@ -465,6 +523,42 @@ def step(self, segment, t, dl):
             pguess = p1
         return (t1, pguess, remainder)
 
+    def interpolateCameraTransforms(self):
+        # TODO: This approach works for the vtkMRMLMarkupsFiducialNode case.  We need to handle the
+        # vtkMRMLMarkupsCurveNode / vtkMRMLMarkupsClosedCurveNode case too, which is more difficult because it doesn't
+        # (yet) provide us self.pathIndices information.
```
jcfr, line 529: Since the `vtkMRMLMarkupsCurveNode`[^1] is based of `vtkMRMLMarkupsNode`[^2], it provides a convenient API to retrieve markup based of their index. See method named after the pattern `SetNthControlPoint/GetNthControlPoint`. The following functions should also be useful:
* `SetNthControlPointOrientationMatrix/GetNthControlPointOrientationMatrix`
* `SetNthControlPointOrientationMatrixWorld/GetNthControlPointOrientationMatrixWorld`

There is also an API
* 

[^1: https://github.com/Slicer/Slicer/blob/main/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsCurveNode.h
[^2]: https://github.com/Slicer/Slicer/blob/main/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsNode.h

jcfr:
@Leengit  Can you update the pull request description to provide a high level description ?

Leengit:
> @Leengit Can you update the pull request description to provide a high level overview of the proposed improvement ?

Done.

jcfr:
_Copied from https://github.com/Slicer/Slicer/issues/6502#issuecomment-1666450082_

> Markups curve provide a robust and smooth curve coordinate system (tangent and normal direction for each curve point), so it makes sense to use this and apply any specify camera pose in this coordinate system.

The description[^1] of the `vtkMRMLMarkupsCurveNode` class is helpful to understand the different coordinate system:

[^1]: https://github.com/Slicer/Slicer/blob/3bd4b9a2d0900612f560bd0c1a804ea010e2e35d/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsCurveNode.h#L53-L58

```
/// Coordinate systems used:
///   - Local: Current node's coordinate system where the position of the control and curve points are defined.
///            Local coordinates can be converted to world by concatenating all parent transforms on the current node.
///   - Surface: Model node's coordinate system where the polydata used for ShortestDistanceOnSurface pathfinding is defined.
///            Surface coordinates can be converted to world by concatenating all parent transforms on the surface node.
///   - World: Patient coordinate system (RAS)
```

> user to specify a camera orientation (and possibly a camera translation too) at specified points. 

For let's handle the orientation of the key frame by leveraging the `OrientationMatrix` associated with the markups.

jcfr:
To summarize our discussion, let's have the "path" be represented as curve node, this will allow to simplify the code.

We may still keep an export path as model to allow existing workflow to work.

jcfr:
> Draw tube

To help with testing, after installing the "SegmentEditorExtraEffects" extension, you will have access to the DrawTube[^1] effect. 

[^1]: https://github.com/lassoan/SlicerSegmentEditorExtraEffects/tree/master#draw-tube

Modules/Scripted/Endoscopy/Endoscopy.py
Leengit: Thank you.  I am using this interface now.

Leengit:
> To summarize our discussion, let's have the "path" be represented as curve node

Thank you.  We are now doing that.

Leengit:
> For let's handle the orientation of the key frame by leveraging the `OrientationMatrix` associated with the markups.

Thank you.  We are doing that now.

Leengit:
The version I just pushed, 22e109707594a9cf33ab1057364f5677bfd593cb, runs in that it successfully reproduces the _old behavior_, but with a new engine underneath.  The new engine

1. Uses a `vtkMRMLMarkupsCurveNode` to store and manipulate fly through data.
    a. Uses `vtkMRMLMarkupsCurveNode` functionality to do resampling to a user-supplied stepsize `dl` rather than re-inventing spline interpolation.
3. Organizes logic within an `EndoscopyLogic` class.
4. User cannot change the orientations from the default orientations (because there is no GUI), but these orientations are used and interpolated (via `vtkQuaternionInterpolator`) in the fly through.

Next steps include:

1. Add a GUI so that the user can specify the orientations at specific frames of the control path.

Leengit:
Rebased to the current main branch.

Leengit:
Squashed commits to ease rebase to latest `main` branch, which incorporates additional translatable strings, and rebased.

lassoan:
Most questions are rooted in the problem that [we would need a "transform" widget](https://github.com/Slicer/Slicer/issues/2579) but since there is no such widget and so we repurpose a plane widget. Probably the code cannot be much cleaner than it is now, unless a transform widget is implemented.

Leengit:
After discussion with @jcfr, I just force pushed a new set of commits, so that STYLE and ENH changes are in separate commits.

jcfr:
> force pushed a new set of commits

:pray: @Leengit 

Could you create the following standalone pull request:

------------

Create branch `documentation-update-md-files-replacing-literal-newlines` with a single commit titled:

```
DOC: Update markdown files replacing literal \n with newline

Facilitate visual parsing of the raw markdown files replacing literal "\n"
with actual newlines.
```

Update all markdown files found in [Docs](https://github.com/Slicer/Slicer/tree/main/Docs) directory

Leengit:
> Could you create the following standalone pull request:
> 
> Create branch `documentation-update-md-files-replacing-literal-newlines` with a single commit titled:

Done, please see #7340.

jcfr:
Can you rebase the topic against latest `main` branch ? This could be done by cherry-picking `Clarify variable names in Endoscopy.py` and `ENH: Improve Endoscopy flythrough`.

Leengit:
With this latest force push, I rebased to the current `main` branch.

jcfr:
To summarize our last discussion:
* [ ] Since `RemoveViewNodeID` expects to work on list including all IDs, add a logic function called `hideDisplayNodeFromView(viewNodeID)` to `EndoscopyLogic` and update the code to leverage it.
  * Ideally, we should refactor `qMRMLDisplayNodeViewComboBox` to use functions available in a logic class. We will create an issue to document this.
* [ ] Remove debug statements
* [ ] Update `Improve Endoscopy flythrough` commit to describe which improvement have been done

Leengit:
This is now at a point where I have reduced my bug list to zero.  (Next, I am going to make a nice video, so I may end up discovering additional bugs.)  I have not squashed the commits yet, in case folks wish to review only the latest changes.  Regardless, of which of the changes you will be reviewing, now would be a good time for reviews.  Thanks!

Leengit:
In case you want to comment on the interface.  First, here is the `main` branch's Endoscopy module interface:
Screenshot from 2023-11-14 08-39-36

Second, here is the pull request's Endoscopy module interface:
Screenshot from 2023-11-14 08-41-00

Leengit:
> * Remove "Create flythough and model" and observe modified event on the markup

... is still pending.  Because there could be several curves and closed curves in the Data, I don't want to create flythroughs for all of them simply because the Endoscopy module page is loaded.

jcfr:
>  Because there could be several curves and closed curves in the Data

It is true that there could be multiple curve, but only one is associated with the endoscopy module, and this is the one you would be observing.

Leengit:
Here's an updated interface view.
Screenshot from 2023-11-17 12-31-52

Leengit:
Do you think that "Use this camera" should instead be "Use this camera and curve"?

Leengit:
I have removed the "Draft" status and label from this pull request.  Please consider for merging.

Leengit:
With this last force push, I have rebased to the `upstream/main` branch.

jcfr:
We will also need to update the documentation of the module description the new controls and improved features.

See https://github.com/Slicer/Slicer/blob/main/Docs/user_guide/modules/endoscopy.md

Leengit:
> We will also need to update the documentation of the module description the new controls and improved features.

Thank you -- I've done that with the latest commit.

jcfr:
Thanks @Leengit, I will move forward the integration and follow-up if any questions

jcfr:
There is still more work to do to have functional module (e.g restore attribute saved into a scene)

Leengit:
Thank you for the merge early this morning.  It looks like it missed this morning's overnight build, so I will hold off until tomorrow to announce it on discourse.slicer.org.

jcfr:
Once the packages for  Slicer 5.6 are finalized, it will be available . See status at https://github.com/Slicer/Slicer/issues/7391

Leengit:
It would be good to have the functionality available in a nightly build so that people can see it at RSNA.  Is that going to be possible?

jcfr:
The updated Endoscopy module has been integrated and is available in the latest stable release. 

And will be available in tomorrow Preview build as well, hope this clarifies things.

jcfr:
For reference, issues associated with the topic originally associated with this pull request are described below.

To reproduce these, all Endoscopy related commits currently integrated in `main` have been reverted and the ones corresponding to a topic based of 0a5af4e0e6 are used[^1] instead.

[^1]: https://github.com/Slicer/Slicer/compare/57087f3c568...0a5af4e0e

### Automatic update of a flythrough path upon update of the curve

Steps to reproduce

* Start application
* Switch to `Dual 3D` view
* Create Open curve
* Select the Endoscopy module
* Click `Use this curve`
* Click `Play flythrough`
* Attempt to move a control point
* Error reported in the terminal :x: 

image

Same issue happen using these sequence:
* Select the Endoscopy module
* Click `Use this curve`
* Update control point position
* Click `Play flythrough`

### Curve is hidden if switching to `Dual 3D` after entering Endoscopy

* Start application
* Layout is `Conventional`
* Create Open curve
* Select the Endoscopy module
* Click `Use this curve`
* Select `Dual 3D` Layout
* Curve is not visible in the second view :x: 

image

Leengit:
A few changes happened between my pushed branches and the final merge.  In case this proves useful for the next steps, I offer these clarifications.

> commit 717fd6f2b89995a8503f63b888a18f081e5329d2
> Author: Lee Newberg <lee.newberg@kitware.com>
> Date:   Sun Nov 26 02:57:41 2023 -0500
> 
> ENH: Add support for managing orientation keyframe in Endoscopy
> 
> * Add support for saving, deleting and browsing keyframes
> * Store camera orientations via node attribute "Endoscopy.Path.CameraOrientations"
> * Interpolate saved orientations via quaternion
> 
> Co-authored-by: Jean-Christophe Fillion-Robin <jchris.fillionr@kitware.com>
> Co-authored-by: Harald Scheirich <harald.scheirich@kitware.com>

The interpolation is by quaternion slerp as provided by VTK's `vtkQuaternionInterpolator`.  We use the "linear spherical interpolation" rather than the "modified Kochanek basis" because the latter seems to use the supplied quaternions as guide points, but doesn't necessarily take a path through them.

> commit e3eceb6f0642cf90ebe034d445d90511359751af
> Author: Jean-Christophe Fillion-Robin <jchris.fillionr@kitware.com>
> Date:   Sun Nov 26 01:11:36 2023 -0500
> 
> ENH: Refactor EndoscopyLogic introducing setControlPoints()
> 
> * Simplify access to "resampledCurve" and "planeNormal" by directly
> accessing it though the logic instance.
> 
> * Remove static function calculatePlaneNormal() and directly call EndoscopyLogic.planeFit()
> 
> * Remove guard variable "ignoreInputCurveModified" and remove reset of logic
> when the curve node is modified.
> 
> Co-authored-by: Lee Newberg <lee.newberg@kitware.com>

The guard variable `ignoreInputCurveModified` was restored in a subsequent pull request so that processing in response to an event would not cause additional events to be processed.

PR #7167, opened 2023-08-12
BUG: Fix home shortcut not working when python console visible

### LINKED ISSUE #7154: Ctrl+H shortcut unexpectedly not working to bring up Home module
## Summary
I came across using the "Ctrl+H" shortcut to bring up the Home module where it wasn't working as I was expecting. Clicking View->Home worked, but the shortcut was not working. I observed in the python console area that it was reporting `[Qt] QAction::event: Ambiguous shortcut overload: Ctrl+H`. 

image

After searching for "Ctrl+H" and only finding 1 usage of it in Slicer core code, I had to search for it through the entire Slicer build tree and found that "Ctrl+H" is also a registered shortcut for the python console in [ctkConsole](https://github.com/commontk/CTK/blob/5657c58a72c2db3e9fcc04f88211859eccd21067/Libs/Widgets/ctkConsole.cpp#L145-L148). Therefore when I hid the python console the Ctrl+H shortcut began to work to bring up the Home module. It was confusing getting up to this point where the shortcut is for multiple actions where "Ctrl+H" for the python console was unknown to me however I did find one mention of it in the Keyboard/Shortcuts reference with not much mention of what it does or how it works.
image

In the Slicer documentation, "Ctrl+h" is listed as the shortcut for opening the default module (see [generic shortcuts](https://slicer.readthedocs.io/en/latest/user_guide/user_interface.html#generic-shortcuts)). There is also a section on shortcuts for the Python console (see [here](https://slicer.readthedocs.io/en/latest/user_guide/user_interface.html#python-console)), however it makes no mention of the "Ctrl+h" shortcut for the python console.

## Steps to reproduce

1. Open Slicer and switch to some other module that is not the Default/Home module. 
2. Make sure the python console is currently shown and docked in the main window.
3. Press "Ctrl+h" and observe that the home module is not brought up and `[Qt] QAction::event: Ambiguous shortcut overload: Ctrl+H` is reported in the python console.

## Expected behavior

I would expect that even with the python console shown that "Ctrl+H" would bring up the Default/Home module, or at least if my last mouse click was in the main Slicer interface and not in the python console that "Ctrl+H" would bring up the Default/Home module.

It appears the ctkConsole "Ctrl+H" functionality was added in https://github.com/commontk/CTK/commit/b3935fd5b6ed74e976dab405f1cfd24b8fce9fa6. Is this still important functionality? Can this shortcut be removed completely? Or removed from the context of 3D Slicer?

## Environment
- Slicer version: Slicer-5.3.0-2023-08-03
- Operating system: Windows

jamesobutler:
This closes #7154.

This removes the "Ctrl+H" shortcut defined for the ctkConsole so that it does not conflict with the "Ctrl+H" shortcut to go to the Slicer "Home" module.

Currently when the python console is visible "Ctrl+H" does not work for either of the defined slots and the following error is reported: [Qt] QAction::event: Ambiguous shortcut overload: Ctrl+H

### DIFF
--- a/Base/QTGUI/qSlicerApplication.cxx
+++ b/Base/QTGUI/qSlicerApplication.cxx
@@ -308,6 +308,16 @@ void qSlicerApplicationPrivate::init()
       << "slicer.mrmlScene"
       << "qt.QPushButton";
     q->pythonConsole()->completer()->setAutocompletePreferenceList(autocompletePreferenceList);
+    foreach(QAction* action, q->pythonConsole()->actions())
+      {
+      if (action->shortcut() == QKeySequence("Ctrl+H"))
+        {
+        // Remove action as "Ctrl+H" is reserved for going to the "Home" module
+        q->pythonConsole()->removeAction(action);
+        action->setParent(nullptr);
+        break;
+        }
+      }
     }
 #endif
 


jcfr (approved):
Thanks for the fix 🙏

This makes sense and is consistent with the documented behavior. See https://slicer.readthedocs.io/en/latest/user_guide/user_interface.html#generic-shortcuts

Issue #7092, opened 2023-07-14
Thick Slab Reconstruction corner-annotation not visible by default in Slice viewer


jcfr:
_Text below adapted from comment originally posted in https://github.com/Slicer/Slicer/issues/7021#issuecomment-1635212545_

## Is your feature request related to a problem? Please describe.

Following the integration of https://github.com/Slicer/Slicer/pull/7021, it is now possible to enable thick slab reconstruction from the Slice Controller.

The problem is that after enabling it, it is not shown by default in the top-right corner.

This happen because the corresponding corner is not active in the `Data Probe` module:

<img width=50% src="https://github.com/Slicer/Slicer/assets/219043/9fdd57c9-1b54-4bf8-8f76-088d41b9e1a9"/>

## Describe the solution you'd like

There are few paths forward to address this:

1. Change the default value for the top-right corner
2. Automatically toggle top-right corner to active
   1. And leave it enabled even after the slab reconstruction is toggled back off (should a confirmation dialog be displayed?)
   2. Or have a tri-state like settings
3. Update display logic to display the corner if the corner is active or if the associated function `vtkMRMLSliceNode::GetSlabReconstructionEnabled()` return True

It seems that moving forward with either `1` or `2.i` is sensible.

### Previous discussions

From @lassoan: 

> The Data Probe module's corner annotation handling is so bad that I don't mind whatever option is chosen. What we really need is a proper displayable manager that can display various dynamic text (orientation, slice thickness, DICOM fields, etc.) and content from any `vtkMRMLTextNode`.

_Originally posted by @lassoan in https://github.com/Slicer/Slicer/issues/7021#issuecomment-1635222773_

From @jamesobutler: 

> `1` or `2.i` is fine with me. 
>
> Our group doesn't use corner text annotations at all as our volume names have generally been long and we didn't like it getting in the way of actually viewing the image data.

_Originally posted by @jamesobutler in https://github.com/Slicer/Slicer/issues/7021#issuecomment-1635231784_

## Describe alternatives you've considered

Update documentation without implementing any changes.

## Additional context

* https://github.com/Slicer/Slicer/pull/7021
* https://github.com/Slicer/Slicer/issues/6965
* https://github.com/Slicer/Slicer/issues/7090

PR #7174, opened 2023-08-17
BUG: Sync DataProbe UI initialization and "Restore Defaults" action

jcfr:
Create a consolidated `DEFAULTS` dictionary to centralize default values. This guarantees consistent utilization of these values during both UI initialization and after clicking on the "Restore Defaults" button.

Fix a regression introduced in 2e8238c3e (`ENH: Added orientation marker to slice view annotations.`) where the values set during UI initialization and after the "Restore Defaults" action became different.

----


jcfr:
Merging as local testing was :heavy_check_mark:

PR #7175, opened 2023-08-17
ENH: Support saving and restoring SliceViewAnnotations display level

jcfr:


jcfr:
## CheckLists

* [x] No error after deleting main settings file, starting Slicer and selecting DataProbe
* [x] Start Slicer, set display level, exit and restart. Confirm value is restored
* [x] Start Slicer, click on "Restore Default"

@lassoan @pieper  Any reason we should not save/restore the `displayLevel` ?

PR #7176, opened 2023-08-17
ENH: Support restoring SliceViewAnnotations enabled state

jcfr:
Also simplify setting of checkbox state in SliceViewAnnotations

----

### DIFF
--- a/Modules/Scripted/DataProbe/DataProbeLib/SliceViewAnnotations.py
+++ b/Modules/Scripted/DataProbe/DataProbeLib/SliceViewAnnotations.py
@@ -193,10 +193,7 @@ class SliceAnnotations(VTKObservationMixin):
             self.layoutManager.connect("destroyed()", self.onLayoutManagerDestroyed)
 
     def onSliceViewAnnotationsCheckBox(self):
-        if self.sliceViewAnnotationsCheckBox.checked:
-            self.sliceViewAnnotationsEnabled = 1
-        else:
-            self.sliceViewAnnotationsEnabled = 0
+        self.sliceViewAnnotationsEnabled = int(self.sliceViewAnnotationsCheckBox.checked)
 
         settings = qt.QSettings()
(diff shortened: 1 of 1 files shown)


jcfr:
## CheckLists

* [ ] No error after deleting main settings file, starting Slicer and selecting DataProbe
* [ ] Start Slicer, set display level, exit and restart. Confirm value is restored
* [ ] Start Slicer, click on "Restore Default"

@lassoan @pieper  Any reason we should not save/restore the `displayLevel` ?

jcfr:
## CheckList

**Status:** Updated topic works as expected

* [x] No error after deleting main settings file, starting Slicer and selecting DataProbe
* [x] Start Slicer, disable, exit and restart. Confirm UI is still disabled
* [x] Click on "Restore Default"

jcfr:
@lassoan @pieper  Any reason we should not restore the `enabled` state as well ?

pieper:
> @lassoan @pieper Any reason we should not restore the `enabled` state as well ?

I'm not sure without looking in more detail but if you think it should be restored then I'm okay with it.

lassoan:
Slice view annotations is a mess. It was a quick prototype that got integrated to have this important feature in Slicer, but the design is really fragile and limited. We would need to rebuild it from scratch, using proper MRML-based design and flexibility and extensibility in mind, as overlay on image corners is very important space for displaying information.

pieper:
Many viewers also use these as UI controls/widgets too (e.g. control reslice rotation or window/level tools by clicking on the corresponding corner annotation can be logical and discoverable).

jcfr:
## CheckList

**Status:** Updated topic works as expected

* [x] No error after deleting main settings file, starting Slicer and selecting DataProbe
* [x] Start Slicer, disable, exit and restart. Confirm UI is still disabled
* [x] Click on "Restore Default"

sjh26 (approved):
lgtm

PR #7177, opened 2023-08-17
ENH: Ensure slab reconstruction thickness is visible in slice view corner

### LINKED ISSUE #7092: Thick Slab Reconstruction corner-annotation not visible by default in Slice viewer

jcfr:
Ensure that the visibility of the corner annotation displaying the thickness is automatically toggled without requiring the user to explicitly do so in the DataProbe module.

If the visibility of the corner annotation was enabled by "forcing", it will not be saved in the settings.

Fixes #7092

### DIFF
--- a/Modules/Scripted/DataProbe/DataProbeLib/SliceViewAnnotations.py
+++ b/Modules/Scripted/DataProbe/DataProbeLib/SliceViewAnnotations.py
@@ -111,6 +111,10 @@ class SliceAnnotations(VTKObservationMixin):
 
         self.maximumTextLength = 35
 
+        # Support forcing display of top right corner annotations if slab reconstruction is enabled
+        # See https://github.com/Slicer/Slicer/issues/7092
+        self.allowForcingTopRightVisibility = True
+
         self.create()
 
         if self.sliceViewAnnotationsEnabled:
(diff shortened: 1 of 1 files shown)


jcfr:
## CheckList

* [ ] No error after deleting main settings file and starting Slicer.
* [ ] Start Slicer, toggle slab reconstruction visibility, confirm corner annotation is visible and DataProbe "right-corner" checkbox is set to on
* [ ] Restart Slicer, confirm that DataProbe "right-corner" checkbox is still set to on, toggle slab reconstruction visibility, confirm corner annotation is visible

jcfr:
Following error is reported after step (3):

```
  File "/home/jcfr/Projects/Slicer-Release/Slicer-build/lib/Slicer-5.3/qt-scripted-modules/DataProbeLib/SliceViewAnnotations.py", line 364, in updateViewAnnotations
    self.makeAnnotationText(caller)
  File "/home/jcfr/Projects/Slicer-Release/Slicer-build/lib/Slicer-5.3/qt-scripted-modules/DataProbeLib/SliceViewAnnotations.py", line 480, in makeAnnotationText
    self.topRightCheckBox.checked = True
AttributeError: Trying to set property 'checked' on a destroyed QCheckBox object
```

PR #7187, opened 2023-08-22
BUG: Include closing segment of closed curve when appropriate.

Leengit:
Fix a bug where the closing segment (from the last control point to the first control point) in a closed curve is not incorporated into a `vtkMRMLMarkupsCurveNode::GetCurveLength` computation in all cases that it should be.  Specifically, a computation where
```c++
    startCurvePointIndex <= lastCurvePointIndex &&
    startCurvePointIndex + numberOfCurvePoints - 1 > lastCurvePointIndex
```
should include the closing segment's length even when the requested `numberOfCurvePoints` is not strictly greater than the number of control points.

### DIFF
--- a/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsCurveNode.cxx
+++ b/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsCurveNode.cxx
@@ -310,10 +310,12 @@ double vtkMRMLMarkupsCurveNode::GetCurveLength(vtkPoints* curvePoints, bool clos
     vtkGenericWarningMacro("Invalid startCurvePointIndex=" << startCurvePointIndex << ", using 0 instead");
     startCurvePointIndex = 0;
     }
-  vtkIdType lastCurvePointIndex = curvePoints->GetNumberOfPoints()-1;
-  if (numberOfCurvePoints >= 0 && startCurvePointIndex + numberOfCurvePoints - 1 < lastCurvePointIndex)
+  vtkIdType lastCurvePointIndex = curvePoints->GetNumberOfPoints() - 1;
+  bool addClosingSegment = closedCurve & startCurvePointIndex <= lastCurvePointIndex;
+  if (numberOfCurvePoints >= 0 && startCurvePointIndex + numberOfCurvePoints - 1 <= lastCurvePointIndex)
     {
     lastCurvePointIndex = startCurvePointIndex + numberOfCurvePoints - 1;
+    addClosingSegment = false;
     }
 
   double length = 0.0;
@@ -329,7 +331,7 @@ double vtkMRMLMarkupsCurveNode::GetCurveLength(vtkPoints* curvePoints, bool clos
     previousPoint[2] = nextPoint[2];
     }
   // Add length of closing segment
-  if (closedCurve && (numberOfCurvePoints < 0 || numberOfCurvePoints >= curvePoints->GetNumberOfPoints()))
+  if (addClosingSegment)
     {
     curvePoints->GetPoint(0, nextPoint);
     length += sqrt(vtkMath::Distance2BetweenPoints(previousPoint, nextPoint));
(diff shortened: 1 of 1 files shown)


Modules/Loadable/Markups/MRML/vtkMRMLMarkupsCurveNode.cxx
```diff
@@ -329,7 +329,8 @@ double vtkMRMLMarkupsCurveNode::GetCurveLength(vtkPoints* curvePoints, bool clos
     previousPoint[2] = nextPoint[2];
     }
   // Add length of closing segment
-  if (closedCurve && (numberOfCurvePoints < 0 || numberOfCurvePoints >= curvePoints->GetNumberOfPoints()))
+  if (closedCurve && startCurvePointIndex <= lastCurvePointIndex &&
```
lassoan, line 332: This entire logic is becoming just too hard to read (and therefore error-prone and may be misinterpreted by users when they try to understand what the code does). What do you think about setting an `addClosingSegment` Boolean flag during the computation of lastCurvePointIndex? For example, the flag could be enabled by default if curve is closed but it would be set to false if it is detected that only a partial curve length is requested that does not include the closing segment (e.g., it could be set around line 316 where we update the `lastCurvePointIndex`).

In the end then we would just check the `addClosingSegment` flag and add the closing segment length.

Leengit: Sounds good.  I just force pushed.

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsCurveNode.cxx
```diff
@@ -310,10 +310,12 @@ double vtkMRMLMarkupsCurveNode::GetCurveLength(vtkPoints* curvePoints, bool clos
     vtkGenericWarningMacro("Invalid startCurvePointIndex=" << startCurvePointIndex << ", using 0 instead");
     startCurvePointIndex = 0;
     }
-  vtkIdType lastCurvePointIndex = curvePoints->GetNumberOfPoints()-1;
-  if (numberOfCurvePoints >= 0 && startCurvePointIndex + numberOfCurvePoints - 1 < lastCurvePointIndex)
+  vtkIdType lastCurvePointIndex = curvePoints->GetNumberOfPoints() - 1;
+  bool addClosingSegment = closedCurve & startCurvePointIndex <= lastCurvePointIndex;
```
lassoan, line 314: Why do we need the `startCurvePointIndex <= lastCurvePointIndex` check?

Leengit: My thinking ... if, say, there are exactly 5 control points and the user asks for 7 points starting with number 9, we want to return 0.0 rather than the length of the closing segment.

lassoan: OK, let's take care of that explicitly, earlier - where we check for `startCurveIndex < 0`. We could add a check for `startCurveIndex >= curvePoints->GetNumberOfPoints()` for an open curve and `startCurveIndex > curvePoints->GetNumberOfPoints()` for a closed curve and return 0 immediately.

Leengit:
To my novice eyes, the build failure appears to be unrelated.  Is that something that you can confirm @lassoan?

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsCurveNode.cxx
```diff
@@ -305,15 +305,22 @@ double vtkMRMLMarkupsCurveNode::GetCurveLength(vtkPoints* curvePoints, bool clos
     {
     return 0.0;
     }
+  if (startCurvePointIndex >= curvePoints->GetNumberOfPoints() + closedCurve)
```
lassoan, line 308: I know this is correct but making things more explicit would improve readability:
```suggestion
  // In case of closed curve we allow one more point beyond the curve points
  // (corresponding to the closing segment).
  if (startCurvePointIndex >= curvePoints->GetNumberOfPoints() + (closedCurve ? 1 : 0))
```

lassoan:
Thank you @Leengit, this is now merged. The build error was unrelated.

Issue #6177, opened 2022-02-11
Some strings cannot be translated to other languages


lassoan:
Not all strings that are visible on the GUI show up in the list of translatable strings. Some strings show up but they do not get translated.

- [x] Module names that are visible on the GUI (i.e., module title) are not translated. For example: Volumes, Models, Markups (although they are available in the module widget) - see https://github.com/Slicer/Slicer/issues/6826
- [x] Python scripted modules are not translatable yet (except the GUI that is loaded from .ui file)
  - https://github.com/Slicer/Slicer/commit/60c659f33b672d8e8a7193f0741a2f56aaac080c (originally reviewed through https://github.com/Slicer/Slicer/pull/6920)
- [x] CLI modules are not translatable yet
  - fixed in https://github.com/Slicer/Slicer/pull/6681
- [x] Extensions are not translatable yet - done: see instructions at https://github.com/Slicer/SlicerLanguagePacks/blob/main/DevelopersManual.md#how-to-make-an-extension-translatable
- [x] Various examples of non-translatable strings: https://discourse.slicer.org/t/some-words-are-not-extracted-to-crowdin/21908 -- screenshots are no longer visible, but probably we already fixed those issues
- [x] Translation file must be added for CTK
  - https://github.com/Slicer/Slicer/pull/6926
  - https://github.com/commontk/CTK/pull/1076
  - https://github.com/commontk/CTK/pull/1079
  - https://github.com/commontk/CTK/pull/1077
  - https://github.com/commontk/CTK/pull/1068
- [x] Color table names are not translatable (see in Volumes module / Display section)
  - the issue is now tracked in https://github.com/Slicer/SlicerLanguagePacks/issues/59
- [x] Some strings appear in weblate but are actually non-translatable. This includes string properties in .ui files, such as `nodeTypes` in `qMRMLNodeComboBox`.

Documentation improvements:
- [x] Add documentation for Data Probe (describe what L, F, B labels mean = label, foreground, background) -- this is now tracked in https://github.com/Slicer/SlicerLanguagePacks/issues/60

lassoan:
Yes, these issues still persist. They are due to that in current Slicer source code some dynamically created strings are not marked as translatable and we would need to implement collection of translatable strings from extensions.

lassoan:
Non-translatable strings:
- [ ] Color table names are not translatable (see in Volumes module / Display section)

Documentation improvements:
[...]

PR #7188, opened 2023-08-23
ENH: Add translation function for logic classes

mhdiop:
* Slicer/SlicerLanguagePacks#12
* Slicer/Slicer#6647
* Slicer/Slicer#6177

### DIFF
--- a/Base/QTCore/qMRMLTranslator.h
+++ b/Base/QTCore/qMRMLTranslator.h
@@ -1,0 +1,16 @@
+#ifndef qMRMLTranslator_h
+#define qMRMLTranslator_h
+
+#include "vtkMRMLTranslator.h"
+
+#include "qSlicerBaseQTCoreExport.h"
+
+class Q_SLICER_BASE_QTCORE_EXPORT qMRMLTranslator: public vtkMRMLTranslator
+{
+public:
+
+  /// Translation function for logic classes
+  std::string Translate(const char *context, const char *sourceText, const char *disambiguation = nullptr, int n = -1) override;
+};
+
+#endif
--- a/Base/QTCore/qSlicerCoreApplication.cxx
+++ b/Base/QTCore/qSlicerCoreApplication.cxx
@@ -85,6 +85,7 @@
 #include "qSlicerModuleFactoryManager.h"
 #include "qSlicerModuleManager.h"
 #include "qSlicerUtils.h"
+#include "qMRMLTranslator.h"
 
 // SlicerLogic includes
 #include "vtkDataIOManagerLogic.h"
@@ -351,6 +352,12 @@ void qSlicerCoreApplicationPrivate::init()
     modifiedRequestCallback->SetClientData(this->AppLogic);
     modifiedRequestCallback->SetCallback(vtkSlicerApplicationLogic::RequestModifiedCallback);
     vtkEventBroker::GetInstance()->SetRequestModifiedCallback(modifiedRequestCallback);
+  }
+
+  // Create the translation object used for string translation in logic classes
+  if (!this->AppLogic->GetTranslatorInstance())
+  {
+    this->AppLogic->SetTranslatorInstance(new qMRMLTranslator());
   }
 
   // Ensure that temporary folder is writable
--- a/Libs/MRML/Logic/vtkMRMLApplicationLogic.h
+++ b/Libs/MRML/Logic/vtkMRMLApplicationLogic.h
@@ -27,6 +27,8 @@
 #include "vtkMRMLLogicExport.h"
 #include "vtkMRMLSliceCompositeNode.h"
 
+#include "vtkMRMLTranslator.h"
+
 class vtkMRMLColorLogic;
 class vtkMRMLModelDisplayNode;
 class vtkMRMLSliceNode;
@@ -289,6 +291,15 @@ public:
   /// Font file path is set to the one specified in FontFileName property in this object.
   void UseCustomFontFile(vtkTextProperty* textProperty);
 
+  /// Translation function for logic classes
+  std::string Translate(const char *context, const char *sourceText, const char *disambiguation = nullptr, int n = -1);
+
+  /// @{
+  /// Set the object used for string translation in logic classes
+  static void SetTranslatorInstance(vtkMRMLTranslator* translatorInstance);
+  static vtkMRMLTranslator* GetTranslatorInstance();
+  /// @}
+
 protected:
 
   vtkMRMLApplicationLogic();
@@ -315,6 +326,8 @@ private:
 
   class vtkInternal;
   vtkInternal* Internal;
+
+  static vtkMRMLTranslator* TranslatorInstance;
 
 };
 
--- a/Libs/MRML/Logic/vtkMRMLTranslator.cxx
+++ b/Libs/MRML/Logic/vtkMRMLTranslator.cxx
@@ -1,0 +1,6 @@
+#include "vtkMRMLTranslator.h"
+
+std::string vtkMRMLTranslator::Translate(const char *vtkNotUsed(context), const char *sourceText, const char *vtkNotUsed(disambiguation), int vtkNotUsed(n))
+{
+  return sourceText;
+}
--- a/Libs/MRML/Logic/vtkMRMLTranslator.h
+++ b/Libs/MRML/Logic/vtkMRMLTranslator.h
@@ -1,0 +1,18 @@
+#ifndef vtkMRMLTranslator_h
+#define vtkMRMLTranslator_h
+
+#include "vtkMRMLLogicExport.h"
+
+// STD includes
+
+#include <string>
+
+class VTK_MRML_LOGIC_EXPORT vtkMRMLTranslator
+{
+public:
+
+  /// Default noop translation function for logic classes
+  virtual std::string Translate(const char *context, const char *sourceText, const char *disambiguation = nullptr, int n = -1);
+};
+
+#endif
--- a/Base/QTCore/qMRMLTranslator.cxx
+++ b/Base/QTCore/qMRMLTranslator.cxx
@@ -1,0 +1,8 @@
+#include <QCoreApplication>
+
+#include "qMRMLTranslator.h"
+
+std::string qMRMLTranslator::Translate(const char *context, const char *sourceText, const char *disambiguation, int n)
+{
+  return QCoreApplication::translate(context, sourceText, disambiguation, n).toStdString();
+}
--- a/Libs/MRML/Logic/vtkMRMLApplicationLogic.cxx
+++ b/Libs/MRML/Logic/vtkMRMLApplicationLogic.cxx
@@ -1254,3 +1254,32 @@ std::string vtkMRMLApplicationLogic::GetFontsDirectory()
   std::string fullPath = vtksys::SystemTools::JoinPath(filesVector);
   return fullPath;
 }
+
+//----------------------------------------------------------------------------
+std::string vtkMRMLApplicationLogic::Translate(const char *context, const char *sourceText, const char *disambiguation, int n)
+{
+  if (vtkMRMLApplicationLogic::TranslatorInstance)
+  {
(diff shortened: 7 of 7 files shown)


jcfr:
@mhdiop If you edit the description of the issue like this, it will render nicely:

| Current | Proposed |
|--|--|
| image | image | 
|  image | image |

jcfr:
As outlined in the build log[^1], the linking is currently failing on Linux:

[^1]: https://github.com/Slicer/Slicer/actions/runs/5946372990/job/16126820100?pr=7188

```
[1537/5178] Linking CXX shared library bin/libMRMLLogic.so
FAILED: bin/libMRMLLogic.so 
...
Libs/MRML/Logic/CMakeFiles/MRMLLogic.dir/vtkMRMLApplicationLogic.cxx.o: In function `vtkMRMLApplicationLogic::GetTranslatorInstance()':
vtkMRMLApplicationLogic.cxx:(.text+0x1c97): undefined reference to `vtkMRMLApplicationLogic::TranslatorInstance'
Libs/MRML/Logic/CMakeFiles/MRMLLogic.dir/vtkMRMLApplicationLogic.cxx.o: In function `vtkMRMLApplicationLogic::SetTranslatorInstance(vtkMRMLTranslator*)':
vtkMRMLApplicationLogic.cxx:(.text+0x1cd7): undefined reference to `vtkMRMLApplicationLogic::TranslatorInstance'
Libs/MRML/Logic/CMakeFiles/MRMLLogic.dir/vtkMRMLApplicationLogic.cxx.o: In function `vtkMRMLApplicationLogic::Translate(char const*, char const*, char const*, int)':
vtkMRMLApplicationLogic.cxx:(.text+0x1d1b): undefined reference to `vtkMRMLApplicationLogic::TranslatorInstance'
collect2: error: ld returned 1 exit status
[1538/5178] Linking CXX shared library lib/Slicer-5.5/ITKFactories/libMRMLIDIOPlugin.so
[1539/5178] Building CXX object Libs/MRML/Core/CMakeFiles/MRMLCorePython.dir/MRMLCorePythonInit.cxx.o
ninja: build stopped: subcommand failed.
```

Libs/MRML/Logic/vtkMRMLTranslator.h
```diff
@@ -0,0 +1,13 @@
+#ifndef vtkMRMLTranslator_h
+#define vtkMRMLTranslator_h
+
```
jcfr, line 3:
```suggestion

#include "vtkMRMLLogicExport.h"

// STD includes
#include <string>
```

Libs/MRML/Logic/vtkMRMLTranslator.h
```diff
@@ -0,0 +1,13 @@
...
+class vtkMRMLTranslator
```
jcfr, line 4:
```suggestion
class VTK_MRML_LOGIC_EXPORT vtkMRMLTranslator
```

Base/QTCore/qMRMLTranslator.h
```diff
@@ -0,0 +1,14 @@
+#ifndef qMRMLTranslator_h
+#define qMRMLTranslator_h
+
```
jcfr, line 3:
```suggestion

#include "qSlicerBaseQTCoreExport.h"

```

Base/QTCore/qMRMLTranslator.h
```diff
@@ -0,0 +1,14 @@
...
+#include "vtkMRMLTranslator.h"
+
+class qMRMLTranslator: public vtkMRMLTranslator
```
jcfr, line 6:
```suggestion
class Q_SLICER_BASE_QTCORE_EXPORT qMRMLTranslator: public vtkMRMLTranslator
```

Base/QTCore/qMRMLTranslator.h
```diff
@@ -0,0 +1,14 @@
...
+{
+public:
+
+  // Translation function for logic classes
```
jcfr, line 10:
```suggestion
  /// Translation function for logic classes
```

Libs/MRML/Logic/vtkMRMLTranslator.cxx
```diff
@@ -0,0 +1,6 @@
+#include "vtkMRMLTranslator.h"
+
+std::string vtkMRMLTranslator::Translate(const char *context, const char *sourceText, const char *disambiguation, int n)
```
jcfr, line 3:
```suggestion
std::string vtkMRMLTranslator::Translate(const char *vtkNotUsed(context), const char *sourceText, const char *vtkNotUsed(disambiguation), int vtkNotUsed(n))
```

Libs/MRML/Logic/vtkMRMLTranslator.h
```diff
@@ -0,0 +1,13 @@
...
+{
+public:
+
+  // Default noop translation function for logic classes
```
jcfr, line 8:
```suggestion
  /// Default noop translation function for logic classes
```

Base/QTCore/qSlicerCoreApplication.cxx
```diff
@@ -353,6 +354,14 @@ void qSlicerCoreApplicationPrivate::init()
...
   }
 
+  // Create the translation object used for string translation in logic classes
+  {
+    if (!this->AppLogic->GetTranslatorInstance())
+    {
+      this->AppLogic->SetTranslatorInstance(new qMRMLTranslator());
+    }
+  }
+
```
jcfr, line 364:
```suggestion
  if (!this->AppLogic->GetTranslatorInstance())
  {
    this->AppLogic->SetTranslatorInstance(new qMRMLTranslator());
  }
```

jcfr: There is no need for extra scope because there are no _stack allocated_[^1] object that need to be removed before reaching the end of the function. Above and below, we wanted the `vtkNew` and `QTemporaryFile` object to be "deleted".

[^1]: https://courses.engr.illinois.edu/cs225/fa2022/resources/stack-heap/

Libs/MRML/Logic/vtkMRMLApplicationLogic.h
```diff
@@ -289,6 +291,15 @@ class VTK_MRML_LOGIC_EXPORT vtkMRMLApplicationLogic
   /// Font file path is set to the one specified in FontFileName property in this object.
   void UseCustomFontFile(vtkTextProperty* textProperty);
 
+  // Translation function for logic classes
```
jcfr, line 294:
```suggestion
  /// Translation function for logic classes
```

jcfr: I suggest you review the other comments so that they are consistently specified.

Libs/MRML/Logic/vtkMRMLApplicationLogic.h
```diff
@@ -289,6 +291,15 @@ class VTK_MRML_LOGIC_EXPORT vtkMRMLApplicationLogic
...
+  std::string Translate(const char *context, const char *sourceText, const char *disambiguation = nullptr, int n = -1);
+
+  // Get the object used for string translation in logic classes
```
jcfr, line 297: Consider grouping (using `/// @{` and `/// }@`) the Set/Get like done elsewhere in the file.

jcfr (changes requested):

Libs/MRML/Logic/vtkMRMLTranslator.h
lassoan: Adding `VTK_MRML_LOGIC_EXPORT` designates the class to be visible on the public API of the shared library. `VTK_MRML_LOGIC_EXPORT` is a compiler/linker-specific macro defined in an automatically generated .h file.

mhdiop: Thank you, I better understand now !

Base/QTCore/qMRMLTranslator.h
```diff
@@ -0,0 +1,16 @@
+#ifndef qMRMLTranslator_h
+#define qMRMLTranslator_h
+
+#include "vtkMRMLTranslator.h"
+
+#include "qSlicerBaseQTCoreExport.h"
+
+class Q_SLICER_BASE_QTCORE_EXPORT qMRMLTranslator: public vtkMRMLTranslator
```
lassoan, line 8: Since this class is derived from a VTK base class, it has to be a VTK class, i.e., `vtkMRMLTranslator`.

Base/QTCore/qSlicerCoreApplication.cxx
```diff
@@ -353,6 +354,12 @@ void qSlicerCoreApplicationPrivate::init()
     vtkEventBroker::GetInstance()->SetRequestModifiedCallback(modifiedRequestCallback);
   }
 
+  // Create the translation object used for string translation in logic classes
+  if (!this->AppLogic->GetTranslatorInstance())
+  {
+    this->AppLogic->SetTranslatorInstance(new qMRMLTranslator());
```
lassoan, line 360: qMRMLTranslator is not a Qt class, and it is not taking care of destroying itself if it is created by new. Instead, it is a VTK class, which can be created for example by vtkNew:

```python
vtkNew<vtkMRMLTranslator> translator;
this->AppLogic->SetTranslatorInstance(translator);
```

Libs/MRML/Logic/vtkMRMLApplicationLogic.h
```diff
@@ -316,6 +327,8 @@ class VTK_MRML_LOGIC_EXPORT vtkMRMLApplicationLogic
   class vtkInternal;
   vtkInternal* Internal;
 
+  static vtkMRMLTranslator* TranslatorInstance;
```
lassoan, line 330: When is this going to be deleted?
How can you ensure that it is deleted at the right time (not too early - when something may still call it; but not too late - after the VTK leak detection mechanism already counted all the non-deleted VTK objects)?

How would we implement this correctly on Windows, Linux, and macOS? We should not reinvent this from scratch but follow a singleton pattern already used in Slicer or VTK.

jcfr: In this particular case, I don't think we need to use [vtkSingleton.h](https://github.com/Slicer/Slicer/blob/main/Libs/MRML/DisplayableManager/vtkSingleton.h) help class.

Since the lifecycle of this class is driven by `vtkMRMLApplicationLogic` (itself instantiated and deleted by the main application), the reference to `vtkMRMLTranslator` will be released upon destruction of `vtkMRMLApplicationLogic`.

Base/QTCore/qMRMLTranslator.h
jcfr: This particular class should probably be called `qSlicerQtVTKTranslationBridge` ?

Note that since its role is to establish a "bridge" so that VTK logic classes can indirectly leverage Qt translation capabilities through polymorphism, it doesn't have to strictly start with the prefix `vtk`.

That said, may be the name `vtkSlicerQtVTKTranslationBridge` would be better ?

lassoan: `vtk` prefix is an important hint about how the class is allowed to be constructed and used. `vtk...` classes are based on `vtkObject` and rely on internal reference counting. So, for clarity and consistency we either use a proper VTK class (with appropriate naming, internal reference counting, private constructor, etc.) or we don't make the translator a VTK class at all. We could make it a Qt class (derived from QObject, following Qt conventions) then, or just make it a simple C++ class (we establishing all rules).

Since initialization and destruction of singleton classes is a difficult topic, I would use already established singleton patterns in Slicer or VTK. The current design seems to be some custom implementation, which may not work correctly in a large application, composed of many shared libraries.

Libs/MRML/Logic/vtkMRMLApplicationLogic.h
pieper: +1 for making a simple vtk class that is owned by the application logic.

lassoan:
Superseded by https://github.com/Slicer/Slicer/pull/7210

PR #7210, opened 2023-08-31
Make text in MRML library translatable

lassoan:
Allows using Qt translation infrastructure in MRML to translate text, without directly depending on Qt.

Fixes https://github.com/Slicer/Slicer/issues/6647


pieper (approved):
Nice, very clean!

It seems like a lot of bookkeeping, but I can't think of a better way to do it.

pieper:
Feel free to merge this whenever you think it's done.

jcfr:
> re: Singleton

We have encapsulated the logic into `Libs/MRML/DisplayableManager/vtkSingleton.h`, I am wondering if we should move it into `vtkAddon` so that it can be more broadly used,

PR #7212, opened 2023-09-02
BUG: Fix "No input data" message in Segment Editor paint effect

lassoan:
If there is no input or if the input has no points, the vtkTransformPolyDataFilter will display an error message on every update: "No input data", polluting the log and slowing down the application. To prevent logging the error, an error callback command is set that discards the message.

### DIFF
--- a/Modules/Loadable/Segmentations/EditorEffects/qSlicerSegmentEditorPaintEffect.cxx
+++ b/Modules/Loadable/Segmentations/EditorEffects/qSlicerSegmentEditorPaintEffect.cxx
@@ -45,6 +45,7 @@
 #include <vtkCellPicker.h>
 #include <vtkCollection.h>
 #include <vtkCommand.h>
+#include <vtkCallbackCommand.h>
 #include <vtkGlyph2D.h>
 #include <vtkGlyph3D.h>
 #include <vtkIdList.h>
@@ -153,6 +154,15 @@ public:
     feedbackActorProperty->SetOpacity(0.5);
     this->FeedbackActor->SetMapper(this->FeedbackMapper);
     this->FeedbackActor->VisibilityOff();
+
+    // If there is no input or if the input has no points, the vtkTransformPolyDataFilter will display an error message
+    // on every update: "No input data", polluting the log and slowing down the application.
+    // To prevent logging the error, an error callback command is set that discards the message.
+    // Ideally, vtkTransformPolyDataFilter should be updated to not log this unnecessary error message.
+    this->ErrorSinkCallbackCommand = vtkSmartPointer<vtkCallbackCommand>::New();
+    this->BrushWorldToSliceTransformer->AddObserver(vtkCommand::ErrorEvent, this->ErrorSinkCallbackCommand);
+    this->FeedbackWorldToSliceTransformer->AddObserver(vtkCommand::ErrorEvent, this->ErrorSinkCallbackCommand);
+
     };
   ~BrushPipeline2D() override = default;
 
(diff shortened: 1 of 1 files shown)


Modules/Loadable/Segmentations/EditorEffects/qSlicerSegmentEditorPaintEffect.cxx
```diff
@@ -153,6 +154,15 @@ class BrushPipeline2D : public BrushPipeline
     feedbackActorProperty->SetOpacity(0.5);
     this->FeedbackActor->SetMapper(this->FeedbackMapper);
     this->FeedbackActor->VisibilityOff();
+
+    // If there is no input or if the input has no points, the vtkTransformPolyDataFilter will display an error message
+    // on every update: "No input data", polluting the log and slowing down the application.
+    // To prevent logging the error, an error callback command is set that discards the message.
+    // Ideally, vtkTransformPolyDataFilter should be updated to not log this unnecessary error message.
```
pieper, line 161: maybe make an upstream merge request and then cherry pick it into our vtk fork?

Modules/Loadable/Segmentations/EditorEffects/qSlicerSegmentEditorPaintEffect.cxx
```diff
@@ -177,6 +187,7 @@ class BrushPipeline2D : public BrushPipeline
   vtkSmartPointer<vtkTransformPolyDataFilter> BrushWorldToSliceTransformer;
   vtkSmartPointer<vtkCutter> FeedbackCutter;
   vtkSmartPointer<vtkTransformPolyDataFilter> FeedbackWorldToSliceTransformer;
+  vtkSmartPointer<vtkCallbackCommand> ErrorSinkCallbackCommand;
```
pieper, line 190: Won't this mask other potentially meaningful errors?

lassoan: The filter does not provide any meaningful error messages. It logs an error if there are no points in the input polydata (which is plain wrong) or if the input transform is not defined (this might be marginally useful, but still unnecessary, as the filter could simple interpret "no transformation is specifed" as "no transformation should be performed).

Modules/Loadable/Segmentations/EditorEffects/qSlicerSegmentEditorPaintEffect.cxx
lassoan: Yes, of course, that would be better. We have 3 other workarounds because of this VTK error (search for "No input data" in the Slicer source code). I'll try to put together a pull reuest that removes the message in VTK.

pieper: It looks like VTK is pretty inconsistent, but probably all of these should be `vtkDebugMacro`

https://gitlab.kitware.com/search?search=No%20input%20data&nav_source=navbar&project_id=13&group_id=16&search_code=true&repository_ref=master

lassoan: VTK just logs this error in 23 classes, out of a total of about 3000 classes. In vtkTransformPolyDataFilter the message was added when the filter was originally developed in 1995. The other occurrences may be just copy-paste or time saving for the developer (the developer decided to just return an error and let the user deal with the situation instead of making sure the class works well with empty input). I've created a merge request in VTK to remove the error message (the filter seems to work well for empty input):

https://gitlab.kitware.com/vtk/vtk/-/merge_requests/10472

cpinter:
Thank you for working on this Andras!

jcfr:
The corresponding change has been back-ported to `Slicer/VTK` fork:
* https://github.com/Slicer/Slicer/pull/7227

I then suggest we close this pull request.

PR #7214, opened 2023-09-04
Empty folder SH visibility

cpinter:
- The SH tree view saves the "current tree view" pointer in the SH plugin handler
- When the user creates a new folder, the folder plugin enables showing empty folders in the sort filter proxy model of the current tree view, and stores the tree view pointer in a list
- Right-clicking on scene or any item, the folder plugin shows a new "Show empty folders" action if the user has created a new folder from that tree view (i.e. the current tree view is in the list)
- The DICOM plugin (providing the Subject and Study hierarchy items) behaves the same way as the Folder plugin
- New plugin created for the Expand to depth feature (used to be a core SH tree view feature)

### DIFF
--- a/Modules/Loadable/SubjectHierarchy/Widgets/CMakeLists.txt
+++ b/Modules/Loadable/SubjectHierarchy/Widgets/CMakeLists.txt
@@ -49,6 +49,8 @@ set(${KIT}_SRCS
   qSlicerSubjectHierarchyVisibilityPlugin.h
   qSlicerSubjectHierarchyExportPlugin.cxx
   qSlicerSubjectHierarchyExportPlugin.h
+  qSlicerSubjectHierarchyExpandToDepthPlugin.cxx
+  qSlicerSubjectHierarchyExpandToDepthPlugin.h
   )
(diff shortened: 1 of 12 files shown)


lassoan (commented):
Thank you, the code looks good, I'll do some testing and if everything works well then I'll merge.

lassoan:
It seems that the original issue is still present: clicking the eye icon for an empty folder in Markups module's SH tree makes the new folder disappear:

https://github.com/Slicer/Slicer/assets/307929/f2c94948-ec87-40e2-83df-e0f54492c2cf

cpinter:
Hm, wow. I'll check now. Thanks a lot for testing.

lassoan:
Thank you for the fix!

Could you merge this fix with the original commit that introduced the feature? The commit comment could be added as explanation/comment in the code if needed (if not explained there sufficiently already).

cpinter:
Done. I squashed the three folder (and patient/study) related commits into one, and left the expand action one separately.

lassoan (approved):
Thank you, this looks good

cpinter:
Thank you very much for the review!

PR #7216, opened 2023-09-05
BUG: Make translatable Editor Effects titles in Segmentations

mhdiop:
Enables fixing of these issues:
* Slicer/SlicerLanguagePacks#12
* Slicer#6177

### DIFF
--- a/SuperBuild/External_JsonCpp.cmake
+++ b/SuperBuild/External_JsonCpp.cmake
@@ -75,7 +75,12 @@ if(NOT DEFINED ${proj}_DIR AND NOT Slicer_USE_SYSTEM_${proj})
     set(lib_prefix "lib")
     set(lib_ext "so")
   endif()
-  set(${proj}_LIBRARY ${${proj}_DIR}/src/lib_json/${CMAKE_CFG_INTDIR}/${lib_prefix}jsoncpp.${lib_ext})
+  if(DEFINED CMAKE_CONFIGURATION_TYPES)
+    set(lib_cfg_dir "$<CONFIG>")
+  else()
+    set(lib_cfg_dir ".")
+  endif()
+  set(${proj}_LIBRARY ${${proj}_DIR}/src/lib_json/${lib_cfg_dir}/${lib_prefix}jsoncpp.${lib_ext})
 
   #-----------------------------------------------------------------------------
   # Launcher setting specific to build tree


Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorDrawEffect.py
```diff
@@ -16,6 +17,7 @@ class SegmentEditorDrawEffect(AbstractScriptedSegmentEditorLabelEffect):
 
     def __init__(self, scriptedEffect):
         scriptedEffect.name = 'Draw'
```
lassoan, line 19: Add a comment that this should not be translated (to prevent accidental translation by other developers in the future):
```suggestion
        scriptedEffect.name = 'Draw'  # no tr (effect name must not be translated because modules find effects by this name)
```

Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorFillBetweenSlicesEffect.py
```diff
@@ -15,6 +17,7 @@ class SegmentEditorFillBetweenSlicesEffect(AbstractScriptedSegmentEditorAutoComp
     def __init__(self, scriptedEffect):
         AbstractScriptedSegmentEditorAutoCompleteEffect.__init__(self, scriptedEffect)
         scriptedEffect.name = 'Fill between slices'
```
lassoan, line 19: Add the no tr comment everywhere

Modules/Loadable/Segmentations/EditorEffects/qSlicerSegmentEditorAbstractEffect.cxx
```diff
@@ -142,6 +142,31 @@ void qSlicerSegmentEditorAbstractEffect::setName(QString name)
...
+    {
+    return this->m_Title;
+    }
+  else if (!this->m_Name.isEmpty())
+    {
+    return this->m_Name;
+    }
+  else
+    {
+    qCritical() << Q_FUNC_INFO << ": Empty effect title!";
```
lassoan, line 158: qWarning is probably sufficient for this

Modules/Loadable/Segmentations/EditorEffects/qSlicerSegmentEditorAbstractEffect.cxx
```diff
@@ -142,6 +142,31 @@ void qSlicerSegmentEditorAbstractEffect::setName(QString name)
...
+    return QString();
+    }
+}
+
+//-----------------------------------------------------------------------------
+void qSlicerSegmentEditorAbstractEffect::setTitle(QString title)
+{
+  Q_UNUSED(title);
+  qCritical() << Q_FUNC_INFO << ": Cannot set effect title by method, only in constructor!";
```
lassoan, line 167: If you don't want to make effect title editable then remove this method.

Modules/Loadable/Segmentations/EditorEffects/qSlicerSegmentEditorAbstractEffect.h
```diff
@@ -70,6 +70,11 @@ class Q_SLICER_SEGMENTATIONS_EFFECTS_EXPORT qSlicerSegmentEditorAbstractEffect :
   /// \sa name(), \sa setName()
   Q_PROPERTY(QString name READ name WRITE setName)
 
+  /// This property stores the title of the effect
+  /// Cannot be empty.
+  /// \sa title(), \sa setTitle()
+  Q_PROPERTY(QString title READ title WRITE setTitle)
```
lassoan, line 76:
```suggestion
  Q_PROPERTY(QString title READ title)
```

Modules/Loadable/Segmentations/EditorEffects/qSlicerSegmentEditorAbstractEffect.h
```diff
@@ -243,6 +248,12 @@ public slots:
   /// NOTE: name must be defined in constructor in C++ effects, this can only be used in python scripted ones
   virtual void setName(QString name);
 
+  /// Get title of effect. Returns the effect's name when the title is empty
+  virtual QString title()const;
+  /// Set the title of the effect
+  /// NOTE: title must be defined in constructor in C++ effects, this can only be used in python scripted ones
+  virtual void setTitle(QString title);
```
lassoan, line 255: Why Python and C++ effects behave differently? We should keep the behavior the same for both (either both required to set the title in the constructor or neither).

Modules/Loadable/Segmentations/EditorEffects/qSlicerSegmentEditorAbstractEffect.h
```diff
@@ -420,6 +431,9 @@ public slots:
   QString m_Name;
   bool m_Active{false};
 
+  /// Title of the effect
```
lassoan, line 434:
```suggestion
  /// Title of the effect. This is displayed on the user interface and translated to the current language.
```

Modules/Loadable/Segmentations/EditorEffects/qSlicerSegmentEditorAbstractEffect.h
```diff
@@ -420,6 +431,9 @@ public slots:
   QString m_Name;
```
lassoan, line 431: Clarify the description of this property:

/// Name of the effect. This name is used by various modules for getting access to an effect. This string is not displayed on the user interface and not translated.

lassoan (commented):
This is very good, just added a few comments, mostly should be easy to fix.
The only topic that may need some more discussion is how to initialize the title (in the constructor).

Modules/Loadable/Segmentations/EditorEffects/qSlicerSegmentEditorAbstractEffect.h
mhdiop: > Why Python and C++ effects behave differently?

I made this to be coherent with the current definition of the class in which `name` and `perSegment` properties are meant, in C++ effects, to only be initialized from the constructor, not through a setter.

The different behaviour is justified by the fact that the `setTitle` method is required for scripted effects to make the `title` property writable (even in the constructor). What is not the case for C++ effects in which initialization may be done in constructors without using the setter.

I will however do what you suggest and keep the same setter's behaviour for all effects types. It should not be a problem since that property is not as critical as the `name` which is used in other contexts.

Modules/Loadable/Segmentations/EditorEffects/qSlicerSegmentEditorAbstractEffect.cxx
mhdiop: This method is required by scripted effects so that their `title` property is writable.

Modules/Loadable/Segmentations/EditorEffects/qSlicerSegmentEditorAbstractEffect.h
pieper: Yes, we want to keep the C++ and Python APIs as similar as possible to help people more easily go back and forth between languages.

mhdiop: Thank you for your feedback, that's what is done in the last submitted commit.
`setTitle` is now only defined in `qSlicerSegmentEditorAbstractEffect` base class and inherited by all other derived effect classes.

lassoan:
Parent folder name is added to the context name is if the folder contains `__init__.py`. Since the folder structure is different in the source tree (where strings are extracted from) and the install tree (where strings are replaced at runtime), the context strings are different and therefore translation fails.

To resolve this, I'll reorganize the folder structure in the source tree to match the install tree. Do not merge until then.

PR #7219, opened 2023-09-07
ENH: Remove fixed size buttons in view controller for better restyling

jamesobutler:
If restyling ToolButtons as a whole through QSS with a fixed vertical size set, but a variable width allowed, the fixed size specification in qMRMLViewControllerBar was still forcing these buttons to be 15px wide.

| `main` | This PR |
|--------|----------|
|image|image|

Now these buttons are sized based on their contents which in this case is their icon. This results in a slightly different spacing and sizing as previously the FixedSize call was actually cutting off part of the ToolButton's normal margin/padding. However I think the end result is near identical and functionally acts the same. This change ultimately allows for better ways of restyling QToolButtons through QSS.

---------------------------------------
I could also issue a separate change to allow these QToolButton's in the slice view controller to be their default size based on their normal 24px icons (remove any fixed size or iconSize specifications for these QToolButtons). This will result in the buttons being the same size as those in the Slice controller floating pop-up widget which I think is more consistent and the larger size buttons make the accessibility better to avoid mis-clicks and speeds up the time it takes to click successfully. It would look like the following:

| `main` | Default size for 24px icon |
|--------|----------|
image|image|

### DIFF
--- a/Libs/MRML/Widgets/qMRMLPlotViewControllerWidget.cxx
+++ b/Libs/MRML/Widgets/qMRMLPlotViewControllerWidget.cxx
@@ -113,7 +113,6 @@ void qMRMLPlotViewControllerWidgetPrivate::init()
   this->FitToWindowToolButton = new QToolButton(q);
   this->FitToWindowToolButton->setAutoRaise(true);
   this->FitToWindowToolButton->setDefaultAction(this->actionFit_to_window);
-  this->FitToWindowToolButton->setIconSize(QSize(12, 12));
   this->BarLayout->insertWidget(2, this->FitToWindowToolButton);
 }
 
--- a/Libs/MRML/Widgets/qMRMLPlotViewControllerWidget.cxx
+++ b/Libs/MRML/Widgets/qMRMLPlotViewControllerWidget.cxx
@@ -113,7 +113,7 @@ void qMRMLPlotViewControllerWidgetPrivate::init()
   this->FitToWindowToolButton = new QToolButton(q);
   this->FitToWindowToolButton->setAutoRaise(true);
   this->FitToWindowToolButton->setDefaultAction(this->actionFit_to_window);
-  this->FitToWindowToolButton->setFixedSize(15, 15);
+  this->FitToWindowToolButton->setIconSize(QSize(12, 12));
   this->BarLayout->insertWidget(2, this->FitToWindowToolButton);
 }
 
--- a/Libs/MRML/Widgets/qMRMLSliceControllerWidget.cxx
+++ b/Libs/MRML/Widgets/qMRMLSliceControllerWidget.cxx
@@ -117,11 +117,8 @@ qMRMLSliceControllerWidgetPrivate::~qMRMLSliceControllerWidgetPrivate() = defaul
 //---------------------------------------------------------------------------
 void qMRMLSliceControllerWidgetPrivate::setColor(QColor barColor)
 {
-  //this->SliceOffsetSlider->spinBox()->setAutoFillBackground(true);
   this->Superclass::setColor(barColor);
-  QPalette spinBoxPalette( this->SliceOffsetSlider->spinBox()->palette());
-  spinBoxPalette.setColor(QPalette::Base, barColor.lighter(130));
-  this->SliceOffsetSlider->spinBox()->setPalette(spinBoxPalette);
+  this->SliceOffsetSlider->spinBox()->setStyleSheet("background-color: transparent");
 }
 
 //---------------------------------------------------------------------------
--- a/Libs/MRML/Widgets/qMRMLSliceControllerWidget.cxx
+++ b/Libs/MRML/Widgets/qMRMLSliceControllerWidget.cxx
@@ -390,7 +390,6 @@ void qMRMLSliceControllerWidgetPrivate::init()
   //this->FitToWindowToolButton->setIcon(fitToWindowIcon);
   this->FitToWindowToolButton->setAutoRaise(true);
   this->FitToWindowToolButton->setDefaultAction(this->actionFit_to_window);
-  this->FitToWindowToolButton->setIconSize(QSize(12, 12));
   this->BarLayout->insertWidget(2, this->FitToWindowToolButton);
 
   this->SliderSpacer = new ctkDynamicSpacer(q);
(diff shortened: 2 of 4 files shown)


pieper:
We like to keep as much space as possible for content in those views, especially for small screen scenarios, but these do look like small changes that may be worth it for customizability.

I agree on the accessibility issue, since these small icons can be hard to hit.  Making everything more consistently respond to styling makes sense too.  Some apps define "cozy" mode vs a more spread out mode.  Maybe we could have defaults for that kind of things to keep the compact layout available.

lassoan:
Thanks for working on this. I've tried this and the control bar does not take up significantly more space, so it is good. The thicker control bar also make it a bit easier to click the pushpin icon, so that's good, too.

The only things that don't not look nice now:
- there doesn't seem to be a margin around the view label
- there is a relatively wide gap between the center and maximize buttons (it would not be a big issue, but the spacing between buttons is now very uneven - maybe due to the same issue of wrong margin size for the label)

image

Previously it looked like a nice square labels and uniform (very tight) spacing:

image

jamesobutler:
> We like to keep as much space as possible for content in those views, especially for small screen scenarios

@pieper Do you know how small of a screen resolution Slicer should support? I could set my monitor to it for testing. 
Looking at https://gs.statcounter.com/screen-resolution-stats/desktop/worldwide it appears the most common size is 1920x1080. That means the current 15px tall slice controllers in the default Four-Up layout would consume 30px total in the vertical dimension which is approximately 3% of the vertical dimension. If the buttons in the slice controller are their default size based on icon as they are within the slice controller pop-up area they would be approximately 30px tall or consume 60px total in the vertical dimension which is approximately 6% of the vertical dimension.

@lassoan I will take a look to give the slice view label some left/right padding. I assume it is to be square when the slice view label is only 1 letter, but allowed to be wider when the label is multiple letters?

pieper:
I don't think there's any sure-fire way to know what screensizes we need to target, it more a general sense of wanting to be mindful of space.  Removing hard-coded pixel sizes is definitely a step in the right direction and ideally it should allow us to go both directions, making things more compact for small screens or less compact for visibility and ease of clicking, without making anything ugly in the process.

I'm fine with this change and can approve unless others want to weigh in more.

jamesobutler:
I can aim to make sure things work well for the most common desktop screen resolutions from https://gs.statcounter.com/screen-resolution-stats/desktop/worldwide as a best effort. This should help make the slice controllers buttons both easy to click without having to slow down with to click with precision and while also being mindful of space.

I'll continue my work regarding the slice view label. I will also post a second commit with the default size for these buttons (30px tall slice controller instead of 15px). With this second commit folks can try it out locally.
image

lassoan:
Note that that the 15px in Qt is device independent pixels. It may be mapped to 15 pixels on screen, but it may be mapped to 60 pixels - depending on high-dpi settings, screen size, display scaling, font size, etc. We need to test on several different systems (if not before merging then shortly after).

> @lassoan I will take a look to give the slice view label some left/right padding. I assume it is to be square when the slice view label is only 1 letter, but allowed to be wider when the label is multiple letters?

Yes, approximately square-shaped label for single-character view label would look nice.

jamesobutler:
https://github.com/Slicer/Slicer/blob/6fc6a2abe964263f7cdef4c57fd1675a561256f4/Libs/MRML/Widgets/qMRMLViewControllerBar.cxx#L119-L123
@lassoan This is the code that has been making the slice view label approximately square. I have restored this code which I had previously removed. 

`ENH: Remove fixed size buttons in view controller for better restyling` looks like this:
image

However if I customize the size of the QToolButtons so that the Slice Controller vertical height is increased the label will once again become a tall rectangle as this current code is based on font.
- I could use QSS to style the slice view label directly to be have a similar width to my QSS styling of QToolButton height. 
- We could also consider moving the slice view label to the left of the push pin button so that isn't so jarring that the width of the item is different from the QToolButtons
image

--------------------------------------------------------
`ENH: Slice controller toolbuttons sized larger to icon size` currently looks like this:
image

--------------------------------------------------------
For additional reference, my Slicer custom app currently has the Slice controller styled like (still a WIP as I need to fix up some things related to icon size). Sizing decisions are primarily based on accessibility user interface standards that my group's UX folks are following to set the guidelines that they provide to us developers.
image

lassoan:
The first image in the previous is close, but still not quite right. The others look much worse than the current controller bar (due to the tight cut around the letter and variable spacing between the buttons and label).

Maybe it is a good time to re-think the slider orientation and the whole view controller. Having to push the pushpin button is unusal and not very convenient. Making the button bigger would not improve things much. We should consider showing a vertical toolbar for scrolling the slice (all other software that I've seen uses a vertical slider for slice selection) and moving the slice view controller menu content to context menu items and/or to interactive corner annotations.

jamesobutler:
A vertical slice offset slider was added in https://github.com/Slicer/Slicer/commit/2ed9451921d937f2a73d846bd1bf19b956478492 if you’re interested in changing the Slicer standard.

lassoan:
Switching to vertical slider in Slicer core must be coupled with reworking the slice view controller, because above/below the vertical slider there is no place for wide widgets, such as text labels (slice position, orientation, view label, etc.), so they should be in corner annotations and/or context menu items. That's why we only use the vertical slider in custom applications.

jamesobutler:
I have tried some more things out and not having the lighter color background behind the slice view label (and also the slice offset) definitely helps things look more consistent.

`main`:
400x95 762x50
image

`ENH: Remove different color background for slice view label and offset`:
image

`ENH: Remove fixed size buttons in view controller for better restyling`:
The spacing between buttons and labels looks good here and helps with the styling in my Slicer custom app by removing the fixed size objects
image

`ENH: Slice controller toolbuttons sized larger to icon size`:
The spacing between buttons and labels still looks good here with the sizes that match the toolbutton sizes 
image
image

jcfr:
| Without explicit iconsize | With explicit iconsize |
|--|--|
| image | image |
| | Preview after reverting only [ENH: Slice controller toolbuttons sized larger to icon size](https://github.com/Slicer/Slicer/pull/7219/commits/c3ee05962f7b729043df251918294db55f5c689b) |

jcfr (approved):
I suggest we move forward and revisit the reverting of the last commit based on community feedback.

jamesobutler:
During the Tuesday hangout when this PR was discussed we did get some initial feedback from @jacob-moore22 that from a new user perspective he did have initial trouble finding the center slice button which he realized was important for regular use of slice view manipulation. He also mentioned that the increased slice controller size was appreciated to help with visual accessibility for him.

PR #7226, opened 2023-09-11
BUG: Fix closed surface disappearing when segment is removed

### LINKED ISSUE #7220: Deleting the first segment removes closed surface representation
## Summary

When removing the first segment in a segmentation and show 3D was enabled, after the segment was removed, 3D is no longer displayed.

## Steps to reproduce

- Load a volume
- Create a segmentation with 3 segments
- Paint something into each segment
- Click Show 3D button to see the segmentation in 3D
- Remove the first segment

### Expected behavior

The first segment should be removed but all the other segments should be still visible in 3D.

### Actual behavior

All the segments disappear. Show 3D button is untoggled.

## Environment
- Slicer version: Slicer-5.4
- Operating system: Windows and Mac

Sunderlandkyl:
When a segment that is in the same layer as other segments is removed from a vtkSegmentation, it causes the shared labelmap to be modified. This triggered a SourceRepresentationModified on the remaining segments, which invalidated non-source representations. Fixed by calling SetSourceRepresentationModifiedEnabled(false) before editing the labelmap, and restoring the original value after.

Fixes #7220

### DIFF
--- a/Libs/vtkSegmentationCore/vtkSegmentation.cxx
+++ b/Libs/vtkSegmentationCore/vtkSegmentation.cxx
@@ -1552,34 +1552,49 @@ void vtkSegmentation::SeparateSegmentLabelmap(std::string segmentId)
 
   vtkOrientedImageData* labelmap = vtkOrientedImageData::SafeDownCast(
     segment->GetRepresentation(vtkSegmentationConverter::GetBinaryLabelmapRepresentationName()));
-  if (labelmap)
+  if (!labelmap)
     {
-    vtkNew<vtkImageThreshold> threshold;
-    threshold->SetInputData(labelmap);
-    threshold->ThresholdBetween(segment->GetLabelValue(), segment->GetLabelValue());
-    threshold->SetOutValue(0);
-    threshold->SetInValue(DEFAULT_LABEL_VALUE);
-    threshold->Update();
-
-    vtkSmartPointer<vtkOrientedImageData> tempImage = vtkSmartPointer<vtkOrientedImageData>::New();
-    tempImage->ShallowCopy(threshold->GetOutput());
-    tempImage->CopyDirections(labelmap);
-
-    segment->AddRepresentation(vtkSegmentationConverter::GetBinaryLabelmapRepresentationName(), tempImage);
-
-    vtkNew<vtkImageThreshold> thresholdErase;
-    thresholdErase->SetInputData(labelmap);
-    thresholdErase->ThresholdBetween(segment->GetLabelValue(), segment->GetLabelValue());
-    thresholdErase->SetInValue(0);
-    thresholdErase->ReplaceOutOff();
-    thresholdErase->Update();
-    labelmap->ShallowCopy(thresholdErase->GetOutput());
+    return;
     }
+
+  vtkNew<vtkImageThreshold> threshold;
+  threshold->SetInputData(labelmap);
+  threshold->ThresholdBetween(segment->GetLabelValue(), segment->GetLabelValue());
+  threshold->SetOutValue(0);
+  threshold->SetInValue(DEFAULT_LABEL_VALUE);
+  threshold->Update();
+
+  vtkSmartPointer<vtkOrientedImageData> tempImage = vtkSmartPointer<vtkOrientedImageData>::New();
+  tempImage->ShallowCopy(threshold->GetOutput());
+  tempImage->CopyDirections(labelmap);
+
+  segment->AddRepresentation(vtkSegmentationConverter::GetBinaryLabelmapRepresentationName(), tempImage);
+
+  vtkNew<vtkImageThreshold> thresholdErase;
+  thresholdErase->SetInputData(labelmap);
+  thresholdErase->ThresholdBetween(segment->GetLabelValue(), segment->GetLabelValue());
+  thresholdErase->SetInValue(0);
+  thresholdErase->ReplaceOutOff();
+  thresholdErase->Update();
+
+  // Although the source labelmap will be modified, vtkSegmentation::onSourceRepresentationModified would unnecessarily invalidate the non-source
+  // representations (because separating a labelmap into a new layer does not change the segment's shape).
+  // This prevents vtkSegmentation::onSourceRepresentationModified from being called. We will later invoke events to let observers know about the change.
+  bool wasSourceRepresentationModifiedEnabled = this->SetSourceRepresentationModifiedEnabled(false);
+  labelmap->ShallowCopy(thresholdErase->GetOutput());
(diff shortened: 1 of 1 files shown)


lassoan (commented):
Thank you for the quick fix. Are there any usages of `SeparateSegmentLabelmap` that would need to be similarly fixed up? Even better, could we call `SetSourceRepresentationModifiedEnabled` inside `SeparateSegmentLabelmap` (or that could suppress update events and cause issues)?

Sunderlandkyl:
I just tested it, and there is another use that causes the same issue. Trying to paint over another segment with "Allow overlap" will cause the 3d surface to disappear. It makes sense to put SetSourceRepresentationModifiedEnabled inside SeparateSegmentLabelmap.

jcfr:
I did some testing to identify what introduced the regression:

| revision | hash | status |
|--|--|--|
| 31503 | 7f5f83e | :white_check_mark: |
| 31739 | 9ae54bd | :white_check_mark: |
| 31800 | 7195f78 | :white_check_mark: |
| 31807 | d7c9a0d | :white_check_mark: |
| 31810 | 1bd1dcf | :white_check_mark: |
| 31813 | d12bcab | :x: |
| 31843 | 5a5de07 | :x: |

The regression was likely introduced in d12bcabc2b777512ae310c2117fd2195ab0cccfb (`ENH: Update VTK from 9.1.20220125 to 9.2.20230607`) integrated through:
* https://github.com/Slicer/Slicer/pull/7010

image

Sunderlandkyl:
Thanks @jcfr. It looks like ShallowCopy didn't trigger a ModifiedEvent with the previous version of VTK.

@lassoan I've moved SetSourceRepresentationModifiedEnabled inside SeparateSegmentLabelmap.
I've also removed InvokeEvent "SourceRepresentationModified" and "ContainedRepresentationNamesModified" as I don't think they should be triggered manually by the function.

Libs/vtkSegmentationCore/vtkSegmentation.cxx
```diff
@@ -1573,13 +1573,24 @@ void vtkSegmentation::SeparateSegmentLabelmap(std::string segmentId)
...
     }
   segment->SetLabelValue(DEFAULT_LABEL_VALUE);
 
+  std::string sourceRepresentationName = this->GetSourceRepresentationName();
+  if (strcmp(sourceRepresentationName.c_str(), vtkSegmentationConverter::GetBinaryLabelmapRepresentationName()) == 0)
+    {
+    // We were blocking onSourceRepresentationModified, however the source labelmap was modified so we need to manually invoke the event.
+    this->InvokeEvent(vtkSegmentation::SourceRepresentationModified, labelmap);
+    }
+  this->InvokeEvent(vtkSegmentation::RepresentationModified, labelmap);
```
lassoan, line 1591: Could you please check if these 3 events (vtkSegmentation::SourceRepresentationModified, vtkSegmentation::RepresentationModified, this->Modified()) do not result in performing any expensive operation (e.g., regeneration of closed surface representation) multiple times? (most probably not, just to make sure)

Libs/vtkSegmentationCore/vtkSegmentation.cxx
```diff
@@ -1573,13 +1573,24 @@ void vtkSegmentation::SeparateSegmentLabelmap(std::string segmentId)
     thresholdErase->SetInValue(0);
     thresholdErase->ReplaceOutOff();
     thresholdErase->Update();
+
+    // Although the source labelmap will be modified, onSourceRepresentationModified would invalidate the non-source representations.
```
lassoan, line 1577: vtkMRMLSegmentationNode::OnSourceRepresentationModified() or vtkSegmentation::OnSourceRepresentationModified()?

lassoan, line 1577:
```suggestion
    // Although the source labelmap will be modified, onSourceRepresentationModified would unnecessarily invalidate the non-source representations
    // (because separating a labelmap into a new layer does not change the segment's shape).
```

Libs/vtkSegmentationCore/vtkSegmentation.cxx
```diff
@@ -1573,13 +1573,24 @@ void vtkSegmentation::SeparateSegmentLabelmap(std::string segmentId)
...
+    // This prevents onSourceRepresentationModified from being called.
```
lassoan, line 1578:
```suggestion
    // This prevents onSourceRepresentationModified from being called. We will later invoke events to let observers know about the change. 
```

Libs/vtkSegmentationCore/vtkSegmentation.cxx
```diff
@@ -1573,13 +1573,24 @@ void vtkSegmentation::SeparateSegmentLabelmap(std::string segmentId)
     thresholdErase->SetInValue(0);
     thresholdErase->ReplaceOutOff();
     thresholdErase->Update();
+
```
lassoan, line 1576: I've just noticed above in line 555: `if (labelmap)...` - why don't we return immediately if `labelmap == nullptr` ?

lassoan (commented):
Thank you, this looks good! I've just added a few (mostly trivial) comments and suggestions.

Libs/vtkSegmentationCore/vtkSegmentation.cxx
Sunderlandkyl: Summary of event observations:
- SourceRepresentationModified:
  - updateWidgetFromMRML in qSlicerSegmentationsModuleWidget
  - updateSliceRotateWarningButtonVisibility in qMRMLSegmentEditorWidget.
  - populateRepresentationsList in qMRMLSegmentationRepresentationsListView.
  - Segmentation history
  - Update storage node supported write file types.
  - Propagate event through segmentation node.

- RepresentationModified:
  - Displayable managers.
  - Update StorableModifiedTime.
  - Propagate event through segmentation node..

- ModifiedEvent:
  - Not observed anywhere

Libs/vtkSegmentationCore/vtkSegmentation.cxx
Sunderlandkyl: vtkSegmentation::OnSourceRepresentationModified()

Sunderlandkyl:
Thanks @lassoan, changes made!

lassoan (approved):
Thank you, it all looks good.

jcfr:
> ContainedRepresentationNamesModified is still removed

For reference, this was first introduced in fcb40b0738c65ab418731a1a34292f3c85b2133a (`ENH: Add support for shared binary labelmap segmentations`)

PR #7227, opened 2023-09-12
BUG: Update VTK to remove unnecessary messages in vtkTransformPolyDataFilter

jcfr:
`vtkTransformPolyDataFilter` is often used in pipelines where it is completely valid to have empty input.
For example, when slicing through a mesh with a plane, the output may be empty if the plane does not happen to intersect any part of the mesh. Logging an error pollutes the error log and if performed too frequently then it may slow down an application.

List of VTK changes:

```
$ git shortlog b261e0c1b..b142c0d08 --no-merges
Andras Lasso (1):
      [Backport] Remove unnecessary "No input data" error message in vtkTransformPolyDataFilter
```


jcfr:
Local testing confirmed that the issue is resolved :white_check_mark:

lassoan:
Thank you. I've labelled this as need to be backported to Slicer-5.4.1.

PR #7231, opened 2023-09-18
COMP: Use gold linker by default on Linux

achataigner:
To avoid this kind of ld linker warnings /path/to/libQt5WebEngineCore.so.5.15.2: .dynsym local symbol at index 3 (>= sh_info of 3).
To reduce linking time since gold is faster than ld.

### DIFF
--- a/Docs/developer_guide/build_instructions/linux.md
+++ b/Docs/developer_guide/build_instructions/linux.md
@@ -224,31 +224,6 @@ Instead of `cmake`, one can use `ccmake`, which provides a text-based interface,
 
 :::
 
-:::{admonition} Tip -- Speed up 3D Slicer build with `ccache`
-
-`ccache` is a compiler cache that can speed up subsequent builds of 3D Slicer. This can be useful if 3D Slicer is built often and there are no large divergences between subsequent builds. This requires `ccache` installed on the system (e.g., `sudo apt install ccache`).
-
-The first time `ccache` is used, the compilation time can marginally increase as it includes the first caching. After the first build, subsequent build times will decrease significantly.
-
-`ccache` is not detected as a valid compiler by the 3D Slicer building process. You can generate local symbolic links to disguise the use of `ccache` as valid compilers:
-
-```console
-ln -s /usr/bin/ccache ~/.local/bin/c++
-ln -s /usr/bin/ccache ~/.local/bin/cc
-```
-
-Then, the Slicer build can be configured to use these compilers:
-
-```console
-cmake \
-  -DCMAKE_BUILD_TYPE:STRING=Release \
-  -DCMAKE_CXX_COMPILER:STRING=$HOME/.local/bin/c++ \
-  -DCMAKE_C_COMPILER:STRING=$HOME/.local/bin/cc \
-  ../Slicer
-```
-
-:::
-
 ## Build Slicer
 
 Once the Slicer build project files have been generated, the Slicer project can
@@ -273,6 +248,12 @@ Increasing the number of parallel builds generally increases the memory required
 :::{admonition} Tip -- Error detection during parallel build
 
 Using parallel builds makes finding compilation errors difficult due to the fact that all parallel build processes use the same screen output, as opposed to sequential builds, where the compilation process will stop at the error. A common technique to have parallel builds and easily find errors is to launch a parallel build followed by a sequential build. For the parallel build, it is advised to run `make -j<N> -k` to have the parallel build keep going as far as possible before doing the sequential build with `make`.
+
+:::
+
+:::{admonition} Tip -- Speed up build
+
+Build and rebuild can be accelerated thanks to tips which are given in the [advanced section](#advanced).
 
 :::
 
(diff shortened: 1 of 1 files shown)


jcfr:
Thanks for the patch :pray:

Instead of explicitly setting the flag, I instead suggest the following:
* update linux prerequisites to document that gold linker should be installed
* indicate how the `update-alternatives` command should be used to set `gold` as the main linker

This would ensure the same linker is also used across all external projects.

achataigner:
> Thanks for the patch 🙏
> 
> Instead of explicitly setting the flag, I instead suggest the following:
> 
> * update linux prerequisites to document that gold linker should be installed
> * indicate how the `update-alternatives` command should be used to set `gold` as the main linker
> 
> This would ensure the same linker is also used across all external projects.

Here is what I did, I did not find how to do it on archlinux ?

achataigner:
Any feedback ?

pieper (approved):
Seems fine to me.  @RafaelPalomar do you have any comments?

RafaelPalomar:
@achataigner, thanks for suggesting this iimprovement. 👍🏻 Offering new options to developers is generally good, and these changes don't introduce much more documentation. In my view, there's a couple of points worth discussing:

  - @achataigner, is this offering a significant performance increase? (i.e., we introduced instructions for `ccache` based on the potential performance improvements https://discourse.slicer.org/t/speeding-up-slicer-builds-with-ccache/20497). 
  
  - There are some yellow lights 🟡 about the use of gold in the Gentoo wiki.
  > :warning:gold is [less active](https://www.phoronix.com/scan.php?page=news_item&px=GNU-Gold-Stagnate-F31) than it once was, e.g. gold's commit [history](https://sourceware.org/git/?p=binutils-gdb.git;a=history;f=gold;hb=HEAD) vs bfd's commit [history](https://sourceware.org/git/?p=binutils-gdb.git;a=history;f=ld;hb=HEAD). Users seeking an alternative linker may be interested in LLVM's [lld](https://wiki.gentoo.org/wiki/Clang).


- @jcfr, wouldn't it be better to instruct the user to pass the linker path to the Slicer configuration? What we are suggesting here is to change the configuration at the system level to just build Slicer differently. Passing a configuration flag is also a distribution-neutral move.

pieper:
> pass the linker path to the Slicer build configuration

I like this idea 👍

achataigner:
I agree with Rafael and Steve about passing the linker path to slicer build config.

About performance:
- in my 3d slicer custom app: If I simply change the current git commit then the build is faster: 30 seconds using gold rather than 45 seconds using bfd. (As you know changing git commit causes some files to be compiled even if no c++ code is changed: https://discourse.slicer.org/t/build-when-git-commit-is-changed/29607/24)
- in 3d slicer: I will compare times of incremental build tomorrow

lassoan:
45 to 30 sec build time after a commit is probably not worth the complication for most developers, as you probably don't commit or update more than a few times a day. Recommending changing the linker may also mean extra efforts for Slicer maintainers (to help out developers who ran into issues due to messing with their linker), and we would make the build instructions longer and more complex. Anyway, I'm not against adding this note to the instructions, along with the `ccache` tip, but these performance tips should maybe put into a separate section (to make it easier for skip over it) and also describe what performance improvement they can provide and what if any extra cost/complication they can expect.

In general, I find linux build times are all right. On my 3-year-old computer a full Slicer build from scratch takes 30 minutes and a rebuild after changing a few files takes just a few seconds. I'm not sure there is much room for improvement here that would worth the trouble. I admit that my view may be a bit skewed, as I mostly use Windows, where the full build time is 2.5-3 hours and an application-level build without any file modification takes about 20 seconds. Compared to this, linux builds - even with default options - are already really fast.

achataigner:
Thank you Andras for your feedback.
Just let me know if you want me to do something.
Regards

lassoan:
> Just let me know if you want me to do something.

The two performance improvement options (ccache, gold) takes about one full screen. I'm sure that in the future other developers will suggest more performance optimization options that some developers may find useful. Skimming through all this non-essential information is extra time, forces developes to make decisions, and cause uncertainty (have I made the right decision? why it is so complicated to build this application at all?...).

What do you think about the followings?
- Move performance tips to a separate "advanced" section after the main build instructions (to the very end of the page or just before "Run Slicer")
- Add a short tip in the main instruction text that describes how long it takes typically to build Slicer and how much time can be saved by using the special instructions (and add a link to the special instructions section)

jcfr:
> Move performance tips to a separate "advanced" section
> Add a short tip in the main instruction ...

:+1: 

>  What we are suggesting here is to change the configuration at the system level to just build Slicer differently

Ditto. Using the gold linker addresses the following linker error for Slicer and any other projects linking against the corresponding libraries. Removing the extra noise in the output is sensible to ensure "real" warnings are easily identified.

`/path/to/libQt5WebEngineCore.so.5.15.2: .dynsym local symbol at index 3 (>= sh_info of 3)`

> to pass the linker path to the Slicer build configuration?

To supporting this, we would have to first make sure Slicer properly propagate `CMAKE_LINKER` for all CMake and non-CMake external projects (like we are already doing for `CMAKE_C_COMPILER` and `CMAKE_CXX_COMPILER`). 

To test this, a clean build would have to be done after setting `LD` env variable to `/dev/null` and explicitly passing `CMAKE_LINKER` to the top-level Slicer project.

This would ensure that both passing `CMAKE_LINKER` CMake option or setting `LD` env variable would work as expected.

achataigner:
Branch updated.
I did not write any number about performance improvements, just let me know if we really need to write this.

RafaelPalomar:
I have opened Slicer/Slicer#7258 where I'm following the suggestions of @jcfr to enable the linker flag propagation as alternative to change the linker globally. I suggest we proceed with Slicer/Slicer#7258 first and update this accordingly.

lassoan:
> did not write any number about performance improvements, just let me know if we really need to write this.

Yes, please write some numbers. Without that developers cannot make an informed decision if it is worth the trouble for them to install and configure additional tools. Lack of information may instill doubts and uncertainty about the build process.

@RafaelPalomar it would be nice if you could add more specific information about expected benefits of ccache, too. For example, typical build times, and mire explicit description of when ccache can speed up subsequent builds of Slicer (my understanding is that only if you build multiple Slicer branches on the same computer).

Thank you both for taking the time to share with the community your experience with Linux build tools.

RafaelPalomar:
Some updates about my efforts making the gold linker available in the Slicer build process:

1. Splitting the compiler and linker to be able to choose custom linkers is problematic. You  can read more about my findings in Slicer/Slicer#7258. I chose to close that up in favor of Slicer/Slicer#7294, as I think having those changes as documented experience can have its value for the future.
2. In Slicer/Slicer#7294, I have enabled a new CMake option (`USE_GOLD_LINKER`), to enable this functionality. @achataigner, could you give it a try?

If we move on with Slicer/Slicer#7294, I could add the documentation for the `USE_GOLD_LINKER` option in CMake.

jcfr:
Closing.

This PR was superseded by:
* https://github.com/Slicer/Slicer/pull/7258
* https://github.com/Slicer/Slicer/pull/7294

Issue #7233, opened 2023-09-18
Improve apperance of of slice view controller


lassoan:
The pull request that removed some styling from the slice view controllers (https://github.com/Slicer/Slicer/pull/7219) introduced some cosmetic/usability issues:
- white-on-yellow text in the yellow's slice controller
- blurry push-pin icon (the icon might have been low-resolution before, but it was not this big, so it was not that noticeable)

![image](https://github.com/Slicer/Slicer/assets/307929/e14d847f-c181-4dd4-8a03-6b11cdd15084)

jamesobutler:
> white-on-yellow text in the yellow's slice controller

Is this specifically Slicer Dark style? I have a memory of everything being dark text by I’m typically using Slicer light style.

lassoan:
Yes, the default dark style.

jamesobutler:
@lassoan I can't really tell that the PushPin looks any blurrier than the original PNG file when I view them side-by-side. Maybe the blurriness you are seeing is the same as what is inherently in the PNG file? 
- PushPinIn.png (32x32px) https://github.com/Slicer/Slicer/blob/c721e9c647b4f97243ede001674f4e0b7756b5b0/Libs/MRML/Widgets/Resources/Icons/PushPinIn.png
- PushPinOut.png (32x32px) https://github.com/Slicer/Slicer/blob/c721e9c647b4f97243ede001674f4e0b7756b5b0/Libs/MRML/Widgets/Resources/Icons/PushPinOut.png

I'm not sure what I could do if it the issue is inherently in the original source file. I could swap it out for the Material Symbol's [Push Pin SVG](https://fonts.google.com/icons?icon.platform=web&icon.set=Material+Symbols&selected=Material+Symbols+Outlined:push_pin:FILL@1;wght@400;GRAD@0;opsz@24&icon.query=push+pin).
![image](https://github.com/Slicer/Slicer/assets/15837524/4c35cc39-e148-4060-9dc1-19cf6a13f866)


Is it only the PushPin icons that look blurry? Or do any of the other icons in the Slice view controller look blurry to you?

- SlicesFitToWindow.png (19x19px) https://github.com/Slicer/Slicer/blob/c721e9c647b4f97243ede001674f4e0b7756b5b0/Libs/MRML/Widgets/Resources/Icons/SlicesFitToWindow.png
[...]

PR #7234, opened 2023-09-18
BUG: Fix difficult to see slice offset text in Slicer dark mode

jamesobutler:
This is work for #7233.

https://github.com/Slicer/Slicer/commit/8e70f763b0b351db00d8851871e68f11469d1a2e introduced this issue of light text being used for the slice offset text when in Slicer dark mode. Dark text was correctly being used in Slicer light mode. This change uses dark text regardless of current Slicer style.

|`main` | This PR |
|--------|---------|
|image|image|

### DIFF
--- a/Libs/MRML/Widgets/qMRMLSliceControllerWidget.cxx
+++ b/Libs/MRML/Widgets/qMRMLSliceControllerWidget.cxx
@@ -118,7 +118,6 @@ qMRMLSliceControllerWidgetPrivate::~qMRMLSliceControllerWidgetPrivate() = defaul
 void qMRMLSliceControllerWidgetPrivate::setColor(QColor barColor)
 {
   this->Superclass::setColor(barColor);
-  this->SliceOffsetSlider->spinBox()->setStyleSheet("background-color: transparent");
 }
 
 //---------------------------------------------------------------------------
@@ -409,6 +408,8 @@ void qMRMLSliceControllerWidgetPrivate::init()
     ctkDoubleSpinBox::DecimalsByShortcuts |
     ctkDoubleSpinBox::DecimalsByKey |
     ctkDoubleSpinBox::DecimalsAsMin );
+  // Slice controller background color is independent from the color palette, therefore the color of text and controls are hardcoded to black
+  this->SliceOffsetSlider->spinBox()->setStyleSheet("color: black; background-color: transparent;");
 
   //this->SliceOffsetSlider->spinBox()->setParent(this->PopupWidget);
   ctkDoubleSpinBox* spinBox = this->SliceOffsetSlider->spinBox();
--- a/Libs/MRML/Widgets/qMRMLViewControllerBar.cxx
+++ b/Libs/MRML/Widgets/qMRMLViewControllerBar.cxx
@@ -115,6 +115,8 @@ void qMRMLViewControllerBarPrivate::init()
   this->ViewLabel = new QLabel(q);
   this->ViewLabel->setObjectName("ViewLabel");
   this->ViewLabel->setAlignment(Qt::AlignHCenter | Qt::AlignCenter);
+  // Slice controller background color is independent from the color palette, therefore the color of text and controls are hardcoded to black
+  this->ViewLabel->setStyleSheet("color: black; background-color: transparent;");
 #if (QT_VERSION >= QT_VERSION_CHECK(5, 11, 0))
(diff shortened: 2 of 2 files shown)


lassoan (commented):
Thank you for the fix.

Why this hardcoding to black was not necessary before? Had the color been set using palette before? If stylesheet overwrites palette colors anyway then could we keep the palette-based coloring? We should not patch coloring issues using stylesheet when it can be fixed using the currently official coloring method (by using palettes). When we decide to stop using palette-based coloring then it's fine to remove palette-related code, but not before then.

jamesobutler:
The dark palette specifies for text to be white and the light palette specifies for text to be black which is what `main` is currently doing. Previously it was doing it based on the default light style palette which resulted in the text always being dark.

In this case it makes sense to define the text color to be something static and not changed dynamically. A Slicer custom app could define a Base (background) that is Blue and for Text to be green, but we would still want to define text on the slice controller differently because the background color of the slice controller is not changing with style updates. You could consider the Slicer controller decision to use black icons instead of white icons for the toolbuttons to be under the same concept. We’re not going to dynamically change those icon colors with style changes.

lassoan:
Thanks for the explanation, it clarifies some things, but I'm still not sure where the special behavior of the spinbox text color is coming from. Why do we need to hardcode the black color of the spinbox text but not the ViewLabel ("R")? Could we color the spinbox text the same way as it is done for the ViewLabel?

jamesobutler:
> Why do we need to hardcode the black color of the spinbox text but not the ViewLabel ("R")? Could we color the spinbox text the same way as it is done for the ViewLabel?

@lassoan Reviewing the code below shows that the Slice view label always has black text because it uses its own custom palette for the widget and does not use the application wide color palette. I could style the Spinbox with a custom palette or a custom stylesheet.

https://github.com/Slicer/Slicer/blob/31ba8942a031b907d4be2f2f5d81db717b347a1b/Libs/MRML/Widgets/qMRMLViewControllerBar.cxx#L195-L200

Libs/MRML/Widgets/qMRMLSliceControllerWidget.cxx
```diff
@@ -409,6 +408,7 @@ void qMRMLSliceControllerWidgetPrivate::init()
     ctkDoubleSpinBox::DecimalsByShortcuts |
     ctkDoubleSpinBox::DecimalsByKey |
     ctkDoubleSpinBox::DecimalsAsMin );
+  this->SliceOffsetSlider->spinBox()->setStyleSheet("color: black; background-color: transparent;");  // Slicer light/dark mode does not manipulate Slice controller colors
```
lassoan, line 411:
```suggestion
  // Slice controller background color is independent from the color palette, therefore the color of text and controls are hardcoded to black
  this->SliceOffsetSlider->spinBox()->setStyleSheet("color: black; background-color: transparent;");
```

Libs/MRML/Widgets/qMRMLViewControllerBar.cxx
```diff
@@ -115,6 +115,7 @@ void qMRMLViewControllerBarPrivate::init()
   this->ViewLabel = new QLabel(q);
   this->ViewLabel->setObjectName("ViewLabel");
   this->ViewLabel->setAlignment(Qt::AlignHCenter | Qt::AlignCenter);
+  this->ViewLabel->setStyleSheet("color: black; background-color: transparent;");  // Slicer light/dark mode does not manipulate Slice controller colors
```
lassoan, line 118:
```suggestion
  // Slice controller background color is independent from the color palette, therefore the color of text and controls are hardcoded to black
  this->ViewLabel->setStyleSheet("color: black; background-color: transparent;");
```

lassoan (commented):
OK, the implementation is now consistent and the hardcoded color is explained. Please have a look at the suggested comment changes and if it looks good then we can merge. Thank you!

Libs/MRML/Widgets/qMRMLSliceControllerWidget.cxx
jamesobutler: Thanks for the updated comment suggestion. I have proceeded with this version!

jamesobutler:
Unclear why the build step is failing do to lost connection to server?
image

lassoan (approved):
Thank you, it all looks good.

lassoan:
I'll override the CI build requirement because CI does not seem to work now.

PR #7237, opened 2023-09-21
BUG: Use log filename to get app version and revision

nhjohnston:
Introduced in https://github.com/Slicer/Slicer/pull/7225#issuecomment-1729384982

This PR uses the log filename to get app version and revision number to report in the error report dialog:
image

`ENH: Report custom application revision in log filename` uses the custom-app revision number rather than the slicer revision number. This won't change anything from the slicer side, but could be beneficial for custom applications. (needs to be tested)

### DIFF
--- a/Base/QTGUI/qSlicerApplication.cxx
+++ b/Base/QTGUI/qSlicerApplication.cxx
@@ -1021,7 +1021,7 @@ void qSlicerApplication::setupFileLogging()
     .arg(tempDir)
     .arg(this->applicationName())
     .arg(qSlicerApplication::application()->applicationVersion())
-    .arg(this->revision())
+    .arg(qSlicerApplication::application()->mainApplicationRevision())
     .arg(QDateTime::currentDateTime().toString("yyyyMMdd_hhmmss"))
     .arg(QRandomGenerator::global()->generate() % 1000, 3, 10, QLatin1Char('0'));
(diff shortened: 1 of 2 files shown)


Base/QTGUI/qSlicerApplication.cxx
```diff
@@ -1021,7 +1021,7 @@ void qSlicerApplication::setupFileLogging()
     .arg(tempDir)
     .arg(this->applicationName())
     .arg(qSlicerApplication::application()->applicationVersion())
-    .arg(this->revision())
+    .arg(qSlicerApplication::application()->revision())
```
jamesobutler, line 1024: To support custom apps this should be the `mainApplicationRevision`. For regular 3D Slicer `revision` and `mainApplicationRevision` are equivalent, but for Slicer custom apps this will make sure the log file name and corresponding info in the table in the custom app's revision rather than the underlying Slicer version revision.
```suggestion
    .arg(qSlicerApplication::application()->mainApplicationRevision())
```

jamesobutler (approved):
Thanks for the bug fix and the improvement for the custom app revision being used for the log file name.

jcfr:
For reference, the before/after these changes is illustrated below:

| Before (Slicer 5.4.0) | After (Slicer 5.6.0) |
|--|--|
| image | image |

PR #7248, opened 2023-09-26
BUG: Fix "Show 3D" button not updating during undo/redo

### LINKED ISSUE #7246: Segmentation undo/redo does not update "Show 3D" button state
## Summary

Undo/redo in Segment Editor may add/remove closed surface representation, but the "Show 3D" button state (which should be pushed if and only if the segmentation contains closed surface representation) is not updated accordingly.

## Steps to reproduce

- Create segmentation, add segment -> paint in slide -> paint in slide -> enable Show 3D -> undo
- This makes the segmentation disappear from the 3D view but "Show 3D" button is still enabled (it should be disabled)

## Environment
- Slicer version: Slicer-5.5.0-2023-09-25
- Operating system: Windows

Sunderlandkyl:
During undo/redo, representations could be added or removed from the segment, however the "Show 3D" would not be updated. Fixed by invoking ContainedRepresentationNamesModified on the segmentation if the current and restored representation lists are different.

Fix #7246

### DIFF
--- a/Libs/vtkSegmentationCore/vtkSegmentationHistory.cxx
+++ b/Libs/vtkSegmentationCore/vtkSegmentationHistory.cxx
@@ -220,6 +220,8 @@ bool vtkSegmentationHistory::RestoreState(unsigned int stateIndex)
 {
   this->RestoreStateInProgress = true;
 
+  bool containedRepresentationNamesModified = false;
+
   SegmentationState restoredState = this->SegmentationStates[stateIndex];
 
   std::set<std::string> segmentIDsToKeep;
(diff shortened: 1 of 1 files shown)


lassoan (commented):
Thanks a lot for the quick fix! Why the missing event had to be invoked from vtkSegmentationHistory? The event should have been invoked automatically by the segmentation object.

Sunderlandkyl:
The RestoreState function doesn't invoke any of the methods in vtkSegmentation that would invoke ContainedRepresentationNamesModified. The RestoreState function removes representations on each segment individually.

Another solution that would probably work would be to call RemoveRepresentation on the vtkSegmentation object instead of on each of the segments.

lassoan:
OK. I think the implementation is good as is after all. We have similar APIs in VTK when manipulating objects at low level we need to invoke modification events manually.

lassoan (approved):

PR #7250, opened 2023-09-27
Fix crashes in subject hierarchy and DICOM reading

lassoan:
This pull request contains fixes for two independent problems that caused Slicer crashes.

### DIFF
--- a/Libs/vtkITK/vtkITKArchetypeImageSeriesReader.cxx
+++ b/Libs/vtkITK/vtkITKArchetypeImageSeriesReader.cxx
@@ -325,6 +325,14 @@ itk::ImageIOBase::Pointer vtkITKArchetypeImageSeriesReader::GetImageIO(const cha
     this->SetErrorCode(vtkErrorCode::FileFormatError);
     return nullptr;
     }
+  catch (const std::string& err)
+    {
+    // GDCM can throw std::string exception if some verification fail
+    vtkDebugMacro("vtkITKArchetypeImageSeriesReader::ExecuteInformation: Cannot open " << fileNameCollapsed.c_str() << ". "
+      << "Error: " << err);
+    this->SetErrorCode(vtkErrorCode::FileFormatError);
+    return nullptr;
+    }
 
   return imageIO;
 }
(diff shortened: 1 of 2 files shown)


jcfr (approved):
Good catch :+1: 

Nitpick:

I suggest to move the indentation changes into a `STYLE` commit.

Using `https://github.com/Slicer/Slicer/pull/7250/commits/dcf4b7d144584be099d95885e3d9368f9db1dbdc?w=1` (not the use of `?w=1`) helped me confirm the change as smaller than anticipated.

lassoan:
> I suggest to move the indentation changes into a `STYLE` commit.

The indentation had to be changed because the added `try {} catch {}` block. If I moved the indentation fix to a separate commit then my commit wouldn't pass style checks (if they were enabled for C++) and would add noise to the commit history.

It is great that you've found this `?w=1` option because that makes review easier without forcing whitespace changes into a separate commit.

jcfr:
> The indentation had to be changed because the added `try {} catch {}` block

Makes sense. 

> re: `w=1`

And reading https://github.blog/changelog/2021-10-14-hiding-whitespace-is-now-remembered-for-each-pull-request/

reminded me it was now available as part of the GitHub UI

image

Libs/vtkITK/vtkITKArchetypeImageSeriesReader.cxx
```diff
@@ -777,189 +790,210 @@ int vtkITKArchetypeImageSeriesReader::RequestInformation(
...
+            {
+            scalarType = VTK_FLOAT;
+            }
+          else if( max <= std::numeric_limits<double>::max() )
+            {
+            scalarType = VTK_DOUBLE;
+            }
+          else
+            {
+            throw std::string("Cannot find suitable scalar type for value range");
```
jcfr, line 939:
```suggestion
            // While throwing `std::string` is not considered to be best-practice, it
            // is done here to be consistent with how GDCM throw exception.
            throw std::string("Cannot find suitable scalar type for value range");
```

PR #7253, opened 2023-09-29
PERF: Improve performance when removing segmentation from SH tree

Sunderlandkyl:
When removing a vtkMRMLSegmentationNode using the SubjectHierarchy tree, all of the virtual children would be removing before the segmentation. This caused a large number of updates in various segmentation widgets as all of the widgets were regenerated when each of the segments were removed from the segmentation node.

This commit improves performance by blocking modification events on the segmentation node (or any other node with virtual children), until all of the children have been removed. Rendering is also blocked during the item removal to prevent unnecessary updates of the views.

### DIFF
--- a/Libs/MRML/Core/vtkMRMLSubjectHierarchyNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLSubjectHierarchyNode.cxx
@@ -1245,6 +1245,10 @@ void vtkSubjectHierarchyItem::ReparentChildrenToParent()
 //---------------------------------------------------------------------------
 void vtkSubjectHierarchyItem::RemoveAllChildren()
 {
+  // If the item has many children, then it may invoke a large number of events.
+  // To avoid this, block Modified events on the data node until all of the children are removed.
+  MRMLNodeModifyBlocker blocker(this->DataNode);
+
   std::vector<vtkIdType> childIDs;
   this->GetAllChildren(childIDs);
   while (childIDs.size())
--- a/Modules/Loadable/SubjectHierarchy/Widgets/qSlicerSubjectHierarchyPluginLogic.cxx
+++ b/Modules/Loadable/SubjectHierarchy/Widgets/qSlicerSubjectHierarchyPluginLogic.cxx
@@ -327,6 +327,8 @@ void qSlicerSubjectHierarchyPluginLogic::onNodeAboutToBeRemoved(vtkObject* scene
   vtkIdType itemID = shNode->GetItemByDataNode(dataNode);
   if (itemID)
     {
+    // Block render to avoid unnecessary view updates.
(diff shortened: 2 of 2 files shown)


Libs/MRML/Core/vtkMRMLSubjectHierarchyNode.cxx
```diff
@@ -1119,6 +1119,9 @@ bool vtkSubjectHierarchyItem::RemoveChild(vtkSubjectHierarchyItem* item)
   // as they represent the item's data node's content), then remove virtual branch
   if (removedItem->IsVirtualBranchParent())
     {
+    // If the item has many virtual children, then it may invoke a large number of events.
+    // To avoid this, block modified events on the data node until all of the children are removed.
+    MRMLNodeModifyBlocker blocker(removedItem->DataNode);
     removedItem->RemoveAllChildren();
```
jcfr, line 1125: In addition of `tkSubjectHierarchyItem::RemoveChild(vtkSubjectHierarchyItem* item)`, should this be done in these functions as well where `RemoveAllChildren` is also called? 
* `vtkSubjectHierarchyItem::DeepCopy(vtkSubjectHierarchyItem* item, bool copyChildren/*=true*/)`
* `vtkSubjectHierarchyItem::~vtkSubjectHierarchyItem()`
* `vtkSubjectHierarchyItem::RemoveChild(vtkIdType itemID)`
* `vtkMRMLSubjectHierarchyNode::Copy(vtkMRMLNode *anode)`

Even better, should we consider updating `RemoveAllChildren` to systematically block the event ?

Sunderlandkyl: That's a good point, the same performance issue could occur with any of those functions.
I've moved the MRMLNodeModifyBlocker into RemoveAllChildren.

Libs/MRML/Core/vtkMRMLSubjectHierarchyNode.cxx
```diff
@@ -1245,6 +1245,10 @@ void vtkSubjectHierarchyItem::ReparentChildrenToParent()
 //---------------------------------------------------------------------------
 void vtkSubjectHierarchyItem::RemoveAllChildren()
 {
+  // If the item has many children, then it may invoke a large number of events.
+  // To avoid this, block modified events on the data node until all of the children are removed.
```
jcfr, line 1249:
```suggestion
  // To avoid this, block Modified events on the data node until all of the children are removed.
  // Once all items have been removed, if it applies, a single Modified event is invoked.

jcfr (approved):

lassoan (approved):
Looks good to me, thank you.

PR #7260, opened 2023-10-02
ENH: Expose authentication parameters for DICOMweb networking

tbirdso:
These changes aim to expose authentication parameters in Python utilities
to aid in DICOM networking while avoiding implementation-specific
changes to the Slicer user interface. Generic methods such as a
user/password combination, OAuth token, or other application-specific
token can now (theoretically) be used in DICOMweb utility methods.

- A username/password combination can be persisted as a global setting
  with the `DICOMWEBUSER` and `DICOMWEBPWD` keys.
- A generic `requests.auth.AuthBase` object can be passed into Slicer
  utility methods for use in retrieving and stowing DICOM objects.
- DICOMSender is refactored for clarity. The DICOMSender public interface is
  limited to initialization and the `send` method. A DICOMweb
  authentication parameter is exposed, otherwise no functional changes
  are made.
- A method for verifying a Python module runtime requirement is
  generalized and moved from `DICOMSender` to `slicer.util`.

Manual testing results:
- DIMSE: ran `storescp` locally, exported sample data from the Slicer DICOM database to localhost and confirmed files populated in the storescp working directory
- DICOMweb: Ran a local [Orthanc](https://www.orthanc-server.com/download-windows.php) instance, confirmed that export with DICOMSender fails with missing/incorrect global authentication information and succeeds with correct auth information.

### DIFF
--- a/Modules/Scripted/DICOMLib/DICOMProcesses.py
+++ b/Modules/Scripted/DICOMLib/DICOMProcesses.py
@@ -1,12 +1,18 @@
 import logging
 import os
+import requests
 import subprocess
 import time
+from typing import Callable, Optional
+from requests.auth import HTTPBasicAuth
 
 import ctk
 import qt
 
+import dicomweb_client
 import slicer
+
+from DICOMLib.DICOMUtils import getGlobalDICOMAuth
 
 #########################################################
 #
@@ -34,33 +40,41 @@ class DICOMProcess:
     Code here depends only on python and DCMTK executables
     """
 
+    POSSIBLE_DCMTK_PATHS = [
+        '/../DCMTK-build/bin/Debug',
+        '/../DCMTK-build/bin/Release',
+        '/../DCMTK-build/bin/RelWithDebInfo',
+        '/../DCMTK-build/bin/MinSizeRel',
+        '/../DCMTK-build/bin',
+        '/../CTK-build/CMakeExternals/Install/bin',
+        '/bin'
+    ]
+    PROCESS_STATE_NAMES = {0: 'NotRunning', 1: 'Starting', 2: 'Running', }
+
+    @classmethod
+    def getDCMTKToolsPath(
+        cls,
+        additionalPaths: list[str] = []
+    ) -> str:
+        """
+        Get the first candidate directory that may contain DCMTK tools
+        that is available on the system.
+        """
+        relSearchPaths = cls.POSSIBLE_DCMTK_PATHS
+        relSearchPaths.extend(additionalPaths)
+
+        try:
+            absSearchPaths = (f'{slicer.app.slicerHome}{path}' for path in relSearchPaths)
+            return next(path for path in absSearchPaths
+                        if os.path.exists(path))
+        except StopIteration:
+            raise UserWarning("Could not find a valid path to DICOM helper applications")
+
     def __init__(self):
         self.process = None
         self.connections = {}
-        pathOptions = (
-            '/../DCMTK-build/bin/Debug',
-            '/../DCMTK-build/bin/Release',
-            '/../DCMTK-build/bin/RelWithDebInfo',
-            '/../DCMTK-build/bin/MinSizeRel',
-            '/../DCMTK-build/bin',
-            '/../CTK-build/CMakeExternals/Install/bin',
-            '/bin'
-        )
-
-        self.exeDir = None
-        for path in pathOptions:
-            testPath = slicer.app.slicerHome + path
-            if os.path.exists(testPath):
-                self.exeDir = testPath
-                break
-        if not self.exeDir:
-            raise UserWarning("Could not find a valid path to DICOM helper applications")
-
-        self.exeExtension = ""
-        if os.name == 'nt':
-            self.exeExtension = '.exe'
-
-        self.QProcessState = {0: 'NotRunning', 1: 'Starting', 2: 'Running', }
+        self.exeDir = self.getDCMTKToolsPath()
+        self.exeExtension = ".exe" if os.name == "nt" else ""
 
     def __del__(self):
         self.stop()
@@ -79,10 +93,10 @@ class DICOMProcess:
         self.process.start(cmd, args)
         return self.process
 
-    def onStateChanged(self, newState):
+    def onStateChanged(self, newState: int):
         """Callback to indicate that the process state has changed
         and is now either starting, running, or not running."""
-        logging.debug(f"Process {self.cmd} now in state {self.QProcessState[newState]}")
+        logging.debug(f"Process {self.cmd} now in state {self.PROCESS_STATE_NAMES[newState]}")
         if newState == 0 and self.process:
             stdout = self.process.readAllStandardOutput()
             stderr = self.process.readAllStandardError()
@@ -94,7 +108,7 @@ class DICOMProcess:
 
     def stop(self) -> None:
         """Stop the running standalone process."""
-        if hasattr(self, "process") and self.process:
+        if self.process:
             logging.debug("stopping DICOM process")
             self.process.kill()
             # Wait up to 3 seconds for the process to stop
@@ -107,7 +121,7 @@ class DICOMCommand(DICOMProcess):
     Run a generic dcmtk command and return the stdout
     """
 
-    def __init__(self, cmd, args):
+    def __init__(self, cmd: str, args: list[str]):
         super().__init__()
         self.executable = self.exeDir + '/' + cmd + self.exeExtension
         self.args = args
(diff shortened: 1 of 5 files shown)


Modules/Scripted/DICOMLib/DICOMUtils.py
```diff
@@ -911,11 +913,25 @@ def onNodeAdded(caller, event, calldata):
...
+
+
+def getGlobalDICOMAuth() -> Optional[requests.auth.HTTPBasicAuth]:
+    """Get the global authentication settings for DICOM networking, if initialized."""
+    user = slicer.util.settingsValue(GLOBAL_DICOMWEB_USER_KEY, "")
+    pwd = slicer.util.settingsValue(GLOBAL_DICOMWEB_PASSWORD_KEY, "")
+    return requests.auth.HTTPBasicAuth(user, pwd) if user or pwd else None
+
+
+# ------------------------------------------------------------------------------
```
jcfr, line 928:
```suggestion
```

Modules/Scripted/DICOMLib/DICOMUtils.py
```diff
@@ -911,11 +913,25 @@ def onNodeAdded(caller, event, calldata):
     return loadedNodeIDs
 
 
+# ------------------------------------------------------------------------------
```
jcfr, line 916:
```suggestion
```

Modules/Scripted/DICOMLib/DICOMUtils.py
```diff
@@ -911,11 +913,25 @@ def onNodeAdded(caller, event, calldata):
...
+GLOBAL_DICOMWEB_USER_KEY = "DICOMWEBUSER"
+GLOBAL_DICOMWEB_PASSWORD_KEY = "DICOMWEBPWD"
```
jcfr, line 918:
```suggestion
GLOBAL_DICOMWEB_USER_KEY = "GLOBAL_DICOMWEB_USER_KEY"
GLOBAL_DICOMWEB_PASSWORD_KEY = "GLOBAL_DICOMWEB_PASSWORD_KEY"
```

jcfr (commented):
Thanks for the contribution.

Waiting we can further test, here are a few minor comments

Base/Python/slicer/util.py
```diff
@@ -3795,6 +3795,64 @@ def pip_uninstall(requirements):
     _executePythonModule('pip', args)
 
 
+def validate_module_version(
+    module_name: str,
+    minimum_version: str,
+    attempt_fix: bool = False,
+) -> None:
```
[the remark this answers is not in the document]
tbirdso: I agree with the suggested flake8 configurations, but I'm not sure I understand the suggested change to remove the `None` return type annotation? IMO that type hint is preferable to indicate that the `None` return value is intended, such as in this case where a validation function will throw an error, produce a side effect (restart the app), or otherwise do nothing. Removing the return type hint indicates that the return type is unknown, which is not correct.

Note that `None` return type is used throughout Python examples: https://peps.python.org/pep-0484/#type-aliases

But, if this is the standard for Slicer type annotation convention then I will change it.

Modules/Scripted/DICOMLib/DICOMProcesses.py
```diff
@@ -116,14 +120,17 @@ def __init__(self, cmd, args):
...
+        """
+        Run the DICOM process and block until return.
 
-    def start(self):
-        # run the process!
-        self.process = qt.QProcess()
+        Returns:
+            A qt.QByteArray representing standard output (stdout)
+            from the DICOM process.
+            https://doc.qt.io/qt-5/qprocess.html#readAllStandardOutput
```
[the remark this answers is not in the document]
tbirdso: Updated

Modules/Scripted/DICOMLib/DICOMUtils.py
tbirdso: Updated. Note that other functions in this file are broken up with this divider though.

tbirdso:
From offline conversation with @jcfr I will work to add instructions for testing `slicer.util` and DICOMweb contributions. Unfortunately test automation is beyond the scope of this effort for now.

These changes include refactoring of `DICOMSender` DIMSE code paths. It would be good to test that functionality to ensure that refactoring did not introduce DIMSE behavioral changes. @pieper , do you maybe have suggestions on how to test DIMSE stow functionality?

lassoan:
Probably the easiest way to test DICOM store is to run DCMTK's storescp (C-STORE service class provider).

pieper:
>  @pieper , do you maybe have suggestions on how to test DIMSE stow functionality?

You can have a look at how query / retrieve are tested in CTK.  We run the dcmtk executables as @lassoan suggested and configure it all via cmake:

https://github.com/commontk/CTK/blob/master/Applications/Testing/Cpp/CMakeLists.txt

https://github.com/commontk/CTK/blob/master/Applications/Testing/Cpp/ctkDICOMApplicationTest1.cpp

Base/Python/slicer/util.py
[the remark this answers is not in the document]
jcfr: > I'm not sure I understand the suggested change to remove the None return type annotation?

You are correct. This was on oversight on my side.

The one we should omit is the annotation for `Self`. See https://docs.astral.sh/ruff/rules/missing-type-self/

tbirdso:
Thanks @lassoan @pieper for pointing me in the right direction. I've tested refactored DICOM networking code paths with the following approaches:
- DIMSE: ran `storescp` on my local system at @lassoan 's suggestion, exported sample data from the Slicer DICOM database to `localhost` and confirmed files populate in the `storescp` working directory
- DICOMweb: I set up a local [Orthanc](https://www.orthanc-server.com/download-windows.php) instance with its DICOMweb plugin and tested both with and without authentication. Confirmed that export with `DICOMSender` fails with missing/incorrect global authentication information and succeeds with correct auth information.

Unfortunately I do not have resources to pursue automated testing at this time. However, I've added a note to DICOM developer documentation to help reproduce similar manual testing in the future.

I've added a few extra documentation updates and QOL improvements reflective of that testing. Please take a moment to (re-)review at your convenience. cc @jcfr

Modules/Scripted/DICOMLib/DICOMProcesses.py
```diff
@@ -392,118 +412,135 @@ def processStdoutLine(self, line):
...
+
+        :raises ModuleNotFoundError: if `dicomweb_client<0.51` and DICOMweb STOW-RS protocol is requested.
+        :raises UserWarning: if a transfer is cancelled.
+        """
+        # Ensure that correct version of dicomweb-client Python package is installed
+        # pythonweb-client 0.50 was broken (https://github.com/MGHComputationalPathology/dicomweb-client/issues/41)
+        minimumDicomwebClientVersion = "0.51"
+        slicer.util.validate_module_version(
+            "dicomweb_client", minimumDicomwebClientVersion
+        )
```
jcfr, line 509:
```suggestion
```

The original logic introduced in https://github.com/Slicer/Slicer/commit/b89b3c8a4b2a209ba3654377ee3050da62a14877 and updated in https://github.com/Slicer/Slicer/commit/c25b4e6212849b14f8d2c69324b4ee88b6abd558.

It was intended to check if the version of `dicomweb_client` provided  by Slicer needed to be updated.

Since then, the version has been updated[^1] and this logic can simply be removed.

[^1]: https://github.com/Slicer/Slicer/commits/main/SuperBuild/External_python-dicom-requirements.cmake

Base/Python/slicer/util.py
```diff
@@ -3795,6 +3795,71 @@ def pip_uninstall(requirements):
...
+        slicer.app.processEvents()
+        try:
+            failure_message = f"Failed to install {module_name}>={minimum_version}."
+            with slicer.util.tryWithErrorDisplay(failure_message):
+                pip_install(f"{module_name}>={minimum_version}")
+        finally:
+            progressDialog.close()
+
+    if needRestart:
+        restart()
```
jcfr, line 3860:
```suggestion
```

I suggest we revisit the generalization of in a another pull request and simply remove the check currently added in `DICOMProcesses`

Instead of `validate_module_version`, having a method called `is_python_module_installed` would be more explicit and focused on doing on task: _Checking that a specific module (and optinally specified version) is installed_

Strictly speaking[^1], raising `ModuleNotFoundError` is not _valid_ in the sense that a module could be installed without having the correct version. I am not sure redefining the meaning of this specific exception in he context of this function is sensible.

Installation, confirmation[^2], progress reporting and restart[^3] should be handled outside of this function.

### Example

An idea would be to support the following:

```
moduleName="dicomweb_client"
versionSpecifier = ">=0.51"

if not slicer.util.is_python_module_installed(
    moduleName, version_specifier=versionSpecifier
):
    if slicer.util.confirmOkCancelDisplay("This current operation requires '{module_name}' Python package. Click OK to install it now."):
        slicer.util.pip_install(f"{moduleName}{versionSpecifier}")
```


[^1]: https://docs.python.org/3/library/exceptions.html#ModuleNotFoundError

[^2]: Handle using the approach suggested in at https://slicer.readthedocs.io/en/latest/developer_guide/script_repository.html#install-a-python-package

[^3]: Restart should also be handled with a confirmation dialog.

Modules/Scripted/DICOMLib/DICOMProcesses.py
tbirdso: Thanks for the background. I agree that this check should be unnecessary in the current Slicer build. My goal in this PR has been to introduce authentication paths for DICOMweb integration and otherwise improve code structure with minimal possible changes to functionality/interface. But, if it is more convenient I will remove the unneeded check here.

Base/Python/slicer/util.py
tbirdso: > I suggest we revisit the generalization of in a another pull request

That's fine. I won't have bandwidth to drive refinement in that separate PR, but if you think this is a worthwhile utility to pursue I can open a draft PR for someone else to take across the finish line.

> and simply remove the check currently added

Will do.

tbirdso: Tracked in https://github.com/Slicer/Slicer/pull/7268 and removed from this PR.

Modules/Scripted/DICOMLib/DICOMProcesses.py
tbirdso: The `dicomweb_client` version check is now removed from `DICOMSender` in this PR.

sjh26 (approved):
lgtm

tbirdso:
On review of `CONTRIBUTING.md` guidelines I will wait for an additional approval before merge.

> Consensus is reached. This usually means that at least two reviewers approved the changes

tbirdso:
@jcfr Ping for review

jcfr (approved):
Thanks for revisiting the patches :pray: :100: :tada:

PR #7270, opened 2023-10-10
ENH: Propagate args / kwargs to loadLoadables

marcus-wirtz-snkeos:
Would be great to have the progressbar option of `loadSeriesByUID` also for the higher level functions.

### DIFF
--- a/Modules/Scripted/DICOMLib/DICOMUtils.py
+++ b/Modules/Scripted/DICOMLib/DICOMUtils.py
@@ -22,7 +22,7 @@ comment = """
 
 
 # ------------------------------------------------------------------------------
-def loadPatientByUID(patientUID):
+def loadPatientByUID(patientUID, messages=None, progressCallback=None):
     """ Load patient by patient UID from DICOM database.
     Returns list of loaded node ids.
 
@@ -76,7 +76,7 @@ def loadPatientByUID(patientUID):
     if len(seriesUIDs) == 0:
         raise OSError('No series found in patient with DICOM database UID ' + patientUIDstr)
 
-    return loadSeriesByUID(seriesUIDs)
+    return loadSeriesByUID(seriesUIDs, messages, progressCallback)
 
 
 # ------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


marcus-wirtz-snkeos:
> Thank you for your contribution, I agree that it can be useful to expose `messages` and `progressCallback` arguments at higher-level functions, too. I would prefer to explicitly specify these arguments, as it would make the API more clear (and 2 extra arguments are not too many and I don't expect that `loadLoadables` will have many new options in the future).

Alright, I explicitly specified the arguments. I realized that the main time consumer (~80%) in most of my cases is actually the function `getLoadablesFromFileLists` rather than `loadLoadable`. Does that match your experience? I updated the PR to pass the progressbar to the longer lasting function call.

jcfr (approved):

jcfr:
Nitpick: Adding type annotation would be nice as well though this is outside of the immediate scope of this pull request

marcus-wirtz-snkeos:
> Nitpick: Adding type annotation would be nice as well though this is outside of the immediate scope of this pull request

I don't see a single type hint in this entire module so far (even in the entire `Scripted` folder?) :-)

jcfr:
We indeed only recently started to introduce them in new code.

For example:
* `Modules/Scripted/DICOMLib`[^1]
* `Base/Python/slicer/parameterNodeWrapper`[^2]

In this case, adding annotation for `messages` would be sufficient. Adding the type for callback would be more involved.

[^1]: https://github.com/Slicer/Slicer/blob/main/Modules/Scripted/DICOMLib/DICOMProcesses.py#L55-L58
[^2]: https://github.com/Slicer/Slicer/blob/main/Base/Python/slicer/parameterNodeWrapper/wrapper.py

Modules/Scripted/DICOMLib/DICOMUtils.py
```diff
@@ -22,7 +22,7 @@
 
 
 # ------------------------------------------------------------------------------
-def loadPatientByUID(patientUID):
+def loadPatientByUID(patientUID, messages=None, progressCallback=None):
```
jcfr, line 25:
```suggestion
def loadPatientByUID(patientUID, messages: Optional[list[str]]=None, progressCallback=None):
```

You  also need to add the following at the top:

```
from typing import Optional
```

jcfr:
Integrating. Annotation type will be introduced later.

PR #7271, opened 2023-10-10
BUG: Use main project version for macOS short version string in bundle

sjh26:
This ensures that the custom application version number is used to the macOS 'Quick Look' feature

Follow on fix to: https://github.com/Slicer/Slicer/commit/d004e4e394bba81d92df506b70b4b263656c094d for another version string that was missed in the initial fix.

Example of the bug:

Screenshot 2023-10-10 10 17 21


pieper (approved):
Good catch 👍

jcfr (approved):

PR #7272, opened 2023-10-10
ENH: Simplify single-image capture in ScreenCapture module

lassoan:
Instead of having an obscure "single" button next to the "number of frames" selector, add a "none" animation mode and make it the default. Options that are not relevant are not just disabled but hidden, to make it easier for the user to find the relevant options.

| Before | After |
|--|--|
| image | image |

### DIFF
--- a/Docs/user_guide/modules/screencapture.md
+++ b/Docs/user_guide/modules/screencapture.md
@@ -8,9 +8,10 @@ This module is for creating videos, image sequences, or lightbox image from 3D a
 
 ### Input
 
-- **Main view:** This view will be changed during the animation (rotated, sweeped, etc.).
-- **Capture all views:** If enabled then all the view in the view layout will be captured. If disabled then only the main view will be captured. By capturing all views, it is possible to see the animated view (such as a moving slice) in 3D and in other slice views.
-- **Animation mode:** specifies how the main view will be modified during capture.
+- **Main view:** This view is used for capturing (rotated, sweeped, etc.).
+- **Capture all views:** When enabled, all views in the layout will be captured. Otherwise, only the main view is captured. By capturing all views, it is possible to see the animated view (such as a moving slice) in 3D and in other slice views.
+- **Capture mode:** specifies how the main view will be modified during capture.
+  - **single frame:** Acquire a single frame of selected main view or all views if "Capture all views" is enabled.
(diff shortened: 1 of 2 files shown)


pieper (dismissed):
This is a great idea.  I'm not sure "Animation mode" of None will make sense to people.  How about just "Capture mode" of "Single Frame", "Sweep animation", "Rotate animation" etc to be explicit?

jamesobutler:
I agree with Steve. User's likely don't expect to have to make a decision about anything Animation related when desiring a single capture. However if it was named "Capture Mode" then it being "Single Frame" makes sense with other options being different animation types. The question would be whether users who are looking for animation would be confused about it being in a Capture mode setting.

pieper:
I'm thinking we might also want a mode that records a frame whenever a node is updated.  Like if you wanted to make a recording of a simulation.

lassoan:
Good suggestions, I'll use "Capture mode".

For recording on node change: we already have a sequence mode that kind of does this. It is a bit complicated because you need to set what nodes are recorded. How would you simplify selection of the observed nodes? Would you record on real-time? How would it be different from a screen capture?

pieper:
Regarding the triggering on a node change, I was thinking it could be just a node selector.  It would just trigger saving a frame whenever that node was modified.  

But now that I think about it maybe a better would be to record a frame every time a render completes.  That way, for example, you could be interacting with something that is slow to compute or render, but then playback in at whatever speed you want.  Or if it was an automated thing, like a simulation or iterative segmentation that takes a long time you could easily record the evolution of it to replay later.

jcfr:
The changes have been rebased against current `main` to account for recent changes related to double-quotes[^1] and docstrings[^2]. 

[^1]: https://github.com/Slicer/Slicer/pull/7345
[^2]: https://github.com/Slicer/Slicer/pull/7387

jcfr:
To follow-up, the last commit rename "Animation mode" to "Capture mode" and change the default "none" animation mode to "single frame".

jcfr:
@lassoan @pieper @jamesobutler Now that the pull request was updated and that a new issue[^1] was created to capture the suggestion, any objection to move forward with integration ?

[^1]: https://github.com/Slicer/Slicer/issues/7405

lassoan (commented):
Thank you very much, it all looks good. Since it was my own PR, I cannot submit a review or approve.

jcfr (approved):

PR #7275, opened 2023-10-11
BUG: Fix a typo in qSlicerSigmentationSettingsPanel signal/slot connection

MichaelColonel:
There is no signal called setDefaultOverwriteMode in the qSlicerSigmentationSettingPanel.
Qt emits a warning during the Slicer execution process.


lassoan (approved):
Great, thank you for the fix.

PR #7276, opened 2023-10-11
ENH: Avoid floating SH combobox popup by default

malbi:
Align pop-up vertically to combo box can lead to pop-up being rendered outside the limit of the main window.
Here is a proposal to disable it by default and render the pop-up tree below the combo box to be consistent with other combo boxes behavior.


lassoan:
To me, it seems that the current default behavior (alignPopupVertically=True) is better for keeping the popup in a reachable location. Could you please provide an example screenshot of a case when alignPopupVertically=False is better than alignPopupVertically=True?

malbi:
With `alignPopupVertically=true`, when selecting an item in a long list and scrolling farther away from this item I got this behavior: 
alignPopupVerticallyTrue_1
alignPopupVerticallyTrue_2
alignPopupVerticallyTrue_3


With alignPopupVertically=false, in the same context I got a consistent display of the selection tree:
alignPopupVerticallyFalse_1
alignPopupVerticallyFalse_2
alignPopupVerticallyFalse_3

I did not spent any time to create this combobox in an appropriate location in the Slicer UI, but in our use case, it is located in the upper part of our custom window and having to change the alignVerticallyPopup property for each combobox to not have them floating away from the screen in case of big number of item seems a bit weird.

PR #7278, opened 2023-10-11
ENH: Simplify paint/erase diameter options for more concise UI

jamesobutler:
Currently the diameter of the tool could be changed in 3 different ways - by SpinBox, Slider, and a set of buttons. This change simplifies the options to be more concise by only have a slider and spinbox along with the button to toggle between absolute and relative sizing. The design is based on the existing Markups module glyph size widget options.

| `main` | This PR |
|--------|----------|
|image|image|
|image|image|

Markups module glyph size widget option for reference:
image

### DIFF
--- a/Modules/Loadable/Segmentations/EditorEffects/qSlicerSegmentEditorPaintEffect.cxx
+++ b/Modules/Loadable/Segmentations/EditorEffects/qSlicerSegmentEditorPaintEffect.cxx
@@ -35,6 +35,7 @@
 #include <QLabel>
 #include <QToolButton>
 #include <QVBoxLayout>
+#include <QFormLayout>
 
 // VTK includes
 #include <vtkActor.h>
@@ -87,7 +88,7 @@
 // Slicer includes
 #include "qMRMLSliceView.h"
 #include "qMRMLSliceWidget.h"
-#include "qMRMLSpinBox.h"
+#include "qMRMLSliderWidget.h"
 #include "qMRMLThreeDView.h"
 #include "qMRMLThreeDWidget.h"
 #include "qSlicerLayoutManager.h"
@@ -234,9 +235,9 @@ qSlicerSegmentEditorPaintEffectPrivate::qSlicerSegmentEditorPaintEffectPrivate(q
   , ActiveViewWidget(nullptr)
   , PaintOptionsFrame(nullptr)
   , BrushDiameterFrame(nullptr)
-  , BrushDiameterSpinBox(nullptr)
-  , BrushDiameterSlider(nullptr)
-  , BrushDiameterRelativeToggle(nullptr)
+  , BrushDiameterSliderWidget(nullptr)
+  , BrushDiameterSizeFrame(nullptr)
+  , BrushDiameterIsAbsoluteButton(nullptr)
   , BrushSphereCheckbox(nullptr)
   , EditIn3DViewsCheckbox(nullptr)
   , ColorSmudgeCheckbox(nullptr)
@@ -618,20 +619,6 @@ void qSlicerSegmentEditorPaintEffectPrivate::onDiameterUnitsClicked()
   {
     q->setCommonParameter("BrushDiameterIsRelative", 0);
   }
-}
-
-//-----------------------------------------------------------------------------
-void qSlicerSegmentEditorPaintEffectPrivate::onQuickDiameterButtonClicked()
-{
-  QToolButton* senderButton = dynamic_cast<QToolButton*>(sender());
-  if (!senderButton)
-  {
-    qWarning() << Q_FUNC_INFO << " failed: invalid sender button";
-    return;
-  }
-  int diameter = senderButton->property("BrushDiameter").toInt();
-
-  this->onDiameterValueChanged(diameter);
 }
 
 //-----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


Modules/Loadable/Segmentations/EditorEffects/qSlicerSegmentEditorPaintEffect.cxx
```diff
@@ -1558,49 +1536,40 @@ void qSlicerSegmentEditorPaintEffect::updateGUIFromMRML()
...
+    d->BrushDiameterSliderWidget->setSuffix("%");
+    d->BrushDiameterSliderWidget->setDecimals(0);
     }
   else
     {
-    d->BrushDiameterSpinBox->setQuantity("length");
-    d->BrushDiameterSpinBox->setUnitAwareProperties(qMRMLSpinBox::Prefix | qMRMLSpinBox::Suffix);
-    int decimals = (int)(log10(d->BrushDiameterSlider->minimum()));
+    d->BrushDiameterSliderWidget->setQuantity("length");
+    d->BrushDiameterSliderWidget->setSuffix(tr("mm"));
```
lassoan, line 1564: This is not necessary. If the MRML scene is set in the widget, suffix is set to be unit-aware, and quantity is set to `length` then the widget gets the current suffix from the scene.

lassoan (commented):
I never use those buttons and I don't think those preset buttons are very helpful, so I don't mind if they are removed. I always use shift-scrollwheel for gross adjustment and zoom in/out the view for fine adjustment.

Even the slider has limited value without a preview of the brush size in views, but since this is the only easily discoverable way for changing the brush size (mouse/keyboard gestures are more effective but not discoverable) we need to keep the slider. We should briefly show the brush size in each view as the slider is moved.

Modules/Loadable/Segmentations/EditorEffects/qSlicerSegmentEditorPaintEffect.cxx
jamesobutler: Ok I can remove these. This wasn't new code as the `BrushDiameterSpinBox`, a qMRMLSpinBox, was previously having the quantity and suffix properties being set. See 
https://github.com/Slicer/Slicer/blob/3f425197ee857eb1ebffb13f9c24bdc3b0066b69/Modules/Loadable/Segmentations/EditorEffects/qSlicerSegmentEditorPaintEffect.cxx#L1589-L1590

jamesobutler:
> Even the slider has limited value without a preview of the brush size in views

I think the slider's general value is that when using shift-scrollwheel that you can see the slider change and you can know why the display isn't getting any bigger or smaller because it is at the slider bounds. That is probably more noticeable than seeing the up/down controls of the spinbox becoming disabled when at the bounds.

Modules/Loadable/Segmentations/EditorEffects/qSlicerSegmentEditorPaintEffect.cxx
lassoan: Suffix need to be set for %, but not for physical unit (then the widget sets for itself).

jamesobutler: I've pushed to remove the following lines
```c++
    d->BrushDiameterSliderWidget->setQuantity("length");
    d->BrushDiameterSliderWidget->setSuffix(tr("mm"));
```

Modules/Loadable/Segmentations/EditorEffects/qSlicerSegmentEditorPaintEffect.cxx
```diff
@@ -1558,49 +1536,38 @@ void qSlicerSegmentEditorPaintEffect::updateGUIFromMRML()
...
+    d->BrushDiameterSliderWidget->setQuantity("");
+    d->BrushDiameterSliderWidget->setSuffix("%");
+    d->BrushDiameterSliderWidget->setDecimals(0);
     }
   else
     {
-    d->BrushDiameterSpinBox->setQuantity("length");
-    d->BrushDiameterSpinBox->setUnitAwareProperties(qMRMLSpinBox::Prefix | qMRMLSpinBox::Suffix);
-    int decimals = (int)(log10(d->BrushDiameterSlider->minimum()));
+    d->BrushDiameterSliderWidget->setUnitAwareProperties(qMRMLSliderWidget::Prefix | qMRMLSliderWidget::Suffix);
```
lassoan, line 1563: Since `quantity` was set to empty, we need to restore it here. Or keep `quantity` property unchanged all the time and only change `unitAwareProperties` (and set `suffix` to `%` in relative mode).

cpinter:
For the record I also don't think the presets are useful, so agree with the proposed changes. Thanks @jamesobutler !

Modules/Loadable/Segmentations/EditorEffects/qSlicerSegmentEditorPaintEffect.cxx
jamesobutler: I've now restored the setting of quantity and suffix.

jamesobutler:
@lassoan I’m looking back on this and wondering my if there is still something wrong with the units stuff for these mrml widgets. I’m not familiar with these types so any small change you can provide I’ll take it as a learning! Thanks

cpinter:
Hi @jamesobutler and @lassoan ! I'm wondering if there is anything in this PR that still needs to be fixed.

I'm working on a project where the inconsistencies of this widget are causing problems, and removing the buttons would help (i.e. max button is 40% but slider max is 25%). Another issue with % is that it does not seem to actually be related to screen size, at least on Linux. If I find something interesting I'll let you know. Thanks!

lassoan:
I'll test it a bit more today, but it should be good to go.

@cpinter Have you tested the changes proposed in this PR?

> Another issue with % is that it does not seem to actually be related to screen size, at least on Linux.

The % is computed from the screen height:

```
screenSizePixel = threeDWidget->threeDView()->renderWindow()->GetScreenSize()[1];
mmPerPixel = sqrt(xyToSlice->GetElement(0, 1)*xyToSlice->GetElement(0, 1)
      + xyToSlice->GetElement(1, 1)*xyToSlice->GetElement(1, 1)
      + xyToSlice->GetElement(2, 1)*xyToSlice->GetElement(2, 1));
double newBrushAbsoluteDiameter = screenSizePixel * (brushRelativeDiameter / 100.0) * mmPerPixel; 
```

It is possible that VTK render window does not get the correct screen size in all environments, especially in unusual cases (e.g., offscreen rendering, when you move the application from one screen to another, etc.).

Modules/Loadable/Segmentations/EditorEffects/qSlicerSegmentEditorPaintEffect.cxx
```diff
@@ -234,8 +235,7 @@ qSlicerSegmentEditorPaintEffectPrivate::qSlicerSegmentEditorPaintEffectPrivate(q
   , ActiveViewWidget(nullptr)
   , PaintOptionsFrame(nullptr)
   , BrushDiameterFrame(nullptr)
-  , BrushDiameterSpinBox(nullptr)
-  , BrushDiameterSlider(nullptr)
+  , BrushDiameterSliderWidget(nullptr)
```
lassoan, line 238: Add initialization of `BrushDiameterSizeFrame`.
To avoid all these initializations here (and risk missing some of them), it would be probably better to initialize all in the header file.

Modules/Loadable/Segmentations/EditorEffects/qSlicerSegmentEditorPaintEffect.cxx
```diff
@@ -1416,39 +1402,32 @@ void qSlicerSegmentEditorPaintEffect::setupOptionsFrame()
...
-    d->BrushDiameterFrame->layout()->addWidget(quickDiameterButton);
-    QObject::connect(quickDiameterButton, SIGNAL(clicked()), d, SLOT(onQuickDiameterButtonClicked()));
-    }
-
-  d->BrushDiameterRelativeToggle = new QToolButton();
-  d->BrushDiameterRelativeToggle->setText("%");
+  d->BrushDiameterRelativeToggle = new QToolButton(d->BrushDiameterSizeFrame);
+  d->BrushDiameterRelativeToggle->setCheckable(true);
+  d->BrushDiameterRelativeToggle->setText("absolute");
+  d->BrushDiameterRelativeToggle->setChecked(false); // relative
```
lassoan, line 1421: There is a contradiction here. If "Relative" button is set to "false" then its state is "absolute".
I would recommend to use the same naming convention as in markups module: `brushDiameterIsAbsoluteButton`. If it is checked then it means diameter is absolute (if unchecked then diameter is relative).

Modules/Loadable/Segmentations/EditorEffects/qSlicerSegmentEditorPaintEffect.cxx
```diff
@@ -1416,39 +1402,32 @@ void qSlicerSegmentEditorPaintEffect::setupOptionsFrame()
...
   d->BrushDiameterRelativeToggle->setToolTip(tr("Toggle diameter quick set buttons between percentage of window size / absolute size in millimeters"));
-  d->BrushDiameterFrame->layout()->addWidget(d->BrushDiameterRelativeToggle);
 
-  d->BrushDiameterSlider = new ctkDoubleSlider();
-  d->BrushDiameterSlider->setOrientation(Qt::Horizontal);
-  paintOptionsLayout->addWidget(d->BrushDiameterSlider);
+  d->BrushDiameterSliderWidget = new qMRMLSliderWidget(d->BrushDiameterSizeFrame);
+  d->BrushDiameterSliderWidget->setToolTip(tr("Set the paint brush size as percentage of screen size or as fixed length"));
```
lassoan, line 1425:
```suggestion
  d->BrushDiameterSliderWidget->setToolTip(tr("Set the paint brush size as percentage of screen size or as physical size"));
```

lassoan (commented):
Thank you, it works well and the code looks mostly good, just added a few minor comments inline.

cpinter:
Just for the record, it is confirmed that the relative brush size works correctly, The only issue is that the tooltip in the % button says “percentage of window size”, which should say screen size, and possibly make it clear that it refers to the height.

cpinter (approved):
LGTM! Sorry for the delay.

Issue #7288, opened 2023-10-18
Fix py_SubjectHierarchyGenericSelfTest  test


jcfr:
## Summary

The test `py_SubjectHierarchyGenericSelfTest` passed on all platform when executed against a build of Slicer 31ba8942a031b907d4be2f2f5d81db717b347a1b  and started to fail with Slicer  80ad0a04dacf134754459557bf2638c63f3d1d1b.

| | :white_check_mark: | :x: |
|---|---|---|
| CDash link | https://slicer.cdash.org/test/25604578 | https://slicer.cdash.org/test/25610246 | 
| Slicer Source | 31ba8942a031b907d4be2f2f5d81db717b347a1b | 80ad0a04dacf134754459557bf2638c63f3d1d1b |

### Useful links

* Compare link: https://github.com/Slicer/Slicer/compare/31ba8942a031b907d4be2f2f5d81db717b347a1b...80ad0a04dacf134754459557bf2638c63f3d1d1b

* Associated pull requests:
  * https://github.com/Slicer/Slicer/pull/7214
  * https://github.com/Slicer/Slicer/pull/7234

* Test sources: https://github.com/Slicer/Slicer/blob/main/Modules/Loadable/SubjectHierarchy/Testing/Python/SubjectHierarchyGenericSelfTest.py

## Steps to reproduce

```
ctest -C Release -R py_SubjectHierarchyGenericSelfTest

[...]

308: FAIL: test_SubjectHierarchyGenericSelfTest_FullTest1 (SubjectHierarchyGenericSelfTest.SubjectHierarchyGenericSelfTestTest)
308: ----------------------------------------------------------------------
308: Traceback (most recent call last):
308:   File "/home/jcfr/Projects/Slicer-Release/Slicer-build/lib/Slicer-5.5/qt-scripted-modules/SubjectHierarchyGenericSelfTest.py", line 113, in test_SubjectHierarchyGenericSelfTest_FullTest1
308:     self.section_AttributeFilters()
308:   File "/home/jcfr/Projects/Slicer-Release/Slicer-build/lib/Slicer-5.5/qt-scripted-modules/SubjectHierarchyGenericSelfTest.py", line 565, in section_AttributeFilters
308:     testAttributeFilters(shProxyModel, shProxyModel)
[...]

PR #7289, opened 2023-10-18
BUG: Fix SubjectHierarchyGenericSelfTest

cpinter:
- The test assumed incorrect shown item count including the empty folder (which flag is enabled by default), expected item counts has been fixed
- Add test for the showEmptyHierarchyItems flag, which has not been tested until now

Re #7288


jcfr:
Thanks for the PR, OI further tweaked by generalizing `testAttributeFilters` to accept a `showEmptyHierarchyItems` argment.

If that looks good, I will squash the commit.

cpinter:
Sure, looks good, thank you!

jcfr (approved):
Local testing on Ubuntu confirmed that the test is now passing.

Issue #4982, opened 2020-06-12
Support loading of images with long int scalar type


lassoan:
This is a follow-up to the issue described here: https://github.com/Slicer/Slicer/issues/4714

VTK does not support reslicing of images with long int scalar type. Image readers would need to cast the image to "int" to make everything work properly. It is highly unlikely that the image would actually contain values that cannot be represented on a regular "int", but if we really want to be safe then a warning could be displayed if this happens.

pieper:
It would be easy enough to check the scalar range and only warn the user if there is data loss.  We could also consider adding a flag in the storage node to save in a particular datatype, and this could be set to match the input.

fepegar:
Are all the Mantis issues gone? I'm pretty sure I reported this there some months/years ago.

lassoan:
All the Mantis issues are preserved at the original bugtracker. Moreover, we have placeholders for all Mantis issues in this github repository (same issue number and headline).

Maybe you mean that search feature is broken on https://mantisarchive.slicer.org/?
@jcfr is this expected? If mantisarchive site is not dynamic, could we somehow encourage google to index it? 
Currently, googling for `volume site:mantisarchive.slicer.org` returns 12 hits results, while I would expect hundreds.

ClaudioHoffmann:
Is this just about 64-bit images? At least as of Slicer Version `5.3.0-2023-06-28 r31836 / d1385c5`, 32-bit images seem to work fine when streamed over OpenIGTLink, and it's only this related check that blocks them from being appropriately resliced:
https://github.com/Slicer/Slicer/blob/7ef59614c2458e4f693ab77834e6726cf6ae0540/Libs/MRML/Logic/vtkMRMLSliceLayerLogic.cxx#L655-L664

The commit description of 7e10c1f48160f874ade53b3fb76ba0b7eee74d97 suggests that this limitation arises from [vtkImageInterpolator's use of `double`s](https://gitlab.kitware.com/vtk/vtk/-/blob/81482725188df9e8e3ee4b4674d46537e7b8c44e/Imaging/Core/vtkImageInterpolator.cxx#L12-17), but a double can losslessly fit 32-bit integers. [The original Mantis issue had an attachment](https://mantisarchive.slicer.org/view.php?id=4714), but the download doesn't seem to work for me anymore.

Steps to reproduce working reslicing of 32-bit images:
* Build [OpenIGTLink](https://github.com/openigtlink/OpenIGTLink)'s ImagerServer example
[...]

PR #7292, opened 2023-10-19
BUG: Allow reslicing of 32-bit integer images

ClaudioHoffmann:
Follow-up to #4982. 32-bit integer data types can fit losslessly into the 53-bit mantissa of a 64-bit `double`, and this fix has been working fine for us for the last few months.


lassoan (approved):
This change is correct, the check was stricter than necessary.

VTK indeed does not support 64-bit integers in many image filters (search for `#undef VTK_USE_INT64` in VTK source code to see all those classes). However, VTK_LONG and VTK_UNSIGNED_LONG refer to 32-bit integer types, which VTKsupports, so Slicer does not need to reject them.

PR #7299, opened 2023-10-23
BUG: Ensure DICOMCommand.start() returns process output 

jcfr:
Resolve `py_nomainwindow_DCMTKPrivateDictTest` failure caused by a regression introduced in commit adb997269 (`COMP: Refactor DICOMProcess and DICOMCommand`).

The issue stemmed from the refactoring commit, which guaranteed that the `start()` function of the base class was consistently invoked. This resulted in the connection of the "onStateChanged()" signal and the immediate reading of standard output and error.

Subsequently, any further calls to `self.process.readAllStandardOutput()` were returning an empty string. This fix restores the expected behavior.

### Test results

_Command below executed on local build done on Ubuntu 20.04_

```
ctest -R "(py_nomainwindow_qSlicerDICOMModuleGenericTest|py_nomainwindow_qSlicerDICOMPatcherModuleGenericTest|py_DICOMPatcher|CreateDICOMSeriesTest|py_DICOMReaders|py_nomainwindow_DCMTKPrivateDictTest)"

Test project /path/to/Slicer-Release/Slicer-build
    Start 450: py_nomainwindow_qSlicerDICOMModuleGenericTest
1/6 Test #450: py_nomainwindow_qSlicerDICOMModuleGenericTest ..........   Passed    4.39 sec
    Start 451: py_nomainwindow_qSlicerDICOMPatcherModuleGenericTest
2/6 Test #451: py_nomainwindow_qSlicerDICOMPatcherModuleGenericTest ...   Passed    3.60 sec
    Start 452: py_DICOMPatcher
3/6 Test #452: py_DICOMPatcher ........................................   Passed    4.96 sec
    Start 523: CreateDICOMSeriesTest
4/6 Test #523: CreateDICOMSeriesTest ..................................   Passed    0.35 sec
    Start 599: py_nomainwindow_DCMTKPrivateDictTest
5/6 Test #599: py_nomainwindow_DCMTKPrivateDictTest ...................   Passed    2.26 sec
    Start 607: py_DICOMReaders
6/6 Test #607: py_DICOMReaders ........................................   Passed    8.00 sec

100% tests passed, 0 tests failed out of 6

```

### DIFF
--- a/Modules/Scripted/DICOMLib/DICOMProcesses.py
+++ b/Modules/Scripted/DICOMLib/DICOMProcesses.py
@@ -75,6 +75,8 @@ class DICOMProcess:
         self.connections = {}
         self.exeDir = self.getDCMTKToolsPath()
         self.exeExtension = ".exe" if os.name == "nt" else ""
+        self.stdout = None
+        self.stderr = None
 
     def __del__(self):
         self.stop()
@@ -97,14 +99,20 @@ class DICOMProcess:
         """Callback to indicate that the process state has changed
         and is now either starting, running, or not running."""
         logging.debug(f"Process {self.cmd} now in state {self.PROCESS_STATE_NAMES[newState]}")
+        stdout = None
+        stderr = None
         if newState == 0 and self.process:
             stdout = self.process.readAllStandardOutput()
             stderr = self.process.readAllStandardError()
-            logging.debug('DICOM process error code is: %d' % self.process.error())
-            logging.debug('DICOM process standard out is: %s' % stdout)
-            logging.debug('DICOM process standard error is: %s' % stderr)
-            return stdout, stderr
-        return None, None
+
+            logging.debug(f"DICOM process exit status is: {self.process.exitStatus()}")
+            logging.debug(f"DICOM process exit code is: {self.process.exitCode()}")
+            logging.debug(f"DICOM process error is: {self.process.error()}")
+            logging.debug(f"DICOM process standard out is: {stdout}")
+            logging.debug(f"DICOM process standard error is: {stderr}")
+        self.stdout = stdout
+        self.stderr = stderr
+        return self.stdout, self.stderr
 
     def stop(self) -> None:
         """Stop the running standalone process."""
@@ -138,16 +146,8 @@ class DICOMCommand(DICOMProcess):
         logging.debug(("DICOM process running: ", self.executable, self.args))
         self.process.waitForFinished()
         if self.process.exitStatus() == qt.QProcess.CrashExit or self.process.exitCode() != 0:
-            stdout = self.process.readAllStandardOutput()
-            stderr = self.process.readAllStandardError()
-            logging.debug('DICOM process exit status is: %d' % self.process.exitStatus())
-            logging.debug('DICOM process exit code is: %d' % self.process.exitCode())
-            logging.debug('DICOM process error is: %d' % self.process.error())
-            logging.debug('DICOM process standard out is: %s' % stdout)
-            logging.debug('DICOM process standard error is: %s' % stderr)
             raise UserWarning(f"Could not run {self.executable} with {self.args}")
-        stdout = self.process.readAllStandardOutput()
-        return stdout
+        return self.stdout
 
 
 class DICOMStoreSCPProcess(DICOMProcess):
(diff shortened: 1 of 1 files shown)


Modules/Scripted/DICOMLib/DICOMProcesses.py
```diff
@@ -97,14 +99,39 @@ def onStateChanged(self, newState: int):
...
+                    qt.QProcess.ReadError: "ReadError",
+                    qt.QProcess.UnknownError: "UnknownError",
+                }.get(self.process.error())
+                logging.debug(f"DICOM process error is: {processError}")
+
+            logging.debug(f"DICOM process standard out is: {stdout}")
+            logging.debug(f"DICOM process standard error is: {stderr}")
+
+        self.stdout = stdout
+        self.stderr = stderr
```
tbirdso, line 133: If we prefer to persist `self.stdout` and `self.stderr` rather than requiring subscription to `onStateChanged`, then it would be more predictable to append and persist all output over the entire process lifetime rather than discarding prior output on async state change.
```suggestion
        self.stdout += str(stdout)
        self.stderr += str(stderr)
```

tbirdso (commented):
Thanks for tracking this down @jcfr .

Modules/Scripted/DICOMLib/DICOMProcesses.py
jcfr: Thanks for the review & suggestion :pray:

The goal of this change is to have the `stdout` and `stderr` variables capture the output of `readAllStandardOutput` and `readAllStandardError` once the process has completed its execution. The subscription to `onStateChanged` is handled internally and is triggered when the process transitions from the `Running` (or `Starting`) state to `NotRunning`. This ensures that these variables will contain the complete output of the process.

Updating these variables by appending to them across multiple calls to `start()` could result in a growing buffer and create complexity at the application level, as it would become challenging to distinguish between the outputs of different runs.

Moreover, considering that there is already an existing API, specific to `DICOMStoreSCPProcess`, for reading continuous output while the process is running (e.g., using `DICOMStoreSCPProcess.readFromStandardOutput()`), further refactoring to support continuous reading in the base class would provide a cleaner and more comprehensive solution for all scenarios.

tbirdso: Thanks for the fast response.

> The subscription to onStateChanged is handled internally and is triggered when the process transitions from the Running (or Starting) state to NotRunning. This ensures that these variables will contain the complete output of the process.

There are two clear scenarios in which this does _not_ ensure complete output:
1. If there were printouts to `stdout` or `stderr` during the `Starting` stage they will be read on transition to `Running` and then overwritten on transition to `NotRunning`. I do not know whether we actually expect output during `Starting` or not.
2. If some external user calls `process.readAllStandardOutput` or `process.readAllStandardError` at any point before the end of execution, the buffer will be wiped and only partial output will be available when the process ends. This is a much more likely scenario.

> Updating these variables by appending to them across multiple calls to start() could result in a growing buffer

It would be reasonable to clear the `self.stdout` and `self.stderr` buffers on `start` as long as that is documented somewhere in the `DICOMProcess` API.

> further refactoring to support continuous reading in the base class would provide a cleaner and more comprehensive solution for all scenarios.

I agree that further refactoring of `DICOMProcesses` classes would be ideal. There are design inconsistencies between `DICOMProcess` and `DICOMCommand` (among others) that predate https://github.com/Slicer/Slicer/commit/adb997269bacf3dfe15b597480391ce8e880d12e and #7299 and which I consider to be the root cause of this issue.

If we want to enforce consistent behavior in making `stdout` and `stderr` available from `DICOMProcess`, we would need to prevent the user from having direct access to the underlying QProcess so that they cannot clear `stdout` or `stderr` buffers and instead force them to call through intermediate `readStdout` and `readStderr` methods that persist buffer data. Alternatively, we could maintain behavior that avoids persisting `stdout` and `stderr` altogether and make the user responsible for managing the buffers.

If the goal is a quick fix to resolve regression failures then the code in this PR is likely functional, but IMO it introduces inconsistencies in `DICOMProcess` and `DICOMCommand` behavior that should be avoided.

pieper (approved):
Looks good to me - if you guys have resolved your discussion points it looks good to merge.

Modules/Scripted/DICOMLib/DICOMProcesses.py
jcfr: To capture the feedback, a dedicated issue has been created:
* https://github.com/Slicer/Slicer/issues/7301

> If there were printouts to stdout or stderr during the Starting stage they will be read on transition to Running

Indeed, that said these would generally be message that are operating system specific and related to failure to load libraries and like. I didn't consider handling that specific scenario.

> external user calls

Ditto. I will revisit the pull request.

pieper (approved):
Sounds good to me 👍

tbirdso (approved):
@jcfr Thanks for making these changes to preserve API compatibility following our offline discussion. 👍

PR #7302, opened 2023-10-25
Add label dilation filter + small fix

lassoan:
This PR contains a new label dilation filter, which is faster and more accurate than the previously used masked median filter.
The PR also contain a small unrelated fix for an error message when drag-and-dropping nodes.

### DIFF
--- a/Modules/Loadable/Segmentations/Logic/CMakeLists.txt
+++ b/Modules/Loadable/Segmentations/Logic/CMakeLists.txt
@@ -17,6 +17,8 @@ set(${KIT}_SRCS
   vtkSlicerSegmentationGeometryLogic.h
   vtkImageGrowCutSegment.cxx
   vtkImageGrowCutSegment.h
+  vtkImageLabelDilate3D.cxx
+  vtkImageLabelDilate3D.h
   FibHeap.cxx
   )
 
--- a/Modules/Loadable/Segmentations/Logic/vtkImageLabelDilate3D.h
+++ b/Modules/Loadable/Segmentations/Logic/vtkImageLabelDilate3D.h
@@ -1,0 +1,74 @@
+/*=========================================================================
+
+  Program:   Visualization Toolkit
+  Module:    vtkImageLabelDilate3D.h
+
+  Copyright (c) Ken Martin, Will Schroeder, Bill Lorensen
+  All rights reserved.
+  See Copyright.txt or http://www.kitware.com/Copyright.htm for details.
+
+     This software is distributed WITHOUT ANY WARRANTY; without even
+     the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+     PURPOSE.  See the above copyright notice for more information.
+
+=========================================================================*/
+/**
+ * @class   vtkImageLabelDilate3D
+ * @brief   Median Filter
+ *
+ * vtkImageLabelDilate3D dilates a labelmap by replacing each background voxel with the
+ * most dominant label voxel in its neighborhood.
+ */
+
+#ifndef vtkImageLabelDilate3D_h
+#define vtkImageLabelDilate3D_h
+
+#include "vtkImageSpatialAlgorithm.h"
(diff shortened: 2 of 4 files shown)


Modules/Loadable/Segmentations/Logic/vtkImageLabelDilate3D.h
```diff
@@ -0,0 +1,74 @@
...
+   * default middle of the neighborhood
+   */
+  void SetKernelSize(int size0, int size1, int size2);
+
+  ///@{
+  /**
+   * Return the number of elements in the median mask
+   */
+  vtkGetMacro(NumberOfElements, int);
+  ///@}
```
pieper, line 47: Not really clear what this refers to without reading the code. Maybe some extra text about why one would want to know this could help.

pieper (approved):
Looks good, just some minor comments.

Not sure why the ModuleInfo.py commit is in this PR.  I guess we should do a rebase and not a squash to keep the history easier to parse.

Modules/Scripted/ExtensionWizard/ExtensionWizardLib/ModuleInfo.py
```diff
@@ -66,10 +69,17 @@ def findModules(path, depth):
...
+                            # Its base class is ScriptedLoadableModule
+                            result.append(ModuleInfo(entry))
             except:
                 # Error while processing the file (e.g., syntax error),
                 # it cannot be a Slicer module.
                 pass
 
+            # We could also detect scripted CLI modules (e.g., by checking for presence of a suitable
+            # module descriptor XML file), but this module type is quite rare, so it was not seem
+            # worth the implementation effort.
```
jcfr, line 83:
```suggestion
            # We have the option to identify scripted CLI modules, such as by examining the existence of a
            # compatible module descriptor XML file. However, this type of module is relatively uncommon, so
            # the decision was made not to invest in implementing this feature.
```

Modules/Loadable/Segmentations/Logic/vtkImageLabelDilate3D.h
lassoan: This accessor was there in the median filter that this filter was adapted from. The line was added by Bill Lorensen in 1998 when he changed the `NumberOfElements` member variable from public to protected. While it does not do any harm, it is just noise, I'll remove it.

Modules/Loadable/Segmentations/Logic/CMakeLists.txt
```diff
@@ -17,6 +17,8 @@ set(${KIT}_SRCS
   vtkSlicerSegmentationGeometryLogic.h
   vtkImageGrowCutSegment.cxx
   vtkImageGrowCutSegment.h
+  vtkImageLabelDilate3D.cxx
+  vtkImageLabelDilate3D.h
```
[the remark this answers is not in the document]
lassoan: OK, I've added this to vtkAddon - https://github.com/Slicer/vtkAddon/pull/41

lassoan:
Thanks for the reviews @jcfr and @pieper.

> Not sure why the ModuleInfo.py commit is in this PR. I guess we should do a rebase and not a squash to keep the history easier to parse.

I just did not want to add a separate PR for a small fix (extra notifications, automatic test runs, etc.). If it's confusing then I'll submit separate PR for each unrelated commits (even if they are just some really small things).

pieper:
> If it's confusing then I'll submit separate PR for each unrelated commits (even if they are just some really small things).

I'm never sure what the cutoff should be.  I know Jc likes to have modular commits so it's easier to summarize changes.  I think if we just do rebase and not squash it will make sense.

lassoan:
Closing this PR, as the commits moved to separate pull requests:
- https://github.com/Slicer/Slicer/pull/7303
- https://github.com/Slicer/vtkAddon/pull/41

jcfr:
Having dedicated pull requests will help write the release notes, we can then get started leveraging this feature:

image


... and generate text like the following:

| |
|--|
| image |
| ... |
| image |

PR #7303, opened 2023-10-25
BUG: Fix error message when drag-and-dropping nodes

lassoan:
Python file detection plugin (that detects if Python files are drag-and-dropped to the application window) logged an error when MRML nodes were drag-and-dropped within the application.

### DIFF
--- a/Modules/Scripted/ExtensionWizard/ExtensionWizardLib/ModuleInfo.py
+++ b/Modules/Scripted/ExtensionWizard/ExtensionWizardLib/ModuleInfo.py
@@ -37,12 +37,15 @@ class ModuleInfo:
         result = []
         if os.path.isfile(path):
             entries = [path]
-        else:
+        elif os.path.isdir(path):
             entries = [os.path.join(path, entry) for entry in os.listdir(path)]
             # If the folder contains __init__.py, it means that this folder
             # is not a Slicer module but an embedded Python library that a module will load.
             if any(entry.endswith('__init__.py') for entry in entries):
                 entries = []
+        else:
+            # not a file or folder
+            return result
 
         if depth > 0:
             for entry in filter(os.path.isdir, entries):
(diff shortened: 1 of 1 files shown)


jcfr:
Should the commit be updated from:

```
BUG: Fix error message when drag-and-dropping nodes
```

to 

```
BUG: Fix error message when drag-and-dropping scripts
```

?

jcfr:
To provide additional context, the handled implemented by the extension wizard in `ExtensionWizardFileDialog.isMimeDataAccepted` was accepting the event before the one implemented in `qSlicerDataDialog::isMimeDataAccepted` could.

jcfr (approved):

lassoan:
> Should the commit be updated from:
> 
> ```
> BUG: Fix error message when drag-and-dropping nodes
> ```
> 
> to
> 
> ```
> BUG: Fix error message when drag-and-dropping scripts
> ```
> 
> ?

The message was logged when _nodes_ were drag-and-dropped (because the URL was empty), so the commit comment title is correct.

lassoan:
> To provide additional context, the handled implemented by the extension wizard in `ExtensionWizardFileDialog.isMimeDataAccepted` was accepting the event before the one implemented in `qSlicerDataDialog::isMimeDataAccepted` could

All FileDialogs gets a chance. There are two phases: during drag, we just check if there is _any_ FileDialog that can accepted it and if there is one then we stop and return that yes, Slicer can accept a drop. If the user actually drops the item on the Slicer window then _all_ the FileDialogs are queried and if there were more than one that could accept it, Slicer shows a popup window where the user can choose which one to use.

The only thing that is not nice is that in the dialog where the user can choose from the various interpretations, the results are not sorted based on confidence (because isMimeDataAccepted just returns a Boolean, so we don't know which FileReader is most likely the best one; it should return a confidence value instead).

Issue #7310, opened 2023-10-28
Improve Event Delegation in qMRMLThreeDView and qMRMLSliceView


jcfr:
## Is your feature request related to a problem? Please describe.

The current event delegation mechanism in Slicer is based on Slicer-specific interactor styles, namely `vtkMRMLThreeDViewInteractorStyle` and `vtkMRMLSliceViewInteractorStyle`, which are derived from `vtkMRMLViewInteractorStyle` and, ultimately, `vtkInteractorStyle3D`.

These styles handle interaction events by adding observers for events like `EnterEvent`, `LeaveEvent`, and `MouseMoveEvent`, and subsequently call functions like `vtkMRMLViewInteractorStyle::OnEnter()`, `vtkMRMLViewInteractorStyle::OnLeave()`, or `vtkMRMLViewInteractorStyle::OnMouseMove()`.

In each of these `On<EventName>()` functions, delegation to displayable managers and their associated widgets is performed, and if a given event isn't intercepted, the default behavior for `<EventName>` implemented in `vtkInteractorStyle3D` is executed by calling `this->Superclass::On<EventName>()`. The snippet below illustrates this:

  ```cpp
  if (!this->DelegateInteractionEventToDisplayableManagers(vtkCommand::<EventName>Event))
    {
    return;
    }
  this->Superclass::On<EventName>();
  ```

However, a limitation arises when default event handling implemented in specific interactor styles (e.g., `vtkOpenXRInteractorStyle` or `vtkOpenVRInteractorStyle`) needs to be executed when not delegated.

#### Refactor Interactor Styles

Refactor `vtkMRMLThreeDViewInteractorStyle` and `vtkMRMLSliceViewInteractorStyle` to derive from `vtkObject` and implement event delegation to MRML displayable managers and associated widgets by observing the interactor style. Depending on the observed event, default interactor style behavior would be executed using one of the following approaches:

1. **Setting Abort Flag**: For events such as `Move3D`, `Button3D`, `Menu3D`, and others where the `vtkInteractorStyle` class checks if the event was aborted, we will set an abort flag.
2. **Explicit Call of `On<EventName>()`**: For other events, we would explicitly call the corresponding `On<EventName>()` function. To support this, the existing method `vtkMRMLAbstractThreeDViewDisplayableManager::PassThroughInteractorStyleEvent` will be used and updated as needed.

The interactor style will be associated with its corresponding displayable manager group, allowing access to the MRML scene and registered displayable managers.

[...]

PR #7311, opened 2023-10-30
Improve Event Delegation in qMRMLThreeDView and qMRMLSliceView

jcfr:
This pull request is aimed at streamlining the review process and testing by consolidating several independent changes, each of which will be contributed separately.

The primary focus for review in this pull request is the last commit, titled `ENH: Improve Event Delegation in qMRMLThreeDView and qMRMLSliceView`.

In this commit, we've refactored `vtkMRMLThreeDViewInteractorStyle` and `vtkMRMLSliceViewInteractorStyle` to inherit from `vtkObject` and have implemented event delegation to MRML displayable managers and their associated widgets by observing the interactor style.

This enhancement allows for the association of specialized interactor styles, such as `vtkOpenXRInteractorStyle` or `vtkOpenVRInteractorStyle`, with the render window while still maintaining MRML-specific event delegation.

### Dependent Changes

The introduction of the commit `ENH: Allow virtual func to be called during MRML(ThreeD|Slice)View pimpl init` necessitates an update to CTK, which can be found in the following pull request:
* https://github.com/commontk/CTK/pull/1153

**Update:** CTK pull request https://github.com/commontk/CTK/pull/1153 has been integrated upstream and corresponding `External_CTK.cmake` update has been added to this pull request by integrating the commits being contributed in the following pull request:
* https://github.com/Slicer/Slicer/pull/7325

Associated independent changes are being contributed in these pull requests:
* https://github.com/Slicer/Slicer/pull/7315
* https://github.com/Slicer/Slicer/pull/7316
* https://github.com/Slicer/Slicer/pull/7318
* https://github.com/Slicer/Slicer/pull/7319
* https://github.com/Slicer/Slicer/pull/7320
* https://github.com/Slicer/Slicer/pull/7321
* https://github.com/Slicer/Slicer/pull/7322 
* https://github.com/Slicer/Slicer/pull/7323 
* https://github.com/Slicer/Slicer/pull/7324
* https://github.com/Slicer/Slicer/pull/7325

### DIFF
--- a/Libs/MRML/DisplayableManager/vtkMRMLThreeDViewInteractorStyle.cxx
+++ b/Libs/MRML/DisplayableManager/vtkMRMLThreeDViewInteractorStyle.cxx
@@ -27,6 +27,7 @@
 #include <vtkCamera.h>
 #include <vtkCellPicker.h>
 #include <vtkEvent.h>
+#include <vtkInteractorStyle.h>
 #include <vtkMath.h>
 #include <vtkObjectFactory.h>
 #include <vtkPoints.h>
@@ -110,7 +111,8 @@ bool vtkMRMLThreeDViewInteractorStyle::DelegateInteractionEventToDisplayableMana
   // Get display and world position
   int* displayPositionInt = this->GetInteractor()->GetEventPosition();
   vtkRenderer* pokedRenderer = this->GetInteractor()->FindPokedRenderer(displayPositionInt[0], displayPositionInt[1]);
-  this->SetCurrentRenderer(pokedRenderer);
+  vtkInteractorStyle* interactorStyle = vtkInteractorStyle::SafeDownCast(this->GetInteractor()->GetInteractorStyle());
+  interactorStyle->SetCurrentRenderer(pokedRenderer);
   if (!pokedRenderer || !inputEventData)
     {
     // can happen during application shutdown
@@ -165,7 +167,8 @@ void vtkMRMLThreeDViewInteractorStyle::SetInteractor(vtkRenderWindowInteractor *
 bool vtkMRMLThreeDViewInteractorStyle::QuickPick(int x, int y, double pickPoint[3])
 {
   vtkRenderer* pokedRenderer = this->GetInteractor()->FindPokedRenderer(x, y);
-  this->SetCurrentRenderer(pokedRenderer);
+  vtkInteractorStyle* interactorStyle = vtkInteractorStyle::SafeDownCast(this->GetInteractor()->GetInteractorStyle());
+  interactorStyle->SetCurrentRenderer(pokedRenderer);
   if (pokedRenderer == nullptr)
   {
     vtkDebugMacro("Pick: couldn't find the poked renderer at event position " << x << ", " << y);
--- a/Libs/MRML/DisplayableManager/vtkMRMLViewInteractorStyle.cxx
+++ b/Libs/MRML/DisplayableManager/vtkMRMLViewInteractorStyle.cxx
@@ -25,6 +25,7 @@
 
 // VTK includes
 #include "vtkCallbackCommand.h"
+#include "vtkInteractorStyle.h"
 #include "vtkObjectFactory.h"
 #include "vtkRenderer.h"
 #include "vtkRenderWindow.h"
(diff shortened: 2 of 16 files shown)


Libs/MRML/DisplayableManager/vtkMRMLViewInteractorStyle.cxx
```diff
@@ -71,31 +76,19 @@ void vtkMRMLViewInteractorStyle::PrintSelf(ostream& os, vtkIndent indent)
 //----------------------------------------------------------------------------
 void vtkMRMLViewInteractorStyle::OnKeyPress()
 {
-  if (this->DelegateInteractionEventToDisplayableManagers(vtkCommand::KeyPressEvent))
-    {
-    return;
-    }
-  this->Superclass::OnKeyPress();
+  this->GetInteractorStyle()->OnKeyPress();
```
lassoan, line 79: If all we do is calling the Superclass implementation then it is better to remove the method completely.

Libs/MRML/DisplayableManager/vtkMRMLViewInteractorStyle.h
```diff
@@ -37,43 +40,44 @@ class vtkTimerLog;
 /// Some additional high-level events (such as click and double-click)
 /// are generated here.
 class VTK_MRML_DISPLAYABLEMANAGER_EXPORT vtkMRMLViewInteractorStyle :
-  public vtkInteractorStyle3D
+  public vtkObject
```
lassoan, line 43: Why it is not based on an interactor style?

Libs/MRML/DisplayableManager/vtkMRMLThreeDViewInteractorStyle.cxx
```diff
@@ -110,7 +111,8 @@ bool vtkMRMLThreeDViewInteractorStyle::DelegateInteractionEventToDisplayableMana
   // Get display and world position
   int* displayPositionInt = this->GetInteractor()->GetEventPosition();
   vtkRenderer* pokedRenderer = this->GetInteractor()->FindPokedRenderer(displayPositionInt[0], displayPositionInt[1]);
```
lassoan, line 113: `FindPokedRenderer` is error-prone. If somebody adds any content to the render window in a new renderer then it may break picking and or interaction with other widgets. It would be safer to pass the main 3D content renderer, not whatever renderer happened to be found at that position.

lassoan (commented):
I don’t like going back to using interactor style, as it is just an unnecessary additional layer that adds complexity and not sufficient to actually handle all interactions in complex applications. Interactor styles should be rather removed from VTK, too (and let VTK widgets handle camera movement).

However, I cannot pinpoint any functional errors with the changes and I see that you worked a lot on this and would not want to hold up things that depend on these changes, so I'm OK if the changes are merged.

Libs/MRML/DisplayableManager/vtkMRMLViewInteractorStyle.h
jcfr: Because our `vtkMRMLViewInteractorStyle` are technically observer and not interactor style anymore. For more details, see section `Refactor Interactor Styles` in https://github.com/Slicer/Slicer/issues/7310

For sake of reducing the amount of changes and allowing easy review, I didn't rename our classes. For more details, see section `Renaming Interactor Styles` in https://github.com/Slicer/Slicer/issues/7310.

Libs/MRML/DisplayableManager/vtkMRMLThreeDViewInteractorStyle.cxx
jcfr: Agreed, this can be improved and I will create a dedicated issue to improve and address this. In prefer to keep the existing behavior and focus this pull request on only revisiting the observation of the interactor style events.

Libs/MRML/DisplayableManager/vtkMRMLViewInteractorStyle.cxx
jcfr: This is intended to keep the same behavior as we had before, I will create a follow-up pull request and/or issue to move forward with simplification

jcfr:
> going back to using interactor style, as it is just an unnecessary additional layer that adds complexity

As a point of clarification, this pull-request does not add any functional changes.
Instead of having:
* `vtkMRMLViewInteractorStyle` derives from `vtkInteractorStyle3D`, it now derives from `vtkObject`
* `vtkMRMLViewInteractorStyle` override callback, it calls corresponding functions by observing the interactor style associated with the render window
* the `qMRMLThreeDView` class set `vtkMRMLThreeDViewInteractorStyle` as interactor style, it sets an instance of `vtkInteractorStyle3D`
* the `qMRMLSliceView` class set `vtkMRMLSliceViewInteractorStyle` as interactor style, it sets an instance of `vtkInteractorStyleUser`

Doing so allows to associated the `OpenVR` or `OpenXR` rendering windows with their corresponding interactor style while maintaining the delegation of event to our displayable manager working as expected.

lassoan (approved):
Thanks for the responses, I have no further questions.

PR #7314, opened 2023-10-30
Improve documentation of obsolete filtering API in SH proxy model

cpinter:
Re https://discourse.slicer.org/t/problems-filtering-subject-hierarchy-tree/32435/5

### DIFF
--- a/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSortFilterSubjectHierarchyProxyModel.cxx
+++ b/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSortFilterSubjectHierarchyProxyModel.cxx
@@ -485,6 +485,8 @@ void qMRMLSortFilterSubjectHierarchyProxyModel::setAttributeNameFilter(QString f
     this->addItemAttributeFilter(filter);
     }
   this->invalidateFilter();
+  qWarning() << "qMRMLSortFilterSubjectHierarchyProxyModel::setAttributeNameFilter is deprecated, "
+                "use addItemAttributeFilter or removeItemAttributeFilter instead";
 }
 
 //-----------------------------------------------------------------------------
@@ -510,6 +512,8 @@ void qMRMLSortFilterSubjectHierarchyProxyModel::setAttributeValueFilter(QString 
 
   d->ItemAttributeFilters[0].AttributeValue = filter;
   this->invalidateFilter();
+  qWarning() << "qMRMLSortFilterSubjectHierarchyProxyModel::setAttributeValueFilter is deprecated,"
+                " use addItemAttributeFilter or removeItemAttributeFilter instead";
 }
 
 //-----------------------------------------------------------------------------
--- a/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSortFilterSubjectHierarchyProxyModel.h
+++ b/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSortFilterSubjectHierarchyProxyModel.h
@@ -85,13 +85,14 @@ class Q_SLICER_MODULE_SUBJECTHIERARCHY_WIDGETS_EXPORT qMRMLSortFilterSubjectHier
   /// Overrides \sa includeNodeAttributeNamesFilter
   Q_PROPERTY(QStringList excludeNodeAttributeNamesFilter READ excludeNodeAttributeNamesFilter WRITE setExcludeNodeAttributeNamesFilter)
 
-  /// Filter to show only items that contain an attribute with this name. Empty by default
-  /// Note: Deprecated, kept only for backwards compatibility. Sets and returns the first attribute in \sa includeNodeAttributeNamesFilter
+  /// Filter to show only items that contain an item attribute with this name. Empty by default.
+  /// Sets and returns the first attribute in \sa includeItemAttributeNamesFilter.
+  /// \deprecated Kept only for backwards compatibility. Use addItemAttributeFilter() or removeItemAttributeFilter() instead.
   Q_PROPERTY(QString attributeNameFilter READ attributeNameFilter WRITE setAttributeNameFilter)
-  /// Filter to show only items that contain any attribute given in \sa includeItemAttributeNamesFilter with the value.
+  /// Filter to show only items that contain any item attribute given in \sa includeItemAttributeNamesFilter with the value.
   /// If empty, then existence of the attributes is enough to show.
-  /// Exact match is required. Empty by default
-  /// Note: Deprecated, kept only for backwards compatibility. Works consistently with the previous operation.
+  /// Exact match is required. Empty by default.
+  /// \deprecated Kept only for backwards compatibility. Use addItemAttributeFilter() or removeItemAttributeFilter() instead.
   Q_PROPERTY(QString attributeValueFilter READ attributeValueFilter WRITE setAttributeValueFilter)
 
 public:
--- a/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyComboBox.h
+++ b/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyComboBox.h
@@ -86,13 +86,14 @@ class Q_SLICER_MODULE_SUBJECTHIERARCHY_WIDGETS_EXPORT qMRMLSubjectHierarchyCombo
   /// Overrides \sa includeNodeAttributeNamesFilter
   Q_PROPERTY(QStringList excludeNodeAttributeNamesFilter READ excludeNodeAttributeNamesFilter WRITE setExcludeNodeAttributeNamesFilter)
 
-  /// Filter to show only items that contain an attribute with this name. Empty by default
-  /// Note: Deprecated, kept only for backwards compatibility. Sets and returns the first attribute in \sa includeNodeAttributeNamesFilter
+  /// Filter to show only items that contain an item attribute with this name. Empty by default.
+  /// Sets and returns the first attribute in \sa includeItemAttributeNamesFilter.
+  /// \deprecated Kept only for backwards compatibility. Use addItemAttributeFilter() or removeItemAttributeFilter() instead.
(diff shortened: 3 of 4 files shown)


Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSortFilterSubjectHierarchyProxyModel.h
```diff
@@ -85,12 +85,12 @@ class Q_SLICER_MODULE_SUBJECTHIERARCHY_WIDGETS_EXPORT qMRMLSortFilterSubjectHier
   /// Overrides \sa includeNodeAttributeNamesFilter
   Q_PROPERTY(QStringList excludeNodeAttributeNamesFilter READ excludeNodeAttributeNamesFilter WRITE setExcludeNodeAttributeNamesFilter)
 
-  /// Filter to show only items that contain an attribute with this name. Empty by default
-  /// Note: Deprecated, kept only for backwards compatibility. Sets and returns the first attribute in \sa includeNodeAttributeNamesFilter
+  /// Filter to show only items that contain an item attribute with this name. Empty by default.
+  /// Note: Deprecated, kept only for backwards compatibility. Sets and returns the first attribute in \sa includeItemAttributeNamesFilter
```
jcfr, line 89:
```suggestion
  /// \deprecated Kept only for backwards compatibility. Sets and returns the first attribute in \sa includeItemAttributeNamesFilter
```

Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSortFilterSubjectHierarchyProxyModel.h
```diff
@@ -85,12 +85,12 @@ class Q_SLICER_MODULE_SUBJECTHIERARCHY_WIDGETS_EXPORT qMRMLSortFilterSubjectHier
...
   Q_PROPERTY(QString attributeNameFilter READ attributeNameFilter WRITE setAttributeNameFilter)
-  /// Filter to show only items that contain any attribute given in \sa includeItemAttributeNamesFilter with the value.
+  /// Filter to show only items that contain any item attribute given in \sa includeItemAttributeNamesFilter with the value.
   /// If empty, then existence of the attributes is enough to show.
-  /// Exact match is required. Empty by default
+  /// Exact match is required. Empty by default.
   /// Note: Deprecated, kept only for backwards compatibility. Works consistently with the previous operation.
```
jcfr, line 94:
```suggestion
  /// \deprecated Kept only for backwards compatibility. Works consistently with the previous operation.
```

Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSortFilterSubjectHierarchyProxyModel.h
cpinter: Ah, thanks! I did a search to find this (in Qt classes) but for some reason I didn't see any. I'll change it now.

jcfr:
To avoid warnings, could you also update `qMRMLSubjectHierarchyTreeView::removeAttributeFilter` to use the recommended API `removeItemAttributeFilter` ?

Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSortFilterSubjectHierarchyProxyModel.cxx
```diff
@@ -510,6 +511,7 @@ void qMRMLSortFilterSubjectHierarchyProxyModel::setAttributeValueFilter(QString
 
   d->ItemAttributeFilters[0].AttributeValue = filter;
   this->invalidateFilter();
+  qWarning() << "qMRMLSortFilterSubjectHierarchyProxyModel::setAttributeValueFilter is deprecated, use addItemAttributeFilter instead";
```
jcfr, line 514:
```suggestion
  qWarning() << "qMRMLSortFilterSubjectHierarchyProxyModel::setAttributeValueFilter is deprecated, use addItemAttributeFilter or removeItemAttributeFilter instead";
```

Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSortFilterSubjectHierarchyProxyModel.cxx
```diff
@@ -485,6 +485,7 @@ void qMRMLSortFilterSubjectHierarchyProxyModel::setAttributeNameFilter(QString f
     this->addItemAttributeFilter(filter);
     }
   this->invalidateFilter();
+  qWarning() << "qMRMLSortFilterSubjectHierarchyProxyModel::setAttributeNameFilter is deprecated, use addItemAttributeFilter instead";
```
jcfr, line 488:
```suggestion
  qWarning() << "qMRMLSortFilterSubjectHierarchyProxyModel::setAttributeNameFilter is deprecated, use addItemAttributeFilter or removeItemAttributeFilter instead";
```

cpinter:
> To avoid warnings, could you also update `qMRMLSubjectHierarchyTreeView::removeAttributeFilter` to use the recommended API `removeItemAttributeFilter` ?

I think it would be better to deprecate the `qMRMLSubjectHierarchyTreeView::removeAttributeFilter` function too. It is from the times when only one name/value item attribute filter was available, and fixing it as suggested would introduce the confusion of having a function apparently for removing a certain attribute filter, but without argument, which removes all the include item attribute filters.

sjh26:
+10e6 for anything that helps clarify the usage of these functions

jcfr:
> I was still a bit confused about `addItemAttributeFilter/addNodeAttributeFilter`

The main idea is that node attributes and subject hierarchy item attributes are different.

See https://discourse.slicer.org/t/problems-filtering-subject-hierarchy-tree/32435/8?u=jcfr

jamesobutler:
The confusion is less about subject hierarchy item vs node, but more about why there is `addItemAttributeFilter` which takes in include/exclude details and there also being `includeItemAttributeNamesFilter`. Why are there 2 different ways trying to accomplish the same thing regarding filtering items (or nodes)? One seemingly supports filtering by name and value. The other only supports filtering by names. 

The documentation is a bit difficult to understand how you might use the two methods. Can I both use `addItemAttributeFilter` and `includeItemAttributeNamesFilter` in a meaningful way for the same tree view? Why doesn't `addItemAttributeFilter` just optionally require value to support the case of only filtering by name and therefore removing the need for `includeItemAttributeNamesFilter`?

jcfr:
> One seemingly supports filtering by name and value. The other only supports filtering by names.

Ditto. Having the different properties allows some flexibility when using Qt Designer.

cpinter:
I believe the PR fulfills the original intention so please approve.

Any further simplification can be discussed later. Concrete suggestions welcome.

Thank you!

jamesobutler:
Is `DOC` the appropriate commit prefix here? There appears to be functional code change. A previous integrated commit with new runtime deprecations was an `ENH` commit prefix (https://github.com/Slicer/Slicer/commit/73867d1c432842ee768ffb7693a4993070f5ff3c).

jamesobutler:
> Having the different properties allows to support setting up name based filtering using Qt Designer.

This is good information about why you might use one method over the other. So something like use `includeItemAttributeNamesFilter` if you would like to define things with Qt Designer. Otherwise you can use `addItemAttributeFilter`. It was unclear from the source why there is 2 ways to do the same thing. I would’ve otherwise issued code to remove the duplicate method to reduce confusion.

cpinter:
> Is DOC the appropriate commit prefix here?

I think it is. I actually explicitly checked here https://slicer.readthedocs.io/en/latest/developer_guide/contributing.html#how-to-write-commit-messages
and it says this:
```
DOC: Documentation change
ENH: New functionality
```

Since there is no new functionality, I chose DOC.

jamesobutler:
Interesting. I specifically read the same thing and interpreted it as not being `DOC` because it wasn’t changing a documentation file like what would be for the readthedocs and would be new functionality because it is a new feature to provide a runtime notification of deprecated API usage.

cpinter:
> changing a documentation file

There is documentation all over the code, in every single file. Docstrincs, doxygen comments, etcetc. No need to change an rst to change documentation.

So you consider this a new feature then?

jamesobutler:
Docstrings, doxygen comments, markdown files for readthedocs - changes to that would all seem to fall under `DOC` change where like `STYLE` it has no impact on functional code. Since this adds changes to code that gets executed it would seem that adding of a new runtime deprecation would be an `ENH` type commit. Where in general a different set of conditions may be applied to `DOC` commits versus `ENH` commits when cherry-picking commits for backport activities.

cpinter:
This was the point that made me think too, that there are runtime consequences. And this is why I looked up the documentation, but there it says "new functionality", and this is not functionality. We can argue about this all day but this is falling down my priority list. @jcfr can you please merge or tell me to amend the commit? Thanks!

jamesobutler:
I'm trying to use this as a learning experience to know what is the appropriate commit message because evidently I did it wrong in the past. Apologies if I'm wasting your time.

cpinter:
I think everyone has done it "wrong" in the past, as the borders between these classes are quite fuzzy, and even the most seniors will say a different thing a few years apart for corner cases like this. I'm curious of what @jcfr or @lassoan would say in this specific case (and let us merge :) ), but I also think that its impact is minimal. Have a great weekend!

lassoan:
Based on my commit history, I tend to choose "ENH", so I would probably choose it this time, too. However, I agree that the lines are blurred and this is already a borderline case. So, I could accept anything:
- ENH: because we are clarifying documentation and making it easier to detect deprecated API usage
- BUG: because we are fixing things (adding documentation and warning messages) that should have been done earlier
- DOC: because the program function practically does not change, we just improve code comments and log messages
- STYLE: because there is almost no functional change
- COMP: because commits related to deprecations are often marked like this

jcfr:
I suggest to use `ENH` as it improves the overall user experience with the user being someone interacting with the Python Console.

jcfr:
cc: @jamesobutler

jcfr (approved):

PR #7315, opened 2023-10-30
ENH: Update MRML interactor style to explicitly use the poked renderer

jcfr:
Instead of relying on the CurrentRenderer ivar internally updated by the "vtkInteractorStyle::FindPokedRenderer()", we explicitly retrieve and use the poked renderer with "vtkRenderWindowInteractor::FindPokedRenderer()"


pieper (approved):
It's not clear to me what is expected to behave differently after this change, but it seems logical.

jcfr:
Thanks for the review. This is helpful with upcoming change related to the refactoring of the event delegation.

lassoan:
There are several renderers drawing into the same render window area: the one that draws the 3D scene and those that display the ruler, orientation marker, markup labels, etc. For most cases, we should probably use the renderer that is explicitly set in the interactor style, not the one that is found by FindPokedRenderer.

Check out implementation of vtkRenderWindowInteractor::FindPokedRenderer. It just returns the first "interactive" renderer that draws in that region. Are you sure that it is the correct one?

What was the problem with the current behavior? In what cases you observed that the poked renderer was different from the current renderer? In those cases, was it correct to use the poked renderer?

Please test the new implementation very thoroughly when you operate in areas over the orientation marker, ruler, and annotation markup labels and interaction handles.

cpinter:
@jcfr These are very good points by @lassoan above; just pinging in case it went unnoticed.

PR #7316, opened 2023-10-30
ENH: Update qMRMLThreeDView to retrieve camera node from displayable manager

jcfr:


lassoan (approved):
The changes look good to me, but it would be even better to deprecate these methods, They should not be at this level. Why are they here (in the widget) instead of logic classes? If they are not used anywhere (Slicer core, CTK,..) then we should deprecate them and remove in a few years.

PR #7317, opened 2023-10-30
BUG: Update CTK to adjust the size of the delete DICOM object dialog

jcfr:
When deleting many patients (or studies or series) it can occur that the confirmation dialog that is created with all the deleted object names is taller than the screen height, so the buttons are cut off and not available.

This ensures that the said dialog cannot be higher than the DICOM browser.

List of CTK changes:

```
$ git shortlog 5657c58a7..b9ef9b8a --no-merges
Csaba Pinter (1):
      BUG: Make sure the delete DICOM object dialog is not too tall
```


cpinter (approved):

pieper:
For reference: https://github.com/commontk/CTK/commit/b9ef9b8aa1938676ce0ac105a3deb0e1bfc235a1

PR #7319, opened 2023-10-30
ENH: Remove unused API to set/get displayable manager interactor abort flags

jcfr:
The `InteractorStyleAbort` API was originally introduced in 8ded9a0bb (`ENH: eliminated conflict between crosshair navigation and window level`).

Then, it was superseded by the `InteractorAbort` API in 0b9b33983 (`BUG: #2629.  Window level controls conflicted with crosshair navigation`)

Ultimately, the last uses of either the `InteractorStyleAbort` or `InteractorAbort` API were removed in dc2c09580 (`ENH: Improved crosshair usability, added display in 3D view`)


lassoan (approved):
Looks good to me. These are no longer needed.

PR #7322, opened 2023-10-30
STYLE: Remove unused ivars from vtkMRMLViewInteractorStyle

jcfr:
Follow-up of 364ff7b6d (`ENH: Use common interactor style for slice and 3D views`) in which these variables became obsolete.


lassoan (approved):
Looks good to me.

PR #7323, opened 2023-10-30
ENH: Simplify vtkMRMLViewInteractorStyle::SetMouseCursor

jcfr:
Instead of relying on the "CurrentRenderer" indirectly set by "vtkInteractorStyle::FindPokedRenderer", directly retrieve the render window from the interactor.


lassoan (approved):
Looks good to me.

PR #7324, opened 2023-10-30
ENH: Set MRML interactor style camera node observing displayable manager

jcfr:


Libs/MRML/DisplayableManager/vtkMRMLThreeDViewInteractorStyle.cxx
```diff
@@ -61,6 +64,29 @@ void vtkMRMLThreeDViewInteractorStyle::PrintSelf(ostream& os, vtkIndent indent)
...
+    }
+
+  this->Superclass::SetDisplayableManagers(displayableManagers);
+
+  // Observe ActiveCameraChangedEvent
+  vtkMRMLAbstractDisplayableManager* cameraDisplayableManager =
+      this->DisplayableManagers->GetDisplayableManagerByClassName("vtkMRMLCameraDisplayableManager");
+  if (cameraDisplayableManager)
+    {
+    // \tbd active camera should be set to view node instead and only observing
```
lassoan, line 82: `tbd` usually means "to be defined", do you mean `\todo`? (I could not find `\tbd` doxygen tag - https://www.doxygen.nl/manual/commands.html)

lassoan, line 82: I would not say it is a TODO. It would make sense to describe why this is here (to detect when the camera displayable manager sets the camera) and maybe make a note on that a simpler alternative would be to directly store a reference to the camera in the view node. However, if we really want to simplify how the camera is mapped to the view (which would make sense) then we should rather add an issue to the bugtracker.

lassoan (commented):
This looks good to me. It is a somewhat more complex design but conceptually more clear. Just added some nitpicks inline.

Libs/MRML/DisplayableManager/vtkMRMLThreeDViewInteractorStyle.cxx
jcfr: I copied the comment from https://github.com/Slicer/Slicer/blob/9f95ec7deb1045167c59644a7634ecde5b60ac8a/Libs/MRML/DisplayableManager/vtkMRMLViewDisplayableManager.cxx#L652-L655

Using `\todo` should indeed better. I will update it at both location

jcfr: Agreed. I will also update the comment at https://github.com/Slicer/Slicer/blob/9f95ec7deb1045167c59644a7634ecde5b60ac8a/Libs/MRML/DisplayableManager/vtkMRMLViewDisplayableManager.cxx#L652-L655 and add an entry in the tracker.

jcfr:
@lassoan  I addressed your comment and also created an issue to document possible path forward to improve how we associated view and camera node:
* https://github.com/Slicer/Slicer/issues/7333

jcfr:
Waiting the following pull request is integrated, I cancelled the `CI (Build)` workflow:
* #7334 

Without it the `CI (Lint)` workflow is failing.

lassoan (approved):
It looks good to me, thank you!

PR #7325, opened 2023-10-30
ENH: Allow virtual func to be called during MRML(ThreeD|Slice)View pimpl init

jcfr:
The CTK update adds protected constructor to `ctkVTKRenderView` and `ctkVTKSliceView` for associating a derived pimpl.

This is required to have the complete virtual function chain executed when virtual function are called in the `qMRMLThreeDViewPrivate::init()` or `qMRMLSliceViewPrivate::init()` function.

### DIFF
--- a/Libs/MRML/Widgets/qMRMLSliceView.cxx
+++ b/Libs/MRML/Widgets/qMRMLSliceView.cxx
@@ -128,7 +128,7 @@ void qMRMLSliceViewPrivate::vtkInternalLightBoxRendererManagerProxy::SetLightBox
 
 //---------------------------------------------------------------------------
 qMRMLSliceViewPrivate::qMRMLSliceViewPrivate(qMRMLSliceView& object)
-  : q_ptr(&object)
+  : ctkVTKSliceViewPrivate(object)
 {
   this->DisplayableManagerGroup = nullptr;
   this->MRMLScene = nullptr;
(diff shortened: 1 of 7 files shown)


pieper (approved):
I'm not following closely but if this works for you I'm fine with it.

lassoan (approved):
Same here. Looks good overall.

jcfr:
This pull request updates CTK in the first commit and update the classes `qMRMLThreeDView` and `qMRMLSliceView` in a second commit, that said the check[^1] attempting to detect SuperBuild changes does not properly see that the first commit is updating `External_CTK.cmake`.

For this reason, I will bypass out usual check and move forward with the integration.

This can be see the build log[^2] associated with the `CI (build)` workflow (also copied below for future reference).

| CI (Build) | Complete list of updated files [^3] |
|--|--|
|image | image |

I think we should also look for updates our check with something like this:

```
      - name: 'Build Slicer'
        id: slicer-build
-        if: steps.changes.outputs.paths-to-include == 'true'
+        if: steps.changes.outputs.paths-to-include == 'true' and steps.changes-superbuild.outputs.paths-to-include == 'false'
        uses: ./.github/actions/slicer-build
```


cc: @jamesobutler 

[^1]: https://github.com/Slicer/Slicer/blob/09a0e50583f85ad80e09f8fbc2a0c115a1bb0f79/.github/workflows/ci.yml#L24-L38
[^2]: https://github.com/Slicer/Slicer/actions/runs/6698340032/job/18200163646?pr=7325#step:3:33
[^3]: https://github.com/Slicer/Slicer/pull/7325/files

PR #7326, opened 2023-10-31
ENH: Make Grow from seeds effect input requirements more clear

lassoan:
Grow from seeds effect requires at least 2 visible segments as input if no editable region is specified, and at least 1 visible segment if editable region is specified. When the requirement was not fulfilled then the user did not know what was wrong, because there was no visible notification on the GUI (only in the application log), see for example https://discourse.slicer.org/t/grow-from-seeds-does-not-work-if-painting-only-one-segment/32452/6

This commit adds display of the error in a popup window if there are not enough input segments. A notification on the status bar is added if segmentation is canceled due to removal of an input segment.


pieper (approved):
LGTM - glad that user's feedback was helpful.

lassoan:
> glad that user's feedback was helpful.

Yes. The user is always right (even when he makes a mistake).

PR #7330, opened 2023-11-01
ENH: Update DICOMUtils loadPatient API to handle messages and progressCallback

jcfr:
_Originally posted by @lassoan in https://github.com/Slicer/Slicer/pull/7270#pullrequestreview-1680877095_

> [...] I agree that it can be useful to expose `messages` and `progressCallback` arguments at higher-level functions, too. I would prefer to explicitly specify these arguments, as it would make the API more clear (and 2 extra arguments are not too many and I don't expect that `loadLoadables` will have many new options in the future).

cc: @marcus-wirtz-snkeos 

---

This pull request was created to include the changes originally included through this pull-requests:
* https://github.com/Slicer/Slicer/pull/7270

For the sake of having a clean history and considering the updated `main` branch has not been referenced externally, `main` was forced push to https://github.com/Slicer/Slicer/commit/b050d39e4f380ae5e3f0a12d77464b8978672fe6

### DIFF
--- a/Modules/Scripted/DICOMLib/DICOMUtils.py
+++ b/Modules/Scripted/DICOMLib/DICOMUtils.py
@@ -24,7 +24,7 @@ comment = """
 
 
 # ------------------------------------------------------------------------------
-def loadPatientByUID(patientUID):
+def loadPatientByUID(patientUID, messages=None, progressCallback=None):
     """ Load patient by patient UID from DICOM database.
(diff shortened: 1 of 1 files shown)


jcfr:
@lassoan @jamesobutler @pieper Since this change was already approved in the original pull request, review should be straightforward. 🙏

jamesobutler (approved):

PR #6418, opened 2022-06-08
BUG: Fix configuration error on macOS due to improper flow control

jamesobutler:
This should resolve the configuration error that happened on macOS https://slicer.cdash.org/viewBuildError.php?buildid=2709158.

The "endif" portion was not removed when it should have been as part of changes in https://github.com/Slicer/Slicer/commit/b9ac6ed61cae9de90986eaf8766c0e8b1b4461c1.

PR #7332, opened 2023-11-01
Trim trailing whitespaces and update pre-commit config to include "trailing-whitespace"

jcfr:
Consistently avoiding trailing whitespaces is helpful to avoid "false differences" and ensure all diffing tools show the "real" changes by default.

By adding `trailing-whitespace` to the pre-commit configuration, it enables the following:
* GitHub Action report: Suggested changes are reported in the error log associated with the pre-commit GitHub Action workflow.
* Consolidated local pre-commit checks and updates by running:

  ```

  ```

Once this is integrated we will add the commit `STYLE: Trim trailing whitespaces` to `.git-blame-ignore-revs`[^1] as it was one:
* https://github.com/Slicer/Slicer/pull/6498
* https://github.com/Slicer/Slicer/pull/6418


jamesobutler (approved):
My group uses this trailing whitespace check in our own work with success. Adding it here makes sense.

For additional reference we also use `- id: check-ast`

jcfr:
Thanks for the feedback. I will gradually introduce new checks.

PR #7338, opened 2023-11-03
ENH: Transition Python linting from flake8 to ruff

jcfr:
Ruff is 10-100x faster than existing linters (like Flake8) and has over 700 built-in rules, with native re-implementations of popular Flake8 plugins.

See https://beta.ruff.rs/docs/


lassoan:
These improvements all sound good, thanks for working on it. What I would really appreciate though if it was possible to:
- easily run the exact same checks in the local build tree without any extra configuration, maybe as part of the build process
- have an option for the the linter to change the source code to make it compliant

jcfr:
> easily run the exact same checks in the local build tree without any extra configuration, maybe as part of the build process

While it is not running as part of the build process, the `.ruff.toml` will be found and used by modern IDE (e.g vscode)

For running the tool locally, it is currently possible doing:
* installing pre-commit in a virtual env
* running `pre-commit run --all`

It should be possible to create a virtual env within the build tree all to install the tooling needed at build time, I wil create an issue,

> have an option for the the linter to change the source code to make it compliant

If possible fixed will be applied running the tool locally.

jcfr:
In the meantime, could it be approved ?

jamesobutler (approved):
The performance looks promising and the community seems to have responded in favor of it.

PR #7340, opened 2023-11-03
DOC: Update sceneviews.md replacing literal \n with newline

Leengit:
Facilitate visual parsing of the raw markdown files replacing literal "\n" with actual newlines.

### DIFF
--- a/Docs/user_guide/modules/sceneviews.md
+++ b/Docs/user_guide/modules/sceneviews.md
@@ -2,7 +2,13 @@
 
 ## Overview
 
-Scene views are a convenience tool for organizing multiple 'live views' of the data in your scene.\n\nYou can create any number of views and control parameters such as the 3D view, model visibility, window layout, and other parameters.\n\nThis can be used to set up a series of predefined starting points for looking at portions of your data in detail.\n\nFor example, you may have one overview scene which shows an external view of the body along with interior views with the skin surface turned off and slice planes visible to highlight a tumor location.
+Scene views are a convenience tool for organizing multiple 'live views' of the data in your scene.
+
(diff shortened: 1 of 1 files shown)


jcfr:
Great progress. 

Can you also take care of

> Update all markdown files found in [Docs](https://github.com/Slicer/Slicer/tree/main/Docs) directory

jcfr (changes requested):
Great progress. 

Can you also take care of

> Update all markdown files found in [Docs](https://github.com/Slicer/Slicer/tree/main/Docs) directory

Leengit:
> Can you also take care of ...

I don't see any other use of `\n` in `Docs/.../*.md` files, except inside quoted C++ or Python strings where it needs to be kept.

jcfr:
It turns out that the files I was looking at generated and consolidated in the `Docs/_moduledescriptions` folder as part of the documentation build process.

Looking at the documentation page of module like "Cast Scalar Volume", we can see that there are literal `\n`.

See https://slicer.readthedocs.io/en/latest/user_guide/modules/castscalarvolume.html

The issue is specific to the function `_generatemd` [^1] taking file like `CastScalarVolume.xml`[^2] as input.

[^1]: https://github.com/Slicer/Slicer/blob/ef38f694dc3675faf3c8203446e0f5434c35c202/Docs/conf.py#L263-L269

[^2]: https://github.com/Slicer/Slicer/blob/ef38f694dc3675faf3c8203446e0f5434c35c202/Modules/CLI/CastScalarVolume/CastScalarVolume.xml#L5

jcfr (approved):
We will squash & merge and address the generation of CLI documentation independently.

PR #7347, opened 2023-11-05
STYLE: Remove unintentionally committed files

jcfr:
Remove files unintentionally integrated in commit 52586f748 (STYLE: Update python scripts to use double quotes in strings)


jamesobutler (approved):
Was this WIP work? Or files that were relocated when they shouldn’t have been?

jcfr:
Nope ... I had the files laying around in my source tree and unintentionally added them when working on the Python pull request.

PR #7350, opened 2023-11-06
ENH: Formalize `WebServer` request handler interface

tbirdso:
## Background

The `WebServer` module extends the Python HTTPServer framework to synchronously handle requests within the Qt event loop. `WebServer` currently implements three request handlers to handler DICOM requests, static page requests, or other Slicer actions.

## Current behavior (status quo)

`SlicerWebServer` assumes that instances in its list of "handlers" implement a common interface to handle requests, but does not explicitly lay out what interface is required. This makes it difficult to develop with `WebServer` as the "contract" between the server and handlers is not formally known.

## Proposed change

This proposal defines the abstract `SlicerBaseRequestHandler` class to formalize the interface that request handlers must implement. By explicitly outlining 1) what methods are required and 2) what data `SlicerWebServer` is guaranteed to provide, we ease the development burden required to extend Slicer with custom request handling behavior. New developers can write request handlers against the given specification and expect that handling will be carried out appropriately.

## Additional Notes

`logMessage` updates appear to fix a small existing bug where the "log to console" setting in `WebServer` was not respected.

Includes type hint updates in `WebServer`.

Related unit test passes locally:
```
> ctest -C Release -R WebServer
Test project C:/repos/sl-b/Slicer-build
    Start 446: py_nomainwindow_qSlicerWebServerModuleGenericTest
1/1 Test #446: py_nomainwindow_qSlicerWebServerModuleGenericTest ...   Passed    2.91 sec

100% tests passed, 0 tests failed out of 1
```


Modules/Scripted/WebServer/CMakeLists.txt
```diff
@@ -6,6 +6,7 @@ set(MODULE_NAME WebServer)
 set(MODULE_PYTHON_SCRIPTS
   ${MODULE_NAME}.py
   ${MODULE_NAME}Lib/__init__
+  ${MODULE_NAME}Lib/base.py
```
pieper, line 9: It would be more consistent to name this BaseRequestHandler.py since in this folder the name Slicer is used to refer to the handler that handles Slicer-related requests.

Modules/Scripted/WebServer/WebServerLib/base.py
```diff
@@ -0,0 +1,73 @@
...
+
+import abc
+from typing import Callable, Optional
+
+
+BaseRequestLoggingFunction = Callable[[list[any]], None]
+"""Function signature for an external handle for message logging."""
+
+
+class SlicerBaseRequestHandler(abc.ABC):
```
pieper, line 13: This is great, but here I think the class should be BaseRequestHandler since we use Slicer in the derived class to indicate it responds to Slicer requests.

pieper (changes requested):
Thanks for your work on this 👍 

Just a couple suggestions but otherwise good.

Modules/Scripted/WebServer/CMakeLists.txt
tbirdso: My preference is not to name submodules with the same class names they implement, because it is easier to confuse new users.

This is an easy mistake to make:
```py
import WebServerLib.BaseRequestHandler

WebServerLib.BaseRequestHandler() # invalid, this is a module!

WebServerLib.BaseRequestHandler.BaseRequestHandler() # this is not intuitive!
```

Whereas using different names avoids the issue:
```py
import WebServerLib.base

WebServerLib.base.BaseRequestHandler() # namespace makes clear what is a module and what is a class
```

With that said, it is true that other `WebServerLib` files are named as `_Handler`. If there is a strong preference for consistency I will make the change. What do you think?

tbirdso: Great, thanks for the feedback. Will make the change 👍

tbirdso: Updated as `BaseRequestHandler.py` and `BaseRequestHandler` class.

pieper, line 9: Looks like you still need to change the name here?

tbirdso: Good catch!

pieper (approved):
LGTM 👍

jcfr (commented):
@tbirdso Thanks for the refactoring 🙏💯 and thanks @pieper for reviewing. 

Following the latest renamjng, can we confirm the builtin webserver work as expected?

(Since the CI doesn't (yet) run tests, j wanted to confirm the latest version  had been manually tested)

tbirdso:
> @tbirdso Thanks for the refactoring 🙏💯 and thanks @pieper for reviewing.
> 
> Following the latest renamjng, can we confirm the builtin webserver work as expected?
> 
> (Since the CI doesn't (yet) run tests, j wanted to confirm the latest version had been manually tested)

Yes, confirmed. I've started the web server in my local build and performed the following actions from the static page in my browser:
1. Loaded MRHead sample data set
2. Downloaded NRRD volume
3. Set viewers 
4. Retrieved application screenshots
5. Listed DICOM studies

Meanwhile, the WebServer unit test passes. Is there additional testing criteria to evaluate?

PR #7351, opened 2023-11-06
BUG: Fix incorrect enter/exit called after scripted module reload

### LINKED ISSUE #7211: Reloading a scripted module causes unexpected behavior when switching modules
## Summary

When reloading a Python scripted module and switching modules after that, the initial widget instance `enter` and `exit` methods are called instead of the current one.

## Steps to reproduce

1. Create a `DummyModule`(otherwise need to change name in the methods) scripted module
2. modify the top of your `setup`, `enter`, and `exit` methods
```
    def setup(self) -> None:
        print(f"{self}.setup")
        ...
        
    def enter(self) -> None:
        print(f"{self}.enter")
        command = "self is slicer.modules.DummyModuleWidget"
        print(f"{command}: {eval(command)}")
        ...

    def exit(self) -> None:
        print(f"{self}.exit")
        command = "self is slicer.modules.DummyModuleWidget"
        print(f"{command}: {eval(command)}")
        ...
        
```
3. Run Slicer
4. Open the module (enter and exit)
5. Reload the module
6. Enter and exit

## Expected behavior

`enter`, and `exit` methods of the currently loaded (reloaded) widget should be called. 

## Environment
Slicer version ...........: 5.4.0 (revision 31938 / 311cb26) macosx-amd64 - installed release
Operating system .........: macOS / 13.4.1 / 22F82 / UTF-8 - 64-bit

Sunderlandkyl:
When a scripted loadable module was reloaded using slicer.util.reloadScriptedModule, pythonSource was not changed in qSlicerScriptedLoadableModuleWidget. As a result, the previous enter/exit methods continued to be used by the scripted module.

Fixed by making setPythonSource accessible in Python, and using it to instantiate the new widget object.

Fix #7211

### DIFF
--- a/Base/Python/slicer/util.py
+++ b/Base/Python/slicer/util.py
@@ -1414,10 +1414,10 @@ def reloadScriptedModule(moduleName):
     for item in items:
         parent.layout().removeItem(item)
 
-    # create new widget inside existing parent
-    widget = eval("reloaded_module.%s(parent)" % widgetName)
-    widget.setup()
-    setattr(slicer.modules, widgetName, widget)
+    # Creates new widget at slicer.modules.{widgetName}.
+    # Also ensures that qSlicerScriptedLoadableModuleWidget has references to updated enter/exit/setup methods.
+    # See https://github.com/Slicer/Slicer/issues/7424
+    widget.parent.reload()
 
     return reloaded_module
 
--- a/Base/QTGUI/qSlicerScriptedLoadableModuleWidget.h
+++ b/Base/QTGUI/qSlicerScriptedLoadableModuleWidget.h
@@ -47,6 +47,8 @@ public:
   QString pythonSource()const;
   bool setPythonSource(const QString& filePath, const QString& className = QLatin1String(""));
 
+  Q_INVOKABLE void reload();
+
   /// Convenience method allowing to retrieve the associated scripted instance
   Q_INVOKABLE PyObject* self() const;
 
--- a/Base/QTGUI/qSlicerScriptedLoadableModuleWidget.cxx
+++ b/Base/QTGUI/qSlicerScriptedLoadableModuleWidget.cxx
@@ -177,6 +177,18 @@ bool qSlicerScriptedLoadableModuleWidget::setPythonSource(const QString& filePat
 }
 
 //-----------------------------------------------------------------------------
+void qSlicerScriptedLoadableModuleWidget::reload()
+{
+  Q_D(qSlicerScriptedLoadableModuleWidget);
+  if (!QFileInfo::exists(d->PythonSourceFilePath))
+    {
+    return;
+    }
+  this->setPythonSource(this->pythonSource());
+  this->setup();
+}
+
+//-----------------------------------------------------------------------------
 PyObject* qSlicerScriptedLoadableModuleWidget::self() const
 {
   Q_D(const qSlicerScriptedLoadableModuleWidget);


Base/Python/slicer/util.py
```diff
@@ -1372,10 +1372,12 @@ def reloadScriptedModule(moduleName):
     for item in items:
         parent.layout().removeItem(item)
 
-    # create new widget inside existing parent
-    widget = eval("reloaded_module.%s(parent)" % widgetName)
+    # Creates new widget at slicer.modules.{widgetName}.
+    # Also ensures that qSlicerScriptedLoadableModuleWidget has references to updated enter/exit/setup methods.
+    parent.setPythonSource(filePath)
```
pieper, line 1377: I realize this method name predates this PR, but I find it really confusing that what looks like a property setter has such profound side effects.  Also it's oddly named since it takes a filePath but the method name kind of implies it would take the literal source code.  Since before it was kind of an internal class that was only available in C++ I think we could expose a new name like `setSourceFilePathAndInstantiate` and deprecate the old method name.

pieper (approved):
My comment is just an idea.  If you need this feature I'm okay with merging as-is.

Base/Python/slicer/util.py
```diff
@@ -1372,10 +1372,12 @@ def reloadScriptedModule(moduleName):
     for item in items:
         parent.layout().removeItem(item)
 
-    # create new widget inside existing parent
-    widget = eval("reloaded_module.%s(parent)" % widgetName)
```
jcfr, line 1376: If we remove the use of the `reloaded_module` ivar, we should also revisit 

https://github.com/Slicer/Slicer/blob/14a4056f2bc935b7bd54e5eaa13941aa4c60aa66/Base/Python/slicer/util.py#L1346-L1356

Base/QTGUI/qSlicerScriptedLoadableModuleWidget.h
```diff
@@ -44,8 +44,8 @@ class Q_SLICER_BASE_QTGUI_EXPORT qSlicerScriptedLoadableModuleWidget
   qSlicerScriptedLoadableModuleWidget(QWidget * parentWidget=nullptr);
   ~qSlicerScriptedLoadableModuleWidget() override;
 
-  QString pythonSource()const;
-  bool setPythonSource(const QString& newPythonSource, const QString& className = QLatin1String(""));
+  Q_INVOKABLE QString pythonSource()const;
+  Q_INVOKABLE bool setPythonSource(const QString& newPythonSource, const QString& className = QLatin1String(""));
```
jcfr, line 48: Before moving forward, it may be sensible to revisit the naming as suggested by @pieper

Base/Python/slicer/util.py
jamesobutler: I agree that `setPythonSource` would mean the input is python code rather than a filepath to python code. An updated name would be appreciated from a developer perspective 🙏🏻

Base/QTGUI/qSlicerScriptedLoadableModuleWidget.h
jcfr: As a step toward this:
* https://github.com/Slicer/Slicer/pull/7352

jcfr (changes requested):
Before moving forward we should revisit code setting the `reloaded_module` internal variable as this is likely not needed anymore.

Base/Python/slicer/util.py
pieper: I agree with @jcfr - this code never seemed like a clean implementation but as I recall it was the best I could figure out at the time.  If there's a cleaner and more maintainable implementation it would be great.

jcfr:
The root cause likes in the fact each scripted module C++ instance hold a reference to a `qSlicerPythonCppAPI` instance associated with the `d->PythonCppAPI` internal variable.

When calling "virtual" functions from Python (e.g when the module manager calls `qSlicerAbstractModuleWidget::enter()`), reference to the previous handle are still used.

We need to address this more systematically by adding a "reload()" function, I will propose a fix.

Base/Python/slicer/util.py
Sunderlandkyl: Current the imp.load_module call still seems to be required. Without it, the module is not actually reloaded.

Sunderlandkyl:
@jcfr 
> The root cause of this issue is that each C++ instance of a scripted module maintains a reference to a qSlicerPythonCppAPI instance through the d->PythonCppAPI internal variable. When virtual functions are invoked from Python (for instance, when the module manager calls qSlicerAbstractModuleWidget::enter()), the previous reference handles are still utilized.
> 
> To comprehensively resolve this, we should implement a systematic solution by introducing a "reload()" function at the C++ level. This function will ensure that the underlying PythonCppAPI object is updated correctly. To prevent any potential "catch 22" scenarios, the reload API should be accessible via the module manager.
> 
> I will propose a fix for this issue.

Should I keep this PR open. or close it in favor of a more comprehensive solution?

jcfr:
> Should I keep this PR open. or close it in favor of a more comprehensive solution?

I will follow-up this afternoon by either:
* creating a pull-request proposing an "cleaner" approach
* or slightly cleaning up this pull-request and creating an issue documenting the "proper" workaround

jcfr (approved):

Issue #7344, opened 2023-11-03
Streamline execution of pre-commit tooling from a build tree.


jcfr:
## Additional context

_Originally posted by @lassoan in https://github.com/Slicer/Slicer/issues/7338#issuecomment-1792301031_
> These improvements all sound good, thanks for working on it. What I would really appreciate though if it was possible to:
> - easily run the exact same checks in the local build tree without any extra configuration, maybe as part of the build process
> - have an option for the the linter to change the source code to make it compliant

jamesobutler:
See https://github.com/InsightSoftwareConsortium/ITK/commit/6fd13e1e9fb8f53668b8c787f47646252005cb03 for how ITK added pre-commit usage to the SetupForDevelopment.sh. Also see further improvement of the install of the pre-commit dependency in SetupForDevelopment.sh by way of https://github.com/InsightSoftwareConsortium/ITK/commit/6f25d918f4e937140e9eac14fb28ecb8409e1da2.

PR #7341, opened 2023-11-03
ENH: Require space and capital letter after colon in commit message first line

Leengit:
Enforce that there be (at least) one space after `BUG:`, `COMP:`, `DOC:`, `ENH:`, `PERF:`, or `STYLE:` at the start of the first line of a commit message.

Current usage shows me the following counts in the `git log` for trailing spaces.  In no cases were there 3 or more spaces.
```text
     39 'BUG:  '
  11109 'BUG: '
     39 'BUG:'
     55 'COMP:  '
   6642 'COMP: '
      6 'COMP:'
      6 'DOC:  '
   1110 'DOC: '
      0 'DOC:'
     86 'ENH:  '
  16342 'ENH: '
     73 'ENH:'
      4 'PERF:  '
    225 'PERF: '
      0 'PERF:'
     14 'STYLE:  '
   3610 'STYLE: '
     11 'STYLE:'
```

PR #7360, opened 2023-11-08
ENH: Add GitHub workflow for automating SetupForDevelopment.sh version updates

jcfr:
This enhancement introduces a scheduled daily workflow that calculates the version by summing the number of revisions associated with `SetupForDevelopment.sh`, its associated scripts, and the revisions of the `hooks`[^1] branch. If the version has changed since the last update, the workflow automatically creates a corresponding pull request.

This ensures that developers are prompted to update their development environments using the `SetupForDevelopment.sh` script.

The version that will be updated can be found here:

https://github.com/Slicer/Slicer/blob/c2e136fa1addcb64185faba49344d8121100264e/Utilities/SetupForDevelopment.sh#L47-L50

[^1]: https://github.com/Slicer/Slicer/tree/hooks


jcfr:
the workflow has been test in the `jcfr/ci-sandbox` project:
* https://github.com/jcfr/ci-sandbox/pull/48

jcfr:
Good point. Since we have all information available, it should indeed be possible to avoid this.

jcfr:
Your concern about polluting the commit history with derived information is sensible.

At the same time, the variable `SetupForDevelopment_VERSION` sees infrequent updates[^1], typically a few times a year. However, we are still forgetting to update it and this led to out of sync hooks[^2], this pull request addresses the problem without re-engineering the mechanism.

For this reason, I would like to move forward with the integration.

Note that I created a dedicated issue with a proposal for improving how we handle versioning:
* https://github.com/Slicer/Slicer/issues/7364

[^1]: https://github.com/Slicer/Slicer/commits/hooks

[^2]: The `hooks` branch was updated on Sept 22th 2021 with https://github.com/Slicer/Slicer/commit/e1b8ebcaa68055479078767f00ce6052bbf0e9d3 but the last version update was from October 2021 (https://github.com/Slicer/Slicer/commits/main/Utilities/SetupForDevelopment.sh)

jcfr:
@lassoan While looking into improving the approach, I realized[^1] implementing an automatic way to check would require a network connection during `pre-commit` and this is not something that can be guaranteed.

Given this limitation[^2], I suggest we move forward with the integration.

While exploring potential alternatives, I also look if there were hooks to run a script after a fetch... but this doesn't seem to be possible[^2][^3]


[^1]: https://github.com/Slicer/Slicer/issues/7364#issuecomment-1803216101
[^2]: https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks
[^3]: https://stackoverflow.com/questions/5623208/how-to-execute-a-command-right-after-a-fetch-or-pull-command-in-git

lassoan:
Could we check in the precommit hook that the commit includes one of the files in a list, and if it does then check that the version is updated then the version in SetuoForDevelopment.sh?

If this is not feasible, could we add one more test to the automated tests in the pull request? 

I still feel that this automated mechanism, pull request, git commit is an overkill for such a low-impact issue. I also worry a bit that this will add some more complexities for repositories of custom applications (hosted on Github or elsewhere). If we added more daily checks to this mechanism then it could justify the added complexity, but going back to daily checks from continuous/preventive checks would be odd, as we should rather go in the other direction. 

I don't have strong objections, though, so if you really want this in then you can merge. But maybe we could have a quick call to discuss this to make sure we don't miss anything important.

jcfr:
> Could we check in the precommit hook that the commit includes one of the files in a list, and if it does then check that the version is updated then the version in SetuoForDevelopment.sh?

I am not sure to understand the suggestion.

> I don't have strong objections, though, so if you really want this in then you can merge. But maybe we could have a quick call to discuss this to make sure we don't miss anything important.

Great, this is a scheduled a workflow that will simply create a pull-request to update the version as we have done (but forgotten) in the past. 


> I don’t see a lot of value with SetupForDevelopment when Slicer development always ultimately goes through PR in which there can be status checks to review commits.

Having the ability to check locally is important especially when developer work on a series of commits. 

As core developer who are used to the process, it surely adds less value. That said, the ability to perform these checks locally is important for new contributors.

Consolidation of how we perform such checks so we leverage the same infrastructure as we do using GitHub Actions has been documented in this issue:
* https://github.com/Slicer/Slicer/issues/7344

lassoan:
>>Could we check in the precommit hook that the commit includes one of the files in a list, and if it does then check that the version is updated then the version in SetuoForDevelopment.sh?
>
>I am not sure to understand the suggestion.

In the pre-commit hook:
- check if the list of modified files include `SetupForDevelopment.sh`, `SetupUser.sh`, `SetupHooks.sh`, or `GitTips.sh`
- if any of these files are modified then check if the `SetupForDevelopment_VERSION` line is modified in `SetupForDevelopment.sh`
- if the version line was not modified in the commit then reject it

pieper:
> Having the ability to check locally is important especially when developer work on a series of commits.

Yes, I think this is useful, but I realize now that I haven't been benefiting from it even on the machine and Slicer checkout I've been developing on for years.

The reason appears to be that I always use `git gui` as a consistent handy tool across platforms, but this led to a couple issues.

To investigate I tried a bad commit message on a local test branch with the cli and found this:

```
 % git commit -m "dfkjdfkjdfkj" README.md 
hint: The '.git/hooks/pre-commit' hook was ignored because it's not set as executable.
hint: You can disable this warning with `git config advice.ignoredHook false`.
[test-branch a1f69bd4be] dfkjdfkjdfkj
 1 file changed, 3 deletions(-)
```

I never got this message via `git gui` so my commits hadn't been tested.

I changed the permissions and now `git gui` does run the hooks and warns me of bad messages, so that's good.

**Suggestion one**: can we change the SetupForDevelopment script to make sure the hooks are executable?

On further investigation I saw that this commit message passes the local commit checks:
```
% git commit -m "BUG: dfkjdfkjdfkj" README.md
[test-branch 5e2860ea91] BUG: dfkjdfkjdfkj
 1 file changed, 3 deletions(-)
```

But our github commit checks complain about a lower case letter after the prefix.

**Suggestion two**: do we have a way to keep these in sync?  If we can't use the same check script in both places, can we at least make them consistent with respect to this capitalization issue?

Sorry that this is a bit off-topic of the PR, but since the conversation about the value of the local hooks led me to investigate.   I can start new issues for these if we want to track them separately.

lassoan:
Update Slicer.crt is a different story, it is good, because:
- it does not generate a new commit from data that is entirely stored in the repository already
- it is doing the check at the right time - while the proposed change does not prevent an incorrect commit (changing scripts without incrementing the version) but it fixes things later in a separate commit; this delay can lead to merge conflicts, or maybe even infinite loops when when synchronizing repositories with forks that also have this mechanism enabled

jamesobutler:
> Suggestion two: do we have a way to keep these in sync? If we can't use the same check script in both places, can we at least make them consistent with respect to this capitalization issue?

@pieper This issue of the local check not matching the PR checks is something I brought up in https://github.com/Slicer/Slicer/pull/7341#pullrequestreview-1712944360, maybe that conversation can continue over there.

lassoan:
I agree with @pieper, it would be great if the checks would be exactly the same.

Lint checks on github are so annoying and time-consuming because I cannot easily replicate the same check locally (admittedly I did not try it much), so I need to keep pushing changes, let the CI run, make changes, push them again, and repeat this until there are no more errors.

I had to disable pre-commit hook a number of times because of too strict/incorrect checks.

If we used the exact same scripts in the PR tests and pre-commit hook then it would be much easier to ensure that checks are good quality and consistent.

This would also address @jamesobutler point that the pre-commit hook should be just an optional convenience feature for frequent committers (for them it is worth investing the time into setting up the local checks), for everyone there are the tests in the pull request CI anyway.

If we consider the pre-commit hook tests to be optional then we could require tools that would be impractical to require all developers to install. For example, we could require configuring a Python environment for the hooks or maybe even docker.

jcfr:
> Suggestion one: can we change the SetupForDevelopment script to make sure the hooks are executable?

The file seems to already be an executable file:



--------

> Suggestion two: do we have a way to keep these in sync?

Yes. Ideally, we should removing the duplication of logic by running `pre-commit` hooks locally. This is described in:
* https://github.com/Slicer/Slicer/issues/7344

In the meantime, we will have to improve the `commit-msg` script.

----------------

> it does not generate a new commit from data that is entirely stored in the repository already
> 

By storing the hooks into a folder called `Utilities/GitSetup` like it is now done for CMake[^1], we should be able to automate the version check.

[^1]: https://gitlab.kitware.com/cmake/cmake/-/tree/master/Utilities/GitSetup

jcfr:
We will revisit the approach by maintaining the hooks into a sub-directories called `Utilities/GitSetup`

PR #7362, opened 2023-11-08
ENH: Support regularization transform hardening in DICOM Scalar plugin 

Punzo:
this PR adds:

- Grouping frames by orientation with a 10^-6 tolerance in examineFiles in DICOM module.
- Add harden transform option in the DICOM settings for load in DICOMScalarVolumePlugin.

### DIFF
--- a/Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
+++ b/Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
@@ -26,10 +26,16 @@ class DICOMScalarVolumePluginClass(DICOMPlugin):
     """ ScalarVolume specific interpretation code
     """
 
-    def __init__(self, epsilon=0.01):
+    def __init__(self, spacingEpsilon=1e-2, orientationEpsilon=1e-6):
+        """
+        :param spacingEpsilon: Tolerance for numeric comparisons for spacing vectors. Defaults to 1.e-2.
+        :param orientationEpsilon: Tolerance for numeric comparisons for orientation vectors. Defaults to 1.e-6, following ITK numeric comparison logic.
+        """
+
         super().__init__()
         self.loadType = _("Scalar Volume")
-        self.epsilon = epsilon
+        self.spacingEpsilon = spacingEpsilon
+        self.orientationEpsilon = orientationEpsilon
         self.acquisitionModeling = None
         self.defaultStudyID = "SLICER10001"  # TODO: What should be the new study ID?
 
@@ -90,7 +96,7 @@ class DICOMScalarVolumePluginClass(DICOMPlugin):
         importFormatsComboBox.addItem(_("default (apply regularization transform)"), "default")
         importFormatsComboBox.addItem(_("none"), "none")
         importFormatsComboBox.addItem(_("apply regularization transform"), "transform")
-        # In the future additional option, such as "resample" (harden the applied transform) may be added.
+        importFormatsComboBox.addItem(_("harden regularization transform"), "hardenTransform")
 
         importFormatsComboBox.currentIndex = 0
         formLayout.addRow(_("Acquisition geometry regularization:"), importFormatsComboBox)
@@ -134,6 +140,10 @@ class DICOMScalarVolumePluginClass(DICOMPlugin):
             comparison += _("Pixel data mismatch") + "\n"
         return comparison
 
+    def hardenAcquisitionGeometryRegularization(self):
+        settings = qt.QSettings()
+        return (settings.value("DICOM/ScalarVolume/AcquisitionGeometryRegularization", "default") == "hardenTransform")
+
     def acquisitionGeometryRegularizationEnabled(self):
         settings = qt.QSettings()
         return (settings.value("DICOM/ScalarVolume/AcquisitionGeometryRegularization", "default") != "none")
@@ -170,6 +180,9 @@ class DICOMScalarVolumePluginClass(DICOMPlugin):
         cleanValue = cleanValue.replace("*", "(star)")
         cleanValue = cleanValue.replace("\\", "-")
         return cleanValue
+
+    def tagValueToVector(self, value):
+        return numpy.array([float(element) for element in value.split("\\")])
 
     def examineFiles(self, files):
         """ Returns a list of DICOMLoadable instances
@@ -212,6 +225,16 @@ class DICOMScalarVolumePluginClass(DICOMPlugin):
             "diffusionGradientOrientation",
         ]
 
+        # Subset of the tags specified in the `subseriesTags` list
+        # that should be compared numerically (rather than using the
+        # built-in string comparison). This is done when checking if
+        # the associated value has already been appended to the
+        # `subseriesValues[tag]` list.
+        vectorTags = [
+            "imageOrientationPatient",
+            "diffusionGradientOrientation",
+        ]
+
         #
         # first, look for subseries within this series
         # - build a list of files for each unique value
@@ -224,9 +247,31 @@ class DICOMScalarVolumePluginClass(DICOMPlugin):
             for tag in subseriesTags:
                 value = slicer.dicomDatabase.fileValue(file, self.tags[tag])
                 value = value.replace(",", "_")  # remove commas so it can be used as an index
+
                 if tag not in subseriesValues:
                     subseriesValues[tag] = []
-                if value not in subseriesValues[tag]:
+
+                if tag in vectorTags:
+                    if value != "":
+                        vector = self.tagValueToVector(value)
+
+                        found = False
+                        for subseriesValue in subseriesValues[tag]:
+                            subseriesVector = self.tagValueToVector(subseriesValue)
+                            # vector numerical comparison by absolute difference as the ITK logic.
+                            # Reference:
+                            #   Class: ITK/Modules/Numerics/Optimizersv4/include/itkObjectToObjectMetric.hxx
+                            #   Method: VerifyDisplacementFieldSizeAndPhysicalSpace
+                            #   URL: https://github.com/InsightSoftwareConsortium/ITK/blob/v5.4rc02/Modules/Numerics/Optimizersv4/include/itkObjectToObjectMetric.hxx#L507-L510.
+
+                            if numpy.allclose(vector, subseriesVector, rtol=0., atol=self.orientationEpsilon):
+                                found = True
+                                value = subseriesValue
+                                break
+
+                        if not found:
+                            subseriesValues[tag].append(value)
+                elif value not in subseriesValues[tag]:
                     subseriesValues[tag].append(value)
                 if (tag, value) not in subseriesFiles:
                     subseriesFiles[tag, value] = []
@@ -307,7 +352,7 @@ class DICOMScalarVolumePluginClass(DICOMPlugin):
         # then adjust confidence values based on warnings
         #
         for loadable in loadables:
-            loadable.files, distances, loadable.warning = DICOMUtils.getSortedImageFiles(loadable.files, self.epsilon)
+            loadable.files, distances, loadable.warning = DICOMUtils.getSortedImageFiles(loadable.files, self.spacingEpsilon)
 
         loadablesBetterThanAllFiles = []
         if allFilesLoadable.warning != "":
@@ -539,7 +584,8 @@ class DICOMScalarVolumePluginClass(DICOMPlugin):
         if volumeNode:
             self.acquisitionModeling = self.AcquisitionModeling()
             self.acquisitionModeling.createAcquisitionTransform(volumeNode,
-                                                                addAcquisitionTransformIfNeeded=self.acquisitionGeometryRegularizationEnabled())
+                                                                addAcquisitionTransformIfNeeded=self.acquisitionGeometryRegularizationEnabled(),
+                                                                hardenAcquisitionTransform=self.hardenAcquisitionGeometryRegularization())
 
         return volumeNode
 
@@ -827,7 +873,7 @@ class DICOMScalarVolumePluginClass(DICOMPlugin):
                         volumeNode.TransformPointToWorld(corners[slice, row, column], worldCorners[slice, row, column])
             return worldCorners
 
-        def createAcquisitionTransform(self, volumeNode, addAcquisitionTransformIfNeeded=True):
+        def createAcquisitionTransform(self, volumeNode, addAcquisitionTransformIfNeeded=True, hardenAcquisitionTransform=False):
             """Creates the actual transform if needed.
             Slice corners are cached for inspection by tests
             """
@@ -846,6 +892,8 @@ class DICOMScalarVolumePluginClass(DICOMPlugin):
                     self.fixedCorners = self.cornersToWorld(volumeNode, self.originalCorners)
                     if not numpy.allclose(self.fixedCorners, self.targetCorners):
                         raise Exception("Acquisition transform didn't fix slice corners!")
+                    if hardenAcquisitionTransform:
+                        volumeNode.HardenTransform()
                 else:
                     logging.warning(warningText + "  Regularization transform is not added, as the option is disabled.")
             elif maxError > 0 and maxError > self.zeroEpsilon:
--- a/Docs/user_guide/modules/dicom.md
+++ b/Docs/user_guide/modules/dicom.md
@@ -183,7 +183,7 @@ DICOM module settings:
         - Load referenced series will give you the option of easily loading, for example, the source volume of a segmentation when you open the segmentation.  This can also be made to happen automatically.
     - DICOMScalarVolumePlugin settings:
         - You can choose what back-end library to use (currently GDCM, DCMTK, or GDCM with DCMTK fallback with the last option being the default.  This is provided in case some data is unsupported by one library or the other.
-        - Acquisition geometry regularization option supports the creation of a nonlinear transform that corrects for things like missing slices or gantry tilt in the acquisition.  See more information [here](https://github.com/Slicer/Slicer/commit/3328b81211cb2e9ae16a0b49097744171c8c71c0)
+        - Acquisition geometry regularization option supports the creation of a nonlinear transform that corrects for things like missing slices or gantry tilt in the acquisition. The regularization transformation can also be hardened to the volume. See more information [here](https://github.com/Slicer/Slicer/commit/3328b81211cb2e9ae16a0b49097744171c8c71c0)
         - Autoloading subseries by time is an option break up some 4D acquisitions into individual volume, but is optional since some volumes are also acquired in time unites and should not be split.
 
 ## Troubleshooting


Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
```diff
@@ -90,6 +91,7 @@ def settingsPanelEntry(panel, parent):
         importFormatsComboBox.addItem(_("default (apply regularization transform)"), "default")
         importFormatsComboBox.addItem(_("none"), "none")
         importFormatsComboBox.addItem(_("apply regularization transform"), "transform")
+        importFormatsComboBox.addItem(_("apply harden regularization transform"), "hardenTransform")
         # In the future additional option, such as "resample" (harden the applied transform) may be added.
```
pieper, line 95: This comment will be out of date so it should be removed with this PR.

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
```diff
@@ -134,6 +136,10 @@ def compareVolumeNodes(volumeNode1, volumeNode2):
             comparison += _("Pixel data mismatch") + "\n"
         return comparison
 
+    def hardenAcquisitionGeometryRegularization(self):
+        settings = qt.QSettings()
+        return (settings.value("DICOM/ScalarVolume/AcquisitionGeometryRegularization", "default") == "hardenTransform")
```
pieper, line 141: This will change the behavior of the module - is this intended?

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
```diff
@@ -224,10 +230,40 @@ def examineFiles(self, files):
...
                 value = value.replace(",", "_")  # remove commas so it can be used as an index
+
                 if tag not in subseriesValues:
                     subseriesValues[tag] = []
-                if not subseriesValues[tag].__contains__(value):
+
+                if tag == "imageOrientationPatient" or tag == "diffusionGradientOrientation":
+                    if value != "":
+                        directionCosine = [float(i) for i in value.split("\\")]
+                        a = math.sqrt(directionCosine[0]*directionCosine[0] + directionCosine[1]*directionCosine[1] + directionCosine[2]*directionCosine[2])
```
pieper, line 240: Stylistically it's odd to mix a very descriptive variable name `directionCosine` and a meaningless single character variable name `a` in the same line, especially when the long one is repeated so much.

I think it would be cleaner to make a numpy array `directionCosines` and use `numpy.linalg.norm` and to give ` a name like `directionMagnitude`.

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
```diff
@@ -224,10 +230,40 @@ def examineFiles(self, files):
             for tag in subseriesTags:
                 value = slicer.dicomDatabase.fileValue(file, self.tags[tag])
                 value = value.replace(",", "_")  # remove commas so it can be used as an index
+
                 if tag not in subseriesValues:
                     subseriesValues[tag] = []
-                if not subseriesValues[tag].__contains__(value):
+
+                if tag == "imageOrientationPatient" or tag == "diffusionGradientOrientation":
+                    if value != "":
```
pieper, line 238: A comment here to define why these tags are unique would help.  Also defining a list or set with the name like `vectorTags` then checking `if tag in vectorTags:` would make the code more readable and easier to cleanly extend to more vector tags in the future.

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
```diff
@@ -224,10 +230,40 @@ def examineFiles(self, files):
...
+                            directionCosine[k] /= a
+                        a = math.sqrt(directionCosine[3]*directionCosine[3] + directionCosine[4]*directionCosine[4] + directionCosine[5]*directionCosine[5])
+                        for k in range(3, 6):
+                            directionCosine[k] /= a
+
+                        found = False
+                        for subseriesValue in subseriesValues[tag]:
+                            subserDirCos = [float(i) for i in subseriesValue.split("\\")]
+                            a = math.sqrt(subserDirCos[0]*subserDirCos[0] + subserDirCos[1]*subserDirCos[1] + subserDirCos[2]*subserDirCos[2])
+                            b = (directionCosine[0]*subserDirCos[0] + directionCosine[1]*subserDirCos[1] + directionCosine[2]*subserDirCos[2]) / a
```
pieper, line 251: Here again, more descriptive names for `a` and `b` and less math noise would make this easier to understand.

pieper (changes requested):
Thanks for working on this 👍 

Another stylistic comment is that the commit message is longer than recommended.

https://slicer.readthedocs.io/en/latest/developer_guide/contributing.html#how-to-write-commit-messages

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
Punzo: thanks, I have actually missed it (:

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
Punzo: yes it is intended, but the default behavior will not be changed (i.e. create and observe transform, without resampling the volume)

Punzo: see https://github.com/Slicer/Slicer/pull/7362/files#diff-ddb2464152946e0fec0e7fcd1ec6cc4aefcd6b8a9ad3ee82a136564525f0c66fR579-R887

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
Punzo: I agree

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
lassoan: I would use `np.linalg.norm(directionCosine)` instead, but I don't think we actually need normalization.

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
pieper: Okay good, I see now.  I do find I almost always want the harden step, but for now I do prefer doing it manually so I know that it's happened.  But I can see this being confusing for most people.

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
Punzo: you are right, comment applied

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
Punzo: you are right, comment applied

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
Punzo: > I would use `np.linalg.norm(directionCosine)` instead, but I don't think we actually need normalization.

it is true that imageOrientationPatient should be already normalized. However I have experienced it is not always the case. The load function in the scalar plugin normalize them as well.

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
lassoan: I don't understand why the comparison is done like this. It looks complex and maybe there is a reason for it but it is not obvious at all.

Comparing orientations looks easy but it is actually super hard. Several ITK core developers struggled with it for about a year and it took them several iterations to get it right - so that they can detect real differences but not harmless differences due to numerical imprecision. I'll try to find the code.

Punzo: I cleaned the code, see https://github.com/Slicer/Slicer/pull/7362/files#diff-ddb2464152946e0fec0e7fcd1ec6cc4aefcd6b8a9ad3ee82a136564525f0c66fR244-R272

it is a simple dot product and check if it is paralle or antiparallel (-1, +1). I took the code from the load function in the vtkITKArchetypeImageSeriesReader, i.e.  https://github.com/Slicer/Slicer/blob/1cbb44434ed36d7b7bdcb9d8a207b673571a4c0c/Libs/vtkITK/vtkITKArchetypeImageSeriesReader.h#L524-L531

Punzo: ok @lassoan @pieper changed the code to use the same behaviour of ITK numeric comparison logic (absolute differences). See:

https://github.com/Slicer/Slicer/pull/7362/files#diff-ddb2464152946e0fec0e7fcd1ec6cc4aefcd6b8a9ad3ee82a136564525f0c66fR250-R266

Punzo:
> Thanks for working on this 👍
> 
> Another stylistic comment is that the commit message is longer than recommended.
> 
> https://slicer.readthedocs.io/en/latest/developer_guide/contributing.html#how-to-write-commit-messages

Thanks for the review Steve. I have addressed all the comments. 

@pieper @lassoan up to you to merge if you think it is ready.

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
```diff
@@ -224,10 +237,39 @@ def examineFiles(self, files):
...
+
                 if tag not in subseriesValues:
                     subseriesValues[tag] = []
-                if not subseriesValues[tag].__contains__(value):
+
+                if tag in vectorTags:
+                    if value != "":
+                        directionCosine = [float(i) for i in value.split("\\")]
+                        directionCosineA = numpy.array(directionCosine[:3])
+                        directionCosineB = numpy.array(directionCosine[-3:])
```
pieper, line 248: The code is much more readable now!  In fact, now that I see what's going on I realize probably these called `rowDirectionCosines` and `columnDirectionCosines` for each better readability.

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
```diff
@@ -224,10 +237,39 @@ def examineFiles(self, files):
...
+                            subserDirCos = [float(i) for i in subseriesValue.split("\\")]
+
+                            subserDirCosA = numpy.array(subserDirCos[:3])
+                            # eps value default to 1.e-6 as same behaviour of ITK numeric comparison logic.
+                            # rtol=0, atol=eps
+                            if not numpy.allclose(directionCosineA, subserDirCosA, 0., eps):
+                                continue
+
+                            subserDirCosB = numpy.array(subserDirCos[-3:])
+                            # eps value default to 1.e-6 as same behaviour of ITK numeric comparison logic.
```
pieper, line 261: I don't think you need to repeat this comment

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
```diff
@@ -224,10 +237,39 @@ def examineFiles(self, files):
...
+                            # rtol=0, atol=eps
```
pieper, line 262: Rather than putting this in the comment, maybe just put them in as keyword arguments?

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
pieper: But also I now realize that this probably won't work for `diffusionGradientOrientation` which has just 3 components. https://dicom.innolitics.com/ciods/enhanced-mr-color-image/enhanced-mr-color-image-multi-frame-functional-groups/52009229/00189117/00189076/00189089

So probably two branches are needed.

pieper (changes requested):
Sorry for all the nitpicks : )

But the diffusion issue is important since I don't think the current code will work.

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
```diff
@@ -90,7 +90,7 @@ def settingsPanelEntry(panel, parent):
         importFormatsComboBox.addItem(_("default (apply regularization transform)"), "default")
         importFormatsComboBox.addItem(_("none"), "none")
         importFormatsComboBox.addItem(_("apply regularization transform"), "transform")
-        # In the future additional option, such as "resample" (harden the applied transform) may be added.
+        importFormatsComboBox.addItem(_("apply harden regularization transform"), "hardenTransform")
```
lassoan, line 93:
```suggestion
        importFormatsComboBox.addItem(_("harden regularization transform"), "hardenTransform")
```

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
lassoan: I don't think we need to split to vectors at all, as we just compare each component anyway.

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
lassoan: I agree. Specify `absoluteTolerance` and `relativeTolerance` variables somewhere above, refer to the ITK implementation there (add least className::methodName, maybe also a full URL).

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
Punzo: right, the splitting was from the commit before (in which I was comparing the arrays by dot product). Not splitting and comparing all the components fix all. Thanks for catchign this. Fixed in https://github.com/Slicer/Slicer/pull/7362/commits/1ccb8ecb92cfb2c2a8db4d2cf5cdc46da5a2c04d

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
Punzo: done in https://github.com/Slicer/Slicer/pull/7362/commits/1ccb8ecb92cfb2c2a8db4d2cf5cdc46da5a2c04d

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
Punzo: done in https://github.com/Slicer/Slicer/pull/7362/commits/1ccb8ecb92cfb2c2a8db4d2cf5cdc46da5a2c04d

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
Punzo: done in https://github.com/Slicer/Slicer/pull/7362/commits/1ccb8ecb92cfb2c2a8db4d2cf5cdc46da5a2c04d

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
```diff
@@ -224,10 +237,34 @@ def examineFiles(self, files):
...
                 value = slicer.dicomDatabase.fileValue(file, self.tags[tag])
                 value = value.replace(",", "_")  # remove commas so it can be used as an index
+
                 if tag not in subseriesValues:
                     subseriesValues[tag] = []
-                if not subseriesValues[tag].__contains__(value):
+
+                if tag in vectorTags:
+                    if value != "":
+                        directionCosine = numpy.array([float(i) for i in value.split("\\")])
```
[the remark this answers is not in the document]
Punzo: I agree!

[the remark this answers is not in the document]
Punzo: fixed in https://github.com/Slicer/Slicer/pull/7362/commits/c3bcf717b74bb5335595bd3d75b874050730bc9b

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
```diff
@@ -224,10 +237,34 @@ def examineFiles(self, files):
...
                     subseriesValues[tag] = []
-                if not subseriesValues[tag].__contains__(value):
+
+                if tag in vectorTags:
+                    if value != "":
+                        element = numpy.array([float(i) for i in value.split("\\")])
+
+                        found = False
+                        for subseriesValue in subseriesValues[tag]:
+                            subseriesElement = numpy.array([float(i) for i in subseriesValue.split("\\")])
```
pieper, line 250: Sorry, I wasn't clear : o

What I meant was that each part of the vector that was previous a cosine could be more generally be called an element, but the list of elements would be a vector.  So my suggestion would be

`vector = numpy.array([float(i) for i in value.split("\\")])`
 
and since we aren't ever working with the individual components of the vector I guess we don't need the word element after all.

Punzo: Sure, not a problem. changed in https://github.com/Slicer/Slicer/pull/7362/commits/64b037eb264de0d6f87c4b0aa79e49bd399176ca 

as 

`vector = numpy.array([float(element) for element in value.split("\\")])`

Punzo: I added element word in the split (: let me know if it is ok (:

pieper (dismissed):
Ahhh, looks nice now : )

Punzo:
> Ahhh, looks nice now : )

perfect, thanks for the review :)

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
```diff
@@ -171,10 +175,12 @@ def cleanNodeName(self, value):
         cleanValue = cleanValue.replace("\\", "-")
         return cleanValue
 
-    def examineFiles(self, files):
+    def examineFiles(self, files, eps = 1.e-6):
```
jcfr, line 178:
```suggestion
    def examineFiles(self, files, eps=1.e-6):
```

While we are not yet using "ruff format" (or black), this will be the recommended style.

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
```diff
@@ -224,10 +237,34 @@ def examineFiles(self, files):
...
+                        for subseriesValue in subseriesValues[tag]:
+                            subseriesVector = numpy.array([float(element) for element in subseriesValue.split("\\")])
+
+                            # The 'eps' value default is set to 1.e-6, which reproduces the same behavior as
+                            # the ITK numeric comparison logic (i.e. absolute difference).
+                            # Reference:
+                            #   Class: ITK/Modules/Numerics/Optimizersv4/include/itkObjectToObjectMetric.hxx
+                            #   Method: VerifyDisplacementFieldSizeAndPhysicalSpace
+                            #   URL: https://github.com/hinerm/ITK/blob/master/Modules/Numerics/Optimizersv4/include/itkObjectToObjectMetric.hxx#L471-L475.
+                            if numpy.allclose(vector, subseriesVector, rtol = 0., atol = eps):
```
jcfr, line 258:
```suggestion
                            if numpy.allclose(vector, subseriesVector, rtol=0.0, atol=eps):
```

While we are not yet using "ruff format" (or black), this will be the recommended style.

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
```diff
@@ -224,10 +237,34 @@ def examineFiles(self, files):
...
+                                found = True
+                                value = subseriesValue
+                                break
+
+                        if not found:
+                            subseriesValues[tag].append(value)
+                elif not subseriesValues[tag].__contains__(value):
                     subseriesValues[tag].append(value)
```
jcfr, line 266: To streamline the overall logic and reduce complexity, I suggest to revisit the approach by:
* adding a `TagValueList` custom type that would be a subclass of `list` with its `__contains__` function overridden to support comparing `np.ndarray`
* adding a function `tagValueToVector`

using something like this:

```python
        subseriesFiles = {}
        subseriesValues = {}
        for file in allFilesLoadable.files:
            # check for subseries values
            for tag in subseriesTags:
                if tag not in subseriesValues:
                    subseriesValues[tag] = TagValueList(epsilon=eps)

                value = slicer.dicomDatabase.fileValue(file, self.tags[tag])

                if tag in subseriesTagsToVectorValues:
                    value = tagValueToVector(value)
                else:
                    value = value.replace(",", "_")  # remove commas so it can be used as an index

                if not subseriesValues[tag].__contains__(value):
                    subseriesValues[tag].append(value)

                if (tag, value) not in subseriesFiles:
                    subseriesFiles[tag, value] = []
                subseriesFiles[tag, value].append(file)
```


with


```
def tagValueToVector(value):
    return numpy.array([float(element) for element in value.split("\\")])
```

and 

```python
class TagValueList(list):
    def __init__(self, epsilon):
        self.epsilon = epsilon
    def __contains__(self, value):
        if isinstance(value, numpy.ndarray):
            return numpy.allclose(vector, subseriesVector, rtol=0.0, atol=self.epsilon):
        return super().__contains__(value):
```

jcfr (changes requested):

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
Punzo: ok thanks!

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
Punzo: ok thanks for the feedback. I agree and I have implemented it in https://github.com/Slicer/Slicer/pull/7362/commits/3b7f8cbe25e39a50dc74c6e17529dcf1e56514ce
however I had to modify a little bit the design. Please have a look.

Punzo:
@jcfr thanks for the review. I have applied your comment, but please have a look at this [comment](https://github.com/Slicer/Slicer/pull/7362#discussion_r1388277594)  regarding the implementation

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
pieper: I'm not following - why are we using `__contains__` and not `in`?

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
```diff
@@ -22,6 +22,42 @@
...
+                #   Method: VerifyDisplacementFieldSizeAndPhysicalSpace
+                #   URL: https://github.com/hinerm/ITK/blob/master/Modules/Numerics/Optimizersv4/include/itkObjectToObjectMetric.hxx#L471-L475.
+                if numpy.allclose(valueVector, subseriesVector, rtol=0.0, atol=self.epsilon):
+                    return True, subseriesValue
+            return False, value
+
+        # The value is a string, use default list __contains__ method
+        value = value.replace(",", "_")  # remove commas so it can be used as an index
+        return super().__contains__(value), value
+
```
jcfr, line 60:
```suggestion


```

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
```diff
@@ -22,6 +22,42 @@
...
+                # Reference:
+                #   Class: ITK/Modules/Numerics/Optimizersv4/include/itkObjectToObjectMetric.hxx
+                #   Method: VerifyDisplacementFieldSizeAndPhysicalSpace
+                #   URL: https://github.com/hinerm/ITK/blob/master/Modules/Numerics/Optimizersv4/include/itkObjectToObjectMetric.hxx#L471-L475.
+                if numpy.allclose(valueVector, subseriesVector, rtol=0.0, atol=self.epsilon):
+                    return True, subseriesValue
+            return False, value
+
+        # The value is a string, use default list __contains__ method
+        value = value.replace(",", "_")  # remove commas so it can be used as an index
```
jcfr, line 58: The contains method should not return a modified value. As I originally suggested, I suggest to keep the update of the value on insertion.

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
jcfr: This should indeed be simplified and could be done in this PR or a follow-up one.

The use of `__contains__` was originally introduced in https://github.com/Slicer/Slicer/commit/f0b2690c328fc0cfc248907d35ab845c3e7c4f4d and was likely on oversight.

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
```diff
@@ -22,6 +22,42 @@
 # plugin architecture.
 #
 
+class TagValueList(list):
```
lassoan, line 25: This looks interesting, elegant, Pythonic but way too complicated for what it does. I usually prefer optimizing for simplicity and readability instead, because those are the most important in the long term. This code looks also potentially slow, which is quite important, because for a single study, these checks are done for thousands of frames, each frame contains several fields that are compared, and  a field may contain vectors with 6 values. I don't want to ruin this elegent Pythonic design, but if there are objective disadvantages (especially slow speed) then we may consider reverting to the previous more direct implementation.

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
pieper: > was likely on oversight.

Yes, that was me in the old days - good time to fix it now : )

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
pieper: yes, I thought I had made basically that same comment but maybe I didn't finish the comment before closing the browser tab.  I find the use of an extra class here clever, but I don't see it improving readability and I don't see any case were we would want to reuse this code.  The speed issue @lassoan raises is also a good point.

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
Punzo: the issue is that we need to return it for the reason explained in https://github.com/Slicer/Slicer/pull/7362#discussion_r1388277594

I can modify the method `__contains__` of `TagValueList` class  into `insert `or `add ` and move the [append ](https://github.com/Slicer/Slicer/pull/7362/files#diff-ddb2464152946e0fec0e7fcd1ec6cc4aefcd6b8a9ad3ee82a136564525f0c66fR272) into this method; then return the modified value,  **needed** for the [lines](https://github.com/Slicer/Slicer/commit/64b037eb264de0d6f87c4b0aa79e49bd399176ca#diff-ddb2464152946e0fec0e7fcd1ec6cc4aefcd6b8a9ad3ee82a136564525f0c66fL231-R269).

Not sure, the issue is that probably we would need a larger extensive refactoring

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
Punzo: @jcfr @lassoan since also the issues listed in  https://github.com/Slicer/Slicer/pull/7362#discussion_r1388508506 

for the moment I will revert to 
https://github.com/Slicer/Slicer/commit/64b037eb264de0d6f87c4b0aa79e49bd399176ca

Then up to you guys how to proceed

Punzo: done, please have a look at https://github.com/Slicer/Slicer/commit/59c63ff45c65291973fd4d343fc53d5a19c6ceaa

Punzo:
@jcfr @lassoan @pieper thank you very much for the reviews and sorry for the noise/spam. 

I have reverted the class implementation commits. Please have a look at the current commit https://github.com/Slicer/Slicer/commit/7ed3e2ebae6aad5381484f2e948e7f8a610a2504 and let me know if there is another issue.

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
```diff
@@ -224,9 +240,32 @@ def examineFiles(self, files):
...
+                        found = False
+                        for subseriesValue in subseriesValues[tag]:
+                            subseriesVector = self.tagValueToVector(subseriesValue)
+
+                            # The 'eps' value default is set to 1.e-6, which reproduces the same behavior as
+                            # the ITK numeric comparison logic (i.e. absolute difference).
+                            # Reference:
+                            #   Class: ITK/Modules/Numerics/Optimizersv4/include/itkObjectToObjectMetric.hxx
+                            #   Method: VerifyDisplacementFieldSizeAndPhysicalSpace
+                            #   URL: https://github.com/hinerm/ITK/blob/master/Modules/Numerics/Optimizersv4/include/itkObjectToObjectMetric.hxx#L471-L475.
```
jcfr, line 260:
```suggestion
                            #   URL: https://github.com/InsightSoftwareConsortium/ITK/blob/v5.4rc02/Modules/Numerics/Optimizersv4/include/itkObjectToObjectMetric.hxx#L507-L510.
```

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
jcfr: Considering that the `numpy.ndarray` is not a hashable type, this prevent us from indexing into the `subseriesFiles` map, the original implementation ends up being more straightforward.

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
```diff
@@ -171,10 +175,15 @@ def cleanNodeName(self, value):
...
-    def examineFiles(self, files):
+    def tagValueToVector(self, value):
+        return numpy.array([float(element) for element in value.split("\\")])
+
+    def examineFiles(self, files, eps=1.e-6):
         """ Returns a list of DICOMLoadable instances
         corresponding to ways of interpreting the
         files parameter.
+
+        eps value default to 1.e-6 as same behaviour of ITK numeric comparison logic.
```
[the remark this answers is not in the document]
jcfr: To be consistent with how other methods accept `epsilon` (for example `DICOMScalarVolumePluginClass.__init__`[^1]), could you instead use the full name `epsilon` ?

Note that the epsilon currently defined when instantiating `DICOMScalarVolumePluginClass` is the used when calling[^2] `DICOMUtils.getSortedImageFiles`[^3] in which we either indicate that a regularization transform may be applied or suggest to enable the corresponding settings.

For this reason, I suggest to simply reuse that `epsilon` and avoid re-introducing a specific one here.

[^2]: https://github.com/Slicer/Slicer/blob/ba80bddd4a6cba9cfb82a527309cda30ada2ba6a/Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py#L310C34-L310C34
[^3]: https://github.com/Slicer/Slicer/blob/ba80bddd4a6cba9cfb82a527309cda30ada2ba6a/Modules/Scripted/DICOMLib/DICOMUtils.py#L705-L713
[^1]: https://github.com/Slicer/Slicer/blob/ba80bddd4a6cba9cfb82a527309cda30ada2ba6a/Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py#L29

jcfr (commented):
Suggested commit message:

```

File: Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py [hunk 20]
```diff
@@ -212,6 +221,13 @@ def examineFiles(self, files):
             "diffusionGradientOrientation",
         ]
 
+        # Duplicate values for these tags will be compared numerically,
+        # instead of checking the values of the dictionary subseriesValues stored as strings.
```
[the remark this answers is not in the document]
Punzo: thanks, commited

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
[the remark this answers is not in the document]
Punzo: > To be consistent with how other methods accept `epsilon` (for example `DICOMScalarVolumePluginClass.__init__`[1](#user-content-fn-1-dc1aa5f8599ae39ad0baef30ca8f7661)), could you instead use the full name `epsilon` ?
> 
> Note that the epsilon currently defined when instantiating `DICOMScalarVolumePluginClass` is the used when calling[2](#user-content-fn-2-dc1aa5f8599ae39ad0baef30ca8f7661) `DICOMUtils.getSortedImageFiles`[3](#user-content-fn-3-dc1aa5f8599ae39ad0baef30ca8f7661) in which we either indicate that a regularization transform may be applied or suggest to enable the corresponding settings.
> 
> For this reason, I suggest to simply reuse that `epsilon` and avoid re-introducing a specific one here.
> 
> ## Footnotes
> 1. https://github.com/Slicer/Slicer/blob/ba80bddd4a6cba9cfb82a527309cda30ada2ba6a/Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py#L29 [↩](#user-content-fnref-1-dc1aa5f8599ae39ad0baef30ca8f7661)
> 2. https://github.com/Slicer/Slicer/blob/ba80bddd4a6cba9cfb82a527309cda30ada2ba6a/Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py#L310C34-L310C34 [↩](#user-content-fnref-2-dc1aa5f8599ae39ad0baef30ca8f7661)
> 3. https://github.com/Slicer/Slicer/blob/ba80bddd4a6cba9cfb82a527309cda30ada2ba6a/Modules/Scripted/DICOMLib/DICOMUtils.py#L705-L713 [↩](#user-content-fnref-3-dc1aa5f8599ae39ad0baef30ca8f7661)

I agree on this. the only issue would be that the default epsilon in https://github.com/Slicer/Slicer/pull/7362/files#diff-ddb2464152946e0fec0e7fcd1ec6cc4aefcd6b8a9ad3ee82a136564525f0c66fL29 is 10^-2 . While here we are using https://github.com/Slicer/Slicer/pull/7362/files#diff-ddb2464152946e0fec0e7fcd1ec6cc4aefcd6b8a9ad3ee82a136564525f0c66fR181 10^-6.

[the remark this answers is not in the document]
Punzo: we could take self.eplison and muliutply it for a factor. I don;t have strong preference on this, but the default value should be 10^-6. 10^-2 is too large.

Punzo:
> Suggested commit message:
> 
> ```
> ENH: Add DICOM grouping frames by orientation and add harden transform option
> 
> Update the ScalarVolume DICOM plugin adding:
> * subseries grouping by orientation 
> * harden transform option in the DICOM settings.
> ```

@jcfr  ok, tomorrow I will squash and change the commit message, thanks!

regarding the eps vs self.epsilon, I don't have strong preferences. However the default value of self.epsilon is 0.01. While we need a default value of 1.e-6 for the orientation check. We could multiply self.epsilon for a factor  in `examineFiles`. Up to you.

jcfr:
To improve on https://github.com/Slicer/Slicer/pull/7362#pullrequestreview-1723605026:

```
ENH: Support regularization transform hardening in DICOM Scalar plugin

Update the logic identifying sub-series (each loaded as independent
volume node) to support grouping by orientation and add harden a new
DICOM settings option to enable automatic hardening of the corresponding
regularization transform.
```

I think this better describe what was implemented and help understand why.

> re: Documentation

Consider also updating:

https://github.com/Slicer/Slicer/blob/ba80bddd4a6cba9cfb82a527309cda30ada2ba6a/Docs/user_guide/modules/dicom.md#L184-L187

jcfr:
> re: epsilon

If the epsilon used in `DICOMUtils.getSortedImageFiles` and the one introduced in the `examineFile` are indeed different, we should name the prefix the parameter like we have done for other epsilon (`cornerEpsilon`, `zeroEpsilon`, ...)

pieper:
> should name the prefix the parameter like we have done for other epsilon

Agreed!

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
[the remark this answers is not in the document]
Punzo: done in https://github.com/Slicer/Slicer/pull/7362/commits/44a503fe000025f3777d064c7735ff0909141068

Punzo:
@jcfr I have applied last round of feedback!

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
```diff
@@ -224,9 +242,32 @@ def examineFiles(self, files):
...
+
+                if tag in vectorTags:
+                    if value != "":
+                        vector = self.tagValueToVector(value)
+
+                        found = False
+                        for subseriesValue in subseriesValues[tag]:
+                            subseriesVector = self.tagValueToVector(subseriesValue)
+
+                            # The 'self.orientationEpsilon' value default is set to 1.e-6, which reproduces the same behavior as
```
lassoan, line 257: Movr this to the top where we define the ne the 1e-6 value

lassoan (commented):
Thank you all for the reviews, this looks good to me. I just added a remark on improving comments.

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
```diff
@@ -26,10 +26,11 @@ class DICOMScalarVolumePluginClass(DICOMPlugin):
     """ ScalarVolume specific interpretation code
     """
 
-    def __init__(self, epsilon=0.01):
+    def __init__(self, spacingEpsilon=1e-2, orientationEpsilon=1e-6):
```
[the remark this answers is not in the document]
Punzo: added in https://github.com/Slicer/Slicer/pull/7362/commits/3c2da06844dc014a62795d35e14c8787fd0dfa42

Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py
Punzo: done in https://github.com/Slicer/Slicer/pull/7362/commits/3c2da06844dc014a62795d35e14c8787fd0dfa42

Punzo:
> Thank you all for the reviews, this looks good to me. I just added a remark on improving comments.

Thanks Andras for the feedback. comments applied!

jcfr (approved):
Pending local testing is successful :rocket:

jcfr:
@Punzo Could you confirm that you performed local testing using the latest version ?

jcfr:
Once this is integrated:
* Add a short discourse post describing the feature (associated with the  `feature`[^1] tag)
* Add a not to the following post cross-referencing this post:
   * https://discourse.slicer.org/t/facing-issue-with-using-volume-rendering/32695
   * https://discourse.slicer.org/t/dicom-file-loading-spacing/5469/4

There are probably other posts that could be updated. I used the query `harden regularization`[^2]


[^1]: https://discourse.slicer.org/tag/feature
[^2]: https://discourse.slicer.org/search?q=harden%20regularization%20order%3Alatest

Punzo:
> @Punzo Could you confirm that you performed local testing using the latest version ?

@jcfr yes I tested locally with several DICOM datasets.

Punzo:
> Once this is integrated:
> 
> * Add a short discourse post describing the feature (associated with the  `feature`[1](#user-content-fn-1-e6274d790bdad8d53b669e0687b7f24b) tag)
> * Add a note to the following post cross-referencing this feature post:
>   
>   * https://discourse.slicer.org/t/facing-issue-with-using-volume-rendering/32695
>   * https://discourse.slicer.org/t/dicom-file-loading-spacing/5469/4
> 
> There are probably other posts that could be updated. I used the query `harden regularization`[2](#user-content-fn-2-e6274d790bdad8d53b669e0687b7f24b)
> 
> ## Footnotes
> 1. https://discourse.slicer.org/tag/feature [↩](#user-content-fnref-1-e6274d790bdad8d53b669e0687b7f24b)
> 2. https://discourse.slicer.org/search?q=harden%20regularization%20order%3Alatest [↩](#user-content-fnref-2-e6274d790bdad8d53b669e0687b7f24b)

ok!

Punzo:
> Once this is integrated:
> 
> * Add a short discourse post describing the feature (associated with the  `feature`[1](#user-content-fn-1-e6274d790bdad8d53b669e0687b7f24b) tag)
> * Add a note to the following post cross-referencing this feature post:
>   
>   * https://discourse.slicer.org/t/facing-issue-with-using-volume-rendering/32695
>   * https://discourse.slicer.org/t/dicom-file-loading-spacing/5469/4
> 
> There are probably other posts that could be updated. I used the query `harden regularization`[2](#user-content-fn-2-e6274d790bdad8d53b669e0687b7f24b)
> 
> ## Footnotes
> 1. https://discourse.slicer.org/tag/feature [↩](#user-content-fnref-1-e6274d790bdad8d53b669e0687b7f24b)
> 2. https://discourse.slicer.org/search?q=harden%20regularization%20order%3Alatest [↩](#user-content-fnref-2-e6274d790bdad8d53b669e0687b7f24b)

Done

PR #7366, opened 2023-11-09
BUG: Fix crash for debug build due to invalid assert in MRMLViewInteractorStyle

HarryDC:
Issue is only visible in debug mode as `assert` is not in release build

In debug mode the assertion fails preventing slicer from starting up

### DIFF
--- a/Libs/MRML/DisplayableManager/vtkMRMLViewInteractorStyle.cxx
+++ b/Libs/MRML/DisplayableManager/vtkMRMLViewInteractorStyle.cxx
@@ -482,7 +482,7 @@ void vtkMRMLViewInteractorStyle::SetInteractor(vtkRenderWindowInteractor *intera
 void vtkMRMLViewInteractorStyle::DisplayableManagerCallback(vtkObject *object, unsigned long event, void *clientData, void *callData)
 {
   vtkMRMLViewInteractorStyle* self = reinterpret_cast<vtkMRMLViewInteractorStyle *>(clientData);
-  assert(!object->IsA("vtkMRMLAbstractDisplayableManager"));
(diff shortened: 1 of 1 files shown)


jcfr:
Good catch and thanks for creating the pull request.

Could you update the commit message to:

```BUG: Fix crash for debug build due to invalid assert

Fix regression introduced in 39c5a0877 (ENH: Set MRML interactor style
camera node observing displayable manager )
```

jcfr:
For context, the code was introduced in https://github.com/Slicer/Slicer/commit/39c5a0877bbdbb449ae6468e7aa12204489431e1 through https://github.com/Slicer/Slicer/pull/7324 by adapting logic from `vtkMRMLViewDisplayableManager.cxx` where there is also an assert (introduced 13 years ago in https://github.com/Slicer/Slicer/commit/dda203142bbc5d5c6c15ffb92bf77a00b72ef3bc)

jcfr:
Pull request will be integrated through `Squash & Merge` with the following commit message:

```
BUG: Fix crash for debug build due to invalid assert in MRMLViewInteractorStyle

Fix regression introduced in 39c5a0877 (ENH: Set MRML interactor style camera
node observing displayable manager )
```

HarryDC:
@jcfr I was suprised to see a bare `assert` wrapping asserts to trigger in release and debug seems to be a better pattern

jcfr:
Instead of using assert in this situation, we will remove them. asserts should be avoided.

jcfr (approved):

lassoan:
Yes, please remove bare asserts - crashing is not a good mechanism to indicate error in a large GUI application.

PR #7371, opened 2023-11-09
DOC: Add instructions for setting up development environment on Ubuntu 23.04

mauriliogenovese:
add instruction for lunar lobster

### DIFF
--- a/Docs/developer_guide/build_instructions/linux.md
+++ b/Docs/developer_guide/build_instructions/linux.md
@@ -50,6 +50,25 @@ The CMake version currently included in Debian 12 Bookworm (Stable) is not compa
 For more details, see the Slicer [CMakeLists.txt](https://github.com/Slicer/Slicer/blob/98c092edb8f5a274277d2e486a4f7e584f58605e/CMakeLists.txt#L3-L5) file. On Debian 12 Bookworm (Stable), you will need to upgrade CMake manually by downloading CMake 3.25.3 or higher from the [CMake website](https://cmake.org/download/) and following the CMake installation instructions.
 :::
 
+### Ubuntu 23.04 (Lunar Lobster)
+
+Install the development tools and the support libraries:
+
+```console
+sudo apt update && sudo apt install git git-lfs subversion build-essential \
+  qtmultimedia5-dev qttools5-dev libqt5xmlpatterns5-dev libqt5svg5-dev qtwebengine5-dev qtscript5-dev \
+  qtbase5-private-dev libqt5x11extras5-dev libxt-dev
+```
+
+Install CMake manually by downloading CMake 3.25.3 or higher from the [CMake website](https://cmake.org/download/)
+and by following the CMake installation instructions.
+
+:::{note}
+The CMake version currently included in Ubuntu 23.04 is CMake 3.25.1 (see [here](https://packages.ubuntu.com/lunar/cmake))
+and is not compatible with the current development version of Slicer.
+For more details, see the Slicer [CMakeLists.txt](https://github.com/Slicer/Slicer/blob/98c092edb8f5a274277d2e486a4f7e584f58605e/CMakeLists.txt#L3-L5) file.
+:::
+
 ### Ubuntu 21.10 (Impish Indri)
 
 Install the development tools and the support libraries:


mauriliogenovese:
instruction for ubuntu lunar lobster

Docs/developer_guide/build_instructions/linux.md
```diff
@@ -76,6 +76,41 @@ sudo apt update && sudo apt install git subversion build-essential cmake cmake-c
...
+tar xf cmake-3.25.3.tar.gz
+
+cd cmake-3.25.3
+./configure
+make
+# Then to install it to your system finally do 
+sudo make install 
+
+# Do this to refresh your console 
+hash -r
```
pieper, line 99: Are these really needed?  It's not even clear to me what they do.

Docs/developer_guide/build_instructions/linux.md
```diff
@@ -76,6 +76,41 @@ sudo apt update && sudo apt install git subversion build-essential cmake cmake-c
   qtbase5-private-dev libqt5x11extras5-dev libxt-dev
 ```
 
+### Ubuntu 23.04 (Lunar Lobster)
+
+:::{warning}
+If you get cmake version error, download build and install cmake 3.25.3:
```
pieper, line 82: I like the idea of getting people the right CMake is good - maybe these instructions below in a separate section about CMake that applies to more than just this ubuntu version, perhaps with notes about how to install the latest CMake on multiple OSes.

Docs/developer_guide/build_instructions/linux.md
pieper: Or maybe it's a typo?

pieper (changes requested):
Thanks for contributing this - just some comments and questions.

Docs/developer_guide/build_instructions/linux.md
```diff
@@ -76,6 +76,41 @@ sudo apt update && sudo apt install git subversion build-essential cmake cmake-c
...
+
+# Then finally check with
+cmake --version 
+```
+:::
+
+Install the development tools and the support libraries:
+
+```console
+sudo apt update && sudo apt install git git-lfs subversion build-essential cmake cmake-curses-gui cmake-qt-gui \
```
jcfr, line 109: Listing `cmake` here and building it above seems to conflict.

Considering that the version of CMake provided by `Ubuntu 23.04` is `3.25.1`[^1] and that the minimum required version by Slicer is `3.16.3`[^2], uninstalling the version provided by the operating system is probably not needed.

[^1]: https://packages.ubuntu.com/lunar/cmake
[^2]: https://github.com/Slicer/Slicer/blob/ba80bddd4a6cba9cfb82a527309cda30ada2ba6a/CMakeLists.txt#L1

jcfr (changes requested):
@mauriliogenovese  Thanks your help for both testing the build process on newer Ubuntu version and improving the build instructions :pray: :100: 

I suggest to revisit the instruction to either build or install the `cmake` package.

If `cmake` needs to be built, we need to explain why and/or update the Slicer build-system to address related issue (if any).

In case we need to update the Slicer build-system, we would be happy to help

Docs/developer_guide/build_instructions/linux.md
mauriliogenovese: Yes it's a typo for bash, maybe we can skip that

Docs/developer_guide/build_instructions/linux.md
mauriliogenovese: When I tried to use the installed cmake I got an error saying that some version were not accepted (including 3.25.1). I have not copied that error but I had to install a newer  release

jcfr: That is helpful details.

To move forward with this, could you add a section like the following ?

https://github.com/Slicer/Slicer/blob/ba80bddd4a6cba9cfb82a527309cda30ada2ba6a/Docs/developer_guide/build_instructions/linux.md#L48-L51

------

For context, it turns out that CMake `3.25.1` had a regression[^1] that was fixed in version `3.25.2` and above. This is why the default version can't be used.

[^1]: https://gitlab.kitware.com/cmake/cmake/-/issues/24567

Docs/developer_guide/build_instructions/linux.md
jcfr, line 99:
```suggestion
```

Removing this section and including a note like we did for `Debian 12 Bookworm (Stable) and Bullseye 11 (OldStable)` should be sufficient.

jcfr (approved):
I revisited the patch based on the suggestions posted at https://github.com/Slicer/Slicer/pull/7371#discussion_r1389719866

For a preview of the change:

| https://slicer--7371.org.readthedocs.build/en/7371/developer_guide/build_instructions/linux.html#ubuntu-23-04-lunar-lobster |
|--|
| image |

pieper (approved):
LGTM 👍

PR #7372, opened 2023-11-09
STYLE: Simplify list membership test in DICOMScalarVolumePlugin

jcfr:
Following-up of f0b2690c3 (ENH: Add support for ctkDICOMDatabase tag precaching) where the original use of "__contains__" was introduced.

See https://docs.python.org/3/reference/expressions.html#in


pieper (approved):
Thanks!

Issue #7363, opened 2023-11-08
`FiberBundleFile` and `ModelFile` data type loading may not be working correctly


jhlegarreta:
## Summary

Following a number of test failures in the `SlicerDMRI` extension, and after digging a little bit more, it turns out the `FiberBundleFile` and `ModelFile` data types are not working properly:
- The former makes the tests crash with a rather uninformative message:
```
error: [/Users/svc-dashboard/D/S/A/Slicer-build/bin/Slicer.app/Contents/MacOS/./Slicer] exit abnormally - Report the problem.
```
This happens at the `SampleData.downloadFromURL` method, which ultimately relies on `SampleDataLogic.downloadFromSource` to load the data. This method does not seem to cover the `FiberBundleFile` and `ModelFile` cases.

- However, in the latter case, if the data type is changed to `ModelFile`, then the data gets loaded into the scene, but the node is not being found immediately after, e.g.:
```
  File "bld/slicerdmri/inner-build/lib/Slicer-5.5/qt-scripted-modules/TractographyDownsample.py", line 653, in test_TractographyDownsample1
    fiberBundleNode = slicer.util.getNode(pattern="fiber_ply_export_test")
  File "bld/slicer/Slicer-build/bin/Python/slicer/util.py", line 1517, in getNode
    raise MRMLNodeNotFoundException("could not find nodes in the scene by name or id '%s'" % (pattern if (isinstance(pattern, str)) else ""))
slicer.util.MRMLNodeNotFoundException: could not find nodes in the scene by name or id 'fiber_ply_export_test'
```

Similar error messages can be found across the failing tests in the 3D Slicer dashboard:
https://slicer.cdash.org/viewTest.php?onlyfailed&buildid=3203300

Some thoughts:
- Andras mentioned that the [reader(s) at issue may be broken in `SlicerDMRI`](https://github.com/Slicer/Slicer/issues/7342#issuecomment-1793967995). However, as said above, the failing methods ultimately rely on 3D Slicer proper methods to load the data, so the failure stems from `3D Slicer` proper. Andras reported crashes of the `FiberBundleFile` data type in 3D Slicer versions as recent as 5.4.0, and 5.5. This matches the dashboard and local tests, where the `HEAD` (dashboard)/a recent (locally) commit is being used. Also, a search in the extension does not provide meaningful results about the readers being present in the SlicerDMRI module.
- A potential workaround may be to use the `slicer.util` module methods to do this:
[...]

PR #7375, opened 2023-11-10
ENH: Add test for the `slicer.util.loadFiberBundle` method

jhlegarreta:
- ENH: Check for extension and registered reader for fiber bundle data
- ENH: Add test for the `slicer.util.loadFiberBundle` method

### DIFF
--- a/Base/Python/slicer/tests/test_slicer_util_load_data.py
+++ b/Base/Python/slicer/tests/test_slicer_util_load_data.py
@@ -13,12 +13,23 @@ class SlicerUtilLoadDataTests(unittest.TestCase):
 
     def test_loadFiberBundle(self):
 
-        def _load(fiberBundleFilePath):
+        def _checkNode(node, returnNode):
 
-            node = slicer.util.loadFiberBundle(fiberBundleFilePath)
+            if returnNode:
+                # Deprecated
+                self.assertIsInstance(node, tuple)
+                self.assertEqual(len(node), 2)
+                self.assertTrue(node[0])
+                node = node[1]
             self.assertIsInstance(node, slicer.vtkMRMLFiberBundleNode)
-            node = slicer.util.loadFiberBundle(fiberBundleFilePath)
-            self.assertIsInstance(node, slicer.vtkMRMLFiberBundleNode)
+
+        def _load(fiberBundleFilePath, returnNode=False):
+
+            node = slicer.util.loadFiberBundle(fiberBundleFilePath, returnNode=returnNode)
+            _checkNode(node, returnNode)
+
+            node = slicer.util.loadFiberBundle(fiberBundleFilePath, returnNode=returnNode)
+            _checkNode(node, returnNode)
 
         hash_algorithm = "SHA256"
         checksum = "e1848e4d0b90d6c7442b2474c513d0464c6f141253c25aee6f5bcda4c828deac"
@@ -35,5 +46,13 @@ class SlicerUtilLoadDataTests(unittest.TestCase):
             if slicer.app.ioManager().registeredFileReaderCount("FiberBundleFile") == 0:
                 with self.assertRaisesRegex(RuntimeError, "FiberBundleFile reader not registered") as cm:
                     _load(fiberBundleFilePath)
+
+                with self.assertRaisesRegex(RuntimeError, "FiberBundleFile reader not registered") as cm:
+                    # Deprecated
+                    _load(fiberBundleFilePath, returnNode=True)
+
             else:
                 _load(fiberBundleFilePath)
+
+                # Deprecated
+                _load(fiberBundleFilePath, returnNode=True)
--- a/Base/Python/slicer/tests/test_slicer_util_load_data.py
+++ b/Base/Python/slicer/tests/test_slicer_util_load_data.py
@@ -9,16 +9,13 @@ import slicer
 class SlicerUtilLoadDataTests(unittest.TestCase):
 
     def setUp(self):
-
         pass
-
 
     def test_loadFiberBundle(self):
 
         def _load(fiberBundleFilePath):
 
-            success, node = slicer.util.loadFiberBundle(fiberBundleFilePath, returnNode=True)
-            self.assertTrue(success)
+            node = slicer.util.loadFiberBundle(fiberBundleFilePath)
             self.assertIsInstance(node, slicer.vtkMRMLFiberBundleNode)
             node = slicer.util.loadFiberBundle(fiberBundleFilePath)
             self.assertIsInstance(node, slicer.vtkMRMLFiberBundleNode)
--- a/Base/Python/slicer/tests/test_slicer_util_load_data.py
+++ b/Base/Python/slicer/tests/test_slicer_util_load_data.py
@@ -1,0 +1,42 @@
+import tempfile
+import unittest
+from pathlib import Path
+from urllib.parse import urljoin
+
+import slicer
+
+
+class SlicerUtilLoadDataTests(unittest.TestCase):
+
+    def setUp(self):
+
+        pass
+
+
+    def test_loadFiberBundle(self):
+
+        def _load(fiberBundleFilePath):
+
+            success, node = slicer.util.loadFiberBundle(fiberBundleFilePath, returnNode=True)
+            self.assertTrue(success)
+            self.assertIsInstance(node, slicer.vtkMRMLFiberBundleNode)
+            node = slicer.util.loadFiberBundle(fiberBundleFilePath)
+            self.assertIsInstance(node, slicer.vtkMRMLFiberBundleNode)
+
+        hash_algorithm = "SHA256"
+        checksum = "e1848e4d0b90d6c7442b2474c513d0464c6f141253c25aee6f5bcda4c828deac"
+        url = urljoin(slicer.util.TESTING_DATA_URL, f"{hash_algorithm}/{checksum}")
+
+        with tempfile.TemporaryDirectory() as tmpdirname:
+
+            file_basename = "tractography.vtk"
+            fiberBundleFilePath = Path(tmpdirname).joinpath(file_basename)
+            downloadSuccess = slicer.util.downloadFile(
+                url, fiberBundleFilePath, checksum=f"{hash_algorithm}:{checksum}", reDownloadIfChecksumInvalid=True)
+            self.assertTrue(downloadSuccess)
+
+            if slicer.app.ioManager().registeredFileReaderCount("FiberBundleFile") == 0:
+                with self.assertRaisesRegex(RuntimeError, "FiberBundleFile reader not registered") as cm:
+                    _load(fiberBundleFilePath)
+            else:
+                _load(fiberBundleFilePath)
--- a/Base/Python/slicer/util.py
+++ b/Base/Python/slicer/util.py
@@ -779,36 +779,53 @@ def loadColorTable(filename, returnNode=False):
     return loadNodeFromFile(filename, "ColorTableFile", {}, returnNode)
 
 
-def loadFiberBundle(filename, returnNode=False):
+def loadFiberBundle(filename):
     """Load fiber bundle node from file.
 
     .. warning::
-    To ensure the FiberBundleFile reader is registered, the ``SlicerDMRI``
-    extension needs to be installed.
+
+      To ensure the FiberBundleFile reader is registered, the ``SlicerDMRI``
+      extension may need to be installed.
 
     :param filename: full path of the file to load.
-    :param returnNode: Deprecated.
-    :return: loaded node (if multiple nodes are loaded then a list of nodes).
-      If returnNode is True then a status flag and loaded node are returned.
-    """
 
-    # Check if SlicerDMRI is installed
-    extension_name = "SlicerDMRI"
-    em = slicer.app.extensionsManagerModel()
-    if not em.isExtensionInstalled(extension_name):
-        errorMessage = f"{extension_name} not installed."
-        errorMessage += "\n" + f"Install the {extension_name} extension"
+    :return: loaded node (if multiple nodes are loaded then a list of nodes).
+
+    :raises RuntimeError: in case of failure
+    """
+    from slicer import app
+
+    readerType = "FiberBundleFile"
 
     # Check if the appropriate reader is registered
-    if slicer.app.ioManager().registeredFileReaderCount("FiberBundleFile") == 0:
-        errorMessage += f"FiberBundleFile reader not registered: failed to read data from file: {filename}"
-        errorMessage += "\n" + f"Try reinstalling the {extension_name} extension"
+    if app.ioManager().registeredFileReaderCount(readerType) == 0:
+        errorMessage = f"{readerType} reader not registered: Failed to load node from file: {filename}"
 
-    if userMessages.GetNumberOfMessages() > 0:
-        errorMessage += "\n" + userMessages.GetAllMessagesAsString()
+        extensionName = "SlicerDMRI"
+        moduleName = "TractographyDisplay"
+        readerClassName = "qSlicerFiberBundleReader"
+
+        errorMessage += (
+            f"\n\n{readerType} reader is implemented in the {readerClassName} class expected to be "
+            f"registered by the {moduleName} module provided by the {extensionName} extension."
+            "\n\nStatus:"
+        )
+
+        if app.moduleManager().module(moduleName) is None:
+            errorMessage += f"\n- {moduleName} module: Not loaded."
+
+        if app.applicationName == "Slicer":
+            # Check if the extension is installed
+            em = app.extensionsManagerModel()
+            if not em.isExtensionInstalled(extensionName):
+                errorMessage += (
+                    f"\n- {extensionName} extension: Not installed."
+                    "\n\nPlease install the extension for proper functionality."
+                )
+
         raise RuntimeError(errorMessage)
 
-    return loadNodeFromFile(filename, "FiberBundleFile", {}, returnNode)
+    return loadNodeFromFile(filename, readerType, {})
 
 
 def loadAnnotationFiducial(filename, returnNode=False):
--- a/Base/Python/slicer/util.py
+++ b/Base/Python/slicer/util.py
@@ -780,13 +780,34 @@ def loadColorTable(filename, returnNode=False):
 
 
 def loadFiberBundle(filename, returnNode=False):
-    """Load node from file.
+    """Load fiber bundle node from file.
+
+    .. warning::
+    To ensure the FiberBundleFile reader is registered, the ``SlicerDMRI``
+    extension needs to be installed.
 
     :param filename: full path of the file to load.
     :param returnNode: Deprecated.
     :return: loaded node (if multiple nodes are loaded then a list of nodes).
       If returnNode is True then a status flag and loaded node are returned.
     """
+
+    # Check if SlicerDMRI is installed
+    extension_name = "SlicerDMRI"
+    em = slicer.app.extensionsManagerModel()
+    if not em.isExtensionInstalled(extension_name):
+        errorMessage = f"{extension_name} not installed."
+        errorMessage += "\n" + f"Install the {extension_name} extension"
+
+    # Check if the appropriate reader is registered
+    if slicer.app.ioManager().registeredFileReaderCount("FiberBundleFile") == 0:
+        errorMessage += f"FiberBundleFile reader not registered: failed to read data from file: {filename}"
+        errorMessage += "\n" + f"Try reinstalling the {extension_name} extension"
+
+    if userMessages.GetNumberOfMessages() > 0:
+        errorMessage += "\n" + userMessages.GetAllMessagesAsString()
+        raise RuntimeError(errorMessage)
+
     return loadNodeFromFile(filename, "FiberBundleFile", {}, returnNode)
 
 
--- a/Base/Python/slicer/util.py
+++ b/Base/Python/slicer/util.py
@@ -779,7 +779,7 @@ def loadColorTable(filename, returnNode=False):
     return loadNodeFromFile(filename, "ColorTableFile", {}, returnNode)
 
 
-def loadFiberBundle(filename):
+def loadFiberBundle(filename, returnNode=False):
     """Load fiber bundle node from file.
 
     .. warning::
@@ -788,8 +788,10 @@ def loadFiberBundle(filename):
       extension may need to be installed.
 
     :param filename: full path of the file to load.
+    :param returnNode: Deprecated.
 
     :return: loaded node (if multiple nodes are loaded then a list of nodes).
+      If returnNode is True then a status flag and loaded node are returned.
 
     :raises RuntimeError: in case of failure
     """
@@ -825,7 +827,7 @@ def loadFiberBundle(filename):
 
         raise RuntimeError(errorMessage)
 
-    return loadNodeFromFile(filename, readerType, {})
+    return loadNodeFromFile(filename, readerType, {}, returnNode)
 
 
 def loadAnnotationFiducial(filename, returnNode=False):
--- a/Applications/SlicerApp/Testing/Python/CMakeLists.txt
+++ b/Applications/SlicerApp/Testing/Python/CMakeLists.txt
@@ -296,6 +296,12 @@ slicer_add_python_unittest(
   )
 
 slicer_add_python_unittest(
+  SCRIPT ${Slicer_SOURCE_DIR}/Base/Python/slicer/tests/test_slicer_util_load_data.py
+  SLICER_ARGS --no-main-window --disable-cli-modules
+  TESTNAME_PREFIX nomainwindow_
+  )
+
+slicer_add_python_unittest(
   SCRIPT ${Slicer_SOURCE_DIR}/Base/Python/slicer/tests/test_slicer_util_save.py
   SLICER_ARGS --no-main-window --disable-cli-modules --disable-scripted-loadable-modules DATA{${INPUT}/MR-head.nrrd}
   TESTNAME_PREFIX nomainwindow_


pieper (approved):
There's a lint issue but otherwise it looks good to me.

jamesobutler:
Should `slicer.util.loadFiberBundle` even still be in Slicer core if the FiberBundle reader is provided by an extension? Shouldn't this method have been removed when Fiber functionality was removed from Slicer core and put into SlicerDMRI?

lassoan:
Good point @jamesobutler - we should replace the function in Slicer core by a function that prints a descriptive error message (need to install SlicerDMRI). And when SlicerDMRI is installed it can replace the method by one that actually reads fiber bundle. Test should be in the extension.

jhlegarreta:
> Should slicer.util.loadFiberBundle even still be in Slicer core if the FiberBundle reader is provided by an extension? Shouldn't this method have been removed when Fiber functionality was removed from Slicer core and put into SlicerDMRI?

I do not know James: I ignore why both exist or why they were added/developed separately, or whether the one in Slicer has been actually used at some point. I just want to avoid any update to 3D Slicer having detrimental consequences on SlicerDMRI, re https://github.com/Slicer/Slicer/issues/7363#issuecomment-1803156288.

IMO, all IO capabilities should be provided by a separate package or module, so that all reading/writing are readily available across all 3D Slicer, and all extensions.

> Test should be in the extension.

https://github.com/SlicerDMRI/SlicerDMRI/pull/210

pieper:
There's a long history of this, since DMRI was once entirely in the core and at the time it was factored out into an extension there were some reasons for some things to remain in the core.  If anyone wants to take another stab at completing the refactoring please jump in.

jcfr:
> re: augment slicer.util from extension

Since are already extending `slicer.modules` through monkey patching (instantiated modules are reachable through that special package), it woupd make sense to extend `slicer.util` in a similar way.

That said, we should think about a systematic way and not just let extension updates `slicer.util`.


For example, function added to the `__all__` list of a package named like `<ModuleName>Lib/util.py` could automatically be added to `slicer.util` (while checking these do not overide existing Slicer function unless allowed)


Related link:
* https://softwareengineering.stackexchange.com/questions/443815/is-adding-functions-to-a-namespace-module-after-imports-an-anti-pattern

Base/Python/slicer/tests/test_slicer_util_load_data.py
```diff
@@ -0,0 +1,33 @@
...
+
+    def setUp(self):
+
+        pass
+
+    def test_loadFiberBundle(self):
+
+        hash_algorithm = "SHA256"
+        checksum = "e1848e4d0b90d6c7442b2474c513d0464c6f141253c25aee6f5bcda4c828deac"
+        url = urljoin(slicer.util.TESTING_DATA_URL, hash_algorithm, checksum)
```
lassoan, line 19:
```suggestion
        url = urljoin(slicer.util.TESTING_DATA_URL, f"{hash_algorithm}/{checksum}")
```

Base/Python/slicer/tests/test_slicer_util_load_data.py
```diff
@@ -0,0 +1,33 @@
...
+
+        with tempfile.TemporaryDirectory() as tmpdirname:
+          file_basename = "tractography.vtk"
+          targetFilePath = Path(tmpdirname).joinpath(file_basename)
+          reDownloadIfChecksumInvalid = True
+          slicer.util.downloadFile(url, targetFilePath, checksum=checksum, reDownloadIfChecksumInvalid=reDownloadIfChecksumInvalid)
```
lassoan, line 25:
```suggestion
          downloadSuccess = slicer.util.downloadFile(url, targetFilePath, checksum=f"{hash_algorithm}:{checksum}", reDownloadIfChecksumInvalid=reDownloadIfChecksumInvalid)
          self.assertTrue(downloadSuccess)
```

Base/Python/slicer/tests/test_slicer_util_load_data.py
```diff
@@ -0,0 +1,33 @@
...
+
+          returnNode = False
```
lassoan, line 27:
```suggestion
```

Base/Python/slicer/tests/test_slicer_util_load_data.py
```diff
@@ -0,0 +1,33 @@
...
+          node = slicer.util.loadFiberBundle(filename, returnNode)
```
lassoan, line 28: `returnNode` is deprecated and `False` is the default value, so it is better not to explicitly add it.

```suggestion
          node = slicer.util.loadFiberBundle(targetFilePath)
```

Base/Python/slicer/tests/test_slicer_util_load_data.py
```diff
@@ -0,0 +1,33 @@
...
+          self.assertFalse(node)
```
lassoan, line 29: If you expect this to succeed then you need to write something like this (node is an object or `None`, we should not cast it to Boolean in the test):

```suggestion
          self.assertIsNotNone(node)
```

If there is a failure then the function throws an exception, so the test will fail, so `assertFalse` is not useful.

Base/Python/slicer/tests/test_slicer_util_load_data.py
```diff
@@ -0,0 +1,33 @@
...
+
+          returnNode = True
```
lassoan, line 31:
```suggestion
```

Base/Python/slicer/tests/test_slicer_util_load_data.py
```diff
@@ -0,0 +1,33 @@
...
+          node = slicer.util.loadFiberBundle(filename, returnNode)
```
lassoan, line 32:
```suggestion
          success, node = slicer.util.loadFiberBundle(filename, returnNode = True)
          self.assertTrue(success)
```

Base/Python/slicer/tests/test_slicer_util_load_data.py
```diff
@@ -0,0 +1,33 @@
...
+          self.assertTrue(node)
```
lassoan, line 33:
```suggestion
          self.assertIsNotNone(node)
```

lassoan (changes requested):
I've tried this and there were many errors with the current Slicer master.
I've suggested changes, which should fix the errors as long as the SlicerDMRI extension is installed. So, I would recommend to move this test into SlicerDMRI extension.

If we add a test into Slicer core then it should be modified so that it passes if reading fails.

It would be also nice to update `slicer.util.loadFiberBundle` function documentation to explain that SlicerDMRI is required and update the function itself to raise an exception that explains that SlicerDMRI is required.

jhlegarreta:
Re https://github.com/Slicer/Slicer/pull/7375#pullrequestreview-1780236104 @lassoan Thanks for having taken the time to suggest fixes to the issues.

> as long as the SlicerDMRI extension is installed. So, I would recommend to move this test into SlicerDMRI extension.

I do see that having SlicerDMRI installed may not be the default case for part of the users, and that pure 3D Slicer builds are not expected to have any extension installed/to rely on any extension, so the next suggestion:

> If we add a test into Slicer core then it should be modified so that it passes if reading fails.

would mean that we would not know whether the test is actually reading the data, and the same applies to

> and update the function itself to raise an exception that explains that SlicerDMRI is required.

since I guess we should catch the exception and make the the CI happy just by printing the exception message about SlicerDMRI not being available.

So, I agree with the proposal of moving the below to SlicerDMRI:
https://github.com/Slicer/Slicer/blob/1d77e473abebc3b554d886614e2958adff34e38e/Base/Python/slicer/util.py#L782-L790

Now, I am not sure where it should fit into SlicerDMRI architecture, as I am still not all that familiar with how things are designed. Maybe @jcfr or @lassoan you have a suggestion of where this fits.

Thanks.

jcfr:
As suggested by @lassoan, we should keep the function in `slicer.util` and report a message/exception if there are no reader registered for the `FiberBundleFile` file type.

The test in Slicer should be failing if the extension is not installed, and passing if it is installed.

Base/Python/slicer/tests/test_slicer_util_load_data.py
jhlegarreta: Changed in 9eabaf7.

Base/Python/slicer/tests/test_slicer_util_load_data.py
jhlegarreta: Changed in 9eabaf7.

Base/Python/slicer/tests/test_slicer_util_load_data.py
jhlegarreta: Changed in 9eabaf7.

Base/Python/slicer/tests/test_slicer_util_load_data.py
jhlegarreta: Changed in 9eabaf7.

Base/Python/slicer/tests/test_slicer_util_load_data.py
jhlegarreta: Changed in 9eabaf7.

Base/Python/slicer/tests/test_slicer_util_load_data.py
jhlegarreta: Changed in 9eabaf7.

Base/Python/slicer/tests/test_slicer_util_load_data.py
jhlegarreta: Changed in 9eabaf7.

jhlegarreta:
Re https://github.com/Slicer/Slicer/pull/7375#issuecomment-1854896608 : commit 004209f and 9eabaf7. Thanks for the reviews.

Base/Python/slicer/util.py
```diff
@@ -780,13 +780,23 @@ def loadColorTable(filename, returnNode=False):
...
     :param filename: full path of the file to load.
     :param returnNode: Deprecated.
     :return: loaded node (if multiple nodes are loaded then a list of nodes).
       If returnNode is True then a status flag and loaded node are returned.
     """
+
+    emm = slicer.app.extensionsManagerModel()
+    is_ext_installed = emm.isExtensionInstalled("SlicerDMRI")
+    if not is_ext_installed:
+        errorMessage = f"SlicerDMRI extension not installed: failed to read data from file: {filename}"
```
jcfr, line 795:
```suggestion
    if slicer.app.ioManager().registeredFileReaderCount("FiberBundleFile") == 0:
        errorMessage = f"FiberBundleFile reader not registered: failed to read data from file: {filename}"
        errorMessage += "\n" + "Try to install the SlicerDMRI extension"
        
```

Since extension may either be installed or bundled (e.g in the context of custom application), the test should be more specific.

Base/Python/slicer/util.py
```diff
@@ -780,13 +780,23 @@ def loadColorTable(filename, returnNode=False):
 
 
 def loadFiberBundle(filename, returnNode=False):
-    """Load node from file.
+    """Load fiber bundle node from file. The ``SlicerDMRI`` needs to be
+    installed to load fiber bundle data.
```
jcfr, line 784:
```suggestion
  """Load fiber bundle node from file.
  
  .. warning::

    To ensure the FiberBundleFile reader is registered, the ``SlicerDMRI``
    extension needs to be installed.
```

Base/Python/slicer/tests/test_slicer_util_load_data.py
```diff
@@ -0,0 +1,36 @@
...
+import slicer
+
+
+class SlicerUtilLoadDataTests(unittest.TestCase):
+
+    def setUp(self):
+
+        emm = slicer.app.extensionsManagerModel()
+        is_ext_installed = emm.isExtensionInstalled("SlicerDMRI")
+        self.assertTrue(is_ext_installed)
```
jcfr, line 15:
```suggestion
```

The test should always pass independently of having the extension installed or not

Base/Python/slicer/tests/test_slicer_util_load_data.py
```diff
@@ -0,0 +1,36 @@
...
+        hash_algorithm = "SHA256"
+        checksum = "e1848e4d0b90d6c7442b2474c513d0464c6f141253c25aee6f5bcda4c828deac"
+        url = urljoin(slicer.util.TESTING_DATA_URL, f"{hash_algorithm}/{checksum}")
+
+        with tempfile.TemporaryDirectory() as tmpdirname:
+            file_basename = "tractography.vtk"
+            targetFilePath = Path(tmpdirname).joinpath(file_basename)
+            reDownloadIfChecksumInvalid = True
+            downloadSuccess = slicer.util.downloadFile(
+                url, targetFilePath, checksum=f"{hash_algorithm}:{checksum}", reDownloadIfChecksumInvalid=reDownloadIfChecksumInvalid)
```
jcfr, line 29:
```suggestion
                url, fiberBundleFilePath, checksum=f"{hash_algorithm}:{checksum}", reDownloadIfChecksumInvalid=True)
```

Base/Python/slicer/tests/test_slicer_util_load_data.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def test_loadFiberBundle(self):
+
+        hash_algorithm = "SHA256"
+        checksum = "e1848e4d0b90d6c7442b2474c513d0464c6f141253c25aee6f5bcda4c828deac"
+        url = urljoin(slicer.util.TESTING_DATA_URL, f"{hash_algorithm}/{checksum}")
+
+        with tempfile.TemporaryDirectory() as tmpdirname:
+            file_basename = "tractography.vtk"
+            targetFilePath = Path(tmpdirname).joinpath(file_basename)
```
jcfr, line 26:
```suggestion
            fiberBundleFilePath = Path(tmpdirname).joinpath(file_basename)
```

Base/Python/slicer/tests/test_slicer_util_load_data.py
```diff
@@ -0,0 +1,36 @@
...
+            reDownloadIfChecksumInvalid = True
+            downloadSuccess = slicer.util.downloadFile(
+                url, targetFilePath, checksum=f"{hash_algorithm}:{checksum}", reDownloadIfChecksumInvalid=reDownloadIfChecksumInvalid)
+            self.assertTrue(downloadSuccess)
+
+            node = slicer.util.loadFiberBundle(targetFilePath)
```
jcfr, line 32:
```suggestion
            node = slicer.util.loadFiberBundle(fiberBundleFilePath)
```

Base/Python/slicer/tests/test_slicer_util_load_data.py
```diff
@@ -0,0 +1,36 @@
...
+            self.assertIsNotNone(node)
+
+            success, node = slicer.util.loadFiberBundle(filename, returnNode=True)
```
jcfr, line 35:
```suggestion
            success, node = slicer.util.loadFiberBundle(fiberBundleFilePath, returnNode=True)
```

Base/Python/slicer/tests/test_slicer_util_load_data.py
```diff
@@ -0,0 +1,36 @@
...
+            self.assertIsNotNone(success)
```
jcfr, line 36:
```suggestion
            self.assertTrue(success)
            self.assertIsInstance(node, slicer.vtkMRMLFiberBundleNode)
```

Base/Python/slicer/tests/test_slicer_util_load_data.py
```diff
@@ -0,0 +1,36 @@
...
+        with tempfile.TemporaryDirectory() as tmpdirname:
+            file_basename = "tractography.vtk"
+            targetFilePath = Path(tmpdirname).joinpath(file_basename)
+            reDownloadIfChecksumInvalid = True
+            downloadSuccess = slicer.util.downloadFile(
+                url, targetFilePath, checksum=f"{hash_algorithm}:{checksum}", reDownloadIfChecksumInvalid=reDownloadIfChecksumInvalid)
+            self.assertTrue(downloadSuccess)
+
+            node = slicer.util.loadFiberBundle(targetFilePath)
+            self.assertIsNotNone(node)
```
jcfr, line 33:
```suggestion
            self.assertIsInstance(node, slicer.vtkMRMLFiberBundleNode)
```

Base/Python/slicer/tests/test_slicer_util_load_data.py
jcfr, line 32: Move code testing loading into an inlined function:

```python
            def _load():
                node = slicer.util.loadFiberBundle(fiberBundleFilePath)
                self.assertIsInstance(node, slicer.vtkMRMLFiberBundleNode)

                ....


            if slicer.app.ioManager().registeredFileReaderCount("FiberBundleFile") == 0:
                with self.assertRaisesRegex(RuntimeError, "FiberBundleFile reader not registered") as cm:
                    _load()

            else:
                _load()
```

jcfr (changes requested):

After addressing the comments, this should be good for integration :rocket:

Base/Python/slicer/util.py
jhlegarreta: Done in de0d64b.

Base/Python/slicer/util.py
jhlegarreta: Done in de0d64b.

Base/Python/slicer/tests/test_slicer_util_load_data.py
jhlegarreta: Removed in bc1d677.

Base/Python/slicer/tests/test_slicer_util_load_data.py
```diff
@@ -0,0 +1,36 @@
...
+    def test_loadFiberBundle(self):
+
+        hash_algorithm = "SHA256"
+        checksum = "e1848e4d0b90d6c7442b2474c513d0464c6f141253c25aee6f5bcda4c828deac"
+        url = urljoin(slicer.util.TESTING_DATA_URL, f"{hash_algorithm}/{checksum}")
+
+        with tempfile.TemporaryDirectory() as tmpdirname:
+            file_basename = "tractography.vtk"
+            targetFilePath = Path(tmpdirname).joinpath(file_basename)
+            reDownloadIfChecksumInvalid = True
```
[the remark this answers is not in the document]
jhlegarreta: Removed in bc1d677.

Base/Python/slicer/tests/test_slicer_util_load_data.py
jhlegarreta: Changed in bc1d677.

Base/Python/slicer/tests/test_slicer_util_load_data.py
jhlegarreta: Changed in bc1d677.

Base/Python/slicer/tests/test_slicer_util_load_data.py
jhlegarreta: Changed in bc1d677.

Base/Python/slicer/tests/test_slicer_util_load_data.py
jhlegarreta: Changed in bc1d677.

Base/Python/slicer/tests/test_slicer_util_load_data.py
jhlegarreta: Changed in bc1d677.

Base/Python/slicer/tests/test_slicer_util_load_data.py
jhlegarreta: Changed in bc1d677.

Base/Python/slicer/tests/test_slicer_util_load_data.py
```diff
@@ -0,0 +1,44 @@
...
+
+        def _load(fiberBundleFilePath, returnNode):
+
+            if returnNode:
+              success, node = slicer.util.loadFiberBundle(fiberBundleFilePath, returnNode=True)
+              self.assertTrue(success)
+            else:
+              node = slicer.util.loadFiberBundle(fiberBundleFilePath)
+
+            self.assertIsInstance(node, slicer.vtkMRMLFiberBundleNode)
```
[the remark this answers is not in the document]
jcfr: To avoid having to rename the parameter to `useReturnNode=False` and call `_load` twice for each case, let's just test the two scenarios together

Base/Python/slicer/tests/test_slicer_util_load_data.py
```diff
@@ -0,0 +1,44 @@
...
+class SlicerUtilLoadDataTests(unittest.TestCase):
+
+    def setUp(self):
+
+        pass
+
+
+    def test_loadFiberBundle(self):
+
+        def _load(fiberBundleFilePath, returnNode):
```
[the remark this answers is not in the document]
jhlegarreta: :+1: Done in 183239e.

Base/Python/slicer/tests/test_slicer_util_load_data.py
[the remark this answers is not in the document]
jhlegarreta: :+1: Done in 183239e.

Base/Python/slicer/tests/test_slicer_util_load_data.py
```diff
@@ -0,0 +1,44 @@
...
+            fiberBundleFilePath = Path(tmpdirname).joinpath(file_basename)
+            downloadSuccess = slicer.util.downloadFile(
+                url, fiberBundleFilePath, checksum=f"{hash_algorithm}:{checksum}", reDownloadIfChecksumInvalid=True)
+            self.assertTrue(downloadSuccess)
+
+            if slicer.app.ioManager().registeredFileReaderCount("FiberBundleFile") == 0:
+                with self.assertRaisesRegex(RuntimeError, "FiberBundleFile reader not registered") as cm:
+                    _load(fiberBundleFilePath, False)
+            else:
+                _load(fiberBundleFilePath, False)
```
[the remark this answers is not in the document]
jhlegarreta: :+1: Done in 183239e.

Base/Python/slicer/tests/test_slicer_util_load_data.py
```diff
@@ -0,0 +1,44 @@
...
+
+            file_basename = "tractography.vtk"
+            fiberBundleFilePath = Path(tmpdirname).joinpath(file_basename)
+            downloadSuccess = slicer.util.downloadFile(
+                url, fiberBundleFilePath, checksum=f"{hash_algorithm}:{checksum}", reDownloadIfChecksumInvalid=True)
+            self.assertTrue(downloadSuccess)
+
+            if slicer.app.ioManager().registeredFileReaderCount("FiberBundleFile") == 0:
+                with self.assertRaisesRegex(RuntimeError, "FiberBundleFile reader not registered") as cm:
+                    _load(fiberBundleFilePath, False)
```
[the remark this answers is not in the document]
jhlegarreta: :+1: Done in 183239e.

Base/Python/slicer/tests/test_slicer_util_load_data.py
jhlegarreta: Removed in 9eabaf7.

jhlegarreta:
A user from the Fetal-Neonatal Neuroimaging Developmental Science Center (@FNNDSC) at Boston Children's Hospital (BCH) believed that there was some error in WMA stemming from this method until I clearly showed that this was working correctly, pointed to this PR and showed that the need to have SlicerDMRI installed is clearly stated in the method docstring. In order to avoid such misunderstandings, and to best use one's time, pinging @jcfr again to merge this, please.

Base/Python/slicer/util.py
```diff
@@ -780,13 +780,26 @@ def loadColorTable(filename, returnNode=False):
...
 
     :param filename: full path of the file to load.
     :param returnNode: Deprecated.
     :return: loaded node (if multiple nodes are loaded then a list of nodes).
       If returnNode is True then a status flag and loaded node are returned.
     """
+
+    if slicer.app.ioManager().registeredFileReaderCount("FiberBundleFile") == 0:
+        errorMessage = f"FiberBundleFile reader not registered: failed to read data from file: {filename}"
+        errorMessage += "\n" + "Try to install the SlicerDMRI extension"
```
jamesobutler, line 797: “Try to install…” seems to invoke a lack of confidence that it might not work. “Install the SlicerDMRI extension…” would be more confident.

Further, checking if the SlicerDMRI is already installed would be helpful. Checking if the file reader is present is a transitive check but not a direct check to see if SlicerDMRI is already installed.

You could also prompt the user directly install the extension as some may not know how to install an extension if they are told this information. https://slicer.readthedocs.io/en/latest/developer_guide/script_repository.html#download-and-install-extension provides an example of such.

jhlegarreta: Re recent push force:
- Added a check for the extension.
- Made the first message be imperative.
- The second one is less sure to be fixed by reinstalling the extension, so left the suggestion mode.
- I am hesitant to trigger an extension installation, especially if this is done silently (not sure if the command involves a prompt) and if this involves restarting the software. If the user does not know how to install an extension, I do not think a loading method should solve that, as it is an essential feature of 3D Slicer that users should be aware of by other means. But let me know if it should still be that way (should probably be made into a policy for any file type then).

lassoan (commented):
Thank you for your patience. The code looks good to me. I'll test the test now and if it passes  then I can merge it.

Isn't it a problem though that when the user is trying to load a fiber bundle using "Add data" module then the user is not informed that SlicerDMRI needs to be installed to be able to load this file as fiber bundle?

jhlegarreta:
> Isn't it a problem though that when the user is trying to load a fiber bundle using "Add data" module then the user is not informed that SlicerDMRI needs to be installed to be able to load this file as fiber bundle?

I agree, but IMO that belongs to a separate PR: this PR was mainly about adding a test. The check could be done when the user chooses to load a file as `FiberBundle` in the dialog that gets opened. This could also be generalized to other file types/extensions that may exist, and the block in https://github.com/Slicer/Slicer/pull/7375/files#diff-562900981ed3d2b01cfc2fc2c8e500e521cd2c40789f9b5159d35af02375225dR795-R800 could be put into a method so that the message is consistent across extensions.

jhlegarreta:
> What is going to call the  loadFiberBundle function?

Loading a fiber bundle when the user chooses such option when adding new data.

> Slicer core does not use it, it is just a convenience function for Python scripting. Users will not see the "FiberBundle" option anywhere until they install SlicerDMRI.

Slicer core not requiring this and potentially moving to SlicerDMRI has already been discussed previously throughout the above comments, and Jean-Christophe said this should remain in `slicer.util`: https://github.com/Slicer/Slicer/pull/7375#issuecomment-1854896608

lassoan:
>>What is going to call the loadFiberBundle function?

> Loading a fiber bundle when the user chooses such option when adding new data.

Loading a fuber bundle node using Slicer GUI ("Add data" window) will not call this function.

It is still useful to make the loadFiberNundle method fail with an informative error message because there are a few extensions that user this function and might not check if SlicerDMRI extension was installed.

jamesobutler:
@jhlegarreta Per the discussion at https://github.com/SlicerDMRI/whitematteranalysis/issues/230, since the user was using white matter analysis, shouldn’t SlicerWMA explicitly list SlicerDMRI as an dependency so that `slicer.util.loadFiberBundle` actually works?

Currenrly [SlicerWMA](https://github.com/SlicerDMRI/SlicerWMA) lists no dependent extensions. See https://github.com/Slicer/ExtensionsIndex/blob/fb045b7b8adffae29b9a1a37c603c97596488145/SlicerWMA.s4ext#L5

Base/Python/slicer/util.py
```diff
@@ -780,13 +780,34 @@ def loadColorTable(filename, returnNode=False):
...
     :return: loaded node (if multiple nodes are loaded then a list of nodes).
       If returnNode is True then a status flag and loaded node are returned.
     """
+
+    # Check if SlicerDMRI is installed
+    extension_name = "SlicerDMRI"
+    em = slicer.app.extensionsManagerModel()
+    if not em.isExtensionInstalled(extension_name):
+        errorMessage = f"{extension_name} not installed."
+        errorMessage += "\n" + f"Install the {extension_name} extension"
```
jcfr, line 800:
```suggestion
```

Base/Python/slicer/util.py
```diff
@@ -780,13 +780,34 @@ def loadColorTable(filename, returnNode=False):
...
+
+    # Check if the appropriate reader is registered
+    if slicer.app.ioManager().registeredFileReaderCount("FiberBundleFile") == 0:
+        errorMessage += f"FiberBundleFile reader not registered: failed to read data from file: {filename}"
+        errorMessage += "\n" + f"Try reinstalling the {extension_name} extension"
+
+    if userMessages.GetNumberOfMessages() > 0:
+        errorMessage += "\n" + userMessages.GetAllMessagesAsString()
+        raise RuntimeError(errorMessage)
```
jcfr, line 809:
```suggestion
    if slicer.app.ioManager().registeredFileReaderCount("FiberBundleFile") == 0:
        message = f"FiberBundleFile reader not registered: failed to read data from file: {filename}"

        extension_name = "SlicerDMRI"
        em = slicer.app.extensionsManagerModel()
        if not em.isExtensionInstalled(extension_name):
            message += "\n" + f"The {extension_name} extension is not installed, but may help work with this data."
        raise RuntimeError(message)
```

jcfr: Additionally I think we should also display an info message like it is done in the DICOMBrowser. See https://github.com/Slicer/Slicer/blob/d7080d6a519df05983f9aebf9e0abd7cbfbc02e6/Modules/Scripted/DICOMLib/DICOMBrowser.py#L234-L243

lassoan: I would not block the execution of the function by displaying a popup. It is a library function. The caller of the function can use `with slicer.util.tryWithErrorDisplay("Failed to read fiber bundle."):` if a popup needs to be displayed.

lassoan:
The test is passed on my computer, so I have no objection for merging this, but @jcfr has some suggestions, so he needs to confirm when this PR is ready to go.

lassoan (dismissed):

Base/Python/slicer/util.py
jhlegarreta: @jcfr Can you please agree with @jamesobutler whether this is required or not? He explicitly requested the check in https://github.com/Slicer/Slicer/pull/7375#discussion_r1451751391

jhlegarreta:
> The test is passed on my computer, so I have no objection for merging this,

@lassoan Thanks for the test. It is appreciated.

> but @jcfr has some suggestions, so he needs to confirm when this PR is ready to go.

Please see https://github.com/Slicer/Slicer/pull/7375#discussion_r1458047071: I was requested to add such a check. I am fine if we remove it, but dealing with contradicting requests is not easy to deal with in this context.

Base/Python/slicer/util.py
jamesobutler: @jhlegarreta I agree with @jcfr. You should only check the status of SlicerDMRI being installed if the file reader is not detected. No need to check if SlicerDMRI is installed every time this utility method is called. Only when there will be a failure.

jcfr: I will follow up asap, and will propose an updated PR. Thanks for the patience :pray:

jcfr:
> I was requested to add such a check. I am fine if we remove it, but dealing with contradicting requests is not easy within this context.

Thanks for revisiting the approach, we need to shuffle things around to call the function only if (1) the IO is not registered and not each we attempt to load a FiberBundle, and (2) only if the main application name is `Slicer`)

And while the test is passing, the fact the code itself is attempting to call `userMessages.GetNumberOfMessages()` without having the variable `userMessages` defined is not correct.

I will follow-up with an update.

jcfr:
> The test is passed on my computer

Additionally, since the test script `test_slicer_util_load_data.py` is not associated with any test in `Applications/SlicerApp/Testing/Python/CMakeLists.txt`[^1], existing tests are passing but the new script was never called through `ctest`.

[^1]: https://github.com/Slicer/Slicer/blob/main/Applications/SlicerApp/Testing/Python/CMakeLists.txt

jcfr:
After adding the test ... attempt to run it illustrate some of the issue:

```
588: Test command: /home/jcfr/Projects/Slicer-Release/Slicer-build/Slicer "--no-splash" "--testing" "--no-main-window" "--disable-cli-modules" "--python-code" "import slicer.testing; slicer.testing.runUnitTest(['/home/jcfr/Projects/Slicer-Release/Slicer-build/Applications/SlicerApp/Testing/Python', '/home/jcfr/Projects/Slicer/Base/Python/slicer/tests'], 'test_slicer_util_load_data')"
588: Test timeout computed to be: 1500
588: When loading module  "CropVolume" , the dependency "ResampleScalarVectorDWIVolume" failed to be loaded.
588: qSlicerMarkupsModulePrivate::addToolBar: no main window is available, toolbar is not added
588: qSlicerSequencesModulePrivate::addToolBar: no main window is available, toolbar is not added
588: -------------------------------------------
588: path: ['/home/jcfr/Projects/Slicer-Release/Slicer-build/Applications/SlicerApp/Testing/Python', '/home/jcfr/Projects/Slicer/Base/Python/slicer/tests']
588: testname: test_slicer_util_load_data
588: -------------------------------------------
588: test_loadFiberBundle (test_slicer_util_load_data.SlicerUtilLoadDataTests) ... ERROR
588: 
588: ======================================================================
588: ERROR: test_loadFiberBundle (test_slicer_util_load_data.SlicerUtilLoadDataTests)
588: ----------------------------------------------------------------------
588: Traceback (most recent call last):
588:   File "/home/jcfr/Projects/Slicer/Base/Python/slicer/tests/test_slicer_util_load_data.py", line 38, in test_loadFiberBundle
588:     _load(fiberBundleFilePath)
588:   File "/home/jcfr/Projects/Slicer/Base/Python/slicer/tests/test_slicer_util_load_data.py", line 18, in _load
588:     success, node = slicer.util.loadFiberBundle(fiberBundleFilePath, returnNode=True)
588:   File "/home/jcfr/Projects/Slicer-Release/Slicer-build/bin/Python/slicer/util.py", line 797, in loadFiberBundle
588:     em = slicer.app.extensionsManagerModel()
588: NameError: name 'slicer' is not defined
588: 
588: ----------------------------------------------------------------------
588: Ran 1 test in 0.416s
588: 
588: FAILED (errors=1)
588: Traceback (most recent call last):
588:   File "<string>", line 1, in <module>
588:   File "/home/jcfr/Projects/Slicer-Release/Slicer-build/bin/Python/slicer/testing.py", line 28, in runUnitTest
588:     exitFailure()
588:   File "/home/jcfr/Projects/Slicer-Release/Slicer-build/bin/Python/slicer/testing.py", line 11, in exitFailure
588:     raise Exception(message)
588: Exception
1/1 Test #588: py_nomainwindow_test_slicer_util_load_data ...***Failed    4.91 sec

0% tests passed, 1 tests failed out of 1
```

jcfr:
We will integrate using `Squash & Merge` with the following message:

````
ENH: Update slicer.util.loadFiberBundle to check for registered reader

This update ensures that the `loadFiberBundle` function checks for the registration
of the appropriate reader before attempting to load fiber bundle data. In case the
reader is not registered, a detailed and informative message is provided, including
the class implementing the reader and the module from the extension that provides it.

If the application is Slicer (and not a custom application), the user is suggested
to install the SlicerDMRI extension.

For instance, the output message may look like the following:

```bash
$ ./Slicer -c "import slicer;print(slicer.util.loadFiberBundle('/tmp/tractography.vtk'));slicer.app.quit()"

The module "TractographyDisplay" has not been registered.
The following modules have been registered: ("EventBroker")
  File "<string>", line 1, in <module>
  File "/path/to/bin/Python/slicer/util.py", line 828, in loadFiberBundle
    raise RuntimeError(errorMessage)

RuntimeError: FiberBundleFile reader not registered: Failed to load node from file: /tmp/tractography.vtk

FiberBundleFile reader is implemented in the qSlicerFiberBundleReader class expected to be registered by the TractographyDisplay module provided by the SlicerDMRI extension.

Status:
- TractographyDisplay module: Not loaded.
- SlicerDMRI extension: Not installed.

Please install the extension for proper functionality.
```
````

jcfr (approved):

jcfr:
@jamesobutler @jhlegarreta  Could you have another look ? :pray: 

Note that these changes have been tested with/without the `TractographyDisplay` loaded.

jamesobutler (approved):

Base/Python/slicer/util.py
jhlegarreta: Solved with the last push force. Thanks @jcfr.

jhlegarreta:
Sorry, was away from the keyboard. Thanks for all your work to improve this, and glad to know that this helped to fix another two issues. Thanks again.

PR #7379, opened 2023-11-11
STYLE: Remove unreachable "break" statement after "return" outside of "switch"

jcfr:
To identify all location if a `break;` statement following `return;`, the entire codebase was inspected using a command like the following:

```
$ ack --ignore-dir=CLI  "break\;" -C2 | ack "return"
```


pieper (approved):
Nice cleanup

jamesobutler (approved):

PR #7383, opened 2023-11-14
DOC: Improve layout of AbstractScriptedSegmentEditorEffect docstrings

jcfr:
These were identified while working on a larger set of improvements for:
* enforcing docstring style
* adding generation of module specific API

Here is a preview of the module specific API that will be integrated as part of a different pull request:

image


jcfr:
I am wondering if instructions for using & instantiating the scripted effect should instead be directly added to https://github.com/Slicer/Slicer/blob/main/Docs/developer_guide/modules/segmenteditor.md?plain=1 ?

lassoan (approved):

PR #7384, opened 2023-11-14
Fix DICOMUtils getSortedImageFiles return value if input file list is empty

jcfr:
These were identified while working on a larger set of improvements for enforcing docstring style.


Modules/Scripted/DICOMLib/DICOMUtils.py
```diff
@@ -613,18 +613,21 @@ def __exit__(self, type, value, traceback):
...
-        epsilon: Maximum difference in distance between slices to consider spacing uniform
+    :param filePaths : Paths of the local DICOM files to sort.
+    :param epsilon: Maximum difference in distance between slices to consider spacing uniform.
+
+    :return: Tuple of (files, distances, warningText)
     """
     warningText = ""
     if len(filePaths) == 0:
-        return filePaths, [], warningText
+        return filePaths, {}, warningText
```
jamesobutler, line 630: It appears there are a couple other returns of `return filePaths, [], warningText` which includes a list type instead of a dict type in this method.

Modules/Scripted/DICOMLib/DICOMUtils.py
```diff
@@ -613,18 +613,21 @@ def __exit__(self, type, value, traceback):
...
-        to the acquisition plane)
+    Use the first file to get the ImageOrientationPatient for the
+    series and calculate the scan direction (assumed to be perpendicular
+    to the acquisition plane)
 
-        epsilon: Maximum difference in distance between slices to consider spacing uniform
+    :param filePaths : Paths of the local DICOM files to sort.
+    :param epsilon: Maximum difference in distance between slices to consider spacing uniform.
+
+    :return: Tuple of (files, distances, warningText)
```
pieper, line 626: It would be nice to have the types here to avoid issues like the one James pointed out.

pieper (approved):
Just a suggestion from me, but I agree with the issue @jamesobutler raised.

Modules/Scripted/DICOMLib/DICOMUtils.py
jcfr: Issue of @jamesobutler has just been addressed.

Documentation of the return type has been addressed in the second commit by introducing the corresponding type annotation.

lassoan (approved):
Thank you, it looks good.

PR #7390, opened 2023-11-15
Docstring fix escape sequence issues

jcfr:
Fix scripts and update Ruff configuration checking if docstrings require raw string prefix

For reference:

> In Python, backslashes are typically used to escape characters in strings. In raw strings (those prefixed with an r), however, backslashes are treated as literal characters.
>
> PEP 257 recommends the use of raw strings (i.e., r"""raw triple double quotes""") for docstrings that include backslashes. The use of a raw string ensures that any backslashes are treated as literal characters, and not as escape sequences, which avoids confusion.

Copied from https://docs.astral.sh/ruff/rules/escape-sequence-in-docstring/

### Documentation of `slicer.util.longPath`

| Context | Illustration |
|--|--|
| Before| image | 
| After|image |

### DIFF
--- a/Base/Python/slicer/util.py
+++ b/Base/Python/slicer/util.py
@@ -3797,10 +3797,10 @@ def pip_uninstall(requirements):
 
 
 def longPath(path):
-    """Make long paths work on Windows, where the maximum path length is 260 characters.
+    r"""Make long paths work on Windows, where the maximum path length is 260 characters.
 
     For example, the files in the DICOM database may have paths longer than this limit.
-    Accessing these can be made safe by prefixing it with the UNC prefix ('\\?\').
+    Accessing these can be made safe by prefixing it with the UNC prefix ('\\\\?\\').
 
     :param string path: Path to be made safe if too long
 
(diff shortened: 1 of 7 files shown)


jcfr:
@lassoan This addresses the use of raw literals we discussed during the developer hangout.

sjh26 (approved):
lgtm

Base/Python/slicer/util.py
```diff
@@ -3797,10 +3797,10 @@ def pip_uninstall(requirements):
 
 
 def longPath(path):
-    """Make long paths work on Windows, where the maximum path length is 260 characters.
+    r"""Make long paths work on Windows, where the maximum path length is 260 characters.
 
     For example, the files in the DICOM database may have paths longer than this limit.
-    Accessing these can be made safe by prefixing it with the UNC prefix ('\\?\').
+    Accessing these can be made safe by prefixing it with the UNC prefix ('\\\\?\\').
```
pieper, line 3803: Not sure this is correct - now that it's a raw string should not be escaping the backslashes with more backslashes.

jcfr: My understanding is that using raw literals avoid on level of escaping.

These are still needed so that sphinx can properly render it ... see before/after in the pull-request description https://github.com/Slicer/Slicer/pull/7390#issue-1993785459

To avoid this, we would need to use sphinx/docutils literals as well. See https://docutils.sourceforge.io/docs/ref/rst/roles.html#literal

pieper: ah, I see - okay that makes sense.

PR #7393, opened 2023-11-15
BUG: Resolve del method override issue in AutoComplete segment editor effect

jcfr:
This commit rectifies the linter error concerning the `__del__` method in the AutoComplete segment editor effect. The specific issue was that the `__del__` method expects one parameter, but two were given. To resolve this, the call to the superclass `__del__` has been removed, as it is not implemented.

Note that while this correction resolves the immediate warning, there may still be issues with proper garbage collection for some effects.

A proposal to address the root cause is documented in:
* Slicer#7392

It also updates the Ruff configuration, re-enabling the PLE0302 rule to ensure that `__del__` functions are called with the expected number of parameters.

### DIFF
--- a/.ruff.toml
+++ b/.ruff.toml
@@ -74,8 +74,6 @@ extend-ignore = [
   "PGH001",  # No builtin `eval()` allowed
   "PGH002",  # `warn` is deprecated in favor of `warning`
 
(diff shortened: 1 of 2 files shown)


lassoan (approved):
Looks good to me.

PR #7396, opened 2023-11-15
COMP: Fix reporting of -Wunused-function related to qSlicerApplicationHelper

jhlegarreta:
Workaround the reporting of -Wunused-function warnings when using
GCC <= 7.3 to build translation units not making use of the
"qSlicerApplicationHelper::postInitializeApplication" function.

This was verified with the following self-contained example
compiled using the "https://godbolt.org/" platform specifying the
-Wunused-function flag and toggling between "x86-64 gcc 7.3" and
"x86-64 gcc 7.4".

```
#include <iostream>

namespace
{
void setEnableQtTesting()
{
  std::cout << "setEnableQtTesting" << std::endl;
}

template<typename SlicerMainWindowType>
void postInitializeApplication()
{
  setEnableQtTesting();
}
int main(int, char*[])
{
  std::cout << "Hello" << std::endl;
  return 0;
}
```

Fixes:
```
In file included from /work/Stable/Slicer-0/Base/QTApp/qSlicerApplicationHelper.h:77:0,
                 from inner-build/Modules/Loadable/InteractiveSeeding/qSlicerTractographyInteractiveSeedingModuleGenericTest.cxx:38:
/work/Stable/Slicer-0/Base/QTApp/qSlicerApplicationHelper.txx:64:6:
 warning: ‘void {anonymous}::splashMessage(QScopedPointer<QSplashScreen>&, const QString&)’ defined but not used [-Wunused-function]
 void splashMessage(QScopedPointer<QSplashScreen>& splashScreen, const QString& message)
      ^~~~~~~~~~~~~
```

and
```
/work/Stable/Slicer-0/Base/QTApp/qSlicerApplicationHelper.txx:53:6:
 warning: ‘void {anonymous}::setEnableQtTesting()’ defined but not used [-Wunused-function]
 void setEnableQtTesting()
      ^~~~~~~~~~~~~~~~~~
```

raised for the SlicerDMRI extenstion build in:
https://slicer.cdash.org/viewBuildError.php?type=1&buildid=3210886

### DIFF
--- a/Base/QTApp/qSlicerApplicationHelper.txx
+++ b/Base/QTApp/qSlicerApplicationHelper.txx
@@ -80,6 +80,17 @@ int qSlicerApplicationHelper::postInitializeApplication(
     QScopedPointer<SlicerMainWindowType>& window
     )
 {
+
+#if defined(Q_CC_GNU) && Q_CC_GNU <= 703
+  // The lines below are needed for suppressing "-Wunused-function" warnings
+  // reported when using GCC<=7.3 to build translation units not making use
+  // of the "qSlicerApplicationHelper::postInitializeApplication" function.
+# ifdef Slicer_USE_QtTesting
+  (void)setEnableQtTesting; // Fix -Wunused-function warning
+# endif
+  (void)splashMessage; // Fix -Wunused-function warning
+#endif
+
   if (app.style())
     {
     app.installEventFilter(app.style());


jhlegarreta:
Not sure if this will make the warnings go away; have not tested locally.

jcfr (changes requested):
It should be fixed using an approach similar to the one implemented in these patches:
* https://github.com/Slicer/Slicer/commit/53fea52c122ad7a1fd6055e3d1c4c96b546a2489 (`COMP: Fix -Wunused-function warnings in qMRMLLayoutManager tests`)
* https://github.com/commontk/CTK/commit/f6e9a1aace53d64669444723fd25bbf8e6815393 (`COMP: Fix -Wunused-function warnings in ctkErrorLogModelTestHelper`)

I also suspect this is specific to `GCC 7` used in the Slicer build environment[^1] `slicer/buildenv-qt5-centos7:latest`

I am not able to reproduce locally using `GCC 9.4.0` despite of making sure the  CXX flag `-Wunused-function` is used.

[^1]: https://github.com/Slicer/SlicerBuildEnvironment

jhlegarreta:
@jcfr thanks for the insight. 

> It should be fixed using an approach similar to the one implemented in these patches:

Are both things necessary? According to your hypothesis, the changes are to be done in `SlicerDMRI`, and not here. Is this correct?

However, I'd say that the error stems from the fact that they were not declared as `static`. I will give this a try first, if you agree.

I ignore why CentOS is used but support for it will end on 2024-06-30.

jcfr:
> I ignore why CentOS is used but support for it will end on 2024-06-30.

It is needed to allow Slicer to run on most linux distribution. This matches https://peps.python.org/pep-0599/#the-manylinux2014-policy

jcfr:
> the changes are to be done in SlicerDMRI

Not exactly, they would likely have to be done in Slicer but in `qSlicerModuleWidgetGenericTest.cxx.in`[^1].

[^1]: https://github.com/Slicer/Slicer/blob/2c7136165ac387985dac9947784ced43b556c37d/Base/QTGUI/Testing/Cxx/qSlicerModuleWidgetGenericTest.cxx.in

jhlegarreta:
I will amend the commit message once the right approach makes the tests pass.

jhlegarreta:
Still happening:
https://github.com/Slicer/Slicer/actions/runs/6894152495/job/18755324113?pr=7396#step:4:13360

jcfr (approved):
Local testing was successful. 

Once the CI also confirm the warning is addressed, this is now good for integration.

PR #7401, opened 2023-11-16
COMP: Call `.c_str()` explicitly to obtain a `const char*`

jhlegarreta:
Call `.c_str()` explicitly to obtain a `const char*`.

Fixes:
```
   const char * originalStringPointer = object->Get##variable(); \
```

and similar errors.

raised for example in:
https://github.com/Slicer/Slicer/actions/runs/6885650499/job/18730113110#step:4:3460

### DIFF
--- a/Libs/MRML/Core/vtkMRMLCoreTestingMacros.h
+++ b/Libs/MRML/Core/vtkMRMLCoreTestingMacros.h
@@ -325,14 +325,14 @@
 /// test a string variable on the object by calling Set/Get
 #define TEST_SET_GET_STRING( object, variable ) \
   { \
-  const char * originalStringPointer = object->Get##variable(); \
+  const char * originalStringPointer = (object->Get##variable()).c_str(); \
   std::string originalString; \
   if( originalStringPointer != nullptr ) \
     { \
     originalString = originalStringPointer; \
     } \
   object->Set##variable( "testing with a const char");                  \
-  if( strcmp(object->Get##variable(), "testing with a const char") != 0) \
+  if( strcmp((object->Get##variable()).c_str(), "testing with a const char") != 0) \
     {                                                                   \
     std::cerr << "Error in Set/Get"#variable << " with a string literal" << std::endl; \
     return EXIT_FAILURE;                                                \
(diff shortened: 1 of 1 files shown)


jcfr (changes requested):
Since `TEST_SET_GET_STRING` is also used for `char*`[^1][^2], addressing this could instead be done using a template based approach where the compiler would choose the specialization based on the type.

[^1]: https://github.com/Slicer/Slicer/blob/b41abc4f69a7f99253f6880733be97186dbfbd77/Libs/MRML/Core/vtkMRMLNode.h#L995-L997
[^2]: https://github.com/Slicer/Slicer/blob/b41abc4f69a7f99253f6880733be97186dbfbd77/Libs/MRML/Core/vtkMRMLCoreTestingUtilities.cxx#L207C13-L209

jhlegarreta:
Not sure about the work that involves. Please go ahead as time permits.

jcfr:
Since the problem and a possible path forward have been documented in a dedicated issue, I am now closing this pull request.

In the future, a new one could be created or this one could be re-opened.

lassoan:
To fix the warnings, we should use the appropriate test macro: `CHECK_STD_STRING`.


`CHECK_STRING` is for classic VTK string properties that are get/set as `char*`.

jhlegarreta:
> To fix the warnings, we should use the appropriate test macro: CHECK_STD_STRING.
> CHECK_STRING is for classic VTK string properties that are get/set as char*.

Excellent. Will give it a try and will open a PR. Thanks for the review @lassoan.

jhlegarreta:
The solution proposed in PR #7411 has fixed the warnings.

The issue with using `CHECK_STRING` is that the `TEST_SET_GET_STRING` macro takes an object instance and the property name to exercise the Set/Get methods; it does not take a pair of lh and rh arguments that are compared, as `CHECK_STRING` does.

So, using `CHECK_STRING` and  `CHECK_STD_STRING` (re https://github.com/Slicer/Slicer/issues/7403#issuecomment-1821009761) requires a different approach in the caller (calling all the Set/Get methods in the caller) unless a more elaborated logic is implemented in  `TEST_SET_GET_STRING` (https://github.com/Slicer/Slicer/issues/7403#issue-1998033130), as we do not know beforehand the type the `object->Get##variable()` call will return.

IMO, PR #7411 is enough as a solution until somebody thinks they want to sophisticate the `TEST_SET_GET_STRING` macro.

PR #7402, opened 2023-11-16
BUG: Fix VTK-based object creation in PythonSlicer regardless of import order

jcfr:
This commit addresses an issue where the creation of VTK-based objects in script executed through the PythonSlicer executable was dependent on the order of module imports. The update ensures consistent and successful instantiation of associated objects, such as `vtkMRMLScene`, regardless of the import order.

For instance, the following code is now supported:

```py
import slicer
import vtk
object = slicer.vtkMRMLScene()
assert isinstance(object, vtk.vtkObject)
```

It doesn't require the developer to always import `vtk` before `slicer`.

The problem stemmed from the wrapping of MRML or Slicer VTK-based C++ classes through `vtkMacroKitPythonWrap` (from vtkAddon). The generated `vtk*PythonInitImpl.cxx` files do not include the names of dependent VTK modules in the list used with `vtkPythonUtil::ImportModule`.

Additionally, a comment is added to `slicerqt.py` to clarify that imports may be necessary to accommodate legacy code in Slicer extensions relying on `from main import vtk, qt, ctk, slicer`.

### DIFF
--- a/Base/Python/mrml.py
+++ b/Base/Python/mrml.py
@@ -2,7 +2,15 @@
 namespace.
 """
 
+# Importing VTK is required for loading its Python modules and ensuring seamless instantiation
+# of VTK-based MRML or Slicer Python modules from code executed in both "PythonSlicer executable"
+# and the "Slicer Python Console" environments. For more details, see Base/Python/slicer/__init__.py
+import vtk  # noqa: F401
+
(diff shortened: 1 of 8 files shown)


jcfr:
Local testing confirmed that all tests are passing :white_check_mark:

PR #7406, opened 2023-11-17
Update Ruff configuration checking trailing commas

jcfr:
These changes are steps toward minimizing the differences between the Slicer python sources and output of formatter like "ruff" or "black".

**Rational:** The presence of a trailing comma can reduce diff size when parameters or elements are added or removed from function calls, function definitions, literals, etc.

See https://docs.astral.sh/ruff/rules/#flake8-commas-com

Once integrated, the commits prefixed with `STYLE: ` will be added to the `.git-blame-ignore-revs`[^1] file.

### DIFF
--- a/.ruff.toml
+++ b/.ruff.toml
@@ -7,6 +7,7 @@ select = [
 #  "ANN",         # flake8-annotations
   "B",           # flake8-bugbear
   "C4",          # flake8-comprehensions
+  "COM",         # flake8-commas
   "E", "F", "W", # flake8
   "D",           # pydocstyle
 #  "EXE",         # flake8-executable
--- a/Applications/SlicerApp/Testing/Python/BRAINSFitRigidRegistrationCrashIssue4139.py
+++ b/Applications/SlicerApp/Testing/Python/BRAINSFitRigidRegistrationCrashIssue4139.py
@@ -125,7 +125,7 @@ class BRAINSFitRigidRegistrationCrashIssue4139Test(ScriptedLoadableModuleTest):
             "movingVolume": moving,
             "linearTransform": outputTransform,
             "outputVolume": outputVolume,
-            "useRigid": True
+            "useRigid": True,
         }
         cmdLineNode = slicer.cli.runSync(slicer.modules.brainsfit, parameters=parameters)
         self.assertIsNotNone(cmdLineNode)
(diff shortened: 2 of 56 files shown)


pieper (approved):
I'm not sure I'm a fan of this style since it is not the same as json but I guess it's okay.

jcfr:
>  not the same as json

Ditto and thanks for the review.

As an aside, while not[^2] supported in "regular" JSON[^1], it is part of `JSON5`[^3].

In the future, we may 


### Additional notes

#### use of short numeric literal in Slicer

Independently of the trailing comma, python code making use of the short numeric literal like the following is invalid JSON. The use of short numeric literal will be fixed in a follow-up pull request (for sake of being compliant with black/ruff)

<table>
<tr><td>
Invalid :x:  (because of short numeric literal (e.g `1.`))
</td>
<td>
Valid :white_check_mark: 
</td>
<td>
Invalid :x: (because of missing trailing comma)
</td></tr>
<tr>
<td>

```py
{
  "Red": [1., 0., 0.],
  "Green": [0., 1., 0.],
  "Yellow": [1., 1., 0.]
}
```

</td>

<td>

```py
{
  "Red": [1.0, 0.0, 0.0],
  "Green": [0.0, 1.0, 0.0],
  "Yellow": [1.0, 1.0, 0.0]
}
```

</td>

<td>

```py
{
  "Red": [1.0, 0.0, 0.0],
  "Green": [0.0, 1.0, 0.0],
  "Yellow": [1.0, 1.0, 0.0],
}
```

</td>

</tr></table>

### From Python object to JSON

I have been using the following to go from Python object representation to JSON. This could be used directly from this webpage providing a Python REPL (https://pyodide.org/en/stable/console.html)

```
import ast
import json

print(json.dumps(ast.literal_eval("""{
  "Red": [1.0, 0.0, 0.0],
  "Green": [0.0, 1.0, 0.0],
  "Yellow": [1.0, 1.0, 0.0],
}"""), indent=2))
```

or 

```
import ast
import json

print(json.dumps(ast.literal_eval("""{
  "Red": [1., 0., 0.],
  "Green": [0., 1., 0.],
  "Yellow": [1., 1., 0.],
}"""), indent=2))
```

### Format related discussions

* https://discuss.python.org/t/toml-development-has-stalled-implications-for-pyproject-toml-and-a-way-forward/26521/20
* [Comparison of configuration file languages](https://gist.github.com/njsmith/78f68204c5d969f8c8bc645ef77d4a8f#summary)

[^1]: https://datatracker.ietf.org/doc/html/rfc7159
[^2]: https://stackoverflow.com/questions/201782/can-you-use-a-trailing-comma-in-a-json-object
[^3]: https://json5.org/

PR #7413, opened 2023-11-23
COMP: Fix unused parameter warnings in `vtkMRMLI18NTest1.cxx`

jhlegarreta:
Fix unused parameter warnings in `vtkMRMLI18NTest1.cxx`

Fixes:
```
/usr/src/Slicer/Libs/MRML/Core/Testing/vtkMRMLI18NTest1.cxx:
 In member function ‘virtual std::string {anonymous}::vtkTestTranslator::Translate(const char*, const char*, const char*, int)’:
/usr/src/Slicer/Libs/MRML/Core/Testing/vtkMRMLI18NTest1.cxx:31:101:
 warning: unused parameter ‘disambiguation’ [-Wunused-parameter]
     std::string Translate(const char* context, const char* sourceText, const char* disambiguation = nullptr, int n = -1) override
                                                                                                     ^~~~~~~
/usr/src/Slicer/Libs/MRML/Core/Testing/vtkMRMLI18NTest1.cxx:31:119:
 warning: unused parameter ‘n’ [-Wunused-parameter]
     std::string Translate(const char* context, const char* sourceText, const char* disambiguation = nullptr, int n = -1) override
                                                                                                                       ^
```

Raised for example at:
https://github.com/Slicer/Slicer/actions/runs/6950843491/job/18911721210#step:4:3066

### DIFF
--- a/Libs/MRML/Core/Testing/vtkMRMLI18NTest1.cxx
+++ b/Libs/MRML/Core/Testing/vtkMRMLI18NTest1.cxx
@@ -28,7 +28,8 @@ namespace
     vtkTypeMacro(vtkTestTranslator, vtkMRMLTranslator);
 
     /// Translation method for testing that returns "translated-(context)(sourceText)" as translation
-    std::string Translate(const char* context, const char* sourceText, const char* disambiguation = nullptr, int n = -1) override
+    std::string Translate(const char* context, const char* sourceText,
+                          const char* vtkNotUsed(disambiguation) /*= nullptr*/, int vtkNotUsed(n) /*= -1*/) override
       {
       return std::string("translated-") + context + sourceText;
       }


jhlegarreta:
The warning is now gone:
https://github.com/Slicer/Slicer/actions/runs/6964552908/job/18951952935?pr=7413

jcfr (approved):

jhlegarreta:
Not sure about the benefit of using `vtkNotUsed` over removing the parameter name. Also, the commit message now does not message the change introduced by the commit.

jcfr:
When reviewing and improving codes, it is helpful to know the name of the parameter without having to look for it. This is what we make use of `vtkNotUsed` and `Q_UNUSED`

jcfr:
> Also, the commit message now does not message the change introduced by the commit

This is being addressed by choosing `Squash & Merge`

jhlegarreta:
> This is addressed by integrating the changes using Squash & Merge

The commit message still says that the names are omitted, whereas they are marked as not used. IMO, those are different things.

jcfr:
If you inspect the commit effectively integrated in the history, we can see it has been updated.  See https://github.com/Slicer/Slicer/commit/93e5fee641b46d6d990968a472dbc95d78308766

jhlegarreta:
> By inspecting the commit effectively integrated in the history, we can see it has been updated. See https://github.com/Slicer/Slicer/commit/93e5fee641b46d6d990968a472dbc95d78308766

Interesting. Unclear what is the path that can lead to that; need to investigate to know about it.

jcfr:
> Unclear what is the path that can lead to that; need to investigate to know about it

There are three ways of integrating pull-requests:
- `Rebase & Merge`
- `Squash & Merge`
- `Merge`

In Slicer, to keep the history linear and easy to follow, we support only `Rebase & Merge` and `Squash & Merge`.

We reserve `Squash & Merge` when keeping a sequence of commits doesn't make sense or when the commit message associated with a pull request introducing a single commit need to be tweaked (we avoid to have to wait for a complete round of CI to complete before the integration happens)

You may read more at https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/about-pull-request-merges#merge-your-commits

PR #7417, opened 2023-11-23
COMP: Fix `-Wstringop-truncation` GCC warning in vtkMRMLROIListNode 

jhlegarreta:
Fixes:

```
In function ‘char* strncpy(char*, const char*, size_t)’,
    inlined from ‘virtual void vtkMRMLROIListNode::ReadXMLAttributes(const char**)’ at slicer/Libs/MRML/Core/vtkMRMLROIListNode.cxx:199:26:
/usr/include/x86_64-linux-gnu/bits/string_fortified.h:95:34: warning: ‘char* __builtin_strncpy(char*, const char*, long unsigned int)’ output truncated before terminating nul copying 1 byte from a string of the same length [-Wstringop-truncation]
   95 |   return __builtin___strncpy_chk (__dest, __src, __len,
      |          ~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~
   96 |                                   __glibc_objsize (__dest));
      |                                   ~~~~~~~~~~~~~~~~~~~~~~~~~
```

raised when compiling on an Ubuntu 22.04 machine with gcc-11.

### DIFF
--- a/Libs/MRML/Core/vtkMRMLROIListNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLROIListNode.cxx
@@ -196,7 +196,7 @@ void vtkMRMLROIListNode::ReadXMLAttributes(const char** atts)
         if (IDPtr != nullptr)
           {
           // replace the space with a carriage return
-          IDPtr = strncpy(IDPtr, "\n", 1);
+          IDPtr[0] = '\n';
           }
         }
       // now parse the string into tokens by the newline


jhlegarreta:
Have not tested locally.

jhlegarreta:
The previous warning is gone, but the compiler still complains:
https://github.com/Slicer/Slicer/actions/runs/6971620227/job/18972063400?pr=7417#step:4:2945

Not sure if that warning is right.

lassoan:
This raw char* buffer manipulation looks very ugly. We should convert to std::string and use its methods.

Note that Annotations module is only kept for backward compatibility with older scenes and will be removed in the next major version of Slicer.

jhlegarreta:
> We should convert to std::string and use its methods.

@lassoan Please go ahead and amend the PR. Thanks.

jcfr (changes requested):
The proposed change introduces `-Wsizeof-pointer-memaccess`[^1] warning:

[^1]: https://gcc.gnu.org/gcc-4.8/porting_to.html

```
vtkMRMLROIListNode.cxx:199:47: 'strncpy' call operates on objects of type 'char' while the size is based on a different type 'char *'
vtkMRMLROIListNode.cxx:199:47: did you mean to provide an explicit length?
```

jcfr (approved):
With the updated fix and after rebasing against `main` including https://github.com/Slicer/Slicer/pull/7419:

```
$ ctest -R "(MRMLROINode|MRMLROIList)"
Test project /home/jcfr/Projects/Slicer-Release/Slicer-build
    Start 65: vtkMRMLROIListNodeTest1
1/2 Test #65: vtkMRMLROIListNodeTest1 ..........   Passed    0.22 sec
    Start 66: vtkMRMLROINodeTest1
2/2 Test #66: vtkMRMLROINodeTest1 ..............   Passed    0.22 sec

100% tests passed, 0 tests failed out of 2

Label Time Summary:
MRMLCore    =   0.44 sec*proc (2 tests)

Total Test time (real) =   0.47 sec
```

PR #7419, opened 2023-11-24
BUG: Revert invalid `-Wdeprecated-declarations` fix in TEST_SET_GET_STRING

jcfr:
This reverts commit e9772b1b1 (COMP: Call `c_str()` to get `const char*` for property comparison tests) introduced through https://github.com/Slicer/Slicer/pull/7411 originally intended to fix warnings like the following:

```
   const char * originalStringPointer = object->Get##variable(); \
```

Initializing a std::string object with a null string is not valid.

cc: @jhlegarreta

### DIFF
--- a/Libs/MRML/Core/vtkMRMLCoreTestingMacros.h
+++ b/Libs/MRML/Core/vtkMRMLCoreTestingMacros.h
@@ -325,36 +325,30 @@
 /// test a string variable on the object by calling Set/Get
 #define TEST_SET_GET_STRING( object, variable ) \
   { \
-  std::string tmpString = std::string(object->Get##variable()); \
-  const char * originalStringPointer = tmpString.c_str(); \
+  const char * originalStringPointer = object->Get##variable(); \
   std::string originalString; \
   if( originalStringPointer != nullptr ) \
     { \
     originalString = originalStringPointer; \
     } \
   object->Set##variable( "testing with a const char");                  \
-  tmpString = std::string(object->Get##variable());                     \
-  if( strcmp(tmpString.c_str(), "testing with a const char") != 0) \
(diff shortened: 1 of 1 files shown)


jcfr:
Nightly build have been manually stopped and re-triggered[^1] to include this fix.

[^1]: https://github.com/Slicer/DashboardScripts/blob/main/maintenance/guides/re-trigger-nightly-build.md

jhlegarreta:
Not sure where the null assignation attempt happens as the object is expect not to be null, and the getter call is expected to be about an existing property. Wondering why tests passed. I had a look to the preview builds in the past days and had not seen errors.

jcfr:
> Wondering why tests passed. I had a look to the preview builds in the past days and had not seen errors.

The Preview build started to fail[^1] after the integration[^2] of the change.

And the passing the test was restored[^3] once it has been reverted.

[^1]: https://slicer.cdash.org/index.php?project=SlicerPreview&date=2023-11-22&filtercount=1&showfilters=1&field1=groupname&compare1=63&value1=Nightly-Packages

[^2]: https://slicer.cdash.org/build/3217540/update

[^3]: https://slicer.cdash.org/index.php?project=SlicerPreview&date=2023-11-24&filtercount=1&showfilters=1&field1=groupname&compare1=63&value1=Nightly-Packages

jhlegarreta:
> The tests associated with the Preview build started to fail[1](https://github.com/Slicer/Slicer/pull/7419#user-content-fn-1-3bded8bd90dbdaf22cb19eb179afe80e) after the integration[2](https://github.com/Slicer/Slicer/pull/7419#user-content-fn-2-3bded8bd90dbdaf22cb19eb179afe80e) of the change.

OK, then I missed that. I am sorry.

PR #7421, opened 2023-11-24
DOC:  Eliminate obsolete Doxygen grouping

jcfr:
This removes the outdated Doxygen grouping, initially implemented to organize classes related to Slicer-specific modules or capabilities.

Since there were no systematic ways of checking that each modules were effectively associated with the correct group, the grouping was either incorrect or incomplete.

This removes all doxygen commands matching `/ingroup` and `/defgroup`


jcfr:
Since these changes doesn't functionally impact the application and local compilation was successful, I will move forward with integration in the next hour.

PR #7426, opened 2023-11-25
STYLE: Add recent formatting changes to ignored list for git blame

jcfr:
This adds recent formatting commits to the git blame ignore list.

For convenience, here is the list of commit added to the `.git-blame-ignore-revs` file. While reviewing this pull request, consider (un)checking if any of these should be added/removed or commenting if I missed any.

* [x] 90164ec63b891a0d9ee89b06af68ba8d9a29dfb6 `STYLE: Align Python scripts with "ruff" formatting updating argument parser`
* [x] 830c33a34ca6d33dc70a4004f674ed640337eee7 `STYLE: Align Python scripts with "ruff" formatting updating if expressions`

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -342,3 +342,7 @@ db4bcddcaca1c9277dcb46e8a787043e43ad8b6e
 6a545c3917ae6c284570593d444136df8ac270f4
 # STYLE: Align Python scripts with "ruff" formatting adjusting quotes
 b19d02ac2119b7457c2341e995ad6353235bbbb4
+# STYLE: Align Python scripts with "ruff" formatting updating argument parser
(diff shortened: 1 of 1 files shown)


jcfr:
Considering these commits are introducing news lines, I am not sure if it makes sense to add them to the list ...

jamesobutler (approved):
These commits made no real functional changes so git ignoring them I think makes sense. There will now be more lines when git blaming associated with the last real change. So 1 line associated with a commit will now be 4 lines associated with that commit in terms of functionality which makes sense.

@jcfr I’ll let you integrate if you have thought any differently since you originally posted.

jcfr:
Thanks for the review :pray: I will then move forward with the integration :rocket:

jcfr:
| | |
|--|--|
| Without ignoring | image |
| With ignoring | image |

PR #7430, opened 2023-11-27
BUG: Fix temporary folder containing special characters

lassoan:
qt.QDateTime().currentDateTime().toString(...) may use Arabic numerals on an operating system set to Arabic language. Always specify en-US locale when using when using date in file/folder names to avoid having to work with paths that may contain unicode characters.

Fixes problem originally reported at:
https://discourse.slicer.org/t/i-tried-everything-to-run-totalsegmentator-pls-help/33023/3


jcfr (dismissed):
I also confirm there are no similar changes needed at the CTK level[^1].

[^1]: Using the commands `ack "QDateTime().+toString"` and `ack "QTime().+toString"`

jcfr (approved):

PR #7433, opened 2023-11-27
BUG: Fix rebuild of Endoscopy flythrough when modifying the input curv

Leengit:
Changes include:

- BUG: Lazily rebuild flythrough if input curve is modified
- BUG: Stop a modification to the input curve from triggering a second modification event
- BUG: Allow the user to specify the name of the model to be created
- BUG: Correct names of vtk.vtkMath functions
- DOC: Use "flythrough" as noun; "fly through" as verb
- DOC: Remove use of degree symbol
- STYLE: Remove explicit set of widget as not collapsed.
- STYLE: Don't instantiate \_\_del\_\_ when it defaults appropriately
- STYLE: Use `p` instead of `ctr` for center position of plane

### DIFF
--- a/Modules/Scripted/Endoscopy/Endoscopy.py
+++ b/Modules/Scripted/Endoscopy/Endoscopy.py
@@ -79,6 +79,7 @@ class EndoscopyWidget(ScriptedLoadableModuleWidget, VTKObservationMixin):
         self.timer = qt.QTimer()
         self.timer.setInterval(20)
         self.timer.connect("timeout()", self.flyToNext)
+        self.ignoreInputCurveModified = 0
 
         self.cameraNode = None
         self.inputCurve = None
@@ -341,7 +342,11 @@ class EndoscopyWidget(ScriptedLoadableModuleWidget, VTKObservationMixin):
 
     def onInputCurveModified(self, observer, eventid):
         """If the input curve  was changed we need to repopulate the keyframe UI"""
-        pass
+        if not self.ignoreInputCurveModified:
+            if self.logic:
+                # We are going to have rebuild the EndoscopyLogic later
+                self.logic.cleanup()
+                self.logic = None
 
     def onCreatePathButtonClicked(self):
         """Connected to 'Use this curve'` button.  It allows to:
@@ -351,6 +356,7 @@ class EndoscopyWidget(ScriptedLoadableModuleWidget, VTKObservationMixin):
         - go to start of the path
         """
         inputCurve = self.inputCurveSelector.currentNode()
+        self.ignoreInputCurveModified += 1
 
         if self.logic:
             self.logic.cleanup()
@@ -361,7 +367,7 @@ class EndoscopyWidget(ScriptedLoadableModuleWidget, VTKObservationMixin):
         # Update frame slider range
         self.frameSlider.maximum = max(0, numberOfControlPoints - 2)
 
-        # Create a cursor so that the user can see where the fly through is progressing.
+        # Create a cursor so that the user can see where the flythrough is progressing.
         self.createCursor()
 
         # Hide the cursor, model and inputCurve from the main 3D view
@@ -377,6 +383,8 @@ class EndoscopyWidget(ScriptedLoadableModuleWidget, VTKObservationMixin):
         enable = numberOfControlPoints > 1
         self.flythroughCollapsibleButton.enabled = enable
         self.advancedCollapsibleButton.enabled = enable
+
+        self.ignoreInputCurveModified -= 1
 
         # Initialize to the start of the path
         self.flyTo(0)
@@ -542,9 +550,11 @@ class EndoscopyWidget(ScriptedLoadableModuleWidget, VTKObservationMixin):
             # Rather than a length-zero or length-two string representing an empty dictionary, clear the string.
             cameraOrientationsString = None
 
+        self.ignoreInputCurveModified += 1
         self.inputCurve.SetAttribute(
             EndoscopyLogic.NODE_PATH_CAMERA_ORIENTATIONS_ATTRIBUTE_NAME, cameraOrientationsString,
         )
+        self.ignoreInputCurveModified -= 1
         if self.logic:
             self.logic.cameraOrientations = cameraOrientations
 
@@ -561,6 +571,9 @@ class EndoscopyWidget(ScriptedLoadableModuleWidget, VTKObservationMixin):
         )
 
     def flyToNext(self):
+        if self.logic is None:
+            # We invalidated self.logic due to an intervening user modification of the input.  We need it now.
+            self.logic = EndoscopyLogic(self.inputCurve)
         currentStep = int(self.frameSlider.value)
         nextStep = currentStep + self.skip + 1
         if nextStep > self.logic.resampledCurve.GetNumberOfControlPoints() - 2:
@@ -570,9 +583,12 @@ class EndoscopyWidget(ScriptedLoadableModuleWidget, VTKObservationMixin):
     def flyTo(self, resampledCurvePointIndex):
         """Apply the resampledCurvePointIndex-th step in the path to the global camera"""
 
+        if self.logic is None:
+            # We invalidated self.logic due to an intervening user modification of the input.  We need it now.
+            self.logic = EndoscopyLogic(self.inputCurve)
+
         if (
-            self.logic is None
-            or self.logic.resampledCurve is None
+            self.logic.resampledCurve is None
             or not 0 <= resampledCurvePointIndex < self.logic.resampledCurve.GetNumberOfControlPoints()
         ):
             return
@@ -769,7 +785,7 @@ class EndoscopyLogic:
 
         # Find a plane that approximately includes the points of the resampled curve,
         # so that we can use its normal to define the "up" direction. This is somewhat
-        # nonsensical if self.numberOfResampledCurveControlPoints < 3, but proceed anyway.
+        # nonsensical if self.resampledCurve.GetNumberOfControlPoints() < 3, but proceed anyway.
         _, self.planeNormal = EndoscopyLogic.planeFit(points.T)
 
         # Interpolate the user-supplied orientations to compute (and assign) an orientation to every control point of


Modules/Scripted/Endoscopy/Endoscopy.py
```diff
@@ -185,11 +187,12 @@ def setupFlythroughUI(self):
 
         # View angle slider
         viewAngleSlider = ctk.ctkSliderWidget()
+        # View angles are in degrees
         viewAngleSlider.decimals = 0
         viewAngleSlider.minimum = 30
         viewAngleSlider.maximum = 180
         viewAngleSlider.connect("valueChanged(double)", self.viewAngleSliderValueChanged)
-        flythroughFormLayout.addRow(_("View Angle:"), viewAngleSlider)
+        flythroughFormLayout.addRow(_("View Angle°:"), viewAngleSlider)
```
jcfr, line 195: For sake of consistency with how angle are set in the application, let's not use the symbol. Instead:
* explained it in the tooltip
* rename variable from `viewAngleSlider` to `viewAngleDegSlider` (consistent with how similar variable are name throughout the code base)

Modules/Scripted/Endoscopy/Endoscopy.py
```diff
@@ -185,11 +187,12 @@ def setupFlythroughUI(self):
 
         # View angle slider
         viewAngleSlider = ctk.ctkSliderWidget()
+        # View angles are in degrees
```
jcfr, line 190: After renaming the variable, remove this comment.

Modules/Scripted/Endoscopy/Endoscopy.py
```diff
@@ -144,6 +145,7 @@ def setupFlythroughUI(self):
         flythroughCollapsibleButton = ctk.ctkCollapsibleButton()
         flythroughCollapsibleButton.text = _("Flythrough")
         flythroughCollapsibleButton.enabled = False
+        flythroughCollapsibleButton.collapsed = False
```
jcfr, line 148: We minimize the number of click required and not collapse this.

Modules/Scripted/Endoscopy/Endoscopy.py
```diff
@@ -341,17 +344,22 @@ def enableOrDisableCreateButton(self):
 
     def onInputCurveModified(self, observer, eventid):
         """If the input curve  was changed we need to repopulate the keyframe UI"""
-        pass
+        if not self.ignoreInputCurveModified:
+            if self.logic:
+                # We are going to have rebuild the EndoscopyLogic later
+                self.logic.cleanup()
+                self.logic = None
```
jcfr, line 351: Re-creating the logic upon modification should be avoided and is an anti-pattern, I have a follow-up change address this.

Modules/Scripted/Endoscopy/Endoscopy.py
```diff
@@ -643,7 +661,9 @@ def _hideOnlyInView(viewNodeID, displayableNodes):
                 if not displayNode:
                     continue
                 with slicer.util.NodeModify(displayNode):
-                    displayNode.SetViewNodeIDs(allViewNodeIDs)
+                    if not displayNode.GetNumberOfViewNodeIDs():
+                        # Currently empty / not initialized, so initialize with all views.
+                        displayNode.SetViewNodeIDs(allViewNodeIDs)
```
jcfr, line 666:
```suggestion
                    displayNode.SetViewNodeIDs(allViewNodeIDs)
```

Since the purpose of the function is to hide in all view expect one, this is not needed.

Modules/Scripted/Endoscopy/Endoscopy.py
```diff
@@ -657,9 +677,11 @@ def createCursor(self):
...
             )
             cursor.CreateDefaultDisplayNodes()
-            cursor.CreateDefaultDisplayNodes()
-            cursor.GetDisplayNode().SetSelectedColor(1, 0, 0)  # red
-            cursor.GetDisplayNode().SetSliceProjection(True)
+            for displayNodeIndex in range(cursor.GetNumberOfDisplayNodes()):
+                displayNode = cursor.GetNthDisplayNode(displayNodeIndex)
+                displayNode.SetSelectedColor(1, 0, 0)  # red
+                displayNode.SetSliceProjection(True)
+                displayNode.BackfaceCullingOn()  # so that the camera can see through the cursor from inside
```
jcfr, line 684: Since the cursor is fiducial with only one display node, this is not needed.

Modules/Scripted/Endoscopy/Endoscopy.py
```diff
@@ -704,6 +726,9 @@ def __init__(self, inputCurve, dl=0.5):
         self.dl = dl  # desired world space step size (in mm)
         self.setControlPoints(inputCurve)
 
+    def __del__(self):
+        self.cleanup()
+
```
jcfr, line 731:
```suggestion
```

The cleanup function is always explicitly called.

Modules/Scripted/Endoscopy/Endoscopy.py
```diff
@@ -720,6 +745,8 @@ def setControlPoints(self, inputCurve: slicer.vtkMRMLMarkupsCurveNode) -> bool:
                 f"inputCurve must be of type '{expectedType}', is type '{type(inputCurve)}' value '{inputCurve}'",
             )
 
+        if not inputCurve.GetNumberOfDisplayNodes():
+            inputCurve.CreateDefaultDisplayNodes()
```
jcfr, line 749:
```suggestion
```

The displayable manager already take care of creating the required display node

Modules/Scripted/Endoscopy/Endoscopy.py
```diff
@@ -852,10 +878,6 @@ def getDefaultOrientation(self, curve, curveControlPointIndex, worldOrientation=
 
         return worldOrientation
 
-    @staticmethod
-    def setWorldOrientation(curve, curveControlPointIndex, worldOrientation):
-        curve.SetNthControlPointOrientation(curveControlPointIndex, worldOrientation)
-
```
jcfr, line 858: This is addressed in https://github.com/Slicer/Slicer/pull/7431

Modules/Scripted/Endoscopy/Endoscopy.py
```diff
@@ -1015,13 +1036,19 @@ class EndoscopyPathModel:
     - Add a single polyline
     """
 
-    def __init__(self, resampledCurve, inputCurve, outputPathNode=None):
+    def __init__(self, resampledCurve, inputCurve, cameraNode, outputPathNode=None, cursorType=None):
```
jcfr, line 1039: The support for alternative cursor type has been removed, and this is not needed. See https://github.com/Slicer/Slicer/commit/cc0ee3744910abf96f1df78d0831bf98e20e25c7

jcfr (changes requested):

Modules/Scripted/Endoscopy/Endoscopy.py
Leengit: Will do.

Modules/Scripted/Endoscopy/Endoscopy.py
Leengit: Will do.

Modules/Scripted/Endoscopy/Endoscopy.py
Leengit: Yes, we can re-architect it later -- good idea! -- but the current preview build does not work unless we do the full rebuild.

Modules/Scripted/Endoscopy/Endoscopy.py
jcfr: Ditto. In that case, this is not needed as the default for collapsible button is to be "open"

Modules/Scripted/Endoscopy/Endoscopy.py
Leengit: Thank you, I'll remove that.

Modules/Scripted/Endoscopy/Endoscopy.py
Leengit: I will remove cursor-type support (but I will keep the outputPathNode support).

Modules/Scripted/Endoscopy/Endoscopy.py
jcfr: > but the current preview build does not work unless we do the full rebuild.

Changes could be tested without rebuilding, you could for example download the latest Stable or Preview build and replace the `Endoscopy.py` file available in the install tree.

Alternatively, you could also remove the `Endoscopy.py` from the install tree and then drag and drop the one in your local checkout. See https://slicer.readthedocs.io/en/latest/developer_guide/extensions.html#run-slicer-with-your-custom-modules

Modules/Scripted/Endoscopy/Endoscopy.py
jcfr: To clarify, the commit referenced above is already integrated into `main`, this means that amending this pull request should be enough.

Modules/Scripted/Endoscopy/Endoscopy.py
jcfr:
```suggestion
```

Modules/Scripted/Endoscopy/Endoscopy.py
jcfr:
```suggestion
        flythroughFormLayout.addRow(_("View Angle:"), viewAngleSlider)
```

Leengit:
I just force pushed to correct a commit message that failed verification.

Modules/Scripted/Endoscopy/Endoscopy.py
jcfr: Properly addressing this presents some challenges (see https://github.com/Slicer/Slicer/issues/7434), waiting these are sorted out, the current approach addresses the problem described [here](https://github.com/Slicer/Slicer/pull/7165#issuecomment-1827205890).

I then suggest to keep the behavior of `_hideOnlyInView` consistent with its name.

```suggestion
                    displayNode.SetViewNodeIDs(allViewNodeIDs)
```

Modules/Scripted/Endoscopy/Endoscopy.py
```diff
@@ -1015,7 +1027,7 @@ class EndoscopyPathModel:
     - Add a single polyline
     """
 
-    def __init__(self, resampledCurve, inputCurve, outputPathNode=None):
+    def __init__(self, resampledCurve, inputCurve, cameraNode, outputPathNode=None):
```
jcfr, line 1028:
```suggestion
    def __init__(self, resampledCurve, inputCurve, outputPathNode=None):
```

The `cameraNode` is not used in the function.

Modules/Scripted/Endoscopy/Endoscopy.py
```diff
@@ -1054,7 +1066,8 @@ def __init__(self, resampledCurve, inputCurve, outputPathNode=None):
                 slicer.mrmlScene.GenerateUniqueName(f"Path-{inputCurve.GetName()}"),
             )
             model.CreateDefaultDisplayNodes()
-            model.GetDisplayNode().SetColor(1, 1, 0)  # yellow
+            for displayNodeIndex in range(model.GetNumberOfDisplayNodes()):
+                model.GetNthDisplayNode(displayNodeIndex).SetColor(1, 1, 0)  # yellow
```
jcfr, line 1068:
```suggestion
            model.GetDisplayNode().SetColor(1, 1, 0)  # yellow
```

Modules/Scripted/Endoscopy/Endoscopy.py
```diff
@@ -977,8 +990,7 @@ def planeFit(points):
         primary_direction = np.abs(n).argmax()
         if n[primary_direction] < 0.0:
             n *= -1.0
-        p = ctr
-        return p, n
+        return ctr, n
```
jcfr, line 991:
```suggestion
        p = ctr
        return p, n
```

I suggest to keep the existing convention (as it was originally introduced) where `p` denotes `position` and `n` denotes `normal`

Leengit:
The most recent force push is to address the failed verification of a commit message.

Modules/Scripted/Endoscopy/Endoscopy.py
Leengit: I will remove it.

Modules/Scripted/Endoscopy/Endoscopy.py
Leengit: Will do.

Modules/Scripted/Endoscopy/Endoscopy.py
Leengit: I will use `p` instead of `ctr`.

Modules/Scripted/Endoscopy/Endoscopy.py
jcfr: > preview build is broken if the user tries to modify the curve after having selected it in the Endoscopy module.

Thanks for clarifying. I will test an follow up.

jcfr (approved):

PR #7436, opened 2023-11-28
Improve Endoscopy module UI self-discovery

jcfr:
This pull request introduces the following improvements:

* All UI elements in the Endoscopy module now have informative tooltips.
* The view angle slider setting is now suffixed with "degree".


Leengit (approved):
Looks good, thanks!

PR #7437, opened 2023-11-28
BUG: Recompute EndoscopyLogic fields immediately instead of lazily

Leengit:
The logic for when to loadCameraOrientations and recompute EndoscopyLogic fields was not working in all cases.  The lazy approach to loading and recomputing could have been modified to fix this, but we did not pursue this approach because the lazy approach has a flaw: the user interface (e.g. number of frames available) should be available immediately, not lazily (e.g. only once the user starts a flythrough).

### DIFF
--- a/Modules/Scripted/Endoscopy/Endoscopy.py
+++ b/Modules/Scripted/Endoscopy/Endoscopy.py
@@ -128,7 +128,6 @@ class EndoscopyWidget(ScriptedLoadableModuleWidget, VTKObservationMixin):
         inputCurveSelector.addEnabled = False
         inputCurveSelector.removeEnabled = False
         inputCurveSelector.connect("currentNodeChanged(bool)", self.enableOrDisableCreateButton)
-        inputCurveSelector.connect("currentNodeChanged(vtkMRMLNode*)", self.setInputCurve)
         pathFormLayout.addRow(_("Curve to modify:"), inputCurveSelector)
         self.inputCurveSelector = inputCurveSelector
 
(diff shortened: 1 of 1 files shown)


Leengit:
Additionally,
```python
inputCurveSelector.connect("currentNodeChanged(vtkMRMLNode*)", self.setInputCurve)
```
is being removed because we do not want to notice that the selected curve has changed, unless the users hits the "Use this curve" button.

lassoan:
> ```python
> ```python
> inputCurveSelector.connect("currentNodeChanged(vtkMRMLNode*)", self.setInputCurve)
> ```
> 
> 
>     
>       
>     
> 
>       
>     
> 
>     
>   
> is being removed because we do not want to notice that the selected curve has changed, unless the users hits the "Use this curve" button.
> ```

Make sure that if the user saves the scene, restarts Slicer, and loads the scene all the node selections are preserved.
For this you usually need to connect a slot to the `currentNodeChanged` signal that saves the selection into the parameter node.

Leengit:
> Make sure that if the user saves the scene, restarts Slicer, and loads the scene all the node selections are preserved.
> For this you usually need to connect a slot to the `currentNodeChanged` signal that saves the selection into the parameter node.

Yes, @jcfr is also recommending use of the parameter node.  Instead currently, we save the keyframe orientations via a `SetAttribute` member function of the curve, which stores a string representation of a dictionary of the needed keyframe orientations.  With this pull request, this information is retrieved when the user loads a scene after restarting Slicer.  However, other attributes are not being stored, including the current frame, the frame skip size, and the timing delay.  Yes, a parameter node can collect all that and more.

However, this pull request is focused on the quick fix that handles an execution path that had failed to reload the keyframe orientations.  Folks at RSNA heard a talk about the updated Endoscopy module this morning from @andinet, and I'm hoping to get this quick fix to them as soon as possible.

Leengit:
I just rebased to the current `main` branch.

PR #7439, opened 2023-11-30
Streamline Endoscopy UX removing "Use this curve" button

jcfr:
Selecting the curve is now sufficient to work with the associated flythrough path. Also listen to the more specific `PointEndInteractionEvent` to only update the curve after user is done updating the curve.

Instantiate one EndoscopyLogic and simplify its API 

... introducing:
* `updateCameraFromOrientationAtIndex()`
* `getNumberOfControlPoints()`
* `saveOrientationAtIndex()`
* `removeOrientationAtIndex()`
* `createTransformFromInputCurve()` / `getTransformFromInputCurve()` / `setInputCurveTransform()`
* `createCursorFromInputCurve()` / `getCursorFromInputCurve()` / `setInputCurveCursor()`

... renaming:
* `setControlPoints` to `setControlPointsByResamplingAndInterpolationFromInputCurve`
* `interpolateOrientations` to `interpolateOrientationsForControlPoints`
* `loadCameraOrientations()` to `getCameraOrientationsFromInputCurve()`
* `saveCameraOrientations()` to `setInputCurveCameraOrientations()`
* `onPlayButtonToggled()` to `setPlaybackEnabled()` (to allow programmatic start/stop of playback)

... updating:

* `interpolateOrientationsForControlPoints()` to require a `cameraOrientations` parameter. This allows to remove the need for maintaining a dedicated `cameraOrientations` instance variable

.. removing:
* `createCursor()`

It also revisits how camera, camera transform ad camera cursor are referenced. These are now directly referenced from the input curve using `Camera`, `CameraTransform` and `CameraCursor` reference role.

Other changes:
* convert relative orientation API to staticmethod
* remove unneeded instance variable related to Endoscopy model export
* consolidate all widget update into `updateWidgetFromMRML()`
* remove `ignoreInputCurveModified` instance variable and instead introduce a `updatingControlPoints` variable to guard `setControlPoints()`
* allow to set and get the input curve `CameraOrientations` independently of the `EndoscopyWidget`. This is done through `getCameraOrientationsFromInputCurve()` and `setInputCurveCameraOrientations()`

### DIFF
--- a/Modules/Scripted/Endoscopy/Endoscopy.py
+++ b/Modules/Scripted/Endoscopy/Endoscopy.py
@@ -797,9 +797,9 @@ class EndoscopyLogic:
         quaternionInterpolator = vtk.vtkQuaternionInterpolator()
         quaternionInterpolator.SetSearchMethod(0)  # binary search
 
-        # # Using a modified Kochanek basis
-        # quaternionInterpolator.SetInterpolationTypeToSpline()
-        # Using linear spherical interpolation
+        # Use the "linear spherical interpolation" rather than the "cubic spline interpolation" (using a modified
+        # Kochanek basis) because the latter seems to use the supplied quaternions as guide points, but doesn't
+        # necessarily take a path through them.
         quaternionInterpolator.SetInterpolationTypeToLinear()
 
         resampledCurveLength = self.resampledCurve.GetCurveLengthWorld()
--- a/Modules/Scripted/Endoscopy/Endoscopy.py
+++ b/Modules/Scripted/Endoscopy/Endoscopy.py
@@ -722,7 +722,7 @@ class EndoscopyLogic:
         self.cameraOrientations = None
         self.cameraOrientationResampledCurveIndices = None
 
-    def setControlPoints(self, inputCurve: slicer.vtkMRMLMarkupsCurveNode) -> bool:
+    def setControlPoints(self, inputCurve: slicer.vtkMRMLMarkupsCurveNode) -> None:
         expectedType = slicer.vtkMRMLMarkupsCurveNode
         if not isinstance(inputCurve, expectedType):
             raise TypeError(
--- a/Modules/Scripted/Endoscopy/Endoscopy.py
+++ b/Modules/Scripted/Endoscopy/Endoscopy.py
@@ -218,24 +218,28 @@ class EndoscopyWidget(ScriptedLoadableModuleWidget, VTKObservationMixin):
         firstOrientationButton.enabled = True
         firstOrientationButton.connect("clicked()", self.onFirstOrientationButtonClicked)
         flythroughOrientationLayout.addWidget(firstOrientationButton)
+        self.firstOrientationButton = firstOrientationButton
 
         backOrientationButton = qt.QPushButton(_("Back"))
         backOrientationButton.toolTip = _("Go to the previous user-supplied keyframe.")
         backOrientationButton.enabled = True
         backOrientationButton.connect("clicked()", self.onBackOrientationButtonClicked)
         flythroughOrientationLayout.addWidget(backOrientationButton)
+        self.backOrientationButton = backOrientationButton
 
         nextOrientationButton = qt.QPushButton(_("Next"))
         nextOrientationButton.toolTip = _("Go to the next user-supplied keyframe.")
         nextOrientationButton.enabled = True
         nextOrientationButton.connect("clicked()", self.onNextOrientationButtonClicked)
         flythroughOrientationLayout.addWidget(nextOrientationButton)
+        self.nextOrientationButton = nextOrientationButton
 
         lastOrientationButton = qt.QPushButton(_("Last"))
         lastOrientationButton.toolTip = _("Go to the last user-supplied keyframe.")
         lastOrientationButton.enabled = True
         lastOrientationButton.connect("clicked()", self.onLastOrientationButtonClicked)
         flythroughOrientationLayout.addWidget(lastOrientationButton)
+        self.lastOrientationButton = lastOrientationButton
 
         flythroughFormLayout.addRow(flythroughOrientationLayout)
 
@@ -350,6 +354,21 @@ class EndoscopyWidget(ScriptedLoadableModuleWidget, VTKObservationMixin):
         self.deleteOrientationButton.enabled = deletable
         self.saveOrientationButton.text = (
             _("Update Keyframe Orientation") if deletable else _("Save Keyframe Orientation")
+        )
+
+        # Update keyframe navigation buttons
+        firstResampledCurvePointIndex = self.logic.getFirstControlPointIndex()
+        lastResampledCurvePointIndex = self.logic.getLastControlPointIndex()
+        (
+            self.firstOrientationButton.enabled,
+            self.backOrientationButton.enabled,
+            self.nextOrientationButton.enabled,
+            self.lastOrientationButton.enabled,
+        ) = (
+            firstResampledCurvePointIndex is not None and firstResampledCurvePointIndex != resampledCurvePointIndex,
+            self.logic.getPreviousControlPointIndex(resampledCurvePointIndex) is not None,
+            self.logic.getNextControlPointIndex(resampledCurvePointIndex) is not None,
+            lastResampledCurvePointIndex is not None and lastResampledCurvePointIndex != resampledCurvePointIndex,
         )
 
         # If there is no input curve available (e.g scene close), stop playblack
--- a/Modules/Scripted/Endoscopy/Endoscopy.py
+++ b/Modules/Scripted/Endoscopy/Endoscopy.py
@@ -301,12 +301,14 @@ class EndoscopyWidget(ScriptedLoadableModuleWidget, VTKObservationMixin):
         """Set and observe a curve node."""
 
         if self.inputCurve is not None:
+            self.removeObserver(self.inputCurve, vtk.vtkCommand.ModifiedEvent, self.updateWidgetFromMRML)
             self.removeObserver(self.inputCurve, slicer.vtkMRMLMarkupsNode.PointModifiedEvent, self.onInputCurveControlPointModified)
             self.removeObserver(self.inputCurve, slicer.vtkMRMLMarkupsNode.PointEndInteractionEvent, self.onInputCurveControlPointEndInteraction)
 
         self.inputCurve = newInputCurve
 
         if newInputCurve is not None:
+            self.addObserver(self.inputCurve, vtk.vtkCommand.ModifiedEvent, self.updateWidgetFromMRML)
             self.addObserver(self.inputCurve, slicer.vtkMRMLMarkupsNode.PointModifiedEvent, self.onInputCurveControlPointModified)
             self.addObserver(self.inputCurve, slicer.vtkMRMLMarkupsNode.PointEndInteractionEvent, self.onInputCurveControlPointEndInteraction)
 
@@ -342,7 +344,9 @@ class EndoscopyWidget(ScriptedLoadableModuleWidget, VTKObservationMixin):
             self.cameraNodeSelector.setCurrentNode(cameraNode)
 
         if cameraNode:
-            self.viewAngleSlider.value = cameraNode.GetViewAngle()
+            viewAngle = EndoscopyLogic.getCameraViewAngleFromInputCurve(self.inputCurve)
+            self.viewAngleSlider.value = viewAngle
+            cameraNode.GetCamera().SetViewAngle(viewAngle)
 
         numberOfControlPoints = self.logic.getNumberOfControlPoints()
 
@@ -388,10 +392,7 @@ class EndoscopyWidget(ScriptedLoadableModuleWidget, VTKObservationMixin):
         self.timer.interval = newValue
 
     def viewAngleSliderValueChanged(self, newValue):
-        cameraNode = EndoscopyLogic.getCameraFromInputCurve(self.inputCurve)
-        if not cameraNode:
-            return
-        cameraNode.GetCamera().SetViewAngle(newValue)
+        EndoscopyLogic.setInputCurveCameraViewAngle(self.inputCurve, newValue)
 
     def setPlaybackEnabled(self, play):
         if play:
@@ -586,6 +587,8 @@ class EndoscopyLogic:
     """
 
     NODE_PATH_CAMERA_ORIENTATIONS_ATTRIBUTE_NAME = "Endoscopy.Path.CameraOrientations"
+    NODE_PATH_VIEW_ANGLE_ATTRIBUTE_NAME = "Endoscopy.Path.ViewAngle"
+    DEFAULT_CAMERA_VIEW_ANGLE = 30.0
 
     def __init__(self, dl=0.5):
         self.dl = dl  # desired world space step size (in mm)
@@ -833,6 +836,19 @@ class EndoscopyLogic:
         if not inputCurve:
             return None
         return inputCurve.GetNodeReference("Camera")
+
+    @staticmethod
+    def setInputCurveCameraViewAngle(inputCurve, viewAngle):
+        if not inputCurve:
+            return
+        inputCurve.SetAttribute(EndoscopyLogic.NODE_PATH_VIEW_ANGLE_ATTRIBUTE_NAME, str(viewAngle))
+
+    @staticmethod
+    def getCameraViewAngleFromInputCurve(inputCurve):
+        if not inputCurve:
+            return
+        value = inputCurve.GetAttribute(EndoscopyLogic.NODE_PATH_VIEW_ANGLE_ATTRIBUTE_NAME)
+        return float(value if value is not None else EndoscopyLogic.DEFAULT_CAMERA_VIEW_ANGLE)
 
     @staticmethod
     def createTransformFromInputCurve(inputCurve):
--- a/Modules/Scripted/Endoscopy/Endoscopy.py
+++ b/Modules/Scripted/Endoscopy/Endoscopy.py
@@ -76,7 +76,6 @@ class EndoscopyWidget(ScriptedLoadableModuleWidget, VTKObservationMixin):
         self.timer = qt.QTimer()
         self.timer.setInterval(20)
         self.timer.connect("timeout()", self.flyToNext)
-        self.ignoreInputCurveModified = 0
 
         self.inputCurve = None
 
@@ -86,12 +85,20 @@ class EndoscopyWidget(ScriptedLoadableModuleWidget, VTKObservationMixin):
         self.setupFlythroughUI()
         self.setupAdvancedUI()
 
+        # Create logic class. Logic implements all computations that should be possible to run
+        # in batch mode, without a graphical user interface.
+        self.logic = EndoscopyLogic()
+
         # Add vertical spacer
         self.layout.addStretch(1)
 
         self.cameraNodeSelector.setMRMLScene(slicer.mrmlScene)
         self.inputCurveSelector.setMRMLScene(slicer.mrmlScene)
         self.outputPathNodeSelector.setMRMLScene(slicer.mrmlScene)
+
+        # Connections
+        self.addObserver(slicer.mrmlScene, slicer.mrmlScene.StartCloseEvent, self.onSceneStartClose)
+        self.addObserver(self.logic.resampledCurve, slicer.vtkMRMLMarkupsNode.PointModifiedEvent, self.updateWidgetFromMRML)
 
     def setupPathUI(self):
         # Path collapsible button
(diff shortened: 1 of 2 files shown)


jcfr:
To ensure all parameters can be updated programmatically and independently of the UI, it would also be nice to either introduce parameter node or set dedicated attributes directly on the curve node for the following:
* frame
* frame skip
* frame delay
* view angle
* plane normal

Association of the following object also needs to be captured at the MRML level (instead of a logic variable):
* camera
* cursor transform
* cursor fiducial

jcfr:
> Association of the following object also needs to be captured at the MRML level (instead of a logic variable)

@lassoan Do we already have an established pattern to "define" the equivalent of reference role from python ?

lassoan:
> Do we already have an established pattern to "define" the equivalent of reference role from python ?

I'm not exactly sure what you mean but we store the user's node selections on the GUI as node references in the parameter node.

Modules/Scripted/Endoscopy/Endoscopy.py
```diff
@@ -984,6 +908,53 @@ def buildCameraMatrix4x4(cameraPosition, inputMatrix3x3):
...
+        # Rather than a length-zero or length-two string representing an empty dictionary, clear the string.
+        if not cameraOrientations:
+            cameraOrientationsString = None
+
+        # JSON turns float `keys` into strings and cannot handle `values` that are numpy arrays.
+        # So, let's convert.
+        convertedCameraOrientations = {str(k): list(v) for k, v in cameraOrientations.items()}
+        if not EndoscopyLogic.cameraOrientationsIsCorrectType(convertedCameraOrientations):
+            raise ValueError("cameraOrientations is not of the correct type")
+        cameraOrientationsString = json.dumps(convertedCameraOrientations)
```
Leengit, line 939: These lines should be in an `else:` block associated with the preceding `if not cameraOrientations:`.

Modules/Scripted/Endoscopy/Endoscopy.py
```diff
@@ -834,11 +680,87 @@ def interpolateOrientations(self):
...
+        focalPoint = cameraNode.GetFocalPoint()
+        viewUp = cameraNode.GetViewUp()
+        worldMatrix3x3 = EndoscopyLogic.buildCameraMatrix3x3(cameraPosition, focalPoint, viewUp)
+        worldOrientation = EndoscopyLogic.matrix3x3ToOrientation(worldMatrix3x3)
+
+        # Add new dictionary key-value pair
+        cameraOrientations = EndoscopyLogic.getCameraOrientationsFromInputCurve(self.inputCurve)
+        cameraOrientations[distanceAlongResampledCurve] = worldOrientation
+
+        EndoscopyLogic.setInputCurveCameraOrientations(self.inputCurve, cameraOrientations)
```
Leengit, line 706: If we change `self.updatingControlPoints` from a `bool` to an `int` then it would be good to wrap this block with
```python
self.updatingControlPoints += 1 # at start of block
self.updatingControlPoints -= 1 # at end of block
```

Modules/Scripted/Endoscopy/Endoscopy.py
```diff
@@ -834,11 +680,87 @@ def interpolateOrientations(self):
...
+
+        # We must freeze the cameraOrientations.keys() generator at the start (by converting it to a list) because we
+        # delete from the cameraOrientations Python dict in this loop.
+        for distanceAlongResampledCurve in list(cameraOrientations.keys()):
+            resampledCurvePointIndex = EndoscopyLogic.indexOfControlPointForDistanceAlongCurve(
+                self.resampledCurve, distanceAlongResampledCurve,
+            )
+            if resampledCurvePointIndex == resampledCurvePointIndexToDelete:
+                del cameraOrientations[distanceAlongResampledCurve]
+                break
```
Leengit, line 721: If the curve is edited, I suppose that it is possible that saved orientations for distinct keyframes now converge to the same keyframe.  (Or maybe that's not possible; I'm not sure.)  If it is possible, and the user asks to delete the orientation from a keyframe, do we want to leave ourselves in a state where there still is an orientation for that frame?  If we want to clear that frame completely then we should delete this `break` statement.

On the other hand, if we are looking for efficiency, we could terminate the loop once we're past the relevant keyframe; or instead do some sort of binary range search; ....

Modules/Scripted/Endoscopy/Endoscopy.py
```diff
@@ -707,28 +551,31 @@ class EndoscopyLogic:
...
+
+    def getNumberOfControlPoints(self):
+        return self.resampledCurve.GetNumberOfControlPoints()
+
+    def setControlPoints(self, inputCurve: slicer.vtkMRMLMarkupsCurveNode) -> None:
+        if self.updatingControlPoints:
+            return
+
         expectedType = slicer.vtkMRMLMarkupsCurveNode
         if not isinstance(inputCurve, expectedType):
```
Leengit, line 573: If the user does "Close scene" while the Endoscopy module is still open, we get an immediate `TypeError` from here because the variable is now `None`.

Modules/Scripted/Endoscopy/Endoscopy.py
```diff
@@ -707,28 +551,31 @@ class EndoscopyLogic:
...
+        self.resampledCurve = slicer.vtkMRMLMarkupsCurveNode()
         self.cameraOrientationResampledCurveIndices = None
 
-    def setControlPoints(self, inputCurve: slicer.vtkMRMLMarkupsCurveNode) -> bool:
+        self.updatingControlPoints = False
+
+    def getNumberOfControlPoints(self):
+        return self.resampledCurve.GetNumberOfControlPoints()
+
+    def setControlPoints(self, inputCurve: slicer.vtkMRMLMarkupsCurveNode) -> None:
```
Leengit, line 568: `getNumberOfControlPoints` is talking about `self.resampledCurve` but `setControlPoints` takes as an argument what will soon be `self.inputCurve`.  It's a little bit confusing, so I'd add a comment to the effect that `setControlPoints` gets its name from the fact that it using the control points of `self.inputCurve` to compute control points for `self.resampledCurve`.

Leengit (changes requested):
This looks great -- thank you for the effort!  I've made a bunch of suggestions.

Modules/Scripted/Endoscopy/Endoscopy.py
Leengit: Perhaps we could change the name of the method from `setControlPoints` to `computeControlPoints`.

Leengit: That could work, though "update from" could mean "copy from", so I was aiming for a word like "compute", "calculate", "infer", "determine", ....

jcfr: Good idea.

There is a precedent for both `calculate` and `compute`, I will go with `calculate`

```$ ack --cpp -i "\:calculate.+From"
Modules/Loadable/Markups/MRML/vtkMRMLMarkupsPlaneNode.cxx
563:void vtkMRMLMarkupsPlaneNode::CalculateAxesFromPoints(const double point0[3], const double point1[3], const double point2[3],

Modules/Loadable/Segmentations/Logic/vtkSlicerSegmentationGeometryLogic.cxx
178:std::string vtkSlicerSegmentationGeometryLogic::CalculateOutputGeometryFromImage()
262:std::string vtkSlicerSegmentationGeometryLogic::CalculateOutputGeometryFromBounds(bool keepCurrentAxisDirections)
```

```$ ack --cpp -i "\:compute.+From"
Libs/MRML/Core/Testing/vtkMRMLVolumeNodeTest1.cxx
84:    vtkMRMLVolumeNode::ComputeIJKToRASFromScanOrder(
89:        vtkMRMLVolumeNode::ComputeScanOrderFromIJKToRAS(ijkToRAS.GetPointer());

Libs/MRML/Core/vtkMRMLVolumeNode.cxx
510:bool vtkMRMLVolumeNode::ComputeIJKToRASFromScanOrder(const char *order,
614:const char* vtkMRMLVolumeNode::ComputeScanOrderFromIJKToRAS(vtkMatrix4x4 *ijkToRAS)

Libs/MRML/Widgets/qMRMLVolumeInfoWidget.cxx
249:    vtkMRMLVolumeNode::ComputeScanOrderFromIJKToRAS(mat.GetPointer())));
363:  if (vtkMRMLVolumeNode::ComputeIJKToRASFromScanOrder(
```

jcfr: I ended up going with the explicit name of `resampleAndInterpolateControlPointsFromInputCurve`

jcfr: Similarly, I think we should rename `interpolateOrientations` to `interpolateControlPointsWithCameraOrientations` ?

Leengit: That's a good one, but I wouldn't start with "interpolate control points" because that is too easy to confuse with what `resampleAndInterpolateControlPointsFromInputCurve` does.  How about `interpolateCameraOrientationsBetweenKeyframes`?

jcfr: Since the concept of `KeyFrame` is not part of the API and what we consider keyframes are in fact labeled as `CameraOrientations` or `Orientations`, I am not sure we want to use `KeyFrames` in the context of `EndoscopyLogic`. This terminology is currently only associated with the user interface.

Since we are literally interpolating between all control points[^1] of the flythrough path (internally represented by `resampledCurve` internal variable), the name `interpolateControlPointsWithOrientations`  nicely capture that and is consistent with:
* `getNumberOfControlPoints`
* `resampleAndInterpolateControlPointsFromInputCurve`
* `saveOrientationAtIndex`
* `removeOrientationAtIndex`
* `updateCameraFromOrientationAtIndex`

Since we are interpolating between all control points while accounting for the _Orientation_ there are either the default ones (or the default ones combined with the ones that were saved) that seems reasonable.

Using `Orientation` (instead of `CameraOrientation`) is also consistent.

[^1]: Resampled control points are also technically keyframes that we are interpolates.

Modules/Scripted/Endoscopy/Endoscopy.py
jcfr: Prior to this changes, we had the following:

```
        for distanceAlongResampledCurve in list(cameraOrientations.keys()):
            resampledCurvePointIndex = EndoscopyLogic.indexOfControlPointForDistanceAlongCurve(
                resampledCurve, distanceAlongResampledCurve,
            )
            if resampledCurvePointIndex == resampledCurvePointIndexToDelete:
                del cameraOrientations[distanceAlongResampledCurve]
```

My understanding is that for a given distance along the curve, we have one or zero saved orientation.

If we have a good reason to keep iterating once we found a matching orientation, the code should be documented to explain the rational.

Modules/Scripted/Endoscopy/Endoscopy.py
```diff
@@ -428,48 +370,28 @@ def onPlayButtonToggled(self, checked):
...
 
     def onDeleteOrientationButtonClicked(self):
-        resampledCurve = self.logic.resampledCurve
+
         resampledCurvePointIndexToDelete = int(self.frameSlider.value)
+
         # The inputCurve (and hence the resampledCurve) could have been modified since these distances were saved,
         # meaning that computing a distance from an index may not now give the same distance value.  However, we can
         # delete the entries associated with this resampledCurvePointIndexToDelete even in the case that its
         # distance has changed.
```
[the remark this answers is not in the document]
jcfr: The API will definitely need to be documented by updating docstrings and by  removing/updating comments, until then, keeping this comment here is sensible. It explains why the entry can be deleted using the function `removeOrientationAtIndex`

Modules/Scripted/Endoscopy/Endoscopy.py
```diff
@@ -707,28 +551,31 @@ class EndoscopyLogic:
...
-        self.resampledCurve = None
+
         self.planeNormal = None
-        self.cameraOrientations = None
+
+        self.resampledCurve = slicer.vtkMRMLMarkupsCurveNode()
         self.cameraOrientationResampledCurveIndices = None
 
-    def setControlPoints(self, inputCurve: slicer.vtkMRMLMarkupsCurveNode) -> bool:
+        self.updatingControlPoints = False
```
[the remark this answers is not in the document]
jcfr: The concern you raised doesn't apply to the current scenario. The boolean nature of `updatingControlPoints` aligns with the intended behavior, and I don't think there is a risk of inadvertently altering the control flow.

The pattern you suggested involving counters is typically reserved for scenarios like `Start/Stop`[^1][^2] (or `Pause/Resume`[^3][^4]) where what indicates if the execution of specific code path should be skipped are distinct function calls/events.

In the context of `updatingControlPoints`, the boolean flag serves a different purpose. It makes sure that the function will always skip if the event is already been processed. This is consistent with all the other widgets[^5] being implemented in Slicer. 

[^1]: https://github.com/Slicer/Slicer/blob/eccb98ce1f58b19d0a429981cc8e3cf131bcf3d5/Libs/MRML/CLI/vtkMRMLCommandLineModuleNode.h#L124-L138
[^2]: https://github.com/Slicer/Slicer/blob/eccb98ce1f58b19d0a429981cc8e3cf131bcf3d5/Libs/MRML/CLI/vtkMRMLCommandLineModuleNode.cxx#L1133-L1153
[^3]: https://github.com/Slicer/Slicer/blob/eccb98ce1f58b19d0a429981cc8e3cf131bcf3d5/Libs/MRML/Widgets/qMRMLLayoutManager.cxx#L1236-L1284
[^4]: https://github.com/Slicer/Slicer/blob/eccb98ce1f58b19d0a429981cc8e3cf131bcf3d5/Libs/MRML/Widgets/qMRMLLayoutManager.h#L253-L265
[^5]: https://github.com/search?type=code&q=repo%3ASlicer%2FSlicer+%2F%28%5BUu%5Dpdating%7CAdding%29.*%3D+true%2F

Modules/Scripted/Endoscopy/Endoscopy.py
```diff
@@ -310,92 +293,51 @@ def setInputCurve(self, newInputCurve):
         """Set and observe a curve node."""
 
         if self.inputCurve is not None:
-            self.removeObserver(self.inputCurve, vtk.vtkCommand.ModifiedEvent, self.onInputCurveModified)
+            self.removeObserver(self.inputCurve, slicer.vtkMRMLMarkupsNode.PointEndInteractionEvent, self.onInputCurvePointModified)
```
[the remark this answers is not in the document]
jcfr: Callbacks have been renamed to include `ControlPoint`

jcfr:
> you mean but we store the user's node selections on the GUI as node references in the parameter node.

@lassoan Flythrough parameters (e.g view angle) are saved as regular note attributes, relationship to nodes are set as node references ("Camera", "CameraTransform" and "CameraCursor")

By convention, should we start these with a lower-case letter ? I kept the upper-case letter for historical reason but this can be changed. I am asking before reference role are currently lower-cased (e.g `display`, `storage`, ...)

Leengit (approved):
To get you a quick review, I have not yet looked at your most recent changes to the code.  However, my testing by running 3D Slicer shows that almost everything is beautiful.  The one problem I am still seeing is that when I `Close Scene`, my console prints out
```text
Close main MRML scene
/home/local/KHQ/lee.newberg/git/Slicer/Slicer-SuperBuild-Release-Qt5.15.2/Slicer-build/bin/Python/slicer/util.py:2786: UserWarning: does not have observer
  warn("does not have observer")
/home/local/KHQ/lee.newberg/git/Slicer/Slicer-SuperBuild-Release-Qt5.15.2/Slicer-build/bin/Python/slicer/util.py:2771: UserWarning: already has observer
  warn("already has observer")
QXcbConnection: XCB error: 3 (BadWindow), sequence: 39328, resource id: 6308414, major code: 40 (TranslateCoords), minor code: 0
```

All told, this represents an improvement so I am marking it as approved.

Modules/Scripted/Endoscopy/Endoscopy.py
```diff
@@ -834,11 +722,186 @@ def interpolateOrientations(self):
...
+        if transform:
+            transform.SetMatrixTransformToParent(worldMatrix4x4)
+
+        return worldMatrix4x4
+
+    @staticmethod
+    def setInputCurveCamera(inputCurve, cameraNode):
+        if not inputCurve:
+            return
+        inputCurve.SetNodeReferenceID("Camera", cameraNode.GetID() if cameraNode else None)
```
Leengit, line 823: Several of these new `EndoscopyLogic` methods are static and accept an `inputCurve` parameter that, in practice, is always the `self.inputCurve` member of the `EndoscopyWidget`.  However, the `EndoscopyLogic` instance already has a `self.inputCurve` member that shadows that from `EndoscopyWidget`.  Perhaps it would be less dangerous to make these methods be non-static and use the `EndoscopyLogic` member variable, rather than accepting a potentially arbitrary `inputCurve` with a static method.

Modules/Scripted/Endoscopy/Endoscopy.py
```diff
@@ -834,11 +722,186 @@ def interpolateOrientations(self):
...
+
+        return cameraOrientations
+
+    def updateCameraFromOrientationAtIndex(self, resampledCurvePointIndex):
+        """Apply the resampledCurvePointIndex-th step in the path to the camera"""
+
+        if (
+            self.resampledCurve is None
+            or not 0 <= resampledCurvePointIndex < self.getNumberOfControlPoints()
+        ):
```
Leengit, line 783: These four lines could be collapsed into one line
```python
if self.resampledCurve is None or not 0 <= resampledCurvePointIndex < self.getNumberOfControlPoints():
```

Leengit (dismissed):
This is looking good.  I've made a few comments about the code; use them as you see fit.

Modules/Scripted/Endoscopy/Endoscopy.py
```diff
@@ -310,92 +301,86 @@ def setInputCurve(self, newInputCurve):
...
+        # If there is no input curve available (e.g scene close), stop playblack
+        if not self.inputCurve:
+            self.setPlaybackEnabled(False)
 
-        # Enable / Disable flythrough button
-        enable = numberOfControlPoints > 1
-        self.flythroughCollapsibleButton.enabled = enable
-        self.advancedCollapsibleButton.enabled = enable
+    def onInputCurveControlPointModified(self, *_unused):
+        if self.inputCurve.GetAttribute("Markups.MovingMarkupIndex"):
```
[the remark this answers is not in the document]
jcfr: This is provided by the markup infrastructure:
* https://github.com/Slicer/Slicer/issues/7450

Modules/Scripted/Endoscopy/Endoscopy.py
jcfr: This was preemptively formatted to comply with `ruff format` (that we will soon fully support in Slicer)

Leengit:
Thank you for your continued efforts.  In case it isn't on your radar, when I `Close Scene` I still get messages in the 3D Slicer Python Console
```text
/home/local/KHQ/lee.newberg/git/Slicer/Slicer-SuperBuild-Release-Qt5.15.2/Slicer-build/bin/Python/slicer/util.py:2786: UserWarning: does not have observer
  warn("does not have observer")
[Qt] void qSlicerSaveDataDialogPrivate::updateStatusIconFromMessageCollection(int, vtkMRMLMessageCollection*, bool) Data save information: ""
[Qt] void qSlicerSaveDataDialogPrivate::updateStatusIconFromMessageCollection(int, vtkMRMLMessageCollection*, bool) Data save information: ""
[Qt] void qSlicerSaveDataDialogPrivate::updateStatusIconFromMessageCollection(int, vtkMRMLMessageCollection*, bool) Data save information: ""
[Qt] void qSlicerSaveDataDialogPrivate::updateStatusIconFromMessageCollection(int, vtkMRMLMessageCollection*, bool) Data save information: ""
[Qt] void qSlicerSaveDataDialogPrivate::updateStatusIconFromMessageCollection(int, vtkMRMLMessageCollection*, bool) Data save information: ""
[Qt] void qSlicerSaveDataDialogPrivate::updateStatusIconFromMessageCollection(int, vtkMRMLMessageCollection*, bool) Data save information: ""
[Qt] Close main MRML scene
```

jcfr:
Can you locally modify `util.py` (e. in the build tree) by adding an exception like this ?

```diff
diff --git a/Base/Python/slicer/util.py b/Base/Python/slicer/util.py
index 6e4c482edd..df3df4a75e 100644
--- a/Base/Python/slicer/util.py
+++ b/Base/Python/slicer/util.py
@@ -2784,6 +2784,7 @@ class VTKObservationMixin:
             obj.RemoveObserver(tag)
         except KeyError:
             warn("does not have observer")
+            raise Exception("does not have observer")
 
     def getObserver(self, obj, event, method, default=None):
         try:

```

Since I am unable to reproduce, this should allow you to identify the source of the message.

Leengit:
> Can you locally modify `util.py`....

I get the below.  (First, I download a CT chest, segment its airway, build a curve inside the airway, play the Endoscopy flythrough, save data to a new directory and then `Close Scene`.)
```text
~/git/Slicer/Slicer-SuperBuild-Release-Qt5.15.2/Slicer-build/bin/Python/slicer/util.py:2786: UserWarning: does not have observer
  warn("does not have observer")
  File "~/git/Slicer/Slicer-SuperBuild-Release-Qt5.15.2/Slicer-build/bin/Python/slicer/util.py", line 2783, in removeObserver
    group, tag, priority = methods.pop(method)
KeyError: <bound method AirwaySegmentationWidget.updateGUIFromParameterNode of <AirwaySegmentation.AirwaySegmentationWidget object at 0x7f9f08336e50>>

During handling of the above exception, another exception occurred:

  File "~/git/Slicer/ExtensionsIndex-SuperBuild-Release-Qt5.15.2/AirwaySegmentation-build/lib/Slicer-5.7/qt-scripted-modules/AirwaySegmentation.py", line 125, in onSceneStartClose
    self.setParameterNode(None)
  File "~/git/Slicer/ExtensionsIndex-SuperBuild-Release-Qt5.15.2/AirwaySegmentation-build/lib/Slicer-5.7/qt-scripted-modules/AirwaySegmentation.py", line 169, in setParameterNode
    self.removeObserver(self._parameterNode, vtk.vtkCommand.ModifiedEvent, self.updateGUIFromParameterNode)
  File "~/git/Slicer/Slicer-SuperBuild-Release-Qt5.15.2/Slicer-build/bin/Python/slicer/util.py", line 2787, in removeObserver
    raise Exception("does not have observer")
Exception: does not have observer
[Qt] Close main MRML scene
```

PR #7442, opened 2023-12-03
BUG: Restore maximization of view on double click event

jcfr:
This commit fixes a regression introduced in 1cc3b2c (`ENH: Improve Event Delegation in qMRMLThreeDView and qMRMLSliceView`) adding missing observers.


pieper (approved):
LGTM 👍 thanks

lassoan:
Thank you, I can confirm confirm that this also fixes touchscreen interactions on Windows that got broken as well.

lassoan:
@jcfr would it be possible to create a 5.6.1 patch release within a few days?

PR #7444, opened 2023-12-05
ENH: Add methods for runTest upgrade

hannah-morilak:
This PR updates the `runTest` method inside of slicer such that now developers should not need to explicitly define each of their tests in a derived class. Additionally, it includes the usage of the setup and tear down methods, so that it functions the same way it would if you were to run them as unittests.

In the case that a develop wants the tests to run in a specific order or exclude certain tests then they can update the `moduleTestNames` variable.

cc: @jamesobutler


jcfr:
@jamesobutler When you have a chance, could you have a look ? :pray:

jamesobutler (approved):

PR #7446, opened 2023-12-07
Fix scripted loadable module importing wrapped classes

jcfr:
1. Fixes Endoscopy module dependency by adding "Markups"

2. Ensure scripted modules can always import wrapped C++ classes even if loading of loadable module is disabled. It fixes the tests `py_nowarning_[no]mainwindow_[nocli_]noloadableTest`.

3. De-duplicate code in loadable module factory by reusing qSlicerScriptedUtils.


jcfr:
Considering that:
* manual build on the branch on the Linux, macOS and Windows factory was successful :white_check_mark:
* the application starts as expected :white_check_mark:
* the failing tests are now passing :white_check_mark:

... I am bypassing our usual checks so that the round of nightly build will thoroughly tests the changes so that there can be backported to the `5.6` branch :rocket:

PR #7451, opened 2023-12-07
ENH: Disable pip version check

jcfr:
This commit addresses the quarterly update cycle of python packages bundled into Slicer. To mitigate unnecessary console clutter caused by the pip version check, the variable `PIP_DISABLE_PIP_VERSION_CHECK` is set, suppressing warnings such as:

```bash
$ PythonSlicer -m pip list
[...]
[notice] A new release of pip is available: X.Y.Z -> Z'.Y'.Z'
[notice] To update, run: pip install --upgrade pip
```

To ensure that `the PIP_DISABLE_PIP_VERSION_CHECK` environment variable is undefined when Slicer is built against a system or external Python installation, this variable is set only in the pip external project.

This also addresses an issue introduced in 028fb31 (COMP: Update python packages to latest) where the pip version was updated from `23.1.2` to `23.3`. This resulted in the following warning message instead of the expected update notification:

```bash
WARNING: There was an error checking the latest version of pip.
```

For additional details about `PIP_DISABLE_PIP_VERSION_CHECK`:
* https://pip.pypa.io/en/stable/cli/pip/#cmdoption-disable-pip-version-check
* https://pip.pypa.io/en/stable/topics/configuration/#environment-variables

### DIFF
--- a/SuperBuild/External_python-pip.cmake
+++ b/SuperBuild/External_python-pip.cmake
@@ -44,6 +44,30 @@ if(NOT Slicer_USE_SYSTEM_${proj})
       ${${proj}_DEPENDENCIES}
     )
 
+  #-----------------------------------------------------------------------------
+  # Slicer Launcher setting specific to build tree
+
+  # environment variables
+  set(${proj}_ENVVARS_LAUNCHER_BUILD
+    "PIP_DISABLE_PIP_VERSION_CHECK=1"
+    )
+  mark_as_superbuild(
+    VARS ${proj}_ENVVARS_LAUNCHER_BUILD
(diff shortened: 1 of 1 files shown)


jcfr:
After re-configuring and re-building, I confirmed that:
* the warning is not reported anymore for both `Slicer` and `PythonSlicer` (e.g by running the `list` command) :white_check_mark: 
* the env. variable `PIP_DISABLE_PIP_VERSION_CHECK` is set in both `SlicerLauncherSettings.ini` and `SlicerLauncherSettingsToInstall.in` :heavy_check_mark: 
* the env. variable is set when running `PythonSlicer`

pieper (approved):
LGTM 👍

PR #7454, opened 2023-12-08
ENH: Display instantiated module name in splashscreen

lassoan:
Module instantiation may take a long time if a module inappropriately performs lengthy operations in its main scope (e.g., pip_install or large import). It has been difficult to determine which modules misbehaved, because the splashscreen only showed the message "Instantiating modules..." but not individual module names.

This commit makes the name of the module that is being instantiated visible in the splashscreen. This way it becomes very clear which module takes disproportionately long time to instantiate.


pieper (approved):
This makes sense.  Good for debugging, but also gives users a better sense that something is happening.

PR #7460, opened 2023-12-11
[5.6] BUG: Fix temporary folder containing special characters

jcfr:
Backports changes originally contributed to `main` through the following pull requests:
* https://github.com/Slicer/Slicer/pull/7430

---------

qt.QDateTime().currentDateTime().toString(...) may use Arabic numerals on an operating system set to Arabic language. Always specify en-US locale when using when using date in file/folder names to avoid having to work with paths that may contain unicode characters (such paths may cause problems on certain file systems and it may not be always possible to pass them as command-line arguments).

Fixes problem originally reported at:
https://discourse.slicer.org/t/i-tried-everything-to-run-totalsegmentator-pls-help/33023/3

(cherry picked from commit eccb98ce1f58b19d0a429981cc8e3cf131bcf3d5)


sjh26 (approved):
lgtm

PR #7470, opened 2023-12-12
ENH: Add convenience function for node reference parameters in segment editor effects

lassoan:


Modules/Loadable/Segmentations/EditorEffects/qSlicerSegmentEditorAbstractEffect.h
```diff
@@ -341,6 +345,13 @@ public slots:
   /// \sa setCommonParameter
   Q_INVOKABLE void setCommonParameterDefault(QString name, double value);
 
+  /// Convenience function to set double parameter
+  /// \param name Parameter name string
+  /// \param value Parameter value double
```
jcfr, line 350:
```suggestion
  /// Convenience function to set node reference parameter
  /// \param name Parameter name string
  /// \param value Parameter node reference
```

Modules/Loadable/Segmentations/EditorEffects/qSlicerSegmentEditorAbstractEffect.h
```diff
@@ -341,6 +345,13 @@ public slots:
...
+  Q_INVOKABLE void setNodeReference(QString name, vtkMRMLNode* node);
+  /// Convenience function to set double common parameter \sa setCommonParameter
```
jcfr, line 352:
```suggestion
  /// Convenience function to set node reference common parameter \sa setCommonParameter
```

jcfr:
I am wondering how we could make the interface for getting/setting parameter value available more generally.

Maybe the class `vtkMRMLScriptedModuleNode`[^1] could be improved adding:
* `SetParameterAsDouble / GetParameterAsDouble`
* `SetParameterAsNodeReference / GetParameterAsNodeReference`

Also since `ModuleName` is already associated with the scripted node, if set it could probably be automatically prepended as `<ModuleName>.` ?

[^1]: https://github.com/Slicer/Slicer/blob/main/Libs/MRML/Core/vtkMRMLScriptedModuleNode.h

lassoan:
> I am wondering how we could make the interface for getting/setting parameter value available more generally.

We should generalize even beyond scripted module nodes and have typed properties in all MRML nodes. That would allow a lot of code simplification, introspection, recording/replay of scene changes, etc. However, this would probably require a Qt-like preprocessor, macros, or waiting for the reflection support in C++.

lassoan:
@jcfr could you review/approve? Thank you!

jcfr (approved):
Thanks for cleanup the implementation and addressing the comments :ok_hand:

PR #7473, opened 2023-12-13
BUG: Fix error handling in slicer.util.downloadFile

lassoan:


Base/Python/slicer/util.py
```diff
@@ -3385,7 +3385,7 @@ def downloadFile(url, targetFilePath, checksum=None, reDownloadIfChecksumInvalid
     try:
         (algo, digest) = extractAlgoAndDigest(checksum)
     except ValueError as excinfo:
-        logging.error("Failed to parse checksum: " + excinfo.message)
+        logging.error("Failed to parse checksum: " + str(excinfo))
```
jcfr, line 3388:
```suggestion
        logging.error(f"Failed to parse checksum: {excinfo}")
```

Base/Python/slicer/util.py
```diff
@@ -3423,9 +3423,9 @@ def downloadFile(url, targetFilePath, checksum=None, reDownloadIfChecksumInvalid
                                   "\n  expected checksum: %s" % (current_digest, digest))
                     return False
             else:
-                logging.info("Requested file has been found and checksum is OK: " + targetFilePath)
+                logging.info("Requested file has been found and checksum is OK: " + str(targetFilePath))
```
jcfr, line 3426:
```suggestion
                logging.info(f"Requested file has been found and checksum is OK: {targetFilePath}")
```

Base/Python/slicer/util.py
```diff
@@ -3423,9 +3423,9 @@ def downloadFile(url, targetFilePath, checksum=None, reDownloadIfChecksumInvalid
...
         else:
-            logging.info("Requested file has been found: " + targetFilePath)
+            logging.info("Requested file has been found: " + str(targetFilePath))
```
jcfr, line 3428:
```suggestion
            logging.info(f"Requested file has been found: {targetFilePath}")
```

lassoan:
Thanks for the comments @jcfr. I've been thinking about switching to using f-strings when I fixed the issue but I could not find a strong reason for it, mostly just to be consistent with similar log message generations in this file. But in those cases the placeholder was inside the string; and f-strings were not used everywhere (and some places used %). Do you have a good reason that justifies switching to f-strings? If yes, then we could add that to our coding guidelines as well and apply it whenever we touch similar code.

jcfr:
Re: f-string
It generally facilitate reading and update of strings with placeholders.

Ar the exception of:
- https://docs.astral.sh/ruff/rules/f-string-in-exception/

PR #7477, opened 2023-12-16
BUG: Fix loading of SubjectHierarchyPlugins at startup

jcfr:
This commit addresses the incorrect loading of SubjectHierarchyPlugins during startup, as well as the regression identified in the `py_SubjectHierarchyCorePluginsSelfTest`. Both issues were introduced in in 1d77e473a (`BUG: Fix language translation of subject hierarchy plugins`).

To resolve these issues, it updates the folder structure to align with the changes introduced in fc85b2f05 (`BUG: Make Segment Editor effect names translatable`).

This structural adjustment ensures proper loading of `SubjectHierarchyPlugins`.


jcfr:
@lassoan I am testing the changes locally :hourglass_flowing_sand: 

In the meantime, I also temporarily disabled the nightly builds. I will re-trigger[^1] them once I have tested and integrated the fix :rocket: .

[^1]: https://github.com/Slicer/DashboardScripts/blob/main/maintenance/guides/re-trigger-nightly-build.md

Modules/Loadable/SubjectHierarchy/Widgets/Python/SubjectHierarchyPlugins/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
+
+# Deploy abstract scripted plugin and init file that
+# imports all deployed scripted plugins deployed by modules
+set(SubjectHierarchyPlugins_PYTHON_SCRIPTS
+  __init__
+  AbstractScriptedSubjectHierarchyPlugin
+  parameterNodeWrapper/__init__
```
lassoan, line 7: `Modules\Loadable\SubjectHierarchy\Widgets\Python\SubjectHierarchyPlugins\parameterNodeWrapper` folder is misplaced: `parameterNodeWrapper` has nothing to do with `SubjectHierarchyPlugins`. It is a ParameterNodeWrapper plugin, therefore it must be in `Modules\Loadable\SubjectHierarchy\Widgets\Python\ParameterNodeWrapperPlugins` folder. Note that the folder name was not following Python package naming conventions either (starting with lowercase), the proposed name fixes that, too.

Modules/Loadable/SubjectHierarchy/Widgets/Python/SubjectHierarchyPlugins/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
...
+  parameterNodeWrapper/guiConnectors
```
lassoan, line 8: Since `ParameterNodeWrapperPlugins` is a shared folder name in the install tree (the same way as all other Python plugin folders), the filenames in it must be unique. Therefore `guiConnectors.py` is not a valid filename, it has to be `SubjectHierarchyGuiConnectors.py`.  Again, the naming convention was violated with guiConnectors.py (started with lowercase). `SubjectHierarchyGuiConnectors.py` is not perfect, because GUI should be all caps, but that would require changing of the class name as well, so I would not do it in this commit.

Modules/Loadable/SubjectHierarchy/Widgets/Python/SubjectHierarchyPlugins/parameterNodeWrapper/__init__.py
lassoan, line 1: This is a shared folder in the install tree, therefore the __init__.py that gets intalled cannot contain hardcoded list of imports but must iterate through the filenames in the folder, similarly to the __init__.py file in the SubjectHieararchyPlugins folder.

lassoan (commented):
Thanks for working on this. Your changes are very similar to what I've been working on, but you have not addressed the numerous issues with the parameterNodeWrapper plugin files (wrong folder name, filename, incorrect import). You can find the full implementation here:

https://github.com/Slicer/Slicer/compare/main...lassoan:Slicer:fix-sh-python-plugins?expand=1

Since the changes affect the same files, it could make sense to address them in a single PR (I would say even in the same commit), but if you prefer then we can fix the parameterNodeWrapper issues in a follow-up PR.

Modules/Loadable/SubjectHierarchy/Widgets/Python/SubjectHierarchyPlugins/CMakeLists.txt
jcfr: Since the `parameterNodeWrapper` package is namespaced within `SubjectHierarchyPlugins`, i don't think it is an issue. This is the intended approach.

jcfr: I suggest we address the `parameterNodeWrapper` in a follow up commit. This PR fixes the regression without impacting the `parameterNodeWrapper`

Modules/Loadable/SubjectHierarchy/Widgets/Python/SubjectHierarchyPlugins/parameterNodeWrapper/__init__.py
jcfr: You are right this was on oversight in 19f787127af40a4ee27e7fdd4b7a19906e814de5 and at that time, I missed the fact there was a "shared" folder.

jcfr:
Merging, I confirmed the regression is fixed in both build and installed tree.

Modules/Loadable/SubjectHierarchy/Widgets/Python/SubjectHierarchyPlugins/CMakeLists.txt
lassoan: It is true that it may work as is, but it is not the right thing to do.

SubjectHierarchyPlugins folder must not contain ParameterNodeWrapper Plugins. Probably it was misleading that the SubjectHierarchyPlugins folder name could be interpreted as "a folder that contains all Python plugins for the SubjectHierarchy module". But in fact this is "a shared folder for storing SubjectHierarchyPlugins in any module". To be consistent with how all the other Python plugins are defined in Slicer, ParameterNodeWrapperPlugins folder should contain ParameterNodeWrapper plugins of all Slicer core modules.

Modules/Loadable/SubjectHierarchy/Widgets/Python/SubjectHierarchyPlugins/CMakeLists.txt
jcfr: I would argue that the approach to register the `parameterNodeWrapper` is more robust as the registration does not require the files to be physically located in the same folder to ensure proper registration. It is done through the `parameterNodeGuiConnector` decorator.

[^1]: https://github.com/Slicer/Slicer/blob/ff58cc7a4e013bb6f46bd867b6d09f920770be63/Base/Python/slicer/parameterNodeWrapper/guiConnectors.py#L117-L152

The current mechanism for `SubjectHierarchyPlugins` is brittle and it may be sensible to also adopt a decorator based registration.

> re: `gui` vs `GUI` naming

This was discussed a while back and if I recall, similarly to the fact we have `vtkSomething`, `qMRMLSomething`, `mrml`, ... it started with `gui` lower cased. I don't think we should change this.

lassoan: ParameterNodeWrapper stuff does not belong into SH plugins folder. It could go to a (moduleName)Lib folder if we don't want a shared folder. I would use a shared folder for consistency with other plugins. If we chose to use (moduleName)Lib folder for plugins instead of a shared folder (which could make sense) then we should do it for all plugins, such as SegmentEditor, SubjectHierarchy, SegmentStatistics, plugins.

jcfr: To follow-up on this discussion, the registration of parameterNodeWrapper plugins has been improved:
* https://github.com/Slicer/Slicer/pull/7480

PR #7479, opened 2023-12-16
ENH: Do not require file type in sample data specification

lassoan:
Previously, `VolumeFile` type was used by default if type was not explicitly specified for a sample data item. Since readers can now provide confidence value, the default type is most often correct and it is no longer necessary to specify file type.

### DIFF
--- a/Modules/Scripted/SampleData/SampleData.py
+++ b/Modules/Scripted/SampleData/SampleData.py
@@ -26,7 +26,7 @@ def downloadFromURL(uris=None, fileNames=None, nodeNames=None, checksums=None, l
     :param checksums: Checksum(s) formatted as ``<algo>:<digest>`` to verify the downloaded file(s). For example, ``SHA256:cc211f0dfd9a05ca3841ce1141b292898b2dd2d3f08286affadf823a7e58df93``.
     :param loadFiles: Boolean indicating if file(s) should be loaded. By default, the function decides.
     :param customDownloader: Custom function for downloading.
-    :param loadFileTypes: file format name(s) ('VolumeFile' by default).
+    :param loadFileTypes: file format name(s) (if not specified then the default file reader will be used).
     :param loadFileProperties: custom properties passed to the IO plugin.
 
     If the given ``fileNames`` are not found in the application cache directory, they
@@ -146,7 +146,8 @@ class SampleDataSource:
     def __init__(self, sampleName=None, sampleDescription=None, uris=None, fileNames=None, nodeNames=None,
                  checksums=None, loadFiles=None,
                  customDownloader=None, thumbnailFileName=None,
-                 loadFileType=None, loadFileProperties=None):
+                 loadFileTypes=None, loadFileProperties=None,
+                 loadFileType=None):
         """
         :param sampleName: Name identifying the data set.
         :param sampleDescription: Displayed name of data set in SampleData module GUI. (default is ``sampleName``)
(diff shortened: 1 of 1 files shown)


Modules/Scripted/SampleData/SampleData.py
```diff
@@ -157,16 +158,22 @@ def __init__(self, sampleName=None, sampleDescription=None, uris=None, fileNames
...
         :param customDownloader: Custom function for downloading.
-        :param loadFileType: file format name(s) ('VolumeFile' by default if node name is specified).
+        :param loadFileTypes: file format name(s) (if not specified then the default file reader will be used).
         :param loadFileProperties: custom properties passed to the IO plugin.
+        :param loadFileType: deprecated, use ``loadFileTypes`` instead.
         """
+
+        # For backward compatibility (allow using "loadFileType" instead of "loadFileTypes")
+        if (loadFileType is not None) and (loadFileTypes is None):
+            loadFileTypes = loadFileType
```
jcfr, line 168: I suggest to raise an exception[^1] if both `loadFileType` and `loadFileTypes` are specified. Updated code should only use `loadFileTypes`

[^1]: I suggest `ValueError`. See https://stackoverflow.com/questions/256222/which-exception-should-i-raise-on-bad-illegal-argument-combinations-in-python

Modules/Scripted/SampleData/SampleData.py
```diff
@@ -178,24 +185,9 @@ def __init__(self, sampleName=None, sampleDescription=None, uris=None, fileNames
...
+            loadFileTypes = [loadFileTypes]
             checksums = [checksums]
 
-        updatedFileType = []
-        for fileName, nodeName, fileType in zip(fileNames, nodeNames, loadFileType):
-            # If not explicitly specified, attempt to guess fileType
-            if fileType is None:
-                if nodeName is not None:
-                    # TODO: Use method from Slicer IO logic ?
-                    fileType = "VolumeFile"
```
jcfr, line 190: What about keeping this logic and instead have something like
```
                    fileType = None  # the fileType will be set automatically based on the registered readers.
```

Modules/Scripted/SampleData/SampleData.py
```diff
@@ -684,6 +692,14 @@ def downloadFromSource(self, source, maximumAttemptsCount=3):
                     continue
                 resultFilePaths.append(filePath)
 
+                # Special case for backward compatibility for loading scene and zip files when loadFileType is not specified
```
jcfr, line 695: I suggest to keep the logic  in the `SampleDataSource` and update the code as described in https://github.com/Slicer/Slicer/pull/7479/files#r1428946872

jcfr: Does the comment mean that if the filetype is explicitly set to `SceneFile` or `ZipFile`, the behavior will be different ? 

I would suggest to have the same behavior .. may be the comment should instead be:

```

                # Special behavior to handle "ZipFile" and "SceneFile"
```

Modules/Scripted/SampleData/SampleData.py
```diff
@@ -916,11 +937,15 @@ def loadScene(self, uri, fileProperties={}):
...
         return True
 
-    def loadNode(self, uri, name, fileType="VolumeFile", fileProperties={}):
+    def loadNode(self, uri, name, fileType=None, fileProperties={}):
         self.logMessage("<b>" + _("Requesting load {name} from {uri} ...").format(name=name, uri=uri) + "</b>")
 
         fileProperties["fileName"] = uri
         fileProperties["name"] = name
+        if not fileType:
+            print(f"loadNode: not fileType, uri = {uri}")
```
jcfr, line 946:
```suggestion
```

Modules/Scripted/SampleData/SampleData.py
```diff
@@ -916,11 +937,15 @@ def loadScene(self, uri, fileProperties={}):
...
+            fileType = slicer.app.coreIOManager().fileType(fileProperties["fileName"])
+            print(f"loadNode: fileType = {fileType}")
```
jcfr, line 948:
```suggestion
```

Modules/Scripted/SampleData/SampleData.py
```diff
@@ -232,14 +224,16 @@ def __str__(self):
...
+            self.fileNames, self.uris, self.nodeNames, self.loadFiles, self.loadFileTypes, self.checksums):
+
             output.extend([
-                " fileName    : %s" % fileName,
-                " uri         : %s" % uri,
-                " checksum    : %s" % checksum,
-                " nodeName    : %s" % nodeName,
-                " loadFile    : %s" % loadFile,
-                " loadFileType: %s" % fileType,
+                " fileName     : %s" % fileName,
```
jcfr, line 231: I suggest to add a first commit titled `STYLE: Improve SampleData source code readability` where all stylistic changes would be done, we will then add to it the `.git-blame-ignore-revs`[^1] file once the MR is merged.

Modules/Scripted/SampleData/SampleData.py
lassoan: The behavior has not changed, the logic just moved to lower level.

Modules/Scripted/SampleData/SampleData.py
lassoan: Thank you, good catch.

Modules/Scripted/SampleData/SampleData.py
lassoan: Thank you, good catch.

lassoan:
@jcfr Thanks for the review and suggestions. I've responded to all comments, please have a look.

PR #7485, opened 2023-12-17
COMP: Fix generation of additional launcher settings for list-based settings

jcfr:
This commit addresses two issues:

## Use of `string(REPLACE ...)`

The use of the `string(REPLACE ...)` command introduced in commits e9f28d6c8  and b1efab8a6 led caused input variable representing a list to be concatenated as a single string. This commit addresses this by transitioning to the `TRANSFORM` action allowing to replace values individually within a list.

Also addresses the missing replacement of `<CMAKE_CFG_INTDIR>` associated with the `PYTHONPATH_LAUNCHER_BUILD` settings.

| Before & After |
|---|
| image |

## Use of incorrect regular expression to replace `$(Configuration)`

Fixes the regular expression used for replacing `$(Configuration)` with `${CMAKE_CFG_INTDIR}` so that the placeholder variable is ultimately replaced through the execution of the custom command associated with the `ConfigureAdditionalLauncherSettings` custom target.

| Before | After |
|--|--|
| image | image |
| image | image |

### DIFF
--- a/Extensions/CMake/SlicerBlockAdditionalLauncherSettings.cmake
+++ b/Extensions/CMake/SlicerBlockAdditionalLauncherSettings.cmake
@@ -92,14 +92,15 @@ if(NOT TARGET ConfigureAdditionalLauncherSettings AND _configure_additional_laun
   # External projects - library paths
   foreach(varname IN LISTS ${SUPERBUILD_TOPLEVEL_PROJECT}_EP_LABEL_LIBRARY_PATHS_LAUNCHER_BUILD)
     set(value ${${varname}})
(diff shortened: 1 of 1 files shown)


jcfr:
:white_check_mark:  Local testing confirmed the fixes work as expected.

jcfr:
While local testing allows to confirm that the list are not concatenated :heavy_check_mark: 

I also noticed that the logic originally introduced to replace may not be working as expected:

image

jcfr:
The additional launcher settings now include the expected value:

image

jamesobutler (approved):

PR #7486, opened 2023-12-17
COMP: Re-generate additional launcher settings if settings are updated

jcfr:
This commit ensures that the custom command responsible for generating `AdditionalLauncherSettings.ini` is rerun whenever the project is re-configured with different options. This is important to account for changes in the set of paths associated with the launcher settings.


jcfr:
This was discovered while testing:
* https://github.com/Slicer/Slicer/pull/7485

jcfr:
Note that the settings associated with the application are not subject to this issue. These are generated using `ctkAppLauncherConfigureForTarget()`[^1] with its custom command dependency properly specified.

[^1]: https://github.com/commontk/AppLauncher/blob/dba5a5c98d2f890550a1641a19ffc1b0d3bb8472/CMake/ctkAppLauncher.cmake#L544-L547

jamesobutler (approved):

PR #7491, opened 2023-12-18
Draft: Enable MRML to be built as a standalone Python Wheel

AlexyPellegrini:
## Goal

Being able to use Slicer MRML as a library in "pure" python projects without the PythonSlicer, Slicer Application GUI, etc...

It also enable Slicer to be built without Qt and CTK. This saves a good amount of build time.

## Build

Build Slicer with `Slicer_BUILD_GUI=OFF` (`Slicer_BUILD_MRML_WHEEL` is `ON` by default for now).  
Working on Windows, to build the wheels I used the following commands:

### VTK:
```
cd VTK-build
py -3.9 -m delvewheel repair --add-path ".\bin\Release;..\tbb-install\redist\intel64\vc14" "dist/vtk_slicer-*-win_amd64.whl"
```

Generated wheel package name is `vtk_slicer`.

### Slicer MRML:
```
cd Slicer-build\MRMLWheel
py -3.9 -m delvewheel repair --add-path ../bin/Release;..\..\VTK-build\bin\Release;..\..\ITK-build\bin\Release;..\..\tbb-install\redist\intel64\vc14;..\..\teem-build\bin\Release;..\..\DCMTK-build\bin\Release;..\..\LibArchive-build\bin\Release;..\..\SlicerExecutionModel-build\ModuleDescriptionParser\bin\Release; dist\Slicer-1.0-py3-none-any.whl
```

## Usage

Install the two wheels.

### 1) DLL mangling

I only tried on Windows. But I think this issues will be the same on other systems.  
Delvewheel mangles DLLs name using both the DLL content but also the package name. This means that `vtk_slicer` and `slicer` both have all required VTK dlls in their `<package>.libs` folder, and they have a different mangling due to the package name.

So we have two times the exact same VTK dll, with different names. I don't know if it is possible for two wheels to share the same native libs, but this would be the best approach, if possible.

### 2) vtkObject

Everything works fine until I run a script that import slicer and vtk, both installed using the built wheels. 
```py
from slicer import *
from vtk import *
object = vtkMRMLScene()
assert isinstance(object, vtkObject) # fails...
```

I read the comment in `__init__.py` that it may fail if slicer is imported before vtk. But even with what seems to be the right import order I still have this issue.
If someone has hints about what could be the issue. Maybe it is related to previous issue tho.

---

@Thibault-Pelletier @finetjul @jcfr @lassoan


pieper:
Thanks for working on this, it's something we've wanted for a long time 👍 

As you mention, it would be great to have a clean layering of dependencies to avoid duplicated libraries, like wheel packages building from the ground up, where VTK and ITK wheels use the same jpeg, zlib, etc and then MRML uses VTK and ITK and then the various Slicer modules and app wheel would use those by adding in Qt and CTK (and then extensions can be built on top of those).  I think that's a bit hard to achieve right now, but this seems like a step in the right direction.  

In the mean time, just having MRML wheels like we have the slicerio package could be very useful for interoperability.

Is there an issue that describes the plan for this work?

jcfr (changes requested):
Thanks for working on this, having these wheels will be important moving forward.

Few suggestions to move forward:

Instead of modifying the complete build system and introducing option like `Slicer_BUILD_GUI` or `Slicer_BUILD_MRML_WHEEL`, we should instead add support for directly building the following directories as independent projects:
* `Libs/vtkSegmentationCore`
* `Libs/vtkITK`
* `Libs/vtkTeem`
* `Libs/MRML/Core`
* `Libs/MRML/Logic`
* `Libs/MRML/DisplayableManager`

This means that it should be possible to do the following:

```
cd Libs/vtkSegmentationCore
cmake \
  -DVTK_DIR:PATH=/path/to/VTK-build \
  -B /tmp/vtkSegmentationCore
  -S .
```

Then for projects depending on other projects (e.g `MRML/Core` depends on `vtkSegmentationCore`, these would expect the `<dependency>_DIR` variable (e.g `vtkSegmentationCore_DIR:PATH=/path/to/vtkSegmentationCore-build`.

To help with this and avoid re-inventing the "wheel", we should rely on the fact we can:
1. already build VTK modules against other VTK modules
2. build such VTK module as python wheels

Such example of projects are `vtkAddon`[^1], `VTKU3DExporter`[^2] or `LookingGlassVTKModule`[^3]

[^1]: https://github.com/Slicer/vtkAddon/tree/main-pythonpackage
[^2]: https://github.com/ClinicalGraphics/VTKU3DExporter
[^3]: https://github.com/Kitware/LookingGlassVTKModule

To help build any regular VTK modules, we have also developed [KitwareMedical/VTKExternalModule](https://github.com/KitwareMedical/VTKExternalModule) which avoid duplicating a lot of boiler plate code and streamline overall maintenance. Ideally, this support (like it is the case for ITK) should be part of VTK proper, but that is topic for another day.

That said, there are some challenges:
* **build-system:**: we can not expect (yet) to convert the regular Slicer build-system to fully adopt the VTK build-system. Instead each one of these project should support both.
* **pyproject.toml**: example listed above are using `setup.py` and scikit-build "classic". These should be updated to use the new `scikit-build` backend[^4] we have been developing and that integrated with `pyproject.toml`. To help with this, we have some scikit-build funding we can allocate to migrate few projects and help identify and fix limitation/issues of the new scikit-build backend.
* **CI**: continuous integration of all of these project need to be centralized and maintainable. To address this, we will integrate dedicated GitHub workflow responsible to build/test/publish each wheels directly in `Slicer/.github/workflows`
* **Naming**: We will have to come up with name for these wheels.
[^4]: https://scikit-build-core.readthedocs.io

finetjul:
Currently Slicer is using a custom fork of VTK. Is the VTK module approach going to work with a VTK official python wheel ? Is the Slicer fork of VTK available on pypi ?

jcfr:
> Is the VTK module approach going to work with a VTK official python wheel ? Is the Slicer fork of VTK available on pypi ?

Yes, that is the plan. The wheels will be built against the official VTK and depends on it. The only Slicer-specific changes we maintain[^1] in our fork are not relevant in this context.

[^1]: https://github.com/Slicer/VTK/pull/46#issuecomment-1859631144

lassoan:
It is awesome to see that this work has begun! If MRML becomes accessible as a standalone Python package then I would find it appropriate to move it out to a separate repository. Slicer could build it the same way as vtkAddon. Having MRML in a separate library would make changes a bit less convenient, but it would encourage to have looser coupling with Slicer, so it would worth it.

I would consider putting MRML core, logic, and displayable manager components into one Python package, as they are quite tightly linked and they all have the same dependencies. This would also limit the proliferation of Python packages when we'll make Slicer modules available as Python packages: we would not want to separately maintain a SlicerMarkupsMRMLCore, SlicerMarkupsMRMLLogic, SlicerMarkupsMRMLDM, SlicerMarkupsVTKWidgets Python packages. Even GUI components could be included in the same package (but as an optional extra), because MRML nodes, logic, and GUI are all quite tightly linked (we would not want to develop, maintain, version them independently).

PR #7497, opened 2023-12-19
ENH: Improve handling of invalid input in MRMLViewInteractorStyle classes

jcfr:
Within the `DelegateInteractionEventToDisplayableManagers` function:

- Introduce a check for `pokedRenderer`, mirroring the one initially incorporated in `vtkMRMLThreeDViewInteractorStyle` via commit 364ff7b6d (ENH: Use common interactor style for slice and 3D views).

- Introduce (or relocate) a check for `inputEventData` at the beginning of the `Delegate` function for consistency with the update to `vtkMRMLSliceViewInteractorStyle` in commit 731be146e (BUG: Fix crash when creating and showing qMRMLSliceWidget using Python). This adjustment also prevents the invocation of `FindPokedRenderer` when there is no `inputEventData`.


sankhesh (approved):
LGTM

PR #7506, opened 2023-12-22
BUG: Fix empty candidate node list in Sequences module

lassoan:
In Sequences module's Edit tab, data node candidate node list ("Add/remove data nodes") sometimes remained empty. Fixed the issue by more robustly checking if the list needs to be updated.

Also fixed a minor issue in auto-advance not working when adding the first data node to the sequence.

### DIFF
--- a/Modules/Loadable/Sequences/Widgets/qMRMLSequenceEditWidget.cxx
+++ b/Modules/Loadable/Sequences/Widgets/qMRMLSequenceEditWidget.cxx
@@ -65,6 +65,7 @@ public:
   vtkWeakPointer<vtkMRMLSequenceNode> SequenceNode;
   // data node class name that was used for populating the candidate node list
   QString DataNodeCandidatesClassName;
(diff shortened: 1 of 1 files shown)


Modules/Loadable/Sequences/Widgets/qMRMLSequenceEditWidget.cxx
```diff
@@ -65,6 +65,7 @@ class qMRMLSequenceEditWidgetPrivate
   vtkWeakPointer<vtkMRMLSequenceNode> SequenceNode;
   // data node class name that was used for populating the candidate node list
   QString DataNodeCandidatesClassName;
+  bool DataNodeCandidatesNeedsUpdate{true};
```
jcfr, line 68: Should the variable be called `DataNodeCandidatesNeedUpdate` ? (since there are more than one candidates)

lassoan: Thanks, fixed.

PR #7515, opened 2023-12-29
ENH: Add option to display shadows in 3D views

lassoan:
Screen-space ambient occlusion has been recently added to VTK for volume rendering (it was added a few years ago for surface rendering). This is a computationally inexpensive way of displaying shadows in 3D views.

This commit adds this as an option to 3D view nodes.

The feature can be enabled in the view controller. It can be enabled by default in application settings (View section).

The option is disabled by default, as it slightly increases computation time and current rendering presets are not yet optimized for it (they tend to be a bit too dark if shadows are added).

Source of ViewShadow.svg:
Google Material icon set (shadow_FILL0_wght400_GRAD0_opsz24.svg) https://fonts.google.com/icons?selected=Material%20Symbols%20Outlined%3Ashadow%3AFILL%400%3Bwght%40400%3BGRAD%400%3Bopsz%4024


pieper (dismissed):
I didn't try compiling this, but other than the spelling issue it looks great - thanks for adding this!

jcfr:
cc: @LucasGandel @sankhesh

jamesobutler:
vtkMRMLCameraDisplayableManagerTest1 is failing (https://slicer.cdash.org/test/26353921) since the saved MRML file includes a few new unexpected entries compared to the baseline scene file:
`shadowsVisibility="false" ambientShadowsSizeScale="0.3" ambientShadowsVolumeOpacityThreshold="0.25"`

lassoan:
@jamesobutler Good catch, thank you, I'll push a fix.

PR #7521, opened 2024-01-05
DOC: Update installation and building instruction for ArchLinux

hubutui:


pieper (changes requested):
I'm not familiar with the AUR approach, but it does look simpler.

But wouldn't there also be people who want to build an a more traditional way by installing the prereqs?  It seems like leaving in the instructions for a local build would be handy, for example if someone wants to make a debug build.

hubutui:
AUR only provides the `PKGBUILD` file, a shell script containing the build information. One could manually run the commands one by one we use in the "prepare" and "build" to compile the package. ArchLinux provides a utility named [makepkg](https://wiki.archlinux.org/title/Makepkg)  to make these steps easier.

pieper (approved):
Okay, if that's what ArchLinux users would expect that's fine with me.

PR #7524, opened 2024-01-08
BUG: Fix crash when module category name matches a module name

### LINKED ISSUE #7523: Cannot start after setting language to Chinese
## Summary
When the language is set to **Simplified Chinese** in **Utilities\Language Tools**, the program cannot be opened after restarting, and the Windows problem report shows that `SlicerApp-real.exe` has stopped working. Switching to **Japanese** will display Japanese normally, switching to **Traditional Chinese** will display English, but it can start normally.

After careful investigation, after deleting the **Slicer_zh-CN.qm** file in the `%LocalAppData%\slicer.org\Slicer 5.6.1\bin\translations`  directory, the program can start normally.

## Steps to reproduce

1. Press Ctrl+4, enter **Extensions Manger**, search for "language", you can see a utility called **LanguagePacks**, restart the software after installation
2. Press Ctrl+2, open **Application Setting**, check **internationalization**, and restart the software.
3. Click **Utilities**→**Language Tools** in the module options, select **Weblate** in the Input translations column, click the **refresh** button, you can see a list of many languages, check **Japanese**, **Simplified Chinese** and **Traditional Chinese**, click **Update translation files** , the translation file will start to download automatically. After downloading, restart the software
4. Open **Language Tools** again, select **Application language** as **Chinese (China)** at the top, click **Restart the application** on the right, and the software will restart.

repo1

5. When starting the software, text such as the initialization module will be displayed, but when it reaches the `Loading module "SegmentEditor"... `stage, it will crash.
repo2

6. Also select the **second** Chinese (China). After restarting, the software can be started normally but it is displayed in **English**.

7. After selecting **Japanese**, the software starts normally and the translation takes effect.
repo3

##  Environment
System: Windows 10 Chinese version
System version: 22H2(19045.3803)
Slicer version: 5.6.1 Stable Release (You see there is no utility named Language Tools in Preview Release )

## Problem log of Windows system
Source: SlicerApp-real.exe
Summary: Stopped working
Date: 2014-01-‎08 23:22
Status: Report Sent
Wrong application path: C:\Users\1233\AppData\Local\slicer.org\Slicer 5.6.1\bin\SlicerApp-real.exe
Problem event name: APPCRASH
Application name: SlicerApp-real.exe
Application version: 0.0.0.0
Application timestamp: 65633afb
Faulty module name: Qt5Widgets.dll
Faulty module version: 5.15.2.0
Faulty module timestamp: 5fa4de1c
Exception code: c0000005
Exception offset: 0000000000032fe9
OS version: 10.0.19045.2.0.0.768.99
Locale ID: 2052
Additional information 1: 6cfd
Additional information 2: 6cfdfc8dd99a7ae0ef76343ef9579a3f
Additional information 3: f94b
Additional information 4: f94bf039ba88192a82b805ca52b38915
Bucket ID: 4ebf9e4a2a29d98d46356509a199b490 (1600296334640395408)

lassoan:
In one of the language translations, "Segmentation" module category name was translated with the same word as "Segmentations" module name, which caused Slicer to crash. Added a check to log a warning if such name conflict occurs and put the module in the top-level category.

fixes #7523

### DIFF
--- a/Base/QTGUI/qSlicerModulesMenu.cxx
+++ b/Base/QTGUI/qSlicerModulesMenu.cxx
@@ -327,7 +327,13 @@ QMenu* qSlicerModulesMenuPrivate::menu(QMenu* menu, QStringList subCategories, b
     {
     if (action->text() == category)
       {
-      return this->menu(action->menu(), subCategories);
+      QMenu* submenu = action->menu();
+      if (!submenu)
+        {
+        // This action is a module (it does not have a menu)
+        return nullptr;
+        }
+      return this->menu(submenu, subCategories);
       }
     }
   // if we are here that means the category has not been found, create it.
@@ -556,6 +562,11 @@ void qSlicerModulesMenu::addModule(qSlicerAbstractCoreModule* moduleToAdd)
   foreach(const QString& category, module->categories())
     {
     QMenu* menu = d->menu(this, category.split('.'), module->isBuiltIn());
+    if (!menu)
+      {
+      qWarning() << "Failed to add module" << module->name() << "to the menu" << category << "because this string is already used as module name.";
+      menu = this;
+      }
     d->addModuleAction(menu, moduleAction, true, module->isBuiltIn());
     }
 


Base/QTGUI/qSlicerModulesMenu.cxx
```diff
@@ -556,6 +562,11 @@ void qSlicerModulesMenu::addModule(qSlicerAbstractCoreModule* moduleToAdd)
   foreach(const QString& category, module->categories())
     {
     QMenu* menu = d->menu(this, category.split('.'), module->isBuiltIn());
+    if (!menu)
+      {
+      qWarning() << "Failed to add module" << module->name() << "to the menu" << category << "because this string is already used as module name.";
```
pieper, line 567:
```suggestion
      qWarning() << "Failed to add module " << module->name() << " to the menu " << category << " because this string is already used as module name.";
```

pieper (approved):
Looks good - thanks for the quick fix 👍 

Just the small change suggestion.

Base/QTGUI/qSlicerModulesMenu.cxx
lassoan: `qWarning` streaming operator adds those spaces automatically (unless some special directive is specified).

I find this behavior unintuitive, too. I see that it can make things a little more convenient, but it also makes some things more complicated (e.g., if you want to use your own separators or you want to put quotes around strings).

pieper: Interesting, okay, I didn't know that. 👍   I searched for qWarning in the Slicer source code and saw we typically use spaces so I thought we should be consistent.  Looking at it now I think we probably shouldn't rely on that feature since it's a special case that could lead to error in other kinds of messages.

KryanWang: How can i use these codes please

lassoan: You don't need to do anything with this code. The fix is integrated and available in the latest Slicer version.

KryanWang: I don't know anything about code at all, but I really need this to fix my 3Dslicer，I would really appreciate it if you could help me fix my 3Dslicer。o(╥﹏╥)o

lassoan: It is already fixed in the latest Slicer Preview Release. It is avaliable for download.

It would be nice if someone would also fix the Chinese translation on [Weblate](https://hosted.weblate.org/translate/3d-slicer/3d-slicer/zh_Hans/?q=&sort_by=-priority%2Cposition&checksum=) by translating "Segmentation" (module category) and "Segmentations" (module name) words with two different words, because the crash is due to that they both translated as 分割

KryanWang: I understand. I downloaded the Stable Release of the software. I'll try reinstalling a Preview Release now. Thank you very much!

KryanWang: I did it! Thank you very much!

PR #7527, opened 2024-01-09
ENH: Allow dragging of splashwindow

lassoan:
Splashwindow is a topmost window that is shown during application startup. Since it is a topmost window that appears in the screen center above all other applications, it may make the computer hard to use for anything else.

This commit makes the splashwindow less obstructive: The splashwindow can be moved by clicking on it (anywhere) and dragging it out of the way. Clicking it also removes "topmost" attribute from splashwindow, so it is no longer forced to appear above everything else.

@muratmaga this pull request (along with [more detailed messages during startup](https://github.com/Slicer/Slicer/commit/f218e8bab2fef7ce7f47a47989825dd4f143e3ce) that pinpoints any bad-beheaving modules) should address your complaints about the Slicer's splashwindow preventing people from using their computers.

### DIFF
--- a/Base/QTApp/qSlicerApplicationHelper.txx
+++ b/Base/QTApp/qSlicerApplicationHelper.txx
@@ -25,6 +25,7 @@
 #include <qSlicerApplication.h>
 
 // Qt includes
+#include <QMouseEvent>
 #include <QSettings>
 #include <QSplashScreen>
 #include <QTimer>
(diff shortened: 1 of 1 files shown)


pieper (approved):
Looks nice 👍

Base/QTApp/qSlicerApplicationHelper.txx
```diff
@@ -48,6 +49,58 @@
...
+  bool eventFilter(QObject* obj, QEvent* event) override
+    {
+    if (event->type() == QEvent::MouseButtonPress && this->Widget)
+      {
+      QMouseEvent* mouseEvent = static_cast<QMouseEvent*>(event);
+      this->PressPosition = mouseEvent->pos();
+      this->Dragging = true;
+      if (this->DisableTopMost)
+        {
+        this->Widget->setWindowFlags(this->Widget->windowFlags() & ~Qt::WindowStaysOnTopHint);
```
jcfr, line 72: To ensure the "widget" is in the front, what about setting `Qt::WindowActive`[^1] in additional of removing `Qt::WindowStaysOnTopHint`

Alternatively, calling `raise`[^2] may be more suitable.

[^1]: https://doc.qt.io/qt-5/qt.html#WindowState-enum
[^2]: https://doc.qt.io/qt-5/qwidget.html#raise

lassoan: The topmost attribute is removed when the user clicks on the widget, so at that point the widget is already at the top. After that we don't want to force the widget to the top. So, I'm not sure why would we need to raise or activate.

jcfr: This was instead of the call to `show()` below which seems to bring back the widget to the top

cpinter:
This is great, thank you!

Base/QTApp/qSlicerApplicationHelper.txx
lassoan: To allow click-and-dragging to work, we need to keep the window displayed at the top. The flag is changed so that when the user stopped moving the window and goes to another application the window will not stay on top anymore. The user click-and-dragging the window is a hint that the window bothers the user.

jcfr: Agreed, unsetting `WindowStaysOnTopHint` is key.

To clarify, I was initially thinking that instead of calling `show()`, setting the window state to active was more appropriate. A quick experiment demonstrated this was not the case.

jcfr:
The splashscreen can not be moved without first releasing the mouse button and clicking again. This is being fixed in:
* https://github.com/Slicer/Slicer/pull/7532

PR #7528, opened 2024-01-10
COMP: Support using Slicer_INSTALL_* vars in external project exporting config

jcfr:
Workaround CMake issue 23312 by ensuring `Slicer_INSTALL_*` variables do no starts with "./". This ensures that on Linux and Windows, external projects in SuperBuild extensions setting variable `CMAKE_INSTALL_LIBDIR` like with value of `Slicer_INSTALL_THIRDPARTY_LIB_DIR` can generate a valid target file with the variable `_IMPORT_PREFIX` properly computed.


jcfr:
After further consideration, I will instead add similar workaround in extension requiring it.

PR #7532, opened 2024-01-10
ENH: Support dragging of the splashwindow just after clicking on it

jcfr:
Follow-up of edfbdb492 (ENH: Allow dragging of splashwindow), this commit further improves the `DraggableWidgetEventFilter` to support dragging the Splashwindow without having to release the mouse button and re-pressing it.

Since just after the mouse button is pressed and before it is released, mouse events are associated with a "QWidgetWindow" instance named after "<objectname>Window" if the widget object name is set or after "<classname>ClassWindow" if the object name is empty (See `QWidgetWindow::updateObjectName`[^1]).

To ensure events from either are being filtered, the event filter should be installed at the application level.

Change locally tested on Linux and macOS

[^1]: https://github.com/qt/qtbase/blob/v5.15.2/src/widgets/kernel/qwidgetwindow.cpp#L1151-L1158

### DIFF
--- a/Base/QTApp/qSlicerApplicationHelper.txx
+++ b/Base/QTApp/qSlicerApplicationHelper.txx
@@ -66,31 +66,9 @@ public:
 protected:
   bool eventFilter(QObject* obj, QEvent* event) override
     {
-    // Determine if the event is from the designated widget or its associated window.
-    bool eventFromWidget = obj == this->Widget;
-
-    // Check if events are from a QWidgetWindow
-    QString windowWidgetName = this->Widget->objectName();
-    if (windowWidgetName.isEmpty())
-    {
-      windowWidgetName = QString::fromUtf8(this->Widget->metaObject()->className()) + QLatin1String("Class");
-    }
-    windowWidgetName += QLatin1String("Window");
-    bool eventFromWidgetWindow = obj->objectName() == windowWidgetName;
-
-    // Ensure events from either the widget or its associated window are being filtered.
-    bool filterEvents = eventFromWidget || eventFromWidgetWindow;
-
-    // Rationale for checking if events are from QWidgetWindow:
-    // Since just after the mouse button is pressed and before it is released, mouse
-    // events are associated with a "QWidgetWindow" instance named after "<objectname>Window"
-    // if the widget object name is set or after "<classname>ClassWindow" if the object name is empty
-    // (See QWidgetWindow::updateObjectName).
-    // To ensure events from either are being filtered, the event filter should be installed at the
-    // application level.
-
-    if (filterEvents && event->type() == QEvent::MouseButtonPress && this->Widget)
+    if (event->type() == QEvent::MouseButtonPress && this->Widget)
       {
+
       // Record the mouse press position for later reference during dragging.
       QMouseEvent* mouseEvent = static_cast<QMouseEvent*>(event);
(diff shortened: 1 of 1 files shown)


pieper (commented):
Since I haven't built this I'll leave it to Andras to review.

jcfr:
@lassoan  If you do not have any issue, I plan on integrating later this afternoon.

lassoan:
I don't understand why the widget type check is needed. If we want to make sure that the message is for a the splash window then we could store the widget pointer in the event filter in a member variable.

jcfr:
>  If we want to make sure that the message is for a the splash window then we could store the widget pointer in the event filter in a member variable

It turns out that internally, the first event does not come from the actual splashscreen but from an `QWidgetWindow` internally created. So storing the pointer wouldn't help here.

jcfr:
This likely happen here because (we have to) break the regular event propagation to deal with the special case of toggling the `WindowStaysOnTopHint` off.

This post talk about `QWidgetWindow`, see https://stackoverflow.com/questions/47712264/qwidgetwindow-as-qobject-in-event-filter-with-qt5

lassoan:
Maybe it is cleaner to remove the topmost hint  in a timer callback.

lassoan:
@jcfr I've added a commit to this branch that uses a timer for changing the window hint. This way we don't need to add an event filter on the application and use a complex (and potentially fragile) logic for analyzing the source of events.

jcfr:
This is indeed much simpler and more robust. I will test late in the morning.

jcfr:
It turns out that the "original" issue preventing from dragging just after clicking is still present.

Despite of using the timer, being able to drag the splashscreen require to click, release, click again.

Is it what you observed ? 

Applying this patch may help give you more time to try ... 

```diff
diff --git a/Modules/Scripted/SampleData/SampleData.py b/Modules/Scripted/SampleData/SampleData.py
index 9d18ed7798..cd442c1e9c 100644
--- a/Modules/Scripted/SampleData/SampleData.py
+++ b/Modules/Scripted/SampleData/SampleData.py
@@ -73,6 +73,14 @@ class SampleData(ScriptedLoadableModule):
 
     def __init__(self, parent):
         ScriptedLoadableModule.__init__(self, parent)
+        import time
+
+        for i in range(10):
+          print("Wait %s / 10 second ..." % i)
+          for j in range(100):
+            time.sleep(0.01)
+            slicer.app.processEvents()
+
         self.parent.title = _("Sample Data")
         self.parent.categories = [translate("qSlicerAbstractCoreModule", "Informatics")]
         self.parent.dependencies = []
```

lassoan:
There are no issues on Windows, but the behavior can be very easily operating-system-dependent. A simple fix could be to start the timer in the button release event handler.

lassoan:
See my latest changes, they further simplify the code (no timer is needed) and works well for me on Windows. It would be great if you could check on linux/macos.

lassoan:
@jcfr This update ended up being quite simple and seems to fix the issue of having to click again to drag after the window hint changed, so if you don't have any objections I would squash&merge this today.

lassoan (approved):
This update ended up being quite simple and seems to fix the issue of having to click again to drag after the window hint changed. @jcfr if you don't have any objections I would squash&merge this today.

lassoan:
Thank you!

Issue #7290, opened 2023-10-18
Can load removed data from database


xriobe:
## Summary

Load button from DICOM database browser can allow to load data previously removed from the database, while it should not.

## Steps to reproduce

- Go to the DICOM database browser
- Select a volume and load it
- Remove the volume from the tree and the database
- Click on Examine/Load (depending on your Advanced checkbox)
=> The volume is examined/loaded

The loading is not even necessary:
- Go to the DICOM database browser
- Select a volume
- Remove the volume from the database
- Click on Examine
=> The volume is examined

## Expected behavior

The volume should not be examined/loaded, and probably the buttons should be disabled

## Environment
- Slicer version: Slicer-5.2.2
- Operating system: Linux Ubuntu 22.04

## Note

In DICOMBrowser.py, self.fileLists should be reset when the selection is removed.

pieper:
Thanks for reporting this 👍 

It sounds like you have a good idea what needs to be fixed - will you provide a PR?

PR #7535, opened 2024-01-11
BUG: Can load removed data from database

cpoetteCertis:
Load button from show DICOM browser can load a deleted data. To solve the bug, remove signals have been created in CTK to be able to connect in slicer. See CTK pull request:

https://github.com/commontk/CTK/pull/1164

This issue is described here: https://github.com/Slicer/Slicer/issues/7290

### DIFF
--- a/Modules/Scripted/DICOMLib/DICOMBrowser.py
+++ b/Modules/Scripted/DICOMLib/DICOMBrowser.py
@@ -316,6 +316,7 @@ class SlicerDICOMBrowser(VTKObservationMixin, qt.QWidget):
         self.loadableTable.setLoadables([])
         self.fileLists = self.getFileListsForRole(seriesUIDList, "SeriesUIDList")
         self.updateButtonStates()
+        self.fileLists = []
 
     def getFileListsForRole(self, uidArgument, role):
         fileLists = []


lassoan (commented):
I'm not sure that the fix is at the best abstraction level. We should not really care about DICOM database changes (it is hard to detect them robustly anyway). Instead, what we want is to empty the filelists when series are no longer selected in the series list in the DICOM browser. Detecting changes in series list selection in the current browser is a much simpler task, so I think we do that instead.

`SlicerDICOMBrowser already` connects to the `seriesSelectionChanged` signal, so there may  be a bug in CTK:  the `seriesSelectionChanged` signal is not emitted when selection is removed in the series table. Maybe something like this: https://github.com/lassoan/CTK/commit/25bbff183290bfbf49c847ff44144517178e4189

cpoetteCertis:
Indeed it is a much simpler thanks to your fix ! I just updated the PR

lassoan:
It seems that Slicer does not need to be changed (only the bug in CTK has to be fixed), so I'm closing this pull request.

There is a big rework of the DICOM browser in the works (148 modified files) with the [pull request](https://github.com/commontk/CTK/pull/1165) is being integrated in these days. I would recommend to wait for this pull request to be merged and then review if the CTK fix is still needed.

jcfr:
Closing. Superseded by the following CTK update:
* https://github.com/Slicer/Slicer/pull/7549

PR #7537, opened 2024-01-12
COMP: Ensure full path ID for macOS libraries installed for fix-up

jcfr:
This commit resolves an issue in Slicer's fix-up process, specifically when updating the installed executable and plugins using `@rpath`. To align with this process, both the installed libraries and their dependent libraries must have a full path. Consequently, we consistently clear `CMAKE_INSTALL_NAME_TOOL` in all external projects. This adjustment is crucial due to the lack of an explicit option in CMake to skip the "install_name_tool" call unless the variable is cleared. Please refer to `cmInstallTargetGenerator::AddInstallNamePatchRule` C++ function found in the CMake source code for a comprehensive understanding.

For historical context, the forceful setting of `CMAKE_INSTALL_NAME_TOOL` to an empty value was originally introduced in commit 4c441a280c (`COMP: Fix MacOSX packaging fixup associated with loadable module`, 2012-10-25). Subsequently, the comment "XXX Probably not needed anymore" was added in commit d2761656de (`BUG: Extension packaging: Ensure @rpath is not applied at build time`, 2016-06-10) and is now removed, as forcing the value is indispensable to support packaging of external projects like VTK modules built externally.


jcfr:
Changes have been locally tested. Once integrated the macOS build of Slicer packages for both the core and extensions will be monitored.

PR #7538, opened 2024-01-12
COMP: Fix macOS extension packaging updating VTK9 python modules fixup rules

jcfr:
This updates Slicer extension fix-up rules adapting a similar fix done for Slicer core in 0b1560371 (`COMP: Fix macOS package setting "path" in VTK9 python modules fixup rules`)


jcfr:
This changes follow-up https://github.com/Slicer/Slicer/pull/7537

Changes have been locally tested :white_check_mark: 

Once integrated, the macOS build for the Slicer packages of both the core and the extensions will be monitored.

PR #7541, opened 2024-01-12
BUG: Update data probe module to avoid resizing issues

hannah-morilak:
This PR updates the Data Probe module to stop hiding/showing elements to avoid awkward resizing.

ZoomySlicerad

This is the look after this PR:

ZoomySlicer

cc: @jamesobutler

### DIFF
--- a/Modules/Scripted/DataProbe/DataProbe.py
+++ b/Modules/Scripted/DataProbe/DataProbe.py
@@ -216,17 +216,11 @@ class DataProbeInfoWidget:
                 self.layerNames[layer].setText("")
                 self.layerIJKs[layer].setText("")
                 self.layerValues[layer].setText("")
-            self.imageLabel.hide()
-            self.viewerColor.hide()
-            self.viewInfo.hide()
-            self.viewerFrame.hide()
-            self.showImageFrame.show()
+                self.displayableManagerInfo.setText("S")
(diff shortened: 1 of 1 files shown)


lassoan:
I agree that the changing window size is not ideal. However, preallocating a larger window sie is not ideal either. Especially because it is almost impossible to know what is the maximum size required. Printing the name of all segments visible at the current mouse position may require several lines of text. How many would you allocate?

image

Does the changing window size cause any specific problems for you? Can you think of other solutions that preallocating a quite large space for the widget (which could go unusued most of the time)?

Most of the time we only have a background volume and allocating 2 empty lines for foreground and label is already a pretty bad waste of valuable vertical space in the module panel. We should try to use the space more efficiently rather than making the situation even worse by allocating more blank space.

lassoan:
Maybe we could revisit showing annotations in the slice view. The data probe was added because rendering of text slowed down view interactions too much. But maybe text rendering speed and generally computer speed has been improved enough over the years that it could be feasible to show all these information in corner annotations. Could you experiment with this? The DataProbe module already manages corner annotations (SliceViewAnnotations.py), so you would just need to tweak the module a bit to change what information is included in corner annotations.

pieper:
I like the idea of moving the segment names into the slice view somewhere, perhaps as a tooltip.

PR #7557, opened 2024-01-24
update the list of supported formats

fedorov:
follow up on https://github.com/QIICR/QuantitativeReporting/pull/274#issuecomment-1897812838


lassoan (approved):
Perfect, thank you for the update!

PR #7562, opened 2024-01-29
ENH: Add interaction handle visualization for linear transform nodes

Sunderlandkyl:
This pull request adds a visualization option for displaying/interacting with linear transforms in both 2D and 3D views using separate rotation/translation/scaling handles. This method of interaction replaces the existing box widget that is used currently to move linear transforms in 3D views only. Users now also have the option to create/toggle display of a parent transform interaction widget through the subject hierarchy by right-clicking on the display column of any transformable node.

## Demo installer
https://1drv.ms/u/s!Al0lwIPqdM2dgql4-5NkWigrmjeU8A?e=lIHd2H

## Existing widgets
| Transforms| Markups |
|----------|:-------------:|
| image | image |

# New widget

| Transforms| Markups |
|----------|:-------------:|
| image | image |

This new widget is based on and expands the functionality of the previous Markups interaction pipeline. The base class vtkMRMLInteractionWidget/vtkMRMLInteractionWidgetRepresentation implements most functionality, and can be overridden to display specialized use cases. Interaction widgets for both Markups and Transforms are derived from these classes.

<video
  controls muted
  src="https://github.com/NA-MIC/ProjectWeek/assets/9222709/43e2d906-a8c0-4909-b357-757d41457d7a"
  style="max-height:640px; min-height: 200px">
</video>


lassoan:
This looks awesome!

Why have you set it to draft? Is there anything that must be down before it can be integrated?

CI build reports a failure, needs to be checked if it is a real error.

Sunderlandkyl:
I created it as a draft so that it couldn't be integrated until we had gathered some feedback.

Looks like the build error a real error.
````
/usr/src/Slicer/Libs/MRML/Core/vtkITKTransformConverter.h:1315:23: error: redeclaration of ‘static vtkAbstractTransform* vtkITKTransformConverter::CreateVTKTransformFromITK(vtkObject*, typename slicer_itk::TransformBaseTemplate<TParametersValueType>::Pointer, double*)’ may not have default arguments [-fpermissive]
````

**Comment by jcfr** [comment]
> This looks awesome!

Ditto. It looks great :100: Thanks for working on this :pray: 

> build error

I just started a local build

**Comment by allemangD** [comment]
I never encountered a build issue, I assume as a result of 731287d. 

The usability of this widget is a great improvement! Some comments in no particular order:

---

There doesn't seem to be a way to perform uniform scaling. Offhand I don't recall if the old widget allowed uniform scaling either. Perhaps some modifier key on the white center dot could do this?

---

There can be a lot of visual noise hiding the subject - perhaps we could reduce the thickness of the rings, or use a line renderer with a fixed pixel width. See:

image

---

While playing with it I often _intuitively_ feel that I should be able to still access a particular rotation, but it's hidden away. Perhaps the fade-out angle could be tightened a bit? For example in this screenshot it _feels_ like I should be able to access the red rotation. The blue rotation is _accessible_ but it's faded to the point of near-invisibility.

image

That might hurt usability with my second point though, adding extra visual noise. Perhaps the "back" side of the rings could be hidden away, so only the "front" of the ball is drawn? See the similar widget in Blender, for example:

image

The arc for each axis is always shown, but only 1/2 of each arc. It also uses a fixed-width line rendering rather than planar geometry which seems to help with visual noise.

---

Currently the white dot allows moving within the view plane - it would be nice if there were also a way to move _perpendicular_ to the view plane, along line-of-sight. My first idea is to use a modifier key with the white dot, but combined with the point about uniform scaling I worry about overloading the white dot with too many actions. Probably better to leave that unless someone else has a more intuitive UX suggestion.

**Comment by allemangD** [comment]
Since the transforms widget is being unified with the markups widget, it would be good to allow disabling certain degrees of freedom like the markups control does. 

image


Additionally there should now be a "View plane" option for rotation, for the new white ring.

**Comment by Sunderlandkyl** [comment]
Thanks for the feedback David!

> There doesn't seem to be a way to perform uniform scaling. Offhand I don't recall if the old widget allowed uniform scaling either. Perhaps some modifier key on the white center dot could do this?

Did some testing, it looks like you can right-click on the box to apply uniform scaling by moving the mouse up and down. A modifier key would probably make sense in this case.

Another option for addressing visual noise that we have discussed is to only show a simplified version of the widget (something like a single ring), unless the mouse is within the range of the widget.

I'll generate some options based on the suggestions and post images of the results.

> While playing with it I often intuitively feel that I should be able to still access a particular rotation, but it's hidden away. Perhaps the fade-out angle could be tightened a bit? For example in this screenshot it feels like I should be able to access the red rotation. The blue rotation is accessible but it's faded to the point of near-invisibility.

In the current version, handles start fading away the angle between the view direction and the axis vector passes 30deg and is completely faded (+non interactive) after 20deg. I'll try lowering the threshold and reducing the range between those values.

> Since the transforms widget is being unified with the markups widget, it would be good to allow disabling certain degrees of freedom like the markups control does.

> Additionally there should now be a "View plane" option for rotation, for the new white ring.

Good point, those should be easy to add.

**Comment by Thibault-Pelletier** [comment]
Hi!

It looks very nice! Definitly more user friendly than the current one!

If I can give my personnal feedbacks :

I find the 3D interaction in the 2D slice not user friendly. It is the same problem for the old interactor and the interaction becomes difficult to understand for users when the views are not aligned with the current orientation of the transform / markup.

With the current widget, what I do is moving the MPR views to be aligned with the current widget on mouse release which makes it easier for users to know how they are manipulating the transform / markup (particularly useful when used in surgery planning apps with moving models with respect to volumes).

Also in 2D what we I often do is prefering using the MPR interaction widget which gives finer precision when moving transforms from the MPR views.

Having an option to have the MPR views follow the transform and / or using the slice interactions widget to move the transforms in MPR views would be nice IMO.

Other, small, feedback : I’m personnaly not a fan of the white interaction circle (feels a little clutered). So it would be nice to have an option to disable it :) . 

If you're going for a blender style interaction, snapped translations, rotations, and scaling would also be nice (but probably way overkill for most users).  

Thx again for the great work!
Best,
Thibault

**Comment by Sunderlandkyl** [comment]
New version with some changes:
https://1drv.ms/u/s!Al0lwIPqdM2dgql7p_gc9DsbDt4New?e=crjT6A

- Reduced the thickness/radius of the handles.
- Handles now begin to fade at 10deg (previous 30), and are invisible at 8def (previous 20).
- View plane rotation is disabled by default
- Transforms module now has a widget to toggle visibility for individual axes.
- Initial center of rotation is now initialized to the bounds center

**Comment by Sunderlandkyl** [comment]
> Other, small, feedback : I’m personnaly not a fan of the white interaction circle (feels a little clutered). So it would be nice to have an option to disable it :) .

I agree that while it is useful, it is probably too much information to display by default. In the latest version it is disabled by default, but can be enabled in the transforms module.

> If you're going for a blender style interaction, snapped translations, rotations, and scaling would also be nice (but probably way overkill for most users).

It's definitely planned for the future, through probably not for the initial version.

> Having an option to have the MPR views follow the transform and / or using the slice interactions widget to move the transforms in MPR views would be nice IMO.

I think this could be done by attaching the volume reslice driver to a transform node.

> The rings have 100% opacity (or maybe they are just darker colors)

The widgets in Slicer are 100% opacity, but use a desaturated color until the user moves the mouse over them.

> I did not find out the modifier key to allow changing the center of rotation of a transform. Maybe that information should be on the project page?

Thanks for pointing that out -- added to the project page.

**Comment by jcfr** [comment]
To confirm there will be no regression in term of build, branch is being build on the "official" build-machine used to build the package:
* macOS (`computron`) :hourglass_flowing_sand: 
* Linux (`metroplex`) :hourglass_flowing_sand:

**Comment by jcfr** [comment]
On macOS, the following was reported:

```
/D/P/S-0/Libs/MRML/DisplayableManager/vtkMRMLInteractionWidgetRepresentation.cxx:308:34: error: non-constant-expression cannot be narrowed from type 'int' to 'double' in initializer list [-Wc++11-narrowi
ng]
  double displayPosition4[4] = { displayPosition[0], displayPosition[1], 0.0, 1.0 };
                                 ^~~~~~~~~~~~~~~~~~
/D/P/S-0/Libs/MRML/DisplayableManager/vtkMRMLInteractionWidgetRepresentation.cxx:308:34: note: insert an explicit cast to silence this issue
  double displayPosition4[4] = { displayPosition[0], displayPosition[1], 0.0, 1.0 };
                                 ^~~~~~~~~~~~~~~~~~
                                 static_cast<double>( )
/D/P/S-0/Libs/MRML/DisplayableManager/vtkMRMLInteractionWidgetRepresentation.cxx:308:54: error: non-constant-expression cannot be narrowed from type 'int' to 'double' in initializer list [-Wc++11-narrowi
ng]
  double displayPosition4[4] = { displayPosition[0], displayPosition[1], 0.0, 1.0 };
                                                     ^~~~~~~~~~~~~~~~~~
/D/P/S-0/Libs/MRML/DisplayableManager/vtkMRMLInteractionWidgetRepresentation.cxx:308:54: note: insert an explicit cast to silence this issue
  double displayPosition4[4] = { displayPosition[0], displayPosition[1], 0.0, 1.0 };
                                                     ^~~~~~~~~~~~~~~~~~
                                                     static_cast<double>( )

```

**Comment by jcfr** [comment]
@Sunderlandkyl While the plan is to `Squash and Merge`, I purposefully integrated following up changes as independent commit to more easily discuss some of the warning fixes

File: Modules/Loadable/Transforms/MRMLDM/vtkMRMLLinearTransformsDisplayableManager2D.cxx [hunk 1]
```diff
@@ -233,7 +233,7 @@ void vtkMRMLLinearTransformsDisplayableManager2D::vtkInternal::RemoveTransformNo
 }
 
 //---------------------------------------------------------------------------
-void vtkMRMLLinearTransformsDisplayableManager2D::vtkInternal::UpdateDisplayableTransforms(vtkMRMLTransformNode* mNode)
+void vtkMRMLLinearTransformsDisplayableManager2D::vtkInternal::UpdateDisplayableTransforms(vtkMRMLTransformNode* vtkNotUsed(mNode))
```
jcfr, line 236: @Sunderlandkyl Should the implementation of this be similar to the one of `vtkMRMLLinearTransformsDisplayableManager3D` ?

See https://github.com/Slicer/Slicer/blob/main/Modules/Loadable/Transforms/MRMLDM/vtkMRMLLinearTransformsDisplayableManager3D.cxx#L260-L275

jcfr (approved):

PR #7566, opened 2024-01-29
ENH: Add DICOM support for Multi-frame Grayscale Byte Secondary Capture Image

Sunderlandkyl:
Adds Multi-frame Grayscale Byte Secondary Capture Image Storage (SOP Class UID: 1.2.840.10008.5.1.4.1.1.7.2) to DICOMImageSequencePlugin.py.


lassoan (approved):

PR #7568, opened 2024-02-02
ENH: Update https support and add WebXR demo as transform controller

pieper:
HTTPS support updated to support both the .pem file and .key file. This allows the server to be configured to support https if the .pem file is shared with a client (note - for Android phones the .pem file may need to be converted to .der format).  Adding a way to help generate and configure these files without editing the code can be addressed in future work.

HTTPS is required to use the WebXR interface on the mobile phone due to security concerns (WebXR uses cameras and other tracking data that is considered a potential privacy issue, so https is required).

The WebXR demo allows the user to manage a MRML linear transform node based on the position and orientation of the phone.  See this page for an example video and more information:

https://projectweek.na-mic.org/PW40_2024_GranCanaria/Projects/AugmentedRealityExperimentsForCardiologyImaging/


lassoan (approved):

PR #7577, opened 2024-02-09
BUG: Fix segment disappearing when sharing segment editor node between widgets

### LINKED ISSUE #6874: Segment disappears after threshold effect preview; two widgets
## Summary

Trying to apply threshold to a segment results in the segment disappearing.  The data is correct as shown in DataProbe and save/reload but it is not visible in slice views.

## Steps to reproduce

1) Install MONAI Label and connect to a server.
2) Run inference
3) Open SegmentEditor and select threshold
4) Either apply or cancel and selected segment is gone.

## Expected behavior

Segment Editor effects should work as normal on segmentations coming from MONAI Label

## Environment
- Operating system: Linux

## More Information

After some debugging I believe the issue is that the MONAI Label has a segment editor widget that shares state via the segment editor node with the Segment  Editor module's widget such that when you enter the Threshold effect in either widget are two instances of the threshold effect class created.  The first instance correctly stores the opacity states and sets the display node to 0, but the second instance records the zero state.  Then on cancel the original non-zero opacity states are not restored.

This would appear to be an issue the threshold any time there would be more than one segment editor widget.  I think the solution would be to store the pre-preview opacity state (`previewedSegmentID, segment2DFillOpacity,segment2DOutlineOpacity`) in the segment editor node and not in the threshold effect instance so that only one of the instances would save/restore the opacity.
        
https://github.com/Slicer/Slicer/blob/5dfb6fb6945d3fb56320c2920571f3ffcc5b48fb/Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorThresholdEffect.py#L99-L145

lassoan:
When a single vtkMRMLSegmentEditorNode was used by multiple qSlicerSegmentEditorWidget instances then after using Threshold effect, the current segment disappeared from slice views.

The reason was that Threshold effects in every segment editor widgets get activated (via sharing the segment editor node) and they all started preview, saved original segment opacity, and then on "Apply" restored the original segment opacity. However, the segment opacity that was restored by second, third, etc. effects were not the original anymore, but it was already the opacity that was set to 0 (it has to be set to 0 to prevent the current segment occluding the preview).

The solution could have been to store the original opacity in a shared location (e.g., in the segment editor node), but that would not have solved the slight rendering issue (preview glow was darker) and performance degradation caused by several effects showing the preview glow at the same time.

Fixed it by storing the object ID of the threshold effect that manipulates the segmentation display node in the node's "SegmentEditor.PreviewingEffect" attribute. All other Threshold effects check this attribute and if they find that another effect already provide preview then they don't display the preview.

This commit also fixes disappearing segment when selecting another segmentation node while the Threshold effect is active.

fixes #6874

### DIFF
--- a/Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorEffects/SegmentEditorThresholdEffect.py
+++ b/Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorEffects/SegmentEditorThresholdEffect.py
@@ -27,6 +27,7 @@ class SegmentEditorThresholdEffect(AbstractScriptedSegmentEditorEffect):
 
         self.segment2DFillOpacity = None
         self.segment2DOutlineOpacity = None
+        self.previewedSegmentationDisplayNode = None
         self.previewedSegmentID = None
 
         # Effect-specific members
@@ -42,7 +43,6 @@ class SegmentEditorThresholdEffect(AbstractScriptedSegmentEditorEffect):
 
         self.previewPipelines = {}
         self.histogramPipeline = None
-        self.setupPreviewDisplay()
 
         # Histogram stencil setup
         self.stencil = vtk.vtkPolyDataToImageStencil()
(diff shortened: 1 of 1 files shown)


pieper (approved):
LGTM 👍

pieper:
Thanks for the fix!

Issue #7512, opened 2023-12-26
"Cannot create accessible child interface for object" warning


jcfr:
## Summary

Warnings like the following are reported when clicking on row associated with the "MRML Node information" attribute table.

```
Cannot create accessible child interface for object:  QTableWidget(0x1f789e73310, name = "NodeAttributesTable")  index:  7
```

This was observed on Windows 11 Pro using:
* Slicer installed using stable installers for `5.2.2` and `5.6.1`
* Slicer built from source (https://github.com/Slicer/Slicer/commit/ecc1f684f30155818f4dcf482b5ebbb0b42661d5 ) using Visual Studio 2022

## Steps to reproduce

For convenience, the snipper below allows to reproduce the issue.

Note that the same warnings are reported when explicitly selecting the row using the mouse.

```python
from SampleData import SampleDataLogic
node = SampleDataLogic().downloadMRHead()
node.SetAttribute("foo", "foo")
node.SetAttribute("bar", "bar")
slicer.util.selectModule("Data")

groupBox = slicer.util.findChildren(name="MRMLNodeInspectorGroupBox")[0]
groupBox.collapsed = False
slicer.app.processEvents()

shTreeView = slicer.util.findChildren(name="SubjectHierarchyTreeView")[0]
shTreeView.setCurrentNode(node)

attributeTableView = slicer.util.findChildren(name="MRMLNodeAttributeTableView")[0]

# Since the wrapped API doesn't allow to directly update the selection, we workaround
# this by getting a reference to the underlying table widget
internalTableWidget= slicer.util.findChildren(attributeTableView, className="QTableWidget")[0]

internalTableWidget.selectRow(0)
slicer.app.processEvents()

internalTableWidget.selectRow(1)
slicer.app.processEvents()
```

## Related
* [QTBUG-33247](https://bugreports.qt.io/browse/QTBUG-33247): `Cannot creat accessible child interface for object` and associated original fix reviewed through [65171](https://codereview.qt-project.org/c/qt/qtbase/+/65171) and integrated as https://github.com/qt/qtbase/commit/e39d629ebe9044b505ac35eaae3ab9c214d452a2

## Environment
- Slicer version: https://github.com/Slicer/Slicer/commit/ecc1f684f30155818f4dcf482b5ebbb0b42661d5
- Operating system: Windows 11 Pro

jcfr:
As discussed in this thread[^1], root cause is probably related to improper handling of indices. 

[...]

PR #7578, opened 2024-02-09
ENH: Improve appearance of node information section in Data module

lassoan:
In Data module, "MRML node information" section had a "Node information" section and an untitled node attributes section. The node attributes section displayed messages (e.g., attribute name already exists) above the table. This all looked quite confusing.

Updated the GUI to have a "Properties" and "Attributes" section under "MRML node information". Also replaced the label above the table by a tooltip for showing "attribute name already exists" message.

image

### DIFF
--- a/Libs/MRML/Widgets/qMRMLNodeAttributeTableView.cxx
+++ b/Libs/MRML/Widgets/qMRMLNodeAttributeTableView.cxx
@@ -191,14 +191,13 @@ void qMRMLNodeAttributeTableView::onAttributeChanged(QTableWidgetItem* changedIt
       changedItem->setText(attributeNameBeforeEditing);
       d->NodeAttributesTable->blockSignals(wasBlocked);
 
-      // Don't set if there is another attribute with the same name (would overwrite it),
-      // revert to the original value and display a message.
+      // Ensure the attribute name is unique; if not, avoid overwriting it by reverting
+      // to the original value and notify the user.
       d->MessageText = tr("There is already an attribute with the same name");
       QRect rect = d->NodeAttributesTable->visualItemRect(changedItem);
       d->MessagePosition = d->NodeAttributesTable->mapToGlobal(rect.bottomLeft());
-      // If we directly show the tooltip in this callback function now then the tooltip
-      // does not appear, therefore we trigger a timer to display it once the attribute change
-      // is fully completed.
+      // Due to limitations in displaying the tooltip immediately within this callback,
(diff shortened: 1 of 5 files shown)


Libs/MRML/Widgets/qMRMLNodeAttributeTableView.cxx
```diff
@@ -193,12 +187,19 @@ void qMRMLNodeAttributeTableView::onAttributeChanged(QTableWidgetItem* changedIt
...
       {
-      // Don't set if there is another attribute with the same name (would overwrite it),
-      // revert to the original value.
-      d->setMessage(tr("There is already an attribute with the same name"));
       bool wasBlocked = d->NodeAttributesTable->blockSignals(true);
       changedItem->setText(attributeNameBeforeEditing);
       d->NodeAttributesTable->blockSignals(wasBlocked);
+
+      // Don't set if there is another attribute with the same name (would overwrite it),
+      // revert to the original value and display a message.
```
jcfr, line 195:
```suggestion
      // Ensure the attribute name is unique; if not, avoid overwriting it by reverting
      // to the original value and notify the user.
```

Libs/MRML/Widgets/qMRMLNodeAttributeTableView.cxx
```diff
@@ -193,12 +187,19 @@ void qMRMLNodeAttributeTableView::onAttributeChanged(QTableWidgetItem* changedIt
...
+      d->MessageText = tr("There is already an attribute with the same name");
+      QRect rect = d->NodeAttributesTable->visualItemRect(changedItem);
+      d->MessagePosition = d->NodeAttributesTable->mapToGlobal(rect.bottomLeft());
+      // If we directly show the tooltip in this callback function now then the tooltip
+      // does not appear, therefore we trigger a timer to display it once the attribute change
+      // is fully completed.
```
jcfr, line 201:
```suggestion
      // Due to limitations in displaying the tooltip immediately within this callback,
      // a timer is used to delay its appearance until after the attribute change is fully processed.
```

jcfr:
The changes makes sense and help clarify the user interface :100:

Does it help address issue https://github.com/Slicer/Slicer/issues/7512 ?

lassoan:
> Does it help address issue #7512 ?

The change only affects the message display (in a separate QLabel) and modifies the layout a bit, so it probably does not fix the "accessible child interface" message.

PR #7587, opened 2024-02-14
BUG: Ensure launcher scripts installed from wheels are accessible

jcfr:
This commit addresses an issue where launcher scripts installed from wheels were not correctly added to the PATH. Specifically, it updates the PATH for both "Slicer" and "PythonSlicer" launcher settings in the build and install tree. This adjustment ensures that convenience executables created after installing wheels with specified entry-points such as "project.scripts" or "project.gui-scripts" can be easily located and accessed within the project environment.

### DIFF
--- a/SuperBuild/External_python.cmake
+++ b/SuperBuild/External_python.cmake
@@ -278,6 +278,14 @@ if((NOT DEFINED PYTHON_INCLUDE_DIR
   mark_as_superbuild(PYTHON_VALGRIND_SUPPRESSIONS_FILE:FILEPATH)
 
   #-----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


jcfr:
Follow-up of:
* https://github.com/ImagingDataCommons/idc-index/pull/42

lassoan (approved):
If I understand correctly, this will add the Python scripts folder to the PATH, which will be very useful, as it makes easier to find scripts that various Python packages install. I cannot comment on the details of the implementation, but it looks good overall.

jcfr:
> If I understand correctly, this will add the Python scripts folder to the PATH,

That is correct. The Slicer python environment used from either the Python Console available in the application or through `PythonSlicer` will now have its PATH setup similarly to a regular python interpreter or python virtual env after it has been activated.

Issue #7570, opened 2024-02-02
Improve interaction handles


lassoan:
## Summary

The new interaction handles are awesome. Here are a few tweaks that would make them even better:

- "Interaction in 3D view" subject hierarchy context menu item was added because the interaction was only possible in 3D, but now that we have it in 2D views as well, the menu item should be just "Interaction" (and turn on/off overall visibility)
- In Transforms module / Display / Interaction: add "Visibility in 3D view" option
- In Transforms module if "Enable translation" is off then I cannot move (cannot even see) the center of rotation anymore. If translation is disabled but rotation is enabled, then left-click-drag should move the center of rotation (with or without pressing Alt)
- Left-click on the rotation center should jump slice views (same as when left-clicking on a markup control point) to make it easier to position the centerpoint in 3D
- If "Enable translation" checkbox is unchecked then disable corresponding x, y, z, in-plane checkboxes to make it more clear that those checkboxes depend on the main "enable" checkbox. Same for rotation and scaling.

## Environment
- Slicer version: Slicer-5.7.0 2024-02-02
- Operating system: Windows

Sunderlandkyl:
> In slice views, I find in-plane rotation very easy to understand, but rotation around parent coordinate system axes is hard to grasp. Therefore, in slice views, default only in-plane rotation handle should be displayed.

Does this apply to translation as well? Out of plane translation is often difficult to understand as well.

lassoan:
Good point, yes, it applies to the translation handles as well. I find the white dot in the middle much easier to use.

All these do not mean that translation and rotation around parent coordinate system axes is not needed, it is actually very important (e.g., for device placement, to translate and rotate around device axes), it is just a different use case.

Sunderlandkyl:
We currently use Red/Green/Blue for the axes colors since that is what is used in other 3D software, but @pieper brought up an interesting suggestion during the project week.

What if we used Yellow/Green/Red for the default axis colors to match the Coronal/Sagittal/Axial views?

Punzo:
@Sunderlandkyl I have a linear transform applied to a volume and if I activate the inetraction handles the center of transformation does not correspond to the center of the volume (I saw that the center can be changed also interactively, alt + left click). Would make sense to calculate the default center on the boundingbox of all the nodes for which the transform is applied?

![image](https://github.com/Slicer/Slicer/assets/7985338/9eb56a53-0d7e-4084-b17c-b6c55bb84aee)

Punzo:
[...]

PR #7590, opened 2024-02-16
ENH: Clean up vtkMRMLLinearTransformsDisplayableManager implementation

Sunderlandkyl:
Refactor and combine both 2D and 3D displayable managers for linear transform nodes into a single class.

Since both 2D and 3D displayable managers are now generating the exact same interaction widget, they ended up sharing all of the same code. To avoid duplication and to reduce maintenance, this commit combines the two displayable managers into a single class.

Re #7570


lassoan (approved):
This is excellent, thank you!

PR #7594, opened 2024-02-20
BUG: Fix unrelated matrix elements overwritten in transform editor

lassoan:
This is a fix for the transformation matrix editing glitch that @RafaelPalomar demonstrated during the weekly developer meeting.

In transforms module, when linear transformation matrix values were edited using spinboxes and then by directly editing the matrix, the value that was previously edited in the spinbox got overwritten by the new value.

The issue was that a single slot was used for all signals coming from all sliders and in the slot implementation the active focus was used to determine which axis should be updated. Fixed by using a separate slot for each axis. This way we don't rely on the GUI widget focus, which may go through complex transitions during transform editing using various widgets.


pieper (approved):
Thanks for the quick fix 👍

RafaelPalomar:
Thank you very much, @lassoan. Super quick!

jcfr (approved):

PR #7598, opened 2024-02-21
BUG: Fixed issue with adding new node after rename cancelation

### LINKED ISSUE #6002: Cancelling "Create new (node) as ..." dialog should not create new object
Cancelling "Create new _X_ as ..." dialogue creates new object with the default name (as if from "Create new _X_"), but clicking `Cancel` should not produce any object.  
I think this happens in several modules, with different types of object, including `Table` and `Volume`.  

## Steps to reproduce

Invoke "Create new Table as ..." dialogue, such as for the _Output table_ of the **Segment Statistics** module.
Click `Cancel`.
This creates a new Table object with a default name.  

## Expected behaviour

Should not create any new object.  

## Environment
- Slicer version: Slicer-4.11.20210226
- Operating system: Windows 10 x64

aramSofthenge:
Fixes #6002

Additional minor improvements

### DIFF
--- a/Libs/MRML/Widgets/qMRMLNodeComboBox.cxx
+++ b/Libs/MRML/Widgets/qMRMLNodeComboBox.cxx
@@ -43,6 +43,15 @@
 #include <vtkMRMLNode.h>
 #include <vtkMRMLScene.h>
 
+// ----------------------------------------------------------------------------
+
+namespace
+{
+  const QString createNew = QObject::tr("Create new ");
+  const QString asSuffix = QObject::tr(" as...");
+  const QString renameTitlePrefix = QObject::tr("Rename ");
+}
+
 // --------------------------------------------------------------------------
 qMRMLNodeComboBoxPrivate::qMRMLNodeComboBoxPrivate(qMRMLNodeComboBox& object)
   : q_ptr(&object)
@@ -308,10 +317,10 @@ void qMRMLNodeComboBoxPrivate::updateActionItems(bool resetRootIndex)
       foreach (QString nodeType, q->nodeTypes())
       {
         QString label = q->nodeTypeLabel(nodeType);
-        extraItems.append(qMRMLNodeComboBox::tr("Create new ") + label);
+        extraItems.append(createNew + label);
         if (this->RenameEnabled)
         {
-          extraItems.append(qMRMLNodeComboBox::tr("Create new ") + label + qMRMLNodeComboBox::tr(" as..."));
+          extraItems.append(createNew + label + asSuffix);
         }
       }
     }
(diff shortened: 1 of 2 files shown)


Libs/MRML/Widgets/qMRMLNodeComboBox.cxx
```diff
@@ -695,16 +708,44 @@ void qMRMLNodeComboBox::renameCurrentNode()
...
+            this->setCurrentNode(newNode);
+            emit this->nodeAddedByUser(newNode);
+        }
+    }
+}
+
+// --------------------------------------------------------------------------
+QString qMRMLNodeComboBox::getNameFromDialog(const QString& dialogTitle, const QString& defaultName = QString(), bool* ok = nullptr)
+{
+  QString newName = QInputDialog::getText(this, dialogTitle, "New name:", QLineEdit::Normal, defaultName, ok);
```
pieper, line 747: Shouldn't "New name" be marked for translation?

Libs/MRML/Widgets/qMRMLNodeComboBox.cxx
```diff
@@ -695,16 +708,44 @@ void qMRMLNodeComboBox::renameCurrentNode()
...
   }
-  node->SetName(newName.toUtf8());
-  emit currentNodeRenamed(newName);
+}
+// --------------------------------------------------------------------------
+void qMRMLNodeComboBox::createNodeAs(const QString& nodeTypeName)
+{
+    QString label = this->nodeTypeLabel(nodeTypeName);
+    bool okPressed = false;
+    QString dialogTitle = tr("Create New %1").arg(label);
```
pieper, line 728: Shouldn't this use the already translated `createNew` from above?  Or do you think having the argument might mean it translates to a different string in different languages.

Libs/MRML/Widgets/qMRMLNodeComboBox.cxx
```diff
@@ -43,6 +43,15 @@
 #include <vtkMRMLNode.h>
 #include <vtkMRMLScene.h>
 
+// ----------------------------------------------------------------------------
+
+namespace
+{
+  const QString createNew = QObject::tr("Create new ");
+  const QString asSuffix = QObject::tr(" as...");
+  const QString renameTitlePrefix = QObject::tr("Rename ");
```
pieper, line 52: I don't know enough about all the languages possibly involved here, but it seems to me some languages may not structure the sentence the same way and these may not be easy to break down as prefix and suffix.

Libs/MRML/Widgets/qMRMLNodeComboBox.cxx
```diff
@@ -695,16 +708,44 @@ void qMRMLNodeComboBox::renameCurrentNode()
...
+    if (okPressed && !nodeName.isEmpty())
+    {
+        vtkMRMLNode* newNode = this->addNode(nodeTypeName);
+        if (newNode != nullptr)
+        {
+            newNode->SetName(nodeName.toUtf8());
+            this->setCurrentNode(newNode);
+            emit this->nodeAddedByUser(newNode);
+        }
+    }
```
pieper, line 741: It seems like this is the key change to fix the reported issue.  Maybe it should be a dedicated commit and the i18n changes would be independent.

Or if the i18n changes at least make it better than it was before we could integrate these and then make further improvements in another PR.

pieper (changes requested):
Thanks for working on this 👍

Libs/MRML/Widgets/qMRMLNodeComboBox.cxx
aramSofthenge: If you think this is essencial, can you please explain the point?
Because I don't see any specific reason of doing this, this will take me extra effort to separate the commits. I also followed the contribution guidelines and did not notice such a requirement to separate different changes.

pieper (approved):

PR #7600, opened 2024-02-22
ENH: Add cancel action for interaction handles

Sunderlandkyl:
This commit adds a cancel action when translating/rotating/scaling using interaction handles. Users can press "Escape" or right-click to cancel the current action and restore the Transform or Markups node back to its original state.

Re #7570


lassoan (approved):
This looks good, thank you!

Issue #6941, opened 2023-04-16
Segmentation closed surface representation generation fails for float voxels


lassoan:
## Summary

When a seg.nrrd file uses `type: float` voxels then closed surface representation is not generated (nothing appears or only a single segment's 3D representation appears).

See full error report here: https://discourse.slicer.org/t/only-one-of-28-segments-is-showing-in-3d/28934/3

## Steps to reproduce

- Load https://github.com/lassoan/PublicTestingData/releases/download/data/InvalidFloatSegmentationArterialAtlas.seg.nrrd into Slicer
- Create closed surface representation
- Only one segment appears in 3D

## Expected behavior

All segments should appear in 3D

## Environment
- Slicer version: Slicer-5.2.2
- Operating system: Windows

lassoan:
Using float voxel type for segmentation does not make much sense (it uses more memory and there are potential issues with comparison of floating-point values), so we don't need to support this. However, we should then display a specific error message to the user when he attempts to load such image and when float/double image is attempted to be used programmatically then we should log an error.

pieper:
Perhaps if someone tries to load float as segmentation we should offer the option of automatically rounding it and casting to the best corresponding datatype based on the data range.  On the other hand I don't think this comes up very often and it's not worth investing much effort.

lassoan:
Good point, logging a warning and automatically casting would be probably a better solution (need to check the scalar range to choose the correct integer type and the image may need to be rejected if does not fit into any integer range).

lassoan:
The situation is even worse for ARM macOS systems, as the ARM processors make the application crash with some floating-point exception when working with a segmentation with float/double value.

@Sunderlandkyl do you think you could implement a fix (log a warning and convert the image into the most suitable integer type; return with an error if it is not possible to convert) in the next few weeks?

Sunderlandkyl:
Sure, I can do that.
Still not sure how those segmentations are being generated.

lassoan:
This is getting more and more common. See for example: https://discourse.slicer.org/t/segmentation-issue-invisible-labels/34460/2

It seems that problems come from that researchers use an AI model to compute the segmentation. The inference result is float, then they use thresholding or sigmoid function to binarize it. But then they forget to cast the result to an integer type in the end.

Sunderlandkyl:
I have a function that will convert any vtkImageData to the smallest possible data type that will include the entire scalar range, excluding scalar point representations.

[...]

PR #7605, opened 2024-02-26
BUG: Fix issues caused by invalid Segmentation binary labelmap scalar types

Sunderlandkyl:
Adds the functions `ValidateSegmentationImageType` and `CastToSmallestUnsignedIntegerType` to vtkOrientedImageDataResample.
- `IsImageScalarTypeValid`: Returns True if the scalar type of the image is valid for representing segmentations. False otherwise.
- `CastToSmallestUnsignedIntegerType`: Casts the contents of the image to the smallest unsigned integer type that can contain all values in the image.
- `ValidateSegmentationImageType`: Checks to see if the scalar type of the image is a valid type, and casts the image to the smallest valid type if it is not.

The function `ValidateSegmentationImageType` is used during binary labelmap read/import to automatically convert the labelmap.
If a labelmap representation with an invalid scalar type is added to a segment, we only log a warning and don't automatically convert the labelmap.
Before performing binary labelmap to closed surface conversion, we check that the scalar type is valid, and log/return an error if it is not.

This should prevent errors from arising due to the use of floating point images for segmentations.

Re #6941

### DIFF
--- a/Libs/MRML/Core/vtkMRMLSegmentationStorageNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLSegmentationStorageNode.cxx
@@ -682,7 +682,7 @@ int vtkMRMLSegmentationStorageNode::ReadBinaryLabelmapRepresentation(vtkMRMLSegm
   }
   else
   {
-    vtkDebugMacro("ReadBinaryLabelmapRepresentation: File is not using a supported format!");
+    vtkDebugMacro("ReadBinaryLabelmapRepresentation: File is not using a supported format");
     return 0;
   }
 
@@ -694,6 +694,33 @@ int vtkMRMLSegmentationStorageNode::ReadBinaryLabelmapRepresentation(vtkMRMLSegm
   }
 
   // Read succeeded
+  int ret = vtkOrientedImageDataResample::IsImageScalarTypeValid(imageData);
+  switch (ret)
+  {
+    case vtkOrientedImageDataResample::TYPE_CONVERSION_TRUNCATION_NEEDED:
+      vtkWarningToMessageCollectionMacro(this->GetUserMessages(), "vtkMRMLSegmentationStorageNode::ReadBinaryLabelmapRepresentation",
+        "Segmentation is a floating point scalar type and will be cast to an integer type by truncation (rounding towards 0).");
+      break;
+    case vtkOrientedImageDataResample::TYPE_CONVERSION_CLAMPING_NEEDED:
+       vtkWarningToMessageCollectionMacro(this->GetUserMessages(), "vtkMRMLSegmentationStorageNode::ReadBinaryLabelmapRepresentation",
+        "Segmentation is a floating point scalar type and are outside of the possible range of integer. Voxel values will be clamped to integer range.");
+      break;
+    case vtkOrientedImageDataResample::TYPE_ERROR:
+      vtkErrorToMessageCollectionMacro(this->GetUserMessages(), "vtkMRMLSegmentationStorageNode::ReadBinaryLabelmapRepresentation",
+        "Failed to cast image to integer type.");
+      return 0;
+    case vtkOrientedImageDataResample::TYPE_OK:
+    default:
+      break;
+  }
+  if (ret != vtkOrientedImageDataResample::TYPE_OK
+    && !vtkOrientedImageDataResample::CastSegmentationToSmallestIntegerType(imageData))
+  {
+    vtkErrorToMessageCollectionMacro(this->GetUserMessages(),
+      "vtkMRMLSegmentationStorageNode::ReadBinaryLabelmapRepresentation",
+      "Failed to cast image to integer type.");
+    return 0;
+  }
 
   int numberOfFrames = imageData->GetNumberOfScalarComponents();
 
--- a/Libs/vtkSegmentationCore/vtkOrientedImageDataResample.cxx
+++ b/Libs/vtkSegmentationCore/vtkOrientedImageDataResample.cxx
@@ -1835,9 +1835,130 @@ bool vtkOrientedImageDataResample::IsLabelInMask(
   return valueFound;
 }
 
+//----------------------------------------------------------------------------
+int vtkOrientedImageDataResample::IsImageScalarTypeValid(vtkImageData* image)
+{
+  if (!image)
+  {
+    vtkErrorWithObjectMacro(nullptr, "IsImageScalarTypeValid: Invalid input image");
+    return TYPE_ERROR;
+  }
+
+  int imageScalarType = image->GetScalarType();
+  switch (imageScalarType)
+  {
+    case VTK_UNSIGNED_CHAR:
+    case VTK_CHAR:
+    case VTK_UNSIGNED_SHORT:
+    case VTK_SHORT:
+    case VTK_UNSIGNED_INT:
+    case VTK_INT:
+      // Current type is valid
+      return TYPE_OK;
+    default:
+      break;
+  }
+
+  double scalarRange[2] = { 0.0, 0.0 };
+  image->GetScalarRange(scalarRange);
+  int smallestScalarType = vtkOrientedImageDataResample::GetSmallestIntegerTypeForSegmentationScalarRange(scalarRange);
+  if (imageScalarType == VTK_FLOAT || imageScalarType == VTK_DOUBLE)
+  {
+    // Floating point type will need to be converted to integer type.
+    // If no integer type can contain the full scalar range, clamping will be needed.
+    // Truncation will be needed if the values are not all integers.
+    return smallestScalarType >= 0 ? TYPE_CONVERSION_TRUNCATION_NEEDED : TYPE_CONVERSION_CLAMPING_NEEDED;
+  }
+
+  if (smallestScalarType < 0)
+  {
+    // No valid integer type can contain the scalar range. The result will need to be clamped to (unsigned) int.
+    return TYPE_CONVERSION_CLAMPING_NEEDED;
+  }
+
+  return TYPE_CONVERSION_NEEDED;
+}
 
 //----------------------------------------------------------------------------
-void vtkOrientedImageDataResample::CastImageForValue(vtkOrientedImageData* image, double value)
+int vtkOrientedImageDataResample::GetSmallestIntegerTypeForSegmentationScalarRange(double scalarRange[2])
+{
+  if (scalarRange[0] >= VTK_UNSIGNED_CHAR_MIN && scalarRange[1] <= VTK_UNSIGNED_CHAR_MAX)
+  {
+    return VTK_UNSIGNED_CHAR;
+  }
+  else if (scalarRange[0] >= VTK_CHAR_MIN && scalarRange[1] <= VTK_CHAR_MAX)
+  {
+    return VTK_CHAR;
+  }
+  else if (scalarRange[0] >= VTK_UNSIGNED_SHORT_MIN && scalarRange[1] <= VTK_UNSIGNED_SHORT_MAX)
+  {
+    return VTK_UNSIGNED_SHORT;
+  }
+  else if (scalarRange[0] >= VTK_SHORT_MIN && scalarRange[1] <= VTK_SHORT_MAX)
+  {
+    return VTK_SHORT;
+  }
+  else if (scalarRange[0] >= VTK_UNSIGNED_INT_MIN && scalarRange[1] <= VTK_UNSIGNED_INT_MAX)
+  {
+    return VTK_UNSIGNED_INT;
+  }
+  else if (scalarRange[0] >= VTK_INT_MIN && scalarRange[1] <= VTK_INT_MAX)
+  {
+    return VTK_INT;
+  }
+
+  // No valid integer type can contain the scalar range.
+  return -1;
+}
+
+//----------------------------------------------------------------------------
+bool vtkOrientedImageDataResample::CastSegmentationToSmallestIntegerType(vtkImageData* image)
+{
+  if (!image)
+  {
+    vtkErrorWithObjectMacro(nullptr, "CastImageToSmallestIntegerType: Invalid input image");
+    return false;
+  }
+
+  double scalarRange[2] = { 0.0, 0.0 };
+  image->GetScalarRange(scalarRange);
+  return vtkOrientedImageDataResample::CastSegmentationToSmallestIntegerType(image, scalarRange);
+}
+
+//----------------------------------------------------------------------------
+bool vtkOrientedImageDataResample::CastSegmentationToSmallestIntegerType(vtkImageData* image, double scalarRange[2])
+{
+  if (!image)
+  {
+    vtkErrorWithObjectMacro(nullptr, "CastImageToSmallestIntegerType: Invalid input image");
+    return false;
+  }
+
+  int oldScalarType = image->GetScalarType();
+  int newScalarType = vtkOrientedImageDataResample::GetSmallestIntegerTypeForSegmentationScalarRange(scalarRange);
+  if (newScalarType == oldScalarType)
+  {
+    // No change needed.
+    return true;
+  }
+
+  if (newScalarType < 0)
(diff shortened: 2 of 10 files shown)


Libs/vtkSegmentationCore/vtkOrientedImageDataResample.cxx
```diff
@@ -1835,9 +1835,100 @@ bool vtkOrientedImageDataResample::IsLabelInMask(
...
+  {
+    VTK_UNSIGNED_CHAR,
+    VTK_UNSIGNED_SHORT,
+    VTK_UNSIGNED_INT
+    //VTK_UNSIGNED_LONG // See issue #5427.
+                        // vtkImageThreshold is not able to fill thresholded images with VTK_LONG_MAX, preventing segment erase.
+  };
+
+  int scalarType = image->GetScalarType();
+  for (int i = 0; i < 3; i++)
```
pieper, line 1857: It would be better not to have the hard-coded number 3 here.  Probably an `#define` would be better.

Libs/vtkSegmentationCore/vtkOrientedImageDataResample.cxx
```diff
@@ -1835,9 +1835,100 @@ bool vtkOrientedImageDataResample::IsLabelInMask(
...
+  {
+    scalarType = VTK_UNSIGNED_SHORT;
+  }
+  else if (scalarRange[1] <= VTK_UNSIGNED_INT_MAX)
+  {
+    scalarType = VTK_UNSIGNED_INT;
+  }
+  else
+  {
+    vtkErrorWithObjectMacro(image, "CastImageToSmallestIntegerType: Scalar range is too large to fit into an integer type");
```
pieper, line 1912: We could enable `ClampOverflow` in `vtkImageCast` and make this a warning rather than an error.  It might be more useful to the user to have something load rather than fail.  Or make it an option if we don't want to lose data with only a warning.

Libs/vtkSegmentationCore/vtkOrientedImageDataResample.cxx
```diff
@@ -1835,9 +1835,100 @@ bool vtkOrientedImageDataResample::IsLabelInMask(
...
+    return false;
+  }
+
+  if (scalarType == image->GetScalarType())
```
pieper, line 1916: Shouldn't we also check that any floating point data is losslessly castable to integer and warn if not?  Otherwise we would be silently losing data.  I know at least some of the float data that a user tried to load as a segmentation recently had non-integral values.

pieper (commented):
Thanks for working on this 👍 Just a few comments/questions.

Libs/vtkSegmentationCore/vtkOrientedImageDataResample.cxx
Sunderlandkyl: Actually, since validScalarTypes is defined locally it probably make sense to use:
```
for (vtkIdType validScalarType : validScalarTypes)
```

Unless we should define the list of valid scalar types elsewhere.

Libs/vtkSegmentationCore/vtkOrientedImageDataResample.cxx
Sunderlandkyl: Ok, we now clamp overflow and log a warning.

Libs/vtkSegmentationCore/vtkOrientedImageDataResample.cxx
Sunderlandkyl: We don't check specifically for lossless castability, however we always log a warning when casting from a floating point type to an integer type so that the user knows that it is a possibility.

The non-integral values are truncated during cast, which is consistent with how they are displayed in labelmap nodes.

Libs/vtkSegmentationCore/vtkOrientedImageDataResample.h
```diff
@@ -183,10 +183,25 @@ class vtkSegmentationCore_EXPORT vtkOrientedImageDataResample : public vtkObject
...
 
+  /// Determine if the segmentation image scalar type is valid for representing segmentations
+  /// \param image Image to validate
+  /// \return True if the image scalar type is valid, false otherwise
+  static bool IsSegmentationScalarTypeValid(vtkImageData* image);
+
+  /// Validate the segmentation image type, casting to another type if necessary.
+  /// \param image Image to validate
+  /// \return True if the image scalar type is valid, or is cast to a valid type, false otherwise
+  static bool ValidateSegmentationImageType(vtkImageData* image);
```
lassoan, line 194: This method performs two quite simple operations, which is probably not used that often (and it makes error reporting a bit more complicated). So, I think we probably don't need this.

Libs/vtkSegmentationCore/vtkOrientedImageDataResample.cxx
```diff
@@ -1835,9 +1835,125 @@ bool vtkOrientedImageDataResample::IsLabelInMask(
   return valueFound;
 }
 
+//----------------------------------------------------------------------------
+bool vtkOrientedImageDataResample::IsSegmentationScalarTypeValid(vtkImageData* image)
```
lassoan, line 1839: It is a bit complex that the scalar types are checked in two slightly different ways (one is segmentation-specific as it excludes long int and float types, the other is somewhat-segmentation-specific as it excludes long int types).

It would be better to have just one segmentation-specific check, which based on the scalar typa and range tells if the the image type is TYPE_OK, TYPE_CONVERSION_NEEDED, TYPE_CONVERSION_TRUNCATION_NEEDED, TYPE_ERROR. Depending on this, warning or error is reported and/or logged and casting is called. Casting does not need to do any warnings (it is a casting function, so the developer must know that casting may cause data loss; but maybe in the future if we need it then we could add an enum return value to distinguish between lossless and potentially lossy conversion).

lassoan (commented):
Thank you, it looks very nice and it is great that you added a specific test for it. We could simplify the checking and error/warning reporting a bit - see my inline comments.

Modules/Loadable/Segmentations/Logic/vtkSlicerSegmentationsModuleLogic.cxx
```diff
@@ -1441,6 +1441,34 @@ bool vtkSlicerSegmentationsModuleLogic::ImportLabelmapToSegmentationNode(vtkMRML
...
+        break;
+      case vtkOrientedImageDataResample::TYPE_CONVERSION_CLAMPING_NEEDED:
+        vtkGenericWarningMacro(
+          "vtkMRMLSegmentationStorageNode::ImportLabelmapToSegmentationNode: "
+          "Segmentation is outside the range of values that can be represented by supported integer types and will be clamped");
+        break;
+      case vtkOrientedImageDataResample::TYPE_ERROR:
+        vtkGenericWarningMacro("vtkMRMLSegmentationStorageNode::ImportLabelmapToSegmentationNode: "
+          "Failed to cast image to integer type.");
+        return 0;
```
lassoan, line 1468:
```suggestion
        return false;
```

Libs/MRML/Core/vtkMRMLSegmentationStorageNode.cxx
```diff
@@ -694,6 +694,25 @@ int vtkMRMLSegmentationStorageNode::ReadBinaryLabelmapRepresentation(vtkMRMLSegm
   }
 
   // Read succeeded
+  int ret = vtkOrientedImageDataResample::IsSegmentationScalarTypeValid(imageData);
+  switch (ret)
+  {
+    case vtkOrientedImageDataResample::TYPE_CONVERSION_TRUNCATION_NEEDED:
+      vtkWarningToMessageCollectionMacro(this->GetUserMessages(), "vtkMRMLSegmentationStorageNode::ReadBinaryLabelmapRepresentation",
+        "Segmentation is a floating point scalar type and will be cast to an integer type. Voxel values may be truncated");
```
lassoan: I don't see where the values are actually truncated or clamped.

Modules/Loadable/Segmentations/Logic/vtkSlicerSegmentationsModuleLogic.cxx
```diff
@@ -1441,6 +1441,34 @@ bool vtkSlicerSegmentationsModuleLogic::ImportLabelmapToSegmentationNode(vtkMRML
     padder->Update();
     labelOrientedImageData->ShallowCopy(padder->GetOutput());
 
+    int ret = vtkOrientedImageDataResample::IsSegmentationScalarTypeValid(labelOrientedImageData);
+    switch (ret)
+    {
+      case vtkOrientedImageDataResample::TYPE_CONVERSION_TRUNCATION_NEEDED:
+        vtkGenericWarningMacro(
```
lassoan, line 1456: It would be nice to able to return this as userMessages to allow the error message to be displayed in the GUI.

Libs/MRML/Core/vtkMRMLSegmentationStorageNode.cxx
```diff
@@ -694,6 +694,25 @@ int vtkMRMLSegmentationStorageNode::ReadBinaryLabelmapRepresentation(vtkMRMLSegm
...
+      break;
+    case vtkOrientedImageDataResample::TYPE_CONVERSION_CLAMPING_NEEDED:
+       vtkWarningToMessageCollectionMacro(this->GetUserMessages(), "vtkMRMLSegmentationStorageNode::ReadBinaryLabelmapRepresentation",
```
lassoan, line 705: It would be nice to able to return this as userMessages to allow the error message to be displayed in the GUI.

lassoan (commented):
I don't see where the casting of float segmentation file to int happens and based on my testing it seems that it does not get converted. Please check!

I also added a commit that returns warning messages from the logic, which allows the warning to be displayed to the user. Maybe we will decide later to not display this warning to the user, to avoid unnecessary confusion, but it may worth giving it a try to educate users that float segmentations are not good.

Libs/MRML/Core/vtkMRMLSegmentationStorageNode.cxx
Sunderlandkyl: Isn't this one returned already?

Libs/vtkSegmentationCore/vtkOrientedImageDataResample.h
```diff
@@ -183,10 +183,39 @@ class vtkSegmentationCore_EXPORT vtkOrientedImageDataResample : public vtkObject
...
     int extent[6]=nullptr, int maskThreshold=0);
 
+  enum
+  {
+    TYPE_OK,
+    TYPE_CONVERSION_NEEDED,
+    TYPE_CONVERSION_TRUNCATION_NEEDED,
+    TYPE_CONVERSION_CLAMPING_NEEDED,
+    TYPE_ERROR
+  };
```
jcfr, line 193:
```suggestion
  };

```

Libs/vtkSegmentationCore/vtkOrientedImageDataResample.h
```diff
@@ -183,10 +183,39 @@ class vtkSegmentationCore_EXPORT vtkOrientedImageDataResample : public vtkObject
   static bool IsLabelInMask(vtkOrientedImageData* binaryLabelmap, vtkOrientedImageData* mask,
     int extent[6]=nullptr, int maskThreshold=0);
 
+  enum
```
jcfr, line 186:
```suggestion
  enum ImageTypeCheckResult
```

Libs/vtkSegmentationCore/vtkOrientedImageDataResample.h
```diff
@@ -183,10 +183,39 @@ class vtkSegmentationCore_EXPORT vtkOrientedImageDataResample : public vtkObject
...
+  {
+    TYPE_OK,
+    TYPE_CONVERSION_NEEDED,
+    TYPE_CONVERSION_TRUNCATION_NEEDED,
+    TYPE_CONVERSION_CLAMPING_NEEDED,
+    TYPE_ERROR
+  };
+  /// Determine if the segmentation image scalar type is valid for representing segmentations
```
jcfr, line 194:
```suggestion
  /// Determine if the scalar type of the image is valid for representing segmentations.
```

Libs/vtkSegmentationCore/vtkOrientedImageDataResample.h
```diff
@@ -183,10 +183,39 @@ class vtkSegmentationCore_EXPORT vtkOrientedImageDataResample : public vtkObject
...
+  /// \param image Image to validate
+  /// \return Value representing if the image scalar type is valid, and if not, what kind of conversion is needed
```
jcfr, line 196:
```
  /// \return ImageTypeCheckResult value representing if the image scalar type is valid, and if not, what kind of conversion is needed.
  /// \sa vtkImageData::GetScalarType()
```

Libs/vtkSegmentationCore/vtkOrientedImageDataResample.h
```diff
@@ -183,10 +183,39 @@ class vtkSegmentationCore_EXPORT vtkOrientedImageDataResample : public vtkObject
...
+  static int IsSegmentationScalarTypeValid(vtkImageData* image);
```
jcfr, line 197:
```suggestion
  static int IsImageScalarTypeValid(vtkImageData* image);
``

File: Libs/vtkSegmentationCore/vtkOrientedImageDataResample.h [hunk 15]
```diff
@@ -183,10 +183,39 @@ class vtkSegmentationCore_EXPORT vtkOrientedImageDataResample : public vtkObject
...
+
+  /// Determine the smallest integer type that can contain the specified scalar range
+  /// \param scalarRange Range of the scalar values that should be representable by the image data type
+  /// \return Smallest integer type that can contain the specified scalar range
```
jcfr, line 201:
```suggestion
  /// Determine the smallest integer type that can contain the specified scalar range.
  /// \param scalarRange Range of the scalar values that should be representable by the image data type
  /// \return Smallest integer type that can contain the specified scalar range. Possible values for are VTK_UNSIGNED_CHAR, VTK_CHAR, VTK_UNSIGNED_SHORT, VTK_SHORT, VTK_UNSIGNED_INT, VTK_INT and -1 if no valid integer type can contain the scalar range.
```

Libs/vtkSegmentationCore/vtkOrientedImageDataResample.h
```diff
@@ -183,10 +183,41 @@ class vtkSegmentationCore_EXPORT vtkOrientedImageDataResample : public vtkObject
...
+
+  /// Determine if the scalar type of the image is valid for representing segmentations.
+  /// \param image Image to validate
+  /// \return ImageTypeCheckResult value representing if the image scalar type is valid, and if not, what kind of conversion is needed.
+  /// \sa vtkImageData::GetScalarType()
+  static int IsImageScalarTypeValid(vtkImageData* image);
+
+  /// Determine the smallest integer type that can contain the specified scalar range.
+  /// \param scalarRange Range of the scalar values that should be representable by the image data type
+  /// \return Smallest integer type that can contain the specified scalar range. Possible values for are VTK_UNSIGNED_CHAR, VTK_CHAR, VTK_UNSIGNED_SHORT, VTK_SHORT, VTK_UNSIGNED_INT, VTK_INT and -1 if no valid integer type can contain the scalar range.
```
jcfr, line 203:
```suggestion
  /// \return Smallest integer type that can contain the specified scalar range. Possible values are VTK_UNSIGNED_CHAR, VTK_CHAR, VTK_UNSIGNED_SHORT, VTK_SHORT, VTK_UNSIGNED_INT, VTK_INT and -1 if no valid integer type can contain the scalar range.
```

jcfr (commented):
Thanks @Sunderlandkyl  , after addressing that last nitpick :+1:

jcfr (approved):

lassoan:
Nice work, thank you @Sunderlandkyl

PR #7606, opened 2024-02-26
[5.6] DOC: Add example script for resampling segmentation to higher resolution

lassoan:
Add an example script to documentation of Slicer-5.6 stable.


jcfr:
Should similar script be added to the current[^1] documentation as well ?

[^1]: https://github.com/Slicer/Slicer/blob/main/Docs/developer_guide/script_repository/segmentations.md?plain=1

lassoan:
It'll be simpler to do this in the main branch - see https://github.com/Slicer/Slicer/pull/7607

PR #7609, opened 2024-02-29
BUG: Fix loading of empty segmentation

### LINKED ISSUE #7576: Threshold Ranger Issue: Not Working When Saving and Restoring Scene with Segment Editor open
## Summary

I downloaded sample data (e.g., MRHead SampleData) and, without further action, saved the scene as a bundle with the segment editor module open. When I loaded the bundle again, I created one segment and attempted to change the threshold value. However, I noticed a bug: the threshold ranger slicer and spinner are not changeable; they are in a disabled state. I have only tested this issue with the threshold effect, and I am not certain if it occurs with other effects as well.

## Steps to reproduce
    1. Navigate to File → Download Sample Data → MRHead.
    2. Open the Segment Editor module.
    3. Press Ctrl+S to save the scene as an .mrb bundle with the segment editor open.
    4. Go to File → Close Scene.
    5. Go to File → Add Data → Choose file(s) and load the previously saved .mrb file.
    6. In the Segment Editor, add one or more segments.
    7. Navigate to the Threshold effect and attempt to change the threshold range.
    8. Notice that it won't work => the Slicer should allow modification of the threshold value.

Video reference: https://drive.google.com/file/d/1tyr15I-fqalwKyngdpUph0i1hq9-lcDI/view?usp=sharing
Screen capture:

image

## Environment
- Slicer version: Slicer-5.6.1-2023-12-12

## References
- Discourse: https://discourse.slicer.org/t/bug-while-saving-data-as-bundle-with-no-segments/34148?u=mjamal

lassoan:
NRRD file cannot be created without voxel data, so an 1x1x1 array was used as image data. However, when the empty segmentation was loaded then this single-voxel geometry was treated as a valid geometry. Fixed by detecting this special case of a single voxel in the segmentation, and considering the geometry invalid in this case.

fixes #7576

### DIFF
--- a/Libs/MRML/Core/Testing/vtkMRMLSegmentationStorageNodeTest1.cxx
+++ b/Libs/MRML/Core/Testing/vtkMRMLSegmentationStorageNodeTest1.cxx
@@ -23,6 +23,7 @@ Care Ontario.
 #include "vtkMRMLScene.h"
 #include "vtkMRMLSegmentationNode.h"
 #include "vtkMRMLSegmentationStorageNode.h"
+#include "vtkOrientedImageData.h"
 #include "vtkSegmentationConverterFactory.h"
 
 // Converter rules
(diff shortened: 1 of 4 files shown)


Libs/MRML/Core/Testing/vtkMRMLSegmentationStorageNodeTest1.cxx
```diff
@@ -114,5 +116,44 @@ int vtkMRMLSegmentationStorageNodeTest1(int argc, char * argv[] )
...
+    // the storage node should ignore that when reading the file (single-voxel volume is a special case).
+    std::string segmentationGeometryString = segmentation->DetermineCommonLabelmapGeometry(
+      vtkSegmentation::EXTENT_UNION_OF_EFFECTIVE_SEGMENTS_AND_REFERENCE_GEOMETRY);
+    CHECK_STD_STRING(segmentationGeometryString, "");
+
+    // Clean up
+    vtksys::SystemTools::RemoveFile(emptySegmentationFilename);
+  }
+
+    return EXIT_SUCCESS;
```
jcfr, line 158:
```suggestion
  return EXIT_SUCCESS;
```

PR #7610, opened 2024-02-29
BUG: Make volume rendering preset registration more robust

lassoan:
When a volume rendering preset was added to the preset scene, the input preset node was moved to that scene, thereby making the scene that contained the input preset node unusable. Fixed by making a copy of the input preset node instead of moving it.

### DIFF
--- a/Modules/Loadable/VolumeRendering/Logic/vtkSlicerVolumeRenderingLogic.cxx
+++ b/Modules/Loadable/VolumeRendering/Logic/vtkSlicerVolumeRenderingLogic.cxx
@@ -1416,12 +1416,13 @@ bool vtkSlicerVolumeRenderingLogic::IsDifferentFunction(vtkColorTransferFunction
 }
 
 //---------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


Modules/Loadable/VolumeRendering/Logic/vtkSlicerVolumeRenderingLogic.cxx
```diff
@@ -292,48 +292,48 @@ void vtkSlicerVolumeRenderingLogic::ChangeVolumeRenderingMethod(const char* disp
 void vtkSlicerVolumeRenderingLogic::UpdateVolumeRenderingDisplayNode(vtkMRMLVolumeRenderingDisplayNode* node)
 {
   if (!node)
-  {
+    {
```
jcfr, line 295: we should avoid reverting back to the old style

jcfr (approved):

Modules/Loadable/VolumeRendering/Logic/vtkSlicerVolumeRenderingLogic.cxx
lassoan: I've fixed it now (I still had the old brace style in my original working copy).

PR #7611, opened 2024-02-29
ENH: Make translations in Python scripts more robust

### LINKED ISSUE #7235: Make Python string translation more robust
If a Python string with placeholders (e.g., `Loading {filename}`) is translated incorrectly so that the placeholder is not in the translated string (e.g., `Carga {nombre de archivo}`) then currently an exception is raised and execution of the Python code stops.

While it is useful to notify the developer about a mismatched placeholder, raising an exception is too disruptive, as it typically makes the entire executed function to fail. See for example error reported here: https://github.com/Slicer/SlicerLanguagePacks/issues/40

We should make string translation more robust: check if placeholders in the source string are all available in the translated string. If there is any mismatch then log a warning and do not use the translated string.

lassoan:
If a Python string with placeholders (e.g., Loading {filename}) is translated incorrectly so that the placeholder is not in the translated string (e.g., Carga {nombre de archivo}) then currently an exception is raised and execution of the Python code stops.

Fixed by only using the translated string if the placeholders match. Otherwise a warning is logged and the translation is not used.

fixes #7235

### DIFF
--- a/Base/Python/slicer/i18n.py
+++ b/Base/Python/slicer/i18n.py
@@ -1,6 +1,7 @@
-import os
 import inspect
-
+import logging
+import os
+import re
 
 def translate(context, text):
     """Translate message to the current application language.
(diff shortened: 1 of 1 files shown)


Base/Python/slicer/i18n.py
```diff
@@ -51,6 +52,25 @@ def tr(text):
       statusText = _("Idle") if idle else _("Running")
 
     """
+
+    def findBracedStrings(text):
+        """Get all strings enclosed in braces, except when enclosed in double-braces."""
```
jcfr, line 57:
```suggestion
"""Get all strings corresponding to f-string expressions.

All expression delimited by curly braces are returned except the ones enclosed in
double-braces.

See https://docs.python.org/3/reference/lexical_analysis.html#f-strings.
"""
```

Base/Python/slicer/i18n.py
```diff
@@ -51,6 +52,25 @@ def tr(text):
...
+
     filename = inspect.stack()[1][1]
     contextName = getContext(filename)
-    return translate(contextName, text)
+    translatedText = translate(contextName, text)
+
+    # Accept the translation only if all placeholders are present in the translated text to prevent runtime errors.
+    # For example:
+    #   text = "{count} items will be deleted"
+    #   translatedText = "{compter} elements seront supprimes" (incorrect, because `count` should not have been translated)
```
jcfr, line 69:
```suggestion
    #   text = f"{count} items will be deleted"
    #   translatedText = f"{compter} éléments seront supprimés" (incorrect, because `count` should not have been translated)
```

lassoan: We use `.format` to make sure that no code is executed, so adding the `f` before the string could be a bit misleading.

Issue #7588, opened 2024-02-14
Display warning when 3D image is saved as TIFF file


lassoan:
## Summary

Users may not know that TIFF file cannot store image orientation and Z spacing and so they lose this information when saving the scene.

See for example this user error report: https://discourse.slicer.org/t/older-scenes-and-specimens-loading-incorrectly/34328

## Steps to reproduce

- Load MRHead example
- Save as TIFF file => ERROR: no warning is displayed
- Load the TIFF file => ERROR: the image is distorted

## Environment
- Slicer version: Slicer-5.6.1
- Operating system: Windows11

lassoan:
It would be better to change the default file format for 3D images loaded from TIFF to nrrd.

pieper:
I believe in the past the mrml node's ijkToRAS was preserved when saving to formats like jpg or tiff (e.g. when the mrml file points to the original data), but avoiding any ambiguity in save/restore is the most important goal, so making nrrd be the default for new storage makes good sense.

lassoan:
Fixed in https://github.com/Slicer/Slicer/pull/7655

PR #7614, opened 2024-03-01
ENH: Add TIFF format warning window

aramSofthenge:
Adds warning for TIFF file format saving. Prevents potential data loss.

Issue #7588

### DIFF
--- a/Base/QTGUI/qSlicerSaveDataDialog.cxx
+++ b/Base/QTGUI/qSlicerSaveDataDialog.cxx
@@ -807,6 +807,7 @@ bool qSlicerSaveDataDialogPrivate::saveNodes()
     // file properties
     QFileInfo file = this->file(row);
     QString format = this->format(row);
+
     qSlicerIOOptions* options = this->options(row);
     vtkMRMLNode* node = vtkMRMLNode::SafeDownCast(this->object(row));
     vtkMRMLStorableNode* const storableNode = vtkMRMLStorableNode::SafeDownCast(this->object(row));
--- a/Libs/MRML/Core/vtkMRMLVolumeArchetypeStorageNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLVolumeArchetypeStorageNode.cxx
@@ -86,6 +86,7 @@ vtkMRMLVolumeArchetypeStorageNode::vtkMRMLVolumeArchetypeStorageNode()
   this->SingleFile  = 0;
   this->UseOrientationFromFile = 1;
   this->DefaultWriteFileExtension = "nrrd";
+  this->TIFFWarningShown = false;
 }
 
 //----------------------------------------------------------------------------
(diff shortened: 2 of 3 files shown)


Base/QTGUI/qSlicerSaveDataDialog.cxx
```diff
@@ -806,7 +814,25 @@ bool qSlicerSaveDataDialogPrivate::saveNodes()
...
-    QString format = this->format(row);
+    QString format = this->format(row).toLower(); // Convert format to lowercase for comparison
+
+    // Check if the format is TIFF, show warning if necessary, and only once
+    if (!tiffWarningShown && (format.contains(".tiff") || format.contains(".tif")))
+    {
+      QMessageBox::StandardButton reply = QMessageBox::warning(this, "TIFF Format Warning",
+        "Saving 3D images in TIFF format may result in loss of orientation and Z-spacing data. "
+        "Consider using .nrrd format for optimal preservation. TIFF is not suitable for 3D images.\n"
+        "Do you want to continue?",
```
pieper, line 825: It would be great if you could mark these strings for translation.  You could follow the pattern i the rest of the code and there are more details here: https://github.com/Slicer/SlicerLanguagePacks/blob/main/DevelopersManual.md

lassoan (commented):
Thanks a lot for working on this. The save dialog does not know about limitations related to specific file formats. Instead, warnings can be added at lower level, in the storage node, as done here: https://github.com/Slicer/Slicer/blob/c7fe8657c6a4bc0666685349b3222ff3c1b4fa02/Libs/MRML/Core/vtkMRMLVolumeArchetypeStorageNode.cxx#L704

aramSofthenge:
> Thanks a lot for working on this. The save dialog does not know about limitations related to specific file formats. Instead, warnings can be added at lower level, in the storage node, as done here:
> 
> https://github.com/Slicer/Slicer/blob/c7fe8657c6a4bc0666685349b3222ff3c1b4fa02/Libs/MRML/Core/vtkMRMLVolumeArchetypeStorageNode.cxx#L704

@lassoan That's a good idea, but when I use `vtkWarningToMessageCollectionMacro` the user basically just has it come out with some sort of error/warning even if the file is saved. Is it okay so we see such an error but the files are still saved. Is this an expected behavior?

image

Base/QTGUI/qSlicerSaveDataDialog.cxx
aramSofthenge: I changed some code. Now, I'm not sure if this request still is relevant. I checked similar parts of that method, and they don't use `tr()`. Should I still add it here? Let me know if I should

pieper: It's not strictly required but we'll want to do it someday.  When people are working on formatting strings it can be easier to add the translation hooks while everything is fresh.  Otherwise someone later needs to make sure they understand the logic.

aramSofthenge:
@lassoan I'd like to point out that the update has been added and so has the location of the code

lassoan:
I had to switch to using placeholders in translations (it is not possible to translate sentence fragments because the order of words is different in each language), made the TIFF check more specific (report error only if actually data is lost by saving into TIFF), and remove the suppression of warning message (no need to suppress warning message - saving is completed, checkbox in Save data dialog is unchecked, so the second click on save will close the dialog).

I could not push to this branch, so I had to create a [new pull request](https://github.com/Slicer/Slicer/pull/7655) and I'm closing this one.

PR #7616, opened 2024-03-02
Small API improvements and bugfixes

lassoan:


Libs/MRML/Core/vtkMRMLSubjectHierarchyNode.cxx
```diff
@@ -1298,6 +1308,26 @@ void vtkSubjectHierarchyItem::SetUID(std::string uidName, std::string uidValue)
   this->Modified();
 }
 
+//---------------------------------------------------------------------------
+bool vtkSubjectHierarchyItem::RemoveUID(std::string uidName)
```
sedrakSofthenge, line 1312: why not passing const std::string& instead here and in other defined functions that get std::string as argument?

Libs/MRML/Core/vtkMRMLSubjectHierarchyNode.cxx
```diff
@@ -1298,6 +1308,26 @@ void vtkSubjectHierarchyItem::SetUID(std::string uidName, std::string uidValue)
...
+{
+  if (this->UIDs.size() == 0)
+  {
+    return false;
+  }
+
+  // Use the find function to prevent adding an empty UID to the map
+  if (this->UIDs.find(uidName) != this->UIDs.end())
```
sedrakSofthenge, line 1320: maybe use iterator to avoid double lookup?
  if (auto it = this->UIDs.find(uidName); it != this->UIDs.end())
  {
    this->UIDs.erase(it);
    this->Modified();
    return true;
  }

Libs/MRML/Core/vtkMRMLSubjectHierarchyNode.cxx
```diff
@@ -1309,6 +1339,25 @@ std::string vtkSubjectHierarchyItem::GetUID(std::string uidName)
   return std::string();
 }
 
+//---------------------------------------------------------------------------
+std::vector<std::string> vtkSubjectHierarchyItem::GetUIDNames()
+{
+  std::vector<std::string> uidNameList;
+  std::map<std::string, std::string>::iterator uidIt;
+  for (uidIt=this->UIDs.begin(); uidIt!=this->UIDs.end(); ++uidIt)
```
sedrakSofthenge, line 1347: maybe use const iteration instead? with const_iterator, cbegin, cend

Libs/MRML/Core/vtkMRMLSubjectHierarchyNode.cxx
```diff
@@ -1309,6 +1339,25 @@ std::string vtkSubjectHierarchyItem::GetUID(std::string uidName)
...
+  {
+    uidNameList.push_back(uidIt->first);
+  }
+  return uidNameList;
+}
+
+//---------------------------------------------------------------------------
+bool vtkSubjectHierarchyItem::HasUID(std::string uidName)
+{
+  return (this->UIDs.size() > 0
```
sedrakSofthenge, line 1357: `this->UIDs.size() > 0` this checking looks redundant, find will always return UIDS.end() if UIDs is empty and it does not provide performance optimization as well.

Libs/MRML/Core/vtkMRMLSubjectHierarchyNode.cxx
lassoan: @sedrakSofthenge I agree with all your comments. I had to choose between consistency (with the existing code that does the same for item attributes) and potential performance gain. Since the performance gain is unlikely to be perceivable in any real-world scenarios (since subject hierarchy typically contains less than 20 items, and never more than a few hundred), I chose to go with less changes (and thus faster development and less chance of breaking anything). I'll see if I can spend a little more time with this to address some of your comments.

lassoan:
I've simplified the code as suggested, except replacing std:string arguments , as it is not clear if that should increase the performance.

Libs/MRML/Core/vtkMRMLSubjectHierarchyNode.cxx
jcfr: Ideally, we should indeed use `const std::string& uidName`

In addition of avoiding unnecessary copy, it also indicates that the `uidName` parameter will not be modified within the function.

That said, the use of `std::string` vs `const std::string&` is distributed ...

```
$ ack --headers "\(.+std::string \w+" --no-filename | wc -l
108

$ ack --headers "const std::string\& \w+" --no-filename | wc -l
199
```

... and aiming on the side of consistency is sensible.

----------------

On the other hand, the use of `const QString&` is dominant:

```
$ ack --headers "\(.+QString \w+" --no-filename | wc -l
47

$ ack --headers "const QString\& \w+" --no-filename | wc -l
491
```

Libs/MRML/Core/vtkMRMLLayoutNode.h
```diff
@@ -139,15 +139,23 @@ class VTK_MRML_EXPORT vtkMRMLLayoutNode : public vtkMRMLAbstractLayoutNode
   /// Modifies a layout description for integer identifier
   /// "layout". Returns false without making any modifications if the
   /// integer identifier "layout" has NOT already been added.
+  /// If layoutDescription is empty then the layout is removed.
```
jcfr, line 142:
```suggestion
  /// If layoutDescription is empty or NULL then the layout is removed.
```

Libs/MRML/Core/vtkMRMLSubjectHierarchyNode.cxx
sedrakSofthenge: overall compiler probably will optimize this to pass reference, but it's worth to use const& in general (with some exceptions, say built-in types) where a variable is not modified inside the function. From the consistency point I actually don't really think its going to break anything, and maybe in further reviews it's worth to ask for adjusting the functions being added.

lassoan: There is no practical performance difference in performance. Also, this is not a public API, so it does not really serve as an example for how to do things elsewhere. So, I would recommend proceding with the merge, while we continue the discussion for our education and for establishing coding best practices.

In cxx11 and later,  const ref is not always better. See lots of discussions on this on Stack Overflow - such as https://stackoverflow.com/questions/10789740/passing-stdstring-by-value-or-reference We should discuss if passing by vale and using std::move is preferable. Also, using const on the *public* API of the method depending on some *private* implementation details (whether that string is used in some operations that may modify it or not) would be very odd, because in general you want to keep internal implementation details modifiable without impacting the public API. This should be also discussed in the context of using const char* and whether we should move towards using std::string everywhere we can. Oerhaos we should create an issue or discourse post to further discuss this. Even if we don't change any code just establish coding guidelines would be a useful outcome.

sedrakSofthenge: I can of course approve, not sure if that is sufficient for merging. My comments were actually more of optional to take into account, not mandatory, so not sure if this needs my approval.

Regarding comments: Thanks for considering them, but I actually don't see the changes in this PR regarding my other comments, but you mentioned they are addressed, am I missing something?
About string vs const string&, I still think it's better to use const&, and the provided link also confirms my point ( the function does not modify the string):
When passing by value (assuming compiler does not optimize the code) it copies the string by calling a copy constructor so we can 2 copies of the string, the original one and the one local to the function where it's passed. When passing by const& we make sure to not modify it inside the function (in the case of functions in this PR and in a lot of other cases in the project) and to not copy, but just pass the memory address to the string. In this case we also cannot rely on move semantics, since it implies all callers of this function have objects to be used with rvalue (temporary objects probably).
Although it's up to you to fix this in this PR or leave it, I think it's worth to follow up on this and make sure in further PRs using const& for classes, structs in general when the passed arg is not being modified inside the function.

char* vs string is another topic, I think it has wider coverage, and honestly I don't know the actual reasons of using char* vs string here as well.

sedrakSofthenge (approved):

jcfr (approved):

jcfr:
Following our discussion during this morning developer meeting, I also created the following issue:
* https://github.com/Slicer/Slicer/issues/7621

PR #7624, opened 2024-03-05
ENH: Support returning the result when executing Python from qSlicerWebWidget

jcfr:
Update `qSlicerWebPythonProxy` to allow specifying `EvalInput` as mode and ensure a result is returned as `QVariant`.

For example:

```js
await slicerPython.evalPython("40+2")
""

await slicerPython.evalPython("40+2", slicerPython.EvalPythonMode.EvalInput)
"42"
```

Note that the associated CTK changes are not strictly required to support passing the evalPython mode parameter.

List of CTK changes:

```
$ git shortlog 5a5c9f62a..a0ea618f1 --no-merges
      ENH: Support using ctkAbstractPythonManager::ExecuteStringMode from Python
```

### DIFF
--- a/Base/QTGUI/qSlicerWebPythonProxy.h
+++ b/Base/QTGUI/qSlicerWebPythonProxy.h
@@ -38,6 +38,22 @@ public:
   /// Constructor
   explicit qSlicerWebPythonProxy(QObject *parent = nullptr);
 
+  /// This enum maps to ctkAbstractPythonManager execution modes Py_eval_input,
+  /// Py_file_input and Py_single_input.
+  ///
+  /// \see https://docs.python.org/3/c-api/veryhigh.html#Py_eval_input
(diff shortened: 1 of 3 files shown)


Base/QTGUI/qSlicerWebPythonProxy.h
```diff
@@ -52,7 +68,7 @@ public slots:
   /// running python code from web pages.
   ///
   /// \sa qSlicerWebWidget::initializeWebEngineProfile
-  QString evalPython(const QString &python);
+  QString evalPython(const QString &python, int mode = FileInput);
```
jcfr, line 71: I ended up using `int` as the type because I didn't find a way to make it work with:

```cpp
QString evalPython(const QString &python, EvalPythonMode mode = FileInput);
```

pieper (approved):
Thanks @jcfr 👍

PR #7625, opened 2024-03-06
BUG: Remove obsolete Webkit-specific integration code from qSlicerWebWidget

jcfr:
This commit updates `qSlicerWebWidget` to remove obsolete code related to Webkit integration. The removal of this code prevents the display of a warning message that occurs when loading a page after opening the extension manager or explicitly instantiating a `qSlicerWebWidget` and setting its URL property.

The warning message was:

```
"JQuery not loaded - Failed to trigger webkitvisibilitychange"
```

This integration was initially introduced in commit 8d7f94e4ee ("BUG: Prevent extensions from being loaded twice when 'Install' tab initially hidden", 2013-02-04), but the support for Webkit was later removed in commits:
* 6bd2db52ef ("COMP: Remove support for building against Qt WebKit module", 2019-03-08)
* f3edc8f609 ("STYLE: Remove obsolete code supporting use of Qt WebKit module", 2019-03-08)

### DIFF
--- a/Base/QTGUI/qSlicerWebWidget.cxx
+++ b/Base/QTGUI/qSlicerWebWidget.cxx
@@ -135,8 +135,6 @@ void qSlicerWebWidgetPrivate::init()
                    this, SLOT(onAppAboutToQuit()));
   this->verticalLayout->insertWidget(0, this->WebView);
 
-  this->WebView->installEventFilter(q);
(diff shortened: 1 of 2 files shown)


pieper (approved):
that's great, thanks for cleaning this up 👍

PR #7628, opened 2024-03-13
BUG: Fix crash in Markups widget when closing scene

Sunderlandkyl:
When closing the scene, qSlicerMarkupsModuleWidget::updateWidgetFromMRML() would access the subjectHierarchyNode without first checking if it exists, causing a crash.

Fixed by adding a nullptr check.

See: https://discourse.slicer.org/t/weird-trouble-with-markups/34872/

### DIFF
--- a/Modules/Loadable/Markups/qSlicerMarkupsModuleWidget.cxx
+++ b/Modules/Loadable/Markups/qSlicerMarkupsModuleWidget.cxx
@@ -650,11 +650,11 @@ void qSlicerMarkupsModuleWidgetPrivate::setMRMLMarkupsNodeFromSelectionNode()
   // Select current markups node
   vtkMRMLMarkupsNode* currentMarkupsNode = vtkMRMLMarkupsNode::SafeDownCast(this->selectionNodeActivePlaceNode());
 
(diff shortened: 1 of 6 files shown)


lassoan (commented):
Thank you, this is perfect as a corrective action.

As preventive actions, could you please do the followings:
- Add to qMRMLSubjectHierarchyTreeView::subjectHierarchyNode() documentation that is may return a nullptr when the scene is being closed.
- Check all other uses of `->subjectHierarchyNode()` in the entire Slicer source tree (in all .cxx files). I did some spot checks and found multiple unprotected uses:
  - qMRMLSubjectHierarchyTreeViewPrivate::firstAppliedTransformToSelectedItems() 
  - qMRMLSubjectHierarchyTreeView::firstSelectedSubjectHierarchyItemInBranch
  - qMRMLSubjectHierarchyTreeView::onCustomContextMenu
  - qSlicerMarkupsModuleWidget::updateWidgetFromMRML()

Sunderlandkyl:
Ok. I've added the documentation to the qMRMLSubjectHierarchyTreeView header, and have added nullptr checks to all other cases without a check that I could find.

lassoan (approved):
Thank you, it looks good.

PR #7649, opened 2024-03-20
ENH: Improve ultrasound reading

lassoan:
Add DICOM patcher rule to split ultrasound series and read pixel spacing for single-region ultrasound images.

@ungi


pieper (approved):
LGTM 👍

PR #7651, opened 2024-03-20
ENH: Add translation for VolumeDisplayPresets.json

mhdiop:
* Slicer/SlicerLanguagePacks#43


lassoan (approved):
Looks great, thank you. I've just generalized it a bit to make it easy to extract translatable strings from any json file.

PR #7655, opened 2024-03-22
Show warning window if data loss occurst when saving to tiff

lassoan:
This pull request supersedes https://github.com/Slicer/Slicer/pull/7614 by @aramSofthenge


Libs/MRML/Core/vtkMRMLI18N.cxx
```diff
@@ -147,3 +147,45 @@ std::string vtkMRMLI18N::Translate(const char *context, const char *sourceText,
...
+      {
+        // Escaped '%'
+        output += "%";
+        ++it;
+        continue;
+      }
+      else if (std::isdigit(nextChar))
+      {
+        // Found a valid placeholder (e.g., %1, %2, %3, ..., %9)
+        if (nextChar == '1' && arg1) { output += arg1; }
```
sedrakSofthenge, line 174: Maybe the using like below is better, to avoid multiple times of arg1 check and if-else blocks
if (arg1) {
switch (nextChar) {
case '1': ... break;
case '2' ... break;
}
}

Libs/MRML/Core/vtkMRMLVolumeArchetypeStorageNode.cxx
```diff
@@ -750,17 +748,49 @@ int vtkMRMLVolumeArchetypeStorageNode::WriteDataInternal(vtkMRMLNode *refNode)
...
+          "VTK file format can only store LPS axis oriented images. Orientation of the saved image may be incorrect."
+          " NRRD file format is recommended for storing this image."));
     }
   }
 
+  // Display warning if saving TIFF file as 3D image
+  if ((vtksys::SystemTools::StringEndsWith(lowerCaseFileName, ".tiff") ||
+       vtksys::SystemTools::StringEndsWith(lowerCaseFileName, ".tif")))
+  {
+    // Display warning if saving VTK file with non-LPS axes (VTK cannot store axis directions)
```
pieper, line 761: Comment should refer to tiff not VTK

pieper, line 761:
```suggestion
    // Display warning if saving tiff file with non-LPS axes (tiff cannot store axis directions)
```

Libs/MRML/Core/vtkMRMLVolumeArchetypeStorageNode.cxx
```diff
@@ -789,8 +819,7 @@ std::string vtkMRMLVolumeArchetypeStorageNode::UpdateFileList(vtkMRMLNode *refNo
   // test whether refNode is a valid node to hold a volume
   if (!refNode->IsA("vtkMRMLScalarVolumeNode") )
   {
-    vtkErrorToMessageCollectionMacro(this->GetUserMessages(), "vtkMRMLVolumeArchetypeStorageNode::UpdateFileList",
-      "Reference node is not a vtkMRMLVolumeNode");
+    vtkErrorMacro("vtkMRMLVolumeArchetypeStorageNode::UpdateFileList: Reference node is not a volume");
```
pieper, line 822: This is a developer-level error so I guess you decided not to report it to the user, but shouldn't the user get some indication of the issue?

pieper (approved):
Just a couple small comments, otherwise LGTM.

Libs/MRML/Core/vtkMRMLVolumeArchetypeStorageNode.cxx
lassoan: Good catch, I will fix it

Libs/MRML/Core/vtkMRMLVolumeArchetypeStorageNode.cxx
lassoan: The user will get a high-level error message that reading of the image failed and more details will be in the log.

I decided not to display these because it would be unnecessary burden for translators to translate these low-level messages that are really only meant for developers, they don't tell anything useful to users. It is also easier for developers if log messages are kept in English.

lassoan:
@pieper please have a look and approve if it looks good. Thank you!

Libs/MRML/Core/vtkMRMLVolumeArchetypeStorageNode.cxx
pieper: Yes, that's what I figured, just wanted to be sure

PR #7656, opened 2024-03-22
ENH: Add DICOM patcher rule to fix compressed ultrasound discolorization

lassoan:
Photometric intepretation field of compressed ultrasound images was ambiguous before DICOM CP156 (https://dicom.nema.org/medical/dicom/Final/cp156_ft.pdf) and there are DICOM images in the wild that contain photometric interpretation value (RGB) that is inconsistent with the compressed transfer syntax (JPEG). See https://groups.google.com/g/comp.protocols.dicom/c/EPkbFOMBOhE for more details.

The new DICOM patcher rule modifies RGB photometric interpretation to YBR_FULL_422 if the transfer syntax of the image is JPEG, because JPEG compression is practically always used with YBR_FULL_422.


pieper (approved):
LGTM 👍

PR #7662, opened 2024-03-25
ENH: Remove DMRIInstall module

jamesobutler:
DMRIInstall module was originally added in https://github.com/Slicer/Slicer/commit/0faa861576d9ceaaf8d5c19499e37671ce5f84f5 and further improved upon on in https://github.com/Slicer/Slicer/commit/474261c7e0ddb2337e6d8a3bd25cef8f129fe4c6 to install the SlicerDMRI extension by way of a Slicer scripted module.

SlicerDMRI extension can be found at https://github.com/SlicerDMRI/SlicerDMRI

This module was unique in that it was the only one to support installing an extension without having to go through the Extensions Manager UI. This module is lacking useful functionality compared to the Extensions Manager as it does not include information regarding the last update date and revision of the extension.

Installing DMRI (removed method):
Module Selector -> Diffusion -> Install Slicer Diffusion Tools (SlicerDMRI)

Installing DMRI (remaining method):
Extensions Manager -> Category "Diffusion" -> Install SlicerDMRI

| DMRIInstall | Extensions Manager SlicerDMRI page|
|--------------|------------------------------------------|
|image|image|


pieper:
@ljod thoughts?

jcfr:
Removing it makes sense. 

Creating an issue describing a generalized version of this extending the module finder and suggesting installation of extensions could be interesting as well.

ljod:
hi everyone thanks. I don't know how many people are using this install pathway, but it is easily findable under our Diffusion menu in case people look there when the module is not yet installed. This is a nice feature and a use case for other modules potentially. However, if maintaining this module in the core is not practical I am ok with removing it.

jamesobutler:
For Slicer Preview, if the SlicerDMRI extension is not building successfully then this module would still allow attempting to install it which will then throw its own error text. For consistent handling of failed extension installs it would seem to be an improvement to consistently handle this through the Extensions Manager interface.

Indeed a new GitHub issue could be created to add an option to the Module Selector to include module names from available Slicer Extensions (would require an internet connection to ping the server). Then users could search for functionality through the Module Selector categories instead of only with the categories in the Extensions Manager.

jcfr:
I was initially thinking about a static index  created based on the extension description, name and list modules. That said, I agree with you that this overkill, the extensions manager fulfills this role nicely.

lassoan:
It could be useful to show all modules provided by tier 1-3 extensions in the module finder. When the user selects it, the module could be automatically installed. I don't know if this is necessary though, how much it is a barrier that users don't know what modules are available in extensions..

pieper:
> show all modules provided by tier 1-3 extensions in the module finder

I would find this useful since it could save many steps.  For example if I want to export a glTF model I could just type `control-F glTF return` to install the OpenAnatomy extension it would be much simpler than the current steps.  It could also help people discover extensions they otherwise wouldn't know about easily.

jamesobutler:
@jcfr per https://discourse.slicer.org/t/2024-03-26-weekly-meeting/35074/2, should we proceed forward with this integration now?

jcfr (approved):

PR #7664, opened 2024-03-28
ENH: Allow aliases for NRRD "type" header in postNRRD

gabuzi:
The NRRD definition lists various aliases
to specify the for the "short" datatype in
the "type" nrrd header field (see
https://teem.sourceforge.net/nrrd/format.html#type).

Pynnrd as a widely used nnrd library
e.g. creates NRRD headers with short
datatypes identified as "int16" and
does not allow to override this.
Consequently, a POST request to the
slicer/volume endpoint will always fail
with data from pynnrrd.

This commit relaxes the type check to
also allow aliases of "short" to pass.

### DIFF
--- a/Modules/Scripted/WebServer/WebServerLib/SlicerRequestHandler.py
+++ b/Modules/Scripted/WebServer/WebServerLib/SlicerRequestHandler.py
@@ -375,7 +375,7 @@ class SlicerRequestHandler(BaseRequestHandler):
                 value = line[colonIndex + 2 :]
                 fields[key] = value
 
-        if fields[b"type"] != b"short":
+        if fields[b"type"] not in [b"short", b"short int", b"signed short", b"signed short int", b"int16", b"int16_t"]:
             raise RuntimeError("Can only read short volumes")
         if fields[b"dimension"] != b"3":
             raise RuntimeError("Can only read 3D, 1 component volumes")


pieper (approved):
Makes sense to me - thanks for the enhancement 👍

pieper:
> any particular reason to restrict to short only? I'd be e.g. interested in float data

No, it was just implementing a single case as a kind of proof-of-concept.  Nrrd has a lot of variants so I only added what seemed like a common case, but it should be pretty easy to add float following the same pattern and some refactoring.

Hope this feature is useful for you.  I use if on and off for some experiments but didn't know if anyone else had ever tried it.  As you may have seen you can manipulate other parts of the scene too, but it's not a comprehensive API at this point.

lassoan (approved):
Thank you, this looks good to me. An even better solution would be to use pynrrd or teem library to read all valid nrrd files, but if this fix is sufficient then it makes sense to integrate as is, because this is a very small, low-risk change.

lassoan:
> almost real-time data visualization

For real-time data visualization Slicer also offers OpenIGTLink connection. It supports many data types (images, transforms, models, point sets, text, commands, etc) at high frame rates (hundreds of frames per second). The protocol is simpler protocol than http, so it is easy to implement in a y environment, even in embedded systems. In Python you can use pyigtl to send/receive data, but there is also C, C++, Java, Matlab, etc. implementations. Slicer can act as client or server, connect to many clients and servers at the same time, and not just receives data but also can automatically send data (on connect and whenever when a node changes).

Modules/Scripted/WebServer/WebServerLib/SlicerRequestHandler.py
```diff
@@ -375,7 +375,7 @@ def postNRRD(self, volumeID, requestBody):
                 value = line[colonIndex + 2 :]
                 fields[key] = value
 
-        if fields[b"type"] != b"short":
+        if fields[b"type"] not in [b"short", b"short int", b"signed short", b"signed short int", b"int16", b"int16_t"]:
             raise RuntimeError("Can only read short volumes")
```
jcfr, line 379:
```suggestion
supported_types = [b"short", b"short int", b"signed short", b"signed short int", b"int16", b"int16_t"]
supported_types_str = [_type.decode("utf-8") for _type in supported_types]
type_str = fields[b"type"].decode("utf-8")
if type_str not in supported_types:
    msg = f"Failed to read volume of type {type_str}. Supported types are {supported_types_str}"
    raise RuntimeError(msg)
```

PR #7669, opened 2024-04-02
Update registration.md

muratmaga:
Replace the reference to ALPACA (which is actually for automated landmarking) with FastModelAlign (based on alpaca, but only for registration) module in SlicerMorph.


pieper (approved):
LGTM 👍

PR #7673, opened 2024-04-02
ENH: Add SH options to control transform interaction handle type visibility

Sunderlandkyl:
Adds the subject hierarchy menu "Interaction options" for enabling/disabling transform handle types (Rotation/Translation/Scaling), and adds all of the interaction options to the view context menu. This commit also fixes an issue where markups interaction handles could not be enabled/disabled correctly, and scale handles could be shown for unsupported markups types.

Re #7570

### DIFF
--- a/Modules/Loadable/Transforms/SubjectHierarchyPlugins/qSlicerSubjectHierarchyTransformsPlugin.cxx
+++ b/Modules/Loadable/Transforms/SubjectHierarchyPlugins/qSlicerSubjectHierarchyTransformsPlugin.cxx
@@ -84,7 +84,12 @@ public:
   QActionGroup* ResetCenterOfTransformationNodeGroup;
 
   QAction* ToggleInteractionAction;
-  QAction* ToggleInteractionItemAction;
+
+  QAction* ToggleInteractionTranslationVisibleAction{ nullptr };
+  QAction* ToggleInteractionRotationVisibleAction{ nullptr };
+  QAction* ToggleInteractionScaleVisibleAction{ nullptr };
+  QMenu* InteractionOptionsMenu{ nullptr };
+  QAction* InteractionOptionsAction{ nullptr };
 
   QVariantMap ViewContextMenuEventData;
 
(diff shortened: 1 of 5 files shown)


Modules/Loadable/Transforms/SubjectHierarchyPlugins/qSlicerSubjectHierarchyTransformsPlugin.cxx
```diff
@@ -57,6 +57,12 @@
 // MRML widgets includes
 #include "qMRMLNodeComboBox.h"
 
+//-----------------------------------------------------------------------------
+const char* INTERACTION_HANDLE_TYPE_PROPERTY = "InteractionHandleType";
+const int INTERACTION_HANDLE_TYPE_TRANSLATION = 0;
```
sedrakSofthenge, line 62: Seems these are mutually exclusive values, if so, then why not using an enum instead?

lassoan: I agree, enum would convey the intent of these values a bit better.

It would also better to move such global symbols into a namespace (private class or anonymous namespace). In this case we have a private implementation class, so we could move all these constants into `qSlicerSubjectHierarchyTransformsPluginPrivate`.

I did a search in the whole Slicer source code and only found a handful of places where constants were in the global namespace. @Sunderlandkyl could you add a separate commit where you add `namespace {}` around these or move them into a private class (if there is one already)?

```
vtkMRMLMarkupsROINode.cxx
vtkMRMLSliceIntersectionInteractionRepresentation.cxx
vtkMRMLStreamingVolumeNode.cxx
vtkMRMLTransformDisplayNode.cxx
vtkMRMLTextNode.cxx
vtkImageGrowCutSegment.cxx
vtkSegmentation.cxx
vtkSlicerPlaneRepresentation3D.cxx
qSlicerSceneViewsModuleWidget.cxx
qSlicerSequencesModuleWidget.cxx
qMRMLExpandingWebView.cxx
qMRMLSequenceEditWidget.cxx
```

Fixes in these additional files are not needed for approval of this PR, so if you prefer no to spend time with it right now then it is fine, too.

Sunderlandkyl: Thanks @sedrakSofthenge and @lassoan, changed to use an enum.

lassoan (approved):
Thank you, it looks good!

PR #7630, opened 2024-03-13
ENH: Re-enable display of interaction handles in VR

Sunderlandkyl:
Improve and re-enable visibility of transform interaction widgets in VR.

Visualization of interaction widgets works for both absolute and relative widget size. Interacting with the widgets is not implemented yet, though the handles hand can be highlighted using the controllers.

Re #7570

## Tested with Acer WMR headset

- ### Adaptive scale:

https://github.com/Slicer/Slicer/assets/9222709/9e79d6ed-68df-4435-ad7f-9b1837177b79

- ### Absolute size:

https://github.com/Slicer/Slicer/assets/9222709/f9e757a5-2ff1-4483-9d80-149d8d73db20

PR #7675, opened 2024-04-03
BUG: Fix markups handle scale not updating

Sunderlandkyl:
Addresses issue introduced in 580e1875b1c0fa20d1db123d5feeb6487fca2c2d which caused markups handle scale to not update. The original fix was meant to prevent a crash when starting VR with Markups in the scene, however it disabled scale update in all situations. Fixed by checking GetNeverRendered() instead of GetGenericContext(), before calling GetScreenSize(). Once the window has been rendered at least once, it is safe to call GetScreenSize().

* https://github.com/Slicer/Slicer/issues/7570
* https://github.com/Slicer/Slicer/pull/7630

### DIFF
--- a/Modules/Loadable/Markups/VTKWidgets/vtkSlicerMarkupsWidgetRepresentation3D.cxx
+++ b/Modules/Loadable/Markups/VTKWidgets/vtkSlicerMarkupsWidgetRepresentation3D.cxx
@@ -1196,7 +1196,6 @@ double vtkSlicerMarkupsWidgetRepresentation3D::GetViewScaleFactorAtPosition(doub
   return viewScaleFactorMmPerPixel;
 }
 
-
 //----------------------------------------------------------------------
 void vtkSlicerMarkupsWidgetRepresentation3D::UpdateViewScaleFactor()
 {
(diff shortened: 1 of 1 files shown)


jcfr (approved):

Modules/Loadable/Markups/VTKWidgets/vtkSlicerMarkupsWidgetRepresentation3D.cxx
```diff
@@ -1206,9 +1205,10 @@ void vtkSlicerMarkupsWidgetRepresentation3D::UpdateViewScaleFactor()
   {
     return;
   }
-  if (!this->Renderer->GetRenderWindow()->GetGenericContext())
+
+  if (this->Renderer->GetRenderWindow()->GetNeverRendered())
```
jcfr, line 1209: @LucasGandel @sankhesh  The change implemented by @Sunderlandkyl but since the function `GetNeverRendered` is only used a few times in VTK itself, I was wondering if this was an anti-pattern of if you had a better idea for handling this.

sankhesh: IMO, it is fine to query the `NeverRendered` flag. There are a few widgets that use that flag too.

PR #7681, opened 2024-04-05
[5.6] BUG: Fix unrelated matrix elements overwritten in transform editor

jcfr:
Backport of https://github.com/Slicer/Slicer/pull/7594

------------

In transforms module, when linear transformation matrix values were edited using spinboxes and then by directly editing the matrix, the value that was previously edited in the spinbox got overwritten by the new value.

The issue was that a single slot was used for all signals coming from all sliders and in the slot implementation the active focus was used to determine which axis should be updated. Fixed by using a separate slot for each axis. This way we don't rely on the GUI widget focus, which may go through complex transitions during transform editing using various widgets.

(cherry picked from commit 325fdbd4c380c1c30f18ded37acdc6c1270e9f1a)


jcfr:
cc: @muratmaga

PR #7687, opened 2024-04-11
ENH: Add Python-wrappable methods for accessing anatomical regions

lassoan:
Terminology logic only offered access to anatomical regions and their modifiers using C++ API (due to CodeIdentifier type is not accessible in Python). This commit adds helper functions that allow developers to iterate through all anatomical regions and their modifiers.


cpinter:
LGTM, thank you! It is true that the API has been a bit rigid...

PR #7691, opened 2024-04-12
ENH: Improve terminology selector

lassoan:
Allow deselecting anatomic region and anatomic region modifier ("None" and "No region modifier" items are added to the comboboxes). Before this commit, once a region or modifier was selected, it was not possible to deselect it.

image

Add the anatomic region to the default generated name, because often it is the only difference between segment names (e.g., "bleeding in Cerebellum" and "bleeding in Subdural space").

image

To reduce occurrences of undefined (gray) color: if color is not specified for the selected type modifier then use the color of the non-modified type.

Added a few missing translation functions to allow full internationalization.


Modules/Loadable/Terminologies/Widgets/qSlicerTerminologyNavigatorWidget.cxx
```diff
@@ -967,11 +973,36 @@ QString qSlicerTerminologyNavigatorWidget::nameFromTerminology(vtkSlicerTerminol
   // Assemble from selection if there is no label
   if (name.isEmpty())
   {
-    name = entry->GetTypeObject()->GetCodeMeaning();
-
+    // Set type and modifier in name
     if (entry->GetTypeObject()->GetHasModifiers() && entry->GetTypeModifierObject() && entry->GetTypeModifierObject()->GetCodeValue())
     {
-      name += QString(", %1").arg(entry->GetTypeModifierObject()->GetCodeMeaning());
+      //: For formatting of terminology entry with a modifier. %1 is structure name (e.g., "Kidney"), %2 is modifier (e.g., "Left")
```
cpinter, line 979: Why the colon? `//:`

cpinter:
These changes look reasonable, thank you!

Modules/Loadable/Terminologies/Widgets/qSlicerTerminologyNavigatorWidget.cxx
lassoan: `//:` is the syntax for message to translators, shown in Weblate and Qt Linguist, for the message in the line below.

PR #7694, opened 2024-04-13
ENH: Update Ruff configuration checking for numpy2 deprecation issues

jamesobutler:
See https://docs.astral.sh/ruff/rules/numpy2-deprecation/

To prep for users potentially using Numpy 2, this runs a numpy2 deprecation check to avoid any incompatibilities. It appears that Slicer's numpy usage is already compatible, but this additional check makes sure that continues.

### DIFF
--- a/.ruff.toml
+++ b/.ruff.toml
@@ -26,6 +26,7 @@ select = [
 #  "SIM",         # flake8-simplify
   "UP",          # pyupgrade
   "YTT",         # flake8-2020
+  "NPY201",      # numpy2-deprecation
 ]
 extend-ignore = [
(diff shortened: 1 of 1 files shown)


.ruff.toml
```diff
@@ -26,6 +26,7 @@ select = [
 #  "SIM",         # flake8-simplify
   "UP",          # pyupgrade
   "YTT",         # flake8-2020
+  "NPY201",      # numpy2-deprecation
```
jcfr, line 29: Since we are already specifying `NPY` above, I think this rule is already enabled by default.

See https://docs.astral.sh/ruff/rules/#numpy-specific-rules-npy

PR #7695, opened 2024-04-14
ENH: Allow disabling histogram (2D views) interactions of ThresholdEffect

mauigna06:
Use case:
- Avoid users changing upper and lower limit threshold values by accidentally clicking on the 2D views

Example:

```
# Load sample data
# Enter segment editor module
# Create an empty segment
# Paste the code below to disable histogram interactions of Threshold effect:
segmentEditorWidget = mainWindow().findChildren('qMRMLSegmentEditorWidget')[0]
segmentEditorWidget.setActiveEffectByName("Threshold")
thresholdEffect = segmentEditorWidget.activeEffect()
thresholdEffect.self().enableViewInteractions = False
```

Thanks


lassoan (approved):
Thank you, the change looks reasonable. I've updated the comment and commit message a bit to explain the motivation for the change and the chosen implementation a bit more clearly.

mauigna06:
Ok, thanks :)

Issue #7661, opened 2024-03-25
Simple Filters do not work in all installations


dzenanz:
## Summary

Trying to run any filter from `Simple Filters` results in an error:
```
---------------------------
Exception before execution of AbsImageFilter
---------------------------
<class 'RuntimeError'>
---------------------------
OK   
---------------------------
```
content in the log:

```log
myFilter = AbsImageFilter()
myFilter.SetDebug(False)
myFilter.SetNumberOfThreads(24)
myFilter.SetNumberOfWorkUnits(0)
Error ImageIO factory did not return an ImageIOBase: MRMLIDImageIO
```
## Steps to reproduce

1. Download `MRHead` sample data
2. Switch to `Simple Filters` module
3. Click `Apply`
4. The above error appears

## Environment
- Error appears for me in `Slicer 5.6.1 r32438 / 117ce5f` on Windows 11 (23H2 build 22631.3296) fully updated
- Works as expected in versions `5.2.1 r31317 / 77da381` and `5.7.0-2024-03-06 r32752 / 024b340` on the same Windows machine
- Works as expected in `Slicer 5.6.1 r32438 / 117ce5f` on Ubuntu 22.04.4 LTS.
- Error appears in `Slicer 5.6.1 r32438 / 117ce5f` on Ubuntu for @andinet.

lassoan:
This error is typically due to that you (or any of thr extensions ypu installed) installed SimpleITK from PyPI and that overwrote Slicer's SimpleITK. SimpleITK installed from PyPI does not include the in-memory transfer image IO.

We have been avoiding reinstaing SimpleITK and if there is a package then insists on replacing it (e.g., nnunetv2) then we rewrite package dependencies on-the-fly (see TotalSegmentator Slicer extension).

Would it be sufficient to check the SimpleITK version ar startup and display a descriptive error message if it is not correct, to make extension developers aware that they broke some features Slicer?

dzenanz:
> check the SimpleITK version ar startup and display a descriptive error message

I think that would be sufficient. The possible culprit (in this instance) is https://github.com/bowang-lab/MedSAMSlicer.

lassoan:
Fixed by https://github.com/Slicer/Slicer/pull/7699

jcfr:
For future reference, the linked pull request was closed and corresponding changes were directly integrated to `main` as commits 03ff788 and 29f6b91

Issue #6711, opened 2022-11-30
Slicer's embedded SimpleITK can be removed by pip


lassoan:
## Summary

When installing a package that requires a specific SimpleITK version (e.g., TotalSegmentator requires SimpleITK==2.0.2) then installation of that package may replace Slicer's SimpleITK.

The new SimpleITK package mostly works, but the ITK IO plugins that Slicer provides (e.g., for in-memory file transfer between Slicer and ITK) don't work anymore.

## Steps to reproduce

Run `pip_install('TotalSegmentator')`.
Slicer's SimpleITK is uninstalled and SimpleITK-2.0.2 is installed.

## Expected behavior

Slicer's SimpleITK should not be uninstalled.

## Environment
- Slicer version: Slicer-5.2.1
- Operating system: Windows 11

lassoan:
The solution could be to freeze SimpleITK version, or specify a constraints file (`-c` argument) in `slicer.util.pip_install` (and make sure all extensions are installed via this `pip_install`).

Locking SimpleITK version would not completely solve cases where a package requires a specific SimpleITK version, because the SimpleITK incompatibility with the requirements would be detected and the package would not run. The following [workaround was implemented in Slicer's TotalSegmentator extension](https://github.com/lassoan/SlicerTotalSegmentator/commit/927ef8fe10cc0a21683684253a0d8577922ad05f):
- Install TotalSegmentator without dependencies
- Modify METADATA file of TotalSegmentator (remove SimpleITK from dependencies)
- Install TotalSegmentator normally (with dependencies.

Probably it would be better to avoid hard requirements by asking third-party package maintainers. We [requested removal of the hard SimpleITK version requirement in TotalSegmentator ](https://github.com/wasserth/TotalSegmentator/issues/32).

jamesobutler:
The issued caused by changing the SimpleITK version in Slicer was originally reported at https://discourse.slicer.org/t/slicer-5-3-preview-warnings-and-error-messages/26457/9.

Another user recently ran into this issue as discussed in https://discourse.slicer.org/t/did-sitkutils-pullvolumefromslicer-break-with-slicer-5-2/26863/6

jcfr:
Related discussion:
* https://discourse.slicer.org/t/proposal-install-simpleitk-from-wheels-instead-of-building-from-source/25635

lassoan:
[...]

PR #7699, opened 2024-04-17
ENH: Use regular file IO in SimpleITK utils if MRMLIDImageIO unavailable

lassoan:
When extensions pip-install SimpleITK and it replaces SimpleITK that is bundled with Slicer, then MRMLIDImageIO ITK IO plugin is no longer available for passing data between SimpleITK and Slicer. As a result, Simple Filters module and most Slicer modules that use SimpleITK no longer work.

This commit makes image passing between SimpleITK and Slicer more robust: if MRMLIDImageIO is not available then images are written to regular files.

see #7661

### DIFF
--- a/Base/Python/sitkUtils.py
+++ b/Base/Python/sitkUtils.py
@@ -2,7 +2,7 @@ import SimpleITK as sitk
 
 import slicer
 
-__sitk__MRMLIDImageIO_Registered__ = False
+__sitk__MRMLIDImageIO_Missing_Reported__ = False
 
 
 def PushVolumeToSlicer(sitkimage, targetNode=None, name=None, className="vtkMRMLScalarVolumeNode"):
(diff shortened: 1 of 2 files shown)


lassoan:
@jcfr @pieper please have a look. This commit may solve (maybe all?) issues with extensions replacing Slicer's SimpleITK.

pieper (approved):
Makes sense to me 👍 

Have we tested the actual performance difference?  If the difference is small we might want to retire the ImageIO method - it seems to have been a consistent maintenance issue.

lassoan:
Thanks for the review. The performance difference can be perceivable for very large volumes (several GB) or real-time updates (dozens of updates per second). If we implement mechanisms like this (fall back to lower-performance but more robust method automatically) then it may worth it to keep MRMLIDImageIO around for now.

jcfr:
Related issues and discussions:
* https://github.com/Slicer/Slicer/issues/6711
* https://github.com/Slicer/Slicer/issues/7661


> Merged as commits 03ff788 and 29f6b91

Let's make sure to integrate changes with pull request, it helps with generating release notes. 

Nitpick: I would even suggest to force push `main` and re-issue a pull request.

PR #7700, opened 2024-04-18
ENH: Log more information when NRRD reading fails

lassoan:
When some required fields in the NRRD file were missing, the user just got a generic error message. This commit logs detailed information about why Teem library could not load the NRRD file.

see https://discourse.slicer.org/t/slicer-rejects-modified-header/35549


pieper (approved):
Makes sense to me. 👍

PR #7714, opened 2024-04-26
ENH: Improve terminology selector speed

lassoan:
When the type selector was populated, all the type IDs were retrieved first, and then the type objects were looked up based on these IDs. This lookup is very slow if there are tens of thousands of types in the terminology.

Improved performance by returning not just all the type IDs but also all the corresponding type objects, therefore avoiding the costly lookup for each type.

see https://discourse.slicer.org/t/exporting-a-segmentation-paired-to-a-custom-colortable/35719/15


cpinter:
Sorry I was away. Looks good @lassoan thank you!

PR #7715, opened 2024-04-26
ENH: Make qMRMLSequenceBrowserPlayWidget shortcuts accessible from python

jamesobutler:
This makes the shortcuts for the qMRMLSequenceBrowserPlayWidget able to be customized from python. I have used this when showing multiple qMRMLSequenceBrowserPlayWidget objects and needing unique shortcuts for each.

### DIFF
--- a/Modules/Loadable/Sequences/Widgets/qMRMLSequenceBrowserPlayWidget.cxx
+++ b/Modules/Loadable/Sequences/Widgets/qMRMLSequenceBrowserPlayWidget.cxx
@@ -37,6 +37,9 @@ protected:
 public:
   qMRMLSequenceBrowserPlayWidgetPrivate(qMRMLSequenceBrowserPlayWidget& object);
   bool RecordingControlsVisible{ true };
+  QString PlayPauseShortcut{ "Ctrl+Shift+Down" };
+  QString PreviousFrameShortcut{ "Ctrl+Shift+Left" };
+  QString NextFrameShortcut{ "Ctrl+Shift+Right" };
   void init();
 
   vtkWeakPointer<vtkMRMLSequenceBrowserNode> SequenceBrowserNode;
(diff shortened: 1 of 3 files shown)


Modules/Loadable/Sequences/Widgets/qMRMLSequenceBrowserPlayWidget.cxx
```diff
@@ -307,23 +307,23 @@ void qMRMLSequenceBrowserPlayWidget::setPlayPauseShortcut(QString keySequence)
 {
   Q_D(qMRMLSequenceBrowserPlayWidget);
   QObject::connect(new QShortcut(QKeySequence(keySequence), this), SIGNAL(activated()), SLOT(onVcrPlayPause()));
-  d->pushButton_VcrPlayPause->setToolTip(d->pushButton_VcrPlayPause->toolTip()+" ("+keySequence+")");
+  d->pushButton_VcrPlayPause->setToolTip("Play/Pause (" + keySequence + ")");
```
lassoan, line 310: Messages that are displayable to users cannot be concatenated (because order of words is different in each language), so always placeholders need to be used. Something like this:

```suggestion
  d->pushButton_VcrPlayPause->setToolTip(tr("Play/Pause (%1)").arg(keySequence));
```

Please update everywhere else.

lassoan (commented):
Thanks for the update, it looks good overall, but the message formatting needs to be adjusted a bit and there is also a build error.

Modules/Loadable/Sequences/Widgets/qMRMLSequenceBrowserPlayWidget.cxx
jamesobutler: I've pushed the according updates to have the translation support.

lassoan (commented):
I'm sorry, I've just realized that these are string setters. We should use Qt properties for these instead of adding Q_INVOKABLE. Since switching from Q_INVOKABLE to Q_PROPERTY changes the Python API, it would be very hard to change it later, so we need to switch to use Q_PROPERTY in this pull request.

jamesobutler:
@lassoan I have made updates to use Q_PROPERTY similar to the existing `RecordingControlsVisible` Q_PROPERTY for the qMRMLSequenceBrowserPlayWidget.

lassoan (approved):
This looks perfect, thank you!

PR #7724, opened 2024-05-03
COMP: Fix main project source dir colliding with built-in extension of same name

jamesobutler:
Here is my situation of names utilized for a Slicer custom app, the custom app's application name and the name of an extension packaged into the custom app:

* `MyNameApp` (Slicer custom app) (github.com/MyOrg/MyNameApp)
* `MyName.exe` (MyNameApp's application name)
* `MyName` (extension name included in custom app) (github.com/MyOrg/MyName)

The following could collide:
* `${Slicer_MAIN_PROJECT_APPLICATION_NAME}_SOURCE_DIR`
* `${${extension_name}_SOURCE_DIR}`

The issue was discovered when MainApplicationRepositoryRevision and MainApplicationRepositoryURL reported info for the extension of same name instead of the application.


CMake/SlicerVersion.cmake
```diff
@@ -25,7 +25,7 @@ set(expected_defined_vars
   GIT_EXECUTABLE
   Slicer_CMAKE_DIR
   Slicer_MAIN_PROJECT_APPLICATION_NAME
-  ${Slicer_MAIN_PROJECT_APPLICATION_NAME}_SOURCE_DIR
+  Slicer_MAIN_PROJECT_SOURCE_DIR
```
jcfr, line 28: Update comment above:

```diff
- <Slicer_MAIN_PROJECT_APPLICATION_NAME>_SOURCE_DIR
+ Slicer_MAIN_PROJECT_SOURCE_DIR
```

jcfr (changes requested):
Considering also updating https://github.com/Slicer/Slicer/blob/main/CMake/SlicerConfigureVersionHeaderTarget.cmake

CMake/SlicerVersion.cmake
jamesobutler: Comment strings have been updated

jcfr (approved):

PR #7735, opened 2024-05-09
ENH: Add API for finding terminology from codes

lassoan:
Often the segment category&type and anatomic region codes are available in multiple contexts. This commit adds FindTerminologyNames and FindAnatomicContextNames methods to the terminology logic that allows getting a list of contexts where the codes are listed in.

The methods can also return the terminology objects, which makes it very easy to get properties (such as name and color) from codes.


cpinter (approved):
LGTM!

PR #7736, opened 2024-05-09
BUG: Fix crash when deleting multiple markups in subject hierarchy

Sunderlandkyl:
Add nullptr check to vtkSlicerMarkupsInteractionWidgetRepresentation::UpdateHandleToWorldTransform.

See discussion: https://discourse.slicer.org/t/reproducible-slicer-crash-deleting-markupsroinodes-from-subject-hierarchy/36007


pieper (approved):
LGTM 👍

PR #7738, opened 2024-05-10
Improve segmentation (.seg.nrrd) file reader robustness

lassoan:
Make the segmentation storage node read .seg.nrrd files that are created outside of Slicer (e.g., using slicerio Python package).
They may use different order of fields (therefore the file detection heuristics had to be tuned), some fields may be missing, and there could be minor syntax differences (e.g., Slicer always terminates a list with separator character, but other software may just use separators between components).

### DIFF
--- a/Libs/MRML/Core/vtkMRMLSegmentationStorageNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLSegmentationStorageNode.cxx
@@ -74,7 +74,7 @@
 #include <sstream>
 
 //----------------------------------------------------------------------------
-static const std::string SERIALIZATION_SEPARATOR = "|";
+static const char LIST_SEPARATOR = '|';
 static const std::string KEY_SEGMENT_ID = "ID";
 static const std::string KEY_SEGMENT_NAME = "Name";
 static const std::string KEY_SEGMENT_COLOR = "Color";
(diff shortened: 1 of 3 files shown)


Libs/MRML/Core/vtkMRMLSegmentationStorageNode.cxx
```diff
@@ -862,35 +862,42 @@ int vtkMRMLSegmentationStorageNode::ReadBinaryLabelmapRepresentation(vtkMRMLSegm
...
-          {
-            vtkWarningToMessageCollectionMacro(this->GetUserMessages(), "vtkMRMLSegmentationStorageNode::ReadBinaryLabelmapRepresentation",
-              "Segment extent is missing for segment " << segmentIndex);
-            for (int i = 0; i < 6; i++)
+            int currentSegmentExtent[6] = { 0, -1, 0, -1, 0, -1 };
+            std::string currentExtentString;
+            if (this->GetSegmentMetaDataFromDicitionary(currentExtentString, dictionary, segmentIndexForExtent, KEY_SEGMENT_EXTENT))
             {
-              currentSegmentExtent[i] = imageExtentInFile[i];
+              GetImageExtentFromString(currentSegmentExtent, currentExtentString);
```
Sunderlandkyl, line 873: If the extent string is invalid, then the extent of the current segment will be skipped.
Should we instead switch to the imageExtentInFile if we encounter an invalid extent string?

lassoan: By invalid do you mean empty (something like [0, -1, 0, -1, 0, -1])? That would mean that the segment is empty and so it should not expand the extent.

Sunderlandkyl: No, I meant that if the string itself was invalid, and could not be correctly parsed.

lassoan: Indeed all these parsing codes are still not fully robust for invalid input. The changes in this commit improve things, so I would integrate it as is, and implement further improvements later if needed.

PR #7739, opened 2024-05-10
BUG: Update vtkMRMLNode.cxx remove unused variable

maddin200:


lassoan:
Thank you! Could you please fix the errors detected by the automatic checks: trailing whitespace in the code, add `STYLE: ` prefix to the commit message. Note that this is not bug just a code style change, because the modification does not cause any functional change in the software (the line is automatically removed by the optimizing compiler).

PR #7740, opened 2024-05-10
ENH: Add API to change screen scale factor

lassoan:
Size of markup control points glyphs and text can be adjusted globally for a view by modifying the view node's ScreenScaleFactor property. This is useful for compensating different physical screen sizes and resolutions or for taking high-resolution screenshots. This scaling factor has already existed but it was hardcoded and now it is exposed in the view node.

Example use:

    getNode('vtkMRMLViewNode1').SetScreenScaleFactor(0.1)
    getNode('vtkMRMLSliceNodeRed').SetScreenScaleFactor(0.1)

Most likely users would not want to save this factor in the scene but they would set it for their computer, therefore the value is not written into the scene file.


Sunderlandkyl (approved):
LGTM!

Issue #7742, opened 2024-05-12
Incorrect geometry after running Orient Scalar Volume 


pieper:
## Summary

Running the CLI module cases the result volume to be in the wrong place.

## Steps to reproduce

1) Download Sample data (MRHead)
2) Run OrientScalarVolume with default LPS output into new volume
3) Put put one volume in foreground and one in background and see the shift

## Environment
- Slicer version: Slicer-5.7.0-2024-05-10, does not happen in 5.6.1
- Operating system: tested on mac

pieper:
Reported and discussed here: https://discourse.slicer.org/t/geometry-all-messed-up-in-latest-5-7-0-pre-release/36072

PR #7627, opened 2024-03-13
ENH: Improve vtkMRMLVolumeArchetypeStorageNode for left handed volumes

Thibault-Pelletier:
* Automatically flip left handed volumes when loading into 3D Slicer to make them compatible with all algorithms. Fixes flipped normals in segmentation node for left handed volumes.

See related discussion here : 
https://discourse.slicer.org/t/slicer-segmentation-inverted-normals-for-closed-surface-representation/34762

PR #7746, opened 2024-05-13
Fix reading of images that have left-handed IJK coordinate system

lassoan:
Fix regression introduced in https://github.com/Slicer/Slicer/pull/7627 that image origin is not computed correctly when flipping K axis of an image to make it right-handed.

Fixes issue discussed in https://discourse.slicer.org/t/geometry-all-messed-up-in-latest-5-7-0-pre-release/36072/12


pieper (approved):
Thanks for the quick fix 👍 

I build and tested this and confirmed it fixed the orient scalar volume issue.

pieper:
Fixes #7742

PR #7750, opened 2024-05-17
BUG: Fix warnings when vtkMRMLModelNode has no mesh data

Sunderlandkyl:
Fixes warnings during rendering such as:

> Algorithm vtkPassThrough (00000223357FFBC0) returned failure for request: vtkInformation (000002233CEA6520)

Fixed by enabling AllowNullInput on the vtkPassThrough in vtkMRMLModelDisplayNode.

This commit also fixes an issue where the views would not be immediately rendered when vtkMRMLModelNode::SetAndObserveMesh(nullptr) was called for a model node that previously had a mesh. Fixed by requesting a render in vtkMRMLModelDisplayableManager whenever an actor that was previously visible is no longer displayable.

### DIFF
--- a/Libs/MRML/DisplayableManager/vtkMRMLModelDisplayableManager.cxx
+++ b/Libs/MRML/DisplayableManager/vtkMRMLModelDisplayableManager.cxx
@@ -785,8 +785,9 @@ bool vtkMRMLModelDisplayableManager::IsModelDisplayable(vtkMRMLDisplayableNode* 
   {
     displayable |= this->IsModelDisplayable(node->GetNthDisplayNode(i));
     if (displayable)
-    {// Optimization: no need to search any further.
-      continue;
+    {
+      // Optimization: no need to search any further.
+      break;
     }
   }
   return displayable;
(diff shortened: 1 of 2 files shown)


Libs/MRML/DisplayableManager/vtkMRMLModelDisplayableManager.cxx
```diff
@@ -818,13 +818,36 @@ bool vtkMRMLModelDisplayableManager::OnMRMLDisplayableModelNodeModifiedEvent(
     return false;
   }
 
+  int ndnodes = modelNode->GetNumberOfDisplayNodes();
   if (!this->IsModelDisplayable(modelNode))
   {
-    return false;
+    bool requestRender = false;
```
lassoan, line 824: The for loop after this looks very similar the the other one below. Could they be merged? If it makes things slower or more complex if we merge them then we should have a comment that explains why it is better to have two loops.

lassoan (commented):
Thank you, it looks good, I just added one comment inline.

Libs/MRML/DisplayableManager/vtkMRMLModelDisplayableManager.cxx
Sunderlandkyl: Good point. I've integrated the loops so it is much cleaner now.

lassoan (dismissed):
Looks great, thank you!

Libs/MRML/DisplayableManager/vtkMRMLModelDisplayableManager.cxx
```diff
@@ -785,8 +785,9 @@ bool vtkMRMLModelDisplayableManager::IsModelDisplayable(vtkMRMLDisplayableNode*
   {
     displayable |= this->IsModelDisplayable(node->GetNthDisplayNode(i));
     if (displayable)
-    {// Optimization: no need to search any further.
-      continue;
+    {
+      // Optimization: no need to search any further.
+      break;
```
Sunderlandkyl, line 790: Fixed an issue that I just noticed in IsModelDisplayable. Should be break not continue.

lassoan (approved):
Good catch. All looks good.

PR #7755, opened 2024-05-21
COMP: Fix version extraction in pre-commit-autoupdate workflow

jcfr:
Add support for versions specified without the "v" prefix.

This fixes the description generated in pull request like the following:
* https://github.com/Slicer/Slicer/pull/7754


jamesobutler (approved):

jcfr:
Thanks for the review, I will re-trigger the relevant workflow once this pull request has been integrated.

PR #7766, opened 2024-05-28
DOC: Match updated interface in SlicerSurfaceToolbox Dynamic Modeler

Leengit:
Make the documentation match the updated interface and add an explanation that the boolean operations combine the negative sides of multiple planes to make the negative side of the output.  See https://github.com/Slicer/SlicerSurfaceToolbox/pull/62.

### DIFF
--- a/Docs/user_guide/modules/dynamicmodeler.md
+++ b/Docs/user_guide/modules/dynamicmodeler.md
@@ -23,7 +23,7 @@ Click on the Apply button to run the tool once, or check the checkbox on the App
 ## Tools
 
 ### Plane cut
-Cut a plane into two separate meshes using any number of markup planes or slice views. The planes can be combined using union, intersection and difference boolean operation.
+Extracts a region of the surface by cutting it with one or more markup planes or slice view planes. The `Output model` contains what is on the negative side of the plane(s). If more than one plane is specified, the negative sides of the planes are combined using a union, intersection, or difference operation. The part of the `input model` that is not included in the `output model` is stored in the `complement output model`.
 #### Input nodes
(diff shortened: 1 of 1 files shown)


Docs/user_guide/modules/dynamicmodeler.md
```diff
@@ -23,16 +23,16 @@ Click on the Apply button to run the tool once, or check the checkbox on the App
 ## Tools
 
 ### Plane cut
-Cut a plane into two separate meshes using any number of markup planes or slice views. The planes can be combined using union, intersection and difference boolean operation.
+Extracts a region of the surface by cutting it with one or markup planes or slice views. If more than one plane is specified, the negative sides of the planes are combined using a union, intersection, or difference boolean operation to give the negative side of the output. The positive side of the output is the complementary region.
```
lassoan, line 26:
```suggestion
Extracts a region of the surface by cutting it with one or more markup planes or slice view planes. The `Output model` contains what is on the negative side of the plane(s). If more than one plane is specified, the negative sides of the planes are combined using a union, intersection, or difference operation. The part that of the `input model` that is not included in the `output model` is stored in the `complement output model`.
```

lassoan (commented):
I've made some adjustments to the "Extracts a region..." section. We also need to update the parameter descriptions in the source code. Thank you!

Leengit:
I have incorporated all the suggestions.

Leengit:
Now that https://github.com/Slicer/SlicerSurfaceToolbox/pull/62 is merged, I believe this is ready for approval and merging.

lassoan (approved):
Thank you!

PR #7772, opened 2024-05-31
BUG: Fix reading of volume property (.vp) file

lassoan:
Color transfer function line in .vp files can be several thousands character long when there are many control points but the reader had a hardcoded limit at 1024 bytes. Fixed by switching to use an API that does not limit line length.

As a preventive action, all other line parsing where a hardcoded line length limit was set is replaced.

Fixes the issue reported at https://discourse.slicer.org/t/saving-volume-properties/36475/12


pieper (approved):
LGTM 👍 

thanks for fixing these!

PR #7788, opened 2024-06-06
ENH: add function call logger

BerDom-Ing:
This is a decorator that must be specified with @slicer.util.log_invocation above the function that you want to be logged


lassoan:
Closing this pull request in favor of https://github.com/Slicer/Slicer/pull/7855.

@BerDom-Ing please review #7855. Contents of this pull request can go to the Telemetry extension. It may make sense to store usage events in an SQLite database (just a table containing `timestamp`, `component`, and `event` columns would suffice), because multiple Slicer instances can be running, and each of them may want to log events at the same time and we don't want to spend too much time with logging. If that seems too complicated then maybe use a CSV file format because that allows simply appending lines to the end of the file (while in a JSON file the events would need to be inserted inside a list).

PR #7798, opened 2024-06-12
DOC: Update Linux documentation

RafaelPalomar:
I have tested the Slicer-5.6.2 binaries in different containers and updated the Linux instructions (mostly adding new packages needed to run the application). This was suggested by @lassoan in [https://discourse.slicer.org/t/cant-start-latest-stable-on-ubuntu-20-04/14029/51](https://discourse.slicer.org/t/cant-start-latest-stable-on-ubuntu-20-04/14029/51)


lassoan (approved):
Perfect, thank you very much. It would be useful if you described somewhere how you did it, so that others can do this, too (or even for you for next time you need it).

PR #7810, opened 2024-06-20
Revert SimpleITK to previous known working hash

blowekamp:
The update to ITK and SimpleITK is causing an unknown loading issue on windows.

This successfully compile on a Ubutu 20.04 linux system.

### DIFF
--- a/SuperBuild/External_SimpleITK.cmake
+++ b/SuperBuild/External_SimpleITK.cmake
@@ -140,7 +140,7 @@ ExternalProject_Execute(${proj} \"install\" \"${PYTHON_EXECUTABLE}\" \"-m\" \"pi
 
   ExternalProject_SetIfNotDefined(
     Slicer_${proj}_GIT_TAG
-    "1236005d4f501aa87457a2b2d8fb07790169dfd5"  # slicer-v2.3.1-2024-05-20-bc4449e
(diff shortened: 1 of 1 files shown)


blowekamp:
@dzenanz @jcfr

jcfr:
I am now looking into the root cause, I will follow-up later today with either a fix or integration of this pull request.

blowekamp:
@jcfr I also have some changes to the SWIG files to accommodate the "slicer_itk" namespace. I didn't think it's related to the issue.

jamesobutler:
@blowekamp Does SimpleITK need newer SWIG (4.2.1) and PCRE (10.40) versions for certain build configurations? I didn't have any problem with latest SimpleITK when I built Slicer locally on Windows (see [here](https://discourse.slicer.org/t/import-simpleitk-fails/36735/3?u=jamesobutler)) which includes these older versions of dependencies, but maybe some configuration doesn't do well with the older dependencies. Only the Slicer factory Windows build machine had problems.
https://github.com/Slicer/Slicer/blob/84b4b105088288ccfb295a7656de3be0bf1efd12/SuperBuild/External_Swig.cmake#L15-L17
https://github.com/Slicer/Slicer/blob/84b4b105088288ccfb295a7656de3be0bf1efd12/SuperBuild/External_PCRE.cmake#L50-L55

PR #7813, opened 2024-06-20
ENH: Only translate python strings when necessary

jamesobutler:
This has performance implications where runtime is faster if English is the desired language for the application because the source code is in English.

## Testing Methods

Segmentation file to extract and put into User Downloads location (or modify script below)
[Segmentation.seg.zip](https://github.com/user-attachments/files/15920920/Segmentation.seg.zip)
```python
import os
import cProfile
cp=cProfile.Profile()
cp.enable()

# Load sample volume
import SampleData
mrHead = sampleDataLogic.downloadMRHead()
slicer.util.loadSegmentation(os.path.join(os.path.expanduser('~'), 'downloads', 'Segmentation.seg.nrrd'))

# Copied from https://slicer.readthedocs.io/en/latest/developer_guide/script_repository.html#get-volume-of-each-segment
segmentationNode = getNode("Segmentation")

# Compute segment statistics
import SegmentStatistics
segStatLogic = SegmentStatistics.SegmentStatisticsLogic()
segStatLogic.getParameterNode().SetParameter("Segmentation", segmentationNode.GetID())
segStatLogic.computeStatistics()
stats = segStatLogic.getStatistics()

# Display volume of each segment
for segmentId in stats["SegmentIDs"]:
  volume_cm3 = stats[segmentId,"LabelmapSegmentStatisticsPlugin.volume_cm3"]
  segmentName = segmentationNode.GetSegmentation().GetSegment(segmentId).GetName()
  print(f"{segmentName} volume = {volume_cm3} cm3")

cp.disable()
cp.dump_stats(os.path.join(os.path.expanduser('~'), 'downloads', 'profile.prof'))
```

## Results

With the changes in this PR, the above script in the testing method is overwhelming quicker going from several seconds to nearly instantaneous.

Using [`snakeviz`](https://jiffyclub.github.io/snakeviz/) for display of the `cProfile` results:
```PS
python -m snakeviz "C:\Users\butlej30383\Downloads\profile.prof"
```

| Slicer 5.7.0-2024-06-18 | This PR |
|---------------------------|---------|
|image|image|

### DIFF
--- a/Base/Python/slicer/i18n.py
+++ b/Base/Python/slicer/i18n.py
@@ -3,6 +3,9 @@ import logging
 import os
 import re
 
+# Cache that stores context name computed from file paths to make translations faster
+FILEPATH_TO_CONTEXT = {}
+
 def translate(context, text):
     """Translate message to the current application language.
     This `translate(context, message)` function is recognized by Qt's lupdate tool.
@@ -25,6 +28,11 @@ def getContext(sourceFile):
     Most helper Python scripts in Slicer are Python packages (subfolders containing addition Python scripts and an `__init__.py` file)
     and their name is constructed as PythonPackageName.SourceFileName (for example, `SegmentEditorEffects.SegmentEditorDrawEffect`).
     """
+
+    # Cache context name to make translations faster
+    if sourceFile in FILEPATH_TO_CONTEXT:
+        return FILEPATH_TO_CONTEXT[sourceFile]
+
     if os.path.isfile(sourceFile):
         parentFolder = os.path.dirname(sourceFile)
         init_file_path = parentFolder + os.path.sep + "__init__.py"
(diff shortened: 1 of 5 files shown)


lassoan:
Thank you, this was a very interesting and useful finding!

Having a fast path for the special case for en-US makes sense, but we can do better than that, because:
1. I18N should not cause any overhead in any languages
2. Translation is used for the source language as well (for example, it can use proper singular/plural depending on number of items)
3. QSettings() is an expensive operation

Instead, I'll send a pull request that optimizes getting the context name from the stack (the current one walked up the entire stack and extracted a whole lot of information that we just discarded), optimize computation of context from filename, and fix the inefficiency in segment statistics logic (each time the logic was instantiated, it translated all string 18x).

Computation times:
- original: 3950ms
- your suggestion: 57ms
- with the pull request that I'll push soon: 11ms

lassoan:
@jamesobutler the proposed changes have been pushed, please test!

Also thanks a lot for the profiling instructions, I haven't used cProfile+snakeviz, they work very well.

jamesobutler:
@lassoan Thanks for your hard work to improve this for the other languages as well! I have tried my original script and can confirm things are still fast.  I did not think about translation being able to handle singular vs plural states even for English. Feel free to provide the approving review to proceed with the integration of this PR.

lassoan (approved):
Thank you @jamesobutler

PR #7816, opened 2024-06-21
BUG: Mark translatable strings in Plots module

MohamedAllali:


lassoan:
All these changes have been already integrated.

PR #7822, opened 2024-06-24
ENH: Allow setting ambient shadows intensity

lassoan:
Shadows intensity can be linearly adjusted by shift and scale sliders in the 3D view controller.

This can be used to make ambient shadows appear darker (by increasing scale), without making the entire 3D scene darker (by increasing shift).

Example: see shadow of the rib casted on the kidney behind it. Left- no shadows. Center - default shadows (100%/0%). Right - strong shadows (200%/20%):

image

### DIFF
--- a/Base/QTGUI/Resources/UI/qSlicerSettingsViewsPanel.ui
+++ b/Base/QTGUI/Resources/UI/qSlicerSettingsViewsPanel.ui
@@ -357,6 +357,70 @@
             </property>
            </widget>
           </item>
+          <item row="10" column="0">
+           <widget class="QLabel" name="ThreeDAmbientShadowsIntensityScaleLabel">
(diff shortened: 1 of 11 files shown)


pieper (approved):
LGTM, but there there are some build errors to resolve before merging.

lassoan:
The build errors are due to that the testing is done on yesterday's docker image, which does not include the [VTK update](https://github.com/Slicer/Slicer/commit/d7eb84f7b71657aa9d4ccc920fd8e1168b6d1bf6) that was merged today.

It is not urgent, so we can wait for the new docker image tomorrow.

pieper:
Since the build error a false positive I'll go ahead and merge so it'll be testable tomorrow.

PR #7829, opened 2024-06-28
ENH: Make SegmentsTableView functions available from Python

cpinter:
This is needed to use the SegmentationVerification extension


cpinter:
See PW project https://projectweek.na-mic.org/PW41_2024_MIT/Projects/SegmentationVerificationModuleForFinalizingMultiLabelAiSegmentations/

Modules/Loadable/Segmentations/Widgets/qMRMLSegmentsTableView.h
```diff
@@ -76,7 +76,7 @@ class Q_SLICER_MODULE_SEGMENTATIONS_WIDGETS_EXPORT qMRMLSegmentsTableView : publ
   Q_INVOKABLE QTableView* tableWidget();
 
   /// Return number of segments (rows) in the table
-  int segmentCount() const;
+  Q_INVOKABLE int segmentCount() const;
```
lassoan, line 79: Shouldn't we make it a property instead?

lassoan (commented):
Thanks, it looks good, just have a look at the question I added inline.

cpinter:
I updated the commit with making `segmentCount` a property.

lassoan (approved):
Looks good, thank you. The CI failure is due to using an older Slicer docker image (up-to-date at the time of the PR was created) with the latest Slicer source code.

PR #7832, opened 2024-07-02
ENH: Hide angle markups when the line doesn't intersect with the slice

Sunderlandkyl:
For line/curve markup types, the lines are hidden whenever they don't intersect with the current slice, however the current behavior for angle markups is that they continue to be visible even when they no longer intersect. This commit updates the behavior of angle lines to be consistent with that of the other markup types.

See discussion: https://discourse.slicer.org/t/visibility-of-markups-angle-in-2d-slice-views-is-unlike-other-markups/37091

### DIFF
--- a/Modules/Loadable/Markups/VTKWidgets/vtkSlicerAngleRepresentation2D.cxx
+++ b/Modules/Loadable/Markups/VTKWidgets/vtkSlicerAngleRepresentation2D.cxx
@@ -238,6 +238,14 @@ void vtkSlicerAngleRepresentation2D::UpdateFromMRMLInternal(vtkMRMLNode* caller,
   this->LineActor->SetVisibility(numberOfDefinedControlPoints >= 2);
   this->ArcActor->SetVisibility(numberOfDefinedControlPoints == 3);
 
+  // Hide the actor if it doesn't intersect the current slice
+  this->LineSliceDistance->Update();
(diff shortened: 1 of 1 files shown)


jamesobutler (approved):
Thanks! I have confirmed with local testing that this change works as expected and is in line with the same type code for the Line and Curve representations in 2D. See https://github.com/Slicer/Slicer/commit/dfefe148939b00d51fa0c4f127c3be303d600cf0. 
https://github.com/Slicer/Slicer/blob/7cfa153a80397d54b9387ba00d0dbdd938386e9c/Modules/Loadable/Markups/VTKWidgets/vtkSlicerLineRepresentation2D.cxx#L104-L109
https://github.com/Slicer/Slicer/blob/7cfa153a80397d54b9387ba00d0dbdd938386e9c/Modules/Loadable/Markups/VTKWidgets/vtkSlicerCurveRepresentation2D.cxx#L109-L114

This PR:

https://github.com/Slicer/Slicer/assets/15837524/0a5769db-c72c-4032-96a1-e5bbdf604b21

jamesobutler:
@lassoan CI failure is due to an older Slicer docker image being used. Is the docker image not getting update appropriately?

jamesobutler:
Could one of the admins help integrate to bypass the status check of the Build check since it is not using an up-to-date docker image?

jcfr:
After fixing the upload of the `slicer/slicer-base`[^1] docker image used for testing, I just re-triggered the build.

Related issue:
* https://github.com/Slicer/Slicer/issues/7835

[^1]: https://hub.docker.com/r/slicer/slicer-base/tags?page=&page_size=&ordering=&name=latest

PR #7833, opened 2024-07-02
COMP: Update acceptable VS2022 MSVC version for TBB install

cpinter:
See https://discourse.slicer.org/t/tbb-dll-problem-with-new-version-of-visual-studio/36936

### DIFF
--- a/CMake/SlicerBlockPlatformCheck.cmake
+++ b/CMake/SlicerBlockPlatformCheck.cmake
@@ -38,7 +38,7 @@ if(Slicer_PLATFORM_CHECK)
     # and https://en.wikipedia.org/wiki/Microsoft_Visual_Studio#Version_history
     # 1910-1919 = VS 15.0 (v141 toolset) 1914 = VS 15.7
     # 1920-1929 = VS 16.0 (v142 toolset)
-    # 1930-1939 = VS 17.0 (v143 toolset)
+    # 1930-1949 = VS 17.0 (v143 toolset, see https://devblogs.microsoft.com/cppblog/msvc-toolset-minor-version-number-14-40-in-vs-2022-v17-10/)
     # VS 15.7 was announced to officially conform with the C++ standard of C++11, C++14 and C++17
     # https://devblogs.microsoft.com/cppblog/announcing-msvc-conforms-to-the-c-standard/
     if(NOT MSVC_VERSION VERSION_GREATER_EQUAL 1914)
--- a/SuperBuild/External_tbb.cmake
+++ b/SuperBuild/External_tbb.cmake
@@ -78,8 +78,10 @@ if (WIN32)
   elseif (NOT MSVC_VERSION VERSION_GREATER 1929)
     # VS2019 is expected to be binary compatible with VS2015
     set(tbb_vsdir vc14)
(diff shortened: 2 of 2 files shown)


jamesobutler (approved):

lassoan:
Can you explain this a bit more? From the discourse link it is not clear for me what is happening and why this fix works.

jamesobutler:
Visual Studio 2015 included MSVC 1900-1909
Visual Studio 2017 included MSVC 1910-1919
Visual Studio 2019 included MSVC 1920-1929
Visual Studio 2022 included MSVC 1930-1939, however 1940 is also Visual Studio 2022 as of version v17.10. Microsoft appears to be using 1940-1949 for Visual Studio 2022 as well.
See https://devblogs.microsoft.com/cppblog/msvc-toolset-minor-version-number-14-40-in-vs-2022-v17-10/

`tbb_vsdir` was not being set when MSVC 1940 was being used. That has been changed with the code logic in this PR.

lassoan:
Thanks for the explanation, this is clear now. I'll add a comment into the code to make things clear there, too.

jcfr:
@cpinter @jamesobutler Can we confirm that the following error is triggered without using this patch ?

> message(FATAL_ERROR "TBB does not support your Visual Studio compiler. [MSVC_VERSION: ${MSVC_VERSION}]")

lassoan (approved):
It all looks good to me now.

We should probably use [MSVC_TOOLSET_VERSION](https://cmake.org/cmake/help/latest/variable/MSVC_TOOLSET_VERSION.html#variable:MSVC_TOOLSET_VERSION), as it would isolate us more from some of these low-level compiler version details, but since Slicer, ITK, and VTK have been only using `MSVC_VERSION`, it is probably better not to switch in Slicer yet.

jamesobutler:
`MSVC_TOOLSET_VERSION` could be used, but it would require updating the minimum required CMake version. The factory build machines are using CMake 3.22 and CMake only knows `1930-1949 = VS 17.0 (v143 toolset)` as of CMake 3.29 where earlier versions know `1930-1939 = VS 17.0 (v143 toolset)`.

jamesobutler:
> @cpinter @jamesobutler Can we confirm that the following error is triggered without using this patch ?
> 
> > message(FATAL_ERROR "TBB does not support your Visual Studio compiler. [MSVC_VERSION: ${MSVC_VERSION}]")

@jcfr Yeah I would've expected @cpinter to have experienced a build error, but if there was none then it was probably not going into that block because `tbb_enabled` wasn't True. That check was added in https://github.com/Slicer/Slicer/commit/3c540129564c1ec79d1f92ad7ec959544ae1717e, when TBB support was originally added, but I'm not sure if that is set by the TBB project or not.

cpinter:
> @cpinter @jamesobutler Can we confirm that the following error is triggered without using this patch ?
> 
> > message(FATAL_ERROR "TBB does not support your Visual Studio compiler. [MSVC_VERSION: ${MSVC_VERSION}]")

This is a good point. I think it does not get triggered, because as I understand this would stop the superbuild, but the build finished successfully, and the error occurred in runtime. Do you want me to double check?

cpinter:
I can try it on Monday if you want to double check this. Let me know

Issue #7835, opened 2024-07-03
Slicer `CI (Build) / Build Slicer (pull_request)` workflow is always failing


jcfr:
## Summary

The `CI (Build) / Build Slicer (pull_request)` workflow is failing.

Notes:
* The [Slicer/SlicerDocker](https://github.com/Slicer/SlicerDocker) project was last updated three months ago
* The build of the associated docker image is likely failing. See https://github.com/Slicer/SlicerDockerUpdate

## Steps to reproduce

Submit a pull request with source files update.

jcfr:
After inspecting the log in `/home/kitware/Packaging/SlicerDockerUpdate/cronjob-log.txt`:

```
[...]

++ make slicer-base.push
docker push slicer/slicer-base:latest
The push refers to a repository [docker.io/slicer/slicer-base]
7abc8dbb34db: Preparing
9a6f568b4b2c: Preparing
bd14cbe6abea: Preparing
2f3e8b38c9af: Preparing
d1af4e865bb4: Preparing
75661c8a984b: Preparing
424fda0cd9ea: Preparing
174f56854903: Preparing
75661c8a984b: Waiting
424fda0cd9ea: Waiting
174f56854903: Waiting
d1af4e865bb4: Retrying in 5 seconds

[...]

7abc8dbb34db: Retrying in 2 seconds
7abc8dbb34db: Retrying in 1 second
x509: certificate has expired or is not yet valid
make: *** [slicer-base.push] Error 1
```

jcfr:
```
$ wget -S -d -O - https://production.cloudflare.docker.com
Setting --output-document (outputdocument) to -
[...]

URI encoding = ‘UTF-8’
Converted file name 'index.html' (UTF-8) -> 'index.html' (UTF-8)
--2024-07-03 10:59:55--  https://production.cloudflare.docker.com/
Resolving production.cloudflare.docker.com (production.cloudflare.docker.com)... 104.16.99.215, 104.16.98.215, 104.16.97.215, ...
Caching production.cloudflare.docker.com => 104.16.99.215 104.16.98.215 104.16.97.215 104.16.101.215 104.16.100.215 2606:4700::6810:62d7 2606:4700::6810:64d7 2606:4700::6810:63d7 2606:4700::6810:65d7 2606:4700::6810:61d7
Connecting to production.cloudflare.docker.com (production.cloudflare.docker.com)|104.16.99.215|:443... connected.
Created socket 3.
[...]

PR #7836, opened 2024-07-03
ENH: Generate unique segment IDs using random identifier

Sunderlandkyl:
Previously adding new segments to a segmentation would generate segment IDs of the format Segment_N unless another segment ID was specified. This commit changes the automatically generated segment IDs to use the format S_XYZ where XYZ is a unique random 16 character alphanumeric string.

This makes segment IDs more meaningful as a unique identifier since previously most segmentations would always contain the same IDs (Segment_1, Segment_1, etc.), now they will be unique across segmentations.

### DIFF
--- a/Libs/vtkSegmentationCore/CMakeLists.txt
+++ b/Libs/vtkSegmentationCore/CMakeLists.txt
@@ -4,6 +4,9 @@ project(vtkSegmentationCore)
 cmake_minimum_required(VERSION 3.16.3...3.19.7 FATAL_ERROR)
 #-----------------------------------------------------------------------------
 
+option(${PROJECT_NAME}_USE_UUID "Use UUIDs for generating segment IDs. Requires gdcm library." ON)
+mark_as_advanced(${PROJECT_NAME}_USE_UUID)
+
 # --------------------------------------------------------------------------
 # Configure headers
 # --------------------------------------------------------------------------
@@ -97,6 +100,14 @@ set(vtkSegmentationCore_INCLUDE_DIRS
 set(vtkSegmentationCore_LIBS
   ${VTK_LIBRARIES}
   )
+
+if (${PROJECT_NAME}_USE_UUID)
+  # In order to use UUIDs for segment IDs, we need to link against the gdcmMSFF library
+  # for gdcm::UIDGenerator.
+  list(APPEND vtkSegmentationCore_LIBS
+    gdcmMSFF
+    )
+endif()
 
 include_directories( ${vtkSegmentationCore_INCLUDE_DIRS} )
 add_library(${PROJECT_NAME} ${vtkSegmentationCore_SRCS})
--- a/Libs/vtkSegmentationCore/vtkSegmentation.cxx
+++ b/Libs/vtkSegmentationCore/vtkSegmentation.cxx
@@ -36,10 +36,12 @@
 #include <vtkImageThreshold.h>
 #include <vtkMath.h>
 #include <vtkMatrix4x4.h>
+#include <vtkMinimalStandardRandomSequence.h>
 #include <vtkNew.h>
 #include <vtkObjectFactory.h>
 #include <vtkPointData.h>
 #include <vtkPolyData.h>
+#include <vtkSingleton.h>
 #include <vtkStringArray.h>
 #include <vtkTransform.h>
 #include <vtkTransformPolyDataFilter.h>
@@ -47,10 +49,31 @@
 
 // STD includes
 #include <algorithm>
+#include <chrono>
 #include <functional>
 #include <sstream>
 
+// GDCM includes
+#ifdef vtkSegmentationCore_USE_UUID
+#include <gdcmUUIDGenerator.h>
+#endif
+
 const int DEFAULT_LABEL_VALUE = 1;
+const int DEFAULT_SEGMENT_ID_LENGTH = 16;
+
+//----------------------------------------------------------------------------
+// The segment ID randomizer singleton instance class.
+// This MUST be default initialized to zero by the compiler and is
+// therefore not initialized here. The classInitialize and classFinalize methods handle this instance.
+class vtkSegmentationRandomSequence : public vtkMinimalStandardRandomSequence
+{
+public:
+  vtkSegmentationRandomSequence() = default;
+  ~vtkSegmentationRandomSequence() = default;
+  static vtkSegmentationRandomSequence* New();
+  VTK_SINGLETON_DECLARE(vtkSegmentationRandomSequence);
+  static vtkSegmentationRandomSequence* GetInstance();
+};
 
 //----------------------------------------------------------------------------
 vtkStandardNewMacro(vtkSegmentation);
@@ -72,6 +95,12 @@ vtkSegmentation::vtkSegmentation()
   this->SegmentModifiedEnabled = true;
 
   this->SegmentIdAutogeneratorIndex = 0;
+
+#ifdef vtkSegmentationCore_USE_UUID
+  this->UUIDSegmentIDs = true;
+#else
+  this->UUIDSegmentIDs = false;
+#endif
 
   this->SetSourceRepresentationName(vtkSegmentationConverter::GetSegmentationBinaryLabelmapRepresentationName());
 }
@@ -271,9 +300,46 @@ bool vtkSegmentation::SetSegmentModifiedEnabled(bool enabled)
 }
 
 //---------------------------------------------------------------------------
-std::string vtkSegmentation::GenerateUniqueSegmentID(std::string id)
+std::string vtkSegmentation::GenerateUniqueSegmentName(std::string base)
 {
-  if (!id.empty() &&  this->Segments.find(id) == this->Segments.end())
+  if (base.empty())
+  {
+    base = "Segment";
+  }
+
+  // try to make it unique by attaching a postfix
+  std::string segmentName = base;
+  int nameIndex = 1;
+  while (true)
+  {
+    std::stringstream nameStream;
+    bool found = false;
+    nameStream << base << "_" << nameIndex;
+    for (auto segment : this->Segments)
+    {
+      if (segment.second->GetName() == nameStream.str())
+      {
+        found = true;
+        break;
+      }
+    }
+
+    if (!found)
+    {
+      segmentName = nameStream.str();
+      break;
+    }
+
+    nameIndex++;
+  }
+
+  return segmentName;
+}
+
+//---------------------------------------------------------------------------
+std::string vtkSegmentation::GenerateUniqueSegmentID(std::string id/*=""*/)
+{
+  if (!id.empty() && this->Segments.find(id) == this->Segments.end())
   {
     // the provided id is already unique
     return id;
@@ -281,8 +347,24 @@ std::string vtkSegmentation::GenerateUniqueSegmentID(std::string id)
 
   if (id.empty())
   {
-    // use a non-empty default prefix if no id is provided
-    id = "Segment";
+    std::string newSegmentID;
+    while (newSegmentID.empty() || this->Segments.find(newSegmentID) != this->Segments.end())
+    {
+      if (this->UUIDSegmentIDs)
+      {
+#ifdef vtkSegmentationCore_USE_UUID
+        newSegmentID = vtkSegmentation::GenerateUUIDDerivedUID();
+#else
+        vtkErrorMacro("GenerateUniqueSegmentID: UUID segment IDs are not supported in this build, using random segment IDs");
+#endif
+      }
+
+      if (newSegmentID.empty())
+      {
+        newSegmentID = vtkSegmentation::GenerateRandomSegmentID(DEFAULT_SEGMENT_ID_LENGTH);
+      }
+    }
+    return newSegmentID;
   }
 
   // try to make it unique by attaching a postfix
@@ -306,6 +388,97 @@ std::string vtkSegmentation::GenerateUniqueSegmentID(std::string id)
 
   // try to make it unique by modifying prefix
   return this->GenerateUniqueSegmentID(id + "_");
+}
+
+//---------------------------------------------------------------------------
+std::string ConvertHexadecimalStringToDecimalString(const std::string& hex)
+{
+  // Start with a vector containing a single zero.
+  std::vector<int> decimals(1, 0);
+  for (char rawDigit : hex)
+  {
+    int hexDigit = std::tolower(rawDigit);
+    int value = 0;
+    if (hexDigit >= '0' && hexDigit <= '9')
+    {
+      value = hexDigit - '0';
+    }
+    else if (hexDigit >= 'a' && hexDigit <= 'f')
+    {
+      value = hexDigit - 'a' + 10;
+    }
+    else
+    {
+      // Skip invalid characters such as '-'.
+      continue;
+    }
+
+    int carry = value;
+    for (size_t i = 0; i < decimals.size(); ++i)
+    {
+      int current = decimals[i] * 16 + carry;
+      decimals[i] = current % 10;
+      carry = current / 10;
+    }
+
+    while (carry > 0)
+    {
+      decimals.push_back(carry % 10);
+      carry /= 10;
+    }
+  }
+
+  // Convert integer vector back to string, working backwards.
+  std::stringstream resultSS;
+  for (auto it = decimals.rbegin(); it != decimals.rend(); ++it)
+  {
+    resultSS << char(*it + '0');
+  }
+
+  return resultSS.str();
+}
+
+//---------------------------------------------------------------------------
+std::string vtkSegmentation::GenerateUUIDDerivedUID()
+{
+#ifndef vtkSegmentationCore_USE_UUID
+  vtkErrorWithObjectMacro(nullptr, "GenerateUUIDDerivedUID: UUID segment IDs are not supported in this build");
+  return "";
+#else
+  std::stringstream uuidStream;
+  uuidStream << "2.25."; // UUID derived UID. See https://dicom.nema.org/medical/dicom/current/output/chtml/part05/sect_b.2.html.
+
+  gdcm::UUIDGenerator uuidGenerator;
+  const char* uuid = uuidGenerator.Generate();
+  // Can't store the result in an integer type since it is 128 bits long.
+  // Resulting string is maximum 39 characters long.
+  uuidStream << ConvertHexadecimalStringToDecimalString(uuid);
+  return uuidStream.str();
+#endif
+}
+
+//---------------------------------------------------------------------------
+std::string vtkSegmentation::GenerateRandomSegmentID(int suffixLength, std::string validCharacters/*=""*/)
+{
+  if (suffixLength <= 0)
+  {
+    vtkErrorWithObjectMacro(nullptr, "GenerateRandomSegmentID: Invalid suffix length, must be greater than 0");
+    return "";
+  }
+
+  if (validCharacters.empty())
+  {
+    validCharacters = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789";
+  }
+
+  vtkMinimalStandardRandomSequence* randomSequence = vtkSegmentation::GetSegmentIDRandomSequenceInstance();
+  std::string randomString = "S_";
+  for (int i = 0; i < suffixLength; ++i)
+  {
+    int index = static_cast<int>(std::floor(std::fmod(randomSequence->GetNextValue(), 1.0) * validCharacters.size()));
+    randomString += validCharacters[index];
+  }
+  return randomString;
 }
 
 //---------------------------------------------------------------------------
@@ -476,8 +649,7 @@ bool vtkSegmentation::AddSegment(vtkSegment* segment, std::string segmentId/*=""
       vtkErrorMacro("AddSegment: Unable to add segment without a key; neither key is given nor segment name is defined!");
       return false;
     }
-    key = segment->GetName();
-    key = this->GenerateUniqueSegmentID(key);
+    key = this->GenerateUniqueSegmentID();
   }
   this->Segments[key] = segment;
   if (insertBeforeSegmentId.empty())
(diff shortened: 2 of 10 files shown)


pieper:
I'm thinking of the case where you sometimes need to type these by hand and 16 random characters sounds like a burden.  How about just three characters, but we still require them to be unique so we aren't relying on UUID statistical-style uniqueness but we make it clear to users that the codes are meaningless.

Sunderlandkyl:
We could reduce it from 16, though I think that 3 would be too short.

Out of curiosity what are the cases where we would type a segment ID manually?
I think that generally we should encourage people to get the segment IDs from the segmentation rather than typing them out.

jcfr:
After fixing the upload of the `slicer/slicer-base`[^1] docker image used for testing, I just re-triggered the build.

Related issue:
* https://github.com/Slicer/Slicer/issues/7835

[^1]: https://hub.docker.com/r/slicer/slicer-base/tags?page=&page_size=&ordering=&name=latest

Libs/vtkSegmentationCore/vtkSegmentation.cxx
```diff
@@ -2514,3 +2588,66 @@ void vtkSegmentation::CopySegment(vtkSegment* destination, vtkSegment* source, v
...
+
+//----------------------------------------------------------------------------
+// Implementation of vtkSegmentationRandomSequenceInitialize class.
+//----------------------------------------------------------------------------
+static vtkSegmentationRandomSequenceInitialize* vtkSegmentationRandomSequenceInitializeInstance;
+
+// Must NOT be initialized.  Default initialization to zero is necessary.
+unsigned int vtkSegmentationRandomSequenceInitialize::Count;
+
+vtkSegmentationRandomSequenceInitialize::vtkSegmentationRandomSequenceInitialize()
```
jcfr, line 2638: We could move `vtkSingleton.h`[^1] into `vtkAddon`[^2] or `MRML/Core` and use it here.

[^1]: https://github.com/Slicer/Slicer/blob/main/Libs/MRML/DisplayableManager/vtkSingleton.h
[^2]: https://github.com/Slicer/vtkAddon

lassoan:
> I'm thinking of the case where you sometimes need to type these by hand and 16 random characters sounds like a burden

Users can always look up the segment ID by segment name, so I don't think that there would be a need for them to type the ID. The ID would act similarly to a memory pointer, which you would sometimes have a look at, compare, but never really need to type in.

By using more characters we would have IDs that would not be the same by chance: matching ID would mean matching content. For example, this could be used for tracking lung nodule instances (each nodule would use the same ID across all time points).

Libs/vtkSegmentationCore/vtkSegmentation.cxx
```diff
@@ -271,18 +280,59 @@ bool vtkSegmentation::SetSegmentModifiedEnabled(bool enabled)
 }
 
 //---------------------------------------------------------------------------
-std::string vtkSegmentation::GenerateUniqueSegmentID(std::string id)
+std::string vtkSegmentation::GenerateUniqueSegmentName(std::string base)
 {
-  if (!id.empty() &&  this->Segments.find(id) == this->Segments.end())
+  if (base.empty())
+  {
+    base = "Segment";
```
lassoan, line 287: The segment name is displayable on the GUI, so it should be translatable. Since order of words and numbers are language specific, instead of using a `base` string, it would be better to pass a `pattern = vtkMRMLTr("vtkSomeClassName", "Segment_%1")`.

Inside this method we should use `vtkMRMLI18N::Format(pattern, nameIndexStr)` to replace the `%1` placeholder. However, this library does not depend on MRML, so maybe we would need to copy `vtkMRMLI18N::Format` method into this class (maybe not the complete method, just a version that handles one placeholder). Maybe we could move `vtkMRMLI18N::Format` to `vtkAddon` with some more generic name like `vtkAddonStringUtilities::FormatString(...)`.

lassoan, line 287:
```suggestion
    pattern = "Segment_%1";
```

Libs/vtkSegmentationCore/vtkSegmentation.cxx
```diff
@@ -271,18 +280,59 @@ bool vtkSegmentation::SetSegmentModifiedEnabled(bool enabled)
...
   {
     // the provided id is already unique
     return id;
   }
 
   if (id.empty())
   {
-    // use a non-empty default prefix if no id is provided
-    id = "Segment";
+    std::string randomString = "";
```
lassoan, line 330:
```suggestion
    std::string randomString;
```

Libs/vtkSegmentationCore/vtkSegmentation.cxx
```diff
@@ -308,6 +358,30 @@ std::string vtkSegmentation::GenerateUniqueSegmentID(std::string id)
...
+  if (validCharacters.empty())
+  {
+    validCharacters = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789";
+  }
+
+  vtkMinimalStandardRandomSequence* randomSequence = vtkSegmentation::GetSegmentIDRandomSequenceInstance();
+  std::string randomString = "S_";
+  for (int i = 0; i < suffix_Length; ++i)
+  {
+    int index = static_cast<int>(std::floor(randomSequence->GetNextValue() * validCharacters.size()));
```
lassoan, line 379: In theory, we may get the `9` character, but practically it would never happen, as we would just get very close to 1.0 but not actually get 1.0. It would be nice if we would get any of the valid characters with about the same probability.

Libs/vtkSegmentationCore/vtkSegmentation.cxx
```diff
@@ -1808,14 +1881,14 @@ std::string vtkSegmentation::AddEmptySegment(std::string segmentId/*=""*/, std::
...
   // Segment ID will be segment name by default
-  segmentId = this->GenerateUniqueSegmentID(segmentId);
   if (!segmentName.empty())
   {
     segment->SetName(segmentName.c_str());
   }
   else
   {
-    segment->SetName(segmentId.c_str());
+    std::string name = this->GenerateUniqueSegmentName("Segment");
```
lassoan, line 1890: We have to pass in this segment name pattern from outside (from i18n-aware code)

Libs/vtkSegmentationCore/vtkSegmentation.cxx
```diff
@@ -2514,3 +2588,66 @@ void vtkSegmentation::CopySegment(vtkSegment* destination, vtkSegment* source, v
...
+    vtkSegmentationRandomSequence = (vtkMinimalStandardRandomSequence*)vtkObjectFactory::CreateInstance("vtkMinimalStandardRandomSequence");
+    // if the factory did not provide one, then create it here
+    if (!vtkSegmentationRandomSequence)
+    {
+      vtkSegmentationRandomSequence = vtkMinimalStandardRandomSequence::New();
+    }
+
+    // Get the current ms since the epoch
+    time_t currentTime = std::chrono::duration_cast<std::chrono::microseconds>(std::chrono::system_clock::now().time_since_epoch()).count();
+    int seed = currentTime % 2147483646;
```
lassoan, line 2622: Add a comment explaining this number (2147483646).

Libs/vtkSegmentationCore/vtkSegmentation.cxx
```diff
@@ -47,10 +48,18 @@
...
 #include <sstream>
 
 const int DEFAULT_LABEL_VALUE = 1;
+const int DEFAULT_SEGMENT_ID_LENGTH = 16;
+
+//----------------------------------------------------------------------------
+// The segment ID randomizer singleton instance.
+// This MUST be default initialized to zero by the compiler and is
+// therefore not initialized here. The classInitialize and classFinalize methods handle this instance.
+static vtkMinimalStandardRandomSequence* vtkSegmentationRandomSequence;
```
lassoan, line 62: `vtkSegmentationRandomSequence` name can be mistaken for a class name. It would be better to change the name that reflects that it is a global variable.

Libs/vtkSegmentationCore/vtkSegmentation.cxx
```diff
@@ -2514,3 +2588,66 @@ void vtkSegmentation::CopySegment(vtkSegment* destination, vtkSegment* source, v
...
+    // if the factory did not provide one, then create it here
+    if (!vtkSegmentationRandomSequence)
+    {
+      vtkSegmentationRandomSequence = vtkMinimalStandardRandomSequence::New();
+    }
+
+    // Get the current ms since the epoch
+    time_t currentTime = std::chrono::duration_cast<std::chrono::microseconds>(std::chrono::system_clock::now().time_since_epoch()).count();
+    int seed = currentTime % 2147483646;
+    vtkSegmentationRandomSequence->SetSeed(currentTime);
```
lassoan, line 2623: Do we really need to seed every time? (I would have assumed that we only need to set the seed when a new vtkSegmentationRandomSequence is created)

Libs/vtkSegmentationCore/vtkSegmentation.cxx
lassoan: Agreed

Modules/Scripted/SegmentStatistics/SegmentStatistics.py
```diff
@@ -802,7 +802,7 @@ def test_SegmentStatisticsBasic(self):
             sphereSource.SetCenter(segmentGeometry[1], segmentGeometry[2], segmentGeometry[3])
             sphereSource.Update()
             uniqueSegmentID = segmentationNode.GetSegmentation().GenerateUniqueSegmentID("Test")
-            segmentationNode.AddSegmentFromClosedSurfaceRepresentation(sphereSource.GetOutput(), uniqueSegmentID)
+            segmentationNode.AddSegmentFromClosedSurfaceRepresentation(sphereSource.GetOutput(), uniqueSegmentID, None, uniqueSegmentID)
```
lassoan, line 805: It obscures the intent of the test a bit that we use segment ID as both segment name and ID. Something like this would be more clear:
```suggestion
            segmentationNode.AddSegmentFromClosedSurfaceRepresentation(sphereSource.GetOutput(), None, None, uniqueSegmentID)
```

Modules/Scripted/SegmentStatistics/SegmentStatistics.py
```diff
@@ -859,7 +859,7 @@ def test_SegmentStatisticsPlugins(self):
             sphereSource.Update()
             segment = vtkSegmentationCore.vtkSegment()
             uniqueSegmentID = segmentationNode.GetSegmentation().GenerateUniqueSegmentID("Test")
-            segmentationNode.AddSegmentFromClosedSurfaceRepresentation(sphereSource.GetOutput(), uniqueSegmentID)
+            segmentationNode.AddSegmentFromClosedSurfaceRepresentation(sphereSource.GetOutput(), uniqueSegmentID, None, uniqueSegmentID)
```
lassoan, line 862:
```suggestion
            segmentationNode.AddSegmentFromClosedSurfaceRepresentation(sphereSource.GetOutput(), None, None, uniqueSegmentID)
```

lassoan (commented):
It looks good to me, just added a number of nitpicky comments inline.

Libs/vtkSegmentationCore/vtkSegmentation.cxx
Sunderlandkyl: Thanks, PR created [here](https://github.com/Slicer/vtkAddon/pull/49) and I've updated this PR to use vtkSingleton macros.

Libs/vtkSegmentationCore/vtkSegmentation.cxx
Sunderlandkyl: It was only set once when the singleton hadn't been defined yet. In the new version this is explicit as it is in classInitialize().

Libs/vtkSegmentationCore/vtkSegmentation.cxx
Sunderlandkyl: Fixed by updating to use the macros.

Libs/vtkSegmentationCore/vtkSegmentation.cxx
Sunderlandkyl: Actually we get 9 as often as the other numbers, however I think that there is a low probability that we will go out of bounds if we get 1.0. Either way I'll update this.

Libs/vtkSegmentationCore/vtkSegmentation.cxx
Sunderlandkyl: I don't think we can use vtkMRMLTr either unless we want to move all of the translation functions into vtkAddon.

lassoan: If that makes things easier, we can move `vtkMRMLI18N` to `vtkAddon`. We would then need to rename to `vtkI18N` or maybe `vtkInternationalization`. It is a singleton class, so maybe we should move the singleton macros to `vtkAddon`, too?

pieper:
> Out of curiosity what are the cases where we would type a segment ID manually?

I know I have typed them at that python console during testing.  I guess I could cut-and-paste in most cases though.  I wouldn't expect a user to have to type them. 

> By using more characters we would have IDs that would not be the same by chance: matching ID would mean matching content. 

If this is the goal then suggest we just use UUIDs so it's clear we don't want a vanishingly low chance of collision.

> For example, this could be used for tracking lung nodule instances (each nodule would use the same ID across all time points).

Yes, we could represent this in dicom sr too.

lassoan:
> If this is the goal then suggest we just use UUIDs so it's clear we don't want a vanishingly low chance of collision

Do you mean [DICOM Part 5 B.2 UUID Derived UID](https://dicom.nema.org/medical/dicom/current/output/chtml/part05/sect_B.2.html)?

It is a good idea, as it would allow us to store a segmentation in DICOM without any data loss (we could preserve segment IDs).

pieper:
Yes, exactly, then it could be used here: https://dicom.innolitics.com/ciods/segmentation/segmentation-image/00620002/00620021

Sunderlandkyl:
Segment IDs are now generated using the form specified in: https://dicom.nema.org/medical/dicom/current/output/chtml/part05/sect_b.2.html.

Ex. `2.25.329800735698586629295641978511506172918`

lassoan (commented):
Thanks a lot, this looks good, but it wpuod not be ideal to add a hard dependency of the segmentation library to gdcm. We should add at least a CMake flag that would choose between a simple random number generator (that does not add a new dependency, lile the one you inplemented earlier) and the GDCM UUID generator.

Sunderlandkyl:
@lassoan Thanks, I've added a flag to switch between uuid/random segment IDs.
By default we compile with UUID support enabled, but it can be disabled.

This PR again now relies on https://github.com/Slicer/vtkAddon/pull/49 to be integrated.

Libs/vtkSegmentationCore/CMakeLists.txt
```diff
@@ -98,10 +98,24 @@ set(vtkSegmentationCore_LIBS
   ${VTK_LIBRARIES}
   )
 
+if (NOT DEFINED ${PROJECT_NAME}_USE_UUID)
+  set(${PROJECT_NAME}_USE_UUID ON)
+endif()
+
+if (${PROJECT_NAME}_USE_UUID)
+  list(APPEND vtkSegmentationCore_LIBS
+    gdcmMSFF
```
lassoan, line 107: Add comment why gdcm is needed

Libs/vtkSegmentationCore/CMakeLists.txt
```diff
@@ -98,10 +98,24 @@ set(vtkSegmentationCore_LIBS
...
+    )
+endif()
+
 include_directories( ${vtkSegmentationCore_INCLUDE_DIRS} )
 add_library(${PROJECT_NAME} ${vtkSegmentationCore_SRCS})
 target_link_libraries( ${PROJECT_NAME} ${vtkSegmentationCore_LIBS} )
 
+if (${PROJECT_NAME}_USE_UUID)
+  target_compile_definitions(${PROJECT_NAME} PRIVATE SEGMENTATION_UUID_SEGMENT_IDS)
```
lassoan, line 116: use `#cmakedefine` to make it more explicit

Libs/vtkSegmentationCore/CMakeLists.txt
```diff
@@ -98,10 +98,24 @@ set(vtkSegmentationCore_LIBS
   ${VTK_LIBRARIES}
   )
 
+if (NOT DEFINED ${PROJECT_NAME}_USE_UUID)
```
lassoan, line 101: Add an `OPTION`

Libs/vtkSegmentationCore/vtkSegmentation.cxx
```diff
@@ -36,21 +36,44 @@
...
 #include <vtkVersion.h>
 
 // STD includes
 #include <algorithm>
+#include <chrono>
 #include <functional>
 #include <sstream>
 
+// GDCM includes
+#if defined(SEGMENTATION_UUID_SEGMENT_IDS) && SEGMENTATION_UUID_SEGMENT_IDS
```
lassoan, line 57: simplify

Libs/vtkSegmentationCore/vtkSegmentation.cxx
```diff
@@ -271,18 +300,73 @@ bool vtkSegmentation::SetSegmentModifiedEnabled(bool enabled)
...
   }
 
   if (id.empty())
   {
-    // use a non-empty default prefix if no id is provided
-    id = "Segment";
+    std::string newSegmentID;
+    while (newSegmentID.empty() || this->Segments.find(newSegmentID) != this->Segments.end())
+    {
+#if !defined(SEGMENTATION_UUID_SEGMENT_IDS) || !SEGMENTATION_UUID_SEGMENT_IDS
```
lassoan, line 353: Maybe simplify

Libs/vtkSegmentationCore/vtkSegmentation.cxx
jamesobutler: Does this also need to take into account whether Slicer is built with DICOM support? The code currently has some compile logic using things such as:

`Slicer_BUILD_DICOM_SUPPORT`

`#ifdef VTKITK_BUILD_DICOM_SUPPORT`

lassoan: We did not want to reimplement an ISO standard UUID generator but did not want to add a new mandatory dependency either. That's why the CMake option was added.

When vtkSegmentationCore is used within Slicer then the UUID generation code in GDCM is should be available even if Slicer is built without DICOM support (because Slicer is always built with ITK and Slicer's ITK always includes GDCM), but it would be nice to test it. If it turns out that the GDCM UUID generator is only available when DICOM support is enabled then UUID generation should be disabled (and the plain random sequence generator will be used).

Sunderlandkyl: Slicer_BUILD_DICOM_SUPPORT only seems to impact DCMTK and not GCDM, so it doesn't look like we need to test if Slicer is built with DICOM support.

Libs/vtkSegmentationCore/CMakeLists.txt
```diff
@@ -4,6 +4,9 @@ project(vtkSegmentationCore)
 cmake_minimum_required(VERSION 3.16.3...3.19.7 FATAL_ERROR)
 #-----------------------------------------------------------------------------
 
+option(${PROJECT_NAME}_USE_UUID "Use UUIDs for generating segment IDs. Requires linking against gdcm" ON)
```
lassoan, line 7:
```suggestion
option(${PROJECT_NAME}_USE_UUID "Use UUIDs for generating segment IDs. Requires gdcm library." ON)
```

Libs/vtkSegmentationCore/vtkSegmentation.cxx
```diff
@@ -271,18 +300,71 @@ bool vtkSegmentation::SetSegmentModifiedEnabled(bool enabled)
...
+      if (this->UUIDSegmentIDs)
+      {
+#ifdef vtkSegmentationCore_USE_UUID
+        newSegmentID = vtkSegmentation::GenerateUUIDDerivedUID();
+#else
+        vtkErrorMacro("GenerateUniqueSegmentID: UUID segment IDs are not supported in this build, using random segment IDs");
+#endif
+      }
+
+      if (!this->UUIDSegmentIDs || newSegmentID.empty())
```
lassoan, line 362: Can we simplify to this?

```suggestion
      if (newSegmentID.empty())
```

Libs/vtkSegmentationCore/vtkSegmentation.cxx
```diff
@@ -308,6 +390,97 @@ std::string vtkSegmentation::GenerateUniqueSegmentID(std::string id)
...
+#ifndef vtkSegmentationCore_USE_UUID
+  vtkErrorWithObjectMacro(nullptr, "GenerateUUIDDerivedUID: UUID segment IDs are not supported in this build");
+  return "";
+#else
+  std::stringstream uuidStream;
+  uuidStream << "2.25."; // UUID derived UID. See https://dicom.nema.org/medical/dicom/current/output/chtml/part05/sect_b.2.html.
+
+  gdcm::UUIDGenerator uuidGenerator;
+  const char* uuid = uuidGenerator.Generate();
+  // Can't store the result in an integer type since it needs a 128-bits.
```
lassoan, line 453:
```suggestion
  // Can't store the result in an integer type since it is 128 bits long.
```

lassoan (commented):
Looks good, thank you, just added a few trivial comments. The only remaining piece now is the i18n compliance.

Libs/vtkSegmentationCore/vtkSegmentation.cxx
lassoan: Good, thanks for checking!

Sunderlandkyl:
I will follow up this PR with another one that includes the i18n compliance changes, including moving the classes to vtkAddon. For now I've made the vtkSegmentation::GenerateUniqueSegmentName() method protected so that it won't be used by users before the i18n changes are made.

lassoan (approved):
This all looks good then. Thank you!

PR #7849, opened 2024-07-12
ENH: Migrate vtkSingleton to vtkAddon

Sunderlandkyl:
This commit moves vtkSingleton from MRMLDM to vtkAddon. The view factories are also updated to remove functions that have been added to the base VTK_SINGLETON_CXX macro.


lassoan (approved):
Thank you, thus looks great! The CI build error is due to that vtkAddon is part of the base image, so I'll curve the merge.

PR #7855, opened 2024-07-16
ENH: Add API to log software usage events

lassoan:
The new qSlicerCoreApplication::logUsageEvent can be called by modules in Slicer core and extensions to indicate that certain software feature is used. The application itself does not process the event, just emits a 'usageEventLogged' signal, which can be connected to slots in external modules.

This API can be used for example for implementing an extension that collects usage data and computes usage statistics of various features. If no extensions are installed that processes software usage data then calling qSlicerCoreApplication::logUsageEvent has no effect.

How to test:

```
# Observe the usage event and print it if any is logged
slicer.app.connect("usageEventLogged(QString,QString)", lambda component, event: print(f"{component} -> {event}"))

# Log an event
slicer.app.logUsageEvent("TotalSegmentator", "segmentation/ct/total")
```

See related discussions at https://discourse.slicer.org/t/should-we-start-collecting-software-usage-data/30873

---

This pull request deprecates https://github.com/Slicer/Slicer/pull/7788, as we do not want to put any actual usage data recording into Slicer core, just a hook that allows implementing extensions in the future (that the user can optionally install) that can collect usage data.

@jcfr @pieper @sjh26 @jamesobutler @cpinter

### DIFF
--- a/Base/QTCore/qSlicerCoreApplication.cxx
+++ b/Base/QTCore/qSlicerCoreApplication.cxx
@@ -42,6 +42,7 @@
 //  - Slicer_BUILD_DICOM_SUPPORT
 //  - Slicer_BUILD_EXTENSIONMANAGER_SUPPORT
 //  - Slicer_BUILD_I18N_SUPPORT
+//  - Slicer_BUILD_USAGE_LOGGING_SUPPORT
 //  - Slicer_BUILD_WIN32_CONSOLE
 //  - Slicer_BUNDLE_LOCATION
 //  - Slicer_CLIMODULES_BIN_DIR
@@ -2395,4 +2396,23 @@ bool qSlicerCoreApplication::loadFiles(const QStringList& filePaths, vtkMRMLMess
 void qSlicerCoreApplication::openUrl(const QString& url)
 {
   emit urlReceived(url);
+}
+
+//------------------------------------------------------------------------------
+bool qSlicerCoreApplication::isUsageLoggingSupported() const
+{
+#ifdef Slicer_BUILD_USAGE_LOGGING_SUPPORT
+  return true;
+#else
+  return false;
+#endif
+}
+
+//------------------------------------------------------------------------------
+void qSlicerCoreApplication::logUsageEvent(const QString& component, const QString& event)
+{
+#ifdef Slicer_BUILD_USAGE_LOGGING_SUPPORT
+  Q_D(const qSlicerCoreApplication);
+  emit usageEventLogged(component, event);
+#endif
 }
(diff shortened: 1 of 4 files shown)


Base/QTCore/qSlicerCoreApplication.h
```diff
@@ -573,6 +573,24 @@ class Q_SLICER_BASE_QTCORE_EXPORT qSlicerCoreApplication : public QApplication
...
+  /// \param event is the name of the event.
+  /// The event name must be shorter than 50 characters to avoid potential performance degradation.
+  /// The event name must not contain any information about the user or any of the processed data to
+  /// alleviate any privacy concerns when handling software usage data.
+  /// It is recommended to use all lowercase and use '/' as separator if events can be organized
+  /// in a hierarchical structure (for example, 'planning/model created', 'planning/model exported', 'segmentation/ct/total', 'segmentation/mr/knee').
+  ///
+  /// For example, an extension can report usage data like this:
+  /// - C++: <code>qSlicerCoreApplication::application()->logUsageEvent("SlicerRT", "dicom/export/sro")</code>
+  /// - Python: <code>slicer.app.logUsageEvent("TotalSegmentator", "segmentation/total")</code>
```
jcfr, line 590: Instead of using `/` as delimiter, I suggest to tweak event name like this:
* use `.` like it is done in https://doc.qt.io/qt-6/qloggingcategory.html
* only accept letter, digit and `_`

Base/QTCore/qSlicerCoreApplication.cxx
```diff
@@ -2396,3 +2396,9 @@ void qSlicerCoreApplication::openUrl(const QString& url)
 {
   emit urlReceived(url);
 }
+
+//------------------------------------------------------------------------------
+void qSlicerCoreApplication::logUsageEvent(const QString& component, const QString& event)
+{
+  emit usageEventLogged(component, event);
```
jcfr, line 2403: Conditionally call this based on a build option, will be useful for custom application

jcfr: Consider skipping when testing is enabled ? and or if build is not in "Release" ?

Base/QTCore/qSlicerCoreApplication.h
jcfr: Should we differentiate between a function called as an effect of using the UI ? and as an effect of using an API call ?

Following today's discussion, developer may add a dedicated gui event.

Base/QTCore/qSlicerCoreApplication.cxx
lassoan: Added `Slicer_BUILD_USAGE_LOGGING_SUPPORT` CMake option to completely disable this feature (`logUsageEvent` has no effect if this flag is disabled).

Further logic would be more suitable to implement at lower levels:
- testing mode: If the feature was completely disabled in testing mode then it would be hard to test the feature. Testing mode may also be used for batch processing, where we may want get usage statistics from. Instead the modules that get the event (such as the data collector in Telemetry extension) could check the state of the application and process the events accordingly (e.g., number of events occurring during automatic testing could be collected independently from events that occur outside of testing mode).
- Debug/release mode: The C++ build mode does not necessarily reflect if the application is being developed. Release mode may be used for development, especially if modules are developed in Python. Also, "Developer mode" could be used for this purpose, too (but that may be enabled by too many people for too long time). Overall, this logic is not simple and would be too rigid to specify in the Slicer core application.

lassoan:
@jcfr Thanks for your comments, I've addressed all of them (updated the code or answered the question). Please review and approve if looks good. Thank you!

PR #7858, opened 2024-07-18
BUG: Force loading images from DICOM using right-handed coordinate system

lassoan:
Automatically flip left handed volumes when loading into 3D Slicer from DICOM to make them compatible with all algorithms. Fixes flipped normals in segmentation node for left handed volumes.

Renamed FlipIJKCoordinateSystemHandedness method to ReverseSliceOrder to make it explicit that flipping of handedness is achieved by reversing slice order. The method was introduced recently (no Slicer Stable Release has been created since then), therefore the old name was not deprecated but removed.

See related commit that applies this regularization when the image is read using volume storage node: https://github.com/Slicer/Slicer/commit/3802d812b1a5d855399d8a19cea0bd1705715d52

See discussion at https://discourse.slicer.org/t/dentalsegmentator-result-mesh-is-inverted/37408/9

### DIFF
--- a/Libs/MRML/Core/vtkMRMLVolumeNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLVolumeNode.cxx
@@ -1362,11 +1362,11 @@ bool vtkMRMLVolumeNode::IsIJKCoordinateSystemRightHanded(vtkMatrix4x4* ijkToRasM
 }
 
 //----------------------------------------------------------------------------
-void vtkMRMLVolumeNode::FlipIJKCoordinateSystemHandedness(vtkImageData* imageData, vtkMatrix4x4* ijkToRasMatrix)
+void vtkMRMLVolumeNode::ReverseSliceOrder(vtkImageData* imageData, vtkMatrix4x4* ijkToRasMatrix)
 {
   if (ijkToRasMatrix == nullptr)
   {
-    vtkGenericWarningMacro("vtkMRMLVolumeNode::FlipIJKCoordinateSystemHandedness failed: ijkToRasMatrix is invalid");
+    vtkGenericWarningMacro("vtkMRMLVolumeNode::ReverseSliceOrder failed: ijkToRasMatrix is invalid");
     return;
   }
   int imageDimensions[3] = { 0, 0, 0 };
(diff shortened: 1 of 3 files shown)


pieper:
I tend to think that algorithms that make assumptions about memory layout be responsible for checking and correcting data before processing (with logic methods that implement common use cases like this one), but I guess I can't think of any specific places where enforcing right-handedness will cause a problem.

pieper (commented):
Somehow it feels like the dicom plugin is the wrong place to be enforcing this right-handedness constraint - i.e. shouldn't other readers also need to perform the same operation for consistency?  

Maybe instead we put this logic in the `vtkMRMLVolumeNode` layer so that only right-handed volumes are allowed and/or that the api for the volume node performs the reshuffling if it's given data that is not right handed (I believe ITK does this internally, but MRML has never had this constraint if you create a node from scratch).  

One option I kind of like would be to allow more flexibility at the volume node level but then add an api like `GetNormalizedImageData` or similar that provided you with a version of the data that met a specific set of constraints.  Then an algorithm that requires a specific type of data, say isotropic right-handed axial data could request it so that no further error checking is required.

lassoan:
The current strategy with negative-determinant IJK to RAS direction matrix is the same as with sheared IJK to RAS direction matri in ITK, VTK, and Slicer. Negative-determinant or sheared matrix is supported by data storage and IO classes, but application-level logic is expected to detect and fix these cases because many operations in ITK and VTK assume positive determinant and orthogonal directions. Developers can still use low-level API to store and use any matrix, but users of GUI and high-level APIs are prevented from unknowingly creating such matrices.

While we are overall consistent in how we handle unexpected direction matrices, I agree that the implementation is not necessarily optimal. For example, it may not be clear what is a low-level API (that enforces orthogonality and handedness constraints) and probably these constraints are not documented clearly enough in at Slicer or MRML level. It could be also nice to combine all geometry checks and normalizations (unfortunately, currently the shear normalization is implemented in Python and it requires more data, so it would require some effort to consolidate them).

> Somehow it feels like the dicom plugin is the wrong place to be enforcing this right-handedness constraint - i.e. shouldn't other readers also need to perform the same operation for consistency?

Yes, of course it is not only done in the DICOM module: the most commonly used code path is [in the archetype storage node](https://github.com/Slicer/Slicer/blob/67a6d2ba896580a43b488716fbbd961cc818f302/Libs/MRML/Core/vtkMRMLVolumeArchetypeStorageNode.cxx#L537-L540). We just missed the normalization in the DICOM module in the first round of implementation, because we assumed that the DICOM reader always povides right-handed IJK, but apparently this is not guaranteed.

> Maybe instead we put this logic in the vtkMRMLVolumeNode layer so that only right-handed volumes are allowed and/or that the api for the volume node performs the reshuffling if it's given data that is not right handed (I believe ITK does this internally, but MRML has never had this constraint if you create a node from scratch).

I agree, that there could be improvements. As I wrote above, it should be clear what is a low-level API (that allows advanced users to do what they want) and what is a high-level API (that is safe and convenient for most users). Doing all normalizations in one `GetNormalizedData`-like method could be useful, too, but then we should move the current grid transform based image resampling from the DICOM module in Python to MRML in C++ (or maybe just implement a linear transform based resampling for sheared volumes). It would be nicer, more consistent, it would allow us to correct sheared transforms when reading NRRD files (not just when reading DICOM), but it may take a several days of work and would not offer significant functional improvements, so I'm not sure it would be the best use of our time.

pieper (approved):

PR #7878, opened 2024-08-07
BUG: Fix issue saving transform including non-origin center of transformation

Sunderlandkyl:
When saving a transform that includes rotation and a non-origin center of transformation, the translation component is shifted. Fixed by setting the itk::AffineTransform center of transformation before the matrix and offset in vtkITKTransformConverter.

Added tests for saving/loading transforms that include a center of rotation.


Libs/MRML/Core/vtkITKTransformConverter.h
```diff
@@ -312,13 +312,13 @@ bool vtkITKTransformConverter::SetITKLinearTransformFromVTK(vtkObject* loggerObj
...
   AffineTransformType::Pointer affine = AffineTransformType::New();
-  affine->SetMatrix(itkmat);
-  affine->SetOffset(itkoffset);
   if (center_RAS)
   {
     double center_LPS[3] = { -center_RAS[0], -center_RAS[1], center_RAS[2] };
     affine->SetCenter(center_LPS);
   }
+  affine->SetMatrix(itkmat);
+  affine->SetOffset(itkoffset);
```
lassoan, line 321: This [note in the ITK API documentation](https://github.com/InsightSoftwareConsortium/ITK/blob/4f675a2f24f496de0f58dcb17c4e80dc2fb462e8/Modules/Core/Transform/include/itkMatrixOffsetTransformBase.h#L246-L247) suggests that we should set matrix+center+translation OR matrix+offset. Here we set center, matrix, and offset. We should at least explain here in a comment why our implementation is appropriate.

lassoan (commented):
Thank you good catch! It is great that you have now added thorough testing of this. I've added a comment inline.

Libs/MRML/Core/vtkITKTransformConverter.h
Sunderlandkyl: I've tried using matrix+center+translation, however that seems to break compatibility with older versions of Slicer. I've added a comment for now explaining why we use center+matrix+offset.

lassoan: I see, we just need to document this well.

Do you know why exactly there is a backward compatibility issue? 

Are older versions of Slicer ignore some important component of the transform?

Are the transforms that we save now loaded correctly in older Slicer versions (it is OK that center of rotation is missing, but is the 4x4 matrix the same)? Is it possible that earlier Slicer versions all loaded certain ITK linear transforms incorrectly?

If we resample an image using an ITK-based CLI with an ITK transform, do we get the same results as we would get if we apply the transform in Slicer?

lassoan (commented):
Thanks a lot, everything looks good, except that I think we interpret the center of rotation differently from ITK.

According to the ITK software guide (3.6.2 section and several other places), the center of the transform is generally placed into the middle of the fixed image:

> The CenteredTransformInitializer supports two modes of operation. In the first mode, the centers
of the images are computed as space coordinates using the image origin, size and spacing. The
center of the fixed image is assigned as the rotational center of the transform 

However, when I load a transform that is initialized by ITK the  centered transform initializer then the center of rotation that Slicer displays is not in the center of the moving image (I believe it should be).

How to reproduce:

- Load this scene: https://1drv.ms/u/s!Arm_AFxB9yqH060cFmt2vwzwysEGFw?e=lS8lMI  - FYI, the scene was created using these steps:
  - load MRBrainTumor1 and 2 example images
  - translate and rotate both so that they get at least 20-30cm from each other and harden the transforms on them
  - go  to `General registration (BRAINS)` module, select 1 as fixed, 2 as moving image, linear transform output, `useGeometryAlign` initial alignment, Rigid (6DOF)
- click Apply
- BRAINS logs that the transform's center is `Center: [-164.355, -255.378, 27.2914]` (in LPS) -> GOOD
- Enable interaction for the result transform
- Center of rotation of the transform does not appear in the middle of the fixed volume -> ERROR
- Go to transforms module, show center of rotation position in world coordinate system: it is (751, 688, 63) => ERROR, it should be [164.355, 255.378, 27.2914] in RAS

Sunderlandkyl:
@lassoan try with the latest changes. The center of rotation is now in local instead of parent coordinates, and we shift the center of rotation by the offset before saving/after loading.

This puts the center of rotation in the center of the image:
image

The test still passes and backwards compatibility is maintained.

lassoan (approved):
This is perfect! It works as expected, the comments clearly describe what is done and why, and the tests help with avoiding any regressions in the future. Thank you for all your patience with answering all the comments and addressing all the issues.

PR #7885, opened 2024-08-12
ENH: Disable compression for CLI output files

lassoan:
CLI modules are primarily used for processing and the output file is just used for temporary storage. Therefore, usually compression has no benefits, but it slows down the execution time.

If it turns out in the future that some modules would benefit from having compressed output, those modules could get an optional "--useCompression" flag, similarly to CreateDICOMSeries module.

See discussion at https://discourse.slicer.org/t/crop-volume-resample-scalar-slowness-for-large-image/34985/6


pieper (approved):
Thanks for working on this!

LGTM 👍

PR #7886, opened 2024-08-13
ENH: Allow disabling layers blend clipping to background in slice views

Punzo:
@lassoan

### DIFF
--- a/Libs/MRML/Logic/vtkMRMLSliceLogic.cxx
+++ b/Libs/MRML/Logic/vtkMRMLSliceLogic.cxx
@@ -34,6 +34,7 @@
 #include <vtkAlgorithmOutput.h>
 #include <vtkCallbackCommand.h>
 #include <vtkCollection.h>
+#include <vtkCollectionIterator.h>
 #include <vtkGeneralTransform.h>
 #include <vtkImageAppendComponents.h>
 #include <vtkImageBlend.h>
@@ -104,9 +105,12 @@ struct BlendPipeline
     //   background > AddSubBackroundCast  /
     //
     //
-    //     ... AddSubOutputCast > AddSubExtractRGB \
-    //                                              > AddSubAppendRGBA > Blend
-    //             background > AddSubExtractAlpha /
+    //     ... AddSubOutputCast > AddSubExtractRGB       \
+    //
+    //         background > AddSubExtractBackgroundAlpha - > AddSubAppendRGBA > Blend
+    //
+    //         foreground > AddSubExtractForegroundAlpha /
+    //
     */
 
     this->AddSubForegroundCast->SetOutputScalarTypeToShort();
@@ -118,18 +122,23 @@ struct BlendPipeline
     this->AddSubMath->SetInputConnection(0, this->AddSubBackgroundCast->GetOutputPort());
     this->AddSubMath->SetInputConnection(1, this->ForegroundFractionMath->GetOutputPort());
     this->AddSubOutputCast->SetInputConnection(this->AddSubMath->GetOutputPort());
+    this->AddSubOutputCast->SetOutputScalarTypeToUnsignedChar();
+    this->AddSubOutputCast->ClampOverflowOn();
 
     this->AddSubExtractRGB->SetInputConnection(this->AddSubOutputCast->GetOutputPort());
     this->AddSubExtractRGB->SetComponents(0, 1, 2);
-    this->AddSubExtractAlpha->SetComponents(3);
-    this->AddSubAppendRGBA->AddInputConnection(this->AddSubExtractRGB->GetOutputPort());
-    this->AddSubAppendRGBA->AddInputConnection(this->AddSubExtractAlpha->GetOutputPort());
+    this->AddSubExtractBackgroundAlpha->SetComponents(3);
+    this->AddSubExtractForegroundAlpha->SetComponents(3);
 
-    this->AddSubOutputCast->SetOutputScalarTypeToUnsignedChar();
-    this->AddSubOutputCast->ClampOverflowOn();
+    this->BlendAlpha->AddInputConnection(this->AddSubExtractBackgroundAlpha->GetOutputPort());
+    this->BlendAlpha->AddInputConnection(this->AddSubExtractForegroundAlpha->GetOutputPort());
+
+    this->AddSubAppendRGBA->AddInputConnection(this->AddSubExtractRGB->GetOutputPort());
+    this->AddSubAppendRGBA->AddInputConnection(this->BlendAlpha->GetOutputPort());
   }
 
-  void AddLayers(std::deque<SliceLayerInfo>& layers, int sliceCompositing,
+  void AddLayers(std::deque<SliceLayerInfo>& layers,
+    int sliceCompositing, bool clipToBackgroundVolume,
     vtkAlgorithmOutput* backgroundImagePort,
     vtkAlgorithmOutput* foregroundImagePort, double foregroundOpacity,
     vtkAlgorithmOutput* labelImagePort, double labelOpacity)
@@ -169,7 +178,8 @@ struct BlendPipeline
     {
       this->AddSubForegroundCast->SetInputConnection(foregroundImagePort);
       this->AddSubBackgroundCast->SetInputConnection(backgroundImagePort);
-      this->AddSubExtractAlpha->SetInputConnection(backgroundImagePort);
+      this->AddSubExtractForegroundAlpha->SetInputConnection(foregroundImagePort);
+      this->AddSubExtractBackgroundAlpha->SetInputConnection(backgroundImagePort);
       if (sliceCompositing == vtkMRMLSliceCompositeNode::Add)
       {
         this->AddSubMath->SetOperationToAdd();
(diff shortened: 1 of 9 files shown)


lassoan (commented):
This looks great, thank you! I'm wondering if it is better to add a new bool flag or a new compositing mode.

It would be simpler to add only one more compositing mode: there would be minimal change in the composite node and if we want then it would be easier to expose it on the GUI. But if the flag is compatible many compositing modes then adding a single flag could be more appropriate.

What modes the new flag is compatible with?
What is the difference between `Alpha` and `ReverseAlpha`? Would it make sense to have unclipped mode for `ReverseAlpha` as well? (if the only difference is foreground opacity is `1-opacity` then maybe it will not be kept in the future)
If `Add`&`Subtract` do not really need a bool flag and `ReverseAlpha` is essentially same as `Alpha` then the only mode where clipping could be meaningful is `Alpha` and then a single new enum like `AlphaFull` or `AlphaComplete` could be simpler.

Punzo:
@lassoan all the modes (`Alpha`, `ReverseAlpha`, `Add`, `Subtract`) are supported by the new flag, see screenshots: 

image
image
image
image

NOTE: of course in the case of `Subtract` the subtraction outside the bounding box of the background will be clamped to zero.

Punzo:
NOTE2: in this PR no UI is provided for enabling the option. I can add a button in the sliceview controller, but I was not sure where to place it since it is already very crowded

Punzo:
To Do:
1) Rename: `BlendClipping` into `ClipToBackgroundVolume`
2) if `ClipToBackgroundVolume` is on, 
thresholding is not applied for `Subtract` and `Add` in https://github.com/Slicer/Slicer/pull/7886/files#diff-d5498fd4b04ced370870a7176ba82f0d624ddcd269f72d27577fddf416f871b3R172-R188

jamesobutler:
This would support viewing 2 volumes side-by-side within the same slice viewer? As previously discussed in https://discourse.slicer.org/t/slice-multiple-volumes/9900?

Punzo:
> This would support viewing 2 volumes side-by-side within the same slice viewer? Previous discussions: https://discourse.slicer.org/t/slice-multiple-volumes/9900 https://discourse.slicer.org/t/showing-both-the-background-and-foreground-layer-volumes-in-their-entirety/13690

do you mean like this?

image

or something else?

thanks for referencing the forum links.

lassoan:
The extent of all the layers are already the same (the size of the slice view).

Layer blending is not disabled, the same blending (compositing) modes are still available. This pull request just makes "clipping to background volume region" feature optional, because this clipping is sometimes useful, but sometimes it is not desirable (e.g., when trying to manually aligning volumes that initially do not overlap).

Libs/MRML/Logic/vtkMRMLSliceLogic.cxx
```diff
@@ -178,7 +178,14 @@ struct BlendPipeline
       {
         this->AddSubMath->SetOperationToSubtract();
       }
-      layers.emplace_back(this->AddSubAppendRGBA->GetOutputPort(), 1.0);
+      if (blendClipping)
```
jcfr, line 181:
```suggestion
      // If blend clipping is disabled, blending occurs over the entire extent
      // of all layers, not just within the background volume region.
      if (blendClipping)
```

jcfr:
> Layer blending is not disabled, the same blending (compositing) modes are still available

Thanks for clarifying. I now understand the follow-up comment  https://github.com/Slicer/Slicer/pull/7886#issuecomment-2288348468 

> UI

This means that by default, the following checkbox would be toggled by default:

Punzo:
> This means that by default, the following checkbox would be toggled by default

Yes!

Punzo:
> To Do:
> 
> 1. Rename: `BlendClipping` into `CropToBackgroundVolume`
> 2. if `CropToBackgroundVolume` is on,
>    thresholding is not applied for `Subtract` and `Add` in https://github.com/Slicer/Slicer/pull/7886/files#diff-d5498fd4b04ced370870a7176ba82f0d624ddcd269f72d27577fddf416f871b3R172-R188
> 3. Add/modify UI:
> 
> 

@lassoan @jcfr @jamesobutler  Done in https://github.com/Slicer/Slicer/pull/7886/commits/cabedb1ef2e3f79fd582f621b0b439ab95980795

Can you review, please? thanks!

video:
https://github.com/user-attachments/assets/33fb8ffc-d481-450b-9166-1acac3eab4b2

lassoan (commented):
Thanks for the updates! A few things still not fully work as expected when crop to background is disabled:
- when foreground opacity is low, the foreground image gets darker instead of fading to the background
- even though label opacity is set to 1.0, the label layer is semitransparent
- enabling/disabling crop to background is not propagated to other views when the views are linked

Punzo:
> Thanks for the updates! A few things still not fully work as expected when crop to background is disabled:
> - when foreground opacity is low, the foreground image gets darker instead of fading to the background
> - even though label opacity is set to 1.0, the label layer is semitransparent
> - enabling/disabling crop to background is not propagated to other views when the views are linked

Thanks for testing and reviewing:

1 and 2), I noticed them too and I also don't like it. It's because of the algorithm that uses the vtk filter in compound mode. I think the only solution is modifying the filter, always using normal mode and simply make the filter ignoring the zeros in the imageData of the first input (background) when blending if the cropToBackground is disabled. 

3) good catch, I forgot to test it. I'll fix it asap.

lassoan:
Modifying the filter be adding some new modes or options sounds like a good approach. We should be able to get the update integrated into VTK quickly.

Punzo:
> 3- good catch, I forgot to test it. I'll fix it asap.

Done in https://github.com/Slicer/Slicer/pull/7886/commits/5748ba4ec61a81585b0932db7987acc42025b7d1

Punzo:
> Modifying the filter be adding some new modes or options sounds like a good approach. We should be able to get the update integrated into VTK quickly.

Done in https://github.com/Slicer/Slicer/pull/7886/commits/36b504bfb3c1247fa4e5fd11b4705316313b0f58 and 
https://github.com/Punzo/VTK/commits/addAlphaBlendingToBlendFilter/
https://github.com/Punzo/VTK/commit/6a1f5a55175c4a7f1ca4efe9c934c9e7c15618e0

Now when crop to background is disabled, the blending behaviour is the same to the previous blending with cropping.

I don't have writing permission on Slicer VTK fork. @lassoan or @jcfr can you please update the VTK fork with 
https://github.com/Punzo/VTK/commit/6a1f5a55175c4a7f1ca4efe9c934c9e7c15618e0
https://github.com/Punzo/VTK/commit/80ccacd348bac7a90e30c53650e976f19ee8cec7

thanks in advance.

Punzo:
> Except exceptional urgent fixes, we always contribute changes to upstream VTK and then merge from there into Slicer's VTK fork. @davide, could you please submit a merge request to https://gitlab.kitware.com/vtk/vtk ? VTK developers will ask for adding a test that exercises this new flag, so I would recommend to include update of a test in the merge request.

PR in VTK opened at https://gitlab.kitware.com/vtk/vtk/-/merge_requests/11403

Punzo:
@jcfr @lassoan 

https://gitlab.kitware.com/vtk/vtk/-/merge_requests/11403 has been merged.

I don't have push permission to https://github.com/Slicer/VTK, so I have opened a PR:
https://github.com/Slicer/VTK/pull/59

and updated this PR to use the branch from https://github.com/Slicer/VTK/pull/59

Punzo:
https://github.com/Slicer/VTK/pull/59 has been merged

updated this PR to use last commit from [ branch from ](https://github.com/Slicer/VTK/commits/slicer-v9.2.20230607-1ff325c54-2/)
https://github.com/Slicer/VTK/commit/57bacae7e9d8613c1a5fc955028211c455f3787a
https://github.com/Slicer/Slicer/pull/7886/commits/00097fd508a27b59629265ec91f1503878b410d9

lassoan:
Thank you, it looks good. The only small thing that would be nice to change is to use "clip" word everywhere instead of sometimes using "clip" sometimes "crop". This would make things more consistent and usually "clip" is used in Slicer for temporary visualization option, while "crop" refers to changing data.

Punzo:
> Thank you, it looks good. The only small thing that would be nice to change is to use "clip" word everywhere instead of sometimes using "clip" sometimes "crop". This would make things more consistent and usually "clip" is used in Slicer for temporary visualization option, while "crop" refers to changing data.

I have renamed the title of this PR. In the code, I already use the word `crop` everywhere. Please let me know if I have missed any `clip` one.

Punzo:
> > Thank you, it looks good. The only small thing that would be nice to change is to use "clip" word everywhere instead of sometimes using "clip" sometimes "crop". This would make things more consistent and usually "clip" is used in Slicer for temporary visualization option, while "crop" refers to changing data.
> 
> I have renamed the title of this PR. In the code, I already use the word `crop` everywhere. Please let me know if I have missed any `clip` one.

I have renamed all `crop` to `clip` one in https://github.com/Slicer/Slicer/pull/7886/commits/467091abda02660d0dbb06bf27d5820775edcfe9

Punzo:
@lassoan @jcfr this PR is ready, but the CI (build) fails. I tested locally and I have built successfully. 
Does the CI rebuild also the superbuild? because it seems it does not use the latest Slicer VTK fork defined in this PR (i.e. https://github.com/Slicer/Slicer/pull/7886/files#diff-6a1489976309ff153180f305389a9b12e745c9ee7fefa7557e2ec92b741174c0R141).

Please let me know if I need to take any other action. Thanks!

lassoan (approved):
It all looks good, thank you. CI is expected to fail, as it only rebuilds Slicer, not the dependencies.

PR #7899, opened 2024-08-20
ENH: Reduce bundle temp directory length when saving scene mrb

Sunderlandkyl:
When saving a scene that includes sequence nodes, it is easy to go over the max path length on windows. This is due to saving the mrb scenes for the sequences inside the temp directory of the full scene. This means that the path will include the same bundle temporary directory string twice in the sequence mrb path.

This commit results in a ~55% length decrease in the bundle directory path.

Previously scene temporary directories were of the format:
```__BundleSaveTemp-2024-08-20_115507_655``` (38 characters)

Now the bundle directory format is:
```20240820115507655``` (17 characters)

### DIFF
--- a/Libs/MRML/Core/vtkMRMLScene.cxx
+++ b/Libs/MRML/Core/vtkMRMLScene.cxx
@@ -3858,6 +3858,20 @@ void vtkMRMLScene::TrimUndoStack()
 }
 
 //----------------------------------------------------------------------------
+std::string vtkMRMLScene::GetTemporaryBundleDirectory()
+{
+  std::stringstream ss;
+  ss << vtksys::SystemTools::GetCurrentDateTime("_tmp%Y%m%d");
+  const char validCharacters[] = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz";
+  int numberOfCharacters = sizeof(validCharacters) - 1;
+  for (int i = 0; i < 5; i++)
+  {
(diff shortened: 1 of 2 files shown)


pieper (dismissed):
Looks fine to me.  Thanks 👍

lassoan:
It could be useful to keep some hints in the folder name that it is a temporary folder for Slicer. To keep the length short, we could remove hhmmss.

For example something like _SlicerTmp20240820655 or _Tmp20240820655 or _Slicer20240820655 would be a bit more informative.

Sunderlandkyl:
If we are writing to the default location on Windows, the full path is something like: "C:/Users/Kyle/AppData/Local/slicer.org/Slicer/cache/SlicerIO/20240820145223141/", so we know that it is a Slicer file. I'm not sure about on macOS/Linux.

If we remove hhmmss, we might need to add some more checks to make sure that there is no duplicate folder in that directory, though it shouldn't be a problem as long as Slicer always deletes the temporary folder after zipping it.

lassoan:
Ok, we don't need to add the application name then. Something like _tmp... could be enough, and it is useful because if cleanup fails for any reason we know that it is safe to remove it. Something like _tmpbw would also tell that the folder was created for bundle writing, which can be useful for debugging. We could add a longer random suffix or use letters&numbers to avoid name clashes.

Sunderlandkyl:
Updated the format, now: _tmpYYMMDD + 5 random alphanumeric characters.

lassoan (approved):
Thank you, this looks good.

PR #7908, opened 2024-08-28
COMP: speed up Linux compilation by using precompiled headers

achataigner:
COMP: Speed up Linux compilation by using precompiled headers

This branch was only tested on Linux Ubuntu 22.04 LTS.

It is quite easy to speed up compilation in Libs and Base thanks to
vtk, itk and qt precompiled headers.

Build test done:
- DBUILD_TESTING=OFF cmake option
- with 12 jobs for the 12 cores of my intel i7 CPU
- commands:
  cd build-dir/Slicer-Build/Libs && make clean && make -j12
  cd build-dir/Slicer-Build/Base && make clean && make -j12

Build times:
  Libs:
    without precompiled headers: 44m 28s
    with precompiled headers: 18m9s

  Base :
    without precompiled headers: 10m 35s
    with precompiled headers: 2m 54s

This branch is simply a proposal to show the benefits of precompiled headers,
do not hesitate to ask questions and propose better code.

Precompiled headers are harder to use in Modules because the target library
 and the source binary must use the same compilation flags (for instance *_EXPORTS flag must be the same)

### DIFF
--- a/Base/CLI/CMakeLists.txt
+++ b/Base/CLI/CMakeLists.txt
@@ -107,6 +107,9 @@ if(SlicerBaseCLI_SRCS)
   set_target_properties(${lib_name} PROPERTIES LABELS ${lib_name})
   set_target_properties(${lib_name} PROPERTIES FOLDER "Core-Base")
 
+  target_precompile_headers(${lib_name} REUSE_FROM PrecompileItkHeaders)
+  target_precompile_headers(${lib_name} REUSE_FROM PrecompileVtkHeaders)
+
   # Apply user-defined properties to the library target.
   if(Slicer_LIBRARY_PROPERTIES)
     set_target_properties(${lib_name} PROPERTIES ${Slicer_LIBRARY_PROPERTIES})
--- a/Base/Logic/CMakeLists.txt
+++ b/Base/Logic/CMakeLists.txt
@@ -174,6 +174,9 @@ target_link_libraries(${lib_name} ${libs})
 set_target_properties(${lib_name} PROPERTIES LABELS ${lib_name})
 set_target_properties(${lib_name} PROPERTIES FOLDER "Core-Base")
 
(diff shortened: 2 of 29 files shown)


achataigner:
@lassoan @jcfr What do you think about this pull request ?

lassoan:
Thank you @achataigner, this looks like a potentially significant speedup!

My computer builds Slicer from scratch (including all dependencies) in about 30 minutes with gcc on Linux, but MSVC on Windows takes hours. So, I'm curious to see how much improvement can be achieved on Windows. I've started a build. I'll let you know when it is done.

lassoan:
I've built Slicer from scratch on Windows in release mode and got the following results:

Slicer main:
- build time: 5 hours 50 minutes
- build tree size: 13.9GB

With precompiled headers in this pull request:
- build time: 5 hours 36 minutes (4% faster)
- build tree size: 14.4GB (3.6% larger)

(I've also tried to speed up the build process on Windows by doing parallel build: `%CMAKE_EXE% --build . --config Release --parallel 16 -- /p:CL_MPcount=16` - with this the build time was 1 hour 47 minutes.)

Based on these results, I'm not sure if this is worth the trouble. But maybe just the proposed changes are ineffective on Windows.

achataigner:
This branch only speeds up Libs and Base build so I am not surprised by your results.
It would be interesting to measure a rebuild of Libs and a rebuild of Base to compare Windows and Linux results.
Naturally I am also curious about MacOS.

Again, this branch is a draft, we could use precompiled header in other libs than Libs and Base and get real improvements on the whole build. Moreover precompiled header is quite easy to use in CMakeLists.txt, it does not really complexify current CMake code.

@jcfr @RafaelPalomar @pieper  What are your opinions ?

achataigner:
I just started a Ninja build.
I've never used this build system before and it sounds promising, do you have any experience with it to build slicer ?

bao-ngy (approved):

PR #7926, opened 2024-09-10
DOC: Add scriptedregistration.md with an example rigid registration

koegl:
I realized there were no examples of (image) registration in the script repository, so I thought it would be helpful to create a section and add examples of scripting registrations.


lassoan (approved):
Thank you for your contribution! I've touched it up (e.g., added loading of sample input data, display fused result in the end, ...). It'll be merged as soon as automatic tests complete.

PR #7928, opened 2024-09-13
BUG: Fix reading of markup point position from integer value

lassoan:
When markup point position was defined as an integer (0) instead of float (0.0) the markup reader rejected the position value. The problem was that the code only accepted double type. Fixed by accepting any numeric type.

Fixes issue reported at https://discourse.slicer.org/t/one-control-point-out-of-many-is-read-incorrectly-into-slicer/38369


pieper (approved):
LGTM 👍

PR #7935, opened 2024-09-16
BUG: Skip disabled plugins when loading DICOM by scripting

### LINKED ISSUE #7931: DICOM module: Inconsistent checking of disabled plugins
## Summary

`DICOMBrowser` checks for user-disabled plugins by referring to `QSettings`. Whereas functions in `DICOMUtil` do not. This causes inconsistent behavior when loading DICOM data programmatically through DICOMUtil functions rather than using the user interface (DICOM Browser).

## Steps to reproduce
Particularly, loading dataset using `DICOMUtils.loadPatientByUID(patientUID)` doesn't work the same way as DICOM Browser user actions would.

There appears to be a discrepancy between [DICOMBrowser.py](https://github.com/Slicer/Slicer/blob/main/Modules/Scripted/DICOMLib/DICOMBrowser.py#L471) and [DICOMUtils.py](https://github.com/Slicer/Slicer/blob/main/Modules/Scripted/DICOMLib/DICOMUtils.py#L738) in how they generate the list of available plugins. DICOMUtils.py doesn't check QSettings for user preferences or disabled plugins. We need to move the logic to a common function so all paths will lead to correctly checking for disabled plugins.

## Environment
- Slicer version: https://github.com/Slicer/Slicer/commit/e8a49d6659898eb75ed0980367affd70c7715be6 (July 25, 2024)
- Operating system: _Windows 11 Pro_ 23H2 22635.4145

lassoan:
Make the default behavior of loading DICOM data consistent between using the GUI and using DICOMUtils Python scripting. Previously, loadSeriesByUID and loadByInstanceUID functions used all available DICOM plugins. Now only those DICOM plugins are used by default that are not disable in the application settings.

fixes #7931

I haven't tested it yet, just created this draft pull request to get quick feedback.

### DIFF
--- a/Modules/Scripted/DICOMLib/DICOMUtils.py
+++ b/Modules/Scripted/DICOMLib/DICOMUtils.py
@@ -24,9 +24,16 @@ comment = """
 
 
 # ------------------------------------------------------------------------------
-def loadPatientByUID(patientUID, messages=None, progressCallback=None):
+def loadPatientByUID(patientUID, messages=None, progressCallback=None, enabledPluginClassNames=None):
     """Load patient by patient UID from DICOM database.
-    Returns list of loaded node ids.
+    :param patientUID: UID of the patient in the database that will be loaded. Note that this is not a DICOM ID but an identifier
+                       created when the patient was added to the DICOM database.
+    :param messages: String list that will contain messages collected during DICOM examination and loading.
+    :param progressCallback: Callback function `progressCallback(pluginClassName, percentageCompleted)`
+                             that is repeatadly called during loading, if it returns True then loading is cancelled.
+    :param enabledPluginClassNames: List of DICOM plugin class names that are enabled for loading.
+                                    If set to `None` then default list is used (all plugins that are not explicitly disabled).
+    :return: List of loaded node ids.
 
     Example: load all data from a DICOM folder (using a temporary DICOM database)
 
@@ -78,7 +85,7 @@ def loadPatientByUID(patientUID, messages=None, progressCallback=None):
     if len(seriesUIDs) == 0:
         raise OSError("No series found in patient with DICOM database UID " + patientUIDstr)
 
-    return loadSeriesByUID(seriesUIDs, messages, progressCallback)
+    return loadSeriesByUID(seriesUIDs, messages, progressCallback, enabledPluginClassNames)
 
 
 # ------------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


pieper (dismissed):
Thanks for jumping in on this 👍 

That looks like the right thing.  It seems it does change behavior a bit and I was worried about changing the behavior but really I agree this is a bug and I doubt anyone is relying on the older behavior.

Modules/Scripted/DICOMLib/DICOMUtils.py
```diff
@@ -745,7 +769,7 @@ def getLoadablesFromFileLists(fileLists, pluginClassNames=None, messages=None, p
         return loadablesByPlugin, loadEnabled
 
     if pluginClassNames is None:
-        pluginClassNames = list(slicer.modules.dicomPlugins.keys())
+        pluginClassNames = getEnabledPlugins()
```
jadh4v, line 772: Wouldn't this break backward compatibility?
For example, previous calls to `DICOMUtils.loadSeriesByUID` will now exclude disabled plugins unless `pluginClassNames` are already provided when calling.

I suppose anyone using the function now has to update to
```py
DICOMUtils.loadSeriesByUID(seriesUIDs, pluginClassNames=list(slicer.modules.dicomPlugins.keys())
```

to restore the previous behavior.

Modules/Scripted/DICOMLib/DICOMUtils.py
```diff
@@ -735,6 +737,28 @@ def refreshDICOMWidget():
...
+        settings.endArray()
+
+    for pluginClass in slicer.modules.dicomPlugins:
+        if pluginClass not in disabledPlugins:
+            enabledPlugins.append(pluginClass)
+
+    return enabledPlugins
+
+
 def getLoadablesFromFileLists(fileLists, pluginClassNames=None, messages=None, progressCallback=None, pluginInstances=None):
```
jadh4v, line 762: Would make sense to rename `pluginClassNames` parameter to `selectedPlugins`

jadh4v (dismissed):
A few comments, but this looks great. Thanks.

Modules/Scripted/DICOMLib/DICOMUtils.py
lassoan: This was a bug (it was not specified in the documentation that even disabled plugins will be used). If someone wants to use all plugins then it is possible to specify a custom list in the newly added parameter.

Modules/Scripted/DICOMLib/DICOMUtils.py
lassoan: `pluginClassNames` is better in that it specifies that it expects class names (and for example not instances). Even if the new name was considerably better, it may not justify a breaking change (developers may use the argument name for clarity and robustness).

Maybe we should use `pluginClassNames` in the newly added arguments, too, to make it more clear if class names are expected?

Modules/Scripted/DICOMLib/DICOMUtils.py
jadh4v: Thanks. That makes sense.

Modules/Scripted/DICOMLib/DICOMUtils.py
jadh4v: Definitely makes sense to set the parameter name to `pluginClassNames` for the new changes to `loadSeriesByUID` and `loadByInstanceUID`. That way we also get consistent name at least within the file level.

PR #7940, opened 2024-09-17
BUG: Fix sitkUtils without MRMLIDImageIO

### LINKED ISSUE #7902: SitkUtils missing import os
## Summary

In Slicer 5.7 seems that sitkUtils.py file has been changed. 
Functions `PullFromSlicer `and `PushToSlicer `requires `os.remove`, but the import os is missing at the beginning of the file, causing an Name error: 

image

The error is raised at **line 34** and **line 56** in sitkUtil.py. 
It can be easily solved by adding an `import os` at the beginning of the file.

## Steps to reproduce

Detected with the following procedure: 
- Create a custom selector for "vtkMRMLLabelMapVolumeNode" in a scripted python module
- get the labelmap from a custom callback, retrieving the image through `sitkUtils.PullFromSlicer` function

## Environment
- Slicer version: Slicer-5.7.0-2024-06-17
- Operating system: Windows 11 Pro 10.0.22631

lassoan:
A fallback mechanism was implemented to keep sitkUtils functional when Slicer's SimpleITK is replaced by a stock SimpleITK (that does not have MRMLIDImageIO). However, this mechanism did not work due to a missing import.

The missing import has been added and a test is now added to check that sitkUtils works with and without MRMLIDImageIO.

fixes #7902

### DIFF
--- a/Base/Python/sitkUtils.py
+++ b/Base/Python/sitkUtils.py
@@ -1,5 +1,5 @@
 import SimpleITK as sitk
-
+import os
 import slicer
 
 __sitk__MRMLIDImageIO_Missing_Reported__ = False
--- a/Base/Python/tests/test_sitkUtils.py
+++ b/Base/Python/tests/test_sitkUtils.py
@@ -2,13 +2,21 @@ import slicer
 import sitkUtils as su
 
 import unittest
+import unittest.mock
 
 
 class SitkUtilsTests(unittest.TestCase):
     def setUp(self):
         pass
 
-    def test_SimpleITK_SlicerPushPull(self):
+    def test_SimpleITK_SlicerPushPullWithMRMLIDImageIO(self):
+        self._test_SimpleITK_SlicerPushPull()
+
+    def test_SimpleITK_SlicerPushPullWithoutMRMLIDImageIO(self):
+        with unittest.mock.patch.object(su, "IsMRMLIDImageIOAvailable", return_value=False):
+            self._test_SimpleITK_SlicerPushPull()
+
+    def _test_SimpleITK_SlicerPushPull(self):
         """Download the MRHead node"""
         import SampleData
 


Base/Python/sitkUtils.py
```diff
@@ -1,11 +1,11 @@
 import SimpleITK as sitk
-
+import os
 import slicer
 
 __sitk__MRMLIDImageIO_Missing_Reported__ = False
 
 
-def PushVolumeToSlicer(sitkimage, targetNode=None, name=None, className="vtkMRMLScalarVolumeNode"):
+def PushVolumeToSlicer(sitkimage, targetNode=None, name=None, className="vtkMRMLScalarVolumeNode", allowMRMLIDImageIO=True):
```
jcfr, line 8: Nitpick: Add missing docstring for `allowMRMLIDImageIO`

Base/Python/sitkUtils.py
```diff
@@ -36,11 +36,11 @@ def PushVolumeToSlicer(sitkimage, targetNode=None, name=None, className="vtkMRML
     return targetNode
 
 
-def PullVolumeFromSlicer(nodeObjectOrName):
+def PullVolumeFromSlicer(nodeObjectOrName, allowMRMLIDImageIO=True):
```
jcfr, line 39:
```
Nitpick: Add missing docstring for `allowMRMLIDImageIO`
```

jcfr (commented):
Instead of introducing a new parameter, I suggest to update the test  to monkey patch `IsMRMLIDImageIOAvailable` doing something like this:

```
with patch.object(su, 'IsMRMLIDImageIOAvailable', return_value=True):
    # 
```

See also https://docs.python.org/3/library/unittest.mock.html#quick-guide

lassoan:
I did not mind adding the extra parameter, as I thought it might be useful to be able to force disabling MRMLIDImageIO for other reasons (performance improvement, reproducibility, troubleshooting). But this `patch.object` would be perfect if we only need this for testing. To keep the API simpler, I'll switch to patching.

PR #7938, opened 2024-09-17
ENH: Improve and Customize URI Argument Handling for Custom Applications

jcfr:
Introduces two enhancements to streamline and customize how URI arguments are handled in custom Slicer applications:

* Add `qSlicerCoreApplication::handleURIArguments`.
* Add the ability to enable or disable post-startup URI argument handling.

Issue #7936, opened 2024-09-17
Streamline command-Line argument parsing for custom applications


jcfr:
## Is your feature request related to a problem? Please describe.

Custom applications leveraging Python-based command-line argument parsing via `ArgumentParser` often trigger debug messages like the following when unrecognized arguments are passed:

```
Ignore argument received via command-line (not a valid URL or existing local file): ...
```

This occurs because the custom argument parser receives arguments that Slicer does not recognize, generating unnecessary log messages.

For reference, custom parsing might be implemented as follows during application startup:

```python
parser = ArgumentParser()
parser.add_argument(...)
return parser.parse_args(list(slicer.app.commandOptions().unparsedArguments))
```

## Describe the solution you'd like

Fix support for standardized argument-handling conventions, such as using `--` (or `--ignore-rest`), would allow custom applications to cleanly separate arguments meant for Slicer from those meant for custom argument parsers. 

## Describe alternatives you've considered

An alternative approach could be to completely remove the debug message when arguments are parsed by a custom application, but this might limit visibility in certain debugging scenarios. 

## Additional context

https://github.com/Slicer/Slicer/blob/db8a5cfd4dda452c10d8d4e30fa3406c1ae38a94/Base/QTCore/qSlicerCoreApplication.cxx#L1142-L1174

jcfr:
> Fix support for standardized argument-handling conventions, such as using `--` (or `--ignore-rest`), would allow custom applications to cleanly separate arguments meant for Slicer from those meant for custom argument parsers.

After further consideration, the implementation in `ctkCommandLineParser::parseArguments`[^1] is sensible.

Instead, we should refactor the code to support disabling the handling of unparsed argument and move the corresponding code into an helper function callable from Python.

[^1]: https://github.com/commontk/CTK/blob/627842c05f262650bf50ff33f8fd31c7f330349a/Libs/Core/ctkCommandLineParser.cpp#L323C26-L424

PR #7941, opened 2024-09-17
BUG: Remove unneeded display of ignored arguments 

jcfr:
Unparsed arguments may be processed after the application startup (e.g., in scripts or custom applications). Always displaying the list of ignored arguments introduces unnecessary noise in the logs.

* https://github.com/Slicer/Slicer/pull/7938
* https://github.com/Slicer/Slicer/issues/7936

### DIFF
--- a/Base/QTCore/qSlicerCoreApplication.cxx
+++ b/Base/QTCore/qSlicerCoreApplication.cxx
@@ -1116,7 +1116,6 @@ void qSlicerCoreApplication::handlePreApplicationCommandLineArguments()
   if (options->ignoreRest())
   {
     qDebug() << "Ignored arguments:" << options->unparsedArguments();
-    return;
   }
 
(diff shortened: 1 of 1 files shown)


jamesobutler (approved):
Removal of ignored arguments log message is fine by me in my custom application as I don't use the command line options frequently. 

Logging ignored arguments does however seem relevant at the debug level of logging which is developer facing. For an end-user using a custom application, it would likely be logging at a higher application logging level where there would be no noise of debugging statements as it would likely be at the INFO or WARNING level.

jcfr:
Thanks for the review. Agreed, I will further improve so that it is only reported in the log files.

Alternatively we could also move the reporting in `slicerqt.py`, would that be enough ?

jadh4v (approved):
LGTM

jcfr:
I will re-introduce the logging of the the ignored argument so that it is only reported using debug level in a follow-up pull request.

PR #7942, opened 2024-09-18
BUG: Fix editing of transform component values in spinbox

### LINKED ISSUE #7574: Editing Decimal Values in Transform Fields Causes Introduction of Negative
## Summary

 Manually editing the fields next to the sliders in the transform module in local reference frame causes the introduction of an unexpected negative when editing the decimal values. Seems to be an issue between the local mode transform and the code that manages the number of decimal places displayed. See: https://discourse.slicer.org/t/manually-adjusting-transform-applied-object-seems-to-jump-around-matrixes-not-edited-as-expected/34086 

## Steps to reproduce

- Load MRHead, create and apply linear transform to it
- Switch to local transformation mode
- Click left of the decimal point in the LR translation slicer text box and type 1 => head moves to patient right, upper right element of matrix is 1
- Use arrow to move to the right of the decimal point and type 5 => head moves to patient left, upper right element is -1.5
- Type more 5s into and the matrix element gets more and more negative
- Setting value back to 0 results in a non-identity matrix

## Environment
- Slicer version: Slicer-5.6.1 (also happens on 5.2.2; notably known to NOT happen in 4.11)
- Operating system: Mac OS Monterey 12.7.1

lassoan:
In Transforms module, when rotate-first mode was enabled and edited translation components and the number of digits changed, the value that was being edited was reset. The problem was that when number of digits change then sibling widgets get a notification and they update their number of digits, too, which results in value change, which leads to resetting of all values (including the one that is being edited). Fixed by preventing resetting of the slider+spinbox for the axis that is being edited using the spinbox.

fixes #7574

---

<del>Draft pull request because it requires merging of https://github.com/commontk/CTK/pull/1219</del>
required CTK update was merged and git hash is updated in this pull request

### DIFF
--- a/Libs/MRML/Widgets/qMRMLTransformSliders.cxx
+++ b/Libs/MRML/Widgets/qMRMLTransformSliders.cxx
@@ -458,19 +458,19 @@ void qMRMLTransformSliders::resetUnactiveSliders()
 {
   Q_D(qMRMLTransformSliders);
 
-  if (!d->ActiveSliders.contains(d->LRSlider))
+  if (!d->ActiveSliders.contains(d->LRSlider) && !d->LRSlider->isSettingValueFromSpinBox())
   {
     bool blocked = d->LRSlider->blockSignals(true);
     d->LRSlider->reset();
     d->LRSlider->blockSignals(blocked);
   }
-  if (!d->ActiveSliders.contains(d->PASlider))
+  if (!d->ActiveSliders.contains(d->PASlider) && !d->PASlider->isSettingValueFromSpinBox())
   {
     bool blocked = d->PASlider->blockSignals(true);
     d->PASlider->reset();
     d->PASlider->blockSignals(blocked);
   }
-  if (!d->ActiveSliders.contains(d->ISSlider))
+  if (!d->ActiveSliders.contains(d->ISSlider) && !d->ISSlider->isSettingValueFromSpinBox())
   {
     bool blocked = d->ISSlider->blockSignals(true);
     d->ISSlider->reset();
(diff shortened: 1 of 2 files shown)


pieper (approved):
LGTM 👍

Libs/MRML/Widgets/qMRMLTransformSliders.cxx
```diff
@@ -458,19 +458,19 @@ void qMRMLTransformSliders::resetUnactiveSliders()
 {
   Q_D(qMRMLTransformSliders);
 
-  if (!d->ActiveSliders.contains(d->LRSlider))
+  if (!d->ActiveSliders.contains(d->LRSlider) && !d->LRSlider->isSettingValueFromSpinBox())
```
finetjul, line 461: isn't `QObject::sender()` sufficient ? (I did not think of it carefully)

lassoan: The direct sender is not that we need, because the series of events looks like this:
- user types one more digit into spinbox A
- number of digits is increased, so spinbox A emits a signal
- spinbox B (sibling widget of A) receives the signal that number of digits are modified and updates its value (the value is determined from slider B, rounded to the new number of digits)
- spinbox B value change triggers reset of all sliders/spinboxes to 0
- we should prevent reset of spinbox A; but the direct sender of the event is spinbox B

jcfr:
> required CTK update was merged and git hash is updated in this pull request

For future reference, see https://github.com/commontk/CTK/compare/627842c05f262650bf50ff33f8fd31c7f330349a...76e21221f81a4483d58f99a4d63d76f13718fb57

jcfr:
I am building & testing locally :hourglass_flowing_sand:

jcfr:
Local testing was :white_check_mark: 

Improvement of `ci.yml` to avoid failure like the one reported here will be suggested in a follow-up pull request.

In the meantime, I will bypass the check and move forward with the integration :rocket:

PR #7943, opened 2024-09-18
ENH: Re-introduce logging of unparsed arguments to log file only

jcfr:
Reintroduces the logging of unparsed command-line arguments, now restricted to the log file as debug entries. This allows developers to access the information for debugging purposes without cluttering the terminal output.

Related pull requests:
* https://github.com/Slicer/Slicer/pull/7941

### DIFF
--- a/Base/QTGUI/qSlicerApplication.cxx
+++ b/Base/QTGUI/qSlicerApplication.cxx
@@ -717,6 +717,14 @@ void qSlicerApplication::handleCommandLineArguments()
     this->errorLogModel()->disableAllMsgHandler();
   }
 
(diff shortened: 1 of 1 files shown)


jcfr:
Given arguments like the following:

```
./Slicer --launcher-verbose --another-arg 12 --path /tmp/this/does/not/exist -- --ignore-this
```

the message starting with `Ignored arguments: ` is only reported in the error log model.

In combination with setting `URIArgumentHandlingEnabled` to `false`, this will allow to nicely handle the case of custom application while reporting useful details in the log.


image

jadh4v (approved):
Looks good.

PR #7946, opened 2024-09-19
BUG: Fixed saving of labelmap volume sequences

### LINKED ISSUE #7124: Sequence of Labelmaps loading as Sequence of Scalar Volume Nodes After Scene Reload
## Summary

When creating a sequence of labelmap nodes, it works perfectly. But upon saving the scene and reloading it, the sequence of labelmaps is transformed into a sequence of scalar volume nodes.

## Steps to reproduce

Sample code:
```
import numpy as np

n = 5

SequenceNode = slicer.mrmlScene.AddNewNodeByClass("vtkMRMLSequenceNode", "sequence example") #Create sequence node

for i in range(n):
    labelmapArray = np.random.randint(4, size=(5,5,5)) #Create small sample with 4 categories
    labelmapNode = slicer.mrmlScene.AddNewNodeByClass("vtkMRMLLabelMapVolumeNode", f"labelmap_{i}") #Create labelmap node
    slicer.util.updateVolumeFromArray(labelmapNode, labelmapArray)
    SequenceNode.SetDataNodeAtValue(labelmapNode, str(i)) #Add labelmap to sequence

browserNode = slicer.mrmlScene.AddNewNodeByClass("vtkMRMLSequenceBrowserNode", "browser example") #Create browser
browserNode.SetAndObserveMasterSequenceNodeID(SequenceNode.GetID()) #Connect browser to sequence

slicer.util.setSliceViewerLayers(background=None, label=browserNode.GetProxyNode(SequenceNode)) #Show proxy node
slicer.modules.sequences.toolBar().setActiveBrowserNode(browserNode)
slicer.modules.sequences.setToolBarVisible(True)
```
1. Run the sample script or create manually a sequence of labemaps.
2. Interact with the sequence toolbar to check if it is working correctly with labelmap.
3. Save and close the scene.
4. Reload the scene and interact with toolbar again. A new scalar proxy node will be created instead of changing the previous proxy.

## Expected behavior

The sequence data type and proxy is correctly loaded as labelmap instead of scalar volume. 

## Environment
- Operating system: Windows 10 x64

lassoan:
Labelmap volume sequences were saved the same way as scalar volume sequences, therefore when a labelmap volume sequence was loaded from file, the proxy nodes were all scalar volumes. Fixes by adding "intent: label" field in the .seq.nrrd file header (inspired by NIFTI file format's intent tag).

fixes #7124

### DIFF
--- a/Docs/developer_guide/modules/index.md
+++ b/Docs/developer_guide/modules/index.md
@@ -12,6 +12,7 @@ plots
 selftests
 segmentations
 segmenteditor
+sequences
 transforms
 volumerendering
 volumes
--- a/Docs/developer_guide/modules/sequences.md
+++ b/Docs/developer_guide/modules/sequences.md
@@ -1,0 +1,49 @@
+# Sequences
+
+## Volume sequence file format (.seq.nrrd)
+
+Whenever it is possible (when origin, spacing, axis directions, and extents are the same in all sequence items), a sequence of volumes is stored in a standard NRRD image file with custom fields to store additional metadata.
+
+### Image data
+
+The image is always stored in 3 spatial dimensions. Even if an image consists of a single slice, the spatial dimension must be still 3 because that is required for specification origin, spacing, and axis directions in 3D space. The file has an additional `list` dimension.
+
+### Metadata
+
+#### Standard fields
+
+- `dimension`: `4`
+- `sizes`: number of time points, followed by image size (I, J, K)
+- `space directions`: `none`, followed by I, J, K axis directions in physical space
+- `kinds`: `list domain domain domain`
+- `labels`: name of the sequence index (e.g., `"time"`), followed by ` "" "" ""`
+- other standard fields (`type`, `space`, `endian`, `encoding`, `space origin`, etc.) can have any value
+
+#### Custom fields
+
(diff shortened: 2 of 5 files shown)


pieper (approved):
LGTM 👍 

But shouldn't we also have a document somewhere that defines the header extensions for seq.nrrd?  (Like the ones for [seg.nrrd](https://slicer.readthedocs.io/en/latest/developer_guide/modules/segmentations.html#segmentation-labelmap-file-format-seg-nrrd)).  If it exists I didn't find it.

lassoan:
> But shouldn't we also have a document somewhere that defines the header extensions for seq.nrrd? (Like the ones for [seg.nrrd](https://slicer.readthedocs.io/en/latest/developer_guide/modules/segmentations.html#segmentation-labelmap-file-format-seg-nrrd)). If it exists I didn't find it.

What an excellent idea! I've created a documentation page to describe custom fields and while writing it I realized that we already store `DataNodeClassName` as a custom attribute in .seq.nrrd files that we can use. This way seq.nrrd writing did not need any change, only the reader needed a small improvement to use this field for determining what MRML node class to instantiate for the items in the sequence.

PR #7961, opened 2024-09-24
ENH: Allow vtkMRMLTextStorageNode to store any text files

lassoan:
Until now, vtkMRMLTextStorageNode could only store files with txt, xml, and json extensions. Even if a filename was set to something else by a file reader/writer plugin, the storage node can override the filename, for example when saving the scene to .mrb format.

This limitation is now lifted, a custom list of file extensions can be specified for both reading and writing.

This will be used for allowing script storage in the scene - see https://github.com/SlicerMorph/SlicerEditor/issues/44


pieper (approved):
LGTM - 👍 

Thanks for working on this!

PR #7962, opened 2024-09-25
BUG: Fix segmentation Show 3D button staying disabled

lassoan:
"Show 3D" button enabled/disabled state was not updated immediately when the button's "locked" state was updated. The problem was that only a member variable was set, but the button update was not triggered.

fixes https://github.com/Project-MONAI/MONAILabel/issues/1624


cpinter:
Looks good to me. I don't have the right to approve but since I already reviewed it...

lassoan:
Thanks for the review!

PR #7967, opened 2024-09-27
Improve file format detection

lassoan:
Improves automatic detection of segmentation and NIFTI transform files.

### DIFF
--- a/Modules/Loadable/Transforms/qSlicerTransformsReader.h
+++ b/Modules/Loadable/Transforms/qSlicerTransformsReader.h
@@ -47,6 +47,13 @@ public:
 
   bool load(const IOProperties& properties) override;
 
+  /// Returns a positive number (>0) if the reader can load this file.
+  /// It uses default confidence value except for NIFTI files.
+  /// For NIFTI file that contain displacement field, the method returns 0.6 which is
+  /// higher than the default (0.5) to make the application prefer reading it as transform.
+  double canLoadFileConfidence(const QString& file)const override;
+
+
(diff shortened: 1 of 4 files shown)


Modules/Loadable/Transforms/qSlicerTransformsReader.h
```diff
@@ -47,6 +47,13 @@ class qSlicerTransformsReader
 
   bool load(const IOProperties& properties) override;
 
+  /// Returns a positive number (>0) if the reader can load this file.
+  /// It uses default confidence value except for NIFTI files.
+  /// For NIFTI file that contain displacement field, the method returns
+  /// higher than default (0.6) to make the application prefer reading it as transform.
```
jcfr, line 53:
```suggestion
  /// For NIFTI file that contain displacement field, the method returns 0.6 which is
  /// higher than default (0.5) to make the application prefer reading it as transform.
```

pieper (approved):
This looks great 👍 

I don't think it's needed for this, but someday if we really care about covering nifti more 'correct' uses of nifti we should assemble a set of 'correct' example files along with common variants that come up in practice.  But for now just dealing with cases as they come up seems good enough.

Did you want to include a link to the discourse thread about SynthMorph which I guess inspired this?

lassoan:
Thanks for the reviews. I've accepted Jcfr's change and added link and description of NIFTI displacement field format as Steve suggested.

pieper (approved):
LGTM 👍

PR #7975, opened 2024-10-04
ENH: Add interaction widget axis labels for markups

cpinter:
Usage:
m = slicer.mrmlScene.AddNewNodeByClass('vtkMRMLMarkupsFiducialNode') m.AddControlPointWorld(0,0,0)
m.SetAxisLabels('X', 'Y', 'Z')
m.GetDisplayNode().SetHandlesInteractive(True)

### DIFF
--- a/Libs/MRML/DisplayableManager/vtkMRMLInteractionWidgetRepresentation.cxx
+++ b/Libs/MRML/DisplayableManager/vtkMRMLInteractionWidgetRepresentation.cxx
@@ -544,6 +544,8 @@ void vtkMRMLInteractionWidgetRepresentation::UpdateHandlePolyData()
     this->Pipeline->Append->RemoveAllInputs();
   }
 
+  vtkNew<vtkTransform> axisLabelTransform;
+
   HandleInfoList handleInfoList = this->GetHandleInfoList();
   for (HandleInfo handleInfo : handleInfoList)
   {
@@ -580,6 +582,7 @@ void vtkMRMLInteractionWidgetRepresentation::UpdateHandlePolyData()
 
     transform->Identity();
     vtkDoubleArray* orientationArray = vtkDoubleArray::SafeDownCast(handlePolyData->GetPointData()->GetArray("orientation"));
+    vtkNew<vtkMatrix4x4> orientationMatrix;
     if (orientationArray)
     {
       double orientation[9] = {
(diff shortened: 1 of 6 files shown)


cpinter:
@Sunderlandkyl You may remember that I asked about axis labels for the interaction widgets a while ago, but it was never integrated into Slicer, because it lacked certain functions and then it got forgotten. We needed this functionality with the new interaction widgets you implemented, and I made it work (your quick answer yesterday helped a lot to finish). Can you please take a look and let me know what may be missing for this to be integrated?

Here's a video about how it looks:
https://github.com/user-attachments/assets/16ac3367-31f5-43c3-b855-e47398b5f3cc

It looks good to me, except for one thing. I couldn't get the text properties from the markup to be propagated to the axis labels, see https://github.com/Slicer/Slicer/pull/7975/files#diff-9b4e55420768a6c4ec0e82307c94048f76f2d8e7f278cae12344b4b055badebeR271  (and https://github.com/Slicer/Slicer/pull/7975/files#diff-eb8c3eb226772bfbd7f81c26ca0c10239a7c7260688123bcc49d09a895780964R700, this makes the labels appear, but with constant parameters).

Thanks a lot for any feedback!

@xriobe @lassoan

Sunderlandkyl (commented):
Strange that only the color is being applied via text property.

A couple of comments:
- AxisLabels isn't included in WriteXML/ReadXML/CopyContent/PrintSelf.
- Should the AxisLabels be included in the display node instead? (I'm not sure which is the best location).
- Should this PR be expanded to include transform handles as well?

cpinter:
Thanks for the feedback, Kyle! Yes there are some missing pieces for sure, it's good you made note of some of them.

It is up for debate where the actual text is stored. I started rebasing the old implementation and for now kept it in the markups node, but it's true that now it could be in the displayable node (subclass of transformable). No problem with moving it there.

Do you have any comment about why setting the text property from `vtkSlicerMarkupsInteractionWidgetRepresentation::UpdateInteractionPipeline` does not have any effect? I think that is a blocking issue before this could be integrated (the rest are small things or decisions to make). Thank you!

cpinter:
Hi @Sunderlandkyl ! Maybe you haven't seen the last comment as I didn't tag you. Can you please look at that issue why the fonts properties cannot be updated in the function I mention above, if you have a little time? I'd greatly appreciate it as it would be a large step towards integrating it. (We'll decide where to define the strings and other issues later, and I'll make the changes.) Thank you very much!

Sunderlandkyl:
I haven't had a chance to look into why the text property isn't updating correctly, and it's not immediately obvious to me from the code what the issue is. I'll need to compile the changes and investigate.

cpinter:
Thanks @Sunderlandkyl ! This is already useful. OK we'll try to do some more investigation. If you happen to have an idea in the meantime please let us know.

PR #7976, opened 2024-10-04
ENH: Improve terminology category selection

cpinter:
- Many users were confused and annoyed by the single-click multi-selection. This changes that style to ExtendedSelection (single-click change selection, Ctrl or Shift modifies selection)
- Add convenience button for selecting all categories again (although Ctrl+A does it as well but it is not obvious)


cpinter:
Here's a video about how it looks. Note that the clicks are not visible in the video, and Ctrl and Shift modifiers are also used to achieve multi-selection
https://github.com/user-attachments/assets/3da1e94c-dc40-420e-b409-365fc22a5783

@muratmaga

lassoan (approved):
Thank you, it looks good!

Issue #6975, opened 2023-05-16
Improvements to the Terminologies module. 


muratmaga:
## Is your feature request related to a problem? Please describe.

Currently both in Data module and segment editor double clicking the segment name allows renaming the segment in place. Clicking the color of the segment bring the Terminologies window, in which the name of the segment can also be changed. This is a confusing behavior and not conducive to using the terminologies more frequently. 

## Describe the solution you'd like

Now that Terminologies module is seeing some improvement, to encourage its usage I suggest changing the renaming behavior in segment editor.  See discussion here. https://discourse.slicer.org/t/terminologies-questions/27384/11?u=muratmaga

- [x] Renaming behavior (double click on segment name brings the navigator, like clicking on color). If it is possible to do this in python, than it can be SlicerMorph customization before we propose to the community ([PR#7983](https://github.com/Slicer/Slicer/pull/7983) issued -> superseded by [PR#8011](https://github.com/Slicer/Slicer/pull/8011), later fix in https://github.com/Slicer/Slicer/pull/8071)
- [x] Allow user to create custom terminology
    - [x] Export color table from terminology (possibly related to https://github.com/Slicer/Slicer/issues/7593)
- [x] Making the change in the terminology selection persistent. So if I choose Uberon, next time it should start with uberon
- [x] Highlight/selection confusion. Easier way to select/deselect things.
- [ ] https://discourse.slicer.org/t/segment-editor-clicking-eye-to-toggle-visibility-should-not-select-segment/20237

muratmaga:
adding this thread for discussions: https://discourse.slicer.org/t/improving-the-usability-of-terminology-module/38501
1. renaming behavior change
2. allow user to create custom terminology
3. Export color table from terminology (possibly related to https://github.com/Slicer/Slicer/issues/7593)

We need these changes for MorphoDepot. 

@cpinter will be working on these, but would be good to obtain some consensus.... @lassoan @pieper

muratmaga:
Also, we need a behavior for user to change the default terminology. I have part of uberon imported, but every time I try to use the terminology module in changing a segment name, I have to click on the dropdown and change it from Segmentation category and type to Uberon.

cpinter:
Another list from recent email of @muratmaga:

4. Making the change in the terminology selection persistent. So if I choose Uberon, next time it should start with uberon
5. Highlight/selection confusion. Easier way to select/deselect things.
[...]

PR #7983, opened 2024-10-08
ENH: Add setting for editing segment name in segments table using ter…

cpinter:
…minology

Re #6975

### DIFF
--- a/Modules/Loadable/Segmentations/Resources/UI/qSlicerSegmentationsSettingsPanel.ui
+++ b/Modules/Loadable/Segmentations/Resources/UI/qSlicerSegmentationsSettingsPanel.ui
@@ -37,23 +37,10 @@
      </property>
     </widget>
    </item>
-   <item row="5" column="0" colspan="2">
-    <spacer name="verticalSpacer">
-     <property name="orientation">
-      <enum>Qt::Vertical</enum>
-     </property>
-     <property name="sizeHint" stdset="0">
-      <size>
-       <width>20</width>
-       <height>40</height>
-      </size>
-     </property>
-    </spacer>
-   </item>
    <item row="1" column="0">
     <widget class="QLabel" name="SurfaceSmoothingLabel">
      <property name="toolTip">
-      <string>Automatically set opacities of the segments when loading from file based on which contains which, so that no segment obscures another</string>
+      <string>Enable surface smoothing during binary labelmap to closed surface conversion in new segmentations. Smoothing improves appearance in 3D views and exported models but makes segment editing considerably slower.</string>
      </property>
      <property name="text">
       <string>Enable surface smoothing by default:</string>
@@ -74,13 +61,36 @@
     </widget>
    </item>
    <item row="2" column="0">
+    <widget class="QLabel" name="SegmentNameTerminologyEditLabel">
+     <property name="toolTip">
+      <string>If enabled, double-clicking the segment name for editing in the segment tables opens the terminology selector. Otherwise the name can be simply edited as text. False by default.</string>
+     </property>
+     <property name="text">
+      <string>Edit segment name as terminology:</string>
+     </property>
+    </widget>
+   </item>
+   <item row="2" column="1">
+    <widget class="QCheckBox" name="SegmentNameTerminologyEditCheckBox">
+     <property name="toolTip">
+      <string>If enabled, double-clicking the segment name for editing in the segment tables opens the terminology selector. Otherwise the name can be simply edited as text. False by default.</string>
+     </property>
+     <property name="text">
+      <string/>
+     </property>
+    </widget>
+   </item>
+   <item row="3" column="0">
     <widget class="QLabel" name="label">
+     <property name="toolTip">
+      <string>This terminology will be used by default for new segments in an empty segmentation.</string>
+     </property>
      <property name="text">
       <string>Default terminology entry:</string>
      </property>
     </widget>
    </item>
-   <item row="2" column="1">
+   <item row="3" column="1">
     <widget class="ctkPushButton" name="EditDefaultTerminologyEntryPushButton">
      <property name="toolTip">
       <string>This terminology will be used by default for new segments in an empty segmentation.</string>
@@ -91,28 +101,51 @@
     </widget>
    </item>
    <item row="4" column="0">
+    <widget class="QLabel" name="label_3">
+     <property name="toolTip">
+      <string>Default mode of editing in areas of other segments.</string>
+     </property>
+     <property name="text">
+      <string>Default overwrite mode:</string>
+     </property>
+    </widget>
+   </item>
+   <item row="4" column="1">
+    <widget class="ctkComboBox" name="DefaultOverwriteModeComboBox">
+     <property name="toolTip">
+      <string>Default mode of editing in areas of other segments.</string>
+     </property>
+    </widget>
+   </item>
+   <item row="5" column="0">
     <widget class="QLabel" name="label_2">
+     <property name="toolTip">
+      <string>This option controls what the application should do if the user edits a segment that is currently not visible. It is meant to prevent unintentional changes to hidden segments.</string>
+     </property>
      <property name="text">
       <string>Edit hidden segments: </string>
      </property>
     </widget>
    </item>
-   <item row="4" column="1">
+   <item row="5" column="1">
     <widget class="ctkComboBox" name="AllowEditingHiddenSegmentComboBox">
      <property name="toolTip">
       <string>This option controls what the application should do if the user edits a segment that is currently not visible. It is meant to prevent unintentional changes to hidden segments.</string>
      </property>
     </widget>
    </item>
-   <item row="3" column="0">
-    <widget class="QLabel" name="label_3">
-     <property name="text">
-      <string>Default overwrite mode:</string>
+   <item row="6" column="0" colspan="2">
+    <spacer name="verticalSpacer">
+     <property name="orientation">
+      <enum>Qt::Vertical</enum>
      </property>
-    </widget>
-   </item>
-   <item row="3" column="1">
-    <widget class="ctkComboBox" name="DefaultOverwriteModeComboBox"/>
+     <property name="sizeHint" stdset="0">
+      <size>
+       <width>20</width>
+       <height>40</height>
+      </size>
+     </property>
+    </spacer>
    </item>
   </layout>
  </widget>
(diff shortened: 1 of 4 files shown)


cpinter:
https://github.com/user-attachments/assets/f430f606-cf63-450a-89fe-fb2eea5b6cf3

The only inconvenience is that the application needs to be restarted after changing the setting. Not sure if we should add a warning about this, and if so, where. Applying the setting immediately does not seem feasible, as there may be many segment tables in the application, and the delegates are set on creation.

cpinter:
Can we integrate this @lassoan ?

@muratmaga

lassoan:
If changing the application settings only takes effect after restart then at least we should mark the setting as it requires restart (if it hasn't marked as such already). Then, if you change the setting then a message is displayed that changing of this setting requires application restart and you are offered the option to restart.

However, this can be quite inconvenient and I'm not sure how would we enforce users to enable this new option.

I think a better approach would be not to use the application settings to customize this behavior, but make it a property of the segment list (and/or segment editor widget). Then each segment editor widget could decide how to set that option. For example, the default Slicer Editor module could use the application settings to customize the behavior, But the MophoDepot Editor would always set up the segment list to bring up the terminology selector with a specific color table or terminology. The MorphoDepot Editor would could also set up many other options and provide other convenience features (e.g., it could ask the user to log in, then it would automatically fetch the next image that has incomplete tasks, and once the user finished with the segmentation it could upload the results).

muratmaga:
> However, this can be quite inconvenient and I'm not sure how would we enforce users to enable this new option.

I think this solution was a compromise (since Csaba was expecting a push back about modifying the rename behavior from the current one) and it is meant to be experimental period to get feedback about this behavior. In SlicerMorph we would do this default behavior through the customizations.

We dont really have MorphoDepotEditor, we will be using the existing functionality and modules as is.

lassoan:
The Segment Editor is just a very short Python script to set up a Segment Editor widget for general purpose usage. The Segment Editor widget is intended to be embedded in specialized modules and it is used auccessfully in many modules.

Since it would be so easy to create a MorphoDepot Editor module and it would address this and many other future customization needs, I think it would be a better solution right now and also in the long term.

Proposed next steps:
- update this PR to just add a Boolean property to the Segment Editor widget that controls the double-click behavior
- clone and rename the Segment Editor module  to MorphoDepot Editor and set the new flag

muratmaga:
MorphoDepot is very specific use case. It is meant for classroom/workshop deployment and meant to make both the students and instructors life easier in working on segmentations (and other collaborative tasks). Most SlicerMorph users are standalone users and will continue to use the SegmentEditor, and may not touch this module. So I don't think this is a good (or at least general enough) solution to the problem this change is addressing. 

Current renaming behavior using terminology navigator is awkward. (You click on color of the segment to change its name from the terminology menu, as opposed to the segment name, which is the natural behavior). This awkwardness has came up multiple times on the forum, particularly from people who were expecting a color palette (like it does in powerpoint or any other application) instead of complex dialog box with names. 

Maybe I am overlooking some technical difficulties unknown to me, but I am not sure why there is such a push back about this behavior change. Since it only add one more click to the current manual renaming behavior, and makes the obscure terminologies module more prominent. My suggestion is to put this as a  application setting (or a segment editor setting, if there is such thing) for people to test and then after sometime we can make a decision about whether to make it permanent or keep the current behavior.

jamesobutler:
> for people to test and then after sometime we can make a decision about whether to make it permanent or keep the current behavior.

It would be good to add updated documentation describing the new setting so that individuals could point other users to it.

https://slicer.readthedocs.io/en/latest/user_guide/modules/segmentations.html

jamesobutler:
I would wish that “tests” like this would be in a Slicer channel only available in non-stable releases of the application so that there isn’t back-and-forth change in available functionality. Going forward could we put these test type changes behind a statement checking the Slicer release type to not display or use the setting if “Stable” type? Could be in “Preview” and “Experimental” release types.

lassoan:
> Maybe I am overlooking some technical difficulties unknown to me, but I am not sure why there is such a push back about this behavior change.

There are several technical issues with the implementation that is currently in the pull request:
- this option is not discoverable
- the option cannot be changed at runtime but you need to restart the application for it to take effect
- the option is global: if the user changes the application settings then it changes the behavior in all Segment Editor widgets in all extensions, which may not be always desirable (segment editor widget is embedded in many modules in various extensions and the behavior change could be unexpected/undesired)

There are separate solutions for each issue, such as:
- move the option somewhere near the segment list to make it discoverable (maybe it could be an item in the right-click menu)
- change how behavior of double-click is remapped (it would be a small design change: do not use a delegate for capturing double-click event, just [connect a slot to the double-click signal of the table view](https://stackoverflow.com/questions/4324005/how-to-detect-doubleclick-in-qtableview) and then open the terminology popup or call `QAbstractItemView::edit` depending on the current application setting)
- add custom `SettingsKey` for each Segment Editor widget to avoid forcing the same setting to be applied globally

Instead of implementing all these, I thought it would be simpler to add a customized Segment Editor module for MorphoDepot that sets this new option, because then the feature does not have to be discovered; no restart is needed (because the widget is updated when the Qt property is changed); the option is not global anymore (only set to the customized editor). Having a customized editor for MorphoDepot seemed inevitable to me, as it is a very specific use case, with some special constraints and features, so users would benefit a lot from an optimized editing environment.

> Current renaming behavior using terminology navigator is awkward. (You click on color of the segment to change its name from the terminology menu, as opposed to the segment name, which is the natural behavior)

I agree, this is inconsistent that double-clicking the color selector brings up the terminology popup (instead of a direct color editor) but double-clicking starts direct text editing (instead of bringing up the terminology popup). 

It would be great to fix this, but then the solution must cover the color selector as well (introduce an option like "Use standard terminology for segmentation", which would enable/disable using the terminology selector for both color and name editing); and we would also need to address the 3 technical issues listed above.

lassoan:
> I would wish that “tests” like this would be in a Slicer channel only available in non-stable releases of the application so that there isn’t back-and-forth change in available functionality. Going forward could we put these test type changes behind a statement checking the Slicer release type to not display or use the setting if “Stable” type? Could be in “Preview” and “Experimental” release types.

Each pull request could automatically provide a "special build" created from the content of that pull request, which could be downloaded as installation package or docker image and it would allow users to test it. For quick testing, it could be also possible to run the docker image in the web browser so that users would not even need to install it (using Binder or similar service).

muratmaga:
> It would be great to fix this, but then the solution must cover the color selector as well (introduce an option like "Use standard terminology for segmentation", which would enable/disable using the terminology selector for both color and name editing); and we would also need to address the 3 technical issues listed above.

That's what I want too!

@cpinter would these changes take considerable time?

pieper:
>  I thought it would be simpler to add a customized Segment Editor module for MorphoDepot

I doubt we'd do this - the MorphoDepot concept is more about organizing and reviewing teamwork.  Once they are working they would use Slicer/SlicerMorph as normal, so terminology features should be orthogonal.  MorphoDepot would just automate setups to make tasks more convenient.

pieper:
Yes, right now the goal of MorphoDepot is to avoid using the github UI.  Only the backend data and work task management would use github, while the actual segmentation or other tasks would rely on users learning Slicer.  But everything is subject to evolution as people actually start using it.

cpinter:
Hi all! Sorry for the late response - I was travelling.

> the option cannot be changed at runtime but you need to restart the application for it to take effect

Yes as you say this is the result of using the delegate. I will look at the event handling alternative you mentioned.

> this option is not discoverable

I think there is no need for this to be easily discoverable, until there is more consensus about the use of terminologies. This change was supposed to facilitate this particular use case in a way that does not affect the rest of the users, and also without introducing much new complexity.

> the option is global: if the user changes the application settings then it changes the behavior in all Segment Editor widgets in all extensions, which may not be always desirable (segment editor widget is embedded in many modules in various extensions and the behavior change could be unexpected/undesired)

I think this is subjective and I don't necessarily agree. Renaming and color changing are such basic operations, that I cannot imagine that developers wanted to rely on the exact functioning of these. So I cannot imagine anything "going wrong" in extensions, since these are atomic small feaures. Besides, I think once someone changes this, they will want to have the change everywhere. Changing certain editors and not others would be confusing (=many accidental clicks), a bit like driving on the left in some parts of a country but on the right in others. One gets used to how a certain tool works...

> It would be great to fix this, but then the solution must cover the color selector as well (introduce an option like "Use standard terminology for segmentation", which would enable/disable using the terminology selector for both color and name editing); and we would also need to address the 3 technical issues listed above.

This could be a good option. However, I would remove the third item (different settings for different editors - see my previous comment).

> @cpinter would these changes take considerable time?

I don't think so. Adding more visibility for the option in the right-click menu should be easy (few hours). I'll look at the suggeston for the second point @lassoan provided, but it should not be more than several hours, a dozen at the most I think.

lassoan:
>>this option is not discoverable

> I think there is no need for this to be easily discoverable, until there is more consensus about the use of terminologies. This change was supposed to facilitate this particular use case in a way that does not affect the rest of the users, and also without introducing much new complexity.

I'm confident that this is a useful feature for all segment list widgets, so we'll need to find a good place for it. Instead of putting the switch to some temporary location (and annoy users later by changing it), we should be able to decide on a location that works in the long term.

A checkbox in the context menu of the segment list could be a good location. It would not just be easier to discover but also address the problem of requiring restart and would be more compatible with allowing custom behavior in segment list and segment editor widgets embedded in various modules.

>> the option is global: if the user changes the application settings then it changes the behavior in all Segment Editor widgets in all extensions, which may not be always desirable

> I think this is subjective and I don't necessarily agree.

There is no need for agreement in this case. If one developer asks to allow the setting to be changed without affecting the main application then the single global flag is not sufficient anymore. I ask for the customizable setting, because we would use that for example in SlicerHeart Valve Segmentation module (we would want to encourage users to use standard names for leaflets without messing with the behavior of the Segment Editor module).

The solution could be trivial: just add a `settingskey` (or `terminologySelectorSettingsKey`) member to the segment list. Segment Editor module and Segmentations module (and also a checkbox in application settings / segmentations, if we want that) would use the same `settingsKey` value. Modules that embed a segment list or segment editor widget could decide to use the same settingsKey as the Segmentations module or use custom setting. If the `settingsKey` is set to empty then there would be no persistence of the option (application settings would not be used in that widget).

cpinter:
> I'm confident that this is a useful feature for all segment list widgets, so we'll need to find a good place for it.

Agreed. If we do this change as you suggested then without a doubt we need to make it more visible. Agreed with the location as well.

> The solution could be trivial: just add a settingskey

I think I understand. In the end @muratmaga needs to decide, but it seems that we have an easy way into Slicer with this if we do these three changes.

cpinter:
FYI the decision is to proceed with the plan of @lassoan . I'll create a new PR in a few days and then close this as superseded. Thanks everyone for the useful discussion!

lassoan:
Sounds good, thank you.

cpinter:
Superseded by https://github.com/Slicer/Slicer/pull/8011

Issue #6159, opened 2022-02-05
Introduce "focused" node concept


lassoan:
## Is your feature request related to a problem? Please describe.

Current unmet needs:
- Cross-reference nodes (and components) between module GUI in the left panel and viewers. For example, if the user clicks on a model in the data tree then it is hard to know which node in the scene it corresponds to.
- Prevent accidental moving of move markup control points. When the users want to rotate the 3D view then often accidentally move a control point.
- Cluttered views due to interaction arrows. Interaction arrows are shown for all nodes, which clutter slice and 3D views.

## Describe the solution you'd like

Introduce the concept of "focused" node and component.

- Component is optional and can be used to identify a control point in a markups node or a segment in a segmentation node.
- Only one node and component in the scene can have the focus. (In contrast, multiple nodes can be "selected".)
- A node can get the focus by the user left-clicking on it in a slice or view.
- A node can get the focus when the user clicks on it in a module GUI. For example, if the user clicks a model in the data tree then that model could get the focus.
- The focused node is displayed somewhat differently: with a highlight, outline, or bounding box - in all views.
- When a node is focused it may display more details. For example it can show node name, interaction arrows, node name, etc.

## Additional context

https://github.com/Slicer/Slicer/issues/5061

jcfr:
For reference, latest work-in-progress from @Sunderlandkyl is organized at https://github.com/Sunderlandkyl/Slicer/tree/focus_pw_41

PR #7985, opened 2024-10-08
ENH: Remove "Active" prefix from some node selection comboboxes

jamesobutler:
As discussed at Slicer hangout today.

There can be confusion that changing the selection of a combobox with label that has the prefix "Active" that it will apply changes to the scene in some way. For examplethere can be confusion that "Active Volume" in the Volume module would change which volume is showing in the layout views, however that is not the case. Changing "ActiveVolume" does not update the selection node using SetActiveVolumeID. Instead this label is just for selecting the volume node for manipulating parameters within the Volumes module widget area only.

| `main` | This PR |
|--------|----------|
|{1FBD831B-10D1-4753-8F13-484BC0BC780C}|{9E4716B9-9B67-4427-AA24-A30448783FB2}|
|{4136E032-F2B2-44CD-8471-6291582CEA4D}|{DFA22558-5BB9-4B80-98AB-AF186D9D3E41}|
|{9BF96458-71E3-4722-B3A8-15B3F919A542}|{BAB6801D-C447-47F2-BFDA-55ADB88A254B}|

Other places in `main` already did not use the "Active" prefix or they used a subject hierarchy tree view instead of a label+combobox:
{EC9A74DD-C93D-4A0C-B1CD-D8987B5274A7}

### DIFF
--- a/Docs/user_guide/modules/segmentations.md
+++ b/Docs/user_guide/modules/segmentations.md
@@ -163,7 +163,7 @@ See Script repository's [Segmentations section](../../developer_guide/script_rep
     - Set visibility and per-segment display settings, opacity, color, segment name
 - Display
     - Segmentations-wide display settings (not per-segment!): visibility, opacity (will be multiplied with per-segment opacity for display)
-    - Views: Individual views to show the active segmentation in
+    - Views: Individual views to show the selected segmentation in
     - Slice intersection thickness
     - Representation in 3D/2D views: The representation to be shown in the 3D and 2D views. Useful if there are multiple representations available, for example if we want to show the closed surface in the 3D view but the labelmap in the slice views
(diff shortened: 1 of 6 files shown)


pieper (approved):
LGTM 👍

lassoan:
For the record, the naming problably comes from Slicer2, where there was an "active" node of each type (e.g., active volume node, active label volume, ...), stored in the "selection" node. Probably the node selector in these modules actually changed the active node, but this is no longer the case (it is only done for markups now), so it was justified to remove the confusing "Active" word.

We can revisit the notion of active/focused/selected nodes when @Sunderlandkyl's [node focus developments](https://github.com/Slicer/Slicer/issues/6159) get integrated.

Issue #7984, opened 2024-10-08
Center slices button not working for first non-null layer


jamesobutler:
## Summary

It appears that the behavior of centering slices to the first non-null layer is no longer a thing as of https://github.com/Slicer/Slicer/commit/a5535f773b4c83556de1191f72173bddec4bf185. @Punzo or @lassoan is this expected behavior or should it still be working as it did in 5.6.2. It is unexpected to me that this change would occur as I believed that commit to be mainly about the alpha blending and what volumes are shown, but that in the case of just a foreground volume and not background volume that the center slices button would update the FOV to that single volume in the foreground layer.

https://github.com/Slicer/Slicer/commit/a5535f773b4c83556de1191f72173bddec4bf185#diff-cf360adf7aa72c06a3cacb1abdbfe85fb5e23f38b0b57a6e97d9bb0fd92c8369L260

https://github.com/Slicer/Slicer/commit/a5535f773b4c83556de1191f72173bddec4bf185#diff-8158e9ae3c4dff65dacd710311fc5c5774037fdd01482129fef60e70b0cfa1a6L2055-R2059
https://github.com/Slicer/Slicer/blob/a5535f773b4c83556de1191f72173bddec4bf185/Libs/MRML/Widgets/qMRMLSliceControllerWidget.cxx#L2052-L2059

## Steps to reproduce

1. Launch Slicer
2. Load MRHead
3. With MRHead in the background layer zoom out and then press Center Slices button in slice viewer and observe that it appropriately updates the FOV.
4. Change the MRHead to the foreground volume layer, change the zoom and press the center slices button and observe that the FOV is not changed.

## Environment
- Slicer version: Slicer-5.7.0-2024-10-07, likely started as of https://github.com/Slicer/Slicer/commit/a5535f773b4c83556de1191f72173bddec4bf185
- Operating system: Windows\

Punzo:
Thanks James for reporting the issue.

If I recall, the idea was that if both a background and foreground volumes are selected and the blending option is on, the centering has to fit only on the background volume extension.

But I agree in the case only the foreground volume (or label map) is selected, the centering should fit the slice view on the foreground. I will open a PR tomorrow.

Punzo:
Fixed in https://github.com/Slicer/Slicer/pull/7986

lassoan:
Associated pull request has been merged.

PR #7986, opened 2024-10-08
BUG: Fix fit slice view if no background volume is selected

Punzo:
Fix regression reported in

https://github.com/Slicer/Slicer/issues/7984

@jamesobutler @lassoan


jamesobutler (approved):
Thanks for the quick fix! I've included these changes in my local build and now it is centering to the volume as expected when there is only a foreground volume layer specified.

Issue #6621, opened 2022-10-27
Transform subject hierachy still visible


riep:
## Summary
The transfom subject hierarchy plugin is always visible even after having black listed it.

## Steps to reproduce
-In python terminal:
```
     dataModuleWidget = slicer.modules.data.widgetRepresentation()
     treeView = slicer.util.findChildren(dataModuleWidget, className='qMRMLSubjectHierarchyTreeView')[0]
     treeView.setPluginWhitelist([''])
```
-Load a volume 
-Go to the subject hierarchy tree of the data module
-Right click on the transform icon of a volume node loaded in the application. 
-Transform menu of the transform plugin is visible 

## Expected behavior
Nothing should be visible on the right click 

## Environment
- Slicer version: Slicer-5.0.3
- Operating system: Linux 20.04

riep:
would you accept 
`      else if (sourceIndex.column() == this->model()->transformColumn())
        {
        vtkIdType itemID = d->SortFilterModel->subjectHierarchyItemFromIndex(index);
        if (itemID) // Valid item is needed for transform actions
          {
          foreach (qSlicerSubjectHierarchyAbstractPlugin *plugin, d->enabledPlugins())
            {
            QString pluginName = plugin->name();

            if (d->SelectedItems.size() > 0 && pluginName == "Transforms")
              {
              this->populateTransformContextMenuForItem(itemID);
              d->TransformMenu->exec(globalPoint);
              return;
              }
            }
          }
        }`

lassoan:
Please send a pull request with the proposed change, that would make it much easier to review and test. Thank you!

PR #7997, opened 2024-10-17
ENH: Add a transform plugin in subject hierarchy tree view

riep:
Transform related functions were originally not included into a specific plugin. Thus some functionalities related to the management of plugins were not available.
Fix https://github.com/Slicer/Slicer/issues/6621

### DIFF
--- a/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyTreeView.cxx
+++ b/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyTreeView.cxx
@@ -85,15 +85,16 @@ public:
 
   virtual void init();
 
-  /// Setup all actions for tree view
-  void setupActions();
+  /// Update list of menu actions in all context menus
+  void updateMenuActions();
+
+  void updateSceneMenuActions();
+  void updateVisibilityMenuActions();
+  void updateNodeMenuActions();
+  void updateTransformMenuActions();
 
   /// Get list of enabled plugins \sa PluginAllowlist \sa PluginBlocklist
   QList<qSlicerSubjectHierarchyAbstractPlugin*> enabledPlugins();
-
-  void applyTransformToItem(vtkIdType itemID, const char* transformNodeID);
-  vtkMRMLTransformNode* appliedTransformToItem(vtkIdType itemID, bool& commonToAllChildren);
-  vtkMRMLTransformNode* firstAppliedTransformToSelectedItems();
 
   void setSubjectHierarchyItemVisibility(vtkIdType itemID, VisibilityAction visibilityAction);
 
@@ -126,16 +127,9 @@ public:
   QActionGroup* SelectPluginActionGroup;
   QMenu* SceneMenu;
   QMenu* VisibilityMenu;
+  QMenu* TransformMenu;
   QStringList PluginAllowlist;
   QStringList PluginBlocklist;
-
-  QMenu* TransformMenu;
-  QAction* TransformInteractionInViewAction;
-  QAction* TransformEditPropertiesAction;
-  QAction* TransformHardenAction;
-  QAction* CreateNewTransformAction;
-  QAction* NoTransformAction;
-  QActionGroup* TransformActionGroup;
 
   /// Subject hierarchy node
   vtkWeakPointer<vtkMRMLSubjectHierarchyNode> SubjectHierarchyNode;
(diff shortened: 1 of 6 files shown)


riep:
Hi all, 
I was a bit confuse about the formatting that slicer is using. Can you help me on that please ?

cpinter:
Before I do a thorough review, can you please
1. Summarize the change in a bit more detail in the commit message. It facilitates future maintenance to have meaningful and informative commit messages that contain more than just the few dozen characters allowed in the first line
2. Remove the style changes. It makes the review much harder (lots of unrelated changes), and much of it seems subjective. Style changes typically arrive in large commits with automatic formatting. Let's focus on the functional part here not the style. If you worry about the style, then I suggest doing it in a separate PR.

Thank you!

lassoan:
> Remove the style changes. 

I agree, I wanted to review the changes, but all the whitespace changes made it much more difficult to see the actual changes. If you want to make style changes then you can put them in a separate commit.

jcfr:

While the following GitHub feature may help ignore the white space changes, it would be helpful to avoid introducing newlines and alike.

image

It will be idea to explain the new API introduces, and how those are used.

cpinter:
In any case the comment of @riep saying "I was a bit confused about the formatting that slicer is using" is completely valid, as I also am not sure about the exact guidelines. Some of his changes seem to go from one acceptable style to another, so I'd not integrate those. Unless there is a clear guideline and the change makes the code conform with that.

riep:
Dear all,
Modifications done regarding formatting. @cpinter  I agree some changes are not essential. I just like to apply a formatting tool and do not think about it while coding. 
There are no additional functional added in this PR. It is only code moving related to transform actions from qMRMLSubjectHierarchyTreeView to a new dedicated plugin.
The behavior MUST be unchanged.

lassoan:
Thanks for the update @riep. I'm refactoring the code to be a bit more consistent with other plugins and will push the proposed changes to this branch soon.

lassoan:
I've finished with the refactoring to make this new plugin more conform to other plugins. It was not trivial, as this plugin was a bit special in that it generates menu items quite dynamically (based what transform nodes are available in the scene).

@riep please test if this works well for you and if you have any comments or suggestions.

riep:
This seams to work fine.

While testing I saw a bug:
Load a volume > apply a transform to the loaded volume > check interaction to display the handler > right click on the volume and select "None" as transform

**after these steps the handler is still visible which is unexpected**

then right click on the volume and select a transform > check interaction to display the handler

**after these step a second handler is visible for the volume**

This is probably not related to this task, if you confirm I can open a bug

lassoan:
> Load a volume > apply a transform to the loaded volume > check interaction to display the handler > right click on the volume and select "None" as transform
after these steps the handler is still visible which is unexpected
then right click on the volume and select a transform > check interaction to display the handler
after these step a second handler is visible for the volume

This is the intended behavior. "Interaction" checkbox includes a convenience feature that it creates a parent transform if the node does not have a parent transform already.

The "Interaction" word may not be ideal (it might suggest that it is a property of the selected node) and overall transform interactions can be complex. Let us know if you have any specific suggestion how to make the behavior more intuitive.

riep:
> > Load a volume > apply a transform to the loaded volume > check interaction to display the handler > right click on the volume and select "None" as transform
> > after these steps the handler is still visible which is unexpected
> > then right click on the volume and select a transform > check interaction to display the handler
> > after these step a second handler is visible for the volume
> 
> This is the intended behavior. "Interaction" checkbox includes a convenience feature that it creates a parent transform if the node does not have a parent transform already.
> 
> The "Interaction" word may not be ideal (it might suggest that it is a property of the selected node) and overall transform interactions can be complex. Let us know if you have any specific suggestion how to make the behavior more intuitive.


> > Load a volume > apply a transform to the loaded volume > check interaction to display the handler > right click on the volume and select "None" as transform
> > after these steps the handler is still visible which is unexpected
> > then right click on the volume and select a transform > check interaction to display the handler
> > after these step a second handler is visible for the volume
> 
> This is the intended behavior. "Interaction" checkbox includes a convenience feature that it creates a parent transform if the node does not have a parent transform already.
> 
> The "Interaction" word may not be ideal (it might suggest that it is a property of the selected node) and overall transform interactions can be complex. Let us know if you have any specific suggestion how to make the behavior more intuitive.

Yes you are right, I understood the feature.
One issue, we now have to possibility to show the interaction/ interaction option when right clicking on the "visible" and "color" column which is not as

lassoan:
We are trying to find a balance between making available features at convenient places, but not to have too complex GUI with too many options exposed.

Let us know if you have any specific suggestions on how to tweak the GUI.

riep:
Current behavior:
"Interaction" check box can be checked by right clicking on "visibility", "color" and "transform" columns
When checking "interaction" by right clicking on "visibility" and  "color" , "interaction options" appears BUT NOT for "transform" column

What I would expect:
"Interaction" check box AND "interaction options" are only visible by right clicking on "transform" column

riep:
> Current behavior: "Interaction" check box can be checked by right clicking on "visibility", "color" and "transform" columns When checking "interaction" by right clicking on "visibility" and "color" , "interaction options" appears BUT NOT for "transform" column
> 
> What I would expect: "Interaction" check box AND "interaction options" are only visible by right clicking on "transform" column

WELL, I think I understand the idea behind this choice. I wont explain what I understood :) ... but I am fine with this.
Thanks for you help Andras !

lassoan:
Good catch, we should not have two transform plugins! I'll merge the two.

If it is not too difficult to add interaction options to the transform column then I'll add it there, too, for consistency.

lassoan (approved):
I've merged the two transform subject hierarchy plugins. It should be ready to merge.
Tweaking of the menu items did not fit into this pull request but would make sense to do it later.

lassoan:
Thanks for the note, I've removed the unnecessary empty lines in https://github.com/Slicer/Slicer/pull/8022.

PR #7998, opened 2024-10-17
BUG: add missing OPTIONS request handler when CORS enabled

Gabsha:
This PR adds an OPTIONS request handler in the Webserver module.

This handler creates a response to preflight requests when CORS is enabled, which is required when executing cross-origin requests.

The response content was inspired from MSDN documentation at:
https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request

Relevant forum discussion:
https://discourse.slicer.org/t/using-webserver-dicomweb-interface-with-ohif-cors-enabled/39736/2

---

To test ->

1. Run a separate OHIF instance (not the built-in one) with the following datasource item in appConfig.js

```js
{...
datasource: [
{
      friendlyName: 'Slicer 3D',
      namespace: '@ohif/extension-default.dataSourcesModule.dicomweb',
      sourceName: 'slicer',
      configuration: {
        name: 'Slicer',
        wadoUriRoot: 'http://localhost:2016/dicom',
        qidoRoot: 'http://localhost:2016/dicom',
        wadoRoot: 'http://localhost:2016/dicom',
        qidoSupportsIncludeField: true,
        imageRendering: 'wadouri',
        thumbnailRendering: 'wadouri',
        enableStudyLazyLoad: true,
        supportsFuzzyMatching: true,
      },
    },
]}
```
2. In slicer, select Webserver module, advanced options, check "Enable CORS".
3. Start server
4. Navigate to your OHIF instance (eg. `http://localhost:3000?dataSources=slicer`)
5. Load a series.

Disabling CORS and restarting the server should prevent loading the series.

### DIFF
--- a/Modules/Scripted/WebServer/WebServer.py
+++ b/Modules/Scripted/WebServer/WebServer.py
@@ -418,7 +418,7 @@ class SlicerHTTPServer(HTTPServer):
                     self.logMessage("Warning, we don't speak %s", version)
                     return
 
-                methods = ["GET", "POST", "PUT", "DELETE"]
+                methods = ["GET", "POST", "PUT", "DELETE", "OPTIONS"]
                 if method not in methods:
                     self.logMessage("Warning, we only handle %s" % methods)
                     return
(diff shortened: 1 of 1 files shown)


lassoan:
This is very helpful thank you! Would it make sense to only respond to these requests if CORS is enabled in Slicer?

lassoan:
Please fix lint an commit message check errors. Let us know if you need any explanation or help.

Gabsha:
> This is very helpful thank you! Would it make sense to only respond to these requests if CORS is enabled in Slicer?

I think that it makes sense to respond even if CORS are not enabled. By responding, but without relevant Headers, the browser will be able to log a message that cross-origin requests are not allowed, and will not issue the subsequent GET. By not responding at all, the browser will just hang with pending requests.... only after timing out will it log a message.

pieper (approved):
LGTM 👍

lassoan (approved):

jcfr:
Thanks @Gabsha for the contribution :pray: :100:

PR #8000, opened 2024-10-19
Fix too frequent display of scene modification confirmation popup

lassoan:
When closing the scene or exiting the application a confirmation popup is often displayed that warns that the scene content is changed.

Often this popup did not seem justified (e.g., the scene has just been saved) and there was no way of telling why the popup was displayed (what exactly was modified).

This pull request adds a "Show details" option that can display the list of all modified nodes that should be saved in data files or the scene file:

image

The pull request also contains a few fixes in detection of node modifications to avoid showing confirmation popups unnecessarily.


pieper (approved):
I like it! 👍

PR #8003, opened 2024-10-23
ENH: Make terminology code finding more robust

### LINKED ISSUE #8002: Excessive warnings reported about terminlologies
## Summary

When loading a seg.nrrd file with different terms there is a lot printed at the terminal.

For example:

```
Generic Warning: In vtkSlicerTerminologiesModuleLogic.cxx, line 244
GetCategoryArrayInTerminology: Failed to find terminology root for context name 'Segmentation category and type - MONAI Auto3DSeg'


Generic Warning: In vtkSlicerTerminologiesModuleLogic.cxx, line 274
GetCategoryInTerminology: Failed to find category array in terminology 'Segmentation category and type - MONAI Auto3DSeg'


GetCategoryInTerminology: Failed to find category 'Anatomical Structure' in terminology 'Segmentation category and type - MONAI Auto3DSeg'


DeserializeTerminologyEntry: Failed to get terminology category


DeserializeTerminologyEntry: Failed to get terminology type
```

## Steps to reproduce

Use Auto3DSeg on any scan and save the result in seg.nrrd and reload.  Also happens for TotalSegmenter results converted from DICOM Seg.

## Environment
- Slicer version: Slicer-5.6.2 and 5.7.0-2024-1014
- Operating system: tested on liniux but probably happens anywhere.

lassoan:
This commit makes terminology code finding more robust and removes unnecessary error messages when loading a segmentation.

- If terminology or anatomic context name is not found then we look for the codes in all available contexts.
- If codes are not available anywhere then we do not display an error when the segmentation is loaded. Only when the unavailable code is attempted to be shown in the terminology selector.
- Made behavior of vtkSlicerModuleLogic::DeserializeTerminologyEntry more clear: now it just parses the terminology string. To get metadata of a terminology entry from all loaded terminology contexts, the vtkSlicerModuleLogic::UpdateEntryFromLoadedTerminologies method can be called.
- When segmentation contained unknown segmentation conversion parameters then an error was logged and the parameter was discarded. Now unknown parameters are not discarded (to avoid unnecessary data loss) and logging of error message is no longer needed.

fixes #8002

### DIFF
--- a/Libs/vtkSegmentationCore/vtkSegmentationConverter.cxx
+++ b/Libs/vtkSegmentationCore/vtkSegmentationConverter.cxx
@@ -49,6 +49,7 @@ vtkStandardNewMacro(vtkSegmentationConverter);
 vtkSegmentationConverter::vtkSegmentationConverter()
 {
   // Get default converter rules from factory
+  this->CustomConversionParameters = vtkSmartPointer<vtkSegmentationConversionParameters>::New();
   vtkSegmentationConverterFactory::GetInstance()->CopyConverterRules(this->ConverterRules);
   this->RebuildRulesGraph();
 }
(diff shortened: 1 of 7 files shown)


cpinter:
Thank you very much Andras, this is really useful! All the code I wrote contain too much error logging, so it makes sense to remove some (the same with SH). The rationale was that we can catch early bugs easier this way. However, some errors were logged in harmless situations and they have not been removed. So thanks.

Is there anything practical to keep in mind after the introduction of `UpdateEntryFromLoadedTerminologies` ?

pieper (approved):
Thanks for looking into this 👍 

I don't know this part of the code well but the changes look good and it's fine with me as long as @cpinter is okay with it.

lassoan:
> Is there anything practical to keep in mind after the introduction of UpdateEntryFromLoadedTerminologies ?

vtkSlicerModuleLogic::DeserializeTerminologyEntry did two things in the past: parse the terminology string into a terminology entry + update the terminology entry with metadata found in the available contexts. Now these are in two different methods (DeserializeTerminologyEntry does the parsing, UpdateEntryFromLoadedTerminology gets updated metadata), because most of the time we only needed the parsing. There was only one place where we actually wanted to look up additional metadata based on the codes, and there UpdateEntryFromLoadedContexts had to be called after DeserializeTerminologyEntry.

Overall, I don't expect that this change impacts any code outside of this pull request, but if you had extensions that called DeserializeTerminologyEntry then you can check if you need to call UpdateEntryFromLoadedContexts after that.

cpinter:
Thanks @lassoan for the explanation. So the Update function gets the rest of the fields from the context(s) based on the parsed category/type identifiers. Maybe I was confused about the name. Would Populate be better? Anyway I'm fine with keeping this as is!

PR #8004, opened 2024-10-23
Remove Slicer-application specific stuff from Libs, Base and Loadable modules

AlexyPellegrini:
This is the first PR for the future SlicerLibrary support. This PR contains small changes that doesn't break anything in slicer and should be mergeable at any time.

This basically remove all hardcoded paths that only make sense for the Slicer application from the Base/, Libs /and Modules/Loadable/[Logic|MRML|MRMLDM|vtkWidgets] modules so they can used as libraries and still have the right behavior as long as the caller gives them the right informations. To do this, I removed all inclusions of `vtkSlicerConfiguration.h` from these modules!

This enables all these modules to be built the same way outside and inside the Slicer application, which will be important for SlicerLibrary! 

I want to integrate the changes one at a time to limit the size and complexity of the future SlicerLibrary Pull Request :) Next on the list is VTK 9.4 support.

Please review @jcfr @lassoan 
FYI: @Thibault-Pelletier @finetjul


lassoan:
Thank you, I'm very happy that we are beginning this work here! Overall the changes look good, there is only one thing that bothers me: using system environment variables for configuring the application. This could be quite fragile, as environment variables are global and very visible and can be changed by anyone at anytime in an uncontrollable manner. Could you think about a better, more controllable alternative for storing application configuration?

For example, we could have a json configuration file for simple key:value storage, implemented in a Slicer or MRML logic class (MRML is probably better, as there are some configuration options at MRML level and then we could use the same at both MRML and Slicer level). When the application is started, it would initialize some ApplicationConfiguration object from the config file at application startup.

AlexyPellegrini:
> Overall the changes look good, there is only one thing that bothers me: using system environment variables for configuring the application. This could be quite fragile, as environment variables are global and very visible and can be changed by anyone at anytime in an uncontrollable manner. Could you think about a better, more controllable alternative for storing application configuration?

I agree that environment variables are fragile, the first approach I tried was to add a member to vtkColorModuleLogic, and initialize it in qSlicerColorsModule::createLogic:
```cpp
  auto* logic = vtkSlicerColorLogic::New();
  qSlicerApplication * app = qSlicerApplication::application();
  logic->SetSlicerShareDirectory(app->slicerSharePath().toUtf8().toStdString());
```
Does this seem go to you?

lassoan:
Various module logic classes may need various application configuration information. Therefore, it would be simpler to just set the application logic object in the module logic object. Then the module logic object could obtain any information it needs from the application logic. I think we always set the application logic in module logic in the C++ Slicer application.

Base/Logic/vtkSlicerApplicationLogic.cxx
```diff
@@ -711,26 +702,28 @@ bool vtkSlicerApplicationLogic::IsEmbeddedModule(const std::string& filePath,
...
-  (void)slicerRevision;
-#endif
+
   return isEmbedded;
 }
 
 //----------------------------------------------------------------------------
 bool vtkSlicerApplicationLogic::IsPluginInstalled(const std::string& filePath,
-                                                  const std::string& applicationHomeDir)
+                                                  const std::string& applicationHomeDir,
```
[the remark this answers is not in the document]
AlexyPellegrini: these functions are static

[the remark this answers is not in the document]
lassoan: I see. Then we have two options: either pass more variables (as implemented in this PR already) or make it a class method. However, considering that this method would now take 4 arguments that all come from the application logic object, making it a method of the application logic class may be more appropriate.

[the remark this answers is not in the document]
AlexyPellegrini: I would require too much time, I will keep it this way.

Modules/Loadable/Colors/Logic/vtkSlicerColorLogic.cxx
```diff
@@ -104,11 +101,20 @@ std::vector<std::string> vtkSlicerColorLogic::FindDefaultColorFiles()
   {
     slicerHome = std::string(vtksys::SystemTools::GetEnv("SLICER_HOME"));
   }
+
+  const char *slicerShareDir = vtksys::SystemTools::GetEnv("SLICER_SHARE_DIR");
```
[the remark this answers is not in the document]
AlexyPellegrini: Done. I introduces more changes to properly integrate this change. I will let you review again.

AlexyPellegrini:
I removed all remaining references to the SLICER_HOME and SLICER_SHARE_DIR variables and put these information in vtkMRMLApplicationLogic instead. I also found out that vtkMRMLLogic class can now be removed, since it was only used for this.
I updated Colors modules initialization and tests to fix them.

AlexyPellegrini:
@lassoan @jcfr

Hi everyone,
Congratulations on the Slicer 5.8 release!

 I rebased this PR which is the first step for slicerlib support!

FYI @finetjul @Thibault-Pelletier

AlexyPellegrini:
@lassoan This is the first PR to look at if you got time!

jcfr:
Thanks for taking the time to put this pull request 🙏✨️

We will test locally and report back. Note that we may shuffle things around in term  of commit history.

finetjul (approved):
Minor comments, otherwise LGTM

Modules/Loadable/Colors/qSlicerColorsModule.cxx
```diff
@@ -107,9 +107,11 @@ void qSlicerColorsModule::setup()
     return;
   }
   vtkSlicerColorLogic* colorLogic = vtkSlicerColorLogic::SafeDownCast(this->logic());
-  if (this->appLogic() != nullptr)
+  vtkSlicerApplicationLogic* appLogic = this->appLogic();
+  if (appLogic != nullptr)
   {
-    this->appLogic()->SetColorLogic(colorLogic);
+    colorLogic->SetMRMLApplicationLogic(appLogic);
```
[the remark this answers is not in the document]
AlexyPellegrini: it will never be, and would break existing code below if it could so I don't think that will be useful

finetjul (dismissed):
LGTM

lassoan:
We need the vtkMRMLLogic class. It contains useful utility functions that may be used by extensions and will be probably used in Slicer core. Some MRML library classes need application home and share folders. I agree that getting the value from environment variable was hacky, but it could get it instead from some singleton class that stores MRML-library-level settings (kind of like QSettings, but for MRML). Those settings could be set by the application at startup.

AlexyPellegrini:
With current changes, the share and home directory are stored within the vtkMRMLApplicationLogic, that should be easily reachable by any module logic that need them. c.f. [vtkSlicerColorLogic.cxx](https://github.com/Slicer/Slicer/pull/8004/files#diff-cd82f3a788a5bd3088c226d939e47944394811f8411969e36ac83810ef5224ea).

lassoan:
`vtkMRMLApplicationLogic` is not accessible from `MRML/Core`, so it is not a trivial change, but if we can avoid the need to access the home and share folders from `MRML/Core` then the moving should OK. Actually, it would be cleaner if MRML core does not need to know where these application folders are, so let's see if this is feasible without breaking things.

History: CLI modules and classes in `MRML/Core` used to rely on the `SLICER_HOME` environment variable to get the application home and share folders. We then introduced `vtkMRMLLogic` to centralize the mechanism of getting these folders (so that we can switch from using environment variables to something cleaner and more controllable). It is in `MRML/Core` to make sure it is accessible everywhere where MRML is used.

Current state: I've checked uses of the application home and share folders (via `vtkMRMLLogic` class and `SLICER_HOME`) and fortunately only a few uses left. Your PR changes Slicer core, so that's good. Use in FreeSurfer color node was changed some time ago to get the shared folder from the module logic. `CMFreg` extension still needs to access the home folder, but it uses the old `SLICER_HOME` mechanism, which will still work (and I think this extension is not used that much anyway), so it can be dealt with later. `SwissSkullStripper` uses `itksys::SystemTools::FindProgramPath` to find the CLI executable location and finds the module's shared data folder by relative path from that, which is not elegant, but workable (and usable for other CLI modules that need resource files).

Overall, the proposed change would not break anything now and it is cleaner design (less coupling) if CLI modules and MRML core classes do not access application settings or resource files, so let's proceed with moving of these methods to the MRML logic. If we later find the access is needed then we can implement a mechanism.

---

The scene cleanup helper functions are still useful, but of course we don't need to maintain a separate class for them. Probably the best new home would be the `vtkMRMLScene` class.

jcfr:
Waiting we come to consensus regarding the following issue, I will keep the conditional building of CLI capabilities:
* https://github.com/Slicer/Slicer/issues/8477

AlexyPellegrini:
Now that we resolved the main point of https://github.com/Slicer/Slicer/issues/8477, I will rebase this branch on main.

AlexyPellegrini:
@jcfr @lassoan Can we move this forward?

finetjul:
> @jcfr @lassoan Can we move this forward?

Isn't this PR blocked by https://github.com/Slicer/Slicer/issues/8477 ?

AlexyPellegrini:
No, only the first part is necessary, the other part can be done at any time, imho

jcfr:
Now that this pull request has been rebased, we will be able to further iterate. 

Thanks for your patience.

jcfr:
@AlexyPellegrini @finetjul  It would be great if you could build and test the following pull request focusing on fixing the logic initialization: 
* https://github.com/Slicer/Slicer/pull/8644

AlexyPellegrini:
This bug only shows up when the `SLICER_SHARE_DIR` and `SLICER_HOME_DIR` macros are no longer used. The Colors module logic never received the app logic, but never accessed it either.
Note that the commit changes alone won't build on master since they reference `vtkMRMLApplicationLogic::[Get|Set][Share|Home]Directory` that is introduced by the first commit (6e5ea5d9).
We could split the commit into two parts, one that correctly propagate the app logic, and one that properly initialized it with the shared and home directory and keep it here.

jcfr:
Thanks for the review, the pull request (https://github.com/Slicer/Slicer/pull/8644) extracting the relevant changes has been updated. I would appreciate if you could help get it finalized :pray:

Modules/Loadable/Colors/Logic/vtkSlicerColorLogic.cxx
```diff
@@ -88,27 +88,28 @@ const char* vtkSlicerColorLogic::GetDefaultChartColorNodeID()
 std::vector<std::string> vtkSlicerColorLogic::FindDefaultColorFiles()
 {
   // get the slicer home dir
-  std::string slicerHome;
-  if (vtksys::SystemTools::GetEnv("SLICER_HOME") == nullptr)
+  auto* app = this->GetMRMLApplicationLogic();
```
jcfr:
```suggestion
  vtkSlicerApplicationLogic* appLogic = this->GetMRMLApplicationLogic();
```

Modules/Loadable/Colors/Logic/vtkSlicerColorLogic.cxx
```diff
@@ -88,27 +88,28 @@ const char* vtkSlicerColorLogic::GetDefaultChartColorNodeID()
...
+  if (!app)
```
jcfr:
```suggestion
  if (!appLogic)
```

Modules/Loadable/Colors/Logic/vtkSlicerColorLogic.cxx
```diff
@@ -88,27 +88,28 @@ const char* vtkSlicerColorLogic::GetDefaultChartColorNodeID()
...
-    else
-    {
-      slicerHome = std::string("");
-    }
+    vtkErrorMacro("GetMRMLApplicationLogic() must not be null");
+    return {};
   }
-  else
+
+  const std::string& shareDir = app->GetShareDirectory();
```
jcfr:
```suggestion
  const std::string& shareDir = appLogic->GetShareDirectory();
```

Libs/MRML/Logic/vtkMRMLApplicationLogic.h
```diff
@@ -295,6 +295,21 @@ class VTK_MRML_LOGIC_EXPORT vtkMRMLApplicationLogic : public vtkMRMLAbstractLogi
...
+  /// represents the current working directory. This is used by modules to access files.
+  void SetHomeDirectory(std::string path);
+  const std::string& GetHomeDirectory() const;
+  /// @}
+
+  /// @{
+  /// Applications must set this variable to a meaningful value.
+  /// This path is relative, modules will generally resolve it from GetHomeDirectory().
+  /// For example, the slicer application will set this to "share/slicer-X-Y".
+  void SetShareDirectory(std::string path);
```
jcfr:
```suggestion
  void SetShareDirectory(const std::string& path);
```

Libs/MRML/Logic/vtkMRMLApplicationLogic.h
```diff
@@ -295,6 +295,21 @@ class VTK_MRML_LOGIC_EXPORT vtkMRMLApplicationLogic : public vtkMRMLAbstractLogi
   /// Font file path is set to the one specified in FontFileName property in this object.
   void UseCustomFontFile(vtkTextProperty* textProperty);
 
+  /// @{
+  /// Applications must set this variable to a meaningful value. By default it is empty, this state
+  /// represents the current working directory. This is used by modules to access files.
+  void SetHomeDirectory(std::string path);
```
jcfr:
```suggestion
  void SetHomeDirectory(const std::string& path);
```

Libs/MRML/Logic/vtkMRMLApplicationLogic.cxx
```diff
@@ -1205,13 +1206,37 @@ void vtkMRMLApplicationLogic::UseCustomFontFile(vtkTextProperty* textProperty)
   textProperty->SetFontFamily(VTK_FONT_FILE);
 }
 
+//----------------------------------------------------------------------------
+void vtkMRMLApplicationLogic::SetHomeDirectory(std::string path)
```
jcfr:
```suggestion
void vtkMRMLApplicationLogic::SetHomeDirectory(const std::string& path)
```

Libs/MRML/Logic/vtkMRMLApplicationLogic.cxx
```diff
@@ -1205,13 +1206,37 @@ void vtkMRMLApplicationLogic::UseCustomFontFile(vtkTextProperty* textProperty)
...
+}
+
+//----------------------------------------------------------------------------
+const std::string& vtkMRMLApplicationLogic::GetHomeDirectory() const
+{
+  return this->Internal->HomeDirectory;
+}
+
+//----------------------------------------------------------------------------
+void vtkMRMLApplicationLogic::SetShareDirectory(std::string path)
```
jcfr:
```suggestion
void vtkMRMLApplicationLogic::SetShareDirectory(const std::string& path)
```

Base/QTCore/qSlicerUtils.cxx
```diff
@@ -203,13 +204,27 @@ QString qSlicerUtils::extractModuleNameFromClassName(const QString& className)
...
 }
 
 //-----------------------------------------------------------------------------
 bool qSlicerUtils::isPluginBuiltIn(const QString& filePath, const QString& applicationHomeDir, const QString& applicationRevision)
 {
-  return vtkSlicerApplicationLogic::IsPluginBuiltIn(filePath.toStdString(), applicationHomeDir.toStdString(), applicationRevision.toStdString());
+  std::string extensionsDirBase;
+#ifdef Slicer_BUILD_EXTENSIONMANAGER_SUPPORT
+  extensionsDirBase = Slicer_EXTENSIONS_DIRBASENAME;
+#endif
```
jcfr, line 227:
```suggestion
#ifdef Slicer_BUILD_EXTENSIONMANAGER_SUPPORT
  std::string extensionsDirBase = Slicer_EXTENSIONS_DIRBASENAME;
#else
  std::string extensionsDirBase;
#endif
```

Base/QTCore/qSlicerUtils.cxx
```diff
@@ -203,13 +204,27 @@ QString qSlicerUtils::extractModuleNameFromClassName(const QString& className)
...
 {
-  return vtkSlicerApplicationLogic::IsPluginInstalled(filePath.toStdString(), applicationHomeDir.toStdString());
+  std::string organisationDomain;
+  std::string organisationName;
+  std::string extensionsDirBase;
+#ifdef Slicer_BUILD_EXTENSIONMANAGER_SUPPORT
+  organisationDomain = Slicer_ORGANIZATION_DOMAIN;
+  organisationName = Slicer_ORGANIZATION_NAME;
+  extensionsDirBase = Slicer_EXTENSIONS_DIRBASENAME;
+#endif
```
jcfr:
```suggestion
#ifdef Slicer_BUILD_EXTENSIONMANAGER_SUPPORT
  std::string organisationDomain = Slicer_ORGANIZATION_DOMAIN;
  std::string organisationName = Slicer_ORGANIZATION_NAME;
  std::string extensionsDirBase = Slicer_EXTENSIONS_DIRBASENAME;
#else
  std::string organisationDomain;
  std::string organisationName;
  std::string extensionsDirBase;
#endif

File: Modules/Loadable/Colors/Logic/Testing/Cxx/vtkSlicerColorLogicTest1.cxx [hunk 13]
```diff
@@ -26,6 +26,8 @@
 #include "vtkMRMLCoreTestingMacros.h"
 #include "vtkMRMLScene.h"
 
+#include "vtkSlicerConfigure.h"
```
jcfr:
```suggestion
#include "vtkSlicerConfigure.h" // For Slicer_SHARE_DIR
```

Modules/Loadable/Colors/Logic/vtkSlicerColorLogic.cxx
```diff
@@ -88,27 +88,28 @@ const char* vtkSlicerColorLogic::GetDefaultChartColorNodeID()
...
   {
-    if (vtksys::SystemTools::GetEnv("PWD") != nullptr)
-    {
-      slicerHome = std::string(vtksys::SystemTools::GetEnv("PWD"));
-    }
-    else
-    {
-      slicerHome = std::string("");
-    }
+    vtkErrorMacro("GetMRMLApplicationLogic() must not be null");
```
jcfr:
```suggestion
    vtkErrorMacro("FindDefaultColorFiles failed: Unable to get application logic");
```

Modules/Loadable/Colors/Logic/vtkSlicerColorLogic.cxx
```diff
@@ -88,27 +88,28 @@ const char* vtkSlicerColorLogic::GetDefaultChartColorNodeID()
...
+    return {};
   }
-  else
+
+  const std::string& shareDir = app->GetShareDirectory();
+  if (shareDir.empty())
   {
-    slicerHome = std::string(vtksys::SystemTools::GetEnv("SLICER_HOME"));
+    vtkWarningMacro("ShareDirectory of vtkSlicerApplication is not defined.\n"
+                    "Default color files won't be loaded!");
```
jcfr:
```suggestion
    vtkWarningMacro("FindDefaultColorFiles failed: ShareDirectory is not set. "
                    "Default color files will not be loaded.");
```

jcfr:
As originally hinted by @lassoan in https://github.com/Slicer/Slicer/pull/8004#issuecomment-2432456593, storing settings in an object would be ideal. That said, this could definitely be done in a follow-up pull request.

After addressing those nitpicks, this will be good for integration :rocket: 

----

> [!NOTE]
> When configuring the corresponding slicer library to be distributed as Python wheels, we will have to further define the meaning of `IsPluginBuiltIn`.

Base/Logic/vtkSlicerApplicationLogic.cxx
```diff
@@ -675,25 +672,27 @@ bool vtkSlicerApplicationLogic::IsEmbeddedModule(const std::string& filePath, co
...
+
   return isEmbedded;
 }
 
 //----------------------------------------------------------------------------
-bool vtkSlicerApplicationLogic::IsPluginInstalled(const std::string& filePath, const std::string& applicationHomeDir)
+bool vtkSlicerApplicationLogic::IsPluginInstalled(const std::string& filePath,
+                                                  const std::string& applicationHomeDir,
+                                                  const std::string& organisationDomain,
+                                                  const std::string& organisationName,
```
jcfr: `organisation` -> `organization` (We prefer american english). I will update the patch accordingly

Base/Logic/vtkSlicerApplicationLogic.h
```diff
@@ -181,20 +181,32 @@ class VTK_SLICER_BASE_LOGIC_EXPORT vtkSlicerApplicationLogic : public vtkMRMLApp
...
 
   /// Return True if \a filePath is a descendant of \a applicationHomeDir.
-  /// \note On macOS extensions are installed in the "<Slicer_EXTENSIONS_DIRBASENAME>-<slicerRevision>"
+  /// \note On macOS extensions are installed in the "<extensionsDirBase>-<slicerRevision>"
   /// folder being a sub directory of the application dir, an extra test is performed to make sure the
-  /// tested filePath doesn't belong to that "<Slicer_EXTENSIONS_DIRBASENAME>-<slicerRevision>" folder.
+  /// tested filePath doesn't belong to that "<extensionsDirBase>-<slicerRevision>" folder.
   /// If this is the case, False will be returned.
-  static bool IsEmbeddedModule(const std::string& filePath, const std::string& applicationHomeDir, const std::string& slicerRevision);
+  /// If \a extensionsDirBase is empty, this function assumes that Slicer was build with
```
jcfr:
```suggestion
  /// If \a extensionsDirBase is empty, this function assumes that Slicer was built
```

I will fix this as well.

jcfr (approved):

AlexyPellegrini:
Thank you JC for finishing up this merge request! I will open a PR for the split of MRML libraries and Slicer application soon, so we can start discussing about it!

PR #8008, opened 2024-10-25
ENH: Add option to display slice edges in 3D view

Punzo:
- Implemented functionality to show slice edges in the 3D view.
- Added user interface controls to control the visibility of slice edges.
- Implemented linking controls to synchronize the visibility of slice edges across different views.

image

Reference: https://discourse.slicer.org/t/show-slice-edges-in-3d-view/11381

### DIFF
--- a/Libs/MRML/Core/vtkMRMLSliceNode.h
+++ b/Libs/MRML/Core/vtkMRMLSliceNode.h
@@ -81,10 +81,17 @@ class VTK_MRML_EXPORT vtkMRMLSliceNode : public vtkMRMLAbstractViewNode
   vtkSetMacro(SliceVisible, int);
 
   ///
+  /// Show colored slice edge in 3D views.
+  vtkGetMacro(SliceEdgeVisibility3D, bool);
+  vtkSetMacro(SliceEdgeVisibility3D, bool);
+  vtkBooleanMacro(SliceEdgeVisibility3D, bool);
+
+  ///
   /// The visibility of the slice plane widget in the 3DViewer.
   vtkGetMacro(WidgetVisible, int);
   vtkSetMacro(WidgetVisible, int);
 
+  ///
   /// The visibility of the slice plane widget outline in the 3DViewer.
   vtkGetMacro(WidgetOutlineVisible, int);
   vtkSetMacro(WidgetOutlineVisible, int);
@@ -469,6 +476,7 @@ public:
   ///    XYZOriginFlag - broadcast the XYZOrigin to all linked viewers
   ///    LabelOutlineFlag - broadcast outlining the labelmaps
   ///    SliceVisibleFlag = broadcast display of slice in 3D
+  ///    SliceEdgeVisibility3DFlag = broadcast display of slice edge in 3D
   ///    ResetOrientationFlag = broadcast a reset to default orientation to all linked viewers
   ///    UpdateSlabReconstructionThicknessFlag = broadcast updating the slab reconstruction thickness
   enum InteractionFlagType
@@ -482,10 +490,11 @@ public:
     XYZOriginFlag = 32,
     LabelOutlineFlag = 64,
     SliceVisibleFlag = 128,
-    SliceSpacingFlag = 256,
-    ResetOrientationFlag = 512,
-    RotateToBackgroundVolumePlaneFlag = 1024,
-    UpdateSlabReconstructionThicknessFlag = 2048,
+    SliceEdgeVisibility3DFlag = 256,
+    SliceSpacingFlag = 512,
+    ResetOrientationFlag = 1024,
+    RotateToBackgroundVolumePlaneFlag = 2048,
+    UpdateSlabReconstructionThicknessFlag = 4096,
   };
 
   /// Get/Set a flag indicating what parameters are being manipulated
@@ -580,6 +589,7 @@ protected:
   int WidgetOutlineVisible;
   int WidgetNormalLockedToCamera;
   int UseLabelOutline;
+  bool SliceEdgeVisibility3D;
 
   double FieldOfView[3];
   double XYZOrigin[3];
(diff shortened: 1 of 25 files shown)


Punzo:
@lassoan @pieper @jcfr 
https://github.com/user-attachments/assets/fbc93a20-6561-42e6-96b6-f7b926c9bbdc

Libs/MRML/Core/vtkMRMLSliceNode.h
```diff
@@ -78,11 +78,27 @@ class VTK_MRML_EXPORT vtkMRMLSliceNode : public vtkMRMLAbstractViewNode
   vtkGetMacro(SliceVisible, int);
   vtkSetMacro(SliceVisible, int);
 
+  ///
+  /// The visibility of the slice edge in the 3DViewer.
+  vtkGetMacro(ThreeDSliceEdgeVisibility, int);
```
lassoan, line 83: Using int instead of bool was a limitation in Tcl decades ago. For new Boolean variables we should use bool type.

lassoan, line 83: Use `SliceEdgeVisibility3D` instead, it is more similar to `SliceVisible` and this suffix is used elsewhere similarly, too.

Libs/MRML/Core/vtkMRMLSliceNode.h
```diff
@@ -78,11 +78,27 @@ class VTK_MRML_EXPORT vtkMRMLSliceNode : public vtkMRMLAbstractViewNode
...
+  vtkSetMacro(ThreeDSliceEdgeVisibility, int);
+
+  ///
+  /// The opacity of the slice edge in the 3DViewer.
+  vtkGetMacro(ThreeDSliceEdgeOpacity, double);
```
lassoan, line 88: Opacity and and radius should not be necessary to tune for each view. It should "just work", the slice view should just look nice. The same scaling factor that is used for relative scaling of markups could be used here (`this->GetScreenScaleFactor() * this->ViewScaleFactorMmPerPixel`). Opacity could be set to 1.0.

Libs/MRML/DisplayableManager/vtkMRMLThreeDSliceEdgeDisplayableManager.cxx
```diff
@@ -0,0 +1,455 @@
...
+#include <vtkMRMLVolumeNode.h>
+#include <vtkMRMLViewNode.h>
+
+// VTK includes
+#include <vtkActor.h>
+#include <vtkCallbackCommand.h>
+#include <vtkColor.h>
+#include <vtkCutter.h>
+#include <vtkFeatureEdges.h>
+#include <vtkImageData.h>
```
lassoan, line 41: remove includes that are no longer needed

Libs/MRML/DisplayableManager/vtkMRMLThreeDSliceEdgeDisplayableManager.cxx
```diff
@@ -0,0 +1,455 @@
...
+#include <vtkRenderer.h>
+#include <vtkRenderWindowInteractor.h>
+#include <vtkStripper.h>
+#include <vtkTubeFilter.h>
+
+// STD includes
+#include <cmath>
+
+//---------------------------------------------------------------------------
+class EdgeRepresentation
```
lassoan, line 60: We use the "pipeline" name for such classes.
```suggestion
class SliceEdgePipeline
```

Libs/MRML/DisplayableManager/vtkMRMLThreeDSliceEdgeDisplayableManager.cxx
```diff
@@ -0,0 +1,455 @@
...
+};
+
+//---------------------------------------------------------------------------
+vtkStandardNewMacro(vtkMRMLThreeDSliceEdgeDisplayableManager);
+
+//---------------------------------------------------------------------------
+class vtkMRMLThreeDSliceEdgeDisplayableManager::vtkInternal
+{
+public:
+  typedef std::map<vtkMRMLSliceNode*, EdgeRepresentation*> SliceNodesLink;
```
lassoan, line 117: Managing raw pointers is very risky.
```suggestion
  typedef std::map<vtkWeakPointer<vtkMRMLSliceNode>, EdgeRepresentation> SliceNodesLink;
```

lassoan (commented):
Very nice work, thank you! I've added a few comments inline.

Punzo:
> Very nice work, thank you! I've added a few comments inline.

thanks for reviewing! I will apply the review on Monday! have a nice weekend!

lassoan:
`GetViewScaleFactorAtPosition()` is somewhat complex operation and it is reimplemented in several widget representation classes already. It would be better not to add yet another copy of this implementation to the slice edge representation but to move the implementation to the MRML camera node and use that everywhere.

Punzo:
@lassoan review applied, please have a look again.

NOTE: I moved the method `GetViewScaleFactorAtPosition` in the `vtkMRMLAbstractThreeDViewDisplayableManager` (as a static one) instead of `vtkMRMLCameraNode`. This because the `vtkMRMLInteractionEventData` (needed for the markups representations) can't be included in `vtkMRMLCameraNode` (i.e. core MRML node do not have access to the classes in the displayManagers folder)

jcfr:
> at the moment the slice edge display manager is only for display,

:100:  This is great. Thanks for working on this.

> but in future we can improve it and we can add interactions (slice translation, etc...)

Ditto. That would be a sensible next step.

lassoan (approved):
I've tested and fixed it up a bit more (removing of view nodes, scene close, etc) and added an application setting to set default visibility. It all looks good, ready for integration.

Punzo:
> I've tested and fixed it up a bit more (removing of view nodes, scene close, etc) and added an application setting to set default visibility. It all looks good, ready for integration.

thanks Andras for the fixes and merging!

lassoan (approved):

PR #8010, opened 2024-10-26
DOC: Update recommended screen resolution in documentation

lassoan:
Screens are wider today than 15 years ago when this requirement was specified, therefore specifying minimum resolution in 4:3 aspect ratio is not relevant anymore.

See discussion in https://discourse.slicer.org/t/full-screen-issue-with-slicer-on-mac-os-15-0-1/39753/18


pieper (approved):
👍

PR #8011, opened 2024-10-26
ENH: Introduce setting for using standard terminology in segments table

cpinter:
Previously the segments table used the terminology selector with double-clicking the color, but allowed direct editing when double-clicking the name. Instead of this inconsistent behavior, now we have a new option in settings for using standard terminology. If turned on, both the name and the color shows the terminology selector, otherwise a simple color picker and text editor, respectively. In addition to the checkbox in Application Settings, this commit adds a checkable action to the context menu of the segments in the segments table. The segments tables may choose a different setting key to have a behavior independent of the default setting (which is Segmentations/SegmentsTableUseStandardTerminology).

Re #6975


cpinter:
Rework of https://github.com/Slicer/Slicer/pull/7983

How the change works:

https://github.com/user-attachments/assets/8352963c-3a47-4553-a930-164d3f7ac557

Modules/Loadable/Segmentations/Widgets/qMRMLSegmentsTableView.cxx
```diff
@@ -460,6 +469,76 @@ void qMRMLSegmentsTableView::onSegmentsTableClicked(const QModelIndex& modelInde
...
+      {
+        return;
+      }
+
+      // Get terminology from segment. Do not check success, as an empty terminology is also a valid starting point
+      vtkNew<vtkSlicerTerminologyEntry> terminologyEntry;
+      std::string serializedTerminology("");
+      if (segment->GetTag(vtkSegment::GetTerminologyEntryTagName(), serializedTerminology))
+      {
+        terminologiesLogic->DeserializeTerminologyEntry(serializedTerminology.c_str(), terminologyEntry);
```
lassoan, line 506: Please check if everything works without calling

    // Get default color and other metadata from loaded terminologies
    logic->UpdateEntryFromLoadedTerminologies(terminologyEntry);

You need to call `UpdateEntryFromLoadedTerminologies` if you need any information that was not in the `serializedTerminology` string.

Modules/Loadable/Segmentations/Widgets/qMRMLSegmentsTableView.h
```diff
@@ -166,6 +170,9 @@ public slots:
   /// Set the text used to filter the segments in the table
   /// \sa textFilter
   void setTextFilter(QString textFilter);
+  /// Set the settings key used to specify whether standard terminologies are used for name and color.
+  /// \sa terminologySelectorSettingsKey
```
lassoan, line 174: Describe what is the default and what happens if the key is set to empty string.

lassoan (commented):
This looks perfect, thank you! I've just added a few trivial comments inline.

muratmaga:
@cpinter when this is merged, can you add an entry to SlicerMorph customizations that will make this enabled by default (when the SlicerMorph customizations is enabled).

https://github.com/SlicerMorph/SlicerMorph/blob/master/MorphPreferences/Resources/SlicerMorphRC.py

cpinter:
@muratmaga The default is to use the terminology both for name and color. The custom setting key is for segment tables that do not want to use it but the simple text/color. Does this change if I need to add anything? My guess would be that we don't need customizations so please clarify.

Modules/Loadable/Segmentations/Widgets/qMRMLSegmentsTableView.cxx
cpinter: It is only for identification, so only the serialized components are needed. I did try it and it works.

cpinter:
I assumed that we'll keep terminology as the default, but it's true that now that the name won't be that freely changeable, it may cause frustration to users. On the other hand, we'd like to encourage the use of terminology.

The mechanism you describe seems to make sense, but @muratmaga 's funds are not that large for this particular development and we may need to prioritize.

@pieper @fedorov Do you have any strong argument against using the terminology selector by default for both the name and color in segment editor?

It's a good question about other data types. I'd discuss that separately, as it's the segmentation that is most sensitive to the issues caused by manual editing, i.e. issues in training models and doing longitudinal comparisons if the same structures are named differently ("Orbit Lt" vs "Lt Orbit" vs "Orbit left", simple typos, etc.).

pieper:
> @pieper @fedorov Do you have any strong argument against using the terminology selector by default for both the name and color in segment editor?

As long as it's easy to define your own terms and pick from a nice set of contrasting colors it's okay with me.  I agree terminologies are preferable for many purposes but often they just get in the way.

For example, I was doing some segmentation this morning had no use for terminologies (the person I was sending the file to specifically requested nrrd with no extra headers because he was afraid his custom reader code might fail - I suggested slicerio but that would have just complicated things on his end).

In any case I looked and didn't see that there was a term for one of my segments (collecting system of the right kidney - does anyone know if that exists in a terminology we provide?).

cpinter:
Thank you Steve! It's in the plans to make editing and adding custom terms easy.

At this point what I'd do is keep the terminologies as default in the nightlies, continue the development, and see where we are just before releasing 5.8. If it seems good enough (all the necessary supplementary features are in and working, no bad feedback on discourse) then we keep it, if not, we turn it off for the stable (and apply the customization in SlicerMorph).

> We could also add a checkbox to the terminology selector that would prevent showing the popup next time (maybe the checkbox could appear when the user entered custom name or color), but that would make the GUI a bit more noisy.

Actually I'd consider this as well.

> In any case I looked and didn't see that there was a term for one of my segments (collecting system of the right kidney - does anyone know if that exists in a terminology we provide?).

Unfortunately there are some basic-looking terms that are not included.

fedorov:
> It's in the plans to make editing and adding custom terms easy.

If those terms are non-standard (not coming from any DICOM-recognized ontology), it is important to make sure that Coding Scheme Designator starts with 99. The tricky part here is that overall, DICOM discourages using non-standard terms. Ideally, it would be preferred to help users to at least try to pick standard terms from ontology resources before they decide to use something that will be non-standard. Not a perfect resource by any stretch, but this page might help - maybe it makes sense to include it in a tooltip or something: https://qiicr.gitbook.io/dcmqi-guide/opening/coding_schemes.

muratmaga:
> If those terms are non-standard (not coming from any DICOM-recognized ontology),

The issue is DICOM-recognized ontology is medical only, and does not have enough coverage non-human organisms. So there are a lot of "custom terms" we need.

muratmaga:
> In any case I looked and didn't see that there was a term for one of my segments (collecting system of the right kidney - does anyone know if that exists in a terminology we provide?).

It does: Both in segmentation, and DICOm terms:
image

muratmaga:
Sorry, your full term is "collecting system of the right kidney". Yeah, I would be surprised if something that specific exists.

jamesobutler:
> At this point what I'd do is keep the terminologies as default in the nightlies, continue the development, and see where we are just before releasing 5.8. If it seems good enough (all the necessary supplementary features are in and working, no bad feedback on discourse) then we keep it, if not, we turn it off for the stable (and apply the customization in SlicerMorph).

@cpinter FYI that Slicer Stable 5.8 is scheduled for release in about the first week of November so this period would only be about a week. This seems to be a short period. It could be possible to maintain existing behavior for the imminent Slicer stable and have the new behavior in the Slicer preview post-5.8 release to provide a longer experimental time frame until say a Slicer 5.10/6.0 stable.

fedorov:
> The issue is DICOM-recognized ontology is medical only, and does not have enough coverage non-human organisms. So there are a lot of "custom terms" we need.

Yes, I completely understand. But unless the code is from one of the coding schemes listed in https://dicom.nema.org/medical/dicom/current/output/chtml/part16/chapter_8.html, CodingSchemeDesignator must start with 99 when serializing into DICOM. And since CodingSchemeDesignator from Terminologies gets reused in DICOM export, it is highly desirable to initialize it to a standard-compliant values.

muratmaga:
> odingSchemeDesignator must start with 99 when serializing into DICOM. And since CodingSchemeDesignator from Terminologies gets reused in DICOM export, it is highly desirable to initialize it to a standard-compliant values.

Thanks for the clarification. Most of our terms "hopefully" will come from Uberon, which is listed under the link you share. But we know for a fact that even that's not enough.

pieper:
> Sorry, your full term is "collecting system of the right kidney". Yeah, I would be surprised if something that specific exists.

Yes, I think that's just the point - it's not the only way to describe the inner part of the kidney, but it's a pretty common term, and it nicely describes the part of the kidney that's not enhanced during the arterial phase of a contrast CT.  I spent a few minutes looking in the various terminology dialog locations before giving up.  Maybe there's another name for the same concept, or maybe parts have specific names like 'renal medulla' and 'renal pelvis' that are defined in a terminology but for this use case I wanted to indicate that I was including both (and I don't believe it's possible to pick two terms in the same segment).

> CodingSchemeDesignator from Terminologies gets reused in DICOM export, it is highly desirable to initialize it to a standard-compliant values

Absolutely, we should be careful not to release non-compliant data.  That's an argument for not assuming that the user's case is going to fit with an existing terminology.  My vote would be to make people explicitly select or design a meaningful terminology that they know matches their use case rather then encouraging them to use an existing one that may be a bad fit.  Otherwise we should assume they are doing something non-standard and give some kind of 99 code.  Selecting standard terms that actually match what you are trying to segment is a difficult business.

muratmaga:
> It could be possible to maintain existing behavior for the imminent Slicer stable and have the new behavior in the Slicer preview post-5.8 release to provide a longer experimental time frame until say a Slicer 5.10/6.0 stable.

We do need this behavior in the 5.8 stable, as our project uses the stable versions for production. As long as the change is integrated, I am open to having it integrated as not the default behavior in stable (we can set it through SlicerMorph customizations as I indicated above).  But it is up to others as well.

muratmaga:
what is the status of this change?

lassoan:
We are waiting on opinion from others. If we don't hear from anyone then I think most of us would found that it would be an acceptable solution to show terminology selector by default but offer switching to a simple selector in a popup after repeatedly choosing custom name or color. If anybody does not agree then please speak up!

@cpinter do you think you could implement this popup before 5.8 release is cut (end of next week)? If not then I may be able to help out with it or we could figure out something else.

cpinter:
I could, with permission from @muratmaga. There are various other improvements needed, and this one has used up a considerable part of the small project with implementing this twice.

muratmaga:
@lassoan @cpinter we really have a small budget for this, as this is quite tangential to our work. @cpinter needs to focus on other project specific asks (which are the custom terminology generation/import) and they are open ended. If you or others can pitch in that will great. Otherwise these changes need to go in without that popup.

lassoan:
OK. I'll add the popup. Probably by Monday. I'll push to this same branch and then we can review and merge early next week.

cpinter:
If there is no way this can get integrated then we really appreciate your help Andras! Unfortunately on the weekend I'll have to deal with our part of two EU grant applications (things always come in groups), there is no time I can volunteer at this time.

By the way @muratmaga what is the reason you are fixed to use the stable?

muratmaga:
Because we deploy Slicer and SlicerMorph on the MorphoCloud and needs to be production level application. We can't constantly bump up the slicer version and the extensions, when we introduce a new feature in SlicerMorph or fix a bug.

cpinter:
I implemented the question as @lassoan suggested. The popup looks like this:
image

The number of times the user needs to select custom name/color before asking the question is hard-coded to 5.
One thing worth noting is that the question is only popped up if the counter is exactly at 5 (static variable so any terminology selection contributes), so that it is not asked again if the user chooses to keep using terminologies.

Please let us know if the PR is good to go with this. Thank you!

lassoan:
This will be good, thank you! However, when I tested it I ran into a crash:
- load an image, create a segmentation, create a few segments
- double-click on name and color columns -> terminology selector comes up (good)
- after 5 or so edits, I get the popup (good)
- clicked to "switch to direct editing"
- I'm not sure if I double-clicked to edit or right after a clicked to switch to direct editing, Slicer crashed because `lineEdit` was nullptr in qSlicerTerminologyItemDelegate::setModelData:

Stack trace:

```
Qt5Widgetsd.dll!QScopedPointer<QObjectData,QScopedPointerDeleter<QObjectData>>::operator->() Line 118
	at C:\Users\qt\work\qt\qtbase\src\corelib\tools\qscopedpointer.h(118)
Qt5Widgetsd.dll!qGetPtrHelper<QScopedPointer<QObjectData,QScopedPointerDeleter<QObjectData>> const>(const QScopedPointer<QObjectData,QScopedPointerDeleter<QObjectData>> & ptr) Line 1135
	at C:\Users\qt\work\qt\qtbase\src\corelib\global\qglobal.h(1135)
Qt5Widgetsd.dll!QLineEdit::d_func() Line 254
	at C:\Users\qt\work\qt\qtbase\src\widgets\widgets\qlineedit.h(254)
Qt5Widgetsd.dll!QLineEdit::text() Line 311
	at C:\Users\qt\work\qt\qtbase\src\widgets\widgets\qlineedit.cpp(311)
qSlicerTerminologiesModuleWidgets.dll!qSlicerTerminologyItemDelegate::setModelData(QWidget * editor, QAbstractItemModel * model, const QModelIndex & index) Line 130
	at C:\D\S4\Modules\Loadable\Terminologies\Widgets\qSlicerTerminologyItemDelegate.cxx(130)
Qt5Widgetsd.dll!QAbstractItemView::commitData(QWidget * editor) Line 2909
	at C:\Users\qt\work\qt\qtbase\src\widgets\itemviews\qabstractitemview.cpp(2909)
Qt5Widgetsd.dll!QAbstractItemView::qt_static_metacall(QObject * _o, QMetaObject::Call _c, int _id, void * * _a) Line 427
	at C:\Users\qt\work\qt\qtbase\src\widgets\.moc\debug\moc_qabstractitemview.cpp(427)
Qt5Cored.dll!doActivate<0>(QObject * sender, int signal_index, void * * argv) Line 3901
	at C:\Users\qt\work\qt\qtbase\src\corelib\kernel\qobject.cpp(3901)
Qt5Cored.dll!QMetaObject::activate(QObject * sender, const QMetaObject * m, int local_signal_index, void * * argv) Line 3947
	at C:\Users\qt\work\qt\qtbase\src\corelib\kernel\qobject.cpp(3947)
Qt5Widgetsd.dll!QAbstractItemDelegate::commitData(QWidget * _t1) Line 206
	at C:\Users\qt\work\qt\qtbase\src\widgets\.moc\debug\moc_qabstractitemdelegate.cpp(206)
qSlicerTerminologiesModuleWidgets.dll!qSlicerTerminologyItemDelegate::commitAndClose() Line 184
	at C:\D\S4\Modules\Loadable\Terminologies\Widgets\qSlicerTerminologyItemDelegate.cxx(184)
qSlicerTerminologiesModuleWidgets.dll!qSlicerTerminologyItemDelegate::qt_static_metacall(QObject * _o, QMetaObject::Call _c, int _id, void * * _a) Line 80
	at C:\D\S4D\Slicer-build\Modules\Loadable\Terminologies\Widgets\moc_qSlicerTerminologyItemDelegate.cpp(80)
Qt5Cored.dll!QMetaCallEvent::placeMetaCall(QObject * object) Line 618
	at C:\Users\qt\work\qt\qtbase\src\corelib\kernel\qobject.cpp(618)
Qt5Cored.dll!QObject::event(QEvent * e) Line 1314
	at C:\Users\qt\work\qt\qtbase\src\corelib\kernel\qobject.cpp(1314)
Qt5Widgetsd.dll!QApplicationPrivate::notify_helper(QObject * receiver, QEvent * e) Line 3632
	at C:\Users\qt\work\qt\qtbase\src\widgets\kernel\qapplication.cpp(3632)
```

image

lassoan:
@cpinter I've fixed the issue (and added a few options to customize the segment list behavior). Please review the code and test if it works well for you.

Modules/Loadable/Segmentations/Widgets/qMRMLSegmentsTableView.cxx
```diff
@@ -313,7 +321,8 @@ void qMRMLSegmentsTableViewPrivate::init()
 
   // Set item delegate to handle color and opacity changes
   this->TerminologyItemDelegate = new qSlicerTerminologyItemDelegate(this->SegmentsTable);
-  this->TerminologyItemDelegate->setTerminologySelectorSettingsKey(this->TerminologySelectorSettingsKey);
+  this->TerminologyItemDelegate->setUseTerminologySelectorCallback([q]{ return q->useTerminologySelector(); });
```
cpinter, line 324: Slightly different style (space or no space before `{`)

Modules/Loadable/Terminologies/Widgets/qSlicerTerminologySelectorDialog.cxx
```diff
@@ -29,21 +29,17 @@
...
-
 //-----------------------------------------------------------------------------
 class qSlicerTerminologySelectorDialogPrivate : public QDialog
 {
   Q_DECLARE_PUBLIC(qSlicerTerminologySelectorDialog);
 protected:
   qSlicerTerminologySelectorDialog* const q_ptr;
 public:
-  qSlicerTerminologySelectorDialogPrivate(qSlicerTerminologySelectorDialog& object, QWidget* parent);
+  qSlicerTerminologySelectorDialogPrivate(qSlicerTerminologySelectorDialog& object);
```
cpinter, line 42: The parent was to improve positioning of the dialog. It's better to appear in the center of the app than the center of the screen, the user needs to move the mouse less.

cpinter:
Thanks @lassoan! I added some comments.

Modules/Loadable/Segmentations/Widgets/qMRMLSegmentsTableView.cxx
```diff
@@ -215,6 +215,14 @@ class qMRMLSegmentsTableViewPrivate: public Ui_qMRMLSegmentsTableView
 
   /// The settings key used to specify whether standard terminologies are used for name and color.
   QString TerminologySelectorSettingsKey{"Segmentations/SegmentsTableUseStandardTerminology"};
+
+  /// Counter for the number of times the user specifies custom name. Offer not using terminology if used often.
+  unsigned int NonStandardTerminologySelectedCounter{0};
+
+  bool TerminologySelectorOptionVisible{ true };
+  // Use terminology selector if TerminologySelectorSettingsKey is empty
```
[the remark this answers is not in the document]
lassoan: I find it a bit misleading to add the third slash to comments in .cxx files (the third slash is a hint for doxygen to add the comment to the documentation, but doxygen ignores content in .cxx files), but all the other methods now use that, so I'll add it for consistency.

Modules/Loadable/Terminologies/Widgets/qSlicerTerminologySelectorDialog.cxx
lassoan: I moved other changes in this file elsewhere and it no longer seemed necessary. I'll add it back. It is important to document this kind of hidden knowledge, to avoid accidental removal.

Modules/Loadable/Segmentations/Widgets/qMRMLSegmentsTableView.cxx
```diff
@@ -215,6 +215,14 @@ class qMRMLSegmentsTableViewPrivate: public Ui_qMRMLSegmentsTableView
 
   /// The settings key used to specify whether standard terminologies are used for name and color.
   QString TerminologySelectorSettingsKey{"Segmentations/SegmentsTableUseStandardTerminology"};
+
+  /// Counter for the number of times the user specifies custom name. Offer not using terminology if used often.
+  unsigned int NonStandardTerminologySelectedCounter{0};
```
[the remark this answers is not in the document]
lassoan: Sharing the counter only makes sense between segments table views that use the exact same flags and settings key. It could have been implemented by computing a hash from these variables and then storing the counters in a static QMap, but I don't think it would have improved the heuristics (it is approximate anyway) and it would have made the behavior a bit less predictable for users (they don't necessarily expect that using a segment selector in one module has an effect of how another module behaves).

lassoan:
@cpinter thanks for the review. I've answered all the comments, please check. We will have to squash all the commits in the end.

fedorov:
> In general I find it a bit annoying that there are monumental commits that need to be treated as one because they were squashed, and sometimes this makes it much more difficult to understand (as opposed to a logical sequence of smaller changes with clear commit message in each), and then there are six of these "COMP: Bump github/codeql-action from X to Y" in a row, which pollute the history, drawing attention from actual changes. This is just a general observation...

I agree. It makes a lot of sense to squash commits within a PR that are incremental, but squashing all of them for a large contribution is not helpful. Especially for experienced developers, it seems to make more sense to let the contributor decide which commits should be kept, and which should be squashed as non-essential for communicating development history and sub-tasks within a PR.

lassoan (approved):
OK, it looks good enough now.

I say "enough" because for me this error is logged whenever I open the terminology dialog (in a debug-mode build):

```
[Qt] QWindowsWindow::setGeometry: Unable to set geometry 200x60+1622+714 (frame: 222x116+1611+669) on QWidgetWindow/"QDialogClassWindow" on "\\.\DISPLAY1". Resulting geometry: 344x448+1622+714 (frame: 366x504+1611+669) margins: 11, 45, 11, 11 minimum size: 172x224 MINMAXINFO maxSize=0,0 maxpos=0,0 mintrack=366,504 maxtrack=0,0)
```

Added https://github.com/Slicer/Slicer/issues/8038 to track this.

I've changed the registry key name to `Segmentations/UseTerminologySelector` to reflect that in the future this key will also be applied to subject hierarchy tree and other places. In the future, we may implement similar mechanism for the subject hierarchy tree for segmentations and for other nodes. I've added https://github.com/Slicer/Slicer/issues/8039 to track this.

cpinter:
Perfect, thank you! I will check the geometry related error once this PR is integrated (then I can just fetch+rebase the main).

lassoan:
Yes, it is in! As you have found, it has an issue with the visibility icon, but otherwise should be ready for testing.

PR #8020, opened 2024-10-30
STYLE: Remove unused MRMLIDImageIO API

### LINKED ISSUE #7087: unreachable code
Libs/MRML/IDImageIO/itkMRMLIDImageIO.cxx line 422

Code:
 throw( "MRML Node does not contain image data." );

  return static_cast< void * >( nullptr ); // <-- unreachable
}

jcfr:
This commit removes the unused virtual functions `CanUseOwnBuffer`, `ReadUsingOwnBuffer`, and `GetOwnBuffer` from `MRMLIDImageIO`. These functions were originally introduced in Slicer through commit `bc99a8cfc1` ("ENH: Modify ITK MRML image reader to re-use buffer when possible", 2010-11-15).

The API was initially added to ITK via InsightSoftwareConsortium/ITK@2ffb32de2f ("ENH: Modify ITK image reader to re-use buffer when possible", 2010-11-15) on the `Slicer4UpdatesToITKv3.20` branch, which was later merged into ITK's `release` branch in commit InsightSoftwareConsortium/ITK@d4d43abf60 ("Merge branch 'Slicer4UpdatesToITKv3.20' into release", 2011-10-24).

Despite a subsequent merge of the `release` branch back into the ITK development branch via commit InsightSoftwareConsortium/ITK@5d6212c8a1 ("Merge branch 'release'", 2011-10-24), the buffer re-use API was not retained, likely due to an "ours" merge strategy (`git checkout master && git merge -s ours release`).

Fixes #7087

### DIFF
--- a/Libs/MRML/IDImageIO/itkMRMLIDImageIO.cxx
+++ b/Libs/MRML/IDImageIO/itkMRMLIDImageIO.cxx
@@ -377,52 +377,6 @@ MRMLIDImageIO
 }
 
 //----------------------------------------------------------------------------
-// Read from the MRML scene
-bool
-MRMLIDImageIO
-::CanUseOwnBuffer()
-{
-  return true;
-}
-
-//----------------------------------------------------------------------------
-// Read from the MRML scene
-void
-MRMLIDImageIO
-::ReadUsingOwnBuffer()
-{
-  return;
-}
-
-//----------------------------------------------------------------------------
-// Read from the MRML scene
-void *
-MRMLIDImageIO
-::GetOwnBuffer()
-{
-  vtkMRMLVolumeNode *node;
-
-  node = this->FileNameToVolumeNodePtr( m_FileName.c_str() );
-  if (node)
-  {
-    // buffer is preallocated, memcpy the data
-    if (vtkMRMLDiffusionImageVolumeNode::SafeDownCast(node) == nullptr)
-    {
-      // Scalar, Diffusion Weighted, or Vector image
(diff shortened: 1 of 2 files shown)


pieper (approved):
Thanks for the detailed research Jc.  I suppose if these had been marked `override` we would have gotten a compiler error earlier in the process.

In any case it's fine to drop this.  Even if it were working the use cases were very limited (trying to write ITK-based CLI modules that only operated on a few pixels of a volume that's already in memory).  As it's worked out that case didn't get a lot of use.

lassoan:
It should be fine to remove this API. For fast data updates we no longer use CLIs but instead use Python and directly share the memory array of vtkImageData via numpy arrays.

jcfr:
>  I suppose if these had been marked override we would have gotten a compiler error earlier in the process.

Indeed, this would have resulted in a compiler error.

> In any case it's fine to drop this. 

Thanks everyone for the review :pray: 

Also, if we either need to resurrect the associated API we know have a good record of the associated history.

> For fast data updates we no longer use CLIs but instead use Python and directly share the memory array of vtkImageData via numpy arrays.

And on the ITK side, there is also functions like `itk.array_view_from_image` internally relying on `itk::Image::GetBufferPointer`

blowekamp:
> And on the ITK side, there is also functions like `itk.array_view_from_image` internally relying on `itk::Image::GetBufferPointer`

Also from the ITK side of things there is the ImportImageFilter which may work in some cases.

PR #8021, opened 2024-10-30
BUG: Fix issue preventing interacting with Markups curves

Sunderlandkyl:
The interaction renderers did not have PreserveDepthBufferOn and EraseOff flags set, causing the depth buffer to be cleared before it could be used for curve picking. Fixed by setting the flags.

In order to ensure the actors are still rendered correctly, this commit also updates the interaction handle actor offset so that the handles are rendered on top of other actors, and updates the position of the actor in the slice views to prevent them from being clipped by the near/far clipping planes.

See discussion:
https://discourse.slicer.org/t/lost-ability-to-interact-with-closedcurves-in-preview/37134/


Libs/MRML/DisplayableManager/vtkMRMLInteractionWidgetRepresentation.cxx
```diff
@@ -971,7 +971,13 @@ vtkMRMLInteractionWidgetRepresentation::InteractionPipeline::InteractionPipeline
   this->Mapper3D->SetLookupTable(this->ColorTable);
   this->Mapper3D->ScalarVisibilityOn();
   this->Mapper3D->UseLookupTableScalarRangeOn();
+
+  // Update offset so that the handles are always rendered on top of other actors.
+  double offsetFactor = 1.0;
+  double offsetUnits = -2500.0;
```
lassoan, line 977: This is very simple, which is good; but hardcoding specific values at many places may become hard to maintain.

What do you think about specifying coincident topology offsets and renderer layer indices at one place, somewhere in MRML Core or DisplayableManager? It should be easily accessible from anywhere, probably with a static function. It could be added for example to `vtkMRMLAbstractDisplayableManager.h`, `vtkMRMLAbstractWidgetRepresentation.h`, `vtkMRML.h`.

lassoan (commented):
Thanks a lot, this looks good. I've just added a comment inline.

Libs/MRML/DisplayableManager/vtkMRMLInteractionWidgetRepresentation.cxx
pieper: I agree with the idea of centralizing these and not hard coding anything.  Also I'd like to see getter/setter methods since the values may depend on the hardware/software used in rendering or the datatype of the framebuffer.

Sunderlandkyl:
Ok, I've moved the values to [vtkMRMLAbstractWidgetRepresentation](https://github.com/Slicer/Slicer/pull/8021/files#diff-b6636fef029da329a81325147969534f31d76ce85fa617507aed2e31c2200003). We now use the AlwaysOnFlag and vtkMRMLAbstractWidgetRepresentation::UpdateRelativeCoincidentTopologyOffsets() function to update the offset values.

Sunderlandkyl:
@lassoan @pieper any additional comments?

pieper:
Looks great to me, thanks 👍

lassoan (approved):

PR #8023, opened 2024-10-31
ENH: Switch highlighted social network to LinkedIn

jamesobutler:
The 3DSlicerApp account on X.com (formerly Twitter) has not been active since March 2023. LinkedIn continues to be a focused place for researchers to post about their work in an academic/professional setting unlike X.com that covers a wide spectrum of topics. Therefore LinkedIn may be considered more appropriate for connecting researchers to others in more helpful ways.

Previously discussed in https://discourse.slicer.org/t/social-media-presence-moving-away-from-twitter/30433/8.

See joint PR to update the Slicer website with similar change:
https://github.com/Slicer/slicer.org/pull/285

### DIFF
--- a/Applications/SlicerApp/qSlicerAppMainWindow.cxx
+++ b/Applications/SlicerApp/qSlicerAppMainWindow.cxx
@@ -102,9 +102,9 @@ void qSlicerAppMainWindowPrivate::setupUi(QMainWindow * mainWindow)
   helpBrowseTutorialsAction->setText(qSlicerAppMainWindow::tr("Browse Tutorials"));
   helpBrowseTutorialsAction->setToolTip(qSlicerAppMainWindow::tr("Raise the training pages in your favorite web browser"));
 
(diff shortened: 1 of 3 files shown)


lassoan:
I'm wondering if we should use the `#3dslicer` hashtag (https://www.linkedin.com/feed/hashtag/?keywords=3dslicer). The main advantage is that people can follow a hashtag, but they cannot easily follow a search term.

Maybe we could also try to find out who registered https://www.linkedin.com/company/3dslicer and then start using it. Linkedin is quite nice in that many people can be appointed who can administer a "company" site.

jamesobutler:
Good idea! I've updated the link to https://www.linkedin.com/feed/hashtag/?keywords=3dslicer and can see that Kitware has used that hashtag as well so it makes sense here.

I've also updated the link for https://github.com/Slicer/slicer.org/pull/285.

lassoan (approved):
It looks good, thank you, let's merge this. Later if we can take ownership of the "3D Slicer" company site then we may switch to that.

PR #8025, opened 2024-11-01
BUG: Fix vtkMRMLTextStorageNode::WriteDataInternal not throwing error

Sunderlandkyl:
When the storage node was unable to open the file for some reason (ex. due to a file name that is too long), the storage node did not report an error, and returned success. Fixed by checking that the file was succesfully opened with std::ofstream::is_open and returning with failure if it isn't open.

### DIFF
--- a/Libs/MRML/Core/vtkMRMLTextStorageNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLTextStorageNode.cxx
@@ -194,15 +194,23 @@ int vtkMRMLTextStorageNode::WriteDataInternal(vtkMRMLNode * refNode)
   {
     if (!vtksys::SystemTools::RemoveFile(fullName.c_str()))
     {
-      vtkErrorMacro("WriteData: Could not overwrite existing file");
+      vtkErrorToMessageCollectionMacro(this->GetUserMessages(), "vtkMRMLTextStorageNode::WriteDataInternal",
+        "Text file '" << fullName.c_str() << "' could not be overwritten while trying to write (" << (this->ID ? this->ID : "(unknown)") << ").");
+      return 0;
     }
   }
 
   std::ofstream file;
   file.open(fullName);
+  if (!file.is_open())
+  {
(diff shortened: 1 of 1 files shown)


Libs/MRML/Core/vtkMRMLTextStorageNode.cxx
```diff
@@ -200,9 +200,14 @@ int vtkMRMLTextStorageNode::WriteDataInternal(vtkMRMLNode * refNode)
 
   std::ofstream file;
   file.open(fullName);
+  if (!file.is_open())
+  {
+    vtkErrorMacro("WriteData: Could not open file for writing");
```
lassoan, line 205: This is an error that we want to report to the users as well, so we should use `vtkErrorToMessageCollectionMacro` (that logs the error and also adds it to the user message collection). See for example in https://github.com/Slicer/Slicer/blob/main/Libs/MRML/Core/vtkMRMLModelStorageNode.cxx

lassoan (commented):
Good catch! It looks good, I've just added a small comment inline.

Libs/MRML/Core/vtkMRMLTextStorageNode.cxx
Sunderlandkyl: Done!

lassoan (commented):
Great thank you! However, I've just realized that the error message does not tell which file or node was problematic. This makes it difficult to determine what went wrong when saving as .mrb or when only the application log is available.

It would be better to do something like this:

```
    vtkErrorToMessageCollectionMacro(this->GetUserMessages(), "vtkMRMLSegmentationStorageNode::ReadDataInternal",
      "Segmentation file '" << fullName.c_str() << "' is not found while trying to read node (" << (this->ID ? this->ID : "(unknown)") << ").");
```

If it is not too much trouble then please include these additional details.

Sunderlandkyl:
@lassoan Ok, done!

lassoan (approved):
Perfect, thank you.

PR #8026, opened 2024-11-01
ENH: Update Ruff configuration to enforce improved python import sorting

### LINKED ISSUE #6272: Improve lint workflow to verify python import ordering

Guidelines for organizing imports are documented in https://github.com/Slicer/Slicer/blob/master/Docs/developer_guide/style_guide.md#python

Some of the ordering has been addressed in https://github.com/Slicer/Slicer/pull/6267 but the lack of an automated process will hinder and slow down the review process.

Evaluate the integration of `flake8-import-order`

See https://pypi.org/project/flake8-import-order/

NA

## Additional context

NA

jamesobutler:
This closes #6272.

This utilizes the isort package set of rules through Ruff. https://docs.astral.sh/ruff/settings/#lintisort

Based on https://github.com/Slicer/Slicer/pull/6267#issue-1169286737:
```
# standard library imports (os, sys, ...)

# third-party imports (ctk, numpy, qt, vtk, ...)

# application imports (slicer, ...)

# local imports
```

The isort portion of the Ruff configuration has lots of options than can be customized. So if we desire any changes from this first work, please let me know. For example I currently have the default of sorting imports alphabetically in each section, but it could also be sorted by line-length, but personally alphabetically is probably easier in terms of finding if an import is already present or not when editing the import sections manually. Otherwise the import line length configuration would change
```python
import collections
import os
```
to 
```python
import os
import collections
```
I have created a special import grouping section for all the "slicer" named imports while I have the local folder section below it containing things like Scripted Loadable module imports, but also vtkITK, vtkAddon and other miscellaneous libraries primarily managed by this Slicer GitHub repository. I could see `vtkITK` maybe making sense in the third party section next to `vtk`, but as Slicer repo specific I have it in this local folder section.

On the Windows platform, I have built this branch locally and run all tests which passed successfully (similar to current nightly).


lassoan (commented):
Thank you for your contribution! Sorting and grouping imports can help a little with reducing duplicate imports. Updating all these files must have been significant effort so we should close the PR quickly.

My only concern is that this grouping and sorting is a quite costly, both in term of mental effort it requires to accuretly follow the requirements and the additional whitespace that is added to the code.

What do you think about the following tweaks:
- Only group imports if it is more than just a few. For example, if there are more than 3 or 5 in at least one category; or more than 10 in total.
- Merge local and application imports, as there are just not enough imports that could sustain 4 categories (and the distinction between them is quite arbitrary anyway)
- Add description of this recommended style to the https://slicer.readthedocs.io/en/latest/developer_guide/style_guide.html

jamesobutler:
> Add description of this recommended style to the https://slicer.readthedocs.io/en/latest/developer_guide/style_guide.html

Thoughts about not doing this specifically but calling out to run pre-commit which autofixes most everything or calls out what is wrong? For example we don’t add to the style guide everything that Ruff is enforcing. I don’t aim to write everything manually in the correct style when pre-commit will auto fix it for me when I run it locally prior to committing.

jamesobutler:
> - Merge local and application imports, as there are just not enough imports that could sustain 4 categories (and the distinction between them is quite arbitrary anyway)
> - Add description of this recommended style to the https://slicer.readthedocs.io/en/latest/developer_guide/style_guide.html

^ The following items have been completed.

jamesobutler:
Abandoning this work based on the previously discussed complexity of different groupings.

Issue #7901, opened 2024-08-22
Cannot save scene to MRB on Windows if some node names are long


cpinter:
## Summary

Posting this as an issue because it comes up more and more often. If I want to save a scene to MRB that has at least one node with a name longer than ~60 characters (which is not at all unreasonably long), then saving the scene fails, even if I save it to the root directory (e.g. `D:\`).

Such long node names are quite frequent for example in RT data (such as `2: RTDOSE: DOSIsoft:RTDOSE:Phase #1 Dosi Dosi 1: Beam setup 1 (GY) [11]` in the data provided in https://discourse.slicer.org/t/creating-a-module-for-electron-bolus/37766). Another case when I have such long names is when I use a module that combines nodes, such as `ComparisonSubtraction_[NodeName1]_[NodeName2]`. So to me it seems that it is quite standard having a node with a name longer than a few dozen characters.

Important to note that if you save the scene as `.mrml` + data files then saving does not fail in this case.

The relevant part of the log:
```
[Qt] void __cdecl qSlicerSaveDataDialogPrivate::updateStatusIconFromMessageCollection(int,class vtkMRMLMessageCollection *,bool) Data save error: "- Error: MRHeadzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzz (vtkMRMLScalarVolumeNode1): Failed to write 'C:/Users/CsabaPintér/AppData/Local/slicer.org/Slicer/cache/SlicerIO/__BundleSaveTemp-2024-08-22_124836_56/2024-08-22-Scene/Data/TempWriteMRHeadzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzz/MRHeadzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzz.nrrd'\n- Error: Failed to save D:/zzzzzzzzzzzzzzzzzzzz/2024-08-22-Scene.mrb: Failed to save scene to data bundle directory\n- Error: Failed to save scene as D:/zzzzzzzzzzzzzzzzzzzz/2024-08-22-Scene.mrb\n- Error: Saving failed with all writers found for file 'D:/zzzzzzzzzzzzzzzzzzzz/2024-08-22-Scene.mrb' of type 'SceneFile'.\n"
```

## Steps to reproduce
[...]

PR #8028, opened 2024-11-04
BUG: Fix saving with long node names

Sunderlandkyl:
When saving nodes with long names on Windows, the file creation and writing would be unsuccessful. Since the error was at the system level, the error message would often be unhelpful for the user.

This PR clamps filename length for nodes with long names using `vtkMRMLScene::ClampFileNameLength`.

The format of the clamped string is:

```
[prefix of first 20 characters of the base name]_[4 character hash code].[extension]
```

The remaining characters after the suffix are used to seed the random number generator used to create the hash code.

Re #7901

### DIFF
--- a/Base/QTCore/Testing/Cxx/qSlicerCoreIOManagerTest1.cxx
+++ b/Base/QTCore/Testing/Cxx/qSlicerCoreIOManagerTest1.cxx
@@ -26,10 +26,63 @@
 #include "qSlicerCoreIOManager.h"
 
 // MRML includes
+#include <vtkMRMLMessageCollection.h>
+#include "vtkMRMLTextNode.h"
 #include "vtkMRMLScene.h"
-#include "vtkMRMLStorableNode.h"
+#include "vtkMRMLStorageNode.h"
 
 #include "vtkMRMLCoreTestingMacros.h"
+
+int TestLongNodeNameSaving(const char* temporaryDirectory)
+{
+  vtkNew<vtkMRMLScene> scene;
+  if (temporaryDirectory)
+  {
+    scene->SetRootDirectory(temporaryDirectory);
+  }
+
+  std::string longNodeName =
+    "Loremipsumdolorsitametconsecteturadipiscingelitseddoeiusmodtemporin"
+    "cididuntutlaboreetdoloremagnaaliquaUtenimadminimveniamquisnostrudex"
+    "ercitationullamcolaborisnisiutaliquipexeacommodoconsequatDuisauteir"
+    "uredolorinreprehenderitinvoluptatevelitessecillumdoloreeufugiatnull"
+    "apariaturExcepteursintoccaecatcupidatatnonproidentsuntinculpaquioff"
+    "iciadeseruntmollitanimidestlaborum";
+  std::string extension = ".txt";
+  std::string longFileName = longNodeName + extension;
+  std::string safeFileName = qSlicerCoreIOManager::forceFileNameValidCharacters(QString::fromStdString(longFileName)).toStdString();
+  safeFileName = qSlicerCoreIOManager::forceFileNameMaxLengthExtension(QString::fromStdString(longFileName), extension.length()).toStdString();
+
+  vtkSmartPointer<vtkMRMLTextNode> textNode = vtkMRMLTextNode::SafeDownCast(
+    scene->AddNewNodeByClass("vtkMRMLTextNode", longNodeName));
+  textNode->SetText(longNodeName);
+  textNode->SetForceCreateStorageNode(true);
+  textNode->AddDefaultStorageNode();
+  vtkMRMLStorageNode* storageNode = textNode->GetStorageNode();
+
+#ifdef Q_OS_WIN
+  std::cout << std::endl << "||||||||||||||||||||" << std::endl;
+  std::cout << std::endl << "Testing long file name: " << longFileName << std::endl;
+  storageNode->SetFileName(longFileName.c_str());
+  CHECK_INT(storageNode->WriteData(textNode), 0); // Writing should fail. File name too long.
+#endif
+
+  std::cout << std::endl << "||||||||||||||||||||" << std::endl;
+  std::cout << std::endl << "Testing safe file name: " << safeFileName << std::endl;
+  storageNode->SetFileName(safeFileName.c_str());
+  storageNode->GetUserMessages()->ClearMessages();
+  CHECK_INT(storageNode->WriteData(textNode), 1); // Writing should succeed.
+
+  std::cout << std::endl << "||||||||||||||||||||" << std::endl;
+  std::cout << std::endl << "Testing scene save with long node name: " << longNodeName << std::endl;
+  storageNode->SetFileName("");
+  std::stringstream scenePathSS;
+  scenePathSS << scene->GetRootDirectory() << "/" << "Loremipsum.mrb";
+  std::string scenePath = scenePathSS.str();
+  CHECK_BOOL(scene->WriteToMRB(scenePath.c_str()), true); // Scene should automatically shorten filename.
+
+  return EXIT_SUCCESS;
+}
 
 int qSlicerCoreIOManagerTest1(int argc, char * argv [])
 {
@@ -104,6 +157,17 @@ int qSlicerCoreIOManagerTest1(int argc, char * argv [])
     }
     qDebug() << "Found extension " << ext << " from file " << testFileNames[i] << " using " << storageNodeClassNames[i];
   }
+
+  const char* temporaryDirectory = nullptr;
+  if (argc > 1)
+  {
+    temporaryDirectory = argv[1];
+  }
+  else
+  {
+    temporaryDirectory = app.mrmlScene()->GetRootDirectory();
+  }
+  CHECK_EXIT_SUCCESS(TestLongNodeNameSaving(temporaryDirectory));
 
   return EXIT_SUCCESS;
 }
--- a/Base/QTCore/qSlicerCoreIOManager.cxx
+++ b/Base/QTCore/qSlicerCoreIOManager.cxx
@@ -482,8 +482,20 @@ QString qSlicerCoreIOManager::forceFileNameValidCharacters(const QString& filena
 
   // Remove leading and trailing spaces
   sanitizedFilename = sanitizedFilename.trimmed();
-
   return sanitizedFilename;
+}
+
+//-----------------------------------------------------------------------------
+QString qSlicerCoreIOManager::forceFileNameMaxLength(const QString& filename, int maxLength/*=-1*/)
+{
+  QString extension = this->extractKnownExtension(filename, nullptr);
+  return qSlicerCoreIOManager::forceFileNameMaxLengthExtension(filename, maxLength, extension.length());
+}
+
+//-----------------------------------------------------------------------------
+QString qSlicerCoreIOManager::forceFileNameMaxLengthExtension(const QString& filename, int extensionLength, int maxLength/*=-1*/)
+{
+  return QString::fromStdString(vtkMRMLStorageNode::ClampFileNameExtension(filename.toStdString(), maxLength, 4, extensionLength));
 }
 
 //-----------------------------------------------------------------------------
--- a/Base/QTCore/qSlicerCoreIOManager.h
+++ b/Base/QTCore/qSlicerCoreIOManager.h
@@ -109,6 +109,15 @@ public:
   /// Remove characters that are likely to cause problems in a filename
   Q_INVOKABLE static QString forceFileNameValidCharacters(const QString& filename);
 
+  /// Clamp the length of a filename to a maximum number of characters.
+  /// The file extension will be detected and excluded from the specified maximum length.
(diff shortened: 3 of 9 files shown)


Libs/MRML/Core/vtkMRMLScene.cxx
```diff
@@ -4678,3 +4680,46 @@ bool vtkMRMLScene::ParseVersion(const char* versionString, std::string& applicat
   revision = atoi(revisionStr.c_str());
   return true;
 }
+
+//-----------------------------------------------------------------------------
+std::string vtkMRMLScene::ClampFileNameLength(const std::string& filename, int maxFileNameLength/*=25*/, int hashLength/*=4*/)
+{
+  std::string sanitizedBaseName = vtksys::SystemTools::GetFilenameWithoutExtension(filename);
```
lassoan, line 4687: remove longest known file extension using the storage node

Libs/MRML/Core/vtkMRMLScene.cxx
```diff
@@ -4678,3 +4680,46 @@ bool vtkMRMLScene::ParseVersion(const char* versionString, std::string& applicat
   revision = atoi(revisionStr.c_str());
   return true;
 }
+
+//-----------------------------------------------------------------------------
+std::string vtkMRMLScene::ClampFileNameLength(const std::string& filename, int maxFileNameLength/*=25*/, int hashLength/*=4*/)
```
lassoan, line 4685: move this method to the storage node (scene class is already very large)

Libs/MRML/Core/vtkMRMLScene.cxx
```diff
@@ -4678,3 +4680,46 @@ bool vtkMRMLScene::ParseVersion(const char* versionString, std::string& applicat
...
+
+  // Generate a random 4 character hash code from 4 upper case letters and numbers
+  std::string hashChars = "ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789";
+  for (int i = 0; i < hashLength; ++i)
+  {
+    int index = static_cast<int>(std::floor(std::fmod(randomSequence->GetNextValue(), 1.0) * hashChars.size()));
+    truncatedFilenameSS << hashChars[index];
+  }
+
+  std::string extension = vtksys::SystemTools::GetFilenameExtension(filename);
```
lassoan, line 4718: Always use the storage node for such operations. vtksys::SystemTools does not know what a file extension is (it only has the option to use the first or last `.`), but cannot recognize actual file extensions.

Base/QTCore/qSlicerCoreIOManager.cxx
```diff
@@ -483,6 +483,9 @@ QString qSlicerCoreIOManager::forceFileNameValidCharacters(const QString& filena
   // Remove leading and trailing spaces
   sanitizedFilename = sanitizedFilename.trimmed();
 
+  // Ensure the filename doesn't exceed the maximum allowed length
+  sanitizedFilename = QString::fromStdString(vtkMRMLScene::ClampFileNameLength(sanitizedFilename.toStdString()));
```
lassoan, line 487: add to the method documentation that valid characters include the total number of characters (so long filenames will be shortened)

lassoan (commented):
This will be great, thank you, just added a few comments inline.

Sunderlandkyl:
@lassoan If the user changes the file name in the save dialog, and it is longer than the max allowed length, should we update the file name or leave it as is, allowing it to potentially fail?

lassoan:
We should let the user choose a long name. We should just try to make the error message as helpful as possible (maybe add to the messages that the long name may be an issue if it is detected that the path length is long).

Base/QTCore/qSlicerCoreIOManager.cxx
```diff
@@ -482,10 +482,22 @@ QString qSlicerCoreIOManager::forceFileNameValidCharacters(const QString& filena
...
   // Remove leading and trailing spaces
   sanitizedFilename = sanitizedFilename.trimmed();
-
   return sanitizedFilename;
 }
 
+//-----------------------------------------------------------------------------
+QString qSlicerCoreIOManager::forceFileNameMaxLength(const QString& filename, int maxLength/*=25*/)
+{
+  QString extension = extractKnownExtension(filename, nullptr);
```
lassoan, line 491: add `this->` or `qSlicerCoreIOManager::` to make it more clear where the function comes from

Libs/MRML/Core/vtkMRMLStorageNode.h
```diff
@@ -375,6 +375,17 @@ class VTK_MRML_EXPORT vtkMRMLStorageNode : public vtkMRMLNode
...
   vtkMRMLMessageCollection *GetUserMessages() { return this->UserMessages; }
 
+  //@{
+  /// Ensures that the file name (excluding the extension) is shorter than the maximum allowed length.
+  /// If the filename is shorter than the maximum allowed length then it is returned unchanged.
+  /// If the filename is longer than the maximum allowed length then the filename is shortened by using the following format:
+  /// [first 20 characters of the base name]_[4 character hash code].[extension]
+  /// The length of the prefix is the maximum allowed length minus the length of the hash code plus one for the added underscore.
+  /// The full base name of the file will be exactly maxFileNameLength characters long.
+  std::string ClampFileName(const std::string& filename, int maxFileNameLength = 25, int hashLength = 4);
```
lassoan, line 385: It is OK to hardcode the 25 but it is error-prone to hardcode it at multiple places. Let's add a static function in the vtkMRMLStorageNode and use that everywhere to specify the filename (and maybe also do it for the hash length). Defaults could be set to `-1`, which would mean that the `value returned by vtkMRMLStorageNode::GetPreferredFileNameLength()` would be used. Maybe `Preferred` is not the best word, it would be just important to qualify what kind of maximum is this (distinguish from the file system or other hard maximum lengths).

Base/QTGUI/qSlicerSaveDataDialog.cxx
```diff
@@ -1370,6 +1387,77 @@ void qSlicerSaveDataDialogPrivate::onItemChanged(QTableWidgetItem* widgetItem)
...
+
+  // Check if the file name is too long
+  QStringList pathLengthWarningMessages;
+  QString shortFileName = qSlicerCoreIOManager::forceFileNameMaxLengthExtension(fileName, extension.length());
+  if (shortFileName != fileName)
+  {
+    pathLengthWarningMessages << qSlicerSaveDataDialog::tr("File name may be too long.");
+  }
+
+  int maxPathLength = 260;
```
lassoan, line 1445: The limit should be lower, around 200, but I think we should only log this warning if saving failed.

lassoan (commented):
Thank you, it is getting close to be ready, just added a couple of comments.

Libs/MRML/Core/vtkMRMLStorageNode.h
Sunderlandkyl: I've used "GetRecommendedFileNameLength" for the function name.

Base/QTGUI/qSlicerSaveDataDialog.cxx
Sunderlandkyl: Done.

lassoan (commented):
It looks good to me, thank you. Just added one question inline.

Base/QTGUI/qSlicerSaveDataDialog_p.h
```diff
@@ -137,6 +137,8 @@ protected slots:
   QIcon WarningIcon;
   QIcon ErrorIcon;
 
+  bool IconUpdating;
```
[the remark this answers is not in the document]
Sunderlandkyl: Nope, not anymore, removed.

Base/QTCore/qSlicerCoreIOManager.h
```diff
@@ -109,6 +109,14 @@ class Q_SLICER_BASE_QTCORE_EXPORT qSlicerCoreIOManager:public QObject
   /// Remove characters that are likely to cause problems in a filename
   Q_INVOKABLE static QString forceFileNameValidCharacters(const QString& filename);
 
+  /// Clamp the length of a filename to a maximum number of characters.
+  /// The file extension will be detected and excluded from the specified maximum length.
```
jcfr, line 113:
```suggestion
  /// The file extension will be detected and excluded from the specified maximum length.
  /// \sa qSlicerCoreIOManager::stripKnownExtension()
```

Base/QTGUI/qSlicerExportNodeDialog.cxx
```diff
@@ -357,7 +357,9 @@ QString qSlicerExportNodeDialogPrivate::defaultFilename(vtkMRMLNode* node, QStri
     qCritical() << Q_FUNC_INFO << "failed: Core IO manager not found.";
     return QString();
   }
-  const QString safeNodeName = qSlicerCoreIOManager::forceFileNameValidCharacters(unsafeNodeName);
+
+  QString safeNodeName = qSlicerCoreIOManager::forceFileNameValidCharacters(unsafeNodeName);
+  safeNodeName = qSlicerCoreIOManager::forceFileNameMaxLengthExtension(safeNodeName, 0);
   return forceFileNameExtension(safeNodeName, extension, node);
```
jcfr, line 363:
```suggestion
  return qSlicerExportNodeDialogPrivate::forceFileNameExtension(safeNodeName, extension, node);
```

Sunderlandkyl: Thanks @jcfr, makes sense to me. Changes made.

Base/QTCore/Testing/Cxx/qSlicerCoreIOManagerTest1.cxx
```diff
@@ -26,11 +26,62 @@
...
+    scene->AddNewNodeByClass("vtkMRMLTextNode", longNodeName));
+  textNode->SetText(longNodeName);
+  textNode->SetForceCreateStorageNode(true);
+  textNode->AddDefaultStorageNode();
+  vtkMRMLStorageNode* storageNode = textNode->GetStorageNode();
+
+  std::cout << std::endl << "||||||||||||||||||||" << std::endl;
+  std::cout << std::endl << "Testing long file name: " << longFileName << std::endl;
+  storageNode->SetFileName(longFileName.c_str());
+  CHECK_INT(storageNode->WriteData(textNode), 0); // Writing should fail. File name too long.
```
lassoan, line 66: I believe this only fails on Windows (and only if long file path support is disabled in the OS, but usually it is).
We could try to make it fail differently on Linux/macOS (for example by injecting some invalid characters into the filepath); or we could just skip these checks on Linux/macOS.

Base/QTCore/Testing/Cxx/qSlicerCoreIOManagerTest1.cxx
```diff
@@ -26,11 +26,62 @@
...
+
+  std::cout << std::endl << "||||||||||||||||||||" << std::endl;
+  std::cout << std::endl << "Testing safe file name: " << safeFileName << std::endl;
+  storageNode->SetFileName(safeFileName.c_str());
+  storageNode->GetUserMessages()->ClearMessages();
+  CHECK_INT(storageNode->WriteData(textNode), 1); // Writing should fail. File name should be shortened.
```
lassoan, line 72:
```suggestion
  CHECK_INT(storageNode->WriteData(textNode), 1); // Writing should succeed.
```

Base/QTGUI/qSlicerSaveDataDialog.cxx
```diff
@@ -868,6 +870,19 @@ bool qSlicerSaveDataDialogPrivate::saveNodes()
...
         snode->GetUserMessages()->AddMessage(vtkCommand::ErrorEvent,
           (qSlicerSaveDataDialog::tr("Cannot write data file: %1.").arg(file.absoluteFilePath())).toStdString());
+
+        // Add warning messages if the file name or path is too long
+        if (file.fileName().length() > vtkMRMLStorageNode::GetRecommendedFileNameLength())
+        {
+          snode->GetUserMessages()->AddMessage(vtkCommand::WarningEvent,
+            (qSlicerSaveDataDialog::tr("File name may be too long: %1.").arg(file.fileName())).toStdString());
+        }
+        if (file.absoluteFilePath().length() > 200)
```
lassoan, line 880:
```suggestion
        // Maximum file path is 260 characters on most Windows systems. Warn the user if the path
        // is long and therefore may fail to be saved or cause compatibility issues later.
        // The warning limit is set to somewhat lower than the 260 limit, to add some safety margin
        // (e.g., to avoid issues when the file is moved into another folder).
        if (file.absoluteFilePath().length() > 200)
```

lassoan (commented):
Thank you, just added a few trivial comments inline.

Base/QTCore/Testing/Cxx/qSlicerCoreIOManagerTest1.cxx
Sunderlandkyl: Added #ifdef _WIN32 around the section.

lassoan (approved):
Thank you, it looks good.

PR #8031, opened 2024-11-06
BUG: Fix incorrect plane interaction scale handles initialization

Sunderlandkyl:
The plane interaction handles and scalar arrays were initialized with an incorrect number of points. Later calls to UpdatePlaneScaleHandles set the number of points correctly, but don't update the scalar arrays to the correct size. This rarely caused a crash when trying to call GetInteractionHandlePositionWorld with an index that was out of range.

Fixed by initializing scale handles with the correct number of points, and by checking to make sure the index is valid in GetInteractionHandlePositionWorld before calling vtkPolyData::GetPoint().

### DIFF
--- a/Modules/Loadable/Markups/VTKWidgets/vtkSlicerMarkupsInteractionWidgetRepresentation.cxx
+++ b/Modules/Loadable/Markups/VTKWidgets/vtkSlicerMarkupsInteractionWidgetRepresentation.cxx
@@ -706,23 +706,22 @@ void vtkSlicerMarkupsInteractionWidgetRepresentation::CreateScaleHandles()
   if (planeNode || roiNode)
   {
     vtkNew<vtkPoints> points;
-    points->InsertNextPoint(distance, 0.0, 0.0); // X-axis +ve
-    points->InsertNextPoint(0.0, distance, 0.0); // Y-axis +ve
-    points->InsertNextPoint(0.0, 0.0, distance); // Z-axis +ve
-    points->InsertNextPoint(-distance, 0.0, 0.0); // X-axis -ve
-    points->InsertNextPoint(0.0, -distance, 0.0); // Y-axis -ve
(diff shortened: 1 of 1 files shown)


Modules/Loadable/Markups/VTKWidgets/vtkSlicerMarkupsInteractionWidgetRepresentation.cxx
```diff
@@ -706,23 +706,29 @@ void vtkSlicerMarkupsInteractionWidgetRepresentation::CreateScaleHandles()
   if (planeNode || roiNode)
   {
     vtkNew<vtkPoints> points;
-    points->InsertNextPoint(distance, 0.0, 0.0); // X-axis +ve
-    points->InsertNextPoint(0.0, distance, 0.0); // Y-axis +ve
-    points->InsertNextPoint(0.0, 0.0, distance); // Z-axis +ve
-    points->InsertNextPoint(-distance, 0.0, 0.0); // X-axis -ve
-    points->InsertNextPoint(0.0, -distance, 0.0); // Y-axis -ve
-    points->InsertNextPoint(0.0, 0.0, -distance); // Z-axis -ve
+    points->InsertNextPoint(distance, 0.0, 0.0); // +ve X
```
lassoan, line 709: I think it would be more clear to have two completely separate branches for planeNode and roiNode. It would just duplicate 4 lines of code, and we could get rid of a couple of if lines, so overall the code may even be shorter.

lassoan (commented):
Thank you, this looks good to me. Just added a minor comment on the organization of the code.

Modules/Loadable/Markups/VTKWidgets/vtkSlicerMarkupsInteractionWidgetRepresentation.cxx
Sunderlandkyl: Ok, done.

lassoan (approved):
Perfect, thank you!

PR #8034, opened 2024-11-11
DOC: Clarify doc of vtkMRMLMarkupsROINode::GetObjectToWorldMatrix

shai-ikko:
Per https://discourse.slicer.org/t/unexpected-interactions-of-roi-objects-with-code/40053/3

### DIFF
--- a/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsROINode.h
+++ b/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsROINode.h
@@ -129,7 +129,7 @@ public:
   //@}
 
   /// 4x4 matrix defining the object center and axis directions within the world coordinate system.
-  /// Changes made to the matrix will not be applied.
+  /// Users should avoid modifying this matrix.
   vtkMatrix4x4* GetObjectToWorldMatrix()
   {
     return this->ObjectToWorldMatrix;
--- a/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsROINode.h
+++ b/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsROINode.h
@@ -129,7 +129,7 @@ public:
   //@}
 
   /// 4x4 matrix defining the object center and axis directions within the world coordinate system.
-  /// Users should avoid modifying this matrix.
+  /// The returned matrix must not be modified.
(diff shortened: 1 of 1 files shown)


Modules/Loadable/Markups/MRML/vtkMRMLMarkupsROINode.h
```diff
@@ -129,7 +129,7 @@ class  VTK_SLICER_MARKUPS_MODULE_MRML_EXPORT vtkMRMLMarkupsROINode : public vtkM
   //@}
 
   /// 4x4 matrix defining the object center and axis directions within the world coordinate system.
-  /// Changes made to the matrix will not be applied.
+  /// Users should avoid modifying this matrix.
   vtkMatrix4x4* GetObjectToWorldMatrix()
```
shai-ikko, line 133: This follows the suggestion of @lassoan on the [forum](https://discourse.slicer.org/t/unexpected-interactions-of-roi-objects-with-code/40053/3). However, it raises the question -- if (external) users are not to modify this matrix, why is the return type not `const vtkMatrix4x4*` (and then, why not make the method itself `const`)?

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsROINode.h
```diff
@@ -129,7 +129,7 @@ class  VTK_SLICER_MARKUPS_MODULE_MRML_EXPORT vtkMRMLMarkupsROINode : public vtkM
   //@}
 
   /// 4x4 matrix defining the object center and axis directions within the world coordinate system.
-  /// Changes made to the matrix will not be applied.
+  /// Users should avoid modifying this matrix.
```
lassoan, line 132:
```suggestion
  /// The returned matrix must not be modified.
```

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsROINode.h
lassoan: VTK is not const correct, so you cannot just add `const` wherever you want, and it would not help anyway when the API is used from Python.

lassoan (approved):
Thank you for your contribution!

shai-ikko:
Thank you!

Issue #7403, opened 2023-11-17
Update TEST_SET_GET_STRING to support both char* and std::string


jcfr:
## Is your feature request related to a problem? Please describe.

When building Slicer using the Centos 7 (GCC 7) build environment[^1], warnings like this one are reported:

[^1]: https://github.com/Slicer/Slicer/pull/7401

```
Slicer/Libs/MRML/Core/Testing/vtkMRMLSceneViewNodeTest1.cxx:
 In function ‘int vtkMRMLSceneViewNodeTest1(int, char**)’:
Slicer/Libs/MRML/Core/vtkMRMLCoreTestingMacros.h:328:62:
 warning: ‘vtkStdString::operator const char*()’ is deprecated: Call `.c_str()` explicitly [-Wdeprecated-declarations]
   const char * originalStringPointer = object->Get##variable(); \
```

also raised for example in https://github.com/Slicer/Slicer/actions/runs/6885650499/job/187301131

## Describe the solution you'd 

Since `TEST_SET_GET_STRING` is also used for `char*`[^2][^3], addressing this could instead be done using a template based approach where the compiler would choose the specialization based on the type.

_Originally posted by @jcfr in https://github.com/Slicer/Slicer/pull/7401#pullrequestreview-1735762477_

[^2]: https://github.com/Slicer/Slicer/blob/b41abc4f69a7f99253f6880733be97186dbfbd77/Libs/MRML/Core/vtkMRMLNode.h#L995-L997
[^3]: https://github.com/Slicer/Slicer/blob/b41abc4f69a7f99253f6880733be97186dbfbd77/Libs/MRML/Core/vtkMRMLCoreTestingUtilities.cxx#L207C13-L209

## Describe alternatives you've considered

NA

## Additional context

* https://github.com/Slicer/Slicer/pull/7401
* https://github.com/Slicer/Slicer/pull/7411
* https://github.com/Slicer/Slicer/pull/7419

lassoan:
To fix the warnings, we should use the appropriate test macros.

- `CHECK_STRING...` macros are for classic VTK string (that are get/set as `char*`) properties
- `CHECK_STD_STRING...` macros are for std::string properties (see for example [here](https://github.com/Slicer/SlicerGitSVNArchive/blob/65e92bb16c2b32ea47a1a66bee71f238891ee1ca/Libs/vtkAddon/vtkAddonTestingMacros.h#L168))

PR #8041, opened 2024-11-12
COMP: Resolve deprecated warning in MRML testing macro for std::string

jcfr:
This commit is follow-up of 9b8c82356b ("BUG: Revert invalid -Wdeprecated-declarations fix in TEST_SET_GET_STRING", 2023-11-24) and is intended to fix warnings like the following by adding `TEST_SET_GET_STD_STRING` MRML testing macro to explicitly test set/get of `std::string` (instead of relying on the deprecated operator const char*().

Warning fixed:

```
   const char * originalStringPointer = object->Get##variable();
```

Related pull requests:
* https://github.com/Slicer/Slicer/pull/7401 
* https://github.com/Slicer/Slicer/pull/7411
* https://github.com/Slicer/Slicer/pull/7419

Related issues:
* https://github.com/Slicer/Slicer/issues/7403

### DIFF
--- a/Libs/MRML/Core/Testing/vtkMRMLSceneViewNodeTest1.cxx
+++ b/Libs/MRML/Core/Testing/vtkMRMLSceneViewNodeTest1.cxx
@@ -41,7 +41,7 @@ int vtkMRMLSceneViewNodeTest1(int , char * [] )
 
   node1->SetAbsentStorageFileNames();
 
-  TEST_SET_GET_STRING( node1.GetPointer(), SceneViewDescription);
+  TEST_SET_GET_STD_STRING( node1.GetPointer(), SceneViewDescription);
 
   node1->SetScreenShot(nullptr);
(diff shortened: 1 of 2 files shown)


jcfr:
Once this is integrated, I will follow-up with another pull request replacing `vtkStdString` by  `std::string` :rocket: 

> In the future, `vtkStdString` itself will be deprecated in preference for just using `std::string` directly.

Source: kitware/VTK@13466f8281bdd41f560d18ab72608e1045d5a632


Related commits:
* kitware/VTK@db852db9712a8ea5a0c240e7742e8cd4840caa95: `Prefer std::string over vtkStdString or const char*`
* kitware/VTK@13466f8281bdd41f560d18ab72608e1045d5a632: `vtkStdString: deprecate the implicit `const char*` conversion`

lassoan:
Thanks for working on this. Switching to std::string from vtkStdString will be great.

lassoan (approved):
It looks good, thank you.

PR #8042, opened 2024-11-13
COMP: Replace vtkStdString with std::string

jcfr:
`vtkStdString`, a subclass of `std::string`, was introduced in the pre-C++11 era to produce shorter symbols in compiled libraries. It is now being deprecated in favor of using `std::string` directly (see Kitware/VTK@13466f828). This commit updates both the implementation and API by replacing `vtkStdString` with `std::string`.

Since `vtkStdString` includes a constructor that accepts an `std::string`, functions expecting `vtkStdString` parameters will seamlessly accept `std::string` as well. For example, this applies to functions associated with `vtkStringArray` class.

### DIFF
--- a/Base/Logic/vtkSlicerScriptedLoadableModuleLogic.cxx
+++ b/Base/Logic/vtkSlicerScriptedLoadableModuleLogic.cxx
@@ -29,7 +29,6 @@
 
 // VTK includes
 #include <vtkPythonUtil.h>
-#include <vtkStdString.h>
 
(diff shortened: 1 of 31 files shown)


jcfr:
Follow-up of https://github.com/Slicer/Slicer/pull/8041

jcfr:
@lassoan Any objection to move forward with this ?

lassoan (approved):
It looks good, thank you.

PR #8044, opened 2024-11-13
STYLE: Fix typo replacing Visibile with Visible

huwqchn:
Rename all occurrences of `Visibile` to `Visible` (for example `savedBoxVisibile` -> `savedBoxVisible`)

### DIFF
--- a/Libs/MRML/Core/vtkMRMLViewNode.h
+++ b/Libs/MRML/Core/vtkMRMLViewNode.h
@@ -290,7 +290,8 @@ public:
     RenderModeFlag,
     BoxVisibleFlag,
     BoxColorFlag,
-    BoxLabelVisibileFlag,
+    BoxLabelVisibleFlag,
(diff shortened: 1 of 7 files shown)


jcfr:
Notes:
* Rename of `onListVisibileInvisiblePushButtonClicked` to `onListVisibleInvisiblePushButtonClicked`: Considering that (1) no public code is referencing[^1] the function (beside of other Slicer fork) and (2) the slot became unused (not connecteD to any GUI element) following e1f905ba758219d7e7271a45cab89416e2525edf (Sept 17, 2019), the function was renamed
* `vtkMRMLViewNode::InteractionFlagType` was updated to keep the entry with the incorrect spelling so that existing code do not break.

[^1]: https://github.com/search?q=onListVisibileInvisiblePushButtonClicked&type=code

lassoan (approved):
Good catch, thanks for the fix!

PR #8052, opened 2024-11-19
ENH: Remove old and unused testing scripts

jamesobutler:
This removes TextureModel.py and ModelRender.py originally added in https://github.com/Slicer/Slicer/commit/bc9333e4ae33beabe193536a48d6bb18d19a811d. As noticed in these scripts there was usage such as:
```python
import Slicer
modelNode = Slicer.slicer.vtkMRMLModelNode()
```
where the import of `Slicer` instead of `slicer` does not work in today's latest Slicer application:
```
Python 3.9.10 (main, Nov 17 2024, 23:19:50) [MSC v.1939 64 bit (AMD64)] on win32
>>> import Slicer
ModuleNotFoundError: No module named 'Slicer'
```


jcfr (approved):

pieper:
Yes, no problem getting rid of these.  They are out of date and the same concepts are covered in the script repository.

PR #8058, opened 2024-11-19
SH folder visibility fix

cpinter:
Two small fixes in SH related to recent SlicerRT improvements and investigation.

No need to handle this until after the stable release.


cpinter:
Now that the stable is out and PW is over, I'd like to propose this PR for integration. Thanks a lot!

cpinter:
@lassoan What do you think? Is this a reasonable change? It fixes a real issue, although does not remove the workaround.

cpinter:
@lassoan Can you please give this a quick look? Thanks!

lassoan:
If there is an update issue then we should be able to address it by adding the missing RequestRender calls (and not with decreasing batch processing probability). Do your have a simple way to reproduce this update issue? 

Have you measured computation time vs. number of items to update; with/without batch processing? I think the breakeven point is way below 50 items, but we should really measure it. The best would be to have  a Python script that creates nodes and shows/hides them and reports the time it took with and without batch processing.

If needed then the logic of counting items can be tuned (e.g., exclude folders), but usually a folder contains at least 5-10 nodes, so excluding folders (or other rare item types that are not displayable nodes) should not make a huge difference.

cpinter:
Thanks for the review! I don't think I'll have time to do this analysis, so maybe I'll just make a fork for this simple change.

Can the commit that fixes the crash be integrated?

lassoan:
I've moved out the fix of the crash to https://github.com/Slicer/Slicer/pull/8701, as it does not require any further discussion or testing.

About the folder visibility: I've made some fixes a few days ago that may fix the folder visibility issues that you encountered. @cpinter could you test if you can still reproduce issues with the latest Slicer Preview Release? (if not, then the changes in the second commit of this pull request might not be needed anymore)

jcfr:
@cpinter @lassoan I rebased this pull request and I am updating the milestone to be `Slicer 5.11`. If you think this is (near-)ready for integration, I will let you update the milestone and move forward with final review & testing.

cpinter:
> could you test if you can still reproduce issues with the latest Slicer Preview Release?

Sorry, again, I missed a comment. I'll do this in the next days. Thanks!

cpinter:
I tested what I thought did not work well (i.e. load the RANDO^ENT dataset from SlicerRtData with everything and show/hide the entire plan), and now it works well, and it is fast too. I am closing this PR, and will open a new one if I ever encounter it again. Thanks a lot for the help and attention @lassoan and @jcfr !

jcfr:
:pray: @cpinter for taking the time to test and update the issue report.

PR #8062, opened 2024-11-21
BUG: Fix hanging progress popup in DICOM browser

jeffmax:
This fixes an issue described within this forum post: https://discourse.slicer.org/t/hanging-popup-some-times-after-loading-dicom-files/6057.  There is a dialog progress popup that appears when attempting to load a DICOM file from the DICOM browser that remains on screen after the load as completed. The original post submitter only had issues intermittently. Using Slicer within [Xpra](https://github.com/Xpra-org/xpra) I was experiencing it consistently and this code change resolved the issue.

### DIFF
--- a/Modules/Scripted/DICOMLib/DICOMBrowser.py
+++ b/Modules/Scripted/DICOMLib/DICOMBrowser.py
@@ -487,6 +487,7 @@ class SlicerDICOMBrowser(VTKObservationMixin, qt.QWidget):
                                                                                 lambda progressLabel, progressValue, progressDialog=progressDialog: progressCallback(progressDialog, progressLabel, progressValue),
                                                                                 self.pluginInstances)
 
+            progressDialog.setParent(None)
             progressDialog.close()
 
         if messages:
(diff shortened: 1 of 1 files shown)


jamesobutler:
I wonder if this is similar to https://github.com/Slicer/Slicer/blob/45366c6d4d6a769c87a4fc8552c50f8c5c3c428d/Base/Python/slicer/util.py#L3095-L3101 where you could still see message boxes in Windows aero peek even after they were closed. For ProgressDialog objects which are different from the message box. Calling `progress_dialog.deleteLater()` or `progress_dialog.delete()` may accomplish a similar thing to `progress_dialog.setParent(None)` which may allow deleting by undefining the parent.

pieper:
> this is similar to

Yes, could be.  I don't know if there's a reason to prefer one method over another.

jeffmax:
Yes, just noting that I did try .deleteLater() after the .close() command. That worked as well, as did doing both methods at once (.setParent(None), .close(), .deleteLater()).

I chose this because it seemed more likely to be getting at the root of the problem directly, but that was just a guess.

jeffmax:
Just wanted to check-in on this PR- happy to test another approach.

pieper (approved):
Hearing no further comments I'm going to go ahead and merge this.

PR #8063, opened 2024-11-22
Fix fit slice to volume

lassoan:
Fixed and fit slice to volume and made it work consistently for both when background clipping is enabled (default) or disabled.

image


Punzo (approved):
LGTM for me!

PR #8067, opened 2024-11-27
BUG: Fix crash when hovering over interaction handles after scene load

Sunderlandkyl:
After OnMRMLSceneEndBatchProcess was called, UpdateFromMRMLRequested was called, but RequestRender was not. This meant that the widget would not be updated until the user tried to mouse over the widget. This caused a crash in ProcessInteractionEvent, since the activeWidget was deleted during processing as UpdateFromMRML was resolved.

Fixed by:
- Calling RequestRender after SetUpdateFromMRMLRequested(True) in OnMRMLSceneEndClose() and OnMRMLSceneEndBatchProcess().
- Using vtkSmartPointer for activeWidget in ProcessInteractionEvent so that the pointer to the active widget can't be made invalid.


lassoan (approved):

PR #8068, opened 2024-11-27
BUG: Fix interaction handle enable/disable not working in views

Sunderlandkyl:
When right clicking in a view to call "toggleInteractionBox()", the current subject hierarchy ID was sometimes invalid, causing nothing to happen. Fixed by getting the transform node from "getTransformNodeForAction()", which works in both SH widgets and in views.

Also ensures that all context menu widgets are updated in "showViewContextMenuActionsForItem()".

### DIFF
--- a/Modules/Loadable/Transforms/SubjectHierarchyPlugins/qSlicerSubjectHierarchyTransformsPlugin.cxx
+++ b/Modules/Loadable/Transforms/SubjectHierarchyPlugins/qSlicerSubjectHierarchyTransformsPlugin.cxx
@@ -848,24 +848,21 @@ void qSlicerSubjectHierarchyTransformsPlugin::showVisibilityContextMenuActionsFo
     vtkMRMLTransformDisplayNode* displayNode = transformNode ? vtkMRMLTransformDisplayNode::SafeDownCast(transformNode->GetDisplayNode()) : nullptr;
 
     d->ToggleInteractionAction->setVisible(true);
-    bool wasBlocked = d->ToggleInteractionAction->blockSignals(true);
+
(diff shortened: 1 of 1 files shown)


Modules/Loadable/Transforms/SubjectHierarchyPlugins/qSlicerSubjectHierarchyTransformsPlugin.cxx
```diff
@@ -919,12 +919,25 @@ void qSlicerSubjectHierarchyTransformsPlugin::showViewContextMenuActionsForItem(
   d->showResetCenterOfTransformationForTransformedNodesActions(transformNode);
 
   vtkMRMLTransformDisplayNode* displayNode = transformNode ? vtkMRMLTransformDisplayNode::SafeDownCast(transformNode->GetDisplayNode()) : nullptr;
+
+  bool wasBlocked = d->ToggleInteractionAction->blockSignals(true);
```
lassoan, line 923: Is there a specific reason for not using QSignalBlocker?

lassoan (approved):
Good catch, thank you!

Modules/Loadable/Transforms/SubjectHierarchyPlugins/qSlicerSubjectHierarchyTransformsPlugin.cxx
Sunderlandkyl: No specific reason, just force of habit. I've updated it to use QSignalBlocker.

lassoan (approved):
Thank you, it looks good.

Issue #8069, opened 2024-11-27
segmentation colors look incorrect, when chosen from terminology 


muratmaga:
This is with today's preview build.

1. Load MRHead and start segmenting.
2. Use threshold to create a segment, and double click to change it is name to Skin from terminology selector
3. Note that there are now two colors. Original segment color (green), and then another color box that shows the color specified in terminology (brown).
4. Create another segment, and choose another term to see that it contiuues to do the same thing. 

The actual displayed segment color is the color on the left (original segment color, not the terminology one)

![image](https://github.com/user-attachments/assets/23131083-3e39-47c5-8e9c-22a7ee594b8d)

@jcfr  @cpinter @lassoan

cpinter:
I tried to reproduce it with r32942 / 861c8b2 and r33123 / 90861e3 (very latest) and the decoration role did not show up in the name column. I tried this on Windows, @muratmaga told me he used Linux, although there should not be a difference.

muratmaga:
This is with r33130 by the way, I think your revisions are older.

muratmaga:
Session start time .......: 20241128_154733
Slicer version ...........: 5.7.0-2024-11-26 (revision 33130 / c5e8d83) linux-amd64 - installed release
Operating system .........: Linux / 6.8.0-49-generic / #49~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Wed Nov  6 17:42:15 UTC 2 / UTF-8 -

cpinter:
OK I install it on Ubuntu and see what happens.

muratmaga:
This was yesterdays (11/27's) build https://download.slicer.org/?offset=-1

muratmaga:
It reproduces with today's linux preview as well.

cpinter:
I double checked the description and this DOES reproduce on windows. My mistake was not double-clicking the name but the color.

My guess is that it is due to this commit, where the `ColorRole` was changed to `Qt::DecorationRole` at some places
https://github.com/Slicer/Slicer/commit/50612e7ca2e5a26858029a93f23005d7e1fc082b#diff-6c0329496f9e1e52921d60a5363b23b64c84e9c6a552996ff838ffa783eaa28cR163
I used the custom `ColorRole` just so that this swatch does not show up in the cell as the decoration role does.

cpinter:
[...]

PR #8071, opened 2024-11-29
Fix segment editor name editing

cpinter:
Hopefully fixes https://github.com/Slicer/Slicer/issues/8069


cpinter:
https://github.com/user-attachments/assets/b71313a9-a449-4c7f-9e1c-7d0b4f315e63

PR #8084, opened 2024-12-10
BUG: Fix ExtensionsIndex build

lassoan:
On Windows, cmake.exe path usually contain space (e.g., "C:/Program Files/CMake/bin/cmake.exe"). If such a string is passed to execute_process as COMMAND directly then the command returns "no such file or directory" error (due to "C:/Program" is not found).

Fixed by converting the semicolon-separated string to a list and using that list as COMMAND.

### DIFF
--- a/Extensions/CMake/SlicerBlockBuildPackageAndUploadExtensions.cmake
+++ b/Extensions/CMake/SlicerBlockBuildPackageAndUploadExtensions.cmake
@@ -146,8 +146,10 @@ foreach(EXTENSION_NAME ${EXTENSION_LIST})
       ${CMAKE_CURRENT_BINARY_DIR}/download_${proj}_wrapper_script.cmake)
     #message(STATUS "Configuring extension download wrapper script: ${download_extension_wrapper_script}")
     file(WRITE ${download_extension_wrapper_script} "
+      # command is a semicolon-separated string that cannot be used as COMMAND directly, because arguments may contain spaces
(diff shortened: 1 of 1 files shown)


lassoan:
I'll proceed with merging this. If it turns out to cause any trouble on the factory machine (unlikely) then we can revert.

Extensions/CMake/SlicerBlockBuildPackageAndUploadExtensions.cmake
```diff
@@ -205,15 +208,18 @@ foreach(EXTENSION_NAME ${EXTENSION_LIST})
   set(build_output_file "${CMAKE_CURRENT_BINARY_DIR}/build_${proj}output.txt")
   set(build_error_file "${CMAKE_CURRENT_BINARY_DIR}/build_${proj}_error.txt")
   #message(STATUS "Configuring extension upload wrapper script: ${build_extension_wrapper_script}")
+  # Note: wrapper_command is a semicolon-separated string that cannot be used as COMMAND directly, because arguments may contain spaces
   file(WRITE ${build_extension_wrapper_script} "
+    set(wrapper_command_list \"${wrapper_command}\")
+    string(REPLACE \";\" \";\" list \${wrapper_command_list})
```
jcfr, line 214: The `list` variable does not seem to be used anywhere

jcfr (changes requested):

jcfr (approved):

PR #8087, opened 2024-12-10
ENH: Move Set/GetSliceClipState methods to vtkMRMLClipNode

Sunderlandkyl:
For backwards compatibility move the slice clipping methods to vtkMRMLClipNode, and mark as deprecated.


lassoan (approved):
This is perfect, thank you.

PR #8088, opened 2024-12-10
DOC: Update dicom.md

pieper:
Add additional suggestions for converting dicom to things Slicer can read (per suggestion from @fedorov)


fedorov:
Yes - dicom2nifti https://github.com/icometrix/dicom2nifti appears to be an actively supported open source python tool maintained by Icometrix. It's been around for quite some time, but also last week I connected with Arne Brys, its lead developer, and I got the impression they are indeed standing behind it.

lassoan (approved):
Looks good to me.

Issue #8110, opened 2024-12-21
3D Slicer build fails on Ubuntu 22.04 due to qt tools path resolution


RafaelPalomar:
## Summary

When building 3D Slicer, the `qt_root_dir` resolution in CMake is incorrect on Ubuntu 22.04 LTS due to the presence of the architecture-specific subdirectory `x86_64-linux-gnu`. This causes issues when building Slicer, as Qt tools like `lconvert` are not found in the expected location.

## Steps to reproduce

1. Attempt to build Slicer on Ubuntu 22.04 LTS with the Qt distributed by the package manager.
2. Observe the following error during the CMake configuration step:

   ```
   CMake Error at CMake/SlicerBlockInstallQtTools.cmake:62 (message):
     Qt tool lconvert not found: /usr/lib/x86_64-linux-gnu/../bin/lconvert
   Call Stack (most recent call first):
     CMake/SlicerCPack.cmake:31 (include)
     CMake/LastConfigureStep/CMakeLists.txt:44 (include)
   ```

3. The error occurs because the `qt_root_dir` is incorrectly resolved as `/usr/lib/../`, leading to an invalid path for `lconvert`. The actual path can be found as follows:

```
find /usr -name lconvert
/usr/bin/lconvert
/usr/lib/qt5/bin/lconvert
/usr/lib/x86_64-linux-gnu/qt5/bin/lconvert
```

### Actual behavior
The `qt_root_dir` resolves to `/usr/lib/../` on Ubuntu 22.04 LTS, which evaluates to `/usr/lib/` instead of `/usr`. As a result, the path to Qt tools such as `lconvert` is incorrect, and the build fails.

### Expected behavior
The `qt_root_dir` should resolve to `/usr` regardless of the presence of architecture-specific subdirectories like `x86_64-linux-gnu`. This would ensure that Qt tools are located correctly, and the build proceeds without errors.

### Suggested Solution
Modify the CMake code to account for architecture-specific subdirectories by adding a check for `x86_64-linux-gnu`. 

## Environment
- Slicer version: N/A (CMake configuration issue)
- Operating system: Ubuntu 22.04 LTS
- System details:
  ```
   lsb_release -a
   No LSB modules are available.
   Distributor ID:	Ubuntu
   Description:	Ubuntu 22.04.5 LTS
   Release:	22.04
   Codename:	jammy
  ```
- Error log:
  ```
  CMake Error at CMake/SlicerBlockInstallQtTools.cmake:62 (message):
[...]

PR #8111, opened 2024-12-21
COMP: Improve system Qt detection on Debian usr-merge and multiarch systems

RafaelPalomar:
This commit fixes Slicer/Slicer#8110 by adding a case for correct detection of `qt_root_dir` for Ubuntu multiarch directory structure.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -716,8 +716,15 @@ set(Slicer_HAVE_QT5 1)
 #
 # See https://issues.slicer.org/view.php?id=3574
 #
+if(NOT DEFINED Slicer_USE_SYSTEM_QT)
   foreach(_path IN ITEMS
       "/usr/lib/"
+      "/usr/lib/${CMAKE_LIBRARY_ARCHITECTURE}/"
+      # The following line is a consequence of multiarch usr merge (https://wiki.debian.org/UsrMerge)
+      # /usr/lib will bye symlinked to /lib. In this case Qt5_DIR could be reported as
+      # /lib/${CMAKE_LIBRARY_ARCHITECTURE}/
+      "/lib/"
+      "/lib/${CMAKE_LIBRARY_ARCHITECTURE}/"
       "/usr/lib32/"
       "/usr/lib64/"
       "/usr/local/lib/"
@@ -730,12 +737,12 @@ set(Slicer_HAVE_QT5 1)
     if("${Qt5_DIR}" MATCHES "^${_path}")
       set(Slicer_USE_SYSTEM_QT ON)
       message(STATUS "")
-      message(STATUS "Forcing Slicer_USE_SYSTEM_QT to ON (Qt5_DIR [${Qt5_DIR}] associated with a system location: ${_path})")
+      message(STATUS "Initializing Slicer_USE_SYSTEM_QT to ON (Qt5_DIR [${Qt5_DIR}] associated with a system location: ${_path})")
       message(STATUS "")
       break()
     endif()
   endforeach()
-
+endif()
 # Always use QVTKOpenGLWidget (instead of QVTKWidget)
 set(Slicer_VTK_USE_QVTKOPENGLWIDGET 1)
 


CMake/SlicerBlockInstallQtTools.cmake
```diff
@@ -1,10 +1,12 @@
...
 get_filename_component(_dir ${_filepath} PATH)
 if(APPLE)
   # "_dir" of the form "<qt_root_dir>/lib/QtCore.framework"
   set(qt_root_dir "${_dir}/../..")
+elseif(${_dir} MATCHES "x86_64-linux-gnu")
+  # "_dir" of the form "<qt_root_dir>/lib/x86_64-linux-gnu"
+  set(qt_root_dir "${_dir}/../../..")
 else()
   # "_dir" of the form "<qt_root_dir>/lib"
   set(qt_root_dir "${_dir}/..")
```
jcfr, line 12:
```suggestion
  set(qt_root_dir "${_dir}/..")
  # "_dir" of the form "<qt_root_dir>/lib/<arch>" (e.g "<qt_root_dir>/lib/x86_64-linux-gnu")
  if("${_dir}" MATCHES "${CMAKE_LIBRARY_ARCHITECTURE}")
    set(qt_root_dir "${qt_root_dir}/..")
  endif()
```

This is a more general approach allowing to support other architecture (e.g arm64). See https://cmake.org/cmake/help/v3.31/variable/CMAKE_LIBRARY_ARCHITECTURE.html#cmake-library-architecture

jcfr (changes requested):

RafaelPalomar:
@jcfr, thanks for your suggestions, they have been integrated in the PR. I tested this in Ubutu 22.04 and it solves Slicer/Slicer#8110

CMake/SlicerBlockInstallQtTools.cmake
```diff
@@ -1,13 +1,19 @@
-
 # Get root directory
 get_property(_filepath TARGET "Qt5::Core" PROPERTY LOCATION_RELEASE)
 get_filename_component(_dir ${_filepath} PATH)
 if(APPLE)
   # "_dir" of the form "<qt_root_dir>/lib/QtCore.framework"
   set(qt_root_dir "${_dir}/../..")
+elseif(${_dir} MATCHES "x86_64-linux-gnu")
+  # "_dir" of the form "<qt_root_dir>/lib/x86_64-linux-gnu"
+  set(qt_root_dir "${_dir}/../..")
```
jcfr, line 9:
```suggestion
```

jcfr (commented):
Thanks for moving forward with this :pray: 

Could you confirm that only relying on

```
  if("${_dir}" MATCHES "${CMAKE_LIBRARY_ARCHITECTURE}")
    set(qt_root_dir "${qt_root_dir}/..")
  endif()
```

is sufficient ?

CMake/SlicerBlockInstallQtTools.cmake
```diff
@@ -1,13 +1,17 @@
...
 get_property(_filepath TARGET "Qt5::Core" PROPERTY LOCATION_RELEASE)
 get_filename_component(_dir ${_filepath} PATH)
 if(APPLE)
   # "_dir" of the form "<qt_root_dir>/lib/QtCore.framework"
   set(qt_root_dir "${_dir}/../..")
 else()
-  # "_dir" of the form "<qt_root_dir>/lib"
-  set(qt_root_dir "${_dir}/..")
+  if("${_dir}" MATCHES "${CMAKE_LIBRARY_ARCHITECTURE}")
+  # "_dir" of the form "<qt_root_dir>/lib/<arch>" (e.g "<qt_root_dir>/lib/x86_64-linux-gnu")
```
jcfr, line 9:
```suggestion
    # "_dir" of the form "<qt_root_dir>/lib/<arch>" (e.g "<qt_root_dir>/lib/x86_64-linux-gnu")
```

jcfr:
while I thinking about the commit message we would write to describe the change, I am realizing that what you we are trying to do is support creating a package in the context of Slicer being built on Qt5 being installed on the system ?

In that particular scenario, I think the "right" approach may be to improve the heuristic here:

https://github.com/Slicer/Slicer/blob/64deee1615824fe821f23bff9b46f3edef37dcd1/CMakeLists.txt#L704-L730

RafaelPalomar:
@jcfr, that's interesting. So in case qt is detected in the system `Slicer_USE_SYSTEM_QT` will be set to `ON`. Can we make the case that the user has QT installed in the system and does not want to use the system qt? I guess `-DQt5_DIR` would do the job for this case.

I'll try playing with the `CMakeLists.txt` and come back to you

jcfr:
Assuming we integrated the patch below, configuring with both `-DQt5_DIR:PATH=/path/to/lib/cmake/Qt5` and explicitly setting `-DSlicer_USE_SYSTEM_QT:BOOL=OFF` would allow to force the packaging of Qt dependencies being installed on the system.

I am curious to see if packaging on Ubuntu 22.04 and then trying to run Slicer on Ubuntu 24.04 without having Qt5 installed on the system would work.

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index c6d54b15ce9..de656b66aab 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -716,6 +716,7 @@ set(Slicer_HAVE_QT5 1)
 #
 # See https://issues.slicer.org/view.php?id=3574
 #
+if(NOT DEFINED Slicer_USE_SYSTEM_QT)
   foreach(_path IN ITEMS
       "/usr/lib/"
       "/usr/lib32/"
@@ -730,11 +731,12 @@ set(Slicer_HAVE_QT5 1)
     if("${Qt5_DIR}" MATCHES "^${_path}")
       set(Slicer_USE_SYSTEM_QT ON)
       message(STATUS "")
-      message(STATUS "Forcing Slicer_USE_SYSTEM_QT to ON (Qt5_DIR [${Qt5_DIR}] associated with a system location: ${_path})")
+      message(STATUS "Initializing Slicer_USE_SYSTEM_QT to ON (Qt5_DIR [${Qt5_DIR}] associated with a system location: ${_path})")
       message(STATUS "")
       break()
     endif()
   endforeach()
+endif()
 
 # Always use QVTKOpenGLWidget (instead of QVTKWidget)
 set(Slicer_VTK_USE_QVTKOPENGLWIDGET 1)
```

RafaelPalomar:
> Assuming we integrated the patch below, configuring with both `-DQt5_DIR:PATH=/path/to/lib/cmake/Qt5` and explicitly setting `-DSlicer_USE_SYSTEM_QT:BOOL=OFF` would allow to force the packaging of Qt dependencies being installed on the system.

I see. This makes sense. There seem to be some implications of the Debian usr-merge (https://wiki.debian.org/UsrMerge). On [f5ab263](https://github.com/Slicer/Slicer/pull/8111/commits/f5ab2633a71e789f40088cfe0e9652169f0bfa68) we expand the list of paths to cover for usr-merge and multiarch variants of paths. This should give good coverage for a wide range of scenarios. Tested on Ubuntu 22.04.

> I am curious to see if packaging on Ubuntu 22.04 and then trying to run Slicer on Ubuntu 24.04 without having Qt5 installed on the system would work.

I'm curious too. I can give this a try.

jcfr (approved):

jcfr:
Thanks for updating this pull request :pray:

I proposed to move forward with `Squash & Merge` specifying the following commit message:

```
COMP: Improve system Qt detection on Debian usr-merge and multiarch systems

On Debian-based distributions that implement usr-merge, `/lib` is a symlink
to `/usr/lib`. As a result, `Qt5_DIR` might end up pointing to a path under
`/lib`, which was not recognized as a system location. Additionally, on
multiarch systems, library directories often include architecture-specific
subfolders under `/usr/lib` or `/lib`.

This commit expands the list of paths checked when initializing `Slicer_USE_SYSTEM_QT`
to include `/lib`, `/lib/<arch>`, and `/usr/lib/<arch>` directories. This
ensures that Slicer correctly detects and uses the system Qt installation
on these configurations.

It also wraps the system Qt detection logic to support overriding the
initialization of `Slicer_USE_SYSTEM_QT`. This allows packaging of Qt
libraries and plugins even if Qt is installed on the system.

```

jcfr:
@RafaelPalomar Note that after trying to configure & package after specifying options as described in https://github.com/Slicer/Slicer/pull/8111#issuecomment-2639982987, we may still have to submit a follow-up pull request to fix `CMake/SlicerBlockInstallQtTools.cmake`, can I let you do so in a follow-up pull request ? 🙏

Issue #7533, opened 2024-01-11
Segmentation visibility settings can be confusing


muratmaga:
## Summary

If a user turns off the visibility of a segmentation from the Data module, and then go to the Segment Editor, and click the Show 3D, no 3D rendering is provided in 3D viewer until they re-enable the visibility from the Data module.  This can be confusing for a enw user to Slicer, since they would have expected to see a 3D rendering (even though slice views not showing any labeled data would have given a clue about visibility). 

 
I often turn on/off the visibility of a segmentation from the Data module, particularly when I am working on multiple segmentations (of different volumes) at the same time. I have to do this since there is actually no object level visibility control neither in the Segment Editor nor in the Segmentations modules for segmentations. 

To alleviate this I would suggest either: 

- adding a visibility button next to do segmentation drop-down menu in Segment Editor and Segmentations (similar to the one in Volume Rendering)
- or implicitly assume that if a user changes the segmentation from the dropdown menus of either module they want to see only that segmentation, and enable visibility for only that automatically and turn off all other segmentation objects. 

I think i would prefer the first one, since it is a behavior that already exists in Volume Rendering. 

## Environment
- Slicer version: Slicer-5.6.1.

muratmaga:
I stand corrected, apparently you can set the object visibility in the Segmentations module, under the Displays. I think it needs to be more prominent and also same option should be exposed in the Segment Editor. Otherwise it is same amount of clicks compare to using Data module for visibility.

lassoan:
There are currently two ways of selecting edited node for a module:
- subject hierarchy tree: Models, Markups
- node combobox: Segmentations, Transforms, Volumes, Volume rendering

Subject hierarchy takes up more space, but makes it easier to select a node (requires one click), allows seeing some more context (e.g., not just the image node name but the study and patient name), and allows quick show/hide or transform of the node.

Could we standardize on using a subject hierarchy tree for node selection in all modules? That would solve the problem described in this issue, as it would show the eye icon right next to the selected node.

PR #8112, opened 2024-12-22
Store and edit terminologies in color table nodes

cpinter:
- Add CSV file format support for color tables
  - Refactoring of storing color properties in color node
  - Save terminology columns into color table CSV file
  - Implement color table terminology CSV reading
- Terminology manual editing from Colors module
- Move Colors widget classes into Colors module
  - The Colors module related widget classes have been in Libs/MRML/Widgets, but that way they cannot possibly use Terminology features. They need to be moved to the Colors module to facilitate this
    - Classes in Testing and DesignerPlugins folders using the moved classes have also been moved to Colors module
  - Remove FileName member from vtkMRMLColorNode, as it was not used (moreover, there were comments explaining why it is not used)
- Create terminology editor widgets for editing from item delegate
- Add functions to color node for get/set terminology as/from string (also add GetColorNameAutoGenerated)
- Style fixes
- Select terminology Colors module from the Terminologies column
- Terminology contexts from color nodes are offered in terminology context selector
- Add new color table node and entry buttons to Colors module
- Add export to color table function to Segment Editor

Re #6975, #7593

@muratmaga

https://github.com/user-attachments/assets/43fc0943-0f9f-4d90-bccc-7d463e90b7bb

https://github.com/user-attachments/assets/40412bd2-e73e-46ca-bf30-fd49f382290f

https://github.com/user-attachments/assets/3c7b3d3f-b9bb-40f4-a78c-c8158345a9e2

PR #8119, opened 2024-12-30
Use subject hierarchy tree as node selector in Volumes, Volume Rendering, Segmentations, Transforms modules

lassoan:
As described in https://github.com/Slicer/Slicer/issues/7533, to make it easier for users to select and show a segmentation (and hide others) the node combobox is now replaced by a subject hierarchy tree. For consistency between all top-level modules, this commit makes the same change for Transforms, Volumes, and Volume Rendering modules.

**Segmentations module:**

image

**Volumes module:**

image

**Volume Rendering module:**

image

**Transforms module:**

image

New transform and segmentation nodes can be created by right-clicking on the empty area in the tree:

image

### DIFF
--- a/Libs/MRML/Widgets/qMRMLClipNodeDisplayWidget.cxx
+++ b/Libs/MRML/Widgets/qMRMLClipNodeDisplayWidget.cxx
@@ -97,7 +97,7 @@ vtkMRMLDisplayNode* qMRMLClipNodeDisplayWidget::mrmlDisplayNode() const
 //------------------------------------------------------------------------------
 void qMRMLClipNodeDisplayWidget::setMRMLDisplayNode(vtkMRMLNode* node)
 {
-  this->setMRMLDisplayNode(vtkMRMLClipNode::SafeDownCast(node));
+  this->setMRMLDisplayNode(vtkMRMLDisplayNode::SafeDownCast(node));
 }
 
 //------------------------------------------------------------------------------
@@ -109,7 +109,6 @@ void qMRMLClipNodeDisplayWidget::setMRMLDisplayNode(vtkMRMLDisplayNode* clipNode
   d->MRMLDisplayNode = clipNode;
   this->updateWidgetFromMRML();
 }
-
 
 //------------------------------------------------------------------------------
 void qMRMLClipNodeDisplayWidget::updateWidgetFromMRML()
--- a/Modules/Loadable/Segmentations/Resources/UI/qSlicerSegmentationsModule.ui
+++ b/Modules/Loadable/Segmentations/Resources/UI/qSlicerSegmentationsModule.ui
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>360</width>
-    <height>909</height>
+    <width>370</width>
+    <height>946</height>
    </rect>
   </property>
   <property name="windowTitle">
@@ -30,43 +30,99 @@
     <number>4</number>
    </property>
    <item>
+    <widget class="ctkExpandableWidget" name="ResizableFrame_2">
+     <property name="orientations">
+      <set>Qt::Vertical</set>
+     </property>
+     <property name="sizeGripInside">
+      <bool>false</bool>
+     </property>
+     <layout class="QVBoxLayout" name="verticalLayout_9">
+      <property name="spacing">
+       <number>4</number>
+      </property>
+      <property name="leftMargin">
+       <number>0</number>
+      </property>
+      <property name="topMargin">
+       <number>0</number>
+      </property>
+      <property name="rightMargin">
+       <number>0</number>
+      </property>
+      <property name="bottomMargin">
+       <number>0</number>
+      </property>
+      <item>
+       <widget class="qMRMLSubjectHierarchyTreeView" name="MRMLNodeSelector_Segmentation">
+        <property name="dragDropMode">
+         <enum>QAbstractItemView::InternalMove</enum>
+        </property>
+        <property name="indentation">
+         <number>12</number>
+        </property>
+        <property name="editMenuActionVisible">
+         <bool>false</bool>
+        </property>
+        <property name="addNodeMenuActionVisible">
+         <bool>true</bool>
+        </property>
+        <property name="multiSelection">
+         <bool>true</bool>
+        </property>
+        <property name="nodeTypes">
+         <stringlist notr="true">
+          <string>vtkMRMLSegmentationNode</string>
+         </stringlist>
+        </property>
+        <property name="hideChildNodeTypes">
+         <stringlist notr="true"/>
+        </property>
+        <property name="idColumnVisible">
+         <bool>false</bool>
+        </property>
+        <property name="colorColumnVisible">
+         <bool>false</bool>
+        </property>
+        <property name="transformColumnVisible">
+         <bool>false</bool>
+        </property>
+        <property name="pluginAllowList">
+         <stringlist notr="true">
+          <string>Segmentations</string>
+          <string>Folder</string>
+          <string>Visibility</string>
+         </stringlist>
+        </property>
+        <property name="excludeItemAttributeNamesFilter">
+         <stringlist notr="true">
+          <string>segmentID</string>
+         </stringlist>
+        </property>
+       </widget>
+      </item>
+     </layout>
+    </widget>
+   </item>
+   <item>
     <layout class="QGridLayout" name="gridLayout_4">
      <property name="spacing">
       <number>4</number>
      </property>
-     <item row="0" column="0">
-      <widget class="QLabel" name="label_ActiveSegmentation">
+     <item row="0" column="1">
+      <widget class="QLabel" name="label_ReferenceVolumeName">
        <property name="text">
-        <string> Segmentation:</string>
+        <string/>
        </property>
       </widget>
      </item>
-     <item row="1" column="0">
+     <item row="0" column="0">
       <widget class="QLabel" name="label_ReferenceVolumeText">
        <property name="toolTip">
         <string>Node that was used for setting the segmentation geometry (origin, spacing, axis directions, and default extent)</string>
        </property>
        <property name="text">
         <string> Source geometry:</string>
-       </property>
-      </widget>
-     </item>
-     <item row="0" column="1">
-      <widget class="qMRMLNodeComboBox" name="MRMLNodeComboBox_Segmentation">
-       <property name="nodeTypes">
-        <stringlist notr="true">
-         <string>vtkMRMLSegmentationNode</string>
-        </stringlist>
-       </property>
-       <property name="renameEnabled">
-        <bool>true</bool>
-       </property>
-      </widget>
-     </item>
-     <item row="1" column="1">
-      <widget class="QLabel" name="label_ReferenceVolumeName">
-       <property name="text">
-        <string/>
        </property>
       </widget>
      </item>
@@ -878,6 +934,11 @@
    <header>qMRMLSegmentationShow3DButton.h</header>
   </customwidget>
   <customwidget>
+   <class>qMRMLSubjectHierarchyTreeView</class>
+   <extends>QTreeView</extends>
+   <header>qMRMLSubjectHierarchyTreeView.h</header>
+  </customwidget>
+  <customwidget>
    <class>qMRMLSubjectHierarchyComboBox</class>
    <extends>ctkComboBox</extends>
    <header>qMRMLSubjectHierarchyComboBox.h</header>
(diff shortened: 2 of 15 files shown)


lassoan:
@pieper @jcfr @jamesobutler @muratmaga I would be interested in any feedback on the general direction where we are going, i.e.:
- In "node modules" (module that allows editing a single node type, such as volumes, segmentations, markups, transforms, tables, etc.): use a tree widget (that shows multiple rows) as main node selector. New nodes can be added by  using the right-click menu.
- In modules where we need to select multiple input and output nodes: use single-line node selectors. Currently, we use qMRMLNodeComboBox, but in the long term this could be replaced by qMRMLSubjectHierarchyCombobox (that shows a tree in the popup instead of a simple list).

jcfr (approved):

jamesobutler (commented):
> New nodes can be added by using the right-click menu.

I think it would be difficult for new users to find this functionality. It could be a convenience feature, but adding buttons that can create a new node would be preferred. 

Reviewing my company’s documentation on UX best practices a selection of 6 or more items is handled by a a dropdown/combobox component. This is what Slicer currently does for the Volumes, Transforms and Volume Rendering modules. I think this makes the most sense as if a user is making a decision to select a single item then once that decision is made there is no need to also display all the other items in the list which would take up space. I see the subject hierarchy only being helpful if the user needs to know information about the parent node/folder organization. Personally I have not utilized these study/patient/series/etc levels of the hierarchy as they are typically created when loading DICOM files. I work in the non-DICOM space where I generally have no hierarchy structure by default when volumes are loaded.

Some guidance items from https://revvity.design/22ce34993/p/067527--dropdown/b/12705f

> Do use when the list of choices is large and/or there is limited space to show all the choices [ISO 9241-161:2016 Section 8.11].

> Include a 'search' field in the dropdown if the list is very expansive, as this can prevent excessive scrolling [ISO 9241-161:2016 Section 8.11].

lassoan:
Thanks for the feedback!

I completely agree that for selection, a single-line combobox is good widget (in Slicer, the popup of the node selector combobox should show a tree instead of just a flat list, but we can discuss that later). In "workflow modules" (that perform some operations on several input and output nodes) this works well.

However, in "node modules" (that manage all instances of a specific node type in the scene), the node selector has to also provide an overview of all the nodes of that type in the scene and allow quick show/hide nodes. A combobox would not work well, because to hide all nodes, you would first need to find which node is shown now (click the combobox + click a node repeatedly until you find the visible node), then click visibility button hide, then click the combobox, click the node that you want to see, and then click visibility button to show. So, when using a combobox, this operation can take up to `(N-1)*2+4` clicks (where N is the number of nodes of that type in the scene). This takes 2 clicks when you have a tree view.

Having a small tree view at the top of each "node module" may not be ideal. Maybe we want to follow a more common pattern of having a "data tree" and show properties for the currently selected item (as in ParaView and many other software). This would be a bigger redesign though - with this pull request I would just want to make one step in the good direction.

jamesobutler:
> Maybe we want to follow a more common pattern of having a "data tree" and show properties for the currently selected item (as in ParaView and many other software).

I should try Paraview out to learn more about their method. 

This discussion reminded me of a recent forum post (https://discourse.slicer.org/t/alternative-to-recommended-way-to-respond-to-volume-loading/40906) about the existing Slicer concept of a selected node. One of the biggest customizations made in my group’s Slicer custom app is the overall selection of volumes to display overlayed or side-by-side in the layout manager. We define a concept of “primary” volume and a “secondary” volume which is displayed in the layout manager. The primary volume selection then applies to all of our utilities such as the Volume module’s display options, the Volume Rendering module’s option and the Segment Editor’s source volume selection. We do not make the user select the volume node manually for all the modules. This is a specific simplified workflow but maybe there is some design that would work to have this be the default but also support more advanced selections in the Slicer application. This was all motivated at reducing number of clicks for users and not allowing them to edit properties of volumes not being shown in the 2D slice viewers which are at the core function of Slicer. This would for example include not turning on Volume Rendering for a volume node not being shown in the 2D slice views.

pieper:
I agree with the comments from @jamesobutler that the proposal seems to take up too much space with items that aren't relevant for the work of the module.  I do like the idea of a 'primary' selection.  One of the most common operations for me is right-clicking and selecting 'edit properties' so I can enter a module with the desired node selected, because the flat-list MRML comboboxes can make it confusing to select the correct node.  Going back and forth to the data module is workable but it is several extra steps.

I also like the idea of being able to control visibility of volumes in the slice views from within the volumes module (and probably models/transforms, etc too from their modules).  Why not have a kind of collapsible widget pops up the subject hierarchy when clicked, closes when a node is selected, and displays the node name when closed?

lassoan:
> seems to take up too much space with items that aren't relevant for the work of the module

This is not happening, because the tree is filtered to show only the nodes the module handles, so there are no irrelevant nodes. Size of the tree view is already adjustable using the draggable splitter below. This dynamic sizing could be made more sophisticated, using some push-pin or expand button, but it is not too bad as it is. Please try it!

I agree that 4 lines instead of 1 is significant in theory, but it does not feel like wasted space, probably because these are already very busy modules and the space has very good use (much easier to see what is in the scene, what is visible, show/hide nodes, and select current node). Overall, seeing more nodes and being able to manipulate them directly in the tree makes the modules easier to use for me. Again, please give it a try.

Of course we could not use a multi-line tree widget as node selector in "workflow modules" (where you select multiple input and output nodes). This PR is only about making node selection consistent across "node modules".

> Going back and forth to the data module is workable but it is several extra steps

Exactly. We are trying to solve this.

Many software have a data tree and when you select an item in the tree then you can edit the item's properties below. This is a "data-oriented" GUI (select data first). In Slicer, the modules implement a "function oriented" GUI (select function first). This PR would kind of bring data-oriented GUI for all "node modules". I don't think the implementation in this PR is the best or final version, but it makes all "node modules" consistent (now some node modules use a combobox, while others use a tree).

I like the "roles" idea and we implement some variant of it in all custom Slicer applications. It would be great if it could be generalized! I'm not sure it is possible, as most general-purpose software uses just a single "selected" role.

pieper:
> so there are no irrelevant nodes

If I'm viewing the properties or adjusting the display of one volume then arguably all the other volumes are irrelevant.

Filtering to just the node type sounds reasonable, but when there's a deep hierarchy (like the default dicom import) then a lot of lines will be used for this, which is essential for picking but takes up space when you aren't.

I agree there's a lot of room for improvement since the current modules are inefficient and inconsistent so it's great to work on this, and I'll try this when I have a chance.  I still think I'll prefer a popup-style of selector with the tree visible (really just making the data module be a popup that acts like a menu would be fine I think).

jamesobutler:
In our Slicer custom application we used comboboxes to select volume nodes for the primary and secondary selection. A button group is used to select to display the volumes in overlay (foreground/background) or side-by-side (1 volume in a given slice viewer). This is how we get around the need of going back to the Data module to turn on/off visibility or drag-drop volumes into viewers. The volume selection comboboxes are always visible as we deem it the most important selection and thing to do in our application as it applies to all the module utilities.

This is a different approach compared to trying to bring the Data module’s usage of the subject hierarchy to many other modules. It creates a little confusion of how to show/hide a volume when there are N number of places which have a subject hierarchy to manipulate it rather than a single place that always looks the same where you go to show/hide volumes. Personally I’d rather have a single place to do an action than have the action repeated in lots of other places. I find it easier to train others and I generally like a consistent experience.

lassoan:
> This is a different approach compared to trying to bring the Data module’s usage of the subject hierarchy to many other modules.

It's not about the Data module. We have been successfully using the node tree in Models and Markups modules for node selection for years. Not having them in the Volume rendering module has been quite painful (especially figuring out which volume is displayed in 3D with volume rendering, not just for me - see #7533), and it was somewhat inconvenient that we did not have a tree in Volumes, Transforms, and Segmentations modules (when working with a complex scene I always had to go to Data module and click "Edit properties" to edit properties of a node).

> In our Slicer custom application we used comboboxes to select volume nodes for the primary and secondary selection. 

Is this generalizable to any use cases? Can you attach a screenshot how this looks/works? (from the description it is not completely clear for me)

Is this for "node modules" or "workflow modules"? The needs of these modules are very different.

For workflow modules, we could keep using a popup, but switch to qMRMLSubjectHierarchyComboBox to see a tree in the popup, not a flat list. We would need to fix up this widget a bit: add rename/remove/add node, highlight rows on hover, make visibility icons work but make selection on single-click, etc. I've quickly exprimented with adding hover highlight and make the visibility icons work (see video below), but probably 1-2 weeks of development would be needed to properly implement everything that is needed.

https://github.com/user-attachments/assets/fc965d64-f906-45dd-a63b-7317a1bcfc57

This PR only about node modules, to bring up all such modules to the same level as Markups and Models modules now. I think I would prefer always seeing multiple lines of the tree (qMRMLSubjectHierarchyTreeView) in the main node selector of a node module, but maybe others would only want to see multiple lines when they click on it (qMRMLSubjectHierarchyComboBox). If this was the case, then it could be possible to develop a combined widget and users somehow could switch between single-line/multi-line modes. It would just require much more work to design and implement such a widget.

muratmaga:
Just coming back from travels. I can comment fresh from running two courses back to back that dropdown selector for volume rendering and volumes modules is one of the most significant stumbling blocks for people new to the Slicer. The moment they have two datasets in the scene, they are almost always confused about which one they are rendering and/or looking at the slice views. Mostly because it is pretty much inconsistent with how other modules (markups, models, 
If the design can be space efficient (from UI perspective), it will be great. But any change from the current setup would be improvement in my experience.

cpinter:
I think this PR is a step in the right direction. Most of these modules are quite busy (and tall) to begin with, so making the most important part of the module a bit more explicit even helps I think. If you use DICOM it is super useful to see which patient/study contains the node you are about to select, and if you don't, then you won't be bothered with those hierarchies (but may still be important if you use folders)

> Why not have a kind of collapsible widget pops up the subject hierarchy when clicked, closes when a node is selected, and displays the node name when closed?
> I still think I'll prefer a popup-style of selector with the tree visible (really just making the data module be a popup that acts like a menu would be fine I think).

If I understand correctly, this is exactly the `qMRMLSubjectHierarchyComboBox` widget. However, I'd prefer the trees, and if they are adjustable in height, then they occupy as much space as you prefer. A reminder that the Models and Markups modules work just like this for years.
And as @muratmaga also says the lack of such clarity (i.e. being able to see the current selection in context) has been an issue. Especially in the Volume Rendering module, not being able to switch between the rendered volume has been a big inconvenience (find currently shown volume, hide, find new one to show, show). Of course it would help if in that module the visibility column showed volume rendering visibility. Which takes me to the next point.

> Primary volume

This issue is part of the really old elephant in the room, which is the concept of shown volumes. As you know every type of displayable object has a single visibility option (separate 2D and 3D as well, but essentially a global visibility) with the option to specify view node IDs in which the object is shown, and if empty the node is shown in all views. Volumes don't work like this, and this is why the visibility column is basically useless in SH. Instead, we have slice composite nodes with background and foreground volumes for 2D, and volume rendering display nodes for 3D, and it is not possible to decide what is "visible" without some heuristics. We haven't been able to solve this problem for more than a decade now, as it is a really complex issue. If we want to make some progress with this I think we'll need a breakout session at PW and/or a recurring topic at the dev meetings about this.
I also like the primary volume concept (I have something like this in all custom apps I have made basically - there is a central parameter node that has reference to "the" volume, which is retrieved by every function that wants to operate on the current volume). But Slicer is different. It is kind of a sandbox application, and there are just too many types of use cases to support to make such simplifications. How would we decide which modules should change selection when the primary volume is changed? In extensions as well? I think this primary volume concept could come together with an overhauled volume visibility mechanism, but it needs a lot of planning.

My vote goes to integrating this PR after the release of 5.8 and keep refining, until we have a good consensus for the overhaul.

lassoan:
OK, I think then we all agree then this PR should be merged and we'll improve node selection in the future (make it collapsible or pinnable).

Is there any objection to include this in 5.8? (this would be a significant usability improvement and if we run into any issues we can fix it in patch releases)

pieper (approved):
I'm okay with it being in 5.8 unless others object.

jamesobutler:
> My vote goes to integrating this PR after the release of 5.8 and keep refining, until we have a good consensus for the overhaul.

^I would generally agree with @cpinter's statement here.

lassoan:
@cpinter probably assumed that 5.8 release has been cut, but still not yet.

@jamesobutler Do you have any specific concern having this in 5.8?

jamesobutler:
Definitely seems like an area to tune and perfect for a preview period to get feedback rather than putting in right before a stable release.

lassoan:
I agree that it is better to put features into stable release that have gone through some testing in the preview release. But it is also better to not add features into patch releases. It is also better to have consistency: use the same kind of node selection in core "node modules"; have similar behavior in Slicer 5.8 and 5.9.

From these tradeoffs, I would choose to risk putting this change into the stable release with minimal testing in preview, because there seems to be a consensus on the overall direction we are going. So, the user feedback we get will help us with just tuning the behavior (and fix any bugs that may be discovered).

jamesobutler:
> But it is also better to not add features into patch releases.

^ Agreed. I would have this come in the next period of Slicer stable 5.10/6.0.

Since this is a feature change I don't think any feature changes should come right before a Slicer stable release without any preview period. This time directly before stable release should be for bug fixes only. 

Since Slicer does not release new stable releases often, if there are issues or changes to be made in this updated experience I would not be confident that a new stable release could be released in a timely manner to address additional feature changes to the experience.

cpinter:
I'm OK with having this in 5.8 as well. I agree getting user feedback is more important than having a small circle consensus. I know it is not cut yet, but was thinking the same as @jamesobutler . Now having mentioned the wide user feedback (and this is a purely usability change) I lean towards including it in the stable.

lassoan:
@jcfr If you don't have a firm day for cutting the Slicer-5.8 release then I think it could make sense to create the 5.8 branch from today's main branch and then if any small fixes come in later those could be cherry-picked into the 5.8 branch. That would allow unblocking the merge of this and similar PRs.

jcfr:
If by Monday January 13th 5pm ET, we don't have the release process initiated, we will move forward with creating a 5.8 branch as suggested by @lassoan.

cpinter:
It is very important to @muratmaga to have a working version of https://github.com/Slicer/Slicer/pull/8112 in 5.8. The functionality is all done, but there are a few small bugs in it that I will try to finish today. I hope we'll suceed with getting this in. Thanks for the consideration.

lassoan:
Slicer-5.8 is created, this can now go into the preview release to get user feedback.

PR #8127, opened 2025-01-08
DOC: Describe responding to change in selected volume

shai-ikko:
This is useful also for responding to volume loading in some cases, so also add a reference to the new section from the existing "Show volume rendering automatically when a volume is loaded".

See https://discourse.slicer.org/t/alternative-to-recommended-way-to-respond-to-volume-loading/40906

### DIFF
--- a/Docs/developer_guide/script_repository/volumes.md
+++ b/Docs/developer_guide/script_repository/volumes.md
@@ -157,6 +157,8 @@ def showVolumeRendering(volumeNode):
 slicer.mrmlScene.AddObserver(slicer.vtkMRMLScene.NodeAddedEvent, onNodeAdded)
 ```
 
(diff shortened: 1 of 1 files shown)


pieper:
Thanks for working on this 👍 

I wonder if we can think of one "best practice" suggestion for detecting that a volume node has been added and that the reading is complete.  That is rather than having two different methods without a clear distinction of when one is needed over the other.

@lassoan @jcfr @sjh26 @jamesobutler does anyone have suggestions for the best event to observe to know that a volume node has been added and that reading is complete?

lassoan:
The IO manager emits the `newFileLoaded` signal, which is fired when loading is completed (even when files are loaded by the DICOM browser). So, a simple way to react to volume loading could be to connect a Python callback to this signal.

PR #8131, opened 2025-01-09
BUG: Fix Model DM crash when adding display node

Thibault-Pelletier:
* Fix vtkMRMLModelDisplayableManager crash if added model display node doesn't derive from vtkMRMLModelDisplayNode


lassoan (approved):
Thank you, good catch! Would you mind checking the method for other unchecked use of the model display node? I quickly scrolled through the code and noticed one more (line 775) but maybe there are more. Thank you!

lassoan:
Thanks for checking it. `IsModelDisplayable` indeed returns false if the model display node is invalid. Color legend display node does not affect model display (it is the responsibility of vtkMRMLColorLegendDisplayableManager), so the check is still correct.

PR #8133, opened 2025-01-10
ENH: Add option to download sample data from custom URL

lassoan:
This commit adds "Custom data download" section to the bottom of Sample Data module. It makes it easier to download data sets when following tutorials: the user does not have to use the web browser, find the downloaded file, or drag the file to the application window, just copy-paste the download URL into the inputbox and hit Enter.

image

@muratmaga you may find this useful for your tutorials

### DIFF
--- a/Modules/Scripted/SampleData/SampleData.py
+++ b/Modules/Scripted/SampleData/SampleData.py
@@ -269,6 +269,27 @@ class SampleDataWidget(ScriptedLoadableModuleWidget):
         if self.developerMode is False:
             self.setCategoryVisible(self.logic.developmentCategoryName, False)
 
+        customFrame = ctk.ctkCollapsibleGroupBox()
+        self.categoryLayout.addWidget(customFrame)
+        customFrame.title = _("Load data from URL")
+        customFrameLayout = qt.QHBoxLayout()
+        customFrame.setLayout(customFrameLayout)
+        self.customSampleLabel = qt.QLabel(_("URL:"))
+        customFrameLayout.addWidget(self.customSampleLabel)
+        self.customSampleName = qt.QLineEdit()
+        customFrameLayout.addWidget(self.customSampleName)
+        self.customDownloadButton = qt.QPushButton(_("Load"))
+        self.customDownloadButton.toolTip = _("Download the dataset from the given URL and import it into the scene")
+        self.customDownloadButton.default = True
+        customFrameLayout.addWidget(self.customDownloadButton)
+        self.showCustomDataFolderButton = qt.QPushButton(_("Show folder"))
+        self.showCustomDataFolderButton.toolTip = _("Show folder where custom data sets are downloaded ({path}).").format(path=slicer.app.cachePath)
+        customFrameLayout.addWidget(self.showCustomDataFolderButton)
+        customFrame.collapsed = True
+        self.customSampleName.connect("returnPressed()", self.onCustomDataDownload)
+        self.customDownloadButton.connect("clicked()", self.onCustomDataDownload)
+        self.showCustomDataFolderButton.connect("clicked()", self.onShowCustomDataFolder)
+
         self.log = qt.QTextEdit()
         self.log.readOnly = True
         self.layout.addWidget(self.log)
@@ -278,6 +299,12 @@ class SampleDataWidget(ScriptedLoadableModuleWidget):
 
     def cleanup(self):
         SampleDataWidget.setCategoriesFromSampleDataSources(self.categoryLayout, {}, self.logic)
+
+    def onCustomDataDownload(self):
+        self.logic.downloadFromURL(self.customSampleName.text)
+
+    def onShowCustomDataFolder(self):
+        qt.QDesktopServices.openUrl(qt.QUrl("file:///" + slicer.app.cachePath, qt.QUrl.TolerantMode))
 
     @staticmethod
     def removeCategories(categoryLayout):


muratmaga:
Is this like the ImportFromURL module from SlicerMorph? ALso what data types are registered?

pieper (approved):
LGTM - maybe iterate with @muratmaga before merging.  To me this look similar to what's in SlicerMorph and better to have it in the core.

lassoan:
> Is this like the ImportFromURL module from SlicerMorph?

Yes, I think so. There may be differences in small details in where the data is downloaded to, is it erased after downloading, how the filename is determined, etc. If you have any specific preference for any of these then we can tune this feature accordingly.

> ALso what data types are registered?

All types of data. The code uses the same heuristics as the "Add data" window. This heuristics previously used only the filename and extension but since sometime in Slicer-5.7, the readers also check the file content, so it is much more reliable.

Modules/Scripted/SampleData/SampleData.py
```diff
@@ -269,6 +269,27 @@ def setup(self):
         if self.developerMode is False:
             self.setCategoryVisible(self.logic.developmentCategoryName, False)
 
+        customFrame = ctk.ctkCollapsibleGroupBox()
+        self.categoryLayout.addWidget(customFrame)
+        customFrame.title = _("Custom data download")
```
jcfr, line 274: For the title, I suggest to use something more specific like "Load data from URL"

```suggestion
        customFrame.title = _("Load data from URL")
```

jcfr: Or even `Add data from URL`. 


This would be consistent with the existing "File -> Add Data" functionality

muratmaga:
This.is great. We have been using this in Slimermorph for a while and it greatly facilitates tutorials by easy access to data.

If you have not considered, i also suggest being able to load mrbs this way.

lassoan:
There was a bug that caused mrb's loaded 3x (mrb loading was falsely detected as failed), but I've fixed it now. It can now load mrb files as well.

Base/QTCore/qSlicerSceneBundleReader.cxx
```diff
@@ -30,6 +30,7 @@
 #include <ctkUtils.h>
 
 // MRML includes
+#include <vtkCollection.h>
```
jcfr, line 33:
```suggestion
```

Add below in the `// VTK includes` section

Base/QTCore/qSlicerSceneBundleReader.cxx
```diff
@@ -90,7 +91,31 @@ bool qSlicerSceneBundleReader::load(const qSlicerIO::IOProperties& properties)
...
+  }
+
   bool success = this->mrmlScene()->ReadFromMRB(file.toUtf8(), clear, this->userMessages());
+
+  // Get all the new nodes
+  QStringList loadedNodeIds;
+  nodes = this->mrmlScene()->GetNodes();
+  for (int n = 0; n < nodes->GetNumberOfItems(); n++)
+  {
+    vtkMRMLNode* node = (vtkMRMLNode*)nodes->GetItemAsObject(n);
```
jcfr, line 110:
```suggestion
  for (int index = 0; n < nodes->GetNumberOfItems(); ++index)
  {
    vtkMRMLNode* node = vtkMRMLNode::SafeDownCast(nodes->GetItemAsObject(index));
```

Base/QTCore/qSlicerSceneBundleReader.cxx
```diff
@@ -90,7 +91,31 @@ bool qSlicerSceneBundleReader::load(const qSlicerIO::IOProperties& properties)
     clear = properties["clear"].toBool();
   }
 
+  // Get all the nodes that have been around before loading
+  QSet<vtkMRMLNode*> nodesPresentBeforeLoading;
+  vtkCollection* nodes = this->mrmlScene()->GetNodes();
+  for (int n = 0; n < nodes->GetNumberOfItems(); n++)
+  {
+    vtkMRMLNode* node = (vtkMRMLNode*)nodes->GetItemAsObject(n);
```
jcfr, line 99:
```suggestion
  for (int index = 0; index < nodes->GetNumberOfItems(); ++index)
  {
    vtkMRMLNode* node = vtkMRMLNode::SafeDownCast(nodes->GetItemAsObject(index));
```

Base/QTCore/qSlicerSceneBundleReader.cxx
lassoan: Good catch, thank you. This was the first time I experienced that AI-generated auto-suggestion injected error into the code (although in this case it was just a style error).

Base/QTCore/qSlicerSceneBundleReader.cxx
lassoan: thank you, fixed

jcfr:
I suggest to squash the following commits together:

image

Ideally, the second set of commit should be submitted into its own pull request, this greatly help with generating release notes.

Title of this independent pull request would be the following:

```
BUG: Ensure MRB loading correctly reports loaded nodes

Previously, the scene reader did not report loaded nodes, leading to
incorrect detection of scene loading failures. This commit ensures
that newly loaded nodes are properly tracked and reported.

Additionally:
- Improved the `slicer.util.loadNode` API to return the first loaded node or the scene if no specific node is reported.
- Refactored `slicer.util.loadScene` to reuse `loadNode`, ensuring consistent behavior.
```

lassoan:
I've moved out the scene loading fix to here: https://github.com/Slicer/Slicer/pull/8135

PR #8134, opened 2025-01-10
BUG: Fix loading and saving .mrb files with trailing space in filename

lassoan:
When a scene mrb file had space before the file extension, for example "myscene .mrb" then on Windows the scene failed to save and load. The issue is due to the folder inside the .mrb file ended with a space ("myscene /"), which kwsys did not handle correctly (due to some Windows file API not supporting it).

This commit fixes saving by removing the trailing space from the internal folder name. It also fixes loading of corrupted scenes created by earlier Slicer versions, by returning with a warning (because the internal folder cannot be deleted) instead of bailing out with an error.


jcfr (approved):
After tweaking the description and commit message 🚀💯

```diff
-due to some Windows file API do not support it
+due to some Windows file API not supporting it
```

lassoan:
Thanks for the review! I've updated the commit message.

PR #8135, opened 2025-01-10
BUG: Ensure MRB loading correctly reports loaded nodes

lassoan:
Previously, the scene reader did not report loaded nodes, leading to
incorrect detection of scene loading failures. This commit ensures
that newly loaded nodes are properly tracked and reported.

Additionally:
- Improved the `SampleDataLogic.loadNode` API to return the first loaded node or the scene if no specific node is reported.
- Refactored `SampleDataLogic.loadScene` to reuse `loadNode`, ensuring consistent behavior.
- Improved `SampleDataLogic.downloadFromSource` to not require node name and file name (it can generate it automatically from the URL)

### DIFF
--- a/Base/QTCore/qSlicerSceneBundleReader.cxx
+++ b/Base/QTCore/qSlicerSceneBundleReader.cxx
@@ -38,6 +38,7 @@
 #include <vtkMRMLApplicationLogic.h>
 
 // VTK includes
+#include <vtkCollection.h>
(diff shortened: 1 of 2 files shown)


jcfr (approved):
After describing the change related to `downloadFromSource` function (in both commit message & PR description), it is good for integration 🙏

lassoan:
Thanks for the review, messages updated.

Issue #264, opened 2020-03-12
TumorGrowth: Slice is not deleted from 3D Viewer


slicerbot:
_This issue was created automatically from an original [Mantis Issue](https://mantisarchive.slicer.org/view.php?id=264). Further discussion may take place here._

PR #8139, opened 2025-01-16
ENH: Adds loadSuccess flag to loadable Qt and VTK DICOM classes

MichaelColonel:
SlicerRT is unable to load DICOM-RT files successfully without it.


MichaelColonel:
SlicerRT issue [#264](https://github.com/SlicerRt/SlicerRT/issues/264).

PR #8141, opened 2025-01-17
ENH: Add JSON format for writing/reading scene and nodes

Punzo:
@lassoan 

### Pull Request: Add JSON Format for MRML scene file and nodes

This is an **early draft version** of the proposed feature. While the functionality is implemented, **performance testing on large MRML files** is still pending, and the **code requires additional cleanup**.

Ideally this would serve as starting point at project week for designing a final solution. [Slicer project week project page.](https://projectweek.na-mic.org/PW42_2025_GranCanaria/Projects/JSONbasedscenefileformat/)

---

#### **Overview**

The primary objective of this feature is to introduce support for **JSON format** to enable writing and reading the full MRML scene file and its nodes. This serves as preparatory work for:

1. Supporting a more structured format for the MRML file with explicit datatypes.
2. Providing APIs in Slicer to facilitate collaborative features, such as retrieving and updating the status of individual nodes.

With this setup, we would start a first step for future adoption and compatibility with standards like OpenUSD and Omniverse. For example, having MRML structured in JSON will make it much easier to convert nodes to OpenUSD.

---

#### **Implemented Features**

- **Node Status Printing**:
  Nodes can now output their state in JSON format using the `WriteMRMLStatusToJSONString` method. For example, in the Python console:

  ```python
  crosshairNode.WriteJSON()
  ```
  Output:
  ```json
  {
    "Crosshair": {
      "id": "vtkMRMLCrosshairNodedefault",
      "name": "Crosshair",
      "hideFromEditors": true,
      "selectable": true,
      "selected": false,
      "singletonTag": "default",
      "crosshairMode": "NoCrosshair",
      "crosshairBehavior": "OffsetJumpSlice",
      "crosshairThickness": "Fine",
      "crosshairRAS": [0.0, 0.0, 0.0]
    }
  }
  ```

- **Node State Reading/Updating**:
  Nodes can now read or update their state (either fully or partially) from a JSON string. For example:

  ```python
  a.ReadJSON('{"Crosshair":{"crosshairRAS":[100.0,100.0,100.0]}}')
  ```

- **Storage nodes, e.g.: Markups**:
```python
  lineMarkupStorageNode.WriteDataToJSONString(lineMarkup)
...
  lineMarkupStorageNode.ReadDataFromJSONString(lineMarkup, jsonString)
  ```

- **Scene-Level JSON Format Support**:
  The format for the entire MRML scene is controlled by the method:

  ```cpp
  vtkMRMLScene::SetUseJSONFormat
  ```

  When enabled, the scene file will be output in JSON format. For example, the attached [sample file](https://github.com/user-attachments/files/18457210/2025-01-17-Scene.zip) demonstrates the current output structure.

---

#### **Underlying Implementation**

The current implementation of the JSON write and read methods builds on the existing XML-based methods. An automatic conversion is performed between the two formats during the process. 

Another approach for avoiding the conversion XML-JSON, could be to change directly the current macros in the Write/ReadXML methods (and rename the methods, maybe?). I am open to feedback and suggestions.

---

#### **Performance**

See 2025 slicer week [project](https://projectweek.na-mic.org/PW42_2025_GranCanaria/Projects/JSONbasedscenefileformat/)

### CODE REVIEW

**Comment by pieper** [comment]
Very nice 👍   this could really demystify mrml (it's currently painful to look at the xmls, but fortunately it's not needed very often).

It would be very nice to have a json schema too.  Looks like there are several tools that can generate schemas from example documents.

**Comment by Punzo** [comment]
> Very nice 👍 this could really demystify mrml (it's currently painful to look at the xmls, but fortunately it's not needed very often).
> 

Indeed (: 

> It would be very nice to have a json schema too. Looks like there are several tools that can generate schemas from example documents.

yes you are right, I still haven't defined a schema. In general, this is still very prelimanry, but the idea was to have a working PR to discuss design, etc..., at the project week.

**Comment by Punzo** [comment]
> I've personally been reluctant to treat MRML as any kind of interchange language since it's quite tied to the Slicer implementation, but there really isn't anything like it, so we should think about perhaps defining a subset of MRML that would be appropriate to use for data interchange, i.e. with USD.

yes currently the idea is that only a limited sets of nodes are used for comunications. MRML -> JSON can help with this. On the long term, we would aim for OpenUSD interoperability (but there are some challanges: e.g. it seems mostly optimized for geometry objects and no official VTK support for a reader/writer)

**Comment by lassoan** [comment]
Among the next steps is implementing node properties similarly to Qt. We could declare property name, type (maybe also default value, range, etc) with macros. These macros could generate code to read, write, copy, print each property (removing a lot of code from MRML data nodes) and the property information could be queried at runtime (which could be useful for automatic GUI generation, recording of node states for history, json schema generation, etc). Since properties could be defined dynamically at runtime, it could be used by Python scripted modules to define new MRML nodes.

**Comment by jcfr** [comment]
Following the discussion on the `JSON-based scene file format` project during the 42nd Project Week[^1], here are some key notes:  

- Supporting partial updates to the scene is an interesting direction.  
- Introducing macros for automatic schema generation would be beneficial.  
- Exploring GraphQL support[^2] could enable batched updates through mutations[^3]. Integration could leverage libraries such as `cppgraphqlgen`[^5], as `libgraphqlparser`[^4] appears unmaintained.
- Investigate VTK serialization[^6] capabilities in recent versions, which might complement this work.

cc: @Punzo @lassoan @sjh26 @diazandr3s 

[^1]: https://projectweek.na-mic.org/PW42_2025_GranCanaria/Projects/JSONbasedscenefileformat/
[^2]: https://graphql.org/learn/
[^3]: https://graphql.org/learn/mutations/
[^4]: https://github.com/graphql/libgraphqlparser
[^5]: https://github.com/microsoft/cppgraphqlgen
[^6]: https://docs.vtk.org/en/latest/advanced/object_manager.html and https://docs.vtk.org/en/latest/release_details/9.4.html#wrapping

**Comment by Punzo** [comment]
@lassoan https://github.com/Slicer/Slicer/pull/8141/commits/daad38441c82fa53271e2cf95294713c93dbcd8a 

renames and moves into MRML core the following classes.

```
vtkMRMLMarkupsJsonElement -> vtkMRMLJsonElement
vtkMRMLMarkupsJsonReader -> vtkMRMLJsonReader
vtkMRMLMarkupsJsonWriter -> vtkMRMLJsonWriter
```

It is really not possible to have this change fully backward-compatible. With this PR the ` vtkMRMLMarkupsJson*` classes are empty child classes of the new ones and the constructor gives a warning the the class is deprecated.

Extensions would need to update as:
https://github.com/chir-set/SlicerExtraMarkups/pull/9

**Comment by lassoan** [comment]
ExtraMarkups extension is the only one that is impacted. If it's a lot of work to keep `vtkMRMLMarkupsJson*` classes around (by making them call `vtkMRMLJson*` classes) then it is OK to specify a precompiler definition like `#define MRML_COMMON_JSON_IO` instead. Then ExtraMarkup extension can use `#ifdef` to decide to use the current `vtkMRMLMarkupsJson*` implementation or use the new `vtkMRMLJson*` classes. We need to send a pull request to ExtraMarkup that implements this so that when this PR is merged then that extension will still compile fine.

**Comment by Punzo** [comment]
> ExtraMarkups extension is the only one that is impacted. If it's a lot of work to keep `vtkMRMLMarkupsJson*` classes around (by making them call `vtkMRMLJson*` classes) then it is OK to specify a precompiler definition like `#define MRML_COMMON_JSON_IO` instead. Then ExtraMarkup extension can use `#ifdef` to decide to use the current `vtkMRMLMarkupsJson*` implementation or use the new `vtkMRMLJson*` classes. We need to send a pull request to ExtraMarkup that implements this so that when this PR is merged then that extension will still compile fine.

I have already done the PR for the extra markups https://github.com/chir-set/SlicerExtraMarkups/pull/9
which simply uses the `vtkMRMLJson*` classes and it can be merged after this one.

I don't think we will need to define the MRML_COMMON_JSON_IO macro. The `vtkMRMLJson*` classes have only additional methods respect to the `vtkMRMLMarkupsJson*` ones and they have simply been moved into the core MRML folder.

The current `vtkMRMLMarkupsJson*` are now child classes of `vtkMRMLJson*`, and they are empty and deprecated (they don't implement different behavior). I kept them only for printing a deprecated warning in the constructor, but we could even remove them.

**Comment by lassoan** [comment]
The extension needs to be compiled with Slicer stable and main branch, unless you confirm with the developer that he will maintain two separate branches.

File: Libs/MRML/Core/vtkMRMLSliceNode.cxx [hunk 1]
```diff
@@ -951,7 +951,17 @@ void vtkMRMLSliceNode::ReadXMLAttributes(const char** atts)
     // Layout color is set in Superclass
     layoutColorFound = true;
   }
+
+
   vtkMRMLReadXMLVectorMacro(fieldOfView, FieldOfView, double, 3);
+  // Adjust the FieldOfView to match the aspect ratio of the slice dimensions.
```
Sunderlandkyl, line 957: Could we move these vtkMRMLSliceNode changes to a separate PR so they could be integrated now?

The changes made in ReadXMLAttributes are also applicable to CopyContent, and would fix slice rendering issues issues when replaying sequences of vtkMRMLSliceNode.

Punzo: sure, it makes sense. I am still working on this PR and It will take a while before being ready for review/merging.

Punzo: @Sunderlandkyl PR is here https://github.com/Slicer/Slicer/pull/8370, if you can test it would be great

Libs/MRML/Core/vtkMRMLAbstractViewNode.cxx
```diff
@@ -333,29 +328,18 @@ bool vtkMRMLAbstractViewNode::SetInteractionNode(vtkMRMLNode* node)
...
-  if (!this->GetAttribute("MappedInLayout"))
-  {
-    return 0;
-  }
-  return strcmp(this->GetAttribute("MappedInLayout"), "1") == 0;
-}
-
 //------------------------------------------------------------------------------
-void vtkMRMLAbstractViewNode::SetMappedInLayout(int value)
+int vtkMRMLAbstractViewNode::IsMappedInLayout()
```
Sunderlandkyl, line 332: Doing some further testing, I think the mapped in layout change is also required for vtkMRMLAbstractViewNode sequences. It would be great if this could also be split.

Punzo: ok I will add it to https://github.com/Slicer/Slicer/pull/8370

Punzo: done in https://github.com/Slicer/Slicer/pull/8370/commits/5d8691e7429a4593c6b99a0394f151747ab7c09f

lassoan:
@Punzo let's start to work towards integrating these changes! A specific reason is that @Sunderlandkyl would like to use the JSON read/write classes for volume rendering presets, but also in general we should just get these merged to avoid risk of merge conflicts later.

To facilitate the merge, could you move these commits to separate pull requests?

1.
- BUG: Do not add WindowLevelPreset if already in the list

2.
- BUG: Fix crash in CollapseBinaryLabelmaps]

3.
BUG: Create temp directory with a timestamp in volume storage node
BUG: Create temp directory with a timestamp in segmentation storage node

The rest of the commits would remain in this pull request.

FYI, @Sunderlandkyl has started to work on splitting the commit `ENH: Add JSON support for MRML scene storage and reading` into two: 1. move the JSON read/write classes to MRLM core; 2. everything else. If you want to make any changes to the files in this commit then please don't force-push but do it in additional commits.

lassoan:
> sure I will do it on Monday.

Perfect, thank you!

Punzo:
@lassoan  PRs created https://github.com/Slicer/Slicer/pull/8448 https://github.com/Slicer/Slicer/pull/8448 https://github.com/Slicer/Slicer/pull/8450

@Sunderlandkyl please let me know if I need to take any other action. Thanks for working on splitting commit https://github.com/Slicer/Slicer/pull/8141/commits/fd3e67924eac4328ec2141ed389d2384f7700826

Punzo:
@lassoan @Sunderlandkyl update:

1) I have rebased and cleaned this PR to latest main. 
2) https://github.com/Slicer/Slicer/pull/8449, I have applied Andras review
3) https://github.com/Slicer/Slicer/pull/8450 is waiting for review
4) https://github.com/Slicer/Slicer/pull/8448 Kyle please let me know if you have time to look into this, thanks in advance!

Punzo:
@chir-set just pinging that the markups json API change has been merged.

Punzo:
Note on https://github.com/Slicer/Slicer/pull/8141/commits/890c6376d84aaf07372bf3dd58e8a13f17be6297:

Saving/Loading nodes async background logic methods can be used as:

```
requestUID = applicationLogic.RequestWriteData(refNode = nodeID, filename = nodeSavePath, backgroundProcess = True)
requestUID = applicationLogic.RequestReadFile(refNode = nodeID, filename = nodeSavePath, displayData = False, deleteFile = False, backgroundProcess = True)
```

the task can be observed to notify the UI as:
```
@vtk.calldata_type(vtk.VTK_LONG)
    def onRequestReadDataDone(caller, eventId, callData):
      if callData != requestUID:
        return
      #notify

requestProcessedEventId = slicer.vtkSlicerApplicationLogic.RequestProcessedEvent
applicationLogic.AddObserver(requestProcessedEventId, onRequestReadDataDone)
```

chir-set:
> @chir-set just pinging that the markups json API change has been merged.

Hello @Punzo ,

I have merged your pull request in the main branch for ExtraMarkups for SlicerPreview. The 5.8 branch of ExtraMarkups does not include your patch since it targets SlicerStable.

Thank you for the PR and for the notification.

lassoan:
@Punzo Let's try to get this in before the new stable release. I see some merge conflicts. Could you fix those? I also see that it is still in draft - is it redy for integration or do you still plan to make more changes? Thanks!

Punzo:
> @Punzo Let's try to get this in before the new stable release. I see some merge conflicts. Could you fix those? I also see that it is still in draft - is it redy for integration or do you still plan to make more changes? Thanks!

yes the PR is ready for review. I can't remove the draft label. I fixed the conflict and updated to latest Slicer. I tested and it looks good.

Libs/MRML/Logic/vtkMRMLSliceLogic.cxx
```diff
@@ -672,27 +672,56 @@ void vtkMRMLSliceLogic::ProcessMRMLLogicsEvents()
...
     double inPoint[4] = { 0, 0, 0, 1 };
-    double outPoint[4];
-    double* outPoint3 = outPoint;
 
     // set the z position to be the active slice (from the lightbox)
     inPoint[2] = this->SliceNode->GetActiveSlice();
 
     vtkPlaneSource* plane = vtkPlaneSource::SafeDownCast(this->SliceModelNode->GetPolyDataConnection()->GetProducer());
 
     int wasModified = this->SliceModelNode->StartModify();
```
jcfr, line 683: Can the change to `vtkMRMLSliceLogic.cxx` be contributed independently ?

Libs/MRML/Core/vtkMRMLLayoutNode.cxx
```diff
@@ -85,6 +85,8 @@ void vtkMRMLLayoutNode::ReadXMLAttributes(const char** atts)
 {
   int disabledModify = this->StartModify();
 
+  this->RemoveAllMaximizedViewNodes();
```
jcfr: Can this fix be contributed independently ?

Libs/MRML/Core/vtkMRMLMarkupsFiducialStorageNode.cxx
```diff
@@ -552,6 +552,7 @@ int vtkMRMLMarkupsFiducialStorageNode::ReadDataInternal(vtkMRMLNode* refNode)
     {
       // clear out the list
       markupsNode->RemoveAllControlPoints();
+      markupsNode->ClearValueForAllMeasurements();
```
jcfr: Similarly, can the change to `vtkMRMLMarkupsNode.cxx` and `vtkMRMLMarkupsFiducialStorageNode.cxx`  be contributed in a separate pull request  ?

Libs/MRML/Core/vtkMRMLNode.cxx
```diff
@@ -395,7 +399,13 @@ void vtkMRMLNode::ReadXMLAttributes(const char** atts)
...
 
   vtkMRMLReadXMLBeginMacro(atts);
-  vtkMRMLReadXMLStringMacro(id, ID);
+  // The node ID should only be set during initial scene loading (when reading a Slicer MRML scene file).
+  // If ReadXMLAttributes is called via ReadJSON for subsequent status updates,
+  // the ID must not be changed to maintain scene consistency.
+  if (this->ID == nullptr)
+  {
+    vtkMRMLReadXMLStringMacro(id, ID);
+  }
```
jcfr, line 408: This may have some unintended side effect. For example, I am wondering if this would impact SceneView.

Interestingly, I am also wondering it this would help address issue discussed in https://github.com/Slicer/Slicer/issues/6287#issuecomment-1096786565

cc: @Sunderlandkyl @lassoan

Punzo: @jcfr thanks for reviewing. 

The main issue is that calling ReadJSON on a node with an ID specified in the JSON that differs from the original will set the new ID on the node. However, the cached NodeIDs in the scene are not updated accordingly, leading to inconsistencies between the node and the scene’s cached information. Additionally, all methods for setting the node ID or updating the cached IDs in the scene are private or protected, so they cannot be called directly to resolve this conflict.

I tested the behavior by removing the fix (the if condition at line 405). If you create scene views and then change the node ID in the JSON and call ReadJSON—either before or after creating a scene view—it breaks the scene view and causes Slicer to crash.

With the fix in place, I tested scene views and everything worked as expected. This suggests that the fix prevents the crash and maintains scene consistency.

Libs/MRML/Core/vtkMRMLMarkupsFiducialStorageNode.cxx
Punzo: ok I will do right now

Libs/MRML/Core/vtkMRMLLayoutNode.cxx
Punzo: ok I will do right now

Libs/MRML/Logic/vtkMRMLSliceLogic.cxx
Punzo: ok I will do right now

Libs/MRML/Core/vtkMRMLNode.cxx
Punzo: of course, I’m open to alternative solutions if you have further insights or suggestions.

Libs/MRML/Logic/vtkMRMLSliceLogic.cxx
Punzo: PR at https://github.com/Slicer/Slicer/pull/8747

Libs/MRML/Core/vtkMRMLMarkupsFiducialStorageNode.cxx
Punzo: PR at https://github.com/Slicer/Slicer/pull/8748

Libs/MRML/Core/vtkMRMLLayoutNode.cxx
Punzo: PR at https://github.com/Slicer/Slicer/pull/8749

Punzo:
@jcfr thanks for reviewing, I have applied/replied to your comments.

New PRs:

- https://github.com/Slicer/Slicer/pull/8748
- https://github.com/Slicer/Slicer/pull/8749
- https://github.com/Slicer/Slicer/pull/8747

this PR also depends on https://github.com/Slicer/Slicer/pull/8450, @lassoan reviewed 8450 and I have applied his review last week. Specifically it is the second commit https://github.com/Slicer/Slicer/pull/8141/commits/8a717adcb5f1a9a3934c4ada5b24e83f60384f09 that needs PR 8450.

I have also improved this PR commit messagges. Jc please let me know if any other action is needed on my side

Punzo:
@jcfr @lassoan I have updated this PR and all the related PRs 

- https://github.com/Slicer/Slicer/pull/8749
- https://github.com/Slicer/Slicer/pull/8748
- https://github.com/Slicer/Slicer/pull/8747
- https://github.com/Slicer/Slicer/pull/8450

please have a look, thanks!

pieper:
Hi @Punzo - I'm glad you are able to work on this, but as I've expressed before my feeling is that this implementation isn't the right approach.  Something simpler could accomplish the goals (and now that I think of it perhaps we can specify the high-level goals a little more explicitly, as in what specific implementation this will support.

To gather arguments for my point of view, I asked Claude for another take.

Here's my prompt:
> Review the PR linked above.  In particular, explain the rationale for such an invasive change to the MRML scene, when the XML descriptions could be unambiguously converted directly to json with a simple single helper function.  Does this PR create a lot of code noise and maintenance burden?  Will extensions tha provide node subclasses now all need to be updated?

And here's what it had to say:

## Review of Slicer PR #8141 — JSON format for MRML scene and nodes

I pulled the PR metadata and diff (27 files, +1818/−225, ~2800 diff lines) and read the key parts. The Copilot review is a neutral file-by-file summary; the substantive critique is below.

### Your premise is correct

Look at the actual base-class implementation in `vtkMRMLNode.cxx` added by the PR:

- `WriteJSON` calls `WriteXML` into a stream, parses that into a `vtkXMLDataElement`, then calls `vtkMRMLJsonWriter::ConvertXMLToJson(...)`.
- `ReadJSON` calls `vtkMRMLJsonReader::ConvertJsonToXML(...)`, parses the resulting XML into a `vtkXMLDataElement`, then walks the attributes and calls `ReadXMLAttributes(...)`.

Both `ConvertXMLToJson` and `ConvertJsonToXML` already exist in `vtkMRMLJsonElement.h`. So for ~95% of nodes the "JSON support" is literally a generic XML round-trip — exactly the single-helper-function approach. It did not need to be expressed as new virtuals on `vtkMRMLNode`. A pair of free functions `WriteNodeAsJson(node)` / `ReadNodeFromJson(node, str)` (and the same for storage nodes) would have delivered the feature with zero base-class surface area and zero header churn.

### Where the genuine code noise lives

1. **A second, parallel scene serializer.** `vtkMRMLScene::Commit` and `LoadIntoScene` are largely duplicated under an `if (this->SetUseJSONFormat) { ... } else { ... }`. The whole user-tags / version / extensions / per-node loop is written twice, in XML and rapidjson dialects. Future scene-format changes must be made in both branches.

2. **Subtle behavior change to XML loading.** `ReadXMLAttributes` is modified to skip `id` when `this->ID != nullptr`. The justification in the comment is "if called via `ReadJSON` for status updates" — but `ReadXMLAttributes` has no way to know who called it, so this also alters XML scene-import behavior in any edge case where a node already has an ID.

3. **Layering inversion.** `vtkMRMLNode.h` adds `friend class vtkSlicerApplicationLogic;`. MRML Core is the bottom layer; Logic sits above it. Friending upward couples them in the wrong direction.

4. **Two unrelated refactors bundled in.** Atomic temp-file/locking helpers (`GenerateTempFilePathForWrite`, `MoveFilesWithLocking`, 250+ lines in `vtkMRMLStorageNode.cxx`) and a background read/write queue on `vtkSlicerApplicationLogic` are independent of JSON support. They should be separate PRs — reviewers currently have to vet three things at once.

5. **Confusing data-member naming.** The flag is exposed as `SetUseJSONFormat`, and the constructor sets `this->SetUseJSONFormat = 0;` — i.e., the member variable shares the setter's name. A future reader will misread that.

### Will extensions need to be updated?

Mostly **no for compilation**, but with caveats:

- `WriteJSON` / `ReadJSON` are `virtual`, not pure-virtual, with working default implementations. Existing `vtkMRMLNode` subclasses in extensions continue to compile and serialize correctly through the XML round-trip — they get JSON "for free."
- `vtkMRMLStorageNode::WriteDataToJSONString` / `ReadDataFromJSONString` default to returning empty/`false`. Extensions providing **custom storage nodes** (e.g., for non-standard data types) will silently produce empty JSON when scene JSON is enabled, with no compile error and no warning at the call site — a functional regression that won't surface until someone tries to save a scene of that type. That's a real maintenance trap.
- Adding virtuals enlarges the vtable, so extension binaries built against older Slicer headers must be **recompiled** (normal ABI requirement for Slicer, but worth noting).
- The `ReadXMLAttributes` id-skip change is a quiet semantic shift that any extension relying on multiple `ReadXMLAttributes` calls per node would silently inherit.

### Bottom line

A single XML↔JSON conversion helper (which the PR already implements internally via `vtkMRMLJsonElement`) is sufficient to deliver the feature for nearly every node, with no base-class API additions and no risk to extensions. The current PR conflates that simple core with (a) a duplicated scene serializer, (b) an unrelated file-locking refactor, and (c) an unrelated background-IO refactor — yielding a large, mixed-purpose change that's harder to review and maintain than the underlying idea requires.

The PR is explicitly marked as an early draft intended as a starting point for the [project-week discussion](https://projectweek.na-mic.org/PW42_2025_GranCanaria/Projects/JSONbasedscenefileformat/), and @Punzo raised the alternative in the description ("change directly the current macros … I am open to feedback"). So the venue exists to push back along these lines.

Punzo:
> Hi @Punzo - I'm glad you are able to work on this, but as I've expressed before my feeling is that this implementation isn't the right approach. Something simpler could accomplish the goals (and now that I think of it perhaps we can specify the high-level goals a little more explicitly, as in what specific implementation this will support.

@pieper thanks for the detailed feedback. I'm absolutely open to design changes.

That said, I'd like to push back on a few points in the Claude analysis, since I don't think the "single XML<->JSON helper" alternative is as clean as it sounds:

- Storage nodes require custom serialization. Storage nodes need per-node customization to produce meaningful JSON (pixel data references, coordinate systems, etc.). Without modifying the node classes themselves, all of that customization would have to live in the helper function, making it just as complex — just in a different file.

- Moving scene serialization to a helper doesn't reduce complexity. The duplicated XML/JSON paths in vtkMRMLScene are a fair observation, but moving that code into a helper just relocates the problem rather than solving it. The serialization logic still needs to exist somewhere.

To clarify the actual scope of this PR, which I realize may not be obvious from the diff, there are really five distinct goals here:

- Serialize nodes, display nodes, and storage nodes to/from JSON via the existing XML <-> JSON conversion.
- Support partial write/read of individual nodes and even partial node state.
- Serialize the full scene to a JSON file (I actually don't need this, but this was explicitly requested during the review process over the past year).
- Load and write segmentation and volume pixel data in the background (with the intent to extend later on to other node types like models).
- Fix a crash caused by concurrent writes to file of segmentation and volume pixel data.

On the bundled commits: the three commits were originally separate PRs, but reviewing and testing them in isolation — especially commits 2 and 3 — was proving very difficult in practice. That's why they ended up together here.

I'm happy to discuss the design further with you and @lassoan. There's no urgency on my end to merge this quickly, but I would avoid changing design again after more than 1 year, unless there is an obvious better solution that can satisfy everyone.

Additional note: Over the past year and a half, we have already incrementally split off and merged several parts of this PR. What remains here is the integration layer that exposes the core serialization functionality already implemented in `vtkMRMLJsonWriter` and `vtkMRMLJsonReader` (see `vtkMRMLJsonElement.cxx`) directly on the node classes, with the intent of making serialization a simple one-liner at the call site. If the consensus is that this integration isn't worth the added API surface, I'm happy to close the PR and maintain the code separately in my own fork, no problem (: !

lassoan:
The long-term goal is to serialize to json and phase out xml. It would allow us to use json data types and schema, which would result in better interoperability. During transition (while we need to support both xml and json) the code will be a bit more complex, but should be temporary (couple of years), and things will be much simpler and cleaner when xml is removed.

We should merge the two other commits separately, but for now it is OK to have them here.

Overall, I don't see this implementation unnecessarily complex. The chatbot has a few small good points, but of course does not get the big picture.

What I miss is that `vtkMRMLStorageNode::WriteDataToJSONString` and `vtkMRMLStorageNode::ReadDataFromJSONString` are not utilized anywhere (or even tested) in Slicer core.

1. We should find a way how they could be utilized in Slicer core. For example we could record scene changes, as it is done in https://github.com/PerkLab/SlicerSandbox/tree/master/SceneRecorder as a macro recorder (and maybe AI analysis).

2. We should have a default implementation that can save any bulk data, if the storage node does not explicitly implement Read/WriteDataFrom/ToJSONString. It is OK if the performance of this default implementation is not very good, therefore it could be implemented by simply using the existing bulk data reading/writing methods to write to files and then encode them (or it could automatically detect if the file is ASCII-only and then it would save without encoding, just with escaping; otherwise it could use Base64 encoding).

pieper:
> The chatbot has a few small good points, but of course does not get the big picture.

I've also been missing the big picture.  It wasn't clear to me that part of your point was to avoid using external storage for things like markups and transforms.  Is that why you are saying storage nodes need special attention?  Why not do that in xml instead and just do the simple xml<->json helper?  I'm sorry, but the title of this PR has always led me to think it was only about xml vs json, but I guess it's really more like "store the bulk data in the text file conveniently".

@Punzo you mentioned that there's logic required to do scene serialization.  I certainly agree we want node-specific logic to stay in the node implementation code and not in some helper function.  My argument would be that it should exist as changes to existing xml methods so that we don't need as many new virtual methods and differing code paths.

Embedding markups and 4x4 transforms in text is good, but still won't help for bulk data since we won't be putting that in json for any realistic use.  The original storage node design was there to allow people to save disk space by referring to existing files.  I'm no sure that requirement still exists since people routinely just save dicom input as nrrd, although it would be nice to be able to refer to things like dicomweb servers or zarr in storage buckets from MRML (or it's json-based successor).

Believe me I don't love xml and will be happy to switch.  We will need an importer for old scenes; we've done that before and it was fine.  But the xml is really just an implementation detail.

The part of the PR to support parallel scene saving is great, and I understand it being bundled for convenience, but I would rather see it in a PR that includes actually using it.  A parallel scene save/restore feature would have a lot of practical benefits and would be easier to review and test on its own.

Punzo:
> What I miss is that `vtkMRMLStorageNode::WriteDataToJSONString` and `vtkMRMLStorageNode::ReadDataFromJSONString` are not utilized anywhere (or even tested) in Slicer core.

you are right those methods are currently not used in the Slicer core. And in general I have added tests only for the background reading/writing (commit 2 and 3). I should add tests for all the new json methods.

> * We should find a way how they could be utilized in Slicer core. For example we could record scene changes, as it is done in https://github.com/PerkLab/SlicerSandbox/tree/master/SceneRecorder as a macro recorder (and maybe AI analysis).
> * We should have a default implementation that can save any bulk data, if the storage node does not explicitly implement Read/WriteDataFrom/ToJSONString. It is OK if the performance of this default implementation is not very good, therefore it could be implemented by simply using the existing bulk data reading/writing methods to write to files and then encode them (or it could automatically detect if the file is ASCII-only and then it would save without encoding, just with escaping; otherwise it could use Base64 encoding).

make sense

> I've also been missing the big picture. It wasn't clear to me that part of your point was to avoid using external storage for things like markups and transforms. Is that why you are saying storage nodes need special attention? Why not do that in xml instead and just do the simple xml<->json helper? I'm sorry, but the title of this PR has always led me to think it was only about xml vs json, but I guess it's really more like "store the bulk data in the text file conveniently".

it is also about using JSON for printing to a string the nodes (e.g. scalar, display, etc...) status. It is easier to "broadcast" the node (for example in the case of partial/delta node status).

> @Punzo you mentioned that there's logic required to do scene serialization. I certainly agree we want node-specific logic to stay in the node implementation code and not in some helper function. My argument would be that it should exist as changes to existing xml methods so that we don't need as many new virtual methods and differing code paths.

ok, I agree that it would be good to limit as much as possible new API methods. What do you propose? we need to mantain both the current XML methods implementation + the json. So I guess, we would need to add like a flag to the readXML/writeXML methods to use JSON. On the other hand, it could be misleading that a methods named "XML", would use JSON. Or maybe you meant something else.

> Embedding markups and 4x4 transforms in text is good, but still won't help for bulk data since we won't be putting that in json for any realistic use. The original storage node design was there to allow people to save disk space by referring to existing files. I'm no sure that requirement still exists since people routinely just save dicom input as nrrd, although it would be nice to be able to refer to things like dicomweb servers or zarr in storage buckets from MRML (or it's json-based successor).

Indeed for the moment I have implemented for the bulk data the JSON methods only for the markups and transform. I like suggestion (2) from Andras in https://github.com/Slicer/Slicer/pull/8141#issuecomment-4453612366, but then probably the methods name should be changed.

> Believe me I don't love xml and will be happy to switch. We will need an importer for old scenes; we've done that before and it was fine. But the xml is really just an implementation detail.

the current PR does support both XML and JSON, but it is set by the flag, probably the reading should be smarter and swtich between the two automatically depending on the input file

> The part of the PR to support parallel scene saving is great, and I understand it being bundled for convenience, but I would rather see it in a PR that includes actually using it. A parallel scene save/restore feature would have a lot of practical benefits and would be easier to review and test on its own.

yes we can split commit 2 and 3 in another PR, no problem for me!

I feel that continue discussing here on github will not be efficient. Please let me know if it something that you would like to discuss at one slicer meeting.

pieper:
@lassoan and I had a chance to discuss in some detail after another meeting was over.  That helped me understand the scope a bit better (supporting some use cases that aren't public).  We agreed to think a bit more about how to move ahead.

Punzo:
@lassoan @pieper here a link for a short document with a new design proposition: https://hackmd.io/@7E7FOOs-R46XBvF5KG2jlw/SJHQFGfezx/edit

please feel free to edit it, then we can discuss in a videocall.

I am converting this PR in draft. We can keep it open since I am still working on it. i,.e. adding some final features to the old design. Then I will start the iterations with the new one after we discuss it.

PR #8143, opened 2025-01-17
Add Support for Configuring Slice View Annotations

### LINKED ISSUE #7184: Refactor Slice Annotation infrasctucture

Since its inception started in 2014, the support for rendering slice annotations has been very useful.

It is ultimately done in the `SliceAnnotations`[^1] class instantiated based on the workflow described below.

### Excerpt copied from the prior comments

_Originally posted by @lassoan in https://github.com/Slicer/Slicer/issues/7176#issuecomment-1684865471_
> Slice view annotations is a mess. It was a quick prototype that got integrated to have this important feature in Slicer, but the design is really fragile and limited. We would need to rebuild it from scratch, using proper MRML-based design and flexibility and extensibility in mind, as overlay on image corners is very important space for displaying information.

### Workflow: Instantiation of `DataProbeLib.SliceAnnotations`

1. `DataProbe` module constructor, `slicer.app.startupCompleted()` is connected to `DataProbe.DataProbe.addView()`[^2]
3. `DataProbe.DataProbe.addView()` instantiates[^3] `DataProbeInfoWidget.DataProbeInfoWidget`[^4] class 
4. In `DataProbeInfoWidget` constructor, the function `DataProbeInfoWidget._createSmall()` which intern instantiates `DataProbeLib.SliceAnnotations()`[^5]

[^1]: https://github.com/Slicer/Slicer/blob/main/Modules/Scripted/DataProbe/DataProbeLib/SliceViewAnnotations.py
[^2]: https://github.com/Slicer/Slicer/blob/8c92a12d6acecde092439981af70b1a27b6d2cf1/Modules/Scripted/DataProbe/DataProbe.py#L40-L41
[^3]: https://github.com/Slicer/Slicer/blob/8c92a12d6acecde092439981af70b1a27b6d2cf1/Modules/Scripted/DataProbe/DataProbe.py#L59-L60
[^4]: https://github.com/Slicer/Slicer/blob/8c92a12d6acecde092439981af70b1a27b6d2cf1/Modules/Scripted/DataProbe/DataProbe.py#L68-L106
[^5]: https://github.com/Slicer/Slicer/blob/8c92a12d6acecde092439981af70b1a27b6d2cf1/Modules/Scripted/DataProbe/DataProbe.py#L401-L406

arhowe00:
## Summary

This Pull Request attempts to address [Refactor Slice View Annotations infrastructure #7184](https://github.com/Slicer/Slicer/issues/7184) by separating the existing annotations infrustructure into components that support flexibilty. The intention is also to rely less on Python wrappers and bring command/observation routines from the the interpreted layer into C++. The functionality is broken down into three main parts:

1. A prescriptive, XML list of tags desired in each corner, currently a `vtkMRMLTextNode`:
  ```XML
  <annotations>
    <corner position="bottom-left">
      <property name="Label" prefix="L: " display-level="always"/>
      <property name="Foreground" prefix="F: " display-level="always"/>
      <property name="Background" prefix="B: " display-level="always"/>
    </corner>
    <edge position="bottom">
      <property name="Label" prefix="L: " display-level="always"/>
    </edge>
  </annotations>
  ```
2. A public API allows plugins to substitute the XML with annotations, given the `<property>` tag's attributes:
  ```C++
  virtual bool
  CanProvideValueForPropertyName(const std::string &propertyName) = 0;
  virtual std::string GetValueForPropertyName(const std::string &propertyName,
                                          const XMLTagAttributes &attributes,
                                          vtkMRMLSliceNode *) = 0;
  virtual std::unordered_set<std::string> GetSupportedProperties() = 0;
  ```
3. Application logic for querying the registered APIs & XML and separate rendering routines, implemented in C++ (see diffs for specifics).

---

## In reference to [#7184](https://github.com/Slicer/Slicer/issues/7184)

This solution accommodates @jcfr's requested solution in the following ways:

- Re-implemented in C++
> The solution is implemented in C++.
- Organize the content to display in `vtkMRMLTextNode` nodes with associated displayable manager
> Each slice node associates with a `vtkMRMLTextNode` where the annotations are prescribed through the simple XML. By default, all slice nodes associate with the same one, as is the current functionality of `SliceViewAnnotations.py`. However, the current form makes it easy for both Layer- and SliceNode specific annotations to be provided
- Allow customization of displayed fields
> Can be handled via the provider & XML
- Support displaying
  - Slice position
  - Anatomical direction codes
  - Image display parameters (window, level, slice thickness, etc)
  - Patient/study/series metadata
> Can be handled by the provider & XML (examples are below of doing this for DICOM data). In sum, a provider knows the specific slice node and any number of XML tag attributes, and the provider can decide what should be displayed specific to a given slice node. Since annotations are rendered on the slice view, we assume that passing the MRML slice node to the provider should be everything that is needed to know what should be displayed.

---

## Commentary

For now, we have not added boolean flags to each corner of `vtkMRMLSliceNode`s for displaying annotations selectively in slice nodes. However, it would be simple to implement this given the modular framework of the refactoring. Each slice node could have boolean flags that are checked by the Displayable Manager to render specific corners on specific slices, something not available with DataProbe's `SliceViewAnnotations.py`.

The [first commit](https://github.com/Slicer/Slicer/commit/e2bbacd44e003ebb64fdc2914f296a0854fdb8b2) contains changes to MRML (CornerText logic and Abstract API), adds a Loadable Module, implements the API in a default case to mimic DataProbe's `SliceViewAnnotations.py`, and removes `SliceViewAnnotations.py` to prevent clashes. To understand why we added anything into MRML core at all, see the Q&A below. The default example uses the API to register a 'Default' annotation provider, which displays the volume names in the bottom-left corner, *but not anything related to DICOM*. You can verify this works by checking out, building, and loading data. For instance:

image

You can also see that by editing the text node directly, within the "Texts" module, you can change the annotations using the prescriptive XML. While we like this approach, we are open to a programmatic one, see the commentary below for more information and discussion. 

image

Noticeably, we did not register a DICOM provider in the first commit. Since this is important, but we were unsure whether this XML + Provider API construct would survive, we made a "high-level" DICOM value provider that intends to exactly emulate the behavior of DataProbe (with some edge-case differences that can also be seen as bug fixes). It is in the [second commit](https://github.com/Slicer/Slicer/commit/e2bbacd44e003ebb64fdc2914f296a0854fdb8b2). As you can see, adding DICOM as a provider shows the extensibility of this approach. The database is queried in Python, and the C++ wraps those calls to provide `vtkMRMLCornerTextLogic` with the values expected in the XML tags. Since the registration process is separate from the rendering routines and logic, the way DICOM tags are provided in `DICOMAnnotationPropertyValueProvider.py` is much more intuitive than in `SliceViewAnnotations.py`. Here is a side-by-side depiction of 3D Slicer 5.6.2 and the feature branch with DICOM loaded in both layers.  

image

Should our proposed API be favored, we will allow tag providers to be registered directly in Python so that the DICOM database calls and caching of results can be implemented in a single Python file.

---

### Q: Why did you add or change anything in MRML Core?

A: To avoid introducing a dependency from the core to a module, the `vtkSlicerCornerTextLogic` class was moved from being in a Loadable module to Libs/MRML/Logic. Some of the dependencies include `vtkMRMLLayoutLogic::UpdateViewCornerAnnotations()` and `vtkMRMLApplicationLogic::GetCornerTextLogic()`. We suggest associating a single `vtkMRMLCornerTextLogic` with the application logic.

---

### Q: Why prescriptive XML?

A: The prescriptive XML, at least currently, allows for high-level and simple modularity available both to developers and users. While a programmatic approach is needed to know what a tag should finally print on the slice view, a user can easily change, e.g., the corner that text is in. We think the XML logic we define is simple enough that all users should be able to manipulate the display of provided tags as desired. However, parsing XML is an inflexibility and could introduce challenges for competing plugins.

---

### Q: Why add an API and not use MRML alone?

A: There is an ongoing discussion of how we can minimize and simplify all exposed functionality. While we think the use case for the API will be isolated enough that developers can efficiently invest their time into implementing the API for their extension, @jcfr and I have been discussing how we can directly address node attributes and make a more MRML-native way to define corner annotation values. However, whatever way we choose, the infrastructure for returning tag values must be implemented somewhere. In the case of this PR, we have decided to do it in a separate API, but we may scrap this idea.

---

### Q: What will be done for modules that want to append annotations as opposed to monolithically overwrite a `vtkMRMLTextNode`?

A: We understand the fragility of the current approach and have thought of a few options. We could make it so that providers also register their own XML (assuming we continue with XML) and then have the `vtkMRMLCornerTextLogic` process all provided XMLs into a final annotation `vtkMRMLTextNode` that is displayed in the current manner.  In this case, We could give each module registering a provider a name (e.g. `Default`, `DICOM` -- perhaps just the name of the module). Annotations could be rendered in the order of a **floating priority** that would a reserved tag attribute (in the XML case). 

If there are clashes, we could additionally add a "rare-case" function that has an API such as setTagOrder(["MyProvider", "Default", "DICOM"]); that will allow a user to impose a second round of ordering based on the descriptive tag order. To minimize confusion, this should be more hidden in the APIs, as this case should be extremely rare. It's pretty clear what the function does and how it works, should the need arise.

---

### Q: What will happen to modules and dependencies that use `SliceViewAnnotations.py`? 

A: There are some codebases we have found that import classes defined in `SliceViewAnnotations.py`. Since this PR will end up deprecating those classes, we should address the ins and outs of those dependencies and whether we should provide backup functionality.


Libs/MRML/Logic/vtkMRMLCornerTextLogic.h
```diff
@@ -0,0 +1,142 @@
...
+// MRML includes
+class vtkMRMLAnnotationNode;
+class vtkMRMLSliceNode;
+class vtkMRMLTextNode;
+class vtkXMLDataElement;
+class vtkMRMLAbstractAnnotationPropertyValueProvider;
+
+// STD includes
+#include <cstdlib>
+#include <array>
```
jcfr, line 40:
```suggestion
#include <array>
#include <cstdlib>
```

Libs/MRML/Logic/vtkMRMLCornerTextLogic.h
```diff
@@ -0,0 +1,142 @@
...
+  static vtkMRMLCornerTextLogic *New();
+  vtkTypeMacro(vtkMRMLCornerTextLogic, vtkMRMLAbstractLogic);
+  void PrintSelf(ostream& os, vtkIndent indent) override;
+
+ /// CornerText Logic functions
+  static vtkMRMLTextNode *GetCornerAnnotations(vtkMRMLScene *,
+                                               const int viewArrangement,
+                                               const std::string& viewName);
+  bool RegisterPropertyValueProvider(const std::string &pluginName,
+                                vtkMRMLAbstractAnnotationPropertyValueProvider *);
```
jcfr, line 69:
```suggestion
                                    vtkMRMLAbstractAnnotationPropertyValueProvider *);
```

Libs/MRML/Logic/vtkMRMLCornerTextLogic.h
```diff
@@ -0,0 +1,142 @@
...
+    EDGE_T,
+    TextLocation_Last
+  };
+
+  /// The Usual vtk class functions
+  static vtkMRMLCornerTextLogic *New();
+  vtkTypeMacro(vtkMRMLCornerTextLogic, vtkMRMLAbstractLogic);
+  void PrintSelf(ostream& os, vtkIndent indent) override;
+
+ /// CornerText Logic functions
```
jcfr, line 64:
```suggestion
  /// CornerText Logic functions
```

Libs/MRML/Logic/vtkMRMLCornerTextLogic.h
```diff
@@ -0,0 +1,142 @@
...
+      {"top", EDGE_T}
+  };
+
+  vtkMRMLCornerTextLogic(const vtkMRMLCornerTextLogic&); // Not implemented
+  void operator=(const vtkMRMLCornerTextLogic&); // Not implemented
+
+  /// Parses the contents of an XMl-based text node into a vtkXMLDataElement.
+  /// The vtkXMLDataElement will be used by other functions to generate the
+  /// actual string intended to be rendered on a given slice view.
+  vtkXMLDataElement* ParseTextNode(vtkMRMLTextNode*);
```
jcfr, line 130: Move to protected

Libs/MRML/Logic/vtkMRMLCornerTextLogic.h
```diff
@@ -0,0 +1,142 @@
...
+
+  bool SliceViewAnnotationsEnabled{true};
+  bool BottomLeftEnabled{true}, TopLeftEnabled{true}, TopRightEnabled{true};
+  int DisplayStrictness, FontSize;
```
jcfr, line 134: Split by having one on each line.

Explicitly initialize `int` as well.

Libs/MRML/Logic/vtkMRMLCornerTextLogic.cxx
```diff
@@ -0,0 +1,358 @@
...
+
+//----------------------------------------------------------------------------
+vtkMRMLCornerTextLogic::vtkMRMLCornerTextLogic()
+{
+}
+
+//----------------------------------------------------------------------------
+vtkMRMLCornerTextLogic::~vtkMRMLCornerTextLogic()
+{
+}
```
jcfr, line 46:
```suggestion
//------------------------------------------------------------------------------
vtkMRMLCornerTextLogic::vtkMRMLCornerTextLogic() = default;

//------------------------------------------------------------------------------
vtkMRMLCornerTextLogic::~vtkMRMLCornerTextLogic() = default;
```

Libs/MRML/Logic/vtkMRMLCornerTextLogic.cxx
```diff
@@ -0,0 +1,358 @@
...
+
+//----------------------------------------------------------------------------
+void vtkMRMLCornerTextLogic::PrintSelf(ostream& os, vtkIndent indent)
+{
```
jcfr, line 50: Add display of current properties

Libs/MRML/Logic/vtkMRMLCornerTextLogic.cxx
```diff
@@ -0,0 +1,358 @@
...
+  if (!sliceNode || !textNode)
+  {
+    if (printWarnings)
+    {
+      vtkErrorWithObjectMacro(sliceNode, "Invalid input nodes.");
+    }
+    return cornerAnnotations;
+  }
+
+  if (textNode->GetText().size() <= 0) return cornerAnnotations;
```
jcfr, line 183:
```suggestion
  if (textNode->GetText().size() <= 0)
  {
    return cornerAnnotations;
  }
```

Libs/MRML/Logic/vtkMRMLCornerTextLogic.cxx
```diff
@@ -0,0 +1,358 @@
...
+      {
+        if (provider->CanProvideValueForPropertyName(propertyName))
+        {
+          propertyValue = provider->GetValueForPropertyName(
+              propertyName, attributes, sliceNode);
+        }
+      }
+
+      const bool valueNotProvided = propertyName != "" && propertyValue == "";
+
```
jcfr, line 327:
```suggestion
```

Libs/MRML/Logic/vtkMRMLCornerTextLogic.cxx
```diff
@@ -0,0 +1,358 @@
...
+            this->DisplayStrictness)
+        {
+          text += propertyValue + '\n';
+        }
+      }
+    }
+
+    // once each property has been parsed, the annotation for that position
+    // is fully specified
+
```
jcfr, line 353:
```suggestion
```

Libs/MRML/Logic/vtkMRMLCornerTextLogic.cxx
```diff
@@ -0,0 +1,358 @@
...
+          vtkErrorWithObjectMacro(textNode,
+                                  "<" + std::string(cornerOrEdge->GetName()) +
+                                      " position=" + position +
+                                      "> has a property with a missing name.");
+        }
+        break;
+      }
+
+      // Get the attribute values out of the XML tag
+
```
jcfr, line 308:
```suggestion
```

Libs/MRML/Logic/vtkMRMLCornerTextLogic.cxx
```diff
@@ -0,0 +1,358 @@
...
+        }
+        break;
+      }
+
+      std::string propertyName = "", propertyValue = "";
+      vtkMRMLAbstractAnnotationPropertyValueProvider::XMLTagAttributes
+          attributes;
+
+      // Get the value for "name" in the tag <property "name"=...> (required)
+
```
jcfr, line 290:
```suggestion
```

Libs/MRML/Logic/vtkMRMLCornerTextLogic.cxx
```diff
@@ -0,0 +1,358 @@
...
+                        "self-closing <property /> tags.");
+        }
+        break;
+      }
+
+      std::string propertyName = "", propertyValue = "";
+      vtkMRMLAbstractAnnotationPropertyValueProvider::XMLTagAttributes
+          attributes;
+
+      // Get the value for "name" in the tag <property "name"=...> (required)
```
jcfr, line 289:
```suggestion
      std::string propertyName = "";
      std::string propertyValue = "";
      vtkMRMLAbstractAnnotationPropertyValueProvider::XMLTagAttributes attributes;

      // Get the value for "name" in the tag <property "name"=...> (required)
```

Libs/MRML/Logic/vtkMRMLCornerTextLogic.cxx
```diff
@@ -0,0 +1,358 @@
...
+            textNode,
+            "<corner> and <edge> tags must specify a position attribute (e.g. "
+            "<corner position=\"bottom-left\"> or <edge "
+            "position=\"bottom\">).");
+      }
+      return cornerAnnotations;
+    }
+
+    std::string position = cornerOrEdge->GetAttribute("position");
+
```
jcfr, line 239:
```suggestion
```

Libs/MRML/Logic/vtkMRMLCornerTextLogic.cxx
```diff
@@ -0,0 +1,358 @@
...
+  }
+
+  // Parse <corner>/<edge> elements within <annotations>
+
+  if (annotations->GetNumberOfNestedElements() == 0)
+  {
+    if (printWarnings)
+    {
+      vtkErrorWithObjectMacro(textNode,
+                              "<annotations> tag had no nested elements.");
```
jcfr, line 204:
```suggestion
      vtkErrorWithObjectMacro(textNode, "<annotations> tag had no nested elements.");
```

Modules/Loadable/CornerText/qSlicerCornerTextDICOMAnnotationPropertyValueProvider.cxx
```diff
@@ -0,0 +1,157 @@
...
+  limitations under the License.
+
+==============================================================================*/
+
+#include <vtkMRMLSliceViewDisplayableManagerFactory.h>
+
+// Slicer includes
+#include <vtkSlicerApplicationLogic.h>
+#include <qSlicerApplication.h>
+#include <qSlicerPythonManager.h>
```
jcfr, line 21:
```suggestion
#include <qSlicerApplication.h>
#include <qSlicerPythonManager.h>
#include <vtkSlicerApplicationLogic.h>
```

Make sure to sort includes alphabetically

Modules/Loadable/CornerText/qSlicerCornerTextDICOMAnnotationPropertyValueProvider.cxx
```diff
@@ -0,0 +1,157 @@
...
+
+// DisplayableManager initialization
+#include <vtkAutoInit.h>
+
+// Qt includes
+#include <QSettings>
+#include <QObject>
+#include <QDebug>
+#include <QJsonObject>
+#include <QJsonDocument>
```
jcfr, line 34:
```suggestion
#include <QDebug>
#include <QJsonObject>
#include <QJsonDocument>
#include <QObject>
#include <QSettings>
```

Modules/Loadable/CornerText/qSlicerCornerTextDICOMAnnotationPropertyValueProvider.cxx
```diff
@@ -0,0 +1,157 @@
...
+// Slicer includes
+#include <vtkSlicerApplicationLogic.h>
+#include <qSlicerApplication.h>
+#include <qSlicerPythonManager.h>
+
+// CornerText includes
+#include "qSlicerCornerTextDICOMAnnotationPropertyValueProvider.h"
+
+// DisplayableManager initialization
+#include <vtkAutoInit.h>
```
jcfr, line 27:
```suggestion
```

This should only be includes from  `qSlicerCornerTextModule.cxx` where I would also expect the macro `VTK_MODULE_INIT` to also be used.

Modules/Loadable/CornerText/qSlicerCornerTextModule.cxx
```diff
@@ -0,0 +1,228 @@
...
+qSlicerCornerTextModule::qSlicerCornerTextModule(QObject* _parent)
+  : Superclass(_parent)
+  , d_ptr(new qSlicerCornerTextModulePrivate)
+{
+}
+
+//-----------------------------------------------------------------------------
+qSlicerCornerTextModule::~qSlicerCornerTextModule()
+{
+  writeSettings();
```
jcfr, line 83:
```suggestion
  this->writeSettings();
```

Modules/Loadable/CornerText/qSlicerCornerTextModule.cxx
```diff
@@ -0,0 +1,228 @@
...
+QString qSlicerCornerTextModule::acknowledgementText() const
+{
+  return "This work was partially funded by NIH grant NXNNXXNNNNNN-NNXN";
+}
+
+//-----------------------------------------------------------------------------
+QStringList qSlicerCornerTextModule::contributors() const
+{
+  QStringList moduleContributors;
+  moduleContributors << QString("John Doe (AnyWare Corp.)");
```
jcfr, line 102:
```suggestion
  moduleContributors << QString("Andrew Howe (Kitware)");
  moduleContributors << QString("Jean-Christophe Fillion-Robin (Kitware)");
```

Modules/Loadable/CornerText/qSlicerCornerTextModule.cxx
```diff
@@ -0,0 +1,228 @@
...
+//-----------------------------------------------------------------------------
+qSlicerCornerTextModule::~qSlicerCornerTextModule()
+{
+  writeSettings();
+}
+
+//-----------------------------------------------------------------------------
+QString qSlicerCornerTextModule::helpText() const
+{
+  return "This is a loadable module that can be bundled in an extension";
```
jcfr, line 89: Update this

Modules/Loadable/CornerText/qSlicerCornerTextModule.cxx
```diff
@@ -0,0 +1,228 @@
...
+  settings->setValue("DataProbe/sliceViewAnnotations.fontFamily", QString::fromStdString(cornerTextLogic->GetFontFamily()));
+
+  settings->setValue("DataProbe/sliceViewAnnotations.bottomLeft", cornerTextLogic->GetBottomLeftEnabled());
+  settings->setValue("DataProbe/sliceViewAnnotations.topLeft", cornerTextLogic->GetTopLeftEnabled());
+  settings->setValue("DataProbe/sliceViewAnnotations.topRight", cornerTextLogic->GetTopRightEnabled());
+}
+
+//-----------------------------------------------------------------------------
+void qSlicerCornerTextModule::onLayoutChanged(int layout)
+{
```
jcfr, line 165:
```suggestion
{
  qSlicerLayoutManager* layoutManager = qSlicerApplication::application()->layoutManager();
```

Modules/Loadable/CornerText/qSlicerCornerTextModule.cxx
```diff
@@ -0,0 +1,228 @@
...
+            ->overlayCornerAnnotation();
+
+    vtkMRMLAbstractDisplayableManager* displayableManager =
+        qSlicerApplication::application()
+            ->layoutManager()
+            ->sliceWidget(sliceViewName)
+            ->sliceView()
+            ->displayableManagerByClassName("vtkMRMLCornerTextDisplayableManager");
+
+    vtkMRMLCornerTextDisplayableManager* cornerTextDisplayableManager;
```
jcfr, line 181: Display error if downcast fails

Modules/Loadable/CornerText/qSlicerCornerTextModule.cxx
```diff
@@ -0,0 +1,228 @@
...
+  this->Superclass::setup();
+
+  // Register displayable managers
+  vtkMRMLSliceViewDisplayableManagerFactory::GetInstance()->RegisterDisplayableManager("vtkMRMLCornerTextDisplayableManager");
+
+  // Register default annotation provider
+  vtkNew<vtkMRMLDefaultAnnotationPropertyValueProvider> defultProvider;
+  defultProvider->SetAppLogic(this->appLogic());
+  this->appLogic()->GetCornerTextLogic()->RegisterPropertyValueProvider(
+      "Default", defultProvider);
```
jcfr, line 203:
```suggestion
  this->appLogic()->GetCornerTextLogic()->RegisterPropertyValueProvider("Default", defultProvider);
```

Modules/Loadable/CornerText/qSlicerCornerTextModule.cxx
```diff
@@ -0,0 +1,228 @@
...
+  {
+    QObject::connect(qSlicerApplication::application()->layoutManager(), SIGNAL(layoutChanged(int)),
+      this, SLOT(onLayoutChanged(int)));
+  }
+
+  // Register the DICOM property value provider
+#ifdef Slicer_USE_PYTHONQT
+  vtkNew<qSlicerCornerTextDICOMAnnotationPropertyValueProvider> DICOMProvider;
+  this->appLogic()->GetCornerTextLogic()->RegisterPropertyValueProvider(
+      "DICOM", DICOMProvider);
```
jcfr, line 219:
```suggestion
  this->appLogic()->GetCornerTextLogic()->RegisterPropertyValueProvider("DICOM", DICOMProvider);
```

Modules/Loadable/CornerText/qSlicerCornerTextModule.cxx
```diff
@@ -0,0 +1,228 @@
...
+
+  // Observe layout manager (may not be available if running without main window)
+  if (qSlicerApplication::application()->layoutManager())
+  {
+    QObject::connect(qSlicerApplication::application()->layoutManager(), SIGNAL(layoutChanged(int)),
+      this, SLOT(onLayoutChanged(int)));
+  }
+
+  // Register the DICOM property value provider
+#ifdef Slicer_USE_PYTHONQT
```
jcfr, line 216:
```suggestion
#ifdef Slicer_BUILD_DICOM_SUPPORT
```

Modules/Loadable/CornerText/Widgets/qSlicerCornerTextSettingsWidget.h
```diff
@@ -0,0 +1,76 @@
...
+#include <QWidget>
+
+// Settings Widgets includes
+#include "ctkPimpl.h"
+#include "ctkVTKObject.h"
+#include "qMRMLWidget.h"
+#include "qSlicerCornerTextModuleWidgetsExport.h"
+#include "vtkMRMLScene.h"
+#include "vtkMRMLLayoutLogic.h"
+#include "vtkMRMLCornerTextLogic.h"
```
jcfr, line 31: split as `// MRML includes`, `// CTK includes`, ...

Libs/MRML/Core/vtkMRMLSliceNode.h
```diff
@@ -565,6 +566,15 @@ class VTK_MRML_EXPORT vtkMRMLSliceNode : public vtkMRMLAbstractViewNode
 
   virtual vtkImplicitFunction* GetImplicitFunctionWorld();
 
+  ///@{
+  /// Node reference role for text that is used in the slice view
+  const char* GetCornerAnnotationsTextNodeReferenceRole() { return "cornerText"; }
+  const char* GetCornerAnnotationsTextNodeReferenceMRMLAttributeName() { return "cornerTextRef"; }
+
```
jcfr, line 573:
```suggestion
  ///}@

  ///@{
  /// Set/Get text node used to configure the slice view corner annotations
```

Libs/MRML/Core/vtkMRMLSliceNode.h
```diff
@@ -18,6 +18,7 @@
 // MRML includes
 #include "vtkMRMLAbstractViewNode.h"
 class vtkMRMLVolumeNode;
+class vtkMRMLTextNode;
```
jcfr, line 21:
```suggestion
class vtkMRMLTextNode;
class vtkMRMLVolumeNode;
```

Modules/Scripted/DataProbe/DataProbeLib/DICOMAnnotationPropertyValueProvider.py
```diff
@@ -0,0 +1,221 @@
...
+                return LAYER_BACKGROUND
+            elif layer == "label" or get_digit(layer) == LAYER_LABEL:
+                return LAYER_LABEL
+        return LAYER_ANY
+
+    # A part of the default annotation provider, not abstract. We need a way to
+    # get app logic
+    @staticmethod
+    def GetAppLogic():
+        return slicer.app.applicationLogic()
```
jcfr, line 30:
```suggestion
```

Modules/Scripted/DataProbe/DataProbeLib/DICOMAnnotationPropertyValueProvider.py
```diff
@@ -0,0 +1,221 @@
...
+
+        if isinstance(sliceNode, str):
+            sliceNode = slicer.mrmlScene.GetNodeByID(sliceNode)
+
+        # Do not attempt to retrieve dicom values if no local database exists
+        if not slicer.dicomDatabase.isOpen:
+            return ""
+
+        # Get the layers
+        sliceLogic = DICOMAnnotationPropertyValueProvider.GetAppLogic().GetSliceLogic(sliceNode)
```
jcfr, line 48:
```suggestion
        sliceLogic = slicer.app.applicationLogic().GetSliceLogic(sliceNode)
```

Modules/Scripted/DataProbe/DataProbe.py
```diff
@@ -403,7 +402,6 @@ def _createSmall(self):
         """
 
         # this method makes SliceView Annotation
```
pieper, line 404: Delete stray comment

Libs/MRML/Logic/vtkMRMLCornerTextLogic.h
arhowe00: What did you want me to move to protected here?

arhowe00: Notes from JC/Andrew 2025.1.20.1157: vtk styling suggests that methods such as `ParseTextNode` 0are suggested to be protected so derived classes can access them.

Libs/MRML/Core/vtkMRMLAbstractAnnotationPropertyValueProvider.h
```diff
@@ -0,0 +1,143 @@
+/*==============================================================================
+
+  Program: 3D Slicer
+
+  Copyright(c) Kitware Inc.
+
+  See COPYRIGHT.txt
+  or http://www.slicer.org/copyright/copyright.txt for details.
+
+  Unless required by applicable law or agreed to in writing, software
```
jcfr, line 10: Test

Libs/MRML/Core/vtkMRMLAbstractAnnotationPropertyValueProvider.h
```diff
@@ -0,0 +1,143 @@
...
+class vtkMRMLSliceNode;
+
+// VTK includes
+#include <vtkObject.h>
+
+// STD includes
+#include <algorithm>
+#include <cctype>
+#include <string>
+#include <unordered_set>
```
jcfr, line 31: Comment on 4 lines

Libs/MRML/Core/vtkMRMLSliceNode.h
arhowe00: Hey, while I make these edits, should I always only declare the class if I never touch the implementation? For instance, if I did not access any members of `vtkMRMLAbstractViewNode` in both the .h and .cxx of `vtkMRMLSliceNode`, should it be

```
class vtkMRMLAbstractViewNode;
```
?

Modules/Loadable/CornerText/Widgets/qSlicerCornerTextSettingsWidget.h
arhowe00: I did this split, but the organization doesn't always seem completely clear to me

arhowe00 (commented):
I rebased all these changes into the previous two commits and force pushed.

jcfr:
Following discussions with @lassoan during the [42nd Slicer Project Week](https://projectweek.na-mic.org/PW42_2025_GranCanaria/), we agreed to proceed with the integration once the following points are addressed:

- **Prevent saving the text node in the scene:** While allowing users to update the text interactively is convenient, we anticipate potential changes to the XML attributes or format before the next Slicer release. To avoid serializing an "unstable" structure in the scene, we prefer to disable saving of the text node for now. For reference, [here](https://github.com/Slicer/Slicer/blob/6a7deea0e133ff7d81bfaa9e770b29899fa08848/Libs/MRML/Logic/vtkMRMLApplicationLogic.cxx#L451-L486) is the current XML structure.

- **Update `display-level` to use integers instead of descriptive labels:** The current values (`always`, `sometimes`, `least`) limit extensibility. Switching to an integer-based approach will facilitate the integration of additional levels in the future.

- **Create an issue outlining follow-up work**: We should document the planned improvements for a future pull request, ensuring a clear roadmap for enhancing the annotation system. This should include further refinements to the API, support for additional customization options, and considerations for integrating annotation updates more seamlessly into MRML.

----------

While discussing with @lassoan, we also considered changing the `layer` attribute to `role`, which would provide greater flexibility—for example, allowing annotations to reference other text nodes based on different roles.

@lassoan, do you have any comments or suggestions regarding this change?

### Current Structure:

```xml
<annotations>
  <!-- For corners -->
  <corner position="bottom-left">
    <!-- Generic properties -->
    <property name="VolumeName" layer="label" prefix="L: " display-level="always"/>
    <property name="VolumeName" layer="foreground" prefix="F: " display-level="always"/>
    <property name="VolumeName" layer="background" prefix="B: " display-level="always"/>
  </corner>
  <corner position="top-left">
    <!-- DICOM-specific properties -->
    <property name="PatientName" display-level="sometimes"/>
    <property name="PatientID" prefix="ID: " display-level="always"/>
    <property name="PatientInfo" display-level="sometimes"/>
    <property name="SeriesDate" prefix="B: " layer="background" display-level="sometimes"/>
    <property name="SeriesDate" prefix="F: " layer="foreground" display-level="sometimes"/>
    <property name="SeriesTime" prefix="B: " layer="background" display-level="least"/>
    <property name="SeriesTime" prefix="F: " layer="foreground" display-level="least"/>
    <property name="SeriesDescription" prefix="B: " layer="background" display-level="least"/>
    <property name="SeriesDescription" prefix="F: " layer="foreground" display-level="least"/>
  </corner>
  [...]
</annotations>
```

### Proposed Generalization Using `role`:

```xml
<annotations>
  <!-- For corners -->
  <corner position="bottom-left">
    <!-- Generic properties -->
    <property name="VolumeName" role="label" prefix="L: " display-level="always"/>
    <property name="VolumeName" role="foreground" prefix="F: " display-level="always"/>
    <property name="VolumeName" role="background" prefix="B: " display-level="always"/>
  </corner>
  <corner position="top-left">
    <!-- DICOM-specific properties -->
    <property name="PatientName" display-level="sometimes"/>
    <property name="PatientID" prefix="ID: " display-level="always"/>
    <property name="PatientInfo" display-level="sometimes"/>
    <property name="SeriesDate" prefix="B: " role="background" display-level="sometimes"/>
    <property name="SeriesDate" prefix="F: " role="foreground" display-level="sometimes"/>
    <property name="SeriesTime" prefix="B: " role="background" display-level="least"/>
    <property name="SeriesTime" prefix="F: " role="foreground" display-level="least"/>
    <property name="SeriesDescription" prefix="B: " role="background" display-level="least"/>
    <property name="SeriesDescription" prefix="F: " role="foreground" display-level="least"/>
  </corner>
  [...]
</annotations>
```

arhowe00:
Thanks @jcfr, I implemented these changes into another commit which I've pushed.

## Summary

The existing refactoring will allow any module to dynamically modify annotations in a slice view. The implementation could be more lean with a few changes, but there are tradeoffs in those changes:

1. MRML attributes could be directly added and referenced when determining what is displayed in a corner, which would reduce the processing that has to occur on part of an annotation provider.  This could potentially constrain an implementor that wants a group of computations to decide what is displayed on the screen.
2. Regarding the XML, it may provide a good middle ground between "our-module-specific-jargon" and a readable way for users to modify the displayed text.  However, removing the XML would make the API less reliant on a custom, albeit lightweight, syntax, allowing complete programmatic display of strings on the slice view.  Such a programmatic view might be more modular, where clashing modules which all want to contribute their own annotations to the slice view could have ranking or floating values associated with the annotations, with annotations displayed in the order of their rank.  Then, annotations from anywhere can be composed and rendered, allowing a more complete answer to all possible needs of users (clinicians, researchers, industry users).

Furthermore, there are a few more things of note:

1. We deferred the decision on removing the "Background DICOM annotations persistence" button.
2. We should keep in mind dependencies who use the class `SliceAnnotations`.  In addition to providing the new flexibility, registration APIs, and refactoring into C++, this PR also implements the annotations features provided by `SliceViewAnnotations.py`, but is not a drop-in-substitute for dependencies of the SliceAnnotations Python class.
3. I like the idea of switching "layer" to "role", but note that both the current implementors ("DICOM" and "Default") use the concept of "role" to mean layer. As you can see, for instance, both providers still use the functions `vtkMRMLSliceLogic::GetForegroundLayer()`, as the "role" which `vtkMRMLDefaultAnnotationPropertyValueProvider.cxx` is interested in.  Other implementors can do whatever they want with "role", but they could also have just made a custom property and parsed it themselves in their provider.
4. Note how the new behavior of "Display Level" is that a lower number will show less information on the screen.  I have tried to make the implementation of "Display Level" as intuitively as I possibly could in the code.
5. We could give text nodes specific to the annotations module a custom property and then select and edit them from within the CornerText module -- this is a WIP.

jcfr:
@arhowe00  Could you have a look at the following failure [reported](https://github.com/Slicer/Slicer/actions/runs/13655531787/job/38173691006) in the CI ?

```
/usr/src/Slicer/Libs/MRML/Logic/vtkMRMLCornerTextLogic.h:118:14: error: ‘unordered_map’ in namespace ‘std’ does not name a template type
   const std::unordered_map<std::string, TextLocation> positionMap =
              ^~~~~~~~~~~~~
/usr/src/Slicer/Libs/MRML/Logic/vtkMRMLCornerTextLogic.h:138:8: error: ‘unordered_map’ in namespace ‘std’ does not name a template type
   std::unordered_map<
        ^~~~~~~~~~~~~
/usr/src/Slicer/Libs/MRML/Logic/vtkMRMLCornerTextLogic.cxx: In member function ‘virtual void vtkMRMLCornerTextLogic::PrintSelf(std::ostream&, vtkIndent)’:
/usr/src/Slicer/Libs/MRML/Logic/vtkMRMLCornerTextLogic.cxx:60:37: error: ‘class vtkMRMLCornerTextLogic’ has no member named ‘registeredProviders’; did you mean ‘RegisterNodes’?
   for (const auto& provider : this->registeredProviders)
                                     ^~~~~~~~~~~~~~~~~~~
                                     RegisterNodes
/usr/src/Slicer/Libs/MRML/Logic/vtkMRMLCornerTextLogic.cxx: In member function ‘bool vtkMRMLCornerTextLogic::RegisterPropertyValueProvider(const string&, vtkMRMLAbstractAnnotationPropertyValueProvider*)’:
/usr/src/Slicer/Libs/MRML/Logic/vtkMRMLCornerTextLogic.cxx:132:7: error: ‘registeredProviders’ was not declared in this scope
   if (registeredProviders.find(pluginName) == registeredProviders.end())
       ^~~~~~~~~~~~~~~~~~~
/usr/src/Slicer/Libs/MRML/Logic/vtkMRMLCornerTextLogic.cxx:132:7: note: suggested alternative: ‘RegisterNodes’
   if (registeredProviders.find(pluginName) == registeredProviders.end())
       ^~~~~~~~~~~~~~~~~~~
       RegisterNodes
/usr/src/Slicer/Libs/MRML/Logic/vtkMRMLCornerTextLogic.cxx: In member function ‘std::array<std::basic_string<char>, 8> vtkMRMLCornerTextLogic::GenerateAnnotations(vtkMRMLSliceNode*, vtkMRMLTextNode*, bool)’:
/usr/src/Slicer/Libs/MRML/Logic/vtkMRMLCornerTextLogic.cxx:260:16: error: ‘class vtkMRMLCornerTextLogic’ has no member named ‘positionMap’
     if (!this->positionMap.count(position))
                ^~~~~~~~~~~
/usr/src/Slicer/Libs/MRML/Logic/vtkMRMLCornerTextLogic.cxx:272:35: error: ‘class vtkMRMLCornerTextLogic’ has no member named ‘positionMap’
     TextLocation location = this->positionMap.at(position);
                                   ^~~~~~~~~~~
/usr/src/Slicer/Libs/MRML/Logic/vtkMRMLCornerTextLogic.cxx:335:44: error: ‘registeredProviders’ was not declared in this scope
       for (const auto [plugin, provider] : registeredProviders)
                                            ^~~~~~~~~~~~~~~~~~~
/usr/src/Slicer/Libs/MRML/Logic/vtkMRMLCornerTextLogic.cxx:335:44: note: suggested alternative: ‘RegisterNodes’
       for (const auto [plugin, provider] : registeredProviders)
                                            ^~~~~~~~~~~~~~~~~~~
                                            RegisterNodes
```

jcfr:
Summary of changes:
- Fix seg fault on startup ensuring pointers are all initialized
- Fix memory leak updating `vtkMRMLCornerTextLogic::ParseTextNode` to accept `vtkXMLDataParser*` as argument and remove calls to `Register`
- Fix memory leak switch `vtkMRMLDefaultAnnotationPropertyValueProvider:AppLogic` to use `vtkWeakPointer`
- Fix formatting of pointers and references (prefer `vtkObject*` and `std::string&` instead of `vtkObject *` and `std::string &`)
- Other minor tweaks

jcfr:
Before integrating, we should dispatch the changes currently associated with the usability commit into their respective ones.

PR #8149, opened 2025-01-20
COMP: Include runtime libraries to ensure compatibility in SlicerExecutionModel

jcfr:
This update resolves a termination fault that occurs when running the CLI executable built with a newer MSVC runtime on systems with older runtime libraries.

By explicitly including the runtime libraries, compatibility is ensured, avoiding reliance on the system's default runtime version.

List of SlicerExecutionModel changes:

```
$ git shortlog 0e28e41aa..91b921bd5 --no-merges
James Butler (1):
      COMP: Include system runtime libraries
```

### DIFF
--- a/SuperBuild/External_SlicerExecutionModel.cmake
+++ b/SuperBuild/External_SlicerExecutionModel.cmake
@@ -85,7 +85,7 @@ if(NOT DEFINED SlicerExecutionModel_DIR AND NOT Slicer_USE_SYSTEM_${proj})
 
   ExternalProject_SetIfNotDefined(
     Slicer_${proj}_GIT_TAG
-    "0e28e41aa167b477bc4889725ca08157c152e584"
+    "91b921bd5977c3384916ba4b03705d87b26067f7"
     QUIET
     )
 


jamesobutler (approved):

lassoan:
How does this fix work? 
Where are these msvc runtime libraries copied to?
Are the libraries copied into each extension's CLI module folder?

jcfr:
> How does this fix work? 
> Where are these msvc runtime libraries copied to?

Similarly to how system libraries are distributed for Slicer binaries installed in the top-level `bin` folder, the system libraries are also distributed along side the CLIs.

Copied from https://github.com/Slicer/SlicerExecutionModel/pull/142#issuecomment-2603323966
> Locally testing the changes, we can observe that the system libraries are now installed along side the CLIs (in addition of being installed in the top-level `bin` directory):
> 
> | Top-level install directory | `lib/Slicer-X.Y/cli-modules` install directory |
> |--|--|
> | image | !image |
> 

------------

> Are the libraries copied into each extension's CLI module folder?

Yes

jcfr:
An alternative approach is proposed in https://github.com/Slicer/Slicer/issues/8150

PR #8151, opened 2025-01-21
COMP: Relocate vtkMRMLInteractionEventData based on its usage

jcfr:
Rehome the class `vtkMRMLInteractionEventData` to `Libs/MRML/Core` so that it is located alongside `vtkMRMLInteractionNode`.

This addresses the dependency introduced in commit 598dea0b5f ("ENH: Add context menu when right-clicking in slice or 3D view", 2021-06-10).


Leengit (approved):
LGTM

lassoan:
MRML core already has interaction related features, but it would be better not to increase this trend. It would be nicer to keep the separation between MRML Core and DisplayableManager as much as possible to keep software complexity lower (it is easier to maintain two smaller highly cohesive packages than one big package). Bringing in some rendering/display/interaction related code to MRML core could even be infeasible due to VTK dependencies, but we are lucky this time that for some reason  `vtkEventDataDevice3D` (the base class of `vtkMRMLInteractionEventData`) is not defined in VTK/Interaction but for some reason it is in VTK/Common/Core. I say we are lucky because VTK's choice of adding `vtkEventDataDevice3D` to VTK/Common/Core is not a very well justified design choice but probably it was done only for convenience (the class is only used in VTK/Interaction, VTK/Rendering/(VR/OpenVR/OpenXR/ZSpace), and VTK/GUISupport).

I don't oppose this change, but I would like to double-check that it is a step in the right direction. The alternative would be to remove even the declaration of `vtkMRMLInteractionEventData` from MRML Core. Could you describe why this change is needed (send a link to the code that uses this)?

Leengit:
> Could you describe why this change is needed (send a link to the code that uses this)?

Greater availability within the build process is needed in order to implement Curved Planar Reformation within `vtkMRMLSliceLogic`.

lassoan:
Can you copy here a link to MRML Core code that uses vtkMRMLInteractionEventData object? (currently we are using an object pointer in MRML Core, so the declaration is enough; so I guess somewhere we need to use the object itself)

jcfr:
After we finalize the discussion related to this pull request, it will have to be rebased & updated once this other one is integrated:
* https://github.com/Slicer/Slicer/pull/8154

jcfr:
> Can you copy here a link to MRML Core code that uses vtkMRMLInteractionEventData object? 

Looks like relocating `vtkMRMLInteractionEventData` is becoming a hard requirement due to the relocation of `vtkMRMLMarkupsDisplayNode` into `Libs/MRML/Core`:


https://github.com/Slicer/Slicer/blob/6c4a45f323293a97eae7655c61f5ccec0010eb7b/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsDisplayNode.cxx#L674-L678

https://github.com/Slicer/Slicer/blob/6c4a45f323293a97eae7655c61f5ccec0010eb7b/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsDisplayNode.cxx#L719-L730

lassoan:
> Looks like relocating `vtkMRMLInteractionEventData` will become a hard requirement due to the relocation of `vtkMRMLMarkupsDisplayNode` into `Libs/MRML/Core`:

OK, this is a good justification. It would would be more clear then to make this move in the same pull request that moves all the other classes. In the future we may revisit how to better isolate the MRML data classes from rendering/3D interaction, but we don't need to do it now.

jcfr:
@lassoan  Thanks for the review :pray:

Per your suggestion, the commit relocating `vtkMRMLInteractionEventData` has been moved into the pull request  taking care of relocating Markups classes from `Modules/Loadable/Markups/MRML` to `Libs/MRML/Core`:
* https://github.com/Slicer/Slicer/pull/8155

I am now closing this pull request.

PR #8152, opened 2025-01-21
COMP: Remove explicit dependency of MarkupsROINode on JSON storage node

jcfr:
Anticipating the relocation of Markups nodes into `Libs/MRML/Core`, this commit removes the include of `vtkMRMLMarkupsROIJsonStorageNode.h` from `vtkMRMLMarkupsROINode.cxx`.

This is consistent with how `vtkMRMLMarkupsPlaneNode` and `vtkMRMLMarkupsNode` are implemented.


lassoan (approved):
The change looks good to me.

PR #8159, opened 2025-01-22
BUG: Suppress "Populating font family aliases" warning on macOS

jcfr:
Suppresses warnings like the following reported at application startup on macOS:
    
```
Populating font family aliases took 108 ms. Replace uses of missing font family "Courier" with one that exists to avoid this cost.
```
    
The root issue, tracked in QTBUG-98369 and QTBUG-99216, has been resolved in newer Qt versions (see qt/qtbase@64dd5a818). This commit conditionally applies a workaround to address the warning for older Qt versions.

### DIFF
--- a/Base/QTApp/qSlicerApplicationHelper.cxx
+++ b/Base/QTApp/qSlicerApplicationHelper.cxx
@@ -22,7 +22,11 @@
 
 // Qt includes
 #include <QFont>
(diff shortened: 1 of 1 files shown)


jcfr:
After locally building the change on macOS, I was able to confirm it prevent the warning from being reported :white_check_mark:

lassoan:
Nice!

PR #8169, opened 2025-01-24
DOC: Add documentation for help menu

### LINKED ISSUE #7634: Add documentation for Help menu

When helping users on the forum we don't have documentation to point to about the Help menu.

Add description of the items in the Help menu, ideally here: https://slicer.readthedocs.io/en/latest/user_guide/user_interface.html#application-menu

Other menus are described but not Help.  As a side note we should remove Twitter since it's not used much as far as I know.

We might also want to extend some of our help text in the documentation so users don't need to ask on the forum for some simpler questions (probably there are lots of people with the same questions who never ask on the forum).

ebrahimebrahim:
Closes #7634


jcfr (approved):

image

PR #8170, opened 2025-01-24
COMP: Ensure setting `run_ctest_with_upload` to `FALSE` skips package upload

### LINKED ISSUE #7825: Ensure setting `run_ctest_with_upload` skip package upload
## Summary

Despite of setting the env. variable `run_ctest_with_upload` to `FALSE`[^1], the windows stable package built running the script `bluestreak-vs2022-slicer_stable_package.bat`[^2] was uploaded:
* https://github.com/Slicer/Slicer/issues/7821

Reviewing `bluestreak-vs2022-slicer_stable_package.txt` confirms that `run_ctest_with_upload` was set:

The build of `packageupload` target is only conditional on the `run_ctest_with_packages` variable:

https://github.com/Slicer/Slicer/blob/b0ed88351af465159b101fc434802c83fd438522/CMake/SlicerDashboardDriverScript.cmake#L561-L572

To address this, two approaches:
* build the `packageupload` target conditional to both `run_ctest_with_packages` and `run_ctest_with_upload`
* update `CMake/SlicerPackageAndUploadTarget.cmake` to check env. variable 

[^1]: https://discourse.slicer.org/t/windows-dashboard-extension-builds-interrupted-due-to-hardware-issue/35688/16?u=jcfr
[^2]: https://github.com/Slicer/DashboardScripts/blob/main/bluestreak-vs2022-slicer_stable_package.bat

## Steps to reproduce

NA

## Environment

NA

jadh4v:
Address the issue mentioned in:
 - #7825 

Moved the building of TARGET `packageupload` under the condition that `run_ctest_with_upload` is true.
The condition `run_ctest_with_packages` is already true under the situated code (the outer if condition).

closes #7825

### DIFF
--- a/CMake/SlicerDashboardDriverScript.cmake
+++ b/CMake/SlicerDashboardDriverScript.cmake
@@ -561,8 +561,14 @@ ${ADDITIONAL_CMAKECACHE_OPTION}
         message(STATUS "Packaging and uploading Slicer to packages server ...")
         set(package_list)
         if(run_ctest_with_packages)
+
+          set(package_target "package")
+          if(run_ctest_with_upload)
+            set(package_target "packageupload")
+          endif()
+
           ctest_build(
-            TARGET packageupload
+            TARGET ${package_target}
             BUILD ${slicer_build_dir}
             APPEND
             )


CMake/SlicerDashboardDriverScript.cmake
```diff
@@ -569,9 +571,7 @@ ${ADDITIONAL_CMAKECACHE_OPTION}
           if(run_ctest_submit)
             ctest_submit(PARTS Build)
           endif()
-        endif()
 
-        if(run_ctest_with_upload)
```
jcfr, line 574:
```suggestion
        endif()

        if(run_ctest_with_upload)
```

CMake/SlicerDashboardDriverScript.cmake
```diff
@@ -560,7 +560,9 @@ ${ADDITIONAL_CMAKECACHE_OPTION}
 
         message(STATUS "Packaging and uploading Slicer to packages server ...")
         set(package_list)
-        if(run_ctest_with_packages)
+
+        if(run_ctest_with_upload)
+
           ctest_build(
             TARGET packageupload
```
jcfr, line 567:
```suggestion
        if(run_ctest_with_packages)

          set(package_target "package")
          if(run_ctest_with_upload)
            set(package_target "packageupload")            
          endif()

          ctest_build(
            TARGET packageupload
```

jcfr (changes requested):

Since created the package without uploading it is useful, I suggest to instead conditionally define the packaging target.

For reference, the `package` target is used here:

https://github.com/Slicer/Slicer/blob/d798402abaa8f47ff042f99f073c10f4f5af0f6a/CMake/SlicerPackageAndUploadTarget.cmake#L179-L184

jcfr (approved):

PR #8173, opened 2025-01-24
ENH: Centralize retrieval of Markups Moving attributes in display node

jadh4v:
# Brief Description

Streamline reuse of attributes `"Markups.MovingInSliceView"` and `"Markups.MovingMarkupIndex"` by declaring `static GetMoving(InSliceView|MarkupIndex)AttributeName()` functions in the `vtkMRMLMarkupsDisplayNode` class as public members. This will allow for consistent use of the attribute strings across C++ and Python scripts.

```cpp
  static const char* GetMovingInSliceViewAttributeName() { return "Markups.MovingInSliceView"; }
  static const char* GetMovingMarkupIndexAttributeName() { return "Markups.MovingMarkupIndex"; }

```

> [!NOTE]
> The function naming convention of `GetxxxxxxxxAttributeName()` is chosen so that we understand from the function name that it is returning a name of an attribute and not the attribute value itself.

### DIFF
--- a/Docs/developer_guide/script_repository/markups.md
+++ b/Docs/developer_guide/script_repository/markups.md
@@ -513,7 +513,7 @@ Event management of Slicer-4.11 version is still subject to change. The example 
 ```python
 def onMarkupChanged(caller,event):
   markupsNode = caller
-  sliceView = markupsNode.GetAttribute(slicer.vtkMRMLMarkupsNode.GetMovingInSliceViewAttributeName())
+  sliceView = markupsNode.GetAttribute(slicer.vtkMRMLMarkupsDisplayNode.GetMovingInSliceViewAttributeName())
   movingMarkupIndex = markupsNode.GetDisplayNode().GetActiveControlPoint()
   if movingMarkupIndex >= 0:
     pos = [0,0,0]
@@ -528,13 +528,13 @@ def onMarkupChanged(caller,event):
 
 def onMarkupStartInteraction(caller, event):
   markupsNode = caller
-  sliceView = markupsNode.GetAttribute(slicer.vtkMRMLMarkupsNode.GetMovingInSliceViewAttributeName())
+  sliceView = markupsNode.GetAttribute(slicer.vtkMRMLMarkupsDisplayNode.GetMovingInSliceViewAttributeName())
   movingMarkupIndex = markupsNode.GetDisplayNode().GetActiveControlPoint()
   logging.info("Start interaction: point ID = {0}, slice view = {1}".format(movingMarkupIndex, sliceView))
 
 def onMarkupEndInteraction(caller, event):
   markupsNode = caller
-  sliceView = markupsNode.GetAttribute(slicer.vtkMRMLMarkupsNode.GetMovingInSliceViewAttributeName())
+  sliceView = markupsNode.GetAttribute(slicer.vtkMRMLMarkupsDisplayNode.GetMovingInSliceViewAttributeName())
   movingMarkupIndex = markupsNode.GetDisplayNode().GetActiveControlPoint()
   logging.info("End interaction: point ID = {0}, slice view = {1}".format(movingMarkupIndex, sliceView))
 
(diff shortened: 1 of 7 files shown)


Modules/Loadable/SubjectHierarchy/Testing/Python/SubjectHierarchyGenericSelfTest.py
```diff
@@ -546,31 +546,31 @@ def testAttributeFilters(filteredObject, proxyModel, showEmptyHierarchyItems=Tru
 
             # Check include node attribute name filter
             self.assertEqual(shProxyModel.acceptedItemCount(shNode.GetSceneItemID()), 9)
-            filteredObject.includeNodeAttributeNamesFilter = ["Markups.MovingInSliceView"]
+            filteredObject.includeNodeAttributeNamesFilter = [slicer.vtkMRMLMarkupsNode.GetMovingInSliceViewAttributeName()]
```
lassoan, line 549: I think these tests could be done with any string. @cpinter was there any specific reason to use the `Markups.MovingInSliceView` string as attribute name here?

Docs/developer_guide/script_repository/markups.md
```diff
@@ -513,7 +513,7 @@ Event management of Slicer-4.11 version is still subject to change. The example
 ```python
 def onMarkupChanged(caller,event):
   markupsNode = caller
-  sliceView = markupsNode.GetAttribute("Markups.MovingInSliceView")
+  sliceView = markupsNode.GetAttribute(slicer.vtkMRMLMarkupsNode.GetMovingInSliceViewAttributeName())
```
lassoan, line 516: If we make `MovingInSliceView` information public. It could be more developer-friendly to add a `GetMovingInSliceView` (maybe also a `SetMovingInSliceView` and `ClearMovingInSliceView`) method to the markups display node.

Modules/Loadable/SubjectHierarchy/Testing/Python/SubjectHierarchyGenericSelfTest.py
cpinter: I think I used this because it was easier to test against an attribute that was easy to change with simple interactions. So no.

Docs/developer_guide/script_repository/markups.md
jcfr: >  It could be more developer-friendly [...] methods

After discussing with @lassoan , we agreed that moving forward it would be sensible to relying on an dedicated API to check the status of Markups and associated control points. That said, this is a more involved task that will be done later:
* https://github.com/Slicer/Slicer/issues/8200

In the meantime, we will move forward with integrating the functions to get the attribute name instead of directly using string.

Modules/Loadable/Markups/MRML/vtkMRMLMarkupsNode.h
```diff
@@ -921,6 +921,10 @@ class  VTK_SLICER_MARKUPS_MODULE_MRML_EXPORT vtkMRMLMarkupsNode : public vtkMRML
     return this->ReplaceListNameInControlPointLabelFormat();
   };
 
+  // Constants for attribute names
+  static const char* GetMovingInSliceViewAttributeName() { return "Markups.MovingInSliceView"; }
+  static const char* GetMovingMarkupIndexAttributeName() { return "Markups.MovingMarkupIndex"; }
```
jcfr, line 926: I will relocate those into the display node.

jcfr:
Once local testing is done, this will be integrated using `Squash & Merge` specifying the following commit message:

```
ENH: Centralize retrieval of Markups Moving attributes in display node

Add static functions in `vtkMRMLMarkupsDisplayNode` to retrieve  
attribute names for `Markups.MovingInSliceView` and `Markups.MovingMarkupIndex`.

These attributes define the interaction context in which a Markups control
point is being moved, making `vtkMRMLMarkupsDisplayNode` a more appropriate
location for them.

Replace all direct occurrences of these attribute strings with the new
static functions, ensuring consistency, improving maintainability, and
simplifying future refactoring.

```

jcfr (approved):

Issue #7721, opened 2024-05-02
Update extension build system to support specifying contributors as a regular CMake list


jcfr:
## Is your feature request related to a problem? Please describe.

> It is OK to enable this syntax, but I find it a bit complicated. Could standard CMake-style lists syntax be used here?
> 
> ```
>   EXTENSION_CATEGORY "Exporter"
>   EXTENSION_CONTRIBUTORS
>     "Jean-Christophe Fillion-Robin (Kitware)"
>     "Pat Marion (Kitware)"
>     "Steve Pieper (Isomics)"
>     "Atsushi Yamada (Shiga University of Medical Science)"
>   EXTENSION_HOMEPAGE "https://www.slicer.org/w/index.php/Documentation/Nightly/Extensions/SlicerToKiwiExporter"
> ```

_Originally posted by @lassoan in https://github.com/Slicer/Slicer/pull/7720#discussion_r1587064965_

## Describe the solution you'd like

Implement approach similar to 1b9bd54b06 (`ENH: Support specifying screenshot url extension metadata as string or list`, 2024-04-16)

## Describe alternatives you've considered

NA

## Additional context

* https://github.com/Slicer/Slicer/pull/7720#pullrequestreview-2034825801

sbelsk:
I can take a look at this

PR #8175, opened 2025-01-24
ENH: Support specifying extension contributors as CMake list

sbelsk:
Resolves https://github.com/Slicer/Slicer/issues/7721.
The `EXTENSION_CONTRIBUTORS` argument to the function [`slicerFunctionGenerateExtensionDescription()`](https://github.com/Slicer/Slicer/blob/80c243f06764d81b0dbb7d40ef4832baf7296f8f/Extensions/CMake/SlicerFunctionGenerateExtensionDescription.cmake#L32-L130) now supports both a string input with the contributors names (to be joined by `, `), as well as a CMake list of the contributors, still applying suitable whitespace stripping. The test for this function has been updated to verify both cases.


jcfr:
@sbelsk Thanks for the contribution 🙏

I slightly revisited the proposed approach so that it can be used for both `screenshots` and `contributors` 🚀

jcfr (approved):

PR #8178, opened 2025-01-24
ENH: Warn when setting volume node with invalid vtkImageData

### LINKED ISSUE #6911: SetAndObserveImageData must have an image with unit spacing
## Summary

Calling `vtkMRMLVolumeNode::SetAndObserveImageData` with `vtkImageData` which has non-unit spacings results in visual bugs. @sjh26 helped me figure out that bounds were computed incorrectly. I was stepping through the C++ code in an attempt to figure out where the spacing needs to be ignored, unsuccessfully.

## Steps to reproduce

Use [this image](https://data.kitware.com/#item/57b5d5d88d777f10f269444b) as input to [BModeFromRF module](https://github.com/KitwareMedical/SlicerITKUltrasound/tree/6ed6b9556a0507abba80648154949649653255d8/BModeFromRF). Comment out this line to observe the problem:
https://github.com/KitwareMedical/SlicerITKUltrasound/blob/6ed6b9556a0507abba80648154949649653255d8/ITKUltrasoundCommon/ITKUltrasoundCommon.py#L213

Only a small part of the image is displayed correctly with small spacing (approximately 0.1x0.01)
0.1 spacing

## Expected behavior

Setting unit spacing to `vtkImageData` before calling `vtkMRMLVolumeNode::SetAndObserveImageData` is not required for proper functioning, and it is enough for `vtkMRMLVolumeNode::SetIJKToRASMatrix` to incorporate spacing.

Image looks normal with unit spacing:
1.0 spacing

## Environment
- Problem reproduced with two slicer versions: Slicer-5.2.1 official release, and a local build of 3d62cbfd81c08e7961d568be18f2bd39a860a543.

ebrahimebrahim:
Print a warning when tryng to get a volume node to observe a vtkImageData that has a non-identity affine (i.e. origin isn't (0,0.0), or spacing isn't (1,1,1), or direction matrix isn't identity)

Closes #6911

### DIFF
--- a/Libs/MRML/Core/vtkMRMLVolumeNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLVolumeNode.cxx
@@ -670,6 +670,38 @@ const char* vtkMRMLVolumeNode::ComputeScanOrderFromIJKToRAS(vtkMatrix4x4 *ijkToR
 }
 
 //----------------------------------------------------------------------------
+namespace
+{
+
+//----------------------------------------------------------------------------
+// Checks that a vtkImageData has origin (0,0,0), spacing (1,1,1), and an identity direction matrix
+bool IsImageDataGeometryValid(vtkImageData* imageData)
+{
+  double spacing[3] = { 0.0 };
+  imageData->GetSpacing(spacing);
+  if (spacing[0] != 1.0 || spacing[1] != 1.0 || spacing[2] != 1.0)
+  {
+    return false;
+  }
+  double origin[3] = { 0.0 };
+  imageData->GetOrigin(origin);
+  if (origin[0] != 0.0 || origin[1] != 0.0 || origin[2] != 0.0)
+  {
+    return false;
+  }
+  vtkMatrix3x3* directionMatrix = imageData->GetDirectionMatrix();
+  vtkNew<vtkMatrix3x3> identityMatrix;
+  identityMatrix->Identity();
+  if (!vtkAddonMathUtilities::MatrixAreEqual(directionMatrix, identityMatrix, /* tolerance= */ 0.0))
+  {
+    return false;
+  }
+  return true;
+}
+
+} // end of anonymous namespace
+
+//----------------------------------------------------------------------------
 void vtkMRMLVolumeNode::SetAndObserveImageData(vtkImageData *imageData)
 {
   if (imageData == nullptr)
@@ -685,6 +717,16 @@ void vtkMRMLVolumeNode::SetAndObserveImageData(vtkImageData *imageData)
   }
   else
   {
+    if (!IsImageDataGeometryValid(imageData))
+    {
+      vtkWarningMacro(
+        "SetAndObserveImageData: The vtkImageData associated with the Volume node "
+        << (this->GetID() != nullptr ? this->GetID() : "(unknown)")
+        << " does not meet the required properties:"
+        << " Origin must be (0,0,0), Spacing must be (1,1,1),"
+        << " and Direction must be an identity matrix."
+      );
+    }
     vtkTrivialProducer* oldProducer = vtkTrivialProducer::SafeDownCast(
       this->GetImageDataConnection() ? this->GetImageDataConnection()->GetProducer() : nullptr);
     if (oldProducer && oldProducer->GetOutputDataObject(0) == imageData)


Libs/MRML/Core/vtkMRMLVolumeNode.cxx
```diff
@@ -669,6 +669,43 @@ const char* vtkMRMLVolumeNode::ComputeScanOrderFromIJKToRAS(vtkMatrix4x4 *ijkToR
   }
 }
 
+//----------------------------------------------------------------------------
+// Checks that a vtkImageData has origin (0,0,0), spacing (1,1,1), and an identity direction matrix
+bool imageDataHasIdentityAffine(vtkImageData *imageData)
+{
+  // Should we be using vtkMathUtilities::FuzzyCompare here?
+  // I am thinking no because we vtkImageData affine should be *exactly* identity
+  // (we don't want "slightly" undefined behavior from slicer).
```
ebrahimebrahim, line 678: Note my question here. I believe we should be checking for exact equality of floats throughout in this function. Just want to double check that everyone agrees.

Libs/MRML/Core/vtkMRMLVolumeNode.cxx
```diff
@@ -669,6 +669,43 @@ const char* vtkMRMLVolumeNode::ComputeScanOrderFromIJKToRAS(vtkMatrix4x4 *ijkToR
   }
 }
 
+//----------------------------------------------------------------------------
+// Checks that a vtkImageData has origin (0,0,0), spacing (1,1,1), and an identity direction matrix
+bool imageDataHasIdentityAffine(vtkImageData *imageData)
```
ebrahimebrahim, line 674: Is this an appropriate place to put this utility function?

dzenanz (commented):
LGTM, but somebody else should review it too.

Libs/MRML/Core/vtkMRMLVolumeNode.cxx
```diff
@@ -669,6 +669,43 @@ const char* vtkMRMLVolumeNode::ComputeScanOrderFromIJKToRAS(vtkMatrix4x4 *ijkToR
   }
 }
 
+//----------------------------------------------------------------------------
```
jcfr, line 672:
```suggestion
//----------------------------------------------------------------------------
namespace
{

//----------------------------------------------------------------------------
```

Libs/MRML/Core/vtkMRMLVolumeNode.cxx
```diff
@@ -669,6 +669,43 @@ const char* vtkMRMLVolumeNode::ComputeScanOrderFromIJKToRAS(vtkMatrix4x4 *ijkToR
...
+        return false;
+      }
+      if (i!=j && directionMatrix->GetElement(i,j)!=0.0)
+      {
+        return false;
+      }
+    }
+  }
+  return true;
+}
```
jcfr, line 707:
```suggestion
}

} // end of anonymous namespace
```

Libs/MRML/Core/vtkMRMLVolumeNode.cxx
jcfr, line 674:
```suggestion
bool HasDefaultSpatialProperties(vtkImageData* imageData)
```

Libs/MRML/Core/vtkMRMLVolumeNode.cxx
```diff
@@ -685,6 +722,13 @@ void vtkMRMLVolumeNode::SetAndObserveImageData(vtkImageData *imageData)
   }
   else
   {
+    if (!imageDataHasIdentityAffine(imageData))
```
jcfr, line 725:
```suggestion
    if (!HasDefaultSpatialProperties(imageData))
```

jcfr (commented):

After addressing few nitpicks, it should be good for integration :ok_hand:

jcfr:
Suggested commit message (and matching pull request title):

```
ENH: Warn when setting volume node with invalid vtkImageData

This commit adds a check in `vtkMRMLVolumeNode::SetAndObserveImageData` to validate 
that the vtkImageData being set has default spatial properties:
- Origin: (0,0,0)
- Spacing: (1,1,1)
- Direction: Identity matrix

A warning is displayed if these conditions are not met.
```

Libs/MRML/Core/vtkMRMLVolumeNode.cxx
lassoan: I would call this method `VerifyImageDataGeometry` or `IsImageDataGeometryValid`, because we usually refer to image origin, spacing, axis directions as "geometry", and the goal of the check is to verify if it is valid (not that it equals to some default; and not that it is specifically identity - because in the future non-identity matrix will be accepted).

Libs/MRML/Core/vtkMRMLVolumeNode.cxx
lassoan: I agree, exact check is appropriate here (especially because both 0.0 and 1.0 can be represented exactly as float).

Libs/MRML/Core/vtkMRMLVolumeNode.cxx
```diff
@@ -669,6 +669,43 @@ const char* vtkMRMLVolumeNode::ComputeScanOrderFromIJKToRAS(vtkMatrix4x4 *ijkToR
...
+      if (i==j && directionMatrix->GetElement(i,j)!=1.0)
+      {
+        return false;
+      }
+      if (i!=j && directionMatrix->GetElement(i,j)!=0.0)
+      {
+        return false;
+      }
+    }
+  }
```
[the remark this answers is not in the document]
ebrahimebrahim: Problem: the `MatrixAreEqual` utility doesn't work with `tolerance=0.0` due to the non-strict `>=` in the comparison [here](https://github.com/Slicer/vtkAddon/blob/63ccc900354dabe7f53c34630d90f04056ca11ea/vtkAddonMathUtilities.cxx#L123).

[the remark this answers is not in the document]
lassoan: Good catch! This is an incorrect (but at least very very unusal, unexpected, and therefore error-prone) interpretation of a "tolerance" value. In this PR, at the minimum we need to add a comment to the code why we did not use `MatrixAreEqual`. But we should fix `MatrixAreEqual` or at least submit an issue to do it later.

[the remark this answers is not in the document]
ebrahimebrahim: Alright, I'll go with that here. I made https://github.com/Slicer/vtkAddon/pull/54 but I have doubts about there as well

Libs/MRML/Core/vtkMRMLVolumeNode.cxx
```diff
@@ -669,6 +669,48 @@ const char* vtkMRMLVolumeNode::ComputeScanOrderFromIJKToRAS(vtkMatrix4x4 *ijkToR
...
+  double origin[3] = { 0.0 };
+  imageData->GetOrigin(origin);
+  if (origin[0] != 0.0 || origin[1] != 0.0 || origin[2] != 0.0)
+  {
+    return false;
+  }
+  vtkMatrix3x3* directionMatrix = imageData->GetDirectionMatrix();
+  // We do not use vtkAddonMathUtilities::MatrixAreEqual and instead
+  // check the matrix elements individually here because of
+  // https://github.com/Slicer/vtkAddon/issues/53
```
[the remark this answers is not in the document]
ebrahimebrahim: Thanks, I updated it with a slightly different phrasing to clarify what the issue is.

jcfr:
Once the following pull request is integrated, we will be able to simplify:
* https://github.com/Slicer/Slicer/pull/8227

ebrahimebrahim:
Simplified now that #8227 is merged. It seems to work well.

Ready for re-review @jcfr and/or @lassoan

jcfr (approved):
Thanks for moving forward with this :pray: :rocket:

dzenanz (approved):

Issue #5827, opened 2021-08-27
Add support for string and bool arrays in `slicer.util.updateTableFromArray()`


shunxinyi2021:
`slicer.util.updateTableFromArray()` only support `float`?
How to add a new column which type is "string" or "bool" to Table in Slicer?

lassoan:
This convenience function supports arrays of floats. You can access all array types from Python, though, so it would not be hard to add support for other array types, too.

pka420:
Hey @lassoan 
I would like to work on this issue.

lassoan:
Awesome! Let us know if you need any help.

Kent-McPhail:
Hi I am trying to start to take on this issue.  One thing I am having troube with starting with is how to best for editing.  I nevr have workd with this large of a codebase and was wondering if there was any documentation on getting started with a first issue.. Whether that includes setting up the environment in VS.  I am just looking for a best practice for development workflow here.  I have got slicer built and running already

PR #8180, opened 2025-01-24
ENH: Add support for additional NumPy types in `slicer.util.updateTableFromArray`

jadh4v:
See updated description below: https://github.com/Slicer/Slicer/pull/8180#issuecomment-2649364818

<details><summary>Previous description</summary>
<p>

Allow additional `dtypes` to be handled when converting to `vtk` array. All types supported by `vtk.util.numpy_support.get_vtk_array_type` can be converted.

However, I don't think this still supports `boolean` or `string` arrays.
We need to figure out a way to generate and possibly use the following VTK array types:
`bool` numpy array --> `vtkBitArray`
`string` numpy array --> `vtkStringArray`

</p>
</details> 

Partially addresses:
 - https://github.com/Slicer/Slicer/issues/5827

### DIFF
--- a/Base/Python/slicer/util.py
+++ b/Base/Python/slicer/util.py
@@ -2500,13 +2500,19 @@ def arrayFromTableColumnModified(tableNode, columnName):
 def updateTableFromArray(tableNode, narrays, columnNames=None):
     """Set values in a table node from a NumPy array or an array-like object (list/tuple of NumPy arrays).
 
+    This function can handle:
+
+      - **Structured (record) array** with named fields. Each field becomes a separate column.
+        If ``columnNames`` is not provided, the field names in the dtype are used.
+      - **1D NumPy array**, which is added as a single column.
+      - **2D NumPy array**, where each column in the array is added as a separate column in the
+        table node (note that the array is transposed).
+      - **List/tuple** of 1D NumPy arrays, which are each added as separate columns.
+
     :param tableNode: The table node to be updated. If ``None``, a new ``vtkMRMLTableNode`` is
       created and added to the scene.
     :type tableNode: vtkMRMLTableNode or None
-    :param narrays: One of:
-        - A 1D NumPy array
-        - A 2D NumPy array (values will be transposed so each column becomes a table column)
-        - A list/tuple of 1D NumPy arrays
+    :param narrays: NumPy array or array-like object (structured array, 1D/2D array, or list/tuple of 1D arrays).
     :type narrays: np.ndarray, tuple, or list
     :param columnNames: Optional string or list of strings specifying names for the columns. If fewer
       names are provided than columns, only the specified columns are named;
@@ -2537,20 +2543,39 @@ def updateTableFromArray(tableNode, narrays, columnNames=None):
 
     if tableNode is None:
         tableNode = slicer.mrmlScene.AddNewNodeByClass("vtkMRMLTableNode")
-    if isinstance(narrays, np.ndarray) and len(narrays.shape) == 1:
+
+    if isinstance(narrays, np.ndarray) and narrays.dtype.names:
+        # Structured array with named fields
+        fieldNames = narrays.dtype.names
+        if columnNames is None:
+            columnNames = list(fieldNames)
+        ncolumns = [narrays[fieldName] for fieldName in fieldNames]
+    elif isinstance(narrays, np.ndarray) and len(narrays.shape) == 1:
+        # Single 1D array
         ncolumns = [narrays]
     elif isinstance(narrays, np.ndarray) and len(narrays.shape) == 2:
+        # 2D array -> transpose so columns become separate arrays
         ncolumns = narrays.T
     elif isinstance(narrays, tuple) or isinstance(narrays, list):
+        # List or tuple of 1D arrays
         ncolumns = narrays
     else:
-        raise ValueError("Expected narrays is a numpy ndarray, or tuple or list of numpy ndarrays, got %s instead." % (str(type(narrays))))
+        # Unrecognized format
+        msg = (
+            "Expected a structured NumPy array (with named fields), "
+            "a 1D/2D NumPy array, or a list/tuple of 1D arrays; "
+            "got {object_type} instead.".format(object_type=type(narrays))
+        )
+        raise ValueError(msg)
+
     tableNode.RemoveAllColumns()
     # Convert single string to a single-element string list
     if columnNames is None:
         columnNames = []
     if isinstance(columnNames, str):
         columnNames = [columnNames]
+
+    # For each extracted column, convert to a VTK array and add to the table.
     for columnIndex, ncolumn in enumerate(ncolumns):
         vtype = vtk.util.numpy_support.get_vtk_array_type(ncolumn.dtype)
         vcolumn = vtk.util.numpy_support.numpy_to_vtk(num_array=ncolumn.ravel(), deep=True, array_type=vtype)
--- a/Base/Python/slicer/util.py
+++ b/Base/Python/slicer/util.py
@@ -2528,7 +2528,7 @@ def _vtkArrayFromNumpyArray(narray, deep=False, arrayType=None):
     return vtk.util.numpy_support.numpy_to_vtk(narray, deep=deep, array_type=arrayType)
 
 
-def updateTableFromArray(tableNode, narrays, columnNames=None):
+def updateTableFromArray(tableNode, narrays, columnNames=None, setBoolAsUchar=False):
     """Set values in a table node from a NumPy array or an array-like object (list/tuple of NumPy arrays).
 
     This function can handle:
@@ -2549,6 +2549,10 @@ def updateTableFromArray(tableNode, narrays, columnNames=None):
       names are provided than columns, only the specified columns are named;
       otherwise columns get default names. If ``None`` is passed, no column names are set.
     :type columnNames: str, list of str, or None
+    :param setBoolAsUchar: If ``True``, boolean arrays are converted to ``uint8`` for plotting compatibility.
+      This leverages the :func:`_vtkArrayFromNumpyArray` function's handling of ``VTK_BIT`` arrays.
+      Default is ``False``.
+    :type setBoolAsUchar: bool
     :return: Updated ``vtkMRMLTableNode``.
     :raises ValueError: If the input ``narrays`` is not a recognized format.
 
@@ -2610,7 +2614,13 @@ def updateTableFromArray(tableNode, narrays, columnNames=None):
     for columnIndex, ncolumn in enumerate(ncolumns):
         vtype = None
         if ncolumn.dtype == np.bool_:
-            vtype = vtk.VTK_BIT
+            if setBoolAsUchar:
+                # Convert boolean arrays to uint to ensure compatibility with VTK plotting infrastructure:
+                # vtkContext2D does not support vtkBitArray, and plotting calls that rely on
+                # vtkArrayDownCast<vtkFloatArray>() would fail with a bit array.
+                ncolumn = ncolumn.astype("uint8")
+            else:
+                vtype = vtk.VTK_BIT
         if vtype is None:
             vtype = vtk.util.numpy_support.get_vtk_array_type(ncolumn.dtype)
         vcolumn = _vtkArrayFromNumpyArray(ncolumn.ravel(), deep=True, arrayType=vtype)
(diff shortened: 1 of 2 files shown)


jcfr:
## Updates

### Add support for additional NumPy types in `slicer.util.updateTableFromArray`

This patch removes the hard-coded use of `VTK_FLOAT` in `updateTableFromArray` and instead calls `vtk.util.numpy_support.get_vtk_array_type` to automatically map NumPy dtypes to the appropriate VTK array types. This allows for proper handling of a broader range of numerical and complex data types, as detailed below:

| NumPy        | VTK                      |
|--------------|--------------------------|
| `uint8`      | `VTK_UNSIGNED_CHAR`      |
| `uint16`     | `VTK_UNSIGNED_SHORT`     |
| `uint32`     | `VTK_UNSIGNED_INT`       |
| `uint64`     | `VTK_UNSIGNED_LONG_LONG` |
| `int8`       | `VTK_CHAR`               |
| `int16`      | `VTK_SHORT`              |
| `int32`      | `VTK_INT`                |
| `int64`      | `VTK_LONG_LONG`          |
| `float32`    | `VTK_FLOAT`              |
| `float64`    | `VTK_DOUBLE`             |
| `complex64`  | `VTK_FLOAT`              |
| `complex128` | `VTK_DOUBLE`             |

By leveraging these mappings, `slicer.util.updateTableFromArray` now supports int, float, and complex arrays without requiring manual conversions.

----

### Support structured (record) arrays in slicer.util.updateTableFromArray

This patch adds support for NumPy structured arrays that have named fields.
When the function detects a structured array (narrays.dtype.names is non-empty),
it iterates over each field and converts that data into a separate column
in the table. If columnNames is not specified, it defaults to using the field
names from the NumPy dtype.

----

### Convert boolean arrays to uint in slicer.util.updateTableFromArray 

This addresses an issue where boolean arrays were mapped to vtkBitArray, which the underlying VTK plotting infrastructure cannot handle. Now, if a column is detected to be of dtype `np.bool_`, it is automatically converted to an unsigned integer array before being added to the table.

jcfr:
For future reference, attempt to directly handle array of boolean was done applying a patch like the following:

```diff
diff --git a/Base/Python/slicer/util.py b/Base/Python/slicer/util.py
index 0882923249..dfa75fbe29 100644
--- a/Base/Python/slicer/util.py
+++ b/Base/Python/slicer/util.py
@@ -2576,15 +2576,30 @@ def updateTableFromArray(tableNode, narrays, columnNames=None):
     if isinstance(columnNames, str):
         columnNames = [columnNames]

+    def _numpy_to_vtk(num_array, deep=0, array_type=None):
+        if array_type == vtk.VTK_BIT:
+            packed_bits = np.packbits(num_array)
+            result_array = vtk.util.numpy_support.create_vtk_array(array_type)
+            result_array.SetNumberOfComponents(1)
+            shape = num_array.shape
+            result_array.SetNumberOfTuples(shape[0])
+            result_array.SetVoidArray(packed_bits, shape[0], 1)
+            # Since packed_bits is a standalone object with no references from the caller.
+            # As such, it will drop out of this scope and cause memory issues if we
+            # do not deep copy its data.
+            copy = result_array.NewInstance()
+            copy.DeepCopy(result_array)
+            result_array = copy
+            return result_array
+        return vtk.util.numpy_support.numpy_to_vtk(num_array, deep=deep, array_type=array_type)
+
     # For each extracted column, convert to a VTK array and add to the table.
     for columnIndex, ncolumn in enumerate(ncolumns):
         if ncolumn.dtype == np.bool_:
-            # Convert boolean arrays to uint to ensure compatibility with VTK plotting infrastructure:
-            # vtkContext2D does not support vtkBitArray, and plotting calls that rely on
-            # vtkArrayDownCast<vtkFloatArray>() would fail with a bit array.
-            ncolumn = ncolumn.astype("uint")
-        vtype = vtk.util.numpy_support.get_vtk_array_type(ncolumn.dtype)
-        vcolumn = vtk.util.numpy_support.numpy_to_vtk(num_array=ncolumn.ravel(), deep=True, array_type=vtype)
+            vtype = vtk.VTK_BIT
+        else:
+            vtype = vtk.util.numpy_support.get_vtk_array_type(ncolumn.dtype)
+        vcolumn = _numpy_to_vtk(num_array=ncolumn.ravel(), deep=True, array_type=vtype)
         if (columnNames is not None) and (columnIndex < len(columnNames)):
             vcolumn.SetName(columnNames[columnIndex])
         tableNode.AddColumn(vcolumn)
diff --git a/Libs/MRML/Widgets/qMRMLPlotView.cxx b/Libs/MRML/Widgets/qMRMLPlotView.cxx
index d29ffd5699..2da8c8bfe3 100644
--- a/Libs/MRML/Widgets/qMRMLPlotView.cxx
+++ b/Libs/MRML/Widgets/qMRMLPlotView.cxx
@@ -258,7 +258,7 @@ vtkSmartPointer<vtkPlot> qMRMLPlotViewPrivate::updatePlotFromPlotSeriesNode(vtkM
     return nullptr;
   }
   int yColumnType = yColumn->GetDataType();
-  if (yColumnType == VTK_STRING || yColumnType == VTK_BIT)
+  if (yColumnType == VTK_STRING /*|| yColumnType == VTK_BIT*/)
   {
     qWarning() << Q_FUNC_INFO << ": Y column has unsupported data type: 'string' or 'bit'";
     return nullptr;
```

and this resulted in runtime error like the following:

```
[VTK] Attempted to paint a line with <2 points.
[VTK] Attempted to paint a line with <2 points.
[VTK] Attempted to paint a line with <2 points.
```

jcfr:
The code used to test support for structure array with boolean column is the following:

```py
def plotTableNode(table_node, plot_name, column_names, y_axis_range=None, plot_type=slicer.vtkMRMLPlotSeriesNode.PlotTypeScatterBar):
    # Create plot
    plotSeriesNode = slicer.mrmlScene.AddNewNodeByClass("vtkMRMLPlotSeriesNode", plot_name)
    plotSeriesNode.SetAndObserveTableNodeID(table_node.GetID())
    plotSeriesNode.SetXColumnName(column_names[0])
    plotSeriesNode.SetYColumnName(column_names[1])
    plotSeriesNode.SetPlotType(plot_type)
    plotSeriesNode.SetColor(0, 0.6, 1.0)

    # Create chart and add plot
    plotChartNode = slicer.mrmlScene.AddNewNodeByClass("vtkMRMLPlotChartNode")
    plotChartNode.AddAndObservePlotSeriesNodeID(plotSeriesNode.GetID())
    if y_axis_range is not None:
        plotChartNode.YAxisRangeAutoOff()
        plotChartNode.SetYAxisRange(0, y_axis_range)
    else:
        plotChartNode.YAxisRangeAutoOn()

    # Show plot in layout
    slicer.modules.plots.logic().ShowChartInLayout(plotChartNode)


import numpy as np

# Create a new table node
tableNodeBool = slicer.mrmlScene.AddNewNodeByClass("vtkMRMLTableNode", "BoolData")

# Define the data types for each column
data_type = np.dtype([('position', int), ('enabled', bool)])

# Create a structured array
boolArray = np.array([
    (1, False),
    (2, True),
    (3, True)
], dtype=data_type)

# Update the table from the boolean array
slicer.util.updateTableFromArray(tableNodeBool, boolArray)

# Rename columns
tableBool = tableNodeBool.GetTable()
tableBool.GetColumn(0).SetName("Position")
tableBool.GetColumn(1).SetName("Enabled")

plotTableNode(tableNodeBool, "Boolean", ["Position", "Enabled"], y_axis_range=2, plot_type=slicer.vtkMRMLPlotSeriesNode.PlotTypeLine)
```

jcfr:
@lassoan To further improved @jadh4v original contribution 🙏 , I ended up updating the API and introducing documentation, we would appreciate if you could review & approve 🚀

jcfr (commented):
Thanks for the contribution 🙏 

Once @lassoan has a chance ot review and approve, we will move forward with the integration 👌

Base/Python/slicer/util.py
```diff
@@ -2498,14 +2498,40 @@ def arrayFromTableColumnModified(tableNode, columnName):
...
+      - **Structured (record) array** with named fields. Each field becomes a separate column.
+        If ``columnNames`` is not provided, the field names in the dtype are used.
+      - **1D NumPy array**, which is added as a single column.
+      - **2D NumPy array**, where each column in the array is added as a separate column in the
+        table node (note that the array is transposed).
+      - **List/tuple** of 1D NumPy arrays, which are each added as separate columns.
+
+    :param tableNode: The table node to be updated. If ``None``, a new ``vtkMRMLTableNode`` is
+                      created and added to the scene.
+    :type tableNode: vtkMRMLTableNode or None
```
lassoan, line 2514: shouldn't we use type annotations instead?

Base/Python/slicer/util.py
```diff
@@ -2520,22 +2546,46 @@ def updateTableFromArray(tableNode, narrays, columnNames=None):
...
     elif isinstance(narrays, tuple) or isinstance(narrays, list):
+        # List or tuple of 1D arrays
         ncolumns = narrays
     else:
-        raise ValueError("Expected narrays is a numpy ndarray, or tuple or list of numpy ndarrays, got %s instead." % (str(type(narrays))))
+        # Unrecognized format
+        msg = (
+            "Expected a structured NumPy array (with named fields), "
+            "a 1D/2D NumPy array, or a list/tuple of 1D arrays; got {} instead."
+        )
```
lassoan, line 2570: This is a bit hard to read, because the placeholder `{}` is too much separated from the value that it is replaced with. It would be better to use a named placeholder; but at least add the .format right after the message (not in a separate instruction).

Base/Python/slicer/util.py
```diff
@@ -2520,22 +2546,46 @@ def updateTableFromArray(tableNode, narrays, columnNames=None):
...
         columnNames = [columnNames]
+
+    # For each extracted column, convert to a VTK array and add to the table.
     for columnIndex, ncolumn in enumerate(ncolumns):
-        vcolumn = vtk.util.numpy_support.numpy_to_vtk(num_array=ncolumn.ravel(), deep=True, array_type=vtk.VTK_FLOAT)
+        if ncolumn.dtype == np.bool_:
+            # Convert boolean arrays to uint to ensure compatibility with VTK plotting infrastructure:
+            # vtkContext2D does not support vtkBitArray, and plotting calls that rely on
+            # vtkArrayDownCast<vtkFloatArray>() would fail with a bit array.
+            ncolumn = ncolumn.astype("uint")
```
lassoan, line 2586: While bool does not play nice with plotting, it is quite rare you want to plot a bool function. Instead, bool is useful if you want to have checkboxes in your table.

So, either remove this special conversion (make it the user's responsibility to convert the column type if he wants bools as int) or make it optional (e.g., with an optional parameter like `setBoolAsUint`).

Base/Python/slicer/util.py
```diff
@@ -2498,14 +2498,40 @@ def arrayFromTableColumnModified(tableNode, columnName):
...
-    :raises ValueError: in case of failure
-
-    Values are copied, therefore if the numpy array  is modified after calling this method,
-    values in the table node will not change.
-    All previous content of the table is deleted.
+    """Set values in a table node from a NumPy array or an array-like object (list/tuple of NumPy arrays).
+
+    This function can handle:
+
+      - **Structured (record) array** with named fields. Each field becomes a separate column.
```
lassoan, line 2505: These are great as reference, but it would be really difficult to figure out what you mean exactly. Please add examples here and/or a `test_slicer_util_updateTableFromArray.py` test (that serve as examples) to demonstrate the ways this function can be used.

@jcfr's code snippet could be a good basis for a test, but instead of adding a new helper function `plotTableNode` we should test with `slicer.util.plot`.

lassoan (commented):
Very nice! Added a couple of suggestions inline.

Base/Python/slicer/util.py
jcfr: Good point. This will be done in a follow-up pull request:
* https://github.com/Slicer/Slicer/issues/8226

Base/Python/slicer/util.py
jcfr: Good point, I will add support for boolean array in a follow-up pull request.

Base/Python/slicer/util.py
jcfr: > re: test_slicer_util_updateTableFromArray

I updated it to check that structured array can be read and that column type and name are initialized accordingly.,

> re: we should test with `slicer.util.plot`
Good point.

jcfr: Tests have also been added

jcfr:
@lassoan @pieper When you have a chance, I would appreciate another review on this. It is ready for integration :rocket:

pieper (commented):
LGTM, but I'l let @lassoan have the final word

lassoan (approved):
It all looks good to me, very nice work!

Issue #7707, opened 2024-04-22
Allow scripted modules to declare and lazily install pip requirements


allemangD:
Continuation of #7697; tightly coupled to #7171 and #6913.

After discussion and feedback at the weekly hangout Apr 16 (see https://github.com/Slicer/Slicer/issues/7697#issuecomment-2059271196 and https://github.com/Slicer/Slicer/issues/7697#issuecomment-2059273077) I think my first issue conflated three separate tasks:

- Declare and install dependencies
- Run modules in an isolated subprocess (CLI modules)
- Run modules in an isolated environment (venv, conda, poetry, etc)

I'm interested in tackling only the first of these, and since CLI have questionable value anyway it doesn't make sense to require CLI usage to get the benefit here. Really, what we want is a way for scripted module's logic to depend on pypi packages, and only install those packages when they are needed. Doing this purely in the Python layer allows all scripted modules to benefit.

## Draft implementation and features

I have a Python 3.9 draft implementation of a `LazyImportGroup` context manager, based on the [`LazyLoader`](https://docs.python.org/3/library/importlib.html#importlib.util.LazyLoader) and [`lazy_import` recipe](https://docs.python.org/3/library/importlib.html#implementing-lazy-imports) in the standard library `importlib` documentation.

https://github.com/allemangD/lazyloader - see the doctests for `LazyImportGroup` for more details.

The important principles here are:

- Consolidate. Keep all `import` statements at top-level, all dependencies in `requirements.txt`.
- Lazy. Do not install or import expensive libraries until they are used.
- Safe. Do not import libraries unless they are guaranteed to be available.

These were not mentioned during the discussion last week, but I think they are important:

- Support static analysis.
- Be explicit about when "magic" occurs.

---

Usage requires the extension developer place a `requirements.txt` in a python package, installed via `RESOURCES` argument of the existing CMake macros. The [anchor and path](https://docs.python.org/3/library/importlib.resources.html#importlib.resources.files) for that `requirements.txt` resource are given as an argument to `LazyImportGroup` context manager, wrapping top-level imports.

```py
with LazyImportGroup('lib:requirements.txt'):
  import lib.logic
  import itk
```

These imports are real import statements that are detected by static analysis, and run-time magic only occurs with imports within the context manager. These imports produce proxy modules that defer install and real import until first usage:

```py
def onApplyButton(self):
  lib.logic.do_the_task(...)
  #        ^ attribute access triggers install and import

def onComputeButton(self):
[...]

PR #8181, opened 2025-01-24
Allow scripted modules to declare requirements and constraints, and enforce Slicer Core constraints.

allemangD:
Work toward #7707. A summary:

- [x] Preinstall https://github.com/astral-sh/uv in the Slicer Python environment. This is needed for faster dependency resolution; other features like python installation management are disabled via environment variables.
- [x] The `LazyImportGroup` from the prior revision is renamed to `slicer.deps.GuardedImports`. I still don't love this name.
- [x] Any occurrence of a `GuardedImports` context manager registers its `requirements.txt` as a constraint that applies to all invocations of `pip_install`. `GuardedImports` also _use_ `pip_install` to install missing dependencies.
- [x] Bundled Slicer packages (`scipy`, `numpy`, etc.) also have a constraint which is enforced by all invocations of `pip_install` and `GuardedImports`.

Together, the intent is for extension developers to declare their dependencies in a `requirements.txt`, have clear error messages and logs if any of those dependencies are incompatible with bundled packages or other extensions, and support static analysis via true `import` syntax. See #7707 for more details.

This PR is a draft for four reasons:

- [ ] First, I have manually copied all the builtin requirements into `slicer.deps:core-constraints.txt`. This file must be generated automatically before the PR should be accepted. I see a couple options to do this, either generate the file via `Utilities/Scripts/python_package_updater.py`; or generate that file in CMake. For example see the `${requirements_file}` for `numpy` in https://github.com/Slicer/Slicer/blob/main/SuperBuild/External_python-numpy.cmake#L30-L31. It is unclear how/when this file should be installed - probably not when the superbuild is configured, since we probably don't want disabled packages to be constrained.

- [x] Second, I need to improve reporting to the user. There are some `# todo` throughout the `pip_install` and `pip_uninstall` implementations. These are points where we would inspect the `uv pip install` return code and stdout to assemble and appropriate message for the user. Critically, it should be possible for the user to cancel the installation.

- [x] Third, I want to evaluate just how necessary the dependency on `uv` is. I suspect that it is worth it, but I want to benchmark dependency resolution on a large system, for example resolving different-but-compatible versions of `torch` and its dependencies. I haven't investigated the `torch` dependency graph enough yet to build this benchmark.

- [x] Fourth, developer documentation must be updated to explain where to place the requirements.txt and how to deal with the new error messages that will appear. To be clear, these are not _new_ modes of failure, but they might not be ones that extension developers have yet considered. The error messages and documentation should be clear enough that a developer can quickly deal with the issue. (The brief advice should be: almost always use `~=` version specifiers.)

---

- [x] I'd also like to clean up the commit history a bit before merging, but that's not as important as the functionality issues above.

I welcome comments on the general approach or advice on how to generate that `core-constraints.txt` file.


allemangD:
After writing the note about "halted import" for f10e7d7 I need to verify:

If module A uses a guarded import on eg. `itk`, and module B _does not_ (say it still uses manual `pip_install` calls), it is possible for module A to halt module B's `import itk`. This seems undesirable, but I think it does indicate a true bug.

allemangD:
cc @jcfr, @lassoan, @pieper. Readthedocs for this PR are now comprehensive. https://slicer--8181.org.readthedocs.build/en/8181/developer_guide/python_packages.html

I'd appreciate any feedback or unsupported edge-cases you foresee on this. The goal is to primarily support things in discrete layers of abstraction:

1. Plain `pip_install`. Most flexible, but a bit of friction.

```python
pip_install('itk')

pip_install('itk', interactive=False)

pip_install(requirements=FileIdentifier('Sample Module', 'requirements.txt'))

register_constraints(FileIdentifier('Sample Module', 'constraints.txt'))

import itk
```

2. `Requirements` groups with manual resolution. Less friction, but not _quite_ as flexible.

```python
requirements = Requirements('requirements.txt', name="Sample Module")

try:
  requirements.resolve()
except InstallAbortedError:
  ...
  
import itk
```

3. `Requirements` groups with manual resolution, and proxy modules for static analysis. IMO the static analysis makes this worth it.

```python
from __future__ import annotations

with Requirements('requirements.txt') as requirements:
  import itk
  
try:
  requirements.resolve()
except InstallAbortedError:
  ...
  
im: itk.Image = itk.imread(...)
```

4. Fully automatic installation and proxy modules. Mostly seamless, as long as the user accepts installation.

```python
from __future__ import annotations

with Requirements('requirements.txt'):
  import itk
  
# use normally
```

---

@jcfr RE my messages yesterday: I did remove import halting mentioned in https://github.com/Slicer/Slicer/pull/8181#issuecomment-2623039227. It causes more problems than it solves.

Now that I think the interface is hardened, I plan to translate my ad-hoc testing modules into comprehensive unit tests. I also should take a look at the API documentation to be sure it's not out of date.

I need some help on automatically generating `core-constraints.txt`. Ideally, only the packages _which are installed to the current build_ will be added to this file. I do not think it is appropriate to do this from `Utilities/Scripts/python_package_updater.py` since it is decoupled from the build and can't know which components are actually in use.

I think it is possible to do this from CMake, but my attempts to do this did not work. Each external python package needs to append `${requirements_file}` to `core-constraints.txt` _as part of its install step_, not as part of the superbuild configure step. I haven't figured out how to get that timing.

Base/Python/slicer/packaging/requirements.py
```diff
@@ -0,0 +1,301 @@
...
+
+        from slicer.packaging import Requirements
+
+        with Requirements('package:requirements.txt'):
+            import dependency
+
+        dependency.function()
+        #         ^ Resolution occurs here.
+
+    See `Automatic Installation <https://slicer.readthedocs.io/en/latest/developer_guide/python_packaging.html#manual-installation>`_.
```
allemangD, line 64: This obviously doesn't work right in the CI readthedocs. I assume there is some way to get reST to do this properly?

Base/Python/slicer/packaging/installer.py
```diff
@@ -0,0 +1,400 @@
...
+        self.name = name
+        self.identifier = identifier
+        self.caller = caller
+
+    def as_file(self) -> AbstractContextManager[Path]:
+        package_name, _, path = self.identifier.rpartition(":")
+
+        # An omitted package_name is a relative import for convenience.
+        if not package_name:
+            package_name = "."
```
allemangD, line 86: Remove this. Do not do relative import.

If the package is not provided, do a path lookup on `sys.path`. Perhaps there is something to help with this in `importlib.util`? But it is not handled by `importlib.resources`.

jcfr, line 86:
```suggestion
        if not package_name:  # No package specified, look in site-packages and Python path
            search_paths = site.getsitepackages() + [site.getusersitepackages()] + sys.path
            for search_dir in search_paths:
                candidate_path = Path(search_dir) / path
                if candidate_path.exists():
                    return contextlib.nullcontext(candidate_path)

            raise FileNotFoundError(f"File '{path}' not found in site-packages or Python path.")
```

mhalle:
I looked at the latest version of uv for capabilities that might help here, and uv workspace seems well suited. Here's Claude's take on it: https://gist.github.com/mhalle/c2e752467d960a123f42ea459c09f73e

PR #8186, opened 2025-01-26
ENH: Define a window icon for all windows 

jamesobutler:
https://doc.qt.io/qt-5/qapplication.html#windowIcon-prop

Defining the QApplication window icon sets the default window icon. Slicer currently sets the QMainWindow window icon which is primarily used for windows that are children of the main window. Before this commit the QApplication window icon was undefined so no icon was used for a widget with no parent.

| `main`  | This PR |
|---------|---------|
|{9738F05D-BFF0-4C2C-A512-300FEE842AE3}|{68C46A8D-C1AC-43EF-B2EF-A7DBA0E95F39}|

### DIFF
--- a/Base/QTGUI/qSlicerApplication.cxx
+++ b/Base/QTGUI/qSlicerApplication.cxx
@@ -360,6 +360,12 @@ void qSlicerApplicationPrivate::init()
 
   this->initStyle();
 
+  QIcon icon;
+  icon.addFile(":/Icons/Medium/DesktopIcon.png");
+  icon.addFile(":/Icons/Large/DesktopIcon.png");
+  icon.addFile(":/Icons/XLarge/DesktopIcon.png");
+  q->setWindowIcon(icon);
+
   this->ToolTipTrapper = new ctkToolTipTrapper(q);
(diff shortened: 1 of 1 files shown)


Base/QTCore/qSlicerCoreApplication.cxx
```diff
@@ -278,6 +279,11 @@ void qSlicerCoreApplicationPrivate::init()
 
   QCoreApplication::setOrganizationDomain(Slicer_ORGANIZATION_DOMAIN);
   QCoreApplication::setOrganizationName(Slicer_ORGANIZATION_NAME);
+  QIcon icon;
```
jcfr, line 282: Since `QIcon` is provided by `Qt GUI`[^1], what do you think of moving the code into `qSlicerApplicationPrivate::init()` ?

[^1]: https://doc.qt.io/qt-5/qicon.html

jamesobutler: Thanks for the review. Yes we can move the code there. I've pushed the corresponding update.

jcfr (approved):

PR #8195, opened 2025-01-29
ENH: Add `ctkColorPickerButton` support with `QColor` in `parameterNodeWrapper`

sjh26:
This commit introduces:
- A new GuiConnector (`ctkColorPickerButtonToQColorConnector`) that binds `ctkColorPickerButton` to a `QColor` property in the parameter node.
- A new serializer (`QColorSerializer`) enabling QColor to be saved and restored from the parameter node.
- An `ObservedQColor` subclass that automatically updates the parameter node whenever a “set*” method modifies the color value.

part of #7308


jcfr (approved):

It works great :100:

I only tweaked the style, added a few docstrings and also added a test (inspired from the existing ones)

PR #8202, opened 2025-01-31
COMP: Update external project to support externally built TBB libraries

jcfr:
Allow using pre-built Threading Building Blocks (TBB) libraries by specifying:
- `TBB_DIR`: Path to CMake config location
- `TBB_BIN_DIR`: Path to runtime libraries (`.dll` on Windows)
- `TBB_LIB_DIR`: Path to linker libraries (`.lib` on Windows, shared libs on Unix)

On Unix systems, `TBB_BIN_DIR` and `TBB_LIB_DIR` should point to the same directory.

### DIFF
--- a/SuperBuild/External_tbb.cmake
+++ b/SuperBuild/External_tbb.cmake
@@ -8,8 +8,26 @@ set(${proj}_DEPENDENCIES "")
 ExternalProject_Include_Dependencies(${proj} PROJECT_VAR proj DEPENDS_VAR ${proj}_DEPENDENCIES)
 
 if(Slicer_USE_SYSTEM_${proj})
+  # When adding support for finding TBB on the system, make sure to set TBB_BIN_DIR and TBB_LIB_DIR
   message(FATAL_ERROR "Enabling Slicer_USE_SYSTEM_${proj} is not supported!")
 endif()
+
+# Sanity checks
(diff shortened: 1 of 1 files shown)


jamesobutler (commented):
Is TBB difficult to build from source? Is there a reason why we’ve been using the pre-built binaries? I assume the pre-built binaries were all x86_64 builds which is why this need to specify externally built TBB is to allow specifying binaries for arm64 support.

jcfr:
> Is TBB difficult to build from source?

I was able to sort this out on Ubuntu 22.04 (aarch64). See https://gist.github.com/jcfr/487f5d846bc86e374969be5565c6d95e

Ultimately, I think we should be able to build it our-self, this give us the flexibility to figure this out.

---------

I tested that building on Linux (after deleting the `tbb-*` directories) worked without issues on Linux.

On macOS and Windows, I confirmed that re-configuring the top-level project worked as expected after checkout this branch.

jamesobutler (approved):
Ok. Yes in a future PR let’s switch to building oneTBB from source for all platforms or at least for the community supported architectures where oneTBB does not provide pre-built binaries.

jcfr:
Re: build from source

Agreed. I will tackle this while working on adding macOS arm64.

PR #8213, opened 2025-02-08
BUG: Fix crash in vtkSlicerApplicationLogic destructor

lassoan:
Fix application logic destruction and two other minor (almost just style) fixes.

---

After threading rework in https://github.com/Slicer/Slicer/commit/47da17d4c90f4b2a0dff3034b5305cc1d864c31f, `vtkSlicerApplicationLogicTest1` test started to fail due to crash in the destructor.
The problem was that the networking threads were deleted before waiting for them to finish.

Fixed by calling `TerminateProcessingThread()` method that joins all created threads (instead of just joining the ProcessingThread as it was implemented in the destructor before).

### DIFF
--- a/Base/Logic/vtkSlicerApplicationLogic.cxx
+++ b/Base/Logic/vtkSlicerApplicationLogic.cxx
@@ -84,19 +84,8 @@ vtkSlicerApplicationLogic::vtkSlicerApplicationLogic()
 //----------------------------------------------------------------------------
 vtkSlicerApplicationLogic::~vtkSlicerApplicationLogic()
 {
-  // Note that we can not kill a thread safely. So we wait
-  // for the thread to finish.  We need to signal the thread that we
-  // want to terminate
-  if (this->ProcessingThread.joinable())
-  {
-    // Signal the processingThread that we are terminating.
-    this->ProcessingThreadActiveLock.lock();
-    this->ProcessingThreadActive = false;
-    this->ProcessingThreadActiveLock.unlock();
-
-    // Wait for the thread to finish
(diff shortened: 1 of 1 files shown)


Base/Logic/vtkSlicerApplicationLogic.cxx
```diff
@@ -98,6 +98,15 @@ vtkSlicerApplicationLogic::~vtkSlicerApplicationLogic()
...
 
+  for (auto& thread : this->NetworkingThreads)
+  {
+    if (thread.joinable())
+    {
+      thread.join();
+    }
+  }
+  this->NetworkingThreads.clear();
+
```
jcfr, line 109:
```suggestion
  this->TerminateProcessingThread();
```

I think should also move the code above into `TerminateProcessingThread`

jcfr: For reference the original code in `TerminateProcessingThread()` and the one in the destructor `~vtkSlicerApplicationLogic` were both introduced in the same commit aef3c715c5 (`BUG: CommandLineModules need to place a request to the main thread to reload data and display results. ENH: Moved processing threading control to vtkSlicerApplicationLogic. Moved most of the interthread communication to vtkSlicerApplicationLogic.  Interthread communication for Warnings/Errors/Information still in vtkSlicerApplication`, 2006-12-12)

lassoan: I've updated the code to use `this->TerminateProcessingThread();`, please review and approve if looks good.

jcfr (approved):

PR #8215, opened 2025-02-09
BUG: Avoid additional error message in DICOMReaders.py

jcfr:
When SampleData.downloadFromURL failed then there was an extra error logged about failing to clean up the database directory, because the directory has not even been created. Added a check so that we only remove the directory if it has been created.


lassoan (approved):
Looks good.

PR #8216, opened 2025-02-09
BUG: Fix scene loading warning message

jcfr:
When the user was warned about potential scene loading issues, the message was formatted incorrectly: the scene file name was specified as argument but the argument was not used.

Since the scene file name is added to the message at higher level, we can simply remove the unneeded argument.

### DIFF
--- a/Modules/Loadable/Data/qSlicerSceneReader.cxx
+++ b/Modules/Loadable/Data/qSlicerSceneReader.cxx
@@ -145,7 +145,7 @@ bool qSlicerSceneReader::load(const qSlicerIO::IOProperties& properties)
     }
     if (!sceneVersionWarningMessages.isEmpty())
     {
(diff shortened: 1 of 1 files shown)


Modules/Loadable/Data/qSlicerSceneReader.cxx
```diff
@@ -145,7 +145,7 @@ bool qSlicerSceneReader::load(const qSlicerIO::IOProperties& properties)
     }
     if (!sceneVersionWarningMessages.isEmpty())
     {
-      sceneVersionWarningMessages.push_front(tr("The scene may not load correctly.").arg(file));
+      sceneVersionWarningMessages.push_back(tr("The scene may not load correctly."));
```
jcfr, line 148:
```suggestion
      sceneVersionWarningMessages << tr("The scene may not load correctly.");
```

lassoan (approved):
Looks good to me.

PR #8232, opened 2025-02-18
ENH: Consolidate SliceLogic calls to SetInterpolateTexture

jcfr:
This pull request removes redundant `vtkMRMLModelDisplayNode::SetInterpolateTexture` call in
`vtkMRMLSliceLogic::ProcessMRMLLogicsEvents()`, by handled the remaining call in `vtkMRMLSliceLogic::UpdatePipeline()` with a more comprehensive check.

### Rationale

After reviewing the logic, checking only for `GetImageDataConnectionUVW()` would be sufficient because:

- The model slice node (`vtkMRMLModelDisplayNode`) is only associated with 3D views (see use of `SetViewNodeIDs` depicted below).
- `vtkMRMLModelDisplayableManager`, responsible for displaying the slice model node, derives from `vtkMRMLAbstractThreeDViewDisplayableManager`.
- `vtkMRMLModelSliceDisplayableManager` is responsible for displaying model intersections on slice planes and does not handle the model associated with the slice viewer.

### Future Optimization

This PR focuses on removing redundancy. Further optimizations should be handled separately.

### Reference: ModelSliceNode Association with 3D Views

https://github.com/Slicer/Slicer/blob/9195b3f94ab121acb9250c21068dea012d33d7e1/Libs/MRML/Logic/vtkMRMLSliceLogic.cxx#L483-L489

### DIFF
--- a/Libs/MRML/Logic/vtkMRMLSliceLogic.cxx
+++ b/Libs/MRML/Logic/vtkMRMLSliceLogic.cxx
@@ -617,19 +617,6 @@ void vtkMRMLSliceLogic::ProcessMRMLLogicsEvents()
     /// \tbd Ideally it should not be fired if the output polydata is not
     /// modified.
     plane->Modified();
-
-    vtkMRMLModelDisplayNode *modelDisplayNode = this->SliceModelNode->GetModelDisplayNode();
-    if ( modelDisplayNode )
-    {
-      if (this->LabelLayer && this->LabelLayer->GetImageDataConnectionUVW())
-      {
-        modelDisplayNode->SetInterpolateTexture(0);
-      }
-      else
-      {
-        modelDisplayNode->SetInterpolateTexture(1);
-      }
-    }
   }
 
   // This is called when a slice layer is modified, so pass it on
@@ -1155,7 +1142,7 @@ void vtkMRMLSliceLogic::UpdatePipeline()
       {
         displayNode->SetTextureImageDataConnection(this->ExtractModelTexture->GetOutputPort());
       }
-        if ( this->LabelLayer && this->LabelLayer->GetImageDataConnection())
+        if (this->LabelLayer && (this->LabelLayer->GetImageDataConnection() || this->LabelLayer->GetImageDataConnectionUVW()))
         {
           displayNode->SetInterpolateTexture(0);
         }


Libs/MRML/Logic/vtkMRMLSliceLogic.cxx
```diff
@@ -617,19 +617,6 @@ void vtkMRMLSliceLogic::ProcessMRMLLogicsEvents()
     /// \tbd Ideally it should not be fired if the output polydata is not
     /// modified.
     plane->Modified();
-
-    vtkMRMLModelDisplayNode *modelDisplayNode = this->SliceModelNode->GetModelDisplayNode();
-    if ( modelDisplayNode )
-    {
-      if (this->LabelLayer && this->LabelLayer->GetImageDataConnectionUVW())
```
lassoan, line 624: In `vtkMRMLSliceLogic::UpdatePipeline()`, the check is `if ( this->LabelLayer && this->LabelLayer->GetImageDataConnection()) ` (note the `UVW`). Which one is correct? (or both are needed?)

lassoan (commented):
It is probably OK but I added a comment, please check

Libs/MRML/Logic/vtkMRMLSliceLogic.cxx
jcfr: For reference:

https://github.com/Slicer/Slicer/blob/9195b3f94ab121acb9250c21068dea012d33d7e1/Libs/MRML/Logic/vtkMRMLSliceLogic.cxx#L1157-L1164

jcfr: The original code adding both check was introduced in https://github.com/Slicer/Slicer/commit/171321d5c49292c52c951f7937de2e9da607a5bf#diff-d5498fd4b04ced370870a7176ba82f0d624ddcd269f72d27577fddf416f871b3

jcfr: Good catch identifying the difference related to the use of `UVW`.

After looking into this, checking only for `GetImageDataConnectionUVW()` would be sufficient as:
* the model slice node (`vtkMRMLModelDisplayNode`) is only associated with the 3D views (see below).
* `vtkMRMLModelDisplayableManager` is responsible to dislay the slice model node but derives from `vtkMRMLAbstractThreeDViewDisplayableManager`
* `vtkMRMLModelSliceDisplayableManager` display intersection of models onto the slice plane (and by definition if does nothing for the the model associated with slice viewer)

That said this is an optimization, that should be done in a different pull request. I will then revisit this patch to remove the duplicated code by checking for both.

### Association of ModelSliceNode with 3D views

https://github.com/Slicer/Slicer/blob/9195b3f94ab121acb9250c21068dea012d33d7e1/Libs/MRML/Logic/vtkMRMLSliceLogic.cxx#L483-L489

jamesobutler (approved):

Issue #7488, opened 2023-12-18
Update VTK from  `slicer-v9.2.20230607-1ff325c54` to a recent version


jcfr:
## Is your feature request related to a problem? Please describe.

The version currently integrated in Slicer corresponds to the branch `Slicer/VTK@slicer-v9.2.20230607-1ff325c54-2`[^1] and is based of kitware/VTK@1ff325c54ad9f214752491222977ff5508d162b4 from June 2023.

While selected fixes and improvement have been integrated through the pull requests listed below, we are missing out on additional fixes and improvements as well as increasing the maintenance burden as time goes.

The last VTK update was done through this pull request:
* https://github.com/Slicer/Slicer/pull/7010

[^1]: https://github.com/Slicer/VTK/tree/slicer-v9.2.20230607-1ff325c54-2

## Describe the solution you'd like

We should create a new branch names like `slicer-v9.3.2024MMDD-abcdefghi` with only a few Slicer specific changes.

:warning: While working on https://github.com/Slicer/Slicer/pull/7487, a performance regression related to `vtkRender::GetZ()` was identified, we should make sure to review its status when working on this update:
* https://github.com/Slicer/Slicer/pull/7487#issuecomment-1859573883
* https://gitlab.kitware.com/vtk/vtk/-/merge_requests/10391#note_1459483
* https://github.com/Slicer/VTK/pull/46#issuecomment-1859596578

## Describe alternatives you've considered

NA

## Additional context

Here are the pull requests backporting changes (from the most recent to the oldest) since the last update through https://github.com/Slicer/Slicer/pull/7010:
* https://github.com/Slicer/Slicer/pull/7487
* https://github.com/Slicer/Slicer/pull/7453
* https://github.com/Slicer/Slicer/pull/7305
* https://github.com/Slicer/Slicer/pull/7286
* https://github.com/Slicer/Slicer/pull/7274
* https://github.com/Slicer/Slicer/pull/7227
* https://github.com/Slicer/Slicer/pull/7195
* https://github.com/Slicer/Slicer/pull/7193
* https://github.com/Slicer/Slicer/pull/7190
* https://github.com/Slicer/Slicer/pull/7140
* https://github.com/Slicer/Slicer/pull/7104
* https://github.com/Slicer/Slicer/pull/7080

PR #8238, opened 2025-02-19
ENH: Update VTK from 9.2.20230607 to 9.4.2

AlexyPellegrini:
These are the changes I identified to successfully build Slicer against VTK 9.4.2.

List of Slicer specific changes (removed all backported commits):
```
git shortlog 6b6b89ee5..38a1f52c5
Jean-Christophe Fillion-Robin (5):
      [Slicer] COMP: Remove custom FindOpenGL to ensure OpenGL_GL_PREFERENCE is considered
      [Slicer] BUG: Update vtkPythonUtil to disable relative import causing crash for Slicer modules
      [Slicer] COMP: Remove FindTBB and expect VTK to be configured with TBB_DIR
      [SlicerSMTK] COMP: Add support for specifying vtk-m library install components
      [SlicerSMTK] vtkm: Use github.com/KitwareMedical/vtk-m: Update diy install rules, support TBB_DIR
```

# Update notes

- Remove RenderingOpenGL module references since this module as been definitively removed from VTK
  - This also removes need of `vtkSlicerConfigure.h` in some files, which is needed for SlicerLibrary
- Update `vtkNumberToString` usages to use new API
- `vtkMRMLMarkupsNode::SetNthControlPointPositionMissing` has been renamed to `vtkMRMLMarkupsNode::SetControlPointPositionMissing`
  - This is needed because VTK wrapping system has been modified in VTK 9.4, and recognise `SetNthControlPointPositionMissing ` as a specific pattern `SetNthX(int i, X x);`, but `SetNthControlPointPositionMissing` prototype does not match.
- Remove VTK>=9.0 references, since they should not be needed anymore

------

Related issues:
* #7488
* https://github.com/Slicer/Slicer/issues/8379

Related pull requests:
* https://github.com/Slicer/Slicer/pull/8239
* https://github.com/commontk/CTK/pull/1229
* https://github.com/Slicer/Slicer/pull/8372
* https://github.com/Slicer/Slicer/pull/8380

-------

> [!NOTE]
> Changes originally integrated through this pull request have been extracted and contributed as independent pull requests:
> * https://github.com/Slicer/Slicer/pull/8244
> * https://github.com/Slicer/Slicer/pull/8249
> * https://github.com/Slicer/Slicer/pull/8251


jcfr:
> Temporary update of vtkAddon to support VTK 9.4.1

Corresponding changes being integrated through the following pull request:

* https://github.com/Slicer/Slicer/pull/8239

AlexyPellegrini:
> I built this branch on Windows (CMake 3.31.4 and Visual Studio 2022 17.12.4) and there was no build errors. However upon running the application from the build tree or after packaging and installing, lots of Slicer modules were not successfully loaded. Seems primarily due to `import vtk` not being successful in the python console.

Yep it worked on my machine because I used a workaround.
I fixed the issue and also rebased the branch on main, now it should be fine everywhere!

Libs/vtkSegmentationCore/vtkSegmentationConverter.cxx
```diff
@@ -143,7 +143,7 @@ std::string vtkSegmentationConverter::SerializeImageGeometry(vtkMatrix4x4* geome
   {
     for (int j=0; j<4; j++)
     {
-      geometryStream << numberToString.Convert(geometryMatrix->GetElement(i, j)) << SERIALIZED_GEOMETRY_SEPARATOR;
+      geometryStream << numberToString(geometryMatrix->GetElement(i, j)) << SERIALIZED_GEOMETRY_SEPARATOR;
```
jcfr, line 146: Turns out that using `Convert` is the recommended way moving forward, I will then revert this in topic called `remove-support-for-vtk-9.1-or-older` I will soon propose for review.

Libs/MRML/Core/vtkCacheManager.cxx
```diff
@@ -641,13 +641,9 @@ int vtkCacheManager::ClearCache()
   {
     this->MarkNodesBeforeDeletingDataFromCache ( this->RemoteCacheDirectory.c_str() );
     vtksys::SystemTools::RemoveADirectory ( this->RemoteCacheDirectory.c_str() );
-  }
-#if (VTK_MAJOR_VERSION >= 9 && VTK_MINOR_VERSION >= 0 && VTK_BUILD_VERSION >= 20210806)
+    }
```
jcfr, line 644:
```suggestion
  }
```

This will be applied in upcoming `remove-support-for-vtk-9.1-or-older` topic contributed as an independent pull request.

Libs/MRML/Core/vtkCacheManager.cxx
```diff
@@ -641,13 +641,9 @@ int vtkCacheManager::ClearCache()
...
   if ( !vtksys::SystemTools::MakeDirectory ( this->RemoteCacheDirectory.c_str() ) )
-#else
-  if ( vtksys::SystemTools::MakeDirectory ( this->RemoteCacheDirectory.c_str() ) == false )
-#endif
-  {
+    {
```
jcfr, line 646:
```suggestion
  {
```

This will be applied in upcoming `remove-support-for-vtk-9.1-or-older` topic contributed as an independent pull request.

Libs/vtkSegmentationCore/vtkSegmentationConverter.cxx
AlexyPellegrini: Indeed, I don't remember why I changed this. Let's revert this :)

jcfr:
---- 

> [!IMPORTANT]
> This pull request should be rebased & integrated *only* after the pull requests referenced below are themselves reviewed & integrated.

> [!NOTE]
> Changes originally integrated through this pull request have been extracted and contributed as independent pull requests:
> * https://github.com/Slicer/Slicer/pull/8244

Libs/vtkITK/vtkITKArchetypeDiffusionTensorImageReaderFile.cxx
```diff
@@ -96,7 +96,8 @@ void vtkITKExecuteDataFromFileDiffusionTensor3D(
   for ( it.GoToBegin(); !it.IsAtEnd() ; ++it )
   {
     const itk::Index<3u> index = it.GetIndex();
-    vtkIdType position = data->FindPoint(index[0], index[1], index[2]);
+    double arr[] = {static_cast<double>(index[0]), static_cast<double>(index[1]), static_cast<double>(index[2])};
+    vtkIdType position = data->FindPoint(arr);
```
jcfr, line 100: Since I couldn't identify the rational for this change, I did not cherry pick this change as part of https://github.com/Slicer/Slicer/pull/8244

SuperBuild/External_VTK.cmake
```diff
@@ -264,17 +264,13 @@ if((NOT DEFINED VTK_DIR OR NOT DEFINED VTK_SOURCE_DIR) AND NOT Slicer_USE_SYSTEM
...
-    endif()
     if(UNIX)
       set(${proj}_PYTHONPATH_LAUNCHER_INSTALLED
-        <APPLAUNCHER_SETTINGS_DIR>/../${_library_install_subdir}/python${Slicer_REQUIRED_PYTHON_VERSION_DOT}/site-packages
+        <APPLAUNCHER_SETTINGS_DIR>/../lib/python${Slicer_REQUIRED_PYTHON_VERSION_DOT}/site-packages
         )
     else()
       set(${proj}_PYTHONPATH_LAUNCHER_INSTALLED
-        <APPLAUNCHER_SETTINGS_DIR>/../${_library_install_subdir}/Lib/site-packages
+        <APPLAUNCHER_SETTINGS_DIR>/../lib/Lib/site-packages
```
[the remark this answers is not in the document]
AlexyPellegrini: Added both locations

Libs/vtkITK/vtkITKArchetypeDiffusionTensorImageReaderFile.cxx
```diff
@@ -96,7 +96,7 @@ void vtkITKExecuteDataFromFileDiffusionTensor3D(
   for ( it.GoToBegin(); !it.IsAtEnd() ; ++it )
   {
     const itk::Index<3u> index = it.GetIndex();
-    vtkIdType position = data->FindPoint(index[0], index[1], index[2]);
+    vtkIdType position = data->vtkDataSet::FindPoint(index[0], index[1], index[2]);
```
jcfr, line 99: The removal of `virtual vtkIdType FindPoint(double x, double y, double z)` from `vtkImageData` seems to have been introduced in 92fc5b2670 (`vtkImageData: Use Implicit structures and check visibility`, 2023-05-18) by @spyridon97

The following pull request is intended to address this. See https://gitlab.kitware.com/vtk/vtk/-/merge_requests/11935

Libs/MRML/Core/vtkMRMLMarkupsNode.h
```diff
@@ -468,7 +468,7 @@ class  VTK_MRML_EXPORT vtkMRMLMarkupsNode : public vtkMRMLDisplayableNode
   void UnsetNthControlPointPosition(int pointIndex);
 
   /// Set control point status to ignored.
-  void SetNthControlPointPositionMissing(int pointIndex);
+  void SetControlPointPositionMissing(int pointIndex);
```
jcfr, line 471:
```suggestion
/// Disable wrapping to prevent matching as a `SET_NTH` function.
VTK_WRAPEXCLUDE
void SetNthControlPointPositionMissing(int pointIndex);
```

jcfr: To ensure existing code works as expected, let's use the `VTK_WRAPEXCLUDE` hint.

Libs/vtkITK/vtkITKArchetypeDiffusionTensorImageReaderFile.cxx
jcfr: Corresponding changes have been backported to https://github.com/Slicer/VTK/commits/slicer-v9.4.1-2024-12-26-6b6b89ee5/

jcfr (commented):
Branch along with recent backports has been pushed at https://github.com/Slicer/VTK/commits/slicer-v9.4.1-2024-12-26-6b6b89ee5/

AlexyPellegrini:
Thanks you for pushing the VTK branch on the fork! There is still one issue, we need to push VTK tags on Slicer/VTK because the version is not detected correctly and this make VTK_VERSION macros unreliable!

I rebased again on master.

jcfr:
>  we need to push VTK tags on Slicer/VTK

Tags `v9.4.0` and `v9.4.1` have been pushed.

AlexyPellegrini:
Everything seems fine on Windows and Ubuntu 22.04 now!

jamesobutler (commented):
My build and package on Windows was successful.

@AlexyPellegrini Do you observe the following weird blue object with the "z" letter in the bottom left of the 3D view?

| `main` | This PR|
|--------|---------|
|image|image|

`ENH: Exclude MarkupsNode::SetNthControlPointPositionMissing from wrappping` -> I would also probably utilize the `COMP` prefix ([source](https://github.com/Slicer/Slicer/blob/main/CONTRIBUTING.md#how-to-write-commit-messages-)) for this fix.

AlexyPellegrini:
We have been observing this axes actors, or at least what looks like a [vtkAxesActor](https://vtk.org/doc/nightly/html/classvtkAxesActor.html).
This issue seems related to the VTK bump indeed.
Maybe there always as been an axes actor that now shows up because something changed in VTK rendering logic.

jamesobutler:
Seems like https://gitlab.kitware.com/vtk/vtk/-/issues/19592 (`9.4: Removing vtkOrientationMarkerWidget and adding a new one leaves blue artifact`) is what I'm seeing. It appears [MR-11874](https://gitlab.kitware.com/vtk/vtk/-/merge_requests/11874) fixed it which was integrated as https://gitlab.kitware.com/vtk/vtk/-/commit/1c617ee5848d44e6d4d8aaa182df55c2c9548af3 (aka https://github.com/Kitware/VTK/commit/1c617ee5848d44e6d4d8aaa182df55c2c9548af3) in January. The current VTK branch we are using is based on 9.4.1 from late December, so we will need to backport this fix. This fix will also be included in VTK 9.4.2 whenever that is released.

jamesobutler:
With the recent backport the vtk orientation marker issue has been resolved.
image

I have update the VTK fork commit hash here and rebased against latest `main` for testing with latest fixes. I have been running the Slicer automated tests locally which have had some failures.

jamesobutler:
One of the last remaining things is a failure of `test_updateTableFromArray` where it appears that the workaround recently added in https://github.com/Slicer/Slicer/commit/d48d7b62f6a38a9422eaa84331c0d5232815d388 is no longer working with the updated VTK version in this PR.

Libs/MRML/Core/Testing/vtkMRMLTableNodeTest1.cxx
```diff
@@ -169,7 +169,7 @@ int vtkMRMLTableNodeTest1(int , char * [] )
 
   checkDefaultArrayValue< vtkUnsignedIntArray, unsigned int >(node3, "unsigned int", "3", 3);
   // this is a bit odd (since an unsigned value accepts a negative value), but this is vtkVariant behavior:
-  checkDefaultArrayValue< vtkUnsignedIntArray, unsigned int >(node3, "unsigned int", "-5", -5);
+  // checkDefaultArrayValue< vtkUnsignedIntArray, unsigned int >(node3, "unsigned int", "-5", -5);
```
jamesobutler, line 172: Unclear if related to the other table issue mentioned in https://github.com/Slicer/Slicer/pull/8238#issuecomment-2715885023, but this commented out line was resulting in vtkDebugLeaks being observed on exit. I assume it might've been exiting abnormally causing those leaks. @lassoan Any ideas on a real solution here? It appears you had already made an code comment about this vtkVariant behavior being odd. Maybe it was changed as part of this VTK update.

lassoan: Is any error (something like `checkDefaultArrayValue failed for...`) logged before the leaks are reported? In that case, the "leaks" can be ignored (they are just a side-effect of the quick exit triggered by the test failure).

lassoan: Great, thank you!

AlexyPellegrini:
Thanks James for taking care of this!

Is there anything else to fix before merging?

jcfr:
Thanks @jamesobutler and @AlexyPellegrini 

Few brief notes:

* Since `VTK_PROPEXCLUDE` is not defined  for older VTK version, how about defining it to nothing in the `vtkMRMLMarkupsNode.h` header if VTK version is older ? 
* Similarly, how about conditionally setting the eggs directory and launcher settings ?

Doing so would allow us to support building against VTK 9.2 and 9.3 cleanly.

AlexyPellegrini:
I agree that `PROP_EXCLUDE` can be defined as an empty macro for VTK < 9.4.0.
Eggs directory shouldn't cause any issue, only one will exist at a given time, but it wouldn't be much work to do it conditionally either, so why not.

AlexyPellegrini:
I backported the fix for test_updateTableFromArray and rebased on main.

jamesobutler:
I have pushed the fix for the `vtkMRMLTableNodeTest1` test again, though I haven't looked through the VTK commits to see which VTK version caused the change in the expected test results. I would expect that someone building with an older VTK version such as 9.2 and maybe 9.3 would observe `vtkMRMLTableNodeTest1` failing even though it passes when using the default VTK 9.4.1 version that this PR switches to.

AlexyPellegrini:
My bad, I forgot to pull/reset before force pushing again. Thank you for pushing the fix again!

jamesobutler:
^ Copilot is now a possible reviewer for PRs, so I was curious to see if it would be able to interpret anything here.

AlexyPellegrini:
Is there anything else to do before merging?

@jcfr

lassoan:
It looks like the Python console cannot get to the specific type (vtkMRMLScalarVolumeNode) but it gets stuck at the base type (vtkMRMLNode). Something related to VTK hierarchy files.

Similar issue used to happen with MRML classes registered in extensions (VMTK, SlicerRT, etc.). We had to call `from ...Python import ...Node` - but this is no longer needed. Can you try if it helps if you run `from MRMLCorePython import vtkMRMLScalarVolumeNode` in the Python console right after startup?

@jcfr

jamesobutler:
{DF938DF0-E204-4241-852F-A1B8C67FBF12}

jcfr:
Once my local build completes, I will review and follow-up.

jcfr:
> re: auto-completion

Starting with VTK 9.4, the Python API has been updated. See https://www.kitware.com/vtk-9-4-a-step-closer-to-the-ways-of-python/

This explains what we observe below:

image


image

jcfr:
I will look at the details to see if we could restore the availability of the VTK methods during auto-completion (in addition of the property based ones), this should then allow us to gradually transition.

jamesobutler:
It would be fine with me if auto completion only supported the new style properties rather than the setter/getters also being available in the autocomplete. The old setter/getter calls still function, they just aren’t in the autocomplete. This would also help with encouraging the new pythonic style while the older style still being functional.

lassoan:
Breaking ties with the C++ API is a big decision and each developer should be given enough time to adapt. To minimize disruption, it could be an application setting to hide the C++ API in auto-complete (for now, show it by default).

jcfr:
VTK completion was previously failing due to an issue where errors were not properly handled when retrieving attributes from VTK classes during auto-completion model generation. This has been resolved in the following pull request:  
* https://github.com/commontk/CTK/pull/1229

Improving the robustness of the completion mechanism also led to VTK warnings being printed in the Python console. These warnings are now excluded in the following pull request:  
* https://github.com/Slicer/Slicer/pull/8372

jcfr:
To provide additional background on the error encountered while building the completion model: the failure to access the `ijk_to_ras_matrix` property occurs because the corresponding getter requires an argument. As a result, the VTK Python wrapper is unable to generate the necessary code to expose it as a readable attribute:

```cpp
void GetIJKToRASMatrix(vtkMatrix4x4* mat);
void SetIJKToRASMatrix(vtkMatrix4x4* mat);
```

In `vtkMRMLVolumeNodePython.cxx`, the wrapper defines a `set` function, but no `get`, making the property effectively write-only:

```cpp
static PyGetSetDef PyvtkMRMLVolumeNode_GetSets[] = {
  {
    /*name=*/pystr("ijk_to_ras_matrix"),
    /*get=*/nullptr,
    /*set=*/[](PyObject* self, PyObject* value, void*) -> int {
      if (PyTuple_Check(value)) {
        auto result = PyvtkMRMLVolumeNode_SetIJKToRASMatrix(self, value);
        return (result == nullptr) ? -1 : 0;
      } else {
        auto args = PyTuple_Pack(1, value);
        auto result = PyvtkMRMLVolumeNode_SetIJKToRASMatrix(self, args);
        Py_DECREF(args);
        return (result == nullptr) ? -1 : 0;
      }
    },
    /*doc=*/pystr("write-only, Calls SetIJKToRASMatrix\n"),
    /*closure=*/nullptr,
  },
  // ...
};
```

To surface these errors during testing, we temporarily replaced internal `PyErr_Clear` calls with `PythonQt::self()->handleError()`, as described in https://github.com/commontk/CTK/pull/1229/files#r2040115294. This revealed a set of attributes that are not properly exposed as readable properties:

```
AttributeError: attribute 'add_to_scene_no_modify' of 'MRMLCorePython.vtkMRMLNode' objects is not readable
AttributeError: attribute 'and_observe_display_node_id' of 'MRMLCorePython.vtkMRMLDisplayableNode' objects is not readable
AttributeError: attribute 'and_observe_image_data' of 'MRMLCorePython.vtkMRMLVolumeNode' objects is not readable
AttributeError: attribute 'and_observe_storage_node_id' of 'MRMLCorePython.vtkMRMLStorableNode' objects is not readable
AttributeError: attribute 'i_to_ras_direction' of 'MRMLCorePython.vtkMRMLVolumeNode' objects is not readable
AttributeError: attribute 'ijk_to_ras_direction_matrix' of 'MRMLCorePython.vtkMRMLVolumeNode' objects is not readable
AttributeError: attribute 'ijk_to_ras_directions' of 'MRMLCorePython.vtkMRMLVolumeNode' objects is not readable
AttributeError: attribute 'ijk_to_ras_matrix' of 'MRMLCorePython.vtkMRMLVolumeNode' objects is not readable
AttributeError: attribute 'j_to_ras_direction' of 'MRMLCorePython.vtkMRMLVolumeNode' objects is not readable
AttributeError: attribute 'k_to_ras_direction' of 'MRMLCorePython.vtkMRMLVolumeNode' objects is not readable
AttributeError: attribute 'memkind_directory' of 'vtkmodules.vtkCommonCore.vtkObjectBase' objects is not readable
AttributeError: attribute 'ras_to_ijk_matrix' of 'MRMLCorePython.vtkMRMLVolumeNode' objects is not readable
AttributeError: attribute 'singleton_off' of 'MRMLCorePython.vtkMRMLNode' objects is not readable
AttributeError: attribute 'singleton_on' of 'MRMLCorePython.vtkMRMLNode' objects is not readable
AttributeError: attribute 'active_attribute_location_from_string' of 'MRMLCorePython.vtkMRMLDisplayNode' objects is not readable
AttributeError: attribute 'add_to_scene_no_modify' of 'MRMLCorePython.vtkMRMLNode' objects is not readable
AttributeError: attribute 'and_observe_clip_node_id' of 'MRMLCorePython.vtkMRMLDisplayNode' objects is not readable
AttributeError: attribute 'and_observe_color_node_id' of 'MRMLCorePython.vtkMRMLDisplayNode' objects is not readable
AttributeError: attribute 'displayable_only_in_view' of 'MRMLCorePython.vtkMRMLDisplayNode' objects is not readable
AttributeError: attribute 'memkind_directory' of 'vtkmodules.vtkCommonCore.vtkObjectBase' objects is not readable
AttributeError: attribute 'scalar_range_flag_from_string' of 'MRMLCorePython.vtkMRMLDisplayNode' objects is not readable
AttributeError: attribute 'singleton_off' of 'MRMLCorePython.vtkMRMLNode' objects is not readable
AttributeError: attribute 'singleton_on' of 'MRMLCorePython.vtkMRMLNode' objects is not readable
AttributeError: attribute 'threshold' of 'MRMLCorePython.vtkMRMLScalarVolumeDisplayNode' objects is not readable
AttributeError: attribute 'window_level' of 'MRMLCorePython.vtkMRMLScalarVolumeDisplayNode' objects is not readable
AttributeError: attribute 'window_level_from_preset' of 'MRMLCorePython.vtkMRMLScalarVolumeDisplayNode' objects is not readable
...
```

jcfr (approved):

AlexyPellegrini:
Thank you for fixing the last issues and merging this!

jamesobutler:
I did an incremental build which was successful. I also ran tests and all the same tests were passing as with the Slicer `main` branch. This was all completed on the Windows platform.

PR #8242, opened 2025-02-19
BUG: Do not load DWMRI volumes as sequences

### LINKED ISSUE #8240: DWI volumes being loaded as sequences
## Summary

.nrrd files containing diffusion weighted MRI data are being interpreted as sequences by default when they should be loaded as volumes.

## Steps to reproduce

Get a dwi nrrd file like this one:

https://github.com/SlicerDMRI/SlicerDMRI/releases/download/v1.0.0/Dti_tutorial_data.zip

In the add data dialog the Sequences option is selected by default and the user must manually pick Volume in order to load correctly for DWI processing.

[This code](https://github.com/Slicer/Slicer/blob/ae95058f068deea17f13595560a341bb590b19f4/Modules/Loadable/Sequences/qSlicerSequencesReader.cxx#L101) should be updated to check for `modality:=DWMRI` in the header and lower the confidence accordingly.

## Environment
- Slicer version: Slicer-5.8
- Operating system: all platforms

pieper:
Improve qSlicerSequencesReader heuristics to know that dwmri modality is not typically loaded as a sequence by lowering the confidence for files of this type.

Also fixes code to now read the entire nrrd header so that other keys won't be missed in files with large headers (dwmri files typically have over 1000 characters before the modality tag).

Fixes #8240

### DIFF
--- a/Modules/Loadable/Volumes/qSlicerVolumesReader.cxx
+++ b/Modules/Loadable/Volumes/qSlicerVolumesReader.cxx
@@ -25,6 +25,8 @@
 // Qt includes
 #include <QDebug>
 #include <QFileInfo>
+#include <QRegularExpression>
+#include <QTextStream>
 
 // Slicer includes
 #include "qSlicerVolumesIOOptionsWidget.h"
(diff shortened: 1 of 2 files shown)


Modules/Loadable/Sequences/qSlicerSequencesReader.cxx
```diff
@@ -118,44 +118,76 @@ double qSlicerSequencesReader::canLoadFileConfidence(const QString& fileName)con
...
-
-        // Supported 4D NRRD files contain "dimension: 4" line.
-        // 2D+t and 3D+color+t files are not yet supported.
-        QRegularExpression dimensionRe("dimension:([^\\n]+)");
-        QRegularExpressionMatch dimensionMatch = dimensionRe.match(line);
-        if (dimensionMatch.hasMatch())
+        QString nrrdHeader = in.readLine();
+        if (!nrrdHeader.startsWith("NRRD"))
+        {
+          confidence = 0.3; // not a valid nrrd file
```
jcfr, line 124: @lassoan For sake of consistency, should not this be `0.4` ?

lassoan:
Is there a reason you added this logic to the sequence reader? The sequence reader should not need to know anything about dwi. Instead, the dwi reader should recognize the dwi volume and return higher confidence.

pieper:
> Is there a reason you added this logic to the sequence reader? The sequence reader should not need to know anything about dwi. Instead, the dwi reader should recognize the dwi volume and return higher confidence.

This reader already has heuristics about RGB and displacement images not being sequences, so it makes sense that it also express lower confidence with other types of nrrd files that are known not to be sequences.

Also we don't have a file reader specific to dwmri, or other volume types, currently.  Instead the mrml volume storage node infrastructure is what sorts out the contents of the nrrd files.  It could make sense to update that architecture but that seems like overkill just to fix this one regression.

lassoan:
Not a huge issue to make the sequence reader DWI-aware, but it would be just as simple to make the general image reader (that currently handles DWI image reading) DWI-aware. It should work just by adding a `qSlicerVolumesReader::canLoadFileConfidence` method like this:

```
double qSlicerVolumesReader::canLoadFileConfidence(const QString& fileName)const
{
  double confidence = Superclass::canLoadFileConfidence(fileName);
  // Confidence for .nrrd and .nhdr file is 0.55 (5 characters in the file extension matched)
  if (confidence > 0)
  {
    // Inspect the content to recognize DWI volumes.
    QString upperCaseFileName = fileName.toUpper();
    if (upperCaseFileName.endsWith("NRRD") || upperCaseFileName.endsWith("NHDR"))
    {
      QFile file(fileName);
      if (file.open(QIODevice::ReadOnly | QIODevice::Text))
      {
        QTextStream in(&file);
        QString line = in.readLine();
        if (line.startsWith("NRRD"))
        {
          // The nrrd header is separated by the data by a blank line, so read everything up to there
          // since diffusion scans can have a long list of gradients and modality can be
          // near the end of the header
          QRegularExpression modalityRe("modality:([^\\n]+)");
          while (!in.atEnd())
          {
            line = in.readLine();
            if (line.trimmed() == "")
            {
              break;
            }
            QRegularExpressionMatch modalityMatch = modalityRe.match(line);
            if (modalityMatch.hasMatch())
            {
              QString modalityStr = modalityMatch.captured(1);
              if (modalityStr.contains("dwmri", Qt::CaseInsensitive))
              {
                // This is a DWMRI image, we are confident that it is not just a general image sequence.
                // Therefore we set higher confidence than the generic sequence reader's confidence of 0.6.
                confidence = 0.7;
                break;
              }
            }
          }
        }
      }
    }
  }
  return confidence;
}
```

pieper:
Good suggestion @lassoan, thanks.  Can you have another look?

If we are doing a patch release it would be nice to cherry pick this fix since it makes SlicerDMRI workflows simpler.

Issue #8241, opened 2025-02-19
Unexpected behavior of save/export option for long file names


muratmaga:
The recent introduction of truncating long node names during saving is inconvenient for those of us who have to rely on actual filenames for their workflows.

Modifying filenames to be saved without user consent is quite unexpected and problematic in terms of data management. While I am assume this is done to avoid the situations of hitting the path limitations with windows, user needs to be explicitly consulted and given an explanation about this behavior at the minimum.

lassoan:
Filenames are generated from the node name, which may be unexpected but not renaming filename when a node is renamed caused more problems. Filename is further modified if the node name contains characters that are unsafe in some file systems (e.g., slash or colon). In recent Slicer versions, the filename is even further modified if it is too long.

Since the filename generation/modification is not new, I don't think it is necessary to explicitly consult the user (it would be too much noise to ask before every node rename). Instead, we could make the maximum file name length configurable.

For robust data management: If files are loaded and saved without using the GUI (save data or export data dialogs) then filenames are not modified. For data processing workflows it may make sense to override the default save dialog.

lassoan:
I've submitted a [pull request](https://github.com/Slicer/Slicer/pull/8245) that:
- sets the default maximum filename length to 1000 on Linux/macOS and 25 on Windows systems
- makes the value editable in application settings (in case someone on linux/macOS wants better Windows compatibility; or someone on Windows wants to relax the length limitations)

Would this be a good fix for you @muratmaga?

PR #8245, opened 2025-02-20
ENH: Make maximum file length configurable

lassoan:
Users who do not need compatibility with Windows systems can use longer file paths. They may prefer allowing longer file paths to keep node names matching file names.

see #8241


cpinter:
LGTM!

Sunderlandkyl (approved):
Thanks @lassoan, LGTM!

jamesobutler:
Linking back to https://github.com/Slicer/Slicer/issues/7901#issuecomment-2451032796 where there was original comments about handling long filepaths the same across platforms. To note is that a scene saved on macOS or Linux may not be able to be loaded on Windows due to long file names. As well as long filepaths could be problematic when saving to removable drives formatted with FAT file system.

lassoan:
See my comment here: https://github.com/Slicer/Slicer/issues/8241#issuecomment-2671875179 - Some people prefer Windows compatibility, some people prefer non-shortened filenames, so we have to make this user-configurable.

I'm not sure what the default value should be. Should be short everywhere? Considering the complaints in the past few weeks, maybe it is better to allow much longer filenames on linux&macos by default.

jamesobutler:
As it relates to saving to disk it probably should be left up to the OS. The user just needs to provided clear text when saving does fail due to long filepath that it was indeed due to long filepath issues and what they can do to resolve it.

In the Slicer UI, I have seen a user complain about the slice viewers fitting on their screen and it was seemingly due to very very long node names which were being shown in the slice controller. So the UI has to consistently do some elide patterns to fit the text in widgets because the widgets can’t be infinitely wide.

lassoan:
The main difficulty is that the filename is automatically updated based on the node name when you open the "Save data" window, . Therefore, we cannot just let the saving fail and ask the user to intervene manually, because he would need to do that each time he saves the scene, for each node that has a long name.

@jamesobutler 
Do you have any other idea to improve this PR?
Do you have any specific concerns (e.g., not preserving filename, too long filenames, etc)

jamesobutler:
> Therefore, we cannot just let the saving fail and ask the user to intervene manually, because he would need to do that each time he saves the scene, for each node that has a long name.

Sorry I don’t have any other ideas other than to do this.

lassoan:
@Sunderlandkyl I've made some more cleanups (hardcode max filename length when saving MRB, while use custom filename length in save and export dialogs; simplified API by always requiring setting of the extension name). Please have a look at it again and approve if looks good. Thank you!

lassoan:
@Sunderlandkyl or @jamesobutler you have already reviewed an earlier version, would you be able to have a look at this latest version and approve if looks good? We would like to get this in the 5.8.1 patch release by the end of the week. Thank you!

Sunderlandkyl (approved):
Sorry, I missed the previous message. New changes LGTM!

PR #8247, opened 2025-02-20
ENH: Add segment visibility toggle option

fedorov:
This PR proposes addition of a context menu item for the segmentations widget to toggle visibility of the selected segments. This feature is very convenient while reviewing segmentations with the large number of segments, when a subset of segments need to be shown/hidden without affecting visibility of other segments.


lassoan (commented):
Thanks @fedorov this is very useful. Toggle visibility/Show/Hide was available in the subject hierarchy tree, but indeed not in the segment list widget.

I would make the "toggle" option the first (above "show only") because often you want to toggle visibility multiple times, while you would not want to do "show only" repeatedly (if the segment is already shown then the action has no effect). I would also add "visbility" to the action name to make it clear what gets toggled.

image

For simplicity, I've added these proposed changes as a commit on this branch, @fedorov please have a look.

fedorov:
@lassoan your changes look great to me, thank you! Ready to go as far as I am concerned.

PR #8251, opened 2025-02-24
COMP: Workaround VTK 9.4 wrapping issue for SetNthControlPointPositionMissing

jcfr:
This commit addresses an issue introduced in VTK 9.4 (kitware/VTK@5672af1f), where `SetNthControlPointPositionMissing` is incorrectly recognized as a `SET_NTH` function by the VTK wrapping system, leading to build failures due to signature mismatches.

### Changes and Workarounds

1. Exclude `SetNthControlPointPositionMissing` from wrapping:
   - Marked with `VTK_WRAPEXCLUDE` to prevent automatic recognition as a `SET_NTH` function.
   - Ensures compatibility until the VTK wrapping infrastructure is improved.
   - Related issue: [VTK Issue #19627](https://gitlab.kitware.com/vtk/vtk/-/issues/19627).

2. Introduce `WrappedSetNthControlPointPositionMissing` as a wrapper method:
   - Serves as an intermediate function that can be replaced in generated Python bindings.
   - Ensures that `SetNthControlPointPositionMissing` remains callable from Python.

3. Patch the generated Python wrapper (`vtkMRMLMarkupsNodePython.cxx`) at build time:
   - Added a CMake custom command to modify the generated wrapper **after VTK wrapping**.
   - Uses a **Python script** to replace occurrences of `WrappedSetNthControlPointPositionMissing`
     with `SetNthControlPointPositionMissing` in `vtkMRMLMarkupsNodePython.cxx`.
   - Ensures that Python bindings remain functional without modifying VTK itself.

4. Ensure the patching process is reliably triggered:
   - `PatchMRMLCorePythonWrappers` is added as a dependency of `MRMLCorePython`.
   - Uses `copy_if_different` to avoid unnecessary rebuilds.

This workaround allows the build to succeed without modifying upstream VTK while ensuring the method remains accessible in Python.

### DIFF
--- a/Libs/MRML/Core/vtkMRMLMarkupsNode.h
+++ b/Libs/MRML/Core/vtkMRMLMarkupsNode.h
@@ -468,7 +468,33 @@ public:
   void UnsetNthControlPointPosition(int pointIndex);
 
   /// Set control point status to ignored.
+  /// Excluded from wrapping to prevent unintended `SET_NTH` recognition.
+  /// This behavior was introduced in VTK 9.4 (kitware/VTK@5672af1f) and requires
+  /// specific function signatures. Until the VTK wrapping infrastructure is improved,
+  /// this method should remain excluded. See https://gitlab.kitware.com/vtk/vtk/-/issues/19627.
+  VTK_WRAPEXCLUDE
   void SetNthControlPointPositionMissing(int pointIndex);
+
+  /// \internal
+  /// Workaround for a VTK 9.4 wrapping limitation that prevents
+  /// `SetNthControlPointPositionMissing` from being correctly recognized.
+  ///
+  /// This method is intentionally introduced as an alias for
+  /// `SetNthControlPointPositionMissing` to be patched in the generated
+  /// Python wrapper (`vtkMRMLMarkupsNodePython.cxx`).
+  ///
+  /// The custom target `PatchMRMLCorePythonWrappers` (defined in `CMakeLists.txt`)
+  /// ensures that occurrences of `WrappedSetNthControlPointPositionMissing`
+  /// in the generated wrapper are replaced with `SetNthControlPointPositionMissing`,
+  /// maintaining Python compatibility.
+  ///
+  /// This function should NOT be called directly in C++ code and will be removed
+  /// once the VTK wrapping infrastructure is improved.
+  ///
+  /// See https://gitlab.kitware.com/vtk/vtk/-/issues/19627 for more details.
+  ///
+  /// \sa SetNthControlPointPositionMissing()
+  void WrappedSetNthControlPointPositionMissing(int pointIndex);
 
   /// Set control point status to preview
   void ResetNthControlPointPosition(int n);
--- a/Libs/MRML/Core/CMakeLists.txt
+++ b/Libs/MRML/Core/CMakeLists.txt
@@ -415,6 +415,43 @@ if(VTK_WRAP_PYTHON)
       set_target_properties(${lib_name}Hierarchy PROPERTIES FOLDER ${${PROJECT_NAME}_FOLDER})
     endif()
   endif()
+
+  # --------------------------------------------------------------------------
+  # Workaround issue https://gitlab.kitware.com/vtk/vtk/-/issues/19627
+  # --------------------------------------------------------------------------
+  set(_generated_wrapper_cxx "vtkMRMLMarkupsNodePython.cxx")
+  set(_patched_wrapper_cxx "${_generated_wrapper_cxx}.patched")
+  set(_patched_done "${CMAKE_CURRENT_BINARY_DIR}/${_generated_wrapper_cxx}.patched.done")
+
+  # Ensure patching happens when the wrapper is regenerated
+  add_custom_command(
+    OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/${_patched_wrapper_cxx}
+    DEPENDS ${CMAKE_CURRENT_BINARY_DIR}/${_generated_wrapper_cxx}
+    COMMAND ${PYTHON_EXECUTABLE} -c
+      "import sys; in_file, out_file, search_text, replace_text = sys.argv[1:]; content = open(in_file).read().replace(search_text, replace_text); open(out_file, 'w').write(content)"
+      ${CMAKE_CURRENT_BINARY_DIR}/${_generated_wrapper_cxx}
+      ${CMAKE_CURRENT_BINARY_DIR}/${_patched_wrapper_cxx}
+      "WrappedSetNthControlPointPositionMissing" "SetNthControlPointPositionMissing"
+    COMMENT "Patching generated Python wrapper source ${_generated_wrapper_cxx}"
+    VERBATIM
+    )
+
+  # Copy the patched file back to the original location, ensuring CMake tracks completion
+  add_custom_command(
+    OUTPUT ${_patched_done}
+    DEPENDS ${CMAKE_CURRENT_BINARY_DIR}/${_patched_wrapper_cxx}
+    COMMAND ${CMAKE_COMMAND} -E copy_if_different
+      ${CMAKE_CURRENT_BINARY_DIR}/${_patched_wrapper_cxx}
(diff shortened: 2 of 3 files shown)


jcfr:
@lassoan @AlexyPellegrini  This is an attempt to address the issue raised in https://github.com/Slicer/Slicer/pull/8238#issue-2862837579 without having to rename the function frmo `SetNthControlPointPositionMissing` to `SetControlPointPositionMissing`.

Considering a search on GitHub[^1] didn't reveal any use from Python, disabling the wrapping seem reasonable.

@dgobbi Is there way to only disable the automatic recognition as a `SET_NTH` function while maintaining the "regular" wrapping of the function ?


[^1]: https://github.com/search?q=SetNthControlPointPositionMissing+AND+NOT++is%3Afork&type=code

jamesobutler (approved):

lassoan (changes requested):
The method is important and the implementation of this method is fairly complex, so simply making it unavailable from Python is not a good solution. We need to make it available in Python. Maybe the simplest is to add an alias - another method that calls this method that is compatible with VTK.

It would be useful if someone can check how and why this method is incompatible with VTK and then either deprecate the method in Slicer to avoid similar issues in the future; or fix the VTK Python wrapper (make it more robust or specific).

jcfr:
> useful if someone can check how and why this method is incompatible with VTK

We identified the issue. It is due to the automatic wrapping described in the commit message and the comment associated with the function.

jcfr:
Having functions like the following should keep the wrapping happy, i think:

```
void SetNthControlPointPositionMissing(int n, bool)
bool GetNthControlPointPositionMissing(int n);
```

lassoan:
@dgobbi The `void SetNthControlPointPositionMissing(int pointIndex)` method sets N-th point position to `Missing`, following the convention in VTK that there are convenience methods to set enumerated values. For example:

```
  vtkSetMacro(BlendMode, int);
  void SetBlendModeToComposite()
  void SetBlendModeToMaximumIntensity()
```

If the wrapper works for `Set(propertyname)To(enumvalue)(...)` then it should also work for `SetNth...` methods, too. Two potential implementation options:
- A. Recognize the `SetNth(propertyname)To(enumvalue)(int index)` pattern. The presence of `To` would be a hint that the wrapper should accept just an index as argument.
- B. The wrapper could always accept `SetNth...` methods with just an index argument.

dgobbi:
@jcfr there is no `WRAPEXCLUDE` hint specific to the new "property" wrapping.  But there should be, so I'll add one.

It looks like the property-wrapping code for `SetNth` was never implemented, and incorrect wrapping code is generated instead.  You probably already figured out that much.  I'll have more time to look things over this evening so that I can get an answer to @lassoan.

jcfr:
Thanks @dgobbi for the update :pray: 

In the short term, supporting a wrapper hint to disable the new "property" wrapping will work well.

Then, in the future, improving the wrapping as @lassoan suggested would be ideal :rocket: 

`#stepbystep`

lassoan:
Thank you @dgobbi for your quick action on this!

jcfr:
> I will create an issue in the VTK tracker to document this.

See https://gitlab.kitware.com/vtk/vtk/-/issues/19627

Libs/MRML/Core/vtkMRMLMarkupsNode.h
```diff
@@ -468,6 +468,9 @@ class  VTK_MRML_EXPORT vtkMRMLMarkupsNode : public vtkMRMLDisplayableNode
   void UnsetNthControlPointPosition(int pointIndex);
 
   /// Set control point status to ignored.
+  /// This method is excluded from wrapping to prevent automatic recognition as a `SET_NTH` function,
+  /// a behavior introduced in VTK 9.4 (kitware/VTK@5672af1f), which requires specific function signatures.
```
jcfr, line 472:
```suggestion
  /// Excluded from wrapping to prevent unintended `SET_NTH` recognition.
  /// This behavior was introduced in VTK 9.4 (kitware/VTK@5672af1f) and requires
  /// specific function signatures. Until the VTK wrapping infrastructure is improved,
  /// this method should remain excluded. See https://gitlab.kitware.com/vtk/vtk/-/issues/19627.
```

jcfr:
@lassoan  Now that we have a path forward, could you approve the pull request so that we can support building against VTK 9.4 ?

Commit message for `Squash & Merge`

```
COMP: Exclude vtkMRMLMarkupsNode method from wrapping for VTK 9.4 compatibility

This commit marks `SetNthControlPointPositionMissing` with `VTK_WRAPEXCLUDE`
to prevent it from being automatically recognized as a `SET_NTH` function.
This behavior was introduced in VTK 9.4 (kitware/VTK@5672af1f) and requires
specific function signatures. Until the VTK wrapping infrastructure is improved,
this method should remain excluded.
See https://gitlab.kitware.com/vtk/vtk/-/issues/19627.

Without this exclusion, the build fails due to a signature mismatch, as
the function does not conform to the expected `SET_NTH` signatures:
- `void SetNthValue(int, type)`
- `void SetNthValue(int, type[])`

Co-authored-by: Alexy Pellegrini <alexy.pellegrini@kitware.com>
```

pieper:
Does that mean the method will be missing from the python bindings?  Is there anyway we can add a manual workaround, like giving it a different name and then defining original name either in python or in a patch to the generated bindings?

jcfr:
> Does that mean the method will be missing from the python bindings? 

With the current approach, it would be missing only until the issue is addressed in VTK. In the interim, I thought this would be okay for now as the method is not used from Python.

I was considering adding the method `_SetNthControlPointPositionMissing` as described below:

```
  /// \internal
  /// This function should NOT be used directly. It serves as a workaround for a limitation
  /// in VTK wrapping introduced in VTK 9.4 (kitware/VTK@5672af1f) that prevents
  /// `SetNthControlPointPositionMissing` from being correctly recognized.
  ///
  /// It is intended **only** for use in `Base/Python/mrml.py`, where it is explicitly
  /// associated with all `vtkMRMLMarkupsNode`-derived classes.
  ///
  /// This method will be removed once the VTK wrapping infrastructure is improved.
  /// See https://gitlab.kitware.com/vtk/vtk/-/issues/19627.
  ///
  /// \sa SetNthControlPointPositionMissing()
  void _SetNthControlPointPositionMissing(int pointIndex);
```

... but it is not possible to add new attributes to VTK classes. The following error is reported ...

```python
import MRMLCorePython

# Attach the new method to all vtkMRMLMarkupsNode-derived classes
for cls_name in dir(MRMLCorePython):
    cls = getattr(MRMLCorePython, cls_name)
    if isinstance(cls, type) and issubclass(cls, MRMLCorePython.vtkMRMLMarkupsNode):
        # Only add if the method doesn't already exist
        if not hasattr(cls, "SetNthControlPointPositionMissing"):
            setattr(
                cls,
                "SetControlPointPositionMissing",
                MRMLCorePython.vtkMRMLMarkupsNode._SetNthControlPointPositionMissing,
            )
```

```
  File "<string>", line 10, in <module>
TypeError: can't set attributes of built-in/extension type 'MRMLCorePython.vtkMRMLMarkupsAngleNode'
```

jcfr:
@AlexyPellegrini Would appreciate if you could test on Windows :pray:

jcfr:
> How can we be sure this will still be possible to trick the wrapping this way?

Addressing https://gitlab.kitware.com/vtk/vtk/-/issues/19627 is the path forward.

The corresponding approach should be applicable using the vtk modules system, I suggest address this when we work on integrating https://github.com/Slicer/Slicer/pull/7491

> Build failed due to python3 not being found.

Thanks for testing.  `${PYTHON_EXECUTABLE}` should be used instead. Once I confirm, I will force push :rocket:

jamesobutler:
Here is the related opened MR https://gitlab.kitware.com/vtk/vtk/-/merge_requests/11941 to address the property wrapping.

jcfr:
Closing. Corresponding changes associated with `MR-11941`[^1] have been backported to https://github.com/Slicer/VTK/commits/slicer-v9.4.1-2024-12-26-6b6b89ee5/

[^1]: https://gitlab.kitware.com/vtk/vtk/-/merge_requests/11941/commits

PR #8252, opened 2025-02-24
BUG: Fix py_UtilTest on Windows

lassoan:
py_UtilTest's test_updateTableFromArray test failed on Windows:

-----

FAIL: test_updateTableFromArray (UtilTest.UtilTestTest) 
---------------------------------------------------------------------- 
  File "C:/D/P/S-0-build/Slicer-build/lib/Slicer-5.9/qt-scripted-modules/UtilTest.py", line 285, in test_updateTableFromArray
    self.assertEqual(tableNode3.GetColumnType("height"), vtk.VTK_LONG_LONG)
AssertionError: 6 != 16

-----

The root cause is that "int" type is mapped to np.int64 on linux and macOS but on Windows it is mapped to np.int32.

Fixed the test by making the behaviour consistent on all platforms, by using np.int32 as column data type.


jcfr (approved):

jcfr:
Thanks for addressing this 🙏

PR #8233, opened 2025-02-18
BUG: Ensure blend pipeline is updated when setting operation Add or S…

jcfr:
…ubtract

PR #8256, opened 2025-02-25
BUG: Avoid unnecessary error message in UpdateAddSubOperation

jcfr:
Fixes an issue introduced in 26fd93d49e ("BUG: Ensure blend pipeline is updated when setting operation Add or Subtract", 2025-02-17) through https://github.com/Slicer/Slicer/pull/8233.

The function `vtkMRMLSliceLogic::UpdateAddSubOperation` may be called with any compositing mode, including unsupported ones. Instead of logging an error, silently return when encountering an unsupported mode, as it does not require special handling.

### DIFF
--- a/Libs/MRML/Logic/vtkMRMLSliceLogic.cxx
+++ b/Libs/MRML/Logic/vtkMRMLSliceLogic.cxx
@@ -963,10 +963,8 @@ bool vtkMRMLSliceLogic::UpdateAddSubOperation(vtkImageMathematics* addSubMath, i
 {
   if (compositing != vtkMRMLSliceCompositeNode::Add && compositing != vtkMRMLSliceCompositeNode::Subtract)
   {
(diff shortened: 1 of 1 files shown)


lassoan (commented):
It is not obvious why it is normal to get an unsupported compositing mode. Could you add a little bit more information to the comment to explain?

Also, if we simply return without explicitly setting an operation in the filter then the behavior becomes kind of unpredictable (the current behavior depends not just on the current state but on what the previous state was). Could we specify a default compositing mode that we revert to in case the input `compositing` is not supported?

jcfr:
Thanks for the review.

> if we simply return without explicitly setting an operation in the filter then the behavior becomes kind of unpredictable

Since my originally intent was to keep the `UpdatePipeline` code flow straightforward, I went with this the naive approach of simply removing the error message.

By doing so, the function would return false in two scenarios: (1) compositing mode set to a value different from `Add` or `Subtract` and (2) compositing mode set to `Add` or `Subtract` .. to remove ambiguity, I am suggesting a different approach in https://github.com/Slicer/Slicer/pull/8257

PR #8257, opened 2025-02-25
BUG: Avoid unnecessary error message in UpdateAddSubOperation

jcfr:
Fixes an issue introduced in 26fd93d ("BUG: Ensure blend pipeline is updated when setting operation Add or Subtract", 2025-02-17).

`vtkMRMLSliceLogic::UpdateAddSubOperation` may be called with any compositing mode, including unsupported ones, leading to unnecessary error messages. This update ensures the function is only called for supported modes (`Add` or `Subtract`), preventing unintended log errors.

### DIFF
--- a/Libs/MRML/Logic/vtkMRMLSliceLogic.cxx
+++ b/Libs/MRML/Logic/vtkMRMLSliceLogic.cxx
@@ -1122,14 +1122,18 @@ void vtkMRMLSliceLogic::UpdatePipeline()
       backgroundImagePortUVW, foregroundImagePortUVW, this->SliceCompositeNode->GetForegroundOpacity(),
       labelImagePortUVW, this->SliceCompositeNode->GetLabelOpacity());
 
-    // Update add/subtract operations in the pipeline
-    if (vtkMRMLSliceLogic::UpdateAddSubOperation(this->Pipeline->AddSubMath.GetPointer(), this->SliceCompositeNode->GetCompositing()))
+    if (this->SliceCompositeNode->GetCompositing() == vtkMRMLSliceCompositeNode::Add
+        || this->SliceCompositeNode->GetCompositing() == vtkMRMLSliceCompositeNode::Subtract)
     {
-      modified = 1;
-    }
-    if (vtkMRMLSliceLogic::UpdateAddSubOperation(this->PipelineUVW->AddSubMath.GetPointer(), this->SliceCompositeNode->GetCompositing()))
-    {
-      modified = 1;
+      // Update add/subtract operations in the pipeline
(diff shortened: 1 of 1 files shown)


lassoan (commented):
This will be good. I just have one more question: when the compositing mode is neither `Add` nor `Subtract` then the `this->Pipeline->AddSubMath` filter is not used at all (and that's why it does not matter what mode is set on it)?

jcfr:
> filter is not used at all (and that's why it does not matter what mode is set on it)?

Indeed, this is depicted below. Setting operation to the `AddSubMath` filter only matter when compositing is either `Add` or `Subtract`:

https://github.com/Slicer/Slicer/blob/26fd93d49e4ce957be4bdc3aa3ed38522254a141/Libs/MRML/Logic/vtkMRMLSliceLogic.cxx#L83-L109

jcfr:
For sake of correctness, we could even move the calls to `UpdateFractions` under the same branch:

```cpp
    if (this->SliceCompositeNode->GetCompositing() == vtkMRMLSliceCompositeNode::Add
        || this->SliceCompositeNode->GetCompositing() == vtkMRMLSliceCompositeNode::Subtract)
    {
      // Update add/subtract operations in the pipeline
      if (vtkMRMLSliceLogic::UpdateAddSubOperation(this->Pipeline->AddSubMath.GetPointer(), this->SliceCompositeNode->GetCompositing()))
      {
        modified = 1;
      }
      if (vtkMRMLSliceLogic::UpdateAddSubOperation(this->PipelineUVW->AddSubMath.GetPointer(), this->SliceCompositeNode->GetCompositing()))
      {
        modified = 1;
      }

      // Update opacity fractions for additional layers in add/subtract blending mode
      if (vtkMRMLSliceLogic::UpdateFractions(this->Pipeline->ForegroundFractionMath.GetPointer(), this->SliceCompositeNode->GetForegroundOpacity()))
      {
        modified = 1;
      }
      if (vtkMRMLSliceLogic::UpdateFractions(this->PipelineUVW->ForegroundFractionMath.GetPointer(), this->SliceCompositeNode->GetForegroundOpacity()))
      {
        modified = 1;
      }
    }
```

jcfr:
> For sake of correctness, we could even move the calls to `UpdateFractions` under the same branch

That said ... I think allowing to set the fraction at anytime is sensible. Then switching the compositing more to Add or Subtract would ensure the fraction set beforehand is considered.

If compositing mode is different from `Add` or `Subtract`, we could effectively set the fraction by also return false as the active pipeline would not effectively be modified ...

lassoan (approved):
OK, thanks for the explanation.

PR #8271, opened 2025-02-25
[5.8] BUG: Avoid additional error message in DICOMReaders.py

jcfr:
Backport from #8215

---

When SampleData.downloadFromURL failed then there was an extra error logged about failing to clean up the database directory, because the directory has not even been created. Added a check so that we only remove the directory if it has been created.

(cherry picked from commit d94ad217a6375e5ed551edf124ad959f5b6becb8)


lassoan (approved):
Thank you, looks good.

PR #8272, opened 2025-02-25
[5.8] BUG: Fix scene loading warning message

jcfr:
Backport from #8216

---

When the user was warned about potential scene loading issues, the message was formatted incorrectly: the scene file name was specified as argument but the argument was not used.

Since the scene file name is added to the message at higher level, we can simply remove the unneeded argument.

(cherry picked from commit 19690712673e96bb91d8eb60f735a5e0e2dffe9c)


lassoan (approved):
Thank you, looks good.

PR #8274, opened 2025-02-25
[5.8] BUG: Do not load DWMRI volumes as sequences

### LINKED ISSUE #8240: DWI volumes being loaded as sequences

jcfr:
Backport from #8242

---

Improve qSlicerSequencesReader heuristics to know that dwmri modality is not typically loaded as a sequence by lowering the confidence for files of this type.

Also fixes code to now read the entire nrrd header so that other keys won't be missed in files with large headers (dwmri files typically have over 1000 characters before the modality tag).

Fixes #8240

(cherry picked from commit ef756a160661be3152831b6744d0dd51e5785fe6)

### DIFF
--- a/Modules/Loadable/Volumes/qSlicerVolumesReader.cxx
+++ b/Modules/Loadable/Volumes/qSlicerVolumesReader.cxx
@@ -25,6 +25,8 @@
 // Qt includes
 #include <QDebug>
 #include <QFileInfo>
+#include <QRegularExpression>
+#include <QTextStream>
 
 // Slicer includes
 #include "qSlicerVolumesIOOptionsWidget.h"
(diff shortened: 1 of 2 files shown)


pieper (approved):
LGTM 👍

PR #8275, opened 2025-02-25
[5.8] BUG: Fix crash in SystemTools::RemoveADirectory

jcfr:
Backport from #8250

---

vtkMRMLScene::GetTemporaryBundleDirectory could return a string that contained a "\0" character. C string manipulation functions mostly ignore what is after the "\0" character.

Since this string was used as a directory name, it was passed to kwSys::SystemTools::RemoveADirectory method, which got into an infinite recursion and crashed with stack overflow, because the appended "/" + filename effectively has not been appended.

This caused several tests to randomly fail, such as vtkMRMLSequenceStorageNodeTest1.

The regression was introduced in commit 47c791c77fe4091a18fb894f9f2d19f5c50ad903 (BUG: Fix loading and saving .mrb files with trailing space in filename) on 2025-01-11.

(cherry picked from commit f24bada8cc0ee2f707425689545a5273f03b90d9)


lassoan (approved):
Thank you, looks good.

PR #8291, opened 2025-03-03
BUG: Fix qSlicerTransformsModuleWidgetTest

lassoan:
The Transforms module widget now uses a Subject Hierarchy tree widget, which requires initialization of the subject hierarchy module. This was missing from the test, therefore the Transforms module widget test failed. Fixed by adding subject hierarchy module initialization to the test.

### DIFF
--- a/Modules/Loadable/Transforms/Testing/Cxx/qSlicerTransformsModuleWidgetTest.cxx
+++ b/Modules/Loadable/Transforms/Testing/Cxx/qSlicerTransformsModuleWidgetTest.cxx
@@ -25,10 +25,20 @@
 #include "ctkTest.h"
 
 // MRML includes
+#include <qMRMLWidget.h>
 #include "qSlicerTransformsModule.h"
 #include "qSlicerTransformsModuleWidget.h"
 #include <vtkMRMLScene.h>
 #include <vtkMRMLTransformNode.h>
+
+// SubjectHierarchy includes
+#include <qSlicerSubjectHierarchyPluginHandler.h>
+#include <qSlicerSubjectHierarchyPluginLogic.h>
+#include <qSlicerSubjectHierarchyTransformsPlugin.h>
+#include <vtkSlicerSubjectHierarchyModuleLogic.h>
+
+// Slicer includes
+#include <qSlicerApplication.h>
 
 // VTK includes
 #include <vtkMatrix4x4.h>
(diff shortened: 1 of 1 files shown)


Modules/Loadable/Transforms/Testing/Cxx/qSlicerTransformsModuleWidgetTest.cxx
```diff
@@ -25,10 +25,16 @@
...
+#include "qMRMLWidget.h"
+#include "qSlicerApplication.h"
+#include "qSlicerSubjectHierarchyPluginHandler.h"
+#include "qSlicerSubjectHierarchyPluginLogic.h"
 #include "qSlicerTransformsModule.h"
 #include "qSlicerTransformsModuleWidget.h"
+#include "qSlicerSubjectHierarchyTransformsPlugin.h"
 #include <vtkMRMLScene.h>
 #include <vtkMRMLTransformNode.h>
+#include "vtkSlicerSubjectHierarchyModuleLogic.h"
```
jcfr, line 37:
```suggestion

// SubjectHierarchy includes
#include <qSlicerSubjectHierarchyPluginHandler.h>
#include <qSlicerSubjectHierarchyPluginLogic.h>
#include <qSlicerSubjectHierarchyTransformsPlugin.h>
#include <vtkSlicerSubjectHierarchyModuleLogic.h>

// Slicer includes
#include <qSlicerApplication.h>
```

Modules/Loadable/Transforms/Testing/Cxx/qSlicerTransformsModuleWidgetTest.cxx
```diff
@@ -25,10 +25,16 @@
 #include "ctkTest.h"
 
 // MRML includes
+#include "qMRMLWidget.h"
+#include "qSlicerApplication.h"
+#include "qSlicerSubjectHierarchyPluginHandler.h"
+#include "qSlicerSubjectHierarchyPluginLogic.h"
 #include "qSlicerTransformsModule.h"
 #include "qSlicerTransformsModuleWidget.h"
+#include "qSlicerSubjectHierarchyTransformsPlugin.h"
```
jcfr, line 34:
```suggestion
```

Modules/Loadable/Transforms/Testing/Cxx/qSlicerTransformsModuleWidgetTest.cxx
```diff
@@ -25,10 +25,16 @@
 #include "ctkTest.h"
 
 // MRML includes
+#include "qMRMLWidget.h"
+#include "qSlicerApplication.h"
+#include "qSlicerSubjectHierarchyPluginHandler.h"
+#include "qSlicerSubjectHierarchyPluginLogic.h"
```
jcfr, line 31:
```suggestion
#include <qMRMLWidget.h>
```

jcfr (commented):

Before of few nitpicks, this is ready for :rocket: 

Consider also explicitly referencing the commit introducing the test failure:

```diff
BUG: Fix qSlicerTransformsModuleWidgetTest
+
+ Addresses a regression introduced in 20bd9d5a4d8 ("ENH: Use subject hierarchy widget as node selector in core modules", 2024-12-29)

The Transforms module widget now uses a Subject Hierarchy tree widget, which requires initialization of the subject hierarchy module.

This was missing from the test, therefore the Transforms module widget test failed. Fixed by adding subject hierarchy module initialization to the test.

Fixes https://github.com/Slicer/Slicer/issues/8285
```

lassoan:
Thanks for the review @jcfr. I've applied all your suggestions and squashed everything into a single commit. Please have a look and approve if looks good.

lassoan:
@jcfr it seems that the includes after your updates are somehow incomplete (causes build error on linux). Maybe it is not due to your changes, as on Windows there were no build errors neither with the original PR content nor with your updates. Can you check if you get build errors on linux when building on your computer? Thank you!

lassoan:
@jcfr I just got the build error notification, but as I see, the latest build is still in progress. So, let's get back to this when the build is completed.

jcfr:
I will test locally shortly and report back.

lassoan:
There is no build error now. Probably github started a build when a few of your suggestions were committed and not yet all of them, and that understandably failed.

So, it is enough if you have a quick look at the code (of course it is always welcome if you build and test).

jamesobutler:
Should we update the build workflow to specify canceling any in progress when an updated push has been made? 
https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/control-the-concurrency-of-workflows-and-jobs#example-using-concurrency-and-the-default-behavior
```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: true
```

jcfr:
Since expecting pull requests to all be based of the current `main` to be merge-able may not be practical, I think we should not  cancel. In practice, pull request touch different module and having each being built even if not the latest `main` is useful and saves time.

jcfr:
Now we could cancel concurrent build for pull request while allowing all push for  `main` to complete:

```
concurrency:
  group: ${{ github.event_name == 'pull_request' && format('pr-{0}', github.event.pull_request.number) || format('push-{0}', github.sha) }}
  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```

jcfr (approved):

PR #8299, opened 2025-03-06
ENH: Reduce unnecessary error reporting in segmentation SH

cpinter:
The error occurs many times and the superfluous error messages make the log harder to process.


pieper (approved):
Definitely! 👍

PR #8300, opened 2025-03-07
BUG: Fix DICOM scene export

cpinter:
It seems that in the new DCMTK version the img2dcm (and probably others) do not have the --no-latin1 flag anymore. This commit removes the use of that flag, which caused an error on MRB DICOM export. I tested it with special characters (Japanese and Hungarian) and export succeeded.


pieper (approved):
LGTM 👍

lassoan (commented):
Thank you, it looks good to me. Could you add to the commit comment a link to the DCMTK change ( https://github.com/DCMTK/dcmtk/commit/f97c55e39593ffae7b02ad993222d7ac5bcf1680) that snuck in the API change under the disguise of "Added multiframe support to img2dcm"? That could help later if we need to improve this (e.g., specify a custom character set). Thank you!

cpinter:
Done, thanks for the reviews.

PR #8313, opened 2025-03-16
DOC: Update model maker documentation

nlbucki:
I noticed what I think is an error in the documentation for the Model Maker module. When talking about the decimation parameter, at the start of the document it says "Use 0 and 1 if you wish no smoothing nor decimation." However, I think this should read "Use 0 and **0** if you wish no smoothing nor decimation."

Later on, the documentation seems to correctly imply a decimation value of 0 produces no decimation:

"Decimate (Decimate): Chose the target reduction in number of polygons as a decimal percentage (between 0 and 1) of the number of polygons. Specifies the percentage of triangles to be removed. For example, 0.1 means 10% reduction and 0.9 means 90% reduction."

Let me know if I misunderstood something (as I'm very new to using 3D Slicer), but this seemed like an easy fix so I figured I'd report it.


pieper (approved):
Yes, makes sense, thanks for the fix 👍

PR #8314, opened 2025-03-17
BUG: Fix 3D view bounds computation with multivolume rendering

lassoan:
When multi-volume rendering was used for displaying volumes in a 3D view, the dummy volume was always placed in the origin. This resulted in the origin always included in the bounds of actors, thereby potentially shifting the bounding box (purple box that is resized when centering the 3D view).

How to reproduce: Load MRHead, apply a transform that it translates to the right by 400mm. Show volume rendering in 3D, center the 3D view => the view is not centered on the MRHead volume but the origin (0,0,0) position is also kept in the field of view.

This commit fixes the problem by moving the dummy volume into the same position as the first volume.


lassoan:
Thanks for the review!

PR #8315, opened 2025-03-17
BUG: Fix crash in multivolume rendering

lassoan:
EchoVolumeRenderer module in SlicerHeart extension crashed the application when multivolume rendering technique was chosen. Fixed it by correcting a null-pointer check.


pieper (approved):
LGTM 👍

PR #8322, opened 2025-03-21
STYLE: Remove obsolete Qt4 code

jamesobutler:
Per https://doc.qt.io/qt-5/portingguide.html, the Q_WS_WIN preprocessor macros was replaced with Q_OS_WIN.

This commit is a follow-up to https://github.com/Slicer/Slicer/commit/ec58ea1cdf9ed15ffd7ddc19b7a27ffe1a4c589c when `updatePythonOsEnviron` was actually removed.


lassoan (approved):
Thank you, good catch!

PR #8326, opened 2025-03-24
COMP: Make vtkSlicerTerminologiesModuleLogic API backward compatible

### LINKED ISSUE #8324: Changed API in Terminologies breaks Quantitative Reporting extension
## Summary

Trying to load DICOM SEG breaks.

https://github.com/Slicer/Slicer/commit/65409219f724bf03bd0307e9c2062c06030f3951

## Steps to reproduce

See issue and PR:

https://github.com/QIICR/QuantitativeReporting/issues/282

https://github.com/QIICR/QuantitativeReporting/pull/281

The PR fixes the extension, but I think it would be better practice to restore the original API so as not to break other code that we may not know about. 

It could be marked as deprecated, but having stubs for the changed method names that call the new methods would be better practice.

## Environment
- Slicer version: Slicer-5.8 and later
- Operating system: all

lassoan:
Recently vtkSlicerTerminologiesModuleLogic methods were renamed to make the API more consistent (mostly using "region" instead of somewhat randomly using anatomicalRegion/anatomy/region in their name). This change caused runtime error in QuantitativeReporting extension.

This commit restores the removed methods. They are deprecated methods, which will be removed in the future, but for now they log a warning message and then call the corresponding new method.

fixes #8324


pieper (approved):
LGTM 👍

PR #8331, opened 2025-03-24
BUG: Fix failing py_nomainwindow_* tests

Sunderlandkyl:
Prevent CaptureToolBar initialization in qSlicerSceneViewsModule unless main window exists. Fixes error/memory leak in py_nomainwindow_* tests added in fb6fb19ef37f60abc7c021c688560fedd1ccfebe.

### DIFF
--- a/Modules/Loadable/SceneViews/qSlicerSceneViewsModule.cxx
+++ b/Modules/Loadable/SceneViews/qSlicerSceneViewsModule.cxx
@@ -23,7 +23,7 @@
 class qSlicerSceneViewsModulePrivate
 {
 public:
-    qMRMLCaptureToolBar* CaptureToolBar;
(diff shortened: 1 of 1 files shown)


lassoan (approved):
Looks good, thank you!

lassoan:
Just one thing - now that `CaptureToolBar` does not always exist, make sure that the variable is initialized to `nullptr`. Also, everywhere the capture toolbar is accessed, it now has to be checked if it is non-nullptr (I could not find any uses outside `qSlicerSceneViewsModule::setup`, but just in case, you could have a look, too).

Sunderlandkyl:
Added the nullptr initialization. I also couldn't find any other references outside qSlicerSceneViewsModule::setup.

lassoan (approved):
Thank you for the update, looks good.

PR #8332, opened 2025-03-25
BUG: Fix invalid scene view indexing

Sunderlandkyl:
SceneViewIndexToSequenceBrowserIndex should only look through SceneViews sequences when trying to determine the index of a specific scene view, however all sequences were being iterated through. This caused the wrong index to be returned when trying to find the index of a scene that contained other sequences.

Fixed by only looking at sequence browser nodes that belong to SceneViews.


lassoan (approved):
Looks good to me, thank you!

Issue #8320, opened 2025-03-20
Modules in different category in Slicer than in documentation


robertf:
The modules **Resample DTI Volume** and **Resample Scalar/Vector/DWI Volume** are under **Diffusion > Utilities** in the Slicer GUI, but are under **Converters** at https://slicer.readthedocs.io/en/latest/user_guide/modules/index.html#converters.

## Environment
- Slicer version: Slicer-5.9.0-2025-03-17
- Operating system: Windows 11

robertf:
I've tried to create a pull request (#8333), but it keeps failing the Commit Message Check, and I don't understand why.

robertf:
I closed pull request #8333 and created a new one #8334 but it still failed the Commit Message Check because it still somehow has the old malformed `Update index.md`. I'm using the GitHub Desktop client, and now I'm having trouble undoing things.

PR #8334, opened 2025-03-26
DOC: Correct error in list of modules

robertf:
DOC: Correct error in list of modules

The modules **Resample DTI Volume** and **Resample Scalar/Vector/DWI Volume** are under **Diffusion > Utilities** in the Slicer 5.9.0 GUI, but are under **Converters** at https://slicer.readthedocs.io/en/latest/user_guide/modules/index.html. I have moved them.

fixed https://github.com/Slicer/Slicer/issues/8320

### DIFF
--- a/Docs/user_guide/modules/index.md
+++ b/Docs/user_guide/modules/index.md
@@ -83,6 +83,8 @@ multivolumeexplorer.md
 :maxdepth: 1
 dwiconvert.md
 brainsdwicleanup.md
+resampledtivolume.md
+resamplescalarvectordwivolume.md
 ```
 
 ## Filtering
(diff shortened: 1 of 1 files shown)


jcfr:
Thanks for working on this 🙏 

If you are creating a new pull request from the Github website, at the time of creating the pull request, you have the option of entering the commit message:

image

Instead of `Update index.md` ... you should enter `DOC: Fix Registration modules category`

Extended description could also be set to something like this:

```
Consistently categorized the modules `Resample DTI Volume` and `Resample Scalar/Vector/DWI Volume`
so that they are listed under `Diffusion > Utilities` in both the application module selector and
the documentation.
```

PR #8341, opened 2025-03-26
ENH: Add invert colors option to Volumes module

lassoan:
If enabled, the order of colors in the lookup table is reversed. This makes it possible to use the same color tables for images regardless of the objects of interest have higher or lower voxel values than the background.

The reversed colors are also applied to volume rendering, if the user chose to synchronize volume rendering transfer functions with scalar volume display.

image

image

@muratmaga

### DIFF
--- a/Libs/MRML/Core/vtkMRMLScalarVolumeDisplayNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLScalarVolumeDisplayNode.cxx
@@ -52,6 +52,7 @@ vtkMRMLScalarVolumeDisplayNode::vtkMRMLScalarVolumeDisplayNode()
   this->AutoWindowLevel = 1;
   this->AutoThreshold = 0;
   this->ApplyThreshold = 0;
+  this->InvertDisplayScalarRange = 0;
 
   // try setting a default grayscale color map
   //this->SetDefaultColorMap(0);
@@ -247,6 +248,11 @@ void vtkMRMLScalarVolumeDisplayNode::WriteXML(ostream& of, int nIndent)
   }
   {
   std::stringstream ss;
+  ss << this->InvertDisplayScalarRange;
+  of << " invertDisplayScalarRange=\"" << ss.str() << "\"";
+  }
+  {
+  std::stringstream ss;
   ss << this->AutoWindowLevel;
   of << " autoWindowLevel=\"" << ss.str() << "\"";
   }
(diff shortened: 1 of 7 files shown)


pieper (dismissed):
Nice 👍

jamesobutler (commented):
@lassoan You were reading my mind. I had to implement a custom way for this recently that was not a full real solution. We called our checkbox "Reverse" rather than "Invert". Are either of those terms more common in this situation?

lassoan:
I've been thinking about using the "inverse" or "reverse" word, too, because the operation has the same effect as _inverting_ voxel values or _reversing_ order of colors.

However, in the end "invert" seemed somewhat more appropriate:
- The volume-oriented term seems more relevant, because the operation is applied to volume display (it is not a property of the color node, but a property of the volume display node).
- When you click "invert", the effect is that the image appears inverted (the colormap itself is not changed).
- The IHE standard uses "invert" word for this purpose, too (uses "reverse" for replaying a time sequence backwards).

Unless some good arguments comes up for using "reverse", I would keep "invert".

muratmaga:
I think this is great. My vote is for "invert", which is the same term used in the ImageJ (well technically what we have is invert LUT).
image

lassoan:
> You are changing voxel values which then map to the same color table order? Our implementation was maintaining the voxel values because they are used for image statistics purposes and reversing the color table that they map to. We were not inverting the volume, but inverting the color table.

The only data that is changed is a flag in the scalar volume display node. Image and color node remain unchanged.

lassoan:
> For example, our sister application uses "Reverse" in the context of the color table

Absolutely, this makes sense. If the flag was a property of the color node I would vote for calling it "reverse", too. It could be possible to add a flag to the color node as well, but most color tables are read-only, so we could not easily modify them. Keeping color tables out of the main scene (similarly to how volume rendering presets are stored in a private scene) would allow us to make color tables editable, improving both the software design and user experience.

lassoan:
I've made color legends aware of the inversion, too, so someone please have a look and approve if looks good.

lassoan:
All checks passed and I think there is a consensus on using "invert" for the volume display node (and later "reverse" for color tables if we add that option). I would just need someone to review&approve latest version so that I don't need to override the merging rules. Thank you!

jamesobutler:
I currently have another build going locally on my system and am planning to review once that finishes. It won't be before 11pm ET though.

jamesobutler (approved):
I have built and tried this out and it looks good 👍🏻

lassoan:
Thank you @jamesobutler for all the feedback and testing.

PR #8354, opened 2025-03-31
ENH: Add option to hide missing display nodes in sequences

Sunderlandkyl:
This commit adds an additional missing item mode to vtkMRMLSequenceBrowserNode to hide proxy display nodes that are not present in a sequence. This option is used by SceneViews to hide display nodes for scene views that were captured before the display node was present.


lassoan (approved):
This looks good to me, thank you!

PR #8358, opened 2025-04-04
BUG: Fix excessive qMRMLSceneViewMenu::resetMenu calls

Sunderlandkyl:
`qMRMLSceneViewMenu::resetMenu` was previously called every time a sequence proxy node was modified. The menu shouldn't need to update if the proxy node is modified, so it makes sense to remove this call to improve performance.


lassoan (approved):
Looks good, thank you!

PR #8361, opened 2025-04-06
BUG: Fix crash in vtkSlicerTerminologiesModuleLogic

### LINKED ISSUE #8360: Editing segment name or colour may crash Slicer
Hello,

Please consider this finding.

If "Use standard terminology for segments" is checked, a double click in a segment entry of the 'Segment editor' to change the name or the colour crashes the application.

If this option is unchecked, no problem occurs.

Using 5.9.0-2025-04-02 r33585 / 191b351 on Linux.

Thanks and regards.

lassoan:
Added missing null-pointer check to avoid crash in case of incomplete terminology.

fixes #8360

### DIFF
--- a/Modules/Loadable/Terminologies/Logic/vtkSlicerTerminologiesModuleLogic.cxx
+++ b/Modules/Loadable/Terminologies/Logic/vtkSlicerTerminologiesModuleLogic.cxx
@@ -2552,7 +2552,7 @@ bool vtkSlicerTerminologiesModuleLogic::FindFirstColorNodeOrTerminology(
   std::string categoryScheme;
   std::string categoryValue;
   vtkSlicerTerminologyCategory* categoryObject = entry->GetCategoryObject();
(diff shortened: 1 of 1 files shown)


chir-set:
Great, I'll build again when the fix is merged, thanks.

lassoan:
Would you mind testing this before before it is merged into the main branch? It seems that there is something unusual about your terminology selection that led to this crash and maybe fixing it will bring up issues elsewhere. If you find something else, it would be nice to include in #8360, too.

chir-set:
> testing this

Ok, I'll start a build and keep you updated tomorrow.

chir-set:
> testing this

The patch fixes the issue. The checked status of said option does not result in a crash when a segment's name or colour is being edited by double-click in the 'Segment editor'.

cpinter (approved):
Definitely, thanks for the fix!

Issue #8355, opened 2025-04-01
Crash when hardening transform on composite transform


lassoan:
_Reported by @che85_

## Summary

Hardening the transform in [this scene](https://github.com/lassoan/PublicTestingData/releases/download/data/bug_1.mrb) freezes Slicer on Linux and MacOS. On Windows I am getting the error in the screenshot.

## Steps to reproduce

- Load [`bug_1.mrb`](https://github.com/lassoan/PublicTestingData/releases/download/data/bug_1.mrb)
- Go to Data module
- Right-click in transform column of `Transform_4_3` node and click `Harden transform`

## Environment
- Slicer version: Slicer-5.8.1
- Operating system: Windows and Linux and Mac

lassoan:
In addition, it was discovered that hardening the transform on `Transform_4_3` changes the transform:
- Load [bug_1.mrb](https://github.com/lassoan/PublicTestingData/releases/download/data/bug_1.mrb)
- Load MRHead sample data set
- Go to Data module
- Apply `Transform_4_3` to MRHead
- Right-click in transform column of `Transform_4_3` node and click `Harden transform` => MRHead image slightly moves/deforms (due to the transform being changed by hardening)

PR #8365, opened 2025-04-08
Fix hardening of inverted composite transform

lassoan:
If parent transform(s) got hardened on an inverted composite transform then the resulting transform was incorrect.

The problem was that the hardened transform components were concatenated to the inverted transform and when Update()
was called on that transform then it was updated from the original (non-inverse) transform, removing all concatenations.

Fixed the problem by creating a new composite transform when a transform is hardened on an inverted composite transform.

fixes https://github.com/Slicer/Slicer/issues/8355

@che85


jcfr (approved):

jcfr:
Nitpick: the typo in comment below could be fixed in a follow-up commit ...

lassoan:
> Nitpick: the typo in comment below could be fixed in a follow-up commit ...
> 

Good catch, fixed.

jcfr (approved):

lassoan:
Thanks for the quick review @jcfr!

PR #8370, opened 2025-04-11
BUG: Adjust the FieldOfView to match the aspect ratio of the slice dimensions.

Punzo:
This is a small fix introduced in https://github.com/Slicer/Slicer/pull/8141

Kyle asked if it can be integrated in a separate PR:
https://github.com/Slicer/Slicer/pull/8141#discussion_r2038099444

to reproduce:
1) enter in dual monitor layout
2) load example nrrd file
3) Copy and paste in the python console
```
a = slicer.mrmlScene.GetNodeByID("vtkMRMLSliceNodeRed")
b = slicer.mrmlScene.GetNodeByID("vtkMRMLSliceNodeRed+")
a.Copy(b)
```
4) you get
image

With this fix:
image

@Sunderlandkyl @lassoan

### DIFF
--- a/Libs/MRML/Core/vtkMRMLAbstractViewNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLAbstractViewNode.cxx
@@ -59,6 +59,8 @@ vtkMRMLAbstractViewNode::vtkMRMLAbstractViewNode()
   {
     this->AxisLabels->InsertNextValue(DEFAULT_AXIS_LABELS[i]);
   }
+
+  this->MappedInLayout = false;
 }
 
 //----------------------------------------------------------------------------
@@ -100,6 +102,9 @@ void vtkMRMLAbstractViewNode::WriteXML(ostream& of, int nIndent)
   // Do not write screenScaleFactor attribute, as we should not use the
   // value that is in the scene but what the user has set in application settings.
   // vtkMRMLWriteXMLFloatMacro(screenScaleFactor, ScreenScaleFactor);
+
+  // Do not write MappedInLayout attribute, as it is set dynamically by the layout manager.
+  // It should not be copied between objects or saved to file.
 
   vtkMRMLWriteXMLEndMacro();
 
@@ -151,6 +156,9 @@ void vtkMRMLAbstractViewNode::ReadXMLAttributes(const char** atts)
   // Do not read screenScaleFactor attribute, as we should not use the
   // value that is in the scene but what the user has set in application settings.
   // vtkMRMLReadXMLFloatMacro(screenScaleFactor, ScreenScaleFactor);
+
+  // Do not read MappedInLayout attribute, as it is set dynamically by the layout manager.
+  // It should not be copied between objects or saved to file.
 
   vtkMRMLReadXMLEndMacro();
 
@@ -205,11 +213,6 @@ void vtkMRMLAbstractViewNode::ReadXMLAttributes(const char** atts)
   }
 #endif
 
-  // Do not restore MappedInLayout state, because the view may not be mapped into the layout just yet.
-  // (the attribute tells that it was mapped into the layout when the scene was saved but the current
-  // layout may be different, see issue #6284).
-  this->SetAttribute("MappedInLayout", nullptr);
-
   this->EndModify(disabledModify);
 }
 
@@ -238,6 +241,10 @@ void vtkMRMLAbstractViewNode::CopyContent(vtkMRMLNode* anode, bool deepCopy/*=tr
   }
   vtkMRMLCopyEnumMacro(RulerColor);
   vtkMRMLCopyFloatMacro(ScreenScaleFactor);
+
+  // Do not copy MappedInLayout attribute, as it is set dynamically by the layout manager.
+  // It should not be copied between objects or saved to file.
+
   vtkMRMLCopyEndMacro();
 
   vtkMRMLAbstractViewNode *node = vtkMRMLAbstractViewNode::SafeDownCast(anode);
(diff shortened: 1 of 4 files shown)


Libs/MRML/Core/vtkMRMLSliceNode.cxx
```diff
@@ -951,7 +951,22 @@ void vtkMRMLSliceNode::ReadXMLAttributes(const char** atts)
     // Layout color is set in Superclass
     layoutColorFound = true;
   }
+
+  // Read Dimensions only if the current Dimensions are the default ones
+  if (this->Dimensions[0] == 256 && this->Dimensions[1] == 256 && this->Dimensions[2] == 1)
```
lassoan, line 956: Why do we read/copy only when default value? Do we really ever want to read this from file?

(256,256,1) is a valid value, so if we want to make decision based on it then we would need to change the default to some invalid value (0,0,0) or maybe (-1,-1,-1).

Sunderlandkyl:
Thanks @Punzo, this change works well for slice sequences!

Libs/MRML/Core/vtkMRMLSliceNode.cxx
Punzo: Done in https://github.com/Slicer/Slicer/pull/8370/commits/52914830db2a06573ec1446dca932bb51ba8e9c9

Punzo:
> Thanks @Punzo, this change works well for slice sequences!

thanks for testing @Sunderlandkyl. I added the MappedInLayout fix as well https://github.com/Slicer/Slicer/pull/8141#discussion_r2039878428

Can you re-test/review? thanks!

@lassoan thanks for reviewing, I have addressed your comment.

Libs/MRML/Core/vtkMRMLAbstractViewNode.h
```diff
@@ -358,6 +360,10 @@ class VTK_MRML_EXPORT vtkMRMLAbstractViewNode
   /// Labels of coordinate system axes
   vtkSmartPointer<vtkStringArray> AxisLabels;
 
+  ///
+  /// Define if the view is mapped in the current layout with a Qwidget.
+  int MappedInLayout;
```
lassoan, line 365: Need to initialize, it can be done here or in the constructor.

Sunderlandkyl:
@Punzo Thanks, it works great!

Libs/MRML/Core/vtkMRMLAbstractViewNode.h
```diff
@@ -132,8 +132,10 @@ class VTK_MRML_EXPORT vtkMRMLAbstractViewNode
   /// \sa GetVisibility()
   /// \sa IsViewVisibleInLayout()
   /// \sa vtkMRMLLayoutNode::SetViewArrangement()
+  vtkSetMacro(MappedInLayout, int);
+  vtkGetMacro(MappedInLayout, int);
+  /// Deprecated. Use GetMappedInLayout instead.
```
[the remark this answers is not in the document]
Punzo: No special reason, before an additional method was necessary because the varibale was saved in the node attributes. Since now it is a variable in the node, we can simply sue the get method. For me there is no difference, I will remove the depracation warning

Libs/MRML/Core/vtkMRMLAbstractViewNode.h
```diff
@@ -132,8 +132,10 @@ class VTK_MRML_EXPORT vtkMRMLAbstractViewNode
   /// \sa GetVisibility()
   /// \sa IsViewVisibleInLayout()
   /// \sa vtkMRMLLayoutNode::SetViewArrangement()
+  vtkSetMacro(MappedInLayout, int);
```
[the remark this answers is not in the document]
Punzo: yes you need it in qMRMLLayoutViewFactory::setupView.

Libs/MRML/Core/vtkMRMLAbstractViewNode.h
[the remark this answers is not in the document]
Punzo: I declared int because the original `IsMappedInLayout` was returning a int. But i agree it should be a bool and `IsMappedInLayout` shoudl return a bool

Punzo: a nice, I forgot, thanks@

Punzo:
> @Punzo Thanks, it works great!

nice!

Punzo:
@lassoan thanks for reviewing the second commit. I have applied your comments.

Punzo:
@Sunderlandkyl @lassoan can you please have a look at this again, and in case merge? I would like to avoid conflicts with https://github.com/Slicer/Slicer/pull/8141

Libs/MRML/Core/vtkMRMLAbstractViewNode.cxx
```diff
@@ -101,6 +103,9 @@ void vtkMRMLAbstractViewNode::WriteXML(ostream& of, int nIndent)
   // value that is in the scene but what the user has set in application settings.
   // vtkMRMLWriteXMLFloatMacro(screenScaleFactor, ScreenScaleFactor);
 
+  // Do not write MappedInLayout attribute, as it is not a property of the view node.
```
lassoan, line 106:
```suggestion
  // Do not write MappedInLayout attribute, as it is set dynamically by the layout manager. It should not be copied between objects or saved to file.
```

Sunderlandkyl:
@Punzo Changes look good to me! @lassoan any other comments?

Libs/MRML/Core/vtkMRMLAbstractViewNode.h
```diff
@@ -358,6 +358,10 @@ class VTK_MRML_EXPORT vtkMRMLAbstractViewNode
   /// Labels of coordinate system axes
   vtkSmartPointer<vtkStringArray> AxisLabels;
 
+  ///
+  /// Define if the view is mapped in the current layout with a Qwidget.
```
lassoan, line 362: MRML is not aware of Qt.

```suggestion
  /// Define if the view is displayed in the current layout.
```

lassoan (approved):
Looks good to me, thank you.

PR #8371, opened 2025-04-11
BUG: Update CTK to fix `ctkDICOMDatabase::fileValue` for non-cached files

jcfr:
Commit commontk/CTK@88ff72b (introduced through f0138cd0b) changed `ctkDICOMDatabase::fileValue` behavior so that it no longer returns DICOM tag value for files that are not already in the database.

This is an issue because this method has been used universally for any DICOM files (it returned result faster for files that are cached in the DICOM database but it worked for files not in the database, too).

This commit restores the previous, preferred behavior.

List of CTK changes:

```
$ git shortlog 82cae578..7f8e1a67 --no-merges
Andras Lasso (1):
      BUG: Fix ctkDICOMDatabase::fileValue for non-cached files
```

### DIFF
--- a/SuperBuild/External_CTK.cmake
+++ b/SuperBuild/External_CTK.cmake
@@ -71,7 +71,7 @@ if(NOT DEFINED CTK_DIR AND NOT Slicer_USE_SYSTEM_${proj})
 
   ExternalProject_SetIfNotDefined(
     Slicer_${proj}_GIT_TAG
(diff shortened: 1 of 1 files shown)


jamesobutler (approved):
Based on the review originally provided at https://github.com/commontk/CTK/pull/1227, updating the CTK commit hash makes sense now.

PR #8372, opened 2025-04-11
BUG: Disable logging of VTK deprecation warnings during Python autocompletion

jcfr:
Follow-up of 3ae750f4714 ("BUG: Avoid logging errors during Python auto-completion with CTK update", 2023-01-19)

Related pull requests:
* https://github.com/Slicer/Slicer/pull/8238

### DIFF
--- a/Base/QTGUI/qSlicerApplication.cxx
+++ b/Base/QTGUI/qSlicerApplication.cxx
@@ -147,33 +147,37 @@ public:
 
   void updateCompletionModel(const QString& completion) override
   {
-    bool wasLoggingEnabled = this->setLoggingEnabled(false);
+    bool wasQtLoggingEnabled = this->setLoggingEnabled(false, ctkErrorLogQtMessageHandler::HandlerName);
+    bool wasVTKLoggingEnabled = this->setLoggingEnabled(false, ctkVTKErrorLogMessageHandler::HandlerName);
     ctkPythonConsoleCompleter::updateCompletionModel(completion);
-    this->setLoggingEnabled(wasLoggingEnabled);
+    this->setLoggingEnabled(wasQtLoggingEnabled, ctkErrorLogQtMessageHandler::HandlerName);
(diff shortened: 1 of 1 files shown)


pieper:
Would it be possible to capture these message and add a note about deprecation on the menu of completions?  Or maybe we can query the python wrapping to know if the method is deprecated?

jcfr:
> Would it be possible to capture these message and add a note about deprecation on the menu of completions? 

Similarly to the warnings reported using `qWarning` are handled during autocompletion, the VTK messages are still reported in the log (and terminal).

In this case, after fixing the completion logic (https://github.com/commontk/CTK/pull/1229), I observed that the following warning was then reported hence this commit.

https://github.com/Slicer/Slicer/blob/c9966885e7ea6cec9e24e974367418f7d5aefa60/Libs/MRML/Core/vtkMRMLScalarVolumeDisplayNode.h#L71-L77

> Or maybe we can query the python wrapping to know if the method is deprecated?


That would be a nice enhancement. However, it would require work on multiple fronts:
1. At the Slicer level, we'd need to consistently label deprecated functions using a macro like `SLICER_DEPRECATED_IN_X_Y_Z`, similar to what's done in VTK.
2. At the VTK wrapping level, the infrastructure would need to expose deprecation metadata in a way that can be queried.

From what I’ve seen, the VTK wrapping system already parses deprecation annotations[^1], so this could be doable.

[^1]: https://github.com/search?q=repo%3AKitware%2FVTK+VTK_PARSE_DEPRECATED&type=code

pieper:
Thanks Jc.  That'll be good to keep in mind for the future.  Cleaning up the logs of these messages is for sure a good step.

jcfr:
> That'll be good to keep in mind for the future

I just created a feature request to keep track of this. In the meantime, I would appreciate it if you could approve so that the changes are integrated once CI completes :rocket:

pieper (approved):
LGTM 👍

PR #8381, opened 2025-04-16
DOC: Update build instructions to warn users about CMake 4 incompatibility

lassoan:
Slicer build currently fails with the recently released CMake 4.x.

### DIFF
--- a/Docs/developer_guide/build_instructions/linux.md
+++ b/Docs/developer_guide/build_instructions/linux.md
@@ -47,7 +47,8 @@ sudo apt update && sudo apt install git build-essential cmake cmake-curses-gui c
 
 :::{note}
 The CMake version currently included in Debian 12 Bookworm (Stable) is not compatible with the current development version of Slicer.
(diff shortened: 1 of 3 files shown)


jcfr (approved):

jamesobutler:
Want to go so far as https://github.com/Slicer/Slicer/blob/de382bd89e0641e7273a92d4e169e336ef772675/CMakeLists.txt#L3-L5?

pieper (approved):
Yes, this looks good.

If updating out build system to support cmake 4 is a problem, we should think about making a "super-superbuild" script that downloads a working version of cmake and then runs the build process so people don't have to do all this stuff manually.  We might also do the same for Qt.

lassoan:
If this kind of cataclysmic CMake event happens once in every 10 years (CMake 3 was released in 2014) then we probably don't need to change our processes.

lassoan:
Added a pull request to 5.8 branch with the same documentation update and a CMake early failure report as @jamesobutler suggested.

https://github.com/Slicer/Slicer/pull/8382

pieper:
> And all the Visual Studio dependencies on Windows? Seems like it could get complicated for such a "super-superbuild" idea.

A bit complicated yes, but it's doable - people do it all the time with github actions.



> If this kind of cataclysmic CMake event happens once in every 10 years (CMake 3 was released in 2014) then we probably don't need to change our processes.

I think a fatal error for cmake 4 at the start of the process like James suggested is good enough for now.

PR #8393, opened 2025-04-24
ENH: Add functions to update the contents of existing scene views

Sunderlandkyl:
This commit adds UpdateNthSceneView functions to update the contents of an existing scene view. These functions can be used to update individual nodes, or replace the content of scene views with the state of the current scene.

### DIFF
--- a/Modules/Loadable/SceneViews/Logic/vtkSlicerSceneViewsModuleLogic.h
+++ b/Modules/Loadable/SceneViews/Logic/vtkSlicerSceneViewsModuleLogic.h
@@ -34,6 +34,7 @@
 
 // MRML includes
 class vtkMRMLSequenceBrowserNode;
+class vtkMRMLSequenceNode;
 class vtkMRMLSceneViewNode;
 class vtkMRMLTextNode;
 class vtkMRMLVolumeNode;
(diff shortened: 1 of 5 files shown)


Modules/Loadable/SceneViews/Logic/vtkSlicerSceneViewsModuleLogic.h
```diff
@@ -73,9 +74,17 @@ class VTK_SLICER_SCENEVIEWS_MODULE_LOGIC_EXPORT vtkSlicerSceneViewsModuleLogic :
...
   int GetSceneViewIndexByName(std::string name);
 
-  /// Modify an existing sceneView.
+  /// Modify the metadata of an existing sceneView.
   void ModifyNthSceneView(int sceneViewIndex, std::string name, std::string, int screenshotType, vtkImageData* screenshot);
 
+  //@{
+  /// Update the contents of a sceneView to match the current state of the scene.
+  /// If new nodes are specified, they will be added to the scene view.
+  void UpdateNthSceneView(int sceneViewIndex, bool updateExistingNodes = true, bool saveDisplayNodes = true, bool saveViewNodes = true);
```
lassoan, line 83: Suggestion from copilot and actually a valid point. Order of arguments is different in the the cxx file:

```
vtkSlicerSceneViewsModuleLogic::UpdateNthSceneView(int sceneViewIndex,
  bool saveDisplayNodes/*=true*/,
  bool saveViewNodes/*=true*/,
  bool updateExistingNodes/*=true*/)
```

lassoan (commented):
Thank you, this works great. Maybe we could clarify the behavior of the checkboxes in tooltips or layout changes. Specifically:
- If "Update existing nodes" is enabled but "Capture display nodes" is disabled and "Capture view nodes" is disabled then are we still going to have display and view nodes in the scene view if originally the scene view contained display and view nodes?
- If "Update existing nodes" is disabled but "Capture display nodes" is enabled then are we adding or updating any display nodes?

It would be also nice to be able to update the thumbnail (currently, that section is disabled).

Sunderlandkyl:
@lassoan any other changes?

lassoan:
It should be good, but I'm testing it now.

lassoan (approved):
This works very nicely, thank you! There are a few small improvements (mainly GUI text, layout, etc.) that would be nice to implement, but let's merge this first. I'll submit a new issue to track those.

lassoan:
Follow-up issue to discuss further enhancements: https://github.com/Slicer/Slicer/issues/8411

PR #8401, opened 2025-04-29
BUG: Restore fallback DICOM SEG upload with alternative storescu config

PereGinebra:
Sending SEG files to a DICOM server using DIMSE currently fails because the first storescu attempt will raise a UserWarning, so the second attempt with the modified storescu configuration doesn't go through.

This is the easiest and least intrusive solution I could come up with, although switching to dcmsend might be a good idea.

---

Catch `UserWarning` raised on failure of the initial `storescu` send attempt, allowing fallback to a custom configuration to proceed.

This restores functionality originally introduced in 21f8ba46beb ("ENH: Add custom DICOM send configuration", 2021-06-14), which was most likely broken by adb997269ba ("COMP: Refactor DICOMProcess and DICOMCommand", 2023-10-02).

### DIFF
--- a/Modules/Scripted/DICOMLib/DICOMProcesses.py
+++ b/Modules/Scripted/DICOMLib/DICOMProcesses.py
@@ -580,9 +580,12 @@ class DICOMSender:
         :param file: Path to the local DICOM file to transmit.
         """
 
-        if self._dicomSendSCU(file):
-            # success
-            return True
+        try:
+            if self._dicomSendSCU(file):
+                # success
+                return True
+        except UserWarning as uw:
+            logging.info(f'Initial DICOM storescu attempt raised: "{uw}". Possible cause: unsupported SOP class.')
 
         # Retry transfer with alternative configuration with presentation contexts which support SEG/SR.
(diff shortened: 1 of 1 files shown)


pieper (dismissed):
Looks good to me, thanks for the fix 👍

jcfr (dismissed):

Modules/Scripted/DICOMLib/DICOMProcesses.py
```diff
@@ -580,9 +580,12 @@ def _sendOneFileWithDIMSE(self, file: str) -> None:
...
 
-        if self._dicomSendSCU(file):
-            # success
-            return True
+        try:
+            if self._dicomSendSCU(file):
+                # success
+                return True
+        except UserWarning as uw:
+            logging.info(f"DICOM storescu raised: \"{uw}\", possibly due to a SEG file sent with a default configuration.")
```
jcfr, line 588:
```suggestion
        except UserWarning as uw:
            logging.info(f'Initial DICOM storescu attempt raised: "{uw}". Possible cause: unsupported SOP class.')
```

jcfr (approved):

PR #8403, opened 2025-04-30
BUG: Fix volume rendering clipping for volumes with negative scalar

Sunderlandkyl:
When applying clipping to the volume, the clipped regions would be filled with "0", which could result in the region not being clipped if the transfer function had visible voxel values at "0". Fixed by adding adding ClippingBlankVoxelValue and AutoClippingBlankVoxelValue attributes. By default the clipped voxel value will be automatically set from vtkMRMLVolumeNode::GetImageBackgroundScalarComponentAsDouble(). The clipped scalar value can be set to any scalar by setting ClippingBlankVoxelValue when AutoClippingBlankVoxelValue is disabled.

image

### DIFF
--- a/Libs/MRML/Core/vtkImageMathematicsAddon.cxx
+++ b/Libs/MRML/Core/vtkImageMathematicsAddon.cxx
@@ -196,7 +196,10 @@ void vtkImageMathematicsAddonExecute2(vtkImageMathematicsAddon* self, vtkImageDa
       }
       for (idxR = 0; idxR < rowLength; idxR++)
       {
-        *outPtr = *inPtr * static_cast<float>((*outPtr - normalizeScalarRange[0]) / normalizeMagnitude);
+        // The formula interpolates between the input value (*inPtr) and the lower bound of the scalar range
+        // (normalizeScalarRange[0]), weighted by the normalized value.
+        float normalizedValue = static_cast<float>((*outPtr - normalizeScalarRange[0]) / normalizeMagnitude);
+        *outPtr = normalizedValue * (*inPtr) + (1.0 - normalizedValue) * normalizeScalarRange[0];
         outPtr++;
         inPtr++;
       }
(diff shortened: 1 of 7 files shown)


lassoan:
> By default the clipped voxel value will be automatically set from the scalar range by default.

From the code it looks like `vtkMRMLVolumeNode::GetImageBackgroundScalarComponentAsDouble`, which is good (it is better than just determine a value from the scalar range). Is this commit message accurate?

Sunderlandkyl:
No, the commit message was wrong. I've updated it to reflect the use of vtkMRMLVolumeNode::GetImageBackgroundScalarComponentAsDouble.

Libs/MRML/Core/vtkImageMathematicsAddon.cxx
```diff
@@ -196,7 +196,9 @@ void vtkImageMathematicsAddonExecute2(vtkImageMathematicsAddon* self, vtkImageDa
       }
       for (idxR = 0; idxR < rowLength; idxR++)
       {
-        *outPtr = *inPtr * static_cast<float>((*outPtr - normalizeScalarRange[0]) / normalizeMagnitude);
+        *outPtr = (*inPtr - normalizeScalarRange[0])
```
lassoan, line 199: Somehow the computation does not seem to work as intended (at least when smooth edge is enabled). We would need a smooth transition from the blank value to the voxel value. Yet, somehow in the transition zone other values appear.

For example: there is a blue "capping" that should not be visible in this volume rendering. Blue correponds to voxel values between -220 and 20. How this blue cap is possible at the boundary of blank value (457) and values of the contrast filling (430)? All values at the cut surface should be around 430-460, and so we should be able to see into the ventricle (similarly how we can see into the ventricle when cropping).

image

Sunderlandkyl: The behavior makes sense in that we are blurring between the transparent range (329-1058), and the background scalar (-1024). If we are applying soft edge blurring, the blurred values would need to go through the intervening scalars (-308-329).

Sunderlandkyl:
The background color should now be correctly applied, which allows users to still be able to look into the cavity. Note that this may also a false "surface" to show up at the edge of the clipping as the interpolation is applied between air and "background" values.
image

Sunderlandkyl:
@lassoan any other changes?

lassoan (approved):
Thank you, this all works well now.

image

PR #8416, opened 2025-05-15
ENH: Set color picker to use basic colors tab by default

jamesobutler:
With application setting "Subject Hierarchy->Use standard terminology" set to Off, then double-clicking on color in the Markups table produces the following result:

| `main` | This PR |
|--------|----------|
|image|image|

I have set the tab to remain the first tab which is the basic colors. This is because when not using standard terminology there is not a need for the "Labels" tab to be used where "GenericAnatomyColors" is essentially picking terminologies again. The "Labels" color picker functionality is a little weird in this dialog as do users really care to pick a specific color from a color table node? I would assume they would likely just be picking a basic color or pick from one of their saved custom colors.


lassoan (approved):
I agree, the label-based color selection is now mostly redundant. We could probably even remove that tab, but for now just de-emphasizing it is enough - as you did in this PR.

lassoan:
Thank you for your contribution @jamesobutler !

PR #8091, opened 2024-12-11
BUG: Make node type labels translatable

mhdiop:
* Slicer/SlicerLanguagePacks#12
* Slicer/SlicerLanguagePacks#21

PR #8420, opened 2025-05-16
Improvements to default node usage across different node creation methods

jamesobutler:
This PR is a collection of commits that aims for default node usage to be respected by various places of Slicer core that create nodes (programmatically through AddNewNodeByClass, GUI through qMRMLNodeComboBox, GUI through Markups Module node creation buttons). I originally started working on these changes after beginning to use a default markups node, but noticing that the Markups module that I was using was not respecting this new default.

I've created the following code snippet to test the various changed logic.

```python
# Start from fresh scene
slicer.mrmlScene.Clear()
# Create node with AddNewNodeByClass (x2) to see how name iterates
markup_node = slicer.mrmlScene.AddNewNodeByClass("vtkMRMLMarkupsLineNode")
print(markup_node.GetName())
markup_node = slicer.mrmlScene.AddNewNodeByClass("vtkMRMLMarkupsLineNode")
print(markup_node.GetName())
# Create node with AddNewNodeByClass with Name prespecified (x2) to see how name iterates
markup_node = slicer.mrmlScene.AddNewNodeByClass("vtkMRMLMarkupsLineNode", "UseThisDefaultNameExactly")
print(markup_node.GetName())
markup_node = slicer.mrmlScene.AddNewNodeByClass("vtkMRMLMarkupsLineNode", "UseThisDefaultNameExactly")
print(markup_node.GetName())
# Create node with Markups module logic (x2) to see how name iterates
markup_node_from_module = slicer.modules.markups.logic().AddNewMarkupsNode("vtkMRMLMarkupsLineNode")
print(markup_node_from_module.GetName())
markup_node_from_module = slicer.modules.markups.logic().AddNewMarkupsNode("vtkMRMLMarkupsLineNode")
print(markup_node_from_module.GetName())
# Create node with Markups module logic with Name prespecified (x2) to see how name iterates
markup_node_from_module = slicer.modules.markups.logic().AddNewMarkupsNode("vtkMRMLMarkupsLineNode", "UseThisDefaultNameExactly")
print(markup_node_from_module.GetName())
markup_node_from_module = slicer.modules.markups.logic().AddNewMarkupsNode("vtkMRMLMarkupsLineNode", "UseThisDefaultNameExactly")
print(markup_node_from_module.GetName())
```

The printed output shows the differences :

|| `main` | This PR |
|-|--------|----------|
|AddNewNodeByClass|MarkupsLine|L|
|AddNewNodeByClass|MarkupsLine_1|L_1|
|AddNewNodeByClass; name specified|UseThisDefaultNameExactly|UseThisDefaultNameExactly|
|AddNewNodeByClass; name specified|UseThisDefaultNameExactly|UseThisDefaultNameExactly|
|Markups Module|L|L_2|
|Markups Module|L_1|L_3|
|Markups Module; name specified|UseThisDefaultNameExactly|UseThisDefaultNameExactly
|Markups Module; name specified|UseThisDefaultNameExactly|UseThisDefaultNameExactly|

-------------------------------------------------

The printed output shows the differences when there is a DEFAULT NODE added:
```python
slicer.mrmlScene.Clear()
# Define a default Markups Line node with a default node name
default_markups_line_node = slicer.vtkMRMLMarkupsLineNode()
default_markups_line_node.SetName("DefaultNodeName")  # Default names are to be used exactly and to not be manipulated through GenerateUniqueName()
slicer.mrmlScene.AddDefaultNode(default_markups_line_node)
# Create node with AddNewNodeByClass (x2) to see how name iterates
markup_node = slicer.mrmlScene.AddNewNodeByClass("vtkMRMLMarkupsLineNode")
print(markup_node.GetName())
markup_node = slicer.mrmlScene.AddNewNodeByClass("vtkMRMLMarkupsLineNode")
print(markup_node.GetName())
# Create node with AddNewNodeByClass with Name prespecified (x2) to see how name iterates
markup_node = slicer.mrmlScene.AddNewNodeByClass("vtkMRMLMarkupsLineNode", "UseThisDefaultNameExactly")
print(markup_node.GetName())
markup_node = slicer.mrmlScene.AddNewNodeByClass("vtkMRMLMarkupsLineNode", "UseThisDefaultNameExactly")
print(markup_node.GetName())
# Create node with Markups module logic (x2) to see how name iterates
markup_node_from_module = slicer.modules.markups.logic().AddNewMarkupsNode("vtkMRMLMarkupsLineNode")
print(markup_node_from_module.GetName())
markup_node_from_module = slicer.modules.markups.logic().AddNewMarkupsNode("vtkMRMLMarkupsLineNode")
print(markup_node_from_module.GetName())
# Create node with Markups module logic with Name prespecified (x2) to see how name iterates
markup_node_from_module = slicer.modules.markups.logic().AddNewMarkupsNode("vtkMRMLMarkupsLineNode", "UseThisDefaultNameExactly")
print(markup_node_from_module.GetName())
markup_node_from_module = slicer.modules.markups.logic().AddNewMarkupsNode("vtkMRMLMarkupsLineNode", "UseThisDefaultNameExactly")
print(markup_node_from_module.GetName())
```

|| `main` | This PR |
|-|--------|----------|
|AddNewNodeByClass|DefaultNodeName|DefaultNodeName|
|AddNewNodeByClass|DefaultNodeName|DefaultNodeName|
|AddNewNodeByClass; name specified|UseThisDefaultNameExactly|UseThisDefaultNameExactly|
|AddNewNodeByClass; name specified|UseThisDefaultNameExactly|UseThisDefaultNameExactly|
|Markups Module|L|DefaultNodeName|
|Markups Module|L_1|DefaultNodeName|
|Markups Module; name specified|UseThisDefaultNameExactly|UseThisDefaultNameExactly|
|Markups Module; name specified|UseThisDefaultNameExactly|UseThisDefaultNameExactly|

---------------------------------------------------
`ENH: Support customizing markups default node prefix` is an additional commit to support customizing Markups objects node prefix. In the example below I change the `DefaultNodeNamePrefix` from "A" to "Angle" for a vtkMRMLMarkupsAngleNode object.

```python
# Start from fresh scene
slicer.mrmlScene.Clear()
# Define a default Markups Angle node with a default node name prefix
default_markups_angle_node = slicer.vtkMRMLMarkupsAngleNode()
default_markups_angle_node.SetDefaultNodeNamePrefix("Angle")
slicer.mrmlScene.AddDefaultNode(default_markups_angle_node)
# Create node with AddNewNodeByClass (x2) to see how name iterates
markup_node = slicer.mrmlScene.AddNewNodeByClass("vtkMRMLMarkupsAngleNode")
print(markup_node.GetName())
markup_node = slicer.mrmlScene.AddNewNodeByClass("vtkMRMLMarkupsAngleNode")
print(markup_node.GetName())
# Create node with AddNewNodeByClass with Name prespecified (x2) to see how name iterates
markup_node = slicer.mrmlScene.AddNewNodeByClass("vtkMRMLMarkupsAngleNode", "UseThisDefaultNameExactly")
print(markup_node.GetName())
markup_node = slicer.mrmlScene.AddNewNodeByClass("vtkMRMLMarkupsAngleNode", "UseThisDefaultNameExactly")
print(markup_node.GetName())
# Create node with Markups module logic (x2) to see how name iterates
markup_node_from_module = slicer.modules.markups.logic().AddNewMarkupsNode("vtkMRMLMarkupsAngleNode")
print(markup_node_from_module.GetName())
markup_node_from_module = slicer.modules.markups.logic().AddNewMarkupsNode("vtkMRMLMarkupsAngleNode")
print(markup_node_from_module.GetName())
# Create node with Markups module logic with Name prespecified (x2) to see how name iterates
markup_node_from_module = slicer.modules.markups.logic().AddNewMarkupsNode("vtkMRMLMarkupsAngleNode", "UseThisDefaultNameExactly")
print(markup_node_from_module.GetName())
markup_node_from_module = slicer.modules.markups.logic().AddNewMarkupsNode("vtkMRMLMarkupsAngleNode", "UseThisDefaultNameExactly")
print(markup_node_from_module.GetName())
```

`main` does not support customizing DefaultNodeNamePrefix, but below is the output using the changes in this PR:

||This PR|
|-|--------|
|AddNewNodeByClass|Angle|
|AddNewNodeByClass|Angle_1|
|AddNewNodeByClass; name specified|UseThisDefaultNameExactly|
|AddNewNodeByClass; name specified|UseThisDefaultNameExactly|
|Markups Module|Angle_2|
|Markups Module|Angle_3|
|Markups Module; name specified|UseThisDefaultNameExactly|
|Markups Module; name specified|UseThisDefaultNameExactly|

### DIFF
--- a/Libs/MRML/Core/vtkMRMLMarkupsNode.h
+++ b/Libs/MRML/Core/vtkMRMLMarkupsNode.h
@@ -1033,9 +1033,6 @@ protected:
 
   std::string PropertiesLabelText;
 
-  /// Store markup type GUI display name. Translated to the application language.
-  std::string TypeDisplayName;
-
   /// Transform that moves the xyz unit vectors and origin of the interaction handles to local coordinates
   vtkSmartPointer<vtkMatrix4x4> InteractionHandleToWorldMatrix;
 
--- a/Libs/MRML/Core/vtkMRMLMarkupsNode.h
+++ b/Libs/MRML/Core/vtkMRMLMarkupsNode.h
@@ -38,10 +38,6 @@
 # define VTK_PROPEXCLUDE
 #endif
 
-// This can be used for writing code that is backward-compatible with older Slicer versions
-// where vtkMRMLMarkupsNode::GetDefaultNodeNamePrefix() returned const char* (instead of std::string).
-#define DEFAULT_NODE_NAME_PREFIX_IS_STD_STRING 1
-
 class vtkMatrix3x3;
 class vtkMRMLUnitNode;
 
@@ -155,13 +151,6 @@ public:
   /// Get markup type internal name. This type name is the same regardless of the
   /// chosen application language and should not be displayed to end users.
   virtual const char* GetMarkupType() { return "Markup"; };
-
-  /// Get markup short name.
-  /// This may be displayed to the user and therefore it is translated to the application language.
-  std::string GetDefaultNodeNamePrefix() override;
-  /// Get markup type GUI display name
-  /// This may be displayed to the user and therefore it is translated to the application language.
-  std::string GetTypeDisplayName() override;
 
   /// Read node attributes from XML file
   void ReadXMLAttributes(const char** atts) override;
@@ -1046,9 +1035,6 @@ protected:
 
   /// Store markup type GUI display name. Translated to the application language.
   std::string TypeDisplayName;
-
-  /// Store markup short name. Translated to the application language.
-  std::string DefaultNodeNamePrefix;
 
   /// Transform that moves the xyz unit vectors and origin of the interaction handles to local coordinates
   vtkSmartPointer<vtkMatrix4x4> InteractionHandleToWorldMatrix;
--- a/Libs/MRML/Core/vtkMRMLMarkupsNode.h
+++ b/Libs/MRML/Core/vtkMRMLMarkupsNode.h
@@ -158,8 +158,7 @@ public:
 
   /// Get markup short name.
   /// This may be displayed to the user and therefore it is translated to the application language.
-  virtual std::string GetDefaultNodeNamePrefix();
-
+  std::string GetDefaultNodeNamePrefix() override;
   /// Get markup type GUI display name
   /// This may be displayed to the user and therefore it is translated to the application language.
   std::string GetTypeDisplayName() override;
--- a/Libs/MRML/Widgets/Testing/qMRMLNodeComboBoxTest8.cxx
+++ b/Libs/MRML/Widgets/Testing/qMRMLNodeComboBoxTest8.cxx
@@ -104,6 +104,7 @@ int qMRMLNodeComboBoxTest8(int argc, char* argv[])
 
   // test a conflict with one of the default actions
   QAction* action2 = new QAction("Create new type of action that conflicts with create new node", &nodeSelector);
+  action2->setData("CREATE/vtkMRMLScalarVolumeNode");
   nodeSelector.addMenuAction(action2);
 
   actionsPlusOne = sceneModel->postItems(sceneModel->mrmlSceneItem()).size();
@@ -225,18 +226,22 @@ int qMRMLNodeComboBoxTest8(int argc, char* argv[])
   }
 
   // add a custom actions
-  const QStringList actionPrefixes = {
-    "Create new ",
-    "Delete current ",
-    "Edit current ",
-    "Rename current ",
+  const QStringList actionPrefixesData = {
+    "Create new |CREATE",
+    "Delete current |DELETE",
+    "Edit current |EDIT",
+    "Rename current |RENAME",
   };
-  for (const QString& actionPrefix : actionPrefixes)
+  for (const QString& actionPrefixData : actionPrefixesData)
   {
+    QStringList parts = actionPrefixData.split("|");
+    QString actionPrefix = parts[0];
+    QString actionData = parts[1];
     startingActions = sceneModel->postItems(sceneModel->mrmlSceneItem()).size();
 
     QString actionName = QString("%1node using custom action").arg(actionPrefix);
     QAction* action = new QAction(actionName, &nodeSelector);
+    action->setData(actionData);
     nodeSelector.addMenuAction(action);
 
     expected = startingActions + 1;
@@ -341,6 +346,7 @@ int qMRMLNodeComboBoxTest8(int argc, char* argv[])
   qMRMLSceneModel* sceneModel2 = nodeSelector2.sceneModel();
 
   QAction* action = new QAction("Rename current node using custom action", &nodeSelector2);
+  action->setData("RENAME");
 
   startingActions = sceneModel2->postItems(sceneModel2->mrmlSceneItem()).size();
   nodeSelector2.addMenuAction(action);
--- a/Libs/MRML/Widgets/qMRMLNodeComboBox.cxx
+++ b/Libs/MRML/Widgets/qMRMLNodeComboBox.cxx
@@ -369,7 +369,7 @@ void qMRMLNodeComboBoxPrivate::updateDelegate(bool force)
 // --------------------------------------------------------------------------
 bool qMRMLNodeComboBoxPrivate::hasPostItem(const QString& name) const
 {
-  for (const QString& item : this->MRMLSceneModel->postItems(this->MRMLSceneModel->mrmlSceneItem()))
+  for (const QString& item : this->MRMLSceneModel->postItemsData(this->MRMLSceneModel->mrmlSceneItem()))
   {
     if (item.startsWith(name))
     {
--- a/Libs/MRML/CLI/vtkMRMLCommandLineModuleNode.cxx
+++ b/Libs/MRML/CLI/vtkMRMLCommandLineModuleNode.cxx
@@ -92,6 +92,8 @@ vtkMRMLNode* vtkMRMLCommandLineModuleNode::CreateNodeInstance()
 //----------------------------------------------------------------------------
 vtkMRMLCommandLineModuleNode::vtkMRMLCommandLineModuleNode()
 {
+  this->TypeDisplayName = vtkMRMLTr("vtkMRMLCommandLineModuleNode", "Command Line Module");
+
   this->Internal = new vtkInternal();
   this->HideFromEditors = true;
   this->Internal->Status = vtkMRMLCommandLineModuleNode::Idle;
--- a/Libs/MRML/CLI/vtkMRMLCommandLineModuleNode.h
+++ b/Libs/MRML/CLI/vtkMRMLCommandLineModuleNode.h
@@ -45,10 +45,6 @@ public:
   /// Get node XML tag name (like Volume, Model)
   const char* GetNodeTagName() override { return "CommandLineModule"; }
 
-  /// Get node type to be displayed to the user.
-  /// It is translated to the application language.
-  virtual std::string GetTypeDisplayName() override { return vtkMRMLTr("vtkMRMLCommandLineModuleNode", "Command Line Module"); };
-
   /// List of events that can be fired on or by the node.
   enum CLINodeEvent
   {
--- a/Libs/MRML/Core/vtkMRMLBSplineTransformNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLBSplineTransformNode.cxx
@@ -25,6 +25,8 @@ vtkMRMLNodeNewMacro(vtkMRMLBSplineTransformNode);
 //----------------------------------------------------------------------------
 vtkMRMLBSplineTransformNode::vtkMRMLBSplineTransformNode()
 {
+  this->TypeDisplayName = vtkMRMLTr("vtkMRMLBSplineTransformNode", "B-Spline Transform");
+
   // Set up the node with a dummy bspline grid (that contains a small set of
   // null-vectors) to make sure the node is valid and can be saved
   double gridSize[3] = { 4, 4, 4 };
--- a/Libs/MRML/Core/vtkMRMLBSplineTransformNode.h
+++ b/Libs/MRML/Core/vtkMRMLBSplineTransformNode.h
@@ -49,10 +49,6 @@ public:
   /// Get node XML tag name (like Volume, Model)
   const char* GetNodeTagName() override { return "BSplineTransform"; };
 
-  /// Get node type to be displayed to the user.
-  /// It is translated to the application language.
-  std::string GetTypeDisplayName() override { return vtkMRMLTr("vtkMRMLBSplineTransformNode", "B-Spline Transform"); };
-
   ///
   /// Create default storage node or nullptr if does not have one
   vtkMRMLStorageNode* CreateDefaultStorageNode() override { return Superclass::CreateDefaultStorageNode(); };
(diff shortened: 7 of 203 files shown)


jamesobutler:
I have added an additional commit which takes the code snippets from the original post in this PR and puts it into an automated test.

Libs/MRML/Widgets/qMRMLNodeComboBox.cxx
```diff
@@ -631,7 +631,7 @@ vtkMRMLNode* qMRMLNodeComboBox::addNode(QString nodeType)
     return nullptr;
   }
   // Create the MRML node via the MRML Scene
-  vtkMRMLNode * newNode = d->MRMLNodeFactory->createNode(nodeType);
+  vtkMRMLNode* newNode = vtkMRMLNode::SafeDownCast(this->mrmlScene()->AddNewNodeByClass(nodeType.toStdString()));
```
jcfr, line 634: Before removing the use of the `qMRMLNodeFactory`, we need to make sure that support for associated features (i.e. `addAttribute/removeAttribute/.../baseName`) and signals (i.e. `nodeInstantiated/nodeInitialized/nodeAdded`) are either obsolete or implemented independently.

jcfr (changes requested):

to avoid blocking the integration due to comment related to the use of the node factory, it may be sensible to integrate changes independently

Libs/MRML/Widgets/qMRMLNodeComboBox.cxx
lassoan: We reviewed this code quickly with @jcfr and found that `MRMLNodeFactory` already uses `CreateNewNodeByClass`, so there should be no need to switch from `MRMLNodeFactory->createNode`.

jamesobutler: I forgot to include qMRMLNodeComboBox node creation to my testing commit. I have pushed an update to include testing for this.

It appears this code change is necessary as if I add a default node such as:
```python
# Define a default Markups Line node with a default node name
default_markups_line_node = slicer.vtkMRMLMarkupsLineNode()
default_markups_line_node.SetName("DefaultNodeName")  # Default names are to be used exactly and to not be manipulated through GenerateUniqueName()
slicer.mrmlScene.AddDefaultNode(default_markups_line_node)
```
and then try to create a Line node through a qMRMLNodeComboBox, it is not respecting my default node defined set name:
```python
markup_line_node_combobox = slicer.qMRMLNodeComboBox()
markup_line_node_combobox.setMRMLScene(slicer.mrmlScene)
markup_line_node_combobox.nodeTypes = ["vtkMRMLMarkupsLineNode"]
markup_node = markup_line_node_combobox.addNode()
if markup_node.GetName() != "DefaultNodeName":
    raise Exception("Unexpected node name: " + markup_node.GetName())
```

lassoan: Maybe if the baseName is empty then the combobox should use the default node for determining the base name?

jamesobutler: As of right now if the Default Node has a name specified it uses that given name exactly. If there is no default specified then it falls back onto the node name prefix like “L” which iterates with the GenerateUniqueName call.

jamesobutler: Is the Markups `DefaultNodeNamePrefix` doing the same thing as the Node factory’s `baseName`? Where the base name is used followed by generating a unique name by appending a numerical value?

lassoan: `DefaultNodeNamePrefix` is not capable of what the node factory is doing, because `DefaultNodeNamePrefix` can only specify a single base name (for each node type) in the entire application.

The node factory is more versatile, because it allows each GUI widget to specify its own base name (and the base name can be a different for each node type).

The only thing I see that can be considered inconsistent is that if the developer forgets to set the basename in the combobox then the node factory in the combobox falls back to `GetTagByClassName` output, instead of `DefaultNodeNamePrefix`. This is not ideal, because the tag is not translatable (hardcoded to English). However, the node factory is generic - does not know about special functions that should be used for a particular node type.

jamesobutler: It appears the use of NodeFactory node creation and customizations were introduced around 2010 (in https://github.com/Slicer/Slicer/commit/77cdc413e9027f8e3ed5993e5a4584a960631e94 and https://github.com/Slicer/Slicer/commit/704faf76a22e136ec7b9109082f250d454764866). While the MRML Scene supporting default nodes was introduced around 2016 (in https://github.com/Slicer/Slicer/commit/4dfe27653dc2d3712c34848638e1c6b4f3e7fb78).

> Maybe if the baseName is empty then the combobox should use the default node for determining the base name?

@lassoan @jcfr:
Should a widget level node defined customization override a MRML scene defined node default?
vs
Should a MRML scene defined node default override a widget level node defined customization?

I'm struggling to determine which should take precedence. It appears that the Node Factory is not specifically related to a given widget as you can instantiate a node factory on its own and use it to add nodes into the scene. So then it results in 2 different ways to define defaults of nodes being added into the scene.

What started the work in this PR was that I was originally thinking if I am adding a default node to the scene, such as `slicer.mrmlScene.AddDefaultNode(..)`, I am making the decision for node creation of this type to always follow this customization.

lassoan: > Should a widget level node defined customization override a MRML scene defined node default?

Yes, definitely. Each widget specifies the relevant default names. For example, if a module computes a set of start and end points: baseName in the start point selector widget is to StartPoints, the in the end point selector widget set to EndPoints.

> It appears that the Node Factory is not specifically related to a given widget 

Each node selector combobox has its own factory, so they are very tightly linked. The node factory stores the preferred base name (and property values to add) for each node type. This is a bit similar to how default unit settings are overridden in widgets. I think the only missing part is that if the widget does not specify baseName then the baseName could be determined from the default node.

jamesobutler: > I think the only missing part is that if the widget does not specify baseName then the baseName could be determined from the default node.

I have pushed an update in `ENH: Use default node name when no baseName for qMRMLNodeFactory` to use the default node name if no base name is specified in the node factory.

jamesobutler:
@jcfr Any additional comments here? You have an active "Requested Change" action applied on this PR.

jamesobutler:
@jcfr I've rebased this PR based on recent clang-format changes to `main`. Since you have requested changes on this PR, would you be willing to re-review to help with this integration? Thanks!

Libs/MRML/Core/vtkMRMLMarkupsNode.h
```diff
@@ -148,10 +148,16 @@ class VTK_MRML_EXPORT vtkMRMLMarkupsNode : public vtkMRMLDisplayableNode
   // MRMLNode methods
   //--------------------------------------------------------------------------
 
+  /// Get node XML tag name (like Volume, Model)
+  const char* GetNodeTagName() override { return this->GetDefaultNodeNamePrefix(); };
```
lassoan: `NodeTagName` and `DefaultNodeNamePrefix` are very different. `NodeTagName` is used for very low level features in the MRML scene and it is not a user-displayable or translatable string. DefaultNodeNamePrefix is user-displayable and translatable. Therefore, we have to keep them as separate values.

It is confusing, because the i18n implementation of the node combobox menu is incomplete. The fix is in [this PR](https://github.com/Slicer/Slicer/pull/8091) that unfortunately still needs some cleanup. I should be able to fix that within a few days, but the PR already shows the idea of how the node type label should be genereated.

jamesobutler: @lassoan Thanks for the review and clarification. I've updated my commit so that `NodeTagName` is kept as the internal non-translatable usage for logic and instead add `DefaultNodeNamePrefix` at the vtkMRMLNode level to use as the translatable thing that can be updated. By default `DefaultNodeNamePrefix` points to `NodeTagName`, but it allows for customization.

I have confirmed locally that my new `MarkupsNodeNameTest` is passing as intended.

jamesobutler: cc: @lassoan to see if you can help with the final review and integration here.

lassoan: I've got https://github.com/Slicer/Slicer/pull/8091 ready, it'll be merged as soon as CI completes.

I've reviewed this PR and it all looks good to me except I don't think the `DefaultNodeNamePrefix` should be written to file. If it was written to file then a scene created in Slicer set to English would always create English node names, instead of using the current application language. It would also pollute the scene file.

I'm just curious, what is the use case for setting an exact node name for new nodes? Wouldn't it create confusion by using the exact same name for many different nodes?

jamesobutler: In our usage we don’t have same name confusion issues in say a qMRMLNodeComboBox because we usually have some attribute filter applied that doesn’t result in that case. We often have nodes under a subject hierarchy where a volume node is the parent for a bunch of child nodes that we keep up with and sometimes want to use similar node names for the child nodes.

lassoan:
I think the best would be to harmonize the implementation of TypeDisplayName and DefaultNodeNamePrefix, something like this: https://github.com/lassoan/Slicer/commit/3416c1b28c7396d9d3d47a415dca172701a3e7d1

If the change looks good to you then I'll rebase on the latest main version and push here.

jamesobutler:
Would I still be able to customize the prefix/display name? For example instead of a vtkMRMLMarkupsLineNode utilizing "Line" I would customize it to either "Ruler" or "Caliper" or like the shortened prefix of "R" or "C" for "R_1" and "C_1" type usage. That would then mean "Ruler" or "Caliper" would be the translatable name based on my customization.

lassoan:
> Would I still be able to customize the prefix/display name? For example instead of a vtkMRMLMarkupsLineNode utilizing "Line" I would customize it to either "Ruler" or "Caliper" or like the shortened prefix of "R" or "C" for "R_1" and "C_1" type usage. That would then mean "Ruler" or "Caliper" would be the translatable name based on my customization

Yes. The change is just generalization of earlier markups-specific mechanism (use different string for default node name than the full node type) and harmonize the implementation (use a member variable instead of requiring overriding a method). In addition to this, as you implemented, automatic node name generation can be avoided by setting node name in the default node.

lassoan (approved):
This should be all good now. @jamesobutler please check if this works well for you.

lassoan (approved):

lassoan:
@jcfr It looks like your concerns have been addressed. Please check and approve if looks good.

Libs/MRML/Widgets/Testing/qMRMLNodeComboBoxTest8.cxx
```diff
@@ -225,18 +226,17 @@ int qMRMLNodeComboBoxTest8(int argc, char* argv[])
...
   // add a custom actions
-  const QStringList actionPrefixes = {
-    "Create new ",
-    "Delete current ",
-    "Edit current ",
-    "Rename current ",
-  };
-  for (const QString& actionPrefix : actionPrefixes)
+  const QStringList actionPrefixesData = { "Create new |CREATE", "Delete current |DELETE", "Edit current |EDIT", "Rename current |RENAME" };
+  for (const QString& actionPrefix : actionPrefixesData)
```
lassoan: I'll fix this by the end of the day.

jamesobutler:
@lassoan Yes your latest changes look good. You can fix up the build error and re-provide review.

Libs/MRML/Widgets/Testing/qMRMLNodeComboBoxTest8.cxx
jcfr: I just addressed the error and updated the topic :rocket:

jcfr (approved):

jamesobutler:
Thanks for the reviews and updates 🙏🏻

PR #8422, opened 2025-05-18
BUG: Fix reordering of subject hierarchy nodes

### LINKED ISSUE #8421: Ordering segments in the subject hierarchy does not change the order in the Segmentations/Segment Editor modules
## Summary
Ordering of the segments determine how segments are going to be flattened in order to export a labelmap. You can reorder the segments easily by dragging them in the SH, but in the segment editor one has to click "move segment up/down" many times (if the segmentation is complex with multiple segments). Therefore it is easier to the reorganization in the SH.

However, it appears that if you change the order in SH, the order in the segmentation (or segment editor) module is not updated. This creates an inconsistency in the workflow and room for confusion and error. It is also true for visibility of individual segments.

It is not clear to me if the order in SH is preserved when i right-click and export a labelmap (as opposed to doing it through the segmentation module). 

This behavior needs to be correct. 

**Easy solution**: Do not allow the segments to be re-ordered in the SH. 
**Better solution:** Implement functionality to keep the state in SH and segmentation module to be in sync (order, visibility, etc)

lassoan:
Subject hierarchy item order are now synchronized between:
- subject hierarchy singleton node in the scene
- subject hierarchy item model of all subject hierarchy Widgets
- segments of each segmentation node

This is implemented by two-way synchronization between: qMRMLSubjectHierarchyModel <-> subject hierarchy node (in qMRMLSubjectHierarchyModel) and subject hierarchy node <-> segmentation nodes (in qSlicerSubjectHierarchySegmentationsPlugin).

fixes #8421

@muratmaga

### DIFF
--- a/Libs/MRML/Core/vtkMRMLSubjectHierarchyNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLSubjectHierarchyNode.cxx
@@ -1031,24 +1031,27 @@ bool vtkSubjectHierarchyItem::Move(vtkSubjectHierarchyItem* beforeItem)
   if (!beforeItem)
   {
     this->Parent->Children.push_back(thisPointer);
-    return true;
+  }
+  else
+  {
+    ChildVector::iterator beforeIt;
+    for (beforeIt = this->Parent->Children.begin(); beforeIt != this->Parent->Children.end(); ++beforeIt)
+    {
+      if (beforeItem == beforeIt->GetPointer())
+      {
+        break;
+      }
+    }
+    if (beforeIt == this->Parent->Children.end())
+    {
+      vtkErrorMacro("Move: Failed to find subject hierarchy item '" << beforeItem->GetName() << "' as insertion position in item '" << this->Parent->GetName() << "'");
+      return false;
+    }
+    this->Parent->Children.insert(beforeIt, thisPointer);
   }
 
-  ChildVector::iterator beforeIt;
-  for (beforeIt = this->Parent->Children.begin(); beforeIt != this->Parent->Children.end(); ++beforeIt)
-  {
-    if (beforeItem == beforeIt->GetPointer())
-    {
-      break;
-    }
-  }
-  if (beforeIt == this->Parent->Children.end())
-  {
-    vtkErrorMacro("Move: Failed to find subject hierarchy item '" << beforeItem->GetName() << "' as insertion position in item '" << this->Parent->GetName() << "'");
-    return false;
-  }
-  this->Parent->Children.insert(beforeIt, thisPointer);
   this->Parent->Modified();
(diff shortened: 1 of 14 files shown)


muratmaga:
Does this also synch the visibility settings?

lassoan:
Segment visibility settings has been kept in sync between subject hierarchy tree widgets and segment list widgets. If you find any problems then please submit a new issue. Thanks!

jcfr (changes requested):
Thanks for working on this 🙏 

After testing locally, I identified few issues that would be nice to address.

### Moving one segment at the end does not work

| Before | After | Expected |
|--|--|--|
| Image | Image | Segment 1 should be at the end |

### Moving multiple segments is not working

| Before | After | Expected |
|--|--|--|
| image | image  |   Segments should be listed as 3, 1, 2 and 4| 

### Moving multiple segments at the end is not working

| Before | After | Expected |
|--|--|--|
| Image | image | Segment 1 and 2 should be at the end |

lassoan:
I've fixed all the reported issues (moving items to the end, moving multiple items). @jcfr please test again. Thank you!

jcfr:
Thanks all reported issues are now address :100: 

One last nitpick, is it expected to be able to move a segment outside of segmentation ?

image

cpinter:
No, segments are only allowed to be under a segmentation.

lassoan:
I'll update the PR, I think I know what is missing.

muratmaga:
Any plan to merge this?

jcfr:
> Any plan to merge this?

Once the last issue[^1] is addressed, it should be ready for integration  :rocket: 

Alternatively, we could create an issue report and address the last "issue" in a follow-up pull request.

cc: @lassoan 

[^1]: https://github.com/Slicer/Slicer/pull/8422#issuecomment-2896755606

lassoan:
It is almost ready, I had to make a lot of changes to fix the remaining issues.

lassoan:
> One last nitpick, is it expected to be able to move a segment outside of segmentation ?

This is fixed now. Everything should work well. All tests that passed before are passing now.

@jcfr please check it out and approve if looks good to you.

lassoan:
I plan to merge this later today if I don't get any further comments.

Libs/MRML/Core/vtkMRMLSubjectHierarchyNode.cxx
```diff
@@ -3751,3 +3758,107 @@ void vtkMRMLSubjectHierarchyNode::ShowItemsInView(vtkIdList* itemIDs, vtkMRMLAbs
...
+    // Add item to new parent
+    childItem->Parent = parentItem;
+    parentItem->Children.push_back(thisPointer);
+
+    modifiedParentItems.insert(parentItem);
+    modifiedParentItems.insert(formerParentItem);
+  }
+
+  // Update order of child items
+
```
jcfr:
```suggestion
```

PR #8428, opened 2025-05-19
ENH: Remove Slicer mention from Markups save to default display tooltip

jamesobutler:
Removing the "Slicer" mention is to support the usage in a Slicer custom application which generally has a different application name.

This string exists in a UI file, so I re-worded it to remove the application name entirely rather than setting the tooltip from the code and utilizing `slicer.app.mainApplicationName`.

### DIFF
--- a/Modules/Loadable/Markups/Resources/UI/qSlicerMarkupsModule.ui
+++ b/Modules/Loadable/Markups/Resources/UI/qSlicerMarkupsModule.ui
@@ -110,7 +110,7 @@
           <item>
            <widget class="QPushButton" name="saveToDefaultDisplayPropertiesPushButton">
             <property name="toolTip">
-             <string>Save current display property settings to defaults, will be saved for when Slicer restarts (see Application Settings)</string>
+             <string>Save current display properties to defaults. These properties will be used even after application restart.</string>
             </property>
             <property name="text">
              <string>Save to Defaults</string>


Modules/Loadable/Markups/Resources/UI/qSlicerMarkupsModule.ui
```diff
@@ -110,7 +110,7 @@
           <item>
            <widget class="QPushButton" name="saveToDefaultDisplayPropertiesPushButton">
             <property name="toolTip">
-             <string>Save current display property settings to defaults, will be saved for when Slicer restarts (see Application Settings)</string>
+             <string>Save current display properties to defaults.</string>
```
jcfr, line 113:
```suggestion
             <string>Save current display properties to defaults for future sessions (see Application Settings).</string>
```

jcfr: @jamesobutler Does it make sense to keep the reference to "Application Settings" ?

lassoan: Instead of "Slicer" we can use "application". Something like this:

```suggestion
             <string>Save current display properties to application settings. These settings will be used as default after the application is restarted.</string>
```

jamesobutler: I noticed that the settings are immediately used and does not need the application to be restarted. Is there a need to re-clarify that these are saved in a way that are re-used in future instances of the application?

lassoan: Maybe it is enough to say "Save current display properties to application settings."?

jamesobutler: I was confused by the mention of "Application Settings" because when I went to "Edit->Application Settings" there is no panel associated with the Markups module with anything about markups display properties.

lassoan: The data is stored in "application settings", but if this is confusing then we can describe the effect - data is preserved after application restart, persistent, etc.

jamesobutler: I've pushed an update to:
"Save current display properties to defaults. These properties will be used even after application restart."

lassoan (approved):
Looks good to me.

PR #8430, opened 2025-05-19
ENH: Make it easier to render large volumes on macOS

lassoan:
On macOS graphics hardware, maximum 3D texture size (along any axis) is typically 2048. This commit sets volume partitioning partitioned to not exceed this size by default.

On other systems volume is maximum size is set to 4096 by default, which will not split up the volume in most cases (which was the behavior before this commit).

### DIFF
--- a/Modules/Loadable/VolumeRendering/MRMLDM/vtkMRMLVolumeRenderingDisplayableManager.h
+++ b/Modules/Loadable/VolumeRendering/MRMLDM/vtkMRMLVolumeRenderingDisplayableManager.h
@@ -65,6 +65,15 @@ public:
   /// Get the MRML ID of the picked node, returns empty string if no pick
   const char* GetPickedNodeID() override;
 
+  /// @{
+  /// Experimental function to adjust the maximum 3D texture size.
+  /// Along each axis, the volume will be split up so that each partition is not larger than this maximum size.
+  /// Default is 2048 on macOS (as Apple hardware typically is limited to this maximum texture size)
+  /// and 4096 on other systems (so that most volumes will not be split up by default).
+  static int GetMaximum3DTextureSize();
+  static void SetMaximum3DTextureSize(int size);
+  /// @}
+
 public:
   static int DefaultGPUMemorySize;
 
@@ -84,6 +93,7 @@ protected:
 
 protected:
   vtkSlicerVolumeRenderingLogic *VolumeRenderingLogic{nullptr};
+  static int Maximum3DTextureSize;
 
 protected:
   vtkMRMLVolumeRenderingDisplayableManager(const vtkMRMLVolumeRenderingDisplayableManager&); // Not implemented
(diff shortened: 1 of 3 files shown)


Modules/Loadable/VolumeRendering/MRMLDM/vtkMRMLVolumeRenderingDisplayableManager.h
```diff
@@ -65,6 +65,15 @@ class VTK_SLICER_VOLUMERENDERING_MODULE_MRMLDISPLAYABLEMANAGER_EXPORT vtkMRMLVol
   /// Get the MRML ID of the picked node, returns empty string if no pick
   const char* GetPickedNodeID() override;
 
+  /// @{
+  /// Experimental function to adjust the maximum 3D texture size.
+  /// Along each axis, the volume will be split up so that each partition is not larger than this maximum size.
+  /// Default is 2048 on macOS (as Apple hardware typically is limited to this maximum texture size)
```
jamesobutler, line 71: Where do you go to find that 2048 is the maximum texture on your Apple hardware?

lassoan: It is based on experimenting on Apple user's computers with the maximum volume size that can be rendered without setting partitions (and then seeing that setting partition>1 fixes the rendering problem). If anyone shows up with an Apple computer where a higher value is supported then we can increase the value.

jamesobutler: It looks like the value of 2048 matches what is listed for “Maximum 3D texture width, height, and depth”

https://developer.apple.com/metal/Metal-Feature-Set-Tables.pdf

lassoan: image

lassoan: Nice find! It is possible in theory that Apple will increase this value, but most likely they will not touch their OpenGL support. Also, VTK's new webGPU backend may implement automatic partitioning. So, the current implementation would be needed probably only for a couple of years.

jamesobutler (dismissed):

jcfr:
I am updating the topic to include the following fix:

```diff
- #include <vtkOpenGLGPUVolumeRaycastMapper.h>
+ #include <vtkOpenGLGPUVolumeRayCastMapper.h>
```

jcfr:
Also replace `unsigned short()` with `static_cast<unsigned short>()` to address error like the following:

```
/path/to/Slicer/Modules/Loadable/VolumeRendering/MRMLDM/vtkMRMLVolumeRenderingDisplayableManager.cxx:1056:7: error: expected primary-expression before ‘unsigned’
 1056 |       unsigned short(dims[0] / vtkMRMLVolumeRenderingDisplayableManager::Maximum3DTextureSize + 1),

```

jcfr (approved):

jcfr (approved):

PR #8433, opened 2025-05-20
ENH: Remove old deprecation warning

jamesobutler:
As indicated by the comment, enough time has passed for developers to update their code accordingly as part of understanding that vtkMRMLScalarVolumeNode does not have a LabelMap attribute anymore.

This warning was originally added in 2015 through https://github.com/Slicer/Slicer/commit/d071d05cd7c4ef91eeced7a9c1ef003bf9751710.

### DIFF
--- a/Libs/MRML/Widgets/qMRMLNodeComboBox.cxx
+++ b/Libs/MRML/Widgets/qMRMLNodeComboBox.cxx
@@ -509,21 +509,6 @@ void qMRMLNodeComboBox::addAttribute(const QString& nodeType,
                                      const QVariant& attributeValue)
 {
   Q_D(qMRMLNodeComboBox);
-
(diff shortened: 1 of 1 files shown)


jcfr:
I was initially thinking we should also remove the reading of the `LabelMap` attribute ...

https://github.com/Slicer/Slicer/blob/673ddca536801eedfdc3287dbc6ab110c494382c/Libs/MRML/Core/vtkMRMLScalarVolumeNode.cxx#L76-L87

that said, it is only used to allow the creation of `vtkMRMLLabelMapVolumeNode`:


https://github.com/Slicer/Slicer/blob/673ddca536801eedfdc3287dbc6ab110c494382c/Libs/MRML/Core/vtkMRMLParser.cxx#L160-L176

jcfr (approved):

PR #8434, opened 2025-05-20
BUG: Fix basename not incrementing for AddNewNodeByClass

jamesobutler:
As originally pointed out in https://discourse.slicer.org/t/addnewnodebyclass-basename-does-not-increment/34386.

This logic setting the name in `AddNewNodeByClass` has seemingly been the same since it was introduced in https://github.com/Slicer/Slicer/commit/31551523b242a03c0f8f5fa17999f0eb7b7cd360. This PR has been made under the assumption that the input argument of "nodeBaseName" means that the intention is for the specified name to serve as a base and increment to provide a unique name.

https://github.com/Slicer/Slicer/blob/673ddca536801eedfdc3287dbc6ab110c494382c/Libs/MRML/Core/vtkMRMLScene.h#L211-L214

It appears that developers have worked around this issue of the name not incrementing by calling `GenerateUniqueName` manually.
https://github.com/Slicer/Slicer/blob/673ddca536801eedfdc3287dbc6ab110c494382c/Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py#L460-L464
https://github.com/Slicer/Slicer/blob/673ddca536801eedfdc3287dbc6ab110c494382c/Modules/Scripted/Endoscopy/Endoscopy.py#L878-L881

However in some places it is unclear if the specified `nodeBaseName` has instead been used as a `nodeName` with expectation that it would not increment and be used exactly as specified. If this is the desire then we could instead change the `AddNewNodeByClass` method to name the input argument `nodeName` instead of `nodeBaseName`.
https://github.com/Slicer/Slicer/blob/673ddca536801eedfdc3287dbc6ab110c494382c/Docs/developer_guide/script_repository/markups.md?plain=1#L687-L689
https://github.com/Slicer/Slicer/blob/673ddca536801eedfdc3287dbc6ab110c494382c/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsJsonStorageNode.cxx#L168-L177

### DIFF
--- a/Libs/MRML/Core/vtkMRMLScene.cxx
+++ b/Libs/MRML/Core/vtkMRMLScene.cxx
@@ -1453,7 +1453,7 @@ vtkMRMLNode* vtkMRMLScene::AddNewNodeByClass(
   }
   if (!nodeBaseName.empty())
   {
-    nodeToAdd->SetName(nodeBaseName.c_str());
+    nodeToAdd->SetName(this->GenerateUniqueName(nodeBaseName).c_str());
   }
   return this->AddNode(nodeToAdd);
 }
@@ -1499,7 +1499,7 @@ vtkMRMLNode* vtkMRMLScene::AddNewNodeByClassWithID(std::string className, std::s
   nodeToAdd->SetID(nodeID.c_str());
   if (!nodeBaseName.empty())
   {
-    nodeToAdd->SetName(nodeBaseName.c_str());
+    nodeToAdd->SetName(this->GenerateUniqueName(nodeBaseName).c_str());
   }
 
   return this->AddNode(nodeToAdd);


jamesobutler:
My guess is the `nodeBaseName` usage in AddNewNodeByClass was trying to mimic the `baseName` usage of the qMRMLNodeFactory which has a `createNode` method that will generate a unique name from the provided basename.
https://github.com/Slicer/Slicer/blob/673ddca536801eedfdc3287dbc6ab110c494382c/Libs/MRML/Widgets/qMRMLNodeFactory.cxx#L83-L93

Libs/MRML/Core/vtkMRMLScene.cxx
```diff
@@ -1453,7 +1453,7 @@ vtkMRMLNode* vtkMRMLScene::AddNewNodeByClass(
   }
   if (!nodeBaseName.empty())
   {
-    nodeToAdd->SetName(nodeBaseName.c_str());
+    nodeToAdd->SetName(this->GenerateUniqueName(nodeBaseName).c_str());
```
jcfr, line 1456: This makes sense. That said, I think existing code like the one you referenced in the pull request description will end up generating name like `<BaseName>_N_N`

for this reason, I am wondering if we should:
1. add a new parameter called `generateUniqueName=false`
2. report a warning if the parameter is not specified & hint user at updating and simplifying their code

Then in few years we would change the default behavior to always generate a unique name ...

cc: @lassoan

jamesobutler: > I think existing code like the one you referenced in the pull request description will end up generating name like `<BaseName>_N_N`

This doesn't appear to be the case as a unique name is only generated if there is one in the scene already. It doesn't iterate the name every time.
```python
slicer.mrmlScene.GenerateUniqueName(slicer.mrmlScene.GenerateUniqueName(slicer.mrmlScene.GenerateUniqueName("First")))
```
^ Produces "First_1" and not "First_1_1_1".
```python
slicer.mrmlScene.Clear()
slicer.mrmlScene.AddNewNodeByClass("vtkMRMLMarkupsLineNode", slicer.mrmlScene.GenerateUniqueName("MyBaseName"))
slicer.mrmlScene.AddNewNodeByClass("vtkMRMLMarkupsLineNode", slicer.mrmlScene.GenerateUniqueName("MyBaseName"))
```
^ Produces node names of "MyBaseName_1" and "MyBaseName_2".

jamesobutler: > This doesn't appear to be the case as a unique name is only generated if there is one in the scene already.

I'm actually wrong in that it is always checking the scene if there is a node of the same name present. Instead it is checking a mapper to see if the name has been produced before, but it doesn't mean that the name is actually being used in the scene currently.
https://github.com/Slicer/Slicer/blob/673ddca536801eedfdc3287dbc6ab110c494382c/Libs/MRML/Core/vtkMRMLScene.cxx#L2603-L2604

This has been discussed previously in https://discourse.slicer.org/t/superfluous-suffix-for-reloads-causes-issues-with-save/13216. Murat had asked:

> "For the particular case I present (i.e., a node is loaded, then deleted, and reloaded) would it be possible the not to add the suffix?"

jamesobutler: > However in some places it is unclear if the specified `nodeBaseName` has instead been used as a `nodeName` with expectation that it would not increment and be used exactly as specified. If this is the desire then we could instead change the `AddNewNodeByClass` method to name the input argument `nodeName` instead of `nodeBaseName`.

I'm inclined to just change the documentation to match how it has been used (`nodeName`, not a `baseName`) where you have to manually use `GenerateUniqueName` for the second input argument of `AddNewNodeByClass`
```python
slicer.mrmlScene.Clear()
slicer.mrmlScene.AddNewNodeByClass("vtkMRMLMarkupsLineNode", slicer.mrmlScene.GenerateUniqueName("MyBaseName"))
slicer.mrmlScene.AddNewNodeByClass("vtkMRMLMarkupsLineNode", slicer.mrmlScene.GenerateUniqueName("MyBaseName"))
```
to get the same type output as:
```python
slicer.mrmlScene.Clear()
node_factory = slicer.qMRMLNodeFactory()
node_factory.setMRMLScene(slicer.mrmlScene)
node_factory.setBaseName("vtkMRMLMarkupsLineNode", "MyBaseName")
node_factory.createNode("vtkMRMLMarkupsLineNode")
node_factory.createNode("vtkMRMLMarkupsLineNode")
```

jamesobutler:
Closing this now. See https://github.com/Slicer/Slicer/pull/8434/files#r2098964573.

PR #8438, opened 2025-05-21
ENH: Add fast clipping shortcut for volume rendering

Sunderlandkyl:
This commit adds a separate path for clipping using the existing vtkAbstractMapper::SetClippingPlanes pathway, when it is compatible. This requires that all of the below conditions are true:
  - Clip type is union
  - Clipping functions contain planes only
  - There are fewer than 7 clipping planes (ROI nodes count as 6 planes)
  - ROI node clipping is negative
  - Soft edge is disabled
  - Cropping is disabled

If any of the conditions are not met, then the slower exiting pathway will be used.

Users can see additional information on the clipping pipeline by clicking on the icon beside the "Clipping" checkbox.

image

### DIFF
--- a/Modules/Loadable/VolumeRendering/MRML/vtkMRMLVolumeRenderingDisplayNode.h
+++ b/Modules/Loadable/VolumeRendering/MRML/vtkMRMLVolumeRenderingDisplayNode.h
@@ -131,7 +131,6 @@ protected:
 
   void ProcessMRMLEvents(vtkObject *caller, unsigned long event, void *callData) override;
 
-  //@{
   /// Check if a fast clipping method can be used with the specified clip node.
   /// Returns true if fast clipping can be utilized, or returns false otherwise.
   /// If userMessages is specified, messages will be added to provide reasons for why fast clipping


Modules/Loadable/VolumeRendering/MRML/vtkMRMLVolumeRenderingDisplayNode.cxx
```diff
@@ -333,3 +337,101 @@ void vtkMRMLVolumeRenderingDisplayNode::ProcessMRMLEvents(vtkObject *caller,
...
+  if (!this->IsClippingShortcutValid(this->GetClipNode(), userMessages))
+  {
+    shortcutValid = false;
+  }
+
+  return shortcutValid;
+}
+
+//---------------------------------------------------------------------------
+bool vtkMRMLVolumeRenderingDisplayNode::IsClippingShortcutValid(vtkMRMLClipNode* clipNode, vtkMRMLMessageCollection* userMessages/*=nullptr*/)
```
lassoan, line 372: "Valid shortcut" name could be somewhat confusing. Something like "fast clipping available" would be more clear.

```suggestion
bool vtkMRMLVolumeRenderingDisplayNode::IsFastClippingAvailable(vtkMRMLClipNode* clipNode, vtkMRMLMessageCollection* userMessages/*=nullptr*/)
```

Modules/Loadable/VolumeRendering/MRML/vtkMRMLVolumeRenderingDisplayNode.h
```diff
@@ -114,6 +115,15 @@ class VTK_SLICER_VOLUMERENDERING_MODULE_MRML_EXPORT vtkMRMLVolumeRenderingDispla
   vtkBooleanMacro(AutoClippingBlankVoxelValue, bool);
   //@}
 
+  //@{
+  /// Check if the clipping shortcut can be used with the display/clipping node.
+  /// Returns true if the clipping shortcut can be utilized, or returns false otherwise.
```
lassoan, line 120: In general, I would avoid using "shortcut" word, as it is not very clear. The important thing is that if the shortcut can be used then the clipping will be _fast_.

Libs/MRML/Core/vtkMRMLClipNode.cxx
```diff
@@ -688,3 +688,68 @@ void vtkMRMLClipNode::SetYellowSliceClipState(int state)
...
 }
+
+//----------------------------------------------------------------------------
+bool vtkMRMLClipNode::GetClippingPlanes(vtkPlaneCollection* planeCollection)
+{
+  return this->GetClippingPlanesFromFunction(this->ImplicitFunction, planeCollection);
+}
+
+//---------------------------------------------------------------------------
+bool vtkMRMLClipNode::GetClippingPlanesFromFunction(vtkImplicitFunction* function, vtkPlaneCollection* planeCollection, bool invert/*=false*/)
```
lassoan, line 699: Could this method have an optional `vtkMRMLMessageCollection*` argument to return details that can be displayed to the user? `vtkMRMLVolumeRenderingDisplayNode` could call this method to get the messages.

The advantage compared to the current implementation would be that it would be easier to keep GetClippingPlanesFromFunction and IsClippingShortcutValid consistent.

lassoan (commented):
Thanks Kyle, this looks great and works well! I've just added a few comments.

Modules/Loadable/VolumeRendering/MRMLDM/vtkMRMLVolumeRenderingDisplayableManager.cxx
```diff
@@ -1260,9 +1289,31 @@ void vtkMRMLVolumeRenderingDisplayableManager::vtkInternal::UpdatePipelineROIs(
...
+  if (!displayNode || !pipeline)
+  {
+    vtkErrorWithObjectMacro(this->External, "UpdateClippingPlanesFromMarkupsROINode: Display node or pipeline is invalid");
+    return;
+  }
+
+  vtkVolumeMapper* volumeMapper = this->GetVolumeMapper(displayNode);
+  if (!volumeMapper)
+  {
+    vtkErrorWithObjectMacro(this->External, "UpdateClippingPlanesFromMarkupsROINode: Unable to get volume mapper");
```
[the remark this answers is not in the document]
lassoan: @Sunderlandkyl this seems to be a valid finding

Libs/MRML/Core/vtkMRMLClipNode.cxx
Sunderlandkyl: What messages would this function return? I could add in messages such as "no planes found", and "fast method cannot clip with more than 6 planes" messages, however the function GetClippingPlanesFromFunction is generic and not specifically related to fast clipping.

Modules/Loadable/VolumeRendering/MRML/vtkMRMLVolumeRenderingDisplayNode.h
```diff
@@ -122,6 +131,13 @@ class VTK_SLICER_VOLUMERENDERING_MODULE_MRML_EXPORT vtkMRMLVolumeRenderingDispla
 
   void ProcessMRMLEvents(vtkObject *caller, unsigned long event, void *callData) override;
 
+  //@{
```
lassoan, line 134:
```suggestion
```

PR #8439, opened 2025-05-22
ENH: Add component selector to volume property node widget

Sunderlandkyl:
This commit adds a spinbox to select the current component in qMRMLVolumePropertyNodeWidget, allowing individually transfer functions to be modified for each scalar component.

### DIFF
--- a/Modules/Loadable/VolumeRendering/qSlicerVolumeRenderingModuleWidget.cxx
+++ b/Modules/Loadable/VolumeRendering/qSlicerVolumeRenderingModuleWidget.cxx
@@ -39,6 +39,7 @@
 #include "vtkMRMLMarkupsROINode.h"
 
 // VTK includes
+#include <vtkImageData.h>
 #include <vtkVolumeProperty.h>
 
 // STD includes
@@ -334,6 +335,17 @@ void qSlicerVolumeRenderingModuleWidget::onCurrentMRMLVolumeNodeChanged(vtkMRMLN
     d->PresetComboBox->setCurrentNode(presetNode);
     d->PresetComboBox->blockSignals(wasBlocking);
   }
+
+  int numberOfComponents = 1;
+  if (volumeNode)
+  {
+    vtkImageData* imageData = volumeNode->GetImageData();
+    if (imageData)
+    {
+      numberOfComponents = imageData->GetNumberOfScalarComponents();
+    }
+  }
+  d->VolumePropertyNodeWidget->setComponentCount(numberOfComponents);
 
   // Update widget from display node of the volume node
   this->updateWidgetFromMRML();
--- a/Modules/Loadable/VolumeRendering/Resources/UI/qMRMLVolumePropertyNodeWidget.ui
+++ b/Modules/Loadable/VolumeRendering/Resources/UI/qMRMLVolumePropertyNodeWidget.ui
@@ -6,19 +6,48 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>240</width>
-    <height>338</height>
+    <width>363</width>
+    <height>362</height>
    </rect>
   </property>
   <property name="windowTitle">
    <string>Volume Property Node</string>
   </property>
-  <layout class="QVBoxLayout" name="verticalLayout">
-   <property name="margin">
+  <layout class="QGridLayout" name="gridLayout">
+   <property name="leftMargin">
     <number>0</number>
    </property>
-   <item>
+   <property name="topMargin">
+    <number>0</number>
+   </property>
+   <property name="rightMargin">
+    <number>0</number>
+   </property>
+   <property name="bottomMargin">
+    <number>0</number>
+   </property>
+   <item row="0" column="0">
+    <widget class="QLabel" name="ComponentLabel">
+     <property name="sizePolicy">
+      <sizepolicy hsizetype="Fixed" vsizetype="Preferred">
+       <horstretch>0</horstretch>
+       <verstretch>0</verstretch>
+      </sizepolicy>
+     </property>
+     <property name="text">
+      <string>Component:</string>
+     </property>
+    </widget>
+   </item>
+   <item row="1" column="0" colspan="2">
     <widget class="ctkVTKVolumePropertyWidget" name="VolumePropertyWidget"/>
+   </item>
+   <item row="0" column="1">
+    <widget class="QSpinBox" name="ComponentSpinBox">
+     <property name="maximum">
+      <number>2</number>
+     </property>
+    </widget>
    </item>
   </layout>
  </widget>
--- a/Modules/Loadable/VolumeRendering/Widgets/qMRMLVolumePropertyNodeWidget.cxx
+++ b/Modules/Loadable/VolumeRendering/Widgets/qMRMLVolumePropertyNodeWidget.cxx
@@ -47,6 +47,7 @@ public:
   virtual void setupUi();
 
   vtkWeakPointer<vtkMRMLVolumePropertyNode> VolumePropertyNode;
+  int ComponentCount{ 1 };
 };
 
 // --------------------------------------------------------------------------
@@ -69,6 +70,8 @@ void qMRMLVolumePropertyNodeWidgetPrivate::setupUi()
                    q, SIGNAL(chartsExtentChanged()));
   QObject::connect(this->VolumePropertyWidget, SIGNAL(thresholdEnabledChanged(bool)),
                    q, SIGNAL(thresholdChanged(bool)));
+  QObject::connect(this->ComponentSpinBox, SIGNAL(valueChanged(int)),
+                   this->VolumePropertyWidget, SLOT(setCurrentComponent(int)));
 }
 
 // --------------------------------------------------------------------------
@@ -177,4 +180,22 @@ void qMRMLVolumePropertyNodeWidget::spreadAllPoints(double factor, bool dontSpre
 {
   Q_D(const qMRMLVolumePropertyNodeWidget);
   return d->VolumePropertyWidget->spreadAllPoints(factor, dontSpreadFirstAndLast);
+}
+
+// --------------------------------------------------------------------------
+int qMRMLVolumePropertyNodeWidget::componentCount()const
+{
+  Q_D(const qMRMLVolumePropertyNodeWidget);
+  return d->ComponentCount;
+}
+
+// --------------------------------------------------------------------------
+void qMRMLVolumePropertyNodeWidget::setComponentCount(int componentCount)
+{
+  Q_D(qMRMLVolumePropertyNodeWidget);
+  d->ComponentCount = componentCount;
+  int currentComponent = d->VolumePropertyWidget->currentComponent();
+  currentComponent = std::clamp(currentComponent, 0, componentCount);
+  d->VolumePropertyWidget->setCurrentComponent(currentComponent);
+  d->ComponentSpinBox->setRange(0, componentCount - 1);
 }
--- a/Modules/Loadable/VolumeRendering/Widgets/qMRMLVolumePropertyNodeWidget.h
+++ b/Modules/Loadable/VolumeRendering/Widgets/qMRMLVolumePropertyNodeWidget.h
@@ -45,6 +45,7 @@ class Q_SLICER_MODULE_VOLUMERENDERING_WIDGETS_EXPORT qMRMLVolumePropertyNodeWidg
   Q_OBJECT
   QVTK_OBJECT
   Q_PROPERTY(bool threshold READ hasThreshold WRITE setThreshold)
+  Q_PROPERTY(int componentCount READ componentCount WRITE setComponentCount)
 
 public:
   /// Constructors
@@ -79,10 +80,14 @@ public slots:
   void setChartsExtent(double extent[2]);
   void setChartsExtent(double min, double max);
 
+  int componentCount()const;
+  void setComponentCount(int component);
+
 signals:
   void thresholdChanged(bool enabled);
   void volumePropertyChanged();
   void chartsExtentChanged();
+  void componentChanged(int component);
 
 protected slots:
   void updateFromVolumePropertyNode();


Modules/Loadable/VolumeRendering/qSlicerVolumeRenderingModuleWidget.cxx
```diff
@@ -335,6 +336,17 @@ void qSlicerVolumeRenderingModuleWidget::onCurrentMRMLVolumeNodeChanged(vtkMRMLN
...
+  int numberOfComponents = 1;
+  if (volumeNode)
+  {
+    vtkImageData* imageData = volumeNode->GetImageData();
+    if (imageData)
+    {
+      numberOfComponents = imageData->GetNumberOfScalarComponents();
+    }
+  }
+  d->VolumePropertyNodeWidget->setComponentCount(numberOfComponents);
```
lassoan, line 348: We should probably limit to VTK_MAX_VRCOMP.

lassoan (commented):
Thank you, this will be good. I added some comments and suggestions.
I'm wondering if we should include the saving of properties of all components to the .vp.json file in this pull request (so that what is displayed is also saved).

Sunderlandkyl:
@lassoan I've added the  new storage node for saving vtkMRMLVolumePropertyNode as json files.
There is an example json in this gist: https://gist.github.com/Sunderlandkyl/2a52e47b9f05cf44e2a63096da3ec1ee

Sunderlandkyl:
Another example of a json using different functions for the first two components: https://gist.github.com/Sunderlandkyl/a7eab71981e500ffdd7d98632412246c

Sunderlandkyl:
@lassoan I've cleaned up the ReadXML/WriteXML calls in vtkMRMLVolumePropertyNode. Let me know if there are any other changes.

Modules/Loadable/VolumeRendering/MRML/vtkMRMLVolumePropertyJsonStorageNode.cxx
```diff
@@ -0,0 +1,735 @@
...
+    vtkSmartPointer<vtkMRMLJsonElement> gradientOpacityElement =
+      vtkSmartPointer<vtkMRMLJsonElement>::Take(componentElement->GetObjectProperty("gradientOpacity"));
+    if (gradientOpacityElement)
+    {
+      vtkNew<vtkPiecewiseFunction> gradientOpacity;
+      success &= this->ReadTransferFunction(gradientOpacity, gradientOpacityElement);
+      volumeProperty->SetGradientOpacity(i, gradientOpacity);
+    }
+
+    // TODO: Read TransferFunction2D (vtkImageData) is not implemented.
```
lassoan: This is not really a todo, but a potential feature to be developed in the future.
```suggestion
    // 2D transfer function support could be implemented here in the future by calling ReadTransferFunction2D.
```

Modules/Loadable/VolumeRendering/MRML/vtkMRMLVolumePropertyJsonStorageNode.cxx
```diff
@@ -0,0 +1,735 @@
...
+    vtkSmartPointer<vtkMRMLJsonElement> pointElement = vtkSmartPointer<vtkMRMLJsonElement>::Take(pointsElement->GetArrayItem(i));
+    if (!pointElement)
+    {
+      vtkErrorToMessageCollectionMacro(this->GetUserMessages(), "vtkMRMLVolumePropertyJsonStorageNode::ReadTransferFunction",
+               "Reading transfer function failed: point element is null.");
+      return false;
+    }
+
+    double x = pointElement->GetDoubleProperty("x");
+    double midpoint = pointElement->GetDoubleProperty("midpoint");
```
lassoan: What happens if this property does not exist? We could have a GetDoubleProperty method that also takes a default value as argument and does not log a warning/error if the property does not exist.

Modules/Loadable/VolumeRendering/MRML/vtkMRMLVolumePropertyJsonStorageNode.cxx
```diff
@@ -0,0 +1,735 @@
...
+  if (!writer->WriteToFileBegin(fullName.c_str(), VOLUME_PROPERTY_SCHEMA.c_str()))
+  {
+    vtkErrorToMessageCollectionMacro(this->GetUserMessages(), "vtkMRMLVolumePropertyJsonStorageNode::WriteDataInternal",
+      "Writing markups node file failed: unable to open file '" << fullName << "' for writing.");
+    return 0;
+  }
+
+  writer->WriteArrayPropertyStart("volumeProperties");
+
+  // Write volume property
```
lassoan:
```suggestion
```

Modules/Loadable/VolumeRendering/MRML/vtkMRMLVolumePropertyJsonStorageNode.cxx
```diff
@@ -0,0 +1,735 @@
...
+  {
+    vtkErrorToMessageCollectionMacro(this->GetUserMessages(), "vtkMRMLVolumePropertyJsonStorageNode::WriteVolumePropertyNode",
+      "Writing volume property node failed: volume property node is null.");
+    return false;
+  }
+
+  writer->WriteObjectStart();
+  writer->WriteVectorProperty("effectiveRange", volumePropertyNode->GetEffectiveRange(), 2);
+  this->WriteVolumeProperty(writer, volumePropertyNode->GetVolumeProperty());
+  writer->WriteObjectEnd(); // volumeProperties
```
lassoan:
```suggestion
  writer->WriteObjectEnd();
```

Modules/Loadable/VolumeRendering/MRML/vtkMRMLVolumePropertyJsonStorageNode.cxx
```diff
@@ -0,0 +1,735 @@
...
+  }
+  if (!volumeProperty)
+  {
+    vtkErrorToMessageCollectionMacro(this->GetUserMessages(), "vtkMRMLVolumePropertyJsonStorageNode::WriteVolumeProperty",
+      "Writing volume property failed: volume property is null.");
+    return false;
+  }
+
+  bool success = true;
+  success &= this->WriteContourValues(writer, volumeProperty->GetIsoSurfaceValues(), "isoSurfaceValues");
```
lassoan: For consistency with other writing methods, such as `writer->WriteBoolProperty...`
```suggestion
  success &= this->WriteContourValues(writer, "isoSurfaceValues", volumeProperty->GetIsoSurfaceValues());
```

Modules/Loadable/VolumeRendering/MRML/vtkMRMLVolumePropertyJsonStorageNode.cxx
```diff
@@ -0,0 +1,735 @@
...
+    writer->WriteDoubleProperty("specular", volumeProperty->GetSpecular(i));
+    writer->WriteDoubleProperty("specularPower", volumeProperty->GetSpecularPower(i));
+    writer->WriteObjectEnd(); // lighting
+
+    writer->WriteBoolProperty("disableGradientOpacity", volumeProperty->GetDisableGradientOpacity(i));
+    writer->WriteDoubleProperty("scalarOpacityUnitDistance", volumeProperty->GetScalarOpacityUnitDistance(i));
+
+    if (volumeProperty->GetColorChannels() > 1)
+    {
+      success &= this->WriteTransferFunction(writer, volumeProperty->GetRGBTransferFunction(i), "rgbTransferFunction");
```
lassoan:
```suggestion
      success &= this->WriteTransferFunction(writer, "rgbTransferFunction", volumeProperty->GetRGBTransferFunction(i));
```

Modules/Loadable/VolumeRendering/MRML/vtkMRMLVolumePropertyJsonStorageNode.cxx
```diff
@@ -0,0 +1,735 @@
...
+    else
+    {
+      success &= this->WriteTransferFunction(writer, volumeProperty->GetGrayTransferFunction(i), "grayTransferFunction");
+    }
+
+
+    success &= this->WriteTransferFunction(writer, volumeProperty->GetScalarOpacity(i), "scalarOpacity");
+    success &= this->WriteTransferFunction(writer, volumeProperty->GetStoredGradientOpacity(i), "gradientOpacity");
+
+    // TODO: Write TransferFunction2D (vtkImageData) is not implemented.
```
lassoan:
```suggestion
    // 2D transfer function support could be implemented here in the future by calling WriteTransferFunction2D.
```

Modules/Loadable/VolumeRendering/MRML/vtkMRMLVolumePropertyJsonStorageNode.cxx
```diff
@@ -0,0 +1,735 @@
...
+  volumeProperty->SetScatteringAnisotropy(volumePropertyElement->GetDoubleProperty("scatteringAnisotropy"));
+
+  vtkSmartPointer<vtkMRMLJsonElement> componentsElement = vtkSmartPointer<vtkMRMLJsonElement>::Take(volumePropertyElement->GetArrayProperty("components"));
+  for (int i = 0; i < componentsElement->GetArraySize() && i < VTK_MAX_VRCOMP; ++i)
+  {
+    vtkSmartPointer<vtkMRMLJsonElement> componentElement = vtkSmartPointer<vtkMRMLJsonElement>::Take(componentsElement->GetArrayItem(i));
+    if (!componentElement)
+    {
+      vtkErrorToMessageCollectionMacro(this->GetUserMessages(), "vtkMRMLVolumePropertyJsonStorageNode::ReadVolumeProperty",
+        "Reading volume property failed: component element is null.");
```
lassoan: Do we require all components to be present? (can't we have a file with only a single component?)

Modules/Loadable/VolumeRendering/Resources/Schema/volume-property-schema-v1.0.0.json
```diff
@@ -0,0 +1,171 @@
...
+                    "type": "array",
+                    "items": {
+                        "type": "object",
+                        "properties": {
+                            "x": { "type": "number" },
+                            "y": { "type": "number" },
+                            "midpoint": { "type": "number" },
+                            "sharpness": { "type": "number" }
+                        },
+                        "required": ["x", "y", "midpoint", "sharpness"]
```
lassoan: it would be better not to require midpoint and sharpness

lassoan (commented):
Thank you, it looks good. It would be nice to be able to read simple, minimal .vp.json files (e.g., only one component, only scalar opacity and color transfer functions) and it would be even nicer to not write everything to file (e.g., if use the volume properties on a single-component volume and never touch other components then it would be better to only write the first component to file).

lassoan (commented):
Thank you, the .vp.json file looks good now (not very long or complex). However, when I tested it, the GUI had some glitches:
- When working with a single-component volume, the `Independent components` checkbox should be hidden (now it is not hidden and when attempting to clicking on it, the checkbox remains checked)
- For an RGBA volume, when `Synchronize with Volumes module` is checked, the `Independent components` checkbox remains visible but clicking on it just makes the GUI flash but the checkbox remains unchecked
- After this, even if `Synchronize with Volumes module` is unchecked, the `Independent components` checkbox remains unchecked when clicking on it
- Component spinbox has no label.
- When checking/unchecking independent components, the GUI layout jumps around. We could either have the checkbox and spinbox in the same line or have them in two separate lines but then just disable the component selector, do not hide it.
- `Independent components` is too hard to grasp. I always need to think about a lot what it means. Something like a combobox or radiobutton: `Components:` with the options `independent` or `RGB` / `color + opacity` (depending on the number of components) would be better.

Sunderlandkyl:
@lassoan I've made some changes to the GUI, and to VP node initialization that should fix the issues.

lassoan:
Thank you! This works well now, I would just ask two last things:

1. When volume rendering a 3 or 4-component volume (especially when `VoxelVectorType: colorRGBA`) then we need to use RGBA components (not independent) by default.

Note that there are two ways to initialize volume rendering, both should use this default:
- when activating volume rendering by drag-and-dropping the volume from the data tree to the 3D view
- when activating volume rendering by clicking the eye icon in the Volume rendering module.

2. Improve the component selection layout

- Arrange the radiobuttons vertically
- Change "Index" label to "Component index"

image

- Try to reduce the flashing when switching between RGBA/Independent. It should be enough to show/hide the "Scalar color mapping" section, but it seems that all the widgets are removed and re-added.

Sunderlandkyl:
> When volume rendering a 3 or 4-component volume (especially when VoxelVectorType: colorRGBA) then we need to use RGBA components (not independent) by default.

This is already done here, but I guess there are a few other places where it is required.
https://github.com/Slicer/Slicer/pull/8439/files#diff-dbac4dec1d88b65de4b7b87bdd25dc8a24f8e73a10270e8f3b600702c7b9ec55R891-R900

lassoan:
I loaded [this RGBA volume](https://1drv.ms/u/c/872af7415c00bfb9/EayjdNbgoSZGno84x5LdPiAB9KIuSXIa1KOmyIQdHRUgeg?e=8IXQFk) by drag-and-drop and then activated volume rendering by drag-and-dropping the image from the subject hierarchy tree to the 3D view and then it did not show up as RGBA volume for me. In this case I think the SH plugin sets up the volume rendering display node.

Sunderlandkyl:
> I loaded [this RGBA volume](https://1drv.ms/u/c/872af7415c00bfb9/EayjdNbgoSZGno84x5LdPiAB9KIuSXIa1KOmyIQdHRUgeg?e=8IXQFk) by drag-and-drop and then activated volume rendering by drag-and-dropping the image from the subject hierarchy tree to the 3D view and then it did not show up as RGBA volume for me. In this case I think the SH plugin sets up the volume rendering display node.

Hmm, I tried the same thing and it showed up as an RGBA volume.

> Try to reduce the flashing when switching between RGBA/Independent. It should be enough to show/hide the "Scalar color mapping" section, but it seems that all the widgets are removed and re-added.

The scalar color mapping section is just hidden by ctkVTKVolumePropertyWidget. My guess is that after the color section is hidden the layout stretches the scalar opacity widget, which is then rendered for some reason before the widget shrinks.

> Improve the component selection layout
Arrange the radiobuttons vertically
Change "Index" label to "Component index"

This part is done.

lassoan:
Thank you, I'll check if I can still reproduce the wrong default setting. If not, then it should be good to go.

lassoan:
> Improve the component selection layout
Arrange the radiobuttons vertically
Change "Index" label to "Component index"

I don't see these updates in the latest commit in the .ui file. Could you please double-check that you committed your latest version?

Sunderlandkyl:
@lassoan sorry I must have messed up my rebase. Should be fixed now.

lassoan:
Thank you, the layout looks good now.

I could still reproduce the black volume by default when drag-and-dropping from Volume Rendering module's subject hierarchy three. It may be because the volume rendering SH plugin may have been disabled in this module to reduce redundancy. Probably we could re-enable it if that's indeed the problem.

While testing, I've also noticed that in RGBA component mode, the scalar opacity transfer function had no effect, if I previously selected an independent component > 0 in Independent mode. Suggested fix: if component mode is RGBA then opacity mapping must always edit the 0-th component.

lassoan:
Also, I have these test failueres in my branch that contains this commit:

```
1>	429 - vtkMRMLVolumePropertyNodeTest1 (Failed)           qSlicerVolumeRenderingModule
1>	431 - vtkMRMLVolumePropertyJsonStorageNodeTest1 (Failed) qSlicerVolumeRenderingModule
```

I have some other changes in my branch, so it is possible that they have some side effects, but most likely they are related to this commit. Could you have a look at these as well? Thank you!

Sunderlandkyl:
Ah I realized that you need to switch to the volume rendering module before clicking and dragging it to the view to reproduce the issue. If you just drag and drop from the data module without switching to the module it will work correctly.

lassoan:
I've rebased this on the latest Slicer main version (took care of all the formatting changes) and tested this. RGBA voIume appears correctly now by default, the GUI looks good! However, I've found the following issues:
- vtkMRMLVolumePropertyNodeTest1 still fails
- when I load an RGBA volume, show volume rendering, then in Volume rendering module I switch to "Independent" then Slicer crashes

Could you look into these? Thank you!

Sunderlandkyl:
@lassoan Sorry, I didn't see the comment and overwrote your changes on the branch. Could you re-push 4b185f639f74bca7faab93af2f4f1cb31d4cbac2, and I will add my changes to that?

I've fixed the test, but I'm still working on fixing the crash.

lassoan:
Done. To fix up an old branch with the new formatting, you can force rebase on upstream, reject all conflicting changes from up upstream, and then run clang-format. I use this code snippet to run clang-format on modified files: https://gist.github.com/lassoan/0d42031ec1ea6a90d56ebedc25cce230

Sunderlandkyl:
Ok, I've committed the test fix, and added a workaround for the crash.
I'll work on a fix for the crash at the VTK level in the coming days.

Sunderlandkyl:
@lassoan the lint error is fixed. Let me know if there are any other changes

lassoan (approved):
Thank you, this should be good now!

PR #8443, opened 2025-05-27
ENH: Fix build-tree environment variables in additional settings

RafaelPalomar:
This fixes the generation of `AdditionalLauncherSettings.ini` file for the build tree. The problem, as documented in [https://discourse.slicer.org/t/configuring-the-running-environment-for-an-extension-build-tree/43103](https://discourse.slicer.org/t/configuring-the-running-environment-for-an-extension-build-tree/43103), is that environment varibles were not included, as opposed to e.g., `PYTHONPATH` .

There seems to be an error when referring to the correct variable in the CMake infrastructure. This PR fixes the variable naming for both the `SlicerBlockAdditionalLauncherSettings.ini`, as well as the `SlicerExtensionGenerateConfig.cmake`. These changes has been tested successfully against the SlicerSOFA extension, which requires the `SOFA_ROOT` environment variable to be set by the launcher.

### DIFF
--- a/CMake/SlicerExtensionGenerateConfig.cmake
+++ b/CMake/SlicerExtensionGenerateConfig.cmake
@@ -4,7 +4,7 @@
 # This file is based on SlicerGenerateSlicerConfig.cmake.
 #
 # Example:
-#
+
 # Add these lines near the end of top-level CMakeLists.txt in Sequences extension
 # to allow other extensions to use it.
 #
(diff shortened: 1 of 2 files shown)


RafaelPalomar:
@jcfr, this has been hanging for some time and think it's ready to merge. I've rebased on `Slicer/main` to update it.

jamesobutler:
@RafaelPalomar Since it has been several months ago since last comment, is this set of changes still working well for you in the context of SlicerSOFA?

RafaelPalomar:
@jamesobutler Yes, the fix is still relevant and working well.

I've rebased the branch on current `main` and verified the fix with a dedicated test:

**Setup:** A minimal SuperBuild extension with an external project (`FakeLib`) that registers environment variables via the standard `mark_as_superbuild(VARS ... LABELS "ENVVARS_LAUNCHER_BUILD")` mechanism.

**Before fix** (on `main`):
- `[EnvironmentVariables]` section in `AdditionalLauncherSettings.ini`: **empty**
- `EnvVarTestExtension_ENVVARS_LAUNCHER_BUILD` in the generated config: `""`

**After fix:**
- `[EnvironmentVariables]` section: correctly populated with the registered variables
- `EnvVarTestExtension_ENVVARS_LAUNCHER_BUILD` in the generated config: correctly exported for dependent extensions

**Root cause:** In `SlicerBlockAdditionalLauncherSettings.cmake`, environment variables are collected into `EXTENSION_LAUNCHER_SETTINGS_ENVVARS` (lines 139–152), but the formatting block (lines 206–209) resets that variable and iterates over `EXTENSION_ENVVARS_BUILD` — which was never defined. The fix renames the collection variable to `EXTENSION_ENVVARS_BUILD`, matching the pattern used by `EXTENSION_LIBRARY_PATHS_BUILD`, `EXTENSION_PATHS_BUILD`, and `EXTENSION_PYTHONPATH_BUILD`.

This fix is actively needed in the context of SlicerSOFA (and any other extension that needs to set environment variables for its external dependencies in the build tree).

jamesobutler (approved):

PR #8448, opened 2025-06-02
BUG: Fix crash in CollapseBinaryLabelmaps

Punzo:

### DIFF
--- a/Libs/vtkSegmentationCore/vtkSegmentation.cxx
+++ b/Libs/vtkSegmentationCore/vtkSegmentation.cxx
@@ -2560,10 +2560,21 @@ void vtkSegmentation::CollapseBinaryLabelmaps(bool forceToSingleLayer /*=false*/
         thresholdedLabelmap->CopyDirections(currentLabelmap);
 
         int effectiveExtent[6] = { 0 };
-        vtkOrientedImageDataResample::CalculateEffectiveExtent(thresholdedLabelmap, effectiveExtent);
+        if (!vtkOrientedImageDataResample::CalculateEffectiveExtent(thresholdedLabelmap, effectiveExtent))
+        {
+          vtkErrorMacro("CollapseBinaryLabelmaps: Failed to calculate effective extent for thresholded labelmap.");
+          continue;
+        }
 
         vtkNew<vtkOrientedImageData> referenceImage;
         referenceImage->ShallowCopy(thresholdedLabelmap);
+
+        // Validate extent before setting it
+        if (effectiveExtent[0] > effectiveExtent[1] || effectiveExtent[2] > effectiveExtent[3] || effectiveExtent[4] > effectiveExtent[5])
+        {
+          vtkErrorMacro("CollapseBinaryLabelmaps: Invalid effective extent calculated.");
+          continue;
+        }
         referenceImage->SetExtent(effectiveExtent);
 
         vtkOrientedImageDataResample::ResampleOrientedImageToReferenceOrientedImage(thresholdedLabelmap, referenceImage, thresholdedLabelmap);


Punzo:
@lassoan

Libs/vtkSegmentationCore/vtkSegmentation.cxx
```diff
@@ -2558,10 +2558,21 @@ void vtkSegmentation::CollapseBinaryLabelmaps(bool forceToSingleLayer/*=false*/)
         thresholdedLabelmap->CopyDirections(currentLabelmap);
 
         int effectiveExtent[6] = { 0 };
```
lassoan, line 2562:
```suggestion
        int effectiveExtent[6] = { 0, -1, 0, -1, 0, -1 };
```

Libs/vtkSegmentationCore/vtkSegmentation.cxx
```diff
@@ -2558,10 +2558,21 @@ void vtkSegmentation::CollapseBinaryLabelmaps(bool forceToSingleLayer/*=false*/)
...
-        vtkOrientedImageDataResample::CalculateEffectiveExtent(thresholdedLabelmap, effectiveExtent);
+        if (!vtkOrientedImageDataResample::CalculateEffectiveExtent(thresholdedLabelmap, effectiveExtent))
+        {
+          vtkErrorMacro("CollapseBinaryLabelmaps: Failed to calculate effective extent for thresholded labelmap.");
```
lassoan, line 2565: This is not an error, it is valid to have an empty segment.

Libs/vtkSegmentationCore/vtkSegmentation.cxx
```diff
@@ -2558,10 +2558,21 @@ void vtkSegmentation::CollapseBinaryLabelmaps(bool forceToSingleLayer/*=false*/)
...
+          continue;
```
lassoan, line 2566: Even if the segment is empty, there are things to do. For example, the segment's layer can be collapsed into the first layer and thus it may need a new layer index and label value. So, I don't think we should just `continue` with the next segment.

I assume what happened is that `vtkOrientedImageDataResample::CalculateEffectiveExtent` returned false (because the segment was empty) and due to the incorrect initialization of the extent (`int effectiveExtent[6] = { 0 };` set all elements to 0, which is not an empty extent), which caused the crash somewhere later.

We should do these corrective and preventive actions:
- review if this method correctly flattens layers for empty segments (`thresholdedLabelmap` extent is `(0,-1,0,-1,0,-1)`).
- `vtkOrientedImageDataResample::ResampleOrientedImageToReferenceOrientedImage` works correctly (and does not crash) if the referenceImage extent is empty (`{0,-1,0,-1,0,-1}`) or all all-zeros (`{0,0,0,0,0,0}`).
- Check all `[6] = { 0 }` initializations in `vtkSegmentationCore`, change them to `{0, -1, 0, -1, 0, -1}` and review the code to make sure the code still works as intended.

Libs/vtkSegmentationCore/vtkSegmentation.cxx
```diff
@@ -2558,10 +2558,21 @@ void vtkSegmentation::CollapseBinaryLabelmaps(bool forceToSingleLayer/*=false*/)
...
+        }
 
         vtkNew<vtkOrientedImageData> referenceImage;
         referenceImage->ShallowCopy(thresholdedLabelmap);
+
+        // Validate extent before setting it
+        if (effectiveExtent[0] > effectiveExtent[1] || effectiveExtent[2] > effectiveExtent[3] || effectiveExtent[4] > effectiveExtent[5])
```
lassoan, line 2573: This is almost the same as above (when vtkOrientedImageDataResample::CalculateEffectiveExtent returns false), so handle it up there (with or right after the vtkOrientedImageDataResample::CalculateEffectiveExtent return value check).

Libs/vtkSegmentationCore/vtkSegmentation.cxx
```diff
@@ -2558,10 +2558,21 @@ void vtkSegmentation::CollapseBinaryLabelmaps(bool forceToSingleLayer/*=false*/)
...
+        {
+          vtkErrorMacro("CollapseBinaryLabelmaps: Invalid effective extent calculated.");
+          continue;
+        }
         referenceImage->SetExtent(effectiveExtent);
 
         vtkOrientedImageDataResample::ResampleOrientedImageToReferenceOrientedImage(thresholdedLabelmap, referenceImage, thresholdedLabelmap);
```
lassoan, line 2580: We should test `vtkOrientedImageDataResample::ResampleOrientedImageToReferenceOrientedImage` method with unusual (`[0,0,0,0,0,0]`)and empty (`[0,-1,0,-1,0,-1]`) `referenceImage` extent to make sure it does not crash. There are some tests that use this method in `SegmentationWidgetsTest1.py`, maybe two simple test cases could be added there.

lassoan (commented):
Thank you @Punzo. I'm not sure why this code crashed but there is definitely problem with the initialization of extents (set to all-zeros by default, which is not empty) and the proposed fix may avoid the crash but may not result in fully correct behavior.

Since this code is quite complex and mostly recently worked on by @Sunderlandkyl, it could be good if @Sunderlandkyl could review the code and implement the corrective and preventive actions that I described in the review comments.

lassoan:
@Sunderlandkyl what do you think, would you have time to thoroughly review the binary labelmap flattening code correctness when some or all segments are empty (and build in safeguards to make sure the code does not crash regardless of the contents of the segmentation)?

Punzo:
> @Sunderlandkyl what do you think, would you have time to thoroughly review the binary labelmap flattening code correctness when some or all segments are empty (and build in safeguards to make sure the code does not crash regardless of the contents of the segmentation)?

Hi @Sunderlandkyl, can you please let me know if you have time to look into this? If not I will try to have a look, please let me know. Thanks!

Libs/vtkSegmentationCore/vtkSegmentation.cxx
Sunderlandkyl: I'm looking into this now, but despite the initialization, it looks like the extent is fully calculated from `vtkOrientedImageDataResample::CalculateEffectiveExtent` and should still be correct. It only returns false if the calculated/returned extent is invalid. All zeroes would return true.

Punzo: ok thanks for looking into it

lassoan:
Thank you, the fix was extended and integrated in https://github.com/Slicer/Slicer/pull/8700

PR #8449, opened 2025-06-02
BUG: Do not add WindowLevelPreset if already in the list

Punzo:


Punzo:
@lassoan

Libs/MRML/Core/vtkMRMLScalarVolumeDisplayNode.cxx
```diff
@@ -664,6 +664,14 @@ void vtkMRMLScalarVolumeDisplayNode::AddWindowLevelPreset(double window, double
   preset.Window = window;
   preset.Level = level;
 
+  for (int p = 0; p < this->GetNumberOfWindowLevelPresets(); p++)
+  {
+    if (fabs(this->WindowLevelPresets[p].Window - window) < 1.e-6 && fabs(this->WindowLevelPresets[p].Level - level) < 1.e-6)
```
lassoan: It would be nicer to avoid specifying an arbitrary tolerance value and silently ignoring a request (user may expect that the preset is moved to the end or may want to get a "false" return value if a new preset is not actually added, etc.).

A solution that would avoid these complications would be to change `vtkMRMLScalarVolumeDisplayNode::CopyContent` to compare `WindowLevelPresets` member variable content and if there is a difference then overwrite it and call `Modified()`.

Punzo: ok I will implement the check at CopyContent and ReadXML level.

Punzo:
@lassoan review applied in https://github.com/Slicer/Slicer/pull/8449/commits/3978622f966556dad53419ce648d4234b818af6f

Punzo:
@lassoan I now clean all the previous levels when copying the node as requested. Please have a look.

lassoan:
I've accidentally closed this PR, with my force-push, therefore created https://github.com/Slicer/Slicer/pull/8542 to get it merged.

PR #8450, opened 2025-06-02
Fix volume seg async save

Punzo:

### DIFF
--- a/Libs/MRML/Core/vtkMRMLSegmentationStorageNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLSegmentationStorageNode.cxx
@@ -1362,8 +1362,31 @@ int vtkMRMLSegmentationStorageNode::WriteBinaryLabelmapRepresentation(vtkMRMLSeg
   }
   vtkOrientedImageDataResample::FillImage(commonGeometryImage, 0);
 
+  // Write to a temporary directory first, then move to the final destination
+  std::string targetDir = vtksys::SystemTools::GetFilenamePath(fullName);
+  std::string tempSubDir = std::string("TempWrite") + vtksys::SystemTools::GetFilenameWithoutExtension(fullName)
+                           + "_" + vtksys::SystemTools::GetCurrentDateTime("%Y%m%d%H%M%6f");
+  tempSubDir.erase(tempSubDir.find_last_not_of(" ") + 1);
+  std::vector<std::string> pathComponents;
+  vtksys::SystemTools::SplitPath(targetDir.c_str(), pathComponents);
+  pathComponents.push_back(tempSubDir);
+  std::string tempDir = vtksys::SystemTools::JoinPath(pathComponents);
+
+  if (vtksys::SystemTools::FileExists(tempDir.c_str()))
+  {
+    vtksys::SystemTools::RemoveADirectory(tempDir.c_str());
+  }
+  if (!vtksys::SystemTools::MakeDirectory(tempDir.c_str()))
+  {
+    vtkErrorToMessageCollectionMacro(this->GetUserMessages(), "vtkMRMLSegmentationStorageNode::WriteBinaryLabelmapRepresentation",
+                                     "Failed to create temporary directory " << tempDir);
+    return 0;
+  }
+
+  std::string tempFile = tempDir + "/" + vtksys::SystemTools::GetFilenameName(fullName);
+
   vtkNew<vtkTeemNRRDWriter> writer;
-  writer->SetFileName(fullName.c_str());
+  writer->SetFileName(tempFile.c_str());
   writer->SetUseCompression(this->GetUseCompression());
   writer->SetSpace(nrrdSpaceLeftPosteriorSuperior);
   writer->SetMeasurementFrameMatrix(nullptr);
@@ -1518,6 +1541,35 @@ int vtkMRMLSegmentationStorageNode::WriteBinaryLabelmapRepresentation(vtkMRMLSeg
   }
 
   writer->Write();
+
+  // Check for write errors
+  if (writer->GetWriteError())
+  {
+    vtkErrorToMessageCollectionMacro(this->GetUserMessages(), "vtkMRMLSegmentationStorageNode::WriteBinaryLabelmapRepresentation",
+      "Error writing temporary NRRD file " << tempFile);
+    vtksys::SystemTools::RemoveFile(tempFile);
+    vtksys::SystemTools::RemoveADirectory(tempDir.c_str());
+    return 0;
+  }
+
+  // Move temporary file to final destination
+  if (vtksys::SystemTools::FileExists(fullName.c_str(), true))
+  {
+    vtksys::SystemTools::RemoveFile(fullName.c_str());
+  }
+  int renameReturn = std::rename(tempFile.c_str(), fullName.c_str());
+  if (renameReturn != 0)
+  {
+    vtkErrorToMessageCollectionMacro(this->GetUserMessages(), "vtkMRMLSegmentationStorageNode::WriteBinaryLabelmapRepresentation",
+      "Failed to move temporary file " << tempFile << " to " << fullName << ", rename returned code " << renameReturn);
+    vtksys::SystemTools::RemoveFile(tempFile.c_str());
+    vtksys::SystemTools::RemoveADirectory(tempDir.c_str());
+    return 0;
+  }
+
+  // Clean up temporary directory
+  vtksys::SystemTools::RemoveADirectory(tempDir.c_str());
+
   this->GetUserMessages()->SetObservedObject(nullptr);
   int writeSuccess = true;
   if (writer->GetWriteError())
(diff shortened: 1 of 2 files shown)


Punzo:
@lassoan

Punzo:
@lassoan

I'm wondering if we should add a common method to the Slicer app logic (or in storage node) for safer file writing, particularly to handle concurrent write scenarios more robustly.

Instead of using the approach proposed in this PR—creating a temporary folder and then overwriting files (with code copied and pasted for each storage node write method)—it might be more reliable to implement a **file-based locking mechanism**.

There are two main use cases for this:

1. **Async node bulk data saving**  
   For example, in [this commit](https://github.com/Slicer/Slicer/pull/8141/commits/890c6376d84aaf07372bf3dd58e8a13f17be6297), async saving can conflict with Slicer's standard synchronous save mechanism.

2. **Multiple concurrent Slicer sessions**  
   When multiple sessions are writing (either async or sync), they may attempt to write to the same file or directory concurrently, leading to race conditions or data corruption.

A centralized, locking-based utility for file output could make these operations more robust and prevent such conflicts.

Would this be a direction worth exploring?

Punzo:
@jcfr @pieper if you have any feedback it would be very welcome!

pieper:
I would love to see a consistent approach to concurrency in the slicer core libraries, and starting with IO makes a lot of sense.  I had done experiments before with lots of model files (dti tractography vtk polydata) and got it working with temporary scenes per-model node that were then merged.  It worked, but it didn't save a lot of time in general cases.  

I think we should profile common use cases to identify bottlenecks.  I know that for large volumes gzip deflating is a single-thread bottleneck, so if compressing .seg.nrrds is in the critical path then spawning it off makes sense.  On the other hand I've been thinking that the zarr of compressing subblocks in parallel is more scalable and a bit cleaner.

Punzo:
> I would love to see a consistent approach to concurrency in the slicer core libraries, and starting with IO makes a lot of sense. I had done experiments before with lots of model files (dti tractography vtk polydata) and got it working with temporary scenes per-model node that were then merged. It worked, but it didn't save a lot of time in general cases.
>
> I think we should profile common use cases to identify bottlenecks. I know that for large volumes gzip deflating is a single-thread bottleneck, so if compressing .seg.nrrds is in the critical path then spawning it off makes sense. On the other hand I've been thinking that the zarr of compressing subbloc

in my use case is more to have a responsive UI than saving time. This is why I have implemented node writing in async in a background thread in https://github.com/Slicer/Slicer/pull/8141. The issue with an async saving is that concurrent writing can occur.

Of course I agree, if we can also boost the perfomances for writing to file would be a plus.

Punzo:
@lassoan by combining a lock file-based approach with writing files to a unique temporary folder and then performing an atomic rename,  the issue of file corruption during concurrent writes is resolved. 

This PR is ready for review. 

NOTE: This PR currently addresses the issue for the **volume** and **segmentation** storage nodes.

Punzo:
Feedback from Andras:

- [x] move the implementation (it is copied and pasted both in volume and segmentation storage node) into the vtkMRMLStorageNode and have a helper method, so it will also be available to all bulk data storage nodes. 
- [x] The method will take a list of input and output files names.
- [x] For storage nodes with multiple filenames, create a lock file for each one and retry until you get the lock for all of them.
- [x] the lockSleepMs shouldn't be costant. add some unifrom random values.

Punzo:
> Feedback from Andras:
> 
> * [x]  move the implementation (it is copied and pasted both in volume and segmentation storage node) into the vtkMRMLStorageNode and have a helper method, so it will also be available to all bulk data storage nodes.
> * [x]  The method will take a list of input and output files names.
> * [x]  For storage nodes with multiple filenames, create a lock file for each one and retry until you get the lock for all of them.
> * [x]  the lockSleepMs shouldn't be costant. add some unifrom random values.

@lassoan done in https://github.com/Slicer/Slicer/pull/8450/commits/be48c6138bb4871dd87f857aee52c923ac257402. This would be ready for review.

Libs/MRML/Core/vtkMRMLSegmentationStorageNode.cxx
```diff
@@ -1624,10 +1652,28 @@ int vtkMRMLSegmentationStorageNode::WritePolyDataRepresentation(vtkMRMLSegmentat
     multiBlockDataset->SetBlock(segmentIndex, currentPolyDataCopy);
   }
 
+  std::string tempDir;
+  vtkNew<vtkStringArray> fullNames;
+  fullNames->InsertNextValue(path);
+  vtkNew<vtkStringArray> tempFiles;
+  if (!this->CreateUniqueTempFilesForWrite(fullNames, tempDir, tempFiles))
```
lassoan: Segmentation may be written to multiple files, even in subfolders.
If file extension is .nhdr then two files are created (.nhdr, .raw or .gz). If the source representation is "closed surface" then in addition to the (filename).vtm file, multiple (segmentname).vtp files are written into the `(filename).seg` subfolder.

Libs/MRML/Core/vtkMRMLStorageNode.cxx
```diff
@@ -1648,3 +1661,223 @@ void vtkMRMLStorageNode::FixFileName()
...
+  {
+    vtkErrorToMessageCollectionMacro(
+      this->GetUserMessages(), "vtkMRMLStorageNode::CreateUniqueTempFilesForWrite", "Failed to create folder: '" << tempDir << "'");
+    return false;
+  }
+
+  tempFiles->Reset();
+  for (vtkIdType fileIndex = 0; fileIndex < fullNames->GetNumberOfValues(); ++fileIndex)
+  {
+    std::string fileName = vtksys::SystemTools::GetFilenameName(fullNames->GetValue(fileIndex));
```
lassoan: Files may be in subfolders (e.g., .seg.vtm files), so it is not enough to get the filename component.

Libs/MRML/Core/vtkMRMLStorageNode.cxx
```diff
@@ -1648,3 +1661,223 @@ void vtkMRMLStorageNode::FixFileName()
     this->SetFileName(secondCollapsedFullPath.c_str());
   }
 }
+
+//----------------------------------------------------------------------------
+bool vtkMRMLStorageNode::CreateUniqueTempFilesForWrite(vtkStringArray* fullNames, std::string& tempDir, vtkStringArray* tempFiles)
```
lassoan: We don't know the list of all filenames that a writer will create in advance (ITK writer does not tell you what files it is going to create) and we cannot specify additional filenames to a writer. All we can do is to specify a folder+filename, so to allow parallel writes, we should not need to create unique filenames, just a unique folder name.

Libs/MRML/Core/vtkMRMLStorageNode.cxx
```diff
@@ -1648,3 +1661,223 @@ void vtkMRMLStorageNode::FixFileName()
...
+  vtksys::Directory dir;
+  dir.Load(sourceDir.c_str());
+
+  vtkNew<vtkStringArray> targetFiles;
+  vtkNew<vtkStringArray> sourceFiles;
+  std::vector<std::string> targetPathComponents;
+  vtksys::SystemTools::SplitPath(targetDir.c_str(), targetPathComponents);
+  std::vector<std::string> sourcePathComponents;
+  vtksys::SystemTools::SplitPath(sourceDir.c_str(), sourcePathComponents);
+  for (size_t fileNum = 0; fileNum < dir.GetNumberOfFiles(); ++fileNum)
```
lassoan: Atomic file locking+file creation is quite complex in itself. We should not try to create multiple lockfiles per node. Instead, we should only use a single lockfile, which is based on the primary filename (that the user specifies, that is the first filename in the filename list).

Libs/MRML/Core/vtkMRMLStorageNode.cxx
```diff
@@ -1648,3 +1661,223 @@ void vtkMRMLStorageNode::FixFileName()
...
+  return true;
+}
+
+//----------------------------------------------------------------------------
+bool vtkMRMLStorageNode::MoveFilesWithLocking(const std::string& sourceDir, const std::string& targetDir)
+{
+  if (sourceDir.empty())
+  {
+    vtkErrorToMessageCollectionMacro(this->GetUserMessages(), "vtkMRMLStorageNode::MoveFilesWithLocking", "Source directory string is empty. Cannot commit files.");
+    vtksys::SystemTools::RemoveADirectory(sourceDir.c_str());
```
lassoan: If move fails, we must not remove the source directory as a side effect, but the source folder must remain in place. The caller may decide to delete the source folder.

Libs/MRML/Core/vtkMRMLStorageNode.cxx
```diff
@@ -1648,3 +1661,223 @@ void vtkMRMLStorageNode::FixFileName()
...
+
+  // Try to acquire all locks
+  int lockAttempts = 0;
+  bool locksAcquired = false;
+  while (lockAttempts < VTKMRMLSTORAGENODE_MAX_LOCK_ATTEMPTS)
+  {
+    locksAcquired = true;
+    for (const auto& lockFile : lockFiles)
+    {
+      if (vtksys::SystemTools::FileExists(lockFile.c_str(), true))
```
lassoan: Several processes may find that the lock file does not exist and then go ahead with the move operation. To prevent this, we must use file API that is guaranteed to create a file with a unique owner (similarly how you must check&acquire mutex in an atomic operation).

lassoan (commented):
This will be good, but we can probably simplify the writing preparation (just create a unique temporary folder) and then we should use just a single lockfile to serialize the moving of written files to their final location.

I'm also wondering why this is only done for segmentation and volume files and not for others. Is it only a problem if multiple files are created (.obj files are written as multiple files, too)? Or if writing takes long time (transforms and sequences can take a long time to write, too)? If they should be implemented for other node types, too, but right now they are just not implemented due to lack of time then this should be documented (maybe enough to describe in the commit comment).

Add more details in the commit message: give a high-level overview of what is done and why.

Punzo:
> This will be good, but we can probably simplify the writing preparation (just create a unique temporary folder) and then we should use just a single lockfile to serialize the moving of written files to their final location.
> 
> I'm also wondering why this is only done for segmentation and volume files and not for others. Is it only a problem if multiple files are created (.obj files are written as multiple files, too)? Or if writing takes long time (transforms and sequences can take a long time to write, too)? If they should be implemented for other node types, too, but right now they are just not implemented due to lack of time then this should be documented (maybe enough to describe in the commit comment).
> 
> Add more details in the commit message: give a high-level overview of what is done and why.

Thank you for your review and suggestions. I agree that this approach should also be extended to models, sequences, and non-linear transforms. Due to development time constraints, the current implementation is limited to segmentations and volumes. My plan is to implement similar handling for the other node types in the future. Additionally, starting with a couple of classes allows us to test and validate the approach before generalizing it further.

I will address the in-line comments and update the commit message with a more detailed high-level overview tomorrow.

Libs/MRML/Core/vtkMRMLSegmentationStorageNode.cxx
Punzo: the method was actually just generating the temp folder and not the temporary files. I cleaned the method (renamed, changed input/output). See new implementation in rebased commit.

Libs/MRML/Core/vtkMRMLStorageNode.cxx
Punzo: I agree, I now use just one lock file for node and I use atomic linux/windows api to open/create the lock file. See new implementation in rebased commit.

Libs/MRML/Core/vtkMRMLStorageNode.cxx
Punzo: the method was actually just generating the temp folder and not the temporary files. I cleaned the method (renamed, changed input/output). See new implementation in rebased commit.

Libs/MRML/Core/vtkMRMLStorageNode.cxx
Punzo: ok removed the remove of the temp directory and moved in the calling methods. See new implementation in rebased commit.

Libs/MRML/Core/vtkMRMLStorageNode.cxx
Punzo: I agree, I now use just one lock file for node and I use atomic linux/windows api to open/create the lock file. See new implementation in rebased commit.

Libs/MRML/Core/vtkMRMLStorageNode.cxx
Punzo: the method was actually just generating the temp folder and not the temporary files. I cleaned the method (renamed, changed input/output). See new implementation in rebased commit.

Punzo:
@lassoan I have applied your review in the rebased commit https://github.com/Slicer/Slicer/pull/8450/commits/3c6ad5f41a2cf48bcc334c4d3c0d666bd0477a14

I have also updated to latest Slicer and added tests.

Punzo:
moved this in https://github.com/Slicer/Slicer/pull/8141, since it is related, and it is difficult to test without the other commits in 8141

PR #8453, opened 2025-06-02
DOC: Add top-level "Extensions" page

jcfr:
This top-level page serves as a landing page to introduce extensions and reference both the User and Developer guides.

---

Related pull requests:
* https://github.com/Slicer/slicer.org/pull/333
* https://github.com/KitwareMedical/slicer-extensions-webapp/pull/196
* https://github.com/Slicer/slicer.org/pull/330


jamesobutler (commented):
It seems like "Extensions" would be a upper level page that would have a child page of "Extensions Manager". Where the upper level describes what extensions are and the child page describes how to manage and install them with the manager. This being different than having them at the same level side-by-side.

fedorov:
I would also include the link to the extensions page, and add a screenshot with a sample of extensions. This page is too dry right now. But we can also refine over time once it is merged. I am happy to help.

jcfr:
> It seems like "Extensions" would be a upper level page that would have a child page of "Extensions Manager"

I initially consider renaming the existing page `extensions_manager.md` to `extensions.md` ... but in interest of (1)  keeping existing link working and (2) addressing initial concern of @fedorov , i kept the approach simple.

hjmjohnson (approved):
This seems like a good initial commit. Refinements will be needed later.  This will allow others to contribute.

lassoan:
I've merged the two pages to reduce noise and added a redirect.

There is only an "Extensions" menu item:

image

After clicking on "Extensions":

image

@jcfr please have a look.

jcfr:
🙏 @lassoan  After you approve, it will be integrated 🚀

lassoan (approved):
Looks good, thank you!

PR #8457, opened 2025-06-03
COMP: Modernize integration of VTK build with Python site-packages

jcfr:
Transition from the legacy `.egg-info` directory to using the modern `.dist-info` format. This ensures the VTK version can be extracted by pip after transitioning from Python 3.9.10 to Python 3.12.10.

See https://packaging.python.org/en/latest/specifications/recording-installed-packages/

To streamline the developer experience doing incremental build, the legacy `vtk.*.egg-info` directory is removed if it exists.


lassoan (approved):
Looks good to me.

PR #8459, opened 2025-06-04
ENH: Replace deprecated imp module with importlib to support Python 3.12+

jcfr:
This commit replaces usage of the deprecated imp module with modern importlib-based APIs. The `imp` module was deprecated in Python 3.4 and removed[^1] in Python 3.12.

[^1]: https://docs.python.org/3/library/imp.html#module-imp

Key updates:
- Use `importlib.util.spec_from_file_location` and `module_from_spec` for loading modules from file paths.
- Use `types.ModuleType` instead of `imp.new_module`.
- Ensure modules are registered and accessed via `sys.modules`.
- Avoid `PyImport_AddModule` when retrieving modules loaded with `importlib`.


jamesobutler (approved):
Makes sense. Only previous Slicer community mention of resolving this deprecation appears to have been in https://discourse.slicer.org/t/package-imp-deprecated/7797.

PR #8463, opened 2025-06-04
COMP: Remove obsolete MRML logic functions related to unreferenced nodes cleanup

jcfr:
The MRML logic functions `RemoveUnreferencedStorageNodes` and `RemoveUnreferencedDisplayNodes` were first introduced in e6ae7b96b16 ("BUG_FIXED: 744 Remove unreferenced storage and display nodes before saving a scene", 2010-04-17) and their last use removed in f8fb837370b ("ENH: SaveData dialog uses qSlicerFileWriters for improved scalability", 2012-06-07)

### DIFF
--- a/Libs/MRML/Core/vtkMRMLLogic.cxx
+++ b/Libs/MRML/Core/vtkMRMLLogic.cxx
@@ -1,10 +1,4 @@
 
-// MRML includes
-#include "vtkMRMLScene.h"
-#include "vtkMRMLStorageNode.h"
-#include "vtkMRMLDisplayableNode.h"
-#include "vtkMRMLDisplayNode.h"
-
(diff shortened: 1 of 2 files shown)


jamesobutler (approved):

lassoan:
Orphan storage and display nodes still appear scenes and cause small complications. It is not clear what leaves them behind, it may be due to incorrectly implemented modules or there may be bugs in Slicer core; but once they are in a scene they stay there forever. If we remove these utility functions now then we'll have to find and re-add them later when we find the time to implement automatic cleanup.

jcfr:
Closing. Superseded by:
* https://github.com/Slicer/Slicer/pull/8470

PR #8464, opened 2025-06-04
COMP: Move vtkSlicerCLIModuleLogic from Base/QTCLI to Base/Logic

jcfr:
Streamline reuse of the `vtkSlicerCLIModuleLogic` class removing its dependency on Qt.

---

Related pull request:
* https://github.com/Slicer/Slicer/pull/8004


jcfr:
We should check if removing the linking to `qSlicerBaseQTCLI` can be removed from `SlicerMacroBuildModuleLogic.cmake`:

https://github.com/Slicer/Slicer/blob/a5a2f3db610a90ec46b8f68f3b116ee4d1a3b99e/CMake/SlicerMacroBuildModuleLogic.cmake#L65-L72

lassoan (approved):
It looks good to me.

PR #8487, opened 2025-06-18
BUG: Avoid invalid default when retrieving active place node class name

jcfr:
Resolves potential runtime issues when `GetActivePlaceNodeClassName()` returns `nullptr`.


jcfr:
I did look for similar issues in the code base and didn't find any.

jamesobutler (approved):

PR #8489, opened 2025-06-18
ENH: Enforce python 3.12 and newer syntax

jamesobutler:
This follows up Slicer's switch from Python 3.9 to Python 3.12 in https://github.com/Slicer/Slicer/commit/5e952f7c5b031dd32858f245e18719e09099117c.

Changes were automatically applied by running:
`python -m pre_commit run --all-files`

PR #8494, opened 2025-06-20
BUG: Fix invalid typing usage in parameter node following pyupgrade

jamesobutler:
This is a follow-up to https://github.com/Slicer/Slicer/commit/52d00b16ee868031029b5cf5f942eb6c24677ab9 where Python syntax was upgraded to a new version automatically through pyupgrade.

Additionally it follows up by defining py312 linting in the ruff configuration.

Local tests have been run confirming that the tests mentioned in https://github.com/Slicer/Slicer/pull/8489#issuecomment-2988233751 are no longer failing.

### DIFF
--- a/.ruff.toml
+++ b/.ruff.toml
@@ -1,4 +1,4 @@
-target-version = "py39"
+target-version = "py312"
 line-length = 550
 
 extend-exclude = [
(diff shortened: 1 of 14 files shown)


jcfr (approved):
🙏 Thanks for addressing this and splitting the changes in multiple commits.

PR #8501, opened 2025-06-24
COMP: Override cmake minimum policy

hjmjohnson:
Many packages with very simple cmake needs
do not require features from newer versions
of cmake, but work just fine with almost all
versions of cmake.

Use -DCMAKE_POLICY_VERSION_MINIMUM:STRING=3.20.6
to override the minimum setting to a higher
policy version.

Set cmake_minimum_required 3.20.6...3.31.8

The range is set to represent the minimum cmake
version that is needed to support building of
python 3.12.

The maximum range is the 3.31.8 which represents
the latest version in the cmake 3 series.

NOTE:  A few versions of cmake within this range
are known to have regressions that are tested
and prevented in the main Slicer/CMakeLists.txt file.


jamesobutler (approved):

CMakeLists.txt
```diff
@@ -8,9 +8,6 @@ if("${CMAKE_VERSION}" VERSION_GREATER_EQUAL "3.25.0" AND "${CMAKE_VERSION}" VERS
   message(FATAL_ERROR "CMake version is ${CMAKE_VERSION} and using CMake >=3.25.0,<=3.25.2 is not supported.\nSee https://gitlab.kitware.com/cmake/cmake/-/issues/24567")
 endif()
 
-if("${CMAKE_VERSION}" VERSION_GREATER_EQUAL "4.0")
-  message(FATAL_ERROR "CMake version is ${CMAKE_VERSION} and using CMake >=4.0 is not supported.")
-endif()
```
jcfr:
```suggestion
if("${CMAKE_VERSION}" VERSION_GREATER_EQUAL "4.0")
  message(FATAL_ERROR "CMake version is ${CMAKE_VERSION} and using CMake >=4.0 is not supported.")
endif()
```

Docs/developer_guide/build_instructions/windows.md
```diff
@@ -8,9 +8,8 @@ Slicer relies on a number of large third-party libraries (such VTK, ITK, DCMTK),
 
 ## Install prerequisites
 
-- [CMake](https://www.cmake.org/cmake/resources/software.html) version >= 3.16.3 and < 4.
+- [CMake](https://www.cmake.org/cmake/resources/software.html) version >= 3.20.6.
```
jcfr:
```suggestion
- [CMake](https://www.cmake.org/cmake/resources/software.html) version >= 3.20.6 and < 4.
```

jcfr (changes requested):

jcfr:
After further consideration, we need to review (and address) all CMake policy warnings reported when using newer version of CMake.

In the meantime, I will revisit this patch to remove the upper bound associated with the `cmake_minimum_required` call.

Note that I also removed from this patch the following:
- setting of `CMAKE_POLICY_VERSION_MINIMUM` to external project. Corresponding project should be updated instead.
- removal of check reporting error if CMake 4.x is used

hjmjohnson:
@jcfr I'll leave this PR to do with what you want.

jcfr (approved):

PR #8504, opened 2025-06-24
COMP: OpenSSL 1.1.1w is needed to fix missing include

hjmjohnson:
Newer compilers require explicit inclusion of string.h for memcmp function usage.

### DIFF
--- a/SuperBuild/External_OpenSSL.cmake
+++ b/SuperBuild/External_OpenSSL.cmake
@@ -48,13 +48,13 @@ if(NOT DEFINED OPENSSL_LIBRARIES
     # Starting with Qt 5.12.4, official Qt binaries are build against OpenSSL 1.1.1
     # See https://www.qt.io/blog/2019/06/17/qt-5-12-4-released-support-openssl-1-1-1
     if("${Qt5_VERSION_MAJOR}.${Qt5_VERSION_MINOR}.${Qt5_VERSION_PATCH}" VERSION_GREATER_EQUAL "5.12.4")
-      set(_default_version "1.1.1g")
+      set(_default_version "1.1.1w")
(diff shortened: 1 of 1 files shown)


jamesobutler (approved):

jcfr:
> This change seems to break the MacOS build: [https://slicer.cdash.org/viewBuildError.php?buildid=3843118](https://slicer.cdash.org/viewBuildError.php?buildid=3843118) @jcfr, @pieper, @jamesobutler, @hjmjohnson   

 _Originally posted by @RafaelPalomar in [ebdd812](https://github.com/Slicer/Slicer/commit/ebdd8127bbffa9ae01c1d37eaadd24ad0cff0fef#commitcomment-160865974)_

> Thanks for pointing this out, I also observed this. I will address this along with generating the package to also use that version on Windows. 

 _Originally posted by @jcfr in [ebdd812](https://github.com/Slicer/Slicer/commit/ebdd8127bbffa9ae01c1d37eaadd24ad0cff0fef#commitcomment-160867440)_

jcfr:
We need to use a patched version of OpenSSL like we did for `1.1.1g`:

> ```
> # Workaround linking error when building against non-system zlib on macOS
> # See https://github.com/openssl/openssl/pull/12238
> set(OpenSSL_1.1.1g_URL https://github.com/Slicer/Slicer-OpenSSL/releases/download/sources/openssl-1.1.1g-pr12238.tar.gz)
> set(OpenSSL_1.1.1g_MD5 4765dcd60bcbed784c59ad7c2ca2b841)
> ```

See https://github.com/Slicer/Slicer/commit/27e98f95ca5ad003d0cd89534e6d2fd8669cdeb8

PR #8516, opened 2025-06-27
ENH: Support fixup overrides in app and extension packaging contexts

jcfr:
This commit introduces support for two new variables that extensions can define in their `CMakeLists.txt` to customize the macOS bundle fixup process:

- `<ExtensionName>_FIXUP_BUNDLE_CANDIDATES_PATTERNS`: Specifies additional glob patterns to locate libraries that should be included in the fixup process.

- `<ExtensionName>_FIXUP_BUNDLE_EMBEDDED_PATH_OVERRIDES`: Specifies rules in the form `<regex>|<relative_path>` to override how `install_name_tool` sets the install name and resolves rpath references.

These variables are processed in both:

- the **application packaging context** (`SlicerCPack.cmake`), for extensions bundled into a custom Slicer application, and
- the **extension packaging context** (`SlicerExtensionCPack.cmake`), for standalone extension packaging workflows.

To be properly consumed by the packaging logic, these variables must be set as `CACHE` entries in the extension’s CMake configuration.

This improves flexibility when packaging extensions that depend on additional third-party libraries or use non-standard directory layouts.


RafaelPalomar (approved):
LGTM

@jcfr, thanks for working on this. This infrastructure is very helpful for superbuild extensions

jamesobutler (approved):

PR #8519, opened 2025-06-30
ENH: Make JSON files in extensions translatable

lassoan:
MONAIAuto3DSeg extension has a model description json file, which contains model title, description, etc., which is displayed on the GUI and therefore have to be translatable to different languages.

A general-purpose mechanism is implemented to detect if a json file is translatable (a filename-schema.json file is present in the same folder) and determine which properties are translatable (those that have their "transatable" custom property set to true).

### DIFF
--- a/Modules/Loadable/Volumes/Logic/vtkSlicerVolumesLogic.cxx
+++ b/Modules/Loadable/Volumes/Logic/vtkSlicerVolumesLogic.cxx
@@ -67,7 +67,7 @@
 
 namespace
 {
-  const std::string MARKUPS_SCHEMA =
+  const std::string VOLUME_DISPLAY_PRESETS_SCHEMA =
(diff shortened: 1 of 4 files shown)


Modules/Loadable/Volumes/Logic/vtkSlicerVolumesLogic.cxx
```diff
@@ -1609,17 +1609,26 @@ void vtkSlicerVolumesLogic::InitializeDefaultVolumeDisplayPresets()
   std::string errorPrefix = "vtkSlicerVolumesLogic::InitializeDefaultVolumeDisplayPresets failed: Error reading '" + displayPresetsFilename + "'.";
 
   // Verify schema
-  if (!jsonRoot->HasMember("@schema"))
+  std::string schemaStr;
+  if (!jsonRoot->HasMember("$schema"))
```
jcfr: May be worth contributing the support for `$schema` in its own pull request. 

Some references related to the use of `@schema` vs `$schema` are also available at https://github.com/Slicer/Slicer/pull/6618#issuecomment-1292107500

PR #8529, opened 2025-07-03
COMP: Update BRAINSTools git tag in SuperBuild.cmake to fix version mis-match

hjmjohnson:
The reported version was not consistent with the git tag.  This is a
minor issue that only changes the recored/reported version.


jcfr:

----

> [!TIP]
> For future contributions, consider pushing the branch to your fork (instead of upstream Slicer) :pray:

jcfr:
After updating the comment associated with `GIT_SHA` in `SuperBuild.cmake` along with the commit message, I just forced push this branch.

As I mentioned in the commit message, my understanding is the following:

> At the time of this update, the upstream project introduced a dedicated Slicer branch `slicer-20250702-postv5.8.0`, corresponding to tag `v5.8.0` + 1 patch.

Since keeping an orphan branch is fragile could you make sure to also create a tag called `v5.8.0.post1`based of that branch ?

PR #8530, opened 2025-07-03
COMP: Remove conditional Qt5 support

hjmjohnson:
The Slicer_HAVE_QT5 flag was used to include code that did
not work with Qt4.  Qt4 is no longer supported.


hjmjohnson:
accidental duplicate of #8528

PR #8535, opened 2025-07-07
ENH: Enable relative import of Slicer Python-wrapped libraries

Thibault-Pelletier:
- Add slicer __init__ kit  relative import when using Slicer as a python library

### DIFF
--- a/Base/Python/slicer/__init__.py
+++ b/Base/Python/slicer/__init__.py
@@ -194,7 +194,11 @@ for kit in available_kits:
     try:
         exec("from %s import *" % (kit))
     except ImportError as detail:
-        print(detail)
+        # Try kit relative import if installed in as a traditional package
+        try:
+            exec("from .%s import *" % (kit))
+        except ImportError:
+            print(detail)
 
     del kit
 


jcfr:
I suggested improving the commit message to something like this ... does this correctly describe the use case you are try to support ?

````
ENH: Enable relative import of Slicer Python-wrapped libraries

Improve the fallback mechanism in `slicer/__init__.py` to support relative import
of Slicer Python-wrapped libraries (kits) when they are distributed as standalone
Python packages (e.g., `slicer-mrml`).

This change allows importing modules using relative paths when absolute imports
fail, supporting package layouts like:

```
slicer_mrml/
  ├── __init__.py
  ├── vtkAddon/
  ├── vtkSegmentationCore.so
  └── qMRMLWidgetsPythonQt/
```

This makes it possible to use Slicer Python kits in isolated package distributions
outside the monolithic Slicer application structure.
````

**Comment by Thibault-Pelletier** [comment]
Yes, your suggested commit message change is great!

File: Base/Python/slicer/__init__.py [hunk 1]
```diff
@@ -194,7 +194,11 @@ def _createModule(name, globals, docstring):
     try:
         exec("from %s import *" % (kit))
     except ImportError as detail:
-        print(detail)
+        # Try kit relative import if installed in as a traditional package
+        try:
+            exec("from .%s import *" % (kit))
```
jcfr, line 199: Should this be done when `standalone_python` is true ? Should we identify which kits support to be imported in a standalone interpreter ?

Thibault-Pelletier: At the moment we tested importing anything which is defined in the kit.py file

Since the file is already configured using the kits.py.in all kits should be importable in the packaged / configured context.

I'm not familiar with the `standalone_python` flag so I don't have any opinion on it.

Thibault-Pelletier: I had a look at the `standalone_python` flag and I'm not sure there is a need to only allow fallback in this case. If we decide at some point to change how the Slicer python pyd files are installed and available in the 3D Slicer application environment, importing the files using relative import will still be relevant.

jcfr: `standalone_python ` means running code directly from `PythonSlicer`[^1]

Can we further improve the commit message to describe how Slicer should be built to support this use case ? Or at least describe from high level how the modified Slicer is organized to justify this.

[^1]: https://slicer.readthedocs.io/en/latest/developer_guide/python_faq.html#python-faq

Thibault-Pelletier: The modified install would be the one you described in your updated message.

Traditional python package install is in a `Python/Lib/site-packages/<package_name>` . If the `slicer` package is installed in a Python `site-packages` folder then the relative import is necessary.

Python/Lib/site-packages/slicer/
  ├── __init__.py
  ├── <any_kit_or_package>
  ├── vtkAddon/
  ├── vtkSegmentationCore.so
  └── qMRMLWidgetsPythonQt/
  
Current Slicer adds in the path to the slicer folder content to the environment path which allows the kit installed in the folder to be importable using a from <kit_name> import * to work.

Should I update the root path of the commit message?

Thibault-Pelletier: Hi @jcfr, 
I just rebased the MR on main. Is there anything you would like me to reword / change to merge this MR?

jcfr: Let's fix the commit message, I incorrectly specified `slicer-mrml` but for this to be relevant. It should be updated like this:

````
Improve the fallback mechanism in `slicer/__init__.py` to support relative import
of Slicer Python-wrapped libraries (kits) when they are distributed as standalone
Python packages installed in `slicer`.

This change allows importing modules using relative paths when absolute imports
fail, supporting package layouts like:

```
slicer/
  ├── __init__.py
  ├── vtkAddon/
  ├── vtkSegmentationCore.so
  └── qMRMLWidgetsPythonQt/
```

This makes it possible to use Slicer Python kits in isolated package distributions
outside the monolithic Slicer application structure.
````

jcfr: I ended up integrating using `Squash & Merge` specifying the message suggested in https://github.com/Slicer/Slicer/pull/8535#discussion_r2206167349

Thibault-Pelletier: Thanks for the suggested message changes and the squash & merge !

PR #8525, opened 2025-07-01
STYLE: Prepare to format c++ codebase with clang-format

hjmjohnson:
In preparation for a large number of qt5->qt6 changes, provide an auto-enforced formatting enforcement.  This pre-commit hook automatically ensures that the codebase is consistent, thereby removing a huge manual effort to make  consistent formatting with qt5->qt6 changes.

Add clang-format configuration & associated exceptions

In preparation for the changes needed for qt6 compatibility, a standardized code formatter that is automated will make the future diffs smaller, easier to review, and easier to make consistent code formatting.

Related-to:
* #7603

Resolves:
* #5211

Update pre-commit config to include "clang-format"

By adding `clang-format` to the pre-commit configuration, it enables the following:

* GitHub Action report: Suggested changes are reported in the error log associated with the pre-commit GitHub Action workflow.
* Consolidated local pre-commit checks and updates by running:

  ```
  pip install pre-commit
  pre-commit run --all-files
  ```

PR #7603, opened 2024-02-23
STYLE: Consistent formatting and indentation of C++ sources

jcfr:
This pull request ensures consistent indentation and formatting of for C++ source files.

It introduces `clang-format` configuration files as well as `pre-commit` configuration ensuring that IDE can appropriately format the code and continuous integration workflow can report when improperly formatted code is contributed.

Once integrated, this commit will be added to the `.git-blame-ignore-revs`[^1] file.

[^1]: https://github.com/Slicer/Slicer/blob/main/.git-blame-ignore-revs

## Open Questions

* Should we set `AlignConsecutiveAssignments`[^2] to `true` ?
* Should we customized the penalties[^3] ?
* Should we add additional `clang-format` exceptions ?

[^2]: https://clang.llvm.org/docs/ClangFormatStyleOptions.html#alignconsecutiveassignments
[^3]: https://clang.llvm.org/docs/ClangFormatStyleOptions.html#penaltybreakassignment

### Penalties

Default value for penalties are defined as such[^4]:

```cpp
  LLVMStyle.PenaltyBreakAssignment = prec::Assignment;
  LLVMStyle.PenaltyBreakBeforeFirstCallParameter = 19;
  LLVMStyle.PenaltyBreakBeforeMemberAccess = 150;
  LLVMStyle.PenaltyBreakComment = 300;
  LLVMStyle.PenaltyBreakFirstLessLess = 120;
  LLVMStyle.PenaltyBreakOpenParenthesis = 0;
  LLVMStyle.PenaltyBreakScopeResolution = 500;
  LLVMStyle.PenaltyBreakString = 1000;
  LLVMStyle.PenaltyBreakTemplateDeclaration = prec::Relational;
  LLVMStyle.PenaltyExcessCharacter = 1'000'000;
  LLVMStyle.PenaltyIndentedWhitespace = 0;
  LLVMStyle.PenaltyReturnTypeOnItsOwnLine = 60;
```

[^4]: https://github.com/llvm/llvm-project/blob/llvmorg-20.1.7/clang/lib/Format/Format.cpp#L1663-L1674

PR #8540, opened 2025-07-08
STYLE: Remove executable permission from non-script text files

jcfr:
Identified using:

```
find . -type f -perm /111 ! -name "*.sh" ! -name "*.sh.in" -exec file {} + | grep ':.*text'
```

Permission removed using:

```
chmod -x /path/to/file
```

These files are plain text and should not be marked as executable.


jcfr:
These files were identified while reviewing changes related to these pull requests:
* https://github.com/Slicer/Slicer/pull/8525
* https://github.com/Slicer/Slicer/pull/7603

PR #8552, opened 2025-07-13
BUG: Incorrect number of arguments given when weightAdjustment was desired.

hjmjohnson:
Assume that the last "0.5" argument was supposed to be the
WeightAdjustment instead of weight.

```c++
setActionPosition(QAction* action, int section, int weight /*=0*/, double weightAdjustment /*=0.0*/)
```

Implicit conversion from 'double' to 'int' changes value from 0.5 to 0


lassoan (approved):
Thank you, it looks good. I've just adjusted the comments to make them consistent and modified the 0.5 to 1.0 (the weight modifier is already scaled to be much smaller than the weight, so there is no need to use <1 value).

hjmjohnson:
@lassoan Thank you!

PR #8554, opened 2025-07-13
COMP: Remove redundant Slicer_HAVE_QT5 definitions

hjmjohnson:
Simplified codebase by eliminating unnecessary Slicer_HAVE_QT5
references, ensuring compatibility with Qt5+ and improving
maintainability.

Slicer_HAVE_QT5=0 was needed for Qt4 support features
Slicer_HAVE_QT5=1 will be required for enabling Qt6 compatibility as well

Prefer #if QT_VERSION < QT_VERSION_CHECK(#, #, #) for specific version
range compatibility checking.


jamesobutler:
This is a follow-up to https://github.com/Slicer/Slicer/commit/8b6d8f5a434a6ce7869bfd6e0267e19e121ff672

jcfr:
As detailed in e88cb66496584d5ac09ca54da6368df38aaf5352, we may have to revert this:

> vtkSlicerConfigure: Introduced Slicer_HAVE_QT5 and Slicer_HAVE_WEBKIT_SUPPORT.
> This was needed to workaround limitation of "moc" compiler preventing it
> from expanding macro like "QT_VERSION_CHECK.

jcfr:
@jamesobutler and @hjmjohnson  Unless we can confirm that the limitation of the moc compiler has been addressed, I suggest to move forward with:
* https://github.com/Slicer/Slicer/pull/8562

PR #8555, opened 2025-07-13
COMP: Use dynamic_cast for inheritance upcasting

hjmjohnson:
This is a followup on the previous isolated case to
apply the same fixes in the Slicer codebase widely.

Even though event is declared as a QEvent*, the compiler cannot
guarantee that it actually points to a QMouseEvent. The classes aren’t
related through inheritance in a static type-safe way.

Casting across siblings in the event class hierarchy is invalid.

Use dynamic_cast<> for casting between heirchies that do *not*
  derive from QObject (i.e. QEvents do not derive from QObjects)

Use qobject_cast<> for casting between QObjects.
  Qt provides qobject_cast<T*>(QObject*) as a safe way to cast between
  QObject-derived types using Qt’s meta-object system (requires Q_OBJECT
  in the class).  It is RTTI‑free, works across shared libraries, and
  is available in both Qt 5 and Qt 6  ￼ ￼.
    •   The function is declared in qobject.h, which is indirectly
        included by <QObject> in both Qt versions

### DIFF
--- a/Base/QTGUI/qSlicerModuleFinderDialog.cxx
+++ b/Base/QTGUI/qSlicerModuleFinderDialog.cxx
@@ -306,9 +306,9 @@ bool qSlicerModuleFinderDialog::eventFilter(QObject* target, QEvent* event)
     // widget in the tab order)
     if (event->type() == QEvent::KeyPress)
     {
-      QKeyEvent* keyEvent = static_cast<QKeyEvent*>(event);
       qSlicerModuleFactoryFilterModel* filterModel = d->ModuleListView->filterModel();
-      if (keyEvent != nullptr && filterModel->rowCount() > 0)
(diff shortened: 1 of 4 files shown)


jcfr:
The use of `static_cast` is appropriate here, as the event type is explicitly verified before casting. Additionally, checking whether the pointer is null is unnecessary in this context.

This approach is consistent with Qt’s own codebase:

```
$ pwd
/path/to/Qt/5.15.2/Src

$ ack "static_cast.+event" | wc -l
1276

$ ack --cpp "dynamic_cast.+event" | wc -l
6
```

lassoan:
The idea of using dynamic_cast was valid and the explanation of the current implementation was useful. The result of this discussion should be not just a closed pull request but comments in the code that explains the rationale of the current implementation for future developers. I've submitted a pull request that does this: https://github.com/Slicer/Slicer/pull/8563

PR #8559, opened 2025-07-13
COMP: Add Qt version check for QProcess start method usage

hjmjohnson:
- Use `startCommand` for Qt6+ while retaining `start` for older versions. Enhances compatibility with updated Qt API.


Base/QTCLI/qSlicerCLIExecutableModuleFactory.cxx
```diff
@@ -141,7 +141,11 @@ QString qSlicerCLIExecutableModuleFactoryItem::runCLIWithXmlArgument()
   QProcessEnvironment env = QProcessEnvironment::systemEnvironment();
   env.insert("ITK_AUTOLOAD_PATH", "");
   cli.setProcessEnvironment(env);
+#if QT_VERSION > QT_VERSION_CHECK(6, 0, 0)
+  cli.startCommand(this->path(), QStringList(QString("--xml")));
```
jamesobutler: Shouldn’t the command be a single string per https://doc.qt.io/qt-6/qprocess.html#startCommand? The second arg is for the OpenMode.

jamesobutler:
Based on latest [push](https://github.com/Slicer/Slicer/compare/c610896b7ca5c933f1642ac67c0d5d8bb502be01..8d732a92f10e5fe743b9bc4b9019e613fab2c16a), the commit message (`COMP: Add Qt version check for QProcess start method usage`) should change because there is no longer a Qt version check. You decided against the `startCommand` usage?

Base/QTCLI/qSlicerCLIExecutableModuleFactory.cxx
hjmjohnson: Good catch! 

I found a way that is more explicit and allows for both forward and backward compatibility, and does not require compilation-time code selection.

hjmjohnson:
> Based on latest [push](https://github.com/Slicer/Slicer/compare/c610896b7ca5c933f1642ac67c0d5d8bb502be01..8d732a92f10e5fe743b9bc4b9019e613fab2c16a), the commit message (`COMP: Add Qt version check for QProcess start method usage`) should change because there is no longer a Qt version check. You decided against the `startCommand` usage?

Yes, I did come across a different route for fixing that did not require the compile time checks.  I fixed the commit message to reflect what is being proposed.

jamesobutler (commented):
Is this change actually needed? The existing usage of `start` here in Slicer is using the 2 inputs of program and arguments rather than the single input of command which was the deprecated method in Qt 5.15 that was then removed in Qt6.

hjmjohnson:
@jamesobutler I was preparing for Qt6.  My understanding was that the 2-argument version is not available in Qt6.  The recommendation I found to future-proof was to use QProcess::setProgram() and setArguments()

```c++
cli.setProgram(this->path());
cli.setArguments(QStringList{"--xml"});
cli.start();
```

I am NOT a Qt expert, so I would take your advice.

jamesobutler:
The passing of the "Command" string to the `start` method is not available in Qt 6 where the command string is a combination of the program and argument flags. However 3D Slicer has not been using this version of the `start` method. It is using the version where the first argument is the program and the second argument is the list of arguments. This is available in both Qt 5.15 and Qt 6. There would only need to be a compatibility fix if the deprecated version of `start` was being used, but that is not the case here. So there appears to be no action required for this piece of code to work for Qt 5.15 and Qt 6.

Qt 5.15:
- [start](https://doc.qt.io/archives/qt-5.15/qprocess.html#start)(const QString &program, const QStringList &arguments, QIODevice::OpenMode mode = ReadWrite)
- [start](https://doc.qt.io/archives/qt-5.14/qprocess.html#start-1)(const QString &command, QIODevice::OpenMode mode = ReadWrite)
  - ^ Deprecated starting with Qt 5.15
- [start](https://doc.qt.io/archives/qt-5.15/qprocess.html#start-2)(QIODevice::OpenMode mode = ReadWrite)

Qt 6:
- [start](https://doc.qt.io/qt-6/qprocess.html#start)(const QString &program, const QStringList &arguments = {}, QIODeviceBase::OpenMode mode = ReadWrite)
- [start](https://doc.qt.io/qt-6/qprocess.html#start-1)(QIODeviceBase::OpenMode mode = ReadWrite)
- [startCommand](https://doc.qt.io/qt-6/qprocess.html#startCommand)(const QString &command, QIODeviceBase::OpenMode mode = ReadWrite)
  - ^ New for Qt6 and replaces the deprecation of `start` from Qt 5.15 where "command" was passed.

jcfr:
Thanks @hjmjohnson and @jamesobutler 

Inspecting the history associated with `QProcess` allows to confirm that the removed deprecated method was `QProcess::start(const QString &command, QIODevice::OpenMode mode)`[^1] and not `QProcess::start(const QString &program, const QStringList &arguments, OpenMode mode)`[^2]. 

For this reason, I am closing this pull request.

[^1]: https://github.com/qt/qtbase/commit/ee92ee7a3cf7ef6017a2950a590005abfd37cf1a
[^2]: https://github.com/qt/qtbase/blob/6.10/src/corelib/io/qprocess.cpp#L2174-L2229

hjmjohnson:
@jcfr @jamesobutler Thanks for the information.

PR #8563, opened 2025-07-14
STYLE: Explain static_cast use in Qt event handling

lassoan:
Qt uses type ID check + static_cast for casting QEvent to more specific classes for performance reasons. Using dynamic_cast would be slower due to complexity of RTTI. qobject_cast does not use RTTI, so it is faster, but Qt event objects are not derived from QObject, so it cannot be used here.

See more discussion at https://github.com/Slicer/Slicer/pull/8555


jcfr (approved):

hjmjohnson (approved):
Thanks Andras.  The comment is very helpful.  Static analysis tools flag these because they are usually a common source of bugs.  This implementation is not a bug.

PR #8567, opened 2025-07-14
COMP: Replace QMap with QMultiMap for field handling consistency

hjmjohnson:
Updated `registerEmptyFieldNames` and related methods to use
QMultiMap instead of QMap for improved handling of multiple
entries per key. Adjusted corresponding usage to ensure
compatibility.

QMap::insertMulti(...) is deprecated. Use QMultiMap instead,
using insert(...) directly.

Replace operator[] with value() and insert()

Refactor usage of `QMap` and `QHash` to replace operator[] with
`value()` for retrieval and `insert()` for insertion. This improves
compatibility with Qt6 and ensures consistent behavior when key access
is involved. All related tests and modules updated accordingly.


Base/QTApp/qSlicerMainWindow.cxx
```diff
@@ -1471,8 +1471,8 @@ void qSlicerMainWindow::onFileSaved(const qSlicerIO::IOProperties& filePropertie
     // which can cause complication when attempted to be stored,
     // therefore we create a new clean property set.
     qSlicerIO::IOProperties properties;
-    properties["fileName"] = fileName;
-    properties["fileType"] = QString("SceneFile");
+    properties.insert("fileName", fileName);
+    properties.insert("fileType", QString("SceneFile"));
```
jcfr, line 1475: Since the use of `insert()` will add duplicated entry, we should revisit to use `replace()`.

> If there is already an item with the same key in the map, this function will simply create a new one. (This behavior is different from replace(), which overwrites the value of an existing item.)

Source:
* https://doc.qt.io/archives/qt-5.15/qmultimap.html#insert
* https://doc.qt.io/qt-6/qmultimap.html#insert-1

jcfr (changes requested):

lassoan:
I would keep `QVariantMap` for `qSlicerIO::IOProperties`. Is there something wrong with it in Qt6?

hjmjohnson:
@lassoan 
```
QMap::unite() is removed in Qt 6.
For duplicating keys, use QMultiMap.
Migrate by replacing unite() calls with explicit insert() or += into a QMultiMap.
```

`QVariantMap` is an alias for `Map<QString, QVariant>`

I think (I could be wrong) we need a `QMultiMap<QString, QVariant>`, but perhaps I had interpreted the previous code incorrectly (See @jcfr comments).   

If I made a mistake here, I likely made the same mistake in CTK   in https://github.com/commontk/CTK/pull/1242. @ https://github.com/commontk/CTK/pull/1242/commits/a3a82e20e2c8bafd20562ad31d818f22872d88dd

lassoan:
Thanks for the clarification. We don't need a multimap, just need to replace `unite()` call with `insert()`.

It is crazy that clazy suggested to switching to an entirely different data structure, instead of just switching the deprecated method to an equivalent non-deprecated method. It may worth sending a bug report to clazy if this idea came from it, because such a bad suggestion could mislead a lot of people.

lassoan:
> QMap::unite() is removed in Qt 6.
For duplicating keys, use QMultiMap.
Migrate by replacing unite() calls with explicit insert() or += into a QMultiMap.

Where does this message come from?

I think it is misleading because of a typo. The last line should be `Migrate by replacing unite() calls with explicit insert() or += into the QMap.`. And I would also suggest removing the line `For duplicating keys, use QMultiMap.` (or at least move it to the end), because it is irrelevant: QMap cannot have duplicate keys, so we know that those who have been using QMap don't need duplicate keys, so they don't need to switch to QMultiMap.

hjmjohnson:
@lassoan I must have gone down the wrong path. Clazy did not make a recommendation, but I did find the recommendation somewhere else.  It is probably my fault.  I'm working on reverting that now.

lassoan:
I've found this in clazy:

https://invent.kde.org/sdk/clazy/-/blob/master/src/checks/manuallevel/qt6-deprecated-api-fixes.cpp#L699

So, I've posted on the KDE forum: https://discuss.kde.org/t/misleading-qt6-upgrade-instructions-for-qmap-unite-deprecation/37054

hjmjohnson:
@jcfr I fixed the comment to reflect what is being changed.  This did not actually use `QMutliMap`, it had been reverted to `QVariantMap` as suggested by @lassoan.

`insert()` is the correct function to use for a `QVariantMap`.

Base/QTGUI/qSlicerDataDialog.cxx
```diff
@@ -350,16 +350,16 @@ QList<qSlicerIO::IOProperties> qSlicerDataDialogPrivate::selectedFiles() const
...
     // TBD: fileType is not good enough to describe what reader to use
-    properties["fileType"] = descriptionComboBox->itemData(descriptionComboBox->currentIndex()).toString();
+    properties.insert("fileType", descriptionComboBox->itemData(descriptionComboBox->currentIndex()).toString());
     qSlicerIOOptionsWidget* optionsItem = dynamic_cast<qSlicerIOOptionsWidget*>(this->FileWidget->cellWidget(row, OptionsColumn));
     if (optionsItem)
     {
-      // The optionsItem contains all the file properties including "fileName"
-      properties.unite(optionsItem->properties());
+      // The optionsItem contains all the file properties, including "fileName"
+      properties.unite(optionsItem->properties()); // Use += after CTK is updated so that the types match
```
jcfr: One last nitpick: Instead of including this comment, could you create an issue ?

lassoan:
We can keep things simple: Leave everything as is (don't change `[]` to `value` or `insert`), just replace those few `unite` method calls with `insert`.

hjmjohnson:
I will revisit this when it becomes important.  It is becoming a distraction from more important changes.

PR #8569, opened 2025-07-14
COMP: Replace Slicer_HAVE_QT5 with versioned Slicer_QT_MAJOR_VERSION define

jcfr:
This commit replaces the legacy `-DSlicer_HAVE_QT5` preprocessor definition with a more general `-DSlicer_QT_MAJOR_VERSION=${Slicer_QT_MAJOR_VERSION}` in all relevant build macros:
* Update `SlicerMacroBuild` macros to pass `Slicer_QT_MAJOR_VERSION` instead of `Slicer_HAVE_QT5`
* Replace `#cmakedefine Slicer_HAVE_QT5` with a versioned define block in `vtkSlicerConfigure.h.in`
* Remove the obsolete `set(Slicer_HAVE_QT5 1)` in the top-level `CMakeLists.txt`

This improves compatibility with future Qt major versions while preserving support for existing extensions relying on `Slicer_HAVE_QT5`.


jcfr:
I need to revisit ...

```
/home/jcfr/Projects/Slicer-Release/Slicer-build/vtkSlicerConfigure.h:47:29: error: operator '==' has no left operand
   47 | #if Slicer_QT_MAJOR_VERSION == 5
      |                             ^~

```

PR #8574, opened 2025-07-15
COMP: signed vs unsigned compiler warning

hjmjohnson:
Remove signed to unsigned comparison warnings

### DIFF
--- a/Libs/MRML/Core/vtkMRMLColorTableNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLColorTableNode.cxx
@@ -1253,10 +1253,10 @@ int vtkMRMLColorTableNode::SetColor(int entry, const char* name, double r, doubl
   }
 
   this->GetLookupTable()->SetTableValue(entry, r, g, b, a);
-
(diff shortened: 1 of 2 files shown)


Libs/MRML/Core/vtkMRMLColorTableNode.cxx
```diff
@@ -1254,9 +1254,14 @@ int vtkMRMLColorTableNode::SetColor(int entry, const char* name, double r, doubl
 
   this->GetLookupTable()->SetTableValue(entry, r, g, b, a);
 
-  if (static_cast<size_t>(entry) >= this->Properties.size())
+  if (entry < 0)
```
lassoan: <0 is already checked above (IsValidColorIndex), so we don't need to check again

lassoan (approved):
I've removed the unnecessary extra checks and forced-push. It looks good now.

Libs/MRML/Core/vtkMRMLColorTableNode.cxx
hjmjohnson: Fixed by lassoan

PR #6708, opened 2022-11-28
COMP: Switch from zlib 1.2.3 to maintained zlib-ng 2.2.4 fork

jamesobutler:
## Background
Slicer is currently using zlib version 1.2.3 originally released July 18th, 2005. Slicer has been using a [fork](https://github.com/commontk/zlib) in the CommonTK repo where zlib was CMake'ified since at least 2012. Since then zlib has made it onto GitHub at https://github.com/madler/zlib and also utilizes CMake. The latest release is [1.2.13](https://github.com/madler/zlib/tree/v1.2.13) released on Oct 13th, 2022.

## Changes in this PR
This PR updates zlib used in Slicer to now be the modern forked version known as [zlib-ng](https://github.com/zlib-ng/zlib-ng).  Read the [history](https://github.com/zlib-ng/zlib-ng#history) about zlib-ng. I have configured zlib-ng to be built in zlib-compat mode rather than zlib-ng native mode. This decision matches what ITK has done as they switched to using zlib-ng as part of the ITK 5.3.0 release. You can review the [PORTING](https://github.com/zlib-ng/zlib-ng/blob/b3dcf11b4204a16fde71fa8224d7b7054e225b93/PORTING.md) instructions to learn more about the advantages/disadvantages/considerations about the two main build options.

The version of zlib-ng being used here is based on https://github.com/zlib-ng/zlib-ng/commit/b3dcf11b4204a16fde71fa8224d7b7054e225b93 which is the same version that ITK 5.3.0 is using. This is along the zlib-ng 2.1.0 development period and zlib-ng is currently using zlib version 1.2.12 as their base. I'm currently using a fork of zlib-ng with a slight code change that matches https://github.com/InsightSoftwareConsortium/ITK/commit/3bcc933052bbe0c5d303af6ab948704450471788 to make it work in Slicer, but someone might find a better configuration or other changes to avoid this slight code change. If the code change is maintained, I suggest creating a new forked version of zlib-ng in the Slicer GitHub organization rather than using the one under my name.

Using the Slicer python console, the zlib version can be reviewed by doing the following:
```python
import zlib
# [Qt] Python console user input: import zlib
zlib.ZLIB_VERSION
# '1.2.12.zlib-ng'
# [Qt] Python console user input: zlib.ZLIB_VERSION
```

## Testing:
I have built on Windows successfully and all tests continued to pass for me (`1>100% tests passed, 0 tests failed out of 662`). If someone could provide additional testing on the macOS and Linux platforms that would be great.

[...]

PR #8575, opened 2025-07-15
COMP: Override cmake minimum policy for external

hjmjohnson:
Many packages with very simple cmake needs do not require features from newer versions of cmake, but work just fine with almost all versions of cmake.

Use `-DCMAKE_POLICY_VERSION_MINIMUM:STRING=3.20.6` to override the minimum setting to a higher policy version.

This is a stopgap measure to decouple building with later versions of CMake without needing to update the upstream code.

---

> [!IMPORTANT]
> This pull request should be rebased & integrated (or closed) *only* after the pull requests referenced below are themselves reviewed & integrated.

> [!NOTE]
> Changes originally integrated through this pull request have been extracted and/or contributed as independent pull request by @jcfr:
> * https://github.com/Slicer/Slicer/pull/8597
>   * https://github.com/Slicer/Slicer/pull/6708
>   * https://github.com/Slicer/Slicer/pull/8596
> * https://github.com/Slicer/Slicer/pull/8576
> * https://github.com/Slicer/Slicer/pull/8594
> * https://github.com/Slicer/Slicer/pull/8595
> * https://github.com/Slicer/Slicer/pull/8629


SuperBuild/External_LibArchive.cmake
```diff
@@ -59,6 +59,7 @@ if((NOT DEFINED LibArchive_INCLUDE_DIR
     BINARY_DIR ${EP_BINARY_DIR}
     INSTALL_DIR ${EP_INSTALL_DIR}
     CMAKE_CACHE_ARGS
+      -DCMAKE_POLICY_VERSION_MINIMUM:STRING=3.20.6
```
jamesobutler, line 62: This increased policy version minimum for LibArchive results in a build error on Windows that can be resolved by updating LibArchive to a newer version that includes https://github.com/libarchive/libarchive/commit/3a5a34b8211b68709090fcac71c45d9b5f59f4f2.

jcfr (changes requested):
Generally speaking, we should update the upstream project and avoid passing such variable.

jamesobutler:
@jcfr I currently have a branch (https://github.com/jamesobutler/Slicer/tree/libarchive-update-for-newer-cmake) I was testing for LibArchive that I was testing, but then realized the branch has the base which has the issues mentioned in https://github.com/Slicer/Slicer/pull/8582#issuecomment-3089720488 so I haven't been able to confirm the automated testing is appropriate.

SuperBuild/External_LibArchive.cmake
jcfr: Update of LibArchive contributed by @jamesobutler is available for review at:
* https://github.com/Slicer/Slicer/pull/8594

jcfr:
Closing. All projects originally referenced in this pull request have been updated independently.

Issue #7621, opened 2024-03-05
Modernize C++ sources using clang-tidy


jcfr:
## Is your feature request related to a problem? Please describe.

Following the update of the minimum C++ version required to build Slicer (now C++17[^1]), existing code can be improved to use simpler and more modern construct now built-in the c++ languages. 

[^1]: https://github.com/Slicer/Slicer/pull/6237

## Describe the solution you'd like

The `modernize-` checks offered by `clang-tidy`[^2] should be selectively applied to update the sources and then systematically executed through a pre-commit hook.

[^2]: https://clang.llvm.org/extra/clang-tidy/checks/list.html

## Describe alternatives you've considered

NA

hjmjohnson:
This is also relevant for updating from Qt5 to Qt6 where some of the migration tools are built upon clang-tidy.

```bash 
# apt-get install clazy clang-tools-19
export CLAZY_EXPORT_FIXES=ON
export CMAKE_EXPORT_COMPILE_COMMANDS=ON
export CMAKE_GENERATOR=Ninja

export CLAZY_CHECKS="qt6-deprecated-api-fixes"
export CLAZY_CHECKS="qt6-header-fixes"
export CLAZY_CHECKS="qt6-qhash-signature"
export CLAZY_CHECKS="qt6-qhash-signature"
export CLAZY_CHECKS="qt6-fwd-fixes"
export CLAZY_CHECKS="missing-qobject-macro"

export SRC_DIR=/localcache/Users/johnsonhj/src/SlicerQt
export BLD_DIR=$(dirname ${SRC_DIR})/Qt5clazySlicer-build
export INNER_BLD_DIR=${BLD_DIR}/Slicer-build


# use clazy as your compiler
#
cmake -GNinja -S ${SRC_DIR} -B ${BLD_DIR} -DCMAKE_CXX_COMPILER=clazy 
#                                                                                                                              ^^^^
cd ${BLD_DIR}
cmake --build ${BLD_DIR}
```

Apply found changes (written by above build to filename.clazy.yaml

```bash
#/bin/bash

clazy_yaml_file=$1
if [ ! -f ${clazy_yaml_file} ]; then
   echo "ERROR: Missing file ${clazy_yaml_file}"
   exit -1
fi
clazy_yaml_dir=$(dirname ${clazy_yaml_file})

clang-apply-replacements-19 ${clazy_yaml_dir}

for changed_file in $(git diff --name-only); do
  # Change tabs to 2 spaces
[...]

PR #8583, opened 2025-07-17
COMP: Prefer explicit const designation

hjmjohnson:
Add const to variables that do not change.

Detect local variables which could be declared as const but are not. Declaring
variables as const is recommended by many coding guidelines, such
as: ES.25 from the C++ Core Guidelines.


jcfr:

Would it be possible to split it into two pull requests ?

### Pull request (1): `STYLE: Normalize const qualifier order`

This PR would include two commits:
- `STYLE: Normalize const qualifier order`
- `ENH: Update clang-format configuration setting QualifierAlignment`

This will allow to add `STYLE: Normalize const qualifier order` to `.git-blame-ignore-revs`

### Pull request (2): `COMP: Prefer explicit const designation`

Could you also indicate how the update was performed ?

hjmjohnson:
I am traveling right now without good Internet access and probably won't be able to do that until Sunday or Monday next week at the earliest.  I have a draft Pohl request with the Scripps that I've been using just for documentation purposes one of the scripts in that PR has a clan mode in the modernization something bash script that was the starting point.

hjmjohnson:
I can't compile slicer on my Mac and I may not have VPN access to my UNIX computer for up to two weeks.  I'll need to see if VPN works in foreign country.

jcfr:
> Pull request (1)

* https://github.com/Slicer/Slicer/pull/8586

lassoan:
Sprinkling const all over the code inside methods seems to have a very low benefit/cost ratio

Some costs:

1. It adds noise - extra code (that takes a little more cognitive load to read and write) without a clear benefit.

For example, what is the point of adding `const` here?

```
QWidget* qMRMLEventBrokerWidgetPlugin::createWidget(QWidget* parentWidget)
{
  qMRMLEventBrokerWidget* const newWidget = new qMRMLEventBrokerWidget(parentWidget);
  return newWidget;
}
```

The current code reflects what the developer wanted to do, just create a new widget of a certain class. The developer could have been less specific just by writing

```
auto = new qMRMLEventBrokerWidget(parentWidget);
```

or 

```
  return new qMRMLEventBrokerWidget(parentWidget);
```

but the developer decided to just give the information to the reader that a the created widget is a `qMRMLEventBrokerWidget`. Const or not does not matter, so it is not mentioned. To me the current implementation makes sense and adding the const would be just extra useless information.

2. Adding const wherever it is possible could get in conflict with modern C++ concepts, such as move semantics.

3. Code churn. It takes time for developers to resolve merge conflicts. Maybe not a huge amount of work, but still annoying (especially when there is no clear benefit for that developer) and chance of making mistakes due to manually having to resolve some merge conflicts.

Limited/no benefits:

1. No benefits are expected in performance. The determination that locally inside a method a pointer can be const, can be made by the compiler just as well as by the linter,

2. No benefits are expected in robustness or security without also using consts in method APIs. Since VTK, the base of most Slicer classes does not do it, const correctess in method arguments is not feasible (as long as VTK is not changed, and I don't think it would ever happen).

Due to very low benefit/cost ration, I would lean towards not making this change and focus on things that are better use of our time. Maybe we could describe Slicer's const correctness considerations (similar to what I described above but a bit more general) in the developer documentation that we can use as a basis before starting on implementing similar changes in the future.

jamesobutler:
My guess why this change is being proposed is because GitHub CoPilot review actively suggests this type of change in its automated review. With more use of automated GitHub CoPilot review doing this type of change may reduce the noise that it provides. See https://github.com/Slicer/Slicer/pull/8580#pullrequestreview-3027403702 for example.

lassoan:
The comments that copilot made may actually be useful, while the changes suggested by tidy are mostly just noise.

hjmjohnson:
@lassoan I respectfully disagree with the assertions made.

TLDR;  I agree with the philospopies and recommendations of https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#S-philosophy.  I try to follow the best practices of giants in the C++ community like [Bjarne Stroustrup](http://www.stroustrup.com/)
[Herb Sutter](http://herbsutter.com/).

-----------

     - I think that it *ADDS* to readability and understanding, especially in large or complicated functions.  Knowing that a variable is const often take it out of play for consideration when tracking down bugs. Knowing that a variable can not change value 
     - It exposes many places where compile time optimizations can be instrumented with constexpr.
     - Move semantics and const-correctness  are not affected in the changes recommended.   I agree that `Move constructor and move assignment operator must take a non-const rvalue reference`, but those are not included in the recommended changes.
     - I do concede that the benefit is small in almost all cases in this first step toward other code improvements.
     - This is the first step towards some (usually small) compiler optimization improvements.  More important, it allows other clang-tidy auto-fixes to take advange of knowledge that the variables are const for performance/readability improvements that can only be done when working with const (constexpr) values.
     - The use of const correctness enables static code analysis tools to do a better job of identifying concerning code.  It has helped identify rare-case bugs, or performance improvements. 

My experience of running these tools on several large code-bases is that having const-correctness has exposed hidden underlying problems (peformance or otherwise).  
  -- The next code-change that "unlocks" easier conformance to  https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines is 
  
  https://clang.llvm.org/extra/clang-tidy/checks/readability/isolate-declaration.html. 
 
The isolate-declaration has almost no intrinsic value on it's own, but it greatly simplifies future best-practices conformanc changes.

------
I have a set of tools that can automate code conversions to move a codebase to meet the best practices guidelines.  I'm willing to assist with implementing these static analysis tools as described in https://github.com/Slicer/Slicer/issues/7621.

lassoan:
We agree in that const correctness has value and in some cases it may actually overweight the costs. However, this refers to the case when the code overall is const correct, most importantly at the API level (method arguments). Achieving significant degree of const correctness would be difficult due to the large and distributed code base (owned by hundreds of people scattered around the world, maintaining various extensions), and practically impossible due to VTK.

We could still add some consts inside methods, but we should not use such a primitive tool for it that cannot distinguish between trivial cases (where const just adds noise) from cases where adding const has actual positive impact (improve performance, security, readability, etc.). @jamesobutler's example above was very good: those const recommendations from copilot were actually worth considering, they did not add noise or result in lot of code churn, and they could have immedate benefits. clang-tidy is just simply not good enough, I don't think it is worth our time dealing with the code churn and noise that its recommendations are causing.

lassoan:
Last week I worked on ITK code for several full days and I had this discussion in mind the whole time. I felt that adding consts made more sense there (because at many places consts and constexprs were needed, for example due to heavy use of templates) and it was easier to add them (more accurately: it was harder not to add them, because of the already strong typing discipline in the existing code).

However, VTK (and all VTK-based tools, such as a large part of Slicer) follow a different approach - maybe for historical reasons, maybe to be more similar to weakly typed languages. I don't think either is worse, they are just different. VTK is closer to the less strongly typed languages, such as Python.

PR #8586, opened 2025-07-18
STYLE: Normalize qualifier order

jcfr:
This applies formatting to C++ source files based on `clang-format` settings introduced in the following commit.

The updated settings define `QualifierOrder` to enforce consistent qualifier placement (e.g., `const T` instead of `T const`) across the codebase.

The specific ordering was chosen to **minimize the number of changes** observed when running `clang-format` via `pre-commit`, resulting in minimal disruption.

This improves code readability and simplifies future refactoring.

Running `pre-commit` required updating the `.clang-format` configuration with
the following patch:

```diff
diff --git a/.clang-format b/.clang-format
--- a/.clang-format
+++ b/.clang-format
@@ -37,6 +37,14 @@ ForEachMacros:
   - Q_FOREACH
   - Q_FOREVER
 
+QualifierAlignment: Custom
+QualifierOrder:
+  - friend
+  - static
+  - inline
+  - constexpr
+  - const
+  - type
 SortIncludes: false
```

### DIFF
--- a/Modules/CLI/ResampleDTIVolume/itkMatrixExtended.h
+++ b/Modules/CLI/ResampleDTIVolume/itkMatrixExtended.h
@@ -37,7 +37,7 @@ public:
   MatrixExtended(const Superclass& matrix);
   // /Cast the matrix
   template <class C, unsigned int NRowsC, unsigned int NColumnsC>
-  operator MatrixExtended<C, NRowsC, NColumnsC> const();
+  operator const MatrixExtended<C, NRowsC, NColumnsC>();
   Self operator=(const Self& matrix);
 
   Self operator=(const Superclass& matrix);
--- a/Modules/Loadable/Segmentations/EditorEffects/qSlicerSegmentEditorPaintEffect.cxx
+++ b/Modules/Loadable/Segmentations/EditorEffects/qSlicerSegmentEditorPaintEffect.cxx
@@ -944,7 +944,7 @@ QIcon qSlicerSegmentEditorPaintEffect::icon()
 }
 
 //---------------------------------------------------------------------------
-QString const qSlicerSegmentEditorPaintEffect::helpText() const
+const QString qSlicerSegmentEditorPaintEffect::helpText() const
 {
   return QString("<html>")
          + tr("Paint with a round brush<br>."
--- a/.clang-format
+++ b/.clang-format
@@ -37,6 +37,14 @@ ForEachMacros:
   - Q_FOREACH
   - Q_FOREVER
 
+QualifierAlignment: Custom
+QualifierOrder:
+  - friend
+  - static
+  - inline
+  - constexpr
+  - const
+  - type
 SortIncludes: false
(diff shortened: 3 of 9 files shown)


Modules/Loadable/Segmentations/EditorEffects/qSlicerSegmentEditorPaintEffect.cxx
```diff
@@ -944,7 +944,7 @@ QIcon qSlicerSegmentEditorPaintEffect::icon()
 }
 
 //---------------------------------------------------------------------------
-QString const qSlicerSegmentEditorPaintEffect::helpText() const
+const QString qSlicerSegmentEditorPaintEffect::helpText() const
```
jcfr, line 947: In a follow-up commit, we will look into changing `const QString` to `QString` as it is meaningless because of `return-by-value`

Modules/CLI/ResampleDTIVolume/itkMatrixExtended.h
```diff
@@ -37,7 +37,7 @@ class MatrixExtended : public Matrix<T, NRows, NColumns>
   MatrixExtended(const Superclass& matrix);
   // /Cast the matrix
   template <class C, unsigned int NRowsC, unsigned int NColumnsC>
-  operator MatrixExtended<C, NRowsC, NColumnsC> const();
+  operator const MatrixExtended<C, NRowsC, NColumnsC>();
```
jcfr, line 40: I suspect the original intent was to have the signature:

```cpp
template <class C, unsigned int NRowsC, unsigned int NColumnsC>
operator MatrixExtended<C, NRowsC, NColumnsC>() const;
```

Modules/Loadable/Segmentations/EditorEffects/qSlicerSegmentEditorPaintEffect.cxx
lassoan: const may also interfere with move semantics so it could make sense to avoid it. However, due to potential use of the current method signature in modules in extensions, the change may not worth the trouble. We should do it if we must do backward-incompatible changes nearby anyway (for example, for splitting Qt-specific parts of segment editor effects from VTK-based logic part).

hjmjohnson: I think that const on the return type is meaningless and ignored for `return-by-value`.  If I remember correctly, the signature of the function is not changed because the return const has been ignored by the compiler (this assumption may not be true for all compilers).

Modules/CLI/ResampleDTIVolume/itkMatrixExtended.h
hjmjohnson: @jcfr I agree,  there are warnings that many functions can be made "const" from clang-tidy, but I have not yet worked on addressing those.

PR #8587, opened 2025-07-18
BUG: Select first component when RGBA volume rendering is selected

Sunderlandkyl:
When selecting RGBA volume rendering the intended behavior was to select the first component, which is used to control the volume opacity. Instead, the first component was selected when "independent" was enabled.

Fixed by using the check state of the RGBColorsRadioButton instead of IndependentRadioButton.


lassoan (approved):
The change looks good. We can integrate this quickly.

However, it would be better if a transient change (e.g., for whatever reason the independentComponents flag is flipped and then flipped back) should not make a change in the widget's state. It would be better if in case RGB is enabled then the spinbox would be simply disabled (value not changed) and its value would be ignored (the 0 component would be used).

PR #8588, opened 2025-07-18
ENH: Avoid changing selected volume rendering component when switching modes

Sunderlandkyl:
This commit leaves the value of the spinbox when switching between RGBA/independent component modes. The value in the spinbox is only used in independent mode, otherwise the selected component is always 0.

See:
https://github.com/Slicer/Slicer/pull/8587#pullrequestreview-3034451745


Sunderlandkyl:
@lassoan I think this PR has the behavior you mentioned in https://github.com/Slicer/Slicer/pull/8587#pullrequestreview-3034451745

lassoan (approved):
Perfect, yes, this was that I had in mind. Thank you!

PR #8589, opened 2025-07-18
BUG: Fix crashes caused by range-based loops over temporary Qt containers

jcfr:
This commit fixes regressions introduced in f06fab25fc ("COMP: Prefer C++11 range-for to Qt macros", 2025-07-16), which replaced `Q_FOREACH` with C++11 range-based `for` loops.

A segmentation fault was reported in `qSlicerWelcomeModuleWidgetPrivate::setupUi`, caused by iterating over a temporary `QWidgetList` constructed inline using the `<<` operator. Since range-based `for` loops do not extend the lifetime of temporaries, the container was destroyed after the loop header executed, leading to a **use-after-free** and a crash when dereferencing the iterator.

Similar crashes were observed in tests on Unix systems involving temporary `QStringList` instances. Although `QStringList` (a `QList<QString>`) uses implicit sharing, range-based iteration over a destroyed temporary is still undefined behavior and may result in a crash.

Additionally, a related use in `qSlicerCoreApplicationPrivate::discoverSlicerHomeDirectory`, which used `QStringList` in the same way, was only compiled on Windows. While this instance did not crash, it was also updated for correctness and consistency.

To address the issue this commit updates all affected range-based loops by:

* Assigning the container to a named local variable prior to iteration, ensuring its lifetime extends through the loop.

For reference, the previously used `Q_FOREACH` macro avoided this issue by explicitly storing the container in a local `_container_` variable:

```cpp
for (auto _container_ = QtPrivate::qMakeForeachContainer(container);
     _container_.i != _container_.e;
     ++_container_.i)
  if (variable = *_container_.i; false) {} else
```

--- 

Follow-up of the following pull request:
* https://github.com/Slicer/Slicer/pull/8580

### DIFF
--- a/Base/QTCore/Testing/Cxx/qSlicerExtensionsManagerModelTest.cxx
+++ b/Base/QTCore/Testing/Cxx/qSlicerExtensionsManagerModelTest.cxx
@@ -909,8 +909,12 @@ void qSlicerExtensionsManagerModelTester::testUninstallExtension()
       this->installHelper(&model, operatingSystem, extensionId, this->Tmp.absolutePath());
     }
 
-    for (const QString& extensionName : QStringList() //
-                                          << "MarkupsToModel")
+    QStringList extensionNames;
+
+    extensionNames = QStringList() //
+                     << "MarkupsToModel";
+
+    for (const QString& extensionName : extensionNames)
     {
       QVERIFY(this->uninstallHelper(&model, extensionName));
       QVERIFY(!model.isExtensionInstalled(extensionName));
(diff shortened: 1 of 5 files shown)


jamesobutler:
I have a local build that has been going for a little bit. Once it is finished I’ll do an incremental build with these changes to confirm the fix. Will aim to integrate before the factory nightly build process begins.

jcfr:
Thanks @jamesobutler :pray:

jamesobutler:
@jcfr Related to the upcoming nightly build will you have the Linux factory build with the newer GCC (for VTK 9.5.0) available to see the build and testing results tomorrow?

jamesobutler (approved):
My build took a little longer to complete than I expected, but I can confirm the application launches successfully now with this change.

PR #8598, opened 2025-07-23
ENH: Add krb5-gssapi-stub project

jcfr:
This stub library provides dummy implementations of the Kerberos-related libraries required by Qt's `libQt5Network.so`, allowing Slicer to run on Linux systems without installing or linking against the full Kerberos stack.


jcfr:
Once this is finalized, it will resolve the test failures associated with the linux Preview build.

jcfr:
After running the Preview build against the `jcfr:add-gssapi-krb5-stub` branch, we can confirm it is working:

image

_Source: https://slicer.cdash.org/index.php?project=SlicerPreview&date=2025-07-24_

jcfr:
@pieper @lassoan @pieper @jamesobutler Et voila, Linux build environment is now functional :100: 

When you have a chance, I would appreciate a review & approval :rocket:

jamesobutler (approved):
Based on the output of the Linux factory now having successful and consistent testing relative to the other platforms, this reasonably seems like a good plan to proceed.

PR #8604, opened 2025-07-29
ENH: Add DM group and render window API to the abstract view node

Thibault-Pelletier:
To support decoupling 3D Slicer logic from the Qt layer, access to the displayable manager group and to the render window is required. This access is currently only possible through the `qMRMLThreeDView` and the `qMRMLSliceViews`.

This change allow direct access of the render window and the displayable manager group using the abstract view node directly.

To avoid adding a dependency from the `MRMLCore` library to the `MRMLDisplayableManager` library, objects are left as type `vtkObject`.

Access can be done using :

```cpp
vtkRenderWindow::SafeDownCast(viewNode->GetRenderWindow())
vtkMRMLDisplayableManagerGroup::SafeDownCast(viewNode->GetDisplayableManagerGroup())
```

### DIFF
--- a/Libs/MRML/Core/vtkMRMLAbstractViewNode.h
+++ b/Libs/MRML/Core/vtkMRMLAbstractViewNode.h
@@ -23,6 +23,7 @@
 
 // VTK includes
 #include <vtkSmartPointer.h>
+#include <vtkWeakPointer.h>
 
 // MRML includes
 #include "vtkMRMLNode.h"
@@ -298,6 +299,18 @@ public:
   vtkSetMacro(ScreenScaleFactor, double);
   //@}
 
+  //@{
+  /// \brief Get/Set the displayable manager group associated with the view.
+  vtkGetObjectMacro(DisplayableManagerGroup, vtkObject);
+  vtkSetObjectMacro(DisplayableManagerGroup, vtkObject);
+  //@}
+
+  //@{
+  /// \brief Get/Set the render window associated with the view.
+  vtkGetObjectMacro(RenderWindow, vtkObject);
+  vtkSetObjectMacro(RenderWindow, vtkObject);
+  //@}
+
 protected:
   vtkMRMLAbstractViewNode();
   ~vtkMRMLAbstractViewNode() override;
@@ -362,6 +375,9 @@ protected:
 
   static const char* ParentLayoutNodeReferenceRole;
   static const char* InteractionNodeReferenceRole;
+
+  vtkWeakPointer<vtkObject> DisplayableManagerGroup = nullptr;
+  vtkWeakPointer<vtkObject> RenderWindow = nullptr;
 };
 
 //------------------------------------------------------------------------------
--- a/Libs/MRML/Widgets/qMRMLSliceView.cxx
+++ b/Libs/MRML/Widgets/qMRMLSliceView.cxx
@@ -342,6 +342,18 @@ void qMRMLSliceView::setMRMLSliceNode(vtkMRMLSliceNode* newSliceNode)
     return;
   }
 
+  if (d->MRMLSliceNode)
+  {
+    d->MRMLSliceNode->SetDisplayableManagerGroup(nullptr);
+    d->MRMLSliceNode->SetRenderWindow(nullptr);
+  }
+
+  if (newSliceNode)
+  {
+    newSliceNode->SetDisplayableManagerGroup(d->DisplayableManagerGroup);
+    newSliceNode->SetRenderWindow(d->RenderWindow);
+  }
+
   d->qvtkReconnect(d->MRMLSliceNode, newSliceNode, vtkCommand::ModifiedEvent, d, SLOT(updateWidgetFromMRML()));
 
   d->MRMLSliceNode = newSliceNode;
--- a/Libs/MRML/Widgets/qMRMLThreeDView.cxx
+++ b/Libs/MRML/Widgets/qMRMLThreeDView.cxx
@@ -390,6 +390,18 @@ void qMRMLThreeDView::setMRMLViewNode(vtkMRMLViewNode* newViewNode)
     return;
   }
 
+  if (d->MRMLViewNode)
+  {
+    d->MRMLViewNode->SetDisplayableManagerGroup(nullptr);
+    d->MRMLViewNode->SetRenderWindow(nullptr);
+  }
+
+  if (newViewNode)
+  {
+    newViewNode->SetDisplayableManagerGroup(d->DisplayableManagerGroup);
+    newViewNode->SetRenderWindow(d->RenderWindow);
+  }
+
   d->qvtkReconnect(d->MRMLViewNode, newViewNode, vtkCommand::ModifiedEvent, d, SLOT(updateWidgetFromMRML()));
 
   d->MRMLViewNode = newViewNode;


pieper:
This does sound logical. 

Will it be possible to use this from python?  I.e. will the wrapper detect that you've retrieved the subclass if it already knows about the `vtkMRMLDisplayableManagerGroup` via the previously imported library?  If not, then I'd add a method to the `vtkMRMLDisplayableManagerGroup` to return the group from a given `viewNode`

Thibault-Pelletier:
Thanks for the feedback!

On Python's side the safe downcast is not necessary (Python's dynamic typing correctly sends back the actual type).
Related console output below :

```pycon
>>> getNode("vtkMRMLSliceNodeRed").GetRenderWindow()
<vtkmodules.vtkRenderingOpenGL2.vtkGenericOpenGLRenderWindow(0x000001946D751E10) at 0x0000019424B05960>
>>> getNode("vtkMRMLSliceNodeRed").GetDisplayableManagerGroup()
<MRMLDisplayableManagerPython.vtkMRMLDisplayableManagerGroup(0x000001946DF206B0) at 0x0000019424B07B80>
```

Thibault-Pelletier:
@jcfr & @lassoan do you have any inputs on this PR?

Thanks in advance 🙏

Libs/MRML/Core/vtkMRMLAbstractViewNode.h
```diff
@@ -362,6 +375,9 @@ class VTK_MRML_EXPORT vtkMRMLAbstractViewNode : public vtkMRMLNode
 
   static const char* ParentLayoutNodeReferenceRole;
   static const char* InteractionNodeReferenceRole;
+
+  vtkWeakPointer<vtkObject> DisplayableManagerGroup = nullptr;
+  vtkWeakPointer<vtkObject> RenderWindow = nullptr;
```
jcfr, line 380:
```suggestion
  vtkWeakPointer<vtkObject> DisplayableManagerGroup;
  vtkWeakPointer<vtkObject> RenderWindow;
```

jcfr:
Can you also:
* update `qMRMLThreeDViewTest1` to check function like `GetDisplayableManagerGroup()` and `GetRenderWindow()` return the expected value.
* update `vtkMRMLViewNodeTest1` to exercise the API

Libs/MRML/Core/vtkMRMLAbstractViewNode.h
```diff
@@ -298,6 +299,18 @@ class VTK_MRML_EXPORT vtkMRMLAbstractViewNode : public vtkMRMLNode
   vtkSetMacro(ScreenScaleFactor, double);
   //@}
 
+  //@{
+  /// \brief Get/Set the displayable manager group associated with the view.
```
jcfr, line 303: For the two pairs of `Set/Get`, could you add a note indicating that there are no consistency checks done. 

The developer is responsible to make sure the group associated with the view node is effectively associated with RendererWindow/Renderer set on the view.

Libs/MRML/Core/vtkMRMLAbstractViewNode.h
```diff
@@ -298,6 +299,18 @@ class VTK_MRML_EXPORT vtkMRMLAbstractViewNode : public vtkMRMLNode
...
+  vtkGetObjectMacro(DisplayableManagerGroup, vtkObject);
+  vtkSetObjectMacro(DisplayableManagerGroup, vtkObject);
```
jcfr, line 305: At the minimum, the `Set` functions should check the type using `IsA()`[^1] ..

```
  if (obj && !obj->IsA("vtkMRMLDisplayableManagerGroup"))
    {
    vtkErrorWithObjectMacro(this, "SetDisplayableManagerGroup: expected vtkMRMLDisplayableManagerGroup");
    return;
    }
```

[^1]: https://vtk.org/doc/nightly/html/classvtkObjectBase.html#a44ecbcd4243dc6d6cbf619c9f983a438

jcfr (changes requested):
Thanks for working on this 🙏 

With adding the notes documenting the risk for getting into an inconsistent state along with the use of `IsA` and updating tests, we would be in a better position for integration.

___

Now, I do have another proposal ...

Since the `vtkRenderWindow` can be retrieved via the displayable manager group, in an ideal layering scenario we could avoid `SetRenderWindow` entirely and instead implement `GetRenderWindow` as:

```cpp
vtkRenderWindow* vtkMRMLAbstractViewNode::GetRenderWindow() const
{
  vtkMRMLDisplayableManagerGroup* displayableManagerGroup = this->GetDisplayableManagerGroup();
  if (!displayableManagerGroup)
  {
    return nullptr;
  }
  vtkRenderer* renderer = displayableManagerGroup->GetRenderer();
  if (!renderer)
  {
    return nullptr;
  }

  return renderer->GetRenderWindow();
}
```

However, as noted in the commit message, `vtkMRMLAbstractViewNode` lives in the `MRMLCore` library, while `vtkMRMLDisplayableManagerGroup` is in the `MRMLDisplayableManager` library, which itself depends on the `MRMLCore` library.

Because of this dependency direction, `MRMLCore` cannot directly include or reference the vtkMRMLDisplayableManagerGroup` class, hence the current design with `vtkWeakPointer<vtkObject>` and requiring the caller to downcast.

To improve integration and reduce the risk of inconsistent state, we could:

1. **Moving** `vtkMRMLDisplayableManagerGroup` into `MRMLCore`, or
2. Introduce a `vtkMRMLAbstractDisplayableManagerGroup` interface in `MRMLCore`, exposing only the minimal API needed (`GetRenderer()`, `GetRenderWindow()`, and possibly `RequestRender()`).


I do have a preference for (2) with the introduction of `vtkMRMLAbstractDisplayableManagerGroup` as it enforces typing at compile time, and removes the need for a separate `SetRenderWindow` API.

pieper:
I like Jc's option 2 also.

cpinter:
+1 to removing `SetRenderWindow` and JC's option 2. Thanks a lot!

Thibault-Pelletier:
Thanks for the feedbacks!
I will make the different changes and update the PR.

Thibault-Pelletier:
Quick update : 

Option 2 of adding an `vtkMRMLAbstractDisplayableManagerGroup` to MRMLCore seems a bit impractical.

The current API for vtkMRMLDisplayableManagerGroup is : 

- `void Initialize(vtkMRMLDisplayableManagerFactory* factory, vtkRenderer* renderer);`
- `void SetAndObserveDisplayableManagerFactory(vtkMRMLDisplayableManagerFactory* factory);`
- `void AddDisplayableManager(vtkMRMLAbstractDisplayableManager* displayableManager);`
- `int GetDisplayableManagerCount();`
- `vtkMRMLAbstractDisplayableManager* GetNthDisplayableManager(int n);`
- `vtkMRMLAbstractDisplayableManager* GetDisplayableManagerByClassName(const char* className);`
- `void SetRenderer(vtkRenderer* newRenderer);`
- `vtkRenderWindowInteractor* GetInteractor();`
- `void RequestRender();`
- `vtkRenderer* GetRenderer();`
- `void SetMRMLDisplayableNode(vtkMRMLNode* newMRMLDisplayableNode);`
- `static bool IsADisplayableManager(const char* displayableManagerName);`
- `static vtkMRMLAbstractDisplayableManager* InstantiateDisplayableManager(const char* displayableManagerName);`
- `virtual void SetLightBoxRendererManagerProxy(vtkMRMLLightBoxRendererManagerProxy*);`
- `virtual vtkMRMLLightBoxRendererManagerProxy* GetLightBoxRendererManagerProxy();`

Creating a proper API would require the following classes defined in DisplayableManager : 

- `vtkMRMLDisplayableManagerFactory`
- `vtkMRMLAbstractDisplayableManager`
- (maybe optional) `vtkMRMLLightBoxRendererManagerProxy`

I feel that it's a lot of boiler plate and moving code around just to have access to a few dynamic properties on the view node and remove a setter.

jcfr:
Closing. Following this morning discussion[^1], this is now superseded by https://github.com/Slicer/Slicer/pull/8658

[^1]: https://discourse.slicer.org/t/2025-08-19-weekly-meeting/44136/2#p-127746-viewnode-renderwindow-in-the-context-of-segmenteditor-refactoring-4

PR #8607, opened 2025-07-29
ENH: Add widget to select any node for scene views

Sunderlandkyl:
Previously users could only select node categories "display", "view" or "existing" (when editing a scene view). With this change, the scene view dialog now contains an "Advanced" section that allows the user to select any nodes in the scene. This node selection section is synchronized with the "display" , "view", and "existing" node check boxes.

image

### DIFF
--- a/Libs/MRML/Widgets/Resources/UI/qMRMLScreenShotDialog.ui
+++ b/Libs/MRML/Widgets/Resources/UI/qMRMLScreenShotDialog.ui
@@ -6,66 +6,178 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>433</width>
-    <height>385</height>
+    <width>636</width>
+    <height>580</height>
    </rect>
   </property>
   <property name="windowTitle">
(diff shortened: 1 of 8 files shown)


lassoan (commented):
Thank you, this will be great!
The only thing that would be nice to improve is the GUI layout. The two "Capture..." checkboxes should be close to each other (probably one above the other) and overall there are lots of very small inconsistencies in the layout (e.g., "Description" is above the textbox, but "Thumbnail" is on the left of the corresponding section; OK&Cancel button seem to be hardcoded to the right as per macOS style, etc.).

Sunderlandkyl:
> OK&Cancel button seem to be hardcoded to the right as per macOS style

We can center the buttons if that configuration is better.

Sunderlandkyl:
Updated the layout:

image

lassoan (approved):
Thank you, it looks much better. I've tweaked it a bit further (used a frame instead of indentation and moved the Save As button to below the image. It looks good to me now and should be ready to merge.

PR #8610, opened 2025-07-30
BUG: Replace u8‑prefixed unit suffix literals with narrow strings

kovalev0:
Fixes build failures on C++20 due to invalid conversion from 'const char8_t*' to 'const char*' when using u8"\u00b5m"‑style literals for Micrometer, Square Micrometer, Cubic Micrometer, and Microsecond units. All u8"\u00b5…" suffixes have been changed to "\u00b5…" so that AddUnitNodeToScene(..., const char* suffix, ...) no longer rejects them.

Error log excerpt:

```
/usr/src/RPM/BUILD/slicer-5.8.0/Modules/Loadable/Units/Logic/vtkSlicerUnitsLogic.cxx:227:33: error: invalid conversion from 'const char8_t*' to 'const char*' [-fpermissive]
  227 |     "Micrometer", "length", "", u8"\u00b5m", 4, …
      |                                 ^~~~~~~~~~~
      |                                 |
      |                                 const char8_t*
/usr/src/RPM/BUILD/slicer-5.8.0/Modules/Loadable/Units/Logic/vtkSlicerUnitsLogic.cxx:121:34:
note:   initializing argument 5 of 'vtkMRMLUnitNode* vtkSlicerUnitsLogic::AddUnitNodeToScene(..., const char* suffix, ...)'
  121 |                      const char* suffix, int precision,
      |                                ~~~~~~~~~^~~~~~

--
```

```
$ g++ --version
x86_64-alt-linux-g++ (GCC) 14.2.1 20241028 (ALT Sisyphus 14.2.1-alt1)
```

### DIFF
--- a/Libs/MRML/Core/Testing/vtkMRMLTextStorageNodeTest1.cxx
+++ b/Libs/MRML/Core/Testing/vtkMRMLTextStorageNodeTest1.cxx
@@ -45,7 +45,7 @@ int vtkMRMLTextStorageNodeTest1(int argc, char* argv[])
   scene->SetRootDirectory(tempDir);
 
   CHECK_EXIT_SUCCESS(TestReadWriteData(scene.GetPointer(), ".txt", "Hello world!", VTK_ENCODING_US_ASCII));
-  CHECK_EXIT_SUCCESS(TestReadWriteData(scene.GetPointer(), "UTF8.txt", u8"Hell\u00F3 vil\u00E1g!", VTK_ENCODING_UTF_8));
+  CHECK_EXIT_SUCCESS(TestReadWriteData(scene.GetPointer(), "UTF8.txt", static_cast<const char*>(u8"Hell\u00F3 vil\u00E1g!"), VTK_ENCODING_UTF_8));
   CHECK_EXIT_SUCCESS(TestReadWriteData(scene.GetPointer(), ".xml", "<Hello World=True/>", VTK_ENCODING_US_ASCII));
   CHECK_EXIT_SUCCESS(TestReadWriteData(scene.GetPointer(), ".json", "{\"Hello\":\"World\"}", VTK_ENCODING_US_ASCII));
 
--- a/Libs/MRML/Widgets/Testing/qMRMLUtf8Test1.cxx
+++ b/Libs/MRML/Widgets/Testing/qMRMLUtf8Test1.cxx
@@ -100,8 +100,8 @@ int qMRMLUtf8Test1(int argc, char* argv[])
   QString actualDisplayNodeName = QString::fromUtf8(volumeDisplayNode->GetName());
   // expectedDisplayNodeName contains a number of unicode characters that are not found in Latin1 character set
   // (the word is "a'rvi'ztu"ro" tu:ko:rfu'ro'ge'p" - https://en.wikipedia.org/wiki/Mojibake#Examples)
-  QString expectedDisplayNodeName =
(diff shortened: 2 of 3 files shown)


pieper:
Thanks for the contribution 👍 

I have zero experience with this, but will this change be backwards compatible with older compilers?  I.e. should there be an `#fdef` for some C++ feature?

kovalev0:
> Thanks for the contribution 👍
> 
> I have zero experience with this, but will this change be backwards compatible with older compilers? I.e. should there be an `#fdef` for some C++ feature?

The root cause is that starting in C++20 the u8"..." literal yields a const char8_t* (a new UTF-8 character type introduced by the standard), whereas prior to C++20 it produced a const char*. Our AddUnitNodeToScene(..., const char* suffix, ...) signature expects a const char*, so the conversion from char8_t* fails.

By removing the u8 prefix and using "\u00B5m", we rely on the universal character name \u00B5 in a narrow string literal, which has been fully supported since C++11 (see https://en.cppreference.com/w/cpp/language/string_literal). All compilers that Slicer supports understand "\u00B5m" and generate the correct UTF-8 bytes (0xC2 0xB5).

Therefore, no #ifdef guard is needed, as this change remains backwards‑compatible with all C++11+ compilers used in Slicer’s ecosystem.

pieper:
Thanks for the explanation 👍 

Does anyone else want to have a look?

lassoan:
I've tested this on Windows, VC++ compiler. The changes actually cause incorrect characters to appear.

image

The problem is that `u8` has an important function. It tells that the string is not specified using the compiler's execution character set but using UTF8.

u8"\u00b5s"
Encoding: Always UTF-8 (by definition)
Type: const char* in C++17
Bytes: 0xC2 0xB5 0x73 (i.e., "µs" encoded as UTF-8)

"\u00b5s"
Encoding: Depends on the compiler's execution character set
On Windows with MSVC, this is often Windows-1252 or Latin-1, not UTF-8
Bytes: 0xB5 0x73 (i.e., "µs" in Windows-1252)

Example:

image

image

s1 is incorrect, not utf8 (it is actually latin1). s2 is correct (utf8).

Unfortunately, it seems that this strict check in Cxx20 makes implementation of "UTF8 everywhere" a bit more complicated.

A simple way would be to just cast this away by doing something like `static_cast<const char*>(u8"\u00b5m")`, but maybe there are more elegant solutions.

kovalev0:
Thanks for the feedback and testing, I’ve updated the patch to use static_cast<const char*> on all u8"\u00B5…" literals.

lassoan:
Thank you for the update, this should work. For consistency, please also update the two tests that uses `u8"`.

kovalev0:
Fixed vtkMRMLTextStorageNodeTest1.cxx and qMRMLUtf8Test1.cxx tests

lassoan (dismissed):

jcfr:
Out of curiosity, does anyone know why the following is reported:

image

jamesobutler:
https://github.blog/changelog/2025-05-01-github-now-provides-a-warning-about-hidden-unicode-text/

> To review a file for which this warning is displayed, open it in an editor that will display the hidden Unicode characters, like [Visual Studio Code](https://code.visualstudio.com/) which highlights the characters by default. Then, verify that the characters are necessary and not disguising text that will be interpreted or compiled differently than it appears.

Seems like need to open in an editor to check.

jcfr:
I fixed the linting error :rocket: 


----

As a quick note for future reference: let's stick to regular hyphens (`-`) in branch names and avoid using the **non-breaking hyphen**[^1], which can be visually confusing and cause issues in tooling.

[^1]: https://www.compart.com/en/unicode/U+2011

PR #8622, opened 2025-08-07
ENH: Remove unnecessary threshold preview pipeline in unused views

jamesobutler:
Rendering the Segment Editor effect threshold preview in slice views that are no longer visible or have no volume in the foreground or background layer is unnecessary. A warning had been spamming the log about this situation, but this commit replaces it by handling creation and deletion of preview pipelines as the layout manager state changes.

To replicate the spammy log message I was observing, copy the following into a new session of Slicer:
```python
slicer.mrmlScene.Clear()
slicer.app.layoutManager().setLayout(slicer.vtkMRMLLayoutNode.SlicerLayoutThreeOverThreeView)
slicer.app.processEvents()
import SampleData
volumeNode = SampleData.SampleDataLogic().downloadMRHead()
slicer.util.selectModule("SegmentEditor")
slicer.modules.SegmentEditorWidget.editor.findChild(qt.QPushButton, "AddSegmentButton").click()

slicer.app.processEvents()
for slice_view in slicer.app.layoutManager().sliceViewNames():
    sliceWidget = slicer.app.layoutManager().sliceWidget(slice_view)
    sliceLogic = sliceWidget.sliceLogic()
    if slice_view not in ['Red', 'Green', 'Yellow']:
        sliceLogic.GetSliceCompositeNode().SetBackgroundVolumeID(None)
        sliceLogic.GetSliceCompositeNode().SetForegroundVolumeID(None)

slicer.app.layoutManager().setLayout(slicer.vtkMRMLLayoutNode.SlicerLayoutFourUpView)
slicer.modules.SegmentEditorWidget.editor.setActiveEffectByName("Threshold")
```

This was a topic I brought up during a Slicer hangout many months ago and I forgot to follow up with this change.

### DIFF
--- a/Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorEffects/SegmentEditorThresholdEffect.py
+++ b/Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorEffects/SegmentEditorThresholdEffect.py
@@ -42,6 +42,7 @@ class SegmentEditorThresholdEffect(AbstractScriptedSegmentEditorEffect):
 
         self.previewPipelines = {}
         self.histogramPipeline = None
+        self._sliceCompositeObservers = []
 
         # Histogram stencil setup
         self.stencil = vtk.vtkPolyDataToImageStencil()
@@ -101,6 +102,9 @@ class SegmentEditorThresholdEffect(AbstractScriptedSegmentEditorEffect):
         # Update intensity range
         self.sourceVolumeNodeChanged()
 
+        # Set up observers for background/foreground volume changes in visible slice widgets
+        self.addSliceCompositeObservers()
+
         # Start preview pulse
         self.timer.start(200)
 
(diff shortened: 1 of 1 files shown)


lassoan:
Thanks for the PR. It looks good overall, but I will have a closer look.

Is there any difference in support of Segment Editor in views that are not managed by the layout manager? (I don't remember if it was supported before, probably not, just please check to make sure)

jamesobutler:
I never considered the Segment Editor operating on a slice view that was not part of the layout manager. Based on the following code below it appears preview pipelines are not setup when the layout manager is not being used.
https://github.com/Slicer/Slicer/blob/5e290f9562926124e957500fa5b9da8f727064f8/Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorEffects/SegmentEditorThresholdEffect.py#L614-L618

lassoan:
It looks good, but I'm wondering if we really need to observe composite node changes or we cold do it simpler. Could we just not display the overlay in a slice view (and not log any warnings) when the source volume is not visible there?

jamesobutler:
Yes though I was handling situations where the preview is initially created and there is a blank slice viewer (only in red, but empty in yellow and green) and then you select a volume to view in that empty viewer (select volume in yellow and green). The composite node updates helped to know when to consider refreshing the preview pipelines. I didn’t want to have to force stopping the preview and then restarting it to generate based on the current slice view population.

lassoan (approved):
While the implementation could be simpler, what is implemented is a bit more efficient (which might be useful if there are many slice views and only a few of them show the source volume).

PR #8630, opened 2025-08-07
BUG: Fix empty volume display histogram group box

jamesobutler:
It appears that if initializing a qSlicerScalarVolumeDisplayWidget during startup initialization, it can result in the ctkCollapsibleGroupBox children breaking their visibility from being adjusted. This fix is inspired by similar code at https://github.com/commontk/CTK/blob/fb91447e0e15327341619cc0cda744724408ebc6/Libs/Widgets/ctkCollapsibleGroupBox.cpp#L281-L285.

With my Slicer custom app:

| Before | After (changes in this PR)|
|---------|-------|
|{1B15D9EC-A112-4328-BFD4-32D80E1D1199}|{9B0C0FD6-0350-41CA-9D09-6D4D4C9437F9}|


Modules/Loadable/Volumes/Widgets/qSlicerScalarVolumeDisplayWidget.cxx
```diff
@@ -241,7 +241,10 @@ void qSlicerScalarVolumeDisplayWidget::updateHistogram()
   vtkDataArray* voxelValues = pointData ? pointData->GetScalars() : nullptr;
 
   // If there are no voxel values then we completely hide the histogram section
-  d->HistogramGroupBox->setVisible(voxelValues != nullptr);
+  if (d->HistogramGroupBox->testAttribute(Qt::WA_WState_Created))
+  {
+    d->HistogramGroupBox->setVisible(voxelValues != nullptr);
+  }
```
jcfr, line 254: @lassoan  Should we remove the second call to `d->Histogram->setDataArray(voxelValues);`  (line 263) introduced in c1ec49d09ffe1d8733bc00272206d3c661dbb03c ?

Modules/Loadable/Volumes/Widgets/qSlicerScalarVolumeDisplayWidget.cxx
```diff
@@ -241,7 +241,10 @@ void qSlicerScalarVolumeDisplayWidget::updateHistogram()
   vtkDataArray* voxelValues = pointData ? pointData->GetScalars() : nullptr;
 
   // If there are no voxel values then we completely hide the histogram section
```
jcfr:
```suggestion
  // Guard against calling setVisible() too early.
  //
  // ctkCollapsibleGroupBox ignores child visibility changes until it has
  // been created (WA_WState_Created is true). Calling setVisible() before
  // this point can cause its internal visibility bookkeeping to become
  // inconsistent, leaving children in the wrong state.
  //
  // See ctkCollapsibleGroupBoxPrivate::setChildVisibility() for details.
```

jcfr:
After adding the comment, this is good for integration :rocket: 

Thanks again for taking the time to address this :100:

Modules/Loadable/Volumes/Widgets/qSlicerScalarVolumeDisplayWidget.cxx
jamesobutler: I've pushed the update with this comment

jcfr (approved):

PR #8643, opened 2025-08-11
ENH: Set Windows registry key to use high performance graphics

jamesobutler:
Beginning with Windows 10 20H1 the Windows OS assigns the graphics processor to use for various application. It will override any settings made in third-party applications such as the NVIDIA Control Panel. If the OS does not have a GPU preference for an application, you can use a third-party app like NVIDIA Control Panel to set preferred graphics processor.

By default Windows has a "Let Windows Decide" option, but it has been observed that it always seems to prefer the power saving integrated graphics option even though there is a high performing graphics available and even when the device is plugged into power and not running on battery. Setting the default to use the "High-Performance Graphics" option avoids issues with new 3D Slicer users who want to do some volume rendering and then face suboptimal performance even when using "VTK GPU Ray Casting" rendering setting is defined because it is actually using the integrated graphics.

I used Windows Task Manager to observe graphics performance where GPU 0 is the integrated graphics and GPU 1 is an NVIDIA T1200 graphics card. This is a typical setup for a mobile workstation where there are 2 graphics options.
image

When using recent Slicer Preview and doing some volume rendering I observed that by default the application was using the integrated graphics - GPU 0.
image

With this PR, the setting of the registry key is observed through Windows Settings->System->Display->Graphics. There is now a predefined entry for the install application that has the "High performance" graphics option selected by default. The "SlicerApp-real.exe" has to be picked instead of the Slicer.exe launcher because otherwise the graphics preference won't work as expected.
image
Now when volume rendering I observe that by default the application is using the NVIDIA T1200 graphics - GPU 1 (it could be any discrete graphics card)
image

Note that when a system only has integrated graphics the "Power Saving" and "High Performance" options are mapped to the same entry so there will be no impact of the registry key.
integrated-graphics-only

### DIFF
--- a/CMake/SlicerCPack.cmake
+++ b/CMake/SlicerCPack.cmake
@@ -503,14 +503,25 @@ if(CPACK_GENERATOR STREQUAL "NSIS")
   set(CPACK_NSIS_FINISH_TITLE_3LINES True)
 
   # -------------------------------------------------------------------------
+  # Graphics Preference
+  # -------------------------------------------------------------------------
+  # Windows Settings -> System -> Display -> Graphics
+  # This area has options for setting custom graphics settings for an individual app with the following options:
+  # GpuPreference=0 -> "Let Windows Decide (default)"
+  # GpuPreference=1 -> "Power Saving" - uses integrated graphics
+  # GpuPreference=2 -> "High Performance" - uses discrete graphics
+  set(CPACK_NSIS_EXTRA_INSTALL_COMMANDS
+"WriteRegStr SHCTX \\\"SOFTWARE\\\\Microsoft\\\\DirectX\\\\UserGpuPreferences\\\" \\\"$INSTDIR\\\\bin\\\\${APPLICATION_NAME}App-real.exe\\\" \\\"GpuPreference=2;\\\"
+")
+  set(CPACK_NSIS_EXTRA_UNINSTALL_COMMANDS
+"DeleteRegValue SHCTX \\\"SOFTWARE\\\\Microsoft\\\\DirectX\\\\UserGpuPreferences\\\" \\\"$INSTDIR\\\\bin\\\\${APPLICATION_NAME}App-real.exe\\\"
+")
+
+  # -------------------------------------------------------------------------
   # File extensions
   # -------------------------------------------------------------------------
   set(FILE_EXTENSIONS .mrml .xcat .mrb)
-
   if(FILE_EXTENSIONS)
-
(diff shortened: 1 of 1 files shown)


CMake/SlicerCPack.cmake
```diff
@@ -502,15 +502,21 @@ if(CPACK_GENERATOR STREQUAL "NSIS")
   set(CPACK_NSIS_FINISH_TITLE "Completed the ${PACKAGE_APPLICATION_NAME} Setup Wizard")
   set(CPACK_NSIS_FINISH_TITLE_3LINES True)
 
+  # -------------------------------------------------------------------------
+  # Graphics Preference
+  # -------------------------------------------------------------------------
+  set(CPACK_NSIS_EXTRA_INSTALL_COMMANDS
+"WriteRegStr SHCTX \\\"SOFTWARE\\\\Microsoft\\\\DirectX\\\\UserGpuPreferences\\\" \\\"$INSTDIR\\\\bin\\\\${APPLICATION_NAME}App-real.exe\\\" \\\"GpuPreference=2;\\\"
```
jcfr, line 514: Is GpuPreference set to 2 always the GPU ? I am wondering if we could include a link to the relevant documentation of any.

jcfr (commented):
Thanks for the thorough analysis and detailed explanation. This will be a great improvement.

CMake/SlicerCPack.cmake
jamesobutler: I have added more code comments describing the available options and that "2" specifically maps to the "High performance" selection. "1" maps to the "Power saving" selection which is generally the integrated graphics and "0" is the default "Let Windows decide" where Windows has been choosing the integrated graphics for Slicer.

Unfortunately I have not found an official Microsoft documentation page describing this graphics preference setting. There are many other third party websites, but none worth memorializing as a link in the code.

jcfr (approved):

jcfr:
The additional details will work well. Thanks for taking the time 🙏👌

CMake/SlicerCPack.cmake
jamesobutler: In some cases as mentioned in the original post, systems with only integrated graphics will have the same option displayed for the "Power saving" and "High performance" option. So "2" means "best - highest performing option" rather than always being "discrete GPU"

image

jcfr: That is also a reasonable pre-selected choice

PR #8645, opened 2025-08-13
BUG: Fix Windows graphics preference not set when installing app with admin privileges

jamesobutler:
This is a follow-up to https://github.com/Slicer/Slicer/commit/fd55fbc130a45d5f85e0e04b294d2a8169bc811d.

SHCTX or SHELL_CONTEXT is set to HKLM if the shell context is for all and is set to HKCU if shell context is set to current user. Slicer is by default set to install with regular user privileges where SHCTX was working by representing HKCU, but then it would not work when installing Slicer with elevated privileges to a system location like "C:/Program Files". This commit fixes the issue by always writing to the HKCU location as this graphics preference setting is a user specific setting.

### DIFF
--- a/CMake/SlicerCPack.cmake
+++ b/CMake/SlicerCPack.cmake
@@ -510,11 +510,12 @@ if(CPACK_GENERATOR STREQUAL "NSIS")
   # GpuPreference=0 -> "Let Windows Decide (default)"
   # GpuPreference=1 -> "Power Saving" - uses integrated graphics
   # GpuPreference=2 -> "High Performance" - uses discrete graphics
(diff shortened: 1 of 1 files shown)


jcfr (approved):

jcfr:
Should the value `SHCTX` vs `HKCU` be conditionally set based on `Slicer_CPACK_NSIS_INSTALL_REQUIRES_ADMIN_ACCOUNT` ?

jamesobutler:
No, it appears the Windows app graphics preference setting is strictly a user setting and not applied at a system level. Therefore HKCU - the current user method seems to be the only viable option.

PR #8649, opened 2025-08-16
BUG: Drop unneeded nsl and nis dependencies to resolve libtirpc symbol error

jcfr:
This pull request removes unnecessary dependencies from Slicer’s build system that could introduce runtime conflicts with system libraries on modern Linux platforms.

### DCMTK

* Update DCMTK to avoid linking against `libnsl` on modern platforms.
* On Linux, BSD, macOS, and Solaris >= 11.4, networking APIs (`gethostbyname`, `getaddrinfo`, `socket`, ...) are provided directly by `libc`; only legacy Solaris required `-lnsl`/`-lsocket`.
* Removing the redundant `libnsl` linkage prevents transitive dependencies (e.g., `libtirpc` and a system OpenSSL version incompatible with Slicer’s bundled one).
* Fixes an observed **symbol resolution error** involving `libtirpc.so.3` and Slicer’s stubbed GSSAPI library.

### Python `nis` Module

* Disable the deprecated `nis` module when building Python on UNIX.
* The module has been deprecated since Python 3.11 and removed entirely in 3.13.
* Disabling it in Python 3.12 simplifies Slicer packaging by avoiding dependencies on `libtirpc` and `libnsl`.

### DIFF
--- a/SuperBuild/External_python.cmake
+++ b/SuperBuild/External_python.cmake
@@ -51,6 +51,8 @@ if((NOT DEFINED PYTHON_INCLUDE_DIR
 
   set(python_SOURCE_DIR "${CMAKE_BINARY_DIR}/Python-${Slicer_REQUIRED_PYTHON_VERSION}")
 
+  # Python version update notes:
+  # - When updating to Python >= 3.13, remove explicit setting of ENABLE_NIS to OFF below.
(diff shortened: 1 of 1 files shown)


jcfr:
Related upstream DCMTK pull request:
* https://github.com/DCMTK/dcmtk/pull/125

Punzo:
tested on ubuntu 25.04, everything looks good! thanks!

PR #8650, opened 2025-08-17
CMake: Replace deprecated TestBigEndian module

petk:
TestBigEndian module can be replaced with `CMAKE_<LANG>_BYTE_ORDER` variable, which is available as of CMake 3.20. In this case the CMAKE_CXX_BYTE_ORDER is used.


jcfr:
Thanks @petk for the contribution :pray: :tada: 

We tweaked the commit message and once the continuous integration completes, it should be good for integration.

---

Also, this prompted us to further review the use of the corresponding macros. This means that instead of (or in addition of) this pull request, we are also considering integrating the following pull request:
* https://github.com/Slicer/Slicer/pull/8653

petk:
Ok, thanks. Even better then.

PR #8656, opened 2025-08-19
DOC: Add dedicated GOVERNANCE document and reference it from the about page

jcfr:
This change extracts the decision-making process and the list of benevolent dictators for life (BDFLs) from `CONTRIBUTING.md` into a new `GOVERNANCE.md` document.

It also updates the about page to include a reference to the new governance document.


pieper (approved):
👍

PR #8658, opened 2025-08-20
ENH: Add API to retrieve view displayable managers without Qt widgets

jcfr:
This adds `vtkSlicerApplicationLogic::GetViewDisplayableManagerByClassName` to access a view’s displayable manager by class name without relying on Qt view classes. The layout factories now propagate each view’s `vtkMRMLDisplayableManagerGroup` into the corresponding MRML logic (`vtkMRMLSliceLogic` / `vtkMRMLViewLogic`), which keeps a weak reference.

Motivation: decouple effects and scripts from Qt to support headless and trame-backed frontends.

Key changes:
- Base/Logic links against `MRMLDisplayableManager` and includes headers.
- Add `DisplayableManagerGroup` setters/getters to Slice/View logics with type validation.
- Refactor Segment Editor and Subject Hierarchy code paths to use the new ApplicationLogic API.
- Update developer docs and script repository examples accordingly.

Notes:
- Returns nullptr if inputs are invalid or the manager is not present. The returned pointer is owned by the displayable manager group.
- Existing Qt-based retrieval continues to work.

### DIFF
--- a/Libs/MRML/Logic/vtkMRMLViewLogic.h
+++ b/Libs/MRML/Logic/vtkMRMLViewLogic.h
@@ -23,6 +23,10 @@
 // MRMLLogic includes
 #include "vtkMRMLAbstractLogic.h"
 
+// VTK includes
+#include <vtkObject.h>
+#include <vtkWeakPointer.h>
+
 // STD includes
 #include <vector>
 #include <deque>
@@ -97,6 +101,12 @@ public:
   /// The name of the Logic is the same of the widget one to which it is associated
   static vtkMRMLCameraNode* GetCameraNode(vtkMRMLScene* scene, const char* layoutName);
 
+  /// @{
+  /// \brief Get/Set the displayable manager group associated with the logic.
+  vtkGetObjectMacro(DisplayableManagerGroup, vtkObject);
+  void SetDisplayableManagerGroup(vtkObject* obj);
+  /// @}
+
 protected:
   vtkMRMLViewLogic();
   ~vtkMRMLViewLogic() override;
@@ -117,6 +127,22 @@ protected:
   vtkMRMLViewNode* ViewNode;
   vtkMRMLCameraNode* CameraNode;
   bool UpdatingMRMLNodes;
+
+  // Weak reference to the view's displayable manager group.
+  //
+  // Stored as vtkObject to keep MRMLLogic independent of
+  // MRMLDisplayableManager headers. The expected dynamic type is
+  // vtkMRMLDisplayableManagerGroup.
+  //
+  // Set by the layout factories when the view widget is created, and used by
+  // vtkSlicerApplicationLogic::GetViewDisplayableManagerByClassName() to
+  // resolve managers without Qt. This works because vtkSlicerApplicationLogic
+  // lives in SlicerBaseLogic, which can depend on MRMLDisplayableManager.
+  //
+  // Ownership: the group is owned by the view/widget; this logic holds only a
+  // weak pointer. May be nullptr (e.g., headless or before view creation).
+  // Type is validated in SetDisplayableManagerGroup().
+  vtkWeakPointer<vtkObject> DisplayableManagerGroup;
 
 private:
   vtkMRMLViewLogic(const vtkMRMLViewLogic&) = delete;
--- a/Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.cxx
+++ b/Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.cxx
@@ -2841,19 +2841,24 @@ void qMRMLSegmentEditorWidget::processEvents(vtkObject* caller, unsigned long ei
     return;
   }
 
+  vtkSlicerApplicationLogic* appLogic = qSlicerApplication::application()->applicationLogic();
+  if (!appLogic)
+  {
+    return;
+  }
+
   // Call processing function of active effect. Handle both interactor and view node events
   vtkRenderWindowInteractor* callerInteractor = vtkRenderWindowInteractor::SafeDownCast(caller);
   vtkMRMLAbstractViewNode* callerViewNode = vtkMRMLAbstractViewNode::SafeDownCast(caller);
-  if (callerInteractor)
+  if (callerInteractor != nullptr)
   {
-
     // Do not process events while a touch gesture is in progress (e.g., do not paint in the view
     // while doing multi-touch pinch/rotate).
     qMRMLSliceWidget* sliceWidget = qobject_cast<qMRMLSliceWidget*>(viewWidget);
     if (sliceWidget)
     {
       vtkMRMLCrosshairDisplayableManager* crosshairDisplayableManager =
-        vtkMRMLCrosshairDisplayableManager::SafeDownCast(sliceWidget->sliceView()->displayableManagerByClassName("vtkMRMLCrosshairDisplayableManager"));
+        vtkMRMLCrosshairDisplayableManager::SafeDownCast(appLogic->GetViewDisplayableManagerByClassName(sliceWidget->mrmlSliceNode(), "vtkMRMLCrosshairDisplayableManager"));
       if (crosshairDisplayableManager)
       {
         int widgetState = crosshairDisplayableManager->GetSliceIntersectionWidget()->GetWidgetState();
@@ -2863,11 +2868,12 @@ void qMRMLSegmentEditorWidget::processEvents(vtkObject* caller, unsigned long ei
         }
       }
     }
+
     qMRMLThreeDWidget* threeDWidget = qobject_cast<qMRMLThreeDWidget*>(viewWidget);
     if (threeDWidget)
     {
       vtkMRMLCameraDisplayableManager* cameraDisplayableManager =
-        vtkMRMLCameraDisplayableManager::SafeDownCast(threeDWidget->threeDView()->displayableManagerByClassName("vtkMRMLCameraDisplayableManager"));
+        vtkMRMLCameraDisplayableManager::SafeDownCast(appLogic->GetViewDisplayableManagerByClassName(threeDWidget->mrmlViewNode(), "vtkMRMLCameraDisplayableManager"));
       if (cameraDisplayableManager)
       {
         int widgetState = cameraDisplayableManager->GetCameraWidget()->GetWidgetState();
@@ -2886,7 +2892,7 @@ void qMRMLSegmentEditorWidget::processEvents(vtkObject* caller, unsigned long ei
       callbackCommand->SetAbortFlag(1);
     }
   }
-  else if (callerViewNode)
+  else if (callerViewNode != nullptr)
   {
     activeEffect->processViewNodeEvents(callerViewNode, eid, viewWidget);
   }
--- a/Base/Logic/CMakeLists.txt
+++ b/Base/Logic/CMakeLists.txt
@@ -58,6 +58,7 @@ set(include_dirs
   ${CMAKE_CURRENT_BINARY_DIR}
   ${MRMLCore_INCLUDE_DIRS}
   ${MRMLLogic_INCLUDE_DIRS}
+  ${MRMLDisplayableManager_INCLUDE_DIRS}
   ${vtkTeem_INCLUDE_DIRS}
   ${RemoteIO_INCLUDE_DIRS}
   ${LibArchive_INCLUDE_DIR}
@@ -138,6 +139,7 @@ add_library(${lib_name} ${srcs})
 
 set(libs
   MRMLLogic
+  MRMLDisplayableManager
   ${VTK_LIBRARIES}
   )
 
--- a/Base/Logic/vtkSlicerApplicationLogic.cxx
+++ b/Base/Logic/vtkSlicerApplicationLogic.cxx
@@ -16,12 +16,19 @@
 #include "vtkSlicerConfigure.h"
 #include "vtkSlicerTask.h"
 
+// MRMLDisplayableManager includes
+#include <vtkMRMLDisplayableManagerGroup.h>
+
 // MRML includes
 #include <vtkCacheManager.h>
 #include <vtkDataIOManagerLogic.h>
 #include <vtkMRMLRemoteIOLogic.h>
 #include <vtkMRMLScene.h>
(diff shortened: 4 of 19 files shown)


jcfr:
Suggested next steps:
* Extend `vtkMRMLLayoutLogic` API to add `GetViewNodesByClassName(const char* className)` to complement the existing `GetViewNodes()`
* Update `qSlicerSegmentEditorAbstractEffect::processInteractionEvents` to accept `vtkMRMLAbstractViewNode` instead of `qMRMLWidget*`, this should help identify additional refactoring to decouple editor effect from Qt.

Thibault-Pelletier:
Hi @jcfr,

Thanks for the follow up!
As discuss during the weekly meeting, the reason for this access to the displayable manager only steams from the [qMRMLSegmentEditorWidget.cxx#L2852](https://github.com/Slicer/Slicer/blob/8a96c0c2b94eb6088a9520c198e4325d9965fdcf/Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.cxx#L2852) widget hack(s).

From our discussions, I think that accessing the DM directly would be an anti-pattern we don't want to propagate in the application. 

Direct access to the DM for their underlying widget's is specially hacky as changing the layout would instantiate new widgets that need to be manually modified if needed.

In the modified examples, access to `widget = cameraDisplayableManager.GetCameraWidget()` (amongst others) to override the interaction events should instead be stored in its underlying MRML Node, in my opinion, and all widgets should follow the MRML Node state as for other node properties.

Factorizing the segment editor effects into proper DM should remove the need for the hacks present in the segment editor and effects classes.

@jcfr, @lassoan & @pieper happy to hear your opinions on this!

jcfr:
> Factorizing the segment editor effects into proper DM should remove the need for the hacks present in the segment editor and effects classes.

:rocket: :100: 

> Direct access to the DM for their underlying widget's is specially hacky as changing the layout would instantiate new widgets that need to be manually modified if needed.

Ditto. I agree that *direct* DisplayableManager (DM) access is not a pattern we want to normalize.

To move forward and ultimately remove the existing Segment Editor hacks (e.g. the crosshair/camera widget state checks in `qMRMLSegmentEditorWidget`), I suggest we treat the new API proposed in this PR as a **transitional bridge**. It removes **Qt** dependencies today and lets us iterate toward a MRML-driven and/or DM-centric design.

Also, having a Qt-independent way to reach a displayable manager is useful for **prototyping** and **debugging** in both the Qt and trame applications.

**Interim steps:**

1. **Layout logic helper**
   Add `vtkMRMLLayoutLogic::GetViewNodesByClassName(const char*)` to retrieve target views.

2. **Effect API decoupling**
   Update `qSlicerSegmentEditorAbstractEffect::processInteractionEvents` to accept `vtkMRMLAbstractViewNode*` (instead of `qMRMLWidget*`). This will allow to identify additional places to decouple effects from Qt.

3. **Rendering backend (non-Qt)**
   - Introduce `vtkMRMLAbstractRenderingBackend` and register per-view backends in `vtkSlicerApplicationLogic` (retrieved via `GetViewRenderingBackend(viewNode)`). This replaces direct `ctkVTKAbstractView` calls without changing behavior.
   - https://github.com/Slicer/Slicer/issues/8659

4. **Segment Editor helper DM**
   Move Segment Editor–specific view logic into a `vtkMRMLSegmentEditorEffectDisplayableManagerHelper` (similar to `vtkMRMLMarkupsDisplayableManagerHelper`).

5. **Rewrite _Scissor_ effect**
   Re-implement on top of closed-curve markups.

6. **Rewrite _Paint_ effect preview**
   Consider also using the new slice multi-layer API.

7. **Module-logic registration + DM**
   Add support for registering editor effects from `vtkSlicerSegmentationsModuleLogic` and finalize the `vtkMRMLSegmentEditorEffectDisplayableManager` class.

> In the modified examples [...]

Note that the updated examples span beyond the segment editor and are now updated to be Qt independent which is something we want to move toward.

Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.cxx
```diff
@@ -2886,7 +2890,7 @@ void qMRMLSegmentEditorWidget::processEvents(vtkObject* caller, unsigned long ei
       callbackCommand->SetAbortFlag(1);
     }
   }
-  else if (callerViewNode)
+  else if (vtkMRMLAbstractViewNode* callerViewNode = vtkMRMLAbstractViewNode::SafeDownCast(caller); callerViewNode != nullptr)
```
cpinter: Readable code is more important than conciseness. I think this is also a practice we should not propagate.

Libs/MRML/Logic/vtkMRMLViewLogic.h
```diff
@@ -118,6 +128,8 @@ class VTK_MRML_LOGIC_EXPORT vtkMRMLViewLogic : public vtkMRMLAbstractLogic
   vtkMRMLCameraNode* CameraNode;
   bool UpdatingMRMLNodes;
 
+  vtkWeakPointer<vtkObject> DisplayableManagerGroup;
```
cpinter, line 145: Why simple `vtkObject`?

cpinter: I see, because the MRMLLogic does not depend on MRMLDM. In this case, this seems also kind of a hack to me.

cpinter:
> Direct access to the DM for their underlying widget's is specially hacky as changing the layout would instantiate new widgets that need to be manually modified if needed.

What does this sentence mean exactly? It is not clear to me how the segment editor widget is a DM's "underlying widget". What difference does it make where you get the same DM from the same class?

> Rewrite Scissor effect: Re-implement on top of closed-curve markups.

This sounds like a major usability setback to me (please don't). What is the motivation?

Thibault-Pelletier:
Hi @cpinter,

> > Direct access to the DM for their underlying widget's is specially hacky as changing the layout would instantiate new widgets that need to be manually modified if needed.
> 
> What does this sentence mean exactly? It is not clear to me how the segment editor widget is a DM's "underlying widget". What difference does it make where you get the same DM from the same class?

My point was that no part of the code should be accessing any of the displayable managers directly. The current usage breaks the encapsulation and separation of concern principles.

Directly accessing the DM to modify their widgets leads to very brittle code.

A very simple (broken) example : 
* Create a 3D only view
* Access the DM to modify the camera interaction
* Switch to 2 or more 3D views > the camera interaction in these views don't match the modified interaction behavior 

> > Rewrite Scissor effect: Re-implement on top of closed-curve markups.
> 
> This sounds like a major usability setback to me (please don't). What is the motivation?

One motivation to rewriting the effects as DM is to have access to them for non Qt usage (i.e. trame Slicer). 
Edit : A second motivation is to have more consistent interaction handling between segmentation effects and other DM (current implementation intercepts the event and doesn't follow the DM priority rules)

@lassoan pointed out that rewriting the effects instead of trying to factorize the Qt / non qt part of the current implementation may be easier and that the new effects could leverage the current markup nodes for user interaction.

jcfr:
Displayable manager were indeed designed to be standalone and only here to update the view based on state captured in MRML nodes and alike.

We should definitely rewrite the editor effect to have a displayable manager.

That said, having the ability to access displayable manager will still be a reality, and doing so in a more systematic way without relying on Qt is sensible hence this pull request.


> Directly accessing the DM to modify their widgets leads to very brittle code.

Indeed, we all agree on this.

Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.cxx
jcfr: Motivation was to avoid using `callerViewNode` in the "wrong" code branch (which I originally did), this construct conveniently avoided this.

That said, I understand the readability argument and reverted back.

Libs/MRML/Logic/vtkMRMLViewLogic.h
jcfr: A comment has been added in both `vtkMRMLViewLogic` and `vtkMRMLSliceLogic`

```
  // Weak reference to the view's displayable manager group.
  //
  // Stored as vtkObject to keep MRMLLogic independent of
  // MRMLDisplayableManager headers. The expected dynamic type is
  // vtkMRMLDisplayableManagerGroup.
  //
  // Set by the layout factories when the view widget is created, and used by
  // vtkSlicerApplicationLogic::GetViewDisplayableManagerByClassName() to
  // resolve managers without Qt. This works because vtkSlicerApplicationLogic
  // lives in SlicerBaseLogic, which can depend on MRMLDisplayableManager.
  //
  // Ownership: the group is owned by the view/widget; this logic holds only a
  // weak pointer. May be nullptr (e.g., headless or before view creation).
  // Type is validated in SetDisplayableManagerGroup().
```

lassoan:
> Factorizing the segment editor effects into proper DM should remove the need for the hacks present in the segment editor and effects classes.

Yes, this is the most important point. If Segment Editor effects could implement all visualizations and interactions using MRML scene manipulations (e.g., displaying preview using segmentation nodes or labelmaps in Threshold effect, use markups for interactions in Scissors effect, etc.) then there would be no need for low-level VTK interaction observations and workarounds.

> We should definitely rewrite the editor effect to have a displayable manager.

We might be able to handle everything with existing DMs. For example, if @cpinter would strongly prefer to keep the single click-and-drag for drawing curve in Scissors, we could improve the curve widget to have a continuous drawing mode (where click-and-drag would drop control points automatically at small distances, similarly to how it is done in ITKSnap).

> override the interaction events should instead be stored in its underlying MRML Node

This is a separate topic. Interaction event translation definitions are designed to be storable as configuration data and therefore could be stored in the MRML scene. However, most of the interactions would be more of a user preference (e.g., some people want right-click to rotate the camera because they got used to that) and therefore it belongs to the application settings. In the Qt application we use the QSettings global object to store configuration data, but in a non-Qt application it could make sense to make the vtkApplicationLogic manage application-wide configuration data (the logic class could have GetValue/SetValue methods, or a separate vtkSettings object could be created).

-----

Overall, I see this PR is an intermediate step towards reducing Qt dependencies. If @Thibault-Pelletier finds it useful then I have no objections for adding these. If @Thibault-Pelletier would instead go directly to the final design (manipulate the scene and let DMs take care of display) then we may not need some of these proposed APIs.

Thibault-Pelletier:
Thanks for the update on this!

Accessing the DM may indeed be required as a transitionning step towards decoupling Qt.

Implementation wise, I think we could directly store the DM group map in the application logic to minimize the modifications.
Synchronizing the map can be done inside the qMRMLSliceView::setMRMLSliceNode and qMRMLThreeDView::setMRMLViewNode which would avoid adding friend classes definition. 

```c++
// 
    vtkMRMLAbstractDisplayableManager* GetViewDisplayableManagerByClassName(vtkMRMLAbstractViewNode* viewNode, const char* className) const;
    void SetViewDisplayableManagerGroup(vtkMRMLAbstractViewNode* viewNode, vtkMRMLDisplayableManagerGroup* );

private:
    std::map<vtkWeakPointer<vtkMRMLAbstractDisplayableManager>,  vtkWeakPointer<vtkMRMLDisplayableManagerGroup>> ViewDisplayableManagers;
    
// qMRMLSliceView::setMRMLSliceNode
  if (d->MRMLSliceNode)
  {
    appLogic->SetViewDisplayableManagerGroup(d->MRMLSliceNode, nullptr);
  }
  if(newSliceNode)
  {
    appLogic->SetViewDisplayableManagerGroup(newSliceNode, d->DisplayableManagerGroup);
  }
  
```

@jcfr let me know what you think of this and if I should amend the PR accordingly

jcfr:
> Implementation wise, I think we could directly store the DM group map in the application logic to minimize the modifications

That is a good idea :ok_hand: 

I initially thought about doing so but I wanted to avoid having to manage another (redundant) mapping.

That said, I can see how it would simply the implementation and would be a sensible trade-off.  

I would then appreciate your help amending this pull request :pray:

cpinter:
> That said, having the ability to access displayable manager will still be a reality, and doing so in a more systematic way without relying on Qt is sensible hence this pull request.

Thanks @jcfr , this answers my question. (Based on your explanations @Thibault-Pelletier the DM was not supposed to be accessed, but it still is, just via a different function)

> For example, if @cpinter would strongly prefer to keep the single click-and-drag for drawing curve in Scissors, we could improve the curve widget to have a continuous drawing mode (where click-and-drag would drop control points automatically at small distances, similarly to how it is done in ITKSnap).

I think many users like the current super-responsive and simple interaction in Scissors. Not sure if we could keep the current one (even in the extra effects extension) and add the new one instead of replacing it? Once the refactoring gets as far as making the older one not viable, I guess there will be no alternative but to removing it, but until then (without seeing how the new one works) I don't see why it could not be kept.

If everyone agrees with the PR I am also in agreement, just want to understand what is happening. Thanks for working on this.

Thibault-Pelletier:
Sorry, while trying to checkout the PR branch I merged the main branch into it...
I will force push in a bit to fix this 🤦

Thibault-Pelletier:
@jcfr I had a look at the branch and tried to implement my proposed API but I'm facing a circular dependency issue.

To register the DM group in the application logic from the qMRMLThreeDView (and slice view) would require the Application. But the qSlicerBaseQTGUI depends on the qMRMLWidgets library...

From what I can see there is no setter on the qMRMLWidget for the application logic (only the mrmlScene). Any qMRML widget would access the application logic or layout manager using the qApplication singleton.

Happy to have your opinion on this!

jcfr:
@Thibault-Pelletier 

I didn’t have a chance to fully re-assess the approach based on your recent proposal (https://github.com/Slicer/Slicer/pull/8658#issuecomment-3219857081) and analysis (https://github.com/Slicer/Slicer/pull/8658#issuecomment-3223742739), but I suggest we go ahead and merge the current implementation so we can ship the intended public API:

```cpp
vtkMRMLAbstractDisplayableManager* 
vtkSlicerApplicationLogic::GetViewDisplayableManagerByClassName(vtkMRMLAbstractViewNode* viewNode, const char* className) const
```

Thibault-Pelletier:
Sounds good!
Since the aim is a transitional change, the implementation can be refined if needs be later on.

Thibault-Pelletier (approved):

PR #8668, opened 2025-08-22
ENH: Match color label title font size to legend font size

Sunderlandkyl:
Previously, the title actor font size was not updated, even as the legend font size changed with viewport size. This commit applies the same font size to both the title and legend actors.

See: https://github.com/Slicer/Slicer/issues/6111#issuecomment-1500592494

Previously:
image

Now:
image


jamesobutler:
Thanks! Yes I have tested and it appears to work well. The main goal was to have the title text and label text use the same sizing convention and relative sizing to the viewport is a viable option. We will see how our users respond to the change and if they request further changes such as making it like the slice view horizontal ruler label which has a fixed screen size that is not relative to the viewport size.

PR #8669, opened 2025-08-25
ENH: Restore OpenGL core profile usage on Windows platform

### LINKED ISSUE #8647: Plot markers not displaying correctly when Slicer uses high performance graphics
## Summary

If I set Windows to run the Slicer application with the "High-Performance" graphics options which uses my T1200 GPU in my mobile workstation instead of the intel integrated graphics I observe that Plot markers are not rendered appropriately.

If this is a VTK issue, does anyone know how to provide a minimal example to the VTK developers related to the rendering option selected (integrated vs non-integrated graphics)? Or is this a NVIDIA driver issue (I'm using 573.06 with a NVIDIA T1200 Laptop GPU)?

## Steps to reproduce

[2025-08-14-Scene.zip](https://github.com/user-attachments/files/21778558/2025-08-14-Scene.zip)

1. Go to Windows Settings -> System -> Display -> Graphics
2. Add the Slicer app and select the option "Power Saving" to select your integrated graphics
3. Launch the Slicer app and load the mrb from the attached zip.
4. Observe that the plot markers are rendered appropriately.
5. Go to Windows Settings -> System -> Display -> Graphics
6. For the Slicer app, select the option "High Performance" to select your non-integrated graphics (NVIDIA card)
7. Re-Launch the Slicer app and load the mrb from the attached zip.
8. Observe that the plot markers are NOT rendered appropriately.

Marker Style "Cross" and "Square" appear as "Circle".
Marker Style "Plus", "Circle" and "Diamond" all appear like "None".

Image

The below results seem to show that it is not a sudden issue with the newer VTK version 9.2->9.5 between these Slicer versions. I guess could be a longer standing VTK issue.

| Slicer Version| "Power Saving" | "High Performance" |
|-|----------------------|-----------------------|
|Slicer 5.8.1 | Image|Image|
|Slicer-5.9.0-2025-08-13 | Image|Image|

## Environment
- Slicer version: Slicer-5.8.1 and Slicer-5.9.0-2025-08-13 (with "High Performance" windows graphics selected).
- Operating system: Windows

jamesobutler:
This closes #8647.

This is a follow-up to https://github.com/Slicer/Slicer/commit/47e89f94da6c8c8faa55de1ddcb50b5608d92a2f which removes the workaround that was setting the OpenGL compatibility profile as default on the Windows platform. The original rendering issues on Intel integrated graphics while use the OpenGL core profile appear to happen less often. Removing the workaround is motivated by rendering issues observed when Nvidia GPUs using the compatibility profile faced rendering issues related to Plot Markers which were not issues upon switching to the core profile.

See https://github.com/Slicer/Slicer/issues/8647#issuecomment-3203747715 for more details.

### DIFF
--- a/Libs/MRML/Widgets/qMRMLWidget.cxx
+++ b/Libs/MRML/Widgets/qMRMLWidget.cxx
@@ -91,23 +91,8 @@ void qMRMLWidget::preInitializeApplication()
 
   QString openGLProfileStr = qgetenv(MRML_APPLICATION_OPENGL_PROFILE_ENV);
   openGLProfileStr = openGLProfileStr.toLower();
-  if (openGLProfileStr.isEmpty() || openGLProfileStr == "default")
+  if (!openGLProfileStr.isEmpty() && openGLProfileStr != "default")
   {
-    // Use default profile
-#ifdef _WIN32
-    // Enable OpenGL compatibility profile on Windows.
-    // It fixes display update issues and should not have any
-    // side effect.
-    // Compatibility profile is only requested for Windows, as it is
-    // not fully supported on Mac, and there is no known issue
-    // on Linux that would require requesting compatibility profile.
-    // More details: https://gitlab.kitware.com/vtk/vtk/issues/17572
-    format.setProfile(QSurfaceFormat::CompatibilityProfile);
-#endif
-  }
-  else
-  {
-    // Force a specific profile
     if (openGLProfileStr == "no")
     {
       format.setProfile(QSurfaceFormat::NoProfile);
@@ -119,6 +104,12 @@ void qMRMLWidget::preInitializeApplication()
     else if (openGLProfileStr == "compatibility")
     {
       format.setProfile(QSurfaceFormat::CompatibilityProfile);
+    }
+    else
+    {
+      qCritical("Invalid OpenGL profile option '%s' set in %s. Valid options are: 'default', 'no', 'core', 'compatibility'.",
+                qPrintable(openGLProfileStr),
+                MRML_APPLICATION_OPENGL_PROFILE_ENV);
     }
   }
 


Libs/MRML/Widgets/qMRMLWidget.cxx
```diff
@@ -94,16 +94,6 @@ void qMRMLWidget::preInitializeApplication()
   if (openGLProfileStr.isEmpty() || openGLProfileStr == "default")
```
Thibault-Pelletier: Since you have removed the exception for win32, and the format is initialized to default, wouldn't it make sense to remove this if block?

lassoan: True. Or we could add a final "else" branch and log an error there, telling that an invalid profile option was set. For example, it would warn the user when he set "none" instead of the valid "no" option.

jamesobutler: I have rearranged the statement to not have an empty if statement area, but directly use a `if not` type logic to go into the block for the explicit handling. In that block I have added a final `else` statement to handle invalid explicit opengl profile string.

Libs/MRML/Widgets/qMRMLWidget.cxx
```diff
@@ -91,35 +91,18 @@ void qMRMLWidget::preInitializeApplication()
...
       format.setProfile(QSurfaceFormat::CoreProfile);
-    }
     else if (openGLProfileStr == "compatibility")
-    {
       format.setProfile(QSurfaceFormat::CompatibilityProfile);
-    }
+    else
+      qCritical("Invalid OpenGL profile option '%s' set in %s. Valid options are: 'default', 'no', 'core', 'compatibility'.",
+                qPrintable(openGLProfileStr),
+                MRML_APPLICATION_OPENGL_PROFILE_ENV);
```
jcfr:
```suggestion
    if (openGLProfileStr == "no")
    {
      format.setProfile(QSurfaceFormat::NoProfile);
    }
    else if (openGLProfileStr == "core")
    {
      format.setProfile(QSurfaceFormat::CoreProfile);
    }
    else if (openGLProfileStr == "compatibility")
    {
      format.setProfile(QSurfaceFormat::CompatibilityProfile);
    }
    else
    {
      qCritical("Invalid OpenGL profile option '%s' set in %s. Valid options are: 'default', 'no', 'core', 'compatibility'.",
                qPrintable(openGLProfileStr),
                MRML_APPLICATION_OPENGL_PROFILE_ENV);
    }
```

jamesobutler: @jcfr Is this something enforceable by the clang-format rules?

lassoan (approved):

Libs/MRML/Widgets/qMRMLWidget.cxx
jcfr: Indeed, here it is:
* https://github.com/Slicer/Slicer/pull/8671

jcfr (approved):

PR #8677, opened 2025-08-25
BUG: Fix VTK errors drawing line profile plot with undefined points

nhjohnston:
Change suggested by @jamesobutler 

This change allows for the line plot to be dynamically updated even before the user places the second point, and greatly reduces VTK warnings logged in the python console.

Main:

https://github.com/user-attachments/assets/7de323b3-2362-41e2-a13f-bd9540a35cdd

This PR:

https://github.com/user-attachments/assets/18c98d3a-848b-42a7-9dbe-96df5b48b93a

### DIFF
--- a/Modules/Scripted/LineProfile/LineProfile.py
+++ b/Modules/Scripted/LineProfile/LineProfile.py
@@ -249,7 +249,7 @@ class LineProfileLogic(ScriptedLoadableModuleLogic):
         """Called when the logic class is instantiated. Can be used for initializing member variables."""
         ScriptedLoadableModuleLogic.__init__(self)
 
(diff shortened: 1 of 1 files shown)


jamesobutler (approved):
Thanks @nhjohnston! I have pushed some minor additional updates to handle the VTK errors when you still had 1 point as shown in your video for this PR beginning at 0:08s. These changes should appropriate clear the table and plot if there is only 1 point present. The enhancement commit here allows for line profile generation when the 2nd point is present regardless if it is defined or undefined yet.

nhjohnston:
@jamesobutler the changes you added look great to me! There are no more logged errors when creating a new line

PR #8681, opened 2025-08-28
ENH: Add keyboard and mouse shortcuts for Module Selector navigation

ruffsl:
> [!Important]
> Testing this pull-request requires the following:
> 
> - A multi or five button mouse with forward and backward navigation buttons

This pull request adds global navigation features to the module selector toolbar, mimicking browser navigation behavior:

- **Global mouse navigation:** Users can now use mouse back/forward buttons anywhere in the application window to switch between previous and next modules.
- **Global keyboard shortcuts:** Added support for `Alt+Left Arrow` and `Alt+Right Arrow` as global shortcuts for previous and next module navigation, matching browser conventions.

These enhancements improve usability, while remaining intuitive for users familiar with file and web browser navigation.

### DIFF
--- a/Base/QTApp/Resources/UI/qSlicerMainWindow.ui
+++ b/Base/QTApp/Resources/UI/qSlicerMainWindow.ui
@@ -519,6 +519,39 @@
     <string>Open settings to the Modules panel where Favorite Modules can be defined.</string>
    </property>
   </action>
+  <action name="ModuleFindAction">
+   <property name="text">
+    <string>Module Finder</string>
+   </property>
+   <property name="toolTip">
+    <string>Find module by name</string>
+   </property>
+   <property name="shortcut">
+    <string>Ctrl+F</string>
+   </property>
+  </action>
+  <action name="ModulePreviousAction">
+   <property name="text">
+    <string>Previous module</string>
+   </property>
+   <property name="toolTip">
+    <string>Switch to previously used module</string>
+   </property>
+   <property name="shortcut">
+    <string>Alt+Left</string>
+   </property>
+  </action>
+  <action name="ModuleNextAction">
+   <property name="text">
+    <string>Next module</string>
+   </property>
+   <property name="toolTip">
+    <string>Switch to next used module</string>
+   </property>
+   <property name="shortcut">
+    <string>Alt+Right</string>
+   </property>
+  </action>
   <action name="EditApplicationSettingsAction">
    <property name="icon">
     <iconset>
--- a/Base/QTApp/Resources/UI/qSlicerMainWindow.ui
+++ b/Base/QTApp/Resources/UI/qSlicerMainWindow.ui
@@ -519,39 +519,6 @@
     <string>Open settings to the Modules panel where Favorite Modules can be defined.</string>
    </property>
   </action>
-  <action name="ModuleFindAction">
-   <property name="text">
-    <string>Module Finder</string>
-   </property>
-   <property name="toolTip">
-    <string>Find module by name</string>
-   </property>
-   <property name="shortcut">
-    <string>Ctrl+F</string>
-   </property>
-  </action>
-  <action name="ModulePreviousAction">
-   <property name="text">
-    <string>Previous module</string>
-   </property>
-   <property name="toolTip">
-    <string>Switch to previously used module</string>
-   </property>
-   <property name="shortcut">
-    <string>Alt+Left</string>
-   </property>
-  </action>
-  <action name="ModuleNextAction">
-   <property name="text">
-    <string>Next module</string>
-   </property>
-   <property name="toolTip">
-    <string>Switch to next used module</string>
-   </property>
-   <property name="shortcut">
-    <string>Alt+Right</string>
-   </property>
-  </action>
   <action name="EditApplicationSettingsAction">
    <property name="icon">
     <iconset>
--- a/Base/QTGUI/qSlicerModuleSelectorToolBar.cxx
+++ b/Base/QTGUI/qSlicerModuleSelectorToolBar.cxx
@@ -21,6 +21,7 @@
 // Qt includes
 #include <QCompleter>
 #include <QDebug>
+#include <QKeyEvent>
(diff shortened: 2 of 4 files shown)


Base/QTApp/Resources/UI/qSlicerMainWindow.ui
```diff
@@ -519,6 +519,54 @@
...
+   <property name="text">
+    <string>Next</string>
+   </property>
+   <property name="toolTip">
+    <string>Next modules</string>
+   </property>
+   <property name="shortcut">
+    <string>Alt+Right</string>
+   </property>
+  </action>
```
ruffsl: I notice shortcuts were only listed under the keyboard shortcut reference menu, so I added them here as well for user discoverability. Not sure if this is appropriate if we switch to using the action stubs instead.

Base/QTGUI/qSlicerModuleSelectorToolBar.cxx
```diff
@@ -187,6 +189,9 @@ void qSlicerModuleSelectorToolBarPrivate::init()
   QObject::connect(q, SIGNAL(toolButtonStyleChanged(Qt::ToolButtonStyle)), this->NextButton, SLOT(setToolButtonStyle(Qt::ToolButtonStyle)));
   this->NextButton->setEnabled(this->NextHistoryMenu->actions().size() > 0);
   this->NextButton->setShortcut(QKeySequence(Qt::ControlModifier + Qt::Key_Right));
+
+  // Install event filter to handle mouse back/forward buttons globally
+  qApp->installEventFilter(q);
```
ruffsl, line 205: I'm not sure why `ViewExtensionsManagerAction` is inserted using a loop over child objects, perhaps that preferred over installing event filters over the entire app? For example, binding the shortcuts this way still alters the module selection even when a pop-up window is in focus and not the main windows. E.g. press Ctrl+F to bring up the Module finder pop-up, then press Ctrl+Left or Right. This action while the main window is not in focus is perhaps not expected. A web browser wouldn't do that if a file-picker pop-up was in focus.

ruffsl (commented):
Could use some refinement, but my Qt foo is a little rusty.

cpinter:
I think this is reasonable. I don't remember any other use of Alt-Left / Right that would have a conflict with this.

@ruffsl can you please make a screenshot of the changed UI? It would be easier to make an educated decision (especially that you ask) based on that rather than just the .ui code (at least for me). Thank you!

ruffsl:
> can you please make a screenshot of the changed UI?

The UI file changes currently just ensure these are included and rendered Keyboard Shortcuts Reference menu: 

image

The current icons are placeholders, but best to reuse the same `qSlicerModuleSelectorToolBar` icons for consistency:

image

- Previous
  - `QStyle::SP_ArrowLeft`
- Next
  - `QStyle::SP_ArrowRight`
- Module Finder
  - `":Icons/ModuleHistory.png"`

Wasn't sure how best to reference a C++ symbol like `QStyle::SP_ArrowLeft` that in QML, so placeholders it was.

cc @cpinter

cpinter:
Thank you! A few nitpicks:
* Previous -> Previous module
* Next -> Next module
* Previous modules -> Switch to previously used module
* Next modules -> accordingly
* Find module -> Find module by name

> Wasn't sure how best to reference a C++ symbol like QStyle::SP_ArrowLeft

I'd probably try to leave it undefined in the ui file and set it in the C++ class in the setup function.

Please remove the commented lines if they will not serve a purpose.

ruffsl:
@cpinter , how this as of f7fce0f :

image

cpinter:
It looks great to me, thank you! I cannot approve the PR on my own, but if it looks good to @jcfr or @lassoan or @pieper then one of them can.

pieper:
Looks like there's a lint issue.  Otherwise I have no problem merging this.  I don't have a working mac build currently to test but we can check it out using a preview build.  Alt should just be the option key on mac.

ruffsl:
@pieper , check if that passes. Odd that pre-commit didn't catch raise the clang-format issue.

pieper:
@jcfr do you think this is good to merge before the release?  It's a small change but I don't want it to break something.

jcfr:
That seems reasonable, I will review carefully and integrate before moving forward with Slicer 5.10.

jcfr:
@ruffsl Thanks again for working on these changes :pray: 

After cleaning up the commit series, I realized that we could de-duplicate and have the actions listed in the dialog.

Also, since the default shortcuts were `CTRL+Left|Right`, I included the generic key sequence  `Back` and `Forward` (mapped to `Alt+Right|Left` on Linux and Windows). See https://doc.qt.io/archives/qt-5.15/qkeysequence.html#standard-shortcuts

Once you had a chance to test, we will integrate.

ruffsl:
@jcfr , thanks for cleanup and de-duplication, much cleaner now, and great to learn from.

> Once you had a chance to test, we will integrate.

I rebuilt Slicer from d21f699ced642fe0da9b58bfe2f2cf5d9cd71d32 and tested out locally on Ubuntu 24.04 with my MX Master 3S mouse, and all seem well:

image

The the mouse buttons still work, as do the arrow shortcuts for Next and Previous with either Alt or Crtl modify keys.

Base/QTGUI/qSlicerModuleSelectorToolBar.cxx
ruffsl: Mouse events are still global, but at least now keyboard events can focus on pop-up windows in case the implement their own navigation/capture behavior.

jcfr: > re: ViewExtensionsManagerAction

Since the action is owner by `qSlicerModuleSelectorToolBar` and not defined in the mainwindows UI file, doing the insertion loop is sensible.

jcfr (approved):

jcfr:
Thanks for testing 🙏 , we will now move forward with the integration 🚀

PR #8686, opened 2025-09-01
BUG: Fix automatic conversion of annotation nodes

lassoan:
Legacy annotation nodes are automatically converted to markup nodes. This conversion had a few issues that are fixed in this commit:
- convert annotation nodes in scene views
- enforce converting legacy annotation nodes before converting legacy scene view nodes (because the scene view conversion function is not prepared to deal with annotation hierarchy)
- orphan annotation display nodes (that were not used by any annotation nodes) are now removed
- prevent converted scene views to be automatically restored (overwriting state stored in the main scene)

As a result, vtkMarkupsAnnotationSceneTest now passes.


Modules/Loadable/Markups/Testing/Cxx/vtkMarkupsAnnotationSceneTest.cxx
```diff
@@ -37,22 +38,36 @@ int vtkMarkupsAnnotationSceneTest(int argc, char* argv[])
...
-  //  | |--- M_1
-  //  |--- ROI list
-  //  | |--- R
-  //  | |--- R_1
-  //  | |--- R_2
-  //  |--- List
-  //    |--- F_4
-  //    |--- F_3
   //
+  // All Annotations (vtkMRMLAnnotationHierarchyNode1) -> vtkMRMLAnnotationDisplayNode1
```
jcfr, line 42: Nitpick: The docstring seems a bit more challenging to visually parse, I am wondering if placeholder could be user (e.g `vtkMRMLAnnotationTextDisplayNode1` -> `A_TextDisplay_1`) ?

lassoan: It is only difficult to parse in the github GUI because it wraps the lines.

lassoan: It is important to keep the node IDs to help finding exact nodes in the example scene. I had to create this mapping to figure out what is happening when I fixed the test and it could be useful for future developers.

jcfr (approved):

Sunderlandkyl:
Sorry, I just noticed this now. Thanks for fixing @lassoan!

PR #8690, opened 2025-09-02
ENH: Show storage path for nodes loaded from file in SH

cpinter:
Re https://discourse.slicer.org/t/data-file-information/44274/8

As discussed in the above forum topic, I added a small feature that shows the file path each node is loaded from (if loaded from file and its storage node contains a path). These types of nodes support this feature for now:

Volume:
image

Model:
image

Segmentation
image

Table
image

Markup
image

Transform
image

### DIFF
--- a/Modules/Loadable/SubjectHierarchy/Widgets/qSlicerSubjectHierarchyAbstractPlugin.cxx
+++ b/Modules/Loadable/SubjectHierarchy/Widgets/qSlicerSubjectHierarchyAbstractPlugin.cxx
@@ -285,6 +285,31 @@ QString qSlicerSubjectHierarchyAbstractPlugin::tooltip(vtkIdType itemID) const
 }
 
 //-----------------------------------------------------------------------------
+QString qSlicerSubjectHierarchyAbstractPlugin::tooltipWithStoragePath(vtkIdType itemID) const
+{
+  vtkMRMLSubjectHierarchyNode* shNode = qSlicerSubjectHierarchyPluginHandler::instance()->subjectHierarchyNode();
+  if (!shNode)
+  {
+    qCritical() << Q_FUNC_INFO << ": Failed to access subject hierarchy node";
+    return QString();
+  }
+
+  // If associated data node is a storable node, get the file path from the URL
+  vtkMRMLNode* dataNode = shNode->GetItemDataNode(itemID);
+  if (dataNode != nullptr)
+  {
+    vtkMRMLStorableNode* storableNode = vtkMRMLStorableNode::SafeDownCast(dataNode);
+    if (storableNode != nullptr && storableNode->GetStorageNode() != nullptr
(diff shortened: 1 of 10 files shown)


pieper:
Hi @cpinter this looks very nice.  Can you have a look at the linting issue?

cpinter:
If I interpreted the log correctly the issue was an extra empty line, which I now removed. Let's see if it passes. Thanks for the review!

cpinter:
@pieper CI seems to have passed

hjmjohnson (commented):
This looks like a very useful addition.

Modules/Loadable/SubjectHierarchy/Widgets/qSlicerSubjectHierarchyAbstractPlugin.cxx
```diff
@@ -284,6 +284,31 @@ QString qSlicerSubjectHierarchyAbstractPlugin::tooltip(vtkIdType itemID) const
...
+  }
+
+  // If associated data node is a storable node, get the file path from the URL
+  vtkMRMLNode* dataNode = shNode->GetItemDataNode(itemID);
+  if (dataNode != nullptr)
+  {
+    vtkMRMLStorableNode* storableNode = vtkMRMLStorableNode::SafeDownCast(dataNode);
+    if (storableNode != nullptr && storableNode->GetStorageNode() != nullptr)
+    {
+      QString tooltipWithStoragePath = tr("Loaded from file: ");
```
lassoan: Sentences cannot be assembled using string append, because order of words is different in each language. So, this should be changed to something like `tr("Loaded from file: %1").arg(...)`.

Modules/Loadable/SubjectHierarchy/Widgets/qSlicerSubjectHierarchyAbstractPlugin.cxx
```diff
@@ -284,6 +284,31 @@ QString qSlicerSubjectHierarchyAbstractPlugin::tooltip(vtkIdType itemID) const
...
+      tooltipWithStoragePath.append(storableNode->GetStorageNode()->GetFileName());
```
lassoan: GetFileName() may return nullptr causing crash, so we need a null-pointer check.
Also, we probably don't want to show the line if the string is empty.

lassoan (commented):
It looks good to me, just added a few comments that should be easy to fix.

I'm just wondering how you use this information. Do you remember it and type it into File Explorer or navigate the folder tree? It would be probably more convenient if the path could be copied to the clipboard or the path could be opened in File Explorer - but I don't know where this could be activated (the subject hierarchy context menu is already quite crowded).

cpinter:
> I'm just wondering how you use this information

In the discourse topic referenced above, the user simply wanted to tell apart the different datasets that were loaded (I think seeing the folders they came from helps more than just the name).

PR #8693, opened 2025-09-04
BUG: Improve scene view storage and conversion, fix model visibility issues

lassoan:
This PR contains several fixes and improvements to loading/saving of scene views and automatic conversion of legacy scenes to modern scene views and model hiearchy.

The changes should also fix the `py_nomainwindow_SubjectHierarchyFoldersTest1` test failure.

---

**Details:**

1. Fixes the saving of scene view screenshots by ensuring storage nodes are checked and replaced if unsuitable.
2. Adds missing folder display nodes from the model hierarchy, ensuring proper association with subject hierarchy tree folder items.
3. Corrects the usage of `MRMLNodeModifyBlocker` in containers by implementing move constructor and move assignment operator.
4. Improves the performance of `vtkSlicerSceneViewsModuleLogic::RemoveSceneView` by suppressing sequence browser node modification events.
5. Fixes a memory leak occurring during the conversion of legacy scene views.
6. Ensures model visibility is updated consistently when toggling visibility by always calling `RequestRender`.

PR #8696, opened 2025-09-05
PERF: Improve sequence node storage performance for simple nodes

jcfr:
Simple non-storable nodes that can save their entire state into the MRML scene (without using additional data files) are now saved as a simple `.mrml` file instead of a `.mrb` file. This improves the reading and writing performance of the sequence nodes, as there is no need to zip/unzip the file.

---

> [!NOTE]
>
> This pull request was created cherry-picking commit from @lassoan that was originally associated with https://github.com/Slicer/Slicer/pull/8693


lassoan (approved):
Thank you!

PR #8697, opened 2025-09-05
ENH: Make vtkITKImageWriter more robust

jcfr:
Implemented heuristics to set vector voxel type if it is not specified. Support color images with signed char scalar type.

---

> [!NOTE]
>
> This pull request was created cherry-picking commit from @lassoan that was originally associated with https://github.com/Slicer/Slicer/pull/8693


lassoan (approved):
Thank you!

PR #8700, opened 2025-09-05
BUG: Fix crash in vtkSegmentation::CollapseBinaryLabelmaps

Sunderlandkyl:
When calling vtkSegmentation::CollapseBinaryLabelmaps with an empty segment, vtkOrientedImageDataResample::CalculateEffectiveExtent would find a maximum "invalid" extent, rather than the minimal one. Ex for an empty 1x1x1 image:
  [1, -1, 1, -1, 1,-1]
Instead of:
  [0, -1, 0, -1, 0, -1]

When calling SetExtent on an image, this could cause an error, since the size of the invalid extent could be negative, which would cause a bad_alloc error in vtkGenericDataArray.txx#428.

Fixed by setting invalid extents from vtkOrientedImageDataResample::CalculateEffectiveExtent to the minimal [0, -1, 0, -1, 0, -1]. Added tests with collapsing empty or small segments to vtkSegmentationTest2. Added proper initialization of extents.


Punzo:
thanks for the fix Kyle!

PR #8702, opened 2025-09-06
COMP: Set BUILD_JOB_SERVER_AWARE for extension ExternalProject on CMake ≥ 3.28

jcfr:
Follow-up to 8cb7fc9d5d (“COMP: Ensure BUILD_JOB_SERVER_AWARE option is set for extensions external projects”, 2025-09-03) introduced through https://github.com/Slicer/Slicer/pull/8691.

Fix the `ExternalProject_Add()` argument: use `BUILD_JOB_SERVER_AWARE` (not `JOB_SERVER_AWARE`). The latter applies to custom steps, whereas the former enables GNU Make jobserver propagation for the ExternalProject’s build when a `BUILD_COMMAND` is used.

This prevents intermittent `gmake[*]: *** read jobs pipe: Bad file descriptor.` errors seen on builders using CMake ≥ 3.28 with parallel GNU Make.

### DIFF
--- a/Extensions/CMake/SlicerBlockBuildPackageAndUploadExtensions.cmake
+++ b/Extensions/CMake/SlicerBlockBuildPackageAndUploadExtensions.cmake
@@ -223,9 +223,9 @@ foreach(EXTENSION_NAME ${EXTENSION_LIST})
     ")
 
   if(CMAKE_VERSION GREATER_EQUAL "3.28")
-    set(maybe_JOB_SERVER_AWARE "JOB_SERVER_AWARE 1")
+    set(maybe_BUILD_JOB_SERVER_AWARE "BUILD_JOB_SERVER_AWARE 1")
   else()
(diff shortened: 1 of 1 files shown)


jcfr:
This should address the remaining errors observed on metroplex 🤞

jcfr:
@lassoan Having this integrated before this evening would be ideal 🙏

jamesobutler (approved):

lassoan:
Awesome, thank you, hopefully the extensions will be successfully built tomorrow!

jcfr:
It turns out there was still a last issue ... which should now be fixed with:
* https://github.com/Slicer/Slicer/pull/8704

Due to quoting issue ... the arguments were incorrectly processed as an install command. For more details, see https://github.com/Slicer/Slicer/pull/8704#issuecomment-3265409461

PR #8703, opened 2025-09-07
ENH: Prevent slice offset set outside bounds of single slice volume

jamesobutler:
This is an attempt to update the handling of slice offset bounds for volumes with a dimension with a single slice voxel. The purpose of this change is to avoid user confusion about why moving the slice offset slider for a volume with a single slice voxel results in the slice view going black due to the offset going outside the volume bounds. For testing I was using a Slicer logo png (see below). 

I also considered hiding the slice offset slider where there is only one voxel in the given dimension such as `slicer.app.layoutManager().sliceWidget("Red").sliceController().setShowSliceOffsetSlider(False)`, but I decided to keep the slider visible at all times allowing to navigate the single voxel using an offset resolution that is half the voxel size.

slicer-logo

Slicer 5.9.0-2025-09-06 (bounds -1mm to 1mm even though single slice volume of 1mm Z spacing is from -0.5mm to 0.5mm):

https://github.com/user-attachments/assets/ba753215-51d6-4ade-94d5-57be4701eb01

This PR (bounds -0.5mm to 0.5mm for single slice volume of 1mm Z spacing):

https://github.com/user-attachments/assets/0c9178c5-02d3-480f-a035-f503734c1264

### DIFF
--- a/Libs/MRML/Widgets/qMRMLSliceControllerWidget.cxx
+++ b/Libs/MRML/Widgets/qMRMLSliceControllerWidget.cxx
@@ -1214,15 +1214,18 @@ void qMRMLSliceControllerWidgetPrivate::onSliceLogicModifiedEvent()
 
   double offsetRange[2] = { -1.0, 1.0 };
   double offsetResolution = 1.0;
-  if (!this->SliceLogic || !this->SliceLogic->GetSliceOffsetRangeResolution(offsetRange, offsetResolution))
+  if (this->SliceLogic)
   {
-    return;
+    this->SliceLogic->GetSliceOffsetRangeResolution(offsetRange, offsetResolution);
   }
 
   bool wasBlocking = this->SliceOffsetSlider->blockSignals(true);
-  q->setSliceOffsetRange(offsetRange[0], offsetRange[1]);
+  q->setSliceOffsetRange(offsetRange[0], offsetRange[1] - 0.00001); // Slightly reduce range to ensure volume is displayed at max offset
   q->setSliceOffsetResolution(offsetResolution);
-  this->SliceOffsetSlider->setValue(this->SliceLogic->GetSliceOffset());
+  if (this->SliceLogic)
+  {
+    this->SliceOffsetSlider->setValue(this->SliceLogic->GetSliceOffset());
+  }
   this->SliceOffsetSlider->blockSignals(wasBlocking);
 
   emit q->renderRequested();
--- a/Libs/MRML/Widgets/qMRMLSliceVerticalControllerWidget.cxx
+++ b/Libs/MRML/Widgets/qMRMLSliceVerticalControllerWidget.cxx
@@ -121,7 +121,7 @@ void qMRMLSliceVerticalControllerWidgetPrivate::onSliceLogicModifiedEvent()
   }
 
   bool wasBlocking = this->SliceVerticalOffsetSlider->blockSignals(true);
-  q->setSliceOffsetRange(offsetRange[0], offsetRange[1]);
+  q->setSliceOffsetRange(offsetRange[0], offsetRange[1] - 0.00001); // Slightly reduce range to ensure volume is displayed at max offset
   q->setSliceOffsetResolution(offsetResolution);
   this->SliceVerticalOffsetSlider->setValue(this->SliceLogic->GetSliceOffset());
   this->SliceVerticalOffsetSlider->blockSignals(wasBlocking);
--- a/Libs/MRML/Logic/vtkMRMLSliceLogic.cxx
+++ b/Libs/MRML/Logic/vtkMRMLSliceLogic.cxx
@@ -2806,32 +2806,21 @@ bool vtkMRMLSliceLogic::GetSliceOffsetRangeResolution(double range[2], double& r
   this->GetLowestVolumeSliceBounds(sliceBounds, true);
 
   const double* sliceSpacing = this->GetLowestVolumeSliceSpacing();
-  if (!sliceSpacing)
+  if (!sliceSpacing || (sliceBounds[5] - sliceBounds[4]) <= 0.0)
   {
-    range[0] = -1.0;
-    range[1] = 1.0;
-    resolution = 1.0;
+    // No valid volume data - use default values
+    range[0] = sliceSpacing ? -sliceSpacing[2] / 2 : -1.0;
+    range[1] = sliceSpacing ? sliceSpacing[2] / 2 : 1.0;
+    resolution = sliceSpacing ? sliceSpacing[2] / 2 : 1.0;
     return false;
   }
 
-  // Set the scale increments to match the z spacing (rotated into slice space)
-  resolution = sliceSpacing ? sliceSpacing[2] : 1.0;
+  // Set resolution based on slice spacing
+  resolution = sliceSpacing[2];
 
-  bool singleSlice = ((sliceBounds[5] - sliceBounds[4]) < resolution);
-  if (singleSlice)
-  {
-    // add one blank slice before and after the current slice to make the slider appear in the center when
-    // we are centered on the slice
-    double centerPos = (sliceBounds[4] + sliceBounds[5]) / 2.0;
-    range[0] = centerPos - resolution;
-    range[1] = centerPos + resolution;
-  }
-  else
-  {
-    // there are at least two slices in the range
-    range[0] = sliceBounds[4];
-    range[1] = sliceBounds[5];
-  }
+  // Use the actual volume bounds as the range
+  range[0] = sliceBounds[4];
+  range[1] = sliceBounds[5];
 
   return true;
 }


hjmjohnson (dismissed):
This seems like a good way to address the issue.

Libs/MRML/Widgets/qMRMLSliceVerticalControllerWidget.cxx
```diff
@@ -121,7 +121,7 @@ void qMRMLSliceVerticalControllerWidgetPrivate::onSliceLogicModifiedEvent()
   }
 
   bool wasBlocking = this->SliceVerticalOffsetSlider->blockSignals(true);
-  q->setSliceOffsetRange(offsetRange[0], offsetRange[1]);
+  q->setSliceOffsetRange(offsetRange[0], offsetRange[1] - 0.00001); // to display volume at max offset, reduce range just inside bounds
```
jcfr:
```suggestion
  q->setSliceOffsetRange(offsetRange[0], offsetRange[1] - 0.00001); // Slightly reduce range to ensure volume is displayed at max offset
```

Libs/MRML/Widgets/qMRMLSliceControllerWidget.cxx
```diff
@@ -1214,15 +1214,18 @@ void qMRMLSliceControllerWidgetPrivate::onSliceLogicModifiedEvent()
...
-  if (!this->SliceLogic || !this->SliceLogic->GetSliceOffsetRangeResolution(offsetRange, offsetResolution))
+  if (this->SliceLogic)
   {
-    return;
+    this->SliceLogic->GetSliceOffsetRangeResolution(offsetRange, offsetResolution);
   }
 
   bool wasBlocking = this->SliceOffsetSlider->blockSignals(true);
-  q->setSliceOffsetRange(offsetRange[0], offsetRange[1]);
+  q->setSliceOffsetRange(offsetRange[0], offsetRange[1] - 0.00001); // to display volume at max offset, reduce range just inside bounds
```
jcfr:
```suggestion
  q->setSliceOffsetRange(offsetRange[0], offsetRange[1] - 0.00001); // Slightly reduce range to ensure volume is displayed at max offset
```

Libs/MRML/Widgets/qMRMLSliceVerticalControllerWidget.cxx
```diff
@@ -121,7 +121,7 @@ void qMRMLSliceVerticalControllerWidgetPrivate::onSliceLogicModifiedEvent()
   }
 
   bool wasBlocking = this->SliceVerticalOffsetSlider->blockSignals(true);
-  q->setSliceOffsetRange(offsetRange[0], offsetRange[1]);
+  q->setSliceOffsetRange(offsetRange[0], offsetRange[1] - 0.00001); // Slightly reduce range to ensure volume is displayed at max offset
```
lassoan, line 124: 0.00001 may be larger than the difference between offsetRange[0] and offsetRange[1]. Instead of some arbitrary fixed value, use a fraction of the range.

Libs/MRML/Widgets/qMRMLSliceControllerWidget.cxx
```diff
@@ -1214,15 +1214,18 @@ void qMRMLSliceControllerWidgetPrivate::onSliceLogicModifiedEvent()
...
-  if (!this->SliceLogic || !this->SliceLogic->GetSliceOffsetRangeResolution(offsetRange, offsetResolution))
+  if (this->SliceLogic)
   {
-    return;
+    this->SliceLogic->GetSliceOffsetRangeResolution(offsetRange, offsetResolution);
   }
 
   bool wasBlocking = this->SliceOffsetSlider->blockSignals(true);
-  q->setSliceOffsetRange(offsetRange[0], offsetRange[1]);
+  q->setSliceOffsetRange(offsetRange[0], offsetRange[1] - 0.00001); // Slightly reduce range to ensure volume is displayed at max offset
```
lassoan, line 1223: Don't use hardcoded value - see below.

lassoan (commented):
I would like to review and test this more, but for now just added two comments

Libs/MRML/Widgets/qMRMLSliceVerticalControllerWidget.cxx
jamesobutler: Or maybe a fraction of the spacing? For a 1mm spacing single slice volume from -0.5mm to 0.5mm, the closer the positive offset was to 0.5000000 the more likely it didn't render the volume at that slice offset as though that bounds position was not inclusive of the volume.

jcfr: Here is a function we could add:

```cpp
// Calculate a small epsilon used to trim the maximum of the slice offset range
// so that the last slice still renders correctly.
// Epsilon is based on:
//   - a fraction of the total offset range width
//   - a fraction of the slice spacing (offsetResolution)
// and is capped to at most a fraction of the offset resolution step.
static double computeSliceOffsetRangeEpsilon(double offsetRangeMin,
                                             double offsetRangeMax,
                                             double offsetResolution)
{
  const double offsetRangeWidth = offsetRangeMax - offsetRangeMin;
  if (offsetRangeWidth <= 0.0)
    {
    return 0.0;
    }

  // Constants controlling epsilon behavior
  const double rangeFraction   = 1e-6;   // one-millionth of range width
  const double spacingFraction = 1e-6;   // one-millionth of slice spacing
  const double stepCapFraction = 0.25;   // cap at one quarter of a step

  double epsilonFromRange   = offsetRangeWidth * rangeFraction;
  double epsilonFromSpacing = (offsetResolution > 0.0)
                            ? offsetResolution * spacingFraction
                            : offsetRangeWidth; // fallback large value
  double epsilonCap         = (offsetResolution > 0.0)
                            ? offsetResolution * stepCapFraction
                            : offsetRangeWidth;

  return qMin(qMin(epsilonFromRange, epsilonFromSpacing), epsilonCap);
}

} // namespace
```

Libs/MRML/Widgets/qMRMLSliceControllerWidget.cxx
jcfr:
```suggestion
  double epsilon = computeSliceOffsetRangeEpsilon(offsetRange[0], offsetRange[1], offsetResolution);
  q->setSliceOffsetRange(offsetRange[0], offsetRange[1] - epsilon);
```

lassoan:
The implementation could be simplified, if we wanted to make the user see an image regardless of where the slider handle is. We could simply change the slice logic to return `[-spacing/2+eps, spacing2-eps]` for single-slice volumes. This way there is no need to add some nudge factor (eps) at the GUI level and thereby avoiding inconsistencies between GUI and logic (for example, slice position could be slightly different when setting it from logic vs. setting it from GUI if some eps is added at the GUI level).

However, this implementation could easily result in the slice view plane positioned at the edge of an image. Having the slice view plane near the edge of an object can lead to several complications: 
- Rendering result would be somewhat unpredictable: if the view rendered only one object (the image) then we could easily find a comfortable nudge factor (eps) that would make the image reliably appear in all slider positions. However, the slice view may display many different objects, such as model intersections, markups, other images, potentially under transforms, etc. It may be impossible to find an eps that results in perfect rendering of all these objects. Too small eps may not result in cutting inside every object, too large eps may start to result in visible differences due to not being exactly at the edge.
- User could place markups at the slice edge (at spacing/2 distance from the slice center) because there would be no clear indication that the view is off center.
- Non-uniform reslice positions: the distance between the first and second slider position (spacing/2-eps) would not be exactly the same as the jump when you use the keyboard (spacing/2), this could lead to markups, etc. not exactly placed at the same position, depending on whether you used your mouse (to move the slider) or your keyboard. Slice spacing can be relatively large, so this error may not be negligible.

It is better to avoid such "edge cases", by making slice views always go through the center of voxels. We would not need to ever nudge the slice position in any directions. This is exactly what the current implementation in slicer/main does - at every slice slider position, the view plane is exactly at the center of an image slice or exactly +/-1 spacing away.

How to solve the cosmetic GUI problem with having a slider that does not really have a useful function (it steps outside of the current image slice)? A few options that does not involve nudge factors:
- Option A: Simply set the slider range to [`offset`, `offset`] and its position to `offset`. As the slider is implemented now, it would not look good, because the slider handle would be displayed on the left. We could introduce a special flag into the `ctkDoubleSlider` widget to render the handle in the center when there is only a single position. However, this would require some work, and the slider would still not have any function. With some more effort, we could probably implement clicking the handle then would jump to the `offset` position, which would be a somewhat useful function. Still, it would not be a very efficient use of space.
- Option B: Hide the slider when it only has a single position. What would we use the space for? We would really need a sequence browser slider in slice views for time sequences (the single shared slider on the toolbar is intuitive when there is only one global time sequence, but when showing independent image sequences in slice views then the shared slider is practically unusable). We could use the slice view controller header to show both the current (spatial) offset slider and a sequence browser slider. This would work perfectly for reviewing 2D+t images (e.g., ultrasound, fluoroscopy). For 3D+t images (e.g., 4D CT and 4D US) we could have two sliders share the same space (stacked horizontally or vertically). Or we could use the vertical slider for spatial offset (as in all other software) and use a horizontal slider for time browsing (as in all other software). We could also consider moving to rendering the sliders over the image instead of as a separate widget outside the slice view - in this case hiding the slider would mean that we would not occlude the image.

Option A. may be a little less work but Option B. would be a much more forward-looking solution. It would make browsing of images in Slicer more similar to other software, solve sequence browsing, and potentially allowing making slice views larger by eliminating the slice view control bars. So, I would vote for Option B (unless there are even better suggestions), implementing after the new stable version is released.

jamesobutler:
Abandoning this work now based on above discussion around redesigning controls for 2D sequences with controls in the slice views and possible reorganization of offset sliders to vertical controls.

PR #8710, opened 2025-09-09
BUG: Fix loading of multi-file volumes in SampleData module

lassoan:
When clicked on sample data sets that consisted of multiple files, such as DTIVolume = DTIVolume.raw.gz and DTIVolume.nhdr then the data set failed to load. This was caused by automatically generating node name for DTIVolume.raw.gz file, while it was not meant to be loaded into the scene.

Fixed by only generating default node names if node name was not specified for any of the files in the sample.

### DIFF
--- a/Modules/Scripted/SampleData/SampleData.py
+++ b/Modules/Scripted/SampleData/SampleData.py
@@ -700,10 +700,16 @@ class SampleDataLogic:
         resultNodes = []
         resultFilePaths = []
 
+        # If some node names are defined and some are left empty then we assume that it is intentional
+        # (e.g., you only want to load the image.nhdr file and not the image.raw file).
+        # In this case default node names are not generated.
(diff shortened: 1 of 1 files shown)


jcfr (commented):
Looks good, I suggest to use `Squash & Merge` specifying the a commit message like this one:

```
BUG: Fix loading of multi-file volumes in SampleData

Sample datasets with multiple files (e.g., `DTIVolume.nhdr` + 
`DTIVolume.raw.gz`) failed to load because a node name was 
automatically generated for the raw companion file, which is
not meant to be loaded.

Fixes regression introduced in 89104423af ("BUG: Ensure MRB 
loading correctly reports loaded nodes", 2025-01-10).
```

lassoan:
Thanks for the review, I've updated the commit message as you suggested.

jcfr (approved):

PR #8713, opened 2025-09-09
BUG: Update SampleData test for multi-file volume loading behavior

jcfr:
Adapt SampleData test expectations to reflect the recent fix for multi-file volume loading. The test now asserts that only a single node is created (e.g., `.nhdr` + `.raw.gz` pair) instead of treating the sidecar file as a separate node.

Follows up on 2218414aeb ("BUG: Fix loading of multi-file volumes in SampleData module", 2025-09-08) introduced through the following pull request:
* https://github.com/Slicer/Slicer/pull/8710

---

This fixes failure like this one: https://slicer.cdash.org/tests/31072271


jcfr:
@lassoan This will fix the last failing test, it would be great to review & approve before the next round of nightly builds :pray:

jamesobutler (approved):
Agreed that the expectation is for 1 node as the .raw.gz won't be loaded as its own, only the .nhdr.

lassoan:
Thank you, the changes look good to me.

PR #8718, opened 2025-09-11
BUG: Fix reading of vector volumes

### LINKED ISSUE #8707: Slicer is now unable to load some 4D nrrd saved with SimpleITK
## Summary

With the recent changes made in https://github.com/Slicer/Slicer/commit/11df881b46c907c92947ae89d1ffb3ef9cd95142, I have observed that some 4D nrrd volumes are now unable to load into the scene as they have been in older versions Slicer 5.9.0-2025-08-18 and earlier. cc: @lassoan @cpinter 

## Steps to reproduce

```python
import os
import numpy as np
import SampleData
SampleData.SampleDataLogic().downloadSample("CTCardioSeq")

# Find the CTCardioSeq volume node by name
volume_node = slicer.util.getNode('CTCardioSeq')

browser = slicer.mrmlScene.GetFirstNodeByName("CTCardioSeq browser")
sequence_node = browser.GetSequenceNode(volume_node)
num_frames = sequence_node.GetNumberOfDataNodes()
arr4d = np.stack([
    slicer.util.arrayFromVolume(sequence_node.GetNthDataNode(i))
    for i in range(num_frames)
])
print("Shape of 4D array:", arr4d.shape)

arr4d_reordered = np.transpose(arr4d, (1, 2, 3, 0))

import SimpleITK as sitk
image = sitk.GetImageFromArray(arr4d_reordered)
writer = sitk.ImageFileWriter()
writer.UseCompressionOn()
file_path = os.getenv("USERPROFILE") + "/Downloads/CTCardioSeq.seq.nrrd"
writer.SetFileName(file_path)
writer.Execute(image)
slicer.mrmlScene.Clear()

nodes = vtk.vtkCollection()
# Requires SlicerIGSIO extension
browser_node = slicer.modules.metafileimporter.logic().ReadVolumeSequence(file_path, nodes)
```

| Slicer 5.9.0-2025-08-18 | Slicer 5.9.0-2025-08-20|
|------|----------|
|Image|Image|

I also have my own 3Dt cardiac ultrasound volume which has the following header information which now results in 3D Slicer crashing on load.
```
NRRD0004
# Complete NRRD file format specification at:
# http://teem.sourceforge.net/nrrd/format.html
type: unsigned char
dimension: 4
space: left-posterior-superior
sizes: 100 128 170 79
space directions: none (-0.10000000000000001,0,0) (0,-0.077150700000000003,0) (0,0,0.14897761153846142)
kinds: vector domain domain domain
encoding: gzip
space origin: (-68.143872099999996,-14.6,3.4982884999999988)
```
Trying to replicate with some Slicer sample data, if I save CT Cardio as a .seq.nhdr and then manually update the nhdr with type changed from short to "unsigned char" and the kinds "list" changed to "vector" results in Slicer 5.9.0-2025-08-20 crashing which did not happen with Slicer 5.9.0-2025-08-18. See header information below that causes this:
```
NRRD0005
# Complete NRRD file format specification at:
# http://teem.sourceforge.net/nrrd/format.html
type: unsigned char
dimension: 4
space: right-anterior-superior
sizes: 10 128 104 72
space directions: none (-1.1499999999999999,0,0) (0,-1.1499999999999999,0) (0,0,2)
kinds: vector domain domain domain
labels: "frame" "" "" ""
endian: little
encoding: gzip
space origin: (62.20000000000001,68,1692.0989989999998)
measurement frame: (1,0,0) (0,1,0) (0,0,1)
data file: CTCardioSeq.seq.raw.gz
DataNodeClassName:=vtkMRMLScalarVolumeNode
axis 0 index type:=numeric
axis 0 index values:=0 1 2 3 4 5 6 7 8 9
```

## Environment
- Slicer version: Slicer-5.9.0-2025-08-20 and later
- Operating system: Windows

lassoan:
This is a preliminary implementation to get quick feedback. Complete implementation will include testing of the sequence reader with a variety of example files.

fixes #8707


cpinter:
This looks reasonable to me, thank you @lassoan !
The only thing I don't understand is passing the `AxesReorder` value directly from one imageIO to the other. Before we used the fixed `SetAxesReorderToUseNonListRangeAxisAsPixel` option. What values do we expect to have now?

lassoan:
Normally we would want to use `UseNonListRangeAxisAsPixel` strategy for axis reordering. However, when there is no list axis then we switch to `UseScalarPixel`, which makes the vector axis used as sequence axis (instead of pixel component axis).

I though not introducing a new `axesReorder` variable makes things simpler (and inherently harmonizes the axis reorder strategy between the metadata reading step and the pixel loading step). But maybe it is better to add that variable.

cpinter:
Thanks! What I don't understand is where the `UseScalarPixel` value is set. Is it automatically set by the image IO in ITK?

lassoan:
These are the places where the axes reorder strategy is set:
- https://github.com/Slicer/Slicer/blob/17be79aa1c049ff45c18c095b17790e2434e9523/Libs/vtkITK/vtkITKImageSequenceReader.cxx#L275
- https://github.com/Slicer/Slicer/blob/17be79aa1c049ff45c18c095b17790e2434e9523/Libs/vtkITK/vtkITKImageSequenceReader.cxx#L316

Libs/MRML/Core/Testing/vtkMRMLVolumeSequenceStorageNodeTest1.cxx
```diff
@@ -0,0 +1,173 @@
...
+#include "vtkMRMLScene.h"
+#include "vtkMRMLSequenceNode.h"
+#include "vtkMRMLVectorVolumeNode.h"
+#include "vtkMRMLVolumeSequenceStorageNode.h"
+
+#include "vtkImageData.h"
+#include "vtkMatrix3x3.h"
+#include "vtkMatrix4x4.h"
+#include "vtkNew.h"
+#include "vtkPointData.h"
```
jcfr:
```suggestion
#include <vtkImageData.h>
#include <vtkMatrix3x3.h>
#include <vtkMatrix4x4.h>
#include <vtkNew.h>
#include <vtkPointData.h>
```

Libs/vtkITK/vtkITKImageSequenceReader.cxx
```diff
@@ -272,43 +283,73 @@ void vtkITKImageSequenceReader::ExecuteDataWithInformation(vtkDataObject* output
     imageIO->SetFileName(this->GetFileName());
     imageIO->ReadImageInformation(); // Read only the header information
 
+    // Workaround for NRRD files in that the range axis has kind=vector:
+    // If there is no axis with kind=list but there is an axis with kind=vector then force using scalar pixel.
```
jcfr:
```suggestion
    // Backward-compatibility for legacy NRRD writers:
    // Some 3D+t sequences were encoded with kinds = {vector, domain, domain, domain},
    // i.e., no "list" axis. In these files the leading "vector" axis does not represent
    // per-voxel components, but the temporal/range axis (frames).
    //
    // If no axis of kind "list" is present but a "vector" axis is present
    // (i.e., the image reports multiple scalar components), interpret the data
    // as a sequence of scalar volumes by forcing scalar pixels
    // (AxesReorderToUseScalarPixel) and re-reading the header. This preserves
    // the intended 3D+t semantics and avoids crashes.
```

Libs/vtkITK/vtkITKImageSequenceReader.cxx
```diff
@@ -338,14 +379,39 @@ void vtkITKImageSequenceReader::ExecuteDataWithInformation(vtkDataObject* output
       }
     } // For all keys in the metadata dictionary
 
-    // Set SequenceAxisLabel and SequenceAxisUnit from AxisLabels and AxisUnits, respectively,
-    // if the frames are stored in the dimensions instead of the components
-    if (listDim >= 0)
+    // Get list dimension, which is the first "list" kind of axis.
+    // If there is no "list" kind axis then we accept the first "vector" kind as well, because
+    // in some legacy files, the range axis kind was set to "vector".
```
jcfr:
```suggestion
    // Identify the sequence (range) axis:
    // Prefer the first axis with kind "list". If none exists, fall back to the first
    // axis with kind "vector" (legacy files that mislabeled the range axis as "vector").
    // Note: this fallback only applies when a proper "list" axis is absent.
```

jamesobutler:
Thanks for the fix and the tests. I can confirm this works well for my own datasets with the ambiguous metadata when writing my 4D array using SimpleITK. I have confirmed my various testing snippets in #8707 now work as expected.

lassoan:
Thanks for the reviews and testing. I've committed all the suggested improvements to comments, made a small change (changed the base class of the reader to match the base class of all the other file readers in vtkITK) and force-pushed.

It would be great if someone could have a final look and approve.

jamesobutler (commented):
@lassoan With your latest changes I'm observing relevant sequences testing failing. This wasn't the case prior to your latest pushes.

The testing volume mentioned at https://github.com/Slicer/Slicer/issues/8707#issuecomment-3285031955 no longer loads successfully.
`[VTK] Input port 0 of algorithm vtkITKImageSequenceReader (000001F52025B370) has 0 connections but is not optional.`

```
TestVolumeSequenceStorage: C:/S5R/Slicer-build/Libs/MRML/Core/Testing/TestData/Volumes-DomainList.seq.nrrd
2025-09-14 10:50:47.592 (   0.026s) [22AF363151C85417]vtkDemandDrivenPipeline:677    ERR| vtkCompositeDataPipeline (00000147BED36870): Input port 0 of algorithm vtkITKImageSequenceReader (00000147BEC963C0) has 0 connections but is not optional.
2025-09-14 10:50:47.592 (   0.026s) [22AF363151C85417]vtkMRMLSequenceNode.cxx:673    ERR| vtkMRMLSequenceNode (00000147BED37D60): vtkMRMLSequenceNode::GetNthDataNode failed: itemNumber 5 is out of range

Line 79 - volumeNode is NULL : CheckNotNull failed
	pointer:0000000000000000

Line 143 - TestVolumeSequenceStorage(volume_DomainList, 1, 32, 27, 15, 10, true, 0, 10, 20, 8, 5, 454, tempDir) != EXIT_SUCCESS : CheckInt failed
	current :1
	expected:0
Assertion failed in C:\S5R\Slicer-build\Libs\MRML\Core\Testing\MRMLCoreCxxTests.cxx:690 - expected 0 error or warning messages, got 2
```

lassoan:
Thanks for testing. Could you test without thr second commit? (the second commit should be the only difference compared to the previous content of the PR, it is not essential, so if that is causing trouble I'll just remove it for now)

jamesobutler:
Yes with `STYLE: Change vtkITKImageSequenceReader parent class to vtkImageAlgorithm` picked off, the tests are passing.

lassoan:
Thank you for the additional testing. I've force-pushed the PR without the second commit. This should be all ready to go then.

cpinter (approved):
I think it can be integrated. Thank you very much!

PR #8722, opened 2025-09-14
BUG: Fix Undo/Redo in the segment editor logic

### LINKED ISSUE #8721: Undo/Redo in the segment editor is not activated when loading a saved scene.
Using `5.9.0-2025-09-12 r33943` / `76db15f` on Linux.

If a scene containing a segmentation is loaded, the `Undo/Redo` buttons are not activated if we modify anything in the segment editor.

In a new segmentation in an empty scene, the buttons do light up when a new segment is added (after setting the segmentation itself for geometry).

No extension is installed and `slicerrc.py` is not loaded.

Thanks for having a look if it can be confirmed.

Thibault-Pelletier:
Fix Undo/Redo in the segment editor not activated when loading a saved scene.

Problem was due to not observing the segmentation node modified event and not synchronizing the segmentation history to the segmentation the first time it was set.

This fix properly calls the observer reconnect and segmentation history synchronization methods on set.

Closes #8721

### DIFF
--- a/Modules/Loadable/Segmentations/Logic/vtkSegmentEditorLogic.cxx
+++ b/Modules/Loadable/Segmentations/Logic/vtkSegmentEditorLogic.cxx
@@ -764,6 +764,9 @@ void vtkSegmentEditorLogic::SetSegmentEditorNode(vtkMRMLSegmentEditorNode* newSe
       });
 
     this->SegmentEditorNodeObs = newSegmentEditorNode->AddObserver(vtkCommand::ModifiedEvent, updateCommand);
+
+    // Update the segment editor's segmentation node observers
+    this->ReconnectSegmentationNodeObserver();
   }
 }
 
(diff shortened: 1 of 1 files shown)


lassoan (approved):
The changes look good to me and because they fix a serious error, they need to be merged now.

However, it would be nice to make this vtkSegmentEditorLogic class follow Slicer conventions:
- vtkSegmentEditorLogic is a logic class, which should derive from vtkMRMLAbstractLogic and it use the observer manager instead of manually adding/removing observations (which is complex and error-prone).
- `Synchronize` is not a commonly used verb and the `SynchronizeSegmentationHistorySegmentation` method actually just simply sets the segmentation into the segmentation history. The method should be called `Set...` but moving out this basic low-level operation to a separate method makes the code less readable and more error-prone. It would be better to get rid of this method and call `SetSegmentation` directly.

lassoan:
@Thibault-Pelletier thanks for monitoring the user reports and quickly delivering the fix!

Thibault-Pelletier:
@lassoan thanks for the merge and the feedbacks! I'll have a look at the vtkMRMLAbstractLogic interface.

PR #8582, opened 2025-07-17
BUG: Fix improper Python initialization causing inconsistent interpreter state

jcfr:
Fixes regression introduced in:
- 7a26aa1b87 ("ENH: Update Python initialization for compatibility with isolated mode")
- 3bace0b4b6 ("ENH: Update Python initialization to use PyConfig API")

These changes inadvertently introduced multiple and inconsistent Python initializations by removing the use of deprecated symbols (`PySys_SetArgvEx`, `Py_NoUserSiteDirectory`) without fully replicating their intended behavior. As a result, over 100 tests failed (e.g., `AttributeError: module 'qt' has no attribute 'QTimer'` on Ubuntu) due to Python being initialized both too late and with conflicting settings.

This commit restores correct initialization behavior by:

1. Initializing Python early in `qSlicerCorePythonManager` using the `PyConfig` API. Rather than explicitly setting `config.isolated = 0`, this relies on the default (non-isolated) mode established by `Py_InitializeFromConfig`, which ensures that environment variables and site directories are respected. Early initialization ensures Python is ready before QApplication starts.

2. Preventing re-initialization by `PythonQt` by setting the `PythonAlreadyInitialized` flag. The previously required clearing of the `IgnoreSiteModule` flag is no longer necessary, as `PythonQt` no longer manages initialization.

3. Moving argument population (`PyConfig_SetArgv`) and interpreter config update (`_PyInterpreterState_SetConfig`) into `qSlicerCoreApplication::handleCommandLineArguments()` to occur after command-line arguments are fully known.

This approach ensures consistent interpreter state across embedded Python and PythonQt integration, and aligns with the best practices of the `PyConfig` API.

----

Follow up of the following pull request(s):
* https://github.com/Slicer/Slicer/pull/8548

PR #8727, opened 2025-09-16
BUG: Preserve `sys.path` during Slicer module discovery & add startup test

RafaelPalomar:
_The original pull request description was update by @jcfr to describe the root cause and associated fix._

This PR fixes a regression where updating `argv` via the PyConfig API caused the live interpreter’s `sys.path` to be replaced with `config.module_search_paths`, dropping entries added during Slicer's module discovery. It also adds an application Python test.

## Background

In fa915a03fc4 ("BUG: Fix improper Python initialization causing inconsistent interpreter state", 2025-07-17) we switched to updating `argv` using `PyConfig_SetArgv()` and `_PyInterpreterState_SetConfig()`,

While correct for `argv`, the latter **repopulates `sys.path` from the config**, which typically lacks path inserted during Slicer module discovery. Result: imports depending on those entries could fail after startup.

## Summary

Add `py_startup_preserve_sys_path` test and update `qSlicerCoreApplication::handleCommandLineArguments` to perform the following steps:
* Before calling `_PyInterpreterState_SetConfig`, snapshot the current `sys.path` using PythonQt.
* Apply the updated config (to update `argv`).
* Restore the exact `sys.path` afterwards (`PythonQt::overwriteSysPath`), preserving order and duplicates.

---

<details><summary>Previous description</summary>
<p>

This adds a new python test which tries to (1) modify `sys.path` on module load and (2) verify preservation of the changes after slicer startup. The mock module structure as follows:

```txt
SlicerStartupModuleTestHelperModule.py                   # Mock module which imports SlicerStartupModuleTestHelperPackage
SlicerStartupModuleTestHelperPackage/__init__.py   # Mock Python package which modifies sys.path
SlicerStartupModuleTestHelperScript.py                      # Startup script which verifies if sys.path was modified accordingly
```

</p>
</details>

### DIFF
--- a/Base/QTCore/qSlicerCoreApplication.cxx
+++ b/Base/QTCore/qSlicerCoreApplication.cxx
@@ -1198,6 +1198,32 @@ CTK_GET_CPP(qSlicerCoreApplication, bool, isURIArgumentHandlingEnabled, URIArgum
 CTK_SET_CPP(qSlicerCoreApplication, bool, setURIArgumentHandlingEnabled, URIArgumentHandlingEnabled);
 
 //-----------------------------------------------------------------------------
+namespace
+{
+
+// Mirrors CPython's PyWideStringList_Clear only available internally:
+// Items in PyWideStringList are allocated with the "raw" allocator;
+// free them with PyMem_RawFree (mirrors CPython).
+void ClearWideStringList(PyWideStringList* list)
+{
+  if (!list || !list->items)
+  {
+    list->length = 0;
+    list->items = nullptr;
+    return;
+  }
+  for (Py_ssize_t i = 0; i < list->length; ++i)
+  {
+    PyMem_RawFree(list->items[i]);
+  }
+  PyMem_RawFree(list->items);
+  list->length = 0;
+  list->items = nullptr;
+}
+
+} // namespace
+
+//-----------------------------------------------------------------------------
 void qSlicerCoreApplication::handleCommandLineArguments()
 {
   Q_D(qSlicerCoreApplication);
@@ -1218,9 +1244,9 @@ void qSlicerCoreApplication::handleCommandLineArguments()
 #else
   if (!qSlicerCoreApplication::testAttribute(qSlicerCoreApplication::AA_DisablePython))
   {
-    // Snapshot current sys.path BEFORE applying the config
-    // Rationale: _PyInterpreterState_SetConfig will replace sys.path from config.
-    // We want an exact restore of the sys.path augmented during Slicer module discovery.
+    // Snapshot current sys.path BEFORE applying the config.
+    // Rationale: _PyInterpreterState_SetConfig replaces sys.path from the config.
+    // We capture the runtime-augmented sys.path so we can install it into the config.
     PythonQtObjectPtr sys;
     sys.setNewRef(PyImport_ImportModule("sys"));
     PythonQtObjectPtr sysPath;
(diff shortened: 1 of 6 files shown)


jcfr:
Related issues:
* https://github.com/Slicer/Slicer/issues/8735

jcfr:
@RafaelPalomar  Thanks again for contributing the test, it was really helpful :pray: 

The regression should now be addressed.

Base/QTCore/qSlicerCoreApplication.cxx
```diff
@@ -1263,13 +1264,38 @@ void qSlicerCoreApplication::handleCommandLineArguments()
...
+    // sys.path from the config's module_search_paths)
     if (_PyInterpreterState_SetConfig(&config) < 0)
     {
       PyConfig_Clear(&config);
       PyErr_Print();
       qSlicerCoreApplication::terminate(EXIT_FAILURE);
     }
 
+    // Restore sys.path exactly as it was before calling "_PyInterpreterState_SetConfig"
+    PythonQt::self()->overwriteSysPath(savedSysPath);
```
lassoan: Shouldn't we somehow merge the sys.path?

jcfr: Thanks for the review. In this case,  there are no paths to be merged. The goal is to ensure the sys paths are not reset. We are saving the list of paths, updating argv and restoring.

lassoan: I would expect `_PyInterpreterState_SetConfig` to set up paths from environment variables, etc. Isn't it the case?
I'm also wondering if we could fix this more cleanly by running `_PyInterpreterState_SetConfig(&config)` earlier (before loading any modules).

jcfr: To further clarify, Python has already been initialized, the point of calling the PyConfig API here is to be able to simply update argv. Nothing more.

Now, I had a slightly more complex approach that was updating  `config.module_search_paths` based on the value of `sys.paths` but I concluded the added complexity was not worth it for the same outcome. (See pull notes in the pull request description)

lassoan: `_PyInterpreterState_SetConfig(&config)` documentation:

```
/* Set the configuration of the current interpreter.

   This function should be called during or just after the Python
   initialization.

   Update the sys module with the new configuration. If the sys module was
   modified directly after the Python initialization, these changes are lost.

   Some configuration like faulthandler or warnoptions can be updated in the
   configuration, but don't reconfigure Python (don't enable/disable
   faulthandler and don't reconfigure warnings filters).

   Return 0 on success. Raise an exception and return -1 on error.

   The configuration should come from _PyInterpreterState_GetConfigCopy(). */
```

If we don't call `_PyInterpreterState_SetConfig` during or just after the Python initialization then we risk introducing subtle errors. We noticed and fixed one error now. But if somebody enables/disables faulthandlers then things may break. It may also be a problem if a module parses argv during startup. New complications could come in each new Python version.

Python should be initialized from command-line arguments at about the same time when Qt application is initialized from command-line arguments (`-style`, `-stylesheet`, `-qwindowgeometry`, etc.), i.e., when command-line arguments are _parsed_. Not in the last minute when command-line arguments are _handled_. That is way too late to mess with the environment.

jcfr: Thanks for the feedback. Based on your suggestion, I switched to a **config-native** approach (see latest commit).

> "Python should be initialized from command-line arguments at about the same time when Qt is initialized"

I agree ... that’s the ideal solution. When I worked on fa915a03fc4 integrated via #8582, implementing this approach wasn’t trivial. I’ve opened a follow-up feature request with a concrete plan to do this cleanly:
* #8752

In the meantime, I’d like to **merge this PR** to fix the regression. I don’t have bandwidth right now to implement the optimal way :pray:

lassoan: Sounds good, thank you!

PR #8731, opened 2025-09-17
ENH: Add customizable application display name property

jamesobutler:
While using a custom slicer app (generated by the [SlicerCustomAppTemplate](https://github.com/KitwareMedical/SlicerCustomAppTemplate)), I observed that modifying the slicer-application-properties.cmake defined `APPLICATION_NAME` to something that contained a space character such as "3D Slicer" would result in a build error as the variable is reused in various programmatic contexts such as CMake target names which cannot have a space character. There can be times when using a space character is needed for specifying the full legal name of the application (e.g. "3D Slicer" while a shortened application name can be used for internal programmatic convenience (e.g. "Slicer").
https://github.com/Slicer/Slicer/blob/967015abbed1756cf3e7bf48b9a473528e72e94b/Applications/SlicerApp/slicer-application-properties.cmake#L2-L4

I observed that there was existing hardcoding of "3D Slicer" in the main application Window title likely because it couldn't use the application name property because that couldn't contain a space character.
https://github.com/Slicer/Slicer/blob/967015abbed1756cf3e7bf48b9a473528e72e94b/Applications/SlicerApp/qSlicerAppMainWindow.cxx#L149

This PR adds a new customizable application property called "APPLICATION_DISPLAY_NAME" which can contain a space character and is used in various user facing contexts such as Window Title, About Dialog, Install directory name, Desktop Shortcuts, StartMenu Shortcuts.

One special note is that [`setApplicationDisplayName`](https://doc.qt.io/archives/qt-5.15/qguiapplication.html#applicationDisplayName-prop) is not called explicitly because it results in the Window title always displaying in the form of "windowTitleText - ApplicationDisplayName" which prevents maintaining the existing Slicer main window title text of "3D Slicer VersionMajor.VersionMinor.VersionPatch". When this existing Qt `applicationDisplayName` property is not defined, it returns the `applicationName` as seen in the image below. For usage around the code, the new `mainApplicationDisplayName` property should be used.
{50C651A1-CDA5-4012-A3E3-F2290F3DACAB}

For Slicer custom apps that do not yet have an "APPLICATION_DISPLAY_NAME" entry in its slicer-application-properties.cmake file, the property will be automatically defined to the existing `mainApplicationName` resulting in the same output as it would be before this PR of changes. A PR has been made to the SlicerCustomAppTemplate to support defining the application display name as part of the template creation process. See https://github.com/KitwareMedical/SlicerCustomAppTemplate/pull/104.

Visible changes to 3D Slicer:
- "3D Slicer" instead of "Slicer" used in various text of the installer
- "3D Slicer" used in start menu shortcuts allowing to search by "3D.." to find the application. Previously typing "3D Slicer" in the Windows start menu would not find "Slicer" named applications because the "3D" part would be excluding it.

|`main` | This PR |
|-------|----------|
|image|image|
|image|image|
|image|image|
|{D6DE1E54-9A0B-4F45-98D8-884292AF58B2}|{04DCCB3F-7CB2-4896-91C8-20CABF9BCFE4}|
|{BC18B93D-3205-46F2-A5CA-C7EE2DADA96B}|{4E81240C-DC54-464F-9B02-4DF8D495F657}|


cpinter:
Looks quite nice! The code looks good as well. Thank you @jamesobutler !

jcfr (approved):

PR #8736, opened 2025-09-20
ENH: Improve custom sample data download

### LINKED ISSUE #8319: Sample Data modules Load from URL need to be revised to work with github data
On github a URL like https://github.com/SlicerMorph/terms-and-colors/blob/main/JaimiGrayTetrapodSkulls.csv

is automatically converted to  

https://raw.githubusercontent.com/SlicerMorph/terms-and-colors/refs/heads/main/JaimiGrayTetrapodSkulls.csv

when downloaded.

This causes problem with the Load from URL feature of the Sample Data. other tools like 3Dviewer has an internal logic that handles this:

https://github.com/kovacsv/Online3DViewer/blob/0d9998a656b84e2a9aebdef18a03759d0bc8df27/source/engine/io/fileutils.js#L102-L115

We should probably update the logic of Load From URL similarly, as this is a confusing behavior for user and would take time to explain...

@lassoan

lassoan:
Allow direct download from github.com and dropbox.com URLs. Allow specifying multiple download URLs at once. Do not alter filename when downloading from a custom URL.

fixes #8319

image

### DIFF
--- a/Modules/Scripted/SampleData/SampleData.py
+++ b/Modules/Scripted/SampleData/SampleData.py
@@ -245,6 +245,20 @@ class SampleDataSource:
 # SampleData widget
 #
 
+class TextEditFinishEventFilter(qt.QObject):
+
+    def __init__(self, *args, **kwargs):
+        qt.QObject.__init__(self, *args, **kwargs)
+        self.textEditFinishedCallback = None
+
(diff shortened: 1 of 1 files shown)


pieper (approved):
LGTM 👍

lassoan:
Thanks for the review!

PR #8737, opened 2025-09-20
ENH: Add volume reorientation to Crop Volume module

### LINKED ISSUE #8415: Add an option to more easily resample linear transforms in Crop Volume

Currently resampling volumes under linear transforms requires additional steps that are cumbersome and difficult to explain. 

Add a non-default option to crop volume that will calculate the ROI bounds without aligning them to the original image axes, and directly resample the image (interpolated cropping is a mandatory option for this). 

You can either manually draw the ROI or use the Volume Rendering to get a ROI big enough to match the new image extends. While they are workarounds, they are not natural part of the expected workflow. 

## Additional context

There is a detailed discussion on the forum showing the complexities: https://discourse.slicer.org/t/resampling-a-linear-transformation/42916/5?u=muratmaga

lassoan:
Orientation of the input volume can now be easily fixed in Crop Volume module using the new "Reorient volume" section.

Options are added to tune "Fit to volume" beheavior. It is now possible to align the cropped volume's ROI to the world coordinate system (or keep the current ROI orientation). This is useful if the user wants to align the cropped volume axes to the world coordinate system, for example to make segmentation in anatomical planes easier.

fixes #8415

image

@muratmaga

### DIFF
--- a/Docs/user_guide/modules/cropvolume.md
+++ b/Docs/user_guide/modules/cropvolume.md
@@ -22,7 +22,9 @@ Most frequently used for these scenarios:
 
 * **Registration of two objects that occupy smaller portions of the image**. In this scenario, cropping will allow you to focus the processing at the region of interest, and simplify registration initialization.
 
-* **Definition of new axis directions** for your image: [Interpolated cropping options](#interpolated-cropping-options) allows the output volume to have different axis directions that the original volume. The output volume's axis directions will be aligned with the ROI widget's axes. ROI widget axes can be rotated using the [Transforms](transforms.md) module.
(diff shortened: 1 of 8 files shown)


pieper (approved):
LGTM 👍 

Only question is if this should be merged after the release so it can get some more testing.  I"m fine if you are confident.

muratmaga:
> LGTM 👍
> 
> Only question is if this should be merged after the release so it can get some more testing. I"m fine if you are confident.

This is something we need for SlicerMorph and MorphoDepot, so it will be great if it can make it into release.

lassoan:
No change in behavior, just more options and extra features, so I don't think it is risky. There will also be at least a couple of days before the rease for a few iterations to test and refine.

I'll merge this to make it easy to try.

PR #8739, opened 2025-09-21
Make volume rendering transfer function adjustment easier

lassoan:
This PR adds interactive window/level and opacity adjustments for volume rendering in 3D views. It works the same way as slice views (activate window/level adjustment mouse mode and then click-and-drag in a view), but works for volumes. Up/down shifts the level, left/right shrinks/expands the window. If you hold down Ctrl then it changes the opacity of the scalar opacity transfer function. It works for even multiple volumes: you adjust the volume that is where you start the click-and-drag. 

https://github.com/user-attachments/assets/9b9fce7c-cc03-44a9-9151-b99c993b60a0

I’ve also added the option to enable/disable volume rendering following slice views in the subject hierarchy menu, but I’m not sure how useful/necessary is this now that the transfer functions are easily editable interactively, even without using Volumes module at all.

image

@muratmaga


muratmaga:
These look great. Hope they can make it into the release.

cpinter:
Thanks Andras, I think this is super useful! The slider does not really cut it anyway for the transfer function "shift", because many times the range is not properly set, and you cannot edit it in a meaningful way. These mouse gestures are always relative, so there is no range problem. I think the SH option is also useful, but you're right that after this change it became questionable. I think it's worth integrating now that you have it, and if we see that it is not too useful after all, we can always remov it...

pieper (approved):
LGTM 👍 

(sorry about the pieper023 review - I've been testing scripts to create and delete repos so I have some dummy accounts and forgot to switch (again)).

lassoan:
Thanks for the reviews!

muratmaga:
> I think the SH option is also useful, but you're right that after this change it became questionable

I think this is important for the visualizations for which one wants to drive 3D rendering solely within Volumes module (with the sync option enabled). The use cases would be putting a custom color LUT and a fixed controls on the W/L and thresholds etc.

jcfr (approved):

Issue #8526, opened 2025-07-03
Convert .txx files to .hxx to allow consistent handling by AUTO_MOC in QT6


hjmjohnson:
There are only a few files that retain the .txx extension.  Converting them to .hxx makes them more consistent with community standards, and can assist IDE and static analysis tools to properly treating these as c++ containing files.

This is motivated by the realization that:

```
CMake’s AUTOMOC mechanism scans files with standard header extensions (.h, .hpp, .hxx, etc.) for Q_OBJECT and automatically runs moc when needed.
However, .txx is not included in that default extension list, so AUTOMOC will ignore .txx files entirely. 
```

PR #8741, opened 2025-09-22
STYLE: Update to use .hxx instead of .txx for tooling

hjmjohnson:
Converting them to .hxx makes them more consistent with community
standards, and can assist IDE and static analysis tools to properly
treating these as c++ containing files.

This is motivated by the realization that:
  CMake’s AUTOMOC mechanism scans files with standard header extensions
  (.h, .hpp, .hxx, etc.) for Q_OBJECT and automatically runs moc when
  needed.  However, .txx is not included in that default extension list,
  so AUTOMOC will ignore .txx files entirely.

Increasingly tools are being created to help the programmer better
understand and analyze, and provide hints to proper coding options
dynamically in the development environment (color coding, auto-completion).

The obscure .txx file requires special configuration of each development
environment tool to indicate that it is really a c++ header.  For new
developers this is often a difficult and frustrating task.  This is
confirmed by the large amount of documentation for just a few of the
common editors.  Tools that recognize and respect the .hxx extension:
uncrustify, vim, emacs, VisualStudios, clang, SlickEdit,
kwrite, etc....)  None of these tools recognize the .txx extension without
end user customizations.

Resolves: #8526

```
find . -name "*.txx" > ../CHANGE_FILE_EXTENSION_LIST
for i in $(cat ../CHANGE_FILE_EXTENSION_LIST ); do
  rm ${i//.txx/.hxx}
  git mv $i ${i//.txx/.hxx}
done
for i in $(cat ../CHANGE_FILE_EXTENSION_LIST ); do
  fn=$(basename $i)
  for reff in  $(git grep -l ${fn}); do
    sed -i "s/${fn//\.txx/}_txx/${fn/\.txx/_hxx}/g" $reff
  done
  for reff in  $(git grep -l ${fn}); do
    sed -i "s/${fn//\.txx/}\.txx/${fn/\.txx/.hxx}/g" $reff
  done
done
```


.pre-commit-config.yaml
```diff
@@ -28,14 +28,9 @@ repos:
...
       - id: clang-format
-        # Explicitly include ".txx" files, which are not recognized as source files
-        # by the "identify" package used by pre-commit. To work around this,
-        # we manually specify file matching via the "files" regex and set "types_or"
-        # to allow formatting of files treated as plain text.
-        #
         # Reference: https://github.com/pre-commit/identify/blob/main/identify/extensions.py
         types_or: [file, text]
-        files: \.(cpp|cxx|h|hpp|txx)$
+        files: \.(cpp|cxx|h|hpp|hxx)$
```
lassoan: A number of Slicer extensions, VTK, ITK (yes, ITK - via GDCM), Plastimatch, etc. all use .txx, so I don't think we are quite ready to remove txx from here yet.

lassoan (commented):
Thanks for working on this. Although the benefits seem quite limited, the cost of change and risk of complications are even smaller, so I think it makes sense to make the proposed change.

.pre-commit-config.yaml
jamesobutler: Since this pre-commit-config is only for Slicer core code, this change would seem appropriate. I don't think Slicer extensions, VTK, ITK, etc are following the Slicer pre-commit-config exactly.

lassoan:
Please also add a not on this to the [Style guide](https://github.com/Slicer/Slicer/blob/main/Docs/developer_guide/style_guide.md).

When this merged then we'll have to add a check to the automated extension checks to warn if files with .txx extensions are found in the repository. (it is enough to submit an issue to https://github.com/Slicer/ExtensionsIndex, no need to implement it right away).

.pre-commit-config.yaml
lassoan: If an extension developer wanted to set up precommit hook then he would by default start from the Slicer core, unless we provide some other official template for extensions. It is harmless to keep txx in this file in addition to hxx.

pieper023:
I agree this is a good idea, but I'd lean toward waiting until the major version release.

hjmjohnson:
> Thanks for working on this. Although the benefits seem quite limited, the cost of change and risk of complications are even smaller, so I think it makes sense to make the proposed change.

The benefits will be more pronounced once Qt6 and cmake AUTOMOC tools are used for building.  AUTOMOC does not know of the txx extension by default.

.pre-commit-config.yaml
hjmjohnson: I re=added txx and hxx to the list.

jamesobutler: If Slicer core no longer uses txx files than it would make sense for the pre-commit-config to reflect that because then a Slicer extension developer would also make sure to match the Slicer core usage of txx instead of hxx if they are  matching the pre-commit-config.

hjmjohnson:
I added to a regex and note .gitignore to help prevent new .txx files from being added to the core repository.

hjmjohnson:
Issue created as requested:   https://github.com/Slicer/ExtensionsIndex/issues/2230

hjmjohnson:
@pieper023 as we approach working on Qt6 upgrades, this will make that work easier.

jamesobutler:
Just as a heads up, the upcoming stable release is planned to support a build option to build with Qt6 (Qt 5 will be default), so if any changes are necessary to support that effort we will want to integrate.

PR #8748, opened 2025-09-24
BUG: Do not reset markups control points when reading XML attributes

Punzo:
Measurement values are properly reset when reading markups data. Control points are preserved during XML attribute reading, maintaining existing markup geometry and preventing unintended data loss.

Ref: https://github.com/Slicer/Slicer/pull/8141#discussion_r2372552810

### DIFF
--- a/Libs/MRML/Core/vtkMRMLMarkupsFiducialStorageNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLMarkupsFiducialStorageNode.cxx
@@ -552,6 +552,7 @@ int vtkMRMLMarkupsFiducialStorageNode::ReadDataInternal(vtkMRMLNode* refNode)
     {
       // clear out the list
       markupsNode->RemoveAllControlPoints();
+      markupsNode->ClearValueForAllMeasurements();
     }
 
     std::string line;
--- a/Libs/MRML/Core/vtkMRMLMarkupsNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLMarkupsNode.cxx
@@ -133,9 +133,6 @@ void vtkMRMLMarkupsNode::ReadXMLAttributes(const char** atts)
 {
   MRMLNodeModifyBlocker blocker(this);
 
-  this->RemoveAllControlPoints();
-  this->ClearValueForAllMeasurements();
-
   Superclass::ReadXMLAttributes(atts);
 
   vtkMRMLReadXMLBeginMacro(atts);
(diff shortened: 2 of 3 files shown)


Punzo:
this fix the issue that when calling a markups node method `ReadJSON -> ReadXMLAttributes` in https://github.com/Slicer/Slicer/pull/8141, which clears the markups control points and it should not (control points should be cleared only when the method `ReadDataInternal` of the storage node is called.

Libs/MRML/Core/vtkMRMLMarkupsFiducialStorageNode.cxx
```diff
@@ -552,6 +552,7 @@ int vtkMRMLMarkupsFiducialStorageNode::ReadDataInternal(vtkMRMLNode* refNode)
     {
       // clear out the list
       markupsNode->RemoveAllControlPoints();
+      markupsNode->ClearValueForAllMeasurements();
```
jcfr, line 555: Should we also call this function from `vtkMRMLMarkupsJsonStorageNode::UpdateMarkupsNodeFromJsonValue` just after `RemoveAllControlPoints`?


https://github.com/Slicer/Slicer/blob/92c6f5dd3a4f7b05d826641fef3a9a22093a59ca/Modules/Loadable/Markups/MRML/vtkMRMLMarkupsJsonStorageNode.cxx#L365-L374

Libs/MRML/Core/vtkMRMLMarkupsNode.cxx
```diff
@@ -133,9 +133,6 @@ void vtkMRMLMarkupsNode::ReadXMLAttributes(const char** atts)
 {
   MRMLNodeModifyBlocker blocker(this);
 
-  this->RemoveAllControlPoints();
-  this->ClearValueForAllMeasurements();
```
jcfr, line 137: > [...] control points should be cleared only when the method `ReadDataInternal` of the storage node is called. [...] 

Since this applies to all markups (not only `vtkMRMLMarkupsFiducialStorageNode`), if we move forward with this change .. we will likely have to call the functions somewhere else.

Punzo: no this is fine, that is the point, it should be the storage nodes to clear all the control points and it should not be the base node. and this is already done. vtkMRMLMarkupsJsonStorageNode and vtkMRMLMarkupsJsonStorageNode clean already the control points before reading the data.

Libs/MRML/Core/vtkMRMLMarkupsFiducialStorageNode.cxx
Punzo: ah good catch Jc, I think yes in line 370 we should also call `ClearValueForAllMeasurements`. But let me double check if it is really needed, maybe they are already cleaned in another way. Today I am off, I will try to be back to this asap.

Punzo: yes indeed it was necessary. Removed the change from original PR https://github.com/Slicer/Slicer/pull/8141/commits/966c2363c07a7ecdd96ad1e689ed381a089c0de5

and I have updated this one.

Punzo:
@jcfr I have applied the review, let me know if I need to take any other action, thanks!

jcfr (approved):

Issue #7434, opened 2023-11-27
Improve vtkMRMLDisplayNode API to support showing object in all view but one.


jcfr:
## Is your feature request related to a problem? Please describe.

The current API presents a challenge when attempting to selectively hide an object in only one view. This limitation necessitates the repetitive re-implementation of logic within modules and custom applications.

There are two cases to address:

1. **Hide in All Views Except One**: Users often need the flexibility to hide an object in all views except a specific one.

2. **Hide in Only the Specified View**: Similarly, there is a need to hide an object exclusively in a designated view. 

### Case (1): Hide in All Views Except One

To facilitate this, a new function, `vtkMRMLDisplayNode::HideOnlyInView(const char *viewNodeID)`[^1], could be introduced. The proposed implementation is as follows:

[^1]: name based of the existing API `vtkMRMLDisplayNode::SetDisplayableOnlyInView(const char *viewNodeID)`).

```cpp
  // Get all viewNodeIDs
  this->GetNumberOfNodesByClass("vtkMRMLAbstractViewNode");
  std::vector<vtkMRMLNode *> viewNodes;
  this->GetScene()->GetNodesByClass("vtkMRMLAbstractViewNode", viewNodes);
  std::vector<std::string> viewNodeIDs;
  for(vtkMRMLNode* viewNode: viewNodes)
    {
    viewNodeIDs.push_back(viewNode->GetID());
    }

  // Hide only in the specified view
  this->SetViewNodeIDs(viewNodeIDs);
  this->RemoveViewNodeID(viewNodeID);
```

### Case (2): Hide in Only the Specified View

Addressing this case is more complex due to the `ViewNodeIDs` convention. We lack a mechanism to distinguish between:

- **Intentional Hiding**: A view from which an object was explicitly hidden.
- **Subsequent Views**: For instance, switching to `Dual 3D` may imply the object should be visible by default.

Further refinement in the API or additional conventions may be necessary to address this.

**ViewNodeIDs Convention**:
- An empty list implies visibility in all views.
- A non-empty list indicates visibility exclusively in the specified view.

## Describe alternatives you've considered

NA

## Additional context

* https://github.com/Slicer/Slicer/pull/7433#discussion_r1406442044

lassoan:
I agree, that the current single list of view node IDs cannot capture what the user wants (show only in the explicitly listed views; or hide in the non-listed views).

[...]

PR #8749, opened 2025-09-24
BUG: Clear existing node references before parsing new ones

Punzo:
Ref: https://github.com/Slicer/Slicer/pull/8141#discussion_r2372549906

### DIFF
--- a/Libs/MRML/Core/vtkMRMLNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLNode.cxx
@@ -474,6 +474,10 @@ void vtkMRMLNode::ParseReferencesAttribute(const char* attValue, std::set<std::s
   /// parse references in the form "role1:id1 id2;role2:id3;"
   std::string attribute(attValue);
 
+  // Parse the new references from the attribute string into a temporary structure
+  // roles -> list of (referenced_node_id, property_names -> property_values)
+  std::map<std::string, std::vector<std::pair<std::string, std::map<std::string, std::string>>>> newReferences;
+
   std::size_t start = 0;
   std::size_t end = attribute.find_first_of(';', start);
   std::size_t sep = attribute.find_first_of(':', start);
@@ -485,6 +489,10 @@ void vtkMRMLNode::ParseReferencesAttribute(const char* attValue, std::set<std::s
     {
       std::string ids = attribute.substr(sep + 1, end - sep - 1);
       std::stringstream ss(ids);
+      // Stores referenced node IDs and properties for the current role in a
+      // list of (referenced_node_id, property_names -> property_values)
+      std::vector<std::pair<std::string, std::map<std::string, std::string>>>& roleReferences = newReferences[role];
+
       while (!ss.eof())
       {
         std::string value;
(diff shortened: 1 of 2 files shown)


Punzo:
this fix the issue that when calling the layout node method `ReadJSON -> ReadXMLAttributes` in https://github.com/Slicer/Slicer/pull/8141, which does not clear previous maximized views

Libs/MRML/Core/vtkMRMLLayoutNode.cxx
```diff
@@ -85,6 +85,8 @@ void vtkMRMLLayoutNode::ReadXMLAttributes(const char** atts)
 {
   int disabledModify = this->StartModify();
 
+  this->RemoveAllMaximizedViewNodes();
```
jcfr, line 88:
```suggestion
  // Reset any previously populated maximized view nodes before reading from XML.
  // This prevents stale state from accumulating across scene loads or node re-use.
  this->RemoveAllMaximizedViewNodes();
```

jcfr: Is this a pattern that we should generalize to other nodes implementing a method calling `RemoveNodeReferenceIDs` ?


Here are such methods ...


```
void vtkMRMLClipNode::RemoveAllClippingNodeIDs()
{
  this->RemoveNodeReferenceIDs(this->GetClippingNodeReferenceRole());
}
```

```
void vtkMRMLDisplayableNode::RemoveAllDisplayNodeIDs()
{
  this->RemoveNodeReferenceIDs(this->GetDisplayNodeReferenceRole());
}
```

```
void vtkMRMLPlotChartNode::RemoveAllPlotSeriesNodeIDs()
{
  this->RemoveNodeReferenceIDs(this->GetPlotSeriesNodeReferenceRole());
}
```

cc: @lassoan

jcfr: If yes, we should add a generic method to `vtkMRMLNode` (e.g. `RemoveAllNodeReferencesForAllRoles` ) and call it from `ReadXMLAttributes`.

If not, let's make sure to add a comment to explain why this is needed only for the `MaximizedView` node IDs.

Punzo: thinking about this, I would need to work more on the PR. This because, when you call ReadJSON -> ReadXMLAttributes, it may be the full node state, but also only a partial update. If it is a partial update, you would call `RemoveAllMaximizedViewNodes` only if in the JSON there is the MaximizedViewNodes information. 

I would need to investigate as well for the reference IDs. Probably I would use the same appoarch, clean them only if the JSON string has their values. Thanks for pointing it out. I will be able to address this PR again, next week.

Punzo:
@jcfr updated this PR to have a proper fix. References are cleared by role if that specific role has any value in the XML/JSON. This ensuring that there are not mismatch between the scene and the JSON/XML descriptor.

I have fully summarized the modification in the commit message.

jcfr (dismissed):
Thanks @Punzo for revisiting the original patch 🙏 

As a bonus, I am wondering if there is a way to update the existing node reference tests to check for the fixed behavior ?

Libs/MRML/Core/vtkMRMLNode.cxx
```diff
@@ -483,6 +483,9 @@ void vtkMRMLNode::ParseReferencesAttribute(const char* attValue, std::set<std::s
     // Only process this role if it has not been encountered already
     if (references.find(role) == references.end())
     {
+      // Clear all existing references for this role before adding new ones
+      this->RemoveNodeReferenceIDs(role.c_str());
```
lassoan: This is very simple, but could cause a lot of noise: it would invoke node reference removed + node reference added events for all the unchanged node references. Instead, we should compare existing and new references and only apply the difference.

jcfr: This made me think of API suggested in the context of:
* https://github.com/Slicer/Slicer/issues/7434

lassoan:
@Punzo please have a look at my proposed change that eliminates all unnecessary changes (e.g., removing and then re-adding node references, and node reference properties if they are not changed).

Checking if something is actually changed takes a little extra computation, but if there are no changes then the checks should be really quick. This known little extra cost is worth it because it avoids making temporary changes that can trigger operations in various modules with an unknown (potentially very high) cost.

Punzo:
> @Punzo please have a look at my proposed change that eliminates all unnecessary changes (e.g., removing and then re-adding node references, and node reference properties if they are not changed).
> 
> Checking if something is actually changed takes a little extra computation, but if there are no changes then the checks should be really quick. This known little extra cost is worth it because it avoids making temporary changes that can trigger operations in various modules with an unknown (potentially very high) cost.


Ok thanks! tomorrow morning I will have a look, and I will test with the JSON infrastructure.

Libs/MRML/Core/vtkMRMLNode.cxx
```diff
@@ -506,41 +514,140 @@ void vtkMRMLNode::ParseReferencesAttribute(const char* attValue, std::set<std::s
...
+        }
+
+        // If this node ID is no longer needed, remove the reference
+        if (!nodeIdStillNeeded)
+        {
+          this->RemoveNthNodeReferenceID(role.c_str(), currentIndex);
+        }
+      }
+    }
+
```
jcfr, line 597:
```suggestion

    // Re-fetch after removals to avoid stale indice
    currentNodeIDs.clear();
    this->GetNodeReferenceIDs(role.c_str(), currentNodeIDs);
```

jcfr: @lassoan 
Since we call `RemoveNthNodeReferenceID`,should we get the updated list IDs ?

Punzo:
> Ok thanks! tomorrow morning I will have a look, and I will test with the JSON infrastructure.

@lassoan @jcfr 

I tested with scalar, layout and clipping nodes and everything seems working on my side with the JSON reading/writing.

Libs/MRML/Core/vtkMRMLNode.cxx
lassoan: Good point, I'll add this.

PR #8757, opened 2025-09-28
ENH: Add support for storing modeling displacement field transforms

lassoan:
Displacement field transform can be defined using a vector field that transforms from the target coordinate system to the source object (a.k.a. resampling transform, because it can be used to transform images by resampling; "from parent" in Slicer naming convention), or that transforms form the source object coordinate system to the target (a.k.a. modeling transform, because it can be used to transform geometric objects by moving their points; "to parent" in Slicer naming convention).

We use ITK readers/writers for storing of transforms in files, and since ITK is an image-oriented toolkit, it always stores resampling transforms in files. Slicer worked around this limitation by introducing special "inverse" transform classes - which specify that those transforms need to be inverted before using them as resampling transform.

However, when displacement field transforms were stored as images. There was no way to specify in the image files that the displacement field should be used for resampling or modeling.

This commit adds the possibility to store displacement fields in NRRD files as either resampling and modeling transform, by introducing a new optional custom field: "displacement field type". This field can be set to "resampling" (default) or "modeling". For other file formats (NIFTI, MetaImage, ...) it is still assumed that the displacement field should be used for resampling.

### DIFF
--- a/Docs/developer_guide/modules/transforms.md
+++ b/Docs/developer_guide/modules/transforms.md
@@ -17,6 +17,70 @@
   - Avoid inverse non-linear transforms: make sure that non-linear transforms are only set as FromParent
 - Transforms module in Slicer shows linear transform matrix values "to parent" (modeling convention) in RAS coordinate system. Therefore to retrieve the same values from an ITK transforms as shown in Slicer GUI, one has switch between RAS/LPS and modeling/resampling. See example [here](../script_repository.md#convert-between-itk-and-slicer-linear-transforms).
 
+## Displacement field files
+
+Slicer can store displacement field transforms (or "grid transforms" in VTK naming convention) in 3D NRRD or NIFTI format; or sequence of transforms in NRRD format. Image file header fields are used to distinguish the displacement field transforms form regular images and specify how to correctly interpret these images as transforms.
+
+Requirements for NRRD files to be correctly recognized and loaded as displacement field transform:
+- `kinds` field is set to:
+  - `vector domain domain domain` for single 3D displacement field transform
+  - `vector domain domain domain list` for displacement field sequence
+- `measurement frame` is recommended to be set to identity to unambiguously specify that vector stores spatial coordinates in the same coordinate system as the image axes are specified in.
+- custom `intent_code` field is set to `1006`. This convention originates from the NIFTI file format and it makes it easy to detect images as displacement fields when using ITK file reader.
+- custom `displacement field type` is set to:
+  - `resampling` for ITK convention, useful for transforming images, in the Slicer transform tree specifies the transform "from parent"
+  - `modeling` (useful for transforming points, markups, models; in the Slicer transform tree specifies the transform "to parent")
+
+Requirements for NIFTI files to be correctly recognized and loaded as displacement field transform:
+- `intent_code` field is set to 1006 (`NIFTI_INTENT_DISPVECT`)
+- displacement vector values are stored in RAS coordinate system
+
+### Example 3D displacement field file header
+
+```txt
+NRRD0005
+# Complete NRRD file format specification at:
+# http://teem.sourceforge.net/nrrd/format.html
+type: double
+dimension: 4
+space: left-posterior-superior
+sizes: 3 256 256 130
+space directions: none (0,1,0) (0,0,-1) (-1.2999954223632812,0,0)
+kinds: vector domain domain domain
+endian: little
+encoding: raw
+space origin: (86.644897460937486,-133.92860412597656,116.78569793701172)
+measurement frame: (1,0,0) (0,1,0) (0,0,1)
+displacement field type:=resampling
+intent_code:=1006
+```
+
+### Example 3D+t displacement field sequence file header
+
+```txt
+NRRD0005
+# Complete NRRD file format specification at:
+# http://teem.sourceforge.net/nrrd/format.html
+type: double
+dimension: 5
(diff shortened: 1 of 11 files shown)


Docs/developer_guide/modules/transforms.md
```diff
@@ -17,6 +17,24 @@
   - Avoid inverse non-linear transforms: make sure that non-linear transforms are only set as FromParent
 - Transforms module in Slicer shows linear transform matrix values "to parent" (modeling convention) in RAS coordinate system. Therefore to retrieve the same values from an ITK transforms as shown in Slicer GUI, one has switch between RAS/LPS and modeling/resampling. See example [here](../script_repository.md#convert-between-itk-and-slicer-linear-transforms).
 
+## Displacement field files
+
+Slicer can store displacement field transforms (or "grid transforms" in VTK naming convention) in 3D NRRD or NIFTI format; or sequence of transforms in NRRD format. Image file header fields are used to distinguish the displacement field transforms form regular images and specify how to correctly interpret these images as transforms.
+
+Requirements for NRRD files to be correctly recognized and loaded as displacement field transform:
+- `kinds` field is set to:
+  - `vector domain domain domain list` for single 3D displacement field transform
```
pieper: Should this be `vector domain domain domain`?

Docs/developer_guide/modules/transforms.md
```diff
@@ -17,6 +17,24 @@
...
+- `measurement frame` is set and it defines the coordinate system the displacement vectors are defined in.
+- custom `intent_code` field is set to `1006`. This convention originates from the NIFTI file format and it makes it easy to detect images as displacement fields when using ITK file reader.
+- custom `displacement field type` is set to:
+  - `resampling` for ITK convention, useful for transforming images, in the Slicer transform tree specifies the transform "from parent"
+  - `modeling` (useful for transforming points, markups, models; in the Slicer transform tree specifies the transform "to parent")
+
+Requirements for NIFTI files to be correctly recognized and loaded as displacement field transform:
+- `intent_code` field is set to 1006 (`NIFTI_INTENT_DISPVECT`)
+- displacement vector values are stored in RAS coordinate system
+
```
pieper, line 37: It would be nice to include a specific example of each type of nrrd header just to be concrete.

pieper (dismissed):
This looks great, and much needed!

I don't know if you saw, but Gordon recently made a new release of teem (https://sourceforge.net/p/teem/mailman/teem-users/?viewmonth=202509).  I'm thinking we should propose this `displacement field type` to him.

Docs/developer_guide/modules/transforms.md
lassoan: Correct. I'll fix.

lassoan:
```suggestion
  - `vector domain domain domain` for single 3D displacement field transform
```

Libs/MRML/Core/vtkMRMLTransformSequenceStorageNode.cxx
```diff
@@ -521,8 +600,8 @@ int vtkMRMLTransformSequenceStorageNode::SupportedFileType(const char* fileName)
   }
 
   // Read the first 800 characters of the file
-  std::string content(800, '\0');
-  file.read(&content[0], 800);
+  std::string content(1500, '\0');
```
Thibault-Pelletier: Comment above is outdated

Docs/developer_guide/modules/transforms.md
```diff
@@ -17,6 +17,70 @@
...
+Slicer can store displacement field transforms (or "grid transforms" in VTK naming convention) in 3D NRRD or NIFTI format; or sequence of transforms in NRRD format. Image file header fields are used to distinguish the displacement field transforms form regular images and specify how to correctly interpret these images as transforms.
+
+Requirements for NRRD files to be correctly recognized and loaded as displacement field transform:
+- `kinds` field is set to:
+  - `vector domain domain domain` for single 3D displacement field transform
+  - `vector domain domain domain list` for displacement field sequence
+- `measurement frame` is recommended to be set to identity to unambiguously specify that vector stores spatial coordinates in the same coordinate system as the image axes are specified in.
+- custom `intent_code` field is set to `1006`. This convention originates from the NIFTI file format and it makes it easy to detect images as displacement fields when using ITK file reader.
+- custom `displacement field type` is set to:
+  - `resampling` for ITK convention, useful for transforming images, in the Slicer transform tree specifies the transform "from parent"
```
Thibault-Pelletier, line 31: Why not using directly "from parent" and "to parent" in the file format?
Anyone knowing transforms would know what is the expected behavior of the displacement field this way.

pieper: I think modeling and resampling are probably clearer and more standard.  I remember working on the first versions of all this with @millerjv and I would have sworn that we intended the term "from parent" to refer to a modeling transform because it expresses moving the object with respect to it's parent and "to parent" to refer to a resampling transform because for any point in child space it tells you where it maps in parent space.  But apparently the implementation is the reverse of this, so now I find the to/from parent language confusing.

Libs/MRML/Core/vtkMRMLTransformSequenceStorageNode.cxx
jcfr: For future reference, approach was revisited to  look for `intent_code:=1006`

jcfr (approved):

PR #8763, opened 2025-10-02
ENH: Expose segment editor effect cursor functions

ThomasKierski:
## PR Summary

This PR adds the `Q_INVOKABLE` macro to `qSlicerSegmentEditorAbstractEffect::createCursor()` and adds a utility function for applying the cursor associated with a given segment editor effect to the active view widgets. 

The motivation for this PR is to address an issue affecting the [local threshold tool in the extra effects repository](https://github.com/lassoan/SlicerSegmentEditorExtraEffects/blob/master/SegmentEditorLocalThreshold/SegmentEditorLocalThreshold.py) which causes the effect-specific cursor to revert to the default arrow-pointer cursor when a user adds an ROI node from the tool's options panel. The utility function provided in this PR provides a route to restore the effect cursor.

## Issue (cursor reverts to default when user adds an ROI)

https://github.com/user-attachments/assets/1b6954a7-9420-4f22-8935-160a0b1859fe

## Example with solution integrated (cursor remains the same)

https://github.com/user-attachments/assets/38f19cb0-983c-4778-86cf-431beb34f8b9

### DIFF
--- a/Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.cxx
+++ b/Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.cxx
@@ -3093,3 +3093,9 @@ void qMRMLSegmentEditorWidget::resumeRender()
 {
   qSlicerApplication::application()->resumeRender();
 }
+
+void qMRMLSegmentEditorWidget::applySegmentEffectCursor(qSlicerSegmentEditorAbstractEffect* effect)
+{
+  Q_D(qMRMLSegmentEditorWidget);
+  d->setEffectCursor(effect);
+}
--- a/Modules/Loadable/Segmentations/EditorEffects/qSlicerSegmentEditorAbstractEffect.h
+++ b/Modules/Loadable/Segmentations/EditorEffects/qSlicerSegmentEditorAbstractEffect.h
@@ -169,7 +169,7 @@ public:
   virtual void setupOptionsFrame() {};
 
   /// Create a cursor customized for the given effect, potentially for each view
-  virtual QCursor createCursor(qMRMLWidget* viewWidget);
+  Q_INVOKABLE virtual QCursor createCursor(qMRMLWidget* viewWidget);
 
   /// Callback function invoked when interaction happens
   /// \param callerInteractor Interactor object that was observed to catch the event
--- a/Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.h
+++ b/Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.h
@@ -276,6 +276,9 @@ public:
   /// Returns true if automatic jump to current segment is enabled.
   bool jumpToSelectedSegmentEnabled() const;
 
+  /// Update the current cursor to incorporate the icon of a segment editor effect tool
+  Q_INVOKABLE void applySegmentEffectCursor(qSlicerSegmentEditorAbstractEffect* effect);
+
 public slots:
   /// Set the MRML \a scene associated with the widget
   void setMRMLScene(vtkMRMLScene* newScene) override;


Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.cxx
```diff
@@ -3093,3 +3093,9 @@ void qMRMLSegmentEditorWidget::resumeRender()
 {
   qSlicerApplication::application()->resumeRender();
 }
+
+void qMRMLSegmentEditorWidget::applySegmentEffectCursor(qSlicerSegmentEditorAbstractEffect* effect)
```
cpinter: Missing separator line between functions.

cpinter:
Thanks for working on this! I only see this new function used in https://github.com/lassoan/SlicerSegmentEditorExtraEffects/pull/65/files
Is there any other place where `applySegmentEffectCursor` could be useful? To me it seems an overkill to extend the API for this one case (this infrastructure is being used since 2017 and until then it was not needed), but if there is no other way to fix the bug, I'm OK with it.
@lassoan what do you think?

jamesobutler:
@lassoan Yeah this would generally be used for any other segment editor effects where there is both interactions in the slice view where the segment editor effect cursor icon is being used, but also has Markups interaction as part of the effect which would change the cursor icon to the markups place mode icon. As more markups objects are used while maintaining the active segment editor effect, this situation trying to be handled will become more common.  

A side conversation that can be handled in a future written up issue or PR would be whether a segment editor effect cursor icon should be integrated with the concept of interaction Mouse Modes so that there isn't conflict between the two. Where mouse modes also define a cursor icon type. Currently the "Place" mouse mode is Markups specific, but it could be considered whether being in say the "Draw" segment editor effect is also a "Place" mode, or some other newly named mouse mode.

lassoan:
The proposed changes look good to me.

In the mid/long term, segment editor effects will use VTK widgets for user interactions in views (development is driven by the KitwareEU/trame team) and then direct cursor setting will probably not be necessary anymore.

jamesobutler (dismissed):

ThomasKierski:
> This should be ready to go, just one small thing could be improved. Exposing a private function by a different name would be an unnecessary extra indirection. Could you please instead move `setEffectCursor` method from the private implementation class to a public method in the `qMRMLSegmentEditorWidget` class? Thank you!

This has been updated. @lassoan, do you have a good rule of thumb for what should be implemented in the private class vs the public class so that I can keep it in mind moving forward?

lassoan (dismissed):
Thank you it looks good to me.

Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.cxx
```diff
@@ -3093,3 +3023,70 @@ void qMRMLSegmentEditorWidget::resumeRender()
...
+    {
+      threeDWidget->threeDView()->setViewCursor(effect->createCursor(threeDWidget));
+      threeDWidget->threeDView()->setDefaultViewCursor(effect->createCursor(threeDWidget));
+      this->CustomCursorInViewNodeIDs.insert(viewNodeID);
+    }
+    else
+    {
+      threeDWidget->threeDView()->setViewCursor(QCursor());
+      threeDWidget->threeDView()->setDefaultViewCursor(QCursor());
+      this->CustomCursorInViewNodeIDs.remove(viewNodeID);
```
jamesobutler: Some relevant build failures as reported by the CI
`/usr/src/Slicer/Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.cxx:3089:13: error: ‘class qMRMLSegmentEditorWidget’ has no member named ‘CustomCursorInViewNodeIDs’`

ThomasKierski:
Rebuilt and tested on my local machine to ensure the feature works as expected after moving the implementation away from the private class, everything seems fine on my end after the latest push

lassoan:
> do you have a good rule of thumb for what should be implemented in the private class vs the public class so that I can keep it in mind moving forward?

If too much is exposed through the public API, the interface becomes complex and fragile, making it difficult for Slicer core developers to evolve the software without breaking backward compatibility. If too little is exposed, extension developers struggle because they don't have access to the necessary features and low-level details. It is generally safer to expose less, as adding new public methods later is easy, while removing or changing existing ones is complex and may be disruptive.

ThomasKierski:
> > do you have a good rule of thumb for what should be implemented in the private class vs the public class so that I can keep it in mind moving forward?
> 
> If too much is exposed through the public API, the interface becomes complex and fragile, making it difficult for Slicer core developers to evolve the software without breaking backward compatibility. If too little is exposed, extension developers struggle because they don't have access to the necessary features and low-level details. It is generally safer to expose less, as adding new public methods later is easy, while removing or changing existing ones is complex and may be disruptive.

Thank you for the explanation!

PR #8765, opened 2025-10-03
ENH: Propagate terminology when exporting segments to models

cpinter:


cpinter:
@lassoan This is a very small and simple change (sets the terminology for models when we export segments). If you agree with its usefulness please approve, otherwise happy to discuss. Thank you!!

lassoan (approved):
Thank you, this is useful and looks good to me, but for consistent behavior it could make sense to copy `NameAutoGenerated` and `ColorAutoGenerated` information, too.

cpinter:
Thank you, makes sense.

cpinter:
@lassoan I wrote yesterday via Teams but here it is in case you see this sooner:

I am not sure we have a `NameAutoGenerated` concept for model nodes already established. We have it for terminology entries, and for color nodes we use the ColorDefined flag (`bool colorAndNameAutoGenerated = !colorNode->GetColorDefined(colorIndex);`), but there doesn't seem to exist attributes for other nodes like models. Did you have anything in mind for copying the Name/ColorAutoGenerated information? Thanks!

lassoan:
Color and name auto-generated flags are already stored for models, too:

image


So, all we need to do is to set not just `Terminology.TerminologyEntry` attribute of the model node but also `Terminology.AutoUpdateNodeColor` and `Terminology.AutoUpdateNodeName` attributes.

lassoan:
These flags are set here:

https://github.com/Slicer/Slicer/blob/d7429c745f2f14cdde96b7aaaface1cb6585f0a9/Modules/Loadable/SubjectHierarchy/Widgets/qSlicerSubjectHierarchyAbstractPlugin.cxx#L378-L387

To make it easier to read/write these flags elsewhere (and to make these flags more discoverable in general), it could make sense to add convenience methods to the terminology logic that would call these functions. These could be additional methods like `GetNameAutoGenerated` / `SetNameAutoGenerated` and/or a new terminology entry getting/setting variant such as `SetTerminologyEntryAsString(vtkMRMLNode* node, std::string entryStr, bool nameAutoGenerated, bool colorAutoGenerated)` method.

cpinter:
Thanks a lot! Strange that I didn't find it. I'll fix this then.

cpinter:
I updated the commit with the `SetTerminologyEntryAsString(vtkMRMLNode* node, std::string entryStr, bool nameAutoGenerated, bool colorAutoGenerated)` function. I decided to call the `SetTerminologyEntryAsString(vtkMRMLNode* node, std::string entryStr)`, from this new one, and change the original function so that when we set an empty entry, it clears these flags as well. It seemed to make sense, because the flags have no meaning if we don't have terminology.

cpinter:
@lassoan Do you think this is ready to go? Thanks!

Modules/Loadable/Segmentations/Logic/vtkSlicerSegmentationsModuleLogic.cxx
```diff
@@ -842,6 +842,13 @@ bool vtkSlicerSegmentationsModuleLogic::ExportSegmentToRepresentationNode(vtkSeg
       modelNode->SetAndObserveTransformNodeID(nullptr);
     }
 
+    // Set terminology of segment as attribute of model node
+    if (!segment->GetTerminology().empty())
+    {
+      // Name auto-generated is set to false, and code auto-generated is set to true, similarly to when we set terminology on displayable nodes
+      vtkSlicerTerminologiesModuleLogic::SetTerminologyEntryAsString(modelNode, segment->GetTerminology(), false, true);
```
lassoan: I've just noticed these hardcoded `false, true` values. Why don't we simply copy these flags from the segment?

```suggestion
      vtkSlicerTerminologiesModuleLogic::SetTerminologyEntryAsString(modelNode, segment->GetTerminology(), segment->GetNameAutoGenerated(), segment->GetColorAutoGenerated());
```

cpinter: Because when we set color manually to models (or other nodes), these flags are explicitly used. As I understood, this means we will not change the model node's name when we choose another term, but the color always follows the terminology. I can change it if you want

lassoan: To me it seems that it is valuable to preserve the flags (not lose this information) when the segment is exported. I've updated the code accordingly and merged. We can tune this behavior later if users find it unintuitive.

PR #8768, opened 2025-10-06
BUG: Fixed cloning of composite transforms

### LINKED ISSUE #8766: Cloning a composite transform yields incorrect result
## Summary
Cloning a composite transform yields an incorrect and non-equivalent clone.  See forum discussion at [https://discourse.slicer.org/t/cloning-a-composite-transform-yields-incorrect-result/44654](https://discourse.slicer.org/t/cloning-a-composite-transform-yields-incorrect-result/44654)

## Steps to reproduce

An example composite transform which does not clone correctly is available here: [Google Drive link to .h5 file](https://drive.google.com/file/d/1AH9RWsc5dIsRNxah0p4COLteh7VvCZs_/view?usp=drive_link)
1. Add a composite transform to the Slicer scene, such as the one linked above.
2. Load MRHead sample data image volume
3. Clone the transform (e.g. by right-clicking on it in the subject hierarchy and clicking “Clone”)
4. Alternately soft-apply the original and the cloned transformation to MRHead. 

Expected result: The result of applying the original transformation should be identical to the result of applying the cloned transformation.
Actual result: Alternately applying the two transformations produce visibly different results.

The example composite transform linked above was generated by an ANTs registration using the "General Registration (ANTs)" module.

On the forum I was asked to create a self-contained script which demonstrates the issue.  I have prepared 2 versions of this. The first, the simpler version, just registers MRHead to itself with no deformation, but using a nonlinear registration, so the result is a composite transform. When you clone that transform, you can see that the internal structure of the transform is different between the clone and the original.  However, since there isn't really any deformation in either one, the original and the clone don't produce visibly different results. Therefore, I also created a more complex version of a self-contained script.  The second version first creates a thin plate spline transform based on deforming a 3x3x3 grid by moving just the central point of the grid. Then this thin plate spline transform is applied to a clone of MRHead to create a deformed version of MRHead. This deformed volume is then also slightly rotated and translated.  Now that we have a more complex true transformation between the two images, we perform a nonlinear registration using ANTs, and receive a composite transform as output.  Cloning this composite transform also creates an incorrect copy, this time with visibly different results when alternately applying the clone and the original to MRHead.  
The two self-contained script files are in [this gist](https://gist.github.com/mikebind/66b023f14ac95dc556a0684d1630e524)

As I noted in the forum post, I suspect the bug is in the DeepCopyTransform function in vtkMRMLTransformNode.cxx, where I wonder if transforms defined by their inverses (where ToParent transform is calculated from FromParent transform) may not be being handled correctly. Maybe concatenations are being Pre-Multiplied when they should be Post-Multiplied (or vice versa) or maybe something isn't being inverted when it should be? I can't read the C++ well enough to understand where exactly it might be going wrong, but hopefully this is enough to help someone rapidly narrow down the problem and find the fix. 

## Environment
- Slicer version: Slicer-5.8.1

lassoan:
Order of composite transform components were reversed when deep-copying a transform node. This error has been in the code for several years, but the error was just recently reported probably because composite transforms are not used very frequently.

fixes #8766

### DIFF
--- a/Libs/MRML/Core/Testing/vtkMRMLNonlinearTransformNodeTest1.cxx
+++ b/Libs/MRML/Core/Testing/vtkMRMLNonlinearTransformNodeTest1.cxx
@@ -33,8 +33,8 @@
 int TestBSplineTransform(const char* filename);
 int TestGridTransform(const char* filename);
 int TestThinPlateSplineTransform(const char* filename);
-int TestCompositeTransformHardenSplit(const char* filename);
-int TestBSplineLinearCompositeTransformSplit(const char* filename);
+int TestCompositeTransformHardenCopySplit(const char* filename);
+int TestBSplineLinearCompositeTransformCopySplit(const char* filename);
 int TestRelativeTransforms(const char* filename);
 int TestGetTransform();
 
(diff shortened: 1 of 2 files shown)


cpinter:
This looks great, thank you!

PR #8771, opened 2025-10-06
BUG: Make mrb file loading more robust

lassoan:
Sequences that don't need data nodes can be saved as `.seq.mrml` files since the recent commit "PERF: Improve sequence node storage performance for simple nodes" (cd151177ac8e27ae61e619b657e5bcc992fa9070). However, these `.mrml` files can be mistaken for the main scene file when loading a scene from an `.mrb` file, causing the scene loading to silently fail (loading only a single sequence instead of all the nodes).

This commit makes finding the main scene `.mrml` file more robust: the `.mrml` file that is the closest to the root of the zip archive will be used as the main scene file.

Thanks to @che85 for reporting the issue.


pieper (approved):
LGTM 👍

PR #8775, opened 2025-10-07
BUG: childwidgetvariables do not return an object with an empty attribute

### LINKED ISSUE #8774: An object with an empty attribute name is returned from childWidgetVariables method
## Summary

`slicer.util.childWidgetVariables` method returns an object with an empty attribute name along all the other attribute names. This result in an empty entry in the python console completion and an empty string when calling `dir` on the returned object.

## Steps to reproduce

- Open Slicer
- Enter `"" in slicer.util.childWidgetVariables(slicer.modules.colors.widgetRepresentation()).__dir__()` in Python console -> Slicer print `True` => ERROR: Slicer should print `False` instead
- Write `slicer.util.childWidgetVariables(slicer.modules.colors.widgetRepresentation()).` and hit tab key to display the autocomplete list. -> Slicer display an empty entry (between `verticalLayout` and `AddColorTableNodeButton` entries) => ERROR: Slicer should not display an empty entry as it is not a valid choice in this context.

## Environment
- Slicer version: Slicer-5.9.0-2025-10-02
- Operating system: Linux Ubuntu 22.04

malbi:
fixes #8774

### DIFF
--- a/Base/Python/slicer/util.py
+++ b/Base/Python/slicer/util.py
@@ -391,7 +391,7 @@ def childWidgetVariables(widget):
     ui = type("", (), {})()  # empty object
     childWidgets = findChildren(widget)
     for childWidget in childWidgets:
-        if hasattr(childWidget, "name"):
+        if hasattr(childWidget, "name") and childWidget.name:
             setattr(ui, childWidget.name, childWidget)
     return ui
 
--- a/Applications/SlicerApp/Testing/Python/CMakeLists.txt
+++ b/Applications/SlicerApp/Testing/Python/CMakeLists.txt
@@ -314,6 +314,12 @@ slicer_add_python_unittest(
   )
 
 slicer_add_python_unittest(
+  SCRIPT ${Slicer_SOURCE_DIR}/Base/Python/slicer/tests/test_slicer_util_childWidgetVariables.py
+  SLICER_ARGS --no-main-window --disable-modules
+  TESTNAME_PREFIX nomainwindow_
+  )
+
+slicer_add_python_unittest(
   SCRIPT ${Slicer_SOURCE_DIR}/Base/Python/slicer/tests/test_slicer_parameter_node_wrapper.py
   SLICER_ARGS --no-main-window --disable-cli-modules --disable-scripted-loadable-modules
   TESTNAME_PREFIX nomainwindow_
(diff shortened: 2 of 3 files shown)


Base/Python/slicer/util.py
```diff
@@ -391,7 +391,7 @@ def childWidgetVariables(widget):
     ui = type("", (), {})()  # empty object
     childWidgets = findChildren(widget)
     for childWidget in childWidgets:
-        if hasattr(childWidget, "name"):
+        if hasattr(childWidget, "name") and childWidget.name:
```
pieper, line 394: Maybe it would make more sense to give them an autogenerated name instead (i.e. based on the class and/or the parents).

malbi: I did not choose this option because the method description states that it set an attribute on the returned object for each **named** child widget. Also, I encounter this issue while using the autocompletion in the python console to access an element with a known name and could not think of a use case where I would access an unnamed child via object attribute. One can still use `findChildren` method to access all the children, named or not of a given widget.

Also on the technical aspect, I only know of `GenerateUniqueName` from `vtkMRMLScene` which store used index and I am not sure if a simpler implementation could do the trick like:
```python
    ui = type("", (), {})()  # empty object
    childWidgets = findChildren(widget)
    for childWidget in childWidgets:
        if hasattr(childWidget, "name"):
            uniqueName = hildWidget.name
            if not childWidget.name:
                baseName = type(childWidget).__name__
                uniqueName = baseName
                i = 0
                while hasattr(ui, uniqueName):
                    uniqueName = baseName + "_" + str(i)
                    i = i + 1
            setattr(ui, uniqueName, childWidget)
    return ui
```

pieper: Okay, sure, that makes sense.  I was thinking of just numbering as they appear in order in their parents but it would be a very fragile thing and only useful for interactive use in the console.  For anything that really matters the widgets should have unique names so they can be found reliably.

lassoan: For getting all child widgets, developers can use `slicer.util.findChildren`.

`slicer.util.childWidgetVariables` function is added for convenient access of widgets based on name, so if a widget has empty name then it makes sense to just ignore it.

PR #8783, opened 2025-10-09
ENH: update_translations script differentiates between tr() and translate() as an alias

xskere:
If in a python file, instead of using the `tr()` function imported as `_()` you do so with the `translate()` function the script will not check if there is already a context given as a parameter and treats it the same as if it were the `tr()` function. 

This means that if you have this in code:

```
from slicer.i18n import translate as _

string = _("Context", "TextToBeTranslated")
```

The current script would substitute the `_("Context", "TextToBeTranslated")` with 
`_("InferredContext", "Context", "TextToBeTranslated")`.

And then, when lupdate is executed the .ts file output won't be as desired.

The updated script in this pull request checks for each source code file whether it's the `tr()` function or the `translate()` function that's being imported with the `_()` alias and based on that adds the context or not. In case of the ReGex not picking up which of the two is being used, it all is treated as if the function being used is `tr()` and always adds the context, same as before this change.

This change would allow free use of either the `tr()` or the `translate()` functions while using the `_()` alias, allowing for more diverse uses in which you still want to use the alias but want to specify the context yourself and not have it be inferred.

### DIFF
--- a/Utilities/Scripts/update_translations.py
+++ b/Utilities/Scripts/update_translations.py
@@ -24,6 +24,19 @@ import shutil
 import subprocess
 import sys
 
+def detect_tr_function_name(source_code_bytes):
+    """
+    Detect if the file uses tr() or translate() as translation function with _() alias.
+    Returns 'tr' if tr is found, 'translate' if translate is found, otherwise 'tr'.
+    """
+    source_code = source_code_bytes.decode(errors="ignore")
+    # Look for function import of tr
+    if re.search(r'(^|\n)\s*from\s+.+\s+import\s+tr(\s)as _', source_code):
+        return "tr"
+    # Look for function import of translate
+    if re.search(r'(^|\n)\s*from\s+.+\s+import\s+translate(\s)as _', source_code):
+        return "translate"
+    return "tr"
 
 def get_python_context(source_file):
     if os.path.isfile(source_file):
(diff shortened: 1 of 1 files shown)


lassoan:
If you want to specify context (don't want the translation infrastructure to automatically determine it) then you can directly use `translate`, as it is done at many places:

https://github.com/Slicer/Slicer/blob/997be79420de826d5250cbe934d7887f52dc266b/Modules/Scripted/Endoscopy/Endoscopy.py#L33

Would this fulfill your needs?

xskere:
That would work. However, in cases in which only `translate()` would be used, it would not be possible to use it with the `_()` alias.

This pull request was more about allowing both functions to be used with the script while also using the alias than about allowing the addition of context.

This change in the script would change nothing in how it's used, but in cases where `translate()` is imported as `_()` it would simply detect and not add context whenever `_()` is used.

cpinter:
Is there an argument against this proposal? It would be very convenient being able to use the `_` alias from classes that allow auto-discovery of the context as well from class that don't. Thank you!

cpinter:
@lassoan Bringing the above question to your attention. Thanks!

xskere:
Thanks to @lassoan helping us on our problem during project week we managed to find a solution on our end and don't need this change anymore. However, I will explain below our solution, to leave some context in case someone in the future has the same problem as us:

If `__init__.py` is in directory, the translation script thinks of it as a module and uses the module name as context. If there is not, it instead uses file name as context.

In our use case we had different modules that all had the same Lib. This made it so that each module had the Lib, but only one of them had a `__init__.py`. 

This caused the tree in source to be different than on build, and thus having a different context on runtime, making some translations don't work due to the context being different.

The solution we arrived at: have the same tree on source and build, having different names for each of those libs.

Issue #8745, opened 2025-09-23
DOC: Document TLS authentication support to DICOM Sender and Listener


jcfr:
## Is your feature request related to a problem? Please describe.

Add documentation associated with this feature:
* https://github.com/Slicer/Slicer/pull/8121

## Describe the solution you'd like

NA

## Describe alternatives you've considered

NA

## Additional context

Related pull requests:
* https://github.com/Slicer/Slicer/pull/8121

PR #8786, opened 2025-10-11
DOC: Add "Slicer integration in hospital/PACS context" to DICOM module docs

jcfr:
This adds a concise, operations-oriented section describing how to integrate Slicer (or Slicer-based apps) with enterprise PACS/VNA environments.

---

> [!TIP]
> For convenience, here is the link to review the generated documentation.
>
> See https://slicer--8786.org.readthedocs.build/en/8786/user_guide/modules/dicom.html#slicer-integration-in-hospital-pacs-context

### DIFF
--- a/Docs/user_guide/modules/dicom.md
+++ b/Docs/user_guide/modules/dicom.md
@@ -352,6 +352,7 @@ Authors:
 - Alireza Mehrtash (BWH)
 - Csaba Pinter (PerkLab, Queen's)
 - Andras Lasso (PerkLab, Queen's)
+- Jean-Christophe Fillion-Robin (Kitware)
 
 ## Acknowledgements
 
@@ -365,3 +366,4 @@ This work is part of the [National Alliance for Medical Image Computing](https:/
 ![](https://github.com/Slicer/Slicer/releases/download/docs-resources/logo_dicom_offis.png)
 ![](https://github.com/Slicer/Slicer/releases/download/docs-resources/logo_spl.png)
 ![](https://github.com/Slicer/Slicer/releases/download/docs-resources/logo_perklab.png)
+![](https://github.com/Slicer/Slicer/releases/download/docs-resources/logo_kitware.png)


jcfr:
Next, I will be updating documentation to address:
* https://github.com/Slicer/Slicer/issues/8745

Docs/user_guide/modules/dicom.md
```diff
@@ -244,6 +244,90 @@ Scanners may create image volumes with varying image slice spacing. Slicer can r
...
+### Supported technical patterns
+
+#### Push to Slicer (`C-STORE` → Slicer)
+
+**Architecture:** A modality, PACS, or routing node pushes specific studies/series to Slicer's AE.
+
+* **Pros:** Simple user experience; deterministic; works well for scheduled research/processing pipelines.
+* **Cons:** Requires routing rules; local copies on the Slicer host can become a shadow archive unless retention is governed; ensure consistent Patient ID/Accession handling.
+* **Requirements**
+  * Register Slicer AE with PACS: **AE Title / Host / Port** (commonly 104 or 11112).
```
jcfr, line 272:
```suggestion
  * Register Slicer AE with PACS: **AE Title / Host / Port**
```

pieper: I guess leaving this out here is good since the same information is elsewhere in the doc.

lassoan:
This is valuable content but it is quite different from the regular module documentation, and it is somewhat subjective and incomplete. In the future, we could add a new "guides" or 'implementation guides" category for such content, but for now we could just create a separate page for it (appear as a sub-page of the DICOM module documentation in the table of contents).

The selection of requirements sections look somewhat arbitrary to me. It is not clear why exactly these requirements are listed. Are there some DICOM or IHE documents that these requirements are based on? If yes, they should be referenced. If not, then maybe "requirement" should be renamed to something softer, such as "recommendations".

By the way, we should document all the new features and workflows that the visual DICOM browser and underlying completely reworked CTK DICOM infrastructure implements. For example, an important fearure is the Q/R via a proxy server, which allows using the container-friendly C-GET via a proxy server and a PACS that only supports C-STORE.

jcfr:
@Punzo To follow up on @lassoan comment, I am wondering how (and if) the list of workflows could be updated to account for capabilities only available in the DICOM Visual Browser.

Punzo:
> @Punzo To follow up on @lassoan comment, I am wondering how (and if) the list of workflows could be updated to account for capabilities only available in the DICOM Visual Browser.

@jcfr do you mean to have some documentation regarding the UI of the visual dicom browser? how to use it? not sure. here the documentation looks like a lot of useful general information regarding the protocols.

> By the way, we should document all the new features and workflows that the visual DICOM browser and underlying completely reworked CTK DICOM infrastructure implements. For example, an important fearure is the Q/R via a proxy server, which allows using the container-friendly C-GET via a proxy server and a PACS that only supports C-STORE.

@lassoan  do you think we should use screenshot? e.g.: screenshots in [link](https://projectweek.na-mic.org/PW42_2025_GranCanaria/Projects/VisualDICOMBrowser/)?

image

I can add a section, just let me know if you would like a pratical guide for the UI and how to setup things or something else.

Punzo:
@jcfr for the moment I will add (tomorrow) an high level paragraph description of visual dicom browser (what is the difference with the classic browser etc...). Then later on I can add more specific docs on the UI and how to use it (after the current UI refactoring).

jcfr:
> do you mean to have some documentation regarding the UI of the visual dicom browser? how to use it? not sure.

In a follow-up PR, adding a high level description for the Visual DICOM Browser and how to use if would be great. 

In the meantime, my specific comment was about reviewing content of this PR and tweaking the workflows based on Visual DICOM Browser capabilities. For example, under _Pull from PACS/VNA (Query/Retrieve)_, we could have a tip section indicating that download using Visual DICOM Browser are parallelized. There are probably other capabilities that could be documented (e.g proxying, ...)

pieper (approved):
LGTM 👍

PR #8791, opened 2025-10-14
DOC: Document the new .vp.json volume property file format

lassoan:
Volume properties are now saved in .vp.json file format by default. Describe this in the documentation.


lassoan:
It is not urgent at all, but I would like to get this merged just not to have so many open PRs.

@jcfr is there any drawbacks of bypassing merge rules and merging it myself (without a review)?

jamesobutler (approved):

PR #8795, opened 2025-10-16
BUG: Fix OpenGL error at startup in debug mode

lassoan:
Window resize events caused vtkMRMLThreeDViewInteractorStyle::DelegateInteractionEventToDisplayableManagers to attempt to look up 3D coordinates using the z buffer, which led to OpenGL errors (visible in debug mode only) when the window has never been rendered.

Fixed the issue by not looking up 3D coordinates for window resize (vtkCommand::ConfigureEvent) events.

### DIFF
--- a/Libs/MRML/DisplayableManager/vtkMRMLThreeDViewInteractorStyle.cxx
+++ b/Libs/MRML/DisplayableManager/vtkMRMLThreeDViewInteractorStyle.cxx
@@ -121,11 +121,13 @@ bool vtkMRMLThreeDViewInteractorStyle::DelegateInteractionEventToDisplayableMana
   }
 
   vtkNew<vtkMRMLInteractionEventData> ed;
-  ed->SetType(inputEventData->GetType());
+  int eventType = inputEventData->GetType();
+  ed->SetType(eventType);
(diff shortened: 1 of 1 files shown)


Libs/MRML/DisplayableManager/vtkMRMLThreeDViewInteractorStyle.cxx
```diff
@@ -121,11 +121,13 @@ bool vtkMRMLThreeDViewInteractorStyle::DelegateInteractionEventToDisplayableMana
   }
 
   vtkNew<vtkMRMLInteractionEventData> ed;
-  ed->SetType(inputEventData->GetType());
+  int eventType = inputEventData->GetType();
+  ed->SetType(eventType);
   int displayPositionCorrected[2] = { displayPositionInt[0] - pokedRenderer->GetOrigin()[0], displayPositionInt[1] - pokedRenderer->GetOrigin()[1] };
   ed->SetDisplayPosition(displayPositionCorrected);
+  bool eventHasPosition = (eventType != vtkCommand::ConfigureEvent);
```
jcfr, line 128: For reference, the `ConfigureEvent` is invoked while resizing the window. See https://github.com/Slicer/VTK/blob/slicer-v9.5.2-2025-09-16-7c0494a68/GUISupport/Qt/QVTKRenderWindowAdapter.cxx#L208-L225

Issue #8793, opened 2025-10-15
Error when selecting Human Orientation Marker


ruffsl:
## Summary

Selection of the Human Orientation Marker for any 3D or slice view results in a error and fails to visualize.

## Steps to reproduce

1. Launch Slicer from terminal
2. Select Human as Orientation Marker for any view
3. Check stdout for error output

<img width="3840" height="2160" alt="Image" src="https://github.com/user-attachments/assets/115df8b0-80ab-4cf9-ae6f-01509a769834" />

## Logs

```
$ /opt/slicer/Slicer --version
Slicer 5.9.0-2025-10-14

$ /opt/slicer/Slicer
Switch to module:  "Welcome"
GetOrientationMarkerModelPath: file does not exist: share/Slicer-5.9/OrientationMarkers/Human.vtp


Error opening file share/Slicer-5.9/OrientationMarkers/Human.vtp


Algorithm vtkXMLPolyDataReader (0x245f1150) returned failure for request: vtkInformation (0x25083220)
  Debug: Off
  Modified Time: 213956
  Reference Count: 1
  Registered Events: (none)
  Request: REQUEST_INFORMATION
  ALGORITHM_AFTER_FORWARD: 1
  FORWARD_DIRECTION: 0
```

```
[DEBUG][Qt] 15.10.2025 21:16:58 [] (unknown:0) - Session start time .......: 20251015_211658
[DEBUG][Qt] 15.10.2025 21:16:58 [] (unknown:0) - Slicer version ...........: 5.9.0-2025-10-14 (revision 33996 / c3ffd7f) linux-amd64 - installed release
[DEBUG][Qt] 15.10.2025 21:16:58 [] (unknown:0) - Operating system .........: Linux / 6.12.52 / #1-NixOS SMP PREEMPT_DYNAMIC Sun Oct 12 10:57:19 UTC 2025 / UTF-8 - 64-bit
[DEBUG][Qt] 15.10.2025 21:16:58 [] (unknown:0) - Memory ...................: 192905 MB physical, 212225 MB virtual
[DEBUG][Qt] 15.10.2025 21:16:58 [] (unknown:0) - CPU ......................: GenuineIntel Intel(R) Core(TM) i9-14900K, 24 cores, 32 logical processors
[DEBUG][Qt] 15.10.2025 21:16:58 [] (unknown:0) - VTK configuration ........: OpenGL2 rendering, TBB threading
[...]

PR #8800, opened 2025-10-21
BUG: Fix loading of orientation marker and fonts

ruffsl:
_Updated description from @lassoan:_

Recent changes in application initialization resulted in incorrectly computed paths for orientation marker and font files.

The root cause of the issue was that `vtkMRMLApplicationLogic::GetShareDirectory()` returns a relative path.

Fixed the issue by properly constructing full path from the relative path. Added a convenience function that makes it easier and less error-prone to get the full path of files and folders in the application share folder.

---

<details><summary><i>Original description by @ruffsl</i></summary>
<p>

This pull request updates how the `ShareDirectory` path is resolved in the `vtkMRMLApplicationLogic` class, improving robustness when handling relative and absolute paths. The main change is that `GetShareDirectory()` now returns an absolute path when possible and gracefully handles errors if required information is missing.

</p>
</details> 

---

Fixes https://github.com/Slicer/Slicer/issues/8793

### DIFF
--- a/Libs/MRML/Logic/vtkMRMLApplicationLogic.cxx
+++ b/Libs/MRML/Logic/vtkMRMLApplicationLogic.cxx
@@ -1233,24 +1233,45 @@ const std::string& vtkMRMLApplicationLogic::GetShareDirectory() const
 //----------------------------------------------------------------------------
 std::string vtkMRMLApplicationLogic::GetFontFilePath(const std::string& fontFileName)
 {
-  std::vector<std::string> filesVector;
-  // Add an empty component because JoinPath does not add a slash for the first two components.
-  filesVector.emplace_back("");
-  filesVector.emplace_back(this->GetShareDirectory());
-  filesVector.emplace_back(FONTS_DIR);
-  filesVector.emplace_back(fontFileName);
-  std::string fullPath = vtksys::SystemTools::JoinPath(filesVector);
-  return fullPath;
+  return this->GetShareFilePath(FONTS_DIR, fontFileName);
 }
 
 //----------------------------------------------------------------------------
 std::string vtkMRMLApplicationLogic::GetFontsDirectory()
 {
+  return this->GetShareFilePath(FONTS_DIR);
+}
+
+//----------------------------------------------------------------------------
+std::string vtkMRMLApplicationLogic::GetShareFilePath(const std::string& fileName) const
+{
+  return this->GetShareFilePath("", fileName);
+}
+
+//----------------------------------------------------------------------------
+std::string vtkMRMLApplicationLogic::GetShareFilePath(const std::string& subfolderName, const std::string& fileName) const
+{
+  if (this->GetHomeDirectory().empty())
+  {
+    vtkWarningMacro("vtkMRMLApplicationLogic::GetShareFilePath: Home directory is not set. Returned path may be invalid.");
+  }
+  if (this->GetShareDirectory().empty())
+  {
+    vtkWarningMacro("vtkMRMLApplicationLogic::GetShareFilePath: Share directory is not set. Returned path may be invalid.");
+  }
   std::vector<std::string> filesVector;
   // Add an empty component because JoinPath does not add a slash for the first two components.
   filesVector.emplace_back("");
+  filesVector.emplace_back(this->GetHomeDirectory());
   filesVector.emplace_back(this->GetShareDirectory());
-  filesVector.emplace_back(FONTS_DIR);
+  if (!subfolderName.empty())
+  {
+    filesVector.emplace_back(subfolderName);
+  }
+  if (!fileName.empty())
+  {
+    filesVector.emplace_back(fileName);
+  }
   std::string fullPath = vtksys::SystemTools::JoinPath(filesVector);
   return fullPath;
 }
--- a/Libs/MRML/Logic/vtkMRMLApplicationLogic.h
+++ b/Libs/MRML/Logic/vtkMRMLApplicationLogic.h
@@ -303,11 +303,19 @@ public:
   /// @}
 
   /// @{
-  /// Applications must set this variable to a meaningful value.
-  /// This path is relative, modules will generally resolve it from GetHomeDirectory().
-  /// For example, the slicer application will set this to "share/slicer-X-Y".
+  /// Applications must set this variable to a meaningful value (typically to something like "share/slicer-X-Y").
+  /// This path is relative to HomeDirectory. Absolute path of a file in the share directory can be constructed
+  /// using GetShareFilePath() convenience method.
+  /// \sa GetHomeDirectory, GetShareFilePath
   void SetShareDirectory(const std::string& path);
   const std::string& GetShareDirectory() const;
+  /// @}
+
+  /// @{
+  /// Get full path to a file (or a relative path) in the Share folder.
+  /// The path is constructed as HomeDirectory/ShareDirectory/subfolderName/shareFileName
+  std::string GetShareFilePath(const std::string& fileName) const;
+  std::string GetShareFilePath(const std::string& subfolderName, const std::string& fileName) const;
   /// @}
 
 protected:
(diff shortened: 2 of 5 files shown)


AlexyPellegrini:
No need to make such complex change. Also GetShareDirectory should stay a simple getter of the property "share dir". 

To fix the issue https://github.com/Slicer/Slicer/issues/8793 we simply have to add `
 filesVector.emplace_back(this->External->GetMRMLApplicationLogic()->GetHomeDirectory()); ` before `
 filesVector.push_back(this->External->GetMRMLApplicationLogic()->GetShareDirectory()); ` in [vtkMRMLOrientationMarkerDisplayableManager L.256](https://github.com/Slicer/Slicer/blob/6791fdda7149786f629b86d0f688cab551075480/Libs/MRML/DisplayableManager/vtkMRMLOrientationMarkerDisplayableManager.cxx#L256).

This is an expected behavior, as stated by the documentation of `GetShareDirectory`:
> This path is relative, modules will generally resolve it from GetHomeDirectory().

ruffsl:
> we simply have to add `filesVector.emplace_back(this->External->GetMRMLApplicationLogic()->GetHomeDirectory());`

Ok @AlexyPellegrini , how about b4fc265 ? I tested that locally, and it also resolves this issue.

lassoan:
> No need to make such complex change. Also GetShareDirectory should stay a simple getter of the property "share dir".

Unfortunately, a simple change that only fixes the orientation marker path will not be sufficient here. What about fixing incorrect font paths (that caused some panic this week before a big Slicer internationalization demo), and all the other places where extensions (some of them in custom applications) access shared resources? We need to fix the issue in a way that all code that worked before by simply prepending the ShareDirectory will keep working.

Preserving backward compatibility as much as possible is really important. Even if just a few hundred people were impacted by the error and it takes each of them a few hours to investigate and resolve, it would waste $10-20k of the community. Even if it takes us several hours to implement a proper fix, it is well worth it.

I'll push a fix on this branch for review, based on @ruffsl's initial implementation.

AlexyPellegrini:
`GetShareDirectory` replaces the `Slicer_SHARE_DIR` macro, and both have never been absolute paths, so I don't understand how this could be related the issues you are describing. Are you sure there is not something else that changes, such as the default working directory of Slicer?
Note that making `GetShareDirectory` an absolute path at runtime would be a breaking change too, as the current code composing paths using `GetHomeDirectory` and `GetShareDirectory` would break.
If you can give me the pointers to the code that actually broke I may be able to help you further, but for what I know now, I recommend keeping `GetShareDirectory` as it is now.

lassoan:
@AlexyPellegrini thanks for the follow-up. I've had a closer look at the code and it seems that all the trouble was caused by not useing the new vtkMRMLApplicationLogic::GetShareDirectory method correctly. This is very good, because we don't need to worry about backward compatibility. (The old features just broke because they got wrong path).

A simple solution would be to fix all the places where this new method was used incorrectly:
- `vtkMRMLOrientationMarkerDisplayableManager::vtkInternal::GetOrientationMarkerModelPath`
- `vtkMRMLApplicationLogic::GetFontFilePath`
- `vtkMRMLApplicationLogic::GetFontsDirectory`

However, the fact that this API got misused in several places indicates that this API is not intuitive enough and future developers may make the same mistake.

As you noted, while almost every paths/directories/folders get/set methods work with absolute paths - the "share" folder stored is an exception. In `qSlicerCoreApplication::slicerSharePath` the path is relative (for 10+ years) and we have not enough reasons to change it (as it would be a backward-incompatible change). Introducing a similar-sounding method in the application logic class returning absolute path would not be good either, as the inconsistency could be a source of misunderstanding.

Perhaps a good enough solution would be to:
1. Keep `vtkMRMLApplicationLogic::SetShareDirectory` and `vtkMRMLApplicationLogic::GetShareDirectory` as is, storing relative path
2. Discourage their use by adding a helper method that is safer and more convenient to use. It would be similar to `std::string vtkMRMLApplicationLogic::GetFontFilePath(const std::string& fontFileName);` something like `std::string vtkMRMLApplicationLogic::GetShareFilePath(const std::string& filename);`.
3. Switch to use this new method everywhere where we currently use `vtkMRMLApplicationLogic::GetShareDirectory` and then construct a full path from it. It will make the code simpler and it serve as a good example for future developers.

@AlexyPellegrini what do you think? If it sounds good to you then it would be great if you could implement it (or ask @ruffsl - but he might be getting a bit tired by us keep asking him different things).

AlexyPellegrini:
Proposed solution sounds good to me. Unfortunately I won't have time to do it.
I think the code from this PR can be merged as is as it fixes one issue, then another follow-up PR can be made to clean everything at once.

ruffsl:
> Perhaps a good enough solution would be to:

I've attempted to implement that, and the human orientation model is working, but not sure how to test everything else.

---

> I think the code from this PR can be merged as is as it fixes one issue, then another follow-up PR can be made to clean everything at once.

I'd also be fine with that, just sticking with the patch set as of b4fc265 .

lassoan:
Thank you @ruffsl . I've squashed your commits, removed the questions, and added one more convenience function to make it really easy to get a file in a subfolder of share.

Libs/MRML/Logic/vtkMRMLApplicationLogic.h
```diff
@@ -303,13 +303,21 @@ class VTK_MRML_LOGIC_EXPORT vtkMRMLApplicationLogic : public vtkMRMLAbstractLogi
...
+  /// \sa GetHomeDirectory, GetShareFilePath
   void SetShareDirectory(const std::string& path);
   const std::string& GetShareDirectory() const;
   /// @}
 
+  /// @{
+  /// Get full path to a file (or a relative path) in the Share folder.
+  /// The path is constructed as HomeDirectory/ShareDirectory/subfolderName/shareFileName
+  std::string GetShareFilePath(const std::string& fileName);
+  std::string GetShareFilePath(const std::string& subfolderName, const std::string& fileName);
```
jcfr:
```suggestion
  std::string GetShareFilePath(const std::string& fileName) const;
  std::string GetShareFilePath(const std::string& subfolderName, const std::string& fileName) const;
```

Libs/MRML/Logic/vtkMRMLApplicationLogic.cxx
```diff
@@ -1233,24 +1233,45 @@ const std::string& vtkMRMLApplicationLogic::GetShareDirectory() const
...
 }
 
 //----------------------------------------------------------------------------
 std::string vtkMRMLApplicationLogic::GetFontsDirectory()
 {
+  return this->GetShareFilePath(FONTS_DIR);
+}
+
+//----------------------------------------------------------------------------
+std::string vtkMRMLApplicationLogic::GetShareFilePath(const std::string& fileName)
```
jcfr:
```suggestion
std::string vtkMRMLApplicationLogic::GetShareFilePath(const std::string& fileName) const
```

Libs/MRML/Logic/vtkMRMLApplicationLogic.cxx
```diff
@@ -1233,24 +1233,45 @@ const std::string& vtkMRMLApplicationLogic::GetShareDirectory() const
...
+{
+  return this->GetShareFilePath("", fileName);
+}
+
+//----------------------------------------------------------------------------
+std::string GetShareFilePath(const std::string& subfolderName, const std::string& fileName)
```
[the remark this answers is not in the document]
jcfr: I  appended `const` to the Copilot suggestion

jcfr (changes requested):
After addressing the remaining nitpicks, it will be good for integration 🚀

lassoan:
Thanks for the review, I've made all the suggested changes. Build is still in progress, I'll then run the tests.

jcfr:
Building locally, when it completes I will integrate :rocket:

lassoan:
Manual and automatic testing did not find any issues.

PR #8193, opened 2025-01-28
COMP: Remove deprecated Jacobian methods for ITK_LEGACY_REMOVE compliance

jcfr:
Removed deprecated Jacobian-related methods to ensure compatibility with ITK configured with ITK_LEGACY_REMOVE set to ON.

Key Changes:
- Removed legacy methods `ComputeJacobianWithRespectToPosition` and `ComputeInverseJacobianWithRespectToPosition` in `vtkITKTransformInverse.h` and `itkWarpTransform3D.h`.
- Simplified code by eliminating outdated API usage.
- Updated `TransformPhysicalPointToIndex` calls for improved clarity and adherence to ITK's current conventions.

This will make it easier to turn on ITK_LEGACY_REMOVE, and eventually move to a newer version of ITK (e.g. 6.0).

PR #8820, opened 2025-11-05
Disable ITKv4 legacy functionality

dzenanz:
Closes #8193.


dzenanz:
I assume we want to integrate this after 5.10 tag, not before?

dzenanz:
The latest force-push is a plain rebase.

dzenanz:
Turning off ITK legacy support breaks BRAINSTools. This should not be integrated until that is fixed, and BRAINSTools updated. Meanwhile, I created a PR with most of these commits here: #8841.

PR #8829, opened 2025-11-07
BUG: Fix not accounting for closed surface rep for temp seg node

nhjohnston:
If a segmentation node has a parent transform node and a closed surface representation, the temporary node will be created. However, no closed surface segmentation stats will be computed for the temp volume because the temporary node does not have a closed surface. https://github.com/Slicer/Slicer/blob/3220bf57f25c367baa45cbfaef9fb97f34837fa4/Modules/Scripted/SegmentStatistics/SegmentStatisticsPlugins/ClosedSurfaceSegmentStatisticsPlugin.py#L30

This PR adds a closed surface to the temporary node if exists, allowing the closed surface segmentation stats to be computed

### DIFF
--- a/Modules/Scripted/SegmentStatistics/SegmentStatistics.py
+++ b/Modules/Scripted/SegmentStatistics/SegmentStatistics.py
@@ -445,6 +445,14 @@ class SegmentStatisticsLogic(ScriptedLoadableModuleLogic):
                 transformedSegmentationNode.HideFromEditorsOn()
                 slicer.mrmlScene.AddNode(transformedSegmentationNode)
                 transformedSegmentationNode.HardenTransform()
+
+                # Check to see if a closed surface representation exists for the original segmentation
+                import vtkSegmentationCorePython as vtkSegmentationCore
+                containsClosedSurfaceRepresentation = segmentationNode.GetSegmentation().ContainsRepresentation(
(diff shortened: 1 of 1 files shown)


jamesobutler:
@lassoan Have you had a chance to take look at the alternative fix you were thinking? Otherwise I will likely proceed with integration.

lassoan:
I could not reproduce this issue. I've created a segmentation, applied a warping transform (not hardened), created closed surface representation, computed segment statistics and I can see the closed surface statistics results.

Could you share a scene (upload somewhere and post the link here) that reproduces the issue?

lassoan:
I can see some inconsistencies with the display node, which is probably due to that the entire node is being copied (including all node references). Instead of `transformedSegmentationNode.Copy` it would be safer to use `CopyContents` and then apply and harden the parent transform of the original segmentation node.

lassoan:
@nhjohnston could you please test if it solves the problem that you reported in https://github.com/Slicer/Slicer/pull/8854?
If yes, then it would be better to merge this one, because #8854 fixes a specific symptom, while this PR fixes the root cause of the problem.

jamesobutler:
Closing as superseded by changes in https://github.com/Slicer/Slicer/pull/8854.

PR #8834, opened 2025-11-09
COMP: Reduce threshold determining use of response file with AUTOMOC

jcfr:
Since the default threshold of 32000 still leads to the "libuv process spawn failed: name too long" being reported when running Automatic MOC and UIC for target like `qSlicerBaseQTCore` ended up passing `-I` and `-D` arguments being in the +37k characters, this commit set the threshold to 16000 to that use of response file is triggered.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -858,6 +858,15 @@ set(CMAKE_MODULE_PATH
 #-----------------------------------------------------------------------------
 set_property(GLOBAL PROPERTY AUTOGEN_SOURCE_GROUP "Generated")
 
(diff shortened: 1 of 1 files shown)


jcfr:
Once I confirm this works as expected on the factory, I will move forward with integration :rocket: 

cc: @jamesobutler @ebrahimebrahim @lassoan 


For reference, last night results:

| |
|--|
| image |
| image |

jcfr:
Setting `CMAKE_AUTOGEN_COMMAND_LINE_LENGTH_MAX` allows to address the issue :tada: 

image

PR #8841, opened 2025-11-11
COMP: Support building with ITK legacy support ON/OFF

dzenanz:
This is a subset of #8820. This should be non-problematic, and can be merged immediately.

### DIFF
--- a/Libs/vtkITK/CMakeLists.txt
+++ b/Libs/vtkITK/CMakeLists.txt
@@ -46,11 +46,6 @@ set(${PROJECT_NAME}_ITK_COMPONENTS
   ITKRegionGrowing
   ITKThresholding
   ITKVTK
-  # Needed for ITKv5 so that SimpleFastMutexLock is an alias to std::mutex in ITKv5
-  # once ITKv4 is not supported, then we can simply replace SimpleFastMutexLock with std::mutex
-  # the use the DeprecatedLib allows building without many preprocessor conditionals
-  # for ITKv5/ITKv4 differences.
-  ITKDeprecated
   )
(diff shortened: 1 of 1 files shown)


dzenanz:
I think this is ready for merging.

jcfr:
I suggest to update the commit message of the last commit: `COMP: We no longer need ITKDeperecated library` -> `COMP: Update vtkITK removing obsolete ITKDeprecated dependency`

Consider removing the entire comment block.

dzenanz:
This force-push is a plain rebase. Changes coming soon.

Libs/MRML/Core/vtkITKTransformInverse.h
```diff
@@ -58,20 +58,22 @@ class InverseBSplineTransform : public BSplineTransform<TScalar, NDimensions, VS
   {
     itkExceptionMacro("Only storage methods are implemented for InverseBSplineTransform");
   }
-  // Deprecated in ITKv5. It should be removed when ITK_LEGACY_REMOVE is set to ON.
+#if !defined(ITK_LEGACY_REMOVE)
```
jcfr, line 61: If I interpret the comment correctly, should not this be `#if defined(ITK_LEGACY_REMOVE)` ?

dzenanz: The condition is correct as-is. This double negation is caused by legacy being on by default.

dzenanz: Also, these declarations should be left in for now, until we get around to turning legacy off. Then we will be able to remove them entirely.

dzenanz:
Can we please merge this?

jamesobutler (approved):

dzenanz:
Thank you James.

PR #8843, opened 2025-11-12
ENH: Last used DICOM import and 10 last used database directories are stored

xskere:
In the [discourse](https://discourse.slicer.org/t/edit-default-dicom-import-directory/44704) there was a request for Slicer to remember the last DICOM directory imported as well as the last few DICOM database directories and to be able to change between them. All of this data is stored with QSettings.

For the first part, storing the last DICOM directory imported. This will be stored in the DICOM group with the key `LastImportDirectory` in the QSettings.

On the `__init__` of SlicerDICOMBrowser this is added, to check if there is a value stored, if so the dialog for the import will point to the directory specified.

```
        if self.settings.value("DICOM/LastImportDirectory") is not None:
            lastImportDirectory = self.settings.value("DICOM/LastImportDirectory")
            if os.path.exists(lastImportDirectory):
                self.changeDialogDirectory(lastImportDirectory)
```

To store this value, also in SlicerDICOMBrowser, this was added to the `onDirectoryImported` method:

```
        if self.useExpertimentalVisualDICOMBrowser:
            browserUsed = self.dicomVisualBrowser
        else:
            browserUsed = self.dicomBrowser

        newDirectory = browserUsed.importDialog().directory().path()
        self.settings.setValue("DICOM/LastImportDirectory", newDirectory)
        self.changeDialogDirectory(newDirectory)
```

For the second part, storing and allowing easy access to the last 10 used DICOM database directories, the DICOM.ui was changed:

<div align='center'>image</div>

A new collapsible button was added, in which there is a ComboBox that displays the 10 last used DICOM database directories, as well as a button to switch between them.

The storage of the 10 directories is done in a similar way to the last import directory, however this time the storage is done whenever `updateDatabaseDirectoryFromWidget` on the DICOMWidget is run.

This is how it all would appear in the .ini file:

```
LastImportDirectory=D:/DICOMTEST
NumberOfDirectoriesToKeep=10
LastDatabaseDirectory\0=D:/DICOMTEST4
LastDatabaseDirectory\1=D:/DICOMTEST5
LastDatabaseDirectory\2=D:/DICOMTEST9
LastDatabaseDirectory\3=D:/DICOMTEST8
LastDatabaseDirectory\4=D:/DICOMTEST11
LastDatabaseDirectory\5=D:/DICOMTEST10
LastDatabaseDirectory\6=D:/DICOMTEST7
LastDatabaseDirectory\7=D:/DICOMTEST6
LastDatabaseDirectory\8=D:/DICOMTEST3
LastDatabaseDirectory\9=D:/DICOMTEST2
```

[Re -> Discourse](https://discourse.slicer.org/t/edit-default-dicom-import-directory/44704)

### DIFF
--- a/Modules/Scripted/DICOM/DICOM.py
+++ b/Modules/Scripted/DICOM/DICOM.py
@@ -674,7 +674,7 @@ class DICOMWidget(ScriptedLoadableModuleWidget):
         # Browser settings
         #
 
-        self.ui.browserSettingsFrame.collapsed = True
+        self.ui.browserSettingsFrame.collapsed = False
 
         self.updateDatabaseDirectoryFromBrowser(self.browserWidget.dicomBrowser.databaseDirectory)
         # Synchronize database selection between browser and this widget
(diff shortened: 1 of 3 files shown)


lassoan:
Thanks for working on this, it is a very useful addition. For path selection we always use ctkPathLine edit for many reasons: consistency (everywhere else this widget or the directory selector is used for selecting a folder), it has auto-complete that helps the user to specify a correct path faster while avoiding some common issues (path separator inconsistencies, extra spaces, quotation marks, typos, ...), it can handle long paths intelligently (while horizontal comobobox sizing is quite limited and it may end up making the GUI layout very wide or not using the available horizontal space; while ctkPathLine can elide the path), and it has a built-in combobox of recently used paths that it can store in the application settings (you just need to specify the settings key and tell it when to store a new path). Could you the PR to use this widget?

We also need to make sure that the user clicks the "Switch" button (it should be renamed to "Apply" or some other word that is more commonly used and is very clear). For this, it could help if at least the "Apply" button appears or becomes enabled only when there is a change to apply (when the current database path is not the same as the one in the GUI). It would be nice to have an observer on the database and if the path is changed then immediately make the "Apply" button available.

For the future:
- If we find that users sometimes forget to click the button then we could validate/apply when the user leaves the widget (when the widget loses focus), but to get started, we can try the apply button.
- Instead of just a combobox, it could be nicer to have a table or a list of named items (a short descriptive name would be displayed instead of just the path) and have `+` and `-` buttons to add/remove paths (instead of automatically remembering the last 10, which may sometimes be too much, in other cases no enough).

cpinter:
> we always use ctkPathLine

Very good point, thanks Andras!

> We also need to make sure that the user clicks the "Switch" button

I suggested this word to make the button as narrow as possible, but Apply is a good idea as well, and short too.

> "Apply" button appears or becomes enabled only when there is a change to apply

Makes sense as well.

@xskere Please make these three changes when you have time and amend the commit. Thanks a lot!

xskere:
Thank you so much for the feedback! I have amended the last commit and changed:

- The comboBox was changed for -> ctkPathLine
- The Switch button text for -> Apply
- Apply button is now disabled on start: If there are changes on the text it checks if what's written is a directory and if it's different than the current database set, if so, the button enables!

pieper:
Thanks for working on this @xskere 👍 

I agree with the points Andras made, so it's great to iterate a bit on this feature.  It does make sense to have a list of database paths, which could be like the additional module paths in the application settings.

Also, since this will a useful feature I think having the database settings collapsible box open by default would be reasonable (also move it above the networking collapsible box).  We aren't really tight on space in the dicom module panel.

pieper:
> Also, since this will a useful feature I think having the database settings collapsible box open by default would be reasonable (also move it above the networking collapsible box). We aren't really tight on space in the dicom module panel.

Another option could be to move just the database directory picker out of the collapsible button and leave the other settings in the collapsible button and leave it collapsed.

xskere:
I fixed an error in the first commit message (Forgot to start message with uppercase, my bad)

Also sorry about the multiple force-pushes, I read this wrong:

> Also, since this will a useful feature I think having the database settings collapsible box open by default would be reasonable (also move it above the networking collapsible box). We aren't really tight on space in the dicom module panel.

And thought it was the previous database locations collapsible box the one that had to be on top the networking collapsible box, with this last amend I put the correct DICOM database settings collapsible box open and on top the networking collapsible box

PR #8742, opened 2025-09-23
COMP: Add support for Visual Studio 2026

jamesobutler:
## Visual Studio 2026 insiders is here
This enables using Visual Studio 2026 to build Slicer. The official release of Visual Studio 2026 is expected later this year, but the product has been announced with insiders builds available to download. This PR aims to make it possible to use the insiders build while Visual Studio 2022 should still be used for production use.
- [Visual Studio 2026 Insiders is here!](https://devblogs.microsoft.com/visualstudio/visual-studio-2026-insiders-is-here/)
- [C++ Language Updates in MSVC Build Tools v14.50](https://devblogs.microsoft.com/cppblog/c-language-updates-in-msvc-build-tools-v14-50/)

## Slicer testing results:
I have tested on the Windows platform using Visual Studio 2026 Insiders 18.0.0 build 11012.119 along with atest CMake dev with recent Visual Studio 2026 handling: https://cmake.org/files/dev/cmake-4.1.20250921-gfe97089-windows-x86_64.msi
{7F407BDB-6475-4855-8719-5C6C02E63C79}
image
- Build ✔️ 
- Package ✔️ 
- Test ✔️

PR #8851, opened 2025-11-13
DOC: Revise Visual Studio installation instructions for Windows

lassoan:
VS2022 is no longer offered by default, but it is needed by Slicer build (e.g., CMake 3 does not support VS2026, CMake 4 does not work with many libraries). Provided instructions how to get VS2022.

### DIFF
--- a/Docs/developer_guide/build_instructions/windows.md
+++ b/Docs/developer_guide/build_instructions/windows.md
@@ -14,7 +14,8 @@ Slicer relies on a number of large third-party libraries (such VTK, ITK, DCMTK),
     - 3.25.0 to 3.25.2 (CMake issues [24180](https://gitlab.kitware.com/cmake/cmake/-/issues/24180), [24567](https://gitlab.kitware.com/cmake/cmake/-/issues/24567))
 - [Git](https://git-scm.com/download/win) >= 1.7.10
   - Note: CMake must be able to find `git.exe` and `patch.exe`. If git is installed in the default location then they may be found there, but if they are not found then either add the folder that contains them to `PATH` environment variable; or set `GIT_EXECUTABLE` and `Patch_EXECUTABLE` as environment variables or as CMake variables at configure time.
-- [Visual Studio](https://visualstudio.microsoft.com/downloads/): any edition can be used (including the free Community edition), when configuring the installer:
+- Visual Studio:
+  - Download [free Community edition](https://aka.ms/vs/17/release/vs_community.exe) (recommended) or a [commercial edition](https://learn.microsoft.com/en-us/visualstudio/releases/2022/release-history#evergreen-bootstrappers) of Visual Studio 2022. Visual Studio 2026 with v143 toolset and latest CMake 4.x is usable for building Slicer core and some of its extensions, but it is not an officially supported build configuration yet.
   - Enable `Desktop development with C++` and in installation details
   - Enable the `MSVC v145 - VS2026 C++ x64...` (Visual Studio 2026 v145 toolset with 64-bit support) component - in some distributions, this option is not enabled by default.
   - Enable the `MSVC v143 - VS2022 C++ x64...` (Visual Studio 2022 v143 toolset with 64-bit support) component - in some distributions, this option is not enabled by default.


lassoan:
@jamesobutler you have commented recently on VS2026 compilation. How far did you get? I got immediately stuck at CMake3 latest version not recognizing VS2026.

jamesobutler:
I successfully built 3D Slicer with Visual Studio 2026 and I used the CMake 4.2.0 dev build (though now there are release candidates that can be used, see https://cmake.org/download/ for 4.2.0-rc3). The results of my Slicer build, package, and test can be found in https://github.com/Slicer/Slicer/pull/8742#issue-3443217828. I also built with the 2026 v145 toolset instead of using Visual Studio 2026 to build as Visual Studio 2022 with the v143 toolset.

lassoan:
Can we tell users to just use vs2026 then?

jamesobutler:
Yes we could simplify the instructions to say use Visual Studio 2026, but use CMake 4.2.0-rc3 or newer. Of course I haven't tested all extensions built with the newer generator, but folks could also still "build as Visual Studio 2022" with the v143 toolset install option as already described at:
https://github.com/Slicer/Slicer/blob/7755c94b3432a98e83892ee4625f17b3af6bf56c/Docs/developer_guide/build_instructions/windows.md?plain=1#L20
That would be in case something about the v145 toolset has issues, but I haven't generally observed failures as it has updated from v141, v142, v143, and now v145. The biggest impact would be the use of CMake 4 if there are old extensions still specifying thing likes `cmake_minimum_required(2.8)` or something which is a hard failure now. Developers will have to update their ancient CMake build system to be compatible with CMake 4.

jamesobutler:
We could also consider soon updating the Slicer factory build machine to use Visual Studio 2026 like new users would also be using so that from factory we are also testing the build process for the new generator. cc: @sjh26 @ebrahimebrahim

Docs/developer_guide/build_instructions/windows.md
```diff
@@ -14,7 +14,7 @@ Slicer relies on a number of large third-party libraries (such VTK, ITK, DCMTK),
...
   - Note: CMake must be able to find `git.exe` and `patch.exe`. If git is installed in the default location then they may be found there, but if they are not found then either add the folder that contains them to `PATH` environment variable; or set `GIT_EXECUTABLE` and `Patch_EXECUTABLE` as environment variables or as CMake variables at configure time.
-- [Visual Studio](https://visualstudio.microsoft.com/downloads/): any edition can be used (including the free Community edition), when configuring the installer:
+- [Visual Studio 2022](https://visualstudio.microsoft.com/downloads/): any edition can be used (including the free Community edition), when configuring the installer. However, Visual Studio 2022 version must be installed, not Visual Studio 2026, which is offered by default (because that is not compatible with many software libraries). Visual Studio 2022 installation is accessible by downloading `VisualStudioSetup.exe` and then run it from the command line using these arguments: `VisualStudioSetup.exe install --channelId VisualStudio.17.Release --productId Microsoft.VisualStudio.Product.Community --channelURI "https://aka.ms/vs/17/release/channel"`
```
jamesobutler: Instead of running that command line call, you can probably get the installer easier from this page:
https://learn.microsoft.com/en-us/visualstudio/releases/2022/release-history#evergreen-bootstrappers such as https://aka.ms/vs/17/release/vs_community.exe

lassoan: Good find, but I'm not sure such bootstrappers exist for Community edition.

sjh26:
> We could also consider soon updating the Slicer factory build machine to use Visual Studio 2026 like new users would also be using so that from factory we are also testing the build process for the new generator. cc: @sjh26 @ebrahimebrahim

Yes, we can start moving toward this

lassoan:
If the build is successful with VS2026 then I'll have to update the instructions accordingly (describe the official and an experimental configuration).

Docs/developer_guide/build_instructions/windows.md
jamesobutler: Visual Studio 2022 17.14 has a "Community" version as well as Professional and Enterprise. Other Visual Studio 2022 version only have Professional and Enterprise downloads.

This is the "Community" one https://aka.ms/vs/17/release/vs_community.exe

jcfr:
> Re: 2026

If build of the stable extensions is done  using 2026, considering setting the toolset to v143 for those.

Docs/developer_guide/build_instructions/windows.md
lassoan: Thank you, the link that I used previously to install VS2022 Community now only offers VS2026, but your https://aka.ms/vs/17/release/vs_community.exe link still offers VS2022!

lassoan:
Thanks to all for the feedbacks. I've pushed an update to include the direct VS2022 download links and clarify but V2026 is not supported but may work. Please have a look and let me know if you think further tweaks are needed.

Docs/developer_guide/build_instructions/windows.md
```diff
@@ -14,7 +14,8 @@ Slicer relies on a number of large third-party libraries (such VTK, ITK, DCMTK),
     - 3.25.0 to 3.25.2 (CMake issues [24180](https://gitlab.kitware.com/cmake/cmake/-/issues/24180), [24567](https://gitlab.kitware.com/cmake/cmake/-/issues/24567))
 - [Git](https://git-scm.com/download/win) >= 1.7.10
   - Note: CMake must be able to find `git.exe` and `patch.exe`. If git is installed in the default location then they may be found there, but if they are not found then either add the folder that contains them to `PATH` environment variable; or set `GIT_EXECUTABLE` and `Patch_EXECUTABLE` as environment variables or as CMake variables at configure time.
-- [Visual Studio](https://visualstudio.microsoft.com/downloads/): any edition can be used (including the free Community edition), when configuring the installer:
+- Visual Studio:
+  - Download [free Community edition](https://aka.ms/vs/17/release/vs_community.exe) (recommended) or a [commercial edition](https://learn.microsoft.com/en-us/visualstudio/releases/2022/release-history#evergreen-bootstrappers) of Visual Studio 2022. Visual Studio 2026 with v143 toolset and latest CMake4 might be usable for building Slicer core and some of its extensions, but it is not an officially supported build configuration yet.
```
jamesobutler: From my testing “it is” usable for building Slicer core, and only “might be” usable for building Slicer extensions since that part is untested.

jamesobutler (approved):
Overall this will be helpful for some specific folks.

Docs/developer_guide/build_instructions/windows.md
jcfr:
```suggestion
  - Download [free Community edition](https://aka.ms/vs/17/release/vs_community.exe) (recommended) or a [commercial edition](https://learn.microsoft.com/en-us/visualstudio/releases/2022/release-history#evergreen-bootstrappers) of Visual Studio 2022. Visual Studio 2026 with v143 toolset and latest CMake 4.x might be usable for building Slicer core and some of its extensions, but it is not an officially supported build configuration yet.
```

lassoan:
Updated, hopefully final version, please approve so that it gets merged. Thanks!

jcfr (approved):

jamesobutler:
CMake 4.2.0 with Visual Studio 2026 generator support has been tagged https://gitlab.kitware.com/cmake/cmake/-/tags/v4.2.0. Should be up and available on https://cmake.org/download/ probably by tomorrow.

jamesobutler:
FYI Microsoft doing some promoting of Visual Studio 2026 in Visual Studio 2022 installs:
{17B9A207-8E26-4DAD-B5C0-2C66F0C12748}

PR #8853, opened 2025-11-14
COMP: Fix crash when setting none display in SH combobox with no scene

cpinter:
Re https://discourse.slicer.org/t/qmrmlsubjecthierarchycombobox-setting-nonedisplay-from-python-segfaults/45052

### DIFF
--- a/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyModel.cxx
+++ b/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyModel.cxx
@@ -275,6 +275,11 @@ void qMRMLSubjectHierarchyModelPrivate::completePendingDragAndDropReverts(QMap<v
     if (!newParentItem)
     {
       newParentItem = q->subjectHierarchySceneItem();
+      if (!newParentItem)
+      {
+        // scene is not set yet
+        continue;
+      }
     }
     // If the item has no parent, then it means it hasn't been put into the hierarchy yet and it will do it automatically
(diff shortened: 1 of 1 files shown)


Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyModel.cxx
```diff
@@ -2037,7 +2037,7 @@ void qMRMLSubjectHierarchyModel::setNoneDisplay(const QString& displayName)
   }
   d->NoneDisplay = displayName;
 
-  if (d->NoneEnabled)
+  if (d->NoneEnabled && d->MRMLScene != nullptr)
```
lassoan: Does it crash because `this->subjectHierarchySceneItem()` returns `nullptr`? Then we should check for that (because maybe it can return `nullptr` in other cases, too).

If `subjectHierarchySceneItem()` crashes when `d->MRMLScene != nullptr` then that crash should be fixed, too (just as preventive action).

cpinter: No it was a crash when getting the scene item, it did not have the chance to return anything.

cpinter: > then that crash should be fixed, too

OK but then it will have to wait.

cpinter:
Thank you!

shai-ikko:
Thank you both!

PR #8854, opened 2025-11-15
Fix missing closed surface segment statistics if segmentation node is transformed

lassoan:
This commit fixes segmentation node transformation (contained representation names are preserved) and thereby fixes the root cause of the issue described in https://github.com/Slicer/Slicer/pull/8829


lassoan:
Thanks for testing!

PR #8858, opened 2025-11-19
BUG: Branding text in the installer was being ignored

dzenanz:
```
CMake Warning (dev) at CMake/SlicerCPack.cmake:477 (elseif):
  ELSEIF called with no arguments, it will be skipped.
  CMake/LastConfigureStep/CMakeLists.txt:44 (include)
```


jcfr:
Since following https://github.com/Slicer/Slicer/commit/c22540839cb55c206331579fc4757b65a2d2540a, the minimum required version is `3.20.6`, how about simplifying and removing the branch ?

Related pull request:
* https://github.com/Slicer/Slicer/pull/8501

jamesobutler (approved):

PR #8862, opened 2025-11-19
ENH: Update ITK to v5.4.5

dzenanz:


jcfr:
Thanks for the update 🙏 

Could you update both commit message and PR description to reference the ITK changes like it was done in https://github.com/Slicer/Slicer/commit/a78db558e7d5cb15f30394e93f81bcb7cb0ec255

In this case, it would be similar to the following:

````
List of ITK changes:

```
$ git shortlog e5dd69339..f51594ad8 --no-merges
<OUTPUT COPIED HERE>
```
````

> [!TIP]
> If the output of `git shortlog` has string like `#NNN` replace the text with `PR-NNN`

File: SuperBuild/External_ITK.cmake [hunk 1]
```diff
@@ -33,7 +33,7 @@ if(NOT DEFINED ITK_DIR AND NOT Slicer_USE_SYSTEM_${proj})
 
   ExternalProject_SetIfNotDefined(
     Slicer_${proj}_GIT_TAG
-    "e5dd69339bf0c436db3650eadd3c2a940c330b77" # slicer-v5.4.4-2025-06-11-f98d5fa
+    "f51594ad88194a72cdc06a314dafff709fcdc2e9" # slicer-v5.4.5-2025-11-19-f51594a
```
jamesobutler: To account for the custom commits on the Slicer fork branch
```suggestion
    "ac65b49c34fcfa3c3422a66026c7fe2bbaa88902" # slicer-v5.4.5-2025-11-19-f51594a
```

jcfr: Thanks for catching this 🙏

dzenanz:
Force-pushed with changes. Thank you James and JC.

jamesobutler:
@jcfr Should the list of ITK changes be to the updated commit hash? I have always wondered this because it will re-include in the shortlog the backported/custom commits on the fork branch even though they aren't actually "new" changes, but just newly on top of the updated upstream code.

dzenanz:
I think this is ready for merging. Are some infrastructure changes needed after this is merged, like updating some docker images or something?

jamesobutler (approved):
Automation nightly builds will appropriately update the build slicer for CI purposes.

PR #8884, opened 2025-12-02
BUG: Fix log10 display of inverted color tables

### LINKED ISSUE #8876: Unable to set log10 lookup table scale when inverting scalar volume display
## Summary

When inverting the color look up table scalar volume display, changing the lookup table scale to log10 instead of linear doesn't work as expected.

## Steps to reproduce

```python
import SampleData
volume_node = SampleData.SampleDataLogic().downloadMRHead()
lookup_table = volume_node.GetDisplayNode().GetColorNode().GetLookupTable()
volume_node.GetDisplayNode().SetInvertDisplayScalarRange(False)
lookup_table.SetScaleToLog10()  # works
lookup_table.SetScaleToLinear()  # works
volume_node.GetDisplayNode().SetInvertDisplayScalarRange(True)
lookup_table.SetScaleToLog10()  # does not work
lookup_table.SetScaleToLinear()  # does not work
```

Likely relevant code:
https://github.com/Slicer/Slicer/blob/4fb8874978c94cb3cf6dcb3bba32df7e6991c76e/Libs/MRML/Core/vtkMRMLScalarVolumeDisplayNode.cxx#L589-L624
## Environment
- Slicer version: Slicer-5.11.0-2025-11-23
- Operating system: Windows

lassoan:
When color table was inverted then a new color table (or transfer function) was instantiated and the log10 flag was not copied into the new object.

Fixed by not making another copy of the color table when InvertDisplayScalarRange is enabled, but only reverse the order of color entries.

fixes #8876

### DIFF
--- a/Libs/MRML/Core/vtkMRMLScalarVolumeDisplayNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLScalarVolumeDisplayNode.cxx
@@ -592,32 +592,31 @@ void vtkMRMLScalarVolumeDisplayNode::UpdateLookupTable(vtkMRMLColorNode* newColo
       if (vtkColorTransferFunction* ctf = vtkColorTransferFunction::SafeDownCast(newLookupTable))
       {
         // For continuous transfer functions, invert the positions
-        vtkNew<vtkColorTransferFunction> invertedCtf;
+        vtkColorTransferFunction* originalCtf = vtkColorTransferFunction::SafeDownCast(lookupTable);
+
+        ctf->RemoveAllPoints();
         double range[2] = { 0, 255 };
-        ctf->GetRange(range);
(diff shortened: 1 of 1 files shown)


lassoan:
> the Lookup Table combobox updates when I change between linear and Log10, but does not appear to update when "Invert" is toggled. The Histogram widget below does update its display of the color table based on the current log/linear and invert on/off state. It would be nice to see the lookup table preview in the combobox update when toggling the "Invert" option

Behavior of the "invert" checkbox is as intended: it inverts the scalar range mapping (does not modify the stored read-only color table).

If support of Log10 will be added to MRML in the future then it should be implemented the same way as other the mapping options (invert scalar range, window/level, threshold): toggling "Log10" checkbox will not modify the shared VTK color table in the color node, but it will only modify the copy of the color table in the volume display node.

PR #8898, opened 2025-12-10
ENH: Disable ITKv4 legacy classes and interfaces

dzenanz:
This disables support for interfaces and classes which existed in ITKv4. This leaves only ITKv5 code enabled. This is a first step towards transitioning to ITKv6, which is currently in beta testing.

PR #8911, opened 2025-12-15
ENH: Enable ITKv4 legacy for Rigid3D and BSplineDeformable transforms

dzenanz:
These transforms were deprecated in ITK, but transforms files which use them still exist. If they ever get removed from ITK, they will need to be moved to Slicer proper. This is a follow-up to: 6b52fb4f787d2eadee9870a3443b199168a0874c. Problem reported here: https://github.com/Slicer/Slicer/pull/8898#issuecomment-3648783628.

### DIFF
--- a/SuperBuild/External_ITK.cmake
+++ b/SuperBuild/External_ITK.cmake
@@ -69,9 +69,8 @@ if(NOT DEFINED ITK_DIR AND NOT Slicer_USE_SYSTEM_${proj})
   set(EP_BINARY_DIR ${CMAKE_BINARY_DIR}/${proj}-build)
 
   list(APPEND EXTERNAL_PROJECT_OPTIONAL_CMAKE_CACHE_ARGS
-    -DITK_LEGACY_REMOVE:BOOL=ON     #<-- Disable LEGACY ITKv4 features in preparation for transition to ITKv6.
+    -DITK_LEGACY_REMOVE:BOOL=OFF    #<-- Enable LEGACY ITKv4 to avoid dealing with removal of Rigid3DTransform and BSplineDeformableTransform.
     -DITK_LEGACY_SILENT:BOOL=OFF    #<-- Use of legacy code will produce compiler warnings
-    -DModule_ITKDeprecated:BOOL=OFF #<-- Needed for ITKv5 now. (itkMultiThreader.h and MutexLock backwards compatibility.)
     )
 
 


lassoan:
From @dzenanz in https://github.com/Slicer/Slicer/pull/8898#issuecomment-3656879542:

> @hjmjohnson in [8cd2517](https://github.com/InsightSoftwareConsortium/ITK/commit/8cd251744a9ad8c18a72bd1821b16c0a09fe0b38) initially removed `Rigid3DTransform` as part of larger v3 feature removal, but soon after [re-added it](https://github.com/InsightSoftwareConsortium/ITK/commit/76bc2bba243f69759f4fb707a6d505e696a). This transform type appears in files, such as https://github.com/Slicer/Slicer/blob/v5.10.0/Modules/CLI/ResampleDTIVolume/Data/Input/rotationAndAffine.tfm#L3. I guess we should not un-deprecate it in ITK. Two possible solutions are:
>
>1. add a copy of `Rigid3DTransform` to Slicer (and register it with transform factory if legacy is disabled)
>2. keep ITK legacy permanently enabled inside Slicer (until `Rigid3DTransform` is removed from Slicer, and then we will have to apply option 1)
>
>Are there some preferences for option 1 or option 2? Option 2 is easy.
>
>The same logic from [Rigid3DTransform](https://github.com/InsightSoftwareConsortium/ITK/blob/v5.4.5/Modules/IO/TransformFactory/src/itkTransformFactoryBaseRegister.hxx#L139-L145) also applies to [BSplineDeformableTransform](https://github.com/InsightSoftwareConsortium/ITK/blob/v5.4.5/Modules/IO/TransformFactory/src/itkTransformFactoryBaseRegister.hxx#L118-L121)

lassoan:
We should re-enable legacy support as a quick fix of the current Slicer core test errors and extension build errors.

As time permits, we should continue to work towards removing dependencies on legacy ITK features to prepare for an ITKv6 upgrade.

jamesobutler:
If `Rigid3DTransform` is deprecated in ITK, should Slicer also consider it deprecated and encourage folks to not use it and plan for the feature to longer be supported in the future? Just wondering how Slicer users will know to not use deprecated ITK functionality.

dzenanz:
I am not sure that Slicer users can directly specify Rigid3DTransform. Slicer developers are hopefully smart enough to use one of [Rigid3DTransform](https://docs.itk.org/projects/doxygen/en/stable/classitk_1_1Rigid3DTransform.html)'s derived classes, which are all more efficient for registration and representation.

lassoan:
> how Slicer users will know to not use deprecated ITK functionality

Slicer developers enabling the legacy removal flag and run tests and then notify extension developers could work well.

Ideally, this would happen way before we want to upgrade to ITK6 so that Slicer core and extension developers have time to work on it.

dzenanz:
ITK 6 is still in beta, so we are at least months away from migration to ITKv6. But transition **away** from ITKv4 legacy functionality can be started today, by developers disabling legacy flag and fixing ensuing errors. I would like to hear @hjmjohnson's opinion before starting option 1.

dzenanz:
Review/merge?

jamesobutler:
@dzenanz Is it correct that with needing ITK_LEGACY_REMOVE on, but successfully having Module_ITKDeprecated off, means Slicer requires some Legacy ITK v4 functionality, but is not using any deprecated features currently in ITK v5?

dzenanz:
Correct. We are not relying on anything in the ITKv5's [Deprecated module](https://github.com/InsightSoftwareConsortium/ITK/tree/release-5.4/Modules/Compatibility/Deprecated/include). And in `main` (from which v6 will be cut), [Deprecated module](https://github.com/InsightSoftwareConsortium/ITK/tree/main/Modules/Compatibility/Deprecated/include) is (currently) empty, containing only placeholder code in order to compile.

hjmjohnson:
@dzenanz Thanks. This looks good to me.

PR #8915, opened 2025-12-16
ENH: Show number of control points in SH tooltip of markup nodes

cpinter:
Many times I'd find it useful to see the number of control points for a markups node in the Data module. This commit adds this.


cpinter:
Copilot was right, and I reverted those changes. Please let me know if this is acceptable now.

PR #8922, opened 2025-12-19
ENH: Improve Git repository detection in SlicerMacroExtractRepository

samuel100u:
Instead of strictly checking for a .git folder in the source directory, use 'git rev-parse --is-inside-work-tree'. This allows the build system to correctly identify the repository even when CMake is run from a subdirectory (e.g. when the .git folder is in a parent directory).

### DIFF
--- a/CMake/SlicerMacroExtractRepositoryInfo.cmake
+++ b/CMake/SlicerMacroExtractRepositoryInfo.cmake
@@ -82,13 +82,25 @@ macro(SlicerMacroExtractRepositoryInfo)
   set(${wc_info_prefix}_WC_REVISION_NAME "NA")
   set(${wc_info_prefix}_WC_REVISION_HASH "NA")
 
-  if(NOT EXISTS ${MY_SOURCE_DIR}/.git)
+  find_package(Git QUIET)
+  set(is_git_checkout FALSE)
+  if(GIT_FOUND)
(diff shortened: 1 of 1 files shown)


pieper (commented):
This looks reasonable to me.

@jcfr are you available to take a quick look?

lassoan:
What problem does this proposed change solve?
Is there any change in behavior if git is not installed or the Slicer source code is not a git repository?

samuel100u:
> What problem does this proposed change solve? Is there any change in behavior if git is not installed or the Slicer source code is not a git repository?

This change removes the restriction that the CMakeLists.txt must be in the same directory as the .git folder. It allows the build system to correctly identify the repository even in nested directory structures or when the source is part of a larger monorepo.

pieper:
Yes, but in what scenario is that functionality required or useful?  Do you have an open source example that would use this feature?

lassoan:
If the copilot findings are fixed then I don't see any problems integrating this. But to make sure that it is worth the effort of fixing up, reviewing, and integrating the change, could you please confirm that this is the only change that prevents using Slicer in a monorepo? I can believe it, I'm just somewhat surprised, as Slicer build relies so heavily on the superbuild mechanism that it is surprising that this is really the only change you need to make.

jamesobutler:
@samuel100u Do you still have interest in this work and would be able to answer the above questions?

jamesobutler:
Closing stale PR based on no response from previous comment.

PR #8930, opened 2025-12-29
Fix mention of SWIG required external project and update to deep copy PushVolumeToSlicer

jamesobutler:
This addresses the items mentioned in https://github.com/Slicer/Slicer/pull/8923#issuecomment-3694033511.

`BUG: Fix appending SWIG as required external project`:
This should have been included in https://github.com/Slicer/Slicer/commit/f87c977465f3b63219456186b7673f5c12e8b514 when SWIG was removed as an external project.

`BUG: Fix PushVolumeToSlicer not using deep copy`:
Prior to https://github.com/Slicer/Slicer/commit/a92c5534b5153b367b1c285077f64bac640bf3fc, PushVolumeToSlicer was utilizing a deep copy during conversion which was accidentally changed to a shallow copy as numpy_to_vtk has a default of a shallow copy. See https://github.com/Kitware/VTK/blob/7c0494a68bff379d32d6b1fbaa3d10d27a73af54/Wrapping/Python/vtkmodules/util/numpy_support.py#L104. This restores the expected behavior as a deep copy.

### DIFF
--- a/Base/Python/sitkUtils.py
+++ b/Base/Python/sitkUtils.py
@@ -15,6 +15,8 @@ def _get_lps_to_ras_matrix():
 def PushVolumeToSlicer(sitkimage, targetNode=None, name=None, className="vtkMRMLScalarVolumeNode"):
     """Given a SimpleITK image, push it back to slicer for viewing
 
+    A deep copy of the image data is created during conversion.
+
     :param targetNode: Target node that will store the image. If None then a new node will be created.
     :param className: if a new target node is created then this parameter determines node class. For label volumes, set it to vtkMRMLLabelMapVolumeNode.
     :param name: if a new target node is created then this parameter will be used as basis of node name.
@@ -123,8 +125,10 @@ def vtk2sitk(vtkimg, origin=None, spacing=None, direction=None):
     return sitkimg
 
 
-def sitk2vtk(img):
+def sitk2vtk(img, deep=1):
     """Convert SimpleITK image to VTK image suitable for MRML volume nodes.
+
+    Input: Image (SimpleITK.Image), deep copy (integer bool, default True)
 
     Note: Returns vtkImageData with origin=(0,0,0), spacing=(1,1,1), and identity direction.
     Spatial information should be set on the MRML volume node via IJKToRASMatrix.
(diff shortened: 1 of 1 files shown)


Base/Python/sitkUtils.py
```diff
@@ -123,9 +125,11 @@ def vtk2sitk(vtkimg, origin=None, spacing=None, direction=None):
     return sitkimg
 
 
-def sitk2vtk(img):
+def sitk2vtk(img, deep=1):
```
lassoan, line 128: It should be Bool.
Since the method name does not include "copy", it would be more clear to name the argument "deepCopy.

```suggestion
def sitk2vtk(img, deepCopy=True):
```

Base/Python/sitkUtils.py
```diff
@@ -146,7 +150,7 @@ def sitk2vtk(img):
     vtk_image.SetOrigin(0.0, 0.0, 0.0)
 
     # Set pixel data
-    depth_array = vtknp.numpy_to_vtk(npdata.ravel())
+    depth_array = vtknp.numpy_to_vtk(npdata.ravel(), deep=deep)
```
lassoan, line 153: Are there only two possible states - deep or shallow? If yes, then we could change this to something more explicit, such as 1 if deepCopy else 0

lassoan (commented):
Looks good, thank you, just added a few nitpicks.

Base/Python/sitkUtils.py
jamesobutler: I did this for consistency of passing the variable to the `numpy_to_vtk` method which has the input argument with name `deep` which is an integer type (https://github.com/Kitware/VTK/blob/7c0494a68bff379d32d6b1fbaa3d10d27a73af54/Wrapping/Python/vtkmodules/util/numpy_support.py#L104). Indeed if the VTK method used the term `deepCopy` then I would have used `deepCopy` here as well.

With the specified code suggestion it would become the following adhering to the matching input argument type used in  the vtk method:

`depth_array = vtknp.numpy_to_vtk(npdata.ravel(), deep=int(deepCopy))`

^This I thought added slight differences which wasn't worth the confusion with the main VTK method being used in this Slicer sitkUtils method.

Base/Python/sitkUtils.py
jamesobutler: Yes only two states.

Yes, it could become the following though as mentioned in https://github.com/Slicer/Slicer/pull/8930#discussion_r2650222386, it seemed to add an extra layer of difference compared to the main VTK method in use that I didn't think was worth the extra complexity here.

`depth_array = vtknp.numpy_to_vtk(npdata.ravel(), deep=1 if deepCopy else 0)`

PR #8931, opened 2025-12-29
ENH: Make transform visualization 2D glyphs thickness and arrow tip length editable on GUI

lassoan:
This PR makes "line thickness" and "arrow tip length" parameters editable in the transforms module GUI. Larger thickness value makes the 2D glyphs thicker, which can be useful for example for creating screenshots that are scaled to smaller size. Shorter arrow tips can help with making the visualization a bit less busy.

image


jamesobutler (dismissed):

lassoan:
@jamesobutler Thanks for the quick review. I've just added configuration of the tip length as well.

PR #8934, opened 2025-12-30
ENH: Make 2D circle glyph resolution adjustable for transform display

lassoan:
Also simplified the GUI layout of qMRMLTransformDisplayNodeWidget.


lassoan:
Thanks @jamesobutler for the review and merge.

PR #8942, opened 2026-01-04
BUG: Fix crash in ExportSegmentsToColorTableNode

lassoan:
Make sure that ExportSegmentsToColorTableNode does not crash if color table is not set.


cpinter:
LGTM! (cannot approve)

PR #8944, opened 2026-01-04
ENH: Replace deprecated AddActor2D/RemoveActor2D usage

jamesobutler:
As of VTK 9.5.0 (per https://github.com/Kitware/VTK/commit/8ec4ae2050c555d1894e2a90029316be4e1ffbea), AddActor2D and RemoveActor2D are deprecated and AddViewProp and RemoveViewProp should be used instead.

- [x] https://github.com/commontk/CTK/pull/1338 should be integrated first
- [x] Update CTK git hash once above is integrated and restore repository to commontk

```
CTK:
$ git shortlog 3811bca..3845903
James Butler (1):
      ENH: Replace deprecated AddActor2D method
```

### DIFF
--- a/SuperBuild/External_CTK.cmake
+++ b/SuperBuild/External_CTK.cmake
@@ -78,7 +78,7 @@ if(NOT DEFINED CTK_DIR AND NOT Slicer_USE_SYSTEM_${proj})
 
   ExternalProject_SetIfNotDefined(
     Slicer_${proj}_GIT_TAG
-    "3845903f333b50ff64db907fd4ae5a6201efc69f"
+    "db2cfae9ed46b48b43d291ee9758a219c87f89f2"
     QUIET
     )
 


Modules/Loadable/Segmentations/EditorEffects/qSlicerSegmentEditorAbstractEffect.cxx
jamesobutler, line 411: This method was originally named based on the main VTK method being used here (`AddActor2D`). Now that the VTK method is deprecated and we are switching to VTK's `AddViewProp` we could consider having Slicer create a new similarly named method `qSlicerSegmentEditorAbstractEffect::addViewProp` and deprecate this named one. This PR doesn't do that, but just a consideration to post here.

jamesobutler:
I've pushed the additional CTK update here to resolve the various issues with QFlags that caused several automated tests to fail recently.

Modules/Loadable/Segmentations/EditorEffects/qSlicerSegmentEditorAbstractEffect.cxx
lassoan: I agree. We don't need to change this method name now but when we touch something in `qSlicerSegmentEditorAbstractEffect` then we can deprecate `AddActor2D` and add `AddViewProp` instead.

lassoan (approved):
It looks good to me, thank you!

Issue #5944, opened 2021-10-12
Unable to build from source on Apple silicon


kliron:
## Steps to reproduce

I followed all the instructions from the [docs](https://slicer.readthedocs.io/en/latest/developer_guide/build_instructions/macos.html#prerequisites) to build slicer from source. 

Build halts with the following error while trying to build DCMTK:

```
make[2]: *** [DCMTK-prefix/src/DCMTK-stamp/DCMTK-configure] Error 1
make[1]: *** [CMakeFiles/DCMTK.dir/all] Error 2
-- OpenSSL: Errors detected - See below.

ld: warning: double slash removed from -install_name (/Users/kliron/opt/slicer/OpenSSL//libcrypto.1.1.dylib)
ld: warning: The i386 architecture is deprecated for macOS (remove from the Xcode build setting: ARCHS)
ld: warning: ignoring file /Users/kliron/opt/slicer/zlib-install/lib/libz.a, building for macOS-i386 but attempting to link with file built for macOS-arm64
ld: warning: ignoring file /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.3.sdk/usr/lib/libSystem.tbd, missing required architecture i386 in file /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.3.sdk/usr/lib/libSystem.tbd (3 slices)

Undefined symbols for architecture i386:

"__DefaultRuneLocale", referenced from:
      _CONF_parse_list in conf_mod.o
  "___bzero", referenced from:
      _ASN1_BIT_STRING_set_bit in a_bitstr.o
      _asn1_item_embed_new in tasn_new.o
      _addr_strings in b_addr.o
      _mem_ctrl in bss_mem.o
      _BLAKE2b_Final in blake2b.o
      _BLAKE2s_Final in blake2s.o
      _bn_div_fixed_top in bn_div.o
      ...

[...an extensive output of a lot more missing symbols...]

ld: symbol(s) not found for architecture i386
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[4]: *** [libcrypto.dylib] Error 1
make[3]: *** [build_libs] Error 2

CMake Error at /Users/kliron/Projects/slicer/CMake/ExternalProjectForNonCMakeProject.cmake:104 (message):
  OpenSSL: build step failed with exit code '2'.
```

## Expected behavior

A successful build.

## Environment
[...]

PR #8945, opened 2026-01-04
Build oneTBB from source and update to 2022.3.0

### LINKED ISSUE #8505: Build TBB from source as an external project to support all architectures

Slicer currently downloads pre-built binaries of Intel oneTBB (Threading Building Blocks) for use as an external project. These binaries are limited to `x86_64` architectures, which presents challenges when building Slicer on other community-supported architectures, such as `aarch64` (e.g., Apple Silicon or Ubuntu 22.04 on ARM). To work around this, Slicer developers currently have to manually specify an externally built version of TBB, increasing setup complexity and reducing consistency across platforms.

Update the TBB external project to build oneTBB from source for all platforms, using the community-supported method provided by Intel. This will:

* Ensure consistent build behavior across all platforms, including macOS arm64 and Linux aarch64.
* Remove the need to manually specify pre-built binaries or external builds in platform-specific configurations.

> [!WARNING]
> To support Hybrid CPU and NUMA platforms, oneTBB relies on the HWLOC* library. See more details at https://uxlfoundation.github.io/oneTBB/GSG/next_steps.html#enable-hybrid-cpu-and-numa-support

* Continue using pre-built binaries for `x86_64` while relying on externally built libraries for `arm64`. However, this introduces extra maintenance burden and makes the build system more brittle.
* Wait for Intel to officially release arm64 binaries for oneTBB. However, this is not guaranteed, and Slicer benefits from controlling its own dependency builds for long-term sustainability.

## Additional context

* See related discussion from @jamesobutler & @jcfr in previous PR: https://github.com/Slicer/Slicer/pull/8202
* Example build of oneTBB on Ubuntu 22.04 aarch64: https://gist.github.com/jcfr/487f5d846bc86e374969be5565c6d95e

jamesobutler:
`COMP: Build oneTBB from source`
This helps support more architectures as there are not arm64 binaries provided by oneTBB releases.
This closes #8505 and is work towards #5944.

`COMP: Update oneTBB to version 2022.3.0`
This updates from [v2021.5.0](https://github.com/uxlfoundation/oneTBB/releases/tag/v2021.5.0) which was originally released December 22nd 2021 to now using [v2022.3.0](https://github.com/uxlfoundation/oneTBB/releases/tag/v2022.3.0) released October 29th 2025. This provides a collection of fixes and improvements for latest platforms.

Local Testing Results:
Windows:
- Build ✔️ 
- Package ✔️ 
- Test ✔️(no new tests failing)
 
macOS:
- Build ✔️ 
- Package ✔️

### DIFF
--- a/SuperBuild/External_tbb.cmake
+++ b/SuperBuild/External_tbb.cmake
@@ -54,7 +54,7 @@ if((NOT DEFINED TBB_DIR
 
   ExternalProject_SetIfNotDefined(
     Slicer_${proj}_GIT_TAG
-    "v2021.5.0"
+    "v2022.3.0"
     QUIET
     )
 


SuperBuild/External_tbb.cmake
```diff
@@ -53,112 +41,91 @@ if((NOT DEFINED TBB_DIR
...
   if(APPLE)
     ExternalProject_Add_Step(${proj} fix_rpath
       # libtbb
-      COMMAND install_name_tool
-        -id ${TBB_INSTALL_DIR}/${tbb_libdir}/libtbb.12.dylib
-            ${TBB_INSTALL_DIR}/${tbb_libdir}/libtbb.12.dylib
-      COMMAND install_name_tool
-        -id ${TBB_INSTALL_DIR}/${tbb_libdir}/libtbb_debug.12.dylib
-            ${TBB_INSTALL_DIR}/${tbb_libdir}/libtbb_debug.12.dylib
+      COMMAND ${CMAKE_COMMAND} -E chdir ${TBB_LIB_DIR}
```
jcfr: For sake of consistency... for the 'fix_rpath` step I suggest to use approach similar across projects. 

See https://github.com/Slicer/Slicer/blob/main/SuperBuild/External_LibArchive.cmake#L101

jamesobutler: Review the most recent push to see if the changes now match what you expect here.

jamesobutler: Actually my former code was handling whether there was release or debug versions. I could remove the debug versions as well here or revert to include the logic to handle whether release or debug have been created.

Something like:
```
COMMAND test ! -f ${TBB_LIB_DIR}/libtbb.12.dylib || install_name_tool -id ${TBB_LIB_DIR}/libtbb.12.dylib ${TBB_LIB_DIR}/libtbb.12.dylib
COMMAND test ! -f ${TBB_LIB_DIR}/libtbb_debug.12.dylib || install_name_tool -id ${TBB_LIB_DIR}/libtbb_debug.12.dylib ${TBB_LIB_DIR}/libtbb_debug.12.dylib
```

jcfr:
Good point. If both debug and release libraries are systemically built, then having the comment for both is sensible.

If not, how about using genex expression to conditionally append the debug suffix ?

jamesobutler:
My latest change will conditionally run the rpath fix depending on the presence of either the release or debug versions

SuperBuild/External_tbb.cmake
```diff
@@ -53,112 +41,85 @@ if((NOT DEFINED TBB_DIR
...
-    BUILD_COMMAND ""
-    INSTALL_COMMAND ""
+    GIT_REPOSITORY "${Slicer_${proj}_GIT_REPOSITORY}"
+    GIT_TAG "${Slicer_${proj}_GIT_TAG}"
+    SOURCE_DIR ${EP_SOURCE_DIR}
+    BINARY_DIR ${EP_BINARY_DIR}
+    CMAKE_CACHE_ARGS
+      -DCMAKE_CXX_COMPILER:FILEPATH=${CMAKE_CXX_COMPILER}
+      -DCMAKE_C_COMPILER:FILEPATH=${CMAKE_C_COMPILER}
+      -DCMAKE_BUILD_TYPE:STRING=${CMAKE_BUILD_TYPE}
```
jcfr: Usually, this is only specified for single config generator.

jamesobutler: Would you like all debug and release versions to be built? Or only the version of the specified build type?

jcfr: It would be only the selected one.

On Unix, it would be by passing the `CMAKE_BUILD_TYPE` option conditionally.

On windows, when using visual Studio... the build type is just specified differently (selected at build time)

jamesobutler: I've updated with the expectation of a single build type that adds the debug suffix if it is a debug build using a generator expression.

jamesobutler:
@jcfr I've retested latest changes on macOS (as well as Windows) and observe build and packaging success for both platforms.

SuperBuild/External_tbb.cmake
```diff
@@ -53,112 +41,80 @@ if((NOT DEFINED TBB_DIR
...
+    Slicer_${proj}_GIT_TAG
+    "v2022.3.0"
+    QUIET
+    )
+
+  set(EXTERNAL_PROJECT_OPTIONAL_CMAKE_CACHE_ARGS)
+
+  # Pass architecture-specific flags
+  if(APPLE)
+    list(APPEND EXTERNAL_PROJECT_OPTIONAL_CMAKE_CACHE_ARGS
```
jcfr: These lines can be removed as the corresponding options are already associated with external projects here. See   https://github.com/Slicer/Slicer/blob/fa5d13972d3f6b00309c3a6cd1aae68250de8e63/CMakeLists.txt#L62

SuperBuild/External_tbb.cmake
```diff
@@ -53,112 +41,80 @@ if((NOT DEFINED TBB_DIR
...
-    BUILD_IN_SOURCE 1
-    CONFIGURE_COMMAND ""
-    BUILD_COMMAND ""
-    INSTALL_COMMAND ""
+    GIT_REPOSITORY "${Slicer_${proj}_GIT_REPOSITORY}"
+    GIT_TAG "${Slicer_${proj}_GIT_TAG}"
+    SOURCE_DIR ${EP_SOURCE_DIR}
+    BINARY_DIR ${EP_BINARY_DIR}
+    CMAKE_CACHE_ARGS
+      -DCMAKE_CXX_COMPILER:FILEPATH=${CMAKE_CXX_COMPILER}
```
jcfr, line 69: Consider adding comments like `# Compiler settings`, `# Options`, `# Install directories`

Examples:
* https://github.com/Slicer/Slicer/blob/main/SuperBuild%2FExternal_LibFFI.cmake
* https://github.com/Slicer/Slicer/blob/main/SuperBuild%2FExternal_bzip2.cmake

SuperBuild/External_tbb.cmake
```diff
@@ -53,112 +41,80 @@ if((NOT DEFINED TBB_DIR
...
+      -DCMAKE_C_COMPILER:FILEPATH=${CMAKE_C_COMPILER}
+      -DCMAKE_BUILD_TYPE:STRING=${CMAKE_BUILD_TYPE}
```
jcfr: You should remove the explicit passing of `CMAKE_BUILD_TYPE`, this is already handled at https://github.com/Slicer/Slicer/blob/main/CMake%2FSlicerInitializeBuildType.cmake

PR #8949, opened 2026-01-06
COMP: Enforce Qt 5.15 minimum across all platforms 

jamesobutler:
Following the drop of documented support for Ubuntu 20.04 (https://github.com/Slicer/Slicer/commit/1d4d0612d78633931dfe3a8bb4fadd8733fb77ed), the minimum Qt version can be enforced to Qt 5.15 across all platforms. This has already been applied for Windows and macOS.

Ubuntu 22.04 comes with Qt 5.15.3.
Ubuntu 24.04 comes with Qt 5.15.10
Ubuntu 25.04 comes with Qt 5.15.15

Based on today Slicer's hangout with discussion of keeping around some Qt5 support, we can at least bump the minimum to be consistent across all platforms where Qt 5.15 can be the minimum where Ubuntu 22.04 is shipped with Qt 5.15. The minimum for Linux had been Qt 5.12 because that is what was shipped with Ubuntu 20.04.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -188,17 +188,7 @@ mark_as_superbuild(Patch_EXECUTABLE)
 # Qt requirements
 #-----------------------------------------------------------------------------
 if(NOT DEFINED Slicer_REQUIRED_QT_VERSION)
-  # Official builds and tests are performed with Qt-5.15 on all platforms,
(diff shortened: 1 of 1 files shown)


RafaelPalomar (approved):
LGTM

Issue #6452, opened 2022-06-28
Add reusable widget for markups control point list


lassoan:
## Is your feature request related to a problem? Please describe.

qSlicerSimpleMarkupsWidget does not support any of the new features that are available in the Markups module's control point table widget.

## Describe the solution you'd like

We would need to create a new qSlicerMarkupsControlPointTableView+Model that could replace both these widgets.

## Additional context

I've started to work on this at some point but then ran out of time. The development branch is available here: https://github.com/lassoan/Slicer/tree/attic/markups-control-points-table-view

PR #8955, opened 2026-01-10
ENH: Add position status column to qSlicerSimpleMarkupsWidget

lassoan:
The new column is hidden by default. Its visibility is controlled by the positionStatusColumnVisible property.

Requested in https://discourse.slicer.org/t/include-functionality-of-control-point-status-in-custom-scripted-module/45658

related to #6452

Example:

```
w=slicer.qSlicerSimpleMarkupsWidget()
w.setMRMLScene(slicer.mrmlScene)
w.setPositionStatusColumnVisible(True)
w.show()
```

image


lassoan:
I haven't added test (it would be difficult to check on the GUI if the column appears/disappears and values are displayed and correct and we do limited testing at GUI level) but addressed the other copilot comments.

PR #8957, opened 2026-01-13
COMP: Skip PIC flag passed to MSVC

jamesobutler:
The -fPIC flag is not available in the MSVC compiler. This commit avoids the warnings such as:

cl : command line warning D9002: ignoring unknown option '-fPIC'

### DIFF
--- a/Libs/vtkSegmentationCore/CMakeLists.txt
+++ b/Libs/vtkSegmentationCore/CMakeLists.txt
@@ -113,10 +113,7 @@ include_directories( ${vtkSegmentationCore_INCLUDE_DIRS} )
 add_library(${PROJECT_NAME} ${vtkSegmentationCore_SRCS})
 target_link_libraries( ${PROJECT_NAME} ${vtkSegmentationCore_LIBS} )
 
-if (CMAKE_SYSTEM_PROCESSOR STREQUAL "x86_64" AND NOT WIN32)
-  set_target_properties(${PROJECT_NAME} PROPERTIES COMPILE_FLAGS "-fPIC")
-endif()
-
+set_target_properties(${PROJECT_NAME} PROPERTIES POSITION_INDEPENDENT_CODE ON)
 # Apply user-defined properties to the library target.
 if(Slicer_LIBRARY_PROPERTIES)
   set_target_properties(${PROJECT_NAME} PROPERTIES ${Slicer_LIBRARY_PROPERTIES})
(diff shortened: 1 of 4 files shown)


Libs/vtkSegmentationCore/CMakeLists.txt
```diff
@@ -113,8 +113,11 @@ include_directories( ${vtkSegmentationCore_INCLUDE_DIRS} )
 add_library(${PROJECT_NAME} ${vtkSegmentationCore_SRCS})
 target_link_libraries( ${PROJECT_NAME} ${vtkSegmentationCore_LIBS} )
 
-if (CMAKE_SYSTEM_PROCESSOR STREQUAL "x86_64" AND NOT WIN32)
-  set_target_properties(${PROJECT_NAME} PROPERTIES COMPILE_FLAGS "-fPIC")
+if (CMAKE_SYSTEM_PROCESSOR STREQUAL "x86_64")
```
lassoan: At all the other places in this PR we just decided based on the result of `check_cxx_compiler_flag(-fPIC...)`.
Why do we use a different pattern here?

jcfr (changes requested):
Since this code was put together before the introduction of `POSITION_INDEPENDENT_CODE`, I suggest to revisit and avoid directly updating compiler flags. 

https://cmake.org/cmake/help/latest/prop_tgt/POSITION_INDEPENDENT_CODE.html

lassoan:
> Since this code was put together before the introduction of POSITION_INDEPENDENT_CODE, I suggest to revisit and avoid directly updating compiler flags.

This can be done for the segmentation library that is in the Slicer repository.

However, how would this work for external projects (curl, lzma, sqlite)? We cannot set target properties from outside (from the superbuild).

jcfr:
> cannot set target properties from outside (from the superbuild).

Setting the corresponding option should work ... https://cmake.org/cmake/help/latest/variable/CMAKE_POSITION_INDEPENDENT_CODE.html

lassoan:
> > cannot set target properties from outside (from the superbuild).
> 
> Setting the corresponding option should work ... https://cmake.org/cmake/help/latest/variable/CMAKE_POSITION_INDEPENDENT_CODE.html

OK, this could work, as we can push variable values to the external projects via CMAKE_CACHE_ARGS.

@jamesobutler could you give this a try? (enable `CMAKE_POSITION_INDEPENDENT_CODE` for the 3 external projects similarly to how it is already done for [zlib](https://github.com/Slicer/Slicer/blob/83e162ff233e9bcd200f6cec08d9176da095e4b1/SuperBuild/External_zlib.cmake#L53))

It seems to me that `POSITION_INDEPENDENT_CODE` is enabled by default for shared libraries in CMake versions that we support, so we can probably simply remove the manual setting of `-fPIC` for `vtkSegmentationCore`.

jamesobutler:
@hjmjohnson I see that you rebased this branch against latest `main`. Do you have review thoughts on this proposed change in its current state?

PR #8861, opened 2025-11-19
Add Dev Container setup for streamlining local development on Linux

ruffsl:
This pull request introduces a comprehensive Dev Container setup for the project, enabling reproducible development and build environments using Docker and BuildKit. It adds multi-stage Dockerfile instructions, configuration files for caching and environment, a detailed README for onboarding, and scripts for initializing the container. The changes are designed to streamline building, testing, and packaging workflows, and to support advanced development scenarios such as hardware acceleration and debugging.

**Dev Container Infrastructure**

* Added a multi-stage `.devcontainer/Dockerfile` that defines separate stages for building, testing, packaging, and running Slicer, with extensive use of BuildKit cache mounts and environment configuration for efficient development.
* Introduced `.devcontainer/devcontainer.json` to configure the Dev Container environment, including image selection, runtime arguments for hardware and debugging, workspace mounts, environment variables, and recommended extensions.

**Developer Experience**

* Added `.devcontainer/README.md` with step-by-step instructions for setting up Docker, Git LFS, NVIDIA toolkit, and development tools, as well as guidance for building, running, and launching the Dev Container using CLI, VSCode, CodeSpaces, etc.

Issue #6388, opened 2022-05-19
Update Slicer to use Qt6


lassoan:
Users are running into issues that are not going to be fixed on Qt5, therefore, we need to switch to Qt6.

Issues reported by users that would be likely fixed by moving to Qt6:
- Proxy issues on mac (https://discourse.slicer.org/t/proxy-issue-in-extensions-manager/23467)

Unfortunately, Qt6 considerably raises minimum software and hardware requirements - see https://github.com/Slicer/Slicer/pull/6108

jamesobutler:
#6237 and #6314 are some of the latest PRs with work towards this goal.

Use https://doc.qt.io/qt-6/portingguide.html as guidance.

Some Action items:
- [x] Update PythonQt with Qt6 support https://github.com/MeVisLab/pythonqt/issues/13
- [x] Update CTK with Qt6 support https://github.com/commontk/CTK/issues/947

jcfr:
## CTK updates


For future reference, the following pull requests fix deprecation warnings and are step toward supporting Qt6 and beyond:
* https://github.com/commontk/CTK/pull/1083
* https://github.com/commontk/CTK/pull/1086
* https://github.com/commontk/CTK/pull/1087
* https://github.com/commontk/CTK/pull/1089
* https://github.com/commontk/CTK/pull/1097
* https://github.com/commontk/CTK/pull/1098
* https://github.com/commontk/CTK/pull/1096
* https://github.com/commontk/CTK/issues/1104 (last release to support Qt4/C++98/VTK8)
* https://github.com/commontk/CTK/pull/1005
* https://github.com/commontk/CTK/pull/1112
* https://github.com/commontk/CTK/pull/1108
* https://github.com/commontk/CTK/pull/1113

In the process also integrated the following:
* https://github.com/commontk/CTK/pull/1088
* https://github.com/commontk/CTK/pull/1084
* https://github.com/commontk/CTK/pull/1085

### Updated dependencies

* PythonQt
  * https://github.com/commontk/CTK/pull/1261

JulienMaille:
Hello there (it's quite fun to stumble upon a fellow CPEen 👋🏼 @jcfr)
[...]

PR #8959, opened 2026-01-14
COMP: Update cURL version to 8.17.0

jamesobutler:
This change is needed to support building against OpenSSL 3.

cURL 7.70.0 was originally released 2020-04-29. This commit updates to 8.17.0 released 2025-11-05. This update provides support for OpenSSL 3 which upon initial search appears to have been introduced/supported beginning in cURL 7.77.0.

Latest cURL 8.18.0 actually bumps  minimum OpenSSL version to 3, but since Slicer doesn't build OpenSSL 3 yet, we will move to this more recent version that supports both OpenSSL 1 and 3.

Note that many of the cURL build options switched from names like "CMAKE_USE_" to "CURL_USE_" based on the change in https://github.com/curl/curl/commit/9108da2c26d18e927b91e33d3729d9cf0f3eb8fa first included in cURL 7.81.0.

Testing results - builds were successful on both Windows and macOS platforms.

### DIFF
--- a/Libs/RemoteIO/CMakeLists.txt
+++ b/Libs/RemoteIO/CMakeLists.txt
@@ -92,10 +92,22 @@ set(srcs ${RemoteIO_SRCS})
 
 add_library(${lib_name} ${srcs})
 
+# Define CURL_STATICLIB when linking against static curl on Windows
+if(WIN32)
+  target_compile_definitions(${lib_name} PRIVATE CURL_STATICLIB)
+endif()
+
 set(libs
   ${CURL_LIBRARIES}
   MRMLCore
   )
+if(WIN32)
+  list(APPEND libs
+    # For curl
+    iphlpapi  # for if_nametoindex
+    ws2_32  # for Windows sockets
+    )
+endif()
 if(Slicer_USE_PYTHONQT_WITH_OPENSSL)
   list(APPEND libs
     ${OPENSSL_LIBRARIES}
--- a/SuperBuild/External_curl.cmake
+++ b/SuperBuild/External_curl.cmake
@@ -52,13 +52,13 @@ if((NOT DEFINED CURL_INCLUDE_DIR
 
   ExternalProject_SetIfNotDefined(
     Slicer_${proj}_GIT_REPOSITORY
-    "${EP_GIT_PROTOCOL}://github.com/Slicer/curl.git"
+    "${EP_GIT_PROTOCOL}://github.com/curl/curl.git"
     QUIET
     )
 
   ExternalProject_SetIfNotDefined(
     Slicer_${proj}_GIT_TAG
-    "d73e360a78d97adda85364e6bd5c504a2eb1572a" # slicer-7.70.0-2020-04-29-53cdc2c
+    "curl-8_17_0"
     QUIET
     )
 
@@ -89,7 +89,7 @@ if((NOT DEFINED CURL_INCLUDE_DIR
       -DCMAKE_INSTALL_LIBDIR:STRING=lib  # Override value set in GNUInstallDirs CMake module
       -DBUILD_TESTING:BOOL=OFF
       -DBUILD_CURL_EXE:BOOL=OFF
-      -DBUILD_SHARED_LIBS:BOOL=OFF  # Before enabling this option, see https://github.com/Slicer/curl/commit/ca5fe8e63df7faea0bfb988ef3fe58f538e6950b
+      -DBUILD_SHARED_LIBS:BOOL=OFF  # Before enabling this option, review CURL_STATICLIB in CFLAGS where cURL is used
       -DENABLE_ARES:BOOL=OFF
       -DCURL_ZLIB:BOOL=ON
       -DZLIB_INCLUDE_DIR:PATH=${ZLIB_INCLUDE_DIR}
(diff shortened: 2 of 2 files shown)


ruffsl:
After re-basing my draft PR with a prior working dev environment using Ubuntu, the build now fails with:

```
$ make build
...
/usr/bin/ld: /opt/slicer_ws/build/Slicer-SuperBuild-Debug/curl-install/lib/libcurl.a(content_encoding.c.o): in function `brotli_do_close':
content_encoding.c:(.text+0x65): undefined reference to `BrotliDecoderDestroyInstance'
/usr/bin/ld: /opt/slicer_ws/build/Slicer-SuperBuild-Debug/curl-install/lib/libcurl.a(content_encoding.c.o): in function `brotli_do_write':
content_encoding.c:(.text+0x161): undefined reference to `BrotliDecoderDecompressStream'
/usr/bin/ld: content_encoding.c:(.text+0x199): undefined reference to `BrotliDecoderDestroyInstance'
/usr/bin/ld: content_encoding.c:(.text+0x1e8): undefined reference to `BrotliDecoderGetErrorCode'
/usr/bin/ld: /opt/slicer_ws/build/Slicer-SuperBuild-Debug/curl-install/lib/libcurl.a(content_encoding.c.o): in function `brotli_do_init':
content_encoding.c:(.text+0x24f): undefined reference to `BrotliDecoderCreateInstance'
collect2: error: ld returned 1 exit status
make[5]: *** [Libs/RemoteIO/CMakeFiles/RemoteIO.dir/build.make:103: bin/libRemoteIO.so] Error 1
make[4]: *** [CMakeFiles/Makefile2:8280: Libs/RemoteIO/CMakeFiles/RemoteIO.dir/all] Error 2
make[4]: *** Waiting for unfinished jobs....
...
```

I ended up having to add a few CMake flags to ensures curl was built without Brotli support to pass the build: 

- https://github.com/Slicer/Slicer/pull/8861
- https://github.com/Slicer/Slicer/pull/8861/commits/4b12447667e7b56dbb7d7011c6c19a9068614606

https://github.com/Slicer/Slicer/blob/4b12447667e7b56dbb7d7011c6c19a9068614606/SuperBuild/External_curl.cmake#L100-L105

Should we upstream this for other systems where curl auto detects Brotli, or fix the super build linking if Brotli is used?

---

Edit: I'm also seeing some runtime OpenSSL issues now. Not sure how the super build even compiled with a mismatch?

```
ubuntu@wox /o/s/b/Slicer-SuperBuild-Debug> ./Slicer-build/Slicer
qt.network.ssl: Incompatible version of OpenSSL (built with OpenSSL >= 3.x, runtime version is < 3.x)
[SSL] SSL support disabled - Failed to load SSL library !
[SSL] Failed to load Slicer.crt

Switch to module:  "Welcome"
Sandboxing disabled by user.
An OpenGL Core Profile was requested, but it is not supported on the current platform. Falling back to a non-Core profile. Note that this might cause rendering issues.
QSslSocket::connectToHostEncrypted: TLS initialization failed
Failed to download extension metadata from server
Uncaught ReferenceError: app is not defined
Property 'verbose'' of object 'qSlicerWebPythonProxy' has no notify signal and is not constant, value updates in HTML will be broken!
Property 'verbose'' of object 'qSlicerWebPythonProxy' has no notify signal and is not constant, value updates in HTML will be broken!
```

jamesobutler:
@ruffsl I think it is ok to issue a PR with the options to turn off Brotli support to help with successful building. It is interesting these build issues haven't come up during the Linux nightly build (https://slicer.cdash.org/index.php?project=SlicerPreview), but that is using AlmaLinux8 and not using the Ubuntu version that you are using.

Yes, OpenSSL 3 is the next thing to update. If you have the time we need to update https://github.com/Slicer/Slicer/blob/main/SuperBuild/External_OpenSSL.cmake. We build OpenSSL from source for Linux and macOS, but it may be good to equally build from source for the Windows platform which is currently pulling pre-built binaries from https://github.com/Slicer/Slicer-OpenSSL/releases. This will support Qt6 binaries which are built against OpenSSL 3 while Qt5 binaries were against OpenSSL 1.1.x which is what the Slicer superbuild currently build. Were you by chance also using Qt6 here in your builds?

ruffsl:
> but that is using AlmaLinux8 and not using the Ubuntu version that you are using.

Yeah, my guess is nothing yet in CI's AlmaLinux8 environment ropes in Brotli, where as even my minimal Ubuntu does.

> I think it is ok to issue a PR with the options to turn off Brotli support to help with successful building.

- https://github.com/Slicer/Slicer/pull/9007

> Yes, OpenSSL 3 is the next thing to update. If you have the time we need to update
> it may be good to equally build from source for the Windows platform

Oh boy, that looks like a big lift in CMake...
Also, don't have a Windows environment ATM, and not sure I'd ever want to go back.
But is there a ticket I should subscribe to? Didn't spot anything recent, other than the general:

- https://github.com/Slicer/Slicer/issues/6388

> This will support Qt6 binaries which are built against OpenSSL 3 while Qt5 binaries were against OpenSSL 1.1.x which is what the Slicer superbuild currently build. Were you by chance also using Qt6 here in your builds?

Oh, that's a good point, as I'm still using Qt5. My development environment is entirely declarative via the Dockerfile:

https://github.com/Slicer/Slicer/blob/4b12447667e7b56dbb7d7011c6c19a9068614606/.devcontainer/Dockerfile#L76-L84

https://github.com/Slicer/Slicer/blob/4b12447667e7b56dbb7d7011c6c19a9068614606/.devcontainer/Dockerfile#L111-L122 

Is the Qt6 build and runtime dependency set for Slicer documented anywhere yet? The developer guide is still Qt5 only.

https://github.com/Slicer/Slicer/blob/c6a6e4eb7aaa6be318b18438757d26e3f8510cfd/Docs/developer_guide/build_instructions/linux.md?plain=1#L27

jamesobutler:
Yes, the OpenSSL upgrade work isn't explicitly mentioned anywhere yet, but is related to https://github.com/Slicer/Slicer/issues/6388. Following along there will provide you updates.

Qt6 instructions have so far been added to the Windows build instructions page, but not yet for the other platforms. I for example have successfully built Slicer with Qt 6.10.1 and latest `main` code. Qt6 is still in the experimental stage for Slicer until we get everything working cross platform. Then we will update the Factory build machines to use Qt6 for the nightly Slicer Preview builds.

PR #8960, opened 2026-01-14
Fix typing issue: Update about.md

benediktjohannes:
computatiton → computation


benediktjohannes:
You're welcome, thanks for the quick answer and the friendly tone! 👍 I've helped many projects fixing typing issues, but many were quite unfriendly, so this is very nice of you, thanks! 👍

lassoan:
Thank you, the fix is included in https://github.com/Slicer/Slicer/commit/055e293a3e5b28c466017242650342cb7f72e3fa

PR #8963, opened 2026-01-14
DOC: Fix spelling error in the word "supercede" -> "supersede"

benediktjohannes:
"Supercede" is a common misspelling of "supersede," the correct word meaning "to replace" or "to displace." Although the spelling "supercede" has existed for centuries and appears in some dictionaries as an archaic or colloquial variant, it is considered an error because the original comes from the Latin supersedere, meaning "to sit above," which also contains an "s" in the stem.

(I'm sorry for interrupting you again, but I just found this when installing the app and thought I have to fix this and I've learned a little bit because I know added "DOC:".)

### DIFF
--- a/License.txt
+++ b/License.txt
@@ -194,6 +194,6 @@ PART C. MISCELLANEOUS
 
 This Agreement shall be governed by and construed in accordance with
 the laws of The Commonwealth of Massachusetts without regard to
-principles of conflicts of law. This Agreement shall supercede and
+principles of conflicts of law. This Agreement shall supersede and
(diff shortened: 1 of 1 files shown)


lassoan:
Thank you, the PR is good (DOC: prefix, etc.), but it impacts a legal document that we cannot change, as it would become inconsistent with official reviewed, approved, archived text of the license (e.g., at https://spdx.github.io/license-list-data/3D-Slicer-1.0.html). But feel free to keep sending corrections if you find issues elsewhere.

PR #8964, opened 2026-01-14
BUG: Fix typo in dicom thumbnail size label

benediktjohannes:
Not quite sure if "Thumbnail size" is correct.

Fixed typo in "Thumbnail size" label (changed to "Thumbnails' sizes", but not quite sure about this one, please correct me if I'm mistaken, thanks).

### DIFF
--- a/Modules/Scripted/DICOM/DICOM.py
+++ b/Modules/Scripted/DICOM/DICOM.py
@@ -614,7 +614,7 @@ class _ui_DICOMSettingsPanel:
         thumbnailsSizeComboBox.addItem(_("Medium"), "medium")
         thumbnailsSizeComboBox.addItem(_("Large"), "large")
         thumbnailsSizeComboBox.currentIndex = 1
-        genericGroupBoxFormLayout.addRow(_("Thumbnails size:"), thumbnailsSizeComboBox)
+        genericGroupBoxFormLayout.addRow(_("Thumbnail size:"), thumbnailsSizeComboBox)
         parent.registerProperty(
             "DICOM/thumbnailsSize", thumbnailsSizeComboBox,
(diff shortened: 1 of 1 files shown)


benediktjohannes:
Oh, wait a moment, this is not a documentation, but something inside the software itself (and I guess that this "DOC" thing is not for the documentation of any kind of changes in the whole software, but just for changes in the documentation (and it doesn't tell you about the fact that you document a change in general)). Unforetunately, I'm pretty new to GitHub and I'm not quite sure how to fix this, sorry, but thanks! 👍

benediktjohannes:
Hi! Sorry for the long delay in replying, I just saw your message and completely forgot to respond. Unfortunately, I don’t think there’s a quick configuration change that would allow you to detect those typos. I use my own scripts (originally written for Wikipedia and later adapted for GitHub), but they currently have no GUI and I’m essentially the only one who can manage them. I might publish them in a few months, but they still need significant improvement, as they’re not very user-friendly and can produce many bugs if you’re not familiar with the code (which, at the moment, only I fully am). So unfortunately, they probably won’t be of much help to you right now. But thanks for the question!

lassoan:
> Unfortunately, I don’t think there’s a quick configuration change that would allow you to detect those typos

Why do you think so? Codespell found hundreds of errors and it is surprising that it missed some. It would be interesting to investigate why codespell missed these typos. It could help you deciding whether it is better to report (and maybe contribute fixes) to codespell or to develop an entirely new solution. If you find that codespell is fundamentally flawed (and codespell developers don't offer a reasonable plan to address it) then introducing a new tool might be justified; but most likely improving codespell would be sufficient. It is also possible that we just not use codespell optimally in the Slicer project or we use an outdated version.

PR #8966, opened 2026-01-15
COMP: Update urllib3 version to latest 2.6.3

jamesobutler:
This satisfies GitHub advisories warned in the Security section of the Slicer repository:
- https://github.com/advisories/GHSA-2xpw-w6gg-jr37
- https://github.com/advisories/GHSA-gm62-xv2j-4w53
- https://github.com/advisories/GHSA-38jv-5279-wg99

This will boost the Slicer OpenSSF Scorecard report which is currently getting docked a few points for these warned advisories.


jcfr (approved):

jamesobutler:
I realized I actually need to update the following to fully address the advisory to bump the OpenSSF score. I’ll follow up with another PR.
https://github.com/Slicer/Slicer/blob/21cf6a6522854c150d16cdbab39a721c85b549f1/requirements-docs.txt#L15

PR #8972, opened 2026-01-18
ENH: Remove support for building against VTK <= 9.3

jamesobutler:
This anticipates upcoming support for VTK 9.6.0 at which there will be support for VTK 9.4, 9.5 and 9.6.

This is a follow-up to https://github.com/Slicer/Slicer/commit/dfde1cfc68f2d8d6da6678fb0a22fa4dd184288c where the VTK minimum version was last updated to 9.2.


lassoan (approved):
Thanks, it looks good to me.

PR #8975, opened 2026-01-19
BUG: Fix Qt6 data dialog description shift-change

codeling:
When building with Qt6, opening the data dialog and adding any file, there is a message in the output:

```
QObject::connect: No such signal QComboBox::activated(QString) in C:\s\src\Base\QTGUI\qSlicerDataDialog.cxx:254
QObject::connect:  (receiver name: 'qSlicerDataDialog')
```
(beceause the `QString` version was removed with Qt6, there now is only an `int` version of that signal).

The result is that the "Description" ComboBox change with Shift clicked (which in previous builds, changed the type for all open datasets) does not work.

This pull request fixes the signal connection.

### DIFF
--- a/Base/QTGUI/qSlicerDataDialog.cxx
+++ b/Base/QTGUI/qSlicerDataDialog.cxx
@@ -251,7 +251,7 @@ void qSlicerDataDialogPrivate::addFile(const QFileInfo& file, const QString& rea
 #else
   QObject::connect(descriptionComboBox, SIGNAL(currentIndexChanged(QString)), this, SLOT(onFileTypeChanged(QString)));
 #endif
-  QObject::connect(descriptionComboBox, SIGNAL(activated(QString)), this, SLOT(onFileTypeActivated(QString)));
+  QObject::connect(descriptionComboBox, &QComboBox::textActivated, this, &qSlicerDataDialogPrivate::onFileTypeActivated);
   this->FileWidget->setCellWidget(row, TypeColumn, descriptionComboBox);
   descriptionComboBox->setCurrentIndex(0);
   if (!readerDescription.isEmpty() && ioProperties != nullptr)


Base/QTGUI/qSlicerDataDialog.cxx
```diff
@@ -391,7 +396,10 @@ void qSlicerDataDialogPrivate::onFileTypeActivated(const QString& description)
         continue;
       }
       QComboBox* selectedComboBox = qobject_cast<QComboBox*>(this->FileWidget->cellWidget(row, TypeColumn));
+#if QT_VERSION >= QT_VERSION_CHECK(5, 15, 0)
+#else
       int descriptionIndex = selectedComboBox ? selectedComboBox->findText(description) : -1;
+#endif
```
jamesobutler: We should no longer have code to support Qt versions < 5.15. Qt 5.15 was set as the minimum required versions across all platforms recently as of https://github.com/Slicer/Slicer/commit/bfc9f1356c05d6fdc1f4d88943231a969732a808. In #8977 I have further made steps to remove existing code handling of Qt versions < 5.15.

codeling: great, simplifies this fix some more. I'll change it accordingly!

Base/QTGUI/qSlicerDataDialog.cxx
```diff
@@ -246,8 +246,12 @@ void qSlicerDataDialogPrivate::addFile(const QFileInfo& file, const QString& rea
   // let's select none, connect the signal and then selecting the first will
   // automatically create the option widget
   descriptionComboBox->setCurrentIndex(-1);
+#if QT_VERSION >= QT_VERSION_CHECK(5, 15, 0)
   QObject::connect(descriptionComboBox, &QComboBox::currentTextChanged, this, &qSlicerDataDialogPrivate::onFileTypeChanged);
-  QObject::connect(descriptionComboBox, SIGNAL(activated(QString)), this, SLOT(onFileTypeActivated(QString)));
+#else
+  QObject::connect(descriptionComboBox, SIGNAL(currentIndexChanged(QString)), this, SLOT(onFileTypeChanged(QString)));
+#endif
+  QObject::connect(descriptionComboBox, &QComboBox::textActivated, this, &qSlicerDataDialogPrivate::onFileTypeActivated);
```
jamesobutler, line 254: Yes textActivated is the correct new signal that replaced the old activated(QString) version. This occurred in https://github.com/qt/qtbase/commit/d8ac4b40b5acb5582b1491f48f3d3c2cfc044666 and has been available since Qt 5.14.

jamesobutler:
@codeling GitHub is currently indicating some merge conflicts to solve. I also noticed your Slicer fork is a fork of JC’s Slicer fork. For clarity, I would suggest re-forking your repo based on the Slicer upstream so that it is easier to tell how many commits behind your branch is relative to the upstream instead of JC’s main branch.

PR #8976, opened 2026-01-19
DOC: Separate configure/build instructions Qt 5/6

codeling:
Provide separate instructions for configuring/building with Qt 5 and Qt 6

### DIFF
--- a/Docs/developer_guide/build_instructions/windows.md
+++ b/Docs/developer_guide/build_instructions/windows.md
@@ -60,6 +60,8 @@ Build takes several hours. Warnings will appear during the build (it is practica
 
 Specify source, build, and Qt location and compiler version and start the build using the following commands (these can be put into a .bat file so that they can be executed again easily), assuming default folder locations:
 
+#### Qt 5
+
 Release mode:
 
 ```bat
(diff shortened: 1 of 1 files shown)


Docs/developer_guide/build_instructions/windows.md
```diff
@@ -86,6 +88,37 @@ cd C:\D
...
+#### Qt 6
+
+Release mode:
+
+```bat
+cd C:\D
+
+"C:\Program Files\CMake\bin\cmake.exe" ^
+  -G "Visual Studio 17 2022" -A x64 ^
+  -DSlicer_REQUIRED_QT_VERSION=6.10 ^
```
jamesobutler, line 101: This is not required. This is the minimum required Qt version which can be maintained as Qt 6.8. Specifying the desired Qt version to use in Qt6_DIR is all that is needed.

https://github.com/Slicer/Slicer/blob/5c803b3deb2019e8c1d5f0ded4278352dcee55d8/CMakeLists.txt#L193

codeling: good to know, thanks!
so the separate sections introduced by the PR are probably a bit overkill then.
Maybe I'll just add a separate example of using Qt6?

PR #8977, opened 2026-01-19
ENH: Remove support for Qt versions < 5.15

jamesobutler:
This is a follow-up to https://github.com/Slicer/Slicer/commit/bfc9f1356c05d6fdc1f4d88943231a969732a808 where Qt 5.15 was defined as the minimum supported version.


lassoan (approved):
Thank you, this all looks good!

PR #8982, opened 2026-01-22
PERF: Fixes performance issues for DICOM import in Qt6

Punzo:
when the copy file option is enabled.
ENH: Cache error log icons and track their state in qSlicerMainWindow to avoid regenerating icons on every log message. ENH: Skip icon updates if the error log state hasn't changed. 
ENH: Automatically disable detailed DICOM logging after 4 hours to avoid accidental performance degradation. 
ENH: Store a timestamp when detailed logging is enabled and check on startup to auto-disable if needed. 
ENH: Set the default thumbnail size to "small" in DICOM settings for better performance. 
ENH: Add a "hidden" option for thumbnail size in DICOM settings.

needs https://github.com/commontk/CTK/pull/1345

### DIFF
--- a/Base/QTApp/qSlicerMainWindow.cxx
+++ b/Base/QTApp/qSlicerMainWindow.cxx
@@ -358,6 +358,15 @@ void qSlicerMainWindowPrivate::setupUi(QMainWindow* mainWindow)
 
   QObject::connect(this->ErrorLogToggleViewAction, SIGNAL(toggled(bool)), q, SLOT(onErrorLogToggled(bool)));
 
+  QIcon defaultIcon = q->style()->standardIcon(QStyle::SP_MessageBoxCritical);
+  this->ErrorLogIconHighlighted = defaultIcon;
+
+  QIcon disabledIcon;
+  disabledIcon.addPixmap(defaultIcon.pixmap(QSize(32, 32), QIcon::Disabled, QIcon::On), QIcon::Active, QIcon::On);
+  this->ErrorLogIconNormal = disabledIcon;
+
+  this->ErrorLogToggleViewAction->setIcon(this->ErrorLogIconNormal);
+
   this->ViewMenu->insertAction(this->ModuleHomeAction, this->ErrorLogToggleViewAction);
 
   // Change orientation depending on where the widget is docked
@@ -699,16 +708,13 @@ void qSlicerMainWindowPrivate::setupStatusBar()
 //-----------------------------------------------------------------------------
 void qSlicerMainWindowPrivate::setErrorLogIconHighlighted(bool highlighted)
 {
-  Q_Q(qSlicerMainWindow);
-  QIcon defaultIcon = q->style()->standardIcon(QStyle::SP_MessageBoxCritical);
-  QIcon icon = defaultIcon;
-  if (!highlighted)
+  if (this->ErrorLogIconIsHighlighted == highlighted)
   {
-    QIcon disabledIcon;
-    disabledIcon.addPixmap(defaultIcon.pixmap(QSize(32, 32), QIcon::Disabled, QIcon::On), QIcon::Active, QIcon::On);
-    icon = disabledIcon;
+    return;
   }
-  this->ErrorLogToggleViewAction->setIcon(icon);
+
+  this->ErrorLogIconIsHighlighted = highlighted;
+  this->ErrorLogToggleViewAction->setIcon(highlighted ? this->ErrorLogIconHighlighted : this->ErrorLogIconNormal);
 }
 
 //-----------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


Base/QTApp/qSlicerMainWindow.cxx
```diff
@@ -699,16 +708,13 @@ void qSlicerMainWindowPrivate::setupStatusBar()
...
   {
-    QIcon disabledIcon;
-    disabledIcon.addPixmap(defaultIcon.pixmap(QSize(32, 32), QIcon::Disabled, QIcon::On), QIcon::Active, QIcon::On);
-    icon = disabledIcon;
+    return;
   }
-  this->ErrorLogToggleViewAction->setIcon(icon);
+
+  this->ErrorLogIconIsHighlighted = highlighted;
+  // this->ErrorLogToggleViewAction->setIcon(highlighted ? this->ErrorLogIconHighlighted : this->ErrorLogIconNormal);
```
lassoan: remove

Modules/Scripted/DICOM/DICOM.py
```diff
@@ -512,6 +517,35 @@ def onBrowserWidgetClosed(self):
...
+        except (ValueError, TypeError):
+            detailedLoggingEnabledTime = 0
+
+        currentTime = time.time()
+        fourHoursInSeconds = 4 * 60 * 60
+
+        # If detailed logging was enabled more than 4 hours ago, disable it
+        if detailedLoggingEnabledTime > 0 and (currentTime - detailedLoggingEnabledTime) > fourHoursInSeconds:
+            logging.info("Detailed DICOM logging was enabled more than 4 hours ago. Auto-disabling to improve performance.")
+            settings.setValue("DICOM/detailedLogging", 0)
```
lassoan: `detailedLoggingEnabledTime` can lead to a somewhat inconsistent state: when detailed logging is enabled but `detailedLoggingEnabledTime` is not set. Instead if we had a `detailedLoggingAutoDisableTime` or `disableDetailedLoggingTime` setting value then the situation would be always clear: if the value is set then detailed logging must be disabled at that time. If it is not set then it means that detailed logging will not be automatically disabled.

Modules/Scripted/DICOM/DICOM.py
```diff
@@ -512,6 +517,35 @@ def onBrowserWidgetClosed(self):
...
+            settings.remove("DICOM/detailedLoggingEnabledTime")
+            # Update the log level immediately
+            ctk.ctk.setDICOMLogLevel(ctk.ctkErrorLogLevel.Warning)
```
lassoan: It would be nice to call a function to enable/disable detailed logging (so that we don't need to hardcode the CTK warning levels corresponding to detailed/normal logging at several places).

Modules/Scripted/DICOM/DICOM.py
```diff
@@ -588,23 +623,30 @@ def __init__(self, parent):
         vBoxLayout.addStretch(1)
 
     def onDetailedLoggingStateChanged(self, detailedLoggingState):
+        settings = qt.QSettings()
+
         if detailedLoggingState == qt.Qt.Checked:
-          ctk.ctk.setDICOMLogLevel(ctk.ctkErrorLogLevel.Debug)
+            ctk.ctk.setDICOMLogLevel(ctk.ctkErrorLogLevel.Debug)
+            # Record the time when detailed logging was enabled
+            settings.setValue("DICOM/detailedLoggingEnabledTime", time.time())
```
lassoan: `registerProperty` may make this method be called at startup. If `DICOM/detailedLogging` was already enabled then  it may result in `DICOM/detailedLoggingEnabledTime` reset to the current time.
Whether the method is called may also depend on what the default state of the checkbox.
Probably it would fix potential issues, if we only set the time if it was not set already.
But maybe the safest would be to add a pushbutton to temporarily enable detailed logging, and a pushbutton to disable detailed logging. Next to it there could be a textbox displaying the current state ("disabled", or "enabled", or "enabled until 15:13, 2026-01-22").

lassoan (commented):
Thank you, very good changes, I just added a few comments to make the temporary detailed logging mechanism a bit more bulletproof.

Base/QTApp/qSlicerMainWindow.cxx
Punzo: nice catch!

Modules/Scripted/DICOM/DICOM.py
Punzo: done

Modules/Scripted/DICOM/DICOM.py
Punzo: done

Modules/Scripted/DICOM/DICOM.py
Punzo: done

Punzo:
@lassoan review applied!

Punzo:
reminder: this PR needs https://github.com/commontk/CTK/pull/1345

Punzo:
@lassoan I have updated to latest CTK this PR and it is ready.

Punzo:
> Thank you, it looks good!

thanks for reviewing/merging!

PR #8928, opened 2025-12-26
ENH: Add support for building against VTK version 9.6.0

jamesobutler:
This updates Slicer to support building with VTK 9.6.0. 

This new version of VTK 9.6.0 includes a possibly beneficial change from https://github.com/Kitware/VTK/commit/73923b5ec1335de1a42488b23a42dbae6fb5b817 related to faster vtkObserver performance. Read https://github.com/Kitware/VTK/blob/v9.6.0.rc1/Documentation/release/9.6/optimize-vtkObserver.md for more details. Performance graph shown below for convenience.

> image

Marking this PR initially as draft for the following items to be completed first:
- [x] #8925 should be integrated first
- [x] #8926 should be integrated first
- [x] #8927 should be integrated first 
- [x] Rebase this branch once the above are integrated
- [x] Mark this PR as ready for integration once VTK 9.6.0 stable is released. [Original plan](https://discourse.vtk.org/t/vtk-9-6-0-release-planning/15760/2?u=jamesobutler) was January 16th 2026, though it is likely to be delayed a couple of weeks.

PR #8984, opened 2026-01-23
BUG: Update VTK with virtual reality fixes

lassoan:
Slicer/VTK:
```
$ git shortlog e21c90bd874fb15f1dc34986c238462b8aab4af8..08210fbecda09bea2544dbb80777d634e1bfea25
Andras Lasso (1):
      [Slicer] Fix compilation error in vtkOpenXRManagerRemoteConnection.cxx

Mathieu Westphal (1):
      [Backport] OpenXR: Adding occulus touch support
```


jamesobutler:
https://github.com/Kitware/VTK/commit/417481cf4294cc0e1f60e85dcdde16ea79c1f52d and https://github.com/Kitware/VTK/commit/0925dfcdb8a8d42acd283b22216201f647d64d5f are both on the VTK `release` branch and will be included in the upcoming VTK 9.6.0 release, so we won't need to backport these commits for that new version that will be added through #8928.

lassoan:
Ok. We needed these fixes urgently because we are going to work on virtual reality features during the project week.

jamesobutler:
Of course, wasn't saying to not include them here. Just making note that with the 9.6.0 release, we won't have to worry about backporting these commits again as they will already be included.

PR #8987, opened 2026-01-23
ENH: Set application icon to SlicerApp-Real.exe executable

jamesobutler:
On the Windows platform there is both the launcher application and SlicerApp-Real.exe application displayed in task manager as a bundle. Previously only the launcher had the application icon set to it, but this commit makes sure the main application executable also has an icon set.

Windows Task Manager: 

| `main` | This PR |
|--------|----------|
|slicerapp-real-icon-original|slicerapp-real-icon|

### DIFF
--- a/CMake/SlicerMacroBuildApplication.cmake
+++ b/CMake/SlicerMacroBuildApplication.cmake
@@ -451,6 +451,26 @@ macro(slicerMacroBuildApplication)
   endif()
   message(STATUS "Setting ${SLICERAPP_APPLICATION_NAME} executable name to '${executable_name}${CMAKE_EXECUTABLE_SUFFIX}'")
 
+  if(WIN32)
+    # Generate a minimal resource file so the icon is always embedded into the
+    # executable at link time without requiring custom apps to provide a .rc file.
+    # This allows for the ${executable_name}App-real (e.g. SlicerApp-real.exe) as shown in
+    # Windows task manager to be displayed with the application icon instead of a missing icon.
+    # On Windows there is both the AppLauncher executable as well as the main application
+    # executable (e.g. Slicer.exe and SlicerApp-real.exe) where we want to define an icon.
+    get_filename_component(_slicerapp_icon_name "${SLICERAPP_WIN_ICON_FILE}" NAME)
+    get_filename_component(_slicerapp_icon_dir "${SLICERAPP_WIN_ICON_FILE}" DIRECTORY)
+    set(_slicerapp_rc "${CMAKE_CURRENT_BINARY_DIR}/${SLICERAPP_NAME}.rc")
+    file(WRITE "${_slicerapp_rc}" "IDI_ICON1 ICON \"${_slicerapp_icon_name}\"\n")
+    # Override AdditionalIncludeDirectories for this source file to contain only
+    # the icon's directory, preventing propagation of all target include directories
+    # to rc.exe which would cause MSB6002 command-line length failures.
+    set_source_files_properties("${_slicerapp_rc}" PROPERTIES
+      VS_SETTINGS "AdditionalIncludeDirectories=${_slicerapp_icon_dir}"
+      )
+    list(APPEND SLICERAPP_SRCS "${_slicerapp_rc}")
+  endif()
+
   ctk_add_executable_utf8(${slicerapp_target}
     ${SLICERAPP_EXE_OPTIONS}
     ${SLICERAPP_SRCS}


jamesobutler:
👋🏻 Thanks for the review @jcfr !

lassoan:
This is nice, thanks for working on this. How does this work when SlicerApp-real.exe keeps changing and CTKResEdit also modifies the same file? How it is ensured that CTKResEdit is always called exactly after SlicerApp-real.exe is changed? It might be more robust if CTKResEdit input and output files are different (then the build system can see from the file timestamps if the icon has been updated or not).

CMake/SlicerMacroBuildApplication.cmake
```diff
@@ -837,8 +837,16 @@ macro(slicerMacroBuildApplication)
           )
         add_dependencies(${SLICERAPP_APPLICATION_NAME}UpdateLauncherIcon ${SLICERAPP_APPLICATION_NAME}ConfigureLauncher)
 
+        # Also embed icon into SlicerApp-real.exe (main application binary)
+        add_custom_target(${SLICERAPP_APPLICATION_NAME}UpdateRealAppIcon ALL
+          COMMAND
+            ${CTKResEdit_EXECUTABLE}
+              --update-resource-ico $<TARGET_FILE:${slicerapp_target}>
+              IDI_ICON1 ${SLICERAPP_WIN_ICON_FILE}
+          )
```
jcfr:
```suggestion
        # Also embed icon into SlicerApp-real.exe (main application binary)
        add_custom_command(
          TARGET ${slicerapp_target}
          POST_BUILD
          COMMAND
            ${CTKResEdit_EXECUTABLE}
              --update-resource-ico $<TARGET_FILE:${slicerapp_target}>
              IDI_ICON1 ${SLICERAPP_WIN_ICON_FILE}
          COMMENT "Update '${slicerapp_target}' application binary icon"
          )
```

jcfr: This would ensure the icon is embedded only when the main application executable is rebuilt.

jcfr: I think adding `DEPENDS ${SLICERAPP_WIN_ICON_FILE}` should ensure the command is also executed when the icon file is modified (to be confirmed)

lassoan: Patching the .exe is justified the CTK app launcher, because we do not build that .exe, so we cannot embed the icon at build time. Also, the risk of race conditions is very low because the binary is not built, but just downloaded (and so almost never changes). Any unnecessary build steps have also lower impact because the launcher patching happens in the top-level build, not in the inner-build.

In contrast, if we patch a file built in the inner-build then the timestamps will be always messed up, which can lead to missed or unnecessary build steps. Even if right now we don't immediately see a failure case, it is unnecessary to expose ourselves to such complications.

Instead, we can cleanly put the correct icon in the .exe at build time. We just need to add an [.ico file](https://github.com/PlusToolkit/PlusApp/blob/master/fCal/Resources/icon_fCal.ico) and an [.rc file](https://github.com/PlusToolkit/PlusApp/blob/master/fCal/fCal.rc) to the source tree and [add the .rc file to the source file list in the CMakeLists.txt file](https://github.com/PlusToolkit/PlusApp/blob/17011e5df15ca2cfdd6ba3f285ab7a8568d942a1/fCal/CMakeLists.txt#L27-L29).

jamesobutler: @lassoan I have come back to this work and implemented the usage of an .rc file as you suggested. Please take a look for latest changes. I have tested in the context of 3D Slicer and my Slicer custom application and it appears to be working successfully.

CMake/SlicerMacroBuildApplication.cmake
```diff
@@ -450,9 +450,25 @@ macro(slicerMacroBuildApplication)
   endif()
   message(STATUS "Setting ${SLICERAPP_APPLICATION_NAME} executable name to '${executable_name}${CMAKE_EXECUTABLE_SUFFIX}'")
 
+  if(WIN32)
```
lassoan, line 454: Shouldn't we initialize SLICERAPP_SRCS?
```suggestion
    SET(SLICERAPP_SRCS "Main.cxx")
    if(WIN32)
```

CMake/SlicerMacroBuildApplication.cmake
```diff
@@ -450,9 +450,25 @@ macro(slicerMacroBuildApplication)
...
+    # Generate a minimal resource file so the icon is always embedded into the
```
lassoan, line 455: Describe why this is only needed (or possible) on Windows and describe where this icon is visible (e.g., Windows Task manager "Apps" list).

CMake/SlicerMacroBuildApplication.cmake
jamesobutler: https://github.com/Slicer/Slicer/blob/eee0da630dee571658434e086a5cd3a6cf12f044/CMake/SlicerMacroBuildApplication.cmake#L300-L305
^This is creating the various `SLICERAPP_` prefixed items whether.

https://github.com/Slicer/Slicer/blob/eee0da630dee571658434e086a5cd3a6cf12f044/Applications/SlicerApp/CMakeLists.txt#L57-L59
^APP_SRCS is passed into SRCS which becomes SLICERAPP_SRCS

In the case that SRCS is passed as undefined, then doing a CMake call to append list to an uninitialized variable will treat it as a empty list and add to it.

CMake/SlicerMacroBuildApplication.cmake
jamesobutler: I've pushed an update to add some additional comments here.

CMake/SlicerMacroBuildApplication.cmake
lassoan: Previously, SRCS was ignored and Main.cxx was used as source files.
Now SRCS is not ignored but used here (which is nice), but it also means that if SRCS is not set (e.g., when building a custom application?) then Main.cxx will not be included in the sources and the build will fail in a not-so-obvious way.

To make things more clear, I would recommend these small changes:
- Split these two changes (fixing of SRCS argument handling and adding application icon) to two separate commits as they are unrelated and not completely trivial.
- Document that SRCS is expected to include the filename that implements `SlicerAppMain`.
- Add `SRCS` to `expected_defined_vars` list. If we want to preserve backward compatibility: if `SLICERAPP_SRCS` is not defined then set it to `Main.cxx`. I would not worry about backward compatibility in this case, because the [custom application template always included setting of SRCS to `Main.cxx`](https://github.com/KitwareMedical/SlicerCAT/blame/e651f197f56e3378ce094378dd29631c08b25a33/Applications/SlicerCATApp/CMakeLists.txt#L81).

CMake/SlicerMacroBuildApplication.cmake
```diff
@@ -450,9 +450,29 @@ macro(slicerMacroBuildApplication)
...
+    # executable at link time without requiring custom apps to provide a .rc file.
+    # This allows for the ${executable_name}App-real (e.g. SlicerApp-real.exe) as shown in
+    # Windows task manager to be displayed with the application icon instead of a missing icon.
+    # On Windows there is both the AppLauncher executable as well as the main application
+    # executable (e.g. Slicer.exe and SlicerApp-real.exe) where we want to define an icon.
+    # Override AdditionalIncludeDirectories for this source file to contain only
```
lassoan: Move the comment on "Override AdditionalIncludeDirectories" closer to where it is actually overridden.

lassoan:
Thanks for the update and additional details. I've added a few comments asking for explaining a bit more clearly what is done and why (to make future changes/fixes easier).

CMake/SlicerMacroBuildApplication.cmake
jamesobutler: Ok I have split into 2 commits and set it to the `expected_defined_vars` and not adding the code to define it if missing since as you observed the Slicer custom app template has always specified it correctly.

lassoan (approved):
Thanks for the updates, it all looks good to me.

PR #8948, opened 2026-01-06
COMP: Update DCMTK to 3.7.0+ commit with binary seg fix

jamesobutler:
Per https://github.com/Slicer/Slicer/pull/8946#issuecomment-3715986049 to sync up with the DCMTK version used by DCMQI. This is additional work for https://github.com/Slicer/Slicer/issues/8387.
```
DCMTK:
$ git shortlog ccfd10b..3e85b37
Joerg Riesmeier (6):
      Fixed versions numbers of external libraries.
      Fixed access rights.
      Added missing "see also" reference to json2dcm.
      Removed wrong reference to "dump2dcm(2)".
      Check pointer before calling setParent().
      Fixed typos in / formatting of API documentation.

Marco Eichelberg (2):
      Updated version information for 3.7.0+ development.
      Updated version information for 3.7.0+ development.

Michael Onken (2):
      Init Frame with number of pixels (not bytes).
      High Bit is 0 not 1 for binary segmentations.
```

PR #8988, opened 2026-01-23
COMP: Update DCMTK to 3.7.0++ commit with additional binary seg fix

fedorov:
This should fix the dashboard linker errors for the dcmqi extension identified while testing the new functionality added in DCMTK (see https://github.com/Slicer/Slicer/pull/8948#issuecomment-3763790260). The actual fix is in this commit: https://github.com/DCMTK/dcmtk/commit/a7369385d91f40e19a9b2d4ef922c61370dfc5b1.

cc: @michaelonken 

```text
$ git shortlog 3e85b37..a736938
Harald Roesen (8):
      Reworked struct OFUUID.
      Added dependencies for reworked OFFUUID struct.
      Removed unnecessary include(s).
      Added few 'negated' tests.
      Finalization of OFUUID class.
      Replaced the old OFUUID implementation with a new one. The old OFUUID represented an UUID version 1, while the new one is an UUID version 7. It offers monotonicity, and improved unguessability. In addition it mitigates security issues because it does not use a Media Access Control (MAC) address as is the case with version 1.
      Check pointer before calling setParent().
      Replaced friend operators with member operators.

Joerg Riesmeier (12):
      Fixed issue with delimiters in Chinese encodings.
      Also test Japanese character sets with "oficonv".
      Updated libiconv version number.
      Added test case with a backslash-like byte.
      Link dcmimage privately to libtiff and libpng.
      Added new DcmSCU loggers to sample config file.
      Minor fixes to comments
      Minor fixes to comments.
      Removed empty lines at the end of the file.
      Replaced leading spaces by a tab character.
      Added OFUUID to list of further classes/utilities.
      Fixed outdated reference to conversion flags.

Marco Eichelberg (6):
      Added dependency check between command line options.
      Added DcmJsonFormat::getMinBulkSize().
      Fixed possible segfault in dcmpschk application.
      Fixed minor gcc -Wextra warning.
      Fixed minor memory leak in oficonv.
      Silently convert path separators in dump2dcm.

Michael Onken (5):
      Added some optional return keys.
      Make SCU dataset logging configurable.
      Updated copyright dates.
      Make new DcmSCU logger names consistent.
      Export symbols of OverlapUtil for shared builds.
```

### DIFF
--- a/SuperBuild/External_DCMTK.cmake
+++ b/SuperBuild/External_DCMTK.cmake
@@ -70,7 +70,7 @@ if(NOT DEFINED DCMTK_DIR AND NOT Slicer_USE_SYSTEM_${proj})
 
   ExternalProject_SetIfNotDefined(
     Slicer_${proj}_GIT_TAG
-    "3e85b37444107e93550167c2284e64b4881b0fcb" # DCMTK 3.7.0+ 2025-12-18
+    "a7369385d91f40e19a9b2d4ef922c61370dfc5b1" # DCMTK 3.7.0++ 2026-01-22
     QUIET
     )
 


fedorov:
@pieper @lassoan @jamesobutler at least 2 projects at the upcoming project week rely on dcmqi, which won't be available in the latest Slicer at least until this PR is merged. We are sorry this issue was not detected earlier. Would be great if you could review sometime soon!

fedorov:
Thank you! Fingers crossed this will get us through the nightly tests!

michaelonken:
It seems there is a problem with the [windows build.](https://slicer.cdash.org/viewBuildError.php?buildid=4075969). Is the full build log somewhere available (including CMake configuration log, full compiler call, ...? Cannot find it when starting from the posted link.

lassoan:
The error message you linked indicates that the problem is a name clash with the Windows `max` macro. DCMTK already uses macro-proof `numeric_limits::max` call at a couple of places (for example [here](https://github.com/DCMTK/dcmtk/blob/6bde69cfa8d69dbde150d4a9909b41ce1ab3069a/ofstd/include/dcmtk/ofstd/oflimits.h#L43)), so a solution could be to use this technique in `ofvriant.h`, too.

I've changed [line 95](https://github.com/DCMTK/dcmtk/blob/6bde69cfa8d69dbde150d4a9909b41ce1ab3069a/ofstd/include/dcmtk/ofstd/ofvriant.h#L95) to this and the error want away:
 
    AlternativeCount <= (std::numeric_limits<IndexAlternative0>::max)(),

Unfortunately, after this, a similar error in https://github.com/InsightSoftwareConsortium/ITK/blob/3fe1f8642c6beb18a6fc82ea627a2159f99743e6/Modules/ThirdParty/VNL/src/vxl/core/vnl/vnl_sse.h#L162 came up.

This is caused by DCMTK including windows.h without defining `NOMINMAX` before the include. The dcmqi build can be fixed by always including an ITK header before DCMTK headers, which ensures that windows.h is properly included.

---

**I've submitted a PR that fixes the Windows dcmqi build: https://github.com/QIICR/dcmqi/pull/525 (it also fixes an incorrect export macro in Bin2Label.h).**

To make DCMTK builds more robust, it could make sense to add a CMake option to call NOMINMAX before including windows.h (and enable this option by default). It can still be useful to switch to macro-proof `numeric_limits::max` call everywhere, just in the very unlikely case that some project relies on the windows min/max macros and so does not want to use `NOMINMAX`; or some project already includes windows.h without NOMINMAX before including DCMTK headers.

jamesobutler:
The Windows DCMQI build is still reporting a failure on the factory machine: https://slicer.cdash.org/viewBuildError.php?buildid=4081433

fedorov:
~Yes, but this, again, is a different error.~ It was indeed the same error, but in a different place! @michaelonken was going to look into this yesterday, but CDash was down.

fedorov:
Thanks to @lassoan efforts, the latest issue was resolved via https://github.com/QIICR/dcmqi/pull/526, and I confirm CDash Windows build is successful! 🎉

PR #8993, opened 2026-01-24
Statement the wrong way round?

benediktjohannes:
I'm really not sure about this one, but doesn't this have to be the other way around? Please correct me if I'm mistaken! I'm really not sure

### DIFF
--- a/Modules/Loadable/SceneViews/GUI/qMRMLSceneViewMenu.cxx
+++ b/Modules/Loadable/SceneViews/GUI/qMRMLSceneViewMenu.cxx
@@ -134,7 +134,7 @@ void qMRMLSceneViewMenuPrivate::onMRMLNodeRemoved(vtkObject* mrmlScene, vtkObjec
     return;
   }
 
-  if (this->SceneViewsLogic->IsSceneViewNode(sequenceBrowserNode))
+  if (!this->SceneViewsLogic->IsSceneViewNode(sequenceBrowserNode))
   {
     return;
   }


lassoan:
The fix is good, please fix the commit comment (use the `BUG:` prefix) and it should be good to go.
To avoid these issues in the future, please follow these instructions: https://slicer.readthedocs.io/en/latest/developer_guide/contributing.html#how-to-submit-a-pr
The `SetupForDevelopment.sh` script sets up a pre-commit hook that runs various checks your commit, including verifying that the commit message is valid.

jamesobutler:
@lassoan I guess a problem in this case is that `SetupForDevelopment.sh` assumes the user has pulled code locally. However as indicated by the default branch name of “patch-X”, this edit and commit was likely made from the GitHub Web Interface where these pre-commit git hooks are not executed. Steps 1,3-5 can all be done through the web interface but not step 2 to run the SetupForDevelopment script. Clarifying web editing vs local editing may be helpful.

lassoan:
The github web GUI will not be sufficient for managing pull requests until editing of commit messages is implemented. I can understand that for a single pull request it may be a significant overhead to clone and configure the Slicer repository. But @benediktjohannes was already kind enough to spend time with creating several pull requests, so setting up a local Slicer repository would worth the little extra time. It would ensure we don't need to reject pull requests due to invalid commit comments, and it would also make it much easier to create cleaner pull requests (as reordering and squashing commits is possible on a local git repository, but unavailable in the github web GUI).

benediktjohannes:
Hi, thanks for the very positive feedback! 👍 I'd like to close this PR and open a new one with the correct commit message because (like you've already said) it's (unforetunately) impossible in GitHub Web UI to change the commit message, but maybe I'll get a cloned repo for git soon, thanks for the feedback! 👍

PR #8998, opened 2026-01-26
Crash closing application after fill between slices

xriobe:
Launch in the terminal Slicer 5.10 (or 5.8), load a volume and go to Segment Editor.
Add a segment and Paint 2 separated slices.
Use Fill between slices, Initialize, Apply.
Close the application and look in the terminal, an exception occurred :
  File "/Slicer-5.10.0-linux-amd64/lib/Slicer-5.10/qt-scripted-modules/SegmentEditorEffects/AbstractScriptedSegmentEditorAutoCompleteEffect.py", line 67, in __del__
  File "/Slicer-5.10.0-linux-amd64/lib/Slicer-5.10/qt-scripted-modules/SegmentEditorEffects/AbstractScriptedSegmentEditorAutoCompleteEffect.py", line 210, in observeSegmentation
ImportError: sys.meta_path is None, Python is likely shutting down

If we just deal with the "import" issue in observeSegmentation, there will be another one:
ValueError: Trying to call 'parameterSetNode' on a destroyed qSlicerSegmentEditorScriptedEffect object

The second commit is produced if instead of closing the application, you click the Undo button until there is nothing left, there will be an error in the terminal:
DeserializeImageGeometry: Failed to de-serialize geometry string

### DIFF
--- a/Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorEffects/AbstractScriptedSegmentEditorAutoCompleteEffect.py
+++ b/Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorEffects/AbstractScriptedSegmentEditorAutoCompleteEffect.py
@@ -410,7 +410,7 @@ class AbstractScriptedSegmentEditorAutoCompleteEffect(AbstractScriptedSegmentEdi
         effectiveGeometryImage = slicer.vtkOrientedImageData()
         effectiveGeometryString = segmentationNode.GetSegmentation().DetermineCommonLabelmapGeometry(
             vtkSegmentationCore.vtkSegmentation.EXTENT_UNION_OF_EFFECTIVE_SEGMENTS, self.selectedSegmentIds)
-        if effectiveGeometryString is None:
+        if not effectiveGeometryString:
             return True
         vtkSegmentationCore.vtkSegmentationConverter.DeserializeImageGeometry(effectiveGeometryString, effectiveGeometryImage)
 


xriobe:
Cc: @lassoan @cpinter

cpinter:
Good catch @xriobe ! I'd say let's integrate the change, but if @lassoan you think we need to test I can do that, just not today.

Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorEffects/AbstractScriptedSegmentEditorAutoCompleteEffect.py
```diff
@@ -3,7 +3,7 @@
 import ctk
 import qt
 import vtk
-
+import vtkSegmentationCorePython as vtkSegmentationCore
```
lassoan: `import vtkSegmentationCorePython as vtkSegmentationCore` is (was?) needed as a workaround, because the Python wrapper did not find specific classes in vtkSegmentationCorePython without importing it first. Unless you have found that this change was a must to fix a problem, don't touch it. If we change this pattern then it has to be changed in all other scripted segment editor effects and the rationale should be described in detail and thoroughly tests.

Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorEffects/AbstractScriptedSegmentEditorAutoCompleteEffect.py
```diff
@@ -397,15 +396,13 @@ def effectiveExtentChanged(self):
...
-        import vtkSegmentationCorePython as vtkSegmentationCore
-
         segmentationNode = self.scriptedEffect.parameterSetNode().GetSegmentationNode()
 
         # The effective extent for the current input segments
         effectiveGeometryImage = slicer.vtkOrientedImageData()
         effectiveGeometryString = segmentationNode.GetSegmentation().DetermineCommonLabelmapGeometry(
             vtkSegmentationCore.vtkSegmentation.EXTENT_UNION_OF_EFFECTIVE_SEGMENTS, self.selectedSegmentIds)
-        if effectiveGeometryString is None:
+        if effectiveGeometryString is None or effectiveGeometryString == "":
```
lassoan:
```suggestion
        if not effectiveGeometryString:
```

xriobe:
Thank you for the review, i simplified both commits.

Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorEffects/AbstractScriptedSegmentEditorAutoCompleteEffect.py
```diff
@@ -207,6 +207,11 @@ def onSegmentationModified(self, caller, event):
             self.delayedAutoUpdateTimer.start()
 
     def observeSegmentation(self, observationEnabled):
+        if not self.scriptedEffect:
+            # This can be the case if the Python interpreter has already begun its shutdown process
+            # when observeSegmentation is called from __dell__ method.
```
lassoan:
```suggestion
            # when observeSegmentation is called from __del__ method.
```

lassoan (commented):
Thank you, looks good to me. Just fix the typo in the comment and it should be good to go.

xriobe:
oops, sorry, corrected and rebased

Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorEffects/AbstractScriptedSegmentEditorAutoCompleteEffect.py
cpinter: Good point

Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorEffects/AbstractScriptedSegmentEditorAutoCompleteEffect.py
cpinter: Interesting. I thought this was not good practice to allow all kinds of "being false". I usually use
`if effectiveGeometryString in [None, '']:`

lassoan: `if not a:` is simpler, more Pythonic, but it may mask some errors.

In this case, `effectiveGeometryString` should never be `0` but if it ends up being that then we don't get an error if we use the simple check. Considering that the probability is low and consequences are probably not severe, I think it is an acceptable little cost for a little improvement in readability and simplicity.

Issue #6811, opened 2023-02-02
Support for building Slicer on Apple arm64


jcfr:
This is meta issues aiming to organize the sub-tasks associated with arm64 support.

* [ ] https://github.com/Slicer/Slicer/issues/5944
* [ ] https://github.com/Slicer/slicer_download/issues/22
* [ ] https://github.com/Slicer/slicer_download_stats/issues/6

pieper:
I was able to build Slicer current git (c6a6e4eb7aaa6be318b18438757d26e3f8510cfd) on a macbook Air M4. Homebrew versions of Qt and CMake.

Some small issues:
* Initial startup is very slow, probably due to shared libs, maybe just in debug mode.  Faster in restarts but still slow.
* The render views are blank at first, but resume to normal after changing the layout
* The mouse is "stuck" in the 3D view, always rotating
* Needed to turn off I18N due to bad path for Qt helper app
* Missing `pydicom` o WebServer and MultiVolumeImporter do not load

```
pieper@slate s % sw_vers
ProductName:		macOS
ProductVersion:		26.2
BuildVersion:		25C56

pieper@slate ~ % qmake --version
QMake version 3.1
Using Qt version 6.10.1 in /opt/homebrew/lib

pieper@slate ~ % cmake --version
cmake version 4.2.3
```

```
Make commands:
mkdir -p /opt/s
cd /opt/s
cmake \
  -DCMAKE_OSX_ARCHITECTURES=arm64 \
  -DSlicer_REQUIRED_QT_VERSION="6.10" \
  -DCMAKE_BUILD_TYPE:STRING=Debug \
  -DSlicer_USE_SimpleITK:BOOL=OFF \
  -DSlicer_BUILD_I18N_SUPPORT:BOOL=OFF \
  -DSlicer_BUILD_DICOM_SUPPORT:BOOL=OFF \
  -DSlicer_VTK_SMP_IMPLEMENTATION_TYPE:STRING=Sequential \
  -DCMAKE_OSX_DEPLOYMENT_TARGET:STRING=26.2 \
  -DCMAKE_OSX_SYSROOT:STRING="/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk" \
  /Users/pieper/slicer/latest/Slicer
```

Then
```
make -j10
```

[...]

PR #9003, opened 2026-01-26
COMP: Pass macOS architecture settings to CTK external project

vacaslistradas:
## Summary
- Passes CMAKE_OSX_ARCHITECTURES, CMAKE_OSX_DEPLOYMENT_TARGET, and CMAKE_OSX_SYSROOT to CTK external project
- Required for native ARM64 builds on Apple Silicon Macs

## Problem
When building Slicer with `-DCMAKE_OSX_ARCHITECTURES=arm64`, CTK's nested PythonQtGenerator builds for x86_64 instead of arm64, causing linker errors.

## Related
- Depends on: commontk/CTK#1346
- Work towards: #6811, #5944

## Test Plan
- [x] Built Slicer with `-DCMAKE_OSX_ARCHITECTURES=arm64` on macOS Tahoe (M4)
- [x] With CTK fix applied, PythonQtGenerator builds successfully for ARM64
- [x] CTK completes without architecture mismatch errors


pieper:
Is this actually needed?  My arm64 mac build worked fine as of a month or so ago.

jamesobutler:
Arm64 builds have been successful without this change. If a specific scenario can be brought up where this is needed, then it can be reconsidered.

PR #9004, opened 2026-01-26
BUG: Fix onMRMLNodeRemoved false return logic

benediktjohannes:
Please review @lassoan , thanks 👍

### DIFF
--- a/Modules/Loadable/SceneViews/GUI/qMRMLSceneViewMenu.cxx
+++ b/Modules/Loadable/SceneViews/GUI/qMRMLSceneViewMenu.cxx
@@ -134,7 +134,7 @@ void qMRMLSceneViewMenuPrivate::onMRMLNodeRemoved(vtkObject* mrmlScene, vtkObjec
     return;
   }
 
-  if (this->SceneViewsLogic->IsSceneViewNode(sequenceBrowserNode))
+  if (!this->SceneViewsLogic->IsSceneViewNode(sequenceBrowserNode))
   {
     return;
   }


benediktjohannes:
@lassoan could you please be as kind and also have a look at SlicerHeart, thanks!

cpinter:
All the core Slicer developers are at the https://projectweek.na-mic.org/PW44_2026_GranCanaria, please be patient

benediktjohannes:
Oh, yeah, of course! Please don‘t stress about it, have a nice conference!

Modules/Loadable/SceneViews/GUI/qMRMLSceneViewMenu.cxx
```diff
@@ -134,7 +134,7 @@ void qMRMLSceneViewMenuPrivate::onMRMLNodeRemoved(vtkObject* mrmlScene, vtkObjec
     return;
   }
 
-  if (this->SceneViewsLogic->IsSceneViewNode(sequenceBrowserNode))
```
jamesobutler: @Sunderlandkyl The existing code in Slicer `main` is correct because we are trying to make sure the node being removed is a sequence browser node and so if it is a scene view node we want to return early? This code was most recently updated in your https://github.com/Slicer/Slicer/commit/fb6fb19ef37f60abc7c021c688560fedd1ccfebe.

Sunderlandkyl: Yes I think the original code was correct since we are only interested in scene view sequences.

jamesobutler: Ok, to confirm we should then close this PR?

benediktjohannes: I am a bit confused Right now. If you Are only interested in scene view sequences, shouldn‘t we then Return if the scene view sequences are Not the case, so !=, or am I mistaken? So the code standing here is Not the one I wanted to have, but the original one, which is false I thought?

lassoan: The proposed change looks necessary to me. We should return early if it is _not_ a scene view browser node.

benediktjohannes: Thanks for the Feedback @lassoan

lassoan:
@Sunderlandkyl please have a look. This may cause subtle errors in scene views.

Sunderlandkyl:
Sorry, I took a look again and it seems like a valid fix. Not sure why I read it the other way around before.

lassoan:
Thank you, merged.

I had to fix the incorrect commit message caused by the merge from the main branch and I've accidentally pushed it directly to the main branch. That's why this shows up as "closed with unmerged commits".

PR #9007, opened 2026-01-27
ENH: Disable Brotli support in CURL configuration

ruffsl:
- Add -DCURL_BROTLI=OFF and -DCURL_USE_BROTLI=OFF to cURL CMake configuration
- Ensures cURL is built without Brotli support, preventing linker errors when Brotli is present on the system

Context:
- https://github.com/Slicer/Slicer/pull/8959#issuecomment-3806496066


SuperBuild/External_curl.cmake
```diff
@@ -97,6 +97,8 @@ if((NOT DEFINED CURL_INCLUDE_DIR
       -DCURL_DISABLE_FILE:BOOL=ON
       -DCURL_DISABLE_TFTP:BOOL=ON
       -DHAVE_LIBIDN2:BOOL=OFF
+      -DCURL_BROTLI:STRING=OFF
```
jamesobutler: Are you sure this is a string type?

Is all that is actually needed is the `-DCURL_USE_BROTLI:BOOL=OFF` configuration in the line below?

ruffsl: > Are you sure this is a string type?

I'm not sure I know enough CMake syntax to know, but here is how it is set via the `CMakeCache.txt` file:

```cmake
...
//Initial cache
CURL_BROTLI:STRING=OFF
...
//STRINGS property for variable: CURL_BROTLI
CURL_BROTLI-STRINGS:INTERNAL=AUTO;ON;OFF
```

[opt_slicer_ws_build_Slicer-SuperBuild-Debug_curl-build_CMakeCache.txt](https://github.com/user-attachments/files/24893225/opt_slicer_ws_build_Slicer-SuperBuild-Debug_curl-build_CMakeCache.txt)

> Is all that is actually needed

I've verified that `-DCURL_USE_BROTLI:BOOL=OFF` is insufficient, as leaving `CURL_BROTLI` on `AUTO` still causes linker failures. Haven't yet tested the inverse though, but figured setting both would be more sure-fire.

ruffsl: That all said, what are the ramifications of the use of `BROTLI` for curl in general?
Just `on-the-fly decompression` for web content?

jamesobutler: Reviewing https://github.com/curl/curl/blob/3003c32cb29347478b8e2917ca5b3e2b37b4ae33/docs/INSTALL-CMAKE.md?plain=1#L360, it does seem the opposite may be all that is needed. So just `-DCURL_BROTLI:STRING=OFF` as it appears that `-DCURL_USE_BROTLI:BOOL=OFF` is not documented. I can try a build.

ebrahimebrahim:
I just ran into this exact problem building on a linux system and had to go disable Brotli support.

Confirmed that

```
-DCURL_BROTLI:STRING=OFF
```

is sufficient to fix.

(And yes it is STRING type. It was AUTO by default for me which enabled it and caused linker errors)

jamesobutler:
Yes I will push the update now. Completed my testing early today.

jamesobutler (approved):

PR #9009, opened 2026-01-28
BUG: Maintain selection in subject hierarchy after reparenting item

cpinter:
Re https://discourse.slicer.org/t/slicer-5-10-qmrmlsubjecthierarchycombobox-turned-to-none-but-no-issue-with-5-8-1/45203

### DIFF
--- a/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyModel.cxx
+++ b/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyModel.cxx
@@ -1031,6 +1031,18 @@ void qMRMLSubjectHierarchyModel::updateItemFromSubjectHierarchyItem(QStandardIte
         // Reparent items
         QList<QStandardItem*> children = parentItem->takeRow(item->row());
         newParentItem->insertRow(newIndex, children);
+
+        // Make sure the selection is maintained after reparenting
+        QList<vtkIdType> childItemIds;
+        for (QStandardItem* child : children)
+        {
+          vtkIdType childItemId = this->subjectHierarchyItemFromItem(child);
+          if (childItemId != vtkMRMLSubjectHierarchyNode::INVALID_ITEM_ID)
+          {
+            childItemIds.append(childItemId);
+          }
+        }
+        emit requestSelectItems(childItemIds);
       }
     }
   }


cpinter:
Hold review: selection is changed in all views, which is not what we want.

cpinter:
The situation is the following

* Each view (tree, combobox) has its own model, and each model reacts to the `vtkMRMLSubjectHierarchyNode::SubjectHierarchyItemReparentedEvent`, doing the reparenting in the Qt model that happened in MRML
* Without the change, all views that have the reparented item selected lose the selection, because when we remove the item from the model (reparenting works by taking the row and re-adding it, see [here](https://github.com/Slicer/Slicer/blob/8e556d8e1a20aa3d7af6a7f42d5ee669922c6d58/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyModel.cxx#L291-L292)), it is also removed from the selection model of the view
  * However, the views that have a different selection are not affected!
* With this change the model that does the reparenting emits a `requestSelectItems` signal, which results in selection being changed to the reparented item even if it was not selected before
  * The model cannot access selection, because selection is managed by the view. So we have no way to know if the reparented item was selected or not (that way we could decide whether to emit the signal or not)

In summary, it seems that we either
1. Lose the selection if the currently selected item is reparented in MRML
2. Every SH view changes selection to the reparented item after reparenting, not only those that had the item selected to begin with

cpinter:
It seems that the proper way to do reparenting, and which ensures that the selections are preserved, is to use `beginMoveRows` and `endMoveRows`. This seems to be the only way to fix this reparenting issue in a good way. My first attempt does not seem to be robust so more work is needed.

xskere:
In my own branch [(https://github.com/xskere/Slicer/tree/reparenting-begin-and-end-moveRows)](https://github.com/xskere/Slicer/tree/reparenting-begin-and-end-moveRows), I attempted to use `beginMoveRows` and `endMoveRows` to solve this issue, it appears to work, however there is a warning stating:

```
QAbstractItemModel::endMoveRows:  Invalid index ( -1 , -1 ) in model qMRMLSubjectHierarchyModel
```

The call stack for this is:
`endMoveRows` -> `QAbstractItemModelPrivate::itemsMoved` -> `QAbstractItemModelPrivate::movePersistentIndexes`

It is there, in the first call of `QAbstractItemModelPrivate::movePersistentIndexes` that the Invalid index is shown (line 787 in qabstractitemmodel.cpp)

However, the invalid index -1 -1 comes from the begin variable

```
    const QVector<QPersistentModelIndexData *>::const_iterator begin = indexes.constBegin();
    const QVector<QPersistentModelIndexData *>::const_iterator end = indexes.constEnd();
```

Which in turn comes from `QAbstractItemModelPrivate::itemsMoved`:

```
    QVector<QPersistentModelIndexData *> moved_explicitly = persistent.moved.pop();
```

This variable is set when calling `beginMoveRows`:

`beginMoveRows` -> `QAbstractItemModelPrivate::itemsAboutToBeMoved`

By appending the data:

```
persistent_moved_explicitly.append(data);
```

When index is appended in `QAbstractItemModelPrivate::itemsAboutToBeMoved` the data looks like this:

`persistent_moved_explicitly` -> data 0x000001b3271f8e60 -> Index r=0 c=0

When it is used later in `QAbstractItemModelPrivate::movePersistentIndexes`, however it looks like this:
data 0x000001b3271f8e60 -> Index r=-1 c= -1

Somehow, the data should be the same, but it's changed and is now not valid.

Still, even with this warning, the reparenting applies correctly and the selection of the node is kept.

cpinter:
Superseded by https://github.com/Slicer/Slicer/pull/9124

Issue #7171, opened 2023-08-17
Improving Support for Python Package Dependencies in Slicer Extensions


jcfr:
_This issue was created based on a comment originally posted by @jcfr in https://github.com/Slicer/Slicer/issues/6913#issuecomment-1636187233_

------

## Is your feature request related to a problem? Please describe.

There are concerns regarding the dependency management and package bundling for Slicer applications, especially in enterprise settings and within tightly controlled environments. Currently, there is a lack of consistency in handling Python package dependencies, leading to potential issues with conflicting package versions and unexpected environment modifications.

## Describe the solution you'd like

It would be beneficial to establish a structured approach for managing Python dependencies within Slicer extensions. This solution could involve supporting two main scenarios: runtime installation of python packages and bundling of python packages into extensions.

1. **Runtime Installation**: For extensions available through the extensions manager, implement a runtime installation mechanism. This would involve documenting the specific Python package requirements for each extension. In addition of being listed in a `Dependencies` section in the extension README, these requirements could be specified in a `pyproject.toml`[^6] or `requirements.txt` file within the extension source tree. The `slicer.util.pip_install` function should be updated to support the `--constraint`[^2] parameter, allowing controlled installation of packages while accounting[^7] for dependencies across enabled extensions.

[^6]: Also worth noting that we are already using this file to organize linting rules for custom application. File is usually `Modules/Scripted/pyproject.toml`
[^7]: Call like `slicer.util.collect_extension_constraint_file()` could be added.
[^2]: https://pip.pypa.io/en/latest/user_guide/#constraints-files

2. **Bundling of Python Packages**: Extensions could provide their own "site-packages" directories appended to the Python path. This would ensure that the Slicer Python environment remains unmodified, with each extension managing its isolated packages. To achieve this, extensions would need to include a `requirements.txt` file that outlines the required packages. The Slicer build-system should be updated to bundle the required Python packages alongside the extension. Additionally, a function could be introduced to read the `dependencies` metadata from the `pyproject.toml` or `requirements.txt` files.

## Immediate next steps

To implement the proposed solution, the following steps are suggested:

1. Update extensions to include a `requirements.txt` file outlining their Python dependencies.
2. Enhance the `ScriptedLoadableModule` to include a property named `python_requirements`. This property would hold dependency information in the form of a dictionary[^5] or a path to a `requirements.txt` file.
3. Provide documentation on how to install Python dependencies for extensions.

[...]

Issue #8226, opened 2025-02-11
Update Slicer Python scripts adding type hints


jcfr:
## Is your feature request related to a problem? Please describe.

Currently, argument types in Slicer Python scripts are primarily inferred from docstrings or naming conventions. This is error-prone and can become outdated. It also limits the effectiveness of static analysis tools and limits IDE autocomplete features.

## Describe the solution you'd like

Introduce Python type hints throughout Slicer’s Python scripts, starting with those located in `Base/Python/slicer`. Type annotations would enhance clarity, maintainability, and tooling support (e.g., autocompletion and linting).

## Describe alternatives you've considered

NA

## Additional context

* https://docs.python.org/3/library/typing.html#module-typing
* https://github.com/Slicer/Slicer/issues/6875
* https://github.com/Slicer/Slicer/issues/6688
* https://github.com/Slicer/Slicer/issues/6689

PR #9010, opened 2026-01-29
Improve python package installation utilities

ebrahimebrahim:
# Python Dependency Handling Improvements

## Summary

This PR introduces a new `slicer.packaging` module for managing Python dependencies in Slicer extensions. The goal is to provide a standardized, user-friendly way to check, prompt for, and install Python packages. The existing `slicer.util.pip_install` and `slicer.util.pip_uninstall` functions are retained as backward-compatible wrappers that delegate to `slicer.packaging`.

**New/updated functions:**

| Function | Purpose |
|----------|---------|
| `slicer.packaging.load_requirements(path)` | Load a `requirements.txt` file into `Requirement` objects |
| `slicer.packaging.load_pyproject_dependencies(path)` | Load `[project.dependencies]` from a `pyproject.toml` file into `Requirement` objects |
| `slicer.packaging.pip_check(reqs)` | Check if requirements are satisfied (pure Python, no subprocess). Accepts a string, a list of strings, a `Requirement`, or a list of `Requirement`s. |
| `slicer.packaging.pip_install(...)` | Canonical home; extended with modal progress dialog, non-blocking mode, status bar feedback, `--no-deps` support, and a discouraged `skip_packages` workaround for transitive dependency conflicts. Also available as `slicer.util.pip_install` (backward-compatible wrapper) |
| `slicer.packaging.pip_ensure(reqs, requester="...")` | High-level: checks, prompts, installs with progress, and offers restart if updated packages were already imported. Accepts the same flexible input as `pip_check`. |

All install functions support optional `constraints` (constraints file), `no_deps_requirements` (packages to install with `--no-deps`), and `skip_packages` (packages to exclude from the transitive dependency tree) parameters.

**Typical usage in an extension:**

```python
import slicer.packaging

reqs = slicer.packaging.load_requirements(self.resourcePath("requirements.txt"))
slicer.packaging.pip_ensure(reqs, requester="MyExtension")
import my_dependency  # Now safe
```

Or, with `pyproject.toml`:

```python
import slicer.packaging

reqs = slicer.packaging.load_pyproject_dependencies(self.resourcePath("pyproject.toml"))
slicer.packaging.pip_ensure(reqs, requester="MyExtension")
import my_dependency  # Now safe
```

With `skip_packages` (for extensions that need to exclude certain transitive dependencies; **discouraged workaround**, see [the script repository](https://github.com/Slicer/Slicer/blob/main/Docs/developer_guide/script_repository/packaging.md) for guidance):

```python
import slicer.packaging

skipped = slicer.packaging.pip_ensure(
    "nnunetv2>=2.3",
    skip_packages=["SimpleITK", "torch", "requests"],
    requester="SlicerNNUNet",
)
# skipped contains the requirement strings that were excluded
```

**Behavior of `pip_install`:**

Four operating modes:

- `show_progress=True, blocking=True` : Modal progress dialog (new default behavior)
- `show_progress=True, blocking=False`: Status bar messages
- `show_progress=False, blocking=True`: Busy cursor only (the previous default behavior, except with a busy cursor now added)
- `show_progress=False, blocking=False`: No visual indication that anything is happening (without looking at python console); specify callbacks to create more reasonable custom behaviors.

## Changes

The PR is organized as 5 commits, each a self-contained change:

1. **`ENH: Add slicer.packaging module for managing Python dependencies`**
   - `Base/Python/slicer/packaging.py` (new) — `load_requirements`, `load_pyproject_dependencies`, `pip_check`, `pip_ensure`, `pip_install`, `pip_uninstall`, the non-blocking infrastructure, the modal progress dialog, the restart prompt, and the `skip_packages` workaround
   - `Base/Python/slicer/util.py` — backward-compatible `pip_install` / `pip_uninstall` wrappers that delegate to `slicer.packaging`; updated error messages to point at the canonical name
   - `Base/Python/slicer/tests/test_slicer_packaging.py` (new) — full unit test suite for the new module
   - `Base/Python/slicer/tests/test_slicer_util_pip.py` (new) — focused tests for the back-compat wrappers
   - `Base/Python/CMakeLists.txt` — register the new module
   - `Applications/SlicerApp/Testing/Python/CMakeLists.txt` — register the test files

2. **`DOC: Document slicer.packaging in the developer guide`**
   - `Docs/developer_guide/script_repository/packaging.md` (new) — full "Python package management" section in the script repository
   - `Docs/developer_guide/script_repository.md` — include directive
   - `Docs/developer_guide/python_faq.md` — point developers at `slicer.packaging`, recommend `pip_ensure` as the high-level workflow, refresh the best-practices list to reflect that `pip_ensure` handles confirmation on its own
   - `Docs/developer_guide/script_repository/gui.md`, `models.md`, `plots.md` — sweep existing examples to use the canonical name
   - `Docs/developer_guide/slicer.md` — wire `slicer.packaging` into the auto-generated API reference

3. **`ENH: Show slicer.packaging usage in the new scripted module template`**
   - `Utilities/Templates/Modules/Scripted/TemplateKey.py` — commented-out `pip_ensure` example at the top of `onApplyButton`
   - `Utilities/Templates/Modules/Scripted/Resources/requirements.txt` (new) — blank scaffold
   - `Utilities/Templates/Modules/Scripted/CMakeLists.txt` — list the requirements file as a module resource
   - `Utilities/Scripts/SlicerWizard/TemplateManager.py` — recognize `requirements.txt` as a template file so it actually rides along when the wizard generates a new module

4. **`BUG: Clean up Python interactor startup`**
   - `Base/Python/slicerqt.py` (renamed from `Base/Python/slicer/slicerqt.py`) — moved out of the `slicer/` package directory so that CTK's `executeFile` no longer puts `bin/Python/slicer/` on `sys.path[0]` (which silently shadowed any same-named PyPI package, including the one we wanted for this PR). The wildcard `from slicer.util import *` is replaced with an explicit 5-name list (`array`, `exit`, `getNode`, `getNodesByClass`, `mainWindow`).
   - `Base/QTGUI/qSlicerPythonManager.cxx` — updated path
   - `Base/Python/CMakeLists.txt` — register the new location

5. **`BUG: Allow i18n translation functions to work without Qt`**
   - `Base/Python/slicer/i18n.py` — `translate()` no longer crashes when `slicer.app` is unavailable; it returns the untranslated text instead. This is what made it possible to wrap user-facing strings in `slicer.packaging` with `_()` while keeping the module headless-safe.

---

<details>
<summary><strong>Testing Snippets (click to expand)</strong></summary>

Convenient snippets to quickly try things. First, run this once in the Python console:

```python
import slicer.packaging
```

### load_requirements

```python
import tempfile, os
with tempfile.NamedTemporaryFile(mode='w', suffix='.txt', delete=False) as f:
    f.write("numpy>=1.20\npandas>=2.0\nscipy\n")
    path = f.name
reqs = slicer.packaging.load_requirements(path)
print([f"{r.name}: {r.specifier}" for r in reqs])
os.unlink(path)
```

### load_pyproject_dependencies

```python
import tempfile, os
with tempfile.NamedTemporaryFile(mode='w', suffix='.toml', delete=False) as f:
    f.write("[project]\ndependencies = [\n")
    f.write('    "numpy>=1.20",\n    "pandas>=2.0",\n    "scipy",\n]\n')
    path = f.name
reqs = slicer.packaging.load_pyproject_dependencies(path)
print([f"{r.name}: {r.specifier}" for r in reqs])
os.unlink(path)
```

### pip_check

```python
# Simple string syntax
slicer.packaging.pip_check("numpy>=1.0")          # True
slicer.packaging.pip_check("numpy>=99999.0")       # False
slicer.packaging.pip_check("nonexistent-package-xyz")  # False

# Multiple requirements in a single space-separated string
slicer.packaging.pip_check("numpy>=1.0 scipy>=1.0")  # True

# Markers that don't apply are considered satisfied
slicer.packaging.pip_check("foo; sys_platform == 'nonexistent'")  # True

# A list of strings also works
slicer.packaging.pip_check(["numpy>=1.0", "scipy>=1.0"])  # True

# Or a packaging.requirements.Requirement object, if you already have one
from packaging.requirements import Requirement
slicer.packaging.pip_check(Requirement("numpy>=1.0"))  # True
```

### pip_install with progress dialog (default)

```python
# Default: shows modal progress dialog, blocks until complete
# (If already installed, completes quickly with "already satisfied")
slicer.packaging.pip_install("scikit-image", requester="ReviewTest")
```

To see the full installation flow, uninstall first:
```python
slicer.packaging.pip_uninstall("scikit-image")
slicer.packaging.pip_install("scikit-image", requester="ReviewTest")
```

If it's still too fast, try putting "torch" as the package :)

### pip_install without progress dialog

```python
# Busy cursor only, no dialog
slicer.packaging.pip_install("scikit-image", show_progress=False)
```

### Non-blocking pip_install with status bar messages

```python
slicer.packaging.pip_install("scikit-image", blocking=False, requester="ReviewTest")
```

### Non-blocking pip_install with custom callbacks

```python
def onLog(line):
    print(f"[pip] {line}")

def onComplete(code):
    print(f"Done! Return code: {code}")

# Using --help to see callbacks firing with lots of output
slicer.packaging.pip_install("--help", blocking=False, show_progress=False, logCallback=onLog, completedCallback=onComplete)
```

### Backward-compat wrappers in slicer.util

The old `slicer.util.pip_install` and `slicer.util.pip_uninstall` still work; they delegate to `slicer.packaging`. Quick sanity check:

```python
slicer.util.pip_install("charset-normalizer")
slicer.util.pip_uninstall("charset-normalizer")
```

### Check if pip install is in progress

```python
slicer.packaging.pip_install("scikit-image", blocking=False, requester="ReviewTest")
print(slicer.packaging.isPipInstallInProgress())  # probably True
```

versus just this:

```python
print(slicer.packaging.isPipInstallInProgress())  # False
```

### pip_ensure

```python
# With prompt dialog (default). Does nothing if already installed --
# uninstall first if you want to see the full install flow:
#   slicer.packaging.pip_uninstall("charset-normalizer")
slicer.packaging.pip_ensure("charset-normalizer>=3.0", requester="ReviewTest")

# Without install prompt
slicer.packaging.pip_ensure("charset-normalizer>=3.0", prompt_install=False, requester="ReviewTest")
```

### pip_ensure restart prompt

After installation, `pip_ensure` checks if any updated packages were already imported in the current session. If so, it shows a "Restart Recommended" dialog with details (old → new versions). The user can restart immediately or continue.

```python
import numpy  # Ensure numpy is in sys.modules

# Reinstall numpy (already imported) -- should trigger restart prompt
slicer.packaging.pip_uninstall("numpy")
slicer.packaging.pip_ensure("numpy>=1.0", requester="ReviewTest")
# A "Restart Recommended" dialog should appear because numpy was already imported
```

To disable the restart prompt: `slicer.packaging.pip_ensure("numpy>=1.0", prompt_restart=False)`

### With constraints file

```python
import tempfile, os

# Create a constraints file that pins charset-normalizer
with tempfile.NamedTemporaryFile(mode='w', suffix='.txt', delete=False) as f:
    f.write("charset-normalizer==3.3.2\n")
    constraints_path = f.name

# Install requests (which depends on charset-normalizer) with the constraint
slicer.packaging.pip_install("requests", constraints=constraints_path)
os.unlink(constraints_path)
```

### With skip_packages (selective dependency installation)

> **Note:** `skip_packages` is a discouraged workaround for transitive dependency conflicts; see the script repository's "Python package management" section for guidance.

```python
# Install scikit-image but skip imageio (one of its dependencies)
skipped = slicer.packaging.pip_install(
    "scikit-image",
    skip_packages=["imageio"],
    requester="ReviewTest",
)
print(f"Skipped: {skipped}")

# Or via pip_ensure:
skipped = slicer.packaging.pip_ensure(
    "scikit-image>=0.20",
    skip_packages=["imageio"],
    requester="ReviewTest",
)
```

### PythonSlicer (command-line) usage

Run this from a terminal using Slicer's PythonSlicer executable (not the Slicer Python console):

```bash
/path/to/PythonSlicer -c "import slicer.packaging; slicer.packaging.pip_install('charset-normalizer')"
```

This verifies that `pip_install` works in the PythonSlicer environment where `slicer.app` and Qt are not available. The function automatically falls back to simple blocking mode.

</details>

---

<details>
<summary><strong>Design Rationale (click to expand)</strong></summary>

### Why requirements.txt? And why also pyproject.toml?

`requirements.txt` is the primary recommended format. Slicer extensions aren't Python packages — they just need "install these things into this environment," which is exactly what `requirements.txt` is for. It's pip's native input format, every Python developer knows it, and it requires no boilerplate beyond the dependency list itself.

That said, `pyproject.toml` is the modern standard in the Python ecosystem (PEP 621). It offers structured parsing via `tomllib` (stdlib) with no ad-hoc text handling, and extensions that already have a `pyproject.toml` for other tooling (ruff, pytest, etc.) can keep dependencies in one file. So we provide `load_pyproject_dependencies` as an alternative for extensions that prefer it.

Both formats boil down to PEP 508 dependency strings. Both loader functions return the same `list[Requirement]` type, so the downstream API (`pip_check`, `pip_ensure`, `pip_install`) works identically regardless of which one you use.

`load_pyproject_dependencies` reads only the `[project.dependencies]` list. Other fields in the `[project]` table (`name`, `version`, etc.) are not read or validated — extensions aren't Python packages and shouldn't need to provide them.

### Why pure-Python pip_check instead of pip --dry-run?

Installing is not done frequently, but _checking_ may be called frequently. A pure-Python implementation using `importlib.metadata` avoids the overhead of a subprocess. If we used `pip --dry-run` it would have to be a subprocess.

### Why explicit pip_ensure instead of lazy import magic?

We considered [LazyImportGroup](https://github.com/Slicer/Slicer/issues/7707) which intercepts first use of imports to trigger installation. It is elegant, but it reduces transparency and makes debugging harder for extension developers. For now we get this pattern which has more boilerplate but is more transparent and simple to debug:

```python
slicer.packaging.pip_ensure(reqs, requester="MyExtension")
import my_dependency  # Explicit, debuggable
```

IDE support works via `TYPE_CHECKING`:
```python
from typing import TYPE_CHECKING
if TYPE_CHECKING:
    import my_dependency  # For type hints only
```

We can still consider the lazy import ideas in the future -- the lower level tools here would still be useful.

### Why `pip_install` now defaults to showing a progress dialog

Before this PR, `pip_install` was always blocking with no progress feedback. The new defaults (`blocking=True`, `show_progress=True`) mean existing code that calls `pip_install` will now get a modal progress dialog "for free" without any code changes. This improves UX for many extensions immediately, while still allowing opt-out via `show_progress=False`.

Hopefully it doesnt' break too many extensions...

### Prevention of multiple non-blocking pip installs

When `pip_install` runs with `blocking=False`, it returns immediately while pip runs in the background. If another non-blocking `pip_install` is started before the first completes then we get chaos.

The solution adopted here is a module-level `_pip_install_in_progress` flag that raises `RuntimeError` if a second non-blocking install is attempted. Developers can check `isPipInstallInProgress()` first if they want to guard against this themselves.

### Why `no_deps_requirements` parameter

Some Python packages declare overly strict dependency requirements that conflict with other packages in Slicer's environment. The standard workaround requires two separate pip calls:

```python
pip_install("--no-deps problematic-package==1.0")  # Ignore its deps
pip_install("numpy scipy")  # Install known-good deps manually
```

The `no_deps_requirements` parameter handles this two-step process internally, making the intent self-documenting:

```python
pip_install(requirements="numpy scipy", no_deps_requirements="problematic-pkg==1.0")
```

This also correctly handles non-blocking mode by chaining the two pip calls internally.

### Why `skip_packages` parameter

`skip_packages` is a discouraged workaround, included reluctantly because it solves a real problem that has no clean alternative for some packages.

Multiple Slicer extensions (SlicerTotalSegmentator, SlicerNNUNet) independently implement ~70-90 lines of recursive selective-install code to install packages while excluding certain transitive dependencies. Common examples: SimpleITK (Slicer bundles a custom version), torch (must be installed via SlicerPyTorch for the correct CUDA/CPU build), and requests (already bundled, replacing it forces an unnecessary restart). For packages like nnUNet with deep transitive dependency trees, manually enumerating the full dependency list (the alternative `no_deps_requirements` approach) would impose a significant maintenance burden across every upstream version bump.

The `skip_packages` parameter centralizes this logic so extensions don't each reinvent it. When provided, each package is installed with `--no-deps`, its dependency tree is walked recursively, and any package matching the skip list is excluded. Package METADATA is updated after installation so that `pip check` doesn't flag skipped packages as missing.

The downsides are real: METADATA scrubbing hides genuine conflicts, `pip show` no longer reflects the true dependency graph, and skip lists drift as Slicer's bundled packages change. Documentation discourages reaching for `skip_packages` first and points developers at upstream fixes or `no_deps_requirements` where those are tractable.

### `skip_packages` vs `no_deps_requirements`

Both are available; they serve different purposes:

- **`no_deps_requirements`**: "Install these packages without any of their deps." You provide the correct deps yourself. Fast (2 pip calls), doesn't modify METADATA. Use when a specific package has broken dependency declarations *and* the dependency tree is small enough to enumerate by hand.
- **`skip_packages`**: "Install everything except these specific packages, anywhere in the dependency tree." Automatic recursive walk with METADATA scrubbing. Slower (one pip call per package). The escape hatch for cases where the dependency tree is too deep to enumerate manually.

They are mutually exclusive — providing both raises `ValueError`.

### Why `pip_install` doesn't accept `list[Requirement]`

It would feel natural to write `pip_install(load_requirements("requirements.txt"))`. The fact that you can't is a little unfortunate — it looks like it *should* work. But `pip_install` is extremely widely used in the wild (via `slicer.util.pip_install`), and we don't want to make breaking changes to its signature. It's also a low-level function that mirrors the pip CLI, taking the same kind of string arguments you'd pass on the command line, so it makes sense to leave it working the way it does.

The new structured `list[Requirement]` input type goes into the new `slicer.packaging` functions instead: `pip_check` and `pip_ensure`. In practice, `slicer.packaging.pip_ensure(load_requirements("requirements.txt"))` is the call you want anyway -- it checks what's already installed, prompts the user, installs only what's missing, and detects whether a restart is needed. So accepting `Requirement` objects at the `pip_ensure` level rather than the `pip_install` level steers developers toward the safer, idempotent workflow.

### Why `pip_` function naming

The new functions follow the existing naming convention established by `slicer.util.pip_install` and `slicer.util.pip_uninstall`. All pip-related functions now live together in `slicer.packaging` with the `pip_` prefix, providing a consistent, discoverable API.

### Why `slicer.packaging` (module naming)

`slicer.packaging` was always the preferred name -- it's descriptive and mirrors the well-known third-party `packaging` library. Getting there involved an unexpected detour.

The first attempt at the name failed: `slicer/packaging.py` shadowed the third-party `packaging` library, breaking `from packaging.requirements import Requirement` everywhere. Tracing the root cause led to CTK: `ctkAbstractPythonManager::executeFile` prepends the executed script's directory to `sys.path[0]` and never removes it. Since Slicer runs `slicerqt.py` at startup, and `slicerqt.py` lived inside `bin/Python/slicer/`, that directory landed on `sys.path[0]` -- which silently promoted every file inside the `slicer/` package to a top-level module name and shadowed any PyPI package of the same name.

The Slicer-side fix is to move `slicerqt.py` out of the `slicer/` package directory (commit 4 in this PR). With `slicerqt.py` at `bin/Python/`, CTK's prepend lands on a path that's already on `sys.path`, so the harmful side effect goes away. The CTK behavior is still a bug and worth fixing upstream separately, but the workaround on the Slicer side is good cleanup on its own merits: `slicerqt.py` is conceptually a startup script, not a member of the `slicer` package.

With `slicerqt.py` moved, `slicer.packaging` is once again a safe name and is what this PR ships.

### Why `import slicer.packaging` is explicit

You need `import slicer.packaging` to use it -- like any normal Python submodule. The reason this is even worth mentioning is that historically `slicer.util` works without an explicit import, because the interactor startup script reaches into the user's `__main__` namespace and pre-populates a handful of names from it. This PR narrows that startup pre-population from a wildcard to an explicit 5-name list (see commit 4), which is part of why the contrast is now sharper. There's no special reason `slicer.packaging` couldn't be added to that startup list; we chose not to because top-level Python module imports in `slicer.packaging` (e.g. `packaging.requirements`, `importlib.metadata`) would be paid by every Slicer session even though most never install a Python package, and "type one extra import line in your extension" is a trivial cost.

### Why `pip_uninstall` was also modified

While the focus of this work is on dependency installation, `pip_uninstall` was updated with the same non-blocking parameters (`blocking`, `logCallback`, `completedCallback`) for API consistency. Since both functions share the same underlying infrastructure (`launchConsoleProcess` and `_executePythonModule`) in `slicer.packaging`, extending non-blocking support to `pip_uninstall` required minimal additional code and ensures users have a symmetric API for both operations.

### Non-blocking implementation

The non-blocking mode uses a QTimer-based polling approach inspired by [SlicerMONAIAuto3DSeg](https://github.com/lassoan/SlicerMONAIAuto3DSeg). A background thread reads process output into a queue while QTimer polls from the main thread, keeping the Qt event loop responsive.

### How the restart prompt works

After `pip_ensure` installs packages, it snapshots all installed package versions (before and after) using `importlib.metadata`. For any packages whose version changed, it checks whether their top-level import names appear in `sys.modules` (meaning they were already imported in the current session). The distribution-to-import-name mapping uses `importlib.metadata.packages_distributions()` (Python 3.11+). If any already-imported packages were updated, a "Restart Recommended" dialog shows the affected packages with their old → new versions. The `prompt_restart=False` parameter disables this check.

</details>

---

<details>
<summary><strong>Interactive Feature Tour (click to expand)</strong></summary>

Build Slicer from the `python-dependency-handling-improvements` branch, launch it, open the Python console, and paste the script below. A menu lets you pick which demos to run — or select "Run All" for the full walkthrough.

```python
import importlib.metadata as _md
import os
import tempfile

import qt
from packaging.requirements import Requirement

import slicer
import slicer.packaging

# ---------------------------------------------------------------------------
# Configuration — change these to try a different demo package
# ---------------------------------------------------------------------------

DEMO_PACKAGE = "scikit-image"        # pip install name
DEMO_IMPORT = "skimage"              # Python import name
DEMO_CONSTRAINT = ">=0.20,<0.25"     # version range for constraints demo
DEMO_SKIP_DEP = "imageio"            # dependency to skip in skip_packages demo

# ---------------------------------------------------------------------------
# State
# ---------------------------------------------------------------------------

try:
    pkg_version_before_tour = _md.version(DEMO_PACKAGE)
except _md.PackageNotFoundError:
    pkg_version_before_tour = None


# ---------------------------------------------------------------------------
# Helpers
# ---------------------------------------------------------------------------

def clear_cache():
    """Purge the pip download cache so installs show real download progress."""
    slicer.packaging._executePythonModule("pip", ["cache", "purge"])


def uninstall_pkg():
    """Remove the demo package if present."""
    try:
        slicer.util.pip_uninstall(DEMO_PACKAGE)
    except Exception:
        pass


def fresh_slate():
    """Uninstall the demo package and clear the cache."""
    uninstall_pkg()
    clear_cache()


# ---------------------------------------------------------------------------
# Demo 1 — Loading Dependencies
# ---------------------------------------------------------------------------

def demo_loading_deps():
    """load_requirements() and load_pyproject_dependencies()"""

    # --- requirements.txt ---
    slicer.util.infoDisplay(
        "load_requirements(path) parses a requirements.txt file into\n"
        "Requirement objects, skipping comments, blanks, and pip options.\n"
        "\n"
        "Watch the Python console for parsed results.",
        windowTitle="Demo 1: Loading Dependencies — requirements.txt",
    )

    with tempfile.NamedTemporaryFile(mode="w", suffix=".txt", delete=False) as f:
        f.write("# Example requirements file\n")
        f.write("numpy>=1.20\n")
        f.write(f"{DEMO_PACKAGE}>=0.20\n")
        f.write("nonexistent-package-xyz>=1.0\n")
        f.write("-c constraints.txt\n")
        f.write("\n")
        path = f.name

    reqs = slicer.packaging.load_requirements(path)
    os.unlink(path)

    lines = [f"  {r.name}  {r.specifier}" for r in reqs]
    print(f"[Tour] load_requirements -> {len(reqs)} requirements:\n" + "\n".join(lines))

    # --- pyproject.toml ---
    with tempfile.NamedTemporaryFile(mode="w", suffix=".toml", delete=False) as f:
        f.write("[project]\n")
        f.write("dependencies = [\n")
        f.write('    "numpy>=1.20",\n')
        f.write(f'    "{DEMO_PACKAGE}>=0.20",\n')
        f.write('    "nonexistent-package-xyz>=1.0",\n')
        f.write("]\n")
        path = f.name

    reqs2 = slicer.packaging.load_pyproject_dependencies(path)
    os.unlink(path)

    lines2 = [f"  {r.name}  {r.specifier}" for r in reqs2]
    print(f"[Tour] load_pyproject_dependencies -> {len(reqs2)} requirements:\n" + "\n".join(lines2))

    slicer.util.infoDisplay(
        f"Loaded {len(reqs)} requirements from requirements.txt\n"
        f"and {len(reqs2)} from pyproject.toml.\n"
        "\n"
        "Both return the same Requirement objects — the downstream\n"
        "API (pip_check, pip_ensure) works identically with either.",
        windowTitle="Demo 1: Loading Dependencies — result",
    )


# ---------------------------------------------------------------------------
# Demo 2 — Checking Requirements
# ---------------------------------------------------------------------------

def demo_checking_reqs():
    """pip_check() — pure-Python requirement validation"""

    slicer.util.infoDisplay(
        "pip_check(req) checks if a requirement is satisfied.\n"
        "Pure Python, no subprocess — fast enough to call frequently.\n"
        "\n"
        "We will test several cases against the current environment.",
        windowTitle="Demo 2: Checking Requirements",
    )

    checks = [
        ("numpy>=1.0", "Bundled with Slicer — should be satisfied"),
        ("numpy>=99999.0", "Impossibly high version — should fail"),
        ("nonexistent-xyz>=1.0", "Package not installed"),
        ('foo; sys_platform == "nonexistent"', "Marker does not apply — treated as satisfied"),
    ]

    results = []
    for spec, desc in checks:
        req = Requirement(spec)
        ok = slicer.packaging.pip_check(req)
        mark = "SATISFIED" if ok else "NOT satisfied"
        results.append(f"  [{mark}]  {spec}\n      {desc}")

    slicer.util.infoDisplay(
        "pip_check results:\n"
        "\n" + "\n\n".join(results),
        windowTitle="Demo 2: Checking Requirements — results",
    )


# ---------------------------------------------------------------------------
# Demo 3 — Installing with Progress
# ---------------------------------------------------------------------------

def demo_install_progress():
    """pip_install() — modal dialog and non-blocking status bar"""

    # --- Part 1: Modal (blocking) ---
    fresh_slate()

    slicer.util.infoDisplay(
        "pip_install() — modal progress dialog (the new default).\n"
        "\n"
        "Try expanding the Details section!",
        windowTitle="Demo 3a: Modal Progress Dialog",
    )

    slicer.util.pip_install(DEMO_PACKAGE, requester="Feature Tour")

    slicer.util.infoDisplay(
        "Modal install done.\n"
        "\n"
        "Now switching to non-blocking mode: the call returns\n"
        "immediately and pip output appears in the status bar.",
        windowTitle="Demo 3a: Modal — done",
    )

    # --- Part 2: Non-blocking ---
    fresh_slate()

    slicer.util.infoDisplay(
        "pip_install() — non-blocking with status bar.\n"
        "\n"
        "Watch the STATUS BAR at the bottom. The UI stays interactive.\n"
        "isPipInstallInProgress() will be printed to the console.",
        windowTitle="Demo 3b: Status Bar Mode",
    )

    loop = qt.QEventLoop()
    _result = [None]

    def on_complete(return_code):
        _result[0] = return_code
        qt.QTimer.singleShot(0, loop.quit)

    slicer.util.pip_install(
        DEMO_PACKAGE,
        blocking=False,
        show_progress=True,
        requester="Feature Tour",
        completedCallback=on_complete,
    )

    # Check in-progress flag shortly after starting
    qt.QTimer.singleShot(500, lambda: print(
        f"[Tour] isPipInstallInProgress() = {slicer.packaging.isPipInstallInProgress()}"
    ))

    loop.exec_()

    slicer.util.infoDisplay(
        f"Non-blocking install finished (return code {_result[0]}).\n"
        "The status bar showed pip output while the UI stayed responsive.",
        windowTitle="Demo 3b: Status Bar — done",
    )


# ---------------------------------------------------------------------------
# Demo 4 — Smart Install Workflow
# ---------------------------------------------------------------------------

def demo_smart_install():
    """pip_ensure() — check, prompt, install, restart detection"""

    # --- Part 1: Normal pip_ensure ---
    fresh_slate()

    slicer.util.infoDisplay(
        "pip_ensure() — the recommended high-level API for extensions.\n"
        "Checks requirements, shows confirmation, installs with progress.\n"
        "\n"
        "You will see a confirmation dialog, then a progress dialog.",
        windowTitle="Demo 4a: pip_ensure",
    )

    reqs = [Requirement(f"{DEMO_PACKAGE}>=0.20")]
    slicer.packaging.pip_ensure(reqs, requester="Feature Tour")

    slicer.util.infoDisplay(
        "pip_ensure done. Now we will call it again immediately.\n"
        "\n"
        "pip_check sees the package is already installed, so\n"
        "pip_ensure skips everything — no dialogs, instant return.",
        windowTitle="Demo 4a: pip_ensure — done",
    )

    slicer.packaging.pip_ensure(reqs, requester="Feature Tour (no-op)")

    slicer.util.infoDisplay(
        "pip_ensure returned instantly — nothing to install.\n"
        "\n"
        "Now: restart prompt demo. We will import the package, uninstall\n"
        "it, then pip_ensure again. Since it is in memory, a restart\n"
        "dialog will appear. (Click NO when asked to restart to continue the tour.)",
        windowTitle="Demo 4a: pip_ensure — no-op verified",
    )

    # --- Part 2: Restart prompt ---
    import importlib
    mod = importlib.import_module(DEMO_IMPORT)
    print(f"[Tour] Imported {DEMO_IMPORT} {mod.__version__}")

    uninstall_pkg()
    clear_cache()

    slicer.packaging.pip_ensure(reqs, requester="Feature Tour (restart demo)")

    slicer.util.infoDisplay(
        "Restart prompt demonstrated.\n"
        "\n"
        "This helps users know when a restart is needed after\n"
        "updating packages that were already imported.",
        windowTitle="Demo 4b: Restart Prompt — done",
    )


# ---------------------------------------------------------------------------
# Demo 5 — Advanced Options
# ---------------------------------------------------------------------------

def demo_advanced_options():
    """Constraints file and skip_packages"""

    # --- Part 1: Constraints ---
    fresh_slate()

    slicer.util.infoDisplay(
        "Constraints file: limits which versions pip may install.\n"
        f"\n"
        f"Installing {DEMO_PACKAGE} constrained to {DEMO_CONSTRAINT}.",
        windowTitle="Demo 5a: Constraints",
    )

    with tempfile.NamedTemporaryFile(mode="w", suffix=".txt", delete=False) as f:
        f.write(f"{DEMO_PACKAGE}{DEMO_CONSTRAINT}\n")
        constraints_path = f.name

    slicer.util.pip_install(
        DEMO_PACKAGE,
        constraints=constraints_path,
        requester="Feature Tour (constrained)",
    )
    os.unlink(constraints_path)

    ver = _md.version(DEMO_PACKAGE)
    slicer.util.infoDisplay(
        f"Installed {DEMO_PACKAGE} {ver} (constrained to {DEMO_CONSTRAINT}).\n"
        "\n"
        f"Now: skip_packages demo — installing {DEMO_PACKAGE}\n"
        f"while skipping '{DEMO_SKIP_DEP}' (one of its dependencies).\n"
        f"\n"
        f"First we uninstall {DEMO_SKIP_DEP} (if present) so we can\n"
        f"verify at the end that skip_packages actually prevented it.",
        windowTitle="Demo 5a: Constraints — done",
    )

    # --- Part 2: skip_packages ---
    fresh_slate()
    try:
        slicer.util.pip_uninstall(DEMO_SKIP_DEP)
    except Exception:
        pass

    skipped = slicer.util.pip_install(
        DEMO_PACKAGE,
        skip_packages=[DEMO_SKIP_DEP],
        requester="Feature Tour (skip_packages)",
    )

    lines = [f"  {s}" for s in (skipped or [])]
    print(f"[Tour] Skipped packages:\n" + "\n".join(lines))

    # Verify the metadata scrub: re-install normally, check the skipped dep
    slicer.util.pip_install(DEMO_PACKAGE, requester="Feature Tour (verify scrub)")
    dep_installed = slicer.packaging.pip_check(Requirement(DEMO_SKIP_DEP))

    slicer.util.infoDisplay(
        f"Skipped {len(skipped or [])} package(s).\n"
        f"{DEMO_SKIP_DEP} installed after normal re-install: {dep_installed}\n"
        "\n"
        "Re-installing normally did NOT pull in the skipped dependency\n"
        "because the metadata scrub removed it.",
        windowTitle="Demo 5b: skip_packages — verified",
    )


# ---------------------------------------------------------------------------
# Cleanup
# ---------------------------------------------------------------------------

def do_cleanup():
    """Restore the user's environment."""
    if pkg_version_before_tour is not None:
        msg = (
            f"{DEMO_PACKAGE} {pkg_version_before_tour} was installed before\n"
            "the tour. Restore it now?"
        )
    else:
        msg = f"Uninstall {DEMO_PACKAGE} to leave your environment clean?"

    if slicer.util.confirmYesNoDisplay(msg, windowTitle="Feature Tour — Cleanup"):
        if pkg_version_before_tour is not None:
            uninstall_pkg()
            slicer.util.pip_install(
                f"{DEMO_PACKAGE}=={pkg_version_before_tour}",
                requester="Feature Tour (restore)",
            )
        else:
            uninstall_pkg()

    slicer.util.infoDisplay(
        "Tour complete! For full API docs:\n"
        "  help(slicer.packaging.pip_ensure)",
        windowTitle="Feature Tour — Done",
    )


# ---------------------------------------------------------------------------
# Menu and main loop
# ---------------------------------------------------------------------------

MENU_ITEMS = [
    "1. Loading Dependencies",
    "2. Checking Requirements",
    "3. Installing with Progress",
    "4. Smart Install Workflow",
    "5. Advanced Options",
    "---",
    "Run All",
    "Exit Tour",
]

DEMO_FUNCS = {
    MENU_ITEMS[0]: demo_loading_deps,
    MENU_ITEMS[1]: demo_checking_reqs,
    MENU_ITEMS[2]: demo_install_progress,
    MENU_ITEMS[3]: demo_smart_install,
    MENU_ITEMS[4]: demo_advanced_options,
}

DEMO_ORDER = MENU_ITEMS[:5]


def run_tour():
    slicer.util.infoDisplay(
        "Welcome to the PR #9010 Feature Tour!\n"
        "\n"
        "Pick demos from the menu. Uses scikit-image as a demo package\n"
        "(configurable via DEMO_PACKAGE at the top of the script).\n"
        'Select "Run All" for the full experience.',
        windowTitle="Feature Tour — Welcome",
    )

    while True:
        dialog = qt.QInputDialog(slicer.util.mainWindow())
        dialog.setWindowTitle("Feature Tour")
        dialog.setLabelText("Choose a demo:")
        dialog.setComboBoxItems(MENU_ITEMS)
        dialog.setComboBoxEditable(False)

        if dialog.exec_() != qt.QDialog.Accepted:
            break

        choice = dialog.textValue()

        if choice == "Exit Tour" or choice == "---":
            break

        if choice == "Run All":
            for key in DEMO_ORDER:
                DEMO_FUNCS[key]()
        else:
            func = DEMO_FUNCS.get(choice)
            if func:
                func()

    do_cleanup()


# ---------------------------------------------------------------------------
# Start the tour
# ---------------------------------------------------------------------------

run_tour()
```

</details>

---

<details>
<summary><strong>References (click to expand)</strong></summary>

- This work is part of the [44th Slicer project week](https://projectweek.na-mic.org/PW44_2026_GranCanaria/Projects/PythonDependenciesInExtensions/)!
- [#7171 — Improving Support for Python Package Dependencies in Slicer Extensions](https://github.com/Slicer/Slicer/issues/7171) — This PR implements the "runtime installation" approach described in the issue: extensions declare dependencies via `requirements.txt`, and the new `slicer.packaging` module handles checking and installation with optional constraints file support for coordinating versions across extensions.
- [#7707 — Allow scripted modules to declare and lazily install pip requirements](https://github.com/Slicer/Slicer/issues/7707) — Proposes a `LazyImportGroup` context manager that intercepts imports and triggers installation on first attribute access. This PR takes a simpler, more explicit approach: developers call `pip_ensure()` at the point dependencies are needed, then import normally. The explicit pattern trades some elegance for transparency and easier debugging. The `LazyImportGroup` approach could potentially be built on top of the primitives provided by `slicer.packaging` (`load_requirements`, `pip_check`, `pip_install` with callbacks) if desired in the future. See #8181.
- At https://github.com/KitwareMedical/SlicerNNUnet/pull/21 is a demo of how some of the features here can simplify python dependency handling in the NNUnet extension.

</details>

### DIFF
--- a/Docs/developer_guide/python_faq.md
+++ b/Docs/developer_guide/python_faq.md
@@ -25,7 +25,7 @@ modules, are not available.
 :::
 
 :::{tip}
-To install additional packages, you can use the {func}`slicer.util.pip_install()` function.
+To install additional packages, you can use the {func}`slicer.packaging.pip_install()` function.
 :::
 
 ## What is the Python Console?
@@ -50,7 +50,7 @@ For example, this applies to the [Segment Editor effects](/user_guide/modules/se
 :::
 
 :::{tip}
-To install additional packages, you can use the {func}`slicer.util.pip_install()` function.
+To install additional packages, you can use the {func}`slicer.packaging.pip_install()` function.
 :::
 
 :::{versionchanged} 5.2.0
@@ -350,14 +350,33 @@ In our example:
 
 You can install any Python package within Slicer's built-in Python environment.
 
-The convenience function {func}`slicer.util.pip_install` can be used to install packages into your Slicer module. To understand its usage, examine the [Install a Python package](/developer_guide/script_repository.md#install-a-python-package) example within the Script Repository.
+### Utility functions for dependency management
+
+:::{versionadded} 5.11
+The `slicer.packaging` module with `pip_ensure`, `pip_check`, `load_requirements`, and `load_pyproject_dependencies`.
+:::
+
+Slicer provides several utility functions to help manage Python dependencies:
+
+| Function | Purpose |
+|----------|---------|
+| `slicer.packaging.load_requirements` | Load a `requirements.txt` file |
+| `slicer.packaging.pip_check` | Check if requirements are already satisfied |
+| `slicer.packaging.pip_ensure` | Ensure requirements are installed, prompting the user if needed |
+| `slicer.packaging.pip_install` | Install packages |
+
+The recommended approach is to use {func}`slicer.packaging.pip_ensure`, which handles checking, user confirmation, and installation with progress display in one call. It also automatically skips installation when running in testing mode (`slicer.app.testingEnabled()`).
+
+For direct installation with visual feedback, use {func}`slicer.packaging.pip_install` with `show_progress=True`. This displays a modal progress dialog during installation.
+
+For detailed usage examples, see [Install a Python package](/developer_guide/script_repository.md#install-a-python-package) in the Script Repository.
 
 :::{warning}
 Since installing packages can have side effects on other extensions or the main application, here are some best practices to adhere to:
 
 **DO:**
-* ✅ Always include a confirmation dialog that clearly communicates the installation process, mirroring the approach in the linked example.
-* ✅ Document the dependencies your module relies upon.
+* ✅ Ask the user before installing. {func}`slicer.packaging.pip_ensure` does this for you; if you go through the lower-level {func}`slicer.packaging.pip_install`, write your own confirmation dialog as shown in the linked example.
+* ✅ Document the dependencies your module relies upon using a `requirements.txt` file.
 * ✅ Consider specifying version requirements using `>=X.Y` to avoid incompatible versions.
 * ✅ Verify that all Python packages are distributed as Python wheels. This is particularly important for dependencies including compiled code, as installing a wheel eliminates the need for users to install a compiler.
 
--- a/Applications/SlicerApp/Testing/Python/CMakeLists.txt
+++ b/Applications/SlicerApp/Testing/Python/CMakeLists.txt
@@ -320,6 +320,18 @@ slicer_add_python_unittest(
   )
 
 slicer_add_python_unittest(
+  SCRIPT ${Slicer_SOURCE_DIR}/Base/Python/slicer/tests/test_slicer_util_pip.py
+  SLICER_ARGS --no-main-window --testing --disable-modules
+  TESTNAME_PREFIX nomainwindow_
+  )
+
+slicer_add_python_unittest(
+  SCRIPT ${Slicer_SOURCE_DIR}/Base/Python/slicer/tests/test_slicer_packaging.py
+  SLICER_ARGS --no-main-window --testing --disable-modules
+  TESTNAME_PREFIX nomainwindow_
+  )
+
+slicer_add_python_unittest(
   SCRIPT ${Slicer_SOURCE_DIR}/Base/Python/slicer/tests/test_slicer_parameter_node_wrapper.py
   SLICER_ARGS --no-main-window --disable-cli-modules --disable-scripted-loadable-modules
   TESTNAME_PREFIX nomainwindow_
--- a/Applications/SlicerApp/Testing/Python/SlicerStartupPreserveSysPathTestHelperScript.py
+++ b/Applications/SlicerApp/Testing/Python/SlicerStartupPreserveSysPathTestHelperScript.py
@@ -1,6 +1,8 @@
 import sys
 import os
 
+from slicer.util import EXIT_SUCCESS
+
 SENTINEL = "/test/slicer-start-preserve-sys-path/directory"
 if SENTINEL not in sys.path:
     raise RuntimeError(f"{SENTINEL} not in sys.path as expected")
--- a/Base/Python/CMakeLists.txt
+++ b/Base/Python/CMakeLists.txt
@@ -17,9 +17,9 @@ set(Slicer_PYTHON_SCRIPTS
   slicer/parameterNodeWrapper/wrapper
   slicer/packaging
   slicer/ScriptedLoadableModule
-  slicer/slicerqt
   slicer/testing
   slicer/util
+  slicerqt
   mrml
   vtkAddon
   vtkITK
--- a/Base/Python/CMakeLists.txt
+++ b/Base/Python/CMakeLists.txt
@@ -15,6 +15,7 @@ set(Slicer_PYTHON_SCRIPTS
   slicer/parameterNodeWrapper/util
   slicer/parameterNodeWrapper/validators
   slicer/parameterNodeWrapper/wrapper
+  slicer/packaging
   slicer/ScriptedLoadableModule
   slicer/slicerqt
   slicer/testing
--- a/Base/Python/slicer/i18n.py
+++ b/Base/Python/slicer/i18n.py
@@ -11,9 +11,11 @@ def translate(context, text):
     This `translate(context, message)` function is recognized by Qt's lupdate tool.
     See example use in DICOM module.
     """
-    from slicer import app
-
-    return app.translate(context, text)
+    try:
+        from slicer import app
+        return app.translate(context, text)
+    except ImportError:
+        return text
 
 
 def getContext(sourceFile):
--- a/Base/Python/slicer/packaging.py
+++ b/Base/Python/slicer/packaging.py
@@ -1,0 +1,1283 @@
+"""Utilities for managing Python package dependencies in 3D Slicer.
+
+This module provides functions for checking, installing, and managing Python
+packages in the Slicer environment. It is the canonical home for all pip-related
+functionality.
+
+**Key functions:**
+
+- :func:`load_requirements` / :func:`load_pyproject_dependencies` -- parse dependency files
+- :func:`pip_check` -- check if requirements are satisfied (pure Python, no subprocess)
+- :func:`pip_ensure` -- high-level: check, prompt, install, restart detection
+- :func:`pip_install` / :func:`pip_uninstall` -- install/uninstall packages
+
+For backward compatibility, :func:`pip_install` and :func:`pip_uninstall` are also
+available as ``slicer.util.pip_install`` and ``slicer.util.pip_uninstall``.
+"""
+
+from __future__ import annotations
+
+import importlib
+import importlib.metadata
+import logging
+import os
+import shlex
+import shutil
+import sys
+import tomllib
+from pathlib import Path
+from subprocess import CalledProcessError
+from typing import TYPE_CHECKING
+
+from packaging.markers import default_environment
+from packaging.requirements import InvalidRequirement, Requirement
+from packaging.utils import canonicalize_name
+
+from slicer.i18n import tr as _
+from slicer.util import launchConsoleProcess, logProcessOutput
+
+if TYPE_CHECKING:
+    from collections.abc import Callable
+
+    import qt
+
+
+def _executePythonModule(
+    module: str,
+    args: list[str],
+    blocking: bool = True,
+    logCallback: Callable[[str], None] | None = None,
+    completedCallback: Callable[[int], None] | None = None,
+) -> None:
+    """Execute a Python module as a script in Slicer's Python environment.
+
+    Internally python -m is called with the module name and additional arguments.
+
+    :param module: Python module name to execute (e.g., "pip")
+    :param args: command-line arguments to pass to the module
+    :param blocking: If True (default), block until completion. If False, return immediately.
+    :param logCallback: When blocking=False, called with each line of output.
+    :param completedCallback: When blocking=False, called when process completes.
+    :raises RuntimeError: if PythonSlicer executable not found
+    :raises CalledProcessError: in blocking mode, if process fails
+    """
+    # Determine pythonSlicerExecutablePath
+    try:
+        from slicer import app  # noqa: F401
+
+        # If we get to this line then import from "app" is succeeded,
+        # which means that we run this function from Slicer Python interpreter.
+        # PythonSlicer is added to PATH environment variable in Slicer
+        # therefore shutil.which will be able to find it.
+        pythonSlicerExecutablePath = shutil.which("PythonSlicer")
+        if not pythonSlicerExecutablePath:
+            raise RuntimeError("PythonSlicer executable not found")
+    except ImportError:
+        # Running from console
+        pythonSlicerExecutablePath = os.path.dirname(sys.executable) + "/PythonSlicer"
+        if os.name == "nt":
+            pythonSlicerExecutablePath += ".exe"
+
+    commandLine = [pythonSlicerExecutablePath, "-m", module, *args]
+
+    if blocking:
+        proc = launchConsoleProcess(commandLine, useStartupEnvironment=False)
+        logProcessOutput(proc, logCallback=logCallback)
+    else:
+        launchConsoleProcess(
+            commandLine,
+            useStartupEnvironment=False,
+            blocking=False,
+            logCallback=logCallback,
+            completedCallback=completedCallback,
+        )
+
+
+def load_requirements(path: str | Path) -> list[Requirement]:
+    """Load requirements from a requirements.txt file.
+
+    Parses a requirements file and returns a list of Requirement objects
+    that can be used with :func:`pip_check` and :func:`pip_ensure`.
+
+    :param path: Path to the requirements.txt file. Can be a string or Path object.
+
+    :returns: List of :class:`packaging.requirements.Requirement` objects.
+
+    Lines starting with ``#`` (comments), empty lines, and lines starting with
+    ``-`` (pip options like ``-r``, ``-c``, ``--index-url``) are skipped.
+
+    Example:
+
+    .. code-block:: python
+
+      from pathlib import Path
+      reqs = slicer.packaging.load_requirements(Path(__file__).parent / "requirements.txt")
+      slicer.packaging.pip_ensure(reqs, requester="MyExtension")
+
+    """
+    reqs = []
+    with open(path) as f:
+        for line in f:
+            line = line.strip()
+            # Skip comments, empty lines, and pip options (-r, -c, --index-url, etc.)
+            if line and not line.startswith("#") and not line.startswith("-"):
+                reqs.append(Requirement(line))
+    return reqs
+
+
+def load_pyproject_dependencies(path: str | Path) -> list[Requirement]:
+    """Load dependencies from a pyproject.toml file.
+
+    Reads the ``[project.dependencies]`` list (`PEP 621
+    <https://peps.python.org/pep-0621/>`_) and returns
+    :class:`~packaging.requirements.Requirement` objects that can be used
+    with :func:`pip_check` and :func:`pip_ensure`.
+
+    Only the ``dependencies`` list inside the ``[project]`` table is read.
+    Other fields in the table (``name``, ``version``, etc.) are not read or
+    validated.
+
+    :param path: Path to the pyproject.toml file. Can be a string or Path object.
+
+    :returns: List of :class:`packaging.requirements.Requirement` objects.
+        Returns an empty list if ``[project]`` exists but has no
+        ``dependencies`` key.
+
+    :raises KeyError: If the file has no ``[project]`` table.
+
+    Example pyproject.toml::
+
+      [project]
+      dependencies = [
+          "numpy>=1.20",
+          "scikit-image>=0.20",
+      ]
+
+    Example:
+
+    .. code-block:: python
+
+      from pathlib import Path
+      reqs = slicer.packaging.load_pyproject_dependencies(
+          Path(__file__).parent / "pyproject.toml"
+      )
+      slicer.packaging.pip_ensure(reqs, requester="MyExtension")
+
+    """
+    with open(path, "rb") as f:
+        data = tomllib.load(f)
+
+    if "project" not in data:
+        msg = f"pyproject.toml has no [project] table: {path}"
+        raise KeyError(msg)
+
+    return [Requirement(dep) for dep in data["project"].get("dependencies", [])]
+
+
+def _to_requirements(
+    reqs: str | Requirement | list[str | Requirement],
+) -> list[Requirement]:
+    """Normalize flexible input into a list of Requirement objects.
+
+    Accepts a single string (one requirement, possibly with a marker; or
+    multiple simple requirements separated by whitespace), a single
+    Requirement, or a list of strings/Requirements. For multiple
+    requirements that include markers, use a list to avoid ambiguity.
+    """
+    if isinstance(reqs, str):
+        # Try the whole string as a single requirement first. This is what
+        # makes "foo; sys_platform == 'win32'" work -- markers contain
+        # whitespace and would be mangled by the split path below.
+        try:
+            return [Requirement(reqs)]
+        except InvalidRequirement:
+            pass
+        # Otherwise treat the string as multiple simple requirements
+        # separated by whitespace, e.g. "numpy>=1.0 scipy>=1.0".
+        return [Requirement(s) for s in reqs.split()]
+    if isinstance(reqs, Requirement):
+        return [reqs]
+    if isinstance(reqs, list):
+        return [Requirement(r) if isinstance(r, str) else r for r in reqs]
+    msg = f"Expected str, Requirement, or list, got {type(reqs).__name__}"
+    raise TypeError(msg)
+
+
+def pip_check(
+    req: str | Requirement | list[str | Requirement],
+    _seen: set[tuple[str, frozenset[str]]] | None = None,
+) -> bool:
+    """Check if requirement(s) are satisfied.
+
+    For requirements with extras like ``package[extra1,extra2]>=1.0``, this:
+
+    1. Checks if the base package is installed at an acceptable version
+    2. Finds which dependencies are activated by the requested extras
+    3. Recursively verifies those dependencies are satisfied
+
+    Markers (e.g., ``; sys_platform == 'win32'``) are evaluated - if a marker
+    doesn't apply to the current environment, the requirement is considered
+    satisfied (since it doesn't need to be installed).
+
+    :param req: Requirement(s) to check. Accepts a space-separated string
+        (e.g., ``"numpy>=1.20 scipy"``), a single
+        :class:`~packaging.requirements.Requirement`, or a list of strings
+        and/or Requirement objects.
+    :param _seen: Internal parameter for tracking circular dependencies.
+        Do not pass this parameter.
+
+    :returns: True if all requirements are satisfied, False otherwise.
+
+    Example:
+
+    .. code-block:: python
+
+      if slicer.packaging.pip_check("numpy>=1.20"):
+          print("numpy is satisfied")
+
+    """
+    from importlib.metadata import PackageNotFoundError, requires, version
+
+    if _seen is None:
+        _seen = set()
+        # Normalize flexible input types at the public entry point
+        reqs = _to_requirements(req)
+        return all(pip_check(r, _seen) for r in reqs)
+
+    # From here, req is a single Requirement (internal recursive call)
+
+    # Check if requirement's marker applies to current environment
+    # If not, consider it satisfied (doesn't need to be installed here)
+    if req.marker is not None:
+        env = default_environment()
+        if not req.marker.evaluate(env):
+            return True
+
+    # Avoid rechecking the same requirement
+    key = (req.name.lower(), frozenset(req.extras))
+    if key in _seen:
+        return True
+    _seen.add(key)
+
+    # Check if base package is installed at acceptable version
+    try:
+        installed = version(req.name)
+    except PackageNotFoundError:
+        return False
+    if installed not in req.specifier:
+        return False
+
+    # If no extras then we are done
+    if not req.extras:
+        return True
+
+    # Find dependencies activated by the requested extras
+    dep_strings = requires(req.name) or []
+    env = default_environment()
+    activated = []
+
+    for dep_str in dep_strings:
+        dep = Requirement(dep_str)
+        if dep.marker is None:
+            continue
+
+        # Check if any requested extra activates this dependency
+        for extra in req.extras:
+            if dep.marker.evaluate({**env, "extra": extra}):
+                # Strip marker before recursive check - we've already determined it applies
+                dep_str_no_marker = str(dep).split(";")[0].strip()
+                activated.append(Requirement(dep_str_no_marker))
+                break  # Don't check other extras for same dep
+
+    # Recursively verify all activated dependencies
+    return all(pip_check(dep, _seen) for dep in activated)
+
+
+# Module-level flag to track whether a non-blocking pip install is in progress.
+# Used to prevent concurrent non-blocking pip operations which could corrupt
+# the Python environment.
+_pip_install_in_progress: bool = False
+
+
+def isPipInstallInProgress() -> bool:
+    """Check if a non-blocking pip install is currently in progress.
+
+    Use this to check before starting an operation that might conflict with
+    an ongoing pip installation.
+
+    :returns: True if a non-blocking pip install is in progress, False otherwise.
+
+    Example:
+
+    .. code-block:: python
+
+      if slicer.packaging.isPipInstallInProgress():
+          slicer.util.warningDisplay(
+              "Package installation is in progress. Please wait."
+          )
+      else:
+          slicer.packaging.pip_install("scipy", blocking=False)
+
+    """
+    return _pip_install_in_progress
+
+
+def _get_installed_versions() -> dict[str, str]:
+    """Return ``{canonical_distribution_name: version}`` for all installed packages.
+
+    Calls :func:`importlib.invalidate_caches` first so that metadata changes made by
+    a pip subprocess are visible to this process.
+    """
+    importlib.invalidate_caches()
+    return {
+        canonicalize_name(dist.metadata["Name"]): dist.metadata["Version"]
+        for dist in importlib.metadata.distributions()
+    }
+
+
+def _find_updated_imported_packages(
+    before_versions: dict[str, str],
+    after_versions: dict[str, str],
+) -> list[tuple[str, str | None, str]]:
+    """Find packages that were updated AND had modules already imported.
+
+    Compares *before_versions* and *after_versions* (both returned by
+    :func:`_get_installed_versions`) and checks whether any changed distribution
+    has top-level import names present in :data:`sys.modules`.
+
+    :param before_versions: ``{canonical_dist_name: version}`` before installation.
+    :param after_versions: ``{canonical_dist_name: version}`` after installation.
+    :returns: List of ``(dist_name, old_version, new_version)`` tuples for packages
+        that have import names present in ``sys.modules`` and either changed version
+        or were freshly installed (``old_version`` is ``None`` in that case).
+    """
+    # Reverse mapping: canonical dist name -> set of top-level import names
+    dist_to_imports: dict[str, set[str]] = {}
+    for import_name, dist_names in importlib.metadata.packages_distributions().items():
+        for dn in dist_names:
+            dist_to_imports.setdefault(canonicalize_name(dn), set()).add(import_name)
+
+    result = []
+    for name, old_ver in before_versions.items():
+        new_ver = after_versions.get(name)
+        if new_ver is None or old_ver == new_ver:
+            continue
+        import_names = dist_to_imports.get(name, set())
+        if any(imp in sys.modules for imp in import_names):
+            result.append((name, old_ver, new_ver))
+
+    # Also check packages that were freshly installed (not in before_versions).
+    # If their import names are already in sys.modules, that means they were
+    # previously installed and imported, then uninstalled, then reinstalled.
+    # The old module objects are still in memory.
+    for name, new_ver in after_versions.items():
+        if name in before_versions:
+            continue  # Already handled above
+        import_names = dist_to_imports.get(name, set())
+        if any(imp in sys.modules for imp in import_names):
+            result.append((name, None, new_ver))
+
+    return result
+
+
+def pip_ensure(
+    requirements: str | Requirement | list[str | Requirement],
+    constraints: str | Path | None = None,
+    skip_packages: list[str] | None = None,
+    prompt_install: bool = True,
+    prompt_restart: bool = True,
+    requester: str | None = None,
+    skip_in_testing: bool = True,
+    show_progress: bool = True,
+) -> list[str] | None:
+    """Ensure requirements are satisfied, installing if needed.
+
+    Call at the point where dependencies are actually needed (e.g., in an
+    ``onApplyButton`` method). This function checks which requirements are
+    missing and installs them, optionally showing a confirmation dialog first.
+
+    After installation, if any updated packages were already imported in the
+    current session, a restart prompt is shown (the old versions remain loaded
+    in memory and may not work correctly until Slicer is restarted).
+
+    :param requirements: Requirement(s) to ensure. Accepts a space-separated
+        string (e.g., ``"flywheel-sdk>=1.0 numpy"``), a single
+        :class:`~packaging.requirements.Requirement`, or a list of strings
+        and/or Requirement objects. Can also be obtained from
+        :func:`load_requirements` or :func:`load_pyproject_dependencies`.
+    :param constraints: Path to a constraints file (string or Path object), or None.
+        When provided, passed to pip as ``-c constraints.txt`` during installation.
+        Constraints files use the same format as requirements files but only constrain
+        versions without triggering installation. Useful for ensuring compatible
+        versions across multiple extensions.
+    :param skip_packages: Discouraged workaround for transitive dependency conflicts;
+        see the script repository for guidance. Forwarded to :func:`pip_install`.
+    :param prompt_install: If True (default), show confirmation dialog before installing.
+    :param prompt_restart: If True (default), check whether any updated packages were
+        already imported and, if so, show a dialog recommending a restart. The user
+        can choose to restart immediately or continue without restarting.
+    :param requester: Name shown in dialog to identify who is requesting the packages
+        (e.g., "TotalSegmentator", "MyFilter", "MyExtension").
+    :param skip_in_testing: If True (default), skip installation when Slicer is running
+        in testing mode (``slicer.app.testingEnabled()``). This prevents tests from
+        modifying the Python environment. Set to False if your test explicitly
+        needs to verify installation behavior.
+    :param show_progress: If True (default), show progress dialog during installation
+        with status updates and collapsible log details. If False, show only
+        a busy cursor during installation.
+
+    :returns: When ``skip_packages`` is provided, a list of skipped requirement strings
+        (forwarded from :func:`pip_install`). Otherwise ``None``.
+
+    :raises RuntimeError: If user declines installation.
+    :raises subprocess.CalledProcessError: If installation fails.
+
+    Example:
+
+    .. code-block:: python
+
+      class MyFilterWidget(ScriptedLoadableModuleWidget):
+
+          def onApplyButton(self):
+              slicer.packaging.pip_ensure("scikit-image>=0.20", requester="MyFilter")
+              import skimage
+
+              # Now safe to use skimage
+              filtered = skimage.filters.gaussian(array, sigma=2.0)
+
+    For loading from a requirements file, use :func:`load_requirements`::
+
+      reqs = slicer.packaging.load_requirements(self.resourcePath("requirements.txt"))
+      slicer.packaging.pip_ensure(reqs, requester="MyFilter")
+
+    For more examples (constraints, skip_packages), see
+    :doc:`/developer_guide/script_repository` (Python package management section).
+
+    """
+    requirements = _to_requirements(requirements)
+    missing = [req for req in requirements if not pip_check(req)]
+
+    if not missing:
+        return None  # All satisfied
+
+    # Check if we're in full Slicer or PythonSlicer
+    if not _isSlicerAppAvailable():
+        # Running in PythonSlicer - just do simple install (prompt not available)
+        return pip_install(
+            [str(req) for req in missing],
+            constraints=constraints,
+            skip_packages=skip_packages,
+        )
+
+    import slicer
+
+    # Skip installation in testing mode to avoid modifying the environment
+    if skip_in_testing and slicer.app.testingEnabled():
+        missing_str = ", ".join(str(req) for req in missing)
+        logging.info(f"Testing mode is enabled: skipping pip_ensure for [{missing_str}]")
+        return None
+
+    if prompt_install:
+        package_list = "\n".join(f"- {req}" for req in missing)
+        if requester:
+            title = _("{requester} - Install Python Packages").format(requester=requester)
+        else:
+            title = _("Install Python Packages")
+        count = len(missing)
+        if count == 1:
+            message = _(
+                "1 Python package needs to be installed.\n\n"
+                "This will modify Slicer's Python environment. Continue?",
+            )
+        else:
+            message = _(
+                "{count} Python packages need to be installed.\n\n"
+                "This will modify Slicer's Python environment. Continue?",
+            ).format(count=count)
+        if not slicer.util.confirmOkCancelDisplay(message, title, detailedText=package_list):
+            raise RuntimeError("User declined package installation")
+
+    # Snapshot installed versions before installation
+    before_versions = _get_installed_versions() if prompt_restart else None
+
+    # Install missing packages with optional progress display
+    skipped = pip_install(
+        [str(req) for req in missing],
+        constraints=constraints,
+        skip_packages=skip_packages,
+        blocking=True,
+        show_progress=show_progress,
+        requester=requester,
+    )
+
+    if not prompt_restart:
+        return skipped
+
+    # Check if any updated packages were already imported
+    after_versions = _get_installed_versions()
+    updated_imported = _find_updated_imported_packages(before_versions, after_versions)
+
+    if updated_imported:
+        detail_lines = [
+            f"- {name}: {old_ver} -> {new_ver}" if old_ver else f"- {name}: (reinstalled) -> {new_ver}"
+            for name, old_ver, new_ver in updated_imported
+        ]
+        detail_text = "\n".join(detail_lines)
+        logging.info(
+            f"Updated packages already imported (restart recommended): {detail_text}",
+        )
+
+        # confirmYesNoDisplay auto-returns True in testing mode, which would
+        # trigger an unwanted restart. Guard with testingEnabled() check and
+        # rely on the log message above for test verification.
+        if not slicer.app.testingEnabled():
+            count = len(updated_imported)
+            if requester:
+                title = _("{requester} - Restart Recommended").format(requester=requester)
+            else:
+                title = _("Restart Recommended")
+            if count == 1:
+                message = _(
+                    "1 updated package was already loaded in memory "
+                    "and may not work correctly until Slicer is restarted.\n\n"
+                    "Would you like to restart now?",
+                )
+            else:
+                message = _(
+                    "{count} updated packages were already loaded in memory "
+                    "and may not work correctly until Slicer is restarted.\n\n"
+                    "Would you like to restart now?",
+                ).format(count=count)
+            if slicer.util.confirmYesNoDisplay(
+                message, title, detailedText=detail_text,
+            ):
+                slicer.util.restart()
+
+    return skipped
+
+
+def _isSlicerAppAvailable() -> bool:
+    """Check if slicer.app is available (running in full Slicer, not PythonSlicer).
+
+    :returns: True if slicer.app is available, False if running in PythonSlicer console.
+    """
+    try:
+        from slicer import app
+        return app is not None
+    except ImportError:
+        return False
+
+
+def pip_install(
+    requirements: str | list[str],
+    constraints: str | Path | None = None,
+    no_deps_requirements: str | list[str] | None = None,
+    skip_packages: list[str] | None = None,
+    blocking: bool = True,
+    show_progress: bool = True,
+    requester: str | None = None,
+    parent: qt.QWidget | None = None,
+    logCallback: Callable[[str], None] | None = None,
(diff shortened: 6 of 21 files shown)


allemangD:
This looks related to https://github.com/Slicer/Slicer/pull/8181 which is/was nearly done, but still needed some cmake/automation written to generate the core constraints file for the current Slicer build config. (eg. if we update the version of Numpy that is packaged with Slicer, then the generated core constraints file needs to also be updated so that package installations correctly resolve to the new numpy version.)

That PR is also pretty stale at this point.. it is all in the Python layer so there *should* not be anything broken there but I have not rebased on main in a long time.

ebrahimebrahim:
@allemangD Oh somehow I missed #8181!
Incorporating `uv` is an amazing thing to do!
I wonder if they are almost compatible actually as this work is more just expanding on the lower level utilities.
You could still stick `uv` in there, or do the lazy import thing, while making use of the changes here, potentially

pieper:
We need to be sure to handle the case of a headless pip_install (i.e. silently skip the dialog and just provide text progress).

ebrahimebrahim:
> We need to be sure to handle the case of a headless pip_install (i.e. silently skip the dialog and just provide text progress).

When I try

```
./PythonSlicer -c "import slicer.util; slicer.util.pip_uninstall('requests')"
./PythonSlicer -c "import slicer.util; slicer.util.pip_install('requests')"
```

this does work currently because in pip_install and other `pip_` commands we [check if we are in headless mode](https://github.com/ebrahimebrahim/Slicer/blob/8001cafc15e589f0d345717984127cf78f368be2/Base/Python/slicer/util.py#L4356-L4360):

```py
    if not _isSlicerAppAvailable():
        # Running in PythonSlicer - use simple blocking mode
        _pip_install_simple(requirements, constraints, no_deps_requirements)
        return
```

[Here](https://github.com/ebrahimebrahim/Slicer/blob/8001cafc15e589f0d345717984127cf78f368be2/Base/Python/slicer/util.py#L4392-L4401) is how we do the check:

```py
def _isSlicerAppAvailable():
    """Check if slicer.app is available (running in full Slicer, not PythonSlicer).


    :returns: True if slicer.app is available, False if running in PythonSlicer console.
    """
    try:
        from slicer import app  # noqa: F401
        return True
    except ImportError:
        return False
```

Does this cover what you mean by "headless"?

mhalle:
I would still recommend moving to a modern uv/pyproject.toml solution with backward compatibility. requirements.txt isn't sufficiently descriptive for the things you'd like to do like separating out dev dependencies. 

UV is so much faster than pip and offers much better dependency resolution. It would be lovely to be able to do static analysis of dependencies, and potentially roll back changes. 

Yes, there are challenges, but it would be nice to have a solution that's in the direction of where the python community is going.

ebrahimebrahim:
> I would still recommend moving to a modern uv/pyproject.toml solution with backward compatibility. requirements.txt isn't sufficiently descriptive for the things you'd like to do like separating out dev dependencies. 
> 
> UV is so much faster than pip and offers much better dependency resolution. It would be lovely to be able to do static analysis of dependencies, and potentially roll back changes. 
> 
> Yes, there are challenges, but it would be nice to have a solution that's in the direction of where the python community is going.


Regarding uv: absolutely 💯! This is just the start.

Regarding toml: There is a justification in # the design rationale section for why that's not necessarily the right choice for dependency specification, see the "Why requirements.txt instead of pyproject.toml?" section

mhalle:
Pip understands pyproject.toml as well, so that isn't the distinguishing factor. Most of python's modern dependency machinery benefits from the richness of pyproject files, so even if modules aren't strictly modules (and perhaps they could be), that isn't a gating factor.

ebrahimebrahim:
Hmm, I am not certain what's the right choice for the dependency specification file format. My current thoughts:

If we're _only_ using the file for dependency specification then I'd say toml doesn't add anything and just throws in more for the user to think about ("is this enough? do i need to add more metadata? what will be used?")

On the other hand if we are planning to create our own custom rich toml format and parse more information out of it (e.g. constraints, things that should be installed with various flags like --no-deps, etc.) then maybe we should go with toml?

ebrahimebrahim:
Question to consider in this PR: after a `pip_install`, should a prompt show up to restart the application? In which of the "modes" of `pip_install` should that happen?

pieper:
> should a prompt show up to restart the application?

No, definitely not.  Most pip installs do not require restart.

lassoan:
The only case when a restart may be needed is when a package that is already imported by some module gets updated. The Slicer module that triggered this update may want to know this, so that it can suggest the user to restart the application. It would be nice if the install function returned a list of updated python packages that had been already imported, and maybe there could be a flag that would show a pop-up suggesting application restart to the user; with "Show details" describing due to which packages this is happening; allowing the users to choose between restarting the application or ignoring the suggestion.

ebrahimebrahim:
Notable updates:

- @lassoan Added a `skip_packages` option to pip_install that helps to handle the difficult situations in NNUnet and TotalSegmentator. A working example of how it can simplify things is at https://github.com/KitwareMedical/SlicerNNUnet/pull/21
- @pieper, @lassoan Added a restart prompt at the end of the `pip_ensure` workflow which only shows up if there were packages whose version was modified and that were previously imported, necessitating a restart.
- Added type annotations to the code that I am adding, to at least work a tiny bit towards #8226
- @mhalle Added `load_pyproject_dependencies()` as a sibling to `load_requirements()`. It reads dependencies from a pyproject.toml via `tomllib` and returns the same `list[Requirement]` type that `load_requirements()` returns, so the downstream API (`pip_check`, `pip_ensure`) works identically with either format. Extensions can use whichever they prefer. The `requirements.txt` path remains the primary recommendation since extensions aren't Python packages and it requires no boilerplate, but pyproject.toml is there for extensions that want the modern standard. Adopting uv is a separate decision from the file format and is out of scope for this PR -- but I am super pro-doing-that.

Additional comment:

- @allemangD Regarding #8181, I think it is mostly complementary rather than conflicting. Here we are focused on expanding the low-level tooling, while #8181 focuses on rethinking the framework at a higher level. The major code conflict is `slicer.util.pip_install`: here I extend it in-place, while #8181 replaces it with a `slicer.packaging.pip_install` backed by `uv`. These are incompatible APIs that would need to be reconciled. The most valuable pieces to salvage from #8181 are the lazy proxy module system (which could call `pip_ensure` under the hood), the global constraint registry, and the `External_python-uv.cmake` superbuild module. The last of these is directly reusable since swapping `pip` for `uv pip` is a small localized change to `_executePythonModule` and `_build_pip_args`.

This is ready for review!

ebrahimebrahim:
Todo from discussion today:
- [x] Split things out into `slicer.packaging`
- [x] Look into reducing docstring verbosity; e.g. maybe some examples can become tests.

(@lassoan I will let you know when these are done and this is ready for review again)

ebrahimebrahim:
@Thibault-Pelletier I took to heart your comment on long docstrings. A couple of examples were moved to being tests, but most were already tested in some form. Many examples were moved to a [new script-repository section](https://slicer--9010.org.readthedocs.build/en/9010/developer_guide/script_repository.html#python-package-management). I think the [API docs](https://slicer--9010.org.readthedocs.build/en/9010/developer_guide/slicer.html#module-slicer.pydeps) look much better now, much more readable as a reference, so ty for the suggestion!

@lassoan I split out all the pip install related functionality into a new module `slicer.pydeps`. With `slicer.packaging` while testing it I found that there are some cases where the `packaging` name conflicts with [packaging](https://pypi.org/project/packaging/) and causes issues. We can continue to discuss what's a good name, but I liked `pydeps` because it is short like `util`, it clearly communicates that this is about _python_ dependencies specifically, and it reads well in commands: `slicer.pydeps.pip_ensure(...)`, `slicer.pydeps.load_requirements(...)`, etc.

@lassoan This should be ready for your review now, thank you!

Docs/developer_guide/python_faq.md
```diff
@@ -350,16 +350,36 @@ In our example:
 
 You can install any Python package within Slicer's built-in Python environment.
 
-The convenience function {func}`slicer.util.pip_install` can be used to install packages into your Slicer module. To understand its usage, examine the [Install a Python package](/developer_guide/script_repository.md#install-a-python-package) example within the Script Repository.
+### Utility functions for dependency management
+
+:::{versionadded} 5.11
```
ebrahimebrahim, line 355: I saw we put these `:::{versionadded}` things in the docs sometimes, but elsewhere it references _stable_ versions only. I put 5.11 here but that's vague -- some 5.11's will have it and some will not. Should we put the name of the next stable version? Do we know the name of the next stable version? 5.12 or 6.0?

ebrahimebrahim:
Would like to get this on the path to integration. No urgency-- I just want to keep it moving along and not forget about it. It can be tested with the snippets provided in the snippets section of the PR description. @lassoan Let me know if there's anything else that would help the review process. TY!

We can finalize the exact name and package-organization of the module after the review and before merging.

ebrahimebrahim:
@lassoan poking to get this looked at :smile:

ebrahimebrahim:
@lassoan let me know if you will have a chance to review this. The testing snippets section in the PR description should help to just see how things work.

(And before merging, but after your review, I will do one more re-organization just to rename the package since we did not like `pydeps`)

lassoan (commented):
Sorry for the delayed response. This is really nice work and it will make Python package installation much simpler and more robust!

I just have a few comments that would be nice to address:
 
1. There are so many comments that it is not clear what were the high-level stages of the development. It would be better to have just one or few commits and document any details that are needed to interpret the code in comments in the code instead of commit comments.
5. It would be nice to simplify the syntax:In addition to this nicely structured but complex syntax:

```
import slicer.pydeps
from packaging.requirements import Requirement
pydeps.pip_ensure([Requirement("flywheel")])
```
We should also allow this trivial syntax:

```
import slicer.pydeps
pydeps.pip_ensure("flywheel")
```

and

```
pydeps.pip_ensure(requirements_file=".../requirements.txt")
```

and

```
if slicer.pydeps.pip_check("flywheel numpy>=2.0"):
    ...
```

It is OK if the simple string syntax does not support everything that `pip install`, it should just allow specifying simple dependencies in the usual simple syntax.

3. LLMs litter the source code with unicode characters. While unicode support has improved in recent years, it may still cause various errors. Let's keep using only ASCII characters in the source code.
 
4. Do not require qt. If qt is available then use it (e.g., prompt the user using a popup window), but make these very useful classes available without qt as well. 

5. All strings that may be displayed to the user must be internationalized (using the `_()` function). See https://github.com/SoniaPujolLab/SlicerLanguagePacks/blob/main/DevelopersManual.md. Text intended only for developers can remain non-translatable. Text intended for very advanced users can remain non-translatable. Application log messages should not be translated.

6. Use these new functions in the Python scripted loadable module template and the Python scripted CLI module template.

7. Finally, just a reminder to not forget the renaming of `pydeps` to `packaging`.

ebrahimebrahim:
Andras, ty for the excellent review. All good points

The name `packaging` isn't so good because there is another package also called packaging and the names can be confusing. There is one point even in this code itself where a name conflict arises:

> Slicer's Python environment adds the `slicer/` directory itself to `sys.path`. This means a file named `slicer/packaging.py` shadows the third-party packaging library for any top-level `from packaging import ...` statement, breaking imports of `packaging.requirements`, `packaging.markers`, etc.

Thibault-Pelletier:
> > Slicer's Python environment adds the `slicer/` directory itself to `sys.path`. This means a file named `slicer/packaging.py` shadows the third-party packaging library for any top-level `from packaging import ...` statement, breaking imports of `packaging.requirements`, `packaging.markers`, etc.

Shouldn't this environment change be fixed and removed?

In the different code snippets the utils, and other sub modules / packages, are always imported using the `slicer` package as reference. It also applies to the `parameterNodeWrapper` package and other kits.

mhalle:
I dug into this a little, and it's a lot of good work to address a real problem.

There are some concerns and functionality gaps though. First, the PR re-implements the dependency resolver that pip and uv build in in order to implement skip packages. That means recursive tree walks. 

The PR scrubs the METADATA to fix things up. But that has a lot of risks. It says "this project's dependencies are met" when they aren't, and there's no way to later figure out what those dependencies actually were. It couples dependencies for a package on the state of the environment at the time when you run the install with no record of that dependency.

For example, say there's a package dependency on SimpleITK. It conficts with slicer's SimpleITK technically, but it works. the PR scrubs the package's metadata and removes that dependency. pip show nnunetv2 won't reveal the dependency in SimpleITK, so the impact of the install is unforeseeable. 

The PR also puts the burden of anticipating conflicts on the caller of pip_install to know about Slicer's installed libraries. By default, it maintains the same behavior as now, which can harm the Slicer installation. Without skip_packages provided, pip_install and pip_ensure offer no protection against overwriting Slicer's bundled packages. Protection is opt-in and requires the caller to know what to skip.

At the same time, skip lists for extensions will go stale over time and would have to be tracked by extension writers to keep up with Slicer's ecosystem.

Here's Claude's suggestion, which on the face of it makes sense:

An alternative would be to auto-generate constraints from Slicer's current environment. Before any pip install, snapshot every installed package and version, write them as name==version pins to a temporary constraints file, and pass it to pip with -c. pip's own resolver then protects the environment — it won't upgrade or downgrade anything already installed. No metadata scrubbing, no recursive tree walk, no skip list to maintain.

That allows the no-new-args `pip_install` to protect the Slicer environment without callers having to do anything new.

If a requirement genuinely conflicts with what Slicer has, pip errors — which is the right behavior, since the conflict is real. For the rare case where you know it works despite the version mismatch, `no_deps_requirements` (already in the PR) handles it.

This makes protection the default rather than opt-in, and it works with both pip and uv without extension developers needing to know what Slicer bundles.

This also matters when Slicer itself upgrades a bundled package in a new release. With scrubbed metadata, there's no way to find out which installed extensions depended on the old version. With intact metadata, you can query the dependency graph after a Slicer upgrade and identify what might break.

allemangD:
> There are some concerns and functionality gaps though. First, the PR re-implements the dependency resolver that pip and uv build in in order to implement skip packages. 
>
> The PR scrubs the METADATA to fix things up. But that has a lot of risks. It says "this project's dependencies are met" when they aren't, and there's no way to later figure out what those dependencies actually were. 
>
> [...]
>
> For example, say there's a package dependency on SimpleITK. It conficts with slicer's SimpleITK technically, but it works. the PR scrubs the package's metadata and removes that dependency. pip show nnunetv2 won't reveal the dependency in SimpleITK, so the impact of the install is unforeseeable.
>
> [...]
>
> Here's Claude's suggestion, which on the face of it makes sense: [...]

@mhalle Indeed, that is the approach we started to use with #8181. As I recall, generating a "core constraints" file was the only thing still blocking that merge. However due to time constraints on my end that PR has sat unmerged and stale since then.

@ebrahimebrahim I suggest removing the "skip" functionality from the scope of this PR entirely, so that this PR strictly focuses on the ergonomics and traceability of package installation. The current behavior that extension can arbitrarily break the Python environment would not change, but the improved logging and similar ergonomics should make it easier to debug.

Then, either by reviving #8181 or by closing it and building a similar constraints implementation based on this PR, we can use a *proper* dependency resolver like `pip` or `uv` with *proper* core-constraints generation in order to guarantee environment correctness as a separate feature.

That could be two parts of one PR, or split into two distinct PRs, but either way I think it's important to consider *ergonomics* and *environment correctness* as two distinct changes.

mhalle:
Claude suggested an alternative implementation that protects slicer packages. I can put that up as an alternative PR, but this isn't my area of expertise so I don't want this to seem like I am steamrolling anything, just throwing out some ideas.

jamesobutler:
Note that as of https://github.com/Slicer/Slicer/commit/f87c977465f3b63219456186b7673f5c12e8b514, the previous issue of installing a different SimpleITK package and it breaking things in Slicer core has been resolved by the switch to installing SimpleITK from PyPI. This is included in latest Slicer Preview 5.11. Preventing any Slicer extension from updating the Slicer core provided python versions is something to probably discuss as some extensions (and/or python packages used by said extension) may require say a specific numpy version different than Slicer core and may want to change it. Generally this is successful, but up to the developer to weigh compatibility with other things versus the features they are targeting. Not all extensions have compatible versions of Python packages with each other or with Slicer core.

ebrahimebrahim:
> There are some concerns and functionality gaps though. First, the PR re-implements the dependency resolver that pip and uv build in in order to implement skip packages. That means recursive tree walks.

It's not a full dependency resolver but just a recursive checker.
There are reasons for the partial re-implementation: we want something that avoids the overhead of a subprocess. It should be fast; `pip_ensure` might be called in a lot of module interactions and it would be convenient to make it cheap to call. `uv` is fast but can I call its dependency resolver without spawning a subprocess? With `pip` I definitely can't -- there's no stable python API for using for example `pip --dry-run`. The CLI is the API. Keeping the `pip_check` as a pure python enables relatively cheap checks. Besides, incorporating `uv` is not the job of this PR but the job of #8181. It can replace `pip_check` with a uv-based thing if the subprocess overhead isn't too bad.

> Here's Claude's suggestion, which on the face of it makes sense: [...]

I don't think it would have helped with the nnunet problem, because the package maintainer had set unreasonably strict version requirements that we needed to get around. If we had imposed on top of them the constraints that needed to be imposed by slicer, then there would just not exist a valid package resolution.

Since as James points out the nnunet problem has vanished, maybe it does make sense to get rid of `skip`. It is indeed not nice to sidestep and hack the python dependency installation framework. If it's no longer solving a problem we need to solve then we may want to get rid of it.

Other points discussed above I think are good and I'll bring them up to discuss at the slicer hangout tomorrow. Thanks everyone for the thoughts and feedback!!

lassoan:
> Since as James points out the nnunet problem has vanished

This happens all the time, unfortunately. Package maintainers often put in lots of unnecessary dependencies list that can cause installation failures, version conflicts, or just unnecessary downloads and storage. Pip has the `--no-deps` option, but that means that we need to take care of _all_ dependencies, while usually there are only 1-2 problematic packages.

Thibault-Pelletier:
In the case of nnUNet and other similar AI packages, a good option would be to be able to create virtual environments from the Slicer Python env. 

Since the extension needs to execute nnUNet in a subprocess, it doesn't need to pollute the core packages with its dependencies.

[ParaView has an option to run](https://www.kitware.com/using-python-virtual-environments-in-paraview-5-13-0/) while using another venv as override to share common Python packages while having local overrides in the specified venv (which would be necessary to share heavy packages such as torch).

This is obviously not to do in the scope of this PR but would resolve a lot of pain points.

lassoan:
Virtual environments are useful and they could be a solution when the main problem is incompatibility between dependencies and processing can run in a separat process. However, virtual environment is not always sufficient or optimal:
- Some packages specify dependencies that are need to be built from source and on many user computers that would fail, even in a virtual environment. If the problematic package is not actually required for the computations that the Slicer package depends on, then simply skipping the package resolves the issue.
- Some computations, such as mesh editing, mechanical simulation, registration, differentiable rendering are better to do in the Slicer process, because the computation is fast and sending updated data from a different process would add significant overhead.

Slicer extensions already need skipping installation of certain Python packages for various reasons (including are several that I maintain) and I don't think the need will go away. We could decide to ignore this need and not provide a solution in Slicer core, but overall we can save a lot of time if we provide one common high-quality implementation that everybody can use.

lassoan:
> If a requirement genuinely conflicts with what Slicer has, pip errors — which is the right behavior, since the conflict is real.

Most often version requirements are more strict than necessary. This is understandable, because many package maintainers choose to test their package with only certain version of dependencies. It is often safe to ignore too strict requirements, as long as proper behavior is tested at application level.

allemangD:
It's worth noting that uv supports dependency overrides: https://docs.astral.sh/uv/reference/cli/#uv-tool-run--overrides

The semantics are similar to constraints files, except that instead of raising a dependency resolution error on failure, they silently override the version of any and all transitive dependencies.

Dependencies can be held back via platform specifiers, for example `torch; sys_platform == 'never'`.

----

Generally I think the ["workspace" concept from uv](https://docs.astral.sh/uv/concepts/projects/workspaces/), combined with these dependency overrides, probably solves all our issues but would mandate a substantially different directory layout for scripted modules.

Suppose each scripted module had a companion `pyproject.toml` which listed the module's dependencies, optional dependencies, and dependency overrides. Then the "core" slicer environment could be a uv workspace which includes `members = ["qt-scripted-modules/*"]`. Then, at all times, uv workspace-level operations (`uv sync`, `uv add`, etc) would honor the declared dependencies and constraints for all modules. uv pip-level operations (`uv pip install`) would not.

So, `slicer.util.pip_install` *called from a scripted module* should use `uv add` against that module's pyproject.toml, generating it if necessary, and raising a developer warning that the module developer should declare the dependency explicitly in pyproject.toml.

It's a little unclear what it should do when *called from the python interactor.* Perhaps it should just use `uv pip install`? Or use `uv add` against the top-level workspace `pyproject.toml`? Or declare a dependency override?

The main blocker to this idea is that it requires each scripted module be laid out as a subdirectory of the workspace source. For example, instead of:

```
qt-scripted-modules
  modulefoo.py
  modulebar.py
```

It would need to change to:

```
qt-scripted-modules
  modulefoo
    modulefoo.py
    pyproject.toml   (autogenerate if not present)
  modulebar
    modulebar.py
    pyproject.toml   (autogenerate if not present)
```

---

I've also decided after further thought that #8181 probably includes too many changes to be merged at once.

- uv integration
- lazy import resolution
- requirements/constraints declaration.

The uv integration should probably be handled separately and soon, since `uv pip install` is a cheap drop-in replacement to `pip install`. The idea above and some other ideas in this discussion are predicated on its availability.

Lazy import resolution will soon(ish) be obviated by https://peps.python.org/pep-0810/. Generally I'd prefer to recommend extension developers use the built-in python syntax than my ad-hoc context manager.

Requirements/constraints declaration are the "primary" feature of #8181, but predicated on the other two.

ebrahimebrahim:
Hi @lassoan, thanks again for your review. Now ready for another review!

---

Changes since your last review:

- **String syntax**: `pip_check` and `pip_ensure` now accept strings (e.g. `pip_check("numpy>=1.0 scipy")`), not just `Requirement` objects.
- **Non-ASCII removed**: Unicode em dashes and so on have been removed from source code. Markdown doc files are still allowed to have some emojis (they already did)
- **Qt was not required**: This was already handled -- all public functions fall back to headless mode when Qt is unavailable. Verified in PythonSlicer. No changes needed.
  - **i18n bugfix**: `slicer.i18n.translate()` now degrades gracefully in PythonSlicer instead of crashing.
- **i18n**: User-facing dialog strings are now wrapped with `_()`.
- **Scripted loadable module template**: Added a commented-out `pip_ensure` example in `onApplyButton` and a blank `Resources/requirements.txt`. The Extension Wizard's file copier was extended to recognize `requirements.txt` so it ships with new modules.
- **Scripted CLI module template**: Decided not to do this. Scripted CLI modules run as headless subprocesses spawned by Slicer, so the natural caller-side place for `pip_ensure` doesn't exist there. The cleanest option would be to put `pip_ensure` at the top of the CLI script's `main()`, but this would install silently on first run with no user confirmation.
- **`skip_packages` discouraged**: Mainly just a change in the docs. It is now clear that it's a workaround for the narrow case where it's needed.
- **sys.path shadowing fix**: Moved `slicerqt.py` out of the `slicer/` package directory. Previously CTK's `executeFile` put `bin/Python/slicer/` on `sys.path[0]`, so every file inside the package shadowed same-named PyPI packages -- the reason we could not use `slicer.packaging`. The move is a worthwhile cleanup on its own: `slicerqt.py` is a top-level startup script, not a member of the `slicer` package, and treating it that way matches its actual role. See also [this fix within CTK](https://github.com/commontk/CTK/pull/1413) so that this doesn't happen again.
- **Console globals**: Replaced `from slicer.util import *` in the interactor startup with an explicit list (`getNode`, `getNodesByClass`, `array`, `mainWindow`, `exit`).
- **Module renamed**: `slicer.pydeps` is now `slicer.packaging`, the originally desired name. The earlier sys.path fix made this safe -- previously it would have shadowed the PyPI `packaging` library.
- **Commits squashed**: There are now just a few hopefully clear commits.
- Updated the PR description to reflect the above changes. The test snippets and everything should work.

lassoan (approved):
This should be good to go! We need this in many extensions. We can tweak it in follow-up commits if needed.

PR #9016, opened 2026-02-03
Add option for logarithmic color mapping

ajallphin:
# TLDR Version
This PR subclasses the current vtkImageMapToWindowLevelColors to add an option for logarithmic color mapping within vtkMRMLScalarVolumeDisplayNodes:
image

# Full Version
## Terminology Specification
This PR relates to the mapping of scalar vtkImageData to RGB colors for vtkMRMLScalarVolumeDisplayNodes. When I refer to "logarithmic color mapping," the following image presents what I mean.
image
In some applications, it is beneficial to map colors in this nonlinear way to enhance small image intensity differences.

## Current color mapping approach (as of c6a6e4eb7aaa6be318b18438757d26e3f8510cfd)
As of c6a6e4eb7aaa6be318b18438757d26e3f8510cfd, the primary components of the scalar mapping pipeline are shown here:
<table>
  <tr>
    <td>Reference Figure</td>
    <td>Label</td>
    <td>Description</td>
  </tr>
  <tr>
    <td rowspan="3">image
    <td>A</td>
    <td>Input ImageData</td>
</td>
  <tr>
    <td>B</td>
    <td>Output from "MapToWindowLevelColors" (ImageData rescaled from 0 - 255)</td>
  </tr>
  <tr>
    <td>C</td>
    <td> ColorNode lookup table used by "MapToColors" to assign each value in B an RGB value.</td>
  </tr>
</table>

And the relevant code here:
https://github.com/Slicer/Slicer/blob/c6a6e4eb7aaa6be318b18438757d26e3f8510cfd/Libs/MRML/Core/vtkMRMLScalarVolumeDisplayNode.cxx#L72-L82

## This PR
This PR's implementation is summarized here:
image

This PR changes the middle section which rescales the image data based on the current window/level. 

The linear mapping equation (unchanged):
<p align="center">
${LookupIndices}_{(0−255)}=\left(\frac{ImageData −WindowMIN}{WINDOW}\right)∗255.0$
</p>

The logarithmic mapping equation (new):
<p align="center">
${LookupIndices}_{(0−255)}=log10\left(1+9\left(\frac{ImageData −WindowMIN}{WINDOW}\right)\right)∗255.0$
</p>

The output of both equations is normalized UCHAR values ranging from 0-255. These values serve as indices to choose a color from the ColorNode.

To facilitate the introduction of this new equation, this PR adds a "vtkImageMapToWindowLevelAddon" which subclasses the existing "vtkImageMapToWindowLevelColors". Alternatively, we could update "vtkImageMapToWindowLevelColors" directly.

## Downstream effects
Because the Color Node is untouched by these changes, this PR also updates some downstream logic to make sure that the Colorbar is displayed in a logarithmic state. I.e.:
1. This PR updates how the color legend is shown
2. This PR updates how the color gradient is shown in the Volumes widget transfer function (the one with a histogram)
3. This PR updates how points are placed when volume display properties are transferred from the Volumes module to the Volume Rendering module. (Triggered by the "Synchronize with Volumes module" checkable button.)

## Related enhancement
The other enhancement in this PR is to reduce the number of control points that get placed in the "Scalar Color Mapping" Volume Rendering widget when synchronizing the Volume module and the Volume Rendering module. Manual editing in that interface becomes extremely tedious with many points placed close together. This is especially apparent when points are placed logarithmically.

The approach taken in this PR is to first initialize a full linear transfer function by placing endpoints.
Then, before we add the next new points, we check to see how different that new color will be from what is already shown in the transfer function. For linear maps like "Grey", we don't need any more control points. For maps like "Rainbow" we will need more points.
| fcd8017 | This PR | This PR (logarithmic applied) |
|--------|--------|--------|
|image|image|image| 
|image|image|image|

Currently, this PR just sets a hard-coded threshold for how much deviation is allowed before we should add a new point. That threshold can certainly be tuned.

### DIFF
--- a/Modules/Loadable/Colors/MRMLDM/vtkMRMLColorLegendDisplayableManager.cxx
+++ b/Modules/Loadable/Colors/MRMLDM/vtkMRMLColorLegendDisplayableManager.cxx
@@ -481,21 +481,33 @@ bool vtkMRMLColorLegendDisplayableManager::vtkInternal::UpdateActor(vtkMRMLColor
     return false;
   }
 
-  // Handle invert flag for scalar volume display nodes
-  if (scalarVolumeDisplayNode && scalarVolumeDisplayNode->GetInvertDisplayScalarRange())
+  // Handle invert flag and different scalar mapping methods
+
+  if (scalarVolumeDisplayNode)
   {
-    // Invert the lookup table colors
-    vtkNew<vtkLookupTable> invertedLut;
-    invertedLut->DeepCopy(lut);
-    // Reverse the order of colors in the table
-    int numColors = invertedLut->GetNumberOfTableValues();
+    bool invertMap = scalarVolumeDisplayNode->GetInvertDisplayScalarRange();
+    auto mappingMode = scalarVolumeDisplayNode->GetWindowMappingMethodFromIndex(scalarVolumeDisplayNode->GetWindowMappingMethod());
+
+    vtkNew<vtkLookupTable> resampledLut;
+    resampledLut->DeepCopy(lut);
+
+    int numColors = resampledLut->GetNumberOfTableValues();
+    double lutScale = (double)(numColors - 1);
+    int lookupIndex = 0;
+    double normalizedWindowLocation = 0.0;
+    double rgba[4] = { 0.0, 0.0, 0.0, 0.0 };
     for (int i = 0; i < numColors; i++)
     {
-      double rgba[4] = { 0.0, 0.0, 0.0, 0.0 };
-      lut->GetTableValue(i, rgba);
-      invertedLut->SetTableValue(numColors - 1 - i, rgba);
+      normalizedWindowLocation = vtkImageMapToWindowLevelAddon::mapScalarToWindow(double(i), 0.0, lutScale, mappingMode);
+      if (invertMap)
+      {
+        normalizedWindowLocation = 1.0 - normalizedWindowLocation;
+      }
+      lookupIndex = round(lutScale * normalizedWindowLocation);
+      lut->GetTableValue(lookupIndex, rgba);
+      resampledLut->SetTableValue(i, rgba);
     }
-    lut = invertedLut;
+    lut = resampledLut;
   }
 
   lut->SetTableRange(range);
--- a/Modules/Loadable/VolumeRendering/Logic/vtkSlicerVolumeRenderingLogic.cxx
+++ b/Modules/Loadable/VolumeRendering/Logic/vtkSlicerVolumeRenderingLogic.cxx
@@ -469,17 +469,18 @@ void vtkSlicerVolumeRenderingLogic::SetThresholdToVolumeProp(double scalarRange[
 }
 
 //----------------------------------------------------------------------------
-void vtkSlicerVolumeRenderingLogic::SetWindowLevelToVolumeProp(double scalarRange[2], double windowLevel[2], vtkScalarsToColors* colors, vtkVolumeProperty* volumeProp)
+void vtkSlicerVolumeRenderingLogic::SetWindowLevelToVolumeProp(double scalarRange[2],
+                                                               double windowLevel[2],
+                                                               vtkScalarsToColors* colors,
+                                                               vtkVolumeProperty* volumeProp,
+                                                               vtkImageMapToWindowLevelAddon::WindowMappingMode mappingMode,
+                                                               bool invertColormap)
 {
   if (!volumeProp || !scalarRange || !windowLevel)
   {
     vtkWarningMacro("SetWindowLevelToVolumeProp: Inputs do not exist.");
     return;
   }
-
-  // Note that window can be negative, which means that the color lookup table should be reversed
-  // (low voxel values are assigned to high values in the color lookup table).
-  bool invertColorTable = (windowLevel[0] < 0.0);
 
   vtkNew<vtkColorTransferFunction> colorTransfer;
 
@@ -496,7 +497,7 @@ void vtkSlicerVolumeRenderingLogic::SetWindowLevelToVolumeProp(double scalarRang
     double color_X_RGB_MS[6] = { 0., 0., 0., 0., 0.5, 1.0 }; // x, RGB, midpoint, sharpness
     for (vtkIdType i = 0; i < colorCount; ++i)
     {
-      vtkIdType colorIndex = (invertColorTable ? colorCount - 1 - i : i);
+      vtkIdType colorIndex = (invertColormap ? colorCount - 1 - i : i);
       inputColorTransfer->GetNodeValue(colorIndex, color_X_RGB_MS);
       colorTransfer->AddRGBPoint(offset + color_X_RGB_MS[0] * scale, color_X_RGB_MS[1], color_X_RGB_MS[2], color_X_RGB_MS[3], color_X_RGB_MS[4], color_X_RGB_MS[5]);
     }
@@ -534,7 +535,7 @@ void vtkSlicerVolumeRenderingLogic::SetWindowLevelToVolumeProp(double scalarRang
     else // if (numberOfColors > 1)
     {
       double color[4] = { 0.0, 0.0, 0.0, 1.0 };
-      vtkIdType firstColorTableIndex = (invertColorTable ? numberOfColors - 1 : 0);
+      vtkIdType firstColorTableIndex = (invertColormap ? numberOfColors - 1 : 0);
       lut->GetTableValue(firstColorTableIndex, color);
       colorTransfer->AddRGBPoint(vtkMRMLVolumePropertyNode::HigherAndUnique(scalarRange[0], previous), color[0], color[1], color[2]);
 
@@ -544,20 +545,26 @@ void vtkSlicerVolumeRenderingLogic::SetWindowLevelToVolumeProp(double scalarRang
       const vtkIdType maxNumberOfPoints = 24;
 
       const vtkIdType numberOfPoints = std::min(numberOfColors, maxNumberOfPoints);
-      // convert from point index to color index
-      double pointIndexScale = static_cast<double>(numberOfColors - 1) / (numberOfPoints - 1);
-      double offset = outputRange[0];
-      double scale = fabs(windowLevel[0]) / (numberOfColors - 1);
+      double pointMaxIndex = static_cast<double>(numberOfPoints - 1);
+      double colorMaxIndex = static_cast<double>(numberOfColors - 1);
+
+      const double window = fabs(windowLevel[0]);
+      const double xOffset = outputRange[0];
+      double xStepSize = window / numberOfPoints;
+
       for (vtkIdType pointIndex = 0; pointIndex < numberOfPoints; ++pointIndex)
       {
-        vtkIdType colorIndex = pointIndex * pointIndexScale;
-        const double value = offset + colorIndex * scale;
-        vtkIdType colorTableIndex = (invertColorTable ? numberOfColors - 1 - colorIndex : colorIndex);
-        lut->GetTableValue(colorTableIndex, color);
-        colorTransfer->AddRGBPoint(vtkMRMLVolumePropertyNode::HigherAndUnique(value, previous), color[0], color[1], color[2]);
+        double xValue = xOffset + pointIndex * xStepSize;
+        double normalizedColorIndex = vtkImageMapToWindowLevelAddon::mapScalarToWindow(double(pointIndex), 0.0, pointMaxIndex, mappingMode);
+        if (invertColormap)
+        {
+          normalizedColorIndex = 1.0 - normalizedColorIndex;
+        }
+        vtkIdType colorIndex = normalizedColorIndex * colorMaxIndex;
+        lut->GetTableValue(colorIndex, color);
+        colorTransfer->AddRGBPoint(vtkMRMLVolumePropertyNode::HigherAndUnique(xValue, previous), color[0], color[1], color[2]);
       }
-
-      vtkIdType lastColorTableIndex = (invertColorTable ? 0 : numberOfColors - 1);
+      vtkIdType lastColorTableIndex = (invertColormap ? 0 : numberOfColors - 1);
       lut->GetTableValue(lastColorTableIndex, color);
       colorTransfer->AddRGBPoint(vtkMRMLVolumePropertyNode::HigherAndUnique(outputRange[1], previous), color[0], color[1], color[2]);
       colorTransfer->AddRGBPoint(vtkMRMLVolumePropertyNode::HigherAndUnique(scalarRange[1], previous), color[0], color[1], color[2]);
@@ -734,17 +741,15 @@ void vtkSlicerVolumeRenderingLogic::CopyScalarDisplayToVolumeRenderingDisplayNod
 
   vtkScalarsToColors* lut = vpNode->GetColorNode() ? vpNode->GetColorNode()->GetScalarsToColors() : nullptr;
   vtkVolumeProperty* prop = vspNode->GetVolumePropertyNode()->GetVolumeProperty();
+  vtkImageMapToWindowLevelAddon::WindowMappingMode scalarMappingMode = vpNode->GetWindowMappingMethodFromIndex(vpNode->GetWindowMappingMethod());
+  bool invertedColormap = vpNode->GetInvertDisplayScalarRange();
 
   int disabledModify = vspNode->StartModify();
   int vpNodeDisabledModify = vspNode->GetVolumePropertyNode()->StartModify();
 
   this->SetThresholdToVolumeProp(scalarRange, threshold, prop, this->UseLinearRamp, ignoreVolumeDisplayNodeThreshold);
 
-  if (vpNode->GetInvertDisplayScalarRange())
-  {
-    windowLevel[0] *= -1;
-  }
-  this->SetWindowLevelToVolumeProp(scalarRange, windowLevel, lut, prop);
+  this->SetWindowLevelToVolumeProp(scalarRange, windowLevel, lut, prop, scalarMappingMode, invertedColormap);
   this->SetGradientOpacityToVolumeProp(scalarRange, prop);
 
   vspNode->GetVolumePropertyNode()->EndModify(vpNodeDisabledModify);
--- a/Modules/Loadable/VolumeRendering/Logic/vtkSlicerVolumeRenderingLogic.cxx
+++ b/Modules/Loadable/VolumeRendering/Logic/vtkSlicerVolumeRenderingLogic.cxx
@@ -535,15 +535,23 @@ void vtkSlicerVolumeRenderingLogic::SetWindowLevelToVolumeProp(double scalarRang
     else // if (numberOfColors > 1)
     {
       double color[4] = { 0.0, 0.0, 0.0, 1.0 };
+      double predictedColor[3] = { 0.0, 0.0, 0.0 };
+      const double colorDeviationThreshold = 0.25;
+
+      // We start with all relevant endpoints.
       vtkIdType firstColorTableIndex = (invertColormap ? numberOfColors - 1 : 0);
       lut->GetTableValue(firstColorTableIndex, color);
       colorTransfer->AddRGBPoint(vtkMRMLVolumePropertyNode::HigherAndUnique(scalarRange[0], previous), color[0], color[1], color[2]);
+      colorTransfer->AddRGBPoint(vtkMRMLVolumePropertyNode::HigherAndUnique(outputRange[0], previous), color[0], color[1], color[2]);
+      vtkIdType lastColorTableIndex = (invertColormap ? 0 : numberOfColors - 1);
+      lut->GetTableValue(lastColorTableIndex, color);
+      colorTransfer->AddRGBPoint(vtkMRMLVolumePropertyNode::HigherAndUnique(outputRange[1], previous), color[0], color[1], color[2]);
+      colorTransfer->AddRGBPoint(vtkMRMLVolumePropertyNode::HigherAndUnique(scalarRange[1], previous), color[0], color[1], color[2]);
 
       // We place up to maxNumberOfPoints points in the color transfer function.
       // The number is high enough to accurately describe most color tables,
       // but not too high so that the user can still edit the function manually.
-      const vtkIdType maxNumberOfPoints = 24;
-
+      const vtkIdType maxNumberOfPoints = 50;
       const vtkIdType numberOfPoints = std::min(numberOfColors, maxNumberOfPoints);
       double pointMaxIndex = static_cast<double>(numberOfPoints - 1);
       double colorMaxIndex = static_cast<double>(numberOfColors - 1);
@@ -560,14 +568,17 @@ void vtkSlicerVolumeRenderingLogic::SetWindowLevelToVolumeProp(double scalarRang
         {
           normalizedColorIndex = 1.0 - normalizedColorIndex;
         }
-        vtkIdType colorIndex = normalizedColorIndex * colorMaxIndex;
+        vtkIdType colorIndex = std::lround(normalizedColorIndex * colorMaxIndex);
         lut->GetTableValue(colorIndex, color);
-        colorTransfer->AddRGBPoint(vtkMRMLVolumePropertyNode::HigherAndUnique(xValue, previous), color[0], color[1], color[2]);
+        colorTransfer->GetColor(xValue, predictedColor);
+        const double d0 = std::fabs(color[0] - predictedColor[0]);
+        const double d1 = std::fabs(color[1] - predictedColor[1]);
+        const double d2 = std::fabs(color[2] - predictedColor[2]);
+        if ((d0 + d1 + d2) > colorDeviationThreshold)
+        {
+          colorTransfer->AddRGBPoint(vtkMRMLVolumePropertyNode::HigherAndUnique(xValue, previous), color[0], color[1], color[2]);
+        }
       }
-      vtkIdType lastColorTableIndex = (invertColormap ? 0 : numberOfColors - 1);
-      lut->GetTableValue(lastColorTableIndex, color);
-      colorTransfer->AddRGBPoint(vtkMRMLVolumePropertyNode::HigherAndUnique(outputRange[1], previous), color[0], color[1], color[2]);
-      colorTransfer->AddRGBPoint(vtkMRMLVolumePropertyNode::HigherAndUnique(scalarRange[1], previous), color[0], color[1], color[2]);
     }
   }
 
--- a/Modules/Loadable/Volumes/Widgets/qSlicerScalarVolumeDisplayWidget.cxx
+++ b/Modules/Loadable/Volumes/Widgets/qSlicerScalarVolumeDisplayWidget.cxx
@@ -16,6 +16,7 @@
 #include "vtkMRMLScalarVolumeDisplayNode.h"
 #include "vtkMRMLScalarVolumeNode.h"
 #include "vtkMRMLScene.h"
+#include "vtkImageMapToWindowLevelAddon.h"
 
 // VTK includes
 #include <vtkAlgorithm.h>
@@ -89,6 +90,17 @@ void qSlicerScalarVolumeDisplayWidgetPrivate::init()
   barsItem->setBarWidth(1.);
   scene->addItem(barsItem);
 
+  // Set up options for linear or logarithmic color mapping
+  this->ScalarMappingComboBox->insertItem( //
+    vtkImageMapToWindowLevelAddon::Linear,
+    qSlicerScalarVolumeDisplayWidget::tr("Linear"));
+  this->ScalarMappingComboBox->insertItem( //
+    vtkImageMapToWindowLevelAddon::Logarithmic,
+    qSlicerScalarVolumeDisplayWidget::tr("Logarithmic"));
+  this->ScalarMappingComboBox->insertItem( //
+    vtkImageMapToWindowLevelAddon::InverseLogarithmic,
+    qSlicerScalarVolumeDisplayWidget::tr("Inverse logarithmic"));
+
   // Add mapping from presets defined in the Volumes module logic (VolumeDisplayPresets.json)
 
   // read volume preset names from volumes logic
@@ -128,6 +140,7 @@ void qSlicerScalarVolumeDisplayWidgetPrivate::init()
 
   QObject::connect(this->InterpolateCheckbox, SIGNAL(toggled(bool)), q, SLOT(setInterpolate(bool)));
   QObject::connect(this->InvertDisplayScalarRangeCheckbox, SIGNAL(toggled(bool)), q, SLOT(setInvert(bool)));
+  QObject::connect(this->ScalarMappingComboBox, SIGNAL(currentIndexChanged(int)), q, SLOT(setScalarMappingMethod(int)));
   QObject::connect(this->ColorTableComboBox, SIGNAL(currentNodeChanged(vtkMRMLNode*)), q, SLOT(setColorNode(vtkMRMLNode*)));
   QObject::connect(this->HistogramGroupBox, SIGNAL(toggled(bool)), q, SLOT(onHistogramSectionExpanded(bool)));
 }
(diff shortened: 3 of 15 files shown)


ajallphin:
Relates to conversation had between me, @jamesobutler, @ebrahimebrahim, and @sjh26 at the Slicer hangout.

Much of the code overlaps with recent "Invert" colormap changes from @lassoan

ajallphin:
Build likely failed due to the updated CMakeLists.txt. My local build ran successfully but obviously required a rerun of CMake.

Modules/Loadable/VolumeRendering/Logic/vtkSlicerVolumeRenderingLogic.cxx
```diff
@@ -744,7 +769,7 @@ void vtkSlicerVolumeRenderingLogic::CopyScalarDisplayToVolumeRenderingDisplayNod
   {
     windowLevel[0] *= -1;
   }
-  this->SetWindowLevelToVolumeProp(scalarRange, windowLevel, lut, prop);
+  this->SetWindowLevelToVolumeProp(scalarRange, windowLevel, vpNode->GetLogCompressWindow(), lut, prop);
```
lassoan: SetWindowLevelToVolumeProp is public API, it cannot be changed in a backward-incompatible way. You can add new optional arguments to the end (or add a new method with a different name and depracate the old method).

Modules/Loadable/Volumes/Widgets/qSlicerScalarVolumeDisplayWidget.cxx
```diff
@@ -381,10 +384,21 @@ void qSlicerScalarVolumeDisplayWidget::updateHistogram()
...
         int resolution = 50;
         double rgb[3] = { 0.0, 0.0, 0.0 };
+        bool logScaled = displayNode->GetLogCompressWindow();
+        double colorMapPosition;
+
         for (int i = 0; i < resolution; i++)
         {
+          if (logScaled)
+          {
+            colorMapPosition = std::log10(1.0 + 9.0 * (double(i) / (resolution - 1)));
```
lassoan: It is not good that we have this formula in 5 different places, at various levels (logic, GUI, ...). Ideally, it should be implemented at one place and that would be accessed from all the other places.

Modules/Loadable/VolumeRendering/Logic/vtkSlicerVolumeRenderingLogic.cxx
```diff
@@ -469,9 +469,9 @@ void vtkSlicerVolumeRenderingLogic::SetThresholdToVolumeProp(double scalarRange[
 }
 
 //----------------------------------------------------------------------------
-void vtkSlicerVolumeRenderingLogic::SetWindowLevelToVolumeProp(double scalarRange[2], double windowLevel[2], vtkScalarsToColors* colors, vtkVolumeProperty* volumeProp)
+void vtkSlicerVolumeRenderingLogic::SetWindowLevelToVolumeProp(double scalarRange[2], double windowLevel[2], bool logAxis, vtkScalarsToColors* colors, vtkVolumeProperty* volProp)
```
lassoan: let's keep `volumeProp` to reduce noise (this word is used in several other methods, too)

Modules/Loadable/Colors/MRMLDM/vtkMRMLColorLegendDisplayableManager.cxx
```diff
@@ -498,6 +498,25 @@ bool vtkMRMLColorLegendDisplayableManager::vtkInternal::UpdateActor(vtkMRMLColor
...
+    vtkNew<vtkLookupTable> compressedLut;
+    compressedLut->DeepCopy(lut);
+    int numColors = compressedLut->GetNumberOfTableValues();
+    double magnitude = (double)(numColors - 1);
+
+    for (int i = 0; i < numColors; i++)
+    {
+      double rgba[4] = { 0.0, 0.0, 0.0, 0.0 };
+      double lookupLocale = magnitude * std::log10(1.0 + 9.0 * (double(i) / magnitude));
+      lut->GetTableValue(lookupLocale, rgba);
```
lassoan: This is effectively `lut->GetTableValue(floor(lookupLocale), rgba);`. Is this OK?
For example, why not `lut->GetTableValue(round(lookupLocale), rgba);` ?
The may be more noticeable if the LUT does not have many points.

Modules/Loadable/Colors/MRMLDM/vtkMRMLColorLegendDisplayableManager.cxx
```diff
@@ -498,6 +498,25 @@ bool vtkMRMLColorLegendDisplayableManager::vtkInternal::UpdateActor(vtkMRMLColor
     lut = invertedLut;
   }
 
+  // Handle window log compression for scalar volume display nodes
+  if (scalarVolumeDisplayNode && scalarVolumeDisplayNode->GetLogCompressWindow())
```
lassoan: Since we are already add vtkImageMapToWindowLevelAddon to handle lin/log, could we also add handling of invert flag to vtkImageMapToWindowLevelAddon? It would be a bit more consistent and probably the overall the code would be simpler.

lassoan (commented):
I've added a couple of comments, but I have a high-level question here. vtkLookupTable can already do logarithmic mapping (via [vtkLookupTable::SetScaleToLog10](https://vtk.org/doc/nightly/html/classvtkLookupTable.html#a93a46edc6ca10b0ba0e7cfff4acb52e9)) could we use that instead of introducing a new filter (vtkImageMapToWindowLevelAddon)?
It would be better to not deviate from how VTK implements log mapping of colors (less Slicer-specific code to maintain, less complications in the long term).

ajallphin:
Thanks for the comments. I will make some updates. 

> I've added a couple of comments, but I have a high-level question here. vtkLookupTable can already do logarithmic mapping (via [vtkLookupTable::SetScaleToLog10](https://vtk.org/doc/nightly/html/classvtkLookupTable.html#a93a46edc6ca10b0ba0e7cfff4acb52e9)) could we use that instead of introducing a new filter (vtkImageMapToWindowLevelAddon)? It would be better to not deviate from how VTK implements log mapping of colors (less Slicer-specific code to maintain, less complications in the long term).

To address the high-level question:
In our custom app, we had originally used the vtkLookupTable::SetScaleToLog10 approach. However, I realized that there was a fundamental flaw with using that approach with the current Slicer visualization pipeline. (see https://slicer.readthedocs.io/en/latest/developer_guide/mrml_overview.html#slice-view-pipeline) In Slicer, the color mapping process has been split into two parts: Scalar rebinning (vtkImageMapToWindowLevelColors) and THEN color assignment (vtkImageMapToColors). In reality, vtk lookup tables can do both simultaneously, but that is not how we are using them. Setting the lookup table itself to Log10 doesn't do what we want since the scalar rebinning was done before going to the lookup table:
| Current slicer pipeline | Adding in SetScaleToLog10 |
|--------|--------|
|image|image|

Changing the lookup table scaling itself would work if and only if the input 0-255 values were themselves identical to the image data

ajallphin:
In theory you could bypass the WindowLevel rebinning entirely by setting the LookupTable "TableRange" to match the current window. However, that leads to other issues. For example, SetScaleToLog10 doesn't work for tables with a negative starting range.

Thus, the best solution I could come up with was the implementation in this PR.

Modules/Loadable/VolumeRendering/Logic/vtkSlicerVolumeRenderingLogic.cxx
ajallphin: Oops thanks for catching that.

Modules/Loadable/Volumes/Widgets/qSlicerScalarVolumeDisplayWidget.cxx
ajallphin: I'm struggling to find a good way to centralize these log10 formulas. Any implementation advice?

Modules/Loadable/Colors/MRMLDM/vtkMRMLColorLegendDisplayableManager.cxx
ajallphin: Probably a good idea to round instead of floor.

Modules/Loadable/Colors/MRMLDM/vtkMRMLColorLegendDisplayableManager.cxx
ajallphin: It is definitely possible. I'm leaning towards keeping them separate because of the slight directional ambiguity if you handle both simultaneously.
image

I like the idea of always keeping the log10 calculations the same.

lassoan: Such details are exactly why I think it would be useful to handle all the remapping in one place.

Modules/Loadable/Volumes/Widgets/qSlicerScalarVolumeDisplayWidget.cxx
lassoan: It must be should be at the lowest level component, which is Libs/MRML/Core (so that it can be accessed from everywhere we need). Since a VTK filter should not depend on a MRML node but a MRML node can depend on a VTK filter, to me it seems that the only option is to add it to vtkImageMapToWindowLevelAddon. It could be a static inline method.

Modules/Loadable/Colors/MRMLDM/vtkMRMLColorLegendDisplayableManager.cxx
ajallphin: For me it boils down to the answer to this question:
Do we allow inversion of both the colormap and the logarithmic compression? If so, then we should remap in one place. If not, it may be simpler to leave the "invert" logic untouched. I could go either way. You have way more experience with long term code maintenance, so I trust your discernment.

| | Normal Colormap | Invert Colormap |
|--------|--------|--------|
| Log |image |image|
| Inverted Log | image|image|

lassoan: Yes, inversion and log mapping are two independent flags, we should allow the user to enable both.

ajallphin: But should they be able to invert the log mapping independently of the colormap?

Colormap options: Normal OR Inverted
Scalar mapping: Linear OR left compressed OR right compressed

lassoan: Colormap inversion flag is separate from log mapping parameters.

Log mapping parameters that could make sense to me would be `float LogBase` and `bool LogCompressLowValues`. If your application does not require these to be configurable now then we don't need to implement these right now - we just need to make sure to choose a design that will not make it very complicated to introduce these later.

I've just realized that instead of having a flag for choosing between linear/log mapping, it would be better to have a `MappingMode` enum with options such as `linear` and `log` (or `log` and `inverseLog` if we want to avoid the `LogCompressLowValues` property). This would allow adding other kind of mappings (e.g., gamma mapping) in the future without worrying about making backward-incompatible changes to the API.

ebrahimebrahim: It's a bit weird that they're represented as independent _flags_. It's more like an order _list_ of independent operations, right? 5 possibilities:

```
no-op

log

log
invert

invert

invert
log
```

2 checkboxes feels like an unnatural way to represent 5 possibilities. Maybe the internal design ought to follow this logic of an ordered list of operations, so that in the future if the UI is also made to follow that logic it becomes simple to hook up. (And for now the UI staying as two checkboxes could be fine-- there is just a canonical order we hard-code for now)

lassoan: There are actually 6 combinations:
- linear
- log
- reverse log
- linear with invert colormap
- log with invert colormap
- reverse log with invert colormap

These properties (that are directly mapped to GUI widgets) look reasonable to me:
- invert colormap [bool, checkbox] - already exists
- mapping mode [enum, combobox]: `Linear` or `Log` - maybe also a `ReverseLog` option
- log mapping options (only enabled if mapping mode is log kind):
  - logarithm base [double, spinbox]
  - compress high values [bool, checkbox] - not needed if a "ReverseLog" mapping mode is added

ajallphin:
For reference this is the diff between the vtkImageMapToWindowLevelColors and the new addon.

<details>
<summary>diff --git vtkImageMapToWindowLevelColors.cxx vtkImageMapToWindowLevelAddon.cxx</summary>

```
index 2d79d1e..fb79ffb 100644
--- a/vtkImageMapToWindowLevelColors.cxx
+++ b/vtkImageMapToWindowLevelAddon.cxx
@@ -1,6 +1,4 @@
-// SPDX-FileCopyrightText: Copyright (c) Ken Martin, Will Schroeder, Bill Lorensen
-// SPDX-License-Identifier: BSD-3-Clause
-#include "vtkImageMapToWindowLevelColors.h"
+#include "vtkImageMapToWindowLevelAddon.h"
 
 #include "vtkDataArray.h"
 #include "vtkImageData.h"
@@ -12,22 +10,19 @@
 
 #include <cmath>
 
-VTK_ABI_NAMESPACE_BEGIN
-vtkStandardNewMacro(vtkImageMapToWindowLevelColors);
+vtkStandardNewMacro(vtkImageMapToWindowLevelAddon);
 
 // Constructor sets default values
-vtkImageMapToWindowLevelColors::vtkImageMapToWindowLevelColors()
+vtkImageMapToWindowLevelAddon::vtkImageMapToWindowLevelAddon()
 {
-  this->Window = 255;
-  this->Level = 127.5;
+  this->MappingMode = WindowMappingMode::Linear;
 }
 
-vtkImageMapToWindowLevelColors::~vtkImageMapToWindowLevelColors() = default;
+vtkImageMapToWindowLevelAddon::~vtkImageMapToWindowLevelAddon() = default;
 
 // This method checks to see if we can simply reference the input data
-int vtkImageMapToWindowLevelColors::RequestData(
-  vtkInformation* request, vtkInformationVector** inputVector, vtkInformationVector* outputVector)
+int vtkImageMapToWindowLevelAddon::RequestData(vtkInformation* request, vtkInformationVector** inputVector, vtkInformationVector* outputVector)
 {
   vtkInformation* outInfo = outputVector->GetInformationObject(0);
   vtkInformation* inInfo = inputVector[0]->GetInformationObject(0);
@@ -37,8 +32,8 @@ int vtkImageMapToWindowLevelColors::RequestData(
 
   // If LookupTable is null and window / level produces no change,
   // then just pass the data
-  if (this->LookupTable == nullptr &&
-    (inData->GetScalarType() == VTK_UNSIGNED_CHAR && this->Window == 255 && this->Level == 127.5))
+  if (this->LookupTable == nullptr && this->MappingMode==WindowMappingMode::Linear &&
+     (inData->GetScalarType() == VTK_UNSIGNED_CHAR && this->Window == 255 && this->Level == 127.5))
   {
     vtkDebugMacro("ExecuteData: LookupTable not set, "
                   "Window / Level at default, "
@@ -68,14 +63,12 @@ int vtkImageMapToWindowLevelColors::RequestData(
 }
 
-int vtkImageMapToWindowLevelColors::RequestInformation(vtkInformation* vtkNotUsed(request),
-  vtkInformationVector** inputVector, vtkInformationVector* outputVector)
+int vtkImageMapToWindowLevelAddon::RequestInformation(vtkInformation* vtkNotUsed(request), vtkInformationVector** inputVector, vtkInformationVector* outputVector)
 {
   vtkInformation* inInfo = inputVector[0]->GetInformationObject(0);
   vtkInformation* outInfo = outputVector->GetInformationObject(0);
 
-  vtkInformation* inScalarInfo = vtkDataObject::GetActiveFieldInformation(
-    inInfo, vtkDataObject::FIELD_ASSOCIATION_POINTS, vtkDataSetAttributes::SCALARS);
+  vtkInformation* inScalarInfo = vtkDataObject::GetActiveFieldInformation(inInfo, vtkDataObject::FIELD_ASSOCIATION_POINTS, vtkDataSetAttributes::SCALARS);
   if (!inScalarInfo)
   {
     vtkErrorMacro("Missing scalar field on input information!");
@@ -84,20 +77,16 @@ int vtkImageMapToWindowLevelColors::RequestInformation(vtkInformation* vtkNotUse
 
   // If LookupTable is null and window / level produces no change,
   // then the data will be passed
-  if (this->LookupTable == nullptr &&
-    (inScalarInfo->Get(vtkDataObject::FIELD_ARRAY_TYPE()) == VTK_UNSIGNED_CHAR &&
-      this->Window == 255 && this->Level == 127.5))
+  if (this->LookupTable == nullptr && (inScalarInfo->Get(vtkDataObject::FIELD_ARRAY_TYPE()) == VTK_UNSIGNED_CHAR && this->Window == 255 && this->Level == 127.5))
   {
     if (inScalarInfo->Get(vtkDataObject::FIELD_ARRAY_TYPE()) != VTK_UNSIGNED_CHAR)
     {
-      vtkErrorMacro(
-        "ExecuteInformation: No LookupTable was set and input data is not VTK_UNSIGNED_CHAR!");
+      vtkErrorMacro("ExecuteInformation: No LookupTable was set and input data is not VTK_UNSIGNED_CHAR!");
     }
     else
     {
       // no lookup table, pass the input if it was VTK_UNSIGNED_CHAR
-      vtkDataObject::SetPointDataActiveScalarInfo(
-        outInfo, VTK_UNSIGNED_CHAR, inScalarInfo->Get(vtkDataObject::FIELD_NUMBER_OF_COMPONENTS()));
+      vtkDataObject::SetPointDataActiveScalarInfo(outInfo, VTK_UNSIGNED_CHAR, inScalarInfo->Get(vtkDataObject::FIELD_NUMBER_OF_COMPONENTS()));
     }
   }
   else // the lookup table was set or window / level produces a change
@@ -105,21 +94,11 @@ int vtkImageMapToWindowLevelColors::RequestInformation(vtkInformation* vtkNotUse
     int numComponents = 4;
     switch (this->OutputFormat)
     {
-      case VTK_RGBA:
-        numComponents = 4;
-        break;
-      case VTK_RGB:
-        numComponents = 3;
-        break;
-      case VTK_LUMINANCE_ALPHA:
-        numComponents = 2;
-        break;
-      case VTK_LUMINANCE:
-        numComponents = 1;
-        break;
-      default:
-        vtkErrorMacro("ExecuteInformation: Unrecognized color format.");
-        break;
+      case VTK_RGBA: numComponents = 4; break;
+      case VTK_RGB: numComponents = 3; break;
+      case VTK_LUMINANCE_ALPHA: numComponents = 2; break;
+      case VTK_LUMINANCE: numComponents = 1; break;
+      default: vtkErrorMacro("ExecuteInformation: Unrecognized color format."); break;
     }
     vtkDataObject::SetPointDataActiveScalarInfo(outInfo, VTK_UNSIGNED_CHAR, numComponents);
   }
@@ -127,113 +106,10 @@ int vtkImageMapToWindowLevelColors::RequestInformation(vtkInformation* vtkNotUse
   return 1;
 }
 
-/*
- * This templated routine calculates effective lower and upper limits
- * for a window of values of type T, lower and upper.
- */
-template <class T>
-void vtkImageMapToWindowLevelClamps(vtkImageData* data, double w, double l, T& lower, T& upper,
-  unsigned char& lower_val, unsigned char& upper_val)
-{
-  double f_lower, f_upper, f_lower_val, f_upper_val;
-  double adjustedLower, adjustedUpper;
-  double range[2];
-
-  data->GetPointData()->GetScalars()->GetDataTypeRange(range);
-
-  f_lower = l - fabs(w) / 2.0;
-  f_upper = f_lower + fabs(w);
-
-  // Set the correct lower value
-  if (f_lower <= range[1])
-  {
-    if (f_lower >= range[0])
-    {
-      lower = static_cast<T>(f_lower);
-      adjustedLower = f_lower;
-    }
-    else
-    {
-      lower = static_cast<T>(range[0]);
-      adjustedLower = range[0];
-    }
-  }
-  else
-  {
-    lower = static_cast<T>(range[1]);
-    adjustedLower = range[1];
-  }
-
-  // Set the correct upper value
-  if (f_upper >= range[0])
-  {
-    if (f_upper <= range[1])
-    {
-      upper = static_cast<T>(f_upper);
-      adjustedUpper = f_upper;
-    }
-    else
-    {
-      upper = static_cast<T>(range[1]);
-      adjustedUpper = range[1];
-    }
-  }
-  else
-  {
-    upper = static_cast<T>(range[0]);
-    adjustedUpper = range[0];
-  }
-
-  // now compute the lower and upper values
-  if (w > 0.0)
-  {
-    f_lower_val = 255.0 * (adjustedLower - f_lower) / w;
-    f_upper_val = 255.0 * (adjustedUpper - f_lower) / w;
-  }
-  else if (w < 0.0)
-  {
-    f_lower_val = 255.0 + 255.0 * (adjustedLower - f_lower) / w;
-    f_upper_val = 255.0 + 255.0 * (adjustedUpper - f_lower) / w;
-  }
-  else
-  {
-    f_lower_val = 0.0;
-    f_upper_val = 255.0;
-  }
-
-  if (f_upper_val > 255)
-  {
-    upper_val = 255;
-  }
-  else if (f_upper_val < 0)
-  {
-    upper_val = 0;
-  }
-  else
-  {
-    upper_val = static_cast<unsigned char>(f_upper_val);
-  }
-
-  if (f_lower_val > 255)
-  {
-    lower_val = 255;
-  }
-  else if (f_lower_val < 0)
-  {
-    lower_val = 0;
-  }
-  else
-  {
-    lower_val = static_cast<unsigned char>(f_lower_val);
-  }
-}
-
 // This non-templated function executes the filter for any type of data.
 template <class T>
-void vtkImageMapToWindowLevelColorsExecute(vtkImageMapToWindowLevelColors* self,
-  vtkImageData* inData, T* inPtr, vtkImageData* outData, unsigned char* outPtr, int outExt[6],
-  int id)
+void vtkImageMapToWindowLevelAddonExecute(vtkImageMapToWindowLevelAddon* self, vtkImageData* inData, T* inPtr, vtkImageData* outData, unsigned char* outPtr, int outExt[6], int id)
 {
   int idxX, idxY, idxZ;
   int extX, extY, extZ;
@@ -249,14 +125,15 @@ void vtkImageMapToWindowLevelColorsExecute(vtkImageMapToWindowLevelColors* self,
   T* inPtr1;
   unsigned char* optr;
   T* iptr;
-  double shift = self->GetWindow() / 2.0 - self->GetLevel();
-  double scale = 255.0 / self->GetWindow();
 
-  T lower, upper;
-  unsigned char lower_val, upper_val, result_val;
-  unsigned short ushort_val;
-  vtkImageMapToWindowLevelClamps(
-    inData, self->GetWindow(), self->GetLevel(), lower, upper, lower_val, upper_val);
+  double window = self->GetWindow();
+  double level = self->GetLevel();
+  T lower = static_cast<T>(level - fabs(window) / 2.0);
+  T upper = lower + static_cast<T>(fabs(window));
+
+  unsigned char result_val;
+  auto mappingMode = self->GetMappingMode();
+  double windowFactor;
 
   // find the region to loop over
   extX = outExt[1] - outExt[0] + 1;
@@ -297,38 +174,23 @@ void vtkImageMapToWindowLevelColorsExecute(vtkImageMapToWindowLevelColors* self,
 
       if (lookupTable)
       {
-        lookupTable->MapScalarsThroughTable(
-          inPtr1, outPtr1, dataType, extX, numberOfComponents, outputFormat);
-
+        lookupTable->MapScalarsThroughTable2(inPtr1, outPtr1, dataType, extX, numberOfComponents, outputFormat);
         for (idxX = 0; idxX < extX; idxX++)
         {
-          if (*iptr <= lower)
-          {
-            ushort_val = lower_val;
-          }
-          else if (*iptr >= upper)
-          {
-            ushort_val = upper_val;
-          }
-          else
-          {
-            ushort_val = static_cast<unsigned char>((*iptr + shift) * scale);
-          }
-          *optr = static_cast<unsigned char>((*optr * ushort_val) >> 8);
+          windowFactor = vtkImageMapToWindowLevelAddon::mapScalarToWindow(*iptr, lower, upper, mappingMode);
+          *optr = static_cast<unsigned char>(*optr * windowFactor);
           switch (outputFormat)
           {
             case VTK_RGBA:
-              *(optr + 1) = static_cast<unsigned char>((*(optr + 1) * ushort_val) >> 8);
-              *(optr + 2) = static_cast<unsigned char>((*(optr + 2) * ushort_val) >> 8);
+              *(optr + 1) = static_cast<unsigned char>(*(optr + 1) * windowFactor);
+              *(optr + 2) = static_cast<unsigned char>(*(optr + 2) * windowFactor);
               *(optr + 3) = 255;
               break;
             case VTK_RGB:
-              *(optr + 1) = static_cast<unsigned char>((*(optr + 1) * ushort_val) >> 8);
-              *(optr + 2) = static_cast<unsigned char>((*(optr + 2) * ushort_val) >> 8);
-              break;
-            case VTK_LUMINANCE_ALPHA:
-              *(optr + 1) = 255;
+              *(optr + 1) = static_cast<unsigned char>(*(optr + 1) * windowFactor);
+              *(optr + 2) = static_cast<unsigned char>(*(optr + 2) * windowFactor);
               break;
+            case VTK_LUMINANCE_ALPHA: *(optr + 1) = 255; break;
           }
           iptr += numberOfComponents;
           optr += numberOfOutputComponents;
@@ -338,18 +200,7 @@ void vtkImageMapToWindowLevelColorsExecute(vtkImageMapToWindowLevelColors* self,
       {
         for (idxX = 0; idxX < extX; idxX++)
         {
-          if (*iptr <= lower)
-          {
-            result_val = lower_val;
-          }
-          else if (*iptr >= upper)
-          {
-            result_val = upper_val;
-          }
-          else
-          {
-            result_val = static_cast<unsigned char>((*iptr + shift) * scale);
-          }
+          result_val = static_cast<unsigned char>(255.0 * vtkImageMapToWindowLevelAddon::mapScalarToWindow(*iptr, lower, upper, mappingMode));
           *optr = result_val;
           switch (outputFormat)
           {
@@ -362,9 +213,7 @@ void vtkImageMapToWindowLevelColorsExecute(vtkImageMapToWindowLevelColors* self,
               *(optr + 1) = result_val;
               *(optr + 2) = result_val;
               break;
-            case VTK_LUMINANCE_ALPHA:
-              *(optr + 1) = 255;
-              break;
+            case VTK_LUMINANCE_ALPHA: *(optr + 1) = 255; break;
           }
           iptr += numberOfComponents;
           optr += numberOfOutputComponents;
@@ -382,28 +231,39 @@ void vtkImageMapToWindowLevelColorsExecute(vtkImageMapToWindowLevelColors* self,
 // This method is passed a input and output data, and executes the filter
 // algorithm to fill the output from the input.
 
-void vtkImageMapToWindowLevelColors::ThreadedRequestData(vtkInformation* vtkNotUsed(request),
-  vtkInformationVector** vtkNotUsed(inputVector), vtkInformationVector* vtkNotUsed(outputVector),
-  vtkImageData*** inData, vtkImageData** outData, int outExt[6], int id)
+void vtkImageMapToWindowLevelAddon::ThreadedRequestData(vtkInformation* vtkNotUsed(request),
+                                                        vtkInformationVector** vtkNotUsed(inputVector),
+                                                        vtkInformationVector* vtkNotUsed(outputVector),
+                                                        vtkImageData*** inData,
+                                                        vtkImageData** outData,
+                                                        int outExt[6],
+                                                        int id)
 {
   void* inPtr = inData[0][0]->GetScalarPointerForExtent(outExt);
   void* outPtr = outData[0]->GetScalarPointerForExtent(outExt);
 
   switch (inData[0][0]->GetScalarType())
   {
-    vtkTemplateMacro(vtkImageMapToWindowLevelColorsExecute(this, inData[0][0],
-      static_cast<VTK_TT*>(inPtr), outData[0], static_cast<unsigned char*>(outPtr), outExt, id));
-    default:
-      vtkErrorMacro(<< "Execute: Unknown ScalarType");
-      return;
+    vtkTemplateMacro(vtkImageMapToWindowLevelAddonExecute(this, inData[0][0], static_cast<VTK_TT*>(inPtr), outData[0], static_cast<unsigned char*>(outPtr), outExt, id));
+    default: vtkErrorMacro(<< "Execute: Unknown ScalarType"); return;
   }
 }
 
-void vtkImageMapToWindowLevelColors::PrintSelf(ostream& os, vtkIndent indent)
+void vtkImageMapToWindowLevelAddon::PrintSelf(ostream& os, vtkIndent indent)
 {
   this->Superclass::PrintSelf(os, indent);
+  os << indent << "WindowMappingMode: " << this->MappingMode << endl;
+}
+
 
-  os << indent << "Window: " << this->Window << endl;
-  os << indent << "Level: " << this->Level << endl;
+vtkImageMapToWindowLevelAddon::WindowMappingMode vtkImageMapToWindowLevelAddon::GetMappingMode() const
+{
+  return this->MappingMode;
+}
+
+void vtkImageMapToWindowLevelAddon::SetMappingMode(vtkImageMapToWindowLevelAddon::WindowMappingMode mappingMode)
+{
+  this->MappingMode = mappingMode;
 }
-VTK_ABI_NAMESPACE_END
```

ajallphin:
Updates incorporated in my latest push:
1. Centralized the log10 calculation into a static inline method
2. Added options for choosing Linear, Left Log, or Right Log.
3. Adjusted some overlapping logic for colormap inversion.

ajallphin:
https://github.com/user-attachments/assets/10a829da-90da-45a9-a55b-2aec557635c9

lassoan:
Thanks for the updates and the ping, I'll review this tomorrow.

jamesobutler:
@lassoan I have gone ahead and rebased this branch against latest `main` for you to review.

jamesobutler:
@lassoan it appears my recent [push](https://github.com/Slicer/Slicer/compare/8f868e9b4c6ca4d145cd29fd66abc31eed11625a..3f06efe3ef94e3a9ecb818de7bc0a085a8309828) resolved the CI build error on the Linux platform due to a missing includes. This was not caught during original local testing on the Windows platform where the build error did not occur.

jamesobutler:
@lassoan Any comments on this set of work?

jamesobutler:
@lassoan It would be great to get this set of work into Slicer 5.12.

lassoan:
@ajallphin I've pushed a few small fixes and improvements to this branch, please have a look and test if everything works well for you. Thanks!

lassoan:
It seems that the names of the log compression options are mixed up.

Linear (for reference):
image

Log compress towards low values: -> low values are actually not compressed but expanded (we can see difference between low-value voxels better)
image

Log compress towards high values: -> low values are compressed (we cannot see the difference between low-value voxels as much as in case of linear)
image

"Compress X" is also not the best name, because it mentions X, but actually X gets suppressed.

What do you think about changing the combobox options:

from these:
- Linear
- Log compress towards low values
- Log compress towards high values

to these:
- Linear
- Logarithmic (emphasize low values)
- Inverse logarithmic (emphasize high values)

(and update the enum names in `vtkImageMapToWindowLevelAddon` accordingly)

jamesobutler:
> from these:
> 
> - Linear
> - Log compress towards low values
> - Log compress towards high values
> to these:
> 
> - Linear
> - Logarithmic (emphasize low values)
> - Inverse logarithmic (emphasize high values)

@lassoan Yes, these new names are good with me. I have made these updates including the enums and have pushed to update the primary commit.

https://github.com/Slicer/Slicer/pull/9016/commits/995b70c84c9233cef7797410f162fd24a844a887 this has now also been squashed into the 2 relevant primary commits with your co-authorship added. Thanks for these updates.

FYI, I have made these updates on behalf of @ajallphin who is out on vacation this week. 

> BUG: Small fixes and improvements
> 
> vtkImageMapToWindowLevelAddon:
> - Log mapping mode incorrectly triggered the pass-through path in
>   RequestInformation, causing wrong output buffer size
> - Negative window was treated as zero-width range instead of an
>   inversion signal. This is a general-purpose filter, so it should be
>   backward-compatible with vtkImageMapToWindowLevelColors even if
>   negative range is not used in Slicer.
> - Per-frame inversion and range normalization are now precomputed once
>   to minimize computations in a very hot loop.
> 
> vtkSlicerVolumeRenderingLogic:
> - Color transfer function endpoints were added out of order, requiring
>   a tracker reset that allowed collisions when the window is narrow
> - LUT color index used truncation causing off-by-one at the upper
>   boundary; use lround instead
> 
> vtkMRMLScalarVolumeDisplayNode: Unrecognized windowMappingMethod XML attribute was silently ignored.
> 
> qSlicerScalarVolumeDisplayWidget: ScalarMappingComboBox labels were not marked for translation.

lassoan (approved):
It looks good to me, thank you! I've added a few lines to the user documentation and rebased on latest main branch. It should be all good to go.

Please consider posting a topic on the Slicer Forum that describes this new feature. Add the `feature` tag to the topic.

ajallphin:
Sorry that I was out of town when this was finalized. It looks like we introduced one little bug. James you mentioned in a previous comment the following change:
>vtkSlicerVolumeRenderingLogic:
>
>Color transfer function endpoints were added out of order, requiring a tracker reset that allowed collisions when the window is narrow

The out of order placement was actually intentional because the "Higher and Unique" method doesn't let us place intermediate control points. So things are not working correctly right now:

PR #9020, opened 2026-02-07
PERF: Enable GENERATECLP_USE_MD5 to speed up builds on Windows

benbennett:
Speed up incremental builds using MD5 checksum.

Depends on Slicer/SlicerExecutionModel#146 for Windows.


jamesobutler (commented):
This option appears to have originally been added to Slicer Exection Model 15 years ago in https://github.com/Slicer/SlicerExecutionModel/commit/4ada01824308bd23a11aac820ce1d2f92cd4b4b7. Based on the option text saying:

> "Enabling that option will teach GenerateCLP to overwrite output file only if needed. Make sure to build ITK with KWSYS_USE_MD5 enabled."

Is there an ITK build option that has to be modified as well? `KWSYS_USE_MD5` still exists as an option in ITK?

lassoan:
https://github.com/Slicer/SlicerExecutionModel/pull/146 has been merged now.

Issue #8947, opened 2026-01-06
Segmentation fault when importing SimpleITK in PythonSlicer


vboussot:
## Summary

Importing SimpleITK in Slicer’s bundled Python (PythonSlicer) causes an immediate segmentation fault on Linux.

This issue affects all Slicer extensions that rely on SimpleITK, since any attempt to import SimpleITK in the Slicer Python environment crashes the interpreter, making such extensions unusable.

## Steps to reproduce

The issue is reproducible both in a Linux source build and in the Slicer Preview Release 5.11.0 (revision 34365, built 2026-01-06) using Python 3.12.10.

1) Launch Slicer or use the bundled Python interpreter (`PythonSlicer`).
2) In the Python interpreter, run: import SimpleITK
3) Actual behavior: Slicer (or `PythonSlicer`) crashes immediately with a segmentation fault.

## Environment
- Slicer version: Slicer-5.11.0 revision 34365 built 2026-01-06 
- Operating system: Linux

pieper:
Do you have a debug build to get a stack trace?

blowekamp:
I was able to reproduce the issue with:
```
./bin/PythonSlicer -c "import SimpleITK"
```

The following did not have the issue:
```
export PYTHONPATH=$(pwd)/./lib/Python/lib/python3.12/site-packages; bin/python-real -c "import SimpleITK"
```

Trying to run GDB with the real executable I am getting a runtime linking error:
```
eval $(./bin/PythonSlicer --launcher-show-set-environment-commands)
gdb --args bin/python-real -c "import SimpleITK"
```

Any suggestions for running gdb here?

Also I am looking for the implementation of PythonSlicer. Is there a C++ file or something?

pieper:
> Any suggestions for running gdb here?

You should be able to start PythonSlicer and then attach gdb to it.  Once you're attached run the import command.

gdb uses python, so it's not compatible with being in the same environment that run Slicer's shared libs.

vboussot:
Steps:
  PythonSlicer -c "import SimpleITK"

GDB:
```text
Program received signal SIGSEGV, Segmentation fault.
0x00007fa3621a7cf5 in slicer_itk::ObjectFactoryBase::CreateObject(char const*) () from /home/vorps/Documents/lib/Slicer-build/ITK-build/lib/./libITKCommon-5.4.so.1
(gdb) thread apply all bt

Thread 1 (Thread 0x7fa37a18f740 (LWP 341592) "python"):
[...]

PR #9023, opened 2026-02-09
COMP: Add option for IDImageIO support in CMake configuration

Punzo:
@pieper @lassoan @blowekamp

Alternative solution for https://github.com/Slicer/Slicer/issues/8947

I have simply added a cmake option to disable the build of the IDImageIO plugin (default OFF), but if you would like to completely remove the plugin from Slicer core code, please let me know.

See also 
https://github.com/Slicer/ITK/pull/14
https://github.com/Slicer/ITK/pull/15
https://github.com/Slicer/Slicer/pull/9021

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -430,6 +430,9 @@ CMAKE_DEPENDENT_OPTION(
   "Slicer_USE_PYTHONQT" OFF)
 mark_as_superbuild(Slicer_USE_SimpleITK)
 
+option(Slicer_USE_IDImageIO "Build Slicer with IDImageIO support" OFF)
+mark_as_superbuild(Slicer_USE_IDImageIO)
+
 option(Slicer_BUILD_PARAMETERSERIALIZER_SUPPORT "Build Slicer with parameter serializer support" OFF)
 mark_as_superbuild(Slicer_BUILD_PARAMETERSERIALIZER_SUPPORT)
 
--- a/Libs/CMakeLists.txt
+++ b/Libs/CMakeLists.txt
@@ -34,8 +34,12 @@ list(APPEND dirs
   RemoteIO
   MRML/Logic
   MRML/DisplayableManager
-  MRML/IDImageIO
   )
+if(Slicer_USE_IDImageIO)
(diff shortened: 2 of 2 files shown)


Punzo:
> This looks good to me. If we run into any problems then it is easy to quickly re-enable it.
> 
> When we remove MRMLIDImageIO fully then we'll need to remove MRMLIDImageIO source code and any setting of ITK_AUTOLOAD_PATH.

Perfect! I have updated the commit messagge to make clear what is the role of IDImageIO MRML plugin on the CLI modules.

blowekamp:
You may just be able to configure ITK with "ITK_DYNAMIC_LOADING=OFF":
https://github.com/InsightSoftwareConsortium/ITK/blob/5b25f7248ffabcb884c1e39f3a9afd340f62252e/Modules/Core/Common/src/itkObjectFactoryBase.cxx#L354

This should disable ITK from trying to load everything in ITK_AUTOLOAD_PATH. You could still keep the ImageIO's around this way, and other executables wouldn't try to load them.

EDIT: Well... Thinking more SimpleITK would still have dynamic loading enabled, and if ITK_AUTOLOAD_PATH was still set to the Slicer ImageIO with itkLoad, we'd have a problem again. Similarly other external ITK applications like ants, elastic, etc may also try to load anything in the specified path. So... I had the direction of the autoload backwards. 

So yes unsetting the ITK_AUTOLOAD should fix third party ITK applications.

blowekamp:
One more  thought :)

If just the "itkLoad" function were removed from the ImageIOs, and the ITK_AUTOLOAD_PATH unset, then the built in slicer extensions could still use the ImageIO, while no third-party ITK applications neither would try, nor have anything to auto load.

lassoan:
This is a simple and safe change, which is useful regardless if we decide to fix the MRMLIDIO mechanism in the future; and it immediately unblocks several developers who use linux, so I'll merge this now.

If there is enough motivation and time to fix the plugin autoloading in ITK and Slicer in the future, then we can re-enable this feature; if we don't think this will happen then we can remove the MRMLIDIO plugin for good.

PR #9030, opened 2026-02-17
ENH: Make Teem a private dependency

AlexyPellegrini:
Teem does not need to be a public dependency as it is an implementation detail that leaked due to historical reasons:
- vtkTeemNRRDReader was implemented first 20 years ago and included teem.h in its header
- vtkTeemNRRDWriter is newer and partially hide the teem include already

This is a small breaking change, as it changes the signature of two protected functions and the type of a protected member, but I do not expect any code to do anything with them in practice.

This will be critical to build Slicer as a SDK in the context of SlicerCore project as Teem CMake install tree is ill-formed, making it a private dependency removes the need to use it and makes packaging of SlicerCore easier.

Please review @Thibault-Pelletier @lassoan 
FYI @finetjul

### DIFF
--- a/Libs/vtkTeem/vtkTeemNRRDReader.cxx
+++ b/Libs/vtkTeem/vtkTeemNRRDReader.cxx
@@ -50,6 +50,7 @@
 #include <vtksys/SystemTools.hxx>
 
 // Teem includes
+#include "teem/nrrd.h"
 #include "teem/ten.h"
 
 vtkStandardNewMacro(vtkTeemNRRDReader);
@@ -72,7 +73,7 @@ vtkTeemNRRDReader::vtkTeemNRRDReader()
 //----------------------------------------------------------------------------
 vtkTeemNRRDReader::~vtkTeemNRRDReader()
 {
-  nrrdNuke(this->nrrd);
+  nrrdNuke(static_cast<Nrrd*>(this->nrrd));
   this->nrrd = nullptr;
 }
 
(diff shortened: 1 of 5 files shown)


lassoan:
@AlexyPellegrini could you look at the lint issue?

Libs/vtkTeem/vtkTeemNRRDReader.cxx
```diff
@@ -243,8 +244,10 @@ int vtkTeemNRRDReader::CanReadFile(const char* filename)
 }
 
 //----------------------------------------------------------------------------
-bool vtkTeemNRRDReader::GetPointType(Nrrd* nrrdTemp, int& pointDataType, int& numOfComponents)
+bool vtkTeemNRRDReader::GetPointType(void* _nrrdTemp, int& pointDataType, int& numOfComponents)
 {
+  auto* nrrdTemp = static_cast<Nrrd*>(_nrrdTemp);
```
finetjul, line 249: shouldn't it be a `reinterpret_cast` and not a `static_cast`?

AlexyPellegrini: Not in this case, casting from `void*` to `T*` never requires using reinterpret_cast.

Fun fact, reinterpreting `T*` to `U*` is strictly equivalent to `static_cast<U*>(static_cast<void*>(t))`. [ref, N°5](https://en.cppreference.com/w/cpp/language/reinterpret_cast.html#Explanation)

AlexyPellegrini:
I fixed the clang-format issue

PR #9032, opened 2026-02-19
BUG: Fix python errors when the application path has special characters

lassoan:
When Slicer was installed in a path that contained special characters then Python logging (lib\Python\Lib\logging\__init__.py) sometimes failed, printing the message: "--- Logging error ---"

Setting PYTHONUTF8=1 environment variable for Slicer completes "UTF8 everywhere", not just for C++ but also for Python, and not just for linux and macOS but also for Windows (where the default locale is still most of the time not UTF8). This change makes the behavior more consistent across different operating systems and configurations.

Effects of the change:
- sys.getfilesystemencoding() returns 'utf-8'.
- locale.getpreferredencoding() returns 'utf-8' (while ignoring the actual system locale).
- Standard Streams: sys.stdin, sys.stdout, and sys.stderr all use UTF-8.


lassoan:
If there are no objections, I'll merge this in the next couple of days.

PR #9035, opened 2026-02-20
ENH: Remove layout node assert impacting tests in debug mode

jamesobutler:
qMRMLLayoutManagerVisibilityTest and some other tests would throw an assert dialog on the Windows platform causing the batch run of automated tests to pause. The handling of an existing layout node already exists, so also having an assert appeared excessive.

image


cpinter:
I think it makes sense

jamesobutler:
@lassoan @ebrahimebrahim Any objections here? Or can we integrate? This will help when triggering all of automated testing with a debug build and letting it run unattended.

PR #9036, opened 2026-02-20
COMP: Unify the use of generated Export.h files in Base and MRML libs

AlexyPellegrini:
This MR simplify and unify how the export headers (vtk...Export.h) files are generated and used. Most of the uses in the code base were already unified around a generated Export.h file using some Slicer specific macros.

MRML libraries used to have manually written code to export symbols on Windows.
This has been replaced a long time ago in the other parts of Slicer using SlicerMacroBuildModuleVTKLibrary or similar.

This will make the transition to SlicerCore easier as the file can be replaced silently with VTK generated export header.

I ran the tests locally and everything seems fine on Windows, which is the only OS in which I expect this change to matter.

Please review @Thibault-Pelletier @lassoan 

FYI @finetjul


lassoan (changes requested):
The linux build error in CI seems to be related to this PR. Please check!

AlexyPellegrini:
> The linux build error in CI seems to be related to this PR. Please check!

Indeed, it is now fixed.

> I've checked and the backward-incompatible renaming of vtk*Configure.h files does not affect extensions, as they only include vtkSlicerConfigure.h.

That is good news 👍

finetjul:
@lassoan is it fine with you ? can it be merged ?

PR #9042, opened 2026-02-26
ENH: Scroll to newly added segment in the segment editor

Punzo:
@lassoan with this PR when a new segment is added, the segment editor automatically scrolls to show it. This lets users immediately rename or recolor the new segment without manual scrolling.


lassoan (approved):
Thank you, this fix makes sense, as adding a new segment already selects that segment, but it may not be visible in the segment list if the list is long.

Punzo:
> Thank you, this fix makes sense, as adding a new segment already selects that segment, but it may not be visible in the segment list if the list is long.

thanks for merging!

PR #9043, opened 2026-02-26
BUG: Fix leaked raw observers in qMRMLThreeDView

mhalle:
## Summary

The `qMRMLThreeDView` constructor creates a local `vtkSmartPointer<vtkCallbackCommand>` and adds 4 observers to the render window interactor (`MouseWheelForwardEvent`, `MouseWheelBackwardEvent`, `InteractionEvent`, `KeyPressEvent`), but:

- The callback command smart pointer is **never stored** as a class member
- The destructor has **no `RemoveObserver` calls**
- `ClickCallbackFunction` does `reinterpret_cast<qMRMLThreeDView*>(clientData)` — if the callback fires after the widget is destroyed, this is a **use-after-free**

## Changes

- Store the callback command in `qMRMLThreeDViewPrivate::ClickCallback` (added as a `vtkSmartPointer<vtkCallbackCommand>` member)
- Remove all observers in the destructor before renderer cleanup via `interactor->RemoveObserver(d->ClickCallback)`

## How this was found

Static analysis of observer lifecycle patterns across the Slicer codebase, checking for `AddObserver` calls without matching `RemoveObserver` in the same class, its private implementation, or ancestor destructors.

## Test plan

- [ ] Verify 3D view mouse wheel zoom, keyboard interaction, and camera interaction events still work
- [ ] Verify no crash on repeated creation/destruction of 3D views (e.g., switching layouts)
- [ ] Run existing `qMRMLThreeDView` tests

### DIFF
--- a/Libs/MRML/Widgets/qMRMLThreeDView.cxx
+++ b/Libs/MRML/Widgets/qMRMLThreeDView.cxx
@@ -260,20 +260,27 @@ qMRMLThreeDView::qMRMLThreeDView(QWidget* _parent)
 
   vtkRenderWindowInteractor* renderWindowInteractor = this->interactor();
 
-  vtkSmartPointer<vtkCallbackCommand> clickCallback = vtkSmartPointer<vtkCallbackCommand>::New();
-  clickCallback->SetClientData(this);
-  clickCallback->SetCallback(ClickCallbackFunction);
+  d->ClickCallback = vtkSmartPointer<vtkCallbackCommand>::New();
+  d->ClickCallback->SetClientData(this);
+  d->ClickCallback->SetCallback(ClickCallbackFunction);
 
-  renderWindowInteractor->AddObserver(vtkCommand::MouseWheelForwardEvent, clickCallback);
-  renderWindowInteractor->AddObserver(vtkCommand::MouseWheelBackwardEvent, clickCallback);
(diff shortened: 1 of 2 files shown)


lassoan:
> The callback command smart pointer is never stored as a class member

It's fine. VTK does internal reference counting. The command object is kept alive by the objects that use it, until they need it.
> The destructor has no RemoveObserver calls

It's fine. When an object is deleted, the observers are deleted automatically. 

> ClickCallbackFunction does reinterpret_cast<qMRMLThreeDView*>(clientData) — if the callback fires after the widget is destroyed, this is a use-after-free

VTK's internal reference counting ensures that the callback command object is kept alive until it is no longer needed.

To show the call stack of deletion of the callback command, I've used a custom class and added a breakpoint in the destructor:

image

The call stack shows that when the internal RenderWindow is deleted, it deletes the RenderWindowInteractor, which deletes all its observers, which automatically deletes the CallbackFunction object.

That said, observers are indeed have to be removed manually quite often (e.g., because when the object is removed that added the observation, the observation is not automatically removed). To avoid the need to keep track of observations manually, it is recommended to use the [EventBroker](https://slicer.readthedocs.io/en/latest/developer_guide/advanced_topics.html#event-broker) instead of AddObserver/RemoveObserver directly. When we touch older code that does not use the EventBroker then we often refactor the code, but we don't refactor working code just for sake of consistency.

PR #9044, opened 2026-02-26
BUG: Add missing Superclass::setup() calls in 13 module widgets

mhalle:
[Edited based on @jamesobutler's comments]

## Summary

`qSlicerAbstractModuleWidget::setup()` sets `objectName`, `windowTitle`, and `windowIcon` from module metadata. 13 module widget subclasses override `setup()` without calling `Superclass::setup()`, causing missing widget, but not module, identity (invisible to Qt debugging tools).

27 other module widgets in the codebase correctly call `this->Superclass::setup()`. These 13 are the exceptions.

## Changes

Added `this->Superclass::setup();` immediately after the `Q_D()` macro and before `d->setupUi(this)` in each affected class, matching the pattern used by all other module widgets:

- `qSlicerCLIModuleWidget`
- `qSlicerEventBrokerModuleWidget`
- `qSlicerCamerasModuleWidget`
- `qSlicerColorsModuleWidget`
- `qSlicerDataModuleWidget`
- `qSlicerPlotsModuleWidget`
- `qSlicerReformatModuleWidget`
- `qSlicerTablesModuleWidget`
- `qSlicerTextsModuleWidget`
- `qSlicerTransformsModuleWidget`
- `qSlicerViewControllersModuleWidget`
- `qSlicerVolumeRenderingModuleWidget`
- `qSlicerVolumesModuleWidget`

## How this was found

Static analysis of lifecycle method implementations across the Slicer codebase, checking for overrides of key virtual methods (`setup`, `enter`, `exit`, `SetMRMLSceneInternal`, etc.) that do not call their `Superclass::` counterpart.

## Test plan

- [ ] Verify no regressions in module UI behavior (all `setupUi` calls remain in place after the super call)
- [ ] Run existing module widget tests

### DIFF
--- a/Modules/Loadable/Colors/qSlicerColorsModuleWidget.cxx
+++ b/Modules/Loadable/Colors/qSlicerColorsModuleWidget.cxx
@@ -133,6 +133,7 @@ qSlicerColorsModuleWidget::~qSlicerColorsModuleWidget() = default;
 void qSlicerColorsModuleWidget::setup()
 {
   Q_D(qSlicerColorsModuleWidget);
+  this->Superclass::setup();
 
   d->setupUi(this);
 
--- a/Base/QTCLI/qSlicerCLIModuleWidget.cxx
+++ b/Base/QTCLI/qSlicerCLIModuleWidget.cxx
@@ -323,6 +323,7 @@ qSlicerCLIModuleWidget::~qSlicerCLIModuleWidget() = default;
 void qSlicerCLIModuleWidget::setup()
 {
   Q_D(qSlicerCLIModuleWidget);
+  this->Superclass::setup();
 
   d->setupUi(this);
 }
--- a/Modules/Core/EventBroker/qSlicerEventBrokerModuleWidget.cxx
+++ b/Modules/Core/EventBroker/qSlicerEventBrokerModuleWidget.cxx
@@ -55,6 +55,7 @@ qSlicerEventBrokerModuleWidget::~qSlicerEventBrokerModuleWidget() = default;
 void qSlicerEventBrokerModuleWidget::setup()
 {
   Q_D(qSlicerEventBrokerModuleWidget);
+  this->Superclass::setup();
   d->setupUi(this);
 }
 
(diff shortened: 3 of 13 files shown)


Modules/Loadable/Colors/qSlicerColorsModuleWidget.cxx
```diff
@@ -133,6 +133,7 @@ qSlicerColorsModuleWidget::~qSlicerColorsModuleWidget() = default;
 void qSlicerColorsModuleWidget::setup()
 {
   Q_D(qSlicerColorsModuleWidget);
+  this->Superclass::setup();
```
jamesobutler, line 136: > causing missing widget identity (wrong title/icon in module panel tab

How so observed?

The following showing of the Colors module widget with current Slicer revealed that the title and icon were correct for the window.
```python
module_widget = slicer.modules.colors.widgetRepresentation()
module_widget.show()
```

Image

Also also currently correct in the module selector menu:

Image

mhalle: You're right, the superclass does this:
```
  // qSlicerAbstractModuleWidget::setup() sets three Qt widget properties:
  this->setObjectName(QString("%1ModuleWidget").arg(m->name()));
  this->setWindowTitle(m->title());
  this->setWindowIcon(m->icon());
```

But Slicer doesn't use them. Qt debugging and accessibility do. 

The patch is valid, though: modules should consistently call their superclass, and these modules are the exceptions.

jamesobutler: > But Slicer doesn't use them. Qt debugging and accessibility do.

What do you mean by this? What failure in the UI is being currently observed with Slicer that this PR is fixing?

mhalle: I'm agreeing that there is not a failure manifested in the Slicer UI, as I thought before. 

It is a straightforward code defect, though, and these modules are inconsistent with the practice in the other Slicer modules.

jamesobutler (approved):

PR #9045, opened 2026-02-26
BUG: Fix copy-paste error in oversampling fuzzy membership function

mhalle:
## Summary

Copy-paste error in `Libs/vtkSegmentationCore/vtkCalculateOversamplingFactor.cxx` where the `oversamplingNormal` piecewise function has a duplicate x-value, creating a degenerate triangle instead of the intended trapezoid.

**Before (line 260, broken):**
```cpp
oversamplingNormal->AddPoint(-0.75, 0);
oversamplingNormal->AddPoint(0.25, 1);   // <-- should be -0.25
oversamplingNormal->AddPoint(0.25, 1);
oversamplingNormal->AddPoint(0.75, 0);
```

Compare with `oversamplingHigh`, which has the correct trapezoid shape:
```cpp
oversamplingHigh->AddPoint(0.25, 0);
oversamplingHigh->AddPoint(0.75, 1);
oversamplingHigh->AddPoint(1.25, 1);
oversamplingHigh->AddPoint(1.75, 0);
```

Both should be symmetric trapezoids with 0.5-unit spacing between control points. The `oversamplingHigh` trapezoid correctly has four distinct x-values (0.25, 0.75, 1.25, 1.75). The `oversamplingNormal` trapezoid should be (-0.75, -0.25, 0.25, 0.75) but the second point was copied from the third without changing the sign.

**After (fixed):**
```cpp
oversamplingNormal->AddPoint(-0.75, 0);
oversamplingNormal->AddPoint(-0.25, 1);   // fixed
oversamplingNormal->AddPoint(0.25, 1);
oversamplingNormal->AddPoint(0.75, 0);
```

## Impact

This is a correctness issue. The `oversamplingNormal` membership function lacks the flat plateau that defines its "definitely normal" region. The fuzzy inference system for automatic oversampling factor calculation during closed surface to binary labelmap conversion does not match its intended design.

## Test plan

- [ ] Convert a medium-sized, low-complexity closed surface segment to binary labelmap with automatic oversampling — verify the oversampling factor is closer to 1x than before
- [ ] Verify no change in behavior for very small structures (should still get high oversampling)
- [ ] Run existing segmentation conversion tests

### DIFF
--- a/Libs/vtkSegmentationCore/vtkCalculateOversamplingFactor.cxx
+++ b/Libs/vtkSegmentationCore/vtkCalculateOversamplingFactor.cxx
@@ -254,7 +254,7 @@ double vtkCalculateOversamplingFactor::DetermineOversamplingFactor()
   vtkSmartPointer<vtkPiecewiseFunction> oversamplingLow = vtkSmartPointer<vtkPiecewiseFunction>::New();
   oversamplingLow->AddPoint(-1.25, 1);
   oversamplingLow->AddPoint(-0.75, 1);
-  oversamplingLow->AddPoint(0.25, 0);
+  oversamplingLow->AddPoint(-0.25, 0);
   vtkSmartPointer<vtkPiecewiseFunction> oversamplingNormal = vtkSmartPointer<vtkPiecewiseFunction>::New();
   oversamplingNormal->AddPoint(-0.75, 0);
   oversamplingNormal->AddPoint(-0.25, 1);


cpinter:
I think this is the function that this code is creating, specifically the red trapezoid:
image
(from paper https://link.springer.com/article/10.1007/s00500-020-05126-w ([PDF](https://labs.cs.queensu.ca/perklab/wp-content/uploads/sites/3/2024/03/Pinter2020a_0.pdf)))

The suggested changes looks correct.

Thank you for the fix!

cpinter (approved):

cpinter:
Not clicking merge yet, @lassoan please do if you agree after seeing the figure I posted.

lassoan:
To me it looks like that then the last point of `oversamplingLow` would need to be adjusted as well. Instead of this:

    oversamplingLow->AddPoint(-1.25, 1);
    oversamplingLow->AddPoint(-0.75, 1);
    oversamplingLow->AddPoint(0.25, 0);

we should have:

    oversamplingLow->AddPoint(-1.25, 1);
    oversamplingLow->AddPoint(-0.75, 1);
    oversamplingLow->AddPoint(-0.25, 0);

cpinter:
Yes, good catch.

Libs/vtkSegmentationCore/vtkCalculateOversamplingFactor.cxx
```diff
@@ -257,7 +257,7 @@ double vtkCalculateOversamplingFactor::DetermineOversamplingFactor()
   oversamplingLow->AddPoint(0.25, 0);
```
cpinter:
```suggestion
  oversamplingLow->AddPoint(-0.25, 0);
```

cpinter: Just playing with the suggestion feature to see if we can amend the existing commit but maybe not...

lassoan (approved):
It all looks good now.

cpinter:
Thank you very much!

PR #9048, opened 2026-02-26
Fix model rendering black under reflection transforms

### LINKED ISSUE #9018: Model turns black under transformation containing reflection
## Summary
Surface models in Slicer turn black and become unlightable when a transformation containing a reflection is applied to them.  This is unexpected and unhelpful.  The root cause is the reflection turns the model inside-out, and to make it lightable again, a vtkReverseSense filter needs to be applied. See discussion at: https://discourse.slicer.org/t/surface-model-turns-black-under-transformation/46065

## Steps to reproduce

1. Load or create a model node containing a surface mesh. 
2. Create a transform node whose transform matrix has a negative determinant, e.g. numpy.diag([-1,1,1,1])
3. Apply the transform to the model node.

The model will turn black and become impervious to lighting and color changes.  Instead, the model should stay the same color as before transformation. 

Image

## Suggested fix
The model's displayable manager should detect and automatically apply vtkReverseSense filter if being put under a transform with negative determinant.  Surface Toolbox module extension Mirror functionality demonstrates this approach.

## Environment
- Slicer version: Slicer-5.10

mhalle:
## Summary

- When a transform with negative determinant (reflection) is applied to a surface model, the model turns black and becomes impervious to lighting and color changes
- Root cause: `SetUserMatrix()` with a reflection matrix reverses polygon winding on the GPU, causing normals to appear inverted
- Fix: insert a `vtkReverseSense` filter (with `ReverseCellsOn` + `ReverseNormalsOn`) into the model display pipeline when a reflection transform is detected; automatically remove it when the transform changes to a non-reflection

## Details

The fix follows existing patterns in the codebase:
- **`vtkDiffusionTensorGlyph.cxx`** — detects negative determinant, flips normals via inverse scale
- **`ModelMaker.cxx`** — uses `vtkReverseSense` when IJK-to-LPS determinant is negative
- **`vtkCapPolyData.cxx`** — reverses cells when polygon normal opposes plane normal

The `vtkReverseSense` filter is managed per-display-node (like existing `DisplayNodeTransformFilters`), cached across updates, and properly cleaned up on node removal.

**Scope:** Handles linear transforms only. Non-linear transforms with local reflections are a much rarer edge case that would require per-cell Jacobian evaluation. Unstructured grid meshes (which use `vtkDataSetMapper` instead of `vtkPolyDataMapper`) are also not handled, as `vtkReverseSense` only supports polydata.

Fixes #9018
See also: https://discourse.slicer.org/t/surface-model-turns-black-under-transformation/46065

## Test plan

- [ ] Load or create a surface model (e.g., via Sample Data → MRHead)
- [ ] Create a linear transform node with negative determinant (e.g., `numpy.diag([-1,1,1,1])`)
- [ ] Apply the transform to the model — model should retain correct colors and lighting
- [ ] Remove the transform — model should still render correctly
- [ ] Toggle the transform on/off rapidly — no rendering artifacts
- [ ] Test with non-reflection transforms — no change in behavior
- [ ] Test model with clipping enabled under reflection transform

### DIFF
--- a/Libs/MRML/DisplayableManager/vtkMRMLModelDisplayableManager.cxx
+++ b/Libs/MRML/DisplayableManager/vtkMRMLModelDisplayableManager.cxx
@@ -28,6 +28,7 @@
 #include <vtkMRMLSliceNode.h>
 #include <vtkMRMLSubjectHierarchyConstants.h>
 #include <vtkMRMLSubjectHierarchyNode.h>
+#include <vtkMRMLTransformableNode.h>
 #include <vtkMRMLTransformNode.h>
 #include <vtkMRMLViewNode.h>
 #include <vtkMRMLVolumeNode.h>
@@ -66,6 +67,7 @@
 #include <vtkProp3DCollection.h>
 #include <vtkProperty.h>
 #include <vtkRenderWindowInteractor.h>
+#include <vtkReverseSense.h>
 #include <vtkSmartPointer.h>
 #include <vtkTexture.h>
 #include <vtkTransform.h>
@@ -114,6 +116,8 @@ public:
   std::map<std::string, vtkSmartPointer<vtkCapPolyData>>        Cappers;
   std::map<std::string, vtkSmartPointer<vtkProp3D>>             DisplayedCapActors;
   std::map<std::string, vtkSmartPointer<vtkTransformFilter>>    DisplayNodeCapTransformFilters;
+  std::map<std::string, vtkSmartPointer<vtkReverseSense>>      DisplayNodeReverseSenseFilters;
+  std::map<std::string, vtkSmartPointer<vtkReverseSense>>      DisplayNodeCapReverseSenseFilters;
   // clang-format on
 
   bool IsUpdatingModelsFromMRML;
@@ -684,6 +688,11 @@ void vtkMRMLModelDisplayableManager::ClearDisplayMaps()
     transformFilter->SetTransform(nullptr);
   }
   this->Internal->DisplayNodeTransformFilters.clear();
+  for (auto iter = this->Internal->DisplayNodeReverseSenseFilters.begin(); iter != this->Internal->DisplayNodeReverseSenseFilters.end(); iter++)
+  {
+    iter->second->SetInputConnection(nullptr);
+  }
+  this->Internal->DisplayNodeReverseSenseFilters.clear();
   this->Internal->Clippers.clear();
   this->Internal->Cappers.clear();
   if (this->GetRenderer())
@@ -701,6 +710,11 @@ void vtkMRMLModelDisplayableManager::ClearDisplayMaps()
     transformFilter->SetTransform(nullptr);
   }
   this->Internal->DisplayNodeCapTransformFilters.clear();
+  for (auto iter = this->Internal->DisplayNodeCapReverseSenseFilters.begin(); iter != this->Internal->DisplayNodeCapReverseSenseFilters.end(); iter++)
+  {
+    iter->second->SetInputConnection(nullptr);
+  }
+  this->Internal->DisplayNodeCapReverseSenseFilters.clear();
 }
 
 //---------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


Libs/MRML/DisplayableManager/vtkMRMLModelDisplayableManager.cxx
```diff
@@ -1317,6 +1331,58 @@ void vtkMRMLModelDisplayableManager::SetModelDisplayProperty(vtkMRMLDisplayableN
     vtkImageActor* imageActor = vtkImageActor::SafeDownCast(prop);
     prop->SetUserMatrix(matrixTransformToWorld);
 
+    // Handle reflection in linear transforms.
```
lassoan: The exact same mechanism has to be applied to the `capActor`.

Libs/MRML/DisplayableManager/vtkMRMLModelDisplayableManager.cxx
```diff
@@ -1317,6 +1331,58 @@ void vtkMRMLModelDisplayableManager::SetModelDisplayProperty(vtkMRMLDisplayableN
...
+    // When a transform matrix has a negative determinant (reflection), polygon winding
+    // reverses on the GPU, causing normals to appear inverted (model renders black).
+    // Insert a vtkReverseSense filter to pre-flip cells and normals so they are correct
+    // after the GPU applies the reflection transform.
+    // See https://github.com/Slicer/Slicer/issues/9018
```
lassoan: The code (with comments) should stand by itself. It should not be necessary to reference an external issue for further details. I would only include links to issues for hacks and temporary solutions, which would be too difficult to explain in the code and things may still evolve, etc. but this will be a permanent solution.

Of course, relevant issues must be referenced in the commit comment, but the issue number is there already.

```suggestion
```

Libs/MRML/DisplayableManager/vtkMRMLModelDisplayableManager.cxx
```diff
@@ -1317,6 +1331,58 @@ void vtkMRMLModelDisplayableManager::SetModelDisplayProperty(vtkMRMLDisplayableN
...
+    if (actor)
+    {
+      vtkPolyDataMapper* polyMapper = vtkPolyDataMapper::SafeDownCast(actor->GetMapper());
+      if (polyMapper && polyMapper->GetNumberOfInputConnections(0) > 0)
+      {
+        bool needReverseSense = (matrixTransformToWorld->Determinant() < 0.0);
+
+        std::string displayNodeID(modelDisplayNode->GetID());
+        auto rsIt = this->Internal->DisplayNodeReverseSenseFilters.find(displayNodeID);
```
lassoan: `rsIt->second` may be `nullptr` (e.g., if somebody ever calls `this->Internal->DisplayNodeReverseSenseFilters[displayNodeID]` instead of `find`). I would set `rsIt` here back to `this->Internal->DisplayNodeReverseSenseFilters.end()` if it stores nullptr, because `rsIt->second` is dereferenced later without any further checks.

lassoan (commented):
Thanks for working on this, this is a very useful feature. I've just added a few comments inline to add some simple improvements.

In addition to those, it is important that if we make this change then **we must also modify the `vtkMRMLModelNode::ApplyTransform`** to apply reverse-sense when an orientation-reversing transform is applied. Without that, the rendering would break when the user chooses to harden the transform on the model.

Also note that the computation in this method is not completely trivial when we have a composite transform that contains linear and non-linear transforms. For now, we can ignore the non-linear transforms or say that the transform is not orientation-reversing if there is a non-linear component, but in the future we may want to improve this. To make sure that reverse-sense is applied consistently in both the displayable manager and in the model node's `ApplyTransform` method, **we should have one single method that decides if reverse-sense should be applied** (e.g., a `bool IsOrientationReversingTransform(vtkAbstractTransform*)` method) and use that method at both places. The `vtkMRMLTransformableNode` would be a good place for this method, as this computation may be needed for other nodes, too (e.g., the GPU raycast mapper cannot render volumes that are under an orientation-reversing transform).

To still allow developers to have full control over what is happening to nodes when they are transformed or rendered, it would be useful to **add to `vtkMRMLTransformableNode` a bool member variable (e.g., `AutoReverseOrientation`)** - enabled by default. Whenever the automatic reversal adds too much complication (e.g., in a custom application that works with composite transforms), the developer can disable it for certain nodes.

mhalle:
I apologize that this PR was premature. I was planning to edit it more, I'll come back to it shortly.

lassoan:
No problem at all, it is a good solution, just not complete (if we want consistent and controllable behavior).

mhalle:
The complete fix is significantly more complex than the original patch. Both Claude and Codex agree on this solution, and both suggested additional tests to look for this issue. The updated PR has tests, which I have not yet run manually.

jamesobutler:
@mhalle Could you take a look at CoPilot's latest review?

PR #9057, opened 2026-03-03
ENH: Add JSON schema for additional extension metadata

Sunderlandkyl:
This commit adds a new metadata field with the new extension JSON schema slicer-extension-catalog-entry-schema-v1.0.2.json.

- dicom_support_rule: Expression to determine DICOM data handling. This string will be evaluated to suggest extensions that can handle specific DICOM modalities

We chose to only update the patch version for this schema to follow previously used patterns in version numbering.


lassoan:
We may consider at some point setting `"additionalProperties": true` to indicate that additional fields may be added to the extension catalog entry files in the future.

PR #9059, opened 2026-03-05
BUG: Fix inability to create multiple scene views

Sunderlandkyl:
After creating the first scene view, the "Capture display nodes" and "Capture view nodes" checkboxes became unchecked for subsequent scene views. As a result, users could not create additional scene views unless they manually re-enabled these options.

The checkboxes became unchecked because UpdateExistingNodesCheckBox was false, which removed already tracked nodes from the list of nodes to capture. However, UpdateExistingNodesCheckBox is not visible when creating a new scene view and should not affect node capture in this case.

Fix: ignore UpdateExistingNodesCheckBox when it is not visible.

### DIFF
--- a/Modules/Loadable/SceneViews/GUI/qSlicerSceneViewsModuleDialog.cxx
+++ b/Modules/Loadable/SceneViews/GUI/qSlicerSceneViewsModuleDialog.cxx
@@ -446,6 +446,10 @@ void qSlicerSceneViewsModuleDialog::loadSceneViewInfo(int index)
   vtkImageData* imageData = this->m_Logic->GetNthSceneViewScreenshot(index);
   this->setImageData(imageData);
 
+  /// Set check states so that updateNodeSelection can update the list of nodes that will be
+  /// included in the scene view.
+  /// Because loadSceneViewInfo is called, we are editing an existing scene.
+  /// Options are left unchecked so that we only modify the metadata and not scene view content by default.
   d->UpdateExistingNodesCheckBox->setVisible(true);
   d->UpdateExistingNodesCheckBox->setChecked(false);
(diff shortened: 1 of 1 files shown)


Modules/Loadable/SceneViews/GUI/qSlicerSceneViewsModuleDialog.cxx
```diff
@@ -415,7 +415,10 @@ void qSlicerSceneViewsModuleDialogPrivate::updateNodeSelection()
   bool wasBlocking = this->NodeSelectorComboBox->blockSignals(true);
   this->updateViewNodesSelection(this->CaptureViewNodesCheckBox->checkState());
   this->updateDisplayNodesSelection(this->CaptureDisplayNodesCheckBox->checkState());
-  this->updateExistingNodesSelection(this->UpdateExistingNodesCheckBox->checkState());
+  if (!this->UpdateExistingNodesCheckBox->isHidden())
```
lassoan: This decision here is not completely trivial. To avoid the need to look up the commit comment related to this change, let's add a short comment to explain.

lassoan (commented):
Thanks, it looks good, I just added one small comment inline.

Modules/Loadable/SceneViews/GUI/qSlicerSceneViewsModuleDialog.cxx
Sunderlandkyl: Added a comment.
```
  // If UpdateExistingNodesCheckBox is hidden then we are creating a new scene view.
  // Don't call updateExistingNodesSelection since it will overwrite the capture state of
  // existing nodes.
```

lassoan (approved):
Looks good, thank you!

PR #9063, opened 2026-03-06
BUG: Fix VTK observer leak in parameterNodeWrapper connectGui/disconnectGui

ebrahimebrahim:
In the parameter node wrapper, calling `disconnectGui()` doesn't remove the VTK `ModifiedEvent` observer that `connectGui()` added. Each connect/disconnect cycle leaks one observer on the underlying MRML node.

(How I ran into this: The stale observers are normally silent, but I have a custom application that pre-creates a lot of module widgets at startup. Clearing the scene during a unit test caused a crash because the stale observers fired while the parameter nodes were being re-initialized.)

The fix stores the observer tag on connect and removes it on disconnect.

### DIFF
--- a/Base/Python/slicer/tests/test_slicer_parameter_node_wrapper_guis.py
+++ b/Base/Python/slicer/tests/test_slicer_parameter_node_wrapper_guis.py
@@ -1171,3 +1171,32 @@ class ParameterNodeWrapperGuiTest(unittest.TestCase):
         param.connectParametersToGui(mapping)
 
         self.impl_parameterPacks_test_connected_parameter_node_wrapper(param, ui)
+
+    def test_disconnectGui_removes_vtk_observer(self):
+        """Verify that disconnectGui removes the ModifiedEvent VTK observer added by connectGui."""
+        @parameterNodeWrapper
+        class ParameterNodeWrapper:
+            alpha: bool
+
+        mappingWidget = qt.QWidget()
+        mappingWidget.setLayout(qt.QVBoxLayout())
+        widgetAlpha = qt.QCheckBox()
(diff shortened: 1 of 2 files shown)


Base/Python/slicer/tests/test_slicer_parameter_node_wrapper_guis.py
```diff
@@ -1171,3 +1171,32 @@ def addWidget(widget, /, packParam):
...
+        mappingWidget = qt.QWidget()
+        mappingWidget.setLayout(qt.QVBoxLayout())
+        widgetAlpha = qt.QCheckBox()
+        widgetAlpha.setProperty(SlicerParameterNamePropertyName, "alpha")
+        mappingWidget.layout().addWidget(widgetAlpha)
+        mappingWidget.deleteLater()
+
+        param = ParameterNodeWrapper(newParameterNode())
+
+        self.assertEqual(len(param._guiVtkObserverTags), 0)
```
allemangD, line 1190: These assertions don't *really* check what we're after, which is that `RemoveObserver` must have been called; however since VTK objects don't really expose anything like "GetNumberOfObservers" I think this is a reasonable proxy.

An alternative *could* be to use `VtkObserverMixin` and peek at the internals of that; but that's an out-of-scope architectural change.

Another alternative could be, somehow, to use `HasObserver`. But, from memory, that doesn't play very nicely with Python functions in the best of times, and in this case the command is a `lambda`. Again, reworking that is probably a more significant architectural change.

ebrahimebrahim: Agreed that this isn't exactly checking what we're after with the fix here, but at least it can catch some ways in which one could break things so I agree it is a reasonable proxy. Thanks for the review!

Issue #9068, opened 2026-03-10
Qt 6 build on mac silicon laptop threeD view stuck in rotation mode


pieper:
## Summary

The build works, as reported in parent issue.  When you start the 3D view tumbles as you move in it and no way out of it.  Apparently the left button is stuck.  This happens on macbook air and may be due to track pad.

## Steps to reproduce

Build and test on mac laptop

## Environment
- Slicer version: Recent git checkout
- Operating system: Mac 
- 

<img width="283" height="482" alt="Image" src="https://github.com/user-attachments/assets/ae69b54f-12c6-4c60-b5bf-22f31afe45e1" />

pieper:
Here's an analysis done by Claude with the slicer-skill

Root Cause Analysis
Bug: On macOS Silicon with Qt6 (6.10.1), the 3D viewer's interactor gets stuck in ROTATE state (as if left mouse button is held down).

Mechanism: The macOS trackpad generates Qt touch events (TouchBegin) alongside regular mouse events. VTK's QVTKInteractorAdapter (line 259-261 in [QVTKInteractorAdapter.cxx](vscode-webview://113go3phljvolkrsfpph4j86iaahm5e3f065k9f7s6v1kvjqpsds/slicer-skill/slicer-dependencies/VTK/GUISupport/Qt/QVTKInteractorAdapter.cxx#L259-L261)) converts TouchBegin into vtkCommand::LeftButtonPressEvent, but the corresponding TouchEnd/LeftButtonReleaseEvent never arrives or arrives out of order.

Evidence from event monitoring:

A single trackpad click generated 5 LeftButtonPressEvent but only 2 LeftButtonReleaseEvent at the VTK level
Qt's TouchBegin event was observed firing after MouseButtonRelease with no corresponding TouchEnd
After disabling WA_AcceptTouchEvents, the state correctly stays at 0
The event flow causing the bug:

User clicks trackpad → Qt generates MouseButtonPress + TouchBegin
Both get forwarded through QVTKOpenGLNativeWidget::event() → QVTKInteractorAdapter::ProcessEvent()
Mouse press → LeftButtonPressEvent (state=ROTATE)
Mouse release → LeftButtonReleaseEvent (state=START)
TouchBegin (arrives late) → LeftButtonPressEvent (state=ROTATE again!)
No TouchEnd arrives → stuck in ROTATE forever
Fix options:

[...]

PR #9069, opened 2026-03-10
BUG: Disable touch events on macOS Qt6 to fix stuck 3D rotation

pieper:
On macOS with Qt6, trackpad touch events (TouchBegin) are converted by VTK's QVTKInteractorAdapter into LeftButtonPressEvent without matching LeftButtonReleaseEvent from TouchEnd, leaving the interactor style permanently stuck in ROTATE state.

Disable WA_AcceptTouchEvents on the VTK widget for both 3D and slice views on macOS Qt6 builds to prevent this.

Fixes https://github.com/Slicer/Slicer/issues/9068

### DIFF
--- a/Libs/MRML/Widgets/qMRMLSliceView.cxx
+++ b/Libs/MRML/Widgets/qMRMLSliceView.cxx
@@ -62,6 +62,7 @@
 #include <vtkRenderWindowInteractor.h>
 #include <vtkInteractorStyleUser.h>
 #include <vtkSmartPointer.h>
+#include <vtkVersionMacros.h>
 
 //--------------------------------------------------------------------------
 // qMRMLSliceViewPrivate methods
@@ -96,6 +97,16 @@ void qMRMLSliceViewPrivate::init()
   Q_Q(qMRMLSliceView);
 
   this->ctkVTKSliceViewPrivate::init();
+
+  // Disable touch event processing on macOS with Qt6 to prevent spurious
+  // LeftButtonPressEvent from trackpad touch events without matching releases.
+  // See https://github.com/Slicer/Slicer/issues/9068
+#if defined(Q_OS_MACOS) && (QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)) && (VTK_VERSION_NUMBER >= VTK_VERSION_CHECK(9, 5, 0))
+  if (q->VTKWidget() != nullptr)
+  {
+    q->VTKWidget()->setEnableTouchEventProcessing(false);
+  }
+#endif
 
   q->setRenderEnabled(this->MRMLScene != nullptr);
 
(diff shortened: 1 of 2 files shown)


pieper:
Note that Claude says:

> The underlying bug is in VTK's QVTKInteractorAdapter which should ideally be fixed upstream to properly handle touch/mouse event coalescing on macOS Qt6, but this Slicer-side workaround is the quickest path to a working build.

But this fix works and slicer is basically unusable without it.

pieper:
Yes, this is cleaner, thanks @jamesobutler 👍 

We may need to revisit this in the future as I saw somewhere that touchscreen mac laptops are coming.

Libs/MRML/Widgets/qMRMLSliceView.cxx
```diff
@@ -97,6 +97,16 @@ void qMRMLSliceViewPrivate::init()
 
   this->ctkVTKSliceViewPrivate::init();
 
+  // Disable touch event processing on macOS with Qt6 to prevent spurious
+  // LeftButtonPressEvent from trackpad touch events without matching releases.
+  // See https://github.com/Slicer/Slicer/issues/9068
+#if defined(Q_OS_MACOS) && (QT_VERSION >= QT_VERSION_CHECK(6, 0, 0))
+  if (q->VTKWidget() != nullptr)
+  {
+    q->VTKWidget()->setEnableTouchEventProcessing(false);
```
jamesobutler: With https://gitlab.kitware.com/vtk/vtk/-/commit/754d6415de48e1829ad7c87ea76aebf9643fa5f3 only being present in VTK 9.5+ and since Slicer currently supports building with VTK 9.4
https://github.com/Slicer/Slicer/blob/16069d367eb549703006a8e709aadba8d7ad1c6a/SuperBuild/External_VTK.cmake#L143-L152
https://github.com/Slicer/Slicer/blob/16069d367eb549703006a8e709aadba8d7ad1c6a/CMakeLists.txt#L448-L452
do you think we should include the following guard check
```
#if (VTK_MAJOR_VERSION >= 9 && VTK_MINOR_VERSION >= 5
```
or make a change to remove VTK 9.4 support? VTK 9.6 support is coming soon through #8928 which will mean still supporting multiple VTK versions for folks

pieper:
The codespell issues are unrelated.  We can fix those in a different PR.

jamesobutler:
@pieper The codespell issues were already resolved in `main`. Looks like you had a [force push](https://github.com/Slicer/Slicer/compare/a0f177a8ea32e3e4e712fbb8abef80dba40e047a..50b96251f56fd947940560b0211b280e39e8771f) that rolled back about 30 some commits from `main` and now have pulled in a merge commit into the branch.

pieper:
Oops, thanks for the careful review!  Let me try again.

lassoan:
> We may need to revisit this in the future as I saw somewhere that touchscreen mac laptops are coming.

The touch gestures have been already quite useful (pinch, zoom, etc.), so we should restore this. I added an issue to track this: https://github.com/Slicer/Slicer/issues/9071

Issue #9077, opened 2026-03-18
ScreenCapture module does not save an output file in Single Frame mode.


mikebind:
## Summary

ScreenCapture does not save an output file in Single Frame mode. 

## Steps to reproduce the issue

Open ScreenCapture module.  Select "single frame" as the Capture mode.  Choose an output directory and output file name (such as Capture.png).  Click the "Capture" button. 
Expected behavior: Image file Capture.png is created in the specified folder.
Actual behavior: Nothing appears in the specified folder.  Error log indicates that a temporary file has been written and then deleted. 

Looking at the ScreenCapture code, I think there should be a relatively simple fix, I'll try to put together a PR and reference this issue. 

## Environment
- Slicer version: Slicer-5.10.0
- Operating system: Windows 11

PR #9078, opened 2026-03-19
BUG: Fix ScreenCapture single frame saving

mikebind:
Summary
ScreenCapture does not save an output file in Single Frame mode.

Steps to reproduce the issue
Open ScreenCapture module. Select "single frame" as the Capture mode. Choose an output directory and output file name (such as Capture.png). Click the "Capture" button.
Expected behavior: Image file Capture.png is created in the specified folder.
Actual behavior: Nothing appears in the specified folder. Error log indicates that a temporary file has been written and then deleted.

This PR addresses this issue while attempting to minimally perturb the existing code. Tested manually and passes automatic tests. 

[Issue 9077](https://github.com/Slicer/Slicer/issues/9077)

### DIFF
--- a/Modules/Scripted/ScreenCapture/ScreenCapture.py
+++ b/Modules/Scripted/ScreenCapture/ScreenCapture.py
@@ -213,6 +213,10 @@ class ScreenCaptureWidget(ScriptedLoadableModuleWidget):
             _("String that defines file name, type, and numbering scheme. Default: image%05d.png."))
         self.fileNamePatternWidget.text = _("image_%05d.png")
 
+        self.singleFrameFileNameWidget = qt.QLineEdit()
+        self.singleFrameFileNameWidget.setToolTip(_("String that defines file name and type for single image capture mode."))
+        self.singleFrameFileNameWidget.text = _("SlicerCaptureSingleFrame.png")
+
         self.videoFileNameWidget = qt.QLineEdit()
         self.videoFileNameWidget.setToolTip(_("String that defines file name and type."))
         self.videoFileNameWidget.text = _("SlicerCapture.mp4")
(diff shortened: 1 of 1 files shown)


lassoan:
The screen capture saving works well without this change. Probably the source of confusion is that you expected that a fixed filename would work. The filename currently requires a placeholder for a counter value, as the user is not expected to type a new filename for each screenshot. Please use `image_%05d.png` as output file name and see if it works for you.

That said, the fact that you had trouble figuring this out means that the implementation is not optimal. We could add a "Reset" button that would set the filename to `image_%05d.png`, add placeholder automatically if the current name does not have one, we could enable filenames without placeholders, rename `Output file name` to `Output file pattern`, etc.

mikebind:
Thanks for the clarification.  I do find the current behavior confusing and also unhelpful.  If I am taking a single capture, it is usually to capture something specific, and I want to name the file in such a way that it records the intention of the capture.  "image_00001.png" does not do that. And, the failure mode of just not saving anything if the name does not contain a numbering placeholder is also confusing.  If that is a requirement, it makes sense to either warn the user, show an error message, or fix it for them.  Of those, an explanatory error message with a suggested correction makes the most sense to me.  As it is, it is not clear to me as a user why a numbering placeholder should be required in a filename when I have selected "single frame" as an output mode.

lassoan:
> it is not clear to me as a user why a numbering placeholder should be required

Certainly not a requirement but essential for creating many screenshots without interruption. I usually take many candidate screenshots in slightly different orientations, visualization settings, etc. and a unique filename is needed for them. It is later easy to pick the screenshots that I want to use and copy/rename them.

The question is how to make things simple and convenient. When the provided filename does not exist then it should be no problem to create a new file by that name. This should fulfill your use case when you manually type a new name for every screenshot. Do you have any further needs or preferences? Maybe when the file already exist then we could offer the user to overwrite, rename, or "auto-rename" (i.e., insert a counter field into the filename pattern).

mikebind:
> Maybe when the file already exist then we could offer the user to overwrite, rename, or "auto-rename" (i.e., insert a counter field into the filename pattern).

I think this sounds like a great plan.  Thanks for thinking about this.

PR #9081, opened 2026-03-21
ENH: Add drag-and-drop segment reordering in segments table

vacaslistradas:
## Summary
Adds drag-and-drop segment reordering to the segments table so segments can be repositioned directly in Segmentations and Segment Editor instead of repeatedly clicking Up/Down.

## Changes
- add custom drag/drop handling in `qMRMLSegmentsTableView`
- expose `moveSegmentsToRow(...)` and `moveSelectedSegmentsToRow(...)`
- mark segment rows draggable and droppable in `qMRMLSegmentsModel`
- add C++ regression coverage for reorder behavior and save/reload persistence
- add Python smoke coverage for programmatic row moves

## Validation
- manual testing on macOS arm64 in Segmentations and Segment Editor
- `qMRMLSegmentsTableViewTest`: 4 passed, 0 failed
- reordered segment order persists after saving and reloading `.seg.nrrd`
- filtered-row behavior matches the existing Up/Down semantics: reordering is relative to visible rows while updating the true underlying segmentation order


vacaslistradas:
Demo of drag-and-drop segment reordering in the segments table.

Local validation for this PR:
- manual testing on macOS arm64 in Segmentations and Segment Editor
- `qMRMLSegmentsTableViewTest`: 4 passed, 0 failed
- reordered segment order persists after saving and reloading `.seg.nrrd`
- filtered-row behavior matches the existing Up/Down semantics: reordering is relative to visible rows while updating the true underlying segmentation order

I will attach a short GIF demo below.

vacaslistradas:
slicer-pr-demo

cpinter:
Thanks for working on this! How do the segments look in the Data module after doing this?

vacaslistradas:
Sorry I didn't think about it at first, but I just checked this locally: after reordering segments in the segments table, the segment items under the segmentation in the Data module get reordered to match the new segmentation order as well. I also verified that after saving and reloading the segmentation, the Data-module order stays consistent with the reordered segment order.

cpinter:
Thanks a lot! In this case it sounds good to me. The code looks nice and clean as well. Let's wait for copilot's review, and probably we want to hear the opinion of @lassoan as well

vacaslistradas:
I addressed the unused-variable comment in fcd9f4c912. For the DragEnter/DragMove handling, I kept that in the custom reorder path, because when I let the default QAbstractItemView handling run during local testing, I was seeing intermittent issues like rows disappearing from the table and the drop state getting confused. Since the actual segment reorder happens explicitly on Drop, I went with the custom handling for reliability, even though that means losing the default drag indicator. Happy to revisit it if you think that tradeoff is not worth it.

cpinter:
Perfect! Please squash the four commits into one. Thank you!

vacaslistradas:
Just squashed, thanks!

Modules/Loadable/Segmentations/Widgets/qMRMLSegmentsTableView.cxx
```diff
@@ -65,21 +65,62 @@
...
 #define VISIBILITY_PROPERTY "Visible"
 #define STATUS_PROPERTY "Status"
 
+namespace
+{
+//------------------------------------------------------------------------------
+const char* SEGMENTS_TABLE_REORDER_MIME_TYPE = "application/x-slicer-segments-table-reorder";
+
+//------------------------------------------------------------------------------
+int segmentDropRowForPosition(QTableView* table, const QPoint& viewportPosition)
```
cpinter: Please move these two functions in the private class. I see no reason to introduce this new namespace

Modules/Loadable/Segmentations/Widgets/qMRMLSegmentsTableView.cxx
```diff
@@ -972,6 +1025,124 @@ bool qMRMLSegmentsTableView::eventFilter(QObject* target, QEvent* event)
...
+    }
+
+    QModelIndex pressedIndex = d->SegmentsTable->indexAt(mouseEvent->pos());
+    d->DragStartPositionValid = pressedIndex.isValid();
+    d->PressedSegmentID = d->DragStartPositionValid ? d->SortFilterModel->segmentIDFromIndex(pressedIndex) : QString();
+    d->DragStartPosition = mouseEvent->pos();
+  }
+  else if (target == d->SegmentsTable->viewport() && event->type() == QEvent::MouseMove)
+  {
+    QMouseEvent* mouseEvent = static_cast<QMouseEvent*>(event);
```
cpinter, line 1308: There are no comments at all in the code. It would be useful to summarize some parts, like here. Humans still read code...

Modules/Loadable/Segmentations/Widgets/qMRMLSegmentsTableView.cxx
```diff
@@ -1534,6 +1707,112 @@ void qMRMLSegmentsTableView::jumpSlices()
   }
 }
 
+//------------------------------------------------------------------------------
+bool qMRMLSegmentsTableView::moveSegmentsToRow(QStringList segmentIDs, int row)
```
cpinter, line 1932: This method seems unnecessarily long for this operation (especially that you just need to call the reorder function in the end). Please try to refactor, or if it is not possible, add comments (but please add comments everywhere, I don't really see any)

cpinter (changes requested):
Sorry I went on vacation then lost sight of this PR. I added some more comments. Please amend the one commit instead of adding a new one. Thank you!

finetjul:
It is to be noted that the order of the segments in the list is currently not related to the order of "rendering" of the segments.
The order of rendering can have an impact when segments overlap.

vacaslistradas:
Thanks, I amended the commit and pushed the requested updates.

Also, to clarify @finetjul's point: I only intended to reorder the segment list / underlying segmentation order, matching the existing Up/Down behavior. This PR does not attempt to change rendering/compositing order for overlapping segments.

cpinter:
There are unresolved conversations. Please review those and make sure you resolve them in some way. If you decide there is nothing to do, explain why. Thanks!

vacaslistradas:
I resolved the outstanding conversations; I believe everything has been addressed in the current commit. The rendering-order point is intentionally out of scope.

cpinter:
Sorry for the delay. Not forgotten just crazy times... If someone else can also review it in the meantime, also great

PR #9107, opened 2026-04-10
ENH: Enable Zlib support in DCMTK build

fedorov:
Lack of ZLIB support in DCMTK makes it impossible to even import DICOM files that use Deflated transfer syntax.

Related to https://github.com/commontk/CTK/issues/25#issuecomment-298687315

I built Slicer with this fix on Linux and confirmed there were no build errors, and I was able to import a Deflated transfer encoded segmentation.

Of note, Zlib has been enabled in DCMTK as part of dcmqi superbuild and there were no issues across platforms, see https://github.com/QIICR/dcmqi/blame/master/CMakeExternals/DCMTK.cmake#L83.


lassoan (commented):
Thank you, this should work! zlib was already set as a dependency of DCMTK. The only thing is missing here that we need to pass [ZLIB_ROOT](https://github.com/Kitware/CMake/blob/3b5342b305d81f75406cf82da0c63f9361b33e02/Modules/FindZLIB.cmake#L53) to DCMTK like this:

https://github.com/Slicer/Slicer/blob/925ccc7513a258792575dbff9fa2aae7da2fa330/SuperBuild/External_LibArchive.cmake#L90

Without that, [`find_package(ZLIB QUIET)` in DCMTK](https://github.com/DCMTK/dcmtk/blob/969c4b6f20e73e26603856bf3a381a3823314422/CMake/3rdparty.cmake#L113) could find some random zlib on the system.

fedorov:
@lassoan `-DZLIB_ROOT:PATH=${ZLIB_ROOT}` added, and build tested on linux!

pieper (approved):
LGTM 👍

Issue #8387, opened 2025-04-19
Add support for labelmap dicom seg


pieper:
## Is your feature request related to a problem? Please describe.

New labelmap DICOM SEG is now available in the standard and the Python highdicom library already support reading and writing so we should support it in Slicer.

Note that the original SEG in is 86MB while the new compressed labelmap version is 1.7MB.

## Describe the solution you'd like

We should make a new DICOM Plugin for Segmentations using highdicom, which is pure python and easy to install in Slicer.  We should make this part of the core so that no extensions are required.

While we are doing this, we should consider switching to highdicom for SR support in the core, and also use it for many other types of dicom objects (presentation states, parametric maps, etc).

## Describe alternatives you've considered

We currently use DCMQI (C++, based on DCMTK) in the Quantitative Reporting extension, so we could add labelmap support there instead.  I've discussed this with @fedorov and there is not a timeline for support of labelmaps in DCMQI so using the existing highdicom code makes sense.

## Additional context

Reference examples of labelmap segs along with example code is available here:

https://zenodo.org/records/15243636

pieper:
Note for implementation: We may be able to use https://highdicom.readthedocs.io/en/latest/volume.html

fedorov:
> We currently use DCMQI (C++, based on DCMTK) in the Quantitative Reporting extension, so we could add labelmap support there instead. I've discussed this with @fedorov and there is not a timeline for support of labelmaps in DCMQI so using the existing highdicom code makes sense.

This is not entirely correct. @michaelonken and @spalte started working on adding LABELMAP support to dcmqi in https://github.com/QIICR/dcmqi/pull/491, but were blocked for a very long time by inability to update DCMTK. The latter issue is now resolved, and @michaelonken will continue that work under IDC, we are already in the process of setting up the arrangements to cover that work.

pieper:
@fedorov I was trying to document exactly what you told me when I explicitly asked you!  Now apparently you are saying the exact opposite : )

fedorov:
I am sorry, maybe I was distracted or misunderstood or forgot. In any case, no harm done, I hope. Just trying to set the record straight.

fedorov:
Update: as of May 29, 2025, @michaelonken is all set to start working on adding support of LABELMAP to dcmqi!

fedorov:
@michaelonken is making great progress with the implementation, with key features already in DCMTK!

[...]

PR #9114, opened 2026-04-16
Add DICOM SEG/PM/SR/M3D plugins to Slicer core

fedorov:
This follows up on the discussion in https://github.com/Slicer/Slicer/issues/8387#issuecomment-4230140358. So far I built and tested this branch on Linux, and confirmed SEG read, SEG write, SR TID1500 read, M3D read all work as expected.

### Overview

This branch migrates four DICOM object-type plugins from the [QuantitativeReporting](https://github.com/QIICR/QuantitativeReporting) extension into Slicer's core `DICOMPlugins` module, so that users no longer need to install a separate extension to load these common DICOM object types. The four plugins added are:

| Plugin | DICOM SOP class |
|---|---|
| `DICOMSegmentationPlugin` | SEG — DICOM Segmentation Storage |
| `DICOMParametricMapPlugin` | PM — Parametric Map Storage |
| `DICOMTID1500Plugin` | SR — Comprehensive/Enhanced SR (TID 1500 Measurement Report) |
| `DICOMM3DPlugin` | M3D — Encapsulated 3D Model (STL) |

---

#### New plugins (`Modules/Scripted/DICOMPlugins/`)

- **[DICOMSegmentationPlugin.py](Modules/Scripted/DICOMPlugins/DICOMSegmentationPlugin.py)** — Loads DICOM SEG objects. Uses the `dcmqi` pip binary `segimage2itkimage` to convert SEG → ITK volumes, then imports them as `vtkMRMLSegmentationNode`.
- **[DICOMParametricMapPlugin.py](Modules/Scripted/DICOMPlugins/DICOMParametricMapPlugin.py)** — Loads DICOM Parametric Map objects using `dcmqi` (`paramap2itkimage`) into scalar volumes.
- **[DICOMTID1500Plugin.py](Modules/Scripted/DICOMPlugins/DICOMTID1500Plugin.py)** — Parses TID 1500 Imaging Measurement Reports using `highdicom` and populates a `vtkMRMLTableNode` with the extracted measurements.
- **[DICOMM3DPlugin.py](Modules/Scripted/DICOMPlugins/DICOMM3DPlugin.py)** — Extracts the encapsulated STL payload from a DICOM M3D object and loads it as a `vtkMRMLSegmentationNode`.
- **[CMakeLists.txt](Modules/Scripted/DICOMPlugins/CMakeLists.txt)** — Registers the four new plugin scripts so they are installed.

#### Base class enhancements (`Modules/Scripted/DICOMLib/DICOMPlugin.py`)

- Added common DICOM tag shortcuts (`seriesInstanceUID`, `modality`, `instanceUID`, `classUID`).
- Added `currentDateTime` property (lazy, cached timestamp for temp directory naming).
- Added `cleanup()` method to remove the plugin's `tempDir`.
- Added `addReferences()` / `_addReferencedSeries()` / `_addReferencedImages()` helpers to populate `loadable.referencedInstanceUIDs`.
- Added `_dcmqi_binary()` static method to locate dcmqi console-script binaries via `shutil.which` with a fallback to the Python `bin/` directory.
- Refactored `examineForImport()` into a caching wrapper that calls a new virtual method `examineFiles()`, reducing boilerplate in subclasses.

#### Dependency changes (`SuperBuild/External_python-dicom-requirements.cmake`)

- Added `dcmqi==0.3.3` with per-platform SHA-256 hashes (macOS arm64/x86_64, Linux manylinux2014, Windows 32/64).
- Added `highdicom==0.27.0` with SHA-256 hash.

#### Extension registry (`Modules/Scripted/DICOM/DICOMExtensions.json`)

- Removed the three QuantitativeReporting entries for SEG, PM, and SR — these object types are now handled natively and no longer prompt the user to install the extension.

#### Integration tests (`Applications/SlicerApp/Testing/Python/DICOMPluginTests.py`)

- Added `DICOMPluginTestsTest` with three end-to-end tests:
  - `test_DICOMSegmentationPlugin` — downloads a ReMIND SEG file from IDC, imports it, and asserts a `vtkMRMLSegmentationNode` with ≥1 segment is created.
  - `test_DICOMTID1500Plugin` — downloads a QIN-PROSTATE SR file from IDC, imports it, and asserts ≥1 `vtkMRMLTableNode` is created.
  - `test_DICOMM3DPlugin` — downloads a Prostate-MRI-US-Biopsy M3D file from IDC, imports it, and asserts a `vtkMRMLSegmentationNode` is created.
- Registered the test module in `Applications/SlicerApp/Testing/Python/CMakeLists.txt`.

---

### Notes for reviewers

- The `dcmqi` pip package ships pre-built platform binaries (macOS 13+, Linux manylinux_2_28 x86_64, Windows). Older macOS versions and non-x86_64 Linux builds will not be able to install it — this is documented in inline comments in the cmake file.
- Integration tests download files from the public IDC GCS bucket at test time; they require network access and are not hermetic.

### DIFF
--- a/Applications/SlicerApp/Testing/Python/DICOMPluginTests.py
+++ b/Applications/SlicerApp/Testing/Python/DICOMPluginTests.py
@@ -1,4 +1,3 @@
-import logging
 import os
 import shutil
 import tempfile
--- a/Applications/SlicerApp/Testing/Python/DICOMPluginTests.py
+++ b/Applications/SlicerApp/Testing/Python/DICOMPluginTests.py
@@ -1,0 +1,264 @@
+import logging
+import os
+import shutil
+import tempfile
+import urllib.request
+
+import slicer
+from slicer.ScriptedLoadableModule import *
+from DICOMLib import DICOMUtils
+
+
+#
+# DICOMPluginTests
+#
+
+
+class DICOMPluginTests(ScriptedLoadableModule):
+    def __init__(self, parent):
+        ScriptedLoadableModule.__init__(self, parent)
+        parent.title = "DICOMPluginTests"
+        parent.categories = ["DICOMPluginTests"]
+        parent.dependencies = []
+        parent.contributors = ["Andrey Fedorov (SPL, BWH)"]
+        parent.helpText = """
+    Integration tests for DICOM plugins added in Slicer core: DICOMSegmentationPlugin,
+    DICOMTID1500Plugin, and DICOMM3DPlugin. Each test downloads a small (~5-32 KB) reference
+    DICOM file from the NCI Imaging Data Commons (IDC, https://portal.imaging.datacommons.cancer.gov/)
+    public GCS bucket, imports it into a temporary DICOM database, and verifies that the plugin
+    correctly examines and loads the object into the MRML scene.
+    """
+        parent.acknowledgementText = (
+            "Test data sourced from the NCI Imaging Data Commons (IDC). "
+            "SEG and SR samples created by the QIICR project (NIH U24 CA180918). "
+            "M3D sample from the Prostate MRI US Biopsy collection."
+        )
+
+
+#
+# DICOMPluginTestsWidget
+#
+
+
+class DICOMPluginTestsWidget(ScriptedLoadableModuleWidget):
+    def setup(self):
+        ScriptedLoadableModuleWidget.setup(self)
+        self.layout.addStretch(1)
+
+
+#
+# DICOMPluginTestsTest
+#
+
+
+class DICOMPluginTestsTest(ScriptedLoadableModuleTest):
+    """Integration tests for DICOMSegmentationPlugin, DICOMTID1500Plugin, and DICOMM3DPlugin."""
+
+    def setUp(self):
+        self.delayDisplay("Closing the scene")
+        slicer.mrmlScene.Clear(0)
+
+    def runTest(self):
+        self.setUp()
+        self.test_DICOMSegmentationPlugin()
+        self.setUp()
+        self.test_DICOMTID1500Plugin()
+        self.setUp()
+        self.test_DICOMM3DPlugin()
+
+    # -------------------------------------------------------------------------
+    # Helpers
+    # -------------------------------------------------------------------------
+
+    def _downloadDICOMFile(self, url, destDir):
+        """Download a single DICOM file via HTTPS into destDir. Returns the local path."""
+        filename = url.split("/")[-1]
+        destPath = os.path.join(destDir, filename)
+        self.delayDisplay(f"Downloading {filename} ...")
+        urllib.request.urlretrieve(url, destPath)
+        return destPath
+
+    # -------------------------------------------------------------------------
+    # Test: DICOMSegmentationPlugin
+    # -------------------------------------------------------------------------
+
+    def test_DICOMSegmentationPlugin(self):
+        """Test loading a DICOM SEG object via DICOMSegmentationPlugin.
+
+        Test data: ReMIND collection, patient ReMIND-004, tumor segmentation
+        (QIICR/dcmqi-generated, Manufacturer: QIICR).
+        SeriesInstanceUID: 1.3.6.1.4.1.14519.5.2.1.27661008822762431370973483522020876264
+        License: CC BY 4.0
+
+        IDC viewer:
+          https://viewer.imaging.datacommons.cancer.gov/v3/viewer/?StudyInstanceUIDs=
+          1.3.6.1.4.1.14519.5.2.1.21953915370182127679153341817124092346
+          &initialSeriesInstanceUID=
+          1.3.6.1.4.1.14519.5.2.1.27661008822762431370973483522020876264
+
+        Acknowledgment:
+          Juvekar, P., et al. (2023). ReMIND: The Brain Resection Multimodal Imaging Database.
+          The Cancer Imaging Archive. https://doi.org/10.7937/3RAG-D070
+        """
+        self.delayDisplay("Starting test_DICOMSegmentationPlugin")
+
+        SEG_URL = (
+            "https://storage.googleapis.com/idc-open-data/"
+            "8676c8bb-3bdd-4106-be09-97223b5126e7/"
+            "6936cea9-f0ac-49bd-91c4-86904f5c3650.dcm"
+        )
+        SEG_SERIES_UID = "1.3.6.1.4.1.14519.5.2.1.27661008822762431370973483522020876264"
+
+        tmpDir = tempfile.mkdtemp()
+        try:
+            self._downloadDICOMFile(SEG_URL, tmpDir)
+
+            with DICOMUtils.TemporaryDICOMDatabase() as db:
+                self.assertTrue(db.isOpen, "Temporary DICOM database failed to open")
+
+                DICOMUtils.importDicom(tmpDir, db)
+
+                files = slicer.dicomDatabase.filesForSeries(SEG_SERIES_UID)
+                self.assertGreater(len(files), 0, "SEG series was not indexed into the DICOM database")
+
+                plugin = slicer.modules.dicomPlugins["DICOMSegmentationPlugin"]()
+                loadables = plugin.examineFiles(files)
+                self.assertGreater(len(loadables), 0,
+                                   "DICOMSegmentationPlugin.examineFiles() returned no loadables")
+
+                self.assertTrue(plugin.load(loadables[0]),
+                                "DICOMSegmentationPlugin.load() returned False")
+
+                segNodes = slicer.util.getNodesByClass("vtkMRMLSegmentationNode")
+                self.assertEqual(len(segNodes), 1,
+                                 f"Expected 1 vtkMRMLSegmentationNode, got {len(segNodes)}")
+                self.assertGreater(segNodes[0].GetSegmentation().GetNumberOfSegments(), 0,
+                                   "Loaded segmentation node contains no segments")
+
+        finally:
+            shutil.rmtree(tmpDir)
+
+        self.delayDisplay("test_DICOMSegmentationPlugin passed!")
+
+    # -------------------------------------------------------------------------
+    # Test: DICOMTID1500Plugin
+    # -------------------------------------------------------------------------
+
+    def test_DICOMTID1500Plugin(self):
+        """Test loading a DICOM SR TID1500 structured report via DICOMTID1500Plugin.
+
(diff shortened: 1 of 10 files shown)


pieper:
Can you add arm64 support for mac to dcmqi easily?  We will be migrating before too long.

pieper (approved):
LGTM 👍

fedorov:
arm64 should already be supported - see https://github.com/QIICR/dcmqi/releases/tag/v1.5.3, https://github.com/ImagingDataCommons/dcmqi-python-distributions/blob/main/dcmqiUrls.cmake#L8 and https://github.com/Slicer/Slicer/pull/9114/changes#diff-dabae6e8317978cad7cb60e6ce80bbd92a127b01064a2c532d2badc1ca4305a3R82-R93!

You can also confirm on your arm64 system with `pip install dcmqi`, as I've just did.

pieper:
I see, I read this backwards.

> Older macOS versions and non-x86_64 Linux builds will not be able to install it — this is documented in inline comments in the cmake file.

I guess it's not a show-stopper, but I guess it's a consequence of going for the pip route instead of superbuilding.

As an intel mac user who uses dcmqi regularly I am okay with needing to use 5.10 with the extension if I need it.  I plan to upgrade to arm before long and I suppose most people will too.

fedorov:
macOS Intel package is also available, see https://github.com/QIICR/dcmqi/releases/tag/v1.5.3.

My interpretation of that comment is that binaries built using GitHub runners for the specific macOS versions may not be compatible with the older macOS versions. dcmqi GitHub action macOS workflow uses macos-15/macos-15-intel runners: https://github.com/QIICR/dcmqi/blob/master/.github/workflows/cmake-macos.yml#L31-L36.

I confess I didn't double check that claim, but it seems to be consistent with what this article says: https://setapp.com/how-to/macos-compatibility-guide.

I do not think it is a big concern if this won't work on macOS older than 2018, is it?

lassoan:
> I do not think it is a big concern if this won't work on macOS older than 2018, is it?

It should be fine. We only guarantee compatibility with hardware released not more than 5 years ago.

Applications/SlicerApp/Testing/Python/DICOMPluginTests.py
```diff
@@ -0,0 +1,263 @@
...
+        SEG_URL = (
+            "https://storage.googleapis.com/idc-open-data/"
+            "8676c8bb-3bdd-4106-be09-97223b5126e7/"
+            "6936cea9-f0ac-49bd-91c4-86904f5c3650.dcm"
+        )
+        SEG_SERIES_UID = "1.3.6.1.4.1.14519.5.2.1.27661008822762431370973483522020876264"
+
+        tmpDir = tempfile.mkdtemp()
+        try:
+            self._downloadDICOMFile(SEG_URL, tmpDir)
```
lassoan, line 113: We should not download a file each time we run the test. Instead, specify the SHA256 of the file here in the script and if we find the file cached already then do not download it again. SampleData module has helper functions for this that can check if the filename is in the cache, hash matches, if yes then reuse if not then download from URL and cache the file.

Applications/SlicerApp/Testing/Python/DICOMPluginTests.py
```diff
@@ -0,0 +1,263 @@
...
+          1.3.6.1.4.1.14519.5.2.1.27661008822762431370973483522020876264
+
+        Acknowledgment:
+          Juvekar, P., et al. (2023). ReMIND: The Brain Resection Multimodal Imaging Database.
+          The Cancer Imaging Archive. https://doi.org/10.7937/3RAG-D070
+        """
+        self.delayDisplay("Starting test_DICOMSegmentationPlugin")
+
+        SEG_URL = (
+            "https://storage.googleapis.com/idc-open-data/"
```
lassoan, line 105: We already send many requests to github=Microsoft while building Slicer, but with this we add Google as well. Does anyone have any concern about this? (I don't mind)

Modules/Scripted/DICOMLib/DICOMPlugin.py
```diff
@@ -89,6 +93,79 @@ def __init__(self):
         self.tags["seriesDescription"] = "0008,103E"
         self.tags["seriesNumber"] = "0020,0011"
         self.tags["frameOfReferenceUID"] = "0020,0052"
+        self.tags["seriesInstanceUID"] = "0020,000E"
+        self.tags["modality"] = "0008,0060"
+        self.tags["instanceUID"] = "0008,0018"
+        self.tags["classUID"] = "0008,0016"
```
lassoan, line 99: Let's use `sopClassUID` for this like (almost) everywhere else.

Modules/Scripted/DICOMLib/DICOMPlugin.py
```diff
@@ -89,6 +93,79 @@ def __init__(self):
...
+
+    @staticmethod
+    def _dcmqi_binary(name):
+        """Return the path to a dcmqi binary installed via the pip 'dcmqi' package.
+
+        Slicer adds its bin directory to PATH, so console_scripts installed in
+        Slicer's embedded Python environment are discoverable via shutil.which().
+        Raises RuntimeError if the binary cannot be found.
+        """
+        path = shutil.which(name)
```
lassoan, line 155: This may find binaries anywhere. Could we search more specifically in the Python scripts folder (or somewhere similar, within the Slicer installation folder) to avoid finding and launching some random binaries on the system?

lassoan (commented):
Thank you, this will be great, I just added a few small comments inline.

Applications/SlicerApp/Testing/Python/DICOMPluginTests.py
pieper: I don't mind.

jamesobutler:
Is the approach of this PR of integrating functionality into Slicer core abandoning the motivation around the work implemented in https://github.com/Slicer/Slicer/commit/44117e93644dfbb0df7d843a812d18a4a536499b which was about informing users which Extensions should be installed for certain DICOM files attempting to be loaded?  Will the newly added `DICOM_SUPPORT_RULE` extension description be less likely to be used now?

lassoan:
> Is the approach of this PR of integrating functionality into Slicer core abandoning the motivation around the work implemented in https://github.com/Slicer/Slicer/commit/44117e93644dfbb0df7d843a812d18a4a536499b which was about informing users which Extensions should be installed for certain DICOM files attempting to be loaded?

There are several other extensions that provide DICOM plugins that will not be integrated into Slicer core (e.g., SlicerRT, SlicerHeart, ...), so the extension recommendation mechanism is still needed.

fedorov:
@lassoan thanks for the review, I updated, and will now do another test on linux.

@jamesobutler definitely we need the extension recommendation infrastructure! This PR covers just one extension.

Applications/SlicerApp/Testing/Python/DICOMPluginTests.py
fedorov: Updated to use SampleData logic.

Modules/Scripted/DICOMLib/DICOMPlugin.py
fedorov: Done. Note that sequences plugin uses `classUID` - I left it in place since it is outside of the scope of this PR.

Modules/Scripted/DICOMLib/DICOMPlugin.py
fedorov: Updated to use `sysconfig.get_path("scripts")`.

Applications/SlicerApp/Testing/Python/DICOMPluginTests.py
fedorov: I can't think of any. GCS S3 buckets are designed to handle many requests.

fedorov:
I should also mention that it should be definitely possible to re-implement these using highdicom. But I would prefer to 1) not deprecate dcmqi so that C++ pathway is maintained and available; 2) not do this as part of this migration, 3) I don't think it is a priority at the moment. In the future we might add pure highdicom-based plugin(s), and adding highdicom as a dependency to the core is a step in that direction.

fedorov:
@lassoan @pieper what are the next steps for this PR? Is there anything I can do?

pieper (approved):
I think this is low risk since these have all been tested as extensions for years.

fedorov:
Thanks Steve! Let's see how it goes...

SuperBuild/External_python-dicom-requirements.cmake
```diff
@@ -75,6 +76,28 @@ if(NOT Slicer_USE_SYSTEM_${proj})
...
+               --hash=sha256:cb007a8991928ff09c864d4c468255f460f6457629a449ad3d52390bf11489f3 \
+               --hash=sha256:5a689b17c1163094a73468d8a624611d460ad86969d24ffd22aeccfee7cb4dc1 \
+               --hash=sha256:daf19ba9326574eb5df6bed91598e6ee55852637c277e37ff29231936526d5ab \
+               --hash=sha256:ca684f09857323389355196cba3c58fce30f4581f1d5220725e2ab08596f6e0f
+  # [/dcmqi]
+  # [highdicom]
+  # Pure-Python package for creating/reading complex DICOM objects (SR, SEG, PM, etc.)
+  # Hashes correspond to the following packages:
+  #  - highdicom-0.27.0-py3-none-any.whl
+  highdicom==0.27.0 --hash=sha256:90a14ef75f67b7df2c3fd83df9c0bddd787dfdfab2b08c8c68828207389aed3f
```
NogginBops, line 99: This broke the build as `highdicom` doesn't specify hashes for their dependenices.
Here is the build log:
```
...
Collecting pyjpegls>=1.0.0 (from highdicom==0.27.0->-r C:/.../python-dicom-requirements-requirements.txt (line 56))
ERROR: In --require-hashes mode, all requirements must have their versions pinned with ==. These do not:
    pyjpegls>=1.0.0 from https://files.pythonhosted.org/packages/50/32/b011c2c33850c8d6f7815cf30257e0e00d9fcb70e0a8052448257636bf19/pyjpegls-1.5.1-cp312-cp312-win_amd64.whl (from highdicom==0.27.0->-r C:/.../python-dicom-requirements-requirements.txt (line 56))
```

fedorov:
@pieper @lassoan I tried a fresh build, and I could reproduce the dashboard error with hashes. No idea what happened when I tested earlier, I was pretty sure I did a clean build. I will try to get it fixed today.

lassoan:
This is great, then the behavior is predictable, so if it works on your computer then it'll likely work on the factory as well. I've started a clean build, too, on Windows - it is still in progress.

PR #9115, opened 2026-04-16
ENH: Add auto-trim of file cache at application startup

lassoan:
Cache folder increased without any limits, despite maximum cache size was specified in application settings. Implemented automatic trimming of cache (deleting the oldest files until the cache size is below the specified limit) and refactored the vtkCacheManager to make it safer and simpler.

- Cache is trimmed at application startup (can be disabled in application settings) by deleting oldest files (by modification time) until cache size is within the specified limit. Default cache limit increased from 200 MB to 1000 MB.
- Sentinel file mechanism: since now the class can delete files, a new safety feature was introduced. Files are only allowed to be deleted if the cache manager owns that folder. Owning is detected by presence of a sentinel file (.slicer-cache), which is only placed automatically in an empty folder or on explicit user approval. Path traversal is also guarded via CanDeletePathInCache() which resolves canonical paths and verifies the target is within the cache root.
- The old "free buffer" concept (keep N MB free) is removed. To keep some space on disk free, it is simpler - and has the same effect - if cache size is reduced by that much. GetFreeCacheSpaceRemaining() is now just RemoteCacheLimit - CurrentCacheSize, and cache-full checks in callers use <= 0.
- Simplified update of cache information. Now a single UpdateCacheInformation() call updates all information about the current state of the cache.
- Settings GUI updates: "Cache free buffer" spinbox removed (concept eliminated). "Auto-prune" checkbox added - persisted in application settings as Cache/AutoPrune setting. "Prune cache" button added to manually trims cache to size limit. "Refresh" button added to updates the used/free size display and file list. setCachePath() now validates the target folder: warns the user if the selected folder contains files but has no sentinel. If confirmed, creates the sentinel. Reverts to previous path if sentinel creation fails. clearCache() now calls updateFromCacheManager() — the old code had this commented out, so the UI was stale after clearing.
- Memory leak fixes: GetFilenameFromURI, EncodeURI, FindCachedFile, and AddCachePathToFilename previously returned new char[]-allocated strings that callers never freed. These all now return std::string. This is the most impactful correctness fix.

image

### DIFF
--- a/Base/Logic/vtkDataIOManagerLogic.cxx
+++ b/Base/Logic/vtkDataIOManagerLogic.cxx
@@ -162,7 +162,6 @@ void vtkDataIOManagerLogic::DeleteDataTransferFromCache(vtkDataTransfer* dt)
         {
           cm->DeleteFromCache(dt->GetDestinationURI());
           dt->SetTransferCached(0);
-          cm->InvokeEvent(vtkCacheManager::CacheDeleteEvent);
         }
       }
     }
@@ -272,7 +271,7 @@ int vtkDataIOManagerLogic::QueueRead(vtkMRMLNode* node)
   }
 
   const char* source = dnode->GetNthStorageNode(storageNodeIndex)->GetURI();
-  const char* dest = cm->GetFilenameFromURI(source);
+  std::string dest = cm->GetFilenameFromURI(source);
   vtkDebugMacro("QueueRead: got the source " << source << " and dest " << dest);
 
   //--- set the destination filename in the node.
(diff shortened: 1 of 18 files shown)


lassoan:
@pieper could please have a look?

jamesobutler:
@lassoan This has resulted in new test failure https://slicer.cdash.org/tests/32989348 due to the presence of a new warning:

`vtkCacheManager.cxx:408   WARN| vtkCacheManager (000001E6521D3FB0): Cache directory does not contain sentinel file .slicer-cache. Destructive cache operations are disabled until the sentinel is created.`

lassoan:
@jamesobutler thanks for the note. The cache manager received a non-empty folder as remote cache folder, so it was valid to log a warning. I've pushed a fix to https://github.com/Slicer/Slicer/pull/9159

PR #9122, opened 2026-04-21
ENH: WebServer socket-takeover hook + threeD camera controls

pieper:
Adds a small, backward-compatible hook that lets a request handler take ownership of the underlying TCP socket after sending initial headers, enabling Server-Sent Events and (later) WebSocket support without touching the per-request write/close lifecycle.  Also fills in a few small gaps in the existing handlers.

BaseRequestHandler
  * New KEEP_OPEN sentinel: a handler returns (contentType, KEEP_OPEN) from handleRequest to signal that it wants to stream to the client rather than emit a single Content-Length-bounded response.
  * New onConnectionTakeover(socket, writeNotifier, readNotifier) default method.  Called after KEEP_OPEN; the handler then owns the socket.  Default implementation disables the notifiers and closes the socket so a buggy handler does not leak a half-open connection.

SlicerHTTPServer.SlicerRequestCommunicator
  * When handleRequest returns KEEP_OPEN, write an HTTP status line plus the caller-supplied Content-Type (no Content-Length), hand the socket and notifiers to the handler, and stop managing the connection.

SlicerRequestHandler.threeD
  * orbitX, orbitY, panX, panY, roll, and a new dolly query parameter now actually apply to the 3D camera instead of being parsed and ignored.  Sign conventions chosen to match Slicer's native 3D drag behaviour.
  * Fixed an AttributeError path: qMRMLThreeDView has no .renderer() accessor -- go through renderWindow().GetRenderers().GetFirstRenderer().
  * ResetCameraClippingRange after camera moves so clipping follows.

StaticPagesRequestHandler
  * Coerce docroot to bytes in __init__ so callers that pass a str no longer trip "TypeError: Can't mix strings and bytes in path components" when composing with bytes URIs elsewhere in the handler.

### DIFF
--- a/Modules/Scripted/WebServer/WebServer.py
+++ b/Modules/Scripted/WebServer/WebServer.py
@@ -16,7 +16,7 @@ from slicer.i18n import translate
 from slicer.ScriptedLoadableModule import *
 from slicer.util import settingsValue, toBool
 
-from WebServerLib.BaseRequestHandler import BaseRequestHandler, BaseRequestLoggingFunction
+from WebServerLib.BaseRequestHandler import BaseRequestHandler, BaseRequestLoggingFunction, KEEP_OPEN
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


lassoan (commented):
It looks good to me, but it would help a lot if you could add a test (but at least an example).

pieper:
I'm going to leave this in draft form for now.  If we can get funding I'm thinking a complete rewrite to modernize the WebServer will be manageable.  There are a lot of new options since this code was first written and probably we drop 90% of the custom implementation for a maintained modern package.  Probably we'll work on some prototyping during project week.

pieper:
Closing: the WebSocket support this hook was meant to enable is instead handled by a standalone QTcpServer (RFC 6455) on its own port in the SlicerLive mrson live channel, which is working well as the path forward for mrson serving. The socket-takeover hook isn't needed for that, so I'm closing to keep the queue clean. The small independent bug fixes here (qMRMLThreeDView renderer lookup, ResetCameraClippingRange, StaticPages docroot bytes coercion) and the threeD camera controls can be re-proposed separately if wanted.

PR #9124, opened 2026-04-23
BUG: When reparenting a selected node maintain selection

xskere:
When a subject hierarchy item is reparented programmatically (e.g. via "Move item here" in the context menu, or through scene loading that triggers a parent change) or when children of a node are reordered, the current selection in the Subject Hierarchy tree view is silently lost.

The root cause is in how Qt handles `QStandardItemModel::takeRow()` followed by `insertRow()`. This remove-then-reinsert sequence causes `QItemSelectionModel` to invalidate and clear the selection for the affected rows, since from the selection model's perspective the original item index no longer exists.

This regression affects any workflow that triggers subject hierarchy reparenting: drag-and-drop already had a workaround via `dropMimeData`, but programmatic reparenting and children reordering did not.

This solution adds two new signals to `qMRMLSubjectHierarchyModel`:

`requestSaveSelection()` emitted immediately before `takeRow()` is called during reparenting or reordering.
`requestRestoreSelection()` emitted immediately after `insertRow()` completes.

Each connected `qMRMLSubjectHierarchyTreeView` responds to these signals independently via two new slots:

`saveSelectionBeforeModelUpdate()` snapshots the current `SelectedItems` list into a separate `SelectedItemsSavedByModel` buffer.
`restoreSelectionAfterModelUpdate()` calls `setCurrentItems()` with the saved list and then clears the buffer.

This design keeps the model free of any view-specific logic. Multiple views connected to the same model each manage their own saved selection independently, so no cross-view coupling is introduced. The save/restore buffer (`SelectedItemsSavedByModel`) is separate from the existing `SelectedItemsToRestore` mechanism (which handles full tree rebuilds during batch scene processing) to avoid interference between the two paths.

Re [Project Week 44, Subject hierarchy combobox selection bug](https://projectweek.na-mic.org/PW44_2026_GranCanaria/Projects/DebugParemeterNodeAndQtmrmlcomboboxSetToNoneAfterAddingToASubjectHierarchy/)

### DIFF
--- a/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyModel.cxx
+++ b/Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyModel.cxx
@@ -1028,9 +1028,22 @@ void qMRMLSubjectHierarchyModel::updateItemFromSubjectHierarchyItem(QStandardIte
       int newIndex = this->subjectHierarchyItemIndex(shItemID);
       if ((newParentItem != nullptr) && (parentItem != newParentItem || newIndex != item->row()))
       {
+        // Save selection before reparenting: takeRow/insertRow is a remove+insert at the Qt
+        // model level, which causes QItemSelectionModel to clear selection for the moved row.
+        // Each connected view independently saves and restores its own selection.
+        // Drag-and-drop operations restore selection separately in dropMimeData, so skip here.
+        if (d->DraggedSubjectHierarchyItems.isEmpty())
+        {
+          emit requestSaveSelection();
+        }
         // Reparent items
         QList<QStandardItem*> children = parentItem->takeRow(item->row());
         newParentItem->insertRow(newIndex, children);
+        // Restore selection (each connected view restores its own previously saved selection)
+        if (d->DraggedSubjectHierarchyItems.isEmpty())
+        {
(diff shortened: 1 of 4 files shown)


cpinter:
Thanks @xskere !

@lassoan What do you think of this change? I spent a lot of time on this during the project week, trying a less "invasive" solution, but what seemed to be the correct solution (as I remember there was a `beginMoveRows` with an end pair) did not work.

I think if we want to fix this bug, this is our best option right now. Note that I have not worked on this using Qt6.

cpinter:
@lassoan This is a usability issue especially with SH comboboxes. If you have a few minutes to look at this, it would be great! Maybe this still fits 5.12. Thank you!

cpinter:
@lassoan Can you please check? Btw this is the follow-up of the PW project https://projectweek.na-mic.org/PW44_2026_GranCanaria/Projects/DebugParemeterNodeAndQtmrmlcomboboxSetToNoneAfterAddingToASubjectHierarchy

Thanks!

cpinter:
@jamesobutler Is this commit a conflict resolution one? I've seen other commits force pushed but no comment about them. Thanks!

jamesobutler:
@cpinter It was a straight "Update with rebase" from the web interface. I think the only commits no longer present were Merge commits which are auto resolved during a rebase event.

image

PR #9132, opened 2026-04-28
ENH: Pin transitive dependencies in python-dicom-requirements

### LINKED ISSUE #9131: Build fails in Linux at highdicom step
It's likely that this [commit](https://github.com/Slicer/Slicer/commit/4ecbcac693a6c7c2694f8bb6d10fb03277444957) is the source of the problem.

Building in Arch Linux stops when the build process tries to install highdicom.

After commenting this requirement, the build completes.

I will check tomorrow if there no runtime issues following this change.

Juste a note.

Regards.

fedorov:
Add pyjpegls and typing_extensions as explicit pinned entries with SHA-256 hashes so that pip --require-hashes does not fail when highdicom pulls them in as transitive dependencies.

Intended to resolve #9131.


pieper (approved):
Looks good to me 👍

fedorov:
I am sorry, you beat me to it -- I forgot to test the build on linux! I will do it now and follow up with another PR if it does not work!

fedorov:
I confirm clean linux build works!

PR #9133, opened 2026-04-28
ENH: Move pip requirements from CMake here-docs into standalone .txt files

fedorov:
## Summary

- Extracts the `file(WRITE ...)` requirements blocks embedded inside each of the 11 `SuperBuild/External_python-*.cmake` files into standalone, version-controlled `.txt` files under `SuperBuild/requirements/`.
- Each cmake file now simply sets `requirements_file` to point at the corresponding source-tree `.txt` file; the `INSTALL_COMMAND` is unchanged (`pip install --require-hashes -r <file>`).
- Adds a new GitHub Actions workflow (`.github/workflows/pip-requirements.yml`) that validates every requirements file in CI: it installs the file with `--require-hashes` in a fresh Python 3.12 venv (with the same baseline packages that the SuperBuild ExternalProject ordering guarantees), ensuring no transitive dependency can slip in without an explicit pinned hash.

## Motivation

Previously, the pinned requirements lived inside CMake bracket strings — they were not visible or testable without running a CMake configure step. Moving them to plain `.txt` files makes them:

- **Auditable**: reviewers and automated tools can read and diff requirements without CMake.
- **Validated in CI**: the new workflow catches missing transitive dependencies before they reach a release build (the immediate trigger was `highdicom` pulling in `pyjpegls` and `typing_extensions` without those being explicitly listed).
- **Easier to update**: `pip-compile` and similar tooling work directly on `.txt` files.

## Test plan

- [ ] All 11 CI jobs in the new `pip-requirements` workflow pass (dry-run via the PR checks).
- [ ] Slicer SuperBuild configures and builds cleanly with the updated cmake files.


.github/workflows/pip-requirements.yml
```diff
@@ -0,0 +1,68 @@
...
+    paths:
+      - "SuperBuild/requirements/*.txt"
+
+permissions:
+  contents: read
+
+jobs:
+  validate:
+    name: ${{ matrix.file }}
+    runs-on: ubuntu-latest
```
jamesobutler, line 24: It would be better to use the nightly Slicer build tree so that the tests using the specific Linux environment for determining if there is a compatible whl for the supported manylinux version. For checking things such as:
https://github.com/Slicer/Slicer/blob/a087af0c6cb44789dcd6df06cf6291d1ecd561a2/SuperBuild/External_python-numpy.cmake#L39

Also this is currently only checking Linux compatibility when historically when we update the Python versions we would run build checks on all platforms prior to integration. So we would want to confirm Windows and macOS package version compatibility as some Python packages occasionally have platform specific package requirements. This is because there are many C Python packages (like numpy) where there are different whl files for different versions and we wouldn’t want to introduce into Slicer core a C Python extension with only Linux whls.

fedorov: We could add testing for multiple platforms, but this would mean 33 CI jobs. The idea is to improve the existing status, where contributor may discover the failure only after the nightly build.

Alternatively, we can scrap the CI workflow altogether, and just keep the part that moves pip requirements into separate files out of CMake strings.

fedorov:
I confirm the build works on linux, so I am marking this out of draft.

.github/workflows/pip-requirements.yml
jamesobutler: To avoid 33 CI jobs, change it up to not be a workflow matrix like it is currently. The types of local testing we've done previously when changing python packages has been a wholistic check because we also have to confirm that one requirements.txt file is compatible with another requirements.txt which is not always guaranteed. Checking that a single requirements (e.g. python-extension-manager-requirements.txt) can be installed by fresh Python 3.12 instance is not sufficient. We have to check its compatibility against other requirements as well (e.g. python-numpy.txt). For example if a package like high dicom began enforcing something like numpy >2.6 while our separate numpy requirements state numpy==2.4.0, then the 2 requirements.txt files are incompatible and we would have a build issue of it not installing the appropriate packages as we are stating in our build system.

--------------------------
As of right now the workflow is not being comprehensive in making sure a change in pip requirements is not going to result in the Slicer factory builds failing overnight. This is only confirming that the Slicer Linux factory build won't fail, but not for the Windows and macOS factory builds as Slicer core uses python packages with whls that are platform specific. Just doing the linux platform would have been ok if we were only using pure python packages (e.g. `highdicom-0.27.0-py3-none-any.whl`), but we're not here.

PR #9134, opened 2026-04-30
BUG: Guard null SegmentationNode in qMRMLSegmentEditorWidget

Sunderlandkyl:
Fix null dereference in qMRMLSegmentEditorWidget::setSourceRepresentationToBinaryLabelmap when no Segmentation node is set by returning early if SegmentationNode is null.

### DIFF
--- a/Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.cxx
+++ b/Modules/Loadable/Segmentations/Widgets/qMRMLSegmentEditorWidget.cxx
@@ -803,6 +803,12 @@ bool qMRMLSegmentEditorWidget::setSourceRepresentationToBinaryLabelmap()
     return d->Logic->TrivialSetSourceRepresentationToBinaryLabelmap();
   }
 
+  if (!d->SegmentationNode)
+  {
+    qDebug() << Q_FUNC_INFO << " failed: No segmentation node.";
+    return false;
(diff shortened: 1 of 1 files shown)


lassoan (approved):
Good catch! The fix looks good to me.

The debug-level message is appropriate, since it is a protected member (we don't need to worry about developers calling it and not noticing that it failed).

Sunderlandkyl:
Seems there is a failure on the nightly Linux build that is causing the CI to fail:

```
CMake Error at /usr/src/Slicer-build/Slicer-build/CMake/LastConfigureStep/cmake_install.cmake:1117 (file):
  file INSTALL cannot find
  "/usr/src/Slicer-build/CTK-build/CMakeExternals/Install/lib/libQtTesting.so":
  No such file or directory.
  /usr/src/Slicer-build/Slicer-build/cmake_install.cmake:162 (include)


/opt/rh/gcc-toolset-14/root/usr/bin/strip: '/usr/src/Slicer-build/Slicer-build/_CPack_Packages/linux-amd64/TGZ/Slicer-5.11.0-2026-04-30-linux-amd64/./lib/Slicer-5.11/libQtTesting.so': No such file
CPack Error: Error when generating package: Slicer
```

Linux nightly CDash failure:
https://slicer.cdash.org/builds/4205217/build

PR #9136, opened 2026-04-30
Fix dcmqi binary lookup

fedorov:
Follow up on #9114.

@sjh26 @lassoan @pieper do you have any idea why this test passed?

https://github.com/Slicer/Slicer/blob/main/Applications/SlicerApp/Testing/Python/DICOMPluginTests.py#L70

https://slicer.cdash.org/tests/32893890

PR #9137, opened 2026-04-30
BUG: Resolve QtTesting install library name from imported target

### LINKED ISSUE #9135: Linux build error due to missing libQtTesting.so
Nightly build fails on linux due to missing libQtTesting.so - https://slicer.cdash.org/builds/4205217/build

```
CMake Error at /usr/src/Slicer-build/Slicer-build/CMake/LastConfigureStep/cmake_install.cmake:1117 (file):
  file INSTALL cannot find
  "/usr/src/Slicer-build/CTK-build/CMakeExternals/Install/lib/libQtTesting.so":
  No such file or directory.
  /usr/src/Slicer-build/Slicer-build/cmake_install.cmake:162 (include)


/opt/rh/gcc-toolset-14/root/usr/bin/strip: '/usr/src/Slicer-build/Slicer-build/_CPack_Packages/linux-amd64/TGZ/Slicer-5.11.0-2026-04-30-linux-amd64/./lib/Slicer-5.11/libQtTesting.so': No such file
CPack Error: Error when generating package: Slicer
```

hjmjohnson:
Fixes #9135. The CTK QtTesting bump (CTK pins QtTesting at [`6d153156`](https://github.com/commontk/QtTesting/commit/6d1531564ae61f702fdd0d63f573fef08853460f), which adds `OUTPUT_NAME qttesting`) renamed the shared library output from `QtTesting` to `qttesting`, causing nightly Linux packaging to fail. This change queries the actual library file name from the imported `qttesting` CMake target instead of hard-coding it, so future renames don't silently break packaging.

<details>
<summary>Root cause</summary>

`CMake/SlicerBlockInstallQtTesting.cmake` hard-coded `libQtTesting.so` / `libQtTesting.dylib` / `QtTesting.dll`. After CTK began pulling in QtTesting commit `6d153156` (which sets `OUTPUT_NAME qttesting`), CTK installs `libqttesting.so` (lowercase), so the install step errors:

```
file INSTALL cannot find
  /usr/src/Slicer-build/CTK-build/CMakeExternals/Install/lib/libQtTesting.so
```

</details>

<details>
<summary>Fix details</summary>

- `find_package(QtTesting CONFIG REQUIRED)` to load the imported target.
- Probe target names (`qttesting`, `QtTesting`, `QtTesting::qttesting`, `CTK::QtTesting`) so a future namespaced rename fails loudly instead of silently.
- Resolve the library basename from `IMPORTED_LOCATION_<CONFIG>` (with fallbacks across known config types and `IMPORTED_LOCATION`).
- Use that basename for both `install(FILES …)` and `slicerStripInstalledLibrary(…)`.
- Documents a known limitation: assumes a single unversioned shared library file (no `libqttesting.so.1` SONAME chain). If upstream introduces SONAME versioning later, the block needs to install the symlink chain too — a comment in the file flags this.

Verified locally: the generated `cmake_install.cmake` now references `libqttesting.so` from the actual CTK install tree.

</details>

## Test plan
- [x] Cross-validated by @fedorov on #9136 — that PR's CI build succeeds with this fix in place.
- [ ] Linux nightly package build (CDash) succeeds: previously failing `cmake_install.cmake:1117` step now installs `libqttesting.so`.
- [ ] macOS package contains `libqttesting.dylib`.
- [ ] Windows package contains `qttesting.dll`.

### DIFF
--- a/CMake/SlicerBlockInstallQtTesting.cmake
+++ b/CMake/SlicerBlockInstallQtTesting.cmake
@@ -1,20 +1,82 @@
 # -------------------------------------------------------------------------
 # Find and install QtTesting
 # -------------------------------------------------------------------------
+#
+# QtTesting is built and installed by the CTK external project. Slicer
+# packages the resulting shared library from the CTK install tree.
+#
+# The library file name is queried from the imported QtTesting CMake
+# target rather than hard-coded, so that future upstream changes to the
+# library output name (for example the Qt5+Qt6 rename of `QtTesting` to
+# `qttesting`) do not silently break Slicer packaging.
+#
+# NOTE: This logic assumes upstream installs a single, unversioned shared
+# library file (no SONAME symlink chain such as libqttesting.so.1 ->
+# libqttesting.so). If a future QtTesting revision introduces SONAME
+# versioning, this block must also install the symlink chain — for
+# example by globbing `libqttesting.so*` or by installing both
+# TARGET_SONAME_FILE and TARGET_FILE names.
 
 set(QtTesting_INSTALL_LIB_DIR "${Slicer_INSTALL_LIB_DIR}")
 
+find_package(QtTesting CONFIG QUIET REQUIRED)
+
+# Probe known QtTesting imported-target names (current and likely-future
+# spellings) so a future namespaced rename upstream is caught loudly.
+set(_qttesting_target "")
+foreach(_candidate qttesting QtTesting QtTesting::qttesting CTK::QtTesting)
+  if(TARGET ${_candidate})
+    set(_qttesting_target ${_candidate})
+    break()
+  endif()
+endforeach()
+if(NOT _qttesting_target)
+  message(FATAL_ERROR
+    "SlicerBlockInstallQtTesting: no imported QtTesting target was "
+    "defined by find_package(QtTesting). Tried: qttesting, QtTesting, "
+    "QtTesting::qttesting, CTK::QtTesting. Upstream may have renamed "
+    "the target; extend the probe list above.")
+endif()
+
+# Resolve the shared library file name (e.g. libqttesting.so,
+# libqttesting.dylib, qttesting.dll) from the imported target.
+set(_qttesting_imported_location "")
+foreach(_cfg
+    "${CMAKE_BUILD_TYPE}"
+    ${CMAKE_CONFIGURATION_TYPES}
+    Release RelWithDebInfo MinSizeRel Debug NOCONFIG
+    )
+  if(NOT _cfg)
+    continue()
+  endif()
+  string(TOUPPER "${_cfg}" _cfg_upper)
+  get_target_property(_qttesting_imported_location
+    ${_qttesting_target} IMPORTED_LOCATION_${_cfg_upper})
+  if(_qttesting_imported_location)
+    break()
+  endif()
+endforeach()
+if(NOT _qttesting_imported_location)
+  get_target_property(_qttesting_imported_location
+    ${_qttesting_target} IMPORTED_LOCATION)
+endif()
+if(NOT _qttesting_imported_location)
+  message(FATAL_ERROR
+    "SlicerBlockInstallQtTesting: failed to read IMPORTED_LOCATION from "
+    "the '${_qttesting_target}' target.")
+endif()
+get_filename_component(_qttesting_libname "${_qttesting_imported_location}" NAME)
+
 if(WIN32)
-  install(FILES ${QtTesting_INSTALL_DIR}/bin/QtTesting.dll
+  install(FILES "${QtTesting_INSTALL_DIR}/bin/${_qttesting_libname}"
     DESTINATION bin COMPONENT Runtime)
 elseif(APPLE)
-  # needs to install symlink version named libraries as well.
-  install(FILES ${QtTesting_INSTALL_DIR}/lib/libQtTesting.dylib
+  install(FILES "${QtTesting_INSTALL_DIR}/lib/${_qttesting_libname}"
     DESTINATION ${QtTesting_INSTALL_LIB_DIR} COMPONENT Runtime)
 elseif(UNIX)
-  install(FILES ${QtTesting_INSTALL_DIR}/lib/libQtTesting.so
+  install(FILES "${QtTesting_INSTALL_DIR}/lib/${_qttesting_libname}"
     DESTINATION ${QtTesting_INSTALL_LIB_DIR} COMPONENT Runtime)
   slicerStripInstalledLibrary(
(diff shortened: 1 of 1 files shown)


hjmjohnson:
@lassoan I don't think the CI environment will hit this failure — the breakage shows up in the **package/install** step (`make package` / CPack), which our regular CI doesn't run. Could you poke the nightly packaging system to verify this fixes it end-to-end? My local testing only confirms the install rule is generated correctly; the actual CPack run on the nightly box is the real validation.

<details>
<summary>Local testing performed</summary>

Single in-tree superbuild on Linux. Limited but covers the failing nightly path.

**Environment**
- Ubuntu 24.04.4 LTS, kernel 6.17.0-1017-oem (x86_64)
- GCC 13.3.0
- CMake 4.2.1, Ninja generator

**Slicer / dependency versions**
- Slicer 5.11.0 (`main` @ `97c6a4971e` + this fix)
- Qt 5.15.13 (system `qtbase5-dev`); `Slicer_REQUIRED_QT_VERSION = 5.15`
- `Slicer_USE_QtTesting = ON`, `CMAKE_BUILD_TYPE = Debug`
- CTK QtTesting tag: `cf5fa4156734edb9c3515c7104c19783b17ed0c1` — the same renamed-library tag that broke the nightly

**What I verified**
1. `cmake .` reconfigure of `Slicer-build` succeeds (no fatal errors from the new `find_package(QtTesting)` / target probe / `IMPORTED_LOCATION` walk).
2. The generated `Slicer-build/CMake/LastConfigureStep/cmake_install.cmake` now references the actual on-disk file:
   ```
   file(INSTALL DESTINATION ".../lib/Slicer-5.11" TYPE FILE FILES
     ".../CTK-build/CMakeExternals/Install/lib/libqttesting.so")
   ```

</details>

<details>
<summary>What was NOT tested locally — needs nightly / CI / reviewer coverage</summary>

- **Full `make package` / CPack on Linux** — would exercise `slicerStripInstalledLibrary` against the installed file. *(This is the one that matters for #9135.)*
- **macOS** package (`libqttesting.dylib`).
- **Windows** package (`qttesting.dll` from `${QtTesting_INSTALL_DIR}/bin/`).
- **Release / RelWithDebInfo** configurations — only Debug was on hand. The config-walk loop is written to handle all of them, but only Debug was actually exercised.
- **Qt6 path** — currently dead code, since Slicer forces `Slicer_USE_QtTesting=OFF` under Qt6 (`CMakeLists.txt:217-221`). Will become relevant when Qt6 QtTesting is re-enabled.

</details>

jamesobutler:
> The CTK QtTesting tag bump (cf5fa41…, "ctk-2026-03-27 Qt5+Qt6 support")

According to what code bumped QtTesting to that commit? Latest CTK specifies at https://github.com/commontk/CTK/blob/549a5b7b72f7b63a70ad91f6d183c7728511f794/CMakeExternals/QtTesting.cmake#L30 a different QtTesting commit hash of https://github.com/commontk/QtTesting/commit/6d1531564ae61f702fdd0d63f573fef08853460f.

fedorov:
I think my PR CI fails exactly due to the issue that this PR is expected to fix!

https://github.com/Slicer/Slicer/pull/9136

fedorov:
@hjmjohnson I tested and confirm #9136 CI build succeeds with this fix in place. I don't think I have the knowledge of CMake or Qt to review the actual changes, but I hope this is helpful.

fedorov (commented):
## Code Review

*This review was prepared with the assistance of [Claude Code](https://claude.ai/code) (Anthropic), using the `slicer-skill` and `review` agents to fetch the PR diff, examine upstream QtTesting CMake sources, and cross-reference the current CTK tag. The review was verified and posted by @fedorov.*

---

### What it does

Fixes a nightly Linux packaging regression where `cmake_install.cmake` fails with "file INSTALL cannot find ... `libQtTesting.so`". The root cause is that upstream QtTesting renamed its shared library from `libQtTesting.so` to `libqttesting.so` (lowercase), but `SlicerBlockInstallQtTesting.cmake` had the old name hard-coded. The fix queries the actual filename from the imported CMake target at configure time instead.

---

### Correctness

The core logic is sound:
- Probing four candidate target names (`qttesting`, `QtTesting`, `QtTesting::qttesting`, `CTK::QtTesting`) with a `FATAL_ERROR` on no-match is appropriate.
- The `IMPORTED_LOCATION` config-walk (checking `CMAKE_BUILD_TYPE`, then `CMAKE_CONFIGURATION_TYPES`, then a fixed list, then bare `IMPORTED_LOCATION`) correctly handles both single-config and multi-config CMake generators.
- Variable quoting is improved over the original code.

---

### Issues found

| Severity | Issue |
|----------|-------|
| Minor | PR description cites the wrong QtTesting commit hash (`cf5fa41`) — the actual breaking commit in the current CTK tag is `6d153156`, which already contains `OUTPUT_NAME qttesting`. The fix is still correct; just the description attribution is inaccurate. |
| Minor | The `find_package(QtTesting CONFIG REQUIRED)` call may be redundant if the target is already defined before this file is included — a comment clarifying intent would help. |
| Low | `NOCONFIG` in the config-walk fallback list (mapping to `IMPORTED_LOCATION_NOCONFIG`) is non-standard and deserves a one-line comment. |
| Low | Windows DLL path relies on `IMPORTED_LOCATION` resolving to the install tree rather than the build tree — noted by the PR author as untested. |
| Info | SONAME symlink limitation is correctly documented in comments; a defensive regex check for versioned filenames would be more robust, but this is a pre-existing concern not introduced by this PR. |

---

### Tests

No automated tests expected or included — this is a CPack/install rule. Validation comes from nightly CDash builds. The related PR #9136 passed CI with this fix applied, providing cross-validation.

---

### Recommendation

**Approve with nits.** The fix correctly resolves the regression and the CMake logic is solid. Suggested before merging:

1. Correct the commit hash in the PR description (`cf5fa41` → `6d153156`).
2. Add a brief comment on the `find_package` guard and the `NOCONFIG` fallback.
3. Confirm on nightly CDash that the Linux packaging job passes.

fedorov (approved):

hjmjohnson:
Thanks for the careful reads, @jamesobutler and @fedorov.

@jamesobutler — you're right, the original PR description had the wrong hash. The actual upstream change that renames the library to lowercase is QtTesting commit [`6d153156`](https://github.com/commontk/QtTesting/commit/6d1531564ae61f702fdd0d63f573fef08853460f), pulled in via CTK `549a5b7` (`External_CTK.cmake:80`). I've corrected the description.

@fedorov — much appreciated for cross-validating on #9136. That's exactly the end-to-end signal I was hoping for. The fix itself is unchanged.

PR #9141, opened 2026-05-05
BUG: Restore removed slicer.util._executePythonModule

lassoan:
`slicer.util._executePythonModule` function has been widely used in many extensions. Recently this function was moved to packaging, and since then these extensions no longer work. The name started with an underscore because it was originally intended to be an internal function, but then it turned out to be essential, and we started using it in extensions. It should be actually a public utility function.

This commit restores `slicer.util._executePythonModule` to make extensions work again, but makes it deprecated. From now on, `slicer.util.executePythonModule` function should be used instead.

### DIFF
--- a/Base/Python/slicer/packaging.py
+++ b/Base/Python/slicer/packaging.py
@@ -21,6 +21,7 @@ import importlib
 import importlib.metadata
 import logging
 import shlex
+import slicer.util
 import sys
 import tomllib
 from pathlib import Path
@@ -32,17 +33,12 @@ from packaging.requirements import InvalidRequirement, Requirement
 from packaging.utils import canonicalize_name
 
 from slicer.i18n import tr as _
-from slicer.util import executePythonModule
 
 if TYPE_CHECKING:
     from collections.abc import Callable
 
     import qt
 
-
-# Tests can monkeypatch _executePythonModule to perform mock actions instead of actually executing
-# the Python module by using slicer.packaging._executePythonModule.
-_executePythonModule = executePythonModule
 def load_requirements(path: str | Path) -> list[Requirement]:
     """Load requirements from a requirements.txt file.
 
@@ -688,11 +684,11 @@ def _pip_install_simple(
     # Handle no_deps_requirements first
     if no_deps_requirements is not None:
         args = _build_pip_args(no_deps_requirements, constraints, no_deps=True)
-        _executePythonModule("pip", args, blocking=True)
+        slicer.util.executePythonModule("pip", args, blocking=True)
 
     # Then install regular requirements
     args = _build_pip_args(requirements, constraints, no_deps=False)
-    _executePythonModule("pip", args, blocking=True)
+    slicer.util.executePythonModule("pip", args, blocking=True)
 
 
 def _pip_install_with_busy_cursor(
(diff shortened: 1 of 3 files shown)


Base/Python/slicer/util.py
```diff
@@ -4082,6 +4082,75 @@ def logProcessOutput(proc, logCallback=None):
...
+    blocking: bool = True,
+    logCallback: Callable[[str], None] | None = None,
+    completedCallback: Callable[[int], None] | None = None,
+) -> None:
+    """Execute a Python module as a script in Slicer's Python environment.
+    .. deprecated:: 5.11.0
+       Use :func:`executePythonModule` function instead.
+    """
+    import warnings
+    warnings.warn("_executePythonModule is deprecated, use executePythonModule instead", DeprecationWarning, stacklevel=2)
```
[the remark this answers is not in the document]
lassoan: The deprecation warning appears in the Python console.

ebrahimebrahim:
Thanks for fixing this. My bad for moving `_executePythonModule`. I definitely agree with deprecating and renaming without the initial underscore, since the initial underscore is what suggested it was okay to move

Will take a look and review this soon

Base/Python/slicer/util.py
ebrahimebrahim, line 4164: Notice that we now narrowly escape circular import thanks to the deferred import here. Because of this I previously wanted to avoid importing `slicer.util` inside `slicer.packaging`.

I was going to suggest leaving `executePythonModule` inside of `slicer.packaging` and adding a convenience wrapper for it here in `slicer.util`, to follow this pattern. But if extensions are using `executePythonModule`, I suppose the proper place for it to live is `slicer.util`. So I agree with the approach here

Base/Python/slicer/packaging.py
```diff
@@ -34,65 +32,17 @@
...
-            useStartupEnvironment=False,
-            blocking=False,
-            logCallback=logCallback,
-            completedCallback=completedCallback,
-        )
-
-
+# Tests can monkeypatch _executePythonModule to perform mock actions instead of actually executing
+# the Python module by using slicer.packaging._executePythonModule.
+_executePythonModule = executePythonModule
```
ebrahimebrahim: Sure they _can_ monkeypatch but there's no reason to. They can mock `executePythonModule` just as well as the can mock `_executePythonModule`.

This line makes the PR smaller which is nice, but I'd really rather we properly rename to executePythonModule in the code.

It should be just a find-and-replace in `Base/Python/slicer/packaging.py` and `Base/Python/slicer/tests/test_slicer_packaging.py`

ebrahimebrahim (changes requested):
Thanks for fixing this. Just one change requested: better stop using the name `_executePythonModule` in `slicer.packaging`, rather than renaming `_executePythonModule = executePythonModule`

lassoan:
I pushed a commit to remove the additional _executePythonModule in packaging, please have a look.

ebrahimebrahim:
Thanks. Running packaging tests on my end...

ebrahimebrahim (approved):
All slicer.packaging and slicer.util tests pass for me on linux, with the exception of `py_nomainwindow_test_slicer_util_load_data` due to something related to highdicom (?) but I suppose that's nothing to do with this PR

Thank you!!

Leaving merge to you

lassoan:
Merged. Thank you for the review and testing @ebrahimebrahim

PR #9146, opened 2026-05-08
Color individual control points

ebrahimebrahim:
Adds per-point color support to `vtkMRMLMarkupsNode`. It is off by default.

When enabled, per-point colors override the Selected/Unselected colors when they are set.
They all start out in an "unset" state.

The challenge with this was that the control points in a vtkMRMLMarkupsNode are not a vtkPolyData but rather a std::vector of ControlPoint structs called `ControlPoints`. 
The way it's done is by adding to vtkMRMLMarkupsNode a new member that is actually a vtkPolyData, called `ControlPointDataSet`, to which we can add color data any other data one might to add in the future. We still have `ControlPoints` as the source of truth, but now we also have `ControlPointDataSet` as a mirror of it. It's duplication, but it doesn't actually duplicate anything unless you go and try to `SetNthControlPointColor`, which is when it's needed.

Adds an optional `"color"` field per `controlPoints[]` entry in the markups JSON.

vid-loop-768x432-15fps

### Try

```python
m = slicer.modules.markups.logic().AddNewMarkupsNode("vtkMRMLMarkupsFiducialNode")
m.AddControlPoint([0, 0, 0])
m.AddControlPoint([40, 0, 0])
m.AddControlPoint([60, 0, 0])

# Override colors on individual control points (RGBA, 0..1)
m.SetNthControlPointColor(0, 1.0, 0.0, 0.0)
m.SetNthControlPointColor(2, 0.0, 0.7, 1.0, 0.5)

# Enable per-point colors on the display node
m.GetDisplayNode().SetUseControlPointColors(True)

# Inspect / clear per-point colors
print(m.IsNthControlPointColorOverridden(0))   # True
print(m.IsNthControlPointColorOverridden(1))   # False (uses Selected/Unselected fallback)
rgba = [0.0, 0.0, 0.0, 0.0]
m.GetNthControlPointColor(0, rgba); print(rgba)  # [1.0, 0.0, 0.0, 1.0]
m.ClearNthControlPointColor(0)                # revert point 0 to fallback
m.ClearAllControlPointColors()                # revert every point
```

### DIFF
--- a/Docs/developer_guide/script_repository/markups.md
+++ b/Docs/developer_guide/script_repository/markups.md
@@ -506,6 +506,36 @@ pointListDisplayNode.SetSelectedColor(1,1,0) # Set color to yellow
 pointListDisplayNode.SetViewNodeIDs(["vtkMRMLSliceNodeRed", "vtkMRMLViewNode1"]) # Only show in red slice view and first 3D view
 ```
 
+### Set per-control-point colors
+
+Each control point can be assigned its own color. Per-point colors apply to
+the control point glyphs and override the display node's Selected/Unselected
+colors for points where a color has been set; the Active color still
+highlights the active control point. The line or curve connecting the
+control points keeps the display node's line color.
+
+```python
+m = slicer.modules.markups.logic().AddNewMarkupsNode("vtkMRMLMarkupsFiducialNode")
+m.AddControlPoint([0, 0, 0])
+m.AddControlPoint([10, 0, 0])
+m.AddControlPoint([20, 0, 0])
+
+# Override colors on individual control points (RGBA, 0..1)
+m.SetNthControlPointColor(0, 1.0, 0.0, 0.0)        # red
+m.SetNthControlPointColor(2, 0.0, 0.7, 1.0, 0.8)   # blue, 80% opacity
+
+# Enable per-point colors on the display node
+m.GetDisplayNode().SetUseControlPointColors(True)
+
+# Inspect / clear per-point colors
+print(m.IsNthControlPointColorOverridden(0))   # True
+print(m.IsNthControlPointColorOverridden(1))   # False (uses Selected/Unselected fallback)
+rgba = [0.0, 0.0, 0.0, 0.0]
+m.GetNthControlPointColor(0, rgba); print(rgba)  # [1.0, 0.0, 0.0, 1.0]
+m.ClearNthControlPointColor(0)                # revert point 0 to fallback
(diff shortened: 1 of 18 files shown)


jamesobutler:
Color columns in subject hierarchy views usually have terminology support along with simple colors. Would it be possible to set individual terminology with its corresponding color for each control point? That way when doing landmarking different features say “left kidney” or “right kidney” the control points would be named and colored accordingly to the terminology.

Is this coloring exclusive to Point List type or also for things like ROI node and Angle node? Interaction handles for things like ROI have colored control points but that aren’t part of the control point list. I assume changing interaction handles color would not be happening here?

ebrahimebrahim:
> Color columns in subject hierarchy views usually have terminology support along with simple colors. Would it be possible to set individual terminology with its corresponding color for each control point? That way when doing landmarking different features say “left kidney” or “right kidney” the control points would be named and colored accordingly to the terminology.

James, I like the idea! But is this something that requires the table to be a subject hierarchy view? The `activeMarkupTableWidget` is a plain old `QTableWidget`. 

> Is this coloring exclusive to Point List type or also for things like ROI node and Angle node? Interaction handles for things like ROI have colored control points but that aren’t part of the control point list. 

It applies to all markups, but only to the point glyphs which is the objective here, but maybe it could be expanded. There are loads of choices to be made if you want to make the underlying curve somehow receive the point colors (interpolate, but then what do you in the 2D views where there is already some fancy projection coloring going on, etc.)



image

image

image

> I assume changing interaction handles color would not be happening here?

Yes, the interaction handle color is not touched

pieper:
This is great, it's a feature I've often wanted.

lassoan:
Thanks for working on this, it will be very useful! You can assign any additional per-controlpoint data to in measurements that are can be computed on the fly (e.g., you could implement a measurement that probes a volume to get color) or it can be static (you can assign constants). This can already be used for coloring the line, but I don't think think it colors control points. Probably all you need in MRML is a new property that specifies which measurement should be used for coloring of control points. I would recommend to allow not just direct RGB color assignment but also coloring via a scalar and a color table (to allow using consistent coloring, terminologies, etc.).

ebrahimebrahim:
Update: Closing this in favor of an improved approach at https://github.com/ebrahimebrahim/Slicer/tree/control-point-colors-revival based on the suggestions here. I will make a PR for that later when I feel it's ready, possibly much later as I haven't had time to work on this.

PR #9148, opened 2026-05-10
Support installing scripted extensions from source

ebrahimebrahim:
(this is quite experimental, don't look at this just yet)

<details>

<summary>ai generated pr description (WIP, will clean up)</summary>

## Summary

This adds first-class support for installing Python-only `source-scripted` extensions directly from source, without building an extension package. Source installs are driven by a required `slicer-extension.json` manifest, copied into the revision-specific extensions install path, and loaded after restart like other Extension Manager installs.

The implementation includes:

- Strict manifest parsing for `type: "source-scripted"`, module `.py` paths, extension dependencies, optional `pythonPaths`, and descriptor-only `installSource`.
- Local directory, local archive, URL archive, and git repository installs through model APIs and the Extension Manager UI.
- Extension Manager UI entry: "Install from source..." with Directory, Archive, Archive URL, and Git choices.
- Trust warning and manifest preview before install, including downloaded URL archives and inspected Git checkouts.
- Clear archive failure messages for unsupported downloads and archives without `slicer-extension.json`.
- Source-scripted path handling that only adds manifest-declared module directories to `Modules/AdditionalPaths` and manifest-declared Python paths to launcher `PYTHONPATH`.
- Git origin tracking with resolved commit SHA stored in the installed metadata and `.s4ext` metadata.
- Update checks for source-scripted git branch origins, with installed and available commit SHAs shown in the update UI; tag and commit origins are treated as fixed snapshots.
- `slicer://extensions/install-source?manifest=...` links for install badges. The remote manifest must declare an installable `installSource`, and the completion dialog reminds users to restart Slicer.
- Badge installs report already-installed extensions and other install failures in dialogs, instead of only logging to the Python console.
- User and developer documentation.

Source-scripted installs remain restart-required and intentionally do not install Python package dependencies. Manifest `depends` still means Slicer extension dependencies.

## How To Try It

Create a minimal source-scripted extension somewhere writable:

```bash
mkdir -p /tmp/SlicerScriptedExtensionExample/Modules/HelloSource
cat > /tmp/SlicerScriptedExtensionExample/Modules/HelloSource/HelloSource.py <<'PY'
from slicer.ScriptedLoadableModule import ScriptedLoadableModule


class HelloSource(ScriptedLoadableModule):
    def __init__(self, parent):
        parent.title = "Hello Source"
        parent.categories = ["Examples"]
        parent.contributors = ["Slicer Developer"]
        parent.helpText = "Minimal source-scripted extension test."
PY
cat > /tmp/SlicerScriptedExtensionExample/slicer-extension.json <<'JSON'
{
  "schema": "https://slicer.org/schemas/source-scripted-extension/v1",
  "type": "source-scripted",
  "extensionname": "SlicerScriptedExtensionExample",
  "description": "Minimal source-scripted extension test.",
  "contributors": "Slicer Developer",
  "modules": [
    {
      "name": "HelloSource",
      "path": "Modules/HelloSource/HelloSource.py"
    }
  ],
  "pythonPaths": [],
  "depends": []
}
JSON
```

Then in Slicer:

1. Open Extensions Manager.
2. Click "Install from source...".
3. Choose "Directory..." and select `/tmp/SlicerScriptedExtensionExample`.
4. Review the warning and manifest preview.
5. Install, then restart Slicer.
6. Confirm the "Hello Source" module appears in the Examples category.

The same source tree can be zipped and installed with "Archive...":

```bash
cd /tmp
zip -r SlicerScriptedExtensionExample.zip SlicerScriptedExtensionExample
```

## URL Archive Flow

Serve the zip from a local or remote web server and choose "Archive URL..." from the same Extension Manager flow. Slicer should download the archive, inspect the manifest, show the interpreted preview, and only install after confirmation.

Try a bad archive too:

```bash
printf '<html>not an archive</html>\n' > /tmp/not-an-extension.zip
```

Installing that file as an archive should report that it is not a supported source-scripted extension archive. A valid archive without `slicer-extension.json` should report that the manifest must be at the archive root or in one top-level directory.

## Git Origin Flow

To test git install and update checks:

```bash
cd /tmp/SlicerScriptedExtensionExample
git init
git checkout -B main
git config user.email slicer@example.invalid
git config user.name "Slicer Test"
git add -A
git commit -m "Initial source-scripted extension"
```

In Slicer:

1. Open Extensions Manager.
2. Click "Install from source...".
3. Choose "Git...".
4. Enter `/tmp/SlicerScriptedExtensionExample` and `main`.
5. Review the preview, install, and restart.

The same flow is available from the Slicer Python console:

```python
repo = "/tmp/SlicerScriptedExtensionExample"
model = slicer.app.extensionsManagerModel()
ok = model.installSourceScriptedExtensionFromGit(repo, "main", False)
print("installed:", ok)
print("metadata:", model.extensionMetadata("SlicerScriptedExtensionExample"))
```

After restart, update the repository with another commit, then use "Check for updates" in Extension Manager. Branch installs should report an available update. Tag and commit installs should remain fixed snapshots.

## Install Badge Flow

A badge link can point to an HTTPS landing page, because GitHub README rendering does not preserve direct `slicer://` links:

Here is an example of a badge to install a minimal example extension I've put at https://github.com/ebrahimebrahim/SlicerScriptedExtensionExample:

[Install in Slicer](https://ebrahimebrahim.github.io/SlicerScriptedExtensionExample/install.html)

The landing page's install button links to:

```text
slicer://extensions/install-source?manifest=https%3A%2F%2Fraw.githubusercontent.com%2Febrahimebrahim%2FSlicerScriptedExtensionExample%2Fmain%2Fslicer-extension.json
```

The remote manifest must include an installable `installSource`. For a Git-backed extension repository, the root `slicer-extension.json` can include:

```json
"installSource": {
  "type": "git",
  "url": "https://github.com/ebrahimebrahim/SlicerScriptedExtensionExample.git",
  "ref": "main"
}
```

Opening the `slicer://` link should download the manifest, show the trust warning and preview, and then install from the declared source if confirmed. Direct Directory, Archive, URL, and Git installs ignore `installSource`; they record the source selected by the user as the installed origin. Protocol registration is automatic for Windows installers and macOS application bundles; portable builds and some Linux desktop environments may need manual `slicer://` association.

A minimal example repository for this badge is `SlicerScriptedExtensionExample`, with:

- `slicer-extension.json` at the repository root.
- A scripted module at `Modules/SlicerScriptedExtensionExample/SlicerScriptedExtensionExample.py`.
- `install.html` served by GitHub Pages as the HTTPS badge target.
- The `installSource` block above so the badge can install from the Git repository.

## Python Smoke Test

From the Slicer Python console, the model API can be exercised directly:

```python
sourcePath = "/tmp/SlicerScriptedExtensionExample"
model = slicer.app.extensionsManagerModel()
ok = model.installSourceScriptedExtension(sourcePath, False)
print("installed:", ok)
print("metadata:", model.extensionMetadata("SlicerScriptedExtensionExample"))
print("module paths:", model.extensionModulePaths("SlicerScriptedExtensionExample"))
```

Restart Slicer after installation to load the module.

## Automated Testing

Locally verified with:

```bash
uvx --from pre-commit pre-commit run --all-files
env CCACHE_DIR=/tmp/slicer-ccache CCACHE_TEMPDIR=/tmp cmake --build . --target qSlicerBaseQTCoreCxxTests qSlicerBaseQTGUI
env QT_QPA_PLATFORM=offscreen XDG_CONFIG_HOME=/tmp/slicer-test-config XDG_CACHE_HOME=/tmp/slicer-test-cache CCACHE_DIR=/tmp/slicer-ccache CCACHE_TEMPDIR=/tmp ctest -R qSlicerExtensionsManagerModelTest --no-tests=error --verbose
```

The focused CTest run passed with `74 passed, 0 failed`.

</details>

## TODO

- [ ] consider adopting a python packaging based approach, rather than creating a new manifest format 
- [ ] add optional Slicer version constraints to manifest format? since installing from source does not require exact revision match.

### DIFF
--- a/Base/QTCore/Testing/Cxx/qSlicerExtensionsManagerModelTest.cxx
+++ b/Base/QTCore/Testing/Cxx/qSlicerExtensionsManagerModelTest.cxx
@@ -20,7 +20,13 @@
 
 // Qt includes
 #include <qSlicerCoreApplication.h>
+#include <QFileInfo>
+#include <QJsonDocument>
+#include <QJsonObject>
 #include <QJSEngine>
+#include <QProcess>
+#include <QTemporaryFile>
+#include <QTextStream>
 
 // CTK includes
 #include <ctkTest.h>
@@ -31,6 +37,7 @@
 
 // QtCore includes
 #include "qSlicerExtensionsManagerModel.h"
+#include "vtkArchive.h"
 #include "vtkSlicerConfigure.h"
 #include "vtkSlicerVersionConfigureMinimal.h"
 
@@ -63,6 +70,24 @@ private:
   void installHelper(qSlicerExtensionsManagerModel* model, const QString& os, int extensionId, const QString& tmp);
 
   bool uninstallHelper(qSlicerExtensionsManagerModel* model, const QString& extensionName);
+
+  bool writeSourceScriptedManifest(const QString& sourcePath,
+                                   const QString& extensionName,
+                                   const QString& modulePath = "SourceModule.py",
+                                   const QStringList& pythonPaths = QStringList(),
+                                   const QString& extensionType = "source-scripted",
+                                   const QStringList& depends = QStringList());
+  bool setSourceScriptedManifestInstallSource(const QString& sourcePath, const QJsonObject& installSource);
+  QString createSourceScriptedExtension(const QString& extensionName,
+                                        const QString& modulePath = "SourceModule.py",
+                                        const QStringList& pythonPaths = QStringList(),
+                                        const QStringList& depends = QStringList());
+  bool runGit(const QString& workingDirectory, const QStringList& arguments, QString* output = nullptr);
+  bool commitGitRepository(const QString& repositoryPath, const QString& message, QString* revision = nullptr);
+  QString createSourceScriptedGitRepository(const QString& extensionName,
+                                            QString* revision = nullptr,
+                                            const QString& tagName = QString(),
+                                            const QJsonObject& installSource = QJsonObject());
 
   bool resetTmp();
   QDir Tmp;
@@ -106,7 +131,17 @@ private slots:
   void testWriteAndParseExtensionDescriptionFile();
   void testWriteAndParseExtensionDescriptionFile_data();
 
+  void testParseSourceScriptedExtensionManifest();
+  void testParseSourceScriptedExtensionInstallSource();
+
   void testInstallExtension();
+
+  void testInspectSourceScriptedExtensionArchiveFailures();
+  void testDownloadAndInspectSourceScriptedExtension();
+  void testInstallSourceScriptedExtension();
+  void testInstallSourceScriptedExtensionFromArchive();
+  void testInspectSourceScriptedExtensionFromGit();
+  void testInstallSourceScriptedExtensionFromGit();
 
   void testUninstallExtension();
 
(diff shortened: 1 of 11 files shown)


lassoan:
Have you had a chance to look at this solution? https://github.com/Slicer/Slicer/pull/9118

It requires no change to existing procedures. We can build standard extension packages within 1 minute, by using a [single-line github workflow](https://github.com/lassoan/SlicerMONAIAuto3DSeg/pull/105/changes), which can then be installed using the extension manager. The main idea is that for Python-only extension building, testing, and packaging we can create a mock Slicer build tree from a Slicer source tree.
The extensions manager can install these packages. We could allow users to specify git repositories where we could download release packages from.

This idea can be easily extended to all extensions, so it would not only work for Python scripted modules - as long as there is a Slicer container that contains a full build tree.

Current limitations:
1. Packaging on macOS requires intel macOS runners on github, which may cause delays (windows and linux packages are built within 1 minute).
2. Building of required additional extensions is not implemented. (this would not be hard to add)

---

The approach installed in this PR has some merits, too. It is nice to just describe what is in the package and generate a package using Python tools (not CMake). However, for this we should not invent some new package format, manifest, etc. because standard Python packages can easily handle all this, and it would be a nice feature that the extensions could be distribute via PyPI (same way as napari). Each Python package that offers 3D Slicer modules could declare the `3d-slicer-extension` keyword in its description, so we could easily find all Slicer extensions on PyPI (similarly how we can [get all the Slicer extension on github](https://github.com/topics/3d-slicer-extension)) and a `[project.entry-points."slicer.extension"]` entry point so that Slicer can get all the Slicer extensions from the installed Python packages extremely quickly. AI chatbots can already automatically create the Python project structure, pyproject.toml, etc. github action to publish PyPI package when a new github release is created - so it should be all really easy and standard. At runtime, Slicer can query all the Python packages that provide Slicer modules like this (takes practically zero time, even if hundreds of Python packages are installed):

```
from importlib.metadata import entry_points
moduleFolders = []
eps = entry_points(group="slicer.extension")
for ep in eps:
    slicerExtension = = ep.load()
    moduleFolders.extend(slicerExtension.get_module_folders())
```

ebrahimebrahim:
I hadn't seen it. It is a great approach!

Especially the fact that the #9118 approach uses the existing extension install path rather than creating a new one, I think makes it desirable.

What I like in the PR here is it moves more of the extension install work to the slicer application side, with less of it on the extension developers' side. Extension developers would not need to worry about packaging. They can instead just point to the branch or tag they want to install, and it can be installed directly.

image

They still have the work of creating the manifest, but it's much less "dev ops" to worry about than building packages.

This also frees people from _having_ to get into the extension index. Easy 1-click install can be on the extension maintainer's side, by pointing to the manifest with a slicer:// URL. I was thinking it could be a badge like

[Install in Slicer](https://ebrahimebrahim.github.io/SlicerScriptedExtensionExample/install.html)

The badge can be clicked from a readme. (Github markdown was refusing to accept `slicer://` in a url, so I had it link to a page for this demo.)

Getting on the extension index would still be desired for discoverability, but it becomes no longer a limitation for people to share their work with others.

> However, for this we should not invent some new package format, manifest, etc. because standard Python packages can easily handle all this, and it would be a nice feature that the extensions could be distribute via PyPI (same way as napari).

Interesting idea! I suppose we could still have the nice easy install link via `pip install git+https://...`? Then it's highly preferable to the json manifest here.

My current thinking:
- I lean towards integrating #9118 sooner, as it will have easy review and provides a huge utility for much lower review effort. It is much lower stakes, with very little design lock-in compared to this other stuff here. And it reduces developer burden a lot
- Then if we later like the approach in the PR here, we can do some variant of it. Possibly with a python packaging based approach.

It will be great to discuss at the developers meeting.

PR #9151, opened 2026-05-11
ENH: Add pip --target fallback for read-only site-packages

hubutui:
On Linux distributions that package Slicer into system directories (e.g. /opt or /usr), the default Python site-packages is owned by root and not writable by regular users. This causes pip_install to fail when extensions try to install additional Python packages.

This change adds a fallback: when the default site-packages is not writable, _build_pip_args automatically appends --target pointing to ~/.local/lib/Slicer/python{X.Y}/site-packages/. The directory is also appended to sys.path at startup so that packages installed there are importable.

This does not affect Windows, macOS, or any scenario where Slicer is installed into a user-writable location.

### DIFF
--- a/Base/Python/slicer/__init__.py
+++ b/Base/Python/slicer/__init__.py
@@ -220,6 +220,20 @@ if not standalone_python:
         print(detail)
 
 # -----------------------------------------------------------------------------
+# Add Slicer user site-packages to sys.path so that packages installed via
+# pip --target are importable.
+from pathlib import Path as _Path
+
+_user_site = str(
+    _Path.home() / ".local" / "lib" / "Slicer"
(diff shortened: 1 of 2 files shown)


ebrahimebrahim:
This is a great solution!

hubutui:
According to this [check](https://github.com/Slicer/Slicer/actions/runs/25654772161/job/75300507507?pr=9151) and this [commit](https://github.com/Slicer/Slicer/commit/83968d6a9212609c2a2bad1a6a819f26f82b82c5), `PYTHONNOUSERSITE=1` is set to ignore user site python packages. Maybe this patch should be applied by packager who build 3dslicer for a linux distribution.

PR #9152, opened 2026-05-11
BUG: Fix GetFileNameWithoutExtension() failing when called with no arguments

jamesobutler:
The std::string overload added in https://github.com/Slicer/Slicer/commit/74d1fa0b9479c304d6dc18050ed94252e833aedb caused the VTK Python wrapper to see two overloads of GetFileNameWithoutExtension, breaking the default argument on the const char* version. As a result, calling the method with no arguments from Python raised:

  TypeError: GetFileNameWithoutExtension() takes exactly 1 argument (0 given)

Guard the std::string overload with #ifndef __VTK_WRAP__ so it is hidden from the VTK wrapper while remaining available to C++ callers.

### DIFF
--- a/Libs/MRML/Core/vtkMRMLStorageNode.h
+++ b/Libs/MRML/Core/vtkMRMLStorageNode.h
@@ -320,7 +320,11 @@ public:
   /// Remove supported extension from filename.
   /// If filename is not specified then the current FileName will be used.
   std::string GetFileNameWithoutExtension(const char* fileName = nullptr);
+// Make sure that the Python wrapper uses the char* version
+// (because the std::string& version does not have a default argument)
+#ifndef __VTK_WRAP__
   std::string GetFileNameWithoutExtension(const std::string& fileName) { return GetFileNameWithoutExtension(fileName.c_str()); }
+#endif
 
   /// Compression parameter that is used to save the node
   vtkSetMacro(CompressionParameter, std::string);


jamesobutler:
@lassoan I discovered this while updating my Slicer custom application to a recent Slicer version. Some of my own private modules were using `GetFileNameWithoutExtension` with no input arguments.

Libs/MRML/Core/vtkMRMLStorageNode.h
```diff
@@ -320,7 +320,9 @@ class VTK_MRML_EXPORT vtkMRMLStorageNode : public vtkMRMLNode
   /// Remove supported extension from filename.
   /// If filename is not specified then the current FileName will be used.
   std::string GetFileNameWithoutExtension(const char* fileName = nullptr);
+#ifndef __VTK_WRAP__
```
lassoan, line 325:
```suggestion
// Make sure that the Python wrapper uses the char* version
// (because the std::string& version does not have a default argument)
#ifndef __VTK_WRAP__
```

lassoan (commented):
Good catch, thanks a lot for the fix. It looks good to me, I just added a comment to make it less likely that future developers will remove the `#ifndef` because they don't know why it is there.

Libs/MRML/Core/vtkMRMLStorageNode.h
jamesobutler: I've added these comments.

lassoan (approved):
Thank you, it looks good.

PR #9154, opened 2026-05-11
BUG: Rebuild model display map after batch updates

Sunderlandkyl:
Clear existing display maps before updating from the MRML scene.

During batch processing, model nodes may be removed without removing their corresponding display pipeline. Rebuilding the map ensures stale pipelines are discarded after the batch operation completes.

### DIFF
--- a/Libs/MRML/DisplayableManager/vtkMRMLModelDisplayableManager.cxx
+++ b/Libs/MRML/DisplayableManager/vtkMRMLModelDisplayableManager.cxx
@@ -407,6 +407,14 @@ void vtkMRMLModelDisplayableManager::OnMRMLSceneEndClose()
 }
 
 //---------------------------------------------------------------------------
+void vtkMRMLModelDisplayableManager::OnMRMLSceneEndBatchProcess()
+{
+  this->PruneMissingNodes();
+  this->SetUpdateFromMRMLRequested(true);
+  Superclass::OnMRMLSceneEndBatchProcess();
+}
+
+//---------------------------------------------------------------------------
 void vtkMRMLModelDisplayableManager::UpdateFromMRMLScene()
 {
   // UpdateFromMRML will be executed only if there has been some actions
@@ -1231,6 +1239,29 @@ void vtkMRMLModelDisplayableManager::RemoveModelObservers(int clearCache)
   if (clearCache)
   {
     this->ClearDisplayMaps();
+  }
+}
+
+//---------------------------------------------------------------------------
+void vtkMRMLModelDisplayableManager::PruneMissingNodes()
+{
+  if (!this->GetMRMLScene())
+  {
+    return;
+  }
+
+  std::vector<vtkMRMLDisplayableNode*> removedDisplayableNodes;
+  for (auto iter = this->Internal->DisplayableNodes.begin(); iter != this->Internal->DisplayableNodes.end(); iter++)
+  {
+    vtkMRMLDisplayNode* modelDisplayNode = vtkMRMLDisplayNode::SafeDownCast(this->GetMRMLScene()->GetNodeByID(iter->first));
+    if (modelDisplayNode == nullptr)
+    {
+      removedDisplayableNodes.push_back(iter->second);
+    }
+  }
+  for (auto displayableNode : removedDisplayableNodes)
+  {
+    this->RemoveDisplayable(displayableNode);
   }
 }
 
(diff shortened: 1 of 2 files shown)


Libs/MRML/DisplayableManager/vtkMRMLModelDisplayableManager.cxx
```diff
@@ -406,6 +406,15 @@ void vtkMRMLModelDisplayableManager::OnMRMLSceneEndClose()
   this->RequestRender();
 }
 
+//---------------------------------------------------------------------------
+void vtkMRMLModelDisplayableManager::OnMRMLSceneEndBatchProcess()
+{
+  // Clear existing node pipelines, as they may be invalid after batch processing
+  this->RemoveModelObservers(1);
```
lassoan: Calling this with `clearCache=1` would mean that if anything activates batch processing for any reason then all the model pipelines are recreated, which may be significant processing if we have a few hundred model nodes in the scene.

We should check what is changed in the scene and do a more specific update (e.g., we could iterate through the display node IDs in the maps and remove just the orphan objects instead of removing them all).

lassoan (commented):
It will be good to fix this, but I'm wondering if we can do it without impacting batch update speed on large scenes.

Libs/MRML/DisplayableManager/vtkMRMLModelDisplayableManager.cxx
```diff
@@ -1234,6 +1242,29 @@ void vtkMRMLModelDisplayableManager::RemoveModelObservers(int clearCache)
...
+{
+  if (!this->GetMRMLScene())
+  {
+    return;
+  }
+
+  std::vector<vtkMRMLDisplayableNode*> removedDisplayableNodes;
+  for (auto iter = this->Internal->DisplayableNodes.begin(); iter != this->Internal->DisplayableNodes.end(); iter++)
+  {
+    vtkMRMLDisplayNode* modelDisplayNode = vtkMRMLDisplayNode::SafeDownCast(this->GetMRMLScene()->GetNodeByID(iter->first));
```
lassoan, line 1256: The variable documentation says `For the following maps, the key is the display node ID` but according to the code it seems that the key is the displayable node ID. Please double-check the variable names and types and maybe also introduce helper variable names, such as `std::string displayableNodeID = iter->first`.

lassoan (commented):
Thanks for the update, this will be good, I added just one comment about a potential mixup of display/displayable node.

Libs/MRML/DisplayableManager/vtkMRMLModelDisplayableManager.cxx
lassoan: It would be a bit less confusing and a bit more consistent with other displayable managers by introducing a Pipeline struct and replacing the several `std::map<std::string, vtkSmartPointer<...>>` maps by one `std::map<std::string, Pipeline>`.

Sunderlandkyl: Yes I noticed the inconsistency in the map key for DisplayableNodes. This was why I put the PR into draft mode until I implemented with a better fix.

I like the idea of creating the Pipeline struct, it should help reduce the possible points of failure.

Libs/MRML/DisplayableManager/vtkMRMLModelDisplayableManager.cxx
Sunderlandkyl, line 610: @lassoan Is there a reason that we create a pipeline for all "vtkMRMLDisplayableNode" in the scene instead of just "vtkMRMLModelNode"?

The DisplayableNodes map contains all sorts of nodes (Markups, Transforms, etc.), which I believe comes from this line.

Sunderlandkyl, line 615: @lassoan Do we still need to order the slice actors separately from the regular model actors, or was this left from a previous VTK version?

lassoan: Yes, I think we still need to have the slice actors in the back. I don't remember it was due to picking or transparent rendering, etc. but when it was not right then things did not work well. I would not touch it now, unless you have time to review the file history to get hints why it is important and willing to test extensively on multiple platforms.

lassoan: The model displayable manager is used for rendering slices, annotations, maybe tensor glyphs, etc. and I'm not sure  if all of them are derived from model nodes. I did a quick search in Slicer core and extensions and it seemed to me that all model display nodes were used with nodes derived from model node. If this is actually the case then it should be fine to only observe model related nodes (model display nodes and probably also model nodes).

lassoan:
As discussed, due to planned rework of displayable managers in Slicer-5.13, let's close this PR.

PR #9155, opened 2026-05-13
BUG: Prevent VTK output window from showing on startup on Windows

cpinter:
After updating Slicer, I noticed that on every startup an empty vtkOutputWindow is shown. This is a workaround to prevent that. This was Claude Code's assessment of the issue:

### Problem

Since updating Slicer to the latest revision (which uses VTK 9.6.1 as the default), an empty Win32 popup titled **"vtkOutputWindow"** appears on startup, accompanied by a system sound. No extensions are needed to reproduce this.

### Root cause

There are two compounding issues.

**1. `vtkWin32OutputWindow::DisplayText()` ignores `DisplayMode`**

`vtkWin32OutputWindow` sets `DisplayMode = NEVER` in its constructor (to suppress stdout/stderr on non-dashboard machines), but its `DisplayText()` override unconditionally calls `AddText()` regardless of the display mode:

```cpp
// vtkWin32OutputWindow::DisplayText() — both VTK 9.5 and 9.6
const auto streamtype = this->GetDisplayStream(...);  // NEVER → Null
// ...
this->AddText(someText);   // always called, regardless of streamtype
this->AddText("\r\n");     // always called
// streamtype only gates cout/cerr, NOT the Win32 GUI window
```

`AddText()` calls `Initialize()` which calls `ShowWindow()` — the popup appears for *any* VTK message that reaches the `vtkWin32OutputWindow` singleton. The base class `vtkOutputWindow::DisplayText()` correctly checks `DisplayMode` and does nothing in `NEVER` mode; the Win32 subclass does not.

**2. CTK's VTK message handler is installed too late**

`ctkVTKErrorLogMessageHandler` (which replaces `vtkWin32OutputWindow` with a silent `ctkVTKOutputWindow`) is installed in `qSlicerApplicationPrivate::init()`:

```cpp
this->ErrorLogModel->registerMsgHandler(new ctkVTKErrorLogMessageHandler);
this->ErrorLogModel->setAllMsgHandlerEnabled(true);  // ← VTK output window replaced here
```

However, a large amount of VTK/MRML initialization happens *before* this point in the same function:

| Step | Code | CTK handler? |
|------|------|-------------|
| `ctkVTKConnectionFactory::setInstance(...)` | line 318 | ✗ |
| `new qSlicerPythonManager()` | line 325 | ✗ |
| `this->Superclass::init()` → creates `vtkSlicerApplicationLogic`, `vtkMRMLScene`, initializes Python, etc. | line 334 | ✗ |
| **CTK VTK handler installed** | lines 393–394 | ✓ |

Any VTK message generated during those unguarded steps goes directly to `vtkWin32OutputWindow`.

**3. Why it is new with VTK 9.6**

VTK 9.6.1 (now the default since the March 2026 SuperBuild update) generates at least one new VTK message during the unguarded initialization phase that VTK 9.5.2 did not. A breakpoint on `vtkWin32OutputWindow::AddText()` will identify the exact trigger.

**4. Why the popup appears empty**

VTK error/warning messages start with `\n` (e.g. `"\nERROR: In file.cxx, line N\n..."`). In `DisplayText()`, the leading `\n` causes `Initialize()` to fire (showing the window) before any visible text is added. The actual message content is then queued via asynchronous `PostMessageW(EM_REPLACESEL, ...)` — since the main thread has not returned to the Win32 message loop yet, the edit control remains visually empty when the user first sees the window.

---

### Fix options

**Option A — Slicer (applied here): suppress `vtkWin32OutputWindow` early**

In `qSlicerApplicationHelper::preInitializeApplication()`, replace the default `vtkWin32OutputWindow` singleton with the base `vtkOutputWindow` before any VTK objects are created. The base class properly does nothing in `NEVER` mode.

```cpp
void qSlicerApplicationHelper::preInitializeApplication(...)
{
  vtkLogger::SetStderrVerbosity(vtkLogger::VERBOSITY_OFF);

#ifdef _WIN32
  // vtkWin32OutputWindow::DisplayText() unconditionally shows a GUI popup for any
  // VTK message, ignoring DisplayMode. Replace it early with the base class which
  // properly suppresses output in NEVER mode, until CTK's error log handler takes over.
  {
    vtkNew<vtkOutputWindow> suppressingWindow;
    suppressingWindow->SetDisplayModeToNever();
    vtkOutputWindow::SetInstance(suppressingWindow);
  }
#endif

  itk::itkFactoryRegistration();
  // ...
```

When CTK's handler is later enabled, it saves this benign base-class instance, replaces it with its own, and restores it on shutdown — all transparently.

**Option B — VTK fork: fix the underlying bug in `vtkWin32OutputWindow`**

In `Common/Core/vtkWin32OutputWindow.cxx`, make `DisplayText()` respect `DisplayMode` for the GUI window by returning early when the stream type is `Null`:

```cpp
void vtkWin32OutputWindow::DisplayText(const char* someText)
{
  std::lock_guard<std::mutex> lock(vtkWin32OutputWindowMutex);
  if (!someText) { return; }
  if (this->PromptUser) { this->PromptText(someText); return; }

  const auto streamtype = this->GetDisplayStream(this->GetCurrentMessageType());

  // Respect NEVER display mode — suppress the GUI window, not just stdout/stderr
  if (streamtype == StreamType::Null)
  {
    return;
  }

  // existing AddText / OutputDebugString / cout/cerr logic ...
```

This should be applied to the `slicer-v9.6.1-2026-03-24` branch of [Slicer/VTK](https://github.com/Slicer/VTK).

### DIFF
--- a/Base/QTApp/qSlicerApplicationHelper.cxx
+++ b/Base/QTApp/qSlicerApplicationHelper.cxx
@@ -68,6 +68,7 @@
 #include <vtksys/SystemTools.hxx>
 #include <vtkNew.h>
 #include <vtkLogger.h>
+#include <vtkOutputWindow.h>
 
 // PythonQt includes
 #ifdef Slicer_USE_PYTHONQT
@@ -97,6 +98,18 @@ void qSlicerApplicationHelper::preInitializeApplication(const char* argv0, ctkPr
 #endif
 
   vtkLogger::SetStderrVerbosity(vtkLogger::VERBOSITY_OFF);
+
+#ifdef _WIN32
+  // vtkWin32OutputWindow::DisplayText() unconditionally shows a GUI popup for any
+  // VTK message, ignoring DisplayMode. Replace it early with the base class which
+  // properly suppresses output in NEVER mode, until CTK's error log handler takes over.
+  {
+    vtkNew<vtkOutputWindow> suppressingWindow;
+    suppressingWindow->SetDisplayModeToNever();
+    vtkOutputWindow::SetInstance(suppressingWindow);
+  }
+#endif
+
   itk::itkFactoryRegistration();
   qMRMLWidget::preInitializeApplication();
 


lassoan:
There seems to be two unrelated issues: VTK output window or console window appearing at startup; chime at startup (and a briefly appearing popup window near the center of the screen).

## 1. VTK output window

I don't see a VTK Win32 output window or Windows console appearing in factory-built Slicer Preview Releases at startup.

> A breakpoint on vtkWin32OutputWindow::AddText() will identify the exact trigger.

I added a breakpoint there and nothing triggered it.

There are two console related build options: they control if you want the launcher and the Slicer app to be console applications (console applications bring up a Windows console window). If you choose the launcher to be a GUI app (`Slicer_BUILD_WIN32_CONSOLE_LAUNCHER:BOOL=OFF`) then the Slicer app should not bring up a console (even if you have it on `Slicer_BUILD_WIN32_CONSOLE:BOOL=ON`). Maybe you have `Slicer_BUILD_WIN32_CONSOLE_LAUNCHER:BOOL=ON`?

## 2. Chime at startup

I went back in the release history to find where the chime appeared at startup (along with a small popup appearing in the center of the screen for a fraction of a second):

"3D Slicer 5.11.0-2026-01-20" git hash 439080 => no chime (OK)
"3D Slicer 5.11.0-2026-01-21" git hash 2d2382 => chime (bad)

There are only two commits in this range:

```
Revision: 2d23822723c1ff4c0cebae745ea58b86dfd1b7ec
Author: Davide Punzo <punzodavide@hotmail.it>
Date: 2025-11-21 10:55:22 AM
Message: ENH: Refactor visual DICOM browser UI

Revision: d11fc4c9521c32f2c36d11b8d1175eea8364ce00
Author: Kyle Sunderland <sunderlandkyl@gmail.com>
Date: 2026-01-21 12:37:56 AM
Message: ENH: Add view node flag to disable Markups occlusion checking
```

`ENH: Add view node flag to disable Markups occlusion checking` cannot cause this, while `ENH: Refactor visual DICOM browser UI` is a huge commit, including a CTK update, so something in that commit must be the culprit.

I've applied the changes in this PR and it did not fix the "chime" issue for me.

I've checked the commit content and it seems that changing of CTK\Libs\Widgets\ctkMessageBox.cpp causes the popup appearance and chime.

---

@cpinter Can you confirm that the console window and chime are unrelated?

@Punzo Could you please check why this messagebox change was necessary and how could you achieve what is needed without the incorrect behavior at startup (chime and briefly appearing popup)?

lassoan:
I can confirm that the chime is caused by the Slicer disclaimer/welcome popup. Due to a change in `ctkMessageBox::setVisible(bool visible)` the popup actually appears and chimes and then it gets hidden immediately:

Original:

```
      // Don't call click now, it would destroy the message box. The calling
      // function might expect the message box to be still valid after
      // setVisible() return.
      QTimer::singleShot(0, autoAcceptButton, SLOT(click()));
```

Modified (causes flashing of popup and chime at startup):

```
      // Show the dialog first, then auto-accept it after the event loop processes
      this->Superclass::setVisible(visible);
      QTimer::singleShot(0, this, [autoAcceptButton]() {
        autoAcceptButton->click();
      });
```

@Punzo Please check why this modification was needed.

Punzo:
> @Punzo Please check why this modification was needed.

@lassoan @cpinter 

It was needed for Qt6. At least in linux without that specific change the "Don't show  this message again and always Delete" was not working, i.e.:
at first delete: 
image
second delete (it continues to open the dialog):
image

lassoan:
Thanks for the information, this is very useful! We need to find an alternative solution that does not temporarily shows the popup but works on Qt6.

Does the the new implementation works well in Qt6 (no chime at startup)? If yes, then we could just add `#ifdef`s to use different implementations on Qt5 and Qt6.

Punzo:
> Does the the new implementation works well in Qt6 (no chime at startup)?

I don't have the chime (but it could be just that my pc is fast enough and I don't see it).

On the other hand, I re-tested and the fix for https://github.com/Slicer/Slicer/pull/9155#issuecomment-4453451861 and it seems that it does not work reliably in Qt6. So I am looking into to restore the original version, and see why that fix was partially working and what can be a real solution. I will open a PR soon

Punzo:
@lassoan PR done in https://github.com/commontk/CTK/pull/1420, It was 2 issues related, see https://github.com/commontk/CTK/pull/1420#issue-4448609587

cpinter:
> Can you confirm that the console window and chime are unrelated?

The chime is unrelated. First of all, the output window appears right after the splash screen, and the chime sounds when the main window shows up. Second, with the workaround in this PR there is no output window but the chime is still there.

> Maybe you have Slicer_BUILD_WIN32_CONSOLE_LAUNCHER:BOOL=ON?

I checked (and then double-checked) the build scripts, and I don't think I have any of these enabled. My build script only uses these two flags:
`%CMAKE_EXE% %SLICER_SOURCE_DIR% -G %CMAKE_GENERATOR_X64% -DQt5_DIR:PATH=%QT5_DIR_X64% -DSlicer_USE_SimpleITK:BOOL=ON -DSlicer_USE_VTK_DEBUG_LEAKS:BOOL=OFF`
However, I do have the console window (in both Qt5 and Qt6 builds).

jamesobutler:
@lassoan I was able to replicate seeing the vtkOutputWindow with Slicer versions built by the factory.

1. Clear out %LOCALAPPDATA%\slicer.org
2. Install Slicer 5.10.0
3. Launch Slicer 5.10.0 from the installer final page and load MRHead from Sample Data and close the application
4. Install Slicer 5.11.0-2026-05-11
5. Launch Slicer 5.11.0-2026-05-11 from the installer final page and observe the vtkOutputWindow appears on startup as well as the presence of the following VTK warnings in the python console seen in images below).
6. Then manually delete "%LOCALAPPDATA%\slicer.org\Slicer\cache"
7. Re-launch Slicer 5.11.0-2026-05-11 and there is no more vtkOutputWindow displayed at startup as "%LOCALAPPDATA%\slicer.org\Slicer\cache\SlicerIO\.slicer-cache" is created.
image
image


This newly observed VTK warning is also being observed in Slicer automated testing and causing a test to failure as described at https://github.com/Slicer/Slicer/pull/9115#issuecomment-4425263247. 

Why the vtkOutputWindow appears empty instead of with text printed? Possibly because [Slicer's VTK](https://github.com/Slicer/VTK/tree/slicer-v9.6.1-2026-03-24-d3ad22cd8) doesn't include recent commits to [vtkWin32OutputWindow](https://gitlab.kitware.com/vtk/vtk/-/commits/master/Common/Core/vtkWin32OutputWindow.cxx) including [Fix output window not displaying text](https://gitlab.kitware.com/vtk/vtk/-/commit/4a52e518ba8eec5e6b797f863a86d340e063767f).

lassoan:
Thanks for all the additional information.

It is important that the `vtkOutputWindow` is not a console window, it is a Win32 GUI window displayed by VTK. So, the problem is not related to `Slicer_BUILD_WIN32_CONSOLE_LAUNCHER` ans `Slicer_BUILD_WIN32_CONSOLE` flags, it is just a window created by `vtkWin32OutputWindow`.

The cache warning is new. Now the cache folder size is actually enforced, but only if we are sure that the folder is created by Slicer (to avoid removing files from a folder that the user choses as cache folder, without understanding that the content will be managed - which includes removing old content - by Slicer). It seems that the warning is just logged too early, before CTK installs the log redirection.

Probably the cleanest solution is to register an error sink to the cache manager to collect all of its messages and log them a bit later (after CTK has set up the log redirection). The fix in this PR drops the messages, which is simple, but we may miss seeing relevant information. Or, if the error sink implementation turns out to be complicated then we could just remove this warning message, as it is not essential.

jamesobutler:
Is there a reason the sentinel .slicer-cache file is not being created and triggering the warning when %LOCALAPPDATA%\slicer.org\Slicer\cache already exists and was created by Slicer 5.10.0? Default Slicer cache locations were in use.

%LOCALAPPDATA%\[COMPANY_NAME]\[MAIN_APPLICATION_NAME]\cache

lassoan:
We don't know anything about already existing cache folders, who created it and what is in them. Some users may have assigned their cache folder to `c:\Users\something\Documents` years ago and have been using it without any issues. Slicer cannot just start managing it now - and start deleting old data from it because the folder size is larger than the permitted cache size.

But we could add a special case that if the cache folder is exactly `%LOCALAPPDATA%\slicer.org\Slicer\cache` then we take ownership, we can be quite confident that this folder does not contain valuable user data and it is fully owned by Slicer.

jamesobutler:
Yes, %LOCALAPPDATA%[COMPANY_NAME][MAIN_APPLICATION_NAME]\cache for a Slicer custom application would be good for identifying a Slicer app defined default location. I had also noticed this output window for my Slicer custom app that I was updating to a newer Slicer version.

jamesobutler:
Superseded by #9159.

PR #9158, opened 2026-05-16
BUG: Fix qMRMLNodeComboBox::nodeAddedByUser emitted twice

### LINKED ISSUE #9143: qMRMLComboBox emitting nodeAddedByUser twice
## Summary

When observing the `nodeAddedByUser` of a qMRMLComboBox, the connected slot is called twice 

Likely caused by these two lines:

https://github.com/Slicer/Slicer/blob/c59af2b3c8635df6b0763604c7007d756642f7c0/Libs/MRML/Widgets/qMRMLNodeComboBox.cxx#L610

https://github.com/Slicer/Slicer/blob/c59af2b3c8635df6b0763604c7007d756642f7c0/Libs/MRML/Widgets/qMRMLNodeComboBox.cxx#L453

## Steps to reproduce

```
import slicer

def onFiducialNodeAdded(node):
    print(f"nodeAddedByUser emitted: {node.GetName()} ({node.GetID()})")

combo = slicer.qMRMLNodeComboBox()
combo.nodeTypes = ["vtkMRMLMarkupsFiducialNode"]
combo.selectNodeUponCreation = True
combo.addEnabled = True
combo.removeEnabled = True
combo.noneEnabled = True
combo.setMRMLScene(slicer.mrmlScene)

combo.nodeAddedByUser.connect(onFiducialNodeAdded)
combo.show()
```

Manually add a new node from the drop down.

## Environment
- Slicer version: going all the way back to at least 5.8.1
- Operating system: Windows / Mac

dparikh79:
## Summary

Fixes #9143.

`qMRMLNodeComboBox::nodeAddedByUser` is emitted twice for every user-initiated node creation through the combo box. The duplicate fire comes from `addNode(QString)` itself emitting the signal on every successful return at `Libs/MRML/Widgets/qMRMLNodeComboBox.cxx:610`, on top of the explicit emit at the two user-action call sites:

- `onComboBoxIndexChanged` slot at line 449 calls `addNode(nodeTypeName)` then emits at line 453.
- `createNodeAs` slot (after the rename dialog) at line 701 calls `addNode(nodeTypeName)` then emits at line 706.

This PR removes the emit inside `addNode(QString)`. Both user-initiated call sites still emit exactly once. The Python reproducer in the issue (`combo.nodeAddedByUser.connect(...)` then add a node from the dropdown) now fires the slot once instead of twice.

## Why remove the inner emit, not one of the outer ones

`addNode(QString)` is a public `Q_INVOKABLE` method (declared in `qMRMLNodeComboBox.h:300`). It can be called from Python, from scripted modules, and from C++ that wants to create a node without going through any user gesture. Emitting `nodeAddedByUser` from inside that method means a programmatic add silently fires a signal whose name explicitly says "by user." The two user-action call sites are the right and only place for the signal.

Behavior change scope:

- User combo dropdown "Create new node" path: ONE emit (line 453). Was TWO.
- User combo dropdown "Create new node as..." path: ONE emit (line 706). Was TWO.
- Programmatic `addNode(QString)` callers (Python, C++, scripted modules): ZERO emits. Was ONE.

Net: every user-driven flow keeps emitting exactly once; programmatic flows stop falsely emitting "by user."

## Test plan

- [x] Read `Libs/MRML/Widgets/qMRMLNodeComboBox.cxx` lines 440-470, 595-625, and 685-710 to confirm the call graph above.
- [x] Read `Libs/MRML/Widgets/qMRMLNodeComboBox.h` to confirm `addNode(QString)` is `Q_INVOKABLE` and `nodeAddedByUser` is the documented signal at line 354.
- [x] Confirm via grep there is no other caller of `addNode(QString)` that would lose a needed emit (only the two user-action paths and the no-arg `addNode()` overload at line 615 which itself delegates to the QString overload).
- [ ] **Could not validate locally**: I do not have a working Slicer build environment (full build is hours). I am relying on the existing CI to confirm the change compiles and passes the existing widget test suite (`Libs/MRML/Widgets/Testing/qMRMLNodeComboBoxTest{1..5}.cxx` plus `qMRMLNodeComboBoxLazyUpdateTest1.cxx` and `qMRMLNodeComboBoxEventTranslatorPlayerTest1.cxx`). None of those existing tests assert emission count on `nodeAddedByUser`, so they will not catch a regression here either way; adding a counting test would require either a full Slicer build or a more focused QSignalSpy-based unit. Happy to add one in a follow-up if a maintainer prefers.
- [ ] @che85, if you can run your Python reproducer from the issue against this branch, the connected slot should print exactly once now.

cc: @che85

## AI Assistance Disclosure

Analysis and the patch were drafted with Claude assistance. I traced the call graph by hand against the repo's source, verified the public Q_INVOKABLE surface of `addNode`, and reviewed every changed line. I am the human contributor accountable for this PR.

### DIFF
--- a/Libs/MRML/Widgets/Testing/qMRMLNodeComboBoxTest5.cxx
+++ b/Libs/MRML/Widgets/Testing/qMRMLNodeComboBoxTest5.cxx
@@ -28,10 +28,13 @@
 
 // qMRML includes
 #include "qMRMLNodeComboBox.h"
+#include "qMRMLSceneModel.h"
 
 // MRML includes
 #include <vtkMRMLCameraNode.h>
+#include <vtkMRMLNode.h>
 #include <vtkMRMLScene.h>
+#include <vtkMRMLScalarVolumeNode.h>
 
 // VTK includes
 #include <vtkNew.h>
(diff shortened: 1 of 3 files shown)


lassoan:
Thank you @dparikh79 for working on this. This would be a breaking change. Even if the current implementation is not consistent with naming or documentation, we need to assess any impact in Slicer core and all extensions in the extensions index. If there are modules that only observe `nodeAddedByUser` signal and expect to be notified when `addNode` is called then those modules need to be updated to observe `nodeAdded`. Alternatively, we could go through a long deprecation process (or we need to introduce a new signal, such as `nodeAddedOnGUI`, deprecate `nodeAddedByUser`, then after a few years remove `nodeAddedByUser`, then after a couple of years we can add back `nodeAddedByUser` with the new behavior and deprecate `nodeAddedOnGUI`, then after a few years remove `nodeAddedOnGUI`). Or we can just fix the duplicate emit issue (emit `nodeAddedByUser` only in `addNode`) and explain the behavior (and unintuitive name) in the documentation.

Could you help with reviewing all usages of `nodeAddedByUser` even in Slicer core and all extensions to see if there is any module that would expect to be notified about a node being added in general (not just via the GUI)?

dparikh79:
Thanks @lassoan and @Copilot for the careful review.

Copilot's catch is correct: `qMRMLNodeComboBoxEventPlayer::playEvent` calls `comboBox->addNode()` (the no-arg overload, which delegates to `addNode(QString)`) when replaying a recorded `nodeAddedByUser` event (`Libs/MRML/Widgets/qMRMLNodeComboBoxEventPlayer.cpp:52`). Before this PR, that path re-emitted the signal; after this PR, replay silently does not, so any recorded workflow or test that depends on the replay-then-observe pattern would regress. I should have grepped the wider tree before sending; my bad.

@lassoan, you are right that this is a breaking change in the public signal contract even if the documentation/name says otherwise. Three ways to land this cleanly, all bigger than the current one-line PR:

**A. Split the API surface** (architecturally cleanest)

Introduce a new method on `qMRMLNodeComboBox`:

```cpp
// New: user-action add. Emits nodeAddedByUser. Use this from any
// user-initiated path (or from EventPlayer when replaying a user action).
Q_INVOKABLE virtual vtkMRMLNode* addNodeByUser(QString nodeType);
Q_INVOKABLE virtual vtkMRMLNode* addNodeByUser();
```

`addNodeByUser` internally calls `addNode` and emits. `onComboBoxIndexChanged`, `createNodeAs`, and `qMRMLNodeComboBoxEventPlayer::playEvent` all migrate to `addNodeByUser`. `addNode(QString)` keeps its current behavior of NOT emitting (preserving the semantic the name suggests) but is now safe for programmatic callers. No extension breaks today because every existing caller of `addNode()` still gets exactly the same observable behavior they got before this PR (since the emit-on-addNode would now happen via `addNodeByUser` at the user-action call sites).

**B. Minimal: EventPlayer emits explicitly**

Keep this PR's change; in `qMRMLNodeComboBoxEventPlayer::playEvent`, after `comboBox->addNode()`, call `comboBox->nodeAddedByUser(node)` directly (or expose a friend / `Q_INVOKABLE emitNodeAddedByUser` helper). Two-line change in the EventPlayer, no public-API additions. Behavior still changes for any extension that was relying on the side-effect emit through `addNode()` programmatic calls, but the EventPlayer regression is closed.

**C. Deprecation path** (most conservative)

Revert this PR. Add a Qt5/6-compatible deprecation warning when `addNode(QString)` emits, with a clear migration note pointing at a new `addNodeByUser` (option A). Drop the emit in a future release (Slicer 5.12 or 6.0). This is the safest for the extensions index but it is a multi-release effort and obviously much bigger than what I came in with.

Open to whichever direction you prefer. Happy to do an audit of the extensions index (or a subset such as SlicerIGT, SlicerRT, SlicerExtension-VMTK, etc.) for direct usages of `addNode()` so you have a concrete impact estimate before you decide. If you would rather close this PR and revisit, that is also fine; I will not push more changes until you signal direction.

lassoan:
Option B would be nice, but the description is not accurate. The impact analysis does not take into account all the usages of these signals outside the Slicer core repository in extensions and private extensions and private Slicer Custom Applications  Please ask an LLM to review all the usage of this signal in Slicer core and all the extensions. If there are places where only `nodeAddByUser` is connected then it may need manual review of the developer intent (I'm not sure if the LLM can reliably determine that) and and submission of pull requests with a fix. The easiest way to get the source code of all extensions is to [build the "ExtensionsIndex"](https://slicer.readthedocs.io/en/latest/developer_guide/extensions.html?utm_source=chatgpt.com#extensions-build-system). If the change would only break a few extensions then we can fix those. There is still some risk that we break Slicer Custom Applications but if we highlight the change in the release notes then that may be sufficient.

I think removing the duplicate emit without change in behavior would be a reasonable tradeoff (not too much effort for minimal gain).

dparikh79:
Done the impact analysis. Two findings, one of which changes the proposed approach materially.

## Impact analysis (extensions index)

Searched GitHub for `nodeAddedByUser` across Slicer extensions and community repos. Filtered out forks of Slicer itself. 12 real extensions reference the signal. For each, checked: does the extension call `addNode()` programmatically AND connect a listener to `nodeAddedByUser`?

| Extension | calls `addNode()` | listener on `nodeAddedByUser` | Breaks under previous Option B? |
|---|---|---|---|
| ALive-research/Slicer-Liver | yes | yes | **YES** |
| HOA-2/SlicerNeuroSegmentation | no | yes | no |
| KitwareMedical/Slicer-CameraPath | no | yes | no |
| ManskeLab/Bone_Analysis_Modules | no | yes | no |
| PerkLab/SlicerSandbox | no | yes | no |
| SlicerDMRI/SlicerDMRI | no | yes | no |
| SlicerIGT/SlicerIGT | no | yes | no |
| SlicerRt/SlicerRT | no | yes | no |
| biomedisa/biomedisa | no | yes | no |
| lchauvin/BetaProbe | no | yes | no |
| mikebind/Slicer_ACPC_Thal | no | no | no |
| vmtk/SlicerExtension-VMTK | no | yes | no |

`ALive-research/Slicer-Liver/LiverSegments/LiverSegments.py:139` connects `endPointsMarkupsSelector.nodeAddedByUser` to `self.newEndpointsListCreated`; line 219 then calls `self.ui.endPointsMarkupsSelector.addNode()` on the SAME combobox. Under my previous PR + Option B, that callback would silently stop firing on the programmatic add, and the markups node would be created without its initialization step.

## Also: Slicer core itself has the same pattern

`Modules/Loadable/CropVolume/qSlicerCropVolumeModuleWidget.cxx:179` calls `this->InputROIComboBox->addNode();` programmatically; line 262 connects that combobox's `nodeAddedByUser` to `initializeInputROI`. Same silent-breakage pattern. So even before talking about extensions, my previous PR + Option B would have broken CropVolume's automatic ROI initialization.

## Revised proposal: remove the OUTER emits, not the inner one

Re-reading the source with the above in mind, I was removing the wrong emit. The duplicate-emit on the user-action paths comes from BOTH:

1. `addNode(QString)` emitting at line 610 (every call), AND
2. The user-action callers explicitly emitting again at lines 453 (`onComboBoxIndexChanged`) and 706 (`createNodeAs`).

If I instead **delete the two outer emits at lines 453 and 706** and leave the inner emit at line 610 untouched, the table becomes:

| Path | Before | After |
|---|---|---|
| User picks "Add new node" from dropdown | 2 emits | 1 emit |
| User picks "Add new node as..." (dialog) | 2 emits | 1 emit |
| Programmatic `addNode(QString)` | 1 emit | 1 emit (unchanged) |
| Programmatic `addNode()` no-arg | 1 emit | 1 emit (unchanged) |
| EventPlayer replay | 1 emit | 1 emit (unchanged) |
| CropVolume `InputROIComboBox->addNode()` | 1 emit | 1 emit (unchanged) |
| Slicer-Liver `endPointsMarkupsSelector.addNode()` | 1 emit | 1 emit (unchanged) |

Net: user-action duplicate is gone, every other caller is bitwise unchanged. No extension impact, no core impact, no deprecation needed.

If you agree, I will revert the current single-line change and push the corrected version (delete lines 453 and 706, leave 610 alone). Will mention this thread in the new commit message so the history is clear.

lassoan:
Sounds good to me. We need to add to the signal's documentation that the name is not accurate (and give explanation why is that). Copy all important additional information from this thread to the commit comment, instead of just referencing this thread, as commit comments are guaranteed to be kept with the source code forever, while github content may become difficult to access in the future.

dparikh79:
Thanks @lassoan. Force-pushed the revised approach in `70d6301`.

Summary of changes vs the prior commit:

- Reverted the inner-emit removal in `qMRMLNodeComboBox::addNode(QString)`; the canonical single emit now lives there.
- Removed the duplicate emits at the two outer user-action call sites: `onComboBoxIndexChanged` (the add-node branch) and `createNodeAs` (after the dialog returns). Each leaves an inline note pointing back to this issue.
- Extended the doxygen comment on `nodeAddedByUser` in `qMRMLNodeComboBox.h` to document that the signal is a misnomer: `addNode` is `Q_INVOKABLE` and `nodeAddedByUser` fires for every successful invocation (user-triggered or programmatic), and renaming the signal would break ABI for every connected slot in the ecosystem, so the name is retained for backward compatibility.
- Rewrote the commit message to capture the full context from this thread (root cause, why outer-emit removal over inner-emit removal, the two concrete in-tree / extension cases that drove the approach change) so future archaeologists do not have to chase the GitHub thread.

The behavior contract is now:
- A user adding a node via the combo box: `nodeAddedByUser` fires exactly once (was: twice).
- A programmatic `addNode(...)` caller (e.g. CropVolume, Slicer-Liver): `nodeAddedByUser` fires exactly once, identical to the pre-PR behavior.
- A programmatic `createNodeAs(...)` caller: `nodeAddedByUser` fires exactly once (was: twice).

Let me know if you want any other clarification in the comments or the commit message.

lassoan:
@dparikh79 I've tweaked your commit: added a new `nodeAddedByUserAction` action that is emitted only when the user adds the node, deprecated `nodeAddedByUser` due to the misleading name, and added a test. Could you please test if everything works well for you? Thanks!

lassoan (approved):
This looks good now.

PR #9159, opened 2026-05-17
Fix CacheManager startup warning

lassoan:
Initialize application logger earlier to capture more messages during startup.
Only log warning about missing cache folder sentinel file if the cache folder location is non-default.

see #9155


jamesobutler (approved):
Thanks. I built locally and can confirm this resolves the issue with the vtkOutputWindow appearing on startup corresponding to the sentinel cache warning. I observe the sentinel cache file gets corrected for my standard Slicer cache directory location that was already present from a Slicer 5.10.0 usage. I also locally ran the automated test that was updated and can confirm vtkSceneViewImportSceneTest passes now too.

lassoan:
Thank you for the review and testing, I'll merge this now.

PR #9162, opened 2026-05-18
Fix test failures

lassoan:
This pull request fixes all failing automated tests on Windows:
237 - qMRMLLinearTransformSliderEventTranslatorPlayerTest1 (Failed) qMRMLWidgets
246 - qMRMLSliceWidgetEventTranslatorPlayerTest1 (Failed) qMRMLWidgets
509 - MergeModelsTestCompare (Failed)                   MergeModels
590 - py_nomainwindow_SlicerOptionDisableSettingsTest (Failed)
609 - py_nomainwindow_test_slicer_packaging (Failed)


jamesobutler (approved):
Thanks for addressing these. The changes make sense. I've run updated tests locally on Windows and can confirm these ones are now resolved.

ebrahimebrahim:
Thank you for fixing these! To me, James's review is sufficient, but if this remains unmerged by the time I get to it in my task log then I will be running local tests on linux as well

lassoan:
Thank you, let's get these merged then and the we'll see the results on the dashboard. If there is any unexpected behavior then we can fix in follow-up commits.

PR #9163, opened 2026-05-18
BUG: Fix segment metadata off-by-one when loading DICOM Labelmap SEG

fedorov:
DICOM Labelmap Segmentation objects reserve pixel value 0 for background and may include it as a defined segment (SegmentNumber 0) in the segment sequence. The DICOMSegmentationPlugin was assigning metadata positionally from the JSON produced by segimage2itkimage, which lists the background entry first. Since ImportLabelmapToSegmentationNode does not create a VTK segment for pixel value 0, every subsequent segment received the metadata of its predecessor (e.g. skin appeared as costal cartilage).

dcmqi 1.5.4 labelmap conversion will generate a segment entry for the background (introduced in https://github.com/QIICR/dcmqi/pull/536, leading to shift of metadata assignment.

Fix: skip any segment entry with labelID == 0 before building the labelAttributes list.

### DIFF
--- a/Modules/Scripted/DICOMPlugins/DICOMSegmentationPlugin.py
+++ b/Modules/Scripted/DICOMPlugins/DICOMSegmentationPlugin.py
@@ -152,6 +152,15 @@ class DICOMSegmentationPluginClass(DICOMPlugin):
           # metadata. ImportLabelmapToSegmentationNode does not create a VTK segment for pixel
           # value 0, so including this entry would shift all subsequent metadata assignments by one.
           if segment.get("labelID", 1) == 0:
+            typeCode, typeCodingScheme, typeCodeMeaning = \
+              self.getValuesFromCodeSequence(segment, "SegmentedPropertyTypeCodeSequence")
+            if typeCode != "125040" or typeCodingScheme != "DCM":
+              logging.warning(
+                f"Segment with labelID 0 has unexpected type code: "
+                f"{typeCodingScheme}:{typeCode} ({typeCodeMeaning}). "
+                f"Expected DCM:125040 (Background). "
+                f"This segment will not be loaded as a Slicer segment.",
+              )
             continue
 
           try:
--- a/Modules/Scripted/DICOMPlugins/DICOMSegmentationPlugin.py
+++ b/Modules/Scripted/DICOMPlugins/DICOMSegmentationPlugin.py
@@ -147,6 +147,13 @@ class DICOMSegmentationPluginClass(DICOMPlugin):
         labelNode.labelAttributes = []
 
         for segment in segmentAttributes:
+          # Skip the background segment (labelID 0). In DICOM Labelmap Segmentation objects,
+          # pixel value 0 is reserved for background and may appear as a defined segment in the
+          # metadata. ImportLabelmapToSegmentationNode does not create a VTK segment for pixel
+          # value 0, so including this entry would shift all subsequent metadata assignments by one.
+          if segment.get("labelID", 1) == 0:
+            continue
+
           try:
             rgb255 = segment["recommendedDisplayRGBValue"]
             rgb = [float(c) / 255. for c in rgb255]


michaelonken:
> DICOM Labelmap Segmentation objects reserve pixel value 0 for background

I think in general it can be any segment. For our labelmap exports, its always the the segment 0 (i.e. pixel value 0) since it is not used for the actual labels.

As the labelmap supplement says:

"2. The need to describe every pixel value implies that the value 0, if used, will also be described, which might
serve as the background"

So in the beginning I forgot about this; I have set all labels in the pixel data and the residual ones stayed at the initialization value that I have used: 0 . Later I realized that I have encode this explicitly.

I think it can also be any other segment, the rule is really just that every pixel value in the pixel data must be described (boiling down to: must have assigned a related segment).

lassoan (dismissed):
Thank you, the change looks good to me.

lassoan:
@fedorov Is it possible that there are gaps between labelIDs or we may get negative labelID?

lassoan:
> I think it can also be any other segment, the rule is really just that every pixel value in the pixel data must be described

It seems that indeed it can be any value, not just 0, but the value that is specified in [pixel padding value attribute](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_c.8.20.2.4.html). However, until we encounter clinical images that use non-zero padding value, we don't need to implement support for reading segmentations with non-zero pixel padding value. Instead, it may be enough if we log a warning if padding value is undefined or non-zero.

michaelonken:
I think Pixel Padding Value alone is not sufficient; this is my take on the issue after reading a bit:

> For a Segmentation Type (0062,0001) of LABELMAP, **if a value is specified in Pixel Padding Value (0028,0120)**, that indicates that the corresponding Segment Number is treated as background; 

and [here](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_A.51.4.html) it says:

> Pixel Padding Value (0028,0120) shall not be present if Segmentation Type (0062,0001) is not LABELMAP. Pixel Padding Value (0028,0010) **may be used** to indicate that a particular segment is background if Segmentation Type (0062,0001) is LABELMAP

and (finally, sorry) in the [explanation of Pixel Padding value (scroll down)](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.7.5.html#table_C.7-8) it says:

> For Segmentations with a Segmentation Type (0062,0001) of LABELMAP, Pixel Padding Value (0028,0120) may be used to indicate that a particular pixel value is to be treated as background. Only a single pixel value is so indicated, and Pixel Padding Range Limit (0028,0121) is not used; see [Section A.51.4](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_A.51.4.html). Typically the Pixel Padding Value (0028,0120) will be 0, but it can be another value, **or may not be specified at all**.

This altogether sounds to me that you might indicate background pixel value by using Pixel Padding value **but you dont have to.** . Together with the rule that every pixel must be backed up by a related segment (in the Segment Sequence), one must also issue the warning if one of these Segments is marked (via code) as "background" **but Pixel Padding Value is not set at all.**

Do you agree?

So IMHO one possible approach would be:
- If a pixel value is denoted as background (using one or both of above indicators):
  - If that pixel value is 0: print a warning
  - If value is != 0: print an error [another option is to reorder segments to move the background to pixel value 0, but thats maybe not worth the effort for now since it may rarely occur]
- If we dont have an error, write ITK labelmap but don't export background segment (for label/pixel value 0) into the JSON metadata.

fedorov:
> is it possible that there are gaps between labelIDs or we may get negative labelID?

@lassoan thank you for bringing this up! Based on my understanding, LABELMAP allows gaps in `SegmentNumber` values (see https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.8.20.2.4.html). BUT dcmqi currently will enforce consecutive numbering - I think this will need to be fixed, issue created: https://github.com/QIICR/dcmqi/issues/537.

I have to say, after spending the time to read various sections of the standard, I cannot say that the guidance strikes me as logical and consistent. But given my understanding:
1. background must be described as a separate segment - no matter whether padding value is indicated or not
2. padding value is not required to be present ("Typically the Pixel Padding Value (0028,0120) will be 0, but it can be another value, or may not be specified at all.")

So IMHO here we are again in a situation common in DICOM - there are too many possibilities of how something can be implemented. Given that background segment must be present in the segment sequence, I am leaning towards using just that - as indicated by DCM "background" code (and let's keep fingers crossed implementations will actually use that code!) - to determine which segment to ignore while loading into Slicer, and not look at the padding value at all. 

But I think it would be good to hear thoughts from Chris Bridge and David Clunie on this. I will reach out to them via email to make sure.

fedorov:
Sharing some notes from an AI-assisted analysis of the DICOM standard, generated using [Claude Code](https://claude.ai/code) (model: `claude-sonnet-4-6`, timestamp: 2026-05-20) against the dcmqi codebase and the DICOM standard. References are all primary standard sources; analysis is Claude own interpretation.

---

### On whether the background value must be described in the Segment Sequence

@fedorov's reading ("background must be described as a separate segment no matter whether padding value is indicated") is the conservative interpretation, but I think there is genuine ambiguity in the standard that is worth flagging.

**The tension:**

[C.8.20.2.3](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.8.20.2.3.html) says:
> *"For a Segmentation Type of LABELMAP, every pixel value **actually encoded** in Pixel Data is required to be described in an Item of Segment Sequence."*

[C.7.5 table C.7-8](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.7.5.html#table_C.7-8) says:
> *"No pixels in the **native image** will have a value equal to Pixel Padding Value."*

The second statement is the traditional CT/MR padding semantics: padding pixels (e.g., outside the circular FOV) are physically in the rectangular pixel array but are not considered "native image" data. If that semantics carries over to LABELMAP, then pixels with the Pixel Padding Value are not "actually encoded" image data, and would be **exempt** from the Segment Sequence requirement.

This is not a hypothetical reading — an intermediate draft of Supplement 243 explicitly raised it as an open question:
> *"The document raises the question of whether Pixel Padding Value or a similar mechanism could be re-used in some way to **avoid the need to describe a pixel value**. This would need a change to the constraint in PS3.3 A.51.4."*

The fact that a standard change was discussed (but not definitively described as made or not made in the text I could access) leaves the question open. It is plausible that the final standard's intent is that Pixel Padding Value *does* exempt the background from requiring a Segment Sequence entry.

A parallel piece of evidence: C.8.20.2.3 has this note:
> *"The need to describe every pixel value implies that the value 0, if used, **will also be described**, which might serve as the background."*

The phrase "if used" suggests that pixel value 0 might not always be considered "used" (i.e., might be treated as outside the native image), and that describing it is an implication of it being "used" rather than an absolute requirement.

**Practical upshot:** The standard is not fully clear. Describing the background value as a Segment Sequence item (dcmqi's current approach) is the safe/conservative choice. But the standard may not require it when Pixel Padding Value is set. Reaching out to Chris Bridge and David Clunie (as @fedorov planned) is the right call.

---

### On segment numbering for LABELMAP

Separately, [C.8.20.2.4](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.8.20.2.4.html) is explicit: the requirement that segment numbers *"shall start at 1 and increase monotonically by 1"* applies only to **BINARY and FRACTIONAL** types, not LABELMAP. For LABELMAP, the only universal requirement is uniqueness within the instance. So gaps in segment numbers are permitted by the standard for LABELMAP — dcmqi's current enforcement of consecutive numbering (in `useLabelIDAsSegmentNumber=true` mode) is more restrictive than required. The dcmqi issue @fedorov linked ([QIICR/dcmqi#537](https://github.com/QIICR/dcmqi/issues/537)) captures this correctly.

---

**Standard sections referenced:**
- [PS3.3 C.8.20.2.3 — Segmentation Type, Fractional Type and Segments Overlap](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.8.20.2.3.html)
- [PS3.3 C.8.20.2.4 — Segment Number](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.8.20.2.4.html)
- [PS3.3 C.7.5 table C.7-8 — General Equipment Module / Pixel Padding Value](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.7.5.html#table_C.7-8)
- [PS3.3 A.51.4 — Segmentation IOD / LABELMAP constraints](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_A.51.4.html)
- [Supplement 243 Final Text — Label Map Segmentation](https://dicom.nema.org/medical/dicom/final/sup243_ft_LabelMapSeg.pdf)

CPBridge:
I remember some discussion about this during the drafting stage. While I pushed for a simpler approach, if memory serves, David preferred this one since it would allow any existing "labelmap" style array in another format to be converted to DICOM regardless of whether segments were numbered consecutively or which segment number was used as background in the original encoding.

My understanding is if PixelPaddingValue is present it gives the value of a segment that should be considered background. If it is not present there is no class that should be considered background, i.e. the labelmap completely segments the image into named foreground labels (a true "total segmentation" situation worthy of the name). So if you are converting this to another format, you might want to skip the segment whose number matches PixelPaddingValue, if PixelPaddingValue is present

FWIW When I implemented it in highdicom I took the following approach when encoded labelmaps segs. Class 0 is always the background and highdicom adds this to the segment sequence (using the DCM Background code) in addition to whichever segment information the user passes about non-background classes and sets the PixelPaddingValue to 0. If the user tries to use segment number 0 for some other class, currently highdicom will refuse to do it (though the standard would allow it).

michaelonken:
Hi Chris,

> My understanding is if PixelPaddingValue is present it gives the value of a segment that should be considered background. If it is not present there is no class that should be considered background, i.e. the labelmap completely segments the image into named foreground labels (a true "total segmentation" situation worthy of the name). So if you are converting this to another format, you might want to skip the segment whose number matches PixelPaddingValue, if PixelPaddingValue is present

I think the standard however does not require Pixel Padding to be present in case of background, its just optional, while it (DICOM) requires every pixel value to be represented by a Segment. So it is desirable that Pixel Padding Value is present and if not that there is no background, but you cannot rely on it, as I understand. So to be safe one would need to check the Segments for a background (coding) as well I guess.

CPBridge:
@michaelonken I'm afraid I disagree :)

From [C.8.20.2.4](https://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_C.8.20.2.4.html)

> For a Segmentation Type (0062,0001) of LABELMAP, if a value is specified in Pixel Padding Value (0028,0120), that indicates that the corresponding Segment Number is treated as background [...] Otherwise, no particular value of Segment Number is deemed to be treated specially as background.

This (especially the last sentence) to me implies that no pixel padding value indicates no background

michaelonken:
No need to be afraid :-D

Hm, I didn't read it that way. I read it as: If Pixel Padding Value is present, then its background. If not, we don't know. 

I understood from the sentence
> Otherwise, no particular value of Segment Number is deemed to be treated specially as background.
that it just makes sure that you cannot derive from a specific value alone (i.e. 0) that it defines the background.

[edit] that there is no other way to know from the pixel value alone that it should be treated as background. [/edit]
But maybe my English language skills and understanding led me into the wrong direction here :-)

michaelonken:
After the discussion, I would like to do the following:

1. I will implement only Pixel Padding Value as *the* required and sufficient indicator that background is present. I.e. if there is a background-coded Segment but no Pixel Padding Value set, the segment will be treated as any other segment.
2. For files that dcmqi creates itself, background, if existing, will always use pixel value 0 and will always have Pixel Padding Value set accordingly.
3. dcmqi will also process files that have Pixel Padding Value set to a different value and print a warning when exporting those files to ITK. Despite the warning it will export all labels/segments (including background). A later option might be to reorder the segments and update the pixel data on export (to put background to 0), if actually those kind of files actually show up in practice.

fedorov:
@michaelonken for dcmqi I suggest we also do the following:
1. user is not required to descibe label 0 in the input
2. if user does not describe label 0 in the input, it is automatically marked as padding value
3. DICOM SEG that includes padding value do not produce an item in the segments list corresponding to the label marked as padding.

In Slicer I plan to implement the following logic:
1. segment that has label 0 in JSON as produced by dcmqi will not appear in Slicer interface after loading segmentation
2. if label 0 segment has a code different from "background", a warning will be logged

fedorov:
> For files that dcmqi creates itself, background, if existing, will always use pixel value 0 and will always have Pixel Padding Value set accordingly.

@michaelonken I do not think we want to change what user passes to dcmqi. Unless you meant that if user does not say anything about label 0, dcmqi will treat it as background. But there should be nothing to prevent user from using 10 as background, or even marking multiple segments as background!

fedorov:
> Is it possible that there are gaps between labelIDs or we may get negative labelID?

@lassoan I don't think gaps in labelID should cause a problem with the current code. While creating segmentation, it iterates over all segment metadata entries and I think we can assume that those are ordered in the increasing order. Do you see any issues from the segmentation node initialization side?

I added the background warning I mentioned earlier, so this is ready for the review I think. @jamesobutler updated dcmqi python package version to latest 0.4.0, which this PR will require.

michaelonken:
> > For files that dcmqi creates itself, background, if existing, will always use pixel value 0 and will always have Pixel Padding Value set accordingly.
> 
> @michaelonken I do not think we want to change what user passes to dcmqi. Unless you meant that if user does not say anything about label 0, dcmqi will treat it as background. But there should be nothing to prevent user from using 10 as background, or even marking multiple segments as background!

Thats fine; I was just referring to the files we really create ourselves from scratch or via conversion from DICOM binary segmentations; those will always choose 0 as the background "color".

If we read from ITK + JSON Metadata dcmqi will use whatever the user provides (if not otherwise broken of course).

pieper (approved):
I was a bit hesitant to jump in on this one after all the attention it has already gotten!  I'm fine with the suggested change, since as everyone has said these non-zero backgrounds are probably unlikely and already well handled.

I have to chime in though and ask why a segment with a zero value would automatically be discarded.  Based on the discussion of the standard it sounds like it is legitimate for the 0 value to have a non-background meaning.

So I'd lean towards keeping the warning (since this is unusual and maybe error prone behavior), but loading the 0 valued segment (with a different label number) and with the coded concept and other data that came from the seg file.  

Did I miss some reason why we shouldn't load this segment?

fedorov:
> but loading the 0 valued segment (with a different label number) and with the coded concept and other data that came from the seg file.
> Did I miss some reason why we shouldn't load this segment?

In practice, from what I have seen, 0 is always background. The reason is to replicate behavior to be the same as when loading a non-DICOM NIfTI etc segmentation - which typically would be the input to create the labelmap - background is not shown as a segment in the segments list. 

I do not like loading background as a different label number for several reasons:
1. it will result in a label shown to the user but not present in the input DICOM, and not present in the non-DICOM segmentation used to create DICOM.
2. we may/will end up in a situation where a segmentation has both 0 label background (created by processing) and some other label background. Propagating proper handling of those will be difficult across tools and conventions used in Slicer.

I think if we want to do it right, then Slicer would need to recognize label 0 for any segmentation being loaded, and allocate a segment to that label, and always explicitly label 0 label as background. But doing that is definitely beyond this PR - if there is agreement to do it at all.

fedorov:
I merged the PR before it again needed to be rebased due to another PR being merged, and I would need someone to approve it again! But I am happy to continue the discussion and revisit the implementation based on it.

pieper:
@fedorov I think you may have missed the point I was making in my review comment.  I don't really want to belabor it, since it's not all that likely, but just for the sake of completeness, the situation I had in mind was that someone creates a segmentation that uses the label value of 0 and gives it a coded concept for something other than background.  The just merged code will ignore that segment with a warning.

My suggestion was that we load the segment and it's coded concept using a different label value (even if we keep the warning in place).

I do agree that the strong convention is that 0 means background, but the text of the standard says that some labelmaps have different conventions (or could have different conventions) so I'm not sure why ignoring the 0 value with a different meaning on input is a good idea.

fedorov:
As Claude likes to say, "you are right to keep pushing back on this".

I am not saying it is a good idea, IMHO it is just a practical compromise around the idiosyncrasy of the standard (where several voices advocated strongly against explicitly listing background as a segment in LABELMAP, but were ignored), and limitation of Slicer (where label 0 in the input is ignored).

If we create a new segment from label 0, with a different label, this will break fidelity between what is in Slicer and what is in the data, and maintaining fidelity of label ID assignment is one of the key reasons for using LABELMAP (BINARY SEG does NOT explicitly call out the background). Further, in my experience, it is common among unscrupulous implementations of DICOM SEG to gloss over code assignment ("Tissue" is probably the most common "catch-all" code assigned to "stuff" folks encode in DICOM SEG). Based on my experience, I would not be surprised at all if we have label 0 assigned non-background code, or background stated in a variety of invalid ways. 

The issue you are describing is valid. But it is not easy to resolve, and it is present outside the scope of this PR. If someone is using NRRD and decided to encode something as label 0, they would not be able to load it into Slicer. And they would not have a warning.

All of that is to say: the proper, or perhaps the only way to solve this issue is for Slicer to load segment 0, no matter what it is claimed to be. Alternatives:
1. reassign the label and load as non-0; cons: may actually be background nevertheless, breaks consistency with the source
2. refuse to load SEG if label 0 is assigned a code different from DCM "Background"; cons: false-positive detection of non-background
3. inventive hack I just came up with: offer a separate loadable for label 0 if it is assigned code different from DCM "Background", disabled by default?

lassoan:
It is a reasonable assumption that label 0 is always background and background is always label 0 in a DICOM Labelmap SEG. We can revise this if we experience that this is often not the case in DICOM images that people try to import.

We should just describe this limitation in the documentation, which we need to update anyway to reflect that QuantitativeReporting extension is no longer needed for DICOM Seg import/export. Ideally, we should add a DICOM Conformance Statement style page to the documentation.

fedorov:
Andras, please let me know where I should contribute to it, and I will look into it. Or better please make a ticket so this does not get lost!

PR #9168, opened 2026-05-19
DOC: Add two commercial products in the about documentation page

cpinter:
As a follow-up of
https://github.com/Slicer/slicer.org/pull/385
can we please add these two products in the Slicer documentation as well? Thanks a lot!


ebrahimebrahim (approved):
Looks good

cpinter:
Thank you from the super fast review!

PR #9169, opened 2026-05-19
BUG: Fix crash when selecting active scalars for several model nodes at once

lassoan:
In Models module, if multiple models were selected in the tree and the user selected a different scalar name in Display/Scalars section then Slicer crashed. The problem was that setActiveScalarName method iterated through d->CurrentDisplayNodes, which was modified during the iteration due to Modified callbacks.

Fixed it by making a copy of the list everywhere where we iterate through the list.

As a preventive action (and to reduce unnecessary updates), setMRMLDisplayNodes method was updated to not modify d->CurrentDisplayNodes if it is already up-to-date.

Thanks to @che85 for reporting and the excellent instructions for reproducing it.

### DIFF
--- a/Libs/MRML/Widgets/qMRMLScalarsDisplayWidget.cxx
+++ b/Libs/MRML/Widgets/qMRMLScalarsDisplayWidget.cxx
@@ -138,7 +138,10 @@ QList<vtkMRMLDisplayNode*> qMRMLScalarsDisplayWidget::mrmlDisplayNodes() const
   QList<vtkMRMLDisplayNode*> displayNodes; // this list will only contain valid (non-null) display node pointers
   for (vtkMRMLDisplayNode* displayNode : d->CurrentDisplayNodes)
   {
-    displayNodes << displayNode;
+    if (displayNode)
+    {
+      displayNodes << displayNode;
+    }
   }
   return displayNodes;
 }
@@ -149,6 +152,24 @@ void qMRMLScalarsDisplayWidget::setMRMLDisplayNodes(QList<vtkMRMLDisplayNode*> d
   Q_D(qMRMLScalarsDisplayWidget);
 
   displayNodes.removeAll(nullptr);
+
+  // Return if the list of display nodes did not change to avoid unnecessary updates.
+  if (displayNodes.size() == d->CurrentDisplayNodes.size())
+  {
+    bool same = true;
+    for (int i = 0; i < displayNodes.size(); ++i)
+    {
+      if (displayNodes[i] != d->CurrentDisplayNodes[i])
+      {
+        same = false;
+        break;
+      }
+    }
+    if (same)
+    {
+      return;
+    }
+  }
 
   // Only the first display node is observed
   qvtkReconnect((d->CurrentDisplayNodes.size() > 0 ? d->CurrentDisplayNodes[0] : nullptr),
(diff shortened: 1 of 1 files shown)


ebrahimebrahim:
Can you link to the instructions on how to reproduce the bug?

lassoan:
> In Models module, if multiple models were selected in the tree and the user selected a different scalar name in Display/Scalars section then Slicer crashed.

Load multiple models with multiple point scalar arrays. Go to models module, select multiple models, select a different array in Display/Scalars section -> crash.

ebrahimebrahim:
Ah, I thought you had a convenient way of creating models that have multiple scalars. Looking into this soon

lassoan:
> Ah, I thought you had a convenient way of creating models that have multiple scalars. Looking into this soon

I have data to reproduce, but it is from patient image, so I cannot post it publicly.

Libs/MRML/Widgets/qMRMLScalarsDisplayWidget.cxx
```diff
@@ -150,6 +153,24 @@ void qMRMLScalarsDisplayWidget::setMRMLDisplayNodes(QList<vtkMRMLDisplayNode*> d
...
+      {
+        same = false;
+        break;
+      }
+    }
+    if (same)
+    {
+      return;
+    }
+  }
```
ebrahimebrahim, line 172: Here it is good that we now skip the full rebuild of the display nodes list, but we additionally end up skipping things like `d->CurrentDisplayNodes[0]->UpdateScalarRange();` and `this->updateWidgetFromMRML();` below, which you'd still want to run if the current display nodes list has not changed but some other thing such as scalar data changes (e.g. changing scalar data should cause the displayed range widgets to update). 

We can skip rebuild but should not skip those updates

ebrahimebrahim (changes requested):
Thanks this looks great

I reproduced the crash by creating models using the following snippet:

<details>

```python
numberOfModels = 5
scalarNames = ["scalar1", "scalar2", "scalar3"]

for modelIndex in range(numberOfModels):
  sphere = vtk.vtkSphereSource()
  sphere.SetRadius(8.0)
  sphere.SetCenter(modelIndex * 25.0, 0.0, 0.0)
  sphere.SetThetaResolution(24)
  sphere.SetPhiResolution(12)
  sphere.Update()

  polyData = vtk.vtkPolyData()
  polyData.DeepCopy(sphere.GetOutput())

  pointData = polyData.GetPointData()
  for scalarIndex, scalarName in enumerate(scalarNames):
    scalarArray = vtk.vtkFloatArray()
    scalarArray.SetName(scalarName)
    scalarArray.SetNumberOfComponents(1)
    scalarArray.SetNumberOfTuples(polyData.GetNumberOfPoints())

    for pointIndex in range(polyData.GetNumberOfPoints()):
      point = polyData.GetPoint(pointIndex)
      scalarArray.SetValue(pointIndex, point[scalarIndex] + modelIndex * 10.0)

    pointData.AddArray(scalarArray)

  pointData.SetActiveScalars(scalarNames[0])

  modelNode = slicer.mrmlScene.AddNewNodeByClass(
    "vtkMRMLModelNode", f"Scalar model {modelIndex + 1}"
  )
  modelNode.SetAndObservePolyData(polyData)
```

</details>

Requesting:

- grammar fix from copilot
- my comment about not skipping widget updates when the display node list does not change

Libs/MRML/Widgets/qMRMLScalarsDisplayWidget.cxx
lassoan: > changing scalar data should cause the displayed range widgets to update

The display node observer takes care of that.
We only need to do an initial update in this method when a different node is selected.

ebrahimebrahim: I see!

Also I see the other fix was made earlier so approving and merging. Thanks!

ebrahimebrahim (approved):

PR #9174, opened 2026-05-21
BUG: Force Qt WebEngine to use ANGLE/D3D11 on Windows

lukas-folle-snkeos:
## Summary

Sets `QTWEBENGINE_CHROMIUM_FLAGS=--use-angle=d3d11` in the Slicer launcher on Windows. Resolves two long-observed Windows symptoms that share a single root cause: 3D view gray-flashing during interaction, and Extensions Manager rendering glitches.

## Symptoms (Windows 11 + NVIDIA Quadro/RTX, OpenGL2 + Qt 5.15)

- **3D view gray-flashes** during mouse interaction (rotate/pan/zoom). Window title flashes `(Not Responding)`; UI thread blocks briefly on each interaction.
- **Extensions Manager renders incorrectly** (visual glitches in the Qt WebEngine catalog view).

Reproduced on Slicer 5.11.0-2026-05-19 (rev 34550), Windows 11 Pro build 22631, NVIDIA Quadro RTX 5000, driver 595.79, Qt 5.15.2.

## Root cause

`Qt::AA_ShareOpenGLContexts` is already set in `qMRMLWidget::preInitializeApplication()` before `QApplication` construction (required for `QVTKOpenGLNativeWidget` + other Qt OpenGL widgets). A side effect on Windows is that Qt 5.15 WebEngine is pulled onto the desktop OpenGL backend rather than its usual ANGLE/D3D11 path. WebEngine's forked Chromium GPU process then competes with VTK for the same OpenGL context on the NVIDIA driver, producing the two symptoms above.

Confirmed empirically: launching with `QTWEBENGINE_CHROMIUM_FLAGS=--use-angle=d3d11` resolves both symptoms while keeping WebEngine hardware-accelerated. The more aggressive `--disable-gpu` resolves it too but at a perf cost to WebEngine UI.

## Fix

Add a Windows-only block to `CMake/SlicerBlockCTKAppLauncherSettings.cmake` (both BUILD and INSTALLED env sections) setting `QTWEBENGINE_CHROMIUM_FLAGS=--use-angle=d3d11`. Mirrors the existing Linux-only `QTWEBENGINE_DISABLE_SANDBOX=1` pattern in the same file. Setting it via the launcher (rather than in C++) is important: the variable must be present in the environment before Chromium's child process forks.

## Alternatives considered

- **`Qt::AA_ShareOpenGLContexts` adjustments** — already set correctly; not the missing piece.
- **`--disable-gpu-compositing`** — less aggressive but still degrades WebEngine perf; ANGLE/D3D11 is the cleaner choice.
- **`--disable-gpu`** — confirmed workaround in the original report but disables WebEngine GPU entirely; unnecessary trade-off.
- **`--use-angle=d3d11` was selected** because Windows ANGLE/D3D11 is the default WebEngine path absent `AA_ShareOpenGLContexts`; making it explicit just restores normal behavior.

## Test plan

- [x] Repro on Windows 11 + NVIDIA Quadro RTX 5000 + driver 595.79: both symptoms reproduced.
- [x] Apply fix: both symptoms resolved; WebEngine still GPU-accelerated.
- [ ] Linux build (NVIDIA + Mesa) — change is `if(WIN32)` guarded; no expected effect.
- [ ] macOS build — change is `if(WIN32)` guarded; no expected effect.
- [ ] Windows + AMD GPU — verify no regression.
- [ ] Windows + Intel iGPU — verify no regression.
- [ ] Windows + GeForce — verify no regression.
- [ ] WebEngine page-load and 3D-view FPS spot check — confirm no perf regression.

## Notes / known limitations

- Verified only on one Windows + NVIDIA configuration; broader hardware coverage welcome from CI / other testers before merge.
- User overrides of `QTWEBENGINE_CHROMIUM_FLAGS` set by the launcher will be overwritten by this default; if that becomes a concern, the launcher's additional-settings file can extend it.
- Qt 6 WebEngine has a different GPU process architecture; this workaround may become unnecessary after the Qt 6 migration but should remain harmless.

### DIFF
--- a/CMake/SlicerBlockCTKAppLauncherSettings.cmake
+++ b/CMake/SlicerBlockCTKAppLauncherSettings.cmake
@@ -149,6 +149,17 @@ if(UNIX AND NOT APPLE)
     "QTWEBENGINE_DISABLE_SANDBOX=1"
     )
 endif()
+if(WIN32)
+  # Route Qt WebEngine's Chromium GPU process through ANGLE/D3D11 instead of
+  # desktop OpenGL. With AA_ShareOpenGLContexts set, WebEngine on Windows can
+  # otherwise be pulled onto the desktop OpenGL backend and contend with VTK's
+  # GL context (observed on NVIDIA), causing 3D view gray-flashes during
+  # interaction and Extensions Manager rendering glitches. ANGLE/D3D11 keeps
+  # WebEngine hardware-accelerated without sharing the GL context.
+  list(APPEND SLICER_ENVVARS_BUILD
+    "QTWEBENGINE_CHROMIUM_FLAGS=--use-angle=d3d11"
+    )
+endif()
 
 # External projects - environment variables
 foreach(varname IN LISTS Slicer_EP_LABEL_ENVVARS_LAUNCHER_BUILD)
@@ -289,6 +300,17 @@ if(UNIX AND NOT APPLE)
   # (see https://github.com/Slicer/Slicer/issues/6577)
   list(APPEND SLICER_ENVVARS_INSTALLED
     "QTWEBENGINE_DISABLE_SANDBOX=1"
+    )
+endif()
+if(WIN32)
+  # Route Qt WebEngine's Chromium GPU process through ANGLE/D3D11 instead of
+  # desktop OpenGL. With AA_ShareOpenGLContexts set, WebEngine on Windows can
+  # otherwise be pulled onto the desktop OpenGL backend and contend with VTK's
+  # GL context (observed on NVIDIA), causing 3D view gray-flashes during
+  # interaction and Extensions Manager rendering glitches. ANGLE/D3D11 keeps
+  # WebEngine hardware-accelerated without sharing the GL context.
+  list(APPEND SLICER_ENVVARS_INSTALLED
+    "QTWEBENGINE_CHROMIUM_FLAGS=--use-angle=d3d11"
     )
 endif()
 


lassoan:
Thank you for investigating this and proposing a change.

Could you give more information on how to reproduce the symptoms? I've used Slicer on about 10 different computers in the past year and I haven't noticed such issues. If we find that the problem is only reproducible on a narrow set of hardware or can cause regression on some other hardware then we would probably want to make this change optional.

I checked there references to see if they give any hint on how to reproduce the symptoms, but they seem to be unrelated. Please review and remove those references to avoid confusion - unless you find that they are actually related to what is discussed here.

You can now build Slicer using Qt6 on Windows ([build instructions](https://slicer.readthedocs.io/en/latest/developer_guide/build_instructions/windows.html) include Qt6). It would be great if you could check if the proposed change is still effective, necessary, and/or causing any issues (we would use this information to add this flag only for Qt5 or also for Qt6).

lukas-folle-snkeos:
@lassoan thanks for getting back on this so quickly!
I will have a look into whether Qt6 does fix this for me. I noticed this error only popped up after I did upgrade my Nvidia GPU Driver to a recent version so my guess was that this might affect more people moving forward.

Essentially the whole screen goes grey for about 2-3 seconds whenever I interact with the 3d rendering. Same is for the Extension Manager, where nothing is rendered at all when I go to the "Install Extensions" tab.

Sorry about the incorrect references, those were wrongly added by my little helper..

lassoan:
> this error only popped up after I did upgrade my NVIDIA GPU Driver to a recent version

It may be a driver bug then. What is the driver version that shows these symptoms? Do you know what version you had before that worked correctly?

lukas-folle-snkeos:
I got this with driver 595 and 596. I don't remember the previous driver though. Must have been quite a bit older.

lassoan:
For reference, everything works well (start Slicer, open the Extensions Manager, close it, interact with 3D views, no artifacts or gray screen) with these hardware and driver versions on Windows 11:
- RTX 3090 with 581.95
- RTX 3060 with 591.86
- RTX 3080 Ti with 572.83
- RTX 3080 Ti with 596.36 Studio Driver

lukas-folle-snkeos:
Interesting!

lassoan:
I've updated to latest NVIDIA Studio drivers (596.36) on one of the computers and everything still looked normal.

The issue might be Quadro specific. We experienced problems with Quadro drivers several times in the past 10-15 years. Quadro drivers are certified to work optimally on a small set of professional applications, while Studio or Game Ready drivers are tested with a wide range of applications and have ~100x more users, so errors are found and fixed much more quickly. It may worth testing if downgrading the driver fixes the problem.

It would be useful if you could provide instructions and post screenshots or screen recordings (maybe with a camera, if screen recording cannot capture the screen content) tp document how to reproduce the error and how it looks.

lukas-folle-snkeos:
I can't really wrap my head around it, but I cannot reproduce the error anymore when I unset the env var. I will monitor and open an issue when I can record it happening again. Thanks for walking me through the steps @lassoan

PR #9175, opened 2026-05-22
BUG: Preserve markups plane normal display settings

### LINKED ISSUE #8518: plane setting not recording
In the markup module, when I create and edit a plane, in the plane settings, I tick the visibility box and set the ruler to 0 out of 100 (I want to hide the arrow). Then, I save the file as a bundle (.mrb). When I reopen the bundle the arrow is displayed, the plane settings show that the visibility is ticked anew and the ruler set to 100.
Thus, it looks as if the save command has forgotten those values.
## Steps to reproduce
-create a plane with the markups module.
-go to the bottom, toward the "plane settings"
-tick the visibility.
-set  the visibility ruller to 100.
-save the file as a bundle.
-reopen the bundle.
-check, in the markup module the settings for this plane, down to the plane settings, box and ruller visibility.

- Slicer version: Slicer-5.0.9-2025-06-19
- Operating system:  Mac + sequoia 15-5

vacaslistradas:
## Summary

This preserves plane markup normal-arrow display settings across scene save/load by serializing `NormalVisibility` and `NormalOpacity` on `vtkMRMLMarkupsPlaneDisplayNode`.

## Root cause

The Markups UI updates these values on the plane display node, but the plane display node did not write or read them from MRML. As a result, saved scenes and MRB reloads restored the default normal-arrow visibility and opacity.

## Changes

- Add `WriteXML` / `ReadXMLAttributes` support for plane normal visibility and opacity.
- Copy the same fields in `CopyContent` and include them in `PrintSelf`.
- Add a focused MRML scene round-trip regression test for these display settings.

Fixes #8518

## Testing

- `git diff --check`
- `pre-commit run --files Libs/MRML/Core/Testing/CMakeLists.txt Libs/MRML/Core/vtkMRMLMarkupsPlaneDisplayNode.cxx Libs/MRML/Core/vtkMRMLMarkupsPlaneDisplayNode.h Libs/MRML/Core/Testing/vtkMRMLMarkupsPlaneDisplayNodeTest1.cxx`

### DIFF
--- a/Docs/developer_guide/script_repository/markups.md
+++ b/Docs/developer_guide/script_repository/markups.md
@@ -623,7 +623,7 @@ base_to_node_matrix[0:3, 1] = axis2
 base_to_node_matrix[0:3, 2] = normal
 
 data = {
-    "@schema": "https://raw.githubusercontent.com/slicer/slicer/master/Modules/Loadable/Markups/Resources/Schema/markups-schema-v1.0.3.json#",
+    "@schema": "https://raw.githubusercontent.com/slicer/slicer/master/Modules/Loadable/Markups/Resources/Schema/markups-schema-v1.0.4.json#",
     "markups": [
         {
             "type": "Plane",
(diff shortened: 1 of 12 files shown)


lassoan:
We usually save markups display settings in the json file as well. Could you add these two new properties to the schema and read/write them in the json file?

vacaslistradas:
Addressed in 2c1eb22: added normalVisibility/normalOpacity to the markups JSON schema, read/write them with the display properties in .mrk.json files, and extended the markups storage test to round-trip the plane display values.

lassoan (approved):
After some tweaks it looks good, thank you!

PR #9177, opened 2026-05-24
DOC: Fix elongation and flatness definitions in segment statistics

gliu2:
The current documentation on how elongation (and flatness) is calculated is poorly worded (https://github.com/slicer/slicer/blob/main/Docs/user_guide/modules/segmentstatistics.md) — for 3D, "second largest" and "second smallest" both refer to the middle principal moment, so the ratio as literally described would just be 1. 

I analyzed with Claude Opus 4.7 the segment statistics module. Claude's output:

Here's what's actually happening under the hood.
What ITK actually computes
Slicer's segment statistics module delegates to ITK's ShapeLabelMapFilter. Looking at the source in itkShapeLabelMapFilter.hxx: cppconst double elongationRatio =
    principalMoments[ImageDimension - 1] / principalMoments[ImageDimension - 2];
elongation = std::sqrt(elongationRatio);
The principal moments come from vnl_symmetric_eigensystem, which returns eigenvalues sorted in ascending order. So with λ₁ ≤ λ₂ ≤ λ₃:

principalMoments[ImageDimension - 1] = [2] = λ₃ (largest) principalMoments[ImageDimension - 2] = [1] = λ₂ (middle)

So in 3D:
Elongation = √(λ₃ / λ₂) — the square root of the ratio of the largest to the second-largest principal moment of inertia. Value is always ≥ 1. A value of 1 means the object has equal spread along its two longest axes (e.g., a sphere, or a perfect prolate spheroid that's rotationally symmetric about its long axis). Larger values mean more cigar-like. Why the docs read the way they do
The wording on the Slicer page is almost certainly a botched generalization from the 2D case in the original Insight Journal paper (Lehmann & Beare, publication 301). In 2D, there are only two moments — so "second largest" is literally "the smaller of the two" and "second smallest" is "the larger of the two." That language gives you λ_small/λ_large ≤ 1 in 2D and makes superficial sense. But it doesn't generalize to 3D, and whoever wrote the Slicer docs just copied the phrasing without checking.


gliu2:
Claude chat thread that suggested this edit: https://claude.ai/share/641ae91f-2fd1-4234-a476-61d96e93ea53

lassoan (approved):
Thank you for your contribution. The original version is copied from the reference documentation. The proposed text is more informative.

PR #9178, opened 2026-05-25
BUG: Fix Extension Wizard module creation

cpinter:
Classes deprecated in Qt6 have been replaced with their new variants. They are available in Qt 5.15.2 as well, so no condition is needed.


lassoan (approved):
Looks good to me, thank you!

PR #9183, opened 2026-05-26
BUG: vtkMRMLMarkupsROINode: Remove implicit cast to float

NogginBops:
In vtkMRMLMarkupsROINode there where a few uses of `vtkNew<vtkPoints>` which defaults to `vktPoints::New()` which creates a list of float32 points, but the data is float64 so we shouldn't do the implicit cast.

### DIFF
--- a/Libs/MRML/Core/vtkMRMLMarkupsROINode.cxx
+++ b/Libs/MRML/Core/vtkMRMLMarkupsROINode.cxx
@@ -1036,6 +1036,7 @@ void vtkMRMLMarkupsROINode::UpdateControlPointsFromBoxROI()
   double center_World[3] = { 0.0, 0.0, 0.0 };
   this->GetCenterWorld(center_World);
   vtkNew<vtkPoints> points_World;
+  points_World->SetDataTypeToDouble();
   points_World->InsertNextPoint(center_World);
   this->SetControlPointPositionsWorld(points_World);
   this->MaximumNumberOfControlPoints = 1;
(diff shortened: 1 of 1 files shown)


lassoan:
Thanks for the PR. Has this come up as an issue for you? Have you built VTK with custom settings?

NogginBops:
Yes, I discovered this when making a plugin for slicer. The center of a region of interest is casted to floating point precision (due to weird interactions between `SetSize` and `SetCenter` eventually calling `UpdateControlPointsFromBoxROI` because of modified events).

The issue is present in the latest preview version of slicer. 
It also appears if you compile slicer according to the build instructions. 
I'm not using a custom build of vtk, this is the default behaviour of vtkPoints: https://gitlab.kitware.com/vtk/vtk/-/blob/master/Common/Core/vtkPoints.cxx#L36-L39

With this patch this is no longer a problem and the double precision floating point values are preserved as they should be.

pieper (approved):
Sounds reasonable to me.  Maybe we should have a small test for this.

lassoan:
I see, this is only about unnecessary loss of precision. It is unlikely to cause any problems in practice, but it makes sense to make these simple changes to preserve precision.

lassoan:
I've checked the code if we should consider forcing using double for point arrays and found the we almost always use the default precision for points (which is float). Using float in vtkPoints is a conscious choice in VTK to reduce the memory footprint and computation cost in all geometry processing. If someone wants to do all this math in double precision then a custom vtkPoint factory can be registered that creates double arrays by default.

Based on this, I don't see a reason to treat markup ROI as a special case and force storing its point coordinates as double. Therefore, **I would vote for not making the changes proposed in this pull request**.

pieper:
I like the idea of making a more comprehensive change that would allow users to pick float or double based on their use cases and making sure that these choices are respected through the whole processing stream.

lassoan:
I agree that changing coordinate precision globally may make sense. This is already supported at VTK level. You can choose what types to use for IDs, point coordinates, etc. some of them with CMake variable, some of them with custom object factories. All we need to do in Slicer is not to interfere with these mechanisms: let `vtkPoints` create the array type it wants, and do not hardcode it to be float or double (unless we have a specific reason to do so).

lassoan:
I'm closing this now to reduce clutter in the PR list, but open to discuss further.

PR #9184, opened 2026-05-27
ENH: Merge the SlicerLayerDisplayableManager extension as a core module

Thibault-Pelletier:
Integrates the LayerDM extension as a core module as a stepping stone to
refactoring and extending Slicer's core Displayable Manager infrastructure.

Merged from https://github.com/KitwareMedical/SlicerLayerDisplayableManager/commit/ee9423d6f4556f6339ea5dbf871f5319c3ae0ec3
while preserving the original extension's git history.

@lassoan @pieper @RafaelPalomar 

Current PR status:
* [x] Fix pre-commit / lint issues
* [x] Remove Slicer extension specific settings
* [x] Migrate documentation
* [x] Migrate examples
* [x] Test slicer-core wheel build with fork
* [ ]  ~Refactor the Model DM with the LayerDM as a concrete example~ 
    * To be implemented in a follow up PR


Thibault-Pelletier:
@RafaelPalomar @lassoan aside from a concrete refactoring, this PR is feature complete and ready for review.

I'll be looking into the current Model DM during the week as a refactoring exercice, but as this may break existing behavior, merging it should be outside the scope of this PR.

With a first look at the problem, the step will probably be to:

- Remove the vtkMRMLModelDisplayableManager from the  Libs\MRML\DisplayableManager target
- Add a dedicated MRMLDM folder to the Modules\Loadable\Models targets
- Implement the Pipeline(s) in the Modules\Loadable\Models\MRMLDM  and register it in the module's init logic (impl in C++ as the C++ layer is missing from the examples at the moment)

Let me know if the way I merged the module works for you / if this plan looks sensible.
If needed, we can defer the discussion until next week's dev meeting and I'll be able to provide the follow up PR in the meantime.

RafaelPalomar (commented):
Thanks for the careful import — diagrams, tests, docs, and green CI all look in good shape. One architectural question before we commit to the integration shape.

In my view,  there are three integration paths in increasing commitment:

  1. Bundle in SuperBuild. Available to extension developers (C++ + Python), CMake stays as-is, full repo decoupling, easy to iterate the API. This move is also reversible
  2. Core as LayerDM + LayerDMPython (loadable+scripted). The split honors the C++/Python module convention. Required if other core DMs (e.g. a Models DM refactor) are going to depend on  LayerDM.
  3. Dissolve into MRML/Libs. It is probably premature. 

The module, as is stands, is a C++/Python hybrid in a place where Slicer core has only had pure-C++ loadables. Four C++ files include `Python.h` and the build gates them with `if(VTK_WRAP_PYTHON)` — a first for a core loadable, and the gate is inconsistent with `Testing/CMakesLists.txt` which uses `Slicer_USE_PYTHONQT`. There's also a parallel set of build files (`vtk.module`, ` wheel.cmake`, `extension.cmake`, checked-in `Export.h`) that the loadable-module path doesn't use; they exist for the slicer-core Python wheel build.e

My preference is (1) as a staging step, with an explicit plan to graduate to (2) in 5.12 / 6.0 once the API has been tested by 1–2 DM migrations done as extensions (Slicer-Liver is walking this path); it is important to consider if there are any short-term plans for refactoring DMs (@Sunderlandkyl). (2) could be the right way now only if a core DM consumer is queued for the next few months and the LayerDM / LayerDMPython split lands in this PR.

I think these points are worth discussion before continuing this PR (@Thibault-Pelletier, @lassoan, @pieper, @sjh26)

pieper:
Starting with this as a superbuild option makes sense to me.  We want to keep the C++ / Python layering clean.

Thibault-Pelletier:
Thanks @RafaelPalomar and @pieper 

I wanted to use this as an exercise to test what it would mean to move the LayerDM into Slicer at the moment.

My preference is also on option (1) which would allow to be leaner on the the LayerDM's API changes if needed during integration / refactoring. 

Regarding the Python / C++ split, the LayerDM complies with how the [Segmentation Loadable Module](https://github.com/Slicer/Slicer/blob/main/Modules/Loadable/Segmentations/EditorEffects/CMakeLists.txt#L35) and the [Libs/MRML/Core](https://github.com/Slicer/Slicer/blob/main/Libs/MRML/Core/CMakeLists.txt#L421) have been implemented for the Scripted trampolines.

As the Scripted version's API are linked to the non-scripted one, centralizing them reduces complexity when udpating.

The usage of Slicer_USE_PYTHONQT and VTK_WRAP_PYTHON mainly depends on where the target is being built (VTK only module vs Qt layer).

The split for the CMake files is only a slicer-core wheel thing. The only requirement there is to have a vtk.module to automatize the build / wrapping / PyPI generation.

If we migrate Slicer's VTK targets to using VTK modules, there would only be one CMake file, but this is probably a discussion for another time.

lassoan:
Thanks a lot @Thibault-Pelletier. Many Slicer core limitations and inconveniences are implemented in these new classes, which would be nice to get integrated into Slicer core.

Integration will take a long time, with potential API changes, so it would not worth trying to do any of it before Slicer-5.12 is released. After that, I would consider merging all the classes in their current form into Slicer core, and then move features into existing Slicer core classes as much as possible and promote the rest to be Slicer core classes. It is quite likely that in the end there would be no need for LayerDM module at all.

I would not try to add an intermediate step of having the extension in SuperBuild.cmake (similar to CompareVolumes, LandmarkRegistration, SurfaceToolbox). This mechanism mostly proved to be a failure to me, as these components are neither Slicer extensions nor reusable libraries, they can only be used when built into Slicer, and updating them is a pain. It would be simpler to keep LayerDM as a proper extension until we are ready to start the integration process. Qt6 and ARM will make the main branch unstable for longer time than usual, which may be a good opportunity to do this integration in the same release (in Slicer-5.13).

Thibault-Pelletier:
@lassoan thank you for comments!
Yes, I guess it would make sense to properly integrate le module into Slicer in one go.
Since we are aiming for a major release after version 5.12, and we will go through additional testing, it would also make sense to use this as an opportunity to do both the integration and refactoring of existing DMs as well.

FYI, contrary to my first plans, I have added the model's DM refactoring as the last commit of this PR to centralize things (I'm still going through a bit of testing).

My implementation for the picking is a bad design choice for this type of feature, but it highlights how you could reuse other DM Pipelines in a more decoupled manner.

Happy to discuss that more at Tuesday's meeting!


EDIT: After a bit of testing some features are broken (clipping / scalar display at least). 

Since the goal is not really to implement the refactoring I'll delay investigation / fix until the integration of the module has been done.

lassoan:
@Thibault-Pelletier let me know if you think it is very close to the final version and ready for review.

Thibault-Pelletier:
@lassoan yes, I just updated the branch with main (both for Slicer and the LayerDM extension) and ran the tests on Windows, so it's ready for review.

As discussed a few weeks ago, I have left out the ModelDisplayableManager refactoring out of this PR so that the actual refactoring can be properly reviewed and address any regressions without being polluted by the +10 000 of the library itself...

Following changes on my end will be limited to handling PR feedbacks / doc and maybe increasing the number of examples if I can find some time tomorrow or on Monday. So you can go ahead with the integration review 👍.

@RafaelPalomar you can use the PR to test it against SlicerLiver.
Let me know if you find any problems 🙏 

EDIT: Documentation is up to date with SlicerLayerDM main

RafaelPalomar:
> @RafaelPalomar you can use the PR to test it against SlicerLiver.
> Let me know if you find any problems 🙏

I'll check it out and let you know. Thanks for working on this

RafaelPalomar:
> @RafaelPalomar you can use the PR to test it against SlicerLiver.
> Let me know if you find any problems 🙏

@Thibault-Pelletier I've tested this against Slicer-Liver as you suggested; results below. Short version: the migration was close to free, and I'm withdrawing concerns I raised in May.

## Setup

Built this branch (`843211e`) on Linux/Guix, Qt6, Release, and migrated a Slicer-Liver working branch carrying ~770 commits of LayerDM-dependent work from the extension to the core module.

## Result: the migration is close to free

| | |
|---|---|
| CMake edits | 7 across 6 files |
| Net lines | **−29** |
| C++ changes | **none** |
| Test suite | **44/44 pass**, 0 fail |

`target_link_libraries` didn't change at all — the core module keeps the extension's target names, so only discovery moved. Every edit was about *deleting* something rather than adding it:

  - `LayerDisplayableManager` out of `EXTENSION_DEPENDS`, plus the comment block
    explaining why launcher-settings forwarding was needed
  - `find_package(LayerDisplayableManager REQUIRED)` in the resections module
  - four `--additional-module-paths ${LayerDisplayableManager_DIR}/...` opt-ins
    in the Python test harnesses

`vtkMRMLLayerDMPipelineI`, `vtkMRMLLayerDMPipelineFactory`,
`vtkMRMLLayerDMPipelineManager`, `vtkMRMLLayerDisplayableManager`,
`vtkMRMLLayerDMLayerManager` and `vtkMRMLLayerDMInteractionLogic` all seem to resolve
from `slicer`, `layerdm` registers with the module manager, and `LayerDMLib` imports.

About  the **CMake gating.** You were right. I checked `Segmentations/EditorEffects` (Qt layer, `Slicer_USE_PYTHONQT`) against  `Libs/MRML/Core` (VTK-only, `VTK_WRAP_PYTHON`), and LayerDM follows the same  rule — VTK-level targets in `MRML`/`MRMLDM`, app-level tests in `Testing`. Not an inconsistency

`Libs/MRML/DisplayableManager/vtkMRMLScriptedDisplayableManager.cxx` already embeds CPython in a core VTK-level library, and it is a scripted DM at that 
  
## AI review 

I did pass the setup through an AI review and raised the following concern.
  
That precedent keeps CPython confined to the `.cxx`, exposing only `SetPythonSource(const std::string&)` in its header. Here, `#include <vtkPython.h>` sits in four *headers*:

  - `MRML/vtkMRMLLayerDMPythonUtil.h`
  - `MRML/vtkMRMLLayerDMObjectEventObserverScripted.h`
  - `MRMLDM/vtkMRMLLayerDMPipelineScriptedCreator.h`
  - `MRMLDM/vtkMRMLLayerDMScriptedPipelineBridge.h`
  
So CPython lands in the library's public include surface rather than its
implementation. Was that deliberate, or would the same pimpl-style split the
scripted DM uses work here? It doesn't bite Slicer-Liver — it only includes
`vtkMRMLLayerDisplayableManager.h` — so this is about the shape of the public
surface, not a problem I hit.
  
## Integration shape

The way you merged the module  works for me.** Consuming it as a core module was mechanical, and leaving the ModelDisplayableManager refactor to a follow-up PR feels correct. I'd be happy to go with merging in current form.

Nice work!

Thibault-Pelletier:
Hi @RafaelPalomar, thx for the review!
I'll check to forward declare vtkPython / vtkPythonUtil and get back to you when it's good (hopefully later today)

Thibault-Pelletier:
@RafaelPalomar I have just moved the vtkPython includes to the implementation

Thibault-Pelletier:
@RafaelPalomar I wanted to edit my previous message but GitHub is having difficulties.
For information, I reverted the forward declaration changes as this breaks the Python wrapping for methods using PyObject* as inputs.

The PR can be considered as complete with its current scope.

Thibault-Pelletier:
@RafaelPalomar and @lassoan, if it's good for you, this PR can be merged

lassoan:
Let's aim for integrating it on Friday, if no other questions or concerns come up until then.

RafaelPalomar (approved):
It's a great integration of the SlicerLayerDM. I appreciate that the documentation follows in the PR. I tested it with our newer SlicerLiver, which was using the external SlicerLayerDM module and migration was easy; everything worked as it should.

Docs/developer_guide/layer_dm/examples/ModelGlowDM.py
```diff
@@ -0,0 +1,655 @@
...
+
+        return cls()
+
+    @classmethod
+    def _GetClassName(cls) -> str:
+        """
+        Convenience method to get the name of the current class.
+        This method will return the actual class name for inheriting classes.
+        """
+        return cls.__name__
```
lassoan, line 199: `cls._GetClassName()` is not shorter or more convenient than `cls.__name__`. I'm not sure if there is a difference in behavior between the two in inheriting classes either. Is there some other reason that could justify the existence of this function? Maybe decoupling of the actual class name from the string that gets stored in the scene? (but then the method name should not be `_GetClassName()` but something like `_GetTypeID`) If we don't have a good reason to add this method then it is better to remove it and use `cls.__name__` directly.

Docs/developer_guide/layer_dm/examples/CustomVR.py
```diff
@@ -0,0 +1,479 @@
...
+        self._imageData = None
+
+    @classmethod
+    def CreateVRNode(cls, volumeNode, scene) -> vtkMRMLScriptedModuleNode:
+        node = cls.GetVRNode(volumeNode, scene)
+        if node:
+            return node
+
+        node = vtkMRMLScriptedModuleNode()
+        node.SetAttribute("PipelineType", cls._GetClassName())
```
lassoan, line 164: Is there a reason for using `SetAttribute` instead of `SetParameter`?

But a more important issue is that the "glow node" should be a display node (because we want to be able to show the custom VR in specific views without reimplementing the display node's view node logic) but here it is a scripted module node. Since the layer DM is explicitly for display, it should use display nodes. We could add a `vtkMRMLScriptedDisplayNode` class that would allow us to implement display nodes in Python.

lassoan (commented):
It looks great, but some changes would be needed to match Slicer conventions, make names more intuitive, and simplify custom layer DM implementations. I'll push a number of changes on this branch that are fairly straightforward (so it is easier to suggest and test a change instead of starting discussions about them) and added some comments inline, but in addition to these it would be nice to:
- Add scripted MRML display node so that we can use display nodes as layer DM nodes (that is the correct type for a displayable manager to process)
- Add a few complete C++ and Python scripted layer DM modules to the build (that are tested and the user can try them; but no need to include them in the Slicer package). Modules in `Docs/layer_dm/examples` would work for scripted modules, but they need to be part of the build process and tested. We need the same for C++.

lassoan:
I've rebased on latest origin/main and added my suggested changes. If you want to make further changes or undo some then please do in additional commits (don't change existing commits -  we can squash commits in the very end).

I've made changes to the API, so code that uses the layer DM will have to be updated accordingly. You can do this automatically by pointing an LLM to the list of commits that I added. It requires some effort, but it is better to fix/improve the API now, because later it would be much more difficult.

lassoan:
It would be good if we could remove these two classes:
- vtkMRMLLayerDMObjectEventObserver -> improve the vtkObserverManager instead
- vtkMRMLLayerDMNodeReferenceObserver -> improve node references in vtkMRMLScene instead

If these classes are only used internally (not needed if someone wants to implement a DM) then the classes can be reworked/removed later; if they are public API then it would be better to remove/reduce them before integration.

Modules/Loadable/LayerDM/Logic/vtkSlicerLayerDMLogic.cxx
```diff
@@ -0,0 +1,120 @@
+/*==============================================================================
+
+  Program: 3D Slicer
+
+  Portions (c) Copyright Brigham and Women's Hospital (BWH) All Rights Reserved.
```
Thibault-Pelletier: @lassoan attribution is wrong as this was fully developed by Kitware Europe.

@finetjul what should be added (including funding)?

Modules/Loadable/LayerDM/MRML/vtkMRMLLayerDMNodeReferenceObserver.h
```diff
@@ -0,0 +1,133 @@
...
+
+/// Reference node observer.
+/// Triggers node ref added / removed when references change in the scene.
+/// Allows to appropriately update pipelines when references to a given display node are added / removed.
+class VTK_SLICER_LAYERDM_MODULE_MRML_EXPORT vtkMRMLLayerDMNodeReferenceObserver : public vtkObject
+{
+public:
+  /// Reference to a node with a given role.
+  ///
+  /// The node is held by a weak pointer, unlike the other LayerDM containers, because these lists can outlive
```
Thibault-Pelletier: @lassoan not a fan of implementation detail in the class docstrings

Modules/Loadable/LayerDM/MRML/vtkMRMLLayerDMNodeReferenceObserver.h
```diff
@@ -0,0 +1,133 @@
...
+  vtkSmartPointer<vtkMRMLLayerDMObjectEventObserver> Observer;
+
+  /// @{
+  /// References from and to each observed node.
+  ///
+  /// The keys own their node, which keeps every node referenced by a RefT of the opposite map alive.
+  ///
+  /// \sa RefT
+  /// \sa OnNodeRemoved
+  std::map<vtkSmartPointer<vtkMRMLNode>, std::vector<RefT>> NodeToReferences;
```
Thibault-Pelletier: Why the change to std::vector instead of set?

Thibault-Pelletier:
@lassoan thank you for the review.

There are a few things that I would like to address before moving forward with this.

- File renaming and public API changes: A few files and API have been renamed in your commits. Renaming the files and changing the API will break existing extensions, including but not limited to trame-slicer, and I'm not sure I see the reason why these changes were made. 
- Copyright: The copyright attribution is wrong. I'll let @finetjul pitch in for the correct one
- Noise: I think there are useful changes in what you committed, but they are buried in a lot of noise including long comments, and what seems like arbitrary changes. I would rather we discussed the changes, either in a Slicer dev meeting or a dedicated meeting to see what makes sense and how to properly address the issues without cluttering the extension.

Docs/developer_guide/layer_dm/examples/ModelGlowDM.py
Thibault-Pelletier: GetClassName is not required by the extension but is only one way to store the information in the node to trigger the TryCreate mechanism.

Here I used cls.__name__, but you could use the __qualname__ for added safety (esp for different modules).

Docs/developer_guide/layer_dm/examples/CustomVR.py
Thibault-Pelletier: No, SetAttribute was used arbitrarily. In trame-slicer, the segmentation effects use the SetParameter to encode all properties into the scripted nodes.

I agree that inheriting the vtkMRMLDisplayNode and having the view node ID mechanism would be nice, but this would also add two parallel similar implementations between the vtkMRMLScriptedModuleNode and the new one.

Since inheritance is not available with VTK / Python, I'm not sure what would be the added value of having a C++ trampoline for this instead as we would still need a dedicated Python class to simplify scripted display node definition.

lassoan: vtkMRMLScriptedDisplayNode would be a real display node, so it would store view node IDs, visibility, clipping, color,  and all other common display properties, so subject hierarchy and all modules would recognize it and would know how to use it.

Thibault-Pelletier:
> It would be good if we could remove these two classes:
> 
> * vtkMRMLLayerDMObjectEventObserver -> improve the vtkObserverManager instead
> * vtkMRMLLayerDMNodeReferenceObserver -> improve node references in vtkMRMLScene instead
> 
> If these classes are only used internally (not needed if someone wants to implement a DM) then the classes can be reworked/removed later; if they are public API then it would be better to remove/reduce them before integration.

Yes, vtkMRMLLayerDMNodeReferenceObserver can be removed by having a two way reference changed invoked in the vtkMRMLNode.
 
vtkMRMLLayerDMObjectEventObserver: There were multiple problems in vtkObserverManager for usage in Python, namely vtkCallbackCommand is not available and  vtkObject** syntaxe will not be available for assignment. So merging the two would be hard to do.

lassoan:
> File renaming and public API changes: A few files and API have been renamed in your commits. Renaming the files and changing the API will break existing extensions, including but not limited to trame-slicer, and I'm not sure I see the reason why these changes were made.

I know it is painful to make these changes, sorry, but this is practically the only time ever that these improvements can be discussed and made. Whatever gets merged now we will be stuck with forever. We can discuss any commit, they are all independent that we can modify or undo based on what we agree on.

> Copyright: The copyright attribution is wrong. I'll let @finetjul pitch in for the correct one

There were no file headers, so I added the default, but you can use any already established patterns.

>Noise: I think there are useful changes in what you committed, but they are buried in a lot of noise including long comments, and what seems like arbitrary changes. I would rather we discussed the changes, either in a Slicer dev meeting or a dedicated meeting to see what makes sense and how to properly address the issues without cluttering the extension.

We can discuss any changes that you find arbitrary, I can see how it may not be always clear what was the motivation and pros/cons considered.

Very long comments are certainly red flags (lengthy explanations should not be necessary, because the code should speak for itself), but I don't see that there are that many, or that long. Let's discuss specific cases.

I did not want to start a discussion about every potential change, as that would have resulted in dozens of discussions, but instead chose to propose an actual implementation for most of my concerns, so we can discuss those ones that we don't agree on.

Creating another branch would have made things more complex (keeping in sync dozens of slightly different commits in two different branches is much harder than having a linear history on a single branch). I added all my commits on top of your branch (did not modify anything except rebase on latest main branch), so your original implementation should be intact and easily identifiable. We can continue adding more commits, keeping the linear history. But we can try a different approach if you prefer.

Dedicated meeting(s) is probably required anyway, I just did not want to hold up this integration process until then.

Docs/developer_guide/layer_dm/examples/CustomVR.py
Thibault-Pelletier: If you want all those properties, they would have to be stored in the C++ layer and may not make sense for some of the DM pipelines. 

For instance, implementing a pipeline only responsible for the camera interaction is pretty straightforward in both Python and C++ using the LayerDM. For this type of DM, 99% of the display properties wouldn't make sense.

There are other uses cases where colors, clipping, and other properties wouldn't make sense (for instance creating a 2D overlay with graphs / mini map display / etc). So if we want to keep the scripted display node not confusing, it should be limited to the ViewNodeID list and the visible in view method IMO.

Thibault-Pelletier:
@lassoan thanks for the clarification.
I'll try to have a deeper look at the diff and add the changes to the top of the current commits so that we can further discuss.

PR #9186, opened 2026-05-28
BUG: Prevent loss of volume rendering opacity transfer function points

### LINKED ISSUE #9129: Volume Property json file is incorrectly saved
As reported in the forum: https://discourse.slicer.org/t/volume-property-json-file-is-incorrectly-saved/46834

I load a 16bit unsigned dataset, set up its volume rendering properties such as

1. Use shift slider
2. Change the material preset to a non-default one (the second brightest)
3. Remove some points from the scalar mapping and shift the colors aroun

At this stage I have perfectly fine rendering. if I save this file to disk and reload and assign this new volume property, the rendering dissappears as the scalar opacity flat lines (all points are 0.0). Reproduces on today preview build on mac.

I asked copilot what is different between the unmodified VP and the one saved and it said:

`Works.vp.json effectiveRange: [0, 65535] (full 16-bit range). doesntwork.json effectiveRange: [3800, 69335]. 69335 > 65535 — outside the actual data range. The min 3800 also chopped off the (0, …) and (2.22e-308, …) baseline points, which probably explains why the start of the tent is gone in the broken file. If effectiveRange was edited in the UI (or auto-recomputed from “current threshold” rather than data range), the saver may have culled all control points outside that range — including the peak’s left base — leaving just a flat-zero function.
`
Not entirely sure what that means, I hope it is helpful. But there is clear bug in saving.

If you want the original volume I used, it is here: https://github.com/muratmaga/Bumblebee_Stained/releases/download/v1/Bumblebee.nrrd

lassoan:
When volume rendering opacity transfer function is shifted using MovePoints, the point x coordinates are temporarily converted to 32-bit float. The small difference that vtkMRMLVolumePropertyNode::NextHigher and vtkMRMLVolumePropertyNode::HigherAndUnique functions introduced was not sufficient to survive this conversion and the x coordinates could end up having the same value. When the points were saved and reloaded, the vtkMRMLVolumePropertyJsonStorageNode did not use vtkMRMLVolumePropertyNode::HigherAndUnique, which resulted in piecewiseFunction->AddPoint overwriting an existing point.

Fix the problem by making vtkMRMLVolumePropertyNode::NextHigher large enough difference that comfortably survives 32-bit float conversion.

In addition to this, vtkMRMLVolumePropertyJsonStorageNode is improved to use vtkMRMLVolumePropertyNode::HigherAndUnique function, which ensures that every point has a unique coordinate regardless of how close the point coordinates are in the file. This is necessary because volume rendering property files may be created by external software, which may not know that x values have to be unique.

fixes #9129

### DIFF
--- a/Modules/Loadable/VolumeRendering/MRML/vtkMRMLVolumePropertyJsonStorageNode.cxx
+++ b/Modules/Loadable/VolumeRendering/MRML/vtkMRMLVolumePropertyJsonStorageNode.cxx
@@ -36,7 +36,9 @@
 #include <vtkVolumeProperty.h>
 
 // std includes
+#include <algorithm>
 #include <sstream>
+#include <vector>
 
 namespace
 {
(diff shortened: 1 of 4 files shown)


jamesobutler:
image

@hjmjohnson You brought in a merge commit which resulted in the commit message check failing. Slicer enforces linear history so merge commits are not allowed.

PR #9190, opened 2026-05-28
ENH: Add 2D slice view in-plane FoV rotation and horizontal flip

ruffsl:
## Summary

Adds two interactions for adjusting the in-plane orientation of a 2D slice view: an in-plane "roll" via **Ctrl+Left-click-and-drag** (the 2D analog of the existing 3D camera-roll gesture in `vtkMRMLCameraWidget`), and a **Flip horizontal** action in the slice view's right-click context menu. Both operate on the slice node's `SliceToRAS` and pivot about the current field-of-view center so that what the user is looking at stays put under pan/zoom.

Three commits, all `ENH:`:

1. `2051d08a` — `ENH: Add in-plane field of view rotation to 2D slice views`
2. `cb27aee5` — `ENH: Add horizontal flip action to slice view context menu`
3. `7faa5572` — `ENH: Flip slice view about field of view center instead of slice origin`

## Motivation

3D views already support a roll/spin gesture (Ctrl+Left-drag in `vtkMRMLCameraWidget::ProcessSpin`). 2D slice views had no equivalent: in-plane orientation could only be changed via the slice-controller orientation dropdown, `RotateToVolumePlane`, the reformat widget, or the multi-slice "rotate intersecting slices" gesture (Ctrl+Alt+Left). None of these gives a quick "tilt this view a few degrees" without committing to an orientation preset or reorienting *other* slices. Similarly, there was no built-in way to flip a slice view to view it from the opposite side (radiological/neurological convention swap), short of constructing a transform.

These two interactions fill those gaps with a UX that mirrors the established 3D roll gesture and a discoverable context-menu entry consistent with `Reset field of view`.

## What this adds

**FoV roll — Ctrl+Left-drag in 2D slice views.** Rolls the current slice in-plane about the renderer center. Implemented as a new `WidgetStateRotateFieldOfView` in `vtkMRMLSliceIntersectionWidget`, gated by the previously-unused `ActionRotate = 4 /* not used */` flag (already included in `ActionAll`, so the gesture is on by default with no extra flag wiring). The handler rotates only the *current* slice node's `SliceToRAS` about its normal through the renderer-center pivot — `ProcessRotateIntersectingSlices` is left untouched, so other slices' orientations are unaffected. The angle-vs-pivot helper `GetSliceRotationAngleRad` and the `rotationDirection` determinant convention are reused from `ProcessRotateIntersectingSlices`. The rotation sign is negated relative to that helper because we are rotating the view's own frame (rather than the *other* slices' frames as `TransformIntersectingSlices` does), so the displayed image follows the cursor.

**Flip horizontal — right-click → Flip horizontal in slice views.** Adds a new `QAction` to `qSlicerSubjectHierarchyViewContextMenuPlugin`, shown only for slice view nodes and positioned alongside `Reset field of view`. The slot mirrors the displayed image left-right by post-multiplying `SliceToRAS` by a 180-degree rotation about the in-plane vertical axis (`vtkTransform::RotateY(180)`), pivoted at the current FoV-center slice coordinate (`XYZOrigin[0]`, which `vtkMRMLSliceNode::UpdateMatrices` shows equals the slice X at the window center because the `-FieldOfView/2` and `spacing * (Dim/2)` terms cancel). This is the same `vtkTransform`-180-degree idiom SlicerIGT's `VolumeResliceDriver` uses for its `Flip` checkbox; the in-plane horizontal axis and the slice normal are negated together so the matrix stays right-handed. The action is a momentary toggle: triggering it again restores the original orientation. Pivoting at the FoV center means content at the window center stays put even when the user has zoomed in and panned off the slice origin; when not panned, this reduces exactly to the simpler `RotateY(180)` form.

## Behavior change — Blend drag relocation

Because Ctrl+Left was previously bound to **Blend** (foreground/label opacity click-and-drag) in `vtkMRMLSliceIntersectionWidget`, the Blend *drag* has been moved to **Ctrl+Shift+Left**. The `t` (foreground) and `g` (label) opacity-toggle keys are unchanged. Users who depend on the previous Blend gesture can disable the new rotate by clearing `ActionRotate` via `SetActionEnabled` — the same per-action bitmask mechanism Slicer already exposes for `Translate`, `Zoom`, `Blend`, etc.

**Why not keep Blend on Ctrl+Left and mode-gate it?** The natural pattern is the one `vtkMRMLWindowLevelWidget` already follows — self-gate on `vtkMRMLInteractionNode::AdjustWindowLevel` mode and win by returning a smaller `distance2` than the intersection widget's `1e10`. The blocker is that Ctrl+Left is *already* claimed in `AdjustWindowLevel` mode by `WidgetEventAdjustWindowLevelAlternativeStart` (the region-based W/L gesture) at `distance2 = 1e9`, so a mode-gated Blend at `1e10` would still be shadowed by region-grow and unreachable. Making it reachable would require also relocating region-grow off Ctrl+Left or introducing Blend as a peer sub-mode of the `AdjustWindowLevelMenu` — both larger surgeries that did not seem worth bundling with this change. Happy to revisit if reviewers prefer a different resolution; see "Open questions" below.

## Files changed

- `Libs/MRML/DisplayableManager/vtkMRMLSliceIntersectionWidget.h` — new `WidgetStateRotateFieldOfView`, `WidgetEventRotateFieldOfViewStart`/`End`, and `ProcessRotateFieldOfViewStart` / `ProcessRotateFieldOfView` method declarations.
- `Libs/MRML/DisplayableManager/vtkMRMLSliceIntersectionWidget.cxx` — event mapping (Ctrl+Left → rotate; Blend relocated to Ctrl+Shift+Left), dispatch in `ProcessInteractionEvent` / `ProcessMouseMove`, drag-state lists in `CanProcessInteractionEvent` / `Leave`, and the two new handlers.
- `Modules/Loadable/SubjectHierarchy/Widgets/qSlicerSubjectHierarchyViewContextMenuPlugin.{h,cxx}` — new `FlipSliceViewHorizontalAction`, `flipSliceViewHorizontal()` slot, and the FoV-center-pivoted `RotateY(180)` flip.

No public MRML API changes; no new MRML node types; no new settings or scene attributes.

## Backward compatibility

- `ActionRotate` (the flag gating the new gesture) was already declared `/* not used */` in `ActionsEnabled` and was already included in `ActionAll`. Existing applications/extensions are unaffected unless they explicitly call `SetActionsEnabled` with a custom mask omitting the rotate bit, in which case the new gesture stays off — matching their explicit configuration.
- Existing 2D pan/zoom/intersecting-slice gestures and the `t`/`g` opacity-toggle keys are unchanged.
- The flip action is purely additive; nothing was renamed or removed in the context-menu plugin.
- Default flip behavior with no pan (`XYZOrigin == 0`) is identical to a plain `RotateY(180)`; the FoV-center pivot only changes behavior when the user has actually panned off the slice origin.

## Test plan

Manual verification I performed locally on Linux (GCC 15, Qt 6.10):

- [x] Ctrl+Left-drag in a slice view rolls the image in-plane about the renderer center; release ends the drag. Confirmed on Axial, Sagittal, and Coronal views (the `rotationDirection` determinant handles both handedness conventions).
- [x] Ctrl+Shift+Left-drag adjusts foreground/label opacity (vertical = foreground, horizontal = label) as Blend did before; `t`/`g` toggles still work.
- [x] Right-click → Flip horizontal mirrors the image left-right. After zooming in and panning off the slice origin, the content at the window center stays put across a flip. Two flips return to the original orientation.
- [x] Linked slices, intersecting-slice lines, and segmentation/volume overlays continue to render correctly after rotate or flip.

No automated tests added. If reviewers prefer, we can add a unit test under `Libs/MRML/DisplayableManager/Testing/` that drives `ProcessRotateFieldOfViewStart` / the flip slot on a synthetic slice node and asserts the resulting `SliceToRAS`.

## i18n

The two new user-facing strings (`"Flip horizontal"` and its tooltip) go through `tr()`, consistent with the rest of the context-menu plugin. The rotate path adds no user-visible strings (it is a gesture only).

## Open questions for reviewers

1. **Blend migration to Ctrl+Shift+Left.** The main behavior change. Alternatives I weighed: (a) mode-gate Blend on `AdjustWindowLevel` (blocked by region-grow at higher priority on Ctrl+Left in that mode); (b) move *Rotate* to a different gesture (e.g. Alt+Left, currently free) so Blend stays on Ctrl+Left — breaks parity with the 3D Ctrl+Left roll; (c) keep migration as-is. Maintainer preference here decides the final binding.
2. **Flip pivot.** Currently FoV center, matching the rotate pivot and the intuition that "what I'm looking at stays put." Alternatives: pivot at the volume center, or at the cursor position at trigger time.
3. **Slice controller button for flip?** The flip lives only in the right-click context menu today. A toolbar button alongside `Rotate to volume plane` in the slice controller would improve discoverability — easy to add as a follow-up commit if desired.

## Screenshots

[Screencast_20260528_153138.webm](https://github.com/user-attachments/assets/31598cb5-481e-4981-8b07-13cad9c70118)

Screenshot_20260528_152949

### DIFF
--- a/Libs/MRML/DisplayableManager/vtkMRMLSliceIntersectionWidget.cxx
+++ b/Libs/MRML/DisplayableManager/vtkMRMLSliceIntersectionWidget.cxx
@@ -161,10 +161,26 @@ void vtkMRMLSliceIntersectionWidget::UpdateInteractionEventMapping()
   {
     this->SetEventTranslation(WidgetStateIdle, vtkCommand::MouseMoveEvent, vtkEvent::ShiftModifier, WidgetEventSetCrosshairPositionBackground);
   }
+  if (this->GetActionEnabled(ActionRotate))
+  {
+    // Ctrl-left-click-and-drag rolls the current slice in-plane about the view center
+    // (the 2D analog of the ctrl-left-click-and-drag camera roll in 3D views).
+    this->SetEventTranslationClickAndDrag(WidgetStateIdle,
+                                          vtkCommand::LeftButtonPressEvent,
+                                          vtkEvent::ControlModifier,
+                                          WidgetStateRotateFieldOfView,
+                                          WidgetEventRotateFieldOfViewStart,
+                                          WidgetEventRotateFieldOfViewEnd);
+  }
   if (this->GetActionEnabled(ActionBlend))
   {
-    this->SetEventTranslationClickAndDrag(
-      WidgetStateIdle, vtkCommand::LeftButtonPressEvent, vtkEvent::ControlModifier, WidgetStateBlend, WidgetEventBlendStart, WidgetEventBlendEnd);
+    // Blend drag uses ctrl+shift+left because ctrl+left is reserved for slice roll (ActionRotate).
+    this->SetEventTranslationClickAndDrag(WidgetStateIdle,
+                                          vtkCommand::LeftButtonPressEvent,
+                                          vtkEvent::ControlModifier + vtkEvent::ShiftModifier,
+                                          WidgetStateBlend,
+                                          WidgetEventBlendStart,
+                                          WidgetEventBlendEnd);
     this->SetKeyboardEventTranslation(WidgetStateIdle, vtkEvent::NoModifier, 0, 0, "g", WidgetEventToggleLabelOpacity);
     this->SetKeyboardEventTranslation(WidgetStateIdle, vtkEvent::NoModifier, 0, 0, "t", WidgetEventToggleForegroundOpacity);
   }
@@ -369,6 +385,7 @@ bool vtkMRMLSliceIntersectionWidget::CanProcessInteractionEvent(vtkMRMLInteracti
   if (this->WidgetState == WidgetStateMoveCrosshair                             //
       || this->WidgetState == WidgetStateTranslate                              //
       || this->WidgetState == WidgetStateRotateIntersectingSlices               //
+      || this->WidgetState == WidgetStateRotateFieldOfView                      //
       || this->WidgetState == WidgetStateBlend                                  //
       || this->WidgetState == WidgetStateTranslateSlice                         //
       || this->WidgetState == WidgetStateZoomSlice                              //
@@ -475,6 +492,15 @@ bool vtkMRMLSliceIntersectionWidget::ProcessInteractionEvent(vtkMRMLInteractionE
       processedEvent = this->ProcessRotateIntersectingSlicesStart(eventData);
       break;
     case WidgetEventRotateIntersectingSlicesEnd:
+      processedEvent = this->ProcessEndMouseDrag(eventData);
+      this->SliceLogic->EndSliceNodeInteraction();
+      break;
+    case WidgetEventRotateFieldOfViewStart:
+      this->SliceLogic->GetMRMLScene()->SaveStateForUndo();
+      this->SliceLogic->StartSliceNodeInteraction(vtkMRMLSliceNode::SliceToRASFlag);
+      processedEvent = this->ProcessRotateFieldOfViewStart(eventData);
+      break;
+    case WidgetEventRotateFieldOfViewEnd:
       processedEvent = this->ProcessEndMouseDrag(eventData);
       this->SliceLogic->EndSliceNodeInteraction();
       break;
@@ -630,6 +656,7 @@ void vtkMRMLSliceIntersectionWidget::Leave(vtkMRMLInteractionEventData* eventDat
       || this->WidgetState == WidgetStateZoomSlice                              //
       || this->WidgetState == WidgetStateBlend                                  //
       || this->WidgetState == WidgetStateRotateIntersectingSlices               //
+      || this->WidgetState == WidgetStateRotateFieldOfView                      //
       || this->WidgetState == WidgetStateTranslateIntersectingSlicesHandle      //
       || this->WidgetState == WidgetStateTranslateSingleIntersectingSliceHandle //
       || this->WidgetState == WidgetStateRotateIntersectingSlicesHandle         //
@@ -663,6 +690,7 @@ bool vtkMRMLSliceIntersectionWidget::ProcessMouseMove(vtkMRMLInteractionEventDat
   switch (this->WidgetState)
   {
     case WidgetStateRotateIntersectingSlices: this->ProcessRotateIntersectingSlices(eventData); break;
+    case WidgetStateRotateFieldOfView: this->ProcessRotateFieldOfView(eventData); break;
     case WidgetStateTranslate:
     {
       const double* worldPos = eventData->GetWorldPosition();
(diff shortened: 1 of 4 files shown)


pieper:
Can you explain the specific use case where people would need this feature?  I ask because I'm very reluctant to introduce changes that make it "easy" to end up in non-radiological convention.  We don't want people to be making left-right flip accidents.  I think this new behavior should be opt-in for people working with data that's not patient-aligned as part of the clinical scanning process.

Also, do you really need to change the binding for the opacity blend?  Why not make the new behavior use the new modifier sequence?

lassoan:
1. Flips

If the reason for the flip is to fix appearance of an incorrectly acquired image (imaging tech set wrong patient orientation at the time of scanning) then the solution is to apply and harden a transform on the volume.

If the reason for horizontal flip is that in a specific application the user prefers radiology/neurology convention then the solution is to use Edit / Application settings / Views / Slice viewer defaults / View orientation (patient right is screen left / patient right is screen right).

For 3D images I don't really know any other good reasons.

In case of 2D images, a flip is more commonly needed. For example, prostate TRUS you would want to flip the image so that the transducer is at the bottom, but for transperineal US you would probably want the transducer to appear at the top. If the ultrasound is not tracked then we don't know patient left/right direction, so horizontal flip may be needed. Reformat module has many reorientation options, including H and V flip, rotations, autorotate based on 3D view, etc. Would this be sufficient? If we make it available more prominently in the slice view controller then we may risk user misusing it on 3D images. If we added default-on orientation markers (A/P, I/S, L/R, and oblique combinations) to slice view sides then it would be less risk to allow flips. Maybe creating reusable widget(s) from controllers in the Reformat module GUI would be a good solution? Then those widgets could be easily added to any custom module.

2. More accessible slice rotation

Slice views preferably should not be rotated (for reasons similar to flip). Also, Ctrl-left-click-and-drag for blending is widely used, so I would not change this in Slicer core.

If you want to remap GUI events to actions, it may be better to implement easier configurability of this mapping.

ruffsl:
> Can you explain the specific use case where people would need this feature?

In developing a UI for image guided surgery, multiple surgeons (most familiar with GE & Siemens systems) we've have requested a matching UX for rotating the FoV of the 2D slice views like they can with the 3D view, so they can refine the planned MPR views to match their current PoV when standing next to patient.

> I think this new behavior should be opt-in 

Is there a pattern of an existing opt-in option I could replicate?

> Also, do you really need to change the binding for the opacity blend?

In the current PR, I opted for consistency across view controls give the Principle of Least Astonishment (POLA) from a UI/UX perspective. As noted it the Open questions, we could try gating or re-prioritizing the actions, but that would touch more of the codebase.

ruffsl:
> If the reason for horizontal flip is that in a specific application the user prefers radiology/neurology convention then the solution is to use Edit / Application settings / Views / Slice viewer defaults / View orientation (patient right is screen left / patient right is screen right).

The use case here was for reorientation view to so a surgeon could match their PoV as they move around the patient, thus being more dynamic and thus not as appropriate for persistent  setting nested in Slicer's user preferences.

> For 3D images I don't really know any other good reasons.

The flip submenu is not applicable to 3D views, and thus doesn't show when right clicking 3D viewports.

 > Reformat module has many reorientation options, including H and V flip, rotations, autorotate based on 3D view, etc. Would this be sufficient? If we make it available more prominently in the slice view controller then we may risk user misusing it on 3D images. If we added default-on orientation markers (A/P, I/S, L/R, and oblique combinations) to slice view sides then it would be less risk to allow flips. Maybe creating reusable widget(s) from controllers in the Reformat module GUI would be a good solution? Then those widgets could be easily added to any custom module.
 
 Are you suggesting something akin to `3D axis labels`/`3D cube` in 3D views but for 2D slice views?
 
 
image

> Also, Ctrl-left-click-and-drag for blending is widely used, so I would not change this in Slicer core.

At first I wanted to keep the key bindings for blending unchanged by gating to `Adjust window/level` thus allow rotation to use the same key bindings when the left-click menu mode is set to `View transforms`, but it seems that `Ctrl-left-click-and-drag` is also already used for AdjustWindowLevelAlternative in `Adjust window/level` mode.
 
 > If you want to remap GUI events to actions, it may be better to implement easier configurability of this mapping.
 
 That could work. Any prior art on that?

lassoan:
> The use case here was for reorientation view to so a surgeon could match their PoV as they move around the patient, thus being more dynamic and thus not as appropriate for persistent setting nested in Slicer's user preferences.

For any so specific use case like this you probably cannot avoid creating a custom module that configures and automates everything. You would not want to ask the user to flip individual views, but the user would just select in the beginning what kind of procedure he is doing and which side of the patient he is standing and all the views would be set up automatically. You can also leverage patient-attached tracking markers (that are attached to the patient in a predefined anatomical orientation) and tool tracking (the views can be automatically set up to the surgeon's viewpoint if he holds up the stylus or cautery).

If you need a GUI just for the initial phantom experiments, to figure out what you need, you can probably manage with existing features (it is just not optimal, as you need to keep switching between several modules).

> > For 3D images I don't really know any other good reasons.
> 
> The flip submenu is not applicable to 3D views, and thus doesn't show when right clicking 3D viewports.

I did not mean 3D views, but 3D images. What we experienced most often was that clinicians were very confused when we showed them oblique or rotated slice views, because they usually look at 3D images in standard orientations.

> Are you suggesting something akin to 3D axis labels/3D cube in 3D views but for 2D slice views?

I'm suggesting the standard letter symbols at least two sides of the slice views.

image

It could be added to the vtkMRMLOrientationMarkerDisplayableManager.

> That could work. Any prior art on that?

In custom applications, usually it is not too difficult to remap keyboard/mouse gestures manually (https://slicer.readthedocs.io/en/latest/developer_guide/script_repository.html#customize-keyboard-shortcuts).
However, it would simplify implementation of custom applications (and would make it easier for every user to remap keyboard/mouse gestures) if default event translation would not be hardcoded in the 3D widget constructors (see for example [1](https://github.com/Slicer/Slicer/blob/b22c472ad0afee6a646e78f1e5c68a35ba163197/Modules/Loadable/Transforms/MRMLDM/vtkMRMLTransformHandleWidget.cxx#L52-L60), [2](https://github.com/Slicer/Slicer/blob/b22c472ad0afee6a646e78f1e5c68a35ba163197/Modules/Loadable/Markups/VTKWidgets/vtkSlicerAngleWidget.cxx#L32-L35), [3](https://github.com/Slicer/Slicer/blob/b22c472ad0afee6a646e78f1e5c68a35ba163197/Libs/MRML/DisplayableManager/vtkMRMLWindowLevelWidget.cxx#L79-L93), [4](https://github.com/Slicer/Slicer/blob/b22c472ad0afee6a646e78f1e5c68a35ba163197/Libs/MRML/DisplayableManager/vtkMRMLSliceIntersectionWidget.cxx#L137-L290)), but the constructors would get them from the application logic (that would read it from a json configuration file).

Issue #7960, opened 2024-09-24
Remove .fcsv as file save as option


muratmaga:
## Summary

Markups module has lots of features that are not fulfilled with legacy fcsv format, and causing unexpected issues: https://discourse.slicer.org/t/trouble-saving-fcsv-file-as-part-of-scene/38464/7

This can be avoided using the mrk.json format. 

I propose to remove fcsv as a file save option, and only keep it as a Export As option. If this is too complex, and remove it all toget and simply point out in the documentation that exporting to a tabular format is possible in Import/Export tab of the Markups module.

lassoan:
I would be happy to get rid of fcsv file support, but we should probably notify users through the Slicer forum (create a new topic for this, explaining what we plan to do and why and ask for feedback). "Export" popup and "Save data" dialog uses the same mechanism (writer plugins), so it would be hard to disable one but not the other. What we could easily do is to keep fcsv reading available but remove the possibility to write or export fcsv.

lassoan:
A warning is already logged when the user chooses to save in .fcsv format. I'll add to the warning message that the format is deprecated and support will be removed in future software versions. Then we can put this issue in the backlog and come back in a couple of years and remove fcsv support.

PR #9192, opened 2026-05-28
ENH: Deprecate fcsv file format

lassoan:
Let users know that fcsv file format is deprecated, in documentation and file saving warning message.

Related to request of removing fcsv support in #7960

image

### DIFF
--- a/Libs/MRML/Core/vtkMRMLMarkupsFiducialStorageNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLMarkupsFiducialStorageNode.cxx
@@ -741,7 +741,13 @@ int vtkMRMLMarkupsFiducialStorageNode::WriteDataInternal(vtkMRMLNode* refNode)
     return 0;
   }
 
-  this->GetUserMessages()->AddMessage(vtkCommand::WarningEvent, "fcsv file format only stores control point coordinates and a limited set of display properties.");
+  vtkWarningToMessageCollectionMacro(this->GetUserMessages(),
+                                     "vtkMRMLMarkupsFiducialStorageNode::WriteDataInternal",
+                                     vtkMRMLTr("vtkMRMLMarkupsFiducialStorageNode",
+                                               "The fcsv format is deprecated and it will not be supported in future software versions.\n"
+                                               "Use JSON file format (.mrk.json) instead.\n"
+                                               "If using fcsv file format then only control point coordinates"
+                                               " and a limited set of display properties will be saved."));
 
   // open the file for writing
   fstream of;
--- a/Docs/developer_guide/modules/markups.md
+++ b/Docs/developer_guide/modules/markups.md
@@ -17,6 +17,16 @@ A simple example that specifies a markups point list with 3 points that can be s
 
 ## Markups fiducial point list file format (.fcsv)
 
+:::{warning}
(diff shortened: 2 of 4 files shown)


muratmaga:
@lassoan there is a small typo on the yellow warning box.  it should be mrk.json not mkp.json.

Libs/MRML/Core/vtkMRMLMarkupsFiducialStorageNode.cxx
```diff
@@ -741,7 +741,13 @@ int vtkMRMLMarkupsFiducialStorageNode::WriteDataInternal(vtkMRMLNode* refNode)
     return 0;
   }
 
-  this->GetUserMessages()->AddMessage(vtkCommand::WarningEvent, "fcsv file format only stores control point coordinates and a limited set of display properties.");
+  vtkWarningToMessageCollectionMacro(this->GetUserMessages(),
+                                     "vtkMRMLMarkupsFiducialStorageNode::WriteDataInternal",
+                                     vtkMRMLTr("vtkMRMLMarkupsFiducialStorageNode",
+                                               "The fcsv format is deprecated and it will not be supported in future software versions.\n"
```
hjmjohnson, line 747: @lassoan I agree with this deprecation.  It would be nice to provide a pointer to a conversion script to convert from .fcsv to .json in the comment.  I have several public datasets that are in active use that have fcsv files from 12 years ago.  Changing the public data set is not feasible.  A simple python script probably already exists somewhere.

Thanks.

hjmjohnson (dismissed):
Approved after fixing minor typo fix.    A pointer to a conversion script would be nice, but not necessary.

jamesobutler:
image

@hjmjohnson You brought in a merge commit which resulted in the commit message check failing. Slicer enforces linear history so merge commits are not allowed.

lassoan:
Thanks for the feedback. I've fixed the typo and added a legacy .fcsv to .mrk.json converter to [slicerio](https://github.com/lassoan/slicerio#convert-legacy-markup-fiducial-file-fcsv-to-current-format-mrkjson).

Libs/MRML/Core/vtkMRMLMarkupsFiducialStorageNode.cxx
lassoan: Done, added to slicerio - https://github.com/lassoan/slicerio#convert-legacy-markup-fiducial-file-fcsv-to-current-format-mrkjson

jamesobutler:
@lassoan The new deprecation warning is causing [`vtkMRMLMarkupsStorageNodeTest2`](https://slicer.cdash.org/tests/33240144) to fail because there are new unexpected numbers of warnings.

https://github.com/Slicer/Slicer/blob/2080f7fc4c71f1dcc172b7041fbf24a0276bb482/Modules/Loadable/Markups/Testing/Cxx/vtkMRMLMarkupsStorageNodeTest2.cxx#L273-L275

Probably need to add the following type to confirm the expected warnings
```
TESTING_OUTPUT_ASSERT_WARNINGS_BEGIN();
[code here]
TESTING_OUTPUT_ASSERT_WARNINGS(1); // Expected warning; writing a fcsv is deprecated
TESTING_OUTPUT_ASSERT_WARNINGS_END();
```

PR #9194, opened 2026-05-29
BUG: Disable default VTK render window keyboard shortcuts

### LINKED ISSUE #8368: Pressing '3' Key Triggers Red/Blue Stereo Effect Even When Stereo Viewing is Disabled
Hi @lassoan  @pieper 

## Summary

When pressing the '3' key in both 2D and 3D views, a red/blue stereo effect appears, even though the Stereo Viewing option is set to 'No Stereo'. The expected behavior is that the stereo effect should not be applied when Stereo Viewing is disabled.

## Steps to reproduce

1. Open the application.
2. Ensure that Stereo Viewing is set to No Stereo in the 3D view settings.
3. Switch to either 2D or 3D view.
4. Press the '3' key.

# Actual Behavior:
- A red/blue stereo effect appears, even though the Stereo Viewing option is disabled.

# Expected Behavior:
- No stereo effect should be displayed when the Stereo Viewing option is set to No Stereo.

# Additional Notes:

- This issue occurs in both 2D and 3D views.
- Confirmed that the Stereo Viewing setting is correctly applied before testing.

# Related discussions

- Slicer post: https://discourse.slicer.org/t/problem-with-stereo-viewing-mode-in-3d-view/42428
- VTK post: https://discourse.vtk.org/t/question-on-vtkrenderwindowinteractor/15469

## Environment
- Slicer version: Slicer-5.8.1-2025-03-03 & older
- Operating system: Windows

lassoan:
VTK render window interactor base class registers default low-level keyboard shortcuts for toggling stereo mode, wireframe rendering, etc., which would interfere with rendering set up by displayable managers.

There was an attempt to disable these shortcuts but it was not effective, as if the custom interactor style does not abort the event then the default one gets to process it.

fixes #8368

### DIFF
--- a/Libs/MRML/DisplayableManager/vtkMRMLViewInteractorStyle.cxx
+++ b/Libs/MRML/DisplayableManager/vtkMRMLViewInteractorStyle.cxx
@@ -87,9 +87,11 @@ void vtkMRMLViewInteractorStyle::OnKeyRelease()
 //----------------------------------------------------------------------------
 void vtkMRMLViewInteractorStyle::OnChar()
 {
-  // Do not call this->GetInteractorStyle->OnChar(), because char OnChar events perform various
-  // low-level operations on the actors (change their rendering style to wireframe, pick them,
-  // change rendering mode to stereo, etc.), which would interfere with displayable managers.
+  // Abort CharEvent so VTK's vtkInteractorStyle::OnChar() does not get invoked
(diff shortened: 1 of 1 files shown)


cpinter:
What a coincidence. Just yesterday I was surprised what happens when I press 3 :) Thanks Andras!

PR #9196, opened 2026-05-29
ENH: Add application icon for linux

### LINKED ISSUE #8389: Put Slicer Icon a fixed path.
In linux there is no launcher for Slicer. One can use .desktop files to create one, however, the only Slicer icon embedded in that Slicer tar file, has a variable path, specifically 
**Slicer-X.Y**/qt-scripted-modules/Resources/SlicerWatermark.png

This breaks the icon path when the minor or major versions change (obviously the former is a bigger concern). 

Obviously an icon is not necessary for launcher, but it will help with the visual recognition. 

I suggest adding the same icon in the same folder where the Slicer launcher is. That way it will be possible to create more robust launchers (they only will need to change the root location of the archive), and don't have to programmatically parse the tree to figure out the correct minor and major versions. 

@jcfr

lassoan:
In linux there is no launcher for Slicer, but one can use .desktop files to create one. However, the only Slicer icon embedded in that Slicer tar file, has a variable path (for example, Slicer-X.Y/qt-scripted-modules/Resources/SlicerWatermark.png).

This breaks the icon path when the minor or major versions change.

This commit adds an icon next to the launcher, with the same filename as the launcher, with the extension of the Slicer application icon (by default a .png file, can be specified by LINUX_ICON_FILE argument in slicerMacroBuildApplication).

fixes #8389


ebrahimebrahim:
I have been wanting this for a while, yay!

pieper (approved):
Sounds nice 👍

Issue #8402, opened 2025-04-30
Segment selection changes in Segments Table when clicking the eye button


cpinter:
## Summary

Selection of the current edited segment changes when toggling visibility of any segment using the eye icon in the Segments Table. This is an annoyance for users, see more detailed description here:

https://discourse.slicer.org/t/segment-editor-clicking-eye-to-toggle-visibility-should-not-select-segment/20237

## Steps to reproduce

1. Load MRHead
2. Go to Segment Editor
3. Create a segment and paint something in it
4. Create another segment and paint something in it
5. Click the eye icon of Segment_1 to hide it
6. Click in the slice viewer to paint
7. Warning appears about the current segment being hidden

It would be better from the usability point of view if clicking the eye didn't change the segment selection.

The problem is that the eye icon is not an actual button, it is a "decoration" of the table cell, clicking on which triggers selection and also visibility.
https://github.com/Slicer/Slicer/blob/3b2948fb04b431468f5ef809f8e02d621eaa6397/Modules/Loadable/Segmentations/Widgets/qMRMLSegmentsTableView.cxx#L467-L472

Maybe a solution would be to add an actual button in the table cell, and handle it in a separate callback function for the button click. A problem I already see is that the button will need more space, but it can probably be prevented with some tricks (maybe negative margins in a style sheet, just an idea).

Related to ticket https://github.com/Slicer/Slicer/issues/6975

## Environment
- Slicer version: Slicer-5.8.1
- Operating system: Windows 10

jamesobutler:
Xref https://github.com/Slicer/Slicer/issues/5290 which has discussed this type of topic of prevent ting selection change when trying to toggle visibility.

Issue #5290, opened 2020-11-02
Toggling visibility shouldn't affect selection (segmentations, markers, models, etc...)


fusentasticus:
I believe this is a bug, not a feature request, in so far as the behavior is inconsistent with standard UI design (afaik).

In the Segmentation Editor, when toggling visibility of a segment, the selection should not be affected; in particular, it should not be changed to that of the segment whose visibility was changed. 

This leads (for me) to very frequent operations of the wrong segment and it's sometimes hard to discover that the operation was misapplied.

The problem is general to all of the 3D slicer UI it seems:  toggling visibility for markups, models, etc. affects the selection. I think the correct behavior is that it shouldn't.

## Environment
- Slicer version: Slicer-4.11
- Operating system: Windows / Linux / Mac + which version

lassoan:
Good point. I've checked Blender and Gimp and they both allow changing visibility of objects/layers without making it selected. (Paraview does not change selection if you turn off visibility, but selects the object when its visibility is turned on.)

lassoan:
We have a pull request under review (https://github.com/Slicer/Slicer/pull/5274) to warn the user if he wants to edit a non-visible segment.

ottensmeyer:
Thanks Andras for the forum response - this appears to be the same issue - adding the comment and response here.

"In the Segment Editor, in 4.10.2, when one segment is selected, one can toggle visibility (change open/closed eye icon state) of another segment without simultaneously selecting the other segment. In 4.11, if I click the eye icon to toggle visibility of some segment, I lose focus on the segment on which I’m actually working, leading me to accidentally start editing the segment that I want to see but not edit.

In 4.10.2, hovering over the eye icon shows a dotted border around the visibility icon, which does not appear in 4.11.

Is this an intended behavior new behavior I’ll need to get used to? At the moment, I keep on clicking the icon to see another segment for reference, and start accidentally editing that one instead of the one I want to change."

lassoan:
"We reworked the segment list to add many new features (searching, filtering, states) and along the way management of checkboxes changed. I agree that the previous behavior was better, but it is not trivial to achieve that again in the new architecture. I would suggest to submit a feature request to the Slicer issue tracker."

DIV-on-github:
[...]

PR #9198, opened 2026-05-30
BUG: Clicking visibility icon does not change item selection

cpinter:
When clicking the eye icon in the subject hierarchy tree view, the base class mousePressEvent was called before the decoration click handler, causing the item selection to change unintentionally. Fixed by checking for a visibility column decoration click first; if handled, the base class event is skipped so the current selection is preserved.

Re #5290, #8402


cpinter:
I put this in the 5.12 milestone because there are two issues related, and also because @muratmaga has requested this quite a lot and would be useful in his projects.

lassoan:
I agree that keeping the current selection is more common, so this change is for the better.

The issues refer to the segment list (in Segment Editor, etc.) as well, so I think we should change the behavior of that widget as well. I already have an implementation, I'll push it to this branch.

lassoan (dismissed):
Thank you . I've fixed the weird behavior when you moved the mouse while the mouse button was still held down and also implemented this for the segment list.

lassoan (dismissed):

cpinter:
Thank you @lassoan !

lassoan:
I worked on this a bit more to make it more similar to all the other buttons in the application: making it perform the action on button release. But then I missed the visual cue of the button being depressed. There were also some edge cases with the widget state getting in inconsistent state when not receiving button up event, etc. I gave it a try adding a standard button in the widgets, and the results look perfect, buttons that appear and behave as all other buttons, and we are less exposed to various complications and regressions as we don't need to do custom low-level manipulations. I've committed the changes, haven't cleaned it up, just pushed here so that it can be tested.

https://github.com/user-attachments/assets/a1c4d7e7-b5ae-4181-8533-2fac7b4c1f5f

@cpinter please have a look. If you agree that these buttons should work like this then I'll finalize the changes.

cpinter:
I like the idea, and the code changes look good as well after a quick review. Using the decorations mechanism is more complicated than it needs to be, so if we have more control using buttons, great. I also like the visual improvement (frame around colors and indication of hover and press).

Building the code now and will test today.

lassoan:
I've reviewed the code more thoroughly, made a few more simplifications, tested a bit more. @cpinter please review and test it (I don't plan to do any more changes other than squashing all commits; the CI build error is due to the build script does not seem to update SurfaceToolbox based on the new git hash).

Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyTreeView.cxx
```diff
@@ -262,10 +264,219 @@ void qMRMLSubjectHierarchyTreeViewPrivate::init()
...
+                           button->setIcon(index.data(qMRMLSubjectHierarchyModel::VisibilityIconRole).value<QIcon>());
+                         }
+                       }
+                       if (topLeft.column() <= colorCol && colorCol <= bottomRight.column())
+                       {
+                         QModelIndex index = this->SortFilterModel->index(row, colorCol, parent);
+                         auto* button = qobject_cast<QToolButton*>(q->indexWidget(index));
+                         if (button)
+                         {
+                           button->setIcon(QIcon(qMRMLUtils::createColorPixmap(qApp->style(), //
```
cpinter, line 312: Why do we need these `//`? I can see them above and in other files as well

SuperBuild.cmake
```diff
@@ -335,7 +335,7 @@ list_conditional_append(Slicer_BUILD_LandmarkRegistration Slicer_REMOTE_DEPENDEN
 
 Slicer_Remote_Add(SurfaceToolbox
   GIT_REPOSITORY "${EP_GIT_PROTOCOL}://github.com/Slicer/SlicerSurfaceToolbox"
-  GIT_TAG 172483af41cf22c39ac77c07485d5f7ef8b4ffdc
```
cpinter, line 338: Is this somehow needed for the change? I started only an inner build, please let me know if this is not just an accidentally included file.

cpinter: Yes it is needed. Starting SuperBuild...

cpinter:
One strange behavior is that in the segments table we always have focus on an item after clicking the terminology button.
Steps: Load MRHead, go to Segment Editor, create three segments, click any of the three color buttons, click Cancel.

Not sure if this is something important to handle now, but in any case it caught my attention. (Ah, and the build I have is with Qt6)

image

I keep testing...

Modules/Loadable/SubjectHierarchy/Widgets/qMRMLSubjectHierarchyTreeView.cxx
lassoan: Because the linter would require so long line that the pre-commit check would fail due to line length. The only way to fix it is to rewrite the code or adding these hard line breaks.

lassoan:
> One strange behavior is that in the segments table we always have focus on an item after clicking the terminology button. 

I don't see this focus behavior on Qt5. We can fine-tune the Qt6 behavior when we switch the Slicer build to use Qt6 (in a couple of weeks).

lassoan:
There was an agreement at the developer meeting that the new behavior looked good, my and Csaba's testing did not uncover issues (and the build error is false alarm - the build system does not update SurfaceToolbox external project), so I'll proceed with the merge.

lassoan (approved):

cpinter:
Excellent, thank you so much @lassoan !

PR #9205, opened 2026-06-03
BUG: Reset view action metadata when opening SH widget context menu

Sunderlandkyl:
Event metadata from view events wasn't cleared when opening the context menu in SH widgets. This caused the wrong transform node to be selected when opening the context menu if it had been previously opened in the views.

Fixed by clearing the view based metadata before opening the context menu.

Fix #9082


lassoan (commented):
Thank you, it looks good to me. Could you please check if similar changes are needed in qSlicerSubjectHierarchyMarkupsPlugin and qSlicerSubjectHierarchyViewContextMenuPlugin as well?

Sunderlandkyl:
qSlicerSubjectHierarchyMarkupsPlugin and qSlicerSubjectHierarchyViewContextMenuPlugin look ok to me. All the show context menu pathways already clear the metadata.

lassoan (approved):
This is all we need then. Thank you!

PR #9206, opened 2026-06-03
COMP: Update FindVcvars with support for old toolsets in newer VS

jamesobutler:
This brings in the changes made in https://github.com/scikit-build/cmake-FindVcvars/commit/f71cb982bf1905b1964080c88c7f36efea04561a.

@lassoan I came across this with my Slicer custom app that makes a FindVcvars call and I was using Visual Studio 2026 but building with the v143 (Visual Studio 2022) toolset and did not have Visual Studio 2022 installed side-by-side.

### DIFF
--- a/CMake/FindVcvars.cmake
+++ b/CMake/FindVcvars.cmake
@@ -390,7 +390,10 @@ function(Vcvars_GetVisualStudioPaths msvc_version msvc_arch output_var)
     string(REGEX REPLACE "^([0-9]+)\.[0-9]+$" "\\1" vs_installer_version ${vs_version})
     # Probe the matching VS version and any newer versions, since a newer VS installation
     # (e.g., VS 2026) may host older toolsets (e.g., v143) as optional components.
-    set(_known_vs_major_versions 15 16 17 18)
+    set(_known_vs_major_versions 15 16 17)  # VS 2017, 2019, 2022
+    if(CMAKE_VERSION VERSION_GREATER_EQUAL "4.2.0")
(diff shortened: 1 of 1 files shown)


lassoan (approved):

The hardcoded `15 16 17 18` is not the nicest, but it is fine, as long as we get a meaningful error message when build fails when trying to use VS 19. The text or location of the error message should make it very easy to figure out that `19` has to be added to this line.

lassoan (changes requested):
One issue with this update is that somebody is not using a very recent CMake, then `cmake_host_system_information(RESULT _vs_dir QUERY VS_18_DIR)` will stop with a hard failure (`cmake_host_system_information does not recognize <key> VS_18_DIR`). It could be worked around by running the `cmake_host_system_information` call in a subprocess. Considering how inconvenient this is, I would consider this a bug that should be reported to CMake developers.

If we want a robust solution that does not require hardcoding of VS versions and can display a meaningful error message (such as `VS_19 is found but not supported by your CMake version 3.31.7`) then we could use `vswhere` to query all the available VS versions.

jamesobutler:
@lassoan In the short term I have added a CMake guard check to only search for VS18 paths with CMake 4.2 or newer which added VS2026 support.

hjmjohnson (dismissed):
LGTM.

jamesobutler:
@lassoan I fixed the typo I discussed earlier today regarding the VS2026 version "18" was still outside the new CMake version guard.

lassoan (approved):
Thanks, it looks good to me.

PR #9220, opened 2026-06-08
BUG: Restore slicer.util Python console imports

### LINKED ISSUE #9182: Restore slicer.util imports in python console
We should restore the legacy behavior in terms of things in `slicer.util` that were previously automatically imported in the slicer python console. Import everything, or almost everything, from slicer.util explicitly [here](https://github.com/Slicer/Slicer/blob/208adb862bc329e6370ae400fef1b18e9ac54cbd/Base/Python/slicerqt.py#L14-L20), or do a wildcard import but explicitly control `__all__` for `slicer.util`. We may also want to consider importing vtk, itk. Ultimately, the python console should be ergonomic.

Behavior was changed in #9010 after fixing a CTK bug and removing a then-problematic wildcard import.

cc @pieper @lassoan @Thibault-Pelletier

ebrahimebrahim:
Define `slicer.util.__all__` with public utility functions and import them into slicerqt.py for Python console convenience, matching the Slicer 5.10 behavior.

Fixes #9182


Base/Python/slicerqt.py
```diff
@@ -11,13 +11,7 @@
...
 import slicer
 
-from slicer.util import (  # noqa: F401
-    array,
-    exit,
-    getNode,
-    getNodesByClass,
-    mainWindow,
-)
+from slicer.util import *
```
lassoan, line 15:
```suggestion
# Import everything from slicer.util to make frequently used functions conveniently available in the Python console
from slicer.util import *
```

lassoan (dismissed):
Thank you, it looks good to me!

lassoan (approved):

ebrahimebrahim:
THanks for review

(Fixed typo in commit message)

PR #9223, opened 2026-06-10
BUG: Fix terminology selector opening with empty content on repeated opens

xskere:
When the terminology selector dialog is opened, it reads the last-used terminology context from the application settings file under `[Terminology] / LastTerminologyContexts` and uses it to populate the dialog content. 

After the very first use of the dialog, this list contains exactly one entry. On every subsequent open, the code detects that the list already contains that one entry and returns early to avoid writing the same value back to the settings file. However, that early return also skipped the initialization of the dialog content, leaving the category list, type list, and color table all empty. 

Selecting a different context from the dropdown worked around the issue for that session by going through a different code path, but the next open would show the problem again. The bug is most noticeable on a clean install, where the settings file starts empty and the list builds from scratch, reaching exactly one entry after the first use and staying there until the user switches between contexts.

<div align=center>
image
</div>

The proposed fix remove the early return that was only there as a small optimization, the logic that follows already handles this case correctly.

To reproduce the bug, set the saved context list to a single entry from the Python console:

`slicer.app.userSettings().setValue("Terminology/LastTerminologyContexts", ["Segmentation category and type - 3D Slicer General Anatomy list"])`

Then open the terminology selector twice and confirm the category list, type list, and color table are all empty.

To verify the fix, apply the change, rebuild, and repeat the same steps.

The dialog should now open with all content correctly populated.

### DIFF
--- a/Modules/Loadable/Terminologies/Widgets/qSlicerTerminologyNavigatorWidget.cxx
+++ b/Modules/Loadable/Terminologies/Widgets/qSlicerTerminologyNavigatorWidget.cxx
@@ -938,21 +938,15 @@ void qSlicerTerminologyNavigatorWidgetPrivate::setCurrentTerminology(QString ter
         terminologyOrColorName = QString::fromUtf8(colorNode->GetName());
       }
     }
-    if (!lastTerminologyContextNames.isEmpty())
+    // Remove the terminology name from the list so that there are no duplicate entries in the list
+    lastTerminologyContextNames.removeOne(terminologyOrColorName);
+    // Prepend terminology name to the list so that the last used terminology is first
+    lastTerminologyContextNames.insert(0, terminologyOrColorName);
+    // Trim to a reasonable maximum to prevent unbounded growth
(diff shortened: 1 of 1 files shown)


cpinter:
@lassoan This fixes a bug that we talked about. Can you please also review? Maybe we can get this in before the stable. Thank you!

cpinter (dismissed):
Looks good to me, thank you!

lassoan:
With this change, the if/else does not really add any value, we could just do:

```
lastTerminologyContextNames.removeOne(terminologyOrColorName);
lastTerminologyContextNames.insert(0, terminologyOrColorName);
```

We might consider dropping the oldest entry if we have entries above some limit (e.g., 30). Just to prevent this most recently used terminology context names growing indefinitely (in case some workflows generate terminology context names dynamically).

xskere:
@lassoan I made some changes taking your comment into consideration!

I removed the if/else and added the dropping the oldest entry when it's above 30

lassoan (approved):
Looks good, thank you!

PR #9228, opened 2026-06-11
BUG: Fix missing openjp2 shared library in macOS package

lassoan:
On macOS, third-party shared libraries are bundled by the fixup_bundle step rather than by explicit install rules, so SlicerBlockInstallOpenJPEG (used on Windows/Linux) has no effect there. Add the OpenJPEG SuperBuild install lib directory to the fixup_bundle search path so libopenjp2 is resolved and copied into the .app, fixing Slicer failing to start with "Library not loaded: @rpath/lib/Slicer-5.11/libopenjp2.7.dylib".

This pull request is a draft, as building of a macOS package is still in progress.


jamesobutler:
I can confirm with today's macOS package from https://slicer.cdash.org/index.php?project=SlicerPreview&date=2026-06-12, that Slicer starts successfully now.

lassoan:
Awesome, thanks for testing. One more step closer to 5.12!

PR #9234, opened 2026-06-15
BUG:  Align CheckableComboBox and comboBox to same column

LiBu981:
Fixes the misalignment of the `qMRMLSegmentSelectorWidget` `CheckableComboBox` (_here the Target volumes_) and `comboBox` (_here the Body volume_) addressed in the SlicerRT [PR 319](https://github.com/SlicerRt/SlicerRT/pull/319).

image

### DIFF
--- a/Libs/MRML/Widgets/Resources/UI/qMRMLSegmentSelectorWidget.ui
+++ b/Libs/MRML/Widgets/Resources/UI/qMRMLSegmentSelectorWidget.ui
@@ -80,7 +80,7 @@
          <property name="spacing">
           <number>4</number>
          </property>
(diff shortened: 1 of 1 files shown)


lassoan (commented):
You cannot have two items in the same row/column positions. With this change, only one of the items show up in Qt Designer and display at runtime may break anytime.

Instead, you can experiment with different layouts and spacings (e.g., instead of a grid layout you can use vertical and horizontal layouts and set the horizontal layout spacing to 0).

jamesobutler:
Closing pending reconsideration of different layouts and spacings.

PR #9238, opened 2026-06-16
BUG: Fix crash loading table with unsupported schema column type

### LINKED ISSUE #9211: Slicer crashes loading CSV+schema into table node
## Summary

I ran into a situation of loading a CSV into a table node which then caused 3D Slicer to crash.

## Steps to reproduce

[stats.csv](https://github.com/user-attachments/files/28648686/stats.csv)
[stats.schema.csv](https://github.com/user-attachments/files/28648688/stats.schema.csv)

The above file pair is a minimal stats table saved into a csv format. My original file had more columns, but the issue can still be replicate with this minimal version.

1. Download the above file pair
2. Drag-and-drop the stats.csv into Slicer and select to load as "Table" type.
3. Observe that Slicer crashes.

If you delete the stats.schema.csv so it is not located next to stats.csv and repeat the process, the CSV successfully loads into a table node.

## Environment
- Slicer version: Slicer-5.11.0-2026-06-03
- Operating system: Windows

mauigna06:
### Problem
A table column whose schema `type` has no `vtkDataArray` representation
(e.g. `variant`, `object`) was mishandled on both load and save:

- **Loading crashed Slicer.** Dropping a `.csv` next to a matching
  `*.schema.csv` that declares such a type crashed the application.
  Without the sidecar the columns default to string, so the crash only
  appeared when both files coexisted — making it look like a
  "two files in the same folder" problem.
- **Saving produced an unloadable file.** A `variant` array was written
  to the schema as type `variant`, which then crashed on reload.

Fixes #9211 

### Root cause
`vtkMRMLTableStorageNode` auto-detects the schema and applies the
declared column types.

- On read, `AddColumnToTable()` did:

  ```cpp
  vtkSmartPointer<vtkDataArray> typedColumn =
      vtkSmartPointer<vtkDataArray>::Take(vtkDataArray::CreateDataArray(valueTypeId));
  typedColumn->SetName(columnName.c_str());   // null deref → crash

`vtkDataArray::CreateDataArray(VTK_VARIANT)` returns `nullptr` because
a variant array is not a `vtkDataArray`, and the return value was
never null-checked.

- On write, `WriteSchema()` saved each column's type verbatim via `GetValueTypeAsString(column->GetDataType())`, emitting `variant` — a type Slicer cannot reconstruct because it has no per-cell data type.

### Fix
- Read ([`AddColumnToTable`](vscode-webview://12o6fpet9apja57d6ijnfmdf31of9s46l9kkkkiaq0alnh2djgof/Libs/MRML/Core/vtkMRMLTableStorageNode.cxx)): probe whether the value type can create a `vtkDataArray`; if not, read the column as string and warn instead of crashing. Guards both the column and per-component array creation.
- Write ([`WriteSchema`](vscode-webview://12o6fpet9apja57d6ijnfmdf31of9s46l9kkkkiaq0alnh2djgof/Libs/MRML/Core/vtkMRMLTableStorageNode.cxx)): if a column is a variant array, save it as `string` and warn, since the cell values are written as text anyway and a per-cell type cannot be preserved.

### Testing
- `MRMLCore` builds and links cleanly.
- Loading `stats.csv` + `stats.schema.csv` (with `type` = `variant`) no longer crashes; the column loads as string with a warning.
- Saving a table containing a variant column writes string in the schema and round-trips without crashing.

### DIFF
--- a/Libs/MRML/Core/vtkMRMLTableStorageNode.cxx
+++ b/Libs/MRML/Core/vtkMRMLTableStorageNode.cxx
@@ -560,6 +560,18 @@ void vtkMRMLTableStorageNode::AddColumnToTable(vtkTable* table, vtkMRMLTableStor
     // schema is not defined or no valid column type is defined for column
     valueTypeId = VTK_STRING;
   }
+  if (valueTypeId != VTK_STRING //
+      && vtkSmartPointer<vtkDataArray>::Take(vtkDataArray::CreateDataArray(valueTypeId)) == nullptr)
+  {
+    // Some value types (such as 'variant' or 'object') cannot be represented by a vtkDataArray,
+    // so vtkDataArray::CreateDataArray() returns nullptr for them. Fall back to reading the column
+    // as string instead of dereferencing the null array (which would crash the application).
+    vtkWarningToMessageCollectionMacro(this->GetUserMessages(),
+                                       "vtkMRMLTableStorageNode::AddColumnToTable",
+                                       "Column '" << columnName << "' has unsupported value type '" //
+                                                  << vtkMRMLTableNode::GetValueTypeAsString(valueTypeId) << "'. The column is read as string.");
+    valueTypeId = VTK_STRING;
+  }
   if (valueTypeId == VTK_STRING)
   {
     if (rawComponentArrays.size() > 0)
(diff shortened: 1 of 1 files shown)


lassoan:
Thanks a lot @mauigna06!
This fix looks good, but it would be nice to prevent generation of a schema that cannot be read. I guess some custom code added a variant array to the table and that was saved. Since Slicer currently cannot save data type per cell, we cannot save a variant array, so when we save the table with a variant array, we should save save it as a string type and log a warning.

mauigna06:
Updated @lassoan. Thank you for the review

jamesobutler:
>  I guess some custom code added a variant array to the table and that was saved.

Yes, reviewing the code that created it, there was the following usage of vtkVariantArray:
```python
# set up stats columns
for stat_info in measurement_info.values():
    column = vtk.vtkVariantArray()
    table_node.AddColumn(column)
    column_name = stat_info.get("name", "")
    column.SetName(column_name)
    table_node.SetColumnTitle(column_name, stat_info.get("title", ""))
    table_node.SetColumnUnitLabel(column_name, stat_info.get("units", ""))
```

lassoan (approved):
This looks good to me, thank you!

PR #9240, opened 2026-06-17
ENH: Add unique Slicerrc file for custom apps with ability to turn off

jamesobutler:
This addresses the items discussed in https://discourse.slicer.org/t/cmake-flag-to-customapp-not-to-read-slicerrc-py-or-define-other-file/39046.

1. Per [comment](https://discourse.slicer.org/t/cmake-flag-to-customapp-not-to-read-slicerrc-py-or-define-other-file/39046/6?u=jamesobutler) the default location when no existing rc files exists is now the slicerHome location rather than in the user directory root.
2. Per [comment](https://discourse.slicer.org/t/cmake-flag-to-customapp-not-to-read-slicerrc-py-or-define-other-file/39046/2?u=jamesobutler) .slicerrc.py is no longer used by Slicer custom applications, but instead use their own application named rc file `.<mainApplicationName>rc.py`. 
3. Per suggestion in [comment](https://discourse.slicer.org/t/cmake-flag-to-customapp-not-to-read-slicerrc-py-or-define-other-file/39046?u=jamesobutler) Slicer custom applications can also configure to turn off application RC startup script usage entirely with a new build option `Slicer_USE_SLICERRC`. This is by default on, but can be configured to be turned off. When the app configures it off, the settings dialog entry for specifying the startup path location is hidden as well as the command line flag `ignore-slicerrc` not being added as it would be irrelevant.

### DIFF
--- a/Base/Python/slicerqt.py
+++ b/Base/Python/slicerqt.py
@@ -97,18 +97,26 @@ initLogging(logging.getLogger())
 
 def getSlicerRCFileName():
     """Return application startup file (Slicer resource script) file name.
-    If a .slicerrc.py file is found in slicer.app.slicerHome folder then that will be used.
-    If that is not found then it will search for a path defined in a SLICERRC environment variable.
-    If that environment variable is not specified or the path does not exist then .slicerrc.py in the user's home folder
-    will be used ('~/.slicerrc.py'). If the path does not exist then the default path in slicer.app.slicerHome will be used.
+    The filename convention is ``.<appname>rc.py`` where ``appname`` is
+    ``slicer.app.mainApplicationName.lower()``. For the base Slicer application
+    this resolves to ``.slicerrc.py``.
+    If such a file is found in the ``slicer.app.slicerHome`` folder then that will be used.
+    If that is not found then the path defined in the ``<APPNAME>RC`` environment variable
+    will be used (e.g. ``SLICERRC`` for the base Slicer application).
+    If that environment variable is not specified then the rc file in the user's home
+    folder will be used (e.g. ``~/.slicerrc.py``).
     """
     import os
 
-    # Ordered candidate paths
+    app_name = slicer.app.mainApplicationName
+    rc_filename = f".{app_name.lower()}rc.py"
+    env_var = f"{app_name.upper()}RC"
+
+    # Ordered candidate paths (highest to lowest priority)
     candidates = [
-        os.path.join(slicer.app.slicerHome, ".slicerrc.py"),
-        os.environ.get("SLICERRC"),
-        os.path.expanduser("~/.slicerrc.py"),
+        os.path.join(slicer.app.slicerHome, rc_filename),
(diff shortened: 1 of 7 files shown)


Base/Python/slicerqt.py
```diff
@@ -97,21 +97,36 @@ def initLogging(logger):
...
+        os.environ.get(env_var),
+        os.path.expanduser(f"~/{rc_filename}"),
+    ]
+
+    # Default to application slicerHome path; pick first existing candidate
+    rcfile = candidates[0]
+    for c in candidates:
+        if c and os.path.exists(c):
+            rcfile = c
+            break
```
ebrahimebrahim, line 127: Note that there is a slight behavior change due to now checking `os.path.exists` on _all_ candidates, which may or may not be intended:

- Previously, if the `SLICERRC` environment variable is set to a non-existent path, then that path is still preferred and so no rc file is loaded.
- Now if `SLICERRC` is set to a non-existent path, it moves on to the next candidate

(I think this behavior change could be fine but just pointing it out)

ebrahimebrahim (changes requested):
This is super useful thank you!

The test https://github.com/Slicer/Slicer/blob/main/Applications/SlicerApp/Testing/Python/SlicerOptionIgnoreSlicerRCTest.py would still be registered and probably fail when `Slicer_USE_SLICERRC=OFF`

Base/Python/slicerqt.py
jamesobutler: Yeah based on the existing path checking of the first location I took that as intention to then later look for the the environment variable and see if it referenced a path that exists before moving on to the next location. Happy to change if others know if the previous behavior was the desired intention.

jamesobutler:
> The test https://github.com/Slicer/Slicer/blob/main/Applications/SlicerApp/Testing/Python/SlicerOptionIgnoreSlicerRCTest.py would still be registered and probably fail when `Slicer_USE_SLICERRC=OFF`

@ebrahimebrahim I've updated the test with the specific application name used instead so that it will pass for custom apps. I also have the test end with success early if `--ignore-slicerrc` is not a command line argument due to building with `Slicer_USE_SLICERRC=OFF`.

ebrahimebrahim (approved):
Looks great, ty!

jcfr (approved):

jcfr:
Will definitely streamline custom app development and avoid patching the associated fork/branch.

Base/Python/slicerqt.py
lassoan: It would be nice to update the [documentation](https://github.com/Slicer/Slicer/blob/main/Docs/user_guide/settings.md) with more details on the current behavior. Maybe it could be also mentioned that the in custom applications, "slicer" is replaced by the application name in "slicerrc" or "SLICERRC" (but this information is mostly relevant to developers, so I'm not sure if the user guide is the right place for it).

PR #9244, opened 2026-06-18
Macos Qt6 packaging and arm64 building

jamesobutler:
https://projectweek.na-mic.org/PW45_2026_Boston/Projects/Qt6AndArm64SlicerBuildsOnTheFactoryMachines/
cc: @sjh26 @lassoan @pieper 

This is a collection of commits that I needed as I worked on successfully building and packaging a Qt6 based native arm64 version of 3D Slicer on macOS.

Build Environment: 
Mac Mini (M2 Pro)
macOS Tahoe 26.5.1
Xcode 26.5 Build 17F42
CMake 4.3.3
Qt 6.11.1

First, I built Qt 6.11.1 from source for arm64. I currently have my Build-Qt6.sh recipe script committed here for reference. This could also be done through an CMake external project of the superbuild since Qt6 uses CMake.
Then I built 3D Slicer from source for arm64 with the updates to pass macOS architecture to a few additional external projects where the process was getting tripped up. The LZMA process kept getting tripped up a pulling in some x86_64 reference so I ended up disabling entirely just to try and observe if the rest of the build would succeed which it did. 
**^UPDATE the LZMA issue has been addressed and restored.**

I used a Qt6 framework bundling commit from @Sunderlandkyl and further expanded upon it to resolve similar issues for Qt Designer that I observed while packaging the application. 

Ultimately with these changes I was able to build successfully, package successfully and install on a different macOS machine and run successfully. I confirmed that the app was showing as an "Apple" process instead of "Intel" using the macOS activity monitor.

xref #6811, #5944. Notably I did not need the commits from https://github.com/Slicer/Slicer/pull/9003.


Utilities/Scripts/Build-qt6.sh
```diff
@@ -0,0 +1,326 @@
...
+# Find the Python that Qt will use
+PYTHON_CMD="/opt/homebrew/Frameworks/Python.framework/Versions/3.12/bin/python3.12"
+if [ ! -f "$PYTHON_CMD" ]; then
+  PYTHON_CMD="/opt/homebrew/bin/python3"
+fi
+if [ ! -f "$PYTHON_CMD" ]; then
+  PYTHON_CMD="/usr/bin/python3"
+fi
+
+echo "Using Python: $PYTHON_CMD"
```
pieper: Use a venv here?

Utilities/Scripts/Build-qt6.sh
```diff
@@ -0,0 +1,326 @@
...
+  echo "✓ html5lib and spdx-tools installed successfully"
+else
+  echo "WARNING: html5lib installation may have failed"
+fi
+
+echo ""
+echo "All dependencies are installed."
+echo ""
+
+check_metal_toolchain() {
```
pieper, line 123: Would be good to move all dependency checks and install instructions before any brew or python installs so that the build is less like to break confusingly in the middle.

Utilities/Scripts/Build-qt6.sh
```diff
@@ -0,0 +1,326 @@
...
+  echo ""
+}
+
+# Set up environment for finding libraries
+export PKG_CONFIG_PATH="/opt/homebrew/opt/ffmpeg/lib/pkgconfig:/opt/homebrew/opt/opus/lib/pkgconfig:/opt/homebrew/opt/libvpx/lib/pkgconfig:/opt/homebrew/opt/icu4c/lib/pkgconfig:/opt/homebrew/opt/re2/lib/pkgconfig:/opt/homebrew/opt/libtiff/lib/pkgconfig:/opt/homebrew/opt/libpng/lib/pkgconfig:/opt/homebrew/opt/harfbuzz/lib/pkgconfig:/opt/homebrew/opt/glib/lib/pkgconfig:$PKG_CONFIG_PATH"
+echo "PKG_CONFIG_PATH set for Homebrew libraries"
+echo ""
+
+# Configuration
+QT_VERSION="6.11.1"
```
pieper: Would be nice to make this a command line argument.  It would be good to be able to put this script into the docs with simple command line examples to test different versions.

SuperBuild/External_python.cmake
```diff
@@ -180,8 +179,8 @@ if((NOT DEFINED PYTHON_INCLUDE_DIR
       -DINSTALL_WINDOWS_TRADITIONAL:BOOL=OFF
       -DBZIP2_INCLUDE_DIR:PATH=${BZIP2_INCLUDE_DIR}
       -DBZIP2_LIBRARIES:FILEPATH=${BZIP2_LIBRARIES}
-      -DLZMA_INCLUDE_PATH:PATH=${LZMA_INCLUDE_DIR}
-      -DLZMA_LIBRARY:FILEPATH=${LZMA_LIBRARY}
+      # -DLZMA_INCLUDE_PATH:PATH=${LZMA_INCLUDE_DIR}
+      # -DLZMA_LIBRARY:FILEPATH=${LZMA_LIBRARY}
```
pieper: Can we just drop these completely?

pieper (dismissed):
This is great progress.  I'll try it when I get a chance.

I'm fine with it going in as is, but others may want to have a look too.

SuperBuild/External_python.cmake
jamesobutler: Many of the changes here including the build Qt6 from source script is a very rough proof of concept that things can build and package. I expect a good bit of cleanup prior to integration. I will be on vacation and unavailable during next week’s project week so I wanted to post my current rough process in case any work on this front was picked up during the week.

jamesobutler:
@pieper I've updated the build script per your review. You can try it out now if you'd like. I have been using the default options which is to build Qt 6.11.1 for arm64

jamesobutler:
@Sunderlandkyl @lassoan Since you also have access to a macOS machine for building, you may want to try this out now for building and packaging a Qt 6.11.1 macOS build for arm64.

lassoan (approved):

Usually ITK vendorizes third-party libraries in a way that targets don't conflict, so it is surprising that there is a conflict. But I would not spend the time with trying to track this down, as it is nicer anyway to avoid having multiple copies of a library

PR #9255, opened 2026-06-25
BUG: Install full Python include/ tree, not just pyconfig.h

### LINKED ISSUE #9254: Embedded Python ships pyconfig.h but not the rest of the dev headers (breaks torch.compile)
I am building a Slicer extension that runs PyTorch models on the GPU, and I hit a wall trying to use torch.compile inside Slicer. After some digging it turns out the bundled Python is missing its development headers - the include/python3.12/ directory only has pyconfig.h, and Python.h (plus everything it pulls in) just isn't there.

That breaks anything that compiles a C extension against the embedded interpreter at runtime. In my case it's Triton: the first time torch.compile runs, it tries to build a little CUDA helper and dies with:

/tmp/.../cuda_utils.c:9:10: fatal error: Python.h: No such file or directory
9 | #include <Python.h>
|^~~~~~~~~~
compilation terminated.
subprocess.CalledProcessError: Command '[... gcc ... -I.../lib/Python/include/python3.12 ...]' returned non-zero exit status 1.

gcc is on the machine and the CUDA driver links fine, the only thing missing is the headers. sysconfig.get_paths()["include"] points straight at that incomplete folder.

I confirmed it's purely a headers problem: if I drop a matching CPython 3.12 Include/ tree next to the existing pyconfig.h, that exact #include <Python.h> compile succeeds and torch.compile works. But doing that by hand is a bad solution to ship to users - the headers have to match the bundled build, and they get wiped on every update, so I'd rather not have my extension reach into the Slicer install and copy files around. It really should come from the build that produces the Python you bundle.

So the ask is just: please include the full set of CPython dev headers for the bundled interpreter version under lib/Python/include/python3.12/ - the whole Include/ tree (Python.h and all the headers it includes, including the cpython/ and internal/ subfolders), alongside the pyconfig.h you already ship.
They're tiny text files. No need for a static libpython; building -shared extensions on Linux/macOS doesn't link against it, so headers alone are enough.

This isn't really nnInteractive-specific — torch.compile/Inductor/Triton is the obvious one, but the same gap probably hits Cython, pybind11, numba, building source-only wheels, anything that compiles against Python at runtime. Right now the only option is to disable torch.compile, which costs a real chunk of inference performance.

Environment, in case it helps:
- Slicer 5.11.0 (nightly, 2026-06-17), linux-amd64
- Bundled Python 3.12.10
- Ubuntu 24.04, gcc 13.3.0, NVIDIA driver present

Best,
Fabian

pieper:
Bundle all Python header files (Python.h, cpython/, internal/, etc.) alongside the previously-installed pyconfig.h so that tools that JIT-compile C extensions against the bundled interpreter — such as torch.compile, Cython, pybind11, and numba — can find the headers they need without requiring a separately installed Python.

Fixes https://github.com/Slicer/Slicer/issues/9254

### DIFF
--- a/CMake/SlicerBlockInstallDCMTKApps.cmake
+++ b/CMake/SlicerBlockInstallDCMTKApps.cmake
@@ -5,7 +5,7 @@
 set(DCMTK_Apps storescu storescp dcmdump dump2dcm img2dcm dcmdjpeg dcmqrscp dcm2xml xml2dcm dsr2html dsr2xml xml2dsr dsrdump echoscu)
 set(int_dir "")
 if(CMAKE_CONFIGURATION_TYPES)
-  set(int_dir "Release/")
+  set(int_dir "$<CONFIG>/")
 endif()
 foreach(dcmtk_App ${DCMTK_Apps})
   install(PROGRAMS ${DCMTK_DIR}/bin/${int_dir}${dcmtk_App}${CMAKE_EXECUTABLE_SUFFIX}
--- a/CMake/SlicerBlockInstallDCMTKLibs.cmake
+++ b/CMake/SlicerBlockInstallDCMTKLibs.cmake
@@ -6,7 +6,7 @@ find_package(DCMTK REQUIRED)
 foreach(dcmtk_Lib ${DCMTK_LIBRARIES})
   if(WIN32)
     if(CMAKE_CONFIGURATION_TYPES)
-      set(int_dir "Release/")
+      set(int_dir "$<CONFIG>/")
     endif()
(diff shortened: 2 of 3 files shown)


lassoan (commented):
Thank you, this looks good to me, but I've tested it on Windows and there `torch.compile` also required `python312.dll`. I've added a few lines to include that in the Python package. I'll test it a bit more and will add it to this pull request.

lassoan (approved):
I've pushed an additional commit that adds python312.lib and with that the fix works on Windows, too. It all looks good to me now.

PR #9258, opened 2026-06-26
ENH: Add an option to allow translation by clicking anywhere in the slice view

lassoan:
If Transforms module / Display / Interaction / More options / Drag from anywhere is enabled then the transform can be moved by click-and-dragging anywhere in the image. This makes is faster and easier to align objects in slice views, as the user does not need to click exactly on the center of the transform widget to translate.

image


Dcaleme:
It would be also nice to allow adjust opacity, motion sensitivity. Thank you!

lassoan:
> It would be also nice to allow adjust opacity, motion sensitivity. Thank you!

done

Sunderlandkyl (approved):
@lassoan I've taken a look, at it seems good to me!

PR #9260, opened 2026-06-29
ENH: Improve visibility of non-stable release version

jamesobutler:
> [!NOTE]
> This is targeting the `5.12` release branch

Related to discussion at https://discourse.slicer.org/t/2026-06-30-weekly-meeting/47493/4?u=jamesobutler

This begins the 5.12.1 development period on the `5.12` maintenance branch. This PR results in all commits after the 5.12.0 tagged commit being indicated like:
"5.12.1-2026-06-29 (Preview)" which indicates a preview build of a future 5.12.1 stable release
image

instead of the current:
"5.12.0-2026-06-29" which indicates a date of a stable (5.12.0) release version?
image

The current is a bit confusing as it just seeing that string when running the application it is hard to know if that version came out before the 5.12.0 release or if it came out after the 5.12.0 release. 

`ENH: Improve visibility of non-stable release version` would also be contributed to the `main` branch as well if we want to proceed with this improved display of knowing if the version of Slicer being used is a preview release or a stable release.

### DIFF
--- a/CMake/SlicerVersion.cmake
+++ b/CMake/SlicerVersion.cmake
@@ -87,7 +87,7 @@ set(Slicer_VERSION      "${Slicer_VERSION_MAJOR}.${Slicer_VERSION_MINOR}")
 set(Slicer_VERSION_FULL "${Slicer_VERSION}.${Slicer_VERSION_PATCH}")
 
 if(NOT "${Slicer_RELEASE_TYPE}" STREQUAL "Stable")
-  set(Slicer_VERSION_FULL "${Slicer_VERSION_FULL}-${Slicer_BUILDDATE}")
+  set(Slicer_VERSION_FULL "${Slicer_VERSION_FULL}-${Slicer_BUILDDATE} (${Slicer_RELEASE_TYPE})")
 endif()
 
 message(STATUS "Configuring Slicer version [${Slicer_VERSION_FULL}]")
@@ -143,7 +143,7 @@ set(Slicer_MAIN_PROJECT_VERSION      "${Slicer_MAIN_PROJECT_VERSION_MAJOR}.${Sli
 set(Slicer_MAIN_PROJECT_VERSION_FULL "${Slicer_MAIN_PROJECT_VERSION}.${Slicer_MAIN_PROJECT_VERSION_PATCH}")
 
 if(NOT "${Slicer_RELEASE_TYPE}" STREQUAL "Stable")
-  set(Slicer_MAIN_PROJECT_VERSION_FULL "${Slicer_MAIN_PROJECT_VERSION_FULL}-${Slicer_MAIN_PROJECT_BUILDDATE}")
+  set(Slicer_MAIN_PROJECT_VERSION_FULL "${Slicer_MAIN_PROJECT_VERSION_FULL}-${Slicer_MAIN_PROJECT_BUILDDATE} (${Slicer_RELEASE_TYPE})")
 endif()
 
 if(NOT "${Slicer_MAIN_PROJECT_APPLICATION_NAME}" STREQUAL "Slicer")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -90,7 +90,7 @@ if(NOT DEFINED Slicer_VERSION_MINOR)
   set(Slicer_VERSION_MINOR "12")
 endif()
 if(NOT DEFINED Slicer_VERSION_PATCH)
-  set(Slicer_VERSION_PATCH "0")
+  set(Slicer_VERSION_PATCH "1")
 endif()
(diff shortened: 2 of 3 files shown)


ebrahimebrahim (commented):
It looks good to me and this is how I'd expect things to work. But let's discuss today as we discuss the even/odd numbering for major versions, so that if we make a change we then follow the same general convention for patch versions within a release branch

Utilities/Scripts/SlicerWizard/__version__.py
```diff
@@ -1,7 +1,8 @@
 __version_info__ = (
     5,
     12,
-    0,
+    1,
+    "dev0",
```
lassoan, line 5: We should probably use the same suffix as for the application (either nothing, as we have been doing; or use `preview` or something similar).

CMakeLists.txt
```diff
@@ -90,7 +90,7 @@ if(NOT DEFINED Slicer_VERSION_MINOR)
   set(Slicer_VERSION_MINOR "12")
 endif()
 if(NOT DEFINED Slicer_VERSION_PATCH)
-  set(Slicer_VERSION_PATCH "0")
+  set(Slicer_VERSION_PATCH "1")
```
lassoan, line 93: agreed!

CMake/SlicerVersion.cmake
```diff
@@ -143,7 +143,7 @@ set(Slicer_MAIN_PROJECT_VERSION      "${Slicer_MAIN_PROJECT_VERSION_MAJOR}.${Sli
 set(Slicer_MAIN_PROJECT_VERSION_FULL "${Slicer_MAIN_PROJECT_VERSION}.${Slicer_MAIN_PROJECT_VERSION_PATCH}")
 
 if(NOT "${Slicer_RELEASE_TYPE}" STREQUAL "Stable")
-  set(Slicer_MAIN_PROJECT_VERSION_FULL "${Slicer_MAIN_PROJECT_VERSION_FULL}-${Slicer_MAIN_PROJECT_BUILDDATE}")
+  set(Slicer_MAIN_PROJECT_VERSION_FULL "${Slicer_MAIN_PROJECT_VERSION_FULL}-${Slicer_MAIN_PROJECT_BUILDDATE} (${Slicer_RELEASE_TYPE})")
```
lassoan, line 146: Same as for `Slicer_VERSION_FULL`

CMake/SlicerVersion.cmake
```diff
@@ -87,7 +87,7 @@ set(Slicer_VERSION      "${Slicer_VERSION_MAJOR}.${Slicer_VERSION_MINOR}")
 set(Slicer_VERSION_FULL "${Slicer_VERSION}.${Slicer_VERSION_PATCH}")
 
 if(NOT "${Slicer_RELEASE_TYPE}" STREQUAL "Stable")
-  set(Slicer_VERSION_FULL "${Slicer_VERSION_FULL}-${Slicer_BUILDDATE}")
+  set(Slicer_VERSION_FULL "${Slicer_VERSION_FULL}-${Slicer_BUILDDATE} (${Slicer_RELEASE_TYPE})")
```
lassoan, line 90: If this is meant to be a version identifier used in code then it has to be modified a bit, as we cannot have space and parentheses in it (it would not be a valid "semantic version").

If it is meant to be displayed for users (in installer, Start menu, application window title, etc.) then we cannot hardcode an English word to it, but it should be translatable. It should also be conveniently configurable for Slicer custom applications.

lassoan (commented):
We should probably integrate the patch version increment separately while we decide on the version suffix.

Utilities/Scripts/SlicerWizard/__version__.py
jamesobutler: We have been making this type of change in SlicerWizard for awhile now. See https://github.com/Slicer/Slicer/commit/d3d3bc46c38ba604f76a8ae1a54cf0d7ef7543c4.

lassoan: I don't know why we do that. There is also a https://github.com/Slicer/Slicer/blob/main/Utilities/Scripts/SlicerWizard/__version__.py.in that avoids the hardcoding completely. The `SlicerWizard` also has its own `Slicer_VERSION_TWEAK_OR_DEV` CMake variable.

Probably the best would be to get rid of this duplicate hardcoded versioning and derive the version from the same place as Slicer (currently the top-level CMakeLists.txt; but maybe it could come from a more easily accessible place, such as some kind of manifest file, which would not require CMake code interpretation, just a json parser).

jamesobutler:
Will reconsider this work again from a fresh start in the future.

PR #9264, opened 2026-07-01
ENH: Add support for building against VTK version 9.7.0

jamesobutler:
This adds initial support for building against VTK 9.7.0. This currently builds the OpenGL rendering backend, but with VTK 9.7 we can then later add the build option to enable the WebGPU rendering backend.

https://discourse.vtk.org/t/vtk-9-7-0-and-10-0-0-release-planning/16371/3

VTK 9.7.0 Release Notes:
https://github.com/Kitware/VTK/blob/master/Documentation/release/9.7.md

Windows Test Results (with Qt 5.15.2 and VTK v9.7.0):
- Build ✔️ 
- Package ✔️  
- Tests ✔️


jamesobutler:
I've rebased this branch against latest Slicer and updated to test against latest VTK 9.7.0.rc2.

jamesobutler:
Issues found by the automated testing has been fixed up. There is a last pending commit (https://github.com/Slicer/Slicer/pull/9264/changes/cbf4531eb4d5ca34d36421cbad573da04db29d5c) with some fixups for some remaining testing. I haven't yet found the corresponding VTK commits leading to the various changes in behavior. I also haven't verified if those changes are specific to VTK 9.7 or would also be compatible with the still supported VTK 9.5 and 9.6 versions.

Sanity testing on other platforms such as Linux would probably be a good next step now.  cc: @ebrahimebrahim

Libs/MRML/Core/vtkMRMLScene.cxx
```diff
@@ -1814,11 +1836,10 @@ int vtkMRMLScene::GetNumberOfNodesByClass(const char* className)
     return 0;
   }
   int num = 0;
-  vtkMRMLNode* node;
-  vtkCollectionSimpleIterator it;
-  for (this->Nodes->InitTraversal(it); (node = (vtkMRMLNode*)this->Nodes->GetNextItemAsObject(it));)
+  for (int i = 0; i < this->Nodes->GetNumberOfItems(); ++i)
   {
-    if (node->IsA(className))
+    auto* node = (vtkMRMLNode*)this->Nodes->GetItemAsObject(i);
```
ebrahimebrahim: It seems vtkCollection went from being linked list backed to being std::vector backed in the transtion to 9.7. It was still linked list in 9.6:

https://github.com/Kitware/VTK/blob/v9.6.2/Common/Core/vtkCollection.h#L28-L39

Since we are retaining 9.5 and 9.6 support in this PR, vtkCollection being a linked list becomes an issue here: the transition from the existing iterator implementation to `vtkCollection::GetItemAsObject(i)` makes several common scene operations go from linear to quadratic.

Libs/MRML/DisplayableManager/vtkMRMLThreeDViewInteractorStyle.cxx
```diff
@@ -192,11 +192,13 @@ bool vtkMRMLThreeDViewInteractorStyle::QuickPick(int x, int y, double pickPoint[
...
-    vtkProp* aProp = nullptr;
-    for (props->InitTraversal(pit); (aProp = props->GetNextProp(pit));)
+    for (int i = 0; i < props->GetNumberOfItems(); ++i)
     {
-      aProp->GetVolumes(pickList);
+      vtkProp* aProp = vtkProp::SafeDownCast(props->GetItemAsObject(i));
+      if (aProp)
+      {
+        aProp->GetVolumes(pickList);
+      }
```
ebrahimebrahim: Similarly to my other comment, for VTK 9.6 which we are still supporting in this PR the `GetItemAsObject` lookup in the loop makes this quadratic rather than linear in the view props collection. 

The code comment saying this runs each time the mouse moves over a 3D view makes it possibly more of an issue

Libs/vtkITK/vtkITKImageToImageFilter.h
```diff
@@ -143,6 +143,20 @@ class VTK_ITK_EXPORT vtkITKImageToImageFilter : public vtkImageAlgorithm
...
+    this->vtkImporter->Update();
+    return true;
+  }
+  bool Update(int port) override
+  {
+    this->vtkCast->Update();
+    this->vtkImporter->Update(port);
+    return true;
+  }
+#else
```
ebrahimebrahim, line 159: In VTK 9.7, `vtkAlgorithm::Update` returns `bool ` indicating success

https://github.com/Slicer/VTK/blob/a675c60fc84f2b7063c89b1d6d9822555e493b38/Common/ExecutionModel/vtkExecutive.h#L74-L81

https://github.com/Slicer/VTK/blob/a675c60fc84f2b7063c89b1d6d9822555e493b38/Common/ExecutionModel/vtkAlgorithm.cxx#L1647-L1660

It is supposed to propagate the success flag down the pipeline.

Here in this override the signature is updated correctly, however we stop propagating the  success signal. i.e. we are throwing away the result of Update()

Libs/vtkITK/vtkITKImageThresholdCalculator.cxx
```diff
@@ -129,7 +129,21 @@ void vtkITKImageThresholdCalculator::PrintSelf(ostream& os, vtkIndent indent)
...
+{
+  this->UpdateImpl();
+  return true;
+}
+#else
 void vtkITKImageThresholdCalculator::Update()
+{
+  this->UpdateImpl();
+}
+#endif
```
ebrahimebrahim, line 143: Kind of similar to my other comment, this is failing to pass along and return the `bool` flag that is returned by `UpdateImpl`

jamesobutler:
I've rebased against the latest VTK 9.7.0.rc3 release candidate for new testing.

ebrahimebrahim (commented):
Linux, VTK 9.6:  

Build: ✔️ 

Linux, VTK 9.7 (rc3):

Build: ✔️ 
Test: ✔️ 
Package: ✔️

Libs/MRML/Core/vtkMRMLScene.cxx
ebrahimebrahim: I see that just now you changed the default minor version to vtk 9.7, so that could be a reason to be okay with this problem. IDK how serious we are when we say "we still support building against vtk 9.6". Or how consequential it would be to just stop supporting 9.6 so that we don't have to keep sticking `#if VTK_VERSION_NUMBER >= VTK_VERSION_CHECK(9, 7, 0)`. This is just my ignorance of the history of slicer and vtk and how customizable developers expect slicer builds to be

jamesobutler: Yeah I think historically we've supported at least 2 versions so that the custom apps and others could still continue to work uninterrupted with latest Slicer until they themselves switch to the newer VTK. As it relates to the specific code change to handle this new deprecation it would be annoying with the version checks for this. In the case of ITK we only support 1 latest version to choose from, but ITK hasn't been changing as much like VTK and its rolling deprecation type policy.

jamesobutler:
FYI VTK 9.7.0 is planned for release on this upcoming Monday August 3rd in case anyone sees any changes that appear to be VTK side related issues that should be fixed before their stable release. See https://discourse.vtk.org/t/vtk-9-7-0-and-10-0-0-release-planning/16371/5 for the announcement.

jamesobutler:
I've picked over the VTK commit from @lassoan  (https://github.com/Slicer/VTK/commit/87a500f5baa3701fa17835760124087aa00b3a1f) over into the v9.7.0.rc3 based branch (https://github.com/Slicer/VTK/commit/06cc281f46bc31289b72fc030b83737ea75bc068) for testing in the context of VTK 9.7 as well. @lassoan Please confirm the cherry-picked commit as there was some merge conflicts I had to handle.

jamesobutler:
I've rebased against the latest VTK 9.7.0.rc4 release candidate for new testing.

ebrahimebrahim:
Thank you for rebasing, is this ready for testing and integration?

jamesobutler:
I had picked off a couple of the last commits https://github.com/jamesobutler/Slicer/commit/ed8d60bba4784bda96b67ec2019b603f53ccb1da and https://github.com/jamesobutler/Slicer/commit/22f91030a4e838127b8ffb9dfa242bd579390ce7 and was running tests locally.

I observed on my Windows build that I saw some weird issues with the layout manager with my most recent build. The yellow slice view of the four-up layout and other views in some other layout were showing as completely white. This is probably why I was seeing some vtk debug leaks on close of some of the automated tests and failures related to layout widgets in several. Unclear whether that is VTK 9.7 initiated or a messed up attempt of rebasing recent Slicer fork branch commit on top of the new VTK 9.7 code (https://github.com/Slicer/VTK/commit/98b6a9bfcfbeeca7eb5d126290621057150a8fa5). Debugging help from others would be appreciated here to narrow things down here.

ferdymercury:
Side note, maybe it could be useful to make a two-step approach? Getting first https://github.com/Slicer/Slicer/pull/9293 running and then this one? To better bisect at what point it starts failing

jamesobutler:
I've rebased against latest Slicer and updated the https://github.com/Slicer/VTK/commits/slicer-v9.7.0-2026-08-14-23f0a0956 branch with several of the recent commits that were added to https://github.com/Slicer/VTK/commits/slicer-v9.6.2-2026-05-15-f49a1dbaf/ since this PR was last updated. 

Work can continue as part of debugging the oddities around the layouts as described in https://github.com/Slicer/Slicer/pull/9264#issuecomment-5361815972.
image

jamesobutler:
> I had picked off a couple of the last commits https://github.com/jamesobutler/Slicer/commit/ed8d60bba4784bda96b67ec2019b603f53ccb1da and https://github.com/jamesobutler/Slicer/commit/22f91030a4e838127b8ffb9dfa242bd579390ce7 and was running tests locally.

It appears the `COMP: Replace deprecated vtkCollectionIterator usage` is what resolved the weird layout issue. That's why I began to see the issue whenever I picked it off from the branch. I originally had pulled it out because of the pending considerations about how to the use the new VTK 9.7 compatible code for this change specifically while also maintaining the existing compatibility for VTK 9.6 and 9.5. Recently we've maintained compatibility for multiple VTK versions to help with transition periods, but maybe with Slicer 6 being the next release that is less necessary because there is likely to be other breaking changes as people migrate from a Slicer 5->Slicer 6 version. Thoughts @ebrahimebrahim (related to previous conversation at https://github.com/Slicer/Slicer/pull/9264#discussion_r3657978139)?

ebrahimebrahim:
I agree that for the transition into Slicer 6, it is not unreasonable to drop compatibility with VTK 9.6. Let me ask around Kitware and see if anyone else has strong opinions, and I'll post more here by tomorrow or so.

ebrahimebrahim:
Outcome: General agreement, no objects, let's go ahead and drop VTK 9.6 support for Slicer 6!

Libs/vtkITK/vtkITKImageThresholdCalculator.cxx
```diff
@@ -152,21 +158,22 @@ void vtkITKImageThresholdCalculator::Update()
...
   }
 
   int inputDataType = pointData->GetScalars()->GetDataType();
   switch (inputDataType)
   {
     vtkTemplateMacro(ITKComputeThresholdFromVTKImage<VTK_TT>(this, inputImage, this->Threshold));
-    default: vtkErrorMacro("Execute: Unknown ScalarType" << inputDataType); return;
+    default: vtkErrorMacro("Execute: Unknown ScalarType" << inputDataType); return false;
   }
+  return true;
```
ebrahimebrahim, line 176: Related to https://github.com/Slicer/Slicer/pull/9264#discussion_r3657462564 above, which was fixed, here we should also return false if the threshold calculation fails. `ITKComputeThresholdFromVTKImage` has a `catch` block above which logs errors, but it does not pass the failure to here.

ebrahimebrahim (changes requested):
(This is an AI assisted review, but the text below is written manually)

**(1)** In our VTK fork, the https://github.com/Slicer/VTK/blob/15f4a0b4db4f387b7fc4e9832b52341d286f7e2b/Wrapping/Python/vtkmodules/util/data_model.py#L55 lazy import change removed `numpy_support`. But it is still called by data_model.py, in the parts of data_model.py that for example enable convenience numpy style assignment. So numpy style assignment for example is now broken; the following snippet now gives an error:

```python
import numpy as np
import vtk

image = vtk.vtkImageData()
image.SetDimensions(2, 2, 1)
image.point_data["values"] = np.arange(4, dtype=np.float32)
```

**(2)** The AI reviewer found that our [OpenXR](https://github.com/Slicer/VTK/tree/15f4a0b4db4f387b7fc4e9832b52341d286f7e2b/Rendering/OpenXR) does not compile. I did not verify, but it is worth a look. If it is an issue then we'd run into it if we build Slicer with `VTK::RenderingOpenXR` enabled.

**(3)** `VTKObservationMixin` uses observed objects as dictionary keys, and so they need to be hashable. But now with VTK 9.7, array wrappers are [no longer hashable](https://github.com/Slicer/VTK/blob/15f4a0b4db4f387b7fc4e9832b52341d286f7e2b/Wrapping/Python/vtkmodules/numpy_interface/_vtk_array_mixin.py#L92-L93). Here is a snippet that now fails as a result:

```python
import vtk
from slicer.util import VTKObservationMixin

array = vtk.vtkFloatArray()
observations = VTKObservationMixin()
observations.addObserver(
    array,
    vtk.vtkCommand.ModifiedEvent,
    lambda caller, event: print("Array modified"),
)
array.Modified()
observations.removeObservers()
```


- (1) should be fixed.
- I am okay with fixing (2) in follow-up work, or if you find it to be a non-issue then ignore me on it.
- I worry that (3) might break some extensions. The fix could be done in a separate, easier-to-review PR to get it through more quickly. It might need some restructuring of the internals of `VTKObservationMixin`.

PR #9266, opened 2026-07-05
ENH: Fix macOS bundle verification and add 'packageverify' target

lassoan:
SlicerCPackBundleVerify.cmake was broken: it was never wired into the build and contained a syntax error. By relying on BundleUtilities' verify_app() it also both missed real packaging errors and raised false alarms:
 - missed errors: verify_app() treats unresolved @rpath references as warnings, so a bundle missing entire frameworks (e.g. Qt not embedded) still reported success.
 - false alarms: it emitted "cannot resolve" warnings for @rpath items even on a correct bundle.

The script was not used in the build process, but it is a useful feature, therefore instead of removing it, I got it rewritten by Claude to perform a strict, dyld-accurate self-containment check, and add a 'packageverify' convenience target that runs it against the bundle staged by CPack.

For every Mach-O file in the bundle the script parses "otool -l" and:
 - FATAL: reports frameworks with a missing binary (empty .framework skeleton) and dependent libraries that do not resolve to a file inside the bundle. @rpath is expanded against each binary's own LC_RPATH plus the main executable's (dyld builds its run-path list from the whole load chain), so plugins that carry no rpath of their own are handled without false positives.
 - WARNING: reports non-relocatable LC_RPATH entries (build-tree leaks or foreign paths baked into bundled Python wheels; harmless because dyld ignores paths that do not exist), optional external backends under /usr/local or /opt (e.g. the Qt SQL drivers), and unresolved weak dependencies.

The script verifies the already-built package, so run 'make package' first:

  make package
  make packageverify

Example output on a bundle where Qt failed to be embedded:

```
kyle@rootmacs-Mac-mini Slicer-build % make packageverify Verifying packaged application bundle is self-contained -- Verifying bundle is self-contained: /Users/kyle/d/Slicer-Release/Slicer-build/_CPack_Packages/macosx-amd64/DragNDrop/Slicer-5.13.0-2026-07-04-macosx-amd64/3D Slicer 5.13.0-2026-07-04.app

ERROR: 27 framework(s) do not contain their binary (empty '.framework' skeleton -- the fixup did not embed the library):
  - QtCore.framework
  - QtCore5Compat.framework
  - QtDesigner.framework ...
  - QtWebEngineWidgets.framework
  - QtWidgets.framework
  - QtXml.framework

ERROR: 30 dependency(ies) do not resolve inside the bundle:
  - [MISSING] @rpath/Frameworks/QtWidgets.framework/Versions/A/QtWidgets referenced by 253 binary(ies), e.g. Contents/bin/Python/vtkmodules/vtkRenderingQt.cpython-312-darwin.so
  - [MISSING] @rpath/Frameworks/QtGui.framework/Versions/A/QtGui referenced by 272 binary(ies), e.g. Contents/lib/QtPlugins/sqldrivers/libqsqlpsql.dylib ...
  - [MISSING] @rpath/QtNetwork.framework/Versions/A/QtNetwork referenced by 6 binary(ies), e.g. Contents/lib/QtPlugins/tls/libqcertonlybackend.dylib
  - [MISSING] @rpath/QtCore.framework/Versions/A/QtCore referenced by 6 binary(ies), e.g. Contents/lib/QtPlugins/tls/libqcertonlybackend.dylib

WARNING: 8 non-relocatable LC_RPATH entry(ies) (absolute paths leaking the build/packaging machine or bundled Python wheels; non-fatal -- dyld ignores paths that do not exist):
  - /Users/kyle/Qt6/6.9.3/macos/lib present in 246 binary(ies), e.g. Contents/bin/Python/vtkmodules/vtkRenderingQt.cpython-312-darwin.so
  - /Users/runner/work/Pillow/Pillow/build/deps/darwin/lib present in 1 binary(ies), e.g. Contents/lib/Python/lib/python3.12/site-packages/PIL/.dylibs/libjpeg.62.4.0.dylib ...
  - /Users/kyle/d/Slicer-Release/teem-build/bin present in 269 binary(ies), e.g. Contents/lib/QtPlugins/designer/libqMRMLWidgetsPlugins.so
  - /Users/kyle/d/Slicer-Release/qRestAPI-build present in 2 binary(ies), e.g. Contents/lib/Slicer-5.13/libqSlicerBaseQTCore.dylib

WARNING: 2 dependency(ies) on optional external libraries (/usr/local, /opt) that are not bundled by design (non-fatal; disables only the corresponding plugin, e.g. a Qt SQL driver):
  - /usr/local/lib/libmimerapi.dylib
  - /usr/local/opt/libiodbc/lib/libiodbc.2.dylib

CMake Error at /Users/kyle/d/Slicer/CMake/SlicerCPackBundleVerify.cmake:504 (message):
  Bundle verification FAILED with 57 error(s): the application bundle is not
  self-contained.  See the report above.

make[3]: *** [CMake/LastConfigureStep/CMakeFiles/packageverify] Error 1
make[2]: *** [CMake/LastConfigureStep/CMakeFiles/packageverify.dir/all] Error 2
make[1]: *** [CMake/LastConfigureStep/CMakeFiles/packageverify.dir/rule] Error 2
make: *** [packageverify] Error 2
```

### DIFF
--- a/CMake/SlicerCPack.cmake
+++ b/CMake/SlicerCPack.cmake
@@ -552,3 +552,33 @@ DeleteRegKey SHCTX \\\"SOFTWARE\\\\Classes\\\\${ext}\\\"
 endif()
 
 include(CPack)
+
+# -------------------------------------------------------------------------
+# macOS: 'packageverify' convenience target
+# -------------------------------------------------------------------------
+# Add a 'packageverify' target that checks whether the already-built application
+# bundle is self-contained by running SlicerCPackBundleVerify.cmake on the bundle
+# staged by CPack. It does NOT (re)build the package; run 'make package' first.
+# This avoids having to locate the staged bundle and pass the required variable
+# by hand.
+if(APPLE)
+
+  # Directory where the CPack DragNDrop generator stages the packaged bundle:
+  #   <build>/_CPack_Packages/<system>/DragNDrop/<package-file-name>/<app>.app
+  # The <system> component is built directly from Slicer_OS/Slicer_ARCHITECTURE
+  # rather than from CPACK_SYSTEM_NAME, since including CPack repurposes the
+  # CPACK_SYSTEM_NAME and CPACK_PACKAGE_FILE_NAME variables for the source
+  # package. The <package-file-name> and bundle name embed a build date and are
+  # discovered by the verify script, so only the DragNDrop directory is passed.
+  set(_packageverify_dir
+    "${CMAKE_BINARY_DIR}/_CPack_Packages/${Slicer_OS}-${Slicer_ARCHITECTURE}/DragNDrop")
+
+  add_custom_target(packageverify
+    COMMAND ${CMAKE_COMMAND}
+      "-DSlicer_INSTALL_DIR=${_packageverify_dir}"
+      -P "${Slicer_CMAKE_DIR}/SlicerCPackBundleVerify.cmake"
+    COMMENT "Verifying packaged application bundle is self-contained"
+    VERBATIM
+    )
+
+endif()
--- a/CMake/SlicerCPackBundleVerify.cmake
+++ b/CMake/SlicerCPackBundleVerify.cmake
@@ -1,17 +1,509 @@
+#-----------------------------------------------------------------------------
+# SlicerCPackBundleVerify
+#-----------------------------------------------------------------------------
+#
+# Standalone helper script that checks whether a macOS application bundle is
+# self-contained, i.e. that every embedded Mach-O binary can be loaded on a
+# machine that only has the bundle (plus the OS), with no dependency on the build
+# tree or on libraries installed on the packaging machine.
+#
+# It is the validation counterpart to SlicerCPackBundleFixup.cmake.in: the fixup
+# script embeds dependencies and rewrites install names / rpaths, and this script
+# verifies the result.
+#
+# What it checks
+# --------------
+# For every Mach-O file in the bundle (executables, *.dylib, *.so and framework
+# binaries) it parses the load commands ("otool -l") and reports:
+#
+#   FATAL (fail the verification):
+#   1. Frameworks whose actual binary is missing (an empty ".framework"
+#      skeleton), which happens when the fixup fails to copy the library.
+#   2. Dependent libraries (LC_LOAD_DYLIB / LC_REEXPORT_DYLIB /
+#      LC_LOAD_UPWARD_DYLIB) that cannot be resolved, or that resolve to a file
+#      *outside* the bundle. Resolution follows the same rules as dyld: @rpath is
+#      expanded against the binary's own LC_RPATH entries as well as the main
+#      executable's (dyld builds its run-path list from the whole load chain,
+#      which always starts at the main executable, so plugins that carry no rpath
+#      of their own still load correctly), and @loader_path / @executable_path are
+#      expanded relative to the binary / main executable. Absolute system
+#      libraries (/usr/lib, /System) are allowed.
+#
+#   WARNING (reported, but do not fail):
+#   3. Non-relocatable LC_RPATH entries, i.e. absolute paths that are not system
+#      locations (build-tree leaks such as ".../CTK-build/.../lib" or a
+#      developer's Qt directory, and foreign paths baked into bundled third-party
+#      Python wheels). A stale rpath is harmless on its own -- dyld skips paths
+#      that do not exist -- so this is a hygiene signal, not a loadability error;
+#      genuinely unsatisfied dependencies are caught by check 2 instead.
+#   4. Dependencies on optional external backends under /usr/local or /opt (e.g.
+#      the ODBC / PostgreSQL / Mimer Qt SQL drivers), and unresolved weak
+#      dependencies. These are not bundled by design and only disable the
+#      corresponding optional plugin.
+#
+# This intentionally does NOT use BundleUtilities' verify_app(): that routine
+# treats unresolved @rpath references as non-fatal warnings and therefore passes
+# bundles that are missing entire frameworks.
+#
+# When to run it
+# --------------
+# Run it *after packaging*, on the fixed-up bundle. The bundle only becomes
+# self-contained once the CPack fixup script has run (as part of "make install"
+# or "make package" / cpack). Do NOT run it on the freshly built application in
+# the build tree: at that point the bundle still links against the build tree, so
+# verification is expected to fail.
+#
+# How to run it
+# -------------
+# The recommended way is the 'packageverify' convenience target (added by
+# SlicerCPack.cmake), which runs this script against the staged bundle with the
+# required variable filled in automatically. It verifies the already-built
+# package, so run 'make package' first:
+#
+#   make package
+#   make packageverify        # or: cmake --build <build> --target packageverify
+#
+# Alternatively, invoke it directly in script mode, providing the directory that
+# contains the bundle:
+#
+#   cmake \
+#     -DSlicer_INSTALL_DIR=/path/to/dir-containing-the-app \
+#     -P /path/to/Slicer/CMake/SlicerCPackBundleVerify.cmake
+#
+# Slicer_INSTALL_DIR may either directly contain the ".app" bundle or contain the
+# CPack staging subdirectory that holds it (e.g. the DragNDrop directory
+# "<build>/_CPack_Packages/<system>/DragNDrop/"). The bundle itself is located by
+# globbing, so its exact (date-dependent) name does not need to be known.
+#
+# On success the script reports the number of binaries verified; on failure it
+# lists every problem found and aborts with a fatal error.
+#-----------------------------------------------------------------------------
 
-include(BundleUtilities)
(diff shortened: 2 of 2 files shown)


jamesobutler:
Were you able to use this work here to validate the packaging fixes in #9244 were correct?

lassoan:
I only tested so far with my broken Qt5 package. It would be awesome if you could test with Qt6. @Sunderlandkyl has successfully created an installable Qt6 based package, so it should be possible to make it work.

pieper:
+1 for updating all of this.  I have no easy way to test it, but I think it would be better to have a version that probably works over one we know does not work.

hjmjohnson:
Tested this branch on an independent machine (macOS arm64, Qt 6.9.1, Ninja/Release superbuild): `packageverify` works as described and caught a genuinely broken package that `make package` had reported as a success (exit 0, DMG generated).

<details>
<summary>Test details</summary>

Setup: existing Qt 6.9.1 (arm64) superbuild tree, inner Slicer-build reconfigured on this branch, then `ninja package` followed by `ninja packageverify`.

- `ninja package` exited 0 and produced the DMG despite `fixup_bundle` failing to embed all Qt frameworks (the fixup log contained "Error copying file @rpath/QtWebEngineWidgets..." messages that did not fail the build) — exactly the silent-failure mode this PR targets.
- `ninja packageverify` failed with 321 errors:
  - 27 empty `.framework` skeletons (all Qt 6 frameworks), matching the example in the commit message.
  - 294 unresolved dependencies: missing `@rpath` Qt loads plus `[OUTSIDE]` references into build trees (DCMTK-build, VTK-build, python-install, teem-build).
  - Non-fatal warnings correctly classified: 11 non-relocatable `LC_RPATH` leaks (including `/Users/<user>/Qt/6.9.1/macos/lib`) and 3 optional `/usr/local`/`/opt` externals (`libmimerapi`, `libiodbc`, Homebrew openjpeg — the Qt SQL driver backends).

Spot-checked two findings by hand — no false positives:
- `Contents/Frameworks/QtCore.framework` in the staged bundle contains zero files (empty skeleton, as reported).
- `otool -L` on `CTKCorePythonQt.so` confirms an absolute build-tree `libpython3.12.dylib` reference and unresolvable `@rpath` QtCore/QtCore5Compat loads.

</details>

hjmjohnson:
Follow-up to my [test report](https://github.com/Slicer/Slicer/pull/9266#issuecomment-4891835729): I also did a code-review of the verification logic. The approach is right and a big improvement over the never-wired-in `verify_app()` script, but the review found five issues to consider fixing before merge (each is small), plus a handful of non-blocking observations.

**Suggested before merge:**

- [x] **CLI module executables are never scanned.** The extensionless-executable scan covers only `Contents/MacOS`, `Contents/bin`, `Contents/libexec` — Slicer's CLI executables live in `Contents/lib/Slicer-X.Y/cli-modules/` (and Python.framework ships `bin/python3`). A fixup regression there would pass verification silently. The old `get_bundle_all_executables()` scanned bundle-wide.
- [x] **`file(READ)` on a dangling symlink hard-aborts the script.** `_bv_is_macho()` has no `EXISTS` guard, and broken framework skeletons — the exact defect this tool diagnoses — commonly contain symlinks to a binary the fixup failed to copy. Reproduced: `cmake -P` dies with a raw "file failed to open for reading" instead of the intended report. An `IS_SYMLINK`/`EXISTS` check fixes it.
- [x] **First-match `@rpath` resolution through leaked build-tree rpaths causes false FATALs on the packaging machine.** A binary with `LC_RPATH [/Users/x/.../CTK-build/bin, @loader_path/../Frameworks]` and a correctly-bundled dep resolves to the build tree first and is reported `[OUTSIDE]` — even though dyld on a user machine would skip the nonexistent path and load the bundled copy (as the script's own "stale rpaths are harmless" warning text says). Developers may never get a clean pass on the machine that built the package. Suggest: when an absolute rpath candidate resolves outside the bundle, keep searching; report `[OUTSIDE]` only if no in-bundle resolution exists.
- [x] **Two staged bundles → FATAL_ERROR.** The `*.app` glob hard-fails on more than one match, which is the normal state after running `package` on two different days (date-stamped names accumulate under `_CPack_Packages`). Picking the newest (or the one matching the most recent DMG) would avoid making everyone clean the staging dir first.
- [x] `/usr/local/` and `/opt/` deps are always warning-only, which includes found Homebrew libraries (e.g. `/opt/homebrew/lib/libomp.dylib`) — a bundle that hard-depends on the packaging machine's Homebrew would PASS. The old `verify_app()` treated these as fatal external prerequisites. Might be worth narrowing the allowlist to the known optional Qt SQL/ODBC backends.

<details>
<summary>Non-blocking observations (fine as follow-ups)</summary>

- `otool` failures are swallowed (`ERROR_QUIET`, and otool exits 0 on non-Mach-O input), so a corrupt embedded binary counts as "verified, zero deps." Checking that the output contains at least one load command (or dropping `ERROR_QUIET`) would close it.
- `global_rpath_dirs` merges `LC_RPATH` from every Mach-O in `Contents/MacOS`, so a helper executable's rpath can satisfy a resolution that fails in the real app's load chain; there's also no check that a main executable exists at all.
- Bare install names with no `/` or `@` prefix (e.g. `libz.1.dylib`, common from older autotools builds) fall through every resolver branch and are always fatal `[MISSING]`; probably worth an explicit classification.
- dyld builds its run-path stack from all intermediate loaders, not just the dependent binary + main executable; a plugin loaded via an intermediate library whose rpath does the resolving could be a false positive. Edge case; the current approximation is reasonable.

</details>

lassoan:
Thank you for testing @hjmjohnson. It would be nice if you could implement these suggestions, if it is not too much trouble (using Claude). We could do it later, too, as it is an optional developer tool that is already more useful with the current changes than it was before.

hjmjohnson:
Implemented all 5 checklist items in 7a6304384f (pushed to this branch, thanks for `maintainerCanModify`) and re-ran the script against the same Qt 6.9.1 staged bundle.

<details>
<summary>What changed and re-test results</summary>

- Extensionless-executable scan is now bundle-wide (`file(GLOB_RECURSE "${app_real}/*")` + magic-number check) — CLI modules under `lib/Slicer-X.Y/cli-modules/` and framework `bin/` dirs are covered. Full-bundle scan adds ~30 s (54 s total on this bundle).
- `_bv_is_macho()` rejects dangling symlinks (`NOT EXISTS` guard) — required for the bundle-wide scan, since empty framework skeletons contain dangling `Versions/Current` symlinks.
- `@rpath` resolution now collects all candidate dirs (own `LC_RPATH` + main executable's) and prefers an in-bundle match; a leaked build-tree rpath no longer produces a false `[OUTSIDE]` on the packaging machine.
- `EXTERNAL` (warning-only) is narrowed to the known optional DB backends (`libiodbc|libodbc|libpq|libmimerapi|libmysqlclient|libmariadb`); everything else under `/usr/local`//`/opt` is fatal. On my bundle this correctly promoted a Homebrew `libopenjp2.7.dylib` dependency from warning to `[OUTSIDE]` error.
- Multiple date-stamped staged bundles: the newest is selected (with a STATUS note) instead of `FATAL_ERROR`.

Re-test on the known-broken bundle: 322 errors (was 321 + the promoted openjpeg dep), no crash, warnings correctly reduced to the two DB drivers.

</details>

hjmjohnson (approved):
Requested polishing changes made.

hjmjohnson:
Independently verified the rebased branch (f18e18e1d0) on a second machine (macOS arm64, Qt 6.9.1): the Qt6 embedding fix works — `packageverify` reports **zero empty framework skeletons** (was 27); all Qt frameworks embed with their binaries. This resolves #9279 (filed before I saw the rebase; the branch now fixes both root causes it describes, so it can close on merge).

<details>
<summary>Verification details</summary>

- Full cycle from clean staging: reconfigure → `ninja package` (exit 0) → `ninja packageverify`.
- "Error copying file" occurrences in the packaging log: 2 (down from dozens) — both pre-existing and unrelated to Qt: a bare `libPythonQt.dylib` install name, and a machine-local `/Applications/Postgres.app/.../libpq.5.dylib` reference in Qt's psql driver.
- `packageverify` still fails with 268 unresolved-dependency errors — all pre-existing packaging debt (absolute build-tree references in DCMTK/VTK-linked binaries, bare CTK/PythonQt install names, Qt tls-plugin rpaths), not Qt embedding. The verifier is doing its job; these are candidates for follow-up issues.
- Note: I had implemented an equivalent fix in parallel before seeing the rebase; the branch's version is equal or better on every point (more general `qt_root_dir` regex, per-framework `Versions/*` search dirs, `-delete_rpath` + `-add_rpath` for Designer), so I dropped mine in favor of the branch as-is.

</details>

w1ebr:
Does this mean that there is an Apple silicon native build now?

lassoan:
Thanks for the improvements @hjmjohnson. It looks good and I'll merge it now.

lassoan:
> Does this mean that there is an Apple silicon native build now?

Yes. There are still a few small bugs, but it is usable. Apple Silicon native Slicer Preview Release builds are coming soon.

PR #9268, opened 2026-07-06
ENH: Name macOS application bundle with the full application name

lassoan:
Previously the macOS package always contained "Slicer.app", so after dragging it to the Applications folder it appeared simply as "Slicer" in Finder and had to be renamed manually to tell versions apart.

As a final packaging step, rename the bundle to the full application name (display name and version, e.g. "3D Slicer 5.10.0"), matching the naming already used by the Windows installer. The executable and all embedded paths keep the internal application name, so only the top-level bundle directory is renamed and the bundle stays relocatable.

For preview releases the version carries a build-date suffix (e.g. "5.13.0-2026-07-04"). This results in a long .app file name, which
Finder elides in the middle, hiding the version number:

image

To avoid this, the implementation in this pull request replaces the separator before the date by a space ("5.13.0-2026-07-04" -> "5.13.0 2026-07-04"):

image

We might consider backporting this to Slicer-5.12.1.

### DIFF
--- a/CMake/SlicerCPackBundleFixup.cmake.in
+++ b/CMake/SlicerCPackBundleFixup.cmake.in
@@ -437,3 +437,28 @@ endfunction()
 
 set(app_name "@Slicer_MAIN_PROJECT_APPLICATION_NAME@.app")
 fixup_bundle_with_plugins(${app_name})
+
+#-----------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


pieper (approved):
Thanks for doing this! 👍 

I'll be happy to test this on the next preview.

lassoan:
OK, let's try this then.

PR #9273, opened 2026-07-07
BUG: Disable pip install UI off main thread

### LINKED ISSUE #9256: Crash running nnInteractive extension
## Summary

Pip install dialog crashes during module setup.

## Steps to reproduce

* Install nnInteractive extension
* restart slicer
* enter module 
* dialog flashes and then crash

## Environment
- Slicer version: Slicer-5.11.0-2026-06-15
- Operating system: Mac + 26.5.1 (25F80)

ebrahimebrahim:
Closes #9256

See also
- https://discourse.slicer.org/t/nninteractive-slicer-client-is-stuck/47550/9
- https://discourse.slicer.org/t/2026-07-07-weekly-meeting/47523/9
- https://github.com/coendevente/SlicerNNInteractive/pull/100

While https://github.com/coendevente/SlicerNNInteractive/pull/100 fixes the issue in SlicerNNInteractive, until extensions are updated this PR introduces a failsafe to `pip_install`: if `pip_install` is being called from outside the main thread, then refrain from making Qt objects.

### DIFF
--- a/Base/Python/slicer/packaging.py
+++ b/Base/Python/slicer/packaging.py
@@ -514,6 +514,17 @@ def _isSlicerAppAvailable() -> bool:
         return False
 
 
+def _isMainQtThread() -> bool:
+    """Check if the current Python call is running on Slicer's main Qt thread."""
+    if not _isSlicerAppAvailable():
+        return False
+
+    import qt
+    import slicer
+
+    return qt.QThread.currentThread() == slicer.app.thread()
+
+
 def pip_install(
     requirements: str | list[str],
     constraints: str | Path | None = None,
@@ -621,6 +632,21 @@ def pip_install(
         return None
 
     import slicer
+
+    if not _isMainQtThread():
+        logging.warning(
+            "pip_install was called from a non-main thread; disabling progress UI.",
+        )
+        if not blocking:
+            raise RuntimeError(
+                "pip_install(blocking=False) is only supported on the main thread",
+            )
+        if skip_packages is not None:
+            raise RuntimeError(
+                "pip_install(skip_packages=...) is only supported on the main thread",
+            )
+        _pip_install_simple(requirements, constraints, no_deps_requirements)
+        return None
 
     # In testing mode, skip UI and use simple blocking install
     if slicer.app.testingEnabled():


ebrahimebrahim:
I confirmed on my linux system that SlicerNNInteractive, without [my fixes](https://github.com/coendevente/SlicerNNInteractive/pull/100), no longer has any setup issues after the changes in this PR.

ebrahimebrahim:
@lassoan does this look good to merge for 5.12.1?

Base/Python/slicer/packaging.py
```diff
@@ -622,6 +633,21 @@ def pip_install(
 
     import slicer
 
+    if not _isMainQtThread():
+        logging.warning(
```
lassoan, line 637: I don't think we need log this. It is enough to have this as a source code comment, but explain why we don't show progress GUI.

Base/Python/slicer/packaging.py
```diff
@@ -622,6 +633,21 @@ def pip_install(
...
+            "pip_install was called from a non-main thread; disabling progress UI.",
+        )
+        if not blocking:
+            raise RuntimeError(
+                "pip_install(blocking=False) is not supported from a non-main thread",
```
lassoan:
```suggestion
                "pip_install(blocking=False) is only supported on the main thread",
```

Base/Python/slicer/packaging.py
```diff
@@ -622,6 +633,21 @@ def pip_install(
...
+            )
+        if skip_packages is not None:
+            raise RuntimeError(
+                "pip_install(skip_packages=...) is not supported from a non-main thread",
```
lassoan: This can be a problem, as the result of the operation is different depending on which thread it is called from. Why skipping packages is related to threads?

lassoan (commented):
Thank you, it looks good except the skip package limitation. It would be nice to fix that, but it should not block the integration of this PR (since the changes make the implementation better, even if not perfect yet).

Base/Python/slicer/packaging.py
ebrahimebrahim: I am trying to keep this change small and localized. The idea is only to support the old usage of `pip_install` from Slicer 5.10 in the cases where it's being called off the main thread. There was no `skip_packages` in Slicer 5.10, so that case does not need to be supported.

Base/Python/slicer/packaging.py
ebrahimebrahim: I guess I was thinking of this PR as helping to support the transition to the new pip_install, not as officially supporting off-the-main-thread use but just not failing so hard in cases where extensions were already doing that. So a warning makes sense, similar to a depracation warning. But maybe you were thinking we should officially support off-the-main-thread usage? IDK... maybe we should at least discourage it with a warning even if the failsafe remains forever

Base/Python/slicer/packaging.py
lassoan: > There was no skip_packages in Slicer 5.10

OK, you are right, skip was implemented in each extension, so there would be no risk of regression.

ebrahimebrahim:
@lassoan Is this good to go as is, or did you want me to remove the warning first? I kept the warning because I am thinking of pip_install-in-a-thread as something that extensions should fix, and this PR is helping us to tell them that without breaking them outright. But if you are thinking of this PR as re-introducing official support for pip_install-in-a-thread, then I can remove the warning.

Since we said we want this commit as part of 5.12.1 I want to finalize it quickly

ebrahimebrahim:
(I am going to go ahead and cherry pick this commit to the release branch for the 5.12.1 patch release, so no more rush. Tweaking the PR a bit before merging to main is still okay.)

lassoan (approved):
This is good to go as is. If we don't want to support `pip_install` from a background thread then the warning is appropriate.

It makes sense to not support pip_install on background thread until developers specifically request this feature. If the need emerges then the feature could be built into pip_install (e.g., introducing a `waitForCompletion` flag).

PR #6662, opened 2022-11-14
ENH: add simple measurements panel

Punzo:
@lassoan here the PR for the simple measurement panel. 

the widget is a simple version of the GUI of the markups module. The idea is to have a widget that can be easly added to a toolbar and provide a simple view of the markups/measurements with minimum controls, e.g.:

```
markupsToolbar = qt.QToolBar("markupsToolbar")
markupsToolbar.objectName = "markupsToolbar"

simpleMeasurementPanel = slicer.qSlicerSimpleMeasurementsWidget()
simpleMeasurementPanel.setMRMLScene(slicer.mrmlScene)
# simpleMeasurementPanel.setMRMLScene(slicer.mrmlScene, ["vtkMRMLMarkupsLineNode"]) to filter the type of available nodes 
simpleMeasurementPanel.setShowMeasurementDisplay(True) # default False
markupsToolbar.addWidget(simpleMeasurementPanel)
    
slicer.util.mainWindow().addToolBar(qt.Qt.RightToolBarArea, markupsToolbar)
markupsToolbar.show()
```

PR #9274, opened 2026-07-07
BUG: Fix lock button not hiding ROI interaction handles

nhjohnston:
When a markups ROI node was locked via the lock button, the scale/rotate/translate interaction handles remained visible and interactive. Fix by:
- Overriding CanInteract() in vtkSlicerMarkupsInteractionWidgetRepresentation to respect the node-level locked state
- Returning false from IsDisplayable() when the node is locked so handles are hidden
- Calling SetLocked() when the per-control-point lock icon is clicked in the table

Slicer 5.12.0:

https://github.com/user-attachments/assets/9e7347c4-670b-493c-8298-011b9c85f602

This PR:

https://github.com/user-attachments/assets/984bf1a4-4581-4f0d-a6d5-cfc7038b6369

### DIFF
--- a/Modules/Loadable/Markups/VTKWidgets/vtkSlicerMarkupsInteractionWidget.cxx
+++ b/Modules/Loadable/Markups/VTKWidgets/vtkSlicerMarkupsInteractionWidget.cxx
@@ -508,6 +508,31 @@ void vtkSlicerMarkupsInteractionWidget::ScaleWidgetROI(double eventPos[2], bool 
     return;
   }
 
+  // Similarly to translation, scaling must not move locked control points.
+  bool anyControlPointLocked = false;
+  for (int controlPointIndex = 0; controlPointIndex < roiNode->GetNumberOfControlPoints(); ++controlPointIndex)
+  {
+    if (roiNode->GetNthControlPointLocked(controlPointIndex))
+    {
+      anyControlPointLocked = true;
+      break;
+    }
+  }
+  if (anyControlPointLocked)
+  {
+    if (roiNode->GetROIType() == vtkMRMLMarkupsROINode::ROITypeBox)
+    {
+      // The only control point of a box ROI is the center.
+      // The locked center is kept in place by scaling symmetrically around it.
+      symmetricScale = true;
+    }
+    else
+    {
+      // Scaling would reposition all control points, do not scale.
+      return;
+    }
+  }
+
   vtkMRMLMarkupsDisplayNode* displayNode = this->GetDisplayNode();
   if (!displayNode)
   {


pieper (dismissed):
This sounds good to me. 👍 

Anyone else want to have a look before we merge?

Modules/Loadable/Markups/qSlicerMarkupsModuleWidget.cxx
```diff
@@ -2013,6 +2013,7 @@ void qSlicerMarkupsModuleWidget::onActiveMarkupTableCellChanged(int row, int col
       item->setData(Qt::DecorationRole, QPixmap(":/Icons/Small/SlicerUnlock.png"));
     }
     d->MarkupsNode->SetNthControlPointLocked(n, flag);
+    d->MarkupsNode->SetLocked(flag);
```
jamesobutler: What about when the ROI type is "BoundingBox" instead of the normal "Box" type? For "BoundingBox" there can be N control points in the table, but this appears that it will lock the entire markups ROI if just one of the individual control points is locked.
image
image

lassoan:
It gives some flexibility that interaction handles and control points editing can be controlled by two separate flags, but it could be also reasonable to say that locking of the markup includes disabling interactions, too. Maybe the solution could be an additional "read-only" flag disables all modifications.

@nhjohnston Can you tell a bit more about your use case? Could you also disable interactions when you lock the markup node?

jamesobutler:
We don't currently expose the interaction handle settings to enable/show/hide/etc in our Slicer custom application.
image

It does seem confusing though why if you have set the "Locked" state of the markup node, why you would still want to enable translation/rotation/scaling. All of those actions will be modifying the markup node where the "Locked" state was generally set to prevent those actions. Ultimately even if only a individual control point is set as locked, I would expect that the interaction handles should never act upon that control point to change its position. In the case of the "Box" type which just has a single control point, it would seem reasonable that translation through the interaction handle should not be possible when the individual control point is set as locked. Scaling and rotation could be potentially enabled since it doesn't modify the individual control point position that was locked.

Modules/Loadable/Markups/qSlicerMarkupsModuleWidget.cxx
Sunderlandkyl: Yes, I think that calling SetLocked on the full markups node when one of the control points is locked is not the correct behavior.

For point lists/curves/representations/etc. that include multiple control points, we can lock some of them and still translate/rotate the unlocked points using the handles.

nhjohnston:
This all makes sense. So it looks like a good solution would be:

1. Keep the bug fix to lock the center control point of the "box" when the interaction control point lock button is checked
2. Revert the changes to hide the "box" interaction handlers when it is locked (I also think this makes sense, since a locked line still has its end point fiducials shown). This will allow for scaling and rotation when locked. 
3. Fix the bug introduced in this PR to make sure the "bounding box" interaction handles are ALL locked when the interaction control point lock button is checked AND only the individual point is locked when the table control point lock button is checked

Does this sound okay?

nhjohnston:
Added the changes described in https://github.com/Slicer/Slicer/pull/9274#issuecomment-4916935694 for testing. If it looks good, I can squash it into one commit!

Modules/Loadable/Markups/qSlicerMarkupsModuleWidget.cxx
```diff
@@ -1915,6 +1915,10 @@ void qSlicerMarkupsModuleWidget::onListLockedUnlockedPushButtonClicked()
   }
   bool locked = d->MarkupsNode->GetLocked();
   d->MarkupsNode->SetLocked(!locked);
+  for (int i = 0; i < d->MarkupsNode->GetNumberOfControlPoints(); ++i)
+  {
+    d->MarkupsNode->SetNthControlPointLocked(i, !locked);
```
Sunderlandkyl: I don't think that the overall lock button should modify the lock state of the individual control points.

Sunderlandkyl:
How about just hide the interaction handles if either the overall markups lock is enabled, or all of the control points in the markup are locked?

Modules/Loadable/Markups/qSlicerMarkupsModuleWidget.cxx
nhjohnston: Updated!

Modules/Loadable/Markups/VTKWidgets/vtkSlicerMarkupsInteractionWidgetRepresentation.cxx
```diff
@@ -150,6 +151,28 @@ bool vtkSlicerMarkupsInteractionWidgetRepresentation::IsDisplayable()
     return false;
   }
 
+  if (markupsNode->GetLocked())
```
Sunderlandkyl: Thanks, just one more thing. You can use GetNumberOfMovableControlPoints to get the number of unlocked control points.

https://apidocs.slicer.org/main/classvtkMRMLMarkupsNode.html#a5488b9eaa3dee060377f2d80e37d80b0

lassoan:
Interaction handles are general-purpose. They allow changing some properties of the markup. Not all properties are necessarily represented in control points. That's why the option to enable/disable the interaction handles is called "Interaction" and not something more specific such as "Transform".

For example, a handle on a custom cardiac device markup node may not change any control points, but instead change the shape of the device; or we may want to show interaction handles on a robotic arm without wanting to store any control points.

Introducing a logic that strongly links interaction handles to control point transformations would sacrifice generalizability. Maybe it is the right thing to do (for example, because the "Interaction" checkbox meaning may not be obvious for the users; or because we try to over-generalize things, which causes extra complexity for users and developers), but then we would need to rethink the entire role of interaction handles.

It would help if we could discuss about what you would need in your use case instead of jumping to the solution.

nhjohnston:
The initial bug, where the overall markup lock was not locking the markup for box ROI, was the main concern. Hiding the interaction handles was added because, to our users, it might be confusing to have no visual queue that the ROI is now locked. However, I am okay leaving interaction handle functionality out of this PR if it is going to negatively impact other areas! @jamesobutler, what do you think?

jamesobutler:
The main thing for us is that if a Markups ROI of “Box” type is locked then interaction handles should not result in the position or extent of the ROI from changing.

lassoan:
OK, thank for your patience and the additional explanations. It looks like that the problem is very specifically with ROI markup scaling: it may move locked control points (translation, rotation, and other markup types are all good).

I've pushed a change to this branch that fixes this specific problem while leaving everything else unchanged. Please test if it works well for you.

nhjohnston:
Thank you for your push!

I am still seeing the initial discrepancy where the overall lock button allows the user to move the ROI via the control point, while the individual control point lock prevents this. My understanding was that the overall lock button would have the same result as having all control points in the table locked, but I might be misunderstanding?

https://github.com/user-attachments/assets/15d0e9f7-da89-4f78-9326-b003416a8c42

lassoan:
Lock every control point has slightly different compared to global locking:
- Locking the markup control points globally disables all direct interactions with control points in slice and 3D views (you cannot highlight, left-click or right-click on control points in any views). This flag does not affect behavior of interaction handles. This is useful behavior, as it makes markup nodes more similar to other nodes (models, etc.) that you can move by using interaction handles but you cannot modify their shape by click-and-drag. It is also somewhat similar to moving nodes by modifying their parent transform using interaction handles - the points move regardless you locked the markup.
- Locking an individual control is respected by interaction handles - except the ROI widget had a bug that this PR now fixes.

Many workflows depend on this behavior, so it would be probably better not to change it, but we can think about how we can make the behavior further customizable if necessary.

jamesobutler:
Historically it has seemed that locking a markup at the node level prevent any position manipulation of the underlying control points. In the case of a Markups ROI, locking it at the node level shouldn't allow interaction handles to change the position of the underlying control point.

lassoan:
> Historically it has seemed that locking a markup at the node level prevent any position manipulation of the underlying control points. In the case of a Markups ROI, locking it at the node level shouldn't allow interaction handles to change the position of the underlying control point.

I agree that it is conceptually cleaner and simpler if "locking every control point of the markup" is equivalent to "lock the markup". However, in many workflows, we need the ability to click on control points but not allow to move them by drag-and-drop. This is currently achieved by enabling the global lock flag while leaving per-control-point lock flag disabled. We cannot take away this functionality, but there are many ways to keep this working while also achieve what you need.

For example, we could deprecate the "Locked" flag and introduce two new flags that are more specific:
- ControlPointInteractionEnabled: do what the current locked flag does (enable/disable all control point interactions). Maybe it could be added to the markup display node instead of the markup node.
- AllControlPointsLocked: override all the per-control-point locked flags (I think this is what you expect the "Locked" flag to do)

Alternatively, we could add some "locking mode" property that would tell what should happen when the markup is locked. The default could be the current behavior (moving enabled & interaction disabled) and we could introduce new modes ("moving disabled", "moving disabled & interaction handles disabled", ...). Or, we could just add a very specific flag like "InteractionHandlesDisabledWhenLocked" to switch between the current behavior and what you would prefer.

jamesobutler:
I'm trying to think of a way of making this general "Markups" module easier to use rather than being able to do a bunch of different things and therefore confusing to use. I'm not sure what to do at this time. We are trying to use the Markups for simple annotations with the measurements functionality, but all this extra functionality is complicating the use of this module.

lassoan:
Could something like this work better? 

https://github.com/Slicer/Slicer/pull/6662

jamesobutler:
Possibly though that’s about a creating a custom workflow and our group can already do that in our custom app. 

The biggest source of confusion from the Slicer core perspective as it relates specifically to the Markups ROI type is that by default when creating the node the interaction handles are shown. The interaction handles are the only way to resize the object as the “Box” type doesn’t define its size by control points but by the interaction handles. This is different than the default line node which defines its size with control points rather than a single point with 2 interaction handles at either end. Then with Markups ROI of “Box” type with the default interaction handles shown, upon pressing the Markup level “lock” button, it appears to result in a no-op/no change which is the confusing part because you can still click on the center which is actually an interaction handles displayed over the control point and still be able to modify the size of the ROI. Locking a line node results in clicking the points that define the size not being able to translate or change the line as one would expect.

lassoan:
Yes, the main issue is that the ROI and plane markups are special in that interaction handles are essential for defining them. Maybe we could add a new `InteractionAllowedWhileLocked` flag to the markup display node, set to `true` by default for all markups but set to `false` in `vtkMRMLMarkupsPlaneDisplayNode` and `vtkMRMLMarkupsROIDisplayNode`.

nhjohnston:
Added a commit to add InteractionAllowedWhileLocked  so that the ROIs can be interacted with via interaction handles while locked, but the ROI "box" control point is not able to be interacted with while locked

jamesobutler:
@lassoan Do you still have plans to push the updates that we had discussed in our last meeting regarding Markups ROI and Markups Plane?

jamesobutler:
xref https://discourse.slicer.org/t/accidentally-moving-all-markups-in-a-pointlist/47685/8 about a similar discussion of whether mouse-wheel click should respect interaction handle enabled movement or whether it should respect control point interaction locked state.

jamesobutler:
@lassoan Regarding this work were you going to push anything different or do you want to re-check the latest that was pushed here?

lassoan:
I'll check this over the weekend.

jamesobutler (approved):
Approving for integration the minimal agreed upon change of `BUG: Prevent scaling interaction move locked markup ROI control points`. Previous state of branch will be temporarily available at https://github.com/jamesobutler/Slicer/tree/nhjohnston-roi-lock-bug-backup.

Issue #9121, opened 2026-04-21
Need native arm64 app launcher for development to correctly inherit arch


pieper:
## Summary

While using the intel app launcher on arm64 machines mostly works, some low-level python packages can misbehave.

## Steps to reproduce

Launch like this:
```
 /opt/s2/Slicer-build/Slicer 
```
Type this in the python console:
```
>>> import os
>>> os.system("arch")
0
```
and in the terminal you see this:
```
Python console user input: import os
Python console user input: os.system("arch")
i386
0
```

If instead you create a launcher script:
```
./Slicer-build/Slicer --launcher-generate-exec-wrapper-script
```
And launch slicer with the generated script:
```
bash /private/var/folders/yk/qm79wxyd1x38xpn3n046gykh0000gn/T/Slicer-JlUjaY.sh /opt/s2/Slicer-build/bin/Slicer.app/Contents/MacOS/./Slicer
```
then arch is printed as `arm64` as expected.

This became an issue when trying to use `ctypes` on a mac build, which failed like this because it have the wrong architecture.

```
[Python]   File "/opt/s2/python-install/lib/python3.12/ctypes/__init__.py", line 397, in __getitem__
[Python]     func = self._FuncPtr((name_or_ordinal, self))
[Python]            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[Python] AttributeError: dlsym(0x9d937ca0, objc_msgSendSuper_stret): symbol not found
```

## Environment
- Slicer version: Slicer-5.11
- Operating system: Only mac arm systems

lassoan:
According to ChatGPT, to launch an arm64 `myexecutable` from a Rosetta2 application, you have to run `arch -arm64 myexecutable`. It may be enough to change the launcher settings ini file. Of course, a native launcher would be nicer.

PR #9276, opened 2026-07-09
macOS arm64 packaging name fixes

jamesobutler:
This updates the generated packaged installer name to end in "-macosx-arm64.dmg" for arm64 builds instead of "-macosx-amd64.dmg". The `Slicer_ARCHITECTURE` variable gets set correctly now to handle this.

Due to the above change it was also needed to define the CTKAppLauncher to continue to be the amd64 version as that is all that is available. Work to track a native arm64 launcher is defined in https://github.com/Slicer/Slicer/issues/9121. We similarly already define the Windows build to always use an i386 32-bit build as an amd64 Windows version is not available.

### DIFF
--- a/CMake/SlicerMacroGetOperatingSystemArchitectureBitness.cmake
+++ b/CMake/SlicerMacroGetOperatingSystemArchitectureBitness.cmake
@@ -25,7 +25,7 @@
 # The macro defines the following variables:
 #  <var-prefix>_BITNESS - bitness of the platform: 32 or 64
 #  <var-prefix>_OS - which is on the this value: linux, macosx, win
-#  <var-prefix>_ARCHITECTURE - which is on the this value: i386, amd64, ppc
+#  <var-prefix>_ARCHITECTURE - which is on the this value: i386, amd64, arm64, ppc
 
 include(${CMAKE_CURRENT_LIST_DIR}/SlicerBlockOperatingSystemNames.cmake)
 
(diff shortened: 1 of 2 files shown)


CMake/SlicerMacroGetOperatingSystemArchitectureBitness.cmake
```diff
@@ -49,6 +49,20 @@ macro(SlicerMacroGetOperatingSystemArchitectureBitness)
     set(${MY_VAR_PREFIX}_ARCHITECTURE amd64)
   endif()
 
+  # On macOS, CMAKE_SYSTEM_PROCESSOR reflects the host machine (and can be
+  # wrong, for example when CMake is run under Rosetta 2 translation), whereas
+  # CMAKE_OSX_ARCHITECTURES reflects the architecture actually being targeted
+  # by the build. Prefer CMAKE_OSX_ARCHITECTURES when it is set.
+  set(_slicer_target_processor "${CMAKE_SYSTEM_PROCESSOR}")
+  if(APPLE AND NOT "${CMAKE_OSX_ARCHITECTURES}" STREQUAL "")
```
lassoan: Add a note on that `CMAKE_OSX_ARCHITECTURES` can contain multiple values (e.g., `x86_64;arm64`) and describe the rationale of using `arm64` then.

CMake/SlicerMacroGetOperatingSystemArchitectureBitness.cmake
```diff
@@ -49,6 +49,20 @@ macro(SlicerMacroGetOperatingSystemArchitectureBitness)
     set(${MY_VAR_PREFIX}_ARCHITECTURE amd64)
   endif()
 
+  # On macOS, CMAKE_SYSTEM_PROCESSOR reflects the host machine (and can be
```
lassoan: If this is a macOS-only thing then put the entire code block under `if(APPLE)` to make it simpler to interpret the code for other platforms.

CMake/SlicerMacroGetOperatingSystemArchitectureBitness.cmake
```diff
@@ -49,6 +49,20 @@ macro(SlicerMacroGetOperatingSystemArchitectureBitness)
...
+  # wrong, for example when CMake is run under Rosetta 2 translation), whereas
+  # CMAKE_OSX_ARCHITECTURES reflects the architecture actually being targeted
+  # by the build. Prefer CMAKE_OSX_ARCHITECTURES when it is set.
+  set(_slicer_target_processor "${CMAKE_SYSTEM_PROCESSOR}")
```
lassoan: The logic is quite convoluted overall. It is difficult to tell what exactly happens on the 3 platforms (what value is set in `${MY_VAR_PREFIX}_ARCHITECTURE`, if any). It would be also nicer to do away the temporary `_slicer_target_processor` value when we want to just set a single variable. Could we set that variable to some default value for all platforms and then have some extra Apple-specific logic to set it to a different value?

SuperBuild/External_CTKAPPLAUNCHER.cmake
lassoan: Move the comment inside the "win" branch for consistency with the "macosx" branch.

CMake/SlicerMacroGetOperatingSystemArchitectureBitness.cmake
```diff
@@ -25,7 +25,7 @@
 # The macro defines the following variables:
 #  <var-prefix>_BITNESS - bitness of the platform: 32 or 64
 #  <var-prefix>_OS - which is on the this value: linux, macosx, win
-#  <var-prefix>_ARCHITECTURE - which is on the this value: i386, amd64, ppc
+#  <var-prefix>_ARCHITECTURE - which is on the this value: i386, amd64, arm64, ppc
```
lassoan, line 28: Is it guaranteed to be always set? If not then add a not on that.
Can we remove the `i386` and `ppc` architectures or it is still possible that we get these values?
Typo: `on the this value` -> `one the these values`

lassoan (commented):
Thank you, it looks good. It would be nice if you could simplify the logic to make the code easier to read.

CMake/SlicerMacroGetOperatingSystemArchitectureBitness.cmake
jamesobutler: I guess that brings up another point should we fail earlier if someone attempts to build for "x86_64;arm64"? I don't believe we are going to aim to support a universal binary option. There will only be support for x86_x64 macOS package or arm64 macOS package.

lassoan: If the macro might not behave correctly with multiple architecture values then it is easier to stop with a failure instead of trying to figure out a meaningful behavior. So, adding a simple check here would be good.

I see that `SlicerInitializeOSXVariables.cmake` already has a check for multiple values. If that is called early enough then we don't need to add another early check.

CMake/SlicerMacroGetOperatingSystemArchitectureBitness.cmake
jamesobutler: Would you like that we have a commit for bulk remove of older 32-bit based architecture logic across all the various 3D Slicer build infrastructure files consider Slicer is no longer supported being built as a 32-bit app? For example there is more of this in places such as the following:
https://github.com/Slicer/Slicer/blob/143b942237aaded8a1eb09b6371573db1e96ecd8/Extensions/CMake/SlicerBlockUploadExtensionPrerequisites.cmake#L12-L17
https://github.com/Slicer/Slicer/blob/143b942237aaded8a1eb09b6371573db1e96ecd8/CMake/SlicerBlockInstallOpenSSL.cmake#L16-L19

CMake/SlicerMacroGetOperatingSystemArchitectureBitness.cmake
jamesobutler: I've updated the logic here to be more generic in handling cases of CMake being amd64 and running through emulation on an arm64 machine where the build is targeting arm64.

CMake/SlicerMacroGetOperatingSystemArchitectureBitness.cmake
jamesobutler: I have adjusted the logic to set in each of the platform statement blocks with their preferred methods of setting the architecture.

CMake/SlicerMacroGetOperatingSystemArchitectureBitness.cmake
jamesobutler: I have confirmed that things are indeed already handled early when trying to specify a universal binary.

At the initial configuration step it will exit early with:
```
.
.
-- Setting C++ standard
-- Setting C++ standard - C++17
CMake Error at CMake/SlicerInitializeOSXVariables.cmake:39 (message):
  error: Only one value should be associated with CMAKE_OSX_ARCHITECTURES
  (for example: arm64 or x86_64).
  CMakeLists.txt:69 (include)


-- Configuring incomplete, errors occurred!
```

CMake/SlicerMacroGetOperatingSystemArchitectureBitness.cmake
hjmjohnson: I do think we should remove all 32-bit-based architecture logic.  It is more noise with no benefit.

jamesobutler: I’ll issue a separate PR for independent review regarding stripping out older 32-bit build support logic.

lassoan (approved):
Thank you, this looks very nice and clear now.

PR #9277, opened 2026-07-09
Set 3D rendering "Shade" to off when synchronized with 2D display settings

ajallphin:
By definition, there is no shading/lighting in the 2D slice views. This PR turns off "Shading" in the volume rendering module whenever synchronized with the 2D slice view display settings. I personally like the way it looks better. Also, with shading on, unless you dial the "Ambient" material setting up to 100, the intensity values shown in the 3D view will always be less than what is shown in the slice views.

| Shade On |
|--------|
||

| Shade Off |
|--------|
|| 

This PR is entirely based on preference, so I won't be offended if it is simply closed and dismissed.

We could also decouple the "Shade" and lighting properties from the 2D to 3D synchronization.

### DIFF
--- a/Modules/Loadable/VolumeRendering/Logic/vtkSlicerVolumeRenderingLogic.cxx
+++ b/Modules/Loadable/VolumeRendering/Logic/vtkSlicerVolumeRenderingLogic.cxx
@@ -607,7 +607,7 @@ void vtkSlicerVolumeRenderingLogic::SetWindowLevelToVolumeProp(double scalarRang
   }
 
   volumeProp->SetInterpolationTypeToLinear();
-  volumeProp->ShadeOn();
+  volumeProp->ShadeOff();
   volumeProp->SetAmbient(0.30);
   volumeProp->SetDiffuse(0.60);
   volumeProp->SetSpecular(0.50);
--- a/Base/Python/slicer/packaging.py
+++ b/Base/Python/slicer/packaging.py
@@ -514,6 +514,17 @@ def _isSlicerAppAvailable() -> bool:
         return False
 
 
+def _isMainQtThread() -> bool:
+    """Check if the current Python call is running on Slicer's main Qt thread."""
+    if not _isSlicerAppAvailable():
+        return False
+
+    import qt
+    import slicer
+
+    return qt.QThread.currentThread() == slicer.app.thread()
+
+
 def pip_install(
     requirements: str | list[str],
     constraints: str | Path | None = None,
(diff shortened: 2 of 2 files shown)


pieper (dismissed):
LGTM 👍 

Anyone see any downside to this?  Maybe make it an opt-in preference for now and make it a breaking change to change it to the default for 6.0?

lassoan:
Right now, if you enable synchronization then shading is forced to be on, which is an unnecessary limitation. We could decouple the `Shade` option from the `Synchronize with Volumes module` option and let the user enable/disable shading independently.

Forcing synchronization to disable shading would not be welcome by all users. For example, we could ask @muratmaga, who have been using synchronization with shading.

Also note that volume rendering can be combined with slice display in 3D (show slice in 3D + set volume rendering clipping with the same slice view), which gives the exactly same cross-sectional display of the volume in 3D as in the slice view, while you would still see nice 3D rendering:

image

Modules/Loadable/VolumeRendering/Logic/vtkSlicerVolumeRenderingLogic.cxx
```diff
@@ -607,7 +607,7 @@ void vtkSlicerVolumeRenderingLogic::SetWindowLevelToVolumeProp(double scalarRang
   }
 
   volumeProp->SetInterpolationTypeToLinear();
-  volumeProp->ShadeOn();
+  volumeProp->ShadeOff();
```
lassoan, line 610: Synchronization is about linking slice view lookup table and volume rendering transfer functions. It is not necessary to also hardcode some unrelated lighting parameters here. I would suggest removing ShadeOff, SetAmbient, SetDiffuse, SetSpecular, SetSpecularPower calls from this method.
```suggestion
```

lassoan:
@ajallphin please confirm that removing the hardcoded lighting settings (instead of changing the hardcoded values) would be an acceptable solution for you.

ajallphin:
Yes, removing the hardcoded settings link is great. However, that will also require changes to make sure the interface isn't locked for editing once "synchronization" is checked.

jamesobutler:
@lassoan would you be able to push the changes regarding the removal of the hardcoded lighting settings based on the latest comments? Then we can integrate/finalize this PR as well.

lassoan (approved):
Thanks, removing shade setting should work well.

lassoan:
I've accidentally closed this PR (the branch has become empty because of the commit was rejected due to long first line), so I created a follow-up at https://github.com/Slicer/Slicer/pull/9390

PR #9284, opened 2026-07-13
ENH: Add Scroll mouse mode to browse slices by click-and-drag

lassoan:
Add a new "Scroll" mouse interaction mode alongside the existing
Translate/rotate view, Adjust window/level, and Place modes.

This is consistent with other radiology viewers (Weasis, OsiriX, etc.)
and an older IHE supplement that many viewers adopted
(https://www.ihe.net/uploadedFiles/Documents/Radiology/IHE_RAD_Suppl_BIR_Rev1.3_TI_2016_09-09.pdf),
and users requested (https://discourse.slicer.org/t/is-it-possible-to-scroll-by-dragging-the-mouse-in-the-current-view/35964/4).

A few additional modes (zoom, pan, crosshair) may be considered,
which can be useful when using single-button mouse or touchscreen.

When this new mode is active, click-and-drag vertically in a slice view scrolls through
slices, with the full view height mapped to the full slice offset slider range.

Implementation:
New mode is added to vtkMRMLInteractionNode.
vtkMRMLSliceIntersectionWidget implements the view interaction.
Added new icon (stack with blue up/down arrow) to the toolbar and shortcut to view context menu.

image


Libs/MRML/DisplayableManager/vtkMRMLSliceIntersectionWidget.cxx
```diff
@@ -1305,6 +1320,59 @@ void vtkMRMLSliceIntersectionWidget::MoveSlice(double delta)
...
+
+  const int* windowSize = this->GetRenderer()->GetRenderWindow()->GetSize();
+  if (!windowSize || windowSize[1] <= 0)
+  {
+    return;
+  }
+
+  // The total height of the view corresponds to the total range of the slice offset slider.
+  double offsetPerPixel = (rangeHigh - rangeLow) / windowSize[1];
+  double newOffset = this->ScrollStartSliceOffset + offsetPerPixel * (eventPosition[1] - this->StartEventPosition[1]);
```
pieper, line 1363: I think allowing either up/down or right/left is more natural feeling.  For the radiologist use case they want to minimize muscle action to avoid fatigue, so any reasonable movement should be translated into useful effect.

Here's an example to try:

https://pieper.github.io/live/viewer.html?scene=scenes/MRHead.json

pieper (changes requested):
The implementation looks good.  I am thinking we should consider making this the default for Slicer6.  There's no real reason not to have a working mouse mode active when the app starts, and this makes more sense than window/level to me.

lassoan:
> The implementation looks good. I am thinking we should consider making this the default for Slicer6. There's no real reason not to have a working mouse mode active when the app starts, and this makes more sense than window/level to me.

If we load 3D data (and we usually do) and we show original views then click-and-drag in slice views moving the slice intersections (instead of shift+mousemove) could be more useful. We could add a new "crosshair" mode for this and making it the default, or we could make the current default mode to move the crosshair in slice views (but still rotate the 3D views).

If we show a single slice view then scrolling that view instead of moving the crosshair may be more useful 

We could add a default mode selector to the application settings, but we would need to decide if we want to implement some smart mode where we switch between scrolling (when in a single slice view), crosshair (when in a slice view with multiple slice views visible), and rotation (when in 3D view) action.

pieper:
Since we are brainstorming about usability, I think we should make switching the mouse modes more discoverable and easier to access.  Currently the mode buttons are buried with dozens of other toolbar buttons.  Something overlayed on the viewers would make more sense for something people are expected to use a lot.

cpinter:
> mouse modes more discoverable and easier to access

The mouse modes are the first options on the top of the context menu when you right-click a view. I think it's as easy as it gets. I also don't like the crowded toolbar, but I don't know where we could move the mouse mode buttons so that they are more prominent but don't steal useful space.

image

lassoan:
I would like to merge this, as it would help a lot with browsing slices of a volume on a phone. We can decide later if we want to make it the default mouse mode later - for now we would keep the current one.

pieper (approved):
👍

jamesobutler:
> When this new mode is active, click-and-drag vertically in a slice view scrolls through
> slices

Thoughts on whether this mouse mode could also scroll through the time dimension for sequences on mouse wheel scroll? Another way to avoid using QWidgets. However this interaction wouldn't translate to phone touch events in terms of going forward/backward for frames in a sequence.

lassoan:
Time sequence browsing would make sense, too. I would add a separate effect for that.

PR #9289, opened 2026-07-15
DOC: Update release issue template checklists

ebrahimebrahim:
Checkboxes are easier to deal with than emojis

emojis can be added just for specific situations


sjh26 (approved):
lgtm ✅

PR #9294, opened 2026-07-17
ENH: Add installer scripts to Slicer

mauigna06:
[Original repository](https://github.com/mauigna06/slicer-installer/tree/main)

As discussed on the Slicer meeting, this is a PR to make the slicer-installer script available. 
A few governance questions I could not answer myself:

> a no-interaction mode, for people who want to use the script in automations

1. When the target version is already installed and you haven’t set SLICER_IF_EXISTING env variable, what should a no-interaction run do: reinstall it (today’s no-terminal behavior) or exit 0 without touching it? This is the idempotency question — it decides whether repeated automation runs re-download ~1 GB or become no-ops.
2. Do you also need quiet output (no logo, no progress bar) for logs, or is suppressing prompts enough? (default: prompts only — logo and progress bars are already suppressed when output isn’t a terminal)
3. Linux/macOS only, or do you automate the Windows installer too?

> an automatic dependency installation mode, for people who want to use the script to create docker images

4. If installing dependencies fails, should the whole install fail (good for RUN steps in a Dockerfile) or warn and continue?
5. If the script is not root, should it attempt sudo (which may password-prompt — colliding with no-interaction mode) or refuse and print the command as it does today?
6. Do you run Slicer headless in the container (e.g. xvfb-run Slicer --no-main-window)? If so, should the mode optionally install headless extras like xvfb, or stay strictly at Slicer’s runtime library list?

> the possibility of packaging an installer script into the tarball, in the linux case

7. Who is it for: people who manually download the tarball from [download.slicer.org](http://download.slicer.org/) and want the wiring (launcher symlink, menu entry, deps hint) afterwards, or is the goal to get upstream Slicer packaging to bundle the script officially?
8. When run from an extracted tree, should it wire that tree up in place (wherever it was extracted), move it to the standard ~/.local/opt location, or ask?
9. Should it be the same install.sh gaining an offline/local mode (one file, could be dropped into the tarball as-is), or a separate minimal script that contains only the wiring logic?
10. Must it work fully offline (no checksum/version lookups at all)?
11. Should it also offer uninstall of that copy? (default: yes — the uninstall logic already exists in uninstall_linux)

> representing system dependencies in a separate file that the script references and that the docs can also reference to report system dependencies, so that they are in one place and never go out of sync

12. Who consumes the file besides this repo’s README — e.g. the official Slicer docs or a Discourse post wanting a stable URL? Decides whether the file format/location becomes a public contract.
13. Plain text lists per package manager (deps/apt.txt, deps/dnf.txt, deps/pacman.txt) or something structured (JSON/YAML) for machine consumption?
14. Is a repo-internal sync acceptable — the lists get embedded into install.sh by a generator, with a CI check that fails on drift — or do you want the script to fetch the file at runtime so even old script copies stay current? Runtime fetching adds a network failure mode to curl | sh; embedding keeps the script self-contained.
15. Should the file format express conditionals (e.g. libasound2t64 vs libasound2 depending on Ubuntu release), or does that logic stay in the script with the file listing the canonical names?


mauigna06:
> > a no-interaction mode, for people who want to use the script in automations 

Added.

> 1. When the target version is already installed and you haven’t set SLICER_IF_EXISTING env variable, what should a no-interaction run do: reinstall it (today’s no-terminal behavior) or exit 0 without touching it? This is the idempotency question — it decides whether repeated automation runs re-download ~1 GB or become no-ops.

We want the installation to succeed when there is no interactive terminal so the current behavior to _reinstall_ Slicer is OK

> 2. Do you also need quiet output (no logo, no progress bar) for logs, or is suppressing prompts enough? (default: prompts only — logo and progress bars are already suppressed when output isn’t a terminal)

You can suppress logs using SLICER_QUIET

> 3. Linux/macOS only, or do you automate the Windows installer too?

Of all them should work

> > an automatic dependency installation mode, for people who want to use the script to create docker images

Added.

> 4. If installing dependencies fails, should the whole install fail (good for RUN steps in a Dockerfile) or warn and continue?

Then whole install fails so RUN in Dockerfile would fail too

> 5. If the script is not root, should it attempt sudo (which may password-prompt — colliding with no-interaction mode) or refuse and print the command as it does today?

If the user is not root or sudo then the dependencies are only suggested and sudo is not requested

> 6. Do you run Slicer headless in the container (e.g. xvfb-run Slicer --no-main-window)? If so, should the mode optionally install headless extras like xvfb, or stay strictly at Slicer’s runtime library list?

Only install Slicer’s runtime library list

mauigna06:
@lassoan 

Could you comment on these:



> > the possibility of packaging an installer script into the tarball, in the linux case
> 
>     7. Who is it for: people who manually download the tarball from [download.slicer.org](http://download.slicer.org/) and want the wiring (launcher symlink, menu entry, deps hint) afterwards, or is the goal to get upstream Slicer packaging to bundle the script officially?
> 
>     8. When run from an extracted tree, should it wire that tree up in place (wherever it was extracted), move it to the standard ~/.local/opt location, or ask?
> 
>     9. Should it be the same install.sh gaining an offline/local mode (one file, could be dropped into the tarball as-is), or a separate minimal script that contains only the wiring logic?
> 
>     10. Must it work fully offline (no checksum/version lookups at all)?
> 
>     11. Should it also offer uninstall of that copy? (default: yes — the uninstall logic already exists in uninstall_linux)
> 
> 
> > representing system dependencies in a separate file that the script references and that the docs can also reference to report system dependencies, so that they are in one place and never go out of sync
> 
>     12. Who consumes the file besides this repo’s README — e.g. the official Slicer docs or a Discourse post wanting a stable URL? Decides whether the file format/location becomes a public contract.
> 
>     13. Plain text lists per package manager (deps/apt.txt, deps/dnf.txt, deps/pacman.txt) or something structured (JSON/YAML) for machine consumption?
> 
>     14. Is a repo-internal sync acceptable — the lists get embedded into install.sh by a generator, with a CI check that fails on drift — or do you want the script to fetch the file at runtime so even old script copies stay current? Runtime fetching adds a network failure mode to curl | sh; embedding keeps the script self-contained.
> 
>     15. Should the file format express conditionals (e.g. libasound2t64 vs libasound2 depending on Ubuntu release), or does that logic stay in the script with the file listing the canonical names?

muratmaga:
@mauigna06  We provision Slicer onto Jetstream2 VMs with Ansible (a fork of exosphere's playbooks),
currently via SlicerDocker's `slicer-download.sh`. So we're one of the
"automation" consumers this script is aimed at.  So some answers with the assistance of Claude:


**Version pinning (not in the list, but the blocker for us).** The script only
exposes `SLICER_VERSION`, which maps to `download.slicer.org?version=`. We pin by
Kitware *revision* (`-r 34627`), because that's the only unambiguous handle once
you're off stable releases, and it's what guarantees every VM in a course gets a
byte-identical build. Could `SLICER_REVISION` be added as an override? Without
it, this script can't replace `slicer-download.sh` for us.

**Q1 (idempotency).** Please make no-interaction + already-installed a no-op that
exits 0. Re-downloading ~1 GB on every converge is the difference between a
config-management tool that can run repeatedly and one that can't. `reinstall`
stays available explicitly via `SLICER_IF_EXISTING`.

**Q4 (deps failure).** Fail the install. A half-provisioned image that fails at
first launch is much more expensive to debug than a failed build step.

**Q5 (sudo).** Attempting sudo is fine, but it must never block on a password
prompt when `SLICER_NONINTERACTIVE` is set — `sudo -n` and a clear error is
better than a hang in CI.

**Q6 (headless extras).** No — we run a real desktop session over TurboVNC.
Please keep the mode at Slicer's runtime library list.

**Q12–15 (dependency file).** Yes, we'd consume it — that's a second consumer
besides your README. We hand-maintain the apt list in our Ansible role and it has
already drifted: your list carries `libsm6`, ours doesn't, and we independently
hit the `libasound2t64` rename on the 24.04 bump. Plain text per package manager
is plenty; JSON buys us nothing. Prefer embedded-in-script + a CI drift check
over a runtime fetch — a `curl | sh` installer shouldn't gain a second network
dependency — but please publish the file at a stable path so tools like ours can
read it directly from the repo.

One note on the script as it stands: it caches packages under
`~/.cache/slicer-installer`, which bakes ~1 GB into a container/VM image unless
`XDG_CACHE_HOME` is redirected. Worth a mention in the docs for the Docker case.

mauigna06:
Hi Murat,

Thanks a lot for the feedback. I'm working on it (here: https://github.com/mauigna06/slicer-installer/tree/main) and will post a message when ready to review. I also had comments from the core-devs on the Tuesday meeting so that is coming too

> Version pinning (not in the list, but the blocker for us). The script only
exposes SLICER_VERSION, which maps to download.slicer.org?version=. We pin by
Kitware revision (-r 34627), because that's the only unambiguous handle once
you're off stable releases, and it's what guarantees every VM in a course gets a
byte-identical build. Could SLICER_REVISION be added as an override? Without
it, this script can't replace slicer-download.sh for us.

Done

> Q1 (idempotency). Please make no-interaction + already-installed a no-op that
exits 0. Re-downloading ~1 GB on every converge is the difference between a
config-management tool that can run repeatedly and one that can't. reinstall
stays available explicitly via SLICER_IF_EXISTING.

Done

mauigna06:
I think I should also add an env variable to set the Slicer language (using the LanguagePacks extension internally)

mauigna06:
> Q4 (deps failure). Fail the install. A half-provisioned image that fails at
first launch is much more expensive to debug than a failed build step.

Done

> Q5 (sudo). Attempting sudo is fine, but it must never block on a password
prompt when SLICER_NONINTERACTIVE is set — sudo -n and a clear error is
better than a hang in CI.

Done

> Q6 (headless extras). No — we run a real desktop session over TurboVNC.
Please keep the mode at Slicer's runtime library list.

Like it is, does not a change

mauigna06:
> I think I should also add an env variable to set the Slicer language (using the LanguagePacks extension internally)

Done

mauigna06:
> One note on the script as it stands: it caches packages under
~/.cache/slicer-installer, which bakes ~1 GB into a container/VM image unless
XDG_CACHE_HOME is redirected. Worth a mention in the docs for the Docker case.

Done. Document how to use the installer script to not bloat a docker image that uses it

mauigna06:
> Q12–15 (dependency file). Yes, we'd consume it — that's a second consumer besides your README. We hand-maintain the apt list in our Ansible role and it has already drifted. Plain text per package manager is plenty. Prefer embedded-in-script + a CI drift check over a runtime fetch — a curl | sh installer shouldn't gain a second network dependency — but please publish the file at a stable path so tools like ours can read it directly from the repo.

@muratmaga Would an environment variable, with a default value you can replace, as the url from where to read the list of to-be-installed dependencies be good enough?

muratmaga:
@mauigna06 Thanks — `SLICER_REVISION` and the `exit 0` on an already-installed
version are exactly what we needed. I confirmed
`download.slicer.org?os=linux&revision=34627` resolves to `Slicer_linux_amd64_34627`
(5.12.3), which is the build our VMs pin today, so that path works end to end.

On the dependency list: an env var holding a URL the script reads at install
time isn't what we'd want, for two reasons.

**It adds a second network dependency.** Right now the script needs the network
once, to fetch the package. Reading the dependency list over the network too
means a `curl | sh` install can fail in a new way — reachable download server,
unreachable list — and that failure lands in the middle of provisioning rather
than up front.

**An overridable URL decides what gets installed as root.** Whatever that list
says gets handed to `apt-get install` under sudo. That turns a formatting
convenience into a way to redirect package installation, which is a lot of
authority for an env var.

The thing that actually helps us is smaller: we don't need the script to fetch
the list, we need the list to **exist at a stable path in the repo** so we can
diff it against the apt list in our Ansible role and catch drift. So:

- Keep the canonical list as plain text in the repo, one file per package manager.
- Embed it into `install.sh` at build time, with a CI check that fails on drift.
- The script stays self-contained and offline for deps; we read the file directly
  from the repo.

If you still want the override, that's fine as an opt-in escape hatch — the
important part is that the default never hits the network.

One thing the file format has to settle either way (your Q15): the
`libasound2t64` / `libasound2` split. We hit that independently on the Ubuntu
24.04 bump. Either the file expresses the condition, or it lists canonical names
and `install.sh` keeps the `apt-cache show` probe it has now — but whichever it
is, a downstream consumer reading the file needs to know which.

mauigna06:
Ok Murat, updated according to your request, please test and let me know if now is fine

mauigna06:
Ok. Hopefully, this satisfies the requirements for inner and outer scripts. I still need to test it by creating a package myself

PR #9296, opened 2026-07-18
BUG: Fix editing of segment opacity by double-click in segments table

lassoan:
Since the visibility/color button rework (94ad4fb3e8), the DoubleClicked edit trigger is disabled in the segments table (double-click on the name and color columns opens the terminology selector instead), which left F2 as the only way to edit the opacity column.

Open the opacity editor explicitly from the double-click handler to restore the previous behavior.

Fixes https://discourse.slicer.org/t/opacity-cell-5-12-2/47652


cpinter (approved):
It works perfectly, thank you very much!

Issue #5040, opened 2020-07-11
Fix custom unit management


lassoan:
For numerical stability of various processing methods, it is important to avoid very small and very large numbers.

By default, Slicer uses millimeter as unit, therefore for microscopy images the image spacing could be very small (for example, 0.00006mm) and derived values even smaller (volume of a voxel would be 0.000000000000216mm3). This could lead to display issues (hard to read the numbers, not enough space to display them) and numerical instabilities and loss of precision. The solution is to use a more appropriate unit, such as micrometer or tenth of a micrometer.

Slicer allows specifying what is the unit of length values of the scene and has unit-aware slider and spinbox widgets to correctly display them. However, the implementation is not fully complete and consistent. Remaining tasks:

- [ ] Use unit-aware widgets (or unit-aware formatting) whenever displaying numerical values with units. There are a number of places where “mm” unit is hardcoded and should be display should be replaced with unit-aware widgets. I remember I had crashes with these widgets in Transform module’s display section and could not easily find the root cause, so just hardcoded mm, but we should give this another go. Examples include: view rulers (horizontal line at the bottom of views) and markups measurements (length of line and curves, surface area of closed curves).
- [ ] Importers should convert distance units. For example, when loading a DICOM file and the length unit is not mm, the voxel size must be scaled accordingly.
- [ ] Exporters should write distance units into files.
- [ ] CLI module interface should be made unit-aware.
- [ ] All transformable data in the scene must be rescaled when length unit is changed.
- [ ] All transformable data in the scene must be rescaled when a scene with a different unit is imported.

Until all these are implemented, a workaround is to enter values in a different unit (e.g., enter 0.06 "mm" instead of the real 0.06 micrometer), but don't change the unit in application settings. Then user would know that all units that Slicer displays in millimeters are actually in the other chosen unit (i.e., micrometer). See some more details and explanation here: https://discourse.slicer.org/t/distance-measurement-and-rendering-of-microscopy-images/791

PR #9300, opened 2026-07-20
DOC: Fix rounding example in unit precision description

anthonyylee:
Corrected the example rounding in the precision description.

### DIFF
--- a/Modules/Loadable/Units/Widgets/Resources/UI/qMRMLUnitWidget.ui
+++ b/Modules/Loadable/Units/Widgets/Resources/UI/qMRMLUnitWidget.ui
@@ -140,7 +140,7 @@ p, li { white-space: pre-wrap; }
 p, li { white-space: pre-wrap; }
 &lt;/style&gt;&lt;/head&gt;&lt;body style=&quot; font-family:'MS Shell Dlg 2'; font-size:8.25pt; font-weight:400; font-style:normal;&quot;&gt;
 &lt;p style=&quot; margin-top:0px; margin-bottom:0px; margin-left:0px; margin-right:0px; -qt-block-indent:0; text-indent:0px;&quot;&gt;&lt;span style=&quot; font-size:8pt;&quot;&gt;Set the precision (i.e. number of significant digits) of the unit.&lt;/span&gt;&lt;/p&gt;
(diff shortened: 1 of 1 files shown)


lassoan (commented):
Thanks for looking into this. I think precision currently refers to significant digits, so for 1.0123 and precision 3 the displayed value will be 1.01 - see

```
const char* vtkMRMLUnitNode::GetDisplayStringFormat()
{
  std::stringstream strstream;
  strstream << this->GetPrefix();
  strstream << "%";
  strstream << "-";                                              // left justify
  strstream << "#";                                              // force decimal point
  strstream << floor(log10(1. + fabs(this->GetMaximumValue()))); // pad
  strstream << "." << this->GetPrecision();                      // decimals
  strstream << "g";
  strstream << this->Suffix;
  strstream >> this->LastDisplayString;
  return this->LastDisplayString.c_str();
}
```

We can fix this in the GUI now as you suggested (but writing `1.01` there instead of `1.012`), but if we want to add good unit support then I think we need to improve the formatting behavior. We can continue the discussion at https://github.com/Slicer/Slicer/issues/5040. Probably the MF initiative could be a good opportunity to work on some improvements.

anthonyylee:
Hey @lassoan, Thanks for the rundown and I agree that it should be 1.01 instead.
I am not familiar with the MF Initiative. Is this related to the Manufacturing Format Specification?

lassoan:
MF is just a new project that may need this feature and provide resources for the development.

Have a look at the current limitations at https://github.com/Slicer/Slicer/issues/5040 and maybe add some information about your use case.

anthonyylee:
I see.
After reviewing #5040, I have a better understanding of how Slicer works and some of its limitations. However, my current workflow and analyses do not require precision that have perceivable issue from numerical instability or unit differences. I have been able to work with `mm`.

The PR was prompted when I was reading the pop-up information box when I was trying to view more decimal points when using the Markups module.

I have also updated the PR to `1.01` instead of the previous `1.012`.

anthonyylee:
@lassoan are there any other changes I should make before merging and closing this PR?

lassoan (commented):
Thank you, it looks good.

lassoan (approved):
Thank you, it looks good.

PR #9293, opened 2026-07-17
COMP: Fix compiler warnings from Qt 6 and VTK 9 deprecated APIs

pieper:
## Summary

Fixes ~150 compiler warnings emitted when building Slicer on macOS against Qt 6.10 and current VTK, while remaining compatible with Qt 5.15 (all replacements compile warning-free on both):

- `QVariant::type()`/`QVariant::Type` comparisons replaced with `userType()` and `QMetaType` enums; deprecated `QVariant::canConvert(int)` replaced with the templated `canConvert<T>()`
- deprecated `QScopedPointer::take()` replaced with `std::unique_ptr`/`release()`
- deprecated `QKeySequence` modifier `operator+` replaced with `operator|`
- `Qt::AA_EnableHighDpiScaling`/`AA_UseHighDpiPixmaps` (always-on no-ops in Qt 6) guarded with `QT_VERSION < 6`; `QImage::flipped()` used on Qt >= 6.9
- deprecated `QMouseEvent` constructor replaced with the overload taking a global position
- deprecated `vtkCellArray` methods replaced: `GetCell()` -> `GetCellAtId()`, `Allocate(EstimateSize())`/`Allocate(GetSize())` -> `AllocateEstimate()`/`AllocateExact()`, `ReplaceCell()` -> `ReplaceCellAtId()`; no-op `SetNumberOfCells()` and vestigial traversal calls removed
- `VTK_FALLTHROUGH` -> `[[fallthrough]]`
- `sprintf` -> `snprintf` (deprecated on macOS for security reasons)
- deprecated `CURLOPT_PUT` -> `CURLOPT_UPLOAD`
- unchecked `QFile::open()` nodiscard warnings fixed by handling failures
- missing `override` added to `vtkMRMLTransformSequenceStorageNode::SupportedFileType()`; implicit int64-to-double conversion made explicit in `vtkITKArchetypeImageSeriesReader`

## Deliberately deferred (hence draft/WIP)

- ~40 `invalidateFilter()` deprecations: the replacement `begin/endFilterChange()` only exists in Qt 6.10, needs a policy decision while Qt 5.15 remains supported
- 5 warnings in `vtkPolyDataToFractionalLabelmapFilter`: require a real rewrite of location-based `vtkCellArray` traversal, best done with test coverage
- warnings in sibling repos (CTK, CTKAppLauncher, vtkAddon, SlicerSurfaceToolbox, MultiVolumeExplorer) and vendored third-party code

[...]

PR #9305, opened 2026-07-22
BUG: Restore dropped Python wrapping hierarchy dependencies

pieper:
With the Makefiles generator, the Python wrapping hierarchy custom command of a module depends on the hierarchy targets of the modules it wraps against, but only if those targets already exist when the module is configured (see vtkWrapHierarchy.cmake in vtkAddon). Transforms and Segmentations are configured before Markups, so their hierarchy rules silently lost the dependency on the Markups MRML hierarchy and parallel builds could fail with:

  vtkWrapHierarchy: couldn't open file .../vtkSlicerMarkupsModuleMRMLHierarchy.txt

Incremental build trees masked the problem because the hierarchy file from a previous build was already on disk; fresh build trees hit the race depending on scheduling.

Restore the two known forward dependencies after all modules are configured, where every hierarchy target exists. An audit of the module list found no other forward references between VTK-wrapped module libraries (the SubjectHierarchy reference to Volumes is in a PythonQt wrapped library, which has no hierarchy target). A generic fix (for example deferring the dependency wiring with cmake_language(DEFER)) belongs in vtkAddon.

### DIFF
--- a/Modules/Loadable/CMakeLists.txt
+++ b/Modules/Loadable/CMakeLists.txt
@@ -40,3 +40,21 @@ foreach(module ${qtmodules})
     add_subdirectory(${module})
   endif()
 endforeach()
+
+# With the Makefiles generator, the Python wrapping hierarchy of a module can
+# only depend on the hierarchy target of another module if that target already
+# exists when the module is configured (see vtkWrapHierarchy.cmake in
+# vtkAddon). For modules that depend on a module configured later in the list
+# above, the dependency is silently dropped and parallel builds can fail with
+# "vtkWrapHierarchy: couldn't open file ...Hierarchy.txt". Restore the known
+# forward dependencies here, where all hierarchy targets exist.
+foreach(_hierarchy_dependency IN ITEMS
+    "vtkSlicerTransformsModuleLogicHierarchy;vtkSlicerMarkupsModuleMRMLHierarchy"
+    "vtkSlicerSegmentationsModuleLogicHierarchy;vtkSlicerMarkupsModuleMRMLHierarchy"
+    )
+  list(GET _hierarchy_dependency 0 _hierarchy_target)
+  list(GET _hierarchy_dependency 1 _hierarchy_dependee)
+  if(TARGET ${_hierarchy_target} AND TARGET ${_hierarchy_dependee})
+    add_dependencies(${_hierarchy_target} ${_hierarchy_dependee})
+  endif()
+endforeach()
--- a/Modules/Loadable/CMakeLists.txt
+++ b/Modules/Loadable/CMakeLists.txt
@@ -41,13 +41,16 @@ foreach(module ${qtmodules})
   endif()
 endforeach()
 
-# With the Makefiles generator, the Python wrapping hierarchy of a module can
+# With the current cmake macros, the Python wrapping hierarchy of a module can
 # only depend on the hierarchy target of another module if that target already
 # exists when the module is configured (see vtkWrapHierarchy.cmake in
 # vtkAddon). For modules that depend on a module configured later in the list
 # above, the dependency is silently dropped and parallel builds can fail with
 # "vtkWrapHierarchy: couldn't open file ...Hierarchy.txt". Restore the known
 # forward dependencies here, where all hierarchy targets exist.
+# Note that this code may change if we use vtk_module_build() / vtk_module_wrap_python() 
+# to compute hierarchy dependencies from declared module metadata (vtk.module files), 
+# which will be independent of configure order
 foreach(_hierarchy_dependency IN ITEMS
     "vtkSlicerTransformsModuleLogicHierarchy;vtkSlicerMarkupsModuleMRMLHierarchy"
     "vtkSlicerSegmentationsModuleLogicHierarchy;vtkSlicerMarkupsModuleMRMLHierarchy"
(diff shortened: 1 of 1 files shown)


Modules/Loadable/CMakeLists.txt
```diff
@@ -40,3 +40,21 @@ foreach(module ${qtmodules})
...
+
+# With the Makefiles generator, the Python wrapping hierarchy of a module can
+# only depend on the hierarchy target of another module if that target already
+# exists when the module is configured (see vtkWrapHierarchy.cmake in
+# vtkAddon). For modules that depend on a module configured later in the list
+# above, the dependency is silently dropped and parallel builds can fail with
+# "vtkWrapHierarchy: couldn't open file ...Hierarchy.txt". Restore the known
+# forward dependencies here, where all hierarchy targets exist.
+foreach(_hierarchy_dependency IN ITEMS
+    "vtkSlicerTransformsModuleLogicHierarchy;vtkSlicerMarkupsModuleMRMLHierarchy"
```
lassoan, line 55: What happens if a module is excluded from the build? Maybe check if the targets exist?

Modules/Loadable/CMakeLists.txt
```diff
@@ -40,3 +40,21 @@ foreach(module ${qtmodules})
     add_subdirectory(${module})
   endif()
 endforeach()
+
+# With the Makefiles generator, the Python wrapping hierarchy of a module can
```
lassoan: Is this a limitation of all generators or only Makefiles? Should we do this workaround only for that generator?

lassoan (commented):
It looks good to me overall, just added two small comments inline.

lassoan:
Remove the irrelevant `(cherry picked from commit 1dd81e00f2667acb771d3e412fce95fcd811a338)` comment.

One more thought: Everybody should use AI assistance in development (it would be irresponsible to not do it). For this reason, I would recommend to add the AI agent as a coauthor only when the human author does not fully understand, review, or test the change; or for any other reason does not take full responsibility for the proposed changes. I expect not everybody would agree with this - we can discuss at the next dev meeting.

Modules/Loadable/CMakeLists.txt
pieper: I don't know the answer to I'm making this PR a draft until we can decide how we want to deal with CMake issues going forward.

Modules/Loadable/CMakeLists.txt
pieper: I'm not sure how we would test that conveniently.  To me this looks like an issue that would happen with all generators, but without a convenient CI that builds every PR it's too much time and effort to check this kind of change manually.

jamesobutler: > Transforms and Segmentations are configured before Markups, so their hierarchy rules silently lost the dependency on the Markups MRML hierarchy and parallel builds could fail with:
> 
> vtkWrapHierarchy: couldn't open file .../vtkSlicerMarkupsModuleMRMLHierarchy.txt

With my parallel builds on Windows and macOS that were clean builds I never ran into this issue. On Windows I have been enabling the `CMAKE_CXX_MP_FLAG` and on macOS with `cmake --build ./S5R -j 10`

pieper:
This PR and https://github.com/Slicer/Slicer/pull/9293 are good ones to evaluate in terms of our AI assistance policy.  My vote would be to always retain the AI co-author tag; if as you say it's irresponsible not not have the AI review then we should be skeptical of any commits that weren't marked as co-authored by an AI!

Rather than discuss in the abstract though, I really think we need to decide what we want to be able to say functionally about any commits to the core in order to accept then.

Candidate requirements would include:
- [ ] was initiated in response to a known issue or feature request
- [ ] builds from source on all platforms with no new warnings or errors
- [ ] passes all tests (core and extension) that were passing without the change
- [ ] includes specific tests for the issues it is intended to fix
- [ ] was visually reviewed by the submitter and by the reviewer
- [ ] has corresponding documentation changes
- [ ] passes review by copilot and any other checks we want to apply

We have always aspired to this level of engineering correctness, but some of these were historically impractical to do manually or autmate, but are more practical with AI help.  Others may still be challenging due to CI build/test time limits, but if that's still the case we should add resources or refactor the code to make the automated tests possible.

Modules/Loadable/CMakeLists.txt
lassoan: It seems to be an issue on Windows, too. `vtkSlicerSegmentationsModuleLogicHierarchy` depends on `vtkSlicerTerminologiesModuleLogicHierarchy` and `vtkSlicerMarkupsModuleMRMLHierarchy` but only `vtkSlicerTerminologiesModuleLogicHierarchy` shows up as a dependency in the generated Visual Studio project file:

Image

Image

I haven't encountered any issues, but I don't use parallel builds because on Windows the speed improvement is not significant (as I remember, maybe few ten percent faster) and it is not very robust (sometimes fails).

So, maybe it is enough to just change the comment to not state that it is only an issue for the Makefiles generator and linux.

lassoan:
According to Claude, based on the analysis of the Windows build tree and source code, it suggested that switching to VTK's module system would fix this:

> vtk_module_build() / vtk_module_wrap_python() compute hierarchy dependencies from declared module metadata (vtk.module files) independent of configure order, so this class of problem doesn't exist there. The vtkAddon macros are VTK8-era ports of KitCommonWrapBlock.cmake; this is the "real" long-term fix but a large migration touching every wrapped Slicer library, so it's a separate project rather than a fix for this issue.

We can use the current workaround until Slicer build switches to the VTK module system sometime in the future.

pieper:
Okay, I changed the comment.  I do hope that in the future we can drastically simplify our whole build process.

pieper:
This one shouldn't be needed once  https://github.com/Slicer/vtkAddon/pull/71 and https://github.com/Slicer/Slicer/pull/9326 are in.

Issue #5945, opened 2021-10-12
Slicer-4.13 hangs at startup on Windows11


lassoan:
Slicer does not start up (hangs, then the operating system terminates it) on Windows11

See discussion here:
https://discourse.slicer.org/t/slicer-on-windows-11/20085/11

## Environment
- Slicer version: Slicer 4.13.0-2021-10-08
- Operating system: Windows 11 Pro, Version: 21H2, OS build: 22000.194

jamesobutler:
I'm glad that the libopenblas information was a helpful hint. Thanks for tracking down a workaround! I'll keep an eye on Windows 11 compatibility issues and if this workaround becomes no longer needed.

Issue #9203, opened 2026-06-02
Improve startup speed


lassoan:
Slicer can take tens of seconds to start. We need to find a way to make the application window to appear more quickly.
More details are described in this issue: https://github.com/Slicer/Slicer/issues/8638

PR #9306, opened 2026-07-22
PERF: Load VTK Python modules lazily and skip startup scipy import

pieper:
The 'vtk.py' shim distributed with VTK eagerly star-imports more than a hundred binary modules, costing ~0.9s at every application startup even though a typical session touches only a fraction of them. Python initialization dominated the fixed part of application startup (~80% of main-thread samples during the core startup phase).

Generate a lazily-loading drop-in replacement at build time (GenerateLazyVtkModule.py): every VTK module is imported once during the build to record which module provides each public attribute, and the generated 'vtk' module resolves attributes on demand through a module __getattr__, importing only the vtkmodules submodule that provides the requested name. The module is placed in bin/Python, which precedes the VTK site-packages directory on the Python path. Submodule imports such as 'import vtk.util.numpy_support' keep working through the same __path__ aliasing used by the eager shim, and 'import vtkmodules.all' remains available to force everything to load.

The VTK modules that provide base classes of the wrapped MRML and Slicer classes are imported explicitly in slicer/__init__.py before the kits, because wrapped classes link their Python bases only if the providing module is already initialized (verified: without this, isinstance(slicer.vtkMRMLScene(), vtk.vtkObject) is False in PythonSlicer). These ~28 core modules cost ~175ms of the original ~810ms; the remaining hundred-plus modules load on first use.

Also restrict the eager numpy/scipy import workaround for issue #5945 to Windows, the only platform that needs it: on macOS and Linux it cost several hundred milliseconds of every startup.

Measured on an M-series MacBook: warm application startup dropped from 5.9s to 3.9s, with 35 of 125 VTK Python modules loaded after startup.


pieper:
Not sure if this is the right long-term solution but it seems to do the trick.

lassoan:
There is no problem with warm startup time. Start in 4s instead of 6s is nice, but 6s is perfectly acceptable already.

The issue is that cold startup can be 30-40 seconds. Have you measured if these changes improve cold startup times? (restart your computer, wait 1 minute after the desktop appears, start Slicer)

jamesobutler:
xref #9203

pieper:
> There is no problem with warm startup time. Start in 4s instead of 6s is nice, but 6s is perfectly acceptable already.
> 

I'm definitely of the opinion that every second counts - users love apps that start fast so I'm trying to squeeze out everything we can.

> The issue is that cold startup can be 30-40 seconds. Have you measured if these changes improve cold startup times? (restart your computer, wait 1 minute after the desktop appears, start Slicer)

Yes, I've been exploring various ways to address the cold start issue - nothing perfect yet, but at least on mac I think we will end up treating it as a kind of "install" step, where on first launch of a new version it runs the various OS-level inspection tasks with a progress screen.  That should be enough for the system to recognize the libs and speed up subsequent starts even after a restart.

jamesobutler:
I locally built Slicer `main` branch on the Windows platform and packaged. Upon running the installer and selecting to launch the application from the last page of the installer I observed it took approximately ~34 seconds to fully load (3 seconds for the first app launcher splash screen to appear, then another 20 seconds until "Registering modules" displayed on the splash screen and then followed by another 11 seconds for all the modules to load and the main window to appear. Subsequent reloads of the application only took approximately 7 seconds.

With the change in this PR I observed it took approximately ~30 seconds to fully load (3 seconds for the first app launcher splash screen to appear, then another 16 seconds until "Registering modules" displayed on the splash screen and then followed by another 11 seconds for all the modules to load and the main window to appear.

^This may possibly suggest that the step prior to "Registering modules" has been reduced from 20 seconds to 16 seconds on a cold start due to the changes in this PR.

pieper:
> ^This may possibly suggest that the step prior to "Registering modules" has been reduced from 20 seconds to 16 seconds on a cold start due to the changes in this PR.

Yes, 4 seconds shaved off a cold import of VTK could be a real thing on windows.

What @lassoan mentioned today about tracking dll load events and displaying those updates in the app launcher splash screen could make a lot of sense just to improve the user experience if not the actual start up time.

pieper:
I think this is worth doing unless anyone objects.

jamesobutler (approved):

lassoan:
I did a lot of experiments on Windows.

I only investigated cold startup time - starting Slicer after the computer was restarted - as restarts takes 5-7 seconds, so there is not much to improve there.

One important finding is that all the slowdown is due to Microsoft Defender malware scan (MsMpEng.exe):
- Slicer startup with malware scan enabled (default system setting): 28sec
- Slicer startup with malware scan disabled: 7sec

We need to optimize the startup time with malware scan enabled, because on most systems it will not be an option to disable it.

Startup time optimization results:
- Current Slicer main branch: **28sec**
- With VTK lazy loading (implemented in this PR) and disabling of scipy import on Windows as well (does not hang anymore): **20.7sec** - enormous improvement!
- After reorganizing Python scripted modules to not import numpy and pydicom at startup: **13.6 sec** - significant improvement (all .py files changes, mostly in Slicer, but two small changes in VTK .py files were needed to prevent unnecessary `import numpy`)

I've implemented parallel loading of DLLs, which makes the malware scan done in parallel. This parallel loading takes about 0.5sec and it makes the loading 1.5sec faster, so it may worth considering, as it doesn't add much complication.

There may be an additional 1-2 seconds that could be saved by moving Qt font database initialization to run in a background thread. This is not complicated at all either, but it might not be necessary with Qt6.

Overall, with all the mentioned optimizations, the cold startup time is about **11sec** on Windows.

lassoan:
This PR is worth to be merged as is and I'll submit additional improvements as separate PRs.

cpinter:
Thank you @pieper for this change and the others for the testing!

jamesobutler:
@pieper One test is failing now because of the change. It is the LandmarkRegistration test from the remote extension bundled into Slicer core.
https://slicer.cdash.org/tests/33793920
```
======================================================================
ERROR: test_LandmarkRegistrationThinPlate (LandmarkRegistration.LandmarkRegistrationTest.test_LandmarkRegistrationThinPlate)
Test the thin plate spline transform
----------------------------------------------------------------------
  File "C:/D/P/S-0-build/Slicer-build/lib/Slicer-5.13/qt-scripted-modules/LandmarkRegistration.py", line 1179, in test_LandmarkRegistrationThinPlate
    w.currentRegistrationInterface.onExportGrid()
  File "C:\D\P\S-0-build\Slicer-build\lib\Slicer-5.13\qt-scripted-modules\RegistrationLib\ThinPlatePlugin.py", line 114, in onExportGrid
    gridTransform.SetDisplacementGridData(toGrid.GetOutput())
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'vtkAddonPython.vtkOrientedGridTransform' object has no attribute 'SetDisplacementGridData'
```

pieper:
@jamesobutler thanks for finding the regression.  I believe Slicer/vtkAddon#71 and #9326 will be the general fix for this.

PR #9307, opened 2026-07-22
ENH: Improve Python package install progress dialog

lassoan:
Extensions often call `pip_install` in quick succession, which makes the dialog nearly impossible to interact with, as each call resets the position, collapsed state, and the the window keeps appearing and disappearing.

This commit changes the pip install progress dialog (_PipProgressDialog) so that installs performed in quick succession appear in a single, stable window:
- Reuse one dialog instance with a deferred hide, so back-to-back installs no longer flash the window closed and open between calls.
- Show the log as one continuous stream across consecutive installs, with a divider between them. Clear it only after the dialog has been idle for a while.
- Preserve the dialog position, size, and scroll position, scroll state, details section collapsed state between installs.

Other minor improvements:
- Auto-scroll: when the end of the log is visible then keep showing the last log lines. Stop auto-scroll when the user scrolls up and resume when return to the bottom.
- Restore popup size when collapsing details section.

Before:

https://github.com/user-attachments/assets/b1690563-66ca-4b03-ae80-ac32111dea3c

After:

https://github.com/user-attachments/assets/87c8cc33-4f39-48b5-9650-6e2635f0deb6

Tested using this script:

```
import slicer, qt

# Obscure, small, pure-Python packages — none are Slicer/pip dependencies,
# so reinstalling them is safe.
packages = [
    "tabulate", "termcolor", "humanize", "more-itertools", "cachetools",
    "decorator", "toml", "inflection", "shortuuid", "emoji",
    "pyfiglet", "cowsay", "parse", "isodate", "natsort",
]

for i, pkg in enumerate(packages):
    slicer.util.pip_uninstall(pkg)

for i, pkg in enumerate(packages):
    slicer.util.pip_install(pkg)
```

### DIFF
--- a/Base/Python/slicer/packaging.py
+++ b/Base/Python/slicer/packaging.py
@@ -751,10 +751,9 @@ def _pip_install_with_dialog(
     import qt
     import slicer
 
-    # Create the progress dialog
-    dialog = _PipProgressDialog(requester=requester, parent=parent)
-    dialog.show()
-    slicer.app.processEvents()  # Ensure dialog is displayed
+    # Acquire the shared progress dialog. It is reused across back-to-back installs so
+    # the window does not flash between them.
+    dialog = _PipProgressDialog.acquire(requester=requester, parent=parent)
 
     completed = threading.Event()
     result = {"returnCode": None, "log": ""}
@@ -779,15 +778,19 @@ def _pip_install_with_dialog(
         slicer.app.processEvents()
         qt.QThread.msleep(10)  # Reduce CPU usage
 
-    dialog.close()
-
     if result["returnCode"] != 0:
+        # On failure, hide immediately so the error dialog is not shown over a lingering
+        # progress dialog.
+        dialog.hideNow()
         slicer.util.errorDisplay(
             _("Package installation failed."),
             windowTitle=_("{requester} - Installation Failed").format(requester=requester) if requester else _("Installation Failed"),
             detailedText=result["log"],
         )
         raise CalledProcessError(result["returnCode"], "pip install")
+
+    # Keep the dialog visible briefly so a following install can reuse it (no flash).
+    dialog.release()
 
 
 def _pip_install_with_statusbar(
@@ -906,18 +909,17 @@ def _pip_install_with_skips_dialog(
     parent: qt.QWidget | None = None,
 ) -> list[str]:
     """Recursive skip-packages install with modal progress dialog."""
-    import slicer
-
-    dialog = _PipProgressDialog(requester=requester, parent=parent)
-    dialog.show()
-    slicer.app.processEvents()
+    # Acquire the shared progress dialog (reused across back-to-back installs so the
+    # window does not flash between them).
+    dialog = _PipProgressDialog.acquire(requester=requester, parent=parent)
 
     try:
         skipped = _pip_install_with_skips(
             requirements, skip_packages, constraints, log_fn=dialog.appendLog,
         )
     finally:
-        dialog.close()
+        # Keep the dialog visible briefly so a following install can reuse it (no flash).
+        dialog.release()
 
     return skipped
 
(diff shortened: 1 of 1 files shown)


Base/Python/slicer/packaging.py
```diff
@@ -1187,19 +1189,57 @@ class _PipProgressDialog:
...
+        """Return a visible progress dialog ready for a new install.
+
+        Reuses the shared dialog if it is still on screen -- avoiding a flash between
+        back-to-back installs -- otherwise creates it (or re-shows the hidden one).
+        """
+        # Cancel any pending hide so a quickly following install reuses this dialog.
+        if cls._hideTimer is not None:
+            cls._hideTimer.stop()
+
+        inst = cls._sharedInstance
```
ebrahimebrahim, line 1222: The dialog gets destroyed by Qt when its parent gets destroyed, but `acquire` could still end up retrieving the `_PipProgressDialog` with the destroyed `QDialog` because it is stored in `_sharedInstance`.  Maybe we could force `parent` to always be the main window, or maybe we could keep `parent` flexible but connect the parent destroy signal to a clearing of `_sharedInstance`.

Base/Python/slicer/packaging.py
```diff
@@ -1187,19 +1189,57 @@ class _PipProgressDialog:
...
+    # and recreating one per install -- avoids the window flashing between back-to-back
+    # installs, keeps its position/size/scroll, and makes the log read as one stream.
+    _sharedInstance = None
+    # Single-shot timer that hides the shared dialog a short while after an install
+    # finishes. A new install arriving before it fires cancels it and reuses the still-
+    # visible dialog, so nothing flashes. If it does fire, the log is cleared so the
+    # next install starts fresh.
+    _hideTimer = None
+    # How long the dialog lingers, visible, after an install finishes (seconds).
+    _hideDelaySeconds = 2
```
ebrahimebrahim: 2 seconds felt a bit long to me, for example when installing something that is already installed and having to wait for no reason. Because it is a modal dialog it Slicer is nonresponsive for no good reason during those 2 seconds. Maybe we can reduce it a bit and still get the benefit? Or maybe we can make it non-modal upon release, and modal again at the start of the next install (idk if that would cause a different sort of flashing effect though...)

ebrahimebrahim (changes requested):
This is an excellent improvement!

Added a couple of comments

ebrahimebrahim:
Also I tested this on linux and it worked well. Modal dialog behavior is a bit different on linux: dragging the window around drags everything (including the main window), so a modal dialog always appears int he same place relative to the main window. But the flashing issue with multiple successive calls to pip_install was definitly a thing

Base/Python/slicer/packaging.py
lassoan: Making the main window the parent could be a good idea (if it exists).

Base/Python/slicer/packaging.py
lassoan: I have not noticed the delay, but we can certainly reduce it to a very small number (0 would work well most cases, but just in case there is some processing in the extension and processEvents is called, using something like 0.5s would be a bit safer). Can you test if 0.5 works well for you?

ebrahimebrahim: 0.5 works well for me with your test snippet, and it is indeed far less annoying when running on something that is already installed.

Base/Python/slicer/packaging.py
lassoan: Done.

Base/Python/slicer/packaging.py
lassoan: Set to 0.5s.

lassoan:
I've addressed all comments, please check again.

Base/Python/slicer/packaging.py
ebrahimebrahim: I guess there is still the same issue here if you launch slicer with --no-main-window

Qt is still avaialble so pip_install still makes dialogs even if you do --no-main-window

So right now if someone uses `pip_install` in a script that is run with `--no-main-window` and they supply a temporary `parent`, then when the `parent` is destroyed, so is the QDialog and you end up with  a broken `_PipProgressDialog`

ebrahimebrahim: Simple solution: `qt.QDialog(slicer.util.mainWindow())` instead of `qt.QDialog(slicer.util.mainWindow() or parent)`. That way a null parent is provided in the `--no-main-window` case and things are okay

ebrahimebrahim (changes requested):

Base/Python/slicer/packaging.py
lassoan: `parent` is None by default, so `qt.QDialog(slicer.util.mainWindow() or parent)` equals `qt.QDialog(slicer.util.mainWindow())`.

ebrahimebrahim: But if someone provides a non-None parent and is --no-main-waindow?

ebrahimebrahim: Would like to merge this! The flashing annoys me too :smile: 

I think this PR looks good and this is the only remaining sticking point. It should just be `qt.QDialog(slicer.util.mainWindow())` I think, removing the option to provide a custom parent. Otherwise it looks like you can always provide a parent when in fact doing so will break things in the --no-main-window scenario.

lassoan:
Thanks for the review. I've removed the `parent` member as suggested.

PR #9309, opened 2026-07-23
Fix quit/save dialog issues on exit

pieper:
## Overview

Fixes two problems in the application-exit confirmation flow in `qSlicerMainWindowPrivate::confirmCloseApplication()`. Both were observed on macOS with Qt6, but the first is a platform-independent logic fragility.

## 1. `BUG:` Save-before-exit dialog didn't honor the chosen button

When the scene has unsaved changes, quitting shows a dialog with **Save**, **Exit (discard modifications)**, and **Cancel exit**. The **Save** and **Exit** buttons were both added with `QMessageBox::ActionRole`, and the result was read via `clickedButton()` pointer identity with no default/escape button set. With two same-role buttons the native macOS/Qt6 button layout could return the wrong button, so clicking **Exit** was handled as **Save** and re-opened the Save Data dialog instead of quitting.

- Give each button a distinct role: `AcceptRole` / `DestructiveRole` / `RejectRole`.
- Set the non-destructive **Cancel exit** as the default and escape button.
- Read `clickedButton()` once.
- Also fixes the long-standing `\todo`: choosing **Save** now exits only if the save data dialog was actually completed, instead of a fire-and-forget action trigger that never exited after saving.

## 2. `STYLE:` Suppressed exit dialog flashed on screen

When there are no unsaved changes and the user previously ticked *"Don't show this message again"*, `ctkMessageBox::confirmExit()` still constructs the dialog and suppresses it by briefly realizing it off-screen and auto-clicking on the next event-loop turn. Because `confirmExit()` runs synchronously from `closeEvent` (unlike `disclaimer()`, which defers `exec()` via `singleShot`), the off-screen suppression loses the race with the window server on macOS/Qt6 and the dialog visibly flashes before the app quits.

- Read the `MainWindow/DontConfirmExit` setting directly and skip constructing the dialog when confirmation has been disabled, so exit is immediate and no dialog is ever realized.

## Testing

Built and tested on macOS (Qt6). With a modified scene: **Exit** quits immediately, **Save** saves then quits (or stays open if the save dialog is cancelled), **Cancel exit**/Esc/Return keep the app open. With no unsaved changes and confirmation disabled: Command-Q quits with no flash.

### DIFF
--- a/Base/QTApp/qSlicerMainWindow.cxx
+++ b/Base/QTApp/qSlicerMainWindow.cxx
@@ -606,7 +606,25 @@ bool qSlicerMainWindowPrivate::confirmCloseApplication()
   }
   else
   {
-    close = ctkMessageBox::confirmExit("MainWindow/DontConfirmExit", q);
+    // If the user previously ticked "Don't show this message again", skip the
+    // confirmation dialog entirely instead of letting ctkMessageBox auto-accept
+    // it. ctkMessageBox suppresses a dialog by briefly realizing it off-screen
+    // and clicking the accept button on the next event-loop turn; when that
+    // happens synchronously from closeEvent (as opposed to the deferred
+    // singleShot(exec()) path used by disclaimer()), the off-screen suppression
+    // loses the race with the window server on macOS/Qt6 and the dialog flashes
+    // on screen. Reading the setting directly avoids constructing the dialog.
+    static const QString dontConfirmExitKey = "MainWindow/DontConfirmExit";
+    QSettings settings;
+    bool confirmationDisabled = settings.value(dontConfirmExitKey, static_cast<int>(QMessageBox::InvalidRole)).toInt() != QMessageBox::InvalidRole;
+    if (confirmationDisabled)
+    {
+      close = true;
+    }
+    else
+    {
+      close = ctkMessageBox::confirmExit(dontConfirmExitKey, q);
+    }
   }
   return close;
 }
--- a/Base/QTApp/qSlicerMainWindow.cxx
+++ b/Base/QTApp/qSlicerMainWindow.cxx
@@ -570,22 +570,35 @@ bool qSlicerMainWindowPrivate::confirmCloseApplication()
                                               qSlicerMainWindow::tr("The scene has been modified. Do you want to save it before exit?"),
                                               QMessageBox::NoButton,
                                               q);
-    QAbstractButton* saveButton = messageBox->addButton(qSlicerMainWindow::tr("Save"), QMessageBox::ActionRole);
-    QAbstractButton* exitButton = messageBox->addButton(qSlicerMainWindow::tr("Exit (discard modifications)"), QMessageBox::ActionRole);
-    messageBox->addButton(qSlicerMainWindow::tr("Cancel exit"), QMessageBox::RejectRole);
+    // Give each button a distinct role. Previously "Save" and
+    // "Exit (discard modifications)" were both added with ActionRole; because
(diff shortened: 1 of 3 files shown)


lassoan:
These will be great!

Have you checked the behavior for failed scene saving or saving with warning? I'll test on windows, but it seems that there may be OS-specific behavior and we should make sure saving logic is solid.

Since this is important, the best would be to have a test for it. Could you try asking Claude to write a test for this?

pieper:
A test is a good idea, I can get Claude to work on that.

I did manually test that the program does quit after a successful save, and that if you cancel the save the quit is also canceled.  I'll doublecheck that a failed save also cancels the quit.

pieper:
Claude says:

Added a test for this: `qSlicerSaveDataDialogSaveOutcomeTest` (Base/QTGUI/Testing/Cxx).

It's a black-box test of the exact contract the exit path relies on: `confirmCloseApplication()` quits only if `openSaveDataDialog()` returns `true`, i.e. the Save Data dialog's accepted/rejected result. The test registers a configurable file writer for a transform node and drives the **real modal dialog** (located via `QApplication::activeModalWidget()`, Save button clicked as a user would — no test-only hooks), asserting:

- **successful save** → dialog accepts → exit allowed;
- **failed save** → dialog does not accept → exit cancelled;
- **save with a warning** → dialog does not accept → exit cancelled (so the user can review).

On the failed/warning question you raised: tracing it through, the behavior is already correct by construction and is **not** OS-specific — in `qSlicerSaveDataDialogPrivate::saveNodes()`, any user message on the storage node (warning *or* error) sets `doneWithSaveDataDialog = false`, so `save()` returns false and `accept()` never closes the dialog. So a failed or warning save keeps the app open regardless of platform. No production change was needed for that; the test pins the behavior so it can't regress.

(One incidental finding while writing it: `qSlicerFileWriter::canWriteObject()` defaults to `false`, so a custom writer must override it to appear in the dialog — noted in the test.)

Passes locally (100%). 🤖 with Claude Code

pieper:
> there may be OS-specific behavior

The issue was that the mac style guidelines move the buttons around and we weren't taking this into account.  That must have changed with Qt6 somehow, but I think the current code is going to be robust for this.

lassoan (approved):
Checking the "dontshowagain" setting value outside of the dialog class is a simple and clean solution for this specific case, but it would need to be repeated everywhere this feature is used. It would ne nice to have a solution that makes the existing code work as is. We would probably need to change the ctkPopupWindow class not to be derived from a Qt popup but intead embed a Qt popup.

Since this dialog is used at every application exit, it could make sense to integrate this PR quickly as is. But if you have time to explore then maybe you could give it a try to implement the final solution that fixes the popup everywhere.

pieper:
Thanks for the review!

I did look into it more and the ctkMessageBox does have the logic to avoid showing base on the settings value, but it also actually invokes the recorded action (e.g. either okay or cancel depending on what the user said to remember) and since it's handled well with the event loop the dialog never shows.  The issue is that for this pathway the message box is being invoked from in another dialog handler so path in ctk ends up on screen instead of off.

So I think we are on a reasonable path with this change.

jamesobutler:
@pieper Likely relevant test case (`py_nomainwindow_SlicerOptionDisableSettingsTest`) that failed last night with these changes:
Windows: https://slicer.cdash.org/tests/33904833
Linux: https://slicer.cdash.org/tests/33903444

pieper:
Thanks @jamesobutler - I'm having a look.

Issue #8638, opened 2025-08-11
Improve Slicer startup experience


jamesobutler:
## Is your feature request related to a problem? Please describe.

It has been described to me from new users of Slicer and derivative Slicer custom apps that the launch of the application seems slow compared to their expectation. This is a poor first impression of the app.

These observations were made using the Windows version of Slicer.

One main component of the startup experience is that the Launcher splash screen appears, then closes, then nothing appears to have for a ~10 seconds or so, and then the main application splash screen appears with the details about initializing and loading modules. During the period between the launcher splash screen and the main application splash screen, the user is confused about what is happening and thinks that something went wrong. This causes them to try to relaunch the application which ultimately produces a 2nd instance.

The launcher shows its splash screen and then hides it based on the time provided at:
https://github.com/Slicer/Slicer/blob/2b9e96c1d0c4fd27ef33f8974ebff90b80731c8c/CMake/SlicerMacroBuildApplication.cmake#L704

It is observed that subsequent launches of the application are faster. However, a bad first launch experience after application installation leaves a bad first impression.

It is unclear if python package importing is included in the main application splash screen loading time or if that happens before the main application splash screen appears. I suspect that the import of the full vtk package as was done in https://github.com/Slicer/Slicer/commit/ca4c71897af242b59d711fd59e0b3ab9df04aa74 is a main contributing factor to the startup time. On a system python I have observed that installing the vtk python package fresh and then importing it results in a import taking ~14 seconds to complete. Subsequent launches of the system python and importing of vtk does not seem to take as long. Additional imports of `numpy` and `scipy` exist as done in https://github.com/Slicer/Slicer/commit/542f53c5ba9bf650119707f990e36d74a6c8d36a and https://github.com/Slicer/Slicer/commit/ad5d20d1a1c14356bb55ab6b4ea50af133c00f57, but I have observed their import time is not nearly as long as vtk.

System Specs: HP ZBook G8 Mobile Workstation (i7-11850H, 32GB RAM, 1TB Storage, Windows 11 Enterprise 23H2)

Startup experience (%AppData%/slicer.org was cleared out of any previous user settings) and using Slicer 5.9.0-2025-08-08:
Timestamps of the video below
- 0:05 Installer is closed with instructions to automatically launch the application
- 0:11 - 0:14 Launcher splash screen appears and then hides
- 0:14 - 0:29 Nothing appears to happen (This user experience is what should be improved)
[...]

PR #9311, opened 2026-07-24
ENH: Pre-warm the macOS first-launch library scan behind a progress splash

pieper:
## What this is

The first time each newly built or freshly downloaded shared library is loaded, the operating system inspects it and caches the verdict per file — on macOS a `syspolicyd` / **XProtect** security scan (~0.13 s per Mach-O, measured), on Windows an antivirus scan, cold filesystem caches everywhere. Slicer ships **~900 libraries**, so the first launch of a new build or a freshly downloaded copy can spend **a minute or two** on this at near-zero CPU, showing a blank Dock icon and no window, before startup even really begins. Later launches take seconds.

This adds a **pre-warm** that performs that inspection up front, behind an optional native progress splash, so the wait is visible and explained instead of a frozen/absent window.

### Honest scope: this is a first-launch **experience** fix, not a speedup

The security scan is **serialized by the system daemon** (single-threaded system-wide — verified: an 8-worker parallel version gave *no* speedup and was removed). So the pre-warm **does not make the total launch faster** — the cost is unavoidable. What it buys:

- the unavoidable wait becomes a **native progress window with an accurate ETA** ("Preloading 3D Slicer libraries… about a minute left") instead of a blank screen, and
- a **stamp file limits the work to when it's actually needed**: after a new build or download, and once per session after a reboot (which clears the OS caches). The splash is only shown when the run is predicted to exceed **2 seconds**, so ordinary launches stay silent.

## How it works

- **macOS primitive:** each library is mapped with `PROT_EXEC`, which makes the kernel run the *identical* scan and cache it — without `dlopen`, dependency resolution, or spawning a process. (Other platforms load in a disposable worker.)
- **`SlicerPrewarm.py`** — the sequential warm-up, a boot-session-aware stamp, and native progress splashes for macOS (AppKit), Windows (user32), and X11, all drawn through `ctypes` with **no GUI-toolkit dependency**.
- Three entry points: an automatic **`prewarm` build target** (freshly built libs at end of build), the **macOS bundle bootstrap** (a small compiled `CFBundleExecutable` shim that pre-warms before `exec`-ing the real app, preserving the process identity), and an **in-app hook** for extension / additional-module paths known only at runtime.
- Disable with `SLICER_DISABLE_STARTUP_PREWARM=1`.

## Relationship to prior discussion

Addresses the first-launch experience described in **#9203** (Improve startup speed) and **#8638** (Improve Slicer startup experience). Concrete points from that discussion this acts on:

- **macOS has no launcher** (and therefore no launcher splash) — so the launcher-splash handoff fix (#8808, Win/Linux) doesn't cover the macOS cold first launch. The bundle bootstrap here is the macOS analog: it puts *something* on screen during the wait.
- The reported failure mode — *"nothing appears for ~10 s, the user thinks it failed and relaunches, producing a 2nd instance"* — is directly what the progress splash is meant to prevent on first launch. (This PR does **not** add single-instance locking; that's a separate mitigation.)
- Per the discussion, this intentionally **does not** add an XML launcher↔app progress protocol — it uses a self-contained native splash instead.

Complementary to **#9306** (lazy VTK Python modules), which reduces the *actual* Python-import work at every startup. That PR is a genuine `PERF` speedup; this one is first-launch transparency for a cost that can't be reduced — the two are independent (no shared files) and address different parts of the same "startup feels slow" problem, which is why they're separate PRs.

## Status / notes

- Draft. Primarily exercised on macOS arm64; the Windows/X11 splashes are implemented but need testing on those platforms.
- The pre-warm is a no-op on a normal warm launch (stamp check only) and never blocks or aborts startup — any failure (unwritable stamp, splash init failure, etc.) is swallowed.
- Not addressed here (possible follow-ups from the same discussion): skipping instantiation of testing modules when developer mode is off, and preventing a second instance during a slow first launch.

### DIFF
--- a/CMake/SlicerAppBootstrap.c.in
+++ b/CMake/SlicerAppBootstrap.c.in
@@ -1,0 +1,191 @@
+/*==============================================================================
+
+  Program: 3D Slicer
+
+  Copyright (c) Kitware Inc.
+
+  See COPYRIGHT.txt
+  or http://www.slicer.org/copyright/copyright.txt for details.
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+
+==============================================================================*/
+
+/* @SLICERAPP_APPLICATION_NAME@ macOS application bundle bootstrap.
+ *
+ * macOS scans every newly built or newly downloaded library the first time it
+ * is loaded (a syspolicyd / XProtect security evaluation) and caches the
+ * verdict per file. Slicer loads ~900 libraries, so the first launch of a new
+ * build or a freshly downloaded copy can spend a couple of minutes on this
+ * before the first window appears, with no feedback. This program runs as the
+ * bundle executable (CFBundleExecutable) before the application binary: it
+ * pre-triggers that evaluation up front while a native splash screen shows
+ * progress (see SlicerPrewarm.py), then replaces itself with the real
+ * application binary so the process keeps the identity expected by Finder and
+ * the Dock. The evaluation is serialized by the system, so this does not make
+ * the launch faster -- it makes the unavoidable wait visible and explained
+ * (and, via a stamp file, limited to when it is actually needed).
+ *
+ * This must be a compiled program rather than a shell script: system shells
+ * are restricted binaries, so dyld would strip the DYLD_* environment
+ * variables set by the application launcher in a build tree.
+ *
+ * Set SLICER_DISABLE_STARTUP_PREWARM=1 to skip the prewarm.
+ */
+
+#include <limits.h>
+#include <mach-o/dyld.h>
+#include <spawn.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/stat.h>
+#include <sys/wait.h>
+#include <unistd.h>
+
+extern char** environ;
+
+static int fileExists(const char* path)
+{
+  return access(path, F_OK) == 0;
+}
+
+static int executableExists(const char* path)
+{
+  return access(path, X_OK) == 0;
+}
+
+int main(int argc, char* argv[])
+{
+  /* The splash is shown by default; some arguments imply that no splash is
+   * wanted, and informational arguments should print without any delay. */
+  int showSplash = 1;
+  int runPrewarm = 1;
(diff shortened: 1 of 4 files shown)


jamesobutler (commented):
I locally built Slicer `main` branch on the Windows platform and packaged. Upon running the installer and selecting to launch the application from the last page of the installer I observed it took approximately ~34 seconds to fully load (3 seconds for the first app launcher splash screen to appear, then another 20 seconds until "Registering modules" displayed on the splash screen and then followed by another 11 seconds for all the modules to load and the main window to appear. Subsequent reloads of the application only took approximately 7 seconds.

When I picked this commit locally and built and packaged and re-ran the installer, I observed that it taking the same amount of time on first startup and same amount of time for subsequent starts. Therefore I have not observed an improvement on the Windows platform with these changes. It was also ~34 seconds for the application to load after restarting my computer and relaunching Slicer.

pieper:
Thanks for the testing James.  This may end up being a mac-specific feature.  So far it seems that the system scans each file sequentially and then saves a flag that survives reboot.  So we may not be able to speed things up but we may be able to provide helpful user feedback.  I'm going to do some more testing before finalizing.

jamesobutler:
Upon rebuilding with latest changes here, I'm actually not observing the `WindowsSplash` object at all on Windows. I do see in the console window that is built with the default Slicer build options:

```
prewarm: 99 libraries to load
prewarm: done
Switch to module:  "Welcome"
```

pieper:
Thanks for testing that.  Based on what I'm finding it's the XProtect service on mac that is unavoidably slow, so that's where the splash screen with progress will make sense.  Preloading the shared libraries on the mac doesn't seem to help so I'm leaving that part out.  

But it may be worth doing on windows: having a multithreaded prewarm that reads all the shared libs into cache might be worth trying on windows.  Do you want to run that test?  For my mac with SSD that part was only 0.5 seconds while the XProtect scan on first run was minutes.

pieper:
As we discussed on the call today, I think this addresses a real user experience problem on the mac.

I'm not 100% convinced this is the best implementation, but I can't think of a better one.

@jamesobutler or @lassoan what do you think?

pieper:
Not time-critical, but this addresses a real user confusion condition: there's no indication that the app is trying to start for a long time.

lassoan (approved):

For reference, Windows startup acceleration (by reading libraries in parallel at every startup) was implemented in https://github.com/Slicer/Slicer/commit/9a81a83f77247f89e3fa7e833b5cab7516fc51a0

PR #9312, opened 2026-07-25
ENH: Remove support for building against VTK <= 9.4

jamesobutler:
This anticipates upcoming support for VTK 9.7.0 at which there will be support for VTK 9.5, 9.6 and 9.7.

This is a follow-up to https://github.com/Slicer/Slicer/commit/67a8c92f54f7972b1bca05673e26e765db11de50 where the VTK minimum version was last updated to 9.4.


lassoan (approved):

PR #9313, opened 2026-07-25
ENH: Allow compiling dcmqi and add additional hash

mauigna06:
If this is not done build fails on Nvidia's Jetson Orin Nano otherwise it builds fine. See https://discourse.slicer.org/t/build-slicer-on-nvidias-jetson-orin-nano


jamesobutler (commented):
Since we don’t generally add the source tar as an option (the [python_package_updater.py](https://github.com/Slicer/Slicer/blob/main/Utilities/Scripts/python_package_updater.py) doesn’t include them in the update when versions are updated), the better thing would be if you can update the CI over at https://github.com/ImagingDataCommons/dcmqi-python-distributions to add support for Linux arm64 whls which probably isn’t too much lift consider macOS arm compatible whls are already available. cc @fedorov who may also support Linux arm64 whls for the `dcmqi` Python package.

jamesobutler:
Closing this PR where `dcmqi` can be updated using https://github.com/Slicer/Slicer/blob/main/Utilities/Scripts/python_package_updater.py where latest version has an available `dcmqi-1.5.7-py3-none-manylinux_2_28_aarch64.whl` now as well.

PR #9320, opened 2026-07-28
BUG: Fix OpenXR rendering not resuming automatically after headset doff/don

Sunderlandkyl:
The git hash containing the previously implemented fix to address headset doff/don wasn't included in External_VTK.cmake.

$ git shortlog 6181bb1223bbc499a340a1644f5356e7e152c318..c9cd17a377cbc7e488329e882a35b70c7f37fe76 --no-merges Andras Lasso (1):
      BUG: Resume OpenXR rendering automatically after headset doff/don


lassoan (approved):
Thanks, good catch!

PR #9322, opened 2026-07-30
ENH: Speed up macOS packaging and make developer builds distributable

pieper:
## What this does

3D Slicer's macOS package is assembled by CPack and made relocatable by CMake's BundleUtilities `fixup_bundle()`, which has produced working, distributable disk images for years. This preserves that result while making three improvements:

**1. Much faster.** `fixup_bundle()` walks the dependency graph one binary at a time, calling `otool` repeatedly and copying/rewriting serially — minutes of wall clock on a bundle with ~1000 Mach-O files, and the dominant cost of `make package`. It is replaced by `Utilities/Scripts/SlicerBundleCloseDeps.py`, which walks the graph once (one `otool -l` per binary) and applies the independent install-name rewrites in parallel: the fixup drops from minutes to about half a minute. The disk image is also written as LZFSE (ULFO) rather than the CPack default (UDZO/zlib) — several times faster to compress and slightly smaller.

**2. Much simpler.** Copying each framework whole (so nested code such as `QtWebEngineCore.framework`'s bundled `Helpers/QtWebEngineProcess` comes along and is fixed up automatically) removes the per-file placement table and the hand-written WebEngine / Designer / mis-embedded-framework special cases — roughly 400 lines of the fixup script go away.

**3. Makes developer builds distributable.** Building against a Homebrew Qt on Apple Silicon — as a developer typically does, rather than the Qt the build farm builds itself — left a few references to the Homebrew tree: Qt plugins installed as symlinks, `QT_PLUGIN_PATH` not set before the `QApplication` is constructed, and standalone executables (PythonSlicer, the CLI modules) carrying only build-machine run-paths. The closure installs real plugin files, sets `QT_PLUGIN_PATH` first, embeds and rewrites every external reference — absolute, `@rpath`, or a bare install name — to `@rpath`, gives the standalone executables a bundle-relative run-path, and strips every run-path that pointed outside the bundle, so the result names no build-machine path.

The bundle is re-signed after the fixup (install_name_tool invalidates each signature it rewrites, and on Apple Silicon an invalid signature faults at load) — inside-out, ad-hoc by default (`--identity` for a Developer ID), in parallel.

## Validation

Verified on a clean arm64 macOS VM with **no Homebrew, no Qt, and no build tree**: the packaged application, `python-real`, and the DCMTK CLI executables all launch, and the bundle references no library outside itself across all ~1000 of its Mach-O files.

## Follow-ups (separate PRs)

- Extend the signer toward a notarizable release (hardened runtime + entitlements + notarize/staple). Ad-hoc signing is enough to load and test; testers of an unsigned build clear the quarantine attribute manually.

### DIFF
--- a/Base/QTApp/qSlicerApplicationHelper.cxx
+++ b/Base/QTApp/qSlicerApplicationHelper.cxx
@@ -42,6 +42,7 @@
 #include "qSlicerLoadableModuleFactory.h"
 #include "qSlicerModuleFactoryManager.h"
 #include "qSlicerModuleManager.h"
+#include "vtkSlicerConfigure.h"        // For Slicer_QtPlugins_DIR
 #include "vtkSlicerVersionConfigure.h" // For Slicer_MAIN_PROJECT_VERSION_FULL
 
 #ifdef Slicer_USE_PYTHONQT
@@ -51,6 +52,7 @@
 #include <vtkSystemInformation.h>
 
 // CTK includes
+#include <ctkAppLauncherEnvironment.h>
 #include <ctkMessageBox.h>
 #include <ctkProxyStyle.h>
 #ifdef Slicer_USE_PYTHONQT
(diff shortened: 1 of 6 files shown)


pieper:
Note that this was heavily AI developed, so I'm leaving this as a draft until we can talk it over.

But I did test it end to end and was able to make a distributable arm64 dev build with homebrew qt that anyone can test here:
https://drive.google.com/file/d/1eHcYpRpqs9ILhZBI4_fl9J-8RZOadiG_/view?usp=drive_link

I tested this using a vanilla tart VM and it seemed fine.

pieper:
This isn't critical but it's nice to have for development.  It's ready if anyone wants to have a look.

lassoan (approved):
I did not review the details, but overall replacing lots of CMake code by Python is good, faster operation is good, and ability to create a usable package with homebrew qt is very useful, too. If it leads to problems we can always revert and revise.

PR #9324, opened 2026-07-30
ENH: Make macOS arm64 packaging repeatable

fynxiu:
## Summary

- dereference Homebrew Qt tool and plugin symlinks during installation
- resolve Qt 6 framework and runtime dependencies without matching framework headers
- select the staged app without following the DMG Applications symlink
- add a reusable verify, architecture-audit, smoke-test, signing, DMG, and installation driver
- support safe side-by-side installation names

## Motivation

A native arm64 package built with Homebrew Qt could retain external dependencies, copy Qt umbrella headers in place of framework binaries, and select applications through the staging Applications symlink. The manual repair steps were also not reusable.

## Testing

- built Slicer 5.12.3 natively for arm64
- verified 1,034 packaged Mach-O files contain arm64
- verified all dependencies for 1,094 Mach-O files resolve inside the bundle
- passed the headless package smoke test (`SLICER_PACKAGE_VERSION=5.12.3-2026-07-31`)
- verified the ad-hoc code signature and generated DMG checksum
- passed `SlicerQtInstallSymlinkTest.cmake` and `SlicerGetPrerequisitesRPathTest.cmake`

The build used Homebrew Qt 6.11.1, so it is intended for local validation on the build host; broadly distributable packages should continue to use the project-built Qt documented by Slicer.


pieper:
Thanks for working on this @fynxiu 👍 

Coincidentally I was working on something similar: https://github.com/Slicer/Slicer/pull/9322

Can you compare and see if we should combine to come up with the best solution?

fynxiu:
Thanks, @pieper. I have now compared the two approaches at a high level.
#9324 was opened earlier than I intended, while I was still validating the arm64/Qt 6 build in a real project. Continued testing has already uncovered additional Qt 6 and macOS-specific issues. For example, Qt 6 uses a different pointing-device API and event-delivery path for the Mac trackpad, and the resulting fix belongs in VTK rather than in the Slicer packaging changes.
#9322 takes a broader, alternative approach to dependency closure and overlaps substantially with this PR. I therefore do not think merging #9324 as-is or combining all the changes immediately would be the safest next step.
I would like to close this draft for now, keep the branch as a reference, and continue validating the arm64 build through real project use. Once the findings are reproducible and tested, I will bring them back as focused contributions: Slicer packaging changes coordinated with #9322, and Qt/VTK input fixes submitted separately to the appropriate repository. This should avoid competing packaging implementations and make each change easier to review.
Thanks for pointing me to #9322.

PR #9326, opened 2026-07-31
PERF: Drop hardcoded VTK base-class imports; rely on pruned wrapping DEPENDS

pieper:
Follow-on to #9306 (lazy `vtk` shim). The shim needs each wrapped subclass's VTK
base class imported before the subclass's Python type is built. That was handled
by importing a hand-maintained list of ~28 core `vtkmodules` explicitly in
`slicer/__init__.py`, because at the time the generated wrapping `DEPENDS`
dropped every VTK dependency.

That list is no longer needed. With the corresponding vtkAddon change, each
wrapped module's generated init imports exactly the VTK modules that provide its
direct base classes (the `DEPENDS`, pruned from the full link line to the
base-class providers), so subclasses resolve their bases automatically as the
kits load.

Removing the hardcoded list is strictly better: it is derived from the actual
class hierarchy rather than hand-maintained (and so covers extension modules
that define their own wrapped classes), and it no longer force-loads heavy
modules (`vtkRenderingOpenGL2`, `vtkRenderingVolumeOpenGL2`, `vtkChartsCore`)
that are not base-class providers at startup.

Measured on macOS: 29 of 125 VTK Python modules loaded after a full GUI startup
(none of the rendering/IO/Charts tail), `isinstance(slicer.vtkMRMLScene(),
vtk.vtkObject)` is True, and warm startup drops about 1s (5.9s → 4.9s).

**Depends on** the vtkAddon DEPENDS-pruning change (Slicer/vtkAddon#71). Before
merge this PR must also bump the vtkAddon `GIT_TAG` in `SuperBuild.cmake` to the
merged vtkAddon commit; without the pruned vtkAddon the wrapped modules still
list every VTK module and startup gains nothing. Opened as a draft until then.

### DIFF
--- a/Base/Python/slicer/__init__.py
+++ b/Base/Python/slicer/__init__.py
@@ -166,45 +166,19 @@ For more details, see the generated Slicer API documentation.
 #   object = slicer.vtkMRMLScene()
 #   assert isinstance(object, vtk.vtkObject)
 #
-# This is needed because the wrapping of the MRML or Slicer VTK-based C++ classes is done
-# through the vtkMacroKitPythonWrap (from vtkAddon) and the generated "vtk*PythonInitImpl.cxx"
-# files do not include the names of the dependent VTK modules in the list used with
-# "vtkPythonUtil::ImportModule".
+# The ``vtk`` module is a lazily-loading shim (see GenerateLazyVtkModule.py)
+# that imports each vtkmodules submodule only when one of its attributes is
+# first accessed, instead of eagerly star-importing all ~125 modules.
 #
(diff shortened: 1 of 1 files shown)


pieper:
this requires the vtkAddon fix.  This should fix a regression that James noticed in a test failure.

lassoan (approved):

jamesobutler:
@pieper This will need a corresponding vtkAddon commit hash bump
https://github.com/Slicer/Slicer/blob/054379095d6dcc52e8211102c1c623b267184a0f/SuperBuild.cmake#L201-L205

PR #9327, opened 2026-08-01
PERF: Skip testing modules unless developer mode is enabled

lassoan:
Modules whose name ends in "Test" or "Tests" did not appear in the module list but they were still loaded, thereby slightly increasing the startup time. On Windows, they increased the cold startup time at least by 0.5 second, but if tests import in numpy, scipy, or pydicom then they could increase the startup time by several seconds.

They are now skipped during registration, which is early enough to avoid the import entirely. The decision is made from the module name, because the module category is only known once the module has been instantiated.

The name is therefore the single rule: qSlicerUtils::isTestingModule() now delegates to qSlicerUtils::isTestingModuleName() instead of inspecting the category, so the registration filter, the module menu and the module list all agree.

To keep name and category consistent, every module in the "Testing" category whose name did not already follow the convention was renamed:

  sceneImport2428, Slicer4Minute, WebEngine, SliceLinkLogic,
  ScenePerformance, RSNAVisTutorial, RSNAQuantTutorial, ShaderProperties,
  ViewControllersSliceInterpolationBug1926, RSNA2012ProstateDemo, JRC2013Vis,
  FiducialLayoutSwitchBug1914, CurvedPlanarReformation,
  BRAINSFitRigidRegistrationCrashIssue4139,
  SlicerRestoreSceneViewCrashIssue3445, VolumeRenderingSceneClose
      -> same name with a "Test" suffix
  SlicerTransformInteractionTest1
      -> SlicerTransformInteractionTest
  SlicerStartupCompletedTestHelperModule
      -> SlicerStartupCompletedHelperModuleTest
  SlicerStartupPreserveSysPathTestHelperModule
      -> SlicerStartupPreserveSysPathHelperModuleTest

Their module, widget, logic and test case classes are renamed to match, as the scripted module conventions require the class names to be derived from the file name.

The LoadableCustomMarkups template declared "Testing.TestCases" as its category, which would place every module generated from it in the Testing category with a name that cannot follow the convention. It now uses "Examples", like the Loadable template.

Testing mode is excluded as well as developer mode, because that is how the automated tests launch the application and they do need their own modules.

see #9203

### DIFF
--- a/Applications/SlicerApp/Testing/Python/BRAINSFitRigidRegistrationCrashIssue4139Test.py
+++ b/Applications/SlicerApp/Testing/Python/BRAINSFitRigidRegistrationCrashIssue4139Test.py
@@ -1,28 +1,30 @@
+import logging
+
 import slicer
 from slicer.ScriptedLoadableModule import *
-from slicer.util import TESTING_DATA_URL
 
 
 #
-# SlicerRestoreSceneViewCrashIssue3445
+# BRAINSFitRigidRegistrationCrashIssue4139Test
 #
 
 
-class SlicerRestoreSceneViewCrashIssue3445(ScriptedLoadableModule):
+class BRAINSFitRigidRegistrationCrashIssue4139Test(ScriptedLoadableModule):
     """Uses ScriptedLoadableModule base class, available at:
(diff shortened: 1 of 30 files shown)


pieper (approved):
Nice idea 👍

PR #9328, opened 2026-08-02
PERF: Load large Python libraries only when they are first used

lassoan:
On Windows, Slicer cold start time was 20 seconds on the latest main branch.
With the fixes in this pull request, the starting time is about 14 seconds.

numpy, scipy, pydicom and the libraries they pull in were imported while the
application started, either from slicer/__init__.py or from the module bodies of
scripted modules that are registered on every launch. Profiling showed these
imports to be a large fraction of Python startup time, and a session that never
touches DICOM or does any array work never needs them.

Move the large package imports out of the application startup:

- Add slicer.util.LazyImport, which resolves a module (or a "pkg.mod:name"
  module attribute) on first attribute access or call, so existing name.attr
  usage stays unchanged.
- Stop importing numpy and scipy eagerly in slicer/__init__.py. The eager import
  was a workaround for an application hang when they were first imported under
  output redirection on Windows 11 (issue #5945); if that recurs, the removed
  block can be restored from the history of this commit.
- Defer the DICOM modules' pydicom, highdicom, requests/dicomweb_client and
  numpy imports to the functions that use them. Importing pydicom in particular
  pulls in its example-data downloader (requests, charset_normalizer) and all
  pixel-data codec handlers. The DICOM SOP Class UID constants that the TID1500
  plugin used to evaluate in its class body (which forced an import of pydicom
  when the class was defined) are now read from pydicom.uid in the two methods
  that compare against them, both of which already work with pydicom datasets.
- Defer numpy in the Endoscopy fly-through and the WebServer request handler,
  and charset_normalizer in SlicerWizard, which the ExtensionWizard module pulls
  in transitively.

Two tests relied on these packages being imported during startup and are updated
accordingly: test_slicer_packaging used numpy as an example of a package that is
present in sys.modules, and SegmentEditorLogicTest used vtk.util.numpy_support
without importing it (which only worked because another module imported it at
startup). Code in extensions may need the same change, as vtk.util is only
available after something imports it.

### DIFF
--- a/Base/Python/slicer/__init__.py
+++ b/Base/Python/slicer/__init__.py
@@ -238,23 +238,6 @@ for kit in available_kits:
     del kit
 
 # -----------------------------------------------------------------------------
-# Import numpy and scipy early, as a workaround for application hang in import
-# of numpy or scipy at application startup on Windows 11 due to output redirection
-# (only needed for embedded Python, not for standalone).
-# See details in https://github.com/Slicer/Slicer/issues/5945
-# The workaround is only needed on Windows: on other platforms these imports
-# cost a noticeable fraction of the application startup time and the packages
-# load on first use instead.
-
-if not standalone_python and os.name == "nt":
-    try:
-        import numpy  # noqa: F401
-        import scipy  # noqa: F401
-        import scipy.linalg  # noqa: F401
-    except ImportError as detail:
-        print(detail)
-
-# -----------------------------------------------------------------------------
 # Cleanup: Removing things the user shouldn't have to see.
 
 del _createModule
--- a/Base/Python/slicer/tests/test_slicer_packaging.py
+++ b/Base/Python/slicer/tests/test_slicer_packaging.py
@@ -713,11 +713,18 @@ class FindUpdatedImportedPackagesTest(unittest.TestCase):
 
     def test_version_changed_and_imported(self):
         """Test that changed AND imported packages are flagged."""
-        before = {"numpy": "1.24.0"}
-        after = {"numpy": "1.26.0"}
+        # The distribution list is mocked below, but sys.modules is not: the tested
+        # function looks up the import name there, so a real package must be imported.
+        # certifi is used because it is always installed (as a requests dependency)
+        # and importing it is cheap (it only provides the path of a CA bundle file).
+        # Heavy packages, such as numpy, cannot be used, as they are not imported
+        # during application startup anymore, but only when they are first used.
+        import certifi  # noqa: F401
 
-        # numpy is already imported in sys.modules in Slicer
-        mock_pkg_dists = {"numpy": ["numpy"]}
+        before = {"certifi": "2025.1.31"}
+        after = {"certifi": "2026.5.20"}
+
+        mock_pkg_dists = {"certifi": ["certifi"]}
         with unittest.mock.patch(
             "importlib.metadata.packages_distributions",
             return_value=mock_pkg_dists,
@@ -725,7 +732,7 @@ class FindUpdatedImportedPackagesTest(unittest.TestCase):
             result = slicer.packaging._find_updated_imported_packages(before, after)
 
         self.assertEqual(len(result), 1)
-        self.assertEqual(result[0], ("numpy", "1.24.0", "1.26.0"))
+        self.assertEqual(result[0], ("certifi", "2025.1.31", "2026.5.20"))
 
     def test_newly_installed_not_flagged(self):
         """Test that packages absent before install are not flagged."""
@@ -742,11 +749,13 @@ class FindUpdatedImportedPackagesTest(unittest.TestCase):
         It won't be in before_versions (uninstalled) but its modules are still
         in sys.modules from the earlier import.
         """
-        before = {}  # Package was uninstalled before pip_ensure ran
-        after = {"numpy": "1.26.0"}  # Now reinstalled
+        # See test_version_changed_and_imported for why certifi is imported here.
+        import certifi  # noqa: F401
 
-        # numpy is already in sys.modules in Slicer
-        mock_pkg_dists = {"numpy": ["numpy"]}
+        before = {}  # Package was uninstalled before pip_ensure ran
+        after = {"certifi": "2026.5.20"}  # Now reinstalled
+
+        mock_pkg_dists = {"certifi": ["certifi"]}
         with unittest.mock.patch(
             "importlib.metadata.packages_distributions",
             return_value=mock_pkg_dists,
@@ -754,7 +763,7 @@ class FindUpdatedImportedPackagesTest(unittest.TestCase):
             result = slicer.packaging._find_updated_imported_packages(before, after)
 
         self.assertEqual(len(result), 1)
-        self.assertEqual(result[0], ("numpy", None, "1.26.0"))
+        self.assertEqual(result[0], ("certifi", None, "2026.5.20"))
 
 
 class PipEnsureRestartPromptTest(unittest.TestCase):
--- a/Base/Python/slicer/util.py
+++ b/Base/Python/slicer/util.py
@@ -81,6 +81,7 @@ __all__ = [
     "itkImageFromVolume",
     "itkImageFromVolumeModified",
     "launchConsoleProcess",
+    "LazyImport",
     "loadAnnotationFiducial",
     "loadAnnotationROI",
     "loadAnnotationRuler",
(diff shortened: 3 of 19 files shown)


pieper (commented):
I did a pass with copilot first just to get a sense and I do think some of the suggestions are good.  Maybe have your copilot or Claude take a look at the text below?


Summary — ready to merge with a few small, actionable fixes and a couple of improvement suggestions.

High-level summary
- Purpose: Reduce Slicer cold-start time by deferring heavy Python package imports (numpy, scipy, pydicom, highdicom, requests/dicomweb_client, charset_normalizer, vtk.util.numpy_support, etc.) until the packages are actually used.
- Main changes:
  - Add slicer.util.LazyImport to resolve a module or module attribute on first use.
  - Remove eager imports of numpy/scipy from startup and replace module-level heavy imports with lazy imports or local imports where needed.
  - Update DICOM-related modules to import pydicom, requests, dicomweb_client, highdicom, charset_normalizer, and numpy lazily in the functions that use them.
  - Adjust tests that assumed numpy was imported during startup (use certifi in one test; explicit import of vtk.util.numpy_support in another).
  - Small build file updates (SuperBuild.cmake, External_VTK.cmake).
- Scope: 19 files changed, +174 / -75 lines (PR metadata). Risk: medium (affects startup behavior and import timing; most changes are conservative but some modules and extensions may rely on side-effects of eager imports).
- CI: I did not fetch CI results. Recommend waiting for CI / full test run.

Core changes (why & how)
- The PR introduces LazyImport to allow existing call/attribute usages to keep working while delaying the actual import until the symbol is first used. This addresses profiling data showing heavy imports (numpy, pydicom, etc.) occur at startup and slow cold start time on Windows.
- The PR converts module-level heavy imports to either:
  - module-level LazyImport assignments (e.g., np = LazyImport("numpy")), or
  - local imports inside functions/methods that only need the package when executed.
- Tests updated to not rely on package presence in sys.modules at startup.

Key diffs (representative, with links)
- LazyImport implementation (Base/Python/slicer/util.py) — adds the new class and exports it from util.
  - PR file listing: https://github.com/Slicer/Slicer/pull/9328/files (see Base/Python/slicer/util.py)
  - What changed: new class LazyImport with __getattr__ and __call__ delegating to importlib.import_module on first use.

- Deferred DICOM/pydicom imports (examples)
  - Modules/Scripted/DICOMLib/DICOMPlugin.py — moved pydicom import into functions that need it.
  - Modules/Scripted/DICOMLib/DICOMProcesses.py and DICOMUtils.py — added "from __future__ import annotations" and moved requests/dicomweb imports to local scope.
  - Modules/Scripted/DICOMPlugins/* — pydicom moved from top-level imports into functions.

- Deferred numpy / vtk.util.numpy_support and related changes
  - WebServer and Endoscopy modules: replace eager numpy imports with LazyImport or local imports; change uses of vtk.util.numpy_support.vtk_to_numpy to import the function inside the function.

- Tests adjusted:
  - Base/Python/slicer/tests/test_slicer_packaging.py — replaced numpy with certifi for tests that require an imported package at test runtime.
  - Modules/Loadable/Segmentations/Testing/Python/SegmentEditorLogicTest.py — explicitly import vtk.util.numpy_support now.

Merge readiness and risk assessment
- Observations:
  - The changes are targeted and conservative: they replace heavy module-level imports with either LazyImport or local imports.
  - The project already recognized the risk in the PR description (some extensions or scripts may have relied on imports at startup); tests were updated where necessary.
  - The LazyImport API is small and the code paths it touches are limited.
- Risks:
  - Extensions or third-party modules that relied on module-level import side effects during application startup may break. The PR description acknowledges this; authors should communicate the change to extension authors.
  - LazyImport is a thin wrapper; a few details could cause surprises (see actionable items below).
  - Minor behavioral change: Import-time exceptions that used to surface during startup now are delayed to first use — this is intended but worth noting.
- Recommended merge status: Approve after addressing the small, concrete comments below and after a successful CI run.


Actionable review points (files & precise locations)
- Base/Python/slicer/util.py — LazyImport._resolve / thread-safety
  - Suggestion: make _resolve thread-safe (use a lock) to avoid two threads racing to import and setting _resolved twice. Python's import machinery serializes imports by module name, but adding a small threading.Lock (or using double-check with lock) avoids a redundant importlib.import_module call and is defensive.
  - Rationale: concurrent GUIs or parallel startup code paths may call the same LazyImport simultaneously; adding a lock is inexpensive and reduces subtle race behavior.

- Modules/Scripted/DICOMLib/DICOMUtils.py — getGlobalDICOMAuth
  - Suggestion: wrap the local import of requests in try/except ImportError and return None when requests is not installed, rather than letting ImportError propagate.
  - Rationale: callers expect None if no credentials are available; if requests is optional (as made by this PR), failing to import requests here could raise an unexpected ImportError during runtime.

- Modules/Scripted/DICOMPlugins/DICOMScalarVolumePlugin.py — naming consistency
  - Suggestion: verify the rest of the module uses the name `numpy` (module-level LazyImport("numpy")) and not `np`. If other code expects `np`, pick a single convention (prefer `np = LazyImport("numpy")`) and keep consistent.
  - Rationale: some code uses `np` throughout the codebase; inconsistent renaming may cause NameError at runtime.

- Modules/Scripted/WebServer/WebServerLib/SlicerRequestHandler.py — use of vtk.util.numpy_support
  - Suggestion: current change imports vtk at module level and imports vtk_to_numpy inside the function. Confirm no other functions in the module expect `vtk.util` to be a module-level symbol. If any other code in the module imports `vtk.util.numpy_support` by name (not via vtk), ensure they still work. Consider adding a small inline comment explaining why the function-level import is used (you already added one — good).

- Base/Python/slicer/tests/test_slicer_packaging.py — use of certifi
  - Suggestion: ensure certifi is available in all CI/test environments where this test runs; it is normally available as a dependency of requests, but confirm the specific CI images used include it. If not guaranteed, prefer a built-in stdlib module (but certifi is a reasonable choice).

Other suggested improvements (not blocking)

- Base/Python/slicer/util.py — LazyImport: add __dir__ and __repr__
  - Suggestion: implement __dir__ to return dir(self._resolve()) and __repr__ that indicates it's a LazyImport to help debugging and improve introspection in interactive sessions.
  - Rationale: improves developer ergonomics while debugging lazy imports.

- Base/Python/slicer/util.py — consider forwarding module metadata
  - Suggestion: after resolving, set attributes like __name__, __package__, __doc__ on the LazyImport instance if appropriate, or provide properties to access them.
  - Rationale: some code may introspect module attributes; making them available after resolution reduces surprises.

- Modules/Scripted/DICOMPlugins/* — consistent import strategy
  - Suggestion: prefer one strategy consistently: either module-level LazyImport for frequently used heavy modules (with a clear alias: np/pydicom) or local imports inside the functions that only occasionally execute. Use module-level LazyImport where a module is used frequently in the file to avoid repeated local imports.
  - Rationale: keeps style consistent and avoids repeated import attempts inside hot functions.

What I checked
- Reviewed diffs for the files changed listed in the PR. Confirmed LazyImport is simple and covers the basic use case (module:attr with ":" separator), plus __getattr__ and __call__ to delegate.
- Verified tests updated reasonably (replaced numpy-based checks with certifi and explicit import of vtk.util.numpy_support where needed).
- Spot-checked DICOM plugin changes to ensure no module-level calls to pydicom or numpy remain (class-body constants that referenced pydicom.uids were moved to runtime checks).

### Want me to…

1. Run a focused check across the repo (code search) to find any other modules that still import heavy packages at module level and propose the same lazy approach.
2. Draft a tiny follow-up patch that adds a threading.Lock + __repr__ + __dir__ to LazyImport and update the PR with that change.
3. Re-run or fetch the CI/test results for this PR and summarize failing tests (if any).

If you want, I can do (1) or (2) now — say which and I’ll run the code search / prepare the patch and push a follow-up commit.

lassoan:
Thank you, I've made some improvements in response to these review comments:
- Thread safety: added locking (to make sure the lazy importer only imports the target once, even if used at the same time from multiple threads)
- LazyImport `__repr__` and `__dir__` implementation: Now `__repr__` shows the target and resolution state without triggering the import (useful for debugging), `__dir__` returns `dir(self._resolve())` (so that tab-completion in the Python console triggers the import).
- LazyImport Metadata forwarding: after resolution, `__name__`, `__doc__`, etc. are now copied into the instance.
- SlicerRequestHandler vtk.util: the function-local `from vtk.util.numpy_support import vtk_to_numpy` usage clarified in a comment

I did not make any code changes in response to the remaining comments:
- Modules/Scripted/DICOMLib/DICOMUtils.py — getGlobalDICOMAuth -> `requests` is expected to be available, it is good to report an error if it is not
- Base/Python/slicer/tests/test_slicer_packaging.py — use of certifi -> certify should be always available in the test environment
- DICOMScalarVolumePlugin naming: the file uses numpy consistently -> there are no `np.` references, the module-level import was import numpy; consistency with other files is not strong enough reason to change many lines of code

pieper (approved):
LGTM 👍

hjmjohnson (approved):

PR #9329, opened 2026-08-02
BUG: Keep 3D picking fast when a large surface is shown

pieper:
### Symptom
Picking in a 3D view is janky whenever a large surface is shown — most commonly
a segmentation closed surface. Markup control-point dragging (default snap mode
"to visible surface") and hover read-out both pick on every mouse move; hiding
the surface makes interaction smooth again. A plain `forceRender()` of the same
scene is fast (~12 ms), so the cost is in picking, not rendering.

### Cause
`vtkCellPicker` tests **every cell of every pickable surface** unless a spatial
locator is registered for that surface's data. Over a segmentation closed
surface with millions of cells each pick costs tens to hundreds of milliseconds
(measured ~184 ms over a 2.4 M-cell surface; a warmed pick is as slow as the
first, confirming it is brute force, not a one-time build).

### Fix
Add **`vtkMRMLAccuratePicker`** (a `vtkCellPicker` subclass) that registers and
caches a `vtkStaticCellLocator` for each large, pickable surface in the renderer
before every pick — rebuilding a locator only when its surface changes and
dropping it when the surface is hidden. Picks become indexed queries; tolerance,
picked position, and returned normal are unchanged.

Following review feedback, this is **shared per view rather than per widget**:
`vtkMRMLThreeDViewInteractorStyle` now creates a `vtkMRMLAccuratePicker` as the
picker it already hands to `vtkMRMLInteractionEventData`, so every widget — and
anything that needs quick localization in world coordinates — reuses one picker
and one set of locators. This also accelerates the interaction-context / hover
pick (`ComputeAccurateWorldPosition`), not just markups. The markups widget
passes the shared picker to its representation for each interaction event, so
control-point dragging uses it too.

| | before | after |
|---|---|---|
| pick over a 2.4 M-cell surface | ~184 ms | ~0.03 ms |
| scripted control-point drag over the surface | ~184 ms/move | ~0.4 ms/move |

The only residual cost is a one-time locator build (~50 ms for a multi-million
cell surface) on the first pick after such a surface appears — a single frame,
then sub-millisecond. Helps any large surface, not only segmentations.

### Test
`vtkMRMLAccuratePickerTest` picks repeatedly over a 1.6 M-cell surface and
asserts the picks stay fast (< 30 ms/pick). It fails (~78 ms/pick) without the
locator and passes (~0.2 ms/pick) with it; the wide gap keeps it insensitive to
hardware speed.

### DIFF
--- a/Libs/MRML/DisplayableManager/CMakeLists.txt
+++ b/Libs/MRML/DisplayableManager/CMakeLists.txt
@@ -107,6 +107,8 @@ set(KIT_SRCS
   vtkMRMLSliceViewInteractorStyle.cxx
   vtkMRMLThreeDViewInteractorStyle.cxx
 
+  vtkMRMLAccuratePicker.cxx
+
   # Widgets
   vtkMRMLAbstractWidget.cxx
   vtkMRMLAbstractWidgetRepresentation.cxx
--- a/Libs/MRML/DisplayableManager/Testing/Cxx/CMakeLists.txt
+++ b/Libs/MRML/DisplayableManager/Testing/Cxx/CMakeLists.txt
@@ -11,6 +11,7 @@ set(KIT_TEST_SRCS
   vtkMRMLThreeDViewDisplayableManagerFactoryTest1.cxx
   vtkMRMLDisplayableManagerFactoriesTest1.cxx
   vtkMRMLSliceViewDisplayableManagerFactoryTest.cxx
+  vtkMRMLAccuratePickerTest.cxx
   )
 create_test_sourcelist(Tests ${KIT}CxxTests.cxx
   ${KIT_TEST_SRCS}
--- a/Libs/MRML/DisplayableManager/Testing/Cxx/vtkMRMLAccuratePickerTest.cxx
+++ b/Libs/MRML/DisplayableManager/Testing/Cxx/vtkMRMLAccuratePickerTest.cxx
@@ -1,0 +1,116 @@
+/*==============================================================================
+
+  Program: 3D Slicer
+
+  Copyright (c) Kitware Inc.
+
+  See COPYRIGHT.txt
+  or http://www.slicer.org/copyright/copyright.txt for details.
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+
+==============================================================================*/
+
+// This guards the interactive performance of accurate picking in 3D views.
+// vtkMRMLAccuratePicker is the picker shared per view (by
+// vtkMRMLThreeDViewInteractorStyle) and used on every mouse move for markup
+// dragging and hover read-out. A plain vtkCellPicker tests every cell of every
+// pickable surface, so over a large mesh (for example a segmentation closed
+// surface with millions of cells) each pick costs tens to hundreds of
+// milliseconds and interaction becomes janky. vtkMRMLAccuratePicker indexes
+// large surfaces with cell locators; this test checks that repeated picks over
+// a large mesh stay fast.
+
+// MRMLDisplayableManager includes
+#include "vtkMRMLAccuratePicker.h"
+
+// VTK includes
+#include <vtkActor.h>
+#include <vtkNew.h>
+#include <vtkPolyData.h>
+#include <vtkPolyDataMapper.h>
+#include <vtkRenderWindow.h>
+#include <vtkRenderer.h>
+#include <vtkSphereSource.h>
+#include <vtkTimerLog.h>
+
+// STD includes
+#include <cstdlib>
+#include <iostream>
+
+int vtkMRMLAccuratePickerTest(int vtkNotUsed(argc), char* vtkNotUsed(argv)[])
+{
+  // A large, pickable surface standing in for a segmentation closed surface.
+  vtkNew<vtkSphereSource> sphere;
+  sphere->SetThetaResolution(900);
+  sphere->SetPhiResolution(900);
+  sphere->Update();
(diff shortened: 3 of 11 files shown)


pieper:
I ran across this behavior when testing something else and I think this is a good fix for it.

lassoan:
This makes a lot of sense. The only thing that looks suboptimal that every 3D widget representation maintains (builds/rebuilds/destroys) its own set of locators. It would be better to share one accurate picker per viewer and allow all widgets to access it (and it could be used not just by markups but anything that needs quick localization in world coordinate system).

pieper:
Good point about centralizing.  It turns out there already was a shared picker, but not with the optimized path.  Claude just did an update.  I also realized we should add DataProbe and SubjectHierarchy support to the 3D view using the fast pick.  I'll make a feature PR for that.

Here's what Claude said:

Great point — done. I moved the locator management into a shared vtkMRMLAccuratePicker (a vtkCellPicker subclass) that vtkMRMLThreeDViewInteractorStyle now owns one of per view and hands to vtkMRMLInteractionEventData, so every widget reuses one picker and one set of locators instead of each representation maintaining its own. A nice side effect: this also accelerates the interaction-context / hover pick (ComputeAccurateWorldPosition), which had the same un-indexed cost, so it's usable by anything that needs quick world-coordinate localization — not just markups. The markups widget passes the shared picker to its representation per event for dragging. Updated the PR; the regression test now targets the shared picker directly.

lassoan (dismissed):

The fallback mechanism (to a instantiate a per-markup picker if a shared picker is not received) is not necessary and could be removed to simplify the code. But it does not cause much harm if it is kept.

Thibault-Pelletier:
Out of curiosity, why are we using a cell picker in this case and not a hardware picker (which should be the fastest because its GPU)?

lassoan:
The hardware picker that was available when we implemented markups (in 2019) was `vtkWorldPointPicker`. It was too inaccurate to be usable and it ignored transparent surfaces and volume rendering. It also could not tell which actor was picked. Some of these limitations may not apply to the `vtkHardwarePicker`, which combines hardware-based approximate picking with cell picking, so it may worth experimenting with it (although transparency may still be an issue and per-pick cost might be still quite high). Although if the cell picker with the locator gets fast enough with this improvement then there may be less need for a better solution.

pieper:
At least on my machine this picker is plenty fast.  My test was MRHead thresholded with default level and then show 3D with default mesh settings, and then drag a markup point back and forth over the surface.  Without this fix the jank was terrible.  With this fix it is silky smooth.  So I don't see any headroom for the GPU picker to help.

lassoan (approved):
This looks great, thank you.

lassoan:
There could be a few niche use cases, such as 4D data replay, where locator rebuilding time may have non-negligible cost. However, for very fast 4D replay we may need to keep the actors in the renderer anyway (just hide them) and then we could adjust the logic to keep the locators even if the actor gets hidden.

We may want to keep the locators for hidden actors (and only delete the locator when the actor gets actually removed from the scene) if we experience noticeable delay in picking after showing/hiding a large branch of model hierarchy. But all these kind of tweaks can be done later, when we notice slowdowns or unexpected behavior.

lassoan:
I've rebased this branch on latest main branch and when I start Slicer I get this error:

```
Python 3.12.10 (main, Mar 20 2026, 23:50:55) [MSC v.1943 64 bit (AMD64)] on win32
>>> 
  File "C:/D/S4R/Slicer-build/lib/Slicer-5.13/qt-scripted-modules/DataProbe.py", line 60, in addView
    self.infoWidget = DataProbeInfoWidget(parent)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:/D/S4R/Slicer-build/lib/Slicer-5.13/qt-scripted-modules/DataProbe.py", line 98, in __init__
    self._createSmall()
  File "C:/D/S4R/Slicer-build/lib/Slicer-5.13/qt-scripted-modules/DataProbe.py", line 421, in _createSmall
    self.sliceAnnotations = DataProbeLib.SliceAnnotations()
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\D\S4R\Slicer-build\lib\Slicer-5.13\qt-scripted-modules\DataProbeLib\SliceViewAnnotations.py", line 121, in __init__
    self.updateSliceViewFromGUI()
  File "C:\D\S4R\Slicer-build\lib\Slicer-5.13\qt-scripted-modules\DataProbeLib\SliceViewAnnotations.py", line 321, in updateSliceViewFromGUI
    self.createCornerAnnotations()
  File "C:\D\S4R\Slicer-build\lib\Slicer-5.13\qt-scripted-modules\DataProbeLib\SliceViewAnnotations.py", line 339, in createCornerAnnotations
    self.addSliceViewObserver(sliceViewName)
  File "C:\D\S4R\Slicer-build\lib\Slicer-5.13\qt-scripted-modules\DataProbeLib\SliceViewAnnotations.py", line 348, in addSliceViewObserver
    renderer = renderWindow.GetRenderers().GetItemAsObject(0)
               ^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'vtkmodules.vtkCommonCore.vtkWindow' object has no attribute 'GetRenderers'
```

It is probably related with other recent changes. I don't have time to investigate this now, so it would be good if you could have a look.

pieper:
Yes, that would be the same vtkAddon and Slicer fix issue that just got merged, so we'll keep an eye on it but it should be resolved now.

pieper:
My local build of the current head of main starts with no problem.

lassoan:
I did not update vtkAddon, so that explains it.

hherhold:
I ran into an issue with this; the picked point lies just above the surface. I (well, Claude and I) have a fix; see https://discourse.slicer.org/t/shift-mouse-not-placing-on-surface-of-segment/48240

PR #9334, opened 2026-08-05
BUG: Overflow when one view in the layout

UlysseDurand:
The problem happens on apps with one view and when these functions are called they make a segmentation fault.

This is because when the layout contains only one view, the number of intersecting views is 0, being a size_t, substracting 1 to it makes it overflow to SIZE_MAX.

Fix by early returning in case of numberOfIntersections being 0.


pieper (approved):
Good catch 👍

lassoan (approved):
Looks good, thank you.

PR #9335, opened 2026-08-05
BUG: Apply slice intersection thickness to planar contour points in 2D

cpinter:
Planar contour segment representations have VERTEX cells using PointSize, not LineWidth, so the "Slice intersection thickness" setting (which only set LineWidth) had no visible effect for this representation, unlike closed surface or ribbon.

image


ferdymercury (approved):
Beautiful, thanks a lot !

lassoan (approved):
Looks good to me, thank you.

PR #9340, opened 2026-08-07
BUG: Restore macOS packaged OpenSSL lookup

### LINKED ISSUE #9339: BUG: macOS Preview cannot initialize TLS after dylibs moved to Contents/lib
Ron and I produced this issue with the latest preview build today, revision 34890. As a consequence, we could not download or update extensions from the extension manager.

> [!NOTE]
>  🤖 Bug report below is AI generated

## Summary

The macOS amd64 Preview `5.13.0-2026-08-06`, revision `34890`, cannot initialize Qt's TLS backend. Consequently, Extension Manager metadata requests time out and extensions cannot be installed.

The regression was introduced by commit [`c3467ecd3f`](https://github.com/Slicer/Slicer/commit/c3467ecd3f8d617b2ef8b6ee469c6da2436c748c), `ENH: Speed up macOS packaging and make developer builds distributable`, committed on August 6, 2026.

That commit changed the location of embedded non-framework dylibs without adding the new location to Slicer's runtime library search paths.

## Steps to reproduce

1. Install `Slicer-5.13.0-2026-08-06-macosx-amd64.dmg`.
2. Copy the application to `/Applications`.
3. Start Slicer and open the Extension Manager.
4. Attempt to install an extension.

```text
QSslSocket::connectToHostEncrypted: TLS initialization failed
Failed to download extension metadata from server
Update extension information from server failed: timed out while waiting for server response from https://slicer-packages.kitware.com/api/v1/app/5f4474d0e1d8c75dfc705482/extension
Download of extension failed, could not find an extension with id = 6a75fb11bb1aa295d413b2df
```

The metadata endpoint currently responds successfully, and the extension ID exists, so the “could not find” error is downstream of the failed metadata refresh.

## Expected behavior

Qt should load the packaged OpenSSL libraries, Extension Manager metadata should download successfully, and extensions should install normally.

## Actual behavior

A direct headless check of the packaged application reports:

```text
[SSL] SSL support disabled - Failed to load SSL library !
[SSL] Failed to load Slicer.crt
SSL_SUPPORTED=False
SSL_BUILD=OpenSSL 1.1.1w 11 Sep 2023
SSL_RUNTIME=
```

Adding only the package's `Contents/lib` directory to `DYLD_LIBRARY_PATH` changes the result to:

```text
SSL_SUPPORTED=True
SSL_BUILD=OpenSSL 1.1.1w 11 Sep 2023
SSL_RUNTIME=OpenSSL 1.1.1w 11 Sep 2023
```

This also removes the SSL initialization warnings.

## Root cause

Before `c3467ecd3f`, `gp_item_default_embedded_path_override()` in `SlicerCPackBundleFixup.cmake.in` placed ordinary `.dylib` and `.so` files in:

```text
Contents/${Slicer_LIB_DIR}
```

For this package, that resolves to:

```text
Contents/lib/Slicer-5.13
```

This directory is included in `SLICER_LIBRARY_PATHS_INSTALLED`.

The new [`SlicerBundleCloseDeps.py`](https://github.com/Slicer/Slicer/blob/c3467ecd3f8d617b2ef8b6ee469c6da2436c748c/Utilities/Scripts/SlicerBundleCloseDeps.py#L194-L250) instead places every external non-framework dylib in:

```text
Contents/lib
```

Accordingly, the affected package contains:

```text
Contents/lib/libssl.1.1.dylib
Contents/lib/libcrypto.1.1.dylib
```

However, [`SLICER_LIBRARY_PATHS_INSTALLED`](https://github.com/Slicer/Slicer/blob/c3467ecd3f8d617b2ef8b6ee469c6da2436c748c/CMake/SlicerBlockCTKAppLauncherSettings.cmake#L232-L240) still includes `Contents/lib/Slicer-5.13` but not `Contents/lib`.

Qt 5.15 loads OpenSSL dynamically rather than through a normal Mach-O `LC_LOAD_DYLIB` dependency. Its macOS loader searches `DYLD_LIBRARY_PATH` and the application Frameworks directory for `libssl.*` and `libcrypto.*`; therefore the libraries in `Contents/lib` are not discovered. See the [Qt 5.15 OpenSSL loader](https://github.com/qt/qtbase/blob/5.15/src/network/ssl/qsslsocket_openssl_symbols.cpp#L715-L765).

This dynamic lookup also explains why ordinary bundle dependency verification did not detect the problem.

As a control, Slicer 5.12 packages these libraries under `Contents/lib/Slicer-5.12`, which is on the runtime library path, and `QSslSocket::supportsSsl()` returns `True`.

## Proposed fix

The preferred fix is to make the installed runtime search path match the new packaging layout by adding `Contents/lib` on macOS:

```cmake
if(APPLE)
  list(APPEND SLICER_LIBRARY_PATHS_INSTALLED
    <APPLAUNCHER_SETTINGS_DIR>/../lib
    )
endif()
```

This could be added after the existing `SLICER_LIBRARY_PATHS_INSTALLED` definition in `CMake/SlicerBlockCTKAppLauncherSettings.cmake`.

This seems preferable because `SlicerBundleCloseDeps.py` intentionally centralizes external dylibs in `Contents/lib`, and other libraries loaded dynamically at runtime could otherwise encounter the same problem.

An alternative fix would be to preserve the previous behavior and embed ordinary dylibs in `${Slicer_LIB_DIR}` instead of `Contents/lib`.

## Suggested regression test

Add a smoke test that runs from the fully packaged `.app`, with a clean inherited `DYLD_LIBRARY_PATH`, and verifies:

```python
assert qt.QSslSocket.supportsSsl()
assert qt.QSslSocket.sslLibraryVersionString()
```

Ideally, it should also perform an HTTPS request.

Testing the packaged application is important: the existing build-tree SSL test has access to build-tree library paths, while the bundle verifier only examines static Mach-O dependencies and cannot detect Qt's runtime OpenSSL lookup.

## Environment

```text
Slicer: 5.13.0-2026-08-06
Revision: 34890
Package: macosx-amd64
Qt: 5.15.18
OpenSSL: 1.1.1w
Host: Apple Silicon macOS, running the x86_64 package under Rosetta
```

All relevant Slicer, QtNetwork, OpenSSL, and libcrypto binaries are consistently x86_64. The DMG checksum verifies, and the installed files match the pristine DMG, so this is not caused by architecture mismatch, corruption, quarantine, or certificate configuration.

ebrahimebrahim:
Fixes #9339 

Adds the Contents/lib directory to the installed macOS library paths

NOTE: While this makes sense to me, I did not test it myself. I am not set up properly to develop on macOS. @pieper if it's easier for you to fix this issue yourself please feel free to close this PR and open your own. Otherwise would appreciate help testing that this fixes the issue

### DIFF
--- a/CMake/SlicerBlockCTKAppLauncherSettings.cmake
+++ b/CMake/SlicerBlockCTKAppLauncherSettings.cmake
@@ -239,6 +239,14 @@ set(SLICER_LIBRARY_PATHS_INSTALLED
   <APPLAUNCHER_SETTINGS_DIR>/../${Slicer_QTLOADABLEMODULES_LIB_DIR}
   )
 
+if(APPLE)
+  # SlicerBundleCloseDeps.py embeds external dylibs in Contents/lib. Add that
+  # directory for libraries loaded dynamically at runtime, such as OpenSSL.
+  list(APPEND SLICER_LIBRARY_PATHS_INSTALLED
+    <APPLAUNCHER_SETTINGS_DIR>/../lib
+    )
+endif()
+
 # The following lines allow Slicer to load a CLI module extension that depends
 # on libraries (i.e a logic class) provided by a loadable module extension.
 list(APPEND SLICER_LIBRARY_PATHS_INSTALLED ../${Slicer_QTLOADABLEMODULES_LIB_DIR})
--- a/CMake/Testing/CMakeLists.txt
+++ b/CMake/Testing/CMakeLists.txt
@@ -22,6 +22,10 @@ add_cmakescript_test(
   slicer_setting_variable_message_test
   CMake/UseSlicerMacros.cmake)
 
+add_cmakescript_test(
+  slicer_block_ctk_app_launcher_settings_test
+  CMake/Testing/SlicerBlockCTKAppLauncherSettingsTest.cmake)
+
 if(UNIX)
   add_test(
     NAME cmake_slicer_extension_cpack_bundle_fixup_path_test
--- a/CMake/Testing/SlicerBlockCTKAppLauncherSettingsTest.cmake
+++ b/CMake/Testing/SlicerBlockCTKAppLauncherSettingsTest.cmake
@@ -1,0 +1,19 @@
+cmake_minimum_required(VERSION 3.28.0...3.28.0 FATAL_ERROR)
+
+# Exercise the macOS launcher settings on every host platform.
+set(APPLE TRUE)
+set(Slicer_USE_SYSTEM_QT TRUE)
+set(Slicer_USE_PYTHONQT FALSE)
+set(Slicer_BIN_DIR bin)
+set(Slicer_LIB_DIR lib/Slicer-test)
+set(Slicer_CLIMODULES_LIB_DIR lib/Slicer-test/cli-modules)
+set(Slicer_QTLOADABLEMODULES_LIB_DIR lib/Slicer-test/qt-loadable-modules)
+
+include("${CMAKE_CURRENT_LIST_DIR}/../SlicerBlockCTKAppLauncherSettings.cmake")
+
+set(root_library_path "<APPLAUNCHER_SETTINGS_DIR>/../lib")
+if(NOT root_library_path IN_LIST SLICER_LIBRARY_PATHS_INSTALLED)
+  message(FATAL_ERROR "macOS installed library paths do not contain ${root_library_path}")
+endif()
+
+message("SUCCESS")


pieper:
Yes, I've been looking at it and it's probably an easy fix.  Sorry for the headache.  I tested local but didn't think to try extensions (wouldn't have been any for my local build).

pieper (approved):
This looks good 👍 

The general problem is that libs loaded but not explicitly linked aren't detected.  So this will fix a known case but we should have a more general test to avoid cases like this were we only discover the issues on the user's machine.

Here's an issue to capture what I think should be done:

https://github.com/Slicer/Slicer/issues/9341

ebrahimebrahim:
ty for review and additional analysis!

Issue #9348, opened 2026-08-10
linux compilation/linking broken


Punzo:
last working commit seems to be https://github.com/Slicer/Slicer/commit/7d4b0430d803e19144c345527a7795bbd72ec565

@pieper @lassoan 

See: https://slicer.cdash.org/builds/4351371/build

I can reproduce locally as well.

ebrahimebrahim:
See also a suspect https://github.com/Slicer/vtkAddon/pull/71

pieper:
Analysis by Claude (Anthropic's Claude Code), working with @pieper — we're investigating further and will validate a fix on a Linux build.

The failure is a link error, not a compile one: bin/vtkAddonPython.so fails with undefined reference to PyTuple_Size / PyErr_Occurred / _Py_NoneStruct / PyType_Type … under -Wl,--no-undefined.

Root cause: not #71 (the DEPENDS-pruning PR), but a sibling commit in the same vtkAddon range that the recent SuperBuild vtkAddon bump ([#9337](https://github.com/Slicer/Slicer/pull/9337)) pulled in: [Slicer/vtkAddon@c44ce00](https://github.com/Slicer/vtkAddon/commit/c44ce005) "COMP: Only search for Python3 Development.Module component." 

That commit:

* switched find_package(Python3 COMPONENTS Development) → Development.Module,
* deleted set(VTK_WRAP_PYTHON_FIND_LIBS 1) in vtkMacroKitPythonWrap.cmake (so the "find the Python library" block in vtkWrapPython.cmake is now dead and VTK_Python3_LIBRARIES is never set), and
* removed ${VTK_Python3_LIBRARIES} from both target_link_libraries calls for the wrapped module.

Development.Module provides only Python3::Module, which intentionally leaves the Python C-API symbols undefined for the host interpreter to supply. The code c44ce00 deleted handled this correctly — it was gated on VTK_UNDEFINED_SYMBOLS_ALLOWED and linked the full libpython when undefined symbols are not allowed.

Why Linux-only: on macOS, add_library(MODULE …) links with -undefined dynamic_lookup, so the missing Python symbols are legal (a full macOS package built and ran fine in our testing). On Linux the wrapped modules link with -Wl,--no-undefined (vtk_undefined_symbols_allowed=FALSE), so the absent libpython is fatal.

[...]

PR #9349, opened 2026-08-10
BUG: Keep the install-tree lib layout in the macOS bundle for extensions

### LINKED ISSUE #9347: macOS Preview: compiled C++ extensions fail to load (library-layout regression from the new bundle fixup)
> **Note:** This analysis was written by Claude (Anthropic's Claude Code), based on an
> investigation carried out together with @pieper on the current macOS Preview.

## Summary

On the macOS `amd64` Preview builds produced after c3467ecd3f ("ENH: Speed up macOS packaging and make developer builds distributable") — e.g. `5.13.0-2026-08-06` (rev 34890) and `5.13.0-2026-08-07` (rev 34896) — **extensions that contain compiled C++ modules fail to load**. Reproduced with SlicerVMTK and ExtraMarkups:

```
Cannot load library .../Extensions-34896/SlicerVMTK/lib/Slicer-5.13/qt-loadable-modules/libqSlicerBranchClipperModule.dylib:
  Library not loaded: @rpath/lib/Slicer-5.13/libitkAdaptiveDenoising-5.4.1.dylib
  ...
  '.../Contents/MacOS/../lib/Slicer-5.13/libitkAdaptiveDenoising-5.4.1.dylib' (no such file)
```

Pure-Python extensions are unaffected. TLS/Extension-Manager networking works (that was #9339 / #9340); this is a separate problem in how compiled extension modules resolve core libraries.

## Root cause

The new `Utilities/Scripts/SlicerBundleCloseDeps.py` **flattens** every embedded external dylib into bare `Contents/lib/` and rewrites its install name to `@rpath/lib/<basename>`. So a core library that the `make install` tree publishes at `lib/Slicer-<ver>/` with install name `@rpath/lib/Slicer-<ver>/<name>` becomes `Contents/lib/<name>` with name `@rpath/lib/<name>` (and `libpython` moves from `lib/Python/lib/` to bare `lib/`).

The core app itself works because all of its own binaries were rewritten to the flat names consistently. But **extensions are built by the factory against the `make install` tree**, so they reference the versioned `@rpath/lib/Slicer-<ver>/<name>` (and `@rpath/lib/Python/lib/libpython…`). At load time dyld can neither match those against the already-loaded core libraries (whose install names are now the flat `@rpath/lib/<name>`) nor find the versioned paths on disk, so every compiled extension module fails.

Confirmed by comparing install names:

| | 5.12 (works) | 5.13 preview (broken) |
| --- | --- | --- |
| core lib location | `lib/Slicer-5.12/libitk…` | bare `lib/libitk…` |
| core lib install name | `@rpath/lib/Slicer-5.12/libitk…` | `@rpath/lib/libitk…` |
| extension references | `@rpath/lib/Slicer-5.x/libitk…` | `@rpath/lib/Slicer-5.13/libitk…` |

## Status / fix direction

A fix is in progress: make the bundle fixup **preserve each embedded library's install-tree-relative location and install name** (the invariant is that the packaged `.app` lib layout and install names must match the `make install` tree the extension factory builds against), rather than flattening to bare `lib/`. This also subsumes #9339 (OpenSSL returns to `lib/Slicer-<ver>/`, already on the runtime path). It will be validated by a factory-style VMTK build against the corrected package, run in a clean VM, before the PR is merged.

Related: #9339, #9340, #9341.

pieper:
Fixes #9347

## Problem

The macOS bundle closer introduced in c3467ecd3f (`SlicerBundleCloseDeps.py`) embedded every external dylib **flat into `Contents/lib`** and derived each install name from that flat location (`@rpath/lib/<name>`). The core application worked because all of its own binaries were rewritten to the flat names consistently — but **compiled C++ extensions could not load**:

```
Cannot load library .../Extensions-34898/SlicerVMTK/lib/Slicer-5.13/qt-loadable-modules/libqSlicerBranchClipperModule.dylib:
  Library not loaded: @rpath/lib/Slicer-5.13/libitkAdaptiveDenoising-5.4.1.dylib
```

Extensions are built by the factory against the standard install tree, where core libraries live in the versioned lib dir and are referenced as `@rpath/lib/Slicer-<ver>/<name>` (and the Python runtime as `@rpath/lib/Python/lib/libpython…`). dyld binds an extension's dependency to the already-loaded core library only when the install name matches exactly, so flattening renamed those libraries and broke every compiled extension (SlicerVMTK, ExtraMarkups, …).

## Fix

Restore the historical BundleUtilities layout the extension ABI depends on:

- ordinary dylibs → the versioned Slicer lib dir (passed from CMake as `--lib-subdir "@Slicer_LIB_DIR@"`, e.g. `lib/Slicer-5.13`)
- the Python runtime → `lib/Python/lib`
- frameworks → `Frameworks` (unchanged)

The `@rpath` install name is derived from the embedded location, so fixing the location restores the names. This also returns OpenSSL to `lib/Slicer-<ver>`, which is already on the runtime library path, so it **independently resolves the TLS lookup from #9339** (the `Contents/lib` search-path entry added in #9340 becomes unnecessary, though harmless).

## Verification

Packaged Slicer with the fix and confirmed in the resulting `.dmg`:

- `libitkAdaptiveDenoising-5.4.1.dylib` → `lib/Slicer-5.13/`, id `@rpath/lib/Slicer-5.13/libitkAdaptiveDenoising-5.4.1.dylib`
- `libpython3.12.dylib` → `lib/Python/lib/`, id `@rpath/lib/Python/lib/libpython3.12.dylib`
- `libssl.1.1.dylib` → `lib/Slicer-5.13/`
- bare `Contents/lib/*.dylib` = 0 (was 250); `lib/Slicer-5.13/*.dylib` = 288

Then built **SlicerVMTK** and **ExtraMarkups** factory-style (`Slicer_DIR` → this build, `ExtraMarkups_DIR` → the ExtraMarkups build) against the package, installed both, and loaded every one of their **18 compiled modules with 0 dependency errors** (`dlopen` in-process), plus a clean interactive launch in a vanilla arm64 VM with both extensions installed.

### DIFF
--- a/CMake/SlicerCPackBundleFixup.cmake.in
+++ b/CMake/SlicerCPackBundleFixup.cmake.in
@@ -137,7 +137,7 @@ endforeach()
 message(STATUS "Closing macOS bundle dependencies")
 execute_process(
   COMMAND python3 "@CMAKE_SOURCE_DIR@/Utilities/Scripts/SlicerBundleCloseDeps.py"
-          --app "${app_dir}" ${search_path_args}
+          --app "${app_dir}" --lib-subdir "@Slicer_LIB_DIR@" ${search_path_args}
   RESULT_VARIABLE _closedeps_result
   )
 if(NOT _closedeps_result EQUAL 0)
--- a/Utilities/Scripts/SlicerBundleCloseDeps.py
+++ b/Utilities/Scripts/SlicerBundleCloseDeps.py
@@ -15,9 +15,12 @@ locations under ``/usr/lib`` and ``/System`` -- is either:
 
   - redirected to a library of the same name already embedded in the bundle
     (avoiding a duplicate copy), or
-  - copied into the bundle (a plain dylib into ``Contents/lib``; a framework,
-    whole, into ``Contents/Frameworks``) and the reference rewritten to
-    ``@rpath/...``.
+  - copied into the bundle (an ordinary dylib into the versioned Slicer lib
+    directory given by ``--lib-subdir``, the Python runtime into
+    ``Contents/lib/Python/lib``, a framework whole into ``Contents/Frameworks``)
+    and the reference rewritten to ``@rpath/...``. The embedded location matches
+    the standard install-tree layout so an embedded library keeps the install
+    name compiled extensions were built against.
 
 The walk follows copied libraries' own dependencies, and every Mach-O *inside* a
 copied framework (for example ``QtWebEngineCore.framework``'s bundled
@@ -191,11 +194,16 @@ def resolve(ref, binary_dir, rpaths, search_dirs, exe_dir):
     return None
 
 
-def close_deps(app, search_dirs):
+def close_deps(app, search_dirs, lib_subdir):
     contents = os.path.join(app, "Contents")
     exe_dir = os.path.join(contents, "MacOS")
     frameworks_dir = os.path.join(contents, "Frameworks")
-    lib_dir = os.path.join(contents, "lib")
+    # Embedded ordinary dylibs go in the versioned Slicer lib dir (lib_subdir,
+    # e.g. lib/Slicer-5.13); the Python runtime keeps its own lib/Python/lib
+    # location. This mirrors the historical BundleUtilities layout that compiled
+    # extensions are built against -- see embed() for why the location matters.
+    lib_dir = os.path.join(contents, lib_subdir)
+    python_lib_dir = os.path.join(contents, "lib", "Python", "lib")
     contents_real = os.path.realpath(contents)
 
     def inside_bundle(path):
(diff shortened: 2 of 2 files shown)


ebrahimebrahim (commented):
Thanks for this!

Have you tested macOS packaging with this?

-- 🤖 Issue described by an LLM reviewer below --

One install-name case remains unresolved. Moving a dylib to the versioned directory does not automatically change its `LC_ID_DYLIB`, and the current closer only resets IDs that begin with an absolute path (`/`). Libraries with a bare or existing `@rpath` ID therefore keep their old identity after relocation.

OpenJPEG is a concrete example: its source ID is `libopenjp2.7.dylib`. This change moves it into `lib/Slicer-5.13`, but its ID remains bare, whereas the historical BundleUtilities packager assigned `@rpath/lib/Slicer-5.13/libopenjp2.7.dylib`. This means the file layout is restored, but the install-name identity that extensions link against is not always restored, potentially defeating the dyld coalescing this change is intended to preserve.

Could the closer set the destination-derived `@rpath/<path-relative-to-Contents>` ID for every copied dylib whenever its existing ID differs, rather than only resetting absolute IDs?

pieper:
> Have you tested macOS packaging with this?

Yes, I did a build and cpack package and then created a VM using [tart](https://tart.run/) and tested that vmtk (which needed custom markups) extension installed and ran with no link error.

The other point it probably not a breakage, but an inconsistency we should fix.  We're testing a fix now.

ebrahimebrahim (approved):

pieper:
CI build failed on this because we are waiting for fix of https://github.com/Slicer/Slicer/issues/9348 to be merged.

PR #9353, opened 2026-08-11
COMP: Reduce actions and pre-commit update frequency

ebrahimebrahim:
From [last week's discussion](https://discourse.slicer.org/t/2026-08-04-weekly-meeting/47761/7#p-135254-dependabot-6).

Reduces the updates to monthly and also groups the actions updates together into one PR

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,8 +5,14 @@ updates:
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
-      # Check for updates to GitHub Actions every week
-      interval: "weekly"
+      # Check for updates to GitHub Actions every month
+      interval: "monthly"
+    cooldown:
+      default-days: 14
+    groups:
+      github-actions:
+        patterns:
+          - "*"
     commit-message:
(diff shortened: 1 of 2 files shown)


ebrahimebrahim:
@jamesobutler Do you think we can get this integrated to reduce the frequency of the automatic actions updates? It's also arguably healthier for us not to always have the latest of everything

jamesobutler:
For dependabot have you seen the recent post on a “cooldown” period? This helps prevent propagation of supply chain attacks, but security updates are not subject to the cooldown so they automatically are issued when they arise.

https://github.blog/changelog/2026-07-14-dependabot-version-updates-introduce-default-package-cooldown/

ebrahimebrahim:
No I hadn't know about that and was actually hoping that a "cooldown" like thing existed. Very happy to see that it now  does!

Added a 2 week cooldown. 

So the updated will be monthly, but if during the monthly check an update isn't at least two weeks old then it should get put off to the next month, for a maximum lag time of ~1.5mo

jamesobutler (approved):

PR #9355, opened 2026-08-13
ENH: Optimize interaction handle rendering with vtkGlyph3DMapper

Sunderlandkyl:
Replace the per-representation vtkAppendPolyData / vtkTransformPolyDataFilter CPU pipeline with a single shared vtkGlyph3DMapper per representation that batches all handle glyphs (rotation, translation, scale) into one GPU instanced draw call, using per-instance orientation/scale/color arrays instead of composing handle polydata by hand on the CPU every frame.

Previously the full handle polydata pipeline (append + transform filter) was rebuilt on every render call, even when nothing about the handles had actually changed - e.g. camera clipping-range updates alone were enough to trigger MTime propagation and force a full rebuild. Add a RenderState struct that caches the camera pose, slice transform MTime, and active component, and only rebuilds the instance arrays when that state actually changes.

Benchmark: 80 vtkMRMLMarkupsROINode instances with HandlesInteractive enabled simultaneously in the 3D view (~1200 handle actors total), orbiting the camera for 100 Render() calls x 2 runs, identical window size (1384x835), RelWithDebInfo build, before/after this commit:

| | main (before) | branch (after) | speedup |
| --- | ---  | --- | --- |
| avg | 568 ms/frame | 76 ms/frame | ~7.5x |
| median | 557 ms/frame | 71 ms/frame | ~7.8x |
| min | 503 ms/frame | 62 ms/frame | ~8.1x |
| frame rate | ~1.8 fps | ~13 fps | ~7.2x  |

### DIFF
--- a/Libs/MRML/DisplayableManager/vtkMRMLInteractionWidgetRepresentation.cxx
+++ b/Libs/MRML/DisplayableManager/vtkMRMLInteractionWidgetRepresentation.cxx
@@ -19,33 +19,22 @@
 ==============================================================================*/
 
 // VTK includes
-#include <vtkAppendPolyData.h>
+#include <vtkCallbackCommand.h>
 #include <vtkCamera.h>
-#include <vtkCellData.h>
 #include <vtkDoubleArray.h>
 #include <vtkEllipseArcSource.h>
 #include <vtkFloatArray.h>
 #include <vtkFocalPlanePointPlacer.h>
-#include <vtkGlyph3D.h>
 #include <vtkLine.h>
-#include <vtkLineSource.h>
-#include <vtkLookupTable.h>
+#include <vtkMath.h>
 #include <vtkMRMLSliceNode.h>
 #include <vtkMRMLViewNode.h>
 #include <vtkPlane.h>
 #include <vtkPointData.h>
-#include <vtkPointSetToLabelHierarchy.h>
+#include <vtkPoints.h>
 #include <vtkRenderer.h>
 #include <vtkRenderWindow.h>
-#include <vtkSphereSource.h>
-#include <vtkStringArray.h>
-#include <vtkTextActor.h>
-#include <vtkTextProperty.h>
-#include <vtkTensorGlyph.h>
 #include <vtkTransform.h>
-#include <vtkTransformPolyDataFilter.h>
-#include <vtkTriangleFilter.h>
-#include <vtkTubeFilter.h>
 
 // MRML includes
 #include <vtkMRMLAbstractThreeDViewDisplayableManager.h>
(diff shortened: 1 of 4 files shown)


Libs/MRML/DisplayableManager/vtkMRMLInteractionWidgetRepresentation.cxx
```diff
@@ -55,6 +44,198 @@
...
+
+  /// The shared actor is rendered by a single representation, but every representation is a prop of
+  /// the renderer and is therefore asked to render itself. The first representation that is asked in
+  /// a frame claims the shared actor and renders it on behalf of all of the others, which render
+  /// nothing. Reset at the beginning of each frame.
+  /// A per-frame "already rendered" flag cannot be used instead, because a render pass may render
+  /// the same pass several times in a single frame (depth peeling renders the translucent geometry
+  /// once to initialize the depth buffers and then once per peel), and the shared actor has to be
+  /// rendered in each of those invocations.
+  vtkMRMLInteractionWidgetRepresentation* RenderingRepresentation{ nullptr };
```
lassoan: vtkWeakPointer (or smart pointer) would be safer

Libs/MRML/DisplayableManager/vtkMRMLInteractionWidgetRepresentation.cxx
```diff
@@ -55,6 +44,198 @@
...
+  vtkSmartPointer<vtkFloatArray> GlyphOrientationArray;
+  vtkSmartPointer<vtkFloatArray> GlyphScaleArray;
+  vtkSmartPointer<vtkIntArray> GlyphSourceIndexArray;
+  vtkSmartPointer<vtkUnsignedCharArray> ColorArray;
+
+  vtkSmartPointer<vtkGlyph3DMapper> Mapper;
+  vtkSmartPointer<vtkProperty> Property;
+  vtkSmartPointer<vtkActor> Actor;
+
+  std::set<vtkMRMLInteractionWidgetRepresentation*> Representations;
```
lassoan: Consider using a smart or weak pointer to ensure memory safety.

```suggestion
  std::set<vtkSmartPointer<vtkMRMLInteractionWidgetRepresentation>> Representations;
```

Libs/MRML/DisplayableManager/vtkMRMLInteractionWidgetRepresentation.cxx
```diff
@@ -55,6 +44,198 @@
...
+  /// the same pass several times in a single frame (depth peeling renders the translucent geometry
+  /// once to initialize the depth buffers and then once per peel), and the shared actor has to be
+  /// rendered in each of those invocations.
+  vtkMRMLInteractionWidgetRepresentation* RenderingRepresentation{ nullptr };
+  bool GlyphSourcesInitialized{ false };
+  unsigned long StartEventTag{ 0 };
+  unsigned long DeleteEventTag{ 0 };
+};
+
+std::map<vtkRenderer*, SharedHandleRenderer*> g_SharedRenderers;
```
lassoan: It would be better to avoid raw pointers.

Libs/MRML/DisplayableManager/vtkMRMLInteractionWidgetRepresentation.cxx
```diff
@@ -468,189 +680,240 @@ void vtkMRMLInteractionWidgetRepresentation::OrthoganalizeTransform(vtkTransform
...
+            {
+              matrix3x3[j][i] = orient9[i * 3 + j];
+            }
+          }
+          vtkMath::Matrix3x3ToQuaternion(matrix3x3, quat);
+        }
+      }
+      else
+      {
+        double viewDir_Handle[3];
```
lassoan: initialize to {0.0,0.0,1.0} here (to ensure it is set; also, then there is no need to set it later)

Libs/MRML/DisplayableManager/vtkMRMLInteractionWidgetRepresentation.cxx
```diff
@@ -468,189 +680,240 @@ void vtkMRMLInteractionWidgetRepresentation::OrthoganalizeTransform(vtkTransform
...
+          worldToHandleTransform->TransformVector(viewDir_World, viewDir_Handle);
+        }
+        else
+        {
+          viewDir_Handle[0] = 0.0;
+          viewDir_Handle[1] = 0.0;
+          viewDir_Handle[2] = 1.0;
+        }
+
+        double xAxis[3], yAxis[3], zAxis[3];
```
lassoan: initialize

Libs/MRML/DisplayableManager/vtkMRMLInteractionWidgetRepresentation.cxx
```diff
@@ -468,189 +680,240 @@ void vtkMRMLInteractionWidgetRepresentation::OrthoganalizeTransform(vtkTransform
...
-    transformOutlineGlyph->SetOutput(handleOutlineGlyphPolyData);
-    transformOutlineGlyph->Update();
-    if (initializeAppendFilter)
-    {
-      this->Pipeline->Append->AddInputData(handleOutlineGlyphPolyData);
-    }
+      if (writeWorldCoordinates)
+      {
+        double point_Handle[3] = { point[0], point[1], point[2] };
+        point[0] = h2w[0][0] * point_Handle[0] + h2w[0][1] * point_Handle[1] + h2w[0][2] * point_Handle[2] + h2w[0][3];
```
lassoan: Al these low-level manipulations are hard to read. If they are needed for performance reasons then add a comment. If they are not critical then it would be nicer to use higher-level VTK and vtkAddon functions.

Libs/MRML/DisplayableManager/vtkMRMLInteractionWidgetRepresentation.cxx
```diff
@@ -468,189 +680,240 @@ void vtkMRMLInteractionWidgetRepresentation::OrthoganalizeTransform(vtkTransform
...
-    vtkSmartPointer<vtkFloatArray> outlineGlyphColorArray = vtkFloatArray::SafeDownCast(handleOutlineGlyphPolyData->GetPointData()->GetAbstractArray("colorIndex"));
-    if (!outlineGlyphColorArray)
-    {
-      outlineGlyphColorArray = vtkSmartPointer<vtkFloatArray>::New();
-      outlineGlyphColorArray->SetName("colorIndex");
-      handleOutlineGlyphPolyData->GetPointData()->AddArray(outlineGlyphColorArray);
+      double scale = this->WidgetScale;
+      if (type == InteractionRotationHandle && index == 3)
+      {
+        scale *= 1.2;
```
lassoan: add comment on why 1.2

Libs/MRML/DisplayableManager/vtkMRMLInteractionWidgetRepresentation.cxx
```diff
@@ -700,44 +1056,190 @@ int vtkMRMLInteractionWidgetRepresentation::RenderOpaqueGeometry(vtkViewport* vi
...
+    this->UpdatePipelineIfNeeded();
+    if (this->HasVisibleHandles)
+    {
+      count += this->Pipeline->Actor->RenderOpaqueGeometry(viewport);
+    }
+    return count;
   }
+
+  // 3D view: shared rendering — one draw call for ALL representations
+  SharedHandleRenderer* shared = GetOrCreateSharedRenderer(this->Renderer);
```
lassoan: use `this->` to match VTK convention

Libs/MRML/DisplayableManager/vtkMRMLInteractionWidgetRepresentation.cxx
```diff
@@ -700,44 +1056,190 @@ int vtkMRMLInteractionWidgetRepresentation::RenderOpaqueGeometry(vtkViewport* vi
...
+    shared->Mapper->SetSourceData(GlyphCrosshair, this->Pipeline->CrosshairPolyData);
+    shared->Mapper->SetSourceData(GlyphArrowOutline, this->Pipeline->ArrowOutlinePolyData);
+    shared->Mapper->SetSourceData(GlyphCircleOutline, this->Pipeline->CircleOutlinePolyData);
+    shared->Mapper->SetSourceData(GlyphRingOutline, this->Pipeline->RingOutlinePolyData);
+    shared->Mapper->SetSourceData(GlyphCrosshairOutline, this->Pipeline->CrosshairOutlinePolyData);
+    this->UpdateRelativeCoincidentTopologyOffsets(shared->Mapper);
+    shared->GlyphSourcesInitialized = true;
+  }
+
+  if (!IsSharedRenderingRepresentation(shared, this))
```
lassoan: use `this->` to match VTK convention (everywhere else, too)

lassoan:
I've reviewed the code also with Claude. Here are the comments that I found to be relevant.

## Global variable handling

[vtkMRMLInteractionWidgetRepresentation.cxx#L89](https://github.com/Slicer/Slicer/blob/3ba8048b7c71cb538fdb258063a467d9dd61635c/Libs/MRML/DisplayableManager/vtkMRMLInteractionWidgetRepresentation.cxx#L89) introduces

```cpp
std::map<vtkRenderer*, SharedHandleRenderer*> g_SharedRenderers;
```

- **Name:** the `g_` prefix is not used anywhere else in the Slicer codebase. The direct precedent for exactly this shape — a per-`vtkRenderer*` cache shared across widget representations — is [`vtkSlicerMarkupsWidgetRepresentation3D::CachedZBuffers`](https://github.com/Slicer/Slicer/blob/1692b62df7ff7276386ecc2dc16abbe9090623b6/Modules/Loadable/Markups/VTKWidgets/vtkSlicerMarkupsWidgetRepresentation3D.h#L198): a PascalCase **static class member** with static accessor/cleanup member functions. Process-wide singletons follow the `<ClassName>Instance` pattern ([vtkEventBroker.cxx#L36](https://github.com/Slicer/Slicer/blob/1692b62df7ff7276386ecc2dc16abbe9090623b6/Libs/MRML/Core/vtkEventBroker.cxx#L36), [vtkMRMLI18N.cxx#L32](https://github.com/Slicer/Slicer/blob/1692b62df7ff7276386ecc2dc16abbe9090623b6/Libs/MRML/Core/vtkMRMLI18N.cxx#L32)).
- **Construction:** existing globals are either zero-initialized POD pointers created lazily (the comment in `vtkMRMLI18N.cxx` explains why: static-init-order safety) or maps holding `vtkSmartPointer` values (`CachedZBuffers`). This map holds raw pointers allocated with `new` — `std::map<vtkRenderer*, std::unique_ptr<SharedHandleRenderer>>` would match the existing smart-pointer-value convention at no cost.
- **Destruction:** existing globals use deterministic teardown (Schwarz-counter `classInitialize()`/`classFinalize()` for the VTK singletons, `...CleanupGlobal` structs on the Qt side, explicit `erase` for `CachedZBuffers`). Here cleanup relies solely on the renderer's `DeleteEvent`:
  - a leaked renderer, or `RemoveAllObservers()` on a renderer, leaks the `SharedHandleRenderer` (the map's exit-time destructor does not delete values);
  - a renderer destroyed after static destruction makes `OnSharedRendererDelete` touch an already-destructed map (UB at shutdown);
  - `StartEventTag`/`DeleteEventTag` ([#L85-L86](https://github.com/Slicer/Slicer/blob/3ba8048b7c71cb538fdb258063a467d9dd61635c/Libs/MRML/DisplayableManager/vtkMRMLInteractionWidgetRepresentation.cxx#L85-L86)) are stored but there is no `RemoveObserver` anywhere — dead members implying cleanup that never happens.

Suggestion: make it a static class member accessed via a static `GetSharedRenderer(vtkRenderer*)`, hold values by `unique_ptr`, and drop the unused tags — matching `CachedZBuffers` in name, placement, and ownership.

## Behavior regressions

1. **Line width doubled.** The shared property hard-codes [`SetLineWidth(2.0)`](https://github.com/Slicer/Slicer/blob/3ba8048b7c71cb538fdb258063a467d9dd61635c/Libs/MRML/DisplayableManager/vtkMRMLInteractionWidgetRepresentation.cxx#L144) where the old `Property3D` used [`1.0`](https://github.com/Slicer/Slicer/blob/1692b62df7ff7276386ecc2dc16abbe9090623b6/Libs/MRML/DisplayableManager/vtkMRMLInteractionWidgetRepresentation.cxx#L995) — every outline/ring/crosshair renders twice as thick in all views, and screenshot test baselines will shift.

2. **Ghost handles after detach.** [`vtkMRMLAbstractWidget::SetRenderer`](https://github.com/Slicer/Slicer/blob/1692b62df7ff7276386ecc2dc16abbe9090623b6/Libs/MRML/DisplayableManager/vtkMRMLAbstractWidget.cxx#L376-L380) only forwards to the representation when the *new* renderer is non-null, so the new [`SetRenderer` override](https://github.com/Slicer/Slicer/blob/3ba8048b7c71cb538fdb258063a467d9dd61635c/Libs/MRML/DisplayableManager/vtkMRMLInteractionWidgetRepresentation.cxx#L624-L631) never runs on `SetRenderer(nullptr)`. The rep stays in `shared->Representations`, and because the shared actor bypasses the renderer's prop list, a detached-but-alive widget's handles keep rendering in the old 3D view (same for `renderer->RemoveViewProp(rep)`). In-tree code happens to destroy widgets instead, but this is a trap for extensions.

3. **Unclamped, truncating color conversion** ([#L857-L860](https://github.com/Slicer/Slicer/blob/3ba8048b7c71cb538fdb258063a467d9dd61635c/Libs/MRML/DisplayableManager/vtkMRMLInteractionWidgetRepresentation.cxx#L857-L860)): `static_cast<unsigned char>(color[i] * 255)` truncates (0.999 → 254; `vtkScalarsToColors::ColorToUChar` rounds), and an opacity > 1 (settable from Python; the old `vtkLookupTable` path clamped) is an out-of-range float→int conversion (UB) — in practice a mostly-transparent handle that also reclassifies the whole shared actor as translucent.

4. **Stale handle size — inputs missing from `RenderState`.** `UpdateViewScaleFactor` now clamps depth to [`GetClippingRange()[0]`](https://github.com/Slicer/Slicer/blob/3ba8048b7c71cb538fdb258063a467d9dd61635c/Libs/MRML/DisplayableManager/vtkMRMLInteractionWidgetRepresentation.cxx#L2202), but clipping range is deliberately excluded from `RenderState` (whose [comment](https://github.com/Slicer/Slicer/blob/3ba8048b7c71cb538fdb258063a467d9dd61635c/Libs/MRML/DisplayableManager/vtkMRMLInteractionWidgetRepresentation.h#L346) still says it "does not affect the appearance of the handles") — in the VR walk-past scenario the clamp was added for, the cache gate stays shut and the handle size freezes. Similarly, [`GetScreenSize()`](https://github.com/Slicer/Slicer/blob/3ba8048b7c71cb538fdb258063a467d9dd61635c/Libs/MRML/DisplayableManager/vtkMRMLInteractionWidgetRepresentation.cxx#L2168) feeds handle sizing but is not in `RenderState`, so moving the window to a different-resolution monitor leaves stale sizes until the camera moves.

5. **`GetActors()` / `ReleaseGraphicsResources()` inconsistency.** In 3D views, [`GetActors`](https://github.com/Slicer/Slicer/blob/3ba8048b7c71cb538fdb258063a467d9dd61635c/Libs/MRML/DisplayableManager/vtkMRMLInteractionWidgetRepresentation.cxx#L926-L932) returns only the per-rep actor (never rendered in 3D) and omits the shared actor that actually draws — breaking prop-collection consumers such as the OBJ/glTF exporters. Meanwhile [`ReleaseGraphicsResources`](https://github.com/Slicer/Slicer/blob/3ba8048b7c71cb538fdb258063a467d9dd61635c/Libs/MRML/DisplayableManager/vtkMRMLInteractionWidgetRepresentation.cxx#L935-L949) on any single rep also releases the shared actor's GPU resources for the whole view, so deleting one node forces a re-upload of every remaining widget's buffers.

## Design/API notes

- Per-representation property state is dead in 3D: all reps render through the one shared `vtkProperty`, so property customization silently works in slice views but not 3D, and the claiming rep's `PropertyKeys` win for everyone (nondeterministic set order).
- The render-state cache mostly helps camera-only renders: `UpdateFromMRML` calls [`this->Modified()` unconditionally](https://github.com/Slicer/Slicer/blob/3ba8048b7c71cb538fdb258063a467d9dd61635c/Libs/MRML/DisplayableManager/vtkMRMLInteractionWidgetRepresentation.cxx#L604-L608), so any MRML event reopens the gate and triggers a full instance-array rebuild plus re-copy/`Modified()` of the shared arrays for **all** reps in the renderer (a GPU re-upload of everything while dragging one widget among N). Consider adding the actual inputs (display/displayable node MTimes) to `RenderState` instead.

## Minor

- The Transforms DM view-node branch ([vtkMRMLLinearTransformsDisplayableManager.cxx#L368-L376](https://github.com/Slicer/Slicer/blob/3ba8048b7c71cb538fdb258063a467d9dd61635c/Modules/Loadable/Transforms/MRMLDM/vtkMRMLLinearTransformsDisplayableManager.cxx#L368-L376)) did not get the `GetNeedToRender()`/`NeedToRenderOff()` treatment the PR added at the other two sites, so view events leave the flag set and the next transform event triggers one spurious render per widget.
- [`GetHandleInfoList()`](https://github.com/Slicer/Slicer/blob/3ba8048b7c71cb538fdb258063a467d9dd61635c/Libs/MRML/DisplayableManager/vtkMRMLInteractionWidgetRepresentation.cxx#L2032-L2036) now runs `UpdateCameraState()` + `UpdateHandleToWorldTransform()` (matrix DeepCopy + re-orthogonalization) on every call — that is per mouse-move per widget on the pick path.
- Latent: the instance-array allocation uses the virtual `GetNumberOfHandles()` while the fill loop iterates polydata point counts; a subclass reporting fewer handles than its polydata contains would write out of bounds.
- Pre-existing (in a rewritten constructor, so noting it): the partial-arc rotation-ring branch writes into `ArrowPolyData`/`ArrowOutlinePolyData`, which are null at that point — dead only because `INTERACTION_HANDLE_ROTATION_ARC_DEGREES` is 360.

lassoan:
The changes look good otherwise and the performance improvement will be awesome!

lassoan (approved):
Thank you, it looks good.

PR #9358, opened 2026-08-17
BUG: Keep Draw cursor while the effect is active

xriobe:
There is a bug with the mouse cursor when mixing segmentation Draw effect with other object types. In the segmentation widget, when using the “Draw” tool, the cursor icon changes when hovering over a markup object (or reorientation controls), but the Draw tool is still active. 

Steps to reproduce: 
- Load a volume.
- In Markups module, create for example a Line.
- In Segment Editor module, add a segment.
- Click on Draw effect => the cursor changes to drawing.
- Move the mouse on the line handle => the cursor changes to "hand" shape indicating we can move it.
- Left-click with mouse and drag => the drawing is active but the cursor does not change, and then never goes back to Draw icon.

The cursor should stay in Draw mode until we deselect the effect.

### DIFF
--- a/Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorEffects/SegmentEditorDrawEffect.py
+++ b/Modules/Loadable/Segmentations/EditorEffects/Python/SegmentEditorEffects/SegmentEditorDrawEffect.py
@@ -52,6 +52,16 @@ class SegmentEditorDrawEffect(AbstractScriptedSegmentEditorLabelEffect):
     def setupOptionsFrame(self):
         pass
 
+    def reassertCursor(self, viewWidget):
+        # Markup widgets set the view's cursor directly on hover, independently of this effect.
+        # Since this tool remains the active one throughout (hovering or clicking on an object
+        # does not switch tools), the cursor should not show their icon instead. Doing so
+        # misleadingly suggests that interacting with that object is possible, so we force here
+        # this effect's own cursor.
+        cursor = self.scriptedEffect.createCursor(viewWidget)
+        viewWidget.sliceView().setViewCursor(cursor)
+        viewWidget.sliceView().setDefaultViewCursor(cursor)
+
     def processInteractionEvents(self, callerInteractor, eventId, viewWidget):
         abortEvent = False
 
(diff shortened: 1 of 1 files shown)


xriobe:
@lassoan @cpinter would you mind reviewing this PR as main contributors of the script?

cpinter:
Thanks for the contribution! I like this fix because it is contained in the effect itself and the other objects do not need to have any special consideration due to the segment editor effects.

Would it make sense to add this to other effects as well?

xriobe:
@cpinter The Draw effect is the only one having this bug where the icon does not go back after.
Some of the other effects let the "hand" cursor appear when hovering though. I had a look at their code and it would require a bigger work to homogenize them, since they don't follow the same pattern : not all have a specific python script in that folder, some do not have processInteractionEvents, etc...
I would rather keep this PR simple, especially since i now found this post https://github.com/Slicer/Slicer/pull/8763 from @lassoan indicating: "In the mid/long term, segment editor effects will use VTK widgets for user interactions in views (development is driven by the KitwareEU/trame team) and then direct cursor setting will probably not be necessary anymore.".

xriobe:
@cpinter @lassoan Should we still merge this fix, or you rather wait for kitware's improvements for a full rework later ?

cpinter:
I suppose the solution does not cause flickering, but maybe an if condition before actually reassuring the cursor would help with performance.

Apart from this I don't see any harm in integrating this, but the decision will be of the core maintainers.

xriobe:
There is no getViewCursor or equivalent in qMRMLSliceView to do "if viewWidget.sliceView().getViewCursor() != cursor:", or you were thinking something else?
I don't see flickering even in debug build so it seemed ok for me.

lassoan:
Setting a cursor is already somewhat complex (platform-specific, managed by both Qt and VTK, there is also an override cursor stack in Qt, Slicer widgets manage them, too) and it involves resources (may leak memory or resource handles if not done correctly or done too frequently). So, overwriting the default cursor very frequently would likely cause much more complications at unexpected times and locations than the problem that we are trying to solve.

I'll try to find a more robust alternative.

lassoan:
I've submitted an alternative, cleaner, more general solution in this pull request: https://github.com/Slicer/Slicer/pull/9389

Probably this PR can be closed.

xriobe:
Yes the other PR is better, and more general to all effects.
We can close this one.
Thanks!

PR #9360, opened 2026-08-17
ENH: Scale OBJ mesh coordinates based on unit specified in file header

lassoan:
Some applications (for example, Materialise Mimics and 3-matic) write the length unit of the vertex coordinates into the OBJ file header as a comment, such as "vertex coordinates are measured in units, where 1 unit = 1000.000000 mm". Vertex coordinates are now automatically scaled to millimeters when such a file is loaded. Only "mm" unit is supported; for any other unit or invalid scale value a warning is logged and the coordinates are loaded without unit conversion.

Added test and documentation. For completeness, also amended the documentation with description of how coordinate system is also read from mesh file header.

### DIFF
--- a/Docs/user_guide/data_loading_and_saving.md
+++ b/Docs/user_guide/data_loading_and_saving.md
@@ -150,13 +150,22 @@ Readers may support 2D, 3D, and 4D images of various types, such as scalar, vect
 
 ### Models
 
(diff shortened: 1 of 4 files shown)


pieper (approved):
LGTM 👍 

github had issues so commit check failed at first but now it's okay.

PR #9368, opened 2026-08-22
COMP: Order lazy vtk module generation after launcher configuration

hjmjohnson:
Generating `bin/Python/vtk.py` fails on Windows with the Ninja generator because `SlicerGenerateLazyVtkModule` declares no ordering against the launcher configuration target that writes the settings file `${PYTHON_EXECUTABLE}` depends on. Adds the missing `add_dependencies` edge.

Supersedes the earlier approach in this PR, which registered the TBB and Qt directories explicitly via a new `--dll-directory` option on `GenerateLazyVtkModule.py`. Per @lassoan's review that mechanism already exists in the launcher; only the build order was wrong.

<details>
<summary>Failure</summary>

```
[2246/4351] Generating lazily-loading vtk module
FAILED: bin/Python/vtk.py
  File ".../Utilities/Scripts/GenerateLazyVtkModule.py", line 36, in module_order
    import vtkmodules.all
  File ".../VTK-build/lib/site-packages/vtkmodules/all.py", line 7, in <module>
    from .vtkCommonCore import *
ImportError: DLL load failed while importing vtkCommonCore: The specified module could not be found.
```

Since Python 3.8 the DLLs an extension module depends on are resolved only from
directories registered with `os.add_dll_directory()`; `PATH` is not searched.
`vtkCommon-9.6.dll` imports `tbb12.dll` (confirmed with `dumpbin /dependents`),
and `vtkRenderingQt` needs the Qt 6 runtime.

</details>

<details>
<summary>Why the launcher already covers this</summary>

The `PythonSlicer` launcher that backs `PYTHON_EXECUTABLE` exports `LibraryPaths`,
and the generated `sitecustomize.py` calls `slicer_dll_directories.add()`, which
turns each entry into an `os.add_dll_directory()` call.

| step | file |
|---|---|
| launcher chains to the app settings file | `SuperBuild/python_configure_python_launcher.cmake` — `ADDITIONAL_SETTINGS_FILEPATH_BUILD`, and `LibraryPaths` in `PYTHONLAUNCHER_ADDITIONAL_PATH_ENVVARS_BUILD` |
| Qt bin and every EP library path land in that file | `CMake/SlicerBlockCTKAppLauncherSettings.cmake` — `SLICER_LIBRARY_PATHS_BUILD` |
| entries become `add_dll_directory` calls | `slicer_dll_directories.py`, invoked from `sitecustomize.py` |

Both directories the previous approach named are already in `SLICER_LIBRARY_PATHS_BUILD`,
so no new dependency has to be spelled out — and a future VTK dependency is covered
without another edit here.

</details>

<details>
<summary>Why Visual Studio was unaffected</summary>

`<AppName>LauncherSettings.ini` is written by the launcher configuration target,
which `ctkAppLauncherConfigureForTarget` makes depend on the application
executable, so it runs late. `SlicerGenerateLazyVtkModule` had no ordering
relation to it.

The Visual Studio generator sorts `ALL_BUILD.vcxproj` references by target name,
so `SlicerConfigureLauncher` precedes `SlicerGenerateLazyVtkModule` and MSBuild
always writes the `.ini` first. Ninja is free to choose the other order, and does.

</details>

<details>
<summary>Placement notes</summary>

Two deviations from the literal `add_dependencies(SlicerGenerateLazyVtkModule SlicerConfigureLauncher)`:

- Placed in the top-level `CMakeLists.txt` after `add_subdirectory(${Slicer_APPLICATIONS_DIR} ...)` rather than in `Base/Python/CMakeLists.txt`. `Base` is added before `Applications`, so the launcher target does not exist yet while `Base/Python` is being processed.
- Uses `${Slicer_MAIN_PROJECT_APPLICATION_NAME}ConfigureLauncher`, since a custom application names that target after itself, and it is that application's settings file the Python launcher chains to.

The pre-existing explicit `PYTHONPATH` on the custom command is left alone. Its
comment describes the same race this change closes, so it may now be redundant —
happy to drop it here or in a follow-up.

</details>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1233,6 +1233,17 @@ add_subdirectory(
   ${CMAKE_CURRENT_BINARY_DIR}/Applications
   )
 
+# ${PYTHON_EXECUTABLE} is a launcher that reads its PYTHONPATH and LibraryPaths
+# from ${Slicer_MAIN_PROJECT_APPLICATION_NAME}LauncherSettings.ini, which the
+# launcher configuration target writes only after the application links. State
+# the ordering so generators free to schedule it earlier do not run the vtk
+# module generator against an unwritten settings file.
+if(TARGET SlicerGenerateLazyVtkModule
+    AND TARGET ${Slicer_MAIN_PROJECT_APPLICATION_NAME}ConfigureLauncher)
+  add_dependencies(SlicerGenerateLazyVtkModule
+    ${Slicer_MAIN_PROJECT_APPLICATION_NAME}ConfigureLauncher)
+endif()
+
 #-----------------------------------------------------------------------------
 # Templates
 #-----------------------------------------------------------------------------


hjmjohnson:

lassoan (changes requested):
Thanks for working on this.

The launcher exports `LibraryPaths`, and the generated `sitecustomize.py` calls  `slicer_dll_directories.add()`, which turns each entry into an `os.add_dll_directory()` call. So, there should be no need for an alternative mechanism for adding these dll directories.

The real problem is that `SlicerLauncherSettings.ini` is written at build time by `SlicerConfigureLauncher`, which depends on `SlicerApp-real` and therefore runs after the app links. `SlicerGenerateLazyVtkModule` has no dependencies, so on a clean build it can run anytime.

The VS generator sorts `ALL_BUILD.vcxproj` references by target name, and so `SlicerConfigureLauncher` comes before `SlicerGenerateLazyVtkModule`, therefore MSBuild always writes the .ini first. However, Ninja happens to do things in a different order (which is also valid, as there is no order forced via dependencies). 

Rather than hardcoding the TBB and Qt paths, it would be better to add `add_dependencies(SlicerGenerateLazyVtkModule SlicerConfigureLauncher)`.

hjmjohnson:
Thank you — you are right, and the launcher path is much better than what I had.

I traced it to confirm before rewriting: `PythonSlicer`'s `ADDITIONAL_SETTINGS_FILEPATH_BUILD` chains to `<AppName>LauncherSettings.ini` and lists `LibraryPaths` in its additional path envvars, and `SLICER_LIBRARY_PATHS_BUILD` already carries `QT_BINARY_DIR` on Windows plus every `<ep>_LIBRARY_PATHS_LAUNCHER_BUILD` — so both directories I was naming explicitly were already there, and a future VTK dependency would have needed yet another edit.

Force-pushed with the `--dll-directory` work dropped and the dependency added instead. Two small deviations from the literal one-liner:

- It sits in the top-level `CMakeLists.txt` after `add_subdirectory(${Slicer_APPLICATIONS_DIR} ...)` rather than in `Base/Python/CMakeLists.txt`, since `Base` is processed before `Applications` and the target does not exist yet at that point.
- It names `${Slicer_MAIN_PROJECT_APPLICATION_NAME}ConfigureLauncher`, so a custom application depends on its own launcher — the one whose settings file the Python launcher actually chains to.

One follow-up question: the explicit `PYTHONPATH` on the custom command is justified by a comment describing this same race ("the launcher settings of `${PYTHON_EXECUTABLE}` may not reference the VTK build tree yet when this rule runs"). With the ordering stated it looks redundant. Drop it here, or leave it?

Re-verifying on the Windows + Ninja tree that reproduced the original failure; will confirm before marking ready.

hjmjohnson:
Verified on Windows at `77f9bfb360`. The ordering fix works, and it is the better fix — thanks @lassoan.

## Targeted test

Deleted **both** generated artifacts first, so ninja had to prove the ordering rather than succeed on a stale file:

```
rm <build>/SlicerLauncherSettings.ini
rm <build>/bin/Python/vtk.py
ninja SlicerGenerateLazyVtkModule      # exit 0
```

```
[319/321] Linking CXX executable bin\SlicerApp-real.exe
[321/321] Generating lazily-loading vtk module        <- last, after the launcher
```

Ordering confirmed by timestamp, not by log position:

| file | written |
|---|---|
| `SlicerLauncherSettings.ini` | 21:17:27.722 |
| `bin/Python/vtk.py` | 21:17:41.106 |

The `.ini` lands 13 seconds first. No `ImportError: DLL load failed while importing vtkCommonCore`.

## Full build

```
ninja        -> exit 0,  [2255/2255],  0 FAILED,  0 ImportError
```

Artifacts: `SlicerApp-real.exe` 199,680 · `bin/Python/vtk.py` 191,449 · `SlicerLauncherSettings.ini` 3,812.

## The mechanism, confirmed

`LibraryPaths` in the generated settings does carry exactly the directories that were missing:

```
2\path=C:/Qt/6.11.2/msvc2022_64/bin
8\path=C:/M/S/tbb-install/bin
```

which the launcher's `sitecustomize.py` turns into `os.add_dll_directory()` calls. So the earlier revision of this PR — passing `--dll-directory` explicitly to the generator — was re-deriving paths Slicer already computes. Eleven lines stating an ordering relation is the right shape; I've dropped the other approach.

## Test-setup notes for anyone reproducing

Two things that would otherwise produce a misleading result:

- **`SlicerLauncherSettings.ini` is at the Slicer-build root**, not in `bin/` beside `vtk.py`. Deleting the wrong path makes this a no-op test that passes regardless of ordering.
- **`77f9bfb360` alone does not build on Windows/Ninja.** Branched off `main`, it lacks #9367, #9369 and #9370; the build stops at `SlicerApp.rc` well before the lazy-vtk step. I tested it as *those fixes + this one*, with the superseded `--dll-directory` commit reverted so this change is genuinely what is under test.

As expected, `SlicerConfigureLauncher` depends on the application executable, so the targeted `ninja SlicerGenerateLazyVtkModule` pulled in 321 steps including a relink of `SlicerApp-real.exe`.

Environment: Windows 11, MSVC 19.38.33130 (VS 2022 17.8), CMake 4.4.2, Ninja, Qt 6.11.2 `msvc2022_64`, Python 3.12.10, Release.

lassoan (approved):
Looks good, thank you!

PR #9373, opened 2026-08-24
BUG: Report the cause of extension package download failures

lassoan:
downloadExtensionByName() returned nullptr without logging anything in four failure paths, and continued execution instead of returning when extension metadata was missing. Each path now reports what went wrong, including the error string from qRestAPI for failed requests.


hjmjohnson (approved):

ebrahimebrahim:
(TY for this! It always bugged me how sparse the messaging was around download failures)

PR #9377, opened 2026-08-31
ENH: Use application display name for macOS pre-warm dialog

jamesobutler:
This is a follow-up to https://github.com/Slicer/Slicer/commit/5bd6bde20360056db0145f8b845f3e21ed008690. For 3D Slicer custom applications, the application display name should be used for the macOS pre-warm dialog instead of hardcoding to be "3D Slicer".

This is the dialog that was hardcoding "3D Slicer" instead of the 3D Slicer custom application display name:
wrong-name-macos-startup


pieper (approved):
LGTM 👍

PR #9380, opened 2026-08-31
ENH: Use application display name for macOS bundle

jamesobutler:
The application name in the macOS menu bar will now use the application display name such as "3D Slicer" which matches the already used application display name in the window title area.

Current observation with `main` code where the application display name is not in use in the menu bar:
image

With this PR, both places (menubar and window title) appropriately use the application display name.

Note that I originally observed this issue while working on my 3D Slicer custom application where I observed it wasn't using the application display name in the menu bar location.


pieper (approved):
LGTM 👍

Issue #9382, opened 2026-09-04
Different vtk object now being returned for plot view chart


jamesobutler:
## Summary

There appears to be something wrong with the VTK python wrapping resulting in a different vtk object type being returned in certain situations. I discovered this while working to update my 3D Slicer custom application to a newer Slicer version and ran into a traceback error trying to use a object type that has now changed.

I have narrowed down the change in behavior coming between the following commits:

[3D Slicer 5.13.0-2026-08-06 Windows installer](https://slicer-packages.kitware.com/api/v1/file/hashsum/SHA512/052d40e3f9022cfed157cf00d60735b15376a7fc94acb6a3e3356dff9cd30734c34ef7a71e1df019a0b4e22097d488cba9cc9ff995f4264487879541b75771a2/download) aka 7d4b043 ✔️ 

[3D Slicer 5.13.0-2026-08-07 Windows installer](https://slicer-packages.kitware.com/api/v1/file/hashsum/SHA512/105715a60ffd38af71883cf0d411a4f89f8d67d06806b5d80f26b00eb16f77154e21eae4332a4df0a739743db8e8b68b71343997dd21f3719180b2aaa6e7386d/download) aka 7191c39 ❌ 

I suspect VTK or vtkAddon related updates from https://github.com/Slicer/Slicer/commit/054379095d6dcc52e8211102c1c623b267184a0f or https://github.com/Slicer/Slicer/commit/bb457fc76d1ca2d4e391d80ee6775d63cde507de. (cc: @pieper @lassoan )

## Steps to reproduce

Execute the following script upon opening 3D Slicer
```python
slicer.app.layoutManager().setLayout(slicer.vtkMRMLLayoutNode.SlicerLayoutFourUpPlotView)
plot_widget = slicer.app.layoutManager().plotWidget(0)
plot_view = plot_widget.plotView()
print(plot_view.chart().GetLegend())
```

When it was working:
```python
>>> 
slicer.app.layoutManager().setLayout(slicer.vtkMRMLLayoutNode.SlicerLayoutFourUpPlotView)
plot_widget = slicer.app.layoutManager().plotWidget(0)
plot_view = plot_widget.plotView()
print(plot_view.chart().GetLegend())
vtkChartLegend (000002005B84F160)
  Debug: Off
  Modified Time: 221678
  Reference Count: 3
  Registered Events: (none)
  Visible: 0
  Interactive: 1
  Scene: 0000020056EA6E80
  Parent: 000002005B50BE30
[...]

PR #9388, opened 2026-09-07
BUG: Wrap objects of lazily imported VTK classes as their actual class

lassoan:
> [!NOTE]
> **Draft pull request.** A prerequisite is to merge https://github.com/Slicer/VTK/pull/65 into the `slicer-v9.6.2-2026-05-15-f49a1dbaf` branch of Slicer/VTK and then update the VTK git hash in `SuperBuild/External_VTK.cmake`.

This is a fix for #9382

Since now VTK Python modules are loaded lazily (#9306, #9326), only the modules providing base classes of the wrapped Slicer classes are imported at startup. This is a problem, because for example a method in a PythonQt class may return a VTK object before the corresponding VTK module is imported. If VTK's Python wrapper does not find the exact class then it silently returns a nearest already-imported base class. For example `plotView().chart()` returned a `vtkContextItem` instead of a `vtkChartXY`, so `GetLegend()` raised `AttributeError`, and `vtk.vtkRenderer()` created a `vtkRenderer` instead of a `vtkOpenGLRenderer`. VTK's Python wrapper also cached the base class alias, so that importing the module afterwards did not repair the class for the rest of the session (`vtkmodules.vtkChartsCore.vtkChartXY` became `vtkContextItem`).

Fix: The VTK wrapping now searches a class-to-module table before falling back to a base class, and repairs a cached alias when the class gets registered (Slicer/VTK#65). On the Slicer side, the generated lazy `vtk` module hands the attribute index it already computes at build time to `vtkmodules.register_class_modules`, so the module wrapping a class is imported on first use.

The PR also contains a style improvement: the GenerateLazyVtkModule.py module contained an embedded Python script, which made the code harder to review (the .py file was not in the file system and the placeholder looked like a Python format field). Switched to the convention used throughout Slicer source code (template file uses `.py.in` file extension, the `.py` file is generated during build; use `@placeholder@` syntax instead of `{placeholder}`).

Tested on Windows:
- Reporter's script from #9382 now prints a `vtkChartLegend`; `renderWindow()` and renderers come back as their OpenGL classes.
- `py_SlicerLazyVtkModuleTest` passes, and fails with 2 failures when the table is cleared, so it detects the regression.
- `py_nomainwindow_SlicerSceneObserverTest`, `py_standalonepython_import_slicer`, `py_nowarning_mainwindowTest` pass.
- The generated `vtk.py` is the same as before (except comments)


jamesobutler:
For testing I was running into issues related to VTK version functionality because your repository fork (containing https://github.com/lassoan/VTK/tree/lazy-class-module-import-vtk96) doesn't have all the necessary git tags of the versions which are needed. I pushed a copy to https://github.com/Slicer/VTK/tree/lazy-class-module-import-vtk96 and was able to test successfully. 

With this change I was also able to confirm the returned object type is now working as expected using my script provided in the original issue. I also re-ran my automated testing that originally identified the issue in my 3D Slicer custom application, and my tests are now passing as expected. I have gone ahead and given the https://github.com/Slicer/VTK/pull/65 an approval.

jamesobutler (approved):

jamesobutler:
@lassoan Do you want to make this PR as "Ready for review" to allow integration or wait on a review from @pieper on the topic?

lassoan:
I would merge this later in the afternoon unless @pieper suggests any changes.

pieper:
I won't have time to test, so I say go ahead and we'll see what happens. Better to test for a fix than leave it broken longer.

Issue #7560, opened 2024-01-25
MIP volume rendering MIP does not appear if shadows visibility is enabled


lassoan:
## Summary

Volume rendering MIP (minimum/maximum intensity projection) is incompatible with shadows option. The volume does not appear if "Shadows visibility" option is enabled in the 3D view.

## Environment
- Slicer version: Slicer-5.7.0 2024-01-25
- Operating system: all

lassoan:
We should document this and probably disable shadows by default and turn off shadows visibility if MIP is enabled (or display a warning message).

lassoan:
It seems that the volume does not appear due to shader compilation problem:

    error C1503: undefined variable "shading_gradient"

Probably the best would be to fix this issue in VTK (e.g., do not try to compute shadows in MIP rendering).

jcfr:
cc: @LucasGandel

jcfr:
When MIP is enabled, we should disable the shadow support.

LucasGandel:
Volume shading is explicitly disabled by VTK in vtkVolumeShaderComposer.h except for the following blend modes, but we need shading to compute SSAO.

```
  // Shading for composite blending only
  int const shadeReqd = volProperty->GetShade() &&
    (mapper->GetBlendMode() == vtkVolumeMapper::COMPOSITE_BLEND ||
      mapper->GetBlendMode() == vtkVolumeMapper::ISOSURFACE_BLEND);
```

Ideally this should be fixed in VTK by checking the volume blend mode in vtkSSAOPass before adding shader replacement. Requires VTK::RenderingOpenGL2 to depend on VTK::RenderingVolume.

PR #9392, opened 2026-09-08
BUG: Preserve render-pass lists during MIP initialization

ebrahimebrahim:
MIP and MinIP volume rendering can leak render passes and their graphics resources, causing Slicer to exit with a leak error.

VTK emits `ModifiedEvent` while creating the empty `RenderPasses` list, before `Append` inserts the pass. Slicer’s callback removes that empty list even when it has not removed an SSAO pass, invalidating the append and leaving references unbalanced.

Only remove an empty list when the callback actually removed an SSAO pass.

Adds a regression test that failed before and passed after this fix.

Confirmed that other volume rendering tests pass.

Related to #7560 (which introduced the callback being corrected here)

### DIFF
--- a/Modules/Loadable/VolumeRendering/MRMLDM/vtkMRMLVolumeRenderingDisplayableManager.cxx
+++ b/Modules/Loadable/VolumeRendering/MRMLDM/vtkMRMLVolumeRenderingDisplayableManager.cxx
@@ -171,17 +171,20 @@ public:
       if (pipeline->VolumeActorInformation->Has(vtkOpenGLRenderPass::RenderPasses()))
       {
         // Iterate on the OpenGL render passes to remove the SSAO pass if attached
+        bool removedSSAOPass = false;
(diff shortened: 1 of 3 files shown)


jamesobutler:
I should try this in the context of my own Slicer custom app because I have for about a year been reverting https://github.com/Slicer/Slicer/commit/e8aece489b7da5c09bebe277b046042f74c0d497 and https://github.com/Slicer/Slicer/commit/76db15f54448c1b9c8c09cd049cfc800735d0a48 because I was observing debug leaks on close as well in my usage of the app.

jamesobutler (approved):
I locally could confirm that the test case could replicate the issue and that the issue resolved it. I additionally ran tests in the context of my Slicer custom app and I can confirm through the run of my own automated tests for my app that I don't appear to be observing the vtkDebugLeaks as observed on exit so I no longer will require the revert commits for the special Slicer branch that I use.

ebrahimebrahim:
Hooray! Thanks for testing!

This was also an issue I had in my custom app

Issue #6919, opened 2023-03-31
Update or improve them to support rendering of all admonition types


jcfr:
That said the icon are not rendered as expected:

| ![image](https://user-images.githubusercontent.com/219043/228971003-5118c592-42e7-4562-8b33-7a72e7d83b50.png) | ![image](https://user-images.githubusercontent.com/219043/228971139-6302af00-6b8a-4112-abd7-0ca31fdd3a06.png) |
|--|--|
| https://myst-parser.readthedocs.io/en/latest/syntax/admonitions.html#admonitions | |

_Originally posted by @jcfr in https://github.com/Slicer/Slicer/issues/6918#issuecomment-1491003360_

jcfr:
This is likely explained by the implementation of the style in `sphinx_rtd_theme`

See https://github.com/readthedocs/sphinx_rtd_theme/blob/1.2.0/src/sass/_theme_rst.sass#L126-L153 where `hint` and `tip` are associated with the same class.

```sass
  // These are the various note pullouts that sphinx applies
  .note, .attention, .caution, .danger, .error, .hint, .important, .tip, .warning, .seealso, .admonition-todo, .admonition
    @extend .wy-alert
    // These block elements are unsuitable for having floating objects next
    // to them
    clear: both
    .last, > *:last-child
      margin-bottom: 0
  .admonition-title
    @extend .wy-alert-title
    @extend .fa
    @extend .fa-exclamation-circle
    &:before
      margin-right: 4px
  .note, .seealso
    @extend .wy-alert.wy-alert-info
  .hint, .tip, .important
    @extend .wy-alert.wy-alert-success
  .error, .danger
    @extend .wy-alert.wy-alert-danger
  .warning, .caution, .attention, .admonition-todo
    @extend .wy-alert.wy-alert-warning
```

We should either improve the theme adapting changes done in other theme:
* https://github.com/pydata/pydata-sphinx-theme/pull/183

PR #9396, opened 2026-09-14
DOC: Render a distinct icon for each admonition type

aaronjji:
## Overview

`sphinx_rtd_theme` styles all admonitions through a single rule:

```css
.fa-exclamation-circle:before, .rst-content .admonition-title:before { content: "\f06a" }
```

so `note`, `tip`, `hint`, `important`, `warning`, `caution`, `danger` and `seealso` all render
the same exclamation-circle glyph and are told apart only by background color. This is the
issue reported in #6919.

The current documentation contains **186 admonitions** across the user and developer guides
(86 note, 54 warning, 36 tip, 7 hint, 3 important), all of which render identically today.

## Changes

A single file is touched, `Docs/_static/css/custom.css`, which is already registered through
`html_css_files` in `Docs/conf.py` and loads after `theme.css`. Per-type `:before` overrides are
added using glyphs from the FontAwesome set the theme already ships, so there is no new
dependency, no new asset, and no change to the build.

| Admonition | Icon | Codepoint |
|---|---|---|
| `note`, `seealso` | info-circle | `\f05a` |
| `tip`, `hint` | lightbulb-o | `\f0eb` |
| `important` | exclamation-circle | `\f06a` (unchanged) |
| `warning`, `caution`, `attention` | exclamation-triangle | `\f071` |
| `danger`, `error` | times-circle | `\f057` |
| `admonition-todo` | check-square-o | `\f046` |

Admonitions with a custom title (plain `.admonition`) are left on the theme default.

This also improves GitHub-style alerts (`> [!TIP]`), which `Docs/_ext/github_alerts.py`
converts into these same admonition types.

## Before / after

Admonitions as rendered on `developer_guide/build_instructions/linux.md`, which uses every
admonition type currently present in the documentation.

**Before** — `note`, `tip`, `hint`, `important` and `warning` all render the same
exclamation-circle glyph:

admonitions-before

**After** — each type renders its own icon:

admonitions-after

## Testing

Built locally with `sphinx-build` against `requirements-docs.txt` (Sphinx 9.0.4,
sphinx_rtd_theme 3.1.0). Verified via computed style in a headless browser that each type
resolves to its intended glyph:

```
before -> note/tip/hint/important/warning/caution/danger/seealso : all U+f06a
after  -> note U+f05a | tip U+f0eb | hint U+f0eb | important U+f06a
          warning U+f071 | caution U+f071 | danger U+f057 | seealso U+f05a
```

Note the overrides use CSS escape sequences rather than literal characters, so they are not
sensitive to how the stylesheet's encoding is interpreted.

Also verified against the Read the Docs build for this PR, which builds in its own
environment rather than the local one:

* all five types resolve to the intended codepoints there,
* the page loads the theme's bundled `fontawesome-webfont.woff2` (FontAwesome 4.x), and
* each codepoint measures a font-specific advance width rather than falling back, so the
  glyphs are present and do not render as missing-character boxes.

Live preview of this PR:
<https://slicer--9396.org.readthedocs.build/en/9396/developer_guide/build_instructions/linux.html>


lassoan (approved):
Thank you, this is a nice improvement.

PR #9399, opened 2026-09-15
Add storage of volume and transform sequences in Nifti format

lassoan:
Despite many limitations and inconveniences of Nifti, it is used too widely to be ignored, and adding better support for Nifti now takes less manual effort. Therefore in this PR adds support for storage of volume and grid transform sequences in Nifti format. This PR also make loading of 4D images and unsupported files more robust and shows more meaningful error messages to users.

Details:
- **Nifti volume and transform sequences:** 4D Nifti images can loaded as volume sequences, 5D Nifti displacement images can be loaded as transform sequences; volume and grid transform sequences can be saved in Nifti format. Sequence index name, unit, and values are set from the Nifti time/spectral unit, spacing, and offset. A warning is reported if some sequence properties cannot be losslessly stored in Nifti. Also fixes saving color images with non-unsigned-char components and spatial vector sequences in Nifti format. Tests are added to cover the newly supported file format variants.
- **Improve messages captured from VTK objects:** messages captured by `vtkMRMLMessageCollection::SetObservedObject` show the message text first, followed by the source code location and object description in square brackets. Warnings and errors of the legacy `.vtk` model writer are now captured.
- **Report why a volume or segmentation file cannot be loaded:** reader errors are displayed to users when volume loading fails. NIfTI-2 files (not supported by ITK) and 4D images loaded as volume or segmentation (which were previously loaded misaligned, using only the first frame) are reported with a clear explanation.
- **Volume sequence geometry and 4D image detection:** volume sequences with left-handed IJK coordinate system are normalized the same way as scalar volumes (`ForceRightHandedIJKCoordinateSystem`, enabled by default, can be disabled using the default storage node). 4D Nifti images and 4D NRRD images without list axis are loaded as sequences by default.


pieper:
Thanks for working on this.  I agree we should do a best-efforts implementation to support things users actually want to do (even if we think it's ill-advised).  I do wish there was some official documentation of the format.  We may want to have another import path that gives people options about how to interpret any ambiguous files.

PR #9400, opened 2026-09-17
BUG: Deleted data can be loaded from database

### LINKED ISSUE #7290: Can load removed data from database

xriobe:
This fixes https://github.com/Slicer/Slicer/issues/7290
@pieper @lassoan


lassoan (commented):
Thanks for working on this. I've checked this implementation and it did not capture the root cause of the error (no selection changes signal emitted when selection is changed due to removing data from the database) and it had undesirable side effects (all loadables discarded just because some unrelated patients/studies/series were removed from the database). Instead of this change, I've submitted a fix to CTK: https://github.com/commontk/CTK/pull/1455

lassoan:
Closing this, a fix is submitted to CTK instead. Thanks for taking up this challenge @xriobe - even though the proposed changes were not necessary this time, I hope you'll keep contributing to Slicer.

xriobe:
Thanks for your fix!

PR #9403, opened 2026-09-21
BUG: Fix sequence browser playback running at ~2x the requested rate

### LINKED ISSUE #9361: Sequence browser playback runs at ~2x the requested PlaybackRateFps
> **Note on provenance:** this issue was written and filed by Claude, an AI assistant, at the
> explicit request of @pieper and reporting what he asked me to report. We found this while
> comparing Slicer's sequence playback against an independent reimplementation of the Sequences
> playback model. The measurements below are ones I ran against Slicer 5.12.0; the analysis is
> mine and should be checked rather than taken on trust.

## Summary

`vtkMRMLSequenceBrowserNode` playback advances at approximately **twice** the requested
`PlaybackRateFps`. A sequence set to 1 fps advances roughly every 0.5 s.

## Steps to reproduce

Load any volume sequence, then in the Python console:

```python
import slicer, time
br = slicer.util.getNodesByClass("vtkMRMLSequenceBrowserNode")[0]
br.SetPlaybackLooped(True)
br.SetPlaybackItemSkippingEnabled(True)

for fps in (0.5, 1.0, 2.0, 5.0, 10.0):
    br.SetPlaybackActive(False); slicer.app.processEvents()
    br.SetPlaybackRateFps(fps); br.SetSelectedItemNumber(0)
    br.SetPlaybackActive(True); slicer.app.processEvents()
    dur = 12.0 if fps <= 1 else 8.0
    marks = []; t0 = time.time(); last = br.GetSelectedItemNumber()
    while time.time() - t0 < dur:
        slicer.app.processEvents()
        c = br.GetSelectedItemNumber()
        if c != last:
            marks.append(time.time() - t0); last = c
    br.SetPlaybackActive(False)
    print(f"requested {fps:5.1f} fps -> measured {len(marks)/dur:6.2f} fps"
          f"  (ratio {(len(marks)/dur)/fps:.2f})")
```

## Observed

| requested | measured | ratio |
|---|---|---|
| 0.5 fps | 0.92 | 1.83× |
| 1 fps | 1.92 | 1.92× |
| 2 fps | 3.75 | 1.88× |
| 5 fps | 9.38 | 1.88× |
| 10 fps | 16.88 | 1.69× |

At 1 fps the median gap between phase advances was 0.508 s rather than 1.000 s.

## Expected

Advancing at the requested rate — 1.0 s between items at 1 fps.

## Cause

[`Modules/Loadable/Sequences/Logic/vtkSlicerSequencesLogic.cxx#L266-L278`](https://github.com/Slicer/Slicer/blob/main/Modules/Loadable/Sequences/Logic/vtkSlicerSequencesLogic.cxx#L266-L278):

```cpp
double elapsedTimeSec = updateStartTimeSec - this->LastSequenceBrowserUpdateTimeSec[browserNode];
int selectionIncrement = floor(elapsedTimeSec * browserNode->GetPlaybackRateFps() + 0.5); // floor with +0.5 is rounding
if (selectionIncrement > 0)
{
  this->LastSequenceBrowserUpdateTimeSec[browserNode] = updateStartTimeSec;
  ...
  browserNode->SelectNextItem(selectionIncrement);
}
```

`floor(x + 0.5)` is round-to-nearest, so `selectionIncrement` first reaches 1 when
`elapsedTimeSec >= 0.5 / fps` — **half** the nominal period. The timer is then reset to
`updateStartTimeSec`, so the next interval starts from zero again and the steady-state period
stays at half the requested one.

The rounding would be correct if `elapsedTimeSec` were measured from a fixed origin, because
the errors would cancel over time. Combined with resetting the reference time on every advance,
it halves the period instead.

The residual deviation from a clean 2× appears to be the 50 Hz update poll
(`UPDATE_VIRTUAL_OUTPUT_NODES_PERIOD_SEC = 0.020`) quantising the half-period: at 10 fps the
half-period is 0.05 s but the poll granularity is 0.02 s, so it fires at ~0.06 s and the ratio
falls to 1.69.

## Suggested fix

Keep the fractional remainder instead of discarding it, so the long-run rate is exact:

```cpp
double fps = browserNode->GetPlaybackRateFps();
int selectionIncrement = floor(elapsedTimeSec * fps);      // no +0.5
if (selectionIncrement > 0)
{
  // advance the reference time by exactly the items consumed, keeping the remainder
  this->LastSequenceBrowserUpdateTimeSec[browserNode] += selectionIncrement / fps;
  ...
}
```

For comparison, an equivalent phase accumulator (`acc += dt * fps; inc = floor(acc); acc -= inc`)
measured 1.000 fps for a 1 fps request, with individual gaps of 0.998–1.001 s.

Two things worth a maintainer's judgement, since they affect whether this is safe to change:

- Some users may have adapted to the current behaviour (choosing 5 fps to get 10).
- `PlaybackItemSkippingEnabled = false` forces `selectionIncrement = 1`, so playback there is
  capped by the poll rate as well as by this bias; the two interact.

## Environment

- Slicer 5.12.0 (revision 34621), macOS, x86_64 build under Rosetta on Apple M4
- Reproduced with a 10-frame volume sequence loaded from `.seq.nrrd`
- The relevant line is unchanged on `main` as of filing

mmustafasenoglu:
Fixes #9361

The increment used round-to-nearest, so it fired after half the nominal period, and resetting the reference clock on every advance kept it there. Truncate instead and carry the remainder forward.

Changes:
- Modules/Loadable/Sequences/Logic/vtkSlicerSequencesLogic.cxx: truncate increment, advance reference time by consumed items


jamesobutler (approved):
Using CT Cardio Sequence from the Sample Data module and the reproducible script of #9361.

With `main`:
```
requested   0.5 fps -> measured   0.92 fps  (ratio 1.83)
requested   1.0 fps -> measured   1.92 fps  (ratio 1.92)
requested   2.0 fps -> measured   3.75 fps  (ratio 1.88)
requested   5.0 fps -> measured   8.00 fps  (ratio 1.60)
requested  10.0 fps -> measured  15.12 fps  (ratio 1.51)
```

With this PR:
```
requested   0.5 fps -> measured   0.42 fps  (ratio 0.83)
requested   1.0 fps -> measured   1.00 fps  (ratio 1.00)
requested   2.0 fps -> measured   1.88 fps  (ratio 0.94)
requested   5.0 fps -> measured   5.00 fps  (ratio 1.00)
requested  10.0 fps -> measured  10.00 fps  (ratio 1.00)
```

